PR #5172, opened 2022-07-01
MD team policy common functions refactor

ldh4:
Refactoring MD team policies interfaces to have them be templated on the team member type and be available for all backends. This is to avoid redundancy in each backend impls when the template specialized implementations of these functions are essentially identical.

PR #4812, opened 2022-02-21
MD Team Policies Implementation for serial

ldh4:
Motivation: #3385 

Implementation for MD TeamPolicy for Serial backend. This includes infrastructures and tests needed for all backends.
This adds parallel_for and parallel_reduce support for MDTeamThreadRange, MDThreadVectorRange and MDTeamVectorRange.

Added in directional supports:
MDTeamThreadRange is single directional
MDThreadVectorRange and MDTeamVectorRange have outer and inner directions

@nliber

PR #5182, opened 2022-07-06
Revert initial md team range policy impl

ldh4:
Reverts #5172, #5164 and #4812.


masterleinad (commented):
Did we want to keep the tests?

masterleinad (approved):

Issue #5071, opened 2022-05-30
[do not close] NGA/task 3: tracking sorting task


fnrizzi:
- [ ] 3.1 Make top level `Kokkos::sort(Kokkos::View)` map toThrust/RocThrust/std::sort/OneAPI as appropriate
   (@jstrzebonski)
- [ ] 3.2 Explore generic implementation of RadixSort and BitonicSort (@PhilMiller)
   - report on potential issues in API design due to need of temporary allocations etc.
   - implement at top level and compare to Thrust/RocThrust 
   - implementation working at Team and Potential Thread level in hierarchical parallelism
   - will potentially down-select optimal algorithm
- [ ] 3.3 (Stretch) Integrate into mainline Kokkos as appropriate with tests (assign to: TBD)

PhilMiller:
Here's a rough list of open issues in Kokkos relating to sorting algorithms and their use cases, to inform 3.2
* #645 
* #1931
* #2914
* #3089
* #3285 
* #4487 
* #4978

PhilMiller:
Possibly related issues in Kokkos Kernels:
* https://github.com/kokkos/kokkos-kernels/issues/10
* https://github.com/kokkos/kokkos-kernels/issues/473

There's various code in Kokkos and Kokkos Kernels that implement various sorting and sorting-adjacent functionality

PhilMiller:
A thought inspired by the Python standard library's sorting function:

It can be quite valuable to distinguish "compute the comparison key of element `i`" from "compare (the keys of) elements `i` and `j`"

In at least some of our prospective use cases, this would potentially let us still use an integer sorting algorithm (like Radix Sort) on non-integer values. E.g. in the bin-particles-by-cell case, the notional key function returns an integer cell index

PhilMiller:
A couple of earlier issues mention interfaces like `sort2` and `sort3` that sort several parallel arrays based on the ordering of one that provides the keys. Here in C++-land, for use cases that don't need to get back the permutation, there's no reason we can't have `sort(Comparator, KeyArray, ValueArrays...)` for an arbitrary number of parallel arrays to be sorted.

Hypothetically, this could even be implemented a bit more efficiently than permuting each array one by one, since we can re-use the loads of the permutation or even re-compute the sort of the key array, depending on pragmatics of the operation in question.

PhilMiller:
[...]

PR #5183, opened 2022-07-07
Dispatch Kokkos::sort(Kokkos::View) to CUDA Thrust

jstrzebonski:

### DIFF
--- a/algorithms/src/Kokkos_Sort.hpp
+++ b/algorithms/src/Kokkos_Sort.hpp
@@ -57,8 +57,8 @@
 
 #if defined(KOKKOS_ENABLE_CUDA)
 
-// Workaround for `Instruction 'shfl' without '.sync' is not supported on .target
-// sm_70 and higher from PTX ISA version 6.4`.
+// Workaround for `Instruction 'shfl' without '.sync' is not supported on
+// .target sm_70 and higher from PTX ISA version 6.4`.
 // Also see https://github.com/NVIDIA/cub/pull/170.
 #if ((__CUDACC_VER_MAJOR__ >= 9) || defined(__NVCOMPILER_CUDA__) || \
      CUDA_VERSION >= 9000) &&                                       \
--- a/algorithms/src/Kokkos_Sort.hpp
+++ b/algorithms/src/Kokkos_Sort.hpp
@@ -629,10 +629,9 @@ sort(const ExecutionSpace& exec,
 }
 
 #if defined(KOKKOS_ENABLE_CUDA)
-template <class ViewType>
-std::enable_if_t<
-    Experimental::Impl::has_random_access_iterator<ViewType>::value>
-sort(Cuda const& space, ViewType const& view, bool const = false) {
+template <class DataType, class... Properties>
+void sort(const Cuda& space, const Kokkos::View<DataType, Properties...>& view,
+          bool const = false) {
   const auto exec = thrust::cuda::par.on(space.cuda_stream());
   auto first      = Experimental::begin(view);
   auto last       = Experimental::end(view);
--- a/algorithms/src/Kokkos_Sort.hpp
+++ b/algorithms/src/Kokkos_Sort.hpp
@@ -60,9 +60,7 @@
 // Workaround for `Instruction 'shfl' without '.sync' is not supported on
 // .target sm_70 and higher from PTX ISA version 6.4`.
 // Also see https://github.com/NVIDIA/cub/pull/170.
-#if ((__CUDACC_VER_MAJOR__ >= 9) || defined(__NVCOMPILER_CUDA__) || \
-     CUDA_VERSION >= 9000) &&                                       \
-    !defined(CUB_USE_COOPERATIVE_GROUPS)
+#if !defined(CUB_USE_COOPERATIVE_GROUPS)
 #define CUB_USE_COOPERATIVE_GROUPS
 #endif
 
--- a/algorithms/src/Kokkos_Sort.hpp
+++ b/algorithms/src/Kokkos_Sort.hpp
@@ -630,13 +630,12 @@ sort(const ExecutionSpace& exec,
 
 #if defined(KOKKOS_ENABLE_CUDA)
 template <class ViewType>
-void sort(Cuda const& space, ViewType const& view) {
-  static_assert(ViewType::rank == 1, "CUDA sort only supports 1D Views.");
-
-  using DevicePtr = thrust::device_ptr<typename ViewType::value_type>;
-  DevicePtr first(view.data());
-  DevicePtr last(view.data() + view.extent(0));
+std::enable_if_t<
+    Experimental::Impl::has_random_access_iterator<ViewType>::value>
+sort(Cuda const& space, ViewType const& view, bool const = false) {
   const auto exec = thrust::cuda::par.on(space.cuda_stream());
+  auto first      = Experimental::begin(view);
+  auto last       = Experimental::end(view);
   thrust::sort(exec, first, last);
 }
 #endif
--- a/algorithms/src/Kokkos_Sort.hpp
+++ b/algorithms/src/Kokkos_Sort.hpp
@@ -63,8 +63,14 @@
 #define CUB_USE_COOPERATIVE_GROUPS
 #endif
 
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wshadow"
+
 #include <thrust/device_ptr.h>
 #include <thrust/sort.h>
+
+#pragma GCC diagnostic pop
+
 #endif
 
 namespace Kokkos {
--- a/algorithms/src/Kokkos_Sort.hpp
+++ b/algorithms/src/Kokkos_Sort.hpp
@@ -51,9 +51,8 @@
 
 #include <Kokkos_Core.hpp>
 #include <Kokkos_NestedSort.hpp>
+#include <std_algorithms/Kokkos_BeginEnd.hpp>
 #include <algorithm>
-
-#include "std_algorithms/Kokkos_BeginEnd.hpp"
 
 #if defined(KOKKOS_ENABLE_CUDA)
 
@@ -636,17 +635,6 @@ sort(const ExecutionSpace& exec,
   bin_sort.sort(exec, view);
 }
 
-#if defined(KOKKOS_ENABLE_CUDA)
-template <class DataType, class... Properties>
-void sort(const Cuda& space, const Kokkos::View<DataType, Properties...>& view,
-          bool const = false) {
-  const auto exec = thrust::cuda::par.on(space.cuda_stream());
-  auto first      = Experimental::begin(view);
-  auto last       = Experimental::end(view);
-  thrust::sort(exec, first, last);
-}
-#endif
-
 template <class ExecutionSpace, class DataType, class... Properties>
 std::enable_if_t<(Kokkos::is_execution_space<ExecutionSpace>::value) &&
                  (SpaceAccessibility<
@@ -657,6 +645,17 @@ sort(const ExecutionSpace&, const Kokkos::View<DataType, Properties...>& view) {
   auto last  = Experimental::end(view);
   std::sort(first, last);
 }
+
+#if defined(KOKKOS_ENABLE_CUDA)
+template <class DataType, class... Properties>
+void sort(const Cuda& space,
+          const Kokkos::View<DataType, Properties...>& view) {
+  const auto exec = thrust::cuda::par.on(space.cuda_stream());
+  auto first      = Experimental::begin(view);
+  auto last       = Experimental::end(view);
+  thrust::sort(exec, first, last);
+}
+#endif
 
 template <class ViewType>
 void sort(ViewType const& view) {
--- a/algorithms/src/Kokkos_Sort.hpp
+++ b/algorithms/src/Kokkos_Sort.hpp
@@ -583,62 +583,6 @@ struct min_max_functor {
 
 }  // namespace Impl
 
-template <class ExecutionSpace>
-class Sort {
- public:
-  template <class ViewType>
-  static void sort(const ExecutionSpace& exec, ViewType const& view) {
-    using CompType = BinOp1D<ViewType>;
-
-    Kokkos::MinMaxScalar<typename ViewType::non_const_value_type> result;
-    Kokkos::MinMax<typename ViewType::non_const_value_type> reducer(result);
-    parallel_reduce("Kokkos::Sort::FindExtent",
-                    Kokkos::RangePolicy<typename ViewType::execution_space>(
-                        exec, 0, view.extent(0)),
-                    Impl::min_max_functor<ViewType>(view), reducer);
-    if (result.min_val == result.max_val) return;
-    // For integral types the number of bins may be larger than the range
-    // in which case we can exactly have one unique value per bin
-    // and then don't need to sort bins.
-    bool sort_in_bins = true;
-    // TODO: figure out better max_bins then this ...
-    int64_t max_bins = view.extent(0) / 2;
-    if (std::is_integral<typename ViewType::non_const_value_type>::value) {
-      // Cast to double to avoid possible overflow when using integer
-      auto const max_val = static_cast<double>(result.max_val);
-      auto const min_val = static_cast<double>(result.min_val);
-      // using 10M as the cutoff for special behavior (roughly 40MB for the
-      // count array)
-      if ((max_val - min_val) < 10000000) {
-        max_bins     = max_val - min_val + 1;
-        sort_in_bins = false;
-      }
-    }
-    if (std::is_floating_point<
-            typename ViewType::non_const_value_type>::value) {
-      KOKKOS_ASSERT(std::isfinite(static_cast<double>(result.max_val) -
-                                  static_cast<double>(result.min_val)));
-    }
-
-    BinSort<ViewType, CompType> bin_sort(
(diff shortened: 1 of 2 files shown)


algorithms/src/Kokkos_Sort.hpp
```diff
@@ -580,48 +585,71 @@ struct min_max_functor {
...
+};
+
+#if defined(KOKKOS_ENABLE_CUDA)
+template <>
+class Sort<Kokkos::Cuda> {
+ public:
+  template <class ViewType>
+  static void sort(const Kokkos::Cuda&, ViewType const& view) {
+    thrust::device_ptr<typename ViewType::value_type> data_ptr(view.data());
+    thrust::sort(data_ptr, data_ptr + view.extent(0));
```
dalg24, line 639: You need to use the underlying stream from the execution space instance.

algorithms/src/Kokkos_Sort.hpp
```diff
@@ -580,48 +585,71 @@ struct min_max_functor {
...
-  if (std::is_floating_point<typename ViewType::non_const_value_type>::value) {
-    KOKKOS_ASSERT(std::isfinite(static_cast<double>(result.max_val) -
-                                static_cast<double>(result.min_val)));
-  }
-
-  BinSort<ViewType, CompType> bin_sort(
-      view, CompType(max_bins, result.min_val, result.max_val), sort_in_bins);
-  bin_sort.create_permute_vector(exec);
-  bin_sort.sort(exec, view);
+  Sort<ExecutionSpace>::sort(exec, view);
```
dalg24, line 652: There is no point on changing the original function, you can define something like
```C++
template <class ViewType>
void sort(Cuda const& space, ViewType const& view) {
  static_assert(ViewType::rank == 1, "some useful error msg");
  auto const n = view.extent(0);
  using ValueType = typename ViewType::value_type;
  auto first = thrust::device_ptr<ValueType>(view.data());
  auto last = thrust::device_ptr<ValueType>(view.data() + n);
  auto const exec = thrust::cuda::par.on(space.cuda_stream());
  thrust::sort(exec, first, last);
}
```
which will be a better match when used with Cuda.

dalg24 (commented):
You can look at

algorithms/src/Kokkos_Sort.hpp
fnrizzi: is that always a better match or can there be compilers that maybe give an ambiguous template error?

algorithms/src/Kokkos_Sort.hpp

algorithms/src/Kokkos_Sort.hpp

algorithms/src/Kokkos_Sort.hpp
```diff
@@ -624,6 +629,19 @@ std::enable_if_t<Kokkos::is_execution_space<ExecutionSpace>::value> sort(
   bin_sort.sort(exec, view);
 }
 
+#if defined(KOKKOS_ENABLE_CUDA)
+template <class ViewType>
+void sort(Cuda const& space, ViewType const& view) {
+  static_assert(ViewType::rank == 1, "CUDA sort only supports 1D Views.");
```
masterleinad, line 635: It looks like the `View` needs to be contiguous in memory, too, right? 
```suggestion
  static_assert(ViewType::rank == 1 && !std::is_same<typename ViewType::array_layout, LayoutStride>::value, "CUDA sort only supports contiguous 1D Views.");
```

fnrizzi: the static assert has advantage of producing a readable message, the sfinea does not

jstrzebonski: I agree. I was thinking in terms of a "fallback" - if "CUDA" requirements are not met then "original" implementation could be used.

masterleinad: I would use a `static_assert` for the restrictions that apply to all overloads and `std::enable_if`/SFINAE if there is a fallback. We assume for all `Kokkos::sort` overloads that we have a 1D-View but only this overload is restricted to contiguous memory, right?

An alternative to restricting this to contiguous memory would be to use `RandomAccessIterator`.

jstrzebonski: I used `RandomAccessIterator`.

masterleinad:
You will need to fix the indentation.

algorithms/src/Kokkos_Sort.hpp
```diff
@@ -46,9 +46,15 @@
 #define KOKKOS_SORT_HPP_
 
 #include <Kokkos_Core.hpp>
+#include "std_algorithms/Kokkos_BeginEnd.hpp"
```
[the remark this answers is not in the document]
jstrzebonski: I'm using `Experimental::begin()` and `Experimental::end()` to get, respectively, the first and last random access iterator of a given View.

jstrzebonski:
> You will need to fix the indentation.

Done.

masterleinad (approved):
Looks good to me.

algorithms/src/Kokkos_Sort.hpp
```diff
@@ -624,6 +632,18 @@ std::enable_if_t<Kokkos::is_execution_space<ExecutionSpace>::value> sort(
   bin_sort.sort(exec, view);
 }
 
+#if defined(KOKKOS_ENABLE_CUDA)
+template <class ViewType>
+std::enable_if_t<
+    Experimental::Impl::has_random_access_iterator<ViewType>::value>
+sort(Cuda const& space, ViewType const& view, bool const = false) {
```
fnrizzi, line 639: if we know that this should accept a kokkos view, I would suggest doing this: 

```
template <class DataType, class ...Properties>
void sort(Cuda const& space, const Kokkos::View<DataType, Properties...> view, bool const = false)
```

algorithms/src/std_algorithms/impl/Kokkos_RandomAccessIterator.hpp
```diff
@@ -54,11 +56,26 @@ namespace Kokkos {
 namespace Experimental {
 namespace Impl {
 
+template <typename ViewType, typename Enabled = void>
+struct has_random_access_iterator : std::false_type {};
```
fnrizzi, line 60: with me suggested change above, we don't need this and it can stay inside the randomAccessIterator class

algorithms/src/Kokkos_Sort.hpp
fnrizzi: I was referring to explicitly use Kokkos::View as an argument rather than using `ViewType`. Makes sense? 
(I personally agree on the defaulted arg being iffy)

dalg24: My bad.  Agreed.

jstrzebonski: 1. As I already wrote, in this case, I treat SFINAE as a possibility of having a fallback function. Also, I follow Daniel's suggestion of `static_assert` vs `enable_if` ([here](https://github.com/kokkos/kokkos/pull/5183#discussion_r917951595)). But I don't have a strong opinion on this. If you guys find the usage of `static_assert` (hidden inside `begin()` and `end()`) more appropriate, I'm okay with removing `enable_if`.

2. And about the function's signature:
* `bool const = false` - yeah, I don't like it either, but the original function has this default boolean parameter, so I could either go with an unnamed default value or two overloads, one with and the other without bool par.
* `ViewType const& view` vs `const Kokkos::View<DataType, Properties...> view` - original function takes a `ViewType const& view` argument, so I followed that. If `Kokkos::View` is more suitable, I'll change that.

fnrizzi: That makes sense to me if we already had a fallback impl, but there isn't one

jstrzebonski: I removed SFINAE from CUDA sort impl.

dalg24: The fallback is the current `BinSort` implementation and SFINAE is a safer approach.

Assuming someone was sorting a view with a custom layout, he would now get a compile-error instead of keep doing the same thing.

dalg24 (approved):

jstrzebonski:
`CUDA-9.2-NVCC` and `CUDA-11.6-NVCC-DEBUG` builds report similar errors, for example:
```
/usr/local/cuda/bin/../targets/x86_64-linux/include/thrust/system/cuda/detail/cub/device/dispatch/dispatch_radix_sort.cuh:1053:1: error: declaration of 'downsweep_kernel' shadows a member of 'thrust::cuda_cub::cub::DispatchRadixSort<false, int, thrust::cuda_cub::cub::NullType, int>::PassConfig<void (*)(const int*, int*, int, int, int, thrust::cuda_cub::cub::GridEvenShare<int>), void (*)(int*, int), void (*)(const int*, int*, const thrust::cuda_cub::cub::NullType*, thrust::cuda_cub::cub::NullType*, int*, int, int, int, thrust::cuda_cub::cub::GridEvenShare<int>)>' [-Werror=shadow]
         cudaError_t InitPassConfig(
 ^       ~~~~~~
/usr/local/cuda/bin/../targets/x86_64-linux/include/thrust/system/cuda/detail/cub/device/dispatch/dispatch_radix_sort.cuh:1040:18: note: shadowed declaration is here
         DownsweepKernelT        downsweep_kernel;
                  ^~~~~~~~~~~~~~~~
/usr/local/cuda/bin/../targets/x86_64-linux/include/thrust/system/cuda/detail/cub/device/dispatch/dispatch_radix_sort.cuh:1053:1: error: declaration of 'scan_kernel' shadows a member of 'thrust::cuda_cub::cub::DispatchRadixSort<false, int, thrust::cuda_cub::cub::NullType, int>::PassConfig<void (*)(const int*, int*, int, int, int, thrust::cuda_cub::cub::GridEvenShare<int>), void (*)(int*, int), void (*)(const int*, int*, const thrust::cuda_cub::cub::NullType*, thrust::cuda_cub::cub::NullType*, int*, int, int, int, thrust::cuda_cub::cub::GridEvenShare<int>)>' [-Werror=shadow]
         cudaError_t InitPassConfig(
 ^       ~~~~~~
/usr/local/cuda/bin/../targets/x86_64-linux/include/thrust/system/cuda/detail/cub/device/dispatch/dispatch_radix_sort.cuh:1038:13: note: shadowed declaration is here
         ScanKernelT             scan_kernel;
             ^~~~~~~~~~~
/usr/local/cuda/bin/../targets/x86_64-linux/include/thrust/system/cuda/detail/cub/device/dispatch/dispatch_radix_sort.cuh:1053:1: error: declaration of 'upsweep_kernel' shadows a member of 'thrust::cuda_cub::cub::DispatchRadixSort<false, int, thrust::cuda_cub::cub::NullType, int>::PassConfig<void (*)(const int*, int*, int, int, int, thrust::cuda_cub::cub::GridEvenShare<int>), void (*)(int*, int), void (*)(const int*, int*, const thrust::cuda_cub::cub::NullType*, thrust::cuda_cub::cub::NullType*, int*, int, int, int, thrust::cuda_cub::cub::GridEvenShare<int>)>' [-Werror=shadow]
         cudaError_t InitPassConfig(
 ^       ~~~~~~
/usr/local/cuda/bin/../targets/x86_64-linux/include/thrust/system/cuda/detail/cub/device/dispatch/dispatch_radix_sort.cuh:1036:16: note: shadowed declaration is here
         UpsweepKernelT          upsweep_kernel;
                ^~~~~~~~~~~~~~
```

but that are caused directly by a library implementation:
https://github.com/NVIDIA/thrust/blob/cuda-9.2/thrust/system/cuda/detail/cub/device/dispatch/dispatch_radix_sort.cuh#L1036
and
https://github.com/NVIDIA/thrust/blob/cuda-9.2/thrust/system/cuda/detail/cub/device/dispatch/dispatch_radix_sort.cuh#L1054

Can it be somehow ignored?

dalg24:
Thrust header are not treated as system header as they should.

fnrizzi (approved):

jstrzebonski:
> Thrust header are not treated as system header as they should.

I tried
```bash
/kokkos/gnu_generate_makefile.bash --with-options=compiler_warnings --cxxflags="-Werror -isystem /usr/local/cuda/targets/x86_64-linux/include/thrust" --cxxstandard=c++14 --with-cuda --with-cuda-options=enable_lambda --arch=Volta70
```
but that didn't help.

nliber (approved):

algorithms/src/Kokkos_Sort.hpp
```diff
@@ -628,6 +644,17 @@ std::enable_if_t<Kokkos::is_execution_space<ExecutionSpace>::value> sort(
   bin_sort.sort(exec, view);
 }
 
+#if defined(KOKKOS_ENABLE_CUDA)
+template <class DataType, class... Properties>
+void sort(const Cuda& space, const Kokkos::View<DataType, Properties...>& view,
+          bool const = false) {
```
PhilMiller, line 650: Could you give the name for `always_use_kokkos_sort` here?

Is there actually a way to call the non-Thrust implementation through the `Kokkos::sort` overload set on these arguments? It seems that the specialization for `Cuda` as `ExecutionSpace` would always win.

We may need to tweak the template parameters to expect a tag like `enum class SORT_IMPLEMENTATION { default, kokkos, backend_library };` and `enable_if` or `if constexpr` on it.

PhilMiller: Or, at least if `always_use_kokkos_sort` is passed `true`, then call `Kokkos::sort<Cuda, decltype(View)>(space, view, true)` to get our generic implementation?

PhilMiller (changes requested):
See my remarks - feel free to dismiss/ignore if I've misunderstood, but that may indicate the need for clarifying comments to document for both developers and users

algorithms/src/Kokkos_Sort.hpp
dalg24: Why do we have this argument in the first place?

jstrzebonski: Wasn't the idea to have a distinct function for CUDA (and other exec_spaces) View? Without any fallback? If that's not the case, then I'll rework this.

jstrzebonski:
> > Thrust header are not treated as system header as they should.
> 
> I tried
> 
> ```shell
> /kokkos/gnu_generate_makefile.bash \
>   --with-options=compiler_warnings \
>   --cxxflags="-Werror -isystem /usr/local/cuda/targets/x86_64-linux/include/thrust" \
>   --cxxstandard=c++14 \
>   --with-cuda \
>   --with-cuda-options=enable_lambda \
>   --arch=Volta70
> ```
> 
> but that didn't help.

If anyone could give me a hand here, I would really appreciate it.

algorithms/src/Kokkos_Sort.hpp
dalg24: I am saying I expect the signature would be
```C++
void sort(const Cuda& space, const Kokkos::View<DataType, Properties...>& view);

Comment (jstrzebonski) [reply algorithms/src/Kokkos_Sort.hpp:650]: @PhilMiller @dalg24 
Here https://github.com/kokkos/kokkos/pull/5372 I removed that argument completely and extracted std::sort to separated overload for serial space. If that PR will be accepted, then I can get back here to CUDA overload.

File: algorithms/src/Kokkos_Sort.hpp [hunk 8]
```diff
@@ -53,6 +53,22 @@
...
 #include <algorithm>
 
+#if defined(KOKKOS_ENABLE_CUDA)
+#include "std_algorithms/Kokkos_BeginEnd.hpp"
+
+// Workaround for `Instruction 'shfl' without '.sync' is not supported on
+// .target sm_70 and higher from PTX ISA version 6.4`.
+// Also see https://github.com/NVIDIA/cub/pull/170.
+#if ((__CUDACC_VER_MAJOR__ >= 9) || defined(__NVCOMPILER_CUDA__) || \
+     CUDA_VERSION >= 9000) &&                                       \
```
masterleinad, line 63: `CUDA_VERSION` will always be larger than `9000`. The (new) minimum is `Cuda 10.0.0`.

jstrzebonski: Removed.

fnrizzi:
@jstrzebonski there are some CI failures

jstrzebonski:
> @jstrzebonski there are some CI failures

Unfortunately yes. And I already tried to fix them, but without any luck -> https://github.com/kokkos/kokkos/pull/5183#issuecomment-1220972539

jstrzebonski:
> @jstrzebonski there are some CI failures

@fnrizzi Unfortunately yes. And I already tried to fix them, but without any luck -> https://github.com/kokkos/kokkos/pull/5183#issuecomment-1220972539

dalg24:
> > @jstrzebonski there are some CI failures
> 
> @fnrizzi Unfortunately yes. And I already tried to fix them, but without any luck -> [#5183 (comment)](https://github.com/kokkos/kokkos/pull/5183#issuecomment-1220972539)

See https://github.com/NVIDIA/thrust/pull/1744#issuecomment-1188564022
We might need to do a pragma warning push/pop

algorithms/src/Kokkos_Sort.hpp
jstrzebonski: That's resolved in https://github.com/kokkos/kokkos/pull/5388 and https://github.com/kokkos/kokkos/pull/5372

PR #5185, opened 2022-07-07
Deprecate nested types in functional

JBludau:
The functions in `Kokkos_Functional.hpp` mimic parts of  `<functional>` but are device portable. See [greater](https://en.cppreference.com/w/cpp/utility/functional/greater) as example. 
In the std lib the nested typenames `result_type`, `first_argument_type`, and `second_argument_type` are deprecated in c++17 and removed in 20. The respective functions in Kokkos should be marked deprecated when bumping to c++17

~Unfortunately, it looks like `using` declarations can not be attributed with `[[deprecated]]` but `typedef` can.~

### DIFF
--- a/containers/src/Kokkos_Functional.hpp
+++ b/containers/src/Kokkos_Functional.hpp
@@ -52,10 +52,12 @@ namespace Kokkos {
 
 template <typename T>
 struct pod_hash {
-  using argument_type        = T;
-  using first_argument_type  = T;
-  using second_argument_type = uint32_t;
-  using result_type          = uint32_t;
+#if defined KOKKOS_ENABLE_DEPRECATED_CODE_3
+  using argument_type KOKKOS_DEPRECATED        = T;
+  using first_argument_type KOKKOS_DEPRECATED  = T;
+  using second_argument_type KOKKOS_DEPRECATED = uint32_t;
+  using result_type KOKKOS_DEPRECATED          = uint32_t;
+#endif
 
   KOKKOS_FORCEINLINE_FUNCTION
   uint32_t operator()(T const& t) const {
@@ -70,9 +72,11 @@ struct pod_hash {
 
 template <typename T>
 struct pod_equal_to {
-  using first_argument_type  = T;
-  using second_argument_type = T;
-  using result_type          = bool;
+#if defined KOKKOS_ENABLE_DEPRECATED_CODE_3
+  using first_argument_type KOKKOS_DEPRECATED  = T;
+  using second_argument_type KOKKOS_DEPRECATED = T;
+  using result_type KOKKOS_DEPRECATED          = bool;
+#endif
 
   KOKKOS_FORCEINLINE_FUNCTION
   bool operator()(T const& a, T const& b) const {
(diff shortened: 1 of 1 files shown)


masterleinad:
> Unfortunately, it looks like `using` declarations can not be attributed with `[[deprecated]]` but `typedef` can.

https://godbolt.org/z/9rYv9xxa8 shows that putting `[[deprecated]]` last seems to work.

JBludau:
> > Unfortunately, it looks like `using` declarations can not be attributed with `[[deprecated]]` but `typedef` can.
> 
> https://godbolt.org/z/9rYv9xxa8 shows that putting `[[deprecated]]` last seems to work.

Interesting

crtrott (approved):

masterleinad (commented):
Do we want to remove it with Kokkos 4 or when using C++20 support? If the former one, we should adjust the comments and wrap it in `#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_3`. In the latter case with `#if defined(KOKKOS_ENABLE_CXX14) || defined(KOKKOS_ENABLE_CXX17)`.

JBludau:
> Do we want to remove it with Kokkos 4 or when using C++20 support? If the former one, we should adjust the comments and wrap it in `#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_3`. In the latter case with `#if defined(KOKKOS_ENABLE_CXX14) || defined(KOKKOS_ENABLE_CXX17)`.

I think we should stay inline with what the std does here

masterleinad (approved):
Fine with me.

JBludau:
CI failure seems unrelated 
` 1/37 Test  #1: KokkosCore_UnitTest_OpenMP ...................***Timeout 1500.69 sec`

containers/src/Kokkos_Functional.hpp
```diff
@@ -46,16 +46,29 @@
...
 
 template <typename T>
 struct pod_hash {
-  using argument_type        = T;
-  using first_argument_type  = T;
-  using second_argument_type = uint32_t;
-  using result_type          = uint32_t;
+#if defined KOKKOS_ENABLE_CXX14 || defined KOKKOS_ENABLE_CXX17
+  using argument_type KOKKOS_DEPRECATED_WITH_COMMENT_17(
+      "(deprecated in C++17)(removed in C++20)") = T;
```
dalg24, line 64: I don't get the deprecation message

JBludau: After some discussion with Damien I propose to drop the message completely. Hinting towards the cause (deprecation of [binary_function](https://en.cppreference.com/w/cpp/utility/functional/binary_function) in the std) is kind of misleading and [this](https://godbolt.org/z/j1zE6fq5W) indicates that compilers are not rushing to remove it.

dalg24 (changes requested):
I would have gone for 
```C++
   using first_argument_type KOKKOS_DEPRECATED  = T;
   using second_argument_type KOKKOS_DEPRECATED = T;
   using result_type KOKKOS_DEPRECATED          = bool;
#endif
```
I would prefer to plan for simple removal in Kokkos 4.

dalg24 (approved):

JBludau:
ci failure unrelated

masterleinad (approved):

ajpowelsnl:
FYI - [cppreference](https://en.cppreference.com/w/cpp/utility/functional/function) details deprecation of `argument_type`, `first_argument_type`, `second_argument_type`. 

What will the usage be after deprecation?

masterleinad:
> What will the usage be after deprecation?

They are not necessary anymore as explained, e.g., in https://stackoverflow.com/a/35907663, because of `decltype` and perfect forwarding.

Issue #3612, opened 2020-11-21
Running older SM kernels on newer archs results in Kokkos::abort


j8asic:
I know you're targeting specially compiled HPC platforms, but for consumer-harware cases the following restrictions are wrong, `Kokkos_Cuda_Instance.cpp, lines 368 to 377`:
```
    if (compiled_major != cudaProp.major || compiled_minor > cudaProp.minor) {
      std::stringstream ss;
      ss << "Kokkos::Cuda::initialize ERROR: running kernels compiled for "
            "compute capability "
         << compiled_major << "." << compiled_minor
         << " on device with compute capability " << cudaProp.major << "."
         << cudaProp.minor << " is not supported by CUDA!" << std::endl;
      std::string msg = ss.str();
      Kokkos::abort(msg.c_str());
    }
```

This should be warning at best, not `abort` if older kernels run on newer devices...
I assume the following test
https://github.com/kokkos/kokkos/blob/03ded4e816da5927b946065b005ebfdf38e5a438/core/src/Cuda/Kokkos_Cuda_Instance.cpp#L368
for the abort case should be written as:
`
if ((compiled_major > cudaProp.major) || (compiled_major == cudaProp.major && compiled_minor > cudaProp.minor)) {
`

crtrott:
No: according to CUDA you are not guaranteed that if you compile for example for 6.5 that it works on 6.0. We actually had runtime failures of code with running Kepler 37 on Kepler 35. From the CUDA Programming Guide:
```
In other words, a cubin object generated for compute capability X.y will only execute on devices of compute capability X.z where z≥y. 
```
https://docs.nvidia.com/cuda/cuda-c-programming-guide/#binary-compatibility

j8asic:
This is also what I'm saying. And I'm trying to say the L368 "if" condition for this is buggy.

dalg24:
@j8asic I am reading the documentation like Christian
```C++
// Binary compatibility is guaranteed from one minor revision to the next one
if (compiled_major != cudaProp.major // but not across major revisions
  || compiled_minor > cudaProp.minor) // nor from one minor revision to the previous one
```

The condition you suggest would allow binary code compiled for 3.5 to run on 6.0 which is not guaranteed to work according to the documentation.

j8asic:
[...]

PR #5187, opened 2022-07-08
Allow CUDA PTX forward compatibility if using Clang

j8asic:
This is a fix for #3612 because Clang always includes PTX


dalg24:
Can one of the admins verify this patch?

masterleinad (approved):
Looks reasonable to me.

PhilMiller:
Ok to test

j8asic:
@crtrott we have been deploying apps compiled using clang 12 and 13, from cuda 9.2 on. I cannot say we had reports on compatibility issues. (Tho we have moved the minimum from 3.5 to 6.1)

Even of they do happen, Clang’s official stance on forward compat should have higher priority over exceptional situations.

PR #5188, opened 2022-07-08
Preserve KOKKOS_INVALID_INDEX in ViewDimension and ArrayLayout construction

### LINKED ISSUE #5175: Nightly Trilinos failure in phalanx unit test (all backends) Phalanx_tViewOfViews_MPI_1: Constructor for Kokkos View mismatched number of arguments
**Describe the bug**
Following merge of #4618 builds of Trilinos (all backends) began failing in the Phalanx_tViewOfViews_MPI_1 unit test:
```
23:35:00 1087/2771 Test #1617: Phalanx_tViewOfViews_MPI_1 ...........................................................................***Failed    2.77 sec
...
23:35:00 4. PhalanxViewOfViews_ViewOfView3_DefaultCtorDtor_UnitTest ... [Passed] (9.1e-06 sec)
23:35:00 Constructor for Kokkos View 'a' has mismatched number of arguments. Number of arguments = 8 but dynamic rank = 1 
23:35:00 
...
```

Were there changes in #4618 that altered View behavior? Adding @rppawlo in case the PR exposed a hidden issue within the Phalanx unit test

**Reproducer** (weaver testbed rhel7W queue):
```bash
# Repos
git clone -b kokkos-promotion https://github.com/trilinos/Trilinos.git
git clone -b develop kokkos
git clone -b develop kokkos-kernels
# Symbolic link to your kokkos and kokkos-kernels repos in Trilinos source directory for source override
cd Trilinos
ln -s <path-to-your-repo>/kokkos kokkos
ln -s <path-to-your-repo>/kokkos-kernels kokkos-kernels

cd $HOME
mkdir -p build
cd build

# Modules and env
export ATDM_CONFIG_REGISTER_CUSTOM_CONFIG_DIR=$TRILINOS_DIR/cmake/std/atdm/contributed/weaver
source $TRILINOS_DIR/cmake/std/atdm/load-env.sh weaver-cuda-10.1-opt
export ATDM_CONFIG_USE_NINJA=OFF
module list
export OMPI_CXX=<path-to-your-kokkos-repo>/bin/nvcc_wrapper

# Configure
cmake \
 -G"Unix Makefiles" \
 -DTrilinos_CONFIGURE_OPTIONS_FILE:STRING=cmake/std/atdm/ATDMDevEnv.cmake \
 -DCMAKE_INSTALL_PREFIX=$PWD/install \
 -DCMAKE_CXX_STANDARD="14" \
 -DTrilinos_ENABLE_TESTS=OFF \
 -DTrilinos_ENABLE_ALL_PACKAGES=ON \
 -DTrilinos_ENABLE_Phalanx=ON \
 -DPhalanx_ENABLE_TESTS=ON \
 -DKokkos_SOURCE_DIR_OVERRIDE:STRING=kokkos \
 -DKokkosKernels_SOURCE_DIR_OVERRIDE:STRING=kokkos-kernels \
$TRILINOS_DIR
```

janciesko:
- Preserves KOKKOS_INVALID_INDEX in ViewDimension and ArrayLayout construction.
- Fixes #5175
- Adds unit test for View construction from Layout. This would have failed previously (as found in the issue).

### DIFF
--- a/core/src/impl/Kokkos_ViewLayoutTiled.hpp
+++ b/core/src/impl/Kokkos_ViewLayoutTiled.hpp
@@ -493,8 +493,14 @@ struct ViewOffset<
   //----------------------------------------
 
   KOKKOS_INLINE_FUNCTION constexpr array_layout layout() const {
-    return array_layout(m_dim.N0, m_dim.N1, m_dim.N2, m_dim.N2, m_dim.N3,
-                        m_dim.N4, m_dim.N5, m_dim.N6, m_dim.N7);
+    return array_layout((VORank > 0 ? m_dim.N0 : KOKKOS_INVALID_INDEX),
+                        (VORank > 1 ? m_dim.N1 : KOKKOS_INVALID_INDEX),
+                        (VORank > 2 ? m_dim.N2 : KOKKOS_INVALID_INDEX),
+                        (VORank > 3 ? m_dim.N3 : KOKKOS_INVALID_INDEX),
+                        (VORank > 4 ? m_dim.N4 : KOKKOS_INVALID_INDEX),
+                        (VORank > 5 ? m_dim.N5 : KOKKOS_INVALID_INDEX),
+                        (VORank > 6 ? m_dim.N6 : KOKKOS_INVALID_INDEX),
+                        (VORank > 7 ? m_dim.N7 : KOKKOS_INVALID_INDEX));
   }
 
   KOKKOS_INLINE_FUNCTION constexpr size_type dimension_0() const {
--- a/core/src/Kokkos_View.hpp
+++ b/core/src/Kokkos_View.hpp
@@ -80,7 +80,7 @@ class ViewMapping {
 };
 
 template <typename IntType>
-KOKKOS_INLINE_FUNCTION std::size_t count_valid_integers(
+constexpr KOKKOS_INLINE_FUNCTION std::size_t count_valid_integers(
     const IntType i0, const IntType i1, const IntType i2, const IntType i3,
     const IntType i4, const IntType i5, const IntType i6, const IntType i7) {
   static_assert(std::is_integral<IntType>::value,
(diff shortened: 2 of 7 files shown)


core/src/impl/Kokkos_ViewLayoutTiled.hpp
```diff
@@ -493,8 +493,14 @@ struct ViewOffset<
...
-    return array_layout(m_dim.N0, m_dim.N1, m_dim.N2, m_dim.N2, m_dim.N3,
-                        m_dim.N4, m_dim.N5, m_dim.N6, m_dim.N7);
+    return array_layout((VORank > 0 ? m_dim.N0 : KOKKOS_INVALID_INDEX),
+                        (VORank > 1 ? m_dim.N1 : KOKKOS_INVALID_INDEX),
+                        (VORank > 2 ? m_dim.N2 : KOKKOS_INVALID_INDEX),
+                        (VORank > 3 ? m_dim.N3 : KOKKOS_INVALID_INDEX),
+                        (VORank > 4 ? m_dim.N4 : KOKKOS_INVALID_INDEX),
+                        (VORank > 5 ? m_dim.N5 : KOKKOS_INVALID_INDEX),
+                        (VORank > 6 ? m_dim.N6 : KOKKOS_INVALID_INDEX),
+                        (VORank > 7 ? m_dim.N7 : KOKKOS_INVALID_INDEX));
```
masterleinad, line 503: Doesn't this also need to be the dynamic rank?

janciesko: From looking at TestViewLayoutTiled and the layout implementation it looks like we do only support static ranks. I think we should be good with that for now.

crtrott (approved):

crtrott:
Note, this does change the behavior to what mdspan does. I wouldn't necessarily document that yet, but I think it is the correct thing to do: i.e. you can either pass dynamik_rank or rank arguments to the View. Note the whole layout() thing is still a bit messy. For example this would still allow this I think:

```c++
View<int*> a("A",N);
View(int*[4]) b("B",a.layout());
```

But lets fix that later.

masterleinad (approved):
I think this is good enough as a fix. We should also check for matching static extents in a follow-up.

PhilMiller:
I expect this will help with the issues I was trying to address in `DynRankView`. I'll need to look at the code, and maybe try to update my PRs related to that to be sure. Probably not a reason to hold this up if it addresses the more pressing issue(s) for core View.

janciesko:
> Note, this does change the behavior to what mdspan does. I wouldn't necessarily document that yet, but I think it is the correct thing to do: i.e. you can either pass dynamik_rank or rank arguments to the View. Note the whole layout() thing is still a bit messy. For example this would still allow this I think:
> 
> ```c++
> View<int*> a("A",N);
> View(int*[4]) b("B",a.layout());
> ```
> 
> But lets fix that later.

Yes, that's fixable.

janciesko:
Retest this please

crtrott:
I wonder whether that timeout is real? I.e. we screw somewhere up and don't get all the information over and a loop nest now is infinite long or so?

nliber (approved):

nliber (approved):

dalg24:
> I wonder whether that timeout is real? I.e. we screw somewhere up and don't get all the information over and a loop nest now is infinite long or so?

I've seen this failure (OpenMP timeout) in other PRs

PR #5192, opened 2022-07-11
Dispatch Kokkos::sort(Kokkos::View) to ROCm Thrust

jstrzebonski:


CMakeLists.txt
```diff
@@ -291,6 +291,11 @@ IF (NOT KOKKOS_HAS_TRILINOS AND NOT Kokkos_INSTALL_TESTING)
 ENDIF()
 INCLUDE(${KOKKOS_SRC_PATH}/cmake/kokkos_install.cmake)
 
+IF (Kokkos_ENABLE_HIP)
+  FIND_PACKAGE(rocthrust)
+  TARGET_LINK_LIBRARIES(kokkos INTERFACE roc::rocthrust)
+ENDIF()
```
jstrzebonski, line 297: What would be a good place for this find_package? Do we want an additional option (maybe Kokkos_ENABLE_ROCTHRUST) enabling the usage of ROCm Thrust?

brian-kelley:
I was just talking to Nick Curtis from AMD about using rocThrust, so I thought I should mention that rocThrust is not necessarily included with the rocm/HIP SDK installation (unlike CUDA, which always comes with Thrust). So this will likely need some TPL detection logic in CMake, and an option like ``Kokkos_ENABLE_ROCTHRUST``.

cmake/kokkos_tpls.cmake
```diff
@@ -44,7 +44,8 @@ IF(KOKKOS_ENABLE_HIP AND NOT KOKKOS_CXX_COMPILER_ID STREQUAL HIPCC)
 ELSE()
   SET(ROCM_DEFAULT OFF)
 ENDIF()
-KOKKOS_TPL_OPTION(ROCM    ${ROCM_DEFAULT})
+KOKKOS_TPL_OPTION(ROCM      ${ROCM_DEFAULT})
+KOKKOS_TPL_OPTION(ROCTHRUST ${ROCM_DEFAULT})
```
dalg24, line 48: We need to think carefully about this.  My initial reaction is that we should have the enable rocthrust option ON by default when HIP is enabled and have the user explicitly disable it if he still wants to build without it.
For reference
https://github.com/arborx/ArborX/blob/3cb2fa329e4ca7beb019612ada8e571d80af0f74/CMakeLists.txt#L18-L30

jstrzebonski: @dalg24 Something like this?

cz4rs:
@dalg24 @fnrizzi Please note that @jstrzebonski no longer works on Kokkos. Does this need to be reassigned?

fnrizzi:
> @dalg24 @fnrizzi Please note that @jstrzebonski no longer works on Kokkos. Does this need to be reassigned?

@cz4rs yes, one of us has to pick this up.

dalg24:
Test this please

dalg24:
OK to test

cz4rs:
The only failing build is `HIP-ROCm-5.2`:
```
23: Exception: terminate called after throwing an instance of 'std::runtime_error'
23:   what():  Desul::Error: init_lock_arrays_hip: post malloc error(hipErrorOutOfMemory): hipErrorOutOfMemory
```

cmake/kokkos_tpls.cmake
```diff
@@ -45,7 +45,18 @@ IF(KOKKOS_ENABLE_HIP AND NOT KOKKOS_CXX_COMPILER_ID STREQUAL HIPCC AND NOT
 ELSE()
   SET(ROCM_DEFAULT OFF)
 ENDIF()
-KOKKOS_TPL_OPTION(ROCM    ${ROCM_DEFAULT})
+KOKKOS_TPL_OPTION(ROCM      ${ROCM_DEFAULT})
+
+INCLUDE(CMakeDependentOption)
+CMAKE_DEPENDENT_OPTION(Kokkos_ENABLE_ROCTHRUST "Enable rocThrust support" ON "Kokkos_ENABLE_HIP" OFF)
```
crtrott, line 51: Does this option conflict with the identical named TPL option below?

crtrott (changes requested):
There is a conflict in the ROCTHRUST enable options

core/src/CMakeLists.txt
```diff
@@ -162,6 +162,7 @@ IF (NOT WIN32)
 ENDIF()
 IF (NOT KOKKOS_ENABLE_COMPILE_AS_CMAKE_LANGUAGE)
   KOKKOS_LINK_TPL(kokkoscore PUBLIC ROCM)
+  KOKKOS_LINK_TPL(kokkoscore PUBLIC ROCTHRUST)
```
crtrott, line 165: doesn't this need to be conditional on whether ROCTHRUST is ON?

cmake/kokkos_tpls.cmake
```diff
@@ -85,6 +96,7 @@ IF (NOT WIN32)
 ENDIF()
 IF (NOT KOKKOS_ENABLE_COMPILE_AS_CMAKE_LANGUAGE)
   KOKKOS_IMPORT_TPL(ROCM INTERFACE)
+  KOKKOS_IMPORT_TPL(ROCTHRUST INTERFACE)
```
crtrott, line 99: I think this needs to be conditional too?

cz4rs:
Note to self - [how to build on a machine without GPU](https://github.com/ROCmSoftwarePlatform/tensorflow-upstream/issues/508#issuecomment-503199297):
> you can trick `rocm_agent_enumerator` inside the docker with a text file under `/opt/rocm/bin/target.lst`. In the file put each line of your desired AMD GCN ISA version inside. For example:
> ```
> gfx803
> gfx900
> gfx906
> ```
> would trick `rocm_agent_enumerator` thinking you have 3 different AMD GPUs installed.

PR #5144, opened 2022-06-24
wip

dalg24:
* Remove `ExecSpaceInitializerBase` and derived classes in each backend.
* Move `InitArguments` to its own header to resolve some nonsense circular dependency issues.
* `ExecutionSpace::impl_initialize(InitArguments const&)`

This is not ready for review.  I am opening mainly to get help on individual backends.

PR #5206, opened 2022-07-12
Test Legion use case

masterleinad:
Addressing https://github.com/kokkos/kokkos/pull/5144#discussion_r919069828. I don't feel strongly about deleting the other test otr not.


core/unit_test/CMakeLists.txt
```diff
@@ -717,6 +716,11 @@ KOKKOS_ADD_EXECUTABLE_AND_TEST(
   SOURCES ${DEFAULT_DEVICE_SOURCES}
 )
 
+KOKKOS_ADD_EXECUTABLE_AND_TEST(
+  UnitTest_InitializeBackends
```
dalg24, line 720: Name is not great

core/unit_test/UnitTest_InitializeBackends.cpp
```diff
@@ -0,0 +1,108 @@
...
+  Kokkos::Threads::impl_initialize(kokkos_init_settings);
+#endif
+#ifdef KOKKOS_ENABLE_OPENMP
+  Kokkos::OpenMP::impl_initialize(kokkos_init_settings);
+#endif
+#ifdef KOKKOS_ENABLE_SERIAL
+  Kokkos::Serial::impl_initialize(kokkos_init_settings);
+#endif
+  Kokkos::Impl::post_initialize(kokkos_init_settings);
+
```
dalg24, line 80:
```suggestion

EXPECT_TRUE(Kokkos::is_initialized());
#ifdef KOKKOS_ENABLE_SYCL
EXPECT_TRUE(Kokkos::Experimental::SYCL::is_initialized());
#endif
// ...

```

core/unit_test/UnitTest_InitializeBackends.cpp
```diff
@@ -0,0 +1,108 @@
...
+  Kokkos::Threads::impl_finalize();
+#endif
+#ifdef KOKKOS_ENABLE_OPENMP
+  Kokkos::OpenMP::impl_finalize();
+#endif
+#ifdef KOKKOS_ENABLE_SERIAL
+  Kokkos::Serial::impl_finalize();
+#endif
+
+  Kokkos::finalize();
```
dalg24, line 106:
```suggestion
  Kokkos::finalize();
  
  EXPECT_FALSE(Kokkos::is_initialized());
  // ...
```

dalg24: Actually now that I look at it again I really dislike the semantics of `is_initialized`

masterleinad:
The actual `Legion` code can be found at https://gitlab.com/StanfordLegion/legion/-/blob/master/runtime/realm/kokkos_interop.cc#L168-288.

masterleinad:
I ended up reordering execution spaces so that host spaces are initialized first since I didn't find a good way to test https://github.com/kokkos/kokkos/blob/d19aab9981a2c447e832a7b4eb7b16992328fb14/core/src/Cuda/Kokkos_Cuda_Instance.cpp#L360-L367 (since it's in a *.cc file and I didn't want to add that define to the library target).

masterleinad:
Only the SYCL build is timing out in `Kokkos_Serial1` which looks spurious.

crtrott (approved):

core/unit_test/TestLegionInitialization.cpp
```diff
@@ -0,0 +1,138 @@
...
+  Kokkos::Experimental::SYCL::impl_initialize(kokkos_init_settings);
+  EXPECT_TRUE(Kokkos::Experimental::SYCL::impl_is_initialized());
+#endif
+#ifdef KOKKOS_ENABLE_OPENMPTARGET
+  EXPECT_FALSE(Kokkos::Experimental::OpenMPTarget::impl_is_initialized());
+  Kokkos::Experimental::OpenMPTarget::impl_initialize(kokkos_init_settings);
+  EXPECT_TRUE(Kokkos::Experimental::OpenMPTarget::impl_is_initialized());
+#endif
+#ifdef KOKKOS_ENABLE_HIP
+  EXPECT_FALSE(Kokkos::Experimental::HIP::impl_is_initialized());
```
Rombur, line 85:
```suggestion
  EXPECT_FALSE(Kokkos::HIP::impl_is_initialized());
```

core/unit_test/TestLegionInitialization.cpp
```diff
@@ -0,0 +1,138 @@
...
+  Kokkos::Experimental::HIP::impl_initialize(kokkos_init_settings);
```
Rombur, line 86:
```suggestion
  Kokkos::HIP::impl_initialize(kokkos_init_settings);
```

core/unit_test/TestLegionInitialization.cpp
```diff
@@ -0,0 +1,138 @@
...
+  EXPECT_TRUE(Kokkos::Experimental::HIP::impl_is_initialized());
```
Rombur, line 87:
```suggestion
  EXPECT_TRUE(Kokkos::HIP::impl_is_initialized());
```

core/unit_test/TestLegionInitialization.cpp
```diff
@@ -0,0 +1,138 @@
...
+#ifdef KOKKOS_ENABLE_SYCL
+  Kokkos::Experimental::SYCL::impl_finalize();
+  EXPECT_FALSE(Kokkos::Experimental::SYCL::impl_is_initialized());
+#endif
+#ifdef KOKKOS_ENABLE_OPENMPTARGET
+  Kokkos::Experimental::OpenMPTarget::impl_finalize();
+  EXPECT_FALSE(Kokkos::Experimental::OpenMPTarget::impl_is_initialized());
+#endif
+#ifdef KOKKOS_ENABLE_HIP
+  Kokkos::Experimental::HIP::impl_finalize();
```
Rombur, line 123:
```suggestion
  Kokkos::HIP::impl_finalize();
```

core/unit_test/TestLegionInitialization.cpp
```diff
@@ -0,0 +1,138 @@
...
+  EXPECT_FALSE(Kokkos::Experimental::HIP::impl_is_initialized());
```
Rombur, line 124:
```suggestion
  EXPECT_FALSE(Kokkos::HIP::impl_is_initialized());
```

Rombur (changes requested):

Rombur (approved):

core/unit_test/TestLegionInitialization.cpp
```diff
@@ -0,0 +1,138 @@
...
+// ************************************************************************
+//@HEADER
+*/
+
+#include <Kokkos_Core.hpp>
+#include <gtest/gtest.h>
+
+namespace {
+// The purpose of this test is to mimic Legion's use case of initializing and
+// finalizing individual backends
```
dalg24, line 50: Out of context may sound reasonable but ends up imposing a constraints on all execution spaces (existence and semantics of member function `ExecutionSpace::impl_initialize` and then probably should not be related to Legion.

dalg24 (changes requested):

core/unit_test/TestLegionInitialization.cpp
masterleinad: I thought that is what we promised `Legion` to maintain, didn't we?

Rombur: > I thought that is what we promised Legion to maintain, didn't we?

That was also my understanding.

dalg24: I meant `ExecutionSpace::impl_{initialize -> is_initialized}` above.  Sorry about the confusion

masterleinad:
What I wanted to achieve with this pull request (mostly) was that the way we promised `Legion` to initialize `Kokkos` works, i.e.,
the sequence of
```
Kokkos::Impl::pre_initialize()
EXECUTION_SPACE::impl_initialize();
Kokkos::Impl::post_initialize();
EXECUTION_SPACE::impl_finalize();
```
is valid. On the way, I discovered that every backend implements `EXECUTION_SPACE::impl_is_initialized()` and so I used it.
If we think that this is not a useful test, I'm happy to bury it.

crtrott:
Maybe we should just use the aliases i.e. call:

```c++
if(!is_same_v<Serial, DefaultHostExecutionSpace>)
  Serial::impl_initialize();
if(!is_same_v<DefaultExecutionSpace, DefaultHostExecutionSpace>)
  DefaultHostExecutionSpace::impl_initialize();
DefaultExecutionSpace::impl_initialize(...);
```

Same for finalize. Then add some actual test in between.
```c++
View<int*> data("d",1000);
deep_copy(d,1);
int result;
Kokkos::parallel_reduce("TestRed", d.extent(0), KOKKOS_LAMBDA(int i, int& sum) {
  sum += d(i);
},result);
EXPECT_EQ(result, d.extent(0));
```

Don't use the `impl_is_initialized`, just check that `Kokkos::is_initialized` is correct.

core/unit_test/TestLegionInitialization.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+  Kokkos::DefaultExecutionSpace::impl_finalize();
+  Kokkos::finalize();
+}
+}  // namespace
+
+int main(int argc, char* argv[]) {
+  ::testing::InitGoogleTest(&argc, argv);
+  return RUN_ALL_TESTS();
+}
```
dalg24, line 89: We already have a version of main that does that.
`core/unit_test/UnitTestMain.cpp`
Any particular reason why we should not use it there too?

core/unit_test/TestLegionInitialization.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+// The purpose of this test is to mimic Legion's use case of initializing and
+// finalizing individual backends
+TEST(initialization, legion_initialization) {
+  Kokkos::InitializationSettings kokkos_init_settings;
+
+  Kokkos::Impl::pre_initialize(kokkos_init_settings);
+
+  // We need to have a host execution space initialized first.
+  Kokkos::HostSpace::execution_space::impl_initialize(kokkos_init_settings);
```
dalg24, line 64:
```suggestion
  Kokkos::DefaultHostExecutionSpace::impl_initialize(kokkos_init_settings);
```

core/unit_test/TestLegionInitialization.cpp
```diff
@@ -0,0 +1,89 @@
...
+
+  if (!std::is_same_v<Kokkos::DefaultExecutionSpace,
+                      Kokkos::HostSpace::execution_space>)
```
dalg24, line 67:
```suggestion
                      Kokkos::DefaultHostExecutionSpace>)
```

core/unit_test/TestLegionInitialization.cpp
```diff
@@ -0,0 +1,89 @@
...
+    Kokkos::DefaultExecutionSpace::impl_initialize(kokkos_init_settings);
+
+  Kokkos::Impl::post_initialize(kokkos_init_settings);
+
```
dalg24, line 71:
```suggestion

  EXPECT_TRUE(Kokkos::is_initialized());
  
```

core/unit_test/TestLegionInitialization.cpp
```diff
@@ -0,0 +1,89 @@
...
+    Kokkos::View<int*> d("d", 1000);
+    Kokkos::deep_copy(d, 1);
+    int result;
+    Kokkos::parallel_reduce("TestRed", d.extent(0), ReductionFunctor{d},
+                            result);
+    EXPECT_EQ(result, d.extent(0));
+  }
+
+  Kokkos::DefaultExecutionSpace::impl_finalize();
+  Kokkos::finalize();
```
dalg24, line 82: Wait what?  Why would we call `Kokkos::finalize().  If you initialize piecemeal by hand you must finalize by hand?
```suggestion
```

core/src/HIP/Kokkos_HIP_Instance.cpp
```diff
@@ -301,6 +301,8 @@ void HIPInternal::release_team_scratch_space(int scratch_pool_id) {
 //----------------------------------------------------------------------------
 
 void HIPInternal::finalize() {
+  if (was_finalized) return;
```
dalg24, line 304: Revert these

dalg24 (changes requested):

core/unit_test/TestLegionInitialization.cpp
masterleinad: I would need to call `Kokkos::finalize` for `is_finalized()` to return `true` and also to run any `finalize_hooks`.

dalg24: Absolutely not ok with changing all backends so you can effectively call `ExecutionSpace::impl_finalize` multiple time nor call it when it was never initialized

core/src/impl/Kokkos_Core.cpp
```diff
@@ -1238,7 +1236,8 @@ void Kokkos::finalize() {
   if (kokkos_finalize_was_called()) {
     Kokkos::abort("Error: Kokkos::finalize() has already been called.\n");
   }
-  finalize_internal();
+  Kokkos::Impl::finalize_without_execution_spaces();
+  Kokkos::Impl::ExecSpaceManager::get_instance().finalize_spaces();
```
dalg24, line 1240: This effectively changes the execution order in the finalization (setting the global variables before finalizing all spaces.  Why is ok to change it?
Also there is a prior terminology `Impl::{pre,post}_initialize`.  Did you consider `{pre,post}_finalize` before you proposed the `finalize_without_execution_spaces` function?

dalg24: One could imagine backend finalization would poll `Kokkos::show_warnings` for instance.
`finalize_without_execution_spaces` is not a good name.

masterleinad: Would you be happy if I introduce `{pre,post}_finalize`?

dalg24: Sure

dalg24 (commented):
Would approve with the changes suggested in masterleinad/kokkos#9

dalg24 (approved):

dalg24:
I just realize the compile test is not being removed any more.  Is it intentional?  I am fine with the removal.

Issue #5205, opened 2022-07-12
Missing error check for some View constructors


lucbv:
@kliegeois 
**Describe the bug**
Kokkos computes wrong extent of subview below is the source code followed by the output:
```c++
const int n = A.extent(0);
ScratchPadMatrixViewType tmp(member.team_scratch(0), n, n + 4);
std::cout << "tmp extents: (" << tmp.extent(0) << ", " << tmp.extent(1) << "), n=" << n << std::endl;
```

```
tmp extents: (3, 1), n=3
unknown file: Failure
C++ exception with description "Kokkos::subview bounds error ( Kokkos::ALL , 1 <= 3 - 0 )" thrown in the test body.
```

**Please also include the following items to support reproducing the bug**
1. compilers (with versions): GNU 10.1.0
2. Kokkos release or commit used (i.e., the sha1 number): a51981a9
3. platform and backend: serial
4. cmake configure command: `cmake -DKokkos_ENABLE_SERIAL=ON -DCMAKE_CXX_FLAGS="-Werror" -DCMAKE_CXX_STANDARD=14 -DKokkos_ENABLE_COMPILER_WARNINGS=ON  -DKokkos_ENABLE_DEBUG_BOUNDS_CHECK:BOOL=ON -DKokkos_ENABLE_DEPRECATED_CODE_3=OFF -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=$PWD/../install ..`
5. output from cmake command 
```
-- Setting default Kokkos CXX standard to 14
-- The CXX compiler identification is AppleClang 13.0.0.13000029
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /Applications/Xcode_13.2.1.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/c++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- The project name is: Kokkos
-- Using -std=gnu++14 for C++14 extensions as feature
-- Built-in Execution Spaces:
--     Device Parallel: NoTypeDefined
--     Host Parallel: NoTypeDefined
--       Host Serial: SERIAL
-- 
-- Architectures:
-- Found TPLLIBDL: /Applications/Xcode_13.2.1.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.1.sdk/usr/include  
[...]

PR #5209, opened 2022-07-12
Assert that layout construction from extents is valid where used

crtrott:
Also introduces a missing layout overload for required_allocation_size

### DIFF
--- a/core/src/Kokkos_View.hpp
+++ b/core/src/Kokkos_View.hpp
@@ -1536,7 +1536,11 @@ class View : public ViewTraits<DataType, Properties...> {
       : View(arg_prop,
              typename traits::array_layout(arg_N0, arg_N1, arg_N2, arg_N3,
                                            arg_N4, arg_N5, arg_N6, arg_N7),
-             check_input_args::yes) {}
+             check_input_args::yes) {
+    static_assert(traits::array_layout::is_extent_constructible,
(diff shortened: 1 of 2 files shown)


crtrott:
Fixes #5205

core/src/Kokkos_View.hpp
```diff
@@ -1536,7 +1536,11 @@ class View : public ViewTraits<DataType, Properties...> {
       : View(arg_prop,
              typename traits::array_layout(arg_N0, arg_N1, arg_N2, arg_N3,
                                            arg_N4, arg_N5, arg_N6, arg_N7),
-             check_input_args::yes) {}
+             check_input_args::yes) {
+    static_assert(traits::array_layout::is_extent_constructible,
+                  "Layout is not extent constructible. Use overload taking a "
```
dalg24, line 1541: Please suggest a message that will be easier to understand for users

crtrott: I don't know, seems pretty obvious to me?

dalg24: "is not constructible from extent arguments.  Use ..."

masterleinad: I would just use the second sentence "Use the overload taking a layout object instead!".

nliber (approved):

PR #5211, opened 2022-07-13
Refactor device selection at initialization

dalg24:
Deprecate `num_devices` and `skip_device` settings, `--kokkos-num-devices` command line argument, `KOKKOS_NUM_DEVICES`, `KOKKOS_SKIP_DEVICES`, and `KOKKOS_RAND_DEVICES` environment variable.

Introduce as a replacement
* `map_device_id_by` setting along with `--kokkos-map-device-id-by` and `KOKKOS_MAP_DEVICE_ID_BY` which can be set to `"random"` or `"mpi_rank"`.
* `KOKKOS_VISIBLE_DEVICES` environment variable to specify device ids as a comma-separated sequence of integers

### DIFF
--- a/core/src/impl/Kokkos_Core.cpp
+++ b/core/src/impl/Kokkos_Core.cpp
@@ -358,11 +358,11 @@ int Kokkos::Impl::get_gpu(const InitializationSettings& settings) {
   // map_device_id provided
   // either random or round-robin assignement based on local MPI rank
   if (!is_valid_map_device_id_by(settings.get_map_device_id_by())) {
-    std::cerr << "Warning: unrecognized map_device_id_by setting \""
-              << settings.get_map_device_id_by() << "\" ignored."
-              << " Raised by Kokkos::initialize(int argc, char* argv[])."
-              << std::endl;
-    return visible_devices[0];
+    std::stringstream ss;
+    ss << "Warning: map_device_id_by setting '"
+       << settings.get_map_device_id_by() << "' is not recognized."
+       << " Raised by Kokkos::initialize(int argc, char* argv[]).\n";
+    Kokkos::abort(ss.str().c_str());
   }
 
   if (settings.get_map_device_id_by() == "random") {
@@ -873,11 +873,11 @@ void Kokkos::Impl::parse_command_line_arguments(
       if (is_valid_map_device_id_by(map_device_id_by)) {
         settings.set_map_device_id_by(map_device_id_by);
       } else {
-        std::cerr << "Warning: unrecognized value for command line argument "
-                     "--kokkos-map-device-id-by=\""
-                  << map_device_id_by << "\" ignored."
-                  << " Raised by Kokkos::initialize(int argc, char* argv[])."
-                  << std::endl;
+        std::stringstream ss;
+        ss << "Warning: command line argument '--kokkos-map-device-id-by="
+           << map_device_id_by << "' is not recognized."
+           << " Raised by Kokkos::initialize(int argc, char* argv[]).\n";
+        Kokkos::abort(ss.str().c_str());
       }
     }
 
@@ -1036,11 +1036,11 @@ void Kokkos::Impl::parse_environment_variables(
     if (is_valid_map_device_id_by(env_map_device_id_by_str)) {
       settings.set_map_device_id_by(env_map_device_id_by_str);
     } else {
-      std::cerr << "Warning: unrecognized value for environment variable "
-                << "KOKKOS_MAP_DEVICE_ID_BY=" << env_map_device_id_by_str
-                << "ignored."
-                << " Raised by Kokkos::initialize(int argc, char* argv[])."
-                << std::endl;
+      std::stringstream ss;
+      ss << "Warning: environment variable 'KOKKOS_MAP_DEVICE_ID_BY="
+         << env_map_device_id_by_str << "' is not recognized."
+         << " Raised by Kokkos::initialize(int argc, char* argv[]).\n";
+      Kokkos::abort(ss.str().c_str());
     }
   }
 }
--- a/core/src/impl/Kokkos_Core.cpp
+++ b/core/src/impl/Kokkos_Core.cpp
@@ -61,6 +61,7 @@
 #include <functional>
 #include <list>
 #include <cerrno>
+#include <random>
 #include <regex>
 #ifndef _WIN32
 #include <unistd.h>
@@ -150,6 +151,32 @@ void combine(Kokkos::Tools::InitArguments& out,
   if (in.has_tools_args()) {
     out.args = in.get_tools_args();
   }
+}
+
+int get_device_count() {
+#if defined(KOKKOS_ENABLE_CUDA)
+  return Kokkos::Cuda::detect_device_count();
+#elif defined(KOKKOS_ENABLE_HIP)
+  return Kokkos::Experimental::HIP::detect_device_count();
+#elif defined(KOKKOS_ENABLE_SYCL)
+  return sycl::device::get_devices(sycl::info::device_type::gpu).size();
+#else
+  // This function is always compiled but should only be ever called when
+  // either CUDA, HIP, or SYCL are enabled.
+  return *reinterpret_cast<int*>(0x8BADF00D);  // implementation bug
+#endif
+}
+
+unsigned get_process_id() {
+#ifdef _WIN32
+  return unsigned(GetCurrentProcessId());
+#else
+  return unsigned(getpid());
+#endif
+}
+
+bool is_valid_map_device_id_by(std::string const& x) {
+  return x == "mpi_rank" || x == "random";
 }
 
 }  // namespace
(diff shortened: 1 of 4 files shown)


crtrott:
Summary: 
```
KOKKOS_VISIBLE_DEVICES -> externally enforced mask

// These also exist as members in the initialization struct
--kokkos-device-id=2
--kokkos-map-device-id-by=random
--kokkos-map-device-id-by=mpi-rank
(--kokkos-map-device-id-by=[specific-device(2)|2])
```

core/src/impl/Kokkos_Core.cpp
```diff
@@ -151,6 +155,32 @@ void combine(Kokkos::Tools::InitArguments& out,
...
+#if defined(KOKKOS_ENABLE_CUDA)
+  return Kokkos::Cuda::detect_device_count();
+#elif defined(KOKKOS_ENABLE_HIP)
+  return Kokkos::Experimental::HIP::detect_device_count();
+#elif defined(KOKKOS_ENABLE_SYCL)
+  return sycl::device::get_devices(sycl::info::device_type::gpu).size();
+#else
+  // This function is always compiled but should only be ever called when
+  // either CUDA, HIP, or SYCL are enabled.
+  return *reinterpret_cast<int*>(0x8BADF00D);  // implementation bug
```
fnrizzi, line 168: why this? why it says implementation bug?

fnrizzi:
@dalg24 can you please post here some examples for `KOKKOS_VISIBLE_DEVICES` ? You said comma separated list of integers right?

core/src/impl/Kokkos_Core.cpp
dalg24: `get_device_count() -> int` is only implemented for CUDA/HIP/SYCL but the function actually gets compiled regardless.  It is called from `Impl::get_gpu(InitializationSettings const&) -> int` which gets only used in these 3 backends.  I wrote `return *reinterpret_cast<int*>(0x8BADF00D);` to cause a segfault if that changes (say OpenMPTarget or OpenACC attempt to call `get_gpu`) so that we don't silently compute garbage and it is obvious where the issue is coming from.

dalg24: If I return say `-1` one would see errors/warnings saying there is no device available and would have to trace back to that function.  I opted for failing early in an attempt to make it easier to debug.  Does that make sense?

fnrizzi: ok, i see. Why not just aborting instead of this reinterpret cast? 
Also, I think it would be great to have your explanation above added as comment there so that one clearly understands things

dalg24:
> @dalg24 You said `KOKKOS_VISIBLE_DEVICES` stores a comma separated list of integers, so this is like `NVIDIA_VISIBLE_DEVICES` env variable right?

I was going for `{CUDA,HIP}_VISIBLE_DEVICES`.  I had actually forgotten about `NVIDIA_VISIBLE_DEVICES`.  That one actually also supports `all`, `none`, `void` or empty or unset but these make little sense for us.

> Since it is a list of integers, what about renaming it `KOKKOS_VISIBLE_DEVICE_IDS`? I realize this departs from the cuda naming, but IMO it is more expressive of what this env var actually stores.

I like `KOKKOS_VISIBLE_DEVICES` better because it is shorter, aligned with current vendor practice, and somehow making "device ids visible" rubs me wrong.  But we will sees if others agree with you.

> When we say "device id", it is implied id with respect to enumeration given the by, e.g., nvidia-smi right?

Yes, in whatever order the vendor runtimes detects them.

core/src/impl/Kokkos_Core.cpp
fnrizzi: >Speculating I would get warnings that function does not return but I can change if you request me to.
ok I am fine keeping the segfault 

currently we have:
```cpp
// This function is always compiled but should only be ever called when
// either CUDA, HIP, or SYCL are enabled.
```
I would suggest changing to:
```cpp
/* get_device_count() -> int is only implemented for CUDA/HIP/SYCL but the function actually gets 
    compiled regardless. It is called from Impl::get_gpu(InitializationSettings const&) -> int 
   which gets only used in these 3 backends. I wrote return *reinterpret_cast<int*>(0x8BADF00D); 
   to cause a segfault if that changes (say OpenMPTarget or OpenACC attempt to call get_gpu) 
   so that we don't silently compute garbage and it is obvious where the issue is coming from. */
```

core/src/impl/Kokkos_Core.cpp
```diff
@@ -151,6 +155,36 @@ void combine(Kokkos::Tools::InitArguments& out,
...
+#elif defined(KOKKOS_ENABLE_SYCL)
+  return sycl::device::get_devices(sycl::info::device_type::gpu).size();
+#else
+  // get_device_count() -> int is only implemented for CUDA/HIP/SYCL but the
+  // function actually gets compiled regardless. It is called from
+  // Impl::get_gpu(InitializationSettings const&) -> int which gets only used
+  // in these 3 backends.  The line below will yield a segfault if that changes
+  // (say OpenMPTarget or OpenACC attempt to call get_gpu) so that we don't
+  // silently compute garbage, and so it is obvious where the actual issue is.
+  return *reinterpret_cast<int*>(0x8BADF00D);
```
masterleinad, line 172: In my opinion,
```suggestion
  Kokkos::abort("get_device_count can only be called when compiling with Cuda, HIP, or SYCL support!");
  return -1;
```
is much clearer. Any reason you can't do that?

core/src/impl/Kokkos_Core.cpp
```diff
@@ -752,6 +802,8 @@ void Kokkos::Impl::parse_command_line_arguments(
         warn_deprecated_command_line_argument("--kokkos-ndevices",
                                               "--kokkos-num-devices");
       }
+      warn_deprecated_command_line_argument(
+          "--kokkos-num-devices", "--kokkos-map-device-id-by=mpi_rank");
```
JBludau, line 806: in the previous lines we recommend the user to use `--kokkos-num-devices`

core/src/impl/Kokkos_Core.cpp
```diff
@@ -627,14 +683,7 @@ Kokkos Core Options:
...
   --kokkos-device-id=INT         : specify device id to be used by Kokkos.
-  --kokkos-num-devices=INT[,INT] : used when running MPI jobs. Specify number of
-                                   devices per node to be used. Process to device
-                                   mapping happens by obtaining the local MPI rank
-                                   and assigning devices round-robin. The optional
-                                   second argument allows for an existing device
-                                   to be ignored. This is most useful on workstations
-                                   with multiple GPUs of which one is used to drive
-                                   screen output.
+  --kokkos-map-devide-id-by=(random|mpi_rank)
```
JBludau, line 687: typo, --kokkos-map-device-id-by=(random|mpi_rank)

core/src/impl/Kokkos_Core.cpp
```diff
@@ -281,57 +310,92 @@ int Kokkos::Impl::get_ctest_gpu(const char* local_rank_str) {
...
+      std::getenv("OMPI_COMM_WORLD_LOCAL_RANK");  // OpenMPI
+  if (!local_rank_str)
+    local_rank_str = std::getenv("MV2_COMM_WORLD_LOCAL_RANK");  // MVAPICH2
+  if (!local_rank_str) local_rank_str = std::getenv("SLURM_LOCALID");  // SLURM
+
+  if (!local_rank_str) {
+    std::cerr << "Warning: unable to detect local MPI rank."
+              << " Raised by Kokkos::initialize(int argc, char* argv[])."
+              << std::endl;
+    return visible_devices[0];
```
JBludau, line 389: I would like to give a hint to the user, that we default to device 0 in this case

core/src/impl/Kokkos_Core.cpp
```diff
@@ -281,57 +310,92 @@ int Kokkos::Impl::get_ctest_gpu(const char* local_rank_str) {
...
+  }
+
+  if (settings.get_map_device_id_by() == "random") {
+    std::default_random_engine gen(get_process_id());
+    std::uniform_int_distribution<int> distribution(0,
+                                                    visible_devices.size() - 1);
+    return visible_devices[distribution(gen)];
+  }
```
JBludau, line 373: could we end up with multiple processes using the same gpu in this case? e.g. multiple kokkos processes getting the same device id randomly ... 
Is there a real use case for the random assignment?

core/src/impl/Kokkos_Core.cpp
```diff
@@ -281,57 +310,92 @@ int Kokkos::Impl::get_ctest_gpu(const char* local_rank_str) {
...
+  if (!settings.has_map_device_id_by()) {
+    return visible_devices[0];
+  }
+  // map_device_id provided
+  // either random or round-robin assignement based on local MPI rank
+  if (!is_valid_map_device_id_by(settings.get_map_device_id_by())) {
+    std::cerr << "Warning: unrecognized map_device_id_by setting \""
+              << settings.get_map_device_id_by() << "\" ignored."
+              << " Raised by Kokkos::initialize(int argc, char* argv[])."
+              << std::endl;
```
JBludau, line 364: Maybe we should hint the user that we fallback to gpu 0 in this case

core/src/impl/Kokkos_Core.cpp
```diff
@@ -281,57 +310,92 @@ int Kokkos::Impl::get_ctest_gpu(const char* local_rank_str) {
...
+                      settings.get_skip_device()),
+          visible_devices.end());
     }
-    // shift assignments over by one so no one is assigned to "skip_device"
-    if (use_gpu >= skip_device) ++use_gpu;
   }
-  return use_gpu;
+  if (visible_devices.empty()) {
+    Kokkos::abort("Error: no GPU available for execution.\n");
+  }
```
JBludau, line 342: maybe 'Error: no GPU set visible for execution.\n'

core/src/impl/Kokkos_Core.cpp
```diff
@@ -151,6 +155,31 @@ void combine(Kokkos::Tools::InitArguments& out,
...
+int get_device_count() {
+#if defined(KOKKOS_ENABLE_CUDA)
+  return Kokkos::Cuda::detect_device_count();
+#elif defined(KOKKOS_ENABLE_HIP)
+  return Kokkos::Experimental::HIP::detect_device_count();
+#elif defined(KOKKOS_ENABLE_SYCL)
+  return sycl::device::get_devices(sycl::info::device_type::gpu).size();
+#else
+  Kokkos::abort("implementation bug");
+  return -1;
```
JBludau, line 167: this could be a better message. Maybe 'Kokkos::get_device_count() not implemented for the enabled backend'

core/src/impl/Kokkos_Core.cpp
```diff
@@ -974,6 +1028,21 @@ void Kokkos::Impl::parse_environment_variables(
       settings.set_tune_internals(false);
     }
   }
+  char* env_map_device_id_by_str = std::getenv("KOKKOS_MAP_DEVICE_ID_BY");
+  if (env_map_device_id_by_str != nullptr) {
+    if (env_device_id_str != nullptr) {
+      std::cerr << "Warning: specified both blah\n";
+    }
```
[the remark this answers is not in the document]
dalg24: `device_id` always wins

core/src/impl/Kokkos_Core.cpp
crtrott: yes we can, but in the primary use case that is not unexpected anyway. The main use case for this is parallel CI. Essentially running ctest with -j. Most unit tests don't tax GPUs dramatically and are not timing sensitive. But it helps if they don't all hit the same GPU. In practice we can run for example most Trilinos testing with -j16 on a 4 GPU machine. But if all 16 hit GPU zero we run out of memory occasionally.

crtrott: The problem there is that ctest doesn't have resource management, and its not MPI i.e. the round robin doesn't work. Random assignement is good enough in spreading out however.

core/src/impl/Kokkos_Core.cpp
```diff
@@ -815,6 +868,17 @@ void Kokkos::Impl::parse_command_line_arguments(
...
+    } else if (check_str_arg(argv[iarg], "--kokkos-map-device-id-by",
+                             map_device_id_by)) {
+      if (is_valid_map_device_id_by(map_device_id_by)) {
+        settings.set_map_device_id_by(map_device_id_by);
+      } else {
+        std::cerr << "Warning: unrecognized value for command line argument "
+                     "--kokkos-map-device-id-by=\""
+                  << map_device_id_by << "\" ignored."
+                  << " Raised by Kokkos::initialize(int argc, char* argv[])."
+                  << std::endl;
```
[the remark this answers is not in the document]
dalg24: Actually this is a good point.  Should we abort instead?

core/src/impl/Kokkos_Core.cpp
JBludau: thanks

core/src/impl/Kokkos_Core.cpp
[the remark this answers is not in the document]
JBludau: then maybe: "Warning: specified both KOKKOS_MAP_DEVICE_ID_BY and KOKKOS_DEVICE_ID. Kokkos will use KOKKOS_DEVICE_ID"

core/src/impl/Kokkos_Core.cpp
JBludau: https://xkcd.com/1024/

dalg24: I will give another pass at cleaning parsing of cmd line args as a follow up if that's ok with you.

crtrott (approved):

fnrizzi (approved):

core/src/impl/Kokkos_Core.cpp
JBludau: sure

JBludau (approved):
looks good

core/src/impl/Kokkos_Core.cpp
```diff
@@ -780,6 +840,7 @@ void Kokkos::Impl::parse_command_line_arguments(
           !kokkos_num_devices_found) {
         num_devices = std::stoi(num1_only);
         settings.set_num_devices(num_devices);
+        settings.set_map_device_id_by("mpi_rank");
```
dalg24, line 843: @xZZx @crtrott 
I suspect the issue in #5252 comes from this line.  Probably needs to move it to L815

core/src/impl/Kokkos_Core.cpp
```diff
@@ -281,57 +310,93 @@ int Kokkos::Impl::get_ctest_gpu(const char* local_rank_str) {
...
-#else
-    return num_devices;
-#endif
-  }(settings.has_num_devices() ? settings.get_num_devices() : -1);
-  const int skip_device =
-      settings.has_skip_device() ? settings.get_skip_device() : 9999;
-
-  // if the exact device is not set, but ndevices was given, assign round-robin
-  // using on-node MPI rank
-  if (use_gpu < 0) {
```
dalg24, line 304: @xzzx @crtrott I think this is what changed.  Will fix.

PR #5212, opened 2022-07-13
Use sycl::device_ptr where appropriate

masterleinad:
See https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/supported/sycl_ext_intel_usm_address_spaces.asciidoc.

### DIFF
--- a/core/src/SYCL/Kokkos_SYCL_Parallel_Team.hpp
+++ b/core/src/SYCL/Kokkos_SYCL_Parallel_Team.hpp
@@ -410,7 +410,7 @@ class ParallelFor<FunctorType, Kokkos::TeamPolicy<Properties...>,
   size_type const m_vector_size;
   int m_shmem_begin;
   int m_shmem_size;
-  sycl::global_ptr<char> m_global_scratch_ptr;
+  sycl::device_ptr<char> m_global_scratch_ptr;
   size_t m_scratch_size[2];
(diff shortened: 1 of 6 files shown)


nliber (approved):

core/src/SYCL/Kokkos_SYCL_Parallel_Team.hpp
```diff
@@ -615,9 +615,9 @@ class ParallelReduce<FunctorType, Kokkos::TeamPolicy<Properties...>,
     // m_result_ptr yet.
     if (size <= 1) {
       results_ptr =
-          static_cast<sycl::global_ptr<value_type>>(instance.scratch_space(
+          static_cast<sycl::device_ptr<value_type>>(instance.scratch_space(
               sizeof(value_type) * std::max(value_count, 1u)));
-      sycl::global_ptr<value_type> device_accessible_result_ptr =
+      sycl::device_ptr<value_type> device_accessible_result_ptr =
           m_result_ptr_device_accessible ? m_result_ptr : nullptr;
```
dalg24, line 621: This one is not obvious to me.  Can you explain why this change is the right thing.

masterleinad: You're right, that should not have been changed. Fixed.

dalg24 (approved):

PR #5217, opened 2022-07-14
Fix cmake default compiler flags for unknown compiler

simon-schlepphorst:
Fixes a bug in `kokkos_functions.cmake` where the default compiler options aren't applied if the current compiler isn't defined in ${COMPILERS} list.


cmake/kokkos_functions.cmake
```diff
@@ -850,6 +850,9 @@ FUNCTION(COMPILER_SPECIFIC_OPTIONS_HELPER)
       ELSEIF(PARSE_DEFAULT)
         SET(COMPILER_SPECIFIC_FLAGS_TMP ${PARSE_DEFAULT})
       ENDIF()
+    # Set default flags even if current compiler isn't in ${COMPILERS}
+    ELSEIF(PARSE_DEFAULT)
+      SET(COMPILER_SPECIFIC_FLAGS_TMP ${PARSE_DEFAULT})
```
masterleinad, line 855: I would prefer setting the default flag in line 843 (if one is given) and removing lines 850-852 so that we only override the flag if a specialized value for the compiler is provided.
It looks like the current solution sets the default value multiple times, right?

simon-schlepphorst: Wouldn't that conflict with line 847-848?
"COMPILER_SPECIFIC_FLAGS_TMP" should be empty if "NO-VALUE-SPECIFIED" was found, right?

masterleinad: Yes, I would prefer adding that case as well to the if-clause started in 847.

masterleinad (approved):
Thanks! Looks good to me now.

crtrott (approved):

crtrott:
Confirming this fixes the issue on my Fujitsu platform.

dalg24:
OK ignoring failures due to one of the machine that ran out of disk space

PR #5219, opened 2022-07-14
OpenMPTarget: Adding a workaound for team scan.

rgayatri23:
The PR avoids the non-working team_scan code path.
The strategy it uses is to make one thread per team do the work.

### DIFF
--- a/core/src/OpenMPTarget/Kokkos_OpenMPTarget_Exec.hpp
+++ b/core/src/OpenMPTarget/Kokkos_OpenMPTarget_Exec.hpp
@@ -1639,11 +1639,6 @@ KOKKOS_INLINE_FUNCTION void parallel_scan(
     member.team_broadcast(accum, team_size - 1);
   }
 #endif
-
-  // for (iType i = 0; i < nchunk; ++i) {
-  // if(omp_get_thread_num() == 0)
-  // lambda(i, accum, true);
-  //}
 }
 
 }  // namespace Kokkos
--- a/core/src/OpenMPTarget/Kokkos_OpenMPTarget_Exec.hpp
+++ b/core/src/OpenMPTarget/Kokkos_OpenMPTarget_Exec.hpp
@@ -59,6 +59,13 @@
 #define KOKKOS_IMPL_HIERARCHICAL_REDUCERS_WORKAROUND
 #endif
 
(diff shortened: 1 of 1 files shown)


core/src/OpenMPTarget/Kokkos_OpenMPTarget_Exec.hpp
```diff
@@ -1584,22 +1591,36 @@ KOKKOS_INLINE_FUNCTION void parallel_scan(
...
+    for (iType i = start; i < end; ++i) {
+      lambda(i, scan_val, true);
+    }
+  }
+#pragma omp barrier
+#else
+  const auto nchunk = (end - start + team_size - 1) / team_size;
+  value_type accum  = 0;
+  each team has to process one or
+      more chunks of the prefix scan for (iType i = 0; i < nchunk; ++i) {
```
Rombur, line 1623: You removed the `//` and now the indentation is off. When we will remove the new code, the old code won't compile.

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Exec.hpp
```diff
@@ -1616,6 +1637,12 @@ KOKKOS_INLINE_FUNCTION void parallel_scan(
     // broadcast last value to rest of the team
     member.team_broadcast(accum, team_size - 1);
   }
+#endif
+
+  // for (iType i = 0; i < nchunk; ++i) {
+  // if(omp_get_thread_num() == 0)
+  // lambda(i, accum, true);
+  //}
```
Rombur, line 1645: Does this need to be removed or in the `#else`?

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Exec.hpp
rgayatri23: Thanks. Fixed it. Did not notice that the comments got uncommented in the blocked code path.

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Exec.hpp
rgayatri23: Thats needs to be removed. Done.

Rombur (approved):

crtrott (approved):

PR #5065, opened 2022-05-26
SYCL: Fence memcpy instead of using in-order queues

masterleinad:
This fixes some issues with recent versions of the compiler. Essentially, we can get rid of in-order queues if we only make sure that we fence before `memcpy` which seems the better workaround.

Issue #4793, opened 2022-02-16
Audit code for incomplete use of execution space arguments


PhilMiller:
Every routine in Kokkos that takes an execution space instance argument should consistently use and synchronize solely that execution space, and not launch operations in the default execution space or call the global `Kokkos::fence()`

PhilMiller:
This is potentially testable through use of the tools and profiling interfaces

stanmoore1:
Can do we a similar audit that checks for the ability to pass `Kokkos::WithoutInitializing` into the relevant parts of the API?

PhilMiller:
That sounds like an excellent suggestion - would you open an issue for it?

stanmoore1:
Done--thanks

PR #5220, opened 2022-07-14
Fix DynamicView::resize_serial

masterleinad:
I was running into a case where the call operator for `DynamicView` would deadlock in https://github.com/kokkos/kokkos/blob/902a50d4486faa036bd88cb30db4c139a70c7a24/containers/src/Kokkos_DynamicView.hpp#L435-L438. As discussed on slack, the code there was only useful before the removal of `resize_parallel` in  768f657d7251b3cdbce0de48a75574f9a2cdd29d.
In particular, it's just an error if the chunk pointer we need to access is a `nullptr`. I took the freedom to remove all that logic in https://github.com/kokkos/kokkos/commit/6b7a1a0bf111e94fc0fbc46903d8806538221ab6 to only check if for out-of-bounds errors and not checking the pointer to be a `nullptr` at all.

The origin of the issue was seeing was that the deep copy in `resize_serial` hadn't finished before we were trying to access the `View`. Conceivably, this can happen if the access happens on a different execution space instance than the one used for deep copy. In this case, it was the same execution space instance but the deep copy wasn't properly enqueued (also see https://github.com/kokkos/kokkos/pull/5065). Since we are supposed to fence after calling `DeepCopy` internally anyway, I didn't bother with adding the workaround there as well.
In the end, it was enough to just fence the deep copy operation in `resize_serial`, and to do that I had to provide `ChunkManager::deep_copy_to` with an execution space instance argument. This also allowed (and enforced) using an execution space instance in a couple of other places using `deep_copy_to`.

### DIFF
--- a/containers/src/Kokkos_DynamicView.hpp
+++ b/containers/src/Kokkos_DynamicView.hpp
@@ -414,33 +414,14 @@ class DynamicView : public Kokkos::ViewTraits<DataType, P...> {
         "space");
 
     // Which chunk is being indexed.
-    const uintptr_t ic = uintptr_t(i0 >> m_chunk_shift);
+    const uintptr_t ic = uintptr_t(i0) >> m_chunk_shift;
 
-    typename traits::value_type* volatile* const ch = m_chunks + ic;
-
-    // Do bounds checking if enabled or if the chunk pointer is zero.
-    // If not bounds checking then we assume a non-zero pointer is valid.
-
-#if !defined(KOKKOS_ENABLE_DEBUG_BOUNDS_CHECK)
-    if (nullptr == *ch)
+#if defined(KOKKOS_ENABLE_DEBUG_BOUNDS_CHECK)
+    const uintptr_t n = *reinterpret_cast<uintptr_t*>(m_chunks + m_chunk_max);
+    if (n <= ic) Kokkos::abort("Kokkos::DynamicView array bounds error");
 #endif
-    {
-      // Verify that allocation of the requested chunk is in progress.
 
-      // The allocated chunk counter is m_chunks[ m_chunk_max ]
-      const uintptr_t n =
-          *reinterpret_cast<uintptr_t volatile*>(m_chunks + m_chunk_max);
-
-      if (n <= ic) {
-        Kokkos::abort("Kokkos::DynamicView array bounds error");
-      }
-
-      // Allocation of this chunk is in progress
-      // so wait for allocation to complete.
-      while (nullptr == *ch)
-        ;
-    }
-
+    typename traits::value_type** const ch = m_chunks + ic;
     return (*ch)[i0 & m_chunk_mask];
   }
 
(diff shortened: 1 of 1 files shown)


containers/src/Kokkos_DynamicView.hpp
```diff
@@ -821,12 +824,13 @@ inline void deep_copy(const Kokkos::Experimental::DynamicView<T, DP...>& dst,
...
   else
-    src.impl_get_chunks().deep_copy_to(dst.impl_get_chunks());
+    src.impl_get_chunks().deep_copy_to(dst_execution_space{},
+                                       dst.impl_get_chunks());
   Kokkos::fence("Kokkos::deep_copy(DynamicView)");
 }
 
 template <class ExecutionSpace, class T, class... DP, class... SP>
-inline void deep_copy(const ExecutionSpace&,
+inline void deep_copy(const ExecutionSpace& exec,
```
PhilMiller, line 833: Adding execution space support here is helpful to https://github.com/DARMA-tasking/magistrate/pull/202/ so thanks!

PhilMiller: Also advances #4793

PhilMiller (approved):

containers/src/Kokkos_DynamicView.hpp
```diff
@@ -564,7 +551,23 @@ class DynamicView : public Kokkos::ViewTraits<DataType, P...> {
       m_chunks_host.template allocate_with_destroy<device_space>(
           label, m_chunks.get_ptr());
       m_chunks_host.initialize();
-      m_chunks_host.deep_copy_to(m_chunks);
+
+      // Add some properties if not provided to avoid need for if constexpr
+      using alloc_prop_input = Kokkos::Impl::ViewCtorProp<Prop...>;
+      using alloc_prop       = Kokkos::Impl::ViewCtorProp<
+          Prop..., std::conditional_t<alloc_prop_input::has_execution_space,
+                                      std::integral_constant<unsigned int, 15>,
```
ldh4, line 559: Where does this '15' come from?

crtrott: oh yeah I overlooked this at first. this seems a bit hacky lol. I get what this is doing but really? What if we want to check in ViewCtorProp that all the arguments you get are actually expected. Maybe at least make it void_t?

crtrott (changes requested):

containers/src/Kokkos_DynamicView.hpp
masterleinad: It's a random integer that we haven't used before so that we can add a dummy argument. This approach is used pretty much everywhere where you need to make sure that `ViewCtorProp` has a certain property.

masterleinad: https://github.com/kokkos/kokkos/blob/5c7a4e3e36912201cbb3a2ec3fe2f3e661c77e84/core/src/impl/Kokkos_ViewCtor.hpp#L115-L126

ldh4 (approved):

PR #5221, opened 2022-07-14
Refactor parsing of command line arguments and environment variables

dalg24:
* Enforce same rule everywhere to parse booleans
  case insensitive `(1,yes,true)` -> `true`
  case insensitive `(0,no,false)` -> `false`
  otherwise call `Kokkos::abort()`
* Changes in behavior
  `--kokkos-disable-warnings=false` would yield `disable_warnings=true` and now actually means `false`
  `KOKKOS_DISABLE_WARNINGS=ON` would yield `disable_warnings=true` but now is not recognized and causes abnormal termination of the program
  `--kokkos-num-devices=3 --num-devices=4` would yield `num_devices=3` and now give `num_devices=4`

### DIFF
--- a/core/src/impl/Kokkos_Core.cpp
+++ b/core/src/impl/Kokkos_Core.cpp
@@ -487,7 +487,7 @@ int Kokkos::Impl::get_gpu(const InitializationSettings& settings) {
     return visible_devices[0];
   }
   // map_device_id provided
-  // either random or round-robin assignement based on local MPI rank
+  // either random or round-robin assignment based on local MPI rank
   if (!is_valid_map_device_id_by(settings.get_map_device_id_by())) {
     std::stringstream ss;
     ss << "Warning: map_device_id_by setting '"
@@ -521,7 +521,7 @@ int Kokkos::Impl::get_gpu(const InitializationSettings& settings) {
     return visible_devices[0];
   }
 
-  // use device assigned by CTest when ressource allocation is activated
+  // use device assigned by CTest when resource allocation is activated
   if (std::getenv("CTEST_KOKKOS_DEVICE_TYPE") &&
       std::getenv("CTEST_RESOURCE_GROUP_COUNT")) {
     return get_ctest_gpu(local_rank_str);
--- a/core/src/impl/Kokkos_Core.cpp
+++ b/core/src/impl/Kokkos_Core.cpp
@@ -910,7 +910,7 @@ void Kokkos::Impl::parse_environment_variables(
       std::stringstream ss;
       ss << "Error: environment variable 'KOKKOS_NUM_THREADS=" << num_threads
          << "' is invalid."
-         << " The number of threads must be greater equal to one."
+         << " The number of threads must be greater than or equal to one."
          << " Raised by Kokkos::initialize(int argc, char* argv[]).\n";
       Kokkos::abort(ss.str().c_str());
     }
@@ -922,7 +922,7 @@ void Kokkos::Impl::parse_environment_variables(
       std::stringstream ss;
       ss << "Error: environment variable 'KOKKOS_DEVICE_ID" << device_id
          << "' is invalid."
-         << " The device id must be greater equal to zero.\n";
+         << " The device id must be greater than or equal to zero.\n";
       Kokkos::abort(ss.str().c_str());
     }
     settings.set_device_id(device_id);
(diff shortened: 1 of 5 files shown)


crtrott:
Yeah abort.

nliber (approved):

core/src/impl/Kokkos_Core.cpp
```diff
@@ -176,6 +176,10 @@ unsigned get_process_id() {
 #endif
 }
 
+bool is_valid_num_threads(int x) { return x > 0; }
+
+bool is_valid_device_id(int x) { return x >= 0; }
```
ldh4, line 181: Should there also be a check here against the upper limit of valid device ids?

core/src/impl/Kokkos_Core.cpp
```diff
@@ -924,133 +901,82 @@ void Kokkos::Impl::parse_environment_variables(
...
-          "Error: cannot specify both KOKKOS_NUM_DEVICES and "
-          "KOKKOS_RAND_DEVICES. "
-          "Raised by Kokkos::initialize(int argc, char* argv[]).");
+  int num_threads;
+  if (check_env_int("KOKKOS_NUM_THREADS", num_threads)) {
+    if (!is_valid_num_threads(num_threads)) {
+      std::stringstream ss;
+      ss << "Error: environment variable 'KOKKOS_NUM_THREADS=" << num_threads
+         << "' is invalid."
+         << " The number of threads must be greater equal to one."
```
ldh4, line 913:
```suggestion
         << " The number of threads must be greater than or equal to one."
```

core/src/impl/Kokkos_Core.cpp
```diff
@@ -924,133 +901,82 @@ void Kokkos::Impl::parse_environment_variables(
...
-      settings.set_num_devices(env_rand_devices);
+    settings.set_num_threads(num_threads);
+  }
+  int device_id;
+  if (check_env_int("KOKKOS_DEVICE_ID", device_id)) {
+    if (!is_valid_device_id(device_id)) {
+      std::stringstream ss;
+      ss << "Error: environment variable 'KOKKOS_DEVICE_ID" << device_id
+         << "' is invalid."
+         << " The device id must be greater equal to zero.\n";
```
ldh4, line 925:
```suggestion
         << " The device id must be greater than or equal to zero.\n";
```

core/src/impl/Kokkos_Core.cpp
dalg24: The limit is the sky

core/src/impl/Kokkos_Core.cpp
```diff
@@ -724,74 +728,59 @@ Report bugs to https://github.com/kokkos/kokkos/issues
...
-      if (check_arg(argv[iarg], "--device-id")) {
-        warn_deprecated_command_line_argument("--device-id",
-                                              "--kokkos-device-id");
-      }
-      if (check_arg(argv[iarg], "--device")) {
-        warn_deprecated_command_line_argument("--device", "--kokkos-device-id");
+      if (!is_valid_device_id(device_id)) {
+        std::stringstream ss;
+        ss << "Error: command line argument '" << argv[iarg] << "' is invalid."
+           << " The device id must be greater equal to zero.\n";
```
ldh4, line 782:
```suggestion
           << " The device id must be greater than or equal to zero.\n";
```

core/src/impl/Kokkos_Core.cpp
```diff
@@ -724,74 +728,59 @@ Report bugs to https://github.com/kokkos/kokkos/issues
...
+        warn_deprecated_command_line_argument(get_flag(argv[iarg]),
                                               "--kokkos-num-threads");
       }
-      if (check_arg(argv[iarg], "--threads")) {
-        warn_deprecated_command_line_argument("--threads",
-                                              "--kokkos-num-threads");
+      if (!is_valid_num_threads(num_threads)) {
+        std::stringstream ss;
+        ss << "Error: command line argument '" << argv[iarg] << "' is invalid."
+           << " The number of threads must be greater equal to one."
```
ldh4, line 766:
```suggestion
           << " The number of threads must be greater than or equal to one."
```

ldh4 (approved):

crtrott (approved):

PR #5222, opened 2022-07-15
Harmonize device selection and allow SYCL to circumvent it when no GPU is available

dalg24:
#5211 may have broken SYCL running on CPUs
This PR make sure backends do not try to interpret on their own the settings and use directly the GPU that is selected by Core.  It allows SYCL to not call `Impl::get_gpu` when it detects there are none.

### DIFF
--- a/core/src/Cuda/Kokkos_Cuda_Instance.cpp
+++ b/core/src/Cuda/Kokkos_Cuda_Instance.cpp
@@ -860,9 +860,7 @@ int Cuda::impl_is_initialized() {
 }
 
 void Cuda::impl_initialize(InitializationSettings const &settings) {
-  int const use_gpu = Impl::get_gpu(settings);
-  Impl::CudaInternal::singleton().initialize(use_gpu > -1 ? use_gpu : 0,
-                                             nullptr);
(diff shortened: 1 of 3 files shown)


masterleinad (commented):
Testing current `develop`, the behavior for initializing SYCL seems as I intended it to be. If you are requesting a specific device it is assumed to be a GPU and we fail if no GPUs are available. This pull request changes that behavior to silently ignore the device id.
I don't necessarily think this change in behavior is a problem; we should just be conscious of it.

dalg24:
> Testing current `develop`, the behavior for initializing SYCL seems as I intended it to be. If you are requesting a specific device it is assumed to be a GPU and we fail if no GPUs are available. This pull request changes that behavior to silently ignore the device id.
> I don't necessarily think this change in behavior is a problem; we should just be conscious of it.

I updated the PR to only consider the fallback when no device id was specified

masterleinad (approved):
Now, this pull request changes the error message when requesting a specific device id when there are no GPUs from
```
  what():  Requested GPU with id 0 but only 0 GPU(s) available!
``` 
to 
```
Error: no GPU available for execution.
```
which is fine with me.

PR #5224, opened 2022-07-15
Use `[[nodiscard]]` on `ScopeGuard`

msimberg:
Unsure if I'm supposed to include `Kokkos_Macros.hpp` directly for the macro or if it's enough to get it through `Kokkos_Core_fwd.hpp`.

### DIFF
--- a/core/src/Kokkos_Core.hpp
+++ b/core/src/Kokkos_Core.hpp
@@ -195,8 +195,11 @@ namespace Kokkos {
  * Kokkos::initialize or Kokkos::finalize it is not copyable or assignable
  */
 
-class ScopeGuard {
+class KOKKOS_ATTRIBUTE_NODISCARD ScopeGuard {
  public:
+#if defined(__has_cpp_attribute) && __has_cpp_attribute(nodiscard) >= 201907
+  KOKKOS_ATTRIBUTE_NODISCARD
+#endif
   ScopeGuard(int& argc, char* argv[]) {
     sg_init = false;
     if (!Kokkos::is_initialized()) {
@@ -205,6 +208,9 @@ class ScopeGuard {
     }
   }
 
+#if defined(__has_cpp_attribute) && __has_cpp_attribute(nodiscard) >= 201907
+  KOKKOS_ATTRIBUTE_NODISCARD
+#endif
   ScopeGuard(
       const InitializationSettings& settings = InitializationSettings()) {
     sg_init = false;


masterleinad (commented):
I'm not sure how this would help with the most common use case/user error, see  https://godbolt.org/z/EfvGba7K6.
We would only warn for a function returning `ScopeGuard` and not storing its return type but constructing an unnamed object of type `ScopeGuard` wouldn't raise a warning.

masterleinad (approved):
It seems to help at least with `clang`, see https://godbolt.org/z/Tv49qnPT8.

masterleinad (changes requested):
Having also the constructor as [[nodiscard]] indeed seems to help at least for clang++, g++, and msvc, see https://godbolt.org/z/absMzfoW7. Please add it there, too.

dalg24 (dismissed):

masterleinad (approved):
Thanks!

msimberg:
> According to https://isocpp.org/std/standing-documents/sd-6-sg10-feature-test-recommendations#nodiscard, we should use
> 
> ```
> #if __has_cpp_attribute(nodiscard) >= 201907
> ```

Thanks for looking that up. Pushed that change but I haven't checked locally if nvcc reports that correctly...

msimberg:
Guarded by both `KOKKOS_ENABLE_CXX17/20` and `__has_cpp_attribute` because of  https://github.com/msimberg/kokkos/runs/7390818883?check_suite_focus=true#step:4:12.

In principle there could be also a separate macro to use the `201603` version of `nodiscard` outside of constructors, but do you think that would be useful?

core/src/Kokkos_Macros.hpp
```diff
@@ -633,7 +633,8 @@ static constexpr bool kokkos_omp_on_host() { return false; }
 #define KOKKOS_ENABLE_CUDA_LDG_INTRINSIC
 #endif
 
-#if defined(KOKKOS_ENABLE_CXX17) || defined(KOKKOS_ENABLE_CXX20)
+#if (defined(KOKKOS_ENABLE_CXX17) || defined(KOKKOS_ENABLE_CXX20)) && \
+    (__has_cpp_attribute(nodiscard) >= 201907)
```
masterleinad, line 637: I think we should just have
```
KOKKOS_ATTRIBUTE_NODISCARD
#endif
```
in front of the constructors and leave the macro alone otherwise.

msimberg:
> > In principle there could be also a separate macro to use the 201603 version of nodiscard outside of constructors, but do you think that would be useful?
> 
> I think we should also provide the macro when ` __has_cpp_attribute(nodiscard)` is less than `201907` (but don't use it for constructors).

What I meant is that one could have:
```
#if defined(__has_cpp_attribute) && __has_cpp_attribute(nodiscard) >= 201603
#define KOKKOS_ATTRIBUTE_NODISCARD [[nodiscard]]
#endif

#define KOKKOS_ATTRIBUTE_NODISCARD_CONSTRUCTOR_OR_2019_OR_SOME_OTHER_APPROPRIATE_NAME [[nodiscard]]
#endif
```

That would avoid having the `#ifdef` at the constructor definition and is a bit nicer if `nodiscard` is useful in other classes in Kokkos as well. On the other hand, it requires knowing that there are two versions of the macro and one can still accidentally use the first macro on constructors (though CI and warnings would hopefully make sure using the wrong one macro would not end up on develop/master).

masterleinad:
We will require C++17 after the 3.7 release anyway and can then replace the current `KOKKOS_ATTRIBUTE_NODISCARD` with `[[nodiscard]]` (and possibly get rid of the macro altogether). Thus, I don't think there is much of a point in modifying the macro.

I'm fine with adding
```
#define KOKKOS_IMPL_ATTRIBUTE_NODISCARD_CONSTRUCTOR [[nodiscard]]
#else
#define KOKKOS_IMPL_ATTRIBUTE_NODISCARD_CONSTRUCTOR
#endif
```
if you prefer that over having the condition at both constructors but wouldn't want to add a macro that users would be encouraged to use downstream.

msimberg:
> We will require C++17 after the 3.7 release anyway and...

Fair enough. In that case it's simpler to just put the ifdefs at the constructors. I've updated this accordingly.

core/src/Kokkos_Macros.hpp
msimberg: Updated according to this suggestion.

masterleinad (approved):

PhilMiller (approved):

PR #5228, opened 2022-07-15
Check execution spaces interface as they are registered to core

dalg24:
Compile-time introspection of the execution spaces to enforce a common API.
The check is only done at compile-time in our own source file and does not impact user code.

### DIFF
--- a/core/src/impl/Kokkos_ExecSpaceManager.hpp
+++ b/core/src/impl/Kokkos_ExecSpaceManager.hpp
@@ -93,7 +93,9 @@ constexpr bool check_valid_execution_space() {
   STATIC_ASSERT(is_detected<print_configuration_t, ExecutionSpace>::value);
   STATIC_ASSERT(is_detected<initialize_finalize_t, ExecutionSpace>::value);
   STATIC_ASSERT(is_detected<fence_t, ExecutionSpace>::value);
+#ifndef KOKKOS_ENABLE_HPX  // FIXME_HPX
   STATIC_ASSERT(sizeof(ExecutionSpace) <= 2 * sizeof(void*));
+#endif
   return true;
 }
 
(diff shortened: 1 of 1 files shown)


crtrott (changes requested):
Looks pretty good, just disable the size check for HPX for now.

msimberg:
> Looks pretty good, just disable the size check for HPX for now.

I suppose it's this check: https://github.com/kokkos/kokkos/blob/9d6dd28eebb935a480aecd2520ae93b7f80b6c66/core/src/impl/Kokkos_ExecSpaceManager.hpp#L96? What is that about? Just a general guarantee that execution space instances are e.g. cheap enough to pass by value or something like that?

In any case, I think getting it down to 2 pointers shouldn't be too hard, and may already have happened in a branch that I'm working on. Not for this PR though of course...

dalg24:
> Just a general guarantee that execution space instances are e.g. cheap enough to pass by value or something like that?

Yes exactly.  That what we have been teaching users.
We use the "pImpl" idiom.  In all backends that actually support execution space instances, the only data member is an `Impl::HostSharedPtr` to the internal implementation details.  When the execution space is the default one you point to the singleton and "disengage" the destructor.

> In any case, I think getting it down to 2 pointers shouldn't be too hard, and may already have happened in a branch that I'm working on. Not for this PR though of course...

That would be great if you can do so.  We can merge with that assertion guarded buy `#ifndef KOKKOS_ENABLE_HPX` as Christian suggested in the meantime.

msimberg:
> > Just a general guarantee that execution space instances are e.g. cheap enough to pass by value or something like that?
> 
> Yes exactly. That what we have been teaching users. We use the "pImpl" idiom. In all backends that actually support execution space instances, the only data member is an `Impl::HostSharedPtr` to the internal implementation details. When the execution space is the default one you point to the singleton and "disengage" the destructor.

Sounds good. I think that's more or less what I've done on that other branch. Currently the HPX backend unnecessarily holds separate references to things that could be accessed through a single reference.

cz4rs (approved):
Looks good to me.

crtrott (approved):

PR #5229, opened 2022-07-15
Dispatch Kokkos::sort(Kokkos::View) to SYCL oneDPL

jstrzebonski:


masterleinad:
Also, see https://github.com/arborx/ArborX/blob/baf4d9480d3a0f7470826011d7a806033c4d02c3/docker/Dockerfile.sycl#L63-L78.

CMakeLists.txt
```diff
@@ -291,6 +291,12 @@ IF (NOT KOKKOS_HAS_TRILINOS AND NOT Kokkos_INSTALL_TESTING)
 ENDIF()
 INCLUDE(${KOKKOS_SRC_PATH}/cmake/kokkos_install.cmake)
 
+# TODO (STRZ):do it gracefully
+IF (KOKKOS_ENABLE_SYCL)
+  FIND_PACKAGE(oneDPL)
+  TARGET_LINK_LIBRARIES(kokkos INTERFACE oneDPL)
+ENDIF()
+
```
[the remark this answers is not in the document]
masterleinad: I would make it a dependent option and include it as a regular tpl in `cmake/kokkos_tpls.cmake` similarly to how we treat `ROCM`.

scripts/docker/Dockerfile.sycl
```diff
@@ -53,3 +53,19 @@ RUN SYCL_VERSION=20220112 && \
...
+RUN ONE_DPL_VERSION=oneDPL-2021.6.0-release && \
+    ONE_DPL_URL=https://github.com/oneapi-src/oneDPL/archive && \
+    ONE_DPL_ARCHIVE=${ONE_DPL_VERSION}.tar.gz && \
+    SCRATCH_DIR=/scratch && mkdir -p ${SCRATCH_DIR} && cd ${SCRATCH_DIR} && \
+    wget --quiet ${ONE_DPL_URL}/${ONE_DPL_ARCHIVE} && \
+    mkdir onedpl && \
+    tar -xf ${ONE_DPL_ARCHIVE} -C onedpl --strip-components=1 && cd onedpl && \
+    echo 'install(CODE "set(OUTPUT_DIR \"${CMAKE_INSTALL_PREFIX}/lib/cmake/oneDPL\")")' >> CMakeLists.txt && \
+    echo 'install(SCRIPT cmake/scripts/generate_config.cmake)' >> CMakeLists.txt && \
+    echo 'install(DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}/include DESTINATION linux)' >> CMakeLists.txt && \
```
masterleinad, line 67: We shouldn't need this in newer releases of `oneDPL` (after https://github.com/oneapi-src/oneDPL/pull/294).

algorithms/src/Kokkos_Sort.hpp
```diff
@@ -628,6 +633,24 @@ std::enable_if_t<Kokkos::is_execution_space<ExecutionSpace>::value> sort(
...
+#if defined(KOKKOS_ENABLE_SYCL)
+template <class DataType, class... Properties>
+void sort(const Experimental::SYCL& space,
+          const Kokkos::View<DataType, Properties...>& view,
+          bool const = false) {
+  using ViewType = Kokkos::View<DataType, Properties...>;
+  static_assert(
+      ViewType::rank == 1 &&
+          !std::is_same<typename ViewType::array_layout, LayoutStride>::value,
+      "SYCL sort only supports contiguous 1D Views.");
```
jstrzebonski, line 645: Should I also assert if the view's memory_space is accessible from SYCL?

CMakeLists.txt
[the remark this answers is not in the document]
jstrzebonski: Done.

scripts/docker/Dockerfile.sycl
jstrzebonski: Removed.

masterleinad:
Retest this please.

algorithms/src/Kokkos_Sort.hpp
```diff
@@ -53,6 +53,11 @@
 
 #include <algorithm>
 
+#if defined(KOKKOS_ENABLE_SYCL)
```
masterleinad, line 56: This should check if the `oneDPL` TPL was enabled.

cmake/Modules/FindTPLONEDPL.cmake
```diff
@@ -0,0 +1,4 @@
+FIND_PACKAGE(oneDPL)
+IF (oneDPL_FOUND)
+  KOKKOS_CREATE_IMPORTED_TPL(ONEDPL INTERFACE LINK_LIBRARIES oneDPL)
+ENDIF()
```
masterleinad, line 4: When you use an `oneAPI` installation, it doesn't provide a find module but still includes the headers and libraries.
Also, we should require finding the package here if necessary. Thus, I would do something like
```suggestion
INCLUDE(CheckIncludeFileCXX)
CHECK_INCLUDE_FILE_CXX(oneapi/dpl/execution KOKKOS_COMPILER_HAS_ONEDPL_EXECUTION_HEADER)
CHECK_INCLUDE_FILE_CXX(oneapi/dpl/algorithm KOKKOS_COMPILER_HAS_ONEDPL_ALGORITHM_HEADER)

IF (KOKKOS_COMPILER_HAS_ONEDPL_EXECUTION_HEADER AND KOKKOS_COMPILER_HAS_ONEDPL_ALGORITHM_HEADER)
  KOKKOS_CREATE_IMPORTED_TPL(ONEDPL INTERFACE)
ELSE()
  FIND_PACKAGE(oneDPL REQUIRED)
  KOKKOS_CREATE_IMPORTED_TPL(ONEDPL INTERFACE LINK_LIBRARIES oneDPL)
ENDIF()
```
instead.

scripts/docker/Dockerfile.sycl
```diff
@@ -10,6 +10,7 @@ RUN apt-get update && apt-get install -y \
         ninja-build \
         python3 \
         git \
+        libomp-dev \
```
masterleinad, line 13: Do you still need this when you are only building for the `dpcpp` backend? Note, that this build doesn't even support `OpenMP` to be enabled in `Kokkos` AFAIK.

masterleinad (changes requested):

algorithms/src/Kokkos_Sort.hpp
```diff
@@ -628,6 +633,24 @@ std::enable_if_t<Kokkos::is_execution_space<ExecutionSpace>::value> sort(
...
+template <class DataType, class... Properties>
+void sort(const Experimental::SYCL& space,
+          const Kokkos::View<DataType, Properties...>& view,
+          bool const = false) {
+  using ViewType = Kokkos::View<DataType, Properties...>;
+  static_assert(
+      ViewType::rank == 1 &&
+          !std::is_same<typename ViewType::array_layout, LayoutStride>::value,
```
PhilMiller, line 644: I think the check should be affirmative that the layout is contiguous, rather than excluding one known bad case

masterleinad:
Also, it's worthwhile keeping https://stackoverflow.com/a/67924408 in mind. I'm not quite sure if we should any kind of workaround here but I always have to deal with it.

algorithms/src/Kokkos_Sort.hpp
jstrzebonski: Do I need it even though oneDPL is `REQUIRED` when `KOKKOS_ENABLE_SYCL` is ON? And if so, should I then add to `FindTPLONEDPL.cmake` something like `add_definitions(-DKOKKOS_ENABLE_ONEDPL)` or is there a better way in Kokkos-cmake-environment?

algorithms/src/Kokkos_Sort.hpp
jstrzebonski: Done.

cmake/Modules/FindTPLONEDPL.cmake
jstrzebonski: Done. Thanks for that!

algorithms/src/Kokkos_Sort.hpp

jstrzebonski:
> Also, it's worthwhile keeping https://stackoverflow.com/a/67924408 in mind. I'm not quite sure if we should any kind of workaround here but I always have to deal with it.

If I understood correctly, changing the order of includes should be sufficient. And that's what I did.

algorithms/src/Kokkos_Sort.hpp
```diff
@@ -51,6 +51,14 @@
 
 #include <Kokkos_Core.hpp>
 
+// The order of includes (firstly oneDPL's, than std's) because of:
+// https://stackoverflow.com/questions/67923287/how-to-resolve-no-member-named-task-in-namespace-tbb-error-when-using-oned/67924408#67924408
```
masterleinad, line 55: It likely doesn't matter that much because we can't control if `<algorithms>` is included before the oneDPL headers in a downstream application anyway.

algorithms/src/Kokkos_Sort.hpp
masterleinad: It's not `REQUIRED` if the `oneDPL` can be found natively by the compiler. Add 
```
#cmakedefine KOKKOS_ENABLE_ONEDPL
```
to `cmake/KokkosCore_config.h.in`.

masterleinad:
> If I understood correctly, changing the order of includes should be sufficient. And that's what I did.

User code can still include `<algorithm>` before any `Kokkos` header. I don't mind keeping the comment, though.

scripts/docker/Dockerfile.sycl
jstrzebonski: Without it, I get a linker error
```
[ 11%] Linking CXX executable KokkosCore_UnitTest_CMakePassCmdLineArgs
/usr/bin/ld: cannot find -lomp
clang-14: error: linker command failed with exit code 1 (use -v to see invocation)
core/unit_test/CMakeFiles/KokkosCore_UnitTest_CMakePassCmdLineArgs.dir/build.make:105: recipe for target 'core/unit_test/KokkosCore_UnitTest_CMakePassCmdLineArgs' failed
make[2]: *** [core/unit_test/KokkosCore_UnitTest_CMakePassCmdLineArgs] Error 1
CMakeFiles/Makefile2:1890: recipe for target 'core/unit_test/CMakeFiles/KokkosCore_UnitTest_CMakePassCmdLineArgs.dir/all' failed
make[1]: *** [core/unit_test/CMakeFiles/KokkosCore_UnitTest_CMakePassCmdLineArgs.dir/all] Error 2

```

algorithms/src/Kokkos_Sort.hpp
jstrzebonski: Yeah, I was a bit too fast with that :- )

algorithms/src/Kokkos_Sort.hpp
```diff
@@ -628,6 +633,24 @@ std::enable_if_t<Kokkos::is_execution_space<ExecutionSpace>::value> sort(
   bin_sort.sort(exec, view);
 }
 
+#if defined(KOKKOS_ENABLE_SYCL)
```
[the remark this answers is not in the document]
jstrzebonski: Done.

algorithms/src/Kokkos_Sort.hpp
jstrzebonski: Thanks. Done.

algorithms/src/Kokkos_Sort.hpp
jstrzebonski: I've put `COMPILE_DEFINITIONS PSTL_USE_PARALLEL_POLICIES=0 _GLIBCXX_USE_TBB_PAR_BACKEND=0` into oneDPL TPL cmake.

masterleinad: Let's see how people like that. :slightly_smiling_face:

algorithms/src/Kokkos_Sort.hpp
```diff
@@ -628,6 +636,30 @@ std::enable_if_t<Kokkos::is_execution_space<ExecutionSpace>::value> sort(
   bin_sort.sort(exec, view);
 }
 
+#if defined(KOKKOS_ENABLE_ONEDPL)
+template <class DataType, class... Properties>
+void sort(const Experimental::SYCL& space,
+          const Kokkos::View<DataType, Properties...>& view,
+          bool const = false) {
```
masterleinad, line 643: The discussion related to https://github.com/kokkos/kokkos/pull/5372 also applies here.

fnrizzi (approved):

cmake/Modules/FindTPLONEDPL.cmake
```diff
@@ -0,0 +1,14 @@
...
+CHECK_INCLUDE_FILE_CXX(oneapi/dpl/algorithm KOKKOS_COMPILER_HAS_ONEDPL_ALGORITHM_HEADER)
+
+IF (KOKKOS_COMPILER_HAS_ONEDPL_EXECUTION_HEADER AND KOKKOS_COMPILER_HAS_ONEDPL_ALGORITHM_HEADER)
+  KOKKOS_CREATE_IMPORTED_TPL(ONEDPL INTERFACE)
+ELSE()
+  FIND_PACKAGE(oneDPL REQUIRED)
+  KOKKOS_CREATE_IMPORTED_TPL(
+    ONEDPL INTERFACE
+    LINK_LIBRARIES oneDPL
+    COMPILE_DEFINITIONS PSTL_USE_PARALLEL_POLICIES=0 _GLIBCXX_USE_TBB_PAR_BACKEND=0
```
fnrizzi, line 12: I am not super familiar with this, but is there an alternative?

masterleinad: If you hit the problem because you are using a gcc-9 or gcc-10 toolchain, there isn't much else users can do (apart from reordering include files which is brittle). The question is whether we should try to fix it by disabling the TBB backend for functions in `<algorithm>` or let the user figure it out themselves in case they are affected.

algorithms/src/Kokkos_Sort.hpp
jstrzebonski: Resolved.

cmake/Modules/FindTPLONEDPL.cmake
masterleinad: I'm OK with this since codes that are using `Kokkos` are probably not interested in using parallel stl but we need at the very least a `FIMXE` so we can remove that at some point.

crtrott (changes requested):
Putting a hold on this: I want to understand the implications better and have a discussion on this. Adding these compile definitions feels fairly iffy. At a minimum we need a big fat warning about this, and an option to overwrite this behavior.

scripts/docker/Dockerfile.sycl
```diff
@@ -62,3 +63,18 @@ RUN SYCL_VERSION=20220112 && \
...
+RUN ONE_DPL_VERSION=oneDPL-2021.7.0-release && \
+    ONE_DPL_URL=https://github.com/oneapi-src/oneDPL/archive && \
+    ONE_DPL_ARCHIVE=${ONE_DPL_VERSION}.tar.gz && \
+    SCRATCH_DIR=/scratch && mkdir -p ${SCRATCH_DIR} && cd ${SCRATCH_DIR} && \
+    wget --quiet ${ONE_DPL_URL}/${ONE_DPL_ARCHIVE} && \
+    mkdir onedpl && \
+    tar -xf ${ONE_DPL_ARCHIVE} -C onedpl --strip-components=1 && cd onedpl && \
+    mkdir build && cd build && \
+    cmake -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_CXX_FLAGS="-w" -DCMAKE_INSTALL_PREFIX=${ONE_DPL_DIR} -DCMAKE_SKIP_INSTALL_ALL_DEPENDENCY=TRUE -DONEDPL_BACKEND="dpcpp_only" .. && \
+    make -j${NPROCS} install && \
```
dalg24, line 79:
```suggestion
    make -j${NPROC} install && \
```

jstrzebonski: Fixed.

jstrzebonski:
> Putting a hold on this: I want to understand the implications better and have a discussion on this. Adding these compile definitions feels fairly iffy. At a minimum we need a big fat warning about this, and an option to overwrite this behavior.

@crtrott any news on this?

masterleinad:
The `oneTBB`  release notes say in https://www.intel.com/content/www/us/en/developer/articles/release-notes/intel-oneapi-threading-building-blocks-release-notes.html:
> An application using Parallel STL algorithms in libstdc++ versions 9 and 10 may fail to compile due to incompatible interface changes between earlier versions of Threading Building Blocks (TBB) and oneAPI Threading Building Blocks (oneTBB). Disable support for Parallel STL algorithms by defining PSTL_USE_PARALLEL_POLICIES (in libstdc++ 9) or _GLIBCXX_USE_TBB_PAR_BACKEND (in libstdc++ 10) macro to zero before inclusion of the first standard header file in each translation unit.

masterleinad (approved):

algorithms/src/Kokkos_Sort.hpp
```diff
@@ -634,6 +640,29 @@ sort(const ExecutionSpace& exec,
...
+#if defined(KOKKOS_ENABLE_ONEDPL)
+template <class DataType, class... Properties>
+void sort(const Experimental::SYCL& space,
+          const Kokkos::View<DataType, Properties...>& view) {
+  using ViewType = Kokkos::View<DataType, Properties...>;
+  static_assert(
+      ViewType::rank == 1 &&
+          (std::is_same<typename ViewType::array_layout, LayoutRight>::value ||
+           std::is_same<typename ViewType::array_layout, LayoutLeft>::value),
+      "SYCL sort only supports contiguous 1D Views.");
```
dalg24, line 652: I think the message is misleading.
This is a restriction for `Kokkos::sort` not something specific from `SYCL`.  We get it transitively from calling `Experimental::{begin,end}(view)` in other specializations of the algorithm.

algorithms/src/Kokkos_Sort.hpp
```diff
@@ -634,6 +640,29 @@ sort(const ExecutionSpace& exec,
...
+
+  static_assert(SpaceAccessibility<Experimental::SYCL,
+                                   typename ViewType::memory_space>::accessible,
+                "SYCL execution space is not able to access the memory space "
+                "of the View argument!");
+
+  auto queue  = space.impl_internal_space_instance()->m_queue.value();
+  auto policy = oneapi::dpl::execution::make_device_policy(queue);
+  const int n = view.extent(0);
+  oneapi::dpl::sort(policy, view.data(), view.data() + n);
```
dalg24, line 662: Why are we using raw pointers rather than iterators?

algorithms/src/Kokkos_Sort.hpp
```diff
@@ -66,6 +65,13 @@
 
 #endif
 
+#if defined(KOKKOS_ENABLE_ONEDPL)
+#include <oneapi/dpl/execution>
+#include <oneapi/dpl/algorithm>
+#endif
+
+#include <algorithm>
```
dalg24, line 73: Why did you move this include?

scripts/docker/Dockerfile.sycl
```diff
@@ -62,3 +63,18 @@ RUN SYCL_VERSION=20220112 && \
     echo "${SYCL_DIR}/lib" > /etc/ld.so.conf.d/sycl.conf && ldconfig && \
     rm -rf ${SCRATCH_DIR}
 ENV PATH=${SYCL_DIR}/bin:$PATH
+
+ARG NPROC=8
+
+ENV ONE_DPL_DIR=/opt/onedpl
+RUN ONE_DPL_VERSION=oneDPL-2021.7.0-release && \
```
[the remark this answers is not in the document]
masterleinad: If CI passes, that is good enough for me. The latest version is https://github.com/oneapi-src/oneDPL/releases/tag/oneDPL-2021.7.1-release released on 11/08/2022 but I don't know if that one works with the compiler version used in the CI.

algorithms/src/Kokkos_Sort.hpp
masterleinad: To minimize the chances that we get clashing `TBB` and `oneTBB`. With the preprocessor macros, this is not necessary anymore, though.

algorithms/src/Kokkos_Sort.hpp
masterleinad: It would be interesting to see if it makes a difference in performance.

algorithms/src/Kokkos_Sort.hpp
dalg24: Then please undo this change.

cmake/Modules/FindTPLONEDPL.cmake
dalg24: Please confirm that these definition are always needed regardless of the toolchain that is being used.
If the answer is no, I would prefer if we only added them when absolutely necessary,

masterleinad: As I said in https://github.com/kokkos/kokkos/pull/5229#discussion_r953731053, you need this if the `<algorithm>` header comes from `gcc-9` or `gcc-10`. Since we can't know at configure-time what toolchain users will use in their code, I would just do it unconditionally. (If we are assuming they are using the same toolchain for everything, we could base the decision on `_GLIBCXX_RELEASE` but that seems more fragile to me.)

masterleinad: It's more fragile than assuming the same compiler was used and I'm trying to say that adding a configure-time check(https://github.com/kokkos/kokkos/pull/5229/commits/08b90bb3ef0e872c971dab7f57fb0e7e18f2c60f) adds more fragility to it. Anyway, I added it and also made sure that we add the definitions if the compiler already includes `oneDPL` implicitly.

dalg24: It was a legitimate question, not me asking to change.
BTW would you remind me if we consider using a push_macro pragma.

masterleinad: I tried and couldn't make it work.

algorithms/src/Kokkos_Sort.hpp
```diff
@@ -634,6 +639,23 @@ sort(const ExecutionSpace& exec,
...
+void sort(const Experimental::SYCL& space,
+          const Kokkos::View<DataType, Properties...>& view) {
+  using ViewType = Kokkos::View<DataType, Properties...>;
+
+  static_assert(SpaceAccessibility<Experimental::SYCL,
+                                   typename ViewType::memory_space>::accessible,
+                "SYCL execution space is not able to access the memory space "
+                "of the View argument!");
+
+  auto queue  = space.impl_internal_space_instance()->m_queue.value();
```
crtrott, line 653: don't we have a public thing to get the queue?

crtrott (changes requested):

algorithms/src/Kokkos_Sort.hpp
masterleinad: Here you go!

dalg24:
Retest this please

crtrott (approved):

crtrott:
SYCL didn't pass test with a compile error:

```c++
In file included from /var/jenkins/workspace/Kokkos/core/unit_test/headers_self_contained/tstHeader.cpp:26:

In file included from /var/jenkins/workspace/Kokkos/algorithms/src/Kokkos_Sort.hpp:70:

In file included from /opt/onedpl/include/oneapi/dpl/execution:61:

In file included from /opt/onedpl/include/oneapi/dpl/pstl/algorithm_impl.h:24:

In file included from /opt/onedpl/include/oneapi/dpl/pstl/algorithm_fwd.h:20:

In file included from /opt/onedpl/include/oneapi/dpl/pstl/iterator_defs.h:21:

In file included from /opt/onedpl/include/oneapi/dpl/pstl/utils.h:31:

/opt/onedpl/include/oneapi/dpl/pstl/hetero/dpcpp/sycl_iterator.h:34:68: error: no type named 'buffer_allocator' in namespace 'sycl'; did you mean 'image_allocator'?

template <access_mode Mode, typename T, typename Allocator = sycl::buffer_allocator>

                                                             ~~~~~~^

/opt/intel/oneapi/compiler/2023.0.0/linux/bin-llvm/../include/sycl/image.hpp:65:7: note: 'image_allocator' declared here

using image_allocator = detail::aligned_allocator<byte>;

      ^

1 error generated.

make[2]: *** [core/unit_test/headers_self_contained/CMakeFiles/Kokkos_HeaderSelfContained_Kokkos_Sort_hpp.dir/build.make:76: core/unit_test/headers_self_contained/CMakeFiles/Kokkos_HeaderSelfContained_Kokkos_Sort_hpp.dir/tstHeader.cpp.o] Error 1

make[1]: *** [CMakeFiles/Makefile2:4704: core/unit_test/headers_self_contained/CMakeFiles/Kokkos_HeaderSelfContained_Kokkos_Sort_hpp.dir/all] Error 2

```

WTF ...

masterleinad:
Works now.

cmake/kokkos_tpls.cmake
```diff
@@ -46,6 +46,7 @@ ELSE()
   SET(ROCM_DEFAULT OFF)
 ENDIF()
 KOKKOS_TPL_OPTION(ROCM    ${ROCM_DEFAULT})
+KOKKOS_TPL_OPTION(ONEDPL  ${Kokkos_ENABLE_SYCL})
```
dalg24, line 49: Does that make that if enabling SYCL, configuration would fail if oneDPL cannot be found unless explicitly disabled?  (I don't have an issue with it, just making sure I understand what this does)

masterleinad: Yes, that is what it's doing.

cmake/Modules/FindTPLONEDPL.cmake
```diff
@@ -0,0 +1,46 @@
...
+    KOKKOS_CREATE_IMPORTED_TPL(
+      ONEDPL INTERFACE
+      # https://stackoverflow.com/questions/67923287/how-to-resolve-no-member-named-task-in-namespace-tbb-error-when-using-oned/
+      COMPILE_DEFINITIONS PSTL_USE_PARALLEL_POLICIES=0 _GLIBCXX_USE_TBB_PAR_BACKEND=0
+    )
+  ENDIF()
+ELSE()
+  FIND_PACKAGE(oneDPL REQUIRED)
+
+  IF(KOKKOS_NO_TBB_CONFLICT)
```
dalg24, line 33: How would that ever be true if we don't link the cxx source above against onedpl

masterleinad: We basically just check if we use a libstdc++ version 9 or 10 implementation and assume that there is a conflict if we do. The source code above doesn't require `oneDPL` AFAICT.

dalg24: Sorry didn't look carefully

dalg24 (approved):

Issue #5223, opened 2022-07-15
`Kokkos::RangePolicy` and `Kokkos_AnalyzePolicy` : compilation should fail if unmatched trait provided


romintomasetti:
**Describe the bug**

The following code snippet should fail to compile
```c++
using rp = Kokkos::RangePolicy<Kokkos::HostSpace>
```
because `Kokkos::HostSpace` doesn't match any of the `execution_policy_trait_specifications`, and we though it should fail on https://github.com/kokkos/kokkos/blob/d19aab9981a2c447e832a7b4eb7b16992328fb14/core/src/impl/Kokkos_AnalyzePolicy.hpp#L173.

It didn't fail, and `rp` is therefore equivalent to `Kokkos::RangePolicy<>`.

**Please also include the following items to support reproducing the bug**
1. `nvcc 11.6` and `gcc 9.4.0`
2. Kokkos release or commit used [ea2a5f8bab8478ac2e450548c809e522ac665fab](https://github.com/kokkos/kokkos/commit/ea2a5f8bab8478ac2e450548c809e522ac665fab)
3. `Ubuntu 22`

dalg24:
It gets interpreted as a tag
```C++
  using Policy = Kokkos::RangePolicy<Kokkos::HostSpace>;
  static_assert(std::is_same<Policy::work_tag, Kokkos::HostSpace>::value, "");
```
(No judgement whether it is right or not here, just saying how it is interpreted by Kokkos)

dalg24:
(Sorry for the digression) This reminds me that I contemplated adding a requirement on the tag that `std::is_empty<WorTag>::value` is `true` to avoid surprising users but was hesitant to proceed because it would potentially break user code.
I believe the current requirements, even though not spelled out in the doc nor by a well placed static assertion, are something like being default constructible on host/device, and of course (sigh) that may depend on the backend implementation.

dalg24:
This is fixed in `develop` and will be in release 3.7

PR #5230, opened 2022-07-16
Raise deprecation warnings if non empty `WorkTag` class is used

dalg24:
Drafted our emitting deprecation warnings when user code specifies non empty class as an execution policy `WorkTag` template argument as a tentative fix for #5223 
The intent is to use it as a base for discussion.
```
In file included from <kokkos>/core/src/Kokkos_Core.hpp:57:
In file included from <kokkos>/cmake-build-default/KokkosCore_Config_DeclareBackend.hpp:47:
In file included from <kokkos>/core/src/decl/Kokkos_Declare_THREADS.hpp:49:
In file included from <kokkos>/core/src/Kokkos_Threads.hpp:199:
In file included from <kokkos>/core/src/Kokkos_ExecPolicy.hpp:60:
<kokkos>/core/src/impl/Kokkos_AnalyzePolicy.hpp:224:3: warning: 'warn_if_deprecated' is deprecated: Invalid WorkTag template argument in execution policy!! [-Wdeprecated-declarations]
  KOKKOS_IMPL_STATIC_WARNING(!std::is_empty<typename base_t::work_tag>::value &&
  ^
<kokkos>/core/src/impl/Kokkos_AnalyzePolicy.hpp:216:7: note: expanded from macro 'KOKKOS_IMPL_STATIC_WARNING'
      warn_if_deprecated(std::integral_constant<bool, __VA_ARGS__>()), "")
      ^
<kokkos>/core/src/Kokkos_ExecPolicy.hpp:99:28: note: in instantiation of template class 'Kokkos::Impl::PolicyTraits<Kokkos::HostSpace>' requested here
class RangePolicy : public Impl::PolicyTraits<Properties...> {
                           ^
<kokkos>/core/unit_test/default/TestDefaultDeviceDevelop.cpp:59:30: note: in instantiation of template class 'Kokkos::RangePolicy<Kokkos::HostSpace>' requested here
  static_assert(std::is_same<Policy::work_tag, Kokkos::HostSpace>::value, "");
                             ^
<kokkos>/core/src/impl/Kokkos_AnalyzePolicy.hpp:211:1: note: 'warn_if_deprecated' has been explicitly marked deprecated here
KOKKOS_DEPRECATED_WITH_COMMENT(
^
<kokkos>/core/src/Kokkos_Macros.hpp:610:51: note: expanded from macro 'KOKKOS_DEPRECATED_WITH_COMMENT'
#define KOKKOS_DEPRECATED_WITH_COMMENT(comment) [[deprecated(comment)]]
                                                  ^
1 warning generated.

### DIFF
--- a/core/src/impl/Kokkos_AnalyzePolicy.hpp
+++ b/core/src/impl/Kokkos_AnalyzePolicy.hpp
@@ -220,8 +220,10 @@ struct PolicyTraits
   using base_t =
       ExecPolicyTraitsWithDefaults<AnalyzeExecPolicy<void, Traits...>>;
   using base_t::base_t;
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_3
   KOKKOS_IMPL_STATIC_WARNING(!std::is_empty<typename base_t::work_tag>::value &&
                              !std::is_void<typename base_t::work_tag>::value);
+#endif
 };
 
 #undef KOKKOS_IMPL_STATIC_WARNING
--- a/core/src/impl/Kokkos_AnalyzePolicy.hpp
+++ b/core/src/impl/Kokkos_AnalyzePolicy.hpp
@@ -205,13 +205,26 @@ struct ExecPolicyTraitsWithDefaults : AnalysisResults {
 };
 
 //------------------------------------------------------------------------------
+
+constexpr bool warn_if_deprecated(std::false_type) { return true; }
+KOKKOS_DEPRECATED_WITH_COMMENT(
+    "Invalid WorkTag template argument in execution policy!!")
+constexpr bool warn_if_deprecated(std::true_type) { return true; }
+#define KOKKOS_IMPL_STATIC_WARNING(...) \
+  static_assert(                        \
+      warn_if_deprecated(std::integral_constant<bool, __VA_ARGS__>()), "")
+
 template <typename... Traits>
 struct PolicyTraits
     : ExecPolicyTraitsWithDefaults<AnalyzeExecPolicy<void, Traits...>> {
   using base_t =
       ExecPolicyTraitsWithDefaults<AnalyzeExecPolicy<void, Traits...>>;
   using base_t::base_t;
+  KOKKOS_IMPL_STATIC_WARNING(!std::is_empty<typename base_t::work_tag>::value &&
+                             !std::is_void<typename base_t::work_tag>::value);
 };
+
+#undef KOKKOS_IMPL_STATIC_WARNING
 
 }  // namespace Impl
 }  // namespace Kokkos
--- a/core/src/traits/Kokkos_WorkTagTrait.hpp
+++ b/core/src/traits/Kokkos_WorkTagTrait.hpp
@@ -109,7 +109,10 @@ struct WorkTagTrait : TraitSpecificationBase<WorkTagTrait> {
   //   we should benchmark this assumption if it becomes a problem.
   template <class T>
   using trait_matches_specification = std::integral_constant<
-      bool, !std::is_void<T>::value && /*is_empty<T>::value &&*/
+      bool, !std::is_void<T>::value &&
+#ifndef KOKKOS_ENABLE_DEPRECATED_CODE_3
+                std::is_empty<T>::value &&
+#endif
                 !type_list_any<_trait_matches_spec_predicate<T>::template apply,
                                _exec_policy_traits_without_work_tag>::value>;
 };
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

File: core/src/impl/Kokkos_AnalyzePolicy.hpp [hunk 1]
```diff
@@ -205,14 +205,27 @@ struct ExecPolicyTraitsWithDefaults : AnalysisResults {
 };
 
 //------------------------------------------------------------------------------
+
+constexpr bool warn_if_deprecated(std::false_type) { return true; }
+KOKKOS_DEPRECATED_WITH_COMMENT(
+    "Invalid WorkTag template argument in execution policy!!")
+constexpr bool warn_if_deprecated(std::true_type) { return true; }
```
crtrott, line 212: shouldn't this return false?

crtrott (changes requested):

ajpowelsnl:
@msimberg would you mind having a look at the possibility of a race condition around `SharedHostPointerTracking`?  We are seeing intermittent test failures.  We are not certain that it's a failure on our side, but it could be.

```
 RUN      ] hpx.host_shared_ptr_tracking
/home/runner/work/kokkos/kokkos/kokkos/core/unit_test/TestHostSharedPtrAccessOnDevice.hpp:208: Failure
  2
  f1.use_count()
    Which is: 3
[  FAILED  ] hpx.host_shared_ptr_tracking (0 ms)
```

core/src/impl/Kokkos_AnalyzePolicy.hpp
crtrott: Ah so maybe if deprecated code is disabled it should return false ...

crtrott (changes requested):

dalg24:
Failure is one of the CUDA build on a machine running out of disk space
"Safe" to ignore

crtrott (approved):

core/src/traits/Kokkos_WorkTagTrait.hpp
```diff
@@ -110,6 +110,9 @@ struct WorkTagTrait : TraitSpecificationBase<WorkTagTrait> {
   template <class T>
   using trait_matches_specification = std::integral_constant<
       bool, !std::is_void<T>::value &&
+#ifndef KOKKOS_ENABLE_DEPRECATED_CODE_3
+                std::is_empty<T>::value &&
```
nliber, line 114: When the deprecation is removed, we should get rid of the `!std::is_void<T>::value` check, correct?  Maybe leave a comment to that effect, or a `#else` on the macro?

nliber (changes requested):

core/src/traits/Kokkos_WorkTagTrait.hpp
dalg24: No, when the deprecation is removed we drop the macro guard that is all.
`void` is not a valid `WorkTag`

dalg24: Gotcha.  Will change.

masterleinad (changes requested):
We should also guard all of the changes in `core/src/impl/Kokkos_AnalyzePolicy.hpp` with `KOKKOS_ENABLE_DEPRECATED_CODE_3`. With the additional condition in `core/src/traits/Kokkos_WorkTagTrait.hpp` it won't really show up in error messages anyway. For me, the main error message then is
```
/tmp/kokkos/core/src/impl/Kokkos_AnalyzePolicy.hpp:172:7: error: implicit instantiation of undefined template 'Kokkos::Impl::show_name_of_invalid_execution_policy_trait<Kokkos::HostSpace>'
      show_name_of_invalid_execution_policy_trait<Trait>{};
      ^
```

nliber (approved):

nliber (approved):

dalg24:
> We should also guard all of the changes in `core/src/impl/Kokkos_AnalyzePolicy.hpp` with `KOKKOS_ENABLE_DEPRECATED_CODE_3`. With the additional condition in `core/src/traits/Kokkos_WorkTagTrait.hpp` it won't really show up in error messages anyway. For me, the main error message then is
> 
> ```
> /tmp/kokkos/core/src/impl/Kokkos_AnalyzePolicy.hpp:172:7: error: implicit instantiation of undefined template 'Kokkos::Impl::show_name_of_invalid_execution_policy_trait<Kokkos::HostSpace>'
>       show_name_of_invalid_execution_policy_trait<Trait>{};
>       ^
> ```

Done

masterleinad (approved):

ajpowelsnl:
What will replace the soon-to-be deprecated non-empty WorkTag class?

masterleinad:
> What will replace the soon-to-be deprecated non-empty WorkTag class?

There is no replacement. Users should just make sure that their tags are empty classes.

ajpowelsnl:
> > What will replace the soon-to-be deprecated non-empty WorkTag class?
> 
> There is no replacement. Users should just make sure that their tags are empty classes.

What was the motivation for this change?

PR #5039, opened 2022-05-19
Last few changes to get all Kokkos UnitTests to compile

crtrott:
This is enough to make all tests compiler with nvc++ as cuda compiler, assuming you use https://github.com/crtrott/desul/tree/refactor-for-nvhpc branch for external desul atomics. 

Not all tests pass yet but this is big step in the right direction.

PR #5232, opened 2022-07-18
Fixup `exp2(long double) -> long double` in NVC++ workaround

dalg24:
Fix https://github.com/kokkos/kokkos/pull/5203#issuecomment-1186732961
Introduced in #5039 `long double` are narrowed to `double`


core/src/Kokkos_MathematicalFunctions.hpp
```diff
@@ -318,6 +318,10 @@ KOKKOS_INLINE_FUNCTION double exp2(double val) {
   constexpr double ln2 = 0.69314718055994530941723212145818;
   return exp(ln2 * val);
 }
+inline long double exp2(double val) {
+  constexpr long double ln2 = 0.69314718055994530941723212145818;
```
masterleinad, line 322:
```suggestion
  constexpr long double ln2 = 0.69314718055994530941723212145818l;
```

dalg24: I updated and used the definition form the math constants.

masterleinad (approved):

core/src/Kokkos_MathematicalFunctions.hpp
```diff
@@ -311,16 +311,20 @@ KOKKOS_IMPL_MATH_UNARY_FUNCTION(exp)
...
-  constexpr float ln2 = 0.69314718055994530941723212145818;
+  constexpr float ln2 = 0.693147180559945309417232121458176568L;
   return exp(ln2 * val);
 }
 KOKKOS_INLINE_FUNCTION double exp2(double val) {
-  constexpr double ln2 = 0.69314718055994530941723212145818;
+  constexpr double ln2 = 0.693147180559945309417232121458176568L;
+  return exp(ln2 * val);
+}
+inline long double exp2(double val) {
```
crtrott, line 321: ??? how is this supposed to work, Should this be long double val?
Isn't this otherwise ambiguous?

crtrott (changes requested):

Issue #3385, opened 2020-09-14
Using MDRangePolicy in Hierarchical Parallelism 


kbthomp1:
It would be nice to be able to use MDRangePolicy within a TeamPolicy to exploit hierarchical parallel with loop tiling on each team's threads.  The specific use case is for structured grid problems where there are multiple grid blocks of different sizes, preventing a tightly nested 4D loop.  Being able to use an MDRangePolicy on each team allows each block to be assigned a team that then uses a 3D MDRangePolicy for the tightly nested loops on each structured grid block.

dalg24:
Tentative 3.5 release

crtrott:
Something like:

```
parallel_for(TeamPolicy<>(...), KOKKOS_LAMBDA(const team_t& team) {
  ....
  parallel_for(TeamVectorMDRange<Iterate::Left>(team, N0, N1, N2) , [=] (int i, int j, int k) {
     ....
  });
  ...
});
```
Essentially the equivalents to our TeamThreadRange, ThreadVectorRange, and TeamVectorRange just with MD. No tiling needed here.

kbthomp1:
@crtrott Yes exactly.  Something that provides MD versions of TeamThreadRange, ThreadVectorRange, and TeamVectorRange, but no explicit tiling needed.  Preferably, with the default template `Iterate` parameter switching from `Iterate::Right` to `Iterate::Left` depending on the default execution space, similar to how MDRangePolicy does.

crtrott:
Yeah we were just prioritizing work and put this on the list for the next release. I just added a bit more description to make clear what the plan is.

vbrunini:
Did this make it into 3.5? Guessing not?

masterleinad:
> Did this make it into 3.5? Guessing not?

No.

PhilMiller:
Delivered in the just-released 4.0.00

PR #5238, opened 2022-07-19
Team MD range policies impl

ldh4:
Motivation: #3385 

An implementation for MD Team policies. 
This adds parallel_for and parallel_reduce supports for MDTeamThreadRange, MDThreadVectorRange and MDTeamVectorRange.
Also adds in single directional supports for the policies.
This requires C++17

Worked with @nliber
Design provided by @crtrott


ldh4:
The starting ranks for threading and vectorization are currently noted in each backend's `ParallelRank` struct. Currently considering some options on how these should be defined to avoid needing to make each of these structs ifdef'ed per backend.

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -857,6 +858,364 @@ KOKKOS_INLINE_FUNCTION_DELETED Impl::ThreadVectorRangeBoundariesStruct<
...
+enum class MDTeamRangeParVector : bool { NotParVector, ParVector };
+enum class MDTeamRangeThreadAndVector : bool { NotBoth, Both };
+
+struct NoReductionTag {
+  explicit NoReductionTag() = default;
+};
+
+inline constexpr NoReductionTag no_reduction_tag{};
+
+template <MDTeamRangeLastRank LR, MDTeamRangeParThread PT,
```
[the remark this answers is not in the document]
crtrott: Is LastRank a good name? In the end its actually one past the last rank or?

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -857,6 +858,364 @@ KOKKOS_INLINE_FUNCTION_DELETED Impl::ThreadVectorRangeBoundariesStruct<
...
+KOKKOS_INLINE_FUNCTION void nested_loop(
+    MDTeamRangeStatus<MDTeamRangeLastRank::NotLastRank, ParThread,
+                      ParVector> const& status,
+    MDTeamRangeNextStatus<CurrentRank, TotalRank, ParThreadRank, ParVectorRank,
+                          Iter>,
+    Policy const& policy, Lambda const& lambda, ReducerValueType& val,
+    Args... args) {
+  constexpr int nextRank = CurrentRank + (Iter == Iterate::Right ? 1 : -1);
+  using NextStatus = MDTeamRangeNextStatus<nextRank, TotalRank, ParThreadRank,
+                                           ParVectorRank, Iter>;
```
crtrott, line 1012: In combination with my comment above maybe it would be better to pass the current rank in here, and do the offsetting inside?

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -857,6 +858,364 @@ KOKKOS_INLINE_FUNCTION_DELETED Impl::ThreadVectorRangeBoundariesStruct<
...
+          MDTeamRangeParVector PV>
+struct MDTeamRangeStatus {
+  static constexpr MDTeamRangeLastRank last_rank   = LR;
+  static constexpr MDTeamRangeParThread par_thread = PT;
+  static constexpr MDTeamRangeParVector par_vector = PV;
+};
+
+template <int CurrentRank, int TotalRank, int ParThreadRank, int ParVectorRank,
+          Iterate Iter>
+struct MDTeamRangeNextStatus {
```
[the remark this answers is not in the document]
crtrott: Also generally the name might be suboptimal, this struct doesn't only compute the next status, it also keeps track of the current rank etc.

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -857,6 +858,364 @@ KOKKOS_INLINE_FUNCTION_DELETED Impl::ThreadVectorRangeBoundariesStruct<
...
+inline constexpr NoReductionTag no_reduction_tag{};
+
+template <MDTeamRangeLastRank LR, MDTeamRangeParThread PT,
+          MDTeamRangeParVector PV>
+struct MDTeamRangeStatus {
+  static constexpr MDTeamRangeLastRank last_rank   = LR;
+  static constexpr MDTeamRangeParThread par_thread = PT;
+  static constexpr MDTeamRangeParVector par_vector = PV;
+};
+
```
crtrott, line 881: add comment explaining this class 
```
// Tag class to keep track of the loop nest level and where to deploy thread and
// vector parallelism
//   - CurrentRank is the current nest level
//   - TotalRank is the total number of loop nests
//   - ParThreadRank is the nesting level on which to deploy thread parallelism
//   - ParVectorRank is the nesting level on which to deploy vector parallelism
//   - Iter is whether to go forward or backward through ranks (i.e. the iteration order for MDRangePolicy)
```

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -857,6 +858,364 @@ KOKKOS_INLINE_FUNCTION_DELETED Impl::ThreadVectorRangeBoundariesStruct<
...
+                    t_and_v> {
+  // For threads deploy thread parallelism on Right most (i.e. slowed index)
+  static constexpr int par_rt  = TotalRank - 1;
+  static constexpr int par_rv  = 0;
+  static constexpr int invalid = -2;
+};
+template <int TotalRank, MDTeamRangeThreadAndVector t_and_v>
+struct ParallelRank<TotalRank, Iterate::Right, DefaultHostExecutionSpace,
+                    t_and_v> {
+  // For threads deploy thread parallelism on left most (i.e. slowed index)
```
crtrott, line 923: same as above

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -857,6 +858,364 @@ KOKKOS_INLINE_FUNCTION_DELETED Impl::ThreadVectorRangeBoundariesStruct<
...
+          Iterate Iter>
+using MDTeamRangeStatus_t =
+    typename MDTeamRangeNextStatus<CurrentRank, TotalRank, ParThreadRank,
+                                   ParVectorRank, Iter>::type;
+
+template <int totalRank, Iterate Iter, typename ExecSpace,
+          MDTeamRangeThreadAndVector t_and_v>
+struct ParallelRank;
+
+// FOR SERIAL ONLY
```
crtrott, line 911: ?? What does this mean?

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -857,6 +858,364 @@ KOKKOS_INLINE_FUNCTION_DELETED Impl::ThreadVectorRangeBoundariesStruct<
...
+  using NextStatus = MDTeamRangeNextStatus<nextRank, TotalRank, ParThreadRank,
+                                           ParVectorRank, Iter>;
+
+  // FIXME Check if recursively calling this is efficient enough compared to an
+  // iterative case
+
+  // This recursively processes ranks from [0..TotalRank-1]
+  // args... is passed by value because it should always be ints
+  parallel_for(
+      nested_policy(status, policy.team, policy.boundaries[CurrentRank]),
```
crtrott, line 1020: This can't possibly work right now or? I think you need another overload of nested_loop for the case where you have no parallelism.

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -857,6 +858,364 @@ KOKKOS_INLINE_FUNCTION_DELETED Impl::ThreadVectorRangeBoundariesStruct<
...
+// void parallel_for(Kokkos::TeamThreadRange const&, Closure const&);
+
+template <typename TeamHandle>
+KOKKOS_INLINE_FUNCTION auto nested_policy(
+    MDTeamRangeStatus<MDTeamRangeLastRank::NotLastRank,
+                      MDTeamRangeParThread::NotParThread,
+                      MDTeamRangeParVector::NotParVector>
+        status,
+    TeamHandle const& team, int count) {
+  return RangePolicy<>(0, count);
```
crtrott, line 939: This can't work right now. Eventually maybe yes if we have an "inline execution space" which is usable inside of parallel contexts, but right now this won't work.

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -857,6 +858,364 @@ KOKKOS_INLINE_FUNCTION_DELETED Impl::ThreadVectorRangeBoundariesStruct<
...
+  using md_team_iter =
+      Impl::MDTeamRangeNextStatus<startRank, TotalRank, par_rank_t::par_rt,
+                                  par_rank_t::invalid, Iter>;
+
+  nested_loop(rangeStatus, md_team_iter(), policy, lambda, val);
+}
+
+}  // namespace Impl
+
+template <Iterate Iter, int TotalRank, typename TeamHandle>
```
crtrott, line 1059: Maybe we should take `Kokkos::Rank<>` here instead of Iterate and TotalRank to make this have the same interface kind of like MDRangePolicy?

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -857,6 +858,364 @@ KOKKOS_INLINE_FUNCTION_DELETED Impl::ThreadVectorRangeBoundariesStruct<
...
+  team_handle_type const& team;
+  boundary_type boundaries[total_rank];
+};
+
+template <typename TeamHandle, typename... Args>
+MDTeamThreadRange(TeamHandle const&, Args&&...)
+    ->MDTeamThreadRange<Kokkos::Iterate::Default, sizeof...(Args), TeamHandle>;
+
+template <Iterate Iter, int TotalRank, typename TeamHandle>
+struct MDThreadVectorRange {
```
crtrott, line 1090: Same thing lets use Kokkos::Rank

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -857,6 +858,364 @@ KOKKOS_INLINE_FUNCTION_DELETED Impl::ThreadVectorRangeBoundariesStruct<
...
+  boundary_type boundaries[total_rank];
+};
+
+template <typename TeamHandle, typename... Args>
+MDThreadVectorRange(TeamHandle const&, Args&&...)
+    ->MDThreadVectorRange<Kokkos::Iterate::Default, sizeof...(Args),
+                          TeamHandle>;
+
+template <Iterate Iter, int TotalRank, typename TeamHandle>
+struct MDTeamVectorRange {
```
crtrott, line 1121: Same thing lets use Kokkos::Rank

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -857,6 +858,364 @@ KOKKOS_INLINE_FUNCTION_DELETED Impl::ThreadVectorRangeBoundariesStruct<
...
+      policy, lambda, rangeStatus, val);
+}
+
+// FIXME Check if passing in NoReductionTag by value vs reference shows any
+// difference
+template <Iterate Iter, int TotalRank, typename TeamHandle, typename Lambda>
+KOKKOS_INLINE_FUNCTION void parallel_for(
+    MDTeamThreadRange<Iter, TotalRank, TeamHandle> const& policy,
+    Lambda const& lambda) {
+  parallel_reduce(policy, lambda, Impl::no_reduction_tag);
```
crtrott, line 1170: Something in me doesn't like this and would prefer calling the impl_parallel function directly, but I guess its a question of code duplication?

crtrott (changes requested):
First pass of feedback

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -857,6 +858,364 @@ KOKKOS_INLINE_FUNCTION_DELETED Impl::ThreadVectorRangeBoundariesStruct<
...
+
+template <int totalRank, Iterate Iter, typename ExecSpace,
+          MDTeamRangeThreadAndVector t_and_v>
+struct ParallelRank;
+
+// FOR SERIAL ONLY
+template <int TotalRank, MDTeamRangeThreadAndVector t_and_v>
+struct ParallelRank<TotalRank, Iterate::Left, DefaultHostExecutionSpace,
+                    t_and_v> {
+  // For threads deploy thread parallelism on Right most (i.e. slowed index)
```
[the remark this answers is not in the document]
ldh4: Added in [ed4859d](https://github.com/kokkos/kokkos/pull/5238/commits/ed4859d102a0817c046a461751a5587dd5fc2a69)

core/src/Kokkos_ExecPolicy.hpp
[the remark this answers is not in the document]
ldh4: Comment added in [ed4859d](https://github.com/kokkos/kokkos/pull/5238/commits/ed4859d102a0817c046a461751a5587dd5fc2a69)

As for the name, I also agree that `LastRank` isn't exactly the right description of how its being used. I'll come up with something else.

core/src/Kokkos_ExecPolicy.hpp
ldh4: Added in [ed4859d](https://github.com/kokkos/kokkos/pull/5238/commits/ed4859d102a0817c046a461751a5587dd5fc2a69)

core/src/Kokkos_ExecPolicy.hpp
ldh4: Added in [ed4859d](https://github.com/kokkos/kokkos/pull/5238/commits/ed4859d102a0817c046a461751a5587dd5fc2a69)

core/src/Kokkos_ExecPolicy.hpp
ldh4: Removed in [ed4859d](https://github.com/kokkos/kokkos/pull/5238/commits/ed4859d102a0817c046a461751a5587dd5fc2a69) for now.

This was just a placeholder. Initially, we were thinking of adding in structs specialized for the other backends as well, but we'll do this part once the base implementation complete.
I was also exploring some options to define each `ParallelRank` struct for other backends in each file located in their respective backend directories, so that we can avoid keep adding in ifdef'ed `ParallelRank` struct all in here.

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -857,6 +858,364 @@ KOKKOS_INLINE_FUNCTION_DELETED Impl::ThreadVectorRangeBoundariesStruct<
...
+      static_cast<MDTeamRangeParThread>(current_rank == par_thread_rank),
+      static_cast<MDTeamRangeParVector>(current_rank == par_vector_rank)>;
+};
+
+template <int CurrentRank, int TotalRank, int ParThreadRank, int ParVectorRank,
+          Iterate Iter>
+using MDTeamRangeStatus_t =
+    typename MDTeamRangeNextStatus<CurrentRank, TotalRank, ParThreadRank,
+                                   ParVectorRank, Iter>::type;
+
```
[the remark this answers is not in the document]
ldh4: Added in [ed4859d](https://github.com/kokkos/kokkos/pull/5238/commits/ed4859d102a0817c046a461751a5587dd5fc2a69).

core/src/Kokkos_ExecPolicy.hpp
ldh4: One of the reasons that we moved away from `Kokkos::Rank`, even in our previous implementation, was because `Kokkos::Rank` is supporting only up to 6 ranks. I also do like the consistency that this change would have with `MDRangePolicy`. So I will change this to use `Kokkos::Rank`, if 6 ranks is enough.

https://github.com/kokkos/kokkos/blob/d19aab9981a2c447e832a7b4eb7b16992328fb14/core/src/Kokkos_Rank.hpp#L54-L60

core/src/Kokkos_ExecPolicy.hpp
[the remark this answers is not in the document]
ldh4: Modified in [4d802ec](https://github.com/kokkos/kokkos/pull/5238/commits/4d802ec0db4591eda75b5565b0042ef04763c875)

Changed that struct's name to `MDTeamRangeRanks`. We tried out both ways on where to do `CurrentRank+1`, and incrementing it internally added unnecessary complexity as we ultimately have to track current, next and next next status in one function. So, we are keeping the offset calculation outside of that struct.

core/src/Kokkos_ExecPolicy.hpp
ldh4: Handled in [4d802ec](https://github.com/kokkos/kokkos/pull/5238/commits/4d802ec0db4591eda75b5565b0042ef04763c875).
See above.

core/src/Kokkos_ExecPolicy.hpp
ldh4: Fixed in [4d802ec](https://github.com/kokkos/kokkos/pull/5238/commits/4d802ec0db4591eda75b5565b0042ef04763c875)

core/src/Kokkos_ExecPolicy.hpp
ldh4: Fixed in [4d802ec](https://github.com/kokkos/kokkos/pull/5238/commits/4d802ec0db4591eda75b5565b0042ef04763c875)

core/src/Kokkos_ExecPolicy.hpp
ldh4: Changed to use `Kokkos::Rank` in [c534937](https://github.com/kokkos/kokkos/pull/5238/commits/c5349378887bbbf53a29cd76eb21095cdd18b83d)

Will make a separate pr to handle `Kokkos::Rank` requiring ranks to be less than 7.

core/src/Kokkos_ExecPolicy.hpp
ldh4: Changed in [c534937](https://github.com/kokkos/kokkos/pull/5238/commits/c5349378887bbbf53a29cd76eb21095cdd18b83d)

core/src/Kokkos_ExecPolicy.hpp
ldh4: Changed in [c534937](https://github.com/kokkos/kokkos/pull/5238/commits/c5349378887bbbf53a29cd76eb21095cdd18b83d)

core/src/Kokkos_ExecPolicy.hpp
ldh4: Also changed in [c534937](https://github.com/kokkos/kokkos/pull/5238/commits/c5349378887bbbf53a29cd76eb21095cdd18b83d)

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -847,6 +847,301 @@ KOKKOS_INLINE_FUNCTION_DELETED Impl::TeamThreadRangeBoundariesStruct<
...
+//     - TotalRank is the total number of loop nests
+//     - Iter is whether to go forward or backward through ranks (i.e. the
+//       iteration order for MDRangePolicy)
+//   - t_and_v determines whether both vector and thread parallelism is in use
+template <typename Rank, typename ExecSpace, MDTeamRangeThreadAndVector t_and_v>
+struct ParallelRank;
+
+#ifdef KOKKOS_ENABLE_SERIAL
+template <typename Rank, MDTeamRangeThreadAndVector t_and_v>
+struct ParallelRank<Rank, DefaultHostExecutionSpace, t_and_v> {
```
crtrott, line 913:
```suggestion
struct ParallelRank<Rank, Serial, t_and_v> {
```

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -847,6 +847,301 @@ KOKKOS_INLINE_FUNCTION_DELETED Impl::TeamThreadRangeBoundariesStruct<
...
+};
+#endif
+
+// TODO check par_rt and par_rv validity
+#ifdef KOKKOS_ENABLE_SYCL
+struct ParallelRank<Rank, SYCL, t_and_v> {
+  static constexpr bool isDirectionLeft =
+      (Rank::outer_direction == Iterate::Left);
+  static constexpr int par_rt  = isDirectionLeft ? Rank::rank - 1 : 0;
+  static constexpr int par_rv  = isDirectionLeft ? 0 : Rank::rank - 1;
```
crtrott, line 994: @masterleinad should this be like CUDA? i.e. do we support separate Thread and Vector parallelism?

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -847,6 +847,301 @@ KOKKOS_INLINE_FUNCTION_DELETED Impl::TeamThreadRangeBoundariesStruct<
...
+#ifdef KOKKOS_ENABLE_CUDA
+template <typename Rank, MDTeamRangeThreadAndVector t_and_v>
+struct ParallelRank<Rank, Cuda, t_and_v> {
+  static constexpr bool isDirectionLeft =
+      (Rank::outer_direction == Iterate::Left);
+  // If vector parallelism is in use deploy thread parallelism on
+  // the second to the last rank, otherwise, thread parallelism on the last rank
+  static constexpr int par_rt =
+      isDirectionLeft ? ((t_and_v) ? 1 : 0)
+                      : ((t_and_v) ? Rank::rank - 2 : Rank::rank - 1);
```
crtrott, line 931: These guys need to check that Rank>1 in order to allow for the 1 and Rank::rank-2 otherwise it needs to be 0, and Rank::rank-1

crtrott: ah I guess we just support rank 2 anyway for now ... maybe a follow up

crtrott (changes requested):
Looks pretty good

core/src/Kokkos_ExecPolicy.hpp
ldh4: Changed in [7361243](https://github.com/kokkos/kokkos/pull/5238/commits/7361243c9f3e76c1328b1c1d27df3f25f0ec410a).

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -847,6 +847,301 @@ KOKKOS_INLINE_FUNCTION_DELETED Impl::TeamThreadRangeBoundariesStruct<
...
+//     - Iter is whether to go forward or backward through ranks (i.e. the
+//       iteration order for MDRangePolicy)
+//   - t_and_v determines whether both vector and thread parallelism is in use
+template <typename Rank, typename ExecSpace, MDTeamRangeThreadAndVector t_and_v>
+struct ParallelRank;
+
+#ifdef KOKKOS_ENABLE_SERIAL
+template <typename Rank, MDTeamRangeThreadAndVector t_and_v>
+struct ParallelRank<Rank, DefaultHostExecutionSpace, t_and_v> {
+  static constexpr bool isDirectionLeft =
```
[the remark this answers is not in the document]
ldh4: Changed in [7361243](https://github.com/kokkos/kokkos/pull/5238/commits/7361243c9f3e76c1328b1c1d27df3f25f0ec410a).

core/src/Kokkos_ExecPolicy.hpp
ldh4: Yep, because we are now using Kokkos::Rank, it will assert that Ranks are range [2, 8].

https://github.com/kokkos/kokkos/blob/4f3ebbe219dd690c3938c56879758e0d55cfb8e2/core/src/Kokkos_Rank.hpp#L59-L72

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -847,6 +847,301 @@ KOKKOS_INLINE_FUNCTION_DELETED Impl::TeamThreadRangeBoundariesStruct<
...
+  return TeamVectorRange(team, count);
+}
+
+// MDTeamRangeRanks is only needed to deduce template parameters
+template <typename Rank, int ParThreadRank, int ParVectorRank, int CurrentRank,
+          typename Policy, typename Lambda, typename... Args>
+KOKKOS_INLINE_FUNCTION void nested_loop(
+    MDTeamRangeMode<MDTeamRangeEndRank::EndRank,
+                    MDTeamRangeParThread::NotParThread,
+                    MDTeamRangeParVector::NotParVector> const&,
```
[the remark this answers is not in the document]
ldh4: Changed in [7361243](https://github.com/kokkos/kokkos/pull/5238/commits/7361243c9f3e76c1328b1c1d27df3f25f0ec410a).

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -847,6 +847,301 @@ KOKKOS_INLINE_FUNCTION_DELETED Impl::TeamThreadRangeBoundariesStruct<
...
+  constexpr auto Iter      = MDTeamPolicy::iter;
+  using TeamHandle         = typename MDTeamPolicy::team_handle_type;
+
+  static_assert(MDTeamPolicy::total_rank >= 2 && MDTeamPolicy::total_rank <= 8);
+
+  using par_rank_t =
+      Impl::ParallelRank<Rank, typename TeamHandle::execution_space,
+                         threadAndVector>;
+
+  constexpr int beginRank = (Iter == Iterate::Right) ? 0 : (TotalRank - 1);
```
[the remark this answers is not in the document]
ldh4: Changed all cases to PascalCases if types or template parameters and to snake_cases if variables in [7361243](https://github.com/kokkos/kokkos/pull/5238/commits/7361243c9f3e76c1328b1c1d27df3f25f0ec410a).

Do we have a common naming style that we try to follow? Since Kokkos is already strictly enforcing a formatting style,  attempting to have a similar naming style could also be nice for consistency.

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -866,6 +1161,181 @@ KOKKOS_INLINE_FUNCTION_DELETED Impl::ThreadVectorRangeBoundariesStruct<
...
+  static constexpr rank_type total_rank = Rank<N, OuterDir, InnerDir>::rank;
+  static constexpr Iterate iter         = OuterDir;
+
+  static constexpr Iterate direction =
+      iter == Iterate::Default
+          ? layout_iterate_type_selector<array_layout>::outer_iteration_pattern
+          : iter;
+
+  template <class... Args>
+  KOKKOS_INLINE_FUNCTION MDTeamThreadRange(team_handle_type const& team_,
```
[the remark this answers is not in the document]
ldh4: Changed in [7361243](https://github.com/kokkos/kokkos/pull/5238/commits/7361243c9f3e76c1328b1c1d27df3f25f0ec410a).

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -847,6 +847,301 @@ KOKKOS_INLINE_FUNCTION_DELETED Impl::TeamThreadRangeBoundariesStruct<
...
+
+#ifdef KOKKOS_ENABLE_HPX
+struct ParallelRank<Rank, Experimental::HPX, t_and_v> {
+  static constexpr bool isDirectionLeft =
+      (Rank::outer_direction == Iterate::Left);
+  static constexpr int par_rt  = isDirectionLeft ? Rank::rank - 1 : 0;
+  static constexpr int par_rv  = isDirectionLeft ? 0 : Rank::rank - 1;
+  static constexpr int invalid = -2;
+};
+#endif
```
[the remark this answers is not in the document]
ldh4: I agree that many of these can use a common strategy.
I'll consolidate them once their configurations are finalized.

core/src/Kokkos_ExecPolicy.hpp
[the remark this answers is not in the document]
ldh4: This has eventually been changed to `EndRank` to keep it similar to an iterator's `begin` and `end`.

ldh4:
Changes to the names in [7361243](https://github.com/kokkos/kokkos/pull/5238/commits/7361243c9f3e76c1328b1c1d27df3f25f0ec410a):
`MDTeamRange...`            -> `TeamMDRange...`
`MDTeamThreadRange`   -> `TeamMDThreadRange`
`MDThreadVectorRange` -> `ThreadMDVectorRange`
`MDTeamVectorRange`    -> `TeamMDVectorRange`

from an opinion in https://github.com/kokkos/kokkos-core-wiki/pull/137#issuecomment-1210622841

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -866,6 +1161,181 @@ KOKKOS_INLINE_FUNCTION_DELETED Impl::ThreadVectorRangeBoundariesStruct<
...
+
+template <typename Rank, typename TeamHandle, typename Lambda,
+          typename ReducerValueType>
+KOKKOS_INLINE_FUNCTION void parallel_reduce(
+    MDTeamThreadRange<Rank, TeamHandle> const& policy, Lambda const& lambda,
+    ReducerValueType& val) {
+  Impl::MDTeamRangeMode<Impl::MDTeamRangeEndRank::NotEndRank,
+                        Impl::MDTeamRangeParThread::ParThread,
+                        Impl::MDTeamRangeParVector::NotParVector>
+      rangeMode;
```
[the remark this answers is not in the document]
ldh4: I was looking into this, because it sounded like a plausible change to make at first.
But the "mode" is dictated by the policy only at its initial state. After that, it is updated based on the current rank that the recursion is operating on. So it is not set just based on the policy type. This is how "mode" is updated in each nested level:
```
  using RangeMode = TeamMDRangeMode<
      (iter == Iterate::Right)
          ? static_cast<TeamMDRangeEndRank>(current_rank == total_rank)
          : static_cast<TeamMDRangeEndRank>(current_rank == -1),
      static_cast<TeamMDRangeParThread>(current_rank == par_thread_rank),
      static_cast<TeamMDRangeParVector>(current_rank == par_vector_rank)>;
 ```
I did have another implementation that didn't use a tag class like this, but I don't believe that it was much better than the current implementation.

File: core/src/Kokkos_ExecPolicy.hpp [hunk 24]
```diff
@@ -847,6 +847,305 @@ KOKKOS_INLINE_FUNCTION_DELETED Impl::TeamThreadRangeBoundariesStruct<
...
+  // the second to the last rank, otherwise, thread parallelism on the last rank
+  static constexpr int par_rt =
+      is_direction_left ? ((ThreadAndVector) ? 1 : 0)
+                        : ((ThreadAndVector) ? Rank::rank - 2 : Rank::rank - 1);
+  // Vector parallelism will always be on the last index
+  static constexpr int par_rv  = is_direction_left ? 0 : Rank::rank - 1;
+  static constexpr int invalid = -2;
+};
+#endif
+
```
crtrott, line 939: O think it might be good to unify here in the categories we got. Essentially GPU vs CPU and do I actually got two nesting levels or not. So HIP and CUDA would do the same OpenMPTarget and SYCL for now do the same, and the CPU backends do the same.

crtrott (changes requested):
Some wording and maybe a bit code simplifications but otherwise I think this is basiclally ready

core/src/Kokkos_ExecPolicy.hpp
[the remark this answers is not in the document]
ldh4: Changed in [e400aa1](https://github.com/kokkos/kokkos/pull/5238/commits/e400aa1fc708dbc461c9f646041c10e82d30d2be). Strategies are temporarily named `HostBasedNestLevel` and `AcceleratorBasedNestLevel`, but I am looking for more suitable names for them.

core/src/Kokkos_ExecPolicy.hpp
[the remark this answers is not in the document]
ldh4: Handled in [54df360](https://github.com/kokkos/kokkos/pull/5238/commits/54df360c6ee32596442f3ad544cd7648e26a6e52). Removed these explicit modes from the interface. They are now initially deduced from the user input policy, but a form of a "mode" is still kept track of in impl.

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -847,6 +847,305 @@ KOKKOS_INLINE_FUNCTION_DELETED Impl::TeamThreadRangeBoundariesStruct<
...
+// vector parallelism
+//   - Rank is Kokkos::Rank<TotalRank, Iter>
+//     - total_rank is the total number of loop nests
+//     - iter is whether to go forward or backward through ranks (i.e. the
+//       iteration order for MDRangePolicy)
+//   - ParThreadRank is the nesting level on which to deploy thread parallelism
+//   - ParVectorRank is the nesting level on which to deploy vector parallelism
+//   - CurrentRank is the current nest level
+template <typename Rank, int ParThreadRank, int ParVectorRank, int CurrentRank>
+struct TeamMDRangeRanks {
```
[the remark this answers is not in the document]
ldh4: Changed in [5d999e8](https://github.com/kokkos/kokkos/pull/5238/commits/5d999e8df62d1cb1b03df179f9eb8e0bd9e67627).
`TeamMDRangeRanks` has been renamed to `TeamMDRangeNestingTracker`.
Other usages of the word `rank` that do not indicate `Kokkos::Rank` have been changed to `nest_level` accordingly as we discussed.

core/src/Kokkos_ExecPolicy.hpp
ldh4: Partially handled in [e400aa1](https://github.com/kokkos/kokkos/pull/5238/commits/e400aa1fc708dbc461c9f646041c10e82d30d2be).
Looking for more fitting names for these "strategy" structs.
I will add in another "strategy" struct for OpenMPTarget and SYCL as applicable.

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -847,6 +847,305 @@ KOKKOS_INLINE_FUNCTION_DELETED Impl::TeamThreadRangeBoundariesStruct<
...
+                        : ((ThreadAndVector) ? Rank::rank - 2 : Rank::rank - 1);
+  // Vector parallelism will always be on the last index
+  static constexpr int par_rv  = is_direction_left ? 0 : Rank::rank - 1;
+  static constexpr int invalid = -2;
+};
+#endif
+
+#ifdef KOKKOS_ENABLE_HIP
+template <typename Rank, TeamMDRangeThreadAndVector ThreadAndVector>
+struct ParallelRank<Rank, Experimental::HIP, ThreadAndVector> {
```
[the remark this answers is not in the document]
ldh4: Fixed in [ef4f479](https://github.com/kokkos/kokkos/pull/5238/commits/ef4f47956a91b0d64e461b3b5ed80265403614b0)

ldh4:
Retest this please

ldh4:
Retest this please

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -864,6 +864,178 @@ ThreadVectorRange(const TeamMemberType&, const iType1& arg_begin,
 
 namespace Impl {
 
+enum class TeamMDRangeLastNestLevel : bool { NotLastNestLevel, LastNestLevel };
+enum class TeamMDRangeParThread : bool { NotParThread, ParThread };
+enum class TeamMDRangeParVector : bool { NotParVector, ParVector };
+enum class TeamMDRangeThreadAndVector : bool { NotBoth, Both };
+
+struct NoReductionTag {
+  explicit NoReductionTag() = default;
```
masterleinad, line 873: What's the point in adding a default constructor (and why make it `expiicit`)?

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -864,6 +864,178 @@ ThreadVectorRange(const TeamMemberType&, const iType1& arg_begin,
...
+};
+
+static KOKKOS_IMPL_DEVICE_FUNCTION NoReductionTag no_reduction_tag{};
```
masterleinad, line 876: It doesn't make sense to me to use `KOKKOS_IMPL_DEVICE_FUNCTION` here just because it does the right thing.
`SYCL`'s support for global device variables is pretty immature and so I would be happy if we could avoid doing this in the first place. If we must, the attribute should have a different name.

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -864,6 +864,178 @@ ThreadVectorRange(const TeamMemberType&, const iType1& arg_begin,
...
+KOKKOS_INLINE_FUNCTION void parallel_reduce(
+    TeamThreadMDRange<Rank, TeamHandle> const& policy, Lambda const& lambda,
+    ReducerValueType& val) {
+  Impl::md_parallel_impl<Rank>(policy, lambda, val);
+}
+
+template <typename Rank, typename TeamHandle, typename Lambda>
+KOKKOS_INLINE_FUNCTION void parallel_for(
+    TeamThreadMDRange<Rank, TeamHandle> const& policy, Lambda const& lambda) {
+  Impl::md_parallel_impl<Rank>(policy, lambda, Impl::no_reduction_tag);
```
masterleinad, line 1006:
```suggestion
  Impl::md_parallel_impl<Rank>(policy, lambda, Impl::NoReductionTag{});
```
here and elsewhere.

crtrott: agreed lets get rid of the global static variable and just create it in-place.

crtrott (changes requested):
Other than that change requested from Daniel too I think this is now good. Definitely like the renamed classes much better, and also the split into separate file. I think this is much more readable now.

core/src/impl/Kokkos_TeamMDPolicy.hpp
```diff
@@ -0,0 +1,374 @@
...
+//   in use
+template <typename Rank, typename ExecSpace,
+          TeamMDRangeThreadAndVector ThreadAndVector>
+struct ThreadAndVectorNestLevel;
+
+#ifdef KOKKOS_ENABLE_SERIAL
+template <typename Rank, TeamMDRangeThreadAndVector ThreadAndVector>
+struct ThreadAndVectorNestLevel<Rank, Serial, ThreadAndVector>
+    : HostBasedNestLevel<Rank, ThreadAndVector> {};
+#endif
```
masterleinad, line 158: This should be in backend-specific files, probably together with other `MDRange` implementation files.

core/src/impl/Kokkos_TeamMDPolicy.hpp
```diff
@@ -0,0 +1,374 @@
...
+struct ThreadAndVectorNestLevel<Rank, Threads, ThreadAndVector>
+    : HostBasedNestLevel<Rank, ThreadAndVector> {};
+#endif
+
+// TODO check par_rt and par_rv validity
+#ifdef KOKKOS_ENABLE_SYCL
+template <typename Rank, TeamMDRangeThreadAndVector ThreadAndVector>
+struct ThreadAndVectorNestLevel<Rank, Experimental::SYCL, ThreadAndVector>
+    : HostBasedNestLevel<Rank, ThreadAndVector> {};
+#endif
```
masterleinad, line 197: Is there any reason this is different from` CUDA1/HIP` or have you just not tested?

masterleinad (commented):
Looks mostly good to me. Just a few comments.

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -847,6 +847,305 @@ KOKKOS_INLINE_FUNCTION_DELETED Impl::TeamThreadRangeBoundariesStruct<
...
+namespace Impl {
+
+enum class TeamMDRangeEndRank : bool { NotEndRank, EndRank };
+enum class TeamMDRangeParThread : bool { NotParThread, ParThread };
+enum class TeamMDRangeParVector : bool { NotParVector, ParVector };
+enum class TeamMDRangeThreadAndVector : bool { NotBoth, Both };
+
+struct NoReductionTag {
+  explicit NoReductionTag() = default;
+};
```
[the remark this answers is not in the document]
ldh4: Changed in [f5f24e3](https://github.com/kokkos/kokkos/pull/5238/commits/f5f24e33bb5c604fb0cbc471850d201f86ca308c).

core/src/Kokkos_ExecPolicy.hpp
ldh4: Changed in [f5f24e3](https://github.com/kokkos/kokkos/pull/5238/commits/f5f24e33bb5c604fb0cbc471850d201f86ca308c).

Removed this explicit constructor.

core/src/Kokkos_ExecPolicy.hpp
ldh4: Changed in [f5f24e3](https://github.com/kokkos/kokkos/pull/5238/commits/f5f24e33bb5c604fb0cbc471850d201f86ca308c).

Removed this global device variable. Now constructs it inline.

core/src/Kokkos_ExecPolicy.hpp
ldh4: Changed in [f5f24e3](https://github.com/kokkos/kokkos/pull/5238/commits/f5f24e33bb5c604fb0cbc471850d201f86ca308c).

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -864,6 +864,178 @@ ThreadVectorRange(const TeamMemberType&, const iType1& arg_begin,
...
 namespace Impl {
 
+enum class TeamMDRangeLastNestLevel : bool { NotLastNestLevel, LastNestLevel };
+enum class TeamMDRangeParThread : bool { NotParThread, ParThread };
+enum class TeamMDRangeParVector : bool { NotParVector, ParVector };
+enum class TeamMDRangeThreadAndVector : bool { NotBoth, Both };
+
+struct NoReductionTag {
+  explicit NoReductionTag() = default;
+};
```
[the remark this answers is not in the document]
ldh4: Removed in [f5f24e3](https://github.com/kokkos/kokkos/pull/5238/commits/f5f24e33bb5c604fb0cbc471850d201f86ca308c).

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -864,6 +864,178 @@ ThreadVectorRange(const TeamMemberType&, const iType1& arg_begin,
...
+  using ArrayLayout    = typename ExecutionSpace::array_layout;
+
+  static constexpr NestLevelType total_nest_level =
+      Rank<N, OuterDir, InnerDir>::rank;
+  static constexpr Iterate iter    = OuterDir;
+  static constexpr auto par_thread = Impl::TeamMDRangeParThread::NotParThread;
+  static constexpr auto par_vector = Impl::TeamMDRangeParVector::ParVector;
+
+  static constexpr Iterate direction =
+      iter == Iterate::Default
```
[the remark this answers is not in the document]
ldh4: Fixed in [b5ca8f6](https://github.com/kokkos/kokkos/pull/5238/commits/b5ca8f61c91e45b64910010183c2d04d51880477).

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -864,6 +864,178 @@ ThreadVectorRange(const TeamMemberType&, const iType1& arg_begin,
...
+  using ArrayLayout    = typename ExecutionSpace::array_layout;
+
+  static constexpr NestLevelType total_nest_level =
+      Rank<N, OuterDir, InnerDir>::rank;
+  static constexpr Iterate iter    = OuterDir;
+  static constexpr auto par_thread = Impl::TeamMDRangeParThread::ParThread;
+  static constexpr auto par_vector = Impl::TeamMDRangeParVector::NotParVector;
+
+  static constexpr Iterate direction =
+      iter == Iterate::Default
```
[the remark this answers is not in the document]
ldh4: Fixed in [b5ca8f6](https://github.com/kokkos/kokkos/pull/5238/commits/b5ca8f61c91e45b64910010183c2d04d51880477).

core/src/impl/Kokkos_TeamMDPolicy.hpp
ldh4: Refactored in [ef6d137](https://github.com/kokkos/kokkos/pull/5238/commits/ef6d137d66aecc586bfe7b9ba673714a8978d40b).

Moved these structs to each backend's `Kokkos_[backend]_MDRangePolicy.hpp` file.

core/src/impl/Kokkos_TeamMDPolicy.hpp
ldh4: It just has not yet been tested for the `SYCL` backend. I changed it to be using the same config as `Cuda/HIP` in [ef6d137](https://github.com/kokkos/kokkos/pull/5238/commits/ef6d137d66aecc586bfe7b9ba673714a8978d40b).
Let me know if you think if this change is ok or if `SYCL` would benefit from a different configuration.

masterleinad: If we are talking about multi-dimensional access, i.e., sycl::nd_range and sycl::nd_item, the order of access should be reversed to what we do for `CUDA`, compare
```
dim3 grid(int(m_league_size), 1, 1);
const dim3 block(int(m_vector_size), int(m_team_size), 1);
```
for `Cuda` with
```
sycl::nd_range<2>(
            sycl::range<2>(m_team_size, m_league_size * final_vector_size),
            sycl::range<2>(m_team_size, final_vector_size)),
```
for `SYCL`. Since the functionality added here is already nested inside a `TeamPolicy`, I would expect that doing the same as for `Cuda` (or `HIP` should be correct). It might be best to measure, though.

masterleinad: It's fine to look at that later, though, if we do, we should create an issue to keep track of it.

core/unit_test/CMakeLists.txt
```diff
@@ -166,6 +166,7 @@ foreach(Tag Threads;Serial;OpenMP;Cuda;HPX;OpenMPTarget;OpenACC;HIP;SYCL)
         Reductions_DeviceView
         Scan
         SharedAlloc
+        TeamMDRange
```
crtrott, line 169: OpenACC doesn't have TeamPolicy yet so lets include it in the remove list for OpenACC (i.e. you leave that change here, to add it to all backends, and then tell OpenACC never mind you are gonna ignore this one).

crtrott (changes requested):
It looks like the SYCL test failure is real. It seems to hang in the test for this. @masterleinad can you take a look at this?

core/unit_test/CMakeLists.txt
ldh4: Added `TeamMDRange` test to the OpenACC unit test remove list. [3ae7bd0](https://github.com/kokkos/kokkos/pull/5238/commits/3ae7bd0ede3bbcb8946cac2efe7bef5dab20d5d9)

Once #5451 gets merged in, I'll remove this from the list.

masterleinad:
I pushed a commit disabling relevant tests for SYC+Cuda.

ldh4:
Retest this please

crtrott (approved):

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -864,6 +864,192 @@ ThreadVectorRange(const TeamMemberType&, const iType1& arg_begin,
...
+struct AcceleratorBasedNestLevel;
+
+// ThreadAndVectorNestLevel determines on which nested level parallelization
+// happens.
+//   - Rank is Kokkos::Rank<TotalNestLevel, Iter>
+//     - TotalNestLevel is the total number of loop nests
+//     - Iter is whether to go forward or backward through ranks (i.e. the
+//       iteration order for MDRangePolicy)
+//   - ThreadAndVector determines whether both vector and thread parallelism is
+//   in use
```
masterleinad, line 885:
```suggestion
//   - Rank is Kokkos::Rank<TotalNestLevel, Iter>
//     - TotalNestLevel is the total number of loop nests
//     - Iter is whether to go forward or backward through ranks (i.e. the
//       iteration order for MDRangePolicy)
//   - ThreadAndVector determines whether both vector and thread parallelism is
//     in use
```

core/src/Kokkos_HPX.hpp
```diff
@@ -977,6 +977,7 @@ class TeamPolicyInternal<Kokkos::Experimental::HPX, Properties...>
     return *this;
   }
 };
+
```
masterleinad, line 980:
```suggestion
```

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Instance.hpp
```diff
@@ -88,6 +88,7 @@ class OpenMPTargetInternal {
   uint32_t m_instance_id = Kokkos::Tools::Experimental::Impl::idForInstance<
       Kokkos::Experimental::OpenMPTarget>(reinterpret_cast<uintptr_t>(this));
 };
+
```
masterleinad, line 91:
```suggestion
```

core/src/SYCL/Kokkos_SYCL_Instance.hpp
```diff
@@ -341,6 +341,7 @@ template <typename Functor, typename Storage>
 auto make_sycl_function_wrapper(const Functor& functor, Storage& storage) {
   return SYCLFunctionWrapper<Functor, Storage>(functor, storage);
 }
+
```
masterleinad, line 344:
```suggestion
```

core/src/impl/Kokkos_TeamMDPolicy.hpp
```diff
@@ -0,0 +1,306 @@
...
+                                                        par_vector_nest_level)>;
+};
+
+// Structs to determine on which nested level parallelization happens.
+//   - Rank is Kokkos::Rank<TotalNestLevel, Iter>
+//     - TotalNestLevel is the total number of loop nests
+//     - Iter is whether to go forward or backward through ranks (i.e. the
+//       iteration order for MDRangePolicy)
+//   - ThreadAndVector determines whether both vector and thread parallelism is
+//   in use
```
masterleinad, line 110:
```suggestion
// Structs to determine on which nested level parallelization happens.
//   - Rank is Kokkos::Rank<TotalNestLevel, Iter>
//     - TotalNestLevel is the total number of loop nests
//     - Iter is whether to go forward or backward through ranks (i.e. the
//       iteration order for MDRangePolicy)
//   - ThreadAndVector determines whether both vector and thread parallelism is
//     in use
```

masterleinad (commented):
I found some formatting suggestions. Did `clang-format` introduce all these blank lines?

crtrott:
I applied them all.

masterleinad (approved):

crtrott:
This is ready for merge the last commit just changed formatting, and the commit before passed testing.

PR #5239, opened 2022-07-19
Revert common view alloc prop

crtrott:
Comment in code to explain this:
```
// This function is required for example for dynamic length embedded ensemble
// types It is used to propagate necessary information (like the ensemble size)
// Most of the time it is called with a single view however.
// Furthermore, the propagated information is not just for view allocations.
// From what I can tell this type of functionality is the equivalent to
// propagating accessors in mdspan, a mechanism we will use to replace this
// clunky approach here when we are finally mdspan based.
// TODO: get rid of this when we have mdspan
```

### DIFF
--- a/core/src/Kokkos_View.hpp
+++ b/core/src/Kokkos_View.hpp
@@ -1961,7 +1961,17 @@ template <class... Views>
 using DeducedCommonPropsType =
     typename Impl::DeduceCommonViewAllocProp<Views...>::prop_type;
 
-// User function
+// This function is required in certain scenarios where users customize
+// Kokkos View internals. One example are dynamic length embedded ensemble
+// types. The function is used to propagate necessary information
+// (like the ensemble size) when creating new views.
+// However, most of the time it is called with a single view.
+// Furthermore, the propagated information is not just for view allocations.
+// From what I can tell, the type of functionality provided by
+// common_view_alloc_prop is the equivalent of propagating accessors in mdspan,
+// a mechanism we will eventually use to replace this clunky approach here, when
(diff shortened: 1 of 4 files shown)


core/src/Kokkos_View.hpp
```diff
@@ -1963,14 +1962,23 @@ using DeducedCommonPropsType =
     typename Impl::DeduceCommonViewAllocProp<Views...>::prop_type;
 
 // User function
+// This function is required for example for dynamic length embedded ensemble
+// types It is used to propagate necessary information (like the ensemble size)
+// Most of the time it is called with a single view however.
+// Furthermore, the propagated information is not just for view allocations.
+// From what I can tell this type of functionality is the equivalent to
```
dalg24, line 1969: I think it is "the equivalent of" or "equivalent to"

dalg24 (approved):

core/src/Kokkos_View.hpp
crtrott: ah I saw you approved anyway, so I am not gonna change that :-D

ndellingwood (approved):

ndellingwood:
Does the appveyor build just need a retest?:
```
Build started
git clone -q --branch=develop https://github.com/kokkos/kokkos.git c:\projects\source
git checkout -qf 2a1ba95f9c83e37404781d4b0dfdb6e4ce5bac2d
fatal: reference is not a tree: 2a1ba95f9c83e37404781d4b0dfdb6e4ce5bac2d
Command exited with code 128
```

PR #5244, opened 2022-07-20
avoid using namespace to improve compatibility

XzzX:
I am working on a project with negligent use of `using namespace` and, therefore, got an ambiguous error for `_1`.
While technically this is not Kokkos's fault, this change might improve overall compatibility.

### DIFF
--- a/core/src/HIP/Kokkos_HIP_BlockSize_Deduction.hpp
+++ b/core/src/HIP/Kokkos_HIP_BlockSize_Deduction.hpp
@@ -249,12 +249,11 @@ unsigned hip_get_preferred_team_blocksize(HIPInternal const *hip_instance,
       get_hip_func_attributes_impl<DriverType, LaunchBounds,
                                    BlockType::Preferred>();
   // get preferred blocksize limited by register usage
-  using namespace std::placeholders;
   const unsigned tperb_reg =
       hip_get_preferred_blocksize<DriverType, LaunchBounds>();
   return hip_internal_get_block_size<BlockType::Preferred, DriverType,
                                      LaunchBounds>(
-      hip_instance, std::bind(f, attr, _1), tperb_reg);
+      hip_instance, std::bind(f, attr, std::placeholders::_1), tperb_reg);
 }
 
 // Standardized blocksize deduction for non-teams parallel constructs with LDS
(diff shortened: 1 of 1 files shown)


dalg24:
Can one of the admins verify this patch?

dalg24 (approved):
Fine by me.
Out of curiosity what was the error you saw downstream?

XzzX:
In fact the problem is not in my code but in boost.
https://github.com/boostorg/bind/blob/develop/include/boost/bind.hpp#L52
Unfortunately, not all boost libraries themselves follow their own very wise suggestion:
```
BOOST_PRAGMA_MESSAGE(
  "The practice of declaring the Bind placeholders (_1, _2, ...) "
  "in the global namespace is deprecated. Please use "
  "<boost/bind/bind.hpp> + using namespace boost::placeholders, "
  "or define BOOST_BIND_GLOBAL_PLACEHOLDERS to retain the current behavior."
)
```

TL;DR
If you happen to include `boost/bind.hpp` and use Kokkos with HIP backend you will get.
```
.../kokkos-src/core/src/HIP/Kokkos_HIP_BlockSize_Deduction.hpp:297:40: error: reference to '_1' is ambiguous
      hip_instance, std::bind(f, attr, _1), tperb_reg);
                                       ^
.../include/c++/9/functional:211:34: note: candidate found by name lookup is 'std::placeholders::_1'
    extern const _Placeholder<1> _1;
                                 ^
.../boost/bind/placeholders.hpp:46:38: note: candidate found by name lookup is 'boost::placeholders::_1'
BOOST_STATIC_CONSTEXPR boost::arg<1> _1;
```

masterleinad:
This is the only place in `Kokkos` that uses `std::bind` and my preference would go towards replacing `std::bind` with a lambda.

dalg24:
> This is the only place in `Kokkos` that uses `std::bind` and my preference would go towards replacing `std::bind` with a lambda.

This is up to @Rombur ultimately.  I have no issue with the resolution proposed here.

PhilMiller (approved):
I'd prefer getting rid of bind usage too, but this addresses the present compatibility issue, and Bruno may not want to deal with the added churn this close to the release.

Issue #5208, opened 2022-07-12
Consider changing the behavior of `Kokkos::is_initialized` and providing `Kokkos::is_finalized`?


dalg24:
Currently `Kokkos::is_initialized` returns
* `false` before calling `Kokkos::initialize` or after calling `Kokkos::finalize`, and
* `true` in-between.

This is unfortunate since it is not legal to call `Kokkos::initialize` more than once, and it means there is no way for programs to figure out if they can safely do so.

We should consider
* changing `Kokkos::is_initialized` so that it returns `false` before calling `Kokkos::initialize`, and `true` after,
* adding `Kokkos::is_finalized` that would return `false` before calling `Kokkos::finalize`, and `true` after.

This would be a breaking change so we need to think carefully about it.

```C++
#include <Kokkos_Core.hpp>

int main(int argc, char* argv[]) {
  assert(!Kokkos::is_initialized());
  assert(!Kokkos::is_finalized());

  Kokkos::initialize(argc, argv);

  assert(Kokkos::is_initialized());
  assert(!Kokkos::is_finalized());

  Kokkos::finalize(argc, argv);

  assert(Kokkos::is_initialized());  // <- breaking change
  assert(Kokkos::is_finalized());
```

masterleinad:
I would be in favor of at least adding `Kokkos::is_finalized`.

JBludau:
having a way to query if finalize was called would probably help with the python issue of not being allowed to have libs in the same context that use an initialized kokkos. Still, this would require the python libs to add the `is_finalized` to their interface, and the users to be aware of the issue ...

stanmoore1:
Also related: https://github.com/kokkos/kokkos/issues/4126

stanmoore1:
I support this change--I had to add logic in LAMMPS to work around the lack of `Kokkos::is_finalized()`, see https://github.com/lammps/lammps/blob/develop/src/KOKKOS/kokkos.cpp#L337-L353, would be nice to simplify the user code.

masterleinad:
I think we agree on introducing the functionality of `Kokkos::finalize`. We are most debating how `Kokkos::initialize` should behave.

dalg24:
I am strongly in favor of Option 2 with naming the new facility `Kokkos::initialized` and `Kokkos::finalized` and marking `Kokkos::is_initialized` as deprecated.  I think it is easy to document and least surprising to users without adding any cognitive load (no new symbols, consistency with MPI, etc.)

[...]

PR #5245, opened 2022-07-20
Deprecate `is_initialized` in favor of `initialized` and `finalized`

dalg24:
*My* proposed resolution for #5208

### DIFF
--- a/core/src/impl/Kokkos_Core.cpp
+++ b/core/src/impl/Kokkos_Core.cpp
@@ -73,7 +73,8 @@
 
 //----------------------------------------------------------------------------
 namespace {
-bool g_is_initialized = false;
+bool g_initialized    = false;
+bool g_finalized      = false;
 bool g_show_warnings  = true;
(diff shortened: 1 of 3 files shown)


core/src/impl/Kokkos_Core.cpp
```diff
@@ -666,7 +667,7 @@ void finalize_internal() {
 
   Kokkos::Impl::ExecSpaceManager::get_instance().finalize_spaces();
 
-  g_is_initialized = false;
+  g_finalized      = false;
```
PhilMiller, line 670: Should be true here?

crtrott (approved):
Please document the deprecation.

nliber (approved):

crtrott (changes requested):
Ok putting a hold on it lets talk about the options today, I thought I missed that the team had agreed on something, and was just a bit done with arguing :-)

dalg24:
Dev team decided to go for option 1

PR #5246, opened 2022-07-20
SYCL: Store device_id passed from initialization

masterleinad:
To the best of my knowledge, there is no other way to obtain unique information about the `sycl::device` used, thus just storing what was given as `device_id` during initialization seems the best we can do. It's debatable what to do if the user passes just a `sycl::queue` add the moment we just use ~~0 by default but we could also use~~ the `device_id` the singleton was initialized with.

### DIFF
--- a/core/src/Kokkos_SYCL.hpp
+++ b/core/src/Kokkos_SYCL.hpp
@@ -134,8 +134,6 @@ class SYCL {
 
   static void impl_initialize(InitializationSettings const&);
 
-  int sycl_device() const;
-
   static bool impl_is_initialized();
 
   static int concurrency();
@@ -161,7 +159,7 @@ struct DeviceTypeTraits<Kokkos::Experimental::SYCL> {
   /// \brief An ID to differentiate (for example) Serial from OpenMP in Tooling
   static constexpr DeviceType id = DeviceType::SYCL;
   static int device_id(const Kokkos::Experimental::SYCL& exec) {
-    return exec.sycl_device();
+    return exec.impl_internal_space_instance()->m_syclDev;
   }
 };
 }  // namespace Experimental
--- a/core/src/SYCL/Kokkos_SYCL.cpp
+++ b/core/src/SYCL/Kokkos_SYCL.cpp
@@ -142,10 +142,6 @@ void SYCL::impl_static_fence(const std::string& name) {
       });
 }
 
-int SYCL::sycl_device() const {
-  return impl_internal_space_instance()->m_syclDev;
-}
-
 void SYCL::impl_initialize(InitializationSettings const& settings) {
   std::vector<sycl::device> gpu_devices =
       sycl::device::get_devices(sycl::info::device_type::gpu);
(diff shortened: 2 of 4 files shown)


core/src/Kokkos_SYCL.hpp
```diff
@@ -134,7 +134,7 @@ class SYCL {
 
   static void impl_initialize(InitializationSettings const&);
 
-  int sycl_device() const;
+  static int sycl_device();
```
dalg24, line 137: Absolutely cannot be called `sycl_device` as it is misleading users.
One could expect it returns `sycl::device` just like `sycl_queue() -> sycl::queue`.
I don't think we need to expose any public member in the execution space class for it.
```suggestion
```
Instead do
```C++
template <>
struct DeviceTypeTraits<Kokkos::Experimental::SYCL> {
  static constexpr DeviceType id = DeviceType::SYCL;
  static int device_id(const Kokkos::Experimental::SYCL& exec) {
    return exec.impl_internal_space_instance().m_syclDev;
  }
};
```

core/src/SYCL/Kokkos_SYCL.cpp
```diff
@@ -157,12 +155,12 @@ void SYCL::impl_initialize(InitializationSettings const& settings) {
...
   if (!settings.has_device_id() && gpu_devices.empty()) {
-    Impl::SYCLInternal::singleton().initialize(sycl::device());
+    Impl::SYCLInternal::singleton().initialize(sycl::device(), 0);
     return;
   }
 #endif
-  using Kokkos::Impl::get_gpu;
-  Impl::SYCLInternal::singleton().initialize(gpu_devices[get_gpu(settings)]);
+  const auto id = ::Kokkos::Impl::get_gpu(settings);
+  Impl::SYCLInternal::singleton().initialize(gpu_devices[id], id);
```
dalg24, line 163: No need to propagate this to `SYCLInternal::initialize`.  Do
```suggestion
  Impl::SYCLInternal::singleton().initialize(gpu_devices[id]);
  Impl::SYCLInternal::m_syclDev = id;
```

dalg24 (changes requested):

masterleinad:
@dalg24 I followed your suggestions.

dalg24 (approved):

masterleinad:
For some reason, `OpenACC` can't find a device in one test
```
Test command: /var/jenkins/workspace/Kokkos/build/core/unit_test/KokkosCore_IncrementalTest_OPENACC
33: Test timeout computed to be: 1500
33: Error: no GPU available for execution.
33:  Raised by Kokkos::initialize().
33: 
33: Backtrace:
33:                         [0x4b2453]
33:                         [0x490463]
33:                         [0x485cf3]
33:                         [0x485e32]
33:                         [0x4b68d1]
33:                         [0x48e81c]
33:                         [0x48a048]
33:                         [0x412fdd]
33: __libc_start_main [0x7f29a846e0b3]
33:                         [0x412e6e]
33/48 Test #33: KokkosCore_IncrementalTest_OPENACC ...........Subprocess aborted***Exception:   0.19 sec
```
That's clearly unrelated to this pull request that only touches `SYCL` fails and I have seen this failure in other pull requests, too.

JBludau (approved):
looks good

dalg24:
Failure was unrelated (OpenACC issue about GPU architecture being wrong that is been resolved on `develop`)

PR #5247, opened 2022-07-20
Add is_finalized

### LINKED ISSUE #5208: Consider changing the behavior of `Kokkos::is_initialized` and providing `Kokkos::is_finalized`?

crtrott:
Fixes #5208.

is_finalized is also checked in the ScopeGuard constructor

Co-Authored-By: Damien L-G <dalg24@gmail.com>

### DIFF
--- a/core/src/Kokkos_Core.hpp
+++ b/core/src/Kokkos_Core.hpp
@@ -204,11 +204,6 @@ class KOKKOS_ATTRIBUTE_NODISCARD ScopeGuard {
   ScopeGuard(int& argc, char* argv[]) {
     sg_init = false;
     if (!is_initialized()) {
-      if (is_finalized()) {
-        Kokkos::abort(
-            "Attempt to create a Kokkos::ScopeGuard after Kokkos::finalize was "
-            "called.");
-      }
       initialize(argc, argv);
       sg_init = true;
     }
@@ -221,18 +216,13 @@ class KOKKOS_ATTRIBUTE_NODISCARD ScopeGuard {
       const InitializationSettings& settings = InitializationSettings()) {
     sg_init = false;
     if (!is_initialized()) {
-      if (is_finalized()) {
-        Kokkos::abort(
-            "Attempt to create a Kokkos::ScopeGuard after Kokkos::finalize was "
-            "called.");
-      }
       initialize(settings);
       sg_init = true;
     }
   }
 
   ~ScopeGuard() {
-    if (!is_finalized() && sg_init) {
+    if (is_initialized() && sg_init) {
       finalize();
     }
   }
--- a/core/src/Kokkos_Core.hpp
+++ b/core/src/Kokkos_Core.hpp
@@ -102,7 +102,8 @@ void declare_configuration_metadata(const std::string& category,
 
 }  // namespace Impl
 
-bool is_initialized() noexcept;
+KOKKOS_ATTRIBUTE_NODISCARD bool is_initialized() noexcept;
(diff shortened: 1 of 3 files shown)


core/src/Kokkos_Core.hpp
```diff
@@ -202,7 +203,12 @@ class KOKKOS_ATTRIBUTE_NODISCARD ScopeGuard {
 #endif
   ScopeGuard(int& argc, char* argv[]) {
     sg_init = false;
-    if (!Kokkos::is_initialized()) {
+    if (!is_initialized()) {
+      if (is_finalized()) {
+        Kokkos::abort(
+            "Attempt to create a Kokkos::ScopeGuard after Kokkos::finalize was "
+            "called.");
+      }
```
dalg24, line 211:
```suggestion
    if (is_finalized()) {
      Kokkos::abort(
          "Attempt to create a Kokkos::ScopeGuard after Kokkos::finalize was "
          "called.");
    }
    if (!is_initialized()) {
```

core/src/Kokkos_Core.hpp
```diff
@@ -214,14 +220,19 @@ class KOKKOS_ATTRIBUTE_NODISCARD ScopeGuard {
   ScopeGuard(
       const InitializationSettings& settings = InitializationSettings()) {
     sg_init = false;
-    if (!Kokkos::is_initialized()) {
+    if (!is_initialized()) {
+      if (is_finalized()) {
+        Kokkos::abort(
+            "Attempt to create a Kokkos::ScopeGuard after Kokkos::finalize was "
+            "called.");
+      }
```
dalg24, line 228:
```suggestion
    if (is_finalized()) {
      Kokkos::abort(
          "Attempt to create a Kokkos::ScopeGuard after Kokkos::finalize was "
          "called.");
    }
    if (!is_initialized()) {
```

core/src/Kokkos_Core.hpp
crtrott: it is one more check in the case of `is_initialized()==true` but I guess thats not that much cost?

dalg24 (approved):

dalg24:
Cuda timing based test failing in one of the CUDA builds

masterleinad (approved):

dalg24 (changes requested):
On 2nd thought I do not agree with the `ScopeGuard` changes.  Please revert them.

dalg24 (approved):

dalg24:
Failure (machine running out of space) are unrelated

PR #5249, opened 2022-07-21
Initial OpenACC support

dalg24:
Revisited #5203 

@seyonglee please review


core/src/Kokkos_OpenACC.hpp
```diff
@@ -0,0 +1,123 @@
...
+#include <string>
+
+namespace Kokkos::Experimental::Impl {
+class OpenACCInternal;
+}
+
+namespace Kokkos::Experimental {
+
+class OpenACC {
+  Impl::OpenACCInternal* m_space_instance = nullptr;
```
crtrott, line 77: This should be Impl::HostSharedPtr<Impl::OpenACCInternal> Or at least make a comment that you need that later.

crtrott (changes requested):

core/src/Kokkos_OpenACC.hpp
crtrott: Otherwise I like it.

dalg24: I meant to do it when we implement multiple instances support

seyonglee:
It seems that macro statement to conditionally include the OpenACC backend (`#if defined(KOKKOS_ENABLE_OPENACC)` ... `#endif` in `Kokkos_OpenACC.h` and `Kokkos_OpenACCSpace.h`) is removed, which will cause compilation errors when targeting non-OpenACC backends, since these header files include OpenACC APIs.

core/src/Kokkos_OpenACC.hpp
seyonglee: Do we want to change this now? or in the later PR?

dalg24:
> It seems that macro statement to conditionally include the OpenACC backend (`#if defined(KOKKOS_ENABLE_OPENACC)` ... `#endif` in `Kokkos_OpenACC.h` and `Kokkos_OpenACCSpace.h`) is removed, which will cause compilation errors when targeting non-OpenACC backends, since these header files include OpenACC APIs.

Yes I will move the file to the `OpenACC` subdirectory which will fix the CI errors.

seyonglee:
Other than the macro statement issue, it looks good to me too.

crtrott (approved):

core/src/OpenACC/Kokkos_OpenACC_Instance.cpp
```diff
@@ -0,0 +1,92 @@
...
+    std::string const& name) const {
+  Kokkos::Tools::Experimental::Impl::profile_fence_event<
+      Kokkos::Experimental::OpenACC>(
+      name,
+      Kokkos::Tools::Experimental::Impl::DirectFenceIDHandle{instance_id()},
+      [&]() {
+        //[DEBUG] disabled due to synchronous behaviors of the current
+        // parallel construct implementations. acc_wait_all();
+      });
+}
```
JBludau, line 86: I personally do not like empty functions for debugging/development, especially if they do not even have side effects. Why not have a fence here, even though it is not necessary at the current state? A `FIXME_OPENACC` would not hurt I guess

JBludau (approved):
No hard feelings concerning my comments, just suggestions. Looks good

cmake/kokkos_arch.cmake
```diff
@@ -124,6 +124,13 @@ IF(KOKKOS_ENABLE_COMPILER_WARNINGS)
     LIST(REMOVE_ITEM COMMON_WARNINGS "-Wsign-compare")
   ENDIF()
 
+  # NVHPC compiler does not support -Wtype-limits.
+  IF(KOKKOS_ENABLE_OPENACC)
+    IF(KOKKOS_CXX_COMPILER_ID STREQUAL NVHPC)
+      LIST(REMOVE_ITEM COMMON_WARNINGS "-Wtype-limits")
+    ENDIF()
+  ENDIF()
```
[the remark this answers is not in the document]
dalg24: Dunno

masterleinad (approved):
Looks OK to me.

cmake/kokkos_arch.cmake
```diff
@@ -490,8 +497,8 @@ FUNCTION(CHECK_CUDA_ARCH ARCH FLAG)
       MESSAGE(FATAL_ERROR "Multiple GPU architectures given! Already have ${CUDA_ARCH_ALREADY_SPECIFIED}, but trying to add ${ARCH}. If you are re-running CMake, try clearing the cache and running again.")
     ENDIF()
     SET(CUDA_ARCH_ALREADY_SPECIFIED ${ARCH} PARENT_SCOPE)
-    IF (NOT KOKKOS_ENABLE_CUDA AND NOT KOKKOS_ENABLE_OPENMPTARGET AND NOT KOKKOS_ENABLE_SYCL)
-      MESSAGE(WARNING "Given CUDA arch ${ARCH}, but Kokkos_ENABLE_CUDA and Kokkos_ENABLE_OPENMPTARGET are OFF. Option will be ignored.")
+    IF (NOT KOKKOS_ENABLE_CUDA AND NOT KOKKOS_ENABLE_OPENMPTARGET AND NOT KOKKOS_ENABLE_SYCL AND NOT KOKKOS_ENABLE_OPENACC)
+      MESSAGE(WARNING "Given CUDA arch ${ARCH}, but Kokkos_ENABLE_CUDA, Kokkos_ENABLE_OPENACC, and Kokkos_ENABLE_OPENMPTARGET are OFF. Option will be ignored.")
```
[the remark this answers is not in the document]
dalg24: I agree with the suggestion but am tempted to ignore for the sake of time.  You can push to my branch if you want it now.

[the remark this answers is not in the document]
JBludau: no, we can exclude it and just do it in another pr ... it is not OpenACC related anyways

PR #5250, opened 2022-07-21
Warn in ScopeGuard about illegal usage

crtrott:
We (Damien and I) are considering a change of behavior and how to use it for ScopeGuard.
Specifically, we believe its current design is problematic in various ways, one of them being the silent ignoring of arguments handed to it if Kokkos was already initialized.

Besides better warnings in cases which were previously already broken (like creating a ScopeGuard after Kokkos::finalize() was called), the primary change is that you can't create a ScopeGuard after Kokkos::initialize was called. This also means that you can't create a ScopeGuard after any other ScopeGuard was created. The error message provided shows a way how to do nested scopeguards (as they might currently be in use) if really desired, but essentially its pushing more of the correctness check behavior on the user. 

Please let us know what you think.

Note: this PR is missing tests, which I am working on.

### DIFF
--- a/core/src/Kokkos_Core.hpp
+++ b/core/src/Kokkos_Core.hpp
@@ -195,7 +195,46 @@ namespace Kokkos {
  *     if Kokkos::is_initialized() in the constructor, don't call
  * Kokkos::initialize or Kokkos::finalize it is not copyable or assignable
  */
+namespace Impl {
 
+inline std::string scopeguard_correct_usage() {
+  return std::string(
+      "Do instead:\n"
+      "  std::unique_ptr<Kokkos::ScopeGuard> guard =\n"
+      "    !Kokkos::is_initialized() && !Kokkos::is_finalized()?\n"
+      "    new ScopeGuard(argc,argv) : nullptr;\n");
+}
+inline std::string scopeguard_create_after_init_warning() {
+  return std::string(
+             "Kokkos Error: Creating a ScopeGuard after Kokkos was initialized "
+             "is illegal.\n")
+      .append(scopeguard_correct_usage());
+}
+
+inline std::string scopeguard_destruct_without_init_warning() {
+  return std::string(
+             "Kokkos Error: Destroying a ScopeGuard after Kokkos was finalized "
+             "is illegal.\n")
+      .append(scopeguard_correct_usage());
+}
+
+inline std::string scopeguard_create_after_finalize_warning() {
+  return std::string(
+             "Kokkos Error: Creating a ScopeGuard after Kokkos was finalized "
+             "is illegal.\n")
+      .append(scopeguard_correct_usage());
+}
+
+inline std::string scopeguard_destruct_after_finalize_warning() {
+  return std::string(
+             "Kokkos Error: Destroying a ScopeGuard after Kokkos was finalized "
+             "is illegal.\n")
+      .append(scopeguard_correct_usage());
+}
+
+}  // namespace Impl
+
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_3
 class KOKKOS_ATTRIBUTE_NODISCARD ScopeGuard {
  public:
 #if defined(__has_cpp_attribute) && __has_cpp_attribute(nodiscard) >= 201907
@@ -203,6 +242,15 @@ class KOKKOS_ATTRIBUTE_NODISCARD ScopeGuard {
 #endif
   ScopeGuard(int& argc, char* argv[]) {
     sg_init = false;
+#ifdef KOKKOS_ENABLE_DEPRECATION_WARNINGS
+    if (is_initialized()) {
+      std::cerr << Impl::scopeguard_create_after_init_warning() << std::endl;
+    }
+    if (is_finalized()) {
+      std::cerr << Impl::scopeguard_create_after_finalize_warning()
+                << std::endl;
+    }
+#endif
     if (!is_initialized()) {
       initialize(argc, argv);
       sg_init = true;
@@ -215,6 +263,15 @@ class KOKKOS_ATTRIBUTE_NODISCARD ScopeGuard {
   ScopeGuard(
       const InitializationSettings& settings = InitializationSettings()) {
     sg_init = false;
+#ifdef KOKKOS_ENABLE_DEPRECATION_WARNINGS
+    if (is_initialized()) {
+      std::cerr << Impl::scopeguard_create_after_init_warning() << std::endl;
+    }
+    if (is_finalized()) {
+      std::cerr << Impl::scopeguard_create_after_finalize_warning()
+                << std::endl;
+    }
+#endif
     if (!is_initialized()) {
       initialize(settings);
       sg_init = true;
@@ -222,6 +279,16 @@ class KOKKOS_ATTRIBUTE_NODISCARD ScopeGuard {
   }
 
   ~ScopeGuard() {
+#ifdef KOKKOS_ENABLE_DEPRECATION_WARNINGS
+    if (!is_initialized() && !is_finalized()) {
+      std::cerr << Impl::scopeguard_destruct_without_init_warning()
+                << std::endl;
+    }
+    if (is_finalized()) {
+      std::cerr << Impl::scopeguard_destruct_after_finalize_warning()
+                << std::endl;
+    }
+#endif
     if (is_initialized() && sg_init) {
       finalize();
     }
@@ -233,6 +300,54 @@ class KOKKOS_ATTRIBUTE_NODISCARD ScopeGuard {
   ScopeGuard& operator=(const ScopeGuard&) = delete;
   ScopeGuard(const ScopeGuard&)            = delete;
 };
+
+#else  // ifndef KOKKOS_ENABLE_DEPRECATED_CODE3
+
+class KOKKOS_ATTRIBUTE_NODISCARD ScopeGuard {
+ public:
+#if defined(__has_cpp_attribute) && __has_cpp_attribute(nodiscard) >= 201907
+  KOKKOS_ATTRIBUTE_NODISCARD
+#endif
+
+  ScopeGuard(int& argc, char* argv[]) {
+    if (is_initialized()) {
+      Kokkos::abort(Impl::scopeguard_create_after_init_warning().c_str());
+    }
+    if (is_finalized()) {
+      Kokkos::abort(Impl::scopeguard_create_after_finalize_warning().c_str());
+    }
+    initialize(argc, argv);
+  }
+
+#if defined(__has_cpp_attribute) && __has_cpp_attribute(nodiscard) >= 201907
+  KOKKOS_ATTRIBUTE_NODISCARD
+#endif
+  ScopeGuard(
+      const InitializationSettings& settings = InitializationSettings()) {
+    if (is_initialized()) {
+      Kokkos::abort(Impl::scopeguard_create_after_init_warning().c_str());
+    }
+    if (is_finalized()) {
+      Kokkos::abort(Impl::scopeguard_create_after_finalize_warning().c_str());
+    }
+    initialize(settings);
+  }
+
+  ~ScopeGuard() {
+    if (!is_initialized() && !is_finalized()) {
+      Kokkos::abort(Impl::scopeguard_destruct_without_init_warning().c_str());
+    }
+    if (is_finalized()) {
+      Kokkos::abort(Impl::scopeguard_destruct_after_finalize_warning().c_str());
+    }
+    finalize();
+  }
+
+  // private:
+  ScopeGuard& operator=(const ScopeGuard&) = delete;
+  ScopeGuard(const ScopeGuard&)            = delete;
+};
+#endif
 
 }  // namespace Kokkos
 
@@ -287,6 +402,7 @@ std::vector<ExecSpace> partition_space(ExecSpace space,
 
 // Specializations requires after core definitions
 #include <KokkosCore_Config_PostInclude.hpp>
+
 //----------------------------------------------------------------------------
 //----------------------------------------------------------------------------
 
--- a/core/src/Kokkos_Core.hpp
+++ b/core/src/Kokkos_Core.hpp
@@ -204,16 +204,10 @@ inline std::string scopeguard_correct_usage() {
       "    !Kokkos::is_initialized() && !Kokkos::is_finalized()?\n"
       "    new ScopeGuard(argc,argv) : nullptr;\n");
 }
-inline std::string scopeguard_create_after_init_warning() {
-  return std::string(
-             "Kokkos Error: Creating a ScopeGuard after Kokkos was initialized "
-             "is illegal.\n")
-      .append(scopeguard_correct_usage());
-}
 
-inline std::string scopeguard_destruct_without_init_warning() {
+inline std::string scopeguard_create_while_initialized_warning() {
   return std::string(
-             "Kokkos Error: Destroying a ScopeGuard after Kokkos was finalized "
+             "Kokkos Error: Creating a ScopeGuard while Kokkos is initialized "
              "is illegal.\n")
       .append(scopeguard_correct_usage());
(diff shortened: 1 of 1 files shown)


core/src/Kokkos_Core.hpp
```diff
@@ -195,14 +195,62 @@ namespace Kokkos {
...
+             "is illegal.\n")
+      .append(scopeguard_correct_usage());
+}
+
+inline std::string scopeguard_destruct_without_init_warning() {
+  return std::string(
+             "Kokkos Error: Destroying a ScopeGuard after Kokkos was finalized "
+             "is illegal.\n")
+      .append(scopeguard_correct_usage());
+}
```
JBludau, line 219: Shouldn't this be: `Destroying a ScopeGuard without Kokkos being initialized is illegal`

core/src/Kokkos_Core.hpp
```diff
@@ -195,14 +195,62 @@ namespace Kokkos {
...
  */
+namespace Impl {
+
+inline std::string scopeguard_correct_usage() {
+  return std::string(
+      "Do instead:\n"
+      "  std::unique_ptr<Kokkos::ScopeGuard> guard =\n"
+      "    !Kokkos::is_initialized() && !Kokkos::is_finalized()?\n"
+      "    new ScopeGuard(argc,argv) : nullptr;\n");
+}
```
JBludau, line 206: Is there a reason not to provide a convenience function for this?

core/src/Kokkos_Core.hpp
```diff
@@ -195,14 +195,62 @@ namespace Kokkos {
...
+}  // namespace Impl
+
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_3
 class KOKKOS_ATTRIBUTE_NODISCARD ScopeGuard {
  public:
 #if defined(__has_cpp_attribute) && __has_cpp_attribute(nodiscard) >= 201907
   KOKKOS_ATTRIBUTE_NODISCARD
 #endif
   ScopeGuard(int& argc, char* argv[]) {
     sg_init = false;
```
nmm0, line 244: Do we still need `sg_init`?

core/src/Kokkos_Core.hpp
```diff
@@ -234,6 +301,54 @@ class KOKKOS_ATTRIBUTE_NODISCARD ScopeGuard {
   ScopeGuard(const ScopeGuard&)            = delete;
 };
 
+#else  // ifndef KOKKOS_ENABLE_DEPRECATED_CODE3
```
nmm0, line 304: In case we ever change the initialization logic, we would have to remember to update the code in two places. I'd suggest delegating the warnings/errors to a function or macro then depending on the state of `KOKKOS_ENABLE_DEPRECATED_CODE3` aborting or warning.

core/src/Kokkos_Core.hpp
```diff
@@ -195,14 +195,62 @@ namespace Kokkos {
  *     if Kokkos::is_initialized() in the constructor, don't call
  * Kokkos::initialize or Kokkos::finalize it is not copyable or assignable
  */
+namespace Impl {
+
+inline std::string scopeguard_correct_usage() {
+  return std::string(
+      "Do instead:\n"
+      "  std::unique_ptr<Kokkos::ScopeGuard> guard =\n"
```
nmm0, line 203: or `std::optional` ;)

core/src/Kokkos_Core.hpp
crtrott: in the deprecated version yes

core/src/Kokkos_Core.hpp
crtrott: But the first branch would go away in 4.0, I just split it completely out so I don't have ifdefs in every second line.

core/src/Kokkos_Core.hpp
```diff
@@ -215,13 +263,32 @@ class KOKKOS_ATTRIBUTE_NODISCARD ScopeGuard {
...
       sg_init = true;
     }
   }
 
   ~ScopeGuard() {
+#ifdef KOKKOS_ENABLE_DEPRECATION_WARNINGS
+    if (!is_initialized() && !is_finalized()) {
+      std::cerr << Impl::scopeguard_destruct_without_init_warning()
+                << std::endl;
+    }
```
dalg24, line 286:
```suggestion
```
Literally cannot happen

dalg24 (commented):
I like the use of function to produce the strings and the general organization besides the comments I made

core/src/Kokkos_Core.hpp
nmm0: ok makes sense, not too worried about it then

core/src/Kokkos_Core.hpp
nmm0: ok, it's a bit mysterious where it comes from but since we're getting rid of the code it doesn't matter

core/src/Kokkos_Core.hpp
```diff
@@ -234,6 +301,54 @@ class KOKKOS_ATTRIBUTE_NODISCARD ScopeGuard {
...
+
+  ScopeGuard(int& argc, char* argv[]) {
+    if (is_initialized()) {
+      Kokkos::abort(Impl::scopeguard_create_after_init_warning().c_str());
+    }
+    if (is_finalized()) {
+      Kokkos::abort(Impl::scopeguard_create_after_finalize_warning().c_str());
+    }
+    initialize(argc, argv);
+  }
```
dalg24, line 320: Commenting so I get a souvenir after Christian renames everything

dalg24 (approved):
Looks good now.
Did you plan on adding tests?

core/src/Kokkos_Core.hpp
```diff
@@ -234,6 +293,53 @@ class KOKKOS_ATTRIBUTE_NODISCARD ScopeGuard {
...
 
+#else  // ifndef KOKKOS_ENABLE_DEPRECATED_CODE3
+
+class KOKKOS_ATTRIBUTE_NODISCARD ScopeGuard {
+ public:
+#if defined(__has_cpp_attribute) && __has_cpp_attribute(nodiscard) >= 201907
+  KOKKOS_ATTRIBUTE_NODISCARD
+#endif
+
+  ScopeGuard(int& argc, char* argv[]) {
```
masterleinad, line 304:
```suggestion
#if defined(__has_cpp_attribute) && __has_cpp_attribute(nodiscard) >= 201907
  KOKKOS_ATTRIBUTE_NODISCARD
#endif
  ScopeGuard(int& argc, char* argv[]) {
```

core/src/Kokkos_Core.hpp
```diff
@@ -234,6 +293,53 @@ class KOKKOS_ATTRIBUTE_NODISCARD ScopeGuard {
...
+  }
+
+  ~ScopeGuard() {
+    if (is_finalized()) {
+      Kokkos::abort(Impl::scopeguard_destruct_after_finalize_warning().c_str());
+    }
+    finalize();
+  }
+
+  // private:
```
masterleinad, line 337:
```suggestion
```

core/src/Kokkos_Core.hpp
```diff
@@ -287,6 +393,7 @@ std::vector<ExecSpace> partition_space(ExecSpace space,
 
 // Specializations requires after core definitions
```
masterleinad, line 394:
```suggestion
// Specializations required after core definitions
```
?

masterleinad (commented):
Looks good to me but you still want to add tests, right?

nliber (changes requested):

core/src/Kokkos_Core.hpp
```diff
@@ -234,6 +293,53 @@ class KOKKOS_ATTRIBUTE_NODISCARD ScopeGuard {
...
+    if (is_finalized()) {
+      Kokkos::abort(Impl::scopeguard_create_after_finalize_warning().c_str());
+    }
+    initialize(argc, argv);
+  }
+
+#if defined(__has_cpp_attribute) && __has_cpp_attribute(nodiscard) >= 201907
+  KOKKOS_ATTRIBUTE_NODISCARD
+#endif
+  ScopeGuard(
```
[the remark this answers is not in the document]
dalg24: Agreed on the `explicit` comment

Interaction with `InitArguments` will be fine https://godbolt.org/z/7scMhjMj6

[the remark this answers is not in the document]
nliber: Before we added `InitializationSettings`, the following three initialization syntaxes work:

```c++
    InitArguments ia;
    
    ScopeGuard sc0 = ia;
    ScopeGuard sc1(ia);
    ScopeGuard sc2{ia};

```

If we do `explicit ScopeGuard(const InitializationSettings& settings) { /* ... */ }`, `sc0` no longer compiles.

For the record, I'm fine with breaking that.  I'm just pointing out that it is a breaking change.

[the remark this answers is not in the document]
dalg24: I tend to think it is fine to break the `ScopeGuard sc0 = ia;` use.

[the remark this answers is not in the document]
masterleinad: I agree.

crtrott:
I won't have time to write tests for 3.7 here is a standalone manual test

```c++
#include <Kokkos_Core.hpp>
#include <cmath>

  {
    int create_after_init = argc > 1 ? atoi(argv[1]) : 0;
    int create_after_final = argc > 2 ? atoi(argv[2]) : 0;
    int create_after_other = argc > 3 ? atoi(argv[3]) : 0;
    int destroy_after_final = argc > 4 ? atoi(argv[4]) : 0;
    if(create_after_other) {
      Kokkos::ScopeGuard guard1(argc,argv);
      Kokkos::ScopeGuard guard2(argc,argv);

    } else {
      if(create_after_init || create_after_final)
      if(create_after_final)
        Kokkos::finalize();
      Kokkos::ScopeGuard guard(argc,argv);
      if((!create_after_final && create_after_init) || destroy_after_final)
        Kokkos::finalize();
    }
  }
}
```


Test results deprecated_code disabled is this, I also confirmed that I get that output with just deprecation warnings on. 
```
[crtrott@kokkos-dev-2 ScopeGuard]$ ./test.cuda 0 0 0 0
[crtrott@kokkos-dev-2 ScopeGuard]$ ./test.cuda 1 0 0 0
Kokkos Error: Creating a ScopeGuard while Kokkos is initialized is illegal.
Do instead:
  std::unique_ptr<Kokkos::ScopeGuard> guard =
    !Kokkos::is_initialized() && !Kokkos::is_finalized()?
    new ScopeGuard(argc,argv) : nullptr;

Backtrace:
                        [0x447473]
                        [0x41b398]
                        [0x41b41b]
                        [0x407731]
                        [0x4058c6]
__libc_start_main [0x7fa3e7bcf555]
                        [0x406d79]
Aborted (core dumped)
[crtrott@kokkos-dev-2 ScopeGuard]$ ./test.cuda 0 1 0 0
Kokkos Error: Creating a ScopeGuard after Kokkos was finalized is illegal.
Do instead:
  std::unique_ptr<Kokkos::ScopeGuard> guard =
    !Kokkos::is_initialized() && !Kokkos::is_finalized()?
    new ScopeGuard(argc,argv) : nullptr;

Backtrace:
                        [0x447473]
                        [0x41b398]
                        [0x41b41b]
                        [0x407763]
                        [0x405895]
__libc_start_main [0x7fa271384555]
                        [0x406d79]
Aborted (core dumped)
[crtrott@kokkos-dev-2 ScopeGuard]$ ./test.cuda 0 0 1 0
Kokkos Error: Creating a ScopeGuard while Kokkos is initialized is illegal.
Do instead:
  std::unique_ptr<Kokkos::ScopeGuard> guard =
    !Kokkos::is_initialized() && !Kokkos::is_finalized()?
    new ScopeGuard(argc,argv) : nullptr;

Backtrace:
                        [0x447473]
                        [0x41b398]
                        [0x41b41b]
                        [0x407731]
                        [0x405965]
__libc_start_main [0x7f8351ae5555]
                        [0x406d79]
Aborted (core dumped)
[crtrott@kokkos-dev-2 ScopeGuard]$ ./test.cuda 0 0 0 1
Kokkos Error: Destroying a ScopeGuard after Kokkos was finalized is illegal.
Do instead:
  std::unique_ptr<Kokkos::ScopeGuard> guard =
    !Kokkos::is_initialized() && !Kokkos::is_finalized()?
    new ScopeGuard(argc,argv) : nullptr;

Backtrace:
                        [0x447473]
                        [0x41b398]
                        [0x41b41b]
                        [0x407798]
                        [0x4058a2]
__libc_start_main [0x7fb800433555]
                        [0x406d79]
Aborted (core dumped)
```

core/src/Kokkos_Core.hpp
```diff
@@ -334,9 +335,11 @@ class KOKKOS_ATTRIBUTE_NODISCARD ScopeGuard {
     finalize();
   }
 
-  // private:
+ private:
```
dalg24, line 338: I am going to pretend I don't see that guy

core/src/Kokkos_Core.hpp
```diff
@@ -195,14 +195,57 @@ namespace Kokkos {
...
  * Kokkos::initialize or Kokkos::finalize it is not copyable or assignable
  */
+namespace Impl {
+
+inline std::string scopeguard_correct_usage() {
+  return std::string(
+      "Do instead:\n"
+      "  std::unique_ptr<Kokkos::ScopeGuard> guard =\n"
+      "    !Kokkos::is_initialized() && !Kokkos::is_finalized()?\n"
+      "    new ScopeGuard(argc,argv) : nullptr;\n");
```
dalg24, line 205: I can live with it but I wanted to point out that this error will be raised even when the `InitializationSettings` argument was used.

core/src/Kokkos_Core.hpp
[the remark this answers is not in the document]
dalg24: @crtrott you forgot to add `explicit` here.  Considering that this is the new code path, I would have been fine with omitting at the other place but it must be added here.

core/src/Kokkos_Core.hpp
```diff
@@ -212,27 +255,93 @@ class KOKKOS_ATTRIBUTE_NODISCARD ScopeGuard {
...
+      Kokkos::abort(Impl::scopeguard_destruct_after_finalize_warning().c_str());
+    }
+    finalize();
+  }
+
+ private:
+  ScopeGuard& operator=(const ScopeGuard&) = delete;
+  ScopeGuard& operator=(ScopeGuard&&) = delete;
+  ScopeGuard(const ScopeGuard&)       = delete;
+  ScopeGuard(ScopeGuard&&)            = delete;
```
masterleinad, line 342: Why are these private?

crtrott: because they were private before

crtrott: or at least meant to be private??

masterleinad: It's more idiomatic not to have private copy constructors if you meant to delete them, right? This feels like part of a pre-C++11 approach where you couldn't use `delete`.

nliber: I agree with @masterleinad here.

Before C++11, the copy operations always existed for every class, so to make them noncopyable one made them `private` and unimplemented (or privately derived from something like `boost::noncopyable`, which did the same thing).  Unimplemented always gave a linker error on an attempted copy, and making them `private` sometimes gave a compile time access error (depending on usage).

No need for that now that we can `delete` functions.

nliber (changes requested):

crtrott:
I removed the private ...

masterleinad (approved):
Thanks!

nliber (approved):

JBludau (approved):

dalg24:
Failures were unrelated (machine running out of space)

PR #5251, opened 2022-07-21
OpenACC select the right device

dalg24:
* Tell the runtime which device to use
* ~~Let `DefaultExecutionSpace` be `OpenACC`~~ withdrawn since it involves more changes than I expected
* ~~Require the user to specify an AMD GPU or NVIDIA architecture at configure time~~
* Pass `-gpu=ccXY` to the NVHPC compiler when the NVIDIA GPU arch is provided


cmake/kokkos_arch.cmake
```diff
@@ -663,9 +663,19 @@ IF (KOKKOS_ENABLE_OPENMPTARGET)
...
+    COMPILER_SPECIFIC_FLAGS(
+      NVHPC -acc -gpu=${NVHPC_CUDA_ARCH}
+    )
+  ELSEIF(KOKKOS_AMDGPU_ARCH_FLAG)
+    # FIXME_OPENACC
+    COMPILER_SPECIFIC_FLAGS(
+      NVHPC -acc
+    )
+  ELSE()
+    MESSAGE(SEND_ERROR "Must specify a GPU architecture with OpenACC")
```
seyonglee, line 677: OpenACC does not requires target device type/architecture at compile time, which can be set at runtime using an environment variable (`ACC_DEVICE_TYPE`) or the implementation can choose a default device type.
For NVHPC compiler, `-gpu` option is optional; if not set, the compiler will generate binary for all available (and supported) architectures.
Therefore, an alternative option for `ELSE()` case is to let the NVHPC compiler to choose available device architecture (`NVHPC -acc`).

core/src/OpenACC/Kokkos_OpenACC_Traits.hpp
```diff
@@ -0,0 +1,68 @@
...
+
+#include <Kokkos_Macros.hpp>
+
+#include <openacc.h>
+
+namespace Kokkos::Experimental::Impl {
+
+struct OpenACC_Traits {
+  static constexpr acc_device_t dev_type =
+#if defined(KOKKOS_ARCH_VOLTA) || defined(KOKKOS_ARCH_AMPERE)
```
seyonglee, line 56: Why limiting only to Volta and Ampere architectures? NVHPC compiler supports several pre-Volta architectures (up to Kepler architecture).

core/src/OpenACC/Kokkos_OpenACC_Traits.hpp
```diff
@@ -0,0 +1,68 @@
...
+      acc_device_nvidia;
+#elif defined(KOKKOS_ARCH_VEGA)
+      acc_device_radeon;
+#else
+      acc_device_host;
```
seyonglee, line 61: How about changing `acc_device_host` to `acc_device_default` so that the OpenACC compiler can choose a target device?

cmake/kokkos_arch.cmake
```diff
@@ -663,9 +663,19 @@ IF (KOKKOS_ENABLE_OPENMPTARGET)
...
-  )
+  IF(KOKKOS_CUDA_ARCH_FLAG)
+    STRING(REPLACE "sm_" "cc" NVHPC_CUDA_ARCH ${KOKKOS_CUDA_ARCH_FLAG})
+    COMPILER_SPECIFIC_FLAGS(
+      NVHPC -acc -gpu=${NVHPC_CUDA_ARCH}
+    )
+  ELSEIF(KOKKOS_AMDGPU_ARCH_FLAG)
+    # FIXME_OPENACC
+    COMPILER_SPECIFIC_FLAGS(
+      NVHPC -acc
```
seyonglee, line 674: NVHPC compiler supports AMD CPUs but NOT AMD GPUs.

dalg24: Duh :)
Should I just error out if no CUDA arch was set ?

core/src/OpenACC/Kokkos_OpenACC_Traits.hpp
dalg24: Sounds good.  I overlooked that it was an option.

core/src/OpenACC/Kokkos_OpenACC_Traits.hpp
dalg24: We are about to drop support for Kepler and Maxwell.  I was going for a cheap way to find out if we are targeting NVIDIA or AMD GPUs.  Not that I had not realized `acc_device_default` existed when I wrote that.

cmake/kokkos_arch.cmake
dalg24: I am fine withdrawing these changes for now.
With the changes in this first version I needed to know at compile-time whether we were targeting NVIDIA or AMD gpu but I will go for `acc_default_device` and drop this for now.
Let's bring it up with the team at the next developer meeting.

seyonglee: What I suggest is not to remove all the architecture-specific options (`-gpu`); instead make it optional.
Then, the code will look like this:

IF(KOKKOS_CUDA_ARCH_FLAG)
     STRING(REPLACE "sm_" "cc" NVHPC_CUDA_ARCH ${KOKKOS_CUDA_ARCH_FLAG})
     COMPILER_SPECIFIC_FLAGS(
       NVHPC -acc -gpu=${NVHPC_CUDA_ARCH}
     )
ELSE()
     COMPILER_SPECIFIC_FLAGS(
       NVHPC -acc
     )
ENDIF()

dalg24: Got it.  Will update.

core/src/OpenACC/Kokkos_OpenACC_Traits.hpp
seyonglee: `acc_device_default` should work all the cases, but specifying a specific device type will be still useful.
If support for Kepler and Maxwell will be dropped, we can use the following code:

static constexpr acc_device_t dev_type =
 #if defined(KOKKOS_ARCH_VOLTA) || defined(KOKKOS_ARCH_AMPERE)
       acc_device_nvidia;
 #else
       acc_device_default;
 #endif

dalg24: Done

core/src/OpenACC/Kokkos_OpenACC_Instance.cpp
```diff
@@ -56,7 +60,12 @@ Kokkos::Experimental::Impl::OpenACCInternal::singleton() {
   return &self;
 }
 
-void Kokkos::Experimental::Impl::OpenACCInternal::initialize() {
+void Kokkos::Experimental::Impl::OpenACCInternal::initialize(
+    InitializationSettings const& settings) {
+  using Kokkos::Impl::get_gpu;
+  int dev_num           = get_gpu(settings);
+  acc_device_t dev_type = OpenACC_Traits::dev_type;
+  acc_set_device_num(dev_num, dev_type);
```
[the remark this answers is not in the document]
seyonglee: `acc_device_default` does not distinguish host from accelerator devices, and thus if we want to allow only accelerator devices as an OpenACC backend target, we will have to change the default device type to `acc_device_no_host` and add extra code to check whether the current device is a host or not for the non-default cases.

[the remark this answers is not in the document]
masterleinad: For `SYCL` we accept any kind of device if no device id was specified. If a device id is specified, we error out if the device id is larger than the number of devices.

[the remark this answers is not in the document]
dalg24: We could do something like
```C++
int num_devices = acc_get_num_devices(
       Kokkos::Experimental::Impl::OpenACC_Traits::dev_type);
if (num_devices == 0 && !settings.has_device_id()) {
  acc_set_device_type(acc_device_host);
} else {
  ...
}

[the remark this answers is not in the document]
Comment (seyonglee) [reply core/src/OpenACC/Kokkos_OpenACC_Instance.cpp:68]: What if `settings.has_device_id()` is not zero?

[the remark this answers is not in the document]
Comment (dalg24) [reply core/src/OpenACC/Kokkos_OpenACC_Instance.cpp:68]: `settings.has_device_id()` is `true` if the user explicitly asked for a device id in which case we do do not want to silently fall back onto the host.
If the runtime reports that they are at least one GPU we defer the selection to `Impl::get_gpu` in core so that alll backends behave the same.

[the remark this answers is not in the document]
Comment (dalg24) [reply core/src/OpenACC/Kokkos_OpenACC_Instance.cpp:68]: I realize now we should also NOT do the fallback if the user specified a GPU arch at configure time :/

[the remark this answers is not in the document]
Comment (seyonglee) [reply core/src/OpenACC/Kokkos_OpenACC_Instance.cpp:68]: @dalg24 you are correct; to fall back to host, compiler option should be configured to support host too.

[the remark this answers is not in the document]
Comment (seyonglee) [reply core/src/OpenACC/Kokkos_OpenACC_Instance.cpp:68]: `Implementations::get_gpu` will handle whether user-specified device id is valid or not, and thus we don't have to check the validity of the device id again here.

File: core/src/OpenACC/Kokkos_OpenACC_Instance.cpp [hunk 6]
```diff
@@ -46,17 +46,40 @@
...
+      std::cerr << "Warning: No GPU available for execution, falling back to"
+                   " using the host!"
+                << std::endl;
+    }
+    acc_set_device_type(acc_device_host);
+  } else {
+    using Kokkos::Impl::get_gpu;
+    int const dev_num = get_gpu(settings);
+    acc_set_device_num(dev_num, OpenACC_Traits::dev_type);
+  }
```
dalg24, line 82: @seyonglee @masterleinad how does that sound?

seyonglee: It sounds reasonable.
By the way, is it possible for `get_gpu(settings)` to return negative value? If so, we may have to insert code to check validity of `dev_num` in line 80 too.

masterleinad: Looks good to me. I think we should also check in https://github.com/kokkos/kokkos/blob/0ef409bab190cdae9417bff34f7cf5c90048f227/core/src/impl/Kokkos_Core.cpp#L355-L357 that `settings.get_device_id()` is less than `visible_devices.size()` but that's independent of this pull request.

dalg24: The intent is that `Impl::get_gpu` will do the validation.  See #5253 
As it is currently implemented, you could technically get in trouble when using `KOKKOS_VISIBLE_DEVICES` but as this is an issue for all backends, we will improve this and handle it in "core".

core/src/OpenACC/Kokkos_OpenACC_Instance.cpp
```diff
@@ -46,7 +46,6 @@
 
 #include <OpenACC/Kokkos_OpenACC_Instance.hpp>
 #include <OpenACC/Kokkos_OpenACC.hpp>
-#include <OpenACC/Kokkos_OpenACC_Traits.hpp>
```
dalg24: Keep that header include please.

dalg24 (commented):
Looks good other than the header include removal

core/src/OpenACC/Kokkos_OpenACC_Instance.cpp
seyonglee: Included again.

seyonglee (approved):
Looks good to me.

dalg24:
(I fixed the compile error from the previous change and amended to the previous commit)
```
"/var/jenkins/workspace/Kokkos/core/src/OpenACC/Kokkos_OpenACC.hpp", line 121: error: name followed by "::" must be a class or namespace name
      return acc_get_device_num(OpenACC_Traits::dev_type);
                                ^

**Comment by dalg24** [comment]
We had a number of these HPX failures recently
```
[70/534] Building CXX object core/unit_test/CMakeFiles/KokkosCore_UnitTest_HPX.dir/hpx/TestHPX_MathematicalFunctions1.cpp.o
[74](https://github.com/kokkos/kokkos/runs/7475763298?check_suite_focus=true#step:9:75)
FAILED: core/unit_test/CMakeFiles/KokkosCore_UnitTest_HPX.dir/hpx/TestHPX_MathematicalFunctions1.cpp.o 
[75](https://github.com/kokkos/kokkos/runs/7475763298?check_suite_focus=true#step:9:76)
/usr/bin/clang++ -DHPX_APPLICATION_EXPORTS -DHPX_APPLICATION_NAME_DEFAULT=KokkosCore_UnitTest_HPX -DHPX_DEBUG -DHPX_PREFIX_DEFAULT=\"/home/runner/work/kokkos/kokkos/hpx/install\" -DKOKKOS_DEPENDENCE -I/home/runner/work/kokkos/kokkos/kokkos/tpls/gtest -I/home/runner/work/kokkos/kokkos/kokkos/build/core/unit_test -I/home/runner/work/kokkos/kokkos/kokkos/core/unit_test -I/home/runner/work/kokkos/kokkos/kokkos/core/unit_test/category_files -I/home/runner/work/kokkos/kokkos/kokkos/build -I/home/runner/work/kokkos/kokkos/kokkos/build/core/src -I/home/runner/work/kokkos/kokkos/kokkos/core/src -I/home/runner/work/kokkos/kokkos/kokkos/core/src/../../tpls/desul/include -I/home/runner/work/kokkos/kokkos/kokkos/build/containers/src -I/home/runner/work/kokkos/kokkos/kokkos/containers/src -I/home/runner/work/kokkos/kokkos/kokkos/build/algorithms/src -I/home/runner/work/kokkos/kokkos/kokkos/algorithms/src -isystem /home/runner/work/kokkos/kokkos/hpx/install/include -Werror -Wall -Wunused-parameter -Wshadow -pedantic -Wsign-compare -Wtype-limits -Wuninitialized -Wimplicit-fallthrough -g -pthread -march=native -mtune=native -std=gnu++14 -MD -MT core/unit_test/CMakeFiles/KokkosCore_UnitTest_HPX.dir/hpx/TestHPX_MathematicalFunctions1.cpp.o -MF core/unit_test/CMakeFiles/KokkosCore_UnitTest_HPX.dir/hpx/TestHPX_MathematicalFunctions1.cpp.o.d -o core/unit_test/CMakeFiles/KokkosCore_UnitTest_HPX.dir/hpx/TestHPX_MathematicalFunctions1.cpp.o -c /home/runner/work/kokkos/kokkos/kokkos/build/core/unit_test/hpx/TestHPX_MathematicalFunctions1.cpp
[76](https://github.com/kokkos/kokkos/runs/7475763298?check_suite_focus=true#step:9:77)
Killed
[77](https://github.com/kokkos/kokkos/runs/7475763298?check_suite_focus=true#step:9:78)
[71/534] Building CXX object core/unit_test/CMakeFiles/KokkosCore_UnitTest_HPX.dir/hpx/TestHPX_Other.cpp.o
[78](https://github.com/kokkos/kokkos/runs/7475763298?check_suite_focus=true#step:9:79)
[72/534] Building CXX object core/unit_test/CMakeFiles/KokkosCore_UnitTest_HPX.dir/hpx/TestHPX_NumericTraits.cpp.o
[79](https://github.com/kokkos/kokkos/runs/7475763298?check_suite_focus=true#step:9:80)
ninja: build stopped: subcommand failed.
```
Obviously this is unrelated to this PR.

JBludau (approved):
looks good

crtrott (approved):

PR #5256, opened 2022-07-22
Warn users when a flag prefixed with `-[-]kokkos` is not recognized and do not remove it

dalg24:
* Do not silently ignore flags we do not recognize when parsing command line arguments if they start with `--kokkos` or `-kokkos`
* Provide a function to tell Kokkos Core not to emit a warning for flags that match some user provided regular expression `Impl::do_not_warn_not_recognized_command_line_argument`.  By default, we only exclude those starting with `--kokkos-tool` but we could expand later as needed.
* Fix bug: do not remove flags prefixed with `--kokkos` that were not actually handled

### DIFF
--- a/core/src/impl/Kokkos_Command_Line_Parsing.hpp
+++ b/core/src/impl/Kokkos_Command_Line_Parsing.hpp
@@ -62,6 +62,7 @@ void warn_deprecated_environment_variable(std::string deprecated,
 void warn_deprecated_command_line_argument(std::string deprecated);
 void warn_deprecated_command_line_argument(std::string deprecated,
                                            std::string use_instead);
+void warn_not_recognized_command_line_argument(std::string not_recognized);
 }  // namespace Impl
 }  // namespace Kokkos
 
--- a/core/src/impl/Kokkos_Command_Line_Parsing.hpp
+++ b/core/src/impl/Kokkos_Command_Line_Parsing.hpp
@@ -46,6 +46,7 @@
 #define KOKKOS_COMMAND_LINE_PARSING_HPP
 
 #include <string>
+#include <regex>
 
 namespace Kokkos {
 namespace Impl {
@@ -63,6 +64,7 @@ void warn_deprecated_command_line_argument(std::string deprecated);
 void warn_deprecated_command_line_argument(std::string deprecated,
                                            std::string use_instead);
 void warn_not_recognized_command_line_argument(std::string not_recognized);
+void do_not_warn_not_recognized_command_line_argument(std::regex ignore);
 }  // namespace Impl
 }  // namespace Kokkos
 
--- a/core/src/impl/Kokkos_Core.cpp
+++ b/core/src/impl/Kokkos_Core.cpp
@@ -874,6 +874,9 @@ void Kokkos::Impl::parse_command_line_arguments(
         Kokkos::abort(ss.str().c_str());
       }
       settings.set_map_device_id_by(map_device_id_by);
+    } else if (std::regex_match(argv[iarg],
+                                std::regex("-?-kokkos.*", std::regex::egrep))) {
+      warn_not_recognized_command_line_argument(argv[iarg]);
     }
 
     // remove flag if prefixed with '--kokkos-'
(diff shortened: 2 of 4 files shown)


dalg24:
PS: feel free to suggest a better name for the function

core/src/impl/Kokkos_Command_Line_Parsing.hpp
```diff
@@ -62,6 +63,8 @@ void warn_deprecated_environment_variable(std::string deprecated,
 void warn_deprecated_command_line_argument(std::string deprecated);
 void warn_deprecated_command_line_argument(std::string deprecated,
                                            std::string use_instead);
+void warn_not_recognized_command_line_argument(std::string not_recognized);
+void do_not_warn_not_recognized_command_line_argument(std::regex ignore);
```
dalg24, line 67: An alternative here would be to taken the prefix the option not to warn about start with (take `std::string` instead of `std::regex`)
Let me know which you prefer

JBludau: I do like the regex. I think it is more flexible than passing a string.
Do you plan on adding a test?

dalg24: Added

JBludau (approved):

core/src/impl/Kokkos_Command_Line_Parsing.hpp
dalg24: So extending the unit test 
```diff
@@ -290,6 +297,21 @@ TEST(defaultdevicetype, cmd_line_args_unrecognized_flag) {
   Kokkos::Impl::parse_command_line_arguments(cla.argc(), cla.argv(), settings);
   captured = ::testing::internal::GetCapturedStderr();
   EXPECT_TRUE(captured.empty() && !settings.has_num_threads()) << captured;
+  EXPECT_EQ(cla.argc(), 1);
+  EXPECT_STREQ(cla.argv()[0], "--kokko-num-threads=4");
+
+  Kokkos::Impl::do_not_warn_not_recognized_command_line_argument(
+      std::regex{"^--kokkos-extension.*"});
+  cla = {{
+      "--kokkos-extension-option=value",  // user explicitly asked not to warn
+                                          // about that prefix
+  }};
+  ::testing::internal::CaptureStderr();
+  Kokkos::Impl::parse_command_line_arguments(cla.argc(), cla.argv(), settings);
+  captured = ::testing::internal::GetCapturedStderr();
+  EXPECT_TRUE(captured.empty()) << captured;
+  EXPECT_EQ(cla.argc(), 1);
+  EXPECT_STREQ(cla.argv()[0], "--kokkos-extension-option=value");
 }

 TEST(defaultdevicetype, env_vars_num_threads) {
```
exposed an issue with the current implementation
```
[ RUN      ] defaultdevicetype.cmd_line_args_unrecognized_flag
/Users/qdi/CLionProjects/kokkos/core/unit_test/TestParseCmdLineArgsAndEnvVars.cpp:313: Failure
  cla.argc()
    Which is: 0
  1
[  FAILED  ] defaultdevicetype.cmd_line_args_unrecognized_flag (1 ms)
```
We are removing all flags that are prefixed with `--kokkos` regardless of whether we handled them or not...

JBludau: we should mark this as a feature ...

dalg24: Fixed

dalg24:
Retest this please

core/src/impl/Kokkos_Core.cpp
```diff
@@ -756,9 +756,13 @@ void Kokkos::Impl::parse_command_line_arguments(
 
   int iarg = 0;
   while (iarg < argc) {
+    bool remove_flag = false;
+
     if (check_arg(argv[iarg], "--kokkos-numa") ||
         check_arg(argv[iarg], "--numa")) {
       warn_deprecated_command_line_argument(get_flag(argv[iarg]));
+      // remove flag if prefixed with '--kokkos-'
+      remove_flag = std::string(argv[iarg]).find("--kokkos-") == 0;
```
PhilMiller, line 765:
```suggestion
      bool begins_with_kokkos = std::string(argv[iarg]).find("--kokkos-") == 0; // move up to top
      remove_flag = begins_with_kokkos;
```

PhilMiller (approved):

crtrott (approved):

Issue #5252, opened 2022-07-22
strange behavior of CudaUVMSpace in develop branch


XzzX:
I am seeing a strange performance regression when changing from Kokkos `3.6.01` to the current `develop` branch.
Code:
 * Cuda 11.4 and run on A100
 * all memory is allocated in CudaUVMSpace
 * I am allocating a bunch of matrices and right hand sides
 * I am calling 5 times `cublasDgelsBatched` on exactly the same data allocated in CudaUVMSpace

Kokkos 3.6.01
![fast](https://user-images.githubusercontent.com/2733900/180416313-559a6633-0427-4d56-b1e4-3ed139e7c08a.png)
This is what I expected to happen.

Kokkos develop
![slow](https://user-images.githubusercontent.com/2733900/180416359-fb2df360-1f39-411a-9fc7-53737f2939bd.png)
All DtoH traffic is evictions?!? Of course, this version is much slower.

Everything is reproducible by changing between the two Kokkos versions. Is anything changed in the way how Kokkos handles UVM? The benchmark code is already a MWE of a larger code and can be found here: https://gitlab.mpcdf.mpg.de/seibl/batched_gels

dalg24:
Do you mind trying 9beba321593b97645b0bda866a8e84c5cc9fc050 ?

XzzX:
> Do you mind trying [9beba32](https://github.com/kokkos/kokkos/commit/9beba321593b97645b0bda866a8e84c5cc9fc050) ?

This commit is fine. Performance similar to `3.6.01`.

dalg24:
What about 7ffb61a105ae513a66ad95a9e9c525411900b8d1

crtrott:
Hm I wasn't able to reproduce it. How did you run it (also I disabled the MKL run)

crtrott:
Even with MKL enabled I was not able to reproduce it. 

Develop:
```
 GPU activities:   58.91%  10.3518s         8  1.29398s  1.28917s  1.30694s  _ZN6Kokkos4Impl33cuda_parallel_launch_local_memoryINS0_11ParallelForIZ12batched_gelsILb0EEvP13cublasContextiiEUliiE_NS_13MDRangePolicyIJNS_4CudaENS_4RankILj2ELNS_7IterateE0ELSA_0EEEEEES8_EEEEvT_
[...]

PR #5259, opened 2022-07-25
Restore old behavior in device selection when user specifies nothing

dalg24:
As reported in #5252, #5211 accidentally changed the default behavior.
The new behavior was "when the user specifies nothing fallback to using the first GPU available".  This PR revert it to the old behavior that is "always attempt to detect local MPI rank and do round-robin assignment".

### DIFF
--- a/core/src/impl/Kokkos_Core.cpp
+++ b/core/src/impl/Kokkos_Core.cpp
@@ -375,7 +375,7 @@ int Kokkos::Impl::get_gpu(const InitializationSettings& settings) {
   }
 
   // either map_device_id_by is not specified or it is mpi_rank
-  if (!settings.has_map_device_id_by() ||
+  if (settings.has_map_device_id_by() &&
       settings.get_map_device_id_by() != "mpi_rank") {
     Kokkos::abort("implementation bug");
   }
--- a/core/src/impl/Kokkos_Core.cpp
+++ b/core/src/impl/Kokkos_Core.cpp
@@ -355,14 +355,10 @@ int Kokkos::Impl::get_gpu(const InitializationSettings& settings) {
   if (settings.has_device_id()) {
     return visible_devices[settings.get_device_id()];
   }
-  // by default use the first GPU available for execution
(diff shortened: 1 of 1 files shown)


JBludau (approved):
Looks good

core/src/impl/Kokkos_Core.cpp
```diff
@@ -355,29 +355,28 @@ int Kokkos::Impl::get_gpu(const InitializationSettings& settings) {
...
+      settings.get_map_device_id_by() == "random") {
     std::default_random_engine gen(get_process_id());
     std::uniform_int_distribution<int> distribution(0,
                                                     visible_devices.size() - 1);
     return visible_devices[distribution(gen)];
   }
 
-  if (settings.get_map_device_id_by() != "mpi_rank") {
+  // either map_device_id_by is not specified or it is mpi_rank
+  if (!settings.has_map_device_id_by() ||
```
masterleinad, line 378: Shouldn't this be
```suggestion
  if (settings.has_map_device_id_by() &&
```

dalg24: No, this is also the code path for "the user specified nothing".  See the comment above.

masterleinad: Yes, the current code aborts if either `has_map_device_id_by` hasn't been specified or it is not "mpi_rank". The one condition we want to prevent is that `has_map_device_id` is specified and it's not "mpi_rank" AFAICT.

Executing the code in its current form I run into the `abort` of not specifying anything, also see https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/Kokkos/detail/Kokkos/9691/pipeline.

dalg24: You're right

masterleinad (approved):

crtrott (approved):
Follow up: warn if mpi_rank is set but can't detect local_rank

Issue #5258, opened 2022-07-23
create_mirror_view_and_copy fails on Windows


j8asic:
Tried after a while Windows compiling and the result is: `Kokkos::create_mirror_view_and_copy` from CUDA space into `Kokkos::HostSpace()` fails, and all the unit tests and algorithms (e.g. scan) that depend on it.

Compilers tried: 
- pure clang++, versions 12 and 13, C++17, C++20
- NVCC 11.7 + MSVC, C++17

Compiled backends: 
- Serial
- CUDA

Some failed unit tests under NVCC+MSVC:
- [  FAILED  ] cuda.parallel_scan_with_reducers
- [  FAILED  ] cuda.view_api_b
- [  FAILED  ] cuda.view_allocation_large_rank
- [  FAILED  ] cuda.view_shmem_size_on_device
- [  FAILED  ] cuda.workgraph_fib 
- [  FAILED  ] cuda.complex_issue_3865
- [  FAILED  ] cuda.crs_count_fill
- [  FAILED  ] cuda.crs_copy_constructor
- [  FAILED  ] cuda.deep_copy_scratch
- [  FAILED  ] cuda.max_within_parfor
- [  FAILED  ] cuda.min_within_parfor
- [  FAILED  ] cuda.minmax_within_parfor
- [  FAILED  ] cuda.clamp_within_parfor
- [  FAILED  ] cuda.range_scan
- [  FAILED  ] cuda.view_mapping

Any ideas what might be the issue?

p.s. There are many annoyances that need to be fixed for (NVCC+MSVC), but that is another subject

masterleinad:
What does the output for the `FAILED` tests look like? Is there some kind of exception/memory access violation or is the function just returning the wrong result?

crtrott:
@j8asic can you try my PR branch? With CUDA 11.4 and Visual Studio 19 that is compiling and passing tests for me.

crtrott:
Just for reference with VS 2022 I started with a standard x64-MSVC config and added these:
```
-DKokkos_ENABLE_CUDA=ON -DKokkos_ENABLE_CUDA_LAMBDA=ON -DKokkos_ARCH_VOLTA70=ON -DKokkos_ENABLE_TESTS=ON -DKokkos_ENABLE_COMPILE_AS_CMAKE_LANGUAGE=ON
``` 
as CMAKE config args. No other editing of the config file.

crtrott:
For reference: the error before the fix:

```
[...]

PR #5261, opened 2022-07-25
Fix CUDA+MSVC build issue

crtrott:
This fixes a build issue I ran into when investigating #5258 using CUDA 11.4 and Visual Studio 19


crtrott:
Ups looks like it need to do the same thing for the CUDA specialization.

core/src/Kokkos_CudaSpace.hpp
```diff
@@ -597,8 +597,26 @@ class SharedAllocationRecord<Kokkos::CudaSpace, void>
...
+        m_space(arg_space) {
+
+    SharedAllocationHeader header;
+
+    this->base_t::_fill_host_accessible_header_info(header, arg_label);
+
+    // Copy to device memory
+    Kokkos::Impl::DeepCopy<CudaSpace, HostSpace>(
+        arg_exec_space, RecordBase::m_alloc_ptr, &header,
+        sizeof(SharedAllocationHeader));
```
masterleinad, line 618: This overload should use the synchronous version, right?
```suggestion
          Impl::checked_allocation_with_header(arg_space, arg_label,
                                               arg_alloc_size),
          sizeof(SharedAllocationHeader) + arg_alloc_size, arg_dealloc,
          arg_label),
      m_tex_obj(0),
      m_space(arg_space) {

  SharedAllocationHeader header;

  this->base_t::_fill_host_accessible_header_info(header, arg_label);

  // Copy to device memory
  Kokkos::Cuda exec;
  Kokkos::Impl::DeepCopy<CudaSpace, HostSpace>(
      exec, RecordBase::m_alloc_ptr, &header, sizeof(SharedAllocationHeader));
  exec.fence(
      "SharedAllocationRecord<Kokkos::CudaSpace, "
      "void>::SharedAllocationRecord(): fence after copying header from "
      "HostSpace");
```

crtrott: Did I make a mistake? I think I just copied over what was getting called?

crtrott:
ust for reference with VS 2022 I started with a standard x64-MSVC config and added these:
```
-DKokkos_ENABLE_CUDA=ON -DKokkos_ENABLE_CUDA_LAMBDA=ON -DKokkos_ARCH_VOLTA70=ON -DKokkos_ENABLE_TESTS=ON -DKokkos_ENABLE_COMPILE_AS_CMAKE_LANGUAGE=ON
``` 
as CMAKE config args. No other editing of the config file. 

This build compiled and passed all tests. Visual Studio 2022, version 17.2.5 with CUDA 11.7.

j8asic:
For now what I have is:
- ErrorReporterViaLambda test cannot compile, it gets confused by `driver_base::m_errorReporter.add_report(work_idx, report);`
- Otherwise it compiles for VS 2022 and tests go okay
- In `Kokkos_FunctorAnalysis.hpp` struct `VOID` makes problems due to unknown reasons, when using kokkos in other projects. When `VOID` is renamed (e.g. to `VOID_TAG`), all is good.
- Kokkos_ENABLE_LAUNCH_COMPILER=OFF is not respected, it is always used and my Windows dont know how to use it. I have to comment out `kokkos_compilation(GLOBAL)` in KokkosConfig.cmake
- I do not what to do with Kokkos_ENABLE_COMPILE_AS_CMAKE_LANGUAGE in my projects :(

core/src/Kokkos_CudaSpace.hpp
```diff
@@ -716,14 +753,27 @@ class SharedAllocationRecord<Kokkos::CudaHostPinnedSpace, void>
...
+                                    void>::s_root_record,
+#endif
+            Impl::checked_allocation_with_header(arg_space, arg_label,
+                                                 arg_alloc_size),
+            sizeof(SharedAllocationHeader) + arg_alloc_size, arg_dealloc,
+            arg_label),
+        m_space(arg_space) {
+    this->base_t::_fill_host_accessible_header_info(*base_t::m_alloc_ptr,
+                                                    arg_label);
+  }
```
dalg24, line 776: Please move the definition to the source and the comment about not deferring to the source file next to the other one.

core/src/Kokkos_CudaSpace.hpp
```diff
@@ -656,14 +680,27 @@ class SharedAllocationRecord<Kokkos::CudaUVMSpace, void>
...
+#endif
+            Impl::checked_allocation_with_header(arg_space, arg_label,
+                                                 arg_alloc_size),
+            sizeof(SharedAllocationHeader) + arg_alloc_size, arg_dealloc,
+            arg_label),
+        m_tex_obj(0),
+        m_space(arg_space) {
+    this->base_t::_fill_host_accessible_header_info(*base_t::m_alloc_ptr,
+                                                    arg_label);
+  }
```
dalg24, line 703: Move next to the other definition in source file

core/src/Kokkos_CudaSpace.hpp
```diff
@@ -610,6 +630,10 @@ class SharedAllocationRecord<Kokkos::CudaSpace, void>
       const size_t arg_alloc_size,
       const RecordBase::function_type arg_dealloc = &base_t::deallocate);
 
+  // helper function to work around MSVC+NVCC issue
+  // https://github.com/kokkos/kokkos/issues/5258
+  void deep_copy_header_no_exec(void*, void*);
```
dalg24, line 635:
```suggestion
  void deep_copy_header_no_exec(void*, void const*);
```

core/src/Kokkos_CudaSpace.hpp
dalg24: OK I realize now it is templated.  In practice we know all execution spaces so we could instantiate everything but that is beyond the scope of this PR.
Please expand the comment to capture why things are implemented the way they are.

core/src/Kokkos_CudaSpace.hpp
dalg24: I was about to comment there is no reason to add here once you move the definitions to the source file but since we won't do it in this PR, please declare that guy as a static member function.

core/src/impl/Kokkos_FunctorAnalysis.hpp
```diff
@@ -890,7 +890,9 @@ struct FunctorAnalysis {
   };
 
   template <class F>
-  struct DeduceTeamShmem<F, std::enable_if_t<0 < sizeof(&F::shmem_size)>> {
+  struct DeduceTeamShmem<F,
+                         std::enable_if_t<(0 < sizeof(&F::shmem_size)) &&
+                                          !(0 < sizeof(&F::team_shmem_size))>> {
```
dalg24, line 895: Did you mean to commit this

masterleinad (commented):
I'm not too fond of the asymmetry created with respect to the other backends too much. I assume we have the same issue for all other backends if used with this compiler combination.
Do we have any idea what the underlying issue really is? Is it a known problem?

I'm OK with this solution for now as a stop-gap assuming there are no other backends that can be used with MSVC+CUDA as of now (what about `Threads`?) but would like to revisit when we have more time.

crtrott:
Daniel: this is only affecting memory spaces, hence it is agnostic to Threads, vs Serial backend and what not. And NVCC is notoriously bad at compiling HIP, SYCL or OpenMPTarget so the compiler combination does not apply :-)

crtrott:
@j8asic can you explain how your windows does not know how to use the LaunchCompiler? What is the failure mode? I guess my build by default uses ninja anyway?
I don't see the compiling issue for ErrorReporter what is your exact CUDA/MSVC version? Are you on the latest (I have no clue how to use older MSVC versions though after updating).

crtrott:
Oh and the language thing: there is really no good way around it, since we would need to figure out how to write something like nvcc_wrapper for windows, and convince VisualStudio that its a good compiler.

crtrott:
Also the VOID problem: it is a nested class (i.e. a class inside a class inside Kokkos::Impl namespace) so are you saying someone else has VOID defined as a macro?

masterleinad (approved):

j8asic:
> What is the failure mode? I guess my build by default uses ninja anyway?

```
CreateProcess failed. Command attempted:
  "C:/Code/kokkos/out/install/x64-Release/bin/kokkos_launch_compiler C:/Code/kokkos/out/install/x64-Release/bin/nvcc_wrapper C:/Program Files/Microsoft Visual Studio/2022/Enterprise/VC/Tools/MSVC/14.32.31326/bin/Hostx64/x64/cl.exe C:\PROGRA~1\MIB055~1\2022\ENTERP~1\VC\Tools\MSVC\1432~1.313\bin\Hostx64\x64\cl.exe  /nologo -DGL_GLEXT_PROTOTYPES=1 -DNOMINMAX -D_CRT_SECURE_NO_WARNINGS ... test.cpp
  ninja: fatal: CreateProcess: %1 is not a valid Win32 application.
```
  
> I don't see the compiling issue for ErrorReporter what is your exact CUDA/MSVC version?

latest msvc in vs 2022, and latest cuda 11.7

> so are you saying someone else has VOID defined as a macro?

this is what I also thought but i didnt dig, I directly renamed it to move forward

masterleinad:
> Daniel: this is only affecting memory spaces, hence it is agnostic to Threads, vs Serial backend and what not. And NVCC is notoriously bad at compiling HIP, SYCL or OpenMPTarget so the compiler combination does not apply :-)

Yes, you're right, we only have one Host memory space (apart from HBM that is possibly enabled only with `OpenMP` and doesn't work anyway right now).

crtrott:
@j8asic ah so the errors come when you are building the app?

core/src/impl/Kokkos_BitOps.hpp
```diff
@@ -82,13 +82,16 @@ inline int int_log2_device(unsigned i) {
 
 KOKKOS_IMPL_HOST_FUNCTION
 inline int int_log2_host(unsigned i) {
-  constexpr int shift = sizeof(unsigned) * CHAR_BIT - 1;
+// duplicating shift to avoid unused warning in else branch
```
PhilMiller, line 85: Could you stick a `FIXME` on this? Once we flip to C++17, we can mark it `[[maybe_unused]]`

crtrott: I really would like to avoid restarting the CI yet again, we need to move this to finally get the release process started.

masterleinad: I think we can replace pretty much all `(void) bla` with `[[maybe_unused]] bla` and don't need an extra comment here.

j8asic:
Yes any app. I am idiot in VS so maybe dont know how to properly setup win app depending on kokkos

core/src/Kokkos_HostSpace.hpp
```diff
@@ -261,13 +261,27 @@ class SharedAllocationRecord<Kokkos::HostSpace, void>
       ;
   SharedAllocationRecord() = default;
 
+  // This constructor does not forward to the one without exec_space arg
+  // in order to work around https://github.com/kokkos/kokkos/issues/5258
+  // This constructor is templated so I can't just put it into the cpp file
+  // like the other constructor.
```
PhilMiller, line 267: Could you describe what the actual failure was? It seems like there was a compiler bug in play of some sort, but that's not clear.

PhilMiller:
All of the messages here are about fixing build issues, but the linked bug and the code changes all seem to be addressing runtime failures in tests. Please document what's going on better, or maybe even split the PR between build issues and functional code changes

crtrott:
I put the primary build error in the issue.

core/src/Kokkos_HostSpace.hpp
```diff
@@ -261,13 +261,27 @@ class SharedAllocationRecord<Kokkos::HostSpace, void>
...
   template <typename ExecutionSpace>
   SharedAllocationRecord(
       const ExecutionSpace& /* exec_space*/, const Kokkos::HostSpace& arg_space,
       const std::string& arg_label, const size_t arg_alloc_size,
       const RecordBase::function_type arg_dealloc = &deallocate)
-      : SharedAllocationRecord(arg_space, arg_label, arg_alloc_size,
-                               arg_dealloc) {}
```
PhilMiller, line 270: Did you try this more explicit syntax to get the forwarding constructor?
```
 : SharedAllocationRecord<Kokkos::HostSpace, void>(arg_space, arg_label, arg_alloc_size,
                               arg_dealloc) {}
```

crtrott: yes

JBludau (approved):
looks good given the circumstances ...

j8asic:
About `VOID` error, I always get:

> C:\Code\kokkos\out\install\x64-Release\include\impl\Kokkos_FunctorAnalysis.hpp(111): error : expected either a definition or a tag name

5ch0:
@j8asic 
 CreateProcess failed. Command attempted:
"E:/kokkos-4.2.01/bin/kokkos_launch_compiler E:/kokkos-4.2.01/bin/nvcc_wrapper E:/Program Files/Microsoft Visual Studio/2022/Community/VC/Tools/MSVC/14.39.33519/bin/Hostx64/x64/cl.exe "E:\Program Files\Microsoft Visual Studio\2022\Community\VC\Tools\MSVC\14.39.33519\bin\Hostx64\x64\cl.exe"  /nologo /TP  -external:IE:\kokkos-4.2.01\build -external:IE:\kokkos-4.2.01\build\core\src -external:IE:\kokkos-4.2.01\core\src -external:IE:\kokkos-4.2.01\core\src\..\..\tpls\desul\include -external:IE:\kokkos-4.2.01\build\containers\src -external:IE:\kokkos-4.2.01\containers\src -external:IE:\kokkos-4.2.01\build\algorithms\src -external:IE:\kokkos-4.2.01\algorithms\src -external:IE:\kokkos-4.2.01\build\simd\src -external:IE:\kokkos-4.2.01\simd\src -external:W0 /DWIN32 /D_WINDOWS /W3 /GR /EHsc /MD /O2 /Ob2 /DNDEBUG -std:c++17 /showIncludes /Fo05\CMakeFiles\05.dir\05.cpp.obj /Fd05\CMakeFiles\05.dir\ /FS -c "E:\Parallel Project\05\05\05.cpp""
ninja: fatal: CreateProcess: %1 is not a valid Win32 application.
I had the same problem as you, have you solved it yet?The error comes when I am building the app.

PR #5265, opened 2022-07-25
HPX: Only use 1 thread in CI

masterleinad:
Trying to fix current build failures.


dalg24:
Thank you for looking into it

crtrott:
Trying two threads, to see if we can have less than 108 minutes build time ...

crtrott (approved):

dalg24 (approved):
Merging as a temporary solution but we should discuss moving that build to the ONRL Jenkins CI to speed things up.  It took 71min to build with 2 threads

Issue #5274, opened 2022-07-27
DynamicView create_mirror[_view[_and_copy]] missing call to resize_serial after construction


PhilMiller:
**Describe the bug**

The implementations of `create_mirror*` for `DynamicView` call the constructor overload that takes the various sizing parameters, but these set the maximum extent for the new instance, not the immediately allocated extent.

Subsequent `deep_copy` calls seem not to notice that that the destination extent is short, possibly because they use the `ViewRemap` call path that copies the minimum of the input and output extents.

**Please also include the following items to support reproducing the bug**
1. compilers (with versions): nvcc V10.1.243, GCC 7.2.0
2. Kokkos release or commit used (i.e., the sha1 number): 6b07e396330cae0220705cfc9538add7b329e3ba
3. platform and backend: Linux x86 with Cuda
4. cmake configure command: `cmake ~/repos/kokkos -DCMAKE_BUILD_TYPE=Debug -DKokkos_ENABLE_CUDA=TRUE `
6. code needed to reproduce the bug

```c++
// Instantiated with ViewT = DynamicView<int*> or whatever
template <typename ViewT, unsigned ndim>
static void compareInner1d(ViewT const& k1, ViewT const& k2) {
  std::cout << "compareInner1d: " << k1.extent(0) << "," << k2.extent(0) << "\n";
  EXPECT_EQ(k1.extent(0), k2.extent(0));

  auto j1 = Kokkos::create_mirror_view_and_copy(Kokkos::HostSpace{}, k1);
  auto j2 = Kokkos::create_mirror_view_and_copy(Kokkos::HostSpace{}, k2);

  for (typename ViewT::size_type i = 0; i < k1.extent(0); i++) {
    EXPECT_EQ(j1.operator()(i), j2.operator()(i));
  }
}
```

PhilMiller:
Cherry-picked fix on 3.7 RC branch is ready to merge

PR #5275, opened 2022-07-27
#5274 DynamicView: Properly resize mirror instances after construction

PhilMiller:
The constructor arguments set the new instance's capacity, but its size is initially small/0

### DIFF
--- a/containers/unit_tests/TestWithoutInitializing.hpp
+++ b/containers/unit_tests/TestWithoutInitializing.hpp
@@ -654,11 +654,16 @@ TEST(TEST_CATEGORY, create_mirror_view_and_copy_dynamicview) {
                                typename TEST_EXECSPACE::memory_space{}),
             mirror_device);
       },
+      [&](BeginFenceEvent event) {
+        if (event.descriptor().find("DynamicView::resize_serial: Fence after "
+                                    "copying chunks to the device") !=
+            std::string::npos)
+          return MatchDiagnostic{false};
+        return MatchDiagnostic{true, {"Found fence event"}};
+      },
+      [&](EndFenceEvent) { return MatchDiagnostic{false}; },
       [&](BeginParallelForEvent) {
         return MatchDiagnostic{true, {"Found parallel_for event"}};
-      },
-      [&](BeginFenceEvent) {
-        return MatchDiagnostic{true, {"Found fence event"}};
       });
   ASSERT_TRUE(success);
 }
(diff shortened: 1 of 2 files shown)


PhilMiller:
Discovered in the course of https://github.com/DARMA-tasking/magistrate/pull/261

crtrott (approved):

crtrott:
hm maybe there should be a constructor which makes the DynamicView start out with a size?

PhilMiller:
Might be useful, but I don't want to get into immediate design for the sake of a bug fix

PhilMiller:
Test failure:

```
53: [ RUN      ] cuda.create_mirror_view_and_copy_dynamicview
53: Test failure: encountered unwanted events
53:   Types didn't match on arguments
53:   Found fence event
53: BeginFence { "DynamicView::resize_serial: Fence after copying chunks to the device", 33554433,94}
53: EndFence { 94}
53: /var/jenkins/workspace/Kokkos/containers/unit_tests/TestWithoutInitializing.hpp:663: Failure
53: Value of: success
53:   Actual: false
53: Expected: true
53: [  FAILED  ] cuda.create_mirror_view_and_copy_dynamicview (0 ms)
```

I need to fix the test's expectations to match the sad reality

containers/unit_tests/TestWithoutInitializing.hpp
```diff
@@ -689,6 +694,14 @@ TEST(TEST_CATEGORY, create_mirror_no_init_dynamicview_view_ctor) {
                                Kokkos::DefaultExecutionSpace{}),
             host_view);
       },
+      [&](BeginFenceEvent event) {
+        if (event.descriptor().find("DynamicView::resize_serial: Fence after "
+                                    "copying chunks to the device") !=
+            std::string::npos)
+          return MatchDiagnostic{false};
```
masterleinad, line 701: We should probably introduce a version that accepts an execution space instance for the `deep_copy` later.

masterleinad (approved):

masterleinad:
Would you mind still adding a test to check that the size is as expected (which failed previously)?

dalg24:
> Would you mind still adding a test to check that the size is as expected (which failed previously)?

Sorry Daniel, I merged as you posted your comment

masterleinad:
> Sorry Daniel, I merged as you posted your comment

That's fine we can add in a follow-up.

PR #3809, opened 2021-02-24
CMake check for C++14

nliber:
Issue #3789 

This adds a CMake check for C++14 (for all compilers) by seeing if `std::remove_cv_t` compiles, as the `_t` versions of type traits were first added in C++14.

Issue #3789, opened 2021-02-10
Kokkos: add cmake check for gcc version now that c++14 is required


jhux2:
(Migrated from https://github.com/trilinos/Trilinos/issues/8720)

There are at least a couple issues (Trilinos  issues 8710 and 8673), as well as reports on the Kokkos slack channel, all related I think to the c++14 change. When someone uses the intel compiler, they need to make sure that a new enough GCC is being used.

Is it possible for Kokkos add a cmake check to make sure that a recent enough GCC is being used?

See the Trilinos issue for some ideas that @jjellio suggested.

DavidPoliakoff:
I'm a fan of the try_compile answer from @jjellio . This is the kind of thing we could provide real value with, if Kokkos is how users _avoid_ inscrutable C++ error messages.

DavidPoliakoff:
So, the error comes from building Kokkos_Core.cpp. Is there a way to:

1) try compiling Kokkos_Core.cpp
2) grep for failures containing `namespace "std" has no member   "remove_cv_t"`
3) If we spot them, call out "hey, your GCC is too old for Intel"

@jrmadsen, aren't you glad you showed initiative in build systems? You get to be a part of these fun conversations!

DavidPoliakoff:
Last comment, I swear: actually, come to think of it, this probably fixes a lot of other compilers, too

jjellio:
David, I would avoid telling them their GCC is too old. In my experience, this problem comes from a broken environment / module - the user thinks they are using Intel, and now you are talking about GNU. Moreover, just because `which g++` points to a valid version, their include or library path may be totally bogus. Just adding a newer GCC to their PATH or something may actually create more hellish problems.

I'd say, try to build something that is oddly specific to your std version (or a choice file from Kokkos), if that fails, print a detailed error message explaining what the problem likely is. My examples using `ldd` can be helpful, but they aren't fool proof - on Mutrino at one point this was broken because of how modules got loaded (and what they did in the module files). To that end, explaining that Intel uses GNU headers and thus needs to be matched to (ideally) the GNU used to install it would atleast be helpful. I don't think I ever opened an issue w/Trilinos about this, as it was resolved via Admin tickets on the machine - which ultimately the user is probably going to have to pursue (if their modules are whacky).

DavidPoliakoff:
[...]

PR #5277, opened 2022-07-27
Require C++17

masterleinad:
This still needs to update compiler requirements.


crtrott (changes requested):

dalg24 (changes requested):
Require C++17, update the CI build but do not start changing the code that you do not absolutely have to in this PR

crtrott (changes requested):
Looks pretty good. Lets decide what to do with the other standard options, and also change the CI to include minimum compilers, and do a manual check with other minimum compilers (e.g. Intel 19.1)

masterleinad:
> Require C++17, update the CI build but do not start changing the code that you do not absolutely have to in this PR

I honestly feel much better if we have the CI pass with all these changes (which for the most part, i.e., basically without the `STATIC_ASSERT` changes, basically only remove guards). I don't mind pushing these changes to a separate pull request in the end.

core/src/Kokkos_Core.hpp
```diff
@@ -102,8 +102,8 @@ void declare_configuration_metadata(const std::string& category,
 
 }  // namespace Impl
 
-KOKKOS_ATTRIBUTE_NODISCARD bool is_initialized() noexcept;
-KOKKOS_ATTRIBUTE_NODISCARD bool is_finalized() noexcept;
+[[nodiscard]] bool is_initialized() noexcept;
```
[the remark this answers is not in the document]
masterleinad: https://godbolt.org/z/5Pz8jWeqo shows that the minimal compiler versions (godbolt has) are `clang-5`, `MSVC 14`, `GCC 7.1`, and `ICC 18.0.0`.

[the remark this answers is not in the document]
crtrott: ok cool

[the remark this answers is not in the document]
masterleinad: Following https://en.cppreference.com/w/cpp/compiler_support/17, these compiler versions are good candidates anyway. Additionally, that would suggest ICC 19.0.0 (if constexpr), msvc 19.15 (std::is_aggregate), no changes for IntelLLVM, whatever HIPCC (we are already testing with C++17 support in one build with the minimum), nvcc 11.0, pgi 19.1.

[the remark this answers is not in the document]
PhilMiller: From my experience doing some of the downstream porting and testing work, you need 19.0.1, and I recommend Intel 19.0.5 or 19.1. The bare 19.0 is painfully buggy, and even 19.0.[1-4] are pretty bad. The Trilinos ATDM configuration I got added was for 19.0.5, so I don't think you'll get any complaints about picking that. 19.1 avoids some further bugs and stupid workarounds that shouldn't be necessary.

.jenkins
```diff
@@ -40,7 +40,7 @@ pipeline {
                         sh '''rm -rf build && mkdir -p build && cd build && \
                               /opt/cmake/bin/cmake \
                                 -DCMAKE_CXX_COMPILER=nvc++ \
-                                -DCMAKE_CXX_STANDARD=17 \
+                                -DCMAKE_CXX_STANDARD=20 \
```
ajpowelsnl, line 43: Do you mean to move to C++20 here, given that we will only move to C++17 in the (future) 4.0 release?

masterleinad: This isn't terribly important and we might just drop it for this pull request but I think bumping the requested standard in the CI by one `c++14->c++17->c++20` makes sense in the end.

core/src/Kokkos_Core.hpp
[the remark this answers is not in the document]
PhilMiller: Apparently, Intel "19.0.5" as reported on the command line `icpc --version` is actually 19.5 as reported by the compiler's macros. We'll need to see what CMake thinks of it

crtrott:
GCC 7.3
Clang 8.0
Intel 19.1
CUDA 11.0
ROCM 5.2

PhilMiller:
> GCC 7.3

There are a number of ICE cases that code generated by nvcc triggers in GCC 7.x for x<5. Since 7.3 was a deployed standard on a lot of lab systems, and 7.5 wasn't, we may want to require GCC 8.y or 9.z instead

PhilMiller:
There's also a noted correction for DRs regarding CTAD in GCC 8 per https://gcc.gnu.org/projects/cxx-status.html#cxx17

masterleinad:
`CUDA-11.7-NVCC` is still failing in `cuda.reducers_bhalf_t`. Otherwise, things are looking good in my opinion.
All that's left then is to finalize the minimum compiler version requirements.

masterleinad:
This still relies on https://github.com/kokkos/kokkos/pull/5286 for updating one CI build.

Makefile.kokkos
```diff
@@ -21,8 +21,8 @@ KOKKOS_ARCH ?= ""
 KOKKOS_DEBUG ?= "no"
 # Options: hwloc,librt,experimental_memkind
 KOKKOS_USE_TPLS ?= ""
-# Options: c++14,c++1y,c++17,c++1z,c++2a
-KOKKOS_CXX_STANDARD ?= "c++14"
+# Options: c++1z,c++2a
```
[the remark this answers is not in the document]
masterleinad: See #5283.

masterleinad:
Apart from updating compiler versions (which we likely could do separately), this should be ready now. I pushed all other cleanup that was initially here and that isn't required to #5295.

.jenkins
```diff
@@ -470,12 +470,11 @@ pipeline {
...
                               cmake \
                                 -DCMAKE_BUILD_TYPE=Release \
-                                -DCMAKE_CXX_STANDARD=14 \
+                                -DCMAKE_CXX_STANDARD=17 \
                                 -DCMAKE_CXX_FLAGS=-Werror \
                                 -DKokkos_ARCH_NATIVE=ON \
                                 -DKokkos_ENABLE_COMPILER_WARNINGS=ON \
-                                -DKokkos_ENABLE_DEPRECATED_CODE_3=ON \
-                                -DKokkos_ENABLE_DEPRECATION_WARNINGS=OFF \
+                                -DKokkos_ENABLE_DEPRECATED_CODE_3=OFF \
```
masterleinad, line 477: I encountered some problems in the `partition_master` test but I didn't care much since we are going to remove deprecated code anyway.

crtrott: partition_master will not be removed until we have a replacement, its a critical dependency for a a very small number of important customers.

masterleinad:
```
[ RUN      ] openmp.partition_master
  what():  Kokkos::Impl::HostThreadTeamData::organize_pool ERROR pool already exists
  what():  Kokkos::Impl::HostThreadTeamData::organize_pool ERROR pool already exists
Aborted
```

Makefile.kokkos
```diff
@@ -326,25 +314,18 @@ endif
...
   else
     ifeq ($(KOKKOS_INTERNAL_COMPILER_CRAY), 1)
-      KOKKOS_INTERNAL_CXX14_FLAG := -hstd=c++14
-      #KOKKOS_INTERNAL_CXX1Y_FLAG := -hstd=c++1y
-      #KOKKOS_INTERNAL_CXX17_FLAG := -hstd=c++17
-      #KOKKOS_INTERNAL_CXX1Z_FLAG := -hstd=c++1z
-      #KOKKOS_INTERNAL_CXX2A_FLAG := -hstd=c++2a
+      KOKKOS_INTERNAL_CXX17_FLAG := -hstd=c++17
+      KOKKOS_INTERNAL_CXX1Z_FLAG := -hstd=c++1z
+      KOKKOS_INTERNAL_CXX2A_FLAG := -hstd=c++2a
```
dalg24, line 327: Same question.  Did you mean to uncomment?

bin/nvcc_wrapper
```diff
@@ -314,8 +314,8 @@ do
     ;;
   # End of Werror handling
   #Handle unsupported standard flags
-  --std=c++1y|-std=c++1y|--std=gnu++1y|-std=gnu++1y|--std=c++1z|-std=c++1z|--std=gnu++1z|-std=gnu++1z|--std=c++2a|-std=c++2a)
-    fallback_std_flag="-std=c++14"
+  --std=gnu++1y|-std=gnu++1y|--std=c++1z|-std=c++1z|--std=gnu++1z|-std=gnu++1z|--std=c++2a|-std=c++2a)
```
dalg24, line 317: What did you change on this line and why?

.jenkins
masterleinad: I still think that this is not a reason to hold the pull request since there are no code changes. I opened https://github.com/kokkos/kokkos/issues/5298 for this.

Rombur: Try adding `OMP_NESTED=TRUE` to the environment variables for the build. Some versions of gcc requires different environment variables to enable nested parallelism.

masterleinad: Thanks, It's weird but it works.

Makefile.kokkos
```diff
@@ -326,25 +314,18 @@ endif
...
 else
   ifeq ($(KOKKOS_INTERNAL_COMPILER_XL), 1)
-     KOKKOS_INTERNAL_CXX14_FLAG := -std=c++14
-     KOKKOS_INTERNAL_CXX1Y_FLAG := -std=c++1y
-     #KOKKOS_INTERNAL_CXX17_FLAG := -std=c++17
-     #KOKKOS_INTERNAL_CXX1Z_FLAG := -std=c++1Z
-     #KOKKOS_INTERNAL_CXX2A_FLAG := -std=c++2a
+     KOKKOS_INTERNAL_CXX17_FLAG := -std=c++17
+     KOKKOS_INTERNAL_CXX1Z_FLAG := -std=c++1z
+     KOKKOS_INTERNAL_CXX2A_FLAG := -std=c++2a
```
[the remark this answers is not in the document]
masterleinad: I think we should discuss XL and PGI support separately anyway.

cmake/compile_tests/cplusplus17.cpp
```diff
@@ -0,0 +1,8 @@
+#include <type_traits>
+
+int main() {
+  // _v versions of type traits were added in C++17
+  constexpr bool same = std::is_same_v<double, int>;
```
PhilMiller, line 5: Is it salient that this is testing a standard library feature, rather than a compiler feature? Variable templates themselves were part of C++14, so a standard library implementation could provide an implementation for this that would compile under C++14.

masterleinad: Previously, we had `std::remove_cv_t<int> i = 0;` for C++14.
I am happy to hear other suggestions. What about `if constexpr`?

PhilMiller: As an edge example, libstdc++ as shipped with GCC 7.1 would pass this, though we're planning to set at least 7.3 as the minimum per https://github.com/kokkos/kokkos/issues/5285

PhilMiller: Given that `if constexpr` is probably one of the most desirable headline features for us, that seems quite suitable as a target.

containers/unit_tests/TestErrorReporter.hpp
```diff
@@ -195,6 +195,7 @@ struct ErrorReporterDriverUseLambda
   void execute(int reporter_capacity, int test_size) {
     Kokkos::parallel_for(
         Kokkos::RangePolicy<execution_space>(0, test_size),
+        // NOLINTNEXTLINE(kokkos-implicit-this-capture)
```
dalg24, line 198: Wait what?  Does that mean there is a bug in the clang-tidy check?

containers/unit_tests/TestErrorReporter.hpp
```diff
@@ -234,7 +235,9 @@ struct ErrorReporterDriverNativeOpenMP
 };
 #endif
 
-#if defined(KOKKOS_CLASS_LAMBDA) && \
+// FIXME_MSVC MSVC just gets confused when using the base class in the
+// KOKKOS_CLASS_LAMBDA
+#if !defined(KOKKOS_COMPILER_MSVC) && defined(KOKKOS_CLASS_LAMBDA) && \
```
dalg24, line 240: Isn't `KOKKOS_CLASS_LAMDA` always defined now?

dalg24 (commented):
Looks mostly ok
I'd like for someone else to look over the `nvcc_wrapper` changes

containers/unit_tests/TestErrorReporter.hpp
masterleinad: Yes, see https://kokkosteam.slack.com/archives/G5CBLMFLP/p1659012380757519.

containers/unit_tests/TestErrorReporter.hpp
masterleinad: Yes, it is, but I avoided doing all the cleanup in this pull request.

masterleinad:
> I'd like for someone else to look over the `nvcc_wrapper` changes

I'm also fine splitting the `nvcc_wrapper` changes from this pull request if we have doubts about dropping C++14 and earlier support from it.

cmake/compile_tests/cplusplus17.cpp
masterleinad: I added `if constexpr` but also kept `std::is_same_v` since the test case was initially designed to detect a working toolchain (with recent enough gcc headers), see https://github.com/kokkos/kokkos/pull/3809 and https://github.com/kokkos/kokkos/issues/3789.

crtrott (changes requested):
Lets leave nvcc_wrapper more or less alone. I don't think we need to make it not deal with cpp14. In particular since some folks use it independent of Kokkos.

masterleinad:
> Lets leave nvcc_wrapper more or less alone. I don't think we need to make it not deal with cpp14. In particular since some folks use it independent of Kokkos.

Sure. I reverted those changes.

dalg24 (approved):

PhilMiller (approved):

masterleinad:
SYCL CI timing out is unrelated. That build required C++17 already.

crtrott (approved):

PR #5281, opened 2022-07-27
Check size in Containers WithoutInitializing test

masterleinad:
Follow-up to https://github.com/kokkos/kokkos/pull/5275#issuecomment-1196677436.

### DIFF
--- a/containers/unit_tests/TestWithoutInitializing.hpp
+++ b/containers/unit_tests/TestWithoutInitializing.hpp
@@ -420,12 +420,16 @@ TEST(TEST_CATEGORY, create_mirror_no_init_dynrankview) {
       [&]() {
         auto mirror_device =
             Kokkos::create_mirror(Kokkos::WithoutInitializing, device_view);
+        ASSERT_EQ(device_view.size(), mirror_device.size());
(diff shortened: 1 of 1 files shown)


PhilMiller:
It feels weird to be comparing `size()` rather than `extent()`, given our semantics. I agree that this is sensitive to the earlier defects we just addressed, though.

ajpowelsnl:
@masterleinad , @PhilMiller - In all of the new testing statements, you assert expectation of equality (amongst Views, mirrors, etc.).  Would it be important to test for conditions of expected non-equality?

ajpowelsnl (approved):

masterleinad:
> @masterleinad , @PhilMiller - In all of the new testing statements, you assert expectation of equality (amongst Views, mirrors, etc.). Would it be important to test for conditions of expected non-equality?

@ajpowelsnl What do you mean? Whenever we create a mirror, we expect the extents to be equal to the ones of the original object, right?

masterleinad:
In one `Cuda` CI test we are running out of disk space and `KokkosCore_IncrementalTest_OPENACC` is failing both should be safe to ignore.

ldh4 (approved):

PhilMiller (approved):

PR #5286, opened 2022-07-28
Support finding libquadmath with native compiler support

masterleinad:
`gcc` oftentimes bundles `libquadmath` and doesn't need any extra libraries or include directories (which would not be contained in `CMake`'s search paths anyway).
Hence, try if the compiler supports `libquadmath` natively first before doing the standard package search.

`KOKKOS_IMPORT_TPL` needed a workaround to be able to create a consistent target whether `KOKKOS_QUADMATH_COMPILER_SUPPORT` is true or not.

This should in particular help with testing `libquadmath` support.


cmake/Modules/FindTPLLIBQUADMATH.cmake
```diff
@@ -1 +1,17 @@
-KOKKOS_FIND_IMPORTED(LIBQUADMATH HEADER quadmath.h LIBRARY quadmath)
+# gcc bundles libquadmath and doesn't need any extra libraries or include
```
dalg24, line 1:
```suggestion
# gcc bundles libquadmath and doesn't need any extra link or include
```
to match better with the cmake vocabulary (`target_{link, include}_directories`).
I got confused with the initial wording and was wondering why you were explicitly linking against libquadmath below even though the comment said we don't need any extra librar[y]ies.

dalg24 (dismissed):
Please apply my suggestion
Looks good other than that

cmake/Modules/FindTPLLIBQUADMATH.cmake
masterleinad: [`f1eae63` (#5286)](https://github.com/kokkos/kokkos/pull/5286/commits/f1eae639c3efacf4b6837111fcd5eca38da18f97)

PhilMiller:
This seems like a good candidate for 3.7

masterleinad:
> Unrelated changes pushed post approval

They are not unrelated but make sure the detection also works with `-Werror` and without extra include path.

masterleinad:
> This seems like a good candidate for 3.7

I think this is not a bug fix so I would abstain from pulling more stuff in.

PhilMiller:
I was looking at it as something that would ease our support burden from users who have trouble with this aspect of the build/installation process

masterleinad:
There are a bunch of CI failures (mostly timing based but OpenMPTarget can't find LIBDL???) but this pull request only affects the `gcc-5.3.1` build which works with
```
-- Performing Test KOKKOS_QUADMATH_COMPILER_SUPPORT
-- Performing Test KOKKOS_QUADMATH_COMPILER_SUPPORT - Success
```

scripts/docker/Dockerfile.openmptarget
```diff
@@ -54,7 +54,7 @@ RUN LLVM_URL=https://github.com/llvm/llvm-project/archive &&\
       -DCMAKE_C_COMPILER=gcc \
       -DCMAKE_CXX_COMPILER=g++ \
       -DLLVM_ENABLE_PROJECTS="clang" \
-      -DLLVM_ENABLE_RUNTIMES="openmp" \
+      -DLLVM_ENABLE_RUNTIMES="openmp;libcxx;libcxxabi;libunwind" \
```
dalg24, line 57: Revert or describe why that would be needed.

dalg24 (changes requested):

scripts/docker/Dockerfile.openmptarget
masterleinad: Dropped, not quite sure how that got there.

cmake/kokkos_functions.cmake
```diff
@@ -281,7 +281,11 @@ MACRO(kokkos_import_tpl NAME)
       MESSAGE(FATAL_ERROR "Find module succeeded for ${NAME}, but did not produce valid target ${TPL_IMPORTED_NAME}")
     ENDIF()
     IF(NOT TPL_NO_EXPORT)
-      KOKKOS_EXPORT_IMPORTED_TPL(${TPL_IMPORTED_NAME})
+      GET_TARGET_PROPERTY(TPL_ORIGINAL_NAME ${TPL_IMPORTED_NAME} ALIASED_TARGET)
+      IF (NOT TPL_ORIGINAL_NAME)
+        SET(TPL_ORIGINAL_NAME ${TPL_IMPORTED_NAME})
+      ENDIF()
+      KOKKOS_EXPORT_IMPORTED_TPL(${TPL_ORIGINAL_NAME})
```
dalg24, line 288: Why is this needed?  Make sure this is in the description of the PR

masterleinad: I added a comment. Without this, we are getting
```
-- Performing Test KOKKOS_QUADMATH_COMPILER_SUPPORT
-- Performing Test KOKKOS_QUADMATH_COMPILER_SUPPORT - Success
CMake Error at cmake/kokkos_functions.cmake:187 (INSTALL):
  INSTALL TARGETS given target "Kokkos::LIBQUADMATH" which is an alias.
  cmake/kokkos_functions.cmake:288 (KOKKOS_EXPORT_IMPORTED_TPL)
  cmake/kokkos_tpls.cmake:86 (KOKKOS_IMPORT_TPL)
  cmake/kokkos_tribits.cmake:247 (INCLUDE)
  CMakeLists.txt:188 (KOKKOS_SETUP_BUILD_ENVIRONMENT)
```

dalg24 (approved):

crtrott (approved):

Issue #5193, opened 2022-07-11
Do we need a HIP_ENABLE_MANAGED/SYCL_ENABLE_SHARED flag like we have for cuda


JBludau:
This is to discuss (and document) if we want to add a flag to cmake that would enable all allocations into HIP or SYCL be done in the HIPManaged/SYCLShared space as we do with the CudaUVM.

I am leaning towards no, but as there is/was a reason to provide this for cuda, we could reevaluate if this still is a thing.

PhilMiller:
As I understand it, the CUDA flag is there for backwards compatibility, since Kokkos used to allocate UVM by default, and then moved away from that because the most performance sensitive client codes all recognized that they needed to avoid implicit data movement.

On the other hand, for less ambitious applications, automatic allocation in shared memory is a huge convenience. I think the way for them to do that going forward should be use of a suitable type alias or template explicit in their code, or even just writing a memory space into every View declaration. Kokkos could potentially provide some portable expression for that, like SharedMemView, that knows the appropriate space to use for the active backend.

JBludau:
Also for porting stuff to Kokkos a shared space could be a drop-in for things that are required for the code to run but are not yet being redesigned to deal with heterogeneous memory. As not all backends do (currently) support a page migrating memory, providing a portable expression could be incomplete.

masterleinad:
I would be in favor of adding an alias that can be used for backends such as `DefaultSharedMemorySpace` as also suggested by @crtrott. I think we should get rid of a configure-time option to make the default memory space shared since bugs are harder to find.

JBludau:
I favor an alias, I think this would be a nice convenience as they are named differently over the backends.
Other than that, I kind of see that users do not have to explicitly specify a MemorySpace as only benefit of a compile time option. For porting users could just use the alias in all places where they are not sure yet which MemorySpace it be in the end. So I agree with @masterleinad and @dalg24 that we should ditch the configure-time option, if there are no compelling reasons to keep it.

If we provide an alias, I think the name should be more specific, as e.g. host pinned/page locked memory is also device accessible and there is even cpu accessible gpu memory. I suggest, that the memory this alias would point to would logically be a `(page)migrating` memory ... thus we could name it `DefaultMigratingMemorySpace`. Shared could be misunderstood as shared between threads.

JBludau:
I think the responses we got here are enough to draft something and propose a pr. This will reignite the discussions anyway and with having a code to look at it becomes less abstract

[...]

Issue #157, opened 2015-12-10
fatal error: Cuda/Kokkos_Cuda_abort.hpp: No such file or directory


ambrad:
I may be building the standalone Kokkos library incorrectly; apologies if this is user error. In any case, when I don't build --with-cuda, in my app I get this compiler error:

/include/impl/Kokkos_Error.hpp:49:38: fatal error: Cuda/Kokkos_Cuda_abort.hpp: No such file or directory

An example Kokkos build that gives this error is this one:

./generate_makefile.bash --with-openmp --with-serial --compiler=g++-4.7 --with-options=aggressive_vectorization --prefix=/home/ambradl/lib/kokkos/cpu

Thanks.

crtrott:
Ah I see the issue. It was introduced on 11/12, and is caused by a missing guard. That doesn't hurt in the case of an embedded build (where all the header files, including unused execution spaces are always available), but it doesn't work for the actual TPL install, since that only installs files for the existing execution spaces. I'll fix it tomorrow and get it into the develop branch.

crtrott:
I pushed a fix to develop. Please let me know if it works for you (if you can work of the develop branch from kokkos). It will take a couple days to get it to Trilinos, since I have to run an integration cycle.

ambrad:
Yes, it works. Thanks.

crtrott:
Reopening until we got it in master.

crtrott:
And thanks for confirming so fast :+1:

crtrott:
Pushed to master

Issue #149, opened 2015-11-30
Check for oversubscription of threads with MPI in Kokkos::initialize


mhoemmen:
Many Trilinos developers have been reporting problems with unit tests timing out or taking a long time.  The cause is oversubscription of the node with too many threads.  Kokkos::initialize can intervene and prevent this by checking the following:

1. Number of cores per node
2. Number of threads that the user requested
3. Number of MPI processes per node

Kokkos already knows (2).  For MVAPICH2 and OpenMPI, it can check environment variables to get (3).  (For example, MV2_COMM_WORLD_LOCAL_SIZE or OMPI_COMM_WORLD_LOCAL_SIZE.)  In fact, it already checks similar environment variables.  This does not require a compile-time dependency on MPI and it works whether or not MPI is enabled.

For (1), the number of cores per node, we will need to research a bit.

nmhamster:
Mark,

I think the best way to solve this is to use a KokkosP connector which at initialization time checks the number of threads requested against the associated process CPU mask. Much of the support for this is already in place including multi-device. I think a lot of debugging can be done without needing to put code directly into the Kokkos runtime itself. This will help maintenance since we can add to the tools *after* a program has been compiled (without needing updates of Kokkos in production) and catch additional errors and reduce performance overheads associated with the checks when running in "real" production.

If the users are using MPI/job subsmission correctly the CPU mask will be generated to ensure that over-subscription does not occur. In OpenMPI and MPICH a user has to actively oversubscribe or the MPI runtime gets upset so it shouldn't happen by accident. Similarly, using SLURM etc from a scheduler perspective will also set masks.

This means we could just perform the check locally within a KokkosP connector and avoid MPI communications. It also gets the problem not just of oversubscribed nodes but incorrectly set affinities on jobs which need to/want to use less than a complete node (for instance, using one thread per core for a parallel run of STREAM gives best performance but would not lead to over-subscription).

Does you think that might work for you?

S.

mhoemmen:
Hi Si!  Christian was thinking of the environment variable approach as a low-overhead fix.  Kokkos already checks environment variables like those so it would be pretty easy to add.  I'm OK with the KokkosP approach as long as it is enabled by default.  The issue is that by default, Trilinos _will_ oversubscribe.  It's easy to fix that (set OMP_NUM_THREADS environment variable manually), but users don't know that, and you know how hard it is to educate users ;-)  Also, the Pthreads back-end won't respect that environment variable.

[...]

PR #5289, opened 2022-07-28
Adding an alias for a page migrating memory space

JBludau:
This pr picks up on the discussion in #5193 and provides an alias for a page migrating memory space.

Even though the participants in the discussion preferred the name `DefaultSharedMemorySpace` I propose `MigratingMemorySpace` but have no problems changing that name. The reasoning was the following:

- Dropping the `Default` as it is to close to the execution space aliases we have (`DefaultExecutionSpace` and `DefaultHostExecutionSpace`). I wanted to prevent it being mistaken as an execution space, thus it follows the convention we have for the (memory)spaces. Furthermore, if we stick with it being migrating the `Default` looses its meaning as there is only one per backend.
- I propose `Migrating`, as it targets developers who actually want the property of a memory that automatically moves to the device and is accessed locally. I expect they want the same behavior independent of the backend and deliberately choose this (especially given that not all backends support this feature). If they switch to another backend that has no page migration, this should not silently chance behavior.
-  An alias for a memoryspace that is always accessible by host and device and available in (almost) all backends is useful. I propose to create a separate alias for this, maybe `UniversalMemorySpace`. I think this would need a good documentation on what the limitations and restrictions are, but it will be clearer what the cost for the universal accessibility is.
- We can actually specify and thus test what we expect from this memory and track if we do any changes that would change behavior.

If we introduce this alias, we should reconsider removing `Kokkos_ENABLE_CUDA_UVM` as there is now a better way to specify that the user wants page migration and we are moving to a major release which includes `HIP` moving out of `Experimental`.

The specification we are testing is:

- Migrate on fist touch in new execution space
- Migrate only if switching to different execution space

Missing:

- [x] write documentation: [#157](https://github.com/kokkos/kokkos-core-wiki/pull/157)


JBludau:
Quick note on the amount of memory we allocate and the number of repetitions the test is running:
It turned out, that Nvidia's V100 had no problem cycling 10 times to 40% of its memory without being at 100% of the core clock rate. 

MI100 (and below) do not support page migration, thus the test is disabled for these archs. 

For SYCL: I need to read more documentation on getting device attributes and looking which devices do support page migration.

masterleinad (changes requested):
Add support for SYCL using `Kokkos::Experimental::SYCLSharedUSMSpace`.

JBludau:
> Add support for SYCL using `Kokkos::Experimental::SYCLSharedUSMSpace`.

added. According to the [doc](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_kinds_of_unified_shared_memory) the initial placement is not specified.
I will look at cuda and hip, especially what they do if we overcommit on migratingMemory.
Might make sense for us to not expect anything about the intital placement

core/unit_test/CMakeLists.txt
```diff
@@ -329,6 +329,30 @@ foreach(PairDeviceSpace HIP-HostPinned;HIP-Managed;Cuda-HostPinned;Cuda-UVM;SYCL
...
+    set(file ${dir}/Test${DEVICE}${SPACE}_${Name}.cpp)
+    # Write to a temporary intermediate file and call configure_file to avoid
+    # updating timestamps triggering unnecessary rebuilds on subsequent cmake runs.
+    file(WRITE ${dir}/dummy.cpp
+        "#include <Test${DEVICE}${SPACE}_Category.hpp>\n"
+        "#include <Test${Name}.hpp>\n"
+    )
+    configure_file(${dir}/dummy.cpp ${file})
+    list(APPEND ${DEVICE}_SOURCES4 ${file})
+    list(APPEND ${DEVICE}_SOURCES ${${DEVICE}_SOURCES4})
```
masterleinad, line 352: `SYCL` doesn't use `${DEVICE}_SOURCES` (since running the device compiler and linking all tests at once is a nightmare) so you would need to define an additional test target for `SYCL` that uses `${DEVICE}_SOURCES4`.

core/unit_test/TestPageMigration.hpp
```diff
@@ -0,0 +1,242 @@
...
+
+  Kokkos::fence();
+  for (unsigned i = 0; i < numRepetitions; ++i) {
+    timer.reset();
+    Kokkos::parallel_for(
+        "increment",
+        Kokkos::RangePolicy<ExecSpace,
+                            Kokkos::IndexType<decltype(view.size())>>{
+            0, view.size()},
+        KOKKOS_LAMBDA(auto i) { ++view(i); });
```
masterleinad, line 92: Does `nvcc` really allow using generic `__host__ __device__` lambdas? I wouldn't be surprised if using `i` here raises a shadowing warning on some compilers.

core/unit_test/TestPageMigration.hpp
```diff
@@ -0,0 +1,242 @@
...
+}
+
+size_t getDeviceMemorySize() {
+#if defined KOKKOS_ENABLE_CUDA
+  return Kokkos::Cuda{}.cuda_device_prop().totalGlobalMem;
+#elif defined KOKKOS_ENABLE_HIP
+  return Kokkos::Experimental::HIP::hip_device_prop().totalGlobalMem;
+#elif defined KOKKOS_ENABLE_SYCL
+  return 1024 * 1024 * 1024 *
+         1024;  // 1gb as I can not access sycl::info::device via Kokkos
```
masterleinad, line 106: `Kokkos::Experimental::SYCL{}.sycl_queue()` gives use the `sycl::queue` member which has a `get_device()` member function returning a `sycl::device` object that has a `get_info()` member function that allows for querying `info::device::
max_mem_alloc_size` or `info::device::global_mem_size`.

masterleinad:
Regarding the name, I'm not sure if people would find it intuitive to understand that this is a space they can access from the host and the device (which is what people would mainly be interested IMHO).

core/unit_test/TestPageMigration.hpp
JBludau: you are right -> changed in [311f08](https://github.com/kokkos/kokkos/pull/5289/commits/311f0811365c3875bcd1bc30241c8536fbcd3f22)

core/unit_test/TestPageMigration.hpp
JBludau: done in [2f1bf3](https://github.com/kokkos/kokkos/pull/5289/commits/2f1bf31480c12c8df25ddbf3405bbcb016fd1b10)

core/unit_test/CMakeLists.txt
JBludau: done in [8ffb58](https://github.com/kokkos/kokkos/pull/5289/commits/8ffb58b4aaf156f79b08cfed392cf3387a210145)

JBludau:
looks like our `rocm` CI is building for `Kokkos_ARCH_VEGA90A` which it does not have hardware for. As the tests need real page migration support to pass, they are failing on our CI. Maybe we should add a label and specify the architecture. -> will do this in another pr

JBludau:
> Regarding the name, I'm not sure if people would find it intuitive to understand that this is a space they can access from the host and the device (which is what people would mainly be interested IMHO).

But this space is far more than just accessible from both. It actively moves the memory and allows local access afterwards. We can just add another alias for space that is accessible everywhere and does not migrate (As I proposed in the description)

masterleinad:
On `Intel GPUs`, I see no migration overhead for `host->device` but only for `device->host`, i.e., the loop time is constant on the device and the first access on the host is slower than the remaining ones. Also, the first access in the first host loop is slower than the first access in the later host loops. Thus, the test currently fails with
```
Initial placement on device is 1 we expect true 
Memory migrates on every space access is 0 we expect true 
Memory migrates only once per access 0 we expect true 
```
and I need
```diff
diff --git a/core/unit_test/TestPageMigration.hpp b/core/unit_test/TestPageMigration.hpp
index 9b6e1a051..ab4e0a494 100644
--- a/core/unit_test/TestPageMigration.hpp
+++ b/core/unit_test/TestPageMigration.hpp
@@ -122,7 +122,7 @@ TEST(TEST_CATEGORY, page_migration) {
   const unsigned int numWarmupRepetitions = 100;
   const unsigned int numDeviceHostCycles  = 3;
   double fractionOfDeviceMemory           = 0.4;
-  double threshold                        = 2.0;
+  double threshold                        = 1.5;
   size_t numBytes       = fractionOfDeviceMemory * getDeviceMemorySize();
   unsigned int numPages = numBytes / getBytesPerPage();
 
@@ -190,7 +190,7 @@ TEST(TEST_CATEGORY, page_migration) {
     if (cycle == 0 && indicatedPageMigrationsDevice == 0)
       initialPlacementOnDevice = true;
     else {
-      if (indicatedPageMigrationsDevice != 1) migratesOnlyOncePerAccess = false;
+      if (indicatedPageMigrationsDevice > 1) migratesOnlyOncePerAccess = false;
     }
 
     unsigned int indicatedPageMigrationsHost = std::count_if(
```
for it to pass.

dalg24:
Please justify that the unit test added here cannot ran faster and achieve the same thing.  (I would probably already have complained for more than a few seconds.)
```
test 7
      Start  7: KokkosCore_UnitTest_Cuda4

7: Test command: /var/jenkins/workspace/Kokkos/build/core/unit_test/KokkosCore_UnitTest_Cuda4
7: Test timeout computed to be: 1500
7: [==========] Running 1 test from 1 test suite.
7: [----------] Global test environment set-up.
7: [----------] 1 test from cuda_uvm
7: [ RUN      ] cuda_uvm.page_migration
7: [       OK ] cuda_uvm.page_migration (38209 ms)
7: [----------] 1 test from cuda_uvm (38209 ms total)
7: 
7: [==========] 1 test from 1 test suite ran. (38209 ms total)
7: [  PASSED  ] 1 test.
 7/61 Test  #7: KokkosCore_UnitTest_Cuda4 ....................   Passed   39.66 sec

**Comment by JBludau** [comment]
just some name ideas for the `temporally local` and `fixed but universal`: (plz don't hate me for this)
`UniversalLocalMemorySpace` and `UniversalFixedMemorySpace` 
`GlobalMovingMemorySpace` and `GlobalPinnedMemorySpace` 
`MoveThenLocalMemorySpace` and `FixedButGlobalMemorySpace`
`SharedMovingMemorySpace` and `SharedFixedMemorySpace`

**Comment by masterleinad** [comment]
I like `Universal` but would try to avoid `Local` and `Global` so maybe `UniversalMovingMemorySpace` and `UniversalPinnedMemorySpace`?

**Comment by PhilMiller** [comment]
UniversalMoving sounds fine to me.

UniversalPinned is more troublesome, because either or both of host or device pinned could exist, and with different properties on both performance and non-compute accessibility

**Comment by crtrott** [comment]
I know its somewhat confusing but how about: `SharedSpace`, and `SharedPinnedSpace`. I know CUDA shared memory is something totally different, however ironically CUDA shared memory is the least shared of all allocations possible in CUDA except for registers ...

**Comment by JBludau** [comment]
> I know its somewhat confusing but how about: `SharedSpace`, and `SharedHostPinnedSpace`. I know CUDA shared memory is something totally different, however ironically CUDA shared memory is the least shared of all allocations possible in CUDA except for registers ...

Damien will fight you on this :-)

**Comment by JBludau** [comment]
The unit test is now based on clock cycles rather than on wall clock time which eliminated the need for warmup runs and thus large memory chunks on AMD and Nvidia GPUs. @masterleinad could you rerun this for intel gpus?

Furthermore, I adapted the name @crtrott suggested but we can still change it. @crtrott could you sum up your reasoning for this here to have it documented?

**Comment by JBludau** [comment]
> Please justify that the unit test added here cannot ran faster and achieve the same thing. (I would probably already have complained for more than a few seconds.)
> 
> ```
> test 7
>       Start  7: KokkosCore_UnitTest_Cuda4
> 
> 7: Test command: /var/jenkins/workspace/Kokkos/build/core/unit_test/KokkosCore_UnitTest_Cuda4
> 7: Test timeout computed to be: 1500
> 7: [==========] Running 1 test from 1 test suite.
> 7: [----------] Global test environment set-up.
> 7: [----------] 1 test from cuda_uvm
> 7: [ RUN      ] cuda_uvm.page_migration
> 7: [       OK ] cuda_uvm.page_migration (38209 ms)
> 7: [----------] 1 test from cuda_uvm (38209 ms total)
> 7: 
> 7: [----------] Global test environment tear-down
> 7: [==========] 1 test from 1 test suite ran. (38209 ms total)
> 7: [  PASSED  ] 1 test.
>  7/61 Test  #7: KokkosCore_UnitTest_Cuda4 ....................   Passed   39.66 sec
> ```

should be in the order of `ms` now

**Comment by masterleinad** [comment]
> The unit test is now based on clock cycles rather than on wall clock time which eliminated the need for warmup runs and thus large memory chunks on AMD and Nvidia GPUs. @masterleinad could you rerun this for intel gpus?

No, it doesn't pass on Intel GPUs and the numbers for device access are much higher than for host access even though the runtime before showed the opposite. Note that this calls different functions on the host and on the device and it's not implemented properly for SYCL+CUDA on the device. We use the function as seed and thus returning 0 is good enough for that but not for the purpose here. CI for SYCL+Cuda shows
```
[ RUN      ] sycl_shared_usm.page_migration
14: Page size as reported by os: 4096 bytes 
14: Allocating 100 pages of memory in pageMigratingMemorySpace.
14: Behavior found: 
14: Initial placement on device is: 1 we expect true 
14: Memory migrates back to GPU is: 0 we expect true 
14: Memory migrates at max once per access: 1 we expect true 
14: 
14: Please look at the following timings. A migration was marked detected if the time was larger than 0 for the device 
14: 
14: device timings of run 0:
14: TimingResult contains 10 results:
14: Duration of loop 0 is 0 clock cycles
14: Duration of loop 1 is 0 clock cycles
14: Duration of loop 2 is 0 clock cycles
14: Duration of loop 3 is 0 clock cycles
14: Duration of loop 4 is 0 clock cycles
14: Duration of loop 5 is 0 clock cycles
14: Duration of loop 6 is 0 clock cycles
14: Duration of loop 7 is 0 clock cycles
14: Duration of loop 8 is 0 clock cycles
14: Duration of loop 9 is 0 clock cycles
14: host timings of run 0:
14: TimingResult contains 10 results:
14: Duration of loop 0 is 20 clock cycles
14: Duration of loop 1 is 15 clock cycles
14: Duration of loop 2 is 15 clock cycles
14: Duration of loop 3 is 15 clock cycles
14: Duration of loop 4 is 15 clock cycles
14: Duration of loop 5 is 15 clock cycles
14: Duration of loop 6 is 14 clock cycles
14: Duration of loop 7 is 15 clock cycles
14: Duration of loop 8 is 14 clock cycles
14: Duration of loop 9 is 14 clock cycles
14: device timings of run 1:
14: TimingResult contains 10 results:
14: Duration of loop 0 is 0 clock cycles
14: Duration of loop 1 is 0 clock cycles
14: Duration of loop 2 is 0 clock cycles
14: Duration of loop 3 is 0 clock cycles
14: Duration of loop 4 is 0 clock cycles
14: Duration of loop 5 is 0 clock cycles
14: Duration of loop 6 is 0 clock cycles
14: Duration of loop 7 is 0 clock cycles
14: Duration of loop 8 is 0 clock cycles
14: Duration of loop 9 is 0 clock cycles
14: host timings of run 1:
14: TimingResult contains 10 results:
14: Duration of loop 0 is 18 clock cycles
14: Duration of loop 1 is 15 clock cycles
14: Duration of loop 2 is 14 clock cycles
14: Duration of loop 3 is 14 clock cycles
14: Duration of loop 4 is 16 clock cycles
14: Duration of loop 5 is 15 clock cycles
14: Duration of loop 6 is 15 clock cycles
14: Duration of loop 7 is 16 clock cycles
14: Duration of loop 8 is 14 clock cycles
14: Duration of loop 9 is 15 clock cycles
14: device timings of run 2:
14: TimingResult contains 10 results:
14: Duration of loop 0 is 0 clock cycles
14: Duration of loop 1 is 0 clock cycles
14: Duration of loop 2 is 0 clock cycles
14: Duration of loop 3 is 0 clock cycles
14: Duration of loop 4 is 0 clock cycles
14: Duration of loop 5 is 0 clock cycles
14: Duration of loop 6 is 0 clock cycles
14: Duration of loop 7 is 0 clock cycles
14: Duration of loop 8 is 0 clock cycles
14: Duration of loop 9 is 0 clock cycles
14: host timings of run 2:
14: TimingResult contains 10 results:
14: Duration of loop 0 is 18 clock cycles
14: Duration of loop 1 is 15 clock cycles
14: Duration of loop 2 is 15 clock cycles
14: Duration of loop 3 is 15 clock cycles
14: Duration of loop 4 is 16 clock cycles
14: Duration of loop 5 is 15 clock cycles
14: Duration of loop 6 is 15 clock cycles
14: Duration of loop 7 is 15 clock cycles
14: Duration of loop 8 is 15 clock cycles
14: Duration of loop 9 is 15 clock cycles
14: /var/jenkins/workspace/Kokkos/core/unit_test/TestPageMigration.hpp:206: Failure
14: Value of: passed
14:   Actual: false
14: Expected: true
14: [  FAILED  ] sycl_shared_usm.page_migration (57 ms)
```

All that is to say that a timing-based test would be better for testing the `SYCL` implementation.

File: core/unit_test/TestPageMigration.hpp [hunk 4]
```diff
@@ -0,0 +1,209 @@
...
+      std::cout << "device timings of run " << cycle << ":\n";
+      printTimings(std::cout, deviceResults[cycle]);
+      std::cout << "host timings of run " << cycle << ":\n";
+      printTimings(std::cout, hostResults[cycle]);
+    }
+  }
+  ASSERT_TRUE(passed);
+}
+#endif
+}  // namespace
```
masterleinad, line 209:
```suggestion
}  // namespace
#endif
```

masterleinad:
Also, the test fails for HIP in the CI with something like
```
4: [ RUN      ] hip_managed.page_migration
4: Page size as reported by os: 4096 bytes 
4: Allocating 100 pages of memory in pageMigratingMemorySpace.
4: Behavior found: 
4: Initial placement on device is: 0 we expect true 
4: Memory migrates back to GPU is: 0 we expect true 
4: Memory migrates at max once per access: 0 we expect true 
4: 
4: Please look at the following timings. A migration was marked detected if the time was larger than 13600 for the device 
4: 
4: device timings of run 0:
4: TimingResult contains 10 results:
4: Duration of loop 0 is 28412 clock cycles
4: Duration of loop 1 is 22968 clock cycles
4: Duration of loop 2 is 23048 clock cycles
4: Duration of loop 3 is 22856 clock cycles
4: Duration of loop 4 is 22896 clock cycles
4: Duration of loop 5 is 23355 clock cycles
4: Duration of loop 6 is 23344 clock cycles
4: Duration of loop 7 is 21965 clock cycles
4: Duration of loop 8 is 21477 clock cycles
4: Duration of loop 9 is 24143 clock cycles
4: host timings of run 0:
4: TimingResult contains 10 results:
4: Duration of loop 0 is 33 clock cycles
4: Duration of loop 1 is 33 clock cycles
4: Duration of loop 2 is 33 clock cycles
4: Duration of loop 3 is 33 clock cycles
4: Duration of loop 4 is 33 clock cycles
4: Duration of loop 5 is 32 clock cycles
4: Duration of loop 6 is 33 clock cycles
4: Duration of loop 7 is 33 clock cycles
4: Duration of loop 8 is 33 clock cycles
4: Duration of loop 9 is 32 clock cycles
4: device timings of run 1:
4: TimingResult contains 10 results:
4: Duration of loop 0 is 22179 clock cycles
4: Duration of loop 1 is 22560 clock cycles
4: Duration of loop 2 is 22308 clock cycles
4: Duration of loop 3 is 21959 clock cycles
4: Duration of loop 4 is 22544 clock cycles
4: Duration of loop 5 is 23154 clock cycles
4: Duration of loop 6 is 22839 clock cycles
4: Duration of loop 7 is 21566 clock cycles
4: Duration of loop 8 is 22493 clock cycles
4: Duration of loop 9 is 22335 clock cycles
4: host timings of run 1:
4: TimingResult contains 10 results:
4: Duration of loop 0 is 32 clock cycles
4: Duration of loop 1 is 33 clock cycles
4: Duration of loop 2 is 33 clock cycles
4: Duration of loop 3 is 33 clock cycles
4: Duration of loop 4 is 33 clock cycles
4: Duration of loop 5 is 33 clock cycles
4: Duration of loop 6 is 33 clock cycles
4: Duration of loop 7 is 33 clock cycles
4: Duration of loop 8 is 33 clock cycles
4: Duration of loop 9 is 33 clock cycles
4: device timings of run 2:
4: TimingResult contains 10 results:
4: Duration of loop 0 is 22573 clock cycles
4: Duration of loop 1 is 22335 clock cycles
4: Duration of loop 2 is 21409 clock cycles
4: Duration of loop 3 is 22105 clock cycles
4: Duration of loop 4 is 22678 clock cycles
4: Duration of loop 5 is 22888 clock cycles
4: Duration of loop 6 is 22534 clock cycles
4: Duration of loop 7 is 22149 clock cycles
4: Duration of loop 8 is 22226 clock cycles
4: Duration of loop 9 is 22630 clock cycles
4: host timings of run 2:
4: TimingResult contains 10 results:
4: Duration of loop 0 is 32 clock cycles
4: Duration of loop 1 is 33 clock cycles
4: Duration of loop 2 is 33 clock cycles
4: Duration of loop 3 is 33 clock cycles
4: Duration of loop 4 is 33 clock cycles
4: Duration of loop 5 is 33 clock cycles
4: Duration of loop 6 is 34 clock cycles
4: Duration of loop 7 is 32 clock cycles
4: Duration of loop 8 is 31 clock cycles
4: Duration of loop 9 is 31 clock cycles
4: /var/jenkins/workspace/Kokkos/core/unit_test/TestPageMigration.hpp:206: Failure
4: Value of: passed
4:   Actual: false
4: Expected: true
4: [  FAILED  ] hip_managed.page_migration (16 ms)
```
which is pretty close to my experience with SYCL.

core/unit_test/CMakeLists.txt
```diff
@@ -545,7 +569,14 @@ if(Kokkos_ENABLE_CUDA)
       cuda/TestCuda_TeamScratchStreams.cpp
       ${Cuda_SOURCES3}
       cuda/TestCuda_Spaces.cpp
-  )
+    )
+
+    KOKKOS_ADD_EXECUTABLE_AND_TEST(
+    UnitTest_Cuda4
```
dalg24, line 575: Any good reason to introduce a new executable for that test?

core/unit_test/TestPageMigration.hpp
JBludau: oh, good catch

core/unit_test/CMakeLists.txt
JBludau: Originated from the time of allocating large chunks of memory. Will add it to another exec

JBludau:
> Also, the test fails for HIP in the CI with something like
> 
> ```
> 4: [ RUN      ] hip_managed.page_migration
> 4: Page size as reported by os: 4096 bytes 
> 4: Allocating 100 pages of memory in pageMigratingMemorySpace.
> 4: Behavior found: 
> 4: Initial placement on device is: 0 we expect true 
> 4: Memory migrates back to GPU is: 0 we expect true 
> 4: Memory migrates at max once per access: 0 we expect true 
> 4: 
> 4: Please look at the following timings. A migration was marked detected if the time was larger than 13600 for the device 
> 4: 
> 4: device timings of run 0:
> 4: TimingResult contains 10 results:
> 4: Duration of loop 0 is 28412 clock cycles
> 4: Duration of loop 1 is 22968 clock cycles
> 4: Duration of loop 2 is 23048 clock cycles
> 4: Duration of loop 3 is 22856 clock cycles
> 4: Duration of loop 4 is 22896 clock cycles
> 4: Duration of loop 5 is 23355 clock cycles
> 4: Duration of loop 6 is 23344 clock cycles
> 4: Duration of loop 7 is 21965 clock cycles
> 4: Duration of loop 8 is 21477 clock cycles
> 4: Duration of loop 9 is 24143 clock cycles
> 4: host timings of run 0:
> 4: TimingResult contains 10 results:
> 4: Duration of loop 0 is 33 clock cycles
> 4: Duration of loop 1 is 33 clock cycles
> 4: Duration of loop 2 is 33 clock cycles
> 4: Duration of loop 3 is 33 clock cycles
> 4: Duration of loop 4 is 33 clock cycles
> 4: Duration of loop 5 is 32 clock cycles
> 4: Duration of loop 6 is 33 clock cycles
> 4: Duration of loop 7 is 33 clock cycles
> 4: Duration of loop 8 is 33 clock cycles
> 4: Duration of loop 9 is 32 clock cycles
> 4: device timings of run 1:
> 4: TimingResult contains 10 results:
> 4: Duration of loop 0 is 22179 clock cycles
> 4: Duration of loop 1 is 22560 clock cycles
> 4: Duration of loop 2 is 22308 clock cycles
> 4: Duration of loop 3 is 21959 clock cycles
> 4: Duration of loop 4 is 22544 clock cycles
> 4: Duration of loop 5 is 23154 clock cycles
> 4: Duration of loop 6 is 22839 clock cycles
> 4: Duration of loop 7 is 21566 clock cycles
> 4: Duration of loop 8 is 22493 clock cycles
> 4: Duration of loop 9 is 22335 clock cycles
> 4: host timings of run 1:
> 4: TimingResult contains 10 results:
> 4: Duration of loop 0 is 32 clock cycles
> 4: Duration of loop 1 is 33 clock cycles
> 4: Duration of loop 2 is 33 clock cycles
> 4: Duration of loop 3 is 33 clock cycles
> 4: Duration of loop 4 is 33 clock cycles
> 4: Duration of loop 5 is 33 clock cycles
> 4: Duration of loop 6 is 33 clock cycles
> 4: Duration of loop 7 is 33 clock cycles
> 4: Duration of loop 8 is 33 clock cycles
> 4: Duration of loop 9 is 33 clock cycles
> 4: device timings of run 2:
> 4: TimingResult contains 10 results:
> 4: Duration of loop 0 is 22573 clock cycles
> 4: Duration of loop 1 is 22335 clock cycles
> 4: Duration of loop 2 is 21409 clock cycles
> 4: Duration of loop 3 is 22105 clock cycles
> 4: Duration of loop 4 is 22678 clock cycles
> 4: Duration of loop 5 is 22888 clock cycles
> 4: Duration of loop 6 is 22534 clock cycles
> 4: Duration of loop 7 is 22149 clock cycles
> 4: Duration of loop 8 is 22226 clock cycles
> 4: Duration of loop 9 is 22630 clock cycles
> 4: host timings of run 2:
> 4: TimingResult contains 10 results:
> 4: Duration of loop 0 is 32 clock cycles
> 4: Duration of loop 1 is 33 clock cycles
> 4: Duration of loop 2 is 33 clock cycles
> 4: Duration of loop 3 is 33 clock cycles
> 4: Duration of loop 4 is 33 clock cycles
> 4: Duration of loop 5 is 33 clock cycles
> 4: Duration of loop 6 is 34 clock cycles
> 4: Duration of loop 7 is 32 clock cycles
> 4: Duration of loop 8 is 31 clock cycles
> 4: Duration of loop 9 is 31 clock cycles
> 4: /var/jenkins/workspace/Kokkos/core/unit_test/TestPageMigration.hpp:206: Failure
> 4: Value of: passed
> 4:   Actual: false
> 4: Expected: true
> 4: [  FAILED  ] hip_managed.page_migration (16 ms)
> ```
> 
> which is pretty close to my experience with SYCL.

This should not even execute for HIP given the hardware in our CI has no proper page migration. Will investigate again. Thought the includeguard on the test would prevent it.

JBludau:
BLOCKED by #5327 as it would break the CI otherwise

core/unit_test/CMakeLists.txt
JBludau: done in [10c9dc](https://github.com/kokkos/kokkos/pull/5289/commits/10c9dcf1fad358cf53df4e5fc3e09d1d72b7ac7f)

JBludau:
Documentation issue [#149](https://github.com/kokkos/kokkos-core-wiki/issues/149)

core/perf_test/CMakeLists.txt
```diff
@@ -64,6 +64,13 @@ IF(Kokkos_ENABLE_OPENMPTARGET)
   )
 ENDIF()
 
+IF((KOKKOS_ENABLE_CUDA AND KOKKOS_ENABLE_CUDA_LAMBDA) OR KOKKOS_ENABLE_HIP OR KOKKOS_ENABLE_SYCL)
```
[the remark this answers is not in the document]
JBludau: The UVM is always usable with enabled Cuda backend. The option you are referring to just makes CudaUVM the default memory space

core/unit_test/CMakeLists.txt
```diff
@@ -655,7 +679,8 @@ if(Kokkos_ENABLE_CUDA)
       cuda/TestCuda_TeamScratchStreams.cpp
       ${Cuda_SOURCES3}
       cuda/TestCuda_Spaces.cpp
-  )
+      ${Cuda_SOURCES_SHAREDSPACE}
+    )
```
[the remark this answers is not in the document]
JBludau: because for hip we just have one big executable for all tests, for the other two backends we have smaller ones (cuda) or separate executables (sycl)

masterleinad (commented):
I would be happy with this if we only keep the performance test part and discard the unit test (since that's very brittle by using `clock_tic`, in particular for `SYCL`).

JBludau:
I see your concern, but would need some more information to address it. Why do you think it is brittle using the `clock_tic`? Is it a problem for `SYCL` with Intel GPUs (we use a function they provide in `clock_tic`)? It is disabled for `SYCL` with other GPUs, as `clock_tic` returns `0` in these cases. 

Some background information:
The initial reason for adding a unit test was, that we should see it if the memory changes the behavior (the behavior being the page migration). 

The timing based test (which was the first proposal for a unit test) does require more resources (especially memory). Furthermore, the timing based test needs a lot of warmup cycles. This is because we need to ensure the core clock of the GPU is **not** the reason for the slower memory access but the page migration is. Unfortunately, this resulted in a test that would not be sustainable as a unit test due to the long runtime and high resource requirements.

After some discussion with @crtrott and @dalg24 and investigation of core clock while running the test we concluded, that a clock count would be a warmup-independent/core-clock-independent measure (given a memory bound kernel).
The `clock_tic` we use for this is not comparable for CPU and GPU, but this was never considered something we would want to do in the unit test. The unit test should just make sure the memory is actually migrating at every first touch and can be accessed locally on subsequent touches in the same `ExecutionSpace`.

To still be able to get more useful information as a user, I proposed adding the timing based test as a perf test (it is built but **not** executed by our CI) and added some transfer bandwidth information. This way a user could run the test to see where the initial placement is and how fast the migration to the GPU and CPU are.
I was thinking about adding information about overcommitment but did not as I thought this would be something the user can look up in documentation of the vendors.

core/perf_test/test_sharedSpace.cpp
```diff
@@ -0,0 +1,300 @@
...
+void printTimings(std::ostream& out, std::vector<double> const& tr,
+                  size_t numBytes, double threshold) {
+  out << "TimingResult contains " << tr.size() << " results:\n";
+  for (auto it = tr.begin(); it != tr.end(); ++it) {
+    out << "Duration of loop " << it - tr.begin() << " is " << *it
+        << " seconds.";
+    if ((*it) > threshold) {
+      out << " Marked as page migation. ";
+    }
+    out << "The transfer rate is " << numBytes / 1000000000 / (*it)
```
Rombur, line 74: Dumb question but should we use `numBytes / (1000)^3` or `numBytes / (1024)^3`?

Rombur (approved):

core/perf_test/test_sharedSpace.cpp
PhilMiller: Whichever one is used, I think it's preferable for readability to write out the multiplication there, rather than writing the product directly. Or, if we stick with powers of 10, you can insert `'` as digit separators

JBludau: AFAIK `GB` is short for Gigabytes, thus `(1000)^3`, `GiB` is short for Gibibytes, thus `(1024)^3`. I think it is mostly a marketing gag for selling hard drives (that is acutally why a 16GB thumbdrive has less than you would expect). 

For something like a pci-e bus I think the most frequent thing is `giga bits per second` ... we could switch to that, but then the user has to think for himself.

For the output format: I will take whatever anyone prefers.

masterleinad (changes requested):
Disable the test for `SYCL` with a `FIXME_SYCL` saying that migration from host to device can't be detected so we can discuss if that is the intended behavior or not later.

core/perf_test/test_sharedSpace.cpp
masterleinad, line 74:
```suggestion
    out << " The transfer rate is " << numBytes / 1000000000 / (*it)
```

JBludau: done in [802214](https://github.com/kokkos/kokkos/pull/5289/commits/802214e37b26b5079401cb94b8d5f21a6959152d)

masterleinad (approved):

JBludau:
I will rewrite the history on this to reduce the amounts of commits

core/perf_test/test_sharedSpace.cpp
mhoemmen: I'm not a Kokkos developer, but here are some thoughts.

1. C++14 lets you use `'` as a digit separator for any decimal integer literal -- please do consider doing that.
2. Would you consider converting these numbers to `double` to avoid truncation when dividing?
3. `1000000000` is uncomfortably close to `numeric_limits<int>::max()`.  Would you consider expressing the literal as a `double` or as `1000000000LL` or `1000000000uLL`?

JBludau:
no changes made except: 

- switching `numPages` from `unsigned int` to `size_t` in perf_test to prevent cutoff errors with large memchunks.
- adding markers to structure output of perf-test
- added minor additional information in output
- report total memory speed, not only the read part

core/perf_test/test_sharedSpace.cpp
JBludau: @mhoemmen I switched to double and doing the literal as `pow(1000.0,3)` which I though is easiest to read together with a variable named `numBytes`.

core/src/Kokkos_Core_fwd.hpp
```diff
@@ -121,9 +121,11 @@ class InitializationSettings;
   [[clang::annotate("DefaultExecutionSpace")]]
 #define KOKKOS_IMPL_DEFAULT_HOST_EXEC_SPACE_ANNOTATION \
   [[clang::annotate("DefaultHostExecutionSpace")]]
+#define KOKKOS_IMPL_SHARED_SPACE_ANNOTATION [[clang::annotate("SharedSpace")]]
```
dalg24, line 124: Why did you use a Clang annotation?

core/src/Kokkos_Core_fwd.hpp
```diff
@@ -184,6 +186,15 @@ using DefaultHostExecutionSpace KOKKOS_IMPL_DEFAULT_HOST_EXEC_SPACE_ANNOTATION =
...
 #endif
 
+#if defined(KOKKOS_ENABLE_CUDA)
+using SharedSpace KOKKOS_IMPL_SHARED_SPACE_ANNOTATION = CudaUVMSpace;
+#elif defined(KOKKOS_ENABLE_HIP)
+using SharedSpace KOKKOS_IMPL_SHARED_SPACE_ANNOTATION = HIPManagedSpace;
+#elif defined(KOKKOS_ENABLE_SYCL)
+using SharedSpace KOKKOS_IMPL_SHARED_SPACE_ANNOTATION =
+    Experimental::SYCLSharedUSMSpace;
+#endif
```
dalg24, line 196: We had discussed the need for both a macro and a boolean variable that can be used in constant expressions to detect the existence of that alias

core/src/Kokkos_Core_fwd.hpp
JBludau: symmetry with `DefaultExecSpace` etc.

core/src/Kokkos_Core_fwd.hpp
JBludau: I should have written this down :-D added in [c28db5](https://github.com/kokkos/kokkos/pull/5289/commits/c28db58c5cb9bc892226a542633e2eeeb440b2ee)

core/src/Kokkos_Core_fwd.hpp
```diff
@@ -184,6 +186,21 @@ using DefaultHostExecutionSpace KOKKOS_IMPL_DEFAULT_HOST_EXEC_SPACE_ANNOTATION =
...
+#define KOKKOS_HAS_SHARED_SPACE true
+#else
+#define KOKKOS_HAS_SHARED_SPACE false
```
dalg24, line 201: No only define it when it's available and set it to `1`
Because our users typically do `#ifdef FOO`

core/src/Kokkos_Core_fwd.hpp
crtrott: I actually want in pure Host builds SharedSpace == HostSpace. Reasoning: it is shared and migrates to every existing ExecutionSpace and thus behaves performance wise basically as expected.

crtrott (changes requested):
I actually want SharedSpace alias defined in pure Host builds (i.e. let it be HostSpace).

Essentially the semantics of SharedSpace are:
(1) every existing execution space type can access it.
(2) if accessing a SharedSpace repeatedly from the same execution space, without accessing it from some other one in between, it will perform close to the performance of the native memory space of that execution space. 

Both of these conditions are fulfilled by HostSpace in a pure Host build. 

Note that the second condition is what makes Cuda/HIPHostPinnedSpace not eligible for this alias.

JBludau:
Okay, I changed the following:

Except for `OpenMPTarget` and `OpenACC` the `SharedSpace` alias is now defined. If there is a device (`Cuda`,`HIP`,`SYCL`) it will point to the corresponding page migrating MemorySpace. If it is a host only build the `SharedSpace` points to `HostSpace`. There is both a preproc define and a constexpr function for checking if the feature is available (We do not have a configure time check, but this would be trivial for users if they need to know)

The unit test now test for the conditions @crtrott proposed:
> Essentially the semantics of SharedSpace are:
> (1) every existing execution space type can access it.
> (2) if accessing a SharedSpace repeatedly from the same execution space, without accessing it from some other one in  between, it will perform close to the performance of the native memory space of that execution space.

Thus we do not evaluate the first access in a new ExecutionSpace and compare the subsequent accesses to the speed of pure local memory. If we detect more than 20% deviation in the memory speed the test fails.

core/src/Kokkos_Core_fwd.hpp
JBludau: done

crtrott (approved):
I did not in detail review the tests, but my concerns are addressed regarding when this is defined.

core/perf_test/CMakeLists.txt
PhilMiller, line 67: It seems like the dependence on `KOKKOS_ENABLE_CUDA_LAMBDA` could be pretty trivially eliminated, since it's just one bit of code that would need to be converted to a named functor

core/src/Kokkos_Core_fwd.hpp
```diff
@@ -184,6 +186,30 @@ using DefaultHostExecutionSpace KOKKOS_IMPL_DEFAULT_HOST_EXEC_SPACE_ANNOTATION =
...
+#define KOKKOS_HAS_SHARED_SPACE 1
+#elif defined(KOKKOS_ENABLE_HIP)
+using SharedSpace KOKKOS_IMPL_SHARED_SPACE_ANNOTATION = HIPManagedSpace;
+#define KOKKOS_HAS_SHARED_SPACE 1
+#elif defined(KOKKOS_ENABLE_SYCL)
+using SharedSpace KOKKOS_IMPL_SHARED_SPACE_ANNOTATION =
+    Experimental::SYCLSharedUSMSpace;
+#define KOKKOS_HAS_SHARED_SPACE 1
+// if only host compile point to HostSpace
+#elif !defined(KOKKOS_ENABLE_OPENACC) && !defined(KOKKOS_ENABLE_OPENMPTARGET)
```
PhilMiller, line 201: This exclusive phrasing feels risky, since anyone doing a future backend addition will have to find and adjust this. It would be preferable if there were a definition somewhere else that described "backend with distinct memory from `HostSpace` is enabled", and then we could check the absence of that.

core/perf_test/CMakeLists.txt
JBludau: good catch, I already did this but forgot to remove the guard

core/src/Kokkos_Core_fwd.hpp
JBludau: yeah, I totally agree, I don't like it ether. I like your suggestion though. We should probably hand the definition to the backends ...

dalg24: We can't do much that is future proof here and that is ok.  We have the same issue with our defining default execution spaces.

core/perf_test/CMakeLists.txt
JBludau: done in [8bf552](https://github.com/kokkos/kokkos/pull/5289/commits/8bf5526aa5baf9b6d2dcb59acf41f2736109a0e6)

core/perf_test/test_sharedSpace.cpp
```diff
@@ -0,0 +1,343 @@
...
+        migratableData, numRepetitions));
+
+    // WARMUP HOST
+    incrementInLoop<Kokkos::DefaultHostExecutionSpace>(
+        hostData,
+        numWarmupRepetitions);  // warming up host
+    // GET RESULTS HOST
+    hostResults.push_back(incrementInLoop<Kokkos::DefaultHostExecutionSpace>(
+        migratableData, numRepetitions));
+  }
```
PhilMiller, line 213: Maybe I'm reading things wrong, but shouldn't the warmup calls here both access `migratableData` - i.e. pull the pages to the space being measured?

Or is the warmup being performed here the clock-speed warmup, to make sure that each core is running at full speed? If so, please elaborate the comments within this loop to clarify.

JBludau: It is indeed to ensure the core-clock is at max when we do the actual measurement that does include the page migration. Therefore, it should not use the `migratableData`

JBludau: hope [3feb4d](https://github.com/kokkos/kokkos/pull/5289/commits/3feb4de9b4700b0d31b4df2e0347c77487aeda20) is helping

JBludau:
Retest this please

core/src/Kokkos_Core_fwd.hpp
dalg24: Drop it, the other one was helpful in the Kokkos specific ClangTidy checks Drew wrote in [kokkos/llvm-project](https://github.com/kokkos/llvm-project)

Rombur (approved):

PR #5294, opened 2022-07-29
Only test product for bhalf_t for N<=5

masterleinad:
Following https://github.com/kokkos/kokkos/blob/39c677e0113998d433dacbd06da8915dfa8b6afb/core/unit_test/TestReducers.hpp#L299-L310,
we can only represent numbers up to 2^8=256. Given that we multiply numbers from 1 to 4 in the product test, it seems that we can only multiply five numbers reliably with the worst case being only 3s, i.e., 3^5=243. The more even numbers we have the more we can represent, of course.


nliber (approved):

ajpowelsnl (approved):

crtrott (approved):

dalg24:
Cuda build failure is unrelated
```
CMake Error: : System Error: No space left on device

PR #5295, opened 2022-07-29
C++17 cleanup

masterleinad:
Depends on #5277. This includes all the cleanup (mostly that `KOKKOS_ENABLE_CXX17` or `KOKKOS_ENABLE_CXX20` is true) done to evaluate the CI setup for #5277.


cmake/kokkos_corner_cases.cmake
```diff
@@ -47,17 +47,14 @@ IF(KOKKOS_CXX_COMPILER_ID STREQUAL AppleClang AND KOKKOS_ENABLE_OPENMP)
 ENDIF()
 
 
-# FIXME_CXX17
-IF (KOKKOS_CXX_STANDARD STREQUAL 17)
-  IF (KOKKOS_CXX_COMPILER_ID STREQUAL GNU AND KOKKOS_CXX_COMPILER_VERSION VERSION_LESS 7)
-    MESSAGE(FATAL_ERROR "You have requested C++17 support for GCC ${KOKKOS_CXX_COMPILER_VERSION}. Although CMake has allowed this and GCC accepts -std=c++1z/c++17, GCC < 7 does not properly support *this capture. Please upgrade the compiler if you do need C++17 support.")
-  ENDIF()
+IF (KOKKOS_CXX_COMPILER_ID STREQUAL GNU AND KOKKOS_CXX_COMPILER_VERSION VERSION_LESS 7)
+  MESSAGE(FATAL_ERROR "You have requested C++17 support for GCC ${KOKKOS_CXX_COMPILER_VERSION}. Although CMake has allowed this and GCC accepts -std=c++1z/c++17, GCC < 7 does not properly support *this capture. Please upgrade the compiler if you do need C++17 support.")
```
crtrott, line 51: Lets update the message here

cmake/kokkos_corner_cases.cmake
```diff
@@ -47,17 +47,14 @@ IF(KOKKOS_CXX_COMPILER_ID STREQUAL AppleClang AND KOKKOS_ENABLE_OPENMP)
...
+ENDIF()
 
-  IF (KOKKOS_CXX_COMPILER_ID STREQUAL NVIDIA AND KOKKOS_CXX_COMPILER_VERSION VERSION_LESS 11)
-    MESSAGE(FATAL_ERROR "You have requested C++17 support for NVCC ${KOKKOS_CXX_COMPILER_VERSION}. NVCC only supports C++17 from version 11 on. Please upgrade the compiler if you need C++17 support.")
-  ENDIF()
-  IF (KOKKOS_CXX_COMPILER_ID STREQUAL NVIDIA AND KOKKOS_ENABLE_CUDA_CONSTEXPR)
-    MESSAGE(WARNING "You have requested -DKokkos_ENABLE_CUDA_CONSTEXPR=ON with C++17 support for NVCC ${KOKKOS_CXX_COMPILER_VERSION} which is known to trigger compiler bugs. See https://github.com/kokkos/kokkos/issues/3496")
-  ENDIF()
+IF (KOKKOS_CXX_COMPILER_ID STREQUAL NVIDIA AND KOKKOS_CXX_COMPILER_VERSION VERSION_LESS 11)
+  MESSAGE(FATAL_ERROR "You have requested C++17 support for NVCC ${KOKKOS_CXX_COMPILER_VERSION}. NVCC only supports C++17 from version 11 on. Please upgrade the compiler if you need C++17 support.")
```
crtrott, line 55: Lets change this message indicating we require C++17

cmake/kokkos_corner_cases.cmake
```diff
@@ -47,17 +47,14 @@ IF(KOKKOS_CXX_COMPILER_ID STREQUAL AppleClang AND KOKKOS_ENABLE_OPENMP)
...
+ENDIF()
+IF (KOKKOS_CXX_COMPILER_ID STREQUAL NVIDIA AND KOKKOS_ENABLE_CUDA_CONSTEXPR)
+  MESSAGE(WARNING "You have requested -DKokkos_ENABLE_CUDA_CONSTEXPR=ON with C++17 support for NVCC ${KOKKOS_CXX_COMPILER_VERSION} which is known to trigger compiler bugs. See https://github.com/kokkos/kokkos/issues/3496")
```
crtrott, line 58: the issue indicates this was fixed in CUDA 11.2 we should check for that and fix the warning as well.

crtrott (changes requested):
Lets fix the cmake messages otherwise looks good.

cmake/kokkos_corner_cases.cmake
masterleinad: I checked that I see errors when compiling `Kokkos` with tests with `nvcc-11.0` but not with `nvcc-11.2`.

masterleinad: Given that problem we could of course also consider requiring cuda-11.2 as minimum compiler version.

core/src/Kokkos_HPX.hpp
```diff
@@ -471,21 +471,21 @@ class HPX {
...
+    reset_on_exit_parallel(HPX const &) = default;
+    ~reset_on_exit_parallel()           = default;
   };
 
-  struct KOKKOS_ATTRIBUTE_NODISCARD reset_count_on_exit_parallel {
-    reset_count_on_exit_parallel() {}
-    ~reset_count_on_exit_parallel() {}
+  struct [[nodiscard]] reset_count_on_exit_parallel {
+    reset_count_on_exit_parallel()  = default;
+    ~reset_count_on_exit_parallel() = default;
```
dalg24, line 488: Yuk

dalg24 (commented):
I am concerned about the `KOKKOS_ATTRIBUTE_NODISCARD` macro removal
Maybe we can emit deprecation warning for a while before we remove it

cmake/kokkos_corner_cases.cmake
crtrott: unfortunately that is an issue because technically supposedly Spectrum MPI is only supported up to CUDA 11.1 maybe?

core/src/Kokkos_Macros.hpp
```diff
@@ -645,13 +633,6 @@ static constexpr bool kokkos_omp_on_host() { return false; }
 #define KOKKOS_ENABLE_CUDA_LDG_INTRINSIC
 #endif
 
-#if defined(KOKKOS_ENABLE_CXX17) || defined(KOKKOS_ENABLE_CXX20) || \
-    defined(KOKKOS_ENABLE_CXX23)
-#define KOKKOS_ATTRIBUTE_NODISCARD [[nodiscard]]
```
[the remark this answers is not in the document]
crtrott: Yeah lets just unconditionally define it this way. Its ok that WE don't use it internally though, and directly write [[nodiscard]]

core/src/Kokkos_HPX.hpp
masterleinad: I needed this to avoid `clang-format` from being really confused (and I found this a better solution than using `clang-format off`).

core/src/Kokkos_Macros.hpp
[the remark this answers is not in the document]
masterleinad: Yeah, I couldn't quite make up my mind about keeping it or not. Restored now.

core/src/Kokkos_HPX.hpp
dalg24: I meant I would do `struct reset_count_on_exit_parallel {};`

masterleinad:
Retest this please.

cmake/kokkos_corner_cases.cmake
dalg24: Ping

masterleinad: See 36251ddb2.

crtrott (approved):

dalg24 (approved):

PR #5296, opened 2022-07-29
Initial support for multiple OpenACC execution space instances:

seyonglee:
- Added a new OpenACC execution space constructor, OpenACC(int
      async_arg), which takes non-negative integer exepression as an
      argument representing the OpenACC device activity queue.
    - Implement asynchronous memory transfers in DeepCopy().
    - Implement fence().


core/src/OpenACC/Kokkos_OpenACCSpace.cpp
```diff
@@ -85,8 +85,6 @@ void *Kokkos::Experimental::OpenACCSpace::impl_allocate(
 
   void *ptr = nullptr;
 
-  // FIXME_OPENACC multiple device instances are not yet supported, and thus
-  // exec_space is ignored for now.
```
dalg24, line 89: If the implementation is the same than the other overload and it is not a FIXME any more, then do
```C+
(void)exec_space;
impl_allocate(arg_label, arg_alloc_size, arg_logical_size, arg_handle);

File: core/src/OpenACC/Kokkos_OpenACC_Instance.cpp [hunk 2]
```diff
@@ -58,31 +58,38 @@ namespace Kokkos {
...
-    // FIXME_OPENACC if multiple execution space instances are supported,
-    // device id variable should be explicitly set to the value returned by
-    // acc_get_device_num(acc_device_host).
-  } else {
-    using Kokkos::Impl::get_gpu;
-    int const dev_num = get_gpu(settings);
-    acc_set_device_num(dev_num, OpenACC_Traits::dev_type);
+int Kokkos::Experimental::Impl::OpenACCInternal::verify_is_initialized(
+    const char* const label) const {
+  if (m_accDev < 0) {
```
dalg24, line 69: I would only check that `m_is_initialized` is `true`.

core/src/OpenACC/Kokkos_OpenACC_Instance.hpp
```diff
@@ -46,24 +46,31 @@
...
 class OpenACCInternal {
   bool m_is_initialized = false;
 
-  OpenACCInternal()                       = default;
   OpenACCInternal(const OpenACCInternal&) = default;
   OpenACCInternal& operator=(const OpenACCInternal&) = default;
 
  public:
-  static OpenACCInternal* singleton();
+  int m_accDev;
```
dalg24, line 65:
```suggestion
  static int m_accDev;
```
You might as well make it static since it is not legal to use anything else than the device selected at initialization.

core/src/OpenACC/Kokkos_OpenACC_Instance.hpp
```diff
@@ -46,24 +46,31 @@
...
+  int m_async_id;
+  uint32_t m_instance_id;
 
-  void initialize(InitializationSettings const& settings);
+  static OpenACCInternal& singleton();
+
+  int verify_is_initialized(const char* const label) const;
+
+  void initialize(int device_id, int async_arg = acc_async_sync);
```
dalg24, line 73: I don't think passing a device ID is a good idea considering that it is only legal to use the device selected at initialization

core/src/OpenACC/Kokkos_OpenACC_Instance.hpp
```diff
@@ -46,24 +46,31 @@
...
 
-  OpenACCInternal()                       = default;
   OpenACCInternal(const OpenACCInternal&) = default;
   OpenACCInternal& operator=(const OpenACCInternal&) = default;
 
  public:
-  static OpenACCInternal* singleton();
+  int m_accDev;
+  int m_async_id;
+  uint32_t m_instance_id;
```
dalg24, line 67: Why do you store the instance ID?

core/src/OpenACC/Kokkos_OpenACC.cpp
```diff
@@ -46,25 +46,47 @@
...
-  Impl::OpenACCInternal::singleton()->initialize(settings);
+  acc_init(Kokkos::Experimental::Impl::OpenACC_Traits::dev_type);
+  Impl::OpenACCInternal::singleton().initialize(
+      Kokkos::Impl::get_gpu(settings));
 }
 
 void Kokkos::Experimental::OpenACC::impl_finalize() {
-  Impl::OpenACCInternal::singleton()->finalize();
+  Impl::OpenACCInternal::singleton().finalize();
+  acc_shutdown(Kokkos::Experimental::Impl::OpenACC_Traits::dev_type);
```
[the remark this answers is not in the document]
seyonglee: `acc_shutdown_device` is also another latest feature introduced in the OpenACC Version 3.2, and thus there exists no OpenACC compiler implementation supporting this feature yet.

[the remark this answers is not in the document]
dalg24: OK about the preferring `acc_shutdown` over `acc_shutdown_device` part.
What about interoperability?  Is not going to be an issue that Kokkos calls `acc_shutdown`?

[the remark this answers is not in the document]
seyonglee: Inserting explicit `acc_init` and acc_shutdown` APIs mean that the Kokkos program will serve as the main program in terms of the interoperability with other OpenACC program. (The Kokkos program will explicitly initialize and shutdown the OpenACC runtime, and other OpenACC code can be inserted only during the execution of the Kokkos program.)
In fact, explicit `acc_init` and `acc_shutdown` API calls are not necessary; if we remove these calls from the Kokkos backend, it will mean that other OpenACC program can serve as the main program controlling when to initialize and/or shutdown the OpenACC runtime.

core/src/OpenACC/Kokkos_OpenACC.hpp
```diff
@@ -103,6 +107,13 @@ class OpenACC {
   static int concurrency() { return 256000; }  // FIXME_OPENACC
   static bool in_parallel() { return acc_on_device(acc_device_not_host); }
   uint32_t impl_instance_id() const noexcept;
+  inline Impl::OpenACCInternal* impl_internal_space_instance() const {
+    return m_space_instance.get();
+  }
+
+  int get_async_id() const;
+  int get_device_id() const;
```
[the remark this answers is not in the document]
seyonglee: Those names were chosen not to be confused with the existing OpenACC APIs and keywords. However, being aligned with the OpenACC terminology is preferred, we can change these names as suggested.

core/src/OpenACC/Kokkos_OpenACC.cpp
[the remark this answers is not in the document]
seyonglee: Deleted `acc_init` and `acc_shutdown` API calls.

core/src/OpenACC/Kokkos_OpenACC.hpp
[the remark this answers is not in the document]
seyonglee: API names are changed as suggested.

core/src/OpenACC/Kokkos_OpenACCSpace.cpp
seyonglee: Updated.

core/src/OpenACC/Kokkos_OpenACC_Instance.cpp
seyonglee: Updated.

core/src/OpenACC/Kokkos_OpenACC_Instance.hpp
seyonglee: Changed.

core/src/OpenACC/Kokkos_OpenACC_Instance.hpp
seyonglee: The `device_id` parameter is removed.

core/src/OpenACC/Kokkos_OpenACC_Instance.hpp
seyonglee: It simply follows the CUDA backend implementation pattern, which may be to save unnecessary type-casting operations.

core/src/OpenACC/Kokkos_OpenACC_Instance.cpp
```diff
@@ -58,31 +58,33 @@ namespace Kokkos {
 bool show_warnings() noexcept;
 }
 
-Kokkos::Experimental::Impl::OpenACCInternal*
+int Kokkos::Experimental::Impl::OpenACCInternal::m_accDev = -1;
```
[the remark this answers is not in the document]
dalg24: Will do.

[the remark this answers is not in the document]
dalg24: Actually I wonder if we should not set it to an arbitrary non-negative value that is unlikely to valid instead since a negative value tells the OpenACC runtime to revert to its (implementation-defined) default behavior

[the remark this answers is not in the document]
seyonglee: A comment is added. 
(Strictly speaking, the OpenACC runtime may not complain an error even if the device number is negative (In OpenACC spec, how to handle negative device number is implementation-defined). However, the Kokkos runtime will issue an error.)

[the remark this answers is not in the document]
dalg24: I was going for
```C++
// Arbitrary value to denote that we don't know yet what device to use.
// The intent is that the value below is invalid.  The OpenACC runtime
// interprets negative values as "revert to (implementation-defined) default
// behavior".
int Kokkos::Experimental::Impl::OpenACCInternal::m_acc_device_num = 255;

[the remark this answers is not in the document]
Comment (dalg24) [reply core/src/OpenACC/Kokkos_OpenACC_Instance.cpp:63]: We actually do not know what the OpenACC runtime use to denote the host.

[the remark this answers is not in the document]
Comment (seyonglee) [reply core/src/OpenACC/Kokkos_OpenACC_Instance.cpp:63]: In OpenACC, handling device number bigger that the number of available devices of a given type is also implementation-defined, and thus using an arbitrary big positive integer (255) will now solve the problem.

File: core/src/OpenACC/Kokkos_OpenACC.cpp [hunk 9]
```diff
@@ -73,8 +74,23 @@ Kokkos::Experimental::OpenACC::OpenACC(int async_arg)
...
+    }
+    acc_set_device_type(acc_device_host);
+    Impl::OpenACCInternal::m_acc_device_num =
+        acc_get_device_num(acc_device_host);
+  } else {
+    using Kokkos::Impl::get_gpu;
+    int const dev_num = get_gpu(settings);
+    acc_set_device_num(dev_num, Impl::OpenACC_Traits::dev_type);
+    Impl::OpenACCInternal::m_acc_device_num = dev_num;
+  }
```
seyonglee, line 93: One issue on moving the host-fallback handling code from `get_device_count()` in `Kokkos_Core.cpp` to here is that even if the host fallback is allowed, the Kokkos device initialization phase can fail if there exists any non-host device since `get_device_count()` will return 0. 
I think that a better way is to create a separate `detect_device_count()` function in the OpenACC backend, which handles the host fallback case, and `get_device_count()` internally calls the `detect_device_count()`. In this case, `get_device_count()` will always return correct device count.

dalg24: That code path you refer to would never be called.  It is exclusively called in `Kokkos::Impl::get_gpu`.

seyonglee: The problem will occur if `Kokkos::Impl:get_gpu` is called at least once in such a case.
As I know, `Kokkos::Impl:get_gpu` is called only for device initialization, but if it is used in any future implementation, this latent problem will occur.

dalg24: I don't anticipate it will change and if we ever did that would break in other backends as well.

core/src/OpenACC/Kokkos_OpenACC.hpp
```diff
@@ -89,6 +90,8 @@ class OpenACC {
 
   OpenACC();
 
+  OpenACC(int async_arg);
+
```
JBludau, line 94: I would prefer an `explicit` on the one argument constructor, users might fuck up some things otherwise.
If we give the passed parameter a default value we could get rid of the default ctor

core/src/OpenACC/Kokkos_OpenACCSpace.cpp
```diff
@@ -80,24 +81,8 @@ void *Kokkos::Experimental::OpenACCSpace::impl_allocate(
...
     const size_t arg_alloc_size, const size_t arg_logical_size,
     const Kokkos::Tools::SpaceHandle arg_handle) const {
-  static_assert(sizeof(void *) == sizeof(uintptr_t),
-                "Error sizeof(void*) != sizeof(uintptr_t)");
-
-  void *ptr = nullptr;
-
-  // FIXME_OPENACC multiple device instances are not yet supported, and thus
-  // exec_space is ignored for now.
   (void)exec_space;
```
JBludau, line 84: Why do we need this?

core/src/OpenACC/Kokkos_OpenACC_Instance.cpp
```diff
@@ -44,45 +44,43 @@
...
+}
+
+bool Kokkos::Experimental::Impl::OpenACCInternal::verify_is_initialized(
+    const char* const label) const {
+  if (!m_is_initialized) {
+    Kokkos::abort((std::string("Kokkos::Experimental::OpenACC::") + label +
+                   " : ERROR device not initialized\n")
+                      .c_str());
+  }
+  return m_is_initialized;
```
JBludau, line 72: I am not sure about the function name if we abort if uninitialized. Maybe we should call it 'abort_if_not_initialized' and return void.

core/src/OpenACC/Kokkos_OpenACCSpace.cpp
dalg24: Silent warnings.  The other way to do it is not name the argument.

core/src/OpenACC/Kokkos_OpenACC_Instance.cpp
dalg24: I plan on revisiting that later.  At the moment just following the other backend current implementation.  Let us merge as is for now.

JBludau: fine with me

core/src/OpenACC/Kokkos_OpenACC.hpp
dalg24: Done

JBludau (approved):
looks good

core/src/OpenACC/Kokkos_OpenACCSpace.cpp
seyonglee: If we use `get_device_count` and `get_gpu` APIs to refer to non-host devices (e.g., GPUs) only, the current implementation will be OK, since it is valid for `get_device_count` to return 0 when the target system does not have any non-host device and the host fallback is allowed, as long as `get_gpu` is not used at all in this case.

core/src/OpenACC/Kokkos_OpenACC.cpp
seyonglee: (I don't know why but my last comment was deleted, and thus write again here)
The current implementation will be OK if we use `get_device_count` and `get_gpu` APIs only to refer to non-host devices (e.g., GPUs). Then, it will be OK for `get_device_count` to return 0 when there is no available non-host device, even if the host fallback is allowed (as long as `get_gpu` is not used at all in this case).

dalg24 (approved):

PR #4973, opened 2022-04-21
add cmake flag to enable mdspan and include mdspan as a tpl

nmm0:
This PR adds `KOKKOS_ENABLE_IMPL_MDSPAN` and `KOKKOS_ENABLE_EXTERNAL_MDSPAN` flags. Additionally, mdspan sources are bundled as a tpl.

This is ongoing work for #4421 

Note: the license on the mdspan sources caused a failure in our CI testing, so I had to exclude the tpl directory from the license check script. Probably the license for mdspan should be updated anyway...

Original text:

This PR adds some changes to add an experimental option to cmake for linking to mdspan.

Qs:
* Should we use the convention of `Kokkos_EXPERIMENTAL_*` for mdspan (and potentially other experimental features)?
* What should our plan be for when these features become core rather than experimental (i.e. will devs have to switch over to non-EXPERIMENTAL flags)?

Issue #4421, opened 2021-10-18
MDSpan design/cleanup


nmm0:
This issue is for tracking the MDSpan refactor issues that still remain

# General
- [ ] Remove unused code
- [ ] Go through all TODOs
- [ ] Verify GPU builds/tests (include std::experimental::mdspan)
- [ ] Verify markup (KOKKOS_INLINE_FUNCTION)

# Accessors
- [ ] Verify subview accounts for every case, clean up w/ partial specializations
- [ ] Write additional accessors
- [ ] MDSpanAccessor includes headers that don't exist (also file isn't used)
- [ ] Verify GPU works (std::pair doesn't work on the GPU)
- [ ] Selecting the correct accessor and solving ambiguities

# BasicView
- [ ] ADL customization points
- [ ] Constructor that takes everything, including Accessor (like mdspan)
- [ ] Ctor mixins don't work, label is blank, maybe other members are blank too

# Layout
- [ ] LayoutTiled is missing

ajpowelsnl:
@nliber , @nliber , @crtrott - where are we on this issue?  Any items we can tick off the list?

nmm0:
> @nliber , @nliber , @crtrott - where are we on this issue? Any items we can tick off the list?

still work in progress

PR #5304, opened 2022-08-02
Type conversion to and from std::extents and Kokkos DataType

nmm0:
Ongoing work on #4421 
This is dependent on #4973 being merged

The implementation is slightly more concise with C++17 and so I've used some 17 constructs


nmm0:
The diff is messy right now because its dependence on #4973 but the files to look at are:

* core/src/View/MDSpan/Kokkos_MDSpan_Extents.hpp
* core/unit_test/TestViewMDSpan.hpp

crtrott:
@nmm0 can you rebase this?

nmm0:
> @nmm0 can you rebase this?

@crtrott Done -- The unit test you added to the TPL PR is in the file TestMDSpan.hpp, while this PR had tests under TestViewMDSpan.hpp . These tests are directly related to view conversion so it may make sense to keep these separate, or we could combine the tests in the same file. What do you think?

core/src/View/MDSpan/Kokkos_MDSpan_Extents.hpp
```diff
@@ -0,0 +1,156 @@
...
+  static constexpr inline std::size_t value = std::size_t{0};
+};
+
+template <class SizeType, class Dimension, class Indices>
+struct ExtentsFromDimension;
+
+template <class SizeType, class Dimension, std::size_t... Indices>
+struct ExtentsFromDimension<SizeType, Dimension,
+                            std::index_sequence<Indices...>> {
+  using dimension_type = Dimension;
```
dalg24, line 115: Not used
```suggestion
```

core/src/View/MDSpan/Kokkos_MDSpan_Extents.hpp
```diff
@@ -0,0 +1,156 @@
...
+      SizeType,
+      ExtentFromDimension<Dimension::static_extent(Indices)>::value...>;
+};
+
+template <class Extents, class Indices>
+struct DimensionsFromExtent;
+
+template <class Extents, std::size_t... Indices>
+struct DimensionsFromExtent<Extents, std::index_sequence<Indices...>> {
+  using extents_type = Extents;
```
dalg24, line 126:
```suggestion
```

core/src/View/MDSpan/Kokkos_MDSpan_Extents.hpp
```diff
@@ -0,0 +1,156 @@
...
+struct DimensionFromExtent {
+  static constexpr inline std::size_t value = N;
+};
+
+template <>
+struct DimensionFromExtent<std::experimental::dynamic_extent> {
+  static constexpr inline std::size_t value = std::size_t{0};
+};
+
+template <class SizeType, class Dimension, class Indices>
```
dalg24, line 109: If we really want to leave control over it, then we better call it `SizeType -> IndexType`
```suggestion
template <class IndexType, class Dimension, class Indices>
```

core/src/View/MDSpan/Kokkos_MDSpan_Extents.hpp
```diff
@@ -0,0 +1,156 @@
...
+struct ExtentsFromDimension;
+
+template <class SizeType, class Dimension, std::size_t... Indices>
+struct ExtentsFromDimension<SizeType, Dimension,
+                            std::index_sequence<Indices...>> {
+  using dimension_type = Dimension;
+  using type           = std::experimental::extents<
+      SizeType,
```
dalg24, line 117:
```suggestion
      IndexType,
```

core/src/View/MDSpan/Kokkos_MDSpan_Extents.hpp
```diff
@@ -0,0 +1,156 @@
...
+struct DimensionsFromExtent<Extents, std::index_sequence<Indices...>> {
+  using extents_type = Extents;
+  using type         = ::Kokkos::Impl::ViewDimension<
+      DimensionFromExtent<extents_type::static_extent(Indices)>::value...>;
+};
+
+template <class DataType>
+struct ExtentsFromDataType {
+  using array_analysis = ::Kokkos::Impl::ViewArrayAnalysis<DataType>;
+  using size_type      = std::size_t;  // Mirrors Kokkos::View's size type
```
dalg24, line 134: Why wouldn't we go ahead and stop pretending that it is configurable in `ExtentsFromDimension` and not have it as a template parameter.  Even if it was we still wouldn't be able to deduce the right thing from just the `DataType`.

core/src/View/MDSpan/Kokkos_MDSpan_Extents.hpp
```diff
@@ -0,0 +1,156 @@
...
+              "Including non-public Kokkos header files is not allowed.");
+#else
+KOKKOS_IMPL_WARNING("Including non-public Kokkos header files is not allowed.")
+#endif
+#endif
+
+#ifndef KOKKOS_EXPERIMENTAL_MDSPAN_EXTENTS_HPP
+#define KOKKOS_EXPERIMENTAL_MDSPAN_EXTENTS_HPP
+
+#include <Kokkos_Core_fwd.hpp>
```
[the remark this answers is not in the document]
nmm0: I think I originally included it to try and get some of the forward declarations from above. It's not needed but the forward decls are

core/src/View/MDSpan/Kokkos_MDSpan_Extents.hpp
```diff
@@ -0,0 +1,156 @@
...
+
+#ifdef KOKKOS_ENABLE_IMPL_MDSPAN
+
+#include <experimental/mdspan>
+
+namespace Kokkos {
+namespace Impl {
+template <class DataType>
+struct ViewArrayAnalysis;
```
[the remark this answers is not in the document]
nmm0: These are forward declarations that the file needs. I can't include View directly since this file is included by View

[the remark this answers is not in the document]
dalg24: I understand they are needed, I am asking whether you thought about where we would eventually move them when the view refactor is done.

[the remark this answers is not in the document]
nmm0: Since everything will be using mdspan at that point, we could even move them into this header. To keep things organized I think keeping a separate header would be better, but it really only should be included by the mdspan stuff.

core/src/View/MDSpan/Kokkos_MDSpan_Extents.hpp
nmm0: Getting rid of it but was named SizeType after the template param in the paper

dalg24: Not sure what version of the paper you used but it has be changed to index type.  See https://eel.is/c++draft/mdspan.extents

core/src/View/MDSpan/Kokkos_MDSpan_Extents.hpp
nmm0: Removing it

core/unit_test/CMakeLists.txt
```diff
@@ -202,6 +202,7 @@ foreach(Tag Threads;Serial;OpenMP;Cuda;HPX;OpenMPTarget;OpenACC;HIP;SYCL)
       ViewLayoutStrideAssignment
       ViewMapping_b
       ViewMapping_subview
+      ViewMDSpan
```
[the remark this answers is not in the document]
nmm0: Future PRs should add to the tests -- I can make these asserts compile-only but should they be separate from runtime MDSpan tests?

core/src/View/MDSpan/Kokkos_MDSpan_Extents.hpp
```diff
@@ -0,0 +1,143 @@
...
+  static constexpr std::size_t value = std::size_t{0};
+};
+
+template <class Dimension, class Indices>
+struct ExtentsFromDimension;
+
+template <class Dimension, std::size_t... Indices>
+struct ExtentsFromDimension<Dimension, std::index_sequence<Indices...>> {
+  using type           = std::experimental::extents<
+      std::size_t,  // Mirrors Kokkos::View's size type
```
crtrott, line 109: the sizetype depends in Kokkos::View on the memory space, so we probably want to pass this in. I.e. ExtentsFromDimension takes the size_type as an argument.

crtrott (changes requested):

core/src/View/MDSpan/Kokkos_MDSpan_Extents.hpp
crtrott: yeah lol now I am contradicting Damien, but I think I am right :-)

dalg24: except that the memory space information is not available so there is no point on exposing

core/unit_test/CMakeLists.txt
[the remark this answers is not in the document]
nmm0: Moved into subdirectory, renamed, and added to compile only tests like we discussed

core/src/View/MDSpan/Kokkos_MDSpan_Extents.hpp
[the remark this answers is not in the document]
nmm0: Commented this in the file

core/src/View/MDSpan/Kokkos_MDSpan_Extents.hpp
nmm0: Added as a template parameter to `ExtentsFromDataType` and it is propagated into this. That way when creating an extents from DataType, you have to pass in the IndexType

nmm0:
Rebased on develop

core/src/View/MDSpan/Kokkos_MDSpan_Extents.hpp
```diff
@@ -0,0 +1,148 @@
...
+              "Including non-public Kokkos header files is not allowed.");
+#else
+KOKKOS_IMPL_WARNING("Including non-public Kokkos header files is not allowed.")
+#endif
+#endif
+
+#ifndef KOKKOS_EXPERIMENTAL_MDSPAN_EXTENTS_HPP
+#define KOKKOS_EXPERIMENTAL_MDSPAN_EXTENTS_HPP
+
+#include <experimental/mdspan>
```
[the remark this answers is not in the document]
nmm0: Yes, maybe it should be delegated to a common headers such as Kokkos_MDSpan.hpp so the logic stays consistent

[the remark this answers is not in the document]
nmm0: Added a new header that has this include logic

nmm0:
rebased

core/unit_test/TestMDSpan.hpp
```diff
@@ -49,14 +49,6 @@
...
 
 #ifdef KOKKOS_ENABLE_IMPL_MDSPAN
-// Look for the right mdspan
-#if __has_include(<mdspan>)
-#include <mdspan>
-namespace mdspan_ns = std;
-#else
-#include <experimental/mdspan>
-namespace mdspan_ns = std::experimental;
-#endif
```
nmm0, line 59: Note that this is including mdspan transitively now

core/src/View/MDSpan/Kokkos_MDSpan_Extents.hpp
```diff
@@ -0,0 +1,148 @@
...
+              "Including non-public Kokkos header files is not allowed.");
+#else
+KOKKOS_IMPL_WARNING("Including non-public Kokkos header files is not allowed.")
+#endif
+#endif
+
+#ifndef KOKKOS_EXPERIMENTAL_MDSPAN_EXTENTS_HPP
+#define KOKKOS_EXPERIMENTAL_MDSPAN_EXTENTS_HPP
+
+#include "Kokkos_MDSpan.hpp"
```
dalg24, line 58: The header name is really unfortunate because one might expect that `<Kokkos_MDSpan.hpp>` is the header to include to get the whole enchilada, that it will transitively include `Kokkos_MDSpan.hpp` and not the opposite.

core/src/View/MDSpan/Kokkos_MDSpan_Extents.hpp
```diff
@@ -0,0 +1,148 @@
...
+
+namespace Kokkos::Impl {
+/*
+ * Forward declarations from impl/Kokkos_ViewMapping.hpp
+ * We cannot include directly since ViewMapping is used elsewhere in View.
+ * After View is fully moved to mdspan we can include it only from here.
+ */
```
dalg24, line 65: Please avoid C-style comment, prefer `//`

core/unit_test/view/TestExtentsDatatypeConversion.cpp
```diff
@@ -0,0 +1,70 @@
...
+                       std::size_t, std::experimental::dynamic_extent,
+                       std::experimental::dynamic_extent, std::size_t{3},
+                       std::size_t{2}, std::size_t{8}>>);
+
+// Conversion from extents to DataType
+// 0-rank extents
+static_assert(
+    std::is_same_v<double,
+                   typename Kokkos::Experimental::Impl::DataTypeFromExtents<
+                       double, std::experimental::extents<std::size_t>>::type>);
```
dalg24, line 40: Introduce a macro or a trait so the test read something
```C++
CHECK_DATA_TYPE_TO_EXTENT(double, extents<std::size_t>);
CHECK_DATA_TYPE_TO_EXTENT(double***, dextents<std::size_t, 3>);
```
essentially eliminates all the boiler plate code and let the reader focus on what is being mapped to what

core/src/View/MDSpan/Kokkos_MDSpan_Extents.hpp
masterleinad: I would also prefer if `Kokkos_View.hpp` includes `Kokkos_MDSpan.hpp` and `Kokkos_MDSpan.hpp` includes `Kokkos_MDSpan_Extents.hpp`.

nmm0: I renamed it to `Kokkos_MDSpan_Header.hpp` but am definitely open to other suggestions :P
- I didn't switch the include order but if we do want to add a `Kokkos_MDSpan.hpp`, what @masterleinad is saying would make sense.

core/src/View/MDSpan/Kokkos_MDSpan_Extents.hpp
nmm0: Done

core/unit_test/view/TestExtentsDatatypeConversion.cpp
nmm0: I used a helper template instead of a macro, hopefully it's clearer.

nmm0:
rebased

nmm0:
rebased

core/src/View/MDSpan/Kokkos_MDSpan_Header.hpp
```diff
@@ -0,0 +1,67 @@
...
+*/
+
+#ifndef KOKKOS_IMPL_PUBLIC_INCLUDE
+#include <Kokkos_Macros.hpp>
+#ifndef KOKKOS_ENABLE_DEPRECATED_CODE_3
+static_assert(false,
+              "Including non-public Kokkos header files is not allowed.");
+#else
+KOKKOS_IMPL_WARNING("Including non-public Kokkos header files is not allowed.")
+#endif
```
masterleinad, line 52: We removed this condition from all the other files, so I would just drop it here (and in the file above_ as well
```suggestion
static_assert(false,
              "Including non-public Kokkos header files is not allowed.");
```

core/src/View/MDSpan/Kokkos_MDSpan_Extents.hpp
```diff
@@ -0,0 +1,144 @@
...
+*/
+
+#ifndef KOKKOS_IMPL_PUBLIC_INCLUDE
+#include <Kokkos_Macros.hpp>
+#ifndef KOKKOS_ENABLE_DEPRECATED_CODE_3
+static_assert(false,
+              "Including non-public Kokkos header files is not allowed.");
+#else
+KOKKOS_IMPL_WARNING("Including non-public Kokkos header files is not allowed.")
+#endif
```
masterleinad, line 52:
```suggestion
static_assert(false,
              "Including non-public Kokkos header files is not allowed.");
```

core/src/View/MDSpan/Kokkos_MDSpan_Header.hpp
nmm0: done

core/src/View/MDSpan/Kokkos_MDSpan_Extents.hpp
nmm0: done

masterleinad (approved):

crtrott (changes requested):
Can we change this to pass the index_type in: we later want to use the index type from the execution/memory space.

crtrott (approved):

PR #2024, opened 2019-03-11
avoid instantiating default constructor of value type when WithoutInitializing is given

dhollman:
should address #1873

PR #5307, opened 2022-08-03
Avoid instantiating default constructor of value type when WithoutInitializing is given

### LINKED ISSUE #1873: Support creation of managed views of classes without a device default constructor when asking for no-initialization
In particular this is coming up for UniqueToken, but the issue should apply to any type whose default constructor is host-only that is being stored in a View. Code like:
```cpp
// Allocate space for unique tokens, do not initialize because the default constructor is not GPU-safe
View<UniqueToken<Cuda>*, CudaSpace> unique_tokens(view_alloc("unique_tokens", Kokkos::WithoutInitializing), size); // Warning here
auto host_mirror = create_mirror_view(unique_tokens);
// Code to create the tokens in the host_mirror and deep_copy them over to the GPU follows here
```
causes warnings about calling a host function from a host device function with nvcc. The warnings show up on the marked line from the functor call that would initialize the values in unique_tokens if we weren’t using Kokkos::WithoutInitializing. Looking at Kokkos_ViewMapping.hpp I see that that functor call is just protected by a regular if statement that checks the (compile time known) value of alloc_prop::initialize, thus the warning.
 
How much would it hurt compile times, maintenance for you guys, etc to move that initialization functor call so that it does not get compiled when we use WithoutInitializing to get rid of the warning?

masterleinad:
Trying #2024 on top of #5277. Fixes #1873.

### DIFF
--- a/core/src/impl/Kokkos_ViewArray.hpp
+++ b/core/src/impl/Kokkos_ViewArray.hpp
@@ -392,25 +392,31 @@ class ViewMapping<Traits, Kokkos::Array<>> {
 
     m_impl_handle = handle_type(reinterpret_cast<pointer_type>(record->data()));
 
-    if constexpr (alloc_prop::initialize) {
+    functor_type functor =
+        execution_space_specified
+            ? functor_type(exec_space, (pointer_type)m_impl_handle,
+                           m_impl_offset.span() * Array_N, alloc_name)
+            : functor_type((pointer_type)m_impl_handle,
+                           m_impl_offset.span() * Array_N, alloc_name);
+
+    if (false) {
+      // Make sure the destroy functor gets instantiated.
+      // This avoids "cudaErrorInvalidDeviceFunction"-type errors.
+      functor.destroy_shared_allocation();
+    }
+
+    //  Only initialize if the allocation is non-zero.
+    //  May be zero if one of the dimensions is zero.
+    if constexpr (alloc_prop::initialize)
       if (alloc_size) {
-        // The functor constructs and destroys
-        record->m_destroy =
-            execution_space_specified
-                ? functor_type(exec_space, (pointer_type)m_impl_handle,
-                               m_impl_offset.span() * Array_N, alloc_name)
-                : functor_type((pointer_type)m_impl_handle,
-                               m_impl_offset.span() * Array_N, alloc_name);
+        // Assume destruction is only required when construction is requested.
+        // The ViewValueFunctor has both value construction and destruction
+        // operators.
+        record->m_destroy = std::move(functor);
 
         // Construct values
-        if (false) {
-          // Make sure the destroy functor gets instantiated.
-          // This avoids "cudaErrorInvalidDeviceFunction"-type errors.
-          record->m_destroy.destroy_shared_allocation();
-        }
         record->m_destroy.construct_shared_allocation();
       }
-    }
 
     return record;
   }
--- a/core/src/impl/Kokkos_ViewArray.hpp
+++ b/core/src/impl/Kokkos_ViewArray.hpp
@@ -399,11 +399,14 @@ class ViewMapping<Traits, Kokkos::Array<>> {
             : functor_type((pointer_type)m_impl_handle,
                            m_impl_offset.span() * Array_N, alloc_name);
 
+#if defined(KOKKOS_ENABLE_CUDA) || defined(KOKKOS_ENABLE_HIP) || \
+    defined(KOKKOS_ENABLE_SYCL) || defined(KOKKOS_ENABLE_OPENMPTARGET)
     if (false) {
       // Make sure the destroy functor gets instantiated.
       // This avoids "cudaErrorInvalidDeviceFunction"-type errors.
       functor.destroy_shared_allocation();
     }
+#endif
 
     //  Only initialize if the allocation is non-zero.
     //  May be zero if one of the dimensions is zero.
--- a/core/src/impl/Kokkos_ViewArray.hpp
+++ b/core/src/impl/Kokkos_ViewArray.hpp
@@ -390,11 +390,10 @@ class ViewMapping<Traits, Kokkos::Array<>> {
                                     alloc_size)
             : record_type::allocate(mem_space, alloc_name, alloc_size);
 
-    if (alloc_size) {
-      m_impl_handle =
-          handle_type(reinterpret_cast<pointer_type>(record->data()));
+    m_impl_handle = handle_type(reinterpret_cast<pointer_type>(record->data()));
 
-      if (alloc_prop::initialize) {
+    if constexpr (alloc_prop::initialize) {
+      if (alloc_size) {
         // The functor constructs and destroys
         record->m_destroy =
             execution_space_specified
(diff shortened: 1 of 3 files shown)


core/src/impl/Kokkos_ViewMapping.hpp
```diff
@@ -2978,19 +2980,22 @@ struct ViewValueFunctor<DeviceType, ValueType, false /* is_scalar */> {
...
       uint64_t kpID = 0;
       if (Kokkos::Profiling::profileLibraryLoaded()) {
         functor_name =
-            (destroy ? "Kokkos::View::destruction [" + functor_name + "]"
-                     : "Kokkos::View::initialization [" + functor_name + "]");
+            (std::is_same_v<Tag, DestroyTag>
+                 ? "Kokkos::View::destruction [" + functor_name + "]"
+                 : "Kokkos::View::initialization [" + functor_name + "]");
         Kokkos::Profiling::beginParallelFor(
             "Kokkos::View::initialization [" + functor_name + "]",
```
PhilMiller, line 3000: It looks like this should just be `functor_name` now, since initialization/destruction was specified in it just above

core/src/impl/Kokkos_ViewMapping.hpp
```diff
@@ -3460,22 +3465,31 @@ class ViewMapping<
...
+            ? functor_type(exec_space, (value_type*)m_impl_handle,
+                           m_impl_offset.span(), alloc_name)
+            : functor_type((value_type*)m_impl_handle, m_impl_offset.span(),
+                           alloc_name);
+
+    if (false) {
+      // Make sure the destroy functor gets instantiated.
+      // This avoids "cudaErrorInvalidDeviceFunction"-type errors.
+      functor.destroy_shared_allocation();
+    }
```
masterleinad, line 3479: This is the critical change that we missed in the previous attempts.

core/src/impl/Kokkos_ViewMapping.hpp
```diff
@@ -3460,22 +3465,31 @@ class ViewMapping<
...
+        execution_space_specified
+            ? functor_type(exec_space, (value_type*)m_impl_handle,
+                           m_impl_offset.span(), alloc_name)
+            : functor_type((value_type*)m_impl_handle, m_impl_offset.span(),
+                           alloc_name);
+
+    if (false) {
+      // Make sure the destroy functor gets instantiated.
+      // This avoids "cudaErrorInvalidDeviceFunction"-type errors.
+      functor.destroy_shared_allocation();
```
PhilMiller, line 3478: Is this not going to generate warnings about unreachable code from one compiler or another? How about `(void)&functor_type::destroy_shared_allocation`?

masterleinad: I would try to deal with it when we run into problems. At least our CI seems to be fine with it.

core/src/impl/Kokkos_ViewMapping.hpp
```diff
@@ -3460,22 +3465,31 @@ class ViewMapping<
...
     m_impl_handle = handle_type(reinterpret_cast<pointer_type>(record->data()));
 
+    functor_type functor =
+        execution_space_specified
+            ? functor_type(exec_space, (value_type*)m_impl_handle,
+                           m_impl_offset.span(), alloc_name)
+            : functor_type((value_type*)m_impl_handle, m_impl_offset.span(),
+                           alloc_name);
+
+    if (false) {
```
dalg24, line 3475: Please guard that code so it is only there when absolutely necessary.
This will be helpful if one tries to revisit later.

dalg24 (commented):
I am glad you figured that one out

core/src/impl/Kokkos_ViewArray.hpp
```diff
@@ -390,22 +390,33 @@ class ViewMapping<Traits, Kokkos::Array<>> {
...
+    }
+
+    //  Only initialize if the allocation is non-zero.
+    //  May be zero if one of the dimensions is zero.
+    if constexpr (alloc_prop::initialize)
+      if (alloc_size) {
+        // Assume destruction is only required when construction is requested.
+        // The ViewValueFunctor has both value construction and destruction
+        // operators.
+        record->m_destroy = std::move(functor);
```
dalg24, line 415: I see that is what we were currently doing but does that mean we don't setup for destructing the elements when a view was constructed specifying the "without initializing"?
Is that a potential issue?

core/src/impl/Kokkos_ViewMapping.hpp
masterleinad: I need the code outside the `if constexpr` block so that I don't run into the `cudaErrorInvalidDeviceFunction` error in the new test.

dalg24: I meant macro guards

core/src/impl/Kokkos_ViewArray.hpp
masterleinad: Yes, we only call the destructor for non-`std::is_scalar`-types and on those only if the `View` constructor also initializes the elements. This design dates back all the way to 95bc2e1657a1f77d725721b1774cb7715ab22907 but it doesn't really give a justification for this assumption. I think it's a question to be discussed separately from this pull request.

masterleinad:
Retest this please.

dalg24 (approved):

core/src/impl/Kokkos_ViewArray.hpp
masterleinad: We discussed on Slack that we want to keep the current behavior.

core/src/impl/Kokkos_ViewArray.hpp
```diff
@@ -390,22 +390,33 @@ class ViewMapping<Traits, Kokkos::Array<>> {
...
+    m_impl_handle = handle_type(reinterpret_cast<pointer_type>(record->data()));
 
+    functor_type functor =
+        execution_space_specified
+            ? functor_type(exec_space, (pointer_type)m_impl_handle,
+                           m_impl_offset.span() * Array_N, alloc_name)
+            : functor_type((pointer_type)m_impl_handle,
+                           m_impl_offset.span() * Array_N, alloc_name);
+
+    if (false) {
```
dalg24, line 402: You forgot to guard that one

core/src/impl/Kokkos_ViewMapping.hpp
```diff
@@ -2892,25 +2892,29 @@ struct ViewValueFunctor;
...
-      new (ptr + i) ValueType();
-    }
+      std::enable_if_t<std::is_default_constructible<_ValueType>::value>
+      operator()(ConstructTag const&, const size_t i) const {
+    new (ptr + i) ValueType();
+  }
+
+  KOKKOS_INLINE_FUNCTION void operator()(DestroyTag const&,
+                                         const size_t i) const {
+    // KOKKOS_IMPL_CUDA_CLANG_WORKAROUND this line causes ptax error
```
crtrott, line 2915: does this comment still has any meaning? Not holding this up because of it but would be good to figure out. I also see that one of hte tests is commented out for HIP too ?

crtrott (changes requested):
Just guard that thing in the ViewArray version too then I approve. The clang stuff we should look at independently.

core/src/impl/Kokkos_ViewMapping.hpp
masterleinad: As far as I can see, the test is also enabled for Cuda+Clang so I removed the comment.

masterleinad:
> Just guard that thing in the ViewArray version too then I approve. The clang stuff we should look at independently.

See https://github.com/kokkos/kokkos/pull/5307/commits/3f3bbc2ec0680d6707dca986197beb737ff5f953.

masterleinad:
Retest this please.

masterleinad:
Only `KokkosCore_UnitTest_CudaTimingBased` is failing in `CUDA-11.6-NVHPC`.

crtrott (approved):

Issue #5290, opened 2022-07-29
Trilinos compilation errors, Intel Classic Compilers (19,2021.3) + OMP: internal compiler error 101003_1112 with Sacado, Intrepid2 unit tests


ndellingwood:
**Describe the bug**
Building Trilinos with intel/19 and the OpenMP backend results in ICE in the Sacado unit test `Trilinos/packages/sacado/test/UnitTests/Fad_Fad_KokkosTests_OpenMP.cpp`

```
[ 28%] Building CXX object packages/sacado/test/UnitTests/CMakeFiles/Sacado_FadFadKokkosTests_OpenMP.dir/Fad_Fad_KokkosTests_OpenMP.cpp.o
[...]

PR #5311, opened 2022-08-04
Increase intel version for workaround macro define

ndellingwood:
KOKKOS_IMPL_WORKAROUND_ICE_IN_TRILINOS_WITH_OLD_INTEL_COMPILERS
This change is a workaround for ICE tested with intel/2021.2.0, similar
to that seen in kokkos/kokkos#5290

### DIFF
--- a/core/src/Kokkos_Macros.hpp
+++ b/core/src/Kokkos_Macros.hpp
@@ -228,8 +228,8 @@
 #define KOKKOS_ENABLE_PRAGMA_SIMD 1
 #endif
 
-// FIXME Workaround for ICE with intel 17,18,19 in Trilinos
-#if (KOKKOS_COMPILER_INTEL <= 1900)
+// FIXME Workaround for ICE with intel 17,18,19,20,21 in Trilinos
+#if (KOKKOS_COMPILER_INTEL <= 2100)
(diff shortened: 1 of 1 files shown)


core/src/Kokkos_Macros.hpp
```diff
@@ -228,8 +228,8 @@
 #define KOKKOS_ENABLE_PRAGMA_SIMD 1
 #endif
 
-// FIXME Workaround for ICE with intel 17,18,19 in Trilinos
-#if (KOKKOS_COMPILER_INTEL <= 1900)
+// FIXME Workaround for ICE with intel 17,18,19,20,21 in Trilinos
+#if (KOKKOS_COMPILER_INTEL <= 2100)
```
dalg24, line 232: Not objecting to the change but wondering if that actually means for all Intel C++ compilers currently released.

masterleinad: `Intel(R) oneAPI DPC++/C++ Compiler 2022.1.0 (2022.1.0.20220316` has `__INTEL_LLVM_COMPILER=20220100`.

ndellingwood: I haven't tested with the 2022 compilers, it's possible the issue is present for that version as well

dalg24 (approved):

masterleinad (approved):

PR #4721, opened 2022-01-26
Add m_stride to layout to support propagation of Padding

janciesko:
Propagate padding when `create_mirror_view` and source view has `Kokkos::AllowPadding`.

Case in point:
```
auto alloc = Kokkos::view_alloc("test", Kokkos::AllowPadding);
auto v = Kokkos::View<double**, Kokkos::CudaSpace>(alloc, 17, 4);
auto h_v = Kokkos::create_mirror_view(v);
ASSERT_EQ(v.stride(1), h_v.stride(1));
```

Fixes #2995

PR #5314, opened 2022-08-05
Remove dummy arguments for ViewCtorProp

masterleinad:
This pull request removes support for the dummy `std::integral_constant<unsigned, I>` `ViewCtor` properties and replaces the mechanism to add properties (that weren't existent already) with a separate function.
This avoids writing much of the boilerplate code and also avoids finding an unused value for `I`.


nliber (approved):

janciesko:
Very nice

janciesko:
https://github.com/kokkos/kokkos/blob/176d5fce6be6ed78613ea62de43d863c4ac23a82/core/src/Kokkos_CopyViews.hpp#L3587-L3590
https://github.com/kokkos/kokkos/blob/176d5fce6be6ed78613ea62de43d863c4ac23a82/core/src/Kokkos_CopyViews.hpp#L3587-L3590

Can be replaced as well.

core/src/impl/Kokkos_ViewCtor.hpp
```diff
@@ -273,6 +260,32 @@ struct ViewCtorProp : public ViewCtorProp<void, P>... {
...
+                    const Property &property,
+                    const Properties &... properties) {
+  if constexpr ((is_execution_space<Property>::value &&
+                 !ViewCtorProp<P...>::has_execution_space) ||
+                (is_memory_space<Property>::value &&
+                 !ViewCtorProp<P...>::has_memory_space) ||
+                (is_view_label<Property>::value &&
+                 !ViewCtorProp<P...>::has_label) ||
+                (std::is_same_v<Property, WithoutInitializing_t> &&
+                 ViewCtorProp<P...>::initialize)) {
```
masterleinad, line 279: I would add more properties here as needed (such as in https://github.com/kokkos/kokkos/pull/4721).

core/unit_test/TestExecutionSpace.hpp
```diff
@@ -54,10 +54,9 @@ struct CheckClassWithExecutionSpaceAsDataMemberIsCopyable {
   Kokkos::DefaultHostExecutionSpace host;
 
   KOKKOS_FUNCTION void operator()(int, int& e) const {
-    auto copy = *this;
```
masterleinad, line 57: This produced problems in the RDC build such as
```
nvlink error   : Undefined reference to '_ZN6Kokkos4Impl13HostSharedPtrINS0_12CudaInternalEEC1ERKS3_' in 'CMakeFiles/KokkosCore_UnitTest_Cuda1.dir/cuda/TestCuda_ExecutionSpace.cpp.o'

make[2]: *** [core/unit_test/CMakeFiles/KokkosCore_UnitTest_Cuda1.dir/build.make:905: core/unit_test/KokkosCore_UnitTest_Cuda1] Error 255
```

janciesko: This was the issue we saw previously?

masterleinad: Yes, and I had to use two calls for MSVC in another place.

core/src/Kokkos_View.hpp
```diff
@@ -1402,22 +1402,13 @@ class View : public ViewTraits<DataType, Properties...> {
...
-        std::conditional_t<alloc_prop_input::has_execution_space,
-                           std::integral_constant<unsigned int, 2>,
-                           typename traits::device_type::execution_space>>;
+    // Append layout and spaces if not input, split for MSVC
+    using alloc_prop_tmp =
+        decltype(Impl::add_properties(arg_prop, std::string{}));
+    using alloc_prop = decltype(
+        Impl::add_properties(std::declval<alloc_prop_tmp>(),
+                             typename traits::device_type::memory_space{},
+                             typename traits::device_type::execution_space{}));
```
PhilMiller, line 1411: Why does this go through defining and extending `alloc_prop_tmp`? It looks like all of these properties can be added in one shot, like you did in `OffsetView`

PhilMiller (approved):

core/src/impl/Kokkos_ViewCtor.hpp
```diff
@@ -273,6 +260,32 @@ struct ViewCtorProp : public ViewCtorProp<void, P>... {
...
 
+template <typename... P>
+auto add_properties(const ViewCtorProp<P...> &view_ctor_prop) {
+  return view_ctor_prop;
+}
+
+template <typename... P, typename Property, typename... Properties>
+auto add_properties(const ViewCtorProp<P...> &view_ctor_prop,
+                    const Property &property,
+                    const Properties &... properties) {
```
dalg24, line 271: Why should we provide a variadic function?
Is adding a single property not sufficient?

core/src/impl/Kokkos_ViewCtor.hpp
```diff
@@ -273,6 +260,32 @@ struct ViewCtorProp : public ViewCtorProp<void, P>... {
...
+                 !ViewCtorProp<P...>::has_label) ||
+                (std::is_same_v<Property, WithoutInitializing_t> &&
+                 ViewCtorProp<P...>::initialize)) {
+    using NewViewCtorProp = ViewCtorProp<P..., Property>;
+    NewViewCtorProp new_view_ctor_prop(view_ctor_prop);
+    static_cast<ViewCtorProp<void, Property> &>(new_view_ctor_prop).value =
+        property;
+    return add_properties(new_view_ctor_prop, properties...);
+  } else
+    return add_properties(view_ctor_prop, properties...);
```
dalg24, line 286: Why is silently ignoring the property the right thing to do in the else-clause?

core/src/Kokkos_View.hpp
```diff
@@ -1402,22 +1402,13 @@ class View : public ViewTraits<DataType, Properties...> {
...
-                           typename traits::device_type::memory_space>,
-        std::conditional_t<alloc_prop_input::has_execution_space,
-                           std::integral_constant<unsigned int, 2>,
-                           typename traits::device_type::execution_space>>;
+    // Append layout and spaces if not input, split for MSVC
+    using alloc_prop_tmp =
+        decltype(Impl::add_properties(arg_prop, std::string{}));
+    using alloc_prop = decltype(
+        Impl::add_properties(std::declval<alloc_prop_tmp>(),
+                             typename traits::device_type::memory_space{},
```
dalg24, line 1410: What does it even mean adding a memory space?

dalg24 (changes requested):
Need to clarify the design

dalg24:
Also this does not eliminate all of the multiple inheritance cast to some arbitrary parent trick to access property value

core/src/impl/Kokkos_ViewCtor.hpp
PhilMiller: Perhaps the function should be called "add default properties"

PhilMiller: Though that only makes sense for the already present case, not the unrecognized argument type case.

dalg24:
Consider having simple abstractions you can compose with such as
* `has_property() -> bool` to check whether `ViewCtorProp` has some given property.  Analogous to the `has_execution_space` or `has_label` members but may come handy in generic context.  Maybe it is not needed, haven't thought much thought into it.
* `get_property() -> PROPERTY const&` to access the underlying property (requires that `has_property()` returns `true`.  Encapsulates the ugly cast to parent trick.
* `set_property() -> ViewCtorProp` that add/overwrite the underlying property

core/src/Kokkos_View.hpp
masterleinad: I tried to say in the comment above that MSVC was giving me errors when defining the alias in one shot.

core/src/Kokkos_View.hpp
masterleinad: This is used for the case of `create_mirror[_view]` to specify the memory space (type) we want a `View` to be allocated in.

core/src/impl/Kokkos_ViewCtor.hpp
masterleinad: The design is motivated by the code to be replaced. We were previously writing stuff like
```
using alloc_prop = Kokkos::Impl::ViewCtorProp<
        P...,
        std::conditional_t<alloc_prop_input::has_label,
                           std::integral_constant<unsigned, 0>, std::string>,
        std::conditional_t<alloc_prop_input::has_memory_space,
                           std::integral_constant<unsigned, 1>,
                           typename traits::device_type::memory_space>,
        std::conditional_t<alloc_prop_input::has_execution_space,
                           std::integral_constant<unsigned, 2>,
                           typename traits::device_type::execution_space>>;
```
and it seems sensible to me to continue supporting to add all desired properties in one go. Adding a single property would be sufficient but a variadic function is convenient and allows for more concise code.

core/src/impl/Kokkos_ViewCtor.hpp
masterleinad: Again, the code that is to be replaced by this function oftentimes expresses "add some property if it doesn't exist already" and that's the reason for ignoring properties in this branch. I'm open to changing the name of the function, of course.

core/src/Kokkos_View.hpp
dalg24: How so?  Isn't the memory space being deduced from the view traits analysis?

core/src/Kokkos_View.hpp
PhilMiller: Could you add a comment to that effect in the code? Otherwise, people may try to refactor it and be disappointed

janciesko (approved):

core/src/Kokkos_View.hpp
masterleinad: I was talking about the overloads taking a memory space argument (possibly included in the `ViewCtorProp` argument or explicitly) for the returned view.

masterleinad:
> Consider having simple abstractions 

I added a `get_property` function that allows to "[...] eliminate all of the multiple inheritance cast to some arbitrary parent trick to access property value" but that could also easily be a follow-up.

containers/src/Kokkos_DynRankView.hpp
```diff
@@ -1100,20 +1100,9 @@ class DynRankView : public ViewTraits<DataType, Properties...> {
...
-                           std::integral_constant<unsigned, 0>, std::string>,
-        std::conditional_t<alloc_prop_input::has_memory_space,
-                           std::integral_constant<unsigned, 1>,
-                           typename traits::device_type::memory_space>,
-        std::conditional_t<alloc_prop_input::has_execution_space,
-                           std::integral_constant<unsigned, 2>,
-                           typename traits::device_type::execution_space>>;
+    using alloc_prop = decltype(Impl::add_properties(
+        arg_prop, std::string{}, typename traits::device_type::memory_space{},
+        typename traits::device_type::execution_space{}));
```
PhilMiller, line 1105: How about
```suggestion
    // Copy the input allocation properties with possibly defaulted properties
    auto prop_copy = Impl::add_properties(
        arg_prop, std::string{}, typename traits::device_type::memory_space{},
        typename traits::device_type::execution_space{});
    using alloc_prop = decltype(prop_copy);
```

It seems to read more cleanly to actually call the function, and then query the type of the resulting object, than to come up with the type based on a hypothetical call, and construct an instance of that

PhilMiller: Most of the other call sites are already doing this

containers/src/Kokkos_OffsetView.hpp
```diff
@@ -1201,21 +1201,9 @@ class OffsetView : public ViewTraits<DataType, Properties...> {
...
-                           std::integral_constant<unsigned, 0>, std::string>,
-        std::conditional_t<alloc_prop_input::has_memory_space,
-                           std::integral_constant<unsigned, 1>,
-                           typename traits::device_type::memory_space>,
-        std::conditional_t<alloc_prop_input::has_execution_space,
-                           std::integral_constant<unsigned, 2>,
-                           typename traits::device_type::execution_space>>;
+    using alloc_prop = decltype(::Kokkos::Impl::add_properties(
+        arg_prop, std::string{}, typename traits::device_type::memory_space{},
+        typename traits::device_type::execution_space{}));
```
PhilMiller, line 1206:
```suggestion
    // Copy the input allocation properties with possibly defaulted properties
    auto prop_copy = ::Kokkos::Impl::add_properties(
        arg_prop, std::string{}, typename traits::device_type::memory_space{},
        typename traits::device_type::execution_space{});
    using alloc_prop = decltype(prop_copy);
```

core/src/Kokkos_View.hpp
PhilMiller: Ditto the suggestion above of creating `prop_copy` and using `decltype(prop_copy)` from above - does that also avoid the MSVC errors?

PhilMiller:
I think the helper functions are a big improvement

PhilMiller (approved):

masterleinad:
Jenkins CI passe din https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/Kokkos/detail/Kokkos/10274/pipeline but doesn't show here.

masterleinad:
Retest this please.

masterleinad:
Retest this please.

masterleinad:
@dalg24 ping

crtrott (changes requested):
Hey, we definitely need a name change: we are not adding a property unconditionally here, its more like: `add_properties_if_not_present` or `add_properties_if_unset`. Also we are not adding it to the existing one: so maybe actually `with_properties_if_unset` is even better.

Longer term we need to clean up the design a bit more, like actually building the "attach multiple properties" being build from "attach one property".

core/src/impl/Kokkos_ViewCtor.hpp
```diff
@@ -325,29 +325,9 @@ KOKKOS_FUNCTION const auto &get_property(
...
-    using pointer_type = typename ViewCtorProp<P...>::pointer_type;
-    return static_cast<ViewCtorProp<void, pointer_type> &>(view_ctor_prop)
-        .value;
-  } else {
-    static_assert(std::is_same_v<Tag, void>, "Invalid property tag!");
-    return view_ctor_prop;
-  }
+  const ViewCtorProp<P...> &tmp = view_ctor_prop;
+  const auto &tmp_return        = get_property<Tag>(tmp);
+  return const_cast<std::decay_t<decltype(tmp_return)> &>(tmp_return);
```
dalg24, line 330: Is that a joke?

dalg24 (changes requested):

core/src/Kokkos_CopyViews.hpp
```diff
@@ -3278,24 +3246,15 @@ impl_realloc(Kokkos::View<T, P...>& v, const size_t n0, const size_t n1,
...
-                             std::integral_constant<unsigned int, 2>,
-                             typename view_type::execution_space>>;
-      alloc_prop arg_prop_copy(arg_prop);
-      auto const& exec_space = static_cast<Kokkos::Impl::ViewCtorProp<
-          void, typename alloc_prop::execution_space> const&>(arg_prop_copy)
-                                   .value;
+    if constexpr (alloc_prop_input::has_execution_space) {
+      using execution_space_type = typename alloc_prop_input::execution_space;
+      const execution_space_type& exec_space =
+          Impl::get_property<Impl::ExecutionSpaceTag>(arg_prop);
```
PhilMiller, line 3257: Using `auto` there's no need to extract the type and name it locally

core/src/impl/Kokkos_ViewCtor.hpp
crtrott: I don't think we should have a non-const reference version of get_property in the first place. I don't think there is any reason for this ting to not be immutable?

crtrott (changes requested):

core/src/Kokkos_View.hpp
```diff
@@ -1402,22 +1402,14 @@ class View : public ViewTraits<DataType, Properties...> {
...
-                           std::integral_constant<unsigned int, 0>,
-                           std::string>,
-        std::conditional_t<alloc_prop_input::has_memory_space,
-                           std::integral_constant<unsigned int, 1>,
-                           typename traits::device_type::memory_space>,
-        std::conditional_t<alloc_prop_input::has_execution_space,
-                           std::integral_constant<unsigned int, 2>,
-                           typename traits::device_type::execution_space>>;
+    // Copy the input allocation properties with possibly defaulted properties
+    // We need to split it into to avoid MSVC compiler errors
```
PhilMiller, line 1406: "split it in two"?

core/src/impl/Kokkos_ViewCtor.hpp
dalg24, line 330:
```suggestion
  // Avoid code duplication by deferring to the const-qualified getter and
  // casting the const away
  const auto &tmp = get_property<Tag>(view_ctor_prop);
  return const_cast<std::decay_t<decltype(tmp)> &>(tmp);
```

masterleinad:
> I don't think we should have a non-const reference version of get_property in the first place. I don't think there is any reason for this thing to not be immutable?

@crtrott https://github.com/masterleinad/kokkos/blob/11842ae0f15b6fe2563b18c106f5177cfcd86570/core/src/Kokkos_CopyViews.hpp#L3820-L3821 is an example where we modify a property by first extracting it from a `ViewCtorProp<>` argument using `get_propetry`.

core/src/impl/Kokkos_ViewCtor.hpp
dalg24: We actually need both versions

masterleinad: That would be calling the same function recursively, right? We need to cast `view_ctor_prop` to `const ViewCtorProp<P...> &` in one way or the other. I'm happy to add a comment if that's sufficient for you.

masterleinad: [masterleinad/kokkos@11842ae/core/src/Kokkos_CopyViews.hpp#L3820-L3821](https://github.com/masterleinad/kokkos/blob/11842ae0f15b6fe2563b18c106f5177cfcd86570/core/src/Kokkos_CopyViews.hpp#L3820-L3821) is an example where we modify a property by first extracting it from a ViewCtorProp<> argument using get_propetry.

dalg24: You are right.  Add a comment though

dalg24: (I have always done it the other way around)

PhilMiller: More constructively, would it make sense to define the const version in terms of the non-const version?

masterleinad: [`017a550` (#5314)](https://github.com/kokkos/kokkos/pull/5314/commits/017a550eacfce297a3806fd4c3a3867e4d03c3d7)

dalg24 (approved):
Fine by me

PhilMiller (approved):

crtrott (approved):
Ok I want to think later again about the non-const thingy but for now its fine ...

PR #5316, opened 2022-08-08
Don't set a flag for MSVC using ARCH_NATIVE

masterleinad:
MSVC doesn't have any flag corresponding to `-march=native`.


cmake/kokkos_arch.cmake
```diff
@@ -221,6 +221,7 @@ ENDIF()
 IF(KOKKOS_ARCH_NATIVE)
   COMPILER_SPECIFIC_FLAGS(
     COMPILER_ID KOKKOS_CXX_HOST_COMPILER_ID
+    MSVC    NO-VALUE-SPECIFIED
```
dalg24, line 224: Why is that the right thing to do rather than error out and say the option is not available?
The proposed resolution is likely so surprise users.

crtrott (changes requested):
Just looked up the default behavior of MSVC: it is not native, it is SSE2: 

https://docs.microsoft.com/en-us/cpp/build/reference/arch-x64?view=msvc-170

Hence we should not do this, and rather error out or. Users can specify Kokkos_ARCH_NONE to get nothing set.

masterleinad:
> Hence we should not do this, and rather error out or. 

Changed.

cmake/kokkos_arch.cmake
```diff
@@ -459,7 +463,7 @@ IF (CMAKE_CXX_COMPILER_ID STREQUAL Clang AND WIN32)
 ENDIF()
 
 # MSVC ABI has many deprecation warnings, so ignore them
-IF (CMAKE_CXX_COMPILER_ID STREQUAL MSVC OR "x${CMAKE_CXX_SIMULATE_ID}" STREQUAL "xMSVC")
+IF (CMAKE_CXX_COMPILER_ID STREQUAL "MSVC" OR "x${CMAKE_CXX_SIMULATE_ID}" STREQUAL "xMSVC")
```
masterleinad, line 466: `MSVC` is a variable in `CMake`, see https://cmake.org/cmake/help/latest/variable/MSVC.html. Thus, we have to use quotes here.

dalg24 (approved):

crtrott (approved):

dalg24:
Failure (CUDA UM perf test) is unrelated

Issue #645, opened 2017-02-16
Concurrently sort many (batched) arrays using Kokkos


tawiesn:
I need to sort a large number of small arrays concurrently. For CUDA i can use
thrust::stable_sort_by_key as, e.g., described in the second answer [here](http://stackoverflow.com/questions/8298728/concurrently-sorting-many-arrays-with-cuda-thrust)
How can i do that with Kokkos? Something like thrust::stable_sort_by_key would be useful.

ibaned:
@tawiesn a while back I found that Kokkos didn't quite have a general enough sorting interface and put together something of my own that either calls Thrust or an OpenMP parallel stable sort written by folks at Intel. What this thing does is to accept an array of keys and return a permutation array that can then be used to sort the many small arrays. There is a bit of complication because I allow keys to be composed of multiple entries.

https://github.com/ibaned/omega_h/blob/master/src/sort.cpp
https://github.com/ibaned/omega_h/tree/master/src/intel_sort

Anyway, maybe you can either adapt this code for your purposes or we can work to get a more production-quality version of it included in Kokkos.

crtrott:
Yeah we do not have a sorting thing for sorting multiple small arrays concurrently. 
Though its something I would want. Maybe we can work on implementing something.

tawiesn:
Thanks @ibaned for pointing me to these algorithms. The sorting part is only a minor step in my overall algorithm. So, i will first make sure that everything else is working. Maybe we can then coordinate our efforts to have some general sorting interface in Kokkos algorithms. I'll keep you informed about the progress.

mhoemmen:
@tawiesn @ibaned If it helps, Tpetra has a (currently not parallel) `sort2` function, that sorts one array and applies the resulting permutation to an array of the same length.

Tpetra could use a thread-parallel "sort all the entries in every row of a sparse matrix."  My plan for now was just to do that on the host, with existing non-CUDA-ized functions.

tawiesn:
@mhoemmen Thanks for the hint. I'll take a look at it. Currently i'm using a non-parallel sort version for my kernels in MueLu, too, playing around with the user interface.

mhoemmen:
@tawiesn If you come up with a good interface, let me know :-)

gmackey:
[...]

PR #4577, opened 2021-12-02
Backport bit manipulation facility from C++20 standard library header `<bit>`

dalg24:
~~Fix #3828~~

**See documentation in kokkos/kokkos-core-wiki#300**

* ~~Deprecate `Kokkos::log2(unsigned) -> int`~~ done in #4595
* ~~Provide alternative `Impl::bit_log2(UnsignedInteger) -> UnsignedInteger` for internal use only~~ withdrawn (might come back with it in a follow up PR)
* Backport bit manipulation facility from C++20 https://eel.is/c++draft/bit
* ~~Remove `Impl::bit_count(unsigned int)` in favor of `Experimental::popcount(UnsignedInteger) -> UnsignedInteger`~~ withdrawn (will propose in a follow up PR)

#### NOTES
* I have already implemented `[bit.rot]` and I intend to remove `Impl::rotate_right(unsigned, int) -> unsigned` in the `Bitset` implementation but I was planing to do that in a follow-up PR
* I intend to incrementally replace everything in `<impl/Kokkos_BitOps.hpp>` until we can get rid of the header
* I will also attempt to replace the `integral_power_of_two*` facility in `impl/Kokkos_Traits.hpp` but I would rather deal with it later
* I only used GCC built-in functions on the host-side for `Experimental::*_builtin` function templates.  We can specialize further and add Intel intrinsics and such later.

PR #5317, opened 2022-08-08
Team- and thread-level sort, sort_by_key

brian-kelley:
(Addresses #645)

Add sort functions that can be called from device,
and exploit team and thread level parallelism. The new functions
use bitonic sort, which is good for this because it's in-place but
highly parallel (when sorting N items, N/2 pairs are compared at once).
It's also comparison-based, so there are versions that can accept an
arbitrary comparison functor (``bool operator()(a, b)`` returns true if key a
goes before key b, otherwise false). Any key type that can be copied in device code will work, as long as there is ``operator<`` for it (or you provide a comparator). So things like Kokkos::pair (see #4487) are fine.

sort_by_key is the same, but in addition to the keys it takes a values
view of the same length. The pairs keys(i) and values(i) are all sorted
according to the key. This is useful for sorting CRS matrices for
example.

The interfaces to sort_[team/thread] and sort_by_key[team/thread] are designed to be similar to Thrust, except they take Kokkos::Views and not iterators.

The new function signatures (8 of them, but all implemented in terms of 2 Impl:: functions):
```
template <class TeamMember, class ViewType>
KOKKOS_INLINE_FUNCTION void sort_team(const TeamMember& t, const ViewType& view);

template <class TeamMember, class ViewType, class Comparator>
KOKKOS_INLINE_FUNCTION void sort_team(const TeamMember& t, const ViewType& view, const Comparator& comp);

template <class TeamMember, class KeyViewType, class ValueViewType>
KOKKOS_INLINE_FUNCTION void sort_by_key_team(const TeamMember& t, const KeyViewType& keyView, const ValueViewType& valueView);

template <class TeamMember, class KeyViewType, class ValueViewType, class Comparator>
KOKKOS_INLINE_FUNCTION void sort_by_key_team(const TeamMember& t, const KeyViewType& keyView, const ValueViewType& valueView, const Comparator& comp);

template <class TeamMember, class ViewType>
KOKKOS_INLINE_FUNCTION void sort_thread(const TeamMember& t, const ViewType& view);

template <class TeamMember, class ViewType, class Comparator>
KOKKOS_INLINE_FUNCTION void sort_thread(const TeamMember& t, const ViewType& view, const Comparator& comp);

template <class TeamMember, class KeyViewType, class ValueViewType>
KOKKOS_INLINE_FUNCTION void sort_by_key_thread(const TeamMember& t, const KeyViewType& keyView, const ValueViewType& valueView);

template <class TeamMember, class KeyViewType, class ValueViewType, class Comparator>
KOKKOS_INLINE_FUNCTION void sort_by_key_thread(const TeamMember& t, const KeyViewType& keyView, const ValueViewType& valueView, const Comparator& comp);
```
BTW, there are 2 full sort implementations now since TeamVectorRange() and ThreadVectorRange() are functions, not types that could be templated on. But once the generic ranges and team/thread handles get added, all the functions could be in terms of 1 implementation.

I tested this myself on architectures PASCAL61, VEGA908, INTEL_XEHP, and made sure that all 8 functions are covered by the new tests.


mhoemmen:
@brian-kelley Have you considered using [CUB](https://docs.nvidia.com/cuda/cub/index.html) for the CUDA back-end?  CUB should always come with your CUDA installation.

brian-kelley:
@mhoemmen Yes, the reasons I didn't use CUB is that its two block-level sorts (Radix and Merge) both require ``__shared__`` temporary storage, and both require the input to be partitioned across threads with a fixed number of items per thread (so the lengths have some small-ish upper bound based on hardware limitations). This bitonic sort doesn't need extra space and any team/thread size can sort any length array (and the data can be any View, any layout, shared or global).

There is still a slight performance hit - when we tried to publish a KokkosKernels paper, this implementation sorted 256-element int arrays 9% slower than BlockRadixSort, using 128 threads per team/block. So it's not a huge slowdown.

algorithms/src/Kokkos_Sort.hpp
```diff
@@ -665,6 +665,231 @@ void sort(ViewType view, size_t const begin, size_t const end) {
...
+template <typename Key>
+struct LessThan {
+  KOKKOS_INLINE_FUNCTION bool operator()(const Key lhs, const Key rhs) const {
+    return lhs < rhs;
+  }
+};
+
+// When just doing sort (not sort_by_key), use nullptr_t for ValueViewType.
+// TODO: when nested policies get refactored, this won't need separate team and
+// thread versions.
```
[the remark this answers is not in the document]
brian-kelley: I should have thought of that before🤦‍♂️
However, there is still the issue of detecting whether a ``t.team_barrier()`` is needed after each parallel_for (true for team-level, false for thread-level). Not sure I can use ``is_same`` on the range's type, because that seems to be backend and index type specific, like for serial it was ``Kokkos::Impl::TeamThreadRangeBoundariesStruct<unsigned long, Kokkos::Impl::HostThreadTeamMember<Kokkos::Serial> >``. Is there another way to check team vs. thread?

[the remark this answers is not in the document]
brian-kelley: Nvm, i think i figured out how to do it

brian-kelley:
@masterleinad Thanks for the review - I just pushed all the suggestions.

algorithms/unit_tests/TestSort.hpp
```diff
@@ -402,6 +403,330 @@ void test_sort_integer_overflow() {
       << "view (" << vh[0] << ", " << vh[1] << ") is not sorted";
 }
 
+// Comparator for sorting in descending order
+template <typename Key>
+struct GreaterThan {
+  KOKKOS_INLINE_FUNCTION bool operator()(const Key lhs, const Key rhs) const {
+    return lhs > rhs;
+  }
```
dalg24, line 411:
```suggestion
  KOKKOS_FUNCTION static constexpr bool operator()(const &Key lhs, const &Key rhs) {
    return lhs > rhs;
  }
```

algorithms/src/Kokkos_Sort.hpp
```diff
@@ -665,6 +667,183 @@ void sort(ViewType view, size_t const begin, size_t const end) {
...
+                                             const ValueViewType& valueView,
+                                             const Comparator& comp,
+                                             const NestedRange<useTeamLevel>&) {
+  using SizeType  = typename KeyViewType::size_type;
+  using KeyType   = typename KeyViewType::non_const_value_type;
+  using Range     = NestedRange<useTeamLevel>;
+  SizeType n      = keyView.extent(0);
+  SizeType npot   = 1;
+  SizeType levels = 0;
+  // FIXME: ceiling power-of-two is a common thing to need - make it a utility
```
dalg24, line 719: Adding a reference to #4577 so we know to update when it eventually lands into `develop` (I am not actively working on it)

algorithms/unit_tests/TestSort.hpp
brian-kelley: @dalg24 I agree about these changes, except for ``static`` - the sorts take an instance of the comparison functor (for maximum flexibility - this lets the functor carry extra data like Views or whatever if needed). Also, the ``StdAlgoLessThanBinaryPredicate`` that is built in (and now used as the default comparison) has non-static operator().

brian-kelley: BTW, Thrust sort/sort_by_key that takes a ``StrictWeakOrdering`` is also an instance, and expects non-static

crtrott (changes requested):
I think this is pretty good. But I'd like us to put it into Experimental, and then hopefully by Kokkos 4.1 we have the more capable execution resource handles, so instead of having:
```c++
sort_team(TeamHandle, ...)
sort_thread(TeamHandle, ...)
```

we simply have:
```c++
sort(TeamHandle, ...)
sort(ThreadHandle, ...)
sort(InlineHandle, ...)
```

dalg24:
> I think this is pretty good. But I'd like us to put it into Experimental, and then hopefully by Kokkos 4.1 we have the more capable execution resource handles, so instead of having:
> 
> ```c++
> sort_team(TeamHandle, ...)
> sort_thread(TeamHandle, ...)
> ```
> 
> we simply have:
> 
> ```c++
> sort(TeamHandle, ...)
> sort(ThreadHandle, ...)
> sort(InlineHandle, ...)
> ```

Agree about `Experimental::`.  Maybe put it in a separate header.

algorithms/src/Kokkos_NestedSort.hpp
```diff
@@ -0,0 +1,230 @@
...
+            KeyType key1 = keyView(elem1);
+            KeyType key2 = keyView(elem2);
+            if (comp(key2, key1)) {
+              keyView(elem1) = key2;
+              keyView(elem2) = key1;
+              if constexpr (!std::is_same_v<ValueViewType, std::nullptr_t>) {
+                Kokkos::Experimental::swap(valueView(elem1), valueView(elem2));
+              }
+            }
+          }
```
masterleinad, line 147:
```suggestion
        SizeType elem2;
        if (j == 0) {
          // first phase (brown box): within a block, compare with the
          // opposite value in the box
          elem2 = boxStart + boxSize - 1 - boxOffset;
        } else {
          // later phases (pink box): within a block, compare with fixed
          // distance (boxSize / 2) apart
          SizeType elem2 = elem1 + boxSize / 2;
        }
        if (elem2 < n) {
          KeyType key1 = keyView(elem1);
          KeyType key2 = keyView(elem2);
          if (comp(key2, key1)) {
            keyView(elem1) = key2;
            keyView(elem2) = key1;
            if constexpr (!std::is_same_v<ValueViewType, std::nullptr_t>) {
              Kokkos::Experimental::swap(valueView(elem1), valueView(elem2));
            }
          }
        }        
```

masterleinad (commented):
I think we can still avoid some duplicated code. Otherwise, this looks good to me.

algorithms/src/Kokkos_NestedSort.hpp
brian-kelley: Nice, pushed.

brian-kelley:
All the suggestions so far are pushed now

masterleinad (approved):

masterleinad:
```
27: [ RUN      ] openmptarget.NestedSort
27: OpenMPTarget backend requires a minimum of 32 threads per team.
```

brian-kelley:
Looks like the test machine fetnat03 has a full disk.

brian-kelley:
This last round of testing exposed a little issue with ``TeamPolicy<OpenMPTarget>`` - ``vector_length_max()`` returned 32, but with that high of a value Kokkos::AUTO can't also satisfy the requirement that team size is at least 32.

algorithms/src/Kokkos_NestedSort.hpp
```diff
@@ -0,0 +1,217 @@
...
+
+// When just doing sort (not sort_by_key), use nullptr_t for ValueViewType.
+// This only takes the NestedRange instance for template arg deduction.
+template <class TeamMember, class KeyViewType, class ValueViewType,
+          class Comparator, bool useTeamLevel>
+KOKKOS_INLINE_FUNCTION void sort_nested_impl(const TeamMember& t,
+                                             const KeyViewType& keyView,
+                                             const ValueViewType& valueView,
+                                             const Comparator& comp,
+                                             const NestedRange<useTeamLevel>&) {
```
dalg24, line 94: Pass tags by value next time (only change if any other change is requested)

algorithms/unit_tests/TestSort.hpp
```diff
@@ -402,6 +403,339 @@ void test_sort_integer_overflow() {
...
+// (inclusive) and ends at offsets(i+1) (exclusive)
+template <typename KeyType, typename KeyViewType, typename OffsetViewType>
+void randomPackedArrays(unsigned n, unsigned k, KeyViewType& keys,
+                        OffsetViewType& offsets, KeyType minKey,
+                        KeyType maxKey) {
+  offsets          = OffsetViewType("Offsets", n + 1);
+  auto offsetsHost = Kokkos::create_mirror_view(Kokkos::HostSpace(), offsets);
+  Kokkos::Random_XorShift64_Pool<Kokkos::DefaultHostExecutionSpace> gHost(1931);
+  Kokkos::fill_random(offsetsHost, gHost, k);
+  Kokkos::DefaultHostExecutionSpace().fence();
```
dalg24, line 555: If you are filling on the host prefer standard library facility

algorithms/unit_tests/TestSort.hpp
```diff
@@ -402,6 +403,339 @@ void test_sort_integer_overflow() {
...
+  offsetsHost(0) = 0;
+  // Prefix-sum
+  for (unsigned i = 1; i <= n; i++) {
+    offsetsHost(i) += offsetsHost(i - 1);
+  }
```
dalg24, line 560: Now that we require C++17 prefer `std::exclusive_scan`

algorithms/unit_tests/TestSort.hpp
```diff
@@ -402,6 +403,339 @@ void test_sort_integer_overflow() {
...
+  Kokkos::deep_copy(offsets, offsetsHost);
+  unsigned totalLength = offsetsHost(n);
+  keys                 = KeyViewType("Keys", totalLength);
+  Kokkos::Random_XorShift64_Pool<typename KeyViewType::execution_space> g(1931);
+  Kokkos::fill_random(keys, g, minKey, maxKey);
+}
+
+// Same as above, but also generate values array with same length as keys
```
dalg24, line 568: Avoid duplicating code.
For instance write 3 functions, one to generate offsets, one for keys, and one for values.

algorithms/unit_tests/TestSort.hpp
```diff
@@ -402,6 +403,339 @@ void test_sort_integer_overflow() {
...
+    std::unordered_multiset<ValueType> outputVals;
+    for (unsigned i = keyStart; i < keyEnd; i++) {
+      correctVals.insert(valuesHost(i));
+      outputVals.insert(valuesOut(i));
+    }
+    // Check one value at a time that they match
+    for (auto it = correctVals.begin(); it != correctVals.end(); it++) {
+      ValueType val = *it;
+      EXPECT_TRUE(outputVals.find(val) != outputVals.end());
+      EXPECT_EQ(correctVals.count(val), outputVals.count(val));
```
dalg24, line 733: Try to produce more meaningful error messages when these fail.
You may use the `operator<<` on the assertion or refactor to have individual unit tests.

algorithms/unit_tests/TestSortCommon.hpp
```diff
@@ -51,5 +51,25 @@ namespace Test {
...
+  // In sort_by_key_X functions, a key view and a value view are both permuted
+  // to make the keys sorted. This means that the value type doesn't need to be
+  // ordered, unlike key
+  Impl::test_nested_sort_by_key<TEST_EXECSPACE, unsigned, unsigned>(
+      161, 0U, UINT_MAX, 0U, UINT_MAX);
+  Impl::test_nested_sort_by_key<TEST_EXECSPACE, float, char>(
+      267, -1e6f, 1e6f, CHAR_MIN, CHAR_MAX);
+  Impl::test_nested_sort_by_key<TEST_EXECSPACE, char, double>(
+      11, CHAR_MIN, CHAR_MAX, 2.718, 3.14);
+}
```
dalg24, line 72: Please create a new unit test source file for these.

algorithms/unit_tests/TestSort.hpp
brian-kelley: I would like to use this, but it seems like the standard tends to be missing some things in the 17 standard. For example, this:
```
#include <numeric>
#include <iostream>
#include <vector>

int main()
{
  std::vector<int> myVec = {1, 2, 3, 4};
  std::exclusive_scan(myVec.begin(), myVec.end(), myVec.begin(), 0);
  for(auto i : myVec)
    std::cout << i << ' ';
  std::cout << '\n';
  return 0;
}
```
doesn't compile for me because all the overloads expect an ``ExecutionPolicy&&`` as the first argument. If I try to ``#include <execution>`` to provide one, I get ``tbb/blocked_range.h: no such file or directory``.
I have GCC 9.1.0, so it's not a super old version (and the new Kokkos minimum is 8.2). Even on kokkos-dev2 with GCC 10.1.0, you can't include ``<execution>`` because the TBB is apparently too old...

algorithms/unit_tests/TestSortCommon.hpp
brian-kelley: But it's OK to keep the actual test logic in TestSort.hpp?

algorithms/unit_tests/TestSort.hpp
dalg24: Looking at https://en.cppreference.com/w/cpp/compiler_support/17#C.2B.2B17_library_features it looks like you are right

algorithms/unit_tests/TestSortCommon.hpp
dalg24: No

dalg24: You can directly put everything in a single new source file

brian-kelley:
@dalg24 I just pushed those suggestions (except I didn't change anything about the prefix sum)

brian-kelley:
@crtrott Could you give this another review? I moved the functions to Kokkos::Experimental.

algorithms/unit_tests/CMakeLists.txt
```diff
@@ -35,6 +35,7 @@ foreach(Tag Threads;Serial;OpenMP;Cuda;HPX;HIP;SYCL;OpenMPTarget)
       "#include <Test${Tag}_Category.hpp>\n"
       "#include <TestRandomCommon.hpp>\n"
       "#include <TestSortCommon.hpp>\n"
+      "#include <TestNestedSort.hpp>\n"
```
dalg24, line 38: Why is it not a separate cpp source file?

algorithms/unit_tests/TestNestedSort.hpp
```diff
@@ -0,0 +1,434 @@
...
+// SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+//
+// Questions? Contact Christian R. Trott (crtrott@sandia.gov)
+//
+// ************************************************************************
+//@HEADER
+*/
+
+#ifndef KOKKOS_ALGORITHMS_UNITTESTS_TEST_NESTED_SORT_HPP
+#define KOKKOS_ALGORITHMS_UNITTESTS_TEST_NESTED_SORT_HPP
```
dalg24, line 46: Why is this a header?

dalg24: Ok because of the category file.  Sorry about the noise

algorithms/unit_tests/CMakeLists.txt
dalg24: I would prefer generating multiple source file but this do not need to be handled here if you open an issue for it

crtrott, line 38: Can you open an issue so we can split this and not end up with a single huge object file.

crtrott (approved):
Just open that additional issue, we also should try what happens if you would use this at the top level (i.e. with an execution space instance ...)

Issue #5313, opened 2022-08-05
Nightly build error: gcc/7.2,8.2 builds with SKX arch, Kokkos_SIMD_AVX512.hpp:680:22: error: ‘_mm512_cmplt_pd_mask’ was not declared in this scope


ndellingwood:
**Describe the bug**
Following the change to require c++17, nightly builds with gcc/7.2, 8.2 for SKX architecture began failing to compile with output:
```
In file included from /home/jenkins/blake-new/workspace/Kokkos_develop_blake_gcc_openmp_pthread_serial_820/kokkos/simd/src/Kokkos_SIMD.hpp:53,
                 from /home/jenkins/blake-new/workspace/Kokkos_develop_blake_gcc_openmp_pthread_serial_820/kokkos/simd/unit_tests/TestSIMD.cpp:47:
/home/jenkins/blake-new/workspace/Kokkos_develop_blake_gcc_openmp_pthread_serial_820/kokkos/simd/src/Kokkos_SIMD_AVX512.hpp: In member function ‘Kokkos::Experimental::simd<double, Kokkos::Experimental::simd_abi::avx512_fixed_size<8> >::mask_type Kokkos::Experimental::simd<double, Kokkos::Experimental::simd_abi::avx512_fixed_size<8> >::operator<(const Kokkos::Experimental::simd<double, Kokkos::Experimental::simd_abi::avx512_fixed_size<8> >&) const’:
/home/jenkins/blake-new/workspace/Kokkos_develop_blake_gcc_openmp_pthread_serial_820/kokkos/simd/src/Kokkos_SIMD_AVX512.hpp:680:22: error: ‘_mm512_cmplt_pd_mask’ was not declared in this scope
     return mask_type(_mm512_cmplt_pd_mask(m_value, other.m_value));
                      ^~~~~~~~~~~~~~~~~~~~
```

**Please also include the following items to support reproducing the bug**
**Reproducer** (Blake testbed):
```
module load cmake/3.19.3 gcc/8.2.0

$KOKKOS_PATH/generate_makefile.bash --with-devices=Serial --arch=SKX --compiler=g++ --cxxflags="-g -Wall -Wunused-parameter -Wshadow -pedantic -Werror -Wsign-compare -Wtype-limits -Wignored-qualifiers -Wempty-body -Wclobbered -Wuninitialized " --cxxstandard="17" --kokkos-path=$KOKKOS_PATH --no-examples   --with-options=enable_large_mem_tests --debug
```

[...]

PR #5318, opened 2022-08-09
Fix SIMD for GCC less than 8

ibaned:
GCC 7.2.0 and older don't provide the special
case _mm512_cmplt_pd_mask and friends, but all
the way down to GCC 5.5.0 do support the
three-argument _mm512_cmp_pd_mask intrinsic.
Instead of cluttering with #ifdef __GNUC__ < 8
everywhere, just use the three-argument instrinsic
always.
The Intel documentation suggests performance will
not suffer...

### DIFF
--- a/simd/src/Kokkos_SIMD_AVX512.hpp
+++ b/simd/src/Kokkos_SIMD_AVX512.hpp
@@ -677,27 +677,27 @@ class simd<double, simd_abi::avx512_fixed_size<8>> {
   }
   KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION mask_type
   operator<(simd const& other) const {
-    return mask_type(_mm512_cmplt_pd_mask(m_value, other.m_value));
+    return mask_type(_mm512_cmp_pd_mask(m_value, other.m_value, _CMP_LT_OS));
   }
   KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION mask_type
   operator>(simd const& other) const {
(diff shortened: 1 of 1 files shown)


ibaned:
Fixes #5313

PhilMiller (changes requested):
https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#expand=727,1063,4909,1062,1062,1063&text=_mm512_mask_cmp_pd_mask&ig_expand=840
The documentation seems to show constants offering any of these comparisons. Could we just use them, rather than making obscure inversions?

ibaned:
> https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#expand=727,1063,4909,1062,1062,1063&text=_mm512_mask_cmp_pd_mask&ig_expand=840 The documentation seems to show constants offering any of these comparisons. Could we just use them, rather than making obscure inversions?

Does 566848f41bb4360e4041b37fe5037252d38ae011 address this?

PhilMiller (approved):

dalg24:
I would have expected we'd target the 3.7 release candidate branch and potentially do nothing in `develop` until we agree on the minimum GCC version we support going forward

ibaned:
Pointed to 3.7 release candidate instead, please let me know if there is anything more I can do. To @dalg24 's point, the minimum GCC version that I know this works for is GCC 5.5.0. Of course I would love minimum GCC 8.1.0  :)

Issue #5257, opened 2022-07-23
Add CTAD (deduction guides) to various types


nliber:

nliber:
For `RangePolicy`, it looks like I can deduce the `ExecutionSpace` and the `IndexType`.

crtrott:
I think we should not: similar to extents(5,5) not deducing int in the C++ standard. I think its too easy to make a mistake.

nliber:
Implemented so far:

RangePolicy only deduces execution space
TeamPolicy only deduces execution space
MDRangePolicy deduces execution space and Rank

nliber:
Adding deduction guides to our standard reducers (`Sum`, `Prod`, etc.) 

It is straightforward to deduce all the template parameters from the `View` constructor.

From the `value_type&` constructor, we cannot deduce the memory space.  We could assume it to be `DefaultExecutionSpace::memory_space`.  Is that useful or error-prone?

nliber:
Open question here on deduction guides for `View`:

Besides the following constructors:

```c++
View(T*, size_t, size_t, size_t, size_t, size_t, size_t, size_t, size_t) -> View<T********>
View(T*, size_t, size_t, size_t, size_t, size_t, size_t, size_t)  -> View<T*******>
View(T*, size_t, size_t, size_t, size_t, size_t, size_t) -> View<T******>
View(T*, size_t, size_t, size_t, size_t, size_t) -> View<T*****>
View(T*, size_t, size_t, size_t, size_t) -> View<T****>
View(T*, size_t, size_t, size_t) -> View<T***>
View(T*, size_t, size_t) -> View<T**>
View(T*, size_t) -> View<T*>
View(T*) -> View<T>

```
are there any other constructors we can deduce the `View` from?

dalg24:
When an execution space is passed via view_alloc
```C++
View<DataType>(view_alloc("v, space), ...) -> View<DataType, decltype(space)>
View<DataType, Layout>(view_alloc("v, space), ...) -> View<DataType, Layout, decltype(space)>

nliber:
@dalg24 `view_alloc("v", space)` doesn't contain the `DataType` needed by `View`, does it?

CTAD has to be `View(...) -> View<...>`; the left hand side cannot have a template argument list.

dalg24:
No it does not.  You're right of course.

ajpowelsnl:
Please summarize how to document deduction guides.

mhoemmen:
[...]

PR #5319, opened 2022-08-09
CTAD (deduction guides) for various types

nliber:
With @ldh4

Added deduction guides for:

```
RangePolicy
TeamPolicy
MDRangePolicy
View
standard reducers
```

This addresses issue #5257.

### DIFF
--- a/core/src/Kokkos_ExecPolicy.hpp
+++ b/core/src/Kokkos_ExecPolicy.hpp
@@ -290,8 +290,7 @@ RangePolicy(int64_t, int64_t, Args...)->RangePolicy<>;
 template <typename ES, typename... Args,
           typename = std::enable_if_t<
               std::is_convertible_v<ES, DefaultExecutionSpace>>>
-RangePolicy(ES const&, int64_t, int64_t, Args...)
-    ->RangePolicy<>;
+RangePolicy(ES const&, int64_t, int64_t, Args...)->RangePolicy<>;
 
 template <typename ES, typename... Args,
           typename = std::enable_if_t<
--- a/core/src/Kokkos_ExecPolicy.hpp
+++ b/core/src/Kokkos_ExecPolicy.hpp
@@ -283,15 +283,15 @@ class RangePolicy : public Impl::PolicyTraits<Properties...> {
 RangePolicy()->RangePolicy<>;
 
 template <typename... Args>
-RangePolicy(int64_t, int64_t, Args...) -> RangePolicy<>;
+RangePolicy(int64_t, int64_t, Args...)->RangePolicy<>;
 
 template <typename... Args>
 RangePolicy(DefaultExecutionSpace const&, int64_t, int64_t, Args...)
-    -> RangePolicy<>;
+    ->RangePolicy<>;
 
 template <typename ES, typename... Args,
           typename = std::enable_if_t<is_execution_space_v<ES>>>
-RangePolicy(ES const&, int64_t, int64_t, Args...) -> RangePolicy<ES>;
+RangePolicy(ES const&, int64_t, int64_t, Args...)->RangePolicy<ES>;
 
 }  // namespace Kokkos
 
@@ -693,28 +693,28 @@ TeamPolicy(DefaultExecutionSpace const&, int, int, Kokkos::AUTO_t const&)
 
 template <typename ES,
           typename = std::enable_if_t<Kokkos::is_execution_space_v<ES>>>
-TeamPolicy(ES const&, int, int) -> TeamPolicy<ES>;
+TeamPolicy(ES const&, int, int)->TeamPolicy<ES>;
 
 template <typename ES,
           typename = std::enable_if_t<Kokkos::is_execution_space_v<ES>>>
-TeamPolicy(ES const&, int, int, int) -> TeamPolicy<ES>;
+TeamPolicy(ES const&, int, int, int)->TeamPolicy<ES>;
 
 template <typename ES,
           typename = std::enable_if_t<Kokkos::is_execution_space_v<ES>>>
-TeamPolicy(ES const&, int, Kokkos::AUTO_t const&) -> TeamPolicy<ES>;
+TeamPolicy(ES const&, int, Kokkos::AUTO_t const&)->TeamPolicy<ES>;
 
 template <typename ES,
           typename = std::enable_if_t<Kokkos::is_execution_space_v<ES>>>
-TeamPolicy(ES const&, int, Kokkos::AUTO_t const&, int) -> TeamPolicy<ES>;
+TeamPolicy(ES const&, int, Kokkos::AUTO_t const&, int)->TeamPolicy<ES>;
 
 template <typename ES,
           typename = std::enable_if_t<Kokkos::is_execution_space_v<ES>>>
 TeamPolicy(ES const&, int, Kokkos::AUTO_t const&, Kokkos::AUTO_t const&)
-    -> TeamPolicy<ES>;
+    ->TeamPolicy<ES>;
 
 template <typename ES,
           typename = std::enable_if_t<Kokkos::is_execution_space_v<ES>>>
-TeamPolicy(ES const&, int, int, Kokkos::AUTO_t const&) -> TeamPolicy<ES>;
+TeamPolicy(ES const&, int, int, Kokkos::AUTO_t const&)->TeamPolicy<ES>;
 
 namespace Impl {
 
--- a/core/src/Kokkos_ExecPolicy.hpp
+++ b/core/src/Kokkos_ExecPolicy.hpp
@@ -285,12 +285,14 @@ RangePolicy()->RangePolicy<>;
 template <typename B, typename E, typename... Args,
           typename = std::enable_if_t<
               !is_execution_space<B>::value &&
+              !std::is_convertible_v<B, DefaultExecutionSpace> &&
               std::is_same_v<std::common_type_t<B, E>, int64_t>>>
 RangePolicy(B, E, Args...) -> RangePolicy<>;
 
 template <typename B, typename E, typename... Args,
           typename = std::enable_if_t<
               !is_execution_space<B>::value &&
+              !std::is_convertible_v<B, DefaultExecutionSpace> &&
               !std::is_same_v<std::common_type_t<B, E>, int64_t>>>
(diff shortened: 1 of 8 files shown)


nmm0 (changes requested):
Would it be possible to split this into separate PRs per type being deduced?

Most of the CI failures may be a result of an incorrect deduction causing a SIGFPE later down the line

core/unit_test/TestViewCtorCTADs.hpp
```diff
@@ -0,0 +1,167 @@
...
+
+struct TestViewCtorCTADs {
+  using DataType                     = double;
+  static constexpr DataType* m       = nullptr;
+  static constexpr const DataType* c = nullptr;
+
+  static void test_viewctorcopymove() {
+    Kokkos::View<DataType> vt;
+    Kokkos::View cd(vt);
+    ASSERT_TRUE((std::is_same_v<decltype(cd), decltype(vt)>));
```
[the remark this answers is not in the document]
nmm0: We may want to consider having this a "compile-only" test (I don't know if we have a way of doing that) since it's a waste of compute time to run these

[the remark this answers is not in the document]
nliber: I think we can, but if there is a bug, it stops all the tests dependent on that file from compiling.  That seems strictly worse to me.

And it still takes compile time either way; it just takes a small bit of run time.  Constructing these objects is fairly cheap, a the `ASSERT_TRUE` statements should all optimize out in correct code (as the answer is known at compile time).

core/src/Kokkos_View.hpp
```diff
@@ -1725,6 +1725,35 @@ class View : public ViewTraits<DataType, Properties...> {
   }
 };
 
+template <typename T>
```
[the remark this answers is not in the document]
nliber: Will do.

I don't think you need the Unmanaged trait, as those constructors should force it to be unmanaged (albeit at run time).

Generally, I'm trying to be minimalistic in adding template parameters.

core/src/Kokkos_Parallel_Reduce.hpp
```diff
@@ -109,6 +109,9 @@ struct Sum {
   bool references_scalar() const { return references_scalar_v; }
 };
 
+template <typename Scalar, typename Space>
+Sum(View<Scalar, Space> const&) -> Sum<Scalar, Space>;
```
crtrott, line 113: This can't possible work, e.g. I hand it `Kokkos::View<double**, Kokkos::LayoutRight>` ...

crtrott (changes requested):

core/src/Kokkos_Parallel_Reduce.hpp
nliber: This seems like an orthogonal issue.  The `Space` argument in the reducer `Sum<Scalar, Space>` is already unconstrained.  I don't think it should be solved as part of adding CTADs.

We should probably open a separate issue for this.  My first thought is that the fix for all the reducers is to replace `Space` with `Properties...` to match the template parameters for `View`.

crtrott: It is not an orthogonal issue. 

I can call the constructor of

```
Sum<int, HostSpace> 
```

With all of the following:
```c++
View<int, HostSpace>
View<int, LayoutRight, HostSpace>
View<int, LayoutLeft, HostSpace>
View<int, LayoutStride, HostSpace, MemoryTraits<Unmanaged>>
...
```

But the deduction would not deduce as written. 
I want to store a plain View, no memory traits no layout (since its scalar, which we may need to fix in the future), no other extraneous information. 
And I want to still have the thing deduced correct. 

So I think we need this:
```c++
template <typename Scalar, class ... Properties>
Sum(View<Scalar, Properties ... > const&) -> Sum<Scalar, typename View<Scalar, Properties ... >::memory_space>;
```

crtrott: In the end we can write deduction guides where the supposed argument of the constructor is anything convertible to the actual constructor argument.

nliber: Oh, I see.  Will try it out.

core/unit_test/TestPolicyConstruction.hpp
```diff
@@ -562,6 +712,164 @@ class TestTeamPolicyConstruction {
...
+    Kokkos::DefaultExecutionSpace des{};
+    ImplicitlyConvertibleToDefaultExecutionSpace notEs{};
+    SomeExecutionSpace ses{};
+    ExecutionSpace es{};
+
+    using PolicyExecSpace = std::conditional_t<
+        std::is_same_v<ExecutionSpace, Kokkos::DefaultExecutionSpace>,
+        Kokkos::TeamPolicy<>, Kokkos::TeamPolicy<ExecutionSpace>>;
+
+    int i{};
```
masterleinad, line 725: Setting the team size to 0 causes some of the test failures.

core/src/Kokkos_Parallel_Reduce.hpp
nliber: I don't think we can quite do anything implicitly convertible to `View<Scalar, Space>` because I have to get the `memory_space` out of the type of the argument passed in, which we cannot do for an arbitrary type.

We can, however, do any `View<Scalar, Properties...>`, which should be sufficient (assuming those are implicitly convertible to `View<Scalar, View<Scalar, Properties...>::memory_space>`, although I could check that too).

In the deduction guides, should I constrain `Scalar` to be a scalar (and by "scalar" I mean non-pointer non-array, or is it something else)?

PhilMiller: I think the sensible phrasing of that latter check is that the View has rank 0?

nliber: And is not `const`.

The question is:  where does the scalar check belong?  IMO, they belong in the reducers, not the deduction guides.

Also, putting SFINAE checks in a deduction guide can make error messages worse, not better, depending on the compiler.

core/unit_test/TestPolicyConstruction.hpp
```diff
@@ -846,6 +1154,103 @@ void more_md_range_policy_construction_test() {
...
+TEST(TEST_CATEGORY, md_range_policy_compile_time_deduction_guides) {
+  Kokkos::DefaultExecutionSpace des{};
+  ImplicitlyConvertibleToDefaultExecutionSpace notEs{};
+  SomeExecutionSpace ses{};
+  TEST_EXECSPACE es{};
+
+  constexpr int t[5]       = {};
+  constexpr int64_t tt[5]  = {};
+  Kokkos::Array<int, 3> a  = {};
+  Kokkos::Array<int, 2> aa = {};
```
ldh4, line 1166:
```suggestion
  Kokkos::Array<int64_t, 3> a  = {};
  Kokkos::Array<int64_t, 2> aa = {};
```

core/unit_test/TestPolicyConstruction.hpp
nliber: Changed the team size to 1.

nliber:
Restest this please

PhilMiller:
Per the in-depth review Nic and I did last month, please split this into separate PRs, at least View, Reducers, Policies, if not subdividing the latter further. It's possible some subset of those will be ready to merge sooner than the others, and would add value if they are.

core/unit_test/TestPolicyConstruction.hpp
nliber: Reverted this change back as it wasn't the issue

core/src/Kokkos_Parallel_Reduce.hpp
```diff
@@ -99,6 +99,10 @@ struct Sum {
   bool references_scalar() const { return references_scalar_v; }
 };
 
+template <typename Scalar, typename... Properties>
+Sum(View<Scalar, Properties...> const&)
+    ->Sum<Scalar, typename View<Scalar, Properties...>::memory_space>;
+
```
crtrott, line 105: Should we have this:
```c++
template<typename Scalar>
Sum(Scalar&) -> Sum<Scalar, AnonymousSpace>;
```
and would we need some sfinae/requires stuff on it to make it not trigger with Views?

core/src/Kokkos_View.hpp
```diff
@@ -1708,6 +1708,36 @@ class View : public ViewTraits<DataType, Properties...> {
...
+View(T*, size_t, size_t, size_t, size_t, size_t, size_t)->View<T******>;
+
+template <typename T>
+View(T*, size_t, size_t, size_t, size_t, size_t, size_t, size_t)
+    ->View<T*******>;
+
+template <typename T>
+View(T*, size_t, size_t, size_t, size_t, size_t, size_t, size_t, size_t)
+    ->View<T********>;
```
crtrott, line 1739: General questions after thinking about these: are they too dangerous? Should we make it use AnonymousSpace instead? I.e. its type erased memory space as opposed to just assuming you gave me Default memory space?

core/src/Kokkos_Parallel_Reduce.hpp
crtrott: Thought that AnonymousSpace would be problematic, since it reads as "device accessible". We would need to switch our logic in the reduction where we check that, to make AnonymousSpace be treated as not not-device accessible maybe?

crtrott (changes requested):
Couple comments with thoughts.

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -280,6 +280,23 @@ class RangePolicy : public Impl::PolicyTraits<Properties...> {
   };
 };
 
+RangePolicy()->RangePolicy<>;
+
+template <typename... Args>
+RangePolicy(int64_t, int64_t, Args...)->RangePolicy<>;
```
crtrott, line 286: What about deducing IndexType, i.e. can we do 
```
template<class Iarg1, class Iarg2, typename ... Args, typename = std::enable_if_t<is_integral_v(Iarg1) && is_integral_v(Iarg2)>>
RangePolicy(Iarg1, Iarg2, Args...) -> RangePolicy<IndexType<common_type_t<Iarg1,Iarg2>>>;
```

nliber:
Refactored into:
#5515 (reducers)
#5516 (MDRangePolicy)
#5517 (RangePolicy)
#5518 (TeamPolicy)
#5519 (View)

PR #5322, opened 2022-08-09
OpenACC `parallel_for` and `parallel_reduce`

JBludau:
This PR advances the integration of the OpenACC backend into kokkos.

It introduces `parallel_for` and `parallel_reduce` for `RangePolicy` and bumps the incremental tests to level 5.
Pedro did the implementation work, I restructured, rebased and deleted includes some unnecessary code.

A lot of the implementation was based on `OpenMPTarget` so please think about if the design chosen there is something we want to stick with.


JBludau:
@pedrovalerolara @seyonglee the branch is now rebased on the current develop now as requested

core/src/OpenACC/Kokkos_OpenACC_ParallelFor_Range.hpp
```diff
@@ -0,0 +1,85 @@
...
+          "Kokkos::Impl::ParallelFor< OpenACC > can not be executed with "
+          "a range <= 0."));
+    }
+
+    int const async_arg = m_policy.space().acc_async_queue();
+    // avoid implicit capture of *this which would yield a memory access
+    // violation when executing on the device
+    auto const& a_functor(m_functor);
+
+#pragma acc parallel loop gang vector copyin(a_functor) async(async_arg)
```
dalg24, line 78: Why "gang vector"?

core/src/OpenACC/Kokkos_OpenACC_ParallelFor_Range.hpp
```diff
@@ -0,0 +1,85 @@
...
+
+  void execute() const {
+    auto const begin = m_policy.begin();
+    auto const end   = m_policy.end();
+
+    if (end <= begin) {
+      Kokkos::Impl::throw_runtime_exception(std::string(
+          "Kokkos::Impl::ParallelFor< OpenACC > can not be executed with "
+          "a range <= 0."));
+    }
```
JBludau, line 71: I added the `throw` here. Turns out, we should not throw when running a parallel for with size 0. Nevertheless, we should think about throwing for negative ranges, as this is something the user should be informed about in my opinion. Could be done at another point in core, as this should be independent of the backend.

core/src/OpenACC/Kokkos_OpenACC_ParallelReduce_Range.hpp
```diff
@@ -0,0 +1,429 @@
...
+    init(ltmp);
+    if constexpr (std::is_same<TagType, void>::value) {
+#pragma acc parallel loop gang vector reduction(+ : ltmp) copyin(a_functor)
+      for (auto i = begin; i < end; i++) a_functor(i, ltmp);
+    } else {
+#pragma acc parallel loop gang vector reduction(+ : ltmp) copyin(a_functor)
+      for (auto i = begin; i < end; i++) a_functor(TagType(), i, ltmp);
+    }
+    tmp = ltmp;
+  }
```
JBludau, line 109: This could be collapsed into a single function returning the value which would be easier to read. Furthermore, this would reduce the number of times we get `begin`, `end`, etc.

core/src/OpenACC/Kokkos_OpenACC_DeepCopy.hpp
```diff
@@ -150,7 +150,7 @@ template <class ExecutionSpace>
 struct Kokkos::Impl::DeepCopy<
     Kokkos::HostSpace, Kokkos::Experimental::OpenACCSpace, ExecutionSpace> {
   DeepCopy(void* dst, const void* src, size_t n) {
-    if (n > 0) acc_memcpy_from_device_async(dst, const_cast<void*>(src), n);
+    if (n > 0) acc_memcpy_from_device(dst, const_cast<void*>(src), n);
```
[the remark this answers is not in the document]
dalg24: Does not compile w/o it?

core/src/OpenACC/Kokkos_OpenACC_ParallelFor_Range.hpp
seyonglee: OpenACC provides three levels of parallelism (gang/worker/vector), but most of existing GPUs (NVIDIA GPU and AMD GPU) provides only two levels of parallelism, and thus mapping OpenACC three level parallelism to the GPU's two level parallelism has been issues for a long time (and each OpenACC compiler (e.g., NVHPC and GCC) also changed their mapping strategies multiple times). For now, just using both gang and vector seems to be the best choice at least for both NVHPC and GCC.

seyonglee: `acc parallel loop …` may work well for NVHPC, but GCC seems to be less efficient in handling auto parallelization and thus may not work well.
A bigger issue is that an OpenACC compiler may not be able to figure out what kind of parallelisms are included in the functor, in which case, the compiler will not be able to find which parallelisms can be allowed to this loop. 
Therefore, it will be better to keep `loop gang vector` as it is.

Rombur:
CI is failing
```
33: [ RUN      ] OPENACC.IncrTest_05_reduction
33: unknown file: Failure
33: C++ exception with description "Kokkos::Impl::ParallelFor< OpenACC > can not be executed with a range <= 0." thrown in the test body.
```

core/src/OpenACC/Kokkos_OpenACC_ParallelFor_Range.hpp
Rombur: Well looks like the CI hits this `throw`

masterleinad: I'm not quite sure if running with `end>begin` (and expecting to do nothing) is necessarily a bug.

pedrovalerolara: According to today's meeting, it looks like we will let this as it is  (the end <= begin is controlled by the back ends), right?

core/src/OpenACC/Kokkos_OpenACC_ParallelReduce_Range.hpp
pedrovalerolara: Sorry Jakob, I need more details. What do you mean with "this could be collapsed"? All the "reduce" functions in one?

core/src/OpenACC/Kokkos_OpenACC_ParallelFor_Range.hpp
pedrovalerolara: It can look weird using the first (gang) and last (vector) OpenACC parallelism level together, but as Seyong commented, using gang and vector works better

core/src/OpenACC/Kokkos_OpenACC_ParallelReduce_Range.hpp
JBludau: I meant, that the body of the `init` function could be put into the reduce function, thus only leaving a single function.

core/src/OpenACC/Kokkos_OpenACC_ParallelFor_Range.hpp
JBludau: > According to today's meeting, it looks like we will let this as it is (the end <= begin is controlled by the back ends), right?

We will just handle the case of `begin=end` in the backends and check for `end<begin` in `RangePolicy` (separate pr)

core/src/OpenACC/Kokkos_OpenACC_ParallelFor_Range.hpp
crtrott: do a ifdef based on compiler: for nvhpc use `acc parallel loop`, and for gcc `acc parallel loop gang vector`

crtrott: and then lets complain to the vendors. So generally I want to see at least a comment what we think we should ideally use, vs what we need to use because the compiler is dumb

seyonglee: `acc parallel loop gang vector` is the best choice for NVHPC too, and even the NVHPC compiler may have the same limit on handling functors; if the compiler cannot figure out available parallelisms in the functor, `acc parallel loop` will not work.

crtrott (approved):
Approving with the understanding that a whole bunch of reduction features are not yet supported.

core/src/OpenACC/Kokkos_OpenACC_ParallelReduce_Range.hpp
```diff
@@ -0,0 +1,394 @@
...
+
+// primary template: catch-all non-implemented custom reducers
+template <class Functor, class Reducer, class Policy,
+          bool = std::is_arithmetic_v<typename Reducer::value_type>>
+struct OpenACCParallelReduceHelper {
+  OpenACCParallelReduceHelper(Functor const&, Reducer const&, Policy const&) {
+    Kokkos::abort(
+        "[ERROR in ParallelReduce<Functor, Reducer, RangePolicy, OpenACC>] not "
+        "implemented.\n");
+  }
```
dalg24, line 63: Note that we could go for a compile error if that is being instantiated (instead of runtime error).
I just refactored the initial approach and did not think it through.  Let me know if you want me to change now.

JBludau: would prefer, as there is no necessity to wait on runtime

core/src/OpenACC/Kokkos_OpenACC_FunctorAdapter.hpp
```diff
@@ -0,0 +1,82 @@
...
+// ************************************************************************
+//@HEADER
+*/
+
+#ifndef KOKKOS_OPENACC_FUNCTOR_ADAPTER_HPP
+#define KOKKOS_OPENACC_FUNCTOR_ADAPTER_HPP
+
+#include <type_traits>
+
+namespace Kokkos::Experimental::Impl {
```
JBludau, line 50: Could be moved into a file that does not have `OpenACC` as there are no details depending on openACC and this could be used for other backends as well

JBludau (commented):
I can't approve as author but looks good to me.

core/src/OpenACC/Kokkos_OpenACC_FunctorAdapter.hpp
dalg24: My intent was to do that eventually (not in this PR)

core/src/OpenACC/Kokkos_OpenACC_ParallelReduce_Range.hpp
dalg24: Done

core/src/OpenACC/Kokkos_OpenACC_ParallelReduce_Range.hpp
```diff
@@ -0,0 +1,392 @@
...
+};
+
+template <class Functor, class Scalar, class Space, class... Traits>
+struct Kokkos::Experimental::Impl::OpenACCParallelReduceHelper<
+    Functor, Kokkos::Prod<Scalar, Space>, Kokkos::RangePolicy<Traits...>,
+    true> {
+  //                 ^^^^
+  using Policy    = RangePolicy<Traits...>;
+  using Reducer   = Sum<Scalar, Space>;
```
dalg24, line 161: Oooops

core/src/OpenACC/Kokkos_OpenACC_FunctorAdapter.hpp
JBludau: sure

core/src/OpenACC/Kokkos_OpenACC_ParallelReduce_Range.hpp
JBludau: looks still okay though :-)

core/src/OpenACC/Kokkos_OpenACC_FunctorAdapter.hpp
```diff
@@ -0,0 +1,82 @@
...
+  using WorkTag = typename Policy::work_tag;
+
+ public:
+  FunctorAdapter(Functor const &functor) : m_functor(functor) {}
+
+  template <class... Args>
+  KOKKOS_FUNCTION void operator()(Args &&... args) const {
+    m_functor(WorkTag(), static_cast<Args &&>(args)...);
+  }
+};
```
masterleinad, line 78:
```suggestion
template <class Functor, class Policy>
class FunctorAdapter {
  Functor m_functor;
  using WorkTag = typename Policy::work_tag;

 public:
  FunctorAdapter(Functor const &functor) : m_functor(functor) {}

  template <class... Args>
  KOKKOS_FUNCTION void operator()(Args &&... args) const {
    if constexpr (std::is_same_v<WorkTag>)
      m_functor(static_cast<Args &&>(args)...);
    else
      m_functor(WorkTag(), static_cast<Args &&>(args)...);
  }
};
```

core/src/OpenACC/Kokkos_OpenACC_ParallelReduce_Range.hpp
pedrovalerolara: Sure! We can do that. Do you want me to do it? There some lines to change here, one per supported reduction type.

core/src/OpenACC/Kokkos_OpenACC_FunctorAdapter.hpp
dalg24: Did you format?

masterleinad: No, I didn't.

dalg24: That's fine I had to edit anyway because the condition on the if constexpr was wrong.
I applied your suggestion though

seyonglee (approved):
It looks good to me too.

dalg24:
```
4: [ RUN      ] openmptarget.int64_t_reduce_dynamic
4: CUDA error: an illegal memory access was encountered 
4: Libomptarget error: Call to targetDataEnd failed, abort target.
4: Libomptarget error: Failed to process data after launching the kernel.
4: Libomptarget error: Consult https://openmp.llvm.org/design/Runtimes.html for debugging options.
4: Kokkos_OpenMPTarget_Parallel.hpp:333:1: Libomptarget fatal error 1: failure of target construct while offloading is mandatory
 4/33 Test  #4: KokkosCore_UnitTest_OpenMPTarget .............Child aborted***Exception:  25.47 sec
```
and
```
7: Time CudaSpace: 0.006351 CudaUVMSpace_1: 0.006383 CudaUVMSpace_2: 0.015805 CudaPinnedHostSpace: 0.244889 CudaUVMSpace_Pinned: 0.018538
7: /var/jenkins/workspace/Kokkos/core/unit_test/cuda/TestCuda_DebugPinUVMSpace.cpp:128: Failure
7: Value of: passed
7:   Actual: false
7: Expected: true
7: [  FAILED  ] cuda.debug_pin_um_to_host (412 ms)
```
are clearly unrelated

Issue #5312, opened 2022-08-05
Clang/10+Cuda test failure: cuda.reducers_complex_double cudaStreamSynchronize(stream) error( cudaErrorIllegalAddress): an illegal memory access was encountered


ndellingwood:
**Describe the bug**
Cuda builds of kokkos (develop branch and release-candidate-3.7.00) are failing in the `cuda.reducers_complex_double` test with clang/10.0.0 and cuda/{10.1,11.0}

```
[ RUN      ] cuda.reducers_complex_double
cudaStreamSynchronize(stream) error( cudaErrorIllegalAddress): an illegal memory access was encountered /ascldap/users/ndellin/kokkos/core/src/Cuda/Kokkos_Cuda_Instance.cpp:172
Backtrace:
                        [0xd23ce0]
                        [0xd1bff8]
                        [0xd1c078]
                        [0x41fae6]
                        [0xd2bd2f]
                        [0xd2b7ff]
                        [0xbb7abc]
                        [0xbb72d6]
                        [0xbb6d97]
                        [0xbaea14]
                        [0xceee44]
                        [0xcc6ac8]
                        [0xcc7d70]
                        [0xcc8657]
                        [0xcd99ea]
                        [0xcefa34]
                        [0xcd940c]
                        [0x41b640]
__libc_start_main [0x7f3628548555]
                        [0x41b552]


0% tests passed, 1 tests failed out of 1

Total Test time (real) =  38.58 sec
```

**Please also include the following items to support reproducing the bug**
**Reproducer**:
```
source /projects/sems/modulefiles/utils/sems-archive-modules-init.sh

module load sems-archive-env sems-archive-gcc/7.3.0 sems-archive-clang/10.0.0 sems-archive-cuda/10.1 sems-archive-cmake/3.19.1

# Option 1 for configure via generate_makefile script
$KOKKOS_PATH/generate_makefile.bash --with-cuda --compiler=clang++ --arch=Volta70`

# Alternatively configure via cmake
[...]

PR #3580, opened 2020-11-11
Fix reduction bug introduced by zero length reduction fix

crtrott:
Looks like a potential compiler issue .. the fix doesn't really make sense to me but its fully reproducible even though executed instructions shouldn't change at all .... (Note I verified that at no point in the failing test is begin()==end() for the policy ...

PR #5324, opened 2022-08-11
#5312: Revert #3580 and try a different workaround (3.7 RC branch)

PhilMiller:
#3580 was an overwrought attempt at working around a bug in NVCC 9.x. On the hypothesis that the original compiler bug was related to re-use/re-assignment of the predicate variable `do_final_reduction`, I tried reverting the workaround and introducing separate variables for the initial and subsequent condition. It seems to have worked for both Clang 10 and NVCC 9.2 and 10.1

Fixes: #5312

### DIFF
--- a/core/src/Cuda/Kokkos_Cuda_Parallel_Range.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_Parallel_Range.hpp
@@ -256,11 +256,17 @@ class ParallelReduce<FunctorType, Kokkos::RangePolicy<Traits...>, ReducerType,
       }
     }
 
-    // Doing code duplication here to fix issue #3428
-    // Suspect optimizer bug??
     // Reduce with final value at blockDim.y - 1 location.
     // Shortcut for length zero reduction
-    if (m_policy.begin() == m_policy.end()) {
+    bool zero_length        = m_policy.begin() == m_policy.end();
+    bool do_final_reduction = true;
+    if (!zero_length)
+      do_final_reduction = cuda_single_inter_block_reduce_scan<false>(
+          final_reducer, blockIdx.x, gridDim.x,
+          kokkos_impl_cuda_shared_memory<word_size_type>(), m_scratch_space,
+          m_scratch_flags);
+
+    if (do_final_reduction) {
       // This is the final block with the final result at the final threads'
       // location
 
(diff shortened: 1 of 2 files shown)


PhilMiller:
Corresponding change to `develop` is in #5325

PhilMiller:
Retest this please

crtrott (approved):

masterleinad:
The relevant changes wrt. pre-3580 look somewhat like (reverse order)
```diff
diff --git a/core/src/Cuda/Kokkos_Cuda_Parallel_Range.hpp b/core/src/Cuda/Kokkos_Cuda_Parallel_Range.hpp
index 987334300..21ced9551 100644
--- a/core/src/Cuda/Kokkos_Cuda_Parallel_Range.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_Parallel_Range.hpp
@@ -258,12 +258,11 @@ class ParallelReduce<FunctorType, Kokkos::RangePolicy<Traits...>, ReducerType,
 
     // Reduce with final value at blockDim.y - 1 location.
     // Shortcut for length zero reduction
-    bool zero_length        = m_policy.begin() == m_policy.end();
-    bool do_final_reduction = true;
-    if (!zero_length)
+    bool do_final_reduction = m_policy.begin() == m_policy.end();
+    if (!do_final_reduction)
       do_final_reduction = cuda_single_inter_block_reduce_scan<false>(
           final_reducer, blockIdx.x, gridDim.x,
-          kokkos_impl_cuda_shared_memory<word_size_type>(), m_scratch_space,
+          kokkos_impl_cuda_shared_memory<size_type>(), m_scratch_space,
           m_scratch_flags);
 
     if (do_final_reduction) {
diff --git a/core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp b/core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
index cdd16085b..6883aeac4 100644
--- a/core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
@@ -759,15 +759,13 @@ class ParallelReduce<FunctorType, Kokkos::TeamPolicy<Properties...>,
     }
 
     // Reduce with final value at blockDim.y - 1 location.
-    bool zero_length        = m_league_size == 0;
-    bool do_final_reduction = true;
-    if (!zero_length)
-      do_final_reduction = cuda_single_inter_block_reduce_scan<false>(
+    bool do_final_reduce = (m_league_size == 0);
+    if (!do_final_reduce)
+      do_final_reduce = cuda_single_inter_block_reduce_scan<false>(
           final_reducer, blockIdx.x, gridDim.x,
           kokkos_impl_cuda_shared_memory<size_type>(), m_scratch_space,
           m_scratch_flags);
-
-    if (do_final_reduction) {
+    if (do_final_reduce) {
       // This is the final block with the final result at the final threads'
       // location
 ```
I wonder if the switch from `size_type` to `word_size_type` also has some impact still recognizing that we introduced past #3580 for the half type reductions. Still, I wonder if we should change the `Team`  implementation to use `word_size_type`, too.
For comparison:
```
// Conditionally set word_size_type to int16_t or int8_t if value_type is
// smaller than int32_t (Kokkos::Cuda::size_type)
// word_size_type is used to determine the word count, shared memory buffer
// size, and global memory buffer size before the reduction is performed.
// Within the reduction, the word count is recomputed based on word_size_type
// and when calculating indexes into the shared/global memory buffers for
// performing the reduction, word_size_type is used again.
// For scalars > 4 bytes in size, indexing into shared/global memory relies
// on the block and grid dimensions to ensure that we index at the correct
// offset rather than at every 4 byte word; such that, when the join is
// performed, we have the correct data that was copied over in chunks of 4
// bytes.
using word_size_type = std::conditional_t<
      sizeof(value_type) < sizeof(Kokkos::Cuda::size_type),
      std::conditional_t<sizeof(value_type) == 2, int16_t, int8_t>,
      Kokkos::Cuda::size_type>;

using size_type      = Cuda::size_type;
```

PhilMiller:
The OpenMPTarget build timed out. Clearly irrelevant to CUDA backend changes here

PhilMiller:
Comments on the partner PR are maybe worth addressing before merging though

crtrott:
The bug was present before the introduction of word_size_type

ndellingwood (approved):
Thanks @PhilMiller !

PhilMiller:
I'm going to suggest leaving the `Team` case with `size_type` for the moment, and merging this PR as is, so that we can unblock the release. If we can put together a test that shows a separate failure for that case because of the `size_type/word_size_type` discrepancy, then we should fix that as a separate effort.

masterleinad (approved):

PR #5325, opened 2022-08-11
#5312: Revert #3580 and try a different workaround

PhilMiller:
https://github.com/kokkos/kokkos/pull/3580 was an overwrought attempt at working around a bug in NVCC 9.x. On the hypothesis that the original compiler bug was related to re-use/re-assignment of the predicate variable do_final_reduction, I tried reverting the workaround and introducing separate variables for the initial and subsequent condition. It seems to have worked for both Clang 10 and NVCC 9.2 and 10.1

Fixes: https://github.com/kokkos/kokkos/issues/5312


PhilMiller:
Retest this please

crtrott (approved):

masterleinad (approved):

PR #5326, opened 2022-08-11
Drop `reciprocal_overflow_threshold` from the numeric traits

dalg24:
LEWG voted its removal when reviewing P2551R0 on 2022/03/29

I propose simple removal (not going through deprecation).


crtrott (approved):

JBludau (approved):
looks good

PR #3941, opened 2021-04-14
[HIP] Use hipcc architecture autodetection when Kokkos_ARCH is not set

Rombur:
If no architecture is selected `hipcc` does the right thing.

Closes #3933

PR #5327, opened 2022-08-11
Enable automatic detection of arch when enabling 'HIP' with 'hipcc'

JBludau:
With #5289 I noticed that we do not autodetect the arch when using `HIP` but silently end up not specifying any.
If I assess this correctly, this triggers `hipcc` to build for any arch it knows. 

This PR uses the autodetection at configure time to set an ARCH and thus the corresponding Kokkos internal flags/defines etc. It errors out if no or an unknown arch was detected during the automatic detection step.


dalg24:
@crtrott @ndellingwood I want this considered for 3.7

dalg24:
Daniel pointed out to me https://github.com/kokkos/kokkos/pull/3941#issue-858004591 which indicates that, at the time, we thought that was OK.  Considering that we do things like
https://github.com/kokkos/kokkos/blob/839df99bcfc94fce4055aa629048b3a7f5b61eee/core/src/impl/Kokkos_SharedAlloc.hpp#L68-L70
I think that the changes proposed in this PR are the right thing to do.

cmake/kokkos_arch.cmake
```diff
@@ -574,15 +574,37 @@ CHECK_AMDGPU_ARCH(VEGA906 gfx906) # Radeon Instinct MI50 and MI60
...
+        SET_AND_CHECK_AMD_ARCH(VEGA906 ${DETECTED_GPU_ARCH})
+      ELSEIF("${DETECTED_GPU_ARCH}" STREQUAL "gfx908")
+        SET_AND_CHECK_AMD_ARCH(VEGA908 ${DETECTED_GPU_ARCH})
+      ELSEIF("${DETECTED_GPU_ARCH}" STREQUAL "gfx90a")
+        SET_AND_CHECK_AMD_ARCH(VEGA90a ${DETECTED_GPU_ARCH})
+      ELSE()
+        MESSAGE(SEND_ERROR "HIP enabled but no AMD GPU architecture could be automatically detected. "
+                         "Please manually enable one AMD GPU architecture via -DKokkos_ARCH_{..}=ON'.")
+      ENDIF()
+      LIST(APPEND KOKKOS_ENABLED_ARCH_LIST ${AMD_ARCHITECTURE})
```
masterleinad, line 607: Move this to `SET_AND_CHECK_AMD_ARCH` to avoid passing the extra variable.

cmake/kokkos_arch.cmake
```diff
@@ -574,15 +574,37 @@ CHECK_AMDGPU_ARCH(VEGA906 gfx906) # Radeon Instinct MI50 and MI60
...
     IF(${len_str} LESS 8)
-      MESSAGE(SEND_ERROR "HIP enabled but no AMD GPU architecture currently enabled. "
-                         "Please enable one AMD GPU architecture via -DKokkos_ARCH_{..}=ON'.")
+      MESSAGE(SEND_ERROR "HIP enabled but no AMD GPU architecture could be automatically detected. "
+                         "Please manually enable one AMD GPU architecture via -DKokkos_ARCH_{..}=ON'.")
+    # check for known gpu archs, otherwise error out
+    ELSE()
+      STRING(REGEX MATCH "(gfx900|gfx906|gfx908|gfx90a)" DETECTED_GPU_ARCH ${GPU_ARCHS})
+      IF("${DETECTED_GPU_ARCH}" STREQUAL "gfx900")
+        SET_AND_CHECK_AMD_ARCH(VEGA900 ${DETECTED_GPU_ARCH})
```
nmm0, line 596: Won't this cause an error because `CHECK_AMDGPU_ARCH` is set multiple times (first on lines 572-575) I might be misunderstanding something.

cmake/kokkos_arch.cmake
```diff
@@ -574,15 +574,37 @@ CHECK_AMDGPU_ARCH(VEGA906 gfx906) # Radeon Instinct MI50 and MI60
...
     STRING(LENGTH "${GPU_ARCHS}" len_str)
     # enumerator always output gfx000 as the first line
     IF(${len_str} LESS 8)
-      MESSAGE(SEND_ERROR "HIP enabled but no AMD GPU architecture currently enabled. "
-                         "Please enable one AMD GPU architecture via -DKokkos_ARCH_{..}=ON'.")
+      MESSAGE(SEND_ERROR "HIP enabled but no AMD GPU architecture could be automatically detected. "
+                         "Please manually enable one AMD GPU architecture via -DKokkos_ARCH_{..}=ON'.")
+    # check for known gpu archs, otherwise error out
+    ELSE()
+      STRING(REGEX MATCH "(gfx900|gfx906|gfx908|gfx90a)" DETECTED_GPU_ARCH ${GPU_ARCHS})
```
masterleinad, line 594: What if there are multiple (different) GPUs? I guess it's fine to take the first one since we are also taking just whatever for Cuda in that case.

cmake/kokkos_arch.cmake
```diff
@@ -574,15 +574,37 @@ CHECK_AMDGPU_ARCH(VEGA906 gfx906) # Radeon Instinct MI50 and MI60
...
+      IF("${DETECTED_GPU_ARCH}" STREQUAL "gfx900")
+        SET_AND_CHECK_AMD_ARCH(VEGA900 ${DETECTED_GPU_ARCH})
+      ELSEIF("${DETECTED_GPU_ARCH}" STREQUAL "gfx906")
+        SET_AND_CHECK_AMD_ARCH(VEGA906 ${DETECTED_GPU_ARCH})
+      ELSEIF("${DETECTED_GPU_ARCH}" STREQUAL "gfx908")
+        SET_AND_CHECK_AMD_ARCH(VEGA908 ${DETECTED_GPU_ARCH})
+      ELSEIF("${DETECTED_GPU_ARCH}" STREQUAL "gfx90a")
+        SET_AND_CHECK_AMD_ARCH(VEGA90a ${DETECTED_GPU_ARCH})
```
masterleinad, line 602:
```suggestion
        SET_AND_CHECK_AMD_ARCH(VEGA90A ${DETECTED_GPU_ARCH})
```

cmake/kokkos_arch.cmake
JBludau: The autodetection is only entered if no user specified ARCH (lines 572-575) was detected as it is guarded by: `NOT AMDGPU_ARCH_ALREADY_SPECIFIED`

cmake/kokkos_arch.cmake
JBludau: Good question. It basically uses the first valid one returned by `rocm_agent_enumerator`. Unfortunately documentation is sparse and it does not provide a `--help`. Maybe @arghdos can provide some insight on how the output is ordered (e.g. is the gpu with the highest capability listed as first one)

cmake/kokkos_arch.cmake
JBludau: good catch

cmake/kokkos_arch.cmake
JBludau: done in [54bca5](https://github.com/kokkos/kokkos/pull/5327/commits/54bca53977483dc918c4bc731a6042e9b94d0bab)

cmake/kokkos_arch.cmake
JBludau: But if we know which order we want I can change the regex to enforce a specific order.

masterleinad: I'm fine with the current version.

masterleinad (approved):

cmake/kokkos_arch.cmake
arghdos: > how the output is ordered (e.g. is the gpu with the highest capability listed as first one)

I believe in the past I was told it's enumerating things by PCIe ID.  I.e., this should *probably* match the order it shows up in lspci.  The reason I believe this is because it doesn't have any dependence on any of the runtimes.  It's _possible_ this matched the view of ROCR_VISIBLE_DEVICES (the HSA layer), but I'd have to test this.

arghdos: Ah, ok -- reading the script (it's just a python script) and it documents the ordering: https://github.com/RadeonOpenCompute/rocminfo/blob/master/rocm_agent_enumerator#L225

Basically it chooses the first of:

1. ROCM_TARGET_LIST -- used in docker
2. tgt.list -- used in docker
3. HSA [this is actually using the kernel driver ordering under the hood]
4. lspci
5. rocminfo
To enforce an order, we _could_ create a temporary file matching the regex order, and export ROCM_TARGET_LIST to that file, and remove it after the fact.

Rombur:
> If I assess this correctly, this triggers hipcc to build for any arch it knows.

That's not true (unless the behavior was changed).

Rombur:
I don't like this PR. We are duplicating what's done upstream for no good reason. It is true that we use `#ifdef KOKKOS_ARCH_VEGA` but we can just set that variable if `HIP` is enable.

arghdos:
@Rombur -- it might matter if we wanted to add support for the RDNA GPUs (e.g., https://www.amd.com/en/products/professional-graphics/amd-radeon-pro-w6800).  The branch I've seen that [ports to RDNA](https://github.com/iotamudelta/kokkos/commits/navi21) uses a different per-processor family `KOKKOS_ARCH_NAVI21`, as this is different family (sort of in the same way as the Ampere / Volta divide)

ajpowelsnl:
@JBludau -- Will this fix go in Kokkos-3.7?

JBludau (commented):
I really dislike the state it is in at the moment (trying to do all in one function). I will redesign in the next days

Rombur (approved):

masterleinad (approved):
Fine with me.

cmake/kokkos_arch.cmake
```diff
@@ -569,26 +577,18 @@ ENDFUNCTION()
...
-  ELSE()
-    MESSAGE(SEND_ERROR "HIP enabled but no AMD GPU architecture currently enabled. "
-                       "Please enable one AMD GPU architecture via -DKokkos_ARCH_{..}=ON'.")
-  ENDIF()
-ENDIF()
+MACRO(SET_AND_CHECK_AMD_ARCH ARCH FLAG)
+  KOKKOS_SET_OPTION(ARCH_${ARCH} ON)
+  CHECK_AMDGPU_ARCH(${ARCH} ${FLAG})
+  LIST(APPEND KOKKOS_ENABLED_ARCH_LIST ${ARCH})
+  SET(KOKKOS_ENABLED_ARCH_LIST  ${KOKKOS_ENABLED_ARCH_LIST})
```
dalg24, line 590: ??

cmake/kokkos_arch.cmake
```diff
@@ -808,11 +808,52 @@ IF (KOKKOS_ARCH_AMPERE80 OR KOKKOS_ARCH_AMPERE86)
...
+    ENDIF()
+  ENDIF()
 ENDIF()
 
+#Regardless of version, make sure we define the general architecture name
+FOREACH(ARCH IN LISTS SUPPORTED_AMD_ARCHS)
+  IF (KOKKOS_ARCH_${ARCH})
+    STRING(REGEX MATCH "(VEGA)" IS_VEGA ${ARCH})
+    IF(IS_VEGA)
+      SET(KOKKOS_ARCH_VEGA ON)
```
dalg24, line 852:
```suggestion
      SET(KOKKOS_ARCH_VEGA ON)
      BREAK()
```

cmake/kokkos_arch.cmake
```diff
@@ -91,10 +91,18 @@ IF(Kokkos_ENABLE_HIP OR Kokkos_ENABLE_OPENMPTARGET OR Kokkos_ENABLE_UNSUPPORTED_
...
+LIST(APPEND SUPPORTED_AMD_GPUS       MI200    MI100    MI50/60 MI25)
+LIST(APPEND SUPPORTED_AMD_ARCHS      VEGA90A  VEGA908  VEGA906 VEGA900)
+LIST(APPEND CORRESPONDING_AMD_FLAGS  gfx90a   gfx908   gfx906  gfx900)
+
+#FIXME CAN BE REPLACED WITH LIST_ZIP IN CMAKE 3.17
+FOREACH(ARCH IN LISTS SUPPORTED_AMD_ARCHS)
+  LIST(FIND SUPPORTED_AMD_ARCHS ${ARCH} LIST_INDEX)
+  LIST(GET SUPPORTED_AMD_GPUS ${LIST_INDEX} GPU)
+  LIST(GET CORRESPONDING_AMD_FLAGS ${LIST_INDEX} FLAG)
+  KOKKOS_ARCH_OPTION(${ARCH}         GPU  "AMD GPU ${GPU} ${FLAG}"      "KOKKOS_SHOW_HIP_ARCHS")
```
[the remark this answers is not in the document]
JBludau: I figured, that this would be okay for a purely informational message given, but I can turn it into caps if wanted

cmake/kokkos_arch.cmake
JBludau: good catch ... forgot to delete when converting the function into a macro ... deleted the set in [50c17e](https://github.com/kokkos/kokkos/pull/5327/commits/50c17e17b94b5114eb9bf5eeb728ba25c4c885f5)

cmake/kokkos_arch.cmake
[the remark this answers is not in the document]
dalg24: This is fine as is

dalg24:
Retest this please

dalg24:
Failures are clearly unrelated (runtime failures on CUDA and OpenMPTarget builds)

dalg24 (approved):

PR #5331, opened 2022-08-12
Fixup let SIMD subpackage be unconditionally enabled since C++17 is now required

dalg24:


masterleinad (approved):

CMakeLists.txt
```diff
@@ -193,14 +193,8 @@ KOKKOS_SETUP_BUILD_ENVIRONMENT()
 
 OPTION(BUILD_SHARED_LIBS "Build shared libraries" OFF)
 
-SET(KOKKOS_EXT_LIBRARIES Kokkos::kokkos Kokkos::kokkoscore Kokkos::kokkoscontainers Kokkos::kokkosalgorithms)
```
crtrott, line 196: why is this gone?

crtrott (changes requested):
Why is KOKKOS_EXT_LIBRARIES now gone?

CMakeLists.txt
crtrott: what was this used for at all?

dalg24: Never used (I checked in Trilinos too)

dalg24: Also gripped for `_EXT_LIBRARIES` and packages don't define it

Rombur: It is used [here](https://github.com/dalg24/kokkos/blob/develop/cmake/KokkosConfig.cmake.in#L12)

Rombur: Sorry it was from an old commit. We don't use this anymore

ibaned (approved):

crtrott (approved):

PR #5169, opened 2022-06-30
Try deducing value type from return type in parallel reduce

masterleinad:
Fixes https://github.com/kokkos/kokkos/issues/5156, improves https://github.com/kokkos/kokkos/issues/2034.

Extending `FunctorAnalysis` to be able to be used with templated call operators appears to be very difficult. On the other hand, the overloads of `parallel_reduce` and `parallel_scan` taking a return value already contain a hint for the `value_type` to use so that we can avoid using the deduction in `FunctorAnalysis` in those cases.

This pull request does exactly that. If the functor has a `value_type` alias, we use that, otherwise, we use the `value_type` deduced from the return type and if that is `void`, too, we do the deduction in `FunctorAnalysis` based on the functor's call operator.

PR #5332, opened 2022-08-12
Pass FunctorAnalysis to ParallelReduce

masterleinad:
Passing `FunctorAnalysis` directly allows for simplifying the backend implementation and simplifies checking requirements on functor and reducer (even though nothing new is introduced here in this regard).
Based on #5169 but pretty much orthogonal. If we decide to merge the functionality in #5169 separately, we should merge #5169 first.
Apart from that (and possibly most of the cleanup in `FunctorAnalysis`), there is little that can be done incrementally. So it's best to just look at the pull requests as a whole.
Of course, it's helpful to review the backends separately.


core/src/Kokkos_Parallel_Reduce.hpp
```diff
@@ -1359,29 +1359,51 @@ struct ParallelReduceAdaptor {
...
+      auto reducer = return_value_adapter::return_value(return_value, functor);
+      typename Analysis::Reducer final_reducer(reducer);
+      ParallelReduceType closure(functor, inner_policy, final_reducer,
+                                 reducer.view());
+      Kokkos::Impl::shared_allocation_tracking_enable();
+      closure.execute();
+    }
+
+    Kokkos::Tools::Impl::end_parallel_reduce<PassedReducerType>(
+        inner_policy, functor, label, kpID);
```
masterleinad, line 1400: These are the most relevant general changes for this pull request.

dalg24:
Need to rebase

core/src/Kokkos_Parallel_Reduce.hpp
```diff
@@ -1359,29 +1359,51 @@ struct ParallelReduceAdaptor {
...
+    if constexpr (std::is_same_v<InvalidType, PassedReducerType>) {
+      typename Analysis::Reducer final_reducer(functor);
+      ParallelReduceType closure(
+          functor, inner_policy, final_reducer,
+          return_value_adapter::return_value(return_value, functor));
+      Kokkos::Impl::shared_allocation_tracking_enable();
+      closure.execute();
+    } else {
+      auto reducer = return_value_adapter::return_value(return_value, functor);
+      typename Analysis::Reducer final_reducer(reducer);
```
crtrott, line 1392: Are you potentially pass two functor copies on below? One for the reducer one for the functor? Are both of those reaching the device?

crtrott (changes requested):
I looked through the Parallel_Reduce, FunctorAnalysis and then a bit of Serial and CUDA RangePolicy. It seems like the functor is now twice copied to the device and part of the ParallelReducce (i.e. the m_functor and the m_reducer which now also potentially contains a copy of the functor?) I think we shouldn't do that in general.

core/src/Kokkos_Parallel_Reduce.hpp
```diff
@@ -1359,29 +1359,51 @@ struct ParallelReduceAdaptor {
                                   const PolicyType& policy,
                                   const FunctorType& functor,
                                   ReturnType& return_value) {
-    uint64_t kpID = 0;
+    using PassedReducerType = typename return_value_adapter::reducer_type;
```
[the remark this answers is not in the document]
masterleinad: The `FunctorAnalysis` object passed here is the same as we were using before. In case there is no reducer, we assume that the functor implements `init`, `join` and `final`. Passing `Sum` would change behavior by requiring that the type has a  `reduction_identity` specialization.

core/src/Kokkos_Parallel_Reduce.hpp
masterleinad: Yes, there are two functor copies if there is no reducer. 
If that is an issue, we would need to combine the functor and the `FunctorAnalysis` object into a different class that we would pass to the `ParallelReduce` implementations.

crtrott: Yeah that is definitely an issue. It effectively halfs the size available to hit the local_launch_path in CUDA/HIP and halfs the size available for constant cache usage.

masterleinad: I'll look into passing a wrapper for both the functor and the reducer to avoid having to copies (which will of course make this pull request even bigger).
If we just want to move the deduction of the value size from the functor signature to the return type first (to avoid doing to make changes in one pull request), we could just do https://github.com/kokkos/kokkos/pull/5169 first (even though changes to `ParallelReduce`'s template parameters will be reverted in this pull request again).

masterleinad: I added the wrapper in [`1c97d20` (#5332)](https://github.com/kokkos/kokkos/pull/5332/commits/1c97d20a84b9ae544928057dc2f7bd16d87f5410) and subsequent commits.

masterleinad:
The pull request is best understood looking at the changes to `Kokkos_Core_fwd.hpp` and then `Kokkos_Parallel_Reduce.hpp`:
- The new signature for the internal classes is
 ```
template <class CombinedFunctorReducerType, class ExecPolicy,
          class ExecutionSpace = typename Impl::FunctorPolicyExecutionSpace<
              typename CombinedFunctorReducerType::functor_type,
              ExecPolicy>::execution_space>
class ParallelReduce;
```
meaning that we pass a `CombinedFunctorReducer` instead of a `Functor` and a `Reducer` separately. That class is currently also defined in `Kokkos_Core_fwd.hpp` but should likely be moved to `FunctorAnalysis.hpp` in the end.
The wrapped functor is the usual functor and the reducer is the reducer wrapped in a `FunctorAnalysis` object (that contains a copy of the reducer). In case, the functor acts as a reducer, the encapsulation gives us the opportunity to avoid having two copies of the functor.  
- In `Kokkos_Parallel_Reduce.hpp`, we determine we basically only create the appropriate `CombinedFunctorReducer` object and call the backend implementations. Also, the `value_type` for the `FunctorAnalysis` object is set according to the return type (if possible). 
- The backend implementations only store a copy of the `CombinedFunctorReducer` object instead of a functor and that is used whenever access to the functor or reducer is required. Again, the value type comes from the wrapped `FunctorAnalysis` object.
- Setting the `value_type` in the `FunctorAnalysis` object as the return type and storing a copy instead of a functor, comes with a bunch of other changes. For the former one see, e.g.,  `cuda_single_inter_block_reduce_scan_shmem`. Changes related to the change in the return type deduction could relatively easily be separated, see #5169. Doing some after the remainder of the changes in this pull request would probably be even easier.

masterleinad:
Superseded by #5982 and #5976.

PR #4156, opened 2021-07-09
Update support for cuda reductions to work with types < 4 bytes in size

e10harvey:
Related to #4101. Merging this PR will merge #4101.

Fixes #4219.

PR #5334, opened 2022-08-12
#5333: CUDA: Use scratch space appropriate to small reduction elements in Team reductions

### LINKED ISSUE #5333: Fix word_size_type inconsistencies in CUDA Team reduction, maybe HIP
In #5324 it was noted that `word_size_type` is used in the `RangePolicy` reduction case, but not the `Team` case. In looking, it seems that's not covered in the HIP implementation for any kind of reduction. Figure this out.

#5975 is a hypothesized consequence of these inconsistencies

PhilMiller:
Follow-on for #4156

Fixes #5333

### DIFF
--- a/core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
@@ -625,6 +625,22 @@ class ParallelReduce<CombinedFunctorReducerType,
 
  public:
   using functor_type = FunctorType;
+  // Conditionally set word_size_type to int16_t or int8_t if value_type is
+  // smaller than int32_t (Kokkos::Cuda::size_type)
+  // word_size_type is used to determine the word count, shared memory buffer
+  // size, and global memory buffer size before the reduction is performed.
+  // Within the reduction, the word count is recomputed based on word_size_type
+  // and when calculating indexes into the shared/global memory buffers for
+  // performing the reduction, word_size_type is used again.
+  // For scalars > 4 bytes in size, indexing into shared/global memory relies
+  // on the block and grid dimensions to ensure that we index at the correct
+  // offset rather than at every 4 byte word; such that, when the join is
+  // performed, we have the correct data that was copied over in chunks of 4
+  // bytes.
+  using word_size_type = std::conditional_t<
+      sizeof(value_type) < sizeof(Kokkos::Cuda::size_type),
+      std::conditional_t<sizeof(value_type) == 2, int16_t, int8_t>,
+      Kokkos::Cuda::size_type>;
   using size_type    = Cuda::size_type;
   using reducer_type = ReducerType;
 
@@ -648,9 +664,11 @@ class ParallelReduce<CombinedFunctorReducerType,
   const pointer_type m_result_ptr;
   const bool m_result_ptr_device_accessible;
   const bool m_result_ptr_host_accessible;
-  size_type* m_scratch_space;
-  size_type* m_scratch_flags;
-  size_type* m_unified_space;
+  word_size_type* m_scratch_space;
+  // m_scratch_flags must be of type Cuda::size_type due to use of atomics
+  // for tracking metadata in Kokkos_Cuda_ReduceScan.hpp
+  Cuda::size_type* m_scratch_flags;
+  word_size_type* m_unified_space;
   size_type m_team_begin;
   size_type m_shmem_begin;
   size_type m_shmem_size;
@@ -694,13 +712,14 @@ class ParallelReduce<CombinedFunctorReducerType,
   }
 
   __device__ inline void run(SHMEMReductionTag&, const int& threadid) const {
-    const integral_nonzero_constant<
-        size_type, ReducerType::static_value_size() / sizeof(size_type)>
+    const integral_nonzero_constant<word_size_type,
+                                    ReducerType::static_value_size() /
+                                        sizeof(word_size_type)>
         word_count(m_functor_reducer.get_reducer().value_size() /
-                   sizeof(size_type));
+                   sizeof(word_size_type));
 
     reference_type value = m_functor_reducer.get_reducer().init(
-        kokkos_impl_cuda_shared_memory<size_type>() +
+        kokkos_impl_cuda_shared_memory<word_size_type>() +
         threadIdx.y * word_count.value);
 
     // Iterate this block through the league
@@ -723,18 +742,19 @@ class ParallelReduce<CombinedFunctorReducerType,
     if (!zero_length)
       do_final_reduction = cuda_single_inter_block_reduce_scan<false>(
           m_functor_reducer.get_reducer(), blockIdx.x, gridDim.x,
-          kokkos_impl_cuda_shared_memory<size_type>(), m_scratch_space,
+          kokkos_impl_cuda_shared_memory<word_size_type>(), m_scratch_space,
           m_scratch_flags);
 
     if (do_final_reduction) {
       // This is the final block with the final result at the final threads'
       // location
 
-      size_type* const shared = kokkos_impl_cuda_shared_memory<size_type>() +
-                                (blockDim.y - 1) * word_count.value;
+      word_size_type* const shared =
+          kokkos_impl_cuda_shared_memory<word_size_type>() +
+          (blockDim.y - 1) * word_count.value;
       size_type* const global =
           m_result_ptr_device_accessible
-              ? reinterpret_cast<size_type*>(m_result_ptr)
+              ? reinterpret_cast<word_size_type*>(m_result_ptr)
               : (m_unified_space ? m_unified_space : m_scratch_space);
 
       if (threadIdx.y == 0) {
@@ -805,13 +825,15 @@ class ParallelReduce<CombinedFunctorReducerType,
           1u, UseShflReduction ? std::min(m_league_size, size_type(1024 * 32))
                                : std::min(int(m_league_size), m_team_size));
 
-      m_scratch_space = cuda_internal_scratch_space(
-          m_policy.space(),
-          m_functor_reducer.get_reducer().value_size() * block_count);
+      m_scratch_space =
+          reinterpret_cast<word_size_type*>(cuda_internal_scratch_space(
+              m_policy.space(),
+              m_functor_reducer.get_reducer().value_size() * block_count));
       m_scratch_flags =
           cuda_internal_scratch_flags(m_policy.space(), sizeof(size_type));
-      m_unified_space = cuda_internal_scratch_unified(
-          m_policy.space(), m_functor_reducer.get_reducer().value_size());
+      m_unified_space =
+          reinterpret_cast<word_size_type*>(cuda_internal_scratch_unified(
+              m_policy.space(), m_functor_reducer.get_reducer().value_size()));
 
       dim3 block(m_vector_size, m_team_size, 1);
       dim3 grid(block_count, 1, 1);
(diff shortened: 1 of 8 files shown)


PhilMiller:
Things to consider
* [ ] HIP analogs
* [ ] `MDRangePolicy`

PhilMiller:
On CUDA, this passes at least the reduction unit tests, including under `cuda-memcheck`, so it's not completely mistaken.

PhilMiller:
On the other hand, it breaks compilation of `TestCuda_TeamBasic.cpp`

e10harvey (approved):

PhilMiller:
#5555 addresses this for at least some scan cases

PhilMiller:
The added test here does not fail without the later implementation change. I'm unsure why

core/unit_test/TestReducers.hpp
```diff
@@ -374,6 +381,62 @@ struct TestReducers {
...
+    {
+      using member_type = typename Kokkos::TeamPolicy<ExecSpace>::member_type;
+
+      Scalar sum_scalar;
+      Kokkos::View<Scalar, ExecSpace> sum_view("result");
+      Kokkos::deep_copy(sum_view, Scalar(1));
+
+      TeamSumFunctor tf;
+
+      if constexpr (sizeof(Scalar) > 1) {
```
e10harvey, line 393: Is there a separate test for `sizeof(Scalar) == 1`?

e10harvey (commented):
Please see questions. Let me know if you want to sit down and triage this together.

masterleinad:
You should probably use a test similar to the one added in https://github.com/kokkos/kokkos/pull/6296 to test the local memory implementation for Cuda and HIP (and small types of course).

cz4rs:
retest this please

cz4rs:
This actually works for me with 5dd2e4d on my desktop (Quadro P1000, cuda-12.2) :man_shrugging:. Further debugging ongoing...

core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
```diff
@@ -691,13 +709,14 @@ class ParallelReduce<CombinedFunctorReducerType,
...
+                                    ReducerType::static_value_size() /
+                                        sizeof(word_size_type)>
         word_count(m_functor_reducer.get_reducer().value_size() /
-                   sizeof(size_type));
+                   sizeof(word_size_type));
 
     reference_type value = m_functor_reducer.get_reducer().init(
-        kokkos_impl_cuda_shared_memory<size_type>() +
+        kokkos_impl_cuda_shared_memory<word_size_type>() +
         threadIdx.y * word_count.value);
```
masterleinad, line 720: This code path is never executed. It would be good to also have a test case that uses array reductions on small types (but that could be a follow-up).

cz4rs:
Retest this please.

cz4rs:
TODO:
- decide what to do with the introduced FIXMEs (some of the newly added tests are skipped for `OpenACC`, `OpenMPTarget`, `SYCL`, `point_t`, see the changes in `TestReducers.hpp` for details)
- [add more tests](https://github.com/kokkos/kokkos/pull/5334#issuecomment-1646097540), possibly after #6296 gets merged (?)
- follow-up for HIP after this PR is merged: https://github.com/kokkos/kokkos/pull/6461

I would still like to get some feedback on the PR in its current shape.

core/src/Cuda/Kokkos_Cuda_ReduceScan.hpp
```diff
@@ -103,7 +103,7 @@ template <class FunctorType>
 __device__ bool cuda_inter_block_reduction(
     typename FunctorType::reference_type value,
     typename FunctorType::reference_type neutral, const FunctorType& reducer,
-    Cuda::size_type* const m_scratch_space,
+    void* const m_scratch_space,
```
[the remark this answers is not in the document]
cz4rs: This seems to work when replacing `void` with the whole `std::conditional_t<sizeof(FunctorType::value_type)...` thing. I will check the tests and look for a nicer way to implement this.

core/unit_test/TestReducers.hpp
```diff
@@ -319,6 +329,83 @@ struct TestReducers {
...
+#ifdef KOKKOS_ENABLE_OPENMPTARGET
+    if constexpr (std::is_same_v<ExecSpace,
+                                 Kokkos::Experimental::OpenMPTarget>) {
+      return;  // FIXME_OPENMPTARGET
+    }
+#endif
+#ifdef KOKKOS_ENABLE_SYCL
+    if constexpr (std::is_same_v<ExecSpace, Kokkos::Experimental::SYCL>) {
+      return;  // FIXME_SYCL
+    }
```
masterleinad, line 362: What's the problem here? The `SYCL` implementation performs even the copies with `value_type`/`pointer_type` (and doesn't use `word_size_type`) so I wouldn't expect there to be a problem.

cz4rs: This is now restricted to skip `reducers_struct` (which uses `array_reduce<float, 1>` etc.) and 
`int8_t` tests.

For `int8_t` we get `CUDA_ERROR_MISALIGNED_ADDRESS` error:
```
5: [ RUN      ] sycl.reducers_int8_t
5: 
5: PI CUDA ERROR:
5: 	Value:           716
5: 	Name:            CUDA_ERROR_MISALIGNED_ADDRESS
5: 	Description:     misaligned address
5: 	Function:        operator()
5: 	Source Location: /root/intel-llvm-mirror/sycl/plugins/cuda/pi_cuda.cpp:2553
5: 
5: 
5: PI CUDA ERROR:
5: 	Value:           716
5: 	Name:            CUDA_ERROR_MISALIGNED_ADDRESS
5: 	Description:     misaligned address
5: 	Function:        cuda_piextUSMFree
5: 	Source Location: /root/intel-llvm-mirror/sycl/plugins/cuda/pi_cuda.cpp:4953
5: 
5: terminate called after throwing an instance of 'sycl::_V1::runtime_error'
5:   what():  Native API failed. Native API returns: -999 (Unknown PI error) -999 (Unknown PI error)
```

`reducers_struct` simply hangs and causes a timeout. I'll be looking closer into those.

core/src/Cuda/Kokkos_Cuda_ReduceScan.hpp
```diff
@@ -103,7 +103,7 @@ template <class FunctorType>
 __device__ bool cuda_inter_block_reduction(
     typename FunctorType::reference_type value,
     typename FunctorType::reference_type neutral, const FunctorType& reducer,
-    Cuda::size_type* const m_scratch_space,
+    typename FunctorType::pointer_type const m_scratch_space,
```
cz4rs, line 106: > `m_scratch_space` is only used as `pointer_type` below so it seems to make most sense to just `reinterpret_cast` the pointer to that type (the size of the memory pointed to by `pointer_type` is the same for the small value case). Ultimately, we use `word_size_type` when we want to copy between different allocations but as soon as we need the actual values (for `init`, `join`, `final`) we need to cast to `pointer_type`/`value_type` anyway. Thus, it might make sense to have `scratch_data` rather as `pointer_type` in the first place and only cast it to `word_size_type` when we copy it to other allocations.

_Originally posted by @masterleinad in https://github.com/kokkos/kokkos/pull/5334#discussion_r1353257202_

Switching to `pointer_type` fits nicely here, so I'm starting with that. I could also go ahead and change it in `Kokkos_Cuda_Parallel_Team.hpp` if that makes more sense :thinking:

core/unit_test/TestReducers.hpp
```diff
@@ -319,6 +329,89 @@ struct TestReducers {
...
+    if constexpr (std::is_same_v<ExecSpace, Kokkos::Experimental::OpenACC> &&
+                  (std::is_same_v<Scalar, size_t> ||
+                   std::is_same_v<Scalar, double>)) {
+      return;  // FIXME_OPENACC
+    }
+#endif
+#ifdef KOKKOS_ENABLE_OPENMPTARGET
+    if constexpr (std::is_same_v<ExecSpace,
+                                 Kokkos::Experimental::OpenMPTarget>) {
+      return;  // FIXME_OPENMPTARGET
```
cz4rs, line 356: TODO: narrow this down to skip the tests for specific types only (if there aren't too many of them).

masterleinad (approved):

core/unit_test/TestReducers.hpp
```diff
@@ -319,6 +329,89 @@ struct TestReducers {
...
+    if constexpr (std::is_same_v<ExecSpace, Kokkos::Experimental::SYCL> &&
+                  (std::is_same_v<Scalar, array_reduce<float, 1>> ||
+                   std::is_same_v<Scalar, array_reduce<float, 2>> ||
+                   std::is_same_v<Scalar, array_reduce<float, 3>> ||
+                   std::is_same_v<Scalar, array_reduce<float, 4>> ||
+                   std::is_same_v<Scalar, array_reduce<float, 7>> ||
+                   std::is_same_v<Scalar, int8_t>)) {
+      return;  // FIXME_SYCL
+    }
+#endif
```
masterleinad, line 369: Hmm.... I only see problems with `arry_reduce<float, 7>` on Intel GPUs.

cz4rs: That's right, I have removed the spurious cases (895201b1b9c69c755b7dac9777d6c0b4fa487755). Still not sure what's the root cause though.

masterleinad:
I pushed a fix for SYCL. For some reason, we were not using the same (proper) way of dealing with the `num_teams_done` variable as for `RangePolicy` and `MDRangePolicy`. Essentially, we were just appending to local memory used for group reduction results with a different type. That leads to unsuitable memory alignment for the `num_teams_done` variable. If preferred, I'm happy to push this commit to a follow-up pull request instead.

cz4rs:
> I pushed a fix for SYCL. For some reason, we were not using the same (proper) way of dealing with the `num_teams_done` variable as for `RangePolicy` and `MDRangePolicy`. Essentially, we were just appending to local memory used for group reduction results with a different type. That leads to unsuitable memory alignment for the `num_teams_done` variable. If preferred, I'm happy to push this commit to a follow-up pull request instead.

Thanks for this! I think there's no need for another PR as long as we can deal with the [CI timeout](https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/Kokkos/detail/Kokkos/14713/pipeline/49):
```
5: [ RUN      ] sycl.reducers_struct
 5/58 Test  #5: Kokkos_CoreUnitTest_SYCL1B ....................***Timeout 1500.19 sec
```

cz4rs:
Retest this please.

masterleinad:
> Looks like the `array_reduce<float, 7>` is still not working. I'll debug that more and restore that guard for that test.

I pushed another fix that somewhat hides the real issue but makes the test work and I expect that fix to be part of the proper solution (which should have its own pull request).

cz4rs:
> I pushed another fix that somewhat hides the real issue but makes the test work and I expect that fix to be part of the proper solution (which should have its own pull request).

Alright! This would only leave FIXMEs for `OpenACC` and `point_t`. I will take a closer look at `point_t` part to at least narrow it down further.

cz4rs:
@masterleinad The only `FIXME` left is for `OpenACC` (`size_t` and `double`). I look forward to getting this merged and dealing with `OpenACC` / adding more tests in a follow-up.

masterleinad:
> @masterleinad The only `FIXME` left is for `OpenACC` (`size_t` and `double`). I look forward to getting this merged and dealing with `OpenACC` / adding more tests in a follow-up.

I have approved it already. 🙂

cz4rs:
Retest this please.

cz4rs:
Retest this please.

cz4rs:
Retest this please.

cz4rs:
Retest this please.

cz4rs:
Rebased on top of `develop` to get fresh Jenkins logs. CI looks good, this needs another approval.

crtrott:
Its not quite clear to me why the point_t default constructor is changed, considering that there is a `reduction_identity` overload. Maybe the team reduce does not use reduction identity??

PR #5335, opened 2022-08-14
OpenACC Parallel Constructs for MDRange Policy

seyonglee:
Add OpenACC/Kokkos_OpenACC_ParallelFor_MDRange.hpp    
Add OpenACC/Kokkos_OpenACC_ParallelReduce_MDRange.hpp
Add KOKKOS_IF_ON_DEVICE and KOKKOS_IF_ON_HOST definitions for the OpenACC backend (Kokkos_Macros.hpp).
Implement device_abort() function for the OpenACC backend (impl/Kokkos_Error.hpp).


core/src/OpenACC/Kokkos_OpenACC_ParallelReduce_MDRange.hpp
```diff
@@ -0,0 +1,1510 @@
...
+
+namespace Kokkos::Experimental::Impl {
+
+// primary template: catch-all non-implemented custom reducers
+template <class Functor, class Reducer, class Policy,
+          bool = std::is_arithmetic_v<typename Reducer::value_type>>
+struct OpenACCParallelReduceMDRangeHelper {
+  OpenACCParallelReduceMDRangeHelper(Functor const&, Reducer const&,
+                                     Policy const&) {
+    static_assert(std::is_void_v<Functor>, "not implemented");
```
Rombur, line 61: Can you improve the error message.

Rombur (approved):

core/src/OpenACC/Kokkos_OpenACC_ParallelFor_MDRange.hpp
```diff
@@ -0,0 +1,243 @@
...
+      }
+    }
+  }
+
+  void execute() const {
+    static_assert(Policy::rank < 7,
+                  "OpenACC Backend MDRangePolicy Error: Unsupported rank...");
+    execute_impl<Policy::rank>();
+  }
+};
```
JBludau, line 241: Please say that only 1<ranks<7 are supported. Rank=1 should trigger the static assert too.

core/src/Kokkos_Macros.hpp
```diff
@@ -563,6 +567,29 @@ static constexpr bool kokkos_omp_on_host() { return false; }
...
+// FIXE acc_on_device is a non-constexpr function
+#define KOKKOS_IF_ON_DEVICE(CODE)                     \
+  if constexpr (acc_on_device(acc_device_not_host)) { \
+    KOKKOS_IMPL_STRIP_PARENS(CODE)                    \
+  }
+#define KOKKOS_IF_ON_HOST(CODE)                   \
+  if constexpr (acc_on_device(acc_device_host)) { \
+    KOKKOS_IMPL_STRIP_PARENS(CODE)                \
+  }
+#endif
```
dalg24, line 590: This does not actually compile

seyonglee: This code is for other OpenACC compilers. In OpenACC, the only portable way to check whether a code is running on a host or a device is to use `acc_on_device` function, which can be converted into a compile-time constant by the OpenACC compiler. Unfortunately, this compile-time conversion is not guaranteed even if the input argument is a compile-time constant, and thus this code may not work in some OpenACC compilers (just best-effort approach for now).

dalg24: Do we know of any compiler that supports it?

core/src/OpenACC/Kokkos_OpenACC_ParallelReduce_MDRange.hpp
seyonglee: Updated the error message.

core/src/OpenACC/Kokkos_OpenACC_ParallelFor_MDRange.hpp
seyonglee: Updated.

core/src/Kokkos_Macros.hpp
seyonglee: I have not checked it; the OpenACC spec says that it can be evaluated to a constant at compile time, but it is optional.

core/src/OpenACC/Kokkos_OpenACC_ParallelFor_MDRange.hpp
```diff
@@ -114,7 +114,8 @@ struct Kokkos::Experimental::Impl::OpenACCParallelForHelper<
         }
       }
     } else {
-      static_assert("implementation bug");
+      static_assert(
+          "Kokkos Error: not supported memory layout for the OpenACC backend");
```
dalg24, line 118: No, I really meant that comment.  If this else-clause gets compiled there is a bug, it is not a matter of being supported or not.

core/src/Kokkos_Macros.hpp
```diff
@@ -563,6 +568,29 @@ static constexpr bool kokkos_omp_on_host() { return false; }
 #endif
 #endif
 
+#ifdef KOKKOS_ENABLE_OPENACC
+#ifdef KOKKOS_COMPILER_NVHPC
+#define KOKKOS_IF_ON_DEVICE(CODE)   \
+  if (__builtin_is_device_code()) { \
+    KOKKOS_IMPL_STRIP_PARENS(CODE)  \
+  }
```
dalg24, line 576: I suspect that this means the fallback to host if no device is detected won't actually work

core/src/OpenACC/Kokkos_OpenACC_ParallelFor_MDRange.hpp
```diff
@@ -0,0 +1,338 @@
...
+#pragma acc parallel loop gang vector collapse(2) copyin(functor) \
+    async(async_arg)
+      for (auto i0 = begin0; i0 < end0; ++i0) {
+        for (auto i1 = begin1; i1 < end1; ++i1) {
+          functor(i0, i1);
+        }
+      }
+    } else {
+      static_assert(
+          "Kokkos Error: not supported memory layout for the OpenACC backend");
```
masterleinad, line 118: Did you mean
```suggestion
      static_assert(false,
          "Kokkos Error: not supported memory layout for the OpenACC backend");
```
? Any string implicitly converts to `true`.

seyonglee: Thank you for catching this; updated.

core/src/OpenACC/Kokkos_OpenACC_ParallelFor_MDRange.hpp
seyonglee: Updated the comment.

core/src/OpenACC/Kokkos_OpenACC_ParallelFor_MDRange.hpp
seyonglee: Changed `static_assert(false, …)` to `Kokkos::abort(…)` since `static_assert(false, …)` will always complain errors.

core/src/Kokkos_Macros.hpp
```diff
@@ -489,6 +492,9 @@
 #define KOKKOS_ENABLE_DEFAULT_DEVICE_TYPE_HIP
 #elif defined(KOKKOS_ENABLE_SYCL)
 #define KOKKOS_ENABLE_DEFAULT_DEVICE_TYPE_SYCL
+// FIXME_OPENACC: will be enabled after fixing bugs
+//#elif defined(KOKKOS_ENABLE_OPENACC)
+//#define KOKKOS_ENABLE_DEFAULT_DEVICE_TYPE_OPENACC
```
seyonglee, line 497: Setting the default device type to the OpenACC device is temporarily disabled, since there are many tests running on the default device, which will fail if using features not implemented in the OpenACC backend.
This default device type setting for the OpenACC backend will be re-enabled after all the basic features are implemented.

core/src/Kokkos_Macros.hpp
seyonglee: There seems to be a bug in `__builtin_is_device_code`, which is not allowed to be called on host (https://forums.developer.nvidia.com/t/nvc-up-to-22-7-undefined-reference-to-builtin-is-device-code-for-empty-if-target-paths/223232). Therefore, we will have to either wait for the next release or use `acc_on_device` API.

core/src/Kokkos_Macros.hpp
dalg24: We absolutely need it for View accesses.  The incremental only pass because `Kokkos::abort` does nothing on the device but the standard output is cluttered with memory access violation error messages.

seyonglee: I created another local branch, which enables the default OpenACC execution space and disable tests evaluating not-implemented OpenACC features. the new branch executes the incremental tests correctly. I will submit the new PR after the this PR (#5335) is merged.

dalg24: I would do it the other way around.  They are many changes on this PR that are controversial.

seyonglee: What you mean is to submit a new PR fixing on the default OpenACC execution space before this PR (#5335)?

dalg24: Yes

seyonglee: Submitted a new PR (#5360) for the default OpenACC execution space.

pedrovalerolara (approved):

core/src/OpenACC/Kokkos_OpenACC_ParallelFor_MDRange.hpp
```diff
@@ -0,0 +1,724 @@
...
+                                     OpenACCMDRangeTile<2> const& tile,
+                                     int async_arg) {
+  int tile0  = tile[0];
+  int tile1  = tile[1];
+  int begin1 = begin[1];
+  int end1   = end[1];
+  int begin0 = begin[0];
+  int end0   = end[0];
+// clang-format off
+#pragma acc parallel loop gang vector tile(tile0,tile1) copyin(functor) async(async_arg)
```
seyonglee, line 120: OpenACC says that the tile size should be a constant positive integer expression, but `tile0` and `tile1` are not constant.

Rombur:
The PR needs to be rebased on develop to fix the conflict

core/src/OpenACC/Kokkos_OpenACC_ParallelFor_MDRange.hpp
dalg24: runtime or compile-time constant?

seyonglee:
> The PR needs to be rebased on develop to fix the conflict

Rebased; thanks!

core/src/OpenACC/Kokkos_OpenACC_ParallelFor_MDRange.hpp
seyonglee: compile-time constant.

core/src/Kokkos_Macros.hpp
seyonglee: PR #5360 is merged into the develop branch and also merged into this PR as a part of rebasing.

JBludau (approved):
looks good

dalg24 (changes requested):
Superseded by #5376

PR #5338, opened 2022-08-15
Fix missing linker language for SIMD

crtrott:
In algorithms we worked around that with a dummy cpp file:

Error:
```
CMake Error: CMake can not determine linker language for target: kokkossimd
CMake Error in kokkos/simd/src/CMakeLists.txt:
  Exporting the target "kokkossimd" is not allowed since its linker language
  cannot be determined
```


dalg24 (approved):
For the record that is an alternative to the `dummy.cpp` file in algorithms.
I also like it better than the "dummy" source file.

CMakeLists.txt
```diff
@@ -67,7 +67,13 @@ ENDFUNCTION()
 
 LIST(APPEND CMAKE_MODULE_PATH cmake/Modules)
 
-IF(NOT KOKKOS_HAS_TRILINOS)
+
+IF(KOKKOS_HAS_TRILINOS)
+  # What language are we compiling Kokkos as
+  # downstream dependencies need to match this!
+  # We do not supporting using CUDA or HIP as cmake language yet in Trilinos
+  SET(KOKKOS_COMPILE_LANGUAGE CXX)
```
PhilMiller, line 75: It looks like this appears in both the `IF` and `ELSE` branches - why not lift it up above?

crtrott: Actually it gets reset again in the else branch if you have CUDA on, I thought its cleaner if the Trilinos path makes clear we always set it to CXX and it doesn't get messed with later again.

PhilMiller: Works for me

PhilMiller (approved):

PhilMiller:
Given the CMake failures in Jenkins, you may have to do the `dummy.cpp` thing

CMakeLists.txt
```diff
@@ -67,7 +67,13 @@ ENDFUNCTION()
 
 LIST(APPEND CMAKE_MODULE_PATH cmake/Modules)
 
-IF(NOT KOKKOS_HAS_TRILINOS)
+
+IF(KOKKOS_HAS_TRILINOS)
+  # What language are we compiling Kokkos as
+  # downstream dependencies need to match this!
+  # We do not supporting using CUDA or HIP as cmake language yet in Trilinos
```
dalg24, line 74:
```suggestion
  # We do not support using CUDA or HIP as CMake language yet in Trilinos
```

PR #5339, opened 2022-08-15
Fix missing linker language for SIMD

crtrott:
In algorithms we worked around that with a dummy cpp file


crtrott:
Companion to #5338

dalg24 (approved):

CMakeLists.txt
```diff
@@ -67,7 +67,13 @@ ENDFUNCTION()
 
 LIST(APPEND CMAKE_MODULE_PATH cmake/Modules)
 
-IF(NOT KOKKOS_HAS_TRILINOS)
+
+IF(KOKKOS_HAS_TRILINOS)
+  # What language are we compiling Kokkos as
+  # downstream dependencies need to match this!
+  # We do not supporting using CUDA or HIP as cmake language yet in Trilinos
```
masterleinad, line 74:
```suggestion
  # We do not support using CUDA or HIP as CMake language yet in Trilinos
```

PR #5344, opened 2022-08-16
Fix UnorderedMapRehash::operator()

masterleinad:
https://kokkos.github.io/kokkos-core-wiki/API/containers/Unordered-Map.html?highlight=unorderedmap
> An UnorderedMap behaves like an unordered set if the template parameter Value is void.

In particular, https://github.com/kokkos/kokkos/pull/5297 removed the deprecated `value_at` if the value type is `void`. https://github.com/kokkos/kokkos/pull/4639 should have touched this code as well (but we apparently don't have a test using it).


crtrott (approved):

crtrott:
we may need to backport a sfinae version of this to 3.7.00 so that it works with  deprecated code off ....

stanmoore1:
Thanks @masterleinad!

Rombur (approved):

PR #5345, opened 2022-08-17
Fix null deref in SharedAllocationRecord::print_records

brian-kelley:
For some memory spaces, the doubly linked list of SharedAllocationRecords starts off with a special root node that doesn't have an actual allocation associated with it (``m_alloc_ptr == nullptr``). So trying to print out ``r->m_alloc_ptr->m_label`` will segfault.

Example reproducer:
```
#include <Kokkos_Core.hpp>
  Kokkos::DefaultExecutionSpace{}.print_configuration(std::cout);
  {
    Kokkos::Impl::SharedAllocationRecord<Kokkos::HostSpace>::print_records( std::cout, Kokkos::HostSpace(), true );
  }
  Kokkos::finalize();
  return 0;
} 
```
Without this PR, the above segfaults. With this PR:
```
Host Serial Execution Space:
  KOKKOS_ENABLE_SERIAL: yes
Serial Atomics:
  KOKKOS_ENABLE_SERIAL_ATOMICS: no

Serial Runtime Configuration:
Host addr( 0x00000067ffa0 ) list( 0x00000067ffa0 0x00000067ffa0 ) extent[ 0x000000000000 + 00000000 ] count(0) dealloc(0x000000000000) [NOT ALLOCATED]
```
(``Kokkos_ENABLE_DEBUG`` needs to be ON for ``print_records`` to work).

Interestingly, this issue seemed to affect HostSpace and CudaUVMSpace, but not CudaSpace. In CudaSpace, this root node still has a null data pointer and 0 extent, but the m_alloc_ptr is NOT null.

### DIFF
--- a/core/src/impl/Kokkos_SharedAlloc.cpp
+++ b/core/src/impl/Kokkos_SharedAlloc.cpp
@@ -338,28 +338,24 @@ void SharedAllocationRecord<void, void>::print_host_accessible_records(
                reinterpret_cast<uintptr_t>(r->m_next),
                reinterpret_cast<uintptr_t>(r->m_alloc_ptr), r->m_alloc_size,
                r->use_count(), reinterpret_cast<uintptr_t>(r->m_dealloc),
-               r->m_alloc_ptr ? r->m_alloc_ptr->m_label : " [NOT ALLOCATED]");
+               r->m_alloc_ptr->m_label);
       s << buffer;
       r = r->m_next;
     }
   } else {
     while (r != root) {
-      if (r->m_alloc_ptr) {
-        // Formatting dependent on sizeof(uintptr_t)
-        const char* format_string;
+      // Formatting dependent on sizeof(uintptr_t)
+      const char* format_string;
 
-        if (sizeof(uintptr_t) == sizeof(unsigned long)) {
-          format_string = "%s [ 0x%.12lx + %ld ] %s\n";
-        } else if (sizeof(uintptr_t) == sizeof(unsigned long long)) {
-          format_string = "%s [ 0x%.12llx + %ld ] %s\n";
-        }
-
-        snprintf(buffer, 256, format_string, space_name,
-                 reinterpret_cast<uintptr_t>(r->data()), r->size(),
-                 r->m_alloc_ptr->m_label);
-      } else {
-        snprintf(buffer, 256, "%s [ 0 + 0 ]\n", space_name);
+      if (sizeof(uintptr_t) == sizeof(unsigned long)) {
+        format_string = "%s [ 0x%.12lx + %ld ] %s\n";
+      } else if (sizeof(uintptr_t) == sizeof(unsigned long long)) {
+        format_string = "%s [ 0x%.12llx + %ld ] %s\n";
       }
+
+      snprintf(buffer, 256, format_string, space_name,
+               reinterpret_cast<uintptr_t>(r->data()), r->size(),
+               r->m_alloc_ptr->m_label);
       s << buffer;
       r = r->m_next;
     }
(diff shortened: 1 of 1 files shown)


masterleinad (approved):
I think it would also make sense to skip the root node for printing.

brian-kelley:
@masterleinad I like that better, just pushed it. I also had missed the same issue in the non-detailed path but that's fixed.

Rombur (approved):

brian-kelley:
I just noticed that the non-detailed path already checks ``if (r->m_alloc_ptr)``. Is it possible for records other than root to not be allocated (I would imagine no, they should get removed when freed). If that understanding is right, I could remove a couple lines from the non-detailed path.

masterleinad:
> I just noticed that the non-detailed path already checks `if (r->m_alloc_ptr)`. Is it possible for records other than root to not be allocated? (I would imagine no, they should get removed from the list when freed). If that understanding is right, I could remove a couple lines from the non-detailed path.

It doesn't hurt having the safeguard of handling non-allocated records since this function is mainly for debugging anyway.

brian-kelley:
@masterleinad Sounds good, I just added the safeguard back in (and still skip the root node).

This is out of scope for this PR, and I'm not sure if it's a real issue but I notice that if detail=true, the printed size includes the header, and if it's false, the size does not include the header.

For example, here's how it prints with true and false for a 4000-byte View:
```
CudaUVM addr( 0x000002a7ea50 ) list( 0x00000071b3e0 0x00000071b3e0 ) extent[ 0x7f983a000000 + 00004128 ] count(1) dealloc(0x00000040b175) asdf
CudaUVM [ 0x7f983a000080 + 4000 ] asdf
```
I wouldn't mind making another PR to get this consistent - maybe the non-detailed path should also include the header size, since [ptr, ptr + sizeof(header) + dataSize] is the actual extent of the total memory?

crtrott:
Retest this please.

core/src/impl/Kokkos_SharedAlloc.cpp
```diff
@@ -336,12 +338,12 @@ void SharedAllocationRecord<void, void>::print_host_accessible_records(
                reinterpret_cast<uintptr_t>(r->m_next),
                reinterpret_cast<uintptr_t>(r->m_alloc_ptr), r->m_alloc_size,
                r->use_count(), reinterpret_cast<uintptr_t>(r->m_dealloc),
-               r->m_alloc_ptr->m_label);
+               r->m_alloc_ptr ? r->m_alloc_ptr->m_label : " [NOT ALLOCATED]");
```
dalg24, line 341: Do we still need that ternary with the excluding the root?
If so you missed L359, otherwise drop that change.

brian-kelley: Probably not, but like Daniel said it doesn't hurt to have the safeguard. If the record headers get corrupted, this could be the difference between a segfault and printing something reasonable to help debug.

dalg24: I do not agree.  At least if it segfault you will notice and the code is less complex

brian-kelley: @masterleinad I don't have a strong feeling either way, want to just take it out?

masterleinad: I still prefer avoiding the segfault and printing "[NOT ALLOCATED]". I won't request changes if you take it out, though.

brian-kelley: Wait no, @dalg24 this is already consistent. Line 359 doesn't need the ternary because the non-detailed path checks ``if (r->m_alloc_ptr)`` on 347, and if that is null it prints "<space> [ 0 + 0 ]". I vote to just keep it how it is

dalg24: > I still prefer avoiding the segfault and printing "[NOT ALLOCATED]". I won't request changes if you take it out, though.

If the intent when not processing the root is that we will never have a "non allocated pointer", then the silently skip `nullptr` or print `"[NOT ALLOCATED]"` makes that we would not realize if `nullptr` are introduced in the linked list.
If they occur in normal operation and it absolutely must be handled (which you don't clearly state) then sure.

brian-kelley: I ended up going @dalg24 's preferred way.

dalg24 (approved):

PR #5350, opened 2022-08-17
Revert "Fix missing linker language for SIMD"

dalg24:
Reverts kokkos/kokkos#5338


dalg24:
For the record we messed up, CI did not pass
the `LINKER_LANGUAGE` property on targets is only available since CMake 3.19

PR #5352, opened 2022-08-17
Simd fix linker language release

crtrott:


crtrott:
Closing since it doesn't work with pre cmake 3.19 which we need for Trilinos ...

PR #5353, opened 2022-08-17
Fix linker language simd

crtrott:


cz4rs (approved):
:crossed_fingers:

crtrott:
Closing since it doesn't work with pre 3.19 cmake in Trilinos

PR #5355, opened 2022-08-17
Add dummy source file to SIMD to allow deduction of linker language.

crtrott:
This is really only necessary for Trilinos ...

Replaces #5352


ndellingwood (approved):

simd/src/Kokkos_SIMD_dummy.cpp
```diff
@@ -0,0 +1,7 @@
+// This file is need in order to get the linker language
```
masterleinad, line 1:
```suggestion
// This file is needed in order to get the linker language
```

fnrizzi (approved):

crtrott:
The failure is unrelated since its in OpenACC incremental testing.

PR #5365, opened 2022-08-19
Bugfix preserve view label when calling `realloc()`

### LINKED ISSUE #5364: `Kokkos::realloc` drops the view label
Bug present 3.7 RC but not in 3.6
```C++
  Kokkos::View<int*> v("v", 0);
  Kokkos::realloc(v, 1);
  ASSERT_EQ(v.label(), "v");
```
yields
```
Failure
  v.label()
    Which is: ""
  "v"
```

dalg24:
Fix #5364

### DIFF
--- a/core/src/Kokkos_CopyViews.hpp
+++ b/core/src/Kokkos_CopyViews.hpp
@@ -3278,26 +3278,25 @@ impl_realloc(Kokkos::View<T, P...>& v, const size_t n0, const size_t n1,
   const bool sizeMismatch = Impl::size_mismatch(v, v.rank_dynamic, new_extents);
 
   if (sizeMismatch) {
-    v = view_type();  // Deallocate first, if the only view to allocation
-    v = view_type(arg_prop, n0, n1, n2, n3, n4, n5, n6, n7);
+    auto const& label = v.label();  // Save the label
+    v = view_type();  // Best effort to deallocate in case no other view refers
+                      // to the shared allocation
+    using alloc_prop = Impl::ViewCtorProp<ViewCtorArgs..., std::string>;
+    alloc_prop arg_prop_copy(arg_prop);
+    static_cast<Kokkos::Impl::ViewCtorProp<void, std::string>&>(arg_prop_copy)
+        .value = label;
+    v          = view_type(arg_prop_copy, n0, n1, n2, n3, n4, n5, n6, n7);
   } else if (alloc_prop_input::initialize) {
     if (alloc_prop_input::has_execution_space) {
-      // Add execution_space if not provided to avoid need for if constexpr
       using alloc_prop = Impl::ViewCtorProp<
           ViewCtorArgs...,
           std::conditional_t<alloc_prop_input::has_execution_space,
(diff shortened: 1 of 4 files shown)


core/src/Kokkos_CopyViews.hpp
```diff
@@ -3278,26 +3278,25 @@ impl_realloc(Kokkos::View<T, P...>& v, const size_t n0, const size_t n1,
...
-      using execution_space_type = typename alloc_prop::execution_space;
-      const execution_space_type& exec_space =
-          static_cast<
-              Kokkos::Impl::ViewCtorProp<void, execution_space_type> const&>(
-              arg_prop_copy)
-              .value;
+      auto const& exec_space = static_cast<Kokkos::Impl::ViewCtorProp<
+          void, typename alloc_prop::execution_space> const&>(arg_prop_copy)
+                                   .value;
       Kokkos::deep_copy(exec_space, v, typename view_type::value_type{});
```
masterleinad, line 3300: #5314 will clean this up some more since we don't really need to modify `arg_prop` when we can use `if constexpr`.

core/src/Kokkos_CopyViews.hpp
```diff
@@ -3278,26 +3278,25 @@ impl_realloc(Kokkos::View<T, P...>& v, const size_t n0, const size_t n1,
...
-    v = view_type();  // Deallocate first, if the only view to allocation
-    v = view_type(arg_prop, n0, n1, n2, n3, n4, n5, n6, n7);
+    auto const& label = v.label();  // Save the label
+    v = view_type();  // Best effort to deallocate in case no other view refers
+                      // to the shared allocation
+    using alloc_prop = Impl::ViewCtorProp<ViewCtorArgs..., std::string>;
+    alloc_prop arg_prop_copy(arg_prop);
+    static_cast<Kokkos::Impl::ViewCtorProp<void, std::string>&>(arg_prop_copy)
+        .value = label;
+    v          = view_type(arg_prop_copy, n0, n1, n2, n3, n4, n5, n6, n7);
```
masterleinad, line 3288: Why not just
```suggestion
    using alloc_prop = Impl::ViewCtorProp<ViewCtorArgs..., std::string>;
    alloc_prop arg_prop_copy(arg_prop);
    static_cast<Kokkos::Impl::ViewCtorProp<void, std::string>&>(arg_prop_copy)
        .value = v.label();  // Save the label
    v = view_type();  // Best effort to deallocate in case no other view refers
                      // to the shared allocation
    v          = view_type(arg_prop_copy, n0, n1, n2, n3, n4, n5, n6, n7);
```

masterleinad (commented):
Thanks for fixing!

core/src/Kokkos_CopyViews.hpp
dalg24: Sure

Rombur (approved):

masterleinad (approved):

dalg24:
Failure (Clang+CUDA build hanging) is unrelated

PR #3929, opened 2021-04-12
Consider std::sort when view is accessible from host

dalg24:
See https://github.com/kokkos/kokkos/issues/3864#issuecomment-8170879

PR #5372, opened 2022-08-22
Dispatch Kokkos::sort(Kokkos::View) to std::sort

jstrzebonski:

As @dalg24 suggested (https://github.com/kokkos/kokkos/pull/5183#discussion_r950460815) I removed `always_use_kokkos_sort` argument. 

`std::sort()` is used ~~only for SERIAL~~ for any execution space that can access host_space.

### DIFF
--- a/algorithms/src/Kokkos_Sort.hpp
+++ b/algorithms/src/Kokkos_Sort.hpp
@@ -630,7 +630,6 @@ std::enable_if_t<(Kokkos::is_execution_space<ExecutionSpace>::value) &&
                                     memory_space>::accessible)>
 sort(const ExecutionSpace& space,
      const Kokkos::View<DataType, Properties...>& view) {
-  space.fence("Kokkos::sort: Fence before sorting on the host");
   auto first = Experimental::begin(view);
   auto last  = Experimental::end(view);
   std::sort(first, last);
--- a/algorithms/src/Kokkos_Sort.hpp
+++ b/algorithms/src/Kokkos_Sort.hpp
@@ -580,7 +580,7 @@ struct min_max_functor {
 
 template <class ExecutionSpace, class DataType, class... Properties>
 std::enable_if_t<(Kokkos::is_execution_space<ExecutionSpace>::value) &&
-                 (not SpaceAccessibility<
+                 (!SpaceAccessibility<
                      HostSpace, typename Kokkos::View<DataType, Properties...>::
                                     memory_space>::accessible)>
 sort(const ExecutionSpace& exec,
--- a/algorithms/src/Kokkos_Sort.hpp
+++ b/algorithms/src/Kokkos_Sort.hpp
@@ -585,7 +585,7 @@ std::enable_if_t<(Kokkos::is_execution_space<ExecutionSpace>::value) &&
                                     memory_space>::accessible)>
 sort(const ExecutionSpace& exec,
      const Kokkos::View<DataType, Properties...>& view) {
-  using ViewType = typename Kokkos::View<DataType, Properties...>;
+  using ViewType = Kokkos::View<DataType, Properties...>;
   using CompType = BinOp1D<ViewType>;
 
   Kokkos::MinMaxScalar<typename ViewType::non_const_value_type> result;
--- a/algorithms/src/Kokkos_Sort.hpp
+++ b/algorithms/src/Kokkos_Sort.hpp
@@ -53,6 +53,10 @@
 
 #include <algorithm>
 
+#if defined(KOKKOS_ENABLE_SERIAL)
+#include "std_algorithms/Kokkos_BeginEnd.hpp"
+#endif
+
 namespace Kokkos {
 
 namespace Impl {
@@ -615,6 +619,17 @@ std::enable_if_t<Kokkos::is_execution_space<ExecutionSpace>::value> sort(
   bin_sort.create_permute_vector(exec);
   bin_sort.sort(exec, view);
 }
+
+#if defined(KOKKOS_ENABLE_SERIAL)
+template <class DataType, class... Properties>
+void sort(const Serial& space,
+          const Kokkos::View<DataType, Properties...>& view) {
+  space.fence("Kokkos::sort: Fence before sorting on the host");
+  auto first = Experimental::begin(view);
+  auto last  = Experimental::end(view);
+  std::sort(first, last);
+}
+#endif
 
 template <class ViewType>
 void sort(ViewType const& view) {
--- a/algorithms/src/Kokkos_Sort.hpp
+++ b/algorithms/src/Kokkos_Sort.hpp
@@ -628,8 +628,7 @@ std::enable_if_t<(Kokkos::is_execution_space<ExecutionSpace>::value) &&
                  (SpaceAccessibility<
                      HostSpace, typename Kokkos::View<DataType, Properties...>::
                                     memory_space>::accessible)>
-sort(const ExecutionSpace& space,
-     const Kokkos::View<DataType, Properties...>& view) {
+sort(const ExecutionSpace&, const Kokkos::View<DataType, Properties...>& view) {
   auto first = Experimental::begin(view);
   auto last  = Experimental::end(view);
   std::sort(first, last);
--- a/algorithms/src/Kokkos_Sort.hpp
+++ b/algorithms/src/Kokkos_Sort.hpp
@@ -53,9 +53,7 @@
 
 #include <algorithm>
 
-#if defined(KOKKOS_ENABLE_SERIAL)
 #include "std_algorithms/Kokkos_BeginEnd.hpp"
-#endif
 
 namespace Kokkos {
 
@@ -580,9 +578,14 @@ struct min_max_functor {
 
 }  // namespace Impl
 
-template <class ExecutionSpace, class ViewType>
-std::enable_if_t<Kokkos::is_execution_space<ExecutionSpace>::value> sort(
-    const ExecutionSpace& exec, ViewType const& view) {
+template <class ExecutionSpace, class DataType, class... Properties>
+std::enable_if_t<(Kokkos::is_execution_space<ExecutionSpace>::value) &&
+                 (not SpaceAccessibility<
+                     HostSpace, typename Kokkos::View<DataType, Properties...>::
+                                    memory_space>::accessible)>
+sort(const ExecutionSpace& exec,
+     const Kokkos::View<DataType, Properties...>& view) {
+  using ViewType = typename Kokkos::View<DataType, Properties...>;
   using CompType = BinOp1D<ViewType>;
 
   Kokkos::MinMaxScalar<typename ViewType::non_const_value_type> result;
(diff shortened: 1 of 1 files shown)


algorithms/src/Kokkos_Sort.hpp
```diff
@@ -596,10 +578,21 @@ std::enable_if_t<Kokkos::is_execution_space<ExecutionSpace>::value> sort(
...
 }
 
+#if defined(KOKKOS_ENABLE_SERIAL)
+template <class DataType, class... Properties>
+void sort(const Serial& space,
+          const Kokkos::View<DataType, Properties...>& view) {
+  space.fence("Kokkos::sort: Fence before sorting on the host");
+  auto first = Experimental::begin(view);
+  auto last  = Experimental::end(view);
+  std::sort(first, last);
```
jstrzebonski, line 588: Should I use `parallel_policy` or `parallel_unsequenced_policy`?

dalg24: No

algorithms/src/Kokkos_Sort.hpp
```diff
@@ -596,10 +578,21 @@ std::enable_if_t<Kokkos::is_execution_space<ExecutionSpace>::value> sort(
   bin_sort.sort(exec, view);
 }
 
+#if defined(KOKKOS_ENABLE_SERIAL)
+template <class DataType, class... Properties>
+void sort(const Serial& space,
+          const Kokkos::View<DataType, Properties...>& view) {
+  space.fence("Kokkos::sort: Fence before sorting on the host");
```
masterleinad, line 585: Why would you need this fence (which is a no-op anyway)?

masterleinad (commented):
What about other execution spaces that can access `HostSpace`? Note in particular that https://github.com/kokkos/kokkos/pull/3929 introduced using `std::sort` for `CudaUVM` space.

algorithms/src/Kokkos_Sort.hpp
jstrzebonski: It was originally in `try_std_sort()`.

jstrzebonski:
> What about other execution spaces that can access `HostSpace`? Note in particular that #3929 introduced using `std::sort` for `CudaUVM` space.

Done.

jstrzebonski:
Retest this please.

algorithms/src/Kokkos_Sort.hpp
```diff
@@ -584,13 +568,14 @@ struct min_max_functor {
...
-    const ExecutionSpace& exec, ViewType const& view,
-    bool const always_use_kokkos_sort = false) {
-  if (!always_use_kokkos_sort) {
-    if (Impl::try_std_sort(view, exec)) return;
-  }
+template <class ExecutionSpace, class DataType, class... Properties>
+std::enable_if_t<(Kokkos::is_execution_space<ExecutionSpace>::value) &&
+                 (not SpaceAccessibility<
+                     HostSpace, typename Kokkos::View<DataType, Properties...>::
+                                    memory_space>::accessible)>
```
masterleinad, line 575: We don't use `not` anywhere else.
```suggestion
std::enable_if_t<(Kokkos::is_execution_space<ExecutionSpace>::value) &&
                 SpaceAccessibility<
                     HostSpace, typename Kokkos::View<DataType, Properties...>::
                                    memory_space>::accessible>
```

algorithms/src/Kokkos_Sort.hpp
masterleinad: If we pass an execution space, we really shouldn't have this fence here.

algorithms/src/Kokkos_Sort.hpp
masterleinad: Realistically, this branch is only executed now for GPU memory spaces and for those we want to call other libraries anyway (if they can be found).

algorithms/src/Kokkos_Sort.hpp
```diff
@@ -628,10 +613,23 @@ std::enable_if_t<Kokkos::is_execution_space<ExecutionSpace>::value> sort(
...
+  auto last  = Experimental::end(view);
+  std::sort(first, last);
+}
+
 template <class ViewType>
-void sort(ViewType const& view, bool const always_use_kokkos_sort = false) {
+void sort(ViewType const& view) {
   typename ViewType::execution_space exec;
-  sort(exec, view, always_use_kokkos_sort);
+  sort(exec, view);
```
masterleinad, line 632: I think this should have a global `fence` before calling the overload taking an execution space but that might be debatable.

masterleinad (commented):
I don't feel qualified to assess if it's OK to remove the option to force using `BinSort` but otherwise this pull request looks mostly good to me.

algorithms/src/Kokkos_Sort.hpp
fnrizzi: why should not fence the space instance here?

algorithms/src/Kokkos_Sort.hpp
fnrizzi: why?

algorithms/src/Kokkos_Sort.hpp
fnrizzi: oh wait this is before calling sort, so I agree

algorithms/src/Kokkos_Sort.hpp
```diff
@@ -628,10 +613,23 @@ std::enable_if_t<Kokkos::is_execution_space<ExecutionSpace>::value> sort(
...
+std::enable_if_t<(Kokkos::is_execution_space<ExecutionSpace>::value) &&
+                 (SpaceAccessibility<
+                     HostSpace, typename Kokkos::View<DataType, Properties...>::
+                                    memory_space>::accessible)>
+sort(const ExecutionSpace& space,
+     const Kokkos::View<DataType, Properties...>& view) {
+  space.fence("Kokkos::sort: Fence before sorting on the host");
+  auto first = Experimental::begin(view);
+  auto last  = Experimental::end(view);
+  std::sort(first, last);
```
fnrizzi, line 626: why are we not using a std policy for this?

algorithms/src/Kokkos_Sort.hpp
masterleinad: Generally, we expect functions without execution space argument to apply all possibly necessary fences and the input for `sort` might depend on the completion of another kernel (for which we don't know the execution space instance).

fnrizzi: but this one here accepts an execution space instance

masterleinad: No, this function is `void sort(ViewType const& view)` (without execution space instance) and we create an execution space instance to call the overload that takes an execution space instance (which should not have any fences).

algorithms/src/Kokkos_Sort.hpp
masterleinad: Also, it's the overload taking an execution space instance.

algorithms/src/Kokkos_Sort.hpp
fnrizzi: ah you are right, sorry I misread

masterleinad:
> I don't feel qualified to assess if it's OK to remove the option to force using BinSort but otherwise this pull request looks mostly good to me.

All that is to say that I think we should just deprecate the overload taking an `always_use_kokkos_sort` argument first before removing it.

algorithms/src/Kokkos_Sort.hpp
jstrzebonski: Removed.

algorithms/src/Kokkos_Sort.hpp
jstrzebonski: Added.

algorithms/src/Kokkos_Sort.hpp
jstrzebonski: `not` changed into `!`.

crtrott:
Ah shit can we get this rebased on develop?

jstrzebonski:
> Ah shit can we get this rebased on develop?

@crtrott 
Rebased.

algorithms/src/Kokkos_Sort.hpp
```diff
@@ -566,9 +568,14 @@ struct min_max_functor {
...
 }  // namespace Impl
 
-template <class ExecutionSpace, class ViewType>
-std::enable_if_t<Kokkos::is_execution_space<ExecutionSpace>::value> sort(
-    const ExecutionSpace& exec, ViewType const& view) {
+template <class ExecutionSpace, class DataType, class... Properties>
+std::enable_if_t<(Kokkos::is_execution_space<ExecutionSpace>::value) &&
+                 (!SpaceAccessibility<
+                     HostSpace, typename Kokkos::View<DataType, Properties...>::
+                                    memory_space>::accessible)>
```
dalg24, line 575: Why do you put parentheses around boolean values?

algorithms/src/Kokkos_Sort.hpp
```diff
@@ -566,9 +568,14 @@ struct min_max_functor {
...
+sort(const ExecutionSpace& exec,
+     const Kokkos::View<DataType, Properties...>& view) {
+  using ViewType = typename Kokkos::View<DataType, Properties...>;
```
dalg24, line 578:
```suggestion
  using ViewType = Kokkos::View<DataType, Properties...>;
```

algorithms/src/Kokkos_Sort.hpp
```diff
@@ -606,8 +613,20 @@ std::enable_if_t<Kokkos::is_execution_space<ExecutionSpace>::value> sort(
   bin_sort.sort(exec, view);
 }
 
+template <class ExecutionSpace, class DataType, class... Properties>
+std::enable_if_t<(Kokkos::is_execution_space<ExecutionSpace>::value) &&
+                 (SpaceAccessibility<
+                     HostSpace, typename Kokkos::View<DataType, Properties...>::
+                                    memory_space>::accessible)>
+sort(const ExecutionSpace&, const Kokkos::View<DataType, Properties...>& view) {
```
dalg24, line 621: I am not sure about that.  Why are you forwarded to `std::sort` for other execution spaces than `Serial`.
`Serial` is the only case where this is a no-brainer.

algorithms/src/Kokkos_Sort.hpp
jstrzebonski: I can easily see where one condition ends. Can be removed though if that brokes any Kokkos's coding style.

algorithms/src/Kokkos_Sort.hpp

crtrott (approved):

dalg24 (approved):

algorithms/src/Kokkos_Sort.hpp
dalg24: I don't like it but we don't have a rule for this

masterleinad (approved):

algorithms/src/Kokkos_Sort.hpp
fnrizzi: @jstrzebonski @dalg24 last comment before approving, but why not using a policy here?

dalg24: Because say we are using the `Serial` backend, we are not allowed to just go grab more resources for sorting.
Say the execution space is any of our host parallel backend, we would have to honor what is available in the execution space instance.

fnrizzi: makes sense

PR #5374, opened 2022-08-22
Don't use 'inline' with KOKKOS_DEFAULTED_FUNCTION or KOKKOS_INLINE_FUNCTION_DELETED

masterleinad:
There doesn't seem to be a good reason to use `inline` with `KOKKOS_DEFAULTED_FUNCTION` or `KOKKOS_INLINE_FUNCTION_DELETED` since both of these are to be used in class scope anyway.


core/src/setup/Kokkos_Setup_HIP.hpp
```diff
@@ -58,8 +58,8 @@
 
 #define KOKKOS_IMPL_FORCEINLINE_FUNCTION __device__ __host__ __forceinline__
 #define KOKKOS_IMPL_INLINE_FUNCTION __device__ __host__ inline
-#define KOKKOS_DEFAULTED_FUNCTION __device__ __host__ inline
-#define KOKKOS_INLINE_FUNCTION_DELETED __device__ __host__ inline
+#define KOKKOS_DEFAULTED_FUNCTION __device__ __host__
+#define KOKKOS_INLINE_FUNCTION_DELETED __device__ __host__
```
dalg24, line 62: Do we need the `__host__ __device__` annotation in the first place?

core/src/setup/Kokkos_Setup_Cuda.hpp
```diff
@@ -110,10 +110,10 @@
 #define KOKKOS_IMPL_HOST_FUNCTION __host__
 #define KOKKOS_IMPL_DEVICE_FUNCTION __device__
 #if defined(KOKKOS_COMPILER_NVCC)
-#define KOKKOS_INLINE_FUNCTION_DELETED inline
+#define KOKKOS_INLINE_FUNCTION_DELETED
 #else
-#define KOKKOS_INLINE_FUNCTION_DELETED __device__ __host__ inline
+#define KOKKOS_INLINE_FUNCTION_DELETED __device__ __host__
```
dalg24, line 115: Are these annotation needed?

core/src/Kokkos_Macros.hpp
```diff
@@ -374,11 +374,11 @@
 #endif
 
 #if !defined(KOKKOS_INLINE_FUNCTION_DELETED)
-#define KOKKOS_INLINE_FUNCTION_DELETED inline
+#define KOKKOS_INLINE_FUNCTION_DELETED
 #endif
 
 #if !defined(KOKKOS_DEFAULTED_FUNCTION)
-#define KOKKOS_DEFAULTED_FUNCTION inline
+#define KOKKOS_DEFAULTED_FUNCTION
```
fnrizzi, line 381: I agree with removing the inline, so that also leads me to ask:  if we do that, it becomes an empty macro, so what is the point of having it at all? just in case we want to redefine to something else later?

fnrizzi:
@masterleinad the reason for this PR is that since both of these are to be used to annotate class member functions, these are lined by default by the compiler anyway so there is no need to have them?

core/src/Kokkos_Macros.hpp
masterleinad: Application codes use at least `KOKKOS_DEFAULTED_FUNCTION` so I don't think we should remove it right away but we could think about deprecating it in some way (or don't in case there will be a need to do something here).

masterleinad:
> @masterleinad IIUC, the reason for this PR is that since those macros are used to annotate class member functions, these are inlined by default by the compiler anyway so there is no need to have them?

The motivation is that an in-class function is implicitly an inline function.

core/src/Kokkos_Macros.hpp
fnrizzi: but do application codes use that macro also for free functions or just for class member fncs?

fnrizzi:
> > @masterleinad IIUC, the reason for this PR is that since those macros are used to annotate class member functions, these are inlined by default by the compiler anyway so there is no need to have them?
> 
> The motivation is that an in-class function is implicitly an inline function.

yes that is what I thought

core/src/Kokkos_Macros.hpp
masterleinad: Free functions cannot be `default` anyway, right? I'm not quite sure how much `delete` is used, though.

fnrizzi: yes, agree! 
In any case my main point was that if we make a macro empty, I think we should deprecate it at some point, since I don't quite see the purpose of an empty macro

core/src/setup/Kokkos_Setup_HIP.hpp
masterleinad: Let's see what 4093f63bc shows.

masterleinad:
Retest this please.

crtrott:
Note this is fairly dangerous, the breakage for marking defaulted functions wrong as far as I remember was almost always downstream and not easy to reproduce. Largely it came down to sometimes the function not being generated for the device, depending exactly on the context.

crtrott (changes requested):
I don't feel comfortable with this change. See my other comment.

masterleinad:
> Note this is fairly dangerous, the breakage for marking defaulted functions wrong as far as I remember was almost always downstream and not easy to reproduce. Largely it came down to sometimes the function not being generated for the device, depending exactly on the context.

This pull request originally only removed `inline` from `KOKKOS_DEFUALTED_FUNCTION` and `KOKKOS_INLINE_FUNCTION_DELETED` but didn't change the `__host__ __device__` annotations. For defaulted functions, we only change `HIP` with the last commit.
Anyway, I'm happy to remove the last commit.

dalg24:
> > Note this is fairly dangerous, the breakage for marking defaulted functions wrong as far as I remember was almost always downstream and not easy to reproduce. Largely it came down to sometimes the function not being generated for the device, depending exactly on the context.
> 
> This pull request originally only removed `inline` from `KOKKOS_DEFUALTED_FUNCTION` and `KOKKOS_INLINE_FUNCTION_DELETED` but didn't change the `__host__ __device__` annotations. For defaulted functions, we only change `HIP` with the last commit. Anyway, I'm happy to remove the last commit.

Did the CI pass before you force pushed?  I still believe the `__host__ __device__` annotations are not needed with the new minimum compiler requirements, neither for defaulted nor for deleted special members.

masterleinad:
> Did the CI pass before you force pushed? I still believe the `__host__ __device__` annotations are not needed with the new minimum compiler requirements, neither for defaulted nor for deleted special members.

Yes the CI was passing (for the relevant builds), see https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/Kokkos/detail/Kokkos/10279/pipeline. I also think that we don't need the annotations but I can also see a point in erring on the side of caution.

JBludau (approved):
I am with Christian here ... As far as I see it we do not gain much from removing the `__host__ __device__`annotations but are on the safe side for downstream stuff

crtrott:
I overlooked that for CUDA the host/device was only on the delete thing. That shouldn't be an actual issue. I was concerned about the defaulted. Which leaves only HIP as an actual proposed change earlier. If we can get word from AMD which says you don't need it, I am ok removing this stuff.

masterleinad:
Waiting on a response from @arghdos.

arghdos:
Tested this on ROCm 5.2, and it appears to do what we want:

```
#include <hip/hip_runtime.h>



class foo {
public:
  void bar() = delete;
  foo() = default;
};


__global__ void foobar(foo f) { f.bar(); }
__global__ void barfoo() { foo f; }


__host__ void foobarh(foo f) { f.bar(); }
__host__ void hbarfoo() { foo f; }


int main() {
  foo f;
  foobar<<<1,1>>>(f);
  barfoo<<<1,1>>>();
  foobarh(f);
  hbarfoo();
}
```

Errors on calling delete'd functions happen regardless of H/D.
Similarly, the default constructor appears seems to be valid regardless of the callsite.

I was told we should be following the same implementation of these as cuda-clang (which appears to try to implement the same as NVCC), so from my perspective I think it should be OK

dalg24:
@masterleinad please update the PR accordingly

dalg24:
Clang+CUDA failure was `No space left on device`

dalg24 (approved):

masterleinad:
@crtrott ping

JBludau (approved):

PR #5376, opened 2022-08-24
OpenACC: implement `parallel_for` with `MDRangePolicy`

dalg24:
Supersedes #5335 
Rebased on develop and pulled out `parallel_reduce` implementation.
Note that the tiling code path is currently disabled


seyonglee (approved):
It looks good to me.

core/src/OpenACC/Kokkos_OpenACC_ParallelFor_MDRange.hpp
```diff
@@ -697,7 +697,7 @@ class Kokkos::Impl::ParallelFor<Functor, Kokkos::MDRangePolicy<Traits...>,
       }
     }
     int const async_arg = m_policy.space().acc_async_queue();
-#if 0  // FIXME_OPENACC
+#if 0  // FIXME_OPENACC: OpenACC requires tile size to be constant.
```
dalg24, line 700: You mean requires them too be known at compile-time?
If so that is a dead end and we can remove the code path altogether

seyonglee: Specwise, yes, it should be compile-time constant. However, NVHPC does not complain this issue, even though other errors occur in the loop with non-constant tile size.

pedrovalerolara (approved):

JBludau (approved):
looks good

dalg24:
Merging with the understanding that we need to revisit the tiling code path

dalg24:
The SYCL build timed out.  Safe to ignore.

PR #5383, opened 2022-08-24
Move HIP out of Experimental

Rombur:
Move the execution and memory spaces associated with HIP outside of experimental. I have added an alias in `core/src/Kokkos_HIP.hpp` so that current codes still work.  The old code is not deprecated, I wasn't sure what was the plan.


core/unit_test/tools/TestEventCorrectness.hpp
```diff
@@ -55,9 +55,9 @@ class Serial;
 class OpenMP;
 class Cuda;
 class Threads;
+class HIP;
 namespace Experimental {
 class SYCL;
-class HIP;
```
Rombur, line 60: This code is not invalid. The compiler compiles that `Experimental::HIP` is already defined. I don't think that many users will hit that problem and the fix in user code is pretty simple.

dalg24: not -> now

In other words, this PR would break code that had forward declarations of `HIP` or `HIPSpace` in `Kokkos::Experimental` namespace.

dalg24 (approved):

dalg24:
Retest this please

JBludau (approved):
looks good.
The only two things I found that still use `Experimental::HIP` are in:
./core/unit_test/TestAbort.hpp:123:  `if (std::is_same<ExecutionSpace, Kokkos::Experimental::HIP>::value) {`
./core/unit_test/TestViewMemoryAccessViolation.hpp:202:  `if (std::is_same<ExecutionSpace, Kokkos::Experimental::HIP>::value) {`

containers/src/Kokkos_ScatterView.hpp
```diff
@@ -1248,11 +1248,11 @@ class ScatterView<DataType, Kokkos::LayoutLeft, DeviceType, Op,
...
-        view_alloc(WithoutInitializing,
-                   std::string("duplicated_") + original_view.label(),
-                   exec_space),
-        arg_N[0], arg_N[1], arg_N[2], arg_N[3], arg_N[4], arg_N[5], arg_N[6],
-        arg_N[7]);
+                              view_alloc(WithoutInitializing,
+                                         std::string("duplicated_") + original_view.label(),
+                                         exec_space),
+                              arg_N[0], arg_N[1], arg_N[2], arg_N[3], arg_N[4], arg_N[5], arg_N[6],
+                              arg_N[7]);
```
crtrott, line 1255: so weird ....

crtrott (changes requested):
Yeah lets get those two guys still fixed and then we can merge this.

containers/src/Kokkos_ScatterView.hpp
Rombur: I used the wrong version of clang-format. Will fix it.

crtrott (approved):

Issue #5382, opened 2022-08-24
Deprecate sort(View, bool always_call_kokkos_sort)


PhilMiller:
Steer users to a new overload that doesn't take the trailing `bool` argument

PR #5384, opened 2022-08-24
[3.7] #5382: Deprecate overloads of Kokkos::sort() taking parameter 'bool always_use_kokkos_sort'

PhilMiller:
A richer sorting interface will be implemented later, probably involving a handle class of some sort. This is just clearing out an obsolete design.

### DIFF
--- a/algorithms/src/Kokkos_Sort.hpp
+++ b/algorithms/src/Kokkos_Sort.hpp
@@ -586,11 +586,7 @@ struct min_max_functor {
 
 template <class ExecutionSpace, class ViewType>
 std::enable_if_t<Kokkos::is_execution_space<ExecutionSpace>::value> sort(
-    const ExecutionSpace& exec, ViewType const& view,
-    bool const always_use_kokkos_sort = false) {
-  if (!always_use_kokkos_sort) {
-    if (Impl::try_std_sort(view, exec)) return;
-  }
+    const ExecutionSpace& exec, ViewType const& view) {
   using CompType = BinOp1D<ViewType>;
 
   Kokkos::MinMaxScalar<typename ViewType::non_const_value_type> result;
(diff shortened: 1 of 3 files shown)


algorithms/src/Kokkos_Sort.hpp
```diff
@@ -628,7 +624,31 @@ std::enable_if_t<Kokkos::is_execution_space<ExecutionSpace>::value> sort(
...
+}
+
+template <class ViewType>
+void sort(ViewType const& view) {
+  typename ViewType::execution_space exec;
+  // FIXME: Do this and the following deprecated overload need to call
+  // Kokkos::fence() first?
+  sort(exec, view);
+  exec.fence("Kokkos::Sort: fence after sorting");
+}
```
masterleinad, line 647: I think it's OK to fix/change that later in `develop` (where I already had the same comment).

masterleinad (commented):
Guard the deprecated functions with `KOKKOS_ENABLE_DEPRECATED_CODE_3` so we can (discuss to) remove them in `develop`.

PhilMiller:
> Guard the deprecated functions with `KOKKOS_ENABLE_DEPRECATED_CODE_3` so we can (discuss to) remove them in `develop`.

Ah, right, thanks

algorithms/src/Kokkos_Sort.hpp
```diff
@@ -628,7 +624,31 @@ std::enable_if_t<Kokkos::is_execution_space<ExecutionSpace>::value> sort(
...
+  } else {
+    sort(exec, view);
+  }
+}
+
+template <class ViewType>
+void sort(ViewType const& view) {
+  typename ViewType::execution_space exec;
+  // FIXME: Do this and the following deprecated overload need to call
+  // Kokkos::fence() first?
```
dalg24, line 644: I don't think so

dalg24: Phil pointed out to me that the data in the few might be filled by some other execution space type

PhilMiller: https://github.com/kokkos/kokkos/issues/5385 To consider/address this

crtrott (approved):

PhilMiller:
Oh, I should add a note to the changelog about this late deprecation

masterleinad (approved):

algorithms/src/Kokkos_Sort.hpp
```diff
@@ -628,12 +624,38 @@ std::enable_if_t<Kokkos::is_execution_space<ExecutionSpace>::value> sort(
...
+  typename ViewType::execution_space exec;
+  sort(exec, view);
+  exec.fence("Kokkos::Sort: fence after sorting");
+}
+
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_3
+template <class ViewType>
+KOKKOS_DEPRECATED_WITH_COMMENT(
+    "Use the overload not taking bool always_use_kokkos_sort")
 void sort(ViewType const& view, bool const always_use_kokkos_sort = false) {
```
masterleinad, line 653:
```suggestion
void sort(ViewType const& view, bool const always_use_kokkos_sort) {
```

PhilMiller:
There's a bit more that needs deprecation-guarding in the test. Fixing now.

Issue #75, opened 2015-08-24
Require scratch_memory_space for shared memory views


hcedwar:

crtrott:
Do we actually want to do this this way? Not sure how that matches our hierarchy of scratches now.

hcedwar:
to be re-evaluated...

Issue #72, opened 2015-08-19
Add Aggressive Vectorization Compilation mode


crtrott:
In principal Intels interpretation of #pragma ivdep can always be added to Kokkos code, since our parallel_for kernels have to be outer loop vectorizable, and intels #pragma ivdep only applies to the immediately following block, and not to nested blocks. That said: this is the current compiler, and the behavior is ill defined, so we don't want to add that by default. This compiler option would add the #pragma ivdep and thus help considerably with vectorization.

nmhamster:
We discussed that I'm nervous about potential errors this may bring to code which will be hard to find and fix, particularly in the initial ports of code to use Kokkos. Its something for us to consider but we should prioritize user experience right now IMHO.

crtrott:
Yeah I know ;-). I only want to open an issue to track the issue, and give a space to leave comments. I got some feedback from the Fluid Dynamics folks who liked the idea of getting that mode (which is not on by default). Btw. I would be interested in an example where it actually fails, so far I couldn't contrive one (which does not mean that I believe that there is none ...).

nmhamster:
Are you asking for an example which is runnable on the GPU under "Kokkos rules" or code which you are actually going to get in the OpenMP-to-Kokkos transition phase? These two things are not born equal.

crtrott:
I was more thinking about in the OpenMP-to-Kokkos transition phase, but if you have an example which actually runs correct with the Cuda backend and fails with OpenMP backend while marking loops with ivdep that would be even more interesting.

crtrott:
After discussions today at the Kokkos meeting I pushed this change in. To enable it either define KOKKOS_OPT_RANGE_AGGRESSIVE_VECTORIZATION to 1 or when building with the Makefile system use KOKKOS_OPTIONS=aggressive_vectorization as an option. There is no CMake option right now. It also only affects the OpenMP backend with the Intel compiler currently.

Slaedr:
@crtrott What are your thoughts on similarly adding an `omp simd` option so that other compilers can also vectorize chunks? I tried a very simple loop body which clang 7 would not vectorize, unless I added the `omp simd` pragma. See [this](https://github.com/Slaedr/kokkos/blob/2de367ac943d1693b043573ab3c92c8dca0656cf/core/src/OpenMP/Kokkos_OpenMP_Parallel.hpp#L90).

Slaedr:
[...]

PR #5386, opened 2022-08-24
Deprecate Subview type from public interface

JBludau:
During documentation @nmm0 and me wrote documentation for this public using declaration.
The doc started a discussion about whether we wanted to have this type in our public interface, as `decltype(Kokkos::subview(...))` would serve the same purpose.

As the discussion stalled, I propose to deprecate it from the public interface. This would also finalize the discussion on [#75](https://github.com/kokkos/kokkos-core-wiki/pull/75) and [#72](https://github.com/kokkos/kokkos-core-wiki/pull/72)

### DIFF
--- a/core/src/Kokkos_View.hpp
+++ b/core/src/Kokkos_View.hpp
@@ -1770,12 +1770,14 @@ void apply_to_view_of_static_rank(Function&& f, View<Args...> a) {
 //----------------------------------------------------------------------------
 //----------------------------------------------------------------------------
 
+#if defined KOKKOS_ENABLE_DEPRECATED_CODE_4
 template <class V, class... Args>
-using Subview =
+using Subview KOKKOS_DEPRECATED =
     typename Kokkos::Impl::ViewMapping<void /* deduce subview type from source
                                                view traits */
                                        ,
                                        typename V::traits, Args...>::type;
+#endif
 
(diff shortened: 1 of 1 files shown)


JBludau:
Retest this please

masterleinad (commented):
I think it's a good idea to at least provide it in an `Impl` namespace or so.

JBludau:
Jeah, I agree ... after I looked through how often it is used internally, I thought it might make sense to keep it for internal use (out of convenience) and just deprecate it from our public interface

JBludau:
I talked about this with @dalg24 and he did agree, that using `decltype(declval<...>(...))` is awkward if used for templates. e.g.:

```
template<typename T>
class bla {
  using ViewType = Kokkos::View<int ***>;
  using mySubView = decltype(Kokkos::subview(std::declval<ViewType>(),0,0,std::declval<T>()); 
  // declval<T> is necessary, as we do not know if T is default constructible 
  // and we only 'construct' it here to get type information
};
```

As the `Subview` alias template gives the required information purely based on types it is a real convenience to have it, thus we should keep it and document it. This should free the way to reconsider [#75](https://github.com/kokkos/kokkos-core-wiki/pull/75) and [#72](https://github.com/kokkos/kokkos-core-wiki/pull/72) on the wiki for merging

PR #5388, opened 2022-08-24
#5382: Delete deprecated overloads of Kokkos::sort() taking parameter 'bool always_use_kokkos_sort'

PhilMiller:
A richer sorting interface will be implemented later, probably involving a handle class of some sort. This is just clearing out an obsolete design.

### DIFF
--- a/algorithms/src/Kokkos_Sort.hpp
+++ b/algorithms/src/Kokkos_Sort.hpp
@@ -624,38 +624,12 @@ std::enable_if_t<Kokkos::is_execution_space<ExecutionSpace>::value> sort(
   bin_sort.sort(exec, view);
 }
 
-#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_3
-template <class ExecutionSpace, class ViewType>
-KOKKOS_DEPRECATED_WITH_COMMENT(
-    "Use the overload not taking bool always_use_kokkos_sort")
-std::enable_if_t<Kokkos::is_execution_space<ExecutionSpace>::value> sort(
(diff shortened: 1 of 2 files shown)


algorithms/src/Kokkos_Sort.hpp
```diff
@@ -628,12 +624,38 @@ std::enable_if_t<Kokkos::is_execution_space<ExecutionSpace>::value> sort(
   bin_sort.sort(exec, view);
 }
 
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_3
+template <class ExecutionSpace, class ViewType>
+KOKKOS_DEPRECATED_WITH_COMMENT(
```
PhilMiller, line 629: Since `develop` is now on track for 4.0, I guess I should drop the deprecated code entirely?

PhilMiller:
I added a commit to just delete the deprecated code. I can drop that commit if we decide to keep it for longer (either 4.1, or 5.0)

masterleinad:
https://github.com/kokkos/kokkos/pull/5372 pretty much does this already. I'm not sure if we need a separate pull request.

masterleinad:
Otherwise, this should also remove `Impl::try_std_sort` (like #5372 does).

PhilMiller:
Until that's ready to merge, this ensures that `develop` is consistent with what we do on the release branch with #5384

masterleinad (approved):

fnrizzi:
what is going on with the CI ?

fnrizzi (approved):

crtrott (approved):

PR #5151, opened 2022-06-27
Finalize deep_copy_space early avoiding printing to std::cerr for Cuda

masterleinad:
We are currently printing an error message to `std::cerr` when finalizing `Cuda` execution spaces as the first commit shows.

The problem is that we are finalizing `deep_copy_space` since that needs to deallocate data which calls a fence on the default execution space and we check if the execution space has already unset `m_dev` at this point. Wd didn't see this so far, since we are only printing to `std::cerr` but not throwing an exception or aborting.
The proposed fix is to finalize `deep_copy_space` before unsettling `m_dev`.

While looking at this, I also found that some of the code in `finalize` is guarded by a check for scratch space which didn't make sense to me.

Throwing in `finalize` is debatable, of course, but makes the problem obvious (in the CI).

PR #5390, opened 2022-08-25
[3.7] Force creation of the cuda_deep_space instance during init

crtrott:
Can't do it inside of the Cuda initialize since that ends up being
recursive.
We may want to find a better solution however.

The test I have passes if run without valgrind, so I didn't add it here
yet.

```c++
  std::atexit(Kokkos::finalize);
  (void) Kokkos::Impl::cuda_get_deep_copy_space(true);
}
```

Also we would need to create a separate executable for this test.

### DIFF
--- a/core/src/Cuda/Kokkos_Cuda_Instance.cpp
+++ b/core/src/Cuda/Kokkos_Cuda_Instance.cpp
@@ -861,6 +861,14 @@ int Cuda::impl_is_initialized() {
 
 void Cuda::impl_initialize(InitializationSettings const &settings) {
   Impl::CudaInternal::singleton().initialize(Impl::get_gpu(settings));
+
+  // In order to support setting an atexit hook for Kokkos::finalize
+  // We need to ensure that the Cuda deep_copy instance is not destroyed
+  // before that atexit hook is getting called.
+  // Thus we create the static instance here, so that it will be deallocated
+  // after the potential atexit call.
+  // This is neccessary since we will access that instance in Kokkos::finalize
+  (void)::Kokkos::Impl::cuda_get_deep_copy_space(true);
 }
 
 std::vector<unsigned> Cuda::detect_device_arch() {


core/src/impl/Kokkos_Core.cpp
```diff
@@ -674,6 +674,15 @@ void post_initialize_internal(const Kokkos::InitializationSettings& settings) {
...
     ::Kokkos::print_configuration(std::cout);
   }
+// In order to support setting an atexit hook for Kokkos::finalize
+// We need to ensure that the Cuda deep_copy instance is not destroyed
+// before that atexit hook is getting called.
+// Thus we create the static instance here, so that it will be deallocated
+// after the potential atexit call.
+// This is neccessary since we will access that instance in Kokkos::finalize
+#ifdef KOKKOS_ENABLE_CUDA
+  (void)::Kokkos::Impl::cuda_get_deep_copy_space(true);
```
dalg24, line 684: Why can't this be in `CUDA::impl_initialize()`?

crtrott: Ah right we have that thing. It can't be called inside the CudaInstance::initialize because that turns out to be recursive, but we can call it in Cuda::impl_initialize, I overlooked that call.

masterleinad (dismissed):
Add a test.

crtrott:
Note the test I have would not trigger in normal testing, you need to run with valgrind to see it fail.

crtrott:
Also we need a new executable for that, and I don't want to add a new executable and a dependency on valgrind in the release candidate.

PhilMiller:
> Note the test I have would not trigger in normal testing, you need to run with valgrind to see it fail.

Does the test also trigger AddressSanitizer? We could modify a CI build to always apply that. I have no idea how much else might fail, though.

PhilMiller (approved):

PhilMiller:
To be clear about my comment, I don't think we should try to do that extension of our testing for this to go into the 3.7 release. Christian's manual testing seems acceptable to me in this instance.

core/src/Cuda/Kokkos_Cuda_Instance.cpp
```diff
@@ -861,6 +861,14 @@ int Cuda::impl_is_initialized() {
 
 void Cuda::impl_initialize(InitializationSettings const &settings) {
   Impl::CudaInternal::singleton().initialize(Impl::get_gpu(settings));
+
+  // In order to support setting an atexit hook for Kokkos::finalize
```
dalg24, line 865:
```suggestion
  // In order to support registering `Kokkos::finalize` to be called at program termination (via std::atexit)
```
(not formatted feel free to ignore)

dalg24 (approved):

masterleinad:
Thinking some more about this, I'm pretty convinced that it's not needed at all. Since we are accessing the default execution space instance in `finalize` (also see https://github.com/kokkos/kokkos/pull/5151), we need to make sure that all other execution space instances have been finalized before finalizing the singleton. From our side, the only problem with that is the static deep copy space instance which we take care of in https://github.com/kokkos/kokkos/blob/7522044c4bffe712b01eb9a7875ea3f44b222cf5/core/src/Cuda/Kokkos_Cuda_Instance.cpp#L724-L728. The problem we were trying to fix in this pull request would only appear if we were to use the two-argument deep copy instead of the three-argument deep copy in https://github.com/kokkos/kokkos/blob/acb34347ac67dc55c5255c1f7dac39af47e11b5f/core/src/impl/Kokkos_SharedAlloc_timpl.hpp#L264-L271 which we did before https://github.com/kokkos/kokkos/commit/43ada2b312c9d1dddcdd8f5c515e6d44aa3f380f (https://github.com/kokkos/kokkos/pull/4478). With this pull request I'm seeing
```
Accessing copy stream
creating copy stream
after initialize, before atexit
after atexit
Accessing copy stream
finalizing singleton
Accessing copy stream
finalizing non-singleton
destroying non-singleton
destroying singleton
```
and without it 
```
after initialize, before atexit
after atexit
Accessing copy stream
creating copy stream
finalizing non-singleton
destroying non-singleton
finalizing singleton
Accessing copy stream
destroying singleton
```
for
```
#include <Kokkos_Core.hpp>

  std::cout << "after initialize, before atexit" << std::endl;
  std::atexit(Kokkos::finalize);
  std::cout << "after atexit" << std::endl;
  (void) Kokkos::Impl::cuda_get_deep_copy_space(true);
}
```
Note that in both cases the copy stream (non-singleton) is finalized before the singleton (as desired).

PR #5391, opened 2022-08-25
Don't rely on synchronization behavior of default stream in CUDA and HIP

masterleinad:
https://docs.nvidia.com/cuda/cuda-runtime-api/stream-sync-behavior.html:
> The legacy default stream is an implicit stream which synchronizes with all other streams in the same CUcontext except for non-blocking streams, described below. (For applications using the runtime APIs only, there will be one context per device.) When an action is taken in the legacy stream such as a kernel launch or cudaStreamWaitEvent(), the legacy stream first waits on all blocking streams, the action is queued in the legacy stream, and then all blocking streams wait on the legacy stream. 

and HIP behaves similarly, see https://github.com/ROCm-Developer-Tools/HIP/issues/129#issuecomment-318667584. 
As discussed on Slack, the default execution space instance shouldn't have any special synchronization behavior with respect to other execution space instances. Hence, this pull request also creates a stream for the singleton that is used for the default execution space instance.

### DIFF
--- a/core/src/Cuda/Kokkos_Cuda_Instance.cpp
+++ b/core/src/Cuda/Kokkos_Cuda_Instance.cpp
@@ -551,11 +551,11 @@ Kokkos::Cuda::initialize WARNING: Cuda is allocating into UVMSpace by default
   cudaDeviceSetCacheConfig(cudaFuncCachePreferShared);
 
   // Init the array for used for arbitrarily sized atomics
-  if (stream == nullptr) Impl::initialize_host_cuda_lock_arrays();
+  if (this == &singleton()) Impl::initialize_host_cuda_lock_arrays();
 
   // Allocate a staging buffer for constant mem in pinned host memory
   // and an event to avoid overwriting driver for previous kernel launches
-  if (stream == nullptr) {
+  if (this == &singleton()) {
     KOKKOS_IMPL_CUDA_SAFE_CALL(
         cudaMallocHost(reinterpret_cast<void **>(&constantMemHostStaging),
                        CudaTraits::ConstantMemoryUsage));
@@ -835,7 +835,10 @@ int Cuda::impl_is_initialized() {
 }
 
 void Cuda::impl_initialize(InitializationSettings const &settings) {
-  Impl::CudaInternal::singleton().initialize(Impl::get_gpu(settings));
+  cudaStream_t singleton_stream;
+  KOKKOS_IMPL_CUDA_SAFE_CALL(cudaStreamCreate(&singleton_stream));
+  Impl::CudaInternal::singleton().initialize(Impl::get_gpu(settings),
+                                             singleton_stream, /*manage*/ true);
 }
 
 std::vector<unsigned> Cuda::detect_device_arch() {
--- a/core/unit_test/cuda/TestCuda_Spaces.cpp
+++ b/core/unit_test/cuda/TestCuda_Spaces.cpp
@@ -266,9 +266,9 @@ TEST(cuda, uvm) {
 
     *uvm_ptr = 42;
 
-    Kokkos::Cuda().fence();
+    Kokkos::fence();
     test_cuda_spaces_int_value<<<1, 1>>>(uvm_ptr);
-    Kokkos::Cuda().fence();
+    Kokkos::fence();
 
     EXPECT_EQ(*uvm_ptr, int(2 * 42));
 
--- a/core/src/Cuda/Kokkos_Cuda_Instance.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_Instance.hpp
@@ -142,8 +142,7 @@ class CudaInternal {
     return nullptr != m_scratchSpace && nullptr != m_scratchFlags;
   }
 
-  void initialize(int cuda_device_id, cudaStream_t stream = nullptr,
-                  bool manage_stream = false);
+  void initialize(int cuda_device_id, cudaStream_t stream, bool manage_stream);
   void finalize();
 
   void print_configuration(std::ostream&) const;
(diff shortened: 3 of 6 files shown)


crtrott (approved):

core/src/HIP/Kokkos_HIP_Space.cpp
```diff
@@ -500,7 +500,10 @@ int HIP::impl_is_initialized() {
 }
 
 void HIP::impl_initialize(InitializationSettings const& settings) {
-  Impl::HIPInternal::singleton().initialize(::Kokkos::Impl::get_gpu(settings));
+  hipStream_t singleton_stream;
+  hipStreamCreate(&singleton_stream);
```
JBludau, line 504: Hmm maybe a name like `base_stream`, `general_stream` or `default_stream` is more appropriate?

masterleinad: IMHO, any of these suggestions could imply that this should be the default stream which is precisely what we don't want. Also, we are initializing the singleton here.

PhilMiller:
It would be helpful to document or reference the particular behavior in the PR description. I believe this relates to  kernel launches on CUDA/HIP's default stream implicitly synchronizing more broadly than launches on any other stream that gets created explicitly?

PhilMiller:
Also, this will pose a merge conflict with #5390 if/when that gets cherry-picked to `develop`

masterleinad:
> It would be helpful to document or reference the particular behavior in the PR description. I believe this relates to kernel launches on CUDA/HIP's default stream implicitly synchronizing more broadly than launches on any other stream that gets created explicitly?

This was still just a draft. :slightly_smiling_face:

core/src/Cuda/Kokkos_Cuda_Instance.cpp
```diff
@@ -836,7 +836,10 @@ int Cuda::impl_is_initialized() {
 }
 
 void Cuda::impl_initialize(InitializationSettings const &settings) {
-  Impl::CudaInternal::singleton().initialize(Impl::get_gpu(settings));
+  cudaStream_t singleton_stream;
+  cudaStreamCreate(&singleton_stream);
```
dalg24, line 840:
```suggestion
  KOKKOS_IMPL_CUDA_SAFE_CALL(cudaStreamCreate(&singleton_stream));
```

core/src/HIP/Kokkos_HIP_Space.cpp
dalg24, line 504:
```suggestion
  KOKKOS_IMPL_HIP_SAFE_CALL(hipStreamCreate(&singleton_stream));
```

dalg24 (dismissed):
For the record at the moment none of the CUDA nor the HIP builds passes

core/src/HIP/Kokkos_HIP_Space.cpp
JBludau: Makes sense.

masterleinad:
Retest this please.

masterleinad:
Retest this please.

core/unit_test/cuda/TestCuda_Spaces.cpp
```diff
@@ -266,9 +266,9 @@ TEST(cuda, uvm) {
 
     *uvm_ptr = 42;
 
-    Kokkos::Cuda().fence();
+    Kokkos::fence();
     test_cuda_spaces_int_value<<<1, 1>>>(uvm_ptr);
-    Kokkos::Cuda().fence();
```
crtrott, line 271: Ah ok this makes sense.

crtrott (approved):

crtrott:
`hip_DeathTest.abort_from_device` is timeing out.

masterleinad:
> `hip_DeathTest.abort_from_device` is timeing out.

Yeah, I couldn't reproduce on `crusher` or `spock` easily and need to dig a little deeper.

core/unit_test/TestAbort.hpp
```diff
@@ -59,8 +59,10 @@ template <class ExecutionSpace>
 struct TestAbortPrintingToStdout {
   TestAbortPrintingToStdout() {
     ::testing::internal::CaptureStdout();
-    Kokkos::parallel_for(Kokkos::RangePolicy<ExecutionSpace>(0, 1), *this);
-    Kokkos::fence();
+    ExecutionSpace exec;
+    Kokkos::parallel_for(Kokkos::RangePolicy<ExecutionSpace>(exec, 0, 1),
+                         *this);
+    exec.fence();
```
masterleinad, line 65: I'm still not quite sure why the `HIP` Ci was timing out without changing this and I'm not quite sure how much we care. 
I couldn't reproduce on `crusher` or `spock`.

masterleinad: Also, @JBludau can only reproduce with ROCm < 5.

crtrott: So you are saying this is not hanging with 5.2?

crtrott (approved):

crtrott (changes requested):
Ok lets revert that abort test change again, and move testing to 5.2

masterleinad:
As expected this works after updating CI to `ROCm 5.2`.

crtrott (approved):

masterleinad:
Waiting for #5416.

masterleinad:
Rebased after #5416 has been merged.

masterleinad:
Rebased after #5410 has been merged.

masterleinad:
Only the `OpenMPTarget` CI is failing with
```
[ RUN      ] openmptarget.unique_token_global
4: /var/jenkins/workspace/Kokkos/core/unit_test/TestUniqueToken.hpp:151: Failure
4:   sum
4:     Which is: 9999990
4:   int64_t(N) * R
4:     Which is: 10000000
4: [  FAILED  ] openmptarget.unique_token_global (267 ms)
```
Since only CUDA and HIP were changed, this is clearly unrelated.

Rombur (approved):

Issue #5397, opened 2022-08-26
`std::hash`, `operator<` and `operator==` for execution spaces


cwpearson:
Currently, if you try to do an `std::map`, `std::set`, or `std::unordered_map` of Kokkos execution spaces you'll get a variety of errors about missing operators:

* std::unordered_set
  * `operator==`
  * `std::hash`
* std::map
  * `operator<`
  * `operator==`
* std::set
  * `operator<`

Someone may want to do this to have a straightforward way to associate other program data with a particular execution space.

It's possible for the user to define their own comparison and hashing operators, but to my reading Kokkos does not define what makes a space equivalent to another in [Execution Space Concept](https://github.com/kokkos/kokkos/wiki/ExecutionSpaceConcept). A user might be tempted to use `Space::m_instance_id` for example (which might even work).

I propose Kokkos defines `operator==`, an std::hash specialization, and `operator<` for its execution spaces. This will make them behave more like value types when used in a straightforward manner with `std` containers. It will also provide a blueprint for anyone who needs something more esoteric and is willing to read Kokkos source code.

If there is general agreement along these lines, I can draft a PR.

dalg24:
I think we should make execution spaces be regular (that is add `operator==`)
I am not convinced we need to support anything beyond that.

cwpearson:
Based on discussion with @crtrott as well, a better plan is to just implement `operator==`, which will make Spaces more well-behaved value types but not address this container question.

Perhaps some guidance for users who want to create their own Space operators for `std` containers can be provided in documentation somewhere as a separate matter.

cwpearson:
After further discussion with @crtrott and @PhilMiller we can try for `operator==` and `std::hash`, so as to avoid implying an ordering of execution spaces but we still get `std::unordered_map`

PR #5398, opened 2022-08-26
Make execution spaces regular

dalg24:
Related to #5397 
* Add `operator==` and `operator!=` for all execution spaces and check that they meet the syntactic requirements of *regular* types

Drive-by changes:
* add missing `is_device_v` helper variable
* more C++17

### DIFF
--- a/core/src/Kokkos_Cuda.hpp
+++ b/core/src/Kokkos_Cuda.hpp
@@ -245,6 +245,13 @@ class Cuda {
   uint32_t impl_instance_id() const noexcept;
 
  private:
+  friend bool operator==(Cuda const& lhs, Cuda const& rhs) {
+    return lhs.impl_internal_space_instance() ==
+           rhs.impl_internal_space_instance();
+  }
+  friend bool operator!=(Cuda const& lhs, Cuda const& rhs) {
+    return !(lhs == rhs);
+  }
   Kokkos::Impl::HostSharedPtr<Impl::CudaInternal> m_space_instance;
 };
 
--- a/core/src/Kokkos_Serial.hpp
+++ b/core/src/Kokkos_Serial.hpp
@@ -212,6 +212,13 @@ class Serial {
 #else
   Kokkos::Impl::HostSharedPtr<Impl::SerialInternal> m_space_instance;
 #endif
+  friend bool operator==(Serial const& lhs, Serial const& rhs) {
+    return lhs.impl_internal_space_instance() ==
+           rhs.impl_internal_space_instance();
+  }
+  friend bool operator!=(Serial const& lhs, Serial const& rhs) {
+    return !(lhs == rhs);
+  }
   //--------------------------------------------------------------------------
 };
 
--- a/core/src/impl/Kokkos_ExecSpaceManager.hpp
+++ b/core/src/impl/Kokkos_ExecSpaceManager.hpp
@@ -57,40 +57,40 @@ namespace {
 
 template <class T>
 using public_member_types_t = std::enable_if_t<
-    Kokkos::is_execution_space<typename T::execution_space>::value &&
-    Kokkos::is_memory_space<typename T::memory_space>::value &&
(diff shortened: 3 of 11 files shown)


core/src/impl/Kokkos_ExecSpaceManager.hpp
```diff
@@ -83,6 +84,39 @@ using fence_t = std::enable_if_t<
...
+  static_assert(is_detected_exact_v<bool, not_equal_to_t, T>);
+  return true;
+}
+
+template <class T>
+constexpr bool check_is_regular() {
+  static_assert(check_is_semiregular<T>() && check_is_equality_comparable<T>());
+  return true;
+}
+
```
PhilMiller, line 119: Would it make sense to define and `static_assert` all of this in a test or maybe something like `Kokkos_Core.cpp`, so that these declarations don't add compilation cost to every client code?

core/src/Kokkos_Serial.hpp
```diff
@@ -212,6 +212,13 @@ class Serial {
 #else
   Kokkos::Impl::HostSharedPtr<Impl::SerialInternal> m_space_instance;
 #endif
+  friend bool operator==(Serial const& lhs, Serial const& rhs) {
+    return lhs.impl_internal_space_instance() ==
+           rhs.impl_internal_space_instance();
+  }
+  friend bool operator!=(Serial const& lhs, Serial const& rhs) {
+    return !(lhs == rhs);
+  }
```
PhilMiller, line 221: What the heck are distinct `Serial` execution space instances?

core/src/impl/Kokkos_ExecSpaceManager.hpp
dalg24: This is a good question.  I was really careful when I wrote this and this header file is exclusively included in the backend source file that does the backend registration, meaning in no way visible when you include `<Kokkos_Core.hpp>` or any other public header.

dalg24:
Some interoperability considerations:  If you construct two execution spaces from the same CUDA/HIP stream, or from the same SYCL queue, these instances will compare not equal.

PhilMiller:
> Some interoperability considerations: If you construct two execution spaces from the same CUDA/HIP stream, or from the same SYCL queue, these instances will compare not equal.

Is there anything else about how those instances function that makes them not equivalent? Fencing behavior, profiling data? That seems like it's a potential distinction without a difference, that presents a trap for exactly the sort of users who may have a need for this.

core/src/Kokkos_Serial.hpp
crtrott: I could imagine a refactor where we actually associate a single thread with the Serial execution space and have a "inline" or "CallingThread" execution space.

crtrott:
If you create two execution spaces from the same CUDA stream, you still end up with distinct mutexes for dispatch, distinct scratch allocations etc.

JBludau (approved):
looks good

core/src/Kokkos_Cuda.hpp
```diff
@@ -245,6 +245,13 @@ class Cuda {
   uint32_t impl_instance_id() const noexcept;
 
  private:
+  friend bool operator==(Cuda const& lhs, Cuda const& rhs) {
+    return lhs.impl_internal_space_instance() ==
+           rhs.impl_internal_space_instance();
```
masterleinad, line 250: Why don't you use `impl_instance_id()`?

dalg24: Because the pointer to the implementation is what guarantees you get the same scratch allocations and such.
And from that definition one cannot tell how the instance id is being computed.  I would have gone for the actual pointer but I decided to use the `impl_internal_space_instance` member function since it is both available and defined a few lines above

crtrott (approved):

crtrott:
Test failure was timing related

PR #5399, opened 2022-08-26
Fix some C-style pointer casts

masterleinad:
Let's see if people like us commit to this style.


crtrott:
I personally am not a huge fan of concat nested spaces, I wouldn't block anyone from using them, but I also don't feel we need to enforce that or refactor our code to use them ...

PR #5408, opened 2022-08-30
Release candidate 3.7.00

ndellingwood:

PR #5409, opened 2022-08-30
Add static_assert(false) for join(volatile) if deprecated code is disabled

### LINKED ISSUE #5194: Deprecate join(volatile)
Following #4077, client code declaring a reduction `join()` operation taking only `volatile` parameters is both a potential performance pitfall and an anachronism of an API requirement we carried for an obscure reason.

Should we deprecate the older method signature and plan to remove support for it, to push downstream code to fix the potential performance bug and clean up the debris?

PhilMiller:
Fixes #5194

### DIFF
--- a/core/src/impl/Kokkos_FunctorAnalysis.hpp
+++ b/core/src/impl/Kokkos_FunctorAnalysis.hpp
@@ -656,7 +656,8 @@ struct FunctorAnalysis {
       : public has_volatile_join_no_tag_function<F> {
     enum : bool { value = true };
 #ifndef KOKKOS_ENABLE_DEPRECATED_CODE_3
-    static_assert(Impl::dependent_false_v<F>, "Reducer with a join() operator taking "
+    static_assert(Impl::dependent_false_v<F>,
+                  "Reducer with a join() operator taking "
                   "volatile-qualified parameters is no longer supported");
 #endif
   };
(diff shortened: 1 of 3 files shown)


core/src/impl/Kokkos_FunctorAnalysis.hpp
```diff
@@ -673,6 +678,11 @@ struct FunctorAnalysis {
                                          detected_volatile_join_tag<F>::value)>>
       : public has_volatile_join_tag_function<F> {
     enum : bool { value = true };
+#ifndef KOKKOS_ENABLE_DEPRECATED_CODE_3
```
PhilMiller, line 681: I failed to record whether we wanted this to be 3 or 4

PhilMiller: And, whether we wanted to get this into 3.7

core/src/impl/Kokkos_FunctorAnalysis.hpp
```diff
@@ -655,6 +655,11 @@ struct FunctorAnalysis {
                         detected_volatile_join_no_tag<F>::value)>>
       : public has_volatile_join_no_tag_function<F> {
     enum : bool { value = true };
+#ifndef KOKKOS_ENABLE_DEPRECATED_CODE_3
+    static_assert(Impl::dependent_false_v<F>,
```
dalg24, line 659: I would have gone for `std::is_void_v` but I don't necessarily have an issue with your helper variable template since it convey the intent.

core/src/impl/Kokkos_FunctorAnalysis.hpp
dalg24: You got it right.  I think this missed the train as #5408 is open now.  @crtrott what do you think?

masterleinad (approved):
Looks good to me. I don't think it's crucial to port it back to 3.7. At least, the deprecation warning is there anyway.

crtrott:
Yeah I think its a bit unfortunate but its not critical.

core/src/impl/Kokkos_FunctorAnalysis.hpp
e10harvey: FYI - https://github.com/kokkos/kokkos-kernels/runs/8123690491?check_suite_focus=true

PR #5411, opened 2022-08-31
Disable OpenMPTarget unit tests that cause ICEs with icpx

masterleinad:
Trying to compile those unit tests with `icpx` for `OpenMPTarget` (and Intel GPUs) causes internal compiler errors.

### DIFF
--- a/algorithms/unit_tests/CMakeLists.txt
+++ b/algorithms/unit_tests/CMakeLists.txt
@@ -137,7 +137,8 @@ foreach(Tag Threads;Serial;OpenMP;Cuda;HPX;HIP;SYCL;OpenMPTarget)
   endif()
 endforeach()
 
-# FIXME_OPENMPTARGET
+# FIXME_OPENMPTARGET These tests cause internal compiler errors as of 09/01/22
+# when compiling for Intel's Xe-HP GPUs.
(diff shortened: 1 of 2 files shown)


JBludau:
Could you comment more details of which version of icpx/GPU combinations this is an issue for? This way we would have it documented here

crtrott (approved):

masterleinad:
> Could you comment more details of which version of icpx/GPU combinations this is an issue for? This way we would have it documented here

I don't think that I can add anything meaningful. As usual, the compiler version is some compiler drop that doesn't mean anything outside the testbeds.

JBludau:
can you at least mark the disabled tests with a hint to the problem in the `FIXME` comment? Otherwise we probably will forget what the reason was for disabling it

fnrizzi (approved):

masterleinad:
> can you at least mark the disabled tests with a hint to the problem in the `FIXME` comment? Otherwise we probably will forget what the reason was for disabling it

Here you go.

JBludau (approved):
looks good

masterleinad:
Retest this please.

crtrott:
Retest this please.

Issue #4077, opened 2021-06-07
Remove need for volatile overloads for types in Kokkos reductions


fnrizzi:
- [ ] identify where reductions are implemented
- [ ] remove volatile overloads
- [ ] ensure all tests work 

@PhilMiller commented: the tests should ideally not just pass, but also exhibit no errors from ThreadSanitizer or a comparable dynamic data race analysis. This may be complicated by the fact that ThreadSanitizer doesn't directly support atomic fence operations, only actual atomic loads and stores (cf https://github.com/google/sanitizers/issues/1352 since this deficiency isn't actually documented otherwise!)

PhilMiller:
The casts to `volatile` types occur in places like `Kokkos_{Cuda,HIP}_Team.hpp` in their global reductions to obtain the final results. Conversion of the non-`volatile` `join` operators to use `atomic` operations would seemingly accomplish the same  goals, though possibly at the expense of performance for the steps of the reduction that occur in non-concurrent shared memory access.

PhilMiller:
Conversely, a lot of operations on compound types can be performed piece-wise atomically, whereever the member elements don't interact in how the operator behaves. E.g. sum, difference, point-wise min, max

masterleinad:
> The casts to `volatile` types occur in places like `Kokkos_{Cuda,HIP}_Team.hpp` in their global reductions to obtain the final results. Conversion of the non-`volatile` `join` operators to use `atomic` operations would seemingly accomplish the same goals, though possibly at the expense of performance for the steps of the reduction that occur in non-concurrent shared memory access.

AFAICT, `__syncthread` and `__syncwarp` should be enough to make sure that access to shared memory works correctly. I would be curious to hear why this might not be sufficient.

PhilMiller:
The same overloads appear in documentation and examples for `parallel_scan` functors as well. I haven't yet gone over their implementation to see whether the `volatile` bits can just be dropped there, too.

PhilMiller:
In my ongoing implementation effort, I’ve reached a point where I can displace the volatile qualifiers from the join operations on every reduce functor to copy and/or assignment operations on the underlying value types. If forming a reference through a volatile T* is sufficient to actually guarantee the code will access the shared memory location, then I don’t need volatile-qualified copy/assignment of the value type, either

PhilMiller:
[...]

PR #5412, opened 2022-08-31
#4077 Follow-on cleanup: drop volatile-qualified members from Pair and Complex

PhilMiller:
Inspired by documentation work in https://github.com/kokkos/kokkos-core-wiki/issues/151, whose corresponding PR will reflect this

### DIFF
--- a/core/src/Kokkos_Complex.hpp
+++ b/core/src/Kokkos_Complex.hpp
@@ -279,11 +279,7 @@ class
     return *this;
   }
 
-  //---------------------------------------------------------------------------
-  // TODO: refactor Kokkos reductions to remove dependency on
-  // volatile member overloads since they are being deprecated in c++20
-  //---------------------------------------------------------------------------
-
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
   //! Copy constructor from volatile.
   template <
       class RType,
@@ -436,8 +432,7 @@ class
     re_ *= src;
     im_ *= src;
   }
-
-  // TODO DSH 2019-10-7 why are there no volatile /= and friends?
+#endif  // KOKKOS_ENABLE_DEPRECATED_CODE_4
 };
 
 //==============================================================================
--- a/core/src/Kokkos_Complex.hpp
+++ b/core/src/Kokkos_Complex.hpp
@@ -284,7 +284,8 @@ class
   template <
       class RType,
       std::enable_if_t<std::is_convertible<RType, RealType>::value, int> = 0>
-  KOKKOS_INLINE_FUNCTION complex(const volatile complex<RType>& src) noexcept
+  KOKKOS_INLINE_FUNCTION KOKKOS_DEPRECATED
+  complex(const volatile complex<RType>& src) noexcept
       // Intentionally do the conversions implicitly here so that users don't
       // get any warnings about narrowing, etc., that they would expect to get
       // otherwise.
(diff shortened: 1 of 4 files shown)


PhilMiller:
> Please guard it with `DEPRECATED_CODE_4`

I expect they're probably unused. 

Could we go with 3, and see if anyone complains once 4.0 ships, so we can drop them in 4.1? I know we're technically late, but it may be immaterial.

dalg24:
> > Please guard it with `DEPRECATED_CODE_4`
> 
> I expect they're probably unused.
> 
> Could we go with 3, and see if anyone complains once 4.0 ships, so we can drop them in 4.1? I know we're technically late, but it may be immaterial.

I would be fine with 4.  I think Christian was leaning towards 3.  @crtrott please advise.

core/src/Kokkos_Complex.hpp
```diff
@@ -278,11 +278,7 @@ class
     return *this;
   }
 
-  //---------------------------------------------------------------------------
-  // TODO: refactor Kokkos reductions to remove dependency on
-  // volatile member overloads since they are being deprecated in c++20
-  //---------------------------------------------------------------------------
-
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_3
```
PhilMiller, line 281: Marked with 3 for the moment, but of course trivial to change if that's where consensus points

PhilMiller:
Following comments from @dalg24 I concluded the following
>  I’m considering the possibility of code that was written to work with 3.6 and later adapted to also compile without warnings or deprecation errors on 3.7/4.0 - i.e. It defines some custom join or operator+= for both volatile qualified arguments, and non-qualified arguments. At present, the code will compile without warnings against 3.7/develop, and use the non-qualified overloads. If the overloads in that PR are made unavailable, then the old disused volatile qualified overloads will now fail to compile, because the operators they might call will be missing. So, that leads me to agree to mark it 4 and not 3.

I will amend and rebase

dalg24 (approved):

masterleinad (commented):
Should we deprecate the properly with `[[deprecated]]`?

PhilMiller:
> Should we deprecate them properly with `KOKKOS_DEPRECATED`?

... Yes

cz4rs (approved):

core/src/Kokkos_Macros.hpp
```diff
@@ -603,8 +603,9 @@ static constexpr bool kokkos_omp_on_host() { return false; }
 // Guard intel compiler version 19 and older
 // intel error #2651: attribute does not apply to any entity
 // using <deprecated_type> KOKKOS_DEPRECATED = ...
-#if defined(KOKKOS_ENABLE_DEPRECATION_WARNINGS) && !defined(__NVCC__) && \
-    (!defined(KOKKOS_COMPILER_INTEL) || KOKKOS_COMPILER_INTEL >= 2021)
+#if defined(KOKKOS_ENABLE_DEPRECATION_WARNINGS) && !defined(__NVCC__) &&  \
+    (!defined(KOKKOS_COMPILER_INTEL) || KOKKOS_COMPILER_INTEL >= 2021) && \
+    !defined(KOKKOS_ENABLE_OPENACC)
```
dalg24, line 608: I find it interesting that it works in the NVHPC build and not in the OPENACC one.
They use nvc++ version 22.7 and 22.3 respectively.
Let's check if it works with more recent version of the compiler and bump the version if that happens to do the trick
@seyonglee @pedrovalerolara

dalg24: Nevermind I see the same issue with 22.9
We need to report this.

dalg24: #5604 fixes the issue

Issue #5414, opened 2022-08-31
Algorithms RandomAndSort test should be split into multiple cpp files


brian-kelley:
Right now, KokkosAlgorithms_UnitTest_RandomAndSort is compiled from a single cpp file. This file (dummy.cpp) is generated by cmake once for each backend to include TestRandomCommon.hpp, TestSortCommon.hpp and TestNestedSort.hpp. Instead, each of these should be its own .cpp file and compiled independently to object files, which are linked to make the final test executable.

PR #5415, opened 2022-08-31
Split Random/Sort/NestedSort test into multiple cpps

brian-kelley:
(Address #5414 )

Instead of including the Sort, NestedSort and Random tests into a single .cpp and compiling that into UnitTest_RandomAndSort, create one cpp file for each test header. Compile these into the same single test executable.

Note: I'm assuming that it's intentional that the final test only runs on one backend (whichever enabled backend is last in the list ``Threads;Serial;OpenMP;Cuda;HPX;HIP;SYCL;OpenMPTarget``). If not, the line ``set(SOURCES_A)`` should be moved outside this loop over backends, so that it doesn't get reset to empty for each one.

### DIFF
--- a/algorithms/unit_tests/CMakeLists.txt
+++ b/algorithms/unit_tests/CMakeLists.txt
@@ -31,7 +31,11 @@ foreach(Tag Threads;Serial;OpenMP;Cuda;HPX;HIP;SYCL;OpenMPTarget)
     # Each of these inputs is an .hpp file.
     # Generate a .cpp file for each one that runs it on the current backend (Tag),
     # and add this .cpp file to the sources for UnitTest_RandomAndSort.
-    foreach(SOURCES_A_Input TestRandomCommon;TestSortCommon;TestNestedSort)
+    foreach(SOURCES_A_Input
+        TestRandomCommon
+        TestSortCommon
+        TestNestedSort
+    )
       set(file ${dir}/${SOURCES_A_Input}.cpp)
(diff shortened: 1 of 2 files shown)


algorithms/unit_tests/CMakeLists.txt
```diff
@@ -28,17 +28,20 @@ foreach(Tag Threads;Serial;OpenMP;Cuda;HPX;HIP;SYCL;OpenMPTarget)
...
-      "#include <TestRandomCommon.hpp>\n"
-      "#include <TestSortCommon.hpp>\n"
-      "#include <TestNestedSort.hpp>\n"
-      )
-    configure_file(${dir}/dummy.cpp ${file})
-    list(APPEND SOURCES_A ${file})
+    # Each of these inputs is an .hpp file.
+    # Generate a .cpp file for each one that runs it on the current backend (Tag),
+    # and add this .cpp file to the sources for UnitTest_RandomAndSort.
+    foreach(SOURCES_A_Input TestRandomCommon;TestSortCommon;TestNestedSort)
```
dalg24, line 34: I would prefer if you broke lines (one per test) for readability and for cleaner diff when we add/rm tests
```suggestion
    foreach(SOURCES_A_Input
        TestRandomCommon
        TestSortCommon
        TestNestedSort
    )
```

dalg24 (approved):
Looks good.  I'd prefer if you applied the change I suggested

fnrizzi (approved):

PR #5416, opened 2022-08-31
Require ROCm 5.2.0

masterleinad:
We need a newer `ROCm` version for #5391 anyway and it appears that Sandia is about ready to support `ROCm 5.2.0`.


cmake/kokkos_compiler_id.cmake
```diff
@@ -189,8 +189,7 @@ ELSEIF(KOKKOS_CXX_COMPILER_ID STREQUAL NVIDIA)
   ENDIF()
   SET(CMAKE_CXX_EXTENSIONS OFF CACHE BOOL "Kokkos turns off CXX extensions" FORCE)
 ELSEIF(KOKKOS_CXX_COMPILER_ID STREQUAL HIPCC)
-  # Note that ROCm 4.5 reports as version 4.4
```
masterleinad, line 192: I am seeing
```
-- Compiler Version: 5.2.21151
```
with `rocm/5.2.0`.

crtrott (approved):

JBludau (approved):
looks good to me

dalg24 (dismissed):
You need to update the actual Dockerfile too.
There seems to be some misunderstanding on when to update the build argument and when to also change the Dockerfile.  The rule of thumb is that building without specifying anything must yield a working image.  Build arguments are used to build variant of the image and sometime repeat the explicitly defaults for readability in the Jenkinsfile

masterleinad:
> You need to update the actual Dockerfile too.

I did change `Dockerfile.hip`, too, see https://github.com/kokkos/kokkos/pull/5416/files#diff-c487a6a2ca5e33b1b80d0da585611bf7c0b0ea46618e1673a5cc30e9c3e2f68f.

dalg24:
> > You need to update the actual Dockerfile too.
> 
> I did change `Dockerfile.hip`, too, see https://github.com/kokkos/kokkos/pull/5416/files#diff-c487a6a2ca5e33b1b80d0da585611bf7c0b0ea46618e1673a5cc30e9c3e2f68f.

Duh...  My bad.  Sorry.

dalg24:
Waiting so Bruno gets a chance to review.  LGTM.

fnrizzi (approved):

masterleinad:
Retest this please.

masterleinad:
> You also need to update `scripts/testing_scripts/test_all_sandia`

I don't feel very comfortable doing that and would rather have @ndellingwood (or someone else that has access to the relevant machines) do that in a follow-up.

dalg24:
I agree that these can be updated in a follow-up PR.  Bruno, any objection other than the Sandia scripts?

Rombur (approved):

PR #5418, opened 2022-09-01
Remove code only used by ROCm < 5.0

Rombur:


core/src/HIP/Kokkos_HIP_Abort.hpp
```diff
@@ -50,17 +50,7 @@
 
 #include <hip/hip_runtime.h>
 
-// FIXME_HIP ROCm 4.5 version header include would be <rocm/rocm_version.h>
-#if __has_include(<rocm_version.h>)
 #include <rocm_version.h>
```
masterleinad, line 53:
```suggestion
```
What's the reason for keeping this include file?

dalg24: I don't think you get the version if you don't include

masterleinad: Right, but we are not using the version anywhere anymore AFAICT. I don't care much either way but just want this to be a conscious choice.

Rombur: You are right, this is not needed anymore. I've removed it.

core/src/HIP/Kokkos_HIP_Abort.hpp
```diff
@@ -50,18 +50,6 @@
 
 #include <hip/hip_runtime.h>
 
-// FIXME_HIP ROCm 4.5 version header include would be <rocm/rocm_version.h>
-#if __has_include(<rocm_version.h>)
-#include <rocm_version.h>
-#define KOKKOS_IMPL_ROCM_VERSION \
-  ROCM_VERSION_MAJOR * 10000 + ROCM_VERSION_MINOR * 100 + ROCM_VERSION_PATCH
```
dalg24, line 57: @Rombur you might want to include this in `HIP::print_configuration()` (see #5419 )

masterleinad (approved):

core/src/HIP/Kokkos_HIP_Abort.hpp
masterleinad: If so, it would probably more sense to have it somewhere else, right?

dalg24: Yes

Issue #5426, opened 2022-09-02
Faulty warning about missing return statement in CUDA


crtrott:
With NVCC 11.2 and GCC 9.1 I get the following:
```
/ascldap/users/crtrott/Kokkos/kokkos/core/src/impl/Kokkos_ViewCtor.hpp(289): warning: missing return statement at end of non-void function "Kokkos::Impl::with_properties_if_unset(const Kokkos::Impl::ViewCtorProp<P...> &, const Property &, const Properties &...) [with P=<std::string>, Property=std::string, Properties=<>]"
          detected during:
            instantiation of "auto Kokkos::Impl::with_properties_if_unset(const Kokkos::Impl::ViewCtorProp<P...> &, const Property &, const Properties &...) [with P=<std::string>, Property=std::string, Properties=<>]"
/ascldap/users/crtrott/Kokkos/kokkos/core/src/Kokkos_View.hpp(1408): here
            instantiation of "Kokkos::View<DataType, Properties...>::View(const Kokkos::Impl::ViewCtorProp<P...> &, const std::enable_if_t<<expression>, Kokkos::ViewTraits<DataType, Properties...>::array_layout> &, Kokkos::View<DataType, Properties...>::check_input_args) [with DataType=uint32_t *, Properties=<Kokkos::CudaSpace>, P=<std::string>]"
/ascldap/users/crtrott/Kokkos/kokkos/core/src/Kokkos_View.hpp(1567): here
            instantiation of "Kokkos::View<DataType, Properties...>::View(const Label &, std::enable_if_t<Kokkos::Impl::is_view_label<Label>::value, const size_t>, size_t, size_t, size_t, size_t, size_t, size_t, size_t) [with DataType=uint32_t *, Properties=<Kokkos::CudaSpace>, Label=char [29]]"
/ascldap/users/crtrott/Kokkos/kokkos/core/src/Cuda/Kokkos_Cuda_UniqueToken.hpp(82): here
```

Note the code definitely has a return statement in both branches. 

cmake: 
```
cmake -DCMAKE_INSTALL_PREFIX=${HOME}/Kokkos/install/develop/Cuda/NVCC/11.2 -DKokkos_ENABLE_CUDA=ON -DKokkos_ENABLE_CUDA_LAMBDA=ON -DKokkos_ARCH_VOLTA70=ON ~/Kokkos/kokkos
```

The warning does not appear with CUDA 11.7 nor 11.6 and GCC 9.1

[...]

PR #5427, opened 2022-09-02
(Develop Only) Fixes issue 5426: spurious warning for NVCC < 11.5 about missing return statement

crtrott:
See #5426

### DIFF
--- a/core/src/Kokkos_Macros.hpp
+++ b/core/src/Kokkos_Macros.hpp
@@ -116,7 +116,11 @@
 // Code is parsed and separated into host and device code.
 // Host code is compiled again with another compiler.
 // Device code is compile to 'ptx'.
-#define KOKKOS_COMPILER_NVCC __NVCC__
+// NOTE: There is no __CUDACC_VER_PATCH__ officially, its __CUDACC_VER_BUILD__
+// which does have more than one digit (potentially undefined number of them).
+// This macro definition is in line with our other compiler defs
+#define KOKKOS_COMPILER_NVCC \
+  __CUDACC_VER_MAJOR__ * 100 + __CUDACC_VER_MINOR__ * 10
 #endif  // #if defined( __NVCC__ )
 
 #if !defined(KOKKOS_LAMBDA)
--- a/core/src/impl/Kokkos_ViewCtor.hpp
+++ b/core/src/impl/Kokkos_ViewCtor.hpp
@@ -335,7 +335,7 @@ KOKKOS_FUNCTION const auto &get_property(
 #if defined(KOKKOS_COMPILER_NVCC) && (KOKKOS_COMPILER_NVCC < 1150)
 // pragma pop is getting a warning from the underlying GCC
 // for unknown pragma if -pedantic is used
-#ifdef __CUDAR_ARCH__
+#ifdef __CUDA_ARCH__
 #pragma pop
 #endif
 #endif
--- a/core/src/impl/Kokkos_ViewCtor.hpp
+++ b/core/src/impl/Kokkos_ViewCtor.hpp
@@ -267,6 +267,16 @@ auto with_properties_if_unset(const ViewCtorProp<P...> &view_ctor_prop) {
   return view_ctor_prop;
 }
 
+// Suppress faulty warning from NVCC see:
+// https://github.com/kokkos/kokkos/issues/5426
+#if defined(KOKKOS_COMPILER_NVCC) && (KOKKOS_COMPILER_NVCC < 1150)
+// pragma pop is getting a warning from the underlying GCC
+// for unknown pragma if -pedantic is used
+#ifdef __CUDA_ARCH__
+#pragma push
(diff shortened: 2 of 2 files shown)


core/src/impl/Kokkos_ViewCtor.hpp
```diff
@@ -267,6 +267,12 @@ auto with_properties_if_unset(const ViewCtorProp<P...> &view_ctor_prop) {
   return view_ctor_prop;
 }
 
+// Suppress faulty warning from NVCC see:
+// https://github.com/kokkos/kokkos/issues/5426
+#if defined(KOKKOS_COMPILER_NVCC) && (KOKKOS_COMPILER_NVCC < 1150)
+#pragma push
+#pragma diag_suppress = implicit_return_from_non_void_function
```
dalg24, line 274:
```suggestion
#pragma diag_suppress implicit_return_from_non_void_function
```

crtrott:
So: the warning we are getting (and which is turned into an error) comes from GCC for an unknown pragma. I can go one step further and make this device code specific however ...

dalg24:
> So: the warning we are getting (and which is turned into an error) comes from GCC for an unknown pragma. I can go one step further and make this device code specific however ...

Do you know why the warning was only emitted for one of the builds?

crtrott:
Only one build which used NVCC older than 11.5 to build. The other ones are newer (CUDA 11.6/11.7) or use Clang as CUDA compiler.

core/src/Kokkos_Macros.hpp
```diff
@@ -116,7 +116,11 @@
 // Code is parsed and separated into host and device code.
 // Host code is compiled again with another compiler.
 // Device code is compile to 'ptx'.
-#define KOKKOS_COMPILER_NVCC __NVCC__
+// NOTE: There is no __CUDACC_VER_PATCH__ officially, its __CUDACC_VER_BUILD__
+// which does have more than one digit (potentially undefined number of them).
+// This macro definition is in line with our other compiler defs
+#define KOKKOS_COMPILER_NVCC \
+  __CUDACC_VER_MAJOR__ * 100 + __CUDACC_VER_MINOR__ * 10
```
masterleinad, line 123: Is there a good reason to not define this as `CUDA_VERSION`?

core/src/impl/Kokkos_ViewCtor.hpp
```diff
@@ -322,6 +332,13 @@ KOKKOS_FUNCTION const auto &get_property(
     return view_ctor_prop;
   }
 }
+#if defined(KOKKOS_COMPILER_NVCC) && (KOKKOS_COMPILER_NVCC < 1150)
+// pragma pop is getting a warning from the underlying GCC
+// for unknown pragma if -pedantic is used
+#ifdef __CUDAR_ARCH__
```
PhilMiller, line 338:
```suggestion
#ifdef __CUDA_ARCH__
```
Typo in the macro name

core/src/Kokkos_Macros.hpp
crtrott: This is about which version of NVCC this is independent of whether we compile with clang or nvc++. Furthermore, I thought that KOKKOS_COMPILER_FOO is for most other compilers defined as FOO_VER_MAJOR*100+FOO_VER_MINOR*10+FOO_VER_PATCH.
Essentially there is a difference between checking for CUDA version and compiler version for the compiler you use to compile CUDA. Just for NVCC its kinda the same (except our fuckup with being inconsistent with KOKKOS_COMPILER_CUDA_VERSION)

core/src/impl/Kokkos_ViewCtor.hpp
crtrott: gah ...

core/src/Kokkos_Macros.hpp
masterleinad: Yeah, we should probably try to remove `KOKKOS_COMPILER_CUDA_VERSION`.

core/src/impl/Kokkos_ViewCtor.hpp
```diff
@@ -267,6 +267,16 @@ auto with_properties_if_unset(const ViewCtorProp<P...> &view_ctor_prop) {
...
+// Suppress faulty warning from NVCC see:
+// https://github.com/kokkos/kokkos/issues/5426
+#if defined(KOKKOS_COMPILER_NVCC) && (KOKKOS_COMPILER_NVCC < 1150)
+// pragma pop is getting a warning from the underlying GCC
+// for unknown pragma if -pedantic is used
+#ifdef __CUDA_ARCH__
+#pragma push
+#pragma diag_suppress implicit_return_from_non_void_function
+#endif
+#endif
```
masterleinad, line 279: I think this is better than trying to do any tag dispatch or SFINAE.

masterleinad (approved):
Thanks! Looks good to me.

dalg24 (approved):

Issue #4767, opened 2022-02-08
Current status of mathematical functions


dalg24:
Summarize below what is currently provided by Kokkos and what is missing.

x means already there
A denotes functions being added shortly

# Basic operations
functions | Kokkos | notes
--------- | ------ | -----
`abs` | x
`fabs` | x
`fmod` | x
`remainder` | x
`remquo` | | `remquo( Arithmetic1 x, Arithmetic2 y, int* quo) -> Promoted` available
`fma` |  |  `fma(Arithmetic1 x, Arithmetic2 y, Arithmetic3 z -> Promoted` available
`fmax` | x
`fmin` | x
`nan` | x

# Exponential functions
functions | Kokkos
--------- | ------
`exp` | x
`exp2` | x
`expm1` | x
`log` | x
`log10` | x
`log2` | x
`log1p` | x

# Power functions
functions | Kokkos
--------- | ------
`pow` | x
`sqrt` | x
`cbrt` | x
`hypot` | x

# Trigonometric functions
functions | Kokkos
--------- | ------
`sin` | x
`cos` | x
`tan` | x
`asin` | x
`acos` | x
`atan` | x
`atan2` | x
# Hyperbolic functions
functions | Kokkos
--------- | ------
`sinh` | x
`cosh` | x
`tanh` | x
`asinh` | x
`acosh` | x
`atanh` | x

# Error and gamma functions
functions | Kokkos
--------- | ------
`erf` | x
`erfc` | x
`tgamma` | x
`lgamma` | x

# Nearest integer floating point operations
functions | Kokkos | notes
--------- | ------ | -----
`ceil` | x
`floor` | x
`trunc` | x
`round` | A
`lround` | A | not available in SYCL
`llround` | A | not available in SYCL
`nearbyint` | x | not available in SYCL
`rint` | A
`lrint` | A | not available in SYCL
`llrint` | A | not available in SYCL

# Floating point manipulation functions
functions | Kokkos | notes
--------- | ------ | -----
`frexp` | | `frexp(Arithmetic arg, int* exp) -> Promoted` available in CUDA and SYCL
`ldexp` | | `ldexp(Arithmetic arg, int exp) -> Promoted` available in CUDA and SYCL
`modf` | | `modf(FloatingPoint arg, FloatingPoint* iptr) -> FloatingPoint` available in CUDA and SYCL
`scalbn` | | `scalbn(Arithmetic arg, int exp) -> Promoted` available in CUDA
[...]

PR #5428, opened 2022-09-06
adding Kokkos::fma(x, y, z)

ibaned:
This is for consistency with providing
the same functions as <cmath> in
namespace Kokkos, and important to
ensure that eventual users of SIMD
can use Kokkos::fma in templated code
and get SIMD FMA instructions when
expected


dalg24 (changes requested):
Please add it there
https://github.com/kokkos/kokkos/blob/3be4cef23d41b91353e045af2f28fe498097d4ce/core/src/Kokkos_MathematicalFunctions.hpp#L351

Also we need to add a unit test.  Let me know if you are unable to or need help with it.

core/unit_test/TestMathematicalFunctions.hpp
```diff
@@ -424,7 +424,7 @@ DEFINE_BINARY_FUNCTION_EVAL(copysign, 1);
     template <typename T, typename U, typename V>                             \
     static KOKKOS_FUNCTION auto eval(T x, U y, V z) {                         \
       static_assert(                                                          \
-          std::is_same<decltype(Kokkos::FUNC((T)0, (U)0), (V)0),              \
+          std::is_same<decltype(Kokkos::FUNC((T)0, (U)0, (V)0)),              \
```
crtrott, line 427: Uhoh ...

crtrott (approved):
I think this is good.

dalg24 (approved):

core/unit_test/TestMathematicalFunctions.hpp
ibaned: The good old comma operator...

dalg24: Was going through the 2-arg version...  Thanks for fixing

dalg24:
Adding reference to #4767

dalg24:
HIP failure on the new machine that I was setting up yesterday is clearly unrelated

PR #5430, opened 2022-09-06
Remove KOKKOS_COMPILER_CUDA_VERSION

masterleinad:
The few cases are in fact not necessary anymore and the proper replacement would be using `KOKKOS_COMPILER_NVCC` with the updates in #5427. Also, note that this wasn't consistently defined in the `Makefile` vs `CMake` approach.
It seems that the macro doesn't have much use downstream but keeping it is of course up for discussion.


dalg24 (approved):

PhilMiller (approved):

PhilMiller:
I corrected the issue reference in the pr description

PR #5434, opened 2022-09-07
Promote math constants `Kokkos::{Experimental -> numbers}` and add missing variables without the `_v` suffix

dalg24:
Rational: There is no good reason to keep them in `Kokkos::Experimental::` since it is just a backport of the C++20 facility.  I opted for the `Kokkos::numbers` namespace for consistency with the C++ standard.  Let me know if you want to capitalize the "N".

Drive-by changes:
* add missing math constant variables without the `_v` suffix

### DIFF
--- a/core/src/impl/Kokkos_QuadPrecisionMath.hpp
+++ b/core/src/impl/Kokkos_QuadPrecisionMath.hpp
@@ -208,8 +208,7 @@ inline bool signbit(__float128 x) { return ::signbitq(x); }
 //</editor-fold>
 
 //<editor-fold desc="Mathematical constants __float128 specializations">
-namespace Kokkos {
-namespace Experimental {
+namespace Kokkos::numbers {
 // clang-format off
 template <> constexpr __float128 e_v         <__float128> = 2.718281828459045235360287471352662498Q;
 template <> constexpr __float128 log2e_v     <__float128> = 1.442695040888963407359924681001892137Q;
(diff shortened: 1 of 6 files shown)


Rombur (approved):

core/src/impl/Kokkos_QuadPrecisionMath.hpp
```diff
@@ -208,8 +208,7 @@ inline bool signbit(__float128 x) { return ::signbitq(x); }
 //</editor-fold>
 
 //<editor-fold desc="Mathematical constants __float128 specializations">
-namespace Kokkos {
-namespace Experimental {
+namespace Kokkos::numbers {
```
ldh4, line 211: I believe every Kokkos namespace starts with a capital letter (`Kokkos`, `Impl`, `Experimental`, `Test`, etc). I think it'll be better if we keep that consistent here as well.

crtrott:
Lets discuss the namespace question today in the meeting.

core/src/impl/Kokkos_QuadPrecisionMath.hpp
dalg24: This was discussed at the developer meeting.  Consensus was for `Kokkos::numbers` lowercase for consistency with C++20.

ldh4 (approved):

dalg24:
HIP failure was my rebooting the new machine

PR #5435, opened 2022-09-07
HIP: Do not warn about `XNACK` when page migration is not supported

dalg24:
Spotted in the CI on gfx906
```
4: Kokkos::HIP::allocation WARNING: The combination of device and system configuration
4:                                  does not support page migration between device and host.
4:                                  HIPManagedSpace might not work as expected.
4:                                  Please refer to the ROCm documentation on unified/managed memory.
4: 
4: Kokkos::HIP::runtime WARNING: Kokkos did not find an environment variable 'HSA_XNACK'
4:                               for the current process.
4:                               Nevertheless, xnack is enabled for all processes if
4:                               amdgpu.noretry=0 was set in the Linux kernel boot line.
4:                               Without xnack enabled, Kokkos::HIPManaged might not behave
4:                               as expected.
```
It feels silly to me to emit the 2nd warning after the first one.

### DIFF
--- a/core/src/HIP/Kokkos_HIP_Space.cpp
+++ b/core/src/HIP/Kokkos_HIP_Space.cpp
@@ -180,32 +180,36 @@ void* HIPManagedSpace::impl_allocate(
   if (arg_alloc_size > 0) {
     if (is_first_hip_managed_allocation.exchange(false) &&
         Kokkos::show_warnings()) {
-      if (!hip_driver_check_page_migration(m_device)) {
-        std::cerr << R"warning(
+      do {  // hack to avoid spamming users with too many warnings
+        if (!hip_driver_check_page_migration(m_device)) {
+          std::cerr << R"warning(
 Kokkos::HIP::allocation WARNING: The combination of device and system configuration
                                  does not support page migration between device and host.
                                  HIPManagedSpace might not work as expected.
                                  Please refer to the ROCm documentation on unified/managed memory.)warning"
-                  << std::endl;
-      }
+                    << std::endl;
+          break;  // do not warn about HSA_XNACK environement variable
+        }
 
-      // check for correct runtime environment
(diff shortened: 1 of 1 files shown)


Rombur (approved):

core/src/HIP/Kokkos_HIP_Space.cpp
```diff
@@ -180,32 +180,36 @@ void* HIPManagedSpace::impl_allocate(
...
-      if (!hsa_xnack)
-        std::cerr << R"warning(
+                    << std::endl;
+          break;  // do not warn about HSA_XNACK environement variable
+        }
+
+        // check for correct runtime environment
+        const char* hsa_xnack = std::getenv("HSA_XNACK");
+        if (!hsa_xnack)
+          std::cerr << R"warning(
```
JBludau, line 197: why the do-while hack and not just changing the `if` to an `else if` here and move the `getenv` to the top? I think it would be easier to read

JBludau (commented):
maybe I am not seeing something here, but it looks to me as if there would be a less hacky solution

core/src/HIP/Kokkos_HIP_Space.cpp
dalg24: I didn't want to have a 4-level nested if-statement which I find arguably harder to read.
Let me know if you want me to change.

JBludau: I leave this to your judgement

JBludau (approved):

core/src/HIP/Kokkos_HIP_Space.cpp
dalg24: @Rombur Your call.  Do you want to change or keep as is?

crtrott (approved):

PR #3286, opened 2020-08-17
cuda_abort: remove [[no return]] with KOKKOS_ENABLE_DEBUG_BOUNDS_CHECK

ndellingwood:
addresses cuda random number generator test failures on Pre-Volta archs

PR #3106, opened 2020-06-17
Apply [[noreturn]] to Kokkos::abort when applicable

AndrewGaspar:
I haven't run any tests beyond just building. Pushing to run CI.

I removed the `__APPLE__` check for `cuda_abort` since I don't think macOS supports CUDA in recent versions anyway. I can put it back if this configuration is still supported, though.

PR #5441, opened 2022-09-07
Suppress warning: function declared with "noreturn" does return for CUDA and Debug mode

### LINKED ISSUE #5440: Suppress warning: function declared with "noreturn" does return
With latest `develop` when `KOKKOS_ENABLE_DEBUG_BOUNDS_CHECK` is enabled, I get a ton of

```
/kokkos/core/src/impl/Kokkos_Error.hpp(122): warning: function declared with "noreturn" does return
```

i.e. one warning for every Kokkos file that includes this header, which is pretty distracting. It would be good to suppress the warning, since it isn't helpful.

This warning also shows up for v3.7 RC, but not for v3.6.

masterleinad:
Fixes #5440.
For compilers that do two passes for device functions, we need to make sure that also the host pass for the device function gets the correct attribute. Notably, that is `SYCL` and `CUDA` for `macOS` or if `KOKKOS_ENABLE_DEBUG_BOUNDS_CHECK` is enabled.

### DIFF
--- a/core/src/Cuda/Kokkos_CudaSpace.cpp
+++ b/core/src/Cuda/Kokkos_CudaSpace.cpp
@@ -122,14 +122,7 @@ void DeepCopyAsyncCuda(void *dst, const void *src, size_t n) {
 
 namespace Kokkos {
 
-bool CudaUVMSpace::available() {
-#if defined(CUDA_VERSION) && !defined(__APPLE__)
-  enum : bool { UVM_available = true };
-#else
-  enum : bool { UVM_available = false };
-#endif
-  return UVM_available;
-}
+bool CudaUVMSpace::available() { return true; }
 
 /*--------------------------------------------------------------------------*/
 
--- a/core/src/Cuda/Kokkos_Cuda_Instance.cpp
+++ b/core/src/Cuda/Kokkos_Cuda_Instance.cpp
@@ -202,8 +202,8 @@ void cuda_internal_error_abort(cudaError e, const char *name, const char *file,
   if (file) {
     out << " " << file << ":" << line;
   }
-  // Call Kokkos::Impl::host_abort instead of Kokkos::abort to avoid a warning
-  // about Kokkos::abort returning in some cases.
+  // FIXME Call Kokkos::Impl::host_abort instead of Kokkos::abort to avoid a
+  // warning about Kokkos::abort returning in some cases.
   host_abort(out.str().c_str());
 }
 
--- a/core/src/Cuda/Kokkos_Cuda_Instance.cpp
+++ b/core/src/Cuda/Kokkos_Cuda_Instance.cpp
@@ -202,9 +202,7 @@ void cuda_internal_error_abort(cudaError e, const char *name, const char *file,
   if (file) {
     out << " " << file << ":" << line;
   }
-  // Call Kokkos::Impl::host_abort instead of Kokkos::abort to avoid a warning
-  // about Kokkos::abort returning in some cases.
-  host_abort(out.str().c_str());
+  abort(out.str().c_str());
 }
 
 //----------------------------------------------------------------------------
--- a/core/src/Cuda/Kokkos_Cuda_Instance.cpp
+++ b/core/src/Cuda/Kokkos_Cuda_Instance.cpp
@@ -202,7 +202,9 @@ void cuda_internal_error_abort(cudaError e, const char *name, const char *file,
   if (file) {
     out << " " << file << ":" << line;
   }
-  abort(out.str().c_str());
+  // Call Kokkos::Impl::host_abort instead of Kokkos::abort to avoid a warning
+  // about Kokkos::abort returning in some cases.
+  host_abort(out.str().c_str());
 }
 
 //----------------------------------------------------------------------------
--- a/core/src/impl/Kokkos_Error.hpp
+++ b/core/src/impl/Kokkos_Error.hpp
@@ -69,7 +69,14 @@ namespace Impl {
 
 #if defined(KOKKOS_ENABLE_CUDA) && defined(__CUDA_ARCH__)
 
+#if defined(KOKKOS_ENABLE_DEBUG_BOUNDS_CHECK)
+// required to workaround failures in random number generator unit tests with
+// pre-volta architectures
+#define KOKKOS_IMPL_ABORT_NORETURN
+#else
+// cuda_abort aborts when building for other platforms than macOS
 #define KOKKOS_IMPL_ABORT_NORETURN [[noreturn]]
+#endif
 
 #elif defined(KOKKOS_COMPILER_NVHPC)
 
@@ -89,7 +96,10 @@ namespace Impl {
 #define KOKKOS_IMPL_ABORT_NORETURN
 #endif
 
-#ifdef KOKKOS_ENABLE_SYCL  // FIXME_SYCL
+// FIXME_SYCL
+// Accomodate host pass for device functions that are not [[noreturn]]
+#if defined(KOKKOS_ENABLE_SYCL) || \
+    (defined(KOKKOS_ENABLE_CUDA) && defined(KOKKOS_ENABLE_DEBUG_BOUNDS_CHECK))
 #define KOKKOS_IMPL_ABORT_NORETURN_DEVICE
 #else
 #define KOKKOS_IMPL_ABORT_NORETURN_DEVICE KOKKOS_IMPL_ABORT_NORETURN
--- a/core/src/impl/Kokkos_Error.hpp
+++ b/core/src/impl/Kokkos_Error.hpp
@@ -69,8 +69,7 @@ namespace Impl {
 
 #if defined(KOKKOS_ENABLE_CUDA) && defined(__CUDA_ARCH__)
 
-#if defined(__APPLE__) || defined(KOKKOS_ENABLE_DEBUG_BOUNDS_CHECK)
-// cuda_abort does not abort when building for macOS.
+#if defined(KOKKOS_ENABLE_DEBUG_BOUNDS_CHECK)
 // required to workaround failures in random number generator unit tests with
 // pre-volta architectures
 #define KOKKOS_IMPL_ABORT_NORETURN
@@ -99,9 +98,8 @@ namespace Impl {
 
 // FIXME_SYCL
 // Accomodate host pass for device functions that are not [[noreturn]]
-#if defined(KOKKOS_ENABLE_SYCL) ||  \
-    (defined(KOKKOS_ENABLE_CUDA) && \
-     (defined(__APPLE__) || defined(KOKKOS_ENABLE_DEBUG_BOUNDS_CHECK)))
+#if defined(KOKKOS_ENABLE_SYCL) || \
+    (defined(KOKKOS_ENABLE_CUDA) && defined(KOKKOS_ENABLE_DEBUG_BOUNDS_CHECK))
 #define KOKKOS_IMPL_ABORT_NORETURN_DEVICE
 #else
 #define KOKKOS_IMPL_ABORT_NORETURN_DEVICE KOKKOS_IMPL_ABORT_NORETURN
--- a/core/src/impl/Kokkos_Error.hpp
+++ b/core/src/impl/Kokkos_Error.hpp
@@ -69,14 +69,7 @@ namespace Impl {
 
 #if defined(KOKKOS_ENABLE_CUDA) && defined(__CUDA_ARCH__)
 
-#if defined(KOKKOS_ENABLE_DEBUG_BOUNDS_CHECK)
-// required to workaround failures in random number generator unit tests with
-// pre-volta architectures
-#define KOKKOS_IMPL_ABORT_NORETURN
-#else
-// cuda_abort aborts when building for other platforms than macOS
 #define KOKKOS_IMPL_ABORT_NORETURN [[noreturn]]
-#endif
 
 #elif defined(KOKKOS_COMPILER_NVHPC)
 
@@ -96,10 +89,7 @@ namespace Impl {
 #define KOKKOS_IMPL_ABORT_NORETURN
 #endif
 
-// FIXME_SYCL
-// Accomodate host pass for device functions that are not [[noreturn]]
-#if defined(KOKKOS_ENABLE_SYCL) || \
-    (defined(KOKKOS_ENABLE_CUDA) && defined(KOKKOS_ENABLE_DEBUG_BOUNDS_CHECK))
+#ifdef KOKKOS_ENABLE_SYCL  // FIXME_SYCL
 #define KOKKOS_IMPL_ABORT_NORETURN_DEVICE
 #else
 #define KOKKOS_IMPL_ABORT_NORETURN_DEVICE KOKKOS_IMPL_ABORT_NORETURN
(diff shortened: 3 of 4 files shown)


PhilMiller (approved):

core/src/impl/Kokkos_Error.hpp
```diff
@@ -97,7 +97,11 @@ namespace Impl {
 #define KOKKOS_IMPL_ABORT_NORETURN
 #endif
 
-#ifdef KOKKOS_ENABLE_SYCL  // FIXME_SYCL
+// FIXME_SYCL
+// Accomodate host pass for device functions that are not [[noreturn]]
+#if defined(KOKKOS_ENABLE_SYCL) ||  \
+    (defined(KOKKOS_ENABLE_CUDA) && \
+     (defined(__APPLE__) || defined(KOKKOS_ENABLE_DEBUG_BOUNDS_CHECK)))
```
dalg24, line 104: How did you come up with this?  I thought CUDA is not supported on OSX.

masterleinad: See https://github.com/kokkos/kokkos/blob/d0f710de8010a476abb93f9f63a0b0bc27f7541f/core/src/impl/Kokkos_Error.hpp#L72-L80 and https://github.com/kokkos/kokkos/blob/d0f710de8010a476abb93f9f63a0b0bc27f7541f/core/src/Cuda/Kokkos_Cuda_abort.hpp#L67-L94.

dalg24:
It certainly does not support CUDA 11
I did some archeology and found #3106
There is no more reason to have it

masterleinad:
> t certainly does not support CUDA 11
> I did some archeology and found #3106
> There is no more reason to have it

What about clang++ using CUDA 10?

stanmoore1:
@masterleinad Looks great now, thanks!

dalg24:
> > t certainly does not support CUDA 11
> > I did some archeology and found #3106
> > There is no more reason to have it
> 
> What about clang++ using CUDA 10?

Does that actually work?

masterleinad:
> Does that actually work?

I don't know about `macOS` support but we test `clang-11` with `CUDA-10.1` in our CI.

masterleinad:
Also, see https://llvm.org/docs/CompileCudaWithLLVM.html#id3:
> CUDA is supported since llvm 3.9. Clang currently supports CUDA 7.0 through 11.5. If clang detects a newer CUDA version, it will issue a warning and will attempt to use detected CUDA SDK it as if it were CUDA 11.5. 
> [...]
> CUDA compilation is supported on Linux. Compilation on MacOS and Windows may or may not work and currently have no maintainers.

That being said, I don't have a problem with dropping special casing `MacOS+CUDA` whether it works or not.

crtrott (changes requested):
Lets drop the CUDA/APPLE combo. Don't think that matters right now, and might not in the future either

masterleinad:
> Lets drop the CUDA/APPLE combo. Don't think that matters right now, and might not in the future either

I removed the related macro guards.

core/src/Cuda/Kokkos_Cuda_Instance.cpp
```diff
@@ -202,7 +202,9 @@ void cuda_internal_error_abort(cudaError e, const char *name, const char *file,
   if (file) {
     out << " " << file << ":" << line;
   }
-  abort(out.str().c_str());
+  // Call Kokkos::Impl::host_abort instead of Kokkos::abort to avoid a warning
+  // about Kokkos::abort returning in some cases.
+  host_abort(out.str().c_str());
```
dalg24, line 207: I am not sure about that one.  I don't like spreading use of internal details if we can avoid.  This would be the first use of `Impl::host_abort` in the code base besides the definition of `Kokkos::abort`.
What other option do we have?

core/src/Cuda/Kokkos_CudaSpace.cpp
```diff
@@ -122,14 +122,7 @@ void DeepCopyAsyncCuda(void *dst, const void *src, size_t n) {
...
 
-bool CudaUVMSpace::available() {
-#if defined(CUDA_VERSION) && !defined(__APPLE__)
-  enum : bool { UVM_available = true };
-#else
-  enum : bool { UVM_available = false };
-#endif
-  return UVM_available;
-}
+bool CudaUVMSpace::available() { return true; }
```
dalg24, line 125: OK.  This is a candidate for DEPRECATED_CODE_4.  (Not asking you to do anything in this PR.)

core/src/Cuda/Kokkos_Cuda_Instance.cpp
masterleinad: >  What other option do we have?

We could just keep this one. It's only compiled once anyway.

cz4rs (approved):

core/src/impl/Kokkos_Error.hpp
```diff
@@ -69,8 +69,7 @@ namespace Impl {
 
 #if defined(KOKKOS_ENABLE_CUDA) && defined(__CUDA_ARCH__)
 
-#if defined(__APPLE__) || defined(KOKKOS_ENABLE_DEBUG_BOUNDS_CHECK)
-// cuda_abort does not abort when building for macOS.
+#if defined(KOKKOS_ENABLE_DEBUG_BOUNDS_CHECK)
```
dalg24, line 72: This comes from #3286
@ndellingwood would you please check if that is still an issue with our new minimum requirements?
I would much rather drop that silly workaround if we are able to.

ndellingwood: @dalg24 @masterleinad  I tested the PR sha 33038b6b855175cbe7a92a5bb2cd756f035cd7ca as-is with cuda/11.1+gcc/8.3.0 and I am seeing the warning:
```
/ascldap/users/ndellin/kokkos/core/src/impl/Kokkos_ViewCtor.hpp(304): warning: missing return statement at end of non-void function "Kokkos::Impl::with_properties_if_unset(const Kokkos::Impl::ViewCtorProp<P...> &, const Property &, const Properties &...) [with P=<Kokkos::Cuda, Kokkos::CudaSpace, std::__cxx11::string, Kokkos::Impl::WithoutInitializing_t>, Property=Kokkos::CudaSpace::execution_space, Properties=<>]"
...
/ascldap/users/ndellin/kokkos/core/src/impl/Kokkos_ViewCtor.hpp(339): warning: missing return statement at end of non-void function "Kokkos::Impl::get_property<Tag,P...>(const Kokkos::Impl::ViewCtorProp<P...> &) [with Tag=Kokkos::Impl::LabelTag, P=<Kokkos::Cuda, Kokkos::CudaSpace, std::__cxx11::string, Kokkos::Impl::WithoutInitializing_t>]"
...
```

I re-compiled with removal of the guards flagged in the message (line 72-79), leaving just
```
#define KOKKOS_IMPL_ABORT_NORETURN [[noreturn]]
```
and I'm seeing multiple warnings like
```
/ascldap/users/ndellin/kokkos/core/src/impl/Kokkos_Error.hpp(230): warning: function declared with "noreturn" does return
```

masterleinad: @ndellingwood The question was if removing the workaround for `KOKKOS_ENABLE_DEBUG_BOUNDS_CHECK` still breaks "CUDA random number generator tests on Pre-Volta archs" (#3286) and stuff in KokkosKernels (https://github.com/kokkos/kokkos-kernels/issues/785).

ndellingwood: The only pre-Volta architecture I have access to is Kepler but that is no longer a supported arch right?

masterleinad: @ndellingwood Kepler 3.5 and later should still be supported in CUDA 11.0.

ndellingwood: I tested the PR after making these changes to remove the `KOKKOS_ENABLE_DEBUG_BOUNDS_CHECK` guards:

```diff
diff --git a/core/src/impl/Kokkos_Error.hpp b/core/src/impl/Kokkos_Error.hpp
index 869bfb1..0447844 100644
--- a/core/src/impl/Kokkos_Error.hpp
+++ b/core/src/impl/Kokkos_Error.hpp
@@ -69,14 +69,8 @@ namespace Impl {
 
 #if defined(KOKKOS_ENABLE_CUDA) && defined(__CUDA_ARCH__)
 
-#if defined(KOKKOS_ENABLE_DEBUG_BOUNDS_CHECK)
-// required to workaround failures in random number generator unit tests with
-// pre-volta architectures
-#define KOKKOS_IMPL_ABORT_NORETURN
-#else
 // cuda_abort aborts when building for other platforms than macOS
 #define KOKKOS_IMPL_ABORT_NORETURN [[noreturn]]
-#endif
 
 #elif defined(KOKKOS_COMPILER_NVHPC)
```

with cuda/11.1+gcc/8.3.0 on Kepler35 with this configuration:
```bash
module load sems-gcc/8.3.0 sems-cuda/11.1.0 sems-cmake/3.23.1

cmake -DCMAKE_CXX_COMPILER=$KOKKOS_PATH/bin/nvcc_wrapper -DKokkos_ENABLE_CUDA=ON -DKokkos_ARCH_KEPLER35=ON -DKokkos_ENABLE_TESTS=ON -DKokkos_ENABLE_EXAMPLES=OFF -DKokkos_ENABLE_CUDA_LAMBDA=ON -DCMAKE_VERBOSE_MAKEFILE=ON -DCMAKE_CXX_EXTENSIONS=OFF -DCMAKE_BUILD_TYPE=RELEASE -DKokkos_ENABLE_DEBUG_BOUNDS_CHECK=ON -DKokkos_ENABLE_DEPRECATION_WARNINGS=OFF -DKokkos_ENABLE_DEPRECATED_CODE_4=OFF $KOKKOS_PATH
```

All algorithms tests (including random num generator) passed, saw repeats of this warning during compilation:

```
/ascldap/users/ndellin/kokkos/core/src/impl/Kokkos_Error.hpp(231): warning: function declared with "noreturn" does return
```

masterleinad: I think the crucial changes to test would be to remove the checks in `core/src/Cuda/Kokkos_Cuda_abort.hpp`, in particular activating  `while(true);`.

masterleinad: @ndellingwood Could you test with the commit I pushed?

ndellingwood: Tests also passed when removing the `while(true);` hack, i.e.

```diff
diff --git a/core/src/Cuda/Kokkos_Cuda_abort.hpp b/core/src/Cuda/Kokkos_Cuda_abort.hpp
index 642581b..94cba76 100644
--- a/core/src/Cuda/Kokkos_Cuda_abort.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_abort.hpp
@@ -81,10 +81,6 @@ __device__ inline void cuda_abort(const char *const message) {
   // __assertfail is not marked as [[noreturn]], even though it does not return.
   //  Disable with KOKKOS_ENABLE_DEBUG_BOUNDS_CHECK to workaround failures
   //  in random number generator unit tests with pre-volta architectures
-#if !defined(KOKKOS_ENABLE_DEBUG_BOUNDS_CHECK)
-  while (true)
-    ;
-#endif
 }
 
 }  // namespace Impl
```

I'll test one more incremental change, removing the other `KOKKOS_ENABLE_DEBUG_BOUNDS_CHECK` guard from `core/src/Cuda/Kokkos_Cuda_abort.hpp`

ndellingwood: Removing the last `KOKKOS_ENABLE_DEBUG_BOUNDS_CHECK` guarded region from `core/src/Cuda/Kokkos_Cuda_abort.hpp `, that is this diff (in addition to the changes previously listed)

```diff
-// required to workaround failures in random number generator unit tests with
-// pre-volta architectures
-#if defined(KOKKOS_ENABLE_DEBUG_BOUNDS_CHECK)
-__device__ inline void cuda_abort(const char *const message) {
-#else
 [[noreturn]] __device__ inline void cuda_abort(const char *const message) {
-#endif
   const char empty[] = "";
 
   __assertfail((const void *)message, (const void *)empty, (unsigned int)0,
                (const void *)empty, sizeof(char));
```

lead to test failures in the algorithms tests, in particular with the random num tests:
```
    Start 1: KokkosAlgorithms_UnitTest_RandomAndSort

1: Test command: /ascldap/users/ndellin/kokkos/testing/Pr5441/algorithms/unit_tests/KokkosAlgorithms_UnitTest_RandomAndSort
1: Test timeout computed to be: 10000000
1: [==========] Running 5 tests from 1 test suite.
1: [----------] Global test environment set-up.
1: [----------] 5 tests from cuda
1: [ RUN      ] cuda.Random_XorShift64
1: Test Seed:1665600133315178171
1: Test Scalar=int
1:  -- Testing randomness properties
1: /ascldap/users/ndellin/kokkos/algorithms/unit_tests/TestRandom.hpp:310: Failure
1: Expected: (std::abs(mean_eps)) < (tolerance), actual: 0.0406466 vs 0.000139188
1: /ascldap/users/ndellin/kokkos/algorithms/unit_tests/TestRandom.hpp:311: Failure
1: Expected: (std::abs(variance_eps)) < (1.5 * tolerance), actual: 0.0406786 vs 0.000208782
1:  -- Testing 1-D histogram
1: cudaStreamSynchronize(stream) error( cudaErrorIllegalInstruction): an illegal instruction was encountered /ascldap/users/ndellin/kokkos/core/src/Cuda/Kokkos_Cuda_Instance.cpp:172
1: Backtrace:
1:                         [0x55f293]
1:                         [0x5577c8]
1:                         [0x55782b]
1:                         [0x5676fe]
1:                         [0x568271]
1:                         [0x4328ca]
1:                         [0x4344d3]
1:                         [0x46884c]
1:                         [0x46abf9]
1:                         [0x53a21a]
1:                         [0x529268]
1:                         [0x52968a]
1:                         [0x529c0d]
1:                         [0x531aea]
1:                         [0x531f78]
1:                         [0x410b50]
1: __libc_start_main [0x7f8e0e9bf555]
1:                         [0x4156ed]
1/1 Test #1: KokkosAlgorithms_UnitTest_RandomAndSort ...Subprocess aborted***Exception:   6.72 sec
```

masterleinad:
`HIP` failing with 
```
[ RUN      ] hip.reduce_device_view_range_policy
4: /var/jenkins/workspace/Kokkos/core/unit_test/TestReductions_DeviceView.hpp:89: Failure
4: Expected: (time1) < (time_fence1), actual: 0.000702964 vs 0.000493095
```
is clearly unrelated (and spurious).

core/src/Cuda/Kokkos_Cuda_Instance.cpp
dalg24: Just prefix with FIXME and I will approve

masterleinad: f6634bcd39e9bc04966ecb5a28ecbf4b22310086

dalg24 (approved):

stanmoore1:
Looks like this didn't make it into 3.7.1?

dalg24:
Correct.

dalg24:
Keeping the warnings to remind you to upgrade to 4.0 when he gets out

PR #5442, opened 2022-09-08
Cuda: Only initialize device-related variables by the singleton

masterleinad:
This should fix the problem @stanmoore1 was observing that the default stream was created before calling `cudaSetDevice`.
More precisely all device-related member variables are made `static` and initialized outside the instance's `initialize` only for the singleton.

### DIFF
--- a/core/src/Cuda/Kokkos_Cuda_Instance.cpp
+++ b/core/src/Cuda/Kokkos_Cuda_Instance.cpp
@@ -719,7 +719,6 @@ int Cuda::impl_is_initialized() {
 }
 
 void Cuda::impl_initialize(InitializationSettings const &settings) {
-  auto &cuda_singleton     = Impl::CudaInternal::singleton();
   const int cuda_device_id = Impl::get_gpu(settings);
   const auto &dev_info     = Impl::CudaInternalDevices::singleton();
 
@@ -832,6 +831,7 @@ void Cuda::impl_initialize(InitializationSettings const &settings) {
   cudaStream_t singleton_stream;
   KOKKOS_IMPL_CUDA_SAFE_CALL(cudaStreamCreate(&singleton_stream));
 
+  auto &cuda_singleton = Impl::CudaInternal::singleton();
   cuda_singleton.initialize(singleton_stream, /*manage*/ true);
 }
 
--- a/core/src/Cuda/Kokkos_Cuda_Instance.cpp
+++ b/core/src/Cuda/Kokkos_Cuda_Instance.cpp
@@ -722,13 +722,10 @@ void Cuda::impl_initialize(InitializationSettings const &settings) {
   auto &cuda_singleton     = Impl::CudaInternal::singleton();
   const int cuda_device_id = Impl::get_gpu(settings);
   const auto &dev_info     = Impl::CudaInternalDevices::singleton();
-  const bool ok_id =
-      0 <= cuda_device_id && cuda_device_id < dev_info.m_cudaDevCount;
 
   // Need device capability 3.0 or better
-  const bool ok_dev =
-      ok_id && (3 <= dev_info.m_cudaProp[cuda_device_id].major &&
-                0 <= dev_info.m_cudaProp[cuda_device_id].minor);
+  const bool ok_dev = 3 <= dev_info.m_cudaProp[cuda_device_id].major &&
+                      0 <= dev_info.m_cudaProp[cuda_device_id].minor;
   if (ok_dev) {
     const struct cudaDeviceProp &cudaProp = dev_info.m_cudaProp[cuda_device_id];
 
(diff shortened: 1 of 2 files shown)


PhilMiller:
Could you correct the typo in the commit message?

masterleinad:
> Could you correct the typo in the commit message?

I rephrased a bit, too.

masterleinad:
We could of course have a smaller (and in my opinion somewhat hackier) fix by setting the device before creating the singleton stream and don't change anything else.

stanmoore1:
@masterleinad I confirm this fixes the crash I was seeing.

crtrott:
Why do you think its hackier? I would envision basically just pulling out the cudaSetDevice and the deviceProp acquisitaion from CudaInternal::initialize to Cuda::impl_initialize. You do that anyway.

core/src/Cuda/Kokkos_Cuda_Instance.cpp
```diff
@@ -291,6 +291,20 @@ const CudaInternalDevices &CudaInternalDevices::singleton() {
...
+std::array<CudaInternal::size_type, 3> CudaInternal::m_maxBlock;
+unsigned CudaInternal::m_maxSharedWords;
+uint32_t CudaInternal::m_maxConcurrency;
+int CudaInternal::m_shmemPerSM;
+int CudaInternal::m_maxShmemPerBlock;
+int CudaInternal::m_maxBlocksPerSM;
+int CudaInternal::m_maxThreadsPerSM;
+int CudaInternal::m_maxThreadsPerBlock;
+cudaDeviceProp CudaInternal::m_deviceProp;
+CudaInternal::size_type CudaInternal::m_scratchUnifiedSupported;
```
dalg24, line 307: Same comment as in #5444 consider declaring these `inline`

core/src/Cuda/Kokkos_Cuda_Instance.cpp
```diff
@@ -835,10 +719,130 @@ int Cuda::impl_is_initialized() {
 }
 
 void Cuda::impl_initialize(InitializationSettings const &settings) {
+  auto &cuda_singleton     = Impl::CudaInternal::singleton();
+  const int cuda_device_id = Impl::get_gpu(settings);
+  const auto &dev_info     = Impl::CudaInternalDevices::singleton();
+  const bool ok_id =
+      0 <= cuda_device_id && cuda_device_id < dev_info.m_cudaDevCount;
```
dalg24, line 726: `device_id` is being validated by Core

dalg24 (changes requested):

core/src/Cuda/Kokkos_Cuda_Instance.cpp
masterleinad: Would you mind elaborating on the syntax you are proposing? 
Are you saying you just want to declare them as, e.g.,  `inline static int m_cudaDev;` and don't change anything else?

dalg24: Initialize them in the body of the class

dalg24: Daniel pointed out to me that this would lead to a bug because it would have different translation units

core/src/Cuda/Kokkos_Cuda_Instance.cpp
```diff
@@ -835,10 +719,120 @@ int Cuda::impl_is_initialized() {
 }
 
 void Cuda::impl_initialize(InitializationSettings const &settings) {
+  auto &cuda_singleton     = Impl::CudaInternal::singleton();
```
dalg24, line 722: Move it closer to its use.  We should avoid constructing instances (singleton comprised) before everything else is ready.

dalg24 (approved):

masterleinad:
Only `CUDA-11.0-NVCC-C++17-RDC` is failing with
```
25: Time MDRangePolicy: NonOverlap: 0.004693 Time Overlap: 0.002397
25: /var/jenkins/workspace/Kokkos/core/perf_test/PerfTest_ExecSpacePartitioning.cpp:475: Failure
25: Expected: (time_overlapped_reduce) < (1.5 * time_no_overlapped_reduce), actual: 0.0015471 vs 0.00131185
25: [  FAILED  ] default_exec.overlap_mdrange_policy (840 ms)
```
That looks unrelated/spurious.

JBludau (approved):

stanmoore1:
ExaMiniMD GPU regression tests are passing again, thanks!

core/src/Cuda/Kokkos_Cuda_Instance.cpp
masterleinad: Even then, I would only use `inline static` if I can provide a value that won't change (but the type doesn't allow `constexpr`).

PR #5443, opened 2022-09-08
unit_tests: disable Abort and ViewMemoryAccessViolation with Trilinos

ndellingwood:
workaround deathtest issues in MPI builds on Power9 systems running tests through jsrun


masterleinad (approved):

PhilMiller:
Do we have a bigger picture plan to either make these work, or narrow down the exclusion, so that we don't end up in a situation where for some reason, they're actually failing across the board in Trilinos and we don't notice?

dalg24:
Not worth our time.  Debugging them is a time sink and these death tests a not critical, they are more there to check the quality of implementation (checking that error message gets printed to the output stream before abnormal termination of the program)

Issue #5432, opened 2022-09-07
Nightly build failures on develop with intel/19, "missing return statement at end of non-void function"


ndellingwood:
**Describe the bug**
Nightly builds with intel/19 (all backends) fail to compile with output:
```
14:28:58 /home/jenkins/node/workspace/kokkos_osrhel7_intel_1905_openmp_pthread_serial/kokkos/core/src/impl/Kokkos_ViewCtor.hpp(289): error #1011: missing return statement at end of non-void function "Kokkos::Impl::with_properties_if_unset(const Kokkos::Impl::ViewCtorProp<P...> &, const Property &, const Properties &...) [with P=<std::__cxx11::string>, Property=std::__cxx11::string, Properties=<>]"
14:28:58   }
14:28:58   ^
14:28:58           detected during:
14:28:58             instantiation of "auto Kokkos::Impl::with_properties_if_unset(const Kokkos::Impl::ViewCtorProp<P...> &, const Property &, const Properties &...) [with P=<std::__cxx11::string>, Property=std::__cxx11::string, Properties=<>]" at line 1408 of "/home/jenkins/node/workspace/kokkos_osrhel7_intel_1905_openmp_pthread_serial/kokkos/core/src/Kokkos_View.hpp"
14:28:58             instantiation of "Kokkos::View<DataType, Properties...>::View(const Kokkos::Impl::ViewCtorProp<P...> &, const std::enable_if_t<<expression>, Kokkos::ViewTraits<DataType, Properties...>::array_layout> &, Kokkos::View<DataType, Properties...>::check_input_args) [with DataType=int *, Properties=<exec_space>, P=<std::__cxx11::string>]" at line 1567 of "/home/jenkins/node/workspace/kokkos_osrhel7_intel_1905_openmp_pthread_serial/kokkos/core/src/Kokkos_View.hpp"
14:28:58             instantiation of "Kokkos::View<DataType, Properties...>::View(const Label &, std::enable_if_t<Kokkos::Impl::is_view_label<Label>::value, const size_t={unsigned long}>, size_t={unsigned long}, size_t={unsigned long}, size_t={unsigned long}, size_t={unsigned long}, size_t={unsigned long}, size_t={unsigned long}, size_t={unsigned long}) [with DataType=int *, Properties=<exec_space>, Label=char [6]]" at line 25 of
[...]

PR #5447, opened 2022-09-09
test_all_sandia: Add diagnostic suppression flags for intel builds

ndellingwood:
Add flags -diag-disable=1011,-diag-disable=869 for intel builds Workaround kokkos/kokkos#5432 for nightly builds


ndellingwood:
CI failures unrelated to changes in this PR (only changes to nightly testing script, unused in Kokkos and unused by CI)

dalg24 (approved):

cz4rs (approved):

masterleinad (commented):
Wouldn't adding the code at https://github.com/kokkos/kokkos/blob/4477a25ebe12b655cd5da273eec4ab954fbf32d5/core/src/impl/Kokkos_ViewCtor.hpp#L280-L284 for warning 869 be sufficient, too?

ndellingwood:
> Wouldn't adding the code at
> 
> https://github.com/kokkos/kokkos/blob/4477a25ebe12b655cd5da273eec4ab954fbf32d5/core/src/impl/Kokkos_ViewCtor.hpp#L280-L284
> for warning 869 be sufficient, too?

No, the current push/pop for 1011 is either not recognized or not respected by the compiler; adding disable for warning 869 does not have an impact on the -Werror

ndellingwood:
@dalg24 @crtrott this should be ready for merge

crtrott:
is maybe the push in the code for the other warning number just missing. I think we really need to try figure out how to fix that, because otherwise downstream codes will run into the warning.

dalg24:
> is maybe the push in the code for the other warning number just missing. I think we really need to try figure out how to fix that, because otherwise downstream codes will run into the warning.

Only the ones that do "inline" builds.  You would not see them when following the recommended `find_package(Kokkos)` and `target_link_libraries(MyExeOrLib PUBLIC Kokkos::kokkos)`

ndellingwood:
PR unnecessary, underlying -Werror were resolved within the source code

Issue #4861, opened 2022-03-10
HIP backend: LocalMemory launch mechanism should pass functor by value


brian-kelley:
I noticed that in the HIP backend, launches using the LocalMemory launch mechanism actually seem to take the functor/driver as a pointer to host-pinned, not by value like on Cuda:

Kokkos_HIP_KernelLaunch.hpp:
```
101 template <class DriverType>
102 __global__ static void hip_parallel_launch_local_memory(
103     const DriverType *driver) {
104   // FIXME_HIP driver() pass by copy
105   driver->operator()();
106 }
107 
108 template <class DriverType, unsigned int maxTperB, unsigned int minBperSM>
109 __global__ __launch_bounds__(
110     maxTperB,
111     minBperSM) static void hip_parallel_launch_local_memory(const DriverType
112                                                                 *driver) {
113   // FIXME_HIP driver() pass by copy
114   driver->operator()();
115 }
```

Is this something that can be resolved easily? I would imagine it would give a noticeable performance boost, since if I'm understanding right, launches have to cross PCIe at least 3 times (?)

- Host launches kernel and sends driver pointer to device
- Device reads the driver out of host pinned
- Driver comes back from host pinned

whereas if it were pass-by-value, it would just be "host launches with the full driver".

I tried to hack this change in, to see how performance changed, but I ran into a couple issues:
- The HIP specializations of some drivers have ``std::lock_guards`` so the drivers can't be copy-constructed (e.g. ``Impl::ParallelReduce`` has a lock ``m_shared_memory_lock``). This is different from Cuda, where the drivers don't acquire locks
- When I change the ``__global__ hip_parallel_launch_local_memory(...)`` function to take the driver by value, hipcc seems to think it's an ambiguous overload down below in ``get_kernel_func()``. This happens even if I give different names to the versions with and without the ``__launch_bounds__``.

Rombur:
The reason it is done that way is that sometimes you would hit a compiler bug at runtime. I don't remember the details maybe @arghdos  does. We can change the code but we will need to tests with several apps to check that the bug has been fixed.

arghdos:
I believe this should be fixed at this point (this was... way back in like ROCm 3.5 days), but I've never gone back to check it, but it used to cause the compiler to behave _terribly_ (think: T-Rex on a unicycle trying to hold a glass of water... spills everywhere :P)

[...]

PR #5448, opened 2022-09-09
HIP: Pass functor by value when using LocalMemory launch mechanism

Rombur:
In the past, we had to pass the functor by reference because of compiler bugs. These have been fixed. To ensure that multiple teams did not use the same scratch memory, the HIP backend was using a lock. I have replaced the lock with a memory pool (this is what is done in CUDA). As noticed https://github.com/kokkos/kokkos/issues/4861#issuecomment-1228924111, the class `ParallelScanHIPBase` cannot have a pure virtual function. Using a dummy implementation fixes the problem.

Related to #4861

### DIFF
--- a/core/src/HIP/Kokkos_HIP_Instance.cpp
+++ b/core/src/HIP/Kokkos_HIP_Instance.cpp
@@ -237,10 +237,6 @@ void HIPInternal::initialize(int hip_device_id, hipStream_t stream,
 
     m_stream        = stream;
     m_manage_stream = manage_stream;
-    for (int i = 0; i < m_n_team_scratch; ++i) {
-      m_team_scratch_current_size[i] = 0;
-      m_team_scratch_ptr[i]          = nullptr;
-    }
 
     // number of multiprocessors
     m_multiProcCount = hipProp.multiProcessorCount;
--- a/core/src/HIP/Kokkos_HIP_Instance.cpp
+++ b/core/src/HIP/Kokkos_HIP_Instance.cpp
@@ -235,10 +235,12 @@ void HIPInternal::initialize(int hip_device_id, hipStream_t stream,
 
     KOKKOS_IMPL_HIP_SAFE_CALL(hipSetDevice(m_hipDev));
 
-    m_stream                    = stream;
-    m_manage_stream             = manage_stream;
-    m_team_scratch_current_size = 0;
-    m_team_scratch_ptr          = nullptr;
+    m_stream        = stream;
+    m_manage_stream = manage_stream;
+    for (int i = 0; i < m_n_team_scratch; ++i) {
+      m_team_scratch_current_size[i] = 0;
+      m_team_scratch_ptr[i]          = nullptr;
+    }
 
     // number of multiprocessors
     m_multiProcCount = hipProp.multiProcessorCount;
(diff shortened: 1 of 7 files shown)


core/src/HIP/Kokkos_HIP_Parallel_Range.hpp
```diff
@@ -614,7 +602,10 @@ class ParallelScanHIPBase {
   }
 
   // Determine block size constrained by shared memory:
-  virtual inline unsigned local_block_size(const FunctorType& f) = 0;
+  virtual inline unsigned local_block_size(const FunctorType&) {
+    // Making this function pure virtual, trips the compiler.
+    return 0;
```
JBludau, line 607: hmm... as we can not make it pure, can we at least throw an `exception` or an `assert` ... just for sanity ... should never be reached anyways, but if it does we would see it

core/src/HIP/Kokkos_HIP_Instance.hpp
```diff
@@ -130,9 +116,14 @@ class HIPInternal {
...
-  mutable void *m_team_scratch_ptr            = nullptr;
-  mutable std::mutex m_team_scratch_mutex;
+  int m_n_team_scratch                            = 10;
+  mutable int64_t m_team_scratch_current_size[10] = {0, 0, 0, 0, 0,
+                                                     0, 0, 0, 0, 0};
+  mutable void *m_team_scratch_ptr[10] = {nullptr, nullptr, nullptr, nullptr,
+                                          nullptr, nullptr, nullptr, nullptr,
+                                          nullptr, nullptr};
+  mutable std::atomic_int m_team_scratch_pool[10] = {0, 0, 0, 0, 0,
+                                                     0, 0, 0, 0, 0};
```
dalg24, line 126:
```suggestion
  mutable int64_t m_team_scratch_current_size[10] = {};
  mutable void *m_team_scratch_ptr[10]            = {};
  mutable std::atomic_int m_team_scratch_pool[10] = {};
```

core/src/HIP/Kokkos_HIP_Parallel_Range.hpp
masterleinad: I  would just get rid of this function and pass the block size directly to the base class.

Rombur: we already throw if the block size is zero

core/src/HIP/Kokkos_HIP_Parallel_Range.hpp
```diff
@@ -614,7 +602,10 @@ class ParallelScanHIPBase {
   }
 
   // Determine block size constrained by shared memory:
-  virtual inline unsigned local_block_size(const FunctorType& f) = 0;
+  virtual inline unsigned local_block_size(const FunctorType&) {
+    // Making this function pure virtual, trips the compiler.
```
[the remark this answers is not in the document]
Rombur: That's we already throw if the size is zero

core/src/HIP/Kokkos_HIP_Instance.cpp
```diff
@@ -362,20 +356,37 @@ Kokkos::HIP::size_type *HIPInternal::scratch_flags(const std::size_t size) {
...
-        "Kokkos::HIPSpace::TeamScratchMemory", m_team_scratch_current_size);
+std::pair<void *, int> HIPInternal::resize_team_scratch_space(
+    std::int64_t bytes, bool force_shrink) {
+  // Multiple ParallelFor/Reduce Teams can call this function at the same time
+  // and invalidate the m_team_scratch_ptr. We use a pool to avoid any race
+  // condition.
+
+  int current_team_scratch = 0;
+  int zero                 = 0;
+  int one                  = 1;
```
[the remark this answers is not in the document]
Rombur: I need `zero` to be a variable. I could use `1` instead of `one`,  if you prefer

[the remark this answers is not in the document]
dalg24: Oh right.  Yes please one -> 1

core/src/HIP/Kokkos_HIP_Instance.cpp
```diff
@@ -362,20 +356,40 @@ Kokkos::HIP::size_type *HIPInternal::scratch_flags(const std::size_t size) {
...
-      ((bytes < m_team_scratch_current_size) && (force_shrink))) {
-    m_team_scratch_current_size = bytes;
-    m_team_scratch_ptr          = Kokkos::kokkos_realloc<Kokkos::HIPSpace>(
-        m_team_scratch_ptr, m_team_scratch_current_size);
+  if (m_team_scratch_current_size[current_team_scratch] == 0) {
+    m_team_scratch_current_size[current_team_scratch] = bytes;
+    m_team_scratch_ptr[current_team_scratch] =
+        Kokkos::kokkos_malloc<Kokkos::HIPSpace>(
+            "Kokkos::HIPSpace::TeamScratchMemory",
+            m_team_scratch_current_size[current_team_scratch]);
```
JBludau, line 376: would it make sense to move the assignment of the size to after the allocation was done? may be nitpicking

core/src/HIP/Kokkos_HIP_Instance.cpp
```diff
@@ -396,68 +410,34 @@ void HIPInternal::finalize() {
     RecordHIP::decrement(RecordHIP::get_record(m_scratchFlags));
     RecordHIP::decrement(RecordHIP::get_record(m_scratchSpace));
 
-    if (m_team_scratch_current_size > 0)
-      Kokkos::kokkos_free<Kokkos::HIPSpace>(m_team_scratch_ptr);
+    for (int i = 0; i < m_n_team_scratch; ++i) {
+      if (m_team_scratch_current_size[i] > 0)
+        Kokkos::kokkos_free<Kokkos::HIPSpace>(m_team_scratch_ptr[i]);
+    }
```
JBludau, line 416: note sure if we should reset the size to 0 here, too

Rombur: It's done below.

Rombur:
Retest this please

core/src/HIP/Kokkos_HIP_Instance.hpp
```diff
@@ -130,9 +116,10 @@ class HIPInternal {
   bool m_manage_stream = false;
 
   // Team Scratch Level 1 Space
-  mutable int64_t m_team_scratch_current_size = 0;
-  mutable void *m_team_scratch_ptr            = nullptr;
-  mutable std::mutex m_team_scratch_mutex;
+  int m_n_team_scratch                            = 10;
```
ldh4, line 119: Is this variable set to 10 arbitrarily or is it derived from something?

dalg24 (approved):
Looks good.  Please consider the changes I suggested (follow up PR is fine)

Rombur:
> Looks good. Please consider the changes I suggested (follow up PR is fine)

I'll' do it in a follow up PR. I'll change the CUDA code at the same time.

PR #5450, opened 2022-09-09
Move `Kokkos::reduction_identity` into its own header file

dalg24:
Move `reduction_identity` from `<Kokkos_NumericTraits.hpp>` to a new (public?) header `<Kokkos_ReductionIdentity.hpp>`

Admittedly this may break code downstream :/


JBludau (approved):
looks good.

masterleinad (approved):
Let's see if someone complains. In the end, `reduction_identity` should only be relevant if people call `parallel_reduce` but, of course, they might just have included this header for specializing `reduction_identity`. Anyway, if we merge early, we can hope that users complain early and it would be easy to revert.

dalg24:
> Let's see if someone complains. In the end, `reduction_identity` should only be relevant if people call `parallel_reduce` but, of course, they might just have included this header for specializing `reduction_identity`. Anyway, if we merge early, we can hope that users complain early and it would be easy to revert.

I was not clear enough.  We **know** that it will break code downstream (for instance KokkosKernels and ArborX).  If we decide to merge it, we are not doing it to see if people complain loud enough and revert, we expect them to update their code.
The main question is do we agree that it is the right thing to do.

masterleinad:
If this breaks too many projects, we could, of course, include `ReductionIdentity.hpp` in `Kokkos_NumericTraits.hpp` again.

dalg24:
In Trilinos
https://github.com/trilinos/Trilinos/blob/597f531b9be215f69c06b58c97bbd6a2bd0ccddf/packages/stokhos/src/sacado/kokkos/pce/Sacado_UQ_PCE.hpp#L1315-L1336
https://github.com/trilinos/Trilinos/blob/597f531b9be215f69c06b58c97bbd6a2bd0ccddf/packages/stokhos/src/sacado/kokkos/vector/Sacado_MP_Vector.hpp#L2063-L2084
would need to be updated.  That's all I could find.

Rombur:
I think it's a good change but we should deprecate the current code since we know it will break user code.

dalg24:
> I think it's a good change but we should deprecate the current code since we know it will break user code.

OK as a compromise what if I include the new `<Kokkos_ReductionIdentity.hpp>` header from `<Kokkos_NumericTraits.hpp>` (96a9c76) when `Kokkos_ENABLE_DEPRECATED_CODE_4` is `ON`

Rombur (approved):

PR #5452, opened 2022-09-12
Don't require user-defined volatile operator= in Kokkos::single for aggregates

cwpearson:
Prior to this commit, users would have to define the following overload to pass an aggregate T to `Kokkos::single`.
* `KOKKOS_INLINE_FUNCTION void operator=(const volatile T &rhs) volatile {...}`

This was annoying because it has to have a void return type to prevent g++ from issuing a warning with the following usage:

```c++
T a, b;
a = b; // here
```

g++ will issue a warning if the return type was the expected `T&` because `a=b` will have a type of `T&`, which results in a spurious memory access if T is volatile (in g++).

it is also annoying because if the user wants to use T in any usual value-type ways, they will have to follow the rule of 7 now that they have defined one operator.

### DIFF
--- a/core/src/impl/Kokkos_HostThreadTeam.hpp
+++ b/core/src/impl/Kokkos_HostThreadTeam.hpp
@@ -498,7 +498,7 @@ class HostThreadTeamMember {
                                              const int source_team_rank) const
       noexcept {
     KOKKOS_IF_ON_HOST((if (1 < m_data.m_team_size) {
-      T volatile* const shared_value = (T*)m_data.team_reduce();
+      T* const shared_value = (T*)m_data.team_reduce();
 
       // Don't overwrite shared memory until all threads arrive
 
@@ -507,13 +507,15 @@ class HostThreadTeamMember {
         // only this thread returned from 'team_rendezvous'
         // with a return value of 'true'
 
-        *shared_value = value;
+        Kokkos::Impl::atomic_store(shared_value, value,
+                                   Kokkos::Impl::memory_order_release);
 
         m_data.team_rendezvous_release();
         // This thread released all other threads from 'team_rendezvous'
         // with a return value of 'false'
       } else {
-        value = *shared_value;
+        value = Kokkos::Impl::atomic_load(shared_value,
+                                          Kokkos::Impl::memory_order_acquire);
       }
     }))
 
@@ -528,7 +530,7 @@ class HostThreadTeamMember {
                                              const int source_team_rank) const
       noexcept {
     KOKKOS_IF_ON_HOST((
-        T volatile* const shared_value = (T*)m_data.team_reduce();
+        T* const shared_value = (T*)m_data.team_reduce();
 
         // Don't overwrite shared memory until all threads arrive
 
@@ -540,13 +542,17 @@ class HostThreadTeamMember {
           f(value);
 
           if (1 < m_data.m_team_size) {
-            *shared_value = value;
+            Kokkos::Impl::atomic_store(shared_value, value,
+                                       Kokkos::Impl::memory_order_release);
           }
 
           m_data.team_rendezvous_release();
           // This thread released all other threads from 'team_rendezvous'
           // with a return value of 'false'
-        } else { value = *shared_value; }))
+        } else {
+          value = Kokkos::Impl::atomic_load(shared_value,
+                                            Kokkos::Impl::memory_order_acquire);
+        }))
 
     KOKKOS_IF_ON_DEVICE(
         ((void)f; (void)value; (void)source_team_rank;
--- a/core/unit_test/TestTeamBasic.hpp
+++ b/core/unit_test/TestTeamBasic.hpp
@@ -181,7 +181,8 @@ struct TestTeamForAggregate {
         Kokkos::parallel_for(team_policy_t(leagueSize, teamSize), test);
         int numErrs = 0;
         Kokkos::parallel_reduce(range_policy_t(0, problemSize), test, numErrs);
-        EXPECT_EQ(numErrs, 0);
+        EXPECT_EQ(numErrs, 0)
+            << " teamSize=" << teamSize << " problemSize=" << problemSize;
       }
     }
   }
--- a/core/unit_test/TestTeamBasic.hpp
+++ b/core/unit_test/TestTeamBasic.hpp
@@ -114,6 +114,83 @@ TEST(TEST_CATEGORY, team_reduce_large) {
   }
 }
 
+/*! \brief Test passing an aggregate to Kokkos::single in a parallel_for with
+           team policy
+*/
+template <typename ExecutionSpace>
+struct TestTeamForAggregate {
+  using range_policy_t = Kokkos::RangePolicy<ExecutionSpace>;
+  using team_policy_t  = Kokkos::TeamPolicy<ExecutionSpace>;
+  using member_t       = typename team_policy_t::member_type;
+  using memory_space   = typename ExecutionSpace::memory_space;
+  using results_type   = Kokkos::View<double*, memory_space>;
+
+  static constexpr double INIT_VALUE   = -1.0;
+  static constexpr double EXPECT_VALUE = 1.0;
+
+  struct Agg {
+    double d;
+  };
+  results_type results_;
+
+  TestTeamForAggregate(const size_t size) : results_("results", size) {}
+  TestTeamForAggregate() : TestTeamForAggregate(0) {}
+
+  KOKKOS_INLINE_FUNCTION
+  void operator()(const member_t& t) const {
+    Agg lagg;
+    lagg.d = INIT_VALUE;
+    Kokkos::single(
+        Kokkos::PerTeam(t), [&](Agg& myAgg) { myAgg.d = EXPECT_VALUE; }, lagg);
+    size_t i = t.league_rank() * t.team_size() + t.team_rank();
+    if (i < results_.size()) {
+      results_(i) = lagg.d;
+    }
+  }
+
+  KOKKOS_INLINE_FUNCTION
+  void operator()(const int i, int& lNumErrs) const {
+    if (EXPECT_VALUE != results_(i)) {
+      ++lNumErrs;
+    }
+  }
+
+  static void run() {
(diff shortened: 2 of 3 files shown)


PhilMiller:
Is there testing covering this code? There really ought to be. Ideally, you would run it for many repetitions to see that it never fails, to try to detect any rare race condition

cwpearson:
> Is there testing covering this code? There really ought to be. Ideally, you would run it for many repetitions to see that it never fails, to try to detect any rare race condition

@PhilMiller I don't believe there was, so I added one. How would you like to see it modified to (hopefully) catch race conditions? Then I will squash and format the commit.

PhilMiller:
Looks like the test is actually showing failures in CI, e.g. on fedora Release OpenMP

Separately, the corresponding DEBUG build is failing due to a syntax quirk. If you're declaring a variable to be default constructed, write either `Foo myFoo;` or `Foo myFoo{};` and not `Foo myFoo()`, which gets parsed as a function declaration rather than a variable declaration.

PhilMiller:
Also, per recent discussion in last week's meeting, I note that the Programming Guide describes `single`, but the API reference only illustrates its usage in examples on the `Team{Thread,Vector}Range` and `ThreadVectorRange` pages. Please open an issue reflecting the missing reference page, and consider addressing it.

PhilMiller (approved):

cwpearson:
Am I understanding correctly that
* we like that `single` is shown in the e.g. [TeamThreadRange](https://kokkos.github.io/kokkos-core-wiki/API/core/policies/TeamThreadRange.html) docs
* we'd like a standalone reference for it, like the many listed on the [API Core](https://kokkos.github.io/kokkos-core-wiki/API/core-index.html) page?

PhilMiller:
> Am I understanding correctly that
> 
> * we like that `single` is shown in the e.g. [TeamThreadRange](https://kokkos.github.io/kokkos-core-wiki/API/core/policies/TeamThreadRange.html) docs
> * we'd like a standalone reference for it, like the many listed on the [API Core](https://kokkos.github.io/kokkos-core-wiki/API/core-index.html) page?

That's my thinking, yeah

cwpearson:
It's not obvious to me what is wrong with the failing test above.

core/unit_test/TestTeamBasic.hpp
```diff
@@ -114,6 +114,75 @@ TEST(TEST_CATEGORY, team_reduce_large) {
...
+  static void run() {
+    // dummy objects to determine max team size
+    int maxTeamSize;
+    {
+      TestTeamForAggregate test;
+      maxTeamSize =
+          team_policy_t(1, 1).team_size_max(test, Kokkos::ParallelForTag());
+    }
+
+    for (int teamSize = 1; teamSize <= maxTeamSize; teamSize *= 2) {
```
masterleinad, line 167: OpenMPTarget has
```
 RUN      ] openmptarget.team_parallel_single
4: OpenMPTarget backend requires a minimum of 32 threads per team.
4: .
```
so you would need to start this loop with 32 for that backend.

cwpearson: https://github.com/kokkos/kokkos/issues/5458

cwpearson:
Here's an issue to track the `Kokkos::single` documentation: https://github.com/kokkos/kokkos/issues/5459

cwpearson:
Looks like there was an infrastructure failure of some kind in the test - can this be marked for retest once that is resolved?

masterleinad:
> Looks like there was an infrastructure failure of some kind in the test - can this be marked for retest once that is resolved?

No, https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/Kokkos/detail/Kokkos/10560/pipeline is failing with
```Building CXX object core/unit_test/CMakeFiles/KokkosCore_UnitTest_OpenMPTarget.dir/openmptarget/TestOpenMPTarget_TeamBasic.cpp.o
In file included from /var/jenkins/workspace/Kokkos/build/core/unit_test/openmptarget/TestOpenMPTarget_TeamBasic.cpp:2:
/var/jenkins/workspace/Kokkos/core/unit_test/TestTeamBasic.hpp:168:58: error: no member named 'OpenMPTarget' in namespace 'Kokkos'
      if constexpr (std::is_same<ExecutionSpace, Kokkos::OpenMPTarget>::value) {
                                                 ~~~~~~~~^
```
It should be `Kokkos::Experimental::OpenMPTarget` instead of `Kokkos::OpenMPTarget`.

core/unit_test/TestTeamBasic.hpp
```diff
@@ -114,6 +114,85 @@ TEST(TEST_CATEGORY, team_reduce_large) {
...
+
+  static void run() {
+    // dummy objects to determine valid team sizes
+    int minTeamSize, maxTeamSize;
+    {
+      TestTeamForAggregate test;
+      minTeamSize = 1;
+      /* OpenMPTarget hard-codes 32 as the minimum size
+         if TeamPolicy::min_team_size() is ever provided, use that instead
+      */
```
PhilMiller, line 166: This should probably have FIXME OPENMPTARGET in the comment, so that we know to clear it out when the backend limitation is lifted

core/unit_test/TestTeamBasic.hpp
```diff
@@ -114,6 +114,85 @@ TEST(TEST_CATEGORY, team_reduce_large) {
...
+  void operator()(const int i, int& lNumErrs) const {
+    if (EXPECT_VALUE != results_(i)) {
+      ++lNumErrs;
+    }
+  }
+
+  static void run() {
+    // dummy objects to determine valid team sizes
+    int minTeamSize, maxTeamSize;
+    {
```
dalg24, line 161: Why do you add a scope here?
```suggestion
```

core/unit_test/TestTeamBasic.hpp
```diff
@@ -114,6 +114,85 @@ TEST(TEST_CATEGORY, team_reduce_large) {
...
+        minTeamSize = 32;
+      }
+#endif
+
+      maxTeamSize =
+          team_policy_t(1, 1).team_size_max(test, Kokkos::ParallelForTag());
+    }
+
+    for (int teamSize = minTeamSize; teamSize <= maxTeamSize; teamSize *= 2) {
+      for (int problemSize : {1, 100, 100 * 100, 100 * 100 * 100}) {
```
dalg24, line 178:
```suggestion
      for (int problemSize : {1, 100, 10'000, 1'000'000}) {
```
more readable IMO

core/unit_test/TestTeamBasic.hpp
```diff
@@ -114,6 +114,85 @@ TEST(TEST_CATEGORY, team_reduce_large) {
...
+        const int leagueSize = (problemSize + teamSize - 1) / teamSize;
+        TestTeamForAggregate::results_type results("results", problemSize);
+        TestTeamForAggregate test(results);
```
dalg24, line 181: Why do you have both `test` and `result`?

core/unit_test/TestTeamBasic.hpp
cwpearson: The scope is to let any temporary objects solely used to set `minTeamSize` and `maxTeamSize` go away, but I can rework this to be a bit cleaner.

core/unit_test/TestTeamBasic.hpp
dalg24: Yes.  I somehow had read the code wrong.  I feel like it is a bit silly to expose the `results` view outside the functor.

core/src/impl/Kokkos_HostThreadTeam.hpp
```diff
@@ -498,7 +498,7 @@ class HostThreadTeamMember {
                                              const int source_team_rank) const
       noexcept {
     KOKKOS_IF_ON_HOST((if (1 < m_data.m_team_size) {
-      T volatile* const shared_value = (T*)m_data.team_reduce();
```
masterleinad, line 501: In the end, `volatile` was only ever relevant for `Cuda` but this code is never executed in `Cuda` device code. Thus, I would expect that the unit tests would equally pass if we don't use atomics at all. In fact, it seems that `HPX` and `Threads` have their own implementation such that this one would only be used for `OpenMP` and `Serial`.

I agree that using atomics here makes sense, though.

masterleinad (approved):

core/src/impl/Kokkos_HostThreadTeam.hpp
PhilMiller: If this gets compiled by MSVC, that largely treats `volatile T` as `atomic<T>`

masterleinad:
Only `default_exec.overlap_mdrange_policy` is failing for two `Cuda` builds which looks unrelated/spurious.

cwpearson:
^ squash and rebase on develop

core/unit_test/TestTeamBasic.hpp
```diff
@@ -114,6 +114,83 @@ TEST(TEST_CATEGORY, team_reduce_large) {
...
+    }
+
+    for (int teamSize = minTeamSize; teamSize <= maxTeamSize; teamSize *= 2) {
+      for (int problemSize : {1, 100, 10'000, 1'000'000}) {
+        const int leagueSize = (problemSize + teamSize - 1) / teamSize;
+        TestTeamForAggregate test(problemSize);
+        Kokkos::parallel_for(team_policy_t(leagueSize, teamSize), test);
+        int numErrs = 0;
+        Kokkos::parallel_reduce(range_policy_t(0, problemSize), test, numErrs);
+        EXPECT_EQ(numErrs, 0);
```
dalg24, line 184: Consider printing `teamSize` and `problemSize`
```suggestion
        EXPECT_EQ(numErrs, 0) << " teamSize=" << teamSize << " problemSize=" << problemSize;
```

crtrott (approved):
I committed Damiens suggestions because I wanted it to be in. Lets see if the formatting check is happy.

crtrott:
Fixed the clang-format

dalg24:
Ignoring failures

PR #5455, opened 2022-09-13
Try removing volatile from AtomicDataElement

masterleinad:
In the last hackathon I was mentoring, my team wanted to use `MemorytTraits:::Atomic` with a custom type. This currently requires defining all kinds of `volatile` qualified member functions which is not very user-friendly.

It appears that none of the `volatile` overloads in `AtomicDataElement` are used anymore, though, and there doesn't seem to be a point in combing `volatile` with `atomic`.
Also, `Kokkos_Atomic_is_only_allowed_with_32bit_and_64bit_scalars` is not used at all.

### DIFF
--- a/core/src/impl/Kokkos_Atomic_View.hpp
+++ b/core/src/impl/Kokkos_Atomic_View.hpp
@@ -248,19 +248,6 @@ class AtomicViewDataHandle {
   operator typename ViewTraits::value_type*() const { return ptr; }
 };
 
-template <unsigned Size>
-struct Kokkos_Atomic_is_only_allowed_with_32bit_and_64bit_scalars;
-
-template <>
-struct Kokkos_Atomic_is_only_allowed_with_32bit_and_64bit_scalars<4> {
-  using type = int;
-};
-
-template <>
-struct Kokkos_Atomic_is_only_allowed_with_32bit_and_64bit_scalars<8> {
-  using type = int64_t;
-};
-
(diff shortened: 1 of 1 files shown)


core/src/impl/Kokkos_Atomic_View.hpp
```diff
@@ -284,63 +197,33 @@ class AtomicDataElement {
...
-  operator const_value_type() const {
-    // return Kokkos::atomic_load(ptr);
-    return *ptr;
-  }
 
   KOKKOS_INLINE_FUNCTION
-  operator non_const_value_type() volatile const {
-    return Kokkos::Impl::atomic_load(ptr);
+  operator value_type() const {
+    return Kokkos::Impl::atomic_load(ptr, Kokkos::Impl::memory_order_relaxed);
```
masterleinad, line 226: This and `operator=` might be the only significant changes.

masterleinad:
> Is that UniqueToken failure in OpenMPTarget a thing which occasionally goes wrong because of a pre-existing bug, or is there something real there changing due to this PR?

We have seen that failing before. That test doesn't even use ` Kokkos::MemoryTraits<Kokkos::Atomic>` AFAICT.

crtrott (approved):

crtrott:
I think this is good.

dalg24:
Ignoring the OpenMPTarget failure

PR #5462, opened 2022-09-15
Provide configure error if Kokkos_ENABLE_UNSUPPORTED_ARCHS is not set when compiling for SYCL+NVIDIA GPUs

masterleinad:
If `KOKKOS_SHOW_CUDA_ARCHS` is `OFF` trying to set any CUDA architecture is simply ignored. This is problematic for SYCL since we want to force the user to use Kokkos_ENABLE_UNSUPPORTED_ARCHS with it. Displaying the CUDA architectures always when SYCL is enabled seems to be a good compromise to me.


JBludau:
Does `SYCL` have any restrictions on which archs it supports? might be worth considering only showing the working ones ... same for `OpenMPTarget` and `OpenACC`

masterleinad:
> Does `SYCL` have any restrictions on which archs it supports? might be worth considering only showing the working ones ... same for `OpenMPTarget` and `OpenACC`

No, SYCL itself doesn't have restrictions (unknown what works) but we said that we decided to hide support for non-Intel GPU backends behind a `Kokkos_ENABLE_UNSUPPORTED_ARCHS` switch.

dalg24 (commented):
Please clarify how you expect users to compile and what happens with the current Kokkos develop/with the changes proposed.
Why does this need to be updated and not L90 or L107

masterleinad:
> Please clarify how you expect users to compile and what happens with the current Kokkos develop/with the changes proposed.
> Why does this need to be updated and not L90 or L107

Nothing changed. Users you need to set `Kokkos_ENABLE_UNSUPPORTED_ARCHS=ON` to be able to set a `CUDA` architecture when using `SYCL`. We just accidentally disabled the configuration error when that option wasn't set (but silently ignored the CUDA architecture).
Let me update the other locations as well for consistency.

masterleinad:
The difference is that SYCL can only be used with Intel and CUDA architectures and currently `Kokkos_ENABLE_UNSUPPORTED_ARCHS` is only used for `SYCL`. The problem fixed in this pull request suggests that it's not a good idea to avoid displaying architectures based on that flag, though.

dalg24:
I though the intent was you must configure with `-DKokkos_ENABLE_SYCL=ON -DKokkos_ENABLE_AMPERE80=ON -D Kokkos_ENABLE_UNSUPPORTED_ARCHS=ON` to run say on A100.
Why should we change what as far as I remember was a conscious design decision?

masterleinad:
> Why should we change what as far as I remember was a conscious design decision?

It doesn't change the validation is done in https://github.com/kokkos/kokkos/blob/d19aab9981a2c447e832a7b4eb7b16992328fb14/cmake/kokkos_arch.cmake#L556-L564.

dalg24:
I still not quite get what issue is addressed here.
* `-DKokkos_ENABLE_SYCL=ON -DKokkos_ENABLE_AMPERE80=ON -D Kokkos_ENABLE_UNSUPPORTED_ARCHS=ON` works as expected
* `-DKokkos_ENABLE_SYCL=ON -DKokkos_ENABLE_AMPERE80=ON` warns the user with something like
```
CMake Warning:

    Kokkos_ENABLE_AMPERE80
```

What are you trying to achieve?  Turn the CMake standard warning about unused variable into a Kokkos configuration error?
Is your point that https://github.com/kokkos/kokkos/blob/d19aab9981a2c447e832a7b4eb7b16992328fb14/cmake/kokkos_arch.cmake#L563 is unreachable?

(PS it is on the PR author to lay things out, not for the reviewer to figure out what is going on)

masterleinad:
```
CMake Warning:

    Kokkos_ENABLE_AMPERE80
```
is not enough for people to see that the option was ignored. With this pull request, I get
```
CMake Error at cmake/kokkos_arch.cmake:682 (MESSAGE):
  Setting a CUDA architecture for SYCL is only allowed with
  Kokkos_ENABLE_UNSUPPORTED_ARCHS=ON!
  cmake/kokkos_tribits.cmake:244 (INCLUDE)
  CMakeLists.txt:194 (KOKKOS_SETUP_BUILD_ENVIRONMENT)
```
which was the intention of implementing the error.

crtrott:
I changed the title does this look good?

masterleinad:
> I changed the title does this look good?

Being pedantic, it's something that worked before we tried hiding architectures so having "fix" in there made sense to me. Otherwise, this is fine.

dalg24 (approved):

crtrott (approved):

PR #5466, opened 2022-09-19
Remove HIP-only parameter for a test

Rombur:
With HIP we had a test that was much slower than the CUDA version and we had to use a different parameter for the test to run in a reasonable time. On Crusher, the test takes the same time than on V100. Let's see if the CI is still slow.

### DIFF
--- a/core/unit_test/TestWorkGraph.hpp
+++ b/core/unit_test/TestWorkGraph.hpp
@@ -159,12 +159,7 @@ struct TestWorkGraph {
 }  // anonymous namespace
 
 TEST(TEST_CATEGORY, workgraph_fib) {
-  // FIXME_HIP The test is very slow with HIP and it causes the CI to timeout
-#ifdef KOKKOS_ENABLE_HIP
-  int limit = 7;
-#else
   int limit = 27;
-#endif
   for (int i = 0; i < limit; ++i) {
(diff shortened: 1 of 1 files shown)


Rombur:
The timings look reasonable, it takes about 1sec to run the test.

masterleinad (approved):
https://cloud.cees.ornl.gov/jenkins-ci/blue/rest/organizations/jenkins/pipelines/Kokkos/runs/10584/nodes/50/steps/249/log/?start=0 shows:
```
4: [ RUN      ] hip.workgraph_fib
4: [       OK ] hip.workgraph_fib (1184 ms)
```
and https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/Kokkos/detail/Kokkos/10584/pipeline/51/ shows
```
4: [ RUN      ] hip.workgraph_fib
4: [       OK ] hip.workgraph_fib (1159 ms)
```

dalg24:
> The timings look reasonable, it takes about 1sec to run the test.

Was it on MI60 or MI100?  Is there any reason to think it matters?

Rombur:
> Was it on MI60 or MI100? 

I think it was slow on both MI60 and MI100. I am sure it was slow on MI100.

>  Is there any reason to think it matters?

No I think it was the compiler that was generating bad code.

dalg24 (approved):

PR #5468, opened 2022-09-19
Move acquisition of memory scratch space to its own function

Rombur:
Fixes comment https://github.com/kokkos/kokkos/pull/5448#discussion_r969649443

### DIFF
--- a/core/src/Cuda/Kokkos_Cuda_Instance.cpp
+++ b/core/src/Cuda/Kokkos_Cuda_Instance.cpp
@@ -563,37 +563,44 @@ Cuda::size_type *CudaInternal::scratch_functor(const std::size_t size) const {
   return m_scratchFunctor;
 }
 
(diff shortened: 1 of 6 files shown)


dalg24 (approved):
Thanks for the refactor.  (You need to fix the format.)
Did you consider whether some sort of scope guard that acquire scratch memory at construction and release it upon destruction would improve the code further?  (Not asking you to implement in this PR, just wondering if you explored that idea.)

masterleinad (changes requested):
We use the same type of code in `SYCL` so change it uniformly.

masterleinad:
Let me know if you prefer me to push the `SYCL` changes to this branch.

masterleinad (approved):
Looking somewhat deeper, we don't actually need the acquisition part in SYCL at the moment anyway.

PR #5478, opened 2022-09-21
HIP simplify code using C++17

Rombur:
This PR does two things:
 - we had a few `FIXME_HIP` to remind us that some functions could be `constexpr`
 - using `if constexpr` we can use a single function to compute the team size. Instead of having a function for `ParallelFor` and one for `ParallelReduce`

### DIFF
--- a/core/src/HIP/Kokkos_HIP_Parallel_Team.hpp
+++ b/core/src/HIP/Kokkos_HIP_Parallel_Team.hpp
@@ -103,7 +103,8 @@ class TeamPolicyInternal<HIP, Properties...>
     using closure_type =
         Impl::ParallelFor<FunctorType, TeamPolicy<Properties...>>;
 
-    return internal_team_size_common<BlockType::Max, closure_type>(f);
+    return internal_team_size_common<BlockType::Max, closure_type, FunctorType,
+                                     ParallelForTag>(f);
   }
 
   template <class FunctorType>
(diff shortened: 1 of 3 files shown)


PhilMiller (approved):

core/src/HIP/Kokkos_HIP_Parallel_Team.hpp
```diff
@@ -345,58 +351,20 @@ class TeamPolicyInternal<HIP, Properties...>
...
-  int internal_team_size_common_reduce(const FunctorType& f) const {
-    using Interface =
-        typename Impl::DeduceFunctorPatternInterface<ClosureType>::type;
-    using Analysis =
-        Impl::FunctorAnalysis<Interface, typename ClosureType::Policy,
-                              FunctorType>;
-
+  template <BlockType BlockSize, class ClosureType, class FunctorType,
+            class Tag>
+  int internal_team_size_common(FunctorType const& f) const {
```
dalg24, line 356: Did you consider keeping `FunctorType` as the last template argument so you don't need to specify it at call site?

masterleinad (approved):

core/src/HIP/Kokkos_HIP_Parallel_Team.hpp
Rombur: Yes, I did. Since we have to specify the other template arguments anyway, I didn't see the point in doing that.

dalg24:
```C++
internal_team_size_common<BlockType::Max, closure_type, FunctorType, ParallelReduceTag>(f);
// VS
internal_team_size_common<BlockType::Max, closure_type, ParallelReduceTag>(f);  // more concise, let the compiler deduce the functor type
```

Rombur: I don't understand your comment. If the goal is to show me that if you don't type `FunctorType` the line is shorter, you wasted your time I already knew that. If you want me to change it, just say it.

dalg24: Didn't understand your comment either.  Yes I would like you to change it.

PR #3962, opened 2021-04-23
Disable hypot long double test on power arches

ndellingwood:

PR #5484, opened 2022-09-22
Disable kk3_hypot in Power9 testing

ndellingwood:
Disable test on Power9 (and Power8) arches
Similar treatment for kk3_hypot as kk_hypot in #3962


ndellingwood:
Workaround for testing failures with kk3_hypot on Power9 with Serial and OpenMP (at least with gcc/8.3.0)

Sample output prior to disable:
```
[[0;32m[ RUN      ] ^[[mopenmp.mathematical_functions_power_functions
relative difference exceeds tolerance [9.155487e-33 > 9.881313e-324]
value at 2.000000, 3.000000, 4.000000 which is 5.385165 was expected to be 5.385165
/ascldap/users/ndellin/trilinos/Trilinos/kokkos/core/unit_test/TestMathematicalFunctions.hpp:562: Failure
  errors
    Which is: 1
  0
Failed check no error for hypot(long double, long double, long double)
```

dalg24 (approved):

cz4rs (approved):

PR #5485, opened 2022-09-22
Make sure we don't add '-cuda' to the link line with NVC++ when CUDA is not enabled

### LINKED ISSUE #5483: NVHPC compiler issues an error when compiled with the serial backend on ppc64le
The NVHPC compiler will issue an error if the '-cuda' option is used on a system without a proper NVIDIA GPU:

cmake options are:

`-DCMAKE_CXX_COMPILER=nvc++ -DCMAKE_CXX_FLAGS="-fast -mp " -DCMAKE_VERBOSE_MAKEFILE=ON -DKokkos_ENABLE_COMPILER_WARNINGS=ON -DKokkos_ENABLE_TESTS=ON -DKokkos_ENABLE_SERIAL=ON -DCMAKE_INSTALL_PREFIX=/tmp -DCMAKE_CXX_STANDARD=17    -DCMAKE_INSTALL_LIBDIR=lib64
...
[  6%] Linking CXX executable KokkosCore_TestCompileOnly
...
nvc++-Fatal-The `-cuda` option is available only on systems with NVIDIA GPUs with compute capability '>= cc60`

This occurs with the 3.7.00 release. I would recommend that the '-cuda' option only be applied when the CUDA backend is compiled.

dalg24:
Fix #5483

Obviously, in an ideal world we don't process that entire block when the CUDA backend is not enabled.  Wrapping it in a big `if(KOKKOS_ENABLE_CUDA)` makes the code less readable and I don't want to get into refactoring the whole thing.  I considered introducing a `KOKKOS_CUDA_LINK_OPTIONS` variables but that would force us to duplicate some target_link_options workaround for CMake version less than 3.18


masterleinad (approved):
Fine with me.

dalg24:
Performance test failing on `CUDA-11.0-NVCC-C++17-RDC` build is clearly unrelated

JBludau (commented):
looks good so far

JBludau (approved):

PR #5486, opened 2022-09-22
Fix configuring with Threads support when rerunning CMake

masterleinad:
This has been bugging me for a while. If `Kokkos` has been configured with the `Threads` backend, any attempt to reconfigure would fail since `Threads::Threads` would not be a target anymore.
This pull request fixes this bug, by pulling the relevant logic from `FindTPLPthread.cmake` into `FindTPLTHREADS.cmake"` and, thus, making the two independent. The former one is only used for `Trilinos` builds anyway (and the latter one is not used for `Trilinos` builds).


crtrott (approved):

cmake/Modules/FindTPLTHREADS.cmake
```diff
@@ -1,5 +1,5 @@
 INCLUDE(FindPackageHandleStandardArgs)
-INCLUDE("${CMAKE_CURRENT_SOURCE_DIR}/cmake/tpls/FindTPLPthread.cmake")
+FIND_PACKAGE(Threads)
```
dalg24, line 2: Remove the file if we don't use it any more

dalg24 (changes requested):

cmake/Modules/FindTPLTHREADS.cmake
masterleinad: AFAIK, it's used when building `Kokkos` as part of `Trilinos`.

dalg24: OK

dalg24 (approved):

PR #5488, opened 2022-09-22
Fix memory spaces in create_mirror_view overloads using view_alloc

masterleinad:
While working on https://github.com/arborx/ArborX/pull/749, I noticed that the new `Kokkos::view_alloc` overloads for `create_mirror_view` give the wrong memory space in some cases. This pull request tests the memory spaces and makes sure they are all correct. This is also a candidate for a patch release for 3.7 and thus no cleanup using C++17 is done here (but I certainly want to do that later).


containers/unit_tests/TestWithoutInitializing.hpp
```diff
@@ -465,28 +490,58 @@ TEST(TEST_CATEGORY, create_mirror_no_init_dynrankview) {
...
+            Kokkos::view_alloc(Kokkos::WithoutInitializing, TEST_EXECSPACE{},
+                               device_memory_space),
             host_view);
         ASSERT_EQ(host_view.size(), mirror_host.size());
+    // FIXME_MSVC
+#ifndef KOKKOS_COMPILER_MSVC
+        static_assert(
+            std::is_same_v<typename decltype(mirror_host)::memory_space,
+                           decltype(device_memory_space)>);
+#endif
```
masterleinad, line 521: I don't understand why these `static_assert`s fail with `MSVC`. They work fine on `Serial` only builds with other compilers.
I would propose to keep this as-is for now unless someone can help me debug this with MSVC.

core/unit_test/tools/TestWithoutInitializing.cpp
```diff
@@ -82,26 +108,46 @@ TEST(kokkosp, create_mirror_no_init_view_ctor) {
...
+            std::is_same_v<typename decltype(mirror_host)::memory_space,
+                           decltype(device_memory_space)>);
+#endif
         auto mirror_device_view = Kokkos::create_mirror_view(
             Kokkos::view_alloc(Kokkos::HostSpace{},
                                Kokkos::WithoutInitializing),
             device_view);
+        static_assert(
+            std::is_same_v<typename decltype(mirror_device_view)::memory_space,
+                           Kokkos::HostSpace>);
```
ldh4, line 139: There are mixed usages of `std::is_same` and `std::is_same_v` throughout this PR. If you want this patch to be a candidate patch for 3.7, I don't think you can use `std::is_same_v`?

masterleinad: Answering anyway: I expect that some minor adjustments would be needed for 3.7 anyway.

masterleinad:
Current `develop` fails with
```
/tmp/kokkos/core/src/Kokkos_View.hpp(1336): error: static assertion failed with "Incompatible View copy construction"
          detected during instantiation of "Kokkos::View<DataType, Properties...>::View(const Kokkos::View<RT, RP...> &, std::enable_if_t<Kokkos::Impl::ViewMapping<Kokkos::View<DataType, Properties...>::traits, Kokkos::View<RT, RP...>::traits, Kokkos::ViewTraits<DataType, Properties...>::specialize>::is_assignable_data_type, void> *) [with DataType=int, Properties=<Kokkos::CudaSpace>, RT=int, RP=<Kokkos::Serial::array_layout, Kokkos::Device<Kokkos::HostSpace::execution_space, Kokkos::HostSpace::memory_space>, Kokkos::Experimental::DefaultViewHooks>]" 
/tmp/kokkos/example/build_cmake_in_tree/cmake_example.cpp(6): here

/tmp/kokkos/core/src/impl/Kokkos_ViewMapping.hpp(3571): error: static assertion failed with "View assignment must have compatible spaces"
          detected during:
            instantiation of "void Kokkos::Impl::ViewMapping<DstTraits, SrcTraits, std::enable_if_t<<expression>, void>>::assign(Kokkos::Impl::ViewMapping<DstTraits, SrcTraits, std::enable_if_t<<expression>, void>>::DstType &, const Kokkos::Impl::ViewMapping<DstTraits, SrcTraits, std::enable_if_t<<expression>, void>>::SrcType &, const Kokkos::Impl::ViewMapping<DstTraits, SrcTraits, std::enable_if_t<<expression>, void>>::TrackType &) [with DstTraits=Kokkos::ViewTraits<int, Kokkos::CudaSpace>, SrcTraits=Kokkos::ViewTraits<int, Kokkos::Serial::array_layout, Kokkos::Device<Kokkos::HostSpace::execution_space, Kokkos::HostSpace::memory_space>, Kokkos::Experimental::DefaultViewHooks>]" 
/tmp/kokkos/core/src/Kokkos_View.hpp(1338): here
            instantiation of "Kokkos::View<DataType, Properties...>::View(const Kokkos::View<RT, RP...> &, std::enable_if_t<Kokkos::Impl::ViewMapping<Kokkos::View<DataType, Properties...>::traits, Kokkos::View<RT, RP...>::traits, Kokkos::ViewTraits<DataType, Properties...>::specialize>::is_assignable_data_type, void> *) [with DataType=int, Properties=<Kokkos::CudaSpace>, RT=int, RP=<Kokkos::Serial::array_layout, Kokkos::Device<Kokkos::HostSpace::execution_space, Kokkos::HostSpace::memory_space>, Kokkos::Experimental::DefaultViewHooks>]" 
/tmp/kokkos/example/build_cmake_in_tree/cmake_example.cpp(6): here
```
for
```
#include <Kokkos_Core.hpp>

  Kokkos::ScopeGuard scope_guard;
  Kokkos::View<int, Kokkos::HostSpace> v("bla");
  Kokkos::View<int, Kokkos::CudaSpace> v_mirror = Kokkos::create_mirror_view(Kokkos::view_alloc(Kokkos::CudaSpace{}), v);
}
```
because we accidentally use https://github.com/kokkos/kokkos/blob/b152efa94bc38d597870e5a275194cc00ed5afbf/core/src/Kokkos_CopyViews.hpp#L3658-L3670 instead of https://github.com/kokkos/kokkos/blob/b152efa94bc38d597870e5a275194cc00ed5afbf/core/src/Kokkos_CopyViews.hpp#L3681-L3691 and we try to assign a device view (the correct one we want) to a host view (which is wrong).

The fix here is to make sure that the second overload mentioned also only takes `ViewCtorProp` arguments, i.e, no extra `space` argument and the is called whenever a memory space is specified where as the first overload is only called if there is no memory space specified.

containers/unit_tests/TestWithoutInitializing.hpp
masterleinad: Fixed it.

PhilMiller (approved):

containers/src/Kokkos_OffsetView.hpp
```diff
@@ -2050,14 +2065,15 @@ template <class Space, class T, class... P,
...
 }
 
 template <class Space, class T, class... P>
 inline auto create_mirror_view(
-    Kokkos::Impl::WithoutInitializing_t wi, const Space& space,
+    Kokkos::Impl::WithoutInitializing_t wi, const Space&,
     const Kokkos::Experimental::OffsetView<T, P...>& src) {
-  return Impl::create_mirror_view(space, src, Kokkos::view_alloc(wi));
+  return Impl::create_mirror_view(
+      src, Kokkos::view_alloc(typename Space::memory_space{}, wi));
```
crtrott, line 2076: why not pass the space on?

containers/unit_tests/TestWithoutInitializing.hpp
```diff
@@ -437,21 +437,46 @@ TEST(TEST_CATEGORY, create_mirror_no_init_dynrankview) {
...
         ASSERT_EQ(device_view.size(), mirror_device.size());
+#ifndef KOKKOS_ENABLE_CUDA_UVM
+        static_assert(
+            std::is_same_v<typename decltype(mirror_device)::memory_space,
+                           Kokkos::HostSpace>);
+#else
+        static_assert(
+            std::is_same_v<typename decltype(mirror_device)::memory_space,
+                           MemorySpace>);
+#endif
```
crtrott, line 455: This is not the right condition to check. It happens to work but is not the right thing. You should check that if the device_view is host accessible, then its the same memory space, otherwise its hostspace.

core/unit_test/tools/TestWithoutInitializing.cpp
```diff
@@ -82,26 +107,46 @@ TEST(kokkosp, create_mirror_no_init_view_ctor) {
...
             device_view);
+        static_assert(
+            std::is_same_v<typename decltype(mirror_device)::memory_space,
+                           Kokkos::HostSpace>);
         auto mirror_host = Kokkos::create_mirror(
-            Kokkos::view_alloc(Kokkos::HostSpace{}, Kokkos::WithoutInitializing,
+            Kokkos::view_alloc(device_memory_space, Kokkos::WithoutInitializing,
                                Kokkos::DefaultExecutionSpace{}),
             host_view);
+    // FIXME_MSVC
```
crtrott, line 126: why would this fail on MSVC?

crtrott: ad a minimum we need a comment whats going on.

crtrott (changes requested):

containers/src/Kokkos_OffsetView.hpp
masterleinad: Because `space` might be an execution space and thus this would change behavior.

core/unit_test/tools/TestWithoutInitializing.cpp
masterleinad: I missed removing those as well in the last comment. The issue was `decltype` returning a reference for `MSVC.

containers/unit_tests/TestWithoutInitializing.hpp
masterleinad: I would say the test is designed for this to be the right thing to check but sure I'm happy to change the condition.

core/unit_test/tools/TestWithoutInitializing.cpp
```diff
@@ -47,25 +47,43 @@
...
+template <typename InputView, typename OutputView>
+void check_host_mirror(const InputView&, const OutputView&) {
+  using InputSpace  = typename InputView::memory_space;
+  using OutputSpace = typename OutputView::memory_space;
+  if constexpr (Kokkos::SpaceAccessibility<Kokkos::HostSpace,
+                                           InputSpace>::accessible)
+    static_assert(std::is_same_v<OutputSpace, InputSpace>);
+  else
+    static_assert(std::is_same_v<OutputSpace, Kokkos::HostSpace>);
+}
```
PhilMiller, line 59: I did some reading on this recently that leads me to think an unsatisfied `static_assert` in the discarded branch of an `if constexpr` makes the program 'ill formed, no diagnostic required'. I'm not entirely sure if I was reading the exact conditions involved correctly though.

Conversely, the non-C++17 version of this that doesn't use `if constexpr` would raise no such concerns. We'd probably want to make sure that the whole team understands them for future use cases, though.

With that in mind, could you stick each of the expressions in `constexpr bool` variables, and then just use them in a single `static_assert`?


Something like
```
(is_accessible && is_input) || (!is_accessible && is_host)
```

PhilMiller: Now that I've written that out, I think a ternary might be clearer

masterleinad: > I did some reading on this recently that leads me to think an unsatisfied static_assert in the discarded branch of an if constexpr makes the program 'ill formed, no diagnostic required'. I'm not entirely sure if I was reading the exact conditions involved correctly though.

Would you mind finding a reference to that? https://stackoverflow.com/a/62469617 cites the standard to say
> The validity of a template may be checked prior to any instantiation. [...] The program is ill-formed, no diagnostic required, if:
>
>  - no valid specialization can be generated for a template or a substatement of a constexpr if statement within a template >and the template is not instantiated, or...
> [...]

but here there are cases where both branches are valid.

containers/unit_tests/TestWithoutInitializing.hpp
masterleinad: @crtrott Changed in [`9c62e10` (#5488)](https://github.com/kokkos/kokkos/pull/5488/commits/9c62e102b8015cd0edd5f234dc2f50f66bbffa77) and [`0ed8de1` (#5488)](https://github.com/kokkos/kokkos/pull/5488/commits/0ed8de1abbf546c60eb675e2d1270f934b9af97d).

core/src/Kokkos_CopyViews.hpp
```diff
@@ -3657,37 +3658,47 @@ create_mirror_view(const Kokkos::View<T, P...>& src,
...
 
 // Create a mirror view in a new space (specialization for same space)
-template <class Space, class T, class... P, class... ViewCtorArgs>
-std::enable_if_t<Impl::MirrorViewType<Space, T, P...>::is_same_memspace,
-                 typename Impl::MirrorViewType<Space, T, P...>::view_type>
-create_mirror_view(const Space&, const Kokkos::View<T, P...>& src,
+template <class T, class... P, class... ViewCtorArgs,
+          typename = std::enable_if_t<
+              Impl::ViewCtorProp<ViewCtorArgs...>::has_memory_space>>
+std::enable_if_t<Impl::MirrorViewType<
```
dalg24, line 3678: Any good reason for doing the SFINAE in two separate places?

core/unit_test/tools/TestWithoutInitializing.cpp
```diff
@@ -47,25 +47,43 @@
...
+        static_assert(
+            std::is_same_v<decltype(mirror_host)::memory_space, MemorySpace>);
         auto mirror_device_view = Kokkos::create_mirror_view(
             Kokkos::WithoutInitializing, device_view);
+        check_host_mirror(device_view, mirror_device_view);
         auto mirror_host_view = Kokkos::create_mirror_view(
             Kokkos::WithoutInitializing, Kokkos::DefaultExecutionSpace{},
             host_view);
+        static_assert(std::is_same_v<decltype(mirror_host_view)::memory_space,
+                                     MemorySpace>);
```
dalg24, line 86: This unit test is about using the tools hooks to check that we allocate, fence, or launch kernels only when appropriate.  The compile time testing of `create_mirror` does not really belong there.  Did you consider moving these elsewhere and asserting in a more systematic way?

core/src/Kokkos_CopyViews.hpp
```diff
@@ -3657,37 +3658,47 @@ create_mirror_view(const Kokkos::View<T, P...>& src,
...
   return src;
 }
 
 // Create a mirror view in a new space (specialization for different space)
-template <class Space, class T, class... P, class... ViewCtorArgs>
-std::enable_if_t<!Impl::MirrorViewType<Space, T, P...>::is_same_memspace,
-                 typename Impl::MirrorViewType<Space, T, P...>::view_type>
-create_mirror_view(const Space&, const Kokkos::View<T, P...>& src,
+template <class T, class... P, class... ViewCtorArgs,
+          typename = std::enable_if_t<
```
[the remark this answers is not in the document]
masterleinad: I don't understand why that matters much but I'm happy to change.

core/src/Kokkos_CopyViews.hpp
masterleinad: I need to check for the existence of memory space before using its type in another restriction. Thus, the first check must be in the template arguments and the second somewhere later, e.g., in the return type.

core/unit_test/tools/TestWithoutInitializing.cpp
masterleinad: For the container tests, all these overloads are tested in the `TestWithoutInitializing` file. Thus, it feels like a good choice to extend those tests slightly instead of duplicating the setup somewhere else. By extension, it appears sensible to me to add the compile-time tests to a corresponding file in `core`.

core/src/Kokkos_CopyViews.hpp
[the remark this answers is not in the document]
masterleinad: See e6461b669.

crtrott:
I have been staring at this for a while, and while its only a limited complication above and beyond what was already there, I think we should use this opportunity to refactor and do better - more systematic - testing of this facility.

First: this file has become an unmitigated disaster. We need to split it. 

At a minimum that means: create_mirror[_view], local_deep_copy, resize and realloc should all get their own files. Maybe inside the src/View sub-directory?

Second: we need to have new test file for create_mirror[_view] instead of just sprinkling its testing everywhere. That is where we should both check all the compile time properties, and that the runtime behavior is correct. And we should design a systematic test where its much more obvious that we actually do test all combinations.

Right now I have no idea whether we do or we don't. 

We also should consolidate our testing of deep_copy. Which also is kinda happening in multiple places or? 

Thoughts?

masterleinad:
@crtrott I played with splitting `Kokkos_CopyViews.hpp` a bit, see [`b6643ef` (#5488)](https://github.com/kokkos/kokkos/pull/5488/commits/b6643ef0d126877a6f547335ddd3f06ec9302f91) but decided to revert (in https://github.com/kokkos/kokkos/pull/5488/commits/00aaba7b08716334f71d332ecb7a02799344528e) since it just obfuscates other changes made here. I think it makes more sense to do that in a follow-up pull request that doesn't change code. What do you think?

Secondly, I avoided modifying the `TestWithoutInitializing.hpp` files and introduced `TestCreateMirror` files that should cover the vast majority of use cases for `create_mirror*`. Is that what you had in mind?

dalg24:
> @crtrott I played with splitting `Kokkos_CopyViews.hpp` a bit, see [`b6643ef` (#5488)](https://github.com/kokkos/kokkos/pull/5488/commits/b6643ef0d126877a6f547335ddd3f06ec9302f91) but decided to revert (in [00aaba7](https://github.com/kokkos/kokkos/commit/00aaba7b08716334f71d332ecb7a02799344528e)) since it just obfuscates other changes made here. I think it makes more sense to do that in a follow-up pull request that doesn't change code. What do you think?

Yes, definitely go for minimal diff in the implementation and refactor in follow up PRs

> Secondly, I avoided modifying the `TestWithoutInitializing.hpp` files and introduced `TestCreateMirror` files that should cover the vast majority of use cases for `create_mirror*`. Is that what you had in mind?

Yes

containers/unit_tests/TestCreateMirror.hpp
```diff
@@ -0,0 +1,292 @@
...
+  auto mirror = Kokkos::create_mirror(std::forward<Args>(args)..., view);
+  check_host_mirror(view, mirror);
+}
+
+template <typename DeviceMemorySpace, typename View, typename... Args>
+void check_create_mirror_device(const View& view, Args&&... args) {
+  auto mirror = Kokkos::create_mirror(std::forward<Args>(args)..., view);
+  static_assert(std::is_same_v<typename decltype(mirror)::memory_space,
+                               DeviceMemorySpace>);
+}
```
dalg24, line 72: Consider doing something along these lines
```diff
diff --git a/core/unit_test/TestCreateMirror.hpp b/core/unit_test/TestCreateMirror.hpp
index 6ea99084e..631500230 100644
--- a/core/unit_test/TestCreateMirror.hpp
+++ b/core/unit_test/TestCreateMirror.hpp
@@ -44,6 +44,42 @@

 #include <Kokkos_Core.hpp>

+#define CHECK_MEMORY_SPACE(CREATE_MIRROR, MEMORY_SPACE)                        \
+  static_assert(std::is_same_v<typename decltype(CREATE_MIRROR)::memory_space, \
+                               MEMORY_SPACE>)
+template <class View>
+using conditional_if_accessible_from_host = std::conditional_t<
+    Kokkos::SpaceAccessibility<Kokkos::HostSpace,
+                               typename View::memory_space>::accessible,
+    typename View::memory_space, Kokkos::HostSpace>;
+
+// need to think more about the name
+#define COND_IF_HOST_ACCESSIBLE(VIEW) \
+  conditional_if_accessible_from_host<decltype(VIEW)>
+
+void test_create_mirror() {
+  using namespace Kokkos;
+  using DeviceExecutionSpace = TEST_EXECSPACE;
+  using DeviceMemorySpace    = TEST_EXECSPACE::memory_space;
+  DeviceExecutionSpace device_exec_space;
+  using HostExecutionSpace   = HostSpace::execution_space;
+  using HostMemorySpace      = HostSpace;
+  View<int, TEST_EXECSPACE> device_view("device");
+  View<int, HostSpace> host_view("host");
+  // clang-format off
+  CHECK_MEMORY_SPACE(create_mirror(                                                    device_view), COND_IF_HOST_ACCESSIBLE(device_view));
+  CHECK_MEMORY_SPACE(create_mirror(WithoutInitializing,                                device_view), COND_IF_HOST_ACCESSIBLE(device_view));
+  CHECK_MEMORY_SPACE(create_mirror(view_alloc(),                                       device_view), COND_IF_HOST_ACCESSIBLE(device_view));  // Just realized this compiles, not suggesting that's by design
+  CHECK_MEMORY_SPACE(create_mirror(view_alloc(WithoutInitializing),                    device_view), COND_IF_HOST_ACCESSIBLE(device_view));
+  CHECK_MEMORY_SPACE(create_mirror(                                 device_exec_space, device_view), DeviceMemorySpace);
+  CHECK_MEMORY_SPACE(create_mirror(WithoutInitializing,             device_exec_space, device_view), DeviceMemorySpace);
+
+  CHECK_MEMORY_SPACE(create_mirror_view(device_view), HostSpace);
+  CHECK_MEMORY_SPACE(create_mirror_view(device_view), HostSpace);
+  CHECK_MEMORY_SPACE(create_mirror_view(device_view), HostSpace);
+  // clang-format on
+}
+
 template <typename InputView, typename OutputView>
 void check_host_mirror(const InputView&, const OutputView&) {
   using InputSpace  = typename InputView::memory_space;

File: core/unit_test/TestCreateMirror.cpp [hunk 11]
```diff
@@ -0,0 +1,158 @@
...
+
+  // clang-format on
+}
+
+void test() {
+  Kokkos::View<int*, Kokkos::DefaultExecutionSpace> device_view("device view",
+                                                                10);
+  Kokkos::View<int*, Kokkos::HostSpace> host_view("host view", 10);
+
+  test_create_mirror_properties(device_view, host_view);
```
crtrott, line 157: If I see the code above correct its absolutely symmetric between host_view and device_view.
Couldn't we just have half as many calls above? I.e. you would use DefaultHostExecutionSpace and DefaultExecutionSpace replacing the Device/HostExecutionSpace in above code, you can get the HostMemorySpace from that too if so desired (or just use HostSpace since that is all we instantiate anyway).

And then just have here:
```
test_create_mirror_properties(device_view);
test_create_mirror_properties(host_view);
```

crtrott (changes requested):
I had a thought on reducing the test size a bit more since it seems to be basically completely symmetric for device_view/host_view

core/unit_test/TestCreateMirror.cpp
```diff
@@ -0,0 +1,125 @@
...
+  return std::conditional_t<
+      Kokkos::SpaceAccessibility<Kokkos::HostSpace,
+                                 typename View::memory_space>::accessible,
+      typename View::memory_space, Kokkos::HostSpace>{};
+}
+
+template <typename View>
+void test_create_mirror_properties(const View& view) {
+  using namespace Kokkos;
+  using DeviceMemorySpace = typename DefaultExecutionSpace::memory_space;
```
dalg24, line 63:
```suggestion
  using DeviceMemorySpace = DefaultExecutionSpace::memory_space;
```

core/unit_test/TestDetectionIdiom.cpp
```diff
@@ -86,5 +86,3 @@ static_assert(std::is_same<difference_type<Woof>, int>::value,
 static_assert(std::is_same<difference_type<Bark>, std::ptrdiff_t>::value,
               "Bark's difference_type should be ptrdiff_t!");
 }  // namespace Example
-
-int main() {}
```
dalg24, line 90: Not blocking but this change is unnecessary

crtrott (approved):

PR #5491, opened 2022-09-23
Allow as_view_of_rank_n() to be overloaded for "special" scalar types

etphipp:
as_view_of_rank_n() needs to be overloaded for special scalar types like Sacado that need to pass a hidden dimension.  This change adds the necessary SFINAE stuff to allow that to happen.  This will require adding the new overload in Trilinos/Sacado.

For Trilinos issue [#11018](https://github.com/trilinos/Trilinos/issues/11018) causing runtime failures in Albany.

@PhilMiller, @ndellingwood

### DIFF
--- a/core/src/Kokkos_View.hpp
+++ b/core/src/Kokkos_View.hpp
@@ -1730,7 +1730,10 @@ struct RankDataType<ValueType, 0> {
 };
 
 template <unsigned N, typename... Args>
-KOKKOS_FUNCTION std::enable_if_t<N == View<Args...>::Rank, View<Args...>>
+KOKKOS_FUNCTION std::enable_if_t<
+    N == View<Args...>::Rank &&
+        std::is_same<typename ViewTraits<Args...>::specialize, void>::value,
+    View<Args...>>
 as_view_of_rank_n(View<Args...> v) {
   return v;
 }
@@ -1738,8 +1741,9 @@ as_view_of_rank_n(View<Args...> v) {
 // Placeholder implementation to compile generic code for DynRankView; should
 // never be called
 template <unsigned N, typename T, typename... Args>
-std::enable_if_t<
-    N != View<T, Args...>::Rank,
+KOKKOS_FUNCTION std::enable_if_t<
+    N != View<T, Args...>::Rank &&
+        std::is_same<typename ViewTraits<T, Args...>::specialize, void>::value,
     View<typename RankDataType<typename View<T, Args...>::value_type, N>::type,
          Args...>>
 as_view_of_rank_n(View<T, Args...>) {
--- a/containers/src/Kokkos_DynRankView.hpp
+++ b/containers/src/Kokkos_DynRankView.hpp
@@ -1691,7 +1691,8 @@ template <unsigned N, typename T, typename... Args>
 KOKKOS_FUNCTION auto as_view_of_rank_n(
     DynRankView<T, Args...> v,
     typename std::enable_if<std::is_same<
-        typename ViewTraits<T, Args...>::specialize, void>::value>::type* = 0) {
+        typename ViewTraits<T, Args...>::specialize, void>::value>::type* =
+        nullptr) {
   if (v.rank() != N) {
     KOKKOS_IF_ON_HOST(
         const std::string message =
--- a/containers/src/Kokkos_DynRankView.hpp
+++ b/containers/src/Kokkos_DynRankView.hpp
@@ -1688,7 +1688,11 @@ namespace Impl {
    underlying memory, to facilitate implementation of deep_copy() and
    other routines that are defined on View */
 template <unsigned N, typename T, typename... Args>
-KOKKOS_FUNCTION auto as_view_of_rank_n(DynRankView<T, Args...> v) {
+KOKKOS_FUNCTION auto as_view_of_rank_n(
+  DynRankView<T, Args...> v,
+  typename std::enable_if<
+    std::is_same< typename ViewTraits<T,Args...>::specialize, void >::value
+  >::type * = 0) {
   if (v.rank() != N) {
     KOKKOS_IF_ON_HOST(
         const std::string message =
--- a/containers/src/Kokkos_DynRankView.hpp
+++ b/containers/src/Kokkos_DynRankView.hpp
@@ -1689,10 +1689,9 @@ namespace Impl {
    other routines that are defined on View */
 template <unsigned N, typename T, typename... Args>
 KOKKOS_FUNCTION auto as_view_of_rank_n(
-  DynRankView<T, Args...> v,
-  typename std::enable_if<
-    std::is_same< typename ViewTraits<T,Args...>::specialize, void >::value
-  >::type * = 0) {
+    DynRankView<T, Args...> v,
+    typename std::enable_if<std::is_same<
+        typename ViewTraits<T, Args...>::specialize, void>::value>::type* = 0) {
   if (v.rank() != N) {
     KOKKOS_IF_ON_HOST(
         const std::string message =


masterleinad:
You probably also want to change the original template https://github.com/kokkos/kokkos/blob/b152efa94bc38d597870e5a275194cc00ed5afbf/core/src/Kokkos_View.hpp#L1732-L1749, right?

masterleinad:
Also, the indentation is off: https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/Kokkos/detail/Kokkos/10673/pipeline
You'll need to use `clang-format-8` (or apply the patch in the link).

etphipp:
> You probably also want to change the original template
> 
> https://github.com/kokkos/kokkos/blob/b152efa94bc38d597870e5a275194cc00ed5afbf/core/src/Kokkos_View.hpp#L1732-L1749
> , right?

No, I don't think so.  Those overloads are fine.

masterleinad:
> No, I don't think so. Those overloads are fine.

You are saying/promising that you will never want to specialize them?

etphipp:
> > No, I don't think so. Those overloads are fine.
> 
> You are saying/promising that you will never want to specialize them?

I can't see any reason why they would need to be specialized.

PhilMiller:
> > > No, I don't think so. Those overloads are fine.
> > 
> > 
> > You are saying/promising that you will never want to specialize them?
> 
> I can't see any reason why they would need to be specialized.

That doesn't really make sense to me. If Sacado needs special treatment of a DynRankView, what lets it do without the same special treatment of a common View? Is it purely coincidence of the current implementations in Kokkos, or something structural?

etphipp:
> That doesn't really make sense to me. If Sacado needs special treatment of a DynRankView, what lets it do without the same special treatment of a common View? Is it purely coincidence of the current implementations in Kokkos, or something structural?

These functions don't create a new view.  If either did, it likely would need specialization.

masterleinad:
This pull request makes `as_view_of_rank_n` a customization point for `DynRankView` but not for any other type. I think it's safer to do it consistently so that if we ever decide to have another specialization, we also take into consideration that someone else (I think `Sacado` is the only one doing these kinds of things at te moment) might want to use it as a customization point.

PhilMiller:
That's my preference too. If any overload of a function needs to be guarded for specialization, please apply the guard to all of them. Otherwise, we end up with way more opportunities for changes on our side to produce unexpected errors on yours

masterleinad (changes requested):
Make the same restrictions on the original template.

ndellingwood:
Added patch release label for 3.7.01 patch release

masterleinad (approved):

masterleinad:
```
/__w/kokkos/kokkos/containers/unit_tests/../src/Kokkos_DynRankView.hpp:1694:77: error: use nullptr [modernize-use-nullptr,-warnings-as-errors]
```

PhilMiller (approved):
In a few years, all this SFINAE specialization cruft can go away with C++20 `requires`. Of course, by then, we expect that we'll have used MDSpan support to drastically simplify how Sacado builds FAD support onto Kokkos Views.

C'est la vie

ndellingwood:
CUDA-11.0-NVCC-C++17-RDC failure is CI issue, unrelated to PR:
```
+ docker build -t 558bd9e52e241f408d7b44ab2e4548b114eac94a --build-arg BASE=nvidia/cuda:11.0-devel --build-arg ADDITIONAL_PACKAGES=g++-8 gfortran clang --build-arg CMAKE_VERSION=3.17.3 -f scripts/docker/Dockerfile.nvcc scripts/docker

Sending build context to Docker daemon  23.04kB

Step 1/10 : ARG BASE=nvidia/cuda:9.2-devel

Step 2/10 : FROM $BASE

manifest for nvidia/cuda:11.0-devel not found: manifest unknown: manifest unknown

script returned exit code 1
```

Rombur:
Retest this please

ndellingwood:
fedora:rawhide, clang++, Release/Debug, OPENMP failure is a CI issue?
```
 -- Build files have been written to: /__w/kokkos/kokkos/example/build_cmake_installed/builddir
Scanning dependencies of target example
[ 33%] Building CXX object CMakeFiles/example.dir/cmake_example.cpp.o
[ 66%] Building Fortran object CMakeFiles/example.dir/foo.f.o
[100%] Linking CXX executable example
/usr/bin/ld: CMakeFiles/example.dir/foo.f.o: relocation R_X86_64_32S against `.rodata' can not be used when making a PIE object; recompile with -fPIE
/usr/bin/ld: failed to set dynamic section sizes: bad value
clang-15: error: linker command failed with exit code 1 (use -v to see invocation)
gmake[2]: *** [CMakeFiles/example.dir/build.make:119: example] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:83: CMakeFiles/example.dir/all] Error 2
gmake: *** [Makefile:101: all] Error 2
Error: Process completed with exit code 2.
```

masterleinad:
> fedora:rawhide, clang++, Release/Debug, OPENMP failure is a CI issue?

Yes, see https://github.com/kokkos/ci-containers/pull/28.

ndellingwood:
@etphipp thanks for you work on this! With this update approved+merged I think work on the corresponding Sacado updates can move forward (and without corresponding Sacado updates we will have compilation failures when testing Trilinos against Kokkos' develop branch, e.g. from intrepid2):
```
/ascldap/users/ndellin/trilinos/Trilinos/kokkos/containers/src/Kokkos_DynRankView.hpp(1740): error: no instance of overloaded function "Kokkos::Impl::as_view_of_rank_n" matches the argument list
            argument types are: (const Kokkos::DynRankView<Sacado::Fad::SFad<double, 10>, Kokkos::Device<Kokkos::CudaSpace::execution_space, Kokkos::CudaSpace::memory_space>>)
          detected during:
            instantiation of "Kokkos::ViewTraits<DataType *******, Properties...>::array_layout Kokkos::DynRankView<DataType, Properties...>::layout() const [with DataType=Sacado::Fad::SFad<double, 10>, Properties=<Kokkos::Device<Kokkos::CudaSpace::execution_space, Kokkos::CudaSpace::memory_space>>]"
/ascldap/users/ndellin/trilinos/Trilinos/packages/sacado/src/Kokkos_DynRankView_Fad.hpp(1198): here
            instantiation of "Kokkos::DynRankView<T, P...>::HostMirror Kokkos::create_mirror(const Kokkos::DynRankView<T, P...> &, std::enable_if<<expression>, void>::type *) [with T=Sacado::Fad::SFad<double, 10>, P=<Kokkos::Device<Kokkos::CudaSpace::execution_space, Kokkos::CudaSpace::memory_space>>]"
```

If you put in a PR to Trilinos please include these Kokkos-related changes as well (they'll get clobbered by the identical changes during the next release). We are moving closer to cutting the next release candidate, but the release process can take some time to complete (especially for a major release)

core/src/Kokkos_View.hpp
```diff
@@ -1730,16 +1730,20 @@ struct RankDataType<ValueType, 0> {
...
 as_view_of_rank_n(View<Args...> v) {
   return v;
 }
 
 // Placeholder implementation to compile generic code for DynRankView; should
 // never be called
 template <unsigned N, typename T, typename... Args>
-std::enable_if_t<
-    N != View<T, Args...>::Rank,
+KOKKOS_FUNCTION std::enable_if_t<
```
dalg24, line 1744: This causes
```
warning: calling a __host__ function from a __host__ __device__ function is not allowed
```
reported by @ldh4

PR #5492, opened 2022-09-23
OpenMPTarget: adding implementation to set device id.

rgayatri23:
The PR adds logic to set device id in the OpenMPTarget backend. 
Also updated the CI testing for OpenMPTarget to use llvm/15

### DIFF
--- a/core/src/Kokkos_OpenMPTarget.hpp
+++ b/core/src/Kokkos_OpenMPTarget.hpp
@@ -121,6 +121,8 @@ class OpenMPTarget {
     return m_space_instance;
   }
 
+  static int detect_device_count() { return omp_get_num_devices(); }
+
   OpenMPTarget();
   uint32_t impl_instance_id() const noexcept;
 
--- a/core/src/Kokkos_OpenMPTarget.hpp
+++ b/core/src/Kokkos_OpenMPTarget.hpp
@@ -121,7 +121,7 @@ class OpenMPTarget {
     return m_space_instance;
   }
 
-  static int detect_device_count() { return omp_get_num_devices(); }
+  inline static int detect_device_count() { return omp_get_num_devices(); }
 
   OpenMPTarget();
   uint32_t impl_instance_id() const noexcept;
--- a/core/src/Kokkos_OpenMPTarget.hpp
+++ b/core/src/Kokkos_OpenMPTarget.hpp
@@ -121,8 +121,6 @@ class OpenMPTarget {
     return m_space_instance;
   }
 
-  inline static int detect_device_count() { return omp_get_num_devices(); }
-
   OpenMPTarget();
   uint32_t impl_instance_id() const noexcept;
 
--- a/core/src/OpenMPTarget/Kokkos_OpenMPTarget_Instance.cpp
+++ b/core/src/OpenMPTarget/Kokkos_OpenMPTarget_Instance.cpp
@@ -47,6 +47,7 @@
 #endif
 
 #include <Kokkos_Macros.hpp>
+#include <impl/Kokkos_DeviceManagement.hpp>
 
 #if defined(KOKKOS_ENABLE_OPENMPTARGET) && defined(_OPENMP)
 
@@ -109,7 +110,8 @@ void OpenMPTargetInternal::impl_finalize() {
     Kokkos::kokkos_free<Kokkos::Experimental::OpenMPTargetSpace>(
         space.m_uniquetoken_ptr);
 }
-void OpenMPTargetInternal::impl_initialize() {
+void OpenMPTargetInternal::impl_initialize(
+    InitializationSettings const& settings) {
   m_is_initialized = true;
 
   // FIXME_OPENMPTARGET:  Only fix the number of teams for NVIDIA architectures
@@ -120,6 +122,10 @@ void OpenMPTargetInternal::impl_initialize() {
   omp_set_num_teams(512);
 #endif
 #endif
+
+  using Kokkos::Impl::get_gpu;
+  const int device_num = get_gpu(settings);
+  omp_set_default_device(device_num);
 }
 int OpenMPTargetInternal::impl_is_initialized() {
   return m_is_initialized ? 1 : 0;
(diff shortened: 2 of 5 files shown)


core/src/impl/Kokkos_Core.cpp
```diff
@@ -166,6 +166,8 @@ int get_device_count() {
 #elif defined(KOKKOS_ENABLE_OPENACC)
   return acc_get_num_devices(
       Kokkos::Experimental::Impl::OpenACC_Traits::dev_type);
+#elif defined(KOKKOS_ENABLE_OPENMPTARGET)
+  return Kokkos::Experimental::OpenMPTarget::detect_device_count();
```
masterleinad, line 170: If this is the only place where this function is used, I would prefer just inlining it.

masterleinad (commented):
Apart from the one comment, this looks good to me.

core/src/impl/Kokkos_Core.cpp
rgayatri23: Yes this is the only place that this function is called for now. I was just following the style of other backends.

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Instance.cpp
```diff
@@ -120,6 +122,10 @@ void OpenMPTargetInternal::impl_initialize() {
   omp_set_num_teams(512);
 #endif
 #endif
+
+  using Kokkos::Impl::get_gpu;
+  const int device_num = get_gpu(settings);
+  omp_set_default_device(device_num);
```
dalg24, line 128: Move this to ` OpenMPTarget::impl_initialize`.  We only want to run it once in the program.

dalg24 (changes requested):

core/src/Kokkos_OpenMPTarget.hpp
```diff
@@ -121,6 +121,8 @@ class OpenMPTarget {
     return m_space_instance;
   }
 
+  inline static int detect_device_count() { return omp_get_num_devices(); }
+
```
dalg24, line 125:
```suggestion
```

core/src/impl/Kokkos_Core.cpp
dalg24, line 170:
```suggestion
#elif defined(KOKKOS_ENABLE_OPENMPTARGET)
  return omp_get_num_devices();
```

dalg24 (approved):

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Instance.cpp
rgayatri23: This is already in `impl_initialize` right? or do you mean somewhere elase?

rgayatri23:
Addressed the comments.

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Instance.cpp
masterleinad: This is `OpenMPTargetInternal::impl_initialize` not `OpenMPTarget::impl_initialize`. Currently, the latter calls the former for the singleton but every other execution space instance would only call the former. Setting the device should be done by the latter, though, so it's only done once by the singleton.
Hence, it's better to move this call to `OpenMPTarget::impl_initialize`.

scripts/docker/Dockerfile.openmptarget
```diff
@@ -38,7 +38,7 @@ RUN CMAKE_URL=https://github.com/Kitware/CMake/releases/download/v${CMAKE_VERSIO
     rm ${CMAKE_SCRIPT}
 ENV PATH=${CMAKE_DIR}/bin:$PATH
 
-ARG LLVM_VERSION=llvmorg-14.0.0
```
masterleinad, line 41: Is this necessary for this pull request?

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Instance.cpp
rgayatri23: Thanks @masterleinad for the explanation. 
I moved the routine to `OpenMPTarget::impl_initialize`

scripts/docker/Dockerfile.openmptarget
rgayatri23: Not necessary for this. I reverted back to 14. Will update that in a different PR.

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Instance.cpp
```diff
@@ -109,7 +110,8 @@ void OpenMPTargetInternal::impl_finalize() {
     Kokkos::kokkos_free<Kokkos::Experimental::OpenMPTargetSpace>(
         space.m_uniquetoken_ptr);
 }
-void OpenMPTargetInternal::impl_initialize() {
+void OpenMPTargetInternal::impl_initialize(
+    InitializationSettings const& settings) {
```
masterleinad, line 114: This is now unused, right? I would expect the CI to complain about that.

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Instance.cpp
```diff
@@ -164,8 +166,12 @@ void OpenMPTarget::impl_static_fence(const std::string& name) {
...
 }
 
-void OpenMPTarget::impl_initialize(InitializationSettings const&) {
-  Impl::OpenMPTargetInternal::impl_singleton()->impl_initialize();
+void OpenMPTarget::impl_initialize(InitializationSettings const& settings) {
+  Impl::OpenMPTargetInternal::impl_singleton()->impl_initialize(settings);
+
+  using Kokkos::Impl::get_gpu;
+  const int device_num = get_gpu(settings);
+  omp_set_default_device(device_num);
```
dalg24, line 174: I would do it in that order to avoid the risk of bugs down the road
```suggestion
  using Kokkos::Impl::get_gpu;
  const int device_num = get_gpu(settings);
  omp_set_default_device(device_num);
  
    Impl::OpenMPTargetInternal::impl_singleton()->impl_initialize(settings);

```

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Instance.cpp
dalg24, line 114:
```suggestion
void OpenMPTargetInternal::impl_initialize() {
```

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Instance.hpp
```diff
@@ -79,7 +79,7 @@ class OpenMPTargetInternal {
   int impl_is_initialized();
   uint32_t impl_get_instance_id() const noexcept;
   //! Initialize, telling the CUDA run-time library which device to use.
-  void impl_initialize();
+  void impl_initialize(InitializationSettings const&);
```
dalg24, line 82:
```suggestion
  void impl_initialize();
```

dalg24 (approved):

masterleinad (approved):

PR #5499, opened 2022-09-27
Draft: Add mdspan-compatible layouts

nmm0:
This is the next step for https://github.com/kokkos/kokkos/issues/4421

Still WIP

### DIFF
--- a/core/src/View/MDSpan/Kokkos_MDSpan_Layout.hpp
+++ b/core/src/View/MDSpan/Kokkos_MDSpan_Layout.hpp
@@ -69,7 +69,7 @@ namespace Kokkos::Experimental {
 template <class Extents>
 class MDSpanLayoutLeft {
   using extents_type = Extents;
-  using size_type    = typename extents_type::size_type;
+  using index_type    = typename extents_type::index_type;
   using rank_type    = typename extents_type::rank_type;
   using layout_type  = LayoutLeft;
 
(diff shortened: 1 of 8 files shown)


core/src/View/MDSpan/Kokkos_MDSpan_Layout.hpp
```diff
@@ -0,0 +1,143 @@
...
+
+#ifdef KOKKOS_ENABLE_IMPL_MDSPAN
+
+namespace Kokkos {
+struct LayoutLeft;
+}
+
+namespace Kokkos::Experimental {
+template <class Extents>
+class MDSpanLayoutLeft {
```
crtrott, line 70: `MDSpanLayoutLeftMapping`
Also we should consider to put this in Impl namespace. Generally in `mdspan` the mappings are nested templates in the layouts. So the only portable way to get them is from Layout::mapping<extents>, and they are not an otherwise named type in the standard. I think we should do the same with our custom layouts or?

core/src/View/MDSpan/Kokkos_MDSpan_Layout.hpp
```diff
@@ -0,0 +1,143 @@
...
+  using dimension_type =
+      typename Impl::DimensionsFromExtents<extents_type>::type;
+  using offset_type = ::Kokkos::Impl::ViewOffset<dimension_type, layout_type>;
+
+  // This assumption should hold since we are using size_t everywhere in Kokkos
+  // but if we ever change this the assert should fail
+  static_assert(std::is_same_v<typename offset_type::size_type, index_type>);
+
+  extents_type m_extents;
+  offset_type m_offset;
```
crtrott, line 137: No we shouldn't store m_offset here. Doesn't that duplicate all the extents?

crtrott (changes requested):

core/src/View/MDSpan/Kokkos_MDSpan_Layout.hpp
nmm0: Yeah that's being changed (per the stuff we discussed at our meeting), still WIP

crtrott:
Closing this since I think with the padded layouts this is obsolete.

PR #5509, opened 2022-09-30
Replace nvidia/cuda:11.0-devel->nvidia/cuda:11.0.3-devel-ubuntu18.04

masterleinad:
Looking sat https://hub.docker.com/r/nvidia/cuda/tags?page=1&name=11.0 it seems `nvidia/cuda:11.0-devel` no longer exists and we also see
```
manifest for nvidia/cuda:11.0-devel not found: manifest unknown: manifest unknown
```
in our CI. This pull request suggests replacing it with `nvidia/cuda:11.0.3-devel-ubuntu18.04` which seems to be close enough.

### DIFF
--- a/.jenkins
+++ b/.jenkins
@@ -353,7 +353,7 @@ pipeline {
                         dockerfile {
                             filename 'Dockerfile.nvcc'
                             dir 'scripts/docker'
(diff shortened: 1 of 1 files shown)


JBludau (approved):
hmm seems like the bumped all their images of version 11 to 0.3

cz4rs (approved):

PR #5405, opened 2022-08-29
`SharedHostPinnedSpace` alias in fwd declaration

JBludau:
Will need a rebase after #5289 is merged -> done

At the moment there is no `SharedHostPinnedSpace` for `OpenMPTarget` and `OpenACC`.
The unittest checks the existance of the alias and asserts that the changes made in one ExecutionSpace are visible in the other.
Some of the examples/tutorials could be changed to use the new `SharedSpace` and `SharedHostPinnedSpace` -> separate pr

- [x] write documentation -> https://github.com/kokkos/kokkos-core-wiki/pull/177
- [ ] add to changelog

PR #5511, opened 2022-10-01
Fixup shared spaces

JBludau:
Fixup for: 

- discussion https://github.com/kokkos/kokkos/pull/5405#discussion_r982473816 (continued in the dev-meeting + decision to only define the symbols `KOKKOS_HAS_SHARED_SPACE` and `KOKKOS_HAS_SHARED_HOST_PINNED_SPACE` but not define them to `1` for enabled. Decision mainly to keep it consistent in Kokkos). 
- Removed some unnecessary brackets in cmake. 
- Fixed bug in `TestSharedSpace.cpp`, it now compares to the right timings on the host.


core/unit_test/TestSharedSpace.cpp
```diff
@@ -156,62 +155,57 @@ TEST(defaultdevicetype, shared_space) {
...
-      incrementInLoop<Kokkos::DefaultHostExecutionSpace>(defaultHostExecData,
-                                                         numRepetitions);
+  auto hostLocalTimings =
+      incrementInLoop<HostExecutionSpace>(hostData, numRepetitions);
 
   // GET PAGE MIGRATING TIMINGS DATA
-  std::vector<decltype(defaultExecLocalTimings)> defaultExecSharedTimings{};
-  std::vector<decltype(defaultExecLocalTimings)> defaultHostExecSharedTimings{};
+  std::vector<decltype(deviceLocalTimings)> deviceSharedTimings{};
+  std::vector<decltype(deviceLocalTimings)> hostSharedTimings{};
```
dalg24, line 180: I know they are the same type but I'd prefer 
```suggestion
  std::vector<decltype(hostLocalTimings)> hostSharedTimings{};
```

core/unit_test/TestSharedSpace.cpp
```diff
@@ -156,62 +155,57 @@ TEST(defaultdevicetype, shared_space) {
...
+                  deviceSharedTimings[cycle].end(), [&](const uint64_t timing) {
+                    (timing < threshold * deviceLocalMean)
                         ? fastAsLocalOnRepeatedAccess &= true
                         : fastAsLocalOnRepeatedAccess &= false;
                   });
 
-    std::for_each(std::next(defaultHostExecSharedTimings[cycle].begin()),
-                  defaultHostExecSharedTimings[cycle].end(),
-                  [&](const uint64_t timing) {
-                    (timing < threshold * defaultExecLocalMean)
```
dalg24, line 214: So that was the issue?  `defaultExecLocalMean` instead of `hostLocalMean` in the line above?

JBludau: yup ... it was comparing to device timings instead of host timings but host timings is what we wanted

dalg24 (approved):

janciesko:
LGTM

janciesko (approved):

PR #5512, opened 2022-10-03
SIMD AVX2 backend

ibaned:
This work adds an AVX2 backend for SIMD types which is activated when Kokkos_ARCH_BDW is used.

It also adds additional common fallback implementations of SIMD operations such as horizontal reductions which are not supported by the AVX2 instruction set.

### DIFF
--- a/simd/src/Kokkos_SIMD_Common.hpp
+++ b/simd/src/Kokkos_SIMD_Common.hpp
@@ -338,6 +338,20 @@ template <class T, class Abi>
 }
 
 template <class T, class Abi>
+[[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION T hmax(
+    const_where_expression<simd_mask<T, Abi>,
+                           simd<T, Abi>> const&
+        x) {
+  auto const& v = x.value();
+  auto const& m = x.mask();
+  auto result = Kokkos::reduction_identity<T>::max();
+  for (std::size_t i = 0; i < v.size(); ++i) {
+    if (m[i]) result = Kokkos::max(result, v[i]);
+  }
+  return result;
+}
+
+template <class T, class Abi>
 [[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION T reduce(
     const_where_expression<simd_mask<T, Abi>,
                            simd<T, Abi>> const&
--- a/simd/src/Kokkos_SIMD_Common.hpp
+++ b/simd/src/Kokkos_SIMD_Common.hpp
@@ -154,6 +154,14 @@ template <class T>
   return const_where_expression(mask, value);
 }
 
+// fallback simd multiplication using generator constructor
+
+template <class T, class Abi>
+[[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
+simd<T, Abi> operator*(simd<T, Abi> const& lhs, simd<T, Abi> const& rhs) {
+  return simd<T, Abi>([&] (std::size_t i) { return lhs[i] * rhs[i]; });
+}
+
 // The code below provides:
 // operator@(simd<T, Abi>, Arithmetic)
 // operator@(Arithmetic, simd<T, Abi>)
--- a/simd/src/Kokkos_SIMD_Common.hpp
+++ b/simd/src/Kokkos_SIMD_Common.hpp
@@ -323,6 +323,19 @@ template <class T, class Abi>
   return a == simd_mask<T, Abi>(false);
 }
 
+template <class T, class Abi>
+[[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION double hmin(
+    const_where_expression<simd_mask<T, Abi>,
+                           simd<T, Abi>> const&
+        x) {
+  auto const v = x.value();
+  auto result = v[0];
+  for (std::size_t i = 1; i < v.size(); ++i) {
+    result = Kokkos::min(result, v[1]);
+  }
+  return result;
+}
+
 }  // namespace Experimental
 
 template <class T, class Abi>
--- a/simd/src/Kokkos_SIMD_Common.hpp
+++ b/simd/src/Kokkos_SIMD_Common.hpp
@@ -328,10 +328,26 @@ template <class T, class Abi>
     const_where_expression<simd_mask<T, Abi>,
                            simd<T, Abi>> const&
         x) {
-  auto const v = x.value();
-  auto result = v[0];
-  for (std::size_t i = 1; i < v.size(); ++i) {
-    result = Kokkos::min(result, v[1]);
+  auto const& v = x.value();
+  auto const& m = x.mask();
+  auto result = Kokkos::reduction_identity<T>::min();
+  for (std::size_t i = 0; i < v.size(); ++i) {
+    if (m[i]) result = Kokkos::min(result, v[i]);
+  }
+  return result;
+}
+
+template <class T, class Abi>
+[[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION double reduce(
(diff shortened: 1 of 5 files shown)


simd/src/Kokkos_SIMD_Scalar.hpp
```diff
@@ -326,7 +326,7 @@ template <class T, class BinaryOp>
 [[nodiscard]] KOKKOS_FORCEINLINE_FUNCTION T
 reduce(const_where_expression<simd_mask<T, simd_abi::scalar>,
                               simd<T, simd_abi::scalar>> const& x,
-       T identity_element, BinaryOp) {
+       T identity_element, std::plus<>) {
```
dalg24, line 329: Did you mean to change?  `BinaryOp` template argument is not used any more and FTAD cannot kick in any more

simd/src/Kokkos_SIMD_Common.hpp
```diff
@@ -154,6 +154,14 @@ template <class T>
   return const_where_expression(mask, value);
 }
 
+// fallback simd multiplication using generator constructor
+
+template <class T, class Abi>
+[[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION simd<T, Abi> operator*(
```
dalg24, line 160: Why is this needed?  What about other arithmetic operators?

simd/src/Kokkos_SIMD_Common.hpp
```diff
@@ -323,6 +331,43 @@ template <class T, class Abi>
...
+reduce(const_where_expression<simd_mask<T, Abi>, simd<T, Abi>> const& x, T,
+       std::plus<>) {
+  auto const& v = x.value();
+  auto const& m = x.mask();
+  auto result   = Kokkos::reduction_identity<T>::sum();
+  for (std::size_t i = 0; i < v.size(); ++i) {
+    if (m[i]) result += v[i];
+  }
+  return result;
+}
```
dalg24, line 369: These changes seems to not be necessarily tied to AVX2.  Make sure the description of your PR capture their being added

crtrott (changes requested):
That one change looks pretty suspicious. Also still want to do a side by side comparison with the AVX2/AVX512 file to check all stuff is covered and consistently organized.

simd/src/Kokkos_SIMD_Scalar.hpp
ibaned: Ah good catch, thank you! I should've removed the template parameter as well. I'll do this now.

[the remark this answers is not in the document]
ibaned: Yes, but I forgot to remove the template parameter as @dalg24 points out. Making it specific to summation resolves ambiguity between this and the fallback in `Kokkos_SIMD_Common.hpp`.

simd/src/Kokkos_SIMD_Common.hpp
ibaned: It is needed because AVX2 does not support multiplication of a vector of 64-bit integers. We could add the other operators for consistency if desired, there are just no backends yet that would use them.

simd/src/Kokkos_SIMD_Common.hpp
dalg24: Yes please do that

ibaned:
I removed the leftover `BinaryOp` template parameter and changed the PR description. Please let me know if you want fallback implementations for all four basic arithmetic operators, or any other changes. Thanks!

ibaned:
The `OPENMPTARGET` failure in the CI seems unrelated to these changes

ibaned:
@stanmoore1 the branch is functional and my Sandia application runs correctly with it. We're just going through Kokkos team review but no features are missing (I think).

simd/src/Kokkos_SIMD_Common.hpp
```diff
@@ -323,6 +331,43 @@ template <class T, class Abi>
...
+  auto result   = Kokkos::reduction_identity<T>::max();
+  for (std::size_t i = 0; i < v.size(); ++i) {
+    if (m[i]) result = Kokkos::max(result, v[i]);
+  }
+  return result;
+}
+
+template <class T, class Abi>
+[[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION T
+reduce(const_where_expression<simd_mask<T, Abi>, simd<T, Abi>> const& x, T,
```
PhilMiller, line 360: Where did the choice of the name `reduce` come from? This would seem to be more sensibly `hsum` (I'm assuming the `h` prefix is 'horizontal' across the elements of the vector)

simd/src/Kokkos_SIMD_Common.hpp
ibaned: @PhilMiller do you mean a comment in the C++ source where the fallback is defined or in the merge request message?

simd/src/Kokkos_SIMD_Common.hpp
PhilMiller: I see that's a pre-existing name in `Kokkos_SIMD_Scalar.hpp`. This is still in `namespace Experimental`, so what do others think about renaming this?

ibaned: @PhilMiller it comes from ISO C++ document N4808, the parallelism TS, which in turn is trying to be somewhat consistent with how ISO C++ reduction already look today. The inconsistency you're pointing out can be traced back to existing ISO C++ parallel algorithms, where there is nothing called `sum`, instead one uses `reduce` with `plus` as the reduction binary operation. Similarly, there is no binary operation usable for computing the minimum using `reduce` in ISO C++, instead one needs to use the `min` overload that takes a range. In my opinion, `min`, `max`, `hmin`, and `hmax` should all use `reduce` instead, but the design objective with Kokkos SIMD is to be consistent with the parallelism TS.

ibaned: I'm okay with defining an `hsum` that is only available in Kokkos, @crtrott and other can probably comment on the overall design and whether that is something we could submit as an ISO C++ paper or something.

simd/src/Kokkos_SIMD_Common.hpp
PhilMiller: I meant a source code comment

simd/src/Kokkos_SIMD_Common.hpp
PhilMiller: Ok, I hadn't realized the context of what we were aligning the design with. A possible amendment is worth talking about briefly, but it definitely doesn't play a role for this PR.

ibaned:
I just addressed the comment requested by @PhilMiller . Please let me know if any further changes are desired!

ibaned:
@crtrott are we waiting on the side by side comparison you mentioned?

ibaned:
@ajpowelsnl I haven't heard anything back from @crtrott , I assume he is just swamped

crtrott:
I actually was looking for you today, because i spend an hour reviewing this morning and have a few questions.

ibaned:
Ah okay! Today is my part-time telecommuting day but I'll send you a calendar invite?

crtrott (approved):
Approving for now and then have follow up PR to harmonize AVX2/AVX512 interface. Note: the AVX512 had already inconsistencies of the same nature as here, between the different scalar implementations. Discussed with Dan, and decided it is better to do a harmonization PR.

PhilMiller:
We may want to consider support for targeting this SIMD variant even when we're optimizing for micro-architectures that support AVX-512. Depending on intensity of usage, AVX-512 can end up being detrimental to performance because of clock speed reductions while it's active, and the transition times as the vector units spin up/down. So, not necessary for this PR, but we shouldn't strictly tie it to what architecture we say we're targeting.

PR #5516, opened 2022-10-03
CTAD (deduction guides) for MDRangePolicy

nliber:

### DIFF
--- a/core/src/KokkosExp_MDRangePolicy.hpp
+++ b/core/src/KokkosExp_MDRangePolicy.hpp
@@ -155,9 +155,20 @@ TileSizeProperties get_tile_size_properties(const ExecutionSpace&) {
 
 // multi-dimensional iteration pattern
 template <typename... Properties>
-struct MDRangePolicy : public Kokkos::Impl::PolicyTraits<Properties...> {
-  using traits       = Kokkos::Impl::PolicyTraits<Properties...>;
-  using range_policy = RangePolicy<Properties...>;
+struct MDRangePolicy;
+
+// Note: If MDRangePolicy has a primary template, implicit CTAD (deduction
+// guides) are generated -> MDRangePolicy<> by some compilers, which is
+// incorrect.  By making it a template specialization instead, no implicit CTAD
+// is generated.  This works because there has to be at least one property
+// specified (which is Rank<...>); otherwise, we'd get the static_assert
+// "Kokkos::Error: MD iteration pattern not defined".  This template
+// specialization uses <P, Properties...> in all places for correctness.
+template <typename P, typename... Properties>
+struct MDRangePolicy<P, Properties...>
+    : public Kokkos::Impl::PolicyTraits<P, Properties...> {
+  using traits       = Kokkos::Impl::PolicyTraits<P, Properties...>;
+  using range_policy = RangePolicy<P, Properties...>;
 
   typename traits::execution_space m_space;
 
@@ -166,8 +177,8 @@ struct MDRangePolicy : public Kokkos::Impl::PolicyTraits<Properties...> {
                   typename traits::schedule_type, typename traits::index_type>;
 
   using execution_policy =
-      MDRangePolicy<Properties...>;  // needed for is_execution_space
-                                     // interrogation
+      MDRangePolicy<P, Properties...>;  // needed for is_execution_space
+                                        // interrogation
 
   template <class... OtherProperties>
   friend struct MDRangePolicy;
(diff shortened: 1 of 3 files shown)


dalg24:
Virtually all the builds are failing on the ORNL Jenkins CI server (does not compile)
https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/Kokkos/detail/Kokkos/10820/pipeline

dalg24:
Retest this please

nliber:
Tests are all compile-time now.

masterleinad (approved):
Looks OK to me.

core/src/KokkosExp_MDRangePolicy.hpp
```diff
@@ -150,9 +150,20 @@ TileSizeProperties get_tile_size_properties(const ExecutionSpace&) {
 
 // multi-dimensional iteration pattern
 template <typename... Properties>
-struct MDRangePolicy : public Kokkos::Impl::PolicyTraits<Properties...> {
-  using traits       = Kokkos::Impl::PolicyTraits<Properties...>;
-  using range_policy = RangePolicy<Properties...>;
+struct MDRangePolicy;
+
+// Note: If MDRangePolicy has a primary template, implicit CTAD (deduction
+// guides) are generated -> MDRangePolicy<> by some compilers, which is
```
fnrizzi, line 156: just curious, which compilers do this?

core/src/KokkosExp_MDRangePolicy.hpp
```diff
@@ -358,6 +369,60 @@ struct MDRangePolicy : public Kokkos::Impl::PolicyTraits<Properties...> {
...
+
+template <typename ES, typename LT, size_t N, typename UT,
+          typename = std::enable_if_t<is_execution_space_v<ES>>>
+MDRangePolicy(ES const&, const LT (&)[N], const UT (&)[N])
+    ->MDRangePolicy<ES, Rank<N>>;
+
+template <typename ES, typename LT, size_t N, typename UT, typename TT,
+          size_t TN, typename = std::enable_if_t<is_execution_space_v<ES>>>
+MDRangePolicy(ES const&, const LT (&)[N], const UT (&)[N], const TT (&)[TN])
+    ->MDRangePolicy<ES, Rank<N>>;
```
fnrizzi, line 396: are the guides above going to be pick up the case when user pass an initializer_list? 

https://kokkos.github.io/kokkos-core-wiki/API/core/policies/MDRangePolicy.html#_CPPv4I000E13MDRangePolicyRKNSt16initializer_listI2OTEERKNSt16initializer_listI2ITEE

masterleinad: That should work but only if you don't use `std::initializer_list` directly (but in that case, you can't know the number of elements at compile-time anyway since that's not part of the type), see https://godbolt.org/z/cMd64vsxT.

core/src/KokkosExp_MDRangePolicy.hpp
```diff
@@ -161,8 +172,8 @@ struct MDRangePolicy : public Kokkos::Impl::PolicyTraits<Properties...> {
                   typename traits::schedule_type, typename traits::index_type>;
 
   using execution_policy =
-      MDRangePolicy<Properties...>;  // needed for is_execution_space
-                                     // interrogation
+      MDRangePolicy<P, Properties...>;  // needed for is_execution_space
```
masterleinad, line 175:
```suggestion
      MDRangePolicy<P, Properties...>;  // needed for is_execution_policy
```

core/unit_test/TestMDRangePolicyCTAD.cpp
```diff
@@ -0,0 +1,111 @@
...
+      std::is_same_v<Kokkos::MDRangePolicy<Kokkos::Rank<std::size(t)>>,
+                     decltype(Kokkos::MDRangePolicy(des, t, tt))>);
+  static_assert(
+      std::is_same_v<Kokkos::MDRangePolicy<Kokkos::Rank<std::size(t)>>,
+                     decltype(Kokkos::MDRangePolicy(notEs, t, t))>);
+
+  static_assert(
+      std::is_same_v<
+          Kokkos::MDRangePolicy<SomeExecutionSpace, Kokkos::Rank<std::size(t)>>,
+          decltype(Kokkos::MDRangePolicy(ses, t, t))>);
```
masterleinad, line 78: Can you duplicate this to use inline initializer lists without specifying the number of elements explicitly?

core/src/KokkosExp_MDRangePolicy.hpp
nliber: Implicit guides are generated from the primary template.

Because there are no template parameters to deduce in a parameter pack, both clang and gcc deduce a matching ctor to `MDRangePolicy<>`, while nvcc will generate an error.  Unfortunately, I find the standard wording on implicit guides to be fairly inscrutable (and cppreference doesn't help, as all it does is quote the standard) :-(.  While I'd certainly like to know, practically speaking, it doesn't matter to us which one is correct, as we have to support all those compilers anyway.

Since MDRangePolicy always requires at least one template parameter (`Rank`), any guides that deduce to `MDRangePolicy<>` are just wrong.

core/src/KokkosExp_MDRangePolicy.hpp
nliber: Done.

core/unit_test/TestMDRangePolicyCTAD.cpp
nliber: Added.

masterleinad (approved):
Looks OK to me. Maybe fix the double whitespace on the comments.

dalg24 (approved):

crtrott (approved):

crtrott (approved):

nliber:
Old typing habits last forever... :-)

core/src/KokkosExp_MDRangePolicy.hpp
```diff
@@ -155,9 +155,20 @@ TileSizeProperties get_tile_size_properties(const ExecutionSpace&) {
...
 // multi-dimensional iteration pattern
 template <typename... Properties>
-struct MDRangePolicy : public Kokkos::Impl::PolicyTraits<Properties...> {
-  using traits       = Kokkos::Impl::PolicyTraits<Properties...>;
-  using range_policy = RangePolicy<Properties...>;
+struct MDRangePolicy;
+
+// Note: If MDRangePolicy has a primary template, implicit CTAD (deduction
+// guides) are generated -> MDRangePolicy<> by some compilers, which is
+// incorrect.  By making it a template specialization instead, no implicit CTAD
```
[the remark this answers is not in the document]
dalg24: I ignored because I also often use double space and there are more occurrences in that file.

dalg24:
Ignoring the one HIP failure.

PR #5517, opened 2022-10-03
CTAD (deduction guides) for RangePolicy

nliber:

### DIFF
--- a/core/src/Kokkos_ExecPolicy.hpp
+++ b/core/src/Kokkos_ExecPolicy.hpp
@@ -261,6 +261,23 @@ class RangePolicy : public Impl::PolicyTraits<Properties...> {
   };
 };
 
+RangePolicy()->RangePolicy<>;
+
+template <typename... Args>
+RangePolicy(int64_t, int64_t, Args...)->RangePolicy<>;
+
+// enable_if is used in these deduction guides in order
+// to get around a gcc8 deduction guide overload set bug
+template <typename ES, typename... Args,
+          typename = std::enable_if_t<
+              std::is_convertible_v<ES, DefaultExecutionSpace>>>
+RangePolicy(ES const&, int64_t, int64_t, Args...)->RangePolicy<>;
+
+template <typename ES, typename... Args,
+          typename = std::enable_if_t<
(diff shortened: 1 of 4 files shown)


dalg24:
All the CUDA builds with NVCC are failing (compiler segfault)

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -280,6 +280,23 @@ class RangePolicy : public Impl::PolicyTraits<Properties...> {
   };
 };
 
+RangePolicy()->RangePolicy<>;
+
+template <typename... Args>
+RangePolicy(int64_t, int64_t, Args...)->RangePolicy<>;
```
crtrott, line 286: Question: should we deduce IndexType too?

I.e. 
```
template <typename IndexT, typename ... Args>
RangePolicy(int64_t, IndexT, Args...) -> RangePolicy<IndexType<IndexT>>;
```

Note I explicitly used IndexT for the second thing: since the second number is larger than the first anyway. 

So `RangePolicy(0,4000000000L)` does the right thing. 
I guess we could also use two template parameters and `common_type`?

crtrott (changes requested):
More discussion

core/src/Kokkos_ExecPolicy.hpp
nmm0: In practice if we deduce IndexType, in the majority of cases when users input literals, they won't add suffixes and we will deduce `int` and we could run into situations with a potential overflow like you mentioned at the meeting. Honestly if a user wants a smaller type they should make it explicit by specifying the template params.

crtrott: For RangePolicy I think it would be fine, since the maximum number we need to compute is right there. Same for TeamPolicy I think. MDRangePolicy is the one where it becomes a problem.

fnrizzi: I think it would make sense to template for the index type or also specialize for size_t, when someone does: 
```cpp
std::size_t b(0);
std::size_t e(111);
RangePolicy r(b, e);
```

crtrott:
Super-ceded by #6850

PR #5519, opened 2022-10-03
CTAD (deduction guides) for View

nliber:

### DIFF
--- a/core/src/Kokkos_View.hpp
+++ b/core/src/Kokkos_View.hpp
@@ -1708,6 +1708,36 @@ class View : public ViewTraits<DataType, Properties...> {
   }
 };
 
+// These deduction guides result in an unmanaged View
+template <typename T>
+View(T*)->View<T>;
+
+template <typename T>
+View(T*, size_t)->View<T*>;
+
+template <typename T>
+View(T*, size_t, size_t)->View<T**>;
+
+template <typename T>
+View(T*, size_t, size_t, size_t)->View<T***>;
+
+template <typename T>
+View(T*, size_t, size_t, size_t, size_t)->View<T****>;
+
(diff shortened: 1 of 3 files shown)


core/unit_test/TestViewCtorCTADs.hpp
```diff
@@ -0,0 +1,167 @@
...
+
+struct TestViewCtorCTADs {
+  using DataType                     = double;
+  static constexpr DataType* m       = nullptr;
+  static constexpr const DataType* c = nullptr;
+
+  static void test_viewctorcopymove() {
+    Kokkos::View<DataType> vt;
+    Kokkos::View cd(vt);
+    ASSERT_TRUE((std::is_same_v<decltype(cd), decltype(vt)>));
```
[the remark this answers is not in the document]
nliber: All the CTAD tests can be compile-only.  Why this particular one?

core/src/Kokkos_View.hpp
```diff
@@ -1708,6 +1708,36 @@ class View : public ViewTraits<DataType, Properties...> {
   }
 };
 
+// These deduction guides result in an unmanaged View
+template <typename T>
+View(T*)->View<T>;
```
crtrott, line 1713: Bringing up the question again: should this be AnonymousSpace?

crtrott (changes requested):
Blocking for now until we come to a resolution which memory space should be deduced.

core/src/Kokkos_View.hpp
dalg24: Or better: Do nothing ie do not provide deduction guidelines at all

dalg24:
Consensus to close

PR #5522, opened 2022-10-05
Remove support for MI25 and add support for NAVI 1030

Rombur:
- ROCm does not support MI25 anymore so removing support. 
 - Adding support for AMD Radeon Pro W6800 and V620. This was asked on slack but I have no way to check that it works.


crtrott:
Consider this for a 3.7.01 patch release?

crtrott:
We don't a user on Slack will test.

crtrott (approved):

core/src/impl/Kokkos_SharedAlloc.hpp
```diff
@@ -65,7 +65,7 @@ class SharedAllocationHeader {
  private:
   using Record = SharedAllocationRecord<void, void>;
 
-#ifdef KOKKOS_ARCH_VEGA
+#if defined(KOKKOS_ARCH_VEGA) || defined(KOKKOS_ARCH_NAVI)
```
dalg24, line 68: Is that something that will need more investigation or that is definitely the right tuned value?

crtrott: We don't know. The larger value is in principle the safer value, but wastes memory if not needed.

crtrott: RDNA 1 architecture has only 128 byte L2 cache line size. But this is RDNA 2 and I couldn't find a similar in-depth architecture whitepaper for it.

ozturkosu:
I have tested it. It is failing some cases which seems to us, it is successfully running on AMD GPU however output of test cases are mismatch. I will share detailed log file with you asap

ozturkosu:
> I have tested it. It is failing some cases which seems to us, it is successfully running on AMD GPU however output of test cases are mismatch. I will share detailed log file with you asap

You can find logs here

https://drive.google.com/drive/folders/1GigrT7OaVtEMX8TXOkRVQdFe6BvmUXot?usp=sharing

Thanks

dalg24:
Removed "patch release" tag because MI25 are still supported in ROCm 4.5 (minimum required in Kokkos 3.7) and NAVI support needs more work

Rombur:
I've rebased the PR. This is ready to be merged.

Makefile.kokkos
```diff
@@ -13,7 +13,7 @@ KOKKOS_DEVICES ?= "Threads"
 # NVIDIA:   Kepler,Kepler30,Kepler32,Kepler35,Kepler37,Maxwell,Maxwell50,Maxwell52,Maxwell53,Pascal60,Pascal61,Volta70,Volta72,Turing75,Ampere80,Ampere86,Hopper90
 # ARM:      ARMv80,ARMv81,ARMv8-ThunderX,ARMv8-TX2,A64FX
 # IBM:      BGQ,Power7,Power8,Power9
-# AMD-GPUS: Vega900,Vega906,Vega908,Vega90A
+# AMD-GPUS: Vega906,Vega908,Vega90A,NAVI1030
```
dalg24, line 16: Any particular reason why to spell it `NAVI1030` instead of `Navi1030`?
I am asking because, why I don't necessary love how we have capitalized things in the past, this change would introduce more inconsistency.

cmake/kokkos_arch.cmake
```diff
@@ -93,9 +93,9 @@ IF(Kokkos_ENABLE_HIP OR Kokkos_ENABLE_OPENMPTARGET)
 ENDIF()
 
 # AMD archs ordered in decreasing priority of autodetection
-LIST(APPEND SUPPORTED_AMD_GPUS       MI200    MI100    MI50/60 MI25)
-LIST(APPEND SUPPORTED_AMD_ARCHS      VEGA90A  VEGA908  VEGA906 VEGA900)
-LIST(APPEND CORRESPONDING_AMD_FLAGS  gfx90a   gfx908   gfx906  gfx900)
+LIST(APPEND SUPPORTED_AMD_GPUS       MI200    MI100    MI50/60 V620/W6800)
+LIST(APPEND SUPPORTED_AMD_ARCHS      VEGA90A  VEGA908  VEGA906 NAVI1030)
+LIST(APPEND CORRESPONDING_AMD_FLAGS  gfx90a   gfx908   gfx906  gfx1030)
```
dalg24, line 98: Can't help myself noting that the indentation was using double space between columns and that this change has a single space.  Please fix the indentation.

core/src/impl/Kokkos_SharedAlloc.hpp
dalg24: @arghdos can you advise here?

core/src/HIP/Kokkos_HIP_Instance.hpp
```diff
@@ -56,9 +56,15 @@ namespace Kokkos {
 namespace Impl {
 
 struct HIPTraits {
+#if defined(KOKKOS_ARCH_VEGA)
   static int constexpr WarpSize       = 64;
   static int constexpr WarpIndexMask  = 0x003f; /* hexadecimal for 63 */
   static int constexpr WarpIndexShift = 6;      /* WarpSize == 1 << WarpShift*/
+#elif defined(KOKKOS_ARCH_NAVI)
+  static int constexpr WarpSize       = 32;
+  static int constexpr WarpIndexMask  = 0x001f; /* hexadecimal for 63 */
```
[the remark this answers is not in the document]
Rombur: It's hard to know what the mask looks like when you use a base-ten number.

arghdos:
Not sure why GH won't let me reply inline @dalg24 -- I have to check w/ some of the gaming folks (I don't have a Navi handy), will get back to you.  I don't think it will _hurt_ to leave it at 256B either way, just might waste some space at the beginning of an allocation.

arghdos:
Based on my reading, I'm pretty sure we want `256` here for Navi as well @dalg24

core/src/HIP/Kokkos_HIP_Instance.hpp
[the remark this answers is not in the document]
masterleinad: I find it also very hard to know what the mask looks like in hexadecimal format (that's why we need a comment) and would prefer binary format.

dalg24 (approved):
LGTM

dalg24:
> Based on my reading, I'm pretty sure we want `256` here for Navi as well @dalg24

Thank you for checking

Issue #5514, opened 2022-10-03
Trilinos nightly failure, OpenMP backend: simpleBuildAgainstTrilinos app builds against install fails with "Target "MyApp" links to target "OpenMP::OpenMP_CXX" but the target was not   found."


ndellingwood:
**Describe the bug**
Attempting to build the demo "app" `simpleBuildAgainstTrilinos` from Trilinos with an OpenMP build, gcc/8.3.0 as compiler fail to configuration with output:
```
CMake Error at CMakeLists.txt:53 (add_executable):
  Target "MyApp" links to target "OpenMP::OpenMP_CXX" but the target was not
  found.  Perhaps a find_package() call is missing for an IMPORTED target, or
  an ALIAS target is missing?
```

This does not occur with kokkos@3.7.00 snapshot into Trilinos, so I'm filing as a kokkos issue as this follows an apparent change in kokkos. 

**Please also include the following items to support reproducing the bug**
**Reproducer**:
trilinos/Trilinos@820615d4fa40f9fb72d27bd689b54797ea5ee7cb
kokkos/kokkos@2ce06b9754972818d8b023b5530732367ef28288
kokkos/kokkos-kernels@a438bcc72fb85fff36f32519c255960eee2bc804

```bash
# Repos
git clone -b develop kokkos
git clone -b develop kokkos-kernels
# Symbolic link to your kokkos and kokkos-kernels repos in Trilinos source directory for source override
cd Trilinos
ln -s <path-to-your-repo>/kokkos kokkos
ln -s <path-to-your-repo>/kokkos-kernels kokkos-kernels

# Trilinos configure
cd $HOME
mkdir -p build
cd build

module purge
module load openmpi/4.1.1/gcc/8.3.1/cuda/11.2.2 openblas/0.3.18/gcc/8.3.1 cmake/3.21.2
export OMPI_CXX=`which g++`
export OMPI_CC=`which gcc`
export OMPI_FC=`which gfortran`

cmake \
 -D CMAKE_INSTALL_PREFIX="$PWD/install" \
 -D CMAKE_CXX_COMPILER=`which mpicxx` \
 -D CMAKE_CXX_FLAGS="-g" \
 -D CMAKE_C_COMPILER=`which mpicc` \
 -D CMAKE_C_FLAGS="-g" \
 -D CMAKE_Fortran_COMPILER=`which mpifort` \
 -D CMAKE_Fortran_FLAGS="-g" \
[...]

PR #4105, opened 2021-06-18
Try using FindOpenMP instead of figuring out flags manually

masterleinad:
Trying to address #3256 by relying on `CMake`'s `FindOpenMP`.

PR #5526, opened 2022-10-06
Fix using OpenMP in code downstream from Trilinos

masterleinad:
Fixes https://github.com/kokkos/kokkos/issues/5514. As the comment says, the way we create `KokkosConfig.cmake` in `Trilinos` is very different from what we are doing in a standalone build. In particular, we didn't include https://github.com/kokkos/kokkos/blob/61d7db55fceac3318c987a291f77b844fd94c165/cmake/KokkosConfig.cmake.in#L10-L15 which is required for `OpenMP` after #4105.

This pull request just adds the required lines in case `Kokkos` was configured with `OpenMP` in the most straightforward way. Unfortunately, this requires some synchronization between `cmake/kokkos_install.cmake` and `cmake/KokkosConfig.cmake` in case we change anything with the way to detect and link to `OpenMP`. I still opted for this way to avoid a major refactor of the way `KokkosConfig.cmake` is created in `Trilinos` (which I would be afraid of doing anyway) and also in anticipation that `Kokkos` would be treated as a regular dependency in `Trilinos` in the not so far future.


masterleinad:
Superseded #5532.

Issue #5505, opened 2022-09-29
Kokkos does not export AMD flags to Trilinos


lucbv:
**Describe the bug**
When configuring Trilinos to build Kokkos with HIP enabled the `--offload-arch=gfx90N` flag is apparently not added to Trilinos' flags.
This leads to obvious compilation issues, setting the environment variable `ROCM_TARGET_LST` fixes the issue but defeats the point of specifying the ARCH to Kokkos at configure time.
Adding the following logic in CMakeList.txt at line 239 also fixes this issue:
```
  IF(KOKKOS_ENABLE_HIP)
    FOREACH(OPTION ${KOKKOS_AMDGPU_OPTIONS})
      STRING(FIND "${OPTION}" " " OPTION_HAS_WHITESPACE)
      IF(OPTION_HAS_WHITESPACE EQUAL -1)
        LIST(APPEND KOKKOS_COMPILE_OPTIONS_TMP "${OPTION}")
      ELSE()
        LIST(APPEND KOKKOS_COMPILE_OPTION_TMP "\"${OPTION}\"")
      ENDIF()
    ENDFOREACH()
  ENDIF()
```

jjellio:
Seems like this would get fixed if kokkos eliminated it's special case of HIPCC (which cmake hates anyway). #5453 
I'm guessing if you used the AMD compiler directly, Kokkos/Cmake would find ROCm (another huge headache) - but it would probably propagate the flags. (I haven't tested it)

lucbv:
In my case above the underlying issue is that I am compiling on a head node without GPUs where `rocm_agent_enumerator` does not find the target architecture. If I build on a compute node there no problem anymore.
The main issue is that Kokkos does not appear to tell Trilinos which target should be used based on the Kokkos_ARCH. My CMake attempt above works fine but is probably not the exact fix that Kokkos wants to implement?

jjellio:
[...]

PR #5528, opened 2022-10-06
Export the flags in KOKKOS_AMDGPU_OPTIONS when using Trilinos

Rombur:
With this PR, we export the flags stored in `KOKKOS_AMDGPU_OPTIONS` when using Trilinos.

Related to #5505


crtrott (approved):

CMakeLists.txt
```diff
@@ -228,6 +229,9 @@ IF (KOKKOS_HAS_TRILINOS)
   # we have to match the annoying behavior, also we have to preserve quotes
   # which needs another workaround.
   SET(KOKKOS_COMPILE_OPTIONS_TMP)
+  IF (KOKKOS_ENABLE_HIP)
+    LIST(APPEND KOKKOS_COMPILE_OPTIONS ${KOKKOS_AMDGPU_OPTIONS})
```
dalg24, line 233: Can you confirm that is the right thing to do
CUDA seems to be using `KOKKOS_ALL_COMPILE_OPTIONS`
https://github.com/kokkos/kokkos/blob/1a63570f2b74ff921e4b0dcb3f3d8e9cb4c0f05e/CMakeLists.txt#L246

crtrott: I think this has something to do with whether its ; separated list or not or so?

dalg24: I am asking because I saw https://github.com/kokkos/kokkos/blob/1a63570f2b74ff921e4b0dcb3f3d8e9cb4c0f05e/cmake/KokkosTrilinosConfig.cmake.in#L14
Should we comment then to make sure it does not look like a mistake?

crtrott: KOKKOS_COMPILER_OPTIONS is appended to KOKKOS_ALL_COMPILE_OPTIONS just a few lines below.

dalg24: OK I see https://github.com/kokkos/kokkos/blob/1a63570f2b74ff921e4b0dcb3f3d8e9cb4c0f05e/CMakeLists.txt#L244
I don't even want to look why we would need both `KOKKOS_ALL_COMPILE_OPTIONS` and `KOKKOS_COMPILER_OPTIONS `

dalg24 (approved):

masterleinad (commented):
Any reason not to just use `COMPILER_SPECIFIC_FLAGS`? That doesn't seem to require doing something special for `Trilinos`.

Rombur:
> Any reason not to just use COMPILER_SPECIFIC_FLAGS? That doesn't seem to require doing something special for Trilinos.

It does not work. I also though it would be cleaner but then I would need to rewrite more of the build system which I don't want to do.

PR #4618, opened 2021-12-15
Fix failing detection of view params bugfix

janciesko:
This is a stab at resolving the issue [View instantiation segfaults when instantiating with wrong number of args](https://github.com/kokkos/kokkos/issues/4561).

This PR moves the runtime_check_rank to the View constructor overload that calls into m_map.allocate_shared and places the check before m_map.allocate_shared but after m_track and m_map allocation. A view parameter is added that allows to turn the check on or off depending on the code path.  

Changes:
- Move runtime_check_rank to the allocating View constructor
- Initializes default dimensions to KOKKOS_IMPL_CTOR_DEFAULT_ARG in Kokkos_Layout (drive-by change)
- Add View param to disable param check for some code paths (see note below)
- PR includes the extra unit-test so we make sure this is doing what it we think it is doing

Note: Disabling runtime_check_rank for some code paths, e.g. TestViewLayoutStrideAssignment.hpp:638 is required and would fail otherwise:
[ RUN      ] cuda_DeathTest.view_layoutstride_right_to_layoutleft_assignment
view_layoutstride_right_to_layoutleft_assignment: srand(1639811970)
Constructor for Kokkos View 'LayoutStride_mirror' has mismatched number of arguments. Number of arguments = 8 but dynamic rank = 1

Issue #4722, opened 2022-01-26
Kokkos build failure with intel/19 and shared library support: unresolvable R_X86_64_TPOFF32 relocation against symbol


ndellingwood:
Kokkos builds with intel/19 and shared library support fail with the following type of output:

```
ld: CMakeFiles/KokkosCore_PerformanceTest_Mempool.dir/test_mempool.cpp.o(.text._ZN6Kokkos4Impl21ParallelReduceAdaptorINS_11RangePolicyIJNS_6SerialEN11TestFunctor15TagAllocDeallocEEEES4_lE7executeERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEERKS6_RKS4_Rl[_ZN6Kokkos4Impl21ParallelReduceAdaptorINS_11RangePolicyIJNS_6SerialEN11TestFunctor15TagAllocDeallocEEEES4_lE7executeERKNSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEERKS6_RKS4_Rl]+0x3e8): unresolvable R_X86_64_TPOFF32 relocation against symbol `_ZN6Kokkos4Impl22SharedAllocationRecordIvvE18t_tracking_enabledE'
ld: final link failed: Nonrepresentable section on output
make[2]: *** [core/perf_test/KokkosCore_PerformanceTest_Mempool] Error 1
make[2]: Leaving directory `/home/ndellin/kokkos/testing/Intel19'
make[1]: *** [core/perf_test/CMakeFiles/KokkosCore_PerformanceTest_Mempool.dir/all] Error 2
```

**Reproducer** (kokkos-dev* modules referenced):
```
module load sems-archive-cmake/3.17.1 sems-archive-gcc/8.3.0 sems-archive-intel/19.0.5

cmake -DKokkos_ENABLE_SERIAL=ON -DCMAKE_CXX_COMPILER=icpc -DBUILD_SHARED_LIBS=ON -DKokkos_ENABLE_TESTS=ON $KOKKOS_PATH
```

PhilMiller:
This also showed up in Trilinos, https://github.com/trilinos/Trilinos/pull/10047#issuecomment-1018921116

My digging online suggests a missing `-fPIC` on some compilation unit.

PhilMiller:
Or, conversely, there could be `-fPIE` on a file that's ending up in a shared library rather than directly in a binary

PhilMiller:
[...]

PR #5529, opened 2022-10-06
Always check rank in View construction 

masterleinad:
https://github.com/kokkos/kokkos/pull/4618 introduced checks for the rank in View constructors except that it disabled them for the constructor that takes a `view_alloc`  argument. After #5188 and a minor fix, this should not be necessary anymore and we can just get rid of the option.
Related to https://github.com/arborx/ArborX/pull/758.

### DIFF
--- a/core/src/Kokkos_CopyViews.hpp
+++ b/core/src/Kokkos_CopyViews.hpp
@@ -3468,11 +3468,8 @@ struct MirrorType {
 };
 
 template <class T, class... P, class... ViewCtorArgs>
-inline std::enable_if_t<
-    !std::is_same<typename Kokkos::ViewTraits<T, P...>::array_layout,
-                  Kokkos::LayoutStride>::value &&
-        !Impl::ViewCtorProp<ViewCtorArgs...>::has_memory_space,
-    typename Kokkos::View<T, P...>::HostMirror>
+inline std::enable_if_t<!Impl::ViewCtorProp<ViewCtorArgs...>::has_memory_space,
+                        typename Kokkos::View<T, P...>::HostMirror>
 create_mirror(const Kokkos::View<T, P...>& src,
               const Impl::ViewCtorProp<ViewCtorArgs...>& arg_prop) {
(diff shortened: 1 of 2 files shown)


core/src/Kokkos_CopyViews.hpp
```diff
@@ -3533,24 +3533,10 @@ create_mirror(const Kokkos::View<T, P...>& src,
...
-  layout.stride[2] = src.stride_2();
-  layout.stride[3] = src.stride_3();
-  layout.stride[4] = src.stride_4();
-  layout.stride[5] = src.stride_5();
-  layout.stride[6] = src.stride_6();
-  layout.stride[7] = src.stride_7();
+  for (unsigned int i = 0; i < src_type::rank; ++i) {
+    layout.dimension[i] = src.extent(i);
+    layout.stride[i]    = src.stride(i);
+  }
```
PhilMiller, line 3539: Would there be something wrong with just `Kokkos::LayoutStride layout = src.layout();`, or just passing `src.layout()` as an argument and forgoing the copy? We know from the `enable_if` that the source layout is `LayoutStride`

masterleinad: No, I think we don't even need to specialize for `LayoutStride`.

dalg24 (approved):

janciesko:
This looks good to me. I believe rebasing this PR vs this #4721 is slightly less work, so I'd prefer to merge #4721 first in case both PRs start to pass CI.

janciesko (approved):

dalg24:
> This looks good to me. I believe rebasing this PR vs this #4721 is slightly less work, so I'd prefer to merge #4721 first in case both PRs start to pass CI.

I thought about it and
* This PR is ready and #4722 is not
* Merging this one first will reduce the diff in #4722 
* The conflicts in `Kokkos_CopyViews.hpp` are manageable

Therefore I am merging this one first

dalg24:
I took resolved the conflicts in the other PR

PR #5532, opened 2022-10-07
Trilinos: Pass OpenMP flags instead of linking with the OpenMP target

masterleinad:
Alternative to #5526 fixing #5514. We discovered that properly linking with the `OpenMP` target in `Trilinos` requires more changes in `Trilinos` itself (and it's not clear to me what the best path there is), also see https://github.com/kokkos/kokkos/issues/5514#issuecomment-1270644310. Those changes would include forcing `Trilinos_ENABLE_OpenMP` for `Kokkos_ENABLE_OPENMP` and enabling `TriBITS`' `CMake` flag detection for `Kokkos` again, see https://github.com/trilinos/Trilinos/blob/master/cmake/ProjectCompilerPostConfig.cmake and https://github.com/kokkos/kokkos/issues/5514#issuecomment-1270848852. These changes might break downstream `Trilinos` users.

Instead, this pull request just tries to extract the `OpenMP` flags similar to what we did before, see https://github.com/kokkos/kokkos/pull/4105/files, and what `TriBITS` does anyway, see https://github.com/trilinos/Trilinos/blob/fef3ccfae5e6cc865aad5025d7d8f99740aa5aa8/cmake/tribits/core/package_arch/TribitsGlobalMacros.cmake#L1685-L1700.

I couldn't restrict the changes to `core/src/CMakeLists.txt` since adding to the compiler flags there is too late to be recognized.

### DIFF
--- a/core/src/CMakeLists.txt
+++ b/core/src/CMakeLists.txt
@@ -170,13 +170,13 @@ IF (Kokkos_ENABLE_IMPL_DESUL_ATOMICS AND KOKKOS_ENABLE_OPENMPTARGET)
   target_link_libraries(kokkoscore PUBLIC atomic)
 ENDIF()
 
-# FIXME_TRILINOS Trilinos doesn't allow for Kokkos to use find_dependency so we
-# just append the flags in cmake/kokkos_tpls.cmake instead of linking with the
-# OpenMP target.
 IF (Kokkos_ENABLE_IMPL_DESUL_ATOMICS AND desul_FOUND)
   target_link_libraries(kokkoscore PUBLIC desul_atomics)
 ENDIF()
 
+# FIXME_TRILINOS Trilinos doesn't allow for Kokkos to use find_dependency so we
+# just append the flags in cmake/kokkos_tpls.cmake instead of linking with the
+# OpenMP target.
 IF(Kokkos_ENABLE_OPENMP AND NOT KOKKOS_HAS_TRILINOS)
   target_link_libraries(kokkoscore PUBLIC OpenMP::OpenMP_CXX)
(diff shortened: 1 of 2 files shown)


ndellingwood:
@masterleinad with these changes I am able to successfully configure, build and tests pass with the demo app from Trilinos in an OpenMP build :)

ndellingwood (approved):

ndellingwood:
CI failures are known issues with the clang++ fedora builds (unrelated to this PR)

core/src/CMakeLists.txt
```diff
@@ -170,11 +170,14 @@ IF (Kokkos_ENABLE_IMPL_DESUL_ATOMICS AND KOKKOS_ENABLE_OPENMPTARGET)
   target_link_libraries(kokkoscore PUBLIC atomic)
 ENDIF()
 
+# FIXME_TRILINOS Trilinos doesn't allow for Kokkos to use find_dependency so we
+# just append the flags in cmake/kokkos_tpls.cmake instead of linking with the
+# OpenMP target.
```
Rombur, line 175: Why is comment above the DESUL code and not the OpenMP code?

masterleinad: Moved it.

crtrott:
@ndellingwood @masterleinad is this something we need in 3.7.01?

masterleinad:
> @ndellingwood @masterleinad is this something we need in 3.7.01?

No, https://github.com/kokkos/kokkos/pull/4105 changed the way we detect `OpenMP` and it's not in 3.7.00 but only in `develop`.

Issue #5534, opened 2022-10-07
View assignment 1D mismatched layout with static extents fails to compile


crtrott:
```
#include <Kokkos_Core.hpp>
#include <cmath>

int main(int argc, char* argv[]) {
  Kokkos::initialize(argc, argv);
  {
    int N = argc > 1 ? atoi(argv[1]) : 1000000;
    int R = argc > 2 ? atoi(argv[2]) : 10;

    Kokkos::View<double[5],Kokkos::LayoutLeft> a("A");
    Kokkos::View<double[5],Kokkos::LayoutRight> b("B");
    a = b;
    b = a;

  }
  Kokkos::finalize();
}
```
```
/ascldap/users/crtrott/Kokkos/kokkos/core/src/Kokkos_View.hpp:1351:28: error: static assertion failed: Incompatible View copy assignment
 1351 |     static_assert(Mapping::is_assignable, "Incompatible View copy assignment");
```

crtrott:
Failed forever ... (at least back to 3.0)

crtrott:
Hoping this is easy to fix .. @lucbv ran into this.

PhilMiller:
I can totally see why that assignment should be workable, but is it something that actually should be allowed?

Is there an implicit statement in this report that the corresponding case with dynamic extents does compile, and works as expected, and is the behavior we want to expand?

crtrott:
yeah I found the places already. And yes this should work, mdspan does allow it.

PhilMiller:
Ok, I guess I shouldn't go off and derail mdspan standardization by requestioning this :-D

crtrott:
It also works already for matching layouts in Kokkos View.

crtrott:
And the actual part in the code which is responsible for checking extents compatibility already did the right thing.

PhilMiller:
Given that this is long standing behavior, shipping a change in a patch release is nevertheless iffy, even if it's safe

crtrott:
yeah but it only enables stuff which didn't compile before, i.e. very unlikely to break any code. Note that the error is a static assert not an SFINAE can't find the overload, so it shouldn't mess with overload sets.

PhilMiller:
Fixed in 3.7.01 with #5566

PR #5535, opened 2022-10-07
Fix static extents assignment for LayoutLeft/LayoutRight assignment

crtrott:
We always supported the following:
```c++
Kokkos::View<int[5]> a("A");
Kokkos::View<int[5]> b("A");
Kokkos::View<int*> c("C",5);
a = b;
b = a;
b = c;
```

We also supported for always:
```c++
Kokkos::View<int*,Kokkos::LayoutLeft> a("A",5);
Kokkos::View<int*,Kokkos::LayoutRight> b("B",5);
a = b;
b = a;
```

What apparently doesn't work is:
```c++
Kokkos::View<int[5],Kokkos::LayoutLeft> a("A");
Kokkos::View<int[5],Kokkos::LayoutRight> b("B");
a = b;
b = a;
```

Turns out we do some extra checks for dynamic rank match, when accounting for LayoutLeft/LayoutRight convertibility of Rank 0 and Rank 1 Views.
That check is unnecessary, since it will get checked via the dimension assignability check already.
I.e. we hadn't properly orthogonalized the extents assignability from the Layout assignability for mixing LayoutLeft/LayoutRight.

This fixes this by getting rid of the erroneous condition.

### DIFF
--- a/core/src/impl/Kokkos_ViewMapping.hpp
+++ b/core/src/impl/Kokkos_ViewMapping.hpp
@@ -1128,9 +1128,8 @@ struct ViewOffset<
   KOKKOS_INLINE_FUNCTION constexpr ViewOffset(
       const ViewOffset<DimRHS, Kokkos::LayoutRight, void>& rhs)
       : m_dim(rhs.m_dim.N0, 0, 0, 0, 0, 0, 0, 0) {
-    static_assert((DimRHS::rank == 0 && dimension_type::rank == 0) ||
-                      (DimRHS::rank == 1 && dimension_type::rank == 1 &&
-                       dimension_type::rank_dynamic == 1),
+    static_assert(((DimRHS::rank == 0 && dimension_type::rank == 0) ||
+                   (DimRHS::rank == 1 && dimension_type::rank == 1)),
                   "ViewOffset LayoutLeft and LayoutRight are only compatible "
                   "when rank <= 1");
   }
@@ -1778,8 +1777,7 @@ struct ViewOffset<
       const ViewOffset<DimRHS, Kokkos::LayoutLeft, void>& rhs)
       : m_dim(rhs.m_dim.N0, 0, 0, 0, 0, 0, 0, 0) {
     static_assert((DimRHS::rank == 0 && dimension_type::rank == 0) ||
-                      (DimRHS::rank == 1 && dimension_type::rank == 1 &&
-                       dimension_type::rank_dynamic == 1),
+                      (DimRHS::rank == 1 && dimension_type::rank == 1),
                   "ViewOffset LayoutRight and LayoutLeft are only compatible "
                   "when rank <= 1");
   }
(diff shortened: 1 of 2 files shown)


crtrott:
Note the proper dimension check happens around line 281 or so.

crtrott:
Fixes #5534 @lucbv

core/src/impl/Kokkos_ViewMapping.hpp
```diff
@@ -1128,9 +1128,8 @@ struct ViewOffset<
   KOKKOS_INLINE_FUNCTION constexpr ViewOffset(
       const ViewOffset<DimRHS, Kokkos::LayoutRight, void>& rhs)
       : m_dim(rhs.m_dim.N0, 0, 0, 0, 0, 0, 0, 0) {
-    static_assert((DimRHS::rank == 0 && dimension_type::rank == 0) ||
-                      (DimRHS::rank == 1 && dimension_type::rank == 1 &&
-                       dimension_type::rank_dynamic == 1),
```
crtrott, line 1133: That line 1133 is the offending check.

crtrott:
Note this behavior was always intended (as the fact that it works for matching layouts attest too) and this is also how mdspan works.

lucbv (approved):
I talked this through with @crtrott and agree with the logic

core/unit_test/TestViewIsAssignable.hpp
```diff
@@ -94,6 +94,14 @@ TEST(TEST_CATEGORY, view_is_assignable) {
   // Layout assignment
   Impl::TestAssignability<View<int*, left, d_exec>,
                           View<int*, right, d_exec>>::test(true, true, 10);
+  Impl::TestAssignability<View<int[5], left, d_exec>,
+                          View<int*, right, d_exec>>::test(false, false, 10);
+  Impl::TestAssignability<View<int[10], left, d_exec>,
+                          View<int*, right, d_exec>>::test(false, true, 10);
+  Impl::TestAssignability<View<int*, left, d_exec>,
+                          View<int[5], right, d_exec>>::test(true, true, 5);
```
dalg24, line 102: Wait why do you pass `5`?  Isn't it something we only relaxed not so long ago?  https://github.com/kokkos/kokkos-core-wiki/pull/108
Please drop it here and below.

dalg24 (commented):
The view assignable unit test could use some refactor (not requesting changes just saying)
Your changes should enable rank-0 view assignments as well isn't it?  If I am not mistake please add tests for it.

core/unit_test/TestViewIsAssignable.hpp
crtrott: yeah unnescessary here, gonna change that.

crtrott: Also added the rank-0 mixed layout case.

core/unit_test/TestViewIsAssignable.hpp
```diff
@@ -92,8 +92,18 @@ TEST(TEST_CATEGORY, view_is_assignable) {
                           View<double*, left, d_exec>>::test(false, false, 10);
 
   // Layout assignment
+  Impl::TestAssignability<View<int, left, d_exec>,
+                          View<int, right, d_exec>>::test(true, true);
```
dalg24, line 96: No right -> left?

dalg24 (approved):

core/unit_test/TestViewIsAssignable.hpp
crtrott: I guess we could add them everywhere, but mostly this is about mismatch, not any specific layouts.

PR #5538, opened 2022-10-09
Add Hopper support

crtrott:
Addresses issue #5524


dalg24 (changes requested):
https://github.com/kokkos/kokkos/blob/61d7db55fceac3318c987a291f77b844fd94c165/core/src/Cuda/Kokkos_Cuda_BlockSize_Deduction.hpp#L218-L232

crtrott:
Added the shared config, and also the printconfig thing. Confirmed in tuning guide that it also can do 8kB shared memory.

masterleinad (commented):
What about `cmake/compile_tests/cuda_compute_capability.cc`?

crtrott:
Fixed: also now actually checked for each use of ARCH_AMPERE and ARCH_VOLTA in our code base and made the appropriate adjustments.

masterleinad (approved):

crtrott:
Fixed the logic mistake in the half precision thing.

dalg24 (approved):

Issue #5489, opened 2022-09-22
HIP Link issue


crtrott:
Using: 

Configure with ROCM 5.2.0:
```
cmake -DKokkos_ENABLE_HIP=ON -DKokkos_ARCH_VEGA908=ON -DKokkos_ENABLE_TESTS=ON ../
```

```
ld.lld: error: undefined symbol: __truncdfhf2
>>> referenced by amd_hip_fp16.h:101 (/home/projects/ROCm/rocm-5.2.0/include/hip/amd_detail/amd_hip_fp16.h:101)
>>>               CMakeFiles/KokkosCore_UnitTest_Default.dir/default/TestDefaultDeviceType.cpp.o:(Test::Functor_TestHalfOperators<Kokkos::View<double*, Kokkos::HostSpace>, Kokkos::Experimental::Impl::floating_point_wrapper<__half> >::operator()(int) const)
>>> referenced by amd_hip_fp16.h:101 (/home/projects/ROCm/rocm-5.2.0/include/hip/amd_detail/amd_hip_fp16.h:101)
>>>               CMakeFiles/KokkosCore_UnitTest_Default.dir/default/TestDefaultDeviceType.cpp.o:(Test::Functor_TestHalfOperators<Kokkos::View<double*, Kokkos::HostSpace>, Kokkos::Experimental::Impl::floating_point_wrapper<__half> >::operator()(int) const)
>>> referenced by amd_hip_fp16.h:101 (/home/projects/ROCm/rocm-5.2.0/include/hip/amd_detail/amd_hip_fp16.h:101)
>>>               CMakeFiles/KokkosCore_UnitTest_Default.dir/default/TestDefaultDeviceType.cpp.o:(Test::Functor_TestHalfOperators<Kokkos::View<double*, Kokkos::HostSpace>, Kokkos::Experimental::Impl::floating_point_wrapper<__half> >::operator()(int) const)
>>> referenced 1 more times
clang-14: error: linker command failed with exit code 1 (use -v to see invocation)
```

crtrott:
This is also noted in the crusher known issues section:

https://docs.olcf.ornl.gov/systems/crusher_quick_start_guide.html#olcfdev-1167-kokkos-build-failures-with-prgenv-amd

crtrott:
@Rombur

Rombur:
Yeah, I know. It's on my TODO list.

jjellio:
[...]

PR #5539, opened 2022-10-10
Fix linking when using amdclang

Rombur:
With this PR I can compile and run the tests on Crusher using `amdclang` instead of `hipcc`

Related to #5489

### DIFF
--- a/cmake/Modules/FindTPLROCM.cmake
+++ b/cmake/Modules/FindTPLROCM.cmake
@@ -3,9 +3,19 @@ include(FindPackageHandleStandardArgs)
 FIND_LIBRARY(AMD_HIP_LIBRARY amdhip64 PATHS ENV ROCM_PATH PATH_SUFFIXES lib)
 FIND_LIBRARY(HSA_RUNTIME_LIBRARY hsa-runtime64 PATHS ENV ROCM_PATH PATH_SUFFIXES lib)
 
-find_package_handle_standard_args(TPLROCM DEFAULT_MSG AMD_HIP_LIBRARY HSA_RUNTIME_LIBRARY)
+# clang_rt.builtins is necessary to use half precision. The following code to
+# find clang_rt.buitins is based on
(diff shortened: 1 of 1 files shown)


dalg24 (approved):

masterleinad (approved):
Looks reasonable.

masterleinad:
```
./scripts/docker/check_format_cpp.sh
awk: read error (Is a directory)
diff --git a/cmake/Modules/FindTPLROCM.cmake b/cmake/Modules/FindTPLROCM.cmake
index 792223e3f..aacdfcaf1 100644
--- a/cmake/Modules/FindTPLROCM.cmake
+++ b/cmake/Modules/FindTPLROCM.cmake
@@ -7,8 +7,8 @@ FIND_LIBRARY(HSA_RUNTIME_LIBRARY hsa-runtime64 PATHS ENV ROCM_PATH PATH_SUFFIXES
 # find clang_rt.buitins is based on
 # https://github.com/ROCm-Developer-Tools/HIP/blob/develop/hip-lang-config.cmake.in#L99-L111
 file(GLOB_RECURSE CLANG_RT_DIR "$ENV{ROCM_PATH}/llvm/lib/clang/*/lib/*/*clang_rt.builtins*")
-FIND_LIBRARY(CLANG_RT_LIBRARY 
-  NAMES 
+FIND_LIBRARY(CLANG_RT_LIBRARY
+  NAMES
   clang_rt.builtins
   clang_rt.builtins-x86_64
   PATHS "${CLANG_RT_DIR}/..")
script returned exit code 1
```

crtrott:
bloody hell. Do we have a report to AMD?

Rombur:
> bloody hell. Do we have a report to AMD?

I have asked Nick about it. Here is his response 

amdclang does not automatically link against the clang_rt for you, this is theoretically handled by the find_package(HIP) in CMake but that's often been broken due to the silly relative pathing scheme they use in there (work is in progress to use LLVM's exported CMake targets to make this less prone to failure),

Issue #5475, opened 2022-09-20
Team parallel_reduce may need an impl-side team_barrier between adjacent reductions.


ambrad:
In possibly just non-scalar-accumulator cases, adjacent team parallel_reduce calls with no application-side data dependencies and thus no application-side team_barrier can lead to the wrong answer nondeterministically. This implies that there is an impl-side data dependency in bookkeeping.

Following is a reproducer. The comments at the top show the machine, modules, Kokkos version, config, build, and run instructions.

```C++
/*
  Reproducer for the following issue: Application calls two team reductions in a
  row on array accumulators. Application side has no data dependencies between
  reductions and thus does not call team_barrier between reductions.
  Nonetheless, the accumulators do not always have the right value, implying
  there is an implementation-side data dependency.

  On Summit:

  $ cat source.sh
  module purge
  module load gcc/9.1.0 cuda/11.0.3 cmake/3.18.4
  export NVCC_WRAPPER_DEFAULT_COMPILER=g++

  $ git clone git@github.com:kokkos/kokkos.git
  $ cd kokkos
  $ git log --oneline -n 1
  d19aab998 (HEAD -> master, tag: 3.6.01, origin/master, origin/HEAD) Merge pull request #5051 from kokkos/release-candidate-3.6.01
  $ cd ../build-kokkos
  $ cat config.sh
  cmake \
      -D CMAKE_CXX_COMPILER=~/summit/reduction-barrier-reproducer/kokkos/bin/nvcc_wrapper \
      -D CMAKE_BUILD_TYPE=RELEASE \
      -D Kokkos_ENABLE_SERIAL=ON \
      -D Kokkos_ARCH_VOLTA70=ON \
      -D Kokkos_ENABLE_CUDA=ON \
      -D Kokkos_ENABLE_CUDA_LAMBDA=On \
      -D Kokkos_ENABLE_CUDA_CONSTEXPR=On \
      -D CMAKE_INSTALL_PREFIX=../kokkos-install \
      ../kokkos
  $ bash config.sh; make -j4 install
  
  $ cd ..
  $ K=./kokkos-install; $K/bin/nvcc_wrapper -O3 -arch=sm_70 --expt-extended-lambda -pedantic -Wall reduction_barrier_reproducer.cpp -I$K/include -L$K/lib64 -lkokkoscore
  $ ./a.out  # should fail
 */

#include "Kokkos_Core.hpp"

struct Accum {
[...]

PR #5540, opened 2022-10-10
Fix repeated team_reduce without barrier

masterleinad:
Fixes https://github.com/kokkos/kokkos/issues/5475.
`cuda_inter_warp_reduction` uses shared memory and we need to ensure that we are done using it before exiting the function. Otherwise, repeated calls to team_reduce might fail as the test demonstrates.

### DIFF
--- a/core/unit_test/TestTeam.hpp
+++ b/core/unit_test/TestTeam.hpp
@@ -1664,6 +1664,7 @@ struct TestRepeatedTeamReduce {
     // Choose a non-recommened (non-power of two for GPUs) team size
     int team_size = team_size_recommended > 1 ? team_size_recommended - 1 : 1;
 
+    // The failure was non-deterministic so run the test a bunch of times
     for (int it = 0; it < 100; ++it) {
       Kokkos::parallel_for(
           Kokkos::TeamPolicy<ExecutionSpace>(ncol, team_size, 1), *this);
--- a/core/unit_test/TestTeam.hpp
+++ b/core/unit_test/TestTeam.hpp
@@ -1629,6 +1629,8 @@ struct TestRepeatedTeamReduce {
     constexpr auto pi  = Kokkos::numbers::pi;
     double b           = 0.;
     for (int ri = 0; ri < 10; ++ri) {
+      // The contributions here must be sufficiently complex, simply adding ones
+      // wasn't enough to trigger the bug.
       const auto g1 = [&](const int k, double &acc) {
         acc += Kokkos::cos(pi * double(k) / nlev);
       };
--- a/core/unit_test/TestTeam.hpp
+++ b/core/unit_test/TestTeam.hpp
@@ -1616,6 +1616,69 @@ struct TestTeamPolicyHandleByValue {
 
 }  // namespace
 
+namespace {
+template <typename ExecutionSpace>
+struct TestRepeatedTeamReduce {
+  using TeamPolicy          = Kokkos::TeamPolicy<ExecutionSpace>;
+  static constexpr int ncol = 1500;  // nothing special, just some work
+
+  KOKKOS_FUNCTION void operator()(
+      const typename TeamPolicy::member_type &team) const {
+    // non-divisible by power of two to make triggering problems easier
+    constexpr int nlev = 129;
+    constexpr auto pi  = Kokkos::numbers::pi;
+    double b           = 0.;
+    for (int ri = 0; ri < 10; ++ri) {
+      const auto g1 = [&](const int k, double &acc) {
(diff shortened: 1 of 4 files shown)


core/unit_test/TestTeam.hpp
```diff
@@ -1616,6 +1616,69 @@ struct TestTeamPolicyHandleByValue {
...
+      KOKKOS_IMPL_DO_NOT_USE_PRINTF("Failing at %d!\n", i);
+    }
+  }
+
+  TestRepeatedTeamReduce() : v("v", ncol) { test(); }
+
+  void test() {
+    int team_size_recommended =
+        TeamPolicy(1, 1).team_size_recommended(*this, Kokkos::ParallelForTag());
+    // Choose a non-recommened (non-power of two for GPUs) team size
```
nmm0, line 1663: Is this also more likely to trigger the bug?

nmm0 (approved):
Looks good to me!

core/unit_test/TestTeam.hpp
masterleinad: I would think so. The original test case used `nlev` (129) also for the team size but we can't really do that across all backends. I just checked that this version would also trigger the bug for me.

ldh4 (approved):

core/unit_test/TestTeam.hpp
```diff
@@ -1616,6 +1616,69 @@ struct TestTeamPolicyHandleByValue {
...
+
+  KOKKOS_FUNCTION void operator()(
+      const typename TeamPolicy::member_type &team) const {
+    // non-divisible by power of two to make triggering problems easier
+    constexpr int nlev = 129;
+    constexpr auto pi  = Kokkos::numbers::pi;
+    double b           = 0.;
+    for (int ri = 0; ri < 10; ++ri) {
+      const auto g1 = [&](const int k, double &acc) {
+        acc += Kokkos::cos(pi * double(k) / nlev);
```
dalg24, line 1633: I think that using trigonometric functions is unnecessary complexity in that test

core/unit_test/TestTeam.hpp
```diff
@@ -1616,6 +1616,69 @@ struct TestTeamPolicyHandleByValue {
...
+
+  TestRepeatedTeamReduce() : v("v", ncol) { test(); }
+
+  void test() {
+    int team_size_recommended =
+        TeamPolicy(1, 1).team_size_recommended(*this, Kokkos::ParallelForTag());
+    // Choose a non-recommened (non-power of two for GPUs) team size
+    int team_size = team_size_recommended > 1 ? team_size_recommended - 1 : 1;
+
+    for (int it = 0; it < 1000; ++it) {
```
dalg24, line 1666: Why do we repeat 1000x ?

masterleinad: We need to repeat a bunch of times since the failure was non-deterministic. I guess we could go to 100, though.

dalg24: Make it 100 and comment then

core/unit_test/TestTeam.hpp
masterleinad: I would prefer to keep it since it was in the original reproducer unless we have a better idea.

dalg24: Are you not able to trigger the issue with a simple reduction that sums ones?

crtrott: I could imagine that the cos call stuff makes the threads get out of sync, because they funnel through limited special function units?

masterleinad: I can't reproduce the failure with something like
```
    const auto g1 = [&](const int k, double &acc) {
        acc += 1. / nlev;
      };
      const auto g2 = [&](const int k, double &acc) {
        acc += .5 / nlev;
      };
      double a1, a2;
      Kokkos::parallel_reduce(Kokkos::TeamThreadRange(team, nlev), g1, a1);
      Kokkos::parallel_reduce(Kokkos::TeamThreadRange(team, 2*nlev), g2, a2);
```

dalg24: Please add a comment to that effect in the code

crtrott: Thanks for checking!

masterleinad: [`087d767` (#5540)](https://github.com/kokkos/kokkos/pull/5540/commits/087d767855d571dd42a8818c67b0fac2d4862158)

PhilMiller (approved):

PR #5541, opened 2022-10-11
Min max pragma push

### LINKED ISSUE #5446: Kokkos and `#include <windows.h>`
The problem is that as soon as we want to use `min` and `max` as function names (e.g. in `Kokkos_NumericTraits.hpp`) we get compilation errors. But a user (or in my case a testcase) might want to include `windows.h`. This creates an interesting problem:

`min` and `max` are defined in `windows.h` as preproc macros. Thus, as soon as someone wants to include `windows.h` in code that also uses kokkos, he would have to define `NOMINMAX` before the include. This prevents `windows.h` from defining the macros. We are basically failing to export this constraint that we add by using the names `min` and `max`.
As failing to define `NOMINMAX` mixes macros with namespace stuff there are nasty compile errors happening far away from the point that needs the `NOMINMAX` fix. 

I can see the following ways we could tackle this (please comment if you know more):

- define `NOMINMAX` in Kokkos and document, that `Kokkos_Core.hpp` has to be included before any`windows.h` include
- just document that users have to do the `NOMINMAX` defines themselfs ...
- export `DNOMINMAX` in our compile commands
- make every call to `min` or `max` in kokkos not being translated by the macro by doing `(min)(x,y)`

JBludau:
Fix #5446 by trying to `pragma_push` the macros `min` and `max` when entering `Kokkos_Core.hpp` and redefining with `pragma_pop` at end. The macros being defined in e.g. `windows.h` lead to syntax errors otherwise. I hope this fixes the problem

### DIFF
--- a/core/src/Kokkos_Core.hpp
+++ b/core/src/Kokkos_Core.hpp
@@ -50,6 +50,22 @@
 #endif
 
 //----------------------------------------------------------------------------
+// In the case windows.h is included before Kokkos_Core.hpp there might be
+// errors due to the potentially defined macros with name "min" and "max" in
+// windows.h. These collide with the use of "min" and "max" in names inside
+// Kokkos. The macros will be redefined at the end of Kokkos_Core.hpp
+#if defined(min)
+#pragma push_macro("min")
+#undef min
+#define KOKKOS_PUSH_MACRO_MIN
+#endif
+#if defined(max)
+#pragma push_macro("max")
+#undef max
+#define KOKKOS_PUSH_MACRO_MAX
+#endif
+
+//----------------------------------------------------------------------------
 // Include the execution space header files for the enabled execution spaces.
 
 #include <Kokkos_Core_fwd.hpp>
@@ -312,6 +328,16 @@ std::vector<ExecSpace> partition_space(ExecSpace space,
 
 // Specializations required after core definitions
 #include <KokkosCore_Config_PostInclude.hpp>
+
+//----------------------------------------------------------------------------
+// Redefinition of the macros min and max if we pushed them at entry of
+// Kokkos_Core.hpp
+#if defined(KOKKOS_PUSH_MACRO_MIN)
+#pragma pop_macro("min")
+#endif
+#if defined(KOKKOS_PUSH_MACRO_MAX)
+#pragma pop_macro("max")
+#endif
 
 //----------------------------------------------------------------------------
 //----------------------------------------------------------------------------
--- a/core/src/Kokkos_Core.hpp
+++ b/core/src/Kokkos_Core.hpp
@@ -57,12 +57,12 @@
 #if defined(min)
 #pragma push_macro("min")
 #undef min
-#define KOKKOS_PUSH_MACRO_MIN
+#define KOKKOS_IMPL_PUSH_MACRO_MIN
 #endif
 #if defined(max)
 #pragma push_macro("max")
 #undef max
-#define KOKKOS_PUSH_MACRO_MAX
+#define KOKKOS_IMPL_PUSH_MACRO_MAX
 #endif
 
 //----------------------------------------------------------------------------
@@ -332,13 +332,13 @@ std::vector<ExecSpace> partition_space(ExecSpace space,
 //----------------------------------------------------------------------------
 // Redefinition of the macros min and max if we pushed them at entry of
 // Kokkos_Core.hpp
-#if defined(KOKKOS_PUSH_MACRO_MIN)
+#if defined(KOKKOS_IMPL_PUSH_MACRO_MIN)
 #pragma pop_macro("min")
-#undef KOKKOS_PUSH_MACRO_MIN
+#undef KOKKOS_IMPL_PUSH_MACRO_MIN
 #endif
-#if defined(KOKKOS_PUSH_MACRO_MAX)
+#if defined(KOKKOS_IMPL_PUSH_MACRO_MAX)
 #pragma pop_macro("max")
-#undef KOKKOS_PUSH_MACRO_MAX
+#undef KOKKOS_IMPL_PUSH_MACRO_MAX
 #endif
 
 //----------------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


PhilMiller:
You still need to call them like `(max)()` at the call site

PhilMiller (approved):

core/src/Kokkos_Core.hpp
```diff
@@ -49,6 +49,22 @@
...
+#if defined(min)
+#pragma push_macro("min")
+#undef min
+#define KOKKOS_IMPL_PUSH_MACRO_MIN
+#endif
+#if defined(max)
+#pragma push_macro("max")
+#undef max
+#define KOKKOS_IMPL_PUSH_MACRO_MAX
+#endif
```
masterleinad, line 66: Let's only do that when we could possibly use the windows runtime, i.e, guard it with `#if defined(_WIN32)`.

JBludau: My reasoning was, that the problem is not windows specific but just becomes painfully apparent when including `windows.h`. Now we don't want to guard against any name being used by us ... but is there a good reason to restrict it to windows only?

masterleinad: We are dealing with a lot of different compilers and `push_macro` is not standardized (even though it's supported by many compilers). That's why I would rather want to be careful and avoid using it without checking that the compiler actually supports it.

JBludau: makes sense, but it checks if `min` or `max` are defined and only then does the push. If we can not push the defined macros the compiler will fail anyways later. With the given implementation it will at least point to a line that actually explains what is wrong (the macro being defined).

PhilMiller: I'm initiall inclined to disagree - anything that's defined macros for those common function names needs to be protected against.

On the other hand, if we see such definitions without `defined _WIN32`, then perhaps we should just error out explicitly, because we're on an even weirder platform, or the user is doing something that will screw us up

PhilMiller: On the flip side, we could just guard all of the definitions and uses inside Kokkos with parentheses, like HPX did, and not mess around with macros, whether they're present or not.

JBludau: if the user wants to define something like `min` or `max` macro why should he not be allowed to use it? My intent was just to protect all our definitions in `Kokkos_Core.hpp` to not suffer from the problem without making a ton of changes by putting `()` everywhere. For future development workflow I think it is nice to not have to remember to put `()`.

dalg24 (commented):
I am not sure what the intent of this PR is.  What it protect from (including `<windows.h>` before `<Kokkos_Core.hpp>`?) and what are the limits of this approach (I can only assume Kokkos will be fine but not user code)?
If we really care if probably should make sure one can also include `<Kokkos_MathematicalFunctions>` (or any other public header) in the same conditions as `<Kokkos_Core.hpp>`.
Lastly, how would we go about testing that the changes proposed here achieve their goal.

JBludau:
> I am not sure what the intent of this PR is. What it protect from (including `<windows.h>` before `<Kokkos_Core.hpp>`?) and what are the limits of this approach (I can only assume Kokkos will be fine but not user code)?
Yeah, any code that includes `windows.h` and does not define `NOMINMAX` will define the macros `min` and `max`. If these are visible in `Kokkos_Core.hpp` we get syntax errors that do not immediately make sense or point to the problem.
The limits are of course, that we can only protect our stuff, users have to "sanitize" their code (e.g. putting `()` or pushing and poping the macros). I can see no way of us doing it for them in their code here.

> If we really care if probably should make sure one can also include `<Kokkos_MathematicalFunctions>` (or any other public header) in the same conditions as `<Kokkos_Core.hpp>`. 

Yeah, we would need to put this for all our public headers. Do we have a list of what is public?

> Lastly, how would we go about testing that the changes proposed here achieve their goal.
If we see a unit test/ perf test that we compile as an external application, we can just do it there. The two tests that ran into exactly this problem (even with `Kokkos_Core.hpp` included before `windows.h`, which I still don't know why) are fixed in this pr (meaning, they now do not use NOMINMAX, thus having the macros defined)

JBludau:
I added a compile only test that shows and tests if we can include `Kokkos_Core.hpp` after `windows.h` without the `NOMINMAX` define. 
As this pr so far only guards `Kokkos_Core.hpp` the test is a separate executable. The other compile only target does include subsets of kokkos and this way they do not get f'd over.
Nevertheless, we should decide which headers are unofficially allowed to be included and think about guarding them too.

core/unit_test/CMakeLists.txt
```diff
@@ -90,6 +90,15 @@ KOKKOS_ADD_EXECUTABLE(
...
 )
 
+if (WIN32)
+  KOKKOS_ADD_EXECUTABLE(
+    TestCompileOnlyWindowsInclude
+    SOURCES
+    TestWindowsInclude.cpp
+    TestCompileMain.cpp
+    )
+endif (WIN32)
```
dalg24, line 100: Just conditionally append it to `COMPILE_ONLY_SOURCES` above
(and please ~~`endif (WIN32)`~~ -> `endif()`

JBludau: I think this will be very brittle. Reasoning is the following: the other tests include headers that are not `Kokkos_Core.hpp` and thus do not have the guard against the macro problem. We can either guard every single one of them with the same macro push or we stick to the "we fix it for including `Kokkos_Core.hpp`" ... in which case we should add the test as a new executable to not break the other tests.

dalg24: You misunderstood, I meant do
```CMake
if (WIN32)
  list(APPEND COMPILE_ONLY_SOURCES TestWindowsInclude.cpp)
endif()
```
There is no point on adding a new **executable** (adding a new source file was the right thing of course)

JBludau: done in [1a0977](https://github.com/kokkos/kokkos/pull/5541/commits/1a0977bb7daf273b2298b75c009dc49cbded7a3d)

dalg24 (approved):
Do you want to rewrite history or just squash merge?

JBludau:
I think you can squash it ... it seems simple enough to be readable as one commit

PR #5544, opened 2022-10-11
Check compatibility of execution space and memory space in View creation

masterleinad:
Allowing to pass execution space instances to view constructors it is of course possible for the execution space to be incompatible with the memory space. This pull request adds `static_assert`s to that effect.

This was quite helpful in https://github.com/arborx/ArborX/pull/749 and #5488.


PhilMiller:
The spur for my comment was seeing the changes to the tests - if such calls previously worked, should we be prohibiting them?

core/unit_test/tools/TestLogicalSpaces.hpp
```diff
@@ -189,14 +189,5 @@ TEST(defaultdevicetype, chained_logical_spaces) { test_chained_spaces(); }
...
   test_allowed_access<fake_memory_space>();
 }
-// FIXME_SYCL
-#if !(defined(KOKKOS_COMPILER_INTEL) && defined(KOKKOS_ENABLE_SYCL))
-TEST(defaultdevicetype_DeathTest, access_forbidden) {
-  ::testing::FLAGS_gtest_death_test_style = "threadsafe";
-  ASSERT_DEATH(
-      { test_allowed_access<semantically_independent_logical_space>(); },
-      "Kokkos::View ERROR: attempt to access inaccessible memory space");
-}
```
dalg24, line 199: Why is it ok to remove? 
Then the `semantically_independent_logical_space` alias is never used.

dalg24: Also how is int related to this PR?

core/unit_test/TestCreateMirror.cpp
```diff
@@ -89,16 +89,16 @@ void test_create_mirror_properties(const View& view) {
   check_memory_space(create_mirror_view(view_alloc(                     DeviceMemorySpace{}), view), DeviceMemorySpace{});
 
   // create_mirror view_alloc + execution space
-  check_memory_space(create_mirror(view_alloc(DefaultExecutionSpace{}, WithoutInitializing),                        view), host_mirror_test_space(view));
-  check_memory_space(create_mirror(view_alloc(DefaultHostExecutionSpace{}),                                         view), host_mirror_test_space(view));
-  check_memory_space(create_mirror(view_alloc(DefaultExecutionSpace{},   WithoutInitializing, DeviceMemorySpace{}), view), DeviceMemorySpace{});
-  check_memory_space(create_mirror(view_alloc(DefaultExecutionSpace{},                        DeviceMemorySpace{}), view), DeviceMemorySpace{});
+  check_memory_space(create_mirror(view_alloc(DefaultHostExecutionSpace{}, WithoutInitializing),                      view), host_mirror_test_space(view));
+  check_memory_space(create_mirror(view_alloc(DefaultHostExecutionSpace{}),                                           view), host_mirror_test_space(view));
```
dalg24, line 93: Why is it not just `Kokkos::HostSpace`?

PhilMiller: OpenMP vs Serial would be distinct execution spaces. I'm tempted to say that Serial should be tested explicitly, along with this

dalg24: I am referring to the trailing argument used in the static assertion

dalg24: The `host_mirror_test_space(view)`

masterleinad: If the memory space of the original view is already accessible from the host (HostSpace, CudaUVM, ...), a View in the same memory space is created.

core/unit_test/tools/TestLogicalSpaces.hpp
masterleinad: It fails the `static_assert` and thus doesn't even compile. We could replace with a compile check for accessibility instead.

masterleinad:
> The spur for my comment was seeing the changes to the tests - if such calls previously worked, should we be prohibiting them?

They only worked when we didn't try to initialize the View. For allocating passing an incompatible execution space would just fall back to using the overload that doesn't take an execution space. Thus, I could see a point in moving the static_assert past the check if we actually initialize the View.

core/unit_test/TestCreateMirror.cpp
dalg24: for `create_mirror`?  (not `create_mirror_view`)

masterleinad: That's what the test already checks, right? Also, see https://github.com/kokkos/kokkos/blob/cc6504d21a0ebc3f17c30aeeaa44c3690fb40d47/core/src/Kokkos_HostSpace.hpp#L193-L219,  https://github.com/kokkos/kokkos/blob/cc6504d21a0ebc3f17c30aeeaa44c3690fb40d47/core/src/Kokkos_View.hpp#L287-L290, https://github.com/kokkos/kokkos/blob/cc6504d21a0ebc3f17c30aeeaa44c3690fb40d47/core/src/Kokkos_View.hpp#L351, and https://github.com/kokkos/kokkos/blob/cc6504d21a0ebc3f17c30aeeaa44c3690fb40d47/core/src/Kokkos_View.hpp#L632-L636.

core/unit_test/tools/TestLogicalSpaces.hpp
dalg24: This test is about memory access violation.  Why are we not able to build the view anymore?
It is out of the question to merge that change without going to the bottom of this.

containers/unit_tests/TestWithoutInitializing.hpp
```diff
@@ -113,43 +113,51 @@ TEST(TEST_CATEGORY, resize_realloc_no_alloc_dualview) {
...
-          return MatchDiagnostic{true, {"Found end event"}};
-        return MatchDiagnostic{false};
-      });
-  ASSERT_TRUE(success);
-  listen_tool_events(Config::DisableAll());
+  constexpr bool exec_space_can_access_host_space =
+      Kokkos::SpaceAccessibility<TEST_EXECSPACE, Kokkos::HostSpace>::accessible;
+  if constexpr (!exec_space_can_access_host_space) {
+    GTEST_SKIP() << "skipping since execution space can't access HostSpace";
+  } else {
```
dalg24, line 120: Please elaborate on this change and the one below where you commented.

core/unit_test/tools/TestLogicalSpaces.hpp
masterleinad: This test tries to create a View in a memory space that is not accessible from its execution space and then tries to access it in a `parallel_for` using that memory space. We expect an access error that is currently happening in the `parallel_for` although we initialize the view. It doesn't already fail in the View initialization because we are using `memset` which doesn't check for memory access (I have been running into (non-Kokkos) Cuda runtime errors because of that). With the original version of this patch, we don't even allow the View initialization to happen which I think is the right thing to do here. To still keep the test, I am now using an unmanaged view.

In the end, the open question still is "Should we use the static_assert only if the View is actually initialized?".
If yes, it would be sufficient to use `Kokkos::WithoutInitializing` instead of an unmanaged View.

containers/unit_tests/TestWithoutInitializing.hpp
masterleinad: In case we pass an execution space, it is used for View construction of both views. Thus, this can only work if the execution space can access `HostSpace`.

containers/unit_tests/TestWithoutInitializing.hpp
```diff
@@ -113,43 +113,60 @@ TEST(TEST_CATEGORY, resize_realloc_no_alloc_dualview) {
...
+                       Config::EnableKernels());
+    Kokkos::DualView<int*** * [1][2][3][4], TEST_EXECSPACE> bla("bla", 8, 7, 6,
+                                                                5);
+    auto success = validate_absence(
+        [&]() {
+          constexpr bool host_exec_space_can_access_memory_space =
+              Kokkos::SpaceAccessibility<
+                  Kokkos::DefaultHostExecutionSpace,
+                  TEST_EXECSPACE::memory_space>::accessible;
+          if constexpr (host_exec_space_can_access_memory_space) {
```
dalg24, line 137: I that not guarded by the `if constexpr` L122?

masterleinad: No, I couldn't capture the `constexpr` variable as such and needed to repeat it.

dalg24: The compiler is too smart and sees the `static_assert(<EVAL-TO-FALSE>)` in the else-branch of the `if constexpr`.
If we template the unit test on `ExecutionSpace` it will work

core/src/impl/Kokkos_ViewMapping.hpp
```diff
@@ -3443,6 +3443,9 @@ class ViewMapping<
     const memory_space& mem_space =
         Impl::get_property<Impl::MemorySpaceTag>(arg_prop);
 
+    static_assert(
+        SpaceAccessibility<execution_space, memory_space>::accessible);
```
dalg24, line 3447: By the way why did you place this static assertion here rather than right after `execution_space` and `memory_space` definitions?

core/unit_test/tools/TestLogicalSpaces.hpp
```diff
@@ -131,7 +131,7 @@ void test_malloc_free() {
   auto* temp =
       Kokkos::kokkos_malloc<fake_memory_space>("does_malloc_work", 1000);
   expect_deallocation_event("does_malloc_work", "TestSpace", "Error in free");
-  Kokkos::kokkos_free(temp);
+  Kokkos::kokkos_free<fake_memory_space>(temp);
```
dalg24, line 134: Why do you have this change.  This was fixed in #5567

containers/unit_tests/TestWithoutInitializing.hpp
dalg24: This gives
```
/var/jenkins/workspace/Kokkos/containers/unit_tests/TestWithoutInitializing.hpp: In lambda function:
/var/jenkins/workspace/Kokkos/containers/unit_tests/TestWithoutInitializing.hpp:133:16: error: declaration of 'host_exec_space_can_access_memory_space' shadows a previous local [-Werror=shadow]
  133 |           constexpr bool host_exec_space_can_access_memory_space =
      |                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/var/jenkins/workspace/Kokkos/containers/unit_tests/TestWithoutInitializing.hpp:119:16: note: shadowed declaration is here
  119 |   constexpr bool host_exec_space_can_access_memory_space =
      |                ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
cc1plus: all warnings being treated as errors
```
Just template the unit test like I suggested, it works.

core/src/impl/Kokkos_ViewMapping.hpp
masterleinad: I didn't want to interrupt the block of aliases and it felt like a good place since we are declaring related variables close to it.

core/unit_test/tools/TestLogicalSpaces.hpp
masterleinad: I am adding a commit merging with `upstream/develop` now.

core/src/impl/Kokkos_ViewMapping.hpp
dalg24: Move it right after their definition (L3421)

containers/unit_tests/TestWithoutInitializing.hpp
```diff
@@ -112,44 +112,62 @@ TEST(TEST_CATEGORY, resize_realloc_no_alloc_dualview) {
...
-      [&](EndParallelForEvent event) {
-        if (event.descriptor().find("initialization") != std::string::npos)
-          return MatchDiagnostic{true, {"Found end event"}};
-        return MatchDiagnostic{false};
-      });
-  ASSERT_TRUE(success);
-  listen_tool_events(Config::DisableAll());
+TEST(TEST_CATEGORY, resize_exec_space_dualview) {
+  // The if constexpr doesn't properly guard if we don't template the test so we
+  // introduce a separate templated function for it.
```
dalg24, line 169: Did you see failures other than NVCC?

masterleinad: It's also failing in a SYCL build.

dalg24 (approved):

crtrott (changes requested):
We need to fix DualView resize for this. The constructor already works (i.e. you can only hand it an execution space for the device view, the host view is created with `create_mirror_view` without passing on `arg_props`. 
But resize has a runtime check on the modification flags to decide which side to actually run the resize on. I think we need to have an additional if constexpr (or enable_if based) thing which supercedes the modified flags for the decision on where to run the resize, and which side then simply gets updated via `create_mirror_view`.

We do not want to change the test for DualView. 
Specifically we can't just forward the execution space to both things and need to make sure that we do the

masterleinad:
@crtrott See https://github.com/kokkos/kokkos/pull/5544/commits/15088ff8da5a49cb66736d3a04945e48b38c6e59.

crtrott (approved):

containers/src/Kokkos_DualView.hpp
```diff
@@ -1050,6 +1053,37 @@ class DualView : public ViewTraits<DataType, Arg1Type, Arg2Type, Arg3Type> {
...
         ++modified_flags(0);
       }
+    };
+
+    constexpr bool has_execution_space = alloc_prop_input::has_execution_space;
+
+    if constexpr (has_execution_space) {
+      using ExecSpace = typename alloc_prop_input::execution_space;
+      const auto& exec_space =
+          Impl::get_property<Impl::ExecutionSpaceTag>(arg_prop);
```
ajpowelsnl, line 1063: Should `Impl:: ...` be user-facing (assuming that `Kokkos_DualView.hpp` is a public header)?  Also, should it be `(arg_prop)` or `(properties)`, as in line 1044 above?

containers/src/Kokkos_DualView.hpp
```diff
@@ -1050,6 +1053,37 @@ class DualView : public ViewTraits<DataType, Arg1Type, Arg2Type, Arg3Type> {
...
+      constexpr bool exec_space_can_access_device =
+          SpaceAccessibility<ExecSpace,
+                             typename t_dev::memory_space>::accessible;
+      constexpr bool exec_space_can_access_host =
+          SpaceAccessibility<ExecSpace,
+                             typename t_host::memory_space>::accessible;
+      static_assert(exec_space_can_access_device || exec_space_can_access_host);
+      if constexpr (exec_space_can_access_device) {
+        sync<typename t_dev::memory_space>(exec_space);
+        resize_on_device(arg_prop);
```
ajpowelsnl, line 1073: Same question as above about `(arg_prop)` vs. `(properties)`  for lines 1073, 1078, 1083 and 1085?

containers/unit_tests/TestCreateMirror.cpp
```diff
@@ -106,20 +106,20 @@ void test_create_mirror_properties(const View& view) {
...
-    check_memory_space(create_mirror_view(view_alloc(DefaultHostExecutionSpace{}),                                       view), host_mirror_test_space(view));
+    check_memory_space(create_mirror_view(view_alloc(DefaultHostExecutionSpace{}, WithoutInitializing),                      view), host_mirror_test_space(view));
+    check_memory_space(create_mirror_view(view_alloc(DefaultHostExecutionSpace{}),                                           view), host_mirror_test_space(view));
   }
-  check_memory_space(create_mirror_view(view_alloc(DefaultExecutionSpace{},   WithoutInitializing, DeviceMemorySpace{}), view), DeviceMemorySpace{});
-  check_memory_space(create_mirror_view(view_alloc(DefaultExecutionSpace{},                        DeviceMemorySpace{}), view), DeviceMemorySpace{});
+  check_memory_space(create_mirror_view(view_alloc(DefaultExecutionSpace{},       WithoutInitializing, DeviceMemorySpace{}), view), DeviceMemorySpace{});
+  check_memory_space(create_mirror_view(view_alloc(DefaultExecutionSpace{},                            DeviceMemorySpace{}), view), DeviceMemorySpace{});
```
ajpowelsnl, line 123: Do lines 117 - 122 duplicate the code in 108 - 113?  If so, why?

containers/unit_tests/TestWithoutInitializing.hpp
```diff
@@ -476,7 +476,7 @@ TEST(TEST_CATEGORY, create_mirror_no_init_dynrankview_viewctor) {
         ASSERT_EQ(device_view.size(), mirror_device.size());
         auto mirror_host = Kokkos::create_mirror(
             Kokkos::view_alloc(Kokkos::WithoutInitializing,
-                               Kokkos::DefaultExecutionSpace{}),
+                               Kokkos::DefaultHostExecutionSpace{}),
```
ajpowelsnl, line 479: Is it advantageous to specify `DefaultHostExecutionSpace{}` vs. the previous version, `DefaultExecutionSpace{}`?  In other words, isn't the `ExecutionSpace` something that should remain generic, vs. a sort of predetermining  `ExecutionSpace` ?  The same question for lines 550 - 552, 558, 586, 594, etc.

ajpowelsnl (commented):
Looks good, but I did have a few questions

containers/src/Kokkos_DualView.hpp
```diff
@@ -1021,10 +1021,11 @@ class DualView : public ViewTraits<DataType, Arg1Type, Arg2Type, Arg3Type> {
     if (modified_flags.data() == nullptr) {
       modified_flags = t_modified_flags("DualView::modified_flags");
     }
-    if (modified_flags(1) >= modified_flags(0)) {
+
+    [[maybe_unused]] auto resize_on_device = [&](const auto& properties) {
```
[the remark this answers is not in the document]
masterleinad: Yes, https://en.cppreference.com/w/cpp/language/attributes/maybe_unused.

containers/src/Kokkos_DualView.hpp
masterleinad: No, `properties` is an argument to the generic lambda. This is not inside the lambda.

containers/unit_tests/TestCreateMirror.cpp
masterleinad: Lines 108-113 test `create_mirror`, lines 117-122 test `create_mirror_view`.

containers/unit_tests/TestWithoutInitializing.hpp
masterleinad: We create a mirror in a host-accessible space here. Hence, the provided execution space must also be a host execution space.

Issue #5453, opened 2022-09-12
Kokkos allow HIP compilers that aren't called hipcc


jjellio:
This is a request!
Kokkos (through Trilinos - or always?) - wants to enable a TPL of ROCM if the compiler is not `hipcc`.
https://github.com/kokkos/kokkos/blob/develop/cmake/kokkos_tpls.cmake#L42

This is kinda annoying, and I religiously patch my Trilinos pulls to `SET(ROCM_DEFAULT OFF).`

The builds all work fine with amdclang or crayclang directly ... as these compilers imply ROCM (do any modern HIP compilers not?).  I just manually merged the latest kokkos promotion (since Trilinos' PR tester is .... having issues) .. and I keep hoping with all the Cmake updates, that this nuisance  will go away.... but it doesn't.

Will this ultimately go away at some point? 

**Please also include the following items to support reproducing the bug**
1. anything but `hipcc`
2. Kokkos promotion in PR testing at Trilinos (3.7)
3. any AMD machine using HIP
4. cmake configure command
```
 "-DKokkos_ARCH_ZEN3:BOOL=ON" \
 "-DKokkos_ENABLE_SERIAL:BOOL=ON" \
 "-DKokkos_ENABLE_OPENMP:BOOL=OFF" \
 "-DKokkos_ENABLE_HIP:BOOL=ON" \
 "-DKokkos_ARCH_VEGA90A:BOOL=ON" \
```
6. output from cmake command
7. code needed to reproduce the bug
8. command line needed to reproduce the bug
7. please also attach the `KokkosCore_config.h` header file (generated during the build);
**Any additional info**
Please provide any additional context about the issue here.

Rombur:
I've just tried to set `SET(ROCM_DEFAULT OFF)` with `amdclang` and Kokkos does not compile.
```
[  5%] Linking CXX executable KokkosCore_UnitTest_PushFinalizeHook                                                                                                                                                 
ld.lld: error: undefined symbol: __hipRegisterVar                                                                                                                                                                  
>>> referenced by UnitTest_PushFinalizeHook.cpp                                                                                                                                                                    
>>>               CMakeFiles/KokkosCore_UnitTest_PushFinalizeHook.dir/UnitTest_PushFinalizeHook.cpp.o:(__hip_register_globals)                 
>>> referenced by UnitTest_PushFinalizeHook.cpp                                                                                                                                                                    
[...]

PR #5546, opened 2022-10-12
Do not error out when using amdclang with Trilinos

Rombur:
Currently it is impossible to compile Trilinos using `amdclang` without patching Kokkos. The problem is that we detect that we are using `amdclang` and we enable the ROCM TPL. Currently our build system does not support TPL when using Trilinos and we just error. With this patch the configuration does not error but it is up to the user to set the correct flags. For a proper fix we need to rework the build system.

Related to https://github.com/kokkos/kokkos/issues/5453


crtrott (approved):

masterleinad (approved):

dalg24:
Failures are unrelated

PR #5550, opened 2022-10-13
Atomic support for OpenACC back-end

marcgt1972:
We are aware that the current implementation is not fully in accordance to the current conventions for integration within the atomic-intrinsics support in Kokkos. We will work with Damien Lebrun to solve this issue.


crtrott:
We need to fix this inside of desul, the other code path is supposed to go away completely from Kokkos Core.

PR #5555, opened 2022-10-13
fix incorrect offset in cuda parallel scan for < 4 byte types

### LINKED ISSUE #5545: Cuda Parallel Scan has incorrect offset for types smaller than `size_type`
**Describe the bug**
https://github.com/kokkos/kokkos/blob/0fb8b8ab8bd01cfc7dd27da37f20deb0904afe30/core/src/Cuda/Kokkos_Cuda_Parallel_Range.hpp#L542-L544

This value is 0 if the size of `value_type` is < 4 bytes (`sizeof(size_type)`)

This is the underlying cause for https://github.com/kokkos/kokkos/issues/5513 and partially fixes the issue

nmm0:
Fixes #5545

### DIFF
--- a/core/src/Cuda/Kokkos_Cuda_Parallel_Range.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_Parallel_Range.hpp
@@ -77,7 +77,7 @@ class ParallelFor<FunctorType, Kokkos::RangePolicy<Traits...>, Kokkos::Cuda> {
   const FunctorType m_functor;
   const Policy m_policy;
 
-  ParallelFor()        = delete;
+  ParallelFor()                              = delete;
   ParallelFor& operator=(const ParallelFor&) = delete;
 
   template <class TagType>
@@ -463,8 +463,24 @@ class ParallelScan<FunctorType, Kokkos::RangePolicy<Traits...>, Kokkos::Cuda> {
  public:
   using pointer_type   = typename Analysis::pointer_type;
   using reference_type = typename Analysis::reference_type;
+  using value_type     = typename Analysis::value_type;
   using functor_type   = FunctorType;
-  using size_type      = Cuda::size_type;
+  // Conditionally set word_size_type to int16_t or int8_t if value_type is
+  // smaller than int32_t (Kokkos::Cuda::size_type)
+  // word_size_type is used to determine the word count, shared memory buffer
+  // size, and global memory buffer size before the scan is performed.
+  // Within the reduction, the word count is recomputed based on word_size_type
+  // and when calculating indexes into the shared/global memory buffers for
+  // performing the reduction, word_size_type is used again.
+  // For scalars > 4 bytes in size, indexing into shared/global memory relies
+  // on the block and grid dimensions to ensure that we index at the correct
+  // offset rather than at every 4 byte word; such that, when the join is
+  // performed, we have the correct data that was copied over in chunks of 4
+  // bytes.
+  using word_size_type = std::conditional_t<
+      sizeof(value_type) < sizeof(Kokkos::Cuda::size_type),
+      std::conditional_t<sizeof(value_type) == 2, int16_t, int8_t>,
+      Kokkos::Cuda::size_type>;
 
  private:
   // Algorithmic constraints:
@@ -475,9 +491,9 @@ class ParallelScan<FunctorType, Kokkos::RangePolicy<Traits...>, Kokkos::Cuda> {
 
   const FunctorType m_functor;
   const Policy m_policy;
-  size_type* m_scratch_space;
-  size_type* m_scratch_flags;
-  size_type m_final;
+  word_size_type* m_scratch_space;
+  Cuda::size_type* m_scratch_flags;
+  Cuda::size_type m_final;
 #ifdef KOKKOS_IMPL_DEBUG_CUDA_SERIAL_EXECUTION
   bool m_run_serial;
 #endif
@@ -499,12 +515,12 @@ class ParallelScan<FunctorType, Kokkos::RangePolicy<Traits...>, Kokkos::Cuda> {
   __device__ inline void initial() const {
     typename Analysis::Reducer final_reducer(&m_functor);
 
-    const integral_nonzero_constant<size_type, Analysis::StaticValueSize /
-                                                   sizeof(size_type)>
-        word_count(Analysis::value_size(m_functor) / sizeof(size_type));
+    const integral_nonzero_constant<word_size_type, Analysis::StaticValueSize /
+                                                        sizeof(word_size_type)>
+        word_count(Analysis::value_size(m_functor) / sizeof(word_size_type));
 
-    size_type* const shared_value =
-        kokkos_impl_cuda_shared_memory<size_type>() +
+    word_size_type* const shared_value =
+        kokkos_impl_cuda_shared_memory<word_size_type>() +
         word_count.value * threadIdx.y;
 
     final_reducer.init(reinterpret_cast<pointer_type>(shared_value));
@@ -530,7 +546,7 @@ class ParallelScan<FunctorType, Kokkos::RangePolicy<Traits...>, Kokkos::Cuda> {
     // gridDim.x
     cuda_single_inter_block_reduce_scan<true>(
         final_reducer, blockIdx.x, gridDim.x,
-        kokkos_impl_cuda_shared_memory<size_type>(), m_scratch_space,
+        kokkos_impl_cuda_shared_memory<word_size_type>(), m_scratch_space,
         m_scratch_flags);
   }
 
@@ -539,21 +555,21 @@ class ParallelScan<FunctorType, Kokkos::RangePolicy<Traits...>, Kokkos::Cuda> {
   __device__ inline void final() const {
     typename Analysis::Reducer final_reducer(&m_functor);
 
-    const integral_nonzero_constant<size_type, Analysis::StaticValueSize /
-                                                   sizeof(size_type)>
-        word_count(Analysis::value_size(m_functor) / sizeof(size_type));
+    const integral_nonzero_constant<word_size_type, Analysis::StaticValueSize /
+                                                        sizeof(word_size_type)>
+        word_count(Analysis::value_size(m_functor) / sizeof(word_size_type));
 
     // Use shared memory as an exclusive scan: { 0 , value[0] , value[1] ,
     // value[2] , ... }
-    size_type* const shared_data = kokkos_impl_cuda_shared_memory<size_type>();
-    size_type* const shared_prefix =
+    word_size_type* const shared_data = kokkos_impl_cuda_shared_memory<word_size_type>();
+    word_size_type* const shared_prefix =
         shared_data + word_count.value * threadIdx.y;
-    size_type* const shared_accum =
+    word_size_type* const shared_accum =
         shared_data + word_count.value * (blockDim.y + 1);
 
     // Starting value for this thread block is the previous block's total.
     if (blockIdx.x) {
-      size_type* const block_total =
+      word_size_type* const block_total =
           m_scratch_space + word_count.value * (blockIdx.x - 1);
       for (unsigned i = threadIdx.y; i < word_count.value; ++i) {
         shared_accum[i] = block_total[i];
@@ -600,7 +616,7 @@ class ParallelScan<FunctorType, Kokkos::RangePolicy<Traits...>, Kokkos::Cuda> {
           typename Analysis::pointer_type(shared_data + word_count.value));
 
       {
-        size_type* const block_total =
+        word_size_type* const block_total =
             shared_data + word_count.value * blockDim.y;
         for (unsigned i = threadIdx.y; i < word_count.value; ++i) {
           shared_accum[i] = block_total[i];
@@ -688,10 +704,10 @@ class ParallelScan<FunctorType, Kokkos::RangePolicy<Traits...>, Kokkos::Cuda> {
       // How many block are really needed for this much work:
       const int grid_x = (nwork + work_per_block - 1) / work_per_block;
 
-      m_scratch_space = cuda_internal_scratch_space(
-          m_policy.space(), Analysis::value_size(m_functor) * grid_x);
+      m_scratch_space = reinterpret_cast< word_size_type * >( cuda_internal_scratch_space(
+          m_policy.space(), Analysis::value_size(m_functor) * grid_x) );
       m_scratch_flags =
-          cuda_internal_scratch_flags(m_policy.space(), sizeof(size_type) * 1);
+          cuda_internal_scratch_flags(m_policy.space(), sizeof(Cuda::size_type) * 1);
 
       dim3 grid(grid_x, 1, 1);
       dim3 block(1, block_size, 1);  // REQUIRED DIMENSIONS ( 1 , N , 1 )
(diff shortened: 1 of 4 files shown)


nmm0:
@cwsmith this should solve the issue though there is additional work to be done in ensuring the correct type is used for the internal parallel scan when using `Kokkos::Experimental::inclusive_scan"

core/src/Cuda/Kokkos_Cuda_Parallel_Range.hpp
```diff
@@ -1026,7 +1061,8 @@ class ParallelScanWithTotal<FunctorType, Kokkos::RangePolicy<Traits...>,
         if (!m_result_ptr_device_accessible)
           DeepCopy<HostSpace, CudaSpace, Cuda>(
               m_policy.space(), m_result_ptr,
-              m_scratch_space + (grid_x - 1) * size / sizeof(int), size);
```
nmm0, line 1029: I don't actually know where `int` came from in the original, but `sizeof(word_size_type)` is important here otherwise the result will be incorrect

PhilMiller (commented):
Does any corresponding change need to be done for HIP?

nmm0:
> Does any corresponding change need to be done for HIP?

From looking at the code it certainly appears so -- in fact the logic bug (size < `HIP::size_type`) appears to also be present in HIP's reduce

core/src/Cuda/Kokkos_Cuda_Parallel_Range.hpp
```diff
@@ -463,8 +463,24 @@ class ParallelScan<FunctorType, Kokkos::RangePolicy<Traits...>, Kokkos::Cuda> {
...
+  // word_size_type is used to determine the word count, shared memory buffer
+  // size, and global memory buffer size before the scan is performed.
+  // Within the reduction, the word count is recomputed based on word_size_type
+  // and when calculating indexes into the shared/global memory buffers for
+  // performing the reduction, word_size_type is used again.
+  // For scalars > 4 bytes in size, indexing into shared/global memory relies
+  // on the block and grid dimensions to ensure that we index at the correct
+  // offset rather than at every 4 byte word; such that, when the join is
+  // performed, we have the correct data that was copied over in chunks of 4
+  // bytes.
```
masterleinad, line 479:
```suggestion
  // Conditionally set word_size_type to int16_t or int8_t if value_type is
  // smaller than int32_t (Kokkos::Cuda::size_type)
  // word_size_type is used to determine the word count, shared memory buffer
  // size, and global memory buffer size before the scan is performed.
  // Within the scan, the word count is recomputed based on word_size_type
  // and when calculating indexes into the shared/global memory buffers for
  // performing the scan, word_size_type is used again.
  // For scalars > 4 bytes in size, indexing into shared/global memory relies
  // on the block and grid dimensions to ensure that we index at the correct
  // offset rather than at every 4 byte word; such that, when the join is
  // performed, we have the correct data that was copied over in chunks of 4
  // bytes.
```

core/src/Cuda/Kokkos_Cuda_Parallel_Range.hpp
```diff
@@ -754,7 +772,22 @@ class ParallelScanWithTotal<FunctorType, Kokkos::RangePolicy<Traits...>,
...
+  // word_size_type is used to determine the word count, shared memory buffer
+  // size, and global memory buffer size before the reduction is performed.
+  // Within the reduction, the word count is recomputed based on word_size_type
+  // and when calculating indexes into the shared/global memory buffers for
+  // performing the reduction, word_size_type is used again.
+  // For scalars > 4 bytes in size, indexing into shared/global memory relies
+  // on the block and grid dimensions to ensure that we index at the correct
+  // offset rather than at every 4 byte word; such that, when the join is
+  // performed, we have the correct data that was copied over in chunks of 4
+  // bytes.
```
masterleinad, line 786:
```suggestion
  // Conditionally set word_size_type to int16_t or int8_t if value_type is
  // smaller than int32_t (Kokkos::Cuda::size_type)
  // word_size_type is used to determine the word count, shared memory buffer
  // size, and global memory buffer size before the scan is performed.
  // Within the scan, the word count is recomputed based on word_size_type
  // and when calculating indexes into the shared/global memory buffers for
  // performing the scan, word_size_type is used again.
  // For scalars > 4 bytes in size, indexing into shared/global memory relies
  // on the block and grid dimensions to ensure that we index at the correct
  // offset rather than at every 4 byte word; such that, when the join is
  // performed, we have the correct data that was copied over in chunks of 4
  // bytes.
```

masterleinad:
I think we should make the same changes for `HIP`.

nmm0:
> I think we should make the same changes for `HIP`.

Filed https://github.com/kokkos/kokkos/issues/5556

crtrott:
Yeah lets get the HIP thing fixed here too, since the test doesn't pass anyway.

core/unit_test/TestScan.hpp
```diff
@@ -170,11 +170,108 @@ struct TestScan {
...
+      if (update != static_cast<value_type>(i)) {
+        int fail = errors()++;
+
+        // Limit the amount of output
+        if (fail < 20) {
+          KOKKOS_IMPL_DO_NOT_USE_PRINTF(
+              "TestSmallSizeTypeScan(%d) = %ld != %ld\n", i,
+              static_cast<long>(update), static_cast<long>(i));
+        }
+      }
```
dalg24, line 214: Do not mix computation and verification.  Launch another kernel for that.

core/unit_test/TestScan.hpp
```diff
@@ -170,11 +170,108 @@ struct TestScan {
...
+  TestSmallSizeTypeScan<TEST_EXECSPACE, std::int8_t>(5, false);
+  TestSmallSizeTypeScan<TEST_EXECSPACE, std::int8_t>(100, false);
+  TestSmallSizeTypeScan<TEST_EXECSPACE, std::int8_t>(
+      std::numeric_limits<std::int8_t>::max(), false);
+  TestSmallSizeTypeScan<TEST_EXECSPACE, std::int16_t>(0, false);
+  TestSmallSizeTypeScan<TEST_EXECSPACE, std::int16_t>(5, false);
+  TestSmallSizeTypeScan<TEST_EXECSPACE, std::int16_t>(10000, false);
+  TestSmallSizeTypeScan<TEST_EXECSPACE, std::int16_t>(
+      std::numeric_limits<std::int16_t>::max(), false);
+  TEST_EXECSPACE().fence();
```
dalg24, line 262: Why fence?

dalg24 (changes requested):
Please beef up the description of the PR and relate it to #4156

core/unit_test/TestScan.hpp
nmm0: That's what the existing test does, I tried to copy it as closely as possible. I can remove both fences though

core/unit_test/TestScan.hpp
nmm0: Similarly, should I remove from the existing test?

Rombur:
I have a patch that fixes the problem with HIP. How should we proceed?

dalg24:
Please go ahead and push to this PR

core/src/HIP/Kokkos_HIP_ReduceScan.hpp
```diff
@@ -390,13 +391,12 @@ __device__ void hip_intra_block_reduce_scan(
  *  Global reduce result is in the last threads' 'shared_data' location.
  */
 
-template <bool DoScan, class FunctorType>
+template <bool DoScan, class FunctorType, typename SizeType>
```
dalg24, line 394: Do not mix `class` and `typename`.  Pick your poison.

core/unit_test/TestScan.hpp
```diff
@@ -170,11 +170,108 @@ struct TestScan {
...
+    check_error(res, N, use_total);
+  }
+
+  KOKKOS_INLINE_FUNCTION
+  void operator()(int i, value_type& update, bool final) const {
+    const auto val = x(i);
+    if (final) {
+      x(i) = update;
+
+      // We should have a nice count from 0 to 1...
```
masterleinad, line 204:
```suggestion
      // We should have a nice count from 0 to N
```
?

core/unit_test/TestScan.hpp
```diff
@@ -170,11 +170,108 @@ struct TestScan {
...
+              static_cast<long>(update), static_cast<long>(i));
+        }
+      }
+    }
+    update += val;
+  }
+
+  KOKKOS_INLINE_FUNCTION
+  void init(value_type& update) const { update = 0; }
+
```
masterleinad, line 221:
```suggestion
```

core/unit_test/TestScan.hpp
```diff
@@ -170,11 +170,108 @@ struct TestScan {
...
   TestScan<TEST_EXECSPACE>::test_range(1, 1000);
   TestScan<TEST_EXECSPACE>(0);
   TestScan<TEST_EXECSPACE>(100000);
   TestScan<TEST_EXECSPACE>(10000000);
   TEST_EXECSPACE().fence();
 }
-}  // namespace Test
+
+TEST(TEST_CATEGORY, small_size_scan) {
+  TestSmallSizeTypeScan<TEST_EXECSPACE, std::int8_t>(0, false);
```
masterleinad, line 252: It's not intuitive to me to understand what the second parameter could do just by seeing these lines. Would you mind introducing another named variable to make this clear?

core/src/Cuda/Kokkos_Cuda_Parallel_Range.hpp
```diff
@@ -463,8 +463,24 @@ class ParallelScan<FunctorType, Kokkos::RangePolicy<Traits...>, Kokkos::Cuda> {
  public:
   using pointer_type   = typename Analysis::pointer_type;
   using reference_type = typename Analysis::reference_type;
+  using value_type     = typename Analysis::value_type;
   using functor_type   = FunctorType;
-  using size_type      = Cuda::size_type;
```
dalg24, line 467: Nitpick: I would have kept that guy and not update the definitions of `m_scratch_flags` and `m_final` below.
Same comment in `ParallelScanWithTotal`.

core/unit_test/TestScan.hpp
```diff
@@ -169,12 +170,31 @@ struct TestScan {
...
-  TestScan<TEST_EXECSPACE>::test_range(1, 1000);
-  TestScan<TEST_EXECSPACE>(0);
-  TestScan<TEST_EXECSPACE>(100000);
-  TestScan<TEST_EXECSPACE>(10000000);
+  constexpr auto imbalance_size = 1000;
+  TestScan<TEST_EXECSPACE, int64_t, imbalance_size>::test_range(1, 1000);
+  TestScan<TEST_EXECSPACE, int64_t, imbalance_size>(0);
+  TestScan<TEST_EXECSPACE, int64_t, imbalance_size>(100000);
+  TestScan<TEST_EXECSPACE, int64_t, imbalance_size>(10000000);
+  TEST_EXECSPACE().fence();
```
dalg24, line 181: Please drop these fences.  They are pointless.

dalg24 (approved):

core/unit_test/TestScan.hpp
```diff
@@ -45,20 +45,21 @@
...
   Kokkos::View<int, Device, Kokkos::MemoryTraits<Kokkos::Atomic> > errors;
 
   KOKKOS_INLINE_FUNCTION
   void operator()(const int iwork, value_type& update,
                   const bool final_pass) const {
-    const value_type n         = iwork + 1;
-    const value_type imbalance = ((1000 <= n) && (0 == n % 1000)) ? 1000 : 0;
+    const value_type n = iwork + 1;
+    const value_type imbalance =
+        ((ImbalanceSz <= n) && (0 == n % ImbalanceSz)) ? ImbalanceSz : 0;
```
dalg24, line 62: HPX builds fails with
```
In file included from /home/runner/work/kokkos/kokkos/kokkos/build/core/unit_test/hpx/TestHPX_Scan.cpp:2:
[88](https://github.com/kokkos/kokkos/actions/runs/3341905420/jobs/5533579527#step:9:89)
/home/runner/work/kokkos/kokkos/kokkos/core/unit_test/TestScan.hpp:62:23: error: comparison of integers of different signs: 'unsigned long' and 'const (anonymous namespace)::TestScan<Kokkos::Experimental::HPX, long, 1000>::value_type' (aka 'const long') [-Werror,-Wsign-compare]
[89](https://github.com/kokkos/kokkos/actions/runs/3341905420/jobs/5533579527#step:9:90)
        ((ImbalanceSz <= n) && (0 == n % ImbalanceSz)) ? ImbalanceSz : 0;
[90](https://github.com/kokkos/kokkos/actions/runs/3341905420/jobs/5533579527#step:9:91)
          ~~~~~~~~~~~ ^  ~

File: core/unit_test/TestScan.hpp (hunk 9 above)
Comment (nmm0) [reply core/unit_test/TestScan.hpp]: removed this value, but did similarly with the new ImbalanceSz param

File: core/unit_test/TestScan.hpp (hunk 7 above)
Comment (nmm0) [reply core/unit_test/TestScan.hpp]: was removed when I combined the testing harasses

File: core/unit_test/TestScan.hpp (hunk 11 above)
Comment (nmm0) [reply core/unit_test/TestScan.hpp]: done

File: core/unit_test/TestScan.hpp [hunk 13]
```diff
@@ -45,20 +45,21 @@
...
   Kokkos::View<int, Device, Kokkos::MemoryTraits<Kokkos::Atomic> > errors;
 
   KOKKOS_INLINE_FUNCTION
   void operator()(const int iwork, value_type& update,
                   const bool final_pass) const {
-    const value_type n         = iwork + 1;
-    const value_type imbalance = ((1000 <= n) && (0 == n % 1000)) ? 1000 : 0;
+    const value_type n = iwork + 1;
+    const value_type imbalance =
+        ((ImbalanceSz <= n) && (value_type(0) == n % ImbalanceSz)) ? ImbalanceSz : value_type(0);
```
dalg24, line 62:
```suggestion
        ((ImbalanceSz <= n) && (value_type(0) == n % ImbalanceSz))
            ? ImbalanceSz
            : value_type(0);
```

dalg24:
The HPX build is now passing.  On the ORNL Jenkins CI server, there is one (likely unrelated) CUDA timing-based failure which I am willing to ignore.
We need one more approval though.

masterleinad (approved):

core/src/Cuda/Kokkos_Cuda_Parallel_Range.hpp
dalg24: I went ahead and applied the suggestion to reduce the diff and be consistent with the HIP backend

masterleinad (approved):
Still looks good to me.

core/src/Cuda/Kokkos_Cuda_Parallel_Range.hpp
```diff
@@ -987,10 +1021,11 @@ class ParallelScanWithTotal<FunctorType, Kokkos::RangePolicy<Traits...>,
...
 
-      m_scratch_space = cuda_internal_scratch_space(
-          m_policy.space(), Analysis::value_size(m_functor) * grid_x);
-      m_scratch_flags =
-          cuda_internal_scratch_flags(m_policy.space(), sizeof(size_type) * 1);
+      m_scratch_space =
+          reinterpret_cast<word_size_type*>(cuda_internal_scratch_space(
+              m_policy.space(), Analysis::value_size(m_functor) * grid_x));
+      m_scratch_flags = cuda_internal_scratch_flags(
+          m_policy.space(), sizeof(Cuda::size_type) * 1);
```
ldh4, line 1028:
```suggestion
      m_scratch_flags =
          cuda_internal_scratch_flags(m_policy.space(), sizeof(size_type) * 1);
```

This change seems unnecessary. Doesn't change anything functionally, but I suggest reverting back to using `sizeof(size_type)` here to keep it consistent with other `[backend]_internal_scratch_flags` calls.

ldh4 (approved):
Looks good to me. Just nitpicking on one thing.

core/src/Cuda/Kokkos_Cuda_Parallel_Range.hpp
dalg24: Agreed.  I missed that one.  Will revert.

PR #5564, opened 2022-10-17
remove RDC flags when using CMake language CUDA

ibaned:
When compiling with CUDA as a CMake language,
the RDC aspect is handled by CMake's
CUDA_SEPARABLE_COMPILATION property on targets and the associated CMAKE_CUDA_SEPARABLE_COMPILATION
variable that sets the default value for that
property.
The arguments set by this part of Kokkos' CMake
actually conflict with the above.


ibaned:
> Should `KOKKOS_ENABLE_CUDA_RELOCATABLE_DEVICE_CODE=ON` then set `CMAKE_CUDA_SEPARABLE_COMPILATION=ON` when compiling with `CUDA` as a `CMake` language?

Yes, that makes sense! Can you help me identify where in the Kokkos CMake files that should be set? I can add it to this PR then if you'd like.

crtrott:
Yeah agreed, we should set that option then, and/or warn if there is a mismatch of that with the Kokkos setting.

ibaned:
Okay, I added some code to ensure that, if we are building using CMake's CUDA language, then setting either of those options on will set both of them on. Let me know if anything else is missing from this.

cmake/kokkos_enable_options.cmake
```diff
@@ -130,6 +130,24 @@ IF (KOKKOS_ENABLE_AGGRESSIVE_VECTORIZATION)
...
+# Force consistency of KOKKOS_ENABLE_CUDA_RELOCATABLE_DEVICE_CODE
+# and CMAKE_CUDA_SEPARABLE_COMPILATION when we are compiling
+# using the CMake CUDA language support.
+# Either one being on will turn the other one on.
+IF (KOKKOS_COMPILE_LANGUAGE STREQUAL CUDA)
+  IF (KOKKOS_ENABLE_CUDA_RELOCATABLE_DEVICE_CODE)
+    IF (CMAKE_CUDA_SEPARABLE_COMPILATION)
+    ELSE()
+      SET(CMAKE_CUDA_SEPARABLE_COMPILATION ON)
+    ENDIF()
```
masterleinad, line 142:
```suggestion
    IF (NOT CMAKE_CUDA_SEPARABLE_COMPILATION)
      SET(CMAKE_CUDA_SEPARABLE_COMPILATION ON)
    ENDIF()
```

cmake/kokkos_enable_options.cmake
```diff
@@ -130,6 +130,24 @@ IF (KOKKOS_ENABLE_AGGRESSIVE_VECTORIZATION)
...
+  ELSE()
+    IF (CMAKE_CUDA_SEPARABLE_COMPILATION)
+      SET(KOKKOS_ENABLE_CUDA_RELOCATABLE_DEVICE_CODE ON)
+    ELSE()
+    ENDIF()
```
masterleinad, line 147:
```suggestion
    IF (CMAKE_CUDA_SEPARABLE_COMPILATION)
      SET(KOKKOS_ENABLE_CUDA_RELOCATABLE_DEVICE_CODE ON)
    ENDIF()
```

cmake/kokkos_arch.cmake
```diff
@@ -450,11 +450,13 @@ IF (KOKKOS_ARCH_POWER9)
 ENDIF()
 
 IF (KOKKOS_ENABLE_CUDA_RELOCATABLE_DEVICE_CODE)
-  COMPILER_SPECIFIC_FLAGS(
-    Clang  -fcuda-rdc
-    NVIDIA --relocatable-device-code=true
-    NVHPC -gpu=rdc
-  )
+  IF (NOT KOKKOS_COMPILE_LANGUAGE STREQUAL CUDA)
```
dalg24, line 453: This does not seem right
Assume RDC is OFF but we are using NVHPC L462 we'd still be adding the `-gpu=nordc` flag

cmake/kokkos_enable_options.cmake
```diff
@@ -130,6 +130,22 @@ IF (KOKKOS_ENABLE_AGGRESSIVE_VECTORIZATION)
...
 
+# Force consistency of KOKKOS_ENABLE_CUDA_RELOCATABLE_DEVICE_CODE
+# and CMAKE_CUDA_SEPARABLE_COMPILATION when we are compiling
+# using the CMake CUDA language support.
+# Either one being on will turn the other one on.
+IF (KOKKOS_COMPILE_LANGUAGE STREQUAL CUDA)
+  IF (KOKKOS_ENABLE_CUDA_RELOCATABLE_DEVICE_CODE)
+    IF (NOT CMAKE_CUDA_SEPARABLE_COMPILATION)
+      SET(CMAKE_CUDA_SEPARABLE_COMPILATION ON)
+    ENDIF()
```
dalg24, line 141: I think we might want to raise a warning (or display a STATUS message) when setting the `CMAKE_CUDA_SEPARABLE_COMPILATION` variable

cmake/kokkos_arch.cmake
masterleinad: I agree. This if condition should guard from setting any flags.

ibaned:
Thank you @masterleinad and @dalg24 for your reviews! I believe I've just addressed both of your suggestions, please let me know if other changes are desired.

masterleinad (approved):

dalg24 (approved):

PR #5537, opened 2022-10-07
Only link against libatomic in gnu-make OpenMPTarget build

crtrott:
This mirrors the CMake logic

PR #5565, opened 2022-10-17
[3.7.01] Only link against libatomic in gnu-make OpenMPTarget build

masterleinad:
Cherry-picking #5537.


dalg24 (approved):

dalg24:
Ignoring failures (Cuda timing based on one build and OpenACC)

PR #5570, opened 2022-10-18
[3.7.01] Add support for detecting MPI local rank with MPICH

dalg24:
Cherry-pick #5525 into RC 3.7.01

PR #5576, opened 2022-10-18
Increase support to detect local MPI ranks

dalg24:
Following up on https://github.com/kokkos/kokkos/pull/5570#issuecomment-1282893923

For now only proposing support for the Process Management Interface (PMI) since I could find multiple sources mentioning it.  I am not sure yet whether to also add `MPT_LRANK`.  Anyone knows more about this one?

### DIFF
--- a/core/src/impl/Kokkos_Core.cpp
+++ b/core/src/impl/Kokkos_Core.cpp
@@ -434,6 +434,7 @@ int Kokkos::Impl::get_gpu(const InitializationSettings& settings) {
            "MV2_COMM_WORLD_LOCAL_RANK",   // MVAPICH2
            "MPI_LOCALRANKID",             // MPICH
            "SLURM_LOCALID",               // SLURM
+           "PMI_LOCAL_RANK"               // PMI
        }) {
     local_rank_str = std::getenv(env_var);
     if (local_rank_str) break;


stanmoore1:
@dalg24 I can't find anything about `MPT_LRANK` either, so unless someone complains I'd say leave it out for now.

masterleinad (approved):

PhilMiller (approved):
Totally aside from this PR, I notice that the code doesn't make any attempt to handle `local_rank_str` not actually parsing as an integer

dalg24:
> Totally aside from this PR, I notice that the code doesn't make any attempt to handle `local_rank_str` not actually parsing as an integer

Yes we only check if any of the environment variables is set and later do
https://github.com/kokkos/kokkos/blob/63399759a7dd38bff471ef2df58c423a46ce4cdd/core/src/impl/Kokkos_Core.cpp#L458
which would throw an exception if the conversion was to fail.  At the moment, I am not convinced we should explicitly handle the case when the conversion cannot actually be performed.  I think the exception from `std::stoi` is good enough.

PhilMiller:
I see now that `std::stoi` throws an exception if it can't parse the input. That's more than sufficient, I think.

Issue #5502, opened 2022-09-29
Trilinos nightly test failure, cuda/11.2 Power9+Volta70 arch: defaultdevicetype.shared_space failed


ndellingwood:
**Describe the bug**
The `defaultdevicetype.shared_space` test is failing in cuda/11.2 builds of Trilinos tested with kokkos' develop branch.
Testing with Trilinos in this run is executed with `ctest -j2`, I'm not sure if that would have an impact on the test

**Output snip**:
```
20:32:12 [ RUN      ] defaultdevicetype.shared_space
20:32:12 Page size as reported by os: 65536 bytes 
20:32:12 Allocating 100 pages of memory in SharedSpace.
20:32:12 Behavior found: 
20:32:12 SharedSpace is as fast as local space on repeated access: 0, we expect true 
20:32:12 
20:32:12 Please look at the following timings. The first access in a different ExecutionSpace is not evaluated for the test. As we expect the memory to migrate during the first access it might have a higher cycle count than subsequent accesses, depending on your hardware. If the cycles are more than 1.5 times the cycles for pure local memory access, we assume a page migration happened.
20:32:12 
20:32:12 ################SHARED SPACE####################
20:32:12 DefaultExectionSpace timings of run 0:
20:32:12 TimingResult contains 10 results:
20:32:12 Duration of loop 0 is 1343 clock cycles. Migration assumed.
20:32:12 Duration of loop 1 is 1369 clock cycles. Migration assumed.
20:32:12 Duration of loop 2 is 1392 clock cycles. Migration assumed.
20:32:12 Duration of loop 3 is 1375 clock cycles. Migration assumed.
20:32:12 Duration of loop 4 is 1397 clock cycles. Migration assumed.
20:32:12 Duration of loop 5 is 1376 clock cycles. Migration assumed.
20:32:12 Duration of loop 6 is 1395 clock cycles. Migration assumed.
20:32:12 Duration of loop 7 is 1364 clock cycles. Migration assumed.
20:32:12 Duration of loop 8 is 1346 clock cycles. Migration assumed.
20:32:12 Duration of loop 9 is 1376 clock cycles. Migration assumed.
20:32:12 DefaultHostExecutionSpace timings of run 0:
20:32:12 TimingResult contains 10 results:
20:32:12 Duration of loop 0 is 8 clock cycles. 
20:32:12 Duration of loop 1 is 7 clock cycles. 
20:32:12 Duration of loop 2 is 7 clock cycles. 
[...]

PR #5577, opened 2022-10-18
Change powerPC clocktic into 64 bit version and remove 32 bit version

JBludau:
In https://github.com/kokkos/kokkos/issues/5502 we discovered that sometimes one of our tests fails on `PowerPC`. The failing test uses the `ClockTic` functionality to count the cycles that elapsed. The cycles counted as elapsed in the failing test is (in this case) `11'258'999'065'812`. The test `TestSharedSpace` uses a difference of `uint64_t` returned by the `ClockTic` to measure timings independent of the current frequency the core is running on. The differences is then added in a `parallel_reduce` to compute the number of cycles needed for the memory accesses.
As the difference of unsigned should not be affected by any overflow, I suspect something being wrong with how the `ClockTic` is read from the registers. I ended up looking at the documentation I found for the registers [here](https://www.ibm.com/docs/en/SSGH2K_13.1.3/com.ibm.compilers.aix.doc/compiler.pdf)

> ## Move to/from register functions
> **__mftb Purpose**
> Move from Time Base
> In 32-bit compilation mode, returns the lower word of the time base register. In 64-bit mode, returns the entire doubleword of the time base register.
> **Prototype**
> unsigned long __mftb (void);
> **Usage**
> In 32-bit mode, this function can be used in conjunction with the__mftbu built-in function to read the entire time base register. In 64-bit mode, the entire doubleword of the time base register is returned, so separate use of __mftbu is unnecessary
>It is recommended that you insert the __fence built-in function before and after the __mftb built-in function.

and 
> **__mftbu Purpose**
> Move from Time Base Upper
> Returns the upper word of the time base register. Prototype
unsigned int __mftbu (void);
> **Usage**
In 32-bit mode you can use this function in conjunction with the __mftb built-in function to read the entire time base register
It is recommended that you insert the __fence built-in function before and after the __mftbu built-in function.

So I tried to split the implementation for PowerPC into a 32 and 64 bit version, as the 64 bit version apparently can read the register as one `unsigned long` value while the 32 bit needs to read two registers and add them like on `x86`.
If we have anyone who is good with `asm` and knows this kind of stuff: any help/insight/thing here is greatly appreciated.
Also I might be looking in a completely wrong direction here ...

### DIFF
--- a/core/src/impl/Kokkos_ClockTic.hpp
+++ b/core/src/impl/Kokkos_ClockTic.hpp
@@ -118,17 +118,6 @@ KOKKOS_IMPL_HOST_FUNCTION inline uint64_t clock_tic_host() noexcept {
 
   return (uint64_t)cycles;
 
-#elif defined(__powerpc) || defined(__powerpc__) || defined(__POWERPC__) || \
-    defined(__ppc__)
-
-  unsigned int lower_word = 0;
-  unsigned int upper_word = 0;
-
-  asm volatile("mftb %0" : "=r"(lower_word));
-  asm volatile("mftbu %0" : "=r"(upper_word));
-
-  return (((uint64_t)upper_word) << 32 | ((uint64_t)lower_word));
-
 #else
 
   return std::chrono::high_resolution_clock::now().time_since_epoch().count();
--- a/core/src/impl/Kokkos_ClockTic.hpp
+++ b/core/src/impl/Kokkos_ClockTic.hpp
@@ -110,14 +110,24 @@ KOKKOS_IMPL_HOST_FUNCTION inline uint64_t clock_tic_host() noexcept {
 
   return ((uint64_t)a) | (((uint64_t)d) << 32);
 
-#elif defined(__powerpc) || defined(__powerpc__) || defined(__powerpc64__) || \
-    defined(__POWERPC__) || defined(__ppc__) || defined(__ppc64__)
+#elif defined(__powerpc64__) || defined(__ppc64__)
 
-  unsigned int cycles = 0;
+  unsigned long cycles = 0;
 
   asm volatile("mftb %0" : "=r"(cycles));
 
   return (uint64_t)cycles;
+
+#elif defined(__powerpc) || defined(__powerpc__) || defined(__POWERPC__) || \
+    defined(__ppc__)
+
+  unsigned int lower_word = 0;
+  unsigned int upper_word = 0;
+
+  asm volatile("mftb %0" : "=r"(lower_word));
+  asm volatile("mftbu %0" : "=r"(upper_word));
+
+  return (((uint64_t)upper_word) << 32 | ((uint64_t)lower_word));
 
 #else
 


core/src/impl/Kokkos_ClockTic.hpp
```diff
@@ -110,15 +110,25 @@ KOKKOS_IMPL_HOST_FUNCTION inline uint64_t clock_tic_host() noexcept {
...
+    defined(__ppc__)
+
+  unsigned int lower_word = 0;
+  unsigned int upper_word = 0;
+
+  asm volatile("mftb %0" : "=r"(lower_word));
+  asm volatile("mftbu %0" : "=r"(upper_word));
+
+  return (((uint64_t)upper_word) << 32 | ((uint64_t)lower_word));
+
```
PhilMiller, line 131: The 32-bit PPC code path is probably completely irrelevant - Kokkos is never expecting to be run on machines like this, since they haven't been used in HPC in probably 15+ years

core/src/impl/Kokkos_ClockTic.hpp
```diff
@@ -110,15 +110,25 @@ KOKKOS_IMPL_HOST_FUNCTION inline uint64_t clock_tic_host() noexcept {
...
 #else
 
   return std::chrono::high_resolution_clock::now().time_since_epoch().count();
```
PhilMiller, line 134: Separate from this PR, `high_resolution_clock` is not exactly deprecated, but its use is discouraged by knowledgeable folks: https://en.cppreference.com/w/cpp/chrono/high_resolution_clock

High quality software chronometry is unduly hard

core/src/impl/Kokkos_ClockTic.hpp
PhilMiller: Since we don’t have a 32-bit PPC system to test against, I’d be hesitant to introduce code for it that may have some subtle issue. Given that, I’d be more inclined to just fall back to the `#else` case and not introduce this `#elif`

JBludau: dropped in [363df9](https://github.com/kokkos/kokkos/pull/5577/commits/363df9228444cc6f2636232cfc09a62c0ec2ae5e)

core/src/impl/Kokkos_ClockTic.hpp
JBludau: thanks for creating the issue

PhilMiller (approved):

dalg24:
@ndellingwood would you please confirm this resolves #5502

ndellingwood:
@JBludau @dalg24 I'll build this PR on our Power9+Volta70 system and rerun the tests several times.
Since the failure is spurious with standalone kokkos testing we'll need to continue monitoring the test

ndellingwood:
Repeat runs of tests without failures, a good sign :)

dalg24 (approved):

PhilMiller:
Later introduction of the failing test aside, this may be worth cherry picking into 3.7.01

PR #5580, opened 2022-10-19
Enable extended lambdas support by default with CUDA

dalg24:
Rational: `--expt-extended-lambda` is not an experimental flag any more for a long time with NVCC, it is an alias for the `--extended-lambda` flag and is officially supported.  We had agreed for a while that we would use it by default once it would not be considered experimental but somehow never got around to actually implement it.  So here it is.

* Update `nvcc_wrapper` so it can handle `-[-]extended-lambda` (w/o the `expt-` prefix)
* Change `Kokkos_ENABLE_CUDA_LAMBDA` option default `OFF -> ON`
* ~~Prefer `-extended-lambda` flag (that change might be problematic if someone was using a pinned version of `nvcc_wrapper`)~~ **edit** this controversial bit was dropped from this PR

NOTE: The default value with generated makefiles has not been updated.  I am not sure how we would go about that one.


PhilMiller:
Is there an inherent reason to shift to preferring the newer form `-extended-lambda`, given that in the short term it may cause friction?

On the other hand, it would be a nice nudge for people to update a pinned/vendored `nvcc_wrapper` if they have one

PhilMiller:
I'm +1 on accepting the non-`-expt` flag and enabling the feature by default

JBludau:
I think this is more then reasonable considering it will be in a major release, even if some people have a version pinned

dalg24:
> Is there an inherent reason to shift to preferring the newer form `-extended-lambda`, given that in the short term it may cause friction?

No reason to prefer it.  I realize that might be an issue as I was opening the PR.

> On the other hand, it would be a nice nudge for people to update a pinned/vendored/snapshot `nvcc_wrapper` if they have one

Let me know if you want me to drop the commit that makes us add the non prefixed flag

PhilMiller:
Let's get the entirely non-controversial bits merged quickly, and then we can figure out the other bit

JBludau (approved):

PR #5583, opened 2022-10-19
CMakeLists.txt: Making it possible for downstream code to see the Kokkos_VERSION

csiefer2:
Change Kokkos_VERSION to global_set from SET, so it can be accessed by downstream code.


csiefer2:
> > Change Kokkos_VERSION to global_set from SET, so it can be accessed by downstream code.
> 
> So when you say "downstream" do you mean those who do `add_subdirectory(kokkos)`? Because if you do `find_package(Kokkos)` you definitely already get these variables being exported.

That's correct.  add_subdirectory is what I have in mind.

PR #5586, opened 2022-10-20
Fix Kokkos_LIFO include

kliegeois:
The file core/src/impl/Kokkos_ChaseLev.hpp does not include Kokkos_LIFO.hpp as the other files do.
While it does not prevent the compilation of the library, it is an inconsistency fixed in this PR.

### DIFF
--- a/core/src/impl/Kokkos_ChaseLev.hpp
+++ b/core/src/impl/Kokkos_ChaseLev.hpp
@@ -58,7 +58,7 @@
 #include <impl/Kokkos_LinkedListNode.hpp>  // KOKKOS_EXPECTS
 
 #include <Kokkos_Atomic.hpp>  // atomic_compare_exchange, atomic_fence
-#include "Kokkos_LIFO.hpp"
+#include <impl/Kokkos_LIFO.hpp>
 
 //----------------------------------------------------------------------------
 //----------------------------------------------------------------------------


dalg24 (approved):
Sure

Rombur (approved):

kliegeois:
@dalg24 the real reason behind this PR is that, when using a Clang LibTooling library called binder https://github.com/RosettaCommons/binder , the fact that this include uses a double quote without the relative path impl/ throws an include error which is fixed by this PR.

dalg24:
The standard says https://eel.is/c++draft/cpp.include#3
> source file is searched for in an implementation-defined manner

for the `""` form but all the implementations I know about search first in the current directory.

dalg24:
Perhaps a better read https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rs-incform which also indicates that this change in not really necessary.  In any case I will merge once the CI finished running

kliegeois:
@dalg24 I agree with that, the issue though is that the above mentioned tool requires angle bracket includes to deduce some file names (example of issue associated to that https://github.com/RosettaCommons/binder/issues/77 ).
Therefore,  in order to better use the tool, I had to write a script that loops over all the files included by the library that I am using (Trilinos/Tpetra in this case) and replace the quote include by an angular include and store them in a temporary folder. I can then use the tool that generates some .cpp files for me and then compile those files including the unmodified headers.

In this particular case, a quote include works but an angular include does not (at least not with the compilers that I tried).
In other words:
- `#include "Kokkos_LIFO.hpp"` works,
- `#include "impl/Kokkos_LIFO.hpp"` works,
- `#include <impl/Kokkos_LIFO.hpp>` works,
- `#include <Kokkos_LIFO.hpp>` does not work.

dalg24:
Not waiting for the SYCL build

PR #5595, opened 2022-10-21
SYCL: Update AOT architectures

masterleinad:
I need this after a compiler update on the testbeds. `Intel` promised to come up with something better in the future...


dalg24 (commented):
You forgot the Makefile.  Does that work with older toolchains?

masterleinad:
> You forgot the Makefile. Does that work with older toolchains?

Fixed. No, this is incompatible.

dalg24 (approved):

cz4rs (approved):

Issue #5581, opened 2022-10-19
Performance regression of Kokkos::subview (especially managed views) inside kernels


brian-kelley:
**Describe the bug**
There has been a performance regression involving ``Kokkos::subview`` inside of kernels, running on Serial. There are 2 dimensions to this:
- the strange part: it seems to be creeping up over multiple versions, instead of jumping up in one specific change
- subviewing managed vs. unmanaged views used to take the same amount of time, but it started to diverge around 3.6 release

This was originally noticed from a regression in a Trilinos performance test running on the CTS-1 clusters after the 3.7 update. This made a significant (~50%) slowdown in a nontrivial kernel. Replacing managed with unmanaged views gets back to the 3.6.1 speeds. But in both that test and this reproducer, the view's ``use_count()`` is not changing inside the kernel. So I'm not sure why managed makes a difference.

CTS-1 is the Broadwell machine used below.

```
Versions/SHAs
develop 9a8ea2f
3.7     61d7db5
3.6.1   d19aab9
3.6     b52f8c8
3.5     ae5fc64

g++ 9.1.0, Intel(R) Xeon(R) W-2155 (SKX)

develop: Normal: 1.85181 s.    Unmanaged: 1.17398 s.
3.7:     Normal: 1.26016 s.    Unmanaged: 0.93661 s.
3.6.1:   Normal: 1.38607 s.    Unmanaged: 0.91535 s.
3.6:     Normal: 1.16579 s.    Unmanaged: 1.14086 s.
3.5:     Normal: 0.928741 s.   Unmanaged: 0.93753 s.

icpc 19.1, Intel(R) Xeon(R) CPU E5-2695 v4 (BDW)

develop: Normal: 3.93745 s.    Unmanaged: 2.64626 s.
3.7:     Normal: 4.11707 s.    Unmanaged: 2.63066 s.
3.6.1:   Normal: 3.3886 s.     Unmanaged: 2.62778 s.
3.6:     Normal: 2.8558 s.     Unmanaged: 2.50448 s.
3.5:     Normal: 2.74807 s.    Unmanaged: 2.75499 s.
```

Reproducer (example/build_cmake_in_tree/cmake_example.cpp):
```
#include <Kokkos_Core.hpp>
int main()
{
  Kokkos::initialize();
  Kokkos::print_configuration(std::cout);
  {
    Kokkos::View<int*, Kokkos::DefaultExecutionSpace> v("V", 1000);
    Kokkos::Timer t;
[...]

PR #5064, opened 2022-05-26
Remove KOKKOS_THREAD_LOCAL

masterleinad:
There doesn't really seem any need to define this macro when we can use `thread_local` instead.

PR #5105, opened 2022-06-09
Add partition_space to OpenMP

Rombur:
This is the last part of #4935

The PR does the following:
 - add a new constructor for `OpenMP()`
 - use the `ExecutionSpace` in `impl_thread_pool_size()` 
 - use the new `impl_thread_pool_size()` where necessary
 - pass the `ExecutionSpace` by const reference instead of by copy in the default implementation of `partition_space`. This was already done in the backend implementations
 - add `partition_space` for the OpenMP backend. I had to modify a test because the new function emits a warning when the variadic arguments are different types. So you cannot mix `int` and `double` anymore. I don't think that's a big deal and the function is still in `Experimental`.

@dalg24 please do not merge until @crtrott  has time to take a look. It has written the `partition_space` for all the other backends and I would like to have his feedback

PR #5601, opened 2022-10-26
Avoid static/extern thread_local 

masterleinad:
Partially addressing #5581. The regression was introduced in https://github.com/kokkos/kokkos/pull/5064. https://stackoverflow.com/a/13123870 might provide some explanations.

Drive-by: Change `extern int g_openmp_hardware_max_threads;` in the same way for consistency.

### DIFF
--- a/core/src/OpenMP/Kokkos_OpenMP_Instance.hpp
+++ b/core/src/OpenMP/Kokkos_OpenMP_Instance.hpp
@@ -69,11 +69,11 @@ namespace Impl {
 
 class OpenMPInternal;
 
-extern int g_openmp_hardware_max_threads;
+inline int g_openmp_hardware_max_threads = 1;
 
-extern thread_local int t_openmp_hardware_id;
+inline thread_local int t_openmp_hardware_id = 0;
 // FIXME_OPENMP we can remove this after we remove partition_master
-extern thread_local OpenMPInternal* t_openmp_instance;
+inline thread_local OpenMPInternal* t_openmp_instance = nullptr;
 
 struct OpenMPTraits {
   static int constexpr MAX_THREAD_COUNT = 512;
--- a/core/src/OpenMP/Kokkos_OpenMP_Instance.cpp
+++ b/core/src/OpenMP/Kokkos_OpenMP_Instance.cpp
@@ -61,12 +61,6 @@
 namespace Kokkos {
 namespace Impl {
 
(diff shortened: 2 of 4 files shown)


dalg24:
Drive-by changes conflict with #5105

core/src/OpenMP/Kokkos_OpenMP_Instance.hpp
```diff
@@ -69,11 +69,11 @@ namespace Impl {
...
 class OpenMPInternal;
 
-extern int g_openmp_hardware_max_threads;
+inline int g_openmp_hardware_max_threads = 1;
 
-extern thread_local int t_openmp_hardware_id;
+inline thread_local int t_openmp_hardware_id = 0;
 // FIXME_OPENMP we can remove this after we remove partition_master
-extern thread_local OpenMPInternal* t_openmp_instance;
+inline thread_local OpenMPInternal* t_openmp_instance = nullptr;
```
crtrott, line 76: how does this work now? Doesn't each translation unit now get its own version of this??

crtrott (changes requested):
I don't understand yet how this is legal? How do these variables get set in each translation unit?

core/src/OpenMP/Kokkos_OpenMP_Instance.hpp
PhilMiller: Every object file has the variable in an initialized data section, and the linker folds them all to one instance

PhilMiller: If you know old Fortran, think `COMMON` block

dalg24 (approved):

PhilMiller:
I haven't gone to check, but if any of those variables are only accessed from within a single compilation unit, then it would be preferable to just define them as `static` within that, so that the compiler can generate the cheapest form of TLS access code.

Rombur (approved):

crtrott (approved):
OK this apparently is a C++17 language feature change that is fine with me.

crtrott:
> I haven't gone to check, but if any of those variables are only accessed from within a single compilation unit, then it would be preferable to just define them as `static` within that, so that the compiler can generate the cheapest form of TLS access code.

All of these are used from multiple (user defined) compilation units. So we could wrap them into accessor functions, but calling a non-inlined function is probably not a win over this.

PR #5146, opened 2022-06-24
parallel_scan with View as result type

masterleinad:
Fixes #1561. Based on top of #5141.

PR #5607, opened 2022-10-31
[3.7.01] fix incorrect offset in cuda parallel scan for < 4 byte types

dalg24:
Cherry-picking #5555 into RC 3.7.01

WARNING I had to do non-trivial conflict resolution in `Kokkos_Cuda_Parallel_Range.hpp`, `Kokkos_HIP_Parallel_Range.hpp`, and `Kokkos_HIP_ReduceScan.hpp` because of #5146 and #5383

Please review carefully.

### DIFF
--- a/core/src/Cuda/Kokkos_Cuda_Parallel_Range.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_Parallel_Range.hpp
@@ -465,8 +465,24 @@ class ParallelScan<FunctorType, Kokkos::RangePolicy<Traits...>, Kokkos::Cuda> {
  public:
   using pointer_type   = typename Analysis::pointer_type;
   using reference_type = typename Analysis::reference_type;
+  using value_type     = typename Analysis::value_type;
   using functor_type   = FunctorType;
   using size_type      = Cuda::size_type;
+  // Conditionally set word_size_type to int16_t or int8_t if value_type is
+  // smaller than int32_t (Kokkos::Cuda::size_type)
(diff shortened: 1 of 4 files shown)


nmm0 (changes requested):
On line [Kokkos_Cuda_ParallelRange.hpp:1060](https://github.com/dalg24/kokkos/blob/rc371-cherrypick5555-fixincorrectoffsetparallelscansmalltypescudaandhip/core/src/Cuda/Kokkos_Cuda_Parallel_Range.hpp#L1060) (I can't comment directly on the diff since the line wasn't changed) I think we need:

```cpp
m_scratch_space + (grid_x - 1) * size / sizeof(word_size_type),
              size);
```

`size` there is the value size so without this we may run into the same problem (offset of 0).

core/src/Cuda/Kokkos_Cuda_Parallel_Range.hpp
```diff
@@ -756,6 +774,21 @@ class ParallelScanWithTotal<FunctorType, Kokkos::RangePolicy<Traits...>,
   using reference_type = typename Analysis::reference_type;
```
nmm0, line 774: One more conflict resolution issue:

```cpp
using value_type     = typename Analysis::value_type;
```
Should go above `using pointer_type`

nmm0 (dismissed):

nmm0 (approved):
Should be good to go

masterleinad (approved):

Rombur (approved):

PR #5608, opened 2022-11-01
Deprecate `Kokkos_ENABLE_CUDA_UVM`

JBludau:
As we have the new alias `SharedSpace` we can deprecate `Kokkos_ENABLE_CUDA_UVM`.


Makefile.kokkos
```diff
@@ -642,6 +642,7 @@ ifeq ($(KOKKOS_INTERNAL_USE_CUDA), 1)
   endif
 
   ifeq ($(KOKKOS_INTERNAL_CUDA_USE_UVM), 1)
+	  # deprecated
```
dalg24, line 645: Use whitespaces instead of a tab

dalg24 (commented):
I am on board with the deprecation of `Kokkos_ENABLE_CUDA_UVM` in 4.0
We will need to elaborate on the why and what we want people to do instead in the documentation (we really need these deprecated code {2,3,4} pages)

I would have done this in two separate PRs as these are pretty much unrelated.
The deprecating/removing `CudaUVMSpace::available()` is uncontroversial as the function has become pointless (always returning `true`, not usable in constant expression, not provided by any other memory space, etc.)

Makefile.kokkos
dalg24, line 645:
```suggestion
    # deprecated
```

JBludau: oh, sry, new computer

dalg24 (approved):

ldh4 (approved):
I also think these should be two separate commits. One has to do with a cmake option while the other has to do with an obsolete function that just happens to be distantly relevant.

JBludau:
Split into two prs. Still working on the documentation in https://github.com/kokkos/kokkos-core-wiki/pull/201

crtrott (approved):

cmake/kokkos_arch.cmake
```diff
@@ -873,7 +873,12 @@ FOREACH (_BACKEND Cuda OpenMPTarget HIP SYCL OpenACC)
     ENDIF()
     IF (${_BACKEND} STREQUAL "Cuda")
        IF(KOKKOS_ENABLE_CUDA_UVM)
-          SET(_DEFAULT_DEVICE_MEMSPACE "Kokkos::${_BACKEND}UVMSpace")
+          MESSAGE(DEPRECATION "Setting Kokkos_ENABLE_CUDA_UVM is deprecated - use SharedSpace as MemorySpace in your code instead")
+          IF(KOKKOS_ENABLE_DEPRECATED_CODE_4)
+            SET(_DEFAULT_DEVICE_MEMSPACE "Kokkos::${_BACKEND}UVMSpace")
+          ELSE()
+            MESSAGE(FATAL_ERROR "Kokkos_ENABLE_DEPRECATED_CODE must be set to use Kokkos_ENABLE_CUDA_UVM")
```
masterleinad, line 880:
```suggestion
            MESSAGE(FATAL_ERROR "Kokkos_ENABLE_DEPRECATED_CODE_4 must be set to use Kokkos_ENABLE_CUDA_UVM")
```

JBludau: good catch

JBludau:
Please take a look at the documentation before merging
https://github.com/kokkos/kokkos-core-wiki/pull/201

JBludau:
ci seems unrelated (openmp.repeated_team_reduce)

masterleinad (approved):

cmake/kokkos_arch.cmake
```diff
@@ -873,7 +873,12 @@ FOREACH (_BACKEND Cuda OpenMPTarget HIP SYCL OpenACC)
     ENDIF()
     IF (${_BACKEND} STREQUAL "Cuda")
        IF(KOKKOS_ENABLE_CUDA_UVM)
-          SET(_DEFAULT_DEVICE_MEMSPACE "Kokkos::${_BACKEND}UVMSpace")
+          MESSAGE(DEPRECATION "Setting Kokkos_ENABLE_CUDA_UVM is deprecated - use SharedSpace as MemorySpace in your code instead")
```
PhilMiller, line 876:
```suggestion
          MESSAGE(DEPRECATION "Setting Kokkos_ENABLE_CUDA_UVM is deprecated - use the portable Kokkos::SharedSpace as an explicit memory space in your code instead")
```

PhilMiller (approved):

dalg24 (commented):
Need to coordinate with #5623
Same question here do we mean to define the option regardless of whether deprecated code 4 is enabled or not

(@JBludau I am not actually requesting any change at this time.  I am just writing things down for bookkeeping.  Maybe that is something we can harmonize as a follow up)

JBludau:
@dalg24 sure, let's harmonize it as soon as we have a verdict

PR #5612, opened 2022-11-02
Fix help text in generate_makefile

brian-kelley:
--cxxstandard values should be "17" instead of "c++17", etc.


dalg24 (approved):
Sure

Should we start spamming users of that script with a warning notice that this is not maintained and that they better start using cmake directly?

brian-kelley:
@dalg24 I have to say, I actually like this script too. I didn't know it was unmaintained. It makes configure scripts more readable than raw cmake commands.

ldh4 (approved):

Issue #5472, opened 2022-09-20
Deprecate `CudaUVMSpace::available()`


dalg24:
Rational: not portable nor really useful.  Prefer facility introduced in #5289 instead

_Originally posted by @dalg24 in https://github.com/kokkos/kokkos/pull/5441#discussion_r975336647_

JBludau:
I was thinking about proposing the deprecation in a separate pr together with some changes in examples/tutorials.

dalg24:
Sounds good

dalg24:
Done

PR #5614, opened 2022-11-04
Deprecate `CudaUVMSpace::available()`

JBludau:
`CudaUVMSpace::available` was proposed for deprecation by @dalg24 in https://github.com/kokkos/kokkos/issues/5472

Function is not available in other backends and is obsolete as it always returns `true`.


masterleinad (approved):

dalg24:
Failures are unrelated.

Issue #5596, opened 2022-10-21
Crash in Kokkos::Initialize from #4682


stanmoore1:
After updating to 3.7, we are seeing a crash in `Kokkos::Initialize` in LAMMPS with a `DEBUG` build and shared libs. I bisected the issue to https://github.com/kokkos/kokkos/pull/4682, which matches the error messages. If we use a `RELEASE` build or don't enable shared libs, then the code runs fine.

```
what():  Desul::Error: init_lock_arrays_cuda: post init kernel error(cudaErrorLaunchFailure): unspecified launch failure

***FAULT Warp Illegal Address
+++ID CUDA count 832 grid 3 block (0,0,0) thread (0,0,0) dev0/sm0/warp2/lane0
+++STACK
desul::(anonymousnamespace)::init_lock_arrays_cuda_kernel<<<(512,1,1),(256,1,1)>>>()@/lammps_master/lib/kokkos/tpls/desul/src/Lock_Array_CUDA.cpp:32

========= Invalid __global__ write of size 4
=========     at 0x000000d0 in /lammps_master/lib/kokkos/tpls/desul/src/Lock_Array_CUDA.cpp:31:desul::_GLOBAL__N__50_tmpxft_0001a4f7_00000000_7_Lock_Array_CUDA_cpp1_ii_4d2bf3a0::init_lock_arrays_cuda_kernel(void)
=========     by thread (238,0,0) in block (80,0,0)
=========     Address 0x000143b8 is out of bounds
=========     Saved host backtrace up to driver entry point at kernel launch time
```

stanmoore1:
@bathmatt

stanmoore1:
@masterleinad any update on this? Do we need to revert #4682?

masterleinad:
> @masterleinad any update on this? Do we need to revert https://github.com/kokkos/kokkos/pull/4682?

I don't really understand the issue, yet, and reverting seems to be the best option at the moment but we are still debating how to deal with it.

masterleinad:
Reproducer using `LAMMPS`:
```
git clone https://github.com/lammps/lammps.git
cd lammps
mkdir build
cd build
[...]

PR #4682, opened 2022-01-16
Avoid macros for CUDA lock arrays

masterleinad:
Trying the suggestion is #2235.

PR #5619, opened 2022-11-07
Fix initialization of Cuda lock arrays

masterleinad:
Fixes https://github.com/kokkos/kokkos/issues/5596. At the moment, reverting #4682 seems to be the best option.

### DIFF
--- a/core/src/Cuda/Kokkos_Cuda_Locks.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_Locks.hpp
@@ -39,7 +39,7 @@ struct CudaLockArrays {
 
 /// \brief This global variable in Host space is the central definition
 ///        of these arrays.
-extern CudaLockArrays g_host_cuda_lock_arrays;
+extern Kokkos::Impl::CudaLockArrays g_host_cuda_lock_arrays;
 
 /// \brief After this call, the g_host_cuda_lock_arrays variable has
 ///        valid, initialized arrays.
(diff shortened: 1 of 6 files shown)


masterleinad:
Reproducer using `LAMMPS`:
```
git clone https://github.com/lammps/lammps.git
cd lammps
mkdir build
cd build
cmake -C ../cmake/presets/basic.cmake -C ../cmake/presets/kokkos-cuda.cmake -D BUILD_SHARED_LIBS=on -DPKG_KOKKOS=ON -DCMAKE_BUILD_TYPE=DEBUG -DKokkos_ARCH_VOLTA70=ON -DKokkos_ARCH_PASCAL60=OFF ../cmake/
make -j16
cd ../bench
../build/lmp -in in.lj -k on g 1 -sf kk -pk kokkos
```

ajpowelsnl:
TODO:
  - Fix conflicts
  - Sync with Desul

PhilMiller:
Jenkins failure was irrelevant stuff in just HIP.

PhilMiller (approved):

crtrott (approved):

core/src/Cuda/Kokkos_Cuda_Locks.hpp
```diff
@@ -121,49 +123,45 @@ namespace {
...
+   putting this in an inline function will NOT do the right thing! */
+#define KOKKOS_COPY_CUDA_LOCK_ARRAYS_TO_DEVICE()                      \
+  {                                                                   \
+    if (::Kokkos::Impl::lock_array_copied == 0) {                     \
+      KOKKOS_IMPL_CUDA_SAFE_CALL(                                     \
+          cudaMemcpyToSymbol(Kokkos::Impl::g_device_cuda_lock_arrays, \
+                             &Kokkos::Impl::g_host_cuda_lock_arrays,  \
+                             sizeof(Kokkos::Impl::CudaLockArrays)));  \
+    }                                                                 \
+    lock_array_copied = 1;                                            \
```
trws, line 140: Does it fail if this is fixed to be an inline static function to preserve the capture semantics of the original macro?  As written `copy_cuda_lock_arrays_to_device()` would get de-duplicated by the linker and only actually update one copy of `g_device_cuda_lock_arrays`, rather than one per TU or shared library.

masterleinad: That seems to work!

PR #4875, opened 2022-03-15
SYCL RangePolicy: manually specify workgroup size through chunk size

masterleinad:
In some cases, selecting a custom workgroup size gives significantly better performance than relying on the compiler to choose.

PR #5625, opened 2022-11-09
SYCL: Set RangePolicy default chunk_size to 1

masterleinad:
https://github.com/kokkos/kokkos/pull/4875 allowed using `chunk_size` to force a workgroup size.
It turns out that for very large ranges, the default `chunk_size` is greater than 1 (see for the full logic below) resulting in performance regressions for some apps.
This pull request just forces the default chunk size for `SYCL` to be 1 so that the workgroup size is determined by the compiler.

### DIFF
--- a/core/src/Kokkos_ExecPolicy.hpp
+++ b/core/src/Kokkos_ExecPolicy.hpp
@@ -213,11 +213,11 @@ class RangePolicy : public Impl::PolicyTraits<Properties...> {
  private:
   /** \brief finalize chunk_size if it was set to AUTO*/
   inline void set_auto_chunk_size() {
-    // chunk_size <=1 lets the compiler choose the workgroup size when launching
-    // kernels
 #ifdef KOKKOS_ENABLE_SYCL
     if (std::is_same_v<typename traits::execution_space,
                        Kokkos::Experimental::SYCL>) {
+      // chunk_size <=1 lets the compiler choose the workgroup size when
+      // launching kernels
       m_granularity      = 1;
       m_granularity_mask = 0;
       return;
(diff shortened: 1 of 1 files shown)


Rombur (approved):

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -213,6 +213,16 @@ class RangePolicy : public Impl::PolicyTraits<Properties...> {
  private:
   /** \brief finalize chunk_size if it was set to AUTO*/
   inline void set_auto_chunk_size() {
+    // chunk_size <=1 lets the compiler choose the workgroup size when launching
+    // kernels
```
dalg24, line 217: The comment belong with the `if(exec space is SYCL)`
Please move it to L221

dalg24 (commented):
I dislike our introducing backend specific code in that code but I don't have a better idea just now.

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -213,6 +213,16 @@ class RangePolicy : public Impl::PolicyTraits<Properties...> {
  private:
   /** \brief finalize chunk_size if it was set to AUTO*/
   inline void set_auto_chunk_size() {
+#ifdef KOKKOS_ENABLE_SYCL
+    if (std::is_same_v<typename traits::execution_space,
```
cz4rs, line 217: Possibly:
```suggestion
    if constexpr (std::is_same_v<typename traits::execution_space,
```
(makes it explicit that this is a static check)

cz4rs (approved):
Looks good to me.

core/src/Kokkos_ExecPolicy.hpp
masterleinad: Hmm... I don't see much of an improvement. I'll change it if someone requests it, otherwise, I would leave it as is.

dalg24:
Merging with the understanding this is a temporary solution until we figure out how to address https://github.com/kokkos/kokkos/pull/5625#pullrequestreview-1176300959

PR #5627, opened 2022-11-10
Remove pre C++17 workarounds and use fold expressions

dalg24:
Fold expressions are supported in C++17.  This PR is removing workarounds that are not necessary any more.
* Drop comma fold emulations facility that was introduced in the implementation of the combined reducers
* Drop unnecessary `ifdef __cpp_fold_expressions` guards in `partition_space`
* Cleanup tool testing utilities

### DIFF
--- a/core/unit_test/tools/include/ToolTestingUtilities.hpp
+++ b/core/unit_test/tools/include/ToolTestingUtilities.hpp
@@ -131,7 +131,7 @@ struct function_traits<R (*)(A...)> {
   constexpr static int num_arguments = sizeof...(A);
   template <class Call, class... Args>
   static auto invoke_as(const Call& call, Args&&... args) {
-    if ((... && !std::dynamic_pointer_cast<A>(std::forward<Args>(args)))) {
+    if ((!std::dynamic_pointer_cast<A>(std::forward<Args>(args)) && ...)) {
       return MatchDiagnostic{false, {"Types didn't match on arguments"}};
     }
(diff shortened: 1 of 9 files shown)


PhilMiller (approved):

masterleinad (approved):

dalg24:
Retest this please

core/unit_test/tools/include/ToolTestingUtilities.hpp
```diff
@@ -153,7 +131,7 @@ struct function_traits<R (*)(A...)> {
   constexpr static int num_arguments = sizeof...(A);
   template <class Call, class... Args>
   static auto invoke_as(const Call& call, Args&&... args) {
-    if (!are_valid(std::dynamic_pointer_cast<A>(std::forward<Args>(args))...)) {
+    if ((!std::dynamic_pointer_cast<A>(std::forward<Args>(args)) && ...)) {
```
PhilMiller, line 134: This change looks like a mixup of where the negation is applied. It's now testing that all the pointers are the wrong type, rather than any of them

PhilMiller (changes requested):

core/unit_test/tools/include/ToolTestingUtilities.hpp
dalg24: Thx.  It's interesting that virtually all the builds had all test passings with that bug...

PhilMiller (approved):

crtrott (approved):

Issue #2478, opened 2019-10-17
query MPI context better


jeffhammond:
You have three instances of the same pattern:
* https://github.com/kokkos/kokkos/blob/develop/core/src/impl/Kokkos_Core.cpp#L101
* https://github.com/kokkos/kokkos/blob/develop/core/src/impl/Kokkos_CPUDiscovery.cpp#L88
* https://github.com/kokkos/kokkos/blob/develop/example/multi_fem/ParallelMachine.cpp#L82

I don't care what you do in examples but [Kokkos_Core.cpp](https://github.com/kokkos/kokkos/blob/develop/core/src/impl/Kokkos_Core.cpp#L101) should use `Kokkos::Impl:: mpi_ranks_per_node()` and/or `Kokkos::Impl::mpi_local_rank_on_node` instead of reinventing the wheel.

Regardless of how many times or where you do it, the right way to query the local MPI environment is `MPI_Comm_split_type` but that means you have a dependency on MPI that you don't want.  At the very least, you should support other MPI libraries.

I will try to work on this, but would like feedback from @crtrott or other authority on the MPI library aspect before proceeding with refactoring the implementation.

crtrott:
I really don't want a dependency on MPI. Technically this is more of a convenience thing anyway. It would be better for people to explicitly tell Kokkos initialize which device to use after they initialized MPI. That said,  I know how convenience crutches become often the defacto standard. I put it on the agenda for our next meeting and see what other people think.

PhilMiller:
@dalg24 @crtrott This hole was dug deeper in #5525

jeffhammond:
> I really don't want a dependency on MPI.

There is `dlopen` if you want to make MPI optional.

jeffhammond:
Have you tried this with MPI_Comm_spawn?

How about MPI-4 sessions, which was driven into MPI by LANL, among others?

What happens when someone uses Kokkos Remote Spaces and the NVSHMEM back end?

jeffhammond:
My point was that this method will be incorrect in the context dynamic process management, and anything like it.  Your users might not care, but it's not portable.

dalg24:
Were you willing to explore the dlopen route?

jeffhammond:
well, i just spent a month writing a dlopen wrapper for everything in MPI so i certainly know how to do it (https://github.com/jeffhammond/mukautuva/blob/main/libinit.c).

if you are interested in going the dlopen route here, i can work on it later.  probably April.

PR #5630, opened 2022-11-11
Reuse code for detecting MPI context

dalg24:
Address https://github.com/kokkos/kokkos/issues/2478#issuecomment-1311114148
Avoid duplicating the logic in our code base

### DIFF
--- a/core/src/OpenMP/Kokkos_OpenMP_Instance.cpp
+++ b/core/src/OpenMP/Kokkos_OpenMP_Instance.cpp
@@ -338,11 +338,8 @@ void OpenMPInternal::initialize(int thread_count) {
   }
 
   // Check for over-subscription
-  auto const mpi_local_size = []() {
-    auto ranks = mpi_ranks_per_node();
-    if (ranks < 0) return 1;
-    return ranks;
-  }();
+  auto const reported_ranks = mpi_ranks_per_node();
+  auto const mpi_local_size = reported_ranks < 0 ? 1 : reported_ranks;
   if (Kokkos::show_warnings() &&
       (mpi_local_size * long(thread_count) > Impl::processors_per_node())) {
     std::cerr << "Kokkos::OpenMP::initialize WARNING: You are likely "
--- a/core/src/OpenMP/Kokkos_OpenMP_Instance.cpp
+++ b/core/src/OpenMP/Kokkos_OpenMP_Instance.cpp
@@ -338,17 +338,20 @@ void OpenMPInternal::initialize(int thread_count) {
   }
 
   // Check for over-subscription
+  auto const mpi_local_size = []() {
+    auto ranks = mpi_ranks_per_node();
+    if (ranks < 0) return 1;
+    return ranks;
+  }();
   if (Kokkos::show_warnings() &&
-      (Impl::mpi_ranks_per_node() * long(thread_count) >
-       Impl::processors_per_node())) {
+      (mpi_local_size * long(thread_count) > Impl::processors_per_node())) {
     std::cerr << "Kokkos::OpenMP::initialize WARNING: You are likely "
                  "oversubscribing your CPU cores."
               << std::endl;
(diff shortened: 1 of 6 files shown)


core/src/OpenMP/Kokkos_OpenMP_Instance.cpp
```diff
@@ -338,17 +338,20 @@ void OpenMPInternal::initialize(int thread_count) {
   }
 
   // Check for over-subscription
+  auto const mpi_local_size = []() {
+    auto ranks = mpi_ranks_per_node();
+    if (ranks < 0) return 1;
+    return ranks;
+  }();
```
PhilMiller, line 345: Why the unusual arrangement of calling a one-off lambda here and in Threads?

core/src/impl/Kokkos_CPUDiscovery.cpp
```diff
@@ -46,6 +46,8 @@
 #define KOKKOS_IMPL_PUBLIC_INCLUDE
 #endif
 
+#include <impl/Kokkos_CPUDiscovery.hpp>
+
 #ifdef _WIN32
 #define WIN32_LEAN_AND_MEAN
 #include <windows.h>
```
PhilMiller, line 53: It would seem that we're not actually doing anything for the Windows case right now.

https://stackoverflow.com/questions/22767620/mingw-missing-sysconf-sc-nprocessors-onln
Has a suggestion for the rough equivalent to what the current code is doing.

core/src/OpenMP/Kokkos_OpenMP_Instance.cpp
dalg24: We have a fair amount of immediately invoked lambdas throughout the code base.  It enables declaring the variable as `const`.  I don't feel strongly about it if you want to suggest changes.

As a side note.  I expect we will eventually refactor the host parallel backends initialization and share more facility between the OpenMP and Threads backends.  The checking for over-subscription is an obvious candidate for that shared facility.

core/src/impl/Kokkos_CPUDiscovery.cpp
dalg24: This is a good point but this would need to be handled elsewhere.

core/src/OpenMP/Kokkos_OpenMP_Instance.cpp
PhilMiller, line 345:
```suggestion
  const auto reported_ranks = mpi_ranks_per_node();
  const auto mpi_local_size = reported_ranks < 0 ? 1 : reported_ranks;
```

This seems much less roundabout

If C++ were a more expression-oriented language, this would of course be less troublesome.

I'm just concerned about generated code size and representation, if the compiler isn't aggressive about inlining constructs like the lambda.

dalg24: > I'm just concerned about generated code size and representation, if the compiler isn't aggressive about inlining constructs like the lambda.

I am not convinced there is anything to worry about here.
https://godbolt.org/z/3c79KeKb8

core/src/impl/Kokkos_CPUDiscovery.cpp
PhilMiller: Could you please open an issue for that before this gets merged?

PhilMiller (approved):

PhilMiller:
CI failure was a single timing failure on HIP. It can be ignored.

core/src/impl/Kokkos_CPUDiscovery.cpp
dalg24: Done (#5631)

PhilMiller:
retest this please

PhilMiller:
On retest, the jenkins failure was in a CUDA timing-based test.

crtrott (approved):

PR #5640, opened 2022-11-23
Update copyright

crtrott:


crtrott:
OK I think its good. Here is my check mechanism.
- check the first 3 commits, which make the copyright stuff more uniform
- generate a diff for the last 7 commits: `git diff HEAD~7 &> whole_diff`
- filter out all lines coming from "diff" itself or are known to be ok; that is they start with one of:
  - `diff `, `index `, `+++ `, `--- `, `@@ `, `+//`, `-//`, `-/*`, `-*/`
```
cat whole_diff | sed '/^diff /d' | sed '/^index /d'| sed '/^+++ /d' | sed '/^--- /d' | sed '/^@@ /d' | sed '/^+\/\//d' | sed '/^-\/\//d' | sed '/^-\/\*/d' | sed '/^-\*\//d' &> whole_diff_small
``` 

Then check `whole_diff_small` for '+' and '-' to see other changes.

LICENSE_FILE_HEADER
```diff
@@ -0,0 +1,15 @@
...
+//
+//                        Kokkos v. 4.0
+//       Copyright (2022) National Technology & Engineering
+//               Solutions of Sandia, LLC (NTESS).
+//
+// Under the terms of Contract DE-NA0003525 with NTESS,
+// the U.S. Government retains certain rights in this software.
+//
+// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
+//
```
dalg24, line 12: Should we add a reference to the actual license file here?
If not a URL then state that the file is to be found at the root of the Kokkos source dir

LICENSE_FILE_HEADER
```diff
@@ -0,0 +1,15 @@
+//@HEADER
+// ************************************************************************
+//
+//                        Kokkos v. 4.0
+//       Copyright (2022) National Technology & Engineering
+//               Solutions of Sandia, LLC (NTESS).
+//
+// Under the terms of Contract DE-NA0003525 with NTESS,
+// the U.S. Government retains certain rights in this software.
```
dalg24, line 9: Are we required to keep these lines in every file or can we get away with the SPDX ID?

scripts/check-copyright
```diff
@@ -0,0 +1,11 @@
+files=`find . -path './tpls' -prune -o -name '*.cpp' -o -name '*.pc.in' -o -name '*.hpp' -o -name '*.cc'`
```
dalg24, line 1:
```
cmake/compile_tests/cuda_compute_capability.cc
tpls/gtest/gtest/gtest-all.cc
```
I am tempted to say let us just rename `cuda_compute_capability.{cc -> cpp}` and drop the `*.cc`

Also why `*.pc.in`?

scripts/check-copyright
```diff
@@ -0,0 +1,11 @@
...
+echo "" &> scripts/diff_files
+for file in $files; do
+  head -n +15 $file &> header
```
dalg24, line 4: probably should use `wc -l LICENSE_FILE_HEADER` instead of a magic number

dalg24:
Retest this please

LICENSE_FILE_HEADER
crtrott: yeah we could add that, just wasn't sure what a stable URL for that is?

LICENSE_FILE_HEADER
crtrott: Yeah I think we need those.

scripts/check-copyright
crtrott: not sure, was in the file from before.

scripts/check-copyright
crtrott: yeah I guess we could do that.

LICENSE_FILE_HEADER
dalg24: I think the best would be to upload to kokkos.org

scripts/diff_files
```diff
@@ -0,0 +1,3 @@
+
+./tpls
+./core/src/Cuda/Kokkos_Cuda_Atomic_Intrinsics.hpp
```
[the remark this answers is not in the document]
crtrott: yes, that is the files of expected diffs.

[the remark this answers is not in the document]
dalg24: Can you elaborate about the adding the build directory?   What (presumably generated) files were problematic?
I am asking because if these files are being installed we might still want the copyright in them.

Note that this is interesting because my initial reaction was "if you used `git ls-files`" like I suggested that would not be an issue, but it would not help if we really wanted to enforce that all installed files have our copyright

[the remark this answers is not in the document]
crtrott: It is an interesting question, for now the problem was that it did find stuff like some config files which weren't correct and the way our CI is set up it created a build directory inside the checked out directory.

crtrott:
We should merge this so we can update all PRs which add new files.

dalg24 (approved):

Issue #5638, opened 2022-11-22
provide wrappers for native functions supported by Cuda


XzzX:
Cuda provides more device optimized math functions (https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#mathematical-functions-appendix) than currently supported by Kokkos (#4767). In principle they can be provided if available and otherwise a fallback will be used.

XzzX:
I can provide a MR, however, I have some questions regarding the implementation.  @dalg24 

Currently, a namespace is set depending on the enabled backend and functions from this namespace are used? Does this work if I enable some device backend and afterwards execute on the host? Or do I end up with device functions on host?  

Do you want it to be implemented similar to `abs` by providing overloads for a certain set of argument types?

dalg24:
You probably will need to use the `KOKKOS_IF_ON_{HOST,DEVICE}` macros.

crtrott:
Core is not there fft support belongs so no - fft doesn't go heee.

XzzX:
The unit tests check against a reference function in `std` namespace. What do you want to do if the function is not available in `std` namespace? Inject a dummy function into `std` namespace? Since the unit tests are defined in a header file, I have some headaches with this.

dalg24:
No you will need to have a different logic for these functions
```C++
KOKKOS_INLINE_FUNCTION double rcbrt(double x) {
#if defined(KOKKOS_ENABLE_HIP) && defined(KOKKOS_ENABLE_CUDA)
  KOKKOS_IF_ON_DEVICE( return ::rcbrt(x); )
  KOKKOS_IF_ON_HOST( ... )
#else
  ...
#endif
}

dalg24:
Do not define any function in the `std::` namespace.  Yes we need custom logic for these new functions.

ajpowelsnl:
@XzzX -- Have we addressed your needs with the merging of #5644?

PR #5644, opened 2022-11-29
Add non-standard `rsqrt` math function

XzzX:
PR for #5638  

Also related to #4767

### DIFF
--- a/core/src/Kokkos_MathematicalFunctions.hpp
+++ b/core/src/Kokkos_MathematicalFunctions.hpp
@@ -402,23 +402,6 @@ KOKKOS_IMPL_MATH_UNARY_FUNCTION(log1p)
 // Power functions
 KOKKOS_IMPL_MATH_BINARY_FUNCTION(pow)
 KOKKOS_IMPL_MATH_UNARY_FUNCTION(sqrt)
-KOKKOS_INLINE_FUNCTION float rsqrt(float val) {
-#if defined(KOKKOS_ENABLE_CUDA)
-  return rsqrtf(val);
-#else
-  return 1.0f / Kokkos::sqrt(val);
-#endif
-}
-KOKKOS_INLINE_FUNCTION double rsqrt(double val) {
-#if defined(KOKKOS_ENABLE_CUDA)
-  return rsqrt(val);
-#else
-  return 1.0 / Kokkos::sqrt(val);
-#endif
-}
-inline long double rsqrt(long double val) {
-  return 1L / Kokkos::sqrt(val);
-}
 KOKKOS_IMPL_MATH_UNARY_FUNCTION(cbrt)
 KOKKOS_IMPL_MATH_BINARY_FUNCTION(hypot)
 #if defined(KOKKOS_ENABLE_CUDA) || defined(KOKKOS_ENABLE_HIP) || \
@@ -529,6 +512,34 @@ KOKKOS_IMPL_MATH_UNARY_PREDICATE(signbit)
 #undef KOKKOS_IMPL_MATH_UNARY_PREDICATE
 #undef KOKKOS_IMPL_MATH_BINARY_FUNCTION
 #undef KOKKOS_IMPL_MATH_TERNARY_FUNCTION
+
+// non-standard math functions provided by CUDA/HIP/SYCL
+KOKKOS_INLINE_FUNCTION float rsqrt(float val) {
+#if defined(KOKKOS_ENABLE_CUDA) || defined(KOKKOS_ENABLE_HIP)
+  return rsqrtf(val);
+#elif defined(KOKKOS_ENABLE_SYCL)
+  return sycl::rsqrt(val);
+#else
+  return 1.0f / Kokkos::sqrt(val);
+#endif
+}
+KOKKOS_INLINE_FUNCTION double rsqrt(double val) {
+#if defined(KOKKOS_ENABLE_CUDA) || defined(KOKKOS_ENABLE_HIP)
+  return rsqrt(val);
+#elif defined(KOKKOS_ENABLE_SYCL)
+  return sycl::rsqrt(val);
+#else
+  return 1.0 / Kokkos::sqrt(val);
+#endif
+}
+inline long double rsqrt(long double val) { return 1l / Kokkos::sqrt(val); }
+KOKKOS_INLINE_FUNCTION float rsqrtf(float x) { return Kokkos::rsqrt(x); }
+inline long double rsqrtl(long double x) { return Kokkos::rsqrt(x); }
+template <class T>
+KOKKOS_INLINE_FUNCTION std::enable_if_t<std::is_integral_v<T>, double> rsqrt(
+    T x) {
+  return Kokkos::rsqrt(static_cast<double>(x));
+}
 
 }  // namespace Kokkos
 
--- a/core/src/Kokkos_MathematicalFunctions.hpp
+++ b/core/src/Kokkos_MathematicalFunctions.hpp
@@ -532,7 +532,7 @@ KOKKOS_INLINE_FUNCTION double rsqrt(double val) {
   return 1.0 / Kokkos::sqrt(val);
 #endif
 }
-inline long double rsqrt(long double val) { return 1l / Kokkos::sqrt(val); }
+inline long double rsqrt(long double val) { return 1.0l / Kokkos::sqrt(val); }
 KOKKOS_INLINE_FUNCTION float rsqrtf(float x) { return Kokkos::rsqrt(x); }
 inline long double rsqrtl(long double x) { return Kokkos::rsqrt(x); }
 template <class T>
--- a/core/src/Kokkos_MathematicalFunctions.hpp
+++ b/core/src/Kokkos_MathematicalFunctions.hpp
@@ -516,7 +516,7 @@ KOKKOS_IMPL_MATH_UNARY_PREDICATE(signbit)
 // non-standard math functions provided by CUDA/HIP/SYCL
 KOKKOS_INLINE_FUNCTION float rsqrt(float val) {
 #if defined(KOKKOS_ENABLE_CUDA) || defined(KOKKOS_ENABLE_HIP)
-  return rsqrtf(val);
+  return ::rsqrtf(val);
 #elif defined(KOKKOS_ENABLE_SYCL)
   return sycl::rsqrt(val);
 #else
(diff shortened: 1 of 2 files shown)


XzzX:
Cuda math functions seemingly also work in host code.

dalg24:
OK to test

XzzX:
I propose renaming `DEFINE_UNARY_FUNCTION_EVAL` to `DEFINE_UNARY_FUNCTION_EVAL_STD` and `eval_std` to `eval_ref`.

XzzX:
@dalg24 What do you think about the structure? If this is ok, I can include more overloads and also the other functions and HIP support.

core/src/Kokkos_MathematicalFunctions.hpp
```diff
@@ -402,6 +402,23 @@ KOKKOS_IMPL_MATH_UNARY_FUNCTION(log1p)
...
+}
+KOKKOS_INLINE_FUNCTION double rsqrt(double val) {
+#if defined(KOKKOS_ENABLE_CUDA)
+  return rsqrt(val);
+#else
+  return 1.0 / Kokkos::sqrt(val);
+#endif
+}
+inline long double rsqrt(long double val) {
+  return 1L / Kokkos::sqrt(val);
```
dalg24, line 420:
```suggestion
  return 1.0l / Kokkos::sqrt(val);
```
`1L` is `long int`

dalg24 (commented):
A few questions/comments
* Did you plan to introduce `rcbrt`?  `rhypot`?  (these are provided by CUDA/HIP)
  * SYCL provides `rsqrt` too (not `rcbrt` nor `rhypot`)
* We also need to consider providing 
  * a set of overloads or a function template accepting arguments of integral types
  `rsqrt(IntegralType) -> double`
  * suffixed versions
  `rsqrtf(float) -> float` and `rsqrtl(long double) -> long double`
* We should probably group these elsewhere instead of interleaving them between the standard functions

core/src/Kokkos_MathematicalFunctions.hpp
XzzX: I can provide all functions that have a simple enough `std` fallback representation.  

additional overloads -> done  
grouping -> done

XzzX: done

XzzX:
> A few questions/comments
> 
> * Did you plan to introduce `rcbrt`?  `rhypot`?  (these are provided by CUDA/HIP)
>   
>   * SYCL provides `rsqrt` too (not `rcbrt` nor `rhypot`)
> * We also need to consider providing
>   
>   * a set of overloads or a function template accepting arguments of integral types
>     `rsqrt(IntegralType) -> double`
>   * suffixed versions
>     `rsqrtf(float) -> float` and `rsqrtl(long double) -> long double`
> * We should probably group these elsewhere instead of interleaving them between the standard functions

I can provide all functions that have a simple enough std fallback representation.

additional overloads -> done
grouping -> done

core/unit_test/TestMathematicalFunctions.hpp
```diff
@@ -893,6 +919,20 @@ TEST(TEST_CATEGORY, mathematical_functions_hyperbolic_functions) {
...
 #endif
 }
+
+TEST(TEST_CATEGORY, mathematical_functions_non_standard) {
+  TEST_MATH_FUNCTION(rsqrt)({0, 1, 2, 3, 5, 7, 11});
+  TEST_MATH_FUNCTION(rsqrt)({0l, 1l, 2l, 3l, 5l, 7l, 11l});
+  TEST_MATH_FUNCTION(rsqrt)({0ll, 1ll, 2ll, 3ll, 5ll, 7ll, 11ll});
+  TEST_MATH_FUNCTION(rsqrt)({0u, 1u, 2u, 3u, 5u, 7u});
+  TEST_MATH_FUNCTION(rsqrt)({0ul, 1ul, 2ul, 3ul, 5ul, 7ul});
+  TEST_MATH_FUNCTION(rsqrt)({0ull, 1ull, 2ull, 3ull, 5ull, 7ull});
```
dalg24, line 929:
```suggestion
  TEST_MATH_FUNCTION(rsqrt)({1, 2, 3, 5, 7, 11});
  TEST_MATH_FUNCTION(rsqrt)({1l, 2l, 3l, 5l, 7l, 11l});
  TEST_MATH_FUNCTION(rsqrt)({1ll, 2ll, 3ll, 5ll, 7ll, 11ll});
  TEST_MATH_FUNCTION(rsqrt)({1u, 2u, 3u, 5u, 7u});
  TEST_MATH_FUNCTION(rsqrt)({1ul, 2ul, 3ul, 5ul, 7ul});
  TEST_MATH_FUNCTION(rsqrt)({1ull, 2ull, 3ull, 5ull, 7ull});
```
(dividing by zero)

core/src/Kokkos_MathematicalFunctions.hpp
```diff
@@ -513,6 +513,34 @@ KOKKOS_IMPL_MATH_UNARY_PREDICATE(signbit)
...
+KOKKOS_INLINE_FUNCTION double rsqrt(double val) {
+#if defined(KOKKOS_ENABLE_CUDA) || defined(KOKKOS_ENABLE_HIP)
+  return rsqrt(val);
+#elif defined(KOKKOS_ENABLE_SYCL)
+  return sycl::rsqrt(val);
+#else
+  return 1.0 / Kokkos::sqrt(val);
+#endif
+}
+inline long double rsqrt(long double val) { return 1l / Kokkos::sqrt(val); }
```
dalg24, line 535: Same comment `1l` is `long int` not `long double`.  You probably meant `1.l` (or `1.0l` for consistency with your using `1.0f` and `1.0` in the previous overloads)

core/unit_test/TestMathematicalFunctions.hpp
XzzX: Yes, good point. Didn't think about that.

core/unit_test/TestMathematicalFunctions.hpp
```diff
@@ -338,6 +363,7 @@ DEFINE_UNARY_FUNCTION_EVAL(log2, 2);
 DEFINE_UNARY_FUNCTION_EVAL(log1p, 2);
 
 DEFINE_UNARY_FUNCTION_EVAL(sqrt, 2);
+DEFINE_UNARY_FUNCTION_EVAL_CUSTOM(rsqrt, 2, 1 / std::sqrt(x));
```
XzzX, line 366: The type of `1` here is not clean. `decltype(x)(1)` will fix it for `float` and `double` but not for integer types. Nevertheless, it will work because `std::sqrt` returns `double`. Any suggestions?

XzzX: `decltype(std::sqrt(x))(1) / std::sqrt(x)` should do the trick.

XzzX:
@dalg24 If you are fine with the current structure I will add more functions.

XzzX:
The current CI error seems strange to me.

```
+ ./scripts/docker/check_format_cpp.sh
awk: read error (Is a directory)
```

crtrott:
Thiss seems to be wrong?

```
/var/jenkins/workspace/Kokkos/core/src/Kokkos_MathematicalFunctions.hpp:526:49: error: all paths through this function will call itself [clang-diagnostic-infinite-recursion]

KOKKOS_INLINE_FUNCTION double rsqrt(double val) {
```

core/src/Kokkos_MathematicalFunctions.hpp
```diff
@@ -513,6 +513,34 @@ KOKKOS_IMPL_MATH_UNARY_PREDICATE(signbit)
...
+  return rsqrtf(val);
+#elif defined(KOKKOS_ENABLE_SYCL)
+  return sycl::rsqrt(val);
+#else
+  return 1.0f / Kokkos::sqrt(val);
+#endif
+}
+KOKKOS_INLINE_FUNCTION double rsqrt(double val) {
+#if defined(KOKKOS_ENABLE_CUDA) || defined(KOKKOS_ENABLE_HIP)
+  return rsqrt(val);
```
crtrott, line 528: This probably recurses: try calling `::rsqrt`?

crtrott (changes requested):

dalg24:
Retest this please

XzzX:
Is there interest in implementing more functions like this or do you want to merge it like this? I only need `rsqrt`.

core/unit_test/TestMathematicalFunctions.hpp
```diff
@@ -320,6 +320,31 @@ struct math_function_name;
...
+  struct MathUnaryFunction_##FUNC {                                   \
+    template <typename T>                                             \
+    static KOKKOS_FUNCTION auto eval(T x) {                           \
+      static_assert(                                                  \
+          std::is_same<decltype(Kokkos::FUNC((T)0)),                  \
+                       math_unary_function_return_type_t<T>>::value); \
+      return Kokkos::FUNC(x);                                         \
+    }                                                                 \
+    template <typename T>                                             \
+    static auto eval_std(T x) {                                       \
```
dalg24, line 333: is never called

core/unit_test/TestMathematicalFunctions.hpp
```diff
@@ -320,6 +320,31 @@ struct math_function_name;
...
+      static_assert(                                                  \
+          std::is_same<decltype(REF_FUNC),                            \
+                       math_unary_function_return_type_t<T>>::value); \
+      return REF_FUNC;                                                \
```
dalg24, line 337: passing `decltype(std::sqrt(x))(1) / std::sqrt(x)` below which is a type, not a value

dalg24 (changes requested):
eval_std is

core/unit_test/TestMathematicalFunctions.hpp
XzzX: It is called via https://github.com/kokkos/kokkos/blob/90bf117bc9890453018f62853cc15473f7d3324e/core/unit_test/TestMathematicalFunctions.hpp#L509

core/unit_test/TestMathematicalFunctions.hpp
XzzX: This can be evaluated to a value. ?

core/unit_test/TestMathematicalFunctions.hpp
XzzX: Regarding the macro and function naming please also see my previous comment, https://github.com/kokkos/kokkos/pull/5644#issuecomment-1330928711

core/unit_test/TestMathematicalFunctions.hpp
dalg24: Sorry clearly did not read carefully and thought the decltype encompassed the entire expression.
Which led me to think that the code could not possibly compile...

dalg24 (approved):

Rombur (approved):

masterleinad (approved):
I think the `Kokkos` namespace is fine.

Issue #5631, opened 2022-11-15
Need to revisit detection of number of processors on Windows


dalg24:
It would seem that we're not actually doing anything for the Windows case right now.
https://github.com/kokkos/kokkos/blob/c4df87cd35b162620c8e0f294dffa25980f35fa7/core/src/impl/Kokkos_CPUDiscovery.cpp#L51-L53
https://github.com/kokkos/kokkos/blob/c4df87cd35b162620c8e0f294dffa25980f35fa7/core/src/impl/Kokkos_CPUDiscovery.cpp#L82-L84

https://stackoverflow.com/questions/22767620/mingw-missing-sysconf-sc-nprocessors-onln
Has a suggestion for the rough equivalent to what the current code is doing.

_Originally posted by @PhilMiller in https://github.com/kokkos/kokkos/pull/5630#discussion_r1020524656_

dalg24:
Is there a good reason not to just use `std::thread::hardware_concurrency()`?

PhilMiller:
It if works, that sounds like it would be great. Ideally, it could be used to replace `processors_per_node` entirely

mhoemmen:
> Is there a good reason not to just use `std::thread::hardware_concurrency()`?

Works pretty well for me with MSVC: https://godbolt.org/z/sEPa5GaTK

I haven't followed the latest MSVC, but I know they were doing a lot of work on their OpenMP implementation.  That might change how you would want to implement processor discovery.

PR #5645, opened 2022-11-30
Prefer C++11 built-in support for threads to detect the number of processors

dalg24:
Proposed resolution for #5631 
Drop `Impl::processors_per_node()` and use `std::thread::hardware_concurrency` instead.

One concern we had was that GCC requires linking against pthread to use `std::thread` but it turns out it is not needed for `hardware_concurrency()` so I can't think of any good reason not to use it instead of implementing our own facility to detect the number of processors.

Note that we should consider renaming `Kokkos_CPUDiscovery.{hpp,cpp}` to something more sensible but I decided to abstain for now so I don't create conflicts with #5640

### DIFF
--- a/core/src/OpenMP/Kokkos_OpenMP_Instance.cpp
+++ b/core/src/OpenMP/Kokkos_OpenMP_Instance.cpp
@@ -57,6 +57,7 @@
 #include <iomanip>
 #include <iostream>
 #include <sstream>
+#include <thread>
 
 namespace Kokkos {
 namespace Impl {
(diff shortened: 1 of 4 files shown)


PhilMiller:
The code change looks good. We'll need some sort of explicit confirmation that it's not changing behavior from what we get now on Linux and macOS.

dalg24:
> The code change looks good. We'll need some sort of explicit confirmation that it's not changing behavior from what we get now on Linux and macOS.

On my MacBook that has the Apple M1 pro chip, using AppleClang I get the same number of processors (10).
I didn't actually check on Linux but I looked at the [libcxx implementation](https://github.com/llvm/llvm-project/blob/bb1eca6bbb470e4fea111f4a6929be6f7a4524b3/libcxx/src/thread.cpp#L73-L99) and they also use `sysconf(_SC_NPROCESSORS_ONLN )`.  The [libstdc++ implementation](https://github.com/gcc-mirror/gcc/blob/f57ff1895725753f325552fc95f0baf8dd708f96/libstdc%2B%2B-v3/src/c%2B%2B11/thread.cc#L196-L203) is more convoluted (a lot of `#ifdef` that are [not super easy to follow](https://github.com/gcc-mirror/gcc/blob/f57ff1895725753f325552fc95f0baf8dd708f96/libstdc%2B%2B-v3/src/c%2B%2B11/thread.cc#L47-L82)).

cz4rs (approved):
Looks good to me, great +/- ratio too :)

ldh4 (approved):

PhilMiller (approved):

PR #5650, opened 2022-11-30
Define new type trait to identify cuda-type execution space in deep copy

ElisabethGiem:
I am creating a resilient Cuda execution space. Currently, a DeepCopy from a resilient Cuda space (ResCudaSpace) to a ResCudaSpace using a resilient Cuda execution space will get caught by the template `template <class MemSpace1, class MemSpace2, class ExecutionSpace>` that filters `ExecutionSpace` as `!Cuda`.

I need it to filter into the template that has the `ExecutionSpace ` type `Cuda`. This could be achieved by changing the filter from `std::enable_if_t<!std::is_same<ExecutionSpace, Cuda>::value>` to `std::enable_if_t<!std::derived_from<ExecutionSpace, Cuda>::value>`.

It has been mentioned that this might be a suboptimal extension point from a Trilinos perspective, so I propose this type trait to identify a Cuda type execution space. There is already a type trait to identify a Cuda type memory space and this follows a similar pattern. This would keep the modification in `Impl `and limit the impact: Only the current `Cuda` space and the `ResCuda` execution space need be defined as Cuda type execution spaces.

### DIFF
--- a/core/src/Kokkos_Concepts.hpp
+++ b/core/src/Kokkos_Concepts.hpp
@@ -447,6 +447,20 @@ struct SpaceAccessibility {
 
 }  // namespace Kokkos
 
+namespace Kokkos {
+namespace Impl{
+
+//! Grant Kokkos Resilience permission to specialize this trait to control the behavior of deep_copy
+template <class ExecutionSpace, class Enabled = void >
+struct ExecutionSpaceAlias{
+
+  using type = ExecutionSpace;
+
+};
+
+}
+}
+
 //----------------------------------------------------------------------------
 
 #endif  // KOKKOS_CORE_CONCEPTS_HPP
--- a/core/src/Kokkos_Concepts.hpp
+++ b/core/src/Kokkos_Concepts.hpp
@@ -450,11 +450,10 @@ struct SpaceAccessibility {
 namespace Kokkos {
 namespace Impl {
 
-//! Grant Kokkos Resilience permission to specialize this trait to control 
+//! Grant Kokkos Resilience permission to specialize this trait to control
 //! the behavior of deep_copy
 template <class ExecutionSpace, class Enabled = void>
 struct ExecutionSpaceAlias {
-
   using type = ExecutionSpace;
 };
 
--- a/core/src/Kokkos_Concepts.hpp
+++ b/core/src/Kokkos_Concepts.hpp
@@ -448,19 +448,18 @@ struct SpaceAccessibility {
 }  // namespace Kokkos
 
 namespace Kokkos {
-namespace Impl{
+namespace Impl {
 
-//! Grant Kokkos Resilience permission to specialize this trait to control the behavior of deep_copy
-template <class ExecutionSpace, class Enabled = void >
-struct ExecutionSpaceAlias{
+//! Grant Kokkos Resilience permission to specialize this trait to control 
+//! the behavior of deep_copy
+template <class ExecutionSpace, class Enabled = void>
+struct ExecutionSpaceAlias {
 
   using type = ExecutionSpace;
-
 };
 
-}
-}
-
+}  // namespace Impl
+}  // namespace Kokkos
 //----------------------------------------------------------------------------
 
 #endif  // KOKKOS_CORE_CONCEPTS_HPP
--- a/core/src/Kokkos_Cuda.hpp
+++ b/core/src/Kokkos_Cuda.hpp
@@ -81,7 +81,6 @@ struct CudaDispatchProperties {
 };
 }  // namespace Experimental
 }  // namespace Impl
-
 /// \class Cuda
 /// \brief Kokkos Execution Space that uses CUDA to run on GPUs.
 ///
--- a/core/src/Kokkos_Cuda.hpp
+++ b/core/src/Kokkos_Cuda.hpp
@@ -81,6 +81,7 @@ struct CudaDispatchProperties {
 };
 }  // namespace Experimental
 }  // namespace Impl
+
 /// \class Cuda
 /// \brief Kokkos Execution Space that uses CUDA to run on GPUs.
 ///
--- a/core/src/Kokkos_CudaSpace.hpp
+++ b/core/src/Kokkos_CudaSpace.hpp
@@ -50,6 +50,9 @@ namespace Kokkos {
 namespace Impl {
 
 template <typename T>
+struct is_cuda_type_ex_space : public std::false_type {};
+
+template <typename T>
 struct is_cuda_type_space : public std::false_type {};
 
 }  // namespace Impl
@@ -120,6 +123,9 @@ class CudaSpace {
 
 template <>
 struct Impl::is_cuda_type_space<CudaSpace> : public std::true_type {};
+
+template <>
+struct Impl::is_cuda_type_ex_space<Cuda> : public std::true_type {};
 
 }  // namespace Kokkos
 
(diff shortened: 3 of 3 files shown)


core/src/Kokkos_Cuda.hpp
```diff
@@ -109,6 +109,7 @@ struct CudaDispatchProperties {
 };
 }  // namespace Experimental
 }  // namespace Impl
+
```
nmm0, line 112: We can probably remove this change from the diff

nmm0 (commented):
This looks good but since I helped on it I would like others to approve

core/src/Kokkos_CudaSpace.hpp
```diff
@@ -77,6 +77,9 @@ extern "C" void kokkos_impl_cuda_set_pin_uvm_to_host(bool);
 namespace Kokkos {
 namespace Impl {
 
+template <typename T>
+struct is_cuda_type_ex_space : public std::false_type {};
```
dalg24, line 81: Just observing `is_cuda_type_space` name was not a great name in the first place but adding `is_cuda_type_ex_space` next to it make things even worse.  We can get back to fit later after we have clarified other comments.

Also you must comment about the intent here, that we grant permission to Kokkos Resilience to specialize that trait.

dalg24 (changes requested):
You need to fix the format.  See https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/Kokkos/detail/Kokkos/11264/pipeline

What would be the plan for other device backends?
Are they other places in Kokkos besides `deep_copy` that have similar issues?

core/src/Kokkos_CudaSpace.hpp
dalg24: I was referring to HIP or SYCL

nmm0: Yeah, HIP and SYCL have the same overly-constrained specialization:

https://github.com/kokkos/kokkos/blob/5dbe8eaeaaa93425c63012db26a1c8b7ba40c8bf/core/src/HIP/Kokkos_HIP_DeepCopy.hpp#L89-L93

https://github.com/kokkos/kokkos/blob/5dbe8eaeaaa93425c63012db26a1c8b7ba40c8bf/core/src/HIP/Kokkos_HIP_DeepCopy.hpp#L114-L117

https://github.com/kokkos/kokkos/blob/5dbe8eaeaaa93425c63012db26a1c8b7ba40c8bf/core/src/HIP/Kokkos_HIP_DeepCopy.hpp#L137-L140

https://github.com/kokkos/kokkos/blob/5dbe8eaeaaa93425c63012db26a1c8b7ba40c8bf/core/src/SYCL/Kokkos_SYCL_DeepCopy.hpp#L114-L120

https://github.com/kokkos/kokkos/blob/5dbe8eaeaaa93425c63012db26a1c8b7ba40c8bf/core/src/SYCL/Kokkos_SYCL_DeepCopy.hpp#L141-L146

https://github.com/kokkos/kokkos/blob/5dbe8eaeaaa93425c63012db26a1c8b7ba40c8bf/core/src/SYCL/Kokkos_SYCL_DeepCopy.hpp#L166-L171

dalg24: Then did you consider introducing a single trait that we can reuse
```C++
namespace Kokkos::Impl {
// Granting Kokkos Resilience the permission to specialize that trait to control the behavior of deep_copy
template <class ExecutionSpace>
struct UnderlyingExecutionSpace { // hopefully you can come up with a better name
  static_assert(is_execution_space_v<ExecutionSpace>);
  using type = ExecutionSpace;
};
}

Comment (nmm0) [reply core/src/Kokkos_CudaSpace.hpp:82]: That makes sense -- should the PR also include the specializations for HIP and SYCL or just the single extensible trait?

**Comment by ajpowelsnl** [comment]
Hi @ElisabethGiem -- did you see Christian's Dec. 5, 2022 comment:

```
I definitely don't want to start intoducing more is_cuda_type stuff etc. as Damien said the original name was poor. Also we need to figure out if there are other places where you want to do stuff like this, as Damien said, in order to decide on best way forward.
```

Next question:  can you rebase your changes on Kokkos `develop`?  It should be as simple as something along these lines (I can walk you through the steps):

```
git fetch upstream
git pull
git rebase develop
```

I think a backend - agnostic solution was requested.

File: core/src/Kokkos_CudaSpace.hpp [hunk 3]
```diff
@@ -120,7 +120,6 @@ class CudaSpace {
 
 template <>
 struct Impl::is_cuda_type_space<CudaSpace> : public std::true_type {};
-
```
dalg24, line 123: Undo

core/src/Kokkos_Concepts.hpp
```diff
@@ -447,6 +447,18 @@ struct SpaceAccessibility {
...
 }  // namespace Kokkos
 
+namespace Kokkos {
+namespace Impl {
+
+//! Grant Kokkos Resilience permission to specialize this trait to control
+//! the behavior of deep_copy
+template <class ExecutionSpace, class Enabled = void>
+struct ExecutionSpaceAlias {
+  using type = ExecutionSpace;
```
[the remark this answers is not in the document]
dalg24: Seriously?!  Do these occur in Kokkos?  How bad is it?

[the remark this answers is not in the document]
dalg24: > some of the deep copies are called with non-execution spaces as template parameters

I don't think that is accurate.  I think it is some header include order issue or whatnot, something like only having a forward declaration of the execution space at the point the assertion happen.

I would still prefer us handling it, whether it means placing the traits definition elsewhere or figuring out what's wrong with our spagetthi code.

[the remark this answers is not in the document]
nmm0: Ok so, I investigated further. I think I was wrong about non-execution spaces being instantiated. I get this error with cuda 11.4:

```
kokkos/core/src/Kokkos_Concepts.hpp(457): error: static assertion failed
          detected during instantiation of class "Kokkos::Impl::ExecutionSpaceAlias<ExecutionSpace, Enabled> [with ExecutionSpace=Kokkos::HostSpace::execution_space, Enabled=void]"
```

However, with clang cuda everything compiles fine.

ajpowelsnl:
@ElisabethGiem , @nmm0 -- what is the status of this issue?

nmm0:
> @ElisabethGiem , @nmm0 -- what is the status of this issue?

@ajpowelsnl We're going to circle back to it once Liz's implementation in Kokkos-resilience is done. That way we can make sure everything is sufficient for our needs.

ajpowelsnl:
> > @ElisabethGiem , @nmm0 -- what is the status of this issue?
> 
> @ajpowelsnl We're going to circle back to it once Liz's implementation in Kokkos-resilience is done. That way we can make sure everything is sufficient for our needs.

@ElisabethGiem , @nmm0 - what else needs to be done to bring this item in for a smooth landing?

nmm0:
@ajpowelsnl This can go on the backburner for now. We will return to CUDA resilience work soon, so lets keep it open

masterleinad:
@nmm0 What's the status here?

nmm0:
> @nmm0 What's the status here?

@masterleinad It's on hold until Liz switches back to the Resilient Cuda work. Right now we're focusing on OpenMP.

crtrott:
Hey I gonna close this for now, trying to do a little bit of cleanup. You can reopen if and when you get back to it.

Issue #5655, opened 2022-12-01
Make `ExecutionSpace::concurrency()` a non-static member function


dalg24:
`ExecutionSpace::concurrency()` is currently declared as a static member function in all backends and this is an issue because the concurrency is really a property of an execution space instance, not from its type.

This is something we knew we would have to deal with for a while but we can't procrastinate much longer.
In particular, it is problematic as we add support for execution space instances to parallel host backends.
Apropos, #5105 attempts to work around that issue by providing `OpenMP::concurrency(OpenMP const& = OpenMP()) -> int`

The change from static to non-static member function is a breaking change but the fix on the user-side is rather trivial
```C++
int concurrency = Kokkos::DefaultExecutionSpace::concurrency();  // does not compile any more
// change to
int concurrency = Kokkos::DefaultExecutionSpace().concurrency();  // using default-constructed execution space
// or
int concurrency = execution_space.concurrency();  // calling member function on an instance
```

The intent of this issue is to announce our intent to make this change and to formulate a plan for the transition.

PhilMiller:
4.0 would be a really nice timeframe to make such a change. We didn't deprecate to provide a warning or anything, but I think users would survive without too much grumbling if we did it ASAP

dalg24:
Unfortunately I can't think of any way to emit warnings because we would not be able to "overload" on the static member function.

PR #5598, opened 2022-10-24
Refactor Policy constructor tests

tcclevenger:
Begin the execution policy unit test refactor by having all `RangePolicy` constructor tests live in a single TestRangePolicyConstructors.hpp` file.

See https://github.com/kokkos/kokkos/issues/5560 for details on the execution policy unit test refactor.

Edit: This changed in scope based on comments, now refactoring all Policy constructor tests.

PR #5656, opened 2022-12-01
Fixup call `ExecutionSpace::concurrency()` from an actual execution space

dalg24:
Partially addressing #5655 by ensuring we call `ExecutionSpace::concurrency()` from an actual execution space in code that is not backend-specific.

Also adding the detection of the member function to static checks for all backends.  Note that I elected to only check that `concurrency()` is callable and that, even though that is something that we would be able to check later after we update all backends, I don't actually plan on enforcing that backends do not declare the member function to be static.

### DIFF
--- a/core/perf_test/PerfTest_CustomReduction.cpp
+++ b/core/perf_test/PerfTest_CustomReduction.cpp
@@ -58,8 +58,8 @@ void custom_reduction_test(int N, int R, int num_trials) {
   Scalar max;
 
   int team_size = 32;
-  if (team_size > Kokkos::DefaultExecutionSpace::concurrency())
-    team_size = Kokkos::DefaultExecutionSpace::concurrency();
+  if (team_size > Kokkos::DefaultExecutionSpace().concurrency())
+    team_size = Kokkos::DefaultExecutionSpace().concurrency();
   // Warm up
   Kokkos::parallel_reduce(
       Kokkos::TeamPolicy<>(N / 1024, team_size),
--- a/core/src/Kokkos_ExecPolicy.hpp
+++ b/core/src/Kokkos_ExecPolicy.hpp
@@ -223,8 +223,7 @@ class RangePolicy : public Impl::PolicyTraits<Properties...> {
       return;
     }
 #endif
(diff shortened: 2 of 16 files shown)


core/perf_test/PerfTest_CustomReduction.cpp
```diff
@@ -58,8 +58,8 @@ void custom_reduction_test(int N, int R, int num_trials) {
   Scalar max;
 
   int team_size = 32;
-  if (team_size > Kokkos::DefaultExecutionSpace::concurrency())
-    team_size = Kokkos::DefaultExecutionSpace::concurrency();
+  if (team_size > Kokkos::DefaultExecutionSpace().concurrency())
+    team_size = Kokkos::DefaultExecutionSpace().concurrency();
```
PhilMiller, line 62: This raises a question: how confident are we that we can generally promise that constructing a default instance of an execution space is and will remain 'cheap'? It's something that happens in a lot of places, but it's never been clear to me if that's actually a commitment Kokkos makes.

masterleinad: I mean that's the reason why we are using a static (`Impl`) execution space object for a default-constructed execution space instance.

dalg24:
Failure is unrelated (`KokkosCore_UnitTest_CudaTimingBased` in the NVHPC build)
```
7: Time CudaSpace: 0.006675 CudaUVMSpace_1: 0.011526 CudaUVMSpace_2: 0.006572 CudaPinnedHostSpace: 0.269566 CudaUVMSpace_Pinned: 0.019738
7: /var/jenkins/workspace/Kokkos/core/unit_test/cuda/TestCuda_DebugPinUVMSpace.cpp:128: Failure
7: Value of: passed
7:   Actual: false
7: Expected: true

dalg24:
I forgot to mention this PR introduces conflicts with #5598
They are no big deal but I wanted to acknowledge I was aware of these.  @tcclevenger

ldh4 (approved):
Looks fine to me.

crtrott (approved):

PR #5666, opened 2022-12-07
Try reproducing Thrust issue

masterleinad:


masterleinad:
This successfully reproduces the problem and shows that it's fixed in `develop`.

PR #5669, opened 2022-12-07
Fixup make `{Cuda,HIP}::concurrency()` member function non static

dalg24:
Partial fix for #5655

### DIFF
--- a/core/src/Cuda/Kokkos_Cuda_Locks.cpp
+++ b/core/src/Cuda/Kokkos_Cuda_Locks.cpp
@@ -59,7 +59,7 @@ void initialize_host_cuda_lock_arrays() {
                  sizeof(int) * (CUDA_SPACE_ATOMIC_MASK + 1)));
   Impl::cuda_device_synchronize(
       "Kokkos::Impl::initialize_host_cuda_lock_arrays: Pre Init Lock Arrays");
-  g_host_cuda_lock_arrays.n = Cuda::concurrency();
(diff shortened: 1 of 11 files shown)


Rombur (approved):

core/src/Cuda/Kokkos_Cuda_Locks.cpp
```diff
@@ -59,7 +59,7 @@ void initialize_host_cuda_lock_arrays() {
                  sizeof(int) * (CUDA_SPACE_ATOMIC_MASK + 1)));
   Impl::cuda_device_synchronize(
       "Kokkos::Impl::initialize_host_cuda_lock_arrays: Pre Init Lock Arrays");
-  g_host_cuda_lock_arrays.n = Cuda::concurrency();
+  g_host_cuda_lock_arrays.n = CudaInternal::concurrency();
```
ldh4, line 62: It looks like there are mixed usages between `CudaInternal::concurrency()` and `Cuda().concurrency()`. Is there a reason to prefer `CudaInternal::` over `Cuda()...` here?

ldh4 (approved):

core/src/Cuda/Kokkos_Cuda_Locks.cpp
dalg24: Yes.  Chicken or the egg dilemma:  `initialize_host_cuda_lock_arrays` is called as you initialize the backend so we really can't construct an execution space here

masterleinad (approved):

PR #5670, opened 2022-12-08
#5667: dont install std algorithm headers multiple times

### LINKED ISSUE #5667: Trilinos builds: Kokkos: std algorithms installed twice
**Describe the bug**
This is a copy of issue https://github.com/trilinos/Trilinos/issues/11358 opened by @maartenarnst 

> The headers for the std algorithms are installed twice. Once in the main trilinos include dir, and then once more in a subdir std_algorithms. The issue might have something to do with how the install is configured (when kokkos is part of trilinos, and in particular with regards to the std_algorithms subdir) in packages/kokkos/algorithms/src/CMakeLists.txt.

> This is a sample of output from the installation:
```
 -- Installing: /opt/Trilinos/GNU-Cuda/include/std_algorithms/Kokkos_BeginEnd.hpp
...
 -- Installing: /opt/Trilinos/GNU-Cuda/include/Kokkos_BeginEnd.hpp
```

See https://github.com/trilinos/Trilinos/issues/11358

nmm0:
Note that this will install them under `<install_dir>/include/std_algorithms`. Any code that depends on them being in `<install_dir>/include` will break.

Fixes #5667


cmake/kokkos_tribits.cmake
```diff
@@ -534,7 +534,7 @@ ELSE()
   CMAKE_PARSE_ARGUMENTS(PARSE
     ""
     ""
-    "HEADERS;SOURCES"
+    "HEADERS;SOURCES;NOINSTALLHEADERS"
```
dalg24, line 537: I don't think that change is needed.  Please undo.

dalg24 (commented):
Thanks for looking into this

fnrizzi:
i am surprised this was never reported in the past ! thanks Nick

fnrizzi (approved):

cmake/kokkos_tribits.cmake
masterleinad: Are the parsed arguments used at all?

nmm0: It is, sadly required because I need to pass it in to tribits, but our function will throw a cmake error on an unrecognized argument in non-tribits mode :shrug:

The argument is used in non-tribits mode. This is yet another thing we will have to strike when we don't ahve to support tribits.

nmm0: Maybe not with multi value keyword args? I get:

```
CMake Error at cmake/kokkos_tribits.cmake:18 (if):
  if given arguments:

    "NOINSTALLHEADERS" <all the headers here>

  Unknown arguments specified
  cmake/kokkos_tribits.cmake:127 (VERIFY_EMPTY)
  cmake/kokkos_tribits.cmake:542 (KOKKOS_INTERNAL_ADD_LIBRARY_INSTALL)
  algorithms/src/CMakeLists.txt:26 (KOKKOS_ADD_INTERFACE_LIBRARY)
```

dalg24: I suppose that is fine.  Thanks for getting to the bottom of it

dalg24 (approved):

cmake/kokkos_tribits.cmake
masterleinad: I still fail to see where the parsed arguments are used below. It seems that only `NAME` is relevant.

nmm0: Oh I see, yes I think we could actually remove the parse arguments. I'm not 100% sure why it is there.

nmm0:
retest this please

cmake/kokkos_tribits.cmake
dalg24: I'd prefer if we removed it

masterleinad (approved):

cmake/kokkos_tribits.cmake
nmm0: I removed it entirely

nmm0:
retest this please

dalg24:
There was no need to re-run.  I would have ignored the Cuda timing based failure :/

nmm0:
> There was no need to re-run. I would have ignored the Cuda timing based failure :/

:_:

ndellingwood:
Cherry-pick to 3.7.02 submitted with #5711

PR #5672, opened 2022-12-08
Fix default memory order for sycl::atomic_ref

masterleinad:
#5105 broke my SYCL builds uncovering invalid defaults for `sycl::atomic_ref`, see https://github.com/intel/llvm/blob/894ce25635c5ead4d1c42e70f66d1970c86a7bf5/sycl/include/sycl/atomic_ref.hpp#L125-L128=:
```
static_assert(
      detail::IsValidDefaultOrder<DefaultOrder>::value,
      "Invalid default memory_order for atomics.  Valid defaults are: "
      "relaxed, acq_rel, seq_cst");
```

### DIFF
--- a/tpls/desul/include/desul/atomics/Adapt_SYCL.hpp
+++ b/tpls/desul/include/desul/atomics/Adapt_SYCL.hpp
@@ -38,6 +38,11 @@ using sycl_memory_scope = std::conditional_t<extended_namespace,
                                              sycl::memory_scope>;
 
 //<editor-fold desc="SYCL memory order">
+// The default memory order for sycl::atomic_ref
+// can be seq_cst, acq_rel, or relaxed according to the
+// "SYCL 2020 Specification (revision 6)", see
+// https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:atomic-references.
+// Thus, we map MemoryOrderAcquire and MemoryOrderRelease to acq_rel.
 template <class MemoryOrder, bool extended_namespace = true>
 struct SYCLMemoryOrder;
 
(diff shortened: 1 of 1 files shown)


masterleinad:
Corresponds to https://github.com/desul/desul/pull/79.

dalg24:
> #5105 broke my SYCL builds uncovering invalid defaults for `sycl::atomic_ref`

I see how `acquire` and `release` are not allowed as template parameters but not how that "broke" your build.
Where did you see the error?

masterleinad:
https://gitlab.jlse.anl.gov/intel-dga/simulation/kokkos/-/jobs/153552 says something like
```
[...]]/atomic_ref.hpp:125:3: error: static assertion failed due to requirement 'detail::integral_constant<bool, false>::value': Invalid default memory_order for atomics.  Valid defaults are: relaxed, acq_rel, seq_cst
  static_assert(
  ^
[...]/atomic_ref.hpp:269:14: note: in instantiation of template class 'sycl::detail::atomic_ref_base<unsigned int, sycl::memory_order::acquire, sycl::memory_scope::device, sycl::access::address_space::generic_space>' requested here
    : public atomic_ref_base<T, DefaultOrder, DefaultScope, AddressSpace> {
             ^
[...]/atomic_ref.hpp:660:27: note: in instantiation of template class 'sycl::detail::atomic_ref_impl<unsigned int, sycl::memory_order::acquire, sycl::memory_scope::device, sycl::access::address_space::generic_space>' requested here
class atomic_ref : public detail::atomic_ref_impl<T, DefaultOrder, DefaultScope,
                          ^
[...]/tpls/desul/include/desul/atomics/Compare_Exchange_SYCL.hpp:32:59: note: in instantiation of template class 'sycl::atomic_ref<unsigned int, sycl::memory_order::acquire, sycl::memory_scope::device, sycl::access::address_space::generic_space>' requested here
  sycl_atomic_ref<unsigned int, MemoryOrder, MemoryScope> dest_ref(
                                                          ^
[...]/tpls/desul/include/desul/atomics/Generic.hpp:37:20: note: in instantiation of function template specialization 'desul::Impl::device_atomic_compare_exchange<int, desul::MemoryOrderAcquire, desul::MemoryScopeDevice>' requested here
      return Impl::device_atomic_compare_exchange(dest, cmp, val, order, scope);)
                   ^
[...]/kokkos/sycl-release-develop/core/src/OpenMP/Kokkos_OpenMP_Instance.cpp:38:22: note: in instantiation of function template specialization 'desul::atomic_compare_exchange<int, desul::MemoryOrderAcquire, desul::MemoryScopeDevice>' requested here
  while (1 == desul::atomic_compare_exchange(&m_pool_mutex, 0, 1,
```

dalg24:
I don't quite understand how that got triggered because the change (hand rolled mutex lock acquire/release in OpenMP backend) were on the host side.
(I would like to understand why we see the error in your nightly build, I don't have a particular issue with the changes you suggest)

crtrott:
Somehow it triggers the device side atomic instantiation too ...

masterleinad:
AFAICT all the code gets processed twice (once with `__SYCL_DEVICE_ONLY__` and one time without it) and there is no guard for `OpenMP` atomics to restrict it to the host part.

crtrott:
Yeah but the thing is that the call to device_atomic_compare_exchange should only happen if the atomic is called on the device, since that call IS protected. But I guess the "throw the device stuff away" only happens at a later point. We should test that.

PR #5682, opened 2022-12-13
Make SYCL::concurrency non-static

masterleinad:
Corresponding to #5669. It seems that we don't need `g_host_cuda_lock_arrays` (probably even after merging the generic atomics pull request). Hence, I haven't included a static `SYCLInternal::concurrency` function.

### DIFF
--- a/core/src/SYCL/Kokkos_SYCL.cpp
+++ b/core/src/SYCL/Kokkos_SYCL.cpp
@@ -65,9 +65,15 @@ SYCL::SYCL(const sycl::queue& stream)
   m_space_instance->initialize(stream);
 }
 
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
 int SYCL::concurrency() {
   return Impl::SYCLInternal::singleton().m_maxConcurrency;
 }
+#else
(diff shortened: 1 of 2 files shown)


core/src/SYCL/Kokkos_SYCL.cpp
```diff
@@ -65,9 +65,15 @@ SYCL::SYCL(const sycl::queue& stream)
   m_space_instance->initialize(stream);
 }
 
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
 int SYCL::concurrency() {
   return Impl::SYCLInternal::singleton().m_maxConcurrency;
 }
+#else
+int SYCL::concurrency() const {
+  return Impl::SYCLInternal::singleton().m_maxConcurrency;
```
crtrott, line 74: why not return `m_space_instance->m_maxConcurrency`?

crtrott (changes requested):

crtrott (approved):

ldh4 (approved):

dalg24:
SYCL build passed.  Ignoring unrelated failures.

PR #5676, opened 2022-12-10
ENH: drop deprecated sprintf usage in Kokkos_Profiling.cpp

neutrinoceros:
fixes a deprecation warning with the latest version of Xcode (14.1)

```
 warning: 'sprintf' is deprecated: This function is provided for compatibility reasons only.  Due to security concerns inherent in the design of sprintf(3), it is highly recommended that you use snprintf(3) instead. [-Wdeprecated-declarations]
  sprintf(envProfileCopy.get(), "%s", envProfileLibrary);
  ^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.0.sdk/usr/include/stdio.h:188:1: note: 'sprintf' has been explicitly marked deprecated here
__deprecated_msg("This function is provided for compatibility reasons only.  Due to security concerns inherent in the design of sprintf(3), it is highly recommended that you use snprintf(3) instead.")
^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX13.0.sdk/usr/include/sys/cdefs.h:215:48: note: expanded from macro '__deprecated_msg'
        #define __deprecated_msg(_msg) __attribute__((__deprecated__(_msg)))
```

I just fixed the one call that is hit in my application; I am not attempting to perform this migration over the whole code base, but I'm willing to do it if considered helpful.

PR #5683, opened 2022-12-13
use snprintf to silence Apple Clang

ibaned:
the version of Apple Clang that I'm
using by default warns that sprintf
is deprecated for security reasons,
so this change switches to snprintf
to silence the warning.


masterleinad:
Duplicate of #5676.

ibaned:
Wow Kokkos has a healthy community if we're getting duplicate fixes! Thanks!

PR #5687, opened 2022-12-15
Fix #4942 Scratch Allocation Alignment

### LINKED ISSUE #4942: Cuda backend fails to allocate a single integer in the PerTeam scratch space.
```c++
#include <Kokkos_Core.hpp>

using TeamPolicy = Kokkos::TeamPolicy<ExecSpace>;
using member_type = TeamPolicy::member_type;
using ScratchSpace = ExecSpace::scratch_memory_space;
using scratch_view_type =
    Kokkos::View<int*, ScratchSpace, Kokkos::MemoryTraits<Kokkos::Unmanaged>>;

int main(int argc, char** argv) {
    {
        TeamPolicy policy(1, 1);
        size_t scratch_size =  sizeof(int);
//        size_t scratch_size = scratch_view_type::shmem_size(1);

        Kokkos::parallel_for(
            policy.set_scratch_size(0, Kokkos::PerTeam(scratch_size)),
            KOKKOS_LAMBDA(const member_type& team) {
                scratch_view_type scratch_view(team.team_scratch(0),
                                               team.team_size());

                if (team.team_rank() == 0) {
                    scratch_view(0) = 1;
                }
            });
        Kokkos::fence();

    }
    Kokkos::finalize();

    return 0;
}
```

This fails with CUDA backend at runtime with illegal memory access errors. 
I need to either allocate 2 integers or use the `shmem_size` routine. 

I used cuda/11.2.1.

crtrott:
This fixes the scratch space allocation alignment, #4942.
Note there is a default over alignment used by views. This makes sure that in the mode in which we run with 64bit shared memory banks that you don't have two conflicting scratch views resulting in really bad bank conflicts.

That default alignment was always there, and our scratch allocation mechanism was now doing a weird (faulty) mix of default alignment and requested alignment. I cleaned that up in a way that `get_shmem` doesn't do any alignment, `get_aligned_shmem` uses only the provided alignment in the argument, and `Kokkos::View` takes the default alignment into account for its scratch allocations.

### DIFF
--- a/core/src/Kokkos_ScratchSpace.hpp
+++ b/core/src/Kokkos_ScratchSpace.hpp
@@ -41,9 +41,8 @@ class ScratchMemorySpace {
       "Instantiating ScratchMemorySpace on non-execution-space type.");
 
  public:
-  // Alignment of memory chunks returned by 'get'
-  // must be a power of two
-  enum { ALIGN = 8 };
+  // Minimal overalignment used by view scratch allocations
+  constexpr static int ALIGN = 8;
 
  private:
   mutable char* m_iter_L0 = nullptr;
@@ -55,7 +54,7 @@ class ScratchMemorySpace {
   mutable int m_offset        = 0;
   mutable int m_default_level = 0;
 
-  enum { MASK = ALIGN - 1 };  // Alignment used by View::shmem_size
+  constexpr static int DEFAULT_ALIGNMENT_MASK = ALIGN - 1;
 
  public:
   //! Tag this class as a memory space
@@ -69,10 +68,13 @@ class ScratchMemorySpace {
 
   static constexpr const char* name() { return "ScratchMemorySpace"; }
 
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  // This function is unused
   template <typename IntType>
-  KOKKOS_INLINE_FUNCTION static IntType align(const IntType& size) {
-    return (size + MASK) & ~MASK;
+  KOKKOS_INLINE_FUNCTION static constexpr IntType align(const IntType& size) {
+    return (size + DEFAULT_ALIGNMENT_MASK) & ~DEFAULT_ALIGNMENT_MASK;
   }
+#endif
 
   template <typename IntType>
   KOKKOS_INLINE_FUNCTION void* get_shmem(const IntType& size,
@@ -93,15 +95,29 @@ class ScratchMemorySpace {
                                                 const ptrdiff_t alignment,
                                                 int level = -1) const {
     if (level == -1) level = m_default_level;
-    auto& m_iter              = (level == 0) ? m_iter_L0 : m_iter_L1;
-    auto& m_end               = (level == 0) ? m_end_L0 : m_end_L1;
-    char* previous            = m_iter;
-    const ptrdiff_t missalign = size_t(m_iter) % alignment;
-    if (missalign) m_iter += alignment - missalign;
+    auto& m_iter = (level == 0) ? m_iter_L0 : m_iter_L1;
+    auto& m_end  = (level == 0) ? m_end_L0 : m_end_L1;
+    if constexpr (aligned) {
+      const ptrdiff_t missalign = size_t(m_iter) % alignment;
+      if (missalign) m_iter += alignment - missalign;
+    }
 
-    void* tmp = m_iter + m_offset * (aligned ? size : align(size));
-    if (m_end < (m_iter += (aligned ? size : align(size)) * m_multiplier)) {
-      m_iter = previous;  // put it back like it was
+    // This is each threads start pointer for their allocation
+    // Note: for team scratch m_offset is 0, since every
+    // thread will get back the same shared pointer
+    void* tmp           = m_iter + m_offset * size;
+    ptrdiff_t increment = size * m_multiplier;
+
+    // increment m_iter first and decrement it again if not
+    // enough memory was available. In the non-failing path
+    // this will safe instructions.
+    m_iter += increment;
+
+    if (m_end < m_iter) {
+      // Request did overflow: reset the base team ptr, and
+      // return nullptr
+      m_iter -= increment;
+      tmp = nullptr;
 #ifdef KOKKOS_ENABLE_DEBUG
       // mfh 23 Jun 2015: printf call consumes 25 registers
       // in a CUDA build, so only print in debug mode.  The
@@ -111,7 +127,6 @@ class ScratchMemorySpace {
           "%ld byte(s); remaining capacity is %ld byte(s)\n",
           long(size), long(m_end - m_iter));
 #endif  // KOKKOS_ENABLE_DEBUG
-      tmp = nullptr;
     }
     return tmp;
   }
--- a/core/src/Kokkos_ScratchSpace.hpp
+++ b/core/src/Kokkos_ScratchSpace.hpp
@@ -99,8 +99,8 @@ class ScratchMemorySpace {
                                                 const ptrdiff_t alignment,
                                                 int level = -1) const {
     if (level == -1) level = m_default_level;
-    auto& m_iter = (level == 0) ? m_iter_L0 : m_iter_L1;
-    auto& m_end  = (level == 0) ? m_end_L0 : m_end_L1;
+    auto& m_iter    = (level == 0) ? m_iter_L0 : m_iter_L1;
+    auto& m_end     = (level == 0) ? m_end_L0 : m_end_L1;
     auto m_iter_old = m_iter;
     if constexpr (alignment_requested) {
       const ptrdiff_t missalign = size_t(m_iter) % alignment;
@@ -116,7 +116,7 @@ class ScratchMemorySpace {
     if (increment > m_end - m_iter) {
       // Request did overflow: return nullptr and reset m_iter
       m_iter = m_iter_old;
-      tmp = nullptr;
+      tmp    = nullptr;
 #ifdef KOKKOS_ENABLE_DEBUG
       // mfh 23 Jun 2015: printf call consumes 25 registers
       // in a CUDA build, so only print in debug mode.  The
--- a/core/src/Kokkos_ScratchSpace.hpp
+++ b/core/src/Kokkos_ScratchSpace.hpp
@@ -101,6 +101,7 @@ class ScratchMemorySpace {
     if (level == -1) level = m_default_level;
     auto& m_iter = (level == 0) ? m_iter_L0 : m_iter_L1;
     auto& m_end  = (level == 0) ? m_end_L0 : m_end_L1;
+    auto m_iter_old = m_iter;
     if constexpr (alignment_requested) {
       const ptrdiff_t missalign = size_t(m_iter) % alignment;
       if (missalign) m_iter += alignment - missalign;
@@ -113,7 +114,8 @@ class ScratchMemorySpace {
     ptrdiff_t increment = size * m_multiplier;
 
     if (increment > m_end - m_iter) {
-      // Request did overflow: return nullptr
+      // Request did overflow: return nullptr and reset m_iter
+      m_iter = m_iter_old;
       tmp = nullptr;
 #ifdef KOKKOS_ENABLE_DEBUG
       // mfh 23 Jun 2015: printf call consumes 25 registers
--- a/core/src/Kokkos_ScratchSpace.hpp
+++ b/core/src/Kokkos_ScratchSpace.hpp
@@ -112,12 +112,7 @@ class ScratchMemorySpace {
     void* tmp           = m_iter + m_offset * size;
     ptrdiff_t increment = size * m_multiplier;
 
-    // increment m_iter first and decrement it again if not
-    // enough memory was available. In the non-failing path
-    // this will save instructions.
-    m_iter += increment;
-
-    if (m_end < m_iter) {
+    if (increment > m_end - m_iter) {
       // Request did overflow: reset the base team ptr, and
       // return nullptr
       m_iter -= increment;
@@ -131,6 +126,8 @@ class ScratchMemorySpace {
           "%ld byte(s); remaining capacity is %ld byte(s)\n",
           long(size), long(m_end - m_iter));
 #endif  // KOKKOS_ENABLE_DEBUG
+    } else {
+      m_iter += increment;
     }
     return tmp;
   }
--- a/core/src/Kokkos_ScratchSpace.hpp
+++ b/core/src/Kokkos_ScratchSpace.hpp
@@ -113,9 +113,7 @@ class ScratchMemorySpace {
     ptrdiff_t increment = size * m_multiplier;
 
     if (increment > m_end - m_iter) {
-      // Request did overflow: reset the base team ptr, and
-      // return nullptr
-      m_iter -= increment;
+      // Request did overflow: return nullptr
       tmp = nullptr;
 #ifdef KOKKOS_ENABLE_DEBUG
       // mfh 23 Jun 2015: printf call consumes 25 registers
--- a/core/src/Kokkos_ScratchSpace.hpp
+++ b/core/src/Kokkos_ScratchSpace.hpp
@@ -54,7 +54,9 @@ class ScratchMemorySpace {
   mutable int m_offset        = 0;
   mutable int m_default_level = 0;
 
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
   constexpr static int DEFAULT_ALIGNMENT_MASK = ALIGN - 1;
+#endif
 
  public:
   //! Tag this class as a memory space
@@ -71,7 +73,8 @@ class ScratchMemorySpace {
 #ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
   // This function is unused
   template <typename IntType>
-  KOKKOS_INLINE_FUNCTION static constexpr IntType align(const IntType& size) {
+  KOKKOS_DEPRECATED KOKKOS_INLINE_FUNCTION static constexpr IntType align(
+      const IntType& size) {
     return (size + DEFAULT_ALIGNMENT_MASK) & ~DEFAULT_ALIGNMENT_MASK;
   }
 #endif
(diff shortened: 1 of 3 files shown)


core/src/Kokkos_ScratchSpace.hpp
```diff
@@ -93,15 +95,29 @@ class ScratchMemorySpace {
...
-    if (m_end < (m_iter += (aligned ? size : align(size)) * m_multiplier)) {
-      m_iter = previous;  // put it back like it was
+    auto& m_iter = (level == 0) ? m_iter_L0 : m_iter_L1;
+    auto& m_end  = (level == 0) ? m_end_L0 : m_end_L1;
+    if constexpr (aligned) {
+      const ptrdiff_t missalign = size_t(m_iter) % alignment;
+      if (missalign) m_iter += alignment - missalign;
+    }
+
+    // This is each threads start pointer for their allocation
```
PhilMiller, line 105:
```suggestion
    // This is each thread's start pointer for its allocation
```

core/src/Kokkos_ScratchSpace.hpp
```diff
@@ -93,15 +95,29 @@ class ScratchMemorySpace {
...
+    // Note: for team scratch m_offset is 0, since every
+    // thread will get back the same shared pointer
+    void* tmp           = m_iter + m_offset * size;
+    ptrdiff_t increment = size * m_multiplier;
+
+    // increment m_iter first and decrement it again if not
+    // enough memory was available. In the non-failing path
+    // this will safe instructions.
```
PhilMiller, line 113:
```suggestion
    // this will save instructions.
```

core/src/Kokkos_View.hpp
```diff
@@ -1638,7 +1639,11 @@ class View : public ViewTraits<DataType, Properties...> {
   static KOKKOS_INLINE_FUNCTION size_t
   shmem_size(typename traits::array_layout const& arg_layout) {
     return map_type::memory_span(arg_layout) +
-           sizeof(typename traits::value_type);
+           // Want to be able to align to minmum alignment or size of elements
+           ::Kokkos::max(
+               sizeof(typename traits::value_type),
+               static_cast<size_t>(
+                   traits::execution_space::scratch_memory_space::ALIGN));
```
PhilMiller, line 1646: Shouldn't this also compare `alignof` in case that's larger?

PhilMiller:
I edited the description slightly. Please check that my change was accurate to your meaning

core/src/Kokkos_ScratchSpace.hpp
```diff
@@ -55,7 +54,7 @@ class ScratchMemorySpace {
   mutable int m_offset        = 0;
   mutable int m_default_level = 0;
 
-  enum { MASK = ALIGN - 1 };  // Alignment used by View::shmem_size
+  constexpr static int DEFAULT_ALIGNMENT_MASK = ALIGN - 1;
```
masterleinad, line 57: This one is also only used in `align()` and can thus be deprecated as well.
```suggestion
#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
  KOKKOS_DEPRECATED constexpr static int DEFAULT_ALIGNMENT_MASK = ALIGN - 1;
#endif
```

core/src/Kokkos_ScratchSpace.hpp
```diff
@@ -69,10 +68,13 @@ class ScratchMemorySpace {
 
   static constexpr const char* name() { return "ScratchMemorySpace"; }
 
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  // This function is unused
   template <typename IntType>
-  KOKKOS_INLINE_FUNCTION static IntType align(const IntType& size) {
-    return (size + MASK) & ~MASK;
+  KOKKOS_INLINE_FUNCTION static constexpr IntType align(const IntType& size) {
```
masterleinad, line 74:
```suggestion
  KOKKOS_DEPRECATED KOKKOS_INLINE_FUNCTION static constexpr IntType align(const IntType& size) {
```

core/src/Kokkos_View.hpp
```diff
@@ -1638,7 +1639,11 @@ class View : public ViewTraits<DataType, Properties...> {
   static KOKKOS_INLINE_FUNCTION size_t
   shmem_size(typename traits::array_layout const& arg_layout) {
     return map_type::memory_span(arg_layout) +
-           sizeof(typename traits::value_type);
+           // Want to be able to align to minmum alignment or size of elements
```
masterleinad, line 1642:
```suggestion
           // Want to be able to align to minimum alignment or size of elements
```

core/src/Kokkos_View.hpp
```diff
@@ -1647,7 +1652,11 @@ class View : public ViewTraits<DataType, Properties...> {
       : View(Impl::ViewCtorProp<pointer_type>(
                  reinterpret_cast<pointer_type>(arg_space.get_shmem_aligned(
                      map_type::memory_span(arg_layout),
-                     sizeof(typename traits::value_type)))),
+                     // Want to align to minmum alignment or size of elements
```
masterleinad, line 1655:
```suggestion
                     // Want to align to minimum alignment or size of elements
```

core/src/Kokkos_View.hpp
```diff
@@ -1665,7 +1674,11 @@ class View : public ViewTraits<DataType, Properties...> {
                      map_type::memory_span(typename traits::array_layout(
                          arg_N0, arg_N1, arg_N2, arg_N3, arg_N4, arg_N5, arg_N6,
                          arg_N7)),
-                     sizeof(typename traits::value_type)))),
+                     // Want to align to minmum alignment or size of elements
```
masterleinad, line 1677:
```suggestion
                     // Want to align to minimum alignment or size of elements
```

core/src/Kokkos_ScratchSpace.hpp
crtrott: its a private variable so I don't think we need to add deprecation?

masterleinad: Sure, then just guard it so we don't forget to remove it later.

PhilMiller (approved):

PhilMiller:
HPX failure is salient - it's in scratch allocation.

core/unit_test/TestTeam.hpp
```diff
@@ -1553,12 +1555,68 @@ struct TestScratchAlignment {
...
+        KOKKOS_LAMBDA(const member_type &team) {
+          int *scratch_ptr1 = (int *)team.team_shmem().get_shmem(24);
+          int *scratch_ptr2 = (int *)team.team_shmem().get_shmem(32);
+          int *scratch_ptr3 = (int *)team.team_shmem().get_shmem(12);
+
+          if ((int(scratch_ptr2 - scratch_ptr1) != 6) ||
+              (int(scratch_ptr3 - scratch_ptr2) != 8))
+            flag() = 1;
+
+          if (((scratch_ptr3 - static_cast<int *>(nullptr)) + 3) % 2 == 1)
```
dalg24, line 1593: Is that not UB?  @nliber please advise

dalg24 (changes requested):

core/unit_test/TestTeam.hpp
PhilMiller: Ooh, yeah, don't do that. Do all the arithmetic and comparisons on `intptr_t`. Forming an invalid pointer is UB.

core/src/Kokkos_View.hpp
```diff
@@ -1635,19 +1636,27 @@ class View : public ViewTraits<DataType, Properties...> {
...
   }
 
+ private:
+  // Want to be able to align to minimum scratch alignment or sizeof or alignof
+  // elements
+  static constexpr size_t scratch_value_alignment =
+      ::Kokkos::max(::Kokkos::max(sizeof(typename traits::value_type),
+                                  alignof(typename traits::value_type)),
+                    static_cast<size_t>(
+                        traits::execution_space::scratch_memory_space::ALIGN));
```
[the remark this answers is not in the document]
crtrott: actually there was an ambiguity with the unqualified cmath max function without it. But I guess that won't hit the multi argument one.

core/src/Kokkos_View.hpp
```diff
@@ -1635,19 +1636,27 @@ class View : public ViewTraits<DataType, Properties...> {
...
+           alignof(typename traits::value_type),
+           static_cast<size_t>(
+               traits::execution_space::scratch_memory_space::ALIGN)});
+
+ public:
   static KOKKOS_INLINE_FUNCTION size_t
   shmem_size(typename traits::array_layout const& arg_layout) {
-    return map_type::memory_span(arg_layout) +
-           sizeof(typename traits::value_type);
+    return map_type::memory_span(arg_layout) + scratch_value_alignment;
```
PhilMiller, line 1651: I don't love that this inherently overestimates (slightly) how much shared memory will be necessary to make such an allocation request succeed, but I guess it's not a big deal.

core/src/Kokkos_ScratchSpace.hpp
```diff
@@ -69,39 +70,58 @@ class ScratchMemorySpace {
...
+    // This is each thread's start pointer for its allocation
+    // Note: for team scratch m_offset is 0, since every
+    // thread will get back the same shared pointer
+    void* tmp           = m_iter + m_offset * size;
+    ptrdiff_t increment = size * m_multiplier;
+
+    // increment m_iter first and decrement it again if not
+    // enough memory was available. In the non-failing path
+    // this will save instructions.
+    m_iter += increment;
```
PhilMiller, line 118: Very pedantically, the failing path leads to UB by calculating a pointer value beyond `m_end` (assuming that's actually the limit of the backing allocation space), but I don't think we actually need to worry about that.

For the record, the 'correct' way to do this would be to check `increment < m_end - m_iter` before the addition.

PhilMiller (approved):

core/src/Kokkos_View.hpp
crtrott: There is no way around though, because we can't predict the order in which allocations are made. And how much alignment buffer you need will depend on that.

crtrott:
Retest this please

core/unit_test/TestTeam.hpp
```diff
@@ -1553,12 +1556,67 @@ struct TestScratchAlignment {
...
+          }
+          scratch_ptr2 = intptr_t(team.team_shmem().get_shmem_aligned(32, 8));
+          scratch_ptr3 = intptr_t(team.team_shmem().get_shmem_aligned(8, 4));
+
+          if ((int(scratch_ptr2 - scratch_ptr1) != 28) &&
+              (int(scratch_ptr2 - scratch_ptr1) != 16))
+            flag() = 1;
+          if (int(scratch_ptr3 - scratch_ptr2) != 32) flag() = 1;
+          if ((int(scratch_ptr1 % 4) != 0) || (int(scratch_ptr2 % 8) != 0) ||
+              (int(scratch_ptr3 % 4) != 0))
```
dalg24, line 1607: Do you need to cast to int here?  Same question above L1602-1603

core/src/Kokkos_ScratchSpace.hpp
```diff
@@ -69,39 +70,58 @@ class ScratchMemorySpace {
...
+
+    // This is each thread's start pointer for its allocation
+    // Note: for team scratch m_offset is 0, since every
+    // thread will get back the same shared pointer
+    void* tmp           = m_iter + m_offset * size;
+    ptrdiff_t increment = size * m_multiplier;
+
+    // increment m_iter first and decrement it again if not
+    // enough memory was available. In the non-failing path
+    // this will save instructions.
```
dalg24, line 117: I am not following.  How is it saving instructions?

core/unit_test/TestTeam.hpp
crtrott: I propbably could write 4lu and 0lu instead ...

core/unit_test/TestTeam.hpp
```diff
@@ -1553,12 +1556,67 @@ struct TestScratchAlignment {
...
+    using member_type = typename Kokkos::TeamPolicy<ExecSpace>::member_type;
+    Kokkos::TeamPolicy<ExecSpace> policy(1, 1);
+    Kokkos::View<int, ExecSpace> flag("Flag");
+
+    Kokkos::parallel_for(
+        policy.set_scratch_size(0, Kokkos::PerTeam(1024)),
+        KOKKOS_LAMBDA(const member_type &team) {
+          intptr_t scratch_ptr1 = intptr_t(team.team_shmem().get_shmem(24));
+          intptr_t scratch_ptr2 = intptr_t(team.team_shmem().get_shmem(32));
+          intptr_t scratch_ptr3 = intptr_t(team.team_shmem().get_shmem(12));
```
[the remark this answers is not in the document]
crtrott: it actually turns out that that doesn't work. https://godbolt.org/z/YhdzrPsaY

core/unit_test/TestTeam.hpp
```diff
@@ -1553,12 +1556,74 @@ struct TestScratchAlignment {
         });
     Kokkos::fence();
   }
+
+  void test_minimal() {
```
PhilMiller, line 1560: This fails on OpenMPTarget, with 
```
4: [ RUN      ] openmptarget.scratch_align
4: OpenMPTarget backend requires a minimum of 32 threads per team.
```

PhilMiller (changes requested):

core/src/Kokkos_ScratchSpace.hpp
crtrott: In the case of the request being successful (which is hopefully the common case) I only one compute m_iter + increment. If I were to first use `m_end<m_iter+increment` as the condition, and then actually increment m_iter in the else branch its an instruction more.

core/src/Kokkos_ScratchSpace.hpp
```diff
@@ -69,39 +70,53 @@ class ScratchMemorySpace {
...
+    // This is each thread's start pointer for its allocation
+    // Note: for team scratch m_offset is 0, since every
+    // thread will get back the same shared pointer
+    void* tmp           = m_iter + m_offset * size;
+    ptrdiff_t increment = size * m_multiplier;
+
+    if (increment > m_end - m_iter) {
+      // Request did overflow: reset the base team ptr, and
+      // return nullptr
+      m_iter -= increment;
```
dalg24, line 118:
```suggestion
```

dalg24: do not forget to update the comment above

core/src/Kokkos_ScratchSpace.hpp
crtrott: changed

core/unit_test/TestTeam.hpp
crtrott: fixed

core/src/Kokkos_ScratchSpace.hpp
crtrott: done

core/src/Kokkos_ScratchSpace.hpp
```diff
@@ -69,39 +70,51 @@ class ScratchMemorySpace {
...
-    if (missalign) m_iter += alignment - missalign;
-
-    void* tmp = m_iter + m_offset * (aligned ? size : align(size));
-    if (m_end < (m_iter += (aligned ? size : align(size)) * m_multiplier)) {
-      m_iter = previous;  // put it back like it was
+    auto& m_iter = (level == 0) ? m_iter_L0 : m_iter_L1;
+    auto& m_end  = (level == 0) ? m_end_L0 : m_end_L1;
+    if constexpr (alignment_requested) {
+      const ptrdiff_t missalign = size_t(m_iter) % alignment;
+      if (missalign) m_iter += alignment - missalign;
```
PhilMiller, line 106: This still happens even if the allocation fails

core/src/Kokkos_ScratchSpace.hpp
```diff
@@ -69,39 +70,53 @@ class ScratchMemorySpace {
...
-
-    void* tmp = m_iter + m_offset * (aligned ? size : align(size));
-    if (m_end < (m_iter += (aligned ? size : align(size)) * m_multiplier)) {
-      m_iter = previous;  // put it back like it was
+    auto& m_iter = (level == 0) ? m_iter_L0 : m_iter_L1;
+    auto& m_end  = (level == 0) ? m_end_L0 : m_end_L1;
+    auto m_iter_old = m_iter;
+    if constexpr (alignment_requested) {
+      const ptrdiff_t missalign = size_t(m_iter) % alignment;
+      if (missalign) m_iter += alignment - missalign;
```
PhilMiller, line 107: I'm sorry to do this to you, but I just noticed, this has the same potential for overflow to evoke UB.

core/src/Kokkos_ScratchSpace.hpp
```diff
@@ -69,39 +70,53 @@ class ScratchMemorySpace {
...
+    }
+
+    // This is each thread's start pointer for its allocation
+    // Note: for team scratch m_offset is 0, since every
+    // thread will get back the same shared pointer
+    void* tmp           = m_iter + m_offset * size;
```
PhilMiller, line 113: And so does this

PhilMiller (commented):
I'll reiterate that I don't think the UB concerns are actually pressing, unless we have a reason to be especially scrupulous in this piece of the codebase.

core/src/Kokkos_ScratchSpace.hpp
crtrott: I think we should just accept this here, its starting to get pretty complicated otherwise, since we also need to check that for example m_end - alignment + missalign is not going below m_start. That is actually also a problem in the other place where we compare m_iter to m_end - size or whatever it was. Because the expression m_end - size may be UB already.

crtrott:
Also this PR does not make the situation worse with regard to UB.

dalg24 (approved):
I think this is good enough.
We can live a bit longer with UB and this is holding the 4.0 release.
I suggest we merge as in and deal with the issues pointed out by Phil in a follow up PR.

PhilMiller (approved):
I'm fine with moving this ahead, and addressing UB concerns in the pointer arithmetic later

PR #5688, opened 2022-12-15
Avoid warnings in NVHPC CI build

dalg24:
Follow-up for #5686 
* Prefer `NVHPC_CUDA_HOME` environment variable with NVHPC 22.9 to get rid of
```
nvc++-Warning-CUDA_HOME has been deprecated. Please, use NVHPC_CUDA_HOME instead.
```
* Drive by change to suppress bogus diagnostics


dalg24:
NVHPC build passed and successfully got rid of those warnings

PhilMiller:
For posterity, could you copy and paste what one of the now-suppressed bogus warnings looked like?

crtrott (approved):

PR #5620, opened 2022-11-07
Tracking performance testing: Port core performance tests

cz4rs:
Port performance tests in `core/perf_test` to use google benchmark lib and the common benchmark infrastructure.

fixes #5070

PR #5689, opened 2022-12-15
Tracking performance results: Collect performance results

### LINKED ISSUE #5070: [do not close] NGA/task 4: tracking performance testing task
# Statement 

- [x] 2.1
  - explore current setup of testing 
  - choose one Kokkos Core performance test and one Kokkos Kernels tests as design vehicle
     -  propose and justify which tests should be used 
  - figure out a plan for how we deal with googlebench dependency 
  - figure out which compilers/backends we want to run all perf tests with 
  - for more details, see #5342
- [x] 2.2 
   - Design metadata context to store system config, toolchain details etc.
  - includes for example the information from Kokkos::print configuration
  - explore integration of Kokkos tools metadata declaration
  - for more details, see #5348 
- [x] 2.3 Add metadata output to design vehicle performance tests
  - whatever comes out of `2.2`, just use `benchmark` API `benchmark::AddCustomContext(const std::string& key, const std::string& value)` to include the metadata context in the output
- [x] 2.4 After design is fully approved convert the bulk of the performance tests in the Kokkos Core repository
  - src/perf_test, containers/performance tests
- [x] 2.5 Propose a mechanism to track performance over time
  - needs to be able to produce timelines specific to system config and toolchain
  - What we already decided: 
    - to have a repo inside the kokkos organization where to store all the json files: let's call this the `database-repo` for now 
    - to push there (somehow) all the json files generated by the benchmark for example every time a PR is merged into main branch 
    - to have a script inside `database-repo` that automatically generates plots every time files are updated,
    - for more details, see #5464 
- [x] 2.6 Implement agreed upon tracking mechanism

### LINKED ISSUE #5464: Tracking performance testing - collect and analyze results of benchmark runs
Related to #5070 2.5(Propose a mechanism to track performance over time):
> - have a repo inside the kokkos organization where to store all the json files: let's call this the `database-repo` for now
> - push there (somehow) all the json files generated by the benchmark for example every time a PR is merged into main branch
> - have a script inside `database-repo` that automatically generates plots every time files are updated

--- 

1. When the Performance Test will be run on CI we can add an additional step which will upload all generated json files to `database-repo`. 
- There is a GitHub Action which can be used: https://github.com/marketplace/actions/push-a-file-to-another-repository
- This GA only needs a `API_TOKEN_GITHUB` to be set in `Secrets` section of your repository options. 
2. After pushing benchmark results to `database-repo` there will be GitHub Action run in `database-repo`. The sole purpose of that action will be to run the analyzing software that will generate the required output data.
3. To analyze the benchmark data we can use: https://github.com/bensanmorris/benchmark_monitor
- By default this tool will generate **chart** with benchmark runs and **HTML index file** based on the **Jinja2** template. This template could be modified to include benchmark context data.
- **Update**: By default the images will be saved using following format `BenchamrkName`-`metric`.`png` so for example `BM_SomeFunction-real_time.png`.
4. Generated files can be pushed using GitHub Actions to GitHub Pages. Github Action that can be used: https://github.com/marketplace/actions/deploy-to-github-pages

--- 

Example chart:
BM_SomeFunction-real_time

---

TODO:
- I will create a quick demo on my repositories

cz4rs:
fixes #5464

Add new GitHub Actions workflow to:
- run all existing benchmarks
  - this is trigerred on PRs as well
- collect benchmark results (JSON files)
- and push them to a separate (storage) repository
  - this step will only happen after a push to `develop` -> i.e. when a PR gets merged to `develop`

---
Some snippets for re-implementing this in Jenkins:
[Adding GitHub token to Jenkins](https://stackoverflow.com/questions/61105368/how-to-use-github-personal-access-token-in-jenkins)
[Jenkins Pipeline Git push](url)

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -93,6 +93,7 @@ jobs:
         run: |
           cmake -B builddir \
             -DCMAKE_INSTALL_PREFIX=/usr \
+            -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
             ${{ matrix.clang-tidy }} \
             -Ddesul_ROOT=/usr/desul-install/ \
             -DKokkos_ARCH_NATIVE=ON \
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -93,7 +93,6 @@ jobs:
         run: |
           cmake -B builddir \
             -DCMAKE_INSTALL_PREFIX=/usr \
-            -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
             ${{ matrix.clang-tidy }} \
             -Ddesul_ROOT=/usr/desul-install/ \
             -DKokkos_ARCH_NATIVE=ON \
@@ -106,6 +105,7 @@ jobs:
             -DKokkos_ENABLE_DEPRECATED_CODE_4=ON \
             -DKokkos_ENABLE_DEPRECATION_WARNINGS=OFF \
             -DCMAKE_CXX_COMPILER=${{ matrix.cxx }} \
+            -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
             -DCMAKE_BUILD_TYPE=${{ matrix.cmake_build_type }}
       - name: Build
         run: |
--- a/.github/workflows/performance-benchmark.yml
+++ b/.github/workflows/performance-benchmark.yml
@@ -16,8 +16,6 @@ jobs:
     runs-on: ubuntu-latest
     container:
       image: ghcr.io/kokkos/ci-containers/${{ matrix.distro }}
-      # see https://github.com/actions/virtual-environments/issues/3812
-      options: --security-opt seccomp=unconfined
     env:
       BUILD_ID: ${{ matrix.distro }}-${{ matrix.cxx }}-${{ matrix.backend }}
     steps:
--- a/.github/workflows/performance-benchmark.yml
+++ b/.github/workflows/performance-benchmark.yml
@@ -1,0 +1,66 @@
+name: github-benchmarks
+on:
+  push:
+    branches:
+      - develop
+      - performance-results-visualization
+
+jobs:
+  CI:
+    continue-on-error: true
+    strategy:
+      matrix:
+        distro: ['ubuntu:latest']
+        cxx: ['g++', 'clang++']
+        backend: ['OPENMP']
+    runs-on: ubuntu-latest
+    container:
+      image: ghcr.io/kokkos/ci-containers/${{ matrix.distro }}
+      # see https://github.com/actions/virtual-environments/issues/3812
+      options: --security-opt seccomp=unconfined
+    env:
+      BUILD_ID: ${{ matrix.distro }}-${{ matrix.cxx }}-${{ matrix.backend }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - uses: actions/cache@v3
+        with:
+          path: ~/.ccache
+          key: kokkos-${{ matrix.distro }}-${{ matrix.cxx }}-${{ matrix.openmp }}-${github.ref}-${{ github.sha }}
+          restore-keys: kokkos-${{ matrix.distro }}-${{ matrix.cxx }}-${{ matrix.openmp }}-${{github.ref}}
+      - name: Configure Kokkos
+        run: |
+          cmake -B builddir \
+            -DCMAKE_INSTALL_PREFIX=/usr \
+            -DKokkos_ARCH_NATIVE=ON \
+            -DKokkos_ENABLE_HWLOC=ON \
+            -DKokkos_ENABLE_${{ matrix.backend }}=ON \
+            -DKokkos_ENABLE_TESTS=ON \
+            -DKokkos_ENABLE_BENCHMARKS=ON \
+            -DKokkos_ENABLE_EXAMPLES=ON \
+            -DKokkos_ENABLE_DEPRECATED_CODE_4=ON \
+            -DKokkos_ENABLE_DEPRECATION_WARNINGS=OFF \
+            -DCMAKE_CXX_COMPILER=${{ matrix.cxx }} \
+            -DCMAKE_BUILD_TYPE=Release
+      - name: Build
+        run: |
+          ccache -z
+          cmake --build builddir --parallel 2
+          ccache -s
+      - name: Tests
+        working-directory: builddir
+        run: ctest --output-on-failure
+      - name: Gather benchmark results
+        run: |
+          mkdir ${{ env.BUILD_ID }}
+          find builddir/core/perf_test/ -name "*.json" -exec mv {} ${{ env.BUILD_ID }}/  \;
+      - name: Push benchmark results
+        uses: dmnemec/copy_file_to_another_repo_action@main
+        env:
+          API_TOKEN_GITHUB: ${{ secrets.API_TOKEN_GITHUB }}
+        with:
+          source_file: ${{ env.BUILD_ID }}
+          destination_repo: 'kokkos/kokkos-benchmark-results'
+          destination_branch: 'master'
+          user_email: 'kokkos@users.noreply.github.com'
+          user_name: 'Kokkos Developers'
--- a/.github/workflows/performance-benchmark.yml
+++ b/.github/workflows/performance-benchmark.yml
@@ -29,6 +29,7 @@ jobs:
       - name: Configure Kokkos
         run: |
           cmake -B builddir \
+            -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
             -DKokkos_ARCH_NATIVE=ON \
             -DKokkos_ENABLE_HWLOC=ON \
             -DKokkos_ENABLE_${{ matrix.backend }}=ON \
--- a/.github/workflows/performance-benchmark.yml
+++ b/.github/workflows/performance-benchmark.yml
@@ -33,9 +33,7 @@ jobs:
             -DKokkos_ARCH_NATIVE=ON \
             -DKokkos_ENABLE_HWLOC=ON \
             -DKokkos_ENABLE_${{ matrix.backend }}=ON \
-            -DKokkos_ENABLE_TESTS=ON \
             -DKokkos_ENABLE_BENCHMARKS=ON \
-            -DKokkos_ENABLE_EXAMPLES=ON \
             -DKokkos_ENABLE_DEPRECATED_CODE_4=ON \
             -DKokkos_ENABLE_DEPRECATION_WARNINGS=OFF \
             -DCMAKE_CXX_COMPILER=${{ matrix.cxx }} \
--- a/.github/workflows/performance-benchmark.yml
+++ b/.github/workflows/performance-benchmark.yml
@@ -54,7 +54,7 @@ jobs:
         if: ${{ github.ref == 'refs/heads/develop' }}
         uses: dmnemec/copy_file_to_another_repo_action@main
         env:
-          API_TOKEN_GITHUB: ${{ secrets.API_TOKEN_GITHUB }}
+          API_TOKEN_GITHUB: ${{ secrets.DALG24_PUSH_BENCHMARK_RESULTS }}
         with:
           source_file: ${{ env.BUILD_ID }}
           destination_repo: 'kokkos/kokkos-benchmark-results'
--- a/.github/workflows/performance-benchmark.yml
+++ b/.github/workflows/performance-benchmark.yml
@@ -41,7 +41,8 @@ jobs:
       - name: Build
         run: |
           ccache -z
-          cmake --build builddir --parallel 2
+          NUM_CPU=$(grep -c processor /proc/cpuinfo)
+          cmake --build builddir --parallel ${NUM_CPU}
           ccache -s
       - name: Tests
         working-directory: builddir
--- a/.github/workflows/performance-benchmark.yml
+++ b/.github/workflows/performance-benchmark.yml
@@ -3,7 +3,7 @@ on:
   push:
     branches:
       - develop
-      - performance-results-visualization
+  pull_request:
 
 jobs:
   CI:
@@ -51,6 +51,7 @@ jobs:
           mkdir ${{ env.BUILD_ID }}
           find builddir/core/perf_test/ -name "*.json" -exec mv {} ${{ env.BUILD_ID }}/  \;
       - name: Push benchmark results
+        if: ${{ github.ref == 'refs/heads/develop' }}
         uses: dmnemec/copy_file_to_another_repo_action@main
         env:
           API_TOKEN_GITHUB: ${{ secrets.API_TOKEN_GITHUB }}
--- a/.github/workflows/performance-benchmark.yml
+++ b/.github/workflows/performance-benchmark.yml
@@ -29,13 +29,10 @@ jobs:
       - name: Configure Kokkos
         run: |
           cmake -B builddir \
-            -DCMAKE_INSTALL_PREFIX=/usr \
             -DKokkos_ARCH_NATIVE=ON \
             -DKokkos_ENABLE_HWLOC=ON \
             -DKokkos_ENABLE_${{ matrix.backend }}=ON \
             -DKokkos_ENABLE_BENCHMARKS=ON \
-            -DKokkos_ENABLE_DEPRECATED_CODE_4=ON \
-            -DKokkos_ENABLE_DEPRECATION_WARNINGS=OFF \
             -DCMAKE_CXX_COMPILER=${{ matrix.cxx }} \
             -DCMAKE_BUILD_TYPE=Release
       - name: Build
--- a/.github/workflows/performance-benchmark.yml
+++ b/.github/workflows/performance-benchmark.yml
@@ -29,12 +29,12 @@ jobs:
       - name: Configure Kokkos
         run: |
           cmake -B builddir \
-            -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
             -DKokkos_ARCH_NATIVE=ON \
             -DKokkos_ENABLE_HWLOC=ON \
             -DKokkos_ENABLE_${{ matrix.backend }}=ON \
             -DKokkos_ENABLE_BENCHMARKS=ON \
             -DCMAKE_CXX_COMPILER=${{ matrix.cxx }} \
+            -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
             -DCMAKE_BUILD_TYPE=Release
       - name: Build
         run: |
--- a/.github/workflows/performance-benchmark.yml
+++ b/.github/workflows/performance-benchmark.yml
@@ -58,6 +58,6 @@ jobs:
         with:
           source_file: ${{ env.BUILD_ID }}
           destination_repo: 'kokkos/kokkos-benchmark-results'
-          destination_branch: 'master'
+          destination_branch: 'main'
           user_email: 'kokkos@users.noreply.github.com'
           user_name: 'Kokkos Developers'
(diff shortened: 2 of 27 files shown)


cmake/Kokkos_Version_Info.hpp
```diff
@@ -14,8 +14,8 @@
 //
 //@HEADER
 
-#ifndef GIT_VERSION_H
-#define GIT_VERSION_H
+#ifndef KOKKOS_GIT_VERSION_INFO_H
```
cz4rs, line 17: Unrelated fix - carry over from previous PR.

.github/workflows/continuous-integration-workflow.yml
```diff
@@ -124,3 +126,20 @@ jobs:
...
+      - name: Push benchmark results
+        if: |
+          github.ref == 'refs/heads/performance-results-visualization' &&
+          matrix.distro == 'ubuntu:latest'
+        uses: dmnemec/copy_file_to_another_repo_action@main
+        env:
+          API_TOKEN_GITHUB: ${{ secrets.API_TOKEN_GITHUB }}
+        with:
+          source_file: ${{ env.BUILD_ID }}
+          destination_repo: 'kokkos/kokkos-benchmark-results'
```
cz4rs, line 142: FIXME: this repository doesn't exist yet.

.github/workflows/continuous-integration-workflow.yml
```diff
@@ -124,3 +126,20 @@ jobs:
           cmake -B builddir -DCMAKE_CXX_COMPILER=${{ matrix.cxx }}
           cmake --build builddir
           cmake --build builddir --target test
+      - name: Gather benchmark results
+        run: |
+          mkdir ${{ env.BUILD_ID }}
+          find builddir/core/perf_test/ -name "*.json" -exec mv {} ${{ env.BUILD_ID }}/  \;
+      - name: Push benchmark results
+        if: |
+          github.ref == 'refs/heads/performance-results-visualization' &&
```
cz4rs, line 135: FIXME: change to `develop` before merging
```suggestion
          github.ref == 'refs/heads/develop' &&
```

.github/workflows/continuous-integration-workflow.yml
```diff
@@ -124,3 +126,20 @@ jobs:
...
+          matrix.distro == 'ubuntu:latest'
```
cz4rs, line 136: TODO: decide what set of conditions should be used here (which builds should report their results)

cz4rs: Use a separate `.yml` file dedicated to benchmarks and decide which build / configuration to use there.

.github/workflows/performance-benchmark.yml
```diff
@@ -0,0 +1,66 @@
...
+        run: |
+          mkdir ${{ env.BUILD_ID }}
+          find builddir/core/perf_test/ -name "*.json" -exec mv {} ${{ env.BUILD_ID }}/  \;
+      - name: Push benchmark results
+        uses: dmnemec/copy_file_to_another_repo_action@main
+        env:
+          API_TOKEN_GITHUB: ${{ secrets.API_TOKEN_GITHUB }}
+        with:
+          source_file: ${{ env.BUILD_ID }}
+          destination_repo: 'cz4rs/kokkos-benchmark-results'
```
cz4rs, line 63:
```suggestion
          destination_repo: 'kokkos/kokkos-benchmark-results'
```

.github/workflows/performance-benchmark.yml
```diff
@@ -0,0 +1,66 @@
+name: github-benchmarks
+on:
+  push:
+    branches:
+      - develop
+      - performance-results-visualization
```
cz4rs, line 6:
```suggestion
```

cz4rs:
Note: regular CI jobs have been temporarily disabled for this one to avoid wasting CI throughput (commits marked with `REMOVE_ME`).

cz4rs:
Sample plots:
- low number of samples (skipping smoothing and linear regression)
image

- complete plot
image

Note: this PR only takes care of gathering and storing the benchmark results. Processing and displaying them will be done in a separete PR.

.github/workflows/performance-benchmark.yml
```diff
@@ -0,0 +1,66 @@
...
+
+jobs:
+  CI:
+    continue-on-error: true
+    strategy:
+      matrix:
```
cz4rs, line 12: Results on GitHub hosted runners (free tier) are wildly unstable. We might as well remove the `matrix` and limit ourselves to single configuration. Leaving at least `g++` and `clang` variants for now.

.github/workflows/performance-benchmark.yml
cz4rs: This should only be triggered on push to `develop`.
Test runs can be observed on my branch, for example https://github.com/cz4rs/kokkos/actions/runs/3741965180.

cz4rs: Another option would be to run the benchmarks on PRs also, just without pushing the results further.

.github/workflows/performance-benchmark.yml
cz4rs: Using private repository for now since `kokkos/kokkos-benchmark-results` (or whatever the name should be) does not exist yet.

core/perf_test/CMakeLists.txt
```diff
@@ -139,6 +139,7 @@ ELSE()
   target_compile_options(benchmark_main PRIVATE -w)
 ENDIF()
 
+ADD_CUSTOM_TARGET(kokkos_all_benchmarks)
```
dalg24, line 142: Is that in order to exclude `PerformanceTest_Benchmark` ?

.github/workflows/performance-benchmark.yml
```diff
@@ -0,0 +1,64 @@
...
+      - name: Configure Kokkos
+        run: |
+          cmake -B builddir \
+            -DCMAKE_INSTALL_PREFIX=/usr \
+            -DKokkos_ARCH_NATIVE=ON \
+            -DKokkos_ENABLE_HWLOC=ON \
+            -DKokkos_ENABLE_${{ matrix.backend }}=ON \
+            -DKokkos_ENABLE_TESTS=ON \
+            -DKokkos_ENABLE_BENCHMARKS=ON \
+            -DKokkos_ENABLE_EXAMPLES=ON \
```
dalg24, line 38: Why bother with tests and examples?

core/perf_test/CMakeLists.txt
cz4rs: With more tests being ported (https://github.com/kokkos/kokkos/pull/5620), I think this might serve as a convenience target to build all benchmark binaries.

dalg24: My point was more how necessary is that?  If you only enable benchmarks and build, is this extra level of granularity actually useful?

cz4rs: Alright, I see you point. First I need to to find a way to build benchmarks without enabling tests.
`core/perf_test/` is currently added with ` KOKKOS_ADD_TEST_DIRECTORIES(perf_test)` which contains:
```CMake
    IF(KOKKOS_ENABLE_TESTS)
      FOREACH(TEST_DIR ${ARGN})
        ADD_SUBDIRECTORY(${TEST_DIR})
      ENDFOREACH()
    ENDIF()
```

dalg24: >  I need to to find a way to build benchmarks without enabling tests.

Oh right.  In ArborX we do 
https://github.com/arborx/ArborX/blob/0633bfdcac44086356a1ffefc143b146dfd40339/CMakeLists.txt#L151-L153

I am not 100% sure what would be the proper way to do that with TriBITS

dalg24: But yes conditionally adding the subdirectories sounds fine to me

.github/workflows/performance-benchmark.yml
cz4rs: Disabled / removed.

core/perf_test/CMakeLists.txt
cz4rs: Building only benchmarks works fine now. Without having to worry about Trilinos support, newly introduced `KOKKOS_ADD_BENCHMARK_DIRECTORY` boils down to calling `add_subdirectory`. This also means that I've had to move away from <angled> includes in the source files.
67a92d3 176ae8b

core/CMakeLists.txt
```diff
@@ -6,10 +6,23 @@ IF (NOT Kokkos_INSTALL_TESTING)
   ADD_SUBDIRECTORY(src)
 ENDIF()
 
+FUNCTION(KOKKOS_ADD_BENCHMARK_DIRECTORY DIR_NAME)
```
dalg24, line 9: Why did you decide to add it here rather than next to the definition of `KOKKOS_ADD_TEST_DIRECTORIES`?
(Not asking for any change at this time)

core/perf_test/CMakeLists.txt
```diff
@@ -15,102 +15,97 @@ IF ((KOKKOS_ENABLE_OPENMPTARGET OR KOKKOS_ENABLE_OPENACC) AND KOKKOS_CXX_COMPILE
   RETURN()
 ENDIF()
 
+IF(KOKKOS_ENABLE_TESTS)
```
dalg24, line 18: We should probably comment about the intent here.  Essentially all the `PerformanceTest_*` executables are considered as being part of the regular tests, not part of the benchmarks.

core/perf_test/PerfTest_ViewCopy_Raw.cpp
```diff
@@ -14,7 +14,7 @@
 //
 //@HEADER
 
-#include <PerfTest_ViewCopy.hpp>
+#include "PerfTest_ViewCopy.hpp"
```
dalg24, line 17: Why did you change?  Same comment everywhere below.

cz4rs: I was getting:
```
error: 'PerfTest_ViewCopy.hpp' file not found with <angled> include; use "quotes" instead
```
 This has worked before because of `  KOKKOS_INCLUDE_DIRECTORIES(${CMAKE_CURRENT_BINARY_DIR})` in the respective `CMakeLists.txt` but I don't see any reason to use that. With the header in the same directory as the `.cpp` file, double quotes work just fine.

core/CMakeLists.txt
cz4rs: It was in the spirit of staying away from Trilinos / Tribits in our CMake (`cmake/kokkos_tribits.cmake` precisely).
I am thinking about creating something like `cmake/kokkos_benchmark.cmake` though (with the content of this function + some general bits from `cmake/core/perf_test/CMakeLists.txt`).

cz4rs: Also keeping it local since it's the only call site for now, although I agree that it looks out of place here.

core/perf_test/CMakeLists.txt
cz4rs: I will add the explanation. Also this should all eventually get ported into benchmarks, which will make this `IF` obsolete.

core/perf_test/PerfTest_ViewCopy_Raw.cpp
dalg24: Mmm actually that file would be in `CMAKE_CURRENT_SOURCE_DIR`.  I feel like we should do `target_include_directories(... ${CMAKE_CURRENT_SOURCE_DIR})` instead

cz4rs: True about the `CMAKE_CURRENT_SOURCE_DIR` :ok_hand: - it's there in the subsequent line:
https://github.com/kokkos/kokkos/blob/d9d43bf62998854617e9224b074775aa1e6fa1ba/core/perf_test/CMakeLists.txt#L61-L62
So this stays intact and `PerformanceTest_*` can continue to work as they were.
Are you ok with NOT adding `target_include_directories(... ${CMAKE_CURRENT_SOURCE_DIR})` for benchmarks? (is there are reason to add that apart from using angled brackets?)

core/perf_test/CMakeLists.txt
cz4rs: Comment added: 327aac5.

core/CMakeLists.txt
```diff
@@ -6,10 +6,23 @@ IF (NOT Kokkos_INSTALL_TESTING)
...
+  IF(NOT Kokkos_ENABLE_BENCHMARKS)
+    RETURN()
+  ENDIF()
+
+  IF(KOKKOS_HAS_TRILINOS)
+    message(
+      STATUS
```
dalg24, line 16: Should it be a warning?

.github/workflows/performance-benchmark.yml
```diff
@@ -0,0 +1,63 @@
...
+            -DKokkos_ENABLE_${{ matrix.backend }}=ON \
+            -DKokkos_ENABLE_BENCHMARKS=ON \
+            -DKokkos_ENABLE_DEPRECATED_CODE_4=ON \
+            -DKokkos_ENABLE_DEPRECATION_WARNINGS=OFF \
+            -DCMAKE_CXX_COMPILER=${{ matrix.cxx }} \
+            -DCMAKE_BUILD_TYPE=Release
+      - name: Build
+        run: |
+          ccache -z
+          cmake --build builddir --parallel 2
```
fnrizzi, line 44: should we detect the parallel rather than hardwiring it to 2 ?

.github/workflows/performance-benchmark.yml
```diff
@@ -0,0 +1,63 @@
...
+    container:
+      image: ghcr.io/kokkos/ci-containers/${{ matrix.distro }}
+    env:
+      BUILD_ID: ${{ matrix.distro }}-${{ matrix.cxx }}-${{ matrix.backend }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v3
+      - uses: actions/cache@v3
+        with:
+          path: ~/.ccache
```
dalg24, line 26: You for got to configure with `-D CMAKE_CXX_COMPILER_LAUNCHER=ccache`

.github/workflows/performance-benchmark.yml
fnrizzi: '''
     - name: Build tests
        run: |
          export NUM_CPU=$(grep -c processor /proc/cpuinfo)
          cmake --build $BUILD_DIR -j $NUM_CPU
'''

.github/workflows/performance-benchmark.yml
```diff
@@ -0,0 +1,63 @@
...
+          find builddir/core/perf_test/ -name "*.json" -exec mv {} ${{ env.BUILD_ID }}/  \;
+      - name: Push benchmark results
+        if: ${{ github.ref == 'refs/heads/develop' }}
+        uses: dmnemec/copy_file_to_another_repo_action@main
+        env:
+          API_TOKEN_GITHUB: ${{ secrets.API_TOKEN_GITHUB }}
+        with:
+          source_file: ${{ env.BUILD_ID }}
+          destination_repo: 'kokkos/kokkos-benchmark-results'
+          destination_branch: 'master'
```
cz4rs, line 61:
```suggestion
          destination_branch: 'master'
```

.github/workflows/performance-benchmark.yml
```diff
@@ -0,0 +1,63 @@
...
+        run: ctest --output-on-failure
+      - name: Gather benchmark results
+        run: |
+          mkdir ${{ env.BUILD_ID }}
+          find builddir/core/perf_test/ -name "*.json" -exec mv {} ${{ env.BUILD_ID }}/  \;
+      - name: Push benchmark results
+        if: ${{ github.ref == 'refs/heads/develop' }}
+        uses: dmnemec/copy_file_to_another_repo_action@main
+        env:
+          API_TOKEN_GITHUB: ${{ secrets.API_TOKEN_GITHUB }}
```
cz4rs, line 57:
```suggestion
          API_TOKEN_GITHUB: ${{ secrets.DALG24_PUSH_BENCHMARK_RESULTS }}
```

core/CMakeLists.txt
cz4rs: As long as skipping the benchmarks is not a big deal for Trilinos, `STATUS` should be fine.

.github/workflows/performance-benchmark.yml
cz4rs: Added that, seems better than hardcoding :ok_hand:

.github/workflows/performance-benchmark.yml
cz4rs: Added that for both `continuous-integration-workflow.yml` and `performance-benchmark.yml`.

.github/workflows/performance-benchmark.yml
```diff
@@ -0,0 +1,63 @@
...
+      - name: Configure Kokkos
+        run: |
+          cmake -B builddir \
+            -DCMAKE_INSTALL_PREFIX=/usr \
+            -DKokkos_ARCH_NATIVE=ON \
+            -DKokkos_ENABLE_HWLOC=ON \
+            -DKokkos_ENABLE_${{ matrix.backend }}=ON \
+            -DKokkos_ENABLE_BENCHMARKS=ON \
+            -DKokkos_ENABLE_DEPRECATED_CODE_4=ON \
+            -DKokkos_ENABLE_DEPRECATION_WARNINGS=OFF \
```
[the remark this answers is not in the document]
cz4rs: Removed that :+1:

.github/workflows/continuous-integration-workflow.yml
```diff
@@ -93,6 +93,7 @@ jobs:
         run: |
           cmake -B builddir \
             -DCMAKE_INSTALL_PREFIX=/usr \
+            -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
```
dalg24, line 96: Did you mean to fix that here?

cmake/Kokkos_Version_Info.hpp
dalg24: I will close my eyes this time but in the future do this in a separate PR

.github/workflows/continuous-integration-workflow.yml
cz4rs: Yes, I figured that this can be handled in the same PR.

dalg24: Also prefer putting it next to the `DCMAKE_CXX_COMPILER` option

.github/workflows/performance-benchmark.yml
```diff
@@ -0,0 +1,62 @@
...
+        uses: actions/checkout@v3
+      - uses: actions/cache@v3
+        with:
+          path: ~/.ccache
+          key: kokkos-${{ matrix.distro }}-${{ matrix.cxx }}-${{ matrix.openmp }}-${github.ref}-${{ github.sha }}
+          restore-keys: kokkos-${{ matrix.distro }}-${{ matrix.cxx }}-${{ matrix.openmp }}-${{github.ref}}
+      - name: Configure Kokkos
+        run: |
+          cmake -B builddir \
+            -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
```
dalg24, line 32: Please move it next to the other CMAKE_CXX_ option

.github/workflows/continuous-integration-workflow.yml
dalg24: > Yes, I figured that this can be handled in the same PR.

OK for this time but move it next to `CMAKE_CXX_COMPILER`

cz4rs: Done (7e651ca).

.github/workflows/performance-benchmark.yml
cz4rs: Done (https://github.com/kokkos/kokkos/commit/7e651ca70ba3c9df2c89da71e54a8fd5c9124309).

dalg24 (approved):

PR #5691, opened 2022-12-15
Update nvcc_wrapper default arch to work with CUDA 12

crtrott:
Our cmake config uses nvcc_wrapper at some point without architecture flag, and thus the default architecture is used. CUDA 12 dropped support for Kepler so 3.5 will not be accepted. arch_70 is supported by CUDA 9 onward, and is unlikely to be dropped anytime soon from CUDA.

### DIFF
--- a/bin/nvcc_wrapper
+++ b/bin/nvcc_wrapper
@@ -10,10 +10,12 @@
 # Default settings: change those according to your machine.  For
 # example, you may have have two different wrappers with either icpc
 # or g++ as their back-end compiler.  The defaults can be overwritten
-# by using the usual arguments (e.g., -arch=sm_30 -ccbin icpc).
+# by using the usual arguments (e.g., -arch=sm_80 -ccbin icpc).
+# sm_70 is supported by every CUDA version from 9-12 and is thus
+# chosen as default
 
-default_arch="sm_35"
-#default_arch="sm_50"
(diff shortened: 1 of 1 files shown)


dalg24:
Please update the description and explain why we need to change and why 70 is the right compute capability to chose as new default.

bin/nvcc_wrapper
```diff
@@ -12,7 +12,7 @@
 # or g++ as their back-end compiler.  The defaults can be overwritten
 # by using the usual arguments (e.g., -arch=sm_30 -ccbin icpc).
 
-default_arch="sm_35"
+default_arch="sm_70"
```
fnrizzi, line 15: Sm_70 is for Volta right ? Which I read online is for cuda9 onward correct?

masterleinad (commented):
Do we need to set a default architecture at all?

crtrott:
> Do we need to set a default architecture at all?

Yeah I thought about just getting rid of it, but then I remembered that there were quite a few people who modify the script to set their arch instead of providing the -arch flag properly ...

fnrizzi (approved):

PhilMiller (approved):

PhilMiller:
Test failures are irrelevant things in HIP and OpenMPTarger

PR #5693, opened 2022-12-15
Add Hopper support and update nvcc_wrapper to work with CUDA 12

crtrott:


PhilMiller:
Should this be a patch candidate?

core/src/Cuda/Kokkos_Cuda_Half_Conversion.hpp
```diff
@@ -290,7 +290,8 @@ KOKKOS_INLINE_FUNCTION
 // Go in this branch if CUDA version is >= 11.0.0 and less than 11.1.0 or if the
 // architecture is not Ampere
 #if CUDA_VERSION >= 11000 && \
-    (CUDA_VERSION < 11010 || !defined(KOKKOS_ARCH_AMPERE))
+    (CUDA_VERSION < 11010 || \
+     !(defined(KOKKOS_ARCH_AMPERE) || defined(KOKKOS_ARCH_HOPPER)))
```
PhilMiller, line 294: Comment above mis-matches.

As a potential refactoring, should we use some sort of define for CUDA compute capability, so we can durably say (e.g.) `KOKKOS_ARCH_CUDA_COMPUTE < 80` rather than having to update the condition for each *new* architecture for which we add support?

PhilMiller (approved):

fnrizzi (approved):

dalg24 (changes requested):
You need to cherry-pick #5686 and #5688 as well

weinbe2:
@crtrott can I help with this in some way? I want to make sure Hopper support makes it into Kokkos 4.0

crtrott:
> @crtrott can I help with this in some way? I want to make sure Hopper support makes it into Kokkos 4.0

This is for the 3.7.02 patch release, we gonna do AFTER 4.0 is out. 4.0 release candidate has the Hopper support in it already. Updating 3.7.02 has been stalled due to some CI issues we don't fully understand, but we also didn't prioritize it since its post 4.0. We are planning on maintaining the 3.7 branch for a while since its the last C++14 compatible release, which means we gonna add bug fixes etc.

weinbe2:
It looks like the two cherry-picks were already merged in @dalg24 ?

weinbe2:
Understood, thank you explaining. I was thrown off by the bifurcation.

crtrott:
Gonna merge this without fixing CI for 3.7 if we do another 3.7.x release we can revisit nvhpc testing.

PR #4295, opened 2021-09-03
Cuda: Adding implementation to use cudaFuncCachePreferEqual when needed.

rgayatri23:
The PR adds an option to use `cudaFuncCachePreferEqual` cache config option instead of always choosing `cudaFuncCachePreferShared`.

PR #5695, opened 2022-12-15
CUDA: Stop setting global preference for shared memory over cache

PhilMiller:
In #4295, this preference was found to hurt performance across a range of benchmarks.


PhilMiller:
@crtrott per [your comment](https://github.com/kokkos/kokkos/pull/4295#issuecomment-1352043001) do we need to do anything special now that the global preference is dropped?

crtrott:
This is not how this works...

PhilMiller:
Per Christian, there's more subtlety to it than this. He's got work in progress, so closing this PR.

PR #5697, opened 2022-12-16
Fixup signature of `hip_parallel_launch_global_memory`

dalg24:
Fixup for #5448 

Need to investigate why this was not caught by the unit tests


crtrott (approved):

Rombur:
I don't think that's the correct solution. We do the same thing in CUDA. For a LocalLaunch we pass the driver by value and for a GlobalLaunch we pass a pointer

Rombur:
I know where the bug is from. I'll create a PR later

crtrott (changes requested):
Sorry I was stupid when I approved this is wrong.

dalg24:
Closing in favor of upcoming fix from Bruno

PR #5702, opened 2022-12-16
cuda: Remove hard error check on compute compatibility

muraj:
CUDA Compute compatibility is defined across architectures via PTX JIT, applications/Kokkos just needs to ensure the features are conditionally compiled for those virtual architectures that don't support them.

Allowing this allows for applications and libraries compiled with older CUDA toolkits to be forward compatible to newer hardware, allowing for more compatibility and utility in Kokkos.


masterleinad:
See the discussion at https://github.com/kokkos/kokkos/issues/3612 and https://github.com/kokkos/kokkos/pull/5527.
Also, pull requests need to target the `develop` branch.

crtrott:
This was addressed in #5527

PR #5705, opened 2022-12-19
SYCL: Pass Xsycl-target-backend* only to the linker

masterleinad:
Without this, I get warnings such as
```
-Xsycl-target-backend -device 12.60.7’ [-Wunused-command-line-argument].
```
The option really only is used by the linker for AOT compilation.


lucbv (approved):
Yep, I want that too!

dalg24 (changes requested):
OK with the changes to the cmake file, not OK with the line addition in the jenkinsfile

masterleinad:
> OK with the changes to the cmake file, not OK with the line addition in the jenkinsfile

Rebased on top of `develop`.

dalg24 (approved):

Issue #5419, opened 2022-09-01
Improve `ExecutionSpace::print_configuration` in all backends


dalg24:
The overarching goal is to have `Kokkos::print_configuration` print something actually useful.  At the moment it is clearly lacking important things like Kokkos commit hash, Desul bundled or external (specify version/commit hash in that case)...

This issue is specifically asking backends developer what information they would themselves like to see in a bug report.  This is the kind of things we want to be printed, at least in verbose mode.

The list below is meant to keep track of our improving in all backends and who may be working on it (some are unassigned at the moment).  Make sure you cross reference this issue when you open a PR.
* [ ] CUDA
* [ ] HIP (@Rombur)
* [ ] SYCL (@masterleinad)
* [ ] OpenMPTarget (@rgayatri23)
* [ ] OpenACC (@seyonglee)
* [ ] HPX (@msimberg)
* [ ] OpenMP
* [ ] Threads
* [ ] Serial

PhilMiller:
For the commit hash, there's a bit of CMake code we can grab from DARMA-tasking/vt that embeds it in a variable in a cpp file. That's preferable to a `#define` or sticking the value in a header, because it avoids forcing a full rebuild when very little has changed.

masterleinad:
For SYCL+CUDA, we currently get output such as
```
Devices:
  KOKKOS_ENABLE_SYCL: yes

Runtime Configuration:
macro  KOKKOS_ENABLE_SYCL : defined
Name: Tesla V100-PCIE-32GB
Driver Version: CUDA 11.7
Is Host: 0
Is CPU: 0
Is GPU: 1
Is Accelerator: 0
Vendor Id: 4318
Max Compute Units: 80
Max Work Item Dimensions: 3
Max Work Group Size: 1024
Preferred Vector Width Char: 1
Preferred Vector Width Short: 1
Preferred Vector Width Int: 1
Preferred Vector Width Long: 1
Preferred Vector Width Float: 1
Preferred Vector Width Double: 1
Preferred Vector Width Half: 0
Native Vector Width Char: 1
Native Vector Width Short: 1
Native Vector Width Int: 1
Native Vector Width Long: 1
Native Vector Width Float: 1
Native Vector Width Double: 1
Native Vector Width Half: 0
Address Bits: 64
Image Support: 1
Max Mem Alloc Size: 1073741824
Max Read Image Args: 128
Image2d Max Width: 140724603584512
Image2d Max Height: 140724603518976
[...]

PR #5707, opened 2022-12-19
Address deprecations in oneAPI 2023.0.0

masterleinad:
Updating to a more recent version in our CI blocks a number of other pull requests for improvements on the relevant systems. Unfortunately, there are various tests in the SYCL+Cuda CI failing after updating the compiler. It appears very unlikely that those are related to fixing the deprecations.
My preference would be to move forward with this, disable failing tests SYCL+Cuda (the tests are still all passing on the testbeds with Intel GPUs), and try fixing those as a follow-up.
Of course, I will continue to run all the tests on the testbeds manually regularly.

The deprecations in this pull request include:
- Various `sycl::device` member functions used in `print_configuration`. I didn't bother replacing them but just removed them. They shouldn't contribute too much interesting information and removing them might also work toward #5419. In detail:
  - `sycl::device::is_host()`
  - `sycl::device::get_info<sycl::device::image_support>()`
  - `sycl::device::get_info<sycl::device::host_unified_memory>()`
  - `sycl::device::get_info<sycl::device::is_endian_little>()`
  - `sycl::device::get_info<sycl::device::is_compiler_available>()`
  - `sycl::device::get_info<sycl::device::is_linker_available>()`
  - `sycl::device::get_info<sycl::device::queue_profiling>()`
  - `sycl::device::get_info<sycl::device::printf_buffer_size>()`
  - `sycl::device::get_info<sycl::device::preferred_interop_user_sync>()`
- `sycl::accessor<value_type, size, sycl::access::mode::read_write, sycl::access::target::local` -> `sycl::local_accessor<value_type, size>`
- `sycl::kernel.get_info<sycl::info::kernel_device_specific::max_sub_group_size>(sycl::device, sycl::range<3>)` drops its last argument which was presumably never used anyway.
- `sycl::default_selector()` -> `sycl::default_selector_v`

Apart from that, I couldn't get the build script working for the `intel/llvm` version selected and just manually copied its actions and simplified it until it worked. In a follow-up pull request, we should explore using the public oneAPI 2023.0.0 release with `Codeplay`'s `SYCL+Cuda` plugin.


crtrott:
Uhm what are all those changes which do not seem to have anything to do with OneAPI?

masterleinad:
> Uhm what are all those changes which do not seem to have anything to do with OneAPI?

This is a draft pull request and WIP. I need to fix the failing tests and prefer not to enable other CI until then.

core/unit_test/sycl/TestSYCL_InterOp_Init.cpp
```diff
@@ -29,8 +29,7 @@ TEST(sycl, raw_sycl_interop) {
   Kokkos::Experimental::SYCL default_space;
   sycl::context default_context = default_space.sycl_queue().get_context();
 
-  sycl::default_selector device_selector;
-  sycl::queue queue(default_context, device_selector);
+  sycl::queue queue(default_context, sycl::default_selector_v);
```
dalg24, line 32: Presumably, this is just a new available syntax, not a deprecation.  Please propose these changes (prefer `sycl::default_selector_v `) in a separate PR.

core/src/SYCL/Kokkos_SYCL_Parallel_Reduce.hpp
```diff
@@ -293,12 +293,8 @@ class ParallelReduce<FunctorType, Kokkos::RangePolicy<Traits...>, ReducerType,
...
 
       auto reduction_lambda_factory =
-          [&](sycl::accessor<value_type, 1, sycl::access::mode::read_write,
-                             sycl::access::target::local>
-                  local_mem,
-              sycl::accessor<unsigned int, 1, sycl::access::mode::read_write,
-                             sycl::access::target::local>
-                  num_teams_done,
+          [&](sycl::local_accessor<value_type, 1> local_mem,
+              sycl::local_accessor<unsigned int, 1> num_teams_done,
```
dalg24, line 297: Same comment here, presumably not actually deprecated.

dalg24 (changes requested):
Please confirm that all changes are related to an actual deprecation or withdraw them.
And discuss our not using the build python scripts any more

dalg24 (approved):

algorithms/unit_tests/TestStdAlgorithmsUniqueCopy.cpp
```diff
@@ -322,6 +322,9 @@ void run_all_scenarios() {
 }
 
 TEST(std_algorithms_mod_seq_ops, unique_copy) {
+#if defined(KOKKOS_ENABLE_SYCL) && !defined(KOKKOS_ARCH_INTEL_GPU)
```
e10harvey, line 325: Should this be `&& !defined(KOKKOS_ENABLE_CUDA)`?

masterleinad: No, you can't enable multiple backends anyway. I could replace `!defined(KOKKOS_ARCH_INTEL_GPU)` with something like `(defined(KOKKOS_ARCH_VEGA) || defined(KOKKOS_ARCH_AMPERE) || ...)` but I'm not convinced that it's worth it.
You need to provide an architecture flag when using `SYCL+Cuda` but otherwise, you don't. Hence, this check also restricts the tests run on, say `Intel GPU`s, if no architecture was specified (but then we don't enable some optimizations).

lucbv (approved):
I am fine with these changes, it's reasonable based on new sycl syntax

crtrott (approved):
I can't judge the deprecations but I am ok with this. Do we have any idea why this broke CUDA builds?

masterleinad:
> I can't judge the deprecations but I am ok with this. Do we have any idea why this broke CUDA builds?

I have no idea what changed for the tests to break.

Issue #5700, opened 2022-12-16
Trilinos configuration errors in some nightly builds


ndellingwood:
**Describe the bug**
Following merge of #5463 some nightly builds of Trilinos fail to configure with output:
```
03:30:30 CMake Error: install(EXPORT "Kokkos" ...) includes target "Kokkos_all_libs" which requires target "impl_git_version" that is not in any export set.
03:30:30 CMake Error in kokkos/CMakeLists.txt:
03:30:30   export called with target "Kokkos_all_libs" which requires target
03:30:30   "impl_git_version" that is not in any export set.
```

Thanks to @brian-kelley , he also pointed this out in https://github.com/kokkos/kokkos-kernels/issues/1631#issuecomment-1355488702 and identified the culprit lines of code in kokkos/CMakeLists.txt:

```
199 # Gather information about the runtime environment
200 INCLUDE(${KOKKOS_SRC_PATH}/cmake/build_env_info.cmake)
201 check_git_setup() #this line errored out with stuff below:
```

**Please also include the following items to support reproducing the bug**
**Reproducer** (kokkos-dev-2):
This can be reproduced with Trilinos' develop branch using Kokkos' develop branch via source override
```bash
# Repos
git clone -b develop https://github.com/trilinos/Trilinos.git
git clone -b develop https://github.com/kokkos/kokkos.git
# Symbolic link to your kokkos repo in Trilinos source directory for source override
cd Trilinos
ln -s <path-to-your-repo>/kokkos kokkos

cd $HOME
mkdir -p build
cd build

# Configure Trilinos
export ATDM_CONFIG_REGISTER_CUSTOM_CONFIG_DIR=${TRILINOS_DIR}/cmake/std/atdm/contributed/kokkos-dev-2
source ${TRILINOS_DIR}/cmake/std/atdm/load-env.sh kokkos-dev-2-cuda-opt

# Env script needs update - just swap modules for now
module swap sems-archive-cmake/3.12.2 sems-archive-cmake/3.17.1
module swap sems-archive-cuda/9.2 sems-archive-cuda/11.1
module swap sems-archive-gcc/7.2.0 sems-archive-gcc/8.3.0

export CXX=$KOKKOS_DIR/bin/nvcc_wrapper

cmake \
 -GNinja \
 -DTrilinos_CONFIGURE_OPTIONS_FILE:STRING=cmake/std/atdm/ATDMDevEnv.cmake \
[...]

PR #5708, opened 2022-12-19
Disable perf tests in Trilinos

crtrott:
the git env generation doesn't work in Trilinos correct, and we don't care to figure out how to fix that.


ndellingwood (approved):
This resolves the Trilinos configuration issues in #5700 , thanks!

cz4rs (approved):

cz4rs:
Note: there's another check here: https://github.com/kokkos/kokkos/blob/f64e5a66db145ffb8c6dd61b649ab914ddbc058a/core/perf_test/CMakeLists.txt#L110-L112
but disabling tests is better than simply generating and error ✅

dalg24:
Ignoring NVHPC timing based test failure and not waiting for SYCL build to finish

PR #5709, opened 2022-12-19
Fix CUDA lock arrays

masterleinad:
Alternative to https://github.com/kokkos/kokkos/pull/5619.

@stanmoore1 Can you please confirm that this works for you?


PhilMiller:
Would we bother with reverting the earlier change in 3.7.01 and applying this for 3.7.02? Or just going with this on `develop` for 4.0?

crtrott:
We may wanna do this in 3.7.02 too so that release is compatible with standalone Desul

crtrott:
Confirmed that this fixes LAMMPS

crtrott (approved):

dalg24 (approved):

Issue #5714, opened 2022-12-20
Seg fault when drawing random samples


dannys4:
**Describe the bug**
Forgive me if I'm using the random interface incorrectly, I haven't found many examples of it being used in practice. I'm encountering a segmentation fault with a simple RNG that looks like it should work. The MWE below seg faults on my computer.

**Please also include the following items to support reproducing the bug**
1. Compiler: g++ 10.3
2. Kokkos release: 3.6.1
3. x86 linux + OpenMP
4. `cmake -DKokkos_ENABLE_OPENMP=ON -DKokkos_ENABLE_SERIAL=ON -DBUILD_SHARED_LIBS=ON -DKokkos_CXX_STANDARD=17`
5. output from cmake command: N/A
6. MWE:
```cpp
#include <Kokkos_Core.hpp>

// Uniform generator on [0,1]
class UniformGenerator {
public:

using PoolType = Kokkos::Random_XorShift64_Pool<Kokkos::DefaultExecutionSpace>;
PoolType rand_pool;

UniformGenerator(): rand_pool() {}

void SampleImpl(Kokkos::View<double**, Kokkos::HostSpace> output) {
    unsigned int N = output.extent(1);
    Kokkos::parallel_for("uniform generator", N, KOKKOS_LAMBDA(const int& j) {
        typename PoolType::generator_type rgen = rand_pool.get_state();
        output(0,j) = rgen.drand();
    });
}
};

int main() {
    auto generator = std::make_shared<UniformGenerator>();
    unsigned int N_pts = 1000;
    Kokkos::View<double**, Kokkos::HostSpace> output ("output", 1, N_pts);
    generator->SampleImpl(output);
}
```
7. `KokkosCore_config.h` header file:
```cpp
#if !defined(KOKKOS_MACROS_HPP) || defined(KOKKOS_CORE_CONFIG_H)
#error \
    "Do not include KokkosCore_config.h directly; include Kokkos_Macros.hpp instead."
#else
#define KOKKOS_CORE_CONFIG_H
#endif

// KOKKOS_VERSION % 100 is the patch level
// KOKKOS_VERSION / 100 % 100 is the minor version
// KOKKOS_VERSION / 10000 is the major version
#define KOKKOS_VERSION 30601

/* Execution Spaces */
#define KOKKOS_ENABLE_SERIAL
#define KOKKOS_ENABLE_OPENMP
/* #undef KOKKOS_ENABLE_OPENMPTARGET */
[...]

PR #4218, opened 2021-08-05
Fix false sharing in random pool

crtrott:
The pool arrays are accessed by all threads, but the
    elements per thread are rather small and thus share cachelines.

    Making the arrays 2D using the second dimension for padding only,
    solves that problem. I have seen up to 200x improvement for
    20 threads on skylake running the random number example with 100k and 1
    as parameters. Serial performance is slightly reduced, but it only makes
    the "grep a generator slower" by the equivalent of one double load more
    and a few integer ops.

PR #5716, opened 2022-12-21
Cleanup random pool special member functions and precondition check in `get_state()`

dalg24:
Tentative fix for #5714
Default-constructed random pool are toxic because they don't properly initialize their "state".
~~I propose to remove the default constructor on the device side and to properly initialize with a default seed on the host.~~
**EDIT** Default-constructed pool have no state by design.  This PR is adding is a precondition check in `Random_XorShift{64,1024}_Pool::get_state()` that they are indeed states.
The device markups are removed from the default constructor and we prefer in-class initializer and defaulted special members.

I dropped the copy constructors and copy assignment operators as the compiler is able to generate them.  Note that the copy constructors were buggy and that we had forgotten to update them in #4218

I haven't actually investigated whether random pools actually get constructed on the device, neither in our codebase nor in Trilinos.

### DIFF
--- a/algorithms/src/Kokkos_Random.hpp
+++ b/algorithms/src/Kokkos_Random.hpp
@@ -897,19 +897,6 @@ class Random_XorShift64_Pool {
     init(seed, execution_space().concurrency());
   }
 
-  KOKKOS_INLINE_FUNCTION
-  Random_XorShift64_Pool(const Random_XorShift64_Pool& src)
-      : locks_(src.locks_), state_(src.state_), num_states_(src.num_states_) {}
-
-  KOKKOS_INLINE_FUNCTION
-  Random_XorShift64_Pool operator=(const Random_XorShift64_Pool& src) {
-    locks_      = src.locks_;
-    state_      = src.state_;
-    num_states_ = src.num_states_;
-    padding_    = src.padding_;
-    return *this;
-  }
-
   void init(uint64_t seed, int num_states) {
     if (seed == 0) seed = uint64_t(1318319);
     // I only want to pad on CPU like archs (less than 1000 threads). 64 is a
@@ -1146,23 +1133,6 @@ class Random_XorShift1024_Pool {
     num_states_ = 0;
 
     init(seed, execution_space().concurrency());
-  }
-
-  KOKKOS_INLINE_FUNCTION
-  Random_XorShift1024_Pool(const Random_XorShift1024_Pool& src)
-      : locks_(src.locks_),
-        state_(src.state_),
-        p_(src.p_),
-        num_states_(src.num_states_) {}
-
-  KOKKOS_INLINE_FUNCTION
-  Random_XorShift1024_Pool operator=(const Random_XorShift1024_Pool& src) {
-    locks_      = src.locks_;
-    state_      = src.state_;
-    p_          = src.p_;
-    num_states_ = src.num_states_;
-    padding_    = src.padding_;
-    return *this;
   }
 
   inline void init(uint64_t seed, int num_states) {
--- a/algorithms/src/Kokkos_Random.hpp
+++ b/algorithms/src/Kokkos_Random.hpp
@@ -895,6 +895,12 @@ class Random_XorShift64_Pool {
     KOKKOS_IF_ON_HOST(init(0, execution_space().concurrency());)
   }
 
+  KOKKOS_DEFAULTED_FUNCTION Random_XorShift64_Pool(
+      Random_XorShift64_Pool const&) = default;
+
+  KOKKOS_DEFAULTED_FUNCTION Random_XorShift64_Pool& operator=(
+      Random_XorShift64_Pool const&) = default;
+
   Random_XorShift64_Pool(uint64_t seed) {
     num_states_ = 0;
 
@@ -1144,6 +1150,12 @@ class Random_XorShift1024_Pool {
         num_states_ = 0;)
     KOKKOS_IF_ON_HOST(init(0, execution_space().concurrency());)
   }
+
+  KOKKOS_DEFAULTED_FUNCTION Random_XorShift1024_Pool(
+      Random_XorShift1024_Pool const&) = default;
+
+  KOKKOS_DEFAULTED_FUNCTION Random_XorShift1024_Pool& operator=(
+      Random_XorShift1024_Pool const&) = default;
 
   Random_XorShift1024_Pool(uint64_t seed) {
     num_states_ = 0;
--- a/algorithms/src/Kokkos_Random.hpp
+++ b/algorithms/src/Kokkos_Random.hpp
@@ -878,41 +878,32 @@ class Random_XorShift64_Pool {
   using execution_space = typename device_type::execution_space;
   using locks_type      = View<int**, device_type>;
   using state_data_type = View<uint64_t**, device_type>;
-  locks_type locks_;
-  state_data_type state_;
-  int num_states_;
-  int padding_;
+
+  locks_type locks_      = {};
+  state_data_type state_ = {};
+  int num_states_        = {};
+  int padding_           = {};
 
  public:
   using generator_type = Random_XorShift64<DeviceType>;
 
 #ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
   KOKKOS_FUNCTION
-  Random_XorShift64_Pool() {
-    KOKKOS_IF_ON_DEVICE((
-        /* nonsensical initialization in the name of backward compatibility */
-        num_states_ = 0; padding_ = 0;))
-    KOKKOS_IF_ON_HOST(init(0, execution_space().concurrency());)
-  }
+  Random_XorShift64_Pool() = default;
 
   KOKKOS_DEFAULTED_FUNCTION Random_XorShift64_Pool(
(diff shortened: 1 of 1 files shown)


algorithms/src/Kokkos_Random.hpp
```diff
@@ -886,29 +886,27 @@ class Random_XorShift64_Pool {
...
   Random_XorShift64_Pool() {
-    num_states_ = 0;
-    padding_    = 0;
+    KOKKOS_IF_ON_DEVICE((
+        /* nonsensical initialization in the name of backward compatibility */
+        num_states_ = 0; padding_ = 0;))
+    KOKKOS_IF_ON_HOST(init(0, execution_space().concurrency());)
   }
+
   Random_XorShift64_Pool(uint64_t seed) {
```
fnrizzi, line 898: should this be marked explicit?

fnrizzi, line 898: should this be marked explicit?

algorithms/src/Kokkos_Random.hpp
```diff
@@ -886,29 +886,27 @@ class Random_XorShift64_Pool {
...
     num_states_ = 0;
 
     init(seed, execution_space().concurrency());
   }
+#else
+  Random_XorShift64_Pool(uint64_t seed = 0) {
```
fnrizzi, line 904: should this be marked explicit?

algorithms/src/Kokkos_Random.hpp
```diff
@@ -886,29 +886,27 @@ class Random_XorShift64_Pool {
...
-    locks_      = src.locks_;
-    state_      = src.state_;
-    num_states_ = src.num_states_;
-    padding_    = src.padding_;
-    return *this;
+    init(seed, execution_space().concurrency());
   }
+#endif
 
   void init(uint64_t seed, int num_states) {
```
fnrizzi, line 911: why is init a public method? I think init should be private

algorithms/src/Kokkos_Random.hpp
```diff
@@ -1139,33 +1136,29 @@ class Random_XorShift1024_Pool {
...
+  }
 
-  inline Random_XorShift1024_Pool(uint64_t seed) {
+  Random_XorShift1024_Pool(uint64_t seed) {
     num_states_ = 0;
 
     init(seed, execution_space().concurrency());
   }
+#else
+  Random_XorShift1024_Pool(uint64_t seed = 0) {
```
fnrizzi, line 1154: should this be marked explicit?

algorithms/src/Kokkos_Random.hpp
```diff
@@ -1139,33 +1136,29 @@ class Random_XorShift1024_Pool {
...
-    p_          = src.p_;
-    num_states_ = src.num_states_;
-    padding_    = src.padding_;
-    return *this;
+    init(seed, execution_space().concurrency());
   }
+#endif
 
-  inline void init(uint64_t seed, int num_states) {
+  void init(uint64_t seed, int num_states) {
```
fnrizzi, line 1161: same comment about this being a public method

algorithms/src/Kokkos_Random.hpp
```diff
@@ -886,29 +886,27 @@ class Random_XorShift64_Pool {
  public:
   using generator_type = Random_XorShift64<DeviceType>;
 
-  KOKKOS_INLINE_FUNCTION
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  KOKKOS_FUNCTION
   Random_XorShift64_Pool() {
```
fnrizzi, line 891: unless i am misunderstanding, given how `KOKKOS_ENABLE_DEPRECATED_CODE_4` is guarding this code now, if we remove the depreacted code there is no constructor for the device.

algorithms/src/Kokkos_Random.hpp
fnrizzi, line 898: if one passes 0 to the seed, this calls `init(0,...` and inside init the seed is changed if it is zero. 
see my comment below inside init

algorithms/src/Kokkos_Random.hpp
```diff
@@ -886,29 +886,27 @@ class Random_XorShift64_Pool {
...
-    state_      = src.state_;
-    num_states_ = src.num_states_;
-    padding_    = src.padding_;
-    return *this;
+    init(seed, execution_space().concurrency());
   }
+#endif
 
   void init(uint64_t seed, int num_states) {
     if (seed == 0) seed = uint64_t(1318319);
```
fnrizzi, line 912: IMO this init method should use the seed passed to it as is without doing this check on 0. 
A proper seed should be put inside the constructors directly otherwise we are separating things that are to follow. 

This is what I mean: 
```
  KOKKOS_FUNCTION
  Random_XorShift64_Pool() {
    // ...
    KOKKOS_IF_ON_HOST(init(uint64_t(1318319), execution_space().concurrency());)
  }

  Random_XorShift64_Pool(uint64_t seed) {
    num_states_ = 0;
    init(seed, execution_space().concurrency());
  }
#else
  Random_XorShift64_Pool(uint64_t seed = 0) {
    num_states_ = 0;
    if (seed == 0){ seed = uint64_t(1318319); }
    init(seed, execution_space().concurrency());
  }
#endif

void init(uint64_t seed, ...)
{ 
  // if (seed == 0) seed = uint64_t(1318319); // remove this 
  // I only want to pad ...
  // ...
}
```

algorithms/src/Kokkos_Random.hpp
dalg24: See my comment above, but yes in general I would declare such constructor with the function specifier explicit.

algorithms/src/Kokkos_Random.hpp
dalg24: Resolving as it is a duplicate from another comment above.
I will make sure `Random_XorShift1024_Pool`'s interface is consistent with `Random_XorShift64_Pool`

algorithms/src/Kokkos_Random.hpp
dalg24: Marking it as revolved, will make sure that we are consistent.

algorithms/src/Kokkos_Random.hpp
```diff
@@ -886,29 +886,27 @@ class Random_XorShift64_Pool {
...
-  Random_XorShift64_Pool(const Random_XorShift64_Pool& src)
-      : locks_(src.locks_), state_(src.state_), num_states_(src.num_states_) {}
-
-  KOKKOS_INLINE_FUNCTION
-  Random_XorShift64_Pool operator=(const Random_XorShift64_Pool& src) {
-    locks_      = src.locks_;
-    state_      = src.state_;
-    num_states_ = src.num_states_;
-    padding_    = src.padding_;
-    return *this;
```
dalg24, line 910: I realize now I should probably preserve the being callable on the device when deprecated code is enabled.
Noting that the copy assignment operator returns by value instead of returning a reference.

algorithms/src/Kokkos_Random.hpp
dalg24: Actually it makes sense to provide a functionality to reseed the random generator engine.

fnrizzi: I think if we want a method to reseed it, I personally would make it: 
```
void reseed(uint64_t newSeed)
```
becaue `init` is too general, and reseeding does not need the num_states

dalg24:
> Default-constructed random pool are toxic because they don't properly initialize their "state".

Christian explained to me that the intent was that default-constructed random pool have no state.  Meaning calling the `get_state()` member function on a default-constructed  pool is a precondition violation.

algorithms/src/Kokkos_Random.hpp
dalg24: Outdated since the PR was updated.

algorithms/src/Kokkos_Random.hpp
dalg24: PR was updated but we are still intentionally removing the device-side default constructor.

algorithms/src/Kokkos_Random.hpp
dalg24: I opted for not adding any explicit function specifier.  I don't think it is worth the trouble.

algorithms/src/Kokkos_Random.hpp
dalg24: Since the default-constructor behavior was restored to an "uninitialized" state, I am not proposing any change to that function.

crtrott (approved):

fnrizzi:
> the intent was that default-constructed random pool have no state. Meaning calling the get_state() member function on a default-constructed pool is a precondition violation.

I don't really understand this... the only difference between the default constructor and the one we offer is the seed. 
why should we make a default constructed random pool be stateless?

dalg24:
Because, (sigh), Christian mentioned some particle-based code that uses static class object with a random generator pool as a data member with essentially that default constructor gettin called before `Kokkos::initialize`.

fnrizzi:
@dalg24 ha! that figure is pretty cool , never seen it before !

Issue #3498, opened 2020-10-19
TeamPolicy team_size_max doesn't account for scratch memory constraints


DavidPoliakoff:
Found via ExaMiniMD (@stanmoore1 ), when attempting to use autotuning. Autotuning just a vector length is functional, but when we try the code

```c++
  Kokkos::TeamPolicy<> policy(nlocal,Kokkos::AUTO,Kokkos::AUTO);

  Kokkos::parallel_for("ForceSNAP::compute",policy
      .set_scratch_size(1,Kokkos::PerThread(thread_scratch_size))
      .set_scratch_size(1,Kokkos::PerTeam(team_scratch_size))
    ,*this);
```

In order to tune team size and vector length for CUDA, a V100 errors out 

```console
what():  Kokkos failed to allocate memory for label "CudaSpace::ScratchMemory".  Allocation using MemorySpace 
named "Cuda" failed with the following error:  Allocation of size 9.978 G failed, likely due to insufficient memory. 
(The allocation mechanism was cudaMalloc().  The Cuda allocation returned the error code 
""cudaErrorMemoryAllocation".)
```

This is because as we sweep the team size, we eventually hit a value for which we can't allocate the requested amount of scratch memory. We need either

1) (ideal) team_size_max to account for memory requirements
2) (okay'ish'ish'ish) an ability to cap the team sizes and vector lengths when tuning (`Kokkos::AUTOLessThan(64)`, conceptually)

At a minimum, this should be documented in the TeamPolicy docs

PR #5718, opened 2022-12-21
Fix max scratch size calculation for level 0 of CUDA and HIP

### LINKED ISSUE #1811: Figure out the maximal number of elements in a shared memory view
I get the "Kokkos::Impl::ParallelFor< Cuda > insufficient shared memory" runtime error when trying to use TeamPolicy::scratch_size_max(0) / sizeof(int) to calculate the maximal number of element in a shared view. This is an example:
```
#include <Kokkos_Core.hpp>
#include <Kokkos_DualView.hpp>
using ExecutionSpace = Kokkos::DefaultExecutionSpace;
using namespace std;

template < class DualView >
void syncToHost(DualView&& data) {
  using MemorySpace = typename std::decay_t<DualView>::memory_space;
  using HostMirrorSpace = typename std::decay_t<DualView>::host_mirror_space;
  std::forward<DualView>(data).template modify<MemorySpace>();
  std::forward<DualView>(data).template sync<HostMirrorSpace>();
}

template < class DualView >
void syncToDevice(DualView&& data) {
  using MemorySpace = typename std::decay_t<DualView>::memory_space;
  using HostMirrorSpace = typename std::decay_t<DualView>::host_mirror_space;
  std::forward<DualView>(data).template modify<HostMirrorSpace>();
  std::forward<DualView>(data).template sync<MemorySpace>();
}

template < class T, class E >
struct F {
  using TeamPolicy = Kokkos::TeamPolicy<E>;
  using Team = typename TeamPolicy::member_type;
  using SharedSpace = typename E::scratch_memory_space;
  using UnmanagedTraits = Kokkos::MemoryTraits<Kokkos::Unmanaged>;
  using SharedView = Kokkos::View<T*,SharedSpace,UnmanagedTraits>;

  Kokkos::DualView<std::size_t,E> N_;

  F() : N_("N") 
  {
    N_.h_view() = TeamPolicy::scratch_size_max(0) / sizeof(T);
    syncToDevice(N_);
    printf("N = %10lu\n",N_.h_view());
  };

  KOKKOS_INLINE_FUNCTION 
  void operator()(const Team& team) const {
    SharedView a(team.team_scratch(0),N_.d_view());
    Kokkos::parallel_for(Kokkos::TeamThreadRange(team,a.extent(0)),
      [&](const int& i){
        a(i) = i;
      });
  }

  std::size_t shmemSize() const {
    return SharedView::shmem_size(N_.h_view());
  }

};


  Kokkos::initialize(argc,argv); {

  const std::size_t nTeams = 10;
  using Functor = F<int,ExecutionSpace>;
  using Policy = Functor::TeamPolicy;
  using Team = Functor::Team;
  Functor f{};
  printf("shmemSize = %10lu\n",f.shmemSize());
  const std::size_t teamSize = Policy(nTeams,Kokkos::AUTO).
    team_size_max(f,Kokkos::ParallelForTag{});
  auto policy = Policy(nTeams,teamSize).
    set_scratch_size(0,Kokkos::PerTeam(f.shmemSize())); 
  Kokkos::parallel_for(policy,f); 

  } Kokkos::finalize();
}
```

crtrott:
This takes a bit better into account the internal shared memory teams are using by default. It does not take into account static shared memory which teams may use for certain types of team level reductions. The amount of shared memory there depends on the `value_type` and thus can't be taken into account without handing over the functor. 

It does otherwise fix #1811

### DIFF
--- a/core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
@@ -203,16 +203,16 @@ class TeamPolicyInternal<Kokkos::Cuda, Properties...>
     // shared ID. Furthermore, they use additional scratch memory in some
     // reduction scenarios, which depend on the size of the value_type and is
     // NOT captured here.
-    constexpr size_t cuda_team_max_reserved_shared_mem =
-        (1024 + 2) * sizeof(double) + sizeof(int64_t);
+    constexpr size_t max_possible_team_size = 1024;
+    constexpr size_t max_reserved_shared_mem_per_team =
+        (max_possible_team_size + 2) * sizeof(double) + sizeof(int64_t);
+    // arbitrarily setting level 1 scratch limit to 20MB, for a
+    // Volta V100 that would give us about 3.2GB for 2 teams per SM
+    constexpr size_t max_l1_scratch_size = 20 * 1024 * 1024;
 
     size_t max_shmem = Cuda().cuda_device_prop().sharedMemPerBlock;
-    return (level == 0
-                ? max_shmem - cuda_team_max_reserved_shared_mem
-                :
-                // arbitrarily setting level 1 scratch limit to 20MB, for a
-                // Volta V100 that would give us about 3.2GB for 2 teams per SM
-                20 * 1024 * 1024);
+    return (level == 0 ? max_shmem - max_reserved_shared_mem_per_team
+                       : max_l1_scratch_size);
   }
 
   //----------------------------------------
--- a/core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
@@ -47,14 +47,6 @@ extern bool show_warnings() noexcept;
 
 namespace Impl {
 
-// Cuda Teams use (team_size + 2)*sizeof(double) shared memory for team
-// reductions. They also use one int64_t in static shared memory for a shared
-// ID. Furthermore, they use additional scratch memory in some reduction
-// scenarios, which depend on the size of the value_type and is NOT captured
-// here.
-constexpr size_t cuda_team_max_reserved_shared_mem =
-    (1024 + 2) * sizeof(double) + sizeof(int64_t);
-
 template <class... Properties>
 class TeamPolicyInternal<Kokkos::Cuda, Properties...>
     : public PolicyTraits<Properties...> {
@@ -206,6 +198,14 @@ class TeamPolicyInternal<Kokkos::Cuda, Properties...>
   }
 
   inline static int scratch_size_max(int level) {
+    // Cuda Teams use (team_size + 2)*sizeof(double) shared memory for team
+    // reductions. They also use one int64_t in static shared memory for a
+    // shared ID. Furthermore, they use additional scratch memory in some
+    // reduction scenarios, which depend on the size of the value_type and is
+    // NOT captured here.
+    constexpr size_t cuda_team_max_reserved_shared_mem =
+        (1024 + 2) * sizeof(double) + sizeof(int64_t);
+
     size_t max_shmem = Cuda().cuda_device_prop().sharedMemPerBlock;
     return (level == 0
                 ? max_shmem - cuda_team_max_reserved_shared_mem
--- a/core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
@@ -48,9 +48,10 @@ extern bool show_warnings() noexcept;
 namespace Impl {
 
 // Cuda Teams use (team_size + 2)*sizeof(double) shared memory for team
-// reductions they also use one int64_t static shared memory for a shared ID
-// Furthermore, they use additional scratch memory in some reduction scenarios,
-// which depend on the size of the value_type and is NOT captured here.
+// reductions. They also use one int64_t in static shared memory for a shared
+// ID. Furthermore, they use additional scratch memory in some reduction
+// scenarios, which depend on the size of the value_type and is NOT captured
+// here.
 constexpr size_t cuda_team_max_reserved_shared_mem =
     (1024 + 2) * sizeof(double) + sizeof(int64_t);
 
--- a/core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
@@ -47,6 +47,13 @@ extern bool show_warnings() noexcept;
 
 namespace Impl {
 
+// Cuda Teams use (team_size + 2)*sizeof(double) shared memory for team
+// reductions they also use one int64_t static shared memory for a shared ID
+// Furthermore, they use additional scratch memory in some reduction scenarios,
+// which depend on the size of the value_type and is NOT captured here.
+constexpr size_t cuda_team_max_reserved_shared_mem =
+    (1024 + 2) * sizeof(double) + sizeof(int64_t);
+
 template <class... Properties>
 class TeamPolicyInternal<Kokkos::Cuda, Properties...>
     : public PolicyTraits<Properties...> {
@@ -198,12 +205,13 @@ class TeamPolicyInternal<Kokkos::Cuda, Properties...>
   }
 
   inline static int scratch_size_max(int level) {
-    return (
-        level == 0 ? 1024 * 40 :  // 48kB is the max for CUDA, but we need some
-                                  // for team_member.reduce etc.
-            20 * 1024 *
-                1024);  // arbitrarily setting this to 20MB, for a Volta V100
-                        // that would give us about 3.2GB for 2 teams per SM
+    size_t max_shmem = Cuda().cuda_device_prop().sharedMemPerBlock;
+    return (level == 0
+                ? max_shmem - cuda_team_max_reserved_shared_mem
+                :
+                // arbitrarily setting level 1 scratch limit to 20MB, for a
+                // Volta V100 that would give us about 3.2GB for 2 teams per SM
+                20 * 1024 * 1024);
   }
 
   //----------------------------------------
--- a/core/src/HIP/Kokkos_HIP_Parallel_Team.hpp
+++ b/core/src/HIP/Kokkos_HIP_Parallel_Team.hpp
@@ -163,16 +163,17 @@ class TeamPolicyInternal<HIP, Properties...>
     // reductions. They also use one int64_t in static shared memory for a
     // shared ID. Furthermore, they use additional scratch memory in some
     // reduction scenarios, which depend on the size of the value_type and is
-    // NOT captured here.
-    constexpr size_t hip_team_max_reserved_shared_mem =
-        (1024 + 2) * sizeof(double) + sizeof(int64_t);
+    // NOT captured here
+    constexpr size_t max_possible_team_size = 1024;
+    constexpr size_t max_reserved_shared_mem_per_team =
+        (max_possible_team_size + 2) * sizeof(double) + sizeof(int64_t);
+    // arbitrarily setting level 1 scratch limit to 20MB, for a
+    // MI250 that would give us about 4.4GB for 2 teams per CU
+    constexpr size_t max_l1_scratch_size = 20 * 1024 * 1024;
 
     size_t max_shmem = HIP().hip_device_prop().sharedMemPerBlock;
-    return (
-        level == 0 ? max_shmem - hip_team_max_reserved_shared_mem :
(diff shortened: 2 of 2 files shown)


core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
```diff
@@ -47,6 +47,12 @@ extern bool show_warnings() noexcept;
 
 namespace Impl {
 
+// Cuda Teams use (team_size + 2)*sizeof(double) shared memory for team
+// reductions they also use one in64_t static shared memory for a shared ID
```
PhilMiller, line 51: int64_t

core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
```diff
@@ -47,6 +47,12 @@ extern bool show_warnings() noexcept;
...
+// Furthermore, they use additional scratch memory in some reduction scenarios,
+// which depend on the size of the value_type and is NOT captured here.
+constexpr size_t cuda_team_max_reserved_shared_mem = (1024 + 2) * 8 + 8;
```
PhilMiller, line 54: Why not use the `sizeof` expressions shown in the comments?

PhilMiller:
This maybe leaves the question of whether we should provide a version that takes the functor for the case where it matters, and make sure the documentation points users to that.

core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
crtrott: Yeah I guess I could do that.

core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
```diff
@@ -47,6 +47,13 @@ extern bool show_warnings() noexcept;
 
 namespace Impl {
 
+// Cuda Teams use (team_size + 2)*sizeof(double) shared memory for team
+// reductions they also use one int64_t static shared memory for a shared ID
```
PhilMiller, line 51:
```suggestion
// reductions. They also use one int64_t in static shared memory for a shared ID.
```

core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
```diff
@@ -47,6 +47,13 @@ extern bool show_warnings() noexcept;
...
+// Furthermore, they use additional scratch memory in some reduction scenarios,
```
PhilMiller, line 52: Scans as well as reductions?

core/src/HIP/Kokkos_HIP_Parallel_Team.hpp
```diff
@@ -30,6 +30,13 @@
 namespace Kokkos {
 namespace Impl {
 
+// HIP Teams use (team_size + 2)*sizeof(double) shared memory for team
+// reductions they also use one int64_t static shared memory for a shared ID
```
PhilMiller, line 34:
```suggestion
// reductions. They also use one int64_t in static shared memory for a shared ID.
```

core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
crtrott: yeah generally when I say "reduction scenarios" I imply scans too, since they are basically doing reductions too.

core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
```diff
@@ -47,6 +47,14 @@ extern bool show_warnings() noexcept;
 
 namespace Impl {
 
+// Cuda Teams use (team_size + 2)*sizeof(double) shared memory for team
+// reductions. They also use one int64_t in static shared memory for a shared
+// ID. Furthermore, they use additional scratch memory in some reduction
+// scenarios, which depend on the size of the value_type and is NOT captured
+// here.
+constexpr size_t cuda_team_max_reserved_shared_mem =
+    (1024 + 2) * sizeof(double) + sizeof(int64_t);
```
dalg24, line 56: Why declare it at namespace scope?   Keep the scope minimal, move it to `TeamPolicyInternal<Cuda>:: scratch_size_max()`

core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
```diff
@@ -198,12 +206,13 @@ class TeamPolicyInternal<Kokkos::Cuda, Properties...>
   }
 
   inline static int scratch_size_max(int level) {
-    return (
-        level == 0 ? 1024 * 40 :  // 48kB is the max for CUDA, but we need some
-                                  // for team_member.reduce etc.
-            20 * 1024 *
-                1024);  // arbitrarily setting this to 20MB, for a Volta V100
-                        // that would give us about 3.2GB for 2 teams per SM
+    size_t max_shmem = Cuda().cuda_device_prop().sharedMemPerBlock;
```
dalg24, line 209: Presumably you have access to an exec space from the policy so don't use a default-constructed one.

core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
```diff
@@ -198,12 +206,13 @@ class TeamPolicyInternal<Kokkos::Cuda, Properties...>
...
+    return (level == 0
+                ? max_shmem - cuda_team_max_reserved_shared_mem
+                :
+                // arbitrarily setting level 1 scratch limit to 20MB, for a
+                // Volta V100 that would give us about 3.2GB for 2 teams per SM
+                20 * 1024 * 1024);
```
dalg24, line 215: Consider turning this guy into a local variable too.

PhilMiller:
(Your HIP commit message got truncated)

dalg24 (changes requested):

core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
dalg24: Ah it is a static member function :/

core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
```diff
@@ -47,6 +47,14 @@ extern bool show_warnings() noexcept;
 
 namespace Impl {
 
+// Cuda Teams use (team_size + 2)*sizeof(double) shared memory for team
+// reductions. They also use one int64_t in static shared memory for a shared
+// ID. Furthermore, they use additional scratch memory in some reduction
+// scenarios, which depend on the size of the value_type and is NOT captured
+// here.
+constexpr size_t cuda_team_max_reserved_shared_mem =
```
PhilMiller, line 55: Why is there `max` in this name? It's a minimum reservation, isn't it?

crtrott: no, its the max you might get because the actual calculation is: `(team_size+2)*sizeof(double)+sizeof(int64_t)`

crtrott: and max team size if 1024

core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
```diff
@@ -198,12 +198,21 @@ class TeamPolicyInternal<Kokkos::Cuda, Properties...>
...
-        level == 0 ? 1024 * 40 :  // 48kB is the max for CUDA, but we need some
-                                  // for team_member.reduce etc.
-            20 * 1024 *
-                1024);  // arbitrarily setting this to 20MB, for a Volta V100
-                        // that would give us about 3.2GB for 2 teams per SM
+    // Cuda Teams use (team_size + 2)*sizeof(double) shared memory for team
+    // reductions. They also use one int64_t in static shared memory for a
+    // shared ID. Furthermore, they use additional scratch memory in some
+    // reduction scenarios, which depend on the size of the value_type and is
+    // NOT captured here.
```
PhilMiller, line 205: Again, please add something in the spirit of this comment as a comment on `Kokkos::TeamPolicy::scratch_size_max` so callers of the public API have an indication that other things may eat into this value.

core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
PhilMiller: Ah, ok. Is there a `int max_team_size = 1024` somewhere? If not, could you make one, and use that by name, rather than the magic constant that I had to ask about?

PhilMiller:
This may also address #3498

core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
crtrott: there isn't a constexpr one, and technically that might not be correct.  In some ways that number is more prone to misinterpretation or misuse than the `reserved_shared_mem`. Essentially, in most cases the max team size will be smaller than that due to stuff like register constraints. Misuse or misunderstanding what a "max_team_size" variable means is very likely to be much more harmful than misunderstanding what cuda_team_max_reserved_shared_mem means. The latter would simply mean you are underestimating a bit how much shared memory you could theoretically use per block, the max_team_size misuse could mean all kinds of problems since it will almost always be wrong.

core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
crtrott: Note that the (admittedly stupid) way this works: this IS the public function, because TeamPolicy inherits from this guy. Turns out there is no scratch_size_max inside the Kokkos_ExecPolicy.hpp ...

crtrott: Oh and the prototype for TeamPolicyInternal in that header is unused, because everything real happens in partial specializations ...

crtrott: that said: working on a PR for the documentation

core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
crtrott: not sure that its worth it.

core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
PhilMiller: Even if it were just a variable local to this function, it would be an improvement. Perhaps `int max_possible_team_size`? `int theoretical_max_team_size`? `int team_size_upper_bound`?

core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
crtrott: https://github.com/kokkos/kokkos-core-wiki/pull/234

core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
crtrott: done

core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
crtrott: anyway done

PhilMiller (approved):

core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
```diff
@@ -198,12 +198,21 @@ class TeamPolicyInternal<Kokkos::Cuda, Properties...>
   }
 
   inline static int scratch_size_max(int level) {
-    return (
-        level == 0 ? 1024 * 40 :  // 48kB is the max for CUDA, but we need some
-                                  // for team_member.reduce etc.
-            20 * 1024 *
-                1024);  // arbitrarily setting this to 20MB, for a Volta V100
-                        // that would give us about 3.2GB for 2 teams per SM
+    // Cuda Teams use (team_size + 2)*sizeof(double) shared memory for team
```
dalg24, line 201: I assume the `sizeof(double)` here is our saying "let the value type for the reduction be double"?

crtrott: No, it actually doesn't I had there an 8 before because we actually say sizeof(double) in the code, but if you happen to do reductions with value_types larger than that they are done in stages.

crtrott: But I think its not a great idea to write the entire explanation of how our reductions work in the description of one variable.

dalg24: OK

Rombur (approved):

dalg24 (approved):

Issue #5719, opened 2022-12-22
Compilation error when using hip_parallel_launch_global_memory


Rombur:
Reported in slack. Fix coming soon.

PR #5720, opened 2022-12-22
[HIP] Fix GlobalMemory launching mechanism

Rombur:
Code is easier to review if you hide white spaces.

Related to #5719

### DIFF
--- a/core/src/HIP/Kokkos_HIP_Instance.cpp
+++ b/core/src/HIP/Kokkos_HIP_Instance.cpp
@@ -236,6 +236,28 @@ Kokkos::HIP::size_type *HIPInternal::scratch_flags(const std::size_t size) {
   return m_scratchFlags;
 }
 
+Kokkos::HIP::size_type *HIPInternal::scratch_functor(
+    const std::size_t size) const {
+  if (verify_is_initialized("scratch_functor") && m_scratchFunctorSize < size) {
+    m_scratchFunctorSize = size;
+
+    using Record = Kokkos::Impl::SharedAllocationRecord<Kokkos::HIPSpace, void>;
+
+    if (m_scratchFunctor)
+      Record::decrement(Record::get_record(m_scratchFunctor));
+
+    Record *const r =
+        Record::allocate(Kokkos::HIPSpace(), "Kokkos::InternalScratchFunctor",
+                         m_scratchFunctorSize);
+
+    Record::increment(r);
+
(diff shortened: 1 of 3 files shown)


core/src/HIP/Kokkos_HIP_Instance.hpp
```diff
@@ -133,6 +135,7 @@ class HIPInternal {
   // Resizing of reduction related scratch spaces
   size_type *scratch_space(const std::size_t size);
   size_type *scratch_flags(const std::size_t size);
+  size_type *scratch_functor(const std::size_t size) const;
```
dalg24, line 138: Why pretend it is const if all other aren't marked as const and it forces you to make member data mutable?

core/src/HIP/Kokkos_HIP_Instance.cpp
```diff
@@ -296,15 +318,18 @@ void HIPInternal::finalize() {
     RecordHIP::decrement(RecordHIP::get_record(m_scratchFlags));
     RecordHIP::decrement(RecordHIP::get_record(m_scratchSpace));
 
-    for (int i = 0; i < m_n_team_scratch; ++i) {
-      if (m_team_scratch_current_size[i] > 0)
-        Kokkos::kokkos_free<Kokkos::HIPSpace>(m_team_scratch_ptr[i]);
-    }
+    if (m_scratchFunctorSize > 0)
+      RecordHIP::decrement(RecordHIP::get_record(m_scratchFunctor));
+  }
```
dalg24, line 323: Did you mean to nest in the `if (nullptr != m_scratchSpace || nullptr != m_scratchFlags)` block?

core/src/HIP/Kokkos_HIP_Instance.hpp
dalg24: Now I see that is also how CUDA does it...

Rombur: The function needs to be `const` because it is called from a const object.

crtrott (approved):
The lock for not overwriting this happens elsewhere right?

PhilMiller:
HPX and CUDA failures are irrelevant.

dalg24:
> The lock for not overwriting this happens elsewhere right?

I think you don't need a lock for this one because the driver is being copied to an instance specific allocation.

dalg24 (approved):

PR #5721, opened 2022-12-23
[3.7.02] Fix static init order issue in InitalizationSettings

crtrott:
Reproducer on Power9 with GCC 8.3.1 and CUDA 11.2

```c++
#include <Kokkos_Core.hpp>
#include <cstdio>
#include <typeinfo>

static Kokkos::InitializationSettings args;

  Kokkos::finalize();
}
```

Crashed when initializing the string members.

### DIFF
--- a/core/src/impl/Kokkos_Core.cpp
+++ b/core/src/impl/Kokkos_Core.cpp
@@ -1165,6 +1165,5 @@ void _kokkos_pgi_compiler_bug_workaround() {}
 #endif
 }  // namespace Kokkos
 
-Kokkos::Impl::InitializationSettingsHelper<std::string>::storage_type const
-    Kokkos::Impl::InitializationSettingsHelper<std::string>::unspecified =
-        "some string we don't expect user would ever provide";
+constexpr char
+    Kokkos::Impl::InitializationSettingsHelper<std::string>::unspecified[];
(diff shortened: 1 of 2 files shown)


dalg24 (approved):

core/src/impl/Kokkos_Core.cpp
```diff
@@ -1165,6 +1165,5 @@ void _kokkos_pgi_compiler_bug_workaround() {}
 #endif
 }  // namespace Kokkos
 
-Kokkos::Impl::InitializationSettingsHelper<std::string>::storage_type const
-    Kokkos::Impl::InitializationSettingsHelper<std::string>::unspecified =
-        "some string we don't expect user would ever provide";
+constexpr char
+    Kokkos::Impl::InitializationSettingsHelper<std::string>::unspecified[];
```
PhilMiller, line 1169: Does this need to be declared here at all now?

PhilMiller (approved):

core/src/impl/Kokkos_Core.cpp
dalg24: I think so.  This is C++14.

dalg24:
Ignoring NVHPC toolchain issues

PR #5618, opened 2022-11-04
OpenACC ParallelReduce with MDRange

seyonglee:
Initial implementation of the OpenACC parallel_reduce() with MDRange
    - Updated the OpenACC parallel_reduce() with Range to be compatible with recent changes

PR #5722, opened 2022-12-24
OpenACC initial implementation of `parallel_reduce` for `MDRange`

dalg24:
Intended to supersede #5618 
It is more a refactor that anything else.  It does not address https://github.com/kokkos/kokkos/pull/5618#discussion_r1018180681 and suffers the same general issues with the implementation in #5618.  I attempted to enable the incremental test 14 and the one that is reducing into a value on the device gives a memory access violation (the other one passes).  The `parallel_reduce` for `Range` has the same bug so I suppose we could tackle that elsewhere.

### DIFF
--- a/core/src/OpenACC/Kokkos_OpenACC_ParallelReduce_MDRange.hpp
+++ b/core/src/OpenACC/Kokkos_OpenACC_ParallelReduce_MDRange.hpp
@@ -1,0 +1,458 @@
+//@HEADER
+// ************************************************************************
+//
+//                        Kokkos v. 4.0
+//       Copyright (2022) National Technology & Engineering
+//               Solutions of Sandia, LLC (NTESS).
+//
+// Under the terms of Contract DE-NA0003525 with NTESS,
+// the U.S. Government retains certain rights in this software.
+//
+// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
+#ifndef KOKKOS_OPENACC_PARALLEL_REDUCE_MDRANGE_HPP
+#define KOKKOS_OPENACC_PARALLEL_REDUCE_MDRANGE_HPP
+
+#include <OpenACC/Kokkos_OpenACC.hpp>
+#include <OpenACC/Kokkos_OpenACC_Macros.hpp>
+#include <OpenACC/Kokkos_OpenACC_FunctorAdapter.hpp>
+#include <OpenACC/Kokkos_OpenACC_MDRangePolicy.hpp>
+#include <Kokkos_Parallel.hpp>
+
+namespace Kokkos::Experimental::Impl {
+
+// primary template: catch-all non-implemented custom reducers
+template <class Functor, class Reducer, class Policy,
+          bool = std::is_arithmetic_v<typename Reducer::value_type>>
+struct OpenACCParallelReduceMDRangeHelper {
+  OpenACCParallelReduceMDRangeHelper(Functor const&, Reducer const&,
+                                     Policy const&) {
+    static_assert(!Kokkos::Impl::always_true<Functor>::value,
+                  "not implemented");
+  }
+};
+}  // namespace Kokkos::Experimental::Impl
+
+template <class Functor, class ReducerType, class... Traits>
+class Kokkos::Impl::ParallelReduce<Functor, Kokkos::MDRangePolicy<Traits...>,
+                                   ReducerType, Kokkos::Experimental::OpenACC> {
+  using Policy = MDRangePolicy<Traits...>;
+
+  using ReducerConditional =
+      if_c<std::is_same_v<InvalidType, ReducerType>, Functor, ReducerType>;
+  using ReducerTypeFwd = typename ReducerConditional::type;
+  using Analysis =
+      FunctorAnalysis<FunctorPatternInterface::REDUCE, Policy, ReducerTypeFwd>;
+
+  using Pointer   = typename Analysis::pointer_type;
+  using ValueType = typename Analysis::value_type;
+
+  Functor m_functor;
+  Policy m_policy;
+  ReducerType m_reducer;
+  Pointer m_result_ptr;
+
+ public:
+  ParallelReduce(Functor const& functor, Policy const& policy,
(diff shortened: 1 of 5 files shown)


seyonglee:
I verified that this PR is semantically equivalent to [PR5618](#5618) regarding `parallel_reduce` construct support.
As @dalg24  said, the current OpenACC backend implementation supports `parallel_reduce` constructs only if the reduction variable is scalar allocated on the host. I will create a separate PR to support scalar reduction variables on the device too, when this PR is merged.

seyonglee (approved):

core/src/OpenACC/Kokkos_OpenACC_ParallelReduce_MDRange.hpp
```diff
@@ -0,0 +1,458 @@
...
+      }                                                                       \
+    }                                                                         \
+    aval = val;                                                               \
+  }                                                                           \
+                                                                              \
+  template <class ValueType, class Functor>                                   \
+  void OpenACCParallelReduce##REDUCER(OpenACCIterateLeft, ValueType& aval,    \
+                                      Functor const& afunctor,                \
+                                      OpenACCMDRangeBegin<3> const& begin,    \
+                                      OpenACCMDRangeEnd<3> const& end,        \
```
rgayatri23, line 155: Can we add the rank as a template parameter to the function call? 
`template <int rank, class ValueType, class Functor>`

dalg24: How would that help?  You still have to explicitly write unpack write the loops.

rgayatri23: Yes, we will have to write the loops explicitly but I was thinking of something like this 
```
template <int Rank, class ValueType, class Functor>
 void std::enable_if_t<Rank == 3> OpenACCParallelReduce(.....) {...}
```
Essentially something to reduce the number of parameters in the function call. But this is not going to help I guess.

rgayatri23:
Is there a plan for tiling too if the user requests it?

core/src/OpenACC/Kokkos_OpenACC_ParallelReduce_MDRange.hpp
dalg24: It does not reduce the numbers of parameters

seyonglee:
> Is there a plan for tiling too if the user requests it?

It is possible to implement the tiling support for `parallel_reduce` construct too, as we did for `parallel_for` construct.
However, the tiling support for `parallel_for` construct is currently disabled since OpenACC requires the arguments of the tile clause to be compile-time constant (and `parallel_reduce` construct will have the same issue).

core/src/OpenACC/Kokkos_OpenACC_ParallelReduce_MDRange.hpp
```diff
@@ -0,0 +1,458 @@
...
+// primary template: catch-all non-implemented custom reducers
+template <class Functor, class Reducer, class Policy,
+          bool = std::is_arithmetic_v<typename Reducer::value_type>>
+struct OpenACCParallelReduceMDRangeHelper {
+  OpenACCParallelReduceMDRangeHelper(Functor const&, Reducer const&,
+                                     Policy const&) {
+    static_assert(!Kokkos::Impl::always_true<Functor>::value,
+                  "not implemented");
+  }
+};
```
masterleinad, line 37: Add a FIXME?

core/src/OpenACC/Kokkos_OpenACC_ParallelReduce_MDRange.hpp
```diff
@@ -0,0 +1,458 @@
...
+        m_result_ptr(result.data()) {}
+
+  void execute() const {
+    static_assert(1 < Policy::rank && Policy::rank < 7);
+    static_assert(Policy::inner_direction == Iterate::Left ||
+                  Policy::inner_direction == Iterate::Right);
+    constexpr int rank = Policy::rank;
+    for (int i = 0; i < rank; ++i) {
+      if (m_policy.m_lower[i] >= m_policy.m_upper[i]) {
+        return;
```
masterleinad, line 83: This should still call `init` and `final` in the end, right? Maybe at a `FIXME` comment.

core/src/OpenACC/Kokkos_OpenACC_ParallelReduce_MDRange.hpp
```diff
@@ -0,0 +1,458 @@
...
+        &ReducerConditional::select(m_functor, m_reducer));
+    final_reducer.init(&val);
+
+    Kokkos::Experimental::Impl::OpenACCParallelReduceMDRangeHelper(
+        Kokkos::Experimental::Impl::FunctorAdapter<Functor, Policy>(m_functor),
+        std::conditional_t<is_reducer_v<ReducerType>, ReducerType,
+                           Sum<ValueType>>(val),
+        m_policy);
+
+    *m_result_ptr = val;
```
masterleinad, line 98:
```suggestion
    final_reducer.final(&val);
    *m_result_ptr = val;
```
or add a FIXME.

masterleinad (commented):
Looks good to me but I would appreciate adding some FIXMEs about missing features (and possibly just adding a call to `final`).

core/src/OpenACC/Kokkos_OpenACC_ParallelReduce_MDRange.hpp
dalg24: Considering that we have the same issue with `parallel_reduce` on a regular `RangePolicy`, I would suggest we just open an issue to track this
https://github.com/kokkos/kokkos/blob/459e8811be9a6ccb984c40a8a414b6f0556841ba/core/src/OpenACC/Kokkos_OpenACC_ParallelReduce_Range.hpp#L81-L83

core/src/OpenACC/Kokkos_OpenACC_ParallelReduce_MDRange.hpp
dalg24: Same comment about opening an issue instead 
https://github.com/kokkos/kokkos/blob/459e8811be9a6ccb984c40a8a414b6f0556841ba/core/src/OpenACC/Kokkos_OpenACC_ParallelReduce_Range.hpp#L96

rgayatri23 (approved):

PR #5726, opened 2023-01-03
Refactor OpenMPTarget backend

rgayatri23:
This PR attempts to re-organize the OpenMPTarget backend code by creating different files for each construct. There is no new code being added but just moving the existing implementations into appropriate construct based file names.


Rombur:
Is there a reason why you decided  to split the files in a different way than what's done in `Serial`, `HIP`, and `Cuda`?

rgayatri23:
> Is there a reason why you decided to split the files in a different way than what's done in `Serial`, `HIP`, and `Cuda`?

The current implementation follows the `Cuda` way, the proposed way in this PR follows the style followed by the `OpenACC` backend. I found it more intuitive since each construct gets its own implementation file.

core/unit_test/CMakeLists.txt
```diff
@@ -351,6 +351,13 @@ if(Kokkos_ENABLE_OPENMPTARGET)
     ${CMAKE_CURRENT_BINARY_DIR}/openmptarget/TestOpenMPTarget_Other.cpp
     ${CMAKE_CURRENT_BINARY_DIR}/openmptarget/TestOpenMPTarget_TeamReductionScan.cpp
     ${CMAKE_CURRENT_BINARY_DIR}/openmptarget/TestOpenMPTarget_WorkGraph.cpp
+
+#FIXME_OPENMPTARGET - These unit tests lead to ICE with llvm-16
```
Rombur, line 355: Is this due to the refactor

scripts/diff_files
```diff
@@ -1,2 +1,4 @@
 
 core/src/Cuda/Kokkos_Cuda_Atomic_Intrinsics.hpp
+core/src/OpenMPTarget/Kokkos_OpenMPTarget_ParallelFor_Range.hpp
+core/src/OpenMPTarget/Kokkos_OpenMPTarget_ParallelFor_Team.hpp
```
Rombur, line 4: What does this do?

core/unit_test/CMakeLists.txt
rgayatri23: No this is not due to the refactor. This is because llvm-16 (against which I was testing) crashes on these unit tests. I did not know how to specialize the loop for Clang/16 in CMAKE . llvm-15 which is used in our CI passes these tests.

scripts/diff_files
rgayatri23: I am not sure why that got added. I will remove this file.

Rombur (approved):

scripts/diff_files
```diff
@@ -1,2 +0,0 @@
-
-core/src/Cuda/Kokkos_Cuda_Atomic_Intrinsics.hpp
```
dalg24, line 2: Why did you touch that file

core/unit_test/CMakeLists.txt
dalg24: You should tackle this else where, not as part of a larger refactor.
The description of the PR states that you are just moving things around.

dalg24 (changes requested):

core/unit_test/CMakeLists.txt
rgayatri23: Rolled back the changes.

scripts/diff_files
dalg24: Did you mean to change?

rgayatri23: No. I reverted back the change.

dalg24 (commented):
There is still a conflict with `diff_files`
Please move `Kokkos_OpenMPTarget[Space].hpp` into the `OpenMPTarget` directory (adjust includes as necessary)

core/unit_test/CMakeLists.txt
vlkale: > > Is there a reason why you decided to split the files in a different way than what's done in `Serial`, `HIP`, and `Cuda`?
> 
> The current implementation follows the `Cuda` way, the proposed way in this PR follows the style followed by the `OpenACC` backend. I found it more intuitive since each construct gets its own implementation file.

To help have a  software/code organization that's maintainable and that can be associated with other backends:  I think  each (OpenMP)  _directive_ gets its own implementation file, right? Not each OpenMP _construct_? 

I don't know the OpenACC organization  in Kokkos (and this would be a very good comparison point by the way given recent developments I have seen with it) but citing the [OpenMP 5.2 Specification](https://www.openmp.org/wp-content/uploads/OpenMPRefCard-5-2-web.pdf
) as guidance: 'An OpenMP executable directive applies to the succeeding structured block or an OpenMP construct. Each directive starts with #pragma omp. ... A construct consists of a directive and, often, additional base language code. C/C++ C directives are formed exclusively with pragmas. C++ .'  


(Also, this follows standard Computer Science terminology, a construct can be any syntax that provides functionality like a conditional or loop. So, 'if' is a lexical token while 'if(true){}' would be a construct. Also, the '{...}' is a construct while '{' is a lexical token. )

vlkale:
There  is still a conflict to be resolved/removed in this branch that are in 'scripts/diff_files'. 

. @rgayatri23  

This seems to be the blocker here, and consideration of the organization of code files. It LGTM overall having reviewed it. 

We may want to just note here - or perhaps keep a separate running Github issue - that we should keep track of any new  directives  for OpenMP offload in future OpenMP Specification releases. Kokkos will support those that are actually implemented across different vendor OpenMP impleemntations, not just ones part of the OpenMP Spec but not available or fully functional in major vendor OpenMP implementations or LLVM OpenMP implementations. 

(And note that it hasn't been properly tested on Aurora yet - and more refactoring may be needed based on this testing. )

dalg24 (changes requested):
Move Kokkos_OpenMPTarget.hpp header too

masterleinad (commented):
It seems that `core/src/OpenMPTarget/Kokkos_OpenMPTarget_Parallel.hpp` both contains the team member implementation (`OpenMPTargetExecTeamMember`) and the `OpenMPTargetExec` class declaration. Would you mind moving the former into an `OpenMPTarget_Team.hpp` header and the latter into `OpenMPTarget_Exec.hpp` (or the `*_Instance` header)?

dalg24 (commented):
Although I agree with https://github.com/kokkos/kokkos/pull/5726#pullrequestreview-1279225024
I suggest we merge as is (provided that the CI passes) and fix in a follow-up PR.

dalg24 (approved):

masterleinad (approved):

PR #5727, opened 2023-01-04
Added multiple reducers support for team-level parallel reduce

### LINKED ISSUE #4510: Multiple reductions for team-level loops
I'd like to be able to use multiple reductions for parallel_reduce calls on TeamThreadRange/TeamVectorRange/ThreadVectorRange levels.

ldh4:
Resolves #4510 

Added an interface to allow parallel_reduce with TeamThreadRange, ThreadVectorRange or TeamVectorRange policy to be called with more than one reducer.

### DIFF
--- a/core/src/impl/Kokkos_Combined_Reducer.hpp
+++ b/core/src/impl/Kokkos_Combined_Reducer.hpp
@@ -515,13 +515,11 @@ KOKKOS_INLINE_FUNCTION void parallel_reduce_combined_reducers_impl(
       combined_value, returnType1, returnType2, returnTypes...);
 
   parallel_reduce(boundaries, combined_functor, combined_reducer);
-  memory_fence();
 
   combined_reducer.write_value_back_to_original_references_on_device(
       combined_value, Impl::_make_reducer_from_arg<mem_space_type>(returnType1),
       Impl::_make_reducer_from_arg<mem_space_type>(returnType2),
       Impl::_make_reducer_from_arg<mem_space_type>(returnTypes)...);
-  memory_fence();
 }
 
 }  // end namespace Impl
--- a/core/src/impl/Kokkos_Combined_Reducer.hpp
+++ b/core/src/impl/Kokkos_Combined_Reducer.hpp
@@ -210,6 +210,8 @@ struct CombinedReducerImpl<std::integer_sequence<size_t, Idxs...>, Space,
      ...);
   }
 
+  KOKKOS_FUNCTION auto& reference() const { return *m_value_view.data(); }
+
   // TODO figure out if we also need to call through to final
 
   KOKKOS_FUNCTION
@@ -245,6 +247,22 @@ struct CombinedReducerImpl<std::integer_sequence<size_t, Idxs...>, Space,
          exec_space, reducers_that_reference_original_values.view(),
          value.template get<Idxs, typename Reducers::value_type>()),
 
+     ...);
+  }
+
+  template <int Idx, class View>
+  KOKKOS_FUNCTION void write_one_value_back_on_device(
+      View const& inputView, typename View::const_value_type& value) noexcept {
+    *inputView.data() = value;
+  }
+
+  template <typename... CombinedReducers>
+  KOKKOS_FUNCTION void write_value_back_to_original_references_on_device(
+      value_type const& value,
+      CombinedReducers const&... reducers_that_reference_original_values) noexcept {
+    (write_one_value_back_on_device<Idxs>(
+         reducers_that_reference_original_values.view(),
+         value.template get<Idxs, typename CombinedReducers::value_type>()),
      ...);
   }
 };
@@ -463,12 +481,12 @@ KOKKOS_INLINE_FUNCTION constexpr auto make_combined_reducer(
   return reducer_type(value,
                       _reducer_from_arg_t<Space, ReferencesOrViewsOrReducers>{
                           (ReferencesOrViewsOrReducers &&) args}...);
-  //----------------------------------------
+  //---------------------------------------
 }
 
-template <class Functor, class Space, class... ReferencesOrViewsOrReducers>
+template <class Space, class Functor, class... ReferencesOrViewsOrReducers>
 KOKKOS_INLINE_FUNCTION constexpr auto make_wrapped_combined_functor(
-    Functor const& functor, Space, ReferencesOrViewsOrReducers&&...) {
+    Functor const& functor, ReferencesOrViewsOrReducers&&...) {
   //----------------------------------------
   return CombinedReductionFunctorWrapper<
       Functor, Space,
@@ -478,6 +496,34 @@ KOKKOS_INLINE_FUNCTION constexpr auto make_wrapped_combined_functor(
 
 template <typename FunctorType>
 using functor_has_value_t = typename FunctorType::value_type;
+
+template <typename MemberType, typename BoundaryStructType, typename Functor,
+          typename ReturnType1, typename ReturnType2, typename... ReturnTypes>
+KOKKOS_INLINE_FUNCTION void parallel_reduce_combined_reducers_impl(
+    BoundaryStructType const& boundaries, Functor const& functor,
+    ReturnType1&& returnType1, ReturnType2&& returnType2,
+    ReturnTypes&&... returnTypes) noexcept {
+  using mem_space_type = typename MemberType::scratch_memory_space;
+
+  auto combined_value = Impl::make_combined_reducer_value<mem_space_type>(
+      returnType1, returnType2, returnTypes...);
+
+  auto combined_functor = Impl::make_wrapped_combined_functor<mem_space_type>(
+      functor, returnType1, returnType2, returnTypes...);
+
+  auto combined_reducer = Impl::make_combined_reducer<mem_space_type>(
+      combined_value, returnType1, returnType2, returnTypes...);
+
+  parallel_reduce(boundaries, combined_functor, combined_reducer);
+  memory_fence();
+
+  combined_reducer.write_value_back_to_original_references_on_device(
+      combined_value, Impl::_make_reducer_from_arg<mem_space_type>(returnType1),
+      Impl::_make_reducer_from_arg<mem_space_type>(returnType2),
+      Impl::_make_reducer_from_arg<mem_space_type>(returnTypes)...);
+  memory_fence();
+}
+
 }  // end namespace Impl
 
 //==============================================================================
@@ -509,8 +555,8 @@ auto parallel_reduce(std::string const& label, PolicyType const& policy,
   auto combined_reducer = Impl::make_combined_reducer<space_type>(
       value, returnType1, returnType2, returnTypes...);
 
-  auto combined_functor = Impl::make_wrapped_combined_functor(
-      functor, space_type{}, returnType1, returnType2, returnTypes...);
+  auto combined_functor = Impl::make_wrapped_combined_functor<space_type>(
+      functor, returnType1, returnType2, returnTypes...);
 
   using combined_functor_type = decltype(combined_functor);
   static_assert(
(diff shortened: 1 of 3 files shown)


ldh4:
Retest this please.

ldh4:
Retest this please.

core/src/impl/Kokkos_Combined_Reducer.hpp
```diff
@@ -478,6 +496,34 @@ KOKKOS_INLINE_FUNCTION constexpr auto make_wrapped_combined_functor(
...
+      returnType1, returnType2, returnTypes...);
+
+  auto combined_functor = Impl::make_wrapped_combined_functor<mem_space_type>(
+      functor, returnType1, returnType2, returnTypes...);
+
+  auto combined_reducer = Impl::make_combined_reducer<mem_space_type>(
+      combined_value, returnType1, returnType2, returnTypes...);
+
+  parallel_reduce(boundaries, combined_functor, combined_reducer);
+  memory_fence();
```
masterleinad, line 518: Should this rather be a `team_barrier`?

core/src/impl/Kokkos_Combined_Reducer.hpp
```diff
@@ -478,6 +496,34 @@ KOKKOS_INLINE_FUNCTION constexpr auto make_wrapped_combined_functor(
...
 template <typename FunctorType>
 using functor_has_value_t = typename FunctorType::value_type;
+
+template <typename MemberType, typename BoundaryStructType, typename Functor,
+          typename ReturnType1, typename ReturnType2, typename... ReturnTypes>
+KOKKOS_INLINE_FUNCTION void parallel_reduce_combined_reducers_impl(
+    BoundaryStructType const& boundaries, Functor const& functor,
+    ReturnType1&& returnType1, ReturnType2&& returnType2,
+    ReturnTypes&&... returnTypes) noexcept {
+  using mem_space_type = DefaultExecutionSpace::memory_space;
```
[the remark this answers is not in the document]
masterleinad:
```suggestion
  using mem_space_type = typename MemberType::execution_space::memory_space;
```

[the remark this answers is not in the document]
dalg24: How comes no test failed?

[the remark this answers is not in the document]
masterleinad: For the outer-level `parallel_reduce`, we decide whether we need to call deep_copy or whether we can directly write the reduced value back. AFAICT, this doesn't apply to the nested one since we know that we can directly write the reduced value.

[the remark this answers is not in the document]
ldh4: Precisely what @masterleinad described. `parallel_reduce` called from the host code needs to consider in which memory space the final reduced value resides before a deep_copy back to the host memory, but the team `parallel_reduce` implementation in this PR expects to be called only from the device code with one of the team policies: `TeamThreadRange`, `ThreadVectorRange` or `TeamVectorRange`.
So, as long as the combined reducers use a consistent device memory space internally, reduction on device side shouldn't be affected.

core/unit_test/TestTeamCombinedReducers.hpp
```diff
@@ -0,0 +1,477 @@
...
+  if constexpr (std::is_same_v<TEST_EXECSPACE,
+                               Kokkos::Experimental::OpenMPTarget>)
+    GTEST_SKIP() << "team_reduce with a generic reducer is not implemented for "
+                 << TEST_EXECSPACE::name();
+
+#elif defined(KOKKOS_ENABLE_OPENACC)
+  if constexpr (std::is_same_v<TEST_EXECSPACE, Kokkos::Experimental::OpenACC>)
+    GTEST_SKIP() << "team_reduce with a generic reducer is not implemented for "
+                 << TEST_EXECSPACE::name();
+#endif
```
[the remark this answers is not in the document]
ldh4: Applied in [d65b5f3](https://github.com/kokkos/kokkos/pull/5727/commits/d65b5f3c6f4f20f540b91ade4aaa55a1e8704685)

core/src/impl/Kokkos_Combined_Reducer.hpp
```diff
@@ -247,6 +249,22 @@ struct CombinedReducerImpl<std::integer_sequence<size_t, Idxs...>, Space,
 
      ...);
   }
+
+  template <int Idx, class View>
+  KOKKOS_FUNCTION void write_one_value_back_on_device(
```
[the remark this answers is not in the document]
ldh4: Converted to a `static` function in [a28463e](https://github.com/kokkos/kokkos/pull/5727/commits/a28463eaae89518a32405a9df9e9cb0394bb3d55).
I recall needing to remove `static` from that function at one point, because something was unable to work with a device static function at that time. But that may no longer be the case now. The CI will confirm this.

core/unit_test/TestTeamCombinedReducers.hpp
```diff
@@ -0,0 +1,478 @@
...
```
tcclevenger, line 429: There is a lot of repeated lines with this identical message in each of the 3 test cases. Would it make sense to just have one test case with the 9 function calls, since they also all use the same struct? Just a thought, feel free to ignore.

core/unit_test/TestTeamCombinedReducers.hpp
```diff
@@ -0,0 +1,478 @@
...
+
+  TeamTeamCombinedReducer tester;
+  tester.test_team_thread_range_only_scalars(5);
```
tcclevenger, line 432: Would there be any value in testing also with base case `n=0` for at least one of the cases?

core/src/impl/Kokkos_Combined_Reducer.hpp
```diff
@@ -210,6 +210,8 @@ struct CombinedReducerImpl<std::integer_sequence<size_t, Idxs...>, Space,
      ...);
   }
 
+  KOKKOS_FUNCTION auto& reference() const { return *m_value_view.data(); }
```
masterleinad, line 213: Where is this used?

core/unit_test/TestTeamCombinedReducers.hpp
ldh4: Good point. I added in base case of `n=0` for each team policy and found some deficiencies in these tests.
It turns out that when `n=0` and when the inner `parallel_reduce` do not run, the outer `parallel_for` loop still runs, initializing intermediate team reduced values based on the corresponding reducer.
Added in if conditions to avoid unnecessary aggregates when the inner `parallel_reduce` do not run.

core/unit_test/TestTeamCombinedReducers.hpp
ldh4: I agree that these lines are quite redundant across the 3 test cases, but I think there is some value in keeping these unit tests separate based on their team policy. If any of these tests fails, I think I would find it simpler to tell which team policy failed based on the failing test name, whereas it may not be as clear at first glance if all 9 cases were in a single test.

core/src/impl/Kokkos_Combined_Reducer.hpp
ldh4: It's used in where each reducer's value is initialized and passed into provided functor.
For example, for `Cuda`:
https://github.com/kokkos/kokkos/blob/15989dd3a1edb72bbc178fc74d081502e32dd3d6/core/src/Cuda/Kokkos_Cuda_Team.hpp#L630-L649

core/src/impl/Kokkos_Combined_Reducer.hpp
ldh4: I'll resume on this shortly.
I recall looking into a workaround to consistently call `team_barrier` for all cases after learning that a reference to `team_member` is not stored in `ThreadVectorBoundariesStruct` for a few backends because of their constructors that do not require `team_member`.

ldh4: @masterleinad I've been giving rounds of thoughts on this, and I'm thinking that the fence operations are not necessary here, both fences on line 518 and 524.

These fences were to ensure that reduced values are ready to be copied back to the original references for all reducers in a combined reducer. This is meaningful if reductions are done on device and if the result from device needs to be copied back to reducers on host (this is what the regular combined reducer does).
However, for team combined reducers, both combined reduction part and copying back to its reducer part are on device and are done in sequence.

For example, for cuda `team_reduce` that is called from `parallel_reduce` that takes in a team policy:
https://github.com/kokkos/kokkos/blob/c09dd1c94861541b2f9651474ecdbec9e074dab7/core/src/Cuda/Kokkos_Cuda_Team.hpp#L192-L204
Syncing threads is already done within the reduction. Then the result from (combined) reduction is assigned in (combined) reducer, and thus another fence afterward shouldn't be necessary to start referencing reduced values on device.

masterleinad: Sounds good!

ldh4: Removed fences in [af212c0](https://github.com/kokkos/kokkos/pull/5727/commits/af212c0bc124056dc22706f18b1ac3da5c06f7ef).

masterleinad (approved):
Looks OK to me.

Rombur (approved):

core/unit_test/TestTeamCombinedReducers.hpp
```diff
@@ -0,0 +1,499 @@
...
+              [=](int const& i, int& localVal0, int& localVal1, int& localVal2,
+                  int& localVal3) {
+                localVal0 += 1;
+                localVal1 += i + 1;
+                localVal2 += (i + 1) * n;
+                localVal3 += n;
+              },
+              teamResult0, teamResult1, teamResult2, teamResult3);
+
+          if (n > 0) {
```
dalg24, line 55: No we don't want to guard this.

core/unit_test/TestTeamCombinedReducers.hpp
```diff
@@ -0,0 +1,499 @@
...
+
+    auto teamView = Kokkos::View<int*, TEST_EXECSPACE::memory_space>("view", 4);
+
+    Kokkos::parallel_for(
+        policy, KOKKOS_LAMBDA(team_member_type const& team) {
+          auto teamThreadRange = Kokkos::TeamThreadRange(team, n);
+          int teamResult0      = 0;
+          int teamResult1      = 0;
+          int teamResult2      = 0;
+          int teamResult3      = 0;
```
dalg24, line 42: Do not initialize, we need to make sure it is being handled properly when reducing over an empty range.

dalg24 (changes requested):

core/unit_test/TestTeamCombinedReducers.hpp
```diff
@@ -0,0 +1,499 @@
...
+          }
+        });
+
+    auto hostView = Kokkos::create_mirror_view_and_copy(
+        Kokkos::DefaultHostExecutionSpace(), teamView);
+
+    EXPECT_EQ(n, hostView(0));
+    EXPECT_EQ((n * (n + 1) / 2), hostView(1));
+    EXPECT_EQ(n * n * (n + 1) / 2, hostView(2));
+    EXPECT_EQ(n * n, hostView(3));
```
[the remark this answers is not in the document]
ldh4: No particular reason. Just wanted to test something more than just 2 reducers.

core/unit_test/TestTeamCombinedReducers.hpp
```diff
@@ -0,0 +1,499 @@
...
+// KOKKOS_ENABLE_CUDA_LAMBDA is off
+#if !defined(KOKKOS_ENABLE_CUDA) || defined(KOKKOS_ENABLE_CUDA_LAMBDA)
+
+struct TeamTeamCombinedReducer {
+ public:
+  void test_team_thread_range_only_scalars(const int n) {
+    auto policy = Kokkos::TeamPolicy<TEST_EXECSPACE>(1, Kokkos::AUTO);
+    using team_member_type = decltype(policy)::member_type;
+
+    auto teamView = Kokkos::View<int*, TEST_EXECSPACE::memory_space>("view", 4);
```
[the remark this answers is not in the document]
ldh4: Changed in [39b91e8](https://github.com/kokkos/kokkos/pull/5727/commits/39b91e8209eba18cecb3d077e97c85c6c7a9d2dc).

core/unit_test/TestTeamCombinedReducers.hpp
ldh4: Changed in [39b91e8](https://github.com/kokkos/kokkos/pull/5727/commits/39b91e8209eba18cecb3d077e97c85c6c7a9d2dc).

core/unit_test/TestTeamCombinedReducers.hpp
ldh4: Removed in [39b91e8](https://github.com/kokkos/kokkos/pull/5727/commits/39b91e8209eba18cecb3d077e97c85c6c7a9d2dc).

core/unit_test/TestTeamCombinedReducers.hpp
```diff
@@ -0,0 +1,499 @@
...
+                localVal0 += 1;
+                localVal1 += i + 1;
+                localVal2 += (i + 1) * n;
+                localVal3 += n;
+              },
+              teamResult0, teamResult1, teamResult2, teamResult3);
+
+          if (n > 0) {
+            Kokkos::single(Kokkos::PerTeam(team), [=]() {
+              teamView(0) += teamResult0;
```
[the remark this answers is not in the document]
ldh4: I don't think there was a particular reason. Changed to assigns in [39b91e8](https://github.com/kokkos/kokkos/pull/5727/commits/39b91e8209eba18cecb3d077e97c85c6c7a9d2dc).

core/unit_test/TestTeamCombinedReducers.hpp
```diff
@@ -0,0 +1,459 @@
...
+        });
+
+    auto hostView = Kokkos::create_mirror_view_and_copy(
+        Kokkos::DefaultHostExecutionSpace(), teamView);
+
+    EXPECT_EQ((n * (n + 1) / 2), hostView(0));
+    EXPECT_EQ((n * (n + 1) / 2), hostView(1));
+    EXPECT_EQ((n == 0) ? Kokkos::reduction_identity<int>::max() : n,
+              hostView(2));
+    EXPECT_EQ(n * n, hostView(3));
```
dalg24, line 392: Technically they all should be reduction_identity if n is equal to 0
```suggestion
    if (n == 0) {
      EXPECT_EQ(Kokkos::reduction_identity<int>::sum(), hostView(0));
      EXPECT_EQ(Kokkos::reduction_identity<int>::sum(), hostView(1));
      EXPECT_EQ(Kokkos::reduction_identity<int>::max(), hostView(2));
      EXPECT_EQ(Kokkos::reduction_identity<int>::sum(), hostView(3));
    } else {
      EXPECT_EQ((n * (n + 1) / 2), hostView(0));
      EXPECT_EQ((n * (n + 1) / 2), hostView(1));
      EXPECT_EQ(n, hostView(2));
      EXPECT_EQ(n * n, hostView(3));
    }
```
Same comment everywhere.

core/unit_test/TestTeamCombinedReducers.hpp
```diff
@@ -0,0 +1,459 @@
...
+// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
+#include <Kokkos_Core.hpp>
+
+#include <iostream>
+#include <limits>
```
dalg24, line 20:
```suggestion
#include <gtest/gtest.h>
```

core/unit_test/TestTeamCombinedReducers.hpp
```diff
@@ -0,0 +1,459 @@
...
+
+namespace Test {
```
dalg24, line 22:
```suggestion
namespace {
```

dalg24 (commented):
Please explicitly compare against reduction identity when the range is empty.
Looks good other than that.

core/unit_test/TestTeamCombinedReducers.hpp
ldh4: Changed in [2b7a12b](https://github.com/kokkos/kokkos/pull/5727/commits/2b7a12b285fce804e0f299e22aa59ce332397810).

dalg24:
I plan on ignoring 
```
5: [ RUN      ] cuda.workgraph_fib
5: /var/jenkins/workspace/Kokkos/core/unit_test/TestWorkGraph.hpp:127: Failure
5: Expected equality of these values:
5:   h_values(0)
5:     Which is: 5778
5:   full_fibonacci(m_input)
5:     Which is: 6765
5: [  FAILED  ] cuda.workgraph_fib (1426 ms)
```
in the CUDA-11.6-NVCC build.  Still waiting on at least one of the HIP build to pass before I merge.

ldh4:
Retest this please.

PR #5706, opened 2022-12-19
Rework CUDA cache config using carveout calculations

crtrott:
This will generally rely on the CUDA runtime to figure out best cache configuration, without us setting it manually anymore. We only set it manually if and when someone requests restricted occupancy. However, I did test the carveout scheme when always manually setting, and it did basically replicate the performance obtained by not setting anything for LAMMPS - and beats what we had before which was more limited. I also wrote a test code to check that the occupancy limitation works more or less as expected. Note that this is fairly iffy since there are many moving parts.

```c++
#include <Kokkos_Core.hpp>
#include <cmath>

void check_range(int N, int occupancy, int work, int size) {
  Kokkos::View<int> count("C");
  Kokkos::View<double**> workarray("A",N,size);

  auto lambda = KOKKOS_LAMBDA(int i, int& lmax) {
    int id = Kokkos::atomic_fetch_add(&count(), 1);
    for(int j=1; j<work; j++) workarray(i,j%size) += workarray(i, (j-1)%size) + j;
    Kokkos::atomic_add(&count(), -1);
    if(id > lmax) lmax = id;
  };

  Kokkos::RangePolicy<> p(0,N);
  auto occ_p = Kokkos::Experimental::prefer(p, Kokkos::Experimental::DesiredOccupancy{occupancy});
  int max_val;
  Kokkos::parallel_reduce(occ_p, lambda, Kokkos::Max<int>(max_val));
  printf("Occ: %i Max: %i\n",occupancy, max_val);
}

void check_team(int N, int occupancy, int work, int size, int team_size, int shmem) {
  Kokkos::View<int> count("C");
  Kokkos::View<double**> workarray("A",N,size);

  auto lambda = KOKKOS_LAMBDA(const typename Kokkos::TeamPolicy<>::member_type& team, int& lmax) {
    int i = team.league_rank()*team_size+team.team_rank();
    int id = Kokkos::atomic_fetch_add(&count(), 1);
    for(int j=1; j<work; j++) workarray(i,j%size) += workarray(i, (j-1)%size) + j;
    Kokkos::atomic_add(&count(), -1);
    if(id > lmax) lmax = id;
  };

  Kokkos::TeamPolicy<> p(N/team_size, team_size);
  p.set_scratch_size(0, Kokkos::PerTeam(shmem));
  auto occ_p = Kokkos::Experimental::prefer(p, Kokkos::Experimental::DesiredOccupancy{occupancy});
  int max_val;
[...]

PR #5731, opened 2023-01-06
Cleanup CUDA block size deduction

dalg24:
* ~~Minimum CUDA required version is 11 so the `#if CUDA_VERSION >= 11000` to get `max_blocks_per_sm` can be removed.~~  **edit** I was reminded that we support CUDA 10 with Clang as the CUDA compiler
* `get_shmem_per_sm_prefer_l1` was only used in code removed in #5706


masterleinad:
For `Clang` we allow `CUDA` 10, right?

dalg24:
> For `Clang` we allow `CUDA` 10, right?

Damn you are right I forgot.

masterleinad (approved):

crtrott (approved):

PR #5735, opened 2023-01-06
Remove KOKKOS_CXX_STANDARD mentioning from BUILD.md

crtrott:


janciesko (approved):

BUILD.md
```diff
@@ -52,6 +52,10 @@ There are numerous device backends, options, and architecture-specific optimizat
 ````
 which activates the OpenMP backend. All of the options controlling device backends, options, architectures, and third-party libraries (TPLs) are given below.
 
+Kokkos requires as a minimum C++17, however C++20 and C++23 are supported depending on the compiler.
```
dalg24, line 55: Sounds good but what does "C++20 and C++23 are supported depending on the compiler" actually mean?

crtrott: whether the compiler supports it?

ajpowelsnl:
@crtrott - can we close?

crtrott:
It is merged so that is the equivalent of closing ...

PR #5738, opened 2023-01-09
Refactor OpenMP backend

Rombur:
This mainly moves code around with the following exceptions: `in_parallel`,  `impl_thread_pool_size`,  and `impl_max_hardware_threads` are not inlined anymore. This is because of cyclical dependency, the only way to keep these functions inlined is to implement then in `Kokkos_OpenMP_Instance.hpp` even though they are declared in `Kokkos_OpenMP.hpp`. These functions are simple so I am counting on the compiler to do the right thing but I haven't done any measurements. I can revert that change.


masterleinad (approved):

cz4rs (approved):
Looks good to me.

PR #5739, opened 2023-01-09
Further update to CUDA occupancy calculation

weinbe2:
This PR is a further correction to the occupancy calculation as a function of the threads per register for the CUDA backend. It is a bugfix relative to #5624 , where testing missed some corner-cases with awkward thread counts and block sizes.

The background of this PR is that, on top of the register allocation granularity of 256 registers per warp (== 8 registers per thread), there is a warp per SM allocation granularity that varies as a function of the architecture (though it has been 4 warps/SM on all recent architectures except CC 6.0). This further constrains the occupancy as a function of the register count. This cannot be programmatically queried via the CUDA device properties, though it is public information included in the CUDA occupancy calculator. I baked in a function to return this value.

An example of this behavior: consider a function requiring 92 registers per thread. Given the register allocation granularity of 256 registers/warp -> 8 registers/thread, this effectively rounds up to the nearest multiple of 8, or 96 registers per thread. Ignoring the warp granularity constraint, the block size that maximizes occupancy is 224 threads/block, which naively corresponds to a residency of 3 blocks -> 21 warps per SM.

However, due to the allocation granularity of 4 warps/SM (on Ampere, for example), this needs to effectively be rounded up to 24 warps per SM, which does not fit, and as such this launch configuration corresponds to 2 blocks -> 14 warps per SM, which is sub-optimal. After factoring in the warp allocation granularity constraint, the maximum warps per SM is 20 warps, and as such a blocksize of 128 or 160 is ideal --- and Kokkos' heuristics pick 128, since it'll pick the smallest block size providing maximum occupancy, capped from below at 128 threads/block.

I propagated this updated heuristic to the `max_tile_size_product` function for `MDRangePolicy`, though it seems like that isn't used except for tile size tuning?


dalg24:
OK to test

core/src/Cuda/Kokkos_Cuda_Parallel_MDRange.hpp
```diff
@@ -63,9 +63,19 @@ class ParallelFor<FunctorType, Kokkos::MDRangePolicy<Traits...>, Kokkos::Cuda> {
     auto const& prop = pol.space().cuda_device_prop();
     // Limits due to registers/SM, MDRange doesn't have
     // shared memory constraints
-    int const regs_per_sm        = prop.regsPerMultiprocessor;
-    int const regs_per_thread    = attr.numRegs;
-    int const max_threads_per_sm = regs_per_sm / regs_per_thread;
+    auto const block_size_to_zero_shmem = [&] (int block_size) {
+      return 0;
+    };
```
dalg24, line 68:
```suggestion
    auto const block_size_to_zero_shmem = [] (int /*block_size*/) {
      return 0;
    };
```
same comment below

core/src/Cuda/Kokkos_Cuda_Parallel_MDRange.hpp
```diff
@@ -63,9 +63,19 @@ class ParallelFor<FunctorType, Kokkos::MDRangePolicy<Traits...>, Kokkos::Cuda> {
...
+    int const optimal_block_size = Kokkos::Impl::cuda_deduce_block_size(false, prop, attr, block_size_to_zero_shmem, LaunchBounds{});
```
dalg24, line 69: This is a bit awkward.  This function was never intended to be exposed beyond the `<Cuda/Kokkos_Cuda_BlockSize_Deduction.hpp>` header.  Maybe we could define an `Impl::cuda_get_opt_block_size_no_shmem()` that does not have the leading boolean parameter nor the lambda.

core/src/Cuda/Kokkos_Cuda_BlockSize_Deduction.hpp
```diff
@@ -25,17 +25,50 @@
...
+
+  int const max_warps_per_sm = warp_granularity * (max_regs_per_block / (allocated_regs_per_warp * warp_granularity));
+
+  return max_warps_per_sm;
+}
+
 inline int cuda_max_active_blocks_per_sm(cudaDeviceProp const& properties,
                                          cudaFuncAttributes const& attributes,
-                                         int block_size, size_t dynamic_shmem) {
+                                         int max_warps_per_sm_registers, int block_size, size_t dynamic_shmem) {
```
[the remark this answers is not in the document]
weinbe2: I didn't specifically benchmark it, but since it was something that was independent of `block_size` and `dynamic_shmem` it seemed like it was something good to compute once, reuse elsewhere, especially in the context of the `for` loop of the block size calculation where this function gets called multiple times.

I should have made it a `const` value, to be fair. I could also make it the last argument and give it a default value (`-1`?) such that when it isn't passed in it'll get computed inside this function. Thoughts?

core/src/Cuda/Kokkos_Cuda_Parallel_MDRange.hpp
weinbe2: Sure, I can do that today/early tomorrow. Good suggestion.

core/src/Cuda/Kokkos_Cuda_Parallel_MDRange.hpp
weinbe2: replied below

weinbe2:
Thanks for the feedback @dalg24 , I'll get to this today or early tomorrow.

core/src/Cuda/Kokkos_Cuda_BlockSize_Deduction.hpp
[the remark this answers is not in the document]
weinbe2: Ultimately ended up pulling the calculation in, since it is indeed cheap.

core/src/Cuda/Kokkos_Cuda_Parallel_MDRange.hpp
weinbe2: Added said new function.

weinbe2:
I addressed (I believe) the review comments, and added some additional inline documentation explaining the logic, since it gets a bit messy. @crtrott , I also propagated these updates to your occupancy control work as best I could tell --- could you please sanity check them, and/or run some of the tests you ran as part of your PR? (Or point me at the best way to do so!)

core/src/Cuda/Kokkos_Cuda_BlockSize_Deduction.hpp
```diff
@@ -25,17 +25,60 @@
...
+  int const regs_per_warp = attributes.numRegs * properties.warpSize;
+  int const warp_granularity = cuda_warp_per_sm_allocation_granularity(properties);
+  // The granularity of register allocation is chunks of 256 registers per warp,
+  // which implies a need to over-allocate, so we round up
+  int const allocated_regs_per_warp = (regs_per_warp + 256 - 1) / 256;
+
+  // The maximum number of warps per SM is constrained from above by register allocation.
+  // To satisfy the constraint that warps per SM is allocated at a finite granularity,
+  // we need to round down.
+  int const max_warps_per_sm = warp_granularity * (max_regs_per_block / (allocated_regs_per_warp * warp_granularity));
```
Rombur, line 57: Does this work if you use `-ffast-math` ?

weinbe2: I haven't explicitly tested it (yet), but based on some quick reading online I don't think `-ffast-math` touches integer division? (If it did, I feel like it would break a handful of other codes...)

Rombur: I am not sure that's why I ask. With floating-point this wouldn't work but I don't know about integer

weinbe2: Just tested it in godbolt, it generates `imul`, `idiv`, `imul` whether or not `-ffast-math` is set: https://godbolt.org/z/G3sz4hjfq

stanmoore1:
Would love to get this merged ASAP so I can backport it into LAMMPS. It fixes a performance regression in the ReaxFF model from https://github.com/kokkos/kokkos/pull/5624. Thanks.

core/src/Cuda/Kokkos_Cuda_BlockSize_Deduction.hpp
PhilMiller: The standard defines integer arithmetic semantics much more concretely than floating point. Overflow is UB, everything else has a precise result.

core/src/Cuda/Kokkos_Cuda_Parallel_MDRange.hpp
```diff
@@ -234,10 +241,18 @@ class ParallelReduce<FunctorType, Kokkos::MDRangePolicy<Traits...>, ReducerType,
...
+    // shared memory constraints
+    int const optimal_block_size = Kokkos::Impl::cuda_get_opt_block_size_no_shmem(attr, LaunchBounds{});
+
+    // Compute how many blocks of this size we can launch, based on warp constraints
+    int const max_warps_per_sm_registers = Kokkos::Impl::cuda_max_warps_per_sm_registers(prop, attr);
+    int const max_num_threads_from_warps = max_warps_per_sm_registers * prop.warpSize;
+    int const max_num_blocks = max_num_threads_from_warps / optimal_block_size;
+
+    // Compute the total number of threads
+    int const max_threads_per_sm = optimal_block_size * max_num_blocks;
```
Rombur, line 254: This code is the same as above right? Is it possible to put it in a free function to avoid duplication.

Rombur (commented):
This looks good. I just have a minor comment

core/src/Cuda/Kokkos_Cuda_Parallel_MDRange.hpp
weinbe2: Yes, it's identical. Were you thinking in this file, or in `Kokkos_Cuda_BlockSize_Deduction.hpp`?

Rombur: Either way. I don't have a preference

weinbe2: https://github.com/kokkos/kokkos/pull/5739/commits/3a23689ba3e861d261b7e251eaa56aaa4e906a1f

weinbe2: Whoops, left some debugging stuff in there...

weinbe2: https://github.com/kokkos/kokkos/pull/5739/commits/dd72c16017f4e02770af40546b4064ae39205716

core/src/Cuda/Kokkos_Cuda_BlockSize_Deduction.hpp
```diff
@@ -25,17 +25,60 @@
...
 inline int cuda_max_active_blocks_per_sm(cudaDeviceProp const& properties,
                                          cudaFuncAttributes const& attributes,
                                          int block_size, size_t dynamic_shmem) {
-  // Limits due do registers/SM
+  // Limits due to registers/SM
   int const regs_per_sm     = properties.regsPerMultiprocessor;
   int const regs_per_thread = attributes.numRegs;
   // The granularity of register allocation is chunks of 256 registers per warp
   // -> 8 registers per thread
   int const allocated_regs_per_thread = 8 * ((regs_per_thread + 8 - 1) / 8);
```
dalg24, line 70: Is that still needed?

Rombur (approved):

Rombur:
You need to apply `clang-format`

core/src/Cuda/Kokkos_Cuda_BlockSize_Deduction.hpp
weinbe2: Yes, they're two separate (albeit highly correlated) constraints.

stanmoore1:
@dalg24-jenkins @PhilMiller is this ready to go? I confirmed it is working as expected.

stanmoore1:
Test failure related to `sprintf` being deprecated looks unrelated to any changes in this PR.

core/src/Cuda/Kokkos_Cuda_BlockSize_Deduction.hpp
```diff
@@ -25,17 +25,69 @@
...
-      regs_per_sm / (allocated_regs_per_thread * block_size);
+  int max_blocks_regs = regs_per_sm / (allocated_regs_per_thread * block_size);
+
+  // Compute the maximum number of warps as a function of the number of
+  // registers
+  int const max_warps_per_sm_registers =
+      cuda_max_warps_per_sm_registers(properties, attributes);
+
+  // Constrain the number of blocks to respect the maximum number of warps per
+  // SM On face value this should be an equality, but due to the warp
```
dalg24, line 84: Missing a "dot"?
Don't really understand the sentence, not sure what "On face value" means.

core/src/Cuda/Kokkos_Cuda_BlockSize_Deduction.hpp
```diff
@@ -179,6 +231,19 @@ int cuda_get_opt_block_size(const CudaInternal* cuda_instance,
                                 LaunchBounds{});
 }
 
+template <class LaunchBounds>
+int cuda_get_opt_block_size_no_shmem(const cudaFuncAttributes& attr,
+                                     LaunchBounds) {
+  auto const& prop = Kokkos::Cuda().cuda_device_prop();
```
dalg24, line 237: Take it as an argument, it is available at call site.

core/src/Cuda/Kokkos_Cuda_BlockSize_Deduction.hpp
```diff
@@ -179,6 +231,19 @@ int cuda_get_opt_block_size(const CudaInternal* cuda_instance,
...
+
+  // Thin version of cuda_get_opt_block_size for cases where there is no shared
+  // memory
+  auto const block_size_to_no_shmem = [&](int /*block_size*/) { return 0; };
+
+  return cuda_deduce_block_size(false, prop, attr, block_size_to_no_shmem,
+                                LaunchBounds{});
+}
```
dalg24, line 245:
```suggestion
// Thin version of cuda_get_opt_block_size for cases where there is no shared
// memory
template <class LaunchBounds>
int cuda_get_opt_block_size_no_shmem(
    const cudaDeviceProp& pro const cudaFuncAttributes& attr, LaunchBounds) {
  return cuda_deduce_block_size(
      false, prop, attr, [](int /*block_size*/) { return 0; }, LaunchBounds{});
}
```

core/src/Cuda/Kokkos_Cuda_Parallel_MDRange.hpp
```diff
@@ -39,6 +39,34 @@
...
+int max_tile_size_product_helper(const Policy& pol, const LaunchBounds&) {
+  cudaFuncAttributes attr =
+      CudaParallelLaunch<ParallelType,
+                         LaunchBounds>::get_cuda_func_attributes();
+  auto const& prop = pol.space().cuda_device_prop();
+
+  // Limits due to registers/SM, MDRange doesn't have
+  // shared memory constraints
+  int const optimal_block_size =
+      Kokkos::Impl::cuda_get_opt_block_size_no_shmem(attr, LaunchBounds{});
```
dalg24, line 52:
```suggestion
      cuda_get_opt_block_size_no_shmem(prop, attr, LaunchBounds{});
```

dalg24 (commented):
Looks good.  I only have a few minor comments.

core/src/Cuda/Kokkos_Cuda_BlockSize_Deduction.hpp
weinbe2: Ah, the missing dot is a byproduct of `clang-format` changing the width of comments. In any case, I'll make this clearer.

core/src/Cuda/Kokkos_Cuda_BlockSize_Deduction.hpp
```diff
@@ -179,6 +231,15 @@ int cuda_get_opt_block_size(const CudaInternal* cuda_instance,
                                 LaunchBounds{});
 }
 
+// Thin version of cuda_get_opt_block_size for cases where there is no shared
+// memory
+template <class LaunchBounds>
+int cuda_get_opt_block_size_no_shmem(
+    const cudaDeviceProp& pro const cudaFuncAttributes& attr, LaunchBounds) {
```
weinbe2, line 238: I realized I accepted this without noticing it wouldn't compile, fixing now...

core/src/Cuda/Kokkos_Cuda_Parallel_MDRange.hpp
```diff
@@ -39,6 +39,34 @@
 namespace Kokkos {
 namespace Impl {
 
+template <typename ParallelType, typename Policy, typename LaunchBounds>
+int max_tile_size_product_helper(const Policy& pol, const LaunchBounds&) {
```
dalg24, line 43: The policy has the launch bounds info so we could deduce it
(Not asking you to change, it is already good enough)

dalg24 (approved):
Assuming the CI passes

weinbe2:
Looking at the failing test (https://github.com/kokkos/kokkos/blob/develop/core/unit_test/TestWorkGraph.hpp , CUDA backend), I can see how it _possibly_ could be related to this PR, but I'm seeing it's failing in other PR tests as well (https://github.com/kokkos/kokkos/pull/5778) for different test values. Looking at that code quickly, I can convince myself at a high level that there could be memory consistency issues... thoughts?

dalg24:
Retest this please

weinbe2:
@dalg24 looks like a different test now...

dalg24:
> @dalg24 looks like a different test now...

The cuda timing-based test occasionally fails (sigh) but since it is on that same NVHPC build, will re-run to be sure...

Retest this please

weinbe2:
> > @dalg24 looks like a different test now...
> 
> The cuda timing-based test occasionally fails (sigh) but since it is on that same NVHPC build, will re-run to be sure...
> 
> Retest this please

Not quite sure what went awry this time...

Rombur:
The error looks unrelated. The problem is from SYCL

dalg24:
SYCL failure is unrelated (fixed in #5821)

dalg24:
Was about to merge...  Waiting for CI now I guess

weinbe2:
> Was about to merge... Waiting for CI now I guess

Ugh, I'm sorry. My old habits of periodically merging in `develop` bit me here.

dalg24:
Having issue with the CI.  Deciding to ignore for the sake of making progress

PR #5740, opened 2023-01-09
OpenACC: Minor bug fixes missing header include and generated makefiles configuration

seyonglee:
Minor bug fixes on CMake and Make configurations for the OpenACC backend

### DIFF
--- a/core/src/Kokkos_Macros.hpp
+++ b/core/src/Kokkos_Macros.hpp
@@ -522,6 +522,7 @@ static constexpr bool kokkos_omp_on_host() { return false; }
     KOKKOS_IMPL_STRIP_PARENS(CODE)   \
   }
 #else
+#include <openacc.h>
 // FIXME_OPENACC acc_on_device is a non-constexpr function
 #define KOKKOS_IF_ON_DEVICE(CODE)                     \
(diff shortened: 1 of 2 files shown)


core/src/Kokkos_Macros.hpp
```diff
@@ -522,6 +522,7 @@ static constexpr bool kokkos_omp_on_host() { return false; }
     KOKKOS_IMPL_STRIP_PARENS(CODE)   \
   }
 #else
+#include <openacc.h>
```
dalg24, line 525: What compiler did you test with?
(If that is sensitive information you can reach out to me directly)

seyonglee: We use Clacc, another OpenACC compiler built on LLVM, which will support both NVIDIA GPUs and AMD GPUs.
(For now, Clacc is still under development, and thus not all Kokkos implementations can be compiled by Clacc.)

seyonglee:
> What CMake bug? Did you mean to commit some other change?
I made a wrong comment; I fixed bugs in both make configuration and Kokkos macros (but not CMake).

dalg24 (approved):

Rombur (approved):

Issue #5600, opened 2022-10-26
`OpenMP` performance regression in `3.7.0`


cz4rs:
**Describe the bug**
Reported via Slack:

> Using OMP_NUM_THREADS=64 on an AMD Threadripper 3970x this is about 3-4x slower using Kokkos 3.7 than Kokkos 3.6
> If I set OMP_NUM_THREADS=32 it is only about 2x slower on Kokkos 3.7

<details>
  <summary>Reproduction code</summary>
  
```cpp
#include <Kokkos_Core.hpp>
#include <list>

using Device = Kokkos::DefaultHostExecutionSpace;

void driver(){
    uint32_t n = 200000000;
    Kokkos::View<uint32_t*, Device> a("test", n);
    std::list<Kokkos::View<uint32_t*, Device>> sources;
    uint32_t t_iter = 10;
    Kokkos::fence();
    Kokkos::Timer t;
    double begin = t.seconds();
    uint32_t segments = 64;
    uint32_t width = n / segments;
    for(uint32_t i = 0; i < t_iter; i++){
        for(uint32_t j = 0; j < segments; j++){
            auto source = Kokkos::subview(a, std::make_pair(width*j, width*(j+1)));
            Kokkos::View<uint32_t*, Device> dest(Kokkos::ViewAllocateWithoutInitializing("dest"), width);
            Kokkos::deep_copy(dest, source);
            sources.push_back(dest);
        }
        Kokkos::fence();
    }
    double end = t.seconds();
    double d = end - begin;
    double avg = d / static_cast<double>(t_iter);
    double bandwidth = static_cast<double>(2*sizeof(uint32_t)*n) / avg;
    printf("uint32_t size: %li\n", sizeof(uint32_t));
    printf("Total time: %.2fs; Avg Time: %.2fs; Bandwidth: %.2fMB/s\n", d, avg, bandwidth / 1000000);
}

int main() {
    Kokkos::initialize();
    {
        driver();
    }
    Kokkos::finalize();
    return 0;
}
```
</details>

**Configuration**
```
../generate_makefile.bash --kokkos-path=/home/user/workspace/kokkos \
--prefix=/home/user/workspace/install/kokkos \
--with-cuda --arch=Ampere86 --compiler=/home/user/workspace/kokkos/bin/nvcc_wrapper \
--with-cuda-options=enable_lambda --with-openmp --with-serial
```
- Ubuntu 20.04 LTS
- `g++` 10.2
[...]

PR #5742, opened 2023-01-10
OpenMP: Don't use chunk size for static scheduling on GNU compiler

rgayatri23:
The PR fixed the regression discussed in #5600

### DIFF
--- a/core/src/OpenMP/Kokkos_OpenMP_Parallel.hpp
+++ b/core/src/OpenMP/Kokkos_OpenMP_Parallel.hpp
@@ -101,8 +101,15 @@ class ParallelFor<FunctorType, Kokkos::RangePolicy<Traits...>, Kokkos::OpenMP> {
   std::enable_if_t<!std::is_same<typename Policy::schedule_type::type,
                                  Kokkos::Dynamic>::value>
   execute_parallel() const {
+// Specifying an chunksize with GCC compiler leads to performance regression
+// with static schedule.
+#ifdef KOKKOS_COMPILER_GNU
+#pragma omp parallel for schedule(static) \
+    num_threads(m_instance->thread_pool_size())
+#else
 #pragma omp parallel for schedule(static KOKKOS_OPENMP_OPTIONAL_CHUNK_SIZE) \
     num_threads(m_instance->thread_pool_size())
+#endif
     KOKKOS_PRAGMA_IVDEP_IF_ENABLED
     for (auto iwork = m_policy.begin(); iwork < m_policy.end(); ++iwork) {
       exec_work(m_functor, iwork);


cmake/kokkos_enable_options.cmake
```diff
@@ -65,6 +65,9 @@ KOKKOS_ENABLE_OPTION(HIP_MULTIPLE_KERNEL_INSTANTIATIONS OFF "Whether multiple ke
 KOKKOS_ENABLE_OPTION(IMPL_DESUL_ATOMICS   ON  "Whether to use desul based atomics - option only during beta")
 KOKKOS_ENABLE_OPTION(DESUL_ATOMICS_EXTERNAL OFF "Whether to use an external desul installation")
 
+# Switched on by default
+KOKKOS_ENABLE_OPTION(IMPL_OPENMP_THREADPOOL_ONLY ON "To use native OpenMP implementation post 3.6 version")
```
crtrott, line 69: Change comment to: "Use legacy OpenMP thread-pool based loop implementation."

crtrott (changes requested):

cmake/kokkos_enable_options.cmake
cz4rs: Should the legacy option be `ON` by default?
I think we could  use refactored (native OpenMP) version by default and only introduce this to deal with the regression for the affected systems.

dalg24 (dismissed):
Please include this in `OpenMP::print_configuration`

core/src/OpenMP/Kokkos_OpenMP_Instance.cpp
```diff
@@ -427,6 +427,13 @@ void OpenMP::print_configuration(std::ostream &os, bool /*verbose*/) const {
   os << "no\n";
 #endif
 
+  os << " KOKKOS_ENABLE_IMPL_OPENMP_THREADPOOL_ONLY: ";
```
dalg24:
```suggestion
  os << "OpenMP Options:\n";
  os << "  KOKKOS_ENABLE_IMPL_OPENMP_THREADPOOL_ONLY: ";
```

core/src/OpenMP/Kokkos_OpenMP_Instance.cpp
```diff
@@ -384,68 +392,4 @@ bool OpenMPInternal::verify_is_initialized(const char *const label) const {
...
-      name, Kokkos::Tools::Experimental::Impl::DirectFenceIDHandle{1}, []() {});
-}
-
-namespace Impl {
-
-int g_openmp_space_factory_initialized =
-    initialize_space_factory<OpenMP>("050_OpenMP");
-
-}  // namespace Impl
-
```
dalg24, line 450: Rebase gone wrong

cz4rs:
Segfault in `KokkosCore_PerfTestExec` (`default_exec.overlap_range_policy`) is caused by https://github.com/kokkos/kokkos/commit/e4cce6e44.
This was not discovered in CI because all of our builds were running native OpenMP implementation.

core/src/OpenMP/Kokkos_OpenMP_Parallel.hpp
```diff
@@ -101,8 +106,13 @@ class ParallelFor<FunctorType, Kokkos::RangePolicy<Traits...>, Kokkos::OpenMP> {
   std::enable_if_t<!std::is_same<typename Policy::schedule_type::type,
                                  Kokkos::Dynamic>::value>
   execute_parallel() const {
+#ifdef KOKKOS_COMPILER_GNU
```
cz4rs, line 109: We already skip the chunk size for NVHPC (not supported) in lines 38 - 42. You could stick with `KOKKOS_OPENMP_OPTIONAL_CHUNK_SIZE` macro by checking for GNU there.

dalg24: It looks like this patch is only wanting to drop the chunk size for static scheduling.  The `KOKKOS_OPENMP_OPTIONAL_CHUNK_SIZE` is also used (12L above) with a dynamic schedule.

cz4rs: True, my bad :grin:

Makefile.kokkos
```diff
@@ -1,7 +1,7 @@
 # Default settings common options.
 
-KOKKOS_VERSION_MAJOR = 3
-KOKKOS_VERSION_MINOR = 7
+KOKKOS_VERSION_MAJOR = 4
+KOKKOS_VERSION_MINOR = 0
```
dalg24, line 4: undo

core/src/OpenMP/Kokkos_OpenMP_Parallel.hpp
rgayatri23: Yeah the dynamic schedule does not show any regression so I did not do anything for it.

rgayatri23:
Retest this please.

cz4rs:
> Retest this please.

@rgayatri23 The history is still messed up. This could probably be just one commit with the changes from `Kokkos_OpenMP_Parallel.hpp`?

cz4rs (approved):
Looks good to me.

crtrott (approved):

PR #5745, opened 2023-01-11
#5641: Fix HIP & CUDA MDRange reduce for sizeof(value_type) < sizeof(int)

### LINKED ISSUE #5641: Parallel reducer, `MDRangePolicy` and HIP or CUDA: doesn't work with bool, char, int16_t
**Describe the bug**

It seems that `Kokkos::LAnd<bool>` does not work with `Kokkos::Experimental::HIP` and `Kokkos::MDRangePolicy`.

* Kokkos release 3.7
* Linux with ROCm 5.3 (vega906)
* code needed to reproduce the bug
   ```c++
   template <typename T, typename execution_space, typename policy_t, typename functor_t>
   T parallel_reducer_LAnd(const policy_t& policy, const functor_t& f)
   {
      T allEqual = 0;
      auto reducer = Kokkos::LAnd<T>(allEqual);
      Kokkos::parallel_reduce(policy,f,reducer);
      execution_space().fence();
      return allEqual;
   }

   TEUCHOS_UNIT_TEST(test,test)
   {
      std::cout << "KOKKOS_VERSION: " << KOKKOS_VERSION << std::endl;

      using execution_space = Kokkos::Experimental::HIP;

      const auto  policy_MD = Kokkos::MDRangePolicy<Kokkos::Rank<2>,execution_space>({0,0},{8,2});
      const auto  policy_RG = Kokkos::  RangePolicy<                execution_space>( 0,    8   );

      const auto res_MD_bool = parallel_reducer_LAnd<bool,execution_space>(policy_MD, KOKKOS_LAMBDA(const size_t, const size_t, bool& res){res = true;});
      const auto res_MD_int  = parallel_reducer_LAnd<int ,execution_space>(policy_MD, KOKKOS_LAMBDA(const size_t, const size_t, int & res){res = 1   ;});

      const auto res_RG_bool = parallel_reducer_LAnd<bool,execution_space>(policy_RG, KOKKOS_LAMBDA(const size_t, bool& res){res = true;});
      const auto res_RG_int  = parallel_reducer_LAnd<int ,execution_space>(policy_RG, KOKKOS_LAMBDA(const size_t, int & res){res = 1   ;});

      TEST_EQUALITY(res_MD_bool,false); // -> I expected true here
      TEST_EQUALITY(res_RG_bool,true);
      TEST_EQUALITY(res_MD_int ,1);
      TEST_EQUALITY(res_RG_int ,1);
   }
   ```
7. please also attach the `KokkosCore_config.h` header file (generated during the build); [KokkosCore_config.hpp.txt](https://github.com/kokkos/kokkos/files/10102708/KokkosCore_config.hpp.txt)

As you can see from the reproducer, using the `Kokkos::LAnd<bool>` with `MDRangePolicy` on HIP seems not to work correctly.

A temporary workaround is to use `int` instead.

Please note that for `OpenMP` and `Cuda` it works fine.

PhilMiller:
Fixes #5641

### DIFF
--- a/core/src/Cuda/Kokkos_Cuda_Parallel_MDRange.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_Parallel_MDRange.hpp
@@ -262,7 +262,7 @@ class ParallelReduce<CombinedFunctorReducerType,
       // Number of blocks is bounded so that the reduction can be limited to two
       // passes. Each thread block is given an approximately equal amount of
       // work to perform. Accumulate the values for this block. The accumulation
-      // ordering does not match the final pass, but is arithmatically
+      // ordering does not match the final pass, but is arithmetically
       // equivalent.
 
       this->exec_range(value);
--- a/core/src/Cuda/Kokkos_Cuda_Parallel_MDRange.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_Parallel_MDRange.hpp
@@ -212,6 +212,23 @@ class ParallelReduce<CombinedFunctorReducerType,
   using size_type      = Cuda::size_type;
   using reducer_type   = ReducerType;
 
+  // Conditionally set word_size_type to int16_t or int8_t if value_type is
+  // smaller than int32_t (Kokkos::Cuda::size_type)
+  // word_size_type is used to determine the word count, shared memory buffer
+  // size, and global memory buffer size before the reduction is performed.
+  // Within the reduction, the word count is recomputed based on word_size_type
+  // and when calculating indexes into the shared/global memory buffers for
+  // performing the reduction, word_size_type is used again.
+  // For scalars > 4 bytes in size, indexing into shared/global memory relies
+  // on the block and grid dimensions to ensure that we index at the correct
+  // offset rather than at every 4 byte word; such that, when the join is
+  // performed, we have the correct data that was copied over in chunks of 4
+  // bytes.
+  static_assert(sizeof(size_type) == 4);
+  using word_size_type = std::conditional_t<
+      sizeof(value_type) < 4,
+      std::conditional_t<sizeof(value_type) == 2, int16_t, int8_t>, size_type>;
+
   // Algorithmic constraints: blockSize is a power of two AND blockDim.y ==
   // blockDim.z == 1
 
@@ -219,9 +236,9 @@ class ParallelReduce<CombinedFunctorReducerType,
   const Policy m_policy;  // used for workrange and nwork
   const pointer_type m_result_ptr;
   const bool m_result_ptr_device_accessible;
-  size_type* m_scratch_space;
+  word_size_type* m_scratch_space;
   size_type* m_scratch_flags;
-  size_type* m_unified_space;
+  word_size_type* m_unified_space;
 
   using DeviceIteratePattern = typename Kokkos::Impl::Reduce::DeviceIterateTile<
       Policy::rank, Policy, FunctorType, typename Policy::work_tag,
@@ -248,15 +265,16 @@ class ParallelReduce<CombinedFunctorReducerType,
   }
 
   inline __device__ void operator()() const {
-    const integral_nonzero_constant<
-        size_type, ReducerType::static_value_size() / sizeof(size_type)>
+    const integral_nonzero_constant<word_size_type,
+                                    ReducerType::static_value_size() /
+                                        sizeof(word_size_type)>
         word_count(m_functor_reducer.get_reducer().value_size() /
-                   sizeof(size_type));
+                   sizeof(word_size_type));
 
     {
       reference_type value =
           m_functor_reducer.get_reducer().init(reinterpret_cast<pointer_type>(
-              kokkos_impl_cuda_shared_memory<size_type>() +
+              kokkos_impl_cuda_shared_memory<word_size_type>() +
               threadIdx.y * word_count.value));
 
       // Number of blocks is bounded so that the reduction can be limited to two
@@ -272,15 +290,16 @@ class ParallelReduce<CombinedFunctorReducerType,
     // Problem: non power-of-two blockDim
     if (cuda_single_inter_block_reduce_scan<false>(
             m_functor_reducer.get_reducer(), blockIdx.x, gridDim.x,
-            kokkos_impl_cuda_shared_memory<size_type>(), m_scratch_space,
+            kokkos_impl_cuda_shared_memory<word_size_type>(), m_scratch_space,
             m_scratch_flags)) {
       // This is the final block with the final result at the final threads'
       // location
-      size_type* const shared = kokkos_impl_cuda_shared_memory<size_type>() +
-                                (blockDim.y - 1) * word_count.value;
-      size_type* const global =
+      word_size_type* const shared =
+          kokkos_impl_cuda_shared_memory<word_size_type>() +
+          (blockDim.y - 1) * word_count.value;
+      word_size_type* const global =
           m_result_ptr_device_accessible
-              ? reinterpret_cast<size_type*>(m_result_ptr)
+              ? reinterpret_cast<word_size_type*>(m_result_ptr)
               : (m_unified_space ? m_unified_space : m_scratch_space);
 
       if (threadIdx.y == 0) {
@@ -342,13 +361,16 @@ class ParallelReduce<CombinedFunctorReducerType,
                        : suggested_blocksize;  // Note: block_size must be less
                                                // than or equal to 512
 
-      m_scratch_space = cuda_internal_scratch_space(
-          m_policy.space(), m_functor_reducer.get_reducer().value_size() *
-                                block_size /* block_size == max block_count */);
+      m_scratch_space =
+          reinterpret_cast<word_size_type*>(cuda_internal_scratch_space(
+              m_policy.space(),
+              m_functor_reducer.get_reducer().value_size() *
+                  block_size /* block_size == max block_count */));
       m_scratch_flags =
           cuda_internal_scratch_flags(m_policy.space(), sizeof(size_type));
-      m_unified_space = cuda_internal_scratch_unified(
-          m_policy.space(), m_functor_reducer.get_reducer().value_size());
+      m_unified_space =
+          reinterpret_cast<word_size_type*>(cuda_internal_scratch_unified(
+              m_policy.space(), m_functor_reducer.get_reducer().value_size()));
 
       // REQUIRED ( 1 , N , 1 )
       const dim3 block(1, block_size, 1);
--- a/core/src/HIP/Kokkos_HIP_Parallel_MDRange.hpp
+++ b/core/src/HIP/Kokkos_HIP_Parallel_MDRange.hpp
@@ -188,6 +188,23 @@ class ParallelReduce<CombinedFunctorReducerType,
   using functor_type   = FunctorType;
   using size_type      = HIP::size_type;
 
+  // Conditionally set word_size_type to int16_t or int8_t if value_type is
+  // smaller than int32_t (Kokkos::HIP::size_type)
+  // word_size_type is used to determine the word count, shared memory buffer
+  // size, and global memory buffer size before the reduction is performed.
+  // Within the reduction, the word count is recomputed based on word_size_type
+  // and when calculating indexes into the shared/global memory buffers for
+  // performing the reduction, word_size_type is used again.
+  // For scalars > 4 bytes in size, indexing into shared/global memory relies
+  // on the block and grid dimensions to ensure that we index at the correct
+  // offset rather than at every 4 byte word; such that, when the join is
+  // performed, we have the correct data that was copied over in chunks of 4
+  // bytes.
+  static_assert(sizeof(size_type) == 4);
+  using word_size_type = std::conditional_t<
+      sizeof(value_type) < 4,
+      std::conditional_t<sizeof(value_type) == 2, int16_t, int8_t>, size_type>;
+
   // Algorithmic constraints: blockSize is a power of two AND blockDim.y ==
   // blockDim.z == 1
 
@@ -195,7 +212,7 @@ class ParallelReduce<CombinedFunctorReducerType,
   const Policy m_policy;  // used for workrange and nwork
   const pointer_type m_result_ptr;
   const bool m_result_ptr_device_accessible;
-  size_type* m_scratch_space;
+  word_size_type* m_scratch_space;
   size_type* m_scratch_flags;
 
   using DeviceIteratePattern = typename Kokkos::Impl::Reduce::DeviceIterateTile<
@@ -210,13 +227,14 @@ class ParallelReduce<CombinedFunctorReducerType,
   inline __device__ void operator()() const {
     const ReducerType& reducer = m_functor_reducer.get_reducer();
 
-    const integral_nonzero_constant<
-        size_type, ReducerType::static_value_size() / sizeof(size_type)>
-        word_count(reducer.value_size() / sizeof(size_type));
+    const integral_nonzero_constant<word_size_type,
+                                    ReducerType::static_value_size() /
+                                        sizeof(word_size_type)>
+        word_count(reducer.value_size() / sizeof(word_size_type));
 
     {
       reference_type value = reducer.init(reinterpret_cast<pointer_type>(
-          kokkos_impl_hip_shared_memory<size_type>() +
+          kokkos_impl_hip_shared_memory<word_size_type>() +
           threadIdx.y * word_count.value));
 
       // Number of blocks is bounded so that the reduction can be limited to two
@@ -232,15 +250,17 @@ class ParallelReduce<CombinedFunctorReducerType,
     // Problem: non power-of-two blockDim
     if (::Kokkos::Impl::hip_single_inter_block_reduce_scan<false>(
             reducer, blockIdx.x, gridDim.x,
-            kokkos_impl_hip_shared_memory<size_type>(), m_scratch_space,
+            kokkos_impl_hip_shared_memory<word_size_type>(), m_scratch_space,
             m_scratch_flags)) {
       // This is the final block with the final result at the final threads'
       // location
-      size_type* const shared = kokkos_impl_hip_shared_memory<size_type>() +
-                                (blockDim.y - 1) * word_count.value;
-      size_type* const global = m_result_ptr_device_accessible
-                                    ? reinterpret_cast<size_type*>(m_result_ptr)
-                                    : m_scratch_space;
+      word_size_type* const shared =
+          kokkos_impl_hip_shared_memory<word_size_type>() +
+          (blockDim.y - 1) * word_count.value;
+      word_size_type* const global =
+          m_result_ptr_device_accessible
+              ? reinterpret_cast<word_size_type*>(m_result_ptr)
+              : m_scratch_space;
 
       if (threadIdx.y == 0) {
         reducer.final(reinterpret_cast<value_type*>(shared));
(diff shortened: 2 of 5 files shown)


PhilMiller:
I'm going to temporarily push a test change to confirm that it fails the HIP build, and then force-push it back.

PhilMiller:
HIP builds with just the assertion but no added test triggering it passed:
https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/Kokkos/detail/Kokkos/11627/pipeline/50#step-231-log-2274

PhilMiller:
The added assertion does trigger with a test modified to call an offending `parallel_reduce`
https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/Kokkos/detail/Kokkos/11630/pipeline#step-65-log-901

However, it also reminds me that the guardrail may be necessary for all possible memory space template arguments of `LAnd`

PhilMiller:
Since I was typo-fixing anyway, I ran ispell over all comments and strings in that file. Everything else is fine.

PhilMiller:
I've only guarded against what's been reported to fail. 

If you can point me at a workable set of modules to test things on kokkos-dev-2 or wherever, I can try it out.

core/src/HIP/Kokkos_HIP_Parallel_MDRange.hpp
```diff
@@ -168,6 +168,18 @@ class ParallelFor<FunctorType, Kokkos::MDRangePolicy<Traits...>, HIP> {
 template <class FunctorType, class ReducerType, class... Traits>
 class ParallelReduce<FunctorType, Kokkos::MDRangePolicy<Traits...>, ReducerType,
                      HIP> {
+  // FIXME HIP #5641
+  static_assert(
+      !std::is_same_v<ReducerType, ::Kokkos::LAnd<bool, ::Kokkos::HostSpace>> &&
```
Rombur, line 173: Why `HostSpace`?

masterleinad (dismissed):

PhilMiller:
> Why only for LAnd? I bet all other reductions would fail too for bool? (Sum, BAnd, LOr, etc. )

It looks like `Kokkos::reduction_identity<bool>` only defines `lor()` and `land()`, so none of the others are possible to meaningfully test as such.

That said, I've reproduced the (a?) failure for `LOr` as well, again with `MDRange`. I'm going to guard off any use `value_type == bool` for HIP `MDRange`.

PhilMiller:
I just tested with `char`, `unsigned char`, `signed char`, `int8_t`,  `uint8_t`, `int16_t`, and `uint16_t`, and those all fail as well. Going up to 32 bits passes across a sweep of `MDRange` sizes.

That suggests the bug is in some sort of sizing or alignment.

PhilMiller:
I adapted the changes in #5333 from `TeamPolicy` to `MDRange` and found that they fixed the failing tests across the range of types smaller than `int`.

I'm committing the sloppy mess now because it's late and I want to get it out. I'll make tests for Team reductions and move them to a sensible place and maybe fix them tomorrow.

PhilMiller:
Test passes on Serial but fails on HIP before the fix. Test passes with the fix.

Will clang-format, then this is ready.

PhilMiller:
Lol, once I got nvcc to compile the damn test, it fails on CUDA in the same way! Fixing that, too

core/src/Cuda/Kokkos_Cuda_Parallel_MDRange.hpp
```diff
@@ -212,16 +212,33 @@ class ParallelReduce<CombinedFunctorReducerType,
...
+  // performing the reduction, word_size_type is used again.
+  // For scalars > 4 bytes in size, indexing into shared/global memory relies
+  // on the block and grid dimensions to ensure that we index at the correct
+  // offset rather than at every 4 byte word; such that, when the join is
+  // performed, we have the correct data that was copied over in chunks of 4
+  // bytes.
+  using word_size_type = std::conditional_t<
+      sizeof(value_type) < sizeof(Kokkos::Cuda::size_type),
+      std::conditional_t<sizeof(value_type) == 2, int16_t, int8_t>,
+      Kokkos::Cuda::size_type>;
```
dalg24, line 230:
```suggestion
      sizeof(value_type) < sizeof(size_type),
      std::conditional_t<sizeof(value_type) == 2, int16_t, int8_t>,
      size_type>;
```

core/unit_test/TestMDRangeReduce.hpp
```diff
@@ -0,0 +1,61 @@
...
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
+#include <gtest/gtest.h>
+
+#include <Kokkos_Core.hpp>
+
+namespace Test {
```
dalg24, line 21:
```suggestion
namespace {
```

core/unit_test/TestMDRangeReduce.hpp
```diff
@@ -0,0 +1,61 @@
...
+    Kokkos::parallel_reduce(
+        policy_MD, KOKKOS_LAMBDA(const int, const int, T& res) { res = 1; },
+        Kokkos::LAnd<T>(land_MD));
+    EXPECT_EQ(land_MD, 1);
+  }
+}
+
+TEST(TEST_CATEGORY, mdrange_parallel_reduce_primitive_types) {
+  for (int bound = 0; bound < 100; bound++) {
+    for (int k = 0; k < bound; ++k) {
```
dalg24, line 50: How long does it take?

core/unit_test/TestMDRangeReduce.hpp
```diff
@@ -0,0 +1,61 @@
...
+      MDRangeReduceTester<bool>(bound, k);
+      MDRangeReduceTester<signed char>(bound, k);
+      MDRangeReduceTester<int8_t>(bound, k);
+      MDRangeReduceTester<int16_t>(bound, k);
```
dalg24, line 54: do we care to test for a user-defined type of size 3?

core/unit_test/TestMDRangeReduce.hpp
```diff
@@ -0,0 +1,61 @@
...
+void MDRangeReduceTester(int bound, int k) {
+  const auto policy_MD = Kokkos::MDRangePolicy<Kokkos::Rank<2>, TEST_EXECSPACE>(
+      {0, 0}, {bound, 2});
+
+  // No explicit fence() calls needed because result is in HostSpace
+  {
+    T lor_MD = 0;
+    Kokkos::parallel_reduce(
+        policy_MD,
+        KOKKOS_LAMBDA(const int i, const int, T& res) { res = res || i == k; },
```
dalg24, line 33: Might need to guard for LAMBDA support enabled.

core/unit_test/TestMDRangeReduce.hpp
```diff
@@ -0,0 +1,61 @@
...
+namespace Test {
+
+template <typename T>
+void MDRangeReduceTester(int bound, int k) {
+  const auto policy_MD = Kokkos::MDRangePolicy<Kokkos::Rank<2>, TEST_EXECSPACE>(
+      {0, 0}, {bound, 2});
+
+  // No explicit fence() calls needed because result is in HostSpace
+  {
+    T lor_MD = 0;
```
dalg24, line 30: Why did you go for logical OR and logical AND?  Presumably that bug can show on other somewhat less obscure reductions right?
(Not necessarily asking you to change, just trying to understand if there is a rational)

core/unit_test/TestMDRangeReduce.hpp
e10harvey: We test this via `reducers_point_t` with a point_t type in TestReducers_d.hpp, if you want to lift that code.

core/unit_test/TestMDRangeReduce.hpp
PhilMiller: The initial bug report was for reducing `LAnd` over `bool`. That just turned out to be a special case of `size_of(value_type) < sizeof(int)`

core/unit_test/TestMDRangeReduce.hpp
PhilMiller: a bit under a second

core/unit_test/TestMDRangeReduce.hpp
PhilMiller: Probably a good idea, but that would require at least some customization of the test. I'm getting really tired of staring at this one at the moment. I already plan to make the same fix for the Team policies, now that it's clearer that it really will be needed

core/src/HIP/Kokkos_HIP_Parallel_MDRange.hpp
```diff
@@ -188,15 +188,32 @@ class ParallelReduce<CombinedFunctorReducerType,
...
   // blockDim.z == 1
 
   const CombinedFunctorReducerType m_functor_reducer;
   const Policy m_policy;  // used for workrange and nwork
   const pointer_type m_result_ptr;
   const bool m_result_ptr_device_accessible;
-  size_type* m_scratch_space;
-  size_type* m_scratch_flags;
+  word_size_type* m_scratch_space;
+  HIP::size_type* m_scratch_flags;
```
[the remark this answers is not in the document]
PhilMiller: How it was written in the `RangePolicy` case. I'm happy to simplify these out.

core/unit_test/TestMDRangeReduce.hpp
PhilMiller: Earnest question: why is that still optional?

core/src/Cuda/Kokkos_Cuda_Parallel_MDRange.hpp
```diff
@@ -212,16 +212,32 @@ class ParallelReduce<CombinedFunctorReducerType,
...
+  // Within the reduction, the word count is recomputed based on word_size_type
+  // and when calculating indexes into the shared/global memory buffers for
+  // performing the reduction, word_size_type is used again.
+  // For scalars > 4 bytes in size, indexing into shared/global memory relies
+  // on the block and grid dimensions to ensure that we index at the correct
+  // offset rather than at every 4 byte word; such that, when the join is
+  // performed, we have the correct data that was copied over in chunks of 4
+  // bytes.
+  using word_size_type = std::conditional_t<
+      sizeof(value_type) < sizeof(size_type),
```
dalg24, line 228:
```suggestion
  static_assert(sizeof(size_type) == 4);
  using word_size_type = std::conditional_t<
      sizeof(value_type) < 4,
```

core/src/HIP/Kokkos_HIP_Parallel_MDRange.hpp
```diff
@@ -188,14 +188,30 @@ class ParallelReduce<CombinedFunctorReducerType,
...
+  // Within the reduction, the word count is recomputed based on word_size_type
+  // and when calculating indexes into the shared/global memory buffers for
+  // performing the reduction, word_size_type is used again.
+  // For scalars > 4 bytes in size, indexing into shared/global memory relies
+  // on the block and grid dimensions to ensure that we index at the correct
+  // offset rather than at every 4 byte word; such that, when the join is
+  // performed, we have the correct data that was copied over in chunks of 4
+  // bytes.
+  using word_size_type = std::conditional_t<
+      sizeof(value_type) < sizeof(size_type),
```
dalg24, line 204:
```suggestion
  // bytes.
  static_assert(sizeof(size_type) == 4);
  using word_size_type = std::conditional_t<
      sizeof(value_type) < 4,
```

core/unit_test/TestMDRangeReduce.hpp
dalg24: I think technically if one configure with `-DKokkos_ENABLE_CUDA_LAMBDA=OFF` we do not pass the `-extended-lambda` flag to NVCC

core/unit_test/TestMDRangeReduce.hpp
```diff
@@ -0,0 +1,63 @@
...
+    Kokkos::parallel_reduce(
+        policy_MD,
+        KOKKOS_LAMBDA(const int i, const int, T& res) { res = res || i == k; },
+        Kokkos::LOr<T>(lor_MD));
+    EXPECT_EQ(lor_MD, 1);
```
dalg24, line 35: Consider putting the type and the arguments value into the stream from the assertion to make failure messages more useful.

core/unit_test/TestMDRangeReduce.hpp
dalg24: It feels excessive.  Ideally, tests should be as small and fast running as can be while still doing the job.
1 second is actually quite a lot for a single unit test.

core/src/Cuda/Kokkos_Cuda_Parallel_MDRange.hpp
```diff
@@ -342,12 +360,12 @@ class ParallelReduce<CombinedFunctorReducerType,
                        : suggested_blocksize;  // Note: block_size must be less
                                                // than or equal to 512
 
-      m_scratch_space = cuda_internal_scratch_space(
+      m_scratch_space = (word_size_type*)cuda_internal_scratch_space(
           m_policy.space(), m_functor_reducer.get_reducer().value_size() *
                                 block_size /* block_size == max block_count */);
```
masterleinad, line 365:
```suggestion
      m_scratch_space = reinterpret_cast<word_size_type*>(cuda_internal_scratch_space(
          m_policy.space(), m_functor_reducer.get_reducer().value_size() *
                                block_size /* block_size == max block_count */));
```

core/src/Cuda/Kokkos_Cuda_Parallel_MDRange.hpp
```diff
@@ -342,12 +360,12 @@ class ParallelReduce<CombinedFunctorReducerType,
...
       m_scratch_flags =
           cuda_internal_scratch_flags(m_policy.space(), sizeof(size_type));
-      m_unified_space = cuda_internal_scratch_unified(
+      m_unified_space = (word_size_type*)cuda_internal_scratch_unified(
           m_policy.space(), m_functor_reducer.get_reducer().value_size());
```
masterleinad, line 369:
```suggestion
      m_unified_space = reinterpret_cast<word_size_type*>(cuda_internal_scratch_unified(
          m_policy.space(), m_functor_reducer.get_reducer().value_size()));
```

core/src/HIP/Kokkos_HIP_Parallel_MDRange.hpp
```diff
@@ -294,7 +313,7 @@ class ParallelReduce<CombinedFunctorReducerType,
                        : suggested_blocksize;  // Note: block_size must be less
                                                // than or equal to 512
 
-      m_scratch_space = hip_internal_scratch_space(
+      m_scratch_space = (word_size_type*)hip_internal_scratch_space(
           m_policy.space(), reducer.value_size() *
                                 block_size /* block_size == max block_count */);
```
masterleinad, line 318:
```suggestion
      m_scratch_space = (reinterpret_cast<word_size_type*>(hip_internal_scratch_space(
          m_policy.space(), reducer.value_size() *
                                block_size /* block_size == max block_count */));
```

core/unit_test/TestMDRangeReduce.hpp
masterleinad: https://github.com/kokkos/kokkos/blob/fc4a9cecfa4d309881847626ebff718bbbe6af8c/cmake/kokkos_arch.cmake#L169-L174

PhilMiller: So, it's optional because Nvidia still considered it 'experimental'?

In CUDA 12, it's not experimental anymore:
https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html
Would we be averse to a PR that deprecates the CMake option, and unconditionally adds the version-appropriate flag and the `#define`?

Is there any benefit to users at this point of being able to *disable* it?

core/unit_test/TestMDRangeReduce.hpp
PhilMiller: Ok. I'll change it to a smattering of values across a broad range.

PhilMiller:
I'm heading out for a trip over the weekend. I'll do all the remaining cleanup by/on Monday.

PhilMiller:
LOL, OpenMPTarget fails this test too!

core/unit_test/TestMDRangeReduce.hpp
```diff
@@ -0,0 +1,63 @@
...
+//@HEADER
+
+#include <gtest/gtest.h>
+
+#include <Kokkos_Core.hpp>
+
+namespace {
+
+template <typename T>
+void MDRangeReduceTester(int bound, int k) {
```
PhilMiller, line 24: Lol, `icpc`
```
/__w/kokkos/kokkos/core/unit_test/TestMDRangeReduce.hpp(24): error #869: parameter "bound" was never referenced
  void MDRangeReduceTester(int bound, int k) {
                               ^
```

What the hell do we even do with that? Mark it `[[maybe_unused]]` and pretend that this compiler is not bonkers?

PhilMiller:
Per @rgayatri23 I'm going to skip the test of OpenMPTarget for now.

core/unit_test/TestMDRangeReduce.hpp
masterleinad: You can bring it up in the developer meeting. 🙂

PhilMiller: #5964

core/unit_test/TestMDRangeReduce.hpp
PhilMiller: It now runs in less than a millisecond on Serial and Cuda

core/unit_test/TestMDRangeReduce.hpp
```diff
@@ -0,0 +1,66 @@
...
+    EXPECT_EQ(land_MD, 1);
+  }
+}
+
+TEST(TEST_CATEGORY, mdrange_parallel_reduce_primitive_types) {
+#if defined(KOKKOS_ENABLE_OPENMPTARGET)
+  GTEST_SKIP() << "FIXME OPENMPTARGET Tests of MDRange reduce over values "
+                  "smaller than int would fail";
+#elif !defined(KOKKOS_ENABLE_CUDA) || defined(KOKKOS_ENABLE_CUDA_LAMBDA)
+  for (int bound : {0, 1, 7, 32, 65, 70000}) {
```
PhilMiller, line 53: Grrr, the large value here seems to be long enough to time out a serial debug build.
```suggestion
  for (int bound : {0, 1, 7, 32, 65, 7000}) {
```

PhilMiller:
Rebased on `develop` with Github CI (hopefully) fixed

Rombur (approved):

core/unit_test/TestMDRangeReduce.hpp
```diff
@@ -0,0 +1,66 @@
...
+        Kokkos::LAnd<T>(land_MD));
+    EXPECT_EQ(land_MD, 1);
+  }
+}
+
+TEST(TEST_CATEGORY, mdrange_parallel_reduce_primitive_types) {
+#if defined(KOKKOS_ENABLE_OPENMPTARGET)
+  GTEST_SKIP() << "FIXME OPENMPTARGET Tests of MDRange reduce over values "
+                  "smaller than int would fail";
+#elif !defined(KOKKOS_ENABLE_CUDA) || defined(KOKKOS_ENABLE_CUDA_LAMBDA)
```
dalg24, line 52: Prefer GTEST_SKIP if the test does not run

PhilMiller, line 52:
```suggestion
#elif !defined(KOKKOS_ENABLE_CUDA) || defined(KOKKOS_ENABLE_CUDA_LAMBDA)
  GTEST_SKIP() << "Skipped ENABLE_CUDA_LAMBDA";
#else
```

core/unit_test/TestMDRangeReduce.hpp
```diff
@@ -0,0 +1,68 @@
...
```
PhilMiller, line 52:
```suggestion
#elif defined(KOKKOS_ENABLE_CUDA) && !defined(KOKKOS_ENABLE_CUDA_LAMBDA)
```
Oops, reversed the condition

dalg24 (approved):

PR #5755, opened 2023-01-11
Fix HIP Global Launch with HSA_XNACK=1

Rombur:
This problem is found with all the versions of ROCm we support.

cc: @arghdos

### DIFF
--- a/core/src/HIP/Kokkos_HIP_Instance.cpp
+++ b/core/src/HIP/Kokkos_HIP_Instance.cpp
@@ -260,17 +260,19 @@ Kokkos::HIP::size_type *HIPInternal::scratch_functor(
 
 Kokkos::HIP::size_type *HIPInternal::scratch_functor_host(
     const std::size_t size) const {
-  if (verify_is_initialized("scratch_functor_host") && m_scratchFunctorSize < size) {
+  if (verify_is_initialized("scratch_functor_host") &&
+      m_scratchFunctorSize < size) {
     m_scratchFunctorSize = size;
 
-    using Record = Kokkos::Impl::SharedAllocationRecord<Kokkos::HIPHostPinnedSpace, void>;
+    using Record =
+        Kokkos::Impl::SharedAllocationRecord<Kokkos::HIPHostPinnedSpace, void>;
 
     if (m_scratchFunctorHost)
       Record::decrement(Record::get_record(m_scratchFunctorHost));
 
-    Record *const r =
-        Record::allocate(Kokkos::HIPHostPinnedSpace(), "Kokkos::InternalScratchFunctorHost",
-                         m_scratchFunctorSize);
+    Record *const r = Record::allocate(Kokkos::HIPHostPinnedSpace(),
+                                       "Kokkos::InternalScratchFunctorHost",
+                                       m_scratchFunctorSize);
 
     Record::increment(r);
 
--- a/core/src/HIP/Kokkos_HIP_Instance.cpp
+++ b/core/src/HIP/Kokkos_HIP_Instance.cpp
@@ -258,6 +258,28 @@ Kokkos::HIP::size_type *HIPInternal::scratch_functor(
   return m_scratchFunctor;
 }
 
+Kokkos::HIP::size_type *HIPInternal::scratch_functor_host(
+    const std::size_t size) const {
+  if (verify_is_initialized("scratch_functor_host") && m_scratchFunctorSize < size) {
+    m_scratchFunctorSize = size;
+
+    using Record = Kokkos::Impl::SharedAllocationRecord<Kokkos::HIPHostPinnedSpace, void>;
+
+    if (m_scratchFunctorHost)
+      Record::decrement(Record::get_record(m_scratchFunctorHost));
+
+    Record *const r =
+        Record::allocate(Kokkos::HIPHostPinnedSpace(), "Kokkos::InternalScratchFunctorHost",
+                         m_scratchFunctorSize);
+
+    Record::increment(r);
+
+    m_scratchFunctorHost = reinterpret_cast<size_type *>(r->data());
+  }
+
+  return m_scratchFunctor;
+}
+
 int HIPInternal::acquire_team_scratch_space() {
   int current_team_scratch = 0;
   int zero                 = 0;
--- a/core/src/HIP/Kokkos_HIP_Instance.cpp
+++ b/core/src/HIP/Kokkos_HIP_Instance.cpp
@@ -236,52 +236,6 @@ Kokkos::HIP::size_type *HIPInternal::scratch_flags(const std::size_t size) {
   return m_scratchFlags;
 }
 
-Kokkos::HIP::size_type *HIPInternal::scratch_functor(
-    const std::size_t size) const {
-  if (verify_is_initialized("scratch_functor") && m_scratchFunctorSize < size) {
-    m_scratchFunctorSize = size;
-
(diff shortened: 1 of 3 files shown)


core/src/HIP/Kokkos_HIP_Instance.cpp
```diff
@@ -258,6 +258,28 @@ Kokkos::HIP::size_type *HIPInternal::scratch_functor(
...
+    Record *const r =
+        Record::allocate(Kokkos::HIPHostPinnedSpace(), "Kokkos::InternalScratchFunctorHost",
+                         m_scratchFunctorSize);
+
+    Record::increment(r);
+
+    m_scratchFunctorHost = reinterpret_cast<size_type *>(r->data());
+  }
+
+  return m_scratchFunctor;
```
dalg24, line 282:
```suggestion
  return m_scratchFunctorHost;
```

dalg24 (changes requested):

core/src/HIP/Kokkos_HIP_Instance.cpp
```diff
@@ -258,6 +258,30 @@ Kokkos::HIP::size_type *HIPInternal::scratch_functor(
   return m_scratchFunctor;
 }
 
+Kokkos::HIP::size_type *HIPInternal::scratch_functor_host(
+    const std::size_t size) const {
+  if (verify_is_initialized("scratch_functor_host") &&
+      m_scratchFunctorSize < size) {
+    m_scratchFunctorSize = size;
```
crtrott, line 265: args this is actually wrong. If you first call `scratch_functor()` and then `scratch_functor_host()` the first call will reset m_scratchFunctorSize so the second will not do anything ...

crtrott (changes requested):

core/src/HIP/Kokkos_HIP_Instance.cpp
crtrott: haha this is why the other issue didn't crash the code. `m_scratchFunctorHost` was always null lol, but this thing made it not crash ...

crtrott:
Maybe we should just have a function which encapsulates the staging. I.e. you hand that member function of the instant the driver, and it returns device ptr.

Rombur:
I have updated the PR

core/src/HIP/Kokkos_HIP_Instance.hpp
```diff
@@ -144,6 +147,50 @@ class HIPInternal {
   void release_team_scratch_space(int scratch_pool_id);
 };
 
+template <typename DriverType>
+Kokkos::HIP::size_type *HIPInternal::scratch_functor(
+    DriverType const &driver) const {
```
dalg24, line 152: We should probably give it a better name now as it does not really capture that this function is copying the driver into a staging area in host pinned memory and returns a pointer to the device where the driver is being copied asynchronously.
Maybe a `transfer_to_device_and_get_pointer`?  Unless someone has a better idea because I am not super proud about that name...

dalg24 (commented):
Looks good besides the name of the member function.

core/src/HIP/Kokkos_HIP_Instance.hpp
Rombur: I agree that the current name is not great but I am not sure `transfer_to_device_and_get_pointer` is better. Unfortunately, I don't have a good name for this functions :-/

dalg24:
Sorry just  realized making the function a template is not a good idea.
Please take void* and size instead. It will save on number of instantiations

core/src/HIP/Kokkos_HIP_Instance.hpp
crtrott: `stage_functor` or `stage_functor_for_execution` ?

core/src/HIP/Kokkos_HIP_Instance.hpp
```diff
@@ -19,6 +19,7 @@
 #ifndef KOKKOS_HIP_INSTANCE_HPP
 #define KOKKOS_HIP_INSTANCE_HPP
 
+#include <HIP/Kokkos_HIP_SharedAllocationRecord.hpp>
```
dalg24, line 22:
```suggestion
```

dalg24 (approved):
Please drop the stray header include

core/src/HIP/Kokkos_HIP_KernelLaunch.hpp
```diff
@@ -378,11 +378,8 @@ struct HIPParallelLaunchKernelInvoker<DriverType, LaunchBounds,
                             dim3 const &block, int shmem,
                             HIPInternal const *hip_instance) {
     DriverType *driver_ptr = reinterpret_cast<DriverType *>(
-        hip_instance->scratch_functor(sizeof(DriverType)));
-
-    hipMemcpyAsync(driver_ptr, &driver, sizeof(DriverType), hipMemcpyDefault,
-                   hip_instance->m_stream);
-
+        hip_instance->stage_functor_for_execution(
```
arghdos, line 381: Does all of this need a lock, as in the constant memory launcher to avoid multiple threads stomping on the same resources?

dalg24: I had the same thought when we discussed it to Bruno but (wrongly) assumed it was locked from the call site.
Good catch

core/src/HIP/Kokkos_HIP_Instance.cpp
```diff
@@ -236,25 +236,45 @@ Kokkos::HIP::size_type *HIPInternal::scratch_flags(const std::size_t size) {
...
+    m_scratchFunctor     = reinterpret_cast<size_type *>(r->data());
+    m_scratchFunctorHost = reinterpret_cast<size_type *>(r_host->data());
   }
 
+  // When using HSA_XNACK=1, it is necessary to copy the driver to the host to
+  // ensure that the driver is not destroyed before the computation is done.
+  // Without this fix, all the atomic tests fail. It is not obvious that this
+  // problem is limited to HSA_XNACK=1 even if all the tests pass when
+  // HSA_XNACK=0. That's why we always copy the driver.
+  KOKKOS_IMPL_HIP_SAFE_CALL(hipStreamSynchronize(m_stream));
```
arghdos, line 272: I might add a note in the comment that the `hipMemcpyAsync` below should guard from anyone overwriting the device `scratchFunctor` while the current kernel is executing. My read is that this synchronization actually functions by guarding anyone from copying to the `scratchFunctorHost` while is still possibly needed for the `hipMemcpyAsync`.

I am not sure I see a better way to ensure this, but it might be useful context if we ever have to monkey with this again.

core/src/HIP/Kokkos_HIP_KernelLaunch.hpp
```diff
@@ -377,12 +377,11 @@ struct HIPParallelLaunchKernelInvoker<DriverType, LaunchBounds,
   static void invoke_kernel(DriverType const &driver, dim3 const &grid,
                             dim3 const &block, int shmem,
                             HIPInternal const *hip_instance) {
+    // Wait until the previous kernel that uses m_scratchFuntor is dne
```
masterleinad, line 380:
```suggestion
    // Wait until the previous kernel that uses m_scratchFuntor is done
```

dalg24 (approved):

crtrott (approved):

dalg24:
Failure was clearly unrelated (no space left on device) and both HIP builds passed

PR #5756, opened 2023-01-12
Desul atomics fixup detect use of SYCL

dalg24:
Related to https://github.com/kokkos/kokkos/pull/5613/files#r1067648253

I remember we only want to use SYCL atomics only on the device side but it is not obvious to me that the macro guard is right.


masterleinad (approved):
I remember problems with this when enabling `OpenMP` and `SYCL` but it seems to work now (possibly due to the refactoring?) on the testbeds.

crtrott (approved):

PR #5757, opened 2023-01-12
Desul atomics drop cuda arch macro guards

dalg24:
Drive-by fixes as I an snooping around in Desul atomics


dalg24:
Failure (running out of disk space in the SYCL build) is unrelated.
Will submit this patch on the Desul side.

crtrott (approved):

rgayatri23 (approved):

PR #5759, opened 2023-01-12
CMake: update package COMPATIBILITY mode

dalg24:
In #5751 we did not anticipate that `find_package(Kokkos 3.7)` would not accept the newer Kokkos version
```
CMake Error at CMakeLists.txt:4 (find_package):
  Could not find a configuration file for package "Kokkos" that is compatible
  with requested version "3.7".

  The following configuration files were considered but not accepted:

    <prefix>/lib/cmake/Kokkos/KokkosConfig.cmake, version: 4.0.99
```
One way to resolve this would be to change the COMPATIBILITY mode when we generate our package version file

### DIFF
--- a/cmake/kokkos_install.cmake
+++ b/cmake/kokkos_install.cmake
@@ -19,7 +19,7 @@ IF (NOT KOKKOS_HAS_TRILINOS AND NOT Kokkos_INSTALL_TESTING)
 
   WRITE_BASIC_PACKAGE_VERSION_FILE("${Kokkos_BINARY_DIR}/KokkosConfigVersion.cmake"
       VERSION "${Kokkos_VERSION}"
-      COMPATIBILITY SameMajorVersion)
+      COMPATIBILITY AnyNewerVersion)
 
   # Install the KokkosConfig*.cmake files
   install(FILES


dalg24:
In case you also wonder how on earth our CI did not catch that before
```
BUILD.md:14:find_package(Kokkos REQUIRED)
cmake/KokkosConfig.cmake.in:32:    # if find_package(Kokkos COMPONENTS launch_compiler) then rely on the
cmake/README.md:42:find_package(Kokkos)
example/build_cmake_installed/CMakeLists.txt:10:find_package(Kokkos REQUIRED)
example/build_cmake_installed_different_compiler/CMakeLists.txt:15:find_package(Kokkos REQUIRED COMPONENTS launch_compiler)
example/build_cmake_installed_kk_as_language/CMakeLists.txt:10:find_package(Kokkos COMPONENTS separable_compilation)
scripts/spack_test/CMakeLists.txt.in:3:find_package(Kokkos REQUIRED)
```
we do not use the `[version]` argument when we do `find_package(Kokkos)`

masterleinad:
Following `CMake`'s guidance on compatibility `SameMajorVersion` would be correct since we have breaking changes between 3.* and 4.0. 
On the other hand, I expect that people expect 4.0 to be compatible if they ask for a minimum of 3.*. Thus, I agree with using `AnyNewerVersion`.

masterleinad (approved):

masterleinad:
On the user side, it probably makes sense to provide ranges of `Kokkos` versions that are known to work.

fnrizzi (approved):

crtrott (approved):
Yeah lets do this until this runs into trouble.

PR #5762, opened 2023-01-12
Fix ScratchSpace pointer comparison for SYCL

### LINKED ISSUE #5761: Tests fail on Intel GPUs following scratch alignment changes
**Describe the bug**

> Looks like the changes to Kokkos_ScratchSpace.hpp in https://github.com/kokkos/kokkos/pull/5687 are breaking tests on Intel GPUs for optimization levels higher than -O1 .

1. compilers (with versions)
2. Kokkos release or commit used (i.e., the sha1 number)
3. platform and backend
4. cmake configure command
5. output from cmake command
7. command line needed to reproduce the bug
7. please also attach the `KokkosCore_config.h` header file (generated during the build);

**Any additional info**
Please provide any additional context about the issue here.

masterleinad:
Fixes #5761. The changes to `Kokkos_ScratchSpace.hpp` in https://github.com/kokkos/kokkos/pull/5687 were breaking tests on Intel GPUs for optimization levels higher than `-O1`.

I'm not entirely sure why the current version breaks. Both sides of the comparison are of type `ptrdiff_t` but trying to display the value for the right-hand side shows a ridiculously small negative number for one of the tests (-4611686018427379892 close to  `LONG_MIN`/2; `ptrdiff_t` is `long`).

edit: Instead of rewriting the comparison, casting to `uintptr_t` seems to work as well.

### DIFF
--- a/core/src/Kokkos_ScratchSpace.hpp
+++ b/core/src/Kokkos_ScratchSpace.hpp
@@ -100,7 +100,6 @@ class ScratchMemorySpace {
       int level = -1) const {
     if (level == -1) level = m_default_level;
     auto& m_iter    = (level == 0) ? m_iter_L0 : m_iter_L1;
-    auto& m_end     = (level == 0) ? m_end_L0 : m_end_L1;
     auto m_iter_old = m_iter;
     if constexpr (alignment_requested) {
       const ptrdiff_t missalign = size_t(m_iter) % alignment;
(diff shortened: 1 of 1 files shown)


crtrott:
Haha, this is one of the things Phil complained about, now every failing case is UB ....

PhilMiller (approved):

cz4rs (approved):

ldh4 (approved):

crtrott (approved):

PR #5765, opened 2023-01-12
Added enable_if conditions to hpx team parallel_reduce

ldh4:
Found that overloaded `parallel_reduce` functions with nested policies for HPX backend do not check if the input result argument is of a correct type (`ReducerType` vs `ValueType`).

Added in enable_ifs to `parallel_reduce` to check templated input argument type.


masterleinad (approved):

msimberg (approved):
Thanks @ldh4.

Are there any unit tests that _should_ be checking this (but clearly aren't...)? Do I need to opt the HPX backend in to some set of tests?

ldh4:
@msimberg I haven't found any unit test that specifically tests for this. From what I've seen so far, reducer tests with team policies either pass in `ValueType` or rvalue `ReducerType` for the reduce result, which is why tests were working even without these enable_ifs.

PR #5767, opened 2023-01-13
Fix sycl.scratch_align test

masterleinad:
Likely also broken by #5687 and related to #5762. Casting to `int` seems to be enough for it to work again.
There are also some drive-by changes using more appropriate`ASSERT_*` macros in some places.

### DIFF
--- a/core/unit_test/TestTeam.hpp
+++ b/core/unit_test/TestTeam.hpp
@@ -1642,9 +1642,12 @@ struct TestScratchAlignment {
           // and since scratch_ptr3 is then already aligned it difference
           // should match that
           if ((scratch_ptr3 - scratch_ptr2) != 32) flag() = 1;
+
           // check actually alignment of ptrs is as requested
-          if (((scratch_ptr1 % 4) != 0) || ((scratch_ptr2 % 8) != 0) ||
-              ((scratch_ptr3 % 4) != 0))
+          // cast to int here to avoid failure with icpx in mixed integer type
+          // comparison
+          if ((int(scratch_ptr1 % 4) != 0) || (int(scratch_ptr2 % 8) != 0) ||
+              (int(scratch_ptr3 % 4) != 0))
             flag() = 1;
         });
     Kokkos::fence();


dalg24 (dismissed):

core/unit_test/TestTeam.hpp
```diff
@@ -1642,15 +1642,16 @@ struct TestScratchAlignment {
           // and since scratch_ptr3 is then already aligned it difference
           // should match that
           if ((scratch_ptr3 - scratch_ptr2) != 32) flag() = 1;
+
           // check actually alignment of ptrs is as requested
-          if (((scratch_ptr1 % 4) != 0) || ((scratch_ptr2 % 8) != 0) ||
-              ((scratch_ptr3 % 4) != 0))
+          if ((int(scratch_ptr1 % 4) != 0) || (int(scratch_ptr2 % 8) != 0) ||
+              (int(scratch_ptr3 % 4) != 0))
```
PhilMiller, line 1648: If they're all `intptr_t`, why do they need to be cast to `int` here?

PhilMiller: Is this just about the compiler mis-behaving?

masterleinad: Yes.

masterleinad:
> How does it relate to/interact with https://github.com/kokkos/kokkos/pull/5762? Please elaborate

I need both for this test to pass with the default compiler drop on the testbeds.

core/unit_test/TestTeam.hpp
crtrott: bloody hell, hate to do that to you but can you add a comment to that effect: i.e. we need to still cast to int here because SYCL compiler fails in mixed integer type comparisons with intptr_t ...

Also next week can we create a reproducer for Intel??

crtrott (dismissed):

core/unit_test/TestTeam.hpp
dalg24: what about a macro `KOKKOS_IMPL_SYCL_CAST_TO_INT`

dalg24: or some other ugly name that we only define when needed

masterleinad:
> Also next week can we create a reproducer for Intel??

Apparently, this doesn't show up in a later compiler version (talking with Intel).

dalg24 (changes requested):
Drive-by changes are totally unrelated, move them to another PR

masterleinad:
Rebased to remove the drive-by changes.

core/unit_test/TestTeam.hpp
dalg24: ping

masterleinad: I added a comment, see https://github.com/kokkos/kokkos/pull/5767/files#diff-f1017e3e2c8e0566e5f1e0277a0150db93555c894cace36cb32aef5f75efb4ccR1647.

dalg24 (approved):
Would have preferred macro but it is good enough

JBludau (approved):
I would consider a Macro more versatile for the future, but looks okay for a fix

Rombur (approved):

core/unit_test/TestTeam.hpp
```diff
@@ -1642,9 +1642,12 @@ struct TestScratchAlignment {
           // and since scratch_ptr3 is then already aligned it difference
```
ajpowelsnl, line 1642: Comment is not very understandable; by, "...and since scratch_ptr3 is already aligned, its difference should match that," do you mean that the scratch_ptr3 shmem allocation, by virtue of being a multiple of 32, will be aligned?

ajpowelsnl (approved):

core/unit_test/TestTeam.hpp
masterleinad: I didn't touch that.

ajpowelsnl: OK, but can we fix it up to make sense (assuming we understand the original intent)?  Just aiming for improvement, while we're in the neighborhood

masterleinad: See https://github.com/kokkos/kokkos/pull/5800.

masterleinad:
@crtrott ping

crtrott (approved):

PR #5770, opened 2023-01-13
Replace HIP_LOCK_ARRAYS macros by functions

masterleinad:
The corresponding pull request on the `desul`-side was merged (https://github.com/desul/desul/pull/88).
The purpose was to get rid of more obscure macros and to make the `CUDA` and `HIP` implementations more similar (again).

### DIFF
--- a/core/src/HIP/Kokkos_HIP_Instance.cpp
+++ b/core/src/HIP/Kokkos_HIP_Instance.cpp
@@ -375,14 +375,6 @@ Kokkos::HIP::size_type *hip_internal_scratch_flags(const HIP &instance,
 
 namespace Kokkos {
 namespace Impl {
-void hip_device_synchronize(const std::string &name) {
-  Kokkos::Tools::Experimental::Impl::profile_fence_event<Kokkos::HIP>(
-      name,
-      Kokkos::Tools::Experimental::SpecialSynchronizationCases::
-          GlobalDeviceSynchronization,
-      [&]() { KOKKOS_IMPL_HIP_SAFE_CALL(hipDeviceSynchronize()); });
-}
-
 void hip_internal_error_throw(hipError_t e, const char *name, const char *file,
                               const int line) {
   std::ostringstream out;
--- a/core/src/HIP/Kokkos_HIP_Locks.cpp
+++ b/core/src/HIP/Kokkos_HIP_Locks.cpp
@@ -67,8 +67,6 @@ void initialize_host_hip_lock_arrays() {
   copy_hip_lock_arrays_to_device();
   init_lock_array_kernel_atomic<<<
       (KOKKOS_IMPL_HIP_SPACE_ATOMIC_MASK + 1 + 255) / 256, 256, 0, nullptr>>>();
-  HIP::impl_static_fence(
-      "Kokkos::Impl::initialize_host_hip_lock_arrays: Post Init Lock Arrays");
 }
 
 void finalize_host_hip_lock_arrays() {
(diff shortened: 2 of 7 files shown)


core/src/HIP/Kokkos_HIP_Instance.cpp
```diff
@@ -375,14 +375,6 @@ Kokkos::HIP::size_type *hip_internal_scratch_flags(const HIP &instance,
...
 namespace Kokkos {
 namespace Impl {
-void hip_device_synchronize(const std::string &name) {
-  Kokkos::Tools::Experimental::Impl::profile_fence_event<Kokkos::HIP>(
-      name,
-      Kokkos::Tools::Experimental::SpecialSynchronizationCases::
-          GlobalDeviceSynchronization,
-      [&]() { KOKKOS_IMPL_HIP_SAFE_CALL(hipDeviceSynchronize()); });
-}
-
```
masterleinad, line 385: Drive-by: this is unused.

core/src/HIP/Kokkos_HIP_Locks.hpp
```diff
@@ -113,45 +113,49 @@ namespace {
...
-    ::Kokkos::Impl::lock_array_copied = 1;                      \
+#ifdef KOKKOS_ENABLE_HIP_RELOCATABLE_DEVICE_CODE
+inline
+#else
+inline static
+#endif
+    void
+    copy_hip_lock_arrays_to_device() {
+  if (lock_array_copied == 0) {
+    KOKKOS_IMPL_HIP_SAFE_CALL(hipMemcpyToSymbol(g_device_hip_lock_arrays,
```
Rombur, line 125: Why did you remove `HIP_SYMBOL`? New compilers don't need this anymore?

masterleinad: I introduced it again. Interesting that it also worked without it.

core/src/HIP/Kokkos_HIP_Locks.cpp
```diff
@@ -65,10 +64,11 @@ void initialize_host_hip_lock_arrays() {
       sizeof(std::int32_t) * (KOKKOS_IMPL_HIP_SPACE_ATOMIC_MASK + 1)));
 
   g_host_hip_lock_arrays.n = HIPInternal::concurrency();
-
-  KOKKOS_COPY_HIP_LOCK_ARRAYS_TO_DEVICE();
+  copy_hip_lock_arrays_to_device();
   init_lock_array_kernel_atomic<<<
       (KOKKOS_IMPL_HIP_SPACE_ATOMIC_MASK + 1 + 255) / 256, 256, 0, nullptr>>>();
+  HIP::impl_static_fence(
+      "Kokkos::Impl::initialize_host_hip_lock_arrays: Post Init Lock Arrays");
```
masterleinad, line 71: This fence is new (and we have it in the `CUDA` implementation).

dalg24 (approved):

core/src/HIP/Kokkos_HIP_Locks.cpp
Rombur: I don't see why this is necessary for the HIP backend

masterleinad: In my opinion, it makes sense to make sure that the kernel finishes before returning so that other streams see the correct values. Noe that we do the same in `desul`: https://github.com/desul/desul/blob/0a0dfd3a4b53030913f970f0a28714a0415e147a/atomics/src/Lock_Array_HIP.cpp#L78-L79
Hopefully, we can get rid of these variables soon anyway and just use the `desul` ones.

Rombur: This has nothing to do with streams. The function is only called when initializing the backend and it is followed by a synchronized HIP function. This just adds pointless code.

masterleinad: Not worth fighting over. I removed the fence.

Rombur (approved):

PR #5775, opened 2023-01-17
SIMD backend of ARM NEON

ibaned:
This adds a backend to the SIMD tools for the ARM NEON intrinsics. It compiles and passes unit tests using the ARM 21.0 `armclang` compilers.
Setting CMake configuration `-DKokkos_ARCH_ARMV8_THUNDERX2=ON`, and possibly other ARM ARCH settings, should enable this NEON backend and make it the "native" SIMD ABI.

### DIFF
--- a/simd/src/Kokkos_SIMD.hpp
+++ b/simd/src/Kokkos_SIMD.hpp
@@ -140,7 +140,7 @@ class abi_set {};
 using host_abi_set = abi_set<simd_abi::scalar, simd_abi::avx512_fixed_size<8>>;
 #elif defined(KOKKOS_ARCH_AVX2)
 using host_abi_set = abi_set<simd_abi::scalar, simd_abi::avx2_fixed_size<4>>;
-#elif defined(KOKKOS_ARCH_NEON)
+#elif defined(__ARM_NEON)
 using host_abi_set = abi_set<simd_abi::scalar, simd_abi::neon_fixed_size<2>>;
 #else
 using host_abi_set = abi_set<simd_abi::scalar>;
--- a/simd/src/Kokkos_SIMD.hpp
+++ b/simd/src/Kokkos_SIMD.hpp
@@ -29,6 +29,10 @@
 #include <Kokkos_SIMD_AVX512.hpp>
 #endif
 
+#ifdef __ARM_NEON
+#include <Kokkos_SIMD_NEON.hpp>
+#endif
+
 namespace Kokkos {
 namespace Experimental {
 
@@ -40,6 +44,8 @@ namespace Impl {
 using host_native = avx512_fixed_size<8>;
 #elif defined(KOKKOS_ARCH_AVX2)
 using host_native  = avx2_fixed_size<4>;
+#elif defined(__ARM_NEON)
+using host_native  = neon_fixed_size<2>;
 #else
 using host_native  = scalar;
 #endif
@@ -134,6 +140,8 @@ class abi_set {};
 using host_abi_set = abi_set<simd_abi::scalar, simd_abi::avx512_fixed_size<8>>;
 #elif defined(KOKKOS_ARCH_AVX2)
 using host_abi_set = abi_set<simd_abi::scalar, simd_abi::avx2_fixed_size<4>>;
+#elif defined(KOKKOS_ARCH_NEON)
+using host_abi_set = abi_set<simd_abi::scalar, simd_abi::neon_fixed_size<2>>;
 #else
 using host_abi_set = abi_set<simd_abi::scalar>;
 #endif
--- a/simd/src/Kokkos_SIMD_NEON.hpp
+++ b/simd/src/Kokkos_SIMD_NEON.hpp
@@ -983,8 +983,8 @@ class const_where_expression<
   }
   KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
   void copy_to(std::int32_t* mem, element_aligned_tag) const {
-    _mm_maskstore_epi32(mem, static_cast<__m128i>(m_mask),
-                        static_cast<__m128i>(m_value));
+    if (m_mask[0]) mem[0] = m_value[0];
+    if (m_mask[1]) mem[1] = m_value[1];
   }
 };
 
@@ -1001,7 +1001,8 @@ class where_expression<simd_mask<std::int32_t, simd_abi::neon_fixed_size<2>>,
       : const_where_expression(mask_arg, value_arg) {}
   KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
   void copy_from(std::int32_t const* mem, element_aligned_tag) {
-    m_value = value_type(_mm_maskload_epi32(mem, static_cast<__m128i>(m_mask)));
+    if (m_mask[0]) m_value[0] = mem[0];
+    if (m_mask[1]) m_value[1] = mem[1];
   }
 };
 
--- a/simd/src/Kokkos_SIMD_NEON.hpp
+++ b/simd/src/Kokkos_SIMD_NEON.hpp
@@ -444,10 +444,10 @@ simd<double, simd_abi::neon_fixed_size<2>> min(
 }
 
 [[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
-simd<double, simd_abi::neon_fixed_size<2>> condition(
-    simd_mask<double, simd_abi::neon_fixed_size<2>> const& a,
-    simd<double, simd_abi::neon_fixed_size<2>> const& b,
-    simd<double, simd_abi::neon_fixed_size<2>> const& c) {
+    simd<double, simd_abi::neon_fixed_size<2>>
+    condition(simd_mask<double, simd_abi::neon_fixed_size<2>> const& a,
+              simd<double, simd_abi::neon_fixed_size<2>> const& b,
+              simd<double, simd_abi::neon_fixed_size<2>> const& c) {
   return simd<double, simd_abi::neon_fixed_size<2>>(
       vbslq_f64(static_cast<uint64x2_t>(a), static_cast<float64x2_t>(b),
                 static_cast<float64x2_t>(c)));
@@ -720,10 +720,10 @@ class simd<std::int64_t, simd_abi::neon_fixed_size<2>> {
 }
 
 [[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
-simd<std::int64_t, simd_abi::neon_fixed_size<2>> condition(
-    simd_mask<std::int64_t, simd_abi::neon_fixed_size<2>> const& a,
-    simd<std::int64_t, simd_abi::neon_fixed_size<2>> const& b,
-    simd<std::int64_t, simd_abi::neon_fixed_size<2>> const& c) {
+    simd<std::int64_t, simd_abi::neon_fixed_size<2>>
+    condition(simd_mask<std::int64_t, simd_abi::neon_fixed_size<2>> const& a,
+              simd<std::int64_t, simd_abi::neon_fixed_size<2>> const& b,
+              simd<std::int64_t, simd_abi::neon_fixed_size<2>> const& c) {
   return simd<std::int64_t, simd_abi::neon_fixed_size<2>>(
       vbslq_s64(static_cast<uint64x2_t>(a), static_cast<int64x2_t>(b),
                 static_cast<int64x2_t>(c)));
@@ -856,10 +856,10 @@ simd<std::int64_t, simd_abi::neon_fixed_size<2>>::simd(
     : m_value(vreinterpretq_s64_u64(static_cast<uint64x2_t>(other))) {}
 
 [[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
-simd<std::uint64_t, simd_abi::neon_fixed_size<2>> condition(
-    simd_mask<std::uint64_t, simd_abi::neon_fixed_size<2>> const& a,
-    simd<std::uint64_t, simd_abi::neon_fixed_size<2>> const& b,
-    simd<std::uint64_t, simd_abi::neon_fixed_size<2>> const& c) {
+    simd<std::uint64_t, simd_abi::neon_fixed_size<2>>
+    condition(simd_mask<std::uint64_t, simd_abi::neon_fixed_size<2>> const& a,
+              simd<std::uint64_t, simd_abi::neon_fixed_size<2>> const& b,
+              simd<std::uint64_t, simd_abi::neon_fixed_size<2>> const& c) {
   return simd<std::uint64_t, simd_abi::neon_fixed_size<2>>(
       vbslq_u64(static_cast<uint64x2_t>(a), static_cast<uint64x2_t>(b),
                 static_cast<uint64x2_t>(c)));
--- a/simd/src/Kokkos_SIMD_NEON.hpp
+++ b/simd/src/Kokkos_SIMD_NEON.hpp
@@ -50,7 +50,8 @@ class simd_mask<double, simd_abi::neon_fixed_size<2>> {
         : m_mask(mask_arg), m_lane(lane_arg) {}
     KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION reference
     operator=(bool value) const {
-      // this switch statement is needed because the lane argument has to be a constant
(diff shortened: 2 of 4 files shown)


PhilMiller:
This should probably be targeting `develop`?

ibaned:
> This should probably be targeting `develop`?

This was something that we were hoping to have in the 4.0 release. Will it still get there if I target `develop`? I realize it is getting late to get things into 4.0 so I'm happy to talk more about it. Thank you!

masterleinad:
> This was something that we were hoping to have in the 4.0 release. Will it still get there if I target `develop`?

You would need to target the `release-candidate-4.0.0` branch. We don't merge pull requests into `master` directly.

ibaned:
> Details

Yep, that's what I think I'm doing right now. Please let me know if I should change anything.

ibaned:
Despite the strange CI errors, this passes all my tests including my internal Sandia application runs correctly. I think this is ready for Kokkos team review

dalg24 (commented):
Please update the description and discuss how one would go about enabling the NEON instructions.  Presumably just enable one of the Kokkos ARM arch?

simd/src/Kokkos_SIMD.hpp
```diff
@@ -29,6 +29,10 @@
 #include <Kokkos_SIMD_AVX512.hpp>
 #endif
 
+#ifdef __ARM_NEON
```
dalg24, line 32: Why did you decide to guard this way rather than follow what we had done for AVX2 and AVX512?
(Not necessarily asking you to changer, just trying to understand if you considered it.  To be fair it seems that these arch macros are only used in the simd package at the moment if we exclude the legacy atomics that we are about to remove anyway.)

ibaned: This is a good question. I think Kokkos had pre-existing macros for AVX and I was able to determine which Intel microarchitectures had support for which extensions. It was less clear to me which ARM architectures support NEON, so it was convenient to let the compiler indicate this. I agree that consistency would be better, perhaps we should use vendor-specific macros for AVX detection as well and reduce the responsibility of Kokkos macros.

dalg24: We discussed briefly with Christian. We think what you did is in the right direction, that we will probably get rid of `KOKKOS_ARCH_AVX*` macros right after we drop the legacy atomics.  One advantage is that it will enable the appropriate SIMD support when users configure with `-DKokkos_ARCH_NATIVE=ON`

ibaned: Sounds great!

PhilMiller: I think I've mentioned this elsewhere, but I think it might be important to maintain a distinction between (in particular, but maybe more generally) building for a microarchitecture that supports AVX-512 and specifically using AVX-512 SIMD.

https://travisdowns.github.io/blog/2020/01/17/avxfreq1.html and some of his other posts give details on why a user may want to compile 'native', but still use AVX2 SIMD.

I wouldn't be surprised if at least one implementation of ARM SVE ends up presenting similar issues down the line.

ibaned: I get your point, but I believe supporting this falls outside the scope of this PR review.

simd/src/Kokkos_SIMD_NEON.hpp
```diff
@@ -0,0 +1,1108 @@
...
+      // this switch statement is needed because the lane argument has to be a
+      // constant
+      switch (m_lane) {
+        case 0:
+          m_mask = vsetq_lane_u64(value ? 0xFFFFFFFFFFFFFFFFULL : 0, m_mask, 0);
+          break;
+        case 1:
+          m_mask = vsetq_lane_u64(value ? 0xFFFFFFFFFFFFFFFFULL : 0, m_mask, 1);
+          break;
+      }
```
Rombur, line 62: You cannot simply do
```c++
m_mask = vsetq_lane_u64(value ? 0xFFFFFFFFFFFFFFFFULL : 0, m_mask, (int const) m_lane);
```
or
```c++
int const const_lane = m_lane ? 0 : 1;
m_mask = vsetq_lane_u64(value ? 0xFFFFFFFFFFFFFFFFULL : 0, m_mask, const_lane);
```
It is easier to see what's the difference between the two calls to `vsetq_lane_u64`

simd/src/Kokkos_SIMD_NEON.hpp
```diff
@@ -0,0 +1,1108 @@
...
+        : m_mask(mask_arg), m_lane(lane_arg) {}
+    KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION reference
+    operator=(bool value) const {
+      switch (m_lane) {
+        case 0:
+          m_mask = vset_lane_u32(value ? 0xFFFFFFFFU : 0, m_mask, 0);
+          break;
+        case 1:
+          m_mask = vset_lane_u32(value ? 0xFFFFFFFFU : 0, m_mask, 1);
+          break;
```
Rombur, line 151: Same as above.

simd/src/Kokkos_SIMD_NEON.hpp
```diff
@@ -0,0 +1,1108 @@
...
+  }
+  KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION bool operator!=(
+      simd_mask const& other) const {
+    return !operator==(other);
+  }
+};
+
+template <>
+class simd_mask<std::int64_t, simd_abi::neon_fixed_size<2>> {
+  uint64x2_t m_value;
```
Rombur, line 217: I cannot see the difference between the code in this `class` and the code in
`class simd_mask<double, simd_abi::neon_fixed_size<2>>`.  What is it?

simd/src/Kokkos_SIMD.hpp
PhilMiller: That was more directed at Damien's remark about selecting SIMD instruction set based on `-DKokkos_ARCH_NATIVE` and dropping separate configuration variables/macros.

simd/src/Kokkos_SIMD_NEON.hpp
ibaned: Yea, that's what I had initially, and the ARM clang compiler gave an error saying that the lane argument must be a compile-time constant.... yea I fully agree with you that is bad design, but the compiler forces us.

simd/src/Kokkos_SIMD_NEON.hpp
ibaned: It may be exactly the same! they are both represented as `uint64x2_t` masks. We could make a common base class and try to derive, although we may have to use the Curiously Recurring Template Pattern to ensure the result types of operators are of the derived type

Rombur: Do we expect the two classes to always be the same? If we do, I think it would be useful to avoid duplicating code. Otherwise, we will need to fix a single bug multiple times.

ibaned: Okay! I've just pushed a couple of commits to de-duplicate this using CRTP base classes

Rombur: thanks

simd/src/Kokkos_SIMD_NEON.hpp
```diff
@@ -0,0 +1,964 @@
...
+      vadd_s32(static_cast<int32x2_t>(lhs), static_cast<int32x2_t>(rhs)));
+}
+
+[[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
+    simd<std::int32_t, simd_abi::neon_fixed_size<2>>
+    condition(simd_mask<std::int32_t, simd_abi::neon_fixed_size<2>> const& a,
+              simd<std::int32_t, simd_abi::neon_fixed_size<2>> const& b,
+              simd<std::int32_t, simd_abi::neon_fixed_size<2>> const& c) {
+  return simd<std::int32_t, simd_abi::neon_fixed_size<2>>(
+      vbsl_s32(static_cast<uint32x2_t>(a), static_cast<int32x2_t>(b),
```
dalg24, line 583: Did you mean `uint32x2_t ` and not `int32x2_t`?

dalg24: Nevermind the answer is yes.

simd/src/Kokkos_SIMD_NEON.hpp
```diff
@@ -0,0 +1,964 @@
...
+  }
+  KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION constexpr explicit neon_mask(
+      uint32x2_t const& value_in)
+      : m_value(value_in) {}
+  template <class U>
+  KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION neon_mask(neon_mask<U, 64> const& other)
+      : m_value(vqmovn_u64(static_cast<uint64x2_t>(other))) {}
+  template <class U>
+  KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION neon_mask(neon_mask<U, 32> const& other)
+      : m_value(static_cast<uint32x2_t>(other)) {}
```
dalg24, line 189: Please move these two definition just above `size()`.
It gives friendlier diff.

simd/src/Kokkos_SIMD_NEON.hpp
```diff
@@ -0,0 +1,964 @@
...
+}
+
+[[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
+    simd<std::int64_t, simd_abi::neon_fixed_size<2>>
+    operator-(simd<std::int64_t, simd_abi::neon_fixed_size<2>> const& a) {
+  return simd<std::int64_t, simd_abi::neon_fixed_size<2>>(
+      vnegq_s64(static_cast<int64x2_t>(a)));
+}
+
+KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
```
dalg24, line 707: The `int32_t` version has a `[[nodiscard]]` attribute.

simd/src/Kokkos_SIMD_NEON.hpp
```diff
@@ -0,0 +1,964 @@
...
+  return simd<std::int64_t, simd_abi::neon_fixed_size<2>>(
+      vsubq_s64(static_cast<int64x2_t>(lhs), static_cast<int64x2_t>(rhs)));
+}
+
+[[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
+    simd<std::int64_t, simd_abi::neon_fixed_size<2>>
+    operator-(simd<std::int64_t, simd_abi::neon_fixed_size<2>> const& a) {
+  return simd<std::int64_t, simd_abi::neon_fixed_size<2>>(
+      vnegq_s64(static_cast<int64x2_t>(a)));
+}
```
dalg24, line 705: The `int32_t` version does not have a "negate" function (unary `operator-`).
Also the "add" function (`binary `operator+`) is missing whereas is provided for `int32_t` .

simd/src/Kokkos_SIMD_NEON.hpp
```diff
@@ -0,0 +1,964 @@
...
+    return m_value;
+  }
+  KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION simd
+  operator<<(unsigned int rhs) const {
+    return simd(vshlq_u64(m_value, vdupq_n_s64(std::int64_t(rhs))));
+  }
+  KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION simd
+  operator>>(unsigned int rhs) const {
+    return simd(vshlq_u64(m_value, vdupq_n_s64(-std::int64_t(rhs))));
+  }
```
dalg24, line 805: operator left/right shift only for unsigned types?

simd/src/Kokkos_SIMD_NEON.hpp
```diff
@@ -0,0 +1,964 @@
...
+    : m_value(
+          vmovn_s64(vreinterpretq_s64_u64(static_cast<uint64x2_t>(other)))) {}
+
+KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
+simd<std::int64_t, simd_abi::neon_fixed_size<2>>::simd(
+    simd<std::uint64_t, simd_abi::neon_fixed_size<2>> const& other)
+    : m_value(vreinterpretq_s64_u64(static_cast<uint64x2_t>(other))) {}
+
+KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
+simd<std::uint64_t, simd_abi::neon_fixed_size<2>> condition(
```
dalg24, line 828: Same comment about the `[[nodiscard]]` attribute

simd/src/Kokkos_SIMD_NEON.hpp
```diff
@@ -0,0 +1,964 @@
...
+  }
+  KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION mask_type
+  operator==(simd const& other) const {
+    return mask_type(vceqq_u64(m_value, other.m_value));
+  }
+  KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION mask_type
+  operator!=(simd const& other) const {
+    return !((*this) == other);
+  }
+};
```
dalg24, line 814: No "add" and "sub" (binary `operator+` and `operator-`)?
No "neg" (unary `operator-`)?

simd/src/Kokkos_SIMD_NEON.hpp
```diff
@@ -0,0 +1,964 @@
...
+ public:
+  where_expression(
+      simd_mask<std::int32_t, simd_abi::neon_fixed_size<2>> const& mask_arg,
+      simd<std::int32_t, simd_abi::neon_fixed_size<2>>& value_arg)
+      : const_where_expression(mask_arg, value_arg) {}
+  KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
+  void copy_from(std::int32_t const* mem, element_aligned_tag) {
+    if (m_mask[0]) m_value[0] = mem[0];
+    if (m_mask[1]) m_value[1] = mem[1];
+  }
```
Rombur, line 958: Why is there no `gather_from` and no move operator like there is for `double`?

dalg24:
Would you say a word about the data types you chose to implement and the one you omitted?

simd/src/Kokkos_SIMD_NEON.hpp
ibaned: Sorry I'm not clear what you're asking: do you want an explanation of why only for these types or are you requesting code changes?

dalg24: only explanation at that point

simd/src/Kokkos_SIMD_NEON.hpp
ibaned: I will add the binary +/- momentarily, but I don't think an unsigned integer type should have unary negation

simd/src/Kokkos_SIMD_NEON.hpp
ibaned: Well I'm not proud of this explanation but to be honest I simply didn't need a shift for any other type in my code.

simd/src/Kokkos_SIMD_NEON.hpp
ibaned: Yea not proud of this explanation but I simply didn't need gathers and loads for this type in my code

ibaned:
Here's the rough justification for the data types:
 - `double`: really the workhorse for HPC codes
 - `int32_t`: meant to be used for indexing to support gather/scatter of `double`
 - `int64_t` and `uint64_t`: supports a special algorithm that decomposes `double` to do associative sums in vectorized parallel

`uint32_t` is omitted because it wasn't needed for any of the above needs in my code, but there is no technical barrier to adding it, only time cost.

ibaned:
Alright, I just pushed commits that I believe address all of @dalg24 's comments

Please let me know if other changes are desired.

Rombur (approved):

PhilMiller (changes requested):
This is still targeting `release-candidate-4.0.0` before the relevant changes are on `develop`. That's not the standard workflow.

ibaned:
rebased onto and pointed to `develop`!

crtrott:
Ok to test

ibaned:
@PhilMiller this will not test unless you mark your requested change as resolved

PhilMiller (approved):

dalg24 (approved):

PR #5778, opened 2023-01-18
Add half_t and bhalf_t limits

e10harvey:
Move these limits from KokkosKernels. Mostly epsilon is used in unit-tests. ArithTraits also uses the other limits.

Related to https://github.com/kokkos/kokkos-kernels/issues/1414.


cz4rs:
Is this relevant?
https://github.com/llvm/llvm-project/commit/059b823e709a90d3b5d2d24663830e59b3bdded9

e10harvey:
> Is this relevant? [llvm/llvm-project@059b823](https://github.com/llvm/llvm-project/commit/059b823e709a90d3b5d2d24663830e59b3bdded9)

It looks relevant. I don't think all toolchains define these constants though.

cz4rs:
> > Is this relevant? [llvm/llvm-project@059b823](https://github.com/llvm/llvm-project/commit/059b823e709a90d3b5d2d24663830e59b3bdded9)
> 
> It looks relevant. I don't think all toolchains define these constants though.

Specifically, changing `MIN_EXP` to `-13` and `MAX_EXP` to `16`.

dalg24 (dismissed):
Why define macros rather than provide specialization of the numeric traits?

crtrott:
Yeah: definitely lets just specialize the numeric traits. We need those anyway ...

e10harvey:
Yea, I was thinking about putting them in numeric traits. Let me know if you want the defines somewhere else. I'd prefer to keep the defines.

crtrott (changes requested):
We don't need the macros for anything. Those should go away. 
Also the numeric traits specialization for half should be in the half header not in the Kokkos_NumericTraits header.

core/src/Kokkos_Half.hpp
```diff
@@ -1005,6 +1007,163 @@ cast_from_bhalf(bhalf_t val) {
...
 #endif  // KOKKOS_IMPL_BHALF_TYPE_DEFINED
+////////////// BEGIN HALF_T (binary16) limits //////////////
+// bits [10,14] set
+// #define KOKKOS_IMPL_HALF_T_HUGE_VALH 0x7c00 0x7c00
+#if defined(KOKKOS_HALF_T_IS_FLOAT) && !KOKKOS_HALF_T_IS_FLOAT
+// Minimum normalized number
+template <>
+struct Kokkos::Experimental::Impl::finite_min_helper<
+    Kokkos::Experimental::half_t> {
+  static constexpr float value = -65504.0F;
```
dalg24, line 1018: float -> half_t

core/src/Kokkos_Half.hpp
```diff
@@ -1005,6 +1007,163 @@ cast_from_bhalf(bhalf_t val) {
...
+#endif
+////////////// END HALF_T (binary16) limits //////////////
+
+////////////// BEGIN BHALF_T (bfloat16) limits //////////////
+#if defined(KOKKOS_BHALF_T_IS_FLOAT) && !KOKKOS_BHALF_T_IS_FLOAT
+// Minimum normalized number
+template <>
+struct Kokkos::Experimental::Impl::finite_min_helper<
+    Kokkos::Experimental::bhalf_t> {
+  static constexpr float value = -3.38953139e38;
```
dalg24, line 1095: float -> bhalf_t

core/src/Kokkos_Half.hpp
e10harvey: Is there a way to instantiate floating_point_wrapper at compile time?

dalg24: you'd need to make sure the constructor is constexpr

e10harvey: @dalg24 Is https://github.com/kokkos/kokkos/pull/5778/commits/1b399bdd43da2b648d5f7cedbade0dfff51f8e43 what you had in mind?

dalg24: something like that yes

masterleinad:
```diff
diff --git a/core/src/Kokkos_Half.hpp b/core/src/Kokkos_Half.hpp
index 656b177cf..59bbac948 100644
--- a/core/src/Kokkos_Half.hpp
+++ b/core/src/Kokkos_Half.hpp
@@ -321,9 +321,7 @@ class alignas(FloatType) floating_point_wrapper {
   KOKKOS_FUNCTION
   floating_point_wrapper(long long rhs) : val(cast_to_wrapper(rhs, val).val) {}
   KOKKOS_FUNCTION
-  constexpr floating_point_wrapper(unsigned short rhs) {
-    reinterpret_cast<unsigned short>(val) = rhs;
-  }
+  constexpr floating_point_wrapper(unsigned short rhs) : val(cast_to_wrapper(rhs, val).val) {}
   KOKKOS_FUNCTION
   floating_point_wrapper(unsigned int rhs)
       : val(cast_to_wrapper(rhs, val).val) {}
@@ -865,7 +863,7 @@ class alignas(FloatType) floating_point_wrapper {
 template <class T>
 static KOKKOS_INLINE_FUNCTION constexpr Kokkos::Experimental::half_t
 cast_to_wrapper(T x, const volatile Kokkos::Impl::half_impl_t::type&) {
-  return Kokkos::Experimental::cast_to_half(x);
+  return Kokkos::Experimental::half_t::impl_type(x);
 }
 
 #ifdef KOKKOS_IMPL_BHALF_TYPE_DEFINED
@@ -1046,7 +1044,7 @@ struct Kokkos::Experimental::Impl::round_error_helper<
 template <>
 struct Kokkos::Experimental::Impl::norm_min_helper<
     Kokkos::Experimental::half_t> {
-  static constexpr int value = 0.000000059604645F;
+  static constexpr Kokkos::Experimental::half_t value = 0.000000059604645F;
 };
 // Quiet not a half precisioin number
 template <>
diff --git a/core/src/SYCL/Kokkos_SYCL_Half_Conversion.hpp b/core/src/SYCL/Kokkos_SYCL_Half_Conversion.hpp
index d7b0271e1..89a167903 100644
--- a/core/src/SYCL/Kokkos_SYCL_Half_Conversion.hpp
+++ b/core/src/SYCL/Kokkos_SYCL_Half_Conversion.hpp
@@ -29,7 +29,7 @@ namespace Experimental {
 KOKKOS_INLINE_FUNCTION
 half_t cast_to_half(half_t val) { return val; }
 
-KOKKOS_INLINE_FUNCTION
+constexpr KOKKOS_INLINE_FUNCTION
 half_t cast_to_half(float val) { return half_t::impl_type(val); }
 KOKKOS_INLINE_FUNCTION
 half_t cast_to_half(double val) { return half_t::impl_type(val); }
```
is enough to fix this for `SYCL`.

e10harvey:
> ```diff
> diff --git a/core/src/Kokkos_Half.hpp b/core/src/Kokkos_Half.hpp
> index 656b177cf..59bbac948 100644
> --- a/core/src/Kokkos_Half.hpp
> +++ b/core/src/Kokkos_Half.hpp
> @@ -321,9 +321,7 @@ class alignas(FloatType) floating_point_wrapper {
>    KOKKOS_FUNCTION
>    floating_point_wrapper(long long rhs) : val(cast_to_wrapper(rhs, val).val) {}
>    KOKKOS_FUNCTION
> -  constexpr floating_point_wrapper(unsigned short rhs) {
> -    reinterpret_cast<unsigned short>(val) = rhs;
> -  }
> +  constexpr floating_point_wrapper(unsigned short rhs) : val(cast_to_wrapper(rhs, val).val) {}
>    KOKKOS_FUNCTION
>    floating_point_wrapper(unsigned int rhs)
>        : val(cast_to_wrapper(rhs, val).val) {}
> @@ -865,7 +863,7 @@ class alignas(FloatType) floating_point_wrapper {
>  template <class T>
>  static KOKKOS_INLINE_FUNCTION constexpr Kokkos::Experimental::half_t
>  cast_to_wrapper(T x, const volatile Kokkos::Impl::half_impl_t::type&) {
> -  return Kokkos::Experimental::cast_to_half(x);
> +  return Kokkos::Experimental::half_t::impl_type(x);
>  }
>  
>  #ifdef KOKKOS_IMPL_BHALF_TYPE_DEFINED
> @@ -1046,7 +1044,7 @@ struct Kokkos::Experimental::Impl::round_error_helper<
>  template <>
>  struct Kokkos::Experimental::Impl::norm_min_helper<
>      Kokkos::Experimental::half_t> {
> -  static constexpr int value = 0.000000059604645F;
> +  static constexpr Kokkos::Experimental::half_t value = 0.000000059604645F;
>  };
>  // Quiet not a half precisioin number
>  template <>
> diff --git a/core/src/SYCL/Kokkos_SYCL_Half_Conversion.hpp b/core/src/SYCL/Kokkos_SYCL_Half_Conversion.hpp
> index d7b0271e1..89a167903 100644
> --- a/core/src/SYCL/Kokkos_SYCL_Half_Conversion.hpp
> +++ b/core/src/SYCL/Kokkos_SYCL_Half_Conversion.hpp
> @@ -29,7 +29,7 @@ namespace Experimental {
>  KOKKOS_INLINE_FUNCTION
>  half_t cast_to_half(half_t val) { return val; }
>  
> -KOKKOS_INLINE_FUNCTION
> +constexpr KOKKOS_INLINE_FUNCTION
>  half_t cast_to_half(float val) { return half_t::impl_type(val); }
>  KOKKOS_INLINE_FUNCTION
>  half_t cast_to_half(double val) { return half_t::impl_type(val); }
> ```
> 
> is enough to fix this for `SYCL`.

The same changes on CUDA give:
```
kokkos/core/src/Kokkos_Half.hpp(1043): error: expression must have a constant value
```

masterleinad:
> The same changes on CUDA give:

Yes, I am seeing
```
/tmp/kokkos/core/src/Kokkos_Half.hpp(1024): error: expression must have a constant value
/tmp/kokkos/core/src/Kokkos_Half.hpp(866): note #2703-D: cannot call non-constexpr function "__half::__half(float)"
/soft/compilers/cuda/cuda-11.6.2/include/cuda_fp16.hpp(201): here
/tmp/kokkos/core/src/Kokkos_Half.hpp(309): note #2693-D: called from:
```
indicating that none of the conversion functions is `constexpr` in `Cuda`.

ajpowelsnl:
@e10harvey - does this issue need some discussion today?

e10harvey:
> @e10harvey - does this issue need some discussion today?

@ajpowelsnl - thanks for following up. I think it needs another review by @crtrott and @dalg24 -- please let me know if you have any questions.

masterleinad:
There are failing tests.

masterleinad:
```
[ RUN      ] defaultdevicetype.bhalf_operators
11: /var/jenkins/workspace/Kokkos/core/unit_test/TestHalfOperators.hpp:926: Failure
11: The difference between f_device_actual_lhs(op_test) and f_device_expected_lhs(op_test) is 0.00020498037338256836, which exceeds epsilon, where
11: f_device_actual_lhs(op_test) evaluates to -0.765625,
11: f_device_expected_lhs(op_test) evaluates to -0.76542001962661743, and
11: epsilon evaluates to 1.1920928955078125e-07.
11: /var/jenkins/workspace/Kokkos/core/unit_test/TestHalfOperators.hpp:926: Failure
11: The difference between f_device_actual_lhs(op_test) and f_device_expected_lhs(op_test) is 3.9145350456237793e-05, which exceeds epsilon, where
11: f_device_actual_lhs(op_test) evaluates to 0.234619140625,
11: f_device_expected_lhs(op_test) evaluates to 0.23457999527454376, and
11: epsilon evaluates to 1.1920928955078125e-07.
11: /var/jenkins/workspace/Kokkos/core/unit_test/TestHalfOperators.hpp:926: Failure
11: The difference between f_device_actual_lhs(op_test) and f_device_expected_lhs(op_test) is 0.00020503997802734375, which exceeds epsilon, where
11: f_device_actual_lhs(op_test) evaluates to 1.234375,
11: f_device_expected_lhs(op_test) evaluates to 1.2345800399780273, and
11: epsilon evaluates to 1.1920928955078125e-07.
11: /var/jenkins/workspace/Kokkos/core/unit_test/TestHalfOperators.hpp:926: Failure
11: The difference between f_device_actual_lhs(op_test) and f_device_expected_lhs(op_test) is 0.00020503997802734375, which exceeds epsilon, where
11: f_device_actual_lhs(op_test) evaluates to 2.234375,
11: f_device_expected_lhs(op_test) evaluates to 2.2345800399780273, and
11: epsilon evaluates to 1.1920928955078125e-07.
11: [  FAILED  ] defaultdevicetype.bhalf_operators (0 ms)
```

e10harvey:
Hi @mhoemmen ! Is there a way to invoke CUDA half and bhalf casting intrinsics at compile time?

e10harvey:
Hi @mhoemmen ! Is there a way to invoke CUDA half and bhalf casting intrinsics at compile time?

e10harvey:
One more test failed with:
```
7: Time CudaSpace: 0.006551 CudaUVMSpace_1: 0.006560 CudaUVMSpace_2: 0.006555 CudaPinnedHostSpace: 0.240870 CudaUVMSpace_Pinned: 0.022919
7: /path/toKokkos/core/unit_test/cuda/TestCuda_DebugPinUVMSpace.cpp:100: Failure
7: Value of: passed
7:   Actual: false
7: Expected: true
7: [  FAILED  ] cuda.debug_pin_um_to_host (367 ms)
7: [----------] 2 tests from cuda (379 ms total)
7: 
7: [==========] 2 tests from 1 test suite ran. (379 ms total)
7: [  PASSED  ] 1 test.
7: [  FAILED  ] 1 test, listed below:
7: [  FAILED  ] cuda.debug_pin_um_to_host
7: 
7:  1 FAILED TEST
 7/37 Test  #7: KokkosCore_UnitTest_CudaTimingBased ..........***Failed  Error regular expression found in output. Regex=[  FAILED  ]  0.87 sec
 ```
 
Is there a way to retest just that build?

**Comment by PhilMiller** [comment]
If it's just a timing based test failing, typically it will just be disregarded and the PR merged.

**Comment by e10harvey** [comment]
@dalg24, @crtrott, @lucbv: Ready to merge?

File: core/unit_test/TestHalfOperators.hpp [hunk 3]
```diff
@@ -910,12 +908,13 @@ struct Functor_TestHalfOperators {
...
-  if (std::is_same<half_type, Kokkos::Experimental::half_t>::value)
-    epsilon = FP16_EPSILON;
-  if (std::is_same<half_type, Kokkos::Experimental::bhalf_t>::value)
-    epsilon = BF16_EPSILON;
+  double epsilon = Kokkos::Experimental::epsilon<half_type>::value;
+#if defined(KOKKOS_BHALF_T_IS_FLOAT) && KOKKOS_BHALF_T_IS_FLOAT
+  double constexpr epsilon_bias = 10e3;
+#else
+  double constexpr epsilon_bias = 1;
+#endif
```
dalg24, line 916: Why do you need a bias?

dalg24 (commented):
How did you come up with these distinguished values?

crtrott (approved):
Other than that test question from Damien I am good with this. Considering that the float/half conversion is not constexpr this is the best we can do for now.

core/unit_test/TestHalfOperators.hpp
e10harvey: Good question. Before moving these constants to numeric traits, when bhalf_t or half_t were aliases to float, the tests were erroneously using FP16_EPSILON and BF16_EPSILON from:
```
 double epsilon = FLT_EPSILON;

  if (std::is_same<half_type, Kokkos::Experimental::half_t>::value)
    epsilon = FP16_EPSILON;
  if (std::is_same<half_type, Kokkos::Experimental::bhalf_t>::value)
    epsilon = BF16_EPSILON;
```

After further triaging, I found that the only test failing without the bias was PASS_BY_REFF. The way the PASS_BY_REF test was written resulted in the compiler optimizing out the cast from float -> double. See the latest commit for a new way of writing that test.

e10harvey:
> 

See above reply.

core/unit_test/TestHalfOperators.hpp
```diff
@@ -928,10 +932,12 @@ void __test_half_operators(half_type h_lhs, half_type h_rhs) {
   Kokkos::deep_copy(f_device_expected_lhs, f_device.expected_lhs);
   for (int op_test = 0; op_test < N_OP_TESTS; op_test++) {
     // printf("op_test = %d\n", op_test);
-    ASSERT_NEAR(f_device_actual_lhs(op_test), f_device_expected_lhs(op_test),
-                epsilon);
-    ASSERT_NEAR(f_host.actual_lhs(op_test), f_host.expected_lhs(op_test),
-                epsilon);
+    if (op_test != PASS_BY_REF) {
```
masterleinad, line 935: What are we checking for that test case now?

e10harvey: This is a typo. One moment.

masterleinad (approved):

e10harvey:
I do not think the `***Exception: Illegal` failures are related to these changes. Can anyone confirm?

e10harvey:
The HIP and SYCL CI failures appear to be unrelated to the changes as well.

crtrott:
Retest this please.

core/unit_test/TestHalfOperators.hpp
```diff
@@ -887,7 +895,7 @@ struct Functor_TestHalfOperators {
     // expected_lhs(TW) = d_lhs <=> d_rhs;  // Need C++20?
 
     actual_lhs(PASS_BY_REF)   = static_cast<double>(accept_ref(h_lhs));
-    expected_lhs(PASS_BY_REF) = d_lhs;
+    expected_lhs(PASS_BY_REF) = accept_ref_expected(h_lhs);
```
dalg24, line 898: I am sorry I don't get that change.  Can you please confirm that the test fail without it.

e10harvey: @dalg24: I have confirmed it fails without it. Please feel free to revert this change to confirm.

crtrott: ok I don't get it either. How are `d_lhs` and `accept_ref_expected(h_lhs)` different? Does this test pass if you just assign `h_lhs` instead of `accept_ref_exepcted(h_lhs)`?

crtrott (changes requested):

core/unit_test/TestHalfOperators.hpp
e10harvey: > Not blocking but also don't fully understand the changes in the unit test

@dalg24, @crtrott: Maybe my understanding is flawed. It appears that the compiler was optimizing out the casts float -> double -> float when we had:
```
expected_lhs(PASS_BY_REF) = d_lhs
```
 
From the compiler writer's perspective, this makes sense since there's no need to spend runtime cycles casting from float -> double -> float again during instantiation: https://github.com/kokkos/kokkos/pull/5778/files/15ed7c0bad93bf7d6040900eb498bfbcf71dcd3e#diff-6980b46765a569d9ae64dffdbe6a0db53b9ed49c8b9157236055a90add5b47afR268. Note that in this case half is an alias to float.

For the unit test though, this means that actual_lhs gets the result of the casts float -> double -> float which is accompanied by rounding while expected_lhs was not cast from float -> double -> float, instead, it was getting a store of float -> double. I believe this is the cause of the unit-test failures that we're previously erroneously addressed with the `epsilon_bias`. I believe this casting difference is propagated by the compiler "helping with optimizations" from the lines we're commenting on here as well as https://github.com/kokkos/kokkos/pull/5778/files/15ed7c0bad93bf7d6040900eb498bfbcf71dcd3e#diff-6980b46765a569d9ae64dffdbe6a0db53b9ed49c8b9157236055a90add5b47afR268. We didn't see this for half_t since half_t always maps to a half precision float on all the Kokkos backends that support half precision (HIP, CUDA, and SYCL). CUDA is the only backend that supports bhalf_t and only CUDA 11+ supports it. This is my understanding. Does that make sense? Do you agree?

e10harvey: Also, I don't think this was failing when half_t or bhalf_t mapped to a 16-bit floating point type since the epsilons for those types are larger than FLT_EPSILON.

e10harvey:
I think this is ready to merge. Let me know if you have any other feedback / questions.

crtrott (approved):
Ok I think I get it now.

crtrott:
Uhoh work graph test failed???

crtrott:
I mean workgraph is clearly not an issue coming from this I would expect but this is weird

dalg24:
> How did you come up with these distinguished values?

Ping

e10harvey:
> 

@dalg24, see response in https://github.com/kokkos/kokkos/pull/5778#discussion_r1113230504.

e10harvey:
> I mean workgraph is clearly not an issue coming from this I would expect but this is weird

I don't see the workgraph failure. Instead, I see that OPENMPTARGET-Clang failed in 0s. How do I triage this?

dalg24:
> @dalg24, see response in [#5778 (comment)](https://github.com/kokkos/kokkos/pull/5778#discussion_r1113230504).

Is there documentation somewhere that you can refer to?

e10harvey:
> Is there documentation somewhere that you can refer to?

@dalg24: I'm confused. These values are removed now.

dalg24:
> > Is there documentation somewhere that you can refer to?
> 
> @dalg24: I'm confused. These values are removed now.

What I am asking is "where would one find out what is the maximum value that is representable with half type".  Presumably you looked that up somewhere or you did some experiment.  Please document how you came up with these numbers.

e10harvey:
> > > Is there documentation somewhere that you can refer to?
> > 
> > 
> > @dalg24: I'm confused. These values are removed now.
> 
> What I am asking is "where would one find out what is the maximum value that is representable with half type". Presumably you looked that up somewhere or you did some experiment. Please document how you came up with these numbers.

Gotcha. I will document this. Should this be documented in the `Kokkos_Half.hpp` file?

dalg24:
> Gotcha. I will document this. Should this be documented in the `Kokkos_Half.hpp` file?

At the very least here in the discussion but if there is a nice reference yes also add as a comment in code.

e10harvey:
> > Gotcha. I will document this. Should this be documented in the `Kokkos_Half.hpp` file?
> 
> At the very least here in the discussion but if there is a nice reference yes also add as a comment in code.

OK, thanks for the clarifications, @dalg24. I did not follow a particular reference. I simply looked up definitions of the descriptions for each type of constant, as well as the format of each half precision type (i.e. N bits for significand and M bits for mantissa) and then computed the constants. For the format of the half type I followed https://ieeexplore.ieee.org/abstract/document/9973611. For computing the values, I used https://observablehq.com/@benaubin/floating-point or python, depending on whether I wanted to set individual bits in the half precision type or evaluate an expression. In some cases, I also set the bits via hex constants in the Kokkos half type and used printf to print the base10 representation.

In short, I do not have a single reference for where these values came from but I did my best to verify each constant is correct. Perhaps there is a better reference for these constants, but I did not find one during my half precision development work.

I also documented the human readable description and equation or high bits above each constants in Kokkos_Half.hpp.

I hope this helps.

e10harvey:
@dalg24: If you think https://github.com/kokkos/kokkos/pull/5778/commits/a509f58386f7628eb7b7f1be6759f645accf418f suffices, I will add similar docs for bhalf_t.

e10harvey:
The GCC 8.4.0 pipeline stage in jenkins timed out.

dalg24:
> The GCC 8.4.0 pipeline stage in jenkins timed out.

We are having some issues with the CPU only testing machines.  I would have merged w/o that one :/

dalg24:
Evan dug out a build that had only GCC 8.4 not run
https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/Kokkos/detail/Kokkos/12521/pipeline/57/
That is good enough for me.

PR #5777, opened 2023-01-18
Fix build on Fedora rawhide

junghans:
uint32_t now gets defined in cstdint

PR #5780, opened 2023-01-18
Fixup GH Actions compiler warnings

dalg24:
Configure with `-DKokkos_ENABLE_COMPILER_WARNINGS=ON`
and fix https://github.com/kokkos/kokkos/pull/5777#issuecomment-1387550429

~~We need to look into fixing other warnings with (ICX for instance) and eventually pass `-Werror`~~


cz4rs:
Two loosely related ideas:
- This silences `ICC` warning, but Intel compiler version is not pinned in fedora image, so we might run into trouble when they decide to remove it. Should we pin `ICC` to some specific version?
- Do you think that enabling `Kokkos_ENABLE_DEBUG_BOUNDS_CHECK` could be good too?

junghans:
Related, I think it would be good to build with `--verbose` to see if the flag are actually there.

.github/workflows/continuous-integration-workflow.yml
```diff
@@ -12,30 +12,27 @@ jobs:
...
+        cxx_extra_flags: ['']
         cmake_build_type: ['Release', 'Debug']
         backend: ['OPENMP']
         clang-tidy: ['']
         include:
           - distro: 'fedora:intel'
             cxx: 'icpc'
+            cxx_extra_flags: '-diag-disable=177,10441'
             cmake_build_type: 'Release'
-            backend: 'OPENMP'
```
junghans, line 22: We can drop that https://github.com/kokkos/kokkos/actions/runs/3961405441/jobs/6786791382#step:11:9 
```
    -DKokkos_ENABLE_HWLOC=ON \
    -DKokkos_ENABLE_=ON \
    -DKokkos_ENABLE_TESTS=ON \
```
empty backend leads to `Kokkos_ENABLE_=ON`.

core/perf_test/BenchmarkMain.cpp
```diff
@@ -14,7 +14,14 @@
 //
 //@HEADER
 
+// Avoid deprecation warning for ICC
+#ifdef __INTEL_COMPILER
+#pragma warning(disable : 1786)
```
junghans, line 19: I think:
```
#pragma warning(push)
#pragma warning(disable: 1786)
...
#pragma warning(pop)
```
is better.

And do we always want to disable the warning or just with ENABLE_COMPILER_WARNINGS=OFF?

.github/workflows/continuous-integration-workflow.yml
```diff
@@ -12,30 +12,27 @@ jobs:
...
         distro: ['fedora:latest', 'fedora:rawhide', 'ubuntu:latest']
         cxx: ['g++', 'clang++']
+        cxx_extra_flags: ['']
         cmake_build_type: ['Release', 'Debug']
         backend: ['OPENMP']
         clang-tidy: ['']
         include:
           - distro: 'fedora:intel'
             cxx: 'icpc'
+            cxx_extra_flags: '-diag-disable=177,10441'
```
masterleinad, line 22: This avoids
```
warning #177: member "<unnamed>::defaultdevicetype_cmd_line_args_unrecognized_flag_Test::test_info_" was declared but never referenced
```
which we seemingly get for every test.

dalg24 (commented):
Please open one separate PR with the `snprintf` changes and one more with the floating point model fix for the math functions

.github/workflows/continuous-integration-workflow.yml
dalg24: Do these actually run?

masterleinad: Yes, see https://github.com/masterleinad/kokkos/actions/runs/4000209392/jobs/6865047046.

dalg24: No I mean the actual executable output `defaultdevicetype.cmd_line_args_unrecognized_flag` from gtest

masterleinad: The error was also there for `defaultdevicetype.initialization_settings` so I picked that one to just `abort()` in it.

junghans: > Related, I think it would be good to build with `--verbose` to see if the flag are actually there.

ping on that, so I can see the flags it is build with.

masterleinad: The flags passed to `CMake` are in the logs.

core/src/Kokkos_Macros.hpp
```diff
@@ -183,8 +183,8 @@
 // Intel compiler macros
 
 #if defined(KOKKOS_COMPILER_INTEL)
-// FIXME_SYCL
-#if !defined(KOKKOS_ENABLE_SYCL)
+// FIXME_ICPX
+#if !defined(__INTEL_LLVM_COMPILER)
```
dalg24, line 187: What the deal with that change?

masterleinad: The compiler is complaining about not knowing the corresponding pragmas. I initially observed that when using `SYCL` but it also seems to appear without it.
I would discuss the existence of these macros anyway. IMHO, they only make sense if they actually expanded to a pragma.

dalg24:
Can't approve my own PR but LGTM now

.github/workflows/continuous-integration-workflow.yml
masterleinad: See https://github.com/masterleinad/kokkos/actions/runs/4001195807/jobs/6867188598 for a run with `ctest -V` that shows `defaultdevicetype.cmd_line_args_unrecognized_flag` and `defaultdevicetype.initialization_settings` being run (among others).

masterleinad:
> Can't approve my own PR but LGTM now

I'll approve in your stead then. 🙂

masterleinad (approved):

Rombur (approved):

PR #5784, opened 2023-01-20
Drop KOKKOS_IMPL_WORKAROUND_INTEL_LLVM_DEFAULT_FLOATING_POINT_MODEL

masterleinad:
Part of #5780. There are more warnings than the ones covered by the guard in question and in my experience unit testing only really gives sensible results with `-fp-model=precise` anyway.

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -12,6 +12,7 @@ jobs:
       matrix:
         distro: ['fedora:latest', 'fedora:rawhide', 'ubuntu:latest']
         cxx: ['g++', 'clang++']
+        cxx_extra_flags: ['']
         cmake_build_type: ['Release', 'Debug']
         backend: ['OPENMP']
         clang-tidy: ['']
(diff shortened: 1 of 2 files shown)


dalg24:
> Part of #5780. There are more warnings than the ones covered by the guard in question and in my experience unit testing only really gives sensible results with `-fp-model=precise` anyway.

Did you attempt to disable the other warnings rather than requiring a specific floating point model?

masterleinad:
> Did you attempt to disable the other warnings rather than requiring a specific floating point model?

Of course, you can just disable all the warnings but I still get failing tests on the testbeds if I don't use `-fp-model=precise`.

PhilMiller:
For some added context (not sure if you guys know it, but others may not), `icpx -fp-model=fast` (the default with higher optimization) is currently *way* more aggressive than `icpc -fp-model=fast`, to the point of just being ridiculously wrong. It sets attributes in the generated LLVM IR that assert the absolute absence of non-normal FP values (infinity, NaN, denormal), and the code is optimized accordingly. Hence, it will evaluate literal `isnan(NaN)` to `false`.

Word needs to get out to the various DOE users, but I think every code, or maybe Trilinos, just needs to be forcing `-fp-model=precise`. Kokkos is well-positioned to just do that, but it's really not the right place for it.

PhilMiller:
Ok, digging back in my email when I tried to file a bug report about this, the recommendation was to add `-fhonor-infinities -fhonor-nans`, rather than avoiding `-fp-model=fast` entirely.

I still think this is a compiler bug, but Intel's position seemed to be that since it's something LLVM does, they're washing their hands of responsibility for telling LLVM to act like this.

dalg24 (approved):

crtrott (approved):

Issue #5789, opened 2023-01-22
Require desul atomics and remove legacy fallback implementation


dalg24:
The goal of this issue is to track the progress of that refactor.  I plan on breaking it up into several PRs to facilitate review.
* [x] update build system
* [x] remove legacy Kokkos atomics
* [x] cleanup

dalg24:
Let me know if you have an issue with the strategy.  I have a draft of the whole refactor and it is gives `47 files changed, 81 insertions(+), 6038 deletions(-)`.  I think that we are more likely to overlook details if I open one big PR.

PR #5790, opened 2023-01-22
Error out at configuration time if attempting to disable desul atomics

dalg24:
First step in #5789 
Prevent disabling desul atomics.  I opted for a hard error at configuration time so this cannot be ignored.  The alternative would be to raise some kind of warning that the user-requested setting will be ignored.

Note that I did not attempt to warn the user if they explicitly configure with `-DKokkos_ENABLE_IMPL_DESUL_ATOMICS=ON`.


dalg24:
> Note that I did not attempt to warn the user if they explicitly configure with `-DKokkos_ENABLE_IMPL_DESUL_ATOMICS=ON`.

A bit more on this.  I opted for that strategy because it makes the diff minimal.  In the next, PR I will prune configuration-time branches that consider whether desul atomics are enabled or not.  When I get to the final cleanup, as we discuss the removal of the `KOKKOS_ENABLE_IMPL_DESUL_ATOMICS` macro, we can revisit it.

dalg24:
Makefile didn't like my appending `// deprecated` to the `KOKKOS_INTERNAL_*` lines...
Fixed now

Rombur:
I would emit a warning instead of an error so users can keep the same script for multiple versions of Kokkos.

dalg24:
> I would emit a warning instead of an error so users can keep the same script for multiple versions of Kokkos.

Proposed #5801 as an alternative resolution

masterleinad (approved):
I like this better. It's an `IMPL` variable and there shouldn't really be a good reason for people to use it. People tend to ignore configure-time warnings anyway.

dalg24:
Closing in favor of #5801

Issue #5690, opened 2022-12-15
Problems in develop with ROCm 5.4


Rombur:
I have strange errors using ROCm 5.4 and develop. The test ` hip.mathspecialfunc_cbesselj1y1` sometimes fails with `Memory access fault by GPU node-4 (Agent handle: 0x67be760) on address 0x7ff7e9dca000. Reason: Unknown.`, sometimes it hangs, and rarely it passes. When the test passes another one fails. Everything works with master. I am going to do a bisect to find where the problem comes from.

Rombur:
The faulty commit is c5f7108644555e8b0526b4fb2d46601dfbe5b8bc

ajpowelsnl:
Should this issue be a blocker for Kokkos-4.0?

romintomasetti:
We are also affected by this issue on ROCm 5.4 (VEGA906).

ajpowelsnl:
> We are also affected by this issue on ROCm 5.4 (VEGA906).

@romintomasetti - Thank you for letting us know.  We're working to resolve the problem as quickly as possible.

Rombur:
@romintomasetti We have talked to AMD and it's a compiler bug. Right now it's best usig ROCm 5.2. We have a fix for the problem but because the problem depends on the user kernel and the performance cost of the fix can be pretty large,  users will have to change the execution policy for the offending kernels. We won't apply the fix indiscriminately.

Rombur:
#5796 introduces a workaround when hitting this bug

fnrizzi:
@Rombur can we close this issue?

maartenarnst:
Just an update :) I've tried with ROCm 6.0, and I found that the following tests now pass:
- `mathspecialfunc_cbesselj0y0`
- `mathspecialfunc_cbesselj1y1`
- `mathspecialfunc_cbesseli0k0`
- `mathspecialfunc_cbesseli1k1`

But the following two remaining tests still fail:
- `mathspecialfunc_cbesselh1stkind`
- `mathspecialfunc_cbesselh2ndkind`

The fix using the property tag doesn't seem to solve the issue for these two remaining tests.

Currently, the tests are disabled for certain hardcoded major/minor versions of ROCm. It may be necessary to add `6.0` for the two remaining failing tests. May be related to #6679.

maartenarnst:
Just wanted to update this issue.

I've tried with ROCm 6.2, released a few days ago.

The two tests that failed with ROCm 6.0 still fail:
[...]

Issue #5606, opened 2022-10-31
Kokkos hangs in develop with ROCm 5.3 and managed memory


Rombur:
I am opening an issue because this blocks the promotion. With ROCm 5.3, `hip_managed.view_allocation_error` segfaults on master. The same test segfaults on develop with `XNACK=0`. The `hip_managed` tests hangs on develop with `XNACK=1`. Everything works as expected with ROCm 5.2

JBludau:
have we checked, if there is some updated requirement for the kernel and driver versions? Feels like stating the obvious

Rombur:
I checked the release note there was nothing about it.

Rombur:
I found the PR that causes the tests to hang: https://github.com/kokkos/kokkos/pull/5448 It's when we decided to pass the functors by value instead of copying them. I need to think about this more but right now, I don't think we have many choices. We either revert the patch or we have both path in the code and we choose one of them depending on the ROCm version.

arghdos:
Interestingly, extracting one of the [hanging tests](https://github.com/kokkos/kokkos/blob/master/core/unit_test/TestViewSubview.hpp#L160) to a standalone test ... doesn't hang.
The hang itself is happening on _some_ lanes of e.g., a store to a managed view: https://github.com/kokkos/kokkos/blob/master/core/unit_test/TestViewSubview.hpp#L160

arghdos:
The segfault w/ HSA_XNACK=0 was comparatively easy to track down, have that reproduced internally.

arghdos:
> The segfault w/ HSA_XNACK=0 was comparatively easy to track down, have that reproduced internally.

Was fixed with https://github.com/ROCm-Developer-Tools/hipamd/commit/df82f201f26e297ae4c0f9e263ab85173a2690dd

arghdos:
I've extracted the `hip_managed.view_subview_auto_1d_left` test out of GTEST and into a standalone code (giving reasonable facsimiles of of `ASSERT_EQ` and `TEST` and friends), and ... it _still_ doesn't hang.

It is starting to seem like GTEST is required... somehow.  No idea how, yet but working on reducing the test case as much as possible.

arghdos:
Been chewing on this in the background.
One thing I noticed is that the whole `DriverType` pointer cycling scheme also went away in https://github.com/kokkos/kokkos/pull/5448.

This will change the performance of e.g., the `GlobalMemory` launcher, but it's not clear that it's related to this hang.

[...]

PR #5796, opened 2023-01-23
Add parameter to force using GlobalMemory launch mechanism using HIP

Rombur:
Add a parameter that users can use to force a kernel to be launched using `GlobalMemory`. This is sometimes necessary due to compiler bugs in ROCm 5.3 and ROCm 5.4 see https://github.com/kokkos/kokkos/issues/5606 and https://github.com/kokkos/kokkos/issues/5690

cc: @arghdos

### DIFF
--- a/core/unit_test/TestViewAPI.hpp
+++ b/core/unit_test/TestViewAPI.hpp
@@ -1542,6 +1542,7 @@ class TestViewAPI {
   }
 
   static void run_test_error() {
+#if !((HIP_VERSION_MAJOR == 5) && (HIP_VERSION_MINOR == 3))
 #ifdef KOKKOS_ENABLE_OPENMPTARGET
     if (std::is_same<typename dView1::memory_space,
                      Kokkos::Experimental::OpenMPTargetSpace>::value)
@@ -1586,6 +1587,7 @@ class TestViewAPI {
       }
 #endif
     }
+#endif
   }
 };
 
--- a/core/unit_test/TestViewAPI.hpp
+++ b/core/unit_test/TestViewAPI.hpp
@@ -1542,7 +1542,6 @@ class TestViewAPI {
   }
 
   static void run_test_error() {
-#if !((HIP_VERSION_MAJOR == 5) && (HIP_VERSION_MINOR == 3))
 #ifdef KOKKOS_ENABLE_OPENMPTARGET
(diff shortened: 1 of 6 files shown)


core/unit_test/TestViewAPI.hpp
```diff
@@ -1542,6 +1542,7 @@ class TestViewAPI {
   }
 
   static void run_test_error() {
+#if !((HIP_VERSION_MAJOR == 5) && (HIP_VERSION_MINOR == 3))
```
dalg24, line 1545: No, instead guard https://github.com/kokkos/kokkos/blob/4271bd9f034f028c9d93691b9506f62063f67f01/core/unit_test/TestViewAPI_d.hpp#L29-L31
and make sure you add a comment

dalg24 (changes requested):

core/unit_test/TestViewAPI.hpp
Rombur: That was for debugging. It shouldn't be there anymore

core/unit_test/TestViewAPI_d.hpp
```diff
@@ -26,11 +26,11 @@ TEST(TEST_CATEGORY, view_api_d) {
   TestViewAPI<double, TEST_EXECSPACE>::run_test_view_operator_c();
 }
 
-// ROCm 5.3 segfaults when trying to allocate too much memory
-#if !((HIP_VERSION_MAJOR == 5) && (HIP_VERSION_MINOR == 3))
 TEST(TEST_CATEGORY, view_allocation_error) {
+#if ((HIP_VERSION_MAJOR == 5) && (HIP_VERSION_MINOR == 3))
+  GTEST_SKIP() << "ROCm 5.3 segfaults when trying to allocate too much memor";
```
dalg24, line 31:
```suggestion
  GTEST_SKIP() << "ROCm 5.3 segfaults when trying to allocate too much memory";
```

dalg24 (approved):

crtrott (changes requested):
Can we name it `ImplForceGlobalLaunch` make it clearer that it has something to do with launch mechanism?

Rombur:
done

crtrott (approved):

dalg24:
@Rombur plz cherry-pick into RC 4.0

PR #5801, opened 2023-01-24
Warn at configuration time if attempting to disable desul atomics and force using it

dalg24:
Alternative to #5790 where we warn and just ignore the user setting instead of erroring out

Yields both
```
CMake Warning at cmake/kokkos_enable_options.cmake:167 (MESSAGE):
  Kokkos_ENABLE_IMPL_DESUL_ATOMICS option has been removed.  Desul atomics
  cannot be disabled.
  cmake/kokkos_tribits.cmake:242 (INCLUDE)
  CMakeLists.txt:213 (KOKKOS_SETUP_BUILD_ENVIRONMENT)

<snip>

CMake Warning:

    Kokkos_ENABLE_IMPL_DESUL_ATOMICS
```

**EDIT** Does not raise the `Manually-specified variables were not used` warning with the 2nd commit.
But I think this is good enough for now.  In follow-up PRs will remove these variables and the warning will be back.


crtrott (approved):
I prefer warning slightly. Would have been ok with just removing the option and have CMake warn about unused variable.

cz4rs (approved):
No strong preference between the two approaches. Looks good from the technical side.

dalg24:
Was step 1 in #5789

PR #5802, opened 2023-01-24
Fix a bug in AVX512 simd_mask::operator[]

ibaned:
this would cause a compiler error when
instantiated because of const-correctness

### DIFF
--- a/simd/src/Kokkos_SIMD_AVX512.hpp
+++ b/simd/src/Kokkos_SIMD_AVX512.hpp
@@ -114,7 +114,7 @@ class simd_mask<T, simd_abi::avx512_fixed_size<8>> {
   }
   KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION value_type
   operator[](std::size_t i) const {
-    return static_cast<value_type>(reference(m_value, int(i)));
+    return reference(const_cast<simd_mask*>(this)->m_value, int(i));
   }
   KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION simd_mask
   operator||(simd_mask const& other) const {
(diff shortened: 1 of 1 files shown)


simd/src/Kokkos_SIMD_AVX512.hpp
```diff
@@ -86,7 +86,7 @@ class simd_mask<T, simd_abi::avx512_fixed_size<8>> {
   }
   KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION value_type
   operator[](std::size_t i) const {
-    return static_cast<value_type>(reference(m_value, int(i)));
+    return reference(const_cast<simd_mask*>(this)->m_value, int(i));
```
crtrott, line 89: Should we just have a const_reference type which takes a const value reference as a parameter in its constructor? I know its more code but I do REALLY dislike casting const away from values. I actually in some ways want to argue for a compiler to have a "const means const" mode, where you can't const_cast and compilers don't have to assume non-const things alias const things.

crtrott (changes requested):

dalg24:
Please open against develop.  We will cherry-pick into the RC branch after it gets approved and merged.

simd/src/Kokkos_SIMD_AVX512.hpp
nmm0: I second this, I really don't like `const_cast`

ibaned: Sure! We don't even need the type really because it's just converted to `bool` immediately so I pushed a change to do this

masterleinad (approved):

Rombur (approved):

crtrott (approved):

PR #5807, opened 2023-01-24
Make `ALL_t` a "public" type `Kokkos::Impl:: -> Kokkos::`

dalg24:
See https://github.com/kokkos/kokkos-core-wiki/pull/273#issuecomment-1402171680

(Will expand description later.)

### DIFF
--- a/core/src/Kokkos_View.hpp
+++ b/core/src/Kokkos_View.hpp
@@ -496,7 +496,7 @@ namespace Kokkos {
 
 namespace {
 
-constexpr Kokkos::Impl::ALL_t ALL = Kokkos::Impl::ALL_t();
+constexpr Kokkos::ALL_t ALL = Kokkos::ALL_t();
 
 constexpr Kokkos::Impl::WithoutInitializing_t WithoutInitializing =
     Kokkos::Impl::WithoutInitializing_t();
--- a/core/src/Kokkos_View.hpp
+++ b/core/src/Kokkos_View.hpp
@@ -494,17 +494,11 @@ constexpr bool is_assignable(const Kokkos::View<ViewTDst...>& dst,
 
 namespace Kokkos {
 
-namespace {
+inline constexpr Kokkos::ALL_t ALL{};
 
-constexpr Kokkos::ALL_t ALL = Kokkos::ALL_t();
+inline constexpr Kokkos::Impl::WithoutInitializing_t WithoutInitializing{};
 
-constexpr Kokkos::Impl::WithoutInitializing_t WithoutInitializing =
-    Kokkos::Impl::WithoutInitializing_t();
-
-constexpr Kokkos::Impl::AllowPadding_t AllowPadding =
-    Kokkos::Impl::AllowPadding_t();
-
-}  // namespace
+inline constexpr Kokkos::Impl::AllowPadding_t AllowPadding{};
 
 /** \brief  Create View allocation parameter bundle from argument list.
  *
--- a/core/src/impl/Kokkos_ViewMapping.hpp
+++ b/core/src/impl/Kokkos_ViewMapping.hpp
@@ -464,7 +464,7 @@ struct SubviewExtents {
   KOKKOS_FORCEINLINE_FUNCTION bool set(unsigned domain_rank,
                                        unsigned range_rank,
                                        const ViewDimension<DimArgs...>& dim,
-                                       const ALL_t, Args... args) {
+                                       ALL_t, Args... args) {
     m_begin[domain_rank] = 0;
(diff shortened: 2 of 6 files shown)


core/src/Kokkos_View.hpp
```diff
@@ -494,17 +494,11 @@ constexpr bool is_assignable(const Kokkos::View<ViewTDst...>& dst,
 
 namespace Kokkos {
 
-namespace {
+inline constexpr Kokkos::ALL_t ALL{};
 
-constexpr Kokkos::Impl::ALL_t ALL = Kokkos::Impl::ALL_t();
+inline constexpr Kokkos::Impl::WithoutInitializing_t WithoutInitializing{};
```
dalg24, line 499: `inline` specifier is not strictly necessary.  It is implicit since C++17.  Let me know if you want me to drop it.

core/src/Kokkos_View.hpp
```diff
@@ -494,17 +494,11 @@ constexpr bool is_assignable(const Kokkos::View<ViewTDst...>& dst,
...
 
-constexpr Kokkos::Impl::WithoutInitializing_t WithoutInitializing =
```
crtrott, line 501: How was this ever usable by the public if it was in an anonymous namespace?

dalg24: just one per TU

dalg24: anonymous namespace means no collision at link time

core/src/Kokkos_View.hpp
fnrizzi: are you sure about c++17 implicitly inlining this? If i read here: https://en.cppreference.com/w/cpp/language/inline
i see "A static member variable (but **not a namespace-scope variable**) declared constexpr is implicitly an inline variable."

core/src/impl/Kokkos_ViewMapping.hpp
```diff
@@ -559,7 +559,7 @@ struct SubviewExtents {
   // std::pair range
   template <size_t... DimArgs, class... Args>
   void error(char* buf, int buf_len, unsigned domain_rank, unsigned range_rank,
-             const ViewDimension<DimArgs...>& dim, const Kokkos::Impl::ALL_t,
+             const ViewDimension<DimArgs...>& dim, const ALL_t,
```
fnrizzi, line 562: do we even need the `const` before ALL_t here?

core/src/Kokkos_View.hpp
dalg24: Looks like I was wrong.  Thanks for pointing that out

core/src/impl/Kokkos_ViewMapping.hpp
dalg24: No.  Want me to remove?

fnrizzi: i would say yes for consistency with the other args but no big deal

core/src/Kokkos_View.hpp
fnrizzi: I did not know that either, just learned it too ..

core/src/impl/Kokkos_ViewMapping.hpp
dalg24: Done

fnrizzi (approved):

Rombur (approved):

core/src/impl/Kokkos_ViewMapping.hpp
```diff
@@ -296,15 +295,19 @@ struct ALL_t {
   constexpr bool operator==(const ALL_t&) const { return true; }
 };
 
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+namespace Impl {
+using Kokkos::ALL_t;
```
masterleinad, line 300:
```suggestion
using All_t KOKKOS_DEPRECATED_WITH_COMMENT("Use Kokkos::ALL_t instead!") = Kokkos::ALL_t;
```

masterleinad: Yes, I agree. It's more for people to get some notification in case they were using it (even if they were not supposed to).

masterleinad (approved):

masterleinad:
`11: [  FAILED  ] defaultdevicetype.shared_space` looks unrelated.

nmm0 (approved):
Looks good to me! Why was this in Impl in the first place?

dalg24:
> Why was this in Impl in the first place?

Not sure

PR #5810, opened 2023-01-25
Move all SYCL headers into SYCL directory

masterleinad:
Cleaning up `core/src`.


core/src/SYCL/Kokkos_SYCL.cpp
```diff
@@ -20,7 +20,7 @@
 
 #include <Kokkos_Concepts.hpp>
 #include <SYCL/Kokkos_SYCL_Instance.hpp>
-#include <Kokkos_SYCL.hpp>
```
ldh4, line 23: It looks like there are currently some inconsistencies between where each backend's `Kokkos_[backend].hpp` is placed in.
`Cuda`, `HPX`, `OpenMPTarget` and `Threads` have their `Kokkos_[backend].hpp` files still in `core/src` whereas some of others are in their respective directory.
Is there a reason why `Kokkos_SYCL.hpp` was placed in `core/src` initially but is being moved now?

ldh4 (approved):

dalg24 (approved):

core/src/SYCL/Kokkos_SYCL.cpp
dalg24: You are right, we are inconsistent at the moment.
What we want is what is implemented in this PR.
HPX and OpenMPTarget have ongoing open refactor PRs which is the main reason why we decided not to do one big PR but defer to individual backend maintainers.

PR #5813, opened 2023-01-25
Try using oneAPI 2023.0.0 in SYCL+Cuda CI

masterleinad:
Follow-up to #5707. Using `oneAPI 2023.0.0` proper together with the `Codeplay` plugin, see https://developer.codeplay.com/products/oneapi/nvidia/2023.0.0/guides/get-started-guide-nvidia, should give us a more solid foundation of what we are testing (which should follow what is currently recommended for `SYCL+Cuda`). 

Additionally, this should improve build times for the docker image and avoids shenanigans used currently because the install scripts for `intel/llvm` didn't work anymore in our CI.

### DIFF
--- a/.jenkins
+++ b/.jenkins
@@ -101,12 +101,13 @@ pipeline {
                     }
                     steps {
                         sh 'ccache --zero-stats'
-                        sh '''rm -rf build && mkdir -p build && cd build && \
+                        sh '''. /opt/intel/oneapi/setvars.sh --include-intel-llvm && \
+                              rm -rf build && mkdir -p build && cd build && \
                               cmake \
                                 -DCMAKE_BUILD_TYPE=Release \
                                 -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
-                                -DCMAKE_CXX_COMPILER=clang++ \
-                                -DCMAKE_CXX_FLAGS="-fsycl-device-code-split=per_kernel -Werror -Wno-gnu-zero-variadic-macro-arguments -Wno-linker-warnings" \
+                                -DCMAKE_CXX_COMPILER=/opt/intel/oneapi/compiler/2023.0.0/linux/bin-llvm/clang++ \
+                                -DCMAKE_CXX_FLAGS="-fsycl-device-code-split=per_kernel -Werror -Wno-gnu-zero-variadic-macro-arguments -Wno-unknown-cuda-version -Wno-sycl-target" \
                                 -DKOKKOS_IMPL_SYCL_DEVICE_GLOBAL_SUPPORTED=0 \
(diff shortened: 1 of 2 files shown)


crtrott (approved):
Fine with me.

scripts/docker/Dockerfile.sycl
```diff
@@ -9,19 +9,7 @@ RUN apt-get update && apt-get install -y \
         ccache \
         ninja-build \
         python3 \
-        git \
-        && \
-    apt-get clean && \
-    rm -rf /var/lib/apt/lists/*
```
dalg24, line 15: Did you mean to drop that guy

dalg24:
Retest this please

.jenkins
```diff
@@ -101,12 +101,13 @@ pipeline {
...
                         sh 'ccache --zero-stats'
-                        sh '''rm -rf build && mkdir -p build && cd build && \
+                        sh '''. /opt/intel/oneapi/setvars.sh --include-intel-llvm && \
+                              rm -rf build && mkdir -p build && cd build && \
                               cmake \
                                 -DCMAKE_BUILD_TYPE=Release \
                                 -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
-                                -DCMAKE_CXX_COMPILER=clang++ \
-                                -DCMAKE_CXX_FLAGS="-fsycl-device-code-split=per_kernel -Werror -Wno-gnu-zero-variadic-macro-arguments -Wno-linker-warnings" \
+                                -DCMAKE_CXX_COMPILER=/opt/intel/oneapi/compiler/2023.0.0/linux/bin-llvm/clang++ \
```
dalg24, line 109: is not in the PATH?

masterleinad: No, it wasn't, `icpx` would be but then it would use the wrong runtime. Locally, I have `clang++` in my path after calling `. /opt/intel/oneapi/setvars.sh --include-intel-llvm` but I don't know why that wasn't enough in the container. I stopped worrying about it once the tests compiled and ran successfully.

dalg24: Fine.  Might be worth exploring sourcing that file in the container entrypoint.
Not a priority though.

Out of curiosity I looked how NVIDIA does its containers entry points
https://gitlab.com/nvidia/container-images/cuda/-/blob/7d552bd45c5d76c9945f655950437672dc7c1f72/nvidia_entrypoint.sh
https://gitlab.com/nvidia/container-images/cuda/-/tree/master/entrypoint.d

Haven't looked whether they are images curated by Intel or Codeplay.

dalg24 (approved):

dalg24:
SYCL build is passing.  The rest does not really matter.

Issue #5815, opened 2023-01-26
Trilinos compilation error (Tpetra) following removal of legacy atomic headers


ndellingwood:
**Describe the bug**
Following removal of the legacy atomic headers compilations of Trilinos began failing in Tpetra:

```
In file included from <https://jenkins-son.sandia.gov/job/KokkosEco_Trilinos_Weaver_CUDA112_Shared_Cplx/ws/Trilinos/packages/tpetra/core/src/Tpetra_MultiVector_def.hpp>:68,
                 from <https://jenkins-son.sandia.gov/job/KokkosEco_Trilinos_Weaver_CUDA112_Shared_Cplx/ws/Build/packages/tpetra/core/src/Tpetra_MultiVector_DOUBLE_INT_LONG_LONG_CUDA.cpp>:69:
<https://jenkins-son.sandia.gov/job/KokkosEco_Trilinos_Weaver_CUDA112_Shared_Cplx/ws/Trilinos/packages/tpetra/core/src/kokkos_refactor/Tpetra_KokkosRefactor_Details_MultiVectorDistObjectKernels.hpp>:61:10: fatal error: impl/Kokkos_Atomic_Generic.hpp: No such file or directory
 #include "impl/Kokkos_Atomic_Generic.hpp"
          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

This is the list of new commits since prior successful build:

Fix sycl.scratch_align test ([details](https://jenkins-son.sandia.gov/job/KokkosEco_Trilinos_Weaver_CUDA112_Shared_Cplx/116/changes#e8c08e2c0cf37786a65574b030679cceb7658108))
    Improve comments in TestTeam.hpp ([details](https://jenkins-son.sandia.gov/job/KokkosEco_Trilinos_Weaver_CUDA112_Shared_Cplx/116/changes#53cc29764415a73732e66a730501f01b3756cd2a))
    Purge logic around desul atomics being enabled at configuration time ([details](https://jenkins-son.sandia.gov/job/KokkosEco_Trilinos_Weaver_CUDA112_Shared_Cplx/116/changes#05cb3f5114897d622c9adaba7e0a17432b8e113d))
    Purge macro guards for desul atomics being enabled or not ([details](https://jenkins-son.sandia.gov/job/KokkosEco_Trilinos_Weaver_CUDA112_Shared_Cplx/116/changes#c3fe1d607b14f526a46c548ee6e1da17099dd554))
[...]

PR #5816, opened 2023-01-26
Add unit test capturing Tpetra custom atomics use case

dalg24:
Related to #5815 

In a nutshell, Tpetra wants a "atomic_max_abs" and they were using implementation details.  I suggested to go through a user-defined type that wraps the comparison operators in trilinos/trilinos#11506

We should probably include something in our unit tests to make sure we remember what they are doing.  This is PR attempts to do that.


core/unit_test/TestAtomics.hpp
```diff
@@ -539,4 +539,45 @@ TEST(TEST_CATEGORY, atomics) {
...
+  Kokkos::View<T> d_{"lbl"};
+  KOKKOS_FUNCTION void operator()(int i) const {
+    // 0, -1, 2, -3, ...
+    auto v_i = static_cast<T>(i);
+    if (i % 2 == 1) v_i = -v_i;
+    Kokkos::atomic_max(reinterpret_cast<WrapScalarAndCompareAbsMax<T>*>(&d_()),
+                       WrapScalarAndCompareAbsMax<T>{v_i});
+  }
+  TpetraUseCase() {
+    Kokkos::deep_copy(d_, Kokkos::Experimental::finite_min_v<T>);
```
masterleinad, line 571:
```suggestion
    Kokkos::deep_copy(d_, 0);
```
or
```suggestion
```
Otherwise, the initial value dominates.

dalg24: No I meant it, I am doing "max", this is the right identity value.

core/unit_test/TestAtomics.hpp
```diff
@@ -539,4 +539,45 @@ TEST(TEST_CATEGORY, atomics) {
...
+  };
+
+  using T = int;
+  Kokkos::View<T> d_{"lbl"};
+  KOKKOS_FUNCTION void operator()(int i) const {
+    // 0, -1, 2, -3, ...
+    auto v_i = static_cast<T>(i);
+    if (i % 2 == 1) v_i = -v_i;
+    Kokkos::atomic_max(reinterpret_cast<WrapScalarAndCompareAbsMax<T>*>(&d_()),
+                       WrapScalarAndCompareAbsMax<T>{v_i});
```
masterleinad, line 568:
```suggestion
    WrapScalarAndCompareAbsMax<T> tmp{d_()};
    Kokkos::atomic_max(&tmp, WrapScalarAndCompareAbsMax<T>{v_i});
    d_() = tmp.value;
```
to avoid undefined behavior, also see https://stackoverflow.com/questions/62765714/is-it-safe-to-reinterpret-cast-between-a-type-and-a-wrappertype. The `reinterpret_cast` is only guaranteed to have the expected behavior if `d_()` would already be wrapped by the temporary `WrapScalarAndCompareAbsMax`.

core/unit_test/TestAtomics.hpp
masterleinad: The CI proves you wrong. 🙂 For integer types `finite_min` is `INT_MIN` (min and lowest are the same).

dalg24: oh right...  ty

core/unit_test/TestAtomics.hpp
dalg24: Would not be atomic then.

masterleinad: Hmmm, yes, that's right.

masterleinad: Doesn't that mean that the functor should use `WrapScalarAndCompareAbsMax` directly then? Or do we think it's reasonable for the compiler to support this technically undefined behavior in the expected way?

dalg24: > Doesn't that mean that the functor should use `WrapScalarAndCompareAbsMax` directly then?

Sure but that would not help for the the Tpetra use case.

> Or do we think it's reasonable for the compiler to support this technically undefined behavior in the expected way?

Acknowledging that it is bad but expecting that it will work

dalg24:
See https://github.com/trilinos/Trilinos/pull/11506#discussion_r1089728089
I am not sure what else to suggest and that would not be making assumptions about how desul atomics are implemented (which is subject to change and hence not something to rely on)

Wondering if we we can make it work if we "abuse" `operator+=`, and binary `operator+` and go through `atomic_add` (not proud of this but don't have better idea at the moment...)

crtrott (approved):

crtrott:
The NVHPC test failure is weird ... how does this get now triggered but wasn't before??

core/unit_test/TestAtomics.hpp
mhoemmen:
```c++
 Kokkos::atomic_max(reinterpret_cast<WrapScalarAndCompareAbsMax<T>*>(&d_()),
                       WrapScalarAndCompareAbsMax<T>{v_i});
```

I ... didn't write that.  Just gonna put that out there in case somebody claims it's my fault ; - P

dalg24:
```
1: [ RUN      ] serial.atomics_tpetra_max_abs
1: /var/jenkins/workspace/Kokkos/core/unit_test/TestAtomics.hpp:580: Failure
1: Expected equality of these values:
1:   v
1:     Which is: 0
1:   9
1: [  FAILED  ] serial.atomics_tpetra_max_abs (28 ms)
```

dalg24:
Retest this please (I don't have a better idea)

dalg24:
SYCL failure is clearly unrelated.

masterleinad (approved):
Fine with me.

PR #5817, opened 2023-01-26
Drop Kokkos device lock arrays

dalg24:
Addressing step 3 in #5789

Device lock arrays were used for lock-based atomic operations and these are now handled by Desul.


dalg24:
OK this is safe to merge.  Sacado was updated in trilinos/trilinos#11597

janciesko (approved):
LGTM

dalg24:
Ignoring SYCL failure
```
In file included from <built-in>:1:
/tmp/Lock_Array_CUDA-header-49aa7e.h:25:36: error: zero size arrays are an extension [-Werror,-Wzero-length-array]
const char* const kernel_names[] = {
                                   ^
1 error generated.

PR #5818, opened 2023-01-26
Spell out Kokkos::ALL_t to avoid deprecation warnings

masterleinad:
Deprecations don't get caught by the CI.

Follow-up to #5807

### DIFF
--- a/core/src/Kokkos_CopyViews.hpp
+++ b/core/src/Kokkos_CopyViews.hpp
@@ -874,7 +874,7 @@ struct ViewRemap<DstType, SrcType, ExecSpace, 2> {
       } else {
         p_type ext1(0, std::min(dst.extent(1), src.extent(1)));
         using sv_adapter_type =
-            CommonSubview<DstType, SrcType, 2, ALL_t, p_type>;
+            CommonSubview<DstType, SrcType, 2, Kokkos::ALL_t, p_type>;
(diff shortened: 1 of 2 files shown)


dalg24:
What compilers are warning?  I would have sworn AppleClang did not yell at me...

masterleinad:
> What compilers are warning? I would have sworn AppleClang did not yell at me...

`Apple clang version 14.0.0 (clang-1400.0.29.202)` complained for me

PhilMiller:
To be clear, the warnings arose because we `[[deprecated]]` `Kokkos::Impl::ALL_t`, and so stuff in that namespace sees the deprecated name and not the non-deprecated public name?

masterleinad:
> To be clear, the warnings arose because we `[[deprecated]]` `Kokkos::Impl::ALL_t`, and so stuff in that namespace sees the deprecated name and not the non-deprecated public name?

Yes, but, of course, it referred to `Kokkos::Impl::ALL_t` all the time. Deprecating it didn't change that. 🙂

dalg24:
I'd rather drop the [[deprecated]] attribute than qualify everywhere.

dalg24 (changes requested):
Add a comment by the deprecated alias to cleanup and remove our qualifying Kokkos::ALL_t when we remove the alias.
Drop the changes to .jenkins.

dalg24 (approved):

crtrott (approved):

Issue #4668, opened 2022-01-12
Deprecate/remove `KOKKOS_ACTIVE_EXECUTION_MEMORY_SPACE_*` macros and the `ActiveExecutionMemorySpace` alias


dalg24:
`ActiveExecutionMemorySpace` is defined in `Kokkos::Impl::` but it is being used downstream in KokkosKernels.

The macros are most commonly used this way
```C++
#ifdef KOKKOS_ACTIVE_EXECUTION_MEMORY_SPACE_HOST
<host code>
#else
<device code>
#endif
```
which will be replaced by
```C++
KOKKOS_IF_ON_HOST((<host code>))
KOKKOS_IF_ON_DEVICE((<device code>))
```
in #4660

PR #5819, opened 2023-01-26
Deprecate `KOKKOS_ACTIVE_EXECUTION_MEMORY_SPACE_*` macros

dalg24:
Oversight in release 3.6
We still define these macros and code downstream are still using them.
See trilinos/Trilinos#11511 trilinos/Trilinos#11508,
trilinos/Trilinos#11511 trilinos/Trilinos#11509,
trilinos/Trilinos#11511 trilinos/Trilinos#11510, and
trilinos/Trilinos#11511 trilinos/Trilinos#11511

Sacado is also using it.  Haven't reported it yet because I am trying to figure out how to get them out of using `Kokkos::Impl::lock_address_*_space` and the two issues are intertwined.

(**edit**: Forgot to mention KokkosKernels still somehow using it in the infamous `<Kokkos_ArithTraits.hpp>` header.)
(**edit2**: Adding reference to #4668 that clearly stated our intent to get rid of these macros)

Our CHANGELOG clearly list these `KOKKOS_ACTIVE_EXECUTION_MEMORY_SPACE_*` macros in the list of deprecation for release 3.6.
I opted for the cautious approach and guarded with `#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4` but I am tempted to retro deprecate and use `KOKKOS_ENABLE_DEPRECATED_CODE_3`.  Any thought on this?

### DIFF
--- a/core/src/Kokkos_Macros.hpp
+++ b/core/src/Kokkos_Macros.hpp
@@ -451,7 +451,7 @@
 
 //----------------------------------------------------------------------------
 // Determine for what space the code is being compiled:
-#if defined(KOKKOS_ENABLE_DEPRECARED_CODE_4)
+#if defined(KOKKOS_ENABLE_DEPRECARED_CODE_3)
 
 #if defined(__CUDACC__) && defined(__CUDA_ARCH__) && defined(KOKKOS_ENABLE_CUDA)
 #define KOKKOS_ACTIVE_EXECUTION_MEMORY_SPACE_CUDA
--- a/core/src/Kokkos_Macros.hpp
+++ b/core/src/Kokkos_Macros.hpp
@@ -451,6 +451,7 @@
 
 //----------------------------------------------------------------------------
 // Determine for what space the code is being compiled:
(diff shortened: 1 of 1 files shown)


masterleinad:
In my opinion, `KOKKOS_ENABLE_DEPRECATED_CODE_3` is a good compromise. There are a few features that still depend on that flag (and haven't been removed yet).

PhilMiller:
If they were described as deprecated in 3.6, then they only kind of fall under our promise of code that compiled without deprecation warnings on 3.7 still working with 4.x. I would favor retro deprecation, and removal as soon as Trilinos can be cleared up.

dalg24:
Updated the PR.  Christian mentioned he also is in favor of DEPRECATED_CODE_3.
Haven't heard anyone objecting so far.

PhilMiller:
Are we going to want this in 4.0? /sigh

crtrott:
Yeah we may want this in 4.0 /sigh and yeah I am ok with deprecated 3

crtrott (approved):

masterleinad (approved):

PR #5821, opened 2023-01-27
Disable global device variables in SYCL+Cuda CI

masterleinad:
Apparently, using `sycl::device_global` variables with the compiler versions in the `SYCL+Cuda` CI results in all kinds of problems (even though the feature detection compiles successfully). Let's just disable that feature for now in the CI manually.

### DIFF
--- a/.jenkins
+++ b/.jenkins
@@ -107,6 +107,7 @@ pipeline {
                                 -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
                                 -DCMAKE_CXX_COMPILER=clang++ \
                                 -DCMAKE_CXX_FLAGS="-fsycl-device-code-split=per_kernel -Werror -Wno-gnu-zero-variadic-macro-arguments -Wno-linker-warnings" \
+                                -DKOKKOS_IMPL_SYCL_DEVICE_GLOBAL_SUPPORTED=0 \
                                 -DKokkos_ARCH_NATIVE=ON \
                                 -DKokkos_ARCH_VOLTA70=ON \
                                 -DKokkos_ENABLE_COMPILER_WARNINGS=ON \


crtrott (approved):
Ok with me.

masterleinad:
For the failing `GitHub` CI runs, I am seeing
```

The job was not started because recent account payments have failed or your spending limit needs to be increased. Please check the 'Billing & plans' section in your settings.
```
I'm not entirely sure what's going on there.

.jenkins
```diff
@@ -107,6 +107,7 @@ pipeline {
                                 -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
                                 -DCMAKE_CXX_COMPILER=clang++ \
                                 -DCMAKE_CXX_FLAGS="-fsycl-device-code-split=per_kernel -Werror -Wno-gnu-zero-variadic-macro-arguments -Wno-linker-warnings" \
+                                -DKOKKOS_IMPL_SYCL_DEVICE_GLOBAL_SUPPORTED=0 \
```
dalg24, line 110: This is a such a hack...

dalg24 (changes requested):
> Apparently, using `sycl::device_global` variables with the compiler versions in the `SYCL+Cuda` CI results in all kinds of problems (even though the feature detection compiles successfully). Let's just disable that feature for now in the CI manually.

I am not following.  What failures is this resolving?  Do you mean if we try another version of the toolchain?

dalg24:
Is that PR related to #5813?

masterleinad:
> Is that PR related to #5813?

No, I also get more test failures when using the `oneAPi 2023.0.0` release. It appears that `sycl::device_gloabl` is just not working for `SYCL+Cuda` (even though the feature test compiles).

.jenkins
masterleinad: Yes, this is mostly a quick fix. I'm happy to look into improving the feature check later.

masterleinad:
> I am not following. What failures is this resolving?

First, we get strange warnings like
```
In file included from <built-in>:1:
/tmp/Lock_Array_CUDA-header-b6db19.h:25:36: error: zero size arrays are an extension [-Werror,-Wzero-length-array]
const char* const kernel_names[] = {
                                   ^
1 error generated.
core/src/CMakeFiles/kokkoscore.dir/build.make:369: recipe for target 'core/src/CMakeFiles/kokkoscore.dir/__/__/tpls/desul/src/Lock_Array_CUDA.cpp.o' failed
make[2]: *** [core/src/CMakeFiles/kokkoscore.dir/__/__/tpls/desul/src/Lock_Array_CUDA.cpp.o] Error 1
In file included from <built-in>:1:
/tmp/Lock_Array_HIP-header-30993a.h:25:36: error: zero size arrays are an extension [-Werror,-Wzero-length-array
const char* const kernel_names[] = {
                                   ^
1 error generated.
```
after ignoring those, every test case is failing with something like `GlobalDevice already initialized. Removing "split by kernel" also didn't help and for the `oneAPI 2023.0.0` release the tests were at least running but some of the atomic tests were timing out.

dalg24:
Did you consider doing "if arch not intel" then turn it off in `cmake/kokkos_arch.cmake`?

dalg24 (approved):
Merging as all the SYCL builds are failing at the moment and this fixes it.
Daniel will investigate and try to come up with a more durable solution.

PR #5822, opened 2023-01-29
Escape double quotes in git description

dalg24:
Realize this was an issue while having the branch from #5816 checked out.
```
<build>/generated/Kokkos_Version_Info.cpp:25:63: error: expected ';' after top level declarator
std::string GIT_COMMIT_DESCRIPTION = "Fixup turns out Tpetra "abs max" operation does not preserve the sign";
```
Haven't looked yet why this is not caught by the CI.
In any case here suggesting we escape the double quote character which in the above case would yield
```C++
std::string GIT_COMMIT_DESCRIPTION = "Fixup turns out Tpetra \"abs max\" operation does not preserve the sign";
```

PS: to be honest I still don't see the point of capturing the git commit message (nor the date), if it was up to me I would drop them altogether.


cz4rs:
I don't see much use for the commit date either, commit message on the other hand is somewhat useful when displayed after hovering your mouse over data point:
image

cz4rs (approved):

dalg24:
> Haven't looked yet why this is not caught by the CI.

In the Jenkins logs, I found
```
--     Could NOT find Git (missing: GIT_EXECUTABLE) 
```
and Cezary pointed out that when they are no conflicts, we actually test the PR merged into develop which means the commit message is not used.

dalg24:
Another option would be a raw string literal.  Let me know which solution you prefer.

cmake/build_env_info.cmake
```diff
@@ -53,6 +53,8 @@ FUNCTION(check_git_version)
     WORKING_DIRECTORY ${KOKKOS_SOURCE_DIR}
     OUTPUT_VARIABLE GIT_COMMIT_DESCRIPTION
     OUTPUT_STRIP_TRAILING_WHITESPACE)
+  string(REPLACE "\"" "\\\"" GIT_COMMIT_DESCRIPTION_ESC_DOUBLE_QUOTES "${GIT_COMMIT_DESCRIPTION}")
+  set(GIT_COMMIT_DESCRIPTION ${GIT_COMMIT_DESCRIPTION_ESC_DOUBLE_QUOTES})
```
masterleinad, line 57:
```suggestion
  # escape double quotes
  string(REPLACE "\"" "\\\"" GIT_COMMIT_DESCRIPTION "${GIT_COMMIT_DESCRIPTION}")
```

cz4rs (changes requested):
> Another option would be a raw string literal. Let me know which solution you prefer.

I think raw string literal would be an even better option here (more concise, does the trick for any other potentially problematic characters as well).

dalg24:
Closing in favor of #5823

PR #5825, opened 2023-01-30
Desul atomics: let pointer to the device lock arrays always be in constant memory

dalg24:
Was in constant memory only when relocatable device code is enabled.

Making sure it actually works before I open a PR on the desul side.


dalg24:
CI is a mess at the moment but all Cuda and HIP builds are passing.

Rombur (approved):
I talked to Damien and I think it's OK but just in case here is a [link](https://forums.developer.nvidia.com/t/constant-memory-variables-give-incorrect-value-when-declared-defined-and-used-in-separate-files/59582) to a potential problem when using `__constant__` without rdc. We shouldn't have a problem but it's better to be aware of this risk

crtrott (approved):

PR #5809, opened 2023-01-25
Fix flags passed to ARM compiler and NVHPC when `Kokkos_ARCH_NATIVE=ON`

dalg24:
Fix #5808 

Haven't actually tested.
Also not sure yet what flags we should be using with the NVHPC compiler.

cc @antoine-morvan does that do what you want?

Issue #5808, opened 2023-01-25
ARCH=NATIVE not supported with ARM Compiler suite & NVHPC on ARM Systems


antoine-morvan:
Hello, 

I am trying Kokkos on an ARM platform using the ARM compiler for Linux (https://developer.arm.com/downloads/-/arm-compiler-for-linux). Also I am using the NATIVE backend. Using this combination, I get the following error during the build:
```
error: the clang compiler does not support '-march=native'
```

Indeed march/mtune/mcpu flags handling is different among arch/compilers (see this article for instance https://community.arm.com/arm-community-blogs/b/tools-software-ides-blog/posts/compiler-flags-across-architectures-march-mtune-and-mcpu)

On ARM we expect the flag to be `-mcpu=native` instead of having `-mtune=native -march=native`. I encountered similar issue when using the NVidia suite (nvhpc 22.9) on ARM platform (`nvc++-Error-Unknown switch: -mtune=native`).

I did not find any reference on ARCH=NATIVE on ARM platforms in the Kokkos documentation, and also tried few keywords in the tracker without any luck. Is there any workaround ?

Below is a sample script to reproduce the issue (just have to adapt compiler env setup).

PS: I would like the atos.net domain to be authorized on the slack or get an invitation 😇

Best.

## compilers (with versions)
arm-compiler-for-linux_22.1_RHEL-8_aarch64

## Kokkos release or commit used (i.e., the sha1 number)
3.7.1 targz from github releases

## platform and backend

Using Kokkos_ARCH=NATIVE on ARM systems:
* Ampere Altra Q8030 (Neoverse N1)
* Fujitsu FX700 (A64FX)

## cmake configure command
```
(cd build && cmake \
        -DKokkos_ARCH_NATIVE=ON \
        -G "Unix Makefiles" \
        ../kokkos-3.7.01)
```

## output from cmake command
```
-- Setting default Kokkos CXX standard to 14
-- The CXX compiler identification is Clang 13.0.1
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: /software/arm/armclang-22.1/arm-linux-compiler-22.1_Generic-AArch64_RHEL-8_aarch64-linux/bin/armclang++ - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Setting build type to 'RelWithDebInfo' as none was specified.
-- The project name is: Kokkos
[...]

PR #5831, opened 2023-01-31
Set native flags according to CMAKE_SYSTEM_PROCESSOR

masterleinad:
Alternative to #5809, fixing #5808, following https://github.com/kokkos/kokkos/pull/5809#issuecomment-1410832647.
For x86 targets, we use `-march=native -mtune=native` and otherwise `-mcpu=native`.


cmake/kokkos_functions.cmake
```diff
@@ -827,7 +827,7 @@ FUNCTION(kokkos_link_tpl TARGET)
 ENDFUNCTION()
 
 FUNCTION(COMPILER_SPECIFIC_OPTIONS_HELPER)
-  SET(COMPILERS NVIDIA NVHPC DEFAULT Cray Intel Clang AppleClang IntelLLVM GNU HIPCC Fujitsu MSVC)
+  SET(COMPILERS NVIDIA NVHPC DEFAULT Cray Intel Clang AppleClang IntelLLVM GNU HIPCC Fujitsu MSVC ARMClang)
```
dalg24, line 830: No point on adding
```suggestion
  SET(COMPILERS NVIDIA NVHPC DEFAULT Cray Intel Clang AppleClang IntelLLVM GNU HIPCC Fujitsu MSVC)
```

cmake/kokkos_arch.cmake
```diff
@@ -229,9 +229,14 @@ IF(KOKKOS_ARCH_NATIVE)
     MESSAGE(FATAL_ERROR "MSVC doesn't support ARCH_NATIVE!")
   ENDIF()
 
+  IF(CMAKE_SYSTEM_PROCESSOR MATCHES "(x86)|(X86)|(amd64)|(AMD64)")
```
dalg24, line 232: Consider making the string all uppercase before to avoid vexing `(Amd64)` not matching

masterleinad:
Only `CUDA-11.7-NVHPC` is failing in
```
- KokkosCore_UnitTest_CudaTimingBased (Failed)
```

masterleinad:
@antoine-morvan Would you mind trying this and report back?

antoine-morvan:
Works fine on x86_64 and aarch64. Tried:
* x86_64: GCC 12; LLVM 14 & 15; OneAPI 2022.3 (icpc & icpx); NVHPC 22.9; AOCC 4.0
* aarch64: GCC 12; LLVM 14 & 15; ARMClang 22.1

Failing with NVHPC 22.9 on aarch64, as `-mcpu/march/mtune` are alias for `-tp` ... but not present on all releases ...
This commit would most likely fix the issue: https://github.com/kokkos/kokkos/commit/3c5e0530d82f807c338b674e51ec704a989b6397

dalg24 (approved):
Lets merge as is and handle NVPHC in a follow-up PR

crtrott (approved):

PR #5832, opened 2023-01-31
Fix macros missing KOKKOS_ prefix, unnecessarily left defined

PhilMiller:


core/src/Kokkos_Crs.hpp
```diff
@@ -304,11 +304,11 @@ struct CountAndFillBase {
   Functor m_functor;
   counts_type m_counts;
   struct Count {};
-  inline void operator()(Count, size_type i) const {
+  KOKKOS_INLINE_FUNCTION void operator()(Count, size_type i) const {
```
masterleinad, line 307: In the spirit of removing `inline` from functions being defined in the class declaration.
```suggestion
  KOKKOS_FUNCTION void operator()(Count, size_type i) const {
```

core/src/Kokkos_Crs.hpp
```diff
@@ -304,11 +304,11 @@ struct CountAndFillBase {
...
     m_counts(i) = m_functor(i, nullptr);
   }
   struct Fill {};
-  inline void operator()(Fill, size_type i) const {
+  KOKKOS_INLINE_FUNCTION void operator()(Fill, size_type i) const {
```
masterleinad, line 311:
```suggestion
  KOKKOS_FUNCTION void operator()(Fill, size_type i) const {
```

core/src/Kokkos_Crs.hpp
PhilMiller: I've probably missed some memo or discussion, but why is that preferable, or what's the guideline?

dalg24: Do not inline function that are defined in the body of a class

PhilMiller: Ok, will revise

core/src/Cuda/Kokkos_Cuda_ReduceScan.hpp
```diff
@@ -62,33 +62,37 @@ template <class ValueType, class ReducerType>
 __device__ inline void cuda_inter_warp_reduction(
     ValueType& value, const ReducerType& reducer,
     const int max_active_thread = blockDim.y) {
-#define STEP_WIDTH 4
+#define KOKKOS_IMPL_STEP_WIDTH 4
```
masterleinad, line 65: I don't see a good reason to use a macro here in the first place.
```suggestion
constexpr int step_width = 4;
```

PhilMiller: Changed.

PhilMiller:
This rebases cleanly onto `release-candidate-4.0` if we want it there.

masterleinad (approved):

rgayatri23:
LGTM for OpenMPTarget stuff.

rgayatri23 (approved):

dalg24 (approved):
For the record Bruno had written on nucleus in the discussion that preceded that PR

> HIP_ENABLE_PRINTF can be removed. It was necessary for older versions of the compiler

crtrott:
The failure is with random test so its likely fine.

crtrott (approved):

PhilMiller:
Do you want me to post the 4.0 PR as well?

dalg24:
Sure

PhilMiller:
4.0 cherry-pick posted in #5842.

PR #5833, opened 2023-01-31
Restrict KOKKOS_IMPL_SYCL_DEVICE_GLOBAL_SUPPORTED feature macro detection to static libraries

masterleinad:
We are seeing test failures on the testbeds if that feature was enabled together with building shared libraries.


dalg24 (approved):
OK (assuming that when `BUILD_SHARED_LIBS` is not defined we build static libraries)

ldh4 (approved):

PR #5834, opened 2023-01-31
Fix unprefixed macros in KokkosExp_Host_IterateTile.hpp

masterleinad:
Similar to #5832.

### DIFF
--- a/core/src/impl/KokkosExp_Host_IterateTile.hpp
+++ b/core/src/impl/KokkosExp_Host_IterateTile.hpp
@@ -37,1167 +37,1278 @@ namespace Impl {
 // Temporary, for testing new loop macros
 #define KOKKOS_ENABLE_NEW_LOOP_MACROS 1
 
-#define LOOP_1L(type, tile)   \
-  KOKKOS_ENABLE_IVDEP_MDRANGE \
+#define KOKKOS_IMPL_LOOP_1L(type, tile) \
+  KOKKOS_ENABLE_IVDEP_MDRANGE           \
   for (type i0 = 0; i0 < static_cast<type>(tile[0]); ++i0)
 
-#define LOOP_2L(type, tile) \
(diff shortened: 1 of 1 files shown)


dalg24 (approved):

core/src/impl/KokkosExp_Host_IterateTile.hpp
```diff
@@ -2721,6 +2848,162 @@ struct HostIterateTile<RP, Functor, Tag, ValueType,
...
+#undef KOKKOS_IMPL_LOOP_1R
+#undef KOKKOS_IMPL_LOOP_2R
+#undef KOKKOS_IMPL_LOOP_3R
+#undef KOKKOS_IMPL_LOOP_4R
+#undef KOKKOS_IMPL_LOOP_5R
+#undef KOKKOS_IMPL_LOOP_6R
+#undef KOKKOS_IMPL_LOOP_7R
+#undef KOKKOS_IMPL_LOOP_8R
+#define KOKKOS_IMPL_LOOP_ARGS_1 i0 + m_offset[0]
+#define KOKKOS_IMPL_LOOP_ARGS_2 KOKKOS_IMPL_LOOP_ARGS_1, i1 + m_offset[1]
```
Rombur, line 2868: Why do you define this? You already define it [earlier](https://github.com/kokkos/kokkos/pull/5834/files?w=1#diff-5a440c37ef49c72e9af80d863acbebe1e4b4eff34464ba94ecedebe303cc34c9R104)

dalg24: Good that someone was less lazy than I was and scrolled to the bottom.  These should be `#undef`s

masterleinad: Fixed.

Rombur (approved):

PR #5836, opened 2023-02-01
Make OpenMP::concurrency and impl_thread_pool_size non-static

Rombur:
When we merged `partition_space`  (#5105) we decided to make non-static in a following PR. Here it is.

### DIFF
--- a/core/src/OpenMP/Kokkos_OpenMP.cpp
+++ b/core/src/OpenMP/Kokkos_OpenMP.cpp
@@ -72,10 +72,10 @@ void OpenMP::print_configuration(std::ostream &os, bool /*verbose*/) const {
 
 #ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
 int OpenMP::concurrency(OpenMP const &instance) {
-  return impl_thread_pool_size(instance);
+  return instance.impl_thread_pool_size();
 }
 #else
-int OpenMP::concurrency() const { return impl_thread_pool_size(*this); }
+int OpenMP::concurrency() const { return impl_thread_pool_size(); }
 #endif
 
 void OpenMP::fence(const std::string &name) const {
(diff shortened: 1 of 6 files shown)


core/src/OpenMP/Kokkos_OpenMP_WorkGraphPolicy.hpp
```diff
@@ -49,7 +49,8 @@ class ParallelFor<FunctorType, Kokkos::WorkGraphPolicy<Traits...>,
     // We need to introduce pool_size to work around NVHPC 22.5 ICE
     // We need to use [[maybe_unused]] to work around an unused-variable warning
     // from HIP
-    [[maybe_unused]] int pool_size = OpenMP::impl_thread_pool_size();
+    OpenMP exec;
```
Rombur, line 52: I need to check the CI to know if `exec` also need `[[maybe_unused]]`

core/src/OpenMP/Kokkos_OpenMP_UniqueToken.hpp
```diff
@@ -22,26 +22,31 @@
...
 template <>
 class UniqueToken<OpenMP, UniqueTokenScope::Instance> {
+ public:
+  using execution_space = OpenMP;
+  using size_type       = int;
+
  private:
   using buffer_type = Kokkos::View<uint32_t*, Kokkos::HostSpace>;
-  int m_count;
+  execution_space const& m_exec;
```
dalg24, line 31: Why a const reference rather than a copy?  This implicitly deletes the copy and move assignment operators and makes it harder to reason about that type.

core/src/OpenMP/Kokkos_OpenMP.hpp
```diff
@@ -127,13 +131,13 @@ class OpenMP {
   /// \brief Free any resources being consumed by the default execution space
   static void impl_finalize();
 
-  static int impl_thread_pool_size(OpenMP const& = OpenMP()) noexcept;
+  int impl_thread_pool_size() const noexcept;
+
+  inline int impl_thread_pool_size(int depth) const;
```
dalg24, line 136: Drop the inline.  It is only needed in the definition, not the declaration.
```suggestion
  int impl_thread_pool_size(int depth) const;
```

core/src/OpenMP/Kokkos_OpenMP_UniqueToken.hpp
Rombur: I hesitated between copy and reference. I used reference because it seemed safer. This ways I know that the execution space was initialized correctly.

core/src/OpenMP/Kokkos_OpenMP_WorkGraphPolicy.hpp
Rombur: The CI seems happy with the current code

dalg24 (approved):

ldh4 (approved):

core/src/OpenMP/Kokkos_OpenMP.cpp
```diff
@@ -94,17 +98,16 @@ bool OpenMP::in_parallel(OpenMP const &exec_space) noexcept {
 #endif
 }
 
-int OpenMP::impl_thread_pool_size(OpenMP const &exec_space) noexcept {
+int OpenMP::impl_thread_pool_size() const noexcept {
 #ifdef KOKKOS_ENABLE_DEPRECATED_CODE_3
   return OpenMP::in_parallel(exec_space)
```
dalg24, line 103: Uggg `exec_space` is not defined

PR #5841, opened 2023-02-01
Desul atomics: cleanup device lock arrays

dalg24:
* Do not attempt to compile the source files when the respective backend is not enabled (already doing the right thing in the Makefiles)
* Drop unnecessary macro guards that make the code difficult to read
* Fixup for #5613 (was not in sync with what actually got merged into desul)


masterleinad (approved):

tpls/desul/include/desul/atomics/Lock_Array_SYCL.hpp
```diff
@@ -128,13 +128,26 @@ inline void unlock_address_sycl(void* ptr, MemoryScopeNode) {
   lock_node_ref.exchange(0);
 }
 #else
+
+template <typename /*AlwaysInt*/ = int>
```
crtrott, line 132: If I get this right, this is the case where locks are not supported right? May want a comment here.

crtrott (changes requested):

tpls/desul/include/desul/atomics/Lock_Array_SYCL.hpp
dalg24: Maybe so, except that this is the sync'ing back up with Desul, I am not technically proposing any change to the SYCL backend.  See the description.

crtrott (approved):

tpls/desul/include/desul/atomics/Lock_Array_SYCL.hpp
dalg24: Added a comment

PR #5847, opened 2023-02-02
Fixup deprecated code 3 code path `OpenMP::impl_thread_pool_size`

dalg24:
Fixup for #5836
See https://github.com/kokkos/kokkos/pull/5836/files#r1094841991

cc @stanmoore1 
Reported on Slack

### DIFF
--- a/core/src/OpenMP/Kokkos_OpenMP.cpp
+++ b/core/src/OpenMP/Kokkos_OpenMP.cpp
@@ -100,7 +100,7 @@ bool OpenMP::in_parallel(OpenMP const &exec_space) noexcept {
 
 int OpenMP::impl_thread_pool_size() const noexcept {
 #ifdef KOKKOS_ENABLE_DEPRECATED_CODE_3
-  return OpenMP::in_parallel(exec_space)
+  return OpenMP::in_parallel()
              ? omp_get_num_threads()
              : (Impl::t_openmp_instance
                     ? Impl::t_openmp_instance->m_pool_size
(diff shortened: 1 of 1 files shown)


masterleinad (commented):
Shouldn't we pass `*this` to `OpenMP::in_parallel`?

Rombur (approved):

core/src/OpenMP/Kokkos_OpenMP.cpp
```diff
@@ -100,7 +100,7 @@ bool OpenMP::in_parallel(OpenMP const &exec_space) noexcept {
 
 int OpenMP::impl_thread_pool_size() const noexcept {
 #ifdef KOKKOS_ENABLE_DEPRECATED_CODE_3
-  return OpenMP::in_parallel(exec_space)
+  return OpenMP::in_parallel()
```
crtrott, line 103:
```suggestion
  return in_parallel(*this)
```

crtrott (commented):
Yeah i think Daniels suggestions sounds good.

stanmoore1:
@dalg24 can you fix the Makefile issue too (shouldn't enable deprecated 3)

dalg24:
> @dalg24 can you fix the Makefile issue too (shouldn't enable deprecated 3)

Not as part of this PR.

Also it is kinda awkward as we do not differentiate between deprecated code 3 and 4
https://github.com/kokkos/kokkos/blob/cba11a18fd70e8cd1c5069c30a9eaad96bab910e/Makefile.kokkos#L523-L526

Rombur (approved):

core/src/OpenMP/Kokkos_OpenMP.cpp
dalg24: Done

stanmoore1:
> Also it is kinda awkward as we do not differentiate between deprecated code 3 and 4
> 
> https://github.com/kokkos/kokkos/blob/cba11a18fd70e8cd1c5069c30a9eaad96bab910e/Makefile.kokkos#L523-L526

Yeah that is what I mean.

masterleinad (approved):

dalg24:
Makefile updated in #5850

PR #5850, opened 2023-02-02
Don't enable deprecated code 3 in Makefile builds anymore

crtrott:

PR #5851, opened 2023-02-02
[4.0.0] Don't enable deprecated code 3 in Makefile builds anymore

crtrott:
Cherry-pick #5850 into the RC 4.0 release branch


dalg24:
Ignoring NVHPC failure (CUDA timing based perf test)
The build that actually uses makefiles passed.

PR #5852, opened 2023-02-02
OpenMPTarget: Adding declare target for constexpr variables on Intel GPUs.

rgayatri23:
Co-authored-by: Daniel Arndt <arndtd@ornl.gov>

### DIFF
--- a/core/src/Kokkos_View.hpp
+++ b/core/src/Kokkos_View.hpp
@@ -496,13 +496,13 @@ namespace Kokkos {
 
 // FIXME_OPENMPTARGET - The `declare target` is needed for the Intel GPUs with
 // the OpenMPTarget backend
-#if defined(KOKKOS_ENABLE_OPENMPTARGET) && defined(KOKKOS_ARCH_INTEL)
+#if defined(KOKKOS_ENABLE_OPENMPTARGET) && defined(KOKKOS_COMPILER_INTEL)
 #pragma omp declare target
 #endif
 
 inline constexpr Kokkos::ALL_t ALL{};
 
-#if defined(KOKKOS_ENABLE_OPENMPTARGET) && defined(KOKKOS_ARCH_INTEL)
+#if defined(KOKKOS_ENABLE_OPENMPTARGET) && defined(KOKKOS_COMPILER_INTEL)
 #pragma omp end declare target
 #endif
 
--- a/core/src/Kokkos_View.hpp
+++ b/core/src/Kokkos_View.hpp
@@ -496,13 +496,13 @@ namespace Kokkos {
 
 // FIXME_OPENMPTARGET - The `declare target` is needed for the Intel GPUs with
 // the OpenMPTarget backend
-#if defined(KOKKOS_ENABLE_OPENMPTARGET) && defined(KOKKOS_ARCH_INTEL_PVC)
+#if defined(KOKKOS_ENABLE_OPENMPTARGET) && defined(KOKKOS_ARCH_INTEL)
 #pragma omp declare target
 #endif
 
 inline constexpr Kokkos::ALL_t ALL{};
 
-#if defined(KOKKOS_ENABLE_OPENMPTARGET) && defined(KOKKOS_ARCH_INTEL_PVC)
+#if defined(KOKKOS_ENABLE_OPENMPTARGET) && defined(KOKKOS_ARCH_INTEL)
 #pragma omp end declare target
 #endif
 
--- a/core/src/Kokkos_View.hpp
+++ b/core/src/Kokkos_View.hpp
@@ -494,7 +494,17 @@ constexpr bool is_assignable(const Kokkos::View<ViewTDst...>& dst,
 
 namespace Kokkos {
 
+// FIXME_OPENMPTARGET - The `declare target` is needed for the Intel GPUs with
+// the OpenMPTarget backend
+#if defined(KOKKOS_ENABLE_OPENMPTARGET) && defined(KOKKOS_ARCH_INTEL_PVC)
+#pragma omp declare target
+#endif
+
 inline constexpr Kokkos::ALL_t ALL{};
+
+#if defined(KOKKOS_ENABLE_OPENMPTARGET) && defined(KOKKOS_ARCH_INTEL_PVC)
+#pragma omp end declare target
+#endif
 
 inline constexpr Kokkos::Impl::WithoutInitializing_t WithoutInitializing{};
 
--- a/core/src/impl/Kokkos_Atomic_Memory_Order.hpp
+++ b/core/src/impl/Kokkos_Atomic_Memory_Order.hpp
@@ -51,7 +51,18 @@ struct memory_order_relaxed_t {
 #endif
   static constexpr auto std_constant = std::memory_order_relaxed;
 };
+
+// FIXME_OPENMPTARGET - The `declare target` is needed for the Intel GPUs with
+// the OpenMPTarget backend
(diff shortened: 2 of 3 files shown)


core/src/impl/Kokkos_Atomic_Memory_Order.hpp
```diff
@@ -51,8 +51,19 @@ struct memory_order_relaxed_t {
...
   static constexpr auto std_constant = std::memory_order_relaxed;
 };
+
+// FIXME_OPENMPTARGET - The `declare target` is needed for the Intel GPUs with
+// the OpenMPTarget backend
+#if defined(KOKKOS_ENABLE_OPENMPTARGET) && defined(KOKKOS_ARCH_INTEL_PVC)
+#pragma omp declare target
+#endif
+
 constexpr memory_order_relaxed_t memory_order_relaxed = {};
```
dalg24, line 61: Did you try declaring it `inline` instead?

crtrott: didn't work on the ALL one, so ...

crtrott (approved):

core/src/impl/Kokkos_Atomic_Memory_Order.hpp
dalg24: sorry did not notice it does not work for `ALL`

dalg24 (approved):
Yuk

core/src/impl/Kokkos_Atomic_Memory_Order.hpp
rgayatri23: @masterleinad Did you try inline? I don't have access to that GPU.

core/unit_test/TestNumericTraits.hpp
```diff
@@ -481,7 +481,13 @@ CHECK_SAME_AS_NUMERIC_LIMITS_MEMBER_FUNCTION(double, round_error);
 CHECK_SAME_AS_NUMERIC_LIMITS_MEMBER_FUNCTION(long double, round_error);
 CHECK_SAME_AS_NUMERIC_LIMITS_MEMBER_FUNCTION(float, denorm_min);
 CHECK_SAME_AS_NUMERIC_LIMITS_MEMBER_FUNCTION(double, denorm_min);
+
+// FIXME_OPENMPTARGET - The static_assert causes issues on Intel GPUs with the
+// OpenMPTarget backend.
+#if !defined(KOKKOS_ENABLE_OPENMPTARGET) && !defined(KOKKOS_ARCH_INTEL_PVC)
```
masterleinad, line 487: I would think this is a general compiler issue than specific to an architecture. I can also see this failing compiling for `XEHP`.

core/src/Kokkos_View.hpp
```diff
@@ -494,8 +494,18 @@ constexpr bool is_assignable(const Kokkos::View<ViewTDst...>& dst,
 
 namespace Kokkos {
 
+// FIXME_OPENMPTARGET - The `declare target` is needed for the Intel GPUs with
+// the OpenMPTarget backend
+#if defined(KOKKOS_ENABLE_OPENMPTARGET) && defined(KOKKOS_ARCH_INTEL_PVC)
+#pragma omp declare target
+#endif
```
masterleinad, line 501: Similarly, here
```suggestion
#if defined(KOKKOS_ENABLE_OPENMPTARGET) && defined(KOKKOS_COMPILER_INTEL)
#pragma omp declare target
#endif
```

rgayatri23: Updated to make it generic to all Intel GPUs with OpenMPTarget backend.

dalg24 (approved):

core/src/Kokkos_View.hpp
```diff
@@ -494,8 +494,18 @@ constexpr bool is_assignable(const Kokkos::View<ViewTDst...>& dst,
 
 namespace Kokkos {
 
+// FIXME_OPENMPTARGET - The `declare target` is needed for the Intel GPUs with
+// the OpenMPTarget backend
+#if defined(KOKKOS_ENABLE_OPENMPTARGET) && defined(KOKKOS_ARCH_INTEL)
```
masterleinad, line 499: That macro doesn't exist
```suggestion
#if defined(KOKKOS_ENABLE_OPENMPTARGET) && defined(KOKKOS_ARCH_INTEL_GPU)
```
but I really think it's more a compiler issue than an architecture issue.

rgayatri23: That should work with `KOKKOS_ENABLE_OPENMPTARGET`

rgayatri23: Or maybe `KOKKOS_ARCH_INTEL_GPU` ?

crtrott: lets do the compiler. But isn't it KOKKOS_COMPILER_INTEL_LLVM or something like that?

crtrott: I thought KOKKOS_COMPILER_INTEL is the classical one?

masterleinad:
```C++
#if defined(__INTEL_COMPILER)
#define KOKKOS_COMPILER_INTEL __INTEL_COMPILER
#elif defined(__INTEL_LLVM_COMPILER)
#define KOKKOS_COMPILER_INTEL __INTEL_LLVM_COMPILER
#elif defined(__ICC)
// Old define
#define KOKKOS_COMPILER_INTEL __ICC
#elif defined(__ECC)
// Very old define
#define KOKKOS_COMPILER_INTEL __ECC
#endif
```
is all we have.

crtrott: oh ok

crtrott: Yeah so that is fine then KOKKOS_COMPILER_INTEL

rgayatri23: Ok edited the code to protect the `declare target` directives with `KOKKOS_ENABLE_OPENMPTARGET` and `KOKKOS_COMPILER_INTEL`

core/unit_test/TestNumericTraits.hpp
```diff
@@ -481,7 +481,13 @@ CHECK_SAME_AS_NUMERIC_LIMITS_MEMBER_FUNCTION(double, round_error);
 CHECK_SAME_AS_NUMERIC_LIMITS_MEMBER_FUNCTION(long double, round_error);
 CHECK_SAME_AS_NUMERIC_LIMITS_MEMBER_FUNCTION(float, denorm_min);
 CHECK_SAME_AS_NUMERIC_LIMITS_MEMBER_FUNCTION(double, denorm_min);
+
+// FIXME_OPENMPTARGET - The static_assert causes issues on Intel GPUs with the
+// OpenMPTarget backend.
+#if !defined(KOKKOS_ENABLE_OPENMPTARGET) && !defined(KOKKOS_COMPILER_INTEL)
 CHECK_SAME_AS_NUMERIC_LIMITS_MEMBER_FUNCTION(long double, denorm_min);
+#endif
```
masterleinad, line 489:
```suggestion
#if !(defined(KOKKOS_ENABLE_OPENMPTARGET) && defined(KOKKOS_COMPILER_INTEL))
CHECK_SAME_AS_NUMERIC_LIMITS_MEMBER_FUNCTION(long double, denorm_min);
#endif
```
Otherwise, this is never compiled for `OpenMPTarget`.

masterleinad (approved):

Issue #5843, opened 2023-02-02
Provide runtime options


kmorel:
It would be helpful to have an API that allows you to query the runtime options that Kokkos was initialized in. This would be helpful for debugging purposes and also to record these options in benchmarking code in a formatting that matches the rest of the logs.

dalg24:
Summarizing what initialization settings we currently support and describing the status quo

| Setting | Note |
| --- | --- |
| num_threads | see below |
| device_id | see below |
| map_device_id_by | useful? |
| disable_warnings | already provided `Kokkos::show_warnings() -> bool` |
| print_configuration | not convinced it would be useful |
| tune_internals | already provided `Kokkos::tune_internals() -> bool` |
| tools_libs | useful? |
| tools_help | not convinced it would be useful |
| tools_args | not convinced it would be useful |

* num_threads
  Currently, one would have to do `Kokkos::DefaultHostExecutionSpace().concurrency()`
* device_id
  No portable solution
  * CUDA: `Kokkos::DefaultExecutionSpace().cuda_device()`
  * HIP: `Kokkos::DefaultExecutionSpace().hip_device()`
  * SYCL: can't get the actual ID but there is `Kokkos::DefaultExecutionSpace().sycl_queue().get_device()`
  * OpenMPTarget: functionality missing at the moment (we should add it)
     In the meantime I suppose could do `omp_get_default_device()`
  * OpenACC: `Kokkos::DefaultExecutionSpace().acc_device_number()`
  * Serial, Threads, OpenMP, HPX: N/A

dalg24:
How does #5853 sound?

dalg24:
> How does #5853 sound? Doc [here](https://github.com/kokkos/kokkos-core-wiki/pull/291/files#diff-5c6293825b06d5bca21aa4e3e0369a0af52c949d1283ddf5bf7c5bb927d90324)

Does not address the issue.  See https://github.com/kokkos/kokkos/pull/5853#discussion_r1095851548

crtrott:
So what you want is a "what initialization settings would I need to provide to reproduce this run with explicit settings?"

kmorel:
That's a fair way of looking at it.

dalg24:
I just remembered we actually have some test where we check our initialization does the right thing
[...]

PR #5853, opened 2023-02-02
Add runtime function to query Kokkos initialization settings

dalg24:
Tentative resolution for #5843 

Adding `Kokkos::initialization_settings()` runtime function that returns the settings including the environment variable parsing.

### DIFF
--- a/core/src/impl/Kokkos_Core.cpp
+++ b/core/src/impl/Kokkos_Core.cpp
@@ -50,6 +50,9 @@ bool g_is_initialized = false;
 bool g_is_finalized   = false;
 bool g_show_warnings  = true;
 bool g_tune_internals = false;
+
+Kokkos::InitializationSettings g_initialization_settings = {};
+
 // When compiling with clang/LLVM and using the GNU (GCC) C++ Standard Library
 // (any recent version between GCC 7.3 and GCC 9.2), std::deque SEGV's during
 // the unwinding of the atexit(3C) handlers at program termination.  However,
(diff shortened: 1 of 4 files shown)


crtrott:
Yeah looks fine.

core/src/impl/Kokkos_Core.cpp
```diff
@@ -767,7 +770,8 @@ void post_initialize_internal(const Kokkos::InitializationSettings& settings) {
   Kokkos::Tools::InitArguments tools_init_arguments;
   combine(tools_init_arguments, settings);
   initialize_profiling(tools_init_arguments);
-  g_is_initialized = true;
+  g_is_initialized          = true;
+  g_initialization_settings = settings;
```
kmorel, line 774: At this point, does `settings` contain information about settings that are implicitly set? For example, if the user doesn't set num threads (for a CPU device), does _settings_ have the number of threads that will actually be used?

dalg24: No.  It contains the settings resolved from the argument passed to `Kokkos::initialize` and the environment variables.

kmorel: Is there a way to fill it with that information? That is something we would like to capture if possible.

dalg24: Can you elaborate on why the version proposed is not sufficient?
If `device_id` is set, you will get that device or Kokkos will call abort if it is not able to.  We will not just decide to pick another one.  Similarly, if `num_threads` is specified, then Kokkos will honor that choice.

kmorel: The point is to get a hint about what kokkos is doing when something is _not_ set. If I don't set `num_threads`, how many threads will Kokkos use? The answer is not obvious. Will it use the number of physical threads, or will it include hyperthreads? Did a job scheduler partition threads in a way that reduces them per process?

Likewise, if I _don't_ set `device_id`, I might want to get an indication of which id Kokkos picked. Did the job scheduler properly allocate different devices for each process on the same node, or did I dork something up with the scheduling flags?

dalg24: OK.  I would explore another direction for this.  I don't like the idea of our returning `InitializationSettings` for that purpose.  Let's take back the discussion to #5843

I will keep that draft open a bit longer to see if there is interest in that facility.

dalg24:
Closing as there does not seem to be much appetite for this facility.  Feel free to comment here if that changes and we can revisit as needed.

PR #5855, opened 2023-02-03
Add runtime functions to query the number of threads and the device ID

dalg24:
Tentative resolution for #5843

Implicit precondition on both that Kokkos is initialized and not finalized.

### DIFF
--- a/core/src/Kokkos_Core.hpp
+++ b/core/src/Kokkos_Core.hpp
@@ -99,6 +99,9 @@ void declare_configuration_metadata(const std::string& category,
 [[nodiscard]] bool is_initialized() noexcept;
 [[nodiscard]] bool is_finalized() noexcept;
 
+[[nodiscard]] int device_id() noexcept;
+[[nodiscard]] int num_threads() noexcept;
+
 bool show_warnings() noexcept;
 bool tune_internals() noexcept;
 
--- a/core/src/impl/Kokkos_Core.cpp
+++ b/core/src/impl/Kokkos_Core.cpp
@@ -165,6 +165,26 @@ bool is_valid_map_device_id_by(std::string const& x) {
 
 }  // namespace
 
+[[nodiscard]] int Kokkos::device_id() noexcept {
+#if defined(KOKKOS_ENABLE_CUDA)
+  return Cuda().cuda_device();
+#elif defined(KOKKOS_ENABLE_HIP)
+  return HIP().hip_device();
+#elif defined(KOKKOS_ENABLE_OPENACC)
+  return Experimental::OpenACC().acc_device_number();
+#elif defined(KOKKOS_ENABLE_OPENMPTARGET)
+  return omp_get_default_device();  // FIXME_OPENMPTARGET
+#elif defined(KOKKOS_ENABLE_SYCL)
+  return Experimental::Impl::SYCLInternal::m_syclDev;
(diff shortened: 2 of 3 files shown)


dalg24:
Retest this please

core/src/Kokkos_Core.hpp
```diff
@@ -99,6 +99,9 @@ void declare_configuration_metadata(const std::string& category,
 [[nodiscard]] bool is_initialized() noexcept;
 [[nodiscard]] bool is_finalized() noexcept;
 
+[[nodiscard]] int device_id() noexcept;
```
fnrizzi, line 102: I was thinking yesterday, if it would make sense to wrap the device id into a class so that one could have a different integer identifier. For example one could have that as far as kokkos is concerned, we have device 0,1,2... but the enumeration is different natively somehow. 
Something like:

```cpp
struct DeviceIdWrapper{
  int kokkos_based_id;
  int native_device_id; 
};

DeviceIdWrapper device_id() noexcept;
```

I don't know if this can be useful always, but it would help us guard against possible differences

dalg24: As you pointed out the native id will depend on the backend, and IMO it belongs to the interoperability facility that the backend is allowed to provide and by definition is not portable.
In the snippet you posted the `native_device_id` would need to be an execution space specific type instead of an `int`.  It might be my lack of imagination but I fail to see how one would do something useful with this information.

PhilMiller: Having the native thing available facilitates interoperability with non-kokkos code that wants to make sure it's using the right resources (either the same ones, or distinct ones)

PhilMiller: We agreed that a native thing would be useful, but is not necessary to address the current need.

fnrizzi:
I will open an issue about the native Id as agreed during our chat @dalg24 @PhilMiller

fnrizzi (approved):
pending the CI passing

dalg24:
Retest this please

dalg24:
Failure is unrelated (` 24 - KokkosContainers_UnitTest_OpenMP (Child killed)` in GCC-8.4.0 build)

crtrott (approved):

core/src/impl/Kokkos_Core.cpp
```diff
@@ -165,6 +165,26 @@ bool is_valid_map_device_id_by(std::string const& x) {
...
 }  // namespace
 
+[[nodiscard]] int Kokkos::device_id() noexcept {
+#if defined(KOKKOS_ENABLE_CUDA)
+  return Cuda().cuda_device();
+#elif defined(KOKKOS_ENABLE_HIP)
+  return HIP().hip_device();
+#elif defined(KOKKOS_ENABLE_OPENACC)
+  return Experimental::OpenACC().acc_device_number();
+#elif defined(KOKKOS_ENABLE_OPENMPTARGET)
```
e10harvey, line 175: I'm guessing OPENMPTARGET is like a device so both Cuda and OpenMP cannot be returned at the same time. No need to check KOKKOS_ENABLE_OPENMP here?

dalg24: OpenMP is a "host" backend so it has no influence on the "device id"

Issue #5858, opened 2023-02-06
Support fill_random for normal distributed random numbers


vqd8a:
Gemma code needs ```fill_random``` for normal distributed random numbers for its Schur-complement PCA preconditioner implementation. And it looks to me Kokkos does not have these yet. I will make a PR for adding these.

ajpowelsnl:
@vqd8a -- what is the status of this issue?

PR #5859, opened 2023-02-06
Add fill_random for normal distributed random numbers

vqd8a:
This PR addresses #5858 for Gemma's need. I am adding unit tests too.


algorithms/src/Kokkos_Random.hpp
```diff
@@ -1518,6 +1531,302 @@ void fill_random(const ExecutionSpace& exec, ViewType a, RandomPool g,
...
+
+template <class ExecutionSpace, class ViewType, class RandomPool,
+          class IndexType = int64_t>
+void fill_random_normal(const ExecutionSpace& exec, ViewType a, RandomPool g,
+                        typename ViewType::const_value_type mean,
+                        typename ViewType::const_value_type std_dev) {
+  int64_t LDA = a.extent(0);
+  if (LDA > 0)
+    parallel_for(
+        "Kokkos::fill_random",
```
fnrizzi, line 1823: the label here is missing "normal"

fnrizzi:
@vqd8a 
(1) why did you specialize `draw_normal` for only double and complex<double> and not any of the others? 

(2) the comment/minidoc at the top was not updated 

(3) beside the above two comments, I personally don't think `rand` should be modified to expose a `draw_normal` because `rand` is a "Template functions to get equidistributed random numbers from a generator for a specific Scalar type" (see top of the file). If we need to add a normal one, i think it should be added as a new class `NormalRand` or better naming that looks someting like:
```
template<class Generator,Scalar>
       struct NormalRand{
         //Returns a value normally distributed with mean = 0 and std = 1
         KOKKOS_INLINE_FUNCTION static Scalar draw(Generator& gen);
         
         //Return value normally distributed with mean = meanVal and std = stdVal
         KOKKOS_INLINE_FUNCTION static Scalar draw(Generator& gen, const Scalar& meanVal, const Scalar& stdVal);
      };
 ```

File: algorithms/src/Kokkos_Random.hpp (hunk 1 above)
Comment (vqd8a) [reply algorithms/src/Kokkos_Random.hpp]: @fnrizzi Thanks for catching this. I added "normal".

**Comment by vqd8a** [comment]
@fnrizzi Thanks for your comments/suggestions.
1) I just saw there were only ```normal()``` with ```double``` available. Moreover, Gemma only needs ```complex<double>```. However, I have added ```fnormal()``` and specializations for ```float``` and ```complex<float>``` too. Hope these are enough.

(2) I updated the comment/minidoc at the top.

(3) This makes sense. So, I made a new class ```normal_rand``` for normally distributed random numbers.

File: algorithms/unit_tests/TestRandom.hpp [hunk 2]
```diff
@@ -486,6 +775,13 @@ void test_random_xorshift64() {
   Impl::TestDynRankView<ExecutionSpace,
                         Kokkos::Random_XorShift64_Pool<ExecutionSpace>>(10000)
       .run();
+#if defined(KOKKOS_ENABLE_SYCL) || defined(KOKKOS_ENABLE_CUDA) || \
+    defined(KOKKOS_ENABLE_HIP)
+  const int N = 134217728;
+#else  // SERIAL, HPX, OPENMP
+  const int N         = 134217728 / 8;
+#endif
```
masterleinad, line 783: How long does this typically take?

algorithms/unit_tests/TestRandom.hpp
```diff
@@ -469,6 +488,276 @@ struct TestDynRankView {
...
+              << ", variance_eps " << variance_eps << ", mean_expect " << mean
+              << ", variance_expect " << stddev * stddev << ", mean "
+              << mean_calc << ", variance " << var_calc << std::endl;
+    if (std::is_same<ScalarA, double>::value) {
+      EXPECT_LE(mean_eps, 1e-2);
+      EXPECT_LE(variance_eps, 1e-2);
+    } else {
+      EXPECT_LE(mean_eps, 1e-1);
+      EXPECT_LE(variance_eps, 1e-1);
+    }
```
masterleinad, line 542: How did you come up with these tolerances? What's the probability for this test to pass given the runtime length?

algorithms/unit_tests/TestRandom.hpp
vqd8a: @masterleinad With these N, the random tests takes 47s on cuda backend, 8s on openmp backend, and 160s on serial backend. I ran the tests on Lassen with V100 GPU and POWER9 CPU.

algorithms/unit_tests/TestRandom.hpp
vqd8a: @masterleinad, When testing on Lassen, this test always pass with the tolerance of 1e-3 on cuda, openmp, serial backends with both float and double. However, it almost fails right away with CI testings on CPUs. So, I chose 1e-2 for double and 1e-1 for float just for the worst test cases pass. Do you have any suggestion?

masterleinad: The tolerances should depend on the sample size, right? The distribution of the sample mean is a normal distribution with the same mean but with a standard deviation scaled by `1./sqrt(n)` and we should use that in the assertion (since we are using different sample sizes on the host and on the device).

masterleinad (approved):
Looks OK to me.

masterleinad:
The OpenMPTarget CI reports
```
28:  -- Testing double Views with Normal distribution 
28: CUDA error: an illegal memory access was encountered 
28: Libomptarget error: Call to targetDataEnd failed, abort target.
28: Libomptarget error: Failed to process data after launching the kernel.
28: Libomptarget error: Consult https://openmp.llvm.org/design/Runtimes.html for debugging options.
28: Kokkos_OpenMPTarget_ParallelFor_Range.hpp:54:1: Libomptarget fatal error 1: failure of target construct while offloading is mandatory
28/34 Test #28: KokkosAlgorithms_UnitTest_RandomAndSort ......Child aborted***Exception:   5.59 sec
test 29
```
I would probably just disable the test for OpenMPTarget with a FIXME.

vqd8a:
@masterleinad Thanks. I have been struggling with it. Do you mean something like this?
```
#ifdef KOKKOS_ENABLE_OPENMPTARGET // FIXME_OPENMPTARGET
  GTEST_SKIP() << "SomeMessage?";
#endif
```

masterleinad:
> @masterleinad Thanks. I have been struggling with it. Do you mean something like this?

Yes, maybe something like
```
 "Skipping only test_random_normal for OpenMPTarget"
```
to clarify that only part of the test is skipped.

algorithms/src/Kokkos_Random.hpp
```diff
@@ -867,6 +954,23 @@ class Random_XorShift64 {
...
+  KOKKOS_INLINE_FUNCTION
+  float fnormal() {
+    double S = 2.0;
+    double U;
+    while (S >= 1.0) {
+      U              = 2.0 * drand() - 1.0;
+      const double V = 2.0 * drand() - 1.0;
+      S              = U * U + V * V;
+    }
+    return static_cast<float>(U * std::sqrt(-2.0 * std::log(S) / S));
```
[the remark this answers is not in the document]
masterleinad: By now, namespace `std` should work with SYCL.

algorithms/unit_tests/TestRandom.hpp
vqd8a: @dalg24 Isn't that 160 seconds on SERIAL backend big enough? I have no problem with just using ```N = 134217728``` for all backends. I am going to change this.

algorithms/src/Kokkos_Random.hpp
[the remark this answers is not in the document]
vqd8a: @dalg24 I am switching to the ones from Kokkos::

algorithms/src/Kokkos_Random.hpp
```diff
@@ -867,6 +954,23 @@ class Random_XorShift64 {
   double normal(const double& mean, const double& std_dev = 1.0) {
     return mean + normal() * std_dev;
   }
+
+  KOKKOS_INLINE_FUNCTION
+  float fnormal() {
+    double S = 2.0;
```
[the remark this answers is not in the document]
vqd8a: @dalg24 I just wanted to keep operations in double precision and only casted the final result to float. I am changing these to ```float``` and see if the accuracy meets.

algorithms/src/Kokkos_Random.hpp
```diff
@@ -1518,6 +1639,301 @@ void fill_random(const ExecutionSpace& exec, ViewType a, RandomPool g,
...
+                             typename ViewType::const_value_type std_dev_)
+      : a(a_), rand_pool(rand_pool_), mean(mean_), std_dev(std_dev_) {}
+
+  KOKKOS_INLINE_FUNCTION
+  void operator()(IndexType i) const {
+    typename RandomPool::generator_type gen = rand_pool.get_state();
+    for (IndexType j = 0; j < loops; j++) {
+      const IndexType idx = i * loops + j;
+      if (idx < static_cast<IndexType>(a.extent(0))) {
+        for (IndexType k = 0; k < static_cast<IndexType>(a.extent(1)); k++)
```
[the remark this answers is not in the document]
vqd8a: @dalg24 I just followed what is being used in the ```fill_random_functor_begin_end```. I can change these to MDRangePolicy. But do you know why ```RangePolicy``` is used below with a loop count of 128? 
https://github.com/kokkos/kokkos/blob/65aa95e8280d1ff845894329a8c3e8b0d138debc/algorithms/src/Kokkos_Random.hpp#L1513-L1518

ajpowelsnl:
Hi @vqd8a - are you still working on this PR?  If so, do you have a timeline for completion?  Many thanks for any update you can provide.

vqd8a:
@ajpowelsnl Thanks for checking. I have not had time to work on this PR for a while. But Gemma still needs this. I planned to complete it by mid November.

masterleinad:
@vqd8a Any news?

vqd8a:
Closing this. I will reopen when I have time to work on.

PR #5865, opened 2023-02-08
Make `HPX::concurrency` a non-static member function

### LINKED ISSUE #5655: Make `ExecutionSpace::concurrency()` a non-static member function

msimberg:
Fixes #5655 for HPX.

### DIFF
--- a/core/src/HPX/Kokkos_HPX.cpp
+++ b/core/src/HPX/Kokkos_HPX.cpp
@@ -153,7 +153,11 @@ void HPX::impl_static_fence(const std::string &name) {
       });
 }
 
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
 int HPX::concurrency() {
+#else
+int HPX::concurrency() const {
+#endif
   hpx::runtime *rt = hpx::get_runtime_ptr();
   if (rt == nullptr) {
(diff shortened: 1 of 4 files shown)


masterleinad (approved):

dalg24:
Ignoring Cuda timing based failure

dalg24 (approved):

PR #5866, opened 2023-02-08
Add cuda ada hopper arch to version 3.7 maintenance branch

akohlmey:
These are the changes make to the Kokkos library copy bundled with LAMMPS to support compiling/running with the CUDA Ada (sm_89)  and Hopper (sm_90) architectures.


dalg24 (changes requested):
You must target the develop branch.  Once merged we would then consider whether to cherry-pick or not.

dalg24:
(You should be able to edit and change the destination branch without closing)

akohlmey:
> (You should be able to edit and change the destination branch without closing)

I have to branch off of develop instead of the 3.7 maintenance and that doesn't need the Hopper related bits.

dalg24:
You could have push forced once you did that.  Anyway no big deal.

PR #5867, opened 2023-02-08
Add CUDA Ada architecture support

akohlmey:
This change adds support to compile Kokkos for Ada generation (sm_89) consumer GPUs (RTX40x0)


cmake/kokkos_arch.cmake
```diff
@@ -866,7 +868,7 @@ IF (KOKKOS_ARCH_VOLTA70 OR KOKKOS_ARCH_VOLTA72)
   SET(KOKKOS_ARCH_VOLTA ON)
 ENDIF()
 
-IF (KOKKOS_ARCH_AMPERE80 OR KOKKOS_ARCH_AMPERE86)
+IF (KOKKOS_ARCH_AMPERE80 OR KOKKOS_ARCH_AMPERE86 OR KOKKOS_ARCH_ADA89)
```
dalg24, line 871: That does not look right.
We do not define `KOKKOS_ENABLE_VOLTA` when using Turing75 and I fail to see why that is different.

dalg24: Ping

akohlmey: You may be correct in principle, but my line of thinking is that there is no likely a need for specific optimizations for those consumer grade GPUs and go through the extra maintenance effort. Then grouping this with the closest feature set seems better than not grouping it into any of those at all. On the Kokkos slack it was confirmed by some Nvidia person that this assignment would be acceptable (I forgot who).

Bottom line, as far as I am concerned, you can take this, drop it, or modify it as you see fit. I personally don't see a need for a change and will not follow up any further. I've primarily submitted it since I consider it best practice to submit changes that we include in bundled software to the upstream version.

masterleinad (commented):
We should also update `pre_initialize_internal()` in `core/src/impl/Kokkos_Core.cpp`.

cmake/kokkos_arch.cmake
```diff
@@ -868,7 +868,7 @@ IF (KOKKOS_ARCH_VOLTA70 OR KOKKOS_ARCH_VOLTA72)
   SET(KOKKOS_ARCH_VOLTA ON)
 ENDIF()
 
-IF (KOKKOS_ARCH_AMPERE80 OR KOKKOS_ARCH_AMPERE86 OR KOKKOS_ARCH_ADA89)
```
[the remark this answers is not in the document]
dalg24: That's a good question.  It likely needs to be added in the first location you mention, although I expect it would compile w/o it.
For all the others in ACC, OMP, and SYCL, we discussed yesterday adding a `KOKKOS_ARCH_NVIDIA_GPU` macro because that is really what was needed there.

crtrott:
Lets take this PR over also in light of the other one introducing general CUDA arch gpu thingy.

Rombur (approved):

janciesko (approved):

Issue #5717, opened 2022-12-21
rank() function for View and DynRankView


maartenarnst:
In Kokkos_View.hpp, there is around lines 644-648 a function `rank()` that is commented out and said "to be implemented".  This function mimics a function also called `rank()` from Kokkos_DynRankView.hpp.  It seems the idea of having the same function in both classes is to allow client code to work generically for `View` and `DynRankView`.  However, because the lines are commented out, this currently doesn't work.

If the lines in  Kokkos_View.hpp are uncommented, compilation errors occur, because, around line 345 in Kokkos_View.hpp, there is an `enum { rank = dimension::rank };`.  Thus, the compiler gets confused by the enum and the function having the same name.

Is having such a function `rank()` still under consideration?  And would there be a solution to the conflict with the enum?  Renaming the function `rank()` might be a solution, but it would likely require many changes in client code.

An example of a use case of such a function could be in Intrepid2.  In fact, Intrepid2 defines functors that are templated generally on a "viewType".  Ideally, these functors could then be instantiated for both View and DynRankView, which is often used in Intrepid2.  But, because some of the member functions of such functors in Intrepid2 call `rank()`, only instantiation for DynRankView is currently possible.  E.g. line 97 in `intrepid2/src/Cell/Intrepid2_CellToolsDefJacobian.hpp`, but there are several other instances where it would be beneficial to have `rank()` work for both `View` and `DynRankView`.

dalg24:
For reference https://github.com/kokkos/kokkos/blob/d9d43bf62998854617e9224b074775aa1e6fa1ba/core/src/Kokkos_View.hpp#L644-L648

dalg24:
> Is having such a function rank() still under consideration? 

Not actively, no.
We will bring it up at the developer meeting.  In the meantime feel free to upvote this if there is more interest in this feature.

Note that can be implemented on the user side as
```C++
template <class ViewLike>
auto rank(ViewLike const& v) {
  if constexpr (Kokkos::is_view_v<ViewLike>)
    return ViewLike::rank;
  else if constexpr (Kokkos::is_dyn_rank_view<ViewLike>)
    return v.rank();
  else
    static_assert(std::is_void_v<ViewLike>);
}

crtrott:
Yeah we probably should fix this, unfortunately we did not take that into consideration during deprecation in 3.x :-(

crtrott:
uhoh I have a nasty thought - not sure if this would fully work lol

```c++
[...]

PR #5870, opened 2023-02-09
Add support for `View::rank[_dynamic]()`

dalg24:
Related to #5717
* Make `View::rank[_dynamic]` implicitly convertible to ~~`int`~~ `size_t` and callable
* ~~Deprecate `View::Rank` (never mentioned in the API reference for View)~~ done in #5882

~~Not sure it is worth the trouble.  Opening so we can discuss.~~ We agreed to move forward at the dev meeting.

`View` member | Before this PR | After this PR
--- | --- | ---
`View::Rank` | implementation-defined integral type | deprecated
`View::rank`  | implementation-defined integral type | unspecified type that is convertible to `size_t`
`View::rank()` | not provided | returns `size_t`
`View::rank_dynamic` | implementation-defined integral type | unspecified type that is convertible to `size_t`
`View::rank_dynamic()` | not provided | returns `size_t`

Two things worth noting:
* One advantage of the `View::rank()` member over the `rank(View)` free function is that it can be used in constant expressions.
* The actual type of `View::rank` as it was defined before was left up to the implementation (up to the compiler not to Kokkos) but in practice it was often `int` which means this change may yield warnings about comparing signed and unsigned integral types.  It may also break code that was using the type of `View::rank`.  The rational for choosing `size_t` was to align with `std::mdspan`.

### DIFF
--- a/algorithms/src/std_algorithms/impl/Kokkos_Constraints.hpp
+++ b/algorithms/src/std_algorithms/impl/Kokkos_Constraints.hpp
@@ -29,7 +29,7 @@ struct is_admissible_to_kokkos_std_algorithms : std::false_type {};
 
 template <typename T>
 struct is_admissible_to_kokkos_std_algorithms<
-    T, std::enable_if_t< ::Kokkos::is_view<T>::value && T::rank == 1 &&
+    T, std::enable_if_t< ::Kokkos::is_view<T>::value && T::rank() == 1 &&
                          (std::is_same<typename T::traits::array_layout,
                                        Kokkos::LayoutLeft>::value ||
                           std::is_same<typename T::traits::array_layout,
--- a/core/src/Kokkos_CopyViews.hpp
+++ b/core/src/Kokkos_CopyViews.hpp
@@ -1636,19 +1636,19 @@ inline void deep_copy(
           "match: ");
       message += dst.label();
       message += "(";
-      for (int r = 0; r < dst_type::rank - 1; r++) {
-        message += std::to_string(dst.extent(r));
+      message += std::to_string(dst.extent(0));
+      for (size_t r = 1; r < dst_type::rank; r++) {
         message += ",";
+        message += std::to_string(dst.extent(r));
       }
-      message += std::to_string(dst.extent(dst_type::rank - 1));
       message += ") ";
       message += src.label();
       message += "(";
-      for (int r = 0; r < src_type::rank - 1; r++) {
-        message += std::to_string(src.extent(r));
+      message += std::to_string(src.extent(0));
+      for (size_t r = 1; r < src_type::rank; r++) {
         message += ",";
+        message += std::to_string(src.extent(r));
       }
-      message += std::to_string(src.extent(src_type::rank - 1));
       message += ") ";
 
       Kokkos::Impl::throw_runtime_exception(message);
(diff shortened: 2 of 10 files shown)


ldh4:
I agree with this change, keeping it consistent with ViewTraits, even though there seem to be handful of comparison of different integer signedness warnings to be handled.

We also happened to have called the struct that contains the iteration pattern `Kokkos::Rank`. So, renaming `View::Rank` to `View::rank` could help reducing potential source of that confusion, especially when used without fully qualified names.

PhilMiller:
Down the line, whatever we do with these interfaces ought to support code that's generic over both `View` and `DynRankView` (and maybe indicate `DynamicView` as rank 1, because that's how it would be accessed and copied)

nmm0:
I added to the MDspan project since it's related (i.e. rank returning `size_t`)

core/src/impl/Kokkos_Utilities.hpp
```diff
@@ -29,6 +29,23 @@
...
 namespace Impl {
 
+// same as std::integral_constant but with __host__ __device__ annotations on
+// the implicit conversion function and the call operator
+template <class T, T v>
+struct integral_constant {
+  using value_type         = T;
+  using type               = integral_constant<T, v>;
+  static constexpr T value = v;
+  KOKKOS_FUNCTION constexpr operator value_type() const noexcept {
```
nmm0, line 39: If we're re-implementing it anyway because of the need for `__host__ __device__` we could eventually deprecate this conversion if we wanted to make the view API more similar to mdspan.

That's assuming this utility class is not used elsewhere...

dalg24: You read my mind

dalg24: But not doing that for a while.  We need to get users off of View::Rank first and we want to make it reasonably easy for them to support a range of Kokkos versions.

core/src/Kokkos_CopyViews.hpp
```diff
@@ -1636,19 +1636,19 @@ inline void deep_copy(
           "match: ");
       message += dst.label();
       message += "(";
-      for (int r = 0; r < dst_type::rank - 1; r++) {
-        message += std::to_string(dst.extent(r));
+      message += std::to_string(dst.extent(0));
+      for (size_t r = 1; r < dst_type::rank; r++) {
```
dalg24, line 1640: These changes are to avoid warning pointless comparison of unsigned integer with zero when the loop is from 0 to rank-1

core/src/Kokkos_View.hpp
```diff
@@ -814,14 +814,14 @@ class View : public ViewTraits<DataType, Properties...> {
 
   template <typename... Is>
   static KOKKOS_FUNCTION void check_access_member_function_valid_args(Is...) {
-    static_assert(traits::rank <= sizeof...(Is), "");
+    static_assert(rank <= sizeof...(Is), "");
```
dalg24, line 817: `traits::rank[_dynamic_rank]` are still valid but it is cleaner without qualifying.

algorithms/src/std_algorithms/impl/Kokkos_Constraints.hpp
```diff
@@ -29,7 +29,7 @@ struct is_admissible_to_kokkos_std_algorithms : std::false_type {};
 
 template <typename T>
 struct is_admissible_to_kokkos_std_algorithms<
-    T, std::enable_if_t< ::Kokkos::is_view<T>::value && T::rank == 1 &&
+    T, std::enable_if_t< ::Kokkos::is_view<T>::value && T::rank() == 1 &&
```
dalg24, line 32: MSVC was choking on that guy.  Was giving failed static assertions all over the place in algorithms

core/src/Kokkos_View.hpp
```diff
@@ -1716,7 +1705,7 @@ struct RankDataType<ValueType, 0> {
 
 template <unsigned N, typename... Args>
 KOKKOS_FUNCTION std::enable_if_t<
-    N == View<Args...>::rank &&
+    N == View<Args...>::rank() &&
```
dalg24, line 1708: This change and the one below was for MSVC :(

dalg24:
Retest this please

dalg24:
Retest this please

crtrott (approved):

masterleinad (approved):
Looks OK to me.

dalg24:
Ignoring HIP failure

PR #5871, opened 2023-02-09
Fix cache configuration in CI

### LINKED ISSUE #5723: `ccache` is not configured properly in CI
**Describe the bug**
We try to use `ccache` to speed up builds in Kokkos CI, but the right option (`CMAKE_CXX_COMPILER_LAUNCHER`) is not provided at the configuration stage.
https://github.com/kokkos/kokkos/blob/f54c04c77250be0cf82f349d451859f4854e4d9e/.github/workflows/continuous-integration-workflow.yml#L109-L113
Statistics from [an actual build](https://github.com/kokkos/kokkos/actions/runs/3762406268/jobs/6395074913):
```
Summary:
  Hits:               0 /  432 (0.00 %)
    Direct:           0 /  432 (0.00 %)
    Preprocessed:     0 /  432 (0.00 %)
  Misses:           432
    Direct:         432
    Preprocessed:   432
  Uncacheable:      146
Primary storage:
  Hits:               0 /  864 (0.00 %)
  Misses:           864
  Cache size (GB): 0.11 / 0.25 (45.65 %)
```

cz4rs:
fixes #5723

---
During `Run actions/cache@v3` step we currently get:
```
Cache not found for input keys: kokkos-ubuntu:latest-g++-Debug--${github.ref}-2272d3b7f6e1027da5d0fd37c67f4063f87daf4e, kokkos-ubuntu:latest-g++-Debug--refs/heads/develop
```
Change `${github.ref}` to `${{ github.ref }}` to access the variable correctly.

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -70,8 +70,8 @@ jobs:
       - uses: actions/cache@v3
         with:
           path: ~/.ccache
-          key: kokkos-${{ matrix.distro }}-${{ matrix.cxx }}-${{ matrix.cmake_build_type }}-${{ matrix.openmp }}-${github.ref}-${{ github.sha }}
-          restore-keys: kokkos-${{ matrix.distro }}-${{ matrix.cxx }}-${{ matrix.cmake_build_type }}-${{ matrix.openmp }}-${{github.ref}}
+          key: kokkos-${{ matrix.distro }}-${{ matrix.cxx }}-${{ matrix.cmake_build_type }}-${{ matrix.openmp }}-${{ github.ref }}-${{ github.sha }}
+          restore-keys: kokkos-${{ matrix.distro }}-${{ matrix.cxx }}-${{ matrix.cmake_build_type }}-${{ matrix.openmp }}-${{ github.ref }}
       - name: maybe_disable_death_tests
         if: ${{ matrix.distro == 'fedora:rawhide' }}
         run: echo "GTEST_FILTER=-*DeathTest*" >> $GITHUB_ENV
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -69,7 +69,7 @@ jobs:
         uses: actions/checkout@v3
       - uses: actions/cache@v3
         with:
-          path: ~/.ccache
+          path: ~/.cache/ccache
           key: kokkos-${{ matrix.distro }}-${{ matrix.cxx }}-${{ matrix.cmake_build_type }}-${{ matrix.openmp }}-${{ github.ref }}-${{ github.sha }}
           restore-keys: kokkos-${{ matrix.distro }}-${{ matrix.cxx }}-${{ matrix.cmake_build_type }}-${{ matrix.openmp }}-${{ github.ref }}
       - name: maybe_disable_death_tests
--- a/.github/workflows/performance-benchmark.yml
+++ b/.github/workflows/performance-benchmark.yml
@@ -24,8 +24,8 @@ jobs:
       - uses: actions/cache@v3
         with:
           path: ~/.ccache
-          key: kokkos-${{ matrix.distro }}-${{ matrix.cxx }}-${{ matrix.openmp }}-${github.ref}-${{ github.sha }}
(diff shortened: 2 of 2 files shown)


cz4rs:
FIXME: 
`Post Run actions/cache@v3` still gives us:
```
Warning: Path Validation Error: Path(s) specified in the action for caching do(es) not exist, hence no cache is being saved.
```

cz4rs:
> FIXME: `Post Run actions/cache@v3` still gives us:
> 
> ```
> Warning: Path Validation Error: Path(s) specified in the action for caching do(es) not exist, hence no cache is being saved.
> ```
> 
> Path mentioned is `~/.ccache`.

After fixing the path:
```
Cache Size: ~52 MB (54562993 B)
Cache saved successfully
Cache saved with key: kokkos-ubuntu:latest-g++-Release--refs/pull/5871/merge-509cf31f25f29e69ddb7ca91db97bd2706ebe6f0
```

.github/workflows/continuous-integration-workflow.yml
```diff
@@ -69,9 +69,9 @@ jobs:
         uses: actions/checkout@v3
       - uses: actions/cache@v3
         with:
-          path: ~/.ccache
-          key: kokkos-${{ matrix.distro }}-${{ matrix.cxx }}-${{ matrix.cmake_build_type }}-${{ matrix.openmp }}-${github.ref}-${{ github.sha }}
-          restore-keys: kokkos-${{ matrix.distro }}-${{ matrix.cxx }}-${{ matrix.cmake_build_type }}-${{ matrix.openmp }}-${{github.ref}}
+          path: ~/.cache/ccache
```
dalg24, line 72: Do you know why the path is not the usual `${HOME}/.ccache`?
How is it set to something else?  I can't find a `CCACHE_DIR` environment variable.

cz4rs: https://ccache.dev/manual/latest.html#_location_of_the_configuration_file

cz4rs: Ubuntu image uses `ccache` version `4.5.1`. From https://ccache.dev/manual/4.5.html#_configuration_options:

> **cache_dir (CCACHE_DIR)**
> 
> This option specifies where ccache will keep its cached compiler outputs. The default is $XDG_CACHE_HOME/ccache if XDG_CACHE_HOME is set, **otherwise $HOME/.cache/ccache**. Exception: If the legacy directory $HOME/.ccache exists then that directory is the default.
> 

We do create the "legacy directory" in Ubuntu Dockerfile:
```Dockerfile
WORKDIR /home/kokkos
RUN mkdir .ccache
```
but when I ran `ccache` in verbose mode, on GitHub we have:
```
Summary:
  Cache directory:          /github/home/.cache/ccache
  Primary config:           /github/home/.config/ccache/ccache.conf
  Secondary config:         /etc/ccache.conf
```

Seems like because `/home/kokkos` is not actually `$HOME`, we fall into `$HOME/.cache/ccache` category.

masterleinad: I think we should either set `XDG_CACHE_HOME` in the Dockerfile then and use it here or don't create `.ccache` in the Dockerfile.

cz4rs:
Random observation: the main CI workflow is triggered on both `push` and `pull_request` unconditionally:
https://github.com/kokkos/kokkos/blob/d39885a2f432af7fa7f8c720068795d5f56b4d37/.github/workflows/continuous-integration-workflow.yml#L1-L2
as opposed to limiting `push` to `develop` branch:
https://github.com/kokkos/kokkos/blob/d39885a2f432af7fa7f8c720068795d5f56b4d37/.github/workflows/performance-benchmark.yml#L1-L6

As far as I know, this means running all the jobs in Kokkos repository (`pull_request`) **and** in the fork (`push`). This seems like wasting a lot of CPU cycles for nothing, since we end up only ever checking the jobs in PR.

.github/workflows/continuous-integration-workflow.yml
cz4rs: > I think we should either set `XDG_CACHE_HOME` in the Dockerfile then and use it here or don't create `.ccache` in the Dockerfile.

I agree, I'd opt for the latter to keep things simple. 
https://github.com/kokkos/ci-containers/pull/34 posted.

masterleinad (approved):

junghans (approved):

dalg24 (approved):

PR #5872, opened 2023-02-09
[4.0.00] Fix version macros

masterleinad:
As noticed by @lucbv, `cmakedefine` variables are undefined if the `CMake` variable is 0.


lucbv (approved):
This update to our CMake yields the proper behavior and since these variables are not set by users but rather hard coded in CMakeLists.txt using `#define` instead of `#cmakedefine` makes sense

crtrott (approved):

PR #5424, opened 2022-09-02
Define `KOKKOS_VERSION_{MAJOR,MINOR,PATCH}` and `KOKKOS_VERSION_{LESS,GREATER,EQUAL}` macros

dalg24:
Add
* `KOKKOS_VERSION_MAJOR`
* `KOKKOS_VERSION_MINOR`
* `KOKKOS_VERSION_PATCH`
macros and draft ~~`KOKKOS_VERSION_COMPARE` so users can write things like~~
```C++
#if KOKKOS_VERSION_COMPARE(>=, 3, 7, 0)
```

**edit per review** `KOKKOS_VERSION_{LESS,GREATER,EQUAL,LESS_EQUAL,GREATER_EQUAL}`
```C++
#if KOKKOS_VERSION_GREATER_EQUAL(3, 7, 0)

PR #5873, opened 2023-02-09
Fix version macros

masterleinad:
As noticed by @lucbv, `cmakedefine` variables are undefined if the `CMake` variable is 0.


dalg24 (approved):
How didn't we notice before...

masterleinad:
> @masterleinad may I add a unit test and push directly to your branch?

Sure, go ahead!

lucbv (approved):
Thanks for the update, the unit-test is also a good enhancement : )

nmm0 (approved):
Looks good to me, we should never have used cmakedefine for this

nmm0:
4.0.0 release branch PR https://github.com/kokkos/kokkos/pull/5872

dalg24:
> Looks good to me, we should never have used cmakedefine for this

You can blame me for that.  Thanks Luc and Daniel for fixing.

dalg24:
For reference this is a fixup for #5424

PR #5874, opened 2023-02-10
Use CombinedFunctorReducerType in ParallelReduce

masterleinad:
Part of #5332. Realizing that converting all backends in one pull request might not be ideal, this pull request introduces a temporary wrapper class ParallelReduceWrapper that allows backends to have either the new or the old interface. This should make discussions easier. So far, this pull request only changes the `Serial` and `Cuda` backend to the new proposed interface.

A general description (largely copied from https://github.com/kokkos/kokkos/pull/5332#issuecomment-1284622238):

The pull request is best understood looking at the changes to Kokkos_Core_fwd.hpp and then Kokkos_Parallel_Reduce.hpp:

The internal classes change from
```
template <class FunctorType, class ExecPolicy, class ReducerType, class ExecutionSpace>
class ParallelReduce;
```
to 
```
template <class CombinedFunctorReducerType, class ExecPolicy, class ExecutionSpace>
class ParallelReduce;
```

meaning that we pass a `CombinedFunctorReducer` instead of a `Functor` and a `Reducer` separately. That class is currently also defined in `Kokkos_Parallel_Reduce.hpp`.
The wrapped functor is the usual functor and the reducer is the reducer wrapped in a `FunctorAnalysis` object (that contains a copy of the reducer). In case, the functor acts as a reducer, the encapsulation gives us the opportunity to avoid having two copies of the functor.

- In `Kokkos_Parallel_Reduce.hpp`, we basically only create the appropriate `CombinedFunctorReducer` object and call the backend implementations.
- The backend implementations only store a copy of the `CombinedFunctorReducer` object (instead of a functor and a reducer separately) that is used whenever access to the functor or reducer is required. Again, the value type comes from the wrapped `FunctorAnalysis` object.
- These changes make it relatively easy to avoid deducing the value type for reductions since we only need to touch the `FunctorAnalysis` class and `Kokkos_Parallel_Reduce.hpp`. This isn't part of the current pull request, though.
- In general, this avoids a lot of code duplication and also removes one of the constructors. It's conceivable for `CombinedFunctorReducer` to also have a call operator that would take care of calling the functor with the correct work tag which would remove more repeated code.
- A caveat with the present implementation is that `FunctorAnalysis` needs to store a copy of the functor which means that there are extra copies in all implementations that don't use the new interface, yet.

### DIFF
--- a/core/src/Cuda/Kokkos_Cuda_Parallel_MDRange.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_Parallel_MDRange.hpp
@@ -260,7 +260,7 @@ class ParallelReduce<FunctorType, Kokkos::MDRangePolicy<Traits...>, ReducerType,
 
   inline __device__ void operator()() const {
     typename Analysis::Reducer final_reducer(
-        &ReducerConditional::select(m_functor, m_reducer));
+        ReducerConditional::select(m_functor, m_reducer));
     const integral_nonzero_constant<size_type, Analysis::StaticValueSize /
                                                    sizeof(size_type)>
         word_count(Analysis::value_size(
@@ -338,7 +338,7 @@ class ParallelReduce<FunctorType, Kokkos::MDRangePolicy<Traits...>, ReducerType,
 
   inline void execute() {
     typename Analysis::Reducer final_reducer(
-        &ReducerConditional::select(m_functor, m_reducer));
+        ReducerConditional::select(m_functor, m_reducer));
 
     const auto nwork = m_policy.m_num_tiles;
     if (nwork) {
--- a/core/src/Cuda/Kokkos_Cuda_Parallel_MDRange.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_Parallel_MDRange.hpp
@@ -188,11 +188,13 @@ class ParallelFor<FunctorType, Kokkos::MDRangePolicy<Traits...>, Kokkos::Cuda> {
       : m_functor(arg_functor), m_rp(arg_policy) {}
 };
 
-template <class FunctorType, class ReducerType, class... Traits>
-class ParallelReduce<FunctorType, Kokkos::MDRangePolicy<Traits...>, ReducerType,
-                     Kokkos::Cuda> {
+template <class CombinedFunctorReducerType, class... Traits>
+class ParallelReduce<CombinedFunctorReducerType,
+                     Kokkos::MDRangePolicy<Traits...>, Kokkos::Cuda> {
  public:
-  using Policy = Kokkos::MDRangePolicy<Traits...>;
+  using Policy      = Kokkos::MDRangePolicy<Traits...>;
+  using FunctorType = typename CombinedFunctorReducerType::functor_type;
+  using ReducerType = typename CombinedFunctorReducerType::reducer_type;
 
  private:
   using array_index_type = typename Policy::array_index_type;
@@ -202,22 +204,10 @@ class ParallelReduce<FunctorType, Kokkos::MDRangePolicy<Traits...>, ReducerType,
   using Member       = typename Policy::member_type;
   using LaunchBounds = typename Policy::launch_bounds;
 
-  using ReducerConditional =
-      Kokkos::Impl::if_c<std::is_same<InvalidType, ReducerType>::value,
-                         FunctorType, ReducerType>;
-  using ReducerTypeFwd = typename ReducerConditional::type;
-  using WorkTagFwd =
-      typename Kokkos::Impl::if_c<std::is_same<InvalidType, ReducerType>::value,
-                                  WorkTag, void>::type;
-
-  using Analysis =
-      Kokkos::Impl::FunctorAnalysis<FunctorPatternInterface::REDUCE, Policy,
-                                    ReducerTypeFwd>;
-
  public:
-  using pointer_type   = typename Analysis::pointer_type;
-  using value_type     = typename Analysis::value_type;
-  using reference_type = typename Analysis::reference_type;
+  using pointer_type   = typename ReducerType::pointer_type;
+  using value_type     = typename ReducerType::value_type;
+  using reference_type = typename ReducerType::reference_type;
   using functor_type   = FunctorType;
   using size_type      = Cuda::size_type;
   using reducer_type   = ReducerType;
@@ -225,9 +215,8 @@ class ParallelReduce<FunctorType, Kokkos::MDRangePolicy<Traits...>, ReducerType,
   // Algorithmic constraints: blockSize is a power of two AND blockDim.y ==
   // blockDim.z == 1
 
-  const FunctorType m_functor;
+  const CombinedFunctorReducerType m_functor_reducer;
   const Policy m_policy;  // used for workrange and nwork
-  const ReducerType m_reducer;
   const pointer_type m_result_ptr;
   const bool m_result_ptr_device_accessible;
   size_type* m_scratch_space;
@@ -241,7 +230,7 @@ class ParallelReduce<FunctorType, Kokkos::MDRangePolicy<Traits...>, ReducerType,
   // Shall we use the shfl based reduction or not (only use it for static sized
   // types of more than 128bit
   static constexpr bool UseShflReduction = false;
-  //((sizeof(value_type)>2*sizeof(double)) && Analysis::StaticValueSize)
+  //((sizeof(value_type)>2*sizeof(double)) && ReducerType::static_value_size())
   // Some crutch to do function overloading
 
  public:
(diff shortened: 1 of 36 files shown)


masterleinad:
One of the HIP CI builds failing looks spurious.

core/src/Kokkos_GraphNode.hpp
```diff
@@ -376,15 +376,38 @@ class GraphNodeRef {
...
+          return_value_adapter::return_value(return_value, functor)});
+    } else {
+      auto reducer = return_value_adapter::return_value(return_value, functor);
+      combined_functor_reducer_type functor_reducer(
+          functor, typename analysis::Reducer(reducer));
+
+      return this->_then_kernel(next_kernel_t{
+          std::move(arg_name), graph_impl_ptr->get_execution_space(),
+          functor_reducer, (Policy &&) policy, reducer.view()});
+    }
```
dalg24, line 410: Can you comment about that change (in GH for now)?  Did you think much about how to simplify that code in the future?

core/src/Kokkos_Parallel_Reduce.hpp
```diff
@@ -1372,6 +1372,115 @@ StdPartitionPoint(View<StdPartPointScalar<Index>, Properties...> const&)
 namespace Kokkos {
 namespace Impl {
 
+template <typename FunctorType, typename FunctorAnalysisReducerType,
```
dalg24, line 1375: Is the "FunctorAnalysis" prefix important here?

core/src/Kokkos_Parallel_Reduce.hpp
```diff
@@ -1372,6 +1372,115 @@ StdPartitionPoint(View<StdPartPointScalar<Index>, Properties...> const&)
...
+
+  void execute() { m_parallel_reduce.execute(); }
+};
+
+template <typename CombinedFunctorReducerType, typename PolicyType,
+          typename ExecutionSpaceType>
+class ParallelReduceWrapper<
+    CombinedFunctorReducerType, PolicyType, ExecutionSpaceType,
+    std::enable_if_t<
+        implements_new_reduce_interface<ExecutionSpaceType>::value>> {
```
dalg24, line 1466: Did you consider inheriting from the "wrapped type" instead?

masterleinad:
@dalg24 Since you are commenting on details, can I assume that you are OK with the overall design?

core/src/OpenMPTarget/Kokkos_OpenMPTarget_ParallelReduce_Range.hpp
```diff
@@ -100,7 +100,7 @@ class ParallelReduce<FunctorType, Kokkos::RangePolicy<Traits...>, ReducerType,
   }
 
   template <class ViewType>
-  ParallelReduce(const FunctorType& arg_functor, Policy& arg_policy,
+  ParallelReduce(const FunctorType& arg_functor, const Policy& arg_policy,
```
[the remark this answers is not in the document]
dalg24: (Not blocking though)

core/src/Kokkos_GraphNode.hpp
masterleinad: That's about the same change as we have in `Parallel_Reduce.hpp`. I'll try to create a `make_combined_functor_reducer_type` function that would avoid the if-clause and could be used here and in `Parallel_Reduce.hpp`

masterleinad:
Implementations:
- Serial: here
- Cuda: here
- Threads: https://github.com/kokkos/kokkos/pull/5894
- OpenMP: https://github.com/kokkos/kokkos/pull/5893
- HPX: https://github.com/kokkos/kokkos/pull/5895
- HIP:
- SYCL:
- OpenMPTarget:
- OpenACC:

crtrott:
Looking into launch latency right now. Also getting this warning a lot:
```
/ascldap/users/crtrott/Kokkos/kokkos/core/src/Kokkos_Parallel_Reduce.hpp(1630): warning: missing return statement at end of non-void function "lambda []()->CombinedFunctorReducerType"
```
in a CUDA build.

crtrott:
Launch latency looks good.

core/src/Cuda/Kokkos_Cuda_Parallel_MDRange.hpp
```diff
@@ -253,24 +242,22 @@ class ParallelReduce<FunctorType, Kokkos::MDRangePolicy<Traits...>, ReducerType,
   inline __device__ void exec_range(reference_type update) const {
     Kokkos::Impl::Reduce::DeviceIterateTile<Policy::rank, Policy, FunctorType,
                                             typename Policy::work_tag,
-                                            reference_type>(m_policy, m_functor,
-                                                            update)
+                                            reference_type>(
+        m_policy, m_functor_reducer.get_functor(), update)
```
crtrott, line 246: we are storing the functor now twice aren't we? I guess we should store the functor_reducer inside the Iterate class thingy?

masterleinad:
> Looking into launch latency right now. Also getting this warning a lot:
/ascldap/users/crtrott/Kokkos/kokkos/core/src/Kokkos_Parallel_Reduce.hpp(1630): warning: missing return statement at end of non-void function "lambda []()->CombinedFunctorReducerType"
in a CUDA build.

Ah, dropping https://github.com/kokkos/kokkos/pull/5874/commits/ef1be2e5d05d6975dbcab84f6901d94dc07d1fc3 should avoid this.

crtrott (changes requested):

core/src/Cuda/Kokkos_Cuda_Parallel_MDRange.hpp
masterleinad: `DeviceIterateTile` stores a reference, not a copy.

masterleinad: @crtrott Do you agree or did I miss something?

masterleinad: For the `Serial` implementation we use `HostIterateTile` which stores a copy indeed but in that implementation, we don't have a `m_functor_reducer` member but only a `m_reducer` and a `m_iter` member (the latter one stores the functor).

masterleinad: Storing `CombinedFunctorReducer` would be quite disruptive since we are using the same class in `ParallelFor`. Hence, I would prefer just keeping the implementation as is. I would expect that stopping an extra copy of the functor if no reducer is given doesn't make much of a difference on the host anyway.

masterleinad: I would be happier to do this if/when we provide `CombinedFunctorReducer` with a call operator.

crtrott (approved):

core/src/Cuda/Kokkos_Cuda_Parallel_MDRange.hpp
masterleinad: We can do
```C++
diff --git a/core/src/Serial/Kokkos_Serial_Parallel_MDRange.hpp b/core/src/Serial/Kokkos_Serial_Parallel_MDRange.hpp
index 0d7143706..c97880483 100644
--- a/core/src/Serial/Kokkos_Serial_Parallel_MDRange.hpp
+++ b/core/src/Serial/Kokkos_Serial_Parallel_MDRange.hpp
@@ -73,9 +73,8 @@ class ParallelReduce<CombinedFunctorReducerType,
   using value_type     = typename ReducerType::value_type;
   using reference_type = typename ReducerType::reference_type;
 
-  using iterate_type =
-      typename Kokkos::Impl::HostIterateTile<MDRangePolicy, FunctorType,
-                                             WorkTag, reference_type>;
+  using iterate_type = typename Kokkos::Impl::HostIterateTile<
+      MDRangePolicy, CombinedFunctorReducerType, WorkTag, reference_type>;
   const iterate_type m_iter;
   const ReducerType m_reducer;
@@ -129,7 +128,7 @@ class ParallelReduce<CombinedFunctorReducerType,
   ParallelReduce(const CombinedFunctorReducerType& arg_functor_reducer,
                  const MDRangePolicy& arg_policy,
                  const ViewType& arg_result_view)
-      : m_iter(arg_policy, arg_functor_reducer.get_functor()),
+      : m_iter(arg_policy, arg_functor_reducer),
         m_reducer(arg_functor_reducer.get_reducer()),
         m_result_ptr(arg_result_view.data()) {
     static_assert(Kokkos::is_view<ViewType>::value,
diff --git a/core/src/impl/KokkosExp_Host_IterateTile.hpp b/core/src/impl/KokkosExp_Host_IterateTile.hpp
index 82604a24c..a44ffefa6 100644
--- a/core/src/impl/KokkosExp_Host_IterateTile.hpp
+++ b/core/src/impl/KokkosExp_Host_IterateTile.hpp
@@ -2093,8 +2093,8 @@ struct HostIterateTile<RP, Functor, Tag, ValueType,
     const bool full_tile = check_iteration_bounds(m_tiledims, m_offset);
 
     Tile_Loop_Type<RP::rank, (RP::inner_direction == Iterate::Left), index_type,
-                   Tag>::apply(val, m_func, full_tile, m_offset, m_rp.m_tile,
-                               m_tiledims);
+                   Tag>::apply(val, m_func.get_functor(), full_tile, m_offset,
+                               m_rp.m_tile, m_tiledims);
   }
 
 #else
```
but only if we convert all host backends simultaneously. Thus, I would prefer not doing it here but later.

dalg24 (approved):

dalg24:
Retest this please

dalg24:
Ignored HIP failure

PR #5875, opened 2023-02-10
Disable stacktrace with musl libc

### LINKED ISSUE #5869: Suggested change to compile Kokkos 3.7 with musl-libc
**Describe the bug**
To compile Kokkos 3.7 on a system running musl-libc instead of GNU libc the following change can be applied.
```
  diff --git a/lib/kokkos/core/src/Kokkos_Macros.hpp b/lib/kokkos/core/src/Kokkos_Macros.hpp
  index 8c2d414a99..3370d908ea 100644
  --- a/lib/kokkos/core/src/Kokkos_Macros.hpp
  +++ b/lib/kokkos/core/src/Kokkos_Macros.hpp
  @@ -650,7 +650,9 @@ static constexpr bool kokkos_omp_on_host() { return false; }
   #if (defined(KOKKOS_COMPILER_GNU) || defined(KOKKOS_COMPILER_CLANG) ||  \
        defined(KOKKOS_COMPILER_INTEL) || defined(KOKKOS_COMPILER_PGI)) && \
       !defined(_WIN32)
  +#if (defined(__linux__) && defined(__GLIBC_MINOR__))
   #define KOKKOS_IMPL_ENABLE_STACKTRACE
  +#endif
   #define KOKKOS_IMPL_ENABLE_CXXABI
   #endif
   
```

There is an equivalent change in upstream that uses a C++17 only feature, so we have implemented the above change in the copy bundled with LAMMPS so we can stick with the C++14 requirement. We use the same heuristic and a few other places in LAMMPS with success and rely the fact that musl-libc is Linux-only and is only partially trying to appear like GNU libc.

This is a suggestion to add to the 3.7 maintenance branch since develop is already fixed and needs C++17 anyway.

dalg24:
Fix #5869

### DIFF
--- a/core/src/Kokkos_Macros.hpp
+++ b/core/src/Kokkos_Macros.hpp
@@ -648,7 +648,10 @@ static constexpr bool kokkos_omp_on_host() { return false; }
 #if (defined(KOKKOS_COMPILER_GNU) || defined(KOKKOS_COMPILER_CLANG) ||  \
      defined(KOKKOS_COMPILER_INTEL) || defined(KOKKOS_COMPILER_PGI)) && \
     !defined(_WIN32)
+// disable stacktrace for musl-libc
+#if !defined(__linux__) || defined(__GLIBC_MINOR__)
 #define KOKKOS_IMPL_ENABLE_STACKTRACE
+#endif
 #define KOKKOS_IMPL_ENABLE_CXXABI
 #endif
 


ldh4 (approved):
I get that this is the sufficient fix, but has this actually been checked against Kokkos 3.7 on a system running musl-libc instead of GNU libc?

dalg24:
> I get that this is the sufficient fix, but has this actually been checked against Kokkos 3.7 on a system running musl-libc instead of GNU libc?

I did not.  I assume @akohlmey did.

akohlmey:
> > I get that this is the sufficient fix, but has this actually been checked against Kokkos 3.7 on a system running musl-libc instead of GNU libc?
> 
> I did not. I assume @akohlmey did.

That is correct. I am using a Linux-to-Linux cross compiler (built using with https://github.com/richfelker/musl-cross-make) that creates fully static LAMMPS executables using musl-libc instead of GNU libc. These executables contain KOKKOS configured for OpenMP. https://download.lammps.org/static/ in case you are curious.

These executables have been succesfully tested running simulations with different paralellization settings (default, native OpenMP, Kokkos OpenMP) on a variety of Linux "flavors" (current and old) using containers or virtual machines.

I have *not* tested compiling and running on a Linux machine that uses musl-libc as the system libc.

cz4rs (approved):
Looks good to me.

stanmoore1:
Looks like this didn't make it into 4.1, will it be in 4.2?

dalg24:
I expect the issue is resolved in 4.0, just using a different workaround.
See https://github.com/kokkos/kokkos/issues/5869#issuecomment-1423247370

stanmoore1:
Ah thanks @dalg24

dalg24:
Also we did not create a release candidate branch for 4.1 yet.

stanmoore1:
> Also we did not create a release candidate branch for 4.1 yet.

I meant 4.0.1

PR #5494, opened 2022-09-26
Parallel scan implementation for OpenACC back end

pedrovalerolara:
First implementation of Parallel scan on OpenACC back end
I used an internal test code using Makefile to validate the code.
Not sure, if I have to change something else to compute the tests using CMake.

PR #5876, opened 2023-02-13
New OpenACC backend implementation for  parallel_scan with  a range policy

seyonglee:
This is a new OpenACC backend implementation for the parallel_scan constructs with Range Policy.
- Support general SCAN types
- Support the total scan result argument


seyonglee:
This new PR supersedes PR #5494

dalg24:
Retest this please

core/unit_test/CMakeLists.txt
```diff
@@ -490,7 +490,7 @@ IF(KOKKOS_ENABLE_OPENACC AND KOKKOS_CXX_COMPILER_ID STREQUAL NVHPC)
     ${CMAKE_CURRENT_BINARY_DIR}/openacc/TestOpenACC_HostSharedPtr.cpp
     ${CMAKE_CURRENT_BINARY_DIR}/openacc/TestOpenACC_HostSharedPtrAccessOnDevice.cpp
     ${CMAKE_CURRENT_BINARY_DIR}/openacc/TestOpenACC_TeamScratch.cpp
-    ${CMAKE_CURRENT_BINARY_DIR}/openacc/TestOpenACC_TestScan.cpp
+    #${CMAKE_CURRENT_BINARY_DIR}/openacc/TestOpenACC_TestScan.cpp
```
masterleinad, line 493:
```suggestion
```

core/unit_test/CMakeLists.txt
```diff
@@ -514,7 +514,7 @@ IF(KOKKOS_ENABLE_OPENACC AND KOKKOS_CXX_COMPILER_ID STREQUAL NVHPC)
     ${CMAKE_CURRENT_BINARY_DIR}/openacc/TestOpenACC_Reducers_e.cpp
     ${CMAKE_CURRENT_BINARY_DIR}/openacc/TestOpenACC_ViewMapping_b.cpp
     ${CMAKE_CURRENT_BINARY_DIR}/openacc/TestOpenACC_TeamBasic.cpp
-    ${CMAKE_CURRENT_BINARY_DIR}/openacc/TestOpenACC_Scan.cpp
+    #${CMAKE_CURRENT_BINARY_DIR}/openacc/TestOpenACC_Scan.cpp
```
masterleinad, line 517:
```suggestion
```

seyonglee:
> Retest this please

Re-ran the tests; CI fails on HIP, but I guess that those fails are not related to this PR.

core/src/OpenACC/Kokkos_OpenACC_ParallelScan_Range.hpp
```diff
@@ -0,0 +1,266 @@
...
+        }
+      }
+    }
+    if (getTotal) {
+      // size_t size = Analysis::value_size(m_functor);
+      DeepCopy<HostSpace, Kokkos::Experimental::OpenACCSpace,
+               Kokkos::Experimental::OpenACC>(m_policy.space(), m_result_ptr,
+                                              m_result_total.data(),
+                                              sizeof(ValueType));
+    }
```
masterleinad, line 194: This should be skipped if `m_result_ptr` is device-accessible because you are not writing to `m_result_total` above but to `m_result_ptr` and this would overwrite that value, right?

seyonglee: This `DeepCopy` is enabled if `getTotal` is true, which will be set to true only if `m_result_ptr` is not device-accessible.

masterleinad: I see. Is there a reason to have `getTotal` then? I there a case where the two are not the same?

seyonglee: `getTotal` was used to use the same `OpenACCParallelScanRangePolicy()` for both `ParallelScan` and `ParallelScanWithTotal` classes.
In facte, as you pointed out, the value of `getTotal` is the same as `!m_result_ptr_device_accessible && m_result_ptr != nullptr`, and thus we can remove `getTotal` parameter.

seyonglee: `getTotal` parameter is removed.

masterleinad (approved):

core/src/OpenACC/Kokkos_OpenACC_ParallelScan_Range.hpp
```diff
@@ -0,0 +1,261 @@
...
+          if (m_result_ptr_device_accessible) {
+            *m_result_ptr = update;
+          } else {
+            m_result_total() = update;
+          }
+        }
+      }
+    }
+    if (!m_result_ptr_device_accessible && m_result_ptr != nullptr) {
+      // size_t size = Analysis::value_size(m_functor);
```
dalg24, line 188: Why did you keep that commented line?

dalg24 (commented):
Please add a few comments telling what phase of the scan is happening and anything that ease understanding the algorithm

core/src/OpenACC/Kokkos_OpenACC_ParallelScan_Range.hpp
seyonglee: Deleted.

core/src/OpenACC/Kokkos_OpenACC_ParallelScan_Range.hpp
```diff
@@ -0,0 +1,261 @@
...
+      DeepCopy<HostSpace, Kokkos::Experimental::OpenACCSpace,
+               Kokkos::Experimental::OpenACC>(m_policy.space(), m_result_ptr,
+                                              m_result_total.data(),
+                                              sizeof(ValueType));
+    }
+
+#pragma acc exit data delete (functor, chunk_values, offset_values, \
+                              final_reducer)async(async_arg)
+    acc_wait(async_arg);
```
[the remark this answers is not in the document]
seyonglee: We need this synchronization since the parallel scan implementation uses several local variables (`chunk_values`, `offset_values`, `m_result_total`, `element_values_owner`) whose lifespan is the enclosing function (`OpenACCParallelScanRangePolicy`). All asynchronous executions here should finish before exiting the enclosing function.

seyonglee:
> Please add a few comments telling what phase of the scan is happening and anything that ease understanding the algorithm
Added a reference to the parallel scan algorithm implemented in the `OpenACCParallelScanRangePolicy` function.

dalg24:
Re-test this please

dalg24 (approved):

dalg24:
Updated your branch that would not build when merging into develop because of #5874
Should be good to merge after the CI runs.

dalg24:
OpenACC build passed.

PR #5877, opened 2023-02-13
Add UnorderedMapInsertOps for coo2crs

e10harvey:
To implement the sparse conversion kernel, coo2crs, I would like to have duplicate key insertions accumulate values in a `Kokkos::UnorderedMap`. This PR adds `UnorderedMapInsertOps::AtomicAdd` and `UnorderedMapInsertOps::NoOp` (the current insertion behavior for duplicate keys).

Related to https://github.com/kokkos/kokkos-kernels/issues/1164.

Going forward, I would also like to extend `Kokkos::UnorderedMap` to support shared memory and add `UnorderedMapInsertOps::Overwrite`.

### DIFF
--- a/containers/src/Kokkos_UnorderedMap.hpp
+++ b/containers/src/Kokkos_UnorderedMap.hpp
@@ -120,22 +120,30 @@ class UnorderedMapInsertResult {
 };
 
 /// \class UnorderedMapInsertOpTypes
+///
 /// \brief Operations applied to the values array upon subsequent insertions.
+///
 /// The default behavior when a k,v pair already exists in the UnorderedMap is
 /// to perform no operation. Alternatively, the caller may select to
 /// instantiate the UnorderedMap with the AtomicAdd insert operator such that
 /// duplicate keys accumulate values into the given values array entry.
-template <class ValueTypeView = int, class ValuesIdxType = uint32_t,
-          class ValueType = int>
+/// \tparam ValueTypeView The UnorderedMap value array type.
+/// \tparam ValuesIdxType The index type for lookups in the value array.
+///
+/// Supported operations:
+///   NoOp:      the first key inserted stores the associated value.
+///   AtomicAdd: duplicate key insertions sum values together.
+template <class ValueTypeView = int, class ValuesIdxType = uint32_t>
 struct UnorderedMapInsertOpTypes {
+  using value_type = typename ValueTypeView::non_const_value_type;
   struct NoOp {
     KOKKOS_FUNCTION
-    void op(ValueTypeView, ValuesIdxType, const ValueType *) const {}
+    void op(ValueTypeView, ValuesIdxType, const value_type *) const {}
   };
   struct AtomicAdd {
     KOKKOS_FUNCTION
     void op(ValueTypeView values, ValuesIdxType values_idx,
-            const ValueType *v) const {
+            const value_type *v) const {
       Kokkos::atomic_add(values.data() + values_idx, *v);
     }
   };
@@ -196,6 +204,11 @@ struct UnorderedMapInsertOpTypes {
 /// \tparam EqualTo Definition of the equality function for instances of
 ///   <tt>Key</tt>.  The default will do a bitwise equality comparison.
 ///
+/// \tparam InsertOp Definition of the insert function for duplicate
+///   insertions of <tt>Key</tt>. The default is to set value to the
+///   first key, value pair that is inserted. Please see
+///   Kokkos::UnorderedMapInsertOpTypes for more details.
+///
 template <typename Key, typename Value,
           typename Device   = Kokkos::DefaultExecutionSpace,
           typename Hasher   = pod_hash<std::remove_const_t<Key>>,
@@ -204,9 +217,7 @@ template <typename Key, typename Value,
               Kokkos::View<std::remove_const_t<std::conditional_t<
                                std::is_void_v<Value>, int, Value>> *,
                            Device>,
-              uint32_t,
-              std::remove_const_t<
-                  std::conditional_t<std::is_void_v<Value>, int, Value>>>::NoOp>
+              uint32_t>::NoOp>
 class UnorderedMap {
  private:
   using host_mirror_space =
@@ -295,16 +306,16 @@ class UnorderedMap {
   //! \name Public member functions
   //@{
 
-  // clang-format off
   /// \brief Constructor
   ///
   /// \param capacity_hint [in] Initial guess of how many unique keys will be
-  /// inserted into the map
-  /// \param hash [in] Hasher function for \c Key instances.
-  /// \param equal_to [in] The operator used for determining if two keys are equal.
-  /// \param insert_op [in] The operator used for combining values if a value already exists. 
-  /// The default value usually suffices.
-  // clang-format on
+  ///                           inserted into the map.
+  /// \param hash          [in] Hasher function for \c Key instances.  The
+  ///                           default value usually suffices.
+  /// \param equal_to      [in] The operator used for determining if two
+  ///                           keys are equal.
+  /// \param insert_op     [in] The operator used for combining values if a
+  ///                           key already exists.
   UnorderedMap(size_type capacity_hint = 0, hasher_type hasher = hasher_type(),
                equal_to_type equal_to   = equal_to_type(),
                insert_op_type insert_op = insert_op_type())
@@ -722,7 +733,7 @@ class UnorderedMap {
       std::enable_if_t<
           Impl::UnorderedMapCanAssign<declared_key_type, declared_value_type,
                                       SKey, SValue>::value,
-          int>                 = 0)
+          int> = 0)
       : m_bounded_insert(src.m_bounded_insert),
         m_hasher(src.m_hasher),
         m_equal_to(src.m_equal_to),
@@ -807,29 +818,6 @@ class UnorderedMap {
 
       *this = tmp;
     }
-  }
-
-  KOKKOS_FORCEINLINE_FUNCTION
-  bool is_op_noop() const {
-    using local_value_type = std::remove_const_t<std::conditional_t<std::is_void_v<Value>, int, Value>>;
-    return std::is_base_of_v<
-        insert_op_type,
-        typename UnorderedMapInsertOpTypes<
-            Kokkos::View<local_value_type *, Device>,
-            uint32_t, local_value_type>::NoOp>;
-  }
-
-  KOKKOS_FORCEINLINE_FUNCTION
-  bool is_op_add() const {
-    return std::is_base_of_v<
-        insert_op_type,
-        typename UnorderedMapInsertOpTypes<
-            Kokkos::View<std::remove_const_t<std::conditional_t<
-                             std::is_void_v<Value>, int, Value>> *,
-                         Device>,
-            uint32_t,
-            std::remove_const_t<std::conditional_t<std::is_void_v<Value>, int,
-                                                   Value>>>::AtomicAdd>;
   }
 
   //@}
--- a/containers/src/Kokkos_UnorderedMap.hpp
+++ b/containers/src/Kokkos_UnorderedMap.hpp
@@ -569,7 +569,9 @@ class UnorderedMap {
         }
 
         result.set_existing(curr, free_existing);
-        if (!is_set) arg_insert_op.op(m_values, curr, v);
+        if constexpr (!is_set) {
+          arg_insert_op.op(m_values, curr, v);
+        }
         not_done = false;
       }
       //------------------------------------------------------------
--- a/containers/src/Kokkos_UnorderedMap.hpp
+++ b/containers/src/Kokkos_UnorderedMap.hpp
@@ -482,6 +482,11 @@ class UnorderedMap {
   KOKKOS_INLINE_FUNCTION insert_result
   insert(key_type const &k, impl_value_type const &v = impl_value_type(),
          InsertOpType arg_insert_op = InsertOpType()) const {
+    if constexpr (is_set) {
+      static_assert(std::is_same_v<InsertOpType, default_op_type>,
+                    "Insert Operations are not supported on sets.");
+    }
+
     insert_result result;
 
     if (!is_insertable_map || capacity() == 0u ||
--- a/containers/src/Kokkos_UnorderedMap.hpp
+++ b/containers/src/Kokkos_UnorderedMap.hpp
@@ -122,18 +122,18 @@ class UnorderedMapInsertResult {
 /// \class UnorderedMapInsertOpTypes
 /// \brief Operations applied to the values array upon subsequent insertions.
 /// The default behavior when a k,v pair already exists in the UnorderedMap is
-/// to perform no operation. Alternatively, the caller may selected to
+/// to perform no operation. Alternatively, the caller may select to
 /// instantiate the UnorderedMap with the AtomicAdd insert operator such that
 /// duplicate keys accumulate values into the given values array entry.
 template <class ValueTypeView = int, class ValuesIdxType = uint32_t,
           class ValueType = int>
 struct UnorderedMapInsertOpTypes {
   struct NoOp {
-    KOKKOS_INLINE_FUNCTION
+    KOKKOS_FUNCTION
     void op(ValueTypeView, ValuesIdxType, const ValueType *) const {}
   };
   struct AtomicAdd {
-    KOKKOS_INLINE_FUNCTION
+    KOKKOS_FUNCTION
     void op(ValueTypeView values, ValuesIdxType values_idx,
             const ValueType *v) const {
       Kokkos::atomic_add(values.data() + values_idx, *v);
@@ -722,8 +722,7 @@ class UnorderedMap {
       std::enable_if_t<
           Impl::UnorderedMapCanAssign<declared_key_type, declared_value_type,
                                       SKey, SValue>::value,
-          int>                 = 0,
-      insert_op_type insert_op = insert_op_type())
+          int>                 = 0)
       : m_bounded_insert(src.m_bounded_insert),
         m_hasher(src.m_hasher),
         m_equal_to(src.m_equal_to),
@@ -812,15 +811,12 @@ class UnorderedMap {
 
   KOKKOS_FORCEINLINE_FUNCTION
   bool is_op_noop() const {
+    using local_value_type = std::remove_const_t<std::conditional_t<std::is_void_v<Value>, int, Value>>;
     return std::is_base_of_v<
         insert_op_type,
         typename UnorderedMapInsertOpTypes<
-            Kokkos::View<std::remove_const_t<std::conditional_t<
-                             std::is_void_v<Value>, int, Value>> *,
-                         Device>,
-            uint32_t,
-            std::remove_const_t<
-                std::conditional_t<std::is_void_v<Value>, int, Value>>>::NoOp>;
+            Kokkos::View<local_value_type *, Device>,
+            uint32_t, local_value_type>::NoOp>;
   }
 
   KOKKOS_FORCEINLINE_FUNCTION
--- a/containers/src/Kokkos_UnorderedMap.hpp
+++ b/containers/src/Kokkos_UnorderedMap.hpp
@@ -481,12 +481,10 @@ class UnorderedMap {
   template <typename InsertOpType = default_op_type>
   KOKKOS_INLINE_FUNCTION insert_result
   insert(key_type const &k, impl_value_type const &v = impl_value_type(),
-         InsertOpType arg_insert_op = InsertOpType()) const {
+         [[maybe_unused]] InsertOpType arg_insert_op = InsertOpType()) const {
     if constexpr (is_set) {
       static_assert(std::is_same_v<InsertOpType, default_op_type>,
                     "Insert Operations are not supported on sets.");
-      // Silence error #869: parameter "arg_insert_op" was never referenced
-      (void)arg_insert_op;
     }
 
     insert_result result;
--- a/containers/src/Kokkos_UnorderedMap.hpp
+++ b/containers/src/Kokkos_UnorderedMap.hpp
@@ -485,6 +485,8 @@ class UnorderedMap {
     if constexpr (is_set) {
       static_assert(std::is_same_v<InsertOpType, default_op_type>,
                     "Insert Operations are not supported on sets.");
+      // Silence error #869: parameter "arg_insert_op" was never referenced
+      (void)arg_insert_op;
     }
 
     insert_result result;
--- a/containers/src/Kokkos_UnorderedMap.hpp
+++ b/containers/src/Kokkos_UnorderedMap.hpp
@@ -204,7 +204,6 @@ struct UnorderedMapInsertOpTypes {
 /// \tparam EqualTo Definition of the equality function for instances of
 ///   <tt>Key</tt>.  The default will do a bitwise equality comparison.
 ///
-///
 template <typename Key, typename Value,
           typename Device  = Kokkos::DefaultExecutionSpace,
           typename Hasher  = pod_hash<std::remove_const_t<Key>>,
--- a/containers/src/Kokkos_UnorderedMap.hpp
+++ b/containers/src/Kokkos_UnorderedMap.hpp
@@ -313,7 +313,6 @@ class UnorderedMap {
         m_hasher(hasher),
         m_equal_to(equal_to),
         m_size(std::make_shared<size_type>()),
-        m_insert_op(insert_op),
         m_available_indexes(calculate_capacity(capacity_hint)),
         m_hash_lists(view_alloc(WithoutInitializing, "UnorderedMap hash list"),
                      Impl::find_hash_size(capacity())),
@@ -769,7 +768,6 @@ class UnorderedMap {
       tmp.m_hasher            = src.m_hasher;
       tmp.m_equal_to          = src.m_equal_to;
       tmp.m_size              = src.m_size;
-      tmp.m_insert_op         = src.m_insert_op;
       tmp.m_available_indexes = bitset_type(src.capacity());
       tmp.m_hash_lists        = size_type_view(
           view_alloc(WithoutInitializing, "UnorderedMap hash list"),
@@ -863,7 +861,6 @@ class UnorderedMap {
   hasher_type m_hasher;
   equal_to_type m_equal_to;
   std::shared_ptr<size_type> m_size;
-  insert_op_type m_insert_op;
   bitset_type m_available_indexes;
   size_type_view m_hash_lists;
   size_type_view m_next_index;
--- a/containers/src/Kokkos_UnorderedMap.hpp
+++ b/containers/src/Kokkos_UnorderedMap.hpp
@@ -119,6 +119,28 @@ class UnorderedMapInsertResult {
   uint32_t m_status;
 };
 
+/// \class UnorderedMapInsertOpTypes
+/// \brief Operations applied to the values array upon subsequent insertions.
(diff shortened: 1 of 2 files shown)


containers/src/Kokkos_UnorderedMap.hpp
```diff
@@ -767,6 +810,32 @@ class UnorderedMap {
...
+  bool is_op_noop() const {
+    return std::is_base_of_v<
+        insert_op_type,
+        typename UnorderedMapInsertOpTypes<
+            Kokkos::View<std::remove_const_t<std::conditional_t<
+                             std::is_void_v<Value>, int, Value>> *,
+                         Device>,
+            uint32_t,
+            std::remove_const_t<
+                std::conditional_t<std::is_void_v<Value>, int, Value>>>::NoOp>;
```
masterleinad, line 823: That's pretty hard to read 🙂 
```suggestion
    using local_value_type = std::remove_const_t<std::conditional_t<std::is_void_v<Value>, int, Value>>;
    return std::is_base_of_v<
        insert_op_type,
        typename UnorderedMapInsertOpTypes<
            Kokkos::View<local_value_type *, Device>,
            uint32_t, local_value_type>::NoOp>;
```
and similarly below.

containers/unit_tests/TestUnorderedMap.hpp
```diff
@@ -68,7 +82,20 @@ struct TestInsert {
...
+      ++failed_count;
+    else {
+      auto idx = map.find(key);
+      auto ptr = expected_values.data();
+      if (map.is_op_add()) {
+        // Kokkos::abort("adding");
+        Kokkos::atomic_add(&((ptr + idx)[0].v), i);
+      } else if (ret.success() && map.is_op_noop()) {
+        // Kokkos::abort("noop");
+        Kokkos::atomic_store(&((ptr + idx)[0].v), i);
```
masterleinad, line 96:
```suggestion
        Kokkos::atomic_add(&((ptr + idx)[0].v), i);
      } else if (ret.success() && map.is_op_noop()) {
        Kokkos::atomic_store(&((ptr + idx)[0].v), i);
```

containers/src/Kokkos_UnorderedMap.hpp
```diff
@@ -767,6 +810,32 @@ class UnorderedMap {
     }
   }
 
+  KOKKOS_FORCEINLINE_FUNCTION
+  bool is_op_noop() const {
```
masterleinad, line 814: Do we need these functions at all?

e10harvey: The main need is for asserting in the unit-test. I am open to alternative approaches, perhaps just in the unit-test. Thoughts? I will try to remove these.

containers/src/Kokkos_UnorderedMap.hpp
```diff
@@ -175,9 +197,16 @@ class UnorderedMapInsertResult {
...
+          typename Device   = Kokkos::DefaultExecutionSpace,
+          typename Hasher   = pod_hash<std::remove_const_t<Key>>,
+          typename EqualTo  = pod_equal_to<std::remove_const_t<Key>>,
+          typename InsertOp = typename UnorderedMapInsertOpTypes<
+              Kokkos::View<std::remove_const_t<std::conditional_t<
+                               std::is_void_v<Value>, int, Value>> *,
+                           Device>,
+              uint32_t,
+              std::remove_const_t<
+                  std::conditional_t<std::is_void_v<Value>, int, Value>>>::NoOp>
```
masterleinad, line 209: You should also document `InsertOp`.

containers/src/Kokkos_UnorderedMap.hpp
```diff
@@ -119,6 +119,28 @@ class UnorderedMapInsertResult {
...
 };
 
+/// \class UnorderedMapInsertOpTypes
+/// \brief Operations applied to the values array upon subsequent insertions.
+/// The default behavior when a k,v pair already exists in the UnorderedMap is
+/// to perform no operation. Alternatively, the caller may select to
+/// instantiate the UnorderedMap with the AtomicAdd insert operator such that
+/// duplicate keys accumulate values into the given values array entry.
+template <class ValueTypeView = int, class ValuesIdxType = uint32_t,
+          class ValueType = int>
```
masterleinad, line 129: Do you need both `ValueTypeView` and `ValueType`? AFAICT, `impl_value_type` is `typename value_type_view::non_const_value_type` in the implementation of `UnorderedMap`, right?

masterleinad:
Apparently, this fails on the `OpenMP` backend and you will also need to fix the indentation. 🙂

containers/src/Kokkos_UnorderedMap.hpp
e10harvey: Thanks for all the helpful feedback, @masterleinad. I think `ValueTypeView` is required since, if the user wants to specify a insert op prior to defining the `map_type`, UnorderedMapInsertOpTypes must be defined before `map_type`. However, perhaps there is a more elegant way to achieve this same result with templates?

containers/src/Kokkos_UnorderedMap.hpp
```diff
@@ -119,6 +119,28 @@ class UnorderedMapInsertResult {
...
+    void op(ValueTypeView, ValuesIdxType, const ValueType *) const {}
+  };
+  struct AtomicAdd {
+    KOKKOS_FUNCTION
+    void op(ValueTypeView values, ValuesIdxType values_idx,
+            const ValueType *v) const {
+      Kokkos::atomic_add(values.data() + values_idx, *v);
+    }
+  };
+};
```
masterleinad, line 142: Yes, I agree that we need `ValueTypeView` but I was thinking that you could do 
```suggestion
/// \class UnorderedMapInsertOpTypes
/// \brief Operations applied to the values array upon subsequent insertions.
/// The default behavior when a k,v pair already exists in the UnorderedMap is
/// to perform no operation. Alternatively, the caller may select to
/// instantiate the UnorderedMap with the AtomicAdd insert operator such that
/// duplicate keys accumulate values into the given values array entry.
template <class ValueTypeView = int, class ValuesIdxType = uint32_t>
struct UnorderedMapInsertOpTypes {
  using ValueType = typename ValueTypeView::non_const_value_type;
  struct NoOp {
    KOKKOS_FUNCTION
    void op(ValueTypeView, ValuesIdxType, const ValueType *) const {}
  };
  struct AtomicAdd {
    KOKKOS_FUNCTION
    void op(ValueTypeView values, ValuesIdxType values_idx,
            const ValueType *v) const {
      Kokkos::atomic_add(values.data() + values_idx, *v);
    }
  };
};
```
avoiding `ValueType` to be a template parameter.

e10harvey: Ah, certainly.

masterleinad:
Looks like there is still a segfault (at least for `OpenMP`).

containers/unit_tests/TestUnorderedMap.hpp
```diff
@@ -55,6 +65,15 @@ struct TestInsert {
...
     execution_space().fence();
+
+    if (!rehash_on_fail) {
+      for (unsigned i = 0; i < map.size(); i++) {
+        auto map_idx = expected_values(i).map_idx;
+        if (map_idx != static_cast<unsigned>(~0)) {
+          ASSERT_EQ(expected_values(map_idx).v, map.value_at(map_idx));
+        }
+      }
+    }
```
masterleinad, line 76: This is running on the host but the VIew is allocated in the default memory space so our GPU CI fails:
```
 Kokkos::View ERROR: attempt to access inaccessible memory space (label="ExpectedValues")
```

containers/src/Kokkos_UnorderedMap.hpp
```diff
@@ -119,6 +119,36 @@ class UnorderedMapInsertResult {
...
+///
+/// Supported operations:
+///   NoOp:      the first key inserted stores the associated value.
+///   AtomicAdd: duplicate key insertions sum values together.
+template <class ValueTypeView = int, class ValuesIdxType = uint32_t>
+struct UnorderedMapInsertOpTypes {
+  using value_type = typename ValueTypeView::non_const_value_type;
+  struct NoOp {
+    KOKKOS_FUNCTION
+    void op(ValueTypeView, ValuesIdxType, const value_type *) const {}
```
dalg24, line 141: Why did you decide to pass a pointer to const rather than passing by value for the last argument?
What types to you expect `value_type` might be?  It is a pessimization for arithmetic types.

containers/src/Kokkos_UnorderedMap.hpp
```diff
@@ -119,6 +119,36 @@ class UnorderedMapInsertResult {
...
+/// to perform no operation. Alternatively, the caller may select to
+/// instantiate the UnorderedMap with the AtomicAdd insert operator such that
+/// duplicate keys accumulate values into the given values array entry.
+/// \tparam ValueTypeView The UnorderedMap value array type.
+/// \tparam ValuesIdxType The index type for lookups in the value array.
+///
+/// Supported operations:
+///   NoOp:      the first key inserted stores the associated value.
+///   AtomicAdd: duplicate key insertions sum values together.
+template <class ValueTypeView = int, class ValuesIdxType = uint32_t>
```
dalg24, line 136: What does that even mean a default `ValueTypeView` that is `int`?  That would not compile.

containers/src/Kokkos_UnorderedMap.hpp
```diff
@@ -296,6 +344,18 @@ class UnorderedMap {
...
+  /// a deeper scope than the original map, the mutable
+  /// m_size member within the original map gets reset
+  /// to 0 when the host mirror object falls out of scope.
+  /// Calling set_flag upon destruction ensures that m_size
+  /// will be recomputed in the parent scope. I think this
+  /// is only a problem when the original map resides on
+  /// the host since some of the host mirror members likely
+  /// point to the same memory locations as the original map
+  /// members.
+  ~UnorderedMap() { set_flag(modified_idx); }
```
dalg24, line 357: Is that a drive-by fix?

e10harvey: No? This was triaged first and the addressed with a destructor.

containers/src/Kokkos_UnorderedMap.hpp
e10harvey: Please suggest an alternative default type.

containers/src/Kokkos_UnorderedMap.hpp
e10harvey: The signatures of `op` must match.

containers/src/Kokkos_UnorderedMap.hpp
dalg24: No default type

containers/src/Kokkos_UnorderedMap.hpp
dalg24: So?  They both could take a value.  The comment applies to both.

crtrott: Agreed there is no reason these guys take the value as a pointer

containers/src/Kokkos_UnorderedMap.hpp
```diff
@@ -174,10 +204,20 @@ class UnorderedMapInsertResult {
...
-          typename Hasher  = pod_hash<std::remove_const_t<Key>>,
-          typename EqualTo = pod_equal_to<std::remove_const_t<Key>>>
+          typename Device   = Kokkos::DefaultExecutionSpace,
+          typename Hasher   = pod_hash<std::remove_const_t<Key>>,
+          typename EqualTo  = pod_equal_to<std::remove_const_t<Key>>,
+          typename InsertOp = typename UnorderedMapInsertOpTypes<
+              Kokkos::View<std::remove_const_t<std::conditional_t<
+                               std::is_void_v<Value>, int, Value>> *,
+                           Device>,
+              uint32_t>::NoOp>
```
crtrott, line 220: Adding a template parameter is a breaking change we can't do that really. We could discuss adding this only if deprecated code off, but we can't simply add it. Alternatively, why can't we have an overload of insert which takes an insert op? Is there a reason this has to be a global template for the class?

containers/src/Kokkos_UnorderedMap.hpp
crtrott: We probably want this fix in a separate PR with a test for whatever this fixes.

containers/src/Kokkos_UnorderedMap.hpp
```diff
@@ -532,6 +593,7 @@ class UnorderedMap {
         }
 
         result.set_existing(curr, free_existing);
+        if (!is_set) m_insert_op.op(m_values, curr, &v);
```
crtrott, line 596: yeah this is the only place where this is used. I think we can just have an overload of insert which takes the insert op. Or maybe even have a different name: insert_accumulate or so. Do you see any issue with that approach?

crtrott (changes requested):
The proposed approach is a breaking change: why can't we just have an insert function overload which takes an op, or add a insert_accumulate function?

containers/src/Kokkos_UnorderedMap.hpp
e10harvey: Thanks for the feedback. I'll make another pass soon.

containers/src/Kokkos_UnorderedMap.hpp
e10harvey: The only issue I see would be a lot of duplicate code.

containers/src/Kokkos_UnorderedMap.hpp
e10harvey: This seems like the lightest-weight way to add this operation as well as following the implementation style of UnorderedMap. What about keeping the current approach with a single constructor overload, rather than a new insert method, the constructor overload could include the InsertOp type and constructor parameter?

e10harvey: Thinking about this more now, I see no reason for making the InsertOp a global template for the class other than following the existing implementation style of UnorderedMap (e.g. EqualTo, etc.). I will try refactoring with a new insert method instead. Can you elaborate on how the InsertOp template parameter is a breaking change?

containers/src/Kokkos_UnorderedMap.hpp
e10harvey: Thinking about this more now, a pointer was used to work around compiler errors when value_type = void for non-insertable UnorderdMaps.

containers/src/Kokkos_UnorderedMap.hpp
e10harvey: After trying to remove the InsertOp as a global template parameter, I found that deep copy no longer works. I will look into using insert method overloads rather than a class member.

containers/unit_tests/TestUnorderedMap.hpp
e10harvey: I have been marking feedback I've addressed as resolved. Is the process for the reviewee to mark comments as resolved or should the reviewer do that?

e10harvey:
This is ready for another round of reviews. CI tests for [87c3691](https://github.com/kokkos/kokkos/pull/5877/commits/87c3691558c9c73599455123705257204f9add3d) will fail until #5899 merges.

crtrott:
Needs a rebase. Question at everyone but especially @masterleinad @nmm0 @nliber @dalg24: Do we like the way the insert ops are grouped here mechanically as nested classes in a separate class?

There are alternatives: 
- nested namespace
- make them nested classes in UnorderedMap itself
- named ops (this is what ScatterView does where the ops are called ScatterAdd etc. in the Kokkos::Experimental namespace)

At least we should check for where we do other stuff like that and what approach we take so we can end up with a consistent way of doing this.

crtrott:
Oh just to confirm: the overload approach of insert with a default insert mechanism is good now.

containers/src/Kokkos_UnorderedMap.hpp
```diff
@@ -186,6 +217,11 @@ class UnorderedMap {
  public:
   //! \name Public types and constants
   //@{
+  using default_op_type = typename UnorderedMapInsertOpTypes<
```
nmm0, line 220: Since you are always passing in `m_values` why not use the type of `m_values` here -- `value_type_view`

nmm0 (changes requested):
Could you add a test for `void` value'd maps (i.e. sets)? What is the expected behavior of the ops when value is `void`?

nmm0:
> > Could you add a test for `void` value'd maps (i.e. sets)? What is the expected behavior of the ops when value is `void`?
> 
> Hi @nmm0. Thanks for your review. The op is not executed when the map is void value'd due to `if(!is_set)`. This is already tested and handled the same way it was before this PR was opened. The specific line of code introduced for getting `void` value'd maps with these insert ops to still build is: https://github.com/kokkos/kokkos/pull/5877/files#diff-2c8acfc1ac5cbbb4cc3da14e466f42c85e415b540c815bf2630c42ec2ac09d59R576.

Thanks for the clarification!

Let's then do:

```cpp
if constexpr (!is_set) {
  arg_insert_op.op(m_values, curr, v);
}
```
Then we don't have to worry about it compiling with a set. Also, in that case, I think we should probably not allow the op to be even specified if inserting into a set.

This is something we probably need to deal with now, because if a user specifies an InsertOp for a set and we clean up the interface later then it would be a breaking change.

I'd like to get other people's opinion on it but one suggestion would be to just do a static_assert, which would be a quick fix rather than making an overload:

```cpp
if constexpr (is_set) {
  static_assert(std::is_same_v<InsertOpType, default_op_type>);
}

**Comment by e10harvey** [comment]
The Jenkins SYCL-OneApi failure is unrelated to these changes:
```
43: [ RUN      ] std_algorithms_sorting_ops_test.is_sorted_until
43: is_sorted_until: dynamic_view, all overloads 
43: /var/jenkins/workspace/Kokkos/algorithms/unit_tests/TestStdAlgorithmsIsSortedUntil.cpp:150: Failure
43: Expected equality of these values:
43:   r3
43:     Which is: 32-byte object <70-0B C3-02 00-00 00-00 80-00 A0-93 9C-7F 00-00 89-8C 01-00 00-00 00-00 F9-02 00-00 00-00 00-00>
43:   gold
43:     Which is: 32-byte object <70-0B C3-02 00-00 00-00 80-00 A0-93 9C-7F 00-00 89-8C 01-00 00-00 00-00 9C-00 00-00 00-00 00-00>
43: /var/jenkins/workspace/Kokkos/algorithms/unit_tests/TestStdAlgorithmsIsSortedUntil.cpp:151: Failure
43: Expected equality of these values:
43:   r4
43:     Which is: 32-byte object <70-0B C3-02 00-00 00-00 80-00 A0-93 9C-7F 00-00 89-8C 01-00 00-00 00-00 01-03 00-00 00-00 00-00>
43:   gold
43:     Which is: 32-byte object <70-0B C3-02 00-00 00-00 80-00 A0-93 9C-7F 00-00 89-8C 01-00 00-00 00-00 9C-00 00-00 00-00 00-00>
43: /var/jenkins/workspace/Kokkos/algorithms/unit_tests/TestStdAlgorithmsIsSortedUntil.cpp:165: Failure
43: Expected equality of these values:
43:   r3
43:     Which is: 32-byte object <70-0B C3-02 00-00 00-00 80-00 A0-93 9C-7F 00-00 89-8C 01-00 00-00 00-00 F9-02 00-00 00-00 00-00>
43:   gold
43:     Which is: 32-byte object <70-0B C3-02 00-00 00-00 80-00 A0-93 9C-7F 00-00 89-8C 01-00 00-00 00-00 9C-00 00-00 00-00 00-00>
43: /var/jenkins/workspace/Kokkos/algorithms/unit_tests/TestStdAlgorithmsIsSortedUntil.cpp:166: Failure
43: Expected equality of these values:
43:   r4
43:     Which is: 32-byte object <70-0B C3-02 00-00 00-00 80-00 A0-93 9C-7F 00-00 89-8C 01-00 00-00 00-00 01-03 00-00 00-00 00-00>
43:   gold
43:     Which is: 32-byte object <70-0B C3-02 00-00 00-00 80-00 A0-93 9C-7F 00-00 89-8C 01-00 00-00 00-00 9C-00 00-00 00-00 00-00>
43: is_sorted_until: stride2_view, all overloads 
43: is_sorted_until: stride3_view, all overloads 
43: [  FAILED  ] std_algorithms_sorting_ops_test.is_sorted_until (72 ms)
```

Full console: https://cloud.cees.ornl.gov/jenkins-ci/blue/rest/organizations/jenkins/pipelines/Kokkos/runs/12424/nodes/49/steps/252/log/?start=0.

**Comment by e10harvey** [comment]
The Jenkins SYCL-OneApi failed again:
```
43: [ RUN      ] std_algorithms_sorting_ops_test.is_sorted_until
43: is_sorted_until: dynamic_view, all overloads 
43: /var/jenkins/workspace/Kokkos/algorithms/unit_tests/TestStdAlgorithmsIsSortedUntil.cpp:148: Failure
43: Expected equality of these values:
43:   r1
43:     Which is: 32-byte object <40-15 04-02 00-00 00-00 80-00 A0-B3 AD-7F 00-00 89-8C 01-00 00-00 00-00 89-8C 01-00 00-00 00-00>
43:   gold
43:     Which is: 32-byte object <40-15 04-02 00-00 00-00 80-00 A0-B3 AD-7F 00-00 89-8C 01-00 00-00 00-00 9C-00 00-00 00-00 00-00>
43: /var/jenkins/workspace/Kokkos/algorithms/unit_tests/TestStdAlgorithmsIsSortedUntil.cpp:163: Failure
43: Expected equality of these values:
43:   r1
43:     Which is: 32-byte object <40-15 04-02 00-00 00-00 80-00 A0-B3 AD-7F 00-00 89-8C 01-00 00-00 00-00 89-8C 01-00 00-00 00-00>
43:   gold
43:     Which is: 32-byte object <40-15 04-02 00-00 00-00 80-00 A0-B3 AD-7F 00-00 89-8C 01-00 00-00 00-00 9C-00 00-00 00-00 00-00>
43: is_sorted_until: stride2_view, all overloads 
43: is_sorted_until: stride3_view, all overloads 
43: [  FAILED  ] std_algorithms_sorting_ops_test.is_sorted_until (143 ms)
```
Can anyone confirm that this failure is unrelated to these changes?

**Comment by masterleinad** [comment]
I would think that the failure is unrelated; the test is not using `Kokkos::UnorderedMap`.

File: containers/src/Kokkos_UnorderedMap.hpp [hunk 14]
```diff
@@ -174,6 +204,7 @@ class UnorderedMapInsertResult {
 /// \tparam EqualTo Definition of the equality function for instances of
 ///   <tt>Key</tt>.  The default will do a bitwise equality comparison.
 ///
+///
```
masterleinad, line 207: Any reason for this change?
```suggestion
///
```

containers/src/Kokkos_UnorderedMap.hpp
```diff
@@ -442,9 +475,20 @@ class UnorderedMap {
...
+  ///                       Kokkos::UnorderedMapInsertOpTypes for more ops.
+  template <typename InsertOpType = default_op_type>
+  KOKKOS_INLINE_FUNCTION insert_result
+  insert(key_type const &k, impl_value_type const &v = impl_value_type(),
+         InsertOpType arg_insert_op = InsertOpType()) const {
+    if constexpr (is_set) {
+      static_assert(std::is_same_v<InsertOpType, default_op_type>,
+                    "Insert Operations are not supported on sets.");
+      // Silence error #869: parameter "arg_insert_op" was never referenced
+      (void)arg_insert_op;
```
masterleinad, line 489:
```suggestion
         [[maybe_unused]] InsertOpType arg_insert_op = InsertOpType()) const {
    if constexpr (is_set) {
      static_assert(std::is_same_v<InsertOpType, default_op_type>,
                    "Insert Operations are not supported on sets.");
```

e10harvey:
I think this is ready to merge. The CI failure appears to be unrelated to these change.

masterleinad (approved):
Looks OK to me.

nmm0:
> Needs a rebase. Question at everyone but especially @masterleinad @nmm0 @nliber @dalg24: Do we like the way the insert ops are grouped here mechanically as nested classes in a separate class?
> 
> There are alternatives:
> 
>     * nested namespace
> 
>     * make them nested classes in UnorderedMap itself
> 
>     * named ops (this is what ScatterView does where the ops are called ScatterAdd etc. in the Kokkos::Experimental namespace)
> 
> 
> At least we should check for where we do other stuff like that and what approach we take so we can end up with a consistent way of doing this.

@crtrott did you get satisfactory feedback on this?

nmm0 (changes requested):
I think this is ready depending on what we decide for the convention for Ops

However it needs a changelog entry ([4.1](https://github.com/kokkos/kokkos/issues/5902)) and a corresponding PR for the docs

e10harvey:
> > I think this is ready depending on what we decide for the convention for Ops
> > However it needs a changelog entry ([4.1](https://github.com/kokkos/kokkos/issues/5902)) and a corresponding PR for the docs
> 
> Thinking about Christian's question some more now and in particular how ScatterView handles insertion operators, we should be consistent if possible. @nmm0, @dalg24: Is ScatterView the only other container that supports "operations" in Kokkos core?

It looks like ScatterView has the concept of a ScatterValue type with and specializes on `ScatterValue` for the supported "operation" types like ScatterSum. Thinking about how to support extension of insertion operations in this way is not clear to me. Before this PR, the `Kokkos::UnorderedMap` only requires `operator=` for the given `value_type`. With this PR, we only change what operation is used if the given `key` already exists. So, from my perspective, I think the current approach using `arg_insert_op` is cleaner than introducing the concept of a `ValueType` with a `ValueOp` to `UnorderedMap` since the body of `insert` is not exposed to the user like operations on `ScatterView`. In a later PR, we could revisit `UnorderedMapInsertOpTypes` if they are found to be difficult to use and extend.

PhilMiller:
Retest this please

PhilMiller:
Rerunning the tests, since it looks like Jenkins had a hiccup

e10harvey:
GCC 8.4.0 timed out and the SYCL-OneApi test failed again:
```
43: is_sorted_until: dynamic_view, all overloads 
43: /var/jenkins/workspace/Kokkos/algorithms/unit_tests/TestStdAlgorithmsIsSortedUntil.cpp:149: Failure
43: Expected equality of these values:
43:   r2
43:     Which is: 32-byte object <10-54 8F-02 00-00 00-00 80-00 A0-D3 9A-7F 00-00 89-8C 01-00 00-00 00-00 E1-02 00-00 00-00 00-00>
43:   gold
43:     Which is: 32-byte object <10-54 8F-02 00-00 00-00 80-00 A0-D3 9A-7F 00-00 89-8C 01-00 00-00 00-00 9C-00 00-00 00-00 00-00>
43: /var/jenkins/workspace/Kokkos/algorithms/unit_tests/TestStdAlgorithmsIsSortedUntil.cpp:164: Failure
43: Expected equality of these values:
43:   r2
43:     Which is: 32-byte object <10-54 8F-02 00-00 00-00 80-00 A0-D3 9A-7F 00-00 89-8C 01-00 00-00 00-00 E1-02 00-00 00-00 00-00>
43:   gold
43:     Which is: 32-byte object <10-54 8F-02 00-00 00-00 80-00 A0-D3 9A-7F 00-00 89-8C 01-00 00-00 00-00 9C-00 00-00 00-00 00-00>
43: is_sorted_until: stride2_view, all overloads 
43: is_sorted_until: stride3_view, all overloads 
43: [  FAILED  ] std_algorithms_sorting_ops_test.is_sorted_until (74 ms)
```

PR #5878, opened 2023-02-13
Remove Aligned memory trait when creating subviews and deprecate subview overload

### LINKED ISSUE #5862: Kokkos::subview() not working with aligned Views on HIP platforms
We have discovered that `Kokkos::subview()` fails with a device-side assert on HIP platforms if the source `Kokkos::View` has the `Kokkos::Aligned` memory trait.  Here is a reproducer for the error:

```
using data_type = double*[5];
using layout = Kokkos::LayoutLeft;
using memory_space = Kokkos::DefaultExecutionSpace::memory_space;
using memory_traits = Kokkos::MemoryTraits<Kokkos::Restrict | Kokkos::Aligned>;

using CustomView = Kokkos::View<data_type, layout, memory_space, memory_traits>;

TEST(defaultdevicetype, development_test) {

  CustomView customView("customView", 10);

  Kokkos::parallel_for("reproducer", customView.extent(0),
    KOKKOS_LAMBDA(int index) {
      const auto customSubView = Kokkos::subview(customView, index, Kokkos::ALL());
    });
}
```
and this is the error when running on a HIP platform (Caraway and Crusher, specifically):
```
:0: : Device-side assertion `Assigning NonAligned View or Pointer to Kokkos::View with Aligned attribute' failed.
:0: : Device-side assertion `Assigning NonAligned View or Pointer to Kokkos::View with Aligned attribute' failed.
:0: : Device-side assertion `Assigning NonAligned View or Pointer to Kokkos::View with Aligned attribute' failed.
:0: : Device-side assertion `Assigning NonAligned View or Pointer to Kokkos::View with Aligned attribute' failed.
:0: : Device-side assertion `Assigning NonAligned View or Pointer to Kokkos::View with Aligned attribute' failed.
:0: : Device-side assertion `Assigning NonAligned View or Pointer to Kokkos::View with Aligned attribute' failed.
:0: : Device-side assertion `Assigning NonAligned View or Pointer to Kokkos::View with Aligned attribute' failed.
:0: : Device-side assertion `Assigning NonAligned View or Pointer to Kokkos::View with Aligned attribute' failed.
:0:rocdevice.cpp            :2652: 3973238735976 us: 502330: [tid:0x7fc0e05a7700] Device::callbackQueue aborting with error : HSA_STATUS_ERROR_EXCEPTION: An HSAIL operation resulted in a hardware exception. code: 0x1016
Aborted (core dumped)
```
This is where the problem occurs:
https://github.com/kokkos/kokkos/blob/a45165d0954ba03fd7ac59e15bfb39a17d0e5dea/core/src/impl/Kokkos_ViewMapping.hpp#L2710-L2721
and it appears that we simply need to add a HIP guard next to the CUDA guard.
```
#ifdef KOKKOS_ENABLE_HIP
                      && (!(std::is_same<typename Traits::memory_space,
                                         Kokkos::Experimental::HIPSpace>::value ||
                            std::is_same<typename Traits::memory_space,
                                         Kokkos::Experimental::HIPManagedSpace>::value))
#endif
```
There appears to be three different versions of the `ViewDataHandle` class in this file that need the same treatment.

masterleinad:
Fixes #5862. It's not safe to assume that subviews of views with the Aligned memory trait are also aligned. Hence, just remove that attribute when creating subviews.

The implementation required to have access to the template argument of `MemoryTraits` which is why I added a static `impl_value` variable. 

This pull request also deprecates the subview overload that takes memory traits template parameters.

### DIFF
--- a/core/src/Kokkos_View.hpp
+++ b/core/src/Kokkos_View.hpp
@@ -1788,8 +1788,10 @@ KOKKOS_INLINE_FUNCTION auto subview(const View<D, P...>& src, Args... args) {
       Args...>::type(src, args...);
 }
 
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
 template <class MemoryTraits, class D, class... P, class... Args>
-KOKKOS_INLINE_FUNCTION auto subview(const View<D, P...>& src, Args... args) {
+KOKKOS_DEPRECATED KOKKOS_INLINE_FUNCTION auto subview(const View<D, P...>& src,
+                                                      Args... args) {
   static_assert(View<D, P...>::rank == sizeof...(Args),
                 "subview requires one argument for each source View rank");
   static_assert(Kokkos::is_memory_traits<MemoryTraits>::value);
@@ -1800,6 +1802,7 @@ KOKKOS_INLINE_FUNCTION auto subview(const View<D, P...>& src, Args... args) {
       typename Impl::RemoveAlignedMemoryTrait<D, P..., MemoryTraits>::type,
       Args...>::type(src, args...);
 }
+#endif
 
 template <class V, class... Args>
 using Subview = decltype(subview(std::declval<V>(), std::declval<Args>()...));
--- a/core/src/Kokkos_View.hpp
+++ b/core/src/Kokkos_View.hpp
@@ -1766,7 +1766,7 @@ struct RemoveAlignedMemoryTrait {
       typename Kokkos::Impl::type_list_remove_first<memory_traits,
                                                     type_list_in>::type;
   using new_memory_traits =
-      Kokkos::MemoryTraits<memory_traits::value & ~Kokkos::Aligned>;
+      Kokkos::MemoryTraits<memory_traits::impl_value & ~Kokkos::Aligned>;
   using new_type_list = typename Kokkos::Impl::concat_type_list<
       type_list_in_wo_memory_traits,
       Kokkos::Impl::type_list<new_memory_traits>>::type;
--- a/core/src/Kokkos_View.hpp
+++ b/core/src/Kokkos_View.hpp
@@ -34,6 +34,7 @@ static_assert(false,
 #include <View/Hooks/Kokkos_ViewHooks.hpp>
 
 #include <impl/Kokkos_Tools.hpp>
+#include <impl/Kokkos_Utilities.hpp>
 
 #ifdef KOKKOS_ENABLE_IMPL_MDSPAN
 #include <View/MDSpan/Kokkos_MDSpan_Extents.hpp>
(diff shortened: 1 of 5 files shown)


core/src/Kokkos_View.hpp
```diff
@@ -1739,45 +1740,64 @@ void apply_to_view_of_static_rank(Function&& f, View<Args...> a) {
...
+  using new_memory_traits =
+      Kokkos::MemoryTraits<memory_traits::value & ~Kokkos::Aligned>;
+  using new_type_list = typename Kokkos::Impl::concat_type_list<
+      type_list_in_wo_memory_traits,
+      Kokkos::Impl::type_list<new_memory_traits>>::type;
+
+ public:
+  using type = typename TypeListToViewTraits<D, new_type_list>::type;
+};
+}  // namespace Impl
```
dalg24, line 1771: This feels more like a hack than a long term solution.
Please do define `Subview` and use it in `subview`

core/src/Kokkos_View.hpp
```diff
@@ -1739,45 +1740,64 @@ void apply_to_view_of_static_rank(Function&& f, View<Args...> a) {
...
+      Args...>::type(src, args...);
 }
 
 template <class MemoryTraits, class D, class... P, class... Args>
-KOKKOS_INLINE_FUNCTION typename Kokkos::Impl::ViewMapping<
-    void /* deduce subview type from source view traits */
-    ,
-    ViewTraits<D, P...>, Args...>::template apply<MemoryTraits>::type
-subview(const View<D, P...>& src, Args... args) {
+KOKKOS_INLINE_FUNCTION auto subview(const View<D, P...>& src, Args... args) {
```
dalg24, line 1786: I did some archaeology and this overload exists since the dawn of time.
It is only used in the tests
```
containers/unit_tests/TestOffsetView.hpp:700:  test_offsetview_subview<int, TEST_EXECSPACE>();
core/unit_test/TestViewMapping_subview.hpp:131:    auto Ad = Kokkos::subview<Kokkos::MemoryUnmanaged>(
core/unit_test/TestViewSubview.hpp:2144:    auto sv = Kokkos::subview<memory_traits_type>(v, range);
```
and I did not find any usage in Trilinos.
It is never mentioned in our documentation.  If we decide to deprecate/remove (which I think we should) it will make things a bit more simple.

core/unit_test/TestViewSubview.hpp
```diff
@@ -2148,6 +2156,22 @@ inline void test_subview_memory_traits_construction() {
   TestSubviewMemoryTraitsConstruction<13>()();
   TestSubviewMemoryTraitsConstruction<14>()();
   TestSubviewMemoryTraitsConstruction<15>()();
+  TestSubviewMemoryTraitsConstruction<16>()();
```
dalg24, line 2159: Implementing in the operator() instead of the constructor was just dumb (not asking you to change, just venting when I l see the syntax)

dalg24 (commented):
I agree with the general direction but I want to discuss the implementation.

core/src/Kokkos_View.hpp
masterleinad: You don't like
```
template <class V, class... Args>
using Subview = decltype(subview(std::declval<V>(), std::declval<Args>()...));
```
?

dalg24: No

masterleinad: Can you please elaborate on why duplicating the type (or declaring the `Subview` alias first) is preferable over deducing the `Subview` alias from the `subview` return value? The latter appears more natural to me since `Subview` is the type that a call to `subview` would return.

core/unit_test/TestViewSubview.hpp
```diff
@@ -2110,15 +2110,30 @@ void test_unmanaged_subview_reset() {
...
   void operator()() const noexcept {
-    using view_type          = Kokkos::View<double*, Kokkos::HostSpace>;
-    using size_type          = view_type::size_type;
     using memory_traits_type = Kokkos::MemoryTraits<MTF>;
+    using view_type =
+        Kokkos::View<double*, Kokkos::HostSpace, memory_traits_type>;
+    using size_type = typename view_type::size_type;
 
-    view_type v("v", 7);
+    Kokkos::View<double*, Kokkos::HostSpace> v_original("v", 7);
```
dalg24, line 2118: Why did you introduce `v_original`

masterleinad: I can't use `Unmanaged` with a managed `View`.

dalg24: Please add a comment.  This is a bit confusing.  All these views are "runtime unmanaged"...

dalg24 (approved):
Make sure you edit the PR title and description to capture the deprecation

masterleinad:
> Make sure you edit the PR title and description to capture the deprecation

Done.

PhilMiller (approved):

PhilMiller:
HIP tests are all failing with this:
```
38: terminate called after throwing an instance of 'std::runtime_error'
38:   what():  Desul::Error: init_lock_arrays_hip: post malloc error(hipErrorOutOfMemory): hipErrorOutOfMemory
```

SYCL had this failure:
```
38: [ RUN      ] sycl.dynamic_view
38: Copy from on-device DynamicView to on-host View failed:
38: deep_copy given views that would require a temporary allocation
38: Copy from on-host View to on-device DynamicView failed:
38: deep_copy given views that would require a temporary allocation
38: Copy from on-device DynamicView to on-host View failed:
38: deep_copy given views that would require a temporary allocation
38: Copy from on-host View to on-device DynamicView failed:
38: deep_copy given views that would require a temporary allocation
38: Copy from on-device DynamicView to on-host View failed:
38: deep_copy given views that would require a temporary allocation
38: Copy from on-host View to on-device DynamicView failed:
38: deep_copy given views that would require a temporary allocation
38: Copy from on-device DynamicView to on-host View failed:
38: deep_copy given views that would require a temporary allocation
38: Copy from on-host View to on-device DynamicView failed:
38: deep_copy given views that would require a temporary allocation
38: /var/jenkins/workspace/Kokkos/containers/unit_tests/TestDynamicView.hpp:269: Failure
38: Expected equality of these values:
38:   new_result_sum
38:     Which is: 6.97873e+07
38:   (value_type)(da_size * (da_size - 1) / 2)
38:     Which is: 7.8745e+07
38: Copy from on-device DynamicView to on-host View failed:
38: deep_copy given views that would require a temporary allocation
38: Copy from on-host View to on-device DynamicView failed:
38: deep_copy given views that would require a temporary allocation
38: Copy from on-device DynamicView to on-host View failed:
38: deep_copy given views that would require a temporary allocation
38: Copy from on-host View to on-device DynamicView failed:
38: deep_copy given views that would require a temporary allocation
38: Copy from on-device DynamicView to on-host View failed:
38: deep_copy given views that would require a temporary allocation
38: Copy from on-host View to on-device DynamicView failed:
38: deep_copy given views that would require a temporary allocation
38: Copy from on-device DynamicView to on-host View failed:
38: deep_copy given views that would require a temporary allocation
38: Copy from on-host View to on-device DynamicView failed:
38: deep_copy given views that would require a temporary allocation
38: Copy from on-device DynamicView to on-host View failed:
38: deep_copy given views that would require a temporary allocation
38: Copy from on-host View to on-device DynamicView failed:
38: deep_copy given views that would require a temporary allocation
38: [  FAILED  ] sycl.dynamic_view (36 ms)
```

It's vaguely plausible that `subview` is implicated in the implementation of `deep_copy` involving `DynamicView`. That looks like something that at least should be deterministic.

masterleinad:
```
Copy from on-device DynamicView to on-host View failed:
deep_copy given views that would require a temporary allocation
```
results from https://github.com/kokkos/kokkos/blob/15989dd3a1edb72bbc178fc74d081502e32dd3d6/containers/unit_tests/TestDynamicView.hpp#L272-L290 and is present on all GPU backends. I can't reproduce the test failure on Intel GPUs. Let's just try to retest and if the test failure persists, disable the test for `SYCL+Cuda`.

Retest this please.

masterleinad:
Retest this please.

crtrott:
Failed right away again. Probably a machine issue? But I am kinda hesitant merging subview stuff if we get memory errors (even though I don't see anything wrong with the PR).

dalg24:
HIP failure is clearly an issue with the machine and I am comfortable ignoring.  SYCL build passed.

Issue #2233, opened 2019-08-13
ScratchMemorySpace should subsume the MemorySpace concept


dhollman:
There's no reason to have two separate concept hierarchies for `MemorySpace` and `ScratchMemorySpace`, but since right now `ScratchMemorySpace` instances in Kokkos don't implement `.name()`, `.allocate()`, and `.deallocate()` (as well as things like `Impl::MemorySpaceAccess` specializations), users have to think of them as a completely separate thing.

If it's not possible to meet all of the requirements of `MemorySpace` for the scratch versions, we should call it something else and remove the `memory_space` tag type alias.

crtrott:
deallocate doesn't work. But it could be a no-op?

dhollman:
Yeah, deallocate should be a no-op, just like any bump allocator, for instance.  But it should be present.  There's no reason for the user to have to think about a completely separate generic entity just because something that you think of as cheap anyway compiles into a no-op.

crtrott:
The question is whether a `no-op` deallocate inside of ScratchMemorySpace confuses people more than the slight non-compliance with a concept. I imagine all the bug reports about "deallocate" not actually deallocating ...

mhoemmen:
> I imagine all the bug reports about "deallocate" not actually deallocating ...

On the other hand, it's fun to have a standard reply that says "read the documentation" ;-)

ajpowelsnl:
@dalg24 , @masterleinad -- are we finished subsuming?  Can we close this issue?

masterleinad:
No, the issue is not addressed. We can, of course, discuss if we ever want to do anything about it or not.

ajpowelsnl:
@ajpowelsnl -- put on items to discuss

PR #5879, opened 2023-02-14
Allow specifying scratch level for ScratchSpace at compile-time

masterleinad:
The oneAPI compiler emits warnings whenever scratch memory is used since we don't expose what the correct address space is. For level 0, we essentially want `sycl::local_ptr<>` and for level 1 `sycl::global_ptr<>`/`sycl::device_ptr<>`. Since these are different types, it's difficult for the compiler to understand what is requested and we end up with more instructions than necessary. This pull request introduces an additional interface for using scratch memory inside kernels that allows to specify the scratch level at compile-time, e.g.
```
teamThread.team_scratch(0)
```
can become
```
teamThread.template team_scratch<0>()
```

Internally, `ScratchSpace` now wraps two `ScratchSpaceBase` classes that can hold pointers for the respective address spaces. For all backends apart from `SYCL` the pointer types are just `char*` and for SYCL we have `sycl::local_ptr<char>` and `sycl::device_ptr<char>`. Even though the interface still uses `void*` pointers, this seems to be enough for the compiler to figure out the correct address space.

All unit tests have been changed to the new interface although that isn't really necessary but merely shows that the interface works as expected.

https://github.com/kokkos/kokkos/issues/2233 already raised the question if the `ScratchSpace` should be a proper `Kokkos` memory space or not which is an open question here as well.


masterleinad:
I'd like to discuss if we an allow an updated interface to allow users to specify the scratch space level at compile-time.
Without this, the `oneAPI` compiler doesn't seem to be able to infer the correct address space (which could either be `local` or `global`/`device`).

stanmoore1:
This could potentially help LAMMPS according to our benchmarking.

masterleinad:
I pushed an alternative implementation that doesn't require duplicating the `ScratchMemorySpace` implementation for `SYCL` but rather allows specifying the pointer types used in `ScratchMemorySpace`.

core/src/Kokkos_View.hpp
```diff
@@ -1695,14 +1699,50 @@ class View : public ViewTraits<DataType, Properties...> {
...
+                 static_cast<pointer_type>(arg_space.template get_shmem_aligned(
+                     map_type::memory_span(arg_layout),
+                     scratch_value_alignment))),
+             arg_layout) {}
+
   explicit KOKKOS_INLINE_FUNCTION View(
       const typename traits::execution_space::scratch_memory_space& arg_space,
       const typename traits::array_layout& arg_layout)
-      : View(Impl::ViewCtorProp<pointer_type>(reinterpret_cast<pointer_type>(
+      : View(Impl::ViewCtorProp<pointer_type>(static_cast<pointer_type>(
```
dalg24, line 1716: The static_cast here and below is a good point but I almost feel like it should be proposed as its own PR

core/unit_test/TestTeam.hpp
```diff
@@ -700,11 +700,11 @@ struct ScratchTeamFunctor {
   KOKKOS_INLINE_FUNCTION
   void operator()(const typename policy_type::member_type &ind,
                   value_type &update) const {
-    const shared_int_array_type scratch_ptr(ind.team_scratch(1),
+    const shared_int_array_type scratch_ptr(ind.template team_scratch<1>(),
```
dalg24, line 703: What's the deal with all the changes in the tests?
You sprinkled changes to showcase the new syntax or are these changes required?
Same questions everywhere.

masterleinad: Just for showcasing/testing the changes. There are no changes in the tests required but I would add some once I get the OK on the design otherwise.

dalg24: Understood

core/src/Kokkos_ScratchSpace.hpp
```diff
@@ -31,103 +31,71 @@ static_assert(false,
...
+  mutable PointerType m_iter = nullptr;
+  PointerType m_end          = nullptr;
 
-  mutable int m_multiplier    = 0;
-  mutable int m_offset        = 0;
-  mutable int m_default_level = 0;
-
-#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
-  constexpr static int DEFAULT_ALIGNMENT_MASK = ALIGN - 1;
-#endif
```
masterleinad, line 59: Looks like this and ...

core/src/Kokkos_ScratchSpace.hpp
```diff
@@ -31,103 +31,71 @@ static_assert(false,
...
-  static constexpr const char* name() { return "ScratchMemorySpace"; }
-
-#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
-  // This function is unused
-  template <typename IntType>
-  KOKKOS_DEPRECATED KOKKOS_INLINE_FUNCTION static constexpr IntType align(
-      const IntType& size) {
-    return (size + DEFAULT_ALIGNMENT_MASK) & ~DEFAULT_ALIGNMENT_MASK;
-  }
-#endif
```
masterleinad, line 80: ... this has been lost over the course of changes to this pull request. I am happy add them back in if preferred.

core/src/Kokkos_ScratchSpace.hpp
```diff
@@ -139,30 +107,102 @@ class ScratchMemorySpace {
...
+};
+
+template <typename ExecSpace, typename PointerTypeL0 = char*,
+          typename PointerTypeL1 = char*>
+class ScratchMemorySpace {
+  ScratchMemorySpaceBase<ExecSpace, PointerTypeL0> m_scratch_L0;
+  ScratchMemorySpaceBase<ExecSpace, PointerTypeL1> m_scratch_L1;
+  mutable int m_default_level = 0;
+
+ public:
```
masterleinad, line 136:
```suggestion
```

core/src/SYCL/Kokkos_SYCL_Team.hpp
```diff
@@ -348,7 +358,9 @@ class SYCLTeamMember {
                  const int arg_league_size)
       : m_team_reduce(shared),
         m_team_shared(static_cast<sycl::local_ptr<char>>(shared) + shared_begin,
-                      shared_size, scratch_level_1_ptr, scratch_level_1_size),
+                      shared_size,
+                      static_cast<sycl::device_ptr<char>>(scratch_level_1_ptr),
```
masterleinad, line 362:
```suggestion
                      static_cast<sycl::global_ptr<char>>(scratch_level_1_ptr),
```

core/src/SYCL/Kokkos_SYCL_Space.hpp
```diff
@@ -301,10 +301,11 @@ struct MemorySpaceAccess<Kokkos::Experimental::SYCLHostUSMSpace,
   enum : bool { deepcopy = true };
 };
 
-template <>
+template <typename LocalPointerType, typename GlobalPointerType>
 struct MemorySpaceAccess<
     Kokkos::Experimental::SYCLDeviceUSMSpace,
-    Kokkos::ScratchMemorySpace<Kokkos::Experimental::SYCL>> {
+    Kokkos::ScratchMemorySpace<Kokkos::Experimental::SYCL, LocalPointerType,
+                               GlobalPointerType>> {
```
masterleinad, line 308: Without this fix, I was getting segmentation faults when the memory space availability check triggered instead of a message through `Kokkos::abort`.

masterleinad:
A more complete version for properly supporting the correct address space can be found at https://github.com/masterleinad/kokkos/tree/specialize_scratch_space_sycl_and_view. 
That branch adds the required `ViewDataHandle` specializations to actually obtain the correct pointer when accessing the memory, updates the requirements on the TeamHandle concept, and introduces more aliases for the new scratch space memory spaces. I'm happy to discuss what of that we already would want in this pull request.

masterleinad:
Talking with @crtrott this morning, we were discussing
- adding an integer level template parameter to `ScratchSpace`, possibly instead of defining `ScratchSpaceBase`
- making these scratch spaces fulfill the memory spaces concept
- adding more aliases to the execution space and team handle classes. Names are up for discussion but `level_0`, `level_1` might not be ideal, maybe `scratch_memory_space_local` and `sratch_memory_space_global` (or `scratch_memory_space_dram`) are better.

ajpowelsnl:
To discuss on Wednesday

core/src/Kokkos_ScratchSpace.hpp
```diff
@@ -31,103 +31,71 @@ static_assert(false,
 
 namespace Kokkos {
 
-/** \brief  Scratch memory space associated with an execution space.
- *
- */
-template <class ExecSpace>
-class ScratchMemorySpace {
+template <typename ExecSpace, typename PointerType>
```
masterleinad, line 34: We could probably do without a `PointerType` template parameter. It does really only matter what we are doing when accessing the data.

core/src/SYCL/Kokkos_SYCL_View.hpp
```diff
@@ -0,0 +1,121 @@
...
+  }
+};
+
+template <class Traits>
+struct ViewDataHandle<
+    Traits,
+    std::enable_if_t<std::is_void<typename Traits::specialize>::value &&
+                     !Traits::memory_traits::is_aligned &&
+                     !Traits::memory_traits::is_restrict &&
+                     !Traits::memory_traits::is_atomic &&
```
masterleinad, line 85: revisit this

core/src/View/Kokkos_ViewMapping.hpp
```diff
@@ -3473,10 +3473,10 @@ struct RuntimeCheckViewMemoryAccessViolation<MemorySpace, AccessSpace, false> {
...
 KOKKOS_FUNCTION void runtime_check_memory_access_violation(
     char const* const msg, Track const& track, Map const& map, Ignore...) {
-  KOKKOS_IF_ON_HOST(
-      ((void)RuntimeCheckViewMemoryAccessViolation<MemorySpace,
-                                                   DefaultHostExecutionSpace>(
-           msg, track, map);))
+  // KOKKOS_IF_ON_HOST(
+  //     ((void)RuntimeCheckViewMemoryAccessViolation<MemorySpace,
+  //                                                  DefaultHostExecutionSpace>(
+  //          msg, track, map);))
```
masterleinad, line 3479:
```suggestion
  KOKKOS_IF_ON_HOST(
      ((void)RuntimeCheckViewMemoryAccessViolation<MemorySpace,
                                                   DefaultHostExecutionSpace>(
           msg, track, map);))
```

core/unit_test/TestTeam.hpp
masterleinad: I think we should actually be explicit about the level whenever we know it at compile-time.

core/src/HIP/Kokkos_HIP.hpp
```diff
@@ -46,6 +46,8 @@ class HIP {
   using size_type    = HIPSpace::size_type;
 
   using scratch_memory_space = ScratchMemorySpace<HIP>;
+  using scratch_memory_space_l0 = ScratchMemorySpaceBase<HIP, L0Tag>;
+    using scratch_memory_space_l1 = ScratchMemorySpaceBase<HIP, L1Tag>;
```
JBludau, line 50:
```suggestion
  using scratch_memory_space_l0 = ScratchMemorySpaceBase<HIP, L0Tag>;
  using scratch_memory_space_l1 = ScratchMemorySpaceBase<HIP, L1Tag>;
```

masterleinad: Sure. I'm more concerned about discussing the interface/design at this point than fixing it all up.

JBludau: just a drive-by as I am trying to get a feel for it :-)

core/src/HIP/Kokkos_HIP_Team.hpp
```diff
@@ -79,8 +81,14 @@ class HIPTeamMember {
 
  public:
   KOKKOS_INLINE_FUNCTION
-  const execution_space::scratch_memory_space& team_shmem() const {
-    return m_team_shared.set_team_thread_mode(0, 1, 0);
+  const auto& team_shmem() const {
```
JBludau, line 84:
```suggestion
  const scratch_memory_space_l0& team_shmem() const {
```

is there a reason why this deviates from the other backends?

core/src/Kokkos_ScratchSpace.hpp
```diff
@@ -31,103 +31,77 @@ static_assert(false,
...
 namespace Kokkos {
 
-/** \brief  Scratch memory space associated with an execution space.
- *
- */
-template <class ExecSpace>
-class ScratchMemorySpace {
+struct L0Tag {};
+
+struct L1Tag {};
```
JBludau, line 36: maybe we should make the name hint that they are used for `Scratch`? like `ScratchL0Tag`?

core/src/Kokkos_ScratchSpace.hpp
```diff
@@ -139,30 +113,103 @@ class ScratchMemorySpace {
...
   }
+
+  template <int Level>
+  KOKKOS_INLINE_FUNCTION const auto& set_team_thread_mode(
+      const int& multiplier, const int& offset) const {
+    if constexpr (Level == 0) {
+      return m_scratch_L0.set_team_thread_mode(multiplier, offset);
+    } else {
+      static_assert(Level == 1);
+      return m_scratch_L1.set_team_thread_mode(multiplier, offset);
```
JBludau, line 199:
```suggestion
    static_assert((Level == 0 or Level==1), "Level of ScratchMemory must be 0 or 1" );
    if constexpr (Level == 0) {
      return m_scratch_L0.set_team_thread_mode(multiplier, offset);
    } else {
      return m_scratch_L1.set_team_thread_mode(multiplier, offset);
```

core/src/Kokkos_UniqueToken.hpp
```diff
@@ -126,7 +126,7 @@ class AcquireTeamUniqueToken {
   using size_type        = typename token_type::size_type;
   using team_member_type = typename TeamPolicy::member_type;
   using scratch_view =
-      Kokkos::View<size_type, typename exec_space::scratch_memory_space,
+      Kokkos::View<size_type, typename exec_space::scratch_memory_space_l0,
```
JBludau, line 129: Would we want to give users the option to put the Token in L1 and use the L0 for something else?

masterleinad: That would be a different discussion. 🙂 This pull request doesn't change where scratch memory is allocated

core/src/HIP/Kokkos_HIP_Team.hpp
masterleinad: No, not really. The real question is if this should return the generic space instead (if we decide that we can assign from the specialized spaces to the generic one) or if changing the type is fine.

core/src/Kokkos_ScratchSpace.hpp
masterleinad: Sure, those names a re up for discussion. I would also be fine with `CacheScratchTag` and `GlobalScratchTag` or similar. Ultimately it should match the name for the alias used for the specialized spaces.
`level=0` and` level=1` is just what we are currently using. 
Keeping integers for the public interfaces for requesting a space is convenient for transitioning but we could use these tags there as well.

core/src/OpenMP/Kokkos_OpenMP.hpp
```diff
@@ -60,10 +60,12 @@ class OpenMP {
...
-  using device_type          = Kokkos::Device<execution_space, memory_space>;
-  using array_layout         = LayoutRight;
-  using size_type            = memory_space::size_type;
-  using scratch_memory_space = ScratchMemorySpace<OpenMP>;
+  using device_type             = Kokkos::Device<execution_space, memory_space>;
+  using array_layout            = LayoutRight;
+  using size_type               = memory_space::size_type;
+  using scratch_memory_space    = ScratchMemorySpace<OpenMP>;
+  using scratch_memory_space_l0 = ScratchMemorySpace<OpenMP>;
+  using scratch_memory_space_l1 = ScratchMemorySpace<OpenMP>;
```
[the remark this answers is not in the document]
masterleinad: The problem with the current implementation is that only one allocation is used. That's something that we could address separately.

masterleinad:
Needs more discussion.

PR #5880, opened 2023-02-16
Deprecate `View::Rank` "uppercase R" in favor of `View::rank` "lowercase r"

dalg24:
Moving the easier part of #5870 into its own PR to facilitate review.
Rational for the change: duplicate facility that was never mentioned in the documentation.


dalg24:
Maybe I should have actually tried compiling before submitting :/
Turns out it cannot be separated like I thought it could...

PR #5884, opened 2023-02-16
Add AMD ROCm support to hpcbind

rbberger:
- takes care of issue on systems where `nvidia-smi` exists but doesn't actually work
- allow hpcbind to work in a similar way on AMD ROCm multi-gpu systems

### DIFF
--- a/bin/hpcbind
+++ b/bin/hpcbind
@@ -36,8 +36,14 @@ fi
 ################################################################################
 declare -i HPCBIND_HAS_NVIDIA=0
 type nvidia-smi >/dev/null 2>&1
-HPCBIND_HAS_NVIDIA=$((!$?))
+HPCBIND_HAS_NVIDIA=$((! $?))
 
+################################################################################
+# Check if rocm-smi exist
+################################################################################
+declare -i HPCBIND_HAS_AMD=0
+type rocm-smi >/dev/null 2>&1
+HPCBIND_HAS_AMD=$((! $?))
 
 ################################################################################
 # Get visible gpu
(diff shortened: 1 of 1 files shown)


dalg24 (changes requested):
Please rebase your commit onto `develop` and update the destination branch.

crtrott (approved):
Looks ok to me but I haven't tried this thing at all. @rbberger would you be willing to write documentation for this - none of us are really using this script :-D

Rombur:
In the current version of the script we have some specific output for CUDA
https://github.com/kokkos/kokkos/blob/5ad609661e570ba6aa7716a26a91cb67d559f8a2/bin/hpcbind#L581
https://github.com/kokkos/kokkos/blob/5ad609661e570ba6aa7716a26a91cb67d559f8a2/bin/hpcbind#L605

Can you add the same thing for ROCm.

rbberger:
> Looks ok to me but I haven't tried this thing at all. @rbberger would you be willing to write documentation for this - none of us are really using this script :-D

:grin: I actually learned about it by accident because someone else was manually doing something similar to this and in a different context someone then pointed out that this existed. Turned out simple enough to then adopt for AMD. @crtrott documentation in what form?

crtrott:
Just like a markdown thingy explaining how this works, what its options are etc. even just as a comment here on github. We can convert that into sphinx and integrate into the kokkos documentation

Rombur (approved):

rbberger:
@crtrott this got merged before I added the docs. What I would have done is below, which is basically just a markdown thingy for the `--help` output.

# hpcbind

Utility wrapper script to set the process mask, OMP environment variables and CUDA/ROCm environment variables of an arbitrary command to sane values if possible. Uses `hwloc` and `nvidia-smi`/`rocm-smi` if available.  Will preserve the current process binding, so it is safe to use with a queuing system or `mpiexec`.

    hpcbind <options> -- command <command-options>

## Options

| Option | Description |
|--------|-------------|
|  `--no-hwloc-bind` | Disable binding |
| `--proc-bind=<LOC>` |     Set the initial process mask for the script.<br> LOC can be any valid location argument for `hwloc-calc`  Default: all |
|  `--whole-system`  | `hpcbind` will ignore the its parent process binding |
|  `--distribute=N`   |  Distribute the current cpuset into N partitions |
| `--distribute-partition=I` | Use the i'th partition (zero based) |
| `--visible-gpus=<L>` |  Comma separated list of gpu ids<br>Default: `CUDA_VISIBLE_DEVICES` / `ROCR_VISIBLE_DEVICES` or all gpus in sequential order |
|  `--ignore-queue` |   Ignore queue job id when choosing visible GPU and partition |
|  `--no-gpu-mapping` |  Do not set `CUDA_VISIBLE_DEVICES` / `ROCR_VISIBLE_DEVICES` |
|  `--openmp=M.m`       |   Set env variables for the given OpenMP version <br> Default: 4.0 |
|  `--openmp-ratio=N/D`  |  Ratio of the cpuset to use for OpenMP<br>Default: 1 |
| `--openmp-places=<Op>` |  Op=`threads`\|`cores`\|`sockets`.<br> Default: `threads` |
| `--openmp-num-threads=N` | Override logic for selecting `OMP_NUM_THREADS` |
| `--openmp-proc-bind=<OP>` | Override logic for selecting `OMP_PROC_BIND` |
|  `--openmp-nested`     |  Set `OMP_NESTED` to true |
|  `--no-openmp-proc-bind` | Set `OMP_PROC_BIND` to `false` and unset `OMP_PLACES` |
| `--output-prefix=<P>` |  Save the output to files of the form `P.hpcbind.N`, `P.stdout.N` and `P.stderr.N` where `P` is the prefix and `N` is the rank (no spaces) |
| `--output-mode=<Op>` |    How console output should be handled. Options are `all`, `rank0`, and `none`.  Default: `rank0` |
| `--lstopo`           |   Show bindings in lstopo |
| `--save-topology=<Xml>` | Save the topology to the given xml file |
| `--load-topology=<Xml>` |  Load a previously saved topology from an xml file |
| `-v\|--verbose`   |       Print bindings and relevant environment variables |
| `-h\|--help` |             Show this message |

## Sample Usage:

### Split the current process cpuset into 4 and use the 3rd partition

    hpcbind --distribute=4 --distribute-partition=2 -v -- command ...

### Launch 16 jobs over 4 nodes with 4 jobs per node using only the even PUs and save the output to rank specific files

    mpiexec -N 16 -npernode 4 hpcbind --whole-system --proc-bind=pu:even \
                                    --distribute=4 -v --output-prefix=output  -- command ...

### Bind the process to all even cores

    hpcbind --proc-bind=core:even -v -- command ...

### Bind the the even cores of socket 0 and the odd cores of socket 1

    hpcbind --proc-bind='socket:0.core:even socket:1.core:odd' -v -- command ...

### Skip GPU 0 when mapping visible devices

    hpcbind --distribute=4 --distribute-partition=0 --visible-gpus=1,2 -v -- command ...

### Display the current bindings

    hpcbind --proc-bind=numa:0 -- command
    
### Display the current bindings using lstopo

    hpcbind --proc-bind=numa:0.core:odd --lstopo

PR #5889, opened 2023-02-17
OpenMPTarget: Enable Cray compiler for the OpenMPTarget backend.

rgayatri23:
The PR does the following : 
1. Adds a macro for Clang based Cray compiler as `KOKKOS_COMPILER_CRAY_CLANG` 
2. Enables appropriate flags for OpenMPTarget backend with Cray compiler in Makefile.kokkos and in cmake/kokkos_arch.cmake.
3. Disable unit tests that do not compile or fail at runtime with the OpenMPTarget backend and Cray compiler.


core/src/OpenMPTarget/Kokkos_OpenMPTarget_Instance.cpp
```diff
@@ -93,13 +93,16 @@ void OpenMPTargetInternal::impl_initialize() {
 
   // FIXME_OPENMPTARGET:  Only fix the number of teams for NVIDIA architectures
   // from Pascal and upwards.
+  // Cray compiler did not yet implement omp_set_num_teams.
```
crtrott, line 96: *sigh*

core/unit_test/CMakeLists.txt
```diff
@@ -353,6 +353,12 @@ if(Kokkos_ENABLE_OPENMPTARGET)
     ${CMAKE_CURRENT_BINARY_DIR}/openmptarget/TestOpenMPTarget_Other.cpp
     ${CMAKE_CURRENT_BINARY_DIR}/openmptarget/TestOpenMPTarget_TeamReductionScan.cpp
     ${CMAKE_CURRENT_BINARY_DIR}/openmptarget/TestOpenMPTarget_WorkGraph.cpp
+    # The following tests fail at compile time when the OpenMPTarget backend is enabled with the Cray compiler.
```
crtrott, line 356: Add a FIXME comment also this is all kinds of atomic stuff - this could be a real issue. As if compare/exchange doesn't work properly.

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Instance.cpp
crtrott: Add a FIXME for Cray

core/unit_test/TestTeamMDRange.hpp
```diff
@@ -1904,6 +1904,12 @@ TEST(TEST_CATEGORY, ThreadVectorMDRangeParallelReduce) {
     GTEST_SKIP() << "skipping because of bug in group_barrier implementation";
 #endif
 
+#if defined(KOKKOS_ENABLE_OPENMPTARGET) && defined(KOKKOS_COMPILER_CRAYCLANG)
+  if (std::is_same_v<TEST_EXECSPACE, Kokkos::Experimental::OpenMPTarget>)
+    GTEST_SKIP() << "Cray compiler fails correctness at runtime with the "
+                    "OpenMPTarget backend.";
+#endif
```
crtrott, line 1911: Add a FIXME comment

core/unit_test/TestTeamMDRange.hpp
```diff
@@ -1937,6 +1943,12 @@ TEST(TEST_CATEGORY, TeamVectorMDRangeParallelReduce) {
...
```
crtrott, line 1950: Add a FIXME comment something with "Cray" in the same line too.

cmake/kokkos_arch.cmake
```diff
@@ -680,12 +680,18 @@ ENDIF()
 IF (KOKKOS_ENABLE_OPENMPTARGET)
   SET(CLANG_CUDA_ARCH ${KOKKOS_CUDA_ARCH_FLAG})
   IF (CLANG_CUDA_ARCH)
+      IF(KOKKOS_CLANG_IS_CRAY)
```
crtrott, line 683: fix identation.

crtrott (dismissed):
Some small changes needed (comments and indentation)

dalg24 (commented):
Can you comment about the choice to distinguish between the legacy and the Clang-based Cray compilers?

Also with Intel we do not define `KOKKOS_COMPILER_CLANG`
https://github.com/kokkos/kokkos/blob/948c6c69fc24879e507c8c86bee36490b4b11dbb/core/src/Kokkos_Macros.hpp#L151-L152
Did you mean to define both `KOKKOS_COMPILER_CRAYCLANG` and `KOKKOS_COMPILER_CLANG`?

rgayatri23:
> 

The `KOKKOS_COMPILER_CLANG` gets defined in the lines above since its already checking for `__clang__` . 
I added a comment to differentiate between classical cray and clang based cray compilers. Also added version number math for `KOKKOS_COMPILER_CRAYCLANG` similar to the clang compiler. 

With respect to Intel compiler, I assume the same `__clang__` will be defined which will define the respective macro (although I am not sure about it).

Makefile.kokkos
```diff
@@ -267,9 +267,13 @@ else
...
-  #KOKKOS_INTERNAL_OPENMPTARGET_FLAG := -DKOKKOS_BUG_WORKAROUND_IBM_CLANG_OMP45_VIEW_INIT -fopenmp-implicit-declare-target -fopenmp-targets=nvptx64-nvidia-cuda -fopenmp -fopenmp=libomp
-  KOKKOS_INTERNAL_OPENMPTARGET_FLAG := -DKOKKOS_WORKAROUND_OPENMPTARGET_CLANG -fopenmp -fopenmp=libomp -Wno-openmp-mapping
-  KOKKOS_INTERNAL_OPENMPTARGET_LIB := -lomptarget
+	# Cray compiler only needs fopenmp flag for target offload.
+  ifeq ($(KOKKOS_INTERNAL_COMPILER_CRAY_CLANG), 1)
+	  KOKKOS_INTERNAL_OPENMPTARGET_FLAG := -DKOKKOS_WORKAROUND_OPENMPTARGET_CLANG -fopenmp 
+  else
+	  KOKKOS_INTERNAL_OPENMPTARGET_FLAG := -DKOKKOS_WORKAROUND_OPENMPTARGET_CLANG -fopenmp -fopenmp=libomp -Wno-openmp-mapping
+       KOKKOS_INTERNAL_OPENMPTARGET_LIB := -lomptarget
+  endif
```
masterleinad, line 276:
```suggestion
  # Cray compiler only needs fopenmp flag for target offload.
  ifeq ($(KOKKOS_INTERNAL_COMPILER_CRAY_CLANG), 1)
    KOKKOS_INTERNAL_OPENMPTARGET_FLAG := -DKOKKOS_WORKAROUND_OPENMPTARGET_CLANG -fopenmp 
  else
    KOKKOS_INTERNAL_OPENMPTARGET_FLAG := -DKOKKOS_WORKAROUND_OPENMPTARGET_CLANG -fopenmp -fopenmp=libomp -Wno-openmp-mapping
    KOKKOS_INTERNAL_OPENMPTARGET_LIB := -lomptarget
  endif
```

dalg24:
> With respect to Intel compiler, I assume the same `__clang__` will be defined which will define the respective macro (although I am not sure about it).

Look at the code snippet I referenced.  We define `KOKKOS_COMPILER_CLANG` only if `KOKKOS_COMPILER_INTEL` is not defined.

Makefile.kokkos
```diff
@@ -972,12 +976,17 @@ ifeq ($(KOKKOS_INTERNAL_USE_CUDA_ARCH), 1)
...
+  ifeq ($(KOKKOS_INTERNAL_COMPILER_CRAY_CLANG), 0)
   ifeq ($(KOKKOS_INTERNAL_COMPILER_CLANG), 1)
     KOKKOS_INTERNAL_CUDA_ARCH_FLAG=-fopenmp-targets=nvptx64 -Xopenmp-target -march
   endif
+  endif
   KOKKOS_INTERNAL_USE_CUDA_ARCH = 1
 endif
 
+# Do not add this flag it its the cray compiler.
+ifeq ($(KOKKOS_INTERNAL_COMPILER_CRAY_CLANG), 0)
```
masterleinad, line 989: Please also indent accordingly.
So the Cray compiler doesn't take any architecture flags at all?!
Does this also apply if compiling for the `Cuda` backend?

dalg24:
> I added a comment to differentiate between classical cray and clang based cray compilers. 

I am asking about clarifying why we want to distinguish them in the first place.  I am not necessary suggesting we should not distinguish them.  I just observing we don't do it for Intel and I am wondering whether you thought about it and why you decided to do it this way.

core/src/Kokkos_Macros.hpp
```diff
@@ -153,6 +153,13 @@
   __clang_major__ * 100 + __clang_minor__ * 10 + __clang_patchlevel__
 #endif
 
+// Macro for the Cray compiler based on clang.
+// Different from classical Cray wihch is defined above.
```
masterleinad, line 157:
```suggestion
// Different from classical Cray which is defined above.
```
Is it a problem that the compiler now identifies as both `CLANG` and `CRAYCLANG`?

rgayatri23:
> > With respect to Intel compiler, I assume the same `__clang__` will be defined which will define the respective macro (although I am not sure about it).
> 
> Look at the code snippet I referenced. We define `KOKKOS_COMPILER_CLANG` only if `KOKKOS_COMPILER_INTEL` is not defined.


Ah yes thanks. I missed the 2nd condition on that line.

rgayatri23:
> > I added a comment to differentiate between classical cray and clang based cray compilers.
> 
> I am asking about clarifying why we want to distinguish them in the first place. I am not necessary suggesting we should not distinguish them. I just observing we don't do it for Intel and I am wondering whether you thought about it and why you decided to do it this way.

Added a comment on the need for a new macro since the newer Clang based Cray compilers have a mature OpenMP offload implementation but we still want the older Classical Cray macro for the host if that is being used somewhere.

Makefile.kokkos
rgayatri23: No Cray compiler does not take any architecture flags. I was also surprised by that update. Why would we compile `Cuda` backend with Cray compiler, do you mean using Cray Wrappers? `Cuda` backend points to `nvcc_wrapper` anyways right?

rgayatri23:
@masterleinad has a good point which even @dalg24 is referring to, that now we define both `Clang` and `CRAYCLANG` and if that might be an issue. I can disable `KOKKOS_COMPILER_CLANG` macro if `KOKKOS_COMPILER_CRAYCLANG` is defined, which I suppose is being done with Intel compilers. Any suggestions? I do not have an opinion on that right now.

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Exec.cpp
```diff
@@ -113,7 +113,9 @@ void OpenMPTargetExec::resize_scratch(int64_t team_size, int64_t shmem_size_L0,
   // max_active_teams is the number of active teams on the given hardware.
   // We set the number of teams to be twice the number of max_active_teams for
   // the compiler to pick the right number in its case.
+#if !defined(KOKKOS_COMPILER_CRAY_LLVM)
```
masterleinad, line 116: Can you duplicate the FIXME you added in `Kokkos_OpenMPTarget_Insstance.cpp` (// FIXME_OPENMPTARGTE: Cray compiler did not yet implement omp_set_num_teams) here?

core/src/OpenMPTarget/Kokkos_OpenMPTarget_ParallelFor_Team.hpp
```diff
@@ -115,7 +115,12 @@ class ParallelFor<FunctorType, Kokkos::TeamPolicy<Properties...>,
     // mode but works in the Debug mode.
 
     // Maximum active teams possible.
+#if !defined(KOKKOS_COMPILER_CRAY_LLVM)
```
masterleinad, line 118: Should there be a FIXME here as well?

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Parallel_Common.hpp
```diff
@@ -357,7 +357,12 @@ struct ParallelReduceSpecialize<FunctorType, TeamPolicyInternal<PolicyArgs...>,
     ValueType result = ValueType();
 
     // Maximum active teams possible.
+#if !defined(KOKKOS_COMPILER_CRAY_LLVM)
```
masterleinad, line 360: Should there be a FIXME here as well?

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Parallel_Common.hpp
```diff
@@ -436,7 +441,12 @@ struct ParallelReduceSpecialize<FunctorType, TeamPolicyInternal<PolicyArgs...>,
     void* scratch_ptr = OpenMPTargetExec::get_scratch_ptr();
 
     // Maximum active teams possible.
+#if !defined(KOKKOS_COMPILER_CRAY_LLVM)
```
masterleinad, line 444: Should there be a FIXME here as well?

cmake/kokkos_arch.cmake
```diff
@@ -682,11 +682,17 @@ ENDIF()
...
+      COMPILER_SPECIFIC_FLAGS(
+      Cray -fopenmp
+    )
+    ELSE()
+      STRING(REPLACE "sm_" "cc" NVHPC_CUDA_ARCH ${CLANG_CUDA_ARCH})
+      COMPILER_SPECIFIC_FLAGS(
       Clang -Xopenmp-target -march=${CLANG_CUDA_ARCH} -fopenmp-targets=nvptx64
       NVHPC -gpu=${NVHPC_CUDA_ARCH}
     )
+    ENDIF()
```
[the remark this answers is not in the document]
rgayatri23: Getting the closing bracket in the same level as the opening one looks weird, I tried to put the opening one in the next line but that does not match the style in the rest of the file. I indented the lines inside the brackets to make the nesting look obvious. Let me know if thats acceptable.

cmake/kokkos_arch.cmake
```diff
@@ -682,11 +682,17 @@ ENDIF()
...
   SET(CLANG_CUDA_ARCH ${KOKKOS_CUDA_ARCH_FLAG})
   IF (CLANG_CUDA_ARCH)
-    STRING(REPLACE "sm_" "cc" NVHPC_CUDA_ARCH ${CLANG_CUDA_ARCH})
-    COMPILER_SPECIFIC_FLAGS(
-      Clang -Xopenmp-target -march=${CLANG_CUDA_ARCH} -fopenmp-targets=nvptx64
-      NVHPC -gpu=${NVHPC_CUDA_ARCH}
+    IF(KOKKOS_CLANG_IS_CRAY)
+      COMPILER_SPECIFIC_FLAGS(
+      Cray -fopenmp
     )
```
masterleinad, line 688:
```suggestion
      COMPILER_SPECIFIC_FLAGS(
        Cray -fopenmp
      )
```

masterleinad (approved):
Thanks for bearing with me! 🙂

PR #5895, opened 2023-02-20
Convert HPX ParallelReduce

masterleinad:
Based on top of https://github.com/kokkos/kokkos/pull/5874.


msimberg (approved):
I can't really comment on the actual reducer interface changes, but I see nothing that seems weird for the HPX backend itself.

dalg24 (approved):

Issue #5637, opened 2022-11-22
Request for cross-platform Bessel function support.


dalg24:
Request for cross-platform Bessel function support.

Edit: Looks like the 0th and 1st order functions for the various kinds are implemented in `Kokkos_MathematicalSpecialFunctions.hpp`, but are undocumented. Implementations for safely evaluating the recurrence relation for Jn, Yn, etc. seem to be missing still.

_Originally posted by @ho37 in https://github.com/kokkos/kokkos/issues/4767#issuecomment-1298977522_

PR #5897, opened 2023-02-21
Bessel function of higher order

Yaraslaut:
This PR adds Bessel functions of higher order (#5637)

TODO:
- [x] update unit test so they cover all the code and ckeck precision for higher orders as well

### DIFF
--- a/core/src/Kokkos_MathematicalSpecialFunctions.hpp
+++ b/core/src/Kokkos_MathematicalSpecialFunctions.hpp
@@ -32,6 +32,7 @@
 
 namespace Kokkos {
 namespace Experimental {
+
 //! Compute exponential integral E1(x) (x > 0).
 template <class RealType>
 KOKKOS_INLINE_FUNCTION RealType expint1(RealType x) {
@@ -1288,9 +1289,9 @@ KOKKOS_INLINE_FUNCTION CmplxType cyl_bessel(const IntType& a,
 
 template <class CmplxType, class RealType, class IntType>
 KOKKOS_INLINE_FUNCTION CmplxType cyl_bessel_j(const IntType& a,
-                                              const CmplxType& z,
-                                              const RealType& joint_val = 25.0,
-                                              const IntType& bw_start   = 70) {
+                                              const CmplxType& z) {
+  constexpr RealType joint_val = 25.0;
+  constexpr IntType bw_start   = 70;
   if ((z.real() == 0.0) && (z.imag() == 0.0)) {
     return CmplxType(0.0, 0.0);
   }
@@ -1304,9 +1305,9 @@ KOKKOS_INLINE_FUNCTION CmplxType cyl_bessel_j(const IntType& a,
 
 template <class CmplxType, class RealType, class IntType>
 KOKKOS_INLINE_FUNCTION CmplxType cyl_bessel_y(const IntType& a,
-                                              const CmplxType& z,
-                                              const RealType& joint_val = 25,
-                                              const IntType& bw_start   = 70) {
+                                              const CmplxType& z) {
+  constexpr RealType joint_val = 25.0;
+  constexpr IntType bw_start   = 70;
   if ((z.real() == 0.0) && (z.imag() == 0.0)) {
     return CmplxType(-infinity<double>::value, 0.0);
   }
@@ -1349,9 +1350,9 @@ KOKKOS_INLINE_FUNCTION CmplxType cyl_bessel_h2(const IntType& a,
 
 template <class CmplxType, class RealType, class IntType>
 KOKKOS_INLINE_FUNCTION CmplxType cyl_bessel_i(const IntType& a,
-                                              const CmplxType& z,
-                                              const RealType& joint_val = 25,
-                                              const IntType& bw_start   = 70) {
+                                              const CmplxType& z) {
+  constexpr RealType joint_val = 25.0;
+  constexpr IntType bw_start   = 70;
   if ((z.real() == 0.0) && (z.imag() == 0.0)) {
     return CmplxType(0.0, 0.0);
   }
@@ -1369,9 +1370,9 @@ KOKKOS_INLINE_FUNCTION CmplxType cyl_bessel_i(const IntType& a,
 
 template <class CmplxType, class RealType, class IntType>
 KOKKOS_INLINE_FUNCTION CmplxType cyl_bessel_k(const IntType& a,
-                                              const CmplxType& z,
-                                              const RealType& joint_val = 9,
-                                              const IntType& bw_start   = 30) {
+                                              const CmplxType& z) {
+  constexpr RealType joint_val = 9.0;
+  constexpr IntType bw_start   = 30;
   using Kokkos::Experimental::infinity_v;
   if ((z.real() == 0.0) && (z.imag() == 0.0)) {
     constexpr auto inf = infinity_v<RealType>;
--- a/core/src/Kokkos_MathematicalSpecialFunctions.hpp
+++ b/core/src/Kokkos_MathematicalSpecialFunctions.hpp
@@ -1325,8 +1325,8 @@ KOKKOS_INLINE_FUNCTION CmplxType cyl_bessel_h1(const IntType& a,
     return CmplxType(0.0, 0.0);
   }
   auto pos_a       = Kokkos::abs(a);
-  CmplxType factor = Kokkos::exp(CmplxType(0.0, 1.0) *
-                                 Kokkos::numbers::pi_v<RealType> * pos_a);
+  auto pi          = Kokkos::numbers::pi_v<RealType>;
+  CmplxType factor = Kokkos::exp(CmplxType(0.0, 1.0) * pi * pos_a);
   return factor *
          cyl_bessel<cyl_bessel_h10<CmplxType>, cyl_bessel_h11<CmplxType>,
                     CmplxType, IntType>(pos_a, z);
@@ -1340,8 +1340,8 @@ KOKKOS_INLINE_FUNCTION CmplxType cyl_bessel_h2(const IntType& a,
   }
 
   auto pos_a       = Kokkos::abs(a);
-  CmplxType factor = Kokkos::exp(-CmplxType(0.0, 1.0) *
-                                 Kokkos::numbers::pi_v<RealType> * pos_a);
+  auto pi          = Kokkos::numbers::pi_v<RealType>;
+  CmplxType factor = Kokkos::exp(-CmplxType(0.0, 1.0) * pi * pos_a);
   return factor *
          cyl_bessel<cyl_bessel_h20<CmplxType>, cyl_bessel_h21<CmplxType>,
                     CmplxType, IntType>(pos_a, z);
--- a/core/src/Kokkos_MathematicalSpecialFunctions.hpp
+++ b/core/src/Kokkos_MathematicalSpecialFunctions.hpp
@@ -1274,12 +1274,16 @@ KOKKOS_INLINE_FUNCTION CmplxType cyl_bessel(const IntType& a,
                                             const CmplxType& z, Args... args) {
   if (a == 0) return BesselZero(z, args...);
   if (a == 1) return BesselOne(z, args...);
-  return CmplxType(2 * (a - 1)) *
-             cyl_bessel<BesselZero, BesselOne, CmplxType, IntType, Args...>(
-                 a - 1, z, args...) /
-             z -
-         cyl_bessel<BesselZero, BesselOne, CmplxType, IntType, Args...>(
-             a - 2, z, args...);
+  CmplxType bessel_im2 = BesselZero(z, args...);
+  CmplxType bessel_im1 = BesselOne(z, args...);
+
+  for (IntType cur_a = 2; cur_a <= a; ++cur_a) {
+    CmplxType tmp = bessel_im1;
+    bessel_im1    = CmplxType(2 * (cur_a - 1)) * bessel_im1 / z - bessel_im2;
+    bessel_im2    = tmp;
+  }
+
+  return bessel_im1;
 }
 
 template <class CmplxType, class RealType, class IntType>
@@ -1365,25 +1369,28 @@ KOKKOS_INLINE_FUNCTION CmplxType cyl_bessel_h2(const IntType& a,
 }
 
 template <class CmplxType, class RealType, class IntType>
-KOKKOS_INLINE_FUNCTION CmplxType cyl_bessel_i(IntType a, const CmplxType& z,
+KOKKOS_INLINE_FUNCTION CmplxType cyl_bessel_i(const IntType& a,
+                                              const CmplxType& z,
                                               const RealType& joint_val = 25,
                                               const IntType& bw_start   = 70) {
   if ((z.real() == 0.0) && (z.imag() == 0.0)) {
     return CmplxType(0.0, 0.0);
   }
-  a = Kokkos::abs(a);
   if (a == 0) return cyl_bessel_i0(z, joint_val, bw_start);
-  if (a == 1) return cyl_bessel_i1(z, joint_val, bw_start);
-  return -CmplxType(2 * (a - 1)) *
-             cyl_bessel_i<CmplxType, RealType, IntType>(a - 1, z, joint_val,
-                                                        bw_start) /
-             z +
-         cyl_bessel_i<CmplxType, RealType, IntType>(a - 2, z, joint_val,
-                                                    bw_start);
+  if (Kokkos::abs(a) == 1) return cyl_bessel_i1(z, joint_val, bw_start);
+  CmplxType bessel_im2 = cyl_bessel_i0(z, joint_val, bw_start);
+  CmplxType bessel_im1 = cyl_bessel_i1(z, joint_val, bw_start);
+  for (IntType cur_a = 2; cur_a <= a; ++cur_a) {
+    CmplxType tmp = bessel_im1;
+    bessel_im1    = -CmplxType(2 * (cur_a - 1)) * bessel_im1 / z + bessel_im2;
+    bessel_im2    = tmp;
+  }
+  return bessel_im1;
 }
 
 template <class CmplxType, class RealType, class IntType>
-KOKKOS_INLINE_FUNCTION CmplxType cyl_bessel_k(IntType a, const CmplxType& z,
+KOKKOS_INLINE_FUNCTION CmplxType cyl_bessel_k(const IntType& a,
+                                              const CmplxType& z,
                                               const RealType& joint_val = 9,
                                               const IntType& bw_start   = 30) {
   using Kokkos::Experimental::infinity_v;
@@ -1391,18 +1398,17 @@ KOKKOS_INLINE_FUNCTION CmplxType cyl_bessel_k(IntType a, const CmplxType& z,
     constexpr auto inf = infinity_v<RealType>;
     return CmplxType(inf, 0.0);
   }
-  a = Kokkos::abs(a);
-  if (a == 0) return cyl_bessel_k0(z, joint_val, bw_start);
-  if (a == 1) {
-    return cyl_bessel_k1(z, joint_val, bw_start);
-  }
 
-  return CmplxType(2 * (a - 1)) *
-             cyl_bessel_k<CmplxType, RealType, IntType>(a - 1, z, joint_val,
-                                                        bw_start) /
-             z +
-         cyl_bessel_k<CmplxType, RealType, IntType>(a - 2, z, joint_val,
-                                                    bw_start);
+  if (a == 0) return cyl_bessel_k0(z, joint_val, bw_start);
+  if (Kokkos::abs(a) == 1) return cyl_bessel_k1(z, joint_val, bw_start);
+  CmplxType bessel_im2 = cyl_bessel_k0(z, joint_val, bw_start);
+  CmplxType bessel_im1 = cyl_bessel_k1(z, joint_val, bw_start);
+  for (IntType cur_a = 2; cur_a <= a; ++cur_a) {
+    CmplxType tmp = bessel_im1;
+    bessel_im1    = CmplxType(2 * (cur_a - 1)) * bessel_im1 / z + bessel_im2;
+    bessel_im2    = tmp;
+  }
+  return bessel_im1;
 }
 
 template <class RealType, class IntType>
--- a/core/src/Kokkos_MathematicalSpecialFunctions.hpp
+++ b/core/src/Kokkos_MathematicalSpecialFunctions.hpp
@@ -1320,11 +1320,11 @@ KOKKOS_INLINE_FUNCTION CmplxType cyl_bessel_h1(const IntType& a,
                                                const CmplxType& z) {
   auto const pos_a  = Kokkos::abs(a);
   constexpr auto pi = Kokkos::numbers::pi_v<RealType>;
-  if (pos_a > 1) {
+  if ((pos_a > 1) && (z.real() == 0.0) && (z.imag() == 0.0)) {
+    return CmplxType(0.0, 0.0);
+  }
+  if (a < 0) {
     auto const factor = Kokkos::exp(CmplxType(0.0, 1.0) * pi * pos_a);
-    if ((z.real() == 0.0) && (z.imag() == 0.0)) {
-      return CmplxType(0.0, 0.0);
-    }
 
     return factor *
            cyl_bessel<cyl_bessel_h10<CmplxType>, cyl_bessel_h11<CmplxType>,
@@ -1340,11 +1340,11 @@ KOKKOS_INLINE_FUNCTION CmplxType cyl_bessel_h2(const IntType& a,
                                                const CmplxType& z) {
   auto const pos_a  = Kokkos::abs(a);
   constexpr auto pi = Kokkos::numbers::pi_v<RealType>;
-  if (pos_a > 1) {
+  if ((pos_a > 1) && (z.real() == 0.0) && (z.imag() == 0.0)) {
+    return CmplxType(0.0, 0.0);
+  }
+  if (a < 0) {
     auto const factor = Kokkos::exp(-CmplxType(0.0, 1.0) * pi * pos_a);
-    if ((z.real() == 0.0) && (z.imag() == 0.0)) {
-      return CmplxType(0.0, 0.0);
-    }
     return factor *
            cyl_bessel<cyl_bessel_h20<CmplxType>, cyl_bessel_h21<CmplxType>,
                       CmplxType, IntType>(pos_a, z);
@@ -1471,6 +1471,9 @@ template <typename RealType, typename IntType,
           typename = std::enable_if<std::is_floating_point<RealType>::value>>
 KOKKOS_INLINE_FUNCTION RealType cyl_bessel_y(const IntType& a,
                                              const RealType& z) {
+#if defined(KOKKOS_ENABLE_CUDA)
+  KOKKOS_IF_ON_DEVICE((return yn(a, z);))
+#endif
   return real(cyl_bessel_y(a, Kokkos::complex<RealType>(z)));
 }
 
--- a/core/src/Kokkos_MathematicalSpecialFunctions.hpp
+++ b/core/src/Kokkos_MathematicalSpecialFunctions.hpp
@@ -1265,7 +1265,6 @@ KOKKOS_INLINE_FUNCTION CmplxType cyl_bessel_h21(const CmplxType& z) {
     cbk1 = cyl_bessel_k1<CmplxType, RealType, int>(ci * z, 18.0, 70);
     ch21 = -2.0 / pi * cbk1;
   }
-
   return ch21;
 }
 
@@ -1283,7 +1282,6 @@ KOKKOS_INLINE_FUNCTION CmplxType cyl_bessel(const IntType& a,
     bessel_im1    = CmplxType(2 * (cur_a - 1)) * bessel_im1 / z - bessel_im2;
     bessel_im2    = tmp;
   }
-
   return bessel_im1;
 }
 
@@ -1292,15 +1290,14 @@ KOKKOS_INLINE_FUNCTION CmplxType cyl_bessel_j(const IntType& a,
                                               const CmplxType& z) {
   constexpr RealType joint_val = 25.0;
   constexpr IntType bw_start   = 70;
-  if ((z.real() == 0.0) && (z.imag() == 0.0)) {
+  auto const pos_a             = Kokkos::abs(a);
+  auto const factor            = a < 0 ? Kokkos::pow(-1.0, pos_a) : 1.0;
+  if ((pos_a > 1) && (z.real() == 0.0) && (z.imag() == 0.0)) {
     return CmplxType(0.0, 0.0);
   }
-  auto pos_a  = Kokkos::abs(a);
-  auto factor = a < 0 ? Kokkos::pow(-1.0, pos_a) : 1.0;
-  return factor *
-         cyl_bessel<cyl_bessel_j0<CmplxType, RealType, IntType>,
-                    cyl_bessel_j1<CmplxType, RealType, IntType>, CmplxType,
-                    IntType, RealType>(pos_a, z, joint_val, bw_start);
+  return factor * cyl_bessel<cyl_bessel_j0<CmplxType, RealType, IntType>,
+                             cyl_bessel_j1<CmplxType, RealType, IntType>,
+                             CmplxType, IntType>(pos_a, z, joint_val, bw_start);
 }
 
 template <class CmplxType, class RealType, class IntType>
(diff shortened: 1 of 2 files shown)


PhilMiller:
If you end up rebasing for whatever reason, there's a typo in the first commit message, 'fucntions'

PhilMiller:
OK to test

Yaraslaut:
> If you end up rebasing for whatever reason, there's a typo in the first commit message, 'fucntions'

oh, will then ask to squash pr to omit typos

mhoemmen:
I'm not a Kokkos dev, but would you consider calling the existing mathematical special functions where appropriate?

* Kokkos requires C++17, which adds [mathematical special functions](https://en.cppreference.com/w/cpp/numeric/special_functions)
* CUDA has [device implementations](https://docs.nvidia.com/cuda/cuda-math-api/group__CUDA__MATH.html#group__CUDA__MATH) of the mathematical special functions

Not using existing implementations where possible risks pessimizing both performance and accuracy.

Yaraslaut:
> I'm not a Kokkos dev, but would you consider calling the existing mathematical special functions where appropriate?
> 
>     * Kokkos requires C++17, which adds [mathematical special functions](https://en.cppreference.com/w/cpp/numeric/special_functions)
> 
>     * CUDA has [device implementations](https://docs.nvidia.com/cuda/cuda-math-api/group__CUDA__MATH.html#group__CUDA__MATH) of the mathematical special functions
> 
> 
> Not using existing implementations where possible risks pessimizing both performance and accuracy.

Hi, here is small example of how this functions can be called
```
// adl
void t()
{
  using namespace std;
  using namespace Kokkos::Experimental;

  auto d{2.0};
  cyl_bessel_j(2,d);

  auto std_complex = std::complex<double>{2.0,2.0};
  //cyl_bessel_j(2,std_complex); // error

  auto kokkos_complex = Kokkos::complex<double>{2.0,2.0};
  //cyl_bessel_j(2,kokkos_complex); //error
  cyl_bessel_j<decltype(kokkos_complex),double,int>(2,kokkos_complex);
}


  Kokkos::ScopeGuard guard(argc, argv);
  auto d = 2.0;
  auto kokkos_complex = Kokkos::complex<double>(2.0,2.0);
  auto std_complex = std::complex<double>(2.0,2.0);
  std::cyl_bessel_i(1, d);
  //std::cyl_bessel_i(1, kokkos_complex);  error
  //std::cyl_bessel_i(1, std_complex); error

  //Kokkos::Experimental::cyl_bessel_j1<decltype(d),double,int>(d); //error
  Kokkos::Experimental::cyl_bessel_j1<decltype(kokkos_complex),double,int>(kokkos_complex);
  //Kokkos::Experimental::cyl_bessel_j1<decltype(std_complex),double,int>(std_complex); //error
}


```
 * bessel functions from kokkos require explicit template arguments since they can not deduce all types from one argument only. 
 * functions from std have real arguments only so if user overlap namespaces and call bessel function directly without namespace specification it will end up calling std functoin ( `void t()` function )
  * kokkos functions can take as argument only `Kokkos::complex<>` type 
  
I think it is a good idea to have bessel functions for real types in Kokkos namespace so users can just call `Kokkos::cyl_besselj(2,2.0)` but that another topic of how properly expose special functions from `<cmath>` in Kokkos namespace so ADL can call functions

fnrizzi:
@Yaraslaut just started looking at this. 
First comment i have: 
```
auto kokkos_complex = Kokkos::complex<double>{2.0,2.0};
cyl_bessel_j<decltype(kokkos_complex),double,int>(2,kokkos_complex);
```  
I think it would be useful to allow doing this: 
```
cyl_bessel_j(2,kokkos_complex);

**Comment by Yaraslaut** [comment]
> @Yaraslaut I just started looking at this, thanks for the PR !
> 
>     1. first comment:
> 
> 
> ```c++
> auto kokkos_complex = Kokkos::complex<double>{2.0,2.0};
> cyl_bessel_j<decltype(kokkos_complex),double,int>(2,kokkos_complex);
> ```
> 
> I think it would be useful to allow doing this:
> 
> ```c++
> auto kokkos_complex = Kokkos::complex<double>{2.0,2.0};
> cyl_bessel_j(2,kokkos_complex); 
> ```
> 
> where this could be handled as:
> 
> ```c++
> template <class RealType, class IntType>
> KOKKOS_INLINE_FUNCTION CmplxType cyl_bessel_j(const IntType a,
>                                               const Kokkos::complex<RealType> & z,
>                                               const RealType& joint_val = 25,
>                                               const IntType& bw_start   = 70) {
> ```
> 
>     2. second comment: I agree with @mhoemmen: ideally we can expose as rich of an API as needed (for example, also to allow users to avoid having to specify all templates like in my example above), but internally we should call native libraries if available and if nothing is avaible have a fallback implementation (effectively what you have in this PR could be that)

Thanks for reviewing this PR, I will try to address this issues, and let you know once it is ready for further review.

**Comment by masterleinad** [comment]
Retest this please.

**Comment by Rombur** [comment]
The new tests fail with ROCm 5.6 but there is a known issue with ROCm 5.5 and 5.6. If the testsuite passes for ROCm 5.2, I am OK  with disabling the new tests for both ROCm 5.5 and 5.6.

**Comment by crtrott** [comment]
@Rombur can you do that fix for turning off the test for ROCM 5.5/5.6?

**Comment by Rombur** [comment]
Sure

**Comment by Rombur** [comment]
@Yaraslaut Your latest commit uses C++ math functions but they are not implemented in libc++ see [here](https://en.cppreference.com/w/cpp/compiler_support/17)

**Comment by Yaraslaut** [comment]
> @Yaraslaut Your latest commit uses C++ math functions but they are not implemented in libc++ see [here](https://en.cppreference.com/w/cpp/compiler_support/17)

thanks for pointing this out, hopefully I fixed it.

**Comment by dalg24** [comment]
> It fixes that error but there are others:
> 
> ```
> /var/jenkins/workspace/Kokkos/install/include/Kokkos_MathematicalSpecialFunctions.hpp(1328): error: identifier "Kokkos::numbers::pi_v<double> " is undefined in device code
> /var/jenkins/workspace/Kokkos/install/include/Kokkos_Complex.hpp(645): error: identifier "Kokkos::numbers::pi_v<double> " is undefined in device code
> /var/jenkins/workspace/Kokkos/install/include/Kokkos_Complex.hpp(599): error: identifier "Kokkos::numbers::pi_v<double> " is undefined in device code
> /var/jenkins/workspace/Kokkos/install/include/Kokkos_Complex.hpp(645): error: identifier "Kokkos::numbers::pi_v<double> " is undefined in device code
> ```
> 
> ```
> /var/jenkins/workspace/Kokkos/core/src/Kokkos_MathematicalSpecialFunctions.hpp(1328): error: identifier "Kokkos::numbers::pi_v<double> " is undefined in device code
> /var/jenkins/workspace/Kokkos/core/src/Kokkos_MathematicalSpecialFunctions.hpp(1343): error: identifier "Kokkos::numbers::pi_v<double> " is undefined in device code
> ```

The issue is that 
https://github.com/kokkos/kokkos/blob/41cf2e51cd8e63a614aa1b4a9564a35934b805d5/core/src/Kokkos_Complex.hpp#L643-L647 takes the scalar argument by reference and then (same problem) calls the complex constructor from the real part that also takes it by reference
https://github.com/kokkos/kokkos/blob/41cf2e51cd8e63a614aa1b4a9564a35934b805d5/core/src/Kokkos_Complex.hpp#L114
and, at the call site you did
```C++
 CmplxType factor = Kokkos::exp(CmplxType(0.0, 1.0) *
                                Kokkos::numbers::pi_v<RealType> * pos_a);
```
where you see the complex `CmplxType(0.0, 1.0) ` multiplies scalar `Kokkos::numbers::pi_v<RealType>` while we unfortunately cannot take the address of the Kokkos math constants on the device side.  This is a known issue:
https://github.com/kokkos/kokkos/blob/41cf2e51cd8e63a614aa1b4a9564a35934b805d5/core/unit_test/TestMathematicalConstants.hpp#L64-L72

We need to document that pitfall.

The way to work around this is to declare a local variable and make a copy.
```C++
auto pi =  Kokkos::numbers::pi_v<RealType>;
CmplxType factor = Kokkos::exp(CmplxType(0.0, 1.0) * pi * pos_a);
```

Yaraslaut:
Thanks @dalg24 fot the help in fixing this odd error, I pushed the fix

masterleinad:
Looks like the the test fails when using `std` functions:
```
4: [ RUN      ] hip.mathspecialfunc_cbesselj2y2
4: /var/jenkins/workspace/Kokkos/core/unit_test/TestMathematicalSpecialFunctions.hpp:2132: Failure
4: Expected: (Kokkos::abs(h_cbj2(i) - h_ref_cbj2(i))) <= (Kokkos::abs(h_ref_cbj2(i)) * 1e-13), actual: 0.947984 vs 1.22779e-13
4: /var/jenkins/workspace/Kokkos/core/unit_test/TestMathematicalSpecialFunctions.hpp:2132: Failure
4: Expected: (Kokkos::abs(h_cbj2(i) - h_ref_cbj2(i))) <= (Kokkos::abs(h_ref_cbj2(i)) * 1e-13), actual: 0.947984 vs 1.22779e-13
4: /var/jenkins/workspace/Kokkos/core/unit_test/TestMathematicalSpecialFunctions.hpp:2132: Failure
4: Expected: (Kokkos::abs(h_cbj2(i) - h_ref_cbj2(i))) <= (Kokkos::abs(h_ref_cbj2(i)) * 1e-13), actual: 0.947984 vs 1.22779e-13
4: /var/jenkins/workspace/Kokkos/core/unit_test/TestMathematicalSpecialFunctions.hpp:2132: Failure
4: Expected: (Kokkos::abs(h_cbj2(i) - h_ref_cbj2(i))) <= (Kokkos::abs(h_ref_cbj2(i)) * 1e-13), actual: 0.947984 vs 1.22779e-13
4: [  FAILED  ] hip.mathspecialfunc_cbesselj2y2 (0 ms)
4: [ RUN      ] hip.mathspecialfunc_cbesseli2k2
4: [       OK ] hip.mathspecialfunc_cbesseli2k2 (0 ms)
4: [ RUN      ] hip.mathspecialfunc_cbesselh12h22
4: /var/jenkins/workspace/Kokkos/core/unit_test/TestMathematicalSpecialFunctions.hpp:2491: Failure
4: Expected: (Kokkos::abs(h_cbh12(i) - h_ref_cbh12(i))) <= (Kokkos::abs(h_ref_cbh12(i)) * 1e-13), actual: 0.432749 vs 2.41179e-13
4: /var/jenkins/workspace/Kokkos/core/unit_test/TestMathematicalSpecialFunctions.hpp:2491: Failure
4: Expected: (Kokkos::abs(h_cbh12(i) - h_ref_cbh12(i))) <= (Kokkos::abs(h_ref_cbh12(i)) * 1e-13), actual: 0.432749 vs 2.50098e-13
4: /var/jenkins/workspace/Kokkos/core/unit_test/TestMathematicalSpecialFunctions.hpp:2491: Failure
4: Expected: (Kokkos::abs(h_cbh12(i) - h_ref_cbh12(i))) <= (Kokkos::abs(h_ref_cbh12(i)) * 1e-13), actual: 0.226039 vs 5.11872e-14
4: /var/jenkins/workspace/Kokkos/core/unit_test/TestMathematicalSpecialFunctions.hpp:2491: Failure
4: Expected: (Kokkos::abs(h_cbh12(i) - h_ref_cbh12(i))) <= (Kokkos::abs(h_ref_cbh12(i)) * 1e-13), actual: 0.226039 vs 5.11872e-14
4: /var/jenkins/workspace/Kokkos/core/unit_test/TestMathematicalSpecialFunctions.hpp:2491: Failure
4: Expected: (Kokkos::abs(h_cbh12(i) - h_ref_cbh12(i))) <= (Kokkos::abs(h_ref_cbh12(i)) * 1e-13), actual: 0.00343646 vs 1.66666e-14
4: /var/jenkins/workspace/Kokkos/core/unit_test/TestMathematicalSpecialFunctions.hpp:2491: Failure
4: Expected: (Kokkos::abs(h_cbh12(i) - h_ref_cbh12(i))) <= (Kokkos::abs(h_ref_cbh12(i)) * 1e-13), actual: 0.00343646 vs 1.66666e-14
4: /var/jenkins/workspace/Kokkos/core/unit_test/TestMathematicalSpecialFunctions.hpp:2497: Failure
4: Expected: (Kokkos::abs(h_cbh22(i) - h_ref_cbh22(i))) <= (Kokkos::abs(h_ref_cbh22(i)) * 1e-13), actual: 0.182976 vs 7.73567e-15
4: /var/jenkins/workspace/Kokkos/core/unit_test/TestMathematicalSpecialFunctions.hpp:2497: Failure
4: Expected: (Kokkos::abs(h_cbh22(i) - h_ref_cbh22(i))) <= (Kokkos::abs(h_ref_cbh22(i)) * 1e-13), actual: 0.182976 vs 7.73567e-15
4: /var/jenkins/workspace/Kokkos/core/unit_test/TestMathematicalSpecialFunctions.hpp:2497: Failure
4: Expected: (Kokkos::abs(h_cbh22(i) - h_ref_cbh22(i))) <= (Kokkos::abs(h_ref_cbh22(i)) * 1e-13), actual: 1.38258e-05 vs 7.45952e-19
4: /var/jenkins/workspace/Kokkos/core/unit_test/TestMathematicalSpecialFunctions.hpp:2497: Failure
4: Expected: (Kokkos::abs(h_cbh22(i) - h_ref_cbh22(i))) <= (Kokkos::abs(h_ref_cbh22(i)) * 1e-13), actual: 1.38258e-05 vs 7.45952e-19
4: /var/jenkins/workspace/Kokkos/core/unit_test/TestMathematicalSpecialFunctions.hpp:2497: Failure
4: Expected: (Kokkos::abs(h_cbh22(i) - h_ref_cbh22(i))) <= (Kokkos::abs(h_ref_cbh22(i)) * 1e-13), actual: 1.81657e-05 vs 6.79118e-19
4: /var/jenkins/workspace/Kokkos/core/unit_test/TestMathematicalSpecialFunctions.hpp:2497: Failure
4: Expected: (Kokkos::abs(h_cbh22(i) - h_ref_cbh22(i))) <= (Kokkos::abs(h_ref_cbh22(i)) * 1e-13), actual: 1.81657e-05 vs 6.79118e-19
4: /var/jenkins/workspace/Kokkos/core/unit_test/TestMathematicalSpecialFunctions.hpp:2497: Failure
4: Expected: (Kokkos::abs(h_cbh22(i) - h_ref_cbh22(i))) <= (Kokkos::abs(h_ref_cbh22(i)) * 1e-13), actual: 1.44286e-05 vs 4.66927e-19
4: /var/jenkins/workspace/Kokkos/core/unit_test/TestMathematicalSpecialFunctions.hpp:2497: Failure
4: Expected: (Kokkos::abs(h_cbh22(i) - h_ref_cbh22(i))) <= (Kokkos::abs(h_ref_cbh22(i)) * 1e-13), actual: 1.44286e-05 vs 4.66927e-19
4: [  FAILED  ] hip.mathspecialfunc_cbesselh12h22 (1 ms)
```

Rombur:
@masterleinad No I don't think so. I think that's the compiler bug that we discussed before in this PR. We just need to disable the failing tests for ROCm 5.5 and 5.6

masterleinad:
> @masterleinad No I don't think so. I think that's the compiler bug that we discussed earlier in this PR. We just need to disable the failing tests for ROCm 5.5 and 5.6

Even better!

Yaraslaut:
I think that PR is somewhat ready for review, sorry it took so long to get it working after initial implementation it :(
not sure whom should i ping to review this changes, @fnrizzi ? @Rombur ?

core/src/Kokkos_MathematicalSpecialFunctions.hpp
```diff
@@ -1268,6 +1268,222 @@ KOKKOS_INLINE_FUNCTION CmplxType cyl_bessel_h21(const CmplxType& z) {
...
+  }
+
+  return bessel_im1;
+}
+
+template <class CmplxType, class RealType, class IntType>
+KOKKOS_INLINE_FUNCTION CmplxType cyl_bessel_j(const IntType& a,
+                                              const CmplxType& z,
+                                              const RealType& joint_val = 25.0,
+                                              const IntType& bw_start   = 70) {
```
masterleinad, line 1293: I would rather not have these extra arguments (and remove the from the current public interface as well). These are implementation details.

core/src/Kokkos_MathematicalSpecialFunctions.hpp
```diff
@@ -1268,6 +1268,222 @@ KOKKOS_INLINE_FUNCTION CmplxType cyl_bessel_h21(const CmplxType& z) {
...
+  if ((z.real() == 0.0) && (z.imag() == 0.0)) {
+    return CmplxType(0.0, 0.0);
+  }
+  auto pos_a  = Kokkos::abs(a);
+  auto factor = a < 0 ? Kokkos::pow(-1.0, pos_a) : 1.0;
+  return factor *
+         cyl_bessel<cyl_bessel_j0<CmplxType, RealType, IntType>,
+                    cyl_bessel_j1<CmplxType, RealType, IntType>, CmplxType,
+                    IntType, RealType>(pos_a, z, joint_val, bw_start);
```
masterleinad, line 1302: Can't we just pass the values for `cyl_bessel_j0` and `cyl_bessel_j1` on? (Not necessarily a problem).

core/unit_test/TestMathematicalSpecialFunctions.hpp
```diff
@@ -694,51 +694,61 @@ struct TestComplexBesselJ0Y0Function {
...
+        Kokkos::Experimental::Impl::cyl_bessel_y0<Kokkos::complex<double>,
+                                                  double, int>(d_z(i));
   }
 
   KOKKOS_INLINE_FUNCTION
   void operator()(const TestLargeArgTag&, const int& /*i*/) const {
     d_cbj0_large(0) =
-        Kokkos::Experimental::cyl_bessel_j0<Kokkos::complex<double>, double,
-                                            int>(d_z_large(0));
+        Kokkos::Experimental::Impl::cyl_bessel_j0<Kokkos::complex<double>,
```
masterleinad, line 708: Why did you move `cyl_bessel_*[0|1]` to `Impl`? If we really want to remove them from the public interface, we should rather deprecate them (in the public interface).

masterleinad (commented):
Not a complete review. Just some obeservations.

core/src/Kokkos_MathematicalSpecialFunctions.hpp
Yaraslaut: previous implementation was using recursion so this functions where required
Not sure that we can change it now since it will require constexpr evaluation of `cyl_ besse{}{} ` and if we pass them by value then for zero and first order instead of one evaluation we end up with two

core/unit_test/TestMathematicalSpecialFunctions.hpp
Yaraslaut: I moved it there so user will not call  `cyl_bessel_*[0|1]`since native functions can exist for cuda or std, and it creates more uniform interface 
if exist better solution to hide this functions please tell me

core/src/Kokkos_MathematicalSpecialFunctions.hpp
masterleinad: I think it's fine.

masterleinad:
We still need to disable the tests failing for `ROCm` 5.5 and 5.6.

core/src/Kokkos_MathematicalSpecialFunctions.hpp
```diff
@@ -1268,6 +1268,304 @@ KOKKOS_INLINE_FUNCTION CmplxType cyl_bessel_h21(const CmplxType& z) {
...
+  }
+
+  return bessel_im1;
+}
+
+template <class CmplxType, class RealType, class IntType>
+KOKKOS_INLINE_FUNCTION CmplxType cyl_bessel_j(const IntType& a,
+                                              const CmplxType& z,
+                                              const RealType& joint_val = 25.0,
+                                              const IntType& bw_start   = 70) {
```
masterleinad, line 1293: In my opinion, these template arguments should only be exposed where they are directly used, i.e., in `cyl_bessel_j0` and `cyl_bessel_j1`.

core/src/Kokkos_MathematicalSpecialFunctions.hpp
```diff
@@ -1268,6 +1268,304 @@ KOKKOS_INLINE_FUNCTION CmplxType cyl_bessel_h21(const CmplxType& z) {
...
+         cyl_bessel<cyl_bessel_j0<CmplxType, RealType, IntType>,
+                    cyl_bessel_j1<CmplxType, RealType, IntType>, CmplxType,
+                    IntType, RealType>(pos_a, z, joint_val, bw_start);
+}
+
+template <class CmplxType, class RealType, class IntType>
+KOKKOS_INLINE_FUNCTION CmplxType cyl_bessel_y(const IntType& a,
+                                              const CmplxType& z,
+                                              const RealType& joint_val = 25,
+                                              const IntType& bw_start   = 70) {
```
masterleinad, line 1309: Similarly here.

core/src/Kokkos_MathematicalSpecialFunctions.hpp
```diff
@@ -1268,6 +1268,304 @@ KOKKOS_INLINE_FUNCTION CmplxType cyl_bessel_h21(const CmplxType& z) {
...
+  return factor *
+         cyl_bessel<cyl_bessel_h20<CmplxType>, cyl_bessel_h21<CmplxType>,
+                    CmplxType, IntType>(pos_a, z);
+}
+
+template <class CmplxType, class RealType, class IntType>
+KOKKOS_INLINE_FUNCTION CmplxType cyl_bessel_i(const IntType& a,
+                                              const CmplxType& z,
+                                              const RealType& joint_val = 25,
+                                              const IntType& bw_start   = 70) {
```
masterleinad, line 1354: and here

core/src/Kokkos_MathematicalSpecialFunctions.hpp
```diff
@@ -1268,6 +1268,304 @@ KOKKOS_INLINE_FUNCTION CmplxType cyl_bessel_h21(const CmplxType& z) {
...
+    bessel_im2    = tmp;
+  }
+  return bessel_im1;
+}
+
+template <class CmplxType, class RealType, class IntType>
+KOKKOS_INLINE_FUNCTION CmplxType cyl_bessel_k(const IntType& a,
+                                              const CmplxType& z,
+                                              const RealType& joint_val = 9,
+                                              const IntType& bw_start   = 30) {
```
masterleinad, line 1374: and here.

core/src/Kokkos_MathematicalSpecialFunctions.hpp
Yaraslaut: it is impossible to omit this parameters here, since templates and default arguments does not play nicely together, see:  https://godbolt.org/z/o4fG5d69f

masterleinad: Right. I suggest specifying the necessary arguments in this function explicitly instead of exposing them as function parameters.

Yaraslaut: I updated PR

core/unit_test/TestMathematicalSpecialFunctions.hpp
```diff
@@ -694,51 +694,61 @@ struct TestComplexBesselJ0Y0Function {
 
   KOKKOS_INLINE_FUNCTION
   void operator()(const int& i) const {
-    d_cbj0(i) = Kokkos::Experimental::cyl_bessel_j0<Kokkos::complex<double>,
-                                                    double, int>(d_z(i));
-    d_cby0(i) = Kokkos::Experimental::cyl_bessel_y0<Kokkos::complex<double>,
-                                                    double, int>(d_z(i));
+    d_cbj0(i) =
+        Kokkos::Experimental::Impl::cyl_bessel_j0<Kokkos::complex<double>,
```
masterleinad, line 698: Can you please change all of these to use/test the public interface?

masterleinad (commented):
Apart from testing the public interface instead of the function in `Impl`, this looks good to me (pending CI).

core/unit_test/TestMathematicalSpecialFunctions.hpp
```diff
@@ -1901,8 +1933,653 @@ struct TestComplexBesselH2Function {
...
+    d_cbj9(0) = Kokkos::Experimental::cyl_bessel_j(9, d_z(0));
+    d_cbj9(1) = Kokkos::Experimental::cyl_bessel_j(9, d_z(1));
+    d_cbj9(2) = Kokkos::Experimental::cyl_bessel_j(9, d_z(2));
+    d_cbj9(3) = Kokkos::Experimental::cyl_bessel_j(9, d_z(3));
+    d_cbj9(4) = Kokkos::Experimental::cyl_bessel_j(9, d_z(4));
+  }
+};
+
+template <class ExecSpace>
+struct TestBesselCalsWithDeduction {
```
fnrizzi, line 2558: is `Cals` a typo? is it supposed to be `Calls`?

core/unit_test/TestMathematicalSpecialFunctions.hpp
```diff
@@ -1901,8 +1933,653 @@ struct TestComplexBesselH2Function {
...
+    Kokkos::Experimental::cyl_bessel_h1(6, Kokkos::complex<double>(1.0, 1.0));
+    Kokkos::Experimental::cyl_bessel_h2(6, Kokkos::complex<double>(1.0, 1.0));
+
+    // Call to vendor functions of possible and ret real type
+    Kokkos::Experimental::cyl_bessel_j(6, 1.0);
+    Kokkos::Experimental::cyl_bessel_i(6, 1.0);
+    Kokkos::Experimental::cyl_bessel_k(6, 1.0);
+    Kokkos::Experimental::cyl_bessel_y(6, 1.0);
+    Kokkos::Experimental::cyl_bessel_h1(6, 1.0);
+    Kokkos::Experimental::cyl_bessel_h2(6, 1.0);
```
fnrizzi, line 2582: what is being tested here? just that it compiles?

core/src/Kokkos_MathematicalSpecialFunctions.hpp
```diff
@@ -1268,6 +1269,304 @@ KOKKOS_INLINE_FUNCTION CmplxType cyl_bessel_h21(const CmplxType& z) {
...
+#if defined(__cpp_lib_math_special_functions)
+  KOKKOS_IF_ON_HOST((return std::cyl_bessel_k(a, z);))
+#endif
+  return real(cyl_bessel_k(a, Kokkos::complex<RealType>(z)));
+}
+
+template <typename RealType, typename IntType>
+KOKKOS_INLINE_FUNCTION RealType cyl_bessel_y(const IntType& a,
+                                             const RealType& z) {
+  return real(cyl_bessel_y(a, Kokkos::complex<RealType>(z)));
```
fnrizzi, line 1465: why here we are not calling the cuda one like you did for `_j`?
I think cuda supports this too: https://docs.nvidia.com/cuda/cuda-math-api/group__CUDA__MATH__DOUBLE.html
or at least the version of the toolkit shown that

core/src/Kokkos_MathematicalSpecialFunctions.hpp
```diff
@@ -1268,6 +1269,304 @@ KOKKOS_INLINE_FUNCTION CmplxType cyl_bessel_h21(const CmplxType& z) {
...
+template <class CmplxType, class RealType, class IntType>
+KOKKOS_INLINE_FUNCTION CmplxType cyl_bessel_j(const IntType& a,
+                                              const CmplxType& z) {
+  constexpr RealType joint_val = 25.0;
+  constexpr IntType bw_start   = 70;
+  if ((z.real() == 0.0) && (z.imag() == 0.0)) {
+    return CmplxType(0.0, 0.0);
+  }
+  auto pos_a  = Kokkos::abs(a);
+  auto factor = a < 0 ? Kokkos::pow(-1.0, pos_a) : 1.0;
```
fnrizzi, line 1299: pos_a and factor can both be marked as const

core/src/Kokkos_MathematicalSpecialFunctions.hpp
```diff
@@ -1268,6 +1269,304 @@ KOKKOS_INLINE_FUNCTION CmplxType cyl_bessel_h21(const CmplxType& z) {
...
+template <class CmplxType, class RealType, class IntType>
+KOKKOS_INLINE_FUNCTION CmplxType cyl_bessel_y(const IntType& a,
+                                              const CmplxType& z) {
+  constexpr RealType joint_val = 25.0;
+  constexpr IntType bw_start   = 70;
+  if ((z.real() == 0.0) && (z.imag() == 0.0)) {
+    return CmplxType(-infinity<double>::value, 0.0);
+  }
+  auto pos_a  = Kokkos::abs(a);
+  auto factor = a < 0 ? Kokkos::pow(-1.0, pos_a) : 1.0;
```
fnrizzi, line 1315: pos_a and factor can both be marked as const

core/src/Kokkos_MathematicalSpecialFunctions.hpp
```diff
@@ -1268,6 +1269,304 @@ KOKKOS_INLINE_FUNCTION CmplxType cyl_bessel_h21(const CmplxType& z) {
...
+
+template <class CmplxType, class RealType, class IntType>
+KOKKOS_INLINE_FUNCTION CmplxType cyl_bessel_h1(const IntType& a,
+                                               const CmplxType& z) {
+  if ((z.real() == 0.0) && (z.imag() == 0.0)) {
+    return CmplxType(0.0, 0.0);
+  }
+  auto pos_a       = Kokkos::abs(a);
+  auto pi          = Kokkos::numbers::pi_v<RealType>;
+  CmplxType factor = Kokkos::exp(CmplxType(0.0, 1.0) * pi * pos_a);
```
fnrizzi, line 1330: pos_a and factor and pi can all be marked as const

core/src/Kokkos_MathematicalSpecialFunctions.hpp
```diff
@@ -1268,6 +1269,304 @@ KOKKOS_INLINE_FUNCTION CmplxType cyl_bessel_h21(const CmplxType& z) {
...
+template <class CmplxType, class RealType, class IntType>
+KOKKOS_INLINE_FUNCTION CmplxType cyl_bessel_h2(const IntType& a,
+                                               const CmplxType& z) {
+  if ((z.real() == 0.0) && (z.imag() == 0.0)) {
+    return CmplxType(0.0, 0.0);
+  }
+
+  auto pos_a       = Kokkos::abs(a);
+  auto pi          = Kokkos::numbers::pi_v<RealType>;
+  CmplxType factor = Kokkos::exp(-CmplxType(0.0, 1.0) * pi * pos_a);
```
fnrizzi, line 1345: pos_a and factor and pi can all be marked as const

masterleinad:
You'll have to fix the indentation.

Yaraslaut:
> You'll have to fix the indentation.

Thanks, for pointing this our,otherwise i would discover it only after a while. I tried to addres all issues, still not sure what to do with large arguments since they require customization of  default variables to get proper result

core/unit_test/TestMathematicalSpecialFunctions.hpp
Yaraslaut: Yes, only that it properly deducing templates, i can also add tests for the numerical value of those

masterleinad: Please do.

core/src/Kokkos_MathematicalSpecialFunctions.hpp
Yaraslaut: Thanks for spotting this issue, I added it there

core/unit_test/TestMathematicalSpecialFunctions.hpp
Yaraslaut: I added tests for higher orders for both real and complex arguments, and I was lucky to find out divergene between results taken from octave and julia. So I need to rethink implementation, if anyone have any ideas i will be happy to hear them

fnrizzi:
making this a draft following the comment: https://github.com/kokkos/kokkos/pull/5897#discussion_r1378422127

Yaraslaut:
Search for an existing implementation led me to discover implementation of Bessel function by Amos[1].  You can see  this is a Fortran code and i am not sure if kokkos wants to link/compile this library, perhaps @fnrizzi  or @masterleinad  can advice  how we can proceed with this PR.  


[1] https://netlib.org/amos/ 
Few other references for the history
[2] https://github.com/JuliaMath/openspecfun/tree/main/amos
[3] COMPUTATION OF BESSEL FUNCTIONS OF COMPLEX ARGUMENT BY D. E. AMOS, SAND83-0083, MAY, 1983.
[4] COMPUTATION OF BESSEL FUNCTIONS OF COMPLEX ARGUMENT AND LARGE ORDER BY D. E. AMOS, SAND83-0643, MAY, 1983
[5] SUBROUTINE PACKAGE FOR BESSEL FUNCTIONS OF A COMPLEX ARGUMENT AND NONNEGATIVE ORDER BY D. E. AMOS, SAND85-1018, MAY, 1985
[6] PORTABLE PACKAGE FOR BESSEL FUNCTIONS OF A COMPLEX ARGUMENT AND NONNEGATIVE ORDER BY D. E. AMOS, TRANS. MATH. SOFTWARE, 1986

masterleinad:
> Search for an existing implementation led me to discover implementation of Bessel function by Amos[1]. You can see this is a Fortran code and i am not sure if kokkos wants to link/compile this library, perhaps @fnrizzi or @masterleinad can advice how we can proceed with this PR.

I don't think we want to use an external library just to be able to use Bessel functions (and does that even work with the GPU backends?). I would prefer just adopting a suitable implementation and converting it to `Kokkos` code (adding the appropriate copyrights, of course).

fnrizzi:
> > Search for an existing implementation led me to discover implementation of Bessel function by Amos[1]. You can see this is a Fortran code and i am not sure if kokkos wants to link/compile this library, perhaps @fnrizzi or @masterleinad can advice how we can proceed with this PR.
> 
> I don't think we want to use an external library just to be able to use Bessel functions (and does that even work with the GPU backends?). I would prefer just adopting a suitable implementation and converting it to `Kokkos` code (adding the appropriate copyrights, of course).

i agree with Daniel!

Yaraslaut:
@masterleinad  @fnrizzi existing libraries does not support complex arguments, for example see boost: https://github.com/boostorg/math/issues/689

Issue #5898, opened 2023-02-21
CMake deduplication issue with `hip::device`


etiennemlb:
Does Kokkos support the following, that is, using the ROCm HIP package and Kokkos simultaneously:
```
add_library(mylib STATIC)
find_package(hip)
find_package(Kokkos)
target_link_library(mylib hip::device Kokkos::kokkos)
```
If this is a use case Kokkos wishes to support, I must report that explicitly switching `-x hip` on Clang based compilers like it is done in `cmake/kokkos_arch.cmake` seems flawed. Indeed, linking with `hip::device` has the side effect of adding the `-x hip` flag which Kokkos does too. For a reason I ignore, CMake's parameter deduplication will remove the `hip` of the `-x hip` provided by Kokkos leaving the `-x` dangling. In my reproducer, this produces the string `-x --rocm-patch` which Clang surely complains about.

The issue shows up in the client code, not when compiling Kokkos.
	
A solution could be to add the [SHELL:](https://gitlab.kitware.com/cmake/cmake/-/merge_requests/1841) prefix like so: `GLOBAL_APPEND(KOKKOS_AMDGPU_OPTIONS "SHELL:-x hip")` (line ~216 in `cmake/kokkos_arch.cmake`). I fear side effects but it did work flawlessly in my case.

If the lack of `SHELL:` is indeed a bug, there might be more places in Kokkos where the same remark as the one in this issue should be applied.

Here is a representation of the issue using ninja (same issue with make).

When linking with Kokkos only:
```
DEFINES = -DKOKKOS_DEPENDENCE -D__HIP_ROCclr__
DEP_FILE = CMakeFiles/mylib.dir/sources/internal/configuration.cpp.o.d
FLAGS = -O3 -DNDEBUG -Wall -Wextra -Wpedantic -O3 -fopenmp -march=znver3 -mtune=znver3 -fno-gpu-rdc -x hip --rocm-path=/opt/rocm-5.2.3 --offload-arch=gfx90a -std=gnu++17
```

When linking with Kokkos and `hip::device`. Note that the `-x hip` becomes `-x --rocm-path`.
```
DEFINES = -DKOKKOS_DEPENDENCE -D__HIP_PLATFORM_AMD__=1 -D__HIP_PLATFORM_HCC__=1 -D__HIP_ROCclr__
DEP_FILE = CMakeFiles/mylib.dir/sources/internal/configuration.cpp.o.d
[...]

PR #5901, opened 2023-02-22
Fix CMake deduplication issue when linking with hip::device

etiennemlb:
Related to #5898; the TLDR being that one does not seem able to link against both `Kokkos::kokkos` and `hip::device`.

I maintain that its more a hack than a complete fix for the issue. An alternative would have been to use `"SHELL:-x hip"`.


crtrott:
Oh are we appending `-x hip` unnecessarily when we build with CMake language support?

crtrott (approved):

crtrott:
Yeah just confirmed with Clang as compiler we also append "-x cuda" even when using CMake's language support.

dalg24:
What prevent us from implementing the suggested `"SHELL:-x hip"` alternative?

crtrott:
> What prevent us from implementing the suggested `"SHELL:-x hip"` alternative?

Nothing, but I am also not sure that it is better/more helpful?

Rombur (approved):

PR #5905, opened 2023-02-22
Fix MSVC/CUDA Build

crtrott:
This sits on top of Damiens #5870


masterleinad (approved):
Looks OK to me.

crtrott:
Rebased

PhilMiller:
The first commit looks like some sort of weird artifact of the rebase, in that it matches the message and authorship of a commit from #5870, but has distinct contents.

crtrott:
Yeah hm, but if the merge thing doesn't complain ...

core/unit_test/TestMathematicalFunctions3.hpp
```diff
@@ -0,0 +1,21 @@
...
+//
+// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
+#define KOKKOS_MATHEMATICAL_FUNCTIONS_SKIP_1
+#define KOKKOS_MATHEMATICAL_FUNCTIONS_SKIP_2
+#include "TestMathematicalFunctions.hpp"
```
PhilMiller, line 19: It would be preferable to use affirmative definitions for each subset, so that further extensions don't require successively deeper nests of conditions.

PhilMiller (approved):

core/unit_test/TestMathematicalFunctions3.hpp
dalg24: No.  It was implemented this way at my request.  The rational is that there is less chance of accidentally disabling tests with a typo for instance.

core/src/Kokkos_View.hpp
```diff
@@ -1687,8 +1687,8 @@ class View : public ViewTraits<DataType, Properties...> {
 };
 
 template <typename D, class... P>
-KOKKOS_INLINE_FUNCTION constexpr unsigned rank(const View<D, P...>& V) {
-  return V.rank();
+KOKKOS_INLINE_FUNCTION constexpr unsigned rank(const View<D, P...>&) {
+  return View<D, P...>::rank();
```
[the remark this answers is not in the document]
crtrott: Yes it does. It did not like `V.rank()` in this case (actually the issue was likely indirectly using the rank function inside a constexpr context).

crtrott:
The HIP OutOfMemory error is unrelated I would guess. Lets just wait for the MSVC build to finish.

dalg24 (approved):

Issue #5285, opened 2022-07-27
Minimum Compiler Version Update


crtrott:
After the 3.7 release, Kokkos is requiring C++17, which necessitates an upgrade in minimum compiler versions.
Furthermore, we are planning to move HIP out of Experimental in the release after 3.7 and start more stable support for ROCM versions. The question arises what that means for minimum versions Kokkos should support.

Here is the proposal we are working with right now, however we are considering lifting the version a bit higher than what we could get away with. The main reason to do so is avoiding known bugs, as well as making our burden for testing/support a bit easier.

| Compiler Line | Current Minimum | Initial Proposal | Considering | Decision | 
| --- | --- | --- | --- | --- | 
| GCC | 5.3 | 7.3 | **8.2** | 8.2 | 
| Clang | 4.0 | **8.0** | 11.1 | 8.0 | 
| Clang as CUDA compiler | 10.0 | **10.0** | 11.1 | 10.0 | 
| Intel `icpc` | 17.0 | 19.1 | **2019.5** (aka `icpc --version` 19.0.5) | 2019.5 | 
| CUDA `nvcc` | 9.2.88 | **11.0** | 11.x | 11.0 |
| Minimum CUDA for use with Clang as CUDA compiler | 10.0.1 | 10.0.1 | -- | 10.0.1 |
| ROCM | 4.5.0 | 5.2.0 | -- | 5.2.0 | 
| IntelLLVM (CPU) `icpx` | 2021.1.1 | 2021.1.1 | -- | 2021.1.1 |
| IntelLLVM (SYCL)| 2022.0.0 | 2022.2.0 | -- |  2022.2.0 |
| NVHPC `nvc++` | 22.3 | 22.3 | -- | 22.3 |
| MSVC | | | | 19.29 |
| IBM XL |  | | | Not supported |
| PGI | | | | now via NVC++ |

Couple notes: 
- Clang as CUDA compiler requires at least 10.x possibly 11.x we test 11.0 currently.
- NVC++ as CUDA compiler is not officially supported yet, we expect official support later this year
- The OpenMPTarget backend has separate requirements for newer compilers
- The upcoming OpenACC backend may have separate requirements for newer compilers

PhilMiller:
For GCC, reasons to consider 8.x or 9.y:
- NVCC with reasonable input can easily generate code that results in ICE crashes of GCC versions up to 7.4, and 7.5 is unlikely to be installed on DOE systems in place of the current widespread 7.3
[...]

PR #5906, opened 2023-02-22
Define KOKKOS_COMPILER_INTEL_LLVM and only define at most one KOKKOS_COMPILER* macro

masterleinad:
As discussed this pull request defines a new compiler macro that allows distinguishing between the Intel Classic Compiler (icpc) and the Intel oneAPI Compiler (icpx). I went through the places that use `KOKKOS_COMPILER_INTEL_LLVM` and replaced/added/omitted where I saw fit.

Also, this pull request makes sure that only at most one compiler macro `KOKKOS_COMPILER*` is defined (apart from `KOKKOS_COMPILER_NVCC` which can appear additionally). I'm happy to separate this change from this pull request if preferred.


core/src/Kokkos_Macros.hpp
```diff
@@ -148,12 +149,14 @@
 #define KOKKOS_COMPILER_APPLECC __APPLE_CC__
 #endif
 
-#if defined(__clang__) && !defined(KOKKOS_COMPILER_INTEL)
+#if defined(__clang__) && !defined(KOKKOS_COMPILER_INTEL) && \
+    !defined(KOKKOS_COMPILER_INTEL_LLVM)
```
PhilMiller, line 153:
```suggestion
#if defined(__clang__) && !defined(KOKKOS_COMPILER_INTEL_LLVM)
```
I don't think there's a case where `KOKKOS_COMPILER_INTEL` is relevant to this, but the `LLVM` macro isn't defined.

core/src/Kokkos_Macros.hpp
```diff
@@ -130,7 +131,7 @@
 #if defined(__INTEL_COMPILER)
 #define KOKKOS_COMPILER_INTEL __INTEL_COMPILER
 #elif defined(__INTEL_LLVM_COMPILER)
-#define KOKKOS_COMPILER_INTEL __INTEL_LLVM_COMPILER
+#define KOKKOS_COMPILER_INTEL_LLVM __INTEL_LLVM_COMPILER
```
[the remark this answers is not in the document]
masterleinad: No, https://www.intel.com/content/www/us/en/developer/articles/technical/use-predefined-macros-for-specific-code-for-intel-dpcpp-compiler-intel-cpp-compiler-intel-cpp-compiler-classic.html.

core/src/Kokkos_Macros.hpp
```diff
@@ -182,19 +186,16 @@
 //----------------------------------------------------------------------------
 // Intel compiler macros
 
-#if defined(KOKKOS_COMPILER_INTEL)
-// FIXME_ICPX
-#if !defined(__INTEL_LLVM_COMPILER)
+// FIXME_INTEL_LLVM
+#if defined(KOKKOS_COMPILER_INTEL_LLVM)
```
PhilMiller, line 189: The sense of the test got inverted here, I think? It had been classic `icpc` only. I think the `FIXME` comment was to see whether we should be enabling these pragmas for that as well.

masterleinad: Fixed.

core/src/Kokkos_Macros.hpp
```diff
@@ -182,19 +185,16 @@
 //----------------------------------------------------------------------------
 // Intel compiler macros
 
+// FIXME_INTEL_LLVM
```
PhilMiller, line 188:
```suggestion
// FIXME_INTEL_LLVM: Figure out whether these pragmas should get set similarly to the classic Intel compiler
```

core/src/Kokkos_Macros.hpp
```diff
@@ -148,12 +149,13 @@
...
 
-#if defined(__clang__) && !defined(KOKKOS_COMPILER_INTEL)
+#if defined(__clang__) && !defined(KOKKOS_COMPILER_INTEL_LLVM)
 #define KOKKOS_COMPILER_CLANG \
   __clang_major__ * 100 + __clang_minor__ * 10 + __clang_patchlevel__
 #endif
 
-#if !defined(__clang__) && !defined(KOKKOS_COMPILER_INTEL) && defined(__GNUC__)
+#if !defined(__clang__) && !defined(KOKKOS_COMPILER_INTEL) && \
+    !defined(KOKKOS_COMPILER_INTEL_LLVM) && defined(__GNUC__)
```
PhilMiller, line 158: I don't we need to check `INTEL_LLVM` here, since `__clang__` will be defined in that case.

PhilMiller: I.e. this line can just be reverted.

core/src/Kokkos_Macros.hpp
```diff
@@ -130,45 +131,39 @@
...
 #define KOKKOS_COMPILER_INTEL __INTEL_COMPILER
 #elif defined(__INTEL_LLVM_COMPILER)
-#define KOKKOS_COMPILER_INTEL __INTEL_LLVM_COMPILER
+#define KOKKOS_COMPILER_INTEL_LLVM __INTEL_LLVM_COMPILER
 #elif defined(__ICC)
 // Old define
 #define KOKKOS_COMPILER_INTEL __ICC
 #elif defined(__ECC)
 // Very old define
 #define KOKKOS_COMPILER_INTEL __ECC
```
PhilMiller, line 140: These are probably no longer relevant, since we bumped the minimum supported version to one that defines `__INTEL_COMPILER`

core/src/Kokkos_Macros.hpp
```diff
@@ -130,45 +131,39 @@
...
-#endif
 
+#elif defined(_CRAYC)
 // CRAY compiler for host code
-#if defined(_CRAYC)
 #define KOKKOS_COMPILER_CRAYC _CRAYC
-#endif
 
-#if defined(__APPLE_CC__)
+#elif defined(__APPLE_CC__)
```
PhilMiller, line 146: Since you're now making these definitions strictly mutually exclusive, please check that any places that were testing `KOKKOS_COMPILER_CLANG` are suitably adapted to check `defined(KOKKOS_COMPILER_CLANG) || defined(KOKKOS_COMPILER_APPLECC)`

core/src/Kokkos_Macros.hpp
```diff
@@ -130,45 +131,39 @@
...
-#endif
 
-#if defined(__clang__) && !defined(KOKKOS_COMPILER_INTEL)
+#elif defined(__clang__)
 #define KOKKOS_COMPILER_CLANG \
   __clang_major__ * 100 + __clang_minor__ * 10 + __clang_patchlevel__
-#endif
 
-#if !defined(__clang__) && !defined(KOKKOS_COMPILER_INTEL) && defined(__GNUC__)
+#elif defined(__GNUC__)
```
PhilMiller, line 153:
```suggestion
// Check this last because many compilers (at least Clang variants and Intel classic) define `__GNUC__` for compatibility
#elif defined(__GNUC__)
```

core/src/Kokkos_Macros.hpp
```diff
@@ -130,45 +131,39 @@
...
 #define KOKKOS_COMPILER_CRAYC _CRAYC
-#endif
 
-#if defined(__APPLE_CC__)
+#elif defined(__APPLE_CC__)
 #define KOKKOS_COMPILER_APPLECC __APPLE_CC__
-#endif
 
-#if defined(__clang__) && !defined(KOKKOS_COMPILER_INTEL)
+#elif defined(__clang__)
```
PhilMiller, line 149:
```suggestion
// Check this after the Clang-based proprietary compilers which will also define __clang__
#elif defined(__clang__)
```

core/src/Kokkos_Macros.hpp
```diff
@@ -130,45 +131,39 @@
...
 #define KOKKOS_COMPILER_CLANG \
   __clang_major__ * 100 + __clang_minor__ * 10 + __clang_patchlevel__
-#endif
 
-#if !defined(__clang__) && !defined(KOKKOS_COMPILER_INTEL) && defined(__GNUC__)
+#elif defined(__GNUC__)
 #define KOKKOS_COMPILER_GNU \
   __GNUC__ * 100 + __GNUC_MINOR__ * 10 + __GNUC_PATCHLEVEL__
 
 #if (530 > KOKKOS_COMPILER_GNU)
```
PhilMiller, line 157: I just noticed - this should be 820 now. I think that's perhaps already enforced in CMake, so it might be redundant. It might not be getting checked with Makefiles, though.

PhilMiller: https://github.com/kokkos/kokkos/issues/5285

core/src/Kokkos_Macros.hpp
```diff
@@ -130,45 +131,39 @@
...
 
-#if defined(__NVCOMPILER)
+#elif defined(__NVCOMPILER)
 #define KOKKOS_COMPILER_NVHPC                                 \
   __NVCOMPILER_MAJOR__ * 10000 + __NVCOMPILER_MINOR__ * 100 + \
       __NVCOMPILER_PATCHLEVEL__
-#endif
 
-#if defined(_MSC_VER) && !defined(KOKKOS_COMPILER_INTEL)
+#elif defined(_MSC_VER)
```
PhilMiller, line 166:
```suggestion
// Check this after Intel because that defines _MSC_VER for compatibility
#elif defined(_MSC_VER)
```

Does Clang on Windows also do that, like it defines `__GNUC__` on platforms where GCC was historically the default?

core/src/Kokkos_Macros.hpp
masterleinad: I did now. We only really check for `Clang` when `CUDA` or `OpenMPTarget` is enabled and `AppleClang`  wouldn't work in these cases anyway. There is of course also the check for builtins that need some love anyway.

core/src/Kokkos_Macros.hpp
masterleinad: Removed.

core/src/Kokkos_Macros.hpp
masterleinad: Yes.

example/build_cmake_in_tree/cmake_example.cpp
```diff
@@ -14,47 +14,8 @@
...
-  }
-
-  count_time = timer.seconds();
-  printf("Sequential: %ld    %10.6f\n", seq_count, count_time);
-
-  Kokkos::finalize();
-
-  return (count == seq_count) ? 0 : -1;
-}
+int main(int argc, char* argv[]) { bla(); }
```
PhilMiller, line 21: This file's changes look like they're addressing Andrey's request for an abort header, swept up in commit -a

PhilMiller (commented):
Things generally look good now. I haven't grepped for all of the uses of the affected macros myself to confirm that there are no missing updates.

example/build_cmake_in_tree/cmake_example.cpp
masterleinad: Fixed.

PhilMiller:
I've checked over all of the uses of `KOKKOS_COMPILER_INTEL` in Kokkos and Kokkos Kernels. Everything else in Kokkos is good, with maybe some `#if defined(KOKKOS_COMPILER_INTEL` gaining `|| defined(KOKKOS_COMPILER_INTEL_LLVM)`, but none of them needing to be substituted or whatever. The two uses in Kernels will end up being simplified, because they're currently `defined(INTEL) && !defined(clang)`.

There's one non-Kokkos usage of the macro in Trilinos to work around an Intel ICE [in the Kokkos 3.5 PR](https://github.com/trilinos/Trilinos/pull/9836/commits)
https://github.com/trilinos/Trilinos/blame/master/packages/intrepid2/src/Shared/Intrepid2_Data.hpp#L1688
I'll just post a PR to try removing that, and see if it blows up in testing.

PhilMiller (approved):

core/src/Kokkos_BitManipulation.hpp
```diff
@@ -183,6 +183,7 @@ rotr(T x, int s) noexcept {
 
 namespace Kokkos::Impl {
 
+// FIXME Try to use intrinsics for more compilers
```
dalg24, line 186: Drive-by change that does not belong to this PR

core/src/Kokkos_Macros.hpp
```diff
@@ -129,46 +130,41 @@
...
 #define KOKKOS_COMPILER_CRAYC _CRAYC
-#endif
 
-#if defined(__APPLE_CC__)
+#elif defined(__APPLE_CC__)
 #define KOKKOS_COMPILER_APPLECC __APPLE_CC__
-#endif
 
-#if defined(__clang__) && !defined(KOKKOS_COMPILER_INTEL)
+#elif defined(__clang__)
```
dalg24, line 144: Your PR is titled "Define KOKKOS_COMPILER_INTEL_LLVM" and this change implies for instance that now an AppleClang build still has `KOKKOS_COMPILER_APPLECC` but not `KOKKOS_COMPILER_CLANG ` any more.
Although I don't necessary object with the direction, this is clearly out of the advertised scope.

dalg24 (changes requested):

PhilMiller:
Would be good to fix the commit message to reflect the full scope of the PR.

core/src/Kokkos_BitManipulation.hpp
dalg24: Ping

dalg24 (dismissed):
Please add a `unit_tests/TestCompilerMacros.cpp` and check that one and only one macro is being defined.

core/src/Kokkos_Macros.hpp
```diff
@@ -176,19 +172,17 @@
 //----------------------------------------------------------------------------
 // Intel compiler macros
 
+// FIXME_INTEL_LLVM: Figure out whether these pragmas should get set similarly
+// to the classic Intel compiler
```
crtrott, line 176: yes

crtrott (changes requested):
I think those other macros like ivdep should be defined for Intel LLVM

core/unit_test/TestCompilerMacros.hpp
```diff
@@ -16,6 +16,70 @@
 
 #include <Kokkos_Core.hpp>
 
+#ifdef KOKKOS_COMPILER_INTEL
```
dalg24, line 19: I was more thinking something a la
https://github.com/kokkos/kokkos/blob/aa1f48f3172069f212ed2c97b74b786999b26f17/core/src/Kokkos_Macros.hpp#L408-L418

core/unit_test/TestCompilerMacros.hpp
```diff
@@ -16,6 +16,17 @@
 
 #include <Kokkos_Core.hpp>
 
+#if 1 < ((defined(KOKKOS_COMPILER_INTEL) ? 1 : 0) +      \
```
dalg24, line 19: not greater than 1, equal to 1

core/unit_test/TestCompilerMacros.hpp
```diff
@@ -51,7 +62,7 @@ struct AddFunctor {
 #pragma vector always
 #endif
 #ifdef KOKKOS_ENABLE_PRAGMA_LOOPCOUNT
-#pragma loop count(128)
+#pragma loop_count(128)
```
dalg24, line 65: uggggg

.github/workflows/continuous-integration-workflow.yml
```diff
@@ -29,12 +29,12 @@ jobs:
             backend: 'OPENMP'
           - distro: 'fedora:intel'
             cxx: 'icpx'
-            cxx_extra_flags: '-fp-model=precise'
+            cxx_extra_flags: '-fp-model=precise -Wno-pass-failed'
```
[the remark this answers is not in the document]
masterleinad: I get the warning that some loops can not be optimized as requested when enabling the pragmas for OpenMP with `icpx`.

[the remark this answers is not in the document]
dalg24: These warnings were not there before?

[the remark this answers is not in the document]
masterleinad: No, we are only now enabling a bunch of pragmas (usable in host code; depending on the compiler version).

masterleinad:
Only `gcc-8.4.0` is failing with
```
KokkosContainers_UnitTest_OpenMP (Child killed)
```

crtrott (approved):

core/src/Kokkos_Macros.hpp
```diff
@@ -123,46 +124,41 @@
...
 #elif defined(__INTEL_LLVM_COMPILER)
-#define KOKKOS_COMPILER_INTEL __INTEL_LLVM_COMPILER
-#elif defined(__ICC)
-// Old define
-#define KOKKOS_COMPILER_INTEL __ICC
-#elif defined(__ECC)
-// Very old define
-#define KOKKOS_COMPILER_INTEL __ECC
-#endif
+#define KOKKOS_COMPILER_INTEL_LLVM __INTEL_LLVM_COMPILER
```
rgayatri23, line 129: Thanks. I will follow this concept for the Cray one too in #5889

rgayatri23 (approved):

dalg24:
Ignoring GCC-8.4.0 failure which is clearly unrelated (running out of memory)
This was resolved elsewhere

PR #5907, opened 2023-02-23
Implement `rot{l,r}` function templates

dalg24:
Following up on #4577

### DIFF
--- a/core/src/Kokkos_BitManipulation.hpp
+++ b/core/src/Kokkos_BitManipulation.hpp
@@ -304,34 +304,6 @@ KOKKOS_IMPL_HOST_FUNCTION
 
 #undef KOKKOS_IMPL_USE_GCC_BUILT_IN_FUNCTIONS
 
-template <class T>
-KOKKOS_FUNCTION T rotl_builtin_host(T x, int s) noexcept {
-  return rotl(x, s);
-}
-
-template <class T>
-KOKKOS_FUNCTION T rotl_builtin_device(T x, int s) noexcept {
-#ifdef KOKKOS_ENABLE_SYCL
-  return sycl::rotate(x, s);
-#else
-  return rotl(x, s);
-#endif
-}
-
-template <class T>
-KOKKOS_FUNCTION T rotr_builtin_host(T x, int s) noexcept {
-  return rotr(x, s);
-}
-
-template <class T>
-KOKKOS_FUNCTION T rotr_builtin_device(T x, int s) noexcept {
-#ifdef KOKKOS_ENABLE_SYCL
-  return sycl::rotate(x, -s);
-#else
-  return rotr(x, s);
-#endif
-}
-
 }  // namespace Kokkos::Impl
 
 namespace Kokkos::Experimental {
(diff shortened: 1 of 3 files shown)


core/src/Kokkos_BitManipulation.hpp
```diff
@@ -278,6 +304,34 @@ KOKKOS_IMPL_HOST_FUNCTION
...
 
+template <class T>
+KOKKOS_FUNCTION T rotl_builtin_host(T x, int s) noexcept {
+  return rotl(x, s);
+}
+
+template <class T>
+KOKKOS_FUNCTION T rotl_builtin_device(T x, int s) noexcept {
+#ifdef KOKKOS_ENABLE_SYCL
+  return sycl::rotate(x, s);
```
dalg24, line 315: Did not compile because the two arguments have different types
https://github.com/intel/llvm/blob/2d47764828724b5c4e11384eb110bde427dcbd25/sycl/include/sycl/builtins.hpp#L1019-L1024

Talked with Daniel and will just use the fallback instead.

dalg24:
Retest this please

dalg24:
Retest this please

Rombur (approved):

core/unit_test/TestBitManipulationBuiltins.hpp
```diff
@@ -425,3 +427,174 @@ TEST(TEST_CATEGORY, bit_manip_bit_width) {
...
+    Kokkos::parallel_reduce(Kokkos::RangePolicy<Space>(0, N), *this, errors);
+    ASSERT_EQ(errors, 0) << "Failed check no error for " << Func::name() << "("
+                         << type_helper<typename Arg::type>::name() << ", int)";
+  }
+  KOKKOS_FUNCTION void operator()(int i, int& e) const {
+    if (Func::eval_builtin(val_[i].x, val_[i].s) !=
+        Func::eval_constexpr(val_[i].x, val_[i].s)) {
+      ++e;
+      KOKKOS_IMPL_DO_NOT_USE_PRINTF(
+          "value at %x rotated by %d which is %x was expected to be %x\n",
```
ajpowelsnl, line 476: Does the statement above clearly convey the expected condition?

ajpowelsnl (approved):

core/unit_test/TestBitManipulationBuiltins.hpp
dalg24: It prints on failure what the arguments were `value at <x> rotated by <s>`, what the function returned `is <returned>`, and what the expected value was `was expected to be <reference>`.

What is missing?

ajpowelsnl: Nothing (in terms of expected vs. actual) is missing, but this formulation (or some variation thereof) is clearer:  "It prints on failure what the arguments were value at <x> rotated by <s>, what the function returned is <returned>, and what the expected value was was expected to be <reference>."

dalg24: I don't understand your suggestion.  Can you suggest a diff?

dalg24:
Ignoring `hipErrorOutOfMemory` failures

PR #5910, opened 2023-02-23
Don't call the functor's destructor on the device for Serial and Cuda

masterleinad:
Changes in #5874 to FunctorAnalysis required some adaptions to `ParallelScan` as well and currently, we are calling the functor's destructor on the device which is problematic.
I will add this kind of fix to all the other implementations in the respective pull requests.

### DIFF
--- a/core/src/Cuda/Kokkos_Cuda_Parallel_Range.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_Parallel_Range.hpp
@@ -421,7 +421,8 @@ class ParallelScan<FunctorType, Kokkos::RangePolicy<Traits...>, Kokkos::Cuda> {
   //  (c) gridDim.x  <= blockDim.y * blockDim.y
   //  (d) gridDim.y  == gridDim.z == 1
 
-  const FunctorType m_functor;
+  const CombinedFunctorReducer<FunctorType, typename Analysis::Reducer>
+      m_functor_reducer;
   const Policy m_policy;
   word_size_type* m_scratch_space;
   size_type* m_scratch_flags;
@@ -433,23 +434,25 @@ class ParallelScan<FunctorType, Kokkos::RangePolicy<Traits...>, Kokkos::Cuda> {
   template <class TagType>
   __device__ inline std::enable_if_t<std::is_void<TagType>::value> exec_range(
       const Member& i, reference_type update, const bool final_result) const {
-    m_functor(i, update, final_result);
+    m_functor_reducer.get_functor()(i, update, final_result);
   }
 
   template <class TagType>
   __device__ inline std::enable_if_t<!std::is_void<TagType>::value> exec_range(
       const Member& i, reference_type update, const bool final_result) const {
-    m_functor(TagType(), i, update, final_result);
+    m_functor_reducer.get_functor()(TagType(), i, update, final_result);
   }
 
   //----------------------------------------
 
   __device__ inline void initial() const {
-    typename Analysis::Reducer final_reducer(m_functor);
+    const typename Analysis::Reducer& final_reducer =
+        m_functor_reducer.get_reducer();
 
     const integral_nonzero_constant<word_size_type, Analysis::StaticValueSize /
                                                         sizeof(word_size_type)>
-        word_count(Analysis::value_size(m_functor) / sizeof(word_size_type));
+        word_count(Analysis::value_size(m_functor_reducer.get_functor()) /
+                   sizeof(word_size_type));
 
     word_size_type* const shared_value =
         kokkos_impl_cuda_shared_memory<word_size_type>() +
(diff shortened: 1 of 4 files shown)


masterleinad:
Retest this please.

masterleinad:
Retest this please.

masterleinad:
This needs #5911.

PhilMiller:
The logic all looks fine, but I don't see how moving the functor instance inside another structure ends up avoiding the call to its destructor. I don't see any local temporary copies, for instance. Could you please provide some more description of what's changed (capturing subtlety?) that accomplishes that?

masterleinad:
> The logic all looks fine, but I don't see how moving the functor instance inside another structure ends up avoiding the call to its destructor. I don't see any local temporary copies, for instance. Could you please provide some more description of what's changed (capturing subtlety?) that accomplishes that?

This pull request basically changes the treatment of the functor and reducer back to what it was before #5874. What changed in that pull request was that we keep a copy of the functor/reducer in the `FunctorAnalysis` object rather than a reference to it. Thus, creating a `FunctorAnalysis` object calls the Functor's copy constructor (and also the destructor later) now. #5874 and related pull requests now aim to create the `FunctorAnalysis` on the host and then copy it to the device as usual. These series of pull requests would avoid calling the destructor for `ParallelReduce` but `FunctorAnalysis` is also used for `ParallelScan` and thus we will also need to address that as well (where it matters which largely `Cuda` and `HIP` but we should probably make the same changes at least for the other GPU backends).

core/unit_test/incremental/Test16_ParallelScan.hpp
```diff
@@ -26,11 +26,45 @@ namespace Test {
...
+    update_value += val_i;
+  }
+
+  NonTrivialScanFunctor()                              = default;
+  NonTrivialScanFunctor(NonTrivialScanFunctor const &) = default;
+  NonTrivialScanFunctor(NonTrivialScanFunctor &&)      = default;
+  NonTrivialScanFunctor &operator=(NonTrivialScanFunctor &&) = default;
+  NonTrivialScanFunctor &operator=(NonTrivialScanFunctor const &) = default;
+  ~NonTrivialScanFunctor() {}
+};
```
dalg24, line 60: The name does not clearly indicate that this is not only about not being trivial but also about no having the host device annotations on the destructor.  Plz comment or come up with a better name

core/unit_test/incremental/Test05_ParallelReduce_RangePolicy.hpp
```diff
@@ -45,7 +45,7 @@ struct NonTrivialReduceFunctor {
   NonTrivialReduceFunctor(NonTrivialReduceFunctor &&)      = default;
   NonTrivialReduceFunctor &operator=(NonTrivialReduceFunctor &&) = default;
   NonTrivialReduceFunctor &operator=(NonTrivialReduceFunctor const &) = default;
-  KOKKOS_FUNCTION ~NonTrivialReduceFunctor() {}
+  ~NonTrivialReduceFunctor() {}
```
dalg24, line 48: Same comment here

e10harvey (commented):
The changes look reasonable but I don't follow how this prevents the functor's destructor from being called on the device.

core/src/Cuda/Kokkos_Cuda_Parallel_Range.hpp
```diff
@@ -421,7 +421,8 @@ class ParallelScan<FunctorType, Kokkos::RangePolicy<Traits...>, Kokkos::Cuda> {
   //  (c) gridDim.x  <= blockDim.y * blockDim.y
   //  (d) gridDim.y  == gridDim.z == 1
 
-  const FunctorType m_functor;
+  const CombinedFunctorReducer<FunctorType, typename Analysis::Reducer>
+      m_functor_reducer;
```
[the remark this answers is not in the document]
masterleinad: `m_functor_reducer` gets created on every invocation of `parallel_scan` on the host (just like `m_functor` before). This member is effectively memcopied to the device. The point of this pull request is that the `FunctorAnalysis` object is not created on the device (since that invokes the functor's constructor and thus also the destructor eventually). Before the changes to the `ParallelReduce` interface, `FunctorAnalysis` was storing a reference so calling its constructor and destructor wasn't a problem (since it didn't invoke the functor's constructor and destructor).

Rombur (approved):

core/src/Cuda/Kokkos_Cuda_Parallel_Range.hpp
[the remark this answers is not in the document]
e10harvey: I see. Thanks for explaining that.

core/src/Cuda/Kokkos_Cuda_Parallel_Range.hpp
```diff
@@ -670,7 +677,7 @@ class ParallelScan<FunctorType, Kokkos::RangePolicy<Traits...>, Kokkos::Cuda> {
   }
 
   ParallelScan(const FunctorType& arg_functor, const Policy& arg_policy)
-      : m_functor(arg_functor),
+      : m_functor_reducer(arg_functor, typename Analysis::Reducer{arg_functor}),
```
e10harvey, line 680: This (and another PR?) are the changes that prevent the FunctorAnalysis object from being created on the device now?

masterleinad: Indirectly. The point really is to store the `FunctorAnaysis::Reducer` object (wrapping the functor) instead of storing the functor directly (and thus creating a copy on the device when creating the `FunctorAnalysis::Reducer` object).

dalg24 (approved):

masterleinad:
Only `HIP-ROCm-5.2` is failing with
```
Desul::Error: init_lock_arrays_hip: post malloc error(hipErrorOutOfMemory): hipErrorOutOfMemory
```
in the `jenkins` CI.

crtrott (approved):

Issue #5674, opened 2022-12-08
SIMD backend fixes


ibaned:
Here are a few things that @crtrott found in looking at SIMD backends

 - [ ] conversions between value types should allow all conversions, use a fallback templated constructor for each type
 - [ ] remove constructors that explicitly set each lane
 - [ ] ensure all types have generator constructors, including mask types
 - [ ] ensure all backends have `simd_mask::abi_type`
 - [ ] can we shift 32-bit integer types for all backends (and 64-bit signed for AVX2)
 - [ ] add `abs()` for all integer types

keichi:
Do you have plans to support fp32? I've been using https://github.com/kokkos/simd-math and recently tried to switch to `Kokkos::Experimental::simd`, but it seems the float support has been removed.

ibaned:
Hi @keichi , fp32 has not been a priority internally at Sandia, but if you open an issue on this repository requesting fp32 support then we can try to prioritize it. Thank you!

ldh4:
Added a task about replaceing int types with a private `simd_size_type` based on https://github.com/kokkos/kokkos/pull/6109#discussion_r1275432904.

ajpowelsnl:
@ldh4 - what's the status / ETA for this issue?

ldh4:
> @ldh4 - what's the status / ETA for this issue?

@ajpowelsnl The two remaining tasks have their respective PR. #6466 needs additional reviews. #6535 may need some adjustments, since its changes are looking to be invasive.

PR #5913, opened 2023-02-24
SIMD: make binary op tests to test against all data types

ldh4:
A prerequisite work for tasks in #5674.

SIMD unit tests are currently only testing for `double` type for all abi sets.
This PR modifies unit tests for binary ops and adds missing functions to simd classes to make them testable against all supported data types.

This PR adds:
- AVX2:
  - `copy_from` function for `int64_t`
  - Add and subtract operators for `uint64_t`
  - `const_where_expression` and `where_experssion` for `int64_t` and `uint64_t`
- AVX512:
  - `copy_to` and `copy_from` functions for `uint32_t` and `uint64_t`
  - Subtract operator for `int32_t`
  - `const_where_expression` and `where_experssion` for `uint32_t`, `int64_t` and `uint64_t`
- NEON:
  - `copy_from` function for `uint64_t`
  - `const_where_expression` and `where_experssion` for `int64_t` and `uint64_t`

### DIFF
--- a/simd/src/Kokkos_SIMD_AVX2.hpp
+++ b/simd/src/Kokkos_SIMD_AVX2.hpp
@@ -824,6 +824,19 @@ class simd<std::uint64_t, simd_abi::avx2_fixed_size<4>> {
   KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION simd(U&& value)
       : m_value(_mm256_set1_epi64x(
             Kokkos::bit_cast<std::int64_t>(value_type(value)))) {}
+  KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION simd(std::uint64_t a, std::uint64_t b,
+                                             std::uint64_t c, std::uint64_t d)
+      : m_value(_mm256_setr_epi64x(a, b, c, d)) {}
+  template <class G,
+            std::enable_if_t<
+                std::is_invocable_r_v<value_type, G,
+                                      std::integral_constant<std::size_t, 0>>,
+                bool> = false>
+  KOKKOS_FORCEINLINE_FUNCTION simd(G&& gen)
+      : simd(gen(std::integral_constant<std::size_t, 0>()),
+             gen(std::integral_constant<std::size_t, 1>()),
+             gen(std::integral_constant<std::size_t, 2>()),
+             gen(std::integral_constant<std::size_t, 3>())) {}
   KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION constexpr simd(__m256i const& value_in)
       : m_value(value_in) {}
   KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION explicit simd(
@@ -887,6 +900,22 @@ KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
 simd<std::int64_t, simd_abi::avx2_fixed_size<4>>::simd(
     simd<std::uint64_t, simd_abi::avx2_fixed_size<4>> const& other)
     : m_value(static_cast<__m256i>(other)) {}
+
+[[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
+    simd<std::uint64_t, simd_abi::avx2_fixed_size<4>>
+    operator+(simd<std::uint64_t, simd_abi::avx2_fixed_size<4>> const& lhs,
+              simd<std::uint64_t, simd_abi::avx2_fixed_size<4>> const& rhs) {
+  return simd<std::uint64_t, simd_abi::avx2_fixed_size<4>>(
+      _mm256_add_epi64(static_cast<__m256i>(lhs), static_cast<__m256i>(rhs)));
+}
+
+[[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
+    simd<std::uint64_t, simd_abi::avx2_fixed_size<4>>
+    operator-(simd<std::uint64_t, simd_abi::avx2_fixed_size<4>> const& lhs,
+              simd<std::uint64_t, simd_abi::avx2_fixed_size<4>> const& rhs) {
+  return simd<std::uint64_t, simd_abi::avx2_fixed_size<4>>(
+      _mm256_sub_epi64(static_cast<__m256i>(lhs), static_cast<__m256i>(rhs)));
+}
 
 KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
 simd<std::uint64_t, simd_abi::avx2_fixed_size<4>> condition(
--- a/simd/src/Kokkos_SIMD_AVX2.hpp
+++ b/simd/src/Kokkos_SIMD_AVX2.hpp
@@ -845,7 +845,8 @@ class simd<std::uint64_t, simd_abi::avx2_fixed_size<4>> {
   }
   KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION void copy_from(value_type const* ptr,
                                                        element_aligned_tag) {
-    m_value = _mm256_loadu_si256(reinterpret_cast<__m256i const*>(ptr));
+    m_value = _mm256_maskload_epi64(reinterpret_cast<long long const*>(ptr),
+                                    static_cast<__m256i>(mask_type(true)));
   }
   KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION simd
   operator>>(unsigned int rhs) const {
@@ -1082,7 +1083,8 @@ class const_where_expression<
 
   KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION void copy_to(
       std::int64_t* mem, element_aligned_tag) const {
-    _mm256_maskstore_epi64(mem, static_cast<__m256i>(m_mask),
+    _mm256_maskstore_epi64(reinterpret_cast<long long*>(mem),
+                           static_cast<__m256i>(m_mask),
                            static_cast<__m256i>(m_value));
   }
 
@@ -1106,8 +1108,8 @@ class where_expression<simd_mask<std::int64_t, simd_abi::avx2_fixed_size<4>>,
       : const_where_expression(mask_arg, value_arg) {}
   KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION void copy_from(std::int64_t const* mem,
                                                        element_aligned_tag) {
-    m_value =
-        value_type(_mm256_maskload_epi64(mem, static_cast<__m256i>(m_mask)));
+    m_value = value_type(_mm256_maskload_epi64(
+        reinterpret_cast<long long const*>(mem), static_cast<__m256i>(m_mask)));
   }
   template <
       class u,
@@ -1145,7 +1147,8 @@ class const_where_expression<
 
   KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION void copy_to(
       std::uint64_t* mem, element_aligned_tag) const {
-    _mm256_maskstore_epi64(mem, static_cast<__m256i>(m_mask),
+    _mm256_maskstore_epi64(reinterpret_cast<long long*>(mem),
+                           static_cast<__m256i>(m_mask),
                            static_cast<__m256i>(m_value));
   }
 
@@ -1169,8 +1172,8 @@ class where_expression<simd_mask<std::uint64_t, simd_abi::avx2_fixed_size<4>>,
       : const_where_expression(mask_arg, value_arg) {}
   KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION void copy_from(std::uint64_t const* mem,
                                                        element_aligned_tag) {
-    m_value =
-        value_type(_mm256_maskload_epi64(mem, static_cast<__m256i>(m_mask)));
+    m_value = value_type(_mm256_maskload_epi64(
+        reinterpret_cast<long long const*>(mem), static_cast<__m256i>(m_mask)));
   }
   template <class u,
             std::enable_if_t<
--- a/simd/src/Kokkos_SIMD_AVX2.hpp
+++ b/simd/src/Kokkos_SIMD_AVX2.hpp
@@ -783,6 +783,14 @@ class simd<std::int64_t, simd_abi::avx2_fixed_size<4>> {
   return simd<std::int64_t, simd_abi::avx2_fixed_size<4>>(0) - a;
 }
 
+[[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
+    simd<std::int64_t, simd_abi::avx2_fixed_size<4>>
+    operator+(simd<std::int64_t, simd_abi::avx2_fixed_size<4>> const& lhs,
+              simd<std::int64_t, simd_abi::avx2_fixed_size<4>> const& rhs) {
+  return simd<std::int64_t, simd_abi::avx2_fixed_size<4>>(
+      _mm256_add_epi64(static_cast<__m256i>(lhs), static_cast<__m256i>(rhs)));
+}
+
 KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
 simd<std::int64_t, simd_abi::avx2_fixed_size<4>> condition(
     simd_mask<std::int64_t, simd_abi::avx2_fixed_size<4>> const& a,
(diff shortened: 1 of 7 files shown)


simd/src/Kokkos_SIMD_AVX2.hpp
```diff
@@ -804,13 +812,26 @@ class simd<std::uint64_t, simd_abi::avx2_fixed_size<4>> {
...
+    return 4;
   }
   template <class U, std::enable_if_t<std::is_convertible_v<U, value_type>,
                                       bool> = false>
   KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION simd(U&& value)
       : m_value(_mm256_set1_epi64x(bit_cast<std::int64_t>(value_type(value)))) {
   }
+  KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION simd(std::uint64_t a, std::uint64_t b,
+                                             std::uint64_t c, std::uint64_t d)
+      : m_value(_mm256_setr_epi64x(a, b, c, d)) {}
```
dalg24, line 824: This constructor is not mentioned in the API and does not exist in the TS Extensions for parallelism v2.
I don't think it is OK to add.

simd/src/Kokkos_SIMD_AVX2.hpp
```diff
@@ -804,13 +812,26 @@ class simd<std::uint64_t, simd_abi::avx2_fixed_size<4>> {
   KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION simd& operator=(simd const&) = default;
   KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION simd& operator=(simd&&) = default;
   KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION static constexpr std::size_t size() {
-    return 8;
+    return 4;
```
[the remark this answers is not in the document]
PhilMiller: Presumably that would be one to cherry-pick to patch release(s)

simd/src/Kokkos_SIMD_AVX2.hpp
ibaned: Agreed, I think @crtrott and I talked about not having this type of constructor on any of the SIMD types (I added it to some but not others)

simd/src/Kokkos_SIMD_AVX2.hpp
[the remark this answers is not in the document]
ibaned: Yes and thank you for this important fix!

simd/src/Kokkos_SIMD_AVX2.hpp
```diff
@@ -1012,6 +1053,148 @@ class where_expression<simd_mask<std::int32_t, simd_abi::avx2_fixed_size<4>>,
...
+  const_where_expression(mask_type const& mask_arg, value_type const& value_arg)
+      : m_value(const_cast<value_type&>(value_arg)), m_mask(mask_arg) {}
+  [[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION constexpr mask_type const&
+  mask() const {
+    return m_mask;
+  }
+  [[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION constexpr value_type const&
+  value() const {
+    return m_value;
+  }
```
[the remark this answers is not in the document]
ibaned: `where_expression` is documented, but there is a bug in the documentation that keeps it from showing up properly: https://github.com/kokkos/kokkos-core-wiki/pull/347

ibaned:
@ldh4 thank you for this work! Everything mostly looks good to me, except the constructor mentioned above and I agree the bug fix (size of i64 AVX-512) can be made a separate bug fix PR.

simd/src/Kokkos_SIMD_AVX2.hpp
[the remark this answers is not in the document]
ldh4: Ok. I will extract this and make into a separate PR.

simd/src/Kokkos_SIMD_AVX2.hpp
ldh4: This constructor is specifically added just to enable multiplication op testing for uint64_t in AVX2 backend. This is because an instruction for 64bit ints multiplication does not exist in AVX2 instruction set and that a fallback approach is currently in place for int64_t.
https://github.com/kokkos/kokkos/blob/9480cb5f8ce56f545f6b693993dad31056d00a4f/simd/src/Kokkos_SIMD_Common.hpp#L129-L137
But I also agree that this does not seem to belong in the public API.

ldh4:
@ibaned As for the generator constructor for uint64_t in AVX2, I think it is preferred to not add it to the public API. I'll remove it from this patch and temporarily disable multiplication op testing for uint64_t in AVX2 (or uint64_t entirely from testing).

simd/src/Kokkos_SIMD_AVX2.hpp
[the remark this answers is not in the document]
dalg24: They are no `const_where_expression::{mask,value}` member functions.  These are not provided in the TS either as far as I can tell.

simd/src/Kokkos_SIMD_AVX2.hpp
ibaned: these are not the same constructor. @dalg24 is pointing to a constructor that takes four `uint64_t` values. You are referring to the constructor that takes a generator (functor).

simd/src/Kokkos_SIMD_AVX2.hpp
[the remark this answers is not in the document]
ibaned: yes, those are implementation-specific to be able to extract the mask and value for other methods that use `where_expression` to work properly. I'm not sure what Kokkos' policy is on non-public member methods. Do we just prefix them with underscores?

ldh4:
> I think I may be confused in my earlier statements... a generator constructor (that takes a functor) is part of the public API and is okay. The constructor that @dalg24 and I are suggesting may be removed is that one that just takes 4 `uint64_t` arguments.

I didn't fully think it through and also didn't describe this very well. Because the generator constructor for `uint64_t` (`KOKKOS_FORCEINLINE_FUNCTION simd(G&& gen)`) currently calls the constructor that takes in four `uint64_t`, Removing that constructor with four `uint64_t` arguments would require a change to the generator constructor to be something like this instead:

```
  KOKKOS_FORCEINLINE_FUNCTION simd(G&& gen)
      : m_value(_mm256_setr_epi64x(
            gen(std::integral_constant<std::size_t, 0>()),
            gen(std::integral_constant<std::size_t, 1>()),
            gen(std::integral_constant<std::size_t, 2>()),
            gen(std::integral_constant<std::size_t, 3>()))) {}
 ```

Alternatively, the constructor that takes in four `uint64_t` can be moved to the private section, but I think the above change may be preferred.

**Comment by ibaned** [comment]
Yes, I prefer the change you described to directly set the values and remove the intermediate constructor

File: simd/src/Kokkos_SIMD_AVX2.hpp (hunk 2 above)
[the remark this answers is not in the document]
Comment (ldh4) [reply simd/src/Kokkos_SIMD_AVX2.hpp:815]: Made in a separate PR (#6004)

File: simd/src/Kokkos_SIMD_AVX2.hpp (hunk 3 above)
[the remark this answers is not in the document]
Comment (ldh4) [reply simd/src/Kokkos_SIMD_AVX2.hpp]: Moved these functions into `Impl::` namespace and declared as friends. [a85f8b5](https://github.com/kokkos/kokkos/pull/5913/commits/a85f8b58eabcad71256787f0deb42f9a8d5bda34)

File: simd/src/Kokkos_SIMD_AVX2.hpp (hunk 1 above)
Comment (ldh4) [reply simd/src/Kokkos_SIMD_AVX2.hpp]: Removed this type of constructor from all simd types in [2d13485](https://github.com/kokkos/kokkos/pull/5913/commits/2d134858504e04b50eebb2a53b3799c0f523d7a0).

**Comment by ldh4** [comment]
Retest this please.

**Comment by ldh4** [comment]
The failing test from `KokkosContainers_UnitTest_OpenMP` seems unrelated to this PR.

**Review by Rombur — APPROVED** [review]

File: simd/unit_tests/TestSIMD.cpp [hunk 4]
```diff
@@ -434,9 +459,28 @@ KOKKOS_INLINE_FUNCTION void device_check_condition() {
...
 }
 
+template <typename Abi>
+inline void host_check_math_ops_all_types(
+    Kokkos::Experimental::Impl::data_types<>) {}
+
+template <typename Abi, typename DataType, typename... RestTypes>
+inline void host_check_math_ops_all_types(
+    Kokkos::Experimental::Impl::data_types<DataType, RestTypes...> =
+        Kokkos::Experimental::Impl::data_type_set()) {
```
dalg24, line 469: Why not using `= {}` instead of spelling it out?

simd/unit_tests/TestSIMD.cpp
```diff
@@ -434,9 +459,28 @@ KOKKOS_INLINE_FUNCTION void device_check_condition() {
   checker.truth(all_of(a == decltype(a)(16)));
 }
 
+template <typename Abi>
+inline void host_check_math_ops_all_types(
+    Kokkos::Experimental::Impl::data_types<>) {}
```
dalg24, line 464: Prefer a fold expression to a recursive implementation.

simd/unit_tests/TestSIMD.cpp
```diff
@@ -277,16 +277,41 @@ class divides {
...
+template <class Abi, size_t n, typename DataType>
+inline void host_check_all_math_ops(DataType const& first_args,
+                                    DataType const& second_args) {
   host_check_binary_op_all_loaders<Abi>(plus(), n, first_args, second_args);
   host_check_binary_op_all_loaders<Abi>(minus(), n, first_args, second_args);
   host_check_binary_op_all_loaders<Abi>(multiplies(), n, first_args,
                                         second_args);
-  host_check_binary_op_all_loaders<Abi>(divides(), n, first_args, second_args);
+
+  if constexpr (std::is_same_v<DataType, double>)
```
dalg24, line 288: Why is division test not available of the integral types?

dalg24 (commented):
Please provide a summary of what this PR is adding/fixing

simd/unit_tests/TestSIMD.cpp
ldh4: That `if` was placed because the division operation is currently only implemented for `double` type in kokkos simd.
And I am guessing that may be because the division instruction is only available for float and double types in intel intrinsics. Knowing all incoming precision types, we could emulate it if needed, but I concluded that should be an incremental work and not be part of this PR.

simd/unit_tests/TestSIMD.cpp
ldh4: Changed in [6c197bc](https://github.com/kokkos/kokkos/pull/5913/commits/6c197bc0c8d74e590b116299bd74a8f0ef90e946).

simd/unit_tests/TestSIMD.cpp
ldh4: Removed in [6c197bc](https://github.com/kokkos/kokkos/pull/5913/commits/6c197bc0c8d74e590b116299bd74a8f0ef90e946).

ldh4 (commented):
> Please provide a summary of what this PR is adding/fixing

This PR updates kokkos simd unit test to run its current tests with every supported data type for available ABIs. The initial intent was to limit modifications to the unit test file and avoid adding any invasive simd classes implementations for this PR. But the minimum set of implementation was added for each simd type to make this conversion process more flexible.

simd/unit_tests/TestSIMD.cpp
```diff
@@ -434,9 +459,21 @@ KOKKOS_INLINE_FUNCTION void device_check_condition() {
...
+inline void host_check_math_ops_all_types(
+    Kokkos::Experimental::Impl::data_types<DataTypes...>) {
+  (host_check_math_ops<Abi, DataTypes>(), ...);
+}
+
+template <typename Abi>
+inline void host_check_math_ops_all_types() {
+  host_check_math_ops_all_types<Abi>(
+      Kokkos::Experimental::Impl::data_type_set());
+}
```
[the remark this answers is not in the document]
ldh4: I'm not sure if that would work as is, since `data_types` is simply an empty class. It wouldn't know about specified template parameters without `data_type_set`. If the goal is to reduce this into calling only one function, I think it can be done in some other way. (With aggregate initialize, I can think of doing it with ctad, but I believe ctad for aggregates is c++20.)

dalg24 (commented):
(not a throughout review)
can you confirm that this PR is up-to-date and ready for review and that the description captures the current status of what is being proposed.

simd/unit_tests/TestSIMD.cpp
```diff
@@ -277,16 +277,41 @@ class divides {
...
 };
 
-template <class Abi>
-inline void host_check_math_ops() {
-  std::size_t constexpr n     = 11;
-  double const first_args[n]  = {1, 2, -1, 10, 0, 1, -2, 10, 0, 1, -2};
-  double const second_args[n] = {1, 2, 1, 1, 1, -3, -2, 1, 13, -3, -2};
+template <class Abi, size_t n, typename DataType>
+inline void host_check_all_math_ops(DataType const& first_args,
+                                    DataType const& second_args) {
```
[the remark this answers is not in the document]
ldh4: Partially addressed in [824c8ac](https://github.com/kokkos/kokkos/pull/5913/commits/824c8ac18eb658409f410b8d24acc689ae63e596).

simd/unit_tests/TestSIMD.cpp
ldh4: Added a `TODO` there for now. Because divisions are only provided for double and float types in both intel and arm intrinsics, I think there'll need to be a decision on what an integer division should produce.

ldh4:
> (not a thorough review) can you confirm that this PR is up-to-date and ready for review and that the description captures the current status of what is being proposed.

Yes, I just rebased this again on the latest `develop`, and it is ready for another round of review.

crtrott:
@ibaned can you give this another review?

ibaned (approved):
A brief skim looks good! I'm somewhat sad that `friend` was used to remove the public members `value` and `mask`, but I understand why and I am okay with the pull request exactly as it is.

simd/src/Kokkos_SIMD_AVX2.hpp
```diff
@@ -1017,6 +1048,148 @@ class where_expression<simd_mask<std::int32_t, simd_abi::avx2_fixed_size<4>>,
   void copy_from(std::int32_t const* mem, element_aligned_tag) {
     m_value = value_type(_mm_maskload_epi32(mem, static_cast<__m128i>(m_mask)));
   }
+  template <
+      class U,
+      std::enable_if_t<std::is_convertible_v<
+                           U, simd<std::int32_t, simd_abi::avx2_fixed_size<4>>>,
+                       bool> = false>
+  KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION void operator=(U&& x) {
```
dalg24, line 1056: I am a bit surprised we have assignment operators with this signature but I see there are other occurrences in develop and we can audit that elsewhere.

dalg24 (approved):

simd/unit_tests/TestSIMD.cpp
```diff
@@ -277,16 +277,42 @@ class divides {
   }
 };
 
-template <class Abi>
-inline void host_check_math_ops() {
-  std::size_t constexpr n     = 11;
-  double const first_args[n]  = {1, 2, -1, 10, 0, 1, -2, 10, 0, 1, -2};
-  double const second_args[n] = {1, 2, 1, 1, 1, -3, -2, 1, 13, -3, -2};
+template <class Abi, size_t n, typename DataType>
```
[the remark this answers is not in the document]
ldh4: Somehow I missed this when adding this line. Fixed in [6a41cd5](https://github.com/kokkos/kokkos/pull/5913/commits/6a41cd53c56eaa4028395c2b968f8babeede35d3).

simd/src/Kokkos_SIMD_AVX2.hpp
```diff
@@ -390,9 +390,6 @@ class simd<double, simd_abi::avx2_fixed_size<4>> {
                                       bool> = false>
   KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION simd(U&& value)
       : m_value(_mm256_set1_pd(value_type(value))) {}
-  KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION simd(double a, double b, double c,
```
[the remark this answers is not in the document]
ldh4: Looks like it. In fact, one of the tasks listed in #5674 is to add generator constructors to all simd types. So this will be addressed in another PR.

PR #5919, opened 2023-02-28
Export `Kokkos_CUDA_ARCHITECTURES` variable with CMake

### LINKED ISSUE #5891: Consider exporting `Kokkos_CUDA_ARCHITECTURES` variable with CMake
Currently we set `Kokkos_ARCH` as a list of enabled architectures (e.g. `"POWER9;VOLTA70"`) and it is on the user to map it back to the GPU compute capability.  Generally it is not needed but I recently came across a use case downstream where the project was setting the language to CUDA for all source files using Kokkos when `Kokkos_ENABLE_CUDA` is `TRUE`. An issue with that approach is that it is awkward to ensure that the proper `CMAKE_CUDA_ARCHITECTURES` is set.
We should consider adding support for this use case. We could set the `Kokkos_CUDA_ARCHITECTURES` variable when CUDA is enabled. I believe it is relevant regardless of whether Kokkos was built with `Kokkos_ENABLE_COMPILE_AS_CMAKE_LANGUAGE` `ON` or `OFF`.

cc @ibaned

dalg24:
Fix #5891 
I intend to deal with HIP in a follow up PR.


Rombur (approved):

dalg24:
Retest this please

PhilMiller (approved):

dalg24:
Ignoring `hipErrorOutOfMemory` failure.

PR #5932, opened 2023-03-01
add cbegin() and cend() to Kokkos::Vector

shaomeng:
This PR improves the compatibility between `Kokkos::vector` and `std::vector` by providing const iterators (cbegin, cend).

### DIFF
--- a/containers/src/Kokkos_Vector.hpp
+++ b/containers/src/Kokkos_Vector.hpp
@@ -196,7 +196,13 @@ class vector : public DualView<Scalar*, LayoutLeft, Arg1Type> {
 
   iterator begin() const { return DV::h_view.data(); }
 
+  const_iterator cbegin() const { return DV::h_view.data(); }
+
   iterator end() const {
+    return _size > 0 ? DV::h_view.data() + _size : DV::h_view.data();
+  }
+
+  const_iterator cend() const {
     return _size > 0 ? DV::h_view.data() + _size : DV::h_view.data();
   }
 


shaomeng:
Could [dalg24](https://github.com/dalg24) review this improved PR?

dalg24:
OK to test

dalg24 (approved):

containers/src/Kokkos_Vector.hpp
```diff
@@ -196,10 +196,16 @@ class vector : public DualView<Scalar*, LayoutLeft, Arg1Type> {
 
   iterator begin() const { return DV::h_view.data(); }
 
+  const_iterator cbegin() const { return DV::h_view.data(); }
+
   iterator end() const {
     return _size > 0 ? DV::h_view.data() + _size : DV::h_view.data();
   }
 
+  const_iterator cend() const {
```
nliber, line 205: I'll approve this because it does the same thing as `end()`, but why the `size_` check?  It is valid to add `0` to a valid pointer, even if that pointer has the null pointer value.

nliber (approved):

containers/src/Kokkos_Vector.hpp
shaomeng: You're correct! Adding 0 to a nullptr is a valid operation [(link)](https://stackoverflow.com/questions/59409034/is-it-allowed-to-add-a-zero-to-a-null-pointer#:~:text=Yes%2C%20you%20can%20add%20zero,Quoting%20Additive%20operators%20%5Bexpr.)! So, I can update this PR and remove the `_size` check if people agree !

dalg24: Not in this PR.

dalg24:
Retest this please

shaomeng:
> Retest this please

Not sure what to retest, since the change discussed above is not implemented?

containers/src/Kokkos_Vector.hpp
shaomeng: Ok, let's not remove the `_size` check in this PR.

dalg24:
> > Retest this please
> 
> Not sure what to retest, since the change discussed above is not implemented?

We are having issues with the testing machines

dalg24:
Ignoring `hipErrorOutOfMemory` failures.

PR #5936, opened 2023-03-02
Do not define `KOKKOS_ARCH_TURING` macro with generated GNU makefiles

dalg24:
There is no counterpart when using CMake.


dalg24:
Likely unrelated
```
37: Expected equality of these values:
37:   new_result_sum
37:     Which is: 6.9938e+07
37:   (value_type)(da_size * (da_size - 1) / 2)
37:     Which is: 7.88956e+07
37: [  FAILED  ] sycl.dynamic_view (36 ms)

Rombur (approved):

cz4rs (approved):

PR #5942, opened 2023-03-02
Tell when Kokkos atomics are disabled in print_configuration

dalg24:
https://github.com/kokkos/kokkos/pull/5940/files#diff-a65afa415b3ce21110d0c6f16b1565b9d832b2b4a804f9a037f7cd47bad36524L152-L158
was pointless because it would always say `no` but we do disable atomics when the `Serial` backend is the only one enabled
https://github.com/kokkos/kokkos/blob/d9fc6cb78688886907938a28a56eb85ca6ef2124/core/src/Kokkos_Atomics_Desul_Wrapper.hpp#L52-L56

This PR makes sure this is reflected in `Kokkos::print_configuration`.
Let me know if you have suggestions to improve the message.


dalg24:
Lovely related to #5789 and #5419

Rombur (approved):

dalg24:
Ignoring `hipErrorOutOfMemory` failure

PR #5943, opened 2023-03-02
Delete .travis.yml

junghans:


dalg24 (changes requested):
We don't accept PRs against `master`
The file has already removed in `develop` (in #5543)

dalg24:
> Yeah `master` always confused me.

Me too.  I am still lobbying to drop it.

PR #5944, opened 2023-03-02
Drop `KOKKOS_ENABLE_RFO_PREFETCH` macro

dalg24:
Related to step 3 in #5789

These are not used any more, not mentioned in the doc.  No usage in current Trilinos develop either.
They were essentially implementation details used in the legacy atomics that we removed.


cz4rs (approved):

dalg24:
Ignoring `hipErrorOutOfMemory` failures

Issue #5921, opened 2023-02-28
Broken default C++ version when enabling HIP backend


etiennemlb:
TLDR; The issue emanate from the following:
```
$ cat main.cc 
#include <cstdio>
int main() { std::printf("%ld\n", __cplusplus); }
$ CC main.cc && ./a.out 
201402
$ CC -xhip main.cc && ./a.out 
201103
```
A solution is to force CMake to produce a `-std=c++XX` using, for instance `-DCMAKE_CXX_STANDARD=14`.

The Kokkos documentation specifies `Kokkos_CXX_STANDARD: STRING Default: 14, The C++ standard for Kokkos to use` (note the default).

Assuming the following reproducer based on this CMake script:
```
cmake_minimum_required(VERSION 3.16)
project("FooProj" LANGUAGES CXX)
set(Kokkos_ENABLE_HIP ON CACHE BOOL "")
set(Kokkos_ARCH_VEGA90A ON CACHE BOOL "")
add_subdirectory(kokkos)
add_library(mylib configuration.cc)
target_link_libraries(mylib Kokkos::kokkos)
```
the following C++ code:
```
#include <cstdio>
int main() { std::printf("%ld\n", __cplusplus); }
```
and the CMake configuration and compilation line:
```
cmake -DCMAKE_CXX_COMPILER=CC -G Ninja ..  && ninja
```
We get a compilation error due to the lack of explicit C++ standard specification (`-std=c++XX`). That is, we try to compile C++14 and up code, using a compiler setup for C++11 (bunch of `std::conditional_t<...>` vs `typename std::conditional<...>::type` kind of error).

I show the issue under different compilation commands. For each command, an excerpt of the CMake configuration and Ninja build script is given:

For `cmake -DCMAKE_CXX_COMPILER=CC                          -G Ninja ..  && ninja` (**FAILS** to compile):
CMake:
```
-- Compiler Version: 15.0.0
-- Using -std=gnu++14 for C++14 extensions as feature
```
Ninja:
```
FLAGS = -fno-gpu-rdc -x hip --rocm-path=/opt/rocm-5.2.3 --offload-arch=gfx90a
```

For `cmake -DCMAKE_CXX_COMPILER=CC -DKokkos_CXX_STANDARD=14 -G Ninja ..  && ninja` (**FAILS** to compile):
CMake:
```
-- Compiler Version: 15.0.0
-- Using -std=gnu++14 for C++14 extensions as feature
```
Ninja:
```
[...]

PR #5945, opened 2023-03-02
Add error message when not using hipcc and  CMAKE_CXX_STANDARD is not set

Rombur:
When using `Cray clang` or `amdclang`, we need to set `CMAKE_CXX_STANDARD` if we want to use C++14. As reported in #5921, there is no problem when using C++17 and therefore no problem with Kokkos 4.0.


cmake/kokkos_arch.cmake
```diff
@@ -212,6 +212,9 @@ GLOBAL_SET(KOKKOS_AMDGPU_OPTIONS)
 IF(KOKKOS_ENABLE_HIP)
   SET(AMDGPU_ARCH_FLAG "--offload-arch")
   IF(NOT KOKKOS_CXX_COMPILER_ID STREQUAL HIPCC)
+    IF(KOKKOS_CXX_STANDARD STREQUAL 14 AND NOT CMAKE_CXX_STANDARD)
+      message(FATAL_ERROR "Set CMAKE_CXX_STANDARD to 14")
```
dalg24, line 216: The title line say "warning" but this is actually erroring out.  Did you mean to?

Rombur: I changed my mind after I created the PR and forgot to update the title.

dalg24 (approved):
Please update the tile line

nliber (approved):

dalg24:
@Rombur please add to the changelog

Issue #5908, opened 2023-02-23
Roadmap for using CombinedFunctorReducerType in ParallelReduce


masterleinad:
Implementations:
- Serial: https://github.com/kokkos/kokkos/pull/5874
- Cuda: https://github.com/kokkos/kokkos/pull/5874
- Threads: https://github.com/kokkos/kokkos/pull/5894
- OpenMP: https://github.com/kokkos/kokkos/pull/5893
- HPX: https://github.com/kokkos/kokkos/pull/5895
- HIP: https://github.com/kokkos/kokkos/pull/5911
- SYCL: https://github.com/kokkos/kokkos/pull/5953
- OpenMPTarget: https://github.com/kokkos/kokkos/pull/5954
- OpenACC: https://github.com/kokkos/kokkos/pull/5949

Follow-ups:
-  Don't call the functor's destructor on the device for Serial and Cuda: https://github.com/kokkos/kokkos/pull/5910
- Fix reducer result check for Serial+HPX: #5971
- Use CombinedReducer in HostIterateTile: #5962
- Deduce `value_type` from return type in `paralle_reduce` and `parallel_scan`: #5976
- Cleanup: Remove `implements_new_reduce_interface` and wrappers. #5982
- Convert `ParallelScan`, too?  Already done in most backends. The remainder can be done together with the `generic_lambdas`.

ajpowelsnl:
@masterleinad -- is all of the work associated with `parallel_reduce` wrapped up?  If so, can we close this issue?

PR #5949, opened 2023-03-03
Convert OpenACC ParallelReduce

masterleinad:
Part of #5908.

### DIFF
--- a/core/src/OpenACC/Kokkos_OpenACC_ParallelReduce_MDRange.hpp
+++ b/core/src/OpenACC/Kokkos_OpenACC_ParallelReduce_MDRange.hpp
@@ -37,40 +37,27 @@ struct OpenACCParallelReduceMDRangeHelper {
 };
 }  // namespace Kokkos::Experimental::Impl
 
-template <class Functor, class ReducerType, class... Traits>
-class Kokkos::Impl::ParallelReduce<Functor, Kokkos::MDRangePolicy<Traits...>,
-                                   ReducerType, Kokkos::Experimental::OpenACC> {
-  using Policy = MDRangePolicy<Traits...>;
+template <class CombinedFunctorReducerType, class... Traits>
+class Kokkos::Impl::ParallelReduce<CombinedFunctorReducerType,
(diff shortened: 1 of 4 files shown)


dalg24 (commented):
What about `OpenACC/Kokkos_OpenACC_ParallelScan_Range.hpp`?

masterleinad:
> What about `OpenACC/Kokkos_OpenACC_ParallelScan_Range.hpp`?

We are constructing the `FunctorAnalysis` object in that case on the host anyway so there is not much of a motivation to change it apart from consistency with the implementation on other backends that must store the `FunctorAnalysis` object instead of the functor. I would rather revisit this when touching all backends again to allow for templated functors in `parallel_scan`.

dalg24 (approved):

core/src/OpenACC/Kokkos_OpenACC_ParallelReduce_MDRange.hpp
```diff
@@ -85,16 +72,18 @@ class Kokkos::Impl::ParallelReduce<Functor, Kokkos::MDRangePolicy<Traits...>,
...
-        std::conditional_t<is_reducer_v<ReducerType>, ReducerType,
-                           Sum<ValueType>>(val),
+        Kokkos::Experimental::Impl::FunctorAdapter<FunctorType, Policy>(
+            m_functor_reducer.get_functor()),
+        std::conditional_t<
+            std::is_same_v<FunctorType, typename ReducerType::functor_type>,
+            Sum<ValueType>, typename ReducerType::functor_type>(val),
         m_policy);
 
+    reducer.final(&val);
```
Rombur, line 86: Why do we call `final` now and not before? Was there a bug?

masterleinad: Yes, this was missing and I tried to enforce it multiple times. I thought this was a good opportunity but, of course, this is somewhat of a drive-by-change.

seyonglee (approved):
It looks good to me too.

dalg24:
Ignoring HIP and SYCL failures

PR #5955, opened 2023-03-05
OpenACC Atomics

seyonglee:
Initial implementation of the OpenACC atomics constructs, where generic fall-back versions work only for serial host for now.


tpls/desul/include/desul/atomics/Fetch_Op_OpenACC.hpp
```diff
@@ -0,0 +1,521 @@
...
+namespace Impl {
+
+// clang-format off
+//<editor-fold desc="device_atomic_fetch_{add,sub,mul,div,lshift,rshift,mod,max,min,and,or,xor}">
+#pragma acc routine seq
+template <class T>
+DESUL_IMPL_ATOMICS_OPENACC_PREFIX(T,T) device_atomic_fetch_add(
+T* const ptr, const T val, MemoryOrderRelaxed, MemoryScopeDevice) {
+  T tmp;
+  T *lptr = const_cast<T *>(ptr);
```
dalg24, line 39: Is that because you can't have a const pointer in the atomic construct?
Why don't you just drop the const in the function signature if that is the case?

tpls/desul/include/desul/atomics/Fetch_Op_OpenACC.hpp
```diff
@@ -0,0 +1,521 @@
...
+  return retval;
+}
+
+#pragma acc routine seq
+template <class T, class MemoryOrder>
+DESUL_IMPL_ATOMICS_OPENACC_PREFIX(T,T) device_atomic_load(
+    const T* const ptr, MemoryOrder, MemoryScopeDevice) {
+  T retval{};
+  retval = device_atomic_load(ptr, MemoryOrderRelaxed(), MemoryScopeDevice());
+  return retval;
```
dalg24, line 510:
```suggestion
  return device_atomic_load(ptr, MemoryOrderRelaxed(), MemoryScopeDevice());
```
Also that is a bug waiting to happen.  Are the ACC atomic construct sequentially consistent?

tpls/desul/include/desul/atomics/Generic.hpp
```diff
@@ -18,18 +18,21 @@ SPDX-License-Identifier: (BSD-3-Clause)
 
 namespace desul {
 
+DESUL_ACC_ROUTINE_DIRECTIVE
```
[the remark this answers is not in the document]
seyonglee: > Does the routine directive need to appear before the `template < parameter-list >`?

Yes. In fact, the OpenACC spec does not explicitly say where to put a routine directive in a template function, but putting the routine directive after the `template<parameter-list>` causes a compile-time error when compiled by NVHPC.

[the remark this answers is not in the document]
seyonglee: > Also is it not sufficient to have the `Impl::device_atomic_*` function with the directive, we actually need it here?

The NVHPC compiler implicitly adds `#pragma acc routine seq` directives to functions called in a compute region but without explicit routine directives. However ,the implicit routine directive feature is added only in the latest OpenACC specification (V3.3), and no existing OpenACC compiler supports OpenACC V3.3 as of now. Therefore, to meet the OpenACC spec V3.2 or older, routine directives should be explicitly added to all device functions.

tpls/desul/include/desul/atomics/Macros.hpp
```diff
@@ -95,6 +99,16 @@ static constexpr bool desul_impl_omp_on_host() { return false; }
 #endif
 #endif
 
+#if defined(DESUL_HAVE_OPENACC_ATOMICS)
+#define DESUL_IF_ON_DEVICE(CODE) \
+  { DESUL_IMPL_STRIP_PARENS(CODE) }
```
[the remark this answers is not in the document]
seyonglee: OpenACC spec says that if the `acc_on_device` routine has a compile-time constant argument, the call evaluates at compile time to a constant, but NVHPC does not follow this rule (`acc_on_device` exists as a runtime function even if its argument is a compile-time constant).

core/unit_test/TestAtomics.hpp
```diff
@@ -498,7 +498,7 @@ TEST(TEST_CATEGORY, atomics) {
   ASSERT_TRUE((TestAtomic::Loop<float, TEST_EXECSPACE>(100, 2)));
   ASSERT_TRUE((TestAtomic::Loop<float, TEST_EXECSPACE>(100, 3)));
 
-#ifndef KOKKOS_ENABLE_OPENMPTARGET
+#if !defined(KOKKOS_ENABLE_OPENMPTARGET) && !defined(KOKKOS_ENABLE_OPENACC)
```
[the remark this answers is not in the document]
seyonglee: Added comments.

tpls/desul/include/desul/atomics/Lock_Based_Fetch_Op_OpenACC.hpp
```diff
@@ -0,0 +1,79 @@
...
+          //   requires !atomic_always_lock_free(sizeof(T))
+          std::enable_if_t<!atomic_always_lock_free(sizeof(T)), int> = 0>
+inline T device_atomic_fetch_oper(const Oper& op,
+                                T* const dest,
+                                dont_deduce_this_parameter_t<const T> val,
+                                MemoryOrder /*order*/,
+                                MemoryScope scope) {
+  if (acc_on_device(acc_device_not_host)) {
+    Kokkos::abort("Kokkos Error in device_atomic_fetch_oper(): Not supported atomic "
+                  "operation in the OpenACC backend");
```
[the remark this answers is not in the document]
seyonglee: Changed `Kokkos::abort()` with `printf()`

tpls/desul/include/desul/atomics/Fetch_Op_OpenACC.hpp
seyonglee: Removed `const_cast<>()` by dropping the const qualifier.

tpls/desul/include/desul/atomics/Fetch_Op_OpenACC.hpp
seyonglee: OpenACC has a weak memory model, and thus OpenACC atomic construct is NOT sequentially consistent.

seyonglee:
CI OpenACC test fails due to some compilation errors; it turned out that some unit tests fail if compiled by old NVHPC compilers (e.g., version 22.5 or older), even though no error occurs when compiled by recent NVHPC compilers (e.g., version 22.11 or later). At this point, it is unknown whether it is a compiler bug or Kokkos OpenACC backend implementation bug. I will investigate the reason for the failures.

dalg24:
OK.  We can upgrade the toolchain used in the CI as needed.

core/unit_test/CMakeLists.txt
```diff
@@ -546,9 +536,17 @@ IF(KOKKOS_ENABLE_OPENACC AND KOKKOS_CXX_COMPILER_ID STREQUAL NVHPC)
...
+    ${CMAKE_CURRENT_BINARY_DIR}/openacc/TestOpenACC_BitManipulationBuiltins.cpp
     )
 endif()
 
+# FIXME_NVHPC: Old NVHPC compilers (V22.5 or older) fail to compile
+IF(KOKKOS_ENABLE_SERIAL AND KOKKOS_CXX_COMPILER_ID STREQUAL NVHPC)
+  list(REMOVE_ITEM Serial_SOURCES2
+    ${CMAKE_CURRENT_BINARY_DIR}/serial/TestSerial_TeamBasic.cpp
+    )
+ENDIF()
```
dalg24, line 548: We don't support older NVHPC compiler.  We can upgrade the SDK in the CI build and require a more recent version.

seyonglee: Then, if we choose NVHPC V22.7 as the minimum version, we can re-enable the disabled unit tests.
(In fact, I couldn't check whether V22.6 works or not because I don't have an access  to NVHPC V22.6)

seyonglee: Restored unit tests that were disabled for old NVHPC compilers (V22.5 or older).
The issues on NVHPC V22.5 are related to supporting atomic operations on float/double variables, which are necessary features. 
Therefore, it will be better to change the minimum required version of NVHPC to V22.7 for the OpenACC backend.

seyonglee:
Changed the required minimum version of the NVHPC compiler from 22.3 to 22.9; now this PR passes all the unit tests for the OpenACC backend.

core/unit_test/TestAtomicOperations_double.hpp
```diff
@@ -21,10 +21,13 @@ TEST(TEST_CATEGORY, atomic_operations_double) {
   const int start = 1;  // Avoid zero for division.
   const int end   = 11;
   for (int i = start; i < end; ++i) {
+#ifndef KOKKOS_ENABLE_OPENACC
+    // FIXME_OPENACC: OpenACC C/C++ does not support atomic min/max operations
```
dalg24, line 25: Did you consider guarding in `AtomicOperationsTestNonIntegralType` instead?

tpls/desul/include/desul/atomics/Compare_Exchange_OpenACC.hpp
```diff
@@ -0,0 +1,266 @@
...
+
+#ifdef KOKKOS_COMPILER_NVHPC
+
+#pragma acc routine seq
+template <class T, class MemoryOrder, class MemoryScope>
+std::enable_if_t<!std::is_arithmetic<T>::value || ((sizeof(T) != 4) && (sizeof(T) != 8)), T> 
+	device_atomic_compare_exchange(T* dest, T compare, T value, MemoryOrder, MemoryScope scope) {
+  T current_val = *dest;
+  if (acc_on_device(acc_device_not_host)) {
+    printf("Kokkos Error in device_atomic_compare_exchange(): Not supported atomic "
```
masterleinad, line 154: This is `desul` not `Kokkos`.

core/unit_test/TestAtomicViews.hpp
```diff
@@ -1392,9 +1392,12 @@ TEST(TEST_CATEGORY, atomic_views_integral) {
     ASSERT_TRUE(
         (TestAtomicViews::AtomicViewsTestIntegralType<int64_t, TEST_EXECSPACE>(
             length, 4)));
+#ifndef KOKKOS_ENABLE_OPENACC
```
dalg24, line 1395: Same question about guarding in `AtomicViewsTestIntegralType` instead.

tpls/desul/include/desul/atomics/Fetch_Op_OpenACC.hpp
```diff
@@ -0,0 +1,490 @@
...
+#define DESUL_ATOMICS_FETCH_OP_OPENACC_HPP_
+
+#include <desul/atomics/Common.hpp>
+#include <algorithm>
+
+#ifdef KOKKOS_COMPILER_NVHPC
+#ifndef DESUL_CUDA_ARCH_IS_PRE_PASCAL
+#define DESUL_IMPL_ATOMICS_OPENACC_PREFIX(T,R) std::enable_if_t<std::is_same_v<T,int> \
+	|| std::is_same_v<T,unsigned int> || std::is_same_v<T,unsigned long long> \
+	|| std::is_same_v<T,float> || std::is_same_v<T,double>, R>
```
masterleinad, line 18: Make sure to fix the indentation when opening the `desul` pull request.

tpls/desul/include/desul/atomics/Compare_Exchange_OpenACC.hpp
```diff
@@ -0,0 +1,266 @@
...
+  //}
+  //device_atomic_thread_fence(MemoryOrderAcquire(), scope);
+  T return_val = *dest;
+  *dest = value;
+  //device_atomic_thread_fence(MemoryOrderRelease(), scope);
+  //unlock_address_openacc((void*)dest, scope);
+  return return_val;
+}
+
+#ifdef KOKKOS_COMPILER_NVHPC
```
dalg24, line 39: You might want to add a comment or define a macro that express better that we will be using CUDA atomic functions in that case.

tpls/desul/include/desul/atomics/Compare_Exchange_OpenACC.hpp
```diff
@@ -0,0 +1,266 @@
...
+}
+
+#pragma acc routine seq
+template <class T, class MemoryOrder, class MemoryScope>
+std::enable_if_t<std::is_integral<T>::value && (sizeof(T) == 4), T> 
+	device_atomic_exchange(T* const dest, T value, MemoryOrder, MemoryScope) {
+  static_assert(sizeof(unsigned int) == 4,
+                "this function assumes an unsigned int is 32-bit");
+  unsigned int return_val = atomicExch(reinterpret_cast<unsigned int*>(dest),
+                                       reinterpret_cast<unsigned int&>(value));
```
dalg24, line 67: So the compiler will emit host code when appropriate?

masterleinad (commented):
You'll need a corresponding pull request on the `desul`-side.

tpls/desul/include/desul/atomics/Compare_Exchange_OpenACC.hpp
```diff
@@ -0,0 +1,266 @@
...
+  T return_val = *dest;
+  *dest = value;
+  //device_atomic_thread_fence(MemoryOrderRelease(), scope);
+  //unlock_address_openacc((void*)dest, scope);
+  return return_val;
+}
+
+#endif
+
+#else
```
dalg24, line 129: Why is it advantageous to use CUDA atomics with NVHPC rather than use acc atomic capture unconditionally?

tpls/desul/include/desul/atomics/Fetch_Op_OpenACC.hpp
```diff
@@ -0,0 +1,490 @@
...
+#else
+#define DESUL_IMPL_ATOMICS_OPENACC_PREFIX(T,R) std::enable_if_t<std::is_same_v<T,int> \
+	|| std::is_same_v<T,unsigned int> || std::is_same_v<T,unsigned long long> \
+	|| std::is_same_v<T,float>, R>
+#endif
+#else
+#define DESUL_IMPL_ATOMICS_OPENACC_PREFIX(T,R) std::enable_if_t<std::is_arithmetic<T>::value, R>
```
dalg24, line 25: I am not sure I like our using a macro for this.
We could use an alias template.

```C++
template <class T>
using acc_enable_if_supported_arithmetic_type = ...
```
or whatnot.

tpls/desul/include/desul/atomics/Fetch_Op_OpenACC.hpp
```diff
@@ -0,0 +1,490 @@
...
+  tmp = *ptr; *ptr = std::max(*ptr, val);
+  return tmp;
+}
+
+#ifdef KOKKOS_COMPILER_NVHPC
+#pragma acc routine seq
+inline int device_atomic_fetch_max(
+int* ptr, const int val, MemoryOrderRelaxed, MemoryScopeDevice) {
+  int tmp;
+  tmp = atomicMax(ptr, val);
```
dalg24, line 144: Presumably this only works with NVHPC

core/unit_test/CMakeLists.txt
```diff
@@ -507,17 +497,10 @@ IF(KOKKOS_ENABLE_OPENACC AND KOKKOS_CXX_COMPILER_ID STREQUAL NVHPC)
     ${CMAKE_CURRENT_BINARY_DIR}/openacc/TestOpenACC_Reducers_d.cpp
     ${CMAKE_CURRENT_BINARY_DIR}/openacc/TestOpenACC_Reductions.cpp
     ${CMAKE_CURRENT_BINARY_DIR}/openacc/TestOpenACC_Reductions_DeviceView.cpp
-    ${CMAKE_CURRENT_BINARY_DIR}/openacc/TestOpenACC_SubView_b.cpp
-    ${CMAKE_CURRENT_BINARY_DIR}/openacc/TestOpenACC_SubView_c02.cpp
-    ${CMAKE_CURRENT_BINARY_DIR}/openacc/TestOpenACC_SubView_c03.cpp
-    ${CMAKE_CURRENT_BINARY_DIR}/openacc/TestOpenACC_SubView_c05.cpp
-    ${CMAKE_CURRENT_BINARY_DIR}/openacc/TestOpenACC_SubView_c08.cpp
-    ${CMAKE_CURRENT_BINARY_DIR}/openacc/TestOpenACC_SubView_c11.cpp
```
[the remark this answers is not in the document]
seyonglee: I previously disabled all unit tests using any atomics regardless of whether host-side or not.

core/unit_test/CMakeLists.txt
```diff
@@ -507,17 +497,10 @@ IF(KOKKOS_ENABLE_OPENACC AND KOKKOS_CXX_COMPILER_ID STREQUAL NVHPC)
...
     ${CMAKE_CURRENT_BINARY_DIR}/openacc/TestOpenACC_TeamBasic.cpp
     ${CMAKE_CURRENT_BINARY_DIR}/openacc/TestOpenACC_TeamScratch.cpp
     ${CMAKE_CURRENT_BINARY_DIR}/openacc/TestOpenACC_TeamTeamSize.cpp
     ${CMAKE_CURRENT_BINARY_DIR}/openacc/TestOpenACC_UniqueToken.cpp
-    ${CMAKE_CURRENT_BINARY_DIR}/openacc/TestOpenACC_ViewMapping_b.cpp
```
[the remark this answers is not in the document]
seyonglee: I previously disabled all unit tests using any atomics regardless of whether host-side or not.

cmake/kokkos_compiler_id.cmake
```diff
@@ -155,7 +155,7 @@ SET(KOKKOS_MESSAGE_TEXT "${KOKKOS_MESSAGE_TEXT}\n    IntelLLVM(CPU)  2021.1.1 or
 SET(KOKKOS_MESSAGE_TEXT "${KOKKOS_MESSAGE_TEXT}\n    IntelLLVM(SYCL) 2023.0.0 or higher")
 SET(KOKKOS_MESSAGE_TEXT "${KOKKOS_MESSAGE_TEXT}\n    NVCC              11.0.0 or higher")
 SET(KOKKOS_MESSAGE_TEXT "${KOKKOS_MESSAGE_TEXT}\n    HIPCC              5.2.0 or higher")
-SET(KOKKOS_MESSAGE_TEXT "${KOKKOS_MESSAGE_TEXT}\n    NVHPC/PGI           22.3 or higher")
+SET(KOKKOS_MESSAGE_TEXT "${KOKKOS_MESSAGE_TEXT}\n    NVHPC/PGI           22.9 or higher")
```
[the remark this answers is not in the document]
seyonglee: As I know, NVHPC is used for both OpenACC and CUDA backends, and CI tests for the CUDA backend use NVHPC V22.9; that's why V22.9 is selected for the minimum required version of NVHPC. (For the OpenACC backend, the minimum required version of NVHPC is V22.7.)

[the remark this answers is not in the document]
masterleinad: > For the OpenACC backend, the minimum required version of NVHPC is V22.7.

I don't find `22.7` anywhere in `Kokkos` `develop`. Anyway, I think raising the minimum should be discussed outside this pull request.

tpls/desul/include/desul/atomics/Compare_Exchange_OpenACC.hpp
seyonglee: CUDA allows the same atomic intrinsic functions to be called on both the host and the device; therefore, the same CUDA atomic-intrinsic-based implementation is used for both the host and the device.

tpls/desul/include/desul/atomics/Compare_Exchange_OpenACC.hpp
seyonglee: CUDA atomics were used to 1) avoid weird compilation errors when compiling some OpenACC atomic constructs and 2) reduce the number of instructions (even though its performance benefit is unclear).

tpls/desul/include/desul/atomics/Compare_Exchange_OpenACC.hpp
```diff
@@ -0,0 +1,266 @@
...
+  return reinterpret_cast<T&>(return_val);
+}
+
+#else
+
+#pragma acc routine seq
+template <class T, class MemoryOrder, class MemoryScope>
+std::enable_if_t<std::is_floating_point<T>::value && (sizeof(T) == 8), T> 
+	device_atomic_exchange(T* const dest, T value, MemoryOrder, MemoryScope scope) {
+  if (acc_on_device(acc_device_not_host)) {
```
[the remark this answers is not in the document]
masterleinad: I understand that but the implementation will also not work correctly on the host (for host-parallel backends) so I would expect to always print an error message if you call this function (and if it's not called on the host, which would be preferable since we have other implementations for that anyway, there is also no point in checking).

tpls/desul/include/desul/atomics/Fetch_Op_OpenACC.hpp
seyonglee: Yes, using CUDA atomic intrinsics will work only with NVHPC.

tpls/desul/include/desul/atomics/Thread_Fence_OpenACC.hpp
```diff
@@ -0,0 +1,27 @@
...
+SPDX-License-Identifier: (BSD-3-Clause)
+*/
+
+#ifndef DESUL_ATOMICS_THREAD_FENCE_OPENACC_HPP_
+#define DESUL_ATOMICS_THREAD_FENCE_OPENACC_HPP_
+
+#include <openacc.h>
+
+#include <desul/atomics/Common.hpp>
+
```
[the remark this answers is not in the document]
seyonglee: Deleted.

tpls/desul/include/desul/atomics/Compare_Exchange_OpenACC.hpp
seyonglee: "Kokkos Error" is changed to "DESUL error".

tpls/desul/include/desul/atomics/Compare_Exchange_OpenACC.hpp
seyonglee: Added a comment.

core/unit_test/TestAtomicOperations_double.hpp
seyonglee: Changed the way to guard unsupported atomic tests for the OpenACC backend.

core/unit_test/TestAtomicViews.hpp
seyonglee: Changed the way to guard unsupported atomic tests for the OpenACC backend.

tpls/desul/include/desul/atomics/Fetch_Op_OpenACC.hpp
seyonglee: Changed the macro to an alias template as suggested.

.jenkins
```diff
@@ -37,7 +37,7 @@ pipeline {
                         }
                     }
                     environment {
-                        CUDA_HOME = '/opt/nvidia/hpc_sdk/Linux_x86_64/22.3/cuda/11.6'
```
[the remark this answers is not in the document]
seyonglee: Reverted the changes on the minimum required version of NVHPC; instead use a new base to build a docker image for OPENACC-NVHPC-CUDA-11.7.

cmake/kokkos_compiler_id.cmake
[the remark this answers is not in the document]
seyonglee: Undo changing the minimum required version of NVHPC.

tpls/desul/include/desul/atomics/Compare_Exchange_OpenACC.hpp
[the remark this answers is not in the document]
seyonglee: The current OpenACC atomic implementation supports the host only as a serial backend due to several unresolved design issues (e.g., distinguishing a serial host, OpenMP-parallel host, and OpenACC-parallel host).

tpls/desul/include/desul/atomics/Compare_Exchange_OpenACC.hpp
dalg24: Not sure I understand.  Is it a mix of things not compiling and places where you compared assembly code and saw benefit to use the CUDA atomics?  Could you elaborate?

tpls/desul/include/desul/atomics/Compare_Exchange_OpenACC.hpp
```diff
@@ -0,0 +1,269 @@
...
+  // Acquire a lock for the address
+  //while (!lock_address_openacc((void*)dest, scope)) {
+  //}
+  //device_atomic_thread_fence(MemoryOrderAcquire(), scope);
+  if (current_val == compare) {
+    *dest = value;
+    //device_atomic_thread_fence(MemoryOrderRelease(), scope);
+  }
+  //unlock_address_openacc((void*)dest, scope);
+  return current_val;
```
dalg24, line 261: Is that not something you can achieve with an acc atomic capture clause?

tpls/desul/include/desul/atomics/Compare_Exchange_OpenACC.hpp
```diff
@@ -0,0 +1,269 @@
...
+
+#pragma acc routine seq
+template <class T, class MemoryOrder, class MemoryScope>
+std::enable_if_t<std::is_floating_point<T>::value && (sizeof(T) == 4), T> device_atomic_compare_exchange(T* const dest, T compare, T value, MemoryOrder, MemoryScope) {
+  static_assert(sizeof(float) == 4,
+                "this function assumes an float is 32-bit");
+  float return_val = atomicCAS(reinterpret_cast<float*>(dest),
+                                      reinterpret_cast<float&>(compare),
+                                      reinterpret_cast<float&>(value));
+  return reinterpret_cast<T&>(return_val);
```
dalg24, line 202: Same question here, is there a good reason to treat floating point separately.

tpls/desul/include/desul/atomics/Lock_Based_Fetch_Op_OpenACC.hpp
```diff
@@ -0,0 +1,79 @@
...
+  // Acquire a lock for the address
+  while (!lock_address((void*)dest, scope)) {
+  }
+
+  device_atomic_thread_fence(MemoryOrderAcquire(), scope);
+  T return_val = *dest;
+  *dest = op.apply(return_val, val);
+  device_atomic_thread_fence(MemoryOrderRelease(), scope);
+  unlock_address((void*)dest, scope);
+  return return_val;
```
dalg24, line 45: This code is pretty much the same as the one in `Lock_Based_Fetch_Op_Host.hpp` except that we have functions prefixed with device_ instead of host_ right?

tpls/desul/include/desul/atomics/Thread_Fence_OpenACC.hpp
```diff
@@ -0,0 +1,23 @@
...
+#ifndef DESUL_ATOMICS_THREAD_FENCE_OPENACC_HPP_
+#define DESUL_ATOMICS_THREAD_FENCE_OPENACC_HPP_
+
+namespace desul {
+namespace Impl {
+
+#pragma acc routine seq
+template <class MemoryOrder, class MemoryScope>
+void device_atomic_thread_fence(MemoryOrder, MemoryScope) {
+}
```
dalg24, line 18: Is that a FIXME or this is the right thing for openacc atomics?

tpls/desul/include/desul/atomics/Fetch_Op_OpenACC.hpp
```diff
@@ -0,0 +1,491 @@
...
+#else
+template <class T, class R>
+using acc_enable_if_supported_arithmetic_type = std::enable_if_t<std::is_same_v<T,int> \
+	|| std::is_same_v<T,unsigned int> || std::is_same_v<T,unsigned long long> \
+	|| std::is_same_v<T,float>, R>;
+#endif
+#else
+template <class T, class R>
+using acc_enable_if_supported_arithmetic_type = std::enable_if_t<std::is_arithmetic<T>::value, R>;
+#endif
```
dalg24, line 29: OK but needs to be defined within the desul::impl:: namespace

tpls/desul/include/desul/atomics/Fetch_Op_OpenACC.hpp
```diff
@@ -0,0 +1,491 @@
...
+
+// clang-format off
+//<editor-fold desc="device_atomic_fetch_{add,sub,mul,div,lshift,rshift,mod,max,min,and,or,xor}">
+#pragma acc routine seq
+template <class T>
+acc_enable_if_supported_arithmetic_type<T,T> device_atomic_fetch_add(
+T* ptr, const T val, MemoryOrderRelaxed, MemoryScopeDevice) {
+  T tmp;
+#pragma acc atomic capture
+  { tmp = ptr[0]; ptr[0] += val; }
```
dalg24, line 42: I am not a big fan of `ptr[0]`, I would prefer `*ptr`

tpls/desul/include/desul/atomics/Fetch_Op_OpenACC.hpp
```diff
@@ -0,0 +1,491 @@
...
+namespace desul {
+namespace Impl {
+
+// clang-format off
+//<editor-fold desc="device_atomic_fetch_{add,sub,mul,div,lshift,rshift,mod,max,min,and,or,xor}">
+#pragma acc routine seq
+template <class T>
+acc_enable_if_supported_arithmetic_type<T,T> device_atomic_fetch_add(
+T* ptr, const T val, MemoryOrderRelaxed, MemoryScopeDevice) {
+  T tmp;
```
dalg24, line 40: Nitpick: For atomic_fetch_op I would prefer `tmp -> old`

tpls/desul/include/desul/atomics/Fetch_Op_OpenACC.hpp
```diff
@@ -0,0 +1,491 @@
...
+  if(acc_on_device(acc_device_not_host)) {
+    printf("Kokkos Error in device_atomic_fetch_max(): Not supported atomic "
+                  "operation in the OpenACC backend\n");
+  }
+  tmp = *ptr; *ptr = std::max(*ptr, val);
+  return tmp;
+}
+
+#ifdef KOKKOS_COMPILER_NVHPC
```
dalg24, line 142: Did you look at the assembly?  Is there any actual benefit from specializing for these rather than doing a regular acc atomic capture?

tpls/desul/include/desul/atomics/Fetch_Op_OpenACC.hpp
dalg24: Now I see that you do `*ptr` below in other functions so definitely change all occurrences of `ptr[0]`

tpls/desul/include/desul/atomics/Fetch_Op_OpenACC.hpp
```diff
@@ -0,0 +1,491 @@
...
+T* ptr, const unsigned int val, MemoryOrderRelaxed, MemoryScopeDevice) {
+  T tmp;
+#pragma acc atomic capture
+  { tmp = *ptr; *ptr = *ptr >> val; }
+  return tmp;
+}
+
+#pragma acc routine seq
+template <class T>
+T device_atomic_fetch_mod(
```
dalg24, line 118: Do I understand correctly that there is no path forward with the current acc specification to support atomic modulo, min, nor max?

tpls/desul/include/desul/atomics/Fetch_Op_OpenACC.hpp
```diff
@@ -0,0 +1,491 @@
...
+    T* const ptr, const T val, MemoryOrderRelaxed, MemoryScopeDevice) {
+#pragma acc atomic write
+    *ptr = val;
+}
+
+#pragma acc routine seq
+template <class T, class MemoryOrder>
+acc_enable_if_supported_arithmetic_type<T,void> device_atomic_store(
+    T* const ptr, const T val, MemoryOrder, MemoryScopeDevice) {
+  device_atomic_store(ptr, val, MemoryOrderRelaxed(), MemoryScopeDevice());
```
dalg24, line 465: Does that mean the atomic write is actually sequentially consistent?

tpls/desul/include/desul/atomics/Fetch_Op_OpenACC.hpp
```diff
@@ -0,0 +1,491 @@
...
+acc_enable_if_supported_arithmetic_type<T,void> device_atomic_store(
+    T* const ptr, const T val, MemoryOrder, MemoryScopeDevice) {
+  device_atomic_store(ptr, val, MemoryOrderRelaxed(), MemoryScopeDevice()); 
+}
+
+#pragma acc routine seq
+template <class T>
+acc_enable_if_supported_arithmetic_type<T,T> device_atomic_load(
+    const T* const ptr, MemoryOrderRelaxed, MemoryScopeDevice) {
+  T retval{};
```
dalg24, line 472:
```suggestion
  T retval;
```

tpls/desul/include/desul/atomics/Compare_Exchange_OpenACC.hpp
```diff
@@ -0,0 +1,269 @@
...
+
+namespace desul {
+namespace Impl {
+
+#pragma acc routine seq
+template <class T, class MemoryOrder, class MemoryScope>
+std::enable_if_t<!std::is_arithmetic<T>::value, T> device_atomic_exchange(T* dest, T value, MemoryOrder, MemoryScope scope) {
+  if (acc_on_device(acc_device_not_host)) {
+    printf("DESUL error in device_atomic_exchange(): Not supported atomic "
+                  "operation in the OpenACC backend\n");
```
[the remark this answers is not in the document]
seyonglee: The current OpenACC standard does not officially support any abort or print functions on the device, but existing OpenACC implementations unofficially support some print functions.

tpls/desul/include/desul/atomics/Compare_Exchange_OpenACC.hpp
seyonglee: Changed `KOKKOS_COMPILER_NVHPC` to `__NVCOMPILER`.

tpls/desul/include/desul/atomics/Fetch_Op_OpenACC.hpp
seyonglee: The current OpenACC standard supports atomic min and max operations only for Fortran, and atomic modulo is not supported at all.

tpls/desul/include/desul/atomics/Fetch_Op_OpenACC.hpp
seyonglee: No, OpenACC has a relaxed memory consistency model; the reason for having this implementation is that the current OpenACC atomics implementation uses the same `device_atomic_store()` API for both host and device.
Host may use memory consistency other than `MemoryOrderRelaxed`, but it is not possible for OpenACC to implement memory consistency models for atomic operations other than the relaxed consistency.
Therefore, the same relaxed consistency implementation is re-used for other Host-side `atomic_store()`.
(This is OK since the current OpenACC atomic implementation supports the host only in a serial execution mode.)

tpls/desul/include/desul/atomics/Fetch_Op_OpenACC.hpp
seyonglee: Renamed the variable `tmp` to `old` in all atomic_fetch_*() functions.

tpls/desul/include/desul/atomics/Fetch_Op_OpenACC.hpp
seyonglee: In this case, the specialization is not for the performance benefits. The OpenACC C/C++ standard does not support atomic min/max operations, but if targeting NVIDIA GPUs with NVHPC, we can use CUDA atomic min/max intrinsics.
This specialization is an NVHPC-specific extension.

tpls/desul/include/desul/atomics/Compare_Exchange_OpenACC.hpp
seyonglee: Using CUDA atomic intrinsics instead of the OpenACC atomic constructs is mainly due to compilation errors; the possible performance benefit is not verified.

tpls/desul/include/desul/atomics/Compare_Exchange_OpenACC.hpp
```diff
@@ -0,0 +1,269 @@
...
+#ifndef DESUL_CUDA_ARCH_IS_PRE_PASCAL
+
+#pragma acc routine seq
+template <class T, class MemoryOrder, class MemoryScope>
+std::enable_if_t<std::is_floating_point<T>::value && (sizeof(T) == 8), T> 
+	device_atomic_exchange(T* const dest, T value, MemoryOrder, MemoryScope) {
+  static_assert(sizeof(double) == 8,
+                "this function assumes an double is 64-bit");
+  double return_val = atomicExch(reinterpret_cast<double*>(dest),
+                                       reinterpret_cast<double&>(value));
```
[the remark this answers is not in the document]
seyonglee: The NVHPC OpenACC compiler supports `double atomicExch(double *, double)`, which is defined in `openacc.h`.

tpls/desul/include/desul/atomics/Compare_Exchange_OpenACC.hpp
seyonglee: If a single implementation is used for both integer and floating types, the NVHPC compiler fails in the integer-to-float casting code (return reinterpret_cast<T&>(return_val);) with the following error:
"parse invalid cast opcode for cast from 'i32' to 'float'"

tpls/desul/include/desul/atomics/Compare_Exchange_OpenACC.hpp
seyonglee: The `acc atomic capture` can not handle "compare" part in the compare-exchange operation.

tpls/desul/include/desul/atomics/Lock_Based_Fetch_Op_OpenACC.hpp
seyonglee: OpenACC does not support "explicit" thread fence operations, and thus this implementation is incomplete and will work only when executed on the host. (The current OpenACC atomic implementation supports the host only in the serial mode.)

tpls/desul/include/desul/atomics/Thread_Fence_OpenACC.hpp
seyonglee: Added FIXME comment.

tpls/desul/include/desul/atomics/Fetch_Op_OpenACC.hpp
seyonglee: Updated.

tpls/desul/include/desul/atomics/Fetch_Op_OpenACC.hpp
seyonglee: Moved into the `desul::Impl::` namespace.

seyonglee:
[OpenACC Atomic Construct Implementation Status (as of August 20, 2023)]
For NVHPC OpenACC compiler, the same device atomic implementations are used for both the host and device executions.
This is because OpenACC acc_on_device() API is not constexpr in NVHPC, where acc_on_device() is converted into 
compile-time constant in the back-end pass (but constexpr evaluation occurs in the NVHPC front-end). 
Therefore, some back-end passes can see the contents of the host atomic implementation and complain some GCC intrinsics.
For LLVM-Clacc compiler, acc_on_device() is constexpr, and thus appropriate implementations can be selected during the compiler front-end.

Below status update is about NVHPC implementations.
```
void atomic_thread_fence(MemoryOrder order, MemoryScope scope)
```
- Not supported except for MemoryScope == MemoryScopeCaller 
      (OpenACC standard does not provide explicit thread fence APIs.)

```
T atomic_exchange(T* dest, T val, MemoryOrder order, MemoryScope scope) 
```
- Supported data type T: std::is_arithmetic<T> && ((sizeof(T) == 4) || (sizeof(T) == 8)) 
    - Arithmetic types of other sizes and non-arithmetic types are non-atomically executed.
- Supported MemoryOrder types: MemoryOrderRelaxed
    - Other MemoryOrder types are non-atomically executed.
- Supported MemoryScope types: MemoryScopeDevice and MemoryScopeCore
    - Other MemoryScore types are non-atomically executed.

```
atomic_compare_exchange(T* dest, T cmp, T val, MemoryOrder order, MemoryScope scope)
```
- Supported data type T: std::is_arithmetic<T> && ((sizeof(T) == 4) || (sizeof(T) == 8)) 
    - Arithmetic types of other sizes and non-arithmetic types are non-atomically executed.
- Supported MemoryOrder types: MemoryOrderRelaxed
    - Other MemoryOrder types are non-atomically executed.
- Supported MemoryScope types: MemoryScopeDevice and MemoryScopeCore
    - Other MemoryScore types are non-atomically executed.

```
T atomic_fetch_add(T* const dest, const T val, MemoryOrder order, MemoryScope scope) 
T atomic_fetch_sub(T* const dest, const T val, MemoryOrder order, MemoryScope scope)
T atomic_fetch_max(T* const dest, const T val, MemoryOrder order, MemoryScope scope)
T atomic_fetch_min(T* const dest, const T val, MemoryOrder order, MemoryScope scope)
T atomic_fetch_mul(T* const dest, const T val, MemoryOrder order, MemoryScope scope)
T atomic_fetch_div(T* const dest, const T val, MemoryOrder order, MemoryScope scope)
T atomic_add_fetch(T* const dest, const T val, MemoryOrder order, MemoryScope scope)
T atomic_sub_fetch(T* const dest, const T val, MemoryOrder order, MemoryScope scope)
T atomic_max_fetch(T* const dest, const T val, MemoryOrder order, MemoryScope scope)
T atomic_min_fetch(T* const dest, const T val, MemoryOrder order, MemoryScope scope)
T atomic_mul_fetch(T* const dest, const T val, MemoryOrder order, MemoryScope scope)
T atomic_div_fetch(T* const dest, const T val, MemoryOrder order, MemoryScope scope)
void atomic_add(T* const dest, const T val, MemoryOrder order, MemoryScope scope)
void atomic_sub(T* const dest, const T val, MemoryOrder order, MemoryScope scope)
void atomic_mul(T* const dest, const T val, MemoryOrder order, MemoryScope scope)
void atomic_div(T* const dest, const T val, MemoryOrder order, MemoryScope scope)
void atomic_min(T* const dest, const T val, MemoryOrder order, MemoryScope scope)
void atomic_max(T* const dest, const T val, MemoryOrder order, MemoryScope scope)
T atomic_inc_fetch(T* const dest, MemoryOrder order, MemoryScope scope)
T atomic_dec_fetch(T* const dest, MemoryOrder order, MemoryScope scope)
T atomic_fetch_inc(T* const dest, MemoryOrder order, MemoryScope scope)
T atomic_fetch_dec(T* const dest, MemoryOrder order, MemoryScope scope)
void atomic_inc(T* const dest, MemoryOrder order, MemoryScope scope)
void atomic_dec(T* const dest, MemoryOrder order, MemoryScope scope)
```
- Supported data type T: int, unsigned int, unsigned long long, float, and double
    - Other arithmetic types of size 4 or 8 are also supported via generic lock-free fetch-op implementation in Lock_Free_Fetch_Op.hpp.
- Supported MemoryOrder types: MemoryOrderRelaxed
    - Other MemoryOrder types are non-atomically executed.
- Supported MemoryScope types: MemoryScopeDevice and MemoryScopeCore
    - Other MemoryScore types are non-atomically executed.

```
T atomic_fetch_add(T* const dest, const T val, MemoryOrder order, MemoryScope scope) 
T atomic_fetch_sub(T* const dest, const T val, MemoryOrder order, MemoryScope scope)
T atomic_fetch_max(T* const dest, const T val, MemoryOrder order, MemoryScope scope)
T atomic_fetch_min(T* const dest, const T val, MemoryOrder order, MemoryScope scope)
T atomic_fetch_mul(T* const dest, const T val, MemoryOrder order, MemoryScope scope)
T atomic_fetch_div(T* const dest, const T val, MemoryOrder order, MemoryScope scope)
T atomic_add_fetch(T* const dest, const T val, MemoryOrder order, MemoryScope scope)
T atomic_sub_fetch(T* const dest, const T val, MemoryOrder order, MemoryScope scope)
T atomic_max_fetch(T* const dest, const T val, MemoryOrder order, MemoryScope scope)
T atomic_min_fetch(T* const dest, const T val, MemoryOrder order, MemoryScope scope)
T atomic_mul_fetch(T* const dest, const T val, MemoryOrder order, MemoryScope scope)
T atomic_div_fetch(T* const dest, const T val, MemoryOrder order, MemoryScope scope)
void atomic_add(T* const dest, const T val, MemoryOrder order, MemoryScope scope)
void atomic_sub(T* const dest, const T val, MemoryOrder order, MemoryScope scope)
void atomic_mul(T* const dest, const T val, MemoryOrder order, MemoryScope scope)
void atomic_div(T* const dest, const T val, MemoryOrder order, MemoryScope scope)
void atomic_min(T* const dest, const T val, MemoryOrder order, MemoryScope scope)
void atomic_max(T* const dest, const T val, MemoryOrder order, MemoryScope scope)
T atomic_inc_fetch(T* const dest, MemoryOrder order, MemoryScope scope)
T atomic_dec_fetch(T* const dest, MemoryOrder order, MemoryScope scope)
T atomic_fetch_inc(T* const dest, MemoryOrder order, MemoryScope scope)
T atomic_fetch_dec(T* const dest, MemoryOrder order, MemoryScope scope)
void atomic_inc(T* const dest, MemoryOrder order, MemoryScope scope)
void atomic_dec(T* const dest, MemoryOrder order, MemoryScope scope)
```
- Supported data type T: int, unsigned int, unsigned long long, float, and double
    - Other arithmetic types of size 4 or 8 are also supported via generic lock-free fetch-op implementation in Lock_Free_Fetch_Op.hpp.
- Supported MemoryOrder types: MemoryOrderRelaxed
    - Other MemoryOrder types are non-atomically executed.
- Supported MemoryScope types: MemoryScopeDevice and MemoryScopeCore
    - Other MemoryScore types are non-atomically executed.

```
T atomic_fetch_and(T* const dest, const T val, MemoryOrder order, MemoryScope scope)
T atomic_fetch_or(T* const dest, const T val, MemoryOrder order, MemoryScope scope)
T atomic_fetch_xor(T* const dest, const T val, MemoryOrder order, MemoryScope scope)
T atomic_fetch_nand(T* const dest, const T val, MemoryOrder order, MemoryScope scope)
T atomic_fetch_lshift(T* const dest, const unsigned int val, MemoryOrder order, MemoryScope scope)
T atomic_fetch_rshift(T* const dest, const unsigned int val, MemoryOrder order, MemoryScope scope)
T atomic_and_fetch(T* const dest, const T val, MemoryOrder order, MemoryScope scope)
T atomic_or_fetch(T* const dest, const T val, MemoryOrder order, MemoryScope scope)
T atomic_xor_fetch(T* const dest, const T val, MemoryOrder order, MemoryScope scope)
T atomic_nand_fetch(T* const dest, const T val, MemoryOrder order, MemoryScope scope)
T atomic_lshift_fetch(T* const dest, const unsigned int val, MemoryOrder order, MemoryScope scope)
T atomic_rshift_fetch(T* const dest, const unsigned int val, MemoryOrder order, MemoryScope scope)
```
- Supported data type T: int, unsigned int, unsigned long long
    - Other integral types of size 4 or 8 are also supported via generic lock-free fetch-op implementation in Lock_Free_Fetch_Op.hpp.
- Supported MemoryOrder types: MemoryOrderRelaxed
    - Other MemoryOrder types are non-atomically executed.
- Supported MemoryScope types: MemoryScopeDevice and MemoryScopeCore
    - Other MemoryScore types are non-atomically executed.

```
T atomic_fetch_mod(T* const dest, const T val, MemoryOrder order, MemoryScope scope)
T atomic_mod_fetch(T* const dest, const T val, MemoryOrder order, MemoryScope scope)
T atomic_fetch_inc_mod(T* const dest, T val, MemoryOrder order, MemoryScope scope)
T atomic_fetch_dec_mod(T* const dest, T val, MemoryOrder order, MemoryScope scope)
```
- Supported data type T: integral types of size 4 or 8 are supported via generic lock-free fetch-op implementation in Lock_Free_Fetch_Op.hpp.
- Supported MemoryOrder types: MemoryOrderRelaxed
    - Other MemoryOrder types are non-atomically executed.
- Supported MemoryScope types: MemoryScopeDevice and MemoryScopeCore
    - Other MemoryScore types are non-atomically executed.

```
T atomic_load(const T* const dest, MemoryOrder order, MemoryScope scope)
```
- Supported data type T: int, unsigned int, unsigned long long, float, and double
    - Other arithmetic types of size 4 or 8 are also supported via generic lock-free fetch-op implementation in Lock_Free_Fetch_Op.hpp.
- Supported MemoryOrder types: MemoryOrderRelaxed (and MemoryOrderAcquire only for Host Serial backend)
    - Other MemoryOrder types are non-atomically executed.
- Supported MemoryScope types: MemoryScopeDevice and MemoryScopeCore
    - Other MemoryScore types are non-atomically executed.

```
void atomic_store(T* const dest, const T val, MemoryOrder order, MemoryScope scope)
```
- Supported data type T: int, unsigned int, unsigned long long, float, and double
    - Other arithmetic types of size 4 or 8 are also supported via generic lock-free fetch-op implementation in Lock_Free_Fetch_Op.hpp.
- Supported MemoryOrder types: MemoryOrderRelaxed (and MemoryOrderRelease only for Host Serial backend)
    - Other MemoryOrder types are non-atomically executed.
- Supported MemoryScope types: MemoryScopeDevice and MemoryScopeCore
    - Other MemoryScore types are non-atomically executed.

tpls/desul/include/desul/atomics/Compare_Exchange_OpenACC.hpp
seyonglee: Now changed to use acc atomic capture.

tpls/desul/include/desul/atomics/Fetch_Op_OpenACC.hpp
seyonglee: Even though OpenACC C/C++ standard does not support atomic min/max/mod operations, the Kokkos OpenACC backend supports them using either CUDA intrinsics (for NVHPC implementation) or generic lock-free fetch-op implementation in Lock_Free_Fetch_Op.hpp.

tpls/desul/include/desul/atomics/Compare_Exchange_OpenACC.hpp
```diff
@@ -0,0 +1,269 @@
...
+
+#pragma acc routine seq
+template <class T, class MemoryOrder, class MemoryScope>
+std::enable_if_t<std::is_floating_point<T>::value && (sizeof(T) == 4), T> 
+	device_atomic_exchange(T* const dest, T value, MemoryOrder, MemoryScope) {
+  static_assert(sizeof(float) == 4,
+                "this function assumes an float is 32-bit");
+  float return_val = atomicExch(reinterpret_cast<float*>(dest),
+                                       reinterpret_cast<float&>(value));
+  return reinterpret_cast<T&>(return_val);
```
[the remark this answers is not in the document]
dalg24: Compiler bug?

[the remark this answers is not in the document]
seyonglee: It seems to be a compile bug that happens only when `reinterpret_cast` is used together with `atomicExch()`.
By the way, the implementation of `device_atomic_exchange()` is changed not to use `atomicExch()`; thus this issue does not occur in `device_atomic_exchange()`. 
However, a similar compiler bug still occurs in `device_atomic_compare_exchange()` because this API uses both `reinterpret_cast` and `atomicCAS()` together.

tpls/desul/include/desul/atomics/Compare_Exchange_OpenACC.hpp
```diff
@@ -0,0 +1,200 @@
...
+}
+
+#endif
+
+#else
+
+#pragma acc routine seq
+template <class T, class MemoryOrder, class MemoryScope>
+T device_atomic_compare_exchange(T* dest, T compare, T value, MemoryOrder, MemoryScope scope) {
```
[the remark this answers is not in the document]
seyonglee: The current OpenACC stand does not support compare-and-exchange atomic operations at all; that's why CUDA intrinsics are used for the NVHPC case. 
For the LLVM-Clacc case, I have not yet checked what kind of alternative options are available. One possibility would be to use OpenMP-specific implementation for the Clacc case since Clacc itself is built on top of LLVM OpenMP implementation. I will dig into this direction in the next PR after this PR is merged.

tpls/desul/include/desul/atomics/Compare_Exchange_OpenACC.hpp
```diff
@@ -0,0 +1,200 @@
...
+  T current_val = *dest;
+  if (acc_on_device(acc_device_not_host)) {
+    printf("DESUL error in device_atomic_compare_exchange(): Not supported atomic "
+                  "operation in the OpenACC backend\n");
+  }
```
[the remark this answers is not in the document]
seyonglee: Updated.

tpls/desul/include/desul/atomics/Compare_Exchange_OpenACC.hpp
```diff
@@ -0,0 +1,200 @@
...
+	&& (std::is_same_v<MemoryScope,MemoryScopeDevice>                 \
+	|| std::is_same_v<MemoryScope,MemoryScopeCore>), T>
+device_atomic_compare_exchange(T* const dest, T compare, T value, MemoryOrderRelaxed, MemoryScope) {
+  static_assert(sizeof(float) == 4,
+                "this function assumes an float is 32-bit");
+  float return_val = atomicCAS(reinterpret_cast<float*>(dest),
+                                      reinterpret_cast<float&>(compare),
+                                      reinterpret_cast<float&>(value));
+  return reinterpret_cast<T&>(return_val);
+}
```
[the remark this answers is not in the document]
seyonglee: Updated.

crtrott (changes requested):
This is supposed to be stale

PR #5962, opened 2023-03-07
Use CombinedReducer in HostIterateTile

masterleinad:
Part of #5908. Addressing https://github.com/kokkos/kokkos/pull/5874#discussion_r1113620098. In particular, this avoids storing the functor twice since recent changes in HostIterateTile made it store the functor to avoid ICEs for ClassIc Intel Compilers.


Rombur (approved):

dalg24 (commented):
Expand the description.
Also this is not helping with code readability.

masterleinad:
@dalg24 I expanded the description.

dalg24 (commented):
Is that (the not storing twice) something that could/should be tested?

crtrott (approved):

crtrott:
Failing tests system issue with HIP.

PR #5964, opened 2023-03-07
Enable Cuda nvcc -extended-lambda unconditionally

PhilMiller:
Nvidia stabilized extended lambdas with the release of CUDA Toolkit 12.0. So, there's no longer a need or benefit to this feature being optional.

In cases yet to be fully characterized, a lambda meant to be invoked on the host (e.g. in a combined OpenMP+CUDA build) may have to be spelled with `[=]` without a `__device__` annotation, rather than using `KOKKOS_LAMBDA`.


PhilMiller:
I've willfully neglected any sort of deprecation or notice about the option not having an effect anymore. I think we should actually leave it that way for at least a few release cycles so that the people still trying to use 3.7.x interchangeably with 4.y.z don't have to see noise about it.

dalg24 (dismissed):
Title line is inaccurate, this is not about enabling by default.
We already do that since #5580 
The changes proposed here are about making impossible to not have the `--extented-lambda` in the compile line which I am not convinced is necessary a good idea

PhilMiller:
Title line fixed.

What are the circumstances in which we'd suggest a user disable extended lambda support?

dalg24:
The merit of what is suggested here is that it considerably simplifies our internal logic but we need to decide whether it is ok to take control away from the end user.

PhilMiller:
So, really this is more a question for Slack `#general` whether anyone sets it to off. I'll post there asking anyone who does to comment here with what app/platform and why

dalg24:
> So, really this is more a question for Slack `#general` whether anyone sets it to `OFF`. I'll post there asking anyone who does to comment here with what app/platform and why

No complaints so far neither here nor on Slack.  We decided to proceed with it.
*`-DKokkos_ENABLE_CUDA_LAMBDA=OFF` -> hard error
*`-DKokkos_ENABLE_CUDA_LAMBDA=ON` -> deprecation warning rather info (status)
  * VOTED 8 VS 3 (Feel free to comment if you have an issue with this)

Use case where `KOKKOS_LAMBDA` is being used with `Kokkos_ENABLE_CUDA_LAMBDA=OFF` and that has code that only works on the host does not sound like something we absolutely need to accommodate.

masterleinad (commented):
You 'll need
```diff
diff --git a/cmake/kokkos_arch.cmake b/cmake/kokkos_arch.cmake
index 83b4f4115..343573a89 100644
--- a/cmake/kokkos_arch.cmake
+++ b/cmake/kokkos_arch.cmake
@@ -176,8 +176,8 @@ IF(KOKKOS_CXX_COMPILER_ID STREQUAL NVIDIA)
   GLOBAL_APPEND(KOKKOS_CUDA_OPTIONS "-Wext-lambda-captures-this")
 ENDIF()
 
-IF(DEFINED KOKKOS_ENABLE_CUDA_LAMBDA)
-  IF(KOKKOS_ENABLE_CUDA_LAMBDA)
+IF(DEFINED Kokkos_ENABLE_CUDA_LAMBDA)
+  IF(Kokkos_ENABLE_CUDA_LAMBDA)
     MESSAGE(DEPRECATION "CUDA extended lambda support is now always enabled. The option Kokkos_ENABLE_CUDA_LAMBDA will be removed")
   ELSE()
     MESSAGE(FATAL_ERROR "Support for disabling CUDA extended lambdas has been removed. Please unset Kokkos_ENABLE_CUDA_LAMBDA, or see #5964 if this is necessary for your application")
diff --git a/cmake/kokkos_enable_options.cmake b/cmake/kokkos_enable_options.cmake
index 7d8026989..1305bc61f 100644
--- a/cmake/kokkos_enable_options.cmake
+++ b/cmake/kokkos_enable_options.cmake
@@ -74,7 +74,6 @@ ELSEIF (KOKKOS_ENABLE_CUDA)
 ELSE()
   SET(CUDA_LAMBDA_DEFAULT OFF)
 ENDIF()
-KOKKOS_ENABLE_OPTION(CUDA_LAMBDA ${CUDA_LAMBDA_DEFAULT} "Whether to activate experimental lambda features")
 IF (Trilinos_ENABLE_Kokkos)
   SET(COMPLEX_ALIGN_DEFAULT OFF)
 ELSE()
```
or similar. If you define `Kokkos_ENABLE_CUDA_LAMDA` in our usual way `KOKKOS_ENABLE_CUDA_LAMDA` is also defined. Further, users should set `Kokkos_ENABLE_CUDA_LAMBDA` and not `KOKKOS_ENABLE_CUDA_LAMBDA`. Therefore, the check should consider ``Kokkos_ENABLE_CUDA_LAMBDA` and not `KOKKOS_ENABLE_CUDA_LAMBDA` since the latter isn't defined anymore internally after removing the option.

PhilMiller:
As I'm going through and clearing stuff out more thoroughly, I'm thinking we maybe leave (effectively) `#define KOKKOS_ENABLE_CUDA_LAMBDA 1` hard-coded, in case downstream code was for some reason checking that.

PhilMiller:
> As I'm going through and clearing stuff out more thoroughly, I'm thinking we maybe leave (effectively) `#define KOKKOS_ENABLE_CUDA_LAMBDA 1` hard-coded, in case downstream code was for some reason checking that.

Also, with the commit I just pushed, we'll (hopefully) be able to see testing pass without the internal code cleanup.

masterleinad:
> As I'm going through and clearing stuff out more thoroughly, I'm thinking we maybe leave (effectively) `#define KOKKOS_ENABLE_CUDA_LAMBDA 1` hard-coded, in case downstream code was for some reason checking that.

I would guard it with `KOKKOS_ENABLE_DEPRECATED_CODE_4`.

masterleinad (approved):
Looks good to me.

config/yaml/volta.yaml
```diff
@@ -1,4 +1,4 @@
 packages:
  kokkos:
-  variants: +cuda +openmp +volta70 +cuda_lambda +wrapper ^cuda@10.1
+  variants: +cuda +openmp +volta70 +wrapper ^cuda@10.1
```
crtrott, line 3: how is this supposed to work. Either we need cuda_lambda because its an older Kokkos ro we need cuda@11.x

crtrott (changes requested):
The spack changes are a bit incoherent since they apply to an older Kokkos version. So if we remove the cuda_lambda we also need to switch out the Kokkos version

PhilMiller:
Ah, sorry, I missed that the Spack bits specified an older version. I think I should just revert those charges for now, and removal of the flag can come with an update to the package

PhilMiller:
I've reverted the Spack-related changes

PhilMiller:
The `view_mapping` test failures in https://cloud.cees.ornl.gov/jenkins-ci/blue/rest/organizations/jenkins/pipelines/Kokkos/runs/12719/nodes/56/steps/155/log/?start=0 look spurious, but I note that this appeared in the output:

```
1: [ RUN      ] serial.mdrange_parallel_reduce_primitive_types
1: /var/jenkins/workspace/Kokkos/core/unit_test/TestMDRangeReduce.hpp:53: Skipped
1: Skipped ENABLE_CUDA_LAMBDA
1: [  SKIPPED ] serial.mdrange_parallel_reduce_primitive_types (0 ms)

...

4: [ RUN      ] cuda.mdrange_parallel_reduce_primitive_types
4: /var/jenkins/workspace/Kokkos/core/unit_test/TestMDRangeReduce.hpp:53: Skipped
4: Skipped ENABLE_CUDA_LAMBDA
4: [  SKIPPED ] cuda.mdrange_parallel_reduce_primitive_types (0 ms)

...


18: [ RUN      ] CUDA.IncrTest_01_execspace
18: Device Execution Space:
18:   KOKKOS_ENABLE_CUDA: yes
18: Cuda Options:
18:   KOKKOS_ENABLE_CUDA_LAMBDA: no
18:   KOKKOS_ENABLE_CUDA_RELOCATABLE_DEVICE_CODE: no
18:   KOKKOS_ENABLE_CUDA_UVM: no
18:   KOKKOS_ENABLE_CXX11_DISPATCH_LAMBDA: yes
18: 
18: Cuda Runtime Configuration:
18: macro  KOKKOS_ENABLE_CUDA      : defined
18: macro  CUDA_VERSION          = 11060 = version 11.6
18: Kokkos::Cuda[ 0 ] NVIDIA A100-PCIE-40GB capability 8.0, Total Global Memory: 39.43 G, Shared Memory per Block: 48 K : Selected
18: Cuda
18: [       OK ] CUDA.IncrTest_01_execspace (0 ms)
```

So, something's not quite right in the header changes

PhilMiller:
Ok, since we're guarding the define behind `KOKKOS_ENABLE_DEPRECATED_CODE_4`, it's not set in that build. I should update the test in question to not be conditional.

PhilMiller:
The `view_mapping` failure is not spurious, and the comment may actually indicate a reason not to integrate this:

```
#if !defined(KOKKOS_ENABLE_CUDA) || !defined(KOKKOS_ENABLE_CUDA_LAMBDA)
    // Cannot launch host lambda when CUDA lambda is enabled.                                                                                                                                                                                                                

    using host_exec_space =
        typename Kokkos::Impl::HostMirror<Space>::Space::execution_space;

    int errors = 0;
    Kokkos::parallel_reduce(
        Kokkos::RangePolicy<host_exec_space>(0, 10),
        KOKKOS_LAMBDA(int, int& e) {
          // an unmanaged copy.  When the parallel dispatch accepts a move for                                                                                                                                                                                               
          // the lambda, this count should become 1.                                                                                                                                                                                                                         

          if (a.use_count() != 2) ++e;
          V x = a;
          if (a.use_count() != 2) ++e;
          if (x.use_count() != 2) ++e;
        },
        errors);
    ASSERT_EQ(errors, 0);
#endif  // #if !defined( KOKKOS_ENABLE_CUDA_LAMBDA )
```

That seems a niche use-case, but maybe not one we can drop entirely (for combined OpenMP-CUDA builds)

masterleinad:
Needs rebasing.

PhilMiller:
Per meeting:
* Add changelog entry
* Guard `view_mapping` test with suitable `FIXME_NVCC` comment
* Further test enablement to happen in follow-ups

PhilMiller:
There was a seemingly spurious failure in `sycl.partitioning_by_args`
```
4: [ RUN      ] sycl.partitioning_by_args
4:   sum1
4:     Which is: 1132494848
4:   sum2
4:     Which is: 1124139008
4: [  FAILED  ] sycl.partitioning_by_args (15 ms)
```

Otherwise waiting on HIP jobs to run, but all the CUDA jobs passed.

crtrott (approved):

PR #5965, opened 2023-03-08
Desul atomics: fix bug in `desul::Impl::numeric_limits_max<uint64_t>` value

dalg24:
Use modulo arithmetic instead of wrong hex number

### DIFF
--- a/tpls/desul/include/desul/atomics/Common.hpp
+++ b/tpls/desul/include/desul/atomics/Common.hpp
@@ -83,11 +83,11 @@ struct numeric_limits_max;
 
 template <>
 struct numeric_limits_max<uint32_t> {
-  static constexpr uint32_t value = 0xffffffffu;
+  static constexpr uint32_t value = -1;
 };
 template <>
 struct numeric_limits_max<uint64_t> {
-  static constexpr uint64_t value = 0xfffffffflu;
+  static constexpr uint64_t value = -1;
 };
 
 constexpr bool atomic_always_lock_free(std::size_t size) {


tpls/desul/include/desul/atomics/Common.hpp
```diff
@@ -83,11 +83,11 @@ struct numeric_limits_max;
 
 template <>
 struct numeric_limits_max<uint32_t> {
-  static constexpr uint32_t value = 0xffffffffu;
+  static constexpr uint32_t value = -1;
 };
 template <>
 struct numeric_limits_max<uint64_t> {
-  static constexpr uint64_t value = 0xfffffffflu;
```
dalg24, line 90: The right value would have been `0xffffffffffffffff`.  The suffixes are pointless here.

tpls/desul/include/desul/atomics/Common.hpp
```diff
@@ -83,11 +83,11 @@ struct numeric_limits_max;
 
 template <>
 struct numeric_limits_max<uint32_t> {
-  static constexpr uint32_t value = 0xffffffffu;
+  static constexpr uint32_t value = -1;
```
PhilMiller, line 86: An alternative expression that makes the 'all ones' nature totally clear, and avoids pedantic details about types and numeric behavior, would be
```suggestion
  static constexpr uint32_t value = ~static_cast<uint32_t>(0);
```

crtrott: I like this.

crtrott: This being Phil's suggestion.

dalg24: Discussion needs to happen on the Desul side and until proof of the contrary no warnings are being emitted with the code I proposed and modulo arithmetic is as easy as it gets so there is no need to try to be clever.

crtrott (approved):

tpls/desul/include/desul/atomics/Common.hpp
crtrott: yeah its fine leave the -1

Rombur (approved):

PR #5967, opened 2023-03-08
Bit manipulation: implement `byteswap`

dalg24:
Follow-up for #4577 and #5907

### DIFF
--- a/core/src/Kokkos_BitManipulation.hpp
+++ b/core/src/Kokkos_BitManipulation.hpp
@@ -31,19 +31,24 @@ KOKKOS_FUNCTION constexpr T byteswap_fallback(T x) {
   } else {
     using U = std::make_unsigned_t<T>;
 
-    size_t diff = CHAR_BIT * (sizeof(T) - 1);
+    size_t shift = CHAR_BIT * (sizeof(T) - 1);
 
-    U mask1 = static_cast<unsigned char>(~0);
-    U mask2 = mask1 << diff;
-    U val   = x;
+    U lo_mask = static_cast<unsigned char>(~0);
+    U hi_mask = lo_mask << shift;
+
+    U val = x;
+
     for (size_t i = 0; i < sizeof(T) / 2; ++i) {
-      U byte1 = val & mask1;
-      U byte2 = val & mask2;
+      U lo_val = val & lo_mask;
+      U hi_val = val & hi_mask;
 
-      val = (val ^ byte1 ^ byte2 ^ (byte1 << diff) ^ (byte2 >> diff));
-      mask1 <<= CHAR_BIT;
-      mask2 >>= CHAR_BIT;
-      diff -= 2 * CHAR_BIT;
+      val = (val & ~lo_mask) | (hi_val >> shift);
+      val = (val & ~hi_mask) | (lo_val << shift);
+
+      lo_mask <<= CHAR_BIT;
+      hi_mask >>= CHAR_BIT;
+
+      shift -= 2 * CHAR_BIT;
     }
     return val;
   }
--- a/core/src/Kokkos_BitManipulation.hpp
+++ b/core/src/Kokkos_BitManipulation.hpp
@@ -26,9 +26,7 @@ namespace Kokkos::Impl {
 
 template <class T>
 KOKKOS_FUNCTION constexpr T byteswap_fallback(T x) {
-  if constexpr (sizeof(T) == 1) {
-    return x;
-  } else {
+  if constexpr (sizeof(T) > 1) {
     using U = std::make_unsigned_t<T>;
 
     size_t shift = CHAR_BIT * (sizeof(T) - 1);
@@ -52,11 +50,7 @@ KOKKOS_FUNCTION constexpr T byteswap_fallback(T x) {
     }
     return val;
   }
-
-  // Workaround for buggy nvcc 11.2 compiler warning: missing return statement
-  // at end of non-void function
-  //
-  // Note: this return statement is unreachable
+  // sizeof(T) == 1
   return x;
 }
 
--- a/core/src/Kokkos_BitManipulation.hpp
+++ b/core/src/Kokkos_BitManipulation.hpp
@@ -19,8 +19,35 @@
 
 #include <Kokkos_Macros.hpp>
 #include <Kokkos_NumericTraits.hpp>
+#include <climits>  // CHAR_BIT
+#include <type_traits>
 
 namespace Kokkos::Impl {
+
+template <class T>
+KOKKOS_FUNCTION constexpr T byteswap_fallback(T x) {
+  if constexpr (sizeof(T) == 1) {
+    return x;
+  } else {
+    using U = std::make_unsigned_t<T>;
+
+    size_t diff = CHAR_BIT * (sizeof(T) - 1);
+
+    U mask1 = static_cast<unsigned char>(~0);
+    U mask2 = mask1 << diff;
+    U val   = x;
+    for (size_t i = 0; i < sizeof(T) / 2; ++i) {
+      U byte1 = val & mask1;
+      U byte2 = val & mask2;
+
+      val = (val ^ byte1 ^ byte2 ^ (byte1 << diff) ^ (byte2 >> diff));
+      mask1 <<= CHAR_BIT;
+      mask2 >>= CHAR_BIT;
+      diff -= 2 * CHAR_BIT;
+    }
+    return val;
+  }
+}
 
 template <class T>
 KOKKOS_FUNCTION constexpr int countl_zero_fallback(T x) {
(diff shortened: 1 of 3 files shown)


Rombur:
Failing with
```
/__w/kokkos/kokkos/core/unit_test/TestBitManipulation.cpp: In function 'constexpr decltype (Kokkos::byteswap(x)) test_byteswap(T)':
/__w/kokkos/kokkos/core/unit_test/TestBitManipulation.cpp:443:55: error: comparison of integer expressions of different signedness: 'std::enable_if_t<true, long int>' {aka 'long int'} and 'long unsigned int' [-Werror=sign-compare]
  443 |   static_assert(byteswap<int64_t>(0x123456789abcdef0) == 0xf0debc9a78563412);
      |                 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~~~~~~
```

core/src/Kokkos_BitManipulation.hpp
```diff
@@ -19,9 +19,36 @@
...
+    for (size_t i = 0; i < sizeof(T) / 2; ++i) {
+      U byte1 = val & mask1;
+      U byte2 = val & mask2;
+
+      val = (val ^ byte1 ^ byte2 ^ (byte1 << diff) ^ (byte2 >> diff));
+      mask1 <<= CHAR_BIT;
+      mask2 >>= CHAR_BIT;
+      diff -= 2 * CHAR_BIT;
+    }
+    return val;
```
masterleinad, line 48:
```suggestion
  using U = std::make_unsigned_t<T>;
    U val = x; 
    std::size_t diff = CHAR_BIT * (sizeof(T) - 1);
    std::size_t byte_mask = static_cast<unsigned char>(~0);
    for (std::size_t i =0; i< sizeof(T)/2; ++i) {
      U low_mask  = byte_mask << i*CHAR_BIT;
      U high_mask = byte_mask << (sizeof(T)-1-i)*CHAR_BIT;
      U low_value = val & low_mask;
      U high_value = val & high_mask;
      val = (val & ~low_mask)  | (high_value >> diff);
      val = (val & ~high_mask) | (low_value << diff);
      diff -= 2* CHAR_BIT;
    }
    return val;
```
is clearer to me and seems to be better recognized by at least `clang`, see https://godbolt.org/z/zb4nEMPYv. I'm not quite sure how
```
val = (val ^ byte1 ^ byte2 ^ (byte1 << diff) ^ (byte2 >> diff));
```
works
```
val = (val & ~low_mask)  | (high_value >> diff);
```
clears the low bits first and then sets them to the desired value.

nliber: I do understand how the @dalg24 version works.  While it may not look like it, this is just applying the regular xor trick for swapping two values.

`byte1 ^ (byte2 >> diff)` results in the low bits that need changing.

`val ^= byte1 ^ (byte2 >> diff)` changes the low bits.

The other two terms do the same thing for the high bits.

That being said, I also find the @masterleinad version easier to understand.

And if you are looking for more possible implementations, I believe this is the one from the paper proposing `byteswap`: https://github.com/bruxisma/byteswap/blob/master/include/ixm/byteswap.hpp

I'll approve it, but if the implementation changes, that's okay too.

nliber (approved):

dalg24:
Fixed the compile-only test and proposed an alternate implementation for `byteswap_fallback`.  Didn't mean to do both in one commit put plan on squash-merge anyway.
@masterleinad let me know if that version is acceptable.

masterleinad (approved):

dalg24:
Seeing this (unrelated) SYCL failure again
```
37: [ RUN      ] sycl.dynamic_view
37: /var/jenkins/workspace/Kokkos/containers/unit_tests/TestDynamicView.hpp:269: Failure
37: Expected equality of these values:
37:   new_result_sum
37:     Which is: 7.89488e+07
37:   (value_type)(da_size * (da_size - 1) / 2)
37:     Which is: 7.93737e+07
37: [  FAILED  ] sycl.dynamic_view (36 ms)
```
but there seems to be an actual issue in the `CUDA-11.0-NVCC-RDC` build
```
[  7%] Building CXX object core/unit_test/CMakeFiles/KokkosCore_TestCompileOnly.dir/TestBitManipulation.cpp.o
/var/jenkins/workspace/Kokkos/core/unit_test/TestBitManipulation.cpp(466): error: static assertion failed

1 error detected in the compilation of "/var/jenkins/workspace/Kokkos/core/unit_test/TestBitManipulation.cpp".
core/unit_test/CMakeFiles/KokkosCore_TestCompileOnly.dir/build.make:134: recipe for target 'core/unit_test/CMakeFiles/KokkosCore_TestCompileOnly.dir/TestBitManipulation.cpp.o' failed
make[2]: *** [core/unit_test/CMakeFiles/KokkosCore_TestCompileOnly.dir/TestBitManipulation.cpp.o] Error 1
CMakeFiles/Makefile2:2140: recipe for target 'core/unit_test/CMakeFiles/KokkosCore_TestCompileOnly.dir/all' failed
make[1]: *** [core/unit_test/CMakeFiles/KokkosCore_TestCompileOnly.dir/all] Error 2

**Comment by dalg24** [comment]
```C++
  static_assert(byteswap<int64_t>(0x123456789abcdef0) ==
                static_cast<int64_t>(0xf0debc9a78563412));
```
Static assertion fails with NVCC up to version 11.2.2 and passes from 11.3.0 https://godbolt.org/z/qrP3rq7Te

core/unit_test/TestBitManipulation.cpp
```diff
@@ -463,8 +463,14 @@ constexpr bool test_byteswap2() {
   static_assert(byteswap<int8_t>(0x12) == 0x12);
   static_assert(byteswap<int16_t>(0x1234) == 0x3412);
   static_assert(byteswap<int32_t>(0x12345678) == 0x78563412);
+
+// FIXME_NVCC This case is overflowing int65_t and the conversion from uint64_t
+// to int64_t here is implementation-defined. This seems to be the reason for
+// the check to fail for nvcc with CUDA <= 11.3.0
```
dalg24, line 469: The comment is inaccurate. For one I donor think it is an overflow, it is a compiler bug.  It is not supposed to be a conversion, it is specifying the type and the underlying bit representation.
And it works with 11.3.0 so it is version less (not version less equal)

crtrott:
I pushed another attempt on top (without clang-format) to see if the github actions pass.

core/src/Kokkos_BitManipulation.hpp
```diff
@@ -187,6 +227,35 @@ namespace Kokkos::Impl {
...
+  } else if constexpr (sizeof(T) == 4) {
+    return __builtin_bswap32(x);
+  } else if constexpr (sizeof(T) == 8) {
+    return __builtin_bswap64(x);
+  } else if constexpr (sizeof(T) == 16) {
+#if __has_builtin(__builtin_bswap128)
+    return __builtin_bswap128(x);
+#else
+    return (__builtin_bswap64(x >> 64) |
+            (static_cast<T>(__builtin_bswap64(x)) << 64));
```
PhilMiller, line 251: This is missing the `make_unsigned` intermediary to avoid UB shifting across the sign bit.

nliber (dismissed):

core/unit_test/TestBitManipulation.cpp
```diff
@@ -429,4 +429,53 @@ constexpr auto test_bit_width(UInt x) -> decltype(Kokkos::bit_width(x)) {
...
+
+constexpr bool test_byteswap2() {
+  using Kokkos::byteswap;
+
+  static_assert(byteswap<int8_t>(0x12) == 0x12);
+  static_assert(byteswap<int16_t>(0x1234) == 0x3412);
+  static_assert(byteswap<int32_t>(0x12345678) == 0x78563412);
+
+  static_assert(byteswap<int64_t>(0x123456789abcdef0ll) ==
+                0xf0debc9a78563412ll);
```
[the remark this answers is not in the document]
PhilMiller: Having learned about the standardized macros `INT64_C` etc al, I'm inclined to favor their use over the suffixes, since the macros specify a size rather than the annoying minimums of the primitive types.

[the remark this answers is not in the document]
nliber: @PhilMiller I'll make that change.  (While I knew there had to be some mechanism for it, I just learned about the macros two days ago.)

nliber:
Retest this please

core/src/Kokkos_BitManipulation.hpp
```diff
@@ -19,9 +19,47 @@
...
+      shift -= 2 * CHAR_BIT;
+    }
+    return val;
+  }
+
+  // Workaround for buggy nvcc 11.2 compiler warning: missing return statement
+  // at end of non-void function
+  //
+  // Note: this return statement is unreachable
+  return x;
```
nmm0, line 60: Could just do
```cpp
if constexpr (sizeof(T) != 1) {
  // ...
  return val;
}

return x;
```

nmm0 (approved):
Looks good to me!

core/src/Kokkos_BitManipulation.hpp
dalg24: I applied your suggestion

core/src/Kokkos_BitManipulation.hpp
PhilMiller: Then I think we may need to do some ugly `reinterpret_cast<uint64_t>(&x)` business, or really technically, to avoid type punning UB, 
```c++
uint64_t as64[2];
memcpy(as64, &x, 16);
// ... bswap64 on those
```

PR #5968, opened 2023-03-08
Drop Kokkos_ARCH_NATIVE=ON because it breaks with ccache

PhilMiller:
Github CI runners seem to have various CPU capabilities, probably differing on AVX-512 support. This causes many jobs to fail with `-march=native` if they get a cached object that doesn't match the current runner.

### DIFF
--- a/.github/workflows/continuous-integration-workflow-hpx.yml
+++ b/.github/workflows/continuous-integration-workflow-hpx.yml
@@ -69,7 +69,6 @@ jobs:
             -DCMAKE_CXX_COMPILER=clang++ \
             -DCMAKE_CXX_FLAGS="-Werror" \
             -DHPX_ROOT=$PWD/../../hpx/install \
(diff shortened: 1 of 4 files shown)


masterleinad (approved):

PhilMiller:
Of course, CI on *this* PR is going to take the full build duration because nothing will be cached with the changed compiler flag

cz4rs (approved):
Looks good to me.

nmm0 (approved):
Looks good to me, hopefully this fixes the random failures.

PhilMiller:
https://github.com/ccache/ccache/issues/824 Surprising that this wasn't recognized upstream sooner. I guess it's only in the last few years that a plethora of x86_64 systems with different instruction set capabilities have been widely deployed in testing environments, and those instructions being generated in common code as opposed to just library specializations or through intrinsics.

PR #5971, opened 2023-03-08
Fix reducer result check for Serial+HPX ParallelReduce

masterleinad:
Part of #5908. We have the same check for `Threads` and `OpenMP` already.


crtrott (approved):

crtrott:
There was a machine failure with HIP, but this doesn't touch that so the failure is no issue.

tcclevenger (approved):

Rombur (approved):

PR #5977, opened 2023-03-10
Allow NVCC 12 to compile using C++20 flag

j8asic:


bin/nvcc_wrapper
```diff
@@ -338,6 +338,24 @@ do
...
+      warn_std_flag
+      shared_args=${shared_args/ $std_flag/}
+    fi
+    # NVCC only has C++20 from version 12 on
+    cuda_main_version=$([[ $(${nvcc_compiler} --version) =~ V([0-9]+) ]] && echo ${BASH_REMATCH[1]})
+    if [ ${cuda_main_version} -lt 12 ]; then
+      fallback_std_flag="-std=c++14"
+      # this is hopefully just occurring in a downstream project during CMake feature tests
+      # we really have no choice here but to accept the flag and change  to an accepted C++ standard
+      echo "nvcc_wrapper does not accept standard flags $1 since partial standard flags and standards after C++14 are not supported. nvcc_wrapper will use $fallback_std_flag instead. It is undefined behavior to use this flag. This should only be occurring during CMake configuration."
```
PhilMiller, line 352: That should be C++17 in the text, the fallback flag, and the comment in the preceding block dealing with partial standards.

PhilMiller: Question for our core folks: would it make sense to fall forward from a partial standard flag to the corresponding full standard if the target version supports it? I.e. convert `-std=c++2a` to 20? I know there are weird interactions here with CMake behavior and such.

dalg24:
OK to test

dalg24 (approved):
Good enough.  The minimum standard fallbacks and comments need updating multiple places and can be handle as a follow-up.

crtrott (approved):

crtrott:
Candidate for 4.0.01 not sure that it makes sense for 3.7.02 since the whole purpose of maintaining 3.7.02 is C++14 support ...

PR #5963, opened 2023-03-07
Fix partition_master test

masterleinad:
`impl_thread_pool_size` is not `static` anymore. The test is only triggered with `KOKKOS_ENABLE_DEPRECATED_CODE_3=ON`.

edit: This test got broken in https://github.com/kokkos/kokkos/pull/5836 which is not included in any release. Hence, there is no need to cherry-pick this unless #5836 is cherry-picked and even then only to be able to run the tests with `KOKKOS_ENABLE_DEPRECATED_CODE_3=ON`.

PR #5978, opened 2023-03-10
Enable OpenMP in CUDA-11.0-NVCC-RDC to test DEPRECATED_CODE_3=ON

masterleinad:
Follow-up to #5963. A lot of functionality guarded by `KOKKOS_ENABLE_DEPRECATED_CODE_3` is related to OpenMP, in particular `partition_master`, but also other changes from static to non-static member functions. We don't have a CI configuration with OpenMP and  `KOKKOS_ENABLE_DEPRECATED_CODE_3=ON`, though.
~~`CUDA-11.0-NVCC-RDC` currently uses `KOKKOS_ENABLE_DEPRECATED_CODE_3=ON` but explicitly disables `OpenMP` (although it configures OpenMP in environment variables). Let's see if this works now or if we have to pick a different build.~~
This pull request enables `DEPRECATED_CODE_3` for the `gcc-8.4.0` jenkins build.

### DIFF
--- a/.jenkins
+++ b/.jenkins
@@ -384,7 +384,7 @@ pipeline {
                                 -DCMAKE_CXX_STANDARD=17 \
                                 -DKokkos_ARCH_NATIVE=ON \
                                 -DKokkos_ENABLE_COMPILER_WARNINGS=ON \
-                                -DKokkos_ENABLE_OPENMP=OFF \
+                                -DKokkos_ENABLE_OPENMP=ON \
                                 -DKokkos_ENABLE_CUDA=ON \
                                 -DKokkos_ENABLE_CUDA_LAMBDA=OFF \
                                 -DKokkos_ENABLE_CUDA_UVM=ON \
--- a/.jenkins
+++ b/.jenkins
@@ -384,7 +384,7 @@ pipeline {
                                 -DCMAKE_CXX_STANDARD=17 \
                                 -DKokkos_ARCH_NATIVE=ON \
                                 -DKokkos_ENABLE_COMPILER_WARNINGS=ON \
-                                -DKokkos_ENABLE_OPENMP=ON \
+                                -DKokkos_ENABLE_OPENMP=OFF \
                                 -DKokkos_ENABLE_CUDA=ON \
                                 -DKokkos_ENABLE_CUDA_LAMBDA=OFF \
                                 -DKokkos_ENABLE_CUDA_UVM=ON \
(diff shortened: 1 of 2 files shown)


.jenkins
```diff
@@ -414,11 +414,7 @@ pipeline {
                                 -DCMAKE_CXX_FLAGS=-Werror \
                                 -DCMAKE_CXX_STANDARD=17 \
                               .. && \
-                              make -j8 && ctest --verbose && \
-                              cd ../.. && \
-                              cmake -B build_cmake_installed_different_compiler/build -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_CXX_FLAGS=-Werror -DCMAKE_CXX_STANDARD=17 build_cmake_installed_different_compiler && \
-                              cmake --build build_cmake_installed_different_compiler/build --target all && \
-                              cmake --build build_cmake_installed_different_compiler/build --target test'''
+                              make -j8 && ctest --verbose'''
```
Rombur, line 417: Why did you change this?

masterleinad: Using different compilers with `Kokkos` is problematic and doesn't work with `OpenMP` at all since we are using different flags.

Rombur (approved):

.jenkins
dalg24: Please elaborate why it is ok to drop that test.

masterleinad: It's just dysfunctional for `OpenMP` right now (which is the backend this CI configuration was set up to use in the first place) and it's not clear to me if this is something we want to address and if it's worth it to test a different configuration with a (significantly) different compiler than `Kokkos` was configured with.

cz4rs (approved):
Looks good to me.
Note: if we want to cover `build_cmake_installed_different_compiler` case, than a separate build would probably be best. Also, if it's just for the sake of different compiler then GitHub Actions could be used (with `g++` + `clang` combo).

masterleinad:
Apparently, we forgot another place where we assumed `impl_thread_pool_size` to be `static`.

core/unit_test/TestDefaultDeviceTypeInit.hpp
```diff
@@ -262,7 +262,7 @@ void check_correct_initialization(const Kokkos::InitArguments& argstruct) {
 #endif
   }
 
-  ASSERT_EQ(Kokkos::HostSpace::execution_space::impl_thread_pool_size(),
+  ASSERT_EQ(Kokkos::HostSpace::execution_space().impl_thread_pool_size(),
```
dalg24, line 265: I don't get how that is not exposed by any other build until now.

.jenkins
```diff
@@ -384,7 +384,7 @@ pipeline {
                                 -DCMAKE_CXX_STANDARD=17 \
                                 -DKokkos_ARCH_NATIVE=ON \
                                 -DKokkos_ENABLE_COMPILER_WARNINGS=ON \
-                                -DKokkos_ENABLE_OPENMP=OFF \
+                                -DKokkos_ENABLE_OPENMP=ON \
```
dalg24, line 387: I would rather not mess with that build in that PR.
I expect a CPU-only build should be able to expose the issue you are fixing in this PR.

core/unit_test/TestDefaultDeviceTypeInit.hpp
dalg24: I suppose I could have read the description.

dalg24:
Plz discuss what kokkos versions are affected by that bug.  Is it develop only or also in 4.0.01?

masterleinad:
> Plz discuss what kokkos versions are affected by that bug. Is it develop only or also in 4.0.01?

The changes to `impl_thread_pool_size` was done after we branched 4.0.0.

dalg24 (approved):

dalg24:
GCC-8.4.0 build passed.  Everything else but the HIP builds passed.  Not waiting on them considering that they might time out since we only have on node with AMD GPU at the moment and that the queue is building up in the CI.

PR #5979, opened 2023-03-10
Print Kokkos version at configuration time

dutkalex:
Hi ! This is just a 1 line add in the CMakeLists.txt which prints the Kokkos version. I thought this would be useful to more people than just me, hence this pull request. Thank you guys for the amazing work you do :)


dalg24:
Do you care to see the version when you configure Kokkos or when you do `find_package(Kokkos)` downstream?
(The changes proposed address the former.)

dutkalex:
Indeed I had the former in mind, but I would say that also displaying that information when doing the `find_package` would be great.

dalg24 (approved):
I find it moderately useful but I have no objection.

masterleinad (approved):

PR #5899, opened 2023-02-21
UnorderedMap: Fix mutable m_size corruption bug

e10harvey:
Using the changes in this PR without the new ~UnorderMap destructor in Kokkos_UnorderedMap.hpp:
```
./containers/unit_tests/KokkosContainers_UnitTest_Cuda --gtest_filter=*UnorderedMap_insert
<snip>
kokkos/containers/unit_tests/TestUnorderedMap.hpp:189: Failure
Expected equality of these values:
  expected_inserts
    Which is: 900
  map_size
    Which is: 0
[  FAILED  ] cuda.UnorderedMap_insert (1507 ms)
[----------] 1 test from cuda (1507 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (1508 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] cuda.UnorderedMap_insert
```

With all the changes in this PR:
```
./containers/unit_tests/KokkosContainers_UnitTest_Cuda --gtest_filter=*UnorderedMap_insert
Note: Google Test filter = *UnorderedMap_insert
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from cuda
[ RUN      ] cuda.UnorderedMap_insert
[       OK ] cuda.UnorderedMap_insert (1495 ms)
[----------] 1 test from cuda (1495 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (1495 ms total)
[  PASSED  ] 1 test.
```

PR #5983, opened 2023-03-14
UnorderedMap: Fix mutable m_size corruption bug

masterleinad:
Alternative to #5899.
`m_size` is about the only member variable that is not shared between copies of `UnorderedMap` objects (but is computed from a shared member). Trying to cache it leads to unexpected problems if different copies modify their member as the test case (stolen from #5899 demonstrates). Making `m_size` a `Kokkos::View` as well fixes that.

### DIFF
--- a/containers/src/Kokkos_UnorderedMap.hpp
+++ b/containers/src/Kokkos_UnorderedMap.hpp
@@ -275,7 +275,7 @@ class UnorderedMap {
       : m_bounded_insert(true),
         m_hasher(hasher),
         m_equal_to(equal_to),
-        m_size(),
+        m_size("m_size"),
         m_available_indexes(calculate_capacity(capacity_hint)),
         m_hash_lists(view_alloc(WithoutInitializing, "UnorderedMap hash list"),
                      Impl::find_hash_size(capacity())),
@@ -315,7 +315,7 @@ class UnorderedMap {
       Kokkos::deep_copy(m_keys, tmp);
     }
     Kokkos::deep_copy(m_scalars, 0);
-    m_size = 0;
+    m_size() = 0;
   }
 
   KOKKOS_INLINE_FUNCTION constexpr bool is_allocated() const {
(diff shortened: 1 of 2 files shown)


containers/src/Kokkos_UnorderedMap.hpp
```diff
@@ -818,7 +818,7 @@ class UnorderedMap {
   bool m_bounded_insert;
   hasher_type m_hasher;
   equal_to_type m_equal_to;
-  mutable size_type m_size;
+  mutable Kokkos::View<size_type, Kokkos::HostSpace> m_size;
```
PhilMiller, line 821: Is `mutable` still necessary?

PhilMiller:
This makes way more sense to me than #5899.

containers/src/Kokkos_UnorderedMap.hpp
PhilMiller: I would lean toward removal.

Since this is strictly host-space, the meaning could be clarified by making it `shared_ptr<size_type>` instead of using `View` as such.

masterleinad: Yeah, let's try that.

nmm0 (changes requested):
Can we add an additional test that also triggers the bug in a more succinct manner?

```cpp
TEST(TEST_CATEGORY, UnorderedMap_consistent_size) {
  using Map =
      Kokkos::UnorderedMap<int, void, Kokkos::DefaultHostExecutionSpace>;

  Map m(11);
  m.insert(7);;
  ASSERT_EQ(1u, m.size());

  {
    auto m2 = m;
    m2.insert(2);
    // This line triggers modified flags to be cleared in both m and m2
    [[maybe_unused]] auto sz = m2.size();
  }

  ASSERT_EQ(2u, m.size());
}
```

I don't like the idea of keeping a separate bit of memory alloc'd somewhere, maybe we should discuss alternatives. One possibility would be to keep m_size as an element of the `m_scalars` view.

I don't even get why m_size is on the host -- it doesn't let us call size() from a device function and we still have to do a deep copy from the device every time we call size() with the current implementation

masterleinad:
@nmm0 I added the other test case as well. We should probably discuss in today's meeting if we want a quick fix to get @brian-kelley moving or redesign this.

e10harvey (approved):

nmm0 (approved):
Looks good to me -- I'm ok with merging this in as a quick fix and really fixing UnorderedMap at a later date

PR #5746, opened 2023-01-11
Thread Safety: Call cudaSetDevice before almost any CUDA API call

crtrott:
When one creates a new thread, that thread will get the default CUDA environment. I.e. any CUDA API call will initally be with respect to device zero unless `cudaSetDevice` is called on that thread.

This means we have to call `cudaSetDevice` again before practically any CUDA API call to be safe.

Currently the device ID we use is a static variable. Subsequently it could be a per-instance ID. This would require us to change some of these places, in particular the `Kokkos::fence()` impl, which would need to iterate over all active devices.

This fixes #5713 and related threading issues

PR #5989, opened 2023-03-16
Encapsulate CudaAPI calls in CudaInternal, call cudaSetDevice for thread safety

### LINKED ISSUE #5713: Nightly test failures with Cuda builds running on device id != 0
**Describe the bug**
Some unit tests are failing in Cuda builds when running with device_id != 0:

**KokkosCore_UnitTest_Cuda1**
```
4: [ RUN      ] cuda.partitioning_by_args
4: /ascldap/users/ndellin/kokkos/core/unit_test/TestExecSpacePartitioning.hpp:101: Failure
4:   sum1
4:     Which is: 0
4:   N * (N - 1) / 2
4:     Which is: 7642095
4: [  FAILED  ] cuda.partitioning_by_args (182 ms)
4: [ RUN      ] cuda.partitioning_by_vector
4: /ascldap/users/ndellin/kokkos/core/unit_test/TestExecSpacePartitioning.hpp:101: Failure
4:   sum1
4:     Which is: 0
4:   N * (N - 1) / 2
4:     Which is: 7642095
4: [  FAILED  ] cuda.partitioning_by_vector (0 ms)
```

One of the following commits began triggering the failures:

Fixup non-static {Serial,OpenACC}::concurrency() member function unless deprecated code 4 is ON ([details](https://jenkins-son.sandia.gov/job/Kokkos_develop_kokkosdev2_cuda110_gcc920_cpp17-deviceid_1/700/changes#052256fa24fd6cb424bef3ab1cd56eb224ef1dff) / [githubweb](https://github.com/kokkos/kokkos/commit/052256fa24fd6cb424bef3ab1cd56eb224ef1dff))
    Make Threads::concurrency() a non-static member function unless deprecated code 4 is ON ([details](https://jenkins-son.sandia.gov/job/Kokkos_develop_kokkosdev2_cuda110_gcc920_cpp17-deviceid_1/700/changes#33762bf5aba1c66f25fa8c5803192baf0f74e462) / [githubweb](https://github.com/kokkos/kokkos/commit/33762bf5aba1c66f25fa8c5803192baf0f74e462))
    5594: Remove two-argument CudaLDFFetch constructor ([details](https://jenkins-son.sandia.gov/job/Kokkos_develop_kokkosdev2_cuda110_gcc920_cpp17-deviceid_1/700/changes#4e08cb2635df5918b857a3b0c8ef89a611005796) / [githubweb](https://github.com/kokkos/kokkos/commit/4e08cb2635df5918b857a3b0c8ef89a611005796))
    Fixup non-static OpenTarget::concurrency() member function unless deprecated code 4 is ON ([details](https://jenkins-son.sandia.gov/job/Kokkos_develop_kokkosdev2_cuda110_gcc920_cpp17-deviceid_1/700/changes#e2efab2a12c766455a57426ae089a147d3d2569a) / [githubweb](https://github.com/kokkos/kokkos/commit/e2efab2a12c766455a57426ae089a147d3d2569a))
    [ci skip] Update changelog ([details](https://jenkins-son.sandia.gov/job/Kokkos_develop_kokkosdev2_cuda110_gcc920_cpp17-deviceid_1/700/changes#14e9887310cb3ca5387a8ac92e8515ed2dd39b57) / [githubweb](https://github.com/kokkos/kokkos/commit/14e9887310cb3ca5387a8ac92e8515ed2dd39b57))
    Turn off classic intel workaround in Serial backend ([details](https://jenkins-son.sandia.gov/job/Kokkos_develop_kokkosdev2_cuda110_gcc920_cpp17-deviceid_1/700/changes#35fa45c52333e29da5cfb82170d7fac9daa0f279) / [githubweb](https://github.com/kokkos/kokkos/commit/35fa45c52333e29da5cfb82170d7fac9daa0f279))
    Fix classic Intel compiler Serial/OpenMP backend build ([details](https://jenkins-son.sandia.gov/job/Kokkos_develop_kokkosdev2_cuda110_gcc920_cpp17-deviceid_1/700/changes#2513fc76770efb3476cb48fc62580c11b586350b) / [githubweb](https://github.com/kokkos/kokkos/commit/2513fc76770efb3476cb48fc62580c11b586350b))
    Add partition_space to OpenMP (#5105) ([details](https://jenkins-son.sandia.gov/job/Kokkos_develop_kokkosdev2_cuda110_gcc920_cpp17-deviceid_1/700/changes#e4cce6e449d482e2bf0653d3b31c38d95d4090e9) / [githubweb](https://github.com/kokkos/kokkos/commit/e4cce6e449d482e2bf0653d3b31c38d95d4090e9))
Drop CudaLDGFetch special member functions ([details](https://jenkins-son.sandia.gov/job/Kokkos_develop_kokkosdev2_cuda110_gcc920_cpp17-deviceid_1/701/changes#8aa23152c5175dd5ebee91d6856b50c9b1567002) / [githubweb](https://github.com/kokkos/kokkos/commit/8aa23152c5175dd5ebee91d6856b50c9b1567002))
    Drop CudaLDGFetch default constructor definition (prefer defaulted one) ([details](https://jenkins-son.sandia.gov/job/Kokkos_develop_kokkosdev2_cuda110_gcc920_cpp17-deviceid_1/701/changes#d9ea15a732762fda6b791547c49b22e3735c00e3) / [githubweb](https://github.com/kokkos/kokkos/commit/d9ea15a732762fda6b791547c49b22e3735c00e3))
    Drop unnecessary inline specifiers in CudaLDGFetch ([details](https://jenkins-son.sandia.gov/job/Kokkos_develop_kokkosdev2_cuda110_gcc920_cpp17-deviceid_1/701/changes#588dda891c01cae8f17671ca00f7bc2eac31246e) / [githubweb](https://github.com/kokkos/kokkos/commit/588dda891c01cae8f17671ca00f7bc2eac31246e))
    Fix {Cuda:: -> Cuda().}concurrency() occurences ([details](https://jenkins-son.sandia.gov/job/Kokkos_develop_kokkosdev2_cuda110_gcc920_cpp17-deviceid_1/701/changes#ad6c3dafec65e6b220a3661fb3ffaf79392c9f1f) / [githubweb](https://github.com/kokkos/kokkos/commit/ad6c3dafec65e6b220a3661fb3ffaf79392c9f1f))
    Replace Impl::CudaInternal::m_maxConcurrency data member by Impl::CudaInternal::concurrency() member function ([details](https://jenkins-son.sandia.gov/job/Kokkos_develop_kokkosdev2_cuda110_gcc920_cpp17-deviceid_1/701/changes#136abca269b6d87de69509c3af59f04eda62093b) / [githubweb](https://github.com/kokkos/kokkos/commit/136abca269b6d87de69509c3af59f04eda62093b))
    Fix chicken and egg Cuda concurrency issue when initializing lock arrays ([details](https://jenkins-son.sandia.gov/job/Kokkos_develop_kokkosdev2_cuda110_gcc920_cpp17-deviceid_1/701/changes#4cfa416c7c4768ab4937419280d6d279f39fea15) / [githubweb](https://github.com/kokkos/kokkos/commit/4cfa416c7c4768ab4937419280d6d279f39fea15))
    Make Cuda::concurrency() member function non static ([details](https://jenkins-son.sandia.gov/job/Kokkos_develop_kokkosdev2_cuda110_gcc920_cpp17-deviceid_1/701/changes#a1b86463689cdd7e798d94794aa0c1568cda2897) / [githubweb](https://github.com/kokkos/kokkos/commit/a1b86463689cdd7e798d94794aa0c1568cda2897))
    Fix {HIP:: -> HIP().}concurrency() occurences ([details](https://jenkins-son.sandia.gov/job/Kokkos_develop_kokkosdev2_cuda110_gcc920_cpp17-deviceid_1/701/changes#965998da585dd96e404dda48474f33df86a1dde4) / [githubweb](https://github.com/kokkos/kokkos/commit/965998da585dd96e404dda48474f33df86a1dde4))
    Introduce Impl::HIPInternal::concurrency static member function ([details](https://jenkins-son.sandia.gov/job/Kokkos_develop_kokkosdev2_cuda110_gcc920_cpp17-deviceid_1/701/changes#969490a83c7930fdb2d8874f93c775a60b01f13e) / [githubweb](https://github.com/kokkos/kokkos/commit/969490a83c7930fdb2d8874f93c775a60b01f13e))
    Make HIP::concurrency() member function non static ([details](https://jenkins-son.sandia.gov/job/Kokkos_develop_kokkosdev2_cuda110_gcc920_cpp17-deviceid_1/701/changes#95fa3a75ebb666056d0b48ea83a3a2082bc53403) / [githubweb](https://github.com/kokkos/kokkos/commit/95fa3a75ebb666056d0b48ea83a3a2082bc53403))
    Fixup HIP header include incomplete type Impl::HIPInternal ([details](https://jenkins-son.sandia.gov/job/Kokkos_develop_kokkosdev2_cuda110_gcc920_cpp17-deviceid_1/701/changes#17170f6eafc7c7c090b77c53ada56e9cd9d56715) / [githubweb](https://github.com/kokkos/kokkos/commit/17170f6eafc7c7c090b77c53ada56e9cd9d56715))
    Add architecture flags for MSVC (#5673) ([details](https://jenkins-son.sandia.gov/job/Kokkos_develop_kokkosdev2_cuda110_gcc920_cpp17-deviceid_1/701/changes#f3cf72a2408b56479a09234fe18262bcacb00c82) / [githubweb](https://github.com/kokkos/kokkos/commit/f3cf72a2408b56479a09234fe18262bcacb00c82))
    #5667: dont install std algorithm headers multiple times (#5670) ([details](https://jenkins-son.sandia.gov/job/Kokkos_develop_kokkosdev2_cuda110_gcc920_cpp17-deviceid_1/701/changes#8303668181ef68e1684a90d25b460188843249c0) / [githubweb](https://github.com/kokkos/kokkos/commit/8303668181ef68e1684a90d25b460188843249c0))
    Work around CUDA+Clang Thrust issue (#5660) ([details](https://jenkins-son.sandia.gov/job/Kokkos_develop_kokkosdev2_cuda110_gcc920_cpp17-deviceid_1/701/changes#b2e2096e933e04bf338007696532009f1d467429) / [githubweb](https://github.com/kokkos/kokkos/commit/b2e2096e933e04bf338007696532009f1d467429))

**Reproducer** (kokkos-dev-2):
```bash
module load sems-archive-env sems-archive-cmake/3.17.1 sems-archive-gcc/9.2.0 cuda/11.0

$KOKKOS_PATH/generate_makefile.bash --with-devices=Cuda,Serial --arch=Volta70 --compiler=$KOKKOS_PATH/bin/nvcc_wrapper --cxxflags="-O3 -Wall -Wunused-parameter -Wshadow -pedantic -Werror -Wsign-compare -Wtype-limits -Wuninitialized " --cxxstandard="17" --kokkos-path=$KOKKOS_PATH --with-cuda-options=enable_lambda  --no-examples

make -j20

KOKKOS_DEVICE_ID=1 ctest -R KokkosCore_UnitTest_Cuda1 -V

```

tcclevenger:
Collect all CudaAPI calls in the `CudaInternal` instance and call `cudaSetDevice()` before CudaAPI calls.

**Reason for `cudaSetDevice()`**
Any new thread created after Kokkos is initialized gets the default env (with `deviceID=0`). If a user requests a `deviceID!=0` (say `1` for example), and then creates new threads, those threads will dispatch with default Cuda env (deviceID=0), but streams are only valid on device 1. See #5713 for a failing example. To be safe, practically all CudaAPI calls need to be proceeded by a call to `cudaSetDevice()`. 

**Reason for CudaAPI calls wrappers**
Motivated by https://github.com/kokkos/kokkos/pull/5746#issuecomment-1380998127. With CudaAPI calls in multiple places, it could be easy to miss adding `cudaSetDevice()`before each, and anyone writing new code using cudaAPI calls may not know this is needed, which wouldn't be an issue if the norm was to use these wrappers instead.

Fixes #5713. 

**Questions**
- Is there a cleaner solution? Any way for a created thread to use the correct Cuda env instead of the default one? 
- Any other changes from the default env that could also cause issues?

**TODO**
- [x] Add for remaining cudaAPI calls
- [x] Update comments to be descriptive
- [x] Verify performance doesn't degrade
- [ ] Add a test which triggers the fail without changes (maybe)


core/src/Cuda/Kokkos_Cuda_Instance.hpp
```diff
@@ -193,6 +193,26 @@ class CudaInternal {
     }
   }
 
+  ///////////////////////////////////////////////////////////////////////////////////////////
+
+  template <typename... InputArgs>
+  void cuda_api_interface_safe_call(
+      cudaError_t (*cuda_api_function)(InputArgs...),
+      InputArgs... cuda_api_input) const {
```
romintomasetti, line 201: I think I would use perfect forwarding here, to make this future-proof ;) (sorry to interrupt, I jumped here from #6008)

PhilMiller: It's a C interface being called down to, so nothing will be large or have constructors or destructors. In general, you'd be right, though.

tcclevenger:
For performance testing, I used a simple program launching parallel_for/reduce kernels, timing with/without fencing, on a single weaver V100 GPU (on my own node, not shared), develop vs. this branch (rebased to same commit). 
Basics:
```
for (loop over M) {
  Kokkos::parallel_for/reduce(N, ...
       {
         for (loop over K) a += b*c;
       }
   );
}
```

For higher workloads (Policy size `N=1e4` with operations per iteration `K=200, 3000`, and `N=1e5` with `K=1, 16, 200, 3000`), almost no increase in run time, less than 1% relative increase in all cases. 

At smaller workloads (Policy size `N=1e4`, operations per iteration `K=1, 16`), anywhere between 0-4% increase.

core/src/Cuda/Kokkos_Cuda.hpp
```diff
@@ -241,28 +240,6 @@ struct DeviceTypeTraits<Cuda> {
...
-  }
-
-  ZeroMemset(const View<DT, DP...>& dst,
-             typename View<DT, DP...>::const_value_type&) {
-    KOKKOS_IMPL_CUDA_SAFE_CALL(
-        cudaMemset(dst.data(), 0,
-                   dst.size() * sizeof(typename View<DT, DP...>::value_type)));
-  }
-};
-}  // namespace Impl
```
tcclevenger, line 265: I moved this to `Kokkos_Cuda_Instance.hpp` to avoid build errors that I think are from circular includes (since `ZeroMemset` now needs `CudaInternal`). If that is not ideal I can spend more time resolving those issues.

core/src/Cuda/Kokkos_Cuda_Instance.hpp
```diff
@@ -193,6 +193,80 @@ class CudaInternal {
...
+  // SharedAllocationRecord::attach_texture_object(). At times the InputArgs may
+  // not be deduced correctly by certain compilers. In most cases, explicitly
+  // providing them solves this issue. For an example, see call to
+  // cudaMemset(void*, int, size_t) in CudaInternal::scratch_flags(). If the
+  // compiler still does not recognize InputArgs, some input vars may need to be
+  // cast to their correct type. For an example, see call to cudaMalloc() in
+  // CudaInternal::initialize().
+  template <bool setCudaDevice, typename... InputArgs>
+  void cuda_api_interface_safe_call(
+      cudaError_t (*cuda_api_function)(InputArgs...),
```
dalg24, line 233: I am wondering how fragile that might be to take the address to a CUDA runtime function.  I am not sure whether they reserve the right to add default arguments.
It is not necessary a blocker but we need to consider it.

dalg24 (commented):
The main drawback of the current approach is that we lose the source location from where the abort was called.  I suppose we still get a stack trace.  Did you consider a macro-based approach.

core/src/Cuda/Kokkos_Cuda_Instance.hpp
PhilMiller: It's a C API, not C++, right? So it shouldn't have the same issues with taking addresses of functions that may later be overloaded or whatever

core/src/Cuda/Kokkos_Cuda.hpp
masterleinad: https://github.com/kokkos/kokkos/pull/6051 has similar changes.

tcclevenger:
@masterleinad I'm planning to implement @dalg24 suggestion in a new PR (`impl/Kokkos_ZeroMemset.hpp` and `Cuda/Kokkos_cuda_ZeroMemset.hpp`). I'll also include HIP implementation. I say just let your PR merge if it's ready and I'll deal with the conflicts.

masterleinad:
>  I say just let your PR merge if it's ready and I'll deal with the conflicts.

No, it's not ready. We should probably get the pull request you are proposing in first.

tcclevenger:
> > I say just let your PR merge if it's ready and I'll deal with the conflicts.
> 
> No, it's not ready. We should probably get the pull request you are proposing in first.

Sounds good, should be straightforward. I'll start working on it tomorrow.

masterleinad:
In general, all places that have access to the execution space instance should use the CudaInternal instance associated with it (even though only one device can be used right now).

masterleinad:
In general, I think that we should not set the device when we don't know the execution space instance/device/stream.

masterleinad (changes requested):
In general, I think that we should not set the device when we don't know the execution space instance/device/stream.

tcclevenger:
@masterleinad Ok, let me think about the best way to do that. It may be the case where function calling API interface need to change to provide the space.

masterleinad:
> @masterleinad Ok, let me think about the best way to do that. It may be the case where function calling API interface need to change to provide the space.

Yes, I agree. For https://github.com/kokkos/kokkos/pull/6091, `get_cuda_func_attributes` seems to be the only function, though. Note that it doesn't really matter for most `Cuda` functions that don't take a `stream` anyway, like `cudaMemcpy` or `cudaDeviceSynchronize`.

masterleinad (commented):
In the end, there are two objectives that are somewhat competing:
1. There is only one device used and that is not the one with device id 0. Every new thread starts off with device id 0. Since we can't control very well when threads are created we want to set the device the one device used for every possible `Cuda` interface call.
2. We want to support using multiple devices in some way (which we still need to discuss). In this case, setting the device to the default id might be problematic since, e.g., stream access and memory access must use the correct device (by default unless we enforce setting up peer access or the like).

I went through the pull request to identify places where we are still using `singleton()` that might cause problems for 2. Since there are a bunch of more changes necessary for 2 that are conflicting with the ones here, it's probably best to focus on 1 to make progress. Still, I would appreciate it if you could either try using the correct device id or adding a FIXME_CUDA_MULTIPLE_DEVICES for the places where I added comments that using the correct device is necessary.

Let's always set the device in `cuda_api_interface_safe_call`, though, unless explicitly requested not to do that, i.e., drop special casing for `singleton()`.

masterleinad:
> The user can specify KOKKOS_DEVICE_ID and that value gets picked up when we initialize Cuda and stored in static int CudaInternal::m_cudaDev. Setting the correct ID in CudaInternal::cuda_api_interface_safe_call() doesn't use any information from the instance calling it, it just sets the value internally to what the user requested.

I'm mostly forward-looking for supporting multiple devices since that is very closely related and that requires `m_CudaDev` to be a non-static member variable.

core/src/Cuda/Kokkos_CudaSpace.cpp
```diff
@@ -43,7 +43,8 @@
 cudaStream_t Kokkos::Impl::cuda_get_deep_copy_stream() {
   static cudaStream_t s = nullptr;
   if (s == nullptr) {
-    cudaStreamCreate(&s);
+    CudaInternal::singleton().cuda_api_interface_safe_call<cudaStream_t *>(
+        &cudaStreamCreate, &s);
```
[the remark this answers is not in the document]
masterleinad: Yes, and it should be fine for the copy_stream to be associated with the default device. 🙂

core/src/Cuda/Kokkos_CudaSpace.cpp
```diff
@@ -66,19 +67,28 @@ static std::atomic<int> num_uvm_allocations(0);
...
 
 void DeepCopyAsyncCuda(void *dst, const void *src, size_t n) {
   cudaStream_t s = cuda_get_deep_copy_stream();
-  KOKKOS_IMPL_CUDA_SAFE_CALL(
-      cudaMemcpyAsync(dst, src, n, cudaMemcpyDefault, s));
+  CudaInternal::singleton()
+      .cuda_api_interface_safe_call<void *, const void *, size_t,
+                                    cudaMemcpyKind, cudaStream_t>(
+          &cudaMemcpyAsync, dst, src, n, cudaMemcpyDefault, s);
+
```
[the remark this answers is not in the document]
masterleinad: Yes, we need to set the device here. 🙂

core/src/Cuda/Kokkos_CudaSpace.cpp
```diff
@@ -240,18 +264,26 @@ void *CudaUVMSpace::impl_allocate(
     Kokkos::Impl::num_uvm_allocations++;
 
     auto error_code =
-        cudaMallocManaged(&ptr, arg_alloc_size, cudaMemAttachGlobal);
+        Impl::CudaInternal::singleton()
+            .cuda_api_interface_return<cudaError_t, void **, size_t,
+                                       unsigned int>(
+                &cudaMallocManaged, &ptr, arg_alloc_size, cudaMemAttachGlobal);
```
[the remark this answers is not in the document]
tcclevenger: A lot of these I let just do the default behavior. I can set cuda dev = false if you like.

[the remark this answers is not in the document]
masterleinad: I think just setting the device even if it doesn't matter for multi-GPU support is the safer/better option.

core/src/Cuda/Kokkos_Cuda_Instance.cpp
```diff
@@ -100,13 +100,21 @@ int cuda_kernel_arch() {
...
   query_cuda_kernel_arch<<<1, 1>>>(d_arch);
 
-  cudaMemcpy(&arch, d_arch, sizeof(int), cudaMemcpyDefault);
-  cudaFree(d_arch);
+  CudaInternal::singleton()
+      .cuda_api_interface_safe_call<false, void *, const void *, size_t,
+                                    cudaMemcpyKind>(
+          &cudaMemcpy, &arch, d_arch, sizeof(int), cudaMemcpyDefault);
+  Impl::CudaInternal::singleton().cuda_api_interface_safe_call<false, void *>(
+      &cudaFree, d_arch);
```
masterleinad, line 117: Please pass the correct device id into this function (and use it) or add a FIXME_CUDA_MULTIPLE_DEVICES.

core/src/Cuda/Kokkos_Cuda_KernelLaunch.hpp
```diff
@@ -136,7 +136,19 @@ const cudaFuncAttributes& get_cuda_kernel_func_attributes(
...
+#if defined(KOKKOS_COMPILER_CLANG)
+    // Clang compilers in the CI were throwing an
+    // error complaining of ambiguous address of overloaded function
+    // "cudaFuncGetAttributes" (templated cudaAPI function).
+    // Manually call cudaAPI functions as a workaround.
     KOKKOS_IMPL_CUDA_SAFE_CALL(cudaFuncGetAttributes(&attr, func));
+#else
+    CudaInternal::singleton()
+        .cuda_api_interface_safe_call<cudaFuncAttributes*, KernelFuncPtr>(
+            &cudaFuncGetAttributes, &attr, func);
```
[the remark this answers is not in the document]
masterleinad: Please add a FIXME_CUDA_MULTIPLE_DEVICES.

core/src/Cuda/Kokkos_Cuda_KernelLaunch.hpp
```diff
@@ -218,8 +230,10 @@ inline void configure_shmem_preference(const KernelFuncPtr& func,
 
   // Set the carveout, but only call it once per kernel or when it changes
   auto set_cache_config = [&] {
-    KOKKOS_IMPL_CUDA_SAFE_CALL(cudaFuncSetAttribute(
-        func, cudaFuncAttributePreferredSharedMemoryCarveout, carveout));
+    CudaInternal::singleton()
+        .cuda_api_interface_safe_call<KernelFuncPtr, cudaFuncAttribute, int>(
+            &cudaFuncSetAttribute, func,
+            cudaFuncAttributePreferredSharedMemoryCarveout, carveout);
```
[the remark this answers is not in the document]
masterleinad: FIXME_CUDA_MULTIPLE_DEVICES

core/src/Cuda/Kokkos_Cuda_ZeroMemset.hpp
```diff
@@ -27,16 +27,19 @@ template <class T, class... P>
...
 
   ZeroMemset(const View<T, P...>& dst,
              typename View<T, P...>::const_value_type&) {
-    KOKKOS_IMPL_CUDA_SAFE_CALL(
-        cudaMemset(dst.data(), 0,
-                   dst.size() * sizeof(typename View<T, P...>::value_type)));
+    Kokkos::Impl::CudaInternal::singleton()
+        .cuda_api_interface_safe_call<void*, int, size_t>(
+            &cudaMemset, dst.data(), 0,
+            dst.size() * sizeof(typename View<T, P...>::value_type));
```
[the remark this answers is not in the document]
masterleinad: FIXME_CUDA_MULTIPLE_DEVICES

masterleinad (approved):
Looks OK to me. I would appreciate it if you could add some more FIXMEs, though.

core/src/Cuda/Kokkos_Cuda_Instance.cpp
tcclevenger: I was passing `false` here since the first cuda api call in this function (which sets the device) ensures all the following calls are set correctly. I'll test again, but I think there should be no major performance gain by setting `false` the few times I do. I change it to all be the default (setting true).

There is only one place where we need `setCudaDevice=false` because we are in the process of querying for the device id. So we still need the option to set false. I'll tag it.

core/src/Cuda/Kokkos_Cuda_Instance.cpp
```diff
@@ -253,15 +279,18 @@ CudaInternalDevices::CudaInternalDevices() {
   // See 'cudaSetDeviceFlags' for host-device thread interaction
   // Section 4.4.2.6 of the CUDA Toolkit Reference Manual
 
-  KOKKOS_IMPL_CUDA_SAFE_CALL(cudaGetDeviceCount(&m_cudaDevCount));
+  CudaInternal::singleton().cuda_api_interface_safe_call<false, int *>(
+      &cudaGetDeviceCount, &m_cudaDevCount);
```
tcclevenger, line 283: Here and the one below are the only  calls that need `setCudaDevice=false` since `CudaInternal::m_CudaDev` has not been set.

core/src/Cuda/Kokkos_Cuda_Task.hpp
```diff
@@ -244,30 +243,29 @@ class TaskQueueSpecialization<SimpleTaskScheduler<Kokkos::Cuda, QueueType>> {
...
     // If not large enough then set the stack size, in bytes:
 
     const size_t larger_stack_size = 1 << 11;
 
     if (previous_stack_size < larger_stack_size) {
-      KOKKOS_IMPL_CUDA_SAFE_CALL(
-          cudaDeviceSetLimit(cudaLimitStackSize, larger_stack_size));
+      CudaInternal::singleton().cuda_api_interface_safe_call<cudaLimit, size_t>(
+          &cudaDeviceSetLimit, cudaLimitStackSize, larger_stack_size);
     }
```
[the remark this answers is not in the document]
tcclevenger: I added to the beginning of this function (line 224). Let me know if you prefer to add in front of the actual api calls

masterleinad:
Thanks @tcclevenger!

crtrott:
DId we do any performance check on whether the function pointer approach is worse than directly calling? My guess is no/no (i.e. we didn't check, but it also doesn't impact).

tcclevenger:
@crtrott I only tested develop vs. this branch kernel launch time (diffs are function ptrs and calling set device) and there wasn't a significant overhead (see https://github.com/kokkos/kokkos/pull/5989#issuecomment-1523639782).

It would be easy to test with only function ptr change if you like.

crtrott (approved):

crtrott:
I tested performance and didn't see any impact over a bunch of scenarios. Still not 100% sure about the function pointer approach instead of having a bunch of explicit functions inside the instance. But I guess this is ok for now?

tcclevenger:
@crtrott I can put together a branch with individual functions if you like (still have an old one with ~50% of them added), but it does create a lot more code. Another alternative could be to take in a lambda with whatever cudaAPI call(s) instead of function pointers.

tcclevenger:
Mentioned in the devel meeting, here is an example of what a runtime error looks like with the current approach.
```
Test: change input arg dst.data() -> nullptr in ZeroMemset() { cudaMemsetAsync() }

Error output:
  what():  cuda_api_function(cuda_api_input...) error( cudaErrorInvalidValue): invalid argument /home/tccleve/Kokkos/install/include/Cuda/Kokkos_Cuda_Instance.hpp:238
Aborted (core dumped)

cuda-gdb bt output:
(cuda-gdb) bt
#0  0x0000200002229628 in raise () from /lib64/glibc-hwcaps/power9/libc-2.28.so
#1  0x0000200002203eb4 in abort () from /lib64/glibc-hwcaps/power9/libc-2.28.so
#2  0x0000200001eab4e8 in __gnu_cxx::__verbose_terminate_handler () at /tmp/jfrye/spack-stage/spack-stage-gcc-9.3.0-lgnjurnhg4eaelqm5jwx7hqucmvmht5j/spack-src/libstdc++-v3/libsupc++/vterminate.cc:95
#3  0x0000200001ea6f34 in __cxxabiv1::__terminate (handler=<optimized out>) at /tmp/jfrye/spack-stage/spack-stage-gcc-9.3.0-lgnjurnhg4eaelqm5jwx7hqucmvmht5j/spack-src/libstdc++-v3/libsupc++/eh_terminate.cc:48
#4  0x0000200001ea7000 in std::terminate () at /tmp/jfrye/spack-stage/spack-stage-gcc-9.3.0-lgnjurnhg4eaelqm5jwx7hqucmvmht5j/spack-src/libstdc++-v3/libsupc++/eh_terminate.cc:58
#5  0x0000200001ea74d8 in __cxxabiv1::__cxa_throw (obj=<optimized out>, tinfo=0x200002056c58 <typeinfo for std::runtime_error>, dest=0x200001ec6a40 <std::runtime_error::~runtime_error()>)
    at /tmp/jfrye/spack-stage/spack-stage-gcc-9.3.0-lgnjurnhg4eaelqm5jwx7hqucmvmht5j/spack-src/libstdc++-v3/libsupc++/eh_throw.cc:95
#6  0x00000000100785cc in Kokkos::Impl::throw_runtime_exception (msg=...) at /home/tccleve/Kokkos/kokkos/core/src/impl/Kokkos_Error.cpp:48
#7  0x000000001008e5c0 in Kokkos::Impl::cuda_internal_error_throw (e=cudaErrorInvalidValue, name=0x100bb8c8 "cuda_api_function(cuda_api_input...)", 
    file=0x100baca8 "/home/tccleve/Kokkos/install/include/Cuda/Kokkos_Cuda_Instance.hpp", line=238) at /home/tccleve/Kokkos/kokkos/core/src/Cuda/Kokkos_Cuda_Instance.cpp:193
#8  0x000000001000b200 in Kokkos::Impl::cuda_internal_safe_call (e=cudaErrorInvalidValue, name=0x100bb8c8 "cuda_api_function(cuda_api_input...)", file=0x100baca8 "/home/tccleve/Kokkos/install/include/Cuda/Kokkos_Cuda_Instance.hpp", 
    line=238) at /home/tccleve/Kokkos/install/include/Cuda/Kokkos_Cuda_Error.hpp:68
#9  0x000000001001a570 in Kokkos::Impl::CudaInternal::cuda_api_interface_safe_call<true, void*, int, unsigned long, CUstream_st*> (this=0x101a67f8 <Kokkos::Impl::CudaInternal::singleton()::self>, 
    cuda_api_function=0x200001cde030 <cudaMemsetAsync>) at /home/tccleve/Kokkos/install/include/Cuda/Kokkos_Cuda_Instance.hpp:238
#10 0x0000000010017c48 in Kokkos::Impl::CudaInternal::cuda_api_interface_safe_call<void*, int, unsigned long, CUstream_st*> (this=0x101a67f8 <Kokkos::Impl::CudaInternal::singleton()::self>, 
    cuda_api_function=0x200001cde030 <cudaMemsetAsync>) at /home/tccleve/Kokkos/install/include/Cuda/Kokkos_Cuda_Instance.hpp:245
#11 0x0000000010015af0 in Kokkos::Impl::ZeroMemset<Kokkos::Cuda, Kokkos::View<double*, Kokkos::CudaSpace, Kokkos::MemoryTraits<1u> > >::ZeroMemset (this=0x7fffffffb330, exec_space_instance=..., dst=...)
    at /home/tccleve/Kokkos/install/include/Cuda/Kokkos_Cuda_ZeroMemset.hpp:30
```
So if you get one line error, it definitely is unhelpful and all points to the same place with minimal information. However, gdb programs have always provided helpful traces in my experience, but would be nice to not have to use them.

Other options
- Macro approach: should be very simple/minimal for case where we need no return value, but the problem still exists where we need to return from cudaAPI function. So we would get a split impl for each case. Not immediately clear to me how to use specific CudaInternal instance from exec space, and not just the singleton (but I'm sure it's possible).
- Each cudaAPI funtion getting a wrapper function in cuda internal: a lot more code, and still one-line error points to CudaInternal, not to where the API function is used.
- New strategy, lambda input:
```
New wrapper in CudaInternal (very similar to api interface function in this PR):

template<bool setCudaDevice, typename FunctorType>
void cuda_api_wrapper(const FunctorType& cuda_api_safe_call) const {
  if (setCudaDevice) set_cuda_device();
  cuda_api_safe_call();
}

Usage in ZeroMemset constructor:

ZeroMemset(const Kokkos::Cuda& exec_space_instance, const View<T, P...>& dst,
           typename View<T, P...>::const_value_type&) {
  exec_space_instance.impl_internal_space_instance()->cuda_api_wrapper<true>(
      [&]() {
        KOKKOS_IMPL_CUDA_SAFE_CALL(cudaMemsetAsync(
            dst.data(), 0,
            dst.size() * sizeof(typename View<T, P...>::value_type),
            exec_space_instance.cuda_stream()));
      });
}
```
This is easy to do for cudaAPI return values as well. Basically requires developer to wrap cudaAPI calls in macro for error handling (already done), and then in CudaInternal wrapper for device handling. And then errors point to the correct place in source code.

tcclevenger:
@dalg24 My latest comment does the following:
- Uses `KOKKOS_IMPL_CUDA_SAFE_CALL()` macro around the cuda interface call
- Remove 2 different interface calls (return v.s no-return), replace with a single `cuda_api_interface()` function which returns what the cudaAPI function returns

Now with my example above (nullptr in ZeroMemset) we get the following error which gives cudaAPI function name, input, and linenumber:
```
  what():  (exec_space_instance.impl_internal_space_instance() ->cuda_api_interface<void*, int, size_t, cudaStream_t>( &cudaMemsetAsync, nullptr, 0, dst.size() * sizeof(typename View<T, P...>::value_type), exec_space_instance.cuda_stream())) error( cudaErrorInvalidValue): invalid argument /home/tccleve/Kokkos/install/include/Cuda/Kokkos_Cuda_ZeroMemset.hpp:35
Aborted (core dumped)
```

If you want to use the return value of the cudaAPI function, there won't be a nice error message, but this is the vast minority of cases, all of which are either returning the last error or getting error string, which shouldn't be high likelihood of errors I don't believe.

Let me know your thoughts.

tcclevenger:
Providing an alternative approach https://github.com/kokkos/kokkos/pull/6299 using individual functions for each cudaAPI functions.

tcclevenger:
Marking this as draft, I think support is more for the approach in https://github.com/kokkos/kokkos/pull/6299.

tcclevenger:
Other PR merged, closing this one.

PR #5990, opened 2023-03-16
Support --compiler-options in nvcc_wrapper

jczhang07:
According to 'nvcc --help', it is just an alias to -Xcompiler.

nvcc --help
...
--compiler-options <options>,...                (-Xcompiler)                    
        Specify options directly to the compiler/preprocessor.

### DIFF
--- a/bin/nvcc_wrapper
+++ b/bin/nvcc_wrapper
@@ -407,7 +407,7 @@ do
   -Woverloaded-virtual)
     ;;
   #strip -Xcompiler because we add it
-  -Xcompiler)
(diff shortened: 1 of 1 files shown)


masterleinad (approved):
Looks OK to me.

tcclevenger (approved):

dalg24:
@jczhang07 Please elaborate why you think this is needed and we need to add support for it.

Assuming that there is a good rationale to add, what about `--linker-options`?

@masterleinad @tcclevenger did you actually try the patch?

crtrott (approved):
Its fine ...

PR #5993, opened 2023-03-17
Introduce KOKKOS_ENABLE_SYCL_RELOCATABLE_DEVICE_CODE as CMake option

masterleinad:
This might improve compiler performance. From https://github.com/intel/llvm/blob/sycl/sycl/doc/UsersManual.md:
> -f[no-]sycl-rdc
Enables/disables relocatable device code. If relocatable device code is
disabled, device code cannot use SYCL_EXTERNAL functions, which allows
the compiler to link device code on a per-translation-unit basis.
This may result in compile time and compiler memory usage improvements.
'-fno-sycl-rdc' used along with '-fsycl-max-parallel-link-jobs' will enable
additional device linking parallism for fat static archives.
Relocatable device code is enabled by default.

SYCL needs SYCL_EXTERNAL for functions used with RDC/separable compilation ([intel.com/content/www/us/en/developer/articles/technical/use-the-sycl-external-macro.html](https://www.intel.com/content/www/us/en/developer/articles/technical/use-the-sycl-external-macro.html), [registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#subsec:syclexternal](https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#subsec:syclexternal)) but doesn't need any annotations for "normal" device functions in contrast to CUDA/HIP so the idea is to add a portability macro, ~~`KOKKOS_EXTERNAL_FUNCTION`~~ `KOKKOS_RELOCATABLE_FUNCTION`.

We replace `Kokkos_ENABLE_IMPL_SYCL_RELOCATABLE_DEVICE_CODE` with `Kokkos_ENABLE_SYCL_RELOCATABLE_DEVICE_CODE` and make the default `OFF` since the feature is properly implemented in with this pull request and the default matches what we do for CUDA and HIP. We could discuss to just define a `Kokkos_ENABLE_RELOCATABLE_DEVICE_CODE` option replacing the backend-specific ones. 

For oneAPI 2023.0.0, we still need to always enable relocatble device code and we make the default and only error out later if the user explicitly disables it.


nliber (approved):

ajpowelsnl:
@masterleinad - is this PR close to being ready, or is it still, in fact, a draft?

masterleinad:
> is this PR close to being ready, or is it still, in fact, a draft?

It's ready but not the highest priority. An open question is what the default value should be. Disabling relocatable device code makes compiling faster but we can't use some low-level level zero functions for getting random values.

ajpowelsnl:
> > is this PR close to being ready, or is it still, in fact, a draft?
> 
> It's ready but not the highest priority. An open question is what the default value should be. Disabling relocatable device code makes compiling faster but we can't use some low-level level zero functions for getting random values.

OK, thanks!  Would you like to discuss enable / disable relocatable device code on Wed.?

masterleinad:
> OK, thanks! Would you like to discuss enable / disable relocatable device code on Wed.?

No, we can do that here.

core/src/Kokkos_Macros.hpp
```diff
@@ -369,10 +373,14 @@
 #define KOKKOS_FORCEINLINE_FUNCTION \
   KOKKOS_IMPL_FORCEINLINE_FUNCTION  \
   __attribute__((annotate("KOKKOS_FORCEINLINE_FUNCTION")))
+#define KOKKOS_EXTERNAL_FUNCTION \
+  KOKKOS_IMPL_EXTERNAL_FUNCTION  \
+  __attribute__((annotate("KOKKOS_EXTERNAL_FUNCTION")))
```
Rombur, line 378: Why are you introducing a new macro?

Rombur (commented):
Don't you need to update the Makefile too?

core/src/Kokkos_Macros.hpp
masterleinad: SYCL needs SYCL_EXTERNAL for functions used with RDC/separable compilation (https://www.intel.com/content/www/us/en/developer/articles/technical/use-the-sycl-external-macro.html, https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#subsec:syclexternal) but doesn't need any annotations for "normal" device functions in contrast to CUDA/HIP so the idea is to add a portability macro.

masterleinad:
> Don't you need to update the Makefile too?

I could do that if requested.

core/src/setup/Kokkos_Setup_SYCL.hpp
```diff
@@ -62,4 +62,12 @@ using sycl_host_ptr = sycl::host_ptr<T>;
 #endif
 }  // namespace Kokkos::Impl
 
+#ifdef KOKKOS_ENABLE_SYCL_RELOCATABLE_DEVICE_CODE
+#define KOKKOS_IMPL_EXTERNAL_FUNCTION SYCL_EXTERNAL
+#else
+#define KOKKOS_IMPL_EXTERNAL_FUNCTION   \
+  @"KOKKOS_EXTERNAL_FUNCTION requires " \
+   "Kokkos_ENABLE_SYCL_RELOCATABLE_DEVICE_CODE=ON"
```
dalg24, line 70: Please print sample error message when this is used

example/build_cmake_rdc/functor.cpp
```diff
@@ -0,0 +1,21 @@
...
+//
+// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
+#include <Kokkos_Macros.hpp>
+
+KOKKOS_EXTERNAL_FUNCTION void count_even(const long i, long& lcount) {
```
dalg24, line 19: How does that work for CUDA and HIP?
As I understand it, with the current definition of `KOKKOS_EXTERNAL_FUNCTION`, the `__host__ __device__` annotations are missing.

example/build_cmake_rdc/CMakeLists.txt
dalg24, line 1: We don't want to include that example from the parent CMakeLists.txt ?

core/src/setup/Kokkos_Setup_SYCL.hpp
masterleinad:
```
/home/darndt/kokkos/example/build_cmake_rdc/functor.cpp:19:1: error: expected unqualified-id
   19 | KOKKOS_EXTERNAL_FUNCTION void count_even(const long i, long& lcount) {
      | ^
/home/darndt/kokkos-install/include/Kokkos_Macros.hpp:383:34: note: expanded from macro 'KOKKOS_EXTERNAL_FUNCTION'
  383 | #define KOKKOS_EXTERNAL_FUNCTION KOKKOS_IMPL_EXTERNAL_FUNCTION
      |                                  ^
/home/darndt/kokkos-install/include/setup/Kokkos_Setup_SYCL.hpp:69:3: note: expanded from macro 'KOKKOS_IMPL_EXTERNAL_FUNCTION'
   69 |   @"KOKKOS_EXTERNAL_FUNCTION requires " \
      |   ^
1 error generated.
```

example/build_cmake_rdc/functor.cpp
masterleinad: Fixed.

example/build_cmake_rdc/CMakeLists.txt
masterleinad: We don't do that for any of the `cmake_build*` examples.

example/build_cmake_rdc/functor.cpp
dalg24, line 19: I thought that macro would only be required on the extern declaration and that the definition would be a regular `KOKKOS_FUNCTION`.  Did I misunderstand how we are supposed to use it?

masterleinad: > When a function is declared with SYCL_EXTERNAL, that macro must be used on the first declaration of that function in the translation unit. Redeclarations of the function in the same translation unit may optionally use SYCL_EXTERNAL, but this is not required.
>
> When a function is declared with SYCL_EXTERNAL, that function must also be defined in some translation unit, where the function is declared with SYCL_EXTERNAL.

Reads to me that you should annotate both the declaration and the definition. If I omit the annotation on the declaration, I am getting
```
/home/darndt/kokkos/example/build_cmake_rdc/cmake_example.cpp:44:54: error: SYCL kernel cannot call an undefined function without SYCL_EXTERNAL attribute
   44 |       n, KOKKOS_LAMBDA(const long i, long& lcount) { count_even(i, lcount); },
      |                                                      ^
/home/darndt/kokkos/example/build_cmake_rdc/cmake_example.cpp:22:22: note: 'count_even' declared here
   22 | KOKKOS_FUNCTION void count_even(const long i, long& lcount);
      |                      ^
/home/darndt/kokkos/example/build_cmake_rdc/cmake_example.cpp:44:10: note: called by 'operator()'
   44 |       n, KOKKOS_LAMBDA(const long i, long& lcount) { count_even(i, lcount); },
      |          ^
/home/darndt/kokkos-install/include/Kokkos_Macros.hpp:128:23: note: expanded from macro 'KOKKOS_LAMBDA'
  128 | #define KOKKOS_LAMBDA [=]
      |                       ^
1 error generated.
```
and if I'm omitting it on the definition, I'm seeing
```
terminate called after throwing an instance of 'sycl::_V1::compile_program_error'
  what():  The program was built for 1 devices
Build program log for 'Intel(R) Data Center GPU Max 1550':
Module <0x2c598f0>:  Unresolved Symbol <_Z10count_evenlRl>
Module <0x2c598f0>:  Unresolved Symbol <_Z10count_evenlRl>
Module <0x2c598f0>:  Unresolved Symbol <_Z10count_evenlRl>
Module <0x2c598f0>:  Unresolved Symbol <_Z10count_evenlRl> -11 (PI_ERROR_BUILD_PROGRAM_FAILURE)
Aborted
``` 
so I really need it in both places.

dalg24: That's interesting https://www.intel.com/content/www/us/en/developer/articles/technical/use-the-sycl-external-macro.html does not have it on the definition

example/build_cmake_rdc/CMakeLists.txt
dalg24: Why do you call it "cmake" example?
It should be a regular example/tutorial with proper guards for setups that support RDC

example/build_cmake_rdc/functor.cpp
dalg24: Maybe the `extern` was missing?

core/src/Kokkos_Macros.hpp
```diff
@@ -357,6 +357,10 @@
 #define KOKKOS_IMPL_DEVICE_FUNCTION
 #endif
 
+#if !defined(KOKKOS_IMPL_EXTERNAL_FUNCTION)
+#define KOKKOS_IMPL_EXTERNAL_FUNCTION KOKKOS_IMPL_FUNCTION
```
dalg24, line 361: I think we should consider including the `extern` storage class specifier in that macro definition.

example/build_cmake_rdc/functor.cpp
masterleinad: That doesn't make a difference for me.

core/src/Kokkos_Macros.hpp
masterleinad: If we need to use in the definition as well, it seems strange to use `extern` but that would work  as well. What case do you have in mind were `extern` would be required and not redundant?

example/build_cmake_rdc/functor.cpp
dalg24: Please confirm that you are saying the example from Intel's website gives an error at runtime

core/src/Kokkos_Macros.hpp
dalg24: How does that seem strange?  It is literally in the macro's name.

example/build_cmake_rdc/functor.cpp
masterleinad: A simple example with
```C++
// file1.cpp
#include <sycl/sycl.hpp>

SYCL_EXTERNAL int f();

int main() {
  sycl::queue q;
  q.single_task([](){ f(); });
  q.wait();
  std::cout << "done" << std::endl;
}
```
and
```C++
// file2.cpp
#include <sycl/sycl.hpp>

/*SYCL_EXTERNAL*/ int f() {
  return 2;
}
```
gives
```
> icpx -fsycl -c file1.cpp && icpx -fsycl -c file2.cpp && icpx -fsycl file1.o file2.o && ./a.out 
terminate called after throwing an instance of 'sycl::_V1::compile_program_error'
  what():  The program was built for 1 devices
Build program log for 'Intel(R) Data Center GPU Max 1550':
Module <0x2d54ec0>:  Unresolved Symbol <_Z1fv>
Module <0x2d54ec0>:  Unresolved Symbol <_Z1fv> -11 (PI_ERROR_BUILD_PROGRAM_FAILURE)
Aborted
```
You can first declare the function as `SYCL_EXTERNAL` in both compilation units but omit it in the definition, though:
```C++
// file2.cpp
#include <sycl/sycl.hpp>

SYCL_EXTERNAL int f();

int f() {
  return 2;
}
```
works.

core/src/Kokkos_Macros.hpp
masterleinad: Yes, `extern` is in the macro's name but it doesn't do the same as `extern` in "regular" C++ code for free functions (which is redundant anyway). It's possible to use `extern` also for the definition of a function but it doesn't carry the intent that the function is defined in a different translation.

All in all, I lean slightly towards not adding `extern`.

crtrott: should we call it relocatable then? I.e. KOKKOS_RELOCTABLE_FUNCTION?

Rombur: I am in favor of renaming the macro to `KOKKOS_RELOCATABLE_FUNCTION`.

example/separable_compilation/CMakeLists.txt
```diff
@@ -0,0 +1,9 @@
+KOKKOS_INCLUDE_DIRECTORIES(${CMAKE_CURRENT_BINARY_DIR})
+KOKKOS_INCLUDE_DIRECTORIES(${CMAKE_CURRENT_SOURCE_DIR})
+
+KOKKOS_ADD_EXECUTABLE(
+  example_separable_compilation
+  SOURCES cmake_example.cpp functor.cpp
+)
+
+add_test(NAME KokkosSeparableCompilation COMMAND Kokkos_example_separable_compilation 10)
```
dalg24, line 9: What's that trailing `10` argument?
```suggestion
add_test(NAME KokkosSeparableCompilation COMMAND Kokkos_example_separable_compilation)
```

dalg24: Nevermind

example/separable_compilation/cmake_example.cpp
```diff
@@ -0,0 +1,64 @@
...
+#include <Kokkos_Core.hpp>
+
+#include <cstdio>
+#include <iostream>
+
+KOKKOS_RELOCATABLE_FUNCTION void count_even(const long i, long& lcount);
+
+int main(int argc, char* argv[]) {
+  Kokkos::initialize(argc, argv);
+  Kokkos::DefaultExecutionSpace().print_configuration(std::cout);
```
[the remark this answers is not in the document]
masterleinad: It's just the prototypical example used in the `cmake_*` examples.

core/src/Kokkos_Macros.hpp
```diff
@@ -369,10 +373,14 @@
...
   KOKKOS_IMPL_FORCEINLINE_FUNCTION  \
   __attribute__((annotate("KOKKOS_FORCEINLINE_FUNCTION")))
+#define KOKKOS_RELOCATABLE_FUNCTION \
+  KOKKOS_IMPL_RELOCATABLE_FUNCTION  \
+  __attribute__((annotate("KOKKOS_RELOCATABLE_FUNCTION")))
 #else
 #define KOKKOS_FUNCTION KOKKOS_IMPL_FUNCTION
 #define KOKKOS_INLINE_FUNCTION KOKKOS_IMPL_INLINE_FUNCTION
 #define KOKKOS_FORCEINLINE_FUNCTION KOKKOS_IMPL_FORCEINLINE_FUNCTION
+#define KOKKOS_RELOCATABLE_FUNCTION KOKKOS_IMPL_RELOCATABLE_FUNCTION
```
crtrott, line 383: I think we should NOT define KOKKOS_RELOCATABLE_FUNCTION if RDC is off with CUDA, HIP and SYCL. If you mark a function as relocatable, I assume you mean it.

crtrott (changes requested):
I think we should not define the KOKKOS_RELOCTABLE_FUNCTION if relocatable code doesn't work. I.e. it will always be defined in host only builds, but will not exist if you compile CUDA, HIP or SYCL without relocatable device code enabled. What about OpenACC and OpenMPTarget? I guess they don't support it at all? Or they support it by default?

example/separable_compilation/cmake_example.cpp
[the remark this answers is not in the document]
dalg24: This is not a cmake_* example so I wouldn't necessarily expect it to follow what these do.

example/separable_compilation/cmake_example.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+  if (argc < 2) {
+    fprintf(stderr, "Usage: %s [<kokkos_options>] <size>\n", argv[0]);
+    Kokkos::finalize();
```
dalg24, line 30: This right here would be an argument for using ScopeGuard

example/separable_compilation/cmake_example.cpp
```diff
@@ -0,0 +1,64 @@
...
+    exit(1);
+  }
+
+  const long n = strtol(argv[1], nullptr, 10);
```
dalg24, line 34: Prefer `stol`

example/separable_compilation/cmake_example.cpp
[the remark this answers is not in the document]
dalg24: I would say drop this and the `<iostream>` header include.

core/src/Kokkos_Macros.hpp
dalg24: The current version poisoned the macro if RDC is not enabled, which yields better error messages and is compatible with host-only builds

masterleinad:
Retest this please.

masterleinad:
Retest this please.

example/separable_compilation/main.cpp
```diff
@@ -0,0 +1,55 @@
...
+//
+// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
+#include <Kokkos_Core.hpp>
+
+#include <cstdio>
```
dalg24, line 19:
```suggestion
```

example/separable_compilation/main.cpp
```diff
@@ -0,0 +1,55 @@
...
+  Kokkos::parallel_reduce(
+      n, KOKKOS_LAMBDA(const long i, long& lcount) { count_even(i, lcount); },
+      count);
+
+  double count_time = timer.seconds();
+  std::cout << "Parallel: " << count << ", " << count_time << "s\n";
+
+  timer.reset();
+
+  // Compare to a sequential loop.
```
dalg24, line 45: Did you try what we discussed that is whether comparing to an inlined version makes any sense at all?

masterleinad: RDC:
```
Number of even integers from 0 to 9
Parallel: 5, 0.0176898s
Sequential: 5, 4.769e-06s
Number of even integers from 0 to 99
Parallel: 50, 0.0175152s
Sequential: 50, 5.4222e-05s
Number of even integers from 0 to 999
Parallel: 500, 0.0180551s
Sequential: 500, 0.000495395s
Number of even integers from 0 to 9999
Parallel: 5000, 0.0179687s
Sequential: 5000, 0.00475208s
Number of even integers from 0 to 99999
Parallel: 50000, 0.018883s
Sequential: 50000, 0.0477251s
Number of even integers from 0 to 999999
Parallel: 500000, 0.0211372s
Sequential: 500000, 0.471865s
Number of even integers from 0 to 9999999
Parallel: 5000000, 0.0416765s
Sequential: 5000000, 4.71598s
Number of even integers from 0 to 99999999
Parallel: 50000000, 0.245804s
Sequential: 50000000, 47.2422s
```
no-rdc:
```
Number of even integers from 0 to 9
Parallel: 5, 0.0176316s
Sequential: 5, 4.708e-06s
Number of even integers from 0 to 99
Parallel: 50, 0.017936s
Sequential: 50, 5.9791e-05s
Number of even integers from 0 to 999
Parallel: 500, 0.0175954s
Sequential: 500, 0.000487341s
Number of even integers from 0 to 9999
Parallel: 5000, 0.0177417s
Sequential: 5000, 0.00474439s
Number of even integers from 0 to 99999
Parallel: 50000, 0.0187472s
Sequential: 50000, 0.0472494s
Number of even integers from 0 to 999999
Parallel: 500000, 0.0195275s
Sequential: 500000, 0.472715s
Number of even integers from 0 to 9999999
Parallel: 5000000, 0.0278132s
Sequential: 5000000, 4.72659s
Number of even integers from 0 to 99999999
Parallel: 50000000, 0.110623s
Sequential: 50000000, 47.1561s
```
Up to 10^6 elements there is hardly any difference after that it gets more than twice as slow.

dalg24: meaning there would be a good story to tell

masterleinad: For SYCL, there is barely any difference:
> 1.000000e+01 2.832492e-02s (RDC) vs. 1.710768e-02s (inline)
1.000000e+02 1.777367e-02s (RDC) vs. 1.779604e-02s (inline)
1.000000e+03 2.295095e-02s (RDC) vs. 2.303725e-02s (inline)
1.000000e+04 2.068359e-02s (RDC) vs. 2.069026e-02s (inline)
1.000000e+05 2.098447e-02s (RDC) vs. 2.057105e-02s (inline)
1.000000e+06 4.073293e-02s (RDC) vs. 4.044686e-02s (inline)
1.000000e+07 4.990597e-02s (RDC) vs. 4.962549e-02s (inline)
1.000000e+08 1.075163e-01s (RDC) vs. 1.074704e-01s (inline)

cmake/kokkos_arch.cmake
```diff
@@ -622,10 +622,15 @@ IF (KOKKOS_ENABLE_SYCL)
...
     DEFAULT -fsycl-unnamed-lambda
   )
-  IF (NOT KOKKOS_ENABLE_IMPL_SYCL_RELOCATABLE_DEVICE_CODE)
-    IF(KOKKOS_CXX_COMPILER_ID STREQUAL IntelLLVM AND KOKKOS_CXX_COMPILER_VERSION VERSION_LESS 2023.1.0)
-      MESSAGE(FATAL_ERROR "Kokkos_ENABLE_IMPL_SYCL_RELOCATABLE_DEVICE_CODE=OFF requires oneAPI 2023.1.0 or later")
+  IF(KOKKOS_CXX_COMPILER_ID STREQUAL IntelLLVM AND KOKKOS_CXX_COMPILER_VERSION VERSION_LESS 2023.1.0)
+    # Before oneAPI 2023.1.0 the compiler didn't support -f[no-]sycl-rdc
+    # but allowed separate compilation by default.
+    IF(NOT KOKKOS_ENABLE_SYCL_RELOCATABLE_DEVICE_CODE)
+      MESSAGE(FATAL_ERROR "Kokkos_ENABLE_SYCL_RELOCATABLE_DEVICE_CODE=OFF requires oneAPI 2023.1.0 or later")
```
crtrott, line 629: Does SYCL_EXTERN exist in older SYCL versions and does it work?

example/CMakeLists.txt
```diff
@@ -1,2 +1,7 @@
 KOKKOS_ADD_EXAMPLE_DIRECTORIES(query_device)
+IF(KOKKOS_ENABLE_CUDA_RELOCATABLE_DEVICE_CODE OR
+   KOKKOS_ENABLE_HIP_RELOCAtABLE_DEVICE_CODE OR
+   KOKKOS_ENABLE_SYCL_RELOCATABLE_DEVICE_CODE)
```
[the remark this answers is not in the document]
crtrott: done

crtrott (changes requested):
Two questions: did SYCL_EXTERNAL always work? Should we rename the example to: "relocatable_function" or so to be aligned with the CMake option, the Macro, and the compiler option used?

cmake/kokkos_arch.cmake
masterleinad: Yes, the CI is using oneAPI 2023.0.0 and builds and runs the example fine.

masterleinad:
> Should we rename the example to: "relocatable_function" or so to be aligned with the CMake option, the Macro, and the compiler option used?

Sure.

example/relocatable_function/main.cpp
```diff
@@ -0,0 +1,49 @@
...
+//@HEADER
+
+#include <Kokkos_Core.hpp>
+
+#include <iostream>
+
+KOKKOS_RELOCATABLE_FUNCTION void count_even(const long i, long& lcount);
+
+int main() {
+  Kokkos::ScopeGuard scope_guard;
```
dalg24, line 24: In examples we really should showcase passing argc and argv through.
Also now that you removed the early exit, it makes little sense to use ScopeGuard

dalg24 (approved):
Approving but would prefer if you fixed the initialization with argc and argv
Also the title line does not capture the addition of a new function annotation

masterleinad:
> Approving but would prefer if you fixed the initialization with argc and argv

I'm happy to address that and the other suggestions in a follow-up.

Issue #5998, opened 2023-03-20
`HIP`: `std::memcpy` compiler warning for virtual class


romintomasetti:
**Describe the bug**

Due to commit e114205066553c534c428a622cb59872213ff4f7, using a `KOKKOS_CLASS_LAMBDA` closure on `HIP` can lead to compiler error (actually a warning, treated as an error in our case due to `-Werror`).

Here is a minimal reproducer:
```c++
class TestBase
{
    public:
        virtual void apply(const size_t N) const = 0;

        virtual ~TestBase() = default;
};

class TestDerived : public TestBase
{
    public:
        KOKKOS_INLINE_FUNCTION
        int test(const int i) const
        {
            return i;
        };

        virtual void apply(const size_t N) const override final
        {
            KOKKOS_PARALLEL_FOR(
                // using 'requires hint heavy weight' is required to trigger the problematic code path
                Kokkos::Experimental::require(
                    Kokkos::RangePolicy<execution_space>(0,N),
                    Kokkos::Experimental::WorkItemProperty::HintHeavyWeight
                ),
                KOKKOS_CLASS_LAMBDA(const int i)
                {
                    this->test(i);
                }
            );
        }
};

/**
 * In include/HIP/Kokkos_HIP_KernelLaunch.hpp:418:26:
 *      error: source of this 'memcpy' call is a pointer to class containing a dynamic class 'TestDerived'; vtable pointer will be copied [-Werror,-Wdynamic-class-memaccess]
 *      std::memcpy(staging, &driver, sizeof(DriverType));
 */
TEST(test,test)
{
    auto t = std::make_shared<TestDerived>();

    t->apply(10);
}
```

and the associated compiler output:
```bash
In file included from /workspaces/HELM/tests/utils/test_KokkosClassLambda.cpp:5:
In file included from /opt/Trilinos/GNU-ROCm/include/Kokkos_Core.hpp:45:
In file included from /opt/Trilinos/GNU-ROCm/include/KokkosCore_Config_DeclareBackend.hpp:23:
In file included from /opt/Trilinos/GNU-ROCm/include/decl/Kokkos_Declare_HIP.hpp:28:
In file included from /opt/Trilinos/GNU-ROCm/include/HIP/Kokkos_HIP_Parallel_Range.hpp:24:
In file included from /opt/Trilinos/GNU-ROCm/include/HIP/Kokkos_HIP_BlockSize_Deduction.hpp:26:
[...]

PR #6000, opened 2023-03-20
Fix warning in some user code when using std::memcpy

Rombur:
This should fix the warning seen by user in https://github.com/kokkos/kokkos/issues/5998

### DIFF
--- a/core/src/HIP/Kokkos_HIP_KernelLaunch.hpp
+++ b/core/src/HIP/Kokkos_HIP_KernelLaunch.hpp
@@ -414,7 +414,8 @@ struct HIPParallelLaunchKernelInvoker<DriverType, LaunchBounds,
 
     // Copy functor (synchronously) to staging buffer in pinned host memory
     unsigned long *staging = hip_instance->constantMemHostStaging;
-    std::memcpy(staging, &driver, sizeof(DriverType));
+    std::memcpy(static_cast<void *>(staging),
+                static_cast<const void *>(&driver), sizeof(DriverType));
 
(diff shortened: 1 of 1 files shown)


core/src/HIP/Kokkos_HIP_KernelLaunch.hpp
```diff
@@ -414,7 +414,7 @@ struct HIPParallelLaunchKernelInvoker<DriverType, LaunchBounds,
 
     // Copy functor (synchronously) to staging buffer in pinned host memory
     unsigned long *staging = hip_instance->constantMemHostStaging;
-    std::memcpy(staging, &driver, sizeof(DriverType));
+    std::memcpy((void *)staging, (void *)&driver, sizeof(DriverType));
```
nliber, line 417: We should use C++ casts and be const correct (as the C-style cast on `&driver` isn't), as in
```c++
std::memcpy(static_cast<void*>(staging), static_cast<const void*>(&driver), sizeof(driver));
```

nliber (changes requested):

dalg24 (approved):

nliber (approved):

dalg24:
clang-format accident :)

dalg24:
Please cherry-pick into 4.0.01 patch release

crtrott:
Is this a problem in 3.7? I don't think so or? @Rombur Please cherry-pick to 4.0.01

Rombur:
It's not a problem in 3.7

PR #6004, opened 2023-03-22
simd: Fixed an incorrectly returning size for uint64_t in avx2

ldh4:
Following up from https://github.com/kokkos/kokkos/pull/5913#discussion_r1137432586.

This fixes an incorrectly returning width size for simd ABI with uint64_t in AVX2 backend.

The added unit test in this PR is temporary that is to quickly test `size()` for each ABI. Because current unit-tests in kokkos simd are set up to only test for `double` for all ABIs, existing tests do not expose this bug. This unit test will be refactored and be integrated with rest of unit tests as part of #5913.

### DIFF
--- a/simd/unit_tests/TestSIMD.cpp
+++ b/simd/unit_tests/TestSIMD.cpp
@@ -505,7 +505,8 @@ TEST(simd, test_size) {
 #else
   constexpr auto width = 1;
   using Abi            = Kokkos::Experimental::simd_abi::scalar;
-  auto uint32_abi_size = Kokkos::Experimental::simd<std::uint32_t, Abi>::size();
+  static_assert(width ==
+                Kokkos::Experimental::simd<std::uint32_t, Abi>::size());
 #endif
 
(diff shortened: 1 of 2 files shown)


ibaned (approved):

simd/unit_tests/TestSIMD.cpp
```diff
@@ -486,3 +486,62 @@ TEST(simd, device) {
   Kokkos::parallel_for(Kokkos::RangePolicy<Kokkos::IndexType<int>>(0, 1),
                        simd_device_functor());
 }
+
+TEST(simd, test_size) {
```
dalg24, line 490: Any good reason to delay assertions to runtime.  These can be done at compile-time via static_assert.

ldh4: Converted to compile time checks. [6f8b0df](https://github.com/kokkos/kokkos/pull/6004/commits/6f8b0dfe179b37f4c9e5a7dc1e7e3e2b636473aa)

dalg24 (approved):

Issue #6005, opened 2023-03-22
Allocation of `Kokkos::View` given a `space` instance and `Cuda UVM`


romintomasetti:
**Describe the bug**

Take the following example (with `Cuda UVM` enabled):
```c++
// Get a Cuda stream different from the default stream
const auto space = Kokkos::Experimental::partition_space(Kokkos::Cuda{}, 1);
// Build a view, giving the stream
Kokkos::View<double*, Kokkos::Device<Kokkos::Cuda, Kokkos::CudaUVMSpace>> view(
    Kokkos::view_alloc(space, "my label"),
    size
);
```

In the general case, it seems it is safe to access the view from host at this point because of the following lines doing a `fence` on the default stream, and the default stream being by default (`legacy mode`) "device-wide synchronizing":
https://github.com/kokkos/kokkos/blob/aa1f48f3172069f212ed2c97b74b786999b26f17/core/src/Kokkos_View.hpp#L1420-L1442

However, if one enables the `per-thread` default stream mode, I think the aforementioned lines are still fencing the default stream, but now the default stream fence does not affect the custom stream used in the allocation. It means that accessing the view from host can be hazardous.

A potential fix could be to determine if a `space` instance was given, and fence on it instead of fencing on the default stream:
```c++
if constexpr (alloc_prop::has_execution_space) {
    const auto& exec_space = Impl::get_property<Impl::ExecutionSpaceTag>(arg_prop);
    exec_space.fence(...);
} else {
    typename traits::device_type::memory_space::execution_space().fence(
          "Kokkos::View<...>::View: fence before allocating UVM");
}
```

Also tagging @maartenarnst as we work on that together.

PhilMiller:
I think your proposed fix does the right thing when passing an explicit execution space instance.

The fence on the default instance may actually not be appropriate anymore if we actually rely on device-wide synchronization, because the default-constructed `Kokkos::Cuda` instance does not use the 'default' CUDA stream, but rather a separate one we create internally, precisely to avoid the implicit global synchronization semantics. When we or users truly want device-wide synchronization, rather than synchronization of the operations on the particular stream, that requires calling `Kokkos::fence()` rather than `exec_space.fence()`.

masterleinad:
@romintomasetti Would you mind creating a pull request for this? Note that `prop_copy` already contains the appropriate execution space instance that should be used to fence.

PhilMiller:
[...]

Issue #4228, opened 2021-08-11
cudaMallocAsync not supported by UCX, may cause failure in OpenMPI+Kokkos+cuda applications


lfmeadow:
See UCX issue https://github.com/openucx/ucx/issues/7194
cudaMallocAsync was added to Kokkos/Cuda sometime in the last few months and results in a application crash in lammps+kokkos+cuda using OpenMPI+UCX.
The UCX issue comments mention 3 PRs to UCX that should resolve the problem.
In the meantime the following patch to Kokkos will fix it:
```diff
diff --git a/core/src/Cuda/Kokkos_CudaSpace.cpp b/core/src/Cuda/Kokkos_CudaSpace.cpp
index 39da8de6b..13c35eb87 100644
--- a/core/src/Cuda/Kokkos_CudaSpace.cpp
+++ b/core/src/Cuda/Kokkos_CudaSpace.cpp
@@ -229,7 +229,7 @@ void *CudaSpace::impl_allocate(
 
 #ifndef CUDART_VERSION
 #error CUDART_VERSION undefined!
-#elif (CUDART_VERSION >= 11020)
+#elif 0 // (CUDART_VERSION >= 11020) UCX issue
   cudaError_t error_code;
   if (arg_alloc_size >= memory_threshold_g) {
     error_code = cudaMallocAsync(&ptr, arg_alloc_size, 0);
@@ -358,7 +358,7 @@ void CudaSpace::impl_deallocate(
   try {
 #ifndef CUDART_VERSION
 #error CUDART_VERSION undefined!
-#elif (CUDART_VERSION >= 11020)
+#elif 0 // (CUDART_VERSION >= 11020) UCX issue
     if (arg_alloc_size >= memory_threshold_g) {
       CUDA_SAFE_CALL(cudaDeviceSynchronize());
       CUDA_SAFE_CALL(cudaFreeAsync(arg_alloc_ptr, 0));
```

maxpkatz:
#4026 is the PR that does this (@matt-stack). We should make the use of cudaMallocAsync opt-in with a CMake flag at least until UCX fully supports that allocator.

BenWibking:
This affects my code as well. It looks like UCX 1.14rc1 may fix this issue: https://github.com/openucx/ucx/blob/d83ef403646473d76be42cbef03fb38652169f78/NEWS#L49.

Do you have a reproducer for the issue that can be tested with the latest UCX?

lfmeadow:
[...]

PR #6008, opened 2023-03-23
view(uvm): don't fence during view's allocation (#6005)

romintomasetti:
This PR is a fix for
* #6005

Basically, when `CudaUVM` is enabled, there is a fence *before* and *after* the view's allocation (whether the `space` instance was given via `arg_prop` or not).

However, the fence was not performed on the *space* instance the user provided (if any).

According to the ["poll" below](https://github.com/kokkos/kokkos/pull/6008#issuecomment-1492054336), it was decided to
> enforce `Kokkos` semantics: Don't ever fence for the view initialization when an execution space instance is given

Therefore, any additional fence that was added for `CudaUVM` in view's allocation has been removed. Note that fencing still occurs in the backend implementations.

### DIFF
--- a/containers/src/Kokkos_DualView.hpp
+++ b/containers/src/Kokkos_DualView.hpp
@@ -239,7 +239,8 @@ class DualView : public ViewTraits<DataType, Arg1Type, Arg2Type, Arg3Type> {
       : modified_flags(t_modified_flags("DualView::modified_flags")),
         d_view(arg_prop, n0, n1, n2, n3, n4, n5, n6, n7) {
     // without UVM, host View mirrors
-    if (Kokkos::Impl::has_type<Impl::WithoutInitializing_t, P...>::value)
+    if constexpr (Kokkos::Impl::has_type<Impl::WithoutInitializing_t,
+                                         P...>::value)
       h_view = Kokkos::create_mirror_view(Kokkos::WithoutInitializing, d_view);
     else
       h_view = Kokkos::create_mirror_view(d_view);
@@ -576,8 +577,8 @@ class DualView : public ViewTraits<DataType, Arg1Type, Arg2Type, Arg3Type> {
         impl_report_host_sync();
       }
     }
-    if (std::is_same<typename t_host::memory_space,
-                     typename t_dev::memory_space>::value) {
+    if constexpr (std::is_same<typename t_host::memory_space,
+                               typename t_dev::memory_space>::value) {
       typename t_dev::execution_space().fence(
           "Kokkos::DualView<>::sync: fence after syncing DualView");
       typename t_host::execution_space().fence(
--- a/containers/src/Kokkos_DynRankView.hpp
+++ b/containers/src/Kokkos_DynRankView.hpp
@@ -1089,36 +1089,11 @@ class DynRankView : public ViewTraits<DataType, Properties...> {
           "execution space");
     }
 
-//------------------------------------------------------------
-#if defined(KOKKOS_ENABLE_CUDA)
-    // If allocating in CudaUVMSpace must fence before and after
-    // the allocation to protect against possible concurrent access
-    // on the CPU and the GPU.
-    // Fence using the trait's executon space (which will be Kokkos::Cuda)
-    // to avoid incomplete type errors from usng Kokkos::Cuda directly.
-    if (std::is_same<Kokkos::CudaUVMSpace,
-                     typename traits::device_type::memory_space>::value) {
-      typename traits::device_type::memory_space::execution_space().fence(
-          "Kokkos::DynRankView<>::DynRankView: fence before UVM allocation");
-    }
-#endif
-    //------------------------------------------------------------
-
     Kokkos::Impl::SharedAllocationRecord<>* record = m_map.allocate_shared(
         prop_copy,
         Impl::DynRankDimTraits<typename traits::specialize>::
             template createLayout<traits, P...>(arg_prop, arg_layout),
         Impl::ViewCtorProp<P...>::has_execution_space);
-
-//------------------------------------------------------------
-#if defined(KOKKOS_ENABLE_CUDA)
-    if (std::is_same<Kokkos::CudaUVMSpace,
-                     typename traits::device_type::memory_space>::value) {
-      typename traits::device_type::memory_space::execution_space().fence(
-          "Kokkos::DynRankView<>::DynRankView: fence after UVM allocation");
-    }
-#endif
-    //------------------------------------------------------------
 
     // Setup and initialization complete, start tracking
     m_track.assign_allocated_record_to_uninitialized(record);
--- a/containers/unit_tests/TestWithoutInitializing.hpp
+++ b/containers/unit_tests/TestWithoutInitializing.hpp
@@ -24,6 +24,19 @@
 
 #include <../../core/unit_test/tools/include/ToolTestingUtilities.hpp>
 
+/// Some tests are skipped for @c CudaUVM memory space.
+/// @todo To be revised according to the future of @c KOKKOS_ENABLE_CUDA_UVM.
+///@{
+#ifdef KOKKOS_ENABLE_CUDA
+#define GTEST_SKIP_IF_CUDAUVM_MEMORY_SPACE                            \
+  if constexpr (std::is_same_v<typename TEST_EXECSPACE::memory_space, \
+                               Kokkos::CudaUVMSpace>)                 \
+    GTEST_SKIP() << "skipping since CudaUVMSpace requires additional fences";
+#else
+#define GTEST_SKIP_IF_CUDAUVM_MEMORY_SPACE
+#endif
+///@}
+
 TEST(TEST_CATEGORY, resize_realloc_no_init_dualview) {
   using namespace Kokkos::Test::Tools;
   listen_tool_events(Config::DisableAll(), Config::EnableKernels());
@@ -125,11 +138,7 @@ TEST(TEST_CATEGORY, resize_exec_space_dualview) {
 }
 
 TEST(TEST_CATEGORY, realloc_exec_space_dualview) {
-#ifdef KOKKOS_ENABLE_CUDA
-  if (std::is_same<typename TEST_EXECSPACE::memory_space,
-                   Kokkos::CudaUVMSpace>::value)
-    GTEST_SKIP() << "skipping since CudaUVMSpace requires additional fences";
-#endif
+  GTEST_SKIP_IF_CUDAUVM_MEMORY_SPACE
 
   using namespace Kokkos::Test::Tools;
   listen_tool_events(Config::DisableAll(), Config::EnableFences());
@@ -221,11 +230,8 @@ TEST(TEST_CATEGORY, resize_exec_space_dynrankview) {
 }
 
 TEST(TEST_CATEGORY, realloc_exec_space_dynrankview) {
-#ifdef KOKKOS_ENABLE_CUDA
-  if (std::is_same<typename TEST_EXECSPACE::memory_space,
-                   Kokkos::CudaUVMSpace>::value)
-    GTEST_SKIP() << "skipping since CudaUVMSpace requires additional fences";
-#endif
+  GTEST_SKIP_IF_CUDAUVM_MEMORY_SPACE
+
 // FIXME_THREADS The Threads backend fences every parallel_for
 #ifdef KOKKOS_ENABLE_THREADS
   if (std::is_same<TEST_EXECSPACE, Kokkos::Threads>::value)
@@ -363,11 +369,8 @@ TEST(TEST_CATEGORY, resize_exec_space_scatterview) {
 }
 
 TEST(TEST_CATEGORY, realloc_exec_space_scatterview) {
-#ifdef KOKKOS_ENABLE_CUDA
-  if (std::is_same<typename TEST_EXECSPACE::memory_space,
-                   Kokkos::CudaUVMSpace>::value)
-    GTEST_SKIP() << "skipping since CudaUVMSpace requires additional fences";
-#endif
+  GTEST_SKIP_IF_CUDAUVM_MEMORY_SPACE
+
 // FIXME_THREADS The Threads backend fences every parallel_for
 #ifdef KOKKOS_ENABLE_THREADS
   if (std::is_same<typename TEST_EXECSPACE, Kokkos::Threads>::value)
@@ -477,11 +480,8 @@ TEST(TEST_CATEGORY, create_mirror_no_init_dynrankview_viewctor) {
 }
 
 TEST(TEST_CATEGORY, create_mirror_view_and_copy_dynrankview) {
-#ifdef KOKKOS_ENABLE_CUDA
-  if (std::is_same<typename TEST_EXECSPACE::memory_space,
-                   Kokkos::CudaUVMSpace>::value)
-    return;
-#endif
+  GTEST_SKIP_IF_CUDAUVM_MEMORY_SPACE
+
   using namespace Kokkos::Test::Tools;
   listen_tool_events(Config::DisableAll(), Config::EnableKernels(),
                      Config::EnableFences());
@@ -584,11 +584,8 @@ TEST(TEST_CATEGORY, create_mirror_no_init_offsetview_view_ctor) {
 }
 
 TEST(TEST_CATEGORY, create_mirror_view_and_copy_offsetview) {
-#ifdef KOKKOS_ENABLE_CUDA
-  if (std::is_same<typename TEST_EXECSPACE::memory_space,
-                   Kokkos::CudaUVMSpace>::value)
-    return;
-#endif
+  GTEST_SKIP_IF_CUDAUVM_MEMORY_SPACE
+
   using namespace Kokkos::Test::Tools;
   listen_tool_events(Config::DisableAll(), Config::EnableKernels(),
                      Config::EnableFences());
@@ -659,11 +656,8 @@ TEST(TEST_CATEGORY, create_mirror_no_init_dynamicview) {
 }
 
 TEST(TEST_CATEGORY, create_mirror_view_and_copy_dynamicview) {
-#ifdef KOKKOS_ENABLE_CUDA
-  if (std::is_same<typename TEST_EXECSPACE::memory_space,
-                   Kokkos::CudaUVMSpace>::value)
-    return;
-#endif
+  GTEST_SKIP_IF_CUDAUVM_MEMORY_SPACE
+
   using namespace Kokkos::Test::Tools;
   listen_tool_events(Config::DisableAll(), Config::EnableKernels(),
                      Config::EnableFences());
--- a/core/src/Cuda/Kokkos_CudaSpace.cpp
+++ b/core/src/Cuda/Kokkos_CudaSpace.cpp
@@ -172,10 +172,11 @@ void *impl_allocate_common(const Cuda &exec_space, const char *arg_label,
     if (exec_space_provided) {
       cudaStream_t stream = exec_space.cuda_stream();
       error_code          = cudaMallocAsync(&ptr, arg_alloc_size, stream);
-      KOKKOS_IMPL_CUDA_SAFE_CALL(cudaStreamSynchronize(stream));
+      exec_space.fence("Kokkos::Cuda: backend fence after async malloc");
     } else {
       error_code = cudaMallocAsync(&ptr, arg_alloc_size, 0);
-      KOKKOS_IMPL_CUDA_SAFE_CALL(cudaDeviceSynchronize());
+      Impl::cuda_device_synchronize(
+          "Kokkos::Cuda: backend fence after async malloc");
     }
   } else {
     error_code = cudaMalloc(&ptr, arg_alloc_size);
@@ -324,9 +325,11 @@ void CudaSpace::impl_deallocate(
 #error CUDART_VERSION undefined!
 #elif (defined(KOKKOS_ENABLE_IMPL_CUDA_MALLOC_ASYNC) && CUDART_VERSION >= 11020)
     if (arg_alloc_size >= memory_threshold_g) {
-      KOKKOS_IMPL_CUDA_SAFE_CALL(cudaDeviceSynchronize());
+      Impl::cuda_device_synchronize(
+          "Kokkos::Cuda: backend fence before async free");
       KOKKOS_IMPL_CUDA_SAFE_CALL(cudaFreeAsync(arg_alloc_ptr, 0));
-      KOKKOS_IMPL_CUDA_SAFE_CALL(cudaDeviceSynchronize());
+      Impl::cuda_device_synchronize(
+          "Kokkos::Cuda: backend fence after async free");
     } else {
       KOKKOS_IMPL_CUDA_SAFE_CALL(cudaFree(arg_alloc_ptr));
     }
--- a/core/src/Kokkos_View.hpp
+++ b/core/src/Kokkos_View.hpp
@@ -1418,33 +1418,8 @@ class View : public ViewTraits<DataType, Properties...> {
         std::is_same<typename traits::specialize, void>::value, i0, i1, i2, i3,
         i4, i5, i6, i7, alloc_name);
 
-//------------------------------------------------------------
-#if defined(KOKKOS_ENABLE_CUDA)
-    // If allocating in CudaUVMSpace must fence before and after
-    // the allocation to protect against possible concurrent access
-    // on the CPU and the GPU.
-    // Fence using the trait's execution space (which will be Kokkos::Cuda)
-    // to avoid incomplete type errors from using Kokkos::Cuda directly.
-    if (std::is_same<Kokkos::CudaUVMSpace,
-                     typename traits::device_type::memory_space>::value) {
-      typename traits::device_type::memory_space::execution_space().fence(
-          "Kokkos::View<...>::View: fence before allocating UVM");
-    }
-#endif
-    //------------------------------------------------------------
-
     Kokkos::Impl::SharedAllocationRecord<>* record = m_map.allocate_shared(
         prop_copy, arg_layout, Impl::ViewCtorProp<P...>::has_execution_space);
-
-//------------------------------------------------------------
-#if defined(KOKKOS_ENABLE_CUDA)
-    if (std::is_same<Kokkos::CudaUVMSpace,
-                     typename traits::device_type::memory_space>::value) {
-      typename traits::device_type::memory_space::execution_space().fence(
-          "Kokkos::View<...>::View: fence after allocating UVM");
-    }
-#endif
-    //------------------------------------------------------------
 
     // Setup and initialization complete, start tracking
     m_track.m_tracker.assign_allocated_record_to_uninitialized(record);
--- a/containers/src/Kokkos_OffsetView.hpp
+++ b/containers/src/Kokkos_OffsetView.hpp
@@ -1191,34 +1191,9 @@ class OffsetView : public ViewTraits<DataType, Properties...> {
           "execution space");
     }
 
-    //------------------------------------------------------------
-#if defined(KOKKOS_ENABLE_CUDA)
-    // If allocating in CudaUVMSpace must fence before and after
-    // the allocation to protect against possible concurrent access
-    // on the CPU and the GPU.
-    // Fence using the trait's executon space (which will be Kokkos::Cuda)
-    // to avoid incomplete type errors from usng Kokkos::Cuda directly.
-    if (std::is_same<Kokkos::CudaUVMSpace,
-                     typename traits::device_type::memory_space>::value) {
-      typename traits::device_type::memory_space::execution_space().fence(
-          "Kokkos::OffsetView::OffsetView(): fence before UVM allocation");
-    }
-#endif
-    //------------------------------------------------------------
-
     Kokkos::Impl::SharedAllocationRecord<>* record = m_map.allocate_shared(
         prop_copy, arg_layout,
         Kokkos::Impl::ViewCtorProp<P...>::has_execution_space);
-
-    //------------------------------------------------------------
-#if defined(KOKKOS_ENABLE_CUDA)
-    if (std::is_same<Kokkos::CudaUVMSpace,
-                     typename traits::device_type::memory_space>::value) {
-      typename traits::device_type::memory_space::execution_space().fence(
-          "Kokkos::OffsetView::OffsetView(): fence after UVM allocation");
-    }
-#endif
-    //------------------------------------------------------------
 
     // Setup and initialization complete, start tracking
     m_track.assign_allocated_record_to_uninitialized(record);


containers/src/Kokkos_DynRankView.hpp
```diff
@@ -1089,19 +1089,8 @@ class DynRankView : public ViewTraits<DataType, Properties...> {
...
-    // to avoid incomplete type errors from usng Kokkos::Cuda directly.
-    if (std::is_same<Kokkos::CudaUVMSpace,
-                     typename traits::device_type::memory_space>::value) {
-      typename traits::device_type::memory_space::execution_space().fence(
-          "Kokkos::DynRankView<>::DynRankView: fence before UVM allocation");
-    }
-#endif
+    //------------------------------------------------------------
+    Impl::allocation_fence_if_shared_space<traits>(arg_prop, "Kokkos::DynRankView<>::DynRankView: fence before UVM allocation");
     //------------------------------------------------------------
```
masterleinad, line 1094: I don't think this fence is necessary. We can't access the new allocation before it was created.

containers/src/Kokkos_DynRankView.hpp
```diff
@@ -1110,14 +1099,8 @@ class DynRankView : public ViewTraits<DataType, Properties...> {
...
-//------------------------------------------------------------
-#if defined(KOKKOS_ENABLE_CUDA)
-    if (std::is_same<Kokkos::CudaUVMSpace,
-                     typename traits::device_type::memory_space>::value) {
-      typename traits::device_type::memory_space::execution_space().fence(
-          "Kokkos::DynRankView<>::DynRankView: fence after UVM allocation");
-    }
-#endif
+    //------------------------------------------------------------
+    Impl::allocation_fence_if_shared_space<traits>(arg_prop,"Kokkos::DynRankView<>::DynRankView: fence after UVM allocation");
```
masterleinad, line 1103: I would probably have a fence here for all host-accessible allocations. This would cover the host-pinned spaces and shared spaces across all GPU backends and `Kokkos::HostSpace`. Note that the only asynchronous host execution space is `HPX`.

containers/src/Kokkos_DynRankView.hpp
romintomasetti: OK, so let's remove the 3 "fence before allocation" (`View`, `OffsetView` and `DynRankView`) :person_fencing:

containers/src/Kokkos_DynRankView.hpp
romintomasetti: Indeed, it would make sense to do so for any host-accessible allocation.

How would you check that? Would something like the following work?

Instead of
```c++
template <typename traits_t, typename alloc_prop_t>
requires std::same_as<Kokkos::CudaUVMSpace,typename traits_t::device_type::memory_space>
constexpr inline void allocation_fence_if_shared_space(alloc_prop_t&& prop, const char* msg)
{ ... }
#endif
```
we would write
```c++
template <typename traits_t, typename alloc_prop_t>
constexpr inline
std::enable_if_t<Kokkos::SpaceAccessibility<Kokkos::Serial, typename traits_t::device_type::memory_space>::accessible>
allocation_fence_if_shared_space(alloc_prop_t&& prop, const char* msg)
{ ... }
```

(I noticed you're not on C++20 yet, so I'll need to remove the `requires` I guess)

masterleinad: I would replace
```C++
 if (std::is_same<Kokkos::CudaUVMSpace,
                     typename traits::device_type::memory_space>::value) {
      typename traits::device_type::memory_space::execution_space().fence(
          "Kokkos::DynRankView<>::DynRankView: fence after UVM allocation");
    }
```
with
```C++
 if (Kokkos::SpaceAccessibility<Kokkos::HostSpace, typename traits::device_type::memory_space>::accessible) {
      Impl::get_property<Impl::ExecutionSpaceTag>(prop_copy).fence(
          "Kokkos::DynRankView<>::DynRankView: fence after host-accessible allocation");
    }
```
since `Serial` is not necessarily available. Instead of `HostSpace`, `DefaultHostExecutionSpace` would also work. Also, note that `prop_copy` already contains the execution space instance you want to use.
Personally, I find it clearer to leave the fence inline but maybe you can make a point for creating a separate function. 🙂

core/src/Kokkos_View.hpp
```diff
@@ -101,6 +101,36 @@ void runtime_check_rank(const size_t rank, const size_t dyn_rank,
...
+ * If a space instance is provided, it is used instead.
+ *
+ * @note Fence using the trait's execution space (which will be @c Kokkos::Cuda)
+ *       to avoid incomplete type errors from using @c Kokkos::Cuda directly.
+ */
+///@{
+#if defined(KOKKOS_ENABLE_CUDA)
+template <typename traits_t, typename alloc_prop_t>
+requires std::same_as<Kokkos::CudaUVMSpace,typename traits_t::device_type::memory_space>
+constexpr inline void allocation_fence_if_shared_space(alloc_prop_t&& prop, const char* msg)
```
dalg24, line 121: The `constexpr` is superfluous.  This is clearly not usable in constant expressions.

romintomasetti: Yes, I noticed. Removed :heavy_check_mark:

core/src/Kokkos_View.hpp
```diff
@@ -101,6 +101,36 @@ void runtime_check_rank(const size_t rank, const size_t dyn_rank,
...
+ * By default, fencing is done on the default space instance.
+ * If a space instance is provided, it is used instead.
+ *
+ * @note Fence using the trait's execution space (which will be @c Kokkos::Cuda)
+ *       to avoid incomplete type errors from using @c Kokkos::Cuda directly.
+ */
+///@{
+#if defined(KOKKOS_ENABLE_CUDA)
+template <typename traits_t, typename alloc_prop_t>
+requires std::same_as<Kokkos::CudaUVMSpace,typename traits_t::device_type::memory_space>
```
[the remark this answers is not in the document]
romintomasetti: Indeed (see https://github.com/kokkos/kokkos/pull/6008#discussion_r1146244889).

I'll use `std::enable_if_t` instead :heavy_check_mark:

containers/src/Kokkos_DynRankView.hpp
romintomasetti: > I would probably have a fence here for all host-accessible allocations. This would cover the host-pinned spaces and shared spaces across all GPU backends and `Kokkos::HostSpace`. Note that the only asynchronous host execution space is `HPX`.

Also, note that the current approach (fencing if allocation is host-accessible) will also fence if the view is allocated in the host space.

Also, we should change the message attached to the fence (`fence after allocating UVM` is not so relevant anymore). What about `fence after host-accessible memory allocation`?

romintomasetti: @masterleinad It seems that tests are failingg because, as mentioned above, now a fence is added e.g. when allocations are done in the host memory space. As this was not the case before, tests are failing.

It looks like we have somehow changed the behaviour e.g. for OpemMP.

What do you suggest? Changing the tests expectations or modify the circumstances under which fencing is required after allocation?

masterleinad: I would change the tests but we should also see what others think. In the end, we are trying to preserve behavior that is inconsistent with the rest of Kokkos's interfaces. If you give an execution space, the call must be assumed to be asynchronous and the result can only be accessed after fencing.

romintomasetti: If I understand it correctly, if we pass in a `space`, the view allocation shouldn't fence since, semantically, it's the user's duty to make appropriate fencing when accessing the view? (not sure I got this right)

core/src/Kokkos_View.hpp
```diff
@@ -101,6 +102,46 @@ void runtime_check_rank(const size_t rank, const size_t dyn_rank,
...
+}
+
+template <typename traits_t, typename alloc_prop_t>
+constexpr inline
+std::enable_if_t<
+  ! Kokkos::SpaceAccessibility<
+    Kokkos::HostSpace,
+    typename traits_t::device_type::memory_space
+  >::accessible>
+allocation_fence_if_shared_space(alloc_prop_t&&, const char *) {}
```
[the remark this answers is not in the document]
romintomasetti: Thanks, I made the change. I just understood what you meant with "`prop_copy` already contains the space you want to use", so `prop_copy` is passed to `fence_for_host_accessible_space`, not `arg_prop` :)

containers/src/Kokkos_DynRankView.hpp
masterleinad: I'll try to bring it up at the next developer meeting.

masterleinad:
The developer meeting asked for more clarification of what we are discussing:

Consider
```C++
Kokkos::Cuda exec_space;
Kokkos::View<int*, Kokkos::CudaUVMSpace> view(Kokkos::view_alloc(exec_space, "view"), 10);
view[0] = 2;
```
should this code work or do we require the user to fence?
What about
```C++
Kokkos::Experimental::HPX exec_space;
Kokkos::View<int*, Kokkos::HostSpace> view(Kokkos::view_alloc(exec_space, "view"), 10);
view[0] = 2;
```
? Currently, we are fencing the default execution space instance(!) before allocation and after (potential initialization) only for `Kokkos::CudaUVMSpace`. Note that fences for the allocation itself are in the backend implementations.
I would think that we can get rid of the first fence for sure. Can we also get rid of the second one in case the user provided an execution space instance? Do we care to check for `Kokkos::WithoutInitializing` for this fence additionally?

dalg24:
> The developer meeting asked for more clarification of what we are discussing:
> 
> Consider
> 
> ```c++
> Kokkos::Cuda exec_space;
> Kokkos::View<int*, Kokkos::CudaUVMSpace> view(Kokkos::view_alloc(exec_space, "view"), 10);
> view[0] = 2;
> ```
> 
> should this code work or do we require the user to fence? What about
> 
> ```c++
> Kokkos::Experimental::HPX exec_space;
> Kokkos::View<int*, Kokkos::HostSpace> view(Kokkos::view_alloc(exec_space, "view"), 10);
> view[0] = 2;
> ```

In my opinion, yes both need fences.

romintomasetti:
> The developer meeting asked for more clarification of what we are discussing:
> 
> Consider
> 
> ```c++
> Kokkos::Cuda exec_space;
> Kokkos::View<int*, Kokkos::CudaUVMSpace> view(Kokkos::view_alloc(exec_space, "view"), 10);
> view[0] = 2;
> ```
> 
> should this code work or do we require the user to fence? What about
> 
> ```c++
> Kokkos::Experimental::HPX exec_space;
> Kokkos::View<int*, Kokkos::HostSpace> view(Kokkos::view_alloc(exec_space, "view"), 10);
> view[0] = 2;
> ```
> 
> ? Currently, we are fencing the default execution space instance(!) before allocation and after (potential initialization) only for `Kokkos::CudaUVMSpace`. Note that fences for the allocation itself are in the backend implementations. I would think that we can get rid of the first fence for sure. Can we also get rid of the second one in case the user provided an execution space instance? Do we care to check for `Kokkos::WithoutInitializing` for this fence additionally?

Hi @dalg24 @masterleinad !

We tried to think about this a bit more. Here is a table that show what we have in mind. Note the introduction of the **opt-in** `Kokkos::Async` flag, that allows the user to say "I'm good with async".

In our opinion, the behavior for a shared memory space should be similar to any other one.

| space instance | `Kokkos::view_alloc` | expected behavior |
|----| ---|---|
| no space instance | :x: | Use default space instance, and fence it at the end of `Kokkos::View` constructor (view is 'ready') |
| no space instance | `Kokkos::Async` | Use default space instance, don't fence it at the end of the constructor (view might not be 'ready' on host, but is ready on the FIFO queue of the default `space` instance) |
| space instance provided |:x: | Use provided `space` instance and fence it at the end of the constructor (view is 'ready') |
| space instance provided | `Kokkos::Async` | Use provided `space` instance and don't fence it (view might not be 'ready' on host but is ready on the FIFO queue of the provided `space` instance)  |

core/src/Cuda/Kokkos_CudaSpace.cpp
romintomasetti, line 1: @dalg24 @masterleinad 

We also wanted to bring your attention on these lines. They are actual fences, but for now resort on the `CUDA API`. Therefore, they don't tell the fence hook that a fence is happening.

Do you agree with these changes?

Rombur:
> Note the introduction of the opt-in Kokkos::Async flag, that allows the user to say "I'm good with async".

There is no need for this flag. If you give an execution space , you agree that the operation may be asynchronous. If you don't provide an execution space. we should fence for you. If you provide an execution space, then it's your responsibility.

masterleinad:
Let's just do a simple poll that might decide pretty much already how we want to proceed:
- enforce `Kokkos` semantics: Don't ever fence for the view initialization when an execution space instance is given 🚀 
- fence in some cases even if an execution space instance is given, e.g.  for UVM allocations ⭐

romintomasetti:
Hi @masterleinad ! :wave:

I guess [your poll](https://github.com/kokkos/kokkos/pull/6008#issuecomment-1492054336) is unanimous: `Kokkos` semantics should always be enforced, which in this case implies this rule:
> Don't ever fence for the view initialization when an execution space instance is given

If you agree, I'll make the changes.

core/src/Cuda/Kokkos_CudaSpace.cpp
masterleinad: Yes, we should probably also use the fence interface here. I'll leave it up to you if you want to tackle this in this pull request, too, or in another one since this might break some tests that check for fences explicitly.

containers/src/Kokkos_DynRankView.hpp
romintomasetti: I'm closing this thread, because it was decided not to fence before allocation (https://github.com/kokkos/kokkos/pull/6008#discussion_r1146208102).

containers/unit_tests/TestWithoutInitializing.hpp
```diff
@@ -24,6 +24,12 @@
 
 #include <../../core/unit_test/tools/include/ToolTestingUtilities.hpp>
 
+//! Get fence envent message for zeromemset
```
masterleinad, line 27:
```suggestion
//! Get fence event message for ZeroMemset
```

romintomasetti: Indeed, and done ;)

core/src/Cuda/Kokkos_CudaSpace.cpp
```diff
@@ -172,10 +172,10 @@ void *impl_allocate_common(const Cuda &exec_space, const char *arg_label,
     if (exec_space_provided) {
       cudaStream_t stream = exec_space.cuda_stream();
       error_code          = cudaMallocAsync(&ptr, arg_alloc_size, stream);
-      KOKKOS_IMPL_CUDA_SAFE_CALL(cudaStreamSynchronize(stream));
+      exec_space.fence("Kokkos::Cuda: backend fence after async malloc");
```
PhilMiller, line 175: Note that this case will need to be tested manually, since `KOKKOS_ENABLE_IMPL_CUDA_MALLOC_ASYNC` is defaulted off in the configuration due to #4228

PhilMiller: I think it looks right, though, and makes things closer to what @tcclevenger is doing in #5989. Hopefully this doesn't create a conflict.

core/src/Cuda/Kokkos_Cuda_Instance.cpp
```diff
@@ -132,13 +132,11 @@ void cuda_device_synchronize(const std::string &name) {
...
       // cudaDeviceSynchronize in device code
       [] __host__() {  // TODO: correct device ID
-        KOKKOS_IMPL_CUDA_SAFE_CALL(cudaDeviceSynchronize());
-      });
 #else
       []() {  // TODO: correct device ID
+#endif
         KOKKOS_IMPL_CUDA_SAFE_CALL(cudaDeviceSynchronize());
       });
-#endif
```
PhilMiller, line 141: This isn't really changing the logic, and again may conflict with #5989. I'd suggest reverting this one.

core/unit_test/TestWithoutInitializing.hpp
```diff
@@ -104,7 +104,7 @@ TEST(TEST_CATEGORY, realloc_exec_space) {
       [&](BeginFenceEvent event) {
         if ((event.descriptor().find("Debug Only Check for Execution Error") !=
              std::string::npos) ||
-            (event.descriptor().find("HostSpace fence") != std::string::npos))
+            (event.descriptor().find(std::string(TEST_EXECSPACE{}.name()).append(": ZeroMemset for")) != std::string::npos))
```
PhilMiller, line 107: I think the string descriptor and the type being matched against don't line up now?

core/src/Cuda/Kokkos_CudaSpace.cpp
romintomasetti: > Note that this case will need to be tested manually, since `KOKKOS_ENABLE_IMPL_CUDA_MALLOC_ASYNC` is defaulted off in the configuration due to #4228

- [ ] I'll test it. We are using `KOKKOS_ENABLE_IMPL_CUDA_MALLOC_ASYNC` in our builds ;)

core/src/Cuda/Kokkos_Cuda_Instance.cpp
romintomasetti: Ok, done.

core/unit_test/TestWithoutInitializing.hpp
romintomasetti: I'm not sure to understand what you mean...

core/src/Cuda/Kokkos_CudaSpace.cpp
romintomasetti: BTW, can you explain how the actions will test if I broke something on Cuda? I only see `OpenMP` and `HPX` in the checks for now...

masterleinad:
OK to test

core/src/Cuda/Kokkos_CudaSpace.cpp
masterleinad: Click on the `jenkin-ci`'s details. It looks like you didn't run clang-format-8: https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/Kokkos/detail/Kokkos/12572/pipeline.

tcclevenger: "Open Blue Ocean" tab in jenkin-ci will show all the configs once clang format passes.

romintomasetti:
@masterleinad It seems only a `gcc` job fails, see https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/Kokkos/detail/Kokkos/12596/pipeline/57. The job shows `Child killed` for `KokkosContainers_UnitTest_OpenMP` Any clue?

masterleinad:
Trying this for myself, it seems you need/want something like
```diff
diff --git a/containers/unit_tests/TestWithoutInitializing.hpp b/containers/unit_tests/TestWithoutInitializing.hpp
index b2257051e..10e1a0a4f 100644
--- a/containers/unit_tests/TestWithoutInitializing.hpp
+++ b/containers/unit_tests/TestWithoutInitializing.hpp
@@ -143,8 +143,6 @@ TEST(TEST_CATEGORY, resize_exec_space_dualview) {
 }
 
 TEST(TEST_CATEGORY, realloc_exec_space_dualview) {
-  GTEST_SKIP_IF_CUDAUVM_MEMORY_SPACE
-
   using namespace Kokkos::Test::Tools;
   listen_tool_events(Config::DisableAll(), Config::EnableFences());
   using view_type = Kokkos::DualView<int*, TEST_EXECSPACE>;
diff --git a/core/src/Cuda/Kokkos_CudaSpace.cpp b/core/src/Cuda/Kokkos_CudaSpace.cpp
index a2fb11b56..c50e6de79 100644
--- a/core/src/Cuda/Kokkos_CudaSpace.cpp
+++ b/core/src/Cuda/Kokkos_CudaSpace.cpp
@@ -175,7 +175,7 @@ void *impl_allocate_common(const Cuda &exec_space, const char *arg_label,
       exec_space.fence("Kokkos::Cuda: backend fence after async malloc");
     } else {
       error_code = cudaMallocAsync(&ptr, arg_alloc_size, 0);
-      cuda_device_synchronize("Kokkos::Cuda: backend fence after async malloc");
+      Impl::cuda_device_synchronize("Kokkos::Cuda: backend fence after async malloc");
     }
   } else {
     error_code = cudaMalloc(&ptr, arg_alloc_size);
@@ -324,9 +324,9 @@ void CudaSpace::impl_deallocate(
 #error CUDART_VERSION undefined!
 #elif (defined(KOKKOS_ENABLE_IMPL_CUDA_MALLOC_ASYNC) && CUDART_VERSION >= 11020)
     if (arg_alloc_size >= memory_threshold_g) {
-      cuda_device_synchronize("Kokkos::Cuda: backend fence before async free");
+      Impl::cuda_device_synchronize("Kokkos::Cuda: backend fence before async free");
       KOKKOS_IMPL_CUDA_SAFE_CALL(cudaFreeAsync(arg_alloc_ptr, 0));
-      cuda_device_synchronize("Kokkos::Cuda: backend fence after async free");
+      Impl::cuda_device_synchronize("Kokkos::Cuda: backend fence after async free");
     } else {
       KOKKOS_IMPL_CUDA_SAFE_CALL(cudaFree(arg_alloc_ptr));
     }
```

masterleinad:
The indentation is off. Otherwise, this looks pretty good to me.

core/src/Cuda/Kokkos_CudaSpace.cpp
romintomasetti: I guess this thread can be closed. I tackled this in this PR :)

core/src/Cuda/Kokkos_CudaSpace.cpp
romintomasetti: @PhilMiller If you agree, I'll close this thread and keep the changes.

masterleinad:
Retest this please.

masterleinad (approved):

romintomasetti:
@masterleinad I guess the reason for failure of `GCC-8.4.0` is the same as before. Do I need to worry about the failure of `SYCL-OneAPI`?

masterleinad:
> @masterleinad I guess the reason for failure of `GCC-8.4.0` is the same as before. Do I need to worry about the failure of `SYCL-OneAPI`?

This doesn't really touch `SYCL` and we have seen the failures elsewhere. I wouldn't worry about it (but we might decide to merge a fix first and rerun CI just to make sure).

romintomasetti:
@masterleinad The last action showed only `GCC-8.4.0` failing. Should we proceed with the PR?

masterleinad:
> @masterleinad The last action showed only `GCC-8.4.0` failing. Should we proceed with the PR?

I already approved it. So you need to bug other people (possibly someone with the power to merge it). 🙂

romintomasetti:
Indeed, sorry @masterleinad ! I guess I need to bug @dalg24 and @PhilMiller :incoming_envelope:

maartenarnst:
Hi @dalg24 and @PhilMiller, would you have a moment to look at this PR, and, if you agree with the changes, approve? Many thanks in advance!

containers/src/Kokkos_DualView.hpp
```diff
@@ -239,7 +239,8 @@ class DualView : public ViewTraits<DataType, Arg1Type, Arg2Type, Arg3Type> {
       : modified_flags(t_modified_flags("DualView::modified_flags")),
         d_view(arg_prop, n0, n1, n2, n3, n4, n5, n6, n7) {
     // without UVM, host View mirrors
-    if (Kokkos::Impl::has_type<Impl::WithoutInitializing_t, P...>::value)
+    if constexpr (Kokkos::Impl::has_type<Impl::WithoutInitializing_t,
+                                         P...>::value)
```
dalg24, line 243: Fine this time but in the future avoid unrelated changes

containers/unit_tests/TestWithoutInitializing.hpp
```diff
@@ -24,6 +24,24 @@
...
+std::string fence_event_message_for_zeromemset(const execution_space& space) {
+  return std::string(space.name()).append(": ZeroMemset for");
+}
+
+//! @name Some tests are skipped for @c CudaUVM memory space.
+///@{
+#ifdef KOKKOS_ENABLE_CUDA
+#define GTEST_SKIP_IF_CUDAUVM_MEMORY_SPACE                            \
+  if constexpr (std::is_same_v<typename TEST_EXECSPACE::memory_space, \
+                               Kokkos::CudaUVMSpace>)                 \
```
dalg24, line 38: Thinking out lout: We recently deprecated the Kokkos_ENABLE_CUDA_UVM option so eventually that code path will not be reachable any more.  If we really care about guaranteeing we support it, we should probably instantiate based on memory spaces rather than relying on the default memory spaces associated to a specific execution space.

core/src/impl/Kokkos_HostSpace_deepcopy.hpp
```diff
@@ -23,7 +23,7 @@ namespace Kokkos {
 
 namespace Impl {
 
-void hostspace_fence(const DefaultHostExecutionSpace& exec);
+void hostspace_fence(const DefaultHostExecutionSpace& exec, const char* msg);
```
dalg24, line 26: Are these changes required for this PR?

containers/src/Kokkos_DualView.hpp
romintomasetti: Ok, taking note of that :spiral_notepad:

core/src/impl/Kokkos_HostSpace_deepcopy.hpp
romintomasetti: This change is somewhat related to the philosophy of this PR (enhance when fencing is done, and how it is done, i.e. messages).

containers/unit_tests/TestWithoutInitializing.hpp
romintomasetti: @dalg24 I guess this is out of scope for this PR?

core/src/impl/Kokkos_HostSpace_deepcopy.hpp
dalg24: "Somewhat related" is an issue here.  I am fine with the rest of the PR but not sure yet I agree with this change.

dalg24:
Make sure you update the title line and the description of this PR to reflect the latest version.

containers/unit_tests/TestWithoutInitializing.hpp
```diff
@@ -24,6 +24,24 @@
...
+//! Get fence event message for @c ZeroMemset.
+template <typename execution_space>
+std::string fence_event_message_for_zeromemset(const execution_space& space) {
+  return std::string(space.name()).append(": ZeroMemset for");
+}
+
+//! @name Some tests are skipped for @c CudaUVM memory space.
+///@{
+#ifdef KOKKOS_ENABLE_CUDA
+#define GTEST_SKIP_IF_CUDAUVM_MEMORY_SPACE                            \
```
PhilMiller, line 36: Other discussion aside, standard practice when defining a macro like this is to wrap the contents in `do { ... } while (false)` so that the use sites have to treat it like a statement with a trailing `;`

core/src/impl/Kokkos_HostSpace_deepcopy.hpp
romintomasetti: I reverted the changes. But I think it would be good to create an issue somewhere to track what I wanted to improve (that basically calling `Kokkos::Impl::hostspace_fence` will just give to the profiling tools a message "Host fence", which is not very helpful).

containers/unit_tests/TestWithoutInitializing.hpp
romintomasetti: It is now
```c++
#ifdef KOKKOS_ENABLE_CUDA
#define GTEST_SKIP_IF_CUDAUVM_MEMORY_SPACE                            \
  if constexpr (std::is_same_v<typename TEST_EXECSPACE::memory_space, \
                               Kokkos::CudaUVMSpace>)                 \
    GTEST_SKIP() << "skipping since CudaUVMSpace requires additional fences"
#else
#define GTEST_SKIP_IF_CUDAUVM_MEMORY_SPACE do {} while(false)
#endif

File: core/src/impl/Kokkos_HostSpace_deepcopy.hpp (hunk 12 above)
Comment (romintomasetti) [reply core/src/impl/Kokkos_HostSpace_deepcopy.hpp]: @dalg24 I reverted the changes, but then I also need to modify
```c++
//! Get fence event message for @c ZeroMemset.
template <typename execution_space>
std::string fence_event_message_for_zeromemset(const execution_space& space) {
  return std::string(space.name()).append(": ZeroMemset for");
}
```
back to
```c++
//! Get fence event message for @c ZeroMemset.
template <typename execution_space>
std::string fence_event_message_for_zeromemset(const execution_space& space) {
  return std::string("HostSpace fence");
}
```

containers/unit_tests/TestWithoutInitializing.hpp
dalg24: I don't agree with Phil's suggestion.  This only apply to macro that take arguments.
I would prefer to revert but I am not blocking on this.

romintomasetti: I also preferred how it was before. I'll revert.

core/unit_test/TestWithoutInitializing.hpp
romintomasetti: I'll close this thread because I reverted the changes :no_entry:

containers/src/Kokkos_DynRankView.hpp
romintomasetti: I'm closing this thread as well, because this was discussed below.

containers/unit_tests/TestWithoutInitializing.hpp
romintomasetti: @dalg24 I think this is the last thread. If you agree with all the changes, this PR is good to go :rocket:

core/src/Cuda/Kokkos_CudaSpace.cpp
```diff
@@ -172,10 +172,11 @@ void *impl_allocate_common(const Cuda &exec_space, const char *arg_label,
...
     } else {
       error_code = cudaMallocAsync(&ptr, arg_alloc_size, 0);
-      KOKKOS_IMPL_CUDA_SAFE_CALL(cudaDeviceSynchronize());
+      Impl::cuda_device_synchronize(
+          "Kokkos::Cuda: backend fence after async malloc");
```
romintomasetti, line 179: @masterleinad This is just a comment about the `clang-format`. Do you think it would make sense (not in this PR of course) to lower the penalty of excess characters? (See https://clang.llvm.org/docs/ClangFormatStyleOptions.html#penaltyexcesscharacter)

I mean, sometimes a line with 87 characters might be nicer than the same line split weirdly across 2 lines :smile: This is just a question, not suggesting to do anything but gathering your opinion on this :smile:

dalg24: We need to update the clang-format version we are using to something more recent in the near future.
This will trigger changes throughout the code base.  When we do so would be a time where we would consider changing the clang-format configuration.  If that's something you are interested in you are welcome to explore options but this needs to be done in an issue elsewhere.
A word of warning, this is a time sink and at times can get controversial.

containers/unit_tests/TestWithoutInitializing.hpp
```diff
@@ -144,7 +162,8 @@ TEST(TEST_CATEGORY, realloc_exec_space_dualview) {
       [&](BeginFenceEvent event) {
         if ((event.descriptor().find("Debug Only Check for Execution Error") !=
              std::string::npos) ||
-            (event.descriptor().find("HostSpace fence") != std::string::npos))
+            (event.descriptor().find(fence_event_message_for_zeromemset(
+                 TEST_EXECSPACE{})) != std::string::npos))
```
dalg24, line 166: This change is unnecessary.  Please undo.

containers/unit_tests/TestWithoutInitializing.hpp
```diff
@@ -24,6 +24,28 @@
 
 #include <../../core/unit_test/tools/include/ToolTestingUtilities.hpp>
 
+//! Get fence event message for @c ZeroMemset.
+template <typename execution_space>
+std::string fence_event_message_for_zeromemset(const execution_space&) {
+  return "HostSpace fence";
+}
```
dalg24, line 31:
```suggestion
```

containers/unit_tests/TestWithoutInitializing.hpp
```diff
@@ -24,6 +24,28 @@
...
+
+/**
+ * @name Some tests are skipped for @c CudaUVM memory space.
+ *
+ * @todo To be revised according to the future of @c KOKKOS_ENABLE_CUDA_UVM .
+ */
+///@{
```
dalg24, line 38: Not requiring changes but these are 5 lines of comment (not even counting the closing comment) when one single line would do and we do not really use Doxygen so chances are it would not even generate correct documentation after other edit the file.

dalg24:
(almost there)

core/src/Cuda/Kokkos_CudaSpace.cpp
romintomasetti: * If I'm not mistaken, the [PenaltyExcessCharacter](https://clang.llvm.org/docs/ClangFormatStyleOptions.html#penaltyexcesscharacter) is available from `clang-format` 3.7, and you're requiring version `8.0` in `scripts/apply-clang-format`. So I guess the option could be readily used.
* Instead of modifying the whole code base when the clang format changes (either version or options), I think you should use `git-clang-format` to check for style only in modified files. This would avoid committing purely stylistic changes. That would avoid messing up the git history.

containers/unit_tests/TestWithoutInitializing.hpp
romintomasetti: reverted

containers/unit_tests/TestWithoutInitializing.hpp
romintomasetti: I've used `///` instead of `/** ... */`. Tell me if it's good for you ;)

core/src/Cuda/Kokkos_CudaSpace.cpp
dalg24: We allow disabling clang-format when appropriate.  You will find all sort of examples throughout the code base such as https://github.com/kokkos/kokkos/blob/4b6d971dce856132961b1f178dcae1256b92b733/core/src/impl/Kokkos_QuadPrecisionMath.hpp#L46-L67 or https://github.com/kokkos/kokkos/blob/4b6d971dce856132961b1f178dcae1256b92b733/core/src/OpenACC/Kokkos_OpenACC_ParallelFor_Range.hpp#L35-L37

In principle you are right but in practice I think it would mean that the entire file would be reformatted as soon as someone touches it and this "spreads" the noise across the history of the repo instead of having one clean "break".

PR #6015, opened 2023-03-27
Fixup 4.0 change log

dalg24:
Need to cherry-pick into 4.0.01 patch release

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -74,6 +74,7 @@
 - Make ExecutionSpace::concurrency() a non-static member function [\#5655](https://github.com/kokkos/kokkos/pull/5655) and related PRs
 
 ### Deprecations
+- Remove code guarded by `KOKKOS_ENABLE_DEPRECATED_CODE_3`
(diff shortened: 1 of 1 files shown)


crtrott:
Shouldn't most of these be there as "now removed"?

dalg24:
> Shouldn't most of these be there as "now removed"?

If we listed the "now removed" then we'd need to expand the list quite a bit.  Also for most of these we are actually being gentle and buying users a bit more time.

CHANGELOG.md
```diff
@@ -74,6 +74,7 @@
 - Make ExecutionSpace::concurrency() a non-static member function [\#5655](https://github.com/kokkos/kokkos/pull/5655) and related PRs
 
 ### Deprecations
+- Remove code guarded by `KOKKOS_ENABLE_DEPRECATED_CODE_3`
```
masterleinad, line 77: I would rather list this under "Incompatibilities".

dalg24: done

masterleinad (approved):

PR #5948, opened 2023-03-02
Introduce `KOKKOS_IMPL_ARCH_NVIDIA_GPU` macro

dalg24:
We already have `KOKKOS_ARCH_INTEL_GPU`.  It is useful for backends like SYCL, OpenMPTarget, or OpenACC, that may target GPUs from different vendors.
I am not sure whether we want it to be a public or a private macro.  We could consider defining it to be the compute capability as an integer but that means more complicate logic in the generated makefiles and in CMake.

Was proposed at last developer meeting when discussing #5867

PR #6022, opened 2023-03-29
[4.0.01] Support NVIDIA Ada architecture

crtrott:


cmake/kokkos_arch.cmake
```diff
@@ -86,6 +86,7 @@ KOKKOS_ARCH_OPTION(VOLTA72         GPU  "NVIDIA Volta generation CC 7.2"   "KOKK
 KOKKOS_ARCH_OPTION(TURING75        GPU  "NVIDIA Turing generation CC 7.5"  "KOKKOS_SHOW_CUDA_ARCHS")
 KOKKOS_ARCH_OPTION(AMPERE80        GPU  "NVIDIA Ampere generation CC 8.0"  "KOKKOS_SHOW_CUDA_ARCHS")
 KOKKOS_ARCH_OPTION(AMPERE86        GPU  "NVIDIA Ampere generation CC 8.6"  "KOKKOS_SHOW_CUDA_ARCHS")
+KOKKOS_ARCH_OPTION(ADA89           GPU  "NVIDIA Ada generation CC 8.9"     "KOKKOS_SHOW_CUDA_ARCHS")
 KOKKOS_ARCH_OPTION(HOPPER90        GPU  "NVIDIA Hopper generation CC 9.0"  "KOKKOS_SHOW_CUDA_ARCHS")
```
fnrizzi, line 91: can we put a comment on line 74 with the link to this table https://en.wikipedia.org/wiki/CUDA#GPUs_supported
so that we have that reference handy here any time we need to verify or have a global view of all the compcap and micro archs? 

(maybe a similar one for AMD too but since this is PR is for cuda we can leave it to tthat for now)

fnrizzi (approved):
left one comment, but not a major deal if skipped

masterleinad:
I think we should also cherry-pick https://github.com/kokkos/kokkos/pull/5948 if we want to cherry-pick this one (or adapt the checks for NVIDIA architectures differently).

crtrott:
The only thing which will not work without the other PR is bhalf. I think this is ok for the patch release. I am hesitant to do the larger rework here.

crtrott:
The comment should go into develop, this is patch release only.

PR #6016, opened 2023-03-28
Use -Wextra in CI

masterleinad:
We should be able to also enable `-Wextra` in our CI avoiding some complaints by users that enable this warning flag in their builds and don't treat `Kokkos` as a system library. This pull request demonstrates that are not that many changes required.

PR #6026, opened 2023-03-30
Add -Wdeprecated-copy warning and fix OMPT scan bug related to assignment operators

masterleinad:
Part of #6016 (adding `-Wextra` to the warning flags we use in the CI). 
We can't easily add `-Wdeprecated-copy` to the warning flags since it depends very much on the compiler and the compiler version if it's supported or not.

### DIFF
--- a/algorithms/src/std_algorithms/impl/Kokkos_PartitionCopy.hpp
+++ b/algorithms/src/std_algorithms/impl/Kokkos_PartitionCopy.hpp
@@ -31,25 +31,6 @@ template <class ValueType>
 struct StdPartitionCopyScalar {
   ValueType true_count_;
   ValueType false_count_;
-
-  // Here we implement the copy assignment operators explicitly for consistency
-  // with how the Scalar structs are implemented inside
-  // Kokkos_Parallel_Reduce.hpp.
-  KOKKOS_FUNCTION
-  void operator=(const StdPartitionCopyScalar& other) {
-    true_count_  = other.true_count_;
-    false_count_ = other.false_count_;
-  }
-
-  // this is needed for
-  // OpenMPTarget/Kokkos_OpenMPTarget_Parallel.hpp:699:21: error: no viable
-  // overloaded '=' m_returnvalue = 0;
-  //
-  KOKKOS_FUNCTION
-  void operator=(const ValueType value) {
-    true_count_  = value;
-    false_count_ = value;
-  }
 };
 
 template <class IndexType, class FirstFrom, class FirstDestTrue,
--- a/algorithms/src/std_algorithms/impl/Kokkos_ValueWrapperForNoNeutralElement.hpp
+++ b/algorithms/src/std_algorithms/impl/Kokkos_ValueWrapperForNoNeutralElement.hpp
@@ -29,12 +29,6 @@ template <class Scalar>
 struct ValueWrapperForNoNeutralElement {
   Scalar val;
   bool is_initial = true;
-
-  KOKKOS_FUNCTION
-  void operator=(const ValueWrapperForNoNeutralElement& rhs) {
-    val        = rhs.val;
-    is_initial = rhs.is_initial;
-  }
 };
 
 }  // namespace Impl
--- a/core/src/Kokkos_Half.hpp
+++ b/core/src/Kokkos_Half.hpp
@@ -226,9 +226,19 @@ class alignas(FloatType) floating_point_wrapper {
 #if defined(_WIN32) && defined(KOKKOS_ENABLE_CUDA)
   KOKKOS_FUNCTION
   floating_point_wrapper(const floating_point_wrapper& rhs) : val(rhs.val) {}
+
+  KOKKOS_FUNCTION
+  floating_point_wrapper& operator=(const floating_point_wrapper& rhs) {
+    val = rhs.val;
+    return *this;
+  }
 #else
   KOKKOS_DEFAULTED_FUNCTION
   floating_point_wrapper(const floating_point_wrapper&) noexcept = default;
+
+  KOKKOS_DEFAULTED_FUNCTION
+  floating_point_wrapper& operator=(const floating_point_wrapper&) noexcept =
+      default;
 #endif
 
   KOKKOS_INLINE_FUNCTION
--- a/core/src/OpenMPTarget/Kokkos_OpenMPTarget_ParallelScan_Range.hpp
+++ b/core/src/OpenMPTarget/Kokkos_OpenMPTarget_ParallelScan_Range.hpp
@@ -241,7 +241,7 @@ class ParallelScanWithTotal<FunctorType, Kokkos::RangePolicy<Traits...>,
             base_t::m_result_ptr, chunk_values.data() + (n_chunks - 1), size);
       }
     } else if (!base_t::m_result_ptr_device_accessible) {
-      *base_t::m_result_ptr = 0;
+      base_t::m_functor_reducer.get_reducer().init(base_t::m_result_ptr);
     }
   }
 
--- a/core/unit_test/incremental/Test14_MDRangeReduce.hpp
+++ b/core/unit_test/incremental/Test14_MDRangeReduce.hpp
@@ -39,6 +39,13 @@ struct MyComplex {
   MyComplex(const MyComplex& src) : _re(src._re), _im(src._im) {}
 
   KOKKOS_INLINE_FUNCTION
+  MyComplex& operator=(const MyComplex& src) {
+    _re = src._re;
+    _im = src._im;
+    return *this;
+  }
+
+  KOKKOS_INLINE_FUNCTION
   void operator+=(const MyComplex& src) {
     _re += src._re;
     _im += src._im;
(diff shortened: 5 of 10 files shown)


core/src/OpenMPTarget/Kokkos_OpenMPTarget_ParallelScan_Range.hpp
```diff
@@ -241,7 +241,7 @@ class ParallelScanWithTotal<FunctorType, Kokkos::RangePolicy<Traits...>,
             base_t::m_result_ptr, chunk_values.data() + (n_chunks - 1), size);
       }
     } else if (!base_t::m_result_ptr_device_accessible) {
-      *base_t::m_result_ptr = 0;
+      base_t::m_functor_reducer.get_reducer().init(base_t::m_result_ptr);
```
masterleinad, line 244: This is related to the changes to `StdPartitionCopyScalar`.

algorithms/src/std_algorithms/impl/Kokkos_PartitionCopy.hpp
```diff
@@ -31,25 +31,6 @@ template <class ValueType>
...
-
-  // this is needed for
-  // OpenMPTarget/Kokkos_OpenMPTarget_Parallel.hpp:699:21: error: no viable
-  // overloaded '=' m_returnvalue = 0;
-  //
-  KOKKOS_FUNCTION
-  void operator=(const ValueType value) {
-    true_count_  = value;
-    false_count_ = value;
-  }
```
masterleinad, line 52: This is fixed directly in `core/src/OpenMPTarget/Kokkos_OpenMPTarget_ParallelScan_Range.hpp`.

algorithms/src/std_algorithms/impl/Kokkos_ValueWrapperForNoNeutralElement.hpp
```diff
@@ -29,12 +29,6 @@ template <class Scalar>
 struct ValueWrapperForNoNeutralElement {
   Scalar val;
   bool is_initial = true;
-
-  KOKKOS_FUNCTION
-  void operator=(const ValueWrapperForNoNeutralElement& rhs) {
-    val        = rhs.val;
-    is_initial = rhs.is_initial;
-  }
```
dalg24, line 37: Why is removing the right thing to do?
Not used?  Unnecessary?  Please elaborate.

algorithms/src/std_algorithms/impl/Kokkos_PartitionCopy.hpp
dalg24, line 52: Please discuss this removal

core/src/Kokkos_Half.hpp
```diff
@@ -226,9 +226,19 @@ class alignas(FloatType) floating_point_wrapper {
 #if defined(_WIN32) && defined(KOKKOS_ENABLE_CUDA)
   KOKKOS_FUNCTION
   floating_point_wrapper(const floating_point_wrapper& rhs) : val(rhs.val) {}
+
+  KOKKOS_FUNCTION
+  floating_point_wrapper& operator=(const floating_point_wrapper& rhs) {
+    val = rhs.val;
+    return *this;
+  }
```
dalg24, line 234: Please give some context for this addition.

dalg24 (commented):
Can you clarify.  Is your point that `-Wdeprecated-copy` being part of `-Wextra` depends on the compiler so we might as well add it ourselves to fix it everywhere?

PhilMiller:
> Can you clarify. Is your point that `-Wdeprecated-copy` being part of `-Wextra` depends on the compiler so we might as well add it ourselves to fix it everywhere?

I think it's the opposite - `-Wdeprecated-copy` only exists on some compilers, so we can't add it explicitly. On ones that have it, it's included in `-Wextra`, so we need to fix what it would report if we want `-Wextra` to build cleanly on those compilers

fnrizzi:
@masterleinad from the title it seemed to me that this PR was just adding decpr-copy flag, but then i saw a bunch of changes all over in different places. So I am guessing all the changes are related to the implicit copy constructor?

algorithms/src/std_algorithms/impl/Kokkos_PartitionCopy.hpp
masterleinad: The changes to `OpenMPTarget` (essentially fixing a bug) make the second assignment operator redundant and the copy assignment operator is a leftover from removing the `volatile` requirement for reductions. `-Wdeprecated-copy` of course only complains about a missing copy constructor but we don't need any of this. The compiler-generated special member functions are sufficient.

algorithms/src/std_algorithms/impl/Kokkos_ValueWrapperForNoNeutralElement.hpp
masterleinad: Again, the compiler-generated special member functions are sufficient. The alternative for fixing the warning would be to define a copy constructor.

core/src/Kokkos_Half.hpp
masterleinad: The warning requires us to define a copy assignment operator if a copy constructor is defined and since we can't make the latter one `default`, it seems reasonable also not to make the former one `default`.

masterleinad:
> I think it's the opposite - `-Wdeprecated-copy` only exists on some compilers, so we can't add it explicitly. On ones that have it, it's included in `-Wextra`, so we need to fix what it would report if we want `-Wextra` to build cleanly on those compilers

Yes, it not only depends on the compiler; the warning is also only available for newer compiler versions for `gcc` and `clang`. If the idea is to support `-Wextra` in the end, it seems not worth it to figure out the exact circumstances when the warning can be used.

masterleinad:
> @masterleinad from the title it seemed to me that this PR was just adding decpr-copy flag, but then i saw a bunch of changes all over in different places. So I am guessing all the changes are related to the copy constructor and copy assign?

Right, the warning is about consistency between copy assignment operators and copy constructors. Admittedly, the OMPT bug fix is somewhat of a drive-by change (that still fits the context IMHO).

PhilMiller (dismissed):

rgayatri23 (approved):
All my questions are answered in the conversation above. 
LGTM if all the relevant unit tests pass

core/src/Kokkos_Half.hpp
e10harvey: It looks like this was added to surpress a compiler warning about copy assignment operators. We could consider using pragma to supress that warning instead. But, as long as the half precision unit tests pass and are extended (or verified) to cover this operator, I think the change is fine. It would also be good to add a comment above these added lines explaining why they are needed.

masterleinad: > It looks like this was added to surpress a compiler warning about copy assignment operators. 

This pull request is about compatibility with a new warning flag that flags exactly set. 🙂
I don't see a good reason why we would try to suppress the warning we are adding here. Note that we could simplify all this significantly if we dropped the `volatile` overloads.

> We could consider using pragma to supress that warning instead. But, as long as the half precision unit tests pass and are extended (or verified) to cover this operator, I think the change is fine.

We get a bunch of compile errors if we tried to delete the copy assignment operator.

PhilMiller: I don't think we have any reason to keep the `volatile` overloads around any more

core/unit_test/incremental/Test14_MDRangeReduce.hpp
```diff
@@ -38,6 +38,13 @@ struct MyComplex {
   KOKKOS_INLINE_FUNCTION
   MyComplex(const MyComplex& src) : _re(src._re), _im(src._im) {}
 
+  KOKKOS_INLINE_FUNCTION
+  MyComplex& operator=(const MyComplex& src) {
+    _re = src._re;
+    _im = src._im;
+    return *this;
+  }
```
dalg24, line 46: What compiler emitted a warning for this?
Same as for `floating_point_wrapper` and `long_wrappper`, I don't fully get why this is related to `-Wdeprecated-copy`
https://godbolt.org/z/av9hbYjah

dalg24 (commented):
Looks mostly fine.  Need clarification why the addition of copy assignment operators belongs to this PR.

core/unit_test/incremental/Test14_MDRangeReduce.hpp
masterleinad: https://godbolt.org/z/beqo1rqoG shows
```
<source>:11:12: error: implicitly-declared 'constexpr Foo& Foo::operator=(const Foo&)' is deprecated [-Werror=deprecated-copy]
   11 |     foo1 = foo2;
      |            ^~~~
<source>:3:5: note: because 'Foo' has user-provided 'Foo::Foo(const Foo&)'
    3 |     Foo(Foo const&) {}
      |     ^~~
```
for `gcc` and
```
<source>:3:5: error: definition of implicit copy assignment operator for 'Foo' is deprecated because it has a user-provided copy constructor [-Werror,-Wdeprecated-copy-with-user-provided-copy]
    Foo(Foo const&) {}
    ^
<source>:11:10: note: in implicit copy assignment operator for 'Foo' first required here
    foo1 = foo2;
         ^
```
for `clang`.

dalg24 (approved):

dalg24:
Ignoring
```
The following tests FAILED:
	 24 - KokkosContainers_UnitTest_OpenMP (Child killed)
```
in GCC-8.4.0 build

core/unit_test/incremental/Test14_MDRangeReduce.hpp
dalg24: Very good.  Thanks.

PR #6029, opened 2023-03-31
Relax scratch space limits for HIP reductions

arghdos:
This change is motivated by profiling of some LJ reduction kernels in LAMMPS.  Essentially we noticed that in some cases, we ended up severely occupancy limited by the "gridDim <= blockDim" restriction.  Omniperf showed very low L1/L2 utilization, but the limiter was simply that we were not launching enough blocks.

This is a simple change to relax the scratch space limit slightly (from blockDim to 4096) to increase occupancy, but still keep the scratch limits well bounded.  I was not particularly scientific about choosing the new limit, and I'm definitely open to better heuristics, but this increased [RHODO](https://github.com/lammps/lammps/blob/develop/bench/in.rhodo.scaled) (4^3) perf in LAMMPS by ~28% (the key kernel is a ParReduce)

### DIFF
--- a/core/src/HIP/Kokkos_HIP_Parallel_Range.hpp
+++ b/core/src/HIP/Kokkos_HIP_Parallel_Range.hpp
@@ -296,10 +296,11 @@ class ParallelReduce<CombinedFunctorReducerType, Kokkos::RangePolicy<Traits...>,
 
       // REQUIRED ( 1 , N , 1 )
       dim3 block(1, block_size, 1);
-      // use a slightly less constrained, but still well bounded limit for scratch
+      // use a slightly less constrained, but still well bounded limit for
+      // scratch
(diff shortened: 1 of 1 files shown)


arghdos:
cc: @Rombur / @dalg24  -- the one we discussed on slack

Rombur (approved):

stanmoore1:
@arghdos nice catch!

dalg24:
I re-ran exercise 4 from the tutorials.  It is looking better with this PR.
arghdos_relax_scratch_limits

arghdos:
That was next on my list of things to check :), I'll re-run my profiling on the new (smaller) dip and see if we can iron it out further

dalg24 (approved):

dalg24:
Ignoring clearly unrelated GCC-8.4.0 failure

PR #6030, opened 2023-04-03
Fix -Wmissing-field-initializers warning

masterleinad:
Part of #6016. The uninitialized member is "error_message", see https://github.com/kokkos/kokkos/blob/5e574380cbca98ec45d92f62fe861d38aab0d825/core/src/impl/Kokkos_Profiling.hpp#L49-L58


dalg24 (approved):

masterleinad:
`gcc-8.4.0` is failing with
```
24 - KokkosContainers_UnitTest_OpenMP (Child killed)
```
which looks unrelated.

cz4rs (approved):

PR #6038, opened 2023-04-06
KOKKOS_COMPILER_PGI -> KOKKOS_COMPILER_NVHPC

masterleinad:
While working on #5906, I realized that we never define KOKKOS_COMPILER_PGI but still use the macro in some places where it seems appropriate to use KOKKOS_COMPILER_NVHPC instead. Also, it seems that we don't need `_kokkos_pgi_compiler_bug_workaround` for NVHPC.

### DIFF
--- a/core/src/Kokkos_Macros.hpp
+++ b/core/src/Kokkos_Macros.hpp
@@ -283,7 +283,7 @@
 
 //----------------------------------------------------------------------------
 
-#if defined(KOKKOS_COMPILER_PGI)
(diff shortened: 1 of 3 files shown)


crtrott (approved):

dalg24 (approved):

dalg24:
Ignoring `24 - KokkosContainers_UnitTest_OpenMP (Child killed)` in the GCC-8.4.0 build

Issue #4247, opened 2021-08-17
Host CPU arch flags not added with nvcc


stanmoore1:
When I set `KOKKOS_ARCH=Power9,Volta70` and compile with `nvcc`, the host CPU arch flags `-mcpu=power9 -mtune=power9` are not added to the build.

PR #6042, opened 2023-04-10
Store a team handle reference in ThreadVectorRangeBoundariesStructs

ldh4:
Storing a reference to `TeamMemberType` handle in `ThreadVectorRangeBoundariesStruct` in serial, cuda and hip for consistencies with other backends' `ThreadVectorRangeBoundariesStruct` implementation.

This allows retrieval of the team member handle from `ThreadVectorRangeBoundariesStruct`.

### DIFF
--- a/core/src/Cuda/Kokkos_Cuda_Team.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_Team.hpp
@@ -369,17 +369,19 @@ struct TeamVectorRangeBoundariesStruct<iType, CudaTeamMember> {
 template <typename iType>
 struct ThreadVectorRangeBoundariesStruct<iType, CudaTeamMember> {
   using index_type = iType;
+  const CudaTeamMember& member;
   const index_type start;
   const index_type end;
 
   KOKKOS_INLINE_FUNCTION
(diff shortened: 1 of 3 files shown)


masterleinad (approved):
I assume this will be used/tested in #4247.

core/src/Cuda/Kokkos_Cuda_Team.hpp
```diff
@@ -369,17 +369,19 @@ struct TeamVectorRangeBoundariesStruct<iType, CudaTeamMember> {
 template <typename iType>
 struct ThreadVectorRangeBoundariesStruct<iType, CudaTeamMember> {
   using index_type = iType;
+  const CudaTeamMember& member;
```
dalg24, line 372: Why a const reference instead of a copy?  (Not asking you to change, trying to figure out if that what discussed as part of the design)

crtrott (changes requested):
Is there any good reason to do this? I don't think consistency is a good driver here, since this is performance critical code, and any extra instructions done, and extra registers being potentially used are harmful.

ldh4:
Closing due to valid performance concerns on this change.

PR #6043, opened 2023-04-10
OpenMPTarget: Update hierarchical parallelism.

rgayatri23:
The PR does the following: 
* Update concurrency based on underlying architecture
* Update max_active_teams based on the updated concurrency
* Introduce a different style of hierarchical parallelism for Intel architectures

Todo:
* Update concurrency specifically for AMD GPUs.

### DIFF
--- a/core/src/OpenMPTarget/Kokkos_OpenMPTarget_Exec.cpp
+++ b/core/src/OpenMPTarget/Kokkos_OpenMPTarget_Exec.cpp
@@ -110,15 +110,16 @@ void OpenMPTargetExec::resize_scratch(int64_t team_size, int64_t shmem_size_L0,
   int max_active_teams =
       std::min(OpenMPTargetExec::MAX_ACTIVE_THREADS / team_size, league_size);
 
-  // Set max_active_teams as the upper bound to the number of teams that can be
-  // generated.
-  omp_set_num_teams(max_active_teams);
+  // max_active_teams is the number of active teams on the given hardware.
+  // We set the number of teams to be twice the number of max_active_teams for
+  // the compiler to pick the right number in its case.
+  omp_set_num_teams(max_active_teams * 2);
 
   // Total amount of scratch memory allocated is depenedent
   // on the maximum number of in-flight teams possible.
   int64_t total_size =
       (shmem_size + OpenMPTargetExecTeamMember::TEAM_REDUCE_SIZE + padding) *
-      max_active_teams;
+      max_active_teams * 2;
 
   if (total_size > m_scratch_size) {
     space.deallocate(m_scratch_ptr, m_scratch_size);
--- a/core/src/OpenMPTarget/Kokkos_OpenMPTarget_Exec.cpp
+++ b/core/src/OpenMPTarget/Kokkos_OpenMPTarget_Exec.cpp
@@ -74,6 +74,8 @@ int64_t OpenMPTargetExec::m_scratch_size      = 0;
 int* OpenMPTargetExec::m_lock_array           = nullptr;
 int64_t OpenMPTargetExec::m_lock_size         = 0;
 uint32_t* OpenMPTargetExec::m_uniquetoken_ptr = nullptr;
+int OpenMPTargetExec::MAX_ACTIVE_THREADS =
+    Kokkos::Experimental::OpenMPTarget().concurrency();
 
 void OpenMPTargetExec::clear_scratch() {
   Kokkos::Experimental::OpenMPTargetSpace space;
@@ -100,11 +102,22 @@ void OpenMPTargetExec::resize_scratch(int64_t team_size, int64_t shmem_size_L0,
   const int64_t shmem_size =
       shmem_size_L0 + shmem_size_L1;  // L0 + L1 scratch memory per team.
   const int64_t padding = shmem_size * 10 / 100;  // Padding per team.
+
+  // Maximum active teams possible.
+  // The number should not exceed the maximum in-flight teams possible or the
+  // league_size.
+  int max_active_teams =
+      std::min(OpenMPTargetExec::MAX_ACTIVE_THREADS / team_size, league_size);
+
+  // Set max_active_teams as the upper bound to the number of teams that can be
+  // generated.
+  omp_set_num_teams(max_active_teams);
+
   // Total amount of scratch memory allocated is depenedent
   // on the maximum number of in-flight teams possible.
   int64_t total_size =
       (shmem_size + OpenMPTargetExecTeamMember::TEAM_REDUCE_SIZE + padding) *
-      std::min(MAX_ACTIVE_THREADS / team_size, league_size);
+      max_active_teams;
 
   if (total_size > m_scratch_size) {
     space.deallocate(m_scratch_ptr, m_scratch_size);
--- a/core/src/OpenMPTarget/Kokkos_OpenMPTarget_Exec.cpp
+++ b/core/src/OpenMPTarget/Kokkos_OpenMPTarget_Exec.cpp
@@ -67,6 +67,8 @@ void OpenMPTargetExec::verify_initialized(const char* const label) {
     msg.append(" ERROR: not initialized");
     Kokkos::Impl::throw_runtime_exception(msg);
   }
+  OpenMPTargetExec::MAX_ACTIVE_THREADS =
+      Kokkos::Experimental::OpenMPTarget().concurrency();
 }
 
 void* OpenMPTargetExec::m_scratch_ptr         = nullptr;
@@ -74,8 +76,7 @@ int64_t OpenMPTargetExec::m_scratch_size      = 0;
 int* OpenMPTargetExec::m_lock_array           = nullptr;
 int64_t OpenMPTargetExec::m_lock_size         = 0;
 uint32_t* OpenMPTargetExec::m_uniquetoken_ptr = nullptr;
-int OpenMPTargetExec::MAX_ACTIVE_THREADS =
-    Kokkos::Experimental::OpenMPTarget().concurrency();
+int OpenMPTargetExec::MAX_ACTIVE_THREADS      = 0;
 
 void OpenMPTargetExec::clear_scratch() {
   Kokkos::Experimental::OpenMPTargetSpace space;
--- a/core/src/OpenMPTarget/Kokkos_OpenMPTarget_Instance.cpp
+++ b/core/src/OpenMPTarget/Kokkos_OpenMPTarget_Instance.cpp
@@ -96,7 +96,7 @@ void OpenMPTargetInternal::print_configuration(std::ostream& os,
                                                bool /*verbose*/) const {
   // FIXME_OPENMPTARGET
   os << "Using OpenMPTarget\n";
-#if defined(KOKKOS_OPENEMPTARGET_HIERARCHICAL_INTEL_GPU)
+#if defined(KOKKOS_IMPL_OPENMPTARGET_HIERARCHICAL_INTEL_GPU)
   os << "Defined KOKKOS_OPENEMPTARGET_HIERARCHICAL_INTEL_GPU: Workaround for "
         "Hierarchical parallelism for Intel GPUs."
 #endif
--- a/core/src/OpenMPTarget/Kokkos_OpenMPTarget_Instance.cpp
+++ b/core/src/OpenMPTarget/Kokkos_OpenMPTarget_Instance.cpp
@@ -97,7 +97,8 @@ void OpenMPTargetInternal::print_configuration(std::ostream& os,
   // FIXME_OPENMPTARGET
   os << "Using OpenMPTarget\n";
 #if defined(KOKKOS_IMPL_OPENMPTARGET_HIERARCHICAL_INTEL_GPU)
-  os << "Defined KOKKOS_IMPL_OPENMPTARGET_HIERARCHICAL_INTEL_GPU: Workaround for "
+  os << "Defined KOKKOS_IMPL_OPENMPTARGET_HIERARCHICAL_INTEL_GPU: Workaround "
+        "for "
         "hierarchical parallelism for Intel GPUs.";
 #endif
 }
--- a/core/src/OpenMPTarget/Kokkos_OpenMPTarget_Instance.cpp
+++ b/core/src/OpenMPTarget/Kokkos_OpenMPTarget_Instance.cpp
@@ -31,6 +31,7 @@
 #include <OpenMPTarget/Kokkos_OpenMPTarget_UniqueToken.hpp>
 #include <OpenMPTarget/Kokkos_OpenMPTarget_Instance.hpp>
 #include <impl/Kokkos_ExecSpaceManager.hpp>
+#include <OpenMPTarget/Kokkos_OpenMPTarget_Parallel.hpp>
 
 #include <sstream>
 
@@ -95,6 +96,10 @@ void OpenMPTargetInternal::print_configuration(std::ostream& os,
                                                bool /*verbose*/) const {
   // FIXME_OPENMPTARGET
   os << "Using OpenMPTarget\n";
+#if defined(KOKKOS_OPENEMPTARGET_HIERARCHICAL_INTEL_GPU)
+  os << "Defined KOKKOS_OPENEMPTARGET_HIERARCHICAL_INTEL_GPU: Workaround for "
+        "Hierarchical parallelism for Intel GPUs."
+#endif
 }
 
 void OpenMPTargetInternal::impl_finalize() {
--- a/core/src/OpenMPTarget/Kokkos_OpenMPTarget_Instance.cpp
+++ b/core/src/OpenMPTarget/Kokkos_OpenMPTarget_Instance.cpp
@@ -84,8 +84,8 @@ int OpenMPTargetInternal::concurrency() const {
 #pragma omp target map(max_threads)
   { max_threads = omp_get_num_procs(); }
 
-  // Multiply the number of processors with teh SIMD length.
-  max_threads *= 64;
+  // Multiply the number of processors with the SIMD length.
+  max_threads *= 32;
 #endif
 
   return max_threads;
--- a/core/src/OpenMPTarget/Kokkos_OpenMPTarget_Instance.cpp
+++ b/core/src/OpenMPTarget/Kokkos_OpenMPTarget_Instance.cpp
@@ -66,7 +66,29 @@ void OpenMPTargetInternal::fence(const std::string& name,
   }
 }
 int OpenMPTargetInternal::concurrency() const {
-  return 128000;  // FIXME_OPENMPTARGET
+  int max_threads = 2048 * 80;
+#if defined(KOKKOS_IMPL_ARCH_NVIDIA_GPU)
+  int max_threads_block = 2048;
+#if defined(KOKKOS_ARCH_AMPERE86)
+  max_threads = max_threads_block * 1536;
+#elif defined(KOKKOS_ARCH_AMPERE80)
+  max_threads = max_threads_block * 108;
+#elif defined(KOKKOS_ARCH_VOLTA72)
+  max_threads = max_threads_block * 84;
+#elif defined(KOKKOS_ARCH_VOLTA70)
+  max_threads = max_threads_block * 80;
+#elif defined(KOKKOS_ARCH_PASCAL60) || defined(KOKKOS_ARCH_PASCAL61)
+  max_threads = max_threads_block * 60;
+#endif
+#elif defined(KOKKOS_ARCH_INTEL_GPU)
+#pragma omp target map(max_threads)
+  { max_threads = omp_get_num_procs(); }
+
+  // Multiply the number of processors with teh SIMD length.
+  max_threads *= 64;
+#endif
+
+  return max_threads;
 }
 const char* OpenMPTargetInternal::name() { return "OpenMPTarget"; }
 void OpenMPTargetInternal::print_configuration(std::ostream& os,
--- a/core/src/OpenMPTarget/Kokkos_OpenMPTarget_Instance.cpp
+++ b/core/src/OpenMPTarget/Kokkos_OpenMPTarget_Instance.cpp
@@ -69,17 +69,17 @@ void OpenMPTargetInternal::fence(const std::string& name,
 int OpenMPTargetInternal::concurrency() const {
   int max_threads = 2048 * 80;
 #if defined(KOKKOS_IMPL_ARCH_NVIDIA_GPU)
-  int max_threads_block = 2048;
+  int max_threads_sm = 2048;
 #if defined(KOKKOS_ARCH_AMPERE86)
-  max_threads = max_threads_block * 1536;
+  max_threads = max_threads_sm * 84;
 #elif defined(KOKKOS_ARCH_AMPERE80)
-  max_threads = max_threads_block * 108;
+  max_threads = max_threads_sm * 108;
 #elif defined(KOKKOS_ARCH_VOLTA72)
-  max_threads = max_threads_block * 84;
+  max_threads = max_threads_sm * 84;
 #elif defined(KOKKOS_ARCH_VOLTA70)
-  max_threads = max_threads_block * 80;
+  max_threads = max_threads_sm * 80;
 #elif defined(KOKKOS_ARCH_PASCAL60) || defined(KOKKOS_ARCH_PASCAL61)
-  max_threads = max_threads_block * 60;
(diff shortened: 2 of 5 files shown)


core/src/OpenMPTarget/Kokkos_OpenMPTarget_Instance.cpp
```diff
@@ -66,7 +66,29 @@ void OpenMPTargetInternal::fence(const std::string& name,
...
+#elif defined(KOKKOS_ARCH_VOLTA72)
+  max_threads = max_threads_block * 84;
+#elif defined(KOKKOS_ARCH_VOLTA70)
+  max_threads = max_threads_block * 80;
+#elif defined(KOKKOS_ARCH_PASCAL60) || defined(KOKKOS_ARCH_PASCAL61)
+  max_threads = max_threads_block * 60;
+#endif
+#elif defined(KOKKOS_ARCH_INTEL_GPU)
+#pragma omp target map(max_threads)
+  { max_threads = omp_get_num_procs(); }
```
masterleinad, line 85: Does this give you the number of EUs? If so, shouldn't the value be a little higher since multiple workgroups could be scheduled by a EU? See https://github.com/intel/llvm/blob/756ba2616111235bba073e481b7f1c8004b34ee6/sycl/source/detail/reduction.cpp#L51-L62.

core/src/OpenMPTarget/Kokkos_OpenMPTarget_ParallelFor_Team.hpp
```diff
@@ -115,44 +121,68 @@ class ParallelFor<FunctorType, Kokkos::TeamPolicy<Properties...>,
...
 // teams-distribute and parallel. Gave a 2x performance boost in test cases with
 // the clang compiler. atomic_compare_exchange can be avoided since the standard
 // guarantees that the number of teams specified in the `num_teams` clause is
 // always less than or equal to the maximum concurrently running teams.
-#pragma omp target teams num_teams(nteams) thread_limit(team_size) \
-    map(to                                                         \
-        : a_functor) is_device_ptr(scratch_ptr)
+#if !defined(KOKKOS_IMPL_HIERARCHICAL_INTEL_GPU)
+#pragma omp target teams thread_limit(team_size) firstprivate(a_functor) \
+    num_teams(max_active_teams) is_device_ptr(scratch_ptr)
```
masterleinad, line 143: This looks similar to https://github.com/kokkos/kokkos/pull/6035 and thus I would expect https://github.com/kokkos/kokkos/pull/6035#issuecomment-1502028488 to also apply. Why is restricting the number of workgroups/teams here, in general, a good idea?

core/src/OpenMPTarget/Kokkos_OpenMPTarget_ParallelFor_Team.hpp
```diff
@@ -115,44 +121,68 @@ class ParallelFor<FunctorType, Kokkos::TeamPolicy<Properties...>,
...
-            m_functor(TagType(), team);
-        }
-      } else
-        Kokkos::abort("`num_teams` clause was not respected.\n");
+      for (int league_id = blockIdx; league_id < league_size;
+           league_id += gridDim) {
+        typename Policy::member_type team(league_id, league_size, team_size,
+                                          vector_length, scratch_ptr, blockIdx,
+                                          shmem_size_L0, shmem_size_L1);
+        if constexpr (std::is_void<TagType>::value)
```
masterleinad, line 160:
```suggestion
        if constexpr (std::is_void_v<TagType>)
```

core/src/OpenMPTarget/Kokkos_OpenMPTarget_ParallelFor_Team.hpp
```diff
@@ -115,44 +121,68 @@ class ParallelFor<FunctorType, Kokkos::TeamPolicy<Properties...>,
...
+    for (int i = 0; i < league_size; i++) {
+#pragma omp parallel num_threads(team_size)
+      {
+        if (omp_get_num_teams() > max_active_teams)
+          Kokkos::abort("`omp_set_num_teams` call was not respected.\n");
+
+        typename Policy::member_type team(i, league_size, team_size,
+                                          vector_length, scratch_ptr, i,
+                                          shmem_size_L0, shmem_size_L1);
+        if constexpr (std::is_void<TagType>::value)
```
masterleinad, line 179:
```suggestion
        if constexpr (std::is_void_v<TagType>)
```

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Parallel_Common.hpp
```diff
@@ -355,42 +362,61 @@ struct ParallelReduceSpecialize<FunctorType, TeamPolicyInternal<PolicyArgs...>,
...
-            f(TagType(), team, result);
-        }
-      } else
-        Kokkos::abort("`num_teams` clause was not respected.\n");
+      for (int league_id = blockIdx; league_id < league_size;
+           league_id += gridDim) {
+        typename PolicyType::member_type team(
+            league_id, league_size, team_size, vector_length, scratch_ptr,
+            blockIdx, shmem_size_L0, shmem_size_L1);
+        if constexpr (std::is_void<TagType>::value)
```
masterleinad, line 393:
```suggestion
        if constexpr (std::is_void_v<TagType>)
```

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Parallel_Common.hpp
```diff
@@ -355,42 +362,61 @@ struct ParallelReduceSpecialize<FunctorType, TeamPolicyInternal<PolicyArgs...>,
...
+    for (int i = 0; i < league_size; i++) {
+#pragma omp parallel reduction(custom : result)
+      {
+        if (omp_get_num_teams() > max_active_teams)
+          Kokkos::abort("`omp_set_num_teams` call was not respected.\n");
+
+        typename PolicyType::member_type team(i, league_size, team_size,
+                                              vector_length, scratch_ptr, i,
+                                              shmem_size_L0, shmem_size_L1);
+        if constexpr (std::is_void<TagType>::value)
```
masterleinad, line 413:
```suggestion
        if constexpr (std::is_void_v<TagType>)
```

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Parallel_Common.hpp
```diff
@@ -416,68 +442,66 @@ struct ParallelReduceSpecialize<FunctorType, TeamPolicyInternal<PolicyArgs...>,
...
-                f(TagType(), team, result);
-            }
-          } else
-            Kokkos::abort("`num_teams` clause was not respected.\n");
+          for (int league_id = blockIdx; league_id < league_size;
+               league_id += gridDim) {
+            typename PolicyType::member_type team(
+                league_id, league_size, team_size, vector_length, scratch_ptr,
+                blockIdx, shmem_size_L0, shmem_size_L1);
+            if constexpr (std::is_void<TagType>::value)
```
masterleinad, line 473:
```suggestion
            if constexpr (std::is_void_v<TagType>)
```

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Parallel_Common.hpp
```diff
@@ -416,68 +442,66 @@ struct ParallelReduceSpecialize<FunctorType, TeamPolicyInternal<PolicyArgs...>,
...
```
masterleinad, line 500:
```suggestion
            if constexpr (std::is_void_v<TagType>)
```

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Parallel_Common.hpp
```diff
@@ -487,28 +511,28 @@ struct ParallelReduceSpecialize<FunctorType, TeamPolicyInternal<PolicyArgs...>,
...
-              f(TagType(), team, result);
-          }
-        } else
-          Kokkos::abort("`num_teams` clause was not respected.\n");
+        for (int league_id = blockIdx; league_id < league_size;
+             league_id += gridDim) {
+          typename PolicyType::member_type team(
+              league_id, league_size, team_size, vector_length, scratch_ptr,
+              blockIdx, shmem_size_L0, shmem_size_L1);
+          if constexpr (std::is_void<TagType>::value)
```
masterleinad, line 531:
```suggestion
          if constexpr (std::is_void_v<TagType>)
```

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Exec.cpp
```diff
@@ -74,6 +74,8 @@ int64_t OpenMPTargetExec::m_scratch_size      = 0;
 int* OpenMPTargetExec::m_lock_array           = nullptr;
 int64_t OpenMPTargetExec::m_lock_size         = 0;
 uint32_t* OpenMPTargetExec::m_uniquetoken_ptr = nullptr;
+int OpenMPTargetExec::MAX_ACTIVE_THREADS =
+    Kokkos::Experimental::OpenMPTarget().concurrency();
```
dalg24, line 78: This is barely legal.
It violates Kokkos semantics (building an execution space before Kokkos is initialized) and just happens to work with the current OpenMPTarget implementation.

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Instance.cpp
```diff
@@ -66,7 +66,29 @@ void OpenMPTargetInternal::fence(const std::string& name,
...
+
+  // Multiply the number of processors with teh SIMD length.
+  max_threads *= 64;
```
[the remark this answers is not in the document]
rgayatri23: I assumed the SIMD length to be 64 to create a maximum number of possible "threads" on Intel architectures.

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Instance.cpp
rgayatri23: As per OpenMP spec, this should give us the number of procs on the device and each proc can then execute a SIMD instruction.

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Exec.cpp
rgayatri23: I will move this into one of the initialize routines.

core/src/OpenMPTarget/Kokkos_OpenMPTarget_ParallelFor_Team.hpp
rgayatri23: No its not a good idea to restrict the number of teams but unfortunately for the OpenMPTarget backend, we need a tight control over the number of teams generated as we have data structures that depend on the maximum number of in-flight teams.

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Exec.cpp
```diff
@@ -100,11 +103,23 @@ void OpenMPTargetExec::resize_scratch(int64_t team_size, int64_t shmem_size_L0,
...
+  // Maximum active teams possible.
+  // The number should not exceed the maximum in-flight teams possible or the
+  // league_size.
+  int max_active_teams =
+      std::min(OpenMPTargetExec::MAX_ACTIVE_THREADS / team_size, league_size);
+
+  // max_active_teams is the number of active teams on the given hardware.
+  // We set the number of teams to be twice the number of max_active_teams for
+  // the compiler to pick the right number in its case.
+  omp_set_num_teams(max_active_teams * 2);
```
dalg24, line 116: `OpenMPTargetExec::resize_scratch` seems like a weird place to set this.

core/src/OpenMPTarget/Kokkos_OpenMPTarget_ParallelFor_Team.hpp
```diff
@@ -22,6 +22,12 @@
 #include <Kokkos_Parallel.hpp>
 #include <OpenMPTarget/Kokkos_OpenMPTarget_Parallel.hpp>
 
+// Intel architectures prefer the classical hierarchical parallelism that relies
+// on OpenMP.
+#if defined(KOKKOS_ARCH_INTEL_GPU)
+#define KOKKOS_IMPL_HIERARCHICAL_INTEL_GPU
+#endif
```
dalg24, line 29: Is that something that it would make sense to define somewhere more "central" and include in print as part of the configuration?

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Instance.cpp
```diff
@@ -66,7 +66,29 @@ void OpenMPTargetInternal::fence(const std::string& name,
...
+#elif defined(KOKKOS_ARCH_PASCAL60) || defined(KOKKOS_ARCH_PASCAL61)
+  max_threads = max_threads_block * 60;
+#endif
+#elif defined(KOKKOS_ARCH_INTEL_GPU)
+#pragma omp target map(max_threads)
+  { max_threads = omp_get_num_procs(); }
+
+  // Multiply the number of processors with the SIMD length.
+  max_threads *= 32;
+#endif
```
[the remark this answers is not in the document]
rgayatri23: The `omp_get_num_proc` is only invoked once for every instance and I think that should be ok. 
The AMD currently has the default 2048*80 number of threads which is fine for now. I will add a FIXME line there to fix it in the future. I don't have a way to get the right number for AMD GPUs right now.

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Exec.cpp
rgayatri23: `resize_scratch` is called for every instance of hierarchical parallelism since we want to control the number of teams generated. And since we allocate data for max_active_teams in the `resize_scratch` routine, I set the number of teams there so that we don't need to save it somewhere or recalculate it.

core/src/OpenMPTarget/Kokkos_OpenMPTarget_ParallelFor_Team.hpp
rgayatri23: We can put that macro in a more central place. I will do that. It will be better since we need it in more than one file. 
But I don't think we need this in configuration printing since the user is not concerned (IMO) with how we implement hierarchical parallelism.

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Exec.cpp
```diff
@@ -100,11 +103,23 @@ void OpenMPTargetExec::resize_scratch(int64_t team_size, int64_t shmem_size_L0,
...
+
+  // Maximum active teams possible.
+  // The number should not exceed the maximum in-flight teams possible or the
+  // league_size.
+  int max_active_teams =
+      std::min(OpenMPTargetExec::MAX_ACTIVE_THREADS / team_size, league_size);
+
+  // max_active_teams is the number of active teams on the given hardware.
+  // We set the number of teams to be twice the number of max_active_teams for
+  // the compiler to pick the right number in its case.
```
dalg24, line 115: I do not understand this comment.  Why are we setting the number to twice the max instead of just once?
Also what do you mean "in its case"?

rgayatri23: The ideal case would be to set an large enough upper bound on the number of teams generated using `omp_set_num_teams` and and let compiler pick up the right number of teams for a give `target` region. We do that in `resize_scratch` where we assign the upper bound to be 2*max_active_teams . However that call is not respected and hence the need to add `num_teams`. 

The idea is to  not hamper compiler's ability to chose the appropriate number of teams (hence a large upper bound) but also to have control over that number so we can allocate data for each team.

core/src/OpenMPTarget/Kokkos_OpenMPTarget_ParallelFor_Team.hpp
```diff
@@ -115,44 +115,68 @@ class ParallelFor<FunctorType, Kokkos::TeamPolicy<Properties...>,
...
-#pragma omp target teams num_teams(nteams) thread_limit(team_size) \
-    map(to                                                         \
-        : a_functor) is_device_ptr(scratch_ptr)
+#if !defined(KOKKOS_IMPL_HIERARCHICAL_INTEL_GPU)
+#pragma omp target teams thread_limit(team_size) firstprivate(a_functor) \
+    num_teams(max_active_teams) is_device_ptr(scratch_ptr)
 #pragma omp parallel
     {
+      if (omp_get_num_teams() > max_active_teams)
+        Kokkos::abort("`omp_set_num_teams` call was not respected.\n");
```
dalg24, line 141: What is the point of that check?  Can this ever fail?  That would indicate a bug in the OpenMP implementation wouldn't it?

core/src/OpenMPTarget/Kokkos_OpenMPTarget_ParallelFor_Team.hpp
```diff
@@ -115,44 +115,68 @@ class ParallelFor<FunctorType, Kokkos::TeamPolicy<Properties...>,
...
+
       const int blockIdx = omp_get_team_num();
       const int gridDim  = omp_get_num_teams();
 
       // Iterate through the number of teams until league_size and assign the
       // league_id accordingly
       // Guarantee that the compilers respect the `num_teams` clause
-      if (gridDim <= nteams) {
```
dalg24, line 142: Not getting warnings about `gridDim` not being used anymore?

dalg24 (commented):
I am pretty sure 2ac6a07676cf71ac8b5882507a4a20f944cb6706 accidentally disabled all the code paths guarded by `#ifdef KOKKOS_IMPL_HIERARCHICAL_INTEL_GPU`

core/src/OpenMPTarget/Kokkos_OpenMPTarget_ParallelFor_Team.hpp
```diff
@@ -115,44 +115,68 @@ class ParallelFor<FunctorType, Kokkos::TeamPolicy<Properties...>,
...
-              blockIdx, shmem_size_L0, shmem_size_L1);
-          if constexpr (std::is_void<TagType>::value)
-            m_functor(team);
-          else
-            m_functor(TagType(), team);
-        }
-      } else
-        Kokkos::abort("`num_teams` clause was not respected.\n");
+      for (int league_id = blockIdx; league_id < league_size;
+           league_id += gridDim) {
```
masterleinad, line 150: `gridDim` is used here.

core/src/OpenMPTarget/Kokkos_OpenMPTarget_ParallelFor_Team.hpp
rgayatri23: Yes thats the intention, that if there is a bug in OpenMP, dont run it because in this case it might lead to race conditions.

dalg24:
> > I am pretty sure [2ac6a07](https://github.com/kokkos/kokkos/commit/2ac6a07676cf71ac8b5882507a4a20f944cb6706) accidentally disabled all the code paths guarded by `#ifdef KOKKOS_IMPL_HIERARCHICAL_INTEL_GPU`
> 
> I don't understand, the macro is now in `core/src/OpenMPTarget/Kokkos_OpenMPTarget_Parallel.hpp` . Am I missing something?

It is `#undef`d at the end of the header and never visible from the parallel constructs implementations

dalg24:
> Ok now `undef`ing from files that include the file that defines the macro.

That means only one of the files will have the new code path enabled, whichever gets included first.

rgayatri23:
> > Ok now `undef`ing from files that include the file that defines the macro.
> 
> That means only one of the files will have the new code path enabled, whichever gets included first.

Oh right thats true. I need to undef it in a common place only once.

rgayatri23:
> > > Ok now `undef`ing from files that include the file that defines the macro.
> > 
> > 
> > That means only one of the files will have the new code path enabled, whichever gets included first.
> 
> Oh right thats true. I need to undef it in a common place only once.

Removed the `undef` and added the macro in print_configuration.

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Instance.cpp
```diff
@@ -66,13 +67,39 @@ void OpenMPTargetInternal::fence(const std::string& name,
...
+
+  return max_threads;
 }
 const char* OpenMPTargetInternal::name() { return "OpenMPTarget"; }
 void OpenMPTargetInternal::print_configuration(std::ostream& os,
                                                bool /*verbose*/) const {
   // FIXME_OPENMPTARGET
   os << "Using OpenMPTarget\n";
+#if defined(KOKKOS_OPENEMPTARGET_HIERARCHICAL_INTEL_GPU)
```
masterleinad, line 99: I guess you want
```suggestion
#if defined(KOKKOS_OPENMPTARGET_HIERARCHICAL_INTEL_GPU)
```
here and elsewhere.

dalg24: Make sure it is a KOKKOS_IMPL_OPENMPTARGET_ macro (implementation only)

masterleinad (approved):
Looks OK to me (apart from the typo you might want to fix).

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Instance.cpp
```diff
@@ -66,13 +67,39 @@ void OpenMPTargetInternal::fence(const std::string& name,
   }
 }
 int OpenMPTargetInternal::concurrency() const {
-  return 128000;  // FIXME_OPENMPTARGET
+  int max_threads = 2048 * 80;
+#if defined(KOKKOS_IMPL_ARCH_NVIDIA_GPU)
+  int max_threads_block = 2048;
+#if defined(KOKKOS_ARCH_AMPERE86)
+  max_threads = max_threads_block * 1536;
```
crtrott, line 74: this doesn't make any sense

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Instance.cpp
```diff
@@ -66,13 +67,39 @@ void OpenMPTargetInternal::fence(const std::string& name,
   }
 }
 int OpenMPTargetInternal::concurrency() const {
-  return 128000;  // FIXME_OPENMPTARGET
+  int max_threads = 2048 * 80;
+#if defined(KOKKOS_IMPL_ARCH_NVIDIA_GPU)
+  int max_threads_block = 2048;
```
crtrott, line 72: you mean max_threads_sm not block (which is 1024)

core/src/OpenMPTarget/Kokkos_OpenMPTarget_ParallelFor_Team.hpp
```diff
@@ -115,44 +115,68 @@ class ParallelFor<FunctorType, Kokkos::TeamPolicy<Properties...>,
     // mode but works in the Debug mode.
 
     // Maximum active teams possible.
-    int max_active_teams = OpenMPTargetExec::MAX_ACTIVE_THREADS / team_size;
-    // nteams should not exceed the maximum in-flight teams possible.
-    const auto nteams =
-        league_size < max_active_teams ? league_size : max_active_teams;
+    int max_active_teams = omp_get_max_teams();
```
crtrott, line 118: Why not `min(nteams, omp_get_max_team())`

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Instance.cpp
rgayatri23: Do you mean change the name to `max_threads_sm`  instead of `_block` ? 
And for AMPERE86 to not multiply by 1536 (which is the number of SMs) ?

crtrott (changes requested):
Couple small things

core/src/OpenMPTarget/Kokkos_OpenMPTarget_ParallelFor_Team.hpp
rgayatri23: That `min` is happening in `resize_scratch` and the resulting value is set in `omp_set_num_teams` . 
The value is then being accessed here using `omp_get_max_teams` rather than setting another variable that is passed between routines.

crtrott (approved):

PR #6045, opened 2023-04-11
Add half math functions

e10harvey:
Related to https://github.com/kokkos/kokkos-kernels/pull/1774.


dalg24 (changes requested):
These need to be added to the <Kokkos_Half.hpp> header

core/src/Kokkos_Half.hpp
```diff
@@ -23,6 +23,7 @@
 
 #include <Kokkos_Macros.hpp>
 #include <Kokkos_NumericTraits.hpp>
+//#include <Kokkos_MathematicalFunctions.hpp>
```
lucbv, line 26: Looks like this can be removed

core/src/Kokkos_Half.hpp
```diff
@@ -1254,6 +1260,19 @@ struct Kokkos::Experimental::Impl::max_exponent_helper<
 
 ////////////// BEGIN BHALF_T (bfloat16) limits //////////////
 #if defined(KOKKOS_BHALF_T_IS_FLOAT) && !KOKKOS_BHALF_T_IS_FLOAT
+/// \brief: Infinity
+///
+///
+/// Bfloat16 encoding:
+///             [s  e  e  e  e  e  e  e  e f f f f f f f]
+///             [0  1  1  1  1  1  1  1  1 0 0 0 0 0 0 0]
+/// bit index:   15 14 13 12 11 10 9 8 7 6 5 4 3 2 1 0
```
lucbv, line 1269: Looks like you need an extra space somewhere to align this properly?
If it looks good in the documentation then its fine.

lucbv (commented):
It looks fine to me just a couple clean ups needed

core/src/Kokkos_Half.hpp
```diff
@@ -1059,15 +1059,20 @@ cast_from_bhalf(bhalf_t val) {
...
 /// Binary16 encoding:
 ///             [s  e  e  e  e  e  f f f f f f f f f f]
 ///             [0  1  1  1  1  1  0 0 0 0 0 0 0 0 0 0]
-/// bit index:  15 14 13 12 11 10 9 8 7 6 5 4 3 2 1 0
-#if defined(KOKKOS_HALF_T_IS_FLOAT) && !KOKKOS_HALF_T_IS_FLOAT
+/// bit index:   15 14 13 12 11 10 9 8 7 6 5 4 3 2 1 0
+///
+template <>
+struct Kokkos::Experimental::Impl::infinity_helper<Kokkos::Experimental::half_t> {
+  static constexpr int value = 0x7C00;
```
dalg24, line 1074: Please add the missing numerics trait in a separate PR

core/src/Kokkos_Half.hpp
```diff
@@ -1328,7 +1346,227 @@ struct Kokkos::Experimental::Impl::max_exponent_helper<
...
+KOKKOS_IMPL_MATH_UNARY_PREDICATE_HALF(isnan)
+KOKKOS_IMPL_MATH_UNARY_PREDICATE_HALF(signbit)
+
+#undef KOKKOS_IMPL_MATH_UNARY_PREDICATE_HALF
+
+KOKKOS_INLINE_FUNCTION Kokkos::Experimental::half_t real(Kokkos::Experimental::half_t x) {
+  return x;
+}
+
+KOKKOS_INLINE_FUNCTION Kokkos::Experimental::half_t imag([[maybe_unused]] Kokkos::Experimental::half_t x) {
```
dalg24, line 1459:
```suggestion
KOKKOS_INLINE_FUNCTION Kokkos::Experimental::half_t imag(Kokkos::Experimental::half_t) {             
```

core/src/Kokkos_Half.hpp
```diff
@@ -1328,7 +1346,227 @@ struct Kokkos::Experimental::Impl::max_exponent_helper<
...
+#endif
+#endif
+
+namespace Kokkos {
+////////////// BEGIN HALF_T (float16) MATH FNS ////////////
+#if defined(KOKKOS_HALF_T_IS_FLOAT) && !KOKKOS_HALF_T_IS_FLOAT
+#define KOKKOS_IMPL_MATH_UNARY_FUNCTION_HALF(FUNC)                \
+  KOKKOS_INLINE_FUNCTION Kokkos::Experimental::half_t FUNC(       \
+      Kokkos::Experimental::half_t x) {                           \
+    using KOKKOS_IMPL_MATH_FUNCTIONS_NAMESPACE::FUNC;             \
```
dalg24, line 1368: If you are going to cast to float then defer to Kokkos::FUNC

core/src/Kokkos_Half.hpp
```diff
@@ -1328,7 +1346,227 @@ struct Kokkos::Experimental::Impl::max_exponent_helper<
...
+    return static_cast<Kokkos::Experimental::half_t>(             \
+        FUNC(static_cast<float>(x)));                             \
```
dalg24, line 1370:
```suggestion
    return static_cast<Kokkos::Experimental::half_t>(             \
        Kokkos::FUNC(static_cast<float>(x)));                             \
```
(not formatted)

core/src/Kokkos_Half.hpp
crtrott, line 1368: Not needed, just call Kokkos::FUNC, after all you anyway call the float version.

core/src/Kokkos_Half.hpp
```diff
@@ -1328,7 +1346,227 @@ struct Kokkos::Experimental::Impl::max_exponent_helper<
...
+  }                                                               \
+  KOKKOS_INLINE_FUNCTION Kokkos::Experimental::half_t FUNC##fp16( \
```
dalg24, line 1372: Please comment about the suffixed versions that are being added.
Is there prior practice of using that suffix?  Do we need/want them?

crtrott, line 1372: We probably shouldn't provide these: there is no clear convention for this (for example NVIDIA has hexp not exp_fp16, and uses hexp for both half and bhalf. So lets leave this overload out.

crtrott (changes requested):
I also think we should split the file, into non-public subfiles and then include them all in Kokkos_Half.hpp

cwpearson (approved):

masterleinad (approved):
Looks OK to me.

e10harvey:
@dalg24, @crtrott: Ready to merge?

dalg24:
> @dalg24, @crtrott: Ready to merge?

How does it relate to #6055 ?
There is no proper description and we are back to one large PR that does things can/should be separated out.

e10harvey:
> > 
> 
> How does it relate to #6055 ?

Good question. This is simply related to 6055 in that the traits and infinity-helper were requested to be put in a separate PR (which is now 6055).

> There is no proper description and we are back to one large PR that does things can/should be separated out.

I am not sure I follow. The original motivation for this PR was to add the half-precision math functions and during review it was requested that the traits and math fns be moved into separate headers. Are you saying you want 3 PRs:
1. Move the floating_point_wrapper to a private header file
2. Move half traits to a private header file
3. Add half math fns to a private header file

If so, that's no problem. I can create 3 PRs.

core/src/impl/Kokkos_Half_MathematicalFunctions.hpp
```diff
@@ -0,0 +1,214 @@
...
+namespace Kokkos {
+////////////// BEGIN HALF_T (float16) MATH FNS ////////////
+#if defined(KOKKOS_HALF_T_IS_FLOAT) && !KOKKOS_HALF_T_IS_FLOAT
+#define KOKKOS_IMPL_MATH_UNARY_FUNCTION_HALF(FUNC)          \
+  KOKKOS_INLINE_FUNCTION Kokkos::Experimental::half_t FUNC( \
+      Kokkos::Experimental::half_t x) {                     \
+    return static_cast<Kokkos::Experimental::half_t>(       \
+        Kokkos::FUNC(static_cast<float>(x)));               \
+  }
+
```
crtrott, line 31: Is there any reason not to do half_t and bhalf_t in one go if both are available? Are there any math functions which don't work in both cases?

```c++
#define KOKKOS_IMPL_MATH_UNARY_FUNCTION_HALF_TYPE(FUNC, TYPE)          \
  KOKKOS_INLINE_FUNCTION TYPE FUNC( TYPE x)                          \
    { return static_cast<TYPE>(  Kokkos::FUNC(static_cast<float>(x)));  }
    
#if defined(KOKKOS_HALF_T_IS_FLOAT) && !KOKKOS_HALF_T_IS_FLOAT
  #if defined(KOKKOS_BHALF_T_IS_FLOAT) && !KOKKOS_BHALF_T_IS_FLOAT
    #define KOKKOS_IMPL_MATH_UNARY_FUNCTION_HALF(FUNC) \
            KOKKOS_IMPL_MATH_UNARY_FUNCTION_HALF_TYPE(FUNC, Kokkos::Experimental::half_t) \
            KOKKOS_IMPL_MATH_UNARY_FUNCTION_HALF_TYPE(FUNC, Kokkos::Experimental::bhalf_t)
  #else
    #define KOKKOS_IMPL_MATH_UNARY_FUNCTION_HALF(FUNC) \
         KOKKOS_IMPL_MATH_UNARY_FUNCTION_HALF_TYPE(FUNC, Kokkos::Experimental::half_t)
  #endif
#endif
```

This way we won't overlook any instantiation.

core/src/impl/Kokkos_Half_MathematicalFunctions.hpp
```diff
@@ -0,0 +1,214 @@
...
+KOKKOS_IMPL_MATH_BINARY_FUNCTION_HALF(nextafter)
+KOKKOS_IMPL_MATH_BINARY_FUNCTION_HALF(copysign)
+
+#undef KOKKOS_IMPL_MATH_BINARY_FUNCTION_HALF
+
+KOKKOS_INLINE_FUNCTION Kokkos::Experimental::half_t abs(
+    Kokkos::Experimental::half_t x) {
+  return static_cast<Kokkos::Experimental::half_t>(
+      Kokkos::abs(static_cast<float>(x)));
+}
```
crtrott, line 91: In this case abs isn't different from the other Unary is it? Abs is a special case on the underlying math functions due to the way how abs is part of the pure integer stuff and the cmath in the standard, but it is not here a special case or?

crtrott (changes requested):
I think we can do the mathematical functions simpler or? Otherwise looks ok.

e10harvey:
@crtrott, @dalg24: I think I addressed your feedback. The CI failure in SYCL is unrelated to these changes. Ready to merge?

dalg24:
> @crtrott, @dalg24: I think I addressed your feedback. The CI failure in SYCL is unrelated to these changes. Ready to merge?

As far as I am concerned no.  My objections are there is no proper description (you need to edit) and this patch is doing 3 things that really should be split up in 3 PRs.  I am not blocking if Christian decides your last changes addressed his concerns and he wants to merge.

At the moment you have two stales approval (approvals but for a version of the patch that is significantly out of date) and one request changes.  There is not enough consensus to merge yet.

e10harvey:
Closing. Please review https://github.com/kokkos/kokkos/pull/6055. I will open 2 more PRs -- one for moving floating point wrapper to a private header and another for the half math functions.

PR #6048, opened 2023-04-12
Remove libnuma

dalg24:
libnuma was documented as a TPL and one could in fact link against it but it is not used in any way hence I suggest simple removal.


masterleinad (approved):
I can't find any use in the history.

dalg24:
> Do we need to do anything about removing a public option?

I'd say no.  Regular CMake warning that option was ignored is good enough and I don't think we need to guard for downstream use of libnuma that would have been coming from the Kokkos include/link

PR #6050, opened 2023-04-12
Fixup typo `nan[q]` in quad precision math function

dalg24:
Suffix was accidentally added


crtrott (approved):

Rombur (approved):

masterleinad (commented):
Are we missing a unit test that should have catched this?

PR #4385, opened 2021-10-06
[WIP] Thread safety tests

Rombur:
This PR adds tests that launch kernels from `std::thread` in order to check basic thread safety. The tests cover:
 - `parallel_for` for `Range`, `MDRange`, and `Team`
 - `parallel_reduce` for `Range`, `MDRange`, and `Team`
 - `parallel_scan` for `Range`
 
The tests are disabled for OpenMP. With the `OpenMP` backend, we cannot launch a kernel from `std::thread`. This is because only the master thread can launch a kernel (possible fix #4119).

I've opened the PR to trigger the CI. It's not ready for review yet.

PR #6051, opened 2023-04-12
Support completion query for execution spaces

### LINKED ISSUE #4209: Support completion query of async operations
We had a few times a request to support querying completion of async operations. In CUDA you can do this with events something like this:

```c++
cudaEvent_t event;
cudaEventCreateWithFlags(&event,cudaEventDisableTiming);
launch_some_kernel<<<grid,block,stream>>>(args);
cudaEventRecord(event,stream);

// check whether the kernel is done?
while(cudaEventQuery(event) != cudaSuccess) {
   // ...
}
```

Now if we were to do this always I think we would increase the kernel launch time slightly. In fact a simple test code (see below) shows an increase from 2.3us to 2.6us for kernel latency benchmark. We could just eat that but I think @stanmoore1 would complain again :-). Also it would not give us individual stuff but only a "last thing done in the execution space instance". 

I.e. something like:
```c++
parallel_for(RangePolicy(exec,0,N),F);

while(exec.still_working()) { ... };
```

Alternatively we can add an opt-in interface. Something like:

```c++
Kokkos::CompletionEvent event;
parallel_for(RangePolicy(0,N).set_completion_event(event), F);

while(!event.is_done());
```

Or:
```c++
Kokkos::CompletionEvent event;
parallel_for(RangePolicy(0,N), F, event);

while(!event.is_done());
```

Here is my benchmark code:

```c++
#include <Kokkos_Core.hpp>
#include <cmath>

struct F {
  Kokkos::View<double> a;
  KOKKOS_FUNCTION
  void operator() (int) const {
    a()+=1.0;
  }
};

  {
    int N = argc > 1 ? atoi(argv[1]) : 10000;
    Kokkos::View<double> a("Foo");
    F f{a};
    Kokkos::parallel_for(1,f);

    Kokkos::fence();
    Kokkos::Timer timer;
    for(int i=0; i<N; i++) {
      Kokkos::parallel_for(1,f);
    }
    Kokkos::fence();
    printf("%lf\n",time*1.e6/N);

    cudaEvent_t event;
    cudaEventCreateWithFlags(&event,cudaEventDisableTiming);
    Kokkos::parallel_for(1,f);
    cudaEventRecord(event);
    Kokkos::fence();
    timer.reset();
    for(int i=0; i<N; i++) {
      Kokkos::parallel_for(1,f);
      cudaEventRecord(event);
    }
    Kokkos::fence();
    time = timer.seconds();
    printf("%lf\n",time*1.e6/N);
  }
  Kokkos::finalize();
}
```

masterleinad:
Fixes #4209.


core/src/Threads/Kokkos_Threads.hpp
```diff
@@ -71,6 +72,12 @@ class Threads {
   //! \name Static functions that all Kokkos devices must implement.
   //@{
 
+  Experimental::ExecutionSpaceStatus get_status() const
+  {
+    // Threads` is synchronous
```
PhilMiller, line 76: Typo

core/src/Cuda/Kokkos_Cuda_Instance.hpp
```diff
@@ -130,6 +131,9 @@ class CudaInternal {
   mutable size_type* m_scratchUnified;
   mutable size_type* m_scratchFunctor;
   cudaStream_t m_stream;
+  cudaEvent_t m_last_event;
```
PhilMiller, line 134: I think this will have to be `mutable` too?

core/src/HIP/Kokkos_HIP_Instance.hpp
```diff
@@ -95,6 +95,8 @@ class HIPInternal {
   inline static std::mutex scratchFunctorMutex;
 
   hipStream_t m_stream = nullptr;
+   hipEvent_t m_last_event;
```
PhilMiller: Ditto `mutable` here?

PhilMiller:
This looks good for what it is. EMPIRE will try using it as soon as it makes its way downstream.

PhilMiller:
I know this is still a draft, but why shouldn't a fence call leave the instance in the completed state?

masterleinad:
> I know this is still a draft, but why shouldn't a fence call leave the instance in the completed state?

It does. Since the implementation immediately queries the event after recording it, the status is either `running` or `complete`. Calling `get_status` after a fence then either already uses the `complete` state or transitions from `running` to `complete`. That way, we don't need to try to have extra logic for `Kokkos::fence`.

core/src/Cuda/Kokkos_Cuda_Instance.hpp
masterleinad: No, we only access this member through a pointer to the internal class.

PhilMiller:
> > I know this is still a draft, but why shouldn't a fence call leave the instance in the completed state?
> 
> It does. Since the implementation immediately queries the event after recording it, the status is either `running` or `complete`. Calling `get_status` after a fence then either already uses the `complete` state or transitions from `running` to `complete`. That way, we don't need to try to have extra logic for `Kokkos::fence`.

Doesn't that entail an extra record/query pair that setting 'complete' in `fence` avoids?

PhilMiller:
From my comments on the issue - from the caller's perspective, is there a meaningful distinction between 'submitted' and 'running', or are they both effectively just 'incomplete'? CUDA and HIP as implemented will never return 'submitted', and that seems fine. If the three states are just an internal detail, and the interface just returned boolean in/complete, would we lose anything?

masterleinad:
> Doesn't that entail an extra record/query pair that setting 'complete' in `fence` avoids?

It's potentially another `query`, yes. My motivation, for now, was to make sure that it doesn't behave differently from calling `Kokkos::fence`.

PhilMiller:
> > Doesn't that entail an extra record/query pair that setting 'complete' in `fence` avoids?
> 
> It's potentially another `query`, yes. My motivation, for now, was to make sure that it doesn't behave differently from calling `Kokkos::fence`.

Could you elaborate on what you mean by that?

I think the desired equivalence, in terms of synchronization semantics, is that `space.get_status() == complete` allows the same subsequent uses as `space.fence()`. Is that understanding correct?

If so, does the extra query avoid some troublesome case I'm missing?

masterleinad:
> > > Doesn't that entail an extra record/query pair that setting 'complete' in `fence` avoids?
> 
> > It's potentially another `query`, yes. My motivation, for now, was to make sure that it doesn't behave differently from calling `Kokkos::fence`.
>
> Could you elaborate on what you mean by that?

We can't (easily) set the status in `Kokkos::fence()` since we don't keep a record of execution space instances (for `CUDA` and `HIP` at least). Instead of adding that (which we have for `SYCL`), I rather wanted to make sure that we would get the desired behavior for fences without setting the status explicitly. Again, `exec.fence` would either already see `complete` or query the (already recorded) event which should also result in `complete`. We could avoid the latter case for `exec.fence` but not for `Kokkos::fence`. 

> I think the desired equivalence, in terms of synchronization semantics, is that space.get_status() == complete allows the same subsequent uses as space.fence(). Is that understanding correct?

Yes, `while(exec.get_status != ExecutionSpaceStatus::complete){}` is essentially equivalent to calling `exec.fence()` (but only `fence` would be recorded in our profiling facilities).

> If so, does the extra query avoid some troublesome case I'm missing?

No, we could add it as an optimization but I would like to get feedback from more people before making changes.

core/unit_test/CMakeLists.txt
```diff
@@ -113,40 +113,7 @@ foreach(Tag Threads;Serial;OpenMP;Cuda;HPX;OpenMPTarget;OpenACC;HIP;SYCL)
...
-        HostSharedPtr
-        HostSharedPtrAccessOnDevice
-        Init
-        JoinBackwardCompatibility
-        LocalDeepCopy
-        MathematicalConstants
-        MathematicalFunctions1
-        MathematicalFunctions2
-        MathematicalFunctions3
-        MathematicalSpecialFunctions
```
PhilMiller, line 149: This needs to be reverted before merge.

masterleinad:
After briefly discussing this in the meeting, it seems that people think that returning the status via a `get_status ` member variable is not quite adequate; querying the status would be more suitable for kernel events. Instead, we were thinking of a boolean return type. Opinions on the name?
- `bool ExdcutionSpace::is_complete()`
- `bool ExecutionSpace::is_running()`
- `bool ExecutionSpace::is_working()`
- `bool ExecutionSPace::poll()`
- ...

core/src/Threads/Kokkos_Threads.hpp
PhilMiller:
```suggestion
    // Threads is synchronous
```

core/unit_test/TestExecutionSpace.hpp
```diff
@@ -50,4 +50,61 @@ TEST(TEST_CATEGORY, execution_space_as_class_data_member) {
...
+                      "functionality in OpenACC backend";
+    }
+#endif
+
+    const int N = 10000;
+
+    Kokkos::parallel_reduce(Kokkos::RangePolicy<ExecutionSpace>(exec, 0, N),
+                            *this, result_view);
+    Kokkos::deep_copy(exec, result, result_view);
+    while (exec.is_running() == true) {
```
dalg24, line 93:
```suggestion
    while (exec.is_running()) {
```

dalg24: Per dev meeting add a comment that this happens to work for now but that we don't want users to write this code and we reserve the right to have a lazy implementation which would make this hang.

rgayatri23:
> After briefly discussing this in the meeting, it seems that people think that returning the status via a `get_status ` member variable is not quite adequate; querying the status would be more suitable for kernel events. Instead, we were thinking of a boolean return type. Opinions on the name?
> 
> * `bool ExecutionSpace::is_complete()`
> * `bool ExecutionSpace::is_running()`
> * `bool ExecutionSpace::is_working()`
> * `bool ExecutionSpace::poll()`
> * ...

I like the `poll` naming, its has less number of characters to type (;-)). But `poll` is not as self explanatory as others. 
Just FYI - OpenMPTarget is also asynchronous when `nowait` clause is added. I am not sure how many compilers respect that clause but llvm certainly does. Right now we only use it for hierarchical parallel_for.

core/src/Cuda/Kokkos_Cuda.hpp
```diff
@@ -241,28 +244,6 @@ struct DeviceTypeTraits<Cuda> {
...
-  }
-
-  ZeroMemset(const View<DT, DP...>& dst,
-             typename View<DT, DP...>::const_value_type&) {
-    KOKKOS_IMPL_CUDA_SAFE_CALL(
-        cudaMemset(dst.data(), 0,
-                   dst.size() * sizeof(typename View<DT, DP...>::value_type)));
-  }
-};
-}  // namespace Impl
```
Rombur, line 265: This will create a conflict with https://github.com/kokkos/kokkos/pull/6087 
cc: @tcclevenger

core/src/Cuda/Kokkos_Cuda_Instance.cpp
```diff
@@ -684,6 +687,26 @@ Cuda::size_type *cuda_internal_scratch_unified(const Cuda &instance,
...
+    case Experimental::ExecutionSpaceStatus::submitted:
+      KOKKOS_IMPL_CUDA_SAFE_CALL(cudaEventRecord(m_space_instance->m_last_event,
+                                                 m_space_instance->m_stream));
+      [[fallthrough]];
+    case Experimental::ExecutionSpaceStatus::running:
+      if (cudaEventQuery(m_space_instance->m_last_event) == cudaSuccess) {
+        m_space_instance->m_internal_status =
+            Experimental::ExecutionSpaceStatus::complete;
+        return false;
+      }
```
Rombur, line 702: You need to add sth like
```c++
else
{
m_space_instance->m_internal_status =
            Experimental::ExecutionSpaceStatus::running;
}
```
otherwise you can never get to the `case` directly.

core/src/HPX/Kokkos_HPX.hpp
```diff
@@ -201,6 +202,12 @@ class HPX {
     return impl_get_instance_data().m_instance_id;
   }
 
+  bool is_running() const {
+    // HPX doesn't provide a query function.
+    fence("HPX::is_running");
+    return false;
```
Rombur, line 208: @msimberg is this the correct solution?

core/src/SYCL/Kokkos_SYCL.hpp
```diff
@@ -78,6 +79,19 @@ class SYCL {
   //! \name Functions that all Kokkos devices must implement.
   //@{
 
+  bool is_running() const {
+    const sycl::info::event_command_status sycl_status =
+        m_space_instance->m_last_event
+            .get_info<sycl::info::event::command_execution_status>();
+    switch (sycl_status) {
+      case sycl::info::event_command_status::submitted:
```
Rombur, line 87: You don't need `[[fallthrough]]`  to avoid warnings?

core/src/Cuda/Kokkos_Cuda.hpp
masterleinad: Of course. 🙂 I expect #6087 to be merged before this and I will rebase then.

core/src/SYCL/Kokkos_SYCL.hpp
masterleinad: No, compilers don't warn for collapsed cases.

core/src/Cuda/Kokkos_Cuda_Instance.cpp
masterleinad: Improved.

rgayatri23 (approved):
LGTM. Does not have any major updates for the OpenMPTarget backend at least.

Rombur (approved):
The PR looks fine but I would like a confirmation that we are doing the right thing for HPX.

core/unit_test/TestExecutionSpace.hpp
```diff
@@ -50,4 +50,63 @@ TEST(TEST_CATEGORY, execution_space_as_class_data_member) {
...
+  void check() const {
+    ExecutionSpace exec;
+    ASSERT_EQ(exec.is_running(), false);
+
+    Kokkos::View<int, typename ExecutionSpace::memory_space> result_view(
+        "result_view");
+    int result;
+
+    Kokkos::deep_copy(exec, result, result_view);
+    while (exec.is_running() == true) {
```
msimberg, line 68:
```suggestion
    while (exec.is_running()) {
```
?

core/src/HPX/Kokkos_HPX.hpp
msimberg: It's not correct, but I guess the question right now is: is it sufficient?

If the primary use case is "fencing" with `while (space.is_running) {}` then this is not going to work (only if `Kokkos_INTERNAL_HPX_ENABLE_ASYNC_DISPATCH=OFF` but that's not the default and shouldn't really be considered for this) for HPX. If a conservative default would be sufficient for now there's an internal counter for the number of kernels submitted to _any_ HPX execution space instance. It would lead to false positives rather than the false negatives that the current `return false` leads to.

The correct solution would require tracking the number of kernels per execution space instance. That shouldn't be hard to add but it's not there at the moment. Is that something you'd like to see added already before this PR is merged or could that be added afterwards?

Finally, note that for the HPX backend `while (space.is_running()) {}` and `space.fence()` would not be 100% equivalent since the latter can yield an HPX thread if the instance is still running a kernel.

msimberg:
I tried to follow the discussion here but I may have missed the answer somewhere. Do I understand correctly that the statuses mentioned in the description

>    ExecutionSpaceStatus::submitted
>    ExecutionSpaceStatus::running
>    ExecutionSpaceStatus::complete

are not actually exposed to users (even though the enum seems to be public), i.e. the only public functionality currently in this PR is `bool ExecSpace::is_running()`?

masterleinad:
@msimberg Yes, the statuses don't need to be exposed to the user (anymore). The question for you is if there is any functionality in `HPX` that lets you query an `HPX` execution space instance for the completion of all kernels submitted to it.

core/src/HPX/Kokkos_HPX.hpp
masterleinad: We discussed that we don't want to promise `while(space.is_running()) {}` to act as a fence but an implementation could deadlock on it (although we test that currently in the unit test). The primary use case is querying if the submitted work has been finished and deciding based on the answer if something else should be done before querying again. 
I would consider `fencing` and returning `false` to be a valid generic implementation that we can improve upon for execution spaces that allow more introspection. I would be happy to see these improvements in another pull request (given reviews agree).

msimberg: Thanks @masterleinad for the explanation. And I'm sorry, I completely missed the `fence` before returning `false`! In that case yes, this works as a fallback, and I can separately add a way to keep track of kernels launched per-instance. I can then update this to remove the fence.

msimberg:
> @msimberg Yes, the statuses don't need to be exposed to the user (anymore). The question for you is if there is any functionality in `HPX` that lets you query an `HPX` execution space instance for the completion of all kernels submitted to it.

You already saw my reply in https://github.com/kokkos/kokkos/pull/6051#discussion_r1187147948, but I'll TL;DR here: There's currenly only a way to query for completion of kernels to _all_ HPX execution spaces instances collectively, but not per-instance. The latter would be easy to add though.

masterleinad:
> Hi Guys -- do you want/need more discussion on this issue in our Wed. meeting?

We just need reviews.

core/src/Cuda/Kokkos_Cuda.hpp
```diff
@@ -39,6 +39,7 @@ static_assert(false,
 #include <Kokkos_Layout.hpp>
 #include <Kokkos_ScratchSpace.hpp>
 #include <Kokkos_MemoryTraits.hpp>
+#include <impl/Kokkos_ExecutionSpaceStatus.hpp>
```
PhilMiller, line 42: This include can now move down to the cpp files, I think

core/src/Cuda/Kokkos_Cuda_Instance.cpp
```diff
@@ -684,6 +687,27 @@ Cuda::size_type *cuda_internal_scratch_unified(const Cuda &instance,
...
+bool Cuda::is_running() const {
+  switch (m_space_instance->m_internal_status) {
+    case Impl::ExecutionSpaceStatus::complete: return false;
+    case Impl::ExecutionSpaceStatus::submitted:
+      KOKKOS_IMPL_CUDA_SAFE_CALL(cudaEventRecord(m_space_instance->m_last_event,
+                                                 m_space_instance->m_stream));
+      m_space_instance->m_internal_status = Impl::ExecutionSpaceStatus::running;
+      [[fallthrough]];
+    case Impl::ExecutionSpaceStatus::running:
+      if (cudaEventQuery(m_space_instance->m_last_event) == cudaSuccess) {
```
PhilMiller, line 699: This can't use the usual macro, I guess, because we're expecting an in-progress return, but any values other than that or success are still hard errors that we should handle accordingly.

PhilMiller:
Ditto for HIP on both of the above, and maybe SYCL?

core/src/SYCL/Kokkos_SYCL.hpp
```diff
@@ -34,6 +34,7 @@ static_assert(false,
 #include <SYCL/Kokkos_SYCL_Space.hpp>
 #include <Kokkos_Layout.hpp>
 #include <Kokkos_ScratchSpace.hpp>
+#include <impl/Kokkos_ExecutionSpaceStatus.hpp>
```
PhilMiller, line 37: I don't think this gets used at all in the SYCL backend.

PhilMiller (commented):
The error checking may be important. Other than that, I think I'm happy with this.

core/src/Cuda/Kokkos_Cuda_Instance.cpp
```diff
@@ -684,6 +688,30 @@ Cuda::size_type *cuda_internal_scratch_unified(const Cuda &instance,
...
+      [[fallthrough]];
+    case Impl::ExecutionSpaceStatus::running:
+      const cudaError_t query_result =
+          cudaEventQuery(m_space_instance->m_last_event);
+      if (query_result == cudaSuccess) {
+        m_space_instance->m_internal_status =
+            Impl::ExecutionSpaceStatus::complete;
+        return false;
+      } else if (query_result != cudaErrorNotReady)
+        KOKKOS_IMPL_CUDA_SAFE_CALL(query_result);
```
PhilMiller, line 707: Stylistically, the lack of braces in just the `else` arm threw me off here. Not a big deal, though

PhilMiller (approved):

core/src/Cuda/Kokkos_Cuda_Instance.cpp
```diff
@@ -684,6 +688,30 @@ Cuda::size_type *cuda_internal_scratch_unified(const Cuda &instance,
...
+      return true;
```
PhilMiller, line 708:
```suggestion
      {
      const cudaError_t query_result =
          cudaEventQuery(m_space_instance->m_last_event);
      if (query_result == cudaSuccess) {
        m_space_instance->m_internal_status =
            Impl::ExecutionSpaceStatus::complete;
        return false;
      } else if (query_result != cudaErrorNotReady) {
        KOKKOS_IMPL_CUDA_SAFE_CALL(query_result);
      }
      return true;
      }
```
Adding the outer braces to hopefully placate CUDA compiler pickiness about `default:` jumping past initialization of `query_result`, and the inner braces that I asked for on the `else` arm.

PhilMiller: This would need to be run through `clang-format`

masterleinad: See 0cb6c12408eebbae73897ea933416aac8021aee1

PhilMiller (dismissed):
The one failing Jenkins test was noise. Compiler had an io error

core/src/OpenMP/Kokkos_OpenMP.hpp
```diff
@@ -84,6 +84,11 @@ class OpenMP {
 
   OpenMP(int pool_size);
 
+  bool is_running() const {
+    // OpenMP is synchronous
+    return false;
+  }
```
crtrott, line 90: this is not correct with multiple instances.

core/src/Serial/Kokkos_Serial.hpp
```diff
@@ -108,6 +108,11 @@ class Serial {
 
   Serial();
 
+  bool is_running() const {
+    // Serial is synchronous
+    return false;
+  }
```
crtrott, line 114: not correct in the context of multiple threads.

core/src/Threads/Kokkos_Threads.hpp
```diff
@@ -71,6 +71,11 @@ class Threads {
   //! \name Static functions that all Kokkos devices must implement.
   //@{
 
+  bool is_running() const {
+    // Threads is synchronous
+    return false;
+  }
```
crtrott, line 77: not correct in the context of multiple threads

crtrott (changes requested):
The statement is_running() return false is not correct for host backends in the context of user controlled threads launching work to Kokkos.

ajpowelsnl:
> The statement is_running() return false is not correct for host backends in the context of user controlled threads launching work to Kokkos.

Is the `is_running()` state for an `ExecutionSpace` highly context dependent?  If so, either a rethink is in order, or at a minimum, a table summarizing the states according to the context.

core/src/Cuda/Kokkos_CudaSpace.cpp
```diff
@@ -73,6 +74,8 @@ void DeepCopyAsyncCuda(const Cuda &instance, void *dst, const void *src,
                        size_t n) {
   KOKKOS_IMPL_CUDA_SAFE_CALL(
       cudaMemcpyAsync(dst, src, n, cudaMemcpyDefault, instance.cuda_stream()));
+  instance.impl_internal_space_instance()->m_internal_status =
+      Kokkos::Impl::ExecutionSpaceStatus::submitted;
```
dalg24, line 78: Technically potential thread safety issue.

masterleinad: Fixed.

masterleinad:
> Is the is_running() state for an ExecutionSpace highly context dependent? If so, either a rethink is in order, or at a minimum, a table summarizing the states according to the context.

@ajpowelsnl Would you mind elaborating? Is https://github.com/kokkos/kokkos-core-wiki/pull/368 sufficient for you?

core/src/Cuda/Kokkos_Cuda_Instance.cpp
```diff
@@ -684,6 +690,32 @@ Cuda::size_type *cuda_internal_scratch_unified(const Cuda &instance,
...
+    case Impl::ExecutionSpaceStatus::complete: return false;
+    case Impl::ExecutionSpaceStatus::submitted:
+      KOKKOS_IMPL_CUDA_SAFE_CALL(cudaEventRecord(m_space_instance->m_last_event,
+                                                 m_space_instance->m_stream));
+      m_space_instance->m_internal_status = Impl::ExecutionSpaceStatus::running;
+      [[fallthrough]];
+    case Impl::ExecutionSpaceStatus::running: {
+      const cudaError_t query_result =
+          cudaEventQuery(m_space_instance->m_last_event);
+      if (query_result == cudaSuccess) {
```
PhilMiller, line 704: With `m_internal_status` being implemented as an atomic, we're not going to have data races resulting from accesses to it.

However, there's still a dangerous semantic race condition, in that another thread could submit a kernel launch here. If that happened, its kernel could be running, but its assignment of `m_internal_status = submitted` could be incorrectly overwritten by `complete`.

PhilMiller:
I was already skeptical that we could define this well in the absence of a broader discussion about what we mean by supporting multiple threads submitting work to a shared execution space instance.

The above identified race condition heightens that skepticism.

My preferred path forward is to merge this substantially as-is, since it provides useful functionality for the many codes not doing the unusual multi-threaded submission thing. Then, address multi-threaded Kokkos usage more holistically.

With that in mind, I think it may make sense to briefly reconsider the name and sense of the function we're providing, to not impede that later consideration of multi-threaded usage. In particular, if we call it `submitted_work_completed`, we can better hold a position that it's relative to only whatever was submitted by the calling thread (and any thread it has `synchronized-with`). I think that holds up for both the GPU-stream-like and the synchronous-host-like backends.

core/src/Cuda/Kokkos_Cuda_Instance.cpp
```diff
@@ -684,6 +690,32 @@ Cuda::size_type *cuda_internal_scratch_unified(const Cuda &instance,
 
 namespace Kokkos {
 
+bool Cuda::is_running() const {
+  switch (m_space_instance->m_internal_status) {
+    case Impl::ExecutionSpaceStatus::complete: return false;
+    case Impl::ExecutionSpaceStatus::submitted:
+      KOKKOS_IMPL_CUDA_SAFE_CALL(cudaEventRecord(m_space_instance->m_last_event,
```
PhilMiller, line 697: Is `cudaEventRecord` even thread-safe with respect to a common stream and event?

I could readily see even `cudaEventQuery` not being thread-safe, if it acts at all like an MPI progress engine in a call to `MPI_Wait` - i.e. actually advancing internal state.

PhilMiller:
Having looked over the CUDA version with a lock, my instincts are that the lock should encompass the `invoke_kernel` and `cudaMemcpyAsync` calls as well. My first pass analysis is that the current version is fine, but I'm concerned that I'm missing a different way of constructing a dangerous race condition.

core/src/Cuda/Kokkos_Cuda_Instance.cpp
PhilMiller: With the added lock, this is no longer a concern.

PhilMiller:
Actually, my question about `cudeEventRecord` and `cudaEventQuery` being thread safe applies to those other CUDA API functions as well - if they're not, then we need the lock to encompass them anyway.

PhilMiller:
Now for the other side of the coin: are we worried that the lock on every kernel launch is going to present unacceptable added overhead?

PhilMiller:
Does SYCL need locks added as well?

masterleinad:
> Now for the other side of the coin: are we worried that the lock on every kernel launch is going to present unacceptable added overhead?

https://github.com/kokkos/kokkos/pull/6151#issuecomment-1564392815 shows that locking and unlocking a mutex without contention guarding `Serial` execution of an empty kernel causes for me a 4% slowdown. I would expect that to be less for any real kernel and in particular for `Cuda`. Of course, we can measure again for this pull request.

PhilMiller:
Crud, there was another comment I thought I'd posted earlier, but Github seems to have eaten it, referencing TOCTTOU issues with multi-threaded callers using this functionality.

Any multi-threaded callers that don't synchronize among themselves are going to be very limited in what meaningful information they can get out of this. A result that all submitted work is done can only possibly refer to kernel launches with a happens-before relationship to the query call. A result that work is not done might refer to unknown kernels launched by concurrent threads, but without synchronizing with those threads, the caller can't know whether all expected work has been launched.

PhilMiller:
If we want to meaningfully support multi-threaded callers without external synchronization between them, we'd need an API more analogous to creating and recording explicit events on the stream, with each thread being able to query the event it actually placed.

PhilMiller:
One partial way to address the multi-threading concerns would be to rename to something like `submitted_work_has_completed` and reverse the sense accordingly. If it returns `true`, then the caller can be sure that any kernel launches that happen-before the query call have completed. If it returns `false`, the caller can try again.

We can't actually guarantee that it will ever return `true` in the face of uncoordinated threads submitting kernel launches, because an 'aggressor' thread could keep stepping on a 'victim' thread, with a repeating sequence where it
- aggressor submits work
- work finishes
- aggressor queries, sees its done, and submits more work

masterleinad:
> [#6151 (comment)](https://github.com/kokkos/kokkos/pull/6151#issuecomment-1564392815) shows that locking and unlocking a mutex without contention guarding `Serial` execution of an empty kernel causes for me a 4% slowdown. I would expect that to be less for any real kernel and in particular for `Cuda`. Of course, we can measure again for this pull request.

```C++
#include <Kokkos_Core.hpp>

  Kokkos::Timer timer;
  timer.reset();

  unsigned int N=10'000'000;

  for (unsigned int i=0; i<N; ++i)
    Kokkos::parallel_for(Kokkos::RangePolicy<Kokkos::Cuda>(0, 1), KOKKOS_LAMBDA(int){});

  Kokkos::fence();
  auto count_time = timer.seconds();
  std::cout << count_time << std::endl;

  Kokkos::finalize();
  return 0;
}
```
takes `39.9217s` with this pull request and `40.5779s` without it on a A100 so it looks like the overhead is negligible.

PhilMiller:
> > Does SYCL need locks added as well?
> 
> I think `SYCL` is fine using atomics since we are never setting the status instead of querying it from the runtime.

I don't think that's right. Consider
- thread 1 launches a kernel, and gets the associated event
- thread 2 launches a kernel, gets the associated event, and assigns it to the atomic instance member
- thread 1 assigns its event to the atomic instance member
- thread 1's kernel finishes
- thread 2 queries, which checks the status of thread 1's kernel, and gets told that it's done

Launching a kernel and storing the associated event have to be an indivisible operation.

masterleinad:
> I don't think that's right. Consider
> [...]
> Launching a kernel and storing the associated event have to be an indivisible operation.

I found another API function that should settle the issue either way. 🙂

masterleinad:
#6151 would allow for a better implementation for `Serial` but the implementation is semantically correct (even though the `Serial` backend might not yet do what we want).

ajpowelsnl:
**TODO:**
- Champions: @masterleinad , @ajpowelsnl 
- Request reviews for:
  - #4385 
  - #6051
  - #6151
- Discuss Kokkos "thread safety" concepts with team
- Decide where to document "thread safety" behavior , using this PR as a starting point

crtrott:
I dismissed the reviews but we gonna come back to this after some of the other instance stuff is done.

PR #6052, opened 2023-04-12
Don't create a shared state for size() in UnorderedMap's deep_copy

masterleinad:
https://github.com/kokkos/kokkos/pull/5983 copied the whole `m_size` `std::shared_ptr` instead of just its value. There should be no shared state created when calling `deep_copy`.


e10harvey (approved):

nmm0 (approved):
Looks good to me. What release is this targeted towards?

masterleinad:
> Looks good to me. What release is this targeted towards?

4.1. I added the label. AFAICT, we didn't cherry-pick #5983 for 4.0.01.

dalg24:
Ignoring likely unrelated failures (GCC-8.4.0, one CUDA build that landed on a new machine that was not setup properly and SYCL)

PR #5504, opened 2022-09-29
SYCL:  Remove workaround for submit_barrier not being enqueued properly

masterleinad:
Needs https://github.com/intel/llvm/pull/6359 and https://github.com/intel/llvm/pull/6888 (or manually setting `SYCL_PI_LEVEL_ZERO_USE_MULTIPLE_COMMANDLIST_BARRIERS=1` as environment variable).
Also updates the AOT architectures.
I would consider merging it already if it passes CI (which I expect).

PR #6056, opened 2023-04-13
Partially reverse #5504: SYCL: Remove workaround for submit_barrier not being enqueued properly

masterleinad:
Apparently, the bug is still in the version we are using for `SYCL+Cuda` (which is surprising since I thought it was limited to Level Zero). I verified that we don't need the workaround for `Intel GPUs` on the testbeds anymore.


dalg24 (approved):

masterleinad:
`KokkosAlgorithms_UnitTest_StdSet_E` is failing but we have seen that before occasionally (and this pull request fixes many more tests). We should try to address that elsewhere.

cz4rs (approved):

PR #6059, opened 2023-04-14
Reduce the size of ScatterView in the CI when using OpenMP

Rombur:
This fixes the CI issues we are having with gcc 8.4.  Oak Ridge started to enforce limits on the memory used by our CI instances.  This PR reduces the memory  footprint of the ScatterView test.


masterleinad (approved):
Looks OK to me.

rgayatri23 (approved):

rgayatri23:
LGTM

crtrott:
The problem with this is that it will likely not surface race conditions. But lets merge this yet to have testing pass, and then discuss this again.

PR #6061, opened 2023-04-17
Import sycl::ext::oneapi::experimental::printf into global namespace

masterleinad:
This allows getting rid of `KOKKOS_IMPL_DO_NOT_USE_PRINTF` (but this pull request keeps it for some backward-compatibility).
It remains to be discussed if importing into the global namespace is acceptable or not.

### DIFF
--- a/core/src/setup/Kokkos_Setup_SYCL.hpp
+++ b/core/src/setup/Kokkos_Setup_SYCL.hpp
@@ -41,4 +41,6 @@
   } while (0)
 #endif
 
+using sycl::ext::oneapi::experimental::printf;
+
 #endif
--- a/containers/src/Kokkos_DynRankView.hpp
+++ b/containers/src/Kokkos_DynRankView.hpp
@@ -221,7 +221,7 @@ KOKKOS_INLINE_FUNCTION bool dyn_rank_view_verify_operator_bounds(
     return (size_t(i) < map.extent(R)) &&
            dyn_rank_view_verify_operator_bounds<R + 1>(rank, map, args...);
   } else if (i != 0) {
-    KOKKOS_IMPL_DO_NOT_USE_PRINTF(
+    printf(
         "DynRankView Debug Bounds Checking Error: at rank %u\n  Extra "
         "arguments beyond the rank must be zero \n",
         R);
--- a/containers/src/Kokkos_UnorderedMap.hpp
+++ b/containers/src/Kokkos_UnorderedMap.hpp
@@ -568,7 +568,7 @@ class UnorderedMap {
           // Previously claimed an unused entry that was not inserted.
           // Release this unused entry immediately.
           if (!m_available_indexes.reset(new_index)) {
-            KOKKOS_IMPL_DO_NOT_USE_PRINTF("Unable to free existing\n");
+            printf("Unable to free existing\n");
           }
         }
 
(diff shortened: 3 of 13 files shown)


masterleinad:
> What happens if someone also does a `#include <cstdio>` (or equivalent)? Does it become ambiguous?

We should already include a header defining the global namespace `printf` in most translation units but that one is a C-style variadic function
```
int printf( const char* format, ... );
```
whereas the imported one has the signature
```
template <typename FormatT, typename... Args>
int printf(const FormatT *__format, Args... args)
```
and should be a better match since it doesn't contain an ellipsis.

crtrott (approved):

ldh4 (approved):

core/src/setup/Kokkos_Setup_SYCL.hpp
```diff
@@ -41,4 +41,6 @@
   } while (0)
 #endif
 
+using sycl::ext::oneapi::experimental::printf;
```
dalg24, line 44: I do not think it is acceptable to import into the global namespace.
Either we provide a `Kokkos::printf` or we do nothing.

nliber:
> > What happens if someone also does a `#include <cstdio>` (or equivalent)? Does it become ambiguous?
> 
> We should already include a header defining the global namespace `printf` in most translation units but that one is a C-style variadic function
> 
> ```
> int printf( const char* format, ... );
> ```
> 
> whereas the imported one has the signature
> 
> ```
> template <typename FormatT, typename... Args>
> int printf(const FormatT *__format, Args... args)
> ```
> 
> and should be a better match since it doesn't contain an ellipsis.

While not ambiguous,

```c++
printf("Hello\n");
```

will resolve to the C version, not the sycl version.  See [Compiler Explorer](https://godbolt.org/z/z7WE9Wano) for an example.

core/src/setup/Kokkos_Setup_SYCL.hpp
nliber: I agree with @dalg24, both because of `printf("Hello");` arguably calling the wrong function and in general we should not pollute the global namespace.

masterleinad: What does it matter if it calls the wrong function if does the right thing? 
I agree that we should avoid polluting the global namespace but this is as good as we can get it for the `SYCL` backend (unless we can convince the `SYCL` working group to support `printf` in the global namespace which appears to be very unlikely). From a user's perspective, this is a much better solution (for an experimental backend) than a `KOKKOS_IMPL_DO_NOT_USE_PRINTF` macro (since they don't need to change their code at all).

In the end, the question is if we tolerate polluting the global namespace for this backend rather than using `KOKKOS_IMPL_DO_NOT_USE_PRINTF` indefinitely or until we provide `Kokkos::printf` (which would require users to change their code once again).

dalg24: Consider this
```C++
// MyLib_DoWork.hpp
#pragma once
namespace MyLib {
KOKKOS_INLINE_FUNCTION void do_work() { /*...*/ printf("whatever"); }
}
```
Are you OK with that code having different meaning based on transitive header includes?

This opens the door to ODR violations (whether `<Kokkos_Core.hpp>` is included before that header change the definition of `MyLib::do_work`)

masterleinad: This file is included by `Kokkos_Macros.hpp`. As soon as you use `KOKKOS_INLINE_FUNCTION`, the imported function is visible.

dalg24: > This file is included by `Kokkos_Macros.hpp`

You are right, I missed that.

masterleinad:
Superseded by #6083.

PR #6063, opened 2023-04-18
Fix bug in Makefile.kokkos

stanmoore1:
Fix copy/paste bug in `Makefile.kokkos`. This caused the SPARTA build to hang in an infinite loop while generating dependencies. Found by @crtrott.


Rombur (approved):

cz4rs (approved):
Looks good to me.

masterleinad (approved):

PR #6064, opened 2023-04-18
SYCL: Improve and simplify parallel_scan implementation

masterleinad:
This pull request changes the previous recursive `parallel_scan` in the `SYCL` backend to a two-pass one as we use for `Cuda` and `HIP` and `SYCL` reductions. This simplifies the code (also making it more uniform) and reduces the memory footprint (since we only need to store intermediate results for all items and group scans but not recursive group scans).
On the way, I made sure that all local operations operate on indices of type `int` avoiding 64-bit index operations. 

A second improvement is switching to an auto-detection of the work group size as we do for reductions by querying a dummy kernel for the maximum group size.

Finally, this fixes a couple of unit tests that were failing with `SYCL+Cuda` since https://github.com/kokkos/kokkos/pull/5707.

### DIFF
--- a/core/src/SYCL/Kokkos_SYCL_Instance.cpp
+++ b/core/src/SYCL/Kokkos_SYCL_Instance.cpp
@@ -280,11 +280,9 @@ sycl::device_ptr<void> SYCLInternal::scratch_flags(const std::size_t size) {
 
     m_scratchFlags = reinterpret_cast<size_type*>(r->data());
   }
-  m_queue->memset(m_scratchFlags, 0, m_scratchFlagsCount * sizeScratchGrain);
-  fence(*m_queue,
-        "Kokkos::Experimental::SYCLInternal::scratch_flags fence after "
-        "initializing m_scratchFlags",
-        m_instance_id);
+  auto memset_event = m_queue->memset(m_scratchFlags, 0,
+                                      m_scratchFlagsCount * sizeScratchGrain);
+  m_queue->ext_oneapi_submit_barrier(std::vector{memset_event});
 
   return m_scratchFlags;
 }
--- a/core/src/SYCL/Kokkos_SYCL_Parallel_Scan.hpp
+++ b/core/src/SYCL/Kokkos_SYCL_Parallel_Scan.hpp
@@ -124,7 +124,7 @@ class ParallelScanSYCLBase {
  private:
   template <typename FunctorWrapper>
   sycl::event sycl_direct_launch(const FunctorWrapper& functor_wrapper,
-                                 sycl::event memcpy_event) const {
+                                 sycl::event memcpy_event) {
     // Convenience references
     const Kokkos::Experimental::SYCL& space = m_policy.space();
     Kokkos::Experimental::Impl::SYCLInternal& instance =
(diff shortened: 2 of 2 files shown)


algorithms/src/std_algorithms/impl/Kokkos_ExclusiveScan.hpp
```diff
@@ -96,6 +96,8 @@ struct ExclusiveScanDefaultFunctor {
 
   KOKKOS_FUNCTION
   void join(value_type& update, const value_type& input) const {
+    if (input.is_initial) return;
+
```
masterleinad, line 100: @fnrizzi With a simplified `parallel_scan` implementation I noticed that
```
reducer.init(value2);
reducer.join(value1, value2);
```
would give wrong results (`value1` shouldn't change but does) as opposed to
```
reducer.init(value2);
reducer.join(value2, value1);
```
where we get `value2==value1` as expected (assuming `value1` doesn't have initial value).

fnrizzi: I am not sure i am following what is going on here... did you try to create a reducer and call init and joint as above? 

Btw, why for `reducer.join(value1, value2);` you say that: "would give wrong results (value1 shouldn't change but does". the first parameter in join is what changes , so value1 changes

masterleinad: Yes, I was running into a situation where `update` had a non-initial value and `input` had an initial value (representing the neutral element). This gave me the wrong results when testing an inclusive scan with a multiplicative functor since the original value was multiplied by zero.

masterleinad:
Requires #6065.

core/src/SYCL/Kokkos_SYCL_Instance.cpp
```diff
@@ -273,11 +273,9 @@ sycl::device_ptr<void> SYCLInternal::scratch_flags(const std::size_t size) {
...
     m_scratchFlags = reinterpret_cast<size_type*>(r->data());
   }
-  m_queue->memset(m_scratchFlags, 0, m_scratchFlagsCount * sizeScratchGrain);
-  fence(*m_queue,
-        "Kokkos::Experimental::SYCLInternal::scratch_flags fence after "
-        "initializing m_scratchFlags",
-        m_instance_id);
+  auto memset_event = m_queue->memset(m_scratchFlags, 0,
+                                      m_scratchFlagsCount * sizeScratchGrain);
+  m_queue->ext_oneapi_submit_barrier(std::vector{memset_event});
```
dalg24, line 278: This change seems unrelated.
Also why `sycl::queue::ext_oneapi_submit_barrier()` instead of regular `sycl::queue::submit_barrier`?

masterleinad: I need this change for the tests checking fences. 
`sycl::queue::submit_barrier` doesn't exist (anymore).

dalg24: > I need this change for the tests checking fences.

I do not understand which tests you are referring to.

masterleinad: https://github.com/kokkos/kokkos/blob/develop/core/unit_test/tools/TestEventCorrectness.hpp#L391-L473

core/src/SYCL/Kokkos_SYCL_Parallel_Scan.hpp
```diff
@@ -32,21 +32,22 @@ namespace Impl {
...
   }
 
-  const auto max_subgroup_size = sg.get_max_local_range()[0];
-  const auto n_active_subgroups =
+  const int max_subgroup_size = sg.get_max_local_range()[0];
+  const int n_active_subgroups =
       (global_range + max_subgroup_size - 1) / max_subgroup_size;
 
-  const auto local_range = sg.get_local_range()[0];
+  const int local_range = sg.get_local_range()[0];
```
crtrott, line 50: did SYCL change so that you need int now?

crtrott (approved):
AS far as I can tell looks good.

core/src/SYCL/Kokkos_SYCL_Parallel_Scan.hpp
masterleinad: No, it's not necessary to do that but I have seen that SYCL is very sensitive to 64-bit index calculations, and local indices surely will never require that.

nliber: Just to be clear, `auto` deduces `size_t`, not `int`.  I agree we don't need that much of an index range.

nliber (approved):

dalg24 (commented):
I do not like the complexity added with the "lambda factory" but I will merge since others seems that it is fine.

dalg24:
The SYCL build passed.  Ignoring the rest.

PR #6065, opened 2023-04-18
Fix join for ValueWrapperForNoNeutralElement

masterleinad:
While working in #6064, I noticed that the `join` implementations don't treat ` ValueWrapperForNoNeutralElement` as neutral element.
With a simplified `parallel_scan` implementation I noticed that
```
reducer.init(value2);
reducer.join(value1, value2);
```
would give wrong results (`value1` shouldn't change but does) as opposed to
```
reducer.init(value2);
reducer.join(value2, value1);
```
where we get `value2==value1` as expected (assuming `value1` doesn't have initial value).

### DIFF
--- a/algorithms/src/std_algorithms/impl/Kokkos_ExclusiveScan.hpp
+++ b/algorithms/src/std_algorithms/impl/Kokkos_ExclusiveScan.hpp
@@ -96,6 +96,8 @@ struct ExclusiveScanDefaultFunctor {
 
   KOKKOS_FUNCTION
   void join(value_type& update, const value_type& input) const {
+    if (input.is_initial) return;
+
     if (update.is_initial) {
       update.val        = input.val;
       update.is_initial = false;
--- a/algorithms/src/std_algorithms/impl/Kokkos_InclusiveScan.hpp
+++ b/algorithms/src/std_algorithms/impl/Kokkos_InclusiveScan.hpp
@@ -90,6 +90,8 @@ struct InclusiveScanDefaultFunctor {
 
   KOKKOS_FUNCTION
   void join(value_type& update, const value_type& input) const {
+    if (input.is_initial) return;
+
     if (update.is_initial) {
       update.val = input.val;
     } else {
--- a/algorithms/src/std_algorithms/impl/Kokkos_TransformExclusiveScan.hpp
+++ b/algorithms/src/std_algorithms/impl/Kokkos_TransformExclusiveScan.hpp
@@ -76,6 +76,8 @@ struct TransformExclusiveScanFunctor {
 
   KOKKOS_FUNCTION
   void join(value_type& update, const value_type& input) const {
+    if (input.is_initial) return;
+
     if (update.is_initial) {
       update.val = input.val;
     } else {
--- a/algorithms/src/std_algorithms/impl/Kokkos_TransformInclusiveScan.hpp
+++ b/algorithms/src/std_algorithms/impl/Kokkos_TransformInclusiveScan.hpp
@@ -67,6 +67,8 @@ struct TransformInclusiveScanNoInitValueFunctor {
 
   KOKKOS_FUNCTION
   void join(value_type& update, const value_type& input) const {
+    if (input.is_initial) return;
+
     if (update.is_initial) {
       update.val = input.val;
     } else {
@@ -118,6 +120,8 @@ struct TransformInclusiveScanWithInitValueFunctor {
 
   KOKKOS_FUNCTION
   void join(value_type& update, const value_type& input) const {
+    if (input.is_initial) return;
+
     if (update.is_initial) {
       update.val = input.val;
     } else {
--- a/algorithms/unit_tests/TestStdAlgorithmsExclusiveScan.cpp
+++ b/algorithms/unit_tests/TestStdAlgorithmsExclusiveScan.cpp
@@ -344,6 +344,46 @@ TEST(std_algorithms_numeric_ops_test, exclusive_scan) {
   run_exclusive_scan_all_scenarios<StridedThreeTag, CustomValueType>();
 }
 
+TEST(std_algorithms_numeric_ops_test, exclusive_scan_functor) {
+  int dummy       = 0;
+  using view_type = Kokkos::View<int*, exespace>;
+  view_type dummy_view("dummy_view", 0);
+  using functor_type = Kokkos::Experimental::Impl::ExclusiveScanDefaultFunctor<
+      exespace, int, int, view_type, view_type>;
+  functor_type functor(dummy, dummy_view, dummy_view);
+  using value_type = functor_type::value_type;
+
+  value_type value1;
+  functor.init(value1);
+  EXPECT_EQ(value1.val, 0);
+  EXPECT_EQ(value1.is_initial, true);
+
+  value_type value2;
+  value2.val        = 1;
+  value2.is_initial = false;
+  functor.join(value1, value2);
+  EXPECT_EQ(value1.val, 1);
+  EXPECT_EQ(value1.is_initial, false);
+
+  functor.init(value1);
+  functor.join(value2, value1);
+  EXPECT_EQ(value2.val, 1);
+  EXPECT_EQ(value2.is_initial, false);
+
+  functor.init(value2);
+  functor.join(value2, value1);
+  EXPECT_EQ(value2.val, 0);
+  EXPECT_EQ(value2.is_initial, true);
+
+  value1.val        = 1;
+  value1.is_initial = false;
+  value2.val        = 2;
+  value2.is_initial = false;
+  functor.join(value2, value1);
+  EXPECT_EQ(value2.val, 3);
+  EXPECT_EQ(value2.is_initial, false);
+}
+
 }  // namespace EScan
 }  // namespace stdalgos
 }  // namespace Test
--- a/algorithms/unit_tests/TestStdAlgorithmsInclusiveScan.cpp
+++ b/algorithms/unit_tests/TestStdAlgorithmsInclusiveScan.cpp
@@ -353,6 +353,45 @@ TEST(std_algorithms_numeric_ops_test, inclusive_scan) {
   run_inclusive_scan_all_scenarios<StridedThreeTag, CustomValueType>();
 }
 
+TEST(std_algorithms_numeric_ops_test, inclusive_scan_functor) {
+  using view_type = Kokkos::View<int*, exespace>;
+  view_type dummy_view("dummy_view", 0);
+  using functor_type = Kokkos::Experimental::Impl::InclusiveScanDefaultFunctor<
+      exespace, int, int, view_type, view_type>;
+  functor_type functor(dummy_view, dummy_view);
+  using value_type = functor_type::value_type;
+
+  value_type value1;
+  functor.init(value1);
+  EXPECT_EQ(value1.val, 0);
+  EXPECT_EQ(value1.is_initial, true);
+
+  value_type value2;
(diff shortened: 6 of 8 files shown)


fnrizzi:
@masterleinad i will take a look today or tomorrow!

crtrott:
I do not quite see how the test would have different results based on whether you make the change to the join or not?

masterleinad:
> I do not quite see how the test would have different results based on whether you make the change to the join or not?

Without the changes, you get
```
[...]
[----------] 8 tests from std_algorithms_numeric_ops_test
[ RUN      ] std_algorithms_numeric_ops_test.exclusive_scan
[       OK ] std_algorithms_numeric_ops_test.exclusive_scan (436 ms)
[ RUN      ] std_algorithms_numeric_ops_test.exclusive_scan_functor
/tmp/kokkos/algorithms/unit_tests/TestStdAlgorithmsExclusiveScan.cpp:376: Failure
  value2.is_initial
    Which is: false
  true
[  FAILED  ] std_algorithms_numeric_ops_test.exclusive_scan_functor (0 ms)
[ RUN      ] std_algorithms_numeric_ops_test.inclusive_scan
[       OK ] std_algorithms_numeric_ops_test.inclusive_scan (375 ms)
[ RUN      ] std_algorithms_numeric_ops_test.inclusive_scan_functor
/tmp/kokkos/algorithms/unit_tests/TestStdAlgorithmsInclusiveScan.cpp:384: Failure
  value2.is_initial
    Which is: false
  true
[  FAILED  ] std_algorithms_numeric_ops_test.inclusive_scan_functor (0 ms)
[ RUN      ] std_algorithms_numeric_ops_test.transform_exclusive_scan
[       OK ] std_algorithms_numeric_ops_test.transform_exclusive_scan (117 ms)
[ RUN      ] std_algorithms_numeric_ops_test.transform_exclusive_scan_functor
/tmp/kokkos/algorithms/unit_tests/TestStdAlgorithmsTransformExclusiveScan.cpp:318: Failure
  value2.val
    Which is: 0
  1
/tmp/kokkos/algorithms/unit_tests/TestStdAlgorithmsTransformExclusiveScan.cpp:324: Failure
  value2.is_initial
    Which is: false
  true
[  FAILED  ] std_algorithms_numeric_ops_test.transform_exclusive_scan_functor (0 ms)
[ RUN      ] std_algorithms_numeric_ops_test.transform_inclusive_scan
[       OK ] std_algorithms_numeric_ops_test.transform_inclusive_scan (38 ms)
[ RUN      ] std_algorithms_numeric_ops_test.transform_inclusive_scan_functor
/tmp/kokkos/algorithms/unit_tests/TestStdAlgorithmsTransformInclusiveScan.cpp:335: Failure
  value2.val
    Which is: 0
  1
/tmp/kokkos/algorithms/unit_tests/TestStdAlgorithmsTransformInclusiveScan.cpp:341: Failure
  value2.is_initial
    Which is: false
  true
/tmp/kokkos/algorithms/unit_tests/TestStdAlgorithmsTransformInclusiveScan.cpp:335: Failure
  value2.val
    Which is: 0
  1
/tmp/kokkos/algorithms/unit_tests/TestStdAlgorithmsTransformInclusiveScan.cpp:341: Failure
  value2.is_initial
    Which is: false
  true
[  FAILED  ] std_algorithms_numeric_ops_test.transform_inclusive_scan_functor (0 ms)
[----------] 8 tests from std_algorithms_numeric_ops_test (967 ms total)

[----------] 1 test from std_algorithms_transform_ops_test
[ RUN      ] std_algorithms_transform_ops_test.transform_unary_op
[       OK ] std_algorithms_transform_ops_test.transform_unary_op (18 ms)
[----------] 1 test from std_algorithms_transform_ops_test (18 ms total)

[----------] Global test environment tear-down
[==========] 29 tests from 7 test suites ran. (1111 ms total)
[  PASSED  ] 25 tests.
[  FAILED  ] 4 tests, listed below:
[  FAILED  ] std_algorithms_numeric_ops_test.exclusive_scan_functor
[  FAILED  ] std_algorithms_numeric_ops_test.inclusive_scan_functor
[  FAILED  ] std_algorithms_numeric_ops_test.transform_exclusive_scan_functor
[  FAILED  ] std_algorithms_numeric_ops_test.transform_inclusive_scan_functor
```
since we overwrite the current value when the value to be joined with is initial (which means that it should be neutral element).

fnrizzi:
1) what did you do in here "With a simplified parallel_scan implementation" ? 
2) why do you say `value1` shouldn't change below? 
```
reducer.init(value2);
reducer.join(value1, value2);
```

masterleinad:
@fnrizzi I basically needed to merge two variables and it happened that one of them had initial value. 

After merging I expect that the result has the value of the non-initial variable.
Say I have `value_initial` and `value_no_initial` then  I expect
```
result = value_initial;
reducer.join(result, value_no_initial);
```
do store `value_no_initial` in `result` and
```
result = value_no_initial;
reducer.join(result, value_initial)
```
to do nothing so that `result` still has the value of `value_no_initial` and the value is the same for both cases.

fnrizzi:
> @fnrizzi I basically needed to merge two variables and it happened that one of them had initial value.
> 
> After merging I expect that the result has the value of the non-initial variable. Say I have `value_initial` and `value_no_initial` then I expect
> 
> ```
> result = value_initial;
> reducer.join(result, value_no_initial);
> ```
> 
> do store `value_no_initial` in `result` and
> 
> ```
> result = value_no_initial;
> reducer.join(result, value_initial)
> ```
> 
> to do nothing so that `result` still has the value of `value_no_initial` and the value is the same for both cases.

isn't this code (as it was before you PR) doing exactly what you say above? 
```
  void join(value_type& update, const value_type& input) const {
    if (update.is_initial) {
      update.val        = input.val;
      update.is_initial = false;
    } else {
      update.val = update.val + input.val;
    }
  }
};
```

fnrizzi:
> @fnrizzi I basically needed to merge two variables and it happened that one of them had initial value.
> 
> After merging I expect that the result has the value of the non-initial variable. Say I have `value_initial` and `value_no_initial` then I expect
> 
> ```
> result = value_initial;
> reducer.join(result, value_no_initial);
> ```
> 
> do store `value_no_initial` in `result` and
> 
> ```
> result = value_no_initial;
> reducer.join(result, value_initial)
> ```
> 
> to do nothing so that `result` still has the value of `value_no_initial` and the value is the same for both cases.

Ok, so I agree on the first thing. 
I don't understand why you expect the second case not to do anything? I think the second one should be default add `value_initial` to `result`

fnrizzi:
as i see it, any argument passed to the second parameter `b` of `join(a,b)` should never have a value_initial to be true

masterleinad:
#6064 fails in the relevant tests without this.

fnrizzi:
I see your point and explanation, let me think about it for a second.

masterleinad:
> ok, I think what you say makes sense. To answer your question about why I said that the second argument should never be allowed to have `is_inital==true`: that is because the init method in the reducer is only called to initialize the reduction result (whether it is a partial one or not), and so in my mind, the second argument was always going to be something contributing to it but never an initial value. does my explanation make sense?

Yes, it does, but I think it's to restrictive and makes it more difficult for implementations using the reducer. 🙂

> I think i am ok with the change, but was wondering: do all other reducers (not the ones using the valueWrapper but all those the preexisted) satisfy your argument about the monoid?

I would rather phrase it as a requirement. All our built-in reducers fulfill this property and user-defined reducers should too.

fnrizzi (approved):

fnrizzi:
> I would rather phrase it as a requirement. 

agree

masterleinad:
> agree! I think we should write this in the wiki more explicitly. I don't there is a detail discussion about the join. But it would be useful to highlight all of this. don't you think?

See https://github.com/kokkos/kokkos-core-wiki/pull/363.

Rombur:
Do I understand correctly that this PR changes the meaning of `init`? Right now it initializes the value to whatever the user decides but with this PR, it has to be the neutral element. So for example right now if I want to define my `Addition` reducer, I can decide that `init` will initialize the value to 2 but the neutral element is still zero. With this PR, both 0 and 2 would become neutral element. Is that correct?

masterleinad:
> Do I understand correctly that this PR changes the meaning of `init`? Right now it initializes the value to whatever the user decides but with this PR, it has to be the neutral element. So for example right now if I want to define my `Addition` reducer, I can decide that `init` will initialize the value to 2 but the neutral element is still zero. With this PR, both 0 and 2 would become neutral element. Is that correct?

No, `init` was never supposed to describe an initial value (to the best of my understanding). It's a value that every thread can be initialized with before merging with other values. That only makes sense (producing results independent of the number of threads and backend implementation) if that value is a neutral element for the operation.

The reducer here does something unusual (compared with built-in reducers) in that it finds a way to define a reduction without knowing about the neutral element of the operation (at the cost of increasing the size of the value type). The value it stores for `is_inital==true` is never supposed to be used.

Rombur:
> No, init was never supposed to describe an initial value (to the best of my understanding).

That's not what I mean. I mean to initialize the value in a valid state.  I know that the value will be overwritten. To me the intent of the current code is to use `init` to create a variable in a valid state and then to overwrite it during the reduce operation. The binary operator still determines the neutral element. The goal of `init` is just to make sure that we have a valid `value`. However with this PR, you try to detect the neutral element and so you need to enforce that `init` initializes the value to the neutral element.

I am not against this PR but it does impact user code in ways that can be surprising by adding a new neutral element.

masterleinad:
> That's not what I mean. I mean to initialize the value in a valid state. I know that the value will be overwritten. To me the intent of the current code is to use init to create a variable in a valid state and then to overwrite it during the reduce operation.

Yes, exactly.

> The binary operator still determines the neutral element. The goal of init is just to make sure that we have a valid value. However with this PR, you try to detect the neutral element and so you need to enforce that init initializes the value to the neutral element.

For an arbitrary binary operator, there is no way to determine its neutral element (but our reducer concept requires one) and that's why the implementation here defines a reducer that doesn't need to know about it but just provides a workaround that acts as one. This pull request doesn't mess with the value when `is_initial==true` (and that value didn't matter before and doesn't matter with this pull request).

> I am not against this PR but it does impact user code in ways that can be surprising by adding a new neutral element.

I don't see how this pull changes anything for users but fixes a bug IMHO. It only touches on the case that the second argument to `join` has been constructed by calling `init`. What other behavior would be valid in this case apart from keeping the current value stored in the first argument?

Rombur:
I need to think more about this. TBH the current behavior makes sense to me.

Rombur (approved):
I think this PR is OK. To me the current code is correct but it's true that the documentation somewhat implies that after calling init the element is neutral.

fnrizzi:
> I think this PR is OK. To me the current code is correct but it's true that the documentation somewhat implies that after calling init the element is neutral.

I agree the current code is  fine, and like Daniel said, if we were to always put the accumulating variable on the left it would still work but makes it a bit harder to use.

masterleinad:
Only `KokkosContainers_UnitTest_OpenMP` is failing in `gcc-8.4.0`.

dalg24:
Test this please

dalg24:
Retest this please (not sure what is going on I can see that the packets were sent to the CI server but somehow it did not trigger the build)

dalg24:
Retest this please

dalg24:
Retest this please

dalg24:
Retest this please

dalg24:
Retest this please

dalg24:
Retest this please

dalg24:
Retest this please

PR #6074, opened 2023-04-21
SYCL: Use in-order queue for SYCL+Cuda

masterleinad:
Apparently, the cause for all the test failures when requiring `oneAPI 2023.0.0` (https://github.com/kokkos/kokkos/pull/5707) was that `submit_barrier` is not quite working as intended for `SYCL+Cuda`.
Reverts #6056.

### DIFF
--- a/core/perf_test/PerfTestGramSchmidt.cpp
+++ b/core/perf_test/PerfTestGramSchmidt.cpp
@@ -175,6 +175,16 @@ static void GramSchmidt(benchmark::State& state) {
   }
 }
 
+// FIXME_SYCL SYCL+Cuda reports "an illegal memory access was encountered"
+#if defined(KOKKOS_ENABLE_SYCL) && defined(KOKKOS_IMPL_ARCH_NVIDIA_GPU)
+BENCHMARK(GramSchmidt<double>)
+    ->ArgName("Count")
+    ->ArgsProduct({
+        benchmark::CreateRange(1 << 10, 1 << 18, 2),
+    })
+    ->UseManualTime()
+    ->Iterations(5);
+#else
 BENCHMARK(GramSchmidt<double>)
     ->ArgName("Count")
     ->ArgsProduct({
@@ -182,5 +192,6 @@ BENCHMARK(GramSchmidt<double>)
     })
     ->UseManualTime()
     ->Iterations(5);
+#endif
 
 }  // namespace Test
--- a/core/src/SYCL/Kokkos_SYCL_Instance.cpp
+++ b/core/src/SYCL/Kokkos_SYCL_Instance.cpp
@@ -75,7 +75,15 @@ void SYCLInternal::initialize(const sycl::device& d) {
       Kokkos::Impl::throw_runtime_exception(
           "There was an asynchronous SYCL error!\n");
   };
-  initialize(sycl::queue{d, exception_handler});
+  // FIXME SYCL+Cuda Apparently, submit_barrier doesn't quite work as expected
+  // for oneAPI 2023.0.0 on NVIDIA GPUs.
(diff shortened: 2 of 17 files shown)


core/src/SYCL/Kokkos_SYCL_Instance.cpp
```diff
@@ -75,7 +75,15 @@ void SYCLInternal::initialize(const sycl::device& d) {
       Kokkos::Impl::throw_runtime_exception(
           "There was an asynchronous SYCL error!\n");
   };
-  initialize(sycl::queue{d, exception_handler});
+  // FIXME SYCL+Cuda Apparently, submit_barrier doesn't quite work as expected
+  // for oneAPI 2023.0.0 on NVIDIA GPUs.
+#if defined(KOKKOS_IMPL_ARCH_NVIDIA_GPU)
+  initialize(
+      sycl::queue{d, exception_handler, sycl::property::queue::in_order()});
```
dalg24, line 82: I thought that was the default

dalg24 (commented):
Please add references to previous related work

core/src/SYCL/Kokkos_SYCL_Instance.cpp
dalg24: Is the "in order" property not the default?
If the answer is "no it depends on the device", then is that not something that we should then be specifying unconditionally?

masterleinad: No, by default `sycl::queus` are out-of-order and I wanted to support that for flexibility in the implementation, e.g., when it comes to our `Graph` interface (which doesn't exploit out-of-order `queues`, yet).

core/perf_test/PerfTestGramSchmidt.cpp
```diff
@@ -175,12 +175,23 @@ static void GramSchmidt(benchmark::State& state) {
   }
 }
 
+// FIXME_SYCL SYCL+Cuda reports "an illegal memory access was encountered"
```
dalg24, line 178: How is that related to using an in-order queue?

core/src/SYCL/Kokkos_SYCL_Instance.cpp
```diff
@@ -75,7 +75,15 @@ void SYCLInternal::initialize(const sycl::device& d) {
       Kokkos::Impl::throw_runtime_exception(
           "There was an asynchronous SYCL error!\n");
   };
-  initialize(sycl::queue{d, exception_handler});
+  // FIXME SYCL+Cuda Apparently, submit_barrier doesn't quite work as expected
+  // for oneAPI 2023.0.0 on NVIDIA GPUs.
```
dalg24, line 79: Was this reported?  How do we keep track on progress on this?

masterleinad: I don't really have a good reproducer (yet). I just found in #6064 that replacing a `submit_barrier` with a `depends_on` fixed issues I was seeing only for `SYCL+Cuda`. I can try to come up with something to report to `intel/llvm`.

core/perf_test/PerfTestGramSchmidt.cpp
masterleinad: I don't quite know yet. It works if put the additional fence before `memcpy` in again but that is the only test that seems affected and I'm leaning towards modifying the test rather than putting the `fence` in which might hide other problems in the future.

dalg24 (approved):
I don't like the Gram-Schmidt benchmark change and do not fully understand how using in-order queue helped but I will approve on the ground that less tests are being skipped.

nliber (approved):

PR #6077, opened 2023-04-24
Update OpenACC FunctorAdapter

seyonglee:
Update OpenACC FunctorAdapter to be able to handle functors containing parallel loops.

In OpenACC, any device function called in a compute region should be annotated with an `acc routine` directive unless it does not have any OpenACC parallel loop, in which case an OpenACC compiler implicitly adds an `acc routine seq` directive. 
The current Kokkos OpenACC backend has two implementations for Kokkos parallel constructs with Team Policy: a version where the hierarchical parallel constructs with the Team Policy are logically flattened using a manual collapsing, and the other version keeps the hierarchical parallel constructs, each of which contains different OpenACC parallel loops.
This PR is to support the second implementation (non-collapsed version), where some Kokkos parallel construct can call a nested lambda containing an OpenACC parallel loop.
If the Kokkos OpenACC backend is compiled by NVHPC OpenACC compiler, the collapsed version is used and thus we didn't have to add `acc routine` directives. However, Clacc/LLVM OpenACC compiler uses the non-collapsed version; this PR is for the Clacc OpenACC compiler. (The non-collapsed version fully exploits hierarchical parallelism available in OpenACC, while the collapsed version does not.)

### DIFF
--- a/core/src/OpenACC/Kokkos_OpenACC_FunctorAdapter.hpp
+++ b/core/src/OpenACC/Kokkos_OpenACC_FunctorAdapter.hpp
@@ -19,9 +19,12 @@
 
 #include <type_traits>
 
+#define KOKKOS_OPENACC_CONTAIN_SEQLOOP 0
+#define KOKKOS_OPENACC_CONTAIN_WORKERLOOP 1
+
 namespace Kokkos::Experimental::Impl {
 
-template <class Functor, class Policy>
+template <class Functor, class Policy, int N, typename = void>
 class FunctorAdapter {
   Functor m_functor;
   using WorkTag = typename Policy::work_tag;
@@ -29,6 +32,28 @@ class FunctorAdapter {
  public:
   FunctorAdapter(Functor const &functor) : m_functor(functor) {}
 
+#pragma acc routine seq
+  template <class... Args>
+  KOKKOS_FUNCTION void operator()(Args &&... args) const {
+    if constexpr (std::is_void_v<WorkTag>) {
+      m_functor(static_cast<Args &&>(args)...);
+    } else {
+      m_functor(WorkTag(), static_cast<Args &&>(args)...);
+    }
+  }
+};
+
+template <class Functor, class Policy, int N>
+class FunctorAdapter<
+    Functor, Policy, N,
+    typename std::enable_if<N == KOKKOS_OPENACC_CONTAIN_WORKERLOOP>::type> {
+  Functor m_functor;
+  using WorkTag = typename Policy::work_tag;
+
+ public:
+  FunctorAdapter(Functor const &functor) : m_functor(functor) {}
+
+#pragma acc routine worker
   template <class... Args>
   KOKKOS_FUNCTION void operator()(Args &&... args) const {
     if constexpr (std::is_void_v<WorkTag>) {
--- a/core/src/OpenACC/Kokkos_OpenACC_FunctorAdapter.hpp
+++ b/core/src/OpenACC/Kokkos_OpenACC_FunctorAdapter.hpp
@@ -17,52 +17,40 @@
 #ifndef KOKKOS_OPENACC_FUNCTOR_ADAPTER_HPP
 #define KOKKOS_OPENACC_FUNCTOR_ADAPTER_HPP
 
+#include <OpenACC/Kokkos_OpenACC_Macros.hpp>
 #include <type_traits>
-
-#define KOKKOS_OPENACC_CONTAIN_SEQLOOP 0
-#define KOKKOS_OPENACC_CONTAIN_WORKERLOOP 1
 
 namespace Kokkos::Experimental::Impl {
 
-template <class Functor, class Policy, int N, typename = void>
-class FunctorAdapter {
-  Functor m_functor;
-  using WorkTag = typename Policy::work_tag;
+enum class RoutineClause { worker, seq };
 
- public:
-  FunctorAdapter(Functor const &functor) : m_functor(functor) {}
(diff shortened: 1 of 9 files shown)


core/src/OpenACC/Kokkos_OpenACC_FunctorAdapter.hpp
```diff
@@ -19,16 +19,41 @@
 
 #include <type_traits>
 
+#define KOKKOS_OPENACC_CONTAIN_SEQLOOP 0
+#define KOKKOS_OPENACC_CONTAIN_WORKERLOOP 1
+
 namespace Kokkos::Experimental::Impl {
 
-template <class Functor, class Policy>
+template <class Functor, class Policy, int N, typename = void>
```
dalg24, line 27: There is no need for these` KOKKOS_OPENACC_CONTAIN_{SEQ,WORKER}LOOP` macros.
Please prefer a strong enumeration (or tag classes).  Something like
```C++
enum class RoutineClause { worker, seq };

template <class Functor, class Policy, RoutineClause>
class FunctorAdapter;

#define KOKKOS_IMPL_ACC_FUNCTOR_ADAPTER(CLAUSE)                    \
  template <class Functor, class Policy>                           \
  class FunctorAdapter<Functor, Policy, RoutineClause::CLAUSE> {   \
    Functor m_functor;                                             \
    using WorkTag = typename Policy::work_tag;                     \
                                                                   \
   public:                                                         \
    FunctorAdapter(Functor const &functor) : m_functor(functor) {} \
    KOKKOS_IMPL_ACC_PRAGMA_HELPER(routine CLAUSE)                  \
    template <class... Args>                                       \
    KOKKOS_FUNCTION void operator()(Args &&...args) const {        \
      if constexpr (std::is_void_v<WorkTag>) {                     \
        m_functor(static_cast<Args &&>(args)...);                  \
      } else {                                                     \
        m_functor(WorkTag(), static_cast<Args &&>(args)...);       \
      }                                                            \
    }                                                              \
  }

KOKKOS_IMPL_ACC_FUNCTOR_ADAPTER(worker);
KOKKOS_IMPL_ACC_FUNCTOR_ADAPTER(seq);

#undef KOKKOS_IMPL_ACC_FUNCTOR_ADAPTER

File: core/unit_test/CMakeLists.txt [hunk 2]
```diff
@@ -546,6 +546,7 @@ IF(KOKKOS_ENABLE_OPENACC AND KOKKOS_CXX_COMPILER_ID STREQUAL NVHPC)
     ${CMAKE_CURRENT_BINARY_DIR}/openacc/TestOpenACC_ViewAPI_f.cpp
     ${CMAKE_CURRENT_BINARY_DIR}/openacc/TestOpenACC_ViewMapping_b.cpp
     ${CMAKE_CURRENT_BINARY_DIR}/openacc/TestOpenACC_ViewResize.cpp
+    ${CMAKE_CURRENT_BINARY_DIR}/openacc/TestOpenACC_BitManipulationBuiltins.cpp
```
dalg24, line 549: How is that change related?

seyonglee: Unrelated to this PR; just added it because NVHPC fails.

core/src/OpenACC/Kokkos_OpenACC_FunctorAdapter.hpp
seyonglee: Updated as suggested.

core/src/OpenACC/Kokkos_OpenACC_ParallelReduce_Team.hpp
```diff
@@ -21,6 +21,14 @@
 #include <OpenACC/Kokkos_OpenACC_FunctorAdapter.hpp>
 #include <OpenACC/Kokkos_OpenACC_Macros.hpp>
 
+#ifdef KOKKOS_ENABLE_OPENACC_COLLAPSE_HIERARCHICAL_CONSTRUCTS
+#define KOKKOS_OPENACC_LOOP_CLAUSE \
```
dalg24, line 25:
```suggestion
#define KOKKOS_IMPL_OPENACC_LOOP_CLAUSE \
```

core/unit_test/CMakeLists.txt
dalg24: Undo and propose in another PR as necessary

dalg24 (commented):
Please elaborate a bit in the PR description.
It is not clear to me why this is needed.  Will user-defined functions needed the acc routine clause as well in the end?

seyonglee:
> Please elaborate a bit in the PR description.
> It is not clear to me why this is needed. Will user-defined functions needed the acc routine clause as well in the end?

Updated the PR description.

core/unit_test/CMakeLists.txt
seyonglee: Removed; now the OpenACC unit test will fail.

core/src/OpenACC/Kokkos_OpenACC_ParallelReduce_Team.hpp
seyonglee: Applied.

core/unit_test/CMakeLists.txt
seyonglee: This failed unit test problem is handled by issue #6088.

dalg24:
Retest this please

dalg24 (approved):

masterleinad (approved):
Looks good to me.

PR #6081, opened 2023-04-26
sorting: add to binsort support for strided views and reorg tests

fnrizzi:
This PR has two parts:

## 1: fixes *bin sort* to allow sorting views with layout stride 

Before this PR, using bin sort for sorting a view with layout stride was giving a compile error because the scratch view used inside the bin sort implementation was declared to use same layout as the values view to sort, but was then constructed from just extents. This is fixed now. 

## 2: reorganizes all sort- and random-related tests 

The tests related to sort were hard to follow. For example, tests for sort, bin sort and random were all lumped together, 
and this was compounded by a few layers to get from cmake to actually see the tests (there was a SortCommon which calls something else which calls something else). Also, tests for OpenMP were hardwired as separate files, but the same tests were also created via cmake via another file.
This PR reorganizes this as follows:
  - tests for random, sort, binsort are split into separate files as follows:
    - `TestBinSortA.hpp`: contains all tests already existing for BinSort 
    - `TestBinSortB.hpp`: contains the new tests added with this PR to verify the sorting for strided view
    - `TestNestedSort.hpp`: contains existing tests for nested sort
    - `TestRandom.hpp`: contains existing tests for random generation stuff

  - executables are split into:
     - `KokkosAlgorithms_UnitTest_Random`: contains all tests for random generation things 
     - `KokkosAlgorithms_UnitTest_Sort`: all tests related to sorting


fnrizzi:
retest this please

fnrizzi:
open for reviews (i know nvhpc fails, will work on it)

fnrizzi:
unrelated HPX CI fails for `1 - Kokkos_CoreUnitTest_HPX (Timeout)`

algorithms/src/Kokkos_Sort.hpp
```diff
@@ -146,8 +146,12 @@ class BinSort {
         Kokkos::is_view<SrcViewType>::value,
         Kokkos::View<typename SrcViewType::const_data_type,
                      typename SrcViewType::array_layout,
-                     typename SrcViewType::device_type,
-                     Kokkos::MemoryTraits<Kokkos::RandomAccess> >,
+                     typename SrcViewType::device_type
+#if not defined(KOKKOS_COMPILER_NVHPC)  // FIXME_NVHPC
```
masterleinad, line 150: We typically don't use `not` in the code base.
```suggestion
#if !defined(KOKKOS_COMPILER_NVHPC)  // FIXME_NVHPC
```

Is that a compile-time error with NVHPC?

algorithms/unit_tests/TestBinSortA.hpp
```diff
@@ -0,0 +1,244 @@
...
+}
+
+template <class ExecutionSpace, class T>
+void test_sort_integer_overflow() {
+  // FRIZZI: this test is meant to test something for BinSort,
+  // but actually uses the kokkos::sort API with the assumption
+  // that underneath it calls binsort. I don't think this is correct,
+  // because if the kokkos::sort API chages impl, this test is not testing
+  // what it meants to test... so need to change this to actually use BinSort
+  // directly.
```
masterleinad, line 205: Should this be a FIXME?

algorithms/src/Kokkos_Sort.hpp
fnrizzi: ah good catch i will use !. 
No runtime error

algorithms/unit_tests/TestBinSortA.hpp
fnrizzi: i will add that , thanks!

masterleinad (approved):

algorithms/src/Kokkos_Sort.hpp
```diff
@@ -344,18 +348,16 @@ class BinSort {
...
     const size_t len        = range_end - range_begin;
     const size_t values_len = values_range_end - values_range_begin;
     if (len != values_len) {
       Kokkos::abort(
           "BinSort::sort: values range length != permutation vector length");
     }
 
+    using scratch_view_type =
+        Kokkos::View<typename ValuesViewType::data_type,
+                     typename ValuesViewType::device_type>;
```
fnrizzi, line 360: this is moved here to be closer to where it is actually used

crtrott (approved):

algorithms/src/Kokkos_Sort.hpp
```diff
@@ -146,8 +146,12 @@ class BinSort {
         Kokkos::is_view<SrcViewType>::value,
         Kokkos::View<typename SrcViewType::const_data_type,
                      typename SrcViewType::array_layout,
-                     typename SrcViewType::device_type,
-                     Kokkos::MemoryTraits<Kokkos::RandomAccess> >,
+                     typename SrcViewType::device_type
+#if !defined(KOKKOS_COMPILER_NVHPC)  // FIXME_NVHPC
+                     ,
+                     Kokkos::MemoryTraits<Kokkos::RandomAccess>
+#endif
```
ldh4, line 153: Is this ifdef guard necessary for unit tests in this PR to work?
I'm curious what caused this ifdef for `NVHPC` to be added, seeing that the ifdef below is still in the unit test.
```
#if defined(KOKKOS_ENABLE_CUDA) && \
    defined(KOKKOS_COMPILER_NVHPC)  // FIXME_NVHPC
  if (!std::is_same_v<ExecutionSpace, Kokkos::Cuda>)
#endif
```
Not asking for any change, just looking for a clarification.

ldh4 (approved):

ldh4 (approved):

algorithms/src/Kokkos_Sort.hpp
fnrizzi: I have not modified the existing tests, but i added TestBinSortB which needs that guard to work. Makes sense?

dalg24:
Retest this please

Rombur (approved):

PR #6083, opened 2023-04-26
Implement Kokkos::printf

masterleinad:
Supersedes #6061. This pull request implements `Kokkos::printf` as discussed.
After trying a bunch of implementations (just wrap `KOKKOS_IMPL_DO_NOT_USE_PRINTF`, define a separate function that calls `printf`), it seems that importing `printf` into the `Kokkos` namespace is the only option that works. 

In my opinion, it's a good idea to keep the `KOKKOS_IMPL_DO_NOT_USE_PRINTF` macro around a little longer to allow some users to transition.

### DIFF
--- a/core/src/impl/Kokkos_Printf.hpp
+++ b/core/src/impl/Kokkos_Printf.hpp
@@ -1,0 +1,51 @@
+//@HEADER
+// ************************************************************************
+//
+//                        Kokkos v. 4.0
+//       Copyright (2022) National Technology & Engineering
+//               Solutions of Sandia, LLC (NTESS).
+//
+// Under the terms of Contract DE-NA0003525 with NTESS,
+// the U.S. Government retains certain rights in this software.
+//
+// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
+#ifndef KOKKOS_IMPL_PRINTF_HPP
+#define KOKKOS_IMPL_PRINTF_HPP
+
+#include <Kokkos_Macros.hpp>
+
+#ifdef KOKKOS_ENABLE_SYCL
+#include <sycl/sycl.hpp>
+#else
+#include <cstdio>
+#endif
+
+namespace Kokkos {
+
+// In contrast to std::printf, return void to get a consistent behavior across
+// backends. The GPU backends always return 1 and NVHPC only compiles if we
+// don't ask for the return value.
+template <typename... Args>
+KOKKOS_FUNCTION void printf(const char* format, Args... args) {
+#ifdef KOKKOS_ENABLE_SYCL
+  // Some compilers warn if "args" is empty and format is not a string literal
+  if constexpr (sizeof...(Args) == 0)
+    sycl::ext::oneapi::experimental::printf("%s", format);
+  else
+    sycl::ext::oneapi::experimental::printf(format, args...);
+#else
+  if constexpr (sizeof...(Args) == 0)
+    ::printf("%s", format);
+  else
+    ::printf(format, args...);
+#endif
+}
+
+}  // namespace Kokkos
+
+#endif /* #ifndef KOKKOS_IMPL_PRINTF_HPP */
--- a/core/unit_test/TestPrintf.hpp
+++ b/core/unit_test/TestPrintf.hpp
@@ -1,0 +1,38 @@
+//@HEADER
+// ************************************************************************
+//
+//                        Kokkos v. 4.0
+//       Copyright (2022) National Technology & Engineering
+//               Solutions of Sandia, LLC (NTESS).
+//
+// Under the terms of Contract DE-NA0003525 with NTESS,
+// the U.S. Government retains certain rights in this software.
+//
+// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
+#include <gtest/gtest.h>
+
+#include <Kokkos_Core.hpp>
+#include <fstream>
+
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <fcntl.h>
+
+template <class ExecutionSpace>
+void test_kokkos_printf() {
+  ::testing::internal::CaptureStdout();
+  Kokkos::parallel_for(
+      Kokkos::RangePolicy<ExecutionSpace>(0, 1),
+      KOKKOS_LAMBDA(int) { Kokkos::printf("Print an integer: %d", 2); });
+  Kokkos::fence();
+  auto const captured = ::testing::internal::GetCapturedStdout();
+  std::string expected_string("Print an integer: 2");
+  ASSERT_EQ(captured, expected_string);
+}
+
+TEST(TEST_CATEGORY, kokkos_printf) { test_kokkos_printf<TEST_EXECSPACE>(); }
--- a/core/unit_test/TestPrintf.hpp
+++ b/core/unit_test/TestPrintf.hpp
@@ -17,11 +17,6 @@
 #include <gtest/gtest.h>
 
 #include <Kokkos_Core.hpp>
-#include <fstream>
-
-#include <sys/types.h>
-#include <sys/stat.h>
-#include <fcntl.h>
 
 template <class ExecutionSpace>
 void test_kokkos_printf() {
--- a/containers/src/Kokkos_DynRankView.hpp
+++ b/containers/src/Kokkos_DynRankView.hpp
@@ -221,7 +221,7 @@ KOKKOS_INLINE_FUNCTION bool dyn_rank_view_verify_operator_bounds(
     return (size_t(i) < map.extent(R)) &&
            dyn_rank_view_verify_operator_bounds<R + 1>(rank, map, args...);
   } else if (i != 0) {
-    KOKKOS_IMPL_DO_NOT_USE_PRINTF(
+    Kokkos::printf(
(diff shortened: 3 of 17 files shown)


masterleinad:
Retest this please.

masterleinad:
NVHPC seems to error out when we are not using a string literal for the formatting string.

cz4rs (approved):
Looks good to me.

core/src/impl/Kokkos_Printf.hpp
```diff
@@ -0,0 +1,32 @@
...
+
+#include <Kokkos_Macros.hpp>
+
+namespace Kokkos {
+
+#ifdef KOKKOS_ENABLE_SYCL
+using sycl::ext::oneapi::experimental::printf;
+#else
+using ::printf;
+#endif
```
cz4rs, line 28: Possibly a short comment (like in https://github.com/kokkos/kokkos-core-wiki/pull/369) could land here?

Rombur (approved):

core/unit_test/TestPrintf.hpp
```diff
@@ -0,0 +1,88 @@
...
+  char* tmp_filename = tmpnam(nullptr);
+  ASSRET_NE(tmp_filename, nullptr);
+#else
+  char tmp_filename[16] = "/tmp/XXXXXXXXXX";
+  int error             = mkstemp(tmp_filename);
+  ASSERT_NE(error, -1);
+#endif
+
+  int backup_no, new_no;
+  fflush(stdout);
```
dalg24, line 38: Any particular reason why you are not using `::testing::internal::CaptureStdout()`
It is not part of the GTest API but we already use it elsewhere and if/when we get into trouble for it we can encapsulate and replace.

masterleinad: I didn't quite realize that we are already reaching into someone else's `internal` namespace. Sure, we can try that instead. It's just for testing anyway.

core/src/impl/Kokkos_Printf.hpp
```diff
@@ -0,0 +1,47 @@
...
+// interface, Kokkos::printf for the Serial backend is implemented in terms of
+// std::printf printing the data specified in format and args... returning the
+// number of characters written if successful or a negative value if an error
+// encountered. The implementation for other backends emulates this behavior.
+#ifdef KOKKOS_ENABLE_SYCL
+using sycl::ext::oneapi::experimental::printf;
+#elif defined(KOKKOS_COMPILER_NVHPC)  // FIXME_NVHPC
+template <typename... Args>
+KOKKOS_FUNCTION int printf(const char* format, Args... args) {
+  ::printf(format, args...);
```
masterleinad, line 38: Returning a value here doesn't work with `NVHPC`.

dalg24: Did we discuss whether we should be providing a checked version or return anything at all?
By that I mean, should we attempt to cause program termination when we detect than an error occurred?
Should we return void instead of "passing through" the return value of `printf` as currently proposed?

I don't necessarily have an issue with what is proposed but I do want to make sure we considered alternatives.

core/src/impl/Kokkos_Printf.hpp
dalg24, line 1: Did we discuss whether that should be a private or public header?

masterleinad: No, my intuition was to do the same as for `core/src/impl/Kokko_Error.hpp` (and it's easier to go from "private" to "public" than the other way around).

core/src/impl/Kokkos_Printf.hpp
masterleinad: I think it's good to stay close to what `printf` does since that's what people are used to (even though I doubt many people use the return value anyway).

core/src/impl/Kokkos_Printf.hpp
```diff
@@ -0,0 +1,49 @@
...
+#include <Kokkos_Macros.hpp>
+
+namespace Kokkos {
+
+// We are forwarding the format string to printf but gcc and clang warn about
+// the format string not being a literal.
+#if defined(__GNUC__)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wformat-security"
+#endif
```
dalg24, line 29: I don't get this one.  Can you share a godbolt link?

masterleinad: I pushed commit to remove these lines and you can see the CI failing with these warnings.

masterleinad: Also, see https://godbolt.org/z/bdM4ozM9r where I found a different workaround when there are no arguments for the format string.

masterleinad:
Failures in `CUDA-11.7-NVHPC`
```
[ 92%] Building CXX object core/unit_test/CMakeFiles/Kokkos_CoreUnitTest_Cuda1.dir/cuda/TestCuda_RangePolicyConstructors.cpp.o
NVC++-S-0000-Internal compiler error. extractor: bad opc       0  (/var/jenkins/workspace/Kokkos/algorithms/unit_tests/TestStdAlgorithmsMoveBackward.cpp: 216)
NVC++-S-0000-Internal compiler error. extractor: bad opc       0  (/var/jenkins/workspace/Kokkos/algorithms/unit_tests/TestStdAlgorithmsMoveBackward.cpp: 216)
NVC++-S-0000-Internal compiler error. extractor: bad opc       0  (/var/jenkins/workspace/Kokkos/algorithms/unit_tests/TestStdAlgorithmsMoveBackward.cpp: 216)
NVC++-S-0000-Internal compiler error. extractor: bad opc       0  (/var/jenkins/workspace/Kokkos/algorithms/unit_tests/TestStdAlgorithmsMoveBackward.cpp: 216)
NVC++-S-0000-Internal compiler error. extractor: bad opc       0  (/var/jenkins/workspace/Kokkos/algorithms/unit_tests/TestStdAlgorithmsMoveBackward.cpp: 216)
NVC++-S-0000-Internal compiler error. extractor: bad opc       0  (/var/jenkins/workspace/Kokkos/algorithms/unit_tests/TestStdAlgorithmsMoveBackward.cpp: 216)
NVC++-S-0000-Internal compiler error. extractor: bad opc       0  (/var/jenkins/workspace/Kokkos/algorithms/unit_tests/TestStdAlgorithmsMoveBackward.cpp: 216)
NVC++-S-0000-Internal compiler error. extractor: bad opc       0  (/var/jenkins/workspace/Kokkos/algorithms/unit_tests/TestStdAlgorithmsMoveBackward.cpp: 216)
[ 92%] Building CXX object core/unit_test/CMakeFiles/Kokkos_CoreUnitTest_OpenMP.dir/openmp/TestOpenMP_ViewCopy_a.cpp.o
NVC++/x86-64 Linux 22.9-0: compilation completed with severe errors
```
are unrelated.

dalg24:
Do these show on other builds?

masterleinad:
> Do these show on other builds?

I have seen that in a couple of other builds.

core/src/impl/Kokkos_Printf.hpp
```diff
@@ -0,0 +1,45 @@
...
+// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
+#ifndef KOKKOS_IMPL_PRINTF_HPP
+#define KOKKOS_IMPL_PRINTF_HPP
+
+#include <Kokkos_Macros.hpp>
```
dalg24, line 20:
```suggestion
#include <Kokkos_Macros.hpp>

#include <cstdio>
#ifdef KOKKOS_ENABLE_SYCL
#include <???>
#endif
```

core/src/impl/Kokkos_Printf.hpp
```diff
@@ -0,0 +1,45 @@
...
+// backends. The GPU backends always return 1 and NVHPC only compiles if we
+// don't ask for the return value.
+template <typename... Args>
+KOKKOS_FUNCTION void printf(const char* format, Args... args) {
+#ifdef KOKKOS_ENABLE_SYCL
+  // Some compilers warn if "args" is empty and format is not a string literal
+  if constexpr (sizeof...(Args) == 0)
+    sycl::ext::oneapi::experimental::printf("%s", format);
+  else
+    sycl::ext::oneapi::experimental::printf(format, args...);
```
dalg24, line 34: Also do we mean to go through that printf rather than the one from `<cstdio>` on the host side?
(we could in principle use `KOKKOS_IF_ON_DEVICE`)

masterleinad: Yes, I don't see a good reason not to also use `sycl::experimental::printf` on the host.

core/src/impl/Kokkos_Printf.hpp
```diff
@@ -0,0 +1,51 @@
...
+#ifndef KOKKOS_IMPL_PRINTF_HPP
+#define KOKKOS_IMPL_PRINTF_HPP
+
+#include <Kokkos_Macros.hpp>
+
+#ifdef KOKKOS_ENABLE_SYCL
+#include <cstdio>
+#else
+#include <sycl/sycl.hpp>
+#endif
```
dalg24, line 26: No that is the opposite

dalg24:
```suggestion
#ifdef KOKKOS_ENABLE_SYCL
#include <sycl/sycl.hpp>
#else
#include <cstdio>
#endif
```

core/unit_test/TestPrintf.hpp
```diff
@@ -0,0 +1,38 @@
...
+//@HEADER
+
+#include <gtest/gtest.h>
+
+#include <Kokkos_Core.hpp>
+#include <fstream>
+
+#include <sys/types.h>
+#include <sys/stat.h>
+#include <fcntl.h>
```
dalg24, line 24:
```suggestion
```

PR #6086, opened 2023-04-27
Fix up SYCL execution space instance creation for Intel GPUs

masterleinad:
Fixes #6074 for Intel GPUs. I missed a "{" for the case that is not tested in the CI.


dalg24 (approved):

fnrizzi (approved):

dalg24:
Not waiting for the CI that is down

PR #6087, opened 2023-04-27
Reorganize ZeroMemset

tcclevenger:
Changes:
- Move forward declaration of `ZeroMemset` to new `impl/Kokkos_ZeroMemset_fwd.hpp` file
- Move impl to exec specific files:
  -  `ZeroMemset<Cuda,...>` to new `Cuda/Kokkos_Cuda_ZeroMemset.hpp` file
  - HIP, Serial, SYCL likewise
  - `ZeroMemset<HostSpace,...>` to new `impl/Kokkos_HostSpace_ZeroMemset.hpp`

### DIFF
--- a/core/src/Cuda/Kokkos_Cuda_ZeroMemset.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_ZeroMemset.hpp
@@ -17,8 +17,7 @@
 #define KOKKOS_CUDA_ZEROMEMSET_HPP
 
 #include <Kokkos_Macros.hpp>
-#ifdef KOKKOS_ENABLE_CUDA
-
+#include <Cuda/Kokkos_Cuda.hpp>
 #include <impl/Kokkos_ZeroMemset_fwd.hpp>
 
 namespace Kokkos {
@@ -46,5 +45,4 @@ struct ZeroMemset<Kokkos::Cuda, DT, DP...> {
 }  // namespace Impl
 }  // namespace Kokkos
 
-#endif  // defined(KOKKOS_ENABLE_CUDA)
 #endif  // !defined(KOKKOS_CUDA_ZEROMEMSET_HPP)
--- a/core/src/Cuda/Kokkos_Cuda_ZeroMemset.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_ZeroMemset.hpp
@@ -1,0 +1,50 @@
+//@HEADER
+// ************************************************************************
+//
+//                        Kokkos v. 4.0
+//       Copyright (2022) National Technology & Engineering
+//               Solutions of Sandia, LLC (NTESS).
+//
+// Under the terms of Contract DE-NA0003525 with NTESS,
+// the U.S. Government retains certain rights in this software.
+//
+// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+#ifndef KOKKOS_CUDA_ZEROMEMSET_HPP
+#define KOKKOS_CUDA_ZEROMEMSET_HPP
+
+#include <Kokkos_Macros.hpp>
+#ifdef KOKKOS_ENABLE_CUDA
+
+#include <impl/Kokkos_ZeroMemset_fwd.hpp>
+
+namespace Kokkos {
+namespace Impl {
+
+template <class DT, class... DP>
+struct ZeroMemset<Kokkos::Cuda, DT, DP...> {
+  ZeroMemset(const Kokkos::Cuda& exec_space_instance,
+             const View<DT, DP...>& dst,
+             typename View<DT, DP...>::const_value_type&) {
+    KOKKOS_IMPL_CUDA_SAFE_CALL(cudaMemsetAsync(
+        dst.data(), 0,
+        dst.size() * sizeof(typename View<DT, DP...>::value_type),
+        exec_space_instance.cuda_stream()));
+  }
+
+  ZeroMemset(const View<DT, DP...>& dst,
+             typename View<DT, DP...>::const_value_type&) {
+    KOKKOS_IMPL_CUDA_SAFE_CALL(
+        cudaMemset(dst.data(), 0,
+                   dst.size() * sizeof(typename View<DT, DP...>::value_type)));
+  }
+};
+
+}  // namespace Impl
+}  // namespace Kokkos
+
+#endif  // defined(KOKKOS_ENABLE_CUDA)
+#endif  // !defined(KOKKOS_CUDA_ZEROMEMSET_HPP)
--- a/core/src/Cuda/Kokkos_Cuda_ZeroMemset.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_ZeroMemset.hpp
@@ -23,22 +23,20 @@
 namespace Kokkos {
 namespace Impl {
 
-template <class DT, class... DP>
-struct ZeroMemset<Kokkos::Cuda, DT, DP...> {
-  ZeroMemset(const Kokkos::Cuda& exec_space_instance,
-             const View<DT, DP...>& dst,
-             typename View<DT, DP...>::const_value_type&) {
+template <class T, class... P>
+struct ZeroMemset<Kokkos::Cuda, View<T, P...>> {
+  ZeroMemset(const Kokkos::Cuda& exec_space_instance, const View<T, P...>& dst,
+             typename View<T, P...>::const_value_type&) {
     KOKKOS_IMPL_CUDA_SAFE_CALL(cudaMemsetAsync(
-        dst.data(), 0,
-        dst.size() * sizeof(typename View<DT, DP...>::value_type),
+        dst.data(), 0, dst.size() * sizeof(typename View<T, P...>::value_type),
         exec_space_instance.cuda_stream()));
   }
 
-  ZeroMemset(const View<DT, DP...>& dst,
-             typename View<DT, DP...>::const_value_type&) {
+  ZeroMemset(const View<T, P...>& dst,
+             typename View<T, P...>::const_value_type&) {
     KOKKOS_IMPL_CUDA_SAFE_CALL(
         cudaMemset(dst.data(), 0,
-                   dst.size() * sizeof(typename View<DT, DP...>::value_type)));
+                   dst.size() * sizeof(typename View<T, P...>::value_type)));
   }
 };
 
--- a/core/src/HIP/Kokkos_HIP_ZeroMemset.hpp
+++ b/core/src/HIP/Kokkos_HIP_ZeroMemset.hpp
@@ -23,21 +23,20 @@
 namespace Kokkos {
 namespace Impl {
 
-template <class DT, class... DP>
-struct ZeroMemset<HIP, DT, DP...> {
-  ZeroMemset(const HIP& exec_space, const View<DT, DP...>& dst,
-             typename View<DT, DP...>::const_value_type&) {
+template <class T, class... P>
+struct ZeroMemset<HIP, View<T, P...>> {
+  ZeroMemset(const HIP& exec_space, const View<T, P...>& dst,
+             typename View<T, P...>::const_value_type&) {
     KOKKOS_IMPL_HIP_SAFE_CALL(hipMemsetAsync(
-        dst.data(), 0,
-        dst.size() * sizeof(typename View<DT, DP...>::value_type),
+        dst.data(), 0, dst.size() * sizeof(typename View<T, P...>::value_type),
         exec_space.hip_stream()));
   }
 
-  ZeroMemset(const View<DT, DP...>& dst,
-             typename View<DT, DP...>::const_value_type&) {
+  ZeroMemset(const View<T, P...>& dst,
+             typename View<T, P...>::const_value_type&) {
     KOKKOS_IMPL_HIP_SAFE_CALL(
         hipMemset(dst.data(), 0,
-                  dst.size() * sizeof(typename View<DT, DP...>::value_type)));
+                  dst.size() * sizeof(typename View<T, P...>::value_type)));
   }
 };
 
--- a/core/src/HIP/Kokkos_HIP_ZeroMemset.hpp
+++ b/core/src/HIP/Kokkos_HIP_ZeroMemset.hpp
@@ -1,0 +1,49 @@
+//@HEADER
+// ************************************************************************
+//
+//                        Kokkos v. 4.0
+//       Copyright (2022) National Technology & Engineering
+//               Solutions of Sandia, LLC (NTESS).
+//
+// Under the terms of Contract DE-NA0003525 with NTESS,
+// the U.S. Government retains certain rights in this software.
+//
+// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
+// See https://kokkos.org/LICENSE for license information.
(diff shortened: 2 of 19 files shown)


core/src/Cuda/Kokkos_Cuda_ZeroMemset.hpp
```diff
@@ -0,0 +1,50 @@
...
+// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+#ifndef KOKKOS_CUDA_ZEROMEMSET_HPP
+#define KOKKOS_CUDA_ZEROMEMSET_HPP
+
+#include <Kokkos_Macros.hpp>
+#ifdef KOKKOS_ENABLE_CUDA
```
dalg24, line 20: You don't need to guard for the Cuda backend being enabled.
The header needs to be self-contained though so you are missing header includes and/or forward definitions.

core/src/Kokkos_HostSpace.hpp
```diff
@@ -38,6 +38,7 @@ static_assert(false,
 
 #include "impl/Kokkos_HostSpace_deepcopy.hpp"
 #include <impl/Kokkos_MemorySpace.hpp>
+#include <impl/Kokkos_ZeroMemset_fwd.hpp>
```
dalg24, line 41: Why did you not move the default host specialization to a separate header?
Not necessarily saying you need to, wondering what was the reasoning.

core/src/SYCL/Kokkos_SYCL_DeepCopy.hpp
```diff
@@ -18,6 +18,7 @@
 #define KOKKOS_SYCLDEEPCOPY_HPP
 
 #include <Kokkos_Core_fwd.hpp>
+#include <impl/Kokkos_ZeroMemset_fwd.hpp>
```
dalg24, line 21: Why no `<SYCL/Kokkos_SYCL_ZeroMemset.hpp>` header?

core/src/impl/Kokkos_ViewMapping.hpp
```diff
@@ -32,6 +32,7 @@
 #include <impl/Kokkos_Atomic_View.hpp>
 #include <impl/Kokkos_Tools.hpp>
 #include <impl/Kokkos_StringManipulation.hpp>
+#include <impl/Kokkos_ZeroMemset_fwd.hpp>
```
dalg24, line 35: Similar question here, why is the fallback generic implementation in that file rather than in its own header?

core/src/impl/Kokkos_ZeroMemset_fwd.hpp
```diff
@@ -0,0 +1,29 @@
...
+//
+//@HEADER
+
+#ifndef KOKKOS_ZEROMEMSET_FWD_HPP
+#define KOKKOS_ZEROMEMSET_FWD_HPP
+
+namespace Kokkos {
+namespace Impl {
+
+template <typename ExecutionSpace, class DT, class... DP>
```
dalg24, line 23: I was about to ask you to spell out what DT and DP... stand for but I am wondering now why it is not just templated on View instead.  Please check if it was intentional when ZeroMemset was introduced.

core/src/HIP/Kokkos_HIP_ZeroMemset.hpp
```diff
@@ -0,0 +1,49 @@
...
+// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+#ifndef KOKKOS_HIP_ZEROMEMSET_HPP
+#define KOKKOS_HIP_ZEROMEMSET_HPP
+
+#include <Kokkos_Macros.hpp>
+#ifdef KOKKOS_ENABLE_HIP
```
fnrizzi, line 20: same comment for cuda above here

fnrizzi (changes requested):

core/src/impl/Kokkos_ZeroMemset_fwd.hpp
```diff
@@ -0,0 +1,29 @@
...
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
+#ifndef KOKKOS_ZEROMEMSET_FWD_HPP
+#define KOKKOS_ZEROMEMSET_FWD_HPP
+
+namespace Kokkos {
+namespace Impl {
```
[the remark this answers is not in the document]
fnrizzi: are we adopting this convention about namespaces everywhere from now on?

core/src/Cuda/Kokkos_Cuda_ZeroMemset.hpp
fnrizzi: "The header needs to be self-contained though"
IIRC, there is a test that should be checking all headers for being self-contained, I think it is : `core/unit_tests/headers_self_contained ` how come this was not caught?

dalg24: Because we only find the header files directly sitting in src/ not the ones in nested subdirectories
https://github.com/kokkos/kokkos/blob/aa1f48f3172069f212ed2c97b74b786999b26f17/core/unit_test/headers_self_contained/CMakeLists.txt#L6-L7

core/src/impl/Kokkos_ZeroMemset_fwd.hpp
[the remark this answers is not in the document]
dalg24: This is just my preference for this use case.  I don't think we had a discussion about it.
 Feel free to ignore if you disagree.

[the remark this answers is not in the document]
tcclevenger: I definitely agree with the change, but the other was what I've seen more of in the repo. I'll leave this as is in this PR and that change can be separate.

core/src/Cuda/Kokkos_Cuda_ZeroMemset.hpp
tcclevenger: Why is this the case? Would it make sense to add subdirectories?

masterleinad: The headers in subdirectories are not necessarily self-contained (yet).

masterleinad: You're welcome to give it a shot (in a separate pull request). 🙂

core/src/Kokkos_HostSpace.hpp
tcclevenger: I was a bit confused by what was considered the default impl. If `HostSpace` is what each backend without a specialization falls back to, I'll add it.

Looks like then only `Serial/` and `SYCL/` are left without this change, might as well add them here?

Other than that, `Kokkos_CopyViews.hpp` has it's own in-house impl of `ZeroMemset`, but the calls to `ZeroMemset` in `Kokkos_CopyViews.hpp` don't actually trigger that impl for me (tested Cuda and Serial). I'm guessing that there is some other configuration where that is used?

core/src/Cuda/Kokkos_Cuda_ZeroMemset.hpp
tcclevenger: Yeah, I'll give it a go.

core/src/Kokkos_HostSpace.hpp
masterleinad: The default implementation is https://github.com/kokkos/kokkos/blob/56ef02c0cd45b62c39b5a789488467331822499d/core/src/Kokkos_CopyViews.hpp#L1329-L1340 and the calls to `ZeroMemnset` should use the specializations instead if provided. This is what you are observing, right?

masterleinad: I think it's a good idea to tackle `Serial` and `SYCL` here as well.

core/src/impl/Kokkos_ZeroMemset_fwd.hpp
masterleinad: The main reason was to follow the design of the other functions in `Kokkos_CopyViews.hpp` (`contiguous_fill`, `contiguous_fill_or_memset`, `deep_copy`, `local_deep_copy`, `local_deep_copy_contiguous`).

core/src/Kokkos_HostSpace.hpp
tcclevenger: Even in `CopyViews`, I observed `HostSpace` being called as default (I incorrectly defined Cuda specialization at first, and `HostSpace` was called).

masterleinad: Yes, that's all the same template.

core/src/HIP/Kokkos_HIP_ZeroMemset.hpp
```diff
@@ -0,0 +1,48 @@
...
+    KOKKOS_IMPL_HIP_SAFE_CALL(
+        hipMemset(dst.data(), 0,
+                  dst.size() * sizeof(typename View<DT, DP...>::value_type)));
+  }
+};
+
+}  // namespace Impl
+}  // namespace Kokkos
+
+#endif  // defined(KOKKOS_ENABLE_HIP)
```
masterleinad, line 47:
```suggestion
```

core/src/Cuda/Kokkos_Cuda_ZeroMemset.hpp
```diff
@@ -0,0 +1,48 @@
...
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+#ifndef KOKKOS_CUDA_ZEROMEMSET_HPP
+#define KOKKOS_CUDA_ZEROMEMSET_HPP
+
+#include <Kokkos_Macros.hpp>
+#include <Cuda/Kokkos_Cuda.hpp>
+#include <impl/Kokkos_ZeroMemset_fwd.hpp>
```
dalg24, line 21: Needs a forward declaration of View as well but maybe a comment is good enough for now.

core/src/impl/Kokkos_ZeroMemset_fwd.hpp
dalg24: I think that we should template on View and then we call always do
```C++
template <class DT, class... DP>
struct ZeroMemset<my-execution-space, View<DT, DP...>> { /* ... */ };
```

and maybe spell out what "DT" ("data type"?? "destination type") and "DP" are

masterleinad: We oftentimes use `View<T, P...>` but we don't really need to know about the template parameters here anyway. I would also be happy with a comment.

crtrott:
Didn't do a careful review, but its a good change.

core/src/impl/Kokkos_ZeroMemset_fwd.hpp
masterleinad: Would you mind trying
```C++
template <typename ExecutionSpace, typename View>
struct ZeroMemset;

template <typename T, typename... P>
struct ZeroMemset<Cuda, Kokkos::View<T, P...>>
{};
```
(or we do it in a follow-up).

core/src/Kokkos_CopyViews.hpp
```diff
@@ -1327,15 +1327,14 @@ inline void contiguous_fill(
 }
 
 // Default implementation for execution spaces that don't provide a definition
-template <typename ExecutionSpace, class DT, class... DP>
+template <typename ExecutionSpace, class ViewType>
 struct ZeroMemset {
-  ZeroMemset(const ExecutionSpace& exec_space, const View<DT, DP...>& dst,
-             typename ViewTraits<DT, DP...>::const_value_type& value) {
+  ZeroMemset(const ExecutionSpace& exec_space, const ViewType& dst,
+             typename ViewType::const_value_type& value) {
```
tcclevenger, line 1333: @masterleinad Am I correct that the default version should also have template 
```
template <typename ExecutionSpace, class ViewType>
```
If so, in Kokkos is it equivalent to use 
```
View<DT, DP...>::const_value_type <-> ViewTraits<DT, DP...>::const_value_type
```

core/src/Cuda/Kokkos_Cuda_ZeroMemset.hpp
```diff
@@ -0,0 +1,46 @@
...
+    KOKKOS_IMPL_CUDA_SAFE_CALL(cudaMemsetAsync(
+        dst.data(), 0, dst.size() * sizeof(typename View<T, P...>::value_type),
+        exec_space_instance.cuda_stream()));
+  }
+
+  ZeroMemset(const View<T, P...>& dst,
+             typename View<T, P...>::const_value_type&) {
+    KOKKOS_IMPL_CUDA_SAFE_CALL(
+        cudaMemset(dst.data(), 0,
+                   dst.size() * sizeof(typename View<T, P...>::value_type)));
```
dalg24, line 39: I realize now this is a bug.  We should be using the stream from the default instance.  Probably not the place to fix though.

core/src/Cuda/Kokkos_Cuda_ZeroMemset.hpp
tcclevenger: Is this in reference to self contained headers? I think that test will pass from includes contained in 
`#include <Cuda/Kokkos_Cuda.hpp>`

core/src/Cuda/Kokkos_Cuda_ZeroMemset.hpp
masterleinad: I wouldn't necessarily call it a bug since the default stream still synchronizes with all other streams but I agree that we could/should use `cudaMemsetAsync` here as well.

core/src/impl/Kokkos_ViewMapping.hpp
masterleinad: The compiler would complain if it couldn't see the default implementation at the point it is used. Yes, having a separate file for the fallback implementation makes sense to me but the question is if you can get the includes right to make that work.

core/src/Kokkos_CopyViews.hpp
masterleinad: Yes, `View<DT, DP...>::const_value_type` and `ViewTraits<DT, DP...>::const_value_type` are the same.

masterleinad:
If you wanted to comment on using a "value" argument, you could add for the generic template:
> Although "value" is required to represent a zero-byte object, we might still call other functions here that require passing an explicit value and we avoid constructing an all-zero-byte object on the fly by keeping this function argument.

dalg24 (commented):
Looks good overall.

tcclevenger:
@masterleinad @dalg24 I agree that the default impl should be in its own file (and available to `Kokkos_ViewMapping.hpp`), but it is a bit messy with dependencies and circular inclusions so it may not be quick. I agree with Daniel and say we merge this and then I'll work on separating the impl parallel with my Cuda API PR, If that is ok.

Alos, what is the meaning for gray dash (as opposed to green check/red "x") for HIP builds in OpenBlueOcean output? I'll commit the suggested change above and maybe this will go away with retesting, but just curious since it seemed to run.

dalg24 (approved):

PR #6088, opened 2023-04-27
Disable failed bit manipulation tests when compiled by NVHPC 

seyonglee:
NVHPC compiler failed to correctly translate accelerator region for tests in
core/unit_test/TestBitManipulationBuiltins.hpp if the tested variable type is char
or short; disable those failed tests.

### DIFF
--- a/core/unit_test/TestBitManipulationBuiltins.hpp
+++ b/core/unit_test/TestBitManipulationBuiltins.hpp
@@ -119,9 +119,14 @@ void test_bit_manip_countl_zero() {
 
 TEST(TEST_CATEGORY, bit_manip_countl_zero) {
 // FIXME_NVHPC: NVC++-W-0155-Compiler failed to translate accelerator region
-#if !defined(KOKKOS_COMPILER_NVHPC)
-  test_bit_manip_countl_zero<unsigned char>();
-  test_bit_manip_countl_zero<unsigned short>();
+#if defined(KOKKOS_ENABLE_OPENACC) && defined(KOKKOS_COMPILER_NVHPC)
+  if constexpr (!std::is_same_v<TEST_EXECSPACE,
+                                Kokkos::Experimental::OpenACC>) {
+#endif
+    test_bit_manip_countl_zero<unsigned char>();
+    test_bit_manip_countl_zero<unsigned short>();
+#if defined(KOKKOS_ENABLE_OPENACC) && defined(KOKKOS_COMPILER_NVHPC)
+  }
 #endif
   test_bit_manip_countl_zero<unsigned int>();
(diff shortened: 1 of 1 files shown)


masterleinad (approved):

core/unit_test/TestBitManipulationBuiltins.hpp
```diff
@@ -118,8 +118,11 @@ void test_bit_manip_countl_zero() {
 }
 
 TEST(TEST_CATEGORY, bit_manip_countl_zero) {
+// FIXME_NVHPC: NVC++-W-0155-Compiler failed to translate accelerator region
+#if !defined(KOKKOS_COMPILER_NVHPC)
```
dalg24, line 122: Did you try to restrict further?  It passes with NVC++ when doing CUDA builds so the right thing to do here is probably something like
```C++
#ifdef KOKKOS_ENABLE_OPENACC
if constexpr (!std::is_same_v<TEST_EXECSPACE, Kokkos::Experimental::OpenACC>)
{
#endif
```

seyonglee: > Did you try to restrict further? It passes with NVC++ when doing CUDA builds so the right thing to do here is probably something like
> 
> ```c++
> #if defined(KOKKOS_ENABLE_OPENACC) && defined(KOKKOS_COMPILER_NVHPC)
> if constexpr (!std::is_same_v<TEST_EXECSPACE, Kokkos::Experimental::OpenACC>)
> {
> #endif
> ```

You are correct; NVHPC fails only when compiling for the OpenACC backend; the serial backend didn't fail. 
Updated the code as suggested.

dalg24 (approved):
Out of curiosity, what version of the toolchain are you using when you see that failure.
I am asking because our current CI setting is passing.  We can discuss updating to something more recent if you like.
Also, did you guys report the issue to the vendors?

seyonglee:
> Out of curiosity, what version of the toolchain are you using when you see that failure.
> I am asking because our current CI setting is passing. We can discuss updating to something more recent if you like.
> Also, did you guys report the issue to the vendors?

I used NVHPC V22.11 for local testing. The root cause for the failure is that the NVHPC OpenACC compiler fails if the reduction variable type is either char or short int. NVIDIA HPC SDK document does not explicitly explain which primitive types are supported for reduction operations (and atomic operations too). I don't have access to the latest NVHPC (V23.3) and thus don't know whether the latest NVHPC supports char/short reduction variables or not. I can report this issue to the vendor after checking the latest NVHPC compiler.

dalg24:
Retest this please

masterleinad:
The HIP CI did not start running before jenkins CI timed out but everything else is passing.

PR #6091, opened 2023-04-28
Allow using multiple CUDA devices 

masterleinad:
Based on top of #5989. This pull request explores how multi-device support could look like using `TestCuda_InterOp_Streams`. The proposed constructor for the execution space instance is
```
Cuda(int device_id, cudaStream_t stream);
```
assuming a `cudaStream_t` is provided as well. It shouldn't be a problem to allow for omitting that argument, though.
Of course, the tricky part is sharing device allocations between devices. `deep_copy` should work independently of the device and we can allocate on the correct device if an execution space instance is provided in the View allocation.
For accessing allocations on a different device, the example uses `cudaDeviceEnablePeerAccess` which comes at a cost, of course. In the end, I think the user should be responsible if they want cross-device access.

Of course, we need a corresponding pull request on the `desul`-side if we decide that we care about supporting to use the arbitrary size atomics on multiple GPUs within this pull request.

~~The other changes in this pull request, show that it's crucial to only set the device in #5989 if we actually know the device to use. Setting the device to the default is more harmful than useful.~~


crtrott:
I don't get your point @masterleinad regarding the default device. We always set a device don't we? So the default execution space instance is gonna be associated with a device which is based on our common decision process (kokkos-device-id, map-by etc.). That default execution space and the associated device needs to be used for any operation which doesn't get an execution space instance passed.

masterleinad:
> I don't get your point @masterleinad regarding the default device. We always set a device don't we? So the default execution space instance is gonna be associated with a device which is based on our common decision process (kokkos-device-id, map-by etc.). That default execution space and the associated device needs to be used for any operation which doesn't get an execution space instance passed.

In the end, we have (or could infer/pass) an execution space instance (or at least the correct device) in most cases anyway so the question would really only be about cases where that is not the case. 
While looking into making a simple example work using multiple devices, I discovered places where setting the device to the default device would be wrong, see https://github.com/kokkos/kokkos/pull/6091/files/a44c11735437a56eab8205b92d3ce5f02637dd94..63270b8f1b9208e130f3fc160a039831b8f5f963 or just makes the implementation more difficult by figuring out at which point we actually set `m_cudaDev`(`cuda_kernel_arch`). A pretty annoying case was `get_cuda_kernel_func_attributes` where we don't have the correct device available and just not changing it would (likely) do the right thing.

Of course, I see the failure in #5713 but we should really see how many places are left where we can't know about the correct device to use. On top of my head, there would be `Kokkos::fence()` which should then just call `cudaDeviceSynchronize` on all devices we know about.

core/src/Cuda/Kokkos_CudaSpace.cpp
```diff
@@ -83,11 +83,12 @@ void DeepCopyAsyncCuda(void *dst, const void *src, size_t n) {
...
           dst, src, n, cudaMemcpyDefault, s)));
-  Impl::cuda_stream_synchronize(
-      s,
+
+  Kokkos::Tools::Experimental::Impl::profile_fence_event<Kokkos::Cuda>(
+      "Kokkos::Impl::DeepCopyAsyncCuda: Deep Copy Stream Sync",
       Kokkos::Tools::Experimental::SpecialSynchronizationCases::
           DeepCopyResourceSynchronization,
-      "Kokkos::Impl::DeepCopyAsyncCuda: Deep Copy Stream Sync");
+      [&]() { KOKKOS_IMPL_CUDA_SAFE_CALL(cudaStreamSynchronize(s)); });
```
masterleinad, line 91: This was the only case where this overload of `cuda_stream_synchronize` was used and it felt easier to just inline the call explicitly.

core/src/Cuda/Kokkos_CudaSpace.cpp
```diff
@@ -168,45 +169,33 @@ void *impl_allocate_common(const Cuda &exec_space, const char *arg_label,
...
+    error_code = cuda_internal.cuda_malloc_async_wrapper(&ptr, arg_alloc_size);
+
     if (exec_space_provided) {
-      error_code =
-          exec_space.impl_internal_space_instance()->cuda_malloc_async_wrapper(
-              &ptr, arg_alloc_size);
       exec_space.fence("Kokkos::Cuda: backend fence after async malloc");
     } else {
-      error_code = Impl::CudaInternal::singleton().cuda_malloc_async_wrapper(
-          &ptr, arg_alloc_size);
```
masterleinad, line 183: I don't see a good reason to use the `singleton` here instead of the provided execution space instance. The default constructed execution space instance corresponds to the singleton anyway.

core/src/Cuda/Kokkos_Cuda.hpp
```diff
@@ -183,6 +183,8 @@ class Cuda {
 
   Cuda(cudaStream_t stream, bool manage_stream = false);
 
+  Cuda(int device_id, cudaStream_t stream, bool manage_stream = false);
```
masterleinad, line 186: That's the proposed new constructor.

core/src/Cuda/Kokkos_Cuda_Error.hpp
```diff
@@ -27,13 +27,7 @@
 namespace Kokkos {
 namespace Impl {
 
-void cuda_stream_synchronize(
-    const cudaStream_t stream,
-    Kokkos::Tools::Experimental::SpecialSynchronizationCases reason,
-    const std::string& name);
 void cuda_device_synchronize(const std::string& name);
-void cuda_stream_synchronize(const cudaStream_t stream,
-                             const std::string& name);
```
masterleinad, line 36: Ultimately, these functions don't need to be exposed since they are only used internally. Also note, that the last overload didn't have an implementation.

core/src/Cuda/Kokkos_Cuda_Instance.cpp
```diff
@@ -101,17 +101,16 @@ int cuda_kernel_arch() {
...
   int *d_arch = nullptr;
 
-  KOKKOS_IMPL_CUDA_SAFE_CALL((CudaInternal::singleton().cuda_malloc_wrapper(
-      reinterpret_cast<void **>(&d_arch), sizeof(int))));
-  KOKKOS_IMPL_CUDA_SAFE_CALL((CudaInternal::singleton().cuda_memcpy_wrapper(
-      d_arch, &arch, sizeof(int), cudaMemcpyDefault)));
+  KOKKOS_IMPL_CUDA_SAFE_CALL(
+      cudaMalloc(reinterpret_cast<void **>(&d_arch), sizeof(int)));
+  KOKKOS_IMPL_CUDA_SAFE_CALL(
+      cudaMemcpy(d_arch, &arch, sizeof(int), cudaMemcpyDefault));
```
masterleinad, line 108: We shouldn't change the device here.

core/src/Cuda/Kokkos_Cuda_Instance.cpp
```diff
@@ -135,7 +134,6 @@ Kokkos::View<uint32_t *, Kokkos::CudaSpace> cuda_global_unique_token_locks(
   return locks;
 }
 
-// FIXME_CUDA_MULTIPLE_DEVICES
 void cuda_device_synchronize(const std::string &name) {
   Kokkos::Tools::Experimental::Impl::profile_fence_event<Kokkos::Cuda>(
```
masterleinad, line 138: `cuda_device_synchronize` is only used in the `impl_static_fence` and thus is now interpreted as "synchronize all devices".

core/src/Cuda/Kokkos_Cuda_Instance.cpp
```diff
@@ -144,39 +142,25 @@ void cuda_device_synchronize(const std::string &name) {
...
-
-void cuda_stream_synchronize(
-    const cudaStream_t stream,
-    Kokkos::Tools::Experimental::SpecialSynchronizationCases reason,
-    const std::string &name) {
-  Kokkos::Tools::Experimental::Impl::profile_fence_event<Kokkos::Cuda>(
-      name, reason, [&]() {
-        KOKKOS_IMPL_CUDA_SAFE_CALL(
-            (Impl::CudaInternal::singleton().cuda_stream_synchronize_wrapper(
-                stream)));
```
masterleinad, line 179: This was just used in one place in `DeepCopy` for the deep copy stream.

core/src/Cuda/Kokkos_Cuda_Instance.cpp
```diff
@@ -364,13 +348,14 @@ CudaInternal &CudaInternal::singleton() {
   return self;
 }
 void CudaInternal::fence(const std::string &name) const {
-  Impl::cuda_stream_synchronize(get_stream(), this, name);
+  Impl::cuda_stream_synchronize(this, name);
```
masterleinad, line 351: We already know the stream to use.

core/src/Cuda/Kokkos_Cuda_Instance.cpp
```diff
@@ -433,21 +420,20 @@ Kokkos::Cuda::initialize WARNING: Cuda is allocating into UVMSpace by default
...
     KOKKOS_IMPL_CUDA_SAFE_CALL((cuda_malloc_host_wrapper(
-        reinterpret_cast<void **>(&constantMemHostStaging),
+        reinterpret_cast<void **>(&constantMemHostStagingPerDevice[m_cudaDev]),
         CudaTraits::ConstantMemoryUsage)));
 
+  if (!constantMemReusablePerDevice[m_cudaDev])
     KOKKOS_IMPL_CUDA_SAFE_CALL(
-        (cuda_event_create_wrapper(&constantMemReusable)));
-  }
+        (cuda_event_create_wrapper(&constantMemReusablePerDevice[m_cudaDev])));
```
masterleinad, line 436: A bunch of variables now need to be per device including the desul lock arrays.

core/src/Cuda/Kokkos_Cuda_Instance.cpp
```diff
@@ -471,19 +457,11 @@ Cuda::size_type *CudaInternal::scratch_flags(const std::size_t size) const {
...
+    KOKKOS_IMPL_CUDA_SAFE_CALL(cudaFree(m_scratchFlags));
     std::size_t alloc_size =
         multiply_overflow_abort(m_scratchFlagsCount, sizeScratchGrain);
-    Record *const r = Record::allocate(
-        Kokkos::CudaSpace(), "Kokkos::InternalScratchFlags", alloc_size);
-
-    Record::increment(r);
-
-    m_scratchFlags = reinterpret_cast<size_type *>(r->data());
+    KOKKOS_IMPL_CUDA_SAFE_CALL(cudaMalloc(&m_scratchFlags, alloc_size));
```
masterleinad, line 464: It's easier to get the device right if we don't go through `SharedAllocationRecord` which generally doesn't know about space instances and could easily set the device wrongly to the one used by the default execution space instance.

core/src/Cuda/Kokkos_Cuda_Instance.hpp
```diff
@@ -102,7 +105,7 @@ class CudaInternal {
  public:
   using size_type = Cuda::size_type;
 
-  inline static int m_cudaDev = -1;
+  int m_cudaDev = -1;
 
   // Device Properties
```
masterleinad, line 110: We still assume that all the devices used have the same properties. Dropping that assumption would require the following variables also to be per device (and some more changes when querying the functor properties for example).

core/src/Cuda/Kokkos_Cuda_Instance.hpp
```diff
@@ -145,11 +148,11 @@ class CudaInternal {
...
-  // FIXME_CUDA: these want to be per-device, not per-stream...  use of 'static'
-  //  here will break once there are multiple devices though
-  inline static unsigned long* constantMemHostStaging = nullptr;
-  inline static cudaEvent_t constantMemReusable       = nullptr;
-  inline static std::mutex constantMemMutex;
+  inline static std::set<int> cuda_devices = {};
+  inline static std::map<int, unsigned long*> constantMemHostStagingPerDevice =
+      {};
+  inline static std::map<int, cudaEvent_t> constantMemReusablePerDevice = {};
+  inline static std::map<int, std::mutex> constantMemMutexPerDevice;
```
masterleinad, line 155: It's convenient to store the `cuda_devices` in a static variable so we can iterate over them. `set` might not be the most performant choice here but is convenient and it shouldn't matter much since we don't use it in performance-critical code. Similarly, `std::map` is mostly convenient.

core/src/Cuda/Kokkos_Cuda_Instance.hpp
```diff
@@ -250,12 +253,6 @@ class CudaInternal {
     return cudaDeviceSetLimit(limit, value);
   }
 
-  template <bool setCudaDevice = true>
-  cudaError_t cuda_device_synchronize_wrapper() const {
-    if constexpr (setCudaDevice) set_cuda_device();
-    return cudaDeviceSynchronize();
-  }
-
```
masterleinad, line 258: Not used anymore.

core/src/Cuda/Kokkos_Cuda_Instance.hpp
```diff
@@ -478,9 +475,9 @@ class CudaInternal {
   }
 
   template <bool setCudaDevice = true>
-  cudaError_t cuda_stream_synchronize_wrapper(cudaStream_t stream) const {
+  cudaError_t cuda_stream_synchronize_wrapper() const {
     if constexpr (setCudaDevice) set_cuda_device();
-    return cudaStreamSynchronize(stream);
+    return cudaStreamSynchronize(get_stream<false>());
```
masterleinad, line 480: We always have a valid stream and that is the only one to use here.

core/src/Cuda/Kokkos_Cuda_KernelLaunch.hpp
```diff
@@ -128,17 +128,17 @@ inline void check_shmem_request(CudaInternal const* cuda_instance, int shmem) {
...
   // Only call cudaFuncGetAttributes once for each unique kernel
   // by leveraging static variable initialization rules
   auto wrap_get_attributes = [&]() -> cudaFuncAttributes {
     cudaFuncAttributes attr;
-    KOKKOS_IMPL_CUDA_SAFE_CALL(
-        (CudaInternal::singleton().cuda_func_get_attributes_wrapper(&attr,
-                                                                    func)));
+
+    KOKKOS_IMPL_CUDA_SAFE_CALL(cudaSetDevice(cuda_device));
+    KOKKOS_IMPL_CUDA_SAFE_CALL(cudaFuncGetAttributes(&attr, func));
```
masterleinad, line 140: It turned out to be problematic to call this function with the wrong device.
Again, the assumption is that all the devices have the same properties so it doesn't matter that this function uses a static variable.

core/unit_test/cuda/TestCuda_InterOp_StreamsMultiGPU.cpp
masterleinad, line 1: The test is basically the same as `TestCuda_InterOp_Streams.cpp` but duplicates all kernels to run on two different devices interleaved.

masterleinad:
With this pull request, we still have 
```
$ git grep -n "FIXME_CUDA_MULTIPLE_DEVICES"
core/src/Cuda/Kokkos_Cuda_Task.hpp:225:  // FIXME_CUDA_MULTIPLE_DEVICES
core/src/Cuda/Kokkos_Cuda_Task.hpp:465:  // FIXME_CUDA_MULTIPLE_DEVICES
core/src/Cuda/Kokkos_Cuda_ZeroMemset.hpp:39:    // FIXME_CUDA_MULTIPLE_DEVICES
```
I think we can discuss fixing tasks later and `ZeroMemset` called without arguments replaces running a functor on the default execution space instance anyway. Selecting the correct device matters here, though.

PR #6097, opened 2023-05-03
change impl of `is_sorted_until` to use reduce

fnrizzi:
Change impl from a scan (which needed an extra allocation) to a reduce without the extra allocation. 

This is a follow up to #6095 and should be merged *AFTER* being rebased on that. 
NOTE: this will fail CI until #6095 is merged and rebase is done.

### DIFF
--- a/algorithms/src/std_algorithms/impl/Kokkos_IsSortedUntil.hpp
+++ b/algorithms/src/std_algorithms/impl/Kokkos_IsSortedUntil.hpp
@@ -28,33 +28,30 @@ namespace Kokkos {
 namespace Experimental {
 namespace Impl {
 
-template <class IteratorType, class IndicatorViewType, class ComparatorType>
+template <class IteratorType, class ComparatorType, class ReducerType>
 struct StdIsSortedUntilFunctor {
-  using index_type = typename IteratorType::difference_type;
+  using index_type     = typename IteratorType::difference_type;
+  using red_value_type = typename ReducerType::value_type;
+
   IteratorType m_first;
-  IndicatorViewType m_indicator;
   ComparatorType m_comparator;
+  ReducerType m_reducer;
 
   KOKKOS_FUNCTION
-  void operator()(const index_type i, int& update, const bool final) const {
+  void operator()(const index_type i, red_value_type& red_value) const {
     const auto& val_i   = m_first[i];
     const auto& val_ip1 = m_first[i + 1];
-
     if (m_comparator(val_ip1, val_i)) {
-      ++update;
-    }
-
-    if (final) {
-      m_indicator(i) = update;
+      m_reducer.join(red_value, i);
     }
   }
 
   KOKKOS_FUNCTION
-  StdIsSortedUntilFunctor(IteratorType _first1, IndicatorViewType indicator,
-                          ComparatorType comparator)
-      : m_first(std::move(_first1)),
-        m_indicator(std::move(indicator)),
-        m_comparator(std::move(comparator)) {}
+  StdIsSortedUntilFunctor(IteratorType first, ComparatorType comparator,
+                          ReducerType reducer)
+      : m_first(std::move(first)),
+        m_comparator(std::move(comparator)),
+        m_reducer(std::move(reducer)) {}
 };
 
 template <class ExecutionSpace, class IteratorType, class ComparatorType>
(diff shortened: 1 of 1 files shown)


algorithms/src/std_algorithms/impl/Kokkos_IsSortedUntil.hpp
```diff
@@ -73,40 +70,31 @@ IteratorType is_sorted_until_impl(const std::string& label,
...
-  return first + (shift + 1);
+    Do a par_reduce computing the *min* index that breaks the sorting.
+    If such an index is found, then the range is sorted until that element.
+    If no such index is found, then the range is sorted until the end.
+  */
+  using index_type = typename IteratorType::difference_type;
+  index_type red_result;
+  index_type red_result_init;
+  ::Kokkos::Min<index_type> reducer(red_result);
+  reducer.init(red_result_init);
```
dalg24, line 81: (Readability)
```suggestion
  ::Kokkos::Min<index_type> reducer(red_result);
```

algorithms/src/std_algorithms/impl/Kokkos_IsSortedUntil.hpp
```diff
@@ -73,40 +70,31 @@ IteratorType is_sorted_until_impl(const std::string& label,
...
+  ::Kokkos::parallel_reduce(
+      label,
+      // use num_elements-1 because each index handles i and i+1
+      RangePolicy<ExecutionSpace>(ex, 0, num_elements - 1),
+      // use CTAD
+      StdIsSortedUntilFunctor(first, comp, reducer), reducer);
+
+  /* If the reduction result is equal to the initial value,
+     it means the range is sorted until the end */
+  if (red_result == red_result_init) {
```
dalg24, line 91:
```suggestion
  index_type red_result_init;
  reducer.init(red_result_init);
  if (red_result == red_result_init) {
```

algorithms/src/std_algorithms/impl/Kokkos_IsSortedUntil.hpp
```diff
@@ -73,40 +70,31 @@ IteratorType is_sorted_until_impl(const std::string& label,
...
+    return last;
+  } else {
+    /* If such an index is found, then the range is sorted until there and
+       we need to return an iterator past the element found so do +1 */
+    return first + (red_result + 1);
```
dalg24, line 96: I was going to ask in the functor why you call `m_reducer.join(i)` instead of `m_reducer.join(i+1)` but I suppose it makes sense now that I read that comment.

algorithms/src/std_algorithms/impl/Kokkos_IsSortedUntil.hpp
```diff
@@ -28,33 +28,30 @@ namespace Kokkos {
 namespace Experimental {
 namespace Impl {
 
-template <class IteratorType, class IndicatorViewType, class ComparatorType>
+template <class IteratorType, class ComparatorType, class ReducerType>
 struct StdIsSortedUntilFunctor {
-  using index_type = typename IteratorType::difference_type;
+  using index_type     = typename IteratorType::difference_type;
+  using red_value_type = typename ReducerType::value_type;
```
masterleinad, line 34: I'm not a big fan of abbreviations that sound like colors. 🙂 Also, `FunctorAnalysis` would try using a `value_type` alias instead of deducing it from the functors join operator but that doesn't quite apply here since we have a reducer anyway.
```suggestion
  using value_type = typename ReducerType::value_type;
```
Feel free to ignore.

algorithms/src/std_algorithms/impl/Kokkos_IsSortedUntil.hpp
```diff
@@ -73,40 +70,31 @@ IteratorType is_sorted_until_impl(const std::string& label,
...
-      KE::find(ex, KE::cbegin(indicator), KE::cend(indicator), sentinel_value);
-  const auto shift = r - ::Kokkos::Experimental::cbegin(indicator);
-
-  return first + (shift + 1);
+    Do a par_reduce computing the *min* index that breaks the sorting.
+    If such an index is found, then the range is sorted until that element.
+    If no such index is found, then the range is sorted until the end.
+  */
+  using index_type = typename IteratorType::difference_type;
+  index_type red_result;
```
masterleinad, line 78: Same here, I would prefer
```suggestion
  index_type reducer_result;
```
or just
```suggestion
  index_type result;
```

fnrizzi:
thanks @dalg24 and @masterleinad for the comments

algorithms/src/std_algorithms/impl/Kokkos_IsSortedUntil.hpp
fnrizzi: next time i will use animal names :)

masterleinad (approved):

dalg24 (approved):

PR #6103, opened 2023-05-04
Allow copying between DynamicView and View in different memory spaces

### LINKED ISSUE #6039: Bad `dynamic_view` container unit test
#5730 introduced a test that is particularly aggravating.
It pollutes the test logs and prints (to the standard error stream :@) messages that could lead to think that the test is silently failing. This gets in the way when grepping for "fail" in the logs.
```
29: [ RUN      ] cuda.dynamic_view
29: Copy from on-device DynamicView to on-host View failed:
29: deep_copy given views that would require a temporary allocation
29: Copy from on-host View to on-device DynamicView failed:
29: deep_copy given views that would require a temporary allocation
29: Copy from on-device DynamicView to on-host View failed:
29: deep_copy given views that would require a temporary allocation
29: Copy from on-host View to on-device DynamicView failed:
29: deep_copy given views that would require a temporary allocation
29: Copy from on-device DynamicView to on-host View failed:
29: deep_copy given views that would require a temporary allocation
29: Copy from on-host View to on-device DynamicView failed:
29: deep_copy given views that would require a temporary allocation
29: Copy from on-device DynamicView to on-host View failed:
29: deep_copy given views that would require a temporary allocation
29: Copy from on-host View to on-device DynamicView failed:
29: deep_copy given views that would require a temporary allocation
29: Copy from on-device DynamicView to on-host View failed:
29: deep_copy given views that would require a temporary allocation
29: Copy from on-host View to on-device DynamicView failed:
29: deep_copy given views that would require a temporary allocation
29: Copy from on-device DynamicView to on-host View failed:
29: deep_copy given views that would require a temporary allocation
29: Copy from on-host View to on-device DynamicView failed:
29: deep_copy given views that would require a temporary allocation
29: Copy from on-device DynamicView to on-host View failed:
29: deep_copy given views that would require a temporary allocation
29: Copy from on-host View to on-device DynamicView failed:
29: deep_copy given views that would require a temporary allocation
29: Copy from on-device DynamicView to on-host View failed:
29: deep_copy given views that would require a temporary allocation
29: Copy from on-host View to on-device DynamicView failed:
29: deep_copy given views that would require a temporary allocation
29: Copy from on-device DynamicView to on-host View failed:
29: deep_copy given views that would require a temporary allocation
29: Copy from on-host View to on-device DynamicView failed:
29: deep_copy given views that would require a temporary allocation
29: Copy from on-device DynamicView to on-host View failed:
29: deep_copy given views that would require a temporary allocation
29: Copy from on-host View to on-device DynamicView failed:
29: deep_copy given views that would require a temporary allocation
29: [       OK ] cuda.dynamic_view (19 ms)
```
The offending code is
https://github.com/kokkos/kokkos/blob/715a6ffd53db387b25c6c408dd98c57790c58176/containers/unit_tests/TestDynamicView.hpp#L272-L290
If the intent is to make sure the code throws, GTest has some macros to do that.  If the intent is to check the error message then they are sting comparison facilities.  Redirecting the error message towards stderr (or stdout) is out of question.

Related to #4463

masterleinad:
Fixes #6039. We simply create a temporary allocation for the `View` if necessary.

### DIFF
--- a/containers/src/Kokkos_DynamicView.hpp
+++ b/containers/src/Kokkos_DynamicView.hpp
@@ -870,8 +870,11 @@ inline void deep_copy(const View<T, DP...>& dst,
     Kokkos::Impl::ViewRemap<dst_type, src_type>(dst, src);
     Kokkos::fence("Kokkos::deep_copy(DynamicView)");
   } else {
-    Kokkos::Impl::throw_runtime_exception(
-        "deep_copy given views that would require a temporary allocation");
+    auto dst_mirror = Kokkos::create_mirror_view(
+        Kokkos::view_alloc(Kokkos::WithoutInitializing, src_memory_space{}),
+        dst);
+    Kokkos::Impl::ViewRemap<decltype(dst_mirror), src_type>(dst_mirror, src);
+    Kokkos::deep_copy(dst, dst_mirror);
   }
 }
 
@@ -896,8 +899,10 @@ inline void deep_copy(const Kokkos::Experimental::DynamicView<T, DP...>& dst,
     Kokkos::Impl::ViewRemap<dst_type, src_type>(dst, src);
     Kokkos::fence("Kokkos::deep_copy(DynamicView)");
   } else {
-    Kokkos::Impl::throw_runtime_exception(
-        "deep_copy given views that would require a temporary allocation");
+    auto src_mirror = Kokkos::create_mirror_view_and_copy(
+        typename dst_type::memory_space{}, src);
+    Kokkos::Impl::ViewRemap<dst_type, decltype(src_mirror)>(dst, src_mirror);
+    Kokkos::fence("Kokkos::deep_copy(DynamicView)");
   }
 }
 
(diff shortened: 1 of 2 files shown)


crtrott:
I think this should not compile i.e. static_assert and the test shouldn't try to do this. Clearly it was intended behavior to NOT allow deep_copy. However, we may need to do that via deprecation. I.e. for now keep throwing, but if deprecated code 4 is off static_assert.

dalg24 (changes requested):
I think the right thing to do is to fix the test.

If we go the changing the `deep_copy()` overload route, then I think we should static_assert that it is legal to call it.

crtrott:
Checked on Kokkos::View LayoutLeft/LayoutRight conversion: it also throws and doesn't assert. I think we can consider changing this to asserts (via deprecation)  but don't think its a high priority. And if we change deep_copy behavior we need to do it in a consistent way.

PR #6108, opened 2023-05-05
Drop pointless `Kokkos{Algorithms,Containers}_config.h` files

dalg24:
They are never included, not copied over to the install directory. They also have no counterpart in the hand-rolled makefiles.

Following up on #6104


masterleinad (approved):

fnrizzi (approved):

dalg24:
Ignoring HIP builds timing out.

PR #6109, opened 2023-05-06
SIMD: add shift ops for all int types

ldh4:
Related to #5674 

This PR adds following shift operators for integer simd types in each backend:
- simd operator>>(const simd& lhs, const simd& rhs)
- simd operator>>(const simd& lhs, const int rhs)
- simd operator<<(const simd& lhs, const simd& rhs)
- simd operator<<(const simd& lhs, const int rhs)

AVX2:
- Added shifts for `int32_t`, `int64_t`[^1] and `uint64_t`

AVX512:
- Added shifts for `int32_t`[^2], `uint32_t`, `int64_t`[^2] and `uint64_t`

NEON:
- Added shifts for `int32_t`, `int64_t` and `uint64_t`

[^1]: 64-bit integer shift right arithmetic intrinsics (`_mm256_srai_epi64` and `_mm256_srav_epi64`) aren't available in AVX2. Because of this, 64-bit integer shift right in AVX2 is currently set to use the fallback method that uses generator constructor.
[^2]: Shift operators are modified to non-member functions to be consistent with [TS extensions for parallelism v2](https://en.cppreference.com/w/cpp/experimental/simd/simd)

### DIFF
--- a/simd/src/Kokkos_SIMD_AVX2.hpp
+++ b/simd/src/Kokkos_SIMD_AVX2.hpp
@@ -661,7 +661,7 @@ class simd<std::int32_t, simd_abi::avx2_fixed_size<4>> {
       _mm_add_epi32(static_cast<__m128i>(lhs), static_cast<__m128i>(rhs)));
 }
 
-KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
+[[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
 simd<std::int32_t, simd_abi::avx2_fixed_size<4>> operator>>(
     simd<std::int32_t, simd_abi::avx2_fixed_size<4>> const& lhs,
     int rhs) noexcept {
@@ -669,7 +669,7 @@ simd<std::int32_t, simd_abi::avx2_fixed_size<4>> operator>>(
       _mm_srai_epi32(static_cast<__m128i>(lhs), rhs));
 }
 
-KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
+[[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
 simd<std::int32_t, simd_abi::avx2_fixed_size<4>> operator>>(
     simd<std::int32_t, simd_abi::avx2_fixed_size<4>> const& lhs,
     simd<std::int32_t, simd_abi::avx2_fixed_size<4>> const& rhs) noexcept {
@@ -677,7 +677,7 @@ simd<std::int32_t, simd_abi::avx2_fixed_size<4>> operator>>(
       _mm_srav_epi32(static_cast<__m128i>(lhs), static_cast<__m128i>(rhs)));
 }
 
-KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
+[[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
 simd<std::int32_t, simd_abi::avx2_fixed_size<4>> operator<<(
     simd<std::int32_t, simd_abi::avx2_fixed_size<4>> const& lhs,
     int rhs) noexcept {
@@ -685,7 +685,7 @@ simd<std::int32_t, simd_abi::avx2_fixed_size<4>> operator<<(
       _mm_slli_epi32(static_cast<__m128i>(lhs), rhs));
 }
 
-KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
+[[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
 simd<std::int32_t, simd_abi::avx2_fixed_size<4>> operator<<(
     simd<std::int32_t, simd_abi::avx2_fixed_size<4>> const& lhs,
     simd<std::int32_t, simd_abi::avx2_fixed_size<4>> const& rhs) noexcept {
@@ -693,7 +693,7 @@ simd<std::int32_t, simd_abi::avx2_fixed_size<4>> operator<<(
       _mm_sllv_epi32(static_cast<__m128i>(lhs), static_cast<__m128i>(rhs)));
 }
 
-KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
+[[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
 simd<std::int32_t, simd_abi::avx2_fixed_size<4>> abs(
     simd<std::int32_t, simd_abi::avx2_fixed_size<4>> const& a) {
   __m128i const rhs = static_cast<__m128i>(a);
@@ -824,7 +824,7 @@ class simd<std::int64_t, simd_abi::avx2_fixed_size<4>> {
 }
 
 // Shift right arithmetic for 64bit packed ints is not availalbe in AVX2
-// KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
+// [[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
 // simd<std::int64_t, simd_abi::avx2_fixed_size<4>> operator>>(
 //     simd<std::int64_t, simd_abi::avx2_fixed_size<4>> const& lhs,
 //     int rhs) noexcept {
@@ -832,7 +832,7 @@ class simd<std::int64_t, simd_abi::avx2_fixed_size<4>> {
 //       _mm256_srai_epi64(static_cast<__m256i>(lhs), rhs));
 // }
 
-// KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
+// [[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
 // simd<std::int64_t, simd_abi::avx2_fixed_size<4>> operator>>(
 //     simd<std::int64_t, simd_abi::avx2_fixed_size<4>> const& lhs,
 //     simd<std::int64_t, simd_abi::avx2_fixed_size<4>> const& rhs) noexcept {
@@ -841,7 +841,7 @@ class simd<std::int64_t, simd_abi::avx2_fixed_size<4>> {
 //       _mm256_cvtepi32_epi64(static_cast<__m128i>(static_cast<__m128i>(rhs))));
 // }
 
-KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
+[[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
 simd<std::int64_t, simd_abi::avx2_fixed_size<4>> operator<<(
     simd<std::int64_t, simd_abi::avx2_fixed_size<4>> const& lhs,
     int rhs) noexcept {
@@ -849,7 +849,7 @@ simd<std::int64_t, simd_abi::avx2_fixed_size<4>> operator<<(
       _mm256_slli_epi64(static_cast<__m256i>(lhs), rhs));
 }
 
-KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
+[[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
 simd<std::int64_t, simd_abi::avx2_fixed_size<4>> operator<<(
     simd<std::int64_t, simd_abi::avx2_fixed_size<4>> const& lhs,
     simd<std::int32_t, simd_abi::avx2_fixed_size<4>> const& rhs) noexcept {
@@ -860,14 +860,14 @@ simd<std::int64_t, simd_abi::avx2_fixed_size<4>> operator<<(
 
 // Manually computing absolute values, because _mm256_abs_epi64
 // is not in AVX2; it's available in AVX512.
-KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
+[[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
 simd<std::int64_t, simd_abi::avx2_fixed_size<4>> abs(
     simd<std::int64_t, simd_abi::avx2_fixed_size<4>> const& a) {
   return simd<std::int64_t, simd_abi::avx2_fixed_size<4>>(
       [&](std::size_t i) { return (a[i] < 0) ? -a[i] : a[i]; });
 }
 
-KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
+[[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
 simd<std::int64_t, simd_abi::avx2_fixed_size<4>> condition(
     simd_mask<std::int64_t, simd_abi::avx2_fixed_size<4>> const& a,
     simd<std::int64_t, simd_abi::avx2_fixed_size<4>> const& b,
@@ -974,7 +974,7 @@ simd<std::int64_t, simd_abi::avx2_fixed_size<4>>::simd(
       _mm256_sub_epi64(static_cast<__m256i>(lhs), static_cast<__m256i>(rhs)));
 }
 
-KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
+[[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
 simd<std::uint64_t, simd_abi::avx2_fixed_size<4>> operator>>(
     simd<std::uint64_t, simd_abi::avx2_fixed_size<4>> const& lhs,
     int rhs) noexcept {
@@ -982,7 +982,7 @@ simd<std::uint64_t, simd_abi::avx2_fixed_size<4>> operator>>(
       _mm256_srli_epi64(static_cast<__m256i>(lhs), rhs));
 }
 
-KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
+[[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
 simd<std::uint64_t, simd_abi::avx2_fixed_size<4>> operator>>(
     simd<std::uint64_t, simd_abi::avx2_fixed_size<4>> const& lhs,
     simd<std::uint64_t, simd_abi::avx2_fixed_size<4>> const& rhs) noexcept {
@@ -990,7 +990,7 @@ simd<std::uint64_t, simd_abi::avx2_fixed_size<4>> operator>>(
       _mm256_srlv_epi64(static_cast<__m256i>(lhs), static_cast<__m256i>(rhs)));
 }
 
-KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
+[[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
 simd<std::uint64_t, simd_abi::avx2_fixed_size<4>> operator<<(
     simd<std::uint64_t, simd_abi::avx2_fixed_size<4>> const& lhs,
     int rhs) noexcept {
@@ -998,7 +998,7 @@ simd<std::uint64_t, simd_abi::avx2_fixed_size<4>> operator<<(
       _mm256_slli_epi64(static_cast<__m256i>(lhs), rhs));
 }
 
-KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
+[[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
 simd<std::uint64_t, simd_abi::avx2_fixed_size<4>> operator<<(
     simd<std::uint64_t, simd_abi::avx2_fixed_size<4>> const& lhs,
     simd<std::uint64_t, simd_abi::avx2_fixed_size<4>> const& rhs) noexcept {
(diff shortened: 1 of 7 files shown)


crtrott:
@ldh4 I assume we are waiting here for the other guy #5913 to merge first?

ldh4:
> @ldh4 I assume we are waiting here for the other guy #5913 to merge first?

@crtrott Yes, while nothing being added in this PR functionally requires #5913, the testing structure added in #5913 is partially used to test newly added shift operations in each non-double simd types, which are otherwise untested.

ldh4:
Retest this please.

simd/unit_tests/TestSIMD.cpp
```diff
@@ -277,6 +277,30 @@ class divides {
...
+ public:
+  template <typename T>
+  auto on_host(T&& a, unsigned int b) const {
+    return a >> b;
+  }
+  template <typename T>
+  KOKKOS_INLINE_FUNCTION auto on_device(T& a, unsigned int b) const {
+    return a >> b;
+  }
+};
```
dalg24, line 290: Where do we test when the rhs is a simd type?

simd/src/Kokkos_SIMD_AVX2.hpp
```diff
@@ -784,6 +813,39 @@ class simd<std::int64_t, simd_abi::avx2_fixed_size<4>> {
...
+  return simd<std::int64_t, simd_abi::avx2_fixed_size<4>>(
+      _mm256_slli_epi64(static_cast<__m256i>(lhs), rhs));
+}
+KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
+simd<std::int64_t, simd_abi::avx2_fixed_size<4>> operator<<(
+    simd<std::int64_t, simd_abi::avx2_fixed_size<4>> const& lhs,
+    simd<int, simd_abi::avx2_fixed_size<4>> const& rhs) noexcept {
+  return simd<std::int64_t, simd_abi::avx2_fixed_size<4>>(
+      _mm256_sllv_epi64(static_cast<__m256i>(lhs),
+                        _mm256_cvtepi32_epi64(static_cast<__m128i>(rhs))));
```
dalg24, line 846: Please explain why we need this.

simd/src/Kokkos_SIMD_AVX2.hpp
```diff
@@ -661,6 +661,35 @@ class simd<std::int32_t, simd_abi::avx2_fixed_size<4>> {
...
+    int rhs) noexcept {
+  return simd<std::int32_t, simd_abi::avx2_fixed_size<4>>(
+      _mm_srai_epi32(static_cast<__m128i>(lhs), rhs));
+}
+KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
+simd<std::int32_t, simd_abi::avx2_fixed_size<4>> operator>>(
+    simd<std::int32_t, simd_abi::avx2_fixed_size<4>> const& lhs,
+    simd<int, simd_abi::avx2_fixed_size<4>> const& rhs) noexcept {
+  return simd<std::int32_t, simd_abi::avx2_fixed_size<4>>(
+      _mm_srav_epi32(static_cast<__m128i>(lhs), static_cast<__m128i>(rhs)));
```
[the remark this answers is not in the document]
ldh4: Because this is for a signed type, I think we would want to use arithmetic right shift to keep shifted negative values still negative. Logical shift wouldn't do this since it fills in the leftmost bits with 0's.

simd/src/Kokkos_SIMD_AVX2.hpp
```diff
@@ -784,6 +813,39 @@ class simd<std::int64_t, simd_abi::avx2_fixed_size<4>> {
       _mm256_add_epi64(static_cast<__m256i>(lhs), static_cast<__m256i>(rhs)));
 }
 
+// Shift right arithmetic for 64bit packed ints is not availalbe in AVX2
```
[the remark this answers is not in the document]
ldh4: Same reason as above.
It seems that the behavior of shift right operation is actually left as "implementation-defined" according to [cpp reference](https://en.cppreference.com/w/cpp/language/operator_arithmetic), but I think changing the sign of a negative value after a shift right op would be unintuitive to use.

simd/src/Kokkos_SIMD_AVX2.hpp
dalg24: Sorry I meant the `_mm256_cvtepi32_epi64(static_cast<__m128i>(rhs))`

simd/src/Kokkos_SIMD_Common.hpp
```diff
@@ -129,10 +129,10 @@ template <class T, class Abi>
 }
 
 // fallback simd shift using generator constructor
-// At the time of this writing, these fallbacks are only used
-// to shift vectors of 64-bit unsigned integers for the NEON backend
+// At the time of this edit, only the fallback for shift vectors of
+// 64-bit signed integers for the AVX2 backend is used
 
-template <class T, class U, class Abi>
+template <class T, class Abi>
```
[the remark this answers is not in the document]
ldh4: Added `static_cast` to each shift operators that checks for T being an integral type in [2aa01ef](https://github.com/kokkos/kokkos/pull/6109/commits/2aa01ef5b992084bda24cb977cb3f485b26de0ec).

simd/unit_tests/TestSIMD.cpp
ldh4: Added test cases that exercise shift ops with simd type rhs in [2aa01ef](https://github.com/kokkos/kokkos/pull/6109/commits/2aa01ef5b992084bda24cb977cb3f485b26de0ec).
Device side ones added in [775b605](https://github.com/kokkos/kokkos/pull/6109/commits/775b6056351fb7edaee60fdae6ff951b94591378).

simd/src/Kokkos_SIMD_AVX2.hpp
ldh4: `_mm256_sllv_epi64` takes in ` __m256i` as its arguments. Because `rhs` is `int32_t` simd typed (`__m128i`), it needs to be extended to 64-bit int simd type. So `_mm256_cvtepi32_epi64` is used to sign extend `rhs` to `__m256i`.

I think the function signature that expects `rhs` as `simd<int32_t, abi>` could be up for a discussion for a change. [Kokkos document](https://kokkos.github.io/kokkos-core-wiki/API/simd/simd.html) has `simd operator>>(const simd& lhs, const simd& rhs)` for the operator, and I think the type of `rhs` may be ambiguous as is.

ldh4:
Retest this please.

simd/src/Kokkos_SIMD_AVX2.hpp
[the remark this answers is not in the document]
crtrott: yeah: the standard says apply the operator. And << and >> preserve the sign when shifting negative values.

simd/src/Kokkos_SIMD_AVX2.hpp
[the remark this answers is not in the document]
crtrott: no its not implementation defined: the wording in https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/p1928r4.pdf says apply the operator to each element. And it looks like << and >> do arithmetic shifts (they keep the sign. But we should check corner cases such as: `-1 >> 1 == -1`

simd/src/Kokkos_SIMD_AVX2.hpp
crtrott: I am pretty sure they just want the same operator on both sides? https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/p1928r4.pdf does not have mixed operator versions. i.e. return type, lhs and rhs are all the same. Now you probably still can pass another thing in and rely on conversion. but it does not explicitly support it. Should we? @ldh4 @ibaned @dalg24 So essentially this should be a signature 
```
simd<std::int64_t, simd_abi::avx2_fixed_size<4>> operator<<(
    simd<std::int64_t, simd_abi::avx2_fixed_size<4>> const& lhs,
    simd<std::int64_t, simd_abi::avx2_fixed_size<4>> const& rhs)
```

simd/src/Kokkos_SIMD_AVX2.hpp
```diff
@@ -925,6 +974,40 @@ simd<std::int64_t, simd_abi::avx2_fixed_size<4>>::simd(
...
+}
+
+KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
+simd<std::uint64_t, simd_abi::avx2_fixed_size<4>> operator>>(
+    simd<std::uint64_t, simd_abi::avx2_fixed_size<4>> const& lhs,
+    simd<std::int32_t, simd_abi::avx2_fixed_size<4>> const& rhs) noexcept {
+  return simd<std::uint64_t, simd_abi::avx2_fixed_size<4>>(
+      _mm256_srlv_epi64(static_cast<__m256i>(lhs),
+                        _mm256_cvtepi32_epi64(static_cast<__m128i>(rhs))));
+}
```
crtrott, line 992: same question as above

simd/src/Kokkos_SIMD_AVX2.hpp
```diff
@@ -925,6 +974,40 @@ simd<std::int64_t, simd_abi::avx2_fixed_size<4>>::simd(
...
+
+KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
+simd<std::uint64_t, simd_abi::avx2_fixed_size<4>> operator<<(
+    simd<std::uint64_t, simd_abi::avx2_fixed_size<4>> const& lhs,
+    simd<std::int32_t, simd_abi::avx2_fixed_size<4>> const& rhs) noexcept {
+  return simd<std::uint64_t, simd_abi::avx2_fixed_size<4>>(
+      _mm256_sllv_epi64(static_cast<__m256i>(lhs),
+                        _mm256_cvtepi32_epi64(static_cast<__m128i>(rhs))));
+}
```
crtrott, line 1009: same as above

crtrott (changes requested):
We are deviating from P1928 the simd operator<<(simd, simd) things take the same simd thing for both arguments, not int32_t as the shift argument.

simd/src/Kokkos_SIMD_AVX2.hpp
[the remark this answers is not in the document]
ldh4: Added a short test case that checks `-1 >>1 == -1` for a quick sanity check.
The test case for having a lane in lhs be `-1` and a lane in rhs be `1` is already part of testing set.

simd/src/Kokkos_SIMD_AVX2.hpp
ldh4: Modified shift operators so that its return type, `lhs` and `rhs` be the same simd type in [63c8055](https://github.com/kokkos/kokkos/pull/6109/commits/63c80558527bb520f6b2fb6c5ae605c94b8c362d).
Shift operators that take in an int as `rhs` still remain.

simd/src/Kokkos_SIMD_AVX2.hpp
ldh4: Changed in [63c8055](https://github.com/kokkos/kokkos/pull/6109/commits/63c80558527bb520f6b2fb6c5ae605c94b8c362d).

simd/src/Kokkos_SIMD_AVX2.hpp
ldh4: Changed in [63c8055](https://github.com/kokkos/kokkos/pull/6109/commits/63c80558527bb520f6b2fb6c5ae605c94b8c362d).

simd/src/Kokkos_SIMD_AVX512.hpp
```diff
@@ -355,6 +384,35 @@ simd<std::uint32_t, simd_abi::avx512_fixed_size<8>> abs(
...
+  return simd<std::uint32_t, simd_abi::avx512_fixed_size<8>>(
+      _mm256_srli_epi32(static_cast<__m256i>(lhs), rhs));
+}
+KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
+simd<std::uint32_t, simd_abi::avx512_fixed_size<8>> operator>>(
+    simd<std::uint32_t, simd_abi::avx512_fixed_size<8>> const& lhs,
+    simd<std::int32_t, simd_abi::avx512_fixed_size<8>> const& rhs) noexcept {
+  return simd<std::uint32_t, simd_abi::avx512_fixed_size<8>>(
+      _mm256_srlv_epi32(static_cast<__m256i>(lhs), static_cast<__m256i>(rhs)));
+}
```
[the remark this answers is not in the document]
ldh4: Changed in [63c8055](https://github.com/kokkos/kokkos/pull/6109/commits/63c80558527bb520f6b2fb6c5ae605c94b8c362d).

simd/unit_tests/TestSIMD.cpp
```diff
@@ -424,11 +476,114 @@ inline void host_check_conversions() {
...
+    std::size_t constexpr width     = simd_type::size();
+    std::size_t constexpr num_cases = 8;
+
+    DataType max = std::numeric_limits<DataType>::max();
+    DataType min = std::numeric_limits<DataType>::min();
+
+    const int shift_by[num_cases] = {
+        0, 1, 3, width / 2, width / 2 + 1, width - 1, width, width + 1};
+    DataType test_vals[width];
+
```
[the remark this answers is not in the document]
ldh4: Added a short test case that checks -1 >>1 == -1 for a quick sanity check in [ee913fa](https://github.com/kokkos/kokkos/pull/6109/commits/ee913fa2fce9a8f71411ea10eda21088de837c16).

More general test case for having a lane in lhs be -1 and a lane in rhs be 1 is already part of testing set.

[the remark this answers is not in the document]
crtrott: I saw the -1 and 1 as part of the math ops check but not the shift?

This is what I am seeing for shift:
```c++
    for (std::size_t i = 0; i < width; ++i) {
      DataType test_val = max / (i + 1);
      test_vals[i]      = (test_val < min) ? min : test_val;
    }
    device_check_shift_op_all_loaders<Abi>(shift_right(), test_vals, shift_by,
                                           num_cases);

    for (std::size_t i = 0; i < width; ++i) {
      DataType test_val = (min == 0) ? 1 : min / (i + 1);
      test_vals[i]      = (test_val > max) ? max : test_val;
    }
    device_check_shift_op_all_loaders<Abi>(shift_left(), test_vals, shift_by,
                                           num_cases);
```

To keep it simple for `char` and a width of 4 wouldn't this lead to only: 127, 63, 31, 15, -128, -64, -32 ,-16 being tested? So neither 1 nor -1?

simd/src/Kokkos_SIMD_AVX2.hpp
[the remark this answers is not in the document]
crtrott: its good to have the sanity check for the scalar but I don't think you test -1 and 1 for shift, only for math ops or?

simd/src/Kokkos_SIMD_AVX2.hpp
```diff
@@ -791,6 +823,41 @@ class simd<std::int64_t, simd_abi::avx2_fixed_size<4>> {
...
+// }
+
+// KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
+// simd<std::int64_t, simd_abi::avx2_fixed_size<4>> operator>>(
+//     simd<std::int64_t, simd_abi::avx2_fixed_size<4>> const& lhs,
+//     simd<std::int64_t, simd_abi::avx2_fixed_size<4>> const& rhs) noexcept {
+//   return simd<std::int64_t, simd_abi::avx2_fixed_size<4>>(
+//       _mm256_srav_epi64(static_cast<__m256i>(lhs),
+//       _mm256_cvtepi32_epi64(static_cast<__m128i>(static_cast<__m128i>(rhs))));
+// }
```
crtrott, line 842: Can you create a follow up issue for this, we may want to add a manual thing here, potentially guarded by some perf warning or something.

simd/src/Kokkos_SIMD_AVX2.hpp
```diff
@@ -661,6 +661,38 @@ class simd<std::int32_t, simd_abi::avx2_fixed_size<4>> {
...
+    simd<std::int32_t, simd_abi::avx2_fixed_size<4>> const& lhs,
+    simd<std::int32_t, simd_abi::avx2_fixed_size<4>> const& rhs) noexcept {
+  return simd<std::int32_t, simd_abi::avx2_fixed_size<4>>(
+      _mm_srav_epi32(static_cast<__m128i>(lhs), static_cast<__m128i>(rhs)));
+}
+
+KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
+simd<std::int32_t, simd_abi::avx2_fixed_size<4>> operator<<(
+    simd<std::int32_t, simd_abi::avx2_fixed_size<4>> const& lhs,
+    int rhs) noexcept {
```
crtrott, line 683: Please create an issue (or task) where we should replace int here (and elsewhere) with a private `simd_size_type`, just so we can in later comparisons with the standard have an easier time. Note `simd_size_type` is an implementation defined signed integral, so its natural for it to be int.

crtrott (changes requested):
Almost there, but I think we are still not testing the right thing.

simd/src/Kokkos_SIMD_Scalar.hpp
```diff
@@ -113,15 +113,15 @@ class simd<T, simd_abi::scalar> {
     return simd(m_value >> rhs);
   }
   KOKKOS_FORCEINLINE_FUNCTION simd
-  operator>>(simd<int, abi_type> const& rhs) const {
-    return simd(m_value >> static_cast<int>(rhs));
+  operator>>(simd<value_type, abi_type> const& rhs) const {
```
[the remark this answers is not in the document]
ldh4: Changed in [892b7c5](https://github.com/kokkos/kokkos/pull/6109/commits/892b7c523754d0941fec7269fd7b4d5bdd2d3079).

simd/src/Kokkos_SIMD_Scalar.hpp
```diff
@@ -113,15 +113,15 @@ class simd<T, simd_abi::scalar> {
...
+    return simd(m_value >> static_cast<int>(rhs[0]));
   }
   KOKKOS_FORCEINLINE_FUNCTION simd operator<<(int rhs) const {
     return simd(m_value << rhs);
   }
   KOKKOS_FORCEINLINE_FUNCTION simd
-  operator<<(simd<int, abi_type> const& rhs) const {
-    return simd(m_value << static_cast<int>(rhs));
+  operator<<(simd<value_type, abi_type> const& rhs) const {
```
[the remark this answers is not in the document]
ldh4: Changed in [892b7c5](https://github.com/kokkos/kokkos/pull/6109/commits/892b7c523754d0941fec7269fd7b4d5bdd2d3079).

simd/src/Kokkos_SIMD_AVX2.hpp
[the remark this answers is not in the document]
ldh4: My mistake. Fixed the test values to include `-1 >> 1`.

simd/unit_tests/TestSIMD.cpp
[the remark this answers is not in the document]
ldh4: I realized that the test values don't represent what I thought they would be. For instance, for shift right op test, it wasn't testing for any negative values.
Changed this test values in [892b7c5](https://github.com/kokkos/kokkos/pull/6109/commits/892b7c523754d0941fec7269fd7b4d5bdd2d3079) and added negative value cases if the data type is a signed type.

simd/src/Kokkos_SIMD_AVX2.hpp
ldh4: Yes, in fact, I think all fallback implementations in `Kokkos_SIMD_Common.hpp` should eventually be moved to their corresponding AVX2 backend impl.

simd/src/Kokkos_SIMD_AVX2.hpp
ldh4: I'll add a task to the issue #5674 so that we maintain a list of simd to-do's in one place instead of having them all scattered.

simd/src/Kokkos_SIMD_AVX2.hpp
ldh4: Added as another task in #5674. Let me know if opening a separate issue is preferred.

crtrott: oh so there is a fallback already in place just elsewhere? That is fine for now

simd/src/Kokkos_SIMD_AVX2.hpp
crtrott: yeah

nliber (changes requested):

simd/src/Kokkos_SIMD_Common.hpp
```diff
@@ -129,30 +129,34 @@ template <class T, class Abi>
...
+  static_assert(std::is_integral_v<T>);
   return simd<T, Abi>([&](std::size_t i) { return lhs[i] << rhs; });
 }
 
-template <class T, class U, class Abi>
+template <typename T, typename Abi>
 [[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION simd<T, Abi> operator>>(
-    simd<T, Abi> const& lhs, simd<U, Abi> const& rhs) {
+    simd<T, Abi> const& lhs, simd<T, Abi> const& rhs) {
+  static_assert(std::is_integral_v<T>);
```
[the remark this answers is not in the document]
ldh4: Removed these `static_assert`s and replaced with `enable_if`s to check the type of `T` in [512ad77](https://github.com/kokkos/kokkos/pull/6109/commits/512ad777c417006e0d86f042e5b0f075e48fa6f7).

These are the fallback implementations for shift operators, but it's currently only needed for AVX2 with `int64_t`. I think this will eventually be moved to AVX2 and be explicitly specialized with `int64_t` instead.

crtrott:
Retest this please.

nliber (approved):

PR #6110, opened 2023-05-08
CUDA_LAMBDA always ON -- Fix `kokkos_check(OPTIONS CUDA_LAMBDA)`

dalg24:
This PR revisits #5964 
The 1st commit is just reverting the PR to make it easier to compare what would be the changes as compared to 4.0

The issue with #5964 is that it broke codes downstream that were checking for CUDA lambda support at configuration time.
```CMake
find_package(Kokkos)
if(Kokkos_ENABLE_CUDA)
  kokkos_check(OPTIONS CUDA_LAMBDA)
endif()
```
Projects such as ArborX, Cabana or deal.II do that for instance.

The diff looks a bit larger than it could because I dropped some of the unnecessary changes (like dropping macro guards in tests or benchmarks).


Makefile.kokkos
```diff
@@ -667,7 +667,7 @@ ifeq ($(KOKKOS_INTERNAL_USE_CUDA), 1)
 
   ifeq ($(KOKKOS_INTERNAL_COMPILER_NVCC), 1)
     tmp := $(call kokkos_append_header,"$H""define KOKKOS_ENABLE_CUDA_LAMBDA")
-    KOKKOS_CXXFLAGS += -expt-extended-lambda
+    KOKKOS_CXXFLAGS += -extended-lambda
```
dalg24, line 670: The `-extended-lambda` flag is supported for all CUDA versions that we support with NVCC.

cmake/kokkos_arch.cmake
```diff
@@ -163,23 +163,10 @@ ENDIF()
 GLOBAL_SET(KOKKOS_CUDA_OPTIONS)
 # Construct the Makefile options
 IF(KOKKOS_CXX_COMPILER_ID STREQUAL NVIDIA)
-  # Extended lambda support was stabilized in nvcc 12
-  IF(KOKKOS_COMPILER_VERSION_MAJOR EQUAL 11)
-    GLOBAL_APPEND(KOKKOS_CUDA_OPTIONS "-expt-extended-lambda")
-  ELSE()
-    GLOBAL_APPEND(KOKKOS_CUDA_OPTIONS "-extended-lambda")
-  ENDIF()
+  GLOBAL_APPEND(KOKKOS_CUDA_OPTIONS "-extended-lambda")
```
dalg24, line 166: Opted for `-extended-lambda` everywhere.  The if-condition is unnecessary here since NVCC 11+ supports that flag.

cmake/kokkos_arch.cmake
```diff
@@ -163,23 +163,10 @@ ENDIF()
...
   GLOBAL_APPEND(KOKKOS_CUDA_OPTIONS "-Wext-lambda-captures-this")
 ENDIF()
 
-IF(DEFINED Kokkos_ENABLE_CUDA_LAMBDA)
-  IF(Kokkos_ENABLE_CUDA_LAMBDA)
-    MESSAGE(DEPRECATION "CUDA extended lambda support is now always enabled. The option Kokkos_ENABLE_CUDA_LAMBDA will be removed")
-  ELSE()
-    MESSAGE(FATAL_ERROR "Support for disabling CUDA extended lambdas has been removed. Please unset Kokkos_ENABLE_CUDA_LAMBDA, or see #5964 if this is necessary for your application")
-  ENDIF()
-ENDIF()
```
dalg24, line 181: This logic was slightly modified and moved to `cmake/kokkos_enable_options.cmake` closer to the definition of the option.

cmake/kokkos_enable_options.cmake
```diff
@@ -67,6 +67,7 @@ mark_as_advanced(Kokkos_ENABLE_IMPL_MDSPAN)
 mark_as_advanced(Kokkos_ENABLE_MDSPAN_EXTERNAL)
 mark_as_advanced(Kokkos_ENABLE_IMPL_SKIP_COMPILER_MDSPAN)
 
+KOKKOS_ENABLE_OPTION(CUDA_LAMBDA ON "Whether to allow lambda expressions on the device with NVCC **DEPRECATED**")
```
dalg24, line 70: The docstring was updated as compared to 4.0 that had `"Whether to activate experimental lambda features"`.

cmake/kokkos_enable_options.cmake
```diff
@@ -160,6 +161,16 @@ IF(Kokkos_ENABLE_CUDA_LDG_INTRINSIC)
     MESSAGE(FATAL_ERROR "Kokkos_ENABLE_CUDA_LDG_INTRINSIC has been removed. LDG intrinsics are always enabled.")
   ENDIF()
 ENDIF()
+IF(NOT Kokkos_ENABLE_CUDA_LAMBDA)
```
dalg24, line 164: Meaning we do not warn when users enable.  I propose we keep it this way for a number of releases and get rid of it closer to 5.0

crtrott:
Not sure I agree with the reverting everything. The resolution itself is fine with me.

cmake/kokkos_enable_options.cmake
nliber: I'm not against this.

However, why wait to warn?  Is it because we expect people to switch between Kokkos 3 and 4 for a while, and we don't want to add friction?

nliber (approved):

Makefile.kokkos
fnrizzi: can we write what you said here as a comment there so that it stays as reminder for later?

PhilMiller: Unnecessary, I think, since the alternative is entirely backwards-looking in time. It's not something that's going to change or need re-consideration.

This is what writing things down in the review is for

dalg24:
@crtrott @PhilMiller my intent was to open a follow-up with the changes to the tests and benchmarks but I am ok including in this PR if you'd like me to.

nmm0 (approved):
This looks good to me, but I think we decided not to warn? I can't find it in the notes. I'm on the "let's warn side" though.

PR #6114, opened 2023-05-09
KokkosTools: Don't call callbacks before backends are initialized

masterleinad:
Fixes the issue seen in https://github.com/kokkos/kokkos-tools/pull/188#issuecomment-1532076044.
In short, the test there sets callbacks before `Kokkos` is initialized and some callbacks are used when initializing backends before the respective tool is initialized. This pull request proposes to pause tools until the backends are initialized which also implies that none of those events get recorded.

### DIFF
--- a/core/src/impl/Kokkos_Core.cpp
+++ b/core/src/impl/Kokkos_Core.cpp
@@ -792,10 +792,10 @@ void post_initialize_internal(const Kokkos::InitializationSettings& settings) {
 }
 
 void initialize_internal(const Kokkos::InitializationSettings& settings) {
-  // Tools should only be used after the backends have been initialized (because
-  // they have not been initialized at that point) but if they have been set via
-  // set_callbacks they need to be present before calling post_initialize so
-  // that they get initialized as well.
+  // The tool initialization is only called in post_initialize_internal.
+  // Pausing tools here, so that if someone has set callbacks programmatically
+  // these callbacks are not called inside the backend initialization, before
+  // the tool initialization happened.
   Kokkos::Tools::Experimental::pause_tools();
   pre_initialize_internal(settings);
   initialize_backends(settings);
--- a/core/src/impl/Kokkos_Core.cpp
+++ b/core/src/impl/Kokkos_Core.cpp
@@ -792,8 +792,14 @@ void post_initialize_internal(const Kokkos::InitializationSettings& settings) {
 }
 
 void initialize_internal(const Kokkos::InitializationSettings& settings) {
+  // Tools should only be used after the backends have been initialized (because
+  // they have not been initialized at that point) but if they have been set via
+  // set_callbacks they need to be present before calling post_initialize so
+  // that they get initialized as well.
+  Kokkos::Tools::Experimental::pause_tools();
   pre_initialize_internal(settings);
   initialize_backends(settings);
+  Kokkos::Tools::Experimental::resume_tools();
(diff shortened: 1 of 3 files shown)


dalg24:
> We need to set the initializer callback before calling `KokkosTools::initialize` anyway.

I do not understand.

masterleinad:
> I do not understand.

The tools provide an `init_library` callback that is used to initialize the tool, see, e.g., https://github.com/kokkos/kokkos-tools/blob/7ab29004cf1d9a05328b704f0717ee686f1b095b/profiling/space-time-stack/kp_space_time_stack.cpp#L802-L806. This callback must be called before any other callback the tool provides and this is (only) done in the call to `Kokkos::Profiling::initialize` which is part of `Kokkos::initialize`. Thus, the (`init_library`) callback must be set before calling `Kokkos::initialize` to have any effect.

Rombur (approved):

core/unit_test/tools/TestToolsInitialization.cpp
```diff
@@ -0,0 +1,212 @@
...
+  }
+  {
+    using execution_space = Kokkos::DefaultExecutionSpace;
+    using memory_space    = typename execution_space::memory_space;
+    Kokkos::View<int*, memory_space> src_view("source", 10);
+    Kokkos::View<int*, memory_space> dst_view("destination", 10);
+    Kokkos::deep_copy(dst_view, src_view);
+    Kokkos::parallel_for("parallel_for",
+                         Kokkos::RangePolicy<execution_space>(0, 1),
+                         [=](int i) { (void)i; });
```
Rombur, line 163:
```suggestion
                        KOKKOS_LAMBDA(int i) { (void)i; });
```

core/unit_test/tools/TestToolsInitialization.cpp
```diff
@@ -0,0 +1,212 @@
...
+    int result;
+    Kokkos::parallel_reduce(
+        "parallel_reduce", Kokkos::RangePolicy<execution_space>(0, 1),
+        [=](int i, int& hold_result) { hold_result += i; }, result);
+    Kokkos::parallel_scan("parallel_scan",
+                          Kokkos::RangePolicy<execution_space>(0, 1),
+                          [=](const int i, int& hold_result, const bool final) {
```
Rombur, line 170:
```suggestion
                          KOKKOS_LAMBDA(const int i, int& hold_result, const bool final) {
```

core/src/impl/Kokkos_Core.cpp
```diff
@@ -792,8 +792,14 @@ void post_initialize_internal(const Kokkos::InitializationSettings& settings) {
 }
 
 void initialize_internal(const Kokkos::InitializationSettings& settings) {
+  // Tools should only be used after the backends have been initialized (because
+  // they have not been initialized at that point) but if they have been set via
+  // set_callbacks they need to be present before calling post_initialize so
+  // that they get initialized as well.
```
crtrott, line 798:
```suggestion
  // The tool initialization is only called in post_initialize_internal.
  // Pausing tools here, so that if someone has set callbacks programmatically
  // these callbacks are not called inside the backend initialization, before
  // the tool initialization happened.
```

crtrott (commented):
Proposed cleaner comment

dalg24 (approved):

dalg24:
Plz make sure you add to the changelog

PR #6116, opened 2023-05-09
hpcbind: check for correct Slurm variable

rbberger:
`hpcbind` was checking for `SLURM_LOCAL_ID` which does not exist. It therefore defaulted to `HPCBIND_QUEUE_RANK=0` (e.g. leading to only using GPU 0)

The correct variable name is `SLURM_LOCALID`. See https://slurm.schedmd.com/srun.html#SECTION_OUTPUT-ENVIRONMENT-VARIABLES


dalg24 (approved):

dalg24:
(no need to trigger tests on the ORNL Jenkins server for this patch)

Rombur (approved):

PR #6118, opened 2023-05-10
Fix nightlies -- workaround compiler bug in GCC 9.1 and 9.2

dalg24:
Resolve #6107 

I was able to reproduce with the `gcc:9.2.0` docker image and did
```
wget https://github.com/Kitware/CMake/releases/download/v3.16.8/cmake-3.16.8-Linux-x86_64.sh
mkdir -p /opt/cmake
sh cmake-3.16.8-Linux-x86_64.sh --skip-license --prefix=/opt/cmake
git clone https://github.com/kokkos/kokkos.git --branch develop --depth 1
cd kokkos/
/opt/cmake/bin/cmake -B build -DKokkos_ENABLE_OPENMP=ON -DKokkos_ENABLE_TESTS=ON
/opt/cmake/bin/cmake --build build --target Kokkos_CoreUnitTest_OpenMP --parallel 6
```
I checked that `gcc:9.3.0` did not need the workaround.

PR #6119, opened 2023-05-10
Kokkos_BitManipulation: KOKKOS_COMPILER_GCC->KOKKOS_COMPILER_GNU

masterleinad:
Found while looking at #6118.
`KOKKOS_COMPILER_GCC` doesn't exist, we use `KOKKOS_COMPILER_GNU` for compilers similar to `g++`.

### DIFF
--- a/core/src/Kokkos_BitManipulation.hpp
+++ b/core/src/Kokkos_BitManipulation.hpp
@@ -259,12 +259,13 @@ KOKKOS_IMPL_HOST_FUNCTION T byteswap_builtin_host(T x) noexcept {
   } else if constexpr (sizeof(T) == 8) {
     return __builtin_bswap64(x);
   } else if constexpr (sizeof(T) == 16) {
+#if defined(__has_builtin)
 #if __has_builtin(__builtin_bswap128)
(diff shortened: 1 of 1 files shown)


PhilMiller (approved):

core/src/Kokkos_BitManipulation.hpp
```diff
@@ -259,7 +259,7 @@ KOKKOS_IMPL_HOST_FUNCTION T byteswap_builtin_host(T x) noexcept {
   } else if constexpr (sizeof(T) == 8) {
     return __builtin_bswap64(x);
   } else if constexpr (sizeof(T) == 16) {
-#if __has_builtin(__builtin_bswap128)
+#if defined (__has_builtin) && __has_builtin(__builtin_bswap128)
```
dalg24, line 262: I don't think that would work, all of it would get evaluated

masterleinad: Seems to work now.

dalg24 (approved):

dalg24:
Not waiting on HIP builds that will likely timeout looking at the build queue on Jenkins.

PR #6120, opened 2023-05-10
dual view: update template types + `Kokkos::is_dual_view` (#6085)

### LINKED ISSUE #6085: `Kokkos::DualView` is still on `<DataType, class, class, class>` instead of `<DataType, Properties...>`
**Describe the bug**

It seems that `Kokkos::DualView` is still templated on
```c++
<class DataType, class Arg1Type = void, class Arg2Type = void, class Arg3Type = void>
```
whereas other containers like `OffsetView` are already following the design of `Kokkos::View` with:
```c++
```

Here are the changes needed to make `DualView` more consistent if you think it's the right thing to do.
I can do a PR if need be.

```patch
@@ -86,22 +86,21 @@ inline const Kokkos::Cuda& get_cuda_space(const NonCudaExecSpace&) {
 #endif  // KOKKOS_ENABLE_CUDA
 
 }  // namespace Impl
-template <class DataType, class Arg1Type = void, class Arg2Type = void,
-          class Arg3Type = void>
-class DualView : public ViewTraits<DataType, Arg1Type, Arg2Type, Arg3Type> {
-  template <class, class, class, class>
+class DualView : public ViewTraits<DataType, Properties...> {
+  template <class, class...>
   friend class DualView;
 
  public:
   //! \name Typedefs for device types and various Kokkos::View specializations.
   //@{
-  using traits = ViewTraits<DataType, Arg1Type, Arg2Type, Arg3Type>;
+  using traits = ViewTraits<DataType, Properties...>;
 
   //! The Kokkos Host Device type;
   using host_mirror_space = typename traits::host_mirror_space;
 
   //! The type of a Kokkos::View on the device.
-  using t_dev = View<typename traits::data_type, Arg1Type, Arg2Type, Arg3Type>;
+  using t_dev = View<typename traits::data_type, traits>;
 
   /// \typedef t_host
   /// \brief The type of a Kokkos::View host mirror of \c t_dev.
@@ -110,7 +109,7 @@ class DualView : public ViewTraits<DataType, Arg1Type, Arg2Type, Arg3Type> {
   //! The type of a const View on the device.
   //! The type of a Kokkos::View on the device.
   using t_dev_const =
-      View<typename traits::const_data_type, Arg1Type, Arg2Type, Arg3Type>;
+      View<typename traits::const_data_type, traits>;
 
   /// \typedef t_host_const
   /// \brief The type of a const View host mirror of \c t_dev_const.
```

romintomasetti:
- This PR closes #6085.
- It introduces `Kokkos::is_dual_view` trait and `Kokkos::is_dual_view_v` helper variable template for consistency with other `Kokkos` containers.

### DIFF
--- a/containers/src/Kokkos_DualView.hpp
+++ b/containers/src/Kokkos_DualView.hpp
@@ -86,22 +86,37 @@ inline const Kokkos::Cuda& get_cuda_space(const NonCudaExecSpace&) {
 #endif  // KOKKOS_ENABLE_CUDA
 
 }  // namespace Impl
-template <class DataType, class Arg1Type = void, class Arg2Type = void,
-          class Arg3Type = void>
-class DualView : public ViewTraits<DataType, Arg1Type, Arg2Type, Arg3Type> {
-  template <class, class, class, class>
+
+template <class DataType, class... Properties>
+class DualView;
+
+template <class>
+struct is_dual_view : public std::false_type {};
+
+template <class DT, class... DP>
+struct is_dual_view<DualView<DT, DP...>> : public std::true_type {};
+
+template <class DT, class... DP>
+struct is_dual_view<const DualView<DT, DP...>> : public std::true_type {};
+
+template <class T>
+inline constexpr bool is_dual_view_v = is_dual_view<T>::value;
+
+template <class DataType, class... Properties>
+class DualView : public ViewTraits<DataType, Properties...> {
+  template <class, class...>
   friend class DualView;
 
  public:
   //! \name Typedefs for device types and various Kokkos::View specializations.
   //@{
-  using traits = ViewTraits<DataType, Arg1Type, Arg2Type, Arg3Type>;
+  using traits = ViewTraits<DataType, Properties...>;
 
   //! The Kokkos Host Device type;
   using host_mirror_space = typename traits::host_mirror_space;
 
   //! The type of a Kokkos::View on the device.
-  using t_dev = View<typename traits::data_type, Arg1Type, Arg2Type, Arg3Type>;
+  using t_dev = View<typename traits::data_type, Properties...>;
 
   /// \typedef t_host
   /// \brief The type of a Kokkos::View host mirror of \c t_dev.
@@ -109,8 +124,7 @@ class DualView : public ViewTraits<DataType, Arg1Type, Arg2Type, Arg3Type> {
 
   //! The type of a const View on the device.
   //! The type of a Kokkos::View on the device.
-  using t_dev_const =
-      View<typename traits::const_data_type, Arg1Type, Arg2Type, Arg3Type>;
+  using t_dev_const = View<typename traits::const_data_type, Properties...>;
 
   /// \typedef t_host_const
   /// \brief The type of a const View host mirror of \c t_dev_const.
@@ -247,15 +261,15 @@ class DualView : public ViewTraits<DataType, Arg1Type, Arg2Type, Arg3Type> {
   }
 
   //! Copy constructor (shallow copy)
-  template <class SS, class LS, class DS, class MS>
-  DualView(const DualView<SS, LS, DS, MS>& src)
+  template <typename DT, typename... DP>
+  DualView(const DualView<DT, DP...>& src)
       : modified_flags(src.modified_flags),
         d_view(src.d_view),
         h_view(src.h_view) {}
 
   //! Subview constructor
-  template <class SD, class S1, class S2, class S3, class Arg0, class... Args>
-  DualView(const DualView<SD, S1, S2, S3>& src, const Arg0& arg0, Args... args)
+  template <class DT, class... DP, class Arg0, class... Args>
+  DualView(const DualView<DT, DP...>& src, const Arg0& arg0, Args... args)
       : modified_flags(src.modified_flags),
         d_view(Kokkos::subview(src.d_view, arg0, args...)),
         h_view(Kokkos::subview(src.h_view, arg0, args...)) {}
@@ -1142,23 +1156,24 @@ class DualView : public ViewTraits<DataType, Arg1Type, Arg2Type, Arg3Type> {
 namespace Kokkos {
 namespace Impl {
 
-template <class D, class A1, class A2, class A3, class... Args>
-struct DualViewSubview {
-  using dst_traits = typename Kokkos::Impl::ViewMapping<
-      void, Kokkos::ViewTraits<D, A1, A2, A3>, Args...>::traits_type;
+template <class V>
+struct V2DV;
 
-  using type = Kokkos::DualView<
-      typename dst_traits::data_type, typename dst_traits::array_layout,
-      typename dst_traits::device_type, typename dst_traits::memory_traits>;
+template <class D, class... P>
+struct V2DV<View<D, P...>> {
+  using type = DualView<D, P...>;
 };
-
 } /* namespace Impl */
 
-template <class D, class A1, class A2, class A3, class... Args>
-typename Impl::DualViewSubview<D, A1, A2, A3, Args...>::type subview(
-    const DualView<D, A1, A2, A3>& src, Args... args) {
-  return typename Impl::DualViewSubview<D, A1, A2, A3, Args...>::type(src,
-                                                                      args...);
+template <class DataType, class... Properties, class... Args>
+auto subview(const DualView<DataType, Properties...>& src, Args&&... args) {
+  // leverage Kokkos::View facilities to deduce the properties of the subview
+  using deduce_subview_type =
+      decltype(subview(std::declval<View<DataType, Properties...>>(),
+                       std::forward<Args>(args)...));
+  // map it back to dual view
+  return typename Impl::V2DV<deduce_subview_type>::type(
+      src, std::forward<Args>(args)...);
 }
 
 } /* namespace Kokkos */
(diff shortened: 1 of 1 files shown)


romintomasetti:
@ajpowelsnl I think this is ready for your approval to test :smile:

containers/src/Kokkos_DualView.hpp
```diff
@@ -86,22 +86,37 @@ inline const Kokkos::Cuda& get_cuda_space(const NonCudaExecSpace&) {
...
+class DualView;
+
+template <class>
+struct is_dual_view : public std::false_type {};
+
+template <class D, class... P>
+struct is_dual_view<DualView<D, P...>> : public std::true_type {};
+
+template <class D, class... P>
+struct is_dual_view<const DualView<D, P...>> : public std::true_type {};
```
dalg24, line 100: Somewhat unrelated change that is not captured in the PR description.
Must clearly state that is being added and why (consistency with other containers) so that it can be easily tracked.

containers/src/Kokkos_DualView.hpp
```diff
@@ -1142,23 +1157,26 @@ class DualView : public ViewTraits<DataType, Arg1Type, Arg2Type, Arg3Type> {
...
+  static_assert(Kokkos::is_dual_view_v<DV>);
+
   using dst_traits = typename Kokkos::Impl::ViewMapping<
-      void, Kokkos::ViewTraits<D, A1, A2, A3>, Args...>::traits_type;
+      void, typename DV::traits, Args...>::traits_type;
 
   using type = Kokkos::DualView<
       typename dst_traits::data_type, typename dst_traits::array_layout,
-      typename dst_traits::device_type, typename dst_traits::memory_traits>;
+      typename dst_traits::device_type, typename dst_traits::hooks_policy,
```
dalg24, line 1169: Comment about that change.

containers/src/Kokkos_DualView.hpp
```diff
@@ -1172,11 +1190,10 @@ namespace Kokkos {
 // Partial specialization of Kokkos::deep_copy() for DualView objects.
 //
 
-template <class DT, class DL, class DD, class DM, class ST, class SL, class SD,
-          class SM>
+template <class DT, class... DProps, class ST, class... SProps>
```
dalg24, line 1193: I'd go for DP and SP for consistency with the rest of Kokkos.  (not critical)

containers/src/Kokkos_DualView.hpp
```diff
@@ -1186,12 +1203,11 @@ void deep_copy(
   }
 }
 
-template <class ExecutionSpace, class DT, class DL, class DD, class DM,
-          class ST, class SL, class SD, class SM>
+template <class ExecutionSpace, class DT, class... DProps, class ST, class... SProps>
 void deep_copy(
     const ExecutionSpace& exec,
-    DualView<DT, DL, DD, DM> dst,  // trust me, this must not be a reference
```
dalg24, line 1193: Comment about why this is ok to delete that comment and not trust whoever wrote that.

dalg24 (commented):
(peer reviewed with Daniel)

containers/src/Kokkos_DualView.hpp
romintomasetti: At the beginning I didn't introduced `is_dual_view`. But in `struct DualViewSubviewType` it was somehow needed to ensure that `typename DV` is a dual view and not something else.

I'll update the description of the PR accordingly for sure!

containers/src/Kokkos_DualView.hpp
romintomasetti: I'm not sure if you mean commenting that in the code or in the PR...

I'll comment it here first.

The whole point of this PR is to allow a `Kokkos::DualView` to accept any `Properties...` that a `Kokkos::View` can accept. Before this PR, it was hardcoded that `DualView` only accepts 4 types:
1. `DataType`
2. `Arg1Type`
3. `Arg2Type`
4. `Arg3Type`

but there was no check enforcing the type of the last 3 (Arg1Type, Arg2Type, Arg3Type). So they could be anything that `Kokkos::ViewTraits` would accept. However, `DualViewSubview::type` would then "hardcode" that when taking a subview of a dual view, only data type, array layout, device type and memory traits would be set. If my understanding is correct, that could have lead to issues if for instance I had created my original dual view with some hooks policy and then took a subview of it because my subview would have the default hooks policy (please correct me if this isn't correct! :wink:).

However, I must admit that I'm not satisfied of my solution because I only added `hooks_policy` while `Kokkos::ViewTraits` may also accept `specialize` and others.

So my question for you @dalg24 is how I can ensure that all traits are passed to `DualViewSubviewType::type`? I guess ideally one should be able to do:
```c++
template <class DV, class... Args>
struct DualViewSubviewType {
  static_assert(Kokkos::is_dual_view_v<DV>);

  using dst_traits = typename Kokkos::Impl::ViewMapping<
      void, typename DV::traits, Args...>::traits_type;

  using type = Kokkos::DualView<dst_traits>;
};
```

where `dst_traits` is a `Kokkos::ViewTraits` that would then embed all the traits given by the user and this would ensure that all types are passed. But I agree this would not work so I wonder how one can ensure that `DualViewSubviewType::type` will always "forward" all types, even in the future when `ViewTraits` would accept some more types...

containers/src/Kokkos_DualView.hpp
romintomasetti: Well, my first feeling was that this code was wrong because it was calling `modify_host`/`modify_device` on a (shallow) copy of `dst`. But as the `modified_flags` member is a (shared) view, it seems this is still OK.

Still I was not "trusting" such a comment like *trust me, this must not be a reference* (reminds me of "Trust me I'm an engineer") because to me this should work with a reference otherwise there must be other cases where taking a dual view by reference is dangerous.

So I wanted to put my hypothesis that in fact it works with a reference to the test.

On my side, it worked for OpenMP and Cuda (without UVM).

dalg24: I am leaning towards saying your change is fine.
My point was this needs to be discussed and not swept under the rug.  It looks like the comment has been there since the initial commit in this project when the code was extracted from Sandia's Trilinos repository.
cc @crtrott

masterleinad: It was introduced in https://github.com/trilinos/trilinos/commit/0806efcf77af0df08b5c82b6e57342e08144dbf4 by @mhoemmen.

containers/src/Kokkos_DualView.hpp
dalg24: I would have tried something like
```C++
namespace Impl {

template <class V>
struct V2DV;

template <class D, class... P>
struct V2DV<View<D, P...>> {
  using type = DualView<D, P...>;
};
} /* namespace Impl */

template <class DataType, class... Properties, class... Args>
auto subview(const DualView<DataType, Properties...>& src, Args... args) {
  // leverage Kokkos::View facilities to deduce the properties of the subview
  using deduce_subview_type =
      decltype(subview(std::declval<View<DataType, Properties...>>(), args...));
  // map it back to dual view
  return typename Impl::V2DV<deduce_subview_type>::type(src, args...);
}
```
to defer the deduction to Kokkos Core.
Ping @nmm0 who introduced the view hooks and @masterleinad who recently refactored `subview(View<...>, ...)`

I creeped for "subview" in the containers unit test and the test coverage looks pretty light.

dalg24: You can use perfect forwarding if you like.

masterleinad (commented):
Looks good to me given that we discuss changing the copy in `deep_copy` to a reference some.

dalg24 (approved):
I am ok with changing to a reference in deep_copy

dalg24:
OK to test

containers/src/Kokkos_DualView.hpp
```diff
@@ -1186,12 +1201,11 @@ void deep_copy(
...
 }
 
-template <class ExecutionSpace, class DT, class DL, class DD, class DM,
-          class ST, class SL, class SD, class SM>
+template <class ExecutionSpace, class DT, class... DP, class ST, class... SP>
 void deep_copy(
     const ExecutionSpace& exec,
-    DualView<DT, DL, DD, DM> dst,  // trust me, this must not be a reference
-    const DualView<ST, SL, SD, SM>& src) {
+    DualView<DT, DP...>& dst,
```
masterleinad, line 1207: I would prefer
```suggestion
    const DualView<DT, DP...>& dst,
```
to mirror the overloads for `Kokkos::View` more closely.

romintomasetti: I'm reverting this because these functions are calling `modify_host` on `dst` that mustn't be `const`:
```bash
In file included from /__w/kokkos/kokkos/containers/unit_tests/TestDualView.hpp:25:
/__w/kokkos/kokkos/containers/unit_tests/../src/Kokkos_DualView.hpp:1194:9: error: no matching member function for call to 'modify_host'
    dst.modify_host();
    ~~~~^~~~~~~~~~~
/__w/kokkos/kokkos/containers/unit_tests/TestDualView.hpp:185:13: note: in instantiation of function template specialization 'Kokkos::deep_copy<int **, Kokkos::LayoutLeft, Kokkos::OpenMP, int **, Kokkos::LayoutLeft, Kokkos::OpenMP>' requested here
    Kokkos::deep_copy(b, a);
```

romintomasetti:
Ready to go :rocket:

PR #6121, opened 2023-05-10
Fix ambiguity for bit_cast

masterleinad:
It turns out that we are seeing ambiguities with SYCL builds since there is also `sycl::bit_cast`. In the `intel/llvm` code base (https://github.com/intel/llvm/blob/sycl/sycl/include/sycl/detail/spirv.hpp) there are unconstrained calls to `bit_cast` with templated types (that are coming from the `Kokkos` namespace in the failing case). For these unconstrained calls, ADL kicks in and the compiler both finds `Kokkos::bit_cast` and `sycl::bit_cast`.
As a workaround, this pull request imports `sycl::bit_cast` into the `Kokkos` namespace directly. It turns out that `sycl::bit_cast` is unconstrained and participates in overload resolution for all types (but there are `static_assert`s in the body of the function). Thus, we also have to disable some tests that check for matching types.

It also turns out that we have some unconstrained calls to `bit_cast` in `Kokkos` that are susceptible to the same ambiguity through ADL problem (also see https://godbolt.org/z/vc5oEeMWa). This is addressed in the second commit.

### DIFF
--- a/core/src/Kokkos_BitManipulation.hpp
+++ b/core/src/Kokkos_BitManipulation.hpp
@@ -100,6 +100,11 @@ inline constexpr bool is_standard_unsigned_integer_type_v =
 namespace Kokkos {
 
 //<editor-fold desc="[bit.cast], bit_cast">
+// FIXME_SYCL intel/llvm has unqualified calls to bit_cast which are ambiguous
+// if we declare our own bit_cast function
+#ifdef KOKKOS_ENABLE_SYCL
+using sycl::bit_cast;
+#else
 template <class To, class From>
 KOKKOS_FUNCTION std::enable_if_t<sizeof(To) == sizeof(From) &&
                                      std::is_trivially_copyable_v<To> &&
@@ -110,6 +115,7 @@ bit_cast(From const& from) noexcept {
   memcpy(&to, &from, sizeof(To));
   return to;
 }
+#endif
 //</editor-fold>
 
 //<editor-fold desc="[bit.byteswap], byteswap">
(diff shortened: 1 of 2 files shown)


core/src/Kokkos_BitManipulation.hpp
```diff
@@ -398,7 +404,7 @@ KOKKOS_FUNCTION std::enable_if_t<sizeof(To) == sizeof(From) &&
                                      std::is_trivially_copyable_v<From>,
                                  To>
 bit_cast_builtin(From const& from) noexcept {
-  return bit_cast<To>(from);  // no benefit to call the _builtin variant
+  return Kokkos::bit_cast<To>(from);  // no benefit to call the _builtin variant
```
dalg24, line 407: Please propose that change on its own because that one is a no brainer.
Add a comment that it is qualified because we don't want ADL to kick in.

masterleinad: See #6122.

nliber (changes requested):
An alternative is that ours just calls the `sycl` one (and eventually the `std` one).  This has the benefit of being constrained.

masterleinad:
> An alternative is that ours just calls the `sycl` one (and eventually the `std` one). This has the benefit of being constrained.

No, that doesn't fix the ambiguity coming from calls within the compiler.

dalg24 (approved):
With the understanding that we will restrict our importing the sycl::bit_cast when it is fixed in dpc++ and we are able to guard it

masterleinad:
Only HIP CI is failing (timing out). Everything else is passing.

Rombur (approved):

nliber (approved):

nliber:
Good point @masterleinad.

dalg24:
Ignoring HIP builds that timed out

PR #6122, opened 2023-05-10
Qualify possibly ambiguous calls to bit_cast

masterleinad:
Split from #6121.
It turns out that we have some unconstrained calls to bit_cast in Kokkos that are susceptible to ambiguity through ADL problem (also see [godbolt.org/z/vc5oEeMWa](https://godbolt.org/z/vc5oEeMWa)).

### DIFF
--- a/simd/src/Kokkos_SIMD_AVX2.hpp
+++ b/simd/src/Kokkos_SIMD_AVX2.hpp
@@ -814,8 +814,8 @@ class simd<std::uint64_t, simd_abi::avx2_fixed_size<4>> {
   template <class U, std::enable_if_t<std::is_convertible_v<U, value_type>,
                                       bool> = false>
   KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION simd(U&& value)
-      : m_value(_mm256_set1_epi64x(bit_cast<std::int64_t>(value_type(value)))) {
-  }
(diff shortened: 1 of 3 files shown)


PhilMiller (approved):

nliber (approved):

dalg24 (approved):

simd/src/Kokkos_SIMD_AVX2.hpp
```diff
@@ -814,8 +814,8 @@ class simd<std::uint64_t, simd_abi::avx2_fixed_size<4>> {
   template <class U, std::enable_if_t<std::is_convertible_v<U, value_type>,
                                       bool> = false>
   KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION simd(U&& value)
-      : m_value(_mm256_set1_epi64x(bit_cast<std::int64_t>(value_type(value)))) {
-  }
+      : m_value(_mm256_set1_epi64x(
+            Kokkos::bit_cast<std::int64_t>(value_type(value)))) {}
```
dalg24, line 818: Ping @ldh4 @ibaned this function has to be qualified to avoid ambiguities with C++20 if the `<bit>` header is transitively included.  Just FYI.

dalg24:
Not waiting for HIP builds that are headed straight to timeout.  Everything else passed.

PR #6123, opened 2023-05-10
core/src: Move floating point wrapper to private header

e10harvey:
With the addition of traits and math functions, Kokkos_Half.hpp was getting large and monolithic. With this change, Kokkos_Half.hpp is still the user facing header but now only includes private implementation headers.


dalg24 (approved):

masterleinad (approved):
Looks good diffing the files.

PR #6124, opened 2023-05-10
core/src: Add half math functions to private header

e10harvey:
## Summary of changes
- core/src: Add half math functions to private header

This PR introduces initial support for half math functions which rely on the floating_point_wrapper explicit conversion constructors to cast half_t and bhalf_t arguments to float in order to fall back on the float implementation of math functions. If this fall back approach is found to be a bottleneck by users, we could add a follow on PR to see if invoking backend-specific implementations (where available) of half_t and bhalf_t math functions helps.

Required for https://github.com/kokkos/kokkos-kernels/pull/1774.


crtrott:
Looks like this needs a rebase? After that I assume its ready for review? We probably should decide which order the two half moving things should come in.

dalg24 (changes requested):
Please provide a proper description.

Make sure you discuss why we cast to float and do not bother using native half precision intrinsics when available.

dalg24:
> @dalg24, @crtrott, @lucbv: Ready to merge?

Not reviewed so no not ready to merge.  You can try to poll for reviews on nucleus.

crtrott:
Do we have no tests for this?

crtrott (changes requested):
I think we need some nesting. Can we reuse the existing one?

e10harvey:
It is tested already via the float math functions. Shall I add tests for these wrappers?

e10harvey:
> Please provide a proper description.
> 
> Make sure you discuss why we cast to float and do not bother using native half precision intrinsics when available.

@dalg24: I made this changes in the description above; does it address your feedback?

e10harvey:
> I think we need some testing. Can we reuse the existing one?

@dalg24: Is it best to add `half_t` and `bhalf_t` to the type lists for the existing math function tests?

e10harvey:
> > I think we need some testing. Can we reuse the existing one?
> 
> @dalg24: Is it best to add `half_t` and `bhalf_t` to the type lists for the existing math function tests?

I do not see a way to reuse the existing one as expected values are computed from `std::FUNC`.

dalg24:
A bare minimum test would be calling a few selected math functions on the device and make sure it compiles and executes.
A more rigorous one would probably compare the result to higher precision reference converted to half types.

e10harvey:
> The CI is failing with warnings like

@masterleinad: I see that `-Wno-gnu-zero-variadic-macro-arguments` is used in the `SYCL-OneAPI` Jenkins pipeline stage which appears to be parsing Kokkos with clang. I cannot reproduce this with clang 15. Which toolchains did you see throwing that warning as an error?

masterleinad:
> Which toolchains did you see throwing that warning as an error?

Like half of the GitHub CI. 🙂

e10harvey:
I did push two commits to address the CI failures but they're not showing up here for some reason.

Once this merges, we can proceed with this code cleanup: https://github.com/kokkos/kokkos-kernels/pull/1774

CC: @lucbv

core/src/impl/Kokkos_Half_MathematicalFunctions.hpp
```diff
@@ -0,0 +1,163 @@
...
+      MACRO(FUNC, Kokkos::Experimental::half_t)
+  #endif  // defined(KOKKOS_BHALF_T_IS_FLOAT) && !KOKKOS_BHALF_T_IS_FLOAT
+#else
+  #if defined(KOKKOS_BHALF_T_IS_FLOAT) && !KOKKOS_BHALF_T_IS_FLOAT
+    #define KOKKOS_IMPL_MATH_HALF_FUNC_WRAPPER(MACRO, FUNC) \
+      MACRO(FUNC, Kokkos::Experimental::bhalf_t)
+  #else
+    #define KOKKOS_IMPL_MATH_HALF_FUNC_WRAPPER(MACRO, FUNC)
+  #endif  // defined(KOKKOS_BHALF_T_IS_FLOAT) && !KOKKOS_BHALF_T_IS_FLOAT
+#endif  // defined(KOKKOS_HALF_T_IS_FLOAT) && !KOKKOS_HALF_T_IS_FLOAT
```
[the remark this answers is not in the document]
masterleinad: ping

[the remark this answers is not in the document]
e10harvey: This is cleaner. I will commit it once CI starts passing.

masterleinad:
MSVC is still failing with
```
C:\projects\source\core\unit_test\TestMathematicalFunctions.hpp(688,3): error C2275: 'Kokkos::Experimental::half_t': illegal use of this type as an expression [C:\projects\source\build\core\unit_test\Kokkos_CoreUnitTest_Serial1.vcxproj]
C:\projects\source\core\unit_test\TestMathematicalFunctions.hpp(688): message : see declaration of 'Kokkos::Experimental::half_t' [C:\projects\source\build\core\unit_test\Kokkos_CoreUnitTest_Serial1.vcxproj]
C:\projects\source\core\unit_test\TestMathematicalFunctions.hpp(688,3): error C3861: 'TEST_HALF_MATH_FUNCTION': identifier not found [C:\projects\source\build\core\unit_test\Kokkos_CoreUnitTest_Serial1.vcxproj]
C:\projects\source\core\unit_test\TestMathematicalFunctions.hpp(689,3): error C2275: 'Kokkos::Experimental::bhalf_t': illegal use of this type as an expression [C:\projects\source\build\core\unit_test\Kokkos_CoreUnitTest_Serial1.vcxproj]
C:\projects\source\core\unit_test\TestMathematicalFunctions.hpp(689): message : see declaration of 'Kokkos::Experimental::bhalf_t' [C:\projects\source\build\core\unit_test\Kokkos_CoreUnitTest_Serial1.vcxproj]
C:\projects\source\core\unit_test\TestMathematicalFunctions.hpp(689,3): error C3861: 'TEST_HALF_MATH_FUNCTION': identifier not found [C:\projects\source\build\core\unit_test\Kokkos_CoreUnitTest_Serial1.vcxproj]
C:\projects\source\core\unit_test\TestMathematicalFunctions.hpp(703,3): error C2275: 'Kokkos::Experimental::half_t': illegal use of this type as an expression [C:\projects\source\build\core\unit_test\Kokkos_CoreUnitTest_Serial1.vcxproj]
[...]
```
and `SYCL` CI is reporting
```
/var/jenkins/workspace/Kokkos/core/src/impl/Kokkos_Half_MathematicalFunctions.hpp:123:36: error: all paths through this function will call itself [-Werror,-Winfinite-recursion]
KOKKOS_IMPL_MATH_HALF_FUNC_WRAPPER(KOKKOS_IMPL_MATH_UNARY_FUNCTION_HALF_TYPE, nearbyint)
```

core/src/impl/Kokkos_Half_MathematicalFunctions.hpp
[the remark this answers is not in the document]
masterleinad: ping

core/src/impl/Kokkos_Half_MathematicalFunctions.hpp
```diff
@@ -0,0 +1,168 @@
...
+KOKKOS_IMPL_MATH_HALF_FUNC_WRAPPER(KOKKOS_IMPL_MATH_BINARY_FUNCTION_HALF, nextafter)
+// nexttoward
+KOKKOS_IMPL_MATH_HALF_FUNC_WRAPPER(KOKKOS_IMPL_MATH_BINARY_FUNCTION_HALF, copysign)
+// Classification and comparison functions
+// fpclassify
+KOKKOS_IMPL_MATH_HALF_FUNC_WRAPPER(KOKKOS_IMPL_MATH_UNARY_PREDICATE_HALF, isfinite)
+KOKKOS_IMPL_MATH_HALF_FUNC_WRAPPER(KOKKOS_IMPL_MATH_UNARY_PREDICATE_HALF, isinf)
+#if !defined(KOKKOS_ENABLE_SYCL) && !defined(KOKKOS_ENABLE_HIP) // FIXME_SYCL, FIXME_HIP
+KOKKOS_IMPL_MATH_HALF_FUNC_WRAPPER(KOKKOS_IMPL_MATH_UNARY_PREDICATE_HALF, isnan)
+#endif
```
masterleinad, line 147: So these are calling `isfinite(float(x))`, `isinf(float(x))`, and `isnan(float(x))`? Does that make sense?

e10harvey: I would expect nan or inf or a finite value to be converted to the correct representation in float. I believe that is the behavior when casting double to float.

masterleinad (approved):

e10harvey:
Retest this please

core/src/impl/Kokkos_Half_MathematicalFunctions.hpp
```diff
@@ -0,0 +1,168 @@
...
+KOKKOS_IMPL_MATH_HALF_FUNC_WRAPPER(KOKKOS_IMPL_MATH_UNARY_PREDICATE_HALF, signbit)
+// isgreater
+// isgreaterequal
+// isless
+// islessequal
+// islessgreater
+// isunordered
+// Complex numer functions
+KOKKOS_IMPL_MATH_HALF_FUNC_WRAPPER(KOKKOS_IMPL_MATH_COMPLEX_REAL_HALF, real)
+KOKKOS_IMPL_MATH_HALF_FUNC_WRAPPER(KOKKOS_IMPL_MATH_COMPLEX_IMAGINARY_HALF, imag)
```
dalg24, line 158: There is a typo.  (numer -> number)

Also how much do we care about complex with half type.  Presumably most complex functions are not currently enabled because `Impl::promote` is not specialized for half types.

core/unit_test/TestMathematicalFunctions.hpp
```diff
@@ -1131,6 +1337,230 @@ TEST(TEST_CATEGORY, mathematical_functions_absolute_value) {
...
+TEST(TEST_CATEGORY, mathematical_functions_ieee_remainder_function) {
+  TestIEEEFloatingPointRemainderFunction<TEST_EXECSPACE>();
+}
+#endif
+
+// TestFpClassify
+
+// TestIsFinite
+
+// TestIsInf
```
dalg24, line 1562: What are these?  A TODO?

e10harvey:
> Did you check it actually does what you expect?

Yes.

> Also how much do we care about complex with half type. Presumably most complex functions are not currently enabled because Impl::promote is not specialized for half types.

We need this for Arith Traits in Kokkos kernels.

> What are these? A TODO?

Yes, the same way we noted them in quad precision. These test are missing for all types, not just half_t and bhalf_t.

> There is a typo. (numer -> number)

can this be corrected without re-triggering the CI checks?

core/src/impl/Kokkos_Half_MathematicalFunctions.hpp
dalg24: > We need this for Arith Traits in Kokkos kernels.

Are you saying all you need is `real` and `imag`?

> can this be corrected without re-triggering the CI checks?

Per https://github.com/kokkos/internal-documents/blob/master/testing/ORNL_JenkinsCI.md#skipping-a-build include `[ci skip]` in your commit message.

core/unit_test/TestMathematicalFunctions.hpp
e10harvey: Not unless @srajama1 or @lucbv requests it.

core/src/impl/Kokkos_Half_MathematicalFunctions.hpp
```diff
@@ -0,0 +1,168 @@
...
+  KOKKOS_INLINE_FUNCTION bool FUNC(HALF_TYPE x) {              \
+    return Kokkos::FUNC(static_cast<float>(x));                \
+  }
+
+
+#define KOKKOS_IMPL_MATH_COMPLEX_REAL_HALF(FUNC, HALF_TYPE) \
+  KOKKOS_INLINE_FUNCTION HALF_TYPE FUNC(HALF_TYPE x) { return x; }
+
+
+#define KOKKOS_IMPL_MATH_COMPLEX_IMAGINARY_HALF(FUNC, HALF_TYPE) \
```
dalg24, line 67: Common practice is to say "imag" which has the same number of characters as "real" and would yield better alignment in the following.

core/src/impl/Kokkos_Half_MathematicalFunctions.hpp
```diff
@@ -0,0 +1,168 @@
...
+  }
+
+
+#define KOKKOS_IMPL_MATH_UNARY_PREDICATE_HALF(FUNC, HALF_TYPE) \
+  KOKKOS_INLINE_FUNCTION bool FUNC(HALF_TYPE x) {              \
+    return Kokkos::FUNC(static_cast<float>(x));                \
+  }
+
+
+#define KOKKOS_IMPL_MATH_COMPLEX_REAL_HALF(FUNC, HALF_TYPE) \
```
dalg24, line 63: Defining these complex related macros after all the other maths functions closer to their use would improve readability IMO

core/src/impl/Kokkos_Half_MathematicalFunctions.hpp
```diff
@@ -0,0 +1,168 @@
...
+// Complex number functions
+KOKKOS_IMPL_MATH_HALF_FUNC_WRAPPER(KOKKOS_IMPL_MATH_COMPLEX_REAL_HALF, real)
+KOKKOS_IMPL_MATH_HALF_FUNC_WRAPPER(KOKKOS_IMPL_MATH_COMPLEX_IMAGINARY_HALF, imag)
+
+
+#undef KOKKOS_IMPL_MATH_UNARY_PREDICATE_HALF
+#undef KOKKOS_IMPL_MATH_BINARY_FUNCTION_HALF
+#undef KOKKOS_IMPL_MATH_UNARY_FUNCTION_HALF_TYPE
+#undef KOKKOS_IMPL_MATH_COMPLEX_REAL_HALF
+#undef KOKKOS_IMPL_MATH_COMPLEX_IMAGINARY_HALF
```
dalg24, line 165: You forgot to undef `KOKKOS_IMPL_MATH_HALF_FUNC_WRAPPER`, `KOKKOS_IMPL_MATH_H_FUNC_WRAPPER` and `KOKKOS_IMPL_MATH_B_FUNC_WRAPPER`

core/unit_test/TestMathematicalFunctions.hpp
```diff
@@ -261,29 +285,34 @@ struct FloatingPointComparison {
...
+      static_assert(                                                    \
+          std::is_same<decltype(Kokkos::FUNC((T)0)),                    \
+                       math_unary_function_return_type_t<T>>::value);   \
+      return Kokkos::FUNC(x);                                           \
+    }                                                                   \
+    template <typename T>                                               \
+    static auto eval_std(T x) {                                         \
+      if constexpr (std::is_same<T, KE::half_t>::value ||               \
+                    std::is_same<T, KE::bhalf_t>::value) {              \
+        return std::FUNC(static_cast<float>(x));                        \
```
dalg24, line 301: Is it intentional to stay in floating point.
Are we comparing results as float or half types?

core/unit_test/TestMathematicalFunctions.hpp
```diff
@@ -380,29 +409,37 @@ DEFINE_UNARY_FUNCTION_EVAL(logb, 2);
...
+      static_assert(                                                        \
+          std::is_same<decltype(Kokkos::FUNC((T)0, (U)0)),                  \
+                       math_binary_function_return_type_t<T, U>>::value);   \
+      return Kokkos::FUNC(x, y);                                            \
+    }                                                                       \
+    template <typename T, typename U>                                       \
+    static auto eval_std(T x, U y) {                                        \
+      if constexpr (std::is_same<T, KE::half_t>::value ||                   \
+                    std::is_same<T, KE::bhalf_t>::value) {                  \
+        return std::FUNC(static_cast<float>(x), static_cast<int>(y));       \
```
dalg24, line 425: Why are we converting the2nd argument to int?  What does that even mean with `hypot` or `pow` for instance.
This looks wrong to me.

core/unit_test/TestMathematicalFunctions.hpp
```diff
@@ -511,6 +554,53 @@ void do_test_math_unary_function(const Arg (&x)[N]) {
...
+    static_cast<T>(d), static_cast<T>(e)})
+#define TEST_HALF_MATH_FUNCTION6(FUNC, T, a, b, c, d, e, f)  \
+  TEST_MATH_FUNCTION(FUNC)                                   \
+  ({static_cast<T>(a), static_cast<T>(b), static_cast<T>(c), \
+    static_cast<T>(d), static_cast<T>(e), static_cast<T>(f)})
+#define TEST_HALF_MATH_FUNCTION7(FUNC, T, a, b, c, d, e, f, g) \
+  TEST_MATH_FUNCTION(FUNC)                                     \
+  ({static_cast<T>(a), static_cast<T>(b), static_cast<T>(c),   \
+    static_cast<T>(d), static_cast<T>(e), static_cast<T>(f),   \
+    static_cast<T>(g)})
```
dalg24, line 601: Why are we using variadic macros instead of variadic function templates?

core/unit_test/TestMathematicalFunctions.hpp
```diff
@@ -1131,6 +1337,230 @@ TEST(TEST_CATEGORY, mathematical_functions_absolute_value) {
...
+    using Kokkos::isnan;
+    if (fabs(-0.) != 0. || !isinf(fabs(-INFINITY)) || !isnan(fabs(-NAN))) {
+      ++e;
+      KOKKOS_IMPL_DO_NOT_USE_PRINTF(
+          "failed fabs(floating_point) special values\n");
+    }
+
+    static_assert(std::is_same<decltype(fabs(static_cast<KE::half_t>(4.f))),
+                               KE::half_t>::value,
+                  "");
```
dalg24, line 1385: There is no point on having an empty error message.  Just drop the argument.

core/unit_test/TestMathematicalFunctions.hpp
```diff
@@ -1131,6 +1337,230 @@ TEST(TEST_CATEGORY, mathematical_functions_absolute_value) {
...
+    static_assert(std::is_same<decltype(fabs(5.)), double>::value, "");
+#ifdef MATHEMATICAL_FUNCTIONS_HAVE_LONG_DOUBLE_OVERLOADS
+    static_assert(std::is_same<decltype(fabs(6.l)), long double>::value, "");
+#endif
+  }
+};
+
+TEST(TEST_CATEGORY, mathematical_functions_floating_point_absolute_value) {
+  TestFloatingPointAbsoluteValueFunction<TEST_EXECSPACE>();
+}
```
dalg24, line 1399: OK even though I would have preferred to add in a separate PR.  (Missing tests for `fabs`)

core/unit_test/TestMathematicalFunctions.hpp
```diff
@@ -1131,6 +1337,230 @@ TEST(TEST_CATEGORY, mathematical_functions_absolute_value) {
...
+#ifdef MATHEMATICAL_FUNCTIONS_HAVE_LONG_DOUBLE_OVERLOADS
+    static_assert(std::is_same<decltype(fmod(6.l, 6.l)), long double>::value,
+                  "");
+#endif
+  }
+};
+
+TEST(TEST_CATEGORY, mathematical_functions_remainder_function) {
+  TestFloatingPointRemainderFunction<TEST_EXECSPACE>();
+}
```
dalg24, line 1474: Same comment probably should be its own PR

core/unit_test/TestMathematicalFunctions.hpp
```diff
@@ -1131,6 +1337,230 @@ TEST(TEST_CATEGORY, mathematical_functions_absolute_value) {
...
+
+#if 0
```
dalg24, line 1476: ???

core/src/impl/Kokkos_Half_MathematicalFunctions.hpp
e10harvey: Good catch

core/unit_test/TestMathematicalFunctions.hpp
e10harvey: Yes, this is intentional. std doesn't define any functions for for T  = half_t or T = bhalf_t. We are comparing results as float. Even when the math functions do return half_t or bhalf_t results in the future, results will be compared as float.

core/unit_test/TestMathematicalFunctions.hpp
e10harvey: We cannot write a implicit conversion constructor for converting half to float since we cannot assign precedence to operator overloads in floating_point_wrapper. That being said, this looks wrong to me too. When either T or U is half_t or bhalf_t, we should upcast both arguments to float.

core/unit_test/TestMathematicalFunctions.hpp
e10harvey: To match the style of the existing tests  and re-use the code which use `TEST_MATH_FUNCTION` for testing built-in types.

core/unit_test/TestMathematicalFunctions.hpp
e10harvey: I could not get this to work for either built-in types of half types. Would you like me to file a issue to expand our math function test coverage?

e10harvey:
Thanks for the detailed reviews, @dalg24 and @masterleinad !

core/src/impl/Kokkos_Half_FloatingPointWrapper.hpp
```diff
@@ -23,6 +23,22 @@
...
+struct kokkos_type_is_half_t {
+  enum : bool { value = false };
+};
+
+/// @brief templated struct for determining if bhalf_t is an alias to float.
+/// @tparam T The type to specialize on.
+template <class T>
+struct kokkos_type_is_bhalf_t {
+  enum : bool { value = false };
+};
```
masterleinad, line 39: Does this need to be user-facing?
```suggestion
/// @brief templated struct for determining if half_t is an alias to float.
/// @tparam T The type to specialize on.
template <class T>
struct kokkos_type_is_half_t : std::false_type {};

/// @brief templated struct for determining if bhalf_t is an alias to float.
/// @tparam T The type to specialize on.
template <class T>
struct kokkos_type_is_bhalf_t : std::false_type {};
```

core/src/impl/Kokkos_Half_FloatingPointWrapper.hpp
```diff
@@ -44,6 +60,10 @@ class floating_point_wrapper;
 // Declare half_t (binary16)
 using half_t = Kokkos::Experimental::Impl::floating_point_wrapper<
     Kokkos::Impl::half_impl_t ::type>;
+template <>
+struct kokkos_type_is_half_t<half_t> {
+  enum : bool { value = true };
+};
```
masterleinad, line 66:
```suggestion
template <>
struct kokkos_type_is_half_t<half_t> : std::true_type{};
```

core/src/impl/Kokkos_Half_FloatingPointWrapper.hpp
```diff
@@ -110,7 +130,10 @@ KOKKOS_INLINE_FUNCTION
 #ifdef KOKKOS_IMPL_BHALF_TYPE_DEFINED
 using bhalf_t = Kokkos::Experimental::Impl::floating_point_wrapper<
     Kokkos::Impl ::bhalf_impl_t ::type>;
-
+template <>
+struct kokkos_type_is_bhalf_t<bhalf_t> {
+  enum : bool { value = true };
+};
```
masterleinad, line 136:
```suggestion
template <>
struct kokkos_type_is_bhalf_t<bhalf_t> : std::true_type {};
```

core/unit_test/TestMathematicalFunctions.hpp
```diff
@@ -380,29 +409,36 @@ DEFINE_UNARY_FUNCTION_EVAL(logb, 2);
...
+                       math_binary_function_return_type_t<T, U>>::value);   \
+      return Kokkos::FUNC(x, y);                                            \
+    }                                                                       \
+    template <typename T, typename U>                                       \
+    static auto eval_std(T x, U y) {                                        \
+      if constexpr (KE::kokkos_type_is_half_t<T>::value ||                  \
+                    KE::kokkos_type_is_bhalf_t<T>::value ||                 \
+                    KE::kokkos_type_is_half_t<U>::value ||                  \
+                    KE::kokkos_type_is_bhalf_t<U>::value) {                 \
+        return std::FUNC(static_cast<float>(x), static_cast<float>(y));     \
```
masterleinad, line 427: Why are we casting all arguments instead of only those that are `half_t`?

core/src/impl/Kokkos_Half_FloatingPointWrapper.hpp
e10harvey: I think it would be helpful to make this user-facing. The current approach of checking the HALF_T_IS_FLOAT macro and then using std::is_same is arduous and messy.

e10harvey:
Retest this please

masterleinad:
Retest this please.

core/unit_test/TestMathematicalFunctions.hpp
e10harvey: @dalg24, @masterleinad: Can we clean this up a bit with another wrapper using a logical-or:
```
template <class T>
struct kokkos_type_is_half : kokkos_type_is_half_t<T>::value || kokkos_type_is_bhalf_t<T>::value;
```
?

e10harvey:
Retest this please

core/src/impl/Kokkos_Half_FloatingPointWrapper.hpp
```diff
@@ -23,6 +23,18 @@
...
+/// @brief templated struct for determining if half_t is an alias to float.
+/// @tparam T The type to specialize on.
+template <class T>
+struct kokkos_type_is_half_t : std::false_type {};
+
+/// @brief templated struct for determining if bhalf_t is an alias to float.
+/// @tparam T The type to specialize on.
+template <class T>
+struct kokkos_type_is_bhalf_t : std::false_type {};
+}  // namespace Kokkos::Experimental
```
dalg24, line 36: Why introduce these traits in this PR?

The code below
```C++
static_assert(Kokkos::Experimental::kokkos_type_is_half_t<Kokkos::Experimental::half_t>::value);
```
would error out for some backends and not for others.
If you really want them you need to make the case why we might want them.

core/unit_test/TestMathematicalFunctions.hpp
dalg24: I don't understand your answer.  I don't see how that matches the style and why the macro approach is better.

core/src/impl/Kokkos_Half_FloatingPointWrapper.hpp
e10harvey: I answered this above... See https://github.com/kokkos/kokkos/pull/6124#discussion_r1244191746

core/unit_test/TestMathematicalFunctions.hpp
e10harvey: The current style uses a macro called `TEST_MATH_FUNCTION`. We cannot use that same function for half types - we need the variadic macro. `TEST_HALF_MATH_FUNCTION` matches this approach.

e10harvey:
@lucbv, @masterleinad, @masterleinad, @crtrott: Ready to merge?

core/unit_test/TestMathematicalFunctions.hpp
masterleinad: Please open an issue. 

I would also prefer a variadic function instead of the variadic macro but otherwise, the pull request is fine with me (we still need to document it well though).

e10harvey: https://github.com/kokkos/kokkos/issues/6275

core/src/impl/Kokkos_Half_FloatingPointWrapper.hpp
dalg24: I saw that and I don't find it convincing.  We can argue about it at the developer meeting tomorrow.

e10harvey: So, T could be float in the math tests and the cleanest alternative I can think of to using the trait is to do this:
```
      constexpr bool const x_is_half = !std::is_same<T, float>::value && (std::is_same<T, KE::bhalf_t>::value || std::is_same<T, KE::half_t>::value);
```
I would have to test whether `std::is_same<T, float>::value` evalutes to true when T is KE::bhalf_t or T is KE::half_t and the half-type aliases to float. If it's not true, we would then need to use `#if defined(KOKKOS_HALF_T_IS_FLOAT) && !KOKKOS_HALF_T_IS_FLOAT`.

So, I can do the above but think everyone would find using the traits easier and less error-prone than having to worry about whether the half-type is an alias to float. I didn't think of adding them till working on this PR.

core/unit_test/TestMathematicalFunctions.hpp
e10harvey: @dalg24: I am getting:
```
kokkos/core/unit_test/TestMathematicalFunctions.hpp:564:51: error: static_cast from 'const float *' to 'float' is not allowed
```

Do you know how to fix it?

dalg24: My bad
```diff
diff --git a/core/unit_test/TestMathematicalFunctions.hpp b/core/unit_test/TestMathematicalFunctions.hpp
index a31232008..84feba7f7 100644
--- a/core/unit_test/TestMathematicalFunctions.hpp
+++ b/core/unit_test/TestMathematicalFunctions.hpp
@@ -560,8 +560,10 @@ void do_test_math_unary_function(const Arg (&x)[N]) {

 template <class Half, class Space, class... Func, class Arg, std::size_t N>
 void do_test_half_math_unary_function(const Arg (&x)[N]) {
+  Half y[N];
+  std::copy(x, x + N, y);  // cast to array of half type
   (void)std::initializer_list<int>{
-      (TestMathUnaryFunction<Space, Func, Arg, N>(static_cast<Half>(x)), 0)...};
+      (TestMathUnaryFunction<Space, Func, Arg, N>(y), 0)...};

   // test if potentially device specific math functions also work on host
   if constexpr (!std::is_same_v<Space, Kokkos::DefaultHostExecutionSpace>)

File: core/unit_test/TestMathematicalFunctions.hpp (hunk 4 above)
Comment (e10harvey) [reply core/unit_test/TestMathematicalFunctions.hpp:1536]: https://github.com/kokkos/kokkos/issues/6279

File: core/unit_test/TestMathematicalFunctions.hpp [hunk 20]
```diff
@@ -511,6 +558,24 @@ void do_test_math_unary_function(const Arg (&x)[N]) {
 #define TEST_MATH_FUNCTION(FUNC) \
   do_test_math_unary_function<TEST_EXECSPACE, MathUnaryFunction_##FUNC>
 
+template <class Half, class Space, class... Func, class Arg, std::size_t N>
+void do_test_half_math_unary_function(const Arg (&x)[N]) {
+  Half y[N];
+  std::copy(x, x + N, y);  // cast to array of half type
+  (void)std::initializer_list<int>{
+      (TestMathUnaryFunction<Space, Func, Arg, N>(y), 0)...};
```
dalg24, line 566:
```suggestion
      (TestMathUnaryFunction<Space, Func, Half, N>(y), 0)...};
```

core/unit_test/TestMathematicalFunctions.hpp
```diff
@@ -511,6 +558,24 @@ void do_test_math_unary_function(const Arg (&x)[N]) {
...
+
+  // test if potentially device specific math functions also work on host
+  if constexpr (!std::is_same_v<Space, Kokkos::DefaultHostExecutionSpace>)
+    (void)std::initializer_list<int>{
+        (TestMathUnaryFunction<Kokkos::DefaultHostExecutionSpace, Func, Arg, N>(
```
dalg24, line 571:
```suggestion
        (TestMathUnaryFunction<Kokkos::DefaultHostExecutionSpace, Func, Half, N>(
```

masterleinad:
You will need to fix the indentation.

masterleinad (approved):
Assuming this is passing testing.

core/unit_test/TestMathematicalFunctions.hpp
```diff
@@ -53,6 +61,14 @@ template <class T>
 using math_unary_function_return_type_t = typename math_unary_function_return_type<T>::type;
 template <class, class>
 struct math_binary_function_return_type;
+#if defined(KOKKOS_HALF_T_IS_FLOAT) && !KOKKOS_HALF_T_IS_FLOAT
+template <> struct math_binary_function_return_type<KE::half_t, KE::half_t> { using type = KE::half_t; };
+template <> struct math_binary_function_return_type<int, KE::half_t> { using type = KE::half_t; };
```
dalg24, line 66: This look fishy

core/unit_test/TestMathematicalFunctions.hpp
dalg24: I don't quite understand why comparing them as float is the right thing to do.

core/unit_test/TestMathematicalFunctions.hpp
dalg24: You would need to provide much more than that.
Where is it being used?

dalg24 (commented):
I realize there is virtually no coverage for this with the fundamental types either but where do we define the promotion rules when mixing half types with other fundamental types in binary functions?

core/unit_test/TestMathematicalFunctions.hpp
e10harvey: @dalg24: I don't know how else to explain this. std math functions are not defined for half types. What else would you compare them to?

core/unit_test/TestMathematicalFunctions.hpp
e10harvey: The math binary function test calls. Please add what's missing.

e10harvey:
> I realize there is virtually no coverage for this with the fundamental types either but where do we define the promotion rules when mixing half types with other fundamental types in binary functions?

I don't understand your question. We always upcast for mixed precision, though. Please feel free to make any further changes you see fit to this PR.

core/unit_test/TestMathematicalFunctions.hpp
dalg24: do we mix int and half_t with the first argument as int and not the second?

dalg24:
> > I realize there is virtually no coverage for this with the fundamental types either but where do we define the promotion rules when mixing half types with other fundamental types in binary functions?
> 
> @dalg24: I don't understand your question. We always upcast for mixed precision, though. Please feel free to make any further changes you see fit to this PR.

Is `Kokkos::hypot(int, half_t) -> half_t` the right return type?
Because we do `hypot(int, float) -> double` (that is what the standard mandates).
Also `Kokkos::hypot(double, half_t) -> ???`

e10harvey:
> I realize there is virtually no coverage for this with the fundamental types either but where do we define the promotion rules when mixing half types with other fundamental types in binary functions?

@dalg24: I think what you're looking for here are the mixed precision operator overloads defined in Kokkos_Half_FloatingPointWrapper.hpp. IIRC, we do not support mixed precision for operators that result in a binary type.

dalg24:
> > > > I realize there is virtually no coverage for this with the fundamental types either but where do we define the promotion rules when mixing half types with other fundamental types in binary functions?
> > > 
> > > 
> > > @dalg24: I don't understand your question. We always upcast for mixed precision, though. Please feel free to make any further changes you see fit to this PR.
> > 
> > 
> > Is `Kokkos::hypot(int, half_t) -> half_t` the right return type? Because we do `hypot(int, float) -> double` (that is what the standard mandates). Also `Kokkos::hypot(double, half_t) -> ???`
> 
> I see. I would expect integral types to be cast to the floating point type, always. I would also expect to always upcast to higher-precision floating point type. In the case of `Kokkos::hypot(int, half_t) -> half_t`, I think it does make the most sense to cast both arguments to float and return a float.

No.  We want to be aligned with standard C++ which states that integral types will be promoted to double
https://eel.is/c++draft/c.math#cmath.syn-3

core/src/impl/Kokkos_Half_MathematicalFunctions.hpp
```diff
@@ -0,0 +1,169 @@
...
+#endif
+
+#define KOKKOS_IMPL_MATH_HALF_FUNC_WRAPPER(MACRO, FUNC) \
+  KOKKOS_IMPL_MATH_H_FUNC_WRAPPER(MACRO, FUNC)          \
+  KOKKOS_IMPL_MATH_B_FUNC_WRAPPER(MACRO, FUNC)
+
+
+#define KOKKOS_IMPL_MATH_UNARY_FUNCTION_HALF_TYPE(FUNC, HALF_TYPE)      \
+  KOKKOS_INLINE_FUNCTION HALF_TYPE FUNC(HALF_TYPE x) {                  \
+    return static_cast<HALF_TYPE>(Kokkos::FUNC(static_cast<float>(x))); \
```
e10harvey, line 46: @dalg24: This is where you want https://eel.is/c++draft/c.math#cmath.syn-3 enforced?

e10harvey:
> I don't quite understand why comparing them as float is the right thing to do.

@dalg24: Can you offer a appropriate alternative? e.g. comparing as double.

core/src/impl/Kokkos_Half_MathematicalFunctions.hpp
dalg24: No, it is about binary functions.  If you mix an integral argument with a half type, the result should be double, not half type.

core/src/impl/Kokkos_Half_MathematicalFunctions.hpp
```diff
@@ -0,0 +1,191 @@
...
+  } \
+  KOKKOS_IMPL_MATH_BINARY_FUNCTION_HALF_MIXED(FUNC, HALF_TYPE, double) \
+  KOKKOS_IMPL_MATH_BINARY_FUNCTION_HALF_MIXED(FUNC, HALF_TYPE, short) \
+  KOKKOS_IMPL_MATH_BINARY_FUNCTION_HALF_MIXED(FUNC, HALF_TYPE, unsigned short) \
+  KOKKOS_IMPL_MATH_BINARY_FUNCTION_HALF_MIXED(FUNC, HALF_TYPE, int) \
+  KOKKOS_IMPL_MATH_BINARY_FUNCTION_HALF_MIXED(FUNC, HALF_TYPE, unsigned int) \
+  KOKKOS_IMPL_MATH_BINARY_FUNCTION_HALF_MIXED(FUNC, HALF_TYPE, long) \
+  KOKKOS_IMPL_MATH_BINARY_FUNCTION_HALF_MIXED(FUNC, HALF_TYPE, unsigned long) \
+  KOKKOS_IMPL_MATH_BINARY_FUNCTION_HALF_MIXED(FUNC, HALF_TYPE, long long) \
+  KOKKOS_IMPL_MATH_BINARY_FUNCTION_HALF_MIXED(FUNC, HALF_TYPE, unsigned long long)
```
dalg24, line 76: Should we be also adding
```C++
  long double FUNC(long double x, HALF_TYPE y) {  \
    return Kokkos::FUNC(x, static_cast<long double>(y)); \
  } \
  long double FUNC(HALF_TYPE x, long double y) {  \
    return Kokkos::FUNC(static_cast<long double>(x), y); \
  } \

File: core/src/impl/Kokkos_Half_MathematicalFunctions.hpp [hunk 25]
```diff
@@ -0,0 +1,191 @@
...
+  #define KOKKOS_IMPL_MATH_B_FUNC_WRAPPER(MACRO, FUNC)
+#endif
+
+#define KOKKOS_IMPL_MATH_HALF_FUNC_WRAPPER(MACRO, FUNC) \
+  KOKKOS_IMPL_MATH_H_FUNC_WRAPPER(MACRO, FUNC)          \
+  KOKKOS_IMPL_MATH_B_FUNC_WRAPPER(MACRO, FUNC)
+
+
+#define KOKKOS_IMPL_MATH_UNARY_FUNCTION_HALF_TYPE(FUNC, HALF_TYPE)      \
+  KOKKOS_INLINE_FUNCTION HALF_TYPE FUNC(HALF_TYPE x) {                  \
```
dalg24, line 45: Just a thought, should we be specializing the function templates rather than adding new overloads?
@nliber please advise

core/src/impl/Kokkos_Half_MathematicalFunctions.hpp
e10harvey: We could do this. It would require adding a cast_from_half specialization that accepts long double as well as a explicit conversion constructor to support static_cast from half to long double. Would you like me to add those changes to this PR?

e10harvey: We discussed this in the developer meeting today and came to a consensus. Since `long double` is not supported in all backend and archuitectures, we will not be adding support for this at this time.

core/src/impl/Kokkos_Half_MathematicalFunctions.hpp
nliber: I think overloads are the way to go.  I keep going back to the advice in [Thou Shalt Not Specialize std Function Templates!](https://wg21.link/P0551) (and his various talks on the subject).  Fewer surprises with overloads.

nliber (approved):

dalg24 (approved):

PR #6129, opened 2023-05-11
Cuda: Remove unused attach_texture_object

masterleinad:
`attach_texture_object` isn't used anymore since https://github.com/kokkos/kokkos/pull/5623. Also, see https://github.com/kokkos/kokkos/pull/5989/files#r1191202517.
Technically, some of the functions are `public` but I don't think it's necessary to deprecate them first.

### DIFF
--- a/core/src/Cuda/Kokkos_CudaSpace.cpp
+++ b/core/src/Cuda/Kokkos_CudaSpace.cpp
@@ -423,49 +423,6 @@ SharedAllocationRecord<void, void>
     SharedAllocationRecord<Kokkos::CudaHostPinnedSpace, void>::s_root_record;
 #endif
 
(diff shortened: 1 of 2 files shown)


Rombur (approved):
As far as I can tell no external code is using these functions.

dalg24 (approved):
`Impl::SharedAllocationRecord<CudaSpace, void>::attach_texture_object[_offset]` are public member functions but `SharedAllocationRecord` belongs to implementation details.
So these functions are not part of Kokkos API and we can change them at will.  That said, thanks Bruno for double checking for usage.

tcclevenger (approved):
Thanks @masterleinad

dalg24:
Ignoring HIP builds that timed out

PR #6130, opened 2023-05-12
sorting an empty view should exit early and not fail

fnrizzi:
## `sort`
The current impl of `sort` does not handle nor test the special case of an empty view. 
Passing an empty view to sort should not fail. 
Currently, the code fails because if we look here: 

https://github.com/kokkos/kokkos/blob/3f602b6f283f82ecee2648f5606214a633d5e737/algorithms/src/Kokkos_Sort.hpp#L607-L640

when `view` is empty, the reduction does not happen and so `result` contains the reduction_identity values. 
Since those values are *not* equal, the following if is *not* satisfied:
```cpp
if (result.min_val == result.max_val) return;
```

and the code fails because it hits this:
```cpp
if (std::is_floating_point<typename ViewType::non_const_value_type>::value) {
    KOKKOS_ASSERT(std::isfinite(static_cast<double>(result.max_val) -
                                static_cast<double>(result.min_val)));
  }
```

## `BinSort`

For BinSort there is a similar problem here:

https://github.com/kokkos/kokkos/blob/3f602b6f283f82ecee2648f5606214a633d5e737/algorithms/src/Kokkos_Sort.hpp#L327-L331

## Comments

Note that an empty view is not a precondition violation. 
Also, the std lib behaves similarly: if we try to call std::sort on an empty vector the code does not crash. 

When merging, need to be careful with #6081

### DIFF
--- a/algorithms/src/Kokkos_Sort.hpp
+++ b/algorithms/src/Kokkos_Sort.hpp
@@ -329,6 +329,10 @@ class BinSort {
   template <class ExecutionSpace, class ValuesViewType>
   void sort(const ExecutionSpace& exec, ValuesViewType const& values,
             int values_range_begin, int values_range_end) const {
+    if (values.extent(0) == 0) {
+      return;
+    }
+
     static_assert(
         Kokkos::SpaceAccessibility<ExecutionSpace,
                                    typename Space::memory_space>::accessible,
--- a/algorithms/src/Kokkos_Sort.hpp
+++ b/algorithms/src/Kokkos_Sort.hpp
@@ -606,6 +606,10 @@ std::enable_if_t<(Kokkos::is_execution_space<ExecutionSpace>::value) &&
                                     memory_space>::accessible)>
 sort(const ExecutionSpace& exec,
      const Kokkos::View<DataType, Properties...>& view) {
+  if (view.extent(0) == 0) {
+    return;
+  }
+
   using ViewType = Kokkos::View<DataType, Properties...>;
   using CompType = BinOp1D<ViewType>;
 
@@ -648,8 +652,11 @@ sort(const ExecutionSpace& exec,
 template <class DataType, class... Properties>
 void sort(const Experimental::SYCL& space,
           const Kokkos::View<DataType, Properties...>& view) {
-  using ViewType = Kokkos::View<DataType, Properties...>;
+  if (view.extent(0) == 0) {
+    return;
+  }
 
+  using ViewType = Kokkos::View<DataType, Properties...>;
   static_assert(SpaceAccessibility<Experimental::SYCL,
                                    typename ViewType::memory_space>::accessible,
                 "SYCL execution space is not able to access the memory space "
@@ -676,6 +683,9 @@ std::enable_if_t<(Kokkos::is_execution_space<ExecutionSpace>::value) &&
                      HostSpace, typename Kokkos::View<DataType, Properties...>::
                                     memory_space>::accessible)>
 sort(const ExecutionSpace&, const Kokkos::View<DataType, Properties...>& view) {
+  if (view.extent(0) == 0) {
+    return;
+  }
   auto first = Experimental::begin(view);
   auto last  = Experimental::end(view);
   std::sort(first, last);
@@ -685,6 +695,9 @@ sort(const ExecutionSpace&, const Kokkos::View<DataType, Properties...>& view) {
 template <class DataType, class... Properties>
 void sort(const Cuda& space,
           const Kokkos::View<DataType, Properties...>& view) {
+  if (view.extent(0) == 0) {
+    return;
+  }
   const auto exec = thrust::cuda::par.on(space.cuda_stream());
   auto first      = Experimental::begin(view);
   auto last       = Experimental::end(view);
@@ -694,6 +707,9 @@ void sort(const Cuda& space,
 
 template <class ViewType>
 void sort(ViewType const& view) {
+  if (view.extent(0) == 0) {
+    return;
+  }
   Kokkos::fence("Kokkos::sort: before");
   typename ViewType::execution_space exec;
   sort(exec, view);
@@ -704,6 +720,10 @@ template <class ExecutionSpace, class ViewType>
 std::enable_if_t<Kokkos::is_execution_space<ExecutionSpace>::value> sort(
     const ExecutionSpace& exec, ViewType view, size_t const begin,
     size_t const end) {
+  if (view.extent(0) == 0) {
+    return;
+  }
+
   using range_policy = Kokkos::RangePolicy<typename ViewType::execution_space>;
   using CompType     = BinOp1D<ViewType>;
 
@@ -725,6 +745,10 @@ std::enable_if_t<Kokkos::is_execution_space<ExecutionSpace>::value> sort(
 
 template <class ViewType>
 void sort(ViewType view, size_t const begin, size_t const end) {
+  if (view.extent(0) == 0) {
+    return;
+  }
+
   Kokkos::fence("Kokkos::sort: before");
   typename ViewType::execution_space exec;
   sort(exec, view, begin, end);
--- a/algorithms/src/Kokkos_Sort.hpp
+++ b/algorithms/src/Kokkos_Sort.hpp
@@ -711,10 +711,12 @@ void sort(const Cuda& space,
 
 template <class ViewType>
 void sort(ViewType const& view) {
+  Kokkos::fence("Kokkos::sort: before");
+
   if (view.extent(0) == 0) {
     return;
   }
-  Kokkos::fence("Kokkos::sort: before");
+
   typename ViewType::execution_space exec;
   sort(exec, view);
   exec.fence("Kokkos::sort: fence after sorting");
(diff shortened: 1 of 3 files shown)


masterleinad (commented):
What about `BinSort::sort`?

fnrizzi:
> What about `BinSort::sort`?

yes that too, i am working on it... have not pushed yet

fnrizzi:
retest this please

algorithms/unit_tests/TestSort.hpp
```diff
@@ -440,6 +440,30 @@ void test_sort_empty_view() {
...
+  Kokkos::BinSort<KeyViewType, BinOp_t> Sorter;
+
+  // does not matter if we use int or something else
+  Kokkos::View<int*, ExecutionSpace> v;
+
+  // test all exposed public sort methods
+  Sorter.sort(ExecutionSpace(), v, 3, 8);
+  ASSERT_TRUE(v.extent(0) == 0);
+
+  Sorter.sort(v, 3, 8);
```
dalg24, line 462: (Commenting here but this is for both occurrences)
In my opinion you are writing a precondition violation.  The range must be valid.  Try [0,0[

fnrizzi: I had to do this to make it work, becuse of PR6131

algorithms/src/Kokkos_Sort.hpp
```diff
@@ -694,6 +711,9 @@ void sort(const Cuda& space,
 
 template <class ViewType>
 void sort(ViewType const& view) {
+  if (view.extent(0) == 0) {
+    return;
+  }
```
dalg24, line 716: Nope I think you need to fence before you exit early,

algorithms/unit_tests/TestSort.hpp
dalg24: Yes exactly

algorithms/src/Kokkos_Sort.hpp
fnrizzi: Why?

dalg24: Similar to deep_copy.  Overload that does not take an execution space is synchronous.

fnrizzi: thanks for pointing this out, i think i fixed it. there was also another overload wihtout exespace agument to fix.

algorithms/src/Kokkos_Sort.hpp
```diff
@@ -704,6 +726,10 @@ template <class ExecutionSpace, class ViewType>
 std::enable_if_t<Kokkos::is_execution_space<ExecutionSpace>::value> sort(
     const ExecutionSpace& exec, ViewType view, size_t const begin,
     size_t const end) {
```
dalg24, line 728: (Not expecting you to resolve here)
Anyone knows why we provide that overload?
Why is it not sufficient to sort a subview?  Also why is it not deferring to the overload that does not take the begin and end range arguments?
@crtrott

algorithms/unit_tests/TestSort.hpp
```diff
@@ -425,6 +425,43 @@ void test_sort(unsigned int N) {
   test_sort_integer_overflow<ExecutionSpace, unsigned long long>();
   test_sort_integer_overflow<ExecutionSpace, int>();
 }
+
+template <class ExecutionSpace>
+void test_sort_empty_view() {
+  // does not matter if we use int or something else
+  Kokkos::View<int*, ExecutionSpace> v;
```
dalg24, line 432:
```suggestion
  Kokkos::View<int*, ExecutionSpace> v("v", 0);
```

algorithms/unit_tests/TestSort.hpp
```diff
@@ -425,6 +425,43 @@ void test_sort(unsigned int N) {
...
+
+  Kokkos::sort(v);
+  // nothing should have changed
+  ASSERT_TRUE(v.extent(0) == 0);
+
+  Kokkos::sort(ExecutionSpace(), v);
+  // nothing should have changed
+  ASSERT_TRUE(v.extent(0) == 0);
```
dalg24, line 440:
```suggestion
  // TODO check the synchronous behavior of the calls below
  ASSERT_NO_THROW(Kokkos::sort(ExecutionSpace(), v));
  ASSERT_NO_THROW(Kokkos::sort(v));

```

algorithms/src/Kokkos_Sort.hpp
fnrizzi: because this is calling the bin sort which has a method taking a "range" defined by indices.

fnrizzi: and it works for dynamic views too. 
my plan is to revisit all this in a separate PR to make it more clear why we hae some things and what happens.

dalg24: My point is that if sorting a subview has the same effect, what is the point of this "non-standard" overload?

algorithms/src/Kokkos_Sort.hpp
```diff
@@ -648,8 +656,11 @@ sort(const ExecutionSpace& exec,
 template <class DataType, class... Properties>
 void sort(const Experimental::SYCL& space,
           const Kokkos::View<DataType, Properties...>& view) {
-  using ViewType = Kokkos::View<DataType, Properties...>;
+  if (view.extent(0) == 0) {
+    return;
+  }
```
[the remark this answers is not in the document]
fnrizzi: you mean this is not necessary because it just happens that oneapi handles that case anyway? 
I think it is important to not rely on assumptions about what happens impl-wise. 
basically, what i am saying is that for constraints and preconditons and corner cases, each overload of a function should be see at a high-level as:
```cpp
// constraints 
void function(...){. //overload 1
 // preconditions 
 // corner cases
 // impl call 
}

// constraints 
void function(...){. //overload 2
 // preconditions 
 // corner cases
 // impl call 
}
```
and so just because overload2 happens to call overload1 or something else, the preconditions or corner cases should be explicitly done in each. Suppose that we did not do this but just rely on the fact that the corner case is handled in overload1. Then someone later changes the impl and forgets about the corner case. This messes everything up.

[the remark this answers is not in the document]
fnrizzi: @dalg24 makes sense?

algorithms/src/Kokkos_Sort.hpp
fnrizzi: because you need to have the begin/end info to access the correct permutation elements that are stored inside binsort.

algorithms/unit_tests/TestSort.hpp
fnrizzi: this slipped , thought i did it too. thanks !

algorithms/src/Kokkos_Sort.hpp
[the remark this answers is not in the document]
dalg24: I would be comfortable with deferring to Thrust/oneDPL/the standard C++ library for something that trivial.
It is less code to maintain on our side.  That said, your approach is fine.

algorithms/src/Kokkos_Sort.hpp
dalg24: This makes sense in the context of BinSort but not for the Kokkos::sort algorithm.

fnrizzi: i totally agree with you. In fact, this seems like this part of the sort api was designed with the assumption that underneath was going to call bin sort.

algorithms/unit_tests/TestSort.hpp
```diff
@@ -425,6 +425,39 @@ void test_sort(unsigned int N) {
...
+void test_binsort_empty_view() {
+  // the bounds and extents used below are totally arbitrary
+  // and, in theory, should have no impact
+
+  using KeyViewType = Kokkos::View<int*, ExecutionSpace>;
+  KeyViewType kv("kv", 20);
+
+  using BinOp_t = Kokkos::BinOp1D<KeyViewType>;
+  BinOp_t binOp(5, 0, 10);
+  Kokkos::BinSort<KeyViewType, BinOp_t> Sorter(ExecutionSpace{}, kv, binOp);
```
masterleinad, line 449: Maybe also test that `kv` can be empty?

fnrizzi: added another test

algorithms/src/Kokkos_Sort.hpp
crtrott: Yeah lets revisit, but not here.

fnrizzi: yes not here, i will work on a new PR (I already started)

dalg24 (approved):

dalg24:
What did you change?

fnrizzi:
> What did you change?

change what? I only rebased on develop

dalg24:
> > What did you change?
> 
> change what? I only rebased on develop

Why bother?  Anyway too late.  I plan on merging after the CI completes

fnrizzi:
> > > What did you change?
> > 
> > 
> > change what? I only rebased on develop
> 
> Why bother? Anyway too late. I plan on merging after the CI completes

habit of keeping things updated so that no surprises come up *after* merging (maybe in this case does not matter, but in general does)

dalg24:
So you know we test your PR merged into develop unless there is conflict in which we can't merge anyway.

fnrizzi:
> So you know we test your PR merged into develop unless there is conflict in which we can't merge anyway.

Now I know :) thanks!

PR #6131, opened 2023-05-12
`BinSort`, `BinOp1D`, `BinOp3D`: make default constructor as deleted

fnrizzi:


dalg24:
Looking at the two commits I see that you initially went for removing the explicitly defaulting the default constructor which is another valid approach.  Why did you settle with marking them as deleted?  I don't necessarily have an issue with your choice but I am curious to hear your reasoning because it is clear that you thought about it.

Both `BinSort` and the `BinOp{1,3}D` functors are in the `Kokkos::` namespace and have been around for a long while.  Please briefly discuss briefly why it is ok to remove without going through deprecation.
For the record, I quickly searched through Trilinos and Cabana and did not find use case that would break.

fnrizzi:
@dalg24 I updated the description

dalg24:
> I don't think any real usage could have used the default constructor since it was totally unusable.

What about default construct and assign later before using?  Consider a case where it is declared as data member of some struct with a defaulted default constructor.  Although I don't think that is particularly good design, we have seen this pattern in the wild with other Kokkos constructs (views for instance).

I am not saying this is blocker, only that is not totally unreasonable that some code is currently relying on the default constructor being available.  You don't have to change.  Let see if others want to weigh in.

fnrizzi:
CI failure unrelated

fnrizzi:
> > I don't think any real usage could have used the default constructor since it was totally unusable.
> 
> What about default construct and assign later before using? Consider a case where it is declared as data member of some struct with a defaulted default constructor. Although I don't think that is particularly good design, we have seen this pattern in the wild with other Kokkos constructs (views for instance).
> 
> I am not saying this is blocker, only that is not totally unreasonable that some code is currently relying on the default constructor being available. You don't have to change. Let see if others want to weigh in.

Ah good point, i did not think about that. do you see an advantage in deprecating this rather than removing?

ajpowelsnl:
@fnrizzi , @dalg24 -- would you like to continue the discussion of this issue in the Wed. meeting?

dalg24:
> @fnrizzi , @dalg24 -- would you like to continue the discussion of this issue in the Wed. meeting?

No, the path forward is pretty clear, it needs to be guarded with `#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4`

fnrizzi:
Yes will do this soon

dalg24:
I took the liberty to apply the changes requested at the group meeting.

fnrizzi:
thanks Damien !

fnrizzi:
CI failure unrelated

dalg24 (approved):

algorithms/src/Kokkos_Sort.hpp
```diff
@@ -535,7 +543,11 @@ struct BinOp3D {
   double mul_[3]   = {};
   double min_[3]   = {};
 
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
   BinOp3D() = default;
```
masterleinad, line 547:
```suggestion
  KOKKOS_DEPRECATED BinOp3D() = default;
```

algorithms/src/Kokkos_Sort.hpp
```diff
@@ -491,7 +495,11 @@ struct BinOp1D {
   double mul_   = {};
   double min_   = {};
 
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
   BinOp1D() = default;
```
masterleinad, line 499:
```suggestion
  KOKKOS_DEPRECATED BinOp1D() = default;
```

algorithms/src/Kokkos_Sort.hpp
```diff
@@ -226,7 +226,11 @@ class BinSort {
   bool sort_within_bins;
 
  public:
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
   BinSort() = default;
```
masterleinad, line 230:
```suggestion
  KOKKOS_DEPRECATED BinSort() = default;
```

masterleinad (approved):

PR #6132, opened 2023-05-15
Increase minimum required HPX version to 1.8.0

msimberg:
This should really already have been part of #5628.

### DIFF
--- a/cmake/Modules/FindTPLHPX.cmake
+++ b/cmake/Modules/FindTPLHPX.cmake
@@ -1,5 +1,5 @@
 
-FIND_PACKAGE(HPX REQUIRED 1.7.0)
+FIND_PACKAGE(HPX REQUIRED 1.8.0)
 #as of right now, HPX doesn't export correctly
 #so let's convert it to an interface target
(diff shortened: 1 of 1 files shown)


dalg24 (approved):
I was going to ask you to update the documentation and just realized there is no such section...  We only mentions compilers/toolchains in https://kokkos.github.io/kokkos-core-wiki/requirements.html, nothing about third party libraries

dalg24 (changes requested):
Probably need to update the CI actually 
https://github.com/kokkos/kokkos/blob/62d2b6c879b74b6ae7bd06eb3e5e80139c4708e6/.github/workflows/continuous-integration-workflow-hpx.yml#L32

dalg24:
> Probably need to update the CI actually
> 
> https://github.com/kokkos/kokkos/blob/62d2b6c879b74b6ae7bd06eb3e5e80139c4708e6/.github/workflows/continuous-integration-workflow-hpx.yml#L32

CI that actually passes.
HPX bizarrely reports
```
-- HPX version: 1.8.0
```

Oh well.
Is that a known behavior?

PR #6133, opened 2023-05-15
Silence HPX deprecation warning from HPX

msimberg:
From HPX version 1.9.0 onwards `hpx::apply` has been renamed to `hpx::post`. `hpx::apply` is still there but it emits a deprecation warning and will be removed in future releases. This uses `hpx::post` when HPX is version 1.9.0 or newer.


dalg24 (approved):

cz4rs (approved):

dalg24:
HPX build passed along with a number of other non-HPX builds.  Not bothering to wait for the rest.

PR #6135, opened 2023-05-15
[ci skip] Add nightly ci for spack

Rombur:
Extracting the part of  #4746 that is working. I still need to configure everything on the jenkins side.

### DIFF
--- a/.jenkins_nightly
+++ b/.jenkins_nightly
@@ -1,0 +1,51 @@
+pipeline {
+    agent none
+
+    options {
+        timeout(time: 6, unit: 'HOURS')
+    }
+
+    stages {
+        stage('Build') {
+            parallel {
+                stage('spack-serial') {
+                    agent {
+                        docker {
+                          image 'ubuntu:22.04'
+                          label 'docker'
+                        }
+                    }
+                    steps {
+                        sh '''
+                          DEBIAN_FRONTEND=noninteractive && \
+                          apt-get update && apt-get upgrade -y && apt-get install -y \
+                          build-essential \
+                          wget \
+                          git \
+                          bc \
+                          python3-dev \
+                          && \
+                          apt-get clean && rm -rf /var/lib/apt/lists/*
+
+                          KEYDUMP_URL=https://cloud.cees.ornl.gov/download && \
+                          KEYDUMP_FILE=keydump && \
+                          wget --quiet ${KEYDUMP_URL}/${KEYDUMP_FILE} && \
+                          wget --quiet ${KEYDUMP_URL}/${KEYDUMP_FILE}.sig && \
+                          gpg --import ${KEYDUMP_FILE} && \
+                          gpg --verify ${KEYDUMP_FILE}.sig ${KEYDUMP_FILE} && \
+                          rm ${KEYDUMP_FILE}*
+
+                          rm -rf spack && \
+                          git clone https://github.com/spack/spack.git && \
+                          . ./spack/share/spack/setup-env.sh && \
+                          spack install kokkos@develop+tests && \
+                          spack load cmake && \
+                          spack test run kokkos && \
+                          spack test results -l
+                          '''
+                    }      
+                }   
+            }
+        }
+    }
+}


.jenkins_nightly
```diff
@@ -0,0 +1,93 @@
...
+                          && \
+                          apt-get clean && rm -rf /var/lib/apt/lists/*
+
+                          KEYDUMP_URL=https://cloud.cees.ornl.gov/download && \
+                          KEYDUMP_FILE=keydump && \
+                          wget --quiet ${KEYDUMP_URL}/${KEYDUMP_FILE} && \
+                          wget --quiet ${KEYDUMP_URL}/${KEYDUMP_FILE}.sig && \
+                          gpg --import ${KEYDUMP_FILE} && \
+                          gpg --verify ${KEYDUMP_FILE}.sig ${KEYDUMP_FILE} && \
+                          rm ${KEYDUMP_FILE}*
```
dalg24, line 36: What are we using from the key dump?

.jenkins_nightly
```diff
@@ -0,0 +1,77 @@
...
+                          bc \
+                          python3-dev \
+                          && \
+                          apt-get clean && rm -rf /var/lib/apt/lists/*
+
+                          rm -rf spack && \
+                          git clone https://github.com/spack/spack.git && \
+                          . ./spack/share/spack/setup-env.sh && \
+                          spack install kokkos@develop+tests && \
+                          spack load cmake && \
```
dalg24, line 34: just curious, is this loading from your path the cmake from the system package manager or cmake is being installed by Spack?

.jenkins_nightly
```diff
@@ -0,0 +1,77 @@
...
+                    steps {
+                        sh '''
+                          DEBIAN_FRONTEND=noninteractive && \
+                          apt-get update && apt-get upgrade -y && apt-get install -y \
+                          build-essential \
+                          wget \
+                          git \
+                          bc \
+                          python3-dev \
+                          gfortran \
```
dalg24, line 56: where is gfortran used?  some of our examples?

Rombur: We need gfortran for openmpi... In spack, `nvcc_wrapper` depends on MPI because we set `OMPI_CXX`, etc. Once I understand better how spack sets variables, I'll remove the MPI dependency.

.jenkins_nightly
```diff
@@ -0,0 +1,77 @@
...
+                          && \
+                          apt-get clean && rm -rf /var/lib/apt/lists/*
+
+                          rm -rf spack && \
+                          git clone https://github.com/spack/spack.git && \
+                          . ./spack/share/spack/setup-env.sh && \
+                          spack install kokkos@develop+cuda+wrapper+tests cuda_arch=80 ^cuda@12.1.0 && \
```
dalg24, line 63: is the cuda_arch required or it can detected?

dalg24: What about the `^cuda@12.1.0`?  Do we need it?

Rombur: I think it could be detected. I am having trouble with cuda  in spack. If you pay attention, we have two cuda installed. We need the docker image to have cuda **and** we need spack to install its own CUDA. When I was only installing CUDA through spack, the concretizer couldn't detect the architecture but it might work now.

.jenkins_nightly
```diff
@@ -0,0 +1,77 @@
...
+                          spack load cmake  && \
+                          spack load kokkos-nvcc-wrapper && \
+                          spack load cuda && \
+                          spack load kokkos && \
+                          spack load cmake && \
+                          spack test run kokkos && \
+                          spack test results -l
```
masterleinad, line 70:
```suggestion
                          spack load cmake  && \
                          spack load kokkos-nvcc-wrapper && \
                          spack load cuda && \
                          spack load kokkos && \
                          spack test run kokkos && \
                          spack test results -l
```
Why do we load `kokkos` here but not in `spack-serial`?

Rombur: Because of `kokkos-nvcc-wrapper`

masterleinad (approved):

dalg24 (approved):

PR #6136, opened 2023-05-15
Don't restrict index type in builtin reducers

### LINKED ISSUE #6134: MaxLoc/MinLoc reductions stopped supporting non-integral for location
There are use cases such as finding the maximum value in a multi-dimensional array plus the index pair. Clearly we don't have a test for that, so we need to unbreak this (i.e. don't require location to be an integral) and test for it. This got broken here: https://github.com/kokkos/kokkos/pull/5698

masterleinad:
Fixes #6134. Removes index type restrictions for `MinLoc`, `MaxLoc`, and `MinMaxLoc` and adds tests that are reduced versions of existing tests.

### DIFF
--- a/core/unit_test/TestReducers.hpp
+++ b/core/unit_test/TestReducers.hpp
@@ -1212,6 +1212,9 @@ struct TestReducers {
 #if defined(KOKKOS_ENABLE_OPENMPTARGET)
 #if defined(KOKKOS_COMPILER_CLANG) && (KOKKOS_COMPILER_CLANG >= 1300)
     test_minmaxloc(10007);
+#else
+    if (!std::is_same_v<ExecSpace, Kokkos::Experimental::OpenMPTarget>)
+      test_minmaxloc(10007);
 #endif
 #else
     test_minmaxloc(10007);
@@ -1230,10 +1233,11 @@ struct TestReducers {
 #if !defined(KOKKOS_ENABLE_OPENACC)
     // FIXME_OPENACC - OpenACC (V3.3) does not support custom reductions.
     test_minloc(10003);
-#if !defined(KOKKOS_COMPILER_NVHPC)  // FIXME_NVHPC misaligned memory
+#if defined(KOKKOS_COMPILER_NVHPC)  // FIXME_NVHPC misaligned memory
+    if (!std::is_same_v<ExecSpace, Kokkos::Cuda>)
     // FIXME_OPENMPTARGET requires custom reductions.
 #if !defined(KOKKOS_ENABLE_OPENMPTARGET)
-    test_minloc_2d(100);
+      test_minloc_2d(100);
 #endif
 #endif
 #endif
@@ -1242,9 +1246,10 @@ struct TestReducers {
     // FIXME_OPENACC - OpenACC (V3.3) does not support custom reductions.
     test_maxloc(10007);
 #if !defined(KOKKOS_COMPILER_NVHPC)  // FIXME_NVHPC misaligned memory
+    if (!std::is_same_v<ExecSpace, Kokkos::Cuda>)
 // FIXME_OPENMPTARGET requires custom reductions.
 #if !defined(KOKKOS_ENABLE_OPENMPTARGET)
-    test_maxloc_2d(100);
+      test_maxloc_2d(100);
 #endif
 #endif
 #endif
@@ -1255,6 +1260,9 @@ struct TestReducers {
 #if defined(KOKKOS_ENABLE_OPENMPTARGET)
 #if defined(KOKKOS_COMPILER_CLANG) && (KOKKOS_COMPILER_CLANG >= 1300)
     test_minmaxloc(10007);
+#else
+      if (!std::is_same_v<ExecSpace, Kokkos::Experimental::OpenMPTarget>)
+        test_minmaxloc(10007);
 #endif
 #else
     test_minmaxloc(10007);
--- a/core/unit_test/TestReducers.hpp
+++ b/core/unit_test/TestReducers.hpp
@@ -1221,13 +1221,19 @@ struct TestReducers {
 #if !defined(KOKKOS_ENABLE_OPENACC)
     // FIXME_OPENACC - OpenACC (V3.3) does not support custom reductions.
     test_minloc(10003);
+// FIXME_NVHPC misaligned memory
+#if !defined(KOKKOS_COMPILER_NVHPC)
     test_minloc_2d(100);
+#endif
 #endif
     test_max(10007);
 #if !defined(KOKKOS_ENABLE_OPENACC)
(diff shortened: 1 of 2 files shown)


fnrizzi:
I personally would keep the static assert for the reducers for the std algorithms

dalg24 (changes requested):
Why relax the constraints for the reducers that are meant to be used in the STL-like algorithms?

masterleinad:
> Why relax the constraints for the reducers that are meant to be used in the STL-like algorithms?

Those are all of IteratorType::difference_type and I wanted to discuss that we are aware of the requirement.
I dropped them now.

dalg24 (approved):
Looks fine.

I was about to nitpick about small things like initializing the value and location, using legacy RNG, etc. but I realized that is just following the style in other unit tests in the file.

fnrizzi (approved):

core/unit_test/TestReducers.hpp
```diff
@@ -22,6 +22,17 @@
 
 //--------------------------------------------------------------------------
 
+namespace Test {
+struct MyPair : Kokkos::pair<int, int> {};
+}  // namespace Test
+
+template <>
+struct Kokkos::reduction_identity<Test::MyPair> {
```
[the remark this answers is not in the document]
fnrizzi: @dalg24 btw, Daniel did use that in here: https://github.com/kokkos/kokkos/blob/62d2b6c879b74b6ae7bd06eb3e5e80139c4708e6/algorithms/src/std_algorithms/impl/Kokkos_Reduce.hpp#L113 already so we can maybe extract this trait and make it an actual functionality usable eveywhere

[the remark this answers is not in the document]
dalg24: Actually our [documentation](https://kokkos.github.io/kokkos-core-wiki/API/core/builtinreducers/MinLoc.html#additional-information) says

> In order to use MinLoc with a custom type of either `Scalar` or `Index`, a template specialization of `Kokkos::reduction_identity<CustomType>` must be defined.

which is decent.  Might be worth specifying only the `min` static member function is actually needed.

core/unit_test/TestReducers.hpp
```diff
@@ -1012,11 +1221,19 @@ struct TestReducers {
 #if !defined(KOKKOS_ENABLE_OPENACC)
     // FIXME_OPENACC - OpenACC (V3.3) does not support custom reductions.
     test_minloc(10003);
+// FIXME_NVHPC misaligned memory
+#if !defined(KOKKOS_COMPILER_NVHPC)
```
dalg24, line 1225: Since it looks like the OpenMPTarget build is failing and more changes will be needed, I would like you to put the FIXME comment on the same line as the `#ifndef`.  Same comment below.

dalg24: The suggestion was meant for comments that fit on the line

core/unit_test/TestReducers.hpp
```diff
@@ -1009,22 +1228,39 @@ struct TestReducers {
...
+#endif
+#endif
+#endif
+#if !defined(KOKKOS_ENABLE_OPENACC)  // FIXME_OPENACC - OpenACC (V3.3) does not
+                                     // support custom reductions.
+#if defined( \
+    KOKKOS_ENABLE_OPENMPTARGET)  // FIXME_OPENMPTARGET - The minmaxloc test
+                                 // fails llvm <= 13 version, the minmaxloc_2d
+                                 // test requires custom reductions.
+#if defined(KOKKOS_COMPILER_CLANG) && (KOKKOS_COMPILER_CLANG > 1300)
```
dalg24, line 1258: Not readable.  Also presumably can be enabled on the host-side and should only guarded for execution space that do not support it.

core/unit_test/TestReducers.hpp
```diff
@@ -1009,22 +1228,39 @@ struct TestReducers {
     test_sum(10001);
     test_prod(sizeof(Scalar) > 4 ? 35 : 19);  // avoid int overflow (see above)
     test_min(10003);
-#if !defined(KOKKOS_ENABLE_OPENACC)
-    // FIXME_OPENACC - OpenACC (V3.3) does not support custom reductions.
+#if !defined(KOKKOS_ENABLE_OPENACC)  // FIXME_OPENACC - OpenACC (V3.3) does not
+                                     // support custom reductions.
     test_minloc(10003);
+#if !defined(KOKKOS_COMPILER_NVHPC)       // FIXME_NVHPC misaligned memory
```
dalg24, line 1234: This is a runtime error so should be a skip test.

core/unit_test/TestReducers.hpp
```diff
@@ -1012,19 +1233,44 @@ struct TestReducers {
...
+// FIXME_OPENACC - OpenACC (V3.3) does not support custom reductions.
 #if !defined(KOKKOS_ENABLE_OPENACC)
-    // FIXME_OPENACC - OpenACC (V3.3) does not support custom reductions.
+// FIXME_OPENMPTARGET - The minmaxloc test fails llvm < 13 version,
+// the minmaxloc_2d test requires custom reductions.
+#if defined(KOKKOS_ENABLE_OPENMPTARGET)
+#if defined(KOKKOS_COMPILER_CLANG) && (KOKKOS_COMPILER_CLANG >= 1300)
+    test_minmaxloc(10007);
+#else
+    if (!std::is_same_v<ExecSpace, Kokkos::Experimental::OpenMPTarget>)
```
dalg24, line 1268: should that be `if constexpr`?  @rgayatri23

dalg24: what is our minimum requirement?

rgayatri23: Yeah `if constexpr` would work there. We don't have a minimum requirement for clang version but I would say not to try anything with less than 14 for all enabled unit tests to pass

masterleinad: Should we then formalize using clang-14 as the minimum in another pull request?

rgayatri23: Yes sure. I can do that.

dalg24: the question was "does it need to be constexpr" meaning was it a compile or runtime error.

rgayatri23: I do not remember. But if we restrict OpenMPTarget backend to use minimum clang-14, then we can avoid that condition

crtrott (approved):

crtrott:
Both HIP tests had failed (system issues), I think it would be good to at least have one of them run and pass to make sure there isn't something funky in the test.

masterleinad:
Retest this please.

PR #6137, opened 2023-05-16
Fix bit_cast for SYCL again

masterleinad:
After getting access to relevant compiler versions again, it turns out that https://github.com/kokkos/kokkos/pull/6121 wasn't quite enough to workaround the ambiguity caused by unqualified calls to `bitcast` in the SYCL compiler since the other candidate considered is `sycl::detail::bit_cast` (https://github.com/intel/llvm/blob/931203f26b8e10d273d89fcd7f90f11885a1baf5/sycl/include/sycl/bit_cast.hpp#L48-L57) that is forwarding to `sycl::bit_cast`. Hence, we get an ambiguity between `sycl::bit_cast` and `sycl::detail::bit_cast` with the current code.
Until https://github.com/intel/llvm/pull/9398 shows up in the compiler releases we are using, we have to import `sycl::detail::bit_cast` instead or drop defining `Kokkos::bit_cast` altogether. Neither of these options is particularly appealing but I lean towards the former.

### DIFF
--- a/.jenkins
+++ b/.jenkins
@@ -107,7 +107,7 @@ pipeline {
                                 -DCMAKE_BUILD_TYPE=Release \
                                 -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
                                 -DCMAKE_CXX_COMPILER=/opt/intel/oneapi/compiler/2023.0.0/linux/bin-llvm/clang++ \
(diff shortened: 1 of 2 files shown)


masterleinad:
`CUDA-11.7-NVHPC` is failing in `cuda.workgraph_fib` which looks unrelated.

dalg24 (approved):

Rombur (approved):

PR #6150, opened 2023-05-19
Drop `Kokkos_ENABLE_PROFILING_LOAD_PRINT` configuration option

dalg24:
`Kokkos_ENABLE_PROFILING_LOAD_PRINT` is exclusively used to control whether we print a debug statement or not when initializing kokkos-tools.
It was introduced in #823 because the print statement was aggravating.  I doubt anyone has been missing it too much and actually enables it  If we really think it is useful, I would suggest we pursue instead a runtime option on the tools side rather than a configuration time one.

I am proposing removal without going through deprecation.

Note to our Spack experts @crtrott and @janciesko that you should probably go ahead and remove that option 
https://github.com/spack/spack/blob/e9e5beee1f066d1ac211740d1d0dbda6154b6e1e/var/spack/repos/builtin/packages/kokkos/package.py#L88
And while you are at it maybe remove `qthread`...


Rombur (approved):
We need to update the wiki

masterleinad (approved):

ajpowelsnl (approved):

ajpowelsnl:
> We need to update the wiki

Thanks for the nudge, @Rombur :  https://github.com/kokkos/kokkos-core-wiki/pull/384

crtrott (approved):

dalg24:
Failure is unrelated (machine issue on one of the CUDA build)

PR #4499, opened 2021-11-03
Avoid hangs in the Threads backend

masterleinad:
This fixes the hangs seen in #4241 (after making deallocate fencing) by not using `execute_serial` for resizing the scratch space but doing that directly.
I'm not quite sure why using `execute_serial` is required or what its purpose is. Maybe, someone that is more familiar with the `Threads` backed can comment on it and tell me if this change is OK or not.

PR #6151, opened 2023-05-19
OpenMP: Ensure kernels submitted by multiple threads to the same instance don't run concurrently

masterleinad:
Related to thread-safety questions in #6051 and #4385.
This pull request ensures that kernels submitted to the same execution space instance from multiple threads don't run concurrently. Also, calling `fence` ensures that any kernel running when `fence` completes before returning. This should be enough to ensure that `is_running` in #6051 can use
```
fence();
return false;
```
as fallback implementation.

There are still some cases where the test added failed but I think it's better to address that in follow-up pull requests to limit the size of this pull request (and unblock #6051).

Note that the behavior we aim for is that of the GPU backends:
Kernels submitted to the same execution space instance don't run concurrently and are executed in submission order (which needs external thread-synchronization to be well-defined if multiple threads submit to the same instance).

### DIFF
--- a/core/src/OpenMP/Kokkos_OpenMP_Instance.cpp
+++ b/core/src/OpenMP/Kokkos_OpenMP_Instance.cpp
@@ -34,6 +34,9 @@
 namespace Kokkos {
 namespace Impl {
 
+std::vector<OpenMPInternal *> OpenMPInternal::all_instances;
+std::mutex OpenMPInternal::all_instances_mutex;
+
 void OpenMPInternal::acquire_lock() {
   while (1 == desul::atomic_compare_exchange(&m_pool_mutex, 0, 1,
                                              desul::MemoryOrderAcquire(),
@@ -304,6 +307,17 @@ void OpenMPInternal::finalize() {
   }
 
   m_initialized = false;
+
+  // guard erasing from all_instances
+  {
+    std::scoped_lock lock(all_instances_mutex);
+
+    auto it = std::find(all_instances.begin(), all_instances.end(), this);
+    if (it == all_instances.end())
+      Kokkos::abort(
+          "Execution space instance to be removed couldn't be found!");
+    all_instances.erase(it);
+  }
 }
 
 void OpenMPInternal::print_configuration(std::ostream &s) const {
--- a/core/src/OpenMP/Kokkos_OpenMP_Instance.cpp
+++ b/core/src/OpenMP/Kokkos_OpenMP_Instance.cpp
@@ -37,19 +37,6 @@ namespace Impl {
 std::vector<OpenMPInternal *> OpenMPInternal::all_instances;
 std::mutex OpenMPInternal::all_instances_mutex;
 
-void OpenMPInternal::acquire_lock() {
-  while (1 == desul::atomic_compare_exchange(&m_pool_mutex, 0, 1,
-                                             desul::MemoryOrderAcquire(),
-                                             desul::MemoryScopeDevice())) {
-    // do nothing
-  }
-}
-
-void OpenMPInternal::release_lock() {
-  desul::atomic_store(&m_pool_mutex, 0, desul::MemoryOrderRelease(),
-                      desul::MemoryScopeDevice());
-}
-
 void OpenMPInternal::clear_thread_data() {
   const size_t member_bytes =
       sizeof(int64_t) *
@@ -126,7 +113,8 @@ void OpenMPInternal::resize_thread_data(size_t pool_reduce_bytes,
       if (nullptr != m_pool[rank]) {
         m_pool[rank]->disband_pool();
 
-        space.deallocate(m_pool[rank], old_alloc_bytes);
+        // impl_deallocate to not fence here
+        space.impl_deallocate("[unlabeled]", m_pool[rank], old_alloc_bytes);
       }
 
       void *ptr = nullptr;
--- a/core/src/OpenMP/Kokkos_OpenMP_Instance.cpp
+++ b/core/src/OpenMP/Kokkos_OpenMP_Instance.cpp
@@ -304,7 +304,8 @@ void OpenMPInternal::finalize() {
     if (it == all_instances.end())
       Kokkos::abort(
           "Execution space instance to be removed couldn't be found!");
-    all_instances.erase(it);
+    *it = all_instances.back();
+    all_instances.pop_back();
   }
 }
 
--- a/core/src/OpenMP/Kokkos_OpenMP_Parallel_For.hpp
+++ b/core/src/OpenMP/Kokkos_OpenMP_Parallel_For.hpp
@@ -338,20 +338,16 @@ class ParallelFor<FunctorType, Kokkos::TeamPolicy<Properties...>,
     const size_t team_shared_size  = m_shmem_size;
     const size_t thread_local_size = 0;  // Never shrinks
 
-    m_instance->acquire_lock();
+    // Serialize kernels on the same execution space instance
+    std::lock_guard<std::mutex> lock(m_instance->m_instance_mutex);
 
     m_instance->resize_thread_data(pool_reduce_size, team_reduce_size,
                                    team_shared_size, thread_local_size);
-
-    // Serialize kernels on the same execution space instance
-    std::lock_guard<std::mutex> lock(m_instance->m_instance_mutex);
 
     if (execute_in_serial(m_policy.space())) {
       ParallelFor::template exec_team<WorkTag>(
           m_functor, *(m_instance->get_thread_data()), 0,
           m_policy.league_size(), m_policy.league_size());
-
-      m_instance->release_lock();
 
       return;
     }
@@ -391,8 +387,6 @@ class ParallelFor<FunctorType, Kokkos::TeamPolicy<Properties...>,
 
       data.disband_team();
     }
-
-    m_instance->release_lock();
   }
 
   inline ParallelFor(const FunctorType& arg_functor, const Policy& arg_policy)
--- a/core/src/OpenMP/Kokkos_OpenMP_Parallel_For.hpp
+++ b/core/src/OpenMP/Kokkos_OpenMP_Parallel_For.hpp
@@ -108,6 +108,8 @@ class ParallelFor<FunctorType, Kokkos::RangePolicy<Traits...>, Kokkos::OpenMP> {
 
  public:
   inline void execute() const {
+    // Serialize kernels on the same execution space instance
+    std::lock_guard<std::mutex> lock(m_instance->m_instance_mutex);
     if (execute_in_serial(m_policy.space())) {
       exec_range(m_functor, m_policy.begin(), m_policy.end());
       return;
@@ -202,6 +204,9 @@ class ParallelFor<FunctorType, Kokkos::MDRangePolicy<Traits...>,
 
  public:
   inline void execute() const {
+    // Serialize kernels on the same execution space instance
+    std::lock_guard<std::mutex> lock(m_instance->m_instance_mutex);
+
 #ifndef KOKKOS_COMPILER_INTEL
     if (execute_in_serial(m_iter.m_rp.space())) {
       exec_range(0, m_iter.m_rp.m_num_tiles);
@@ -337,6 +342,9 @@ class ParallelFor<FunctorType, Kokkos::TeamPolicy<Properties...>,
 
     m_instance->resize_thread_data(pool_reduce_size, team_reduce_size,
                                    team_shared_size, thread_local_size);
+
+    // Serialize kernels on the same execution space instance
+    std::lock_guard<std::mutex> lock(m_instance->m_instance_mutex);
 
     if (execute_in_serial(m_policy.space())) {
       ParallelFor::template exec_team<WorkTag>(
--- a/core/src/OpenMP/Kokkos_OpenMP_Parallel_Reduce.hpp
+++ b/core/src/OpenMP/Kokkos_OpenMP_Parallel_Reduce.hpp
@@ -92,6 +92,9 @@ class ParallelReduce<CombinedFunctorReducerType, Kokkos::RangePolicy<Traits...>,
                                    0  // thread_local_bytes
     );
 
+    // Serialize kernels on the same execution space instance
+    std::lock_guard<std::mutex> lock(m_instance->m_instance_mutex);
+
     if (execute_in_serial(m_policy.space())) {
       const pointer_type ptr =
           m_result_ptr
@@ -106,6 +109,9 @@ class ParallelReduce<CombinedFunctorReducerType, Kokkos::RangePolicy<Traits...>,
           update);
 
       reducer.final(ptr);
+
+      m_instance->release_lock();
+
       return;
     }
     const int pool_size = m_instance->thread_pool_size();
@@ -226,6 +232,9 @@ class ParallelReduce<CombinedFunctorReducerType,
                                    ,
                                    0  // thread_local_bytes
     );
+
+    // Serialize kernels on the same execution space instance
+    std::lock_guard<std::mutex> lock(m_instance->m_instance_mutex);
 
 #ifndef KOKKOS_COMPILER_INTEL
     if (execute_in_serial(m_iter.m_rp.space())) {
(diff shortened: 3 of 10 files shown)


crtrott:
We need to look at overhead for this, also its not clear to me how fence() would help with the is_running thing here?

ajpowelsnl:
@masterleinad - would you like to continue discussing this issue in the Wed. meeting?

masterleinad:
> @masterleinad - would you like to continue discussing this issue in the Wed. meeting?

Yes, we should discuss this, depending on priority tomorrow or later.

masterleinad:
> We need to look at overhead for this, [...]

For an empty lambda using one thread, i.e., no contention, (which should be a worst case for applications using one thread) using
```C++
#include <Kokkos_Core.hpp>

  Kokkos::Timer timer;
  timer.reset();

  unsigned int N=10'000'000;

  for (unsigned int i=0; i<N; ++i)
    Kokkos::parallel_for(Kokkos::RangePolicy<Kokkos::Serial>(0, 1), KOKKOS_LAMBDA(int){});

  Kokkos::fence();
  auto count_time = timer.seconds();
  std::cout << count_time << std::endl;

  Kokkos::finalize();
  return 0;
}
```
I observe locally that execution time increases from `3.47s` to `3.62s`, i.e., it's like 4% slower.

masterleinad:
Needs #6441.

masterleinad:
Retest this please.

nmm0 (commented):
What exactly are the thread safety guarantees we are getting? E.g. are we also letting two threads submit to the same exec space index? If so we should test that

core/src/OpenMP/Kokkos_OpenMP_Parallel_For.hpp
```diff
@@ -202,6 +204,9 @@ class ParallelFor<FunctorType, Kokkos::MDRangePolicy<Traits...>,
 
  public:
   inline void execute() const {
+    // Serialize kernels on the same execution space instance
+    std::lock_guard<std::mutex> lock(m_instance->m_instance_mutex);
```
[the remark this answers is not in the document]
nmm0: I realized that you used the template argument for most calls to the lock constructor, so I think it's fine either way as long as its consistent

core/src/OpenMP/Kokkos_OpenMP_Parallel_Reduce.hpp
```diff
@@ -106,6 +109,9 @@ class ParallelReduce<CombinedFunctorReducerType, Kokkos::RangePolicy<Traits...>,
           update);
 
       reducer.final(ptr);
+
+      m_instance->release_lock();
```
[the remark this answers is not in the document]
masterleinad: I'm happy to rework `OpenMP`'s `acquire_lock`/`release_lock` in a different pull request.

core/src/Serial/Kokkos_Serial_Parallel_Range.hpp
```diff
@@ -191,10 +200,16 @@ class ParallelScan<FunctorType, Kokkos::RangePolicy<Traits...>,
...
         internal_instance->m_thread_team_data_mutex);
+
     internal_instance->resize_thread_team_data(
         pool_reduce_size, team_reduce_size, team_shared_size,
         thread_local_size);
 
+    // Make sure kernels are running sequentially even when using multiple
+    // threads
+    std::lock_guard<std::mutex> instance_lock(
+        internal_instance->m_instance_mutex);
```
dalg24, line 211: You probably should lock 1st thing or comment why you deferred until after the resize.
Same comment below.

core/src/Serial/Kokkos_Serial.cpp
```diff
@@ -59,6 +68,16 @@ void SerialInternal::finalize() {
...
   m_is_initialized = false;
+
+  // guard erasing from all_instances
+  {
+    std::scoped_lock lock(all_instances_mutex);
+    auto it = std::find(all_instances.begin(), all_instances.end(), this);
+    if (it == all_instances.end())
+      Kokkos::abort(
+          "Execution space instance to be removed couldn't be found!");
+    all_instances.erase(it);
```
dalg24, line 79: This is interesting it means we could now diagnose at finalize that some exec space instance has not been destroyed yet.
I am not suggesting you implement it now, just one thought.

dalg24 (commented):
(only reviewed the Serial backend changes so far)

core/src/Serial/Kokkos_Serial.hpp
```diff
@@ -61,6 +64,10 @@ class SerialInternal {
   static SerialInternal& singleton();
 
   std::mutex m_thread_team_data_mutex;
+  std::mutex m_instance_mutex;
```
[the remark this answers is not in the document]
dalg24: (Drop `m_thread_team_data_mutex`)

core/src/Serial/Kokkos_Serial.cpp
```diff
@@ -147,7 +166,9 @@ Serial::Serial(NewInstance)
     : m_space_instance(new Impl::SerialInternal, [](Impl::SerialInternal* ptr) {
         ptr->finalize();
         delete ptr;
-      }) {}
+      }) {
+  m_space_instance->initialize();
```
dalg24, line 170: Ugggg

dalg24 (commented):
I was about to comment that these changes do not guarantee that kernels, fences, and such get enqueued properly as the title line says, only that they don't don't run concurrently.
But your description is more accurate
> This pull request ensures that kernels submitted to the same execution space instance from multiple threads don't run concurrently. 

Please update the title and make sure the description says that it does not actually guarantee proper enqueuing.

masterleinad:
> Please update the title and make sure the description says that it does not actually guarantee proper enqueuing.

How would you define "enqueueing properly"? The definition I'm following is given by the behavior of the GPU backends, i.e., kernels submitted to the same execution space instance don't run concurrently and are executed in submission order (which needs external thread-synchronization to be well-defined if multiple threads submit to the same instance).

dalg24:
> > Please update the title and make sure the description says that it does not actually guarantee proper enqueuing.
> 
> How would you define "enqueueing properly"? The definition I'm following is given by the behavior of the GPU backends, i.e., kernels submitted to the same execution space instance don't run concurrently and are executed in submission order (which needs external thread-synchronization to be well-defined if multiple threads submit to the same instance).

I was thinking we are lacking the "execution in the same ordering as submission" but I suppose we have some playroom if we clearly document the caveat you added in your reply.

masterleinad:
I amended the description some for the expected behavior.

Rombur:
The serial tests are timing out

core/src/Serial/Kokkos_Serial.hpp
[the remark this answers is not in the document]
masterleinad: We need two locks since resize the scratch data might call fence which in turn locks the instance mutex so we are deadlocking if locking the instance mutex before resizing.

masterleinad:
> The serial tests are timing out

Reverting https://github.com/kokkos/kokkos/pull/6151/commits/ac408ddfe4c3ae327f014d2c6dad31b71b3519c0 fixed that.

core/src/Serial/Kokkos_Serial_Parallel_Range.hpp
masterleinad: We need both lock the `m_thread_team_data_mutex` and the `m_instance_mutex` since we want to make sure that there is no other kernel that modifies the scratch allocations while this kernel is executing.
We can't just use the `m_instance_mutex` alone (before `resize_thread_team_data`) since that function might call `deallocate` which implies a global fence and that fence tries locking `instance_lock` resulting in a deadlock.

masterleinad:
```
    frame #12: 0x000000010066c384 Kokkos_CoreUnitTest_Serial1`Kokkos::fence(name="HostSpace::impl_deallocate before free") at Kokkos_Core.cpp:1113:47 [opt]
    frame #13: 0x000000010066d60c Kokkos_CoreUnitTest_Serial1`Kokkos::HostSpace::impl_deallocate(this=<unavailable>, arg_label="Kokkos::Serial::scratch_mem", arg_alloc_ptr=0x000000013980b800, arg_alloc_size=11264, arg_logical_size=0, arg_handle=(name = "Host")) const at Kokkos_HostSpace.cpp:122:5 [opt]
    frame #14: 0x000000010066d560 Kokkos_CoreUnitTest_Serial1`Kokkos::HostSpace::deallocate(this=<unavailable>, arg_label=<unavailable>, arg_alloc_ptr=<unavailable>, arg_alloc_size=<unavailable>, arg_logical_size=<unavailable>) const at Kokkos_HostSpace.cpp:115:3 [opt]
    frame #15: 0x00000001006773bc Kokkos_CoreUnitTest_Serial1`Kokkos::Impl::SerialInternal::resize_thread_team_data(this=0x0000600001798000, pool_reduce_bytes=512, team_reduce_bytes=512, team_shared_bytes=8, thread_local_bytes=0) at Kokkos_Serial.cpp:119:13 [opt]
    frame #16: 0x00000001000477e0 Kokkos_CoreUnitTest_Serial1`Kokkos::Impl::ParallelFor<Test::TestSharedAtomicsFunctor<Kokkos::Serial>, Kokkos::TeamPolicy<Kokkos::Serial>, Kokkos::Serial>::execute(this=0x000000016fdfefe0) const at Kokkos_Serial_Parallel_Team.hpp:255:24 [opt]
    frame #17: 0x0000000100047634 Kokkos_CoreUnitTest_Serial1`void Kokkos::parallel_for<Kokkos::TeamPolicy<Kokkos::Serial>, Test::TestSharedAtomicsFunctor<Kokkos::Serial>, void>(str=<unavailable>, policy=0x000000016fdff0d0, functor=0x000000016fdff080) at Kokkos_Parallel.hpp:144:11 [opt]
    frame #18: 0x0000000100045e30 Kokkos_CoreUnitTest_Serial1`Test::serial_atomic_shared_Test::TestBody() [inlined] void Kokkos::parallel_for<Kokkos::TeamPolicy<Kokkos::Serial>, Test::TestSharedAtomicsFunctor<Kokkos::Serial>>(policy=0x000000016fdff0d0, functor=0x000000016fdff080, (null)=0x0000000000000000) at Kokkos_Parallel.hpp:153:3 [opt]
    frame #19: 0x0000000100045e18 Kokkos_CoreUnitTest_Serial1`Test::serial_atomic_shared_Test::TestBody(this=<unavailable>) at TestAtomicOperations_shared.hpp:49:3 [opt]
    frame #20: 0x000000010063cb14 Kokkos_CoreUnitTest_Serial1`void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) [inlined] void testing::internal::HandleSehExceptionsInMethodIfSupported<testing::Test, void>(object=<unavailable>, method=0x00000000000000010000000000000020, location="the test body") at gtest-all.cc:4101:10 [opt]
    frame #21: 0x000000010063cb04 Kokkos_CoreUnitTest_Serial1`void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(object=0x0000600002e88100, method=<unavailable>, location="the test body") at gtest-all.cc:4137:14 [opt]
    frame #22: 0x000000010063ca68 Kokkos_CoreUnitTest_Serial1`testing::Test::Run(this=0x0000600002e88100) at gtest-all.cc:4176:5 [opt]
    frame #23: 0x000000010063e1d4 Kokkos_CoreUnitTest_Serial1`testing::TestInfo::Run(this=0x0000000139605a00) at gtest-all.cc:4326:11 [opt]
    frame #24: 0x000000010063ed90 Kokkos_CoreUnitTest_Serial1`testing::TestSuite::Run(this=0x00000001396048f0) at gtest-all.cc:4480:28 [opt]
    frame #25: 0x000000010064eb74 Kokkos_CoreUnitTest_Serial1`testing::internal::UnitTestImpl::RunAllTests(this=0x000000013a804080) at gtest-all.cc:7320:44 [opt]
    frame #26: 0x000000010064e208 Kokkos_CoreUnitTest_Serial1`bool testing::internal::HandleExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>(testing::internal::UnitTestImpl*, bool (testing::internal::UnitTestImpl::*)(), char const*) [inlined] bool testing::internal::HandleSehExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>(object=<unavailable>, method=(Kokkos_CoreUnitTest_Serial1`testing::internal::UnitTestImpl::RunAllTests() at gtest-all.cc:7211), location="auxiliary test code (environments or event listeners)") at gtest-all.cc:4101:10 [opt]
    frame #27: 0x000000010064e1f8 Kokkos_CoreUnitTest_Serial1`bool testing::internal::HandleExceptionsInMethodIfSupported<testing::internal::UnitTestImpl, bool>(object=0x000000013a804080, method=<unavailable>, location="auxiliary test code (environments or event listeners)") at gtest-all.cc:4137:14 [opt]
    frame #28: 0x000000010064e178 Kokkos_CoreUnitTest_Serial1`testing::UnitTest::Run(this=0x0000000100854948) at gtest-all.cc:6903:10 [opt]
    frame #29: 0x000000010000369c Kokkos_CoreUnitTest_Serial1`main [inlined] RUN_ALL_TESTS() at gtest.h:12372:46 [opt]
    frame #30: 0x0000000100003694 Kokkos_CoreUnitTest_Serial1`main(argc=1, argv=0x000000016fdff638) at UnitTestMainInit.cpp:26:16 [opt]
    frame #31: 0x0000000189ebe0e0 dyld`start + 2360
```

masterleinad: @Rombur The above backtrace shows that we are calling `Kokkos::fence` in `deallocate` and that calls the static fence that is introduced in this pull request.

Rombur: I saw that but my understanding is that the global fence was added to fix a deadlock in the `Threads` backend and the code has changed quite a bit since then. I want to understand if the fence is still necessary or if we could do something in the `Threads` backend, so that we don't pollute the other backend.

masterleinad: We need the fence (introduced in https://github.com/kokkos/kokkos/pull/4499) for asynchronous host backends like `HPX` so we don't free an allocation that is still used (same reasoning as for the GPU backends). As for the scope of then fence, see https://github.com/kokkos/kokkos/pull/4499/commits/69ce416956a2350de03f8f81206e5f042f834bcc#r747018936.

masterleinad: I'm open to rediscussing that `deallocate` only fences asynchronous execution spaces (that can access the respective allocation).

dalg24 (commented):
Nitpick about the title: isn't it the same deal as for Serial, that we technically only ensure that kernels don't run concurrently, not so much in-order enqueuing?

core/src/OpenMP/Kokkos_OpenMP_Parallel_For.hpp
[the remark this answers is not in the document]
masterleinad: This is the same approach as for the `Serial` backend. We must ensure kernels don't run concurrently even if submitted by multiple threads and for `ParallelFor<[MD]RangePolicy>` there has not been any synchronization mechanism. In the other cases, we replace the `acquire_loock/release_lock` (initially designed to guard the use of scratch memory) with a `lock_guard` as well for consistency and since we need to control a larger scope.

dalg24 (approved):
Please update the PR title (see https://github.com/kokkos/kokkos/pull/6151#pullrequestreview-2164014790)

masterleinad:
> Please update the PR title (see [#6151 (review)](https://github.com/kokkos/kokkos/pull/6151#pullrequestreview-2164014790))

done

core/src/OpenMP/Kokkos_OpenMP_Instance.cpp
```diff
@@ -34,18 +34,8 @@
...
-                                             desul::MemoryOrderAcquire(),
-                                             desul::MemoryScopeDevice())) {
-    // do nothing
-  }
-}
-
-void OpenMPInternal::release_lock() {
-  desul::atomic_store(&m_pool_mutex, 0, desul::MemoryOrderRelease(),
-                      desul::MemoryScopeDevice());
-}
```
Rombur, line 48: After a quick look, it seems that `m_pool_mutex` is not used anymore and we can remove it.

Rombur (approved):

PR #6159, opened 2023-05-24
Check for overflow during backend initialization (Cuda, HIP, SYCL)

### LINKED ISSUE #3805: arithmetic overflow potential
In the file (./src/Cuda/Kokkos_Cuda_Instance.cpp) at line 567 of CudaInternal::scratch_flags function, there exists a integer overflow potential with size added to sizeScratchGrain

```
Cuda::size_type*CudaInternal::scratch_flags(const Cuda::size_type size) const {
  if (verify_is_initialized("scratch_flags") &&  
      m_scratchFlagsCount * sizeScratchGrain < size) {
          m_scratchFlagsCount = (size + sizeScratchGrain - 1) / sizeScratchGrain;

```

the check m_scratchFlagsCount * sizeScratchGrain less than size seems odd, given that m_scratchFlagsCount is set with new value.  but it would seem we could check for overflow by:

```
(size + sizeScratchGrain  - 1 ) > sizeScratchGrain // Not OF
```

cz4rs:
fixes #3805 

---
- refactor the code to avoid potential overflow when possible
- check for overflow when initializing scratch flags / scratch space
  - note: the code will `Kokkos::abort` when an overflow is detected (which should be fine in this case since we cannot allocate that much memory anyways)

### DIFF
--- a/core/src/Cuda/Kokkos_Cuda_Instance.cpp
+++ b/core/src/Cuda/Kokkos_Cuda_Instance.cpp
@@ -110,6 +110,13 @@ int cuda_kernel_arch() {
   return arch;
 }
 
+using ScratchGrain = Cuda::size_type[Impl::CudaTraits::WarpSize];
+static constexpr auto sizeScratchGrain = sizeof(ScratchGrain);
+
+std::size_t scratch_count(const std::size_t size) {
+  return (size + sizeScratchGrain - 1) / sizeScratchGrain;
+}
+
 }  // namespace
 
 Kokkos::View<uint32_t *, Kokkos::CudaSpace> cuda_global_unique_token_locks(
@@ -438,33 +445,25 @@ Kokkos::Cuda::initialize WARNING: Cuda is allocating into UVMSpace by default
 
 //----------------------------------------------------------------------------
 
-using ScratchGrain = Cuda::size_type[Impl::CudaTraits::WarpSize];
-static constexpr auto sizeScratchGrain = sizeof(ScratchGrain);
-
-constexpr std::size_t scratch_count(const std::size_t size) {
-  return (size + sizeScratchGrain - 1) / sizeScratchGrain;
-}
-
 Cuda::size_type *CudaInternal::scratch_flags(const std::size_t size) const {
   if (verify_is_initialized("scratch_flags") &&
       m_scratchFlagsCount < scratch_count(size)) {
-    m_scratchFlagsCount = scratch_count(size);
+    m_scratchFlagsCount   = scratch_count(size);
+    const auto alloc_size = check_mul_of(m_scratchFlagsCount, sizeScratchGrain);
 
     using Record =
         Kokkos::Impl::SharedAllocationRecord<Kokkos::CudaSpace, void>;
 
     if (m_scratchFlags) Record::decrement(Record::get_record(m_scratchFlags));
 
-    Record *const r =
-        Record::allocate(Kokkos::CudaSpace(), "Kokkos::InternalScratchFlags",
-                         (sizeof(ScratchGrain) * m_scratchFlagsCount));
+    Record *const r = Record::allocate(
+        Kokkos::CudaSpace(), "Kokkos::InternalScratchFlags", alloc_size);
 
     Record::increment(r);
 
     m_scratchFlags = reinterpret_cast<size_type *>(r->data());
 
-    KOKKOS_IMPL_CUDA_SAFE_CALL(
-        cudaMemset(m_scratchFlags, 0, m_scratchFlagsCount * sizeScratchGrain));
+    KOKKOS_IMPL_CUDA_SAFE_CALL(cudaMemset(m_scratchFlags, 0, alloc_size));
   }
 
   return m_scratchFlags;
@@ -473,16 +472,16 @@ Cuda::size_type *CudaInternal::scratch_flags(const std::size_t size) const {
 Cuda::size_type *CudaInternal::scratch_space(const std::size_t size) const {
   if (verify_is_initialized("scratch_space") &&
       m_scratchSpaceCount < scratch_count(size)) {
-    m_scratchSpaceCount = scratch_count(size);
+    m_scratchSpaceCount   = scratch_count(size);
+    const auto alloc_size = check_mul_of(m_scratchSpaceCount, sizeScratchGrain);
 
     using Record =
         Kokkos::Impl::SharedAllocationRecord<Kokkos::CudaSpace, void>;
 
     if (m_scratchSpace) Record::decrement(Record::get_record(m_scratchSpace));
 
-    Record *const r =
-        Record::allocate(Kokkos::CudaSpace(), "Kokkos::InternalScratchSpace",
-                         (sizeof(ScratchGrain) * m_scratchSpaceCount));
+    Record *const r = Record::allocate(
+        Kokkos::CudaSpace(), "Kokkos::InternalScratchSpace", alloc_size);
 
     Record::increment(r);
 
@@ -496,6 +495,8 @@ Cuda::size_type *CudaInternal::scratch_unified(const std::size_t size) const {
   if (verify_is_initialized("scratch_unified") && m_scratchUnifiedSupported &&
       m_scratchUnifiedCount < scratch_count(size)) {
     m_scratchUnifiedCount = scratch_count(size);
+    const auto alloc_size =
+        check_mul_of(m_scratchUnifiedCount, sizeScratchGrain);
 
     using Record =
         Kokkos::Impl::SharedAllocationRecord<Kokkos::CudaHostPinnedSpace, void>;
@@ -503,9 +504,9 @@ Cuda::size_type *CudaInternal::scratch_unified(const std::size_t size) const {
     if (m_scratchUnified)
       Record::decrement(Record::get_record(m_scratchUnified));
 
-    Record *const r = Record::allocate(
-        Kokkos::CudaHostPinnedSpace(), "Kokkos::InternalScratchUnified",
-        (sizeof(ScratchGrain) * m_scratchUnifiedCount));
+    Record *const r =
+        Record::allocate(Kokkos::CudaHostPinnedSpace(),
+                         "Kokkos::InternalScratchUnified", alloc_size);
 
     Record::increment(r);
 
--- a/core/src/Cuda/Kokkos_Cuda_Instance.cpp
+++ b/core/src/Cuda/Kokkos_Cuda_Instance.cpp
@@ -32,6 +32,7 @@
 #include <Cuda/Kokkos_Cuda_UniqueToken.hpp>
 #include <impl/Kokkos_Error.hpp>
 #include <impl/Kokkos_Tools.hpp>
+#include <impl/Kokkos_CheckedIntegerOps.hpp>
 #include <impl/Kokkos_DeviceManagement.hpp>
 #include <impl/Kokkos_ExecSpaceManager.hpp>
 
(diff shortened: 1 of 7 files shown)


masterleinad (changes requested):
Please also fix the other backends that have the same code.

core/src/Cuda/Kokkos_Cuda_Instance.cpp
```diff
@@ -439,12 +440,19 @@ Kokkos::Cuda::initialize WARNING: Cuda is allocating into UVMSpace by default
...
 
 using ScratchGrain = Cuda::size_type[Impl::CudaTraits::WarpSize];
-enum { sizeScratchGrain = sizeof(ScratchGrain) };
+static constexpr auto sizeScratchGrain = sizeof(ScratchGrain);
+
+constexpr std::size_t scratch_count(const std::size_t size) {
+  if (size + sizeScratchGrain - 1 > sizeScratchGrain)  // check overflow
+    return (size + sizeScratchGrain - 1) / sizeScratchGrain;
+  else
+    return std::numeric_limits<std::size_t>::max() / sizeScratchGrain;
```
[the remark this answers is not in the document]
cz4rs: Can this case actually occur or should I not bother with the check here?

core/src/Cuda/Kokkos_Cuda_Instance.cpp
```diff
@@ -439,12 +439,16 @@ Kokkos::Cuda::initialize WARNING: Cuda is allocating into UVMSpace by default
 //----------------------------------------------------------------------------
 
 using ScratchGrain = Cuda::size_type[Impl::CudaTraits::WarpSize];
-enum { sizeScratchGrain = sizeof(ScratchGrain) };
+static constexpr auto sizeScratchGrain = sizeof(ScratchGrain);
+
+constexpr std::size_t scratch_count(const std::size_t size) {
```
dalg24, line 444: The `constexpr` specifier is superfluous.  This will never be evaluated at compile-time.
I think making it a static member function is more pertinent.

core/src/Cuda/Kokkos_Cuda_Instance.cpp
```diff
@@ -439,12 +439,16 @@ Kokkos::Cuda::initialize WARNING: Cuda is allocating into UVMSpace by default
 //----------------------------------------------------------------------------
 
 using ScratchGrain = Cuda::size_type[Impl::CudaTraits::WarpSize];
-enum { sizeScratchGrain = sizeof(ScratchGrain) };
+static constexpr auto sizeScratchGrain = sizeof(ScratchGrain);
```
dalg24, line 442: Define it within `scratch_count` because it is only used there.

cz4rs: Note that `m_scratchFlagsCount * sizeScratchGrain` is the expression that was potentially overflowing in the `if` statement. Not sure how this could be fixed here.

core/src/Cuda/Kokkos_Cuda_Instance.cpp
cz4rs: Is it really worth it to move it into the header since it's only used in the .cpp? What about dropping the `constexpr` and moving it to the anonymous namespace at the top of the file (where we already have helpers like `cuda_kernel_arch`)?

core/src/Cuda/Kokkos_Cuda_Instance.cpp
masterleinad: I guess we should really just error out if we detect that the calculation would overflow since we can't allocate that much memory anyway.

core/src/Cuda/Kokkos_Cuda_Instance.cpp
dalg24: I misread and thought it was defined in the body of the class.  But since you are in the anonymous namespace the static specifier is not needed.

core/src/impl/Kokkos_Utilities.hpp
```diff
@@ -202,6 +202,17 @@ template <typename T>
 constexpr bool dependent_false_v = !sizeof(T*);
 //==============================================================================
 
+template <typename T>
+T check_mul_of(T a, T b) {
+  auto product = a * b;
```
masterleinad, line 207: Note that signed integer overflow is undefined behavior so we should make sure that `T` is unsigned if it's an integer type.

cz4rs: Restricting this to unsigned integrals only.

masterleinad:
Please update the title and description to describe the full scope of this pull request.

cz4rs:
> Please update the title and description to describe the full scope of this pull request.

Done :ok_hand:

core/src/impl/Kokkos_Utilities.hpp
```diff
@@ -202,6 +203,18 @@ template <typename T>
 constexpr bool dependent_false_v = !sizeof(T*);
 //==============================================================================
 
+template <typename T>
+std::enable_if_t<std::is_integral_v<T> && std::is_unsigned_v<T>, T>
+check_mul_of(T a, T b) {
```
dalg24, line 208: I don't like the function template name `check_mul_of` and would like to suggest we consider instead following existing practice like GCC builtins and introduce `add_overflow(x, y , &result) -> bool` or rename to `checked_mul` (maybe easier).

If you go for `checked_mul`, please do static assert rather than sfinae.

core/src/HIP/Kokkos_HIP_Instance.cpp
```diff
@@ -59,9 +59,19 @@ Kokkos::View<uint32_t *, HIPSpace> hip_global_unique_token_locks(
...
-
 namespace Impl {
 
+namespace {
+
+using ScratchGrain = Kokkos::HIP::size_type[Impl::HIPTraits::WarpSize];
+constexpr auto sizeScratchGrain = sizeof(ScratchGrain);
+
+std::size_t scratch_count(const std::size_t size) {
+  return (size + sizeScratchGrain - 1) / sizeScratchGrain;
```
dalg24, line 70: I am not convinced defining that function is a good change but I will leave it up to @Rombur

core/src/impl/Kokkos_Utilities.hpp
cz4rs: Ok, `add_overflow` probably makes more sense looking at upcoming `<stdckdint.h>` / N2683.

cz4rs: Live review suggestions:
- separate header in impl: something along the lines of `checked_integer_operations.h`
- SFINAE on integer
- static_assert( not unsigned, "not implemented yet")
- add unit test (!)

cz4rs:
Looks like HIP failures are unrelated.

Rombur:
Retest this please

core/unit_test/TestCheckedIntegerOps.hpp
```diff
@@ -0,0 +1,39 @@
...
+//
+//@HEADER
+
+#include <gtest/gtest.h>
+#include <impl/Kokkos_CheckedIntegerOps.hpp>
+#include <limits>
+
+namespace Test {
+
+TEST(TEST_CATEGORY, test_multiply) {
```
dalg24, line 23:
```suggestion
TEST(TEST_CATEGORY, checked_integer_operations_multiply_overflow) {
```

masterleinad (approved):
Looks OK to me.

core/unit_test/TestCheckedIntegerOps.hpp
```diff
@@ -0,0 +1,39 @@
...
+    auto result      = 1u;
+    auto is_overflow = Kokkos::Impl::multiply_overflow(1u, 2u, result);
+    EXPECT_EQ(result, 2u);
+    EXPECT_FALSE(is_overflow);
+  }
+  {
+    auto result      = 1u;
+    auto is_overflow = Kokkos::Impl::multiply_overflow(
+        std::numeric_limits<unsigned>::max(), 2u, result);
+    EXPECT_EQ(result, 1u);
```
dalg24, line 34: Do we actually want to guarantee anything about the result when overflow occurs?

cz4rs: I think not (which makes this check superfluous).

[gcc built-ins](https://gcc.gnu.org/onlinedocs/gcc/Integer-Overflow-Builtins.html) and upcoming `<stdckdint.h>` do infinite precision and then cast to result type, but I don't think that it's necessary in our use case.

cz4rs:
New unit test is passing fine on failed builds:
```
1: [ RUN      ] serial.checked_integer_operations_multiply_overflow
1: [       OK ] serial.checked_integer_operations_multiply_overflow (0 ms)
```

```
1: [ RUN      ] openmp.checked_integer_operations_multiply_overflow
1: [       OK ] openmp.checked_integer_operations_multiply_overflow (0 ms)
```

```
4: [ RUN      ] hip.checked_integer_operations_multiply_overflow
4: [       OK ] hip.checked_integer_operations_multiply_overflow (0 ms)
```

```
4: [ RUN      ] sycl.checked_integer_operations_multiply_overflow
4: [       OK ] sycl.checked_integer_operations_multiply_overflow (0 ms)
```

cz4rs:
Retest this please

cz4rs:
Investigating the SYCL failures.

core/src/SYCL/Kokkos_SYCL_Instance.cpp
```diff
@@ -20,10 +20,23 @@
...
+#include <impl/Kokkos_CheckedIntegerOps.hpp>
+
 namespace Kokkos {
 namespace Experimental {
 namespace Impl {
 
+namespace {
+
+static constexpr auto sizeScratchGrain =
+    sizeof(Kokkos::Experimental::SYCL::size_type);
```
masterleinad, line 32:
```suggestion

// FIXME_SYCL Should be a multiple of the maximum subgroup size.
static constexpr auto sizeScratchGrain =
    sizeof(Kokkos::Experimental::SYCL::size_type[32]);
```

masterleinad (approved):

core/src/Cuda/Kokkos_Cuda_Instance.cpp
```diff
@@ -438,47 +446,49 @@ Kokkos::Cuda::initialize WARNING: Cuda is allocating into UVMSpace by default
...
 
     if (m_scratchFlags) Record::decrement(Record::get_record(m_scratchFlags));
 
-    Record *const r =
-        Record::allocate(Kokkos::CudaSpace(), "Kokkos::InternalScratchFlags",
-                         (sizeof(ScratchGrain) * m_scratchFlagsCount));
+    std::size_t alloc_size;
+    if (multiply_overflow(m_scratchFlagsCount, sizeScratchGrain, alloc_size)) {
+      Kokkos::abort("Arithmetic overflow detected.");
+    }
```
Rombur, line 462: I don't find this code intuitive. What about we do sth like
```c++
std::size_t alloc_size = multiply_overflow(m_scratchFlagsCount, sizeScratchGrain);
```
and we put the `Kokkos::abort` inside `multiply_overflow`. The code is easier to read and we don't need to repeat the `Kokkos::abort` everywhere.

cz4rs: See ffe1120.

Rombur (approved):

core/src/HIP/Kokkos_HIP_Instance.cpp
crtrott: @rombur approved, and its what CUDA did so there is some consistency.

core/src/impl/Kokkos_CheckedIntegerOps.hpp
```diff
@@ -0,0 +1,53 @@
...
+  static_assert(std::is_unsigned_v<T>,
+                "Operation not implemented for signed integers.");
+  auto product = a * b;
+  if ((a == 0) || (b == 0) || (a == product / b)) {
+    res = product;
+    return false;
+  } else {
+    return true;
+  }
+}
```
crtrott, line 39: Small follow up for someone: there are builtins we could use here: https://gcc.gnu.org/onlinedocs/gcc/Integer-Overflow-Builtins.html

crtrott (approved):

PR #6160, opened 2023-05-24
[HIP] Improve heuristic deciding the number of blocks used in parallel_reduce

Rombur:
https://github.com/kokkos/kokkos/pull/6029 introduced a regression when using `parallel_reduce` to perform a vector dot product (which is done by PETSc). I have a new heuristic based on a vector dot product benchmark.

@arghdos  @stanmoore1 is this change OK for LAMMPS?

### DIFF
--- a/core/src/HIP/Kokkos_HIP_Parallel_Range.hpp
+++ b/core/src/HIP/Kokkos_HIP_Parallel_Range.hpp
@@ -306,9 +306,7 @@ class ParallelReduce<CombinedFunctorReducerType, Kokkos::RangePolicy<Traits...>,
               Kokkos::Experimental::WorkItemProperty::HintLightWeight;
       static constexpr typename Policy::work_item_property property =
           typename Policy::work_item_property();
-      if ((property & light_weight) == light_weight)
-      // Kokkos::Experimental::WorkItemProperty::HintHeavyWeight_t) {
-      {
+      if ((property & light_weight) == light_weight) {
         if (nblocks < block_size) {
           // Keep nblocks as is
         } else if (nblocks < 16 * block_size) {
--- a/core/src/HIP/Kokkos_HIP_Parallel_Range.hpp
+++ b/core/src/HIP/Kokkos_HIP_Parallel_Range.hpp
@@ -299,8 +299,8 @@ class ParallelReduce<CombinedFunctorReducerType, Kokkos::RangePolicy<Traits...>,
       // use a slightly less constrained, but still well bounded limit for
       // scratch
       int nblocks = (nwork + block.y - 1) / block.y;
-      // Heuristic deciding the value of nblocks. The values for the light weight case
-      // have been chosen using a vector product benchmark on MI250.
+      // Heuristic deciding the value of nblocks. The values for the light
+      // weight case have been chosen using a vector product benchmark on MI250.
       constexpr auto light_weight =
           Kokkos::Experimental::WorkItemProperty::HintLightWeight;
       constexpr typename Policy::work_item_property property;
--- a/core/src/HIP/Kokkos_HIP_Parallel_Range.hpp
+++ b/core/src/HIP/Kokkos_HIP_Parallel_Range.hpp
@@ -298,9 +298,28 @@ class ParallelReduce<CombinedFunctorReducerType, Kokkos::RangePolicy<Traits...>,
       dim3 block(1, block_size, 1);
       // use a slightly less constrained, but still well bounded limit for
       // scratch
-      uint32_t nblocks = static_cast<uint32_t>((nwork + block.y - 1) / block.y);
-      nblocks          = std::min(nblocks, 4096u);
-      m_scratch_space  = ::Kokkos::Impl::hip_internal_scratch_space(
+      int nblocks = (nwork + block.y - 1) / block.y;
+      // Heuristic deciding the value of nblocks. The values have been chosen
+      // using a vector product benchmark on MI250.
+      static constexpr Kokkos::Experimental::WorkItemProperty::HintLightWeight_t
+          light_weight =
+              Kokkos::Experimental::WorkItemProperty::HintLightWeight;
+      static constexpr typename Policy::work_item_property property =
+          typename Policy::work_item_property();
+      if ((property & light_weight) == light_weight)
+      // Kokkos::Experimental::WorkItemProperty::HintHeavyWeight_t) {
+      {
+        if (nblocks < block_size) {
+          // Keep nblocks as is
+        } else if (nblocks < 16 * block_size) {
+          nblocks = block_size;
+        } else {
+          nblocks = 4 * block_size;
+        }
+      } else {
+        nblocks = std::min(nblocks, 4096);
+      }
+      m_scratch_space = ::Kokkos::Impl::hip_internal_scratch_space(
           m_policy.space(), reducer.value_size() * nblocks);
       m_scratch_flags = ::Kokkos::Impl::hip_internal_scratch_flags(
           m_policy.space(), sizeof(size_type));
--- a/core/src/HIP/Kokkos_HIP_Parallel_Range.hpp
+++ b/core/src/HIP/Kokkos_HIP_Parallel_Range.hpp
@@ -299,8 +299,8 @@ class ParallelReduce<CombinedFunctorReducerType, Kokkos::RangePolicy<Traits...>,
       // use a slightly less constrained, but still well bounded limit for
       // scratch
       int nblocks = (nwork + block.y - 1) / block.y;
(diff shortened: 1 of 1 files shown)


core/src/HIP/Kokkos_HIP_Parallel_Range.hpp
```diff
@@ -299,8 +299,16 @@ class ParallelReduce<CombinedFunctorReducerType, Kokkos::RangePolicy<Traits...>,
       // use a slightly less constrained, but still well bounded limit for
       // scratch
       uint32_t nblocks = static_cast<uint32_t>((nwork + block.y - 1) / block.y);
-      nblocks          = std::min(nblocks, 4096u);
-      m_scratch_space  = ::Kokkos::Impl::hip_internal_scratch_space(
+      // Heuristic deciding the value of nblocks. The values have been chosen
+      // using a vector probuct benchmark on MI250.
```
masterleinad, line 303:
```suggestion
      // using a vector product benchmark on MI250.
```

arghdos:
I'll give it a shot -- I am curious why it hurts PETSC as well, if you point me in that direction

Rombur:
For what I see with omniperf, this PR improves L1 cache access.

arghdos:
It looks like this drops RHODO back to where it was before my PR, unfortunately:

image

Seemingly because it drops the # of WGs from 4096 to 1024.

Is the PETSc case captured in an issue anywhere? I think we (AMD) needs to take a deeper look to see if we can find a heuristic that works for both.

Rombur:
> Is the PETSc case captured in an issue anywhere for us to reproduce? I think we (AMD) needs to take a deeper look to see if we can find a heuristic that works for both.

It's just a regular dot product.

crtrott:
I am wondering if we should change the heuristic based on "lightweigth kernel" hint, and or simply based on size of functor?

Rombur:
@arghdos I've updated the PR. By default, users get the behavior that you had implemented. If they say the kernel is lightweight then we use the new heuristic.

arghdos:
@Rombur -- ok I can confirm this restores RHODO perf.  I think this is OK in the short term, but it's a bit of a band-aid (heavyweight shouldn't _really_ have anything to do with it as far as I can tell).

I was out on vacation last week, but I am still digging into the root cause of the regression here so we can get a true fix.

core/src/HIP/Kokkos_HIP_Parallel_Range.hpp
```diff
@@ -298,9 +298,28 @@ class ParallelReduce<CombinedFunctorReducerType, Kokkos::RangePolicy<Traits...>,
...
+      int nblocks = (nwork + block.y - 1) / block.y;
+      // Heuristic deciding the value of nblocks. The values have been chosen
+      // using a vector product benchmark on MI250.
+      static constexpr Kokkos::Experimental::WorkItemProperty::HintLightWeight_t
+          light_weight =
+              Kokkos::Experimental::WorkItemProperty::HintLightWeight;
+      static constexpr typename Policy::work_item_property property =
+          typename Policy::work_item_property();
+      if ((property & light_weight) == light_weight)
+      // Kokkos::Experimental::WorkItemProperty::HintHeavyWeight_t) {
```
dalg24, line 310: What is this commented line about?

dalg24 (approved):
I would prefer if you applied my suggestion.

core/src/HIP/Kokkos_HIP_Parallel_Range.hpp
```diff
@@ -298,9 +298,26 @@ class ParallelReduce<CombinedFunctorReducerType, Kokkos::RangePolicy<Traits...>,
...
-      uint32_t nblocks = static_cast<uint32_t>((nwork + block.y - 1) / block.y);
-      nblocks          = std::min(nblocks, 4096u);
-      m_scratch_space  = ::Kokkos::Impl::hip_internal_scratch_space(
+      int nblocks = (nwork + block.y - 1) / block.y;
+      // Heuristic deciding the value of nblocks. The values have been chosen
+      // using a vector product benchmark on MI250.
+      static constexpr Kokkos::Experimental::WorkItemProperty::HintLightWeight_t
+          light_weight =
+              Kokkos::Experimental::WorkItemProperty::HintLightWeight;
+      static constexpr typename Policy::work_item_property property =
```
dalg24, line 307:
```suggestion
      static constexpr typename Policy::work_item_property property;
```

core/src/HIP/Kokkos_HIP_Parallel_Range.hpp
```diff
@@ -298,9 +298,26 @@ class ParallelReduce<CombinedFunctorReducerType, Kokkos::RangePolicy<Traits...>,
       dim3 block(1, block_size, 1);
       // use a slightly less constrained, but still well bounded limit for
       // scratch
-      uint32_t nblocks = static_cast<uint32_t>((nwork + block.y - 1) / block.y);
-      nblocks          = std::min(nblocks, 4096u);
-      m_scratch_space  = ::Kokkos::Impl::hip_internal_scratch_space(
+      int nblocks = (nwork + block.y - 1) / block.y;
+      // Heuristic deciding the value of nblocks. The values have been chosen
+      // using a vector product benchmark on MI250.
+      static constexpr Kokkos::Experimental::WorkItemProperty::HintLightWeight_t
```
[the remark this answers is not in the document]
dalg24:
```suggestion
     constexpr auto light_weight = Kokkos::Experimental::WorkItemProperty::HintLightWeight;
```

core/src/HIP/Kokkos_HIP_Parallel_Range.hpp
```diff
@@ -301,11 +301,9 @@ class ParallelReduce<CombinedFunctorReducerType, Kokkos::RangePolicy<Traits...>,
...
       // Heuristic deciding the value of nblocks. The values have been chosen
       // using a vector product benchmark on MI250.
-      static constexpr Kokkos::Experimental::WorkItemProperty::HintLightWeight_t
-          light_weight =
-              Kokkos::Experimental::WorkItemProperty::HintLightWeight;
-      static constexpr typename Policy::work_item_property property =
-          typename Policy::work_item_property();
+      constexpr auto light_weight =
+          Kokkos::Experimental::WorkItemProperty::HintLightWeight;
+      constexpr auto property = typename Policy::work_item_property();
```
dalg24, line 306: (Optional) drop the parentheses.

(I already approved)

core/src/HIP/Kokkos_HIP_Parallel_Range.hpp
```diff
@@ -298,9 +298,24 @@ class ParallelReduce<CombinedFunctorReducerType, Kokkos::RangePolicy<Traits...>,
       dim3 block(1, block_size, 1);
       // use a slightly less constrained, but still well bounded limit for
       // scratch
-      uint32_t nblocks = static_cast<uint32_t>((nwork + block.y - 1) / block.y);
-      nblocks          = std::min(nblocks, 4096u);
-      m_scratch_space  = ::Kokkos::Impl::hip_internal_scratch_space(
+      int nblocks = (nwork + block.y - 1) / block.y;
+      // Heuristic deciding the value of nblocks. The values have been chosen
+      // using a vector product benchmark on MI250.
```
crtrott, line 303:
```suggestion
      // Heuristic deciding the value of nblocks. The values for the light_weight path
      // have been chosen using a vector product benchmark on MI250.
```

crtrott, line 303:
```suggestion
      // Heuristic deciding the value of nblocks. The values for the light_weigth case
      // have been chosen using a vector product benchmark on MI250.
```

core/src/HIP/Kokkos_HIP_Parallel_Range.hpp
```diff
@@ -298,9 +298,24 @@ class ParallelReduce<CombinedFunctorReducerType, Kokkos::RangePolicy<Traits...>,
       dim3 block(1, block_size, 1);
       // use a slightly less constrained, but still well bounded limit for
       // scratch
-      uint32_t nblocks = static_cast<uint32_t>((nwork + block.y - 1) / block.y);
-      nblocks          = std::min(nblocks, 4096u);
-      m_scratch_space  = ::Kokkos::Impl::hip_internal_scratch_space(
+      int nblocks = (nwork + block.y - 1) / block.y;
+      // Heuristic deciding the value of nblocks. The values for the light weight case
+      // have been chosen using a vector product benchmark on MI250.
```
crtrott, line 303:
```suggestion
      // Heuristic deciding the value of nblocks. The values for the light
      // weight case have been chosen using a vector product benchmark on MI250.
```

crtrott (approved):

PR #6162, opened 2023-05-24
snapshot mdspan namespace changes

### LINKED ISSUE #6161: merge in mdspan namespace changes
This prevents namespace conflicts with mdspan's use of "std"

nmm0:
Closes #6161 

Snapshotted https://github.com/kokkos/mdspan/commit/124b860f458e5c06c9b96d7510dc35b7acdd642b


crtrott (approved):

core/src/View/MDSpan/Kokkos_MDSpan_Extents.hpp
```diff
@@ -63,7 +63,7 @@ struct DimensionFromExtent {
 };
 
 template <>
-struct DimensionFromExtent<std::experimental::dynamic_extent> {
+struct DimensionFromExtent<dynamic_extent> {
```
dalg24, line 66: Any particular reason why you decided to qualify at other places and not here?

nmm0: oh woops -- I assume you meant line 57 but that was left in from something earlier so I'll make it consistent

masterleinad: This line (66) really is the only one where an unqualified name is used and I also would prefer consistency.

nmm0: Just pushed some changes fixing this

masterleinad: You changed it the other way around, though. Now we have two places with unqualified names and all other cases use qualified names.

nmm0: I missed one more, but all the mdspan stuff is now in namespace Kokkos so there is no need to actually qualify. It's only qualified in the tests (and the one more that I accidentally missed).

dalg24 (changes requested):
What version of MDSPAN is being snapshotted ?

nmm0:
> What version of MDSPAN is being snapshotted ?

https://github.com/kokkos/mdspan/pull/271

So best thing would be to review that PR so we can get it merged and then we will be based off stable. That PR is required for this PR to compile.

dalg24:
Please update your PR to reflect was actually got merged into kokkos/mdspan

nmm0:
Only difference from before besides a newer snapshot is I removed `<experimental/mdspan>` and `<experimental/mdarray>` headers as they may conflict and we aren't using them anyway

@crtrott @masterleinad OK with you?

core/src/View/MDSpan/Kokkos_MDSpan_Header.hpp
```diff
@@ -25,10 +25,13 @@ static_assert(false,
 // Look for the right mdspan
 #if __has_include(<mdspan>)
```
crtrott, line 26: Two things: 
- also check for the feature test macro so we don't pull in partial implementation like in clang right now.
- don't we need to do a `using std::layout_right/left/stride` and `using std::mdspan` too?

crtrott (changes requested):
I think the logic for mdspan from compile inclusion is incomplete.

crtrott:
Oh and should we for now disable the "import the compiler provided one" if we compile for CUDA/HIP? So we get the markup? We can later figure out whether we can use the one from nvc++ etc.

masterleinad (approved):
Please make sure we test `mdspan` support in our CI. As far as I can see, there is no build with `Kokkos_ENABLE_IMPL_MDSPAN=ON`.

masterleinad (changes requested):
Please make sure we test `mdspan` support in our CI. As far as I can see, there is no build with `Kokkos_ENABLE_IMPL_MDSPAN=ON`.

core/src/View/MDSpan/Kokkos_MDSpan_Header.hpp
nmm0: done, also added in `std::default_accessor`

masterleinad (commented):
`CUDA-11.7-NVHPC` is failing with
````
"/var/jenkins/workspace/Kokkos/core/src/View/MDSpan/Kokkos_MDSpan_Header.hpp", line 40: catastrophic error: cannot open source file "mdspan/mdspan.hpp"
  #include <mdspan/mdspan.hpp>
                              ^
1 catastrophic error detected in the compilation of "/var/jenkins/workspace/Kokkos/core/src/impl/Kokkos_ExecPolicy.cpp".
Compilation terminated.
"/var/jenkins/workspace/Kokkos/core/src/View/MDSpan/Kokkos_MDSpan_Header.hpp", line 40: catastrophic error: cannot open source file "mdspan/mdspan.hpp"
  #include <mdspan/mdspan.hpp>
                              ^
```
I'm not quite sure if testing with this compiler is crucial for this pull request (given our experience that nvc++ is pretty buggy anyway).

core/src/View/MDSpan/Kokkos_MDSpan_Header.hpp
```diff
@@ -23,12 +23,21 @@ static_assert(false,
 #define KOKKOS_EXPERIMENTAL_MDSPAN_HPP
 
 // Look for the right mdspan
-#if __has_include(<mdspan>)
+#include <version>
```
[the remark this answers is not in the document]
crtrott: The problem is partial implementation in clang or whatever. So having the header doesn't mean all of mdspan is implemented.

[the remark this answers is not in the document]
nmm0: Ah you are right. No we don't expect that, so I'll guard with a C++ version test

crtrott (changes requested):
Still think we don't include standard mdspan for now with CUDA and HIP, and then later see if we can use say the one from nvhpc.

core/src/View/MDSpan/Kokkos_MDSpan_Header.hpp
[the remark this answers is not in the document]
masterleinad: > The problem is partial implementation in clang or whatever. So having the header doesn't mean all of mdspan is implemented.

Right, but at least the feature macro should exist if the header file is available.

masterleinad (approved):
Looks OK to me.

crtrott (approved):

nmm0:
@masterleinad I'm not 100% sure what's going on in either of those. While the submdspan header *used* to not have std:: qualified tuple it should now... I'll investigate after I get to Varna

masterleinad:
> @masterleinad I'm not 100% sure what's going on in either of those. While the submdspan header _used_ to not have std:: qualified tuple it should now... I'll investigate after I get to Varna

Looks like @crtrott took care of it.

crtrott:
Retest this please!

nmm0:
Retest this please

nmm0:
I'm not sure what's going on here, the jenkins CUDA jobs seem to be failing, but without any output.

Rombur:
There is an output:

ERROR: The container started but didn't run the expected command. Please double check your ENTRYPOINT does execute the command passed as docker run argument, as required by official docker images (see https://github.com/docker-library/official-images#consistency for entrypoint consistency requirements).
Alternatively you can force image entrypoint to be disabled by adding option `--entrypoint=''`.

It seems that there is an issue with the images but I have no clue what that is.

crtrott:
Retest this please.

dalg24:
Please make sure that the description is up to date (capture the right version of kokkos/mdspan being snapshotted)

PR #6164, opened 2023-05-26
Avoid using TriBITS in Kokkos

masterleinad:
This corresponds to https://github.com/trilinos/Trilinos/pull/11779.

Currently, we are essentially using three build systems in `Kokkos`:
- Makefile-based (only supported for using Kokkos inline, not standalone)
- raw `CMake`
- through `Trilinos` using `TriBITS`

We discussed that we want to try to lighten that burden by removing a full `TriBITS` option by only providing necessary additions to the raw `CMake` build system. This pull request explores doing this.

The approach taken is to simply delete all paths that mention `KOKKOS_HAS_TRILINOS` or `TRIBITS` directly which seems to work pretty well. The places where we still need to do something special (apart from adding additional aliases or making sure we can link against a Kokkos build directory) are:
- ~~Calling `TRIBITS_PACKAGE(Kokkos)` at the top of `CMakeLists.txt`~~
- Dealing with `Trilinos` options in the new `kokkos_configure_trilinos.cmake`
- ~~Calling `tribits_package_postprocess` in `kokkos_configure_trilinos.cmake`~~
- ~~Dealing with `Trilinos` TPL's `Kokkos` provides/uses.~~
- cmake/Dependencies.cmake

Note that we also remove all the special casing for running `Kokkos` tests in `Trilinos`.


masterleinad:
@bartlettroscoe I revived this after you asked for feedback wrt https://github.com/TriBITSPub/TriBITS/pull/591#issuecomment-1731599860.
It seems we neither need to call 
 - TRIBITS_PACKAGE(Kokkos) at the top of CMakeLists.txt, nor
 - tribits_package_postprocess in kokkos_configure_trilinos.cmake
 
That leaves us basically with additional code for 
  - dealing with `Trilinos` options in the new `kokkos_configure_trilinos.cmake`, and
  - dealing with `Trilinos` TPL's `Kokkos` provides/uses.

Also, `Trilinos` seems to have a problem with us exporting `KokkosTargets.cmake` twice, once to `CMAKE_BINARY_DIR}/cmake_packages/Kokkos/KokkosTargets.cmake`, and once to `export(EXPORT KokkosTargets NAMESPACE Kokkos:: FILE ${Kokkos_BINARY_DIR}/KokkosTargets.cmake)` (which is a more natural choice for supporting linking to the build tree; at least in a single repository build). It seems that `including` the former in the latter provides a workaround.

Apart from that, everything seems to be working here as described in the files you linked.

bartlettroscoe:
@masterleinad, one thing you might consider is to keep using the TriBITS function `tribits_add_test()` under your `kokkos_add_test()` function when building Kokkos under Trilinos.  You can now use those functions apart from the rest of TriBITS as described in the new Howto section:

* https://tribitspub.github.io/TriBITS/users_guide/index.html#how-to-use-tribits-testing-support-in-non-tribits-project

But in the case of building Kokkos under Trilinos, you don't need to include the module `TribitsAddTest.cmake` (the TriBITS framework already includes that before processing the package CMakeLists.txt files) or define the variable `Kokkos_ENABLE_TESTS` (TriBITS defines and sets that automatically).  You just need to call the function `tribits_add_test()`.

Using `tribits_add_test()` will allow the Trilinos framework team and other developers/users to continue to surgically disable Kokkos tests on particular configurations and platforms without having to touch anything under `packages/kokkos/`. Otherwise, it would not be very easy to maintain running the Kokkos test suite on all of the different Trilinos build configurations on all of the different platforms.

Just something to think about.

masterleinad:
Sure, we should discuss with the rest of the `Kokkos` team if we want to keep `tribits_add_test`.

maartenarnst:
Hi @masterleinad. I was just wondering about the current status of the work on this work on the build systems?

I am asking because we are interested in activating `rocthrust` in our `rocm` builds. However, currently Kokkos disables `rocthrust` when Kokkos is built within Trilinos. This was discussed in 

- https://github.com/kokkos/kokkos/pull/6871#issuecomment-1997282161

and this work on the build systems was suggested as a possible way forward for the `rocthrust` issue. Thanks in advance!

masterleinad:
Set to "ready for review" to initiate some discussion.

core/unit_test/CMakeLists.txt
```diff
@@ -1011,14 +1011,15 @@ KOKKOS_ADD_TEST_EXECUTABLE( push_finalize_hook_terminate
...
-)
+# FIXME_REFACTOR_TRIBITS
+#KOKKOS_ADD_ADVANCED_TEST( CoreUnitTest_PushFinalizeHook_terminate
+#  TEST_0
+#    EXEC push_finalize_hook_terminate
+#    NUM_MPI_PROCS 1
+#    PASS_REGULAR_EXPRESSION
+#      "PASSED: I am the custom std::terminate handler."
+#    ALWAYS_FAIL_ON_ZERO_RETURN
+#)
```
dalg24, line 1022: You might want to rebase.  I got rid of this in favor of raw cmake in #7133

cmake/kokkos_tribits.cmake
dalg24, line 1: Just a note that we might want to rename that file.
Can be done as a follow up though.

dalg24 (commented):
I like the general direction

masterleinad:
@ndellingwood Would you mind testing this pulll request with Trilinos?

ndellingwood:
@masterleinad yes, I'll test several configurations sometime later today. Would testing a subset of packages that heavily use kokkos (e.g. kokkos, kokkos-kernels, tpetra, sacado, stokhos), as well testing a mini-app against an install suffice?

ndellingwood:
@masterleinad a couple data points of testing so far - Trilinos build with cuda/11.4.2 was not impacted by this PR. Tests included KokkosKernels, Sacado, Stokhos, Teuchos, Tpetra, Amesos2, Ifpack2, Zoltan2, and MueLu. However, Kokkos tests did not build and run, even when explicitly setting `-D Kokkos_ENABLE_TESTS=ON`, is this to be expected with these updates?

masterleinad:
@ndellingwood Thanks! That is inline with my testing. The only thing not working yet is disabling tests.

cmake/kokkos_install.cmake
```diff
@@ -36,21 +36,18 @@ IF (NOT KOKKOS_HAS_TRILINOS AND NOT Kokkos_INSTALL_TESTING)
             ${Kokkos_BINARY_DIR}/KokkosConfigCommon.cmake
             ${Kokkos_BINARY_DIR}/KokkosConfigVersion.cmake
             DESTINATION ${CMAKE_BINARY_DIR}/cmake_packages/Kokkos)
-  export(EXPORT KokkosTargets NAMESPACE Kokkos:: FILE ${CMAKE_BINARY_DIR}/cmake_packages/Kokkos/KokkosTargets.cmake)
+  file(WRITE ${CMAKE_BINARY_DIR}/cmake_packages/Kokkos/KokkosTargets.cmake "include(${Kokkos_BINARY_DIR}/KokkosTargets.cmake)")
 ELSE()
   CONFIGURE_FILE(cmake/KokkosConfigCommon.cmake.in ${Kokkos_BINARY_DIR}/KokkosConfigCommon.cmake @ONLY)
   file(READ ${Kokkos_BINARY_DIR}/KokkosConfigCommon.cmake KOKKOS_CONFIG_COMMON)
   file(APPEND "${CMAKE_CURRENT_BINARY_DIR}/CMakeFiles/KokkosConfig_install.cmake" "${KOKKOS_CONFIG_COMMON}")
```
dalg24, line 43: Do we want to keep that file?

cmake/kokkos_install.cmake
```diff
@@ -36,21 +36,18 @@ IF (NOT KOKKOS_HAS_TRILINOS AND NOT Kokkos_INSTALL_TESTING)
             ${Kokkos_BINARY_DIR}/KokkosConfigCommon.cmake
             ${Kokkos_BINARY_DIR}/KokkosConfigVersion.cmake
             DESTINATION ${CMAKE_BINARY_DIR}/cmake_packages/Kokkos)
-  export(EXPORT KokkosTargets NAMESPACE Kokkos:: FILE ${CMAKE_BINARY_DIR}/cmake_packages/Kokkos/KokkosTargets.cmake)
+  file(WRITE ${CMAKE_BINARY_DIR}/cmake_packages/Kokkos/KokkosTargets.cmake "include(${Kokkos_BINARY_DIR}/KokkosTargets.cmake)")
```
dalg24, line 39: Not following that change

masterleinad: Without this, I'm getting
```
CMake Error in packages/sacado/CMakeLists.txt:
  export called with target "sacado" which requires target "kokkos" that is
  not in this export set, but in multiple other export sets:
  /Users/6da/Software/trilinos_new/build/cmake_packages/Kokkos/KokkosTargets.cmake,
  /Users/6da/Software/trilinos_new/build/kokkos/KokkosTargets.cmake.

  An exported target cannot depend upon another target which is exported
  multiple times.  Consider consolidating the exports of the "kokkos" target
  to a single export.


CMake Error in packages/teuchos/core/CMakeLists.txt:
  export called with target "teuchoscore" which requires target "kokkos" that
  is not in this export set, but in multiple other export sets:
  /Users/6da/Software/trilinos_new/build/cmake_packages/Kokkos/KokkosTargets.cmake,
  /Users/6da/Software/trilinos_new/build/kokkos/KokkosTargets.cmake.

  An exported target cannot depend upon another target which is exported
  multiple times.  Consider consolidating the exports of the "kokkos" target
  to a single export.


CMake Error in packages/teuchos/kokkoscomm/CMakeLists.txt:
  export called with target "teuchoskokkoscomm" which requires target
  "kokkos" that is not in this export set, but in multiple other export sets:
  /Users/6da/Software/trilinos_new/build/cmake_packages/Kokkos/KokkosTargets.cmake,
  /Users/6da/Software/trilinos_new/build/kokkos/KokkosTargets.cmake.

  An exported target cannot depend upon another target which is exported
  multiple times.  Consider consolidating the exports of the "kokkos" target
  to a single export.


CMake Error in packages/teuchos/kokkoscompat/CMakeLists.txt:
  export called with target "teuchoskokkoscompat" which requires target
  "kokkos" that is not in this export set, but in multiple other export sets:
  /Users/6da/Software/trilinos_new/build/cmake_packages/Kokkos/KokkosTargets.cmake,
  /Users/6da/Software/trilinos_new/build/kokkos/KokkosTargets.cmake.

  An exported target cannot depend upon another target which is exported
  multiple times.  Consider consolidating the exports of the "kokkos" target
  to a single export.
```
when trying to configure `Trilinos`.

masterleinad:
@ndellingwood Would you mind testing that
   * `-D <fullTestName>_DISABLE=ON` 
  * `-D <fullTestName>_EXTRA_ARGS="<arg0>;<arg1>;<arg2>;..."`
   * `-D <fullTestName>_SET_RUN_SERIAL=ON`

still work as expected from within `Trilinos` with this pull request?

ndellingwood:
> @ndellingwood Would you mind testing that
> 
>     * `-D <fullTestName>_DISABLE=ON`
> 
>     * `-D <fullTestName>_EXTRA_ARGS="<arg0>;<arg1>;<arg2>;..."`
> 
>     * `-D <fullTestName>_SET_RUN_SERIAL=ON`
> 
> 
> still work as expected from within `Trilinos` with this pull request?

@masterleinad I retested in a gcc/8.5.0 build and `_DISABLE` and `_EXTRA_ARGS` are working as expected

For `_SET_RUN_SERIAL` it seems to work as expected; I ran with `ctest -j4` and the test I designated with this property was the final test in the list, out of its typical order (I believe the run serial property collects the tests to run sequentially and tests them at the end)
```
      Start  2: Kokkos_CoreUnitTest_Default
41/41 Test  #2: Kokkos_CoreUnitTest_Default ..........................   Passed    0.02 sec
```

I tried the `_ENVIRONMENT` variable as well to filter out a test, and though it was not rejected during configuration it did not filter the test (maybe this is expected)

Here are my relevant configuration options:
```
 -D Kokkos_CoreUnitTest_Serial1_DISABLE=ON \
 -D Kokkos_CoreUnitTest_Default_SET_RUN_SERIAL=ON \
 -D Kokkos_UnitTest_Random_EXTRA_ARGS="--gtest_filter=-serial.Random_XorShift64" \
 -D Kokkos_UnitTest_Random_ENVIRONMENT="GTEST_FILTER=-serial.Random_XorShift1024_0" \
```


Running ctest, I see the environment variable `GTEST_FILTER=-serial.Random_XorShift1024_0` mentioned, but the test runs anyway, only `Random_XorShift64` is removed
```
    Start 22: Kokkos_UnitTest_Random

22: Test command: /home/ndellin/Trilinos/Build/SemsGcc850-kokkos_dev/kokkos/algorithms/unit_tests/Kokkos_UnitTest_Random "--gtest_filter=-serial.Random_XorShift64"
22: Working Directory: /home/ndellin/Trilinos/Build/SemsGcc850-kokkos_dev/kokkos/algorithms/unit_tests
22: Environment variables:
22:  GTEST_FILTER=-serial.Random_XorShift1024_0
22: Test timeout computed to be: 1500
22: Note: Google Test filter = -serial.Random_XorShift64
22: [==========] Running 2 tests from 1 test suite.
22: [----------] Global test environment set-up.
22: [----------] 2 tests from serial
22: [ RUN      ] serial.Random_XorShift1024_0
...
```

dalg24:
It is expected.  The `--gtest_filter=-serial.Random_XorShift64` command line argument takes precedence over the `GTEST_FILTER=-serial.Random_XorShift1024_0` environment variable.
You would need to test `ENVIRONMENT` by itself or set a different option, possibly `GTEST_COLOR=yes`.

masterleinad:
@ndellingwood Are you good with merging this or do you want to pull in someone else from the Trilinos-side before?

ndellingwood:
@masterleinad it would be good to have @bartlettroscoe in the loop and any feedback he may have

ndellingwood:
@masterleinad @dalg24 I think we are hitting a bump with the changes here. So far, the AT2 builds in https://github.com/trilinos/Trilinos/pull/13423 are failing due to a change in behavior triggering enable of the `Kokkos::Threads` backend in those jobs that did not happen prior to these changes. It might be triggered due to an option like `-DTPL_Enable_Pthreads=ON` or something similar, I still need to confirm

masterleinad:
> @masterleinad @dalg24 I think we are hitting a bump with the changes here. So far, the AT2 builds in [trilinos/Trilinos#13423](https://github.com/trilinos/Trilinos/pull/13423) are failing due to a change in behavior triggering enable of the `Kokkos::Threads` backend in those jobs that did not happen prior to these changes. It might be triggered due to an option like `-DTPL_Enable_Pthreads=ON` or something similar, I still need to confirm

Let me also try a local build with the `Threads` backend.

masterleinad:
> Let me also try a local build with the `Threads` backend.

That worked fine.

CMakeLists.txt
```diff
@@ -72,7 +63,6 @@ ENDFUNCTION()
 
 LIST(APPEND CMAKE_MODULE_PATH cmake/Modules)
 
-IF(NOT KOKKOS_HAS_TRILINOS)
   set(CMAKE_DISABLE_SOURCE_CHANGES ON)
   set(CMAKE_DISABLE_IN_SOURCE_BUILD ON)
```
aprokop, line 67: Change `set` -> `SET` for consistency (drive-by).

cmake/kokkos_install.cmake
```diff
@@ -36,21 +36,16 @@ IF (NOT KOKKOS_HAS_TRILINOS AND NOT Kokkos_INSTALL_TESTING)
...
-  file(READ ${Kokkos_BINARY_DIR}/KokkosTrilinosConfig.cmake KOKKOS_TRILINOS_CONFIG)
-  file(APPEND "${CMAKE_CURRENT_BINARY_DIR}/CMakeFiles/KokkosConfig_install.cmake" "${KOKKOS_TRILINOS_CONFIG}")
 
   WRITE_BASIC_PACKAGE_VERSION_FILE("${CMAKE_CURRENT_BINARY_DIR}/KokkosConfigVersion.cmake"
       VERSION "${Kokkos_VERSION}"
       COMPATIBILITY AnyNewerVersion)
 
   install(FILES ${CMAKE_CURRENT_BINARY_DIR}/KokkosConfigVersion.cmake
-      DESTINATION "${${PROJECT_NAME}_INSTALL_LIB_DIR}/cmake/${PACKAGE_NAME}")
+      DESTINATION "${${PROJECT_NAME}_INSTALL_LIB_DIR}/cmake/Kokkos")
```
aprokop, line 48: Why this change? Wouldn't it be better to reference a variable?

core/perf_test/CMakeLists.txt
```diff
@@ -82,7 +78,7 @@ FUNCTION(KOKKOS_ADD_BENCHMARK NAME)
     )
   ENDIF()
 
-  SET(BENCHMARK_NAME ${PACKAGE_NAME}_${NAME})
+  SET(BENCHMARK_NAME Kokkos_${NAME})
```
aprokop, line 81: Why is hardcoded name preferred?

cmake/kokkos_configure_trilinos.cmake
```diff
@@ -0,0 +1,35 @@
+IF(CMAKE_PROJECT_NAME STREQUAL "Trilinos")
+  set(Kokkos_ENABLE_SERIAL ON CACHE BOOL "Whether to build Serial backend" FORCE)
+
+  if(NOT ${Trilinos_ENABLE_OpenMP} STREQUAL "")
+    set(Kokkos_ENABLE_OPENMP ${Trilinos_ENABLE_OpenMP} CACHE BOOL "Whether to build OpenMP backend" FORCE)
+  else()
+    set(Kokkos_ENABLE_OPENMP                       OFF CACHE BOOL "Whether to build OpenMP backend" FORCE)
+  endif()
```
aprokop, line 8: Wonder if this repeated code block could be done with a single macro.

cmake/fake_tribits.cmake
```diff
@@ -34,88 +32,59 @@ MACRO(ADD_INTERFACE_LIBRARY LIB_NAME)
...
+    ADD_TEST(NAME ${TEST_NAME} COMMAND ${EXE} ${TEST_ARGS} ${${TEST_NAME}_EXTRA_ARGS})
+  ENDIF()
+  IF(${TEST_NAME}_SET_RUN_SERIAL)
+    SET_TESTS_PROPERTIES(${TEST_NAME} PROPERTIES RUN_SERIAL ON)
+  ENDIF()
+  # TriBITS doesn't actually currently support `-D <fullTestName>_ENVIRONMENT`
+  # but we decided to add it anyway
+  IF(${TEST_NAME}_ENVIRONMENT)
+    SET_TESTS_PROPERTIES(${TEST_NAME} PROPERTIES ENVIRONMENT "${${TEST_NAME}_ENVIRONMENT}")
+  ENDIF()
```
aprokop, line 73: I don't think I follow the reasoning for adding it.

aprokop (commented):
A lot of `IF()` clauses were removed, but the block inside them has not been shifted left, making it inconsistent with the surrounding text. Would be nice to have it all properly aligned.

What's the role of `kokkos_tribits.cmake` and `fake_tribits.cmake` now?

masterleinad:
> A lot of `IF()` clauses were removed, but the block inside them has not been shifted left, making it inconsistent with the surrounding text. Would be nice to have it all properly aligned.

We could just run https://github.com/cheshirekow/cmake_format after this pull request and enforce it in the CI.

aprokop:
> We could just run [cheshirekow/cmake_format](https://github.com/cheshirekow/cmake_format) after this pull request and enforce it in the CI.

Something to discuss for sure.

bartlettroscoe (commented):
Why not just call `tribits_add_test()` when that function is defined, like when Kokkos is built as part of Trilinos?  Otherwise, a customer may not trust an install of Trilinos with Kokkos if I could not run the native Kokkos tests.  Once this merges to Trilinos 'develop' (i.e., trilinos/Trilinos#13423) and the Kokkos tests are turned off for a while, it could be very hard to turn them on again in the future.  So please consider this carefully.

Otherwise, this is actually a fairly small number of lines of code removed to remove TriBITS.  (That just shows how well Kokkos did with their own CMake function wrappers.)

cmake/fake_tribits.cmake
```diff
@@ -34,88 +32,59 @@ MACRO(ADD_INTERFACE_LIBRARY LIB_NAME)
...
-      RETURN()
-    ENDIF()
-
-    IF(TEST_EXE)
-      SET(EXE_ROOT ${TEST_EXE})
-    ELSE()
-      SET(EXE_ROOT ${TEST_NAME})
-    ENDIF()
-
-    TRIBITS_ADD_TEST(
```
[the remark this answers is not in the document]
ndellingwood: @bartlettroscoe just to confirm the changes are not intended to remove testing of kokkos within Trilinos, I've been able to build the kokkos tests in Trilinos with the changes in this PR, and the AT2 jobs also built the kokkos tests, though maybe that is unexpected since `tribits_add_test` is not used?
It's not clear to me why the tests are not building as expected in the AT1 jobs (there is some issue with which version or how gtest was found)

cmake/fake_tribits.cmake
masterleinad: We can discuss to remove it now that we support `_EXTRA_ARGS`.

core/perf_test/CMakeLists.txt
masterleinad: I didn't see much of a reason to keep a variable that always has the same value for it.

CMakeLists.txt
masterleinad: Let's try to make the CMake style uniform in a later pull request.

cedricchevalier19 (approved):
Looks great with the eyes of a non-trilinos user.

And I am also in favor to put explicitly the Kokkos name instead of macros.

cmake/kokkos_configure_trilinos.cmake
masterleinad: Not sure if the added complexity is worth it for these few uses.

aprokop (approved):
LGTM

cmake/kokkos_configure_trilinos.cmake
```diff
@@ -0,0 +1,38 @@
+IF(CMAKE_PROJECT_NAME STREQUAL "Trilinos")
+  set(Kokkos_ENABLE_SERIAL ON CACHE BOOL "Whether to build Serial backend" FORCE)
+
+  if(NOT ${Trilinos_ENABLE_OpenMP} STREQUAL "")
```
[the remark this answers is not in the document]
dalg24: Whatever it's fine as proposed if it works

cmake/fake_tribits.cmake
```diff
@@ -34,88 +32,61 @@ MACRO(ADD_INTERFACE_LIBRARY LIB_NAME)
...
-    VERIFY_EMPTY(KOKKOS_ADD_TEST ${TEST_UNPARSED_ARGUMENTS})
+  SET(EXE ${PACKAGE_NAME}_${EXE_ROOT})
+  IF(WIN32)
+    ADD_TEST(NAME ${TEST_NAME} WORKING_DIRECTORY ${LIBRARY_OUTPUT_PATH}
+      COMMAND ${EXE}${CMAKE_EXECUTABLE_SUFFIX} ${TEST_ARGS} ${${TEST_NAME}_EXTRA_ARGS})
+  ELSE()
+    ADD_TEST(NAME ${TEST_NAME} COMMAND ${EXE} ${TEST_ARGS} ${${TEST_NAME}_EXTRA_ARGS})
+  ENDIF()
+  # Trilinos testing benefits from labeling the tests as "Kokkos" tests
+  SET_TESTS_PROPERTIES(${TEST_NAME} PROPERTIES LABELS Kokkos)
```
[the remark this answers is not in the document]
bartlettroscoe: > Were these LABELS set when using TriBITS?

Yes, and that is critical so these tests get handled by CDash correctly for the CDash "subproject" Kokkos.  (Forgot to add that to requirements in https://tribitspub.github.io/TriBITS/users_guide/index.html#tribits-compliant-internal-packages.)

CMakeLists.txt
```diff
@@ -72,7 +63,6 @@ ENDFUNCTION()
 
 LIST(APPEND CMAKE_MODULE_PATH cmake/Modules)
 
-IF(NOT KOKKOS_HAS_TRILINOS)
```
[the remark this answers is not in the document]
dalg24: There is ongoing work to systematically format our cmake files #7356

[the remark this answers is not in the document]
nmm0: Sounds good

ndellingwood:
@masterleinad this should get a changelog entry?

masterleinad:
> @masterleinad this should get a changelog entry?

I don't think we need one.

dalg24:
Mention a major refactor in the build system and we can decide whether to keep it or not at the time of the release.

PR #4096, opened 2021-06-15
Adding ExecutionSpace partitioning function

crtrott:
This allows the creation of multiple instances, but only CUDA implements an actual implementation, everyone else returns the same thing multiple times for now.

The design is as a customization point, which would allow for ADL hopefully.

PR #6167, opened 2023-05-30
Allow passing a temporary std::vector to partition_space

masterleinad:
Related to https://kokkosteam.slack.com/archives/C5BGU5NDQ/p1685215235868969. There doesn't seem to be a good reason not to allow passing a temporary `std::vector` to `partition_space`.
Also, this overload is not mentioned at https://kokkos.github.io/kokkos-core-wiki/API/core/spaces/partition_space.html?highlight=partition_space#_CPPv4I0DpE15partition_spaceNSt6vectorI9ExecSpaceEERK9ExecSpaceDp4Args.

### DIFF
--- a/core/src/Cuda/Kokkos_Cuda_Instance.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_Instance.hpp
@@ -240,6 +240,8 @@ std::vector<Cuda> partition_space(const Cuda&, std::vector<T> const& weights) {
       std::is_arithmetic<T>::value,
       "Kokkos Error: partitioning arguments must be integers or floats");
 
+  // We only care about the number of instances to create and ignore weights
+  // otherwise.
   std::vector<Cuda> instances(weights.size());
   Impl::create_Cuda_instances(instances);
   return instances;
--- a/core/src/Cuda/Kokkos_Cuda_Instance.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_Instance.hpp
@@ -235,7 +235,7 @@ std::vector<Cuda> partition_space(const Cuda&, Args...) {
 }
 
 template <class T>
-std::vector<Cuda> partition_space(const Cuda&, std::vector<T>& weights) {
+std::vector<Cuda> partition_space(const Cuda&, std::vector<T> const& weights) {
   static_assert(
       std::is_arithmetic<T>::value,
       "Kokkos Error: partitioning arguments must be integers or floats");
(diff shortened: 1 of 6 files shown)


PhilMiller (approved):

core/src/Cuda/Kokkos_Cuda_Instance.hpp
```diff
@@ -235,7 +235,7 @@ std::vector<Cuda> partition_space(const Cuda&, Args...) {
 }
 
 template <class T>
-std::vector<Cuda> partition_space(const Cuda&, std::vector<T>& weights) {
+std::vector<Cuda> partition_space(const Cuda&, std::vector<T> const& weights) {
```
fnrizzi, line 238: weights is not used anywhere here, just to get its size?

masterleinad: Yes, for the device backends, we ignore the content of `weights`, we only care about the number of instances to create.

fnrizzi: can we add that as a comment so that it is clear for someone who reads it ?

fnrizzi (approved):

core/unit_test/TestExecSpacePartitioning.hpp
```diff
@@ -131,9 +131,8 @@ TEST(TEST_CATEGORY, partitioning_by_args) {
 }
 
 TEST(TEST_CATEGORY, partitioning_by_vector) {
-  std::vector<int> weights{1, 1};
-  auto instances =
-      Kokkos::Experimental::partition_space(TEST_EXECSPACE(), weights);
+  auto instances = Kokkos::Experimental::partition_space(
+      TEST_EXECSPACE(), std::vector<int> /*weights*/ {1, 1});
```
dalg24, line 135: It does not make it obvious to the reader that you mean to call with a temporary.
Consider doing
```C++
static_assert(
  std::is_invocablle_v<
    decltype(Kokkos::Experimental::partition_space<int>),
    std::vector<int>&&>,
  "intended to be callable with temporary"
);
```

masterleinad: I would need overload resolution for `Kokkos::Experimental::partition_space<ExecSpace, int>)` since both overloads are eligible. That means that I would need to simulate the call before testing that the function is callable with the arguments I need to provide for overload resolution which seems to defeat the purpose.

I just added a comment making the intent clear.

dalg24:
Please check on the original PR if taking a non-const reference was a design choice.

masterleinad:
> Please check on the original PR if taking a non-const reference was a design choice.

https://github.com/kokkos/kokkos/pull/4096#discussion_r673061817 asked that question but there was no answer.

PR #6172, opened 2023-05-30
Remove extended_namespace template parameter for SYCLMemoryOrder/Scope

masterleinad:
Recent compiler versions drop the `sycl::ext::oneapi` namespace in favor of just `sycl`, see https://github.com/intel/llvm/commit/17ea6d10f1d5c2796ab4b81911fef73b42658dba.
Corresponds to https://github.com/desul/desul/pull/106.


dalg24 (approved):
Please double check there is no diff with what we merged in desul.

masterleinad:
> Please double check there is no diff with what we merged in desul.

I double-checked that the changes are the same.

Rombur (approved):

PR #6176, opened 2023-06-01
Kokkos: Mark HWLOC as a TriBITS TPL as well (trilinos/Trilinos#11938)

bartlettroscoe:
It was reported by a Trilinos user in trilinos/Trilinos#11938 that after removing Kokkos subpackages that their build of Trilinos failed and by marking 'HWLOC' as a TriBITS TPL this fixed the problem.  HWLOC has always been a TriBITS TPL so it is not quite clear how this worked before this.

I tested this as part of Trilinos locally and it seems to work.


cmake/kokkos_tpls.cmake
```diff
@@ -31,7 +31,7 @@ FUNCTION(KOKKOS_TPL_OPTION PKG DEFAULT)
   ENDIF()
 ENDFUNCTION()
 
-KOKKOS_TPL_OPTION(HWLOC   Off)
+KOKKOS_TPL_OPTION(HWLOC   Off TRIBITS HWLOC)
```
dalg24, line 34: The trailing `HWLOC` does not do anything.
https://github.com/kokkos/kokkos/blob/ab6f7565b3a16403e214fcdc61f6245dea92eacf/cmake/kokkos_tpls.cmake#L5-L32
```suggestion
KOKKOS_TPL_OPTION(HWLOC   Off TRIBITS)
```

bartlettroscoe: >The trailing `HWLOC` does not do anything.

@dalg24, what makes you think that?

From the code:

https://github.com/kokkos/kokkos/blob/9df5a01a80509c63b14c1c68e1b1aec989159797/cmake/kokkos_tpls.cmake#L6-L10

and comparing to the documentation for [`cmake_parse_arguments()` ](https://cmake.org/cmake/help/v3.23/command/cmake_parse_arguments.html?highlight=cmake_parse_arguments) this is the form:

```
cmake_parse_arguments(<prefix> <options> <one_value_keywords>
                      <multi_value_keywords> <args>...)
```

So adding comments to the argument in `cmake_parse_arguments()` (which I always do) above gives:

```
  CMAKE_PARSE_ARGUMENTS(PARSED
    ""  # options
    "TRIBITS"  # one_value_keywords
    "" # multi_value_keywords
    ${ARGN})
```

So when you call:

```
  KOKKOS_TPL_OPTION(HWLOC   Off TRIBITS HWLOC)
```

the value of `PARSED_TRIBITRS` is `HWLOC` so the evaluation of `TPL_ENABLE_${PARSED_TRIBITS}` in:

https://github.com/kokkos/kokkos/blob/9df5a01a80509c63b14c1c68e1b1aec989159797/cmake/kokkos_tpls.cmake#L14-L16

is `TPL_ENABLE_HWLOC`.  To prove this, I made the following changes:

```
diff --git a/packages/kokkos/cmake/kokkos_tpls.cmake b/packages/kokkos/cmake/kokkos_tpls.cmake
index 3b02413b991..55060c12f54 100644
--- a/packages/kokkos/cmake/kokkos_tpls.cmake
+++ b/packages/kokkos/cmake/kokkos_tpls.cmake
@@ -3,6 +3,7 @@ KOKKOS_CFG_DEPENDS(TPLS DEVICES)
 KOKKOS_CFG_DEPENDS(TPLS COMPILER_ID)
 
 FUNCTION(KOKKOS_TPL_OPTION PKG DEFAULT)
+  message("\nKOKKOS_TPL_OPTION(${PKG} ${DEFAULT} ${ARGN})")
   CMAKE_PARSE_ARGUMENTS(PARSED
     ""
     "TRIBITS"
@@ -10,6 +11,8 @@ FUNCTION(KOKKOS_TPL_OPTION PKG DEFAULT)
     ${ARGN})
 
   IF (PARSED_TRIBITS)
+    print_var(PARSED_TRIBITS)
+    print_var(TPL_ENABLE_${PARSED_TRIBITS})
     #this is also a TPL option you can activate with Tribits
     IF (NOT "${TPL_ENABLE_${PARSED_TRIBITS}}" STREQUAL "")
       #Tribits brought its own default that should take precedence
@@ -17,11 +20,15 @@ FUNCTION(KOKKOS_TPL_OPTION PKG DEFAULT)
     ENDIF()
   ENDIF()
 
+  print_var(DEFAULT)
+
   KOKKOS_ENABLE_OPTION(${PKG} ${DEFAULT} "Whether to enable the ${PKG} library")
   KOKKOS_OPTION(${PKG}_DIR "" PATH "Location of ${PKG} library")
   SET(KOKKOS_ENABLE_${PKG} ${KOKKOS_ENABLE_${PKG}} PARENT_SCOPE)
   SET(KOKKOS_${PKG}_DIR  ${KOKKOS_${PKG}_DIR} PARENT_SCOPE)
 
+  print_var(KOKKOS_ENABLE_${PKG})
+
   IF (KOKKOS_HAS_TRILINOS
     AND KOKKOS_ENABLE_${PKG}
     AND NOT PARSED_TRIBITS)
```

and configured Kokkos under Trilinos with `-D TPL_ENABLE_HWLOC=true` which printed:

```
KOKKOS_TPL_OPTION(HWLOC Off TRIBITS;HWLOC)
-- PARSED_TRIBITS='HWLOC'
-- TPL_ENABLE_HWLOC='dummy_true'
-- DEFAULT='dummy_true'
-- KOKKOS_ENABLE_HWLOC='ON'
```

So this proves that the argument `HWLOC` after `TRIBITS` is required.

NOTE: The value of `KOKKOS_ENABLE_HWLOC` is set to `ON` from the non-cache var `Kokkos_ENABLE_HWLOC` which is shown in the Trilinos configure output:

```
-- Setting Kokkos_ENABLE_HWLOC=ON since Trilinos_ENABLE_Kokkos=ON AND TPL_ENABLE_HWLOC=ON
```

(The latter value `"TPL_ENABLE_HWLOC=ON"`  is not showing the actual value of  `TPL_ENABLE_HWLOC` of `dummy_true`, just that it evaluates to `ON` or true in an if-statement.  I should likely fix that.)

The actual cache var value is:

```
$ grep TPL_ENABLE_HWLOC: CMakeCache.txt 
TPL_ENABLE_HWLOC:STRING=dummy_true
```

bartlettroscoe: In my experience, it is a good idea to comment the arguments to `cmake_parse_arguments()` because the code is unreadable otherwise. (I understand why Kitware had to do it this way because of the limits of the CMake language but it does not make for very self-documenting code.)

PR #6187, opened 2023-06-06
Default hip/cuda memset async in zero memset

tcclevenger:
For ZeroMemset, if no execution space is given, use stream from default cuda/hip instance in a `{cuda|hip}MemsetAsync` call instead of `{cuda|hip}Memset`.

Update to https://github.com/kokkos/kokkos/pull/6087, from comment https://github.com/kokkos/kokkos/pull/6087#discussion_r1186314333.


dalg24:
Is this still a work in progress or is it ready for review?

dalg24:
This seems to be the only use and it looks like we should pass a default constructed exec space just like we did in the else-clause with the `fill_contiguous`

tcclevenger:
Superseded by https://github.com/kokkos/kokkos/pull/6764#pullrequestreview-1854097714. Closing.

PR #6183, opened 2023-06-05
SIMD: Replace unavailable loadu/storeu instructions

ldh4:
Resolves #6182.

`_mm512_loadu_epi64` and `_mm512_storeu_epi64` became available in `gcc/avx512fintrin.h` starting GCC 10. Compiling with previous versions of gcc causes undeclared errors on these.

This PR replaces `_mm512_loadu_epi64` with `_mm512_loadu_si512` and `_mm512_storeu_epi64` with `_mm512_storeu_si512`. 
Functionally, both replaced and replacing calls operate on 512 bits of memory that are not expected to be aligned in any boundary.

Issue #5677, opened 2022-12-12
Decouple SIMD variant target from micro-architecture optimization target


PhilMiller:
We may want to consider support for targeting the AVX2 SIMD variant even when we're optimizing for micro-architectures that support AVX-512. Depending on intensity of usage, AVX-512 can end up being detrimental to performance because of clock speed reductions while it's active, and the transition times as the vector units spin up/down. So, not necessary for this PR, but we shouldn't strictly tie it to what architecture we say we're targeting.

_Originally posted by @PhilMiller in https://github.com/kokkos/kokkos/issues/5512#issuecomment-1347210299_

PhilMiller:
https://travisdowns.github.io/blog/2020/01/17/avxfreq1.html for a deep dive into the motivations. The author has other blog posts related to this. The rest of the blog is very informative on low-level operating details of Intel CPUs, too.

ajpowelsnl:
@crtrott , @dalg24 - can this issue be closed?

masterleinad:
Related to https://github.com/kokkos/kokkos/pull/6188 and https://github.com/kokkos/kokkos/pull/6248.

PR #6188, opened 2023-06-06
Allow detecting SIMD types based on compiler macros 

masterleinad:
We have been noticing problems after merging SIMD pull requests that show that our CI support is pretty slim.
This pull request proposes allows `Kokkos_ARCH_NATIVE` ~~add a CMake option that allows~~ enabling SIMD support macro-based.


masterleinad:
Retest this please.

simd/src/Kokkos_SIMD_AVX512.hpp
```diff
@@ -854,11 +854,11 @@ class const_where_expression<simd_mask<double, simd_abi::avx512_fixed_size<8>>,
...
 
-  friend constexpr auto const& Impl::mask<double, abi_type>(
-      const_where_expression<mask_type, value_type> const& x);
+  KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
+  value_type get_value() const { return m_value; }
 
-  friend constexpr auto const& Impl::value<double, abi_type>(
-      const_where_expression<mask_type, value_type> const& x);
+  KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
+  mask_type get_mask() const { return m_mask; }
```
ldh4, line 861: I'm all for reverting these back into member functions. My attempts to make these friend functions to work on host and device have not been successful without vast amount of code duplication.

But based on the [previous comments on these member functions]( https://github.com/kokkos/kokkos/pull/5913#discussion_r1139176096), I think renaming these to be `impl_get_value` and `impl_get_mask` may be preferred.

ldh4 (approved):

simd/src/Kokkos_SIMD_AVX2.hpp
```diff
@@ -957,12 +957,6 @@ class const_where_expression<simd_mask<double, simd_abi::avx2_fixed_size<4>>,
       if (m_mask[lane]) mem[index[lane]] = m_value[lane];
     }
   }
-
-  friend constexpr auto const& Impl::mask<double, abi_type>(
-      const_where_expression<mask_type, value_type> const& x);
-
-  friend constexpr auto const& Impl::value<double, abi_type>(
-      const_where_expression<mask_type, value_type> const& x);
```
ldh4, line 965: Don't we need replacement `impl_get_value()` and `impl_get_mask()` functions here as well? Same for NEON simd.

masterleinad: Possibly, it just doesn't seem to be needed in the tests.

masterleinad: Looks like `reduce`, `hmin`, and `hmax` are not tested.

masterleinad: I verified manually that we need `impl_get_value/mask` for `reduce`, `hmin`, and `hmax` and added those.
We should test those but I haven't done that here.

simd/src/Kokkos_SIMD.hpp
```diff
@@ -139,11 +147,15 @@ class abi_set {};
...
+     (defined(KOKKOS_ENABLE_IMPL_SIMD_NATIVE) && defined(__AVX512F__)))
 using host_abi_set  = abi_set<simd_abi::scalar, simd_abi::avx512_fixed_size<8>>;
 using data_type_set = data_types<std::int32_t, std::uint32_t, std::int64_t,
                                  std::uint64_t, double>;
-#elif defined(KOKKOS_ARCH_AVX2)
+#elif defined(KOKKOS_ARCH_AVX2) || \
+    (defined(KOKKOS_ENABLE_IMPL_SIMD_NATIVE) && defined(__AVX2__))
 using host_abi_set = abi_set<simd_abi::scalar, simd_abi::avx2_fixed_size<4>>;
 using data_type_set =
     data_types<std::int32_t, std::int64_t, std::uint64_t, double>;
```
Rombur, line 161: Why  does `KOKKOS_ENABLE_IMPL_SIMD_NATIVE` not work for NEON> (line 162)?

masterleinad: We are already auto-detecting the architecture for `NEON`. We could discuss if we would hide that behind `KOKKOS_ENABLE_IMPL_SIMD_NATIVE` or not given that we don't have an explicit architecture option for `NEON`.

The focus at the moment is on all the changes not related to the new option and I will split the pull request/remove the changes adding testing before merging so that we can review the proposed option separately.

simd/unit_tests/TestSIMD.cpp
```diff
@@ -168,9 +168,11 @@ void host_check_binary_op_one_loader(BinaryOp binary_op, std::size_t n,
         loader.host_load(second_args + i, nlanes, second_arg);
     if (!(loaded_first_arg && loaded_second_arg)) continue;
     simd_type expected_result;
-    for (std::size_t lane = 0; lane < nlanes; ++lane) {
-      expected_result[lane] =
-          binary_op.on_host(T(first_arg[lane]), T(second_arg[lane]));
+    // gcc 8.4.0 warns if using nlanes as upper bound
```
dalg24, line 171: What was it warning about?

simd/src/Kokkos_SIMD_AVX512.hpp
```diff
@@ -519,11 +519,11 @@ class simd<std::uint64_t, simd_abi::avx512_fixed_size<8>> {
   }
   KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION void copy_from(value_type const* ptr,
                                                        element_aligned_tag) {
-    m_value = _mm512_loadu_epi64(ptr);
+    m_value = _mm512_loadu_si512(ptr);
   }
   KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION void copy_to(
       value_type* ptr, element_aligned_tag) const {
-    _mm512_storeu_epi64(ptr, m_value);
+    _mm512_storeu_si512(ptr, m_value);
```
dalg24, line 526: Why are these showing in the diff?  These changes were in #6183

simd/src/Kokkos_SIMD_Common.hpp
```diff
@@ -92,17 +96,34 @@ class where_expression<bool, T> : public const_where_expression<bool, T> {
...
+[[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION std::enable_if_t<
+    !std::is_same_v<Abi, Kokkos::Experimental::simd_abi::scalar>,
+    where_expression<simd_mask<T, Abi>, simd<T, Abi>>>
+where(typename simd<T, Abi>::mask_type const& mask, simd<T, Abi>& value) {
+  return where_expression(mask, value);
+}
+
+template <class T, class Abi>
+[[nodiscard]] KOKKOS_FORCEINLINE_FUNCTION std::enable_if_t<
+    std::is_same_v<Abi, Kokkos::Experimental::simd_abi::scalar>,
```
dalg24, line 108: Why SFINAE rather than a non-templated `simd_abi::scalar` overload?

dalg24 (dismissed):
Split things up:
Move the fix the host device annotation on `simd_mask` operations to their own PR.  This patch may be a candidate for cherry-picking into the 4.1 RC branch

Make sure you discuss the `Impl::{value,mask}` friend functions being replaced by `impl_*` member functions.  (Probably belongs to its own PR as well)

simd/src/Kokkos_SIMD_Common.hpp
masterleinad: Mostly, because it feels right. These functions are doing the same thing with identical implementation. Also, it looks clearer than replacing `Abi` everywhere.

I checked anyway that we can do a function overload instead.

simd/unit_tests/TestSIMD.cpp
masterleinad: I extended the comment.

simd/src/Kokkos_SIMD_AVX512.hpp
masterleinad: Should disappear after rebasing now.

ibaned:
It's not super clear what this PR does, but I also don't see anything wrong with it

masterleinad:
Depends on #6223.

ajpowelsnl:
@ldh4 , @ndellingwood  #6232 is relevant to this issue.

.jenkins
```diff
@@ -1,5 +1,5 @@
 pipeline {
-    agent none
+agent none
```
masterleinad, line 2:
```suggestion
    agent none
```

masterleinad:
Only `defaultdevicetype.shared_space` is failing for `CUDA-11.0.3-Clang-Tidy`. This looks spurious.

cmake/kokkos_arch.cmake
```diff
@@ -483,6 +471,39 @@ IF (KOKKOS_ARCH_POWER9)
...
+      SET(KOKKOS_ARCH_AVX512XEON ON)
+    ENDIF()
+
+    # Make sure to rerun the check if compile options have changed
+    UNSET(KOKKOS_COMPILER_HAS_AVX2)
+    CHECK_CXX_SYMBOL_EXISTS(__AVX2__ "" KOKKOS_COMPILER_HAS_AVX2)
+    IF(KOKKOS_COMPILER_HAS_AVX2)
+      SET(KOKKOS_ARCH_AVX2 ON)
+    ENDIF()
+
```
dalg24, line 495: What about `KOKKOS_ARCH_AVX`?

cmake/kokkos_arch.cmake
```diff
@@ -483,6 +471,39 @@ IF (KOKKOS_ARCH_POWER9)
...
+IF (KOKKOS_ARCH_NATIVE)
+  INCLUDE(CheckCXXSymbolExists)
+
+  IF(NOT "${KOKKOS_COMPILE_OPTIONS}" STREQUAL "${KOKKOS_COMPILE_OPTIONS_SAVED}")
+    SET(CMAKE_REQUIRED_QUIET ON)
+    SET(CMAKE_REQUIRED_FLAGS "${KOKKOS_COMPILE_OPTIONS}")
+    SET(KOKKOS_COMPILE_OPTIONS_SAVED "${KOKKOS_COMPILE_OPTIONS}" CACHE INTERNAL "")
+
+    # Make sure to rerun the check if compile options have changed
+    UNSET(KOKKOS_COMPILER_HAS_AVX512)
```
dalg24, line 483: Please confirm that this does something.  I did not find any mention of the variable argument being defined influencing the way the check is done in the [doc](https://cmake.org/cmake/help/latest/module/CheckCXXSymbolExists.html).

cmake/kokkos_arch.cmake
```diff
@@ -406,10 +406,7 @@ IF (KOKKOS_ARCH_SKL)
 ENDIF()
 
 IF (KOKKOS_ARCH_SKX)
-  # FIXME_NVHPC nvc++ doesn't seem to support AVX512.
-  IF (NOT KOKKOS_CXX_HOST_COMPILER_ID STREQUAL NVHPC)
-    SET(KOKKOS_ARCH_AVX512XEON ON)
-  ENDIF()
+  SET(KOKKOS_ARCH_AVX512XEON ON)
```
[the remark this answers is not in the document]
masterleinad: Done.

cmake/kokkos_arch.cmake
```diff
@@ -483,6 +471,39 @@ IF (KOKKOS_ARCH_POWER9)
   )
 ENDIF()
 
+IF (KOKKOS_ARCH_NATIVE)
+  INCLUDE(CheckCXXSymbolExists)
```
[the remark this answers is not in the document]
masterleinad: Done.

cmake/kokkos_arch.cmake
masterleinad: Yes, these are cached and the checks are not rerun otherwise.

cmake/kokkos_arch.cmake
```diff
@@ -483,6 +471,39 @@ IF (KOKKOS_ARCH_POWER9)
...
+
+  IF(NOT "${KOKKOS_COMPILE_OPTIONS}" STREQUAL "${KOKKOS_COMPILE_OPTIONS_SAVED}")
+    SET(CMAKE_REQUIRED_QUIET ON)
+    SET(CMAKE_REQUIRED_FLAGS "${KOKKOS_COMPILE_OPTIONS}")
+    SET(KOKKOS_COMPILE_OPTIONS_SAVED "${KOKKOS_COMPILE_OPTIONS}" CACHE INTERNAL "")
```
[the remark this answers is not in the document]
masterleinad: I moved it to the top and combined the two if-clauses.

dalg24 (commented):
It just occurred to me that multiple of these macros might be defined.
What if both `__AVX__` and `__AVX2__` are defined?
Currently that would define both macros `KOKKOS_ARCH_AVX` and `KOKKOS_ARCH_AVX2` which would never happen before.

masterleinad:
> Currently that would define both macros `KOKKOS_ARCH_AVX` and `KOKKOS_ARCH_AVX2` which would never happen before.

Right, but we only use these preprocessor definitions in `Kokkos_SIMD.hpp` and the CI shows that defining multiple of these doesn't cause issues. In fact, it allows users to select more SIMD types than before which should help some with https://github.com/kokkos/kokkos/issues/5677.
I don't think defining multiple of these would cause issues for users. We can of course take a step back and only define one preprocessor definition for more uniformity.

cmake/kokkos_arch.cmake
```diff
@@ -480,24 +480,19 @@ IF (KOKKOS_ARCH_NATIVE AND NOT "${KOKKOS_COMPILE_OPTIONS}" STREQUAL "${KOKKOS_CO
...
-  IF(KOKKOS_COMPILER_HAS_AVX2)
-    SET(KOKKOS_ARCH_AVX2 ON)
-  ENDIF()
-
-  # Make sure to rerun the check if compile options have changed
   UNSET(KOKKOS_COMPILER_HAS_AVX CACHE)
   CHECK_CXX_SYMBOL_EXISTS(__AVX__ "" KOKKOS_COMPILER_HAS_AVX)
-  IF(KOKKOS_COMPILER_HAS_AVX)
+
+  IF(KOKKOS_COMPILER_HAS_AVX512)
```
ldh4, line 491: I suggest leaving a comment here that briefly notes on these macros needing to be defined mutually exclusively for now and that this will probably have to be looked at again in the future.

dalg24:
Retest this please.
Not sure about the ROCm 5.6 failure

dalg24:
```

[ 94%] Building CXX object core/unit_test/CMakeFiles/Kokkos_CoreUnitTest_HIP.dir/hip/TestHIP_SubView_c10.cpp.o
clang-16: error: clang frontend command failed with exit code 70 (use -v to see invocation)
AMD clang version 16.0.0 (https://github.com/RadeonOpenCompute/llvm-project roc-5.6.0 23243 be997b2f3651a41597d7a41441fff8ade4ac59ac)
Thread model: posix
InstalledDir: /opt/rocm-5.6.0/llvm/bin
clang-16: note: diagnostic msg: Error generating preprocessed source(s).
make[2]: *** [simd/unit_tests/CMakeFiles/Kokkos_UnitTest_SIMD.dir/build.make:79: simd/unit_tests/CMakeFiles/Kokkos_UnitTest_SIMD.dir/TestSIMD.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:5168: simd/unit_tests/CMakeFiles/Kokkos_UnitTest_SIMD.dir/all] Error 2
[ 95%] Building CXX object core/unit_test/CMakeFiles/Kokkos_CoreUnitTest_HIP.dir/hip/TestHIP_SubView_c11.cpp.o
```

masterleinad:
> Not sure about the ROCm 5.6 failure

Let's see if we can get the failure reproduced (and print the detected architecture in that case).

Rombur:
Retest this please

masterleinad:
Looks like the problem with `ROCm` 5.6.0 not being able to compile the SIMD tests with AVX2 persists. I disabled AVX2 support in that case now as a workaround. Unfortunately, I don't have access to that compiler version to easily debug this further.

masterleinad:
#6449 should be merged first.

nliber (approved):

PR #6189, opened 2023-06-06
Make in-order queues the default via macro

masterleinad:
This pull request makes in-order queues the default but retains the option for us to change this later easily.
If we are using in-order queues, we don't need any of the `depends_on` or `submit_barrier` calls and for benchmarks up to moderate sizes, I see `zeCommandListAppendBarrier` dominating without this patch.
Related to #6035.

This pull request is in some way reverting back to in-order queues after we lifted the requirement in https://github.com/kokkos/kokkos/pull/5065 (but also guarding a bunch of other synchronization points necessary for out-of-order queues). Thus, this is for performance, not correctness (which was the issue before).

### DIFF
--- a/core/src/SYCL/Kokkos_SYCL_Instance.cpp
+++ b/core/src/SYCL/Kokkos_SYCL_Instance.cpp
@@ -75,9 +75,7 @@ void SYCLInternal::initialize(const sycl::device& d) {
       Kokkos::Impl::throw_runtime_exception(
           "There was an asynchronous SYCL error!\n");
   };
-  // FIXME SYCL+Cuda Apparently, submit_barrier doesn't quite work as expected
-  // for oneAPI 2023.0.0 on NVIDIA GPUs.
-#if defined(KOKKOS_IMPL_ARCH_NVIDIA_GPU)
+#ifdef KOKKOS_IMPL_SYCL_USE_IN_ORDER_QUEUES
   initialize(
       sycl::queue{d, exception_handler, sycl::property::queue::in_order()});
 #else
@@ -282,7 +280,9 @@ sycl::device_ptr<void> SYCLInternal::scratch_flags(const std::size_t size) {
   }
   auto memset_event = m_queue->memset(m_scratchFlags, 0,
                                       m_scratchFlagsCount * sizeScratchGrain);
+#ifndef KOKKOS_IMPL_SYCL_USE_IN_ORDER_QUEUES
   m_queue->ext_oneapi_submit_barrier(std::vector{memset_event});
+#endif
 
   return m_scratchFlags;
 }
--- a/core/src/setup/Kokkos_Setup_SYCL.hpp
+++ b/core/src/setup/Kokkos_Setup_SYCL.hpp
@@ -17,6 +17,11 @@
 #ifndef KOKKOS_SETUP_SYCL_HPP_
 #define KOKKOS_SETUP_SYCL_HPP_
 
+// FIXME_SYCL Using in-order queues currently gives better performance on Intel
+// GPUs and we run into correctness issues with out-of-order queues on NVIDIA
+// GPUs.
+#define KOKKOS_IMPL_SYCL_USE_IN_ORDER_QUEUES
+
 // FIXME_SYCL the fallback assert is temporarily disabled by default in the
 // compiler so we need to force it
 #ifndef SYCL_ENABLE_FALLBACK_ASSERT
--- a/core/src/SYCL/Kokkos_SYCL.cpp
+++ b/core/src/SYCL/Kokkos_SYCL.cpp
@@ -92,6 +92,13 @@ void SYCL::print_configuration(std::ostream& os, bool verbose) const {
 #else
   os << "macro  KOKKOS_IMPL_SYCL_DEVICE_GLOBAL_SUPPORTED : undefined\n";
 #endif
+
+#ifdef KOKKOS_IMPL_SYCL_USE_IN_ORDER_QUEUES
+  os << "macro  KOKKOS_IMPL_SYCL_USE_IN_ORDER_QUEUES : defined\n";
+#else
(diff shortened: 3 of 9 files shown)


core/src/setup/Kokkos_Setup_SYCL.hpp
```diff
@@ -17,6 +17,10 @@
 #ifndef KOKKOS_SETUP_SYCL_HPP_
 #define KOKKOS_SETUP_SYCL_HPP_
 
+// This seems to be necessary for SYCL+Cuda and gives better performance
+// otherwise
+#define KOKKOS_IMPL_SYCL_USE_IN_ORDER_QUEUES
```
dalg24, line 22: Make sure you print whether we are using in order queues or not from `print_configuration`

core/src/SYCL/Kokkos_SYCL_Instance.cpp
```diff
@@ -75,9 +75,7 @@ void SYCLInternal::initialize(const sycl::device& d) {
       Kokkos::Impl::throw_runtime_exception(
           "There was an asynchronous SYCL error!\n");
   };
-  // FIXME SYCL+Cuda Apparently, submit_barrier doesn't quite work as expected
-  // for oneAPI 2023.0.0 on NVIDIA GPUs.
-#if defined(KOKKOS_IMPL_ARCH_NVIDIA_GPU)
+#ifdef KOKKOS_IMPL_SYCL_USE_IN_ORDER_QUEUES
   initialize(
```
dalg24, line 79: Why did you drop the maco guard and the comment?

dalg24 (commented):
Make it more obvious in the description that this is in some sort reverting back to using in-order queues.  Please add a reference to the previous work (PR link) that had lifted the requirement.

masterleinad:
> Make it more obvious in the description that this is in some sort reverting back to using in-order queues. Please add a reference to the previous work (PR link) that had lifted the requirement.

I add some text in the pull request description.

core/src/SYCL/Kokkos_SYCL_Instance.cpp
masterleinad: I moved it to `Kokkos_Setup_SYCL.hpp` since that is the central place where we decide to use in-order or out-of-order queues.

core/src/setup/Kokkos_Setup_SYCL.hpp
```diff
@@ -17,6 +17,10 @@
 #ifndef KOKKOS_SETUP_SYCL_HPP_
 #define KOKKOS_SETUP_SYCL_HPP_
 
+// This seems to be necessary for SYCL+Cuda and gives better performance
+// otherwise
```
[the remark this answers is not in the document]
masterleinad: It gives better performance on Intel GPUs and we observed before that there were some correctness issues on NVIDIA GPUs.

core/src/decl/Kokkos_Declare_SYCL.hpp
```diff
@@ -18,6 +18,7 @@
 #define KOKKOS_DECLARE_SYCL_HPP
 
 #if defined(KOKKOS_ENABLE_SYCL)
+#include <setup/Kokkos_Setup_SYCL.hpp>
```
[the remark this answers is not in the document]
masterleinad: Yeah, I forgot to remove once verifying that everything works as expected. Fixed now.

core/src/setup/Kokkos_Setup_SYCL.hpp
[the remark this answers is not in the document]
dalg24: Try "Using in-order queues currently gives better performance on Intel GPUs and we run into correctness issues with out-of-order queues on NVIDIA GPUs"

core/src/setup/Kokkos_Setup_SYCL.hpp
```diff
@@ -17,6 +17,12 @@
 #ifndef KOKKOS_SETUP_SYCL_HPP_
 #define KOKKOS_SETUP_SYCL_HPP_
 
+// FIXME SYCL+Cuda Apparently, submit_barrier doesn't quite work as expected
```
dalg24, line 20: FIXME is fine but me but just want to make sure you did not want FIXME_SYCL

core/src/setup/Kokkos_Setup_SYCL.hpp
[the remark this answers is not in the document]
masterleinad: da972211c1152f69cb2b38d4654db46888417590

core/src/setup/Kokkos_Setup_SYCL.hpp
masterleinad: Changed it to FIXME_SYCL, see da972211c1152f69cb2b38d4654db46888417590.

dalg24 (approved):

dalg24:
Will ignore HIP timeouts.  Everything else passed on Jenkins.

masterleinad:
I just collapsed all the commits into one.

fnrizzi (approved):

jedbrown:
Hi :wave:. I see 4.1 (https://github.com/kokkos/kokkos/releases/tag/4.1.00) was released without this PR. This continues to be a substantial performance regression for vector operations in PETSc. Was that intentional and is there a time when in-order queues will go in a release?

PhilMiller:
This was merged to `develop` after the branch to stabilize the 4.1 release was prepared. It will definitely be in the 4.2 release (barring problems that lead to reversion), and could potentially be included in a hypothetical 4.1.1 release if the argument for it is strong enough.

jedbrown:
We'll be asking PETSc users to `--download-kokkos --download-kokkos-commit=origin/develop` (and for KK) if they care about performance, but we'd rather use a release and then users can have simpler version management (especially if they want to install Kokkos themselves instead of via `--download-kokkos`).

crtrott:
We could make this part of a 4.1.1

PR #6220, opened 2023-06-14
SIMD: add remaining gather_from and scatter_to

ldh4:
This PR adds missing `gather_from` and `scatter_to` functions:

- `void where_expression::gather_from(const double* mem, simd<std::int32_t, Abi> const& index)`
- `void const_where_expression::scatter_to(double* mem, simd<std::int32_t, Abi> const& index) const`

described in Kokkos documentation on SIMD [`when_expression`](https://kokkos.github.io/kokkos-core-wiki/API/simd/where_expression.html#).

Added for:

AVX2: `int32_t`, `int64_t` and `uint64_t`

AVX512: `int32_t`, `uint32_t`, `int64_t` and `uint64_t`

NEON: `int32_t`, `int64_t` and `uint64_t`

### DIFF
--- a/simd/src/Kokkos_SIMD_AVX512.hpp
+++ b/simd/src/Kokkos_SIMD_AVX512.hpp
@@ -1042,7 +1042,7 @@ class where_expression<simd_mask<double, simd_abi::avx512_fixed_size<8>>,
       double const* mem,
       simd<std::int32_t, simd_abi::avx512_fixed_size<8>> const& index) {
     m_value = value_type(_mm512_mask_i32gather_pd(
-        _mm512_set1_pd(0.0), static_cast<__mmask8>(m_mask),
+        static_cast<__m512d>(m_value), static_cast<__mmask8>(m_mask),
         static_cast<__m256i>(index), mem, 8));
   }
   template <class U, std::enable_if_t<
@@ -1081,6 +1081,14 @@ class const_where_expression<
     _mm256_mask_storeu_epi32(mem, static_cast<__mmask8>(m_mask),
                              static_cast<__m256i>(m_value));
   }
+  KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
+  void scatter_to(
+      std::int32_t* mem,
+      simd<std::int32_t, simd_abi::avx512_fixed_size<8>> const& index) const {
+    _mm256_mask_i32scatter_epi32(mem, static_cast<__mmask8>(m_mask),
+                                 static_cast<__m256i>(index),
+                                 static_cast<__m256i>(m_value), 4);
+  }
 
   [[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION value_type const&
   impl_get_value() const {
@@ -1108,6 +1116,14 @@ class where_expression<simd_mask<std::int32_t, simd_abi::avx512_fixed_size<8>>,
   void copy_from(std::int32_t const* mem, element_aligned_tag) {
     m_value = value_type(_mm256_mask_loadu_epi32(
         _mm256_set1_epi32(0), static_cast<__mmask8>(m_mask), mem));
+  }
+  KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
+  void gather_from(
+      std::int32_t const* mem,
+      simd<std::int32_t, simd_abi::avx512_fixed_size<8>> const& index) {
+    m_value = value_type(_mm256_mmask_i32gather_epi32(
+        static_cast<__m256i>(m_value), static_cast<__mmask8>(m_mask),
+        static_cast<__m256i>(index), mem, 4));
   }
   template <class U,
             std::enable_if_t<
@@ -1147,6 +1163,14 @@ class const_where_expression<
     _mm256_mask_storeu_epi32(mem, static_cast<__mmask8>(m_mask),
                              static_cast<__m256i>(m_value));
   }
+  KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
+  void scatter_to(
+      std::uint32_t* mem,
+      simd<std::int32_t, simd_abi::avx512_fixed_size<8>> const& index) const {
+    _mm256_mask_i32scatter_epi32(mem, static_cast<__mmask8>(m_mask),
+                                 static_cast<__m256i>(index),
+                                 static_cast<__m256i>(m_value), 4);
+  }
 
   [[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION value_type const&
   impl_get_value() const {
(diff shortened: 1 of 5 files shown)


ldh4:
@ibaned It looks like the current behavior of `gather_from` is different per backend. 

For NEON and scalar, if the mask value is false for an index, the `m_value` at that index is maintained.
https://github.com/kokkos/kokkos/blob/8d410bd245d34f7f0ea1750a69082257080d191a/simd/src/Kokkos_SIMD_NEON.hpp#L924-L930

However, for AVX2 and AVX512, `m_value` at the index is set to 0 if the mask value is false.
https://github.com/kokkos/kokkos/blob/8d410bd245d34f7f0ea1750a69082257080d191a/simd/src/Kokkos_SIMD_AVX512.hpp#L880-L887

The description on [where_expression](https://kokkos.github.io/kokkos-core-wiki/API/simd/where_expression.html#) in Kokkos documentation does not mention about zero'ing values. Based on that, should `gather_from` in AVX2 and AVX512 be changed to function consistently with NEON and scalar backends?

ibaned:
@ldh4 good point! I think keeping the previous value for lanes where the mask is `false` makes sense as the behavior for all these methods. It looks like that is how `where_expression::copy_from` is defined as well. Please feel free to change the behavior of all backends to "values stay the same if mask is `false`".

masterleinad (commented):
Apart from the question of how to handle `mask[i]==false` in `gather_from`, this looks good to me.

masterleinad (approved):

simd/src/Kokkos_SIMD_AVX512.hpp
```diff
@@ -921,6 +921,14 @@ class const_where_expression<
     _mm256_mask_storeu_epi32(mem, static_cast<__mmask8>(m_mask),
                              static_cast<__m256i>(m_value));
   }
+  KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
+  void scatter_to(
+      std::int32_t* mem,
+      simd<std::int32_t, simd_abi::avx512_fixed_size<8>> const& index) const {
+    _mm256_mask_i32scatter_epi32(mem, static_cast<__mmask8>(m_mask),
+                                 static_cast<__m256i>(index),
+                                 static_cast<__m256i>(m_value), 4);
```
dalg24, line 930: Should it be 4 -> 8 ?

simd/src/Kokkos_SIMD_NEON.hpp
```diff
@@ -965,6 +965,13 @@ class const_where_expression<
     if (m_mask[0]) mem[0] = m_value[0];
     if (m_mask[1]) mem[1] = m_value[1];
   }
+  KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
+  void scatter_to(
+      std::int32_t* mem,
+      simd<std::int32_t, simd_abi::neon_fixed_size<2>> const& index) const {
+    m_value[0] = (m_mask[0]) ? mem[0] : 0;
```
dalg24, line 972: Wait I thought `scatter_to` would write to `mem` not read from it.
What am I missing?

simd/src/Kokkos_SIMD_AVX512.hpp
dalg24: Actually why is the scale argument not `1` everywhere?

simd/unit_tests/TestSIMD.cpp
```diff
@@ -379,6 +379,76 @@ inline void host_check_condition() {
...
+  for (std::size_t i = 0; i < nlanes; i += 2) {
+    mask[i] = true;
+  }
+
+  DataType simd_val = 40;
+  simd_type test_simd(simd_val);
+  DataType val   = 3;
+  DataType dst[] = {val, val, val, val, val, val, val, val};
+
+  index_type index(0);
```
dalg24, line 400: I think same index and value for all lane is not a good test.  I suspect the scale argument is wrong when you call the intrinsics and your test does not expose the bug(?)

dalg24 (changes requested):
Need to clarify your use of the trailing scale argument with AVX2 and AVX512.
I think the NEON implementation is wrong.
And then, if the above is right, the test needs improvement.

simd/src/Kokkos_SIMD_AVX512.hpp
ldh4: Looking at the pseudo-code for `_mm256_mask_i32scatter_epi32` from [intel intrinsics](https://www.intel.com/content/www/us/en/docs/intrinsics-guide/index.html#text=_mm256_mask_i32scatter_epi32&ig_expand=3781):
```
addr := base_addr + SignExtend64(vindex[m+31:m]) * ZeroExtend64(scale) * 8
MEM[addr+31:addr] := a[i+31:i]
```
`scale` is really used to represent the byte size of type to determine the stride in bits.

In this case, because this simd type is for `int32_t`, we would need to use 4 for `scale` to make it 32 (bit) * index to get the offset.

crtrott (dismissed):

ajpowelsnl:
- **TODO**
- Re-request review from DLG for merging
- Close to being ready for merging

masterleinad (approved):
Looks OK to me.

ldh4:
> Need to clarify your use of the trailing scale argument with AVX2 and AVX512.
I think the NEON implementation is wrong.
And then, if the above is right, the test needs improvement.

@dalg24 Please re-review. The explanation on the trailing scale argument is provided in this comment https://github.com/kokkos/kokkos/pull/6220#discussion_r1236257029.
Previous incorrect NEON implementation on `scatter_to` has been fixed, and the unit test that previously failed to catch this has been modified as well.

simd/unit_tests/include/TestSIMD_WhereExpressions.hpp
```diff
@@ -0,0 +1,195 @@
...
+    mask_type mask(true);
+    mask[idx] = false;
+
+    simd_type dst;
+    index_type index;
+    simd_type expected_result;
+    for (std::size_t i = 0; i < nlanes; ++i) {
+      dst[i]             = (2 + (i * 2));
+      index[i]           = i;
+      expected_result[i] = (mask[i]) ? src[index[i]] : (2 + (i * 2));
```
dalg24, line 152:
```suggestion
      expected_result[i] = (mask[i]) ? src[index[i]] : dst[i];
```

simd/unit_tests/include/TestSIMD_WhereExpressions.hpp
```diff
@@ -0,0 +1,195 @@
...
+  for (std::size_t idx = 0; idx < nlanes; ++idx) {
+    mask_type mask(true);
+    mask[idx] = false;
+
+    simd_type dst;
+    index_type index;
+    simd_type expected_result;
+    for (std::size_t i = 0; i < nlanes; ++i) {
+      dst[i]             = (2 + (i * 2));
+      index[i]           = i;
```
dalg24, line 151: Is that too simple of a mapping?

simd/unit_tests/include/TestSIMD_WhereExpressions.hpp
dalg24: Same comment everywhere

dalg24 (commented):
Almost there

simd/unit_tests/include/TestSIMD_WhereExpressions.hpp
crtrott: I think its fine

simd/unit_tests/include/TestSIMD_WhereExpressions.hpp
ldh4: Changed in [09fd0c2](https://github.com/kokkos/kokkos/pull/6220/commits/09fd0c20743be077e061510fe7a53321675452e1).

dalg24 (approved):

PR #6221, opened 2023-06-15
Draft: split view header into multiple headers

nmm0:
This untangles the knot that is the current View header a bit. It creates two new headers -- there is Kokkos_ViewFwd.hpp that is as private header that forward declares and has the primary templates for some structures View needs. Then there is Kokkos_ViewTraits.hpp which has the ViewTraits struct and its specializations.

This may be a bit bold but I kind of want it to make doing the job of replacing the View implementation with mdspan easier and it honestly makes the code a little more understandable.

Please let me know what you think.

### DIFF
--- a/core/src/View/Kokkos_ViewFwd.hpp
+++ b/core/src/View/Kokkos_ViewFwd.hpp
@@ -1,0 +1,86 @@
+//@HEADER
+// ************************************************************************
+//
+//                        Kokkos v. 4.0
+//       Copyright (2022) National Technology & Engineering
+//               Solutions of Sandia, LLC (NTESS).
+//
+// Under the terms of Contract DE-NA0003525 with NTESS,
+// the U.S. Government retains certain rights in this software.
+//
+// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
+#ifndef KOKKOS_IMPL_PUBLIC_INCLUDE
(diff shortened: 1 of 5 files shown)


core/src/View/Kokkos_ViewFwd.hpp
```diff
@@ -0,0 +1,89 @@
...
+              "Constructor for Kokkos View '" + label +
+              "' has mismatched number of arguments. Number of arguments = " +
+              std::to_string(num_passed_args) +
+              " but dynamic rank = " + std::to_string(dyn_rank) + " \n";
+          Kokkos::abort(message.c_str());)
+      KOKKOS_IF_ON_DEVICE(Kokkos::abort("Constructor for Kokkos View has "
+                                        "mismatched number of arguments.");)
+    }
+  }
+}
```
masterleinad, line 81: I don't think this should be in a header file matching `*_fwd.hpp` but maybe a separate file `Kokkos_ViewUtilities.hpp` or `Kokkos_ViewHelper.hpp`. I'm not quite sure if there would other functions that can go there and would be used by other View-like classes.

masterleinad (commented):
I think breaking up `Kokkos_View.hpp` a little bit, in particular if it helps with integrating `MDSpan`.

core/src/View/Kokkos_ViewFwd.hpp
nmm0: Good point, honestly I don't mind having more, smaller headers that are easier to digest and it doesn't really make sense to put these in a "fwd" header.

dalg24:
Do you plan on revive this or should we close it?

nmm0:
Superseded by MDSpan changes

PR #6223, opened 2023-06-15
Fix SIMD support on GPUs

masterleinad:
This contains changes from #6188 without adding the option and modifying CI.
- NVHPC doesn't seem to support AVX512 SIMD types for the compiler versions tested on the CI, so we don't enable it in that case
- Constructors for the host-only SIMD types AVX512, AVX2, and NEON contained host-device constructors which resulted in warnings. These are host-only now.
- The friend declaration for accessing `m_mask` and `m_value` were missing host-device annotations which were difficult to get right since some of the SIMD types can be only used on the host and others also on the device but the interface must be consistent while avoiding the device compiler seeing vector instructions. Instead, this pull request introduces `impl_get_mask` and `impl_get_value` member functions that have the correct annotations for the respective SIMD types.
- GCC 8.4.0 complained about uninitialized values which look spurious. These can be silenced by an extra if-switch that doesn't change behavior. This change isn't really GPU-specific but still necessary for fixing the CI.
- `where_expressions` and `const_where_expression` also need correct host-device annotations to avoid compiler warnings about calling host functions from host-device functions.


ibaned:
Thank you for doing this! We hit this issue a few months back when trying to enable both Skylake and NVIDIA architectures in Kokkos. I'm glad you were able to resolve it.

ibaned (approved):
I only skimmed the code, but the description sounds like a very positive improvement

ldh4 (approved):

simd/src/Kokkos_SIMD.hpp
```diff
@@ -25,7 +25,8 @@
 #include <Kokkos_SIMD_AVX2.hpp>
 #endif
 
-#ifdef KOKKOS_ARCH_AVX512XEON
+// FIXME_NVHPC nvc++ doesn't seem to support AVX512.
+#if !defined(KOKKOS_COMPILER_NVHPC) && defined(KOKKOS_ARCH_AVX512XEON)
```
dalg24, line 29: Should we instead only define `KOKKOS_ARCH_AVX512XEON` when not using NVC++?

simd/src/Kokkos_SIMD_AVX2.hpp
```diff
@@ -958,11 +958,15 @@ class const_where_expression<simd_mask<double, simd_abi::avx2_fixed_size<4>>,
     }
   }
 
-  friend constexpr auto const& Impl::mask<double, abi_type>(
-      const_where_expression<mask_type, value_type> const& x);
+  [[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION value_type const&
```
dalg24, line 961: What happened to the constexpr?
Why did you add [[nodiscard]]?
(Same questions on all occurences)

simd/src/Kokkos_SIMD_Common.hpp
```diff
@@ -26,6 +26,10 @@ namespace Kokkos {
 
 namespace Experimental {
 
+namespace simd_abi {
+class scalar;
+}
```
dalg24, line 31: Why this instead of having more specialized overloads in `<Kokkos_SIMD_Scalar.hpp>`?

simd/src/Kokkos_SIMD_Common.hpp
```diff
@@ -308,44 +332,52 @@ KOKKOS_FORCEINLINE_FUNCTION where_expression<M, T>& operator/=(
...
+all_of(simd_mask<T, Abi> const& a) {
   return a == simd_mask<T, Abi>(true);
 }
 
 template <class T, class Abi>
-[[nodiscard]] KOKKOS_FORCEINLINE_FUNCTION bool any_of(
-    simd_mask<T, Abi> const& a) {
-  return a != simd_mask<T, Abi>(false);
+[[nodiscard]] KOKKOS_FORCEINLINE_FUNCTION std::enable_if_t<
+    std::is_same_v<Abi, Kokkos::Experimental::simd_abi::scalar>, bool>
```
dalg24, line 343: We are not consistent.  Here using SFINAE and in `where` above providing a more specialized overload.  (I prefer the more specialized overloads moved into the `<Kokkos_SIMD_Scalar.hpp>` header)

dalg24 (commented):
I do not necessarily have an issue with the ` impl_*` member functions instead of friend functions but I don't understand what prevents to use the latter (which would yield a smaller diff)

simd/src/Kokkos_SIMD.hpp
masterleinad: I don't feel strongly about it either way. [`c229f2e` (#6223)](https://github.com/kokkos/kokkos/pull/6223/commits/c229f2eeabcdf829e05b3fb476505e95ebe1fcea) changes it.

masterleinad:
`OPENMPTARGET-Clang` is failing in `openmptarget.partitioning_by_args` and `CUDA_11.7-NVHPC` fails with
```
NVC++-S-0000-Internal compiler error. extractor: bad opc       0  (/var/jenkins/workspace/Kokkos/algorithms/unit_tests/TestStdAlgorithmsMoveBackward.cpp: 216)
NVC++-S-0000-Internal compiler error. extractor: bad opc       0  (/var/jenkins/workspace/Kokkos/algorithms/unit_tests/TestStdAlgorithmsMoveBackward.cpp: 216)
NVC++-S-0000-Internal compiler error. extractor: bad opc       0  (/var/jenkins/workspace/Kokkos/algorithms/unit_tests/TestStdAlgorithmsMoveBackward.cpp: 216)
NVC++-S-0000-Internal compiler error. extractor: bad opc       0  (/var/jenkins/workspace/Kokkos/algorithms/unit_tests/TestStdAlgorithmsMoveBackward.cpp: 216)
NVC++-S-0000-Internal compiler error. extractor: bad opc       0  (/var/jenkins/workspace/Kokkos/algorithms/unit_tests/TestStdAlgorithmsMoveBackward.cpp: 216)
NVC++-S-0000-Internal compiler error. extractor: bad opc       0  (/var/jenkins/workspace/Kokkos/algorithms/unit_tests/TestStdAlgorithmsMoveBackward.cpp: 216)
NVC++-S-0000-Internal compiler error. extractor: bad opc       0  (/var/jenkins/workspace/Kokkos/algorithms/unit_tests/TestStdAlgorithmsMoveBackward.cpp: 216)
NVC++-S-0000-Internal compiler error. extractor: bad opc       0  (/var/jenkins/workspace/Kokkos/algorithms/unit_tests/TestStdAlgorithmsMoveBackward.cpp: 216)
NVC++/x86-64 Linux 22.9-0: compilation completed with severe errors
make[2]: *** [algorithms/unit_tests/CMakeFiles/Kokkos_AlgorithmsUnitTest_StdSet_C.dir/build.make:272: algorithms/unit_tests/CMakeFiles/Kokkos_AlgorithmsUnitTest_StdSet_C.dir/TestStdAlgorithmsMoveBackward.cpp.o] Error 2
```
both of these look unrelated.

simd/src/Kokkos_SIMD_AVX2.hpp
dalg24: I don't like the `[[nodiscard]]` so much but that is fine.

dalg24 (approved):
Retest this please

dalg24:
Retest this please

masterleinad:
Both `OPENMPTARGET-Clang` and `CUDA_11.7-NVHPC` are passing now.

PR #5444, opened 2022-09-08
HIP: Initialize device-related variables only by the singleton

masterleinad:
Analogous to #5442 for HIP.

PR #6224, opened 2023-06-16
fix uninitialized variables when using dynamic libraries with `RTLD_DEEPBIND`

### LINKED ISSUE #6178: `inline static` variables in headers break CUDA kernels of a library loaded with `RTLD_DEEPBIND`
**Describe the bug**

Some variables declared in headers as `inline static` prevent loading dynamic library with the `RTLD_DEEPBIND` flag.

So far I only encountered this bug with the CUDA backend, in my case one dynamic library called `Kokkos::initialize()` while another defined the kernels. Since the variables in [Kokkos::Impl::CudaInternal](https://github.com/kokkos/kokkos/blob/master/core/src/Cuda/Kokkos_Cuda_Instance.hpp#L104-L150) are `inline static`, they are defined in each library defining CUDA kernels as well as in `libkokkoscore.so`.
When loading a library with the `RTLD_DEEPBIND` flag its symbols are not overriden by those in `libkokkoscore.so`, therefore calling `Kokkos::initlialize()` would only initialize the variables in `libkokkoscore.so`, but not those exported by `inline static` and present in the other library.
Kernel launches from the loaded library would then throw an unrelated error.

From a quick `grep -r "inline static .* = .*;$"` in all sources, the HIP backend is also impacted.

1. compilers: clang-11 (host and CUDA)
2. Kokkos release: 4.0.1
3. platform and backend: linux, CUDA
4. cmake configure command:
```
cmake -S . -B ./build -DKokkos_ROOT=$KOKKOS_ROOT -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DKokkos_ENABLE_CUDA=ON -DKokkos_ENABLE_OPENMP=ON -DKokkos_ARCH_AMPERE80=ON
```
5. output from cmake command

<details>

```
-- The C compiler identification is Clang 11.0.1
-- The CXX compiler identification is Clang 11.0.1
-- Check for working C compiler: /usr/bin/clang - skipped
-- Check for working CXX compiler: /usr/bin/clang++ - skipped
-- Found OpenMP_C: -fopenmp=libomp (found version "5.0") 
-- Found OpenMP_CXX: -fopenmp=libomp (found version "5.0") 
-- Found OpenMP: TRUE (found version "5.0")  
-- Enabled Kokkos devices: OPENMP;CUDA
-- kokkos_launch_compiler is enabled globally. C++ compiler commands with -DKOKKOS_DEPENDENCE will be redirected to the appropriate compiler for Kokkos
-- Configuring done
-- Generating done
CMake Warning:

    Kokkos_ARCH_AMPERE80
```

</details>

<details><summary>`kernels.cpp`</summary>

```c++
#include "Kokkos_Core.hpp"

extern "C" double time_step(size_t N, double dx, const Kokkos::View<double*>& u, const Kokkos::View<double*>& c)
{
    double dt_global = INFINITY;

    Kokkos::parallel_reduce(Kokkos::RangePolicy<>(0, N),
    KOKKOS_LAMBDA(const Kokkos::RangePolicy<>::index_type i, double& dt) {
        double max_cx = Kokkos::max(Kokkos::abs(u[i] + c[i]), Kokkos::abs(u[i] - c[i]));
        dt = Kokkos::min(dt, dx / max_cx);
    }, Kokkos::Min<double>(dt_global));

    return dt_global;
}
```

</details>

<details><summary>`main.cpp`</summary>

```c++
#include "Kokkos_Core.hpp"
#include <dlfcn.h>

double time_step(size_t N, double dx, const Kokkos::View<double*>& u, const Kokkos::View<double*>& c);
using time_step_fptr_t = decltype(time_step)*;

int main()
{
    Kokkos::initialize();

    // Removing 'RTLD_DEEPBIND' will prevent the problem
    void* kernel_lib = dlopen("./build/libkernels.so", RTLD_LAZY | RTLD_LOCAL | RTLD_DEEPBIND);
    if (kernel_lib == nullptr) return 1;
    auto time_step_fptr = (time_step_fptr_t) dlsym(kernel_lib, "time_step");
    if (time_step_fptr == nullptr) return 2;

    {
        size_t N = 10;
        double dx = 0.1;

        Kokkos::View<double*> u("u", N), c("c", N);

        time_step_fptr(N, dx, u, c);
    }

    Kokkos::finalize();
}
```

</details>

<details><summary>`CMakeLists.txt`</summary>

```cmake
cmake_minimum_required(VERSION 3.20)
project(bug_repo_kokkos)

set(CMAKE_CXX_STANDARD 17)
set(CMAKE_CXX_EXTENSIONS OFF)
set(BUILD_SHARED_LIBS ON)

find_package(Kokkos REQUIRED)
    
add_library(kernels SHARED kernels.cpp)
target_link_libraries(kernels PUBLIC Kokkos::kokkos)
    
add_executable(main main.cpp)
target_link_libraries(main PUBLIC Kokkos::kokkos)
```

</details>

7. command line needed to reproduce the bug: `./build/main`
8. please also attach the `KokkosCore_config.h` header file (generated during the build);

<details><summary>`KokkosCore_config.h`</summary>

```c++

#if !defined(KOKKOS_MACROS_HPP) || defined(KOKKOS_CORE_CONFIG_H)
#error \
    "Do not include KokkosCore_config.h directly; include Kokkos_Macros.hpp instead."
#else
#define KOKKOS_CORE_CONFIG_H
#endif

// KOKKOS_VERSION % 100 is the patch level
// KOKKOS_VERSION / 100 % 100 is the minor version
// KOKKOS_VERSION / 10000 is the major version
#define KOKKOS_VERSION 40001
#define KOKKOS_VERSION_MAJOR 4
#define KOKKOS_VERSION_MINOR 0
#define KOKKOS_VERSION_PATCH 1

/* Execution Spaces */
/* #undef KOKKOS_ENABLE_SERIAL */
#define KOKKOS_ENABLE_OPENMP
/* #undef KOKKOS_ENABLE_OPENACC */
/* #undef KOKKOS_ENABLE_OPENMPTARGET */
/* #undef KOKKOS_ENABLE_THREADS */
#define KOKKOS_ENABLE_CUDA
/* #undef KOKKOS_ENABLE_HIP */
/* #undef KOKKOS_ENABLE_HPX */
/* #undef KOKKOS_ENABLE_MEMKIND */
/* #undef KOKKOS_ENABLE_LIBRT */
/* #undef KOKKOS_ENABLE_SYCL */

/* General Settings */
#define KOKKOS_ENABLE_CXX17
/* #undef KOKKOS_ENABLE_CXX20 */
/* #undef KOKKOS_ENABLE_CXX23 */

/* #undef KOKKOS_ENABLE_CUDA_RELOCATABLE_DEVICE_CODE */
/* #undef KOKKOS_ENABLE_CUDA_UVM */
#define KOKKOS_ENABLE_CUDA_LAMBDA
#define KOKKOS_ENABLE_CUDA_CONSTEXPR
/* #undef KOKKOS_ENABLE_IMPL_CUDA_MALLOC_ASYNC */
/* #undef KOKKOS_ENABLE_HIP_RELOCATABLE_DEVICE_CODE */
/* #undef KOKKOS_ENABLE_HIP_MULTIPLE_KERNEL_INSTANTIATIONS */
/* #undef KOKKOS_ENABLE_HPX_ASYNC_DISPATCH */
/* #undef KOKKOS_ENABLE_DEBUG */
/* #undef KOKKOS_ENABLE_DEBUG_DUALVIEW_MODIFY_CHECK */
/* #undef KOKKOS_ENABLE_DEBUG_BOUNDS_CHECK */
/* #undef KOKKOS_ENABLE_PROFILING_LOAD_PRINT */
/* #undef KOKKOS_ENABLE_TUNING */
/* #undef KOKKOS_ENABLE_DEPRECATED_CODE_3 */
#define KOKKOS_ENABLE_DEPRECATED_CODE_4
#define KOKKOS_ENABLE_DEPRECATION_WARNINGS
/* #undef KOKKOS_ENABLE_LARGE_MEM_TESTS */
#define KOKKOS_ENABLE_COMPLEX_ALIGN
#define KOKKOS_ENABLE_IMPL_DESUL_ATOMICS
/* #undef KOKKOS_OPT_RANGE_AGGRESSIVE_VECTORIZATION */
/* #undef KOKKOS_ENABLE_AGGRESSIVE_VECTORIZATION */
/* #undef KOKKOS_ENABLE_IMPL_MDSPAN */

/* TPL Settings */
/* #undef KOKKOS_ENABLE_HWLOC */
/* #undef KOKKOS_USE_LIBRT */
/* #undef KOKKOS_ENABLE_HBWSPACE */
#define KOKKOS_ENABLE_LIBDL
/* #undef KOKKOS_ENABLE_LIBQUADMATH */
#define KOKKOS_IMPL_CUDA_CLANG_WORKAROUND

/* #undef KOKKOS_ARCH_SSE42 */
/* #undef KOKKOS_ARCH_ARMV80 */
/* #undef KOKKOS_ARCH_ARMV8_THUNDERX */
/* #undef KOKKOS_ARCH_ARMV81 */
/* #undef KOKKOS_ARCH_ARMV8_THUNDERX2 */
/* #undef KOKKOS_ARCH_A64FX */
/* #undef KOKKOS_ARCH_AMD_AVX2 */
/* #undef KOKKOS_ARCH_AVX */
/* #undef KOKKOS_ARCH_AVX2 */
/* #undef KOKKOS_ARCH_AVX512XEON */
/* #undef KOKKOS_ARCH_KNC */
/* #undef KOKKOS_ARCH_AVX512MIC */
/* #undef KOKKOS_ARCH_POWER7 */
/* #undef KOKKOS_ARCH_POWER8 */
/* #undef KOKKOS_ARCH_POWER9 */
/* #undef KOKKOS_ARCH_INTEL_GEN */
/* #undef KOKKOS_ARCH_INTEL_DG1 */
/* #undef KOKKOS_ARCH_INTEL_GEN9 */
/* #undef KOKKOS_ARCH_INTEL_GEN11 */
/* #undef KOKKOS_ARCH_INTEL_GEN12LP */
/* #undef KOKKOS_ARCH_INTEL_XEHP */
/* #undef KOKKOS_ARCH_INTEL_PVC */
/* #undef KOKKOS_ARCH_INTEL_GPU */
/* #undef KOKKOS_ARCH_KEPLER */
/* #undef KOKKOS_ARCH_KEPLER30 */
/* #undef KOKKOS_ARCH_KEPLER32 */
/* #undef KOKKOS_ARCH_KEPLER35 */
/* #undef KOKKOS_ARCH_KEPLER37 */
/* #undef KOKKOS_ARCH_MAXWELL */
/* #undef KOKKOS_ARCH_MAXWELL50 */
/* #undef KOKKOS_ARCH_MAXWELL52 */
/* #undef KOKKOS_ARCH_MAXWELL53 */
/* #undef KOKKOS_ARCH_PASCAL */
/* #undef KOKKOS_ARCH_PASCAL60 */
/* #undef KOKKOS_ARCH_PASCAL61 */
/* #undef KOKKOS_ARCH_VOLTA */
/* #undef KOKKOS_ARCH_VOLTA70 */
/* #undef KOKKOS_ARCH_VOLTA72 */
/* #undef KOKKOS_ARCH_TURING75 */
#define KOKKOS_ARCH_AMPERE
#define KOKKOS_ARCH_AMPERE80
/* #undef KOKKOS_ARCH_AMPERE86 */
/* #undef KOKKOS_ARCH_ADA89 */
/* #undef KOKKOS_ARCH_HOPPER */
/* #undef KOKKOS_ARCH_HOPPER90 */
/* #undef KOKKOS_ARCH_AMD_ZEN */
/* #undef KOKKOS_ARCH_AMD_ZEN2 */
/* #undef KOKKOS_ARCH_AMD_ZEN3 */
/* #undef KOKKOS_ARCH_VEGA */
/* #undef KOKKOS_ARCH_VEGA906 */
/* #undef KOKKOS_ARCH_VEGA908 */
/* #undef KOKKOS_ARCH_VEGA90A */
/* #undef KOKKOS_ARCH_NAVI */
/* #undef KOKKOS_ARCH_NAVI1030 */
/* #undef KOKKOS_ARCH_NAVI1100 */
```

</details>

**Any additional info**

Launching a kernel would throw this error:
```
  what():  Kokkos::Impl::ParallelReduce< Cuda > requested too much L0 scratch memory
```

Which is caused by the fact that `Kokkos::Impl::m_maxShmemPerBlock` is uninitialized, even though we called `Kokkos::initialize()`.

I would have expected a `"ERROR device not initialized"` error message from [`Kokkos::Impl::CudaInternal::verify_is_initialized`](https://github.com/kokkos/kokkos/blob/7de2b62f79c48905a377590912d9c2a348e70aee/core/src/Cuda/Kokkos_Cuda_Instance.cpp#L320), but this is not the case, since it checks `m_cudaDev`, another `inline static` variable.

We can confirm that the symbol is defined in each library:
```
$ objdump -t ./build/libkernels.so | grep "maxShmem"
00000000002e0818  w    O .bss	0000000000000004              _ZN6Kokkos4Impl12CudaInternal18m_maxShmemPerBlockE

$ objdump -t kokkos_install_dir/libkokkoscore.so | grep "maxShmem"
00000000003a03c0  w    O .bss	0000000000000004              _ZN6Kokkos4Impl12CudaInternal18m_maxShmemPerBlockE
```
Upon loading `libkernels.so` the symbols should be mapped to those in `libkokkoscore.so`, but this is not the case when using `RTLD_DEEPBIND`.
Then when `Kokkos::initialize()` is called, it only initializes `Kokkos::Impl::m_maxShmemPerBlock` in `libkokkoscore.so`, hence the error mentioned above.

By moving the definition of `inline static` variables to `Kokkos_Cuda_Instance.cpp` their symbols are no longer defined in the library:
```
$ objdump -t ./build/libkernels.so | grep "maxShmem"
0000000000000000       O *UND*	0000000000000000              _ZN6Kokkos4Impl12CudaInternal18m_maxShmemPerBlockE
```
Then the kernel launches without any errors. I also tried this fix on a larger application and it did work successfully.

I should emphazise on the fact that loading `libkernels.so` without the `RTLD_DEEPBIND` flag will not cause any problems, therefore this bug is certainly not major but quite difficult to identify from the misleading error message.

In my case I got unlucky since this flag is set by default by Julia's `Libdl.dlopen`.

### LINKED ISSUE #6366: RuntimeError: Kokkos::Impl::ParallelReduce< Cuda > issues in Lightning-Kokkos with CUDA12
**Describe the bug**
[Lightning-Kokkos](https://github.com/PennyLaneAI/pennylane-lightning-kokkos) (a74957281293715d49c1bc925745365b710e5b78) can be built with CUDA12 and Kokkos 4.1.00, but reduce operations fail with
```
E       RuntimeError: Kokkos::Impl::ParallelReduce< Cuda > requested too much L0 scratch memory
```
Keeping everything else identical as possible, the code builds and executes correctly with CUDA11.
Printing a bit more, it appears that `maxShmemPerBlock` is not initialized properly
```
E       RuntimeError: Kokkos::Impl::ParallelReduce< Cuda > requested too much L0 scratch memory
E       minBlockSize = 32
E       reqShmemSize = 272
E       maxShmemPerBlock = 0
```

1. compilers (with versions):
On Perlmutter, this is my env
```
Currently Loaded Modules:
  1) gcc/11.2.0    (c)   3) cray-dsmml/0.2.2     5) craype-network-ofi         7) craype-accel-nvidia80   9) Nsight-Compute/2022.1.1  11) cudatoolkit/12.0                (g,E)  13) evp-patch
  2) craype/2.7.20 (c)   4) libfabric/1.15.2.0   6) PrgEnv-gnu/8.3.3   (cpe)   8) gpu/1.0                10) Nsight-Systems/2022.2.1  12) conda/Miniconda3-py311_23.5.2-0        14) python/3.11 (E,dev)
```
2. Kokkos release or commit used (i.e., the sha1 number): v4.1.00
3. platform and backend: OPENMP;CUDA;SERIAL
4. cmake configure command
```
rm -rf build
ARCH=AMPERE80
VERSION="4.1.00"

CC=cc CXX=CC cmake -B build \
-G Ninja \
-DCMAKE_BUILD_TYPE=RelWithDebInfo \
-DCMAKE_INSTALL_PREFIX=${HOME}/kokkos/${VERSION}_${ARCH}_CUDA12 \
-DBUILD_SHARED_LIBS:BOOL=ON \
-DBUILD_TESTING:BOOL=OFF \
-DCMAKE_CXX_STANDARD=20 \
-DKokkos_ENABLE_SERIAL:BOOL=ON \
-DKokkos_ENABLE_OPENMP:BOOL=ON \
-DKokkos_ENABLE_EXAMPLES:BOOL=OFF \
-DKokkos_ENABLE_TESTS:BOOL=OFF \
-DKokkos_ENABLE_LIBDL:BOOL=OFF \
-DKokkos_ENABLE_LAUNCH_COMPILER=OFF \
-DKokkos_ENABLE_DEPRECATED_CODE_3:BOOL=OFF \
-DKokkos_ENABLE_DEPRECATED_CODE_4:BOOL=OFF \
-DKokkos_ENABLE_CUDA:BOOL=ON \
-DKokkos_ARCH_${ARCH}=ON

cmake --build build -j32
cmake --install build
```
5. output from cmake command
[cmake-log.txt](https://github.com/kokkos/kokkos/files/12371309/cmake-log.txt)

[Lightning-Kokkos](https://github.com/PennyLaneAI/pennylane-lightning-kokkos)
I could not reproduce the bug with a simple program calling `Kokkos::parallel_reduce`. 
7. command line needed to reproduce the bug
```
python -m venv venv
source venv/bin/activate
pip install -r requirements.txt
pip install -e . -vv
pytest -s -x tests/test_expval.py
```
8. please also attach the `KokkosCore_config.h` header file (generated during the build);
[KokkosCore_config.txt](https://github.com/kokkos/kokkos/files/12371297/KokkosCore_config.txt)

**Any additional info**
Please provide any additional context about the issue here.

cz4rs:
fixes #6178 
fixes #6366 

This reverts #5442 and https://github.com/kokkos/kokkos/pull/5444 in order to fix issues with uninitialized static variables when using using dynamic libraries with `RTLD_DEEPBIND`.

The changes here affect Cuda and HIP backend.
In its current shape this might reintroduce the issue that #5442 was fixing - "the default stream was created before calling `cudaSetDevice`".

----
Added unit test (`Kokkos_RtldDeepbind`) reproduces the issue when building with `BUILD_SHARED_LIBS` enabled. Note that for Cuda it only reproduces when compiling with `clang`.

Details for Cuda are described in the original issue. HIP-ROCm error message:
```
15: terminate called after throwing an instance of 'std::runtime_error'
15:   what():  Kokkos::Impl::ParallelReduce< HIP > could not find a valid execution configuration.
15/42 Test #15: Kokkos_RtldDeepbind .........................Child aborted***Exception:   1.25 sec
```
[rocm-rtld-deepbind.log](https://github.com/kokkos/kokkos/files/11897795/rocm-rtld-deepbind.log)

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -23,7 +23,7 @@ jobs:
         cxx_extra_flags: ['']
         cmake_build_type: ['Release', 'Debug']
         backend: ['OPENMP']
-        clang-tidy: ['']
+        cmake_extra_args: ['']
         include:
           - distro: 'fedora:intel'
             cxx: 'icpc'
@@ -49,11 +49,16 @@ jobs:
             cxx: 'clang++'
             cmake_build_type: 'RelWithDebInfo'
             backend: 'THREADS'
-            clang-tidy: '-DCMAKE_CXX_CLANG_TIDY="clang-tidy;-warnings-as-errors=*"'
+            cmake_extra_args: '-DCMAKE_CXX_CLANG_TIDY="clang-tidy;-warnings-as-errors=*"'
           - distro: 'ubuntu:latest'
             cxx: 'g++'
             cmake_build_type: 'RelWithDebInfo'
             backend: 'THREADS'
+          - distro: 'ubuntu:latest'
+            cxx: 'clang++'
+            cmake_build_type: 'Debug'
+            backend: 'OPENMP'
+            cmake_extra_args: '-DBUILD_SHARED_LIBS=ON'
     runs-on: ubuntu-latest
     container:
       image: ghcr.io/kokkos/ci-containers/${{ matrix.distro }}
@@ -87,13 +92,13 @@ jobs:
         if: ${{ matrix.distro == 'ubuntu:latest' }}
         run: sudo apt-get update && sudo apt-get install -y libgtest-dev
       - name: maybe_install_clang_tidy
-        if: ${{ matrix.clang-tidy != '' }}
+        if: ${{ contains(matrix.cmake_extra_args, 'clang-tidy') }}
         run: sudo apt-get update && sudo apt-get install -y clang-tidy
       - name: Configure Kokkos
         run: |
           cmake -B builddir \
             -DCMAKE_INSTALL_PREFIX=/usr \
-            ${{ matrix.clang-tidy }} \
+            ${{ matrix.cmake_extra_args }} \
             -Ddesul_ROOT=/usr/desul-install/ \
             -DKokkos_ENABLE_DESUL_ATOMICS_EXTERNAL=ON \
             -DKokkos_ENABLE_HWLOC=ON \
--- a/.jenkins
+++ b/.jenkins
@@ -151,6 +151,7 @@ pipeline {
                         sh 'echo "/opt/rocm/llvm/lib" > /etc/ld.so.conf.d/llvm.conf && ldconfig'
                         sh '''rm -rf build && mkdir -p build && cd build && \
                               cmake \
+                                -DBUILD_SHARED_LIBS=ON \
                                 -DCMAKE_BUILD_TYPE=Debug \
                                 -DCMAKE_CXX_COMPILER=hipcc \
                                 -DCMAKE_CXX_FLAGS="-Werror -Wno-unused-command-line-argument -DNDEBUG" \
@@ -301,6 +302,7 @@ pipeline {
                         sh 'ccache --zero-stats'
                         sh '''rm -rf build && mkdir -p build && cd build && \
                               cmake \
+                                -DBUILD_SHARED_LIBS=ON \
                                 -DCMAKE_BUILD_TYPE=Release \
                                 -DCMAKE_CXX_CLANG_TIDY="clang-tidy;-warnings-as-errors=*" \
                                 -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
--- a/core/unit_test/CMakeLists.txt
+++ b/core/unit_test/CMakeLists.txt
@@ -1073,6 +1073,17 @@ KOKKOS_ADD_ADVANCED_TEST( CoreUnitTest_PushFinalizeHook_terminate
       tools/TestAllCalls.cpp
     )
 
+    if(BUILD_SHARED_LIBS)
+      add_library(RtldDeepbind_TestKernel SHARED TestKernel.cpp)
+      target_link_libraries(RtldDeepbind_TestKernel PUBLIC Kokkos::kokkos)
+
+      KOKKOS_ADD_EXECUTABLE_AND_TEST(
+        RtldDeepbind
+        SOURCES TestRtldDeepbind.cpp
+      )
+      add_dependencies(Kokkos_RtldDeepbind RtldDeepbind_TestKernel)
+    endif()
+
     KOKKOS_ADD_TEST_EXECUTABLE(
       ToolsInitialization
       UnitTestMain.cpp
--- a/core/unit_test/TestKernel.cpp
+++ b/core/unit_test/TestKernel.cpp
@@ -1,0 +1,33 @@
+//@HEADER
+// ************************************************************************
+//
+//                        Kokkos v. 4.0
+//       Copyright (2022) National Technology & Engineering
+//               Solutions of Sandia, LLC (NTESS).
+//
+// Under the terms of Contract DE-NA0003525 with NTESS,
+// the U.S. Government retains certain rights in this software.
+//
+// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
+#include <Kokkos_Core.hpp>
+
+extern "C" double time_step(size_t N, double dx, const Kokkos::View<double*>& u,
+                            const Kokkos::View<double*>& c) {
+  double dt_global = INFINITY;
+
+  Kokkos::parallel_reduce(
+      Kokkos::RangePolicy<>(0, N),
+      KOKKOS_LAMBDA(const Kokkos::RangePolicy<>::index_type i, double& dt) {
+        double max_cx =
+            Kokkos::max(Kokkos::abs(u[i] + c[i]), Kokkos::abs(u[i] - c[i]));
+        dt = Kokkos::min(dt, dx / max_cx);
+      },
+      Kokkos::Min<double>(dt_global));
+
+  return dt_global;
+}
--- a/core/unit_test/TestRtldDeepbind.cpp
+++ b/core/unit_test/TestRtldDeepbind.cpp
@@ -1,0 +1,50 @@
+//@HEADER
+// ************************************************************************
+//
+//                        Kokkos v. 4.0
+//       Copyright (2022) National Technology & Engineering
+//               Solutions of Sandia, LLC (NTESS).
+//
+// Under the terms of Contract DE-NA0003525 with NTESS,
+// the U.S. Government retains certain rights in this software.
+//
+// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
+#include <Kokkos_Core.hpp>
+#include <dlfcn.h>
+#include <iostream>
+
+double time_step(size_t N, double dx, const Kokkos::View<double*>& u,
+                 const Kokkos::View<double*>& c);
+using time_step_fptr_t = decltype(time_step)*;
+
+int main(int argc, char** argv) {
+  Kokkos::initialize(argc, argv);
+  Kokkos::print_configuration(std::cout);
+
+  void* kernel_lib = dlopen("./libRtldDeepbind_TestKernel.so",
+                            RTLD_LAZY | RTLD_LOCAL | RTLD_DEEPBIND);
+  if (kernel_lib == nullptr) {
+    std::cout << "Could not dlopen libRtldDeepbind_TestKernel.so.\n";
+    return 1;
+  }
+
+  auto time_step_fptr =
+      reinterpret_cast<time_step_fptr_t>(dlsym(kernel_lib, "time_step"));
+  if (time_step_fptr == nullptr) return 2;
+
+  {
+    size_t N  = 10;
+    double dx = 0.1;
+
+    Kokkos::View<double*> u("u", N), c("c", N);
+
+    time_step_fptr(N, dx, u, c);
+  }
+
+  Kokkos::finalize();
+}


core/src/Cuda/Kokkos_Cuda_Instance.cpp
```diff
@@ -359,9 +374,106 @@ void CudaInternal::initialize(cudaStream_t stream, bool manage_stream) {
   }
 #endif
 
+  const CudaInternalDevices &dev_info = CudaInternalDevices::singleton();
+
   const bool ok_init = nullptr == m_scratchSpace || nullptr == m_scratchFlags;
 
-  if (ok_init) {
+  const bool ok_id =
```
cz4rs, line 381: TODO: consider removing as in 8b5a155af024a1dd0bf68957e87973475d036dc9

cz4rs:
Work in progress: using unit tests for reproduction (b99aaa2) didn't work for me, so I have moved the reproducer to `examples`.

Following commands will show `terminate called (...) Kokkos::Impl::ParallelReduce< Cuda > requested too much L0 scratch memory` error when using unfixed version:
```
cmake -B build -DCMAKE_C_COMPILER=clang -DCMAKE_CXX_COMPILER=clang++ -DKokkos_ENABLE_CUDA=ON -DKokkos_ROOT=../../build/install
cmake --build build
./build/main
```

masterleinad:
> Work in progress: using unit tests for reproduction ([b99aaa2](https://github.com/kokkos/kokkos/commit/b99aaa2cfe083626e4fc028f5af75e033f91d604)) didn't work for me, so I have moved the reproducer to `examples`.

You were missing
```diff
diff --git a/core/unit_test/CMakeLists.txt b/core/unit_test/CMakeLists.txt
index 8b5beecdd..78cae63a1 100644
--- a/core/unit_test/CMakeLists.txt
+++ b/core/unit_test/CMakeLists.txt
@@ -957,6 +957,8 @@ KOKKOS_ADD_ADVANCED_TEST( CoreUnitTest_PushFinalizeHook_terminate
       SOURCES TestRtldDeepbind.cpp
     )
 
+    add_dependencies(Kokkos_RtldDeepbind kernels rtld)
+
     KOKKOS_ADD_TEST_EXECUTABLE(
       ToolsInitialization
       UnitTestMain.cpp
```
and making sure we get position-independent code (either building as a shared library or setting the `POSITION_INDEPENDENT_CODE` property for `kokkoscore`)

.jenkins
```diff
@@ -145,12 +145,14 @@ pipeline {
                         OMP_MAX_ACTIVE_LEVELS = 3
                         OMP_PLACES = 'threads'
                         OMP_PROC_BIND = 'spread'
+                        NVCC_WRAPPER_DEFAULT_COMPILER = 'clang++'
```
cz4rs, line 148: Note: this bug only occurs when using `clang`. `NVCC_WRAPPER_DEFAULT_COMPILER` defaults to `g++` which does NOT trigger the bug.

This SO post contains some hints about the difference in implementations: https://stackoverflow.com/a/70673863.

masterleinad: I'm confused. `HIP-ROCm-5.2` doesn't use `nvcc_wrapper` anyway so adding this line should not have any effect.

cz4rs: Ok, maybe I'm overdoing this. This was the case when using `nvcc_wrapper` and Cuda, let me double check HIP locally.

cz4rs: In the meantime, the build seems to fail as expected:
```
15: terminate called after throwing an instance of 'std::runtime_error'
15:   what():  Kokkos::Impl::ParallelReduce< HIP > could not find a valid execution configuration.
```

cz4rs: Line removed as it is not needed to reproduce the issue:
[rocm-rtld-deepbind.log](https://github.com/kokkos/kokkos/files/11897787/rocm-rtld-deepbind.log)

.github/workflows/continuous-integration-workflow.yml
```diff
@@ -49,11 +49,16 @@ jobs:
             cxx: 'clang++'
             cmake_build_type: 'RelWithDebInfo'
             backend: 'THREADS'
-            clang-tidy: '-DCMAKE_CXX_CLANG_TIDY="clang-tidy;-warnings-as-errors=*"'
+            cmake_extra_args: '-DCMAKE_CXX_CLANG_TIDY="clang-tidy;-warnings-as-errors=*"'
```
crtrott, line 52: why this change?

core/src/Cuda/Kokkos_Cuda_Instance.cpp
```diff
@@ -344,12 +356,15 @@ void CudaInternal::fence() const {
...
 
-void CudaInternal::initialize(cudaStream_t stream, bool manage_stream) {
+void CudaInternal::initialize(int cuda_device_id, cudaStream_t stream,
+                              bool manage_stream) {
   if (was_finalized)
     Kokkos::abort("Calling Cuda::initialize after Cuda::finalize is illegal\n");
   was_initialized = true;
   if (is_initialized()) return;
 
+  enum { WordSize = sizeof(size_type) };
```
crtrott, line 366: why not constexpr variable?

.github/workflows/continuous-integration-workflow.yml
cz4rs: To avoid having separate variables for `clang-tidy` and '-DBUILD_SHARED_LIBS=ON'.

vincentmr:
Hi all, this can also be marked as fixing #6366 .

cz4rs:
@stanmoore1 
This currently involves reverting changes from #5442 and can thus reintroduce the issue that you have reported (`that the default stream was created before calling cudaSetDevice`).
Can you advise on how to reproduce that issue it so I can add a unit test for that case?

masterleinad:
It should be enough to set the device before creating the singleton stream which we should do now anyway.

cz4rs:
Note: this PR currently contains only the unit test. The "fix" was to revert #5442 (and this was enough to make the test pass), but there were so many changes in the related code in the meantime that I had to clean up this branch. Unfortunately I still haven't gotten to reapply the fix properly.

masterleinad:
I would be very interested to see if the test added here just works out-of-the-box now.

masterleinad:
Retest this please.

masterleinad:
Locally this seems to work fine without extra changes on `Cuda` but we still have similar complaints when using `MDRangePolicy`, see https://kokkosteam.slack.com/archives/C5BGU5NDQ/p1706727845918379. It might be necessary to do
```C++
diff --git a/core/src/Cuda/Kokkos_Cuda_Instance.cpp b/core/src/Cuda/Kokkos_Cuda_Instance.cpp
index 9a57eb939..fa3f401c2 100644
--- a/core/src/Cuda/Kokkos_Cuda_Instance.cpp
+++ b/core/src/Cuda/Kokkos_Cuda_Instance.cpp
@@ -200,6 +200,12 @@ int Impl::CudaInternal::concurrency() {
   return concurrency;
 }
 
+int CudaInternal::m_cudaArch;
+cudaDeviceProp CudaInternal::m_deviceProp;
+unsigned long *CudaInternal::constantMemHostStaging;
+cudaEvent_t CudaInternal::constantMemReusable;
+std::mutex CudaInternal::constantMemMutex;
+
 void CudaInternal::print_configuration(std::ostream &s) const {
   s << "macro  KOKKOS_ENABLE_CUDA      : defined\n";
diff --git a/core/src/Cuda/Kokkos_Cuda_Instance.hpp b/core/src/Cuda/Kokkos_Cuda_Instance.hpp
index db764fb1b..418f776cf 100644
--- a/core/src/Cuda/Kokkos_Cuda_Instance.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_Instance.hpp
@@ -87,10 +87,10 @@ class CudaInternal {
   int m_cudaDev = -1;
 
   // Device Properties
-  inline static int m_cudaArch = -1;
+  static int m_cudaArch;
   static int concurrency();
 
-  inline static cudaDeviceProp m_deviceProp;
+  static cudaDeviceProp m_deviceProp;
 
   // Scratch Spaces for Reductions
   mutable std::size_t m_scratchSpaceCount;
@@ -118,9 +118,9 @@ class CudaInternal {
 
   // FIXME_CUDA: these want to be per-device, not per-stream...  use of 'static'
   //  here will break once there are multiple devices though
-  inline static unsigned long* constantMemHostStaging = nullptr;
-  inline static cudaEvent_t constantMemReusable       = nullptr;
-  inline static std::mutex constantMemMutex;
+  static unsigned long* constantMemHostStaging;
+  static cudaEvent_t constantMemReusable;
+  static std::mutex constantMemMutex;
 
   static CudaInternal& singleton();
```
in that case.

masterleinad:
Superseded by https://github.com/kokkos/kokkos/pull/7065.

PR #6229, opened 2023-06-20
[HIP] Optimize parallel_reduce

arghdos:
Based on #6160, I spent a bit of time to understand the original regression introduced by #6029.

Essentially, the issue was most exacerbated when each thread was doing a reduction over a single element of the array (i.e., each thread loads two values, multiplies them, sums, and passes to the rest of the reduction).
This was mostly due to the cost of `hip_single_inter_block_reduce_scan`, namely due to runtime int divisions of (e.g.,) `blockDim.x` in `hip_inter_warp_shuffle_reduction` and friends.
After spending some time playing with various reduction algorithms, I discovered that the SHMEM reduction was faster in pretty much all cases, so I flipped the default from `UseShflReduction=true` to `false`.
In addition, I did a bit of cleanup to use `__syncthreads_or` (implemented since ROCm 4.5), which bumps perf a little more, and finally, I tweaked @Rombur's heuristic to have three ranges:

- "small" -> not a lot of parallelism, don't touch it.
- "intermediate" -> this is where we saw the largest regressions, there's not enough parallelism to hide some of the cost in reductions.  Here, we target trying to have four reduction items per-thread (~ what 4.0.01 did)
- "large" -> lots of parallelism (RHODO is here), so we leave it alone

I've plotted the performance (runtime, in seconds) for DOT products over various commits:

image

- 4.0.01 -> the latest stable release
- dev -> the last commit before #6160 (43a797b59b17f9939178ab1f5e2df8c72099a33e)
- rombur (light) -> d78a24f72a4aeb439b19843b0c8b158f96c3a370, with the lightweighthint applied
- new reduce -> shmem reduction
- new reduce + new heur -> shmem reduction + the new heuristic

Generally we see the large perf regressions:

image

are solved, and the "new reduce + new heur" line is fasted _except_ one point where it's a bit slower:

image

this (as you might guess from the above) happens right at the "4096 blocks" crossover point.  Perhaps we could tweak it a bit more, but generally this is a large improvement.

Finally, I compared LAMMPS perf over a # of benchmarks, and this maintains perf (to within the typical noise of 1%) as compared to @Rombur's patch:

```
+---------+-----------------------+--------------------+
|         | lammps-dev-baseline   | lammps-newreduce   |
|---------+-----------------------+--------------------|
| eam     | 0.00%                 | 0.75%              |
| lj      | 0.00%                 | -0.16%             |
| reaxff  | 0.00%                 | -0.80%             |
| rhodo   | 0.00%                 | -0.20%             |
| snap    | 0.00%                 | 0.07%              |
| tersoff | 0.00%                 | 0.04%              |
+---------+-----------------------+--------------------+
```

### DIFF
--- a/core/src/HIP/Kokkos_HIP_Parallel_Range.hpp
+++ b/core/src/HIP/Kokkos_HIP_Parallel_Range.hpp
@@ -322,7 +322,7 @@ class ParallelReduce<CombinedFunctorReducerType, Kokkos::RangePolicy<Traits...>,
       //    2. Have each thread reduce > 1 value to minimize overheads
       //    3. Limit the total # of blocks, to avoid unbounded scratch space
       constexpr int block_max           = 4096;
-      constexpr int preferred_block_min = 256;
+      constexpr int preferred_block_min = 1024;
 
       if (nblocks < preferred_block_min) {
         // keep blocks as is, already have low parallelism
--- a/core/src/HIP/Kokkos_HIP_Parallel_Range.hpp
+++ b/core/src/HIP/Kokkos_HIP_Parallel_Range.hpp
@@ -147,7 +147,7 @@ class ParallelReduce<CombinedFunctorReducerType, Kokkos::RangePolicy<Traits...>,
   const bool m_result_ptr_device_accessible;
   const bool m_result_ptr_host_accessible;
   word_size_type* m_scratch_space = nullptr;
-  size_type* m_scratch_flags = nullptr;
+  size_type* m_scratch_flags      = nullptr;
 
   static bool constexpr UseShflReduction = false;
 
@@ -177,8 +177,9 @@ class ParallelReduce<CombinedFunctorReducerType, Kokkos::RangePolicy<Traits...>,
 
   __device__ inline void run(SHMEMReductionTag) const {
     const ReducerType& reducer = m_functor_reducer.get_reducer();
-    const integral_nonzero_constant<
-        word_size_type, ReducerType::static_value_size() / sizeof(word_size_type)>
+    const integral_nonzero_constant<word_size_type,
+                                    ReducerType::static_value_size() /
+                                        sizeof(word_size_type)>
         word_count(reducer.value_size() / sizeof(size_type));
 
     {
@@ -206,8 +207,8 @@ class ParallelReduce<CombinedFunctorReducerType, Kokkos::RangePolicy<Traits...>,
     if (!do_final_reduction)
       do_final_reduction = hip_single_inter_block_reduce_scan<false>(
           reducer, blockIdx.x, gridDim.x,
-          ::Kokkos::kokkos_impl_hip_shared_memory<word_size_type>(), m_scratch_space,
-          m_scratch_flags);
+          ::Kokkos::kokkos_impl_hip_shared_memory<word_size_type>(),
+          m_scratch_space, m_scratch_flags);
     if (do_final_reduction) {
       // This is the final block with the final result at the final threads'
       // location
@@ -215,9 +216,10 @@ class ParallelReduce<CombinedFunctorReducerType, Kokkos::RangePolicy<Traits...>,
       word_size_type* const shared =
           ::Kokkos::kokkos_impl_hip_shared_memory<word_size_type>() +
           (blockDim.y - 1) * word_count.value;
-      word_size_type* const global = m_result_ptr_device_accessible
-                                    ? reinterpret_cast<word_size_type*>(m_result_ptr)
-                                    : m_scratch_space;
+      word_size_type* const global =
+          m_result_ptr_device_accessible
+              ? reinterpret_cast<word_size_type*>(m_result_ptr)
+              : m_scratch_space;
 
       if (threadIdx.y == 0) {
         reducer.final(reinterpret_cast<value_type*>(shared));
@@ -315,10 +317,11 @@ class ParallelReduce<CombinedFunctorReducerType, Kokkos::RangePolicy<Traits...>,
       int nblocks = (nwork + block.y - 1) / block.y;
       // Heuristic deciding the value of nblocks.
       // The general idea here is we want to:
-      //    1. Not undersubscribe the device (i.e., we want at least preferred_block_min blocks)
+      //    1. Not undersubscribe the device (i.e., we want at least
+      //    preferred_block_min blocks)
       //    2. Have each thread reduce > 1 value to minimize overheads
       //    3. Limit the total # of blocks, to avoid unbounded scratch space
-      constexpr int block_max = 4096;
+      constexpr int block_max           = 4096;
       constexpr int preferred_block_min = 256;
 
       if (nblocks < preferred_block_min) {
@@ -327,18 +330,23 @@ class ParallelReduce<CombinedFunctorReducerType, Kokkos::RangePolicy<Traits...>,
         // "large dispatch" -> already have lots of parallelism
         nblocks = block_max;
       } else {
-        // in the intermediate range, try to have each thread process multiple items to
-        // offset the cost of the reduction (with not enough parallelism to hide it)
-        int items_per_thread = (nwork + nblocks * block_size - 1) / (nblocks * block_size);
+        // in the intermediate range, try to have each thread process multiple
+        // items to offset the cost of the reduction (with not enough
+        // parallelism to hide it)
+        int items_per_thread =
+            (nwork + nblocks * block_size - 1) / (nblocks * block_size);
         if (items_per_thread < 4) {
-          int ratio = std::min((nblocks + preferred_block_min - 1) / preferred_block_min, (4 + items_per_thread - 1) / items_per_thread);
+          int ratio = std::min(
+              (nblocks + preferred_block_min - 1) / preferred_block_min,
+              (4 + items_per_thread - 1) / items_per_thread);
           nblocks /= ratio;
         }
       }
 
       // TODO: down casting these uses more space than required?
-      m_scratch_space  = (word_size_type*)::Kokkos::Impl::hip_internal_scratch_space(
-          m_policy.space(), reducer.value_size() * nblocks);
+      m_scratch_space =
+          (word_size_type*)::Kokkos::Impl::hip_internal_scratch_space(
+              m_policy.space(), reducer.value_size() * nblocks);
       // Intentionally do not downcast to word_size_type since we use HIP
       // atomics in Kokkos_HIP_ReduceScan.hpp
       m_scratch_flags = ::Kokkos::Impl::hip_internal_scratch_flags(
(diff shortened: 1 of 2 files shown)


arghdos:
cc: @Rombur @dalg24

Rombur (dismissed):
This is great. Thanks.

masterleinad:
> After spending some time playing with various reduction algorithms, I discovered that the SHMEM reduction was faster in pretty much all cases, so I flipped the default from `UseShflReduction=true` to `false`.

Do you have any idea why that is? I see similar behavior in https://github.com/kokkos/kokkos/pull/6035 but can't make much sense of it. I mean the shuffle reduction could always be implemented in terms of local memory so I would expect better performance if shuffle reductions can be used.

core/src/HIP/Kokkos_HIP_ReduceScan.hpp
```diff
@@ -395,20 +387,11 @@ __device__ bool hip_single_inter_block_reduce_scan_impl(
...
-    n_done = 1 + atomicInc(global_flags, block_count - 1);
-  }
-  __syncthreads();
-  bool const is_last_block = (n_done == static_cast<int>(block_count));
-
+  int n_done = 0;
+  const bool is_last_block = !__syncthreads_or(
+     threadIdx.y
+         ? 0
+         : (1 + atomicInc(global_flags, block_count - 1) < block_count));
```
dalg24: Sounds like a convoluted way to do `__syncthreads_and(threadIdx.y ? 1 : (atomicInc(global_flags, block_count - 1)  == block_count - 1));`

arghdos:
>Do you have any idea why that is? I see similar behavior in https://github.com/kokkos/kokkos/pull/6035 but can't make much sense of it. I mean the shuffle reduction could always be implemented in terms of local memory so I would expect better performance if shuffle reductions can be used.

Mainly from what I found:

https://github.com/kokkos/kokkos/blob/80dd6adad7b9333eadd318dc8f98b27a3fe4d49c/core/src/HIP/Kokkos_HIP_Shuffle_Reduce.hpp#L69

division by non-compile time constant

https://github.com/kokkos/kokkos/blob/80dd6adad7b9333eadd318dc8f98b27a3fe4d49c/core/src/HIP/Kokkos_HIP_Shuffle_Reduce.hpp#L77

division by non-compile time constant in a loop.

I briefly went down the rabbit hole of "can we make these divisons faster" (the answer is probably yes -- making them unsigned, doing specialty algorithms for bounded numerators, etc.), but then I stumbled across a key different w/r/t the LDS reductions, they **only** do divisons by warp_size which is constexpr (and essentially: free, as compared to runtime divs):

https://github.com/kokkos/kokkos/blob/80dd6adad7b9333eadd318dc8f98b27a3fe4d49c/core/src/HIP/Kokkos_HIP_ReduceScan.hpp#L62

https://github.com/kokkos/kokkos/blob/80dd6adad7b9333eadd318dc8f98b27a3fe4d49c/core/src/HIP/Kokkos_HIP_ReduceScan.hpp#L76

So, looking at how the CUDA backend does it, I decided to not look a gift horse in the mouth and just use the existing impl :)

core/src/HIP/Kokkos_HIP_ReduceScan.hpp
arghdos: heh, I blame our comment from years ago.  IIRC, this is how cuda does it as well: https://github.com/kokkos/kokkos/blob/80dd6adad7b9333eadd318dc8f98b27a3fe4d49c/core/src/Cuda/Kokkos_Cuda_ReduceScan.hpp#L611

dalg24: Yes I saw that.  Do you agree the version I suggested is more readable though?

arghdos: yes, agreed.

arghdos:
Talking with @dalg24 and @Rombur on slack, they reminded me of the kokkos tutorials case (specifically: https://github.com/kokkos/kokkos-tutorials/blob/6cc33429eafcbf212fa84934c1b74f7f503011a2/Exercises/04/Solution/exercise_4_solution.cpp#L126-L134) that we saw large dips in before https://github.com/kokkos/kokkos/pull/6029.

Looked at that, and tweaked the lower bound of the range here a bit to improve perf there, there's still a big dip (this happens right at the 1024) point, but it's far better than with my existing heuristic:

image

this does reduce dot perf a bit for smaller array sizes, but puts is ~ back to what Bruno's patch had (i.e., what was in <= 4.0):

image

Rombur:
Retest this please

arghdos:
I was just thinking about that @masterleinad -- we should probably do the same exercise of flipping `TeamPolicy` to `SHMEM` reductions (right now it defaults to shuffle for any non-zero static size element), and tuning both them and `MDRange`

Rombur:
@arghdos can you explain what's the `relax_scratch` curve on the plot.

arghdos:
> @arghdos can you explain what's the `relax_scratch` curve on the plot.

It's the branch I had in my fork that was the PR for https://github.com/kokkos/kokkos/pull/6029

arghdos:
> I was just thinking about that @masterleinad -- we should probably do the same exercise of flipping `TeamPolicy` to `SHMEM` reductions (right now it defaults to shuffle for any non-zero static size element), and tuning both them and `MDRange`

Let's open another issue and assign to me / @IanBogle to follow up on w/ this one

ajpowelsnl:
@arghdos , @IanBogle -- I will create a Kokkos issue for `parallel_reduce` with `shmem`, and reference this issue.

core/src/HIP/Kokkos_HIP_Parallel_Range.hpp
```diff
@@ -122,6 +122,21 @@ class ParallelReduce<CombinedFunctorReducerType, Kokkos::RangePolicy<Traits...>,
...
+  // and when calculating indexes into the shared/global memory buffers for
+  // performing the scan, word_size_type is used again.
+  // For scalars > 4 bytes in size, indexing into shared/global memory relies
+  // on the block and grid dimensions to ensure that we index at the correct
+  // offset rather than at every 4 byte word; such that, when the join is
+  // performed, we have the correct data that was copied over in chunks of 4
+  // bytes.
+  using word_size_type = std::conditional_t<
+      sizeof(value_type) < sizeof(size_type),
+      std::conditional_t<sizeof(value_type) == 2, int16_t, int8_t>, size_type>;
```
masterleinad, line 139: This and related changes should be a separate pull request. AFAICT, this is not for performance but for correctness with small types, right?

arghdos: >this is not for performance but for correctness with small types, right?

Correct.  I believe what happened here was that when I switched from Shfl->SHMEM reductions, the tests for the small types started failing, so it would be a bit awkward IMO to split this from 

https://github.com/kokkos/kokkos/pull/6229/files#diff-4f2a2e882baf2b7a7ae9265652801ba48521779534ee182a14a1edd8cc1fb164R152

I could see splitting that from the heuristic change tho

Rombur: I think it makes sense to have part of the PR. The PR needs this and it's a really small change.

masterleinad: If this pull request indeed exposes the issue we were trying to trigger, I'm fine with doing it here.

masterleinad (approved):

core/src/HIP/Kokkos_HIP_Parallel_Range.hpp
Rombur: I don't see the point of a cherry-pick in the patch release, since you cannot trigger the bug without the other changes from this PR. The reason we didn't need this change until now is because we did not use the bugged code path.

PhilMiller: That would explain why in #5333/#5641/#5555 some tests failed on CUDA but not HIP. Note from those that there are probably several other reduction paths that still need the corresponding fix.

core/src/HIP/Kokkos_HIP_Parallel_Range.hpp
```diff
@@ -163,13 +177,14 @@ class ParallelReduce<CombinedFunctorReducerType, Kokkos::RangePolicy<Traits...>,
 
   __device__ inline void run(SHMEMReductionTag) const {
     const ReducerType& reducer = m_functor_reducer.get_reducer();
-    const integral_nonzero_constant<
-        size_type, ReducerType::static_value_size() / sizeof(size_type)>
+    const integral_nonzero_constant<word_size_type,
+                                    ReducerType::static_value_size() /
+                                        sizeof(word_size_type)>
         word_count(reducer.value_size() / sizeof(size_type));
```
masterleinad, line 183: Shouldn't this be
```suggestion
        word_count(reducer.value_size() / sizeof(word_size_type));
```
?

masterleinad (commented):
I verified that we indeed need the changes here when forcing reductions to use local memory instead of shuffles.

core/src/HIP/Kokkos_HIP_Parallel_Range.hpp
arghdos: Ahh, yes, you're right.

core/src/HIP/Kokkos_HIP_ReduceScan.hpp
```diff
@@ -395,20 +387,11 @@ __device__ bool hip_single_inter_block_reduce_scan_impl(
...
-  __shared__ int n_done;
-  n_done = 0;
-  __syncthreads();
-  if (threadIdx.y == 0) {
-    n_done = 1 + atomicInc(global_flags, block_count - 1);
-  }
-  __syncthreads();
-  bool const is_last_block = (n_done == static_cast<int>(block_count));
-
+  int n_done               = 0;
```
masterleinad, line 390:
```
/var/jenkins/workspace/Kokkos/core/src/HIP/Kokkos_HIP_ReduceScan.hpp:390:7: error: unused variable 'n_done' [-Werror,-Wunused-variable]
  int n_done               = 0;
      ^
```
```suggestion
```

masterleinad (changes requested):

masterleinad (approved):

dalg24:
Retest this please

crtrott (approved):
I think I am ok with this. I would like to ask folks like Stan Moore to test this out and see if we observe regressions. But we should merge this now and get reports in I guess.

crtrott:
I will dismiss @Rombur approval, just want to confirm with him that its still valid a month later. If he confirms I will merge.

Rombur (approved):

Issue #6217, opened 2023-06-13
compute-sanitizer --tool=racecheck hazard warning in parallel_reduce()


tcclevenger:
`compute-sanitizer --tool=racecheck` gives many "hazard" warnings for `Kokkos::parallel_reduce()`. Here is a very basic reproducer

```
Weaver V100 GPU, modules: gcc/9.3.0, cuda/11.8.0, openmpi/4.1.4, cmake/3.25.1

CODE:

 1 include <Kokkos_Core.hpp>
 2
 3 int main(int argc, char* argv[]) {
 4   Kokkos::initialize(argc, argv);
 5   {
 6     double result;
 7     Kokkos::parallel_reduce("reduce_nothing", 1, KOKKOS_LAMBDA(int, double&) {
 8       // Do nothing;                                                                                                                                                                                                    
 9     }, result);
10   }
11   Kokkos::finalize();
12   printf("Test complete\n");
13 }
14

COMMAND:

compute-sanitizer --tool=racecheck ./racecheck_test

OUTPUT:

========= COMPUTE-SANITIZER
========= Warning: Race reported between Read access at 0x230 in void Kokkos::Impl::cuda_parallel_launch_local_memory<Kokkos::Impl::ParallelReduce<Kokkos::Impl::CombinedFunctorReducer<main::[lambda(int, double &) (instance 1)], Kokkos::Impl::FunctorAnalysis<Kokkos::Impl::FunctorPatternInterface::REDUCE, Kokkos::RangePolicy<Kokkos::Cuda>, main::[lambda(int, double &) (instance 1)], double>::Reducer, void>, Kokkos::RangePolicy<Kokkos::Cuda>, Kokkos::Cuda, Kokkos::Cuda>>(T1)
=========     and Write access at 0x260 in void Kokkos::Impl::cuda_parallel_launch_local_memory<Kokkos::Impl::ParallelReduce<Kokkos::Impl::CombinedFunctorReducer<main::[lambda(int, double &) (instance 1)], Kokkos::Impl::FunctorAnalysis<Kokkos::Impl::FunctorPatternInterface::REDUCE, Kokkos::RangePolicy<Kokkos::Cuda>, main::[lambda(int, double &) (instance 1)], double>::Reducer, void>, Kokkos::RangePolicy<Kokkos::Cuda>, Kokkos::Cuda, Kokkos::Cuda>>(T1) [6272 hazards]
========= 
========= ...7 other very similar Warning: messages...
========= 
[...]

PR #4855, opened 2022-03-09
CUDA Reductions: Fix data races reported by Nvidia compute-sanitizer

PhilMiller:
Detected with Nvidia's compute-sanitizer racecheck tool, with the tiny test case

```c++
#include <Kokkos_Core.hpp>
#include <cstdlib>
#include <iostream>

struct Sum {
  using value_type = double;

  KOKKOS_INLINE_FUNCTION
  void operator()(const int i, double& lsum) const {
    lsum += i;
  }
};

int main(int argc, char** argv)
{
  Kokkos::initialize(argc, argv);

  int n = std::atoi(argv[1]);
  std::cout << "Running with n = " << n << std::endl;
  {
    Kokkos::View<double*> v("foo", n);
    double val;
    Sum sum;
    Kokkos::parallel_reduce(n, sum, val);
    Kokkos::fence();
  }
  Kokkos::finalize();
  return 0;
}
```

Before this change:

```
$ compute-sanitizer --tool=racecheck ./redtest 1
========= COMPUTE-SANITIZER
Running with n = 1
========= WARNING: Race reported between Read access at 0xf10 in /scratch/pbmille/kokkos/install/cuda10/include/impl/Kokkos_FunctorAdapter.hpp:1571:Kokkos::Impl::CudaReductionsFunctor<Sum,void,bool=0,bool=0>::scalar_intra_warp_reduction(Sum const &,double*,bool,int)
=========     and Write access at 0x8a0 in /scratch/pbmille/kokkos/install/cuda10/include/Cuda/Kokkos_Cuda_ReduceScan.hpp:557:Kokkos::Impl::CudaReductionsFunctor<Sum,void,bool=0,bool=0>::scalar_intra_block_reduction(Sum const &,double,bool,double*,int,double*) [3968 hazards]
========= 
========= WARNING: Race reported between Read access at 0x11a0 in /scratch/pbmille/kokkos/install/cuda10/include/Cuda/Kokkos_Cuda_ReduceScan.hpp:548:Kokkos::Impl::CudaReductionsFunctor<Sum,void,bool=0,bool=0>::scalar_intra_warp_reduction(Sum const &,double*,bool,int)
[...]

PR #6236, opened 2023-06-22
Potential data race in Cuda parallel reduce

tcclevenger:
See comment https://github.com/kokkos/kokkos/issues/6217#issuecomment-1603420863. Suggested in https://github.com/kokkos/kokkos/pull/4855#discussion_r822238516 as being a potential issue, I think this is an actual issue. At the very least we avoid doing computation that is unnecessary, and it clears a warning when running `compute-sanitizer --tool=racecheck` (see https://github.com/kokkos/kokkos/issues/6217).

### DIFF
--- a/core/src/Cuda/Kokkos_Cuda_ReduceScan.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_ReduceScan.hpp
@@ -315,7 +315,7 @@ struct CudaReductionsFunctor<FunctorType, false, false> {
     __syncwarp(mask);
 
     for (int delta = skip_vector ? blockDim.x : 1; delta < width; delta *= 2) {
-      if (lane_id + delta < 32) {
+      if ((lane_id + delta < 32) && (lane_id % (delta * 2) == 0)) {
         functor.join(value, value + delta);
       }
       __syncwarp(mask);


Rombur (commented):
We probably need the same fix for HIP

PhilMiller:
> We probably need the same fix for HIP

@Rombur In which case, the fixes from #4855 would likely be applicable as well.

masterleinad:
Have you measured if you see any impact in performance for, say https://github.com/kokkos/kokkos-tutorials/tree/main/Exercises/04/Solution?

tcclevenger:
> Have you measured if you see any impact in performance for, say https://github.com/kokkos/kokkos-tutorials/tree/main/Exercises/04/Solution?

Yes, the differences were negligible (<2% relative increase/decrease between this branch and develop commit it is based on, various problem sizes, launching p_for/reduce kernels).

masterleinad (approved):

cz4rs (approved):
Looks good to me 👍 


> We probably need the same fix for HIP

I propose we merge this as is and fix HIP with separate issue / PR.

PhilMiller:
Retest this please

PhilMiller:
The previous Jenkins failure was long enough ago that the logs have been discarded

PhilMiller:
> Looks good to me 👍
> 
> > We probably need the same fix for HIP
> 
> I propose we merge this as is and fix HIP with a separate issue / PR.

#6348

crtrott (approved):

crtrott:
Retest this please.

PhilMiller:
CUDA builds all passed. None of the pending Jenkins builds is a CUDA backend, so this should be safe to merge.

PR #6235, opened 2023-06-22
#5635: CUDA: Add Overloads for parallel_scan with return value for ThreadVectorRange

thearusable:
Related to #5635 #6453
~Depends on https://github.com/kokkos/kokkos/pull/6292~ (merged)

---

Edit: this now also contains a fix for Cuda parallel_scan ThreadVectorRange range (4a819b6b799b07e818727d49650571a02d87d5d1).

Issue #5635, opened 2022-11-17
Team parallel_scan with return_value


jstrzebonski:
For the purpose of https://github.com/kokkos/kokkos/pull/5150 we miss `parallel_scan` overload that takes parameters:
`const &TeamThreadRange loop_bounds, const FunctorType& lambda, ReturnType& return_value`.

Would that be possible to implement?

fnrizzi:
where is this missing? at the public API level or the impl level?

jstrzebonski:
No, sorry, that's just an example. It's missing for every Team.

jstrzebonski:
Here are examples for CUDA, HIP, SYCL, Threads and HostThread:

https://github.com/kokkos/kokkos/blob/develop/core/src/Cuda/Kokkos_Cuda_Team.hpp#L720
https://github.com/kokkos/kokkos/blob/develop/core/src/HIP/Kokkos_HIP_Team.hpp#L576
https://github.com/kokkos/kokkos/blob/develop/core/src/SYCL/Kokkos_SYCL_Team.hpp#L602
https://github.com/kokkos/kokkos/blob/develop/core/src/Threads/Kokkos_ThreadsTeam.hpp#L1007
https://github.com/kokkos/kokkos/blob/develop/core/src/impl/Kokkos_HostThreadTeam.hpp#L897

dalg24:
yes possible to implement and yes we should add it

fnrizzi:
@mperrinel 
just starting this , to complete: 

Existing overloads for parallel_scan: 

# CUDA
```
template <typename iType, typename FunctorType>
KOKKOS_INLINE_FUNCTION void parallel_scan(
    const Impl::TeamThreadRangeBoundariesStruct<iType, Impl::CudaTeamMember>& loop_bounds,
   const FunctorType& lambda) {
```

mperrinel:
Existing overloads for parallel_scan that can be called from the device directly:

# CUDA
```
template <typename iType, class Closure>
KOKKOS_INLINE_FUNCTION void parallel_scan(
    const Impl::ThreadVectorRangeBoundariesStruct<iType, Impl::CudaTeamMember>& loop_boundaries,
    const Closure& closure)

template <typename iType, typename FunctorType>
KOKKOS_INLINE_FUNCTION void parallel_scan(
    const Impl::TeamThreadRangeBoundariesStruct<iType, Impl::CudaTeamMember>& loop_bounds,
    const FunctorType& lambda)
   
```

[...]

PR #6242, opened 2023-06-27
#5635: HIP: Add Overloads for parallel_scan with return value for ThreadVectorRange

thearusable:
Related to #5635 #6453

~Depends on #6235~ (merged)

### DIFF
--- a/core/src/HIP/Kokkos_HIP_Team.hpp
+++ b/core/src/HIP/Kokkos_HIP_Team.hpp
@@ -797,7 +797,8 @@ parallel_scan(const Impl::ThreadVectorRangeBoundariesStruct<
     // exclusive scan -- the final accumulation
     // of i's val will be included in the second
     // closure call later.
-    if (i < loop_boundaries.end && threadIdx.x > 0) closure(i - 1, val, false);
+    if (i - 1 < loop_boundaries.end && threadIdx.x > 0)
+      closure(i - 1, val, false);
 
     // Bottom up exclusive scan in triangular pattern
     // where each HIP thread is the root of a reduction tree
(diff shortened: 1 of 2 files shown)


masterleinad (changes requested):
Please avoid copying the implementation from the overloads not returning the total.

thearusable:
> Why does this pull request also touch `Cuda`?

I created the PRs as a chain which should be merged in particular order. This one is the last one currently created so it contains the changes from other ones. There will be at least two more PRs created.

masterleinad:
Marking as draft since dependencies have not been merged.

masterleinad (approved):

core/src/HIP/Kokkos_HIP_Team.hpp
```diff
@@ -797,7 +797,8 @@ parallel_scan(const Impl::ThreadVectorRangeBoundariesStruct<
     // exclusive scan -- the final accumulation
     // of i's val will be included in the second
     // closure call later.
-    if (i < loop_boundaries.end && threadIdx.x > 0) closure(i - 1, val, false);
+    if (i - 1 < loop_boundaries.end && threadIdx.x > 0)
+      closure(i - 1, val, false);
```
cz4rs, line 801: See the discussion https://github.com/kokkos/kokkos/pull/6235#discussion_r1331460296 for a similar fix in CUDA backend.

fnrizzi (approved):

fnrizzi:
gcc timed out , rest passes

core/unit_test/TestTeamVector.hpp
```diff
@@ -1016,7 +1016,7 @@ struct checkScan {
 };
 }  // namespace VectorScanReducer
 
-#if !(defined(KOKKOS_IMPL_CUDA_CLANG_WORKAROUND) || defined(KOKKOS_ENABLE_HIP))
+#if !defined(KOKKOS_IMPL_CUDA_CLANG_WORKAROUND)
```
Rombur, line 1019: If it works for HIP and CUDA nvcc, there is a high probability that it also works for CUDA clang. Not for this PR but I think it's worth to check if the still need this condition.

Rombur (approved):

crtrott (approved):

PR #6246, opened 2023-06-28
SYCL: Use in-order queues in InterOp tests

masterleinad:
Follow-up to https://github.com/kokkos/kokkos/pull/6189. Since we might be relying on `sycl::queues` to be in-order now, we should also use in-order queues in the interoperability tests to be on the safe side.

### DIFF
--- a/core/src/SYCL/Kokkos_SYCL.hpp
+++ b/core/src/SYCL/Kokkos_SYCL.hpp
@@ -163,7 +163,8 @@ std::vector<SYCL> partition_space(const SYCL& sycl_space, Args...) {
   std::vector<SYCL> instances;
   instances.reserve(sizeof...(Args));
   for (unsigned int i = 0; i < sizeof...(Args); ++i)
-    instances.emplace_back(sycl::queue(context, device));
+    instances.emplace_back(
+        sycl::queue(context, device, sycl::property::queue::in_order()));
   return instances;
 }
 
@@ -183,7 +184,8 @@ std::vector<SYCL> partition_space(const SYCL& sycl_space,
   // otherwise.
   instances.reserve(weights.size());
   for (unsigned int i = 0; i < weights.size(); ++i)
-    instances.emplace_back(sycl::queue(context, device));
+    instances.emplace_back(
+        sycl::queue(context, device, sycl::property::queue::in_order()));
   return instances;
 }
 }  // namespace Experimental


crtrott (approved):

crtrott:
Retest this please.

dalg24 (commented):
Should we not be that a precondition in SYCL constructor that accept a queue?
By that it means, why aren't we checking that the user-provided queue has the "in order" property?

masterleinad:
> By that it means, why aren't we checking that the user-provided queue has the "in order" property?

Good idea. https://github.com/kokkos/kokkos/pull/6246/commits/55a589927945ae6eb5447d26462064dfaec736e6

core/src/SYCL/Kokkos_SYCL.cpp
```diff
@@ -59,6 +59,8 @@ SYCL::SYCL(const sycl::queue& stream)
         ptr->finalize();
         delete ptr;
       }) {
+  if (!stream.is_in_order())
+    Kokkos::abort("User provided sycl::queues must be in-order!");
```
dalg24, line 63:
```suggestion
#ifdef KOKKOS_IMPL_SYCL_USE_IN_ORDER_QUEUES
  if (!stream.is_in_order())
    Kokkos::abort("User provided sycl::queues must be in-order!");
#endif
```

masterleinad: I thought about it but decided to play it safe and make the check independent of what we do internally. What are your thoughts?

crtrott:
We could make the things which are protected by in-order macro protected by a runtime check instead. Thus it would work either way for interop purposes. We could also wait with that until someone asks and until then we unconditionally check.

core/src/SYCL/Kokkos_SYCL.cpp
```diff
@@ -59,6 +59,11 @@ SYCL::SYCL(const sycl::queue& stream)
         ptr->finalize();
         delete ptr;
       }) {
+  // To be on the safe side if we decide to switch the default to out-of-order
+  // queues again, we require user-provided queues to be in-order
+  // unconditionally.
```
dalg24, line 64:
```suggestion
  // In principle could be guarded with
  // #ifdef KOKKOS_IMPL_SYCL_USE_IN_ORDER_QUEUES
  // but we chose to require user-provided queues to be in-order
  // unconditionally so that code downstream does not break
  // when the backend setting changes.
```

dalg24 (approved):

dalg24:
SYCL tests do not pass (precondition is in-order queue violations)

masterleinad:
> SYCL tests do not pass (precondition is in-order queue violations)

Sorry about that. Fixed!

masterleinad:
For the `OpenMPTarget` CI build we have
```
 1 - Kokkos_CoreUnitTest_Serial1 (Timeout)
 4 - Kokkos_CoreUnitTest_OpenMPTarget (Timeout)
18 - Kokkos_CoreUnitTest_StackTraceTest (Timeout)
```
which looks unrelated.

dalg24:
Please make sure you mention the enforcing the precondition in the changelog

PR #6247, opened 2023-06-28
Remove calling tribits_exclude_autotools_files()

bartlettroscoe:
Kokkos has no autotools files anymore and support for `tribits_exclude_autotools_files()` was stripped out of TriBITS because it is not needed anymore (see https://github.com/TriBITSPub/TriBITS/pull/586).

There will be a slight hack in Trilinos 'develop' to avoid needing to make this change on Trilinos 'develop'.  But the next time Kokkos is snapshotted into Trilinos, we can remove that (minor) hack.


masterleinad (approved):

cz4rs (approved):

Rombur:
OK to test

masterleinad:
> HIP-ROCm-5.2-C++20 build aborted, which was the only CI failure.
@masterleinad @dalg24 should this be retested, or is the aborted build a transient system issue we can ignore?

I don't think we care about CI results here apart from possibly checking that `CMake` runs successfully.

PR #6248, opened 2023-06-28
Check AVX, AVX2, AVX512XEON compiler macros on KokkosCore_config.h

masterleinad:
We discussed that we also want to have the reverse check as the one in #6188, i.e., check that we actually see the expected compiler macro when setting an architecture (for `AVX`, `AVX2`, `AVX512XEON` in this pull request).

### DIFF
--- a/simd/src/Kokkos_SIMD.hpp
+++ b/simd/src/Kokkos_SIMD.hpp
@@ -26,12 +26,14 @@
 #if defined(KOKKOS_ARCH_AVX) && !defined(__AVX__)
 #error "__AVX__ must be defined for KOKKOS_ARCH_AVX"
 #endif
+
 #if defined(KOKKOS_ARCH_AVX2)
 #if !defined(__AVX2__)
 #error "__AVX2__ must be defined for KOKKOS_ARCH_AVX2"
 #endif
 #include <Kokkos_SIMD_AVX2.hpp>
 #endif
+
 #if defined(KOKKOS_ARCH_AVX512XEON)
 #if !defined(__AVX512F__)
 #error "__AVX512F__ must be defined for KOKKOS_ARCH_AVX512XEON"
--- a/simd/src/Kokkos_SIMD.hpp
+++ b/simd/src/Kokkos_SIMD.hpp
@@ -21,11 +21,21 @@
 
 #include <Kokkos_SIMD_Scalar.hpp>
 
-#ifdef KOKKOS_ARCH_AVX2
+#include <Kokkos_Macros.hpp>
+
+#if defined(KOKKOS_ARCH_AVX) && !defined(__AVX__)
+#error "__AVX__ must be defined for KOKKOS_ARCH_AVX"
+#endif
+#if defined(KOKKOS_ARCH_AVX2)
+#if !defined(__AVX2__)
+#error "__AVX2__ must be defined for KOKKOS_ARCH_AVX2"
+#endif
 #include <Kokkos_SIMD_AVX2.hpp>
 #endif
-
-#ifdef KOKKOS_ARCH_AVX512XEON
+#if defined(KOKKOS_ARCH_AVX512XEON)
+#if !defined(__AVX512F__)
+#error "__AVX512F__ must be defined for KOKKOS_ARCH_AVX512XEON"
+#endif
 #include <Kokkos_SIMD_AVX512.hpp>
 #endif
 
--- a/cmake/KokkosCore_config.h.in
+++ b/cmake/KokkosCore_config.h.in
@@ -70,17 +70,8 @@
 #cmakedefine KOKKOS_ARCH_ARMV8_THUNDERX2
 #cmakedefine KOKKOS_ARCH_A64FX
 #cmakedefine KOKKOS_ARCH_AVX
-#if defined(KOKKOS_ARCH_AVX) && !defined(__AVX__)
-#error "__AVX__ must be defined for KOKKOS_ARCH_AVX"
-#endif
 #cmakedefine KOKKOS_ARCH_AVX2
-#if defined(KOKKOS_ARCH_AVX2) && !defined(__AVX2__)
-#error "__AVX2__ must be defined for KOKKOS_ARCH_AVX2"
-#endif
 #cmakedefine KOKKOS_ARCH_AVX512XEON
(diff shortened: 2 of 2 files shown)


crtrott:
What do other folks think about the place where to check this? We could also do it in Kokkos_Macros.hpp, or have a separate file included in Kokkos_Macros.hpp (i.e. add these checks inside: Kokkos_Architecture_Macro_Verification.hpp).

dalg24:
I am not sure about the direction.  I would like to look into dropping these `KOKKOS_ARCH_AVX[2,512XEON]` macros

crtrott:
I am pretty sure we need a way of testing. Say you build Trilinos with AVX512, but then you build an App on top of it with only AVX2. That would lead to issues we should catch.

masterleinad:
> I am not sure about the direction. I would like to look into dropping these `KOKKOS_ARCH_AVX[2,512XEON]` macros

If you don't have these (or similar macros), you could only detect SIMD support directly through the compiler macros and that is the approach we moved away from in https://github.com/kokkos/kokkos/pull/6188 to avoid ODR violations/compiling with different architecture flags.

masterleinad:
Retest this please.

dalg24 (commented):
These seem out of place.
Introduce an `impl/Kokkos_AdvancedVectorExtensionsMacros.hpp` file or whatnot.

dalg24 (approved):

simd/src/impl/Kokkos_AdvancedVectorExtensionsMacros.hpp
```diff
@@ -0,0 +1,34 @@
...
+
+#ifndef KOKKOS_ADVANCED_VECTOR_EXTENSIONS_MACROS_HPP
+#define KOKKOS_ADVANCED_VECTOR_EXTENSIONS_MACROS_HPP
+
+#include <Kokkos_Macros.hpp>
+
+#if defined(KOKKOS_ARCH_AVX) && !defined(__AVX__)
+#error "__AVX__ must be defined for KOKKOS_ARCH_AVX"
+#endif
+#cmakedefine KOKKOS_ARCH_AVX2
```
dalg24, line 25:
```suggestion

```

dalg24 (changes requested):

simd/src/impl/Kokkos_AdvancedVectorExtensionsMacros.hpp
masterleinad: Fixed.

ldh4 (commented):
A few thoughts.
Will there be a separate check for NEON?

I think changing the file name to something like `Kokkos_AVXMacrosCheck.hpp` could be useful as well, since the file doesn't really introduce any macros about AVX.

Also, I think this part could be moved and incorporated into the new file as part of its check:
```
#ifdef KOKKOS_ARCH_AVX2
#include <Kokkos_SIMD_AVX2.hpp>
#endif
#ifdef KOKKOS_ARCH_AVX512XEON
#include <Kokkos_SIMD_AVX512.hpp>
#endif
```

masterleinad:
> Will there be a separate check for NEON?

Yes, once we have this one and #6188 in, I wanted to propose the respective changes for `NEON` that are somewhat backward-incompatible.

simd/src/impl/Kokkos_AdvancedVectorExtensionsMacros.hpp
dalg24: I meant the empty line but anyway I think DongHun's suggestion is better.

masterleinad:
I moved the checks inside `Kokkos_SIMD.hpp`. I like that better than creating a separate file as well.

simd/src/Kokkos_SIMD.hpp
```diff
@@ -21,11 +21,17 @@
...
-#ifdef KOKKOS_ARCH_AVX2
+#include <Kokkos_Macros.hpp>
+
+#if defined(KOKKOS_ARCH_AVX) && !defined(__AVX__)
+#error "__AVX__ must be defined for KOKKOS_ARCH_AVX"
+#endif
+#if defined(KOKKOS_ARCH_AVX2) && !defined(__AVX2__)
+#error "__AVX2__ must be defined for KOKKOS_ARCH_AVX2"
 #include <Kokkos_SIMD_AVX2.hpp>
 #endif
```
ldh4, line 32: I agree with moving the check into `Kokkos_SIMD.hpp`, but is the condition to include each header correct?

Shouldn't it be something like this?
```suggestion
#if defined(KOKKOS_ARCH_AVX2)
#if !defined(__AVX2__)
#error "__AVX2__ must be defined for KOKKOS_ARCH_AVX2"
#else
#include <Kokkos_SIMD_AVX2.hpp>
#endif
#endif
```

masterleinad: Of course, fixed!

ldh4 (approved):

simd/src/Kokkos_SIMD.hpp
```diff
@@ -21,11 +21,21 @@
 
 #include <Kokkos_SIMD_Scalar.hpp>
 
-#ifdef KOKKOS_ARCH_AVX2
+#include <Kokkos_Macros.hpp>
+
+#if defined(KOKKOS_ARCH_AVX) && !defined(__AVX__)
+#error "__AVX__ must be defined for KOKKOS_ARCH_AVX"
+#endif
```
dalg24, line 28: Break a line plz

masterleinad: Here you go!

dalg24 (approved):

masterleinad:
`CUDA-12.2-NVHPC`, `HIP-ROCm-5.2`, and `HIP-ROCm-5.6-C++20` were timing out. Everything else passed.

simd/src/Kokkos_SIMD.hpp
```diff
@@ -21,11 +21,23 @@
 
 #include <Kokkos_SIMD_Scalar.hpp>
 
-#ifdef KOKKOS_ARCH_AVX2
+#include <Kokkos_Macros.hpp>
```
dalg24, line 24: This header include is unnecessary but whatever

PR #6250, opened 2023-06-30
OpenACC CMakechange Clacc

seyonglee:
This PR contains changes on the cmake configurations and OpenACC codes that are necessary for the LLVM-Clacc compiler to compile the OpenACC backend.
- Clacc supports both NVIDIA GPUs and AMD GPUs as the Kokkos OpenACC backend target.
- Clacc can be downloaded from the LLVM-DOE repository (https://github.com/llvm-doe-org/llvm-project/tree/clacc/main).


cmake/kokkos_arch.cmake
```diff
@@ -89,7 +89,7 @@ KOKKOS_ARCH_OPTION(AMPERE86        GPU  "NVIDIA Ampere generation CC 8.6"  "KOKK
 KOKKOS_ARCH_OPTION(ADA89           GPU  "NVIDIA Ada generation CC 8.9"     "KOKKOS_SHOW_CUDA_ARCHS")
 KOKKOS_ARCH_OPTION(HOPPER90        GPU  "NVIDIA Hopper generation CC 9.0"  "KOKKOS_SHOW_CUDA_ARCHS")
 
-IF(Kokkos_ENABLE_HIP OR Kokkos_ENABLE_OPENMPTARGET)
+IF(Kokkos_ENABLE_HIP OR Kokkos_ENABLE_OPENMPTARGET  OR Kokkos_ENABLE_OPENACC)
```
masterleinad, line 92:
```suggestion
IF(Kokkos_ENABLE_HIP OR Kokkos_ENABLE_OPENMPTARGET OR Kokkos_ENABLE_OPENACC)
```

algorithms/CMakeLists.txt
```diff
@@ -2,6 +2,6 @@ IF (NOT Kokkos_INSTALL_TESTING)
   ADD_SUBDIRECTORY(src)
 ENDIF()
 # FIXME_OPENACC: temporarily disabled due to unimplemented features
-IF(NOT ((KOKKOS_ENABLE_OPENMPTARGET OR KOKKOS_ENABLE_OPENACC) AND KOKKOS_CXX_COMPILER_ID STREQUAL NVHPC))
+IF(NOT (((KOKKOS_ENABLE_OPENMPTARGET OR KOKKOS_ENABLE_OPENACC) AND KOKKOS_CXX_COMPILER_ID STREQUAL NVHPC) OR (KOKKOS_ENABLE_OPENACC AND KOKKOS_CXX_COMPILER_ID STREQUAL Clang)))
```
masterleinad, line 5: Is there any compiler for which we expect the unit tests to work? Otherwise, I would just do
```suggestion
IF((NOT KOKKOS_ENABLE_OPENMPTARGET AND KOKKOS_CXX_COMPILER_ID STREQUAL NVHPC) OR KOKKOS_ENABLE_OPENACC)
```

cmake/kokkos_arch.cmake
```diff
@@ -631,8 +631,8 @@ FUNCTION(CHECK_AMDGPU_ARCH ARCH FLAG)
       MESSAGE(FATAL_ERROR "Multiple GPU architectures given! Already have ${AMDGPU_ARCH_ALREADY_SPECIFIED}, but trying to add ${ARCH}. If you are re-running CMake, try clearing the cache and running again.")
     ENDIF()
     SET(AMDGPU_ARCH_ALREADY_SPECIFIED ${ARCH} PARENT_SCOPE)
-    IF (NOT KOKKOS_ENABLE_HIP AND NOT KOKKOS_ENABLE_OPENMPTARGET)
-      MESSAGE(WARNING "Given AMD GPU architecture ${ARCH}, but Kokkos_ENABLE_HIP and Kokkos_ENABLE_OPENMPTARGET are OFF. Option will be ignored.")
+    IF (NOT KOKKOS_ENABLE_HIP AND NOT KOKKOS_ENABLE_OPENMPTARGET  AND NOT KOKKOS_ENABLE_OPENACC)
```
masterleinad, line 634:
```suggestion
    IF (NOT KOKKOS_ENABLE_HIP AND NOT KOKKOS_ENABLE_OPENMPTARGET AND NOT KOKKOS_ENABLE_OPENACC)
```

cmake/kokkos_arch.cmake
```diff
@@ -744,14 +744,32 @@ ENDIF()
...
     COMPILER_SPECIFIC_FLAGS(
       NVHPC -acc
     )
+    # When not compiling for offload to any GPU, we're compiling for kernel
+    # execution on the host.  In that case, memory is shared between the OpenACC
+    # space and the host space.
+    COMPILER_SPECIFIC_DEFS(
+      Clang KOKKOS_OPENACC_WITHOUT_GPU
+      NVHPC KOKKOS_OPENACC_WITHOUT_GPU
+    )
```
masterleinad, line 772: Do we actually care about that case apart from debugging (and then the shared memory space doesn't really help)?

cmake/kokkos_enable_devices.cmake
```diff
@@ -41,6 +41,17 @@ ENDIF()
 KOKKOS_DEVICE_OPTION(OPENMP ${OMP_DEFAULT} HOST "Whether to build OpenMP backend")
 
 KOKKOS_DEVICE_OPTION(OPENACC OFF DEVICE "Whether to build the OpenACC backend")
+IF (KOKKOS_ENABLE_OPENACC)
+  COMPILER_SPECIFIC_FLAGS(
+    Clang -fopenacc -fopenacc-fake-async-wait
+          -Wno-openacc-and-cxx -Wno-openmp-mapping -Wno-unknown-cuda-version
+          -Wno-pass-failed
+          # -Wno-defaulted-function-deleted
```
masterleinad, line 49: What's up with this flag?

core/src/OpenACC/Kokkos_OpenACC_ParallelScan_Range.hpp
```diff
@@ -201,6 +202,164 @@ class ParallelScanOpenACCBase {
                               final_reducer)async(async_arg)
     acc_wait(async_arg);
   }
+#else
+  // Alternative implementation to work around OpenACC features not yet
+  // implemented by Clacc
```
masterleinad, line 207: What's the difference between the two implementations?

core/src/OpenACC/Kokkos_OpenACC_Traits.hpp
```diff
@@ -25,6 +25,9 @@ struct OpenACC_Traits {
 #if defined(KOKKOS_IMPL_ARCH_NVIDIA_GPU)
   static constexpr acc_device_t dev_type     = acc_device_nvidia;
   static constexpr bool may_fallback_to_host = false;
+#elif defined(KOKKOS_ARCH_VEGA)
```
masterleinad, line 28: also `NAVI`?

cmake/kokkos_arch.cmake
seyonglee: A user may want to execute a kernel on a host (in parallel).

core/src/OpenACC/Kokkos_OpenACC_ParallelScan_Range.hpp
seyonglee: The current implementation of Clacc does not support gang-private variables; thus, the alternative implementation allocates the gang-private arrays on GPU global memory using array expansion.

cmake/kokkos_enable_devices.cmake
seyonglee: Deleted.

core/src/OpenACC/Kokkos_OpenACC_Traits.hpp
seyonglee: Added.

dalg24 (changes requested):
Please withdraw the `KOKKOS_OPENACC_WITHOUT_GPU` changes.
You may propose them in another PR but, as far as I can tell, these don't belong here and they are potentially controversial.

cmake/kokkos_enable_devices.cmake
```diff
@@ -41,6 +41,16 @@ ENDIF()
...
 
 KOKKOS_DEVICE_OPTION(OPENACC OFF DEVICE "Whether to build the OpenACC backend")
+IF (KOKKOS_ENABLE_OPENACC)
+  COMPILER_SPECIFIC_FLAGS(
+    Clang -fopenacc -fopenacc-fake-async-wait
+          -Wno-openacc-and-cxx -Wno-openmp-mapping -Wno-unknown-cuda-version
+          -Wno-pass-failed
+  )
+  COMPILER_SPECIFIC_DEFS(
+    Clang KOKKOS_WORKAROUND_OPENMPTARGET_CLANG
```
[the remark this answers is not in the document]
seyonglee: That macro is originally used by OpenMPTarget/Clang to disable unsupported unit tests in `core/unit_test/TestComplex.hpp`, and the same tests are disabled for OpenACC/Clacc too since Clacc is built on top of the same OpenMP implementation in Clang/LLVM. (Clacc internally performs OpenACC-to-OpenMP translation to use the existing OpenMP implementation in LLVM to support OpenACC.)

seyonglee:
> Please withdraw the `KOKKOS_OPENACC_WITHOUT_GPU` changes.
> You may propose them in another PR but, as far as I can tell, these don't belong here and they are potentially controversial.

Changes related to `KOKKOS_OPENACC_WITHOUT_GPU` are removed, which will be addressed by a separate PR.

core/unit_test/CMakeLists.txt
```diff
@@ -1116,7 +1215,14 @@ if (NOT KOKKOS_HAS_TRILINOS)
       INPUT TestDeviceAndThreads.py
       ${USE_SOURCE_PERMISSIONS_WHEN_SUPPORTED}
     )
-    if(NOT Kokkos_ENABLE_OPENMPTARGET)  # FIXME_OPENMPTARGET does not select the right device
+    if(NOT Kokkos_ENABLE_OPENMPTARGET # FIXME_OPENMPTARGET does not select the right device
+       AND
+       # FIXME_OPENACC - does not select specified device beyond 0 until OpenACC
+       # backend is initialized.  For example, adding a Kokkos::parallel_for to
+       # the start of main in core/unit_test/UnitTest_DeviceAndThreads.cpp makes
+       # the test pass.
```
dalg24, line 1223: I am not sure I understand your point.  Are you saying that `acc_set_device_num` does not have effect until a parallel region actually gets executed?
```C++
int dev_num = <some-non-zero-device-num>;
acc_set_device_num(dev_num);
assert(acc_get_device_num() == dev_num);  // fails

File: core/unit_test/CMakeLists.txt [hunk 10]
```diff
@@ -835,6 +921,19 @@ if ((KOKKOS_ENABLE_OPENMPTARGET OR KOKKOS_ENABLE_OPENACC) AND KOKKOS_CXX_COMPILE
   )
 endif()
 
+# FIXME_OPENACC - Comment non-passing tests with the Clang compiler
+if (KOKKOS_ENABLE_OPENACC AND KOKKOS_CXX_COMPILER_ID STREQUAL Clang)
+  SET(DEFAULT_DEVICE_SOURCES
+    UnitTestMainInit.cpp
+    TestInitializationSettings.cpp
+    TestParseCmdLineArgsAndEnvVars.cpp
+    #default/TestDefaultDeviceType.cpp
```
dalg24, line 930:
```suggestion
```

core/unit_test/CMakeLists.txt
```diff
@@ -835,6 +921,19 @@ if ((KOKKOS_ENABLE_OPENMPTARGET OR KOKKOS_ENABLE_OPENACC) AND KOKKOS_CXX_COMPILE
...
+    default/TestDefaultDeviceType_d.cpp
+    default/TestDefaultDeviceTypeResize.cpp
+    default/TestDefaultDeviceTypeViewAPI.cpp
+  )
```
dalg24, line 934: Does `TestCompilerMacros.cpp` really not pass?

I see now that this is how we currently do it but I would really prefer if we expressed it in terms of what tests are not supported (remove the ones that are not currently supported)

dalg24: @rgayatri23 

Please fix https://github.com/kokkos/kokkos/blob/35c1e3c37402ff8505410456421145e0cd50812c/core/unit_test/CMakeLists.txt#L830-L835
and use
```cmake
list(REMOVE_ITEM DEFAULT_DEVICE_SOURCES
  <remove-this>
  <remove-that>
)

File: core/unit_test/CMakeLists.txt [hunk 12]
```diff
@@ -522,6 +528,86 @@ IF(KOKKOS_ENABLE_OPENACC AND KOKKOS_CXX_COMPILER_ID STREQUAL NVHPC)
...
+    ${CMAKE_CURRENT_BINARY_DIR}/openacc/TestOpenACC_MathematicalConstants.cpp
+    ${CMAKE_CURRENT_BINARY_DIR}/openacc/TestOpenACC_MathematicalSpecialFunctions.cpp
+    ${CMAKE_CURRENT_BINARY_DIR}/openacc/TestOpenACC_MinMaxClamp.cpp
+    ${CMAKE_CURRENT_BINARY_DIR}/openacc/TestOpenACC_ViewLayoutStrideAssignment.cpp
+    ${CMAKE_CURRENT_BINARY_DIR}/openacc/TestOpenACC_ViewMapping_a.cpp
+    ${CMAKE_CURRENT_BINARY_DIR}/openacc/TestOpenACC_ViewMemoryAccessViolation.cpp
+    ${CMAKE_CURRENT_BINARY_DIR}/openacc/TestOpenACC_WithoutInitializing.cpp
+    ${CMAKE_CURRENT_BINARY_DIR}/openacc/TestOpenACC_ViewAPI_d.cpp
+    )
+  # Somehow on ExCL's explorer (AMD GPU), these cause clang-linker-wrapper to
```
dalg24, line 584: What's "ExCL's explorer"?

core/src/OpenACC/Kokkos_OpenACC_ParallelScan_Range.hpp
seyonglee: Updated the comment.

core/unit_test/CMakeLists.txt
seyonglee: Explorer is the name of the machine in the ORNL ExCL cluster (https://excl.ornl.gov).
Updated the comment to provide detailed system info instead of the system name.

core/unit_test/CMakeLists.txt
seyonglee: > Does `TestCompilerMacros.cpp` really not pass?
> 
> I see now that this is how we currently do it but I would really prefer if we expressed it in terms of what tests are not supported (remove the ones that are not currently supported)

`TestCompilerMacros.cpp` passes when compiled by NVHPC V22.11 and Clacc (built on Clang V17.0.0) and thus is re-enabled.

core/unit_test/CMakeLists.txt
seyonglee: The comment and condition-checking code are outdated; deleted.
The original issue was about the interaction between Kokkos initialization and the OpenACC backend implementation, which is fixed now.

core/src/OpenACC/Kokkos_OpenACC_ParallelScan_Range.hpp
```diff
@@ -201,6 +202,167 @@ class ParallelScanOpenACCBase {
...
+               Kokkos::Experimental::OpenACC>(m_policy.space(), m_result_ptr,
+                                              m_result_total.data(),
+                                              sizeof(ValueType));
+    }
+
+#pragma acc exit data delete (functor, chunk_values, offset_values, \
+                              final_reducer)async(async_arg)
+    acc_wait(async_arg);
+  }
+#endif
```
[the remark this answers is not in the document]
seyonglee: Merged the two implementations using macros as suggested.

masterleinad (approved):
Fine with me.

cmake/kokkos_arch.cmake
```diff
@@ -744,9 +744,20 @@ ENDIF()
...
+      Clang -Xopenmp-target=nvptx64-nvidia-cuda -march=${CLANG_CUDA_ARCH}
+            -fopenmp-targets=nvptx64-nvidia-cuda
+    )
+  ELSEIF(KOKKOS_AMDGPU_ARCH_FLAG)
+    COMPILER_SPECIFIC_FLAGS(
+      Clang -Xopenmp-target=amdgcn-amd-amdhsa -march=${KOKKOS_AMDGPU_ARCH_FLAG}
+            -fopenmp-targets=amdgcn-amd-amdhsa
+    )
+    COMPILER_SPECIFIC_LIBS(
+      Clang -lm
```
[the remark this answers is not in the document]
seyonglee: It was added for an unexpected linking error when compiling some unit tests on AMD GPUs with old ROCm, but the error no longer exists (when tested with ROCm V5.4.0); deleted.

core/src/OpenACC/Kokkos_OpenACC_Traits.hpp
```diff
@@ -25,6 +25,9 @@ struct OpenACC_Traits {
 #if defined(KOKKOS_IMPL_ARCH_NVIDIA_GPU)
   static constexpr acc_device_t dev_type     = acc_device_nvidia;
   static constexpr bool may_fallback_to_host = false;
+#elif defined(KOKKOS_ARCH_VEGA) || defined(KOKKOS_ARCH_NAVI)
```
dalg24, line 28: cc @Rombur who is working on #6266 
I think it os ok though.

core/unit_test/CMakeLists.txt
```diff
@@ -1116,7 +1216,7 @@ if (NOT KOKKOS_HAS_TRILINOS)
       INPUT TestDeviceAndThreads.py
       ${USE_SOURCE_PERMISSIONS_WHEN_SUPPORTED}
     )
-    if(NOT Kokkos_ENABLE_OPENMPTARGET)  # FIXME_OPENMPTARGET does not select the right device
+    if(NOT Kokkos_ENABLE_OPENMPTARGET) # FIXME_OPENMPTARGET does not select the right device
```
dalg24, line 1219: plz undo

seyonglee: Reverted.

core/src/OpenACC/Kokkos_OpenACC_ParallelScan_Range.hpp
```diff
@@ -19,8 +19,27 @@
...
+// Clacc uses an alternative implementation to work around not-yet-implemented
+// OpenACC features: Clacc does not fully support private clauses for
+// gang-private variables, and the alternative implementation allocates
+// the gang-private arrays on GPU global memory using array expansion,
+// instead of using the private clause.
+/* clang-format off */
+#ifdef KOKKOS_COMPILER_CLANG
+#define ELEMENT_VALUES_SIZE (n_chunks * 2 * chunk_size)
+#define ACCESS_ELEMENTS(THREADID) \
+  element_values[team_id * 2 * chunk_size + THREADID]
```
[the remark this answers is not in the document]
seyonglee: I think that the current approach is better than using an unmanaged view in terms of the number of code changes between two implementations (one implementation uses the element_values as a gang-private array and the other uses it as a gang-shared array).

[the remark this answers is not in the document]
dalg24: You need to prefix all macros with `KOKKOS_IMPL_[ACC_]`.

What prevents you to use a regular variable instead of a macro for the number of elements in the array?

~`ELEMENT_VALUES_SIZE`~
```C++
#ifdef KOKKOS_COMPILER_CLANG
int const num_elements = n_chunks * 2 * chunk_size;
#else
int const num_elements = 2 * chunk_size;
#endif
```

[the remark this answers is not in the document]
seyonglee: Prefixed all macros with `KOKKOS_IMPL_ACC_`.
Removed ELEMENT_VALUES_SIZE.

core/src/OpenACC/Kokkos_OpenACC_ParallelScan_Range.hpp
```diff
@@ -88,22 +86,27 @@ class ParallelScanOpenACCBase {
         functor(m_functor);
     const IndexType N        = end - begin;
     const IndexType n_chunks = (N + chunk_size - 1) / chunk_size;
+#ifdef KOKKOS_COMPILER_CLANG
+    int const num_elements = n_chunks * 2 * chunk_size;
+#else
+    int const num_elements = 2 * chunk_size;
```
dalg24, line 92: I was excepting you would use `num_elements` in the definition of `KOKKOS_IMPL_ACC_ELEMENT_VALUES_CLAUSE`
Not blocking though

dalg24 (approved):

core/src/OpenACC/Kokkos_OpenACC_ParallelScan_Range.hpp
seyonglee: Updated.

dalg24:
Ignoring unrelated SYCL failure that was fixed in #6293

PR #6069, opened 2023-04-19
SIMD: Add abs() for all int types

ldh4:
Related to #5674 

This PR adds abs() function for all integer simd in each backend:
- simd Kokkos::abs(const simd& lhs)

AVX2:
- Added abs() for `int32_t`, `int64_t` and `int64_t`

AVX512:
- Added shifts for `int32_t`, `uint32_t`, `int64_t` and `uint64_t`

NEON:
- Added shifts for `int32_t`, `int64_t` and `uint64_t`

PR #6251, opened 2023-07-01
Ensure that `Kokkos::complex` only gets instantiated for cv-unqualified floating-point types

dalg24:
https://github.com/kokkos/kokkos/pull/6069#discussion_r1243571228 uncovered a bug where we accidentally instantiated `Kokkos::complex` for integral types.

This PR is open as a base for discussion and proposes to static assert that `complex` is being instantiated for a cv-unqualified floating-point type.  The current implementation would not allow for using half types.

### DIFF
--- a/core/src/Kokkos_Complex.hpp
+++ b/core/src/Kokkos_Complex.hpp
@@ -44,6 +44,11 @@ class
     alignas(2 * sizeof(RealType))
 #endif
         complex {
+  static_assert(std::is_floating_point_v<RealType> &&
+                    std::is_same_v<RealType, std::remove_cv_t<RealType>>,
+                "Kokkos::complex can only be instantiated for a cv-unqualified "
+                "floating point type");
+
  private:
   RealType re_{};
   RealType im_{};


fnrizzi (approved):

nliber (approved):

ldh4 (approved):

dalg24:
Do you guys want a comment about half/bhalf type support?

core/src/Kokkos_Complex.hpp
```diff
@@ -44,6 +44,11 @@ class
     alignas(2 * sizeof(RealType))
 #endif
         complex {
+  static_assert(std::is_floating_point_v<RealType> &&
```
e10harvey, line 47: Shall we merge this and then add a follow on PR with:

```suggestion
  static_assert((std::is_floating_point_v<RealType> &&
                        std::is_same_v<RealType, std::remove_cv_t<RealType>) ||
                        kokkos_type_is_half<RealType>::value)
```

e10harvey (approved):

core/src/Kokkos_Complex.hpp
e10harvey: I do not see any usage of `Kokkos::complex` with `half_t` or `bhalf_t` in Trilinos:
https://github.com/search?q=repo%3Atrilinos%2FTrilinos+Kokkos%3A%3Acomplex+half_t&type=code

dalg24: Except that
* `kokkos_type_is_half` is not yet a thing
* that including `<Kokkos_Half.hpp>` is not acceptable

so we would have to explore some other mechanism to enable if we decide to add support.

dalg24:
Ignoring timeouts in OpenMPTarget build

PR #6252, opened 2023-07-01
Deprecate `Kokkos::vector`

dalg24:
Rational: `Kokkos::vector` might have been relevant in the past to ease the transition to Kokkos but this is really not something we want people to use any more.
It is at times the source of confusion (quite different to `Kokkos::array`, `Kokkos::pair`, or `Kokkos::complex`).

I suggest we mark it as deprecated and prepare for removal in Kokkos 5.0

### DIFF
--- a/containers/src/Kokkos_Vector.hpp
+++ b/containers/src/Kokkos_Vector.hpp
@@ -21,6 +21,17 @@
 #define KOKKOS_IMPL_PUBLIC_INCLUDE_NOTDEFINED_VECTOR
 #endif
 
+#if !defined(KOKKOS_ENABLE_DEPRECATED_CODE_4) || \
+    defined(KOKKOS_ENABLE_DEPRECATION_WARNINGS)
+namespace {
+[[deprecated("Deprecated <Kokkos_Vector.hpp> header is included")]] int
+emit_warning_kokkos_vector_deprecated() {
+  return 0;
+}
+static auto do_not_include = emit_warning_kokkos_vector_deprecated();
+}  // namespace
+#endif
+
 #include <Kokkos_Core_fwd.hpp>
 #include <Kokkos_DualView.hpp>
 
--- a/containers/src/Kokkos_Vector.hpp
+++ b/containers/src/Kokkos_Vector.hpp
@@ -21,6 +21,8 @@
 #define KOKKOS_IMPL_PUBLIC_INCLUDE_NOTDEFINED_VECTOR
 #endif
 
+#include <Kokkos_Macros.hpp>
+
 #if !defined(KOKKOS_ENABLE_DEPRECATED_CODE_4) || \
     defined(KOKKOS_ENABLE_DEPRECATION_WARNINGS)
 namespace {
--- a/containers/src/Kokkos_Vector.hpp
+++ b/containers/src/Kokkos_Vector.hpp
@@ -23,8 +23,8 @@
 
 #include <Kokkos_Macros.hpp>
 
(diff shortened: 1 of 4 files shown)


fnrizzi:
some jenkins are failing bc DKokkos_ENABLE_DEPRECATED_CODE_4 is on, so this PR triggers the expected warning. 
the ci failing is a good thing for this PR, isn't it? but seems to me we won't have all test passing for this PR anyway right?

fnrizzi (approved):
caveat the CI comments to be discussed

containers/src/Kokkos_Vector.hpp
```diff
@@ -21,6 +21,17 @@
...
+#if !defined(KOKKOS_ENABLE_DEPRECATED_CODE_4) || \
+    defined(KOKKOS_ENABLE_DEPRECATION_WARNINGS)
+namespace {
+[[deprecated("Deprecated <Kokkos_Vector.hpp> header is included")]] int
+emit_warning_kokkos_vector_deprecated() {
+  return 0;
+}
+static auto do_not_include = emit_warning_kokkos_vector_deprecated();
+}  // namespace
+#endif
```
masterleinad, line 33:
```suggestion
#include <Kokkos_Macros.hpp>

#if defined(KOKKOS_ENABLE_DEPRECATED_CODE_4)
namespace {
KOKKOS_DEPRECATED_WITH_COMMENT("Deprecated <Kokkos_Vector.hpp> header is included") int
emit_warning_kokkos_vector_deprecated() {
  return 0;
}
static auto do_not_include = emit_warning_kokkos_vector_deprecated();
}  // namespace
#else
#error "Deprecated <Kokkos_Vector.hpp> header is included"
#endif
```

The 
```
#include <Kokkos_Macros.hpp>
```
is crucial for fixing CI.

dalg24: Please go ahead and push to my branch

masterleinad:
We confirmed in the meeting that we want to deprecate it.

masterleinad (approved):

dalg24:
There are a couple unrelated failures (machine issues)

containers/src/Kokkos_Vector.hpp
```diff
@@ -21,6 +21,22 @@
 #define KOKKOS_IMPL_PUBLIC_INCLUDE_NOTDEFINED_VECTOR
 #endif
 
+#include <Kokkos_Macros.hpp>
+
+#if defined(KOKKOS_ENABLE_DEPRECATED_CODE_4)
+#if defined(KOKKOS_ENABLE_DEPRECATION_WARNINGS)
+namespace {
+[[deprecated("Deprecated <Kokkos_Vector.hpp> header is included")]] int
```
dalg24, line 29: @masterleinad why did you do this rather than `KOKKOS_DEPRECATED_WITH_COMMENT`?

masterleinad: What I cared about is using "#error" and including "Kokkos_Macros.hpp" but not so much about using `KOKKOS_DEPRECATED_WITH_COMMENT` as opposed to `KOKKOS_ENABLE_DEPRECATION_WARNINGS`. The way now at least only declares extra symbols when they are needed for the deprecation warning.

dalg24: Good enough.

To be pedantic, there will be no symbol, that is why I used an anonymous namespace.

dalg24:
Two failures look unrelated.  One is lost connection with the runner, this other one is failure to start the container.

PR #6256, opened 2023-07-04
team-level std algos: part 10

fnrizzi:
Co-authored-by: Cezary Skrzyński
Co-authored-by: Jakub Strzebonski

## Content 

Team-level implementation and tests for:

- `Kokkos_IsSortedUntil.hpp` 
- `Kokkos_IsSorted.hpp` 
- `Kokkos_IsPartitioned.hpp` 
- `Kokkos_PartitionCopy.hpp` 
- `Kokkos_PartitionPoint.hpp`

## Technical comments

- the change `class` -> `typename` and the enable if defaulted in the template parameters comes from [this request](https://github.com/kokkos/kokkos/pull/6200#issuecomment-1626406816)

- the code duplication is intentional and agreed upon after the long discussion in [here](https://github.com/kokkos/kokkos/pull/6200)

- tests are run for both contiguous views and strided views

## Other comments

- all team-level PRs are tracked in: https://github.com/kokkos/kokkos/issues/6352

- this comes from breaking apart https://github.com/kokkos/kokkos/pull/5150

## Possible conflicts

Implementation-wise, this has not conflict with other team-level PRs so it could be merged independently. 
However, a trivial conflict in the CMakeLists inside the unit test occurs if another of the team-level PRs is merged before this because each team-level PR adds test using a specific "label". Resolving this is trivial. 

## Associated wiki PRs

- `Kokkos_IsSortedUntil.hpp`: see [wiki PR](https://github.com/kokkos/kokkos-core-wiki/pull/436)
- `Kokkos_IsSorted.hpp`, `Kokkos_IsPartitioned.hpp`, `Kokkos_PartitionCopy.hpp`: see [wiki PR](https://github.com/kokkos/kokkos-core-wiki/pull/403) 
- `Kokkos_PartitionPoint.hpp`: see [wiki PR](https://github.com/kokkos/kokkos-core-wiki/pull/440) 

IMPORTANT: wiki PRs are not complete yet but 99% there 

The issue tracking all *wiki PRs* is [here](https://github.com/kokkos/kokkos-core-wiki/issues/379)

PR #6257, opened 2023-07-04
std_algos: fix wrong corner case for `is_partitioned`

fnrizzi:
Fix wrong behavior for corner case in `Kokkos::Experimental::is_partitioned`. 

Currently, if we were to do: 
```cpp
template <class ValueType>
struct IsNegativeFunctor {
  KOKKOS_INLINE_FUNCTION
  bool operator()(const ValueType val) const { return (val < 0); }
};

Kokkos::View<int*> v = {11, 3, 17, 9, 3};
auto res = KE::is_partitioned(exespace(), KE::cbegin(v), KE::cend(v), p);
```
we get `res == false`. 
However, this is wrong wrt std library [see godbolt here](https://godbolt.org/z/3rao6GW3K) and should return `true`.
This PR fixes this and the test that was wrongly set to false. 

Note that the std library documentation `https://en.cppreference.com/w/cpp/algorithm/is_partitioned` is not super clear about this, in the sense that when none of the elements satisfy the predicate one could misinterpret it.

Comment: this bug was found when implementing the team-level api in #6256

### DIFF
--- a/algorithms/src/std_algorithms/impl/Kokkos_IsPartitioned.hpp
+++ b/algorithms/src/std_algorithms/impl/Kokkos_IsPartitioned.hpp
@@ -107,14 +107,27 @@ bool is_partitioned_impl(const std::string& label, const ExecutionSpace& ex,
   constexpr index_type red_id_max =
       ::Kokkos::reduction_identity<index_type>::max();
 
+  // clang-format off
   if (red_result.max_loc_true != red_id_max &&
-      red_result.min_loc_false != red_id_min) {
+      red_result.min_loc_false != red_id_min){
+    // this occurs when the reduction yields nontrivial values
     return red_result.max_loc_true < red_result.min_loc_false;
-  } else if (first + red_result.max_loc_true == --last) {
+  }
+  else if (red_result.max_loc_true == red_id_max &&
+	   red_result.min_loc_false == 0){
+    // this occurs when all values do NOT satisfy
+    // the predicate, and this corner case should also be true
     return true;
-  } else {
+  }
+  else if (first + red_result.max_loc_true == --last) {
+    // this occurs when all values DO satisfy the predicate,
+    // this corner case should also be true
+    return true;
+  }
+  else {
     return false;
   }
+  // clang-format on
 }
 
 }  // namespace Impl
--- a/algorithms/src/std_algorithms/impl/Kokkos_IsPartitioned.hpp
+++ b/algorithms/src/std_algorithms/impl/Kokkos_IsPartitioned.hpp
@@ -107,27 +107,22 @@ bool is_partitioned_impl(const std::string& label, const ExecutionSpace& ex,
   constexpr index_type red_id_max =
       ::Kokkos::reduction_identity<index_type>::max();
 
-  // clang-format off
   if (red_result.max_loc_true != red_id_max &&
-      red_result.min_loc_false != red_id_min){
+      red_result.min_loc_false != red_id_min) {
     // this occurs when the reduction yields nontrivial values
     return red_result.max_loc_true < red_result.min_loc_false;
-  }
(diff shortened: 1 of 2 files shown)


algorithms/src/std_algorithms/impl/Kokkos_IsPartitioned.hpp
```diff
@@ -107,14 +107,27 @@ bool is_partitioned_impl(const std::string& label, const ExecutionSpace& ex,
   constexpr index_type red_id_max =
       ::Kokkos::reduction_identity<index_type>::max();
 
+  // clang-format off
```
dalg24, line 110: Why disable the format here?

dalg24 (commented):
OK with the changes but not with the format disable

algorithms/src/std_algorithms/impl/Kokkos_IsPartitioned.hpp
```diff
@@ -107,14 +107,27 @@ bool is_partitioned_impl(const std::string& label, const ExecutionSpace& ex,
...
+    // the predicate, and this corner case should also be true
+    return true;
+  }
+  else if (first + red_result.max_loc_true == --last) {
+    // this occurs when all values DO satisfy the predicate,
+    // this corner case should also be true
     return true;
-  } else {
+  }
+  else {
```
dalg24, line 127: This
```suggestion
  } else {
```

algorithms/src/std_algorithms/impl/Kokkos_IsPartitioned.hpp
```diff
@@ -107,14 +107,27 @@ bool is_partitioned_impl(const std::string& label, const ExecutionSpace& ex,
   constexpr index_type red_id_max =
       ::Kokkos::reduction_identity<index_type>::max();
 
+  // clang-format off
   if (red_result.max_loc_true != red_id_max &&
-      red_result.min_loc_false != red_id_min) {
+      red_result.min_loc_false != red_id_min){
```
dalg24, line 112: and this
```suggestion
      red_result.min_loc_false != red_id_min) {
```

algorithms/src/std_algorithms/impl/Kokkos_IsPartitioned.hpp
fnrizzi: readability (IMO)

dalg24 (commented):
The disable format introduces inconsistencies (that look accidental) and it is not obvious to me what benefit we get from disabling.

algorithms/src/std_algorithms/impl/Kokkos_IsPartitioned.hpp
fnrizzi: i will revert

dalg24: You can post the diff to try to convince me or see if others will agree with you.
I will not block but would only apply at this time if the format disable is dropped.

cz4rs (approved):
Looks good to me.

algorithms/src/std_algorithms/impl/Kokkos_IsPartitioned.hpp
dalg24: I am not trying to be difficult. I just really cannot tell what was better with disabling the format.

dalg24 (approved):

algorithms/src/std_algorithms/impl/Kokkos_IsPartitioned.hpp
fnrizzi: To me it was easier to glance and see the various ifs/elses without having like 10 lines all condensed but was just a personal preference I guess :) 

Really no big deal to revert it back

Rombur:
Retest this please

PR #6265, opened 2023-07-06
Add default ParallelFor copy constructor for HIP

Rombur:
With ROCm 5.6 and the Cray compiler, we get a lot of warnings because we are using an implicitly define copy constructor. This PR explicitly create the copy constructor. The copy constructor is used  when using local memory [here](https://github.com/kokkos/kokkos/blob/f050fbc66dd15410f26fda1db2a1678075fc9fda/core/src/HIP/Kokkos_HIP_KernelLaunch.hpp#L68-L80).

This PR also makes the deleted default constructor and assignment operator public. They were private before which is not necessary since they are deleted.

### DIFF
--- a/core/src/HIP/Kokkos_HIP_Parallel_MDRange.hpp
+++ b/core/src/HIP/Kokkos_HIP_Parallel_MDRange.hpp
@@ -40,10 +40,11 @@ class ParallelFor<FunctorType, Kokkos::MDRangePolicy<Traits...>, HIP> {
   const FunctorType m_functor;
   const Policy m_policy;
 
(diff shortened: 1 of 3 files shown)


rgayatri23:
Less warnings is good! 
Was this not an issue before rocm/5.6 ?

Rombur:
> Was this not an issue before rocm/5.6 ?

No. I think it's the underlying clang that started to emit the warning and that's why both amdclang and cray clang now emit a warning

rgayatri23 (approved):

dalg24:
ignoring timeout for one of the HIP builds

PR #6280, opened 2023-07-14
Added CTAD and to_Array for Kokkos::Array

nliber:
Added the deduction guide for Kokkos::Array (similar to the C++17 one added to `std::array`) and `Kokkos::to_Array` free functions that convert to `Kokkos::Array` (similar to the C++20 `std::to_array` converting to `std::array`).

Note:  While I played around with also supporting the `Proxy` template parameter to `Kokkos::Array`, but I did not come up with a nice way to make it useful.


core/unit_test/TestArray.cpp
```diff
@@ -49,4 +49,49 @@ KOKKOS_FUNCTION constexpr bool test_array_structured_binding_support() {
...
+    // implicit conversion happens
+    auto a3 = Kokkos::to_Array<long>({0, 1, 3});
+    static_assert(std::is_same_v<decltype(a3), Kokkos::Array<long, 3>>);
+
+    auto a4 = Kokkos::to_Array<std::pair<int, float>>(
+        {{3, 0.0f}, {4, 0.1f}, {4, 0.1e23f}});
+    static_assert(a4.size() == 3);
+
+    return true;
+}
```
crtrott, line 93: are we sure these guys get instantiated on device?

crtrott (changes requested):
Need to make sure we test this on device.

core/unit_test/TestArray.cpp
nliber: Nothing in `TestArray.cpp` currently gets tested outside of compile time as it is in the `COMPILE_ONLY_SOURCES` section.  There are some minor tests for `Array` in `TestAggregate.hpp`.  Should I move those into a `TestArrayOps.hpp` and expand on them to cover `to_Array`?  Adding more unit tests for `Array` is also doable, but I'd rather do that as part of a separate PR.

nliber:
I think this is an ICPC bug...

> /__w/kokkos/kokkos/core/unit_test/TestArray.cpp(95): error: expression must have a constant value
>   static_assert(test_to_array());
>                 ^
> /__w/kokkos/kokkos/core/src/Kokkos_Array.hpp(337): note: attempt to access expired storage
>     return {{std::move(a[I])...}};
>              ^
> /__w/kokkos/kokkos/core/src/Kokkos_Array.hpp(349): note: called from:
>     return Impl::to_Array_impl(std::move(a), std::make_index_sequence<N>{});
>                               ^
> /__w/kokkos/kokkos/core/unit_test/TestArray.cpp(88): note: called from:
>       auto a4 = Kokkos::to_Array<std::pair<int, float>>(
>

core/unit_test/TestArray.cpp
nliber: Added some runtime tests.

(There is still a possible ICPC issue, so this is still a draft)

nliber:
Closing this in favor of #6356, #6372, #6373 and #6375.

Issue #6355 is the more complete description of what needs to be done.

PR #6282, opened 2023-07-14
make Kokkos_CXX_COMPILER_VERSION available to CMake consumers

cwpearson:
This might be generally useful to people, but the specific need is that a new performance test in Kokkos Kernels uses `std::filesystem`, and the rocm 5.2 compiler has that in a different library and `experimental` namespace, so we need to do some different configuration based on the compiler version.


crtrott (approved):
I don't see a good reason not to export KOKKOS_CXX_COMPILER_VERSION, if we export KOKKOS_CXX_COMPILER_ID too. I could see an argument for ONLY exporting Kokkos_CXX_COMPILER to enable check for using the same compiler. But in fact even checking for same compiler to be rigorous probably should check for same compiler version too.

dalg24 (approved):

dalg24:
Ignoring failure to start of one of the CUDA build

dalg24:
Make sure you add to the changelog

cwpearson:
Done, thanks for the reminder.

PR #6290, opened 2023-07-17
Fix a memory bug in the `free_state` function of random pools.

### LINKED ISSUE #6140: Kokkos_Random: Random number sequence seems to get repeated for different threads on CUDA
I can't provide a fully reproducible minimal code at this moment, but here's the scenario: I am trying to initialize say a million particles and set their properties by generating e.g. 7 different doubles for each particle. 

```
Kokkos::Random_XorShift64_Pool<Kokkos::CudaSpace> random_pool(12345);

// 2000 teams, 512 threads per team
Kokkos::parallel_for(teamPolicy,
        KOKKOS_LAMBDA(const TeamHandleType& team)
        {
            unsigned numElems = 512;
            Kokkos::parallel_for(Kokkos::TeamThreadRange(team, 0u, numElems), [&] (const int& i)
            {
                Kokkos::Random_XorShift64<stk::ngp::MemSpace> gen = random_pool.get_state();

                const double x1 = gen.drand(0.0, 1.0);
                const double x2 = gen.drand(0.0, 1.0);
                ...
                const double x7 = gen.drand(0.0, 1.0);

                random_pool.free_state(gen);
           }
     }
```

Sometimes, I would get two particles with exactly same values for all randomly generated properties. It doesn't happen very often- once in every 3 or 4 runs of simulation maybe, and I have only seen it when dealing with around 1million or higher number of particles.

Edit: I do have an unique id for each particle that can be used as a seed to generate the sequences independently for each particle. Is there any Kokkos API that allows that? or any plan to do this in future?

**Details**
1. Modules: `GCC/10.2.0 CUDA/11.2.1 OpenMPI/4.0.5 CMake/3.23.1`
2. Kokkos release: 4.0.1
3. platform: CUDA and linux.
4. cmake configure command:
```
cmake ${ROOT_DIR} \
    -DCMAKE_CXX_COMPILER="${Trilinos_Source_DIR}/packages/kokkos/bin/nvcc_wrapper" \
    -DCMAKE_CXX_FLAGS="-O0 -lmpi" \
    -DTrilinos_DIR:PATH=${Trilinos_Install_DIR} \
    -DKokkos_ROOT:PATH="${Trilinos_Install_DIR}" \
    -DCMAKE_BUILD_TYPE:STRING=Debug \
    -DKokkos_ENABLE_CUDA=ON
```

Shihab-Shahriar:
Fixes #6140. 

The `free_state` function was missing a memory fence between two of it's global memory operations.

```
  KOKKOS_INLINE_FUNCTION
  void free_state(const Random_XorShift64<DeviceType>& state) const {
    state_(state.state_idx_, 0) = state.state_;
    locks_(state.state_idx_, 0) = 0;
  }
```

Since not all devices guarantee order in global memory operations, without a fence, the lock could get released before the `state_` array has been updated in global memory.


masterleinad:
Is there any chance that we can get a unit test for this?

Shihab-Shahriar:
I added an unit test. In summary, this is how it works:

- Since this bug gets triggered only when states get reused by next thread with similar ID, the number of streams (`n_streams`) should be significantly larger than number of states (which equals execution space's `concurrency()`).  I used `n_streams = ExecutionSpace{}.concurrency() * 4`.
- Generate 8 samples for each stream on device, copy back to a `(n_streams ,8)` shaped 2d array on host.
- To quickly find the duplicates- sort all the streams based on first number, if that's equal then second number and so on. (there's some room for optimization here- we don't need to do entire sort if a duplicate is found. But I opted for simplicity.)
- Check each pair of neighbors in the sorted order for duplicates.

I have been able to reliably reproduce the bug for 64 bit random pool in my device (Nvidia V100), but not for 1024 bit one. I think this is because as the pool has 16 states that get updated in `free_state`, the probability of at least one of them reaching global memory before before lock gets released is much higher. 

I also want to note that the test has a narrow focus of reproducing the issue reported in #6140 i.e. checking for exact duplicates. Statistical tests that check for independence of the streams [1] can potentially identify the issue with 1024 bit pool without memory fences. 



_[1] Salmon, John K., Mark A. Moraes, Ron O. Dror, and David E. Shaw. "Parallel random numbers: as easy as 1, 2, 3." In Proceedings of 2011 international conference for high performance computing, networking, storage and analysis, pp. 1-12. 2011._

masterleinad:
PLease fix the indentation (according to `clang-format-8`):
```diff
+ ./scripts/docker/check_format_cpp.sh
diff --git a/algorithms/src/Kokkos_Random.hpp b/algorithms/src/Kokkos_Random.hpp
index aada5fe27..2d7d236d2 100644
--- a/algorithms/src/Kokkos_Random.hpp
+++ b/algorithms/src/Kokkos_Random.hpp
@@ -1210,7 +1210,7 @@ class Random_XorShift1024_Pool {
   KOKKOS_INLINE_FUNCTION
   void free_state(const Random_XorShift1024<DeviceType>& state) const {
     for (int i = 0; i < 16; i++) state_(state.state_idx_, i) = state.state_[i];
-    p_(state.state_idx_, 0)     = state.p_;
+    p_(state.state_idx_, 0) = state.p_;
     // Release the lock only after the state has been updated in memory
     Kokkos::memory_fence();
     locks_(state.state_idx_, 0) = 0;
diff --git a/algorithms/unit_tests/TestRandom.hpp b/algorithms/unit_tests/TestRandom.hpp
index 7a31411b9..fae2f029d 100644
--- a/algorithms/unit_tests/TestRandom.hpp
+++ b/algorithms/unit_tests/TestRandom.hpp
@@ -480,8 +480,7 @@ struct generate_random_stream {
   GeneratorPool rand_pool;
   int samples;
 
-  generate_random_stream(ViewType vals_, GeneratorPool rand_pool_,
-                  int samples_)
+  generate_random_stream(ViewType vals_, GeneratorPool rand_pool_, int samples_)
       : vals(vals_), rand_pool(rand_pool_), samples(samples_) {}
 
   KOKKOS_INLINE_FUNCTION
@@ -494,50 +493,48 @@ struct generate_random_stream {
   }
 };
 
-
 // NOTE: this doesn't test the statistical independence of multiple streams
-// generated by a Random pool, it only tests for complete duplicates. 
+// generated by a Random pool, it only tests for complete duplicates.
 template <class ExecutionSpace, class Pool>
-void test_duplicate_stream(){
-  using ViewType = Kokkos::View<uint64_t**, ExecutionSpace>; 
+void test_duplicate_stream() {
+  using ViewType = Kokkos::View<uint64_t**, ExecutionSpace>;
 
   // Heuristic to create a "large enough" number of streams.
   int n_streams = ExecutionSpace{}.concurrency() * 4;
-  int samples = 8;
+  int samples   = 8;
 
   Pool rand_pool(42);
   ViewType vals_d("Vals", n_streams, samples);
   typename ViewType::HostMirror vals_h = Kokkos::create_mirror_view(vals_d);
 
-  Kokkos::parallel_for(n_streams,generate_random_stream<ExecutionSpace, Pool>(
-                          vals_d, rand_pool, samples));
+  Kokkos::parallel_for(n_streams, generate_random_stream<ExecutionSpace, Pool>(
+                                      vals_d, rand_pool, samples));
 
   Kokkos::fence();
   Kokkos::deep_copy(vals_h, vals_d);
 
   /*
   To quickly find streams that are identical, we sort them by the first number,
-  if that's equal then the second and so on. We then test each neighbor pair 
+  if that's equal then the second and so on. We then test each neighbor pair
   for duplicates.
-  */ 
+  */
   std::vector<size_t> indices(n_streams);
   std::iota(indices.begin(), indices.end(), 0);
 
-  auto comparator = [&](int i, int j){
-                  for (int k=0; k<samples; k++){
-                    if(vals_h(i, k)!=vals_h(j, k)) 
-                      return vals_h(i, k) < vals_h(j, k);
-                  }
-                  return false;
-              };
+  auto comparator = [&](int i, int j) {
+    for (int k = 0; k < samples; k++) {
+      if (vals_h(i, k) != vals_h(j, k)) return vals_h(i, k) < vals_h(j, k);
+    }
+    return false;
+  };
   std::sort(indices.begin(), indices.end(), comparator);
 
-  for(int i=0; i< n_streams-1; i++){
-    int idx1 = indices[i], idx2 = indices[i+1];
+  for (int i = 0; i < n_streams - 1; i++) {
+    int idx1 = indices[i], idx2 = indices[i + 1];
 
     int k = 0;
-    while(k < samples && vals_h(idx1, k)==vals_h(idx2,k)) k++;
-    ASSERT_LT(k, samples)  << "Duplicate streams found";
+    while (k < samples && vals_h(idx1, k) == vals_h(idx2, k)) k++;
+    ASSERT_LT(k, samples) << "Duplicate streams found";
   }
 }
 
@@ -583,8 +580,8 @@ TEST(TEST_CATEGORY, Random_XorShift1024_0) {
 
 TEST(TEST_CATEGORY, Multi_streams) {
   using ExecutionSpace = TEST_EXECSPACE;
-  using Pool64 = Kokkos::Random_XorShift64_Pool<ExecutionSpace>;
-  using Pool1024 = Kokkos::Random_XorShift1024_Pool<ExecutionSpace>;
+  using Pool64         = Kokkos::Random_XorShift64_Pool<ExecutionSpace>;
+  using Pool1024       = Kokkos::Random_XorShift1024_Pool<ExecutionSpace>;
 
   AlgoRandomImpl::test_duplicate_stream<ExecutionSpace, Pool64>();
   AlgoRandomImpl::test_duplicate_stream<ExecutionSpace, Pool1024>();
```

masterleinad (approved):

algorithms/unit_tests/TestRandom.hpp
```diff
@@ -469,6 +472,72 @@ struct TestDynRankView {
...
+
+  // Heuristic to create a "large enough" number of streams.
+  int n_streams = ExecutionSpace{}.concurrency() * 4;
+  int samples   = 8;
+
+  Pool rand_pool(42);
+  ViewType vals_d("Vals", n_streams, samples);
+  typename ViewType::HostMirror vals_h = Kokkos::create_mirror_view(vals_d);
+
+  Kokkos::parallel_for(n_streams, generate_random_stream<ExecutionSpace, Pool>(
```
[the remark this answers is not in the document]
dalg24: Also how is that not a memory access violation because the device view does get the execution space template argument.

algorithms/unit_tests/TestRandom.hpp
```diff
@@ -469,6 +472,72 @@ struct TestDynRankView {
...
+  auto comparator = [&](int i, int j) {
+    for (int k = 0; k < samples; k++) {
+      if (vals_h(i, k) != vals_h(j, k)) return vals_h(i, k) < vals_h(j, k);
+    }
+    return false;
+  };
+  std::sort(indices.begin(), indices.end(), comparator);
+
+  for (int i = 0; i < n_streams - 1; i++) {
+    int idx1 = indices[i], idx2 = indices[i + 1];
```
dalg24, line 533:
```suggestion
    int idx1 = indices[I];
    int idx2 = indices[i + 1];
```

dalg24 (approved):

dalg24:
SYCL failure is unrelated but OpenMPTarget has an issue
```
29: [ RUN      ] openmptarget.Multi_streams
29: CUDA error: an illegal memory access was encountered 
29: Libomptarget error: Call to targetDataEnd failed, abort target.
29: Libomptarget error: Failed to process data after launching the kernel.
29: Libomptarget error: Consult https://openmp.llvm.org/design/Runtimes.html for debugging options.
29: Kokkos_OpenMPTarget_ParallelFor_Range.hpp:54:1: Libomptarget fatal error 1: failure of target construct while offloading is mandatory
```
we could disable if we can't figure it out.  @rgayatri23 please advise.

rgayatri23:
> SYCL failure is unrelated but OpenMPTarget has an issue
> 
> ```
> 29: [ RUN      ] openmptarget.Multi_streams
> 29: CUDA error: an illegal memory access was encountered 
> 29: Libomptarget error: Call to targetDataEnd failed, abort target.
> 29: Libomptarget error: Failed to process data after launching the kernel.
> 29: Libomptarget error: Consult https://openmp.llvm.org/design/Runtimes.html for debugging options.
> 29: Kokkos_OpenMPTarget_ParallelFor_Range.hpp:54:1: Libomptarget fatal error 1: failure of target construct while offloading is mandatory
> ```
> 
> we could disable if we can't figure it out. @rgayatri23 please advise.

Yeah this is very likely a compiler bug. Can we disable it for OpenMPTarget for now? Once it is merged, I will go line-by-line to see what triggers it and take the necessary action.

algorithms/unit_tests/TestRandom.hpp
```diff
@@ -509,5 +579,14 @@ TEST(TEST_CATEGORY, Random_XorShift1024_0) {
       .run();
 }
 
+TEST(TEST_CATEGORY, Multi_streams) {
```
dalg24, line 582: (not formatted)
please add something like
```suggestion
TEST(TEST_CATEGORY, Multi_streams) {
#ifdef KOKKOS_ENABLE_OPENMPTARGET
    if constexpr (std::is_same_v<ExecutionSpace, \
                                 Kokkos::Experimental::OpenMPTarget>) {
      GTEST_SKIP() << "Libomptarget error"; // FIXME_OPENMPTARGET
    }
#endif
```

```

Comment (dalg24) [reply algorithms/unit_tests/TestRandom.hpp:582]: Make sure `ExecutionSpace` is defined before this block or use directly `TEST_EXECSPACE`.

**Comment by dalg24** [comment]
Uggg of course now SYCL fails
```
39: [ RUN      ] sycl.Multi_streams
39: /var/jenkins/workspace/Kokkos/algorithms/unit_tests/TestRandom.hpp:538: Failure
39: Expected: (k) < (samples), actual: 8 vs 8
39: Duplicate streams found
```
@masterleinad please advise

masterleinad:
> @masterleinad please advise

It works on `Intel GPUs` so we might just disable the test for `SYCL` with a FIXME if `KOKKOS_ARCH_INTEL_GPU` is not defined.

dalg24:
> > @masterleinad please advise
> 
> It works on `Intel GPUs` so we might just disable the test for `SYCL` with a FIXME if `KOKKOS_ARCH_INTEL_GPU` is not defined.

What about
```C++
#if defined(KOKKOS_ENABLE_SYCL) && defined(KOKKOS_IMPL_ARCH_NVIDIA_GPU)
  if constexpr (std::is_same_v<ExecutionSpace,
                               Kokkos::Experimental::SYCLt>) {
    GTEST_SKIP() << "Failing on NVIDIA GPUs";  // FIXME_SYCL
  }
#endif

masterleinad:
Sure, we can also just disable for SYCL+CUDA. It boils down to the same restriction anyway for now.

dalg24:
Ignoring HIP builds that timed out.

Thank you for fixing this!

PR #6292, opened 2023-07-18
#5635: Serial/OpenMP: Parallel_scan with return value for ThreadVectorRange

thearusable:
Related to #5635

### DIFF
--- a/core/src/impl/Kokkos_HostThreadTeam.hpp
+++ b/core/src/impl/Kokkos_HostThreadTeam.hpp
@@ -893,16 +893,21 @@ KOKKOS_INLINE_FUNCTION
   }
 }
 
-template <typename iType, class ClosureType, class Member>
+template <typename iType, class ClosureType, class Member, typename ValueType>
 KOKKOS_INLINE_FUNCTION
-    std::enable_if_t<Impl::is_host_thread_team_member<Member>::value>
+    std::enable_if_t<!Kokkos::is_reducer<ValueType>::value &&
+                     Impl::is_host_thread_team_member<Member>::value>
     parallel_scan(Impl::ThreadVectorRangeBoundariesStruct<iType, Member> const&
                       loop_boundaries,
-                  ClosureType const& closure) {
-  using value_type = typename Kokkos::Impl::FunctorAnalysis<
-      Impl::FunctorPatternInterface::SCAN, void, ClosureType, void>::value_type;
+                  ClosureType const& closure, ValueType& return_val) {
+  // Extract ValueType from the Closure
+  using ClosureValueType = typename Kokkos::Impl::FunctorAnalysis<
+      Kokkos::Impl::FunctorPatternInterface::SCAN, void, ClosureType,
+      void>::value_type;
+  static_assert(std::is_same<ClosureValueType, ValueType>::value,
+                "Non-matching value types of closure and return type");
 
-  value_type scan_val = value_type();
+  ValueType scan_val = ValueType();
 
 #ifdef KOKKOS_ENABLE_PRAGMA_IVDEP
 #pragma ivdep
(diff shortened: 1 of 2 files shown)


core/src/impl/Kokkos_HostThreadTeam.hpp
```diff
@@ -893,16 +893,21 @@ KOKKOS_INLINE_FUNCTION
...
+                  ClosureType const& closure, ValueType& return_val) {
+  // Extract ValueType from the Closure
+  using ClosureValueType = typename Kokkos::Impl::FunctorAnalysis<
+      Kokkos::Impl::FunctorPatternInterface::SCAN, void, ClosureType,
+      void>::value_type;
+  static_assert(std::is_same<ClosureValueType, ValueType>::value,
+                "Non-matching value types of closure and return type");
 
-  value_type scan_val = value_type();
+  ValueType scan_val;
```
masterleinad, line 910: This actually needs to be zero-initialized, right?

thearusable: Yes, I was removing the initial assignments and I removed to much. Will fix in the moment

thearusable: Fixed

fnrizzi (approved):

core/unit_test/TestTeamVector.hpp
```diff
@@ -605,6 +605,71 @@ struct functor_vec_scan {
   }
 };
 
+// Temporary: This condition will progressively be reduced when parallel_scan
+// with return value will be implemented for more backends.
+#if defined(KOKKOS_ENABLE_SERIAL) || defined(KOKKOS_ENABLE_OPENMP)
+#if !defined(KOKKOS_ENABLE_CUDA) && !defined(KOKKOS_ENABLE_HIP) &&             \
+    !defined(KOKKOS_ENABLE_OPENACC) && !defined(KOKKOS_ENABLE_SYCL) &&         \
+    !defined(KOKKOS_ENABLE_THREADS) && !defined(KOKKOS_ENABLE_OPENMPTARGET) && \
+    !defined(KOKKOS_ENABLE_HPX)
```
dalg24, line 614: Why are we not testing if for instance OpenMP+CUDA is enabled?
We can easily do something like
```C++
#ifdef KOKKOS_ENABLE_OPENMP
if constexpr (std::is_same_v<TEST_EXECSPACE, OpenMP>)
#endif
#ifdef KOKKOS_ENABLE_SERIAL
if constexpr (std::is_same_v<TEST_EXECSPACE, Serial>)
#endif
ASSERT_TRUE((TestTeamVector::Test<TEST_EXECSPACE>(12)));

Comment (crtrott) [reply core/unit_test/TestTeamVector.hpp:614]: I think all the PRs are ready to introduce this for all backends? So it doesn't seem to matter too much?

File: core/unit_test/TestTeamVector.hpp [hunk 3]
```diff
@@ -605,6 +605,71 @@ struct functor_vec_scan {
...
+template <typename Scalar, class ExecutionSpace>
+struct functor_vec_scan_ret_val {
+  using policy_type     = Kokkos::TeamPolicy<ExecutionSpace>;
+  using execution_space = ExecutionSpace;
+  using view_type       = Kokkos::View<Scalar **, execution_space>;
+
+  Kokkos::View<int, Kokkos::LayoutLeft, ExecutionSpace> flag;
```
dalg24, line 621:
```suggestion
  Kokkos::View<int, ExecutionSpace> flag;
```

core/unit_test/TestTeamVector.hpp
```diff
@@ -605,6 +605,71 @@ struct functor_vec_scan {
...
+  view_type return_values;
+  int team_size;
+
+  functor_vec_scan_ret_val(
+      Kokkos::View<int, Kokkos::LayoutLeft, ExecutionSpace> flag_, int nteams,
```
dalg24, line 626:
```suggestion
      Kokkos::View<int, ExecutionSpace> flag_, int nteams,
```

core/unit_test/TestTeamVector.hpp
```diff
@@ -605,6 +605,71 @@ struct functor_vec_scan {
...
+
+            if (test != val) {
+              Kokkos::printf("FAILED vector_par_scan %i %i %lf %lf\n",
+                             team.league_rank(), team.team_rank(),
+                             static_cast<double>(test),
+                             static_cast<double>(val));
+
+              flag() = 1;
+            }
+          }
```
dalg24, line 654: Mixing computation and checking the result is not good practice.

fnrizzi: btw this is following the similar approach used in the other tests here, so if you want to change the overall approach maybe we can do that in a separate PR?

core/unit_test/TestTeamVector.hpp
fnrizzi: fixed

core/unit_test/TestTeamVector.hpp
fnrizzi: fixed

crtrott (approved):

PR #6272, opened 2023-07-10
Improve SYCL TeamPolicy reduction

masterleinad:
Part of #6035.
This limits the number of workgroups, effectively processing multiple work items per thread similar to what we are doing for `Cuda` and `HIP`.

PR #6293, opened 2023-07-18
SYCL TeamPolicy: Fix sign comparison warning

masterleinad:
We got a little unlucky that the combination of #6272 and #6270 resulted in a sign comparison warning that we didn't notice for the individual pull requests. It seems that we don't test array reductions for TeamPolicy in the CI.

The change matches what we do for the other (shuffle-based) code path.

### DIFF
--- a/core/src/SYCL/Kokkos_SYCL_Parallel_Team.hpp
+++ b/core/src/SYCL/Kokkos_SYCL_Parallel_Team.hpp
@@ -681,7 +681,7 @@ class ParallelReduce<CombinedFunctorReducerType,
                 if constexpr (!use_shuffle_based_algorithm<ReducerType>) {
                   reference_type update =
                       reducer.init(&local_mem[local_id * value_count]);
-                  for (size_t league_rank = group_id; league_rank < league_size;
+                  for (int league_rank = group_id; league_rank < league_size;
                        league_rank += n_wgroups) {
                     const member_type team_member(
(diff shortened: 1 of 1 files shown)


dalg24 (commented):
Where did you see that warning?

> It seems that we don't test array reductions for TeamPolicy in the CI.

Not sure what you are saying.  Are we not seeing the warning because the code path is not being compiled?

fnrizzi:
I see it here: https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/Kokkos/detail/Kokkos/13656/pipeline/

fnrizzi (approved):

dalg24 (approved):
Approving since as Francesco pointed out the warning shows in the CI in other builds but still do not understand the

> It seems that we don't test array reductions for TeamPolicy in the CI.

Was that a drive-by note?

dalg24:
Ignoring failure to start the container in one CUDA build and
```
4:   sum1
4:     Which is: 0
4:   sum2
4:     Which is: 7642095
4: [  FAILED  ] openmptarget.partitioning_by_args (1 ms)

masterleinad:
> Was that a drive-by note?

We would have caught the warning in the TeamPolicy pull request if array reductions were tested for TeamPolicy.
With the workgroup reduction pull request, we always take the non-shuffle (local memory) reduction code path we were previously only using for array reductions.

PR #6299, opened 2023-07-21
Create cudaAPI function wrappers

tcclevenger:
An alternative to https://github.com/kokkos/kokkos/pull/5989. Here I use individual wrappers for each cudaAPI call.

Benefits
- Much easier to read where wrapper is called (vs. function ptrs)
- Reading runtime errors give location of call with name matching cudaAPI function name (difference only in case)
- Less code where function is called (vs. function ptrs where type casts are required)

Negative
- Much more code in `Kokkos_Cuda_Instance.hpp` where wrappers are defined

Notes
- I did not use the wrappers in the unit tests
- A separate impl file could be added for the wrappers to make `CudaInternal` class much more readable (at the cost of more code lines)

Ping @crtrott.

### DIFF
--- a/core/src/Cuda/Kokkos_CudaSpace.cpp
+++ b/core/src/Cuda/Kokkos_CudaSpace.cpp
@@ -75,7 +75,7 @@ void DeepCopyAsyncCuda(const Cuda &instance, void *dst, const void *src,
                        size_t n) {
   KOKKOS_IMPL_CUDA_SAFE_CALL(
       (instance.impl_internal_space_instance()->cuda_memcpy_async_wrapper(
-          dst, src, n, cudaMemcpyDefault, instance.cuda_stream())));
+          dst, src, n, cudaMemcpyDefault)));
 }
 
 void DeepCopyAsyncCuda(void *dst, const void *src, size_t n) {
@@ -174,21 +174,23 @@ void *impl_allocate_common(const Cuda &exec_space, const char *arg_label,
   cudaError_t error_code;
   if (arg_alloc_size >= memory_threshold_g) {
     if (exec_space_provided) {
-      cudaStream_t stream = exec_space.cuda_stream();
       error_code =
           exec_space.impl_internal_space_instance()->cuda_malloc_async_wrapper(
-              &ptr, arg_alloc_size, stream);
+              &ptr, arg_alloc_size);
       exec_space.fence("Kokkos::Cuda: backend fence after async malloc");
     } else {
-      error_code =
-          exec_space.impl_internal_space_instance()->cuda_malloc_async_wrapper(
-              &ptr, arg_alloc_size, 0);
+      error_code = Impl::CudaInternal::singleton().cuda_malloc_async_wrapper(
+          &ptr, arg_alloc_size);
       Impl::cuda_device_synchronize(
           "Kokkos::Cuda: backend fence after async malloc");
     }
   } else {
-    error_code = exec_space.impl_internal_space_instance()->cuda_malloc_wrapper(
-        &ptr, arg_alloc_size);
+    error_code =
+        (exec_space_provided
+             ? exec_space.impl_internal_space_instance()->cuda_malloc_wrapper(
+                   &ptr, arg_alloc_size)
+             : Impl::CudaInternal::singleton().cuda_malloc_wrapper(
+                   &ptr, arg_alloc_size));
   }
 #else
   cudaError_t error_code;
@@ -350,7 +352,7 @@ void CudaSpace::impl_deallocate(
           "Kokkos::Cuda: backend fence before async free");
       KOKKOS_IMPL_CUDA_SAFE_CALL(
           (Impl::CudaInternal::singleton().cuda_free_async_wrapper(
-              arg_alloc_ptr, 0)));
+              arg_alloc_ptr)));
       Impl::cuda_device_synchronize(
           "Kokkos::Cuda: backend fence after async free");
     } else {
@@ -606,7 +608,7 @@ void cuda_prefetch_pointer(const Cuda &space, const void *ptr, size_t bytes,
       space.cuda_device_prop().concurrentManagedAccess) {
     KOKKOS_IMPL_CUDA_SAFE_CALL(
         (space.impl_internal_space_instance()->cuda_mem_prefetch_async_wrapper(
-            ptr, bytes, space.cuda_device(), space.cuda_stream())));
+            ptr, bytes, space.cuda_device())));
   }
 }
 
--- a/core/src/Cuda/Kokkos_CudaSpace.cpp
+++ b/core/src/Cuda/Kokkos_CudaSpace.cpp
@@ -43,7 +43,8 @@
 cudaStream_t Kokkos::Impl::cuda_get_deep_copy_stream() {
   static cudaStream_t s = nullptr;
   if (s == nullptr) {
-    cudaStreamCreate(&s);
+    KOKKOS_IMPL_CUDA_SAFE_CALL(
+        (CudaInternal::singleton().cuda_stream_create_wrapper(&s)));
   }
   return s;
 }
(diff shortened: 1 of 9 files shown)


crtrott:
I personally like this approach better.

tcclevenger:
> I personally like this approach better.

I'm starting to agree. Extra ~100 lines of code are worth the simplicity. Any opinion on putting the function definitions in a separate header or the cpp file?

masterleinad (approved):
Looks reasonable to me.

core/src/Cuda/Kokkos_Cuda_Instance.hpp
```diff
@@ -193,6 +194,329 @@ class CudaInternal {
...
+                                      cudaMemoryAdvise advice,
+                                      int device) const {
+    if (setCudaDevice) set_cuda_device();
+    return cudaMemAdvise(devPtr, count, advice, device);
+  }
+
+  template <bool setCudaDevice = true>
+  cudaError_t cuda_mem_prefetch_async_wrapper(
+      const void* devPtr, size_t count, int dstDevice,
+      cudaStream_t stream = nullptr) const {
```
masterleinad, line 404: Do we need a default argument here? Shouldn't this always be the stream associated with this instance?

tcclevenger: I was just matching the args for cuda api functions. But if there is no case where a different stream would be given then I will just get the stream from the instance.

core/src/Cuda/Kokkos_CudaSpace.cpp
```diff
@@ -66,19 +67,22 @@ static std::atomic<int> num_uvm_allocations(0);
...
 void DeepCopyAsyncCuda(const Cuda &instance, void *dst, const void *src,
                        size_t n) {
   KOKKOS_IMPL_CUDA_SAFE_CALL(
-      cudaMemcpyAsync(dst, src, n, cudaMemcpyDefault, instance.cuda_stream()));
+      (instance.impl_internal_space_instance()->cuda_memcpy_async_wrapper(
+          dst, src, n, cudaMemcpyDefault, instance.cuda_stream())));
 }
 
 void DeepCopyAsyncCuda(void *dst, const void *src, size_t n) {
   cudaStream_t s = cuda_get_deep_copy_stream();
```
tcclevenger, line 82: @masterleinad I don't understand exactly what the deep copy stream and space are. It looks to be separate from the singleton instance?

So then would the call below `CudaInternal::singleton().cuda_memcpy_async_wrapper(()` still need the string as an argument? Since otherwise it would use the singleton stream.

masterleinad: The `deep_copy` stream is managed by the singleton but is not the same as the singleton's stream. So yes, that wrapper would still need a stream argument.

core/src/Cuda/Kokkos_Cuda_GraphNodeKernel.hpp
```diff
@@ -31,7 +31,6 @@
 #include <Kokkos_PointerOwnership.hpp>
 
 #include <Cuda/Kokkos_Cuda.hpp>
-#include <cuda_runtime_api.h>
```
masterleinad, line 34: What happened here (and similar places removing `cuda_runtime_api.h`?

core/src/Cuda/Kokkos_Cuda_Instance.cpp
```diff
@@ -157,16 +165,20 @@ void cuda_stream_synchronize(
...
 
 void cuda_internal_error_throw(cudaError e, const char *name, const char *file,
                                const int line) {
   std::ostringstream out;
-  out << name << " error( " << cudaGetErrorName(e)
-      << "): " << cudaGetErrorString(e);
+  out << name << " error( "
+      << CudaInternal::singleton().cuda_get_error_name_wrapper<false>(e)
+      << "): "
+      << CudaInternal::singleton().cuda_get_error_string_wrapper<false>(e);
```
masterleinad, line 181: Is there a particular reason not to set the device here?

core/src/Cuda/Kokkos_Cuda_Instance.hpp
```diff
@@ -21,6 +21,7 @@
 #include <impl/Kokkos_Tools.hpp>
 #include <atomic>
 #include <Cuda/Kokkos_Cuda_Error.hpp>
+#include <cuda_runtime_api.h>
```
masterleinad, line 24: I guess you remove this include from all files that include this file?

core/src/Cuda/Kokkos_Cuda_Instance.hpp
```diff
@@ -193,6 +194,338 @@ class CudaInternal {
     }
   }
 
+  // Using cudaAPI function/objects will be w.r.t. device 0 unless
+  // cudaSetDevice(device_id) is called with the correct device_id.
+  // The correct device_id is stored in the static variable
+  // CudaInternal::m_cudaDev set in Cuda::impl_initialize(). It is not
```
masterleinad, line 200: Soon that variable will not be `static` anymore.

tcclevenger: I'll remove mention of static in the comment.

core/src/Cuda/Kokkos_Cuda_Instance.hpp
tcclevenger: Yes, since there are no runtime api calls anywhere else now. But maybe this file includes types that are still used in the other `core/src/Cuda/` files?

core/src/Cuda/Kokkos_Cuda_Instance.cpp
tcclevenger: I did this since all the cudaAPI calls wrapped in `KOKKOS_IMPL_CUDA_SAFE_CALL()` will hit this (or below) if there is a runtime error, and we can't be guaranteed that the call that is erroring out could set device or not. For example if `cuda_get_device_count_wrapper<false>(...)` in line 271 of this file had an error, we would not get the message because before error was thrown, these functions would complain that the device was not initialized.

core/src/Cuda/Kokkos_Cuda_Instance.hpp
tcclevenger: I have a branch that adds the test that checks whether or not header files are self contained for the `core/src/Cuda` files. This would catch that case I believe?

core/src/Cuda/Kokkos_Cuda_Instance.cpp
masterleinad: Sounds reasonable.

core/src/Cuda/Kokkos_Cuda_Instance.hpp
masterleinad: I think it's fine.

masterleinad (approved):
Looks good to me. When adding support for multiple devices we will need to again review all the places where the singleton is used anyway.

core/src/Cuda/Kokkos_CudaSpace.cpp
```diff
@@ -43,7 +43,8 @@
 cudaStream_t Kokkos::Impl::cuda_get_deep_copy_stream() {
   static cudaStream_t s = nullptr;
   if (s == nullptr) {
-    cudaStreamCreate(&s);
+    KOKKOS_IMPL_CUDA_SAFE_CALL(
+        (CudaInternal::singleton().cuda_stream_create_wrapper(&s)));
```
fnrizzi, line 47: based on the discussion of Wed 26 july, can we please open an issue where we can the name of this `singleton` into something else since it is not a singleton?

masterleinad:
You will have to resolve conflicts.

crtrott (approved):
As a follow on optimization can we identify calls which always will be preceded by another call so one wouldn't need to call setCudaDevice?

masterleinad:
> As a follow on optimization can we identify calls which always will be preceded by another call so one wouldn't need to call setCudaDevice?

I thought we already convinced ourselves that the performance impact of this pull request is negligible?

tcclevenger:
> As a follow on optimization can we identify calls which always will be preceded by another call so one wouldn't need to call setCudaDevice?

I had that at one point in a previous commit on the other PR, but like Daniel said there wasn't a real performance impact, so I removed it since it made the already complicated looking template params even more complicated. Now that the calls are much simpler (with no templates needed for inputs), I'll add back in a follow up branch and get some performance numbers.

crtrott:
CUDA 11.6 failed with non-available container, but all other CUDA builds passed so I am merging.

tcclevenger:
Thanks @crtrott. Setting up an issue for HIP version.

PR #6300, opened 2023-07-21
 SYCL: Use sycl::bit_cast from oneAPI 2024.0.0 on 

masterleinad:
Follow-up to #6121 and #6137. 
https://github.com/intel/llvm/pull/10215 removes `sycl::detail::biit_cast` which we needed before https://github.com/intel/llvm/pull/9398 to fix ambiguities between `Kokkos::bit_cast` and `sycl::bit_cast` due to some unconstrained calls. This latter pull request will be included in the `oneAPI 2024.0.0` release.
After https://github.com/intel/llvm/pull/9398, we can also enable all the `Kokos::bit_cast` tests for `SYCL` since that pull request also constrained `sycl::bit_cast` properly.
An alternative to importing `sycl::bit_cast` into the `Kokkos` namespace would be to call it inside `Kokkos::bit_cast` of course but it's not obvious to me that this would be better and this way it's (slightly) easier to support multiple `oneAPI` versions.

### DIFF
--- a/core/src/Kokkos_BitManipulation.hpp
+++ b/core/src/Kokkos_BitManipulation.hpp
@@ -100,12 +100,9 @@ inline constexpr bool is_standard_unsigned_integer_type_v =
 namespace Kokkos {
 
 //<editor-fold desc="[bit.cast], bit_cast">
-#ifdef KOKKOS_ENABLE_SYCL
-#if defined(__INTEL_LLVM_COMPILER) && __INTEL_LLVM_COMPILER >= 20240000
-using sycl::bit_cast;
-#else
+#if defined(KOKKOS_ENABLE_SYCL) && defined(__INTEL_LLVM_COMPILER) && \
+    __INTEL_LLVM_COMPILER < 20240000
 using sycl::detail::bit_cast;
-#endif
 #else
 template <class To, class From>
 KOKKOS_FUNCTION std::enable_if_t<sizeof(To) == sizeof(From) &&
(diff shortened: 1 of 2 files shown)


masterleinad:
Only HIP-ROCm-5.2-C++20 CI is timing out.

fnrizzi:
>An alternative to importing sycl::bit_cast into the Kokkos namespace would be to call it inside Kokkos::bit_cast of

I think this a bit subjective, but I personally like this better because it more clearly expresses that users interface with Kokkos code *directly* and kokkos underneath calls native stuff. 
For example, IMO one nice thing is this: https://godbolt.org/ 
Someone tries to call Kokkos::bit_cast but gets types wrong and they get a compile message from std. 
This can be misleading. Whereas if the call to native func is wrapped, we would get a message that immediately tells you it is a kokkos call.

masterleinad:
@fnrizzi Here you go!

fnrizzi:
> @fnrizzi Here you go!

appreciate it :) we still have the previous one but that is fine

fnrizzi (approved):

nliber:
Another reason to wrap it (as this code does) is our version is not `constexpr`, while `sycl::bit_cast` probably is because of `__has_builtin(__builtin_bit_cast)` (see [https://github.com/intel/llvm/blob/sycl/sycl/include/sycl/bit_cast.hpp](https://github.com/intel/llvm/blob/sycl/sycl/include/sycl/bit_cast.hpp)).  If we want portability then ours has to not be `constexpr` until C++23.

nliber (approved):

crtrott (approved):
Test Failures look unrelated.

PR #6311, opened 2023-07-26
Preserve execution space in HostMirror if possible

### LINKED ISSUE #6310: View::HostMirror should preserve execution space for Views constructed with a host-accessible execution space
We noticed that `View::HostMirror` uses the `DefaultHostExecutionSpace` even if the `View`'s execution space can access `Kokkos::HostSpace`. This only makes a difference for the host parallel backends (Threads, OpenMP, HPX). Related to #6309.

masterleinad:
Fixes #6310.

### DIFF
--- a/core/src/Kokkos_View.hpp
+++ b/core/src/Kokkos_View.hpp
@@ -614,11 +614,15 @@ class View : public ViewTraits<DataType, Properties...> {
            typename traits::memory_traits>;
 
   /** \brief  Compatible HostMirror view */
-  using HostMirror =
-      View<typename traits::non_const_data_type, typename traits::array_layout,
-           Device<DefaultHostExecutionSpace,
-                  typename traits::host_mirror_space::memory_space>,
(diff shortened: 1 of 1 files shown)


masterleinad:
It turns out that doing
```
using HostMirror = View<
      typename traits::non_const_data_type, typename traits::array_layout,
      Device<std::conditional_t<
                 SpaceAccessibility<typename traits::execution_space,
                                    typename traits::host_mirror_space::
                                        memory_space>::accessible,
                 typename traits::execution_space, DefaultHostExecutionSpace>,
             typename traits::host_mirror_space::memory_space>,
      typename traits::hooks_policy>;
```
isn't compatible with the `DualView` implementation (for `SharedSpace`/`CudaUVMSpace`) that detects the `View` to work on based on a full match of the device derived from `HostMirror`. The last commit instead only keeps the execution space if it can access `HostSpace`.

masterleinad:
Not important enough

PR #6312, opened 2023-07-27
add helper variable templates `are_*_iterators_v`

fnrizzi:
self-explanatory 

Why needed: for the team-level algorithms PRs, to address comments

### DIFF
--- a/algorithms/src/std_algorithms/impl/Kokkos_Constraints.hpp
+++ b/algorithms/src/std_algorithms/impl/Kokkos_Constraints.hpp
@@ -72,7 +72,7 @@ struct are_iterators<T> {
 template <class Head, class... Tail>
 struct are_iterators<Head, Tail...> {
   static constexpr bool value =
-      are_iterators<Head>::value && are_iterators<Tail...>::value;
+      are_iterators<Head>::value && (are_iterators<Tail>::value && ... && true);
 };
 
 template <class... Ts>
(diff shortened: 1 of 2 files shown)


cz4rs (approved):

masterleinad (approved):

algorithms/unit_tests/TestRandomAccessIterator.cpp
```diff
@@ -217,5 +217,15 @@ TEST_F(random_access_iterator_test, distance) {
...
 
+TEST_F(random_access_iterator_test, traits_helpers) {
+  using T1_t = KE::Impl::RandomAccessIterator<static_view_t>;
+  using T2_t = KE::Impl::RandomAccessIterator<dyn_view_t>;
+  using T3_t = KE::Impl::RandomAccessIterator<strided_view_t>;
+
+  namespace KE = Kokkos::Experimental;
+  static_assert(KE::Impl::are_iterators_v<T1_t, T2_t, T3_t>);
+  static_assert(KE::Impl::are_random_access_iterators_v<T1_t, T2_t, T3_t>);
+}
```
crtrott, line 228:
```suggestion
  static_assert(!KE::Impl::are_iterators_v<T1_t, int, T3_t>);
  static_assert(!KE::Impl::are_random_access_iterators_v<T1_t, int, T3_t>);
}
```

crtrott (approved):
Approve anyway please do follow up PR with added test

crtrott:
Also I would like to replace recursive impl with fold expression. If you care that its not compiling with zero args it would be:

```c++
template <class Head, class... Tail>
struct are_iterators<Head, Tail...> {
  static constexpr bool value =
      are_iterators<Head>::value && (are_iterators<Tail>::value && ... && true);
};

File: algorithms/src/std_algorithms/impl/Kokkos_Constraints.hpp [hunk 2]
```diff
@@ -63,15 +66,18 @@ struct are_iterators;
 
 template <class T>
 struct are_iterators<T> {
-  static constexpr bool value = is_iterator<T>::value;
+  static constexpr bool value = is_iterator_v<T>;
 };
```
crtrott, line 70: This guy could go away, the one below would match, but its fine for now

crtrott (approved):

PR #6313, opened 2023-07-27
Use checked arithmetic builtins for overflow detection

cz4rs:
Follow-up to #6159 and https://github.com/kokkos/kokkos/pull/6159#discussion_r1275548410.

Use checked arithmetic builtins when available. Both `clang` and `gcc` provide those:
- https://clang.llvm.org/docs/LanguageExtensions.html#checked-arithmetic-builtins
- https://gcc.gnu.org/onlinedocs/gcc/Integer-Overflow-Builtins.html

_Note: [Prior to Clang 10, `__has_builtin` could not be used to detect most builtin pseudo-functions](https://clang.llvm.org/docs/LanguageExtensions.html#has-builtin). This should not be an issue here, so I opted to use that instead of explicitly checking for compilers._

### DIFF
--- a/core/src/impl/Kokkos_CheckedIntegerOps.hpp
+++ b/core/src/impl/Kokkos_CheckedIntegerOps.hpp
@@ -29,9 +29,13 @@ std::enable_if_t<std::is_integral_v<T>, bool> constexpr multiply_overflow(
     T a, T b, T& res) {
   static_assert(std::is_unsigned_v<T>,
                 "Operation not implemented for signed integers.");
-#if defined(KOKKOS_COMPILER_CLANG) || defined(KOKKOS_COMPILER_GNU)
+
+#if defined(__has_builtin)
+#if __has_builtin(__builtin_mul_overflow)
   return __builtin_mul_overflow(a, b, &res);
-#else
+#endif
+#endif
+
   auto product = a * b;
   if ((a == 0) || (b == 0) || (a == product / b)) {
     res = product;
@@ -39,7 +43,6 @@ std::enable_if_t<std::is_integral_v<T>, bool> constexpr multiply_overflow(
   } else {
     return true;
   }
-#endif
 }
 
 template <typename T>
--- a/core/src/impl/Kokkos_CheckedIntegerOps.hpp
+++ b/core/src/impl/Kokkos_CheckedIntegerOps.hpp
@@ -29,6 +29,9 @@ std::enable_if_t<std::is_integral_v<T>, bool> constexpr multiply_overflow(
     T a, T b, T& res) {
   static_assert(std::is_unsigned_v<T>,
                 "Operation not implemented for signed integers.");
+#if defined(KOKKOS_COMPILER_CLANG) || defined(KOKKOS_COMPILER_GNU)
+  return __builtin_mul_overflow(a, b, &res);
+#else
   auto product = a * b;
   if ((a == 0) || (b == 0) || (a == product / b)) {
     res = product;
@@ -36,6 +39,7 @@ std::enable_if_t<std::is_integral_v<T>, bool> constexpr multiply_overflow(
   } else {
     return true;
   }
+#endif
 }
 
 template <typename T>
--- a/core/src/impl/Kokkos_CheckedIntegerOps.hpp
+++ b/core/src/impl/Kokkos_CheckedIntegerOps.hpp
@@ -24,18 +24,21 @@
 namespace Kokkos {
 namespace Impl {
 
+#if defined(__has_builtin)
+#if __has_builtin(__builtin_mul_overflow)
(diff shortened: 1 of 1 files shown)


cz4rs:
Note 2: C23 [is going to provide](https://open-std.org/JTC1/SC22/WG14/www/docs/n2683.pdf) a standard way of dealing with overflows in `<stdckdint.h>` header.
```
bool ckd_add(type1 *result, type2 a, type3 b);
bool ckd_sub(type1 *result, type2 a, type3 b);
bool ckd_mul(type1 *result, type2 a, type3 b);
```

We might want to use that in the far future. :wink:

core/src/impl/Kokkos_CheckedIntegerOps.hpp
```diff
@@ -29,6 +29,11 @@ std::enable_if_t<std::is_integral_v<T>, bool> constexpr multiply_overflow(
     T a, T b, T& res) {
   static_assert(std::is_unsigned_v<T>,
                 "Operation not implemented for signed integers.");
+
+#if defined(__has_builtin) && __has_builtin(__builtin_mul_overflow)
+  return __builtin_mul_overflow(a, b, &res);
+#endif
```
crtrott, line 35: I think we should ifdef out the rest via a #else clause

crtrott (changes requested):

crtrott:
> Note 2: C23 [is going to provide](https://open-std.org/JTC1/SC22/WG14/www/docs/n2683.pdf) a standard way of dealing with overflows in `<stdckdint.h>` header.
> 
> ```
> bool ckd_add(type1 *result, type2 a, type3 b);
> bool ckd_sub(type1 *result, type2 a, type3 b);
> bool ckd_mul(type1 *result, type2 a, type3 b);
> ```
> 
> We might want to use that in the far future. 😉

Not actually sure that even C++26 is based on C23, pretty sure C++23 is not.

core/src/impl/Kokkos_CheckedIntegerOps.hpp
cz4rs: We have to use `#if` twice (to avoid errors on compilers that don't have `__has_builtin`), which would make the `#else` approach somewhat ugly.

This is also consistent with existing usage:
https://github.com/kokkos/kokkos/blob/1a3ea28f6e97b4c9dd2c8ceed53ad58ed5f94dfe/core/src/Kokkos_BitManipulation.hpp#L268-L274

cz4rs:
retest me please

cz4rs:
Retest this please

nliber:
FYI:  C++23 is based on C17, as C23 has technically not shipped yet (and probably not until early 2024).

nliber (approved):

cz4rs:
Rebased to include Clang-Tidy fix.

core/src/impl/Kokkos_CheckedIntegerOps.hpp
crtrott: Are you saying this doesn't work: 

```c++
#if defined(__has_builtin) && __has_builtin(__builtin_mul_overflow)
  return __builtin_mul_overflow(a, b, &res);
#else
 ...
 #endif
 ```

Comment (crtrott) [reply core/src/impl/Kokkos_CheckedIntegerOps.hpp:36]: Yeah ok it doesn't work on ICC

Comment (crtrott) [reply core/src/impl/Kokkos_CheckedIntegerOps.hpp:36]: and we don't get unreachable code warnings?

**Comment by crtrott** [comment]
Ah NVC++ does warn about unreachable code, so I would vote for the more ugly ifdef option :
How about this: outside of the function you define a macro:

```c++
#if defined(__has_builtin)
#if __has_builtin(__builtin_mul_overflow)
  #define KOKKOS_IMPL_USE_MUL_OVERFLOW_BUILTIN
#endif
#endif
```

and then inside the function:
```c++
#if defined(KOKKOS_IMPL_USE_MUL_OVERFLOW_BUILTIN
return __builtin_mul_overflow(a, b, &res);
#else
  auto product = a * b;
  if ((a == 0) || (b == 0) || (a == product / b)) {
    res = product;
    return false;
  } else {
    return true;
  }
#endif
```

and then undefine the macro again?

crtrott (changes requested):
I'd like to avoid the warning with NVC++ about unreachable code.

crtrott:
https://godbolt.org/z/18WWrK7s7 with warning, but it also shows up in our CI for the NVHPC build.

cz4rs:
Retest this please

cz4rs:
Retest this please

cz4rs:
Retest this please

Rombur:
Retest this please

crtrott (approved):

PR #6315, opened 2023-07-27
Make temp_dir before running nvcc

sebrowne:
We set TMPDIR for nvcc, but sometimes it doesn't exist.  Seems reasonable to handle that here, because nvcc will crash otherwise.


sebrowne:
Fair enough.  Upon reconsidering I think I agree with not creating it.  Originally my thought was "we redefine the temp dir", but thinking about it, nvcc_wrapper doesn't change the compiler behavior here.  If the client is setting NVCC_WRAPPER_TMPDIR, they should indeed be responsible for creating that directory.

PR #6320, opened 2023-07-29
SIMD: convert binary operators to hidden friends

ldh4:
Follow up from https://github.com/kokkos/kokkos/pull/6109#discussion_r1276669755.

Converted binary operators to hidden friends and modified operator signatures to stay consistent with class overview provided in [P1928R6](https://wg21.link/P1928R6).

### DIFF
--- a/simd/src/Kokkos_SIMD_AVX2.hpp
+++ b/simd/src/Kokkos_SIMD_AVX2.hpp
@@ -472,70 +472,67 @@ class simd<double, simd_abi::avx2_fixed_size<4>> {
       const {
     return m_value;
   }
-  KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION mask_type
-  operator<(simd const& other) const {
-    return mask_type(_mm256_cmp_pd(m_value, other.m_value, _CMP_LT_OS));
+  [[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION simd operator-() const
+      noexcept {
+    return simd(
+        _mm256_sub_pd(_mm256_set1_pd(0.0), static_cast<__m256d>(m_value)));
   }
-  KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION mask_type
-  operator>(simd const& other) const {
-    return mask_type(_mm256_cmp_pd(m_value, other.m_value, _CMP_GT_OS));
+
+  [[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION friend simd operator*(
+      simd const& lhs, simd const& rhs) noexcept {
+    return simd(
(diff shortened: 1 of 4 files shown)


nliber:
Looks good so far...

simd/src/Kokkos_SIMD_Scalar.hpp
```diff
@@ -149,53 +124,76 @@ class simd<T, simd_abi::scalar> {
...
+  }
+  [[nodiscard]] KOKKOS_FORCEINLINE_FUNCTION friend constexpr simd operator>>(
       simd const& lhs, int rhs) noexcept {
     return simd(lhs.m_value >> rhs);
   }
-  KOKKOS_FORCEINLINE_FUNCTION friend constexpr simd operator>>(
+  [[nodiscard]] KOKKOS_FORCEINLINE_FUNCTION friend constexpr simd operator>>(
       simd const& lhs, simd const& rhs) noexcept {
-    return simd(lhs.m_value >> static_cast<int>(rhs[0]));
+    return simd(lhs.m_value >> static_cast<value_type>(rhs));
```
dalg24, line 153: Why are we casting to `value_type` rather than using `rhs.value` directly?

simd/src/Kokkos_SIMD_AVX2.hpp
```diff
@@ -412,71 +412,68 @@ class simd<double, simd_abi::avx2_fixed_size<4>> {
       const {
     return m_value;
   }
-  KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION mask_type
-  operator<(simd const& other) const {
-    return mask_type(_mm256_cmp_pd(m_value, other.m_value, _CMP_LT_OS));
+  [[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION simd operator-() const
```
dalg24, line 415: What about `constexpr`?

dalg24 (commented):
Why do we only have constexpr specifier with the scalar implementation?

simd/src/Kokkos_SIMD_Scalar.hpp
ldh4: Replaced to use `rhs.m_value` instead. [787c301](https://github.com/kokkos/kokkos/pull/6320/commits/787c3010702bb0005ffbc25091b126a2e23cfbc3)

simd/src/Kokkos_SIMD_AVX2.hpp
ldh4: It's because these intrinsic functions are not `constexpr`.
Placing `constexpr` specifier will produce errors: `call to non-'constexpr' function '__m256d _mm256_sub_pd`.

nliber (approved):

dalg24 (approved):

PR #6321, opened 2023-07-30
Allow using the SYCL execution space on AMD GPUs

G-071:
This pull request adds the changes necessary to run the Kokkos SYCL execution space on AMD GPUs (if ```Kokkos_ENABLE_UNSUPPORTED_ARCH=ON```).

Of course, I realize that this is not the intended execution space for AMD GPUs (hence I made sure it cannot be used without ```Kokkos_ENABLE_UNSUPPORTED_ARCH```).  However, I needed these changes as I adapted the toolchain of our own simulation (which uses Kokkos) for SYCL and dpcpp but lacked access to an Intel GPU as yet - hence, I wanted to test it on an AMD MI100.

The changes required to do this with Kokkos are quite minimal; one simply has to make sure the correct AMD architecture flags are passed to dpcpp (as outlined [here](https://github.com/intel/llvm/blob/sycl/sycl/doc/GetStartedGuide.md#run-simple-dpc-application)). Note that the logic to similarly run the SYCL execution space on NVIDIA devices already existed in ```cmake/kokkos_arch.cmake```, I merely adapted it to also work for AMD GPUs.

I tested it (using our simulation [Octo-Tiger](https://github.com/STEllAR-GROUP/octotiger)) with dpcpp/2023-03 on a MI100 and everything seems to work fine! I figured it might be useful to other people as well, even if just for testing/benchmark purposes, hence this PR!


cmake/kokkos_arch.cmake
```diff
@@ -573,7 +573,7 @@ FUNCTION(CHECK_CUDA_ARCH ARCH FLAG)
     ENDIF()
     SET(CUDA_ARCH_ALREADY_SPECIFIED ${ARCH} PARENT_SCOPE)
     IF (NOT KOKKOS_ENABLE_CUDA AND NOT KOKKOS_ENABLE_OPENMPTARGET AND NOT KOKKOS_ENABLE_SYCL AND NOT KOKKOS_ENABLE_OPENACC)
-      MESSAGE(WARNING "Given CUDA arch ${ARCH}, but Kokkos_ENABLE_CUDA, Kokkos_ENABLE_OPENACC, and Kokkos_ENABLE_OPENMPTARGET are OFF. Option will be ignored.")
+      MESSAGE(WARNING "Given CUDA arch ${ARCH}, but Kokkos_ENABLE_CUDA, Kokkos_ENABLE_SYCL, Kokkos_ENABLE_OPENACC, and Kokkos_ENABLE_OPENMPTARGET are OFF. Option will be ignored.")
```
crtrott, line 576: I think we left this off here intentionally since its unsupported.

crtrott: @masterleinad thoughts?

crtrott: On the other hand we print them above, so I guess could go either way.

crtrott (commented):
Looks reasonable to me. I personally wouldn't go so far for us to set CI up, maybe other than a nightly? In the end we are a bit resource constrained during PR testing.

cmake/kokkos_arch.cmake
masterleinad: I think it makes sense to print `SYCL` in the warning as well.

masterleinad (approved):
Looks reasonable to me (but we won't maintain support or promise SYCL to work on AMD GPUs).

crtrott (approved):

masterleinad:
OK to test.

G-071:
Thanks for the feedback! 

One more thing: I noticed that one of the tests failed due to some trailing whitespaces I missed in cmake/kokkos_arch.cmake. I have fixed that in the last commit just now!

PR #6322, opened 2023-07-31
allow sorting via native oneDPL  to support views with stride = 1

fnrizzi:
Before this PR, the call to native sort in oneDPL was guarded with static asserts for views to have layoutRight and layoutLeft. 
Since technically oneDPL can support stride = 1, this PR also allows for LayoutStride but adds a runtime check that stride ==1.

### DIFF
--- a/algorithms/src/sorting/impl/Kokkos_SortImpl.hpp
+++ b/algorithms/src/sorting/impl/Kokkos_SortImpl.hpp
@@ -198,7 +198,7 @@ void sort_onedpl(const Kokkos::Experimental::SYCL& space,
   static_assert(
       (ViewType::rank == 1) &&
           (std::is_same_v<typename ViewType::array_layout, LayoutRight> ||
-               std::is_same_v<typename ViewType::array_layout, LayoutLeft>,
+           std::is_same_v<typename ViewType::array_layout, LayoutLeft> ||
            std::is_same_v<typename ViewType::array_layout, LayoutStride>),
       "SYCL sort only supports contiguous rank-1 Views with LayoutLeft, "
       "LayoutRight or LayoutStride"
--- a/algorithms/src/sorting/impl/Kokkos_SortImpl.hpp
+++ b/algorithms/src/sorting/impl/Kokkos_SortImpl.hpp
@@ -285,8 +285,8 @@ void sort_device_view_without_comparator(
           (std::is_same_v<typename ViewType::array_layout, LayoutRight> ||
            std::is_same_v<typename ViewType::array_layout, LayoutLeft> ||
            std::is_same_v<typename ViewType::array_layout, LayoutStride>),
-      "sort_device_view_without_comparator: support rank-1 Views "
-      " with LayoutLeft, LayoutRight or LayoutStride");
+      "sort_device_view_without_comparator: supports rank-1 Views "
+      "with LayoutLeft, LayoutRight or LayoutStride");
 
   if (view.stride(0) == 1) {
     sort_onedpl(exec, view);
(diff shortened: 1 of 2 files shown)


masterleinad (approved):

nliber (approved):

algorithms/src/sorting/impl/Kokkos_SortImpl.hpp
```diff
@@ -195,18 +195,17 @@ void sort_onedpl(const Kokkos::Experimental::SYCL& space,
...
-  // the data is on the host.
-  static_assert(
-      ViewType::rank == 1 &&
-          (std::is_same<typename ViewType::array_layout, LayoutRight>::value ||
-           std::is_same<typename ViewType::array_layout, LayoutLeft>::value),
-      "SYCL sort only supports contiguous rank-1 Views.");
+  static_assert(ViewType::rank == 1, "SYCL sort only supports rank-1 Views");
+  if (view.stride(0) != 1) {
+    Kokkos::abort("SYCL sort only supports rank-1 Views with stride = 1".);
+  }
```
crtrott, line 201: This ignored LayoutTiled, and would compile with it

crtrott (changes requested):
We don't have a is_always_strided property like mdspan does, but basically that is what you need to check for here. LayoutTiled for example breaks this.

nliber (changes requested):
I agree with @crtrott.  The old `static_assert` needs to be there.

algorithms/src/sorting/impl/Kokkos_SortImpl.hpp
fnrizzi: ops, yes totally, i actually had it that way in the first place and then simplified too much. thanks for catching it !

algorithms/src/sorting/impl/Kokkos_SortImpl.hpp
```diff
@@ -305,14 +313,10 @@ void sort_device_view_with_comparator(
...
     const Kokkos::View<DataType, Properties...>& view,
     const ComparatorType& comparator) {
-  using ViewType = Kokkos::View<DataType, Properties...>;
-  constexpr bool strided =
-      std::is_same_v<LayoutStride, typename ViewType::array_layout>;
-  if constexpr (strided) {
-    // strided views not supported in dpl so use the most generic case
-    copy_to_host_run_stdsort_copy_back(exec, view, comparator);
-  } else {
+  if (view.stride(0) == 1) {
```
masterleinad, line 316: We also need to match the checks in `sort_onedpl` above
```
          rank == 1 && 
          (std::is_same_v<typename ViewType::array_layout, LayoutRight> ||
           std::is_same_v<typename ViewType::array_layout, LayoutLeft> ||
           std::is_same_v<typename ViewType::array_layout, LayoutStride>),
```      
here so we can fallback to `copy_to_host_run_stdsort_copy_back` in case of a different layout without running into the `Kokkos::abort` or `static_assert`.

fnrizzi: ok, so i changed this because it makes sense for this specific function. However, this is within the impl namespace and as far as usability, we have a static assert in the public API that alreay statically checks only for these layouts. so one cannot pass a view with a different layout anyway. makes sense?

masterleinad: > we have a static assert in the public API that alreay statically checks only for these layouts.

That's not true for the overloads without comparator, see https://github.com/kokkos/kokkos/blob/d89140de00f90183dc6618167bd30a0f9470970b/algorithms/src/sorting/Kokkos_SortPublicAPI.hpp#L31-L72, but the ones with comparator have them https://github.com/kokkos/kokkos/blob/d89140de00f90183dc6618167bd30a0f9470970b/algorithms/src/sorting/Kokkos_SortPublicAPI.hpp#L77-L131.

masterleinad: Then we need this check when calling the version without comparator (since it's also hitting `sort_onedpl`), right?

nliber (approved):

algorithms/src/sorting/impl/Kokkos_SortImpl.hpp
fnrizzi: yes, I did not want touch the existing stuff, but makes sense IMO move the constraints as much as possible into the public API rather than having them hidden. So yes i am changing this to also add these constraints to the public API of the sort without comparator.

crtrott (approved):

PR #6328, opened 2023-08-02
Fix gtest when using C++20 with ROCm 5.6

Rombur:
There is a bug in gtest when using C++20 with newer version of ROCm. This has been fixed upstream [here](
https://github.com/google/googletest/commit/be03d00f5f0cc3a997d1a368bee8a1fe93651f48) but it is not part of any release yet.

### DIFF
--- a/tpls/gtest/gtest/gtest-all.cc
+++ b/tpls/gtest/gtest/gtest-all.cc
@@ -11957,7 +11957,7 @@ static const char* GetCharWidthPrefix(unsigned char) {
   return "";
 }
 
-#ifdef __cpp_char8_t
+#ifdef __cpp_lib_char8_t
 static const char* GetCharWidthPrefix(char8_t) {
   return "u8";
 }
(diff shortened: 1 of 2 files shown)


arghdos:
thanks @Rombur!

cz4rs:
Failing test is `Kokkos_CoreUnitTest_Default` -> `defaultdevicetype.shared_space`. Any idea if this is somehow related?

Rombur:
No I don't see how it is related. Especially since this only happens for a single configuration.

arghdos:
Ah, I see -- got a bit down the debugging rabbit hole when I realized, that's a CUDA run that's failing.  The test is... a bit temperamental from looking at it, as in I understand _why_ we'd want to check page migrations by measuring changes in cycle accesses, but I can also see that failing intermittently.

However, I do see a similar failure on MI-210's when `HSA_XNACK` is not set however (perhaps we need to `GTEST_SKIP` that if we ever get one in the CI?), I'll follow up with @Rombur on slack.

crtrott (approved):

masterleinad (approved):

crtrott:
Yeah the failure is ok, this is a timing based test which can occasionally fail. With HSA_XNACK the test is supposed to fail, since we have a strong recommendation for it being ON, if you ever want to use SHARED_SPACE>

PR #6331, opened 2023-08-03
Initial BasicView/mdspan integration

nmm0:
This PR is still in pretty early stages. I have a decent amount of things still left to do, in particularly integration with `deep_copy`. However I do want to get early design feedback.

(Note the CI won't pass unless I snapshot my WIP branch from mdspan which I don't want to do to keep the diff relatively clean)

TODOs:
- [ ] `deep_copy`
- [ ] subview constructor
- [ ] Probably more things...


core/src/View/Kokkos_ViewAlloc.hpp
```diff
@@ -0,0 +1,408 @@
...
+#define KOKKOS_VIEW_ALLOC_HPP
+
+#include <impl/Kokkos_SharedAlloc.hpp>
+#include <impl/Kokkos_ZeroMemset_fwd.hpp>
+
+namespace Kokkos::Impl
+{
+
+  template <typename T>
+  inline bool is_zero_byte(const T& t) {
```
nmm0, line 32: Moved from Kokkos_ViewMapping.hpp

core/src/View/Kokkos_ViewAlloc.hpp
```diff
@@ -0,0 +1,408 @@
...
+  *  called from the shared memory tracking destruction.
+  *  Secondarily to have two fewer partial specializations.
+  */
+  template <class DeviceType, class ValueType,
+            bool IsScalar = std::is_scalar<ValueType>::value>
+  struct ViewAllocFunctor;
+
+  template <class DeviceType, class ValueType>
+  struct ViewAllocFunctor<DeviceType, ValueType, false /* is_scalar */> {
+    using ExecSpace = typename DeviceType::execution_space;
```
nmm0, line 64: Moved from Kokkos_ViewMapping.hpp

core/src/Kokkos_Parallel.hpp
```diff
@@ -137,9 +137,9 @@ inline void parallel_for(const std::string& str, const ExecPolicy& policy,
   ExecPolicy inner_policy = policy;
   Kokkos::Tools::Impl::begin_parallel_for(inner_policy, functor, str, kpID);
 
-  Kokkos::Impl::shared_allocation_tracking_disable();
-  Impl::ParallelFor<FunctorType, ExecPolicy> closure(functor, inner_policy);
-  Kokkos::Impl::shared_allocation_tracking_enable();
+  auto closure = Kokkos::Impl::with_shared_allocation_tracking_disabled([&]() {
+    return Impl::ParallelFor<FunctorType, ExecPolicy>(functor, inner_policy);
+  });
```
crtrott, line 142: What is the rational behind this change? I don't necessarily disagree with it but not clear how this relates to mdspan?

core/src/View/Kokkos_BasicView.hpp
```diff
@@ -0,0 +1,260 @@
...
+};
+
+template <typename ExecutionSpace>
+using preferred_layout_for_exec_space = typename ExecutionSpace::array_layout;
+}  // namespace Impl
+
+template <class ElementType, class Extents,
+          class ExecutionSpace = Kokkos::DefaultExecutionSpace,
+          class MemorySpace    = typename ExecutionSpace::memory_space,
+          class LayoutPolicy =
```
crtrott, line 56: I think we should make basic view take the same things as mdspan in the same order, and essentially map ExecutionSpace/MemorySpace to accessors. Thoughts?

core/src/View/Kokkos_BasicView.hpp
```diff
@@ -0,0 +1,260 @@
...
+
+template <>
+struct kokkos_layout_conversion<LayoutLeft> {
+  using type = Experimental::layout_left_padded<dynamic_extent>;
+};
+
+template <>
+struct kokkos_layout_conversion<LayoutRight> {
+  using type = Experimental::layout_right_padded<dynamic_extent>;
+};
```
crtrott, line 47: I am not convinced that this is a good idea. In the end `layout_type` should be what you hand as a template argument to `View` and I think we should have the nested `mdspan` have the same layout as the original thing.

core/src/View/Kokkos_BasicView.hpp
```diff
@@ -0,0 +1,260 @@
...
+class BasicView {
+ private:
+  using internal_layout_type =
+      typename Impl::kokkos_layout_conversion<LayoutPolicy>::type;
+
+  using default_alloc_params = Impl::alloc_params<MemorySpace, ExecutionSpace>;
+
+ public:
+  using extents_type  = Extents;
+  using accessor_type = default_accessor<ElementType>;
```
crtrott, line 67: so my feeling is that we should use a `Kokkos::SpaceAwareAccessor<DeviceType, MemoryTraits>` as the standard thing.

core/src/View/Kokkos_BasicView.hpp
```diff
@@ -0,0 +1,260 @@
...
+    return m_data.stride(r);
+  }
+
+  KOKKOS_INLINE_FUNCTION constexpr index_type span() const noexcept {
+    return m_data.mapping().required_span_size();
+  }
+
+  KOKKOS_INLINE_FUNCTION constexpr bool span_is_contiguous() const noexcept {
+    return m_data.is_exhaustive();
+  }
```
crtrott, line 172: I think we should use this opportunity and introduce the mdspan names for the functions too, and then a bit closer to 5.0 deprecate the old names.

crtrott (changes requested):
I left a few fundamental comments.

core/src/Kokkos_Parallel.hpp
masterleinad: That's https://github.com/kokkos/kokkos/pull/6289.

nmm0: Yeah my branch is based on top of that please ignore and prefer #6289

crtrott:
A few more general thoughts:

- `BasicView` (or whatever we will call it in the end) should be effectively `mdspan` with `Kokkos::View` owning semantics
  - it owns the tracking object and an mdspan
  - template parameters are mdspan parameters
  - constructors could be the minimal set to support all the functionality we want to begin with
    - e.g. maybe don't have a ctor which takes individual indices but have one taking extents
    - we could even go further and say for now that there are really only two ctors: 
      - `(view_alloc, mapping, accessor)`
      - `(mdspan)`
  - maybe make `BasicView` implementation detail to start with
- `View` provides the traditional Kokkos interface
  - i.e. takes our old template args and has all the ctors
- MemorySpaceAware accessor:
  - I think we should orthogonalize instead of having one accessor which does all the memory traits things
  - the memory space awareness can take a nested accessor and forwards everything except that `access` adds the check for being called from the right place before calling the nested accessors `access` function
  - in addition to that we gonna have stuff like atomic, restrict, etc. accessors
- initially maybe don't provide any new interfaces other than the mdspan conversion and then we make careful decisions about which things we want to make public
  - medium term does `View` allow you to provide `mdspan` like parameters?
  - should `View` get all the `mdspan` properties such as typedefs?
  - Is `BasicView` something public, maybe even long term replacement for `View`?

nmm0:
closing PR for now to stop CI runs

PR #6332, opened 2023-08-03
skip team-level `for_each` tests for nhvpc

fnrizzi:
skip tests to address : https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/Kokkos/detail/Kokkos/13846/pipeline/48/

but worked when we merged the PR https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/Kokkos/detail/Kokkos/13824/pipeline/48

PR #6333, opened 2023-08-03
Run NVHPC CI only on V100

Rombur:
Since the problem with `for_each` and `NVHPC` only happens on waffle, we can force the ci to use fetnat instead.

Alternative to https://github.com/kokkos/kokkos/pull/6332


masterleinad (approved):

crtrott (approved):
I will merge this for now, but I need someone to do some testing on waffle to track this down further. Specifically I would like to test whether the opcode thing goes away if:
- you explicitly compile for AVX not Zen2 (native)
- you compile for Volta instead of Ampere
- you use a newer NVHPC

@masterleinad @Rombur

Issue #695, opened 2017-03-20
CUDA Error when capturing a class member in a lambda


ramanan:
I had an error with a Kokkos kernel when executing with CUDA. I tracked down the issue to using a view that is a member of the class directly. If I make a local copy of that view and then used it, there is no error anymore. I made a simple example program to reproduce the error. That program is below. The error can be reproduced with the latest version of kokkos and cuda/8.0 

I have a parallel_for with a lambda. When using the class member "seq" directly, it will compile fine, but will have a cuda error on execution. When copied to a local variable "lseq", it will execute properly. Do we know what is causing this error?

Thanks
Ramanan

```c++
#include <Kokkos_Core.hpp>

class MyClass {
  public:
    typedef Kokkos::View<int *> IndexType;
    IndexType seq;

  public:
    void set(int NP) {
      seq=IndexType("seq", NP);
      IndexType lseq=seq;
      Kokkos::parallel_for(NP, KOKKOS_LAMBDA(const size_t& n) {
         seq(n) = n; //This compiles fine, but will have a cuda error at runtime
      //lseq(n) = n; //This compiles and executes correctly
      } );

      int sum;
      Kokkos::parallel_reduce(NP, KOKKOS_LAMBDA(const size_t& n, int& lsum) { lsum+=lseq(n); }, sum);
      printf("rsa set %2d\n", sum);
    }
};

int main(int argc, char *argv[]){
  Kokkos::initialize();
  MyClass A;
  A.set(4);
  Kokkos::finalize();
}
```

ibaned:
I have also run into this issue multiple times. As far as I know, it is a defect (?) in the CUDA compiler, not really a problem with Kokkos. I don't know of any better short-term solution than what you pointed out, i.e. make local copies first.

My guess is that `seq(n)` gets expanded to `this->seq(n)`, so `this` is copied by value to the GPU rather than `seq`, and finally the GPU attempts to dereference the host pointer `this`. This is speculation on my part as to exactly what the failure mechanism is.

I hope that this can be fixed by NVIDIA and I wonder if it has been brought to their attention. It may also be the case that the C++ standard mandates this behavior, and that CUDA cannot be simultaneously convenient and compliant. Hopefully other Kokkos developers can confirm or deny my speculations here.

I recommend you rename this issue to "CUDA Error when capturing a class member in a lambda".

crtrott:
[...]

PR #6337, opened 2023-08-04
Fix #6336: capturing `this` implicitly is deprecated in C++20

### LINKED ISSUE #6336: Nightly GCC13 fails in TestStdAlgorithmsTeamAdjacentFind
@fnrizzi Can you take a look. The error message is [here](https://cloud.cees.ornl.gov/jenkins-ci/blue/organizations/jenkins/Kokkos-nightly/detail/Kokkos-nightly/446/pipeline)

fnrizzi:
implicit capture of 'this' via '[=]' is deprecated in C++20 , so making it explicit anyway.

### DIFF
--- a/algorithms/unit_tests/TestStdAlgorithmsTeamAdjacentFind.cpp
+++ b/algorithms/unit_tests/TestStdAlgorithmsTeamAdjacentFind.cpp
@@ -55,7 +55,7 @@ struct TestFunctorA {
       case 0: {
         const auto it = KE::adjacent_find(member, KE::cbegin(myRowViewFrom),
                                           KE::cend(myRowViewFrom));
-        Kokkos::single(Kokkos::PerTeam(member), [=, this]() {
+        Kokkos::single(Kokkos::PerTeam(member), [=, *this]() {
           m_distancesView(myRowIndex) =
               KE::distance(KE::cbegin(myRowViewFrom), it);
         });
@@ -64,7 +64,7 @@ struct TestFunctorA {
 
       case 1: {
         const auto it = KE::adjacent_find(member, myRowViewFrom);
-        Kokkos::single(Kokkos::PerTeam(member), [=, this]() {
+        Kokkos::single(Kokkos::PerTeam(member), [=, *this]() {
           m_distancesView(myRowIndex) =
               KE::distance(KE::begin(myRowViewFrom), it);
         });
@@ -75,7 +75,7 @@ struct TestFunctorA {
         const auto it =
             KE::adjacent_find(member, KE::cbegin(myRowViewFrom),
                               KE::cend(myRowViewFrom), m_binaryPred);
-        Kokkos::single(Kokkos::PerTeam(member), [=, this]() {
+        Kokkos::single(Kokkos::PerTeam(member), [=, *this]() {
           m_distancesView(myRowIndex) =
               KE::distance(KE::cbegin(myRowViewFrom), it);
         });
@@ -84,7 +84,7 @@ struct TestFunctorA {
 
       case 3: {
         const auto it = KE::adjacent_find(member, myRowViewFrom, m_binaryPred);
-        Kokkos::single(Kokkos::PerTeam(member), [=, this]() {
+        Kokkos::single(Kokkos::PerTeam(member), [=, *this]() {
           m_distancesView(myRowIndex) =
               KE::distance(KE::begin(myRowViewFrom), it);
         });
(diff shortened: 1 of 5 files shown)


fnrizzi:
retest this please

Rombur (approved):

algorithms/unit_tests/TestStdAlgorithmsTeamAdjacentFind.cpp
```diff
@@ -55,7 +55,7 @@ struct TestFunctorA {
       case 0: {
         const auto it = KE::adjacent_find(member, KE::cbegin(myRowViewFrom),
                                           KE::cend(myRowViewFrom));
-        Kokkos::single(Kokkos::PerTeam(member), [=]() {
+        Kokkos::single(Kokkos::PerTeam(member), [=, this]() {
```
masterleinad, line 58: Don't you want
```suggestion
        Kokkos::single(Kokkos::PerTeam(member), [=, *this]() {
```
here and below instead?

masterleinad: The `this` pointer points to host data so I find it at the very least confusing to use `this` instead of `*this` here.

Rombur: Windows is not happy with `this` anyway.

fnrizzi: yes both right, i will fix it

masterleinad (approved):

crtrott (approved):

algorithms/unit_tests/TestStdAlgorithmsTeamAdjacentFind.cpp
```diff
@@ -55,7 +55,7 @@ struct TestFunctorA {
       case 0: {
         const auto it = KE::adjacent_find(member, KE::cbegin(myRowViewFrom),
                                           KE::cend(myRowViewFrom));
-        Kokkos::single(Kokkos::PerTeam(member), [=]() {
+        Kokkos::single(Kokkos::PerTeam(member), [=, *this]() {
```
nliber, line 58: `[=]` and `[=, *this]` are not equivalent.

`[=]` and `[=, this]` are equivalent, as `[=]` effectively captured the object by reference.

nliber (changes requested):

algorithms/unit_tests/TestStdAlgorithmsTeamAdjacentFind.cpp
fnrizzi: Yes but I think what we want is *this because _this_ points to the host object and because of https://github.com/kokkos/kokkos/issues/695

PhilMiller: Per @nmm0 [here](https://github.com/DARMA-tasking/vt/issues/925#issuecomment-1259849888), it's possible to end up with warnings coming from compilers passed `-std=c++17` about restating defaults with an explicit capture clause. I don't recall how that got resolved there, but Nic might.

PhilMiller: Also, isn't one of those the expansion of `KOKKOS_CLASS_LAMBDA`, modulo `__host__ __device__` annotations?

fnrizzi: Not sure if the Kokkos macros changes based on the standard. 
But I think solving the warning can be done via ifdeffing it, however we need to resolve if we want to use *this or this. 
I was going off the wiki (see [here](https://kokkos.github.io/kokkos-core-wiki/ProgrammingGuide/HierarchicalParallelism.html#restricting-execution-to-a-single-executor)) but then Daniel's comments made me realize this is not totally correct

masterleinad: Since this uses an explicit functor, rather than a lambda, we copy the whole class to the device anyway and it should be OK to just use `[=, this]` (but this raises warnings for C++17 with some compilers and we have seen problems in particular with MSVC). If changing the functor to a lambda, I could see that using `[=, this]` is problematic and I was preferring `[=, *this]` for consistency (and portability in that case). Ultimately, this is just a test, and copying any of the class members is cheap anyway.

crtrott: Technically in this place it doesn't matter: It must be correct with [=, *this] but if it is it's ok to just capture via [&] because it's a nested lambda. Since this is all tests I don't have a preference and I am happy to go with [=, *this] which is the closest to what the test intended.

PR #6339, opened 2023-08-06
WIP: trying fix for type inconsistency in sort API with bounds when calling BinSort

fnrizzi:


nmm0:
I am closing this as there hasn't been any activity for a long time, if this gets revived we can re-open.

PR #5011, opened 2022-05-11
Remove KOKKOS_ENABLE_POSIX_MEMALIGN

masterleinad:
`KOKKOS_ENABLE_POSIX_MEMALIGN` is only mentioned in
```
commit 8802c19fdfd492b70288daa8849d0d48f96441eb
Author: crtrott <crtrott@sandia.gov>
Date:   Wed Dec 2 12:27:33 2015 -0700

    Core: Add option KOKKOS_ENABLE_PERFORMANCE_POSIX_MEMALIGN
    This option is off by default, if defined we will use posix
    memalign for HostSpace, but it fails to allocate often.
    This only effects the new view implementation.  
```
in the whole `git` history. Hence, it seems to be untested and unused. We should consider removing it (or actually using and testing it).

PR #4889, opened 2022-03-21
Add CI run with HBWSpace

masterleinad:
Apparently, HBWSpace is not tested in our current CI and (therefore unsurprisingly) broken.

PR #6341, opened 2023-08-07
Use std::aligned_alloc for allocations

masterleinad:
Related to https://github.com/kokkos/kokkos/pull/5011. It seems that we don't need all these different allocation mechanisms anymore and can use `std::aligned_alloc`(https://en.cppreference.com/w/cpp/memory/c/aligned_alloc) instead. Also, there were some problems with `INTEL_MM_ALLOC` reported that don't appear anymore when using `std::malloc` instead.
Note that `MSVC` is unable to handle aligned allocations of any kind and provides `_aligned_malloc` (to be freed with `_aligned_free`) instead.

This pull request deprecates all other allocation variants for `HostSpace` to only use `aligned_malloc`.

`HBWSpace` is untested and broken, see https://github.com/kokkos/kokkos/pull/4889. I would address it in a different pull request (maybe that one).

### DIFF
--- a/core/src/Kokkos_HostSpace.hpp
+++ b/core/src/Kokkos_HostSpace.hpp
@@ -64,25 +64,27 @@ class HostSpace {
   //! This memory space preferred device_type
   using device_type = Kokkos::Device<execution_space, memory_space>;
 
-  /**\brief  Default memory space instance */
-  HostSpace();
+  HostSpace()                     = default;
   HostSpace(HostSpace&& rhs)      = default;
   HostSpace(const HostSpace& rhs) = default;
   HostSpace& operator=(HostSpace&&) = default;
   HostSpace& operator=(const HostSpace&) = default;
   ~HostSpace()                           = default;
 
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
   /**\brief  Non-default memory space instance to choose allocation mechansim,
    * if available */
 
-  enum AllocationMechanism {
+  enum KOKKOS_DEPRECATED AllocationMechanism {
     STD_MALLOC,
     POSIX_MEMALIGN,
     POSIX_MMAP,
     INTEL_MM_ALLOC
   };
 
+  KOKKOS_DEPRECATED
   explicit HostSpace(const AllocationMechanism&);
+#endif
 
   /**\brief  Allocate untracked memory in the space */
   void* allocate(const size_t arg_alloc_size) const;
@@ -114,7 +116,9 @@ class HostSpace {
   static constexpr const char* name() { return m_name; }
 
  private:
-  AllocationMechanism m_alloc_mech;
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  KOKKOS_DEPRECATED AllocationMechanism m_alloc_mech = STD_MALLOC;
+#endif
   static constexpr const char* m_name = "Host";
   friend class Kokkos::Impl::SharedAllocationRecord<Kokkos::HostSpace, void>;
 };
@@ -187,7 +191,7 @@ class SharedAllocationRecord<Kokkos::HostSpace, void>
   static RecordBase s_root_record;
 #endif
 
-  const Kokkos::HostSpace m_space;
+  Kokkos::HostSpace m_space;
 
  protected:
   ~SharedAllocationRecord();
--- a/core/src/Kokkos_HostSpace.hpp
+++ b/core/src/Kokkos_HostSpace.hpp
@@ -116,9 +116,6 @@ class HostSpace {
   static constexpr const char* name() { return m_name; }
 
  private:
-#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
-  KOKKOS_DEPRECATED AllocationMechanism m_alloc_mech = STD_MALLOC;
-#endif
   static constexpr const char* m_name = "Host";
   friend class Kokkos::Impl::SharedAllocationRecord<Kokkos::HostSpace, void>;
 };
--- a/core/src/OpenMP/Kokkos_OpenMP_Instance.cpp
+++ b/core/src/OpenMP/Kokkos_OpenMP_Instance.cpp
@@ -256,8 +256,6 @@ void OpenMPInternal::initialize(int thread_count) {
       }
     }
 
-    OpenMP::memory_space space;
-
     // Before any other call to OMP query the maximum number of threads
     // and save the value for re-initialization unit testing.
 
--- a/core/src/impl/Kokkos_Error.cpp
+++ b/core/src/impl/Kokkos_Error.cpp
@@ -102,13 +102,6 @@ void Experimental::RawMemoryAllocationFailure::print_error_message(
   o << "  (The allocation mechanism was ";
   switch (m_mechanism) {
     case AllocationMechanism::StdMalloc: o << "standard malloc()."; break;
-#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
-    case AllocationMechanism::PosixMemAlign: o << "posix_memalign()."; break;
-    case AllocationMechanism::PosixMMap: o << "POSIX mmap()."; break;
-    case AllocationMechanism::IntelMMAlloc:
-      o << "the Intel _mm_malloc() intrinsic.";
-      break;
-#endif
     case AllocationMechanism::CudaMalloc: o << "cudaMalloc()."; break;
     case AllocationMechanism::CudaMallocManaged:
       o << "cudaMallocManaged().";
--- a/core/src/impl/Kokkos_Error.cpp
+++ b/core/src/impl/Kokkos_Error.cpp
@@ -121,6 +121,7 @@ void Experimental::RawMemoryAllocationFailure::print_error_message(
     case AllocationMechanism::SYCLMallocHost:
       o << "sycl::malloc_host().";
       break;
+    default: o << "unsupported.";
   }
   append_additional_error_information(o);
   o << ")" << std::endl;
--- a/core/src/impl/Kokkos_Error.cpp
+++ b/core/src/impl/Kokkos_Error.cpp
@@ -102,11 +102,13 @@ void Experimental::RawMemoryAllocationFailure::print_error_message(
   o << "  (The allocation mechanism was ";
   switch (m_mechanism) {
     case AllocationMechanism::StdMalloc: o << "standard malloc()."; break;
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
     case AllocationMechanism::PosixMemAlign: o << "posix_memalign()."; break;
     case AllocationMechanism::PosixMMap: o << "POSIX mmap()."; break;
     case AllocationMechanism::IntelMMAlloc:
       o << "the Intel _mm_malloc() intrinsic.";
       break;
+#endif
     case AllocationMechanism::CudaMalloc: o << "cudaMalloc()."; break;
     case AllocationMechanism::CudaMallocManaged:
       o << "cudaMallocManaged().";
--- a/core/src/impl/Kokkos_Error.hpp
+++ b/core/src/impl/Kokkos_Error.hpp
@@ -113,11 +113,6 @@ class RawMemoryAllocationFailure : public std::bad_alloc {
   };
   enum class AllocationMechanism {
     StdMalloc,
-#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
-    PosixMemAlign KOKKOS_DEPRECATED,
-    PosixMMap KOKKOS_DEPRECATED,
-    IntelMMAlloc KOKKOS_DEPRECATED,
-#endif
     CudaMalloc,
     CudaMallocManaged,
     CudaHostAlloc,
--- a/core/src/impl/Kokkos_Error.hpp
+++ b/core/src/impl/Kokkos_Error.hpp
@@ -113,6 +113,11 @@ class RawMemoryAllocationFailure : public std::bad_alloc {
   };
   enum class AllocationMechanism {
     StdMalloc,
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+    PosixMemAlign KOKKOS_DEPRECATED,
+    PosixMMap KOKKOS_DEPRECATED,
+    IntelMMAlloc KOKKOS_DEPRECATED,
+#endif
     CudaMalloc,
     CudaMallocManaged,
     CudaHostAlloc,
--- a/core/src/impl/Kokkos_Error.hpp
+++ b/core/src/impl/Kokkos_Error.hpp
@@ -113,9 +113,11 @@ class RawMemoryAllocationFailure : public std::bad_alloc {
   };
   enum class AllocationMechanism {
     StdMalloc,
-    PosixMemAlign,
-    PosixMMap,
-    IntelMMAlloc,
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+    PosixMemAlign KOKKOS_DEPRECATED,
+    PosixMMap KOKKOS_DEPRECATED,
+    IntelMMAlloc KOKKOS_DEPRECATED,
+#endif
     CudaMalloc,
     CudaMallocManaged,
     CudaHostAlloc,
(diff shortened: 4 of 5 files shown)


core/src/Kokkos_HostSpace.hpp
```diff
@@ -64,25 +64,28 @@ class HostSpace {
...
-  enum AllocationMechanism {
+  enum KOKKOS_DEPRECATED AllocationMechanism {
     STD_MALLOC,
     POSIX_MEMALIGN,
     POSIX_MMAP,
     INTEL_MM_ALLOC
   };
 
+  KOKKOS_DEPRECATED
   explicit HostSpace(const AllocationMechanism&);
```
PhilMiller, line 85: It looks like with `DEPRECATED_CODE_4` enabled, it would no longer be possible to default-construct an instance of `HostSpace`, since the default constructor declaration is removed above, and there's no default argument for the parameter this constructor overload takes.

masterleinad: Fixed.

cz4rs (approved):
Looks good to me :+1:

masterleinad:
@https://github.com/zippylab should be able to comment some more on the issue found with using `_mm_malloc`.

zippylab:
> https://github.com/zippylab should be able to comment some more on the issue found with using `_mm_malloc`.

We had been seeing crashes of the XGC application on the Sunspot system (Aurora test system) coming from this line in HostSpace.cpp (in Kokkos::HostSpace::impl_deallocate):
```
(gdb) list
212           void *alloc_ptr = *(reinterpret_cast<void **>(arg_alloc_ptr) - 1);
/213           free(alloc_ptr);
214         }
215     #if defined(KOKKOS_ENABLE_INTEL_MM_ALLOC)
216         else if (m_alloc_mech == INTEL_MM_ALLOC) {
217           _mm_free(arg_alloc_ptr);  <<<<<<<<<<<<<<<<<<HERE<<<<<<
218         }
219     #endif
220       }
221     }
(gdb)
``` 

in a part of XGC that is copying particle data from device to host memory (using Cabana::SoA and Cabana::AoSoA).

After rebuilding Kokkos, Cabana, and XGC with the changes in this PR, that failure mode of XGC went away, and it ran past the point where the SEGV was happening before in the same test case.

core/src/impl/Kokkos_HostSpace.cpp
```diff
@@ -119,30 +87,13 @@ void *HostSpace::impl_allocate(
...
-    }
-#if defined(KOKKOS_ENABLE_INTEL_MM_ALLOC)
-    else if (m_alloc_mech == INTEL_MM_ALLOC) {
-      ptr = _mm_malloc(arg_alloc_size, alignment);
-    }
+    // Over-allocate to to multiple of alignment
+    size_t size_alignment_multiple =
+        (arg_alloc_size + alignment - 1) / alignment * alignment;
+#ifdef KOKKOS_COMPILER_MSVC
+    ptr = _aligned_malloc(size_alignment_multiple, alignment);
```
dalg24, line 94: Does `_aligned_malloc` also require the size argument to be a multiple of the alignment?

core/src/OpenMP/Kokkos_OpenMP_Instance.cpp
```diff
@@ -256,8 +256,6 @@ void OpenMPInternal::initialize(int thread_count) {
       }
     }
 
-    OpenMP::memory_space space;
```
dalg24, line 259: What's the deal with that change?

core/src/Kokkos_HostSpace.hpp
```diff
@@ -187,7 +191,7 @@ class SharedAllocationRecord<Kokkos::HostSpace, void>
   static RecordBase s_root_record;
 #endif
 
-  const Kokkos::HostSpace m_space;
+  Kokkos::HostSpace m_space;
```
dalg24, line 194: Although I agree with that change in general, why is it part of this PR?

dalg24 (commented):
What about `RawMemoryAllocationFailure::AllocationMechanism`?
Should we not deprecate enumerators that are not supported?

core/src/impl/Kokkos_HostSpace.cpp
masterleinad: I couldn't find information about that but preferred doing it anyway for consistency.

core/src/OpenMP/Kokkos_OpenMP_Instance.cpp
masterleinad: It's not used which was revealed by defaulting the Host space constructor.

core/src/Kokkos_HostSpace.hpp
masterleinad: The compiler complained about const default constructed objects. This appeared better than explicitly brace-initializing.

core/src/impl/Kokkos_HostSpace.cpp
crtrott: Does MSVC not support aligned_alloc?

core/src/impl/Kokkos_Error.cpp
```diff
@@ -102,11 +102,13 @@ void Experimental::RawMemoryAllocationFailure::print_error_message(
   o << "  (The allocation mechanism was ";
   switch (m_mechanism) {
     case AllocationMechanism::StdMalloc: o << "standard malloc()."; break;
```
crtrott, line 104: How does this work if KOKKOS_ENABLE_DEPRECATED_CODE_4 is off? Isn't the entire AllocationMechanism thing removed if DEPRECATED_CODE_4 is not enabled?

crtrott (changes requested):
I don't think the print stuff works?

core/src/impl/Kokkos_HostSpace.cpp
PhilMiller: MSVC indeed does not support `aligned_alloc` as noted [on CppReference.com](https://en.cppreference.com/w/cpp/memory/c/aligned_alloc)

It would be nice to be able to use `operator new(size_alignment_multiple, std::align_val_t(alignment))` instead, avoiding the conditional, but then we'd have to store the alignment to pass to the corresponding `operator delete`

core/src/impl/Kokkos_Error.cpp
masterleinad: You are looking at the wrong place, the relevant is `core/src/impl/Kokkos_Error.hpp`, not `core/src/Kokkos_HostSpace.hpp`. Both files contain `AllocationMechanism` enums but they are different. Note that we test with `KOKKOS_DEPRECATED_CODE_4=OFF` in our CI, see https://github.com/kokkos/kokkos/blob/1a3ea28f6e97b4c9dd2c8ceed53ad58ed5f94dfe/.jenkins#L56-L92 for example.

masterleinad:
Note that we get a couple of warnings (if we enable KOKKOS_DEPRECATED_CODE_4 and deprecation warnings; here with warnings as errors)
```
/tmp/kokkos/core/src/impl/Kokkos_Error.cpp: In member function 'void Kokkos::Experimental::RawMemoryAllocationFailure::print_error_message(std::ostream&) const':
/tmp/kokkos/core/src/impl/Kokkos_Error.cpp:106:31: error: 'Kokkos::Experimental::RawMemoryAllocationFailure::AllocationMechanism::PosixMemAlign' is deprecated [-Werror=deprecated-declarations]
  106 |     case AllocationMechanism::PosixMemAlign: o << "posix_memalign()."; break;
      |                               ^~~~~~~~~~~~~
In file included from /tmp/kokkos/core/src/impl/Kokkos_Error.cpp:28:
/tmp/kokkos/core/src/impl/Kokkos_Error.hpp:117:5: note: declared here
  117 |     PosixMemAlign KOKKOS_DEPRECATED,
      |     ^~~~~~~~~~~~~
/tmp/kokkos/core/src/impl/Kokkos_Error.cpp:106:31: error: 'Kokkos::Experimental::RawMemoryAllocationFailure::AllocationMechanism::PosixMemAlign' is deprecated [-Werror=deprecated-declarations]
  106 |     case AllocationMechanism::PosixMemAlign: o << "posix_memalign()."; break;
      |                               ^~~~~~~~~~~~~
/tmp/kokkos/core/src/impl/Kokkos_Error.hpp:117:5: note: declared here
  117 |     PosixMemAlign KOKKOS_DEPRECATED,
      |     ^~~~~~~~~~~~~
/tmp/kokkos/core/src/impl/Kokkos_Error.cpp:107:31: error: 'Kokkos::Experimental::RawMemoryAllocationFailure::AllocationMechanism::PosixMMap' is deprecated [-Werror=deprecated-declarations]
  107 |     case AllocationMechanism::PosixMMap: o << "POSIX mmap()."; break;
      |                               ^~~~~~~~~
/tmp/kokkos/core/src/impl/Kokkos_Error.hpp:118:5: note: declared here
  118 |     PosixMMap KOKKOS_DEPRECATED,
      |     ^~~~~~~~~
/tmp/kokkos/core/src/impl/Kokkos_Error.cpp:107:31: error: 'Kokkos::Experimental::RawMemoryAllocationFailure::AllocationMechanism::PosixMMap' is deprecated [-Werror=deprecated-declarations]
  107 |     case AllocationMechanism::PosixMMap: o << "POSIX mmap()."; break;
      |                               ^~~~~~~~~
/tmp/kokkos/core/src/impl/Kokkos_Error.hpp:118:5: note: declared here
  118 |     PosixMMap KOKKOS_DEPRECATED,
      |     ^~~~~~~~~
/tmp/kokkos/core/src/impl/Kokkos_Error.cpp:108:31: error: 'Kokkos::Experimental::RawMemoryAllocationFailure::AllocationMechanism::IntelMMAlloc' is deprecated [-Werror=deprecated-declarations]
  108 |     case AllocationMechanism::IntelMMAlloc:
      |                               ^~~~~~~~~~~~
/tmp/kokkos/core/src/impl/Kokkos_Error.hpp:119:5: note: declared here
  119 |     IntelMMAlloc KOKKOS_DEPRECATED,
      |     ^~~~~~~~~~~~
/tmp/kokkos/core/src/impl/Kokkos_Error.cpp:108:31: error: 'Kokkos::Experimental::RawMemoryAllocationFailure::AllocationMechanism::IntelMMAlloc' is deprecated [-Werror=deprecated-declarations]
  108 |     case AllocationMechanism::IntelMMAlloc:
      |                               ^~~~~~~~~~~~
/tmp/kokkos/core/src/impl/Kokkos_Error.hpp:119:5: note: declared here
  119 |     IntelMMAlloc KOKKOS_DEPRECATED,
      |     ^~~~~~~~~~~~
cc1plus: all warnings being treated as errors
make[2]: *** [core/src/CMakeFiles/kokkoscore.dir/impl/Kokkos_Error.cpp.o] Error 1
/tmp/kokkos/core/src/impl/Kokkos_HostSpace.cpp:59:60: error: 'AllocationMechanism' is deprecated [-Werror=deprecated-declarations]
   59 | HostSpace::HostSpace(const HostSpace::AllocationMechanism &) : HostSpace() {}
In file included from /tmp/kokkos/core/src/impl/Kokkos_HostSpace.cpp:49:
/tmp/kokkos/core/src/Kokkos_HostSpace.hpp:78:26: note: declared here
   78 |   enum KOKKOS_DEPRECATED AllocationMechanism {
      |                          ^~~~~~~~~~~~~~~~~~~
```
but they all originate from "*.cc" files.

core/src/impl/Kokkos_HostSpace.cpp
```diff
@@ -119,30 +87,13 @@ void *HostSpace::impl_allocate(
...
-        address -= sizeof(void *);
-        *reinterpret_cast<void **>(address) = alloc_ptr;
-      }
-    }
-#if defined(KOKKOS_ENABLE_INTEL_MM_ALLOC)
-    else if (m_alloc_mech == INTEL_MM_ALLOC) {
-      ptr = _mm_malloc(arg_alloc_size, alignment);
-    }
+    // Over-allocate to multiple of alignment
+    size_t size_alignment_multiple =
```
dalg24, line 91: `size_alignment_multiple -> size`

core/src/impl/Kokkos_Error.cpp
```diff
@@ -102,11 +102,13 @@ void Experimental::RawMemoryAllocationFailure::print_error_message(
...
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
     case AllocationMechanism::PosixMemAlign: o << "posix_memalign()."; break;
     case AllocationMechanism::PosixMMap: o << "POSIX mmap()."; break;
     case AllocationMechanism::IntelMMAlloc:
       o << "the Intel _mm_malloc() intrinsic.";
       break;
+#endif
```
dalg24, line 111: Does that make sense to preserve this when we just ignore the alloc mechanism while constructing the host memory space?

core/src/OpenMP/Kokkos_OpenMP_Instance.cpp
dalg24: It is not clear to me how defaulting the default constructor changes things but this is fine.

core/src/impl/Kokkos_HostSpace.cpp
```diff
@@ -55,41 +55,9 @@
...
-    : m_alloc_mech(HostSpace::STD_MALLOC) {
-  if (arg_alloc_mech == STD_MALLOC) {
-    m_alloc_mech = HostSpace::STD_MALLOC;
-  }
-#if defined(KOKKOS_ENABLE_INTEL_MM_ALLOC)
-  else if (arg_alloc_mech == HostSpace::INTEL_MM_ALLOC) {
-    m_alloc_mech = HostSpace::INTEL_MM_ALLOC;
-  }
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+HostSpace::HostSpace(const HostSpace::AllocationMechanism &) : HostSpace() {}
```
dalg24, line 59: Why is it ok to just ignore the argument without warning?

dalg24 (commented):
What about `HBWSpace`?

dalg24:
> > What about `HBWSpace`?
> 
> `HBWSpace` is untested and broken, see #4889. I would address it in a different pull request (maybe that one).

Add a note in the PR description that you intentionally did not update HBWSpace.

core/src/impl/Kokkos_HostSpace.cpp
masterleinad: You get compile-time warnings using this constructor and ultimately the allocation mechanism is opaque to the user anyway.

masterleinad:
> Add a note in the PR description that you intentionally did not update HBWSpace.

Here you go!

core/src/impl/Kokkos_HostSpace.cpp
masterleinad: I don't have a strong opinion, changed in [9373c47](https://github.com/kokkos/kokkos/pull/6341/commits/9373c475814357eee29fbddc8cc2920800b47a9c).

core/src/impl/Kokkos_Error.cpp
dalg24: `RawMemoryAllocationFailure` is defined in the `Experimental::` namespace so it is fair game for us to change.  Considering the warning I think we should get rid of it.

dalg24: Also preserving the error message for an allocation mechanism that we don't honor anyway makes little sense.

masterleinad: fa90d1cfaaa91121b2ef1f2588296a1512d532a2

core/src/impl/Kokkos_Error.hpp
```diff
@@ -113,9 +113,6 @@ class RawMemoryAllocationFailure : public std::bad_alloc {
   };
   enum class AllocationMechanism {
     StdMalloc,
-    PosixMemAlign,
-    PosixMMap,
-    IntelMMAlloc,
```
dalg24, line 118: I had commented on an individual (last?) commit.
No do not remove, deprecate them.

dalg24: it is not about me being happy or not, it is about not potentially breaking user code.

core/src/impl/Kokkos_Error.cpp
crtrott: Ok that seems somewhat unfortunate naming but ok.

crtrott (approved):

barracuda156:
@masterleinad @dalg24 @crtrott Unfortunately, this does not work for all platforms: https://github.com/kokkos/kokkos/issues/6367

PR #6342, opened 2023-08-08
Disable default oneDPL support in Trilinos

masterleinad:
Corresponds to https://github.com/trilinos/Trilinos/pull/12112. We noticed in https://github.com/trilinos/Trilinos/issues/12016 that `oneDPL` is properly set up as a `Trilinos/TriBITS` dependency. 
The easiest fix is to simply disable support by default similar to what we are doing for `ROCM` and `HIP`.


Rombur (approved):

ndellingwood (approved):

dalg24:
Ignoring NVHPC failure.

Issue #6344, opened 2023-08-09
Cuda instance will not compile after PR 6299 (`==` in signature)


romintomasetti:
The following signatures are incorrect.

https://github.com/kokkos/kokkos/blob/11c6cf74bfda55fdf318682c4d234924ef5ca4d7/core/src/Cuda/Kokkos_Cuda_Instance.hpp#L487-L500

It would give:
```c++
/kokkos/core/src/Cuda/Kokkos_Cuda_Instance.hpp(490): error: expected a ")"
                                           cudaStream_t hStream == nullptr) const {
                                                                ^
```

I guess `cudaStream_t hStream = nullptr` is needed instead.

Introduced in https://github.com/kokkos/kokkos/commit/ba6b4d9d1c23dbbfeee1ef5d90bba157d30d7f0c by @tcclevenger. The error only happens if `KOKKOS_ENABLE_IMPL_CUDA_MALLOC_ASYNC` is enabled and `CUDART_VERSION >= 11020` (this is our case).

Related to:
- #6299.

dalg24:
You are welcome to open a PR to fix the default arguments.

@ndellingwood @tcclevenger is there no coverage for `CUDA_MALLOC_ASYNC` in the nightlies?
No test means no support.

ndellingwood:
> @ndellingwood @tcclevenger is there no coverage for `CUDA_MALLOC_ASYNC` in the nightlies?

@dalg24 I suppose not, I was not aware of the option `KOKKOS_ENABLE_IMPL_CUDA_MALLOC_ASYNC` and the need to explicitly enable it for coverate. Will add nightly tests to cover this case

tcclevenger:
@romintomasetti Since this was a quick fix, I went ahead and made a PR. https://github.com/kokkos/kokkos/pull/6346

ndellingwood:
@dalg24 @tcclevenger nightly job added to test `Kokkos_ENABLE_IMPL_CUDA_MALLOC_ASYNC=ON` with cuda/11.2.2, reproduced this error

PhilMiller:
@ndellingwood and others who weren't aware of this non-default option, see #4228 for details of why it's optional and disabled by default.

PR #6346, opened 2023-08-09
Fix cudaAPI wrapper errors for CUDA_MALLOC_ASYNC

tcclevenger:
Fixes build error uncovered in https://github.com/kokkos/kokkos/issues/6344. Uncovered an additional error (typo in input arg).

I verified the following config on Weaver V100:
```
Modules:
cuda11.8, gcc/9.3.0, cmake/3.25.1

CMake env: 
cmake -DKokkos_ENABLE_IMPL_CUDA_MALLOC_ASYNC=ON \
      -DCMAKE_CXX_COMPILER=$KOKKOS_DIR/bin/nvcc_wrapper \
      -DKokkos_ENABLE_SERIAL=ON \
      -DKokkos_ENABLE_CUDA=ON \
      -DKokkos_ARCH_VOLTA70=ON \
      -DKokkos_ENABLE_TESTS=ON \
      -DCMAKE_BUILD_TYPE:STRING=Debug
      $KOKKOS_DIR

Test command:
ctest -R Cuda
``` 
Build fails on develop, but this branch builds successfully and all tests run and pass.

Closes https://github.com/kokkos/kokkos/issues/6344.

### DIFF
--- a/core/src/Cuda/Kokkos_CudaSpace.cpp
+++ b/core/src/Cuda/Kokkos_CudaSpace.cpp
@@ -141,7 +141,7 @@ CudaUVMSpace::CudaUVMSpace() : m_device(Kokkos::Cuda().cuda_device()) {}
 
 CudaHostPinnedSpace::CudaHostPinnedSpace() {}
 
-int memory_threshold_g = 40000;  // 40 kB
+size_t memory_threshold_g = 40000;  // 40 kB
 
 //==============================================================================
 // <editor-fold desc="allocate()"> {{{1
@@ -172,7 +172,7 @@ void *impl_allocate_common(const Cuda &exec_space, const char *arg_label,
 #error CUDART_VERSION undefined!
 #elif (defined(KOKKOS_ENABLE_IMPL_CUDA_MALLOC_ASYNC) && CUDART_VERSION >= 11020)
   cudaError_t error_code;
-  if (arg_alloc_size >= (size_t)memory_threshold_g) {
+  if (arg_alloc_size >= memory_threshold_g) {
     if (exec_space_provided) {
       error_code =
           exec_space.impl_internal_space_instance()->cuda_malloc_async_wrapper(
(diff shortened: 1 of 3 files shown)


masterleinad (changes requested):
Please enable a build in our CI to use this code path.

masterleinad (approved):

crtrott (approved):

core/src/Cuda/Kokkos_CudaSpace.cpp
```diff
@@ -172,7 +172,7 @@ void *impl_allocate_common(const Cuda &exec_space, const char *arg_label,
 #error CUDART_VERSION undefined!
 #elif (defined(KOKKOS_ENABLE_IMPL_CUDA_MALLOC_ASYNC) && CUDART_VERSION >= 11020)
   cudaError_t error_code;
-  if (arg_alloc_size >= memory_threshold_g) {
+  if (arg_alloc_size >= (size_t)memory_threshold_g) {
```
dalg24, line 175: Considering that this and L350 below are the only use of `memory_threshold_g`, please change its definition from `int` to `size_t` instead of casting.

core/src/Cuda/Kokkos_Cuda_Instance.hpp
```diff
@@ -487,16 +487,16 @@ class CudaInternal {
 #if (defined(KOKKOS_ENABLE_IMPL_CUDA_MALLOC_ASYNC) && CUDART_VERSION >= 11020)
   template <bool setCudaDevice = true>
   cudaError_t cuda_malloc_async_wrapper(void** devPtr, size_t size,
-                                        cudaStream_t hStream == nullptr) const {
+                                        cudaStream_t hStream = nullptr) const {
     if constexpr (setCudaDevice) set_cuda_device();
-    return cudaMallocAsync(devPtr, size, get_input_stream(stream));
+    return cudaMallocAsync(devPtr, size, get_input_stream(hStream));
```
dalg24, line 492: what did the "H" stand for?

tcclevenger: I used `hStream` to match the cuda API, but looking at the documentation (this is in the ["Stream Ordered Memory Allocator" module](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY__POOLS.html#group__CUDART__MEMORY__POOLS)) it's not obvious to me exactly why they use both `stream` and `hStream` there, especially since the only function which uses `stream` refers to the input as `hStream` in the description. Maybe this is a typo?

But looking in the ["Stream Management" module](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__STREAM.html#group__CUDART__STREAM), `hStream` inputs are described as "Handle to the stream to be queried" where `stream` inputs are described "The stream to query".

dalg24:
HIP builds timed out

dalg24 (approved):

PR #6347, opened 2023-08-09
SIMD: add generator constructors

ldh4:
Related to #5674 

This PR adds a generator constructor for all simd types and mask types:

- `template <class G> simd(G&& gen)`

Generator constructors for simd allow constructing simd objects using expressions without needing to manually set a value for each vector lane. For internal implementation purpose, generator constructors have been used to implement generic fallback methods for few operators where simd intrinsics aren't adequately provided to implement such operations (i.e. Arithmetic shift right for AVX2 with int64_t).

### DIFF
--- a/simd/src/Kokkos_SIMD_NEON.hpp
+++ b/simd/src/Kokkos_SIMD_NEON.hpp
@@ -86,13 +86,13 @@ class neon_mask<Derived, 64> {
                 std::is_invocable_r_v<value_type, G,
                                       std::integral_constant<std::size_t, 0>>,
                 bool> = false>
-  KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION constexpr explicit simd(
+  KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION constexpr explicit neon_mask(
       G&& gen) noexcept {
-    m_value = vsetq_lane_f64(
+    m_value = vsetq_lane_u64(
         (gen(std::integral_constant<std::size_t, 0>()) ? 0xFFFFFFFFFFFFFFFFULL
                                                        : 0),
         m_value, 0);
-    m_value = vsetq_lane_f64(
+    m_value = vsetq_lane_u64(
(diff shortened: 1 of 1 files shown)


masterleinad:
Can you please elaborate a bit more on why we need this?

ldh4:
@masterleinad Just updated the PR description to provide a bit more on where generator constructors are typically used, but I 
 believe the main motivation for having this task is to fill in missing features to stay as consistent as possible with [P1928R6](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/p1928r6.pdf).

simd/src/Kokkos_SIMD_AVX2.hpp
```diff
@@ -73,6 +73,18 @@ class simd_mask<double, simd_abi::avx2_fixed_size<4>> {
...
   }
+  template <class G,
+            std::enable_if_t<
+                std::is_invocable_r_v<value_type, G,
+                                      std::integral_constant<std::size_t, 0>>,
+                bool> = false>
+  KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION constexpr explicit simd_mask(
+      G&& gen) noexcept
+      : m_value(_mm256_castsi256_pd(_mm256_setr_epi64x(
+            -std::int64_t(gen(std::integral_constant<std::size_t, 0>())),
```
masterleinad, line 84: Why is it correct to negate the argument here?

masterleinad (approved):

simd/src/Kokkos_SIMD_AVX2.hpp
ldh4: It's to set all bits in the lane to 1 when the mask is set to true (in this case, to produce `0xFFFFFFFFFFFFFFFFULL`). I guess this imposes an underlying assumption that setting the mask to be `true` is strictly equivalent to setting it to be `1` rather than any non-zero value.

crtrott (approved):

PR #6200, opened 2023-06-09
team-level std algos: part 1

fnrizzi:
Co-authored-by: Cezary Skrzyński
Co-authored-by: Jakub Strzebonski

## Content 

Add team-level support for:

- `Kokkos_AdjacentFind.hpp`
- `Kokkos_Count.hpp`
- `Kokkos_CountIf.hpp`
- `Kokkos_ForEach.hpp`
- `Kokkos_ForEachN.hpp`
- `Kokkos_LexicographicalCompare.hpp`
- `Kokkos_Mismatch.hpp`

### Public API has now two distinct sections as follows 
```cpp
//
// overload set accepting execution space (as before)
//
template <class ExecutionSpace, /* other templates */>
std::enable_if_t< ::Kokkos::is_execution_space<ExecutionSpace>::value, IteratorType2>
<func_name>(const ExecutionSpace& ex, /*args*/)

// other overloads accepting execution space

------ 

//
// overload set accepting a team handle 
// (all team overloads do not accept a label) 
//
template <class TeamHandleType, /* other templates */>
KOKKOS_FUNCTION
std::enable_if_t<is_team_handle<TeamHandleType>::value>
<func_name>(const TeamHandleType& teamHandle, /*args*/)

// all team level overloads 
```

<br>

## Comments
- this can be merged independently from other others

- corresponding wiki documentation is being tracked [here](https://github.com/kokkos/kokkos-core-wiki/issues/379)

- this comes from breaking apart https://github.com/kokkos/kokkos/pull/5150

PR #5150, opened 2022-06-25
team impl of std algorithms 

fnrizzi:
# Add team-level support for std algorithms

**Note**: even if this might seem a big PR to review, the impl is not majorly changing things since it follows closely the existing impl for the execution space one. 

### Public API has two distinct sections as follows 
```cpp
//
// overload set accepting execution space (as before)
//
template <class ExecutionSpace, class IteratorType1, class IteratorType2>
std::enable_if_t< ::Kokkos::is_execution_space<ExecutionSpace>::value, IteratorType2>
swap_ranges(const ExecutionSpace& ex, IteratorType1 first1, IteratorType1 last1, IteratorType2 first2)

// all exespace overloads 

------ 

//
// overload set accepting a team handle 
// (all team overloads do not accept a label) 
//
template <class TeamHandleType, class IteratorType1, class IteratorType2>
KOKKOS_FUNCTION
std::enable_if_t<is_team_handle<TeamHandleType>::value, IteratorType2>
swap_ranges(const TeamHandleType& teamHandle, IteratorType1 first1, IteratorType1 last1, IteratorType2 first2)

// all team level overloads 
```

### All tests follow a similar structure: 
```cpp
  // -----------------------------------------------
  // 1. prepare data
  // -----------------------------------------------
  // rank-2 view is created and filled in a way that is algo-specific
  // ...

  // -----------------------------------------------
  // 2. launch kokkos kernel
  // -----------------------------------------------
  // to test the team level, we use a Team policy with one team per row 
  using space_t = Kokkos::DefaultExecutionSpace;
  Kokkos::TeamPolicy<space_t> policy(numTeams, Kokkos::AUTO());
  TestFunctorA fnc(/*whatever args needed for a test*/);
  Kokkos::parallel_for(policy, fnc);

  // -----------------------------------------------
  // 3. check correctness
  // -----------------------------------------------
  // in most cases we check results are correct by comparing against the corresponding std algorithm 
  // in some cases, for simple ones like fill(), we do a "manual" check without using the std algorithm 
```

## Caveats/comments

1. For some implementations, we resort to serial because we need #5635 (once this is done, we can just change the impl easily)
  
2. using CTAD and `if constexpr`, where feasible and with some moderation 

3. `exclusive_scan`, `transform_exclusive_scan` do not support custom value types (those without neutral element) because when we wrap them we get error like (tb addressed in a separate PR):
```
[...]

PR #6350, opened 2023-08-10
team-level std algos: part 12

fnrizzi:
Co-authored-by: Cezary Skrzyński
Co-authored-by: Jakub Strzebonski

## Content 

Team-level implementation and tests for:

- `Kokkos_ExclusiveScan.hpp`
- `Kokkos_TransformExclusiveScan.hpp`

## Technical comments

- the change `class` -> `typename` and the enable if defaulted in the template parameters comes from [this request](https://github.com/kokkos/kokkos/pull/6200#issuecomment-1626406816)

- the code duplication is intentional and agreed upon after the long discussion in [here](https://github.com/kokkos/kokkos/pull/6200)

- tests are run for both contiguous views and strided views

## Other comments

- all team-level PRs are tracked in: https://github.com/kokkos/kokkos/issues/6352

- this comes from breaking apart https://github.com/kokkos/kokkos/pull/5150

## Possible conflicts

Implementation-wise, this has not conflict with other team-level PRs so it could be merged independently. 
However, a trivial conflict in the CMakeLists inside the unit test occurs if another of the team-level PRs is merged before this because each team-level PR adds test using a specific "label". Resolving this is trivial. 

## Associated wiki PRs

- `Kokkos_ExclusiveScan.hpp`, `Kokkos_TransformExclusiveScan.hpp`: see [wiki PR](https://github.com/kokkos/kokkos-core-wiki/pull/440)

IMPORTANT: wiki PRs are not complete yet but 99% there 

The issue tracking all *wiki PRs* is [here](https://github.com/kokkos/kokkos-core-wiki/issues/379)


algorithms/src/std_algorithms/Kokkos_ExclusiveScan.hpp
```diff
@@ -23,38 +23,46 @@
...
+// overload set accepting execution space
+//
+
 // overload set 1
 template <class ExecutionSpace, class InputIteratorType,
           class OutputIteratorType, class ValueType>
 std::enable_if_t<::Kokkos::Experimental::Impl::are_iterators<
-                     InputIteratorType, OutputIteratorType>::value,
+                     InputIteratorType, OutputIteratorType>::value&& ::Kokkos::
+                     is_execution_space<ExecutionSpace>::value,
```
nliber, line 35: That is weird formatting...

Does it get any better if you use the `_v` versions instead of `::value` versions?

(Anyway, this is minor.)

algorithms/src/std_algorithms/Kokkos_ExclusiveScan.hpp
```diff
@@ -23,38 +23,46 @@
...
                  OutputIteratorType>
 exclusive_scan(const ExecutionSpace& ex, InputIteratorType first,
                InputIteratorType last, OutputIteratorType first_dest,
                ValueType init_value) {
   static_assert(std::is_move_constructible<ValueType>::value,
                 "ValueType must be move constructible.");
-  return Impl::exclusive_scan_default_op_impl(
+  return Impl::exclusive_scan_default_op_exespace_impl(
```
nliber, line 42: While you assert that `ValueType` must be move constructible, `init_value` is passed to `exclusive_scan_default_op_exespace_impl` by copy constructing it.

(This kind of thing happens all over)

nliber (changes requested):
Note:  I haven't finished reviewing this.

fnrizzi:
@nliber i actually did not finish this , was still draft but thanks for the comments 

re the move constructible, it is because the std requires that see e.g. https://en.cppreference.com/w/cpp/algorithm/exclusive_scan

nliber:
> @nliber i actually did not finish this , was still draft but thanks for the comments
> 
> re the move constructible, it is because the std requires that see e.g. https://en.cppreference.com/w/cpp/algorithm/exclusive_scan

@fnrizzi But this implementation requires it to be copy constructible, as we pass it by (named) value.

We can tighten the implementation to move constructible by passing it as `std::move(init_value)` instead of `init_value`, or even `static_cast<ValueType&&>(init_value)` (the latter has far fewer template instantiations of `std::move`, so better compiler performance, and it works in this case because `init_value` is passed to us by value, so `ValueType` is not deduced to a reference type).

fnrizzi:
@nliber yeah totally right, I thought this by value/move things was fixed already. I will use your suggestion.

fnrizzi:
yes ok i follow, so what you say is: 
- if we impose the move constructible constraint, then we use use move internally and we shuld change how we pass it to `exclusive_scan_default_op_exespace_impl`
- if we drop the move constructilbe constraint, then the current impl is fine 

Right?

nliber:
> yes i agree with that. So what you say is:
> 
> * if we impose the move constructible constraint, then we shuld change how we pass it to `exclusive_scan_default_op_exespace_impl`
> * if we drop the move constructilbe constraint, then the current implementation is ok
> 
> Right?

Correct.

> 
> Btw, just to clarify, the fact i am asserting it is move constuctible at this level is not because i use that here but because inside the impl we use that. makes sense?

Yes.  And that is why the C++ standard has those requirements.  It is actually worse, because those requirements are really on values, not types.  Example 1: an uninitialized `int` is not `MoveConstructible`, even though `int` is.  Example 2:  `std::set<float>` is legal as long as none of the values stored are things like NaNs.

> > Because the algorithm API takes it by value, users can pass it in either by copying it (if it is in a named variable),
> 
> yes but if our API allows that, doesn't that mean that the type has to be copy constructible?

No.  It is up to the user how they want to pass it.  Take this sample:

```c++
struct MoveConstructibleOnly {
    MoveConstructibleOnly() = default;

    MoveConstructibleOnly(MoveConstructibleOnly&&) = default;
};

inline auto MCOFactory() { return MoveConstructibleOnly(); }

template<typename T>
void ByValue(T) {}

int main()
{
    ByValue(MoveConstructibleOnly());
    ByValue(MCOFactory());

    MoveConstructibleOnly mco;
    ByValue(std::move(mco));
    ByValue(static_cast<MoveConstructibleOnly&&>(mco));
    // ByValue(mco);   // fails to compile

    return 0;
}
```

It doesn't matter to us that `ByValue(mco)` fails to compile, as long as we aren't trying to perform that operation.


> I can remove the move constr constraint if it makes sense to do so - I did not have quite enough time yet to go into detail about all the various fine-grained requirements for all algorithms, so if something is missing or imprecise, it is definitely possible and something to revise

I think we should keep it (do as the std algorithms do), and just make sure we got it right.

algorithms/src/std_algorithms/Kokkos_ExclusiveScan.hpp
fnrizzi: i updated it and seems to me it does not improve

algorithms/src/std_algorithms/Kokkos_ExclusiveScan.hpp
fnrizzi: I think i address this in https://github.com/kokkos/kokkos/pull/6350/commits/138f4854bb222596661297650186c73876dae40d

nliber (approved):

algorithms/unit_tests/CMakeLists.txt
```diff
@@ -153,6 +153,18 @@ foreach(Tag Threads;Serial;OpenMP;Cuda;HPX;HIP;SYCL;OpenMPTarget)
...
     endforeach()
 
+    # ------------------------------------------
+    # std team P
+    # ------------------------------------------
+    set(STDALGO_TEAM_SOURCES_P)
+    foreach(Name
+  StdAlgorithmsCommon
+  StdAlgorithmsTeamExclusiveScan
+  StdAlgorithmsTeamTransformExclusiveScan
```
masterleinad, line 163: The indentation looks weird. You are using tabs anywhere else.

algorithms/unit_tests/TestStdAlgorithmsTeamExclusiveScan.cpp
```diff
@@ -0,0 +1,251 @@
...
+    auto rowDest   = Kokkos::subview(stdDestView, i, Kokkos::ALL());
+    auto initValue = initValuesView_h(i);
+
+    ASSERT_TRUE(intraTeamSentinelView_h(i));
+
+#if defined(__GNUC__) && __GNUC__ == 8
+#define exclusive_scan testing_exclusive_scan
+#else
+#define exclusive_scan std::exclusive_scan
+#endif
```
masterleinad, line 197: This should check the standard library and not the compiler.

fnrizzi: Thanks, yes this  is leftover from before , I will change it

algorithms/unit_tests/CMakeLists.txt
fnrizzi: that was a side effect: i lost all my settings on the sandia machine, so I had to use an editor that i don't use regularly :)
fixed now

algorithms/unit_tests/TestStdAlgorithmsTeamExclusiveScan.cpp
fnrizzi: fixed

masterleinad (approved):

Issue #6349, opened 2023-08-10
CI failing for fedora g++ debug due to format-truncation=


fnrizzi:
https://github.com/kokkos/kokkos/actions/runs/5819571007/job/15782009841?pr=6256

```
/__w/kokkos/kokkos/core/src/impl/Kokkos_SharedAlloc.cpp: In static member function 'static void Kokkos::Impl::SharedAllocationRecord<void, void>::print_host_accessible_records(std::ostream&, const char*, const Kokkos::Impl::SharedAllocationRecord<void, void>*, bool)':
/__w/kokkos/kokkos/core/src/impl/Kokkos_SharedAlloc.cpp:345:1: error: '%s' directive output may be truncated writing up to 247 bytes into a region of size 232 [-Werror=format-truncation=]
  345 | } /* namespace Kokkos */
      | ^
/__w/kokkos/kokkos/core/src/impl/Kokkos_SharedAlloc.cpp:326:15: note: 'snprintf' output 26 or more bytes (assuming 273) into a destination of size 256
  326 |       snprintf(buffer, 256, format_string, space_name,
      |       ~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  327 |                reinterpret_cast<uintptr_t>(r->data()), r->size(),
      |                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  328 |                r->m_alloc_ptr->m_label);
      |                ~~~~~~~~~~~~~~~~~~~~~~~~
/__w/kokkos/kokkos/core/src/impl/Kokkos_SharedAlloc.cpp:345:1: error: '%s' directive output may be truncated writing up to 247 bytes into a region of size 125 [-Werror=format-truncation=]
  345 | } /* namespace Kokkos */
      | ^
/__w/kokkos/kokkos/core/src/impl/Kokkos_SharedAlloc.cpp:305:15: note: 'snprintf' output 133 or more bytes (assuming 380) into a destination of size 256
  305 |       snprintf(buffer, 256, format_string, space_name,
      |       ~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
  306 |                reinterpret_cast<uintptr_t>(r),
      |                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
[...]

PR #6354, opened 2023-08-10
Fix -Wformat-truncation warnings in CI

masterleinad:
This attempts to fix https://github.com/kokkos/kokkos/issues/6349. 
Instead of using a fixed-size buffer, this pull request ~~makes sure that we resize the buffer to be at least big enough to hold the full formatted output.~~ directly uses `operator<<`.

Trying enabling `-Wformat-truncation` locally also flagged `Kokkos_Profiling.cpp`, so this is simplified using fewer `C` functions.
```
/tmp/kokkos/core/src/impl/Kokkos_Profiling.cpp: In function 'void Kokkos::Tools::initialize(const std::string&)':
/tmp/kokkos/core/src/impl/Kokkos_Profiling.cpp:633:52: error: 'snprintf' output may be truncated before the last format character [-Werror=format-truncation=]
  633 |   snprintf(envProfileCopy.get(), envProfileLen, "%s", envProfileLibrary);
      |                                                    ^
/tmp/kokkos/core/src/impl/Kokkos_Profiling.cpp:633:11: note: 'snprintf' output 1 or more bytes (assuming 2) into a destination of size 1
  633 |   snprintf(envProfileCopy.get(), envProfileLen, "%s", envProfileLibrary);
      |   ~~~~~~~~^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

### DIFF
--- a/core/src/impl/Kokkos_SharedAlloc.cpp
+++ b/core/src/impl/Kokkos_SharedAlloc.cpp
@@ -282,7 +282,7 @@ SharedAllocationRecord<void, void>* SharedAllocationRecord<
 // FIXME GCC warns that we truncate the output below.
 #ifdef KOKKOS_COMPILER_GNU
 #pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wformat-truncation"
+#pragma GCC diagnostic ignored "-Wformat-truncation="
 #endif
 
 void SharedAllocationRecord<void, void>::print_host_accessible_records(
--- a/core/src/impl/Kokkos_SharedAlloc.cpp
+++ b/core/src/impl/Kokkos_SharedAlloc.cpp
@@ -278,6 +278,13 @@ SharedAllocationRecord<void, void>* SharedAllocationRecord<
 }
 
 #ifdef KOKKOS_ENABLE_DEBUG
+
+// FIXME GCC warns that we truncate the output below.
+#ifdef KOKKOS_COMPILER_GNU
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wformat-truncation"
+#endif
+
 void SharedAllocationRecord<void, void>::print_host_accessible_records(
     std::ostream& s, const char* const space_name,
     const SharedAllocationRecord* const root, const bool detail) {
@@ -331,6 +338,11 @@ void SharedAllocationRecord<void, void>::print_host_accessible_records(
     }
   }
 }
+
+#ifdef KOKKOS_COMPILER_GNU
+#pragma GCC diagnostic pop
+#endif
+
 #else
 void SharedAllocationRecord<void, void>::print_host_accessible_records(
     std::ostream&, const char* const, const SharedAllocationRecord* const,
--- a/core/src/impl/Kokkos_SharedAlloc.cpp
+++ b/core/src/impl/Kokkos_SharedAlloc.cpp
@@ -278,13 +278,6 @@ SharedAllocationRecord<void, void>* SharedAllocationRecord<
 }
 
 #ifdef KOKKOS_ENABLE_DEBUG
-
-// FIXME GCC warns that we truncate the output below.
-#ifdef KOKKOS_COMPILER_GNU
-#pragma GCC diagnostic push
-#pragma GCC diagnostic ignored "-Wformat-truncation="
-#endif
-
 void SharedAllocationRecord<void, void>::print_host_accessible_records(
     std::ostream& s, const char* const space_name,
     const SharedAllocationRecord* const root, const bool detail) {
@@ -292,57 +285,23 @@ void SharedAllocationRecord<void, void>::print_host_accessible_records(
   // allocation.
   const SharedAllocationRecord<void, void>* r = root->m_next;
 
-  char buffer[256];
-
   if (detail) {
     while (r != root) {
-      // Formatting dependent on sizeof(uintptr_t)
-      const char* format_string;
-
-      if (sizeof(uintptr_t) == sizeof(unsigned long)) {
-        format_string =
-            "%s addr( 0x%.12lx ) list( 0x%.12lx 0x%.12lx ) extent[ 0x%.12lx + "
-            "%.8ld ] count(%d) dealloc(0x%.12lx) %s\n";
-      } else if (sizeof(uintptr_t) == sizeof(unsigned long long)) {
-        format_string =
-            "%s addr( 0x%.12llx ) list( 0x%.12llx 0x%.12llx ) extent[ "
-            "0x%.12llx + %.8ld ] count(%d) dealloc(0x%.12llx) %s\n";
-      }
-
-      snprintf(buffer, 256, format_string, space_name,
-               reinterpret_cast<uintptr_t>(r),
-               reinterpret_cast<uintptr_t>(r->m_prev),
-               reinterpret_cast<uintptr_t>(r->m_next),
-               reinterpret_cast<uintptr_t>(r->m_alloc_ptr), r->m_alloc_size,
-               r->use_count(), reinterpret_cast<uintptr_t>(r->m_dealloc),
-               r->m_alloc_ptr->m_label);
-      s << buffer;
+      s << space_name << " addr( " << r << " ) list ( " << r->m_prev << ' '
+        << r->m_next << " ) extent[ " << r->m_alloc_ptr << " + "
+        << r->m_alloc_size << " ] count(" << r->use_count() << ") dealloc("
+        << reinterpret_cast<void*>(r->m_dealloc) << ") "
+        << r->m_alloc_ptr->m_label << '\n';
       r = r->m_next;
     }
   } else {
     while (r != root) {
-      // Formatting dependent on sizeof(uintptr_t)
-      const char* format_string;
-
-      if (sizeof(uintptr_t) == sizeof(unsigned long)) {
-        format_string = "%s [ 0x%.12lx + %ld ] %s\n";
-      } else if (sizeof(uintptr_t) == sizeof(unsigned long long)) {
-        format_string = "%s [ 0x%.12llx + %ld ] %s\n";
-      }
-
-      snprintf(buffer, 256, format_string, space_name,
-               reinterpret_cast<uintptr_t>(r->data()), r->size(),
-               r->m_alloc_ptr->m_label);
-      s << buffer;
+      s << space_name << " [ " << r->data() << " + " << r->size() << " ] "
+        << r->m_alloc_ptr->m_label << '\n';
       r = r->m_next;
     }
   }
 }
-
-#ifdef KOKKOS_COMPILER_GNU
-#pragma GCC diagnostic pop
-#endif
-
 #else
 void SharedAllocationRecord<void, void>::print_host_accessible_records(
     std::ostream&, const char* const, const SharedAllocationRecord* const,
--- a/core/src/impl/Kokkos_SharedAlloc.cpp
+++ b/core/src/impl/Kokkos_SharedAlloc.cpp
@@ -19,6 +19,7 @@
 #endif
 
 #include <Kokkos_Core.hpp>
+#include <iomanip>
 
 namespace Kokkos {
 namespace Impl {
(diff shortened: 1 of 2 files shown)


core/src/impl/Kokkos_SharedAlloc.cpp
```diff
@@ -285,50 +285,54 @@ void SharedAllocationRecord<void, void>::print_host_accessible_records(
...
-      snprintf(buffer, 256, format_string, space_name,
-               reinterpret_cast<uintptr_t>(r->data()), r->size(),
-               r->m_alloc_ptr->m_label);
-      s << buffer;
-      r = r->m_next;
+  std::vector<char> buffer(256);
+
+  while (r != root) {
+    if (detail) {
+      auto printf_lambda = [&](char* buf, std::size_t buf_size) {
```
dalg24, line 292: This code was already bad and it feels even worse now.
C-style string manipulation sprinkled with immediately invoked lambda within another lambda that captures everything...

core/src/impl/Kokkos_SharedAlloc.cpp
```diff
@@ -285,50 +285,54 @@ void SharedAllocationRecord<void, void>::print_host_accessible_records(
...
+        const char* const format_string = [&] {
```
dalg24, line 293: What is being captured?

dalg24 (commented):
Changes to `Kokkos_Profiling.cpp` are fine.

core/unit_test/TestSharedAlloc.hpp
```diff
@@ -82,7 +82,8 @@ void test_shared_alloc() {
     // Sanity check for the whole set of allocation records to which this record
     // belongs.
     RecordBase::is_sane(r[0]);
-    // RecordMemS::print_records( std::cout, s, true );
+    [[maybe_unused]] std::stringstream dummy;
+    RecordMemS::print_records(dummy, s, true);
```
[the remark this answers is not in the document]
masterleinad: No, the warning appeared when compiling `Kokkos_SharedAlloc.cpp`. This is the only place where the function is possibly used and it makes sense to me to actually do that. On the other hand, we don't want more output to `stdout` in the tests so I just let it write into a `std::stringstream` and don't care about its contents at this point.

core/src/impl/Kokkos_SharedAlloc.cpp
masterleinad: To me, this is better but then I also don't care that much about the particular form of initializing `format_string`. I changed to stay closer to the current form now.

fnrizzi:
IMO this is a good short-term solution to get rid of the warnings, but possibly like Damien said long term needs to be revisited for improvment

core/src/impl/Kokkos_SharedAlloc.cpp
```diff
@@ -285,50 +285,55 @@ void SharedAllocationRecord<void, void>::print_host_accessible_records(
...
-      r = r->m_next;
+        if (sizeof(uintptr_t) == sizeof(unsigned long)) {
+          format_string =
+              "%s addr( 0x%.12lx ) list( 0x%.12lx 0x%.12lx ) extent[ 0x%.12lx "
+              "+ %.8ld ] count(%d) dealloc(0x%.12lx) %s\n";
+        } else if (sizeof(uintptr_t) == sizeof(unsigned long long))
+          format_string =
+              "%s addr( 0x%.12llx ) list( 0x%.12llx 0x%.12llx ) extent[ "
+              "0x%.12llx + %.8ld ] count(%d) dealloc(0x%.12llx) %s\n";
+        return snprintf(buf, buf_size, format_string, space_name,
```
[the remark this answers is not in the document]
masterleinad: Because we want to know how many characters to allocate when we first call the lambda with a `nullptr`.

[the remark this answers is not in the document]
fnrizzi: ah right, sorry missed it

fnrizzi (dismissed):
i think i am good with it as a temporary fix for the CI issue

core/src/impl/Kokkos_SharedAlloc.cpp
```diff
@@ -285,50 +285,55 @@ void SharedAllocationRecord<void, void>::print_host_accessible_records(
...
-      // Formatting dependent on sizeof(uintptr_t)
-      const char* format_string;
-
-      if (sizeof(uintptr_t) == sizeof(unsigned long)) {
-        format_string = "%s [ 0x%.12lx + %ld ] %s\n";
-      } else if (sizeof(uintptr_t) == sizeof(unsigned long long)) {
-        format_string = "%s [ 0x%.12llx + %ld ] %s\n";
-      }
+  while (r != root) {
+    if (detail) {
```
ldh4, line 291: Is `detail` updated somewhere during the execution of the while loop?
Seeing that the if-else blocks are largely the same other than `format_strings` within each lambda, I wonder if this condition should be placed within lambda for ternary operator and consolidate these if-else blocks.

masterleinad: `detail` is not changed. I tried to combine both code paths but ultimately it seemed best to group the format strings for `detail` and non-`detail` separately together with the `snprintf` command. Feel free to have a go if you have a better idea. We could also remove/deprecate the `detail` path or something like that but it's not clear to me how much effort it's worth spending on it at this point. Ultimately, this pull request is only for getting the `GitHub` CI running again.

core/src/impl/Kokkos_SharedAlloc.cpp
nmm0, line 304: Honestly I'd rather switch to C++ string manipulation rather than using `snprintf` here. Apparently the `-Wformat-truncation` warning can occur differently depending on whatever heuristic the compiler is using (it can even depend on optimization levels according to the gcc manual). If you think that should be done in a separate PR so as to get CI working, I'm good either way

ldh4 (approved):
Agreed that this is fine as a short-term fix to get the CI moving again.

core/unit_test/TestSharedAlloc.hpp
```diff
@@ -82,7 +82,8 @@ void test_shared_alloc() {
     // Sanity check for the whole set of allocation records to which this record
     // belongs.
     RecordBase::is_sane(r[0]);
-    // RecordMemS::print_records( std::cout, s, true );
+    [[maybe_unused]] std::stringstream dummy;
```
dalg24, line 85: The `[[maybe_unused]]` attribute is unnecessary.
I would prefer to revert these changes.  They don't belong to a PR that claims to be a hotfix for the CI

core/src/impl/Kokkos_SharedAlloc.cpp
dalg24, line 1: Prefer push/pop pragma to disable the warning in this PR and a proper refactor as a follow up

masterleinad: Here you go!

nmm0 (approved):
👍 We should follow up with C++ string processing as mentioned in the meeting

dalg24 (approved):

dalg24:
It looks like warnings are still emitted

masterleinad:
> It looks like warnings are still emitted

It works in https://godbolt.org/z/3PceP84vx.

masterleinad:
I decided to just use `operator<<` instead. Let's see if that finds acceptance.
Before:
```
Host addr( 0x600002be51f0 ) list( 0x0001009fa1c0 0x600002be5180 ) extent[ 0x0001495f7840 + 00009728 ] count(120) dealloc(0x000100939090) test_1199
Host addr( 0x600002be5180 ) list( 0x600002be51f0 0x600002be5110 ) extent[ 0x0001495f5040 + 00009720 ] count(120) dealloc(0x000100939090) test_1198
Host addr( 0x600002be5110 ) list( 0x600002be5180 0x600002be50a0 ) extent[ 0x0001495f2840 + 00009712 ] count(120) dealloc(0x000100939090) test_1197
Host addr( 0x600002be50a0 ) list( 0x600002be5110 0x600002be5030 ) extent[ 0x0001495f0040 + 00009704 ] count(120) dealloc(0x000100939090) test_1196

```
After:
```
Host addr( 0x6000009911f0 ) list ( 0x1009f21c0 0x600000991180 ) extent[ 0x140df7840 = 9728 ] count(120) dealloc(1) test_1199
Host addr( 0x600000991180 ) list ( 0x6000009911f0 0x600000991110 ) extent[ 0x140df5040 = 9720 ] count(120) dealloc(1) test_1198
Host addr( 0x600000991110 ) list ( 0x600000991180 0x6000009910a0 ) extent[ 0x140df2840 = 9712 ] count(120) dealloc(1) test_1197
Host addr( 0x6000009910a0 ) list ( 0x600000991110 0x600000991030 ) extent[ 0x140df0040 = 9704 ] count(120) dealloc(1) test_1196
```

nmm0 (approved):
Looks good to me! :)

masterleinad:
We have a bunch of `Cuda` CI builds that run out of memory but the relevant `GitHub` CI builds don't show the warning anymore.

core/src/impl/Kokkos_SharedAlloc.cpp
```diff
@@ -285,48 +285,19 @@ void SharedAllocationRecord<void, void>::print_host_accessible_records(
...
-               reinterpret_cast<uintptr_t>(r->m_next),
-               reinterpret_cast<uintptr_t>(r->m_alloc_ptr), r->m_alloc_size,
-               r->use_count(), reinterpret_cast<uintptr_t>(r->m_dealloc),
-               r->m_alloc_ptr->m_label);
-      s << buffer;
+      s << space_name << " addr( " << r << " ) list ( " << r->m_prev << ' '
+        << r->m_next << " ) extent[ " << r->m_alloc_ptr << " + "
+        << r->m_alloc_size << " ] count(" << r->use_count() << ") dealloc("
+        << reinterpret_cast<void*>(r->m_dealloc) << ") "
+        << r->m_alloc_ptr->m_label << '\n';
```
dalg24, line 294: Please comment about the formatting.  Did you compare before and after?
Any reason why you drop the precision?  You could have used `std::setw(12)` or whatnot.

masterleinad: See https://github.com/kokkos/kokkos/pull/6354#issuecomment-1681347629 to me that looked close enough.

masterleinad: Please have another look. We now have
```
Host addr( 0x600002be51f0 ) list( 0x0001009fa1c0 0x600002be5180 ) extent[ 0x0001495f7840 + 00009728 ] count(120) dealloc(0x000100939090) test_1199
Host addr( 0x600002be5180 ) list( 0x600002be51f0 0x600002be5110 ) extent[ 0x0001495f5040 + 00009720 ] count(120) dealloc(0x000100939090) test_1198
Host addr( 0x600002be5110 ) list( 0x600002be5180 0x600002be50a0 ) extent[ 0x0001495f2840 + 00009712 ] count(120) dealloc(0x000100939090) test_1197
Host addr( 0x600002be50a0 ) list( 0x600002be5110 0x600002be5030 ) extent[ 0x0001495f0040 + 00009704 ] count(120) dealloc(0x000100939090) test_1196
```
before and
```
Host addr( 0x600001e4c3c0 ) list ( 0x00010294dfb0 0x600001e4c360 ) extent[ 0x0001465f6a40 + 00009728 ] count(120) dealloc(0x0001028f026c) test_1199
Host addr( 0x600001e4c360 ) list ( 0x600001e4c3c0 0x600001e4c300 ) extent[ 0x0001465f4240 + 00009720 ] count(120) dealloc(0x0001028f026c) test_1198
Host addr( 0x600001e4c300 ) list ( 0x600001e4c360 0x600001e4c2a0 ) extent[ 0x0001465f1a40 + 00009712 ] count(120) dealloc(0x0001028f026c) test_1197
Host addr( 0x600001e4c2a0 ) list ( 0x600001e4c300 0x600001e4c240 ) extent[ 0x0001465ef240 + 00009704 ] count(120) dealloc(0x0001028f026c) test_1196
```
after

core/src/impl/Kokkos_SharedAlloc.cpp
```diff
@@ -285,22 +286,28 @@ void SharedAllocationRecord<void, void>::print_host_accessible_records(
   // allocation.
   const SharedAllocationRecord<void, void>* r = root->m_next;
 
+#define KOKKOS_PAD_HEX(ptr)                              \
+  "0x" << std::hex << std::setw(12) << std::setfill('0') \
+       << reinterpret_cast<uintptr_t>(ptr)
```
dalg24, line 291: The issue with the current implementation is that the `setfill`, `hex` and `dec` manipulators are sticky.
So if someone prints the record on `std::cout` then subsequent integers that would be put into it will be written has hexes.

masterleinad: My bad; fixed!

dalg24 (approved):
Good enough.

If you push any other change, please do `KOKKOS_PAD_HEX -> KOKKOS_IMPL_FORMAT_HEX` and add a const qualifier when you save the current flags.  The way everything is piped is not super readable but I can live with it.

masterleinad:
Retest this please.

nmm0 (approved):
:+1:

Issue #6355, opened 2023-08-11
Kokkos::Array fixes and improvements


nliber:
While looking to add deduction guides to `Kokkos::Array`, I had to add unit tests, which uncovered some issues.

`Kokkos::Array<>` is not an array but has two tag types`contiguous` and `strided` which are used in other `Kokkos::Array`s.
`Kokkos::Array<T, N>` is an owning array similar to `std::array` without iterator support.
`Kokkos::Array<T, KOKKOS_INVALID_INDEX, Kokkos::Array<>::contiguous>` is a **_non-owning_** array (a span-like type) around contiguous storage.  The assignment operators have funky semantics (`a=b` only copies all of `b` if `a.size() == b.size()`).
`Kokkos::Array<T, KOKKOS_INVALID_INDEX, Kokkos::Array<>::strided>` is a **_non-owning_** array (a span-like type) around strided storage.  The  assignment operators have funky semantics (`a=b` only copies all of `b` if `a.size() == b.size()`).  The assignment operators are also buggy that it is obvious they have never been used.

Q:  Should these four things be conflated into the same public type?  The only place that contiguous and strided `Kokkos::Arrays are used is in core/src/impl/Kokkos_ViewArray.hpp for defining `ViewMapping<Traits, Kokkos::Array<>>::reference_type`.  My vote would be for `Kokkos::Array` to only support owning semantics, and to deprecate/remove the third (`Proxy`) template parameter.

Q:  I can fix the bugs in the funky strided assignment operators.  Should I, or should I just remove this unused functionality?  My vote is to remove this functionality.

Another issue:  

The subscripting operators are templated on integral and enum types.  I assume (but don't know) this was done to avoid compiler warnings, such as sign/unsigned mismatches.  However, while it deliberately supports enums, it doesn't support scoped enums.  Also, unlike C array subscripting operators, it doesn't work on things convertible to `size_t`.

Q:  Should scoped enums be supported for subscripting?  The reason to do so is that it encourages people to use them over unscoped enums.  (If we intend not to support them, I'll give it a better error message.). My vote is to support scoped enums.

Q:  Should type convertible to `size_t` be supported for subscripting?  I don't see why not, and it meets user expectations (especially for new users).  My vote is to support types convertible to `size_t`.

The plan (breaking up my current PR into these):

- Add `is_scoped_enum` trait with compile time tests.
[...]

PR #6356, opened 2023-08-11
Adding is_scoped_enum & to_underlying

nliber:
Adding to `Impl`:
- C++20 compatible version of `underlying_type` & `underlying_type_t`
- C++23 `to_underlying` (marked as a `KOKKOS_INLINE_FUNCTION`)
- C++23 `is_scoped_enum` & `is_scoped_enum_v`

Q:  Should these be made public?

(Needed as part of issue #6355)

### DIFF
--- a/core/src/impl/Kokkos_Utilities.hpp
+++ b/core/src/impl/Kokkos_Utilities.hpp
@@ -79,6 +79,57 @@ template <class T>
 using remove_cvref_t = typename remove_cvref<T>::type;
 #endif
 
+#if __cplusplus >= 202002
+// C++20 version of std::underlying_type
+//
+// If T is a complete enumeration (enum) type, provides a member typedef type
+// that names the underlying type of T.
+// Otherwise, if T is not an enumeration type, there is no member type.
+// Otherwise (T is an incomplete enumeration type), the program is ill-formed.
+using std::underlying_type;
+using std::underlying_type_t;
+#else
+template <typename E, bool = std::is_enum_v<E>>
+struct underlying_type_impl {};
+
+template <typename E>
+struct underlying_type_impl<E, true> {
+  using type = std::underlying_type_t<E>;
+};
+
+template <typename E>
+using underlying_type = underlying_type_impl<E>;
+
+template <typename E>
+using underlying_type_t = typename underlying_type<E>::type;
+#endif
+
+// same as C++23 std::to_underlying but with __host__ __device__ annotations
+template <typename E>
+KOKKOS_INLINE_FUNCTION constexpr underlying_type_t<E> to_underlying(
+    E e) noexcept {
+  return static_cast<underlying_type_t<E>>(e);
+}
+
+#if defined(__cpp_lib_is_scoped_enum)
+// since C++23
+using std::is_scoped_enum;
+using std::is_scoped_enum_v;
+#else
+template <typename E, bool = std::is_enum_v<E>>
+struct is_scoped_enum_impl : std::false_type {};
+
+template <typename E>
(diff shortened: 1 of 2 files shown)


core/src/impl/Kokkos_Utilities.hpp
```diff
@@ -79,6 +79,57 @@ template <class T>
...
+struct underlying_type_impl<E, true> {
+  using type = std::underlying_type_t<E>;
+};
+
+template <typename E>
+using underlying_type = underlying_type_impl<E>;
+
+template <typename E>
+using underlying_type_t = typename underlying_type<E>::type;
+#endif
```
dalg24, line 105: It is non-obvious to me why the C++11 definition of `std::underlying_type` would not enable implementing `to_underlying` or `is_scoped_enum`.  Out of curiosity I replaced L82 by `#if 1` and your unit test builds.

ldh4:
> Q: Should these be made public?

Do you plan on exposing these `using` statements to the public kokkos namespace as well or just the newly added compatible types?
```
using std::underlying_type;
using std::underlying_type_t;
using std::is_scoped_enum;
using std::is_scoped_enum_v;
```

core/src/impl/Kokkos_Utilities.hpp
nliber: I think you are correct, but we should be able to rely on the C++20 version, especially if we end up exposing it publicly.

nliber:
> > Q: Should these be made public?
> 
> Do you plan on exposing these `using` statements to the public kokkos namespace as well or just the newly added compatible types?
> 
> ```
> using std::underlying_type;
> using std::underlying_type_t;
> using std::is_scoped_enum;
> using std::is_scoped_enum_v;
> ```

I would add all four of those to `namespace Kokkos`.

core/unit_test/TestUtilities.hpp
```diff
@@ -41,4 +41,72 @@ void test_is_specialization_of() {
...
+  static_assert(!is_scoped_enum<EnumBool>::value);
+  static_assert(!is_scoped_enum_v<EnumBool>);
+
+  static_assert(is_scoped_enum<ScopedEnum>{});
+  static_assert(is_scoped_enum<ScopedEnum>::value);
+  static_assert(is_scoped_enum_v<ScopedEnum>);
+
+  static_assert(is_scoped_enum<ScopedEnumShort>{});
+  static_assert(is_scoped_enum<ScopedEnumShort>::value);
+  static_assert(is_scoped_enum_v<ScopedEnumShort>);
```
dalg24, line 109: Add assertions for a user-defined class?

nliber: Added.

core/unit_test/TestUtilities.hpp
```diff
@@ -41,4 +41,94 @@ void test_is_specialization_of() {
...
+
+void test_is_scoped_enum() {
+  using Kokkos::Impl::is_scoped_enum;
+  using Kokkos::Impl::is_scoped_enum_v;
+
+  static_assert(!is_scoped_enum<int>{});
+  static_assert(!is_scoped_enum<int>::value);
+  static_assert(!is_scoped_enum_v<int>);
+  static_assert(
+      is_public_unambiguous_base_of_v<std::false_type, is_scoped_enum<int>>);
```
dalg24, line 100: Why do we care about this part?

dalg24 (approved):

ldh4 (approved):

dalg24:
Ignoring timeouts

PR #6357, opened 2023-08-11
Use `/usr/bin/env bash` in `nvcc_wrapper` and `kokkos_launch_compiler` for portability

msimberg:
This is useful e.g. on NixOS where `bash` is not in `/bin`. As far as I know there's no downside to using `/usr/bin/env bash` as it should be strictly more portable, but am I missing some reason for keeping `/bin/bash` hardcoded in `kokkos_launch_compiler` and `nvcc_wrapper`?

### DIFF
--- a/bin/kokkos_launch_compiler
+++ b/bin/kokkos_launch_compiler
@@ -1,4 +1,4 @@
-#!/bin/bash -e
+#!/usr/bin/env bash
 #
 #   This script allows CMAKE_CXX_COMPILER to be a standard
 #   C++ compiler and Kokkos sets RULE_LAUNCH_COMPILE and
@@ -12,6 +12,8 @@
 #   the original command. Examples of when $2 will not equal
 #   $1 are 'ar', 'cmake', etc. during the linking phase
 #
+
+set -e
 
 # emit a message about the underlying command executed
 : ${DEBUG:=0}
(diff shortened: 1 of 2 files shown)


dalg24 (approved):
Can't think of any good reason not to update the shebang to use the env utility.

msimberg:
Failures look like what is being fixed by https://github.com/kokkos/kokkos/pull/6354.

masterleinad (approved):

cz4rs:
> This is useful e.g. on NixOS where `bash` is not in `/bin`. As far as I know there's no downside to using `/usr/bin/env bash` as it should be strictly more portable, but am I missing some reason for keeping `/bin/bash` hardcoded in `kokkos_launch_compiler` and `nvcc_wrapper`?

Some distros complain about using `/usr/bin/env` - see [Fedora's guideline](https://fedoraproject.org/wiki/Script_Interpreters_(draft)) (this is only a draft!) and rather lenghty discussion [here](https://lists.fedoraproject.org/archives/list/devel@lists.fedoraproject.org/thread/WYBBLP2PVCVVJVKBDVXDQMJRQ4AT4P35/). Main issue seems to be automated dependency tracking.
AFAIK there is no hard requirement to avoid it in Fedora and openSUSE shows rpmlint warning for such scripts.

cz4rs (approved):
Distro packages can be trivially patched, so that's still :+1: for me.
As a bonus, now all the scripts in `bin` directory use the same shebang.

PR #6358, opened 2023-08-14
Initial implementation of gfx942

arghdos:


cmake/kokkos_arch.cmake
```diff
@@ -94,6 +94,9 @@ IF(Kokkos_ENABLE_HIP OR Kokkos_ENABLE_OPENMPTARGET OR Kokkos_ENABLE_OPENACC OR K
 ENDIF()
 
 # AMD archs ordered in decreasing priority of autodetection
+LIST(APPEND SUPPORTED_AMD_GPUS       MI300)
+LIST(APPEND SUPPORTED_AMD_ARCHS      AMD_GFX942)
+LIST(APPEND CORRESPONDING_AMD_FLAGS  gfx942   gfx942)
```
dalg24, line 99:
```suggestion
LIST(APPEND CORRESPONDING_AMD_FLAGS  gfx942)
```

cmake/kokkos_arch.cmake
```diff
@@ -956,6 +959,11 @@ ENDIF()
 
 FOREACH(ARCH IN LISTS SUPPORTED_AMD_ARCHS)
   IF (KOKKOS_ARCH_${ARCH})
+    STRING(REGEX MATCH "942" IS_942 ${ARCH})
+    IF(IS_90A)
```
masterleinad, line 963:
```suggestion
    IF(IS_942)
```

arghdos: Ah, my bad, messed this up in the rebase -- good catch!

dalg24:
Can you say a word about the thread fences that are being added?

arghdos:
>Can you say a word about the thread fences that are being added?

Technically, this was a violation of the memory model, because there was no guarantee that the write for the intermediate reduction values became visible before the read by the last block to do the second stage.  This never bit us because it was quite unlikely on the hardware we're running on, but... shall we say that may not always hold.

I've tested correctness and performance of several LAMMPS benchmarks, a regular dot product, and the yAx tutorial example on MI-250, and saw essentially no impact from unconditionally including it.

cmake/kokkos_arch.cmake
```diff
@@ -956,6 +959,11 @@ ENDIF()
 
 FOREACH(ARCH IN LISTS SUPPORTED_AMD_ARCHS)
   IF (KOKKOS_ARCH_${ARCH})
+    STRING(REGEX MATCH "942" IS_942 ${ARCH})
+    IF(IS_942)
+      SET(KOKKOS_ARCH_AMD_GFX942 ON)
+      BREAK()
+    ENDIF()
```
Rombur, line 966: I don't think this is necessary. The reason it's there for the other architectures is because I don't know if the user used `AMD_GFX` or `VEGA` but whatever his choice, we want to export both macros. That's why I check for the common part of the flag and then set both variables. I am 99% sure you can remove this but you should double check.

arghdos: Ah, yeah -- that tracks.

arghdos: Correct, removed in [18c88f7](https://github.com/kokkos/kokkos/pull/6358/commits/18c88f700c4348e21fe00c34445940f91d6af4c8)

masterleinad (approved):

dalg24:
CUDA failure unrelated

PR #6360, opened 2023-08-16
simd: make mask and condition unit test to check with all data types

ldh4:
Follow on from https://github.com/kokkos/kokkos/pull/6278#discussion_r1275517329.

This PR modifies `simd_mask` ops and `condition` unit tests to test with all supported data types for:

- `bool any_of(const simd_mask&)`
- `simd condition(const mask_type& mask, const simd& a, const simd& b)`

This addition to the tests ensures that `simd_mask` and `condition` are defined properly per supported simd backend.


Rombur (approved):

crtrott (approved):

crtrott:
Test failures are unrelated: NVHPC opcode, and snprintf.

PR #6361, opened 2023-08-16
Cuda: Detect device from stream for multi-GPU support

masterleinad:
Part of #6091. ~~We agreed that~~
```
Cuda(int device_id, cudaStream_t stream);
```
~~is the construction we want (without an option to let `Kokkos` manage the stream).~~

We decided to query the device id to use from the stream passed in.


core/src/Cuda/Kokkos_Cuda_Instance.hpp
```diff
@@ -159,7 +159,7 @@ class CudaInternal {
     return nullptr != m_scratchSpace && nullptr != m_scratchFlags;
   }
 
-  void initialize(cudaStream_t stream, bool manage_stream);
+  void initialize(int cuda_devie, cudaStream_t stream, bool manage_stream);
```
ldh4, line 162:
```suggestion
  void initialize(int cuda_device, cudaStream_t stream, bool manage_stream);
```

crtrott (dismissed):
I would like to see a test for this specifically I would like to see a CUDA specific test where we query the number of GPUs and create a std::thread for each GPU and then have them run independent kernels. As a sanity check I would like that test to be a kernel which uses the entire GPU and runs reasonably long (>100ms) and compare the total time with the n-threads with the individual time: i.e. making sure it overlaps. We probably need to guard that test somehow since its gonna be problematic inside of Trilinos testing or an environment which doesn't guard GPU overuse by multiple processes.

masterleinad:
This really only adds the constructor and prepares `CudaInternal` for multi-GPU support. #6091 is a functional prototype including a test.

core/unit_test/cuda/TestCuda_InterOp_StreamsMultiGPU.cpp
```diff
@@ -0,0 +1,34 @@
...
+//@HEADER
+
+#include <Kokkos_Core.hpp>
+#include <TestCuda_Category.hpp>
+#include <gtest/gtest.h>
+
+namespace Test {
+// Test Interoperability with Cuda Streams and muliple GPUs.
+TEST(cuda, raw_cuda_streams) {
+  Kokkos::ScopeGuard scope_guard;
```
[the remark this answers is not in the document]
masterleinad: This does `Kokkos::initialize/finalize`, see https://kokkos.github.io/kokkos-core-wiki/API/core/initialize_finalize/ScopeGuard.html?highlight=scopeguard#_CPPv410ScopeGuard.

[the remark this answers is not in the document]
masterleinad: I expect to replace this file in #6091. A minimal test like this was proposed by @crtrott.

core/src/Cuda/Kokkos_Cuda_Instance.cpp
```diff
@@ -893,7 +891,18 @@ Cuda::Cuda(cudaStream_t stream, bool manage_stream)
...
+}
+
+Cuda::Cuda(int device_id, cudaStream_t stream)
+    : m_space_instance(new Impl::CudaInternal, [](Impl::CudaInternal *ptr) {
+        ptr->finalize();
+        delete ptr;
+      }) {
+  Impl::CudaInternal::singleton().verify_is_initialized(
+      "Cuda instance constructor");
+  m_space_instance->initialize(device_id, stream, /*manage_stream*/ false);
```
ldh4, line 905: What happens if an invalid device id is passed in? Does one of `KOKKOS_IMPL_CUDA_SAFE_CALLs` in `initialize` correctly throws for this?
Should there be a simple upfront check against `cudaGetDeviceCount`?

masterleinad: Yeah, the error message is not great. I added a check, see fa1aaa712.

ldh4 (approved):

masterleinad:
Only HIP-ROCm-5.6-C++20 is timing out.

masterleinad:
`CUDA-12.2-NVHPC` was timing out. All other `Cuda` CI builds passed.

tcclevenger (commented):
Just a question about using the cudaAPI wrapper: Does it make sense to use them everywhere in `core/src/Cuda` (for consistency), or should we directly call cudaAPI when we want to manually set the device id (less complicated code, easier to interpret)?

~~I lean towards the former (consistency) since I think with the latter it then becomes complicated as to when you should or shouldn't use them, but I'm open to other opinions.~~

Edit: I see your comment in https://github.com/kokkos/kokkos/pull/6392 about other places that must manually set the device id and the interaction with the wrapper. We still could direct all calls through the CudaInternal::singleton() and manually set the device, but I'm not as convinced that would be my opinion.

masterleinad:
> We still could direct all calls through the CudaInternal::singleton() and manually set the device, but I'm not as convinced that would be my opinion.

I find it confusing to use the singleton (possibly with template arguments to not set the device) if it's not used. We could make the wrapper functions `static` or (`global`) but that makes it more awkward to use them where we actually have a `Cuda` instance.

core/src/Cuda/Kokkos_Cuda_Instance.cpp
```diff
@@ -851,7 +848,31 @@ Cuda::Cuda(cudaStream_t stream, Impl::ManageStream manage_stream)
       }) {
   Impl::CudaInternal::singleton().verify_is_initialized(
       "Cuda instance constructor");
-  m_space_instance->initialize(stream, static_cast<bool>(manage_stream));
+  m_space_instance->initialize(Impl::CudaInternal::singleton().m_cudaDev,
+                               stream, static_cast<bool>(manage_stream));
+}
+
+Cuda::Cuda(int device_id, cudaStream_t stream)
```
tcclevenger, line 855: Is there a way to ensure that `stream` was created on `device_id` (I'm not seeing anything obvious in CudaAPI)? I'm guessing it could cause issues if they are mismatched.

Would it make sense that this constructor only taking the device id and internally creating a stream where we can ensure it is on the correct device?

masterleinad: We discussed that we don't want to (allow to) manage the stream if we don't use the default execution space instance.

masterleinad: > Is there a way to ensure that stream was created on device_id (I'm not seeing anything obvious in CudaAPI)? I'm guessing it could cause issues if they are mismatched.

We could possibly use (not tested)
```
CUresult cuStreamGetCtx ( CUstream hStream, CUcontext* pctx );
CUresult cuCtxPushCurrent ( CUcontext ctx )
CUresult cuCtxGetDevice ( CUdevice* device )
CUresult cuCtxPopCurrent ( CUcontext* ctx )
```
but I'm not sure if it's worth it.

masterleinad: @tcclevenger I added such a check in e156d5859e36fbc3c4f84575a0c8734689c1b3eb.

core/src/Cuda/Kokkos_Cuda_Instance.cpp
```diff
@@ -370,13 +370,32 @@ void CudaInternal::fence() const {
...
+  KOKKOS_IMPL_CUDA_SAFE_CALL(cudaError_t(cuCtxGetDevice(&device_for_stream)));
+  KOKKOS_IMPL_CUDA_SAFE_CALL(cudaError_t(cuCtxPopCurrent(&context)));
+
+  if (device_for_stream != cuda_device) {
+    std::stringstream ss;
+    ss << "Error: The provided stream is associated with device "
+       << device_for_stream << " but device " << cuda_device
+       << " was requested in the execution space instance constructor!";
+    Kokkos::abort(ss.str().c_str());
+  }
```
tcclevenger, line 389: So if I'm a user and I want to use multiple devices, how am I meant to do this? If I need to set a device and then create a stream on that device and then call the new `Cuda(dev_id,stream)` to create the space, we can get all that information from just one of these arguments, so would it make since to just take a stream and query the device_id internally?

tcclevenger (approved):

core/src/Cuda/Kokkos_Cuda_Instance.cpp
```diff
@@ -796,12 +807,10 @@ Kokkos::Cuda::initialize WARNING: Cuda is allocating into UVMSpace by default
   //----------------------------------
 
   cudaStream_t singleton_stream;
-  KOKKOS_IMPL_CUDA_SAFE_CALL(
-      (Impl::CudaInternal::singleton().cuda_stream_create_wrapper(
-          &singleton_stream)));
+  KOKKOS_IMPL_CUDA_SAFE_CALL(cudaStreamCreate(&singleton_stream));
```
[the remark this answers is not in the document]
tcclevenger: I think that makes sense based on your previous explanation. Should this be done in a separate PR? I don't mind creating that PR if you don't want to.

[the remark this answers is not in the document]
masterleinad: If someone requests to split it from this pull request, I'll drop it. I would have a slight preference to get all the other related pull requests in before systematically searching for this kind of guard but ultimately I don't care much.

masterleinad:
Only `hpx` is failing with a clearly unrelated error:
```
/home/runner/work/kokkos/kokkos/hpx/libs/core/iterator_support/include/hpx/iterator_support/counting_iterator.hpp:58:53: error: no member named 'intmax_t' in namespace 'std'
                    (sizeof(Integer) >= sizeof(std::intmax_t)),
                                               ~~~~~^
/home/runner/work/kokkos/kokkos/hpx/libs/core/iterator_support/include/hpx/iterator_support/counting_iterator.hpp:61:39: error: expected ';' after alias declaration
                        std::intmax_t>,
                                      ^
                                      ;
/home/runner/work/kokkos/kokkos/hpx/libs/core/iterator_support/include/hpx/iterator_support/counting_iterator.hpp:64:46: error: no member named 'intmax_t' in namespace 'std'
                        std::ptrdiff_t, std::intmax_t>>::type::type;
                                        ~~~~~^
/home/runner/work/kokkos/kokkos/hpx/libs/core/iterator_support/include/hpx/iterator_support/counting_iterator.hpp:64:55: error: expected member name or ';' after declaration specifiers
                        std::ptrdiff_t, std::intmax_t>>::type::type;
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
4 errors generated.
```
that we now see for all pull requests.

core/src/Cuda/Kokkos_Cuda_Instance.cpp
```diff
@@ -692,14 +707,10 @@ void Cuda::impl_initialize(InitializationSettings const &settings) {
 
   const struct cudaDeviceProp &cudaProp = dev_info.m_cudaProp[cuda_device_id];
 
-  Impl::CudaInternal::m_cudaDev    = cuda_device_id;
   Impl::CudaInternal::m_deviceProp = cudaProp;
 
-  Kokkos::Impl::cuda_device_synchronize(
-      "Kokkos::CudaInternal::initialize: Fence on space initialization");
```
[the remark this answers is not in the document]
masterleinad: I didn't see a good reason to have it here. Let me revert this one anyway; it's not really relevant to multi-GPU support in particular. a07c7a2b6c9999020859a7de86004e447790d0c4

core/src/Cuda/Kokkos_Cuda_Instance.cpp
```diff
@@ -97,21 +97,21 @@ __global__ void query_cuda_kernel_arch(int *d_arch) {
 }
 
 /** Query what compute capability is actually launched to the device: */
-int cuda_kernel_arch() {
+int cuda_kernel_arch(int cuda_device) {
```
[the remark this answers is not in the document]
masterleinad: Sure. a07c7a2b6c9999020859a7de86004e447790d0c4

core/src/Cuda/Kokkos_Cuda_Instance.cpp
```diff
@@ -377,6 +377,21 @@ void CudaInternal::initialize(cudaStream_t stream, bool manage_stream) {
...
+  KOKKOS_IMPL_CUDA_SAFE_CALL(cudaSetDevice(m_cudaDev));
+
+  // FIXME_CUDA multiple devices
+  if (m_cudaDev != Cuda().cuda_device())
+    Kokkos::abort(
+        "Currently, the device id must match the device id used when Kokkos "
+        "was initialized!");
+
+  was_initialized = true;
+
```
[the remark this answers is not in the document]
masterleinad: a07c7a2b6c9999020859a7de86004e447790d0c4

core/src/Cuda/Kokkos_Cuda_Instance.cpp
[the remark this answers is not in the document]
dalg24: If you think it is not needed, then make the case for it, don't just remove it without mentioning the change.

masterleinad:
All `CUDA` builds are passing.

core/src/Cuda/Kokkos_Cuda_Instance.cpp
```diff
@@ -672,15 +685,13 @@ void Cuda::impl_initialize(InitializationSettings const &settings) {
...
   const struct cudaDeviceProp &cudaProp = dev_info.m_cudaProp[cuda_device_id];
+  Impl::CudaInternal::m_deviceProp      = cudaProp;
 
-  Impl::CudaInternal::m_cudaDev    = cuda_device_id;
-  Impl::CudaInternal::m_deviceProp = cudaProp;
-
-  Kokkos::Impl::cuda_device_synchronize(
-      "Kokkos::CudaInternal::initialize: Fence on space initialization");
+  KOKKOS_IMPL_CUDA_SAFE_CALL(cudaSetDevice(cuda_device_id));
+  KOKKOS_IMPL_CUDA_SAFE_CALL(cudaDeviceSynchronize());
```
masterleinad, line 691: I guess we could remove but it wouldn't really be related to the intent of this pull request.

tcclevenger (approved):

ldh4 (approved):

masterleinad:
Only `openmptarget.partitioning_by_args` is failing.

crtrott (approved):

core/src/Cuda/Kokkos_Cuda_Instance.cpp
```diff
@@ -377,6 +377,19 @@ void CudaInternal::initialize(cudaStream_t stream, bool manage_stream) {
     Kokkos::abort("Calling Cuda::initialize after Cuda::finalize is illegal\n");
   was_initialized = true;
 
+  // Check that the device associated with the stream matches cuda_device
+  CUcontext context;
+  KOKKOS_IMPL_CUDA_SAFE_CALL(cudaError_t(cuStreamGetCtx(stream, &context)));
+  KOKKOS_IMPL_CUDA_SAFE_CALL(cudaError_t(cuCtxPushCurrent(context)));
+  KOKKOS_IMPL_CUDA_SAFE_CALL(cudaError_t(cuCtxGetDevice(&m_cudaDev)));
+  KOKKOS_IMPL_CUDA_SAFE_CALL(cudaSetDevice(m_cudaDev));
```
crtrott, line 385: Couldn't find a better way of doing this (i.e. not using driver interface) but we should be ok as long as Kokkos::initialize was called. Otherwise there could be issues with the context not being created yet.

core/src/Cuda/Kokkos_Cuda_Instance.cpp
```diff
@@ -97,21 +97,21 @@ __global__ void query_cuda_kernel_arch(int *d_arch) {
...
-int cuda_kernel_arch() {
+int cuda_kernel_arch(int device_id) {
   int arch    = 0;
   int *d_arch = nullptr;
 
-  KOKKOS_IMPL_CUDA_SAFE_CALL((CudaInternal::singleton().cuda_malloc_wrapper(
-      reinterpret_cast<void **>(&d_arch), sizeof(int))));
-  KOKKOS_IMPL_CUDA_SAFE_CALL((CudaInternal::singleton().cuda_memcpy_wrapper(
-      d_arch, &arch, sizeof(int), cudaMemcpyDefault)));
+  KOKKOS_IMPL_CUDA_SAFE_CALL(cudaSetDevice(device_id));
```
crtrott, line 104: the singleton would query the wrong device so not call through its wrapper functions.

core/src/Cuda/Kokkos_Cuda_Instance.cpp
crtrott: can't use the wrapper here because it used to use the static m_cudaDev which isn't static anymore

PR #6368, opened 2023-08-18
Use `desul::atomic_load` to read `m_ready_count` in task queue implementations

### LINKED ISSUE #5748: HPX and OpenMP backends should not use volatile where atomic loads would do in tasking implementations
E.g. https://github.com/kokkos/kokkos/blob/5ad609661e570ba6aa7716a26a91cb67d559f8a2/core/src/OpenMP/Kokkos_OpenMP_Task.hpp#L311 and https://github.com/kokkos/kokkos/blob/5ad609661e570ba6aa7716a26a91cb67d559f8a2/core/src/HPX/Kokkos_HPX_Task.hpp#L266.

msimberg:
Fixes #5748?

I've only changed the loads for `HPX` and `OpenMP`. I don't dare touch the ones for `CUDA` and `OpenMPTarget`, but if someone says it's safe to change those similarly I'll update this. I'm also not going to touch the many other volatile loads/stores and variables (at least in this PR...).

I've made the loads `MemoryOrderAcquire` which I'm ~90% sure is ok here instead of `SeqCst`, but if you want to be on the safe side I'll make those `SeqCst` as well.

### DIFF
--- a/core/src/impl/Kokkos_TaskQueueCommon.hpp
+++ b/core/src/impl/Kokkos_TaskQueueCommon.hpp
@@ -143,14 +143,14 @@ class TaskQueueCommonMixin {
  public:
   KOKKOS_INLINE_FUNCTION
   bool is_done() const noexcept {
-    // TODO @tasking @memory_order DSH Memory order, instead of volatile
-    return (*(volatile int*)(&m_ready_count)) == 0;
+    return desul::atomic_load(&m_ready_count, desul::MemoryOrderAcquire(),
+                              desul::MemoryScopeDevice()) == 0;
   }
 
   KOKKOS_INLINE_FUNCTION
   int32_t ready_count() const noexcept {
-    // TODO @tasking @memory_order DSH Memory order, instead of volatile
-    return (*(volatile int*)(&m_ready_count));
+    return desul::atomic_load(&m_ready_count, desul::MemoryOrderAcquire(),
+                              desul::MemoryScopeDevice());
   }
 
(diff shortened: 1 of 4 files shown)


PhilMiller:
If you're doing OpenMP and HPX, you may as well do Threads as well. They should all be quite similar in this regard.

msimberg:
> If you're doing OpenMP and HPX, you may as well do Threads as well. They should all be quite similar in this regard.

It turns out `Threads` is actually a bit different. I also thought it was similar which is why I included it in the title of #5748 initially. I've removed `Threads` from the issue now. Don't get me wrong, there are still `volatile`s in other places in the `Threads` backend still, but not in the tasking implementation. I want to keep this PR as minimal as possible.

core/src/impl/Kokkos_TaskQueueCommon.hpp
dalg24, line 1: What about https://github.com/kokkos/kokkos/blob/756b7203c2bd05266a8d929b22670a93c70ef81e/core/src/impl/Kokkos_TaskQueueCommon.hpp#L208
and
https://github.com/kokkos/kokkos/blob/756b7203c2bd05266a8d929b22670a93c70ef81e/core/src/impl/Kokkos_TaskQueueCommon.hpp#L220

dalg24: These are beyond the scope this PR anyway.

dalg24 (approved):

Issue #6367, opened 2023-08-18
kokkos-devel does not build now: error: 'aligned_alloc' is not a member of 'std'; did you mean 'aligned_union'?


barracuda156:
Commit 8d8b24a1f165b56c1debe457ef8d6291bb1a8b01 has broken the build with GCC 12.3.0 for me:

```
/opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_devel_kokkos/kokkos-devel/work/kokkos-ca3687e637c12225e824ee626e69fc9fea8cebd2/core/src/impl/Kokkos_HostSpace.cpp: In member function 'void* Kokkos::HostSpace::impl_allocate(const char*, size_t, size_t, Kokkos::Tools::SpaceHandle) const':
/opt/local/var/macports/build/_opt_PPCSnowLeopardPorts_devel_kokkos/kokkos-devel/work/kokkos-ca3687e637c12225e824ee626e69fc9fea8cebd2/core/src/impl/Kokkos_HostSpace.cpp:95:16: error: 'aligned_alloc' is not a member of 'std'; did you mean 'aligned_union'?
   95 |     ptr = std::aligned_alloc(alignment, size);
      |                ^~~~~~~~~~~~~
      |                aligned_union
make[2]: *** [core/src/CMakeFiles/kokkoscore.dir/impl/Kokkos_HostSpace.cpp.o] Error 1
```

P. S. Passing `-std=c++17` does not help.

barracuda156:
I guess, a check should be done if `aligned_alloc` is available, and if not, then `posix_memalign` should be used.

jwakely:
> I guess, a check should be done if `aligned_alloc` is available, and if not, then `posix_memalign` should be used.

Or check for `__cpp_aligned_new` and use the aligned version of `operator new`, which should deal with using aligned_alloc or posix_memalign or whatever the OS supports.

masterleinad:
@barracuda156 Is `aligned_alloc` (in the global namespace) available using your toolchain?

barracuda156:
> @barracuda156 Is `aligned_alloc` (in the global namespace) available using your toolchain?

@masterleinad It is not there before macOS 10.15: https://gcc.gnu.org/bugzilla/show_bug.cgi?id=92143#c6

manuel5975p:
std::aligned_alloc should be standard since C++17: https://en.cppreference.com/w/cpp/memory/c/aligned_alloc
Did you enable -std=c++17?

[...]

PR #6369, opened 2023-08-18
Use aligned version of operator new instead of aligned_malloc

masterleinad:
Fixes https://github.com/kokkos/kokkos/issues/6367. ~~Apparently, `aligned_malloc` is not available on Mac OS X before 10.15 and I don't see much of a problem using `posix_memalign` for Mac OS X unconditionally until we don't support 10.15 anymore.~~
The aligned version of operator new seems to work better (icpc needs some non-standard include file, though.).

### DIFF
--- a/core/src/impl/Kokkos_HostSpace.cpp
+++ b/core/src/impl/Kokkos_HostSpace.cpp
@@ -46,6 +46,10 @@
 #include <sstream>
 #include <cstring>
 
+#ifdef KOKKOS_COMPILER_INTEL
+#include <aligned_new>
+#endif
+
 #include <Kokkos_HostSpace.hpp>
 #include <impl/Kokkos_Error.hpp>
 #include <Kokkos_Atomic.hpp>
@@ -86,15 +90,9 @@ void *HostSpace::impl_allocate(
 
   void *ptr = nullptr;
 
-  if (arg_alloc_size) {
-    // Over-allocate to multiple of alignment
-    size_t size = (arg_alloc_size + alignment - 1) / alignment * alignment;
-#ifdef KOKKOS_COMPILER_MSVC
-    ptr = _aligned_malloc(size, alignment);
-#else
-    ptr = std::aligned_alloc(alignment, size);
-#endif
-  }
+  if (arg_alloc_size)
+    ptr = operator new (arg_alloc_size, std::align_val_t(alignment),
+                        std::nothrow_t{});
 
   if ((ptr == nullptr) || (reinterpret_cast<uintptr_t>(ptr) == ~uintptr_t(0)) ||
       (reinterpret_cast<uintptr_t>(ptr) & alignment_mask)) {
@@ -143,11 +141,9 @@ void HostSpace::impl_deallocate(
       Kokkos::Profiling::deallocateData(arg_handle, arg_label, arg_alloc_ptr,
                                         reported_size);
     }
-#ifdef KOKKOS_COMPILER_MSVC
-    _aligned_free(arg_alloc_ptr);
-#else
-    std::free(arg_alloc_ptr);
-#endif
+    constexpr uintptr_t alignment = Kokkos::Impl::MEMORY_ALIGNMENT;
+    operator delete (arg_alloc_ptr, std::align_val_t(alignment),
+                     std::nothrow_t{});
   }
 }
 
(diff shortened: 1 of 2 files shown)


core/src/impl/Kokkos_HostSpace.cpp
```diff
@@ -91,6 +91,9 @@ void *HostSpace::impl_allocate(
     size_t size = (arg_alloc_size + alignment - 1) / alignment * alignment;
 #ifdef KOKKOS_COMPILER_MSVC
     ptr = _aligned_malloc(size, alignment);
+#elif defined(__APPLE__)
+    // FIXME_APPLE MacOS X doesn't provide std::aligned_alloc before 10.15
+    posix_memalign(&ptr, alignment, size);
```
jwakely, line 96: `posix_memalign` is undefined unless `alignment >= sizeof(void*)`. Is that a precondition of this function already, or do you need to increase the alignment if it's small?

dalg24: The alignment is always 64 as far as I can tell https://github.com/kokkos/kokkos/blob/ca3687e637c12225e824ee626e69fc9fea8cebd2/core/src/Kokkos_Macros.hpp#L190-L192

Looks like `alignment`
> must be a power of two and a multiple `of sizeof(void *)`

So I suppose we're good.

masterleinad: Unless it's user-provided as compile definition (which probably almost never happens). I added a `static_assert` so we are safe either way.

dalg24 (dismissed):
You forgot the deallocation

dalg24:
Why did you chose this approach rather than the one suggested by Jonathan https://github.com/kokkos/kokkos/issues/6367#issuecomment-1683435438

```C++
#if __cpp_aligned_new >= 201606
  ptr = new (std::align_val_t(alignment)) char[size];
#else
  // fallbacks

**Comment by masterleinad** [comment]
> You forgot the deallocation

Can you please elaborate? `posix_memalign` allocations are also deallocated using `free`.

**Comment by masterleinad** [comment]
> Why did you chose this approach rather than the one suggested by Jonathan [#6367 (comment)](https://github.com/kokkos/kokkos/issues/6367#issuecomment-1683435438)

It's not that clear to me what the fallback(s) would be whereas this is clear to me in this approach.

**Comment by dalg24** [comment]
> > You forgot the deallocation
> 
> Can you please elaborate? `posix_memalign` allocations are also deallocated using `free`.

Just me being ignorant.  Sorry about the noise.

**Comment by dalg24** [comment]
> > Why did you chose this approach rather than the one suggested by Jonathan [#6367 (comment)](https://github.com/kokkos/kokkos/issues/6367#issuecomment-1683435438)
> 
> It's not that clear to me what the fallback(s) would be whereas this is clear to me in this approach.

Same fallbacks as you have now

**Comment by barracuda156** [comment]
Thank you for working to fix this!

**Comment by masterleinad** [comment]
@barracuda156 Ca you confirm that the latest version works for you?

**Review by crtrott — APPROVED** [review]
I like this, assuming it passes testing :-)

**Comment by masterleinad** [comment]
@msimberg Can you reproduce the test faiures for the HPX CI here?

**Comment by masterleinad** [comment]
> I like this, assuming it passes testing :-)

Not quite sure why `HPX` is complaining and if special casing for it would make sense.

**Comment by barracuda156** [comment]
@masterleinad Yes, from 183c05a3ca43b3830a5d6dcb28f375cbe85687a8 it builds fine for me. Thank you!

**Review by dalg24 — APPROVED** [review]

**Review by msimberg — APPROVED** [review]

**Comment by masterleinad** [comment]
`HIP-ROCm-5.2-C++20` is timing out. Everything else is passing.

File: core/src/impl/Kokkos_HostSpace.cpp [hunk 2]
```diff
@@ -46,6 +46,10 @@
 #include <sstream>
 #include <cstring>
 
+#ifdef KOKKOS_COMPILER_INTEL
+#include <aligned_new>
```
dalg24, line 50: @masterleinad What was that header needed for?

barracuda156: For Intel I have no idea. I only built this with GCC.

masterleinad: Have a look at https://godbolt.org/z/jThT4h9Ed.

dalg24: Understood.  Thanks.

PR #6370, opened 2023-08-21
Add support for HIP Graph

Rombur:
Similar to CUDA, HIP has  a Graph library (as far as I can tell it was introduced in ROCm 4.5). The API is basically the same as the CUDA API and the new code is very close to CUDA code.

### DIFF
--- a/core/src/HIP/Kokkos_HIP_GraphNodeKernel.hpp
+++ b/core/src/HIP/Kokkos_HIP_GraphNodeKernel.hpp
@@ -1,0 +1,166 @@
+//@HEADER
+// ************************************************************************
+//
+//                        Kokkos v. 4.0
+//       Copyright (2022) National Technology & Engineering
+//               Solutions of Sandia, LLC (NTESS).
+//
+// Under the terms of Contract DE-NA0003525 with NTESS,
+// the U.S. Government retains certain rights in this software.
+//
+// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
+#ifndef KOKKOS_HIP_GRAPHNODEKERNEL_HPP
+#define KOKKOS_HIP_GRAPHNODEKERNEL_HPP
+
+#include <Kokkos_Graph_fwd.hpp>
+
+#include <impl/Kokkos_GraphImpl.hpp>
+#include <impl/Kokkos_SharedAlloc.hpp>
+
+#include <Kokkos_Parallel.hpp>
+#include <Kokkos_Parallel_Reduce.hpp>
+#include <Kokkos_PointerOwnership.hpp>
+
+#include <HIP/Kokkos_HIP_SharedAllocationRecord.hpp>
+
+namespace Kokkos {
+namespace Impl {
+
+template <class CombinedFunctorReducer, class PolicyType>
+struct PatternImplSpecializationFromTag<
+    Kokkos::ParallelReduceTag, CombinedFunctorReducer, PolicyType, Kokkos::HIP>
+    : type_identity<
+          ParallelReduce<CombinedFunctorReducer, PolicyType, Kokkos::HIP>> {};
+
+template <typename PolicyType, typename Functor, typename PatternTag,
+          typename... Args>
+class GraphNodeKernelImpl<Kokkos::HIP, PolicyType, Functor, PatternTag, Args...>
+    : public PatternImplSpecializationFromTag<PatternTag, Functor, PolicyType,
+                                              Args..., Kokkos::HIP>::type {
+ public:
+  using Policy       = PolicyType;
+  using graph_kernel = GraphNodeKernelImpl;
+  using base_t =
+      typename PatternImplSpecializationFromTag<PatternTag, Functor, Policy,
+                                                Args..., Kokkos::HIP>::type;
+  using Record = Kokkos::Impl::SharedAllocationRecord<Kokkos::HIPSpace, void>;
+
+  // TODO use the name and executionspace
+  template <typename PolicyDeduced, typename... ArgsDeduced>
+  GraphNodeKernelImpl(std::string, Kokkos::HIP const&, Functor arg_functor,
+                      PolicyDeduced&& arg_policy, ArgsDeduced&&... args)
+      : base_t(std::move(arg_functor), (PolicyDeduced &&) arg_policy,
+               (ArgsDeduced &&) args...) {}
+
+  template <typename PolicyDeduced>
+  GraphNodeKernelImpl(Kokkos::HIP const& exec_space, Functor arg_functor,
+                      PolicyDeduced&& arg_policy)
+      : GraphNodeKernelImpl("", exec_space, std::move(arg_functor),
+                            (PolicyDeduced &&) arg_policy) {}
+
+  ~GraphNodeKernelImpl() {
+    if (m_driver_storage) {
+      Record::decrement(Record::get_record(m_driver_storage));
+    }
+  }
+
+  void set_hip_graph_ptr(hipGraph_t* arg_graph_ptr) {
+    m_graph_ptr = arg_graph_ptr;
+  }
+
+  void set_hip_graph_node_ptr(hipGraphNode_t* arg_node_ptr) {
+    m_graph_node_ptr = arg_node_ptr;
+  }
+
+  hipGraphNode_t* get_hip_graph_node_ptr() const { return m_graph_node_ptr; }
+
+  hipGraph_t const* get_hip_graph_ptr() const { return m_graph_ptr; }
+
+  Kokkos::ObservingRawPtr<base_t> allocate_driver_memory_buffer() const {
+    KOKKOS_EXPECTS(m_driver_storage == nullptr);
+
+    auto* record = Record::allocate(
+        Kokkos::HIPSpace{}, "GraphNodeKernel global memory functor storage",
+        sizeof(base_t));
+
+    Record::increment(record);
+    m_driver_storage = reinterpret_cast<base_t*>(record->data());
+    KOKKOS_ENSURES(m_driver_storage != nullptr);
+
+    return m_driver_storage;
+  }
+
+ private:
+  Kokkos::ObservingRawPtr<const hipGraph_t> m_graph_ptr    = nullptr;
+  Kokkos::ObservingRawPtr<hipGraphNode_t> m_graph_node_ptr = nullptr;
+  Kokkos::OwningRawPtr<base_t> m_driver_storage            = nullptr;
+};
+
+struct HIPGraphNodeAggregateKernel {
+  using graph_kernel = HIPGraphNodeAggregateKernel;
+
+  // Aggregates don't need a policy, but for the purposes of checking the static
+  // assertions about graph kerenls,
+  struct Policy {
+    using is_graph_kernel = std::true_type;
+  };
+};
+
+template <typename KernelType,
+          typename Tag =
+              typename PatternTagFromImplSpecialization<KernelType>::type>
+struct get_graph_node_kernel_type
+    : type_identity<
+          GraphNodeKernelImpl<Kokkos::HIP, typename KernelType::Policy,
+                              typename KernelType::functor_type, Tag>> {};
+
+template <typename KernelType>
+struct get_graph_node_kernel_type<KernelType, Kokkos::ParallelReduceTag>
+    : type_identity<GraphNodeKernelImpl<
+          Kokkos::HIP, typename KernelType::Policy,
+          CombinedFunctorReducer<typename KernelType::functor_type,
+                                 typename KernelType::reducer_type>,
+          Kokkos::ParallelReduceTag>> {};
+
+template <typename KernelType>
+auto* allocate_driver_storage_for_kernel(KernelType const& kernel) {
+  using graph_node_kernel_t =
+      typename get_graph_node_kernel_type<KernelType>::type;
+  auto const& kernel_as_graph_kernel =
+      static_cast<graph_node_kernel_t const&>(kernel);
+
+  return kernel_as_graph_kernel.allocate_driver_memory_buffer();
+}
+
+template <typename KernelType>
+auto const& get_hip_graph_from_kernel(KernelType const& kernel) {
+  using graph_node_kernel_t =
+      typename get_graph_node_kernel_type<KernelType>::type;
+  auto const& kernel_as_graph_kernel =
+      static_cast<graph_node_kernel_t const&>(kernel);
+  hipGraph_t const* graph_ptr = kernel_as_graph_kernel.get_hip_graph_ptr();
+  KOKKOS_EXPECTS(graph_ptr != nullptr);
+
+  return *graph_ptr;
+}
+
+template <typename KernelType>
+auto& get_hip_graph_node_from_kernel(KernelType const& kernel) {
+  using graph_node_kernel_t =
+      typename get_graph_node_kernel_type<KernelType>::type;
+  auto const& kernel_as_graph_kernel =
+      static_cast<graph_node_kernel_t const&>(kernel);
+  auto* graph_node_ptr = kernel_as_graph_kernel.get_hip_graph_node_ptr();
+  KOKKOS_EXPECTS(graph_node_ptr != nullptr);
+
+  return *graph_node_ptr;
+}
+}  // namespace Impl
+}  // namespace Kokkos
+
+#endif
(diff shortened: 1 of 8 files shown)


core/src/Kokkos_Graph.hpp
```diff
@@ -161,6 +161,7 @@ Graph<ExecutionSpace> create_graph(Closure&& arg_closure) {
 #include <impl/Kokkos_GraphNodeImpl.hpp>
 #include <impl/Kokkos_Default_Graph_Impl.hpp>
 #include <Cuda/Kokkos_Cuda_Graph_Impl.hpp>
+#include <HIP/Kokkos_HIP_Graph_Impl.hpp>
```
masterleinad, line 164: This file doesn't have guards for `KOKKOS_ENABLE_HIP`.
```suggestion
#ifdef KOKKOS_ENABLE_HIP
#include <HIP/Kokkos_HIP_Graph_Impl.hpp>
#endif
```

core/src/HIP/Kokkos_HIP_GraphNodeKernel.hpp
```diff
@@ -0,0 +1,167 @@
...
+#include <Kokkos_Parallel.hpp>
+#include <Kokkos_Parallel_Reduce.hpp>
+#include <Kokkos_PointerOwnership.hpp>
+
+#include <HIP/Kokkos_HIP_SharedAllocationRecord.hpp>
+
+namespace Kokkos {
+namespace Impl {
+
+// FIXME Remove once all backends implement the new reduce interface
```
masterleinad, line 34:
```suggestion
```

masterleinad:
This doesn't work
```
Start  7: Kokkos_UnitTest_HIPGraph

7: Test command: /var/jenkins/workspace/Kokkos/build/core/unit_test/Kokkos_UnitTest_HIPGraph
7: Test timeout computed to be: 1500
7: [==========] Running 5 tests from 1 test suite.
7: [----------] Global test environment set-up.
7: [----------] 5 tests from hip_count_bugs
7: [ RUN      ] hip_count_bugs.launch_one
7: [       OK ] hip_count_bugs.launch_one (11 ms)
7: [ RUN      ] hip_count_bugs.launch_one_rvalue
7: [       OK ] hip_count_bugs.launch_one_rvalue (8 ms)
7: [ RUN      ] hip_count_bugs.launch_six
7: /var/jenkins/workspace/Kokkos/core/unit_test/TestGraph.hpp:144: Failure
7: Expected equality of these values:
7:   6
7:   count_host()
7:     Which is: 8
7: [  FAILED  ] hip_count_bugs.launch_six (24 ms)
7: [ RUN      ] hip_count_bugs.when_all_cycle
7: /var/jenkins/workspace/Kokkos/core/unit_test/TestGraph.hpp:167: Failure
7: Expected equality of these values:
7:   0
7:   bugs_host()
7:     Which is: 3
7: [  FAILED  ] hip_count_bugs.when_all_cycle (9 ms)
7: [ RUN      ] hip_count_bugs.zero_work_reduce
7: [       OK ] hip_count_bugs.zero_work_reduce (8 ms)
7: [----------] 5 tests from hip_count_bugs (63 ms total)
7: 
7: [==========] 5 tests from 1 test suite ran. (63 ms total)
7: [  PASSED  ] 3 tests.
7: [  FAILED  ] 2 tests, listed below:
7: [  FAILED  ] hip_count_bugs.launch_six
7: [  FAILED  ] hip_count_bugs.when_all_cycle
```

Rombur:
Looks like it only fails for 5.2. I will guard the code with the HIP version

dalg24 (commented):
(Partial review.  Haven't looked at the kernel launch changes yet)

core/src/HIP/Kokkos_HIP_GraphNode_Impl.hpp
```diff
@@ -0,0 +1,56 @@
...
+
+  GraphNodeBackendDetailsBeforeTypeErasure(
+      Kokkos::HIP const &, _graph_node_is_root_ctor_tag,
+      GraphNodeBackendSpecificDetails<Kokkos::HIP> &) noexcept {}
+};
+
+}  // namespace Impl
+}  // namespace Kokkos
+
+#include <HIP/Kokkos_HIP_GraphNodeKernel.hpp>
```
masterleinad, line 54: I would prefer to just inline the implementation in `HIP/Kokkos_HIP_GraphNodeKernel.hpp` if we can't do with forward declarations.

masterleinad (commented):
Looks OK to me. The implementation is pretty much copied from the Cuda one.

core/src/HIP/Kokkos_HIP_KernelLaunch.hpp
```diff
@@ -367,6 +379,42 @@ struct HIPParallelLaunchKernelInvoker<DriverType, LaunchBounds,
...
+          &graph_node, graph, /* dependencies = */ nullptr,
+          /* numDependencies = */ 0, &params));
+    } else {
+      // We still need an empty node for the dependency structure
+      KOKKOS_IMPL_HIP_SAFE_CALL(
+          hipGraphAddEmptyNode(&graph_node, graph,
+                               /* dependencies = */ nullptr,
+                               /* numDependencies = */ 0));
+    }
+    KOKKOS_ENSURES(graph_node);
```
dalg24, line 415: Wondering if we need this.

core/src/Kokkos_Graph.hpp
```diff
@@ -161,6 +161,11 @@ Graph<ExecutionSpace> create_graph(Closure&& arg_closure) {
 #include <impl/Kokkos_GraphNodeImpl.hpp>
 #include <impl/Kokkos_Default_Graph_Impl.hpp>
 #include <Cuda/Kokkos_Cuda_Graph_Impl.hpp>
+#if defined(KOKKOS_ENABLE_HIP)
+#if !((HIP_VERSION_MAJOR == 5) && (HIP_VERSION_MINOR == 2))
+#include <HIP/Kokkos_HIP_Graph_Impl.hpp>
```
romintomasetti, line 166: Hi @Rombur ! I can't wait to see the HIP implementation of `Kokkos::Graph`!

I'm just a bit worried about the code duplication here. As you mentioned in the PR description, the HIP API is very similar to the one of CUDA.

So I wonder if having some sort of macro for the common part would help. Some classes are cleary near copy-pastes.

For some functions like `create_aggregate_ptr` it's even more obvious...

romintomasetti (changes requested):

core/src/Kokkos_Graph.hpp
Rombur: It's true that the code is very similar but it's always the case when we first port CUDA code to HIP. Over time the backends are optimized differently and the code diverges.

core/src/HIP/Kokkos_HIP_KernelLaunch.hpp
```diff
@@ -25,6 +25,11 @@
 #include <HIP/Kokkos_HIP_Instance.hpp>
 #include <HIP/Kokkos_HIP_Space.hpp>
 
+#if !((HIP_VERSION_MAJOR == 5) && (HIP_VERSION_MINOR == 2))
```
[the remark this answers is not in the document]
tcclevenger: Kind of related, Cuda has a `KOKKOS_ENABLE_IMPL_CUDA_MALLOC_ASYNC` which is always used with checking version
```
#if (defined(KOKKOS_ENABLE_IMPL_CUDA_MALLOC_ASYNC) && CUDART_VERSION >= 11020)
```
Should an error be thrown if the user defines `KOKKOS_ENABLE_IMPL_CUDA_MALLOC_ASYNC` with version <11.2 (and similar here as well, defining `KOKKOS_IMPL_HIP_GRAPH_ENABLED` with <rocm5.2)? The user might define these macros and not realize they are not being used?

core/src/HIP/Kokkos_HIP_KernelLaunch.hpp
```diff
@@ -481,42 +579,57 @@ struct HIPParallelLaunch<
...
+      }
     }
-  }
 #endif
+  }
 }
 }  // namespace Impl
 }  // namespace Kokkos
 
+#undef KOKKOS_ENABLE_HIP_RELOCATABLE_DEVICE_CODE
```
dalg24, line 631:
```suggestion
#undef KOKKOS_IMPL_HIP_GRAPH_ENABLED
```

masterleinad (approved):
Fine with me.

core/src/Kokkos_Graph.hpp
```diff
@@ -161,6 +161,12 @@ Graph<ExecutionSpace> create_graph(Closure&& arg_closure) {
 #include <impl/Kokkos_GraphNodeImpl.hpp>
 #include <impl/Kokkos_Default_Graph_Impl.hpp>
 #include <Cuda/Kokkos_Cuda_Graph_Impl.hpp>
+#if defined(KOKKOS_ENABLE_HIP)
+// The implementation of hipGraph in ROCm 5.2 is bugged, so we cannot use it.
+#if !((HIP_VERSION_MAJOR == 5) && (HIP_VERSION_MINOR == 2))
```
fnrizzi, line 166: the PR description says that starting from 5.2 it is supported, but then here you check for not 5.2. 
Can we just guard for HIP > 5.2 ? (then i read Damien's comment, i agree on adding a comment to clarify the min version)

core/src/HIP/Kokkos_HIP_Graph_Impl.hpp
```diff
@@ -0,0 +1,187 @@
...
+
+  // Not moveable or copyable; it spends its whole life as a shared_ptr in the
+  // Graph object.
+  GraphImpl()                 = delete;
+  GraphImpl(GraphImpl const&) = delete;
+  GraphImpl(GraphImpl&&)      = delete;
+  GraphImpl& operator=(GraphImpl const&) = delete;
+  GraphImpl& operator=(GraphImpl&&) = delete;
+
+  ~GraphImpl();
```
fnrizzi, line 51: two comments: 
1) why is this not defaulted? 
2) the cuda impl does a few things in the desctructor, why here we don't need to do the same/similar?

core/src/HIP/Kokkos_HIP_Parallel_MDRange.hpp
```diff
@@ -30,7 +30,8 @@ namespace Impl {
 template <class FunctorType, class... Traits>
 class ParallelFor<FunctorType, Kokkos::MDRangePolicy<Traits...>, HIP> {
  public:
-  using Policy = Kokkos::MDRangePolicy<Traits...>;
+  using Policy       = Kokkos::MDRangePolicy<Traits...>;
+  using functor_type = FunctorType;
```
fnrizzi, line 34: was this added becuase needed for the graph?

core/src/HIP/Kokkos_HIP_Parallel_Range.hpp
```diff
@@ -121,6 +121,7 @@ class ParallelReduce<CombinedFunctorReducerType, Kokkos::RangePolicy<Traits...>,
   using value_type     = typename ReducerType::value_type;
   using reference_type = typename ReducerType::reference_type;
   using functor_type   = FunctorType;
+  using reducer_type   = ReducerType;
```
fnrizzi, line 124: was this added becuase needed for the graph?

core/src/HIP/Kokkos_HIP_Parallel_Team.hpp
```diff
@@ -587,7 +587,8 @@ class ParallelReduce<CombinedFunctorReducerType,
   using value_type     = typename ReducerType::value_type;
 
  public:
-  using size_type = HIP::size_type;
+  using functor_type = FunctorType;
+  using size_type    = HIP::size_type;
```
fnrizzi, line 591: was this added becuase needed for the graph?

core/src/HIP/Kokkos_HIP_Parallel_Team.hpp
```diff
@@ -425,7 +425,7 @@ __device__ inline void hip_release_scratch_index(int32_t* scratch_locks,
 template <typename FunctorType, typename... Properties>
 class ParallelFor<FunctorType, Kokkos::TeamPolicy<Properties...>, HIP> {
  public:
-  using Policy       = TeamPolicyInternal<HIP, Properties...>;
+  using Policy       = TeamPolicy<Properties...>;
```
fnrizzi, line 428: i guess this is a correction?

fnrizzi:
@Rombur is all graph stuff tested already as part of a backend-agnostic test somewhere?

core/src/Kokkos_Graph.hpp
Rombur: That's not what the description says, hipGraph was introduced in ROCm 4.5. ROCm 5.2 is the oldest version of ROCm that we support  but unfortunately that in ROCm 5.2m hipGraph is bugged (from what I read older versions worked)

core/src/HIP/Kokkos_HIP_Parallel_MDRange.hpp
Rombur: yes

core/src/HIP/Kokkos_HIP_Parallel_Range.hpp
Rombur: yes

core/src/HIP/Kokkos_HIP_Parallel_Team.hpp
Rombur: yes

core/src/HIP/Kokkos_HIP_Parallel_Team.hpp
Rombur: I don't remember exactly why but I needed this change for the graph to work.

Rombur:
> @Rombur is all graph stuff tested already as part of a backend-agnostic test somewhere?

Yes, we have graph tests.

core/src/Kokkos_Graph.hpp
fnrizzi: sorry i misread the 4.5 in the description for a 4.2! 
I still think that it would be helpful to add to the comment the fact that we are assuming the ROCm minimum version to be 5.2. which i think right now it is implicitly assumed here right?

core/src/HIP/Kokkos_HIP_Parallel_Team.hpp
fnrizzi: just a personal suggestion that maybe this and the other nested typedefs in my other comments above is something that could be done in a separate PR to avoid adding too many drive-by changes to this PR?

core/src/Kokkos_Graph.hpp
Rombur: > I still think that it would be helpful to add to the comment the fact that we are assuming the ROCm minimum version to be 5.2. which i think right now it is implicitly assumed here right?

I don't understand what you mean. Among the versions of ROCm that we support, one of them is bugged and therefore, for that bugged version, we use the fallback implementation. Why does it matter that it's the minimal version that we support?

fnrizzi: because that guard has no base to exist if the minim version is > 5.2. so we could at least say that this is here becaause at the time of this development the min version kokkos supports is 5.2 and should be removed when the min gets updated. so that 2 years from now when the min will be 6 or whatever, anybody reading this will realize this is not just a forgotten thing

fnrizzi: if you want to take my suggestion great, otherwise no big deal ! i can move along

core/src/HIP/Kokkos_HIP_KernelLaunch.hpp
Rombur: We do need it. Like we talked about if you add an empty node, you end up with two empty nodes but it's necessary. If you don't, the code crashes when it tries to build the dependency structure. I don't think it's a requirement from the hip library itself, so we could potentially rework that if we wanted to.

core/src/HIP/Kokkos_HIP_KernelLaunch.hpp
```diff
@@ -367,6 +383,42 @@ struct HIPParallelLaunchKernelInvoker<DriverType, LaunchBounds,
...
+  static void create_parallel_launch_graph_node(
+      DriverType const &driver, dim3 const &grid, dim3 const &block, int shmem,
+      HIPInternal const * /*hip_instance*/) {
+    auto const &graph = get_hip_graph_from_kernel(driver);
+    KOKKOS_EXPECTS(graph);
+    auto &graph_node = get_hip_graph_node_from_kernel(driver);
+    // Expect node not yet initialized
+    KOKKOS_EXPECTS(!graph_node);
+
+    if (!Impl::is_empty_launch(grid, block)) {
```
crtrott, line 397:
```suggestion
    if (!is_empty_launch(grid, block)) {
```

crtrott:
I added the Impl removal suggested by Damien mainly to trigger full CI. If it passes I merge.

crtrott (approved):

Rombur:
Retest this please

PR #6372, opened 2023-08-23
Fixes for Kokkos::Array

nliber:
(Part of issue #6355)

This P/R contains fixes for

`Kokkos::Array<T, KOKKOS_INVALID_INDEX, Kokkos::Array<>::contiguous>::empty()`
`Kokkos::Array<T, KOKKOS_INVALID_INDEX, Kokkos::Array<>::strided>::empty()`
`Kokkos::Array<T, KOKKOS_INVALID_INDEX, Kokkos::Array<>::strided>::operator=(...)`

As well as unit tests for the above and the rest of `Kokkos::Array`.

Added 2023-Aug-23:

Changed the return value for `Kokkos::Array<T, 0>::data()` from `pointer(0)` to `nullptr`.  Technically this wasn't a bug, but it did trigger warnings under some builds.

PR #6373, opened 2023-08-23
Kokkos::Array deduction guide

nliber:
(Part of issue #6355, intended to be applied after #6372, as that contains the initial unit tests for `std::array`.)

This adds a deduction guide for `Kokkos::Array<T, N, void>` (where `0<N`) analogous to the one for `std::array`, along with unit tests.

### DIFF
--- a/core/src/Kokkos_Array.hpp
+++ b/core/src/Kokkos_Array.hpp
@@ -320,6 +320,9 @@ struct Array<T, KOKKOS_INVALID_INDEX, Array<>::strided> {
       : m_elem(arg_ptr), m_size(arg_size), m_stride(arg_stride) {}
 };
 
+template <typename T, typename... Us>
+Array(T, Us...)->Array<T, 1 + sizeof...(Us)>;
+
 }  // namespace Kokkos
 
 //<editor-fold desc="Support for structured binding">
--- a/core/unit_test/TestArray.cpp
+++ b/core/unit_test/TestArray.cpp
@@ -49,4 +49,25 @@ KOKKOS_FUNCTION constexpr bool test_array_structured_binding_support() {
 
 static_assert(test_array_structured_binding_support());
 
+template <typename L, typename R>
+constexpr bool is_equal(L const& l, R const& r) {
+  if (std::size(l) != std::size(r)) return false;
+
+  for (size_t i = 0; i != std::size(l); ++i) {
+    if (l[i] != r[i]) return false;
+  }
+
+  return true;
+}
+
+KOKKOS_FUNCTION constexpr bool test_array_ctad() {
+  constexpr int x = 10;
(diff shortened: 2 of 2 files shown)


core/unit_test/TestArrayOps.hpp
```diff
@@ -0,0 +1,393 @@
...
+
+TEST(TEST_CATEGORY, array_contiguous_capacity) {
+  using A =
+      Kokkos::Array<int, KOKKOS_INVALID_INDEX, Kokkos::Array<>::contiguous>;
+
+  A e(nullptr, 0);
+
+  ASSERT_TRUE(e.empty());
+  ASSERT_EQ(e.size(), 0u);
+  ASSERT_EQ(e.max_size(), 0u);
```
ldh4, line 122: I wonder if there should be a line explicitly testing for `e.data() == nullptr` (even though it is redundant), since that is one of the changes introduced in this PR.

core/src/Kokkos_Array.hpp
```diff
@@ -199,7 +199,7 @@ struct Array<T, KOKKOS_INVALID_INDEX, Array<>::contiguous> {
   using const_pointer   = std::add_const_t<T>*;
 
   KOKKOS_INLINE_FUNCTION constexpr size_type size() const { return m_size; }
-  KOKKOS_INLINE_FUNCTION constexpr bool empty() const { return 0 != m_size; }
+  KOKKOS_INLINE_FUNCTION constexpr bool empty() const { return 0 == m_size; }
```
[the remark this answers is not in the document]
nliber: I fixed that in #6372.  Let me try rebasing again...

core/unit_test/TestArrayOps.hpp
nliber: `data()` will be whatever pointer was passed in.  `nullptr` and a size of `0` aren't special here.

And it isn't part of the changes for this P/R, as the test above only applies to `contiguous` non-owning arrays.

core/unit_test/TestArray.cpp
```diff
@@ -49,4 +49,25 @@ KOKKOS_FUNCTION constexpr bool test_array_structured_binding_support() {
 
 static_assert(test_array_structured_binding_support());
 
+template <typename L, typename R>
+constexpr bool is_equal(L const& l, R const& r) {
```
ldh4, line 53:
```suggestion
KOKKOS_FUNCTION constexpr bool is_equal(L const& l, R const& r) {
```

nliber: Done.

ldh4 (approved):

crtrott:
Retest this please.

crtrott (approved):
Waiting for testing again.

dalg24:
Ignoring failures on Jenkins.

PR #6374, opened 2023-08-23
Adding Kokkos::to_Array

nliber:
(Part of issue #6355, intended to be applied after #6372, as that contains the initial unit tests for `std::array`.)

Adding `Kokkos::to_Array` for `Kokkos::Array`, analogous to `std::to_array` for `std::array`, along with unit tests.


nliber:
This is against the wrong branch.  Closing.

PR #6375, opened 2023-08-23
Adding `Kokkos::to_array`

nliber:
(Part of issue #6355, intended to be applied after #6372, as that contains the initial unit tests for `Kokkos::array`.)

Adding `Kokkos::to_Array` for `Kokkos::Array<T, N, void>`, analogous to `std::to_array` for `std::array`, as well as unit tests.

### DIFF
--- a/core/src/Kokkos_Array.hpp
+++ b/core/src/Kokkos_Array.hpp
@@ -356,6 +356,32 @@ struct KOKKOS_DEPRECATED
 template <typename T, typename... Us>
 Array(T, Us...)->Array<T, 1 + sizeof...(Us)>;
 
+namespace Impl {
+
+template <typename T, size_t N, size_t... I>
+KOKKOS_INLINE_FUNCTION constexpr Array<std::remove_cv_t<T>, N> to_Array_impl(
+    T (&a)[N], std::index_sequence<I...>) {
+  return {{a[I]...}};
+}
+
+template <typename T, size_t N, size_t... I>
+KOKKOS_INLINE_FUNCTION constexpr Array<std::remove_cv_t<T>, N> to_Array_impl(
+    T(&&a)[N], std::index_sequence<I...>) {
+  return {{std::move(a[I])...}};
+}
+
+}  // namespace Impl
+
+template <class T, size_t N>
+KOKKOS_INLINE_FUNCTION constexpr auto to_Array(T (&a)[N]) {
+  return Impl::to_Array_impl(a, std::make_index_sequence<N>{});
+}
+
+template <class T, size_t N>
+KOKKOS_INLINE_FUNCTION constexpr auto to_Array(T(&&a)[N]) {
+  return Impl::to_Array_impl(std::move(a), std::make_index_sequence<N>{});
+}
+
 }  // namespace Kokkos
 
 //<editor-fold desc="Support for structured binding">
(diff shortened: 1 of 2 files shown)


core/unit_test/TestArrayOps.hpp
dalg24, line 1: Does this belong to this PR?

core/unit_test/TestArray.cpp
```diff
@@ -160,4 +160,23 @@ constexpr bool test_array_specialization_kokkos_swap() {
 
 static_assert(test_array_specialization_kokkos_swap());
 
+KOKKOS_FUNCTION constexpr bool test_to_array() {
```
dalg24, line 163:
```suggestion
constexpr bool test_to_array() {
```

core/src/Kokkos_Array.hpp
```diff
@@ -356,6 +356,32 @@ struct KOKKOS_DEPRECATED
...
+template <typename T, size_t N, size_t... I>
+KOKKOS_FUNCTION constexpr Array<std::remove_cv_t<T>, N> to_Array_impl(
+    T(&&a)[N], std::index_sequence<I...>) {
+  return {{std::move(a[I])...}};
+}
+
+}  // namespace Impl
+
+template <typename T, size_t N>
+KOKKOS_INLINE_FUNCTION constexpr auto to_Array(T (&a)[N]) {
```
dalg24, line 376: I would drop the "inline" here and below.

dalg24 (commented):
I am not a big fine of the spelling `to_Array` but I doubt whether `to_array` would get consensus either...

Looks fine to me if you can get the CI to pass.
I was able to reproduce the GCC 8.4 failure https://godbolt.org/z/o9Pf3eKEn but I will let you the honor to find a fix :)

core/unit_test/TestArray.cpp
nliber: Fixed (or will be the next time I push the repo up).

core/src/Kokkos_Array.hpp
nliber: Fixed (or will be the next time I push the repo up).

crtrott:
Also prefer `to_array` conforms more with our standard naming practice, and its not like `create_mirror_view` uses a capital `V`.

dalg24:
> Also prefer `to_array` conforms more with our standard naming practice, and its not like `create_mirror_view` uses a capital `V`.

You can probably find all sorts of example in the codebase depending on what is the point you are trying to make :p

dalg24 (approved):

dalg24:
The Theads build failure on GH action is clearly unrelated
```
 -- Testing 3-D histogram
Density 3D: 7.28572e-05 -0.00120552 0.000316342 || 0.051031 1e+64 -1e+64
Test Scalar=double
 -- Testing randomness properties
/__w/kokkos/kokkos/algorithms/unit_tests/TestRandom.hpp:287: Failure
Expected: (std::abs(variance_eps)) < (1.5 * tolerance), actual: 0.00081575001396583247 vs 0.00075405570881983128
```
Same for the runtimes exception `"Desul::Error: init_lock_arrays_hip: post malloc error(hipErrorOutOfMemory): hipErrorOutOfMemory"` in one of the HIP builds.
Deciding to be impatient and not waiting for the CUDA debug build that is at 80% now

dalg24 (approved):

dalg24:
@nliber make sure to add to the changelog

PR #6376, opened 2023-08-23
Use constexpr West

Rombur:
Use `constexpr West` because  a lot of people in the team have bad taste and prefer it. Most of the changes are in the HIP backend. @ldh4 I also changed the order in the SIMD tests. I can revert that change if you want.


dalg24 (approved):

cz4rs (approved):
Looks good to me. Can we make the change in `.clang-format` as well?

masterleinad:
> Looks good to me. Can we make the change in `.clang-format` as well?

No, our `clang-format` version is not recent enough.

masterleinad (approved):

PR #6380, opened 2023-08-23
OpenMPTarget: clang extensions for dynamic shared memory.

rgayatri23:
The PR does the following:
* Use OpenMP extensions in clang/clang++ for requesting dynamic shared memory. The extensions are not a part of OpenMP spec
* Use the extensions to request L0 scratch memory
* Edit the necessary parts of the implementation accordingly.

Tested 3 kernels that use scratch memory with these extensions on A100 and MI250x:
kernel-1 - >2x speedup
kernel-2 - 10% performance regression
kernel-3 - no change in time-to-solution


core/src/OpenMPTarget/Kokkos_OpenMPTarget_Parallel.hpp
```diff
@@ -32,6 +32,14 @@
 #define KOKKOS_IMPL_OPENMPTARGET_HIERARCHICAL_INTEL_GPU
 #endif
 
+// Define a macro for llvm compiler greater than version 15 and on NVIDIA and
+// AMD GPUs. This would be useful in cases where non-OpenMP standard llvm
+// extensions can be used.
+#if defined(KOKKOS_COMPILER_CLANG) && (KOKKOS_COMPILER_CLANG >= 1500) && \
+    (defined(KOKKOS_ARCH_AMD_GPU) || defined(KOKKOS_IMPL_ARCH_NVIDIA_GPU))
+#define KOKKOS_IMPL_OPENMPTARGET_LLVM_EXTENSIONS
```
dalg24, line 40: Is that something you would want to see in print_configuration?
If we were getting a bug report, would you care to know about that setting?

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Parallel.hpp
```diff
@@ -249,15 +257,37 @@ class OpenMPTargetExecTeamMember {
     // and L1 shmem size. TEAM_REDUCE_SIZE = 512 bytes saved per team for
     // hierarchical reduction. There is an additional 10% of the requested
     // scratch memory allocated per team as padding. Hence the product with 0.1.
+    //
+    // Use llvm extensions for dynamic shared memory with compilers/architecture
+    // combinations where it is supported.
+    //
+    // Size allocated in HBM will now change based on whether we use llvm
+    // extensions.
+#if defined(KOKKOS_IMPL_OPENMPTARGET_LLVM_EXTENSIONS)
```
dalg24, line 266: Are they other LLVM extensions that you might be willing to use?
Should we be more specific in the macro name?

rgayatri23: There are active plans towards working on extensions that might be considered "useful" for accelerators but not in OpenMP spec (for now at-least). Thats the reason I kept the macro name generic.  But there is none that can be used in near future.

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Parallel.hpp
rgayatri23: Thats a good idea. I will include it there.

rgayatri23:
Putting this on a hold for now until CI is updated to 17. That should happen in the next month.

core/src/OpenMPTarget/Kokkos_OpenMPTarget_ParallelFor_Team.hpp
```diff
@@ -140,8 +140,16 @@ class ParallelFor<FunctorType, Kokkos::TeamPolicy<Properties...>,
 // guarantees that the number of teams specified in the `num_teams` clause is
 // always less than or equal to the maximum concurrently running teams.
 #if !defined(KOKKOS_IMPL_OPENMPTARGET_HIERARCHICAL_INTEL_GPU)
+      // Use scratch memory extensions to request dynamic shared memory for the
+      // right compiler/architecture combination.
+#if defined(KOKKOS_IMPL_OPENMPTARGET_LLVM_EXTENSIONS)
+#pragma omp target teams thread_limit(team_size) firstprivate(a_functor) \
+    num_teams(max_active_teams) is_device_ptr(scratch_ptr)               \
+        ompx_dyn_cgroup_mem(shmem_size_L0)
+#else
```
dalg24, line 149: Considering all the use cases are just "add the directive if LLVM extensions are available", you could do something along these lines (not tested)
```C++
#define KOKKOS_IMPL_OMPTARGET_PRAGMA_HELPER(x) _Pragma(#x)
#define KOKKOS_IMPL_OMPTARGET_PRAGMA(x) KOKKOS_IMPL_OMP_PRAGMA_HELPER(omp target x)

#ifdef KOKKOS_IMPL_OPENMPTARGET_LLVM_EXTENSIONS
#define KOKKOS_IMPL_OMPX_DYN_CGROUP_MEM(N) ompx_dyn_cgroup_mem(N)
#else
#define KOKKOS_IMPL_OMPX_DYN_CGROUP_MEM(N)
#endif

KOKKOS_IMPL_OMPTARGET_PRAGMA(teams thread_limit(team_size) firstprivate(a_functor) \
    num_teams(max_active_teams) is_device_ptr(scratch_ptr) \
    KOKKOS_IMPL_OMPX_DYN_CGROUP_MEM(shmem_size_L0))
#pragma omp parallel
    {
       // ...

File: core/src/OpenMPTarget/Kokkos_OpenMPTarget_Parallel.hpp [hunk 4]
```diff
@@ -32,6 +32,14 @@
 #define KOKKOS_IMPL_OPENMPTARGET_HIERARCHICAL_INTEL_GPU
 #endif
 
+// Define a macro for llvm compiler greater than version 15 and on NVIDIA and
+// AMD GPUs. This would be useful in cases where non-OpenMP standard llvm
+// extensions can be used.
+#if defined(KOKKOS_COMPILER_CLANG) && (KOKKOS_COMPILER_CLANG >= 1500) && \
```
dalg24, line 38: Is that the right version to guard for?

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Parallel.hpp
dalg24: Ping

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Parallel.hpp
rgayatri23: No that has to change to 17 to be absolutely sure of the support.

core/src/OpenMPTarget/Kokkos_OpenMPTarget_ParallelFor_Team.hpp
rgayatri23: For now, I wrapped only the pragmas in this file. In a future PR I will put the macro in a central file and update this in all openmp calls.

core/src/OpenMPTarget/Kokkos_OpenMPTarget_ParallelFor_Team.hpp
```diff
@@ -22,6 +22,18 @@
...
+#define KOKKOS_IMPL_OMPTARGET_PRAGMA(x) \
+  KOKKOS_IMPL_OPENMPTARGET_PRAGMA_HELPER(omp target x)
+
+// Use scratch memory extensions to request dynamic shared memory for the
+// right compiler/architecture combination.
+#ifdef KOKKOS_IMPL_OPENMPTARGET_LLVM_EXTENSIONS
+#define KOKKOS_IMPL_OMPX_DYN_CGROUP_MEM(N) ompx_dyn_cgroup_mem(N)
+#else
+#define KOKKOS_IMPL_OMPX_DYN_CGROUP_MEM(N)
+#endif
```
dalg24, line 35: Move these to a `Kokkos_OpenMPTarget_Macros.hpp` header and include/use it where appropriate

rgayatri23: Yeah I was planning on doing it in  a separate PR just dedicated to this, but I can do it now too.

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Parallel.hpp
masterleinad: ping

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Parallel_Common.hpp
masterleinad, line 1: In lines 372 and 469, you are still requesting memory for L0 from the global scratch space even though you want to take it from the local scratch space with the new implementation.
Also, can we relax 616-620?

rgayatri23: Thanks for pointing it out. I think I will have to edit the interface of `resize_scratch` to avoid level-0 completely.

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Parallel_Common.hpp
```diff
@@ -370,7 +371,7 @@ struct ParallelReduceSpecialize<FunctorType, TeamPolicyInternal<PolicyArgs...>,
     const size_t shmem_size_L0 = p.scratch_size(0, team_size);
     const size_t shmem_size_L1 = p.scratch_size(1, team_size);
     OpenMPTargetExec::resize_scratch(PolicyType::member_type::TEAM_REDUCE_SIZE,
-                                     shmem_size_L0, shmem_size_L1, league_size);
+                                     0, shmem_size_L1, league_size);
```
masterleinad, line 374: Don't you still need the L0 allocation here (and below) if `KOKKOS_IMPL_OPENMPTARGET_LLVM_EXTENSIONS` isn't defined?

rgayatri23: Yes, thanks. Fixed it.

rgayatri23:
Also, I changed a few `map(to:...` to `firstprivate(...` while I was editing the pragmas for the sake of performance. Let me know if you want to revert that and add it as a separate PR.

masterleinad:
> Also, I changed a few `map(to:...` to `firstprivate(...` while I was editing the pragmas for the sake of performance. Let me know if you want me to revert that and add it as a separate PR.

Yeah, let's do that in a separate pull request.

rgayatri23:
> > Also, I changed a few `map(to:...` to `firstprivate(...` while I was editing the pragmas for the sake of performance. Let me know if you want me to revert that and add it as a separate PR.
> 
> Yeah, let's do that in a separate pull request.

Restored that change.

masterleinad (approved):

rgayatri23:
Retest this please

dalg24 (commented):
```
git grep -n resize_scratch core/src/OpenMPTarget/
core/src/OpenMPTarget/Kokkos_OpenMPTarget_Exec.cpp:98:void OpenMPTargetExec::resize_scratch(int64_t team_size, int64_t shmem_size_L0,
core/src/OpenMPTarget/Kokkos_OpenMPTarget_Parallel.hpp:748:  static void resize_scratch(int64_t team_reduce_bytes,
core/src/OpenMPTarget/Kokkos_OpenMPTarget_ParallelFor_Team.hpp:107:    OpenMPTargetExec::resize_scratch(team_size, shmem_size_L0, shmem_size_L1,
core/src/OpenMPTarget/Kokkos_OpenMPTarget_ParallelFor_Team.hpp:128:    // omp_set_num_teams in the resize_scratch routine, the call is not
core/src/OpenMPTarget/Kokkos_OpenMPTarget_Parallel_Common.hpp:231:    // parameter of `resize_scratch=1`.
core/src/OpenMPTarget/Kokkos_OpenMPTarget_Parallel_Common.hpp:232:    OpenMPTargetExec::resize_scratch(1, 0, value_count * sizeof(ValueType),
core/src/OpenMPTarget/Kokkos_OpenMPTarget_Parallel_Common.hpp:372:    OpenMPTargetExec::resize_scratch(PolicyType::member_type::TEAM_REDUCE_SIZE,
core/src/OpenMPTarget/Kokkos_OpenMPTarget_Parallel_Common.hpp:462:    OpenMPTargetExec::resize_scratch(PolicyType::member_type::TEAM_REDUCE_SIZE,
core/src/OpenMPTarget/Kokkos_OpenMPTarget_Parallel_Common.hpp:612:    OpenMPTargetExec::resize_scratch(1, 0, value_count * sizeof(ValueType),
```
There are two `resize_scratch` untouched.  Is that a TODO?

I am asking because I noticed that we guard at call site to set the L0 to zero when the LLVM extension is enabled and if it wasn't for these two calls we could guard the implementation instead.

dalg24 (changes requested):
I get it now, the two untouched ones are those that had a L0 argument set to zero in the first place.
Please undo the changes at call site and guard in the implementation of `resize_scratch` to ignore L0 when the LLVM extension is available

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Parallel_Common.hpp
```diff
@@ -369,6 +370,7 @@ struct ParallelReduceSpecialize<FunctorType, TeamPolicyInternal<PolicyArgs...>,
 
     const size_t shmem_size_L0 = p.scratch_size(0, team_size);
     const size_t shmem_size_L1 = p.scratch_size(1, team_size);
+
```
dalg24, line 373:
```suggestion
```

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Parallel_Common.hpp
```diff
@@ -459,6 +463,7 @@ struct ParallelReduceSpecialize<FunctorType, TeamPolicyInternal<PolicyArgs...>,
...
```
dalg24, line 466:
```suggestion
```

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Exec.cpp
```diff
@@ -98,6 +98,11 @@ void OpenMPTargetExec::resize_scratch(int64_t team_size, int64_t shmem_size_L0,
                                       int64_t shmem_size_L1,
                                       int64_t league_size) {
   Kokkos::Experimental::OpenMPTargetSpace space;
+  // FIXME_OPENMPTATGET - Level-0 scratch when using clang/17 and higher comes
```
dalg24, line 101: It is not really a FIXME though is it?
Maybe add that the L0 request is performed via `ompx_dyn_cgroup_mem(shmem_size_L0)` in the omp pragma

dalg24 (approved):

PR #6382, opened 2023-08-24
Add deprecated attribute to `HostSpace(AllocationMechanism)` definition

dalg24:
This is an attempt to silent a deprecated declaration warning when building Kokkos.
The warning shows in the ArborX nightly GCC 13.1 build.

The declaration had the deprecated attribute since #6341
https://github.com/kokkos/kokkos/blob/b26a6c620b4e89d8a842a662fb6547abef105719/core/src/Kokkos_HostSpace.hpp#L85-L86

### DIFF
--- a/core/src/impl/Kokkos_HostSpace.cpp
+++ b/core/src/impl/Kokkos_HostSpace.cpp
@@ -56,7 +56,8 @@
 namespace Kokkos {
 
 #ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
-HostSpace::HostSpace(const HostSpace::AllocationMechanism &) : HostSpace() {}
+KOKKOS_DEPRECATED HostSpace::HostSpace(const HostSpace::AllocationMechanism &)
+    : HostSpace() {}
 #endif
 
 void *HostSpace::allocate(const size_t arg_alloc_size) const {


masterleinad:
How would this change anything? Can you please elaborate?

dalg24:
Without the attribute on the definition we get
```
/root/kokkos/core/src/impl/Kokkos_HostSpace.cpp:63:60: warning: 'AllocationMechanism' is deprecated [-Wdeprecated-declarations]
   63 | HostSpace::HostSpace(const HostSpace::AllocationMechanism &) : HostSpace() {}
In file included from /root/kokkos/core/src/impl/Kokkos_HostSpace.cpp:53:
/root/kokkos/core/src/Kokkos_HostSpace.hpp:78:26: note: declared here
   78 |   enum KOKKOS_DEPRECATED AllocationMechanism {
      |                          ^~~~~~~~~~~~~~~~~~~
```
With the attribute there is no warning.

masterleinad (approved):
Fascinating, I convinced myself that this makes a difference for `gcc` but not `clang`, see https://godbolt.org/z/c1fzPMePv, that doesn't emit a warning either way.

crtrott (approved):

PR #6384, opened 2023-08-24
simd: use kokkos lambda to avoid host device function calling host only lambda

ldh4:
Recently merged in #6347 introduced a few occurrences of `calling a __host__ function from a __host__ __device__ function` problems.

This occurred in two places:
  - In generator ctors unit test where host lambdae are passed into host device generator ctors.
  - Calling a constexpr std::size_t operator in host device lambda (by a generator constructor).
  
because non-scalar backend simds interfaces are strictly host-only, only `Scalar` simd is affected.


crtrott (approved):

dalg24 (approved):

masterleinad (approved):

dalg24:
Ignoring unrelated timeouts in the CI

PR #6386, opened 2023-08-25
Suppress NVCC attribute ignored warning on SIMD operators as hidden friends

### LINKED ISSUE #6383: Nightly build errors with cuda/11.{1,2}+gcc/{8,9}.3.0 with -Werror: error: attribute ignored with [[nodiscard]]
**Describe the bug**
Nightly cuda/11.1+gcc/8.3.0 builds fail to compile with -Werror following merge of #6320 with output:
```
20:33:35 /data/jenkins-new/workspace/kokkos_semsrhel7gpu01_cuda111_gcc830_cpp17/kokkos/simd/src/Kokkos_SIMD_Scalar.hpp:138:1: error: attribute ignored [-Werror=attributes]
20:33:35    [[nodiscard]] KOKKOS_FORCEINLINE_FUNCTION friend constexpr simd operator*(
20:33:35  ^
20:33:35 /data/jenkins-new/workspace/kokkos_semsrhel7gpu01_cuda111_gcc830_cpp17/kokkos/simd/src/Kokkos_SIMD_Scalar.hpp:138:1: note: an attribute that appertains to a type-specifier is ignored
20:33:35 /data/jenkins-new/workspace/kokkos_semsrhel7gpu01_cuda111_gcc830_cpp17/kokkos/simd/src/Kokkos_SIMD_Scalar.hpp:142:1: error: attribute ignored [-Werror=attributes]
20:33:35    [[nodiscard]] KOKKOS_FORCEINLINE_FUNCTION friend constexpr simd operator/(
20:33:35  ^
20:33:35 /data/jenkins-new/workspace/kokkos_semsrhel7gpu01_cuda111_gcc830_cpp17/kokkos/simd/src/Kokkos_SIMD_Scalar.hpp:142:1: note: an attribute that appertains to a type-specifier is ignored
20:33:35 /data/jenkins-new/workspace/kokkos_semsrhel7gpu01_cuda111_gcc830_cpp17/kokkos/simd/src/Kokkos_SIMD_Scalar.hpp:146:1: error: attribute ignored [-Werror=attributes]
20:33:35    [[nodiscard]] KOKKOS_FORCEINLINE_FUNCTION friend constexpr simd operator+(
20:33:35  ^
20:33:35 /data/jenkins-new/workspace/kokkos_semsrhel7gpu01_cuda111_gcc830_cpp17/kokkos/simd/src/Kokkos_SIMD_Scalar.hpp:146:1: note: an attribute that appertains to a type-specifier is ignored
20:33:35 /data/jenkins-new/workspace/kokkos_semsrhel7gpu01_cuda111_gcc830_cpp17/kokkos/simd/src/Kokkos_SIMD_Scalar.hpp:150:1: error: attribute ignored [-Werror=attributes]
20:33:35    [[nodiscard]] KOKKOS_FORCEINLINE_FUNCTION friend constexpr simd operator-(
20:33:35  ^
20:33:35 /data/jenkins-new/workspace/kokkos_semsrhel7gpu01_cuda111_gcc830_cpp17/kokkos/simd/src/Kokkos_SIMD_Scalar.hpp:150:1: note: an attribute that appertains to a type-specifier is ignored
20:33:35 /data/jenkins-new/workspace/kokkos_semsrhel7gpu01_cuda111_gcc830_cpp17/kokkos/simd/src/Kokkos_SIMD_Scalar.hpp:154:1: error: attribute ignored [-Werror=attributes]
20:33:35    [[nodiscard]] KOKKOS_FORCEINLINE_FUNCTION friend constexpr simd operator>>(
...
```

**Reproducer** (kokkos-dev-2 or ):
```bash
module load sems-cmake sems-gcc/8.3.0 sems-cuda/11.1.0

$KOKKOS_PATH/generate_makefile.bash --with-devices=Cuda,Serial --arch=Volta70 --compiler=$KOKKOS_PATH/bin/nvcc_wrapper --cxxflags="-O3 -Wall -Wunused-parameter -Wshadow -pedantic -Werror -Wsign-compare -Wtype-limits -Wuninitialized " --cxxstandard="17" --kokkos-path=$KOKKOS_PATH --with-cuda-options=enable_lambda  --no-examples   --with-options=enable_large_mem_tests
```

dalg24:
Fix #6383 

The alternative would be to implement the operators out of class which yields less readable code IMO.
https://godbolt.org/z/sqn1WP9x1

I opted to suppress the warning before including the various SIMD implementation to avoid guarding everywhere.  I think it is OK because `<Kokkos_Core.hpp>` gets included prior to that.

### DIFF
--- a/simd/src/Kokkos_SIMD.hpp
+++ b/simd/src/Kokkos_SIMD.hpp
@@ -19,6 +19,13 @@
 
 #include <Kokkos_SIMD_Common.hpp>
 
+// suppress NVCC warnings with the [[nodiscard]] attribute on overloaded
+// operators implemented as hidden friends
+#if defined(KOKKOS_COMPILER_NVCC) && KOKKOS_COMPILER_NVCC < 1130
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wattributes"
+#endif
+
 #include <Kokkos_SIMD_Scalar.hpp>
 
 #ifdef KOKKOS_ARCH_AVX2
@@ -31,6 +38,10 @@
 
 #ifdef __ARM_NEON
 #include <Kokkos_SIMD_NEON.hpp>
+#endif
+
+#if defined(KOKKOS_COMPILER_NVCC) && KOKKOS_COMPILER_NVCC < 1130
+#pragma GCC diagnostic pop
 #endif
 
 namespace Kokkos {


crtrott (approved):

masterleinad:
Since `[[nodiscard]]` is not part of the signature and is ignored anyway, I would prefer a more localized solution that just removes `[[nodiscard]]` for the affected compiler versions (probably via a macro).

dalg24:
> Since `[[nodiscard]]` is not part of the signature and is ignored anyway, I would prefer a more localized solution that just removes `[[nodiscard]]` for the affected compiler versions (probably via a macro).

I don't like the idea to introduce a macro to substitute the `[[nodiscard]]` attribute.  For one I am not sure what name we could come up with a name that gets consensus and is not too ridiculously long.  Also there are many places scattered over 4 files and some of the attributes (the ones on math functions for instance) would not need to be replaced.

If @ldh4 agrees with you, since he maintains the SIMD sub package, I would consider changing.

ldh4 (approved):
I think this is okay to suppress the `[[nodiscard]]` problem for now.
The only thought I have is that this may cause to ignore other potential attribute errors if something other than `[[nodiscard]]` would be added into the code.

An alternative would be to encapsulate each hidden friend blocks with this suppression code, but they are scattered across 4 files for now and is probably not a desired solution for extendibility.

crtrott (approved):
Agree with this approach: since it only appears to affect hidden friends, I think we wouldn't want a general KOKKOS_IMPL_NODISCARD macro anyway, because we don't want to disable nodiscard on everything.

PR #6390, opened 2023-08-28
OpenMPTarget: Update CI to clang-17.

rgayatri23:
The PR does the following:
* Update CI testing of OpenMPTarget on NVIDIA GPUs to use clang-17
* Block tests which create ICE with the newer compilers

We want to update the CI, so that newer OpenMP extensions from clang can be used, like the ones in #6380 .
Bug reports for failing tests are being submitted to be fixed in the newer versions of the compiler.

### DIFF
--- a/algorithms/unit_tests/CMakeLists.txt
+++ b/algorithms/unit_tests/CMakeLists.txt
@@ -245,6 +245,13 @@ if(KOKKOS_ENABLE_OPENMPTARGET)
   )
 endif()
 
+# FIXME_OPENMPTARGET - remove sort test as it leads to ICE with clang/17 and above at compile time.
+if(KOKKOS_ENABLE_OPENMPTARGET STREQUAL "Clang" AND KOKKOS_CXX_COMPILER_VERSION VERSION_GREATER 16.0.0)
+    list(REMOVE_ITEM ALGO_SORT_SOURCES
+    TestSort.cpp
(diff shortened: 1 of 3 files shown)


crtrott:
So @rgayatri23 wait until 17.0.0 is actually out or do you want to only update to 16? I am happy to wait with this for real 17.0.0 I would not like to use an RC, but maybe we could just for a few weeks and then update to the real thing? @dalg24 thoughts?

rgayatri23:
> So @rgayatri23 wait until 17.0.0 is actually out or do you want to only update to 16? I am happy to wait with this for real 17.0.0 I would not like to use an RC, but maybe we could just for a few weeks and then update to the real thing? @dalg24 thoughts?

Yeah I can update to 16.0.0 too and that should work too.
But I thought 17.x was out, I see a tag of [release/17.x](https://github.com/llvm/llvm-project/tree/release/17.x)
Am I missing anything?

Rombur:
> But I thought 17.x was out,

The official release is mid-September: https://llvm.org/

crtrott:
From discussions on discord it sounds like the release may be getting pushed by two weeks, they may need another 2 (one this week another in two weeks) RCs.

rgayatri23:
Closing this PR so that I can re-open one with 17 release in a few weeks.

PR #6391, opened 2023-08-28
Fix nightly (gcc 13)

Rombur:
The variable is initialized in a `switch` but the compiler complains that there may be cases where the variable is not initialized.

### DIFF
--- a/algorithms/unit_tests/TestStdAlgorithmsTeamFindEnd.cpp
+++ b/algorithms/unit_tests/TestStdAlgorithmsTeamFindEnd.cpp
@@ -186,7 +186,7 @@ void test_A(const bool sequencesExist, std::size_t numTeams,
     auto rowSearchedSeq =
         Kokkos::subview(searchedSequencesView_h, i, Kokkos::ALL());
 
-    std::size_t stdDistance;
+    std::size_t stdDistance = std::numeric_limits<std::size_t>::max();
     const std::size_t beginEndDistance = KE::distance(rowFromBegin, rowFromEnd);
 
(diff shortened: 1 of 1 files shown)


dalg24:
Does having a default statement that aborts also get rid of the warning?

Rombur:
> Does having a default statement that aborts also get rid of the warning?

yes.

masterleinad:
> Does having a default statement that aborts also get rid of the warning?

I would prefer that.

fnrizzi (approved):
i am fine with either this or the default suggestion

dalg24:
Please prefer the default case.

Rombur:
> apiId doesn't have an enum type so the compiler can't know the allowed values.

Yes, it can know it. The possible values of `apild` are known are compile time:
```
      for (int apiId : {0, 1, 2, 3}) {
        test_A<LayoutTag, ValueType>(sequencesExist, numTeams, numCols, apiId);
```

dalg24:
What if you make the `apild` an enum?

fnrizzi:
@dalg24 @masterleinad aren't we just complicated this too much for just a test?

Rombur:
> What if you make the apild an enum?

So we can start bikeshedding on the name of the `enum` 😛 I can do it if you insist but I don't see how it improves the code.

dalg24:
AppVeyor failure seems unrelated.  Log stopped while running test 28/30 and all success up til then.
Couple failures on Jenkins are `toomanyrequests: You have reached your pull rate limit`.

PR #6392, opened 2023-08-28
Cuda: Allocate using the correct device

masterleinad:
Part of #6091. This pull request introduces additional constructors for `CudaSpace`, `CudaUVMSpace`, and `CudaHostPinnedSpace` that allow using a specific device and stream for allocations.
We don't use execution space instances here since we want to use the same functions for allocations in the initialization of execution space instances. This also means that we can't use the wrappers we introduced in #6299. Note that `cudaFree` also seems to work when called from the "wrong" device.

Additionally, this pull request cleans up the implementation of `impl_allocate_common` some. There was no good reason to explicitly use the singleton instead of the execution space instance argument (that might have been constructed from the singleton).


core/src/Cuda/Kokkos_CudaSpace.cpp
```diff
@@ -161,52 +174,37 @@ void *CudaSpace::allocate(const char *arg_label, const size_t arg_alloc_size,
...
 namespace {
-void *impl_allocate_common(const Cuda &exec_space, const char *arg_label,
-                           const size_t arg_alloc_size,
+void *impl_allocate_common(const int device_id,
+                           [[maybe_unused]] const cudaStream_t stream,
+                           const char *arg_label, const size_t arg_alloc_size,
                            const size_t arg_logical_size,
                            const Kokkos::Tools::SpaceHandle arg_handle,
-                           bool exec_space_provided) {
+                           [[maybe_unused]] bool exec_space_provided) {
```
crtrott, line 182: should we fine a new name for this in particular since we are not passing an exec space anymore? Maybe: `do_device_sync` (which would have the opposite meaning of the current one) or `stream_sync_only` or something like that?

crtrott (changes requested):
I don't think it matters here, but do we need to think about giving memory spaces the same ref count semantics as execution spaces?

For the spaces here I don't think we have an issue since the only stuff getting copied is an int and a stream (basically another int).

core/src/Cuda/Kokkos_CudaSpace.cpp
masterleinad: Sure, see a25a2f0.

masterleinad:
> I don't think it matters here, but do we need to think about giving memory spaces the same ref count semantics as execution spaces?

We normally construct memory spaces on the fly anyway and doing that is very cheap. So I'm not sure if this is problematic. Of course, we would run into problems if a MemorySpace space instance would outlive its stream and would then be used for allocating/deallocating data (without using the overloads using an execution space instance). Giving the memory space ref count semantics wouldn't even help but we would need to give it owning semantics over the stream as well. Ultimately, we only provide the new constructor here to be able to use `SharedAllocationRecord` for internal allocations in the execution space instance initialization so we could rather make it an internal member function. Or we just document the semantics well. 🙂

masterleinad:
Only GCC-8.4.0 is failing because `serial.local_deepcopy_rangepolicy_layoutleft` is timing out. That looks unrelated/spurious.

core/src/Cuda/Kokkos_CudaSpace.cpp
```diff
@@ -161,52 +174,36 @@ void *CudaSpace::allocate(const char *arg_label, const size_t arg_alloc_size,
...
   if (arg_alloc_size >= memory_threshold_g) {
-    if (exec_space_provided) {
-      error_code =
-          exec_space.impl_internal_space_instance()->cuda_malloc_async_wrapper(
-              &ptr, arg_alloc_size);
-      exec_space.fence("Kokkos::Cuda: backend fence after async malloc");
+    error_code = cudaMallocAsync(&ptr, arg_alloc_size, stream);
+
+    if (stream_sync_only) {
+      KOKKOS_IMPL_CUDA_SAFE_CALL(cudaStreamSynchronize(stream));
```
dalg24, line 194: There is an `Impl::cuda_stream_synchronize`

dalg24: Also I don't fully understand the error code handling in that function.

core/src/Cuda/Kokkos_CudaSpace.hpp
```diff
@@ -68,6 +68,7 @@ class CudaSpace {
   /*--------------------------------*/
 
   CudaSpace();
+  CudaSpace(int cuda_device, cudaStream_t cuda_stream);
```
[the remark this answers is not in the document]
masterleinad: > Are these new constructors intended to be exposed to users or are these still considered implementation only?

They should ideally be implementation-only. There shouldn't be much of a reason for users to call this one.
If allocating via `Kokkos::View` we pass in the correct execution space instance already. Just `Kokkos::alloc` doesn't yet provide a way to specify the desired device/stream.

> Also please comment about the argument ordering. Is there a rational for device_id first and then stream?

That's the same as in #6361. To me it makes more sense to specify arguments in the order from coarse to fine.

core/src/Cuda/Kokkos_CudaSpace.cpp
masterleinad: > There is an Impl::cuda_stream_synchronize

Yes, but that would set the device to the default one which would be wrong here.

masterleinad: > Also I don't fully understand the error code handling in that function.

Which function? This one? I tried to preserve behavior.

masterleinad: The only way this function could fail is if an invalid stream was provided. Maybe it's better to guard with
```
if (error_code == cudaSuccess) {
if (stream_sync_only) 
[...]
else
[...]
}
```
instead.

masterleinad: 00f0a73747d89c4402678f5c3360c7195d245e84

crtrott (approved):
I don't see an obvious reason to do the design differently than this (e.g. store an execution space instead of device id or stream). We also could maybe have a constructor independently from execution space instance.

But maybe we can gain that experience and insight with further work on it. Considering that there are no public interface changes here I am ok to go forward like it is.

dalg24 (approved):
Reluctantly approving so we move forward

core/src/Cuda/Kokkos_Cuda_Instance.cpp
```diff
@@ -26,10 +26,10 @@
...
 #include <Kokkos_Core.hpp>
 
-#include <Cuda/Kokkos_Cuda_Error.hpp>
-#include <Cuda/Kokkos_Cuda_BlockSize_Deduction.hpp>
-#include <Cuda/Kokkos_Cuda_Instance.hpp>
-#include <Cuda/Kokkos_Cuda_UniqueToken.hpp>
+//#include <Cuda/Kokkos_Cuda_Error.hpp>
+//#include <Cuda/Kokkos_Cuda_BlockSize_Deduction.hpp>
+//#include <Cuda/Kokkos_Cuda_Instance.hpp>
+//#include <Cuda/Kokkos_Cuda_UniqueToken.hpp>
```
dalg24, line 32: Great

Issue #6330, opened 2023-08-03
Feature request for SIMD rounding operations


vbrunini:
Having floor/ceil/round/trunc available for the SIMD types would be useful.

PR #6393, opened 2023-08-29
simd: add floor, ceil, round, trunc operations

ldh4:
Requested features from #6330.

This PR adds `floor`, `ceil`, `round`, and `trunc` operations for ~~supported simd `double` and `float` types~~ for all supported simd types.

- `simd Kokkos::Experimental::floor(simd const& a)`
- `simd Kokkos::Experimental::ceil(simd const& a)`
- `simd Kokkos::Experimental::round(simd const& a)`
- `simd Kokkos::Experimental::trunc(simd const& a)`

### DIFF
--- a/simd/src/Kokkos_SIMD_AVX2.hpp
+++ b/simd/src/Kokkos_SIMD_AVX2.hpp
@@ -1499,7 +1499,7 @@ class simd<std::uint64_t, simd_abi::avx2_fixed_size<4>> {
   }
 };
 
-[[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
+KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
 simd<std::int64_t, simd_abi::avx2_fixed_size<4>>::simd(
     simd<std::uint64_t, simd_abi::avx2_fixed_size<4>> const& other)
     : m_value(static_cast<__m256i>(other)) {}
--- a/simd/src/Kokkos_SIMD_AVX2.hpp
+++ b/simd/src/Kokkos_SIMD_AVX2.hpp
@@ -635,103 +635,139 @@ class simd<double, simd_abi::avx2_fixed_size<4>> {
 
 }  // namespace Experimental
 
-KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
-Experimental::simd<double, Experimental::simd_abi::avx2_fixed_size<4>> copysign(
-    Experimental::simd<double,
-                       Experimental::simd_abi::avx2_fixed_size<4>> const& a,
-    Experimental::simd<double,
-                       Experimental::simd_abi::avx2_fixed_size<4>> const& b) {
+[[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION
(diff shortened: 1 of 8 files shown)


dalg24:
Please explain why we don't provide overloads for integral type specializations of simd.

ldh4:
> Please explain why we don't provide overloads for integral type specializations of simd.

The primary reason for missing these implementation is because these intrinsics are only provided for floating point types.
As for why corresponding overloads for integral type specializations are not provided may need a bit more discussion, in terms of how it should be handled.

Currently, there already are functions that are only for floating point type: `operator divide`, `cbrt`, `exp`, `log`, etc. For operator divide, a fallback implementation is placed in `Kokkos_SIMD_Common.hpp` which is meant to be for all integral types. For `cbrt`, `exp`, `log`, implementations are only available for floating point types in `AVX2` and `AVX512`.

For now, I'm thinking of adding simple fallback functions for `floor`, `ceil`, `round` and `trunc` for integral types.

dalg24:
But presumably they are intrinsics to convert packed integers into floating-point numbers so the missing overloads would be implemented by composing conversion and floating-point intrinsics.

ldh4:
Added overloads for integral types [3485d2a](https://github.com/kokkos/kokkos/pull/6393/commits/3485d2af818538a5aa34436d620d59609786bea3). Since we eventually want to remove all fallback implementations, it does make more sense to provide overloads for each integral specialization for this as well instead of adding in more fallbacks. However, because simd `floor`, `cell`, `round` and `trunc` are floating point operations, implemented calling these with integral types to simply return input argument.

ldh4:
Changed floating point ops with integral simd types to return as `double` simd type, as discussed during the weekly meeting.

Rombur:
This PR fails for OpenACC

```
NVC++-S-1061-Procedures called in a compute region must have acc routine information - rint (/var/jenkins/workspace/Kokkos/simd/unit_tests/TestSIMD.cpp: 188)
NVC++-S-1061-Procedures called in a compute region must have acc routine information - rint (/var/jenkins/workspace/Kokkos/simd/unit_tests/TestSIMD.cpp: 188)
NVC++-S-1061-Procedures called in a compute region must have acc routine information - rint (/var/jenkins/workspace/Kokkos/simd/unit_tests/TestSIMD.cpp: 188)
NVC++/x86-64 Linux 23.7-0: compilation completed with severe errors
make[2]: *** [simd/unit_tests/CMakeFiles/Kokkos_UnitTest_SIMD.dir/build.make:90: simd/unit_tests/CMakeFiles/Kokkos_UnitTest_SIMD.dir/TestSIMD.cpp.o] Error 2
make[1]: *** [CMakeFiles/Makefile2:3944: simd/unit_tests/CMakeFiles/Kokkos_UnitTest_SIMD.dir/all] Error 2
```

nliber (approved):

ldh4:
Retest this please.

ajpowelsnl (approved):

ldh4:
Retest this please.

masterleinad:
The indentation is off. 🙂

crtrott:
GCC 8.2 failed with some nodiscard attribute issue.

crtrott (approved):

PR #6394, opened 2023-08-29
Add KOKKOS_ARCH_ARM_NEON 

masterleinad:
Based on top of #6188. Currently, we are checking for `__ARM_NEON` to enable the respective SIMD type. This is different from how we do it for all other SIMD types and this pull request makes the behavior more uniform by introducing a non-cache `KOKKOS_ARCH_ARM_NEON` CMake variable and preprocessor define that works analogous to `KOKKOS_ARCH_AVX512XEON` and similar.
Technically, this could be considered as backward-incompatible since users might have relied on the autodetection for the intrinsics.


ldh4 (approved):
Looks fine to me.

Rombur (approved):

dalg24:
What about auto detection?
https://github.com/kokkos/kokkos/blob/8181d707581e32766c9f8b6f35da630ab3e7cb86/cmake/kokkos_arch.cmake#L486-L488

dalg24:
> What about auto detection?
> 
> https://github.com/kokkos/kokkos/blob/8181d707581e32766c9f8b6f35da630ab3e7cb86/cmake/kokkos_arch.cmake#L486-L488

Nevermind I see you added it.  Sorry about the noise.

dalg24 (approved):

dalg24:
The CUDA 11.0 build failure is unrelated (SIMD issue being fixed in #6485).  Both HIP builds timed out.

PR #6401, opened 2023-09-01
Deprecate `{Cuda,Hip}({cuda,hip}Stream_t, bool)`

dalg24:
Whether to manage stream should not have been exposed to the user.
This was only meant as a "private" constructor to implement `Experimental::partition_space()`.
Here I suggest we mark these constructors `Cuda(cudaStream_t, bool)` and `Hip(hipStream_t, bool)` as deprecated and replace them with constructors that take an `Impl::ManageStream::{yes, no}` enumerator instead of the boolean argument.

### DIFF
--- a/core/src/Cuda/Kokkos_Cuda.hpp
+++ b/core/src/Cuda/Kokkos_Cuda.hpp
@@ -80,6 +80,9 @@ struct CudaDispatchProperties {
   CudaLaunchMechanism launch_mechanism = l;
 };
 }  // namespace Experimental
+
+enum class ManageStream : bool { no, yes };
+
 }  // namespace Impl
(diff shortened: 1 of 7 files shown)


dalg24:
Conflicts with #6361

dalg24:
OpenMPTarget failure is clearly unrelated
```
4: [ RUN      ] openmptarget.partitioning_by_vector
4:   sum1
4:     Which is: 7642095
4:   sum2
4:     Which is: 0
4: [  FAILED  ] openmptarget.partitioning_by_vector (0 ms)
```
Everything else passed.

crtrott (approved):

fnrizzi (approved):

PR #6402, opened 2023-09-01
Use `cudaMallocAsync` by default

dalg24:
We decided we would change the default value of the `Kokkos_ENABLE_IMPL_CUDA_MALLOC_ASYNC` option from `OFF` to `ON` a couple weeks ago at the developer meeting but somehow we didn't get around to implement it.  So here it is.

This option was introduced in #4233 because of issues downstream with UCX.  It has been 2 years now and we expect this has been resolved on the UCX side.  We preserve the option a bit longer so it is possible to revert to the old behavior.
I also introduce a `disable_malloc_async` option so it can be disabled with the generated makefiles.

Drive-by change:  Say whether `KOKKOS_ENABLE_IMPL_CUDA_MALLOC_ASYNC` is defined or not from `Kokkos::print_configuration()`.

### DIFF
--- a/Makefile.kokkos
+++ b/Makefile.kokkos
@@ -30,7 +30,7 @@ KOKKOS_TRIBITS ?= "no"
 KOKKOS_STANDALONE_CMAKE ?= "no"
 
 # Default settings specific options.
-# Options: force_uvm,use_ldg,rdc,enable_lambda,enable_constexpr
+# Options: force_uvm,use_ldg,rdc,enable_lambda,enable_constexpr,disable_malloc_async
 KOKKOS_CUDA_OPTIONS ?= ""
 
 # Options: rdc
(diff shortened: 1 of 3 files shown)


dalg24:
cc @maxpkatz @matt-stack @lfmeadow @BenWibking

dalg24:
Adding a reference to https://github.com/kokkos/kokkos/issues/6344#issuecomment-1671842325 

@ndellingwood if this goes in we would want to update the nightlies to have also at least one build with `-D Kokkos_ENABLE_IMPL_CUDA_MALLOC_ASYNC=OFF`.

crtrott (approved):

Makefile.kokkos
```diff
@@ -30,7 +30,7 @@ KOKKOS_TRIBITS ?= "no"
 KOKKOS_STANDALONE_CMAKE ?= "no"
 
 # Default settings specific options.
-# Options: force_uvm,use_ldg,rdc,enable_lambda,enable_constexpr
+# Options: force_uvm,use_ldg,rdc,enable_lambda,enable_constexpr,disable_malloc_async
```
fnrizzi, line 33: why are you making this an optional to "disable" it? 
I would argue that for consistency wit the the `Kokkos_ENABLE_IMPL_CUDA_MALLOC_ASYNC` we should call this `enable_malloc_async` so that the user decided to enable it (not enabling the disabling )

Makefile.kokkos
```diff
@@ -689,6 +690,12 @@ ifeq ($(KOKKOS_INTERNAL_USE_CUDA), 1)
   ifeq ($(KOKKOS_INTERNAL_COMPILER_CLANG), 1)
     tmp := $(call kokkos_append_header,"$H""define KOKKOS_IMPL_CUDA_CLANG_WORKAROUND")
   endif
+
+  ifeq ($(KOKKOS_INTERNAL_CUDA_DISABLE_MALLOC_ASYNC), 0)
+    tmp := $(call kokkos_append_header,"$H""define KOKKOS_ENABLE_IMPL_CUDA_MALLOC_ASYNC")
+  else
+    tmp := $(call kokkos_append_header,"/* $H""undef KOKKOS_ENABLE_IMPL_CUDA_MALLOC_ASYNC */")
+  endif
```
fnrizzi, line 698: if you take my suggestion above this of course change

Makefile.kokkos
dalg24: Because our internal logic just check for existence of the substrings so you can't really do `disable_malloc_async=0`.  And even if we did parse that properly I find it awkward to use.

fnrizzi (approved):

ndellingwood:
> @ndellingwood if this goes in we would want to update the nightlies to have also at least one build with `-D Kokkos_ENABLE_IMPL_CUDA_MALLOC_ASYNC=OFF`.

@dalg24 nightly test has been updated to set  `-D Kokkos_ENABLE_IMPL_CUDA_MALLOC_ASYNC=OFF`.

stanmoore1:
This breaks CUDA-aware MPI for CUDA 11.2 on LLNL Lassen/Sierra and SNL Vortex. And >11.2 is already broken for CUDA-aware MPI, so users will have to revert to 11.1 or below which turns off CUDA_MALLOC_ASYC by default.

PR #2825, opened 2020-03-03
Address issue 2652: not checking for host init in CUDA

crtrott:
This is for a specific way of how Legion wants to interact with Kokkos

PR #6409, opened 2023-09-06
Drop check that the host backend is initialized before the device one

dalg24:
These assertions were only present in the Cuda, HIP, and SYCL backend, presumably just copied over without questioning whether it was necessary.
In #2825, we had guarded the Cuda check with `#ifndef KOKKOS_IMPL_TURN_OFF_CUDA_HOST_INIT_CHECK` so that Legion is able to disable it.  On the issue this PR was addressing it is already suggested that this might very well be unnecessary.  I looked at the `{Cuda,HIP}::impl_initialize` and I could find anything that would actually require the host execution space to be initialized before.  I also tried
```diff
diff --git a/core/src/Serial/Kokkos_Serial.cpp b/core/src/Serial/Kokkos_Serial.cpp
index 6d55dffb7..271b92e39 100644
--- a/core/src/Serial/Kokkos_Serial.cpp
+++ b/core/src/Serial/Kokkos_Serial.cpp
@@ -171,7 +171,7 @@ const char* Serial::name() { return "Serial"; }
 namespace Impl {

 int g_serial_space_factory_initialized =
-    initialize_space_factory<Serial>("100_Serial");
+    initialize_space_factory<Serial>("999_Serial");

 }  // namespace Impl
```
to force the Serial backend to be initialized after HIP and all tests were passing.

**TLDR:** These check are unnecessary and are dictate the order in which to initialize backends.  I suggest to remove them.

### DIFF
--- a/core/src/Cuda/Kokkos_Cuda_Instance.cpp
+++ b/core/src/Cuda/Kokkos_Cuda_Instance.cpp
@@ -376,15 +376,6 @@ void CudaInternal::initialize(cudaStream_t stream, bool manage_stream) {
   was_initialized = true;
   if (is_initialized()) return;
 
-#ifndef KOKKOS_IMPL_TURN_OFF_CUDA_HOST_INIT_CHECK
-  if (!HostSpace::execution_space::impl_is_initialized()) {
-    const std::string msg(
-        "Cuda::initialize ERROR : HostSpace::execution_space is not "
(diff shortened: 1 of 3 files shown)


cz4rs (approved):
Looks good to me.

fnrizzi (approved):

PR #6410, opened 2023-09-06
Remove unused `HIPInternal` static data members

dalg24:
Drop `HIPInternal::{m_maxSharedWords,m_hipArch}` static data members.
`m_maxSharedWords`'s equivalent in the Cuda backend was removed in #6408


fnrizzi (approved):
approving assuming CI passes

PR #6411, opened 2023-09-06
Improve `{Cuda,HIP,SYCL}Internal::initialize()`

dalg24:
`initialized()` being `false` should be a precondition for calling `{Cuda,HIP,SYCL}Internal::initialize()`.
Also if these precondition checks ever failed, that would denote a bug in Kokkos, not an incorrect usage like when we check whether an execution space is being constructed after Kokkos was finalized.

This change simplifies the logic and improves readability.

### DIFF
--- a/core/src/Cuda/Kokkos_Cuda_Instance.cpp
+++ b/core/src/Cuda/Kokkos_Cuda_Instance.cpp
@@ -371,32 +371,24 @@ void CudaInternal::fence() const {
 }
 
 void CudaInternal::initialize(cudaStream_t stream, bool manage_stream) {
+  KOKKOS_EXPECTS(!is_initialized());
+
   if (was_finalized)
     Kokkos::abort("Calling Cuda::initialize after Cuda::finalize is illegal\n");
   was_initialized = true;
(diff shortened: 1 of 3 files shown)


core/src/Cuda/Kokkos_Cuda_Instance.cpp
```diff
@@ -371,32 +371,24 @@ void CudaInternal::fence() const {
...
+  KOKKOS_EXPECTS(!is_initialized());
+
   if (was_finalized)
     Kokkos::abort("Calling Cuda::initialize after Cuda::finalize is illegal\n");
   was_initialized = true;
-  if (is_initialized()) return;
-
-  const bool ok_init = nullptr == m_scratchSpace || nullptr == m_scratchFlags;
 
-  if (ok_init) {
```
dalg24, line 381: For the record
https://github.com/kokkos/kokkos/blob/c692a816df60691482950c54bc88ea8ef51e0db8/core/src/Cuda/Kokkos_Cuda_Instance.hpp#L157-L159

crtrott (approved):

PR #6413, opened 2023-09-06
Drop pre-Kepler logic in `Cuda::impl_initialize`

dalg24:
Fermi microarchitecture and previous generations are not supported by the minimum CUDA version that we require.
Assuming that such a GPU can even be listed as visible device by the CUDA runtime, we would error out here
https://github.com/kokkos/kokkos/blob/7b4d0a6f70e17a89836d949fa16ca56642676fd7/core/src/Cuda/Kokkos_Cuda_Instance.cpp#L751-L762
which is good enough.

Rational for change: Removing that code simplifies our backend initialization logic.


crtrott (approved):

dalg24:
Retest this please

dalg24:
Failure (before re-running) was an issue with the file system on one of the CUDA build.
After re-running we see a few more machine issues.
Deciding to ignore.

PR #6417, opened 2023-09-07
Drop check whether device supports unified addressing

dalg24:
This check is obsolete and only prints a warning when it should probably cause program termination if it ever failed because everything would crash and burn (for example we use `cudaMemcpyDefault` in all our `cudaMemCpy*` calls)

I suggest to simply remove it.


Rombur (approved):

dalg24:
Retest this please

PR #6422, opened 2023-09-08
re-enable unit tests for sort and random via makefile

fnrizzi:
In https://github.com/kokkos/kokkos/pull/6081, where all files for sorting and random were reorganized, I accidentally forgot to update the `Makefile` inside the unit tests accordingly. To fix that, this PR does this: 

1. it adds those files back so that all sort and random tests are rebuilt when using the makefile

2. simplifies `TestBinSortB.hpp`: this file was using duplicated helper functions because those were not accessible in develop at that time. Now, they are accessible in develop, so we get rid of those and simply use them from `TestStdAlgorithmsCommon.hpp` 

3. because of how the Makefile works (for example if seria/cuda are enabled, the test is enabled for both cuda and serial by passing to the test the actual exespace space to use via a define statement), some code needed to be fixed to allow this behavior to work properly (before it was just assumed to use the default execution space in all cases)


cz4rs (approved):
Looks good to me assuming the CI passes.

masterleinad (approved):
Looks OK to me.

cz4rs (approved):
Still looks good to me.

Issue #6254, opened 2023-07-03
compilation(ViewMapping): `nvcc 12.1` warns about `pointless comparison of unsigned integer with zero` for zero-rank view


romintomasetti:
**Describe the bug**

We recently updated our Trilinos snapshot, and thereby got a more recent Kokkos version.

When compiling with Cuda, at some point we get `error #186-D: pointless comparison of unsigned integer with zero` in the following function:
https://github.com/kokkos/kokkos/blob/13948ecf8f0289384ee3a3e53561bbd57cde6080/core/src/impl/Kokkos_ViewMapping.hpp#L3661-L3688

The line that causes the issue is `for (unsigned int i = 0; i < src.Rank; i++)` when `Rank` is zero.
1. I think @masterleinad modified this line in be65fe4298756153373dc1138ea22b43721f4cb5 and went from `int` to `unsigned int` for the loop variable (I think that's why we hadn't any problem before)
2. Only `nvcc V12.1.105` warns, `g++ 11.3.0` or `clang 16` for instance are fine with the code (and I think `nvcc` should as well because `Rank` is a constant expression so it should simply figure out that the `for` can be skipped)
3. I think the proposed fix is a good fix in that for zero rank views, `assignable_layout_check` should always return `true` (which was already the case before), but it should be clearly "stated" by reading the code. My first approach was to have a `if constexpr (src.Rank == 0) return true;` in the function, but `nvcc` would still complain. `nvcc` would stop complaining only if I wrote `if constexpr (rank == 0) return true; else { ... }`. I'm fine if you prefer the solution with `if/else` :smile:

Proposed fix:
```diff
diff --git a/core/src/impl/Kokkos_ViewMapping.hpp b/core/src/impl/Kokkos_ViewMapping.hpp
index 877004ae4..76c113834 100644
--- a/core/src/impl/Kokkos_ViewMapping.hpp
+++ b/core/src/impl/Kokkos_ViewMapping.hpp
@@ -3657,9 +3657,22 @@ class ViewMapping<
   using DstType   = ViewMapping<DstTraits, void>;
   using SrcType   = ViewMapping<SrcTraits, void>;
 
+  //! A zero-rank view is always assignable in this case.
+  template <size_t src_rank = SrcType::Rank,
+            std::enable_if_t<(src_rank == 0)>* = nullptr>
   KOKKOS_INLINE_FUNCTION
   static bool assignable_layout_check(DstType&,
-                                      const SrcType& src)  // Runtime check
[...]

Issue #186, opened 2016-02-09
Kokkos Threads Backend impl_shared_alloc Broken on Intel 16.1 (Shepard Haswell)


nmhamster:
Doing standard test in develop branch on Shepard with Intel 16.1 compiler.

```
$ ./KokkosCore_UnitTest_Threads
[==========] Running 44 tests from 1 test case.
[----------] Global test environment set-up.
[----------] 44 tests from threads
Kokkos::Threads KOKKOS_HAVE_PTHREAD threads[2] threads_per_numa[2] threads_per_core[1] ReduceScratch[0] SharedScratch[0]
 Thread[ 1 : 0.0 ] Fan{ [ 0 : 0.0 ] } is_process
 Thread[ 0 : 0.0 ] Fan{ }
[ RUN      ] threads.init
[       OK ] threads.init (0 ms)
[ RUN      ] threads.dispatch
[       OK ] threads.dispatch (23 ms)
[ RUN      ] threads.impl_shared_alloc
Segmentation fault
```

Backtrace Information:
```
Program received signal SIGSEGV, Segmentation fault.
0x00000000005f469f in Kokkos::Experimental::Impl::SharedAllocationRecord<void, void>::is_sane(Kokkos::Experimental::Impl::SharedAllocationRecord<void, void>*) ()
(gdb) bt
#0  0x00000000005f469f in Kokkos::Experimental::Impl::SharedAllocationRecord<void, void>::is_sane(Kokkos::Experimental::Impl::SharedAllocationRecord<void, void>*) ()
#1  0x0000000000483e64 in void Test::test_shared_alloc<Kokkos::HostSpace, Kokkos::Threads>() ()
#2  0x00000000005dd77d in void testing::internal::HandleExceptionsInMethodIfSupported<testing::Test, void>(testing::Test*, void (testing::Test::*)(), char const*) ()
#3  0x0000000000578899 in testing::Test::Run() ()
#4  0x0000000000576e16 in testing::TestInfo::Run() ()
#5  0x0000000000575db8 in testing::TestCase::Run() ()
#6  0x000000000057484a in testing::internal::UnitTestImpl::RunAllTests() ()
#7  0x00000000005694b6 in testing::UnitTest::Run() ()
#8  0x000000000055eeba in main ()
```

crtrott:
Looking into it

crtrott:
I fixed the actual crash, but the underlying reason for the crash is unexpected (incorrect) behaviour. So the test still fails.

[...]

PR #6423, opened 2023-09-10
(fix for issue #6254) Avoid `#186-D pointless comparison warning`.

maartenarnst:
This is a follow up of issue #6254.

In this issue, @romintomasetti had raised that `nvcc` gives a `#186-D pointless comparison warning` for a comparison of an `unsigned int` with the rank of a zero-rank view.

In #6254, @ajpowelsnl had asked to open a PR with a fix proposed by @dalg24 (cast to `int`), which was an alternative to a fix proposed by @romintomasetti (new specialisation of `assignable_layout_check` for case of zero rank view).

And so this is the PR with @dalg24's proposed fix :)


fnrizzi (approved):

dalg24:
Ignoring HIP builds timeout

PR #6430, opened 2023-09-11
[SYCL][Reduce/Scan] Group counter should use at least memory_order::acq_rel

aelovikov-intel:
From https://en.cppreference.com/w/cpp/atomic/memory_order:

> Atomic operations tagged memory_order_relaxed are not synchronization
> operations; they do not impose an order among concurrent memory
> accesses. They only guarantee atomicity and modification order
> consistency.

Yet we want to use that counter exactly for the synchronization purposes - to decide which WG finished last and should perform the final step.

The same issue had been fixed in DPC++ at https://github.com/intel/llvm/pull/8058.

### DIFF
--- a/core/src/SYCL/Kokkos_SYCL_ParallelReduce_MDRange.hpp
+++ b/core/src/SYCL/Kokkos_SYCL_ParallelReduce_MDRange.hpp
@@ -215,7 +215,7 @@ class Kokkos::Impl::ParallelReduce<CombinedFunctorReducerType,
                     value_count, reducer, false, wgroup_size);
 
                 if (local_id == 0) {
-                  sycl::atomic_ref<unsigned, sycl::memory_order::relaxed,
+                  sycl::atomic_ref<unsigned, sycl::memory_order::acq_rel,
                                    sycl::memory_scope::device,
                                    sycl::access::address_space::global_space>
                       scratch_flags_ref(*scratch_flags);
@@ -260,7 +260,7 @@ class Kokkos::Impl::ParallelReduce<CombinedFunctorReducerType,
                     device_accessible_result_ptr, reducer, false, wgroup_size);
 
                 if (local_id == 0) {
-                  sycl::atomic_ref<unsigned, sycl::memory_order::relaxed,
+                  sycl::atomic_ref<unsigned, sycl::memory_order::acq_rel,
                                    sycl::memory_scope::device,
                                    sycl::access::address_space::global_space>
                       scratch_flags_ref(*scratch_flags);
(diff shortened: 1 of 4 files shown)


masterleinad (dismissed):
Looks reasonable.

masterleinad:
OK to test.

dalg24 (commented):
What about this one
https://github.com/kokkos/kokkos/blob/9081d366c4a67a5900a512993cfd79a884b73674/core/src/SYCL/Kokkos_SYCL_ParallelScan_Range.hpp#L182-L186

dalg24:
Also I don't get it.
Consider
https://github.com/kokkos/kokkos/blob/9081d366c4a67a5900a512993cfd79a884b73674/core/src/SYCL/Kokkos_SYCL_ParallelReduce_Team.hpp#L211-L215
which this PR suggest to change to use `memory_order::acq_rel`.
Why is relaxed not sufficient?  We only atomic increment of the counter.

dalg24:
It is not like with relaxed ordering will break https://github.com/kokkos/kokkos/blob/9081d366c4a67a5900a512993cfd79a884b73674/core/src/SYCL/Kokkos_SYCL_ParallelReduce_Team.hpp#L218
What am I missing?

crtrott (changes requested):
I think this is unnecessary. I am pretty sure this is not needed at all, and the previous memory order was sufficient for the use case here.

aelovikov-intel:
> It is not like with relaxed ordering will break

That's exactly what happens - we see random failures on 
```core/unit_test/Kokkos_CoreUnitTest_SYCL1B --gtest_filter=sycl.mdrange_parallel_reduce_primitive_types```
when running on one of our Intel GPUs. Corresponding issue in DPC++ was reported by one of the IGC (https://github.com/intel/intel-graphics-compiler) developers. What I was told is that `relaxed` ordering only requires the atomic counter to be synced between the threads, but not the actual reduction data. Having `acq_rel` results in the compiler emitting extra fences syncing the memory.

Another quote from the https://en.cppreference.com/w/cpp/atomic/memory_order:

> Release-Acquire ordering
> If an atomic store in thread A is tagged memory_order_release, an atomic load in thread B from the same variable is tagged memory_order_acquire, and the load in thread B reads a value written by the store in thread A, then the store in thread A synchronizes-with the load in thread B.
>
> All memory writes (including non-atomic and relaxed atomic) that happened-before the atomic store from the point of view of thread A, become visible side-effects in thread B.

We need this synchronization of the whole memory, not just the atomic counter that is being incremented, to be visible across all WGs.

crtrott:
We came to the conclusion that this (or something similar) is required.

masterleinad (approved):
Since this has passed testing already, let's move forward with this approach.

crtrott (approved):

PR #6433, opened 2023-09-12
Initialize m_num_scratch_locks for Cuda parallel_for TeamPolicy

### LINKED ISSUE #6398: Invalid memory access when using parallel_for and level 1 scratch memory.
**Describe the bug**
I recently updated to Kokkos 4.1.00 (1a3ea28), and since then I've started running into issues with invalid memory accesses for `Kokkos::parallel_for` loops. If I downgrade to 4.0.00 (aa1f48f), the problem disappears, so it appears to be due to a recent change (or bug?) in the code.

It seems that a `parallel_for` loop of sufficient size with level 1 scratch memory on CUDA will eventually produce an error similar to the following:
`cudaFreeHost(constantMemHostStaging) error( cudaErrorIllegalAddress): an illegal memory access was encountered /home/jacob/projects/experiments/scratch/kokkos/core/src/Cuda/Kokkos_Cuda_Instance.cpp:589`
In my case, the minimum league size producing the error is 7535, though I have no reason to believe that is some magic number fixed across all problem sizes, configurations, and machines.

1. compiler: nvcc 11.7.64
2. Kokkos release or commit used (i.e., the sha1 number): Kokkos 4.1.00 (1a3ea28)
3. platform and backend: WSL2 Ubuntu 20.04, CUDA backend
4. cmake configure command: `cmake -DKokkos_ENABLE_CUDA=ON -DKokkos_ARCH_AMPERE86=ON -DKokkos_ARCH_ZEN2=ON ../`
5. output from cmake command:
```
-- The C compiler identification is GNU 9.4.0
-- The CXX compiler identification is GNU 9.4.0
-- Check for working C compiler: /usr/bin/cc
-- Check for working C compiler: /usr/bin/cc -- works
-- Check for working CXX compiler: /usr/bin/c++
-- Check for working CXX compiler: /usr/bin/c++ -- works
-- Setting default Kokkos CXX standard to 17
-- Kokkos version: 4.1.00
-- The project name is: Kokkos
-- Using internal gtest for testing
-- Configured git information in /home/jacob/projects/experiments/scratch/build/generated/Kokkos_Version_Info.cpp
-- Compiler Version: 11.7.64
-- kokkos_launch_compiler (/home/jacob/projects/experiments/scratch/kokkos/bin/kokkos_launch_compiler) is enabled...
-- SERIAL backend is being turned on to ensure there is at least one Host space. To change this, you must enable another host execution space and configure with -DKokkos_ENABLE_SERIAL=OFF or change CMakeCache.txt
-- Using -std=c++17 for C++17 standard as feature
-- Built-in Execution Spaces:
--     Device Parallel: Kokkos::Cuda
--     Host Parallel: NoTypeDefined
--       Host Serial: SERIAL
--
-- Architectures:
--  ZEN2
--  AMPERE86
-- Found CUDAToolkit: /usr/local/cuda-11.7/include (found version "11.7.64")
-- Looking for pthread.h
-- Looking for pthread.h - found
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD
-- Performing Test CMAKE_HAVE_LIBC_PTHREAD - Failed
-- Looking for pthread_create in pthreads
-- Looking for pthread_create in pthreads - not found
-- Looking for pthread_create in pthread
-- Looking for pthread_create in pthread - found
-- Found Threads: TRUE
-- Found TPLCUDA: TRUE
-- Found TPLLIBDL: /usr/include
-- Using internal desul_atomics copy
-- Kokkos Devices: CUDA;SERIAL, Kokkos Backends: CUDA;SERIAL
-- Configuring done
-- Generating done
-- Build files have been written to: /home/jacob/projects/experiments/scratch/build
```

Here is a minimal reproducible example I constructed:
```c++
#include <Kokkos_Core.hpp>

int main(int argc, char *argv[]) {
  {
  
  int scr_level = 1;
  int scr_size = 4896;

  const int nk = 7535;

  Kokkos::TeamPolicy<> policy(Kokkos::DefaultExecutionSpace(), nk, Kokkos::AUTO);
  policy.set_scratch_size(scr_level, Kokkos::PerTeam(scr_size));
  Kokkos::parallel_for("kernel", policy,
  KOKKOS_LAMBDA(Kokkos::TeamPolicy<>::member_type tmember) {
  });

  }
  Kokkos::finalize();

  return 0;
}
```

Here's the CMakeLists.txt file:
[CMakeLists.txt](https://github.com/kokkos/kokkos/files/12477257/CMakeLists.txt)

The CMake configuration expects a local copy of Kokkos rather than a globally installed one.

8. command line needed to reproduce the bug
```
make
./scratch
```
10. please also attach the `KokkosCore_config.h` header file (generated during the build);
[KokkosCore_config.zip](https://github.com/kokkos/kokkos/files/12477401/KokkosCore_config.zip)

**Any additional info**
Setting `scr_level = 0` fixes the issue, but unfortunately my intended application needs level 1 scratch. I have observed this problem on both my local machine and Perlmutter, which use Ampere-class GPUs. A separate machine using a Volta-class GPU seems to run just fine.

Here is an abbreviated device stack trace from `compute-sanitizer`:
```
========= COMPUTE-SANITIZER
========= Invalid __global__ atomic of size 4 bytes
=========     at 0x5c0 in /usr/local/cuda-11.7/include/device_atomic_functions.hpp:172:_INTERNAL_cf6669a5_8_main_cpp_main::atomicCAS(int *, int, int)
=========     by thread (0,0,0) in block (489,0,0)
=========     Address 0x608b0f600 is out of bounds
=========     and is 4609 bytes after the nearest allocation at 0x608ad2400 of size 245760 bytes
=========     Device Frame:/home/jacob/projects/experiments/scratch/kokkos/core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp:419:Kokkos::Impl::cuda_get_scratch_index(unsigned int, int *, unsigned long) [0x510]
=========     Device Frame:/home/jacob/projects/experiments/scratch/kokkos/core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp:497:Kokkos::Impl::ParallelFor<main::[lambda(Kokkos::Impl::CudaTeamMember) (instance 1)], Kokkos::TeamPolicy<>, Kokkos::Cuda>::operator ()() const [0x50]
=========     Device Frame:/home/jacob/projects/experiments/scratch/kokkos/core/src/Cuda/Kokkos_Cuda_KernelLaunch.hpp:88:void Kokkos::Impl::cuda_parallel_launch_local_memory<Kokkos::Impl::ParallelFor<main::[lambda(Kokkos::Impl::CudaTeamMember) (instance 1)], Kokkos::TeamPolicy<>, Kokkos::Cuda>>(T1) [0x10]
=========     Saved host backtrace up to driver entry point at kernel launch time
```

On further investigation, I wonder if the private variable `m_num_scratch_locks` is actually uninitialized for `ParallelFor` in the CUDA implementation. Doing something analogous to `ParallelReduce` and adding `m_num_scratch_locks = internal_space_instance->m_num_scratch_locks;` after line 559 in `Kokkos_Cuda_Parallel_Team.hpp` seems to fix the issue. With that change, my toy problem and the code I originally found the problem in both seem to work, and `compute-sanitizer` comes back clean now. That being said, I'm unfamiliar with the Kokkos codebase, so I don't know if this breaks other parts of the code.

masterleinad:
Fixes #6398. Since we are missing the initialization we are running into undefined behavior and it's difficult to reliably reproduce the issue in the test suite (although it always failed in a stand-alone reproducer). Disabling other tests around it or executing them in a different order made the test pass or fail for me (without the fix).

### DIFF
--- a/core/unit_test/TestTeamScratch.hpp
+++ b/core/unit_test/TestTeamScratch.hpp
@@ -54,5 +54,23 @@ TEST(TEST_CATEGORY, multi_level_scratch) {
 #endif
 }
 
+struct DummyTeamParallelForFunctor {
+  KOKKOS_FUNCTION void operator()(
+      Kokkos::TeamPolicy<TEST_EXECSPACE>::member_type) const {}
+};
+
+TEST(TEST_CATEGORY, team_scratch_memory_index_parallel_for) {
+  // Requesting per team scratch memory for a largish number of teams, resulted
+  // in problems computing the correct scratch pointer due to missed
+  // initialization of the maximum number of scratch pad indices in the Cuda
+  // baackend.
+  const int scratch_size = 4896;
+  const int league_size  = 7535;
+
+  Kokkos::TeamPolicy<TEST_EXECSPACE> policy(league_size, Kokkos::AUTO);
+  policy.set_scratch_size(1, Kokkos::PerTeam(scratch_size));
+  Kokkos::parallel_for("kernel", policy, DummyTeamParallelForFunctor());
+}
+
 }  // namespace Test
 #endif
--- a/core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
@@ -554,9 +554,10 @@ class ParallelFor<FunctorType, Kokkos::TeamPolicy<Properties...>,
     m_shmem_size =
         (m_policy.scratch_size(0, m_team_size) +
          FunctorTeamShmemSize<FunctorType>::value(m_functor, m_team_size));
-    m_scratch_size[0] = m_policy.scratch_size(0, m_team_size);
-    m_scratch_size[1] = m_policy.scratch_size(1, m_team_size);
-    m_scratch_locks   = internal_space_instance->m_scratch_locks;
+    m_scratch_size[0]   = m_policy.scratch_size(0, m_team_size);
+    m_scratch_size[1]   = m_policy.scratch_size(1, m_team_size);
+    m_scratch_locks     = internal_space_instance->m_scratch_locks;
+    m_num_scratch_locks = internal_space_instance->m_num_scratch_locks;
 
     // Functor's reduce memory, team scan memory, and team shared memory depend
     // upon team size.


core/unit_test/TestTeamScratch.hpp
```diff
@@ -54,5 +54,23 @@ TEST(TEST_CATEGORY, multi_level_scratch) {
...
+struct DummyTeamParallelForFunctor {
+  KOKKOS_FUNCTION void operator()(
+      Kokkos::TeamPolicy<TEST_EXECSPACE>::member_type) const {}
+};
+
+TEST(TEST_CATEGORY, team_scratch_memory_index_parallel_for) {
+  // Requesting per team scratch memory for a largish number of teams, resulted
+  // in problems computing the correct scratch pointer due to missed
+  // initialization of the maximum number of scratch pad indices in the Cuda
+  // baackend.
```
dalg24, line 66:
```suggestion
  // backend.
```

dalg24 (approved):

fnrizzi (approved):

core/unit_test/TestTeamScratch.hpp
masterleinad: I'll fix if there are any other changes required/requested.

dalg24:
@masterleinad did you identify when this bug was introduced?

dalg24:
Please open an issue listing fixes that would be going into an 4.0.01 if we do a patch release and add this one.

PR #6441, opened 2023-09-13
Serial: Allow for distinct execution space instances

masterleinad:
We discussed that we want to allow for distinct `Serial` execution space instances before merging https://github.com/kokkos/kokkos/pull/6151.
This pull request introduces a `NewInstance` tag that gives `Serial` instances that are different from the default constructed one.

Drive-by: Remove a test case that should never be executed because we assert before that `sum1==sum2`.

### DIFF
--- a/core/src/Serial/Kokkos_Serial.hpp
+++ b/core/src/Serial/Kokkos_Serial.hpp
@@ -72,6 +72,8 @@ class SerialInternal {
 };
 }  // namespace Impl
 
+struct NewInstance {};
+
 /// \class Serial
 /// \brief Kokkos device for non-parallel execution
 ///
@@ -107,6 +109,8 @@ class Serial {
   //@}
 
   Serial();
+
+  Serial(NewInstance);
 
   /// \brief True if and only if this method is being called in a
   ///   thread-parallel function.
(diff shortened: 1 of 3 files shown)


core/src/Serial/Kokkos_Serial.hpp
```diff
@@ -72,6 +72,8 @@ class SerialInternal {
 };
 }  // namespace Impl
 
+struct NewInstance {};
```
nliber, line 75: This should be

```c++
struct NewInstance { explicit NewInstance() = default; };
```

to disallow calling it via:

```c++
Serial({});
```

nliber (dismissed):

core/src/Serial/Kokkos_Serial.hpp
masterleinad: 297e7e3

dalg24 (approved):

ldh4 (approved):

Rombur:
clang format is not happy

masterleinad:
> clang format is not happy

Fixed.

masterleinad:
Only the `GCC-8.4.0` CI is timing out. Everything else passed.

nliber (approved):

PR #6443, opened 2023-09-13
Fix compiling code using Kokkos::printf for OpenMPTarget on Intel GPUs

masterleinad:
Compiling the unit tests with `OpenMPTargets` on `Intel GPUs` gives warnings like
```
non-string-literal argument used in printf is not supported for spir64
```
and compilation fails. As a workaround make `Kokkos::printf` a no-op in these cases if there are arguments.

### DIFF
--- a/core/src/impl/Kokkos_Printf.hpp
+++ b/core/src/impl/Kokkos_Printf.hpp
@@ -39,10 +39,13 @@ KOKKOS_FUNCTION void printf(const char* format, Args... args) {
   else
     sycl::ext::oneapi::experimental::printf(format, args...);
 #else
-  if constexpr (sizeof...(Args) == 0)
-    ::printf("%s", format);
+  if constexpr (sizeof...(Args) == 0) ::printf("%s", format);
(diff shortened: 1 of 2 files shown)


rgayatri23 (approved):

core/src/impl/Kokkos_Printf.hpp
```diff
@@ -39,11 +39,14 @@ KOKKOS_FUNCTION void printf(const char* format, Args... args) {
   else
     sycl::ext::oneapi::experimental::printf(format, args...);
 #else
-  if constexpr (sizeof...(Args) == 0)
-    ::printf("%s", format);
```
dalg24, line 43: Why was it not just `::printf(format)`?

masterleinad: See the comment above: 
> Some compilers warn if "args" is empty and format is not a string literal

ldh4 (approved):

crtrott (approved):

dalg24:
Did we report this to Intel?

dalg24:
The openmptarget build passed. Ignoring timeouts.

PR #6444, opened 2023-09-13
OpenMPTarget init-join fix

rgayatri23:
The PR adds verification for OpenMP process and initialization in init-join reducers.
This also fixes the backward_compatibility test failure with the backend.

### DIFF
--- a/core/src/OpenMPTarget/Kokkos_OpenMPTarget_Instance.cpp
+++ b/core/src/OpenMPTarget/Kokkos_OpenMPTarget_Instance.cpp
@@ -115,6 +115,9 @@ void OpenMPTargetInternal::impl_finalize() {
 void OpenMPTargetInternal::impl_initialize() {
   m_is_initialized = true;
 
+  Kokkos::Impl::OpenMPTargetExec::MAX_ACTIVE_THREADS =
+      Kokkos::Experimental::OpenMPTarget().concurrency();
+
   // FIXME_OPENMPTARGET:  Only fix the number of teams for NVIDIA architectures
   // from Pascal and upwards.
   // FIXME_OPENMPTARGTE: Cray compiler did not yet implement omp_set_num_teams.
--- a/core/src/OpenMPTarget/Kokkos_OpenMPTarget_Instance.cpp
+++ b/core/src/OpenMPTarget/Kokkos_OpenMPTarget_Instance.cpp
@@ -112,11 +112,11 @@ void OpenMPTargetInternal::impl_finalize() {
     Kokkos::kokkos_free<Kokkos::Experimental::OpenMPTargetSpace>(
         space.m_uniquetoken_ptr);
 }
+
 void OpenMPTargetInternal::impl_initialize() {
   m_is_initialized = true;
 
-  Kokkos::Impl::OpenMPTargetExec::MAX_ACTIVE_THREADS =
-      Kokkos::Experimental::OpenMPTarget().concurrency();
+  Kokkos::Impl::OpenMPTargetExec::MAX_ACTIVE_THREADS = concurrency();
 
   // FIXME_OPENMPTARGET:  Only fix the number of teams for NVIDIA architectures
   // from Pascal and upwards.
--- a/core/src/OpenMPTarget/Kokkos_OpenMPTarget_Parallel_Common.hpp
+++ b/core/src/OpenMPTarget/Kokkos_OpenMPTarget_Parallel_Common.hpp
@@ -197,6 +197,10 @@ struct ParallelReduceSpecialize<FunctorType, Kokkos::RangePolicy<PolicyArgs...>,
 
   static void execute_init_join(const FunctorType& f, const PolicyType& p,
                                 PointerType ptr, const bool ptr_on_device) {
+    OpenMPTargetExec::verify_is_process(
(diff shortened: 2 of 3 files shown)


dalg24 (commented):
How is calling two functions that check preconditions fixing anything?

dalg24:
If the issue is related to setting the value of `MAX_ACTIVE_THREADS` https://github.com/kokkos/kokkos/blob/6a95b5f3a7f9382d2251fec69dcbfe5cfefa8acc/core/src/OpenMPTarget/Kokkos_OpenMPTarget_Exec.cpp#L70-L71
why is initialization in `OpenMPTargetExec::verify_initialized` the proper place to do that?
Why can't it be done at initialization of the backend?

rgayatri23:
> If the issue is related to setting the value of `MAX_ACTIVE_THREADS`
> 
> https://github.com/kokkos/kokkos/blob/6a95b5f3a7f9382d2251fec69dcbfe5cfefa8acc/core/src/OpenMPTarget/Kokkos_OpenMPTarget_Exec.cpp#L70-L71
> 
> 
> why is initialization in `OpenMPTargetExec::verify_initialized` the proper place to do that?
> Why can't it be done at initialization of the backend?

Yes the issue is related to `MAX_ACTIVE_THREADS`. That variable is used to calculate maximum in flight teams which in-turn determines the scratch space allocated. 
The variable can be set during the backend initialization, which is probably where it should go I guess if we only want to do it only once.

masterleinad:
> The variable can be set during the backend initialization, which is probably where it should go I guess if we only want to do it only once.

Would there be a reason to do it more than once?

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Parallel_Common.hpp
```diff
@@ -197,6 +197,10 @@ struct ParallelReduceSpecialize<FunctorType, Kokkos::RangePolicy<PolicyArgs...>,
 
   static void execute_init_join(const FunctorType& f, const PolicyType& p,
                                 PointerType ptr, const bool ptr_on_device) {
+    OpenMPTargetExec::verify_is_process(
+        "Kokkos::Experimental::OpenMPTarget parallel_for");
+    OpenMPTargetExec::verify_initialized(
+        "Kokkos::Experimental::OpenMPTarget parallel_for");
```
masterleinad, line 203: Are these changes still necessary?

rgayatri23: Firstly, the message label was wrong which I fixed now.
Secondly, I think yes , we are checking for those in other similar routines too. Not having here was an oversight.

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Parallel_Common.hpp
```diff
@@ -84,9 +84,9 @@ struct ParallelReduceSpecialize<FunctorType, Kokkos::RangePolicy<PolicyArgs...>,
   static void execute_reducer(const FunctorType& f, const PolicyType& p,
                               PointerType result_ptr, bool ptr_on_device) {
     OpenMPTargetExec::verify_is_process(
-        "Kokkos::Experimental::OpenMPTarget parallel_for");
+        "Kokkos::Experimental::OpenMPTarget parallel_reduce");
     OpenMPTargetExec::verify_initialized(
-        "Kokkos::Experimental::OpenMPTarget parallel_for");
+        "Kokkos::Experimental::OpenMPTarget parallel_reduce");
```
masterleinad, line 89: Wouldn't it be better to call these functions in `execute` directly before the dispatch to `execute_init_join`, `execute_reducer`,...? In that case, you could also be more precise about the policy used in the error message.

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Instance.cpp
```diff
@@ -115,6 +115,9 @@ void OpenMPTargetInternal::impl_finalize() {
 void OpenMPTargetInternal::impl_initialize() {
   m_is_initialized = true;
 
+  Kokkos::Impl::OpenMPTargetExec::MAX_ACTIVE_THREADS =
+      Kokkos::Experimental::OpenMPTarget().concurrency();
```
masterleinad, line 119:
```suggestion
  Kokkos::Impl::OpenMPTargetExec::MAX_ACTIVE_THREADS =
      concurrency();
```

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Parallel_Common.hpp
rgayatri23: Edited the error message.

masterleinad (approved):
OK with me.

dalg24 (approved):

dalg24:
The openmptarget build passed.  Not waiting on the rest

dalg24:
@rgayatri23 make sure you add to the changelog
When was this bug introduced? Did you investigate?

rgayatri23:
> @rgayatri23 make sure you add to the changelog When was this bug introduced? Did you investigate?

Yeah this was introduced in when we did #6043

Issue #6447, opened 2023-09-14
Inconsistent behavior of `Kokkos::OpenMP` vs `Kokkos::Cuda`/`Kokkos::HIP` when default initialized before `Kokkos` is initialized


romintomasetti:
# Problem description

When an `OpenMP` space instance `Kokkos::OpenMP` is initialized **before** `Kokkos` is initialized, it fails silently, simply putting some warning in `std::cerr`.

However, for `CUDA` and `HIP`, it will stop with `Kokkos::abort`.

I deem this behavior inconsistent and suggest to make it consistent for all spaces (i.e. make it `abort`).

See below. I guess a macro is welcome, since all spaces shoudl have the exact same implementation of `verify_is_initialized`.

https://github.com/kokkos/kokkos/blob/6a95b5f3a7f9382d2251fec69dcbfe5cfefa8acc/core/src/OpenMP/Kokkos_OpenMP_Instance.cpp#L381-L385

https://github.com/kokkos/kokkos/blob/6a95b5f3a7f9382d2251fec69dcbfe5cfefa8acc/core/src/HIP/Kokkos_HIP_Instance.cpp#L131-L136

https://github.com/kokkos/kokkos/blob/6a95b5f3a7f9382d2251fec69dcbfe5cfefa8acc/core/src/Cuda/Kokkos_Cuda_Instance.cpp#L353-L358

# Further discussion and context

I was implementing a benchmark using `Google Benchmark` (`v1.8.3`), and I naïvely used `BENCHMARK_REGISTER_F` to register my benchmark as one usually does.

However, `BENCHMARK_REGISTER_F` will in fact instantiate the benchmark class **before** the `main` (wherein `Kokkos` is initialized), because it creates a static variable at the global scope (see https://github.com/google/benchmark/blob/344117638c8ff7e239044fd0fa7085839fc03021/include/benchmark/benchmark.h#L1614-L1619).

The problem was that my benchmark class have some `execution_space space;` protected member, that therefore got default initialized as soon as the static variable was initialized, i.e. before the `main`. Therefore, the singleton of the internal backend implementation would be retrieved in the default constructor of `space`, but of course the singleton would not be valid since `Kokkos` is not yet initialized.

For `OpenMP`, it was okayish because it would just output to `std::cerr`, but my `space` member variable was then initialized to possibly something wrong (since `verify_is_initialized` would say "nope, not initialized so I'll write to cerr").

[...]

PR #6448, opened 2023-09-14
core(spaces): consistency of verify_is_initialized

romintomasetti:
This PR addresses the main issue discussed in:
- #6447

Basically, `Kokkos::Impl::OpenMPInternal` would not `Kokkos::abort` as others do.

Also, some `...::verify_is_initialized()` would return `bool`, others `int`. I made them all return `bool`.

### DIFF
--- a/core/src/Cuda/Kokkos_Cuda_Instance.cpp
+++ b/core/src/Cuda/Kokkos_Cuda_Instance.cpp
@@ -350,19 +350,21 @@ CudaInternal::~CudaInternal() {
   }
 }
 
-int CudaInternal::verify_is_initialized(const char *const label) const {
+bool CudaInternal::verify_is_initialized(const char *const label) const {
   if (m_cudaDev < 0) {
-    Kokkos::abort((std::string("Kokkos::Cuda::") + label +
-                   " : ERROR device not initialized\n")
+    Kokkos::abort((std::string("Kokkos::Cuda : ") + label +
+                   " : ERROR device not initialized")
                       .c_str());
   }
   return 0 <= m_cudaDev;
 }
+
 uint32_t CudaInternal::impl_get_instance_id() const { return m_instance_id; }
 CudaInternal &CudaInternal::singleton() {
   static CudaInternal self;
   return self;
 }
+
 void CudaInternal::fence(const std::string &name) const {
   Impl::cuda_stream_synchronize(get_stream(), this, name);
 }
--- a/core/src/SYCL/Kokkos_SYCL_Instance.cpp
+++ b/core/src/SYCL/Kokkos_SYCL_Instance.cpp
@@ -62,14 +62,15 @@ SYCLInternal::~SYCLInternal() {
   }
 }
 
-int SYCLInternal::verify_is_initialized(const char* const label) const {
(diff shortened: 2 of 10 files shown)


core/src/Cuda/Kokkos_Cuda_Instance.cpp
```diff
@@ -35,6 +35,7 @@
 #include <impl/Kokkos_CheckedIntegerOps.hpp>
 #include <impl/Kokkos_DeviceManagement.hpp>
 #include <impl/Kokkos_ExecSpaceManager.hpp>
+#include <impl/Kokkos_Space_Instance.hpp>
```
romintomasetti, line 38: @dalg24 I don't know where to put this new macro, so anywhere you like is fine with me :smile:

core/src/SYCL/Kokkos_SYCL_Instance.cpp
```diff
@@ -62,14 +63,8 @@ SYCLInternal::~SYCLInternal() {
...
 
-int SYCLInternal::verify_is_initialized(const char* const label) const {
-  if (!is_initialized()) {
-    Kokkos::abort((std::string("Kokkos::Experimental::SYCL::") + label +
-                   " : ERROR device not initialized\n")
-                      .c_str());
-  }
-  return is_initialized();
-}
+KOKKOS_IMPL_SPACE_VERIFY_IS_INITIALIZED(SYCL, !is_initialized(), is_initialized())
```
romintomasetti, line 66: Todo for me: it seems this one needs to add `::Experimental::` in the error message.

core/src/Cuda/Kokkos_Cuda_Instance.cpp
dalg24: I am not convinced having a macro is a good idea here nor that we should dictate that individual backends have/use `verify_is_initialized`.

romintomasetti: * Instead of a macro, would you prefer a CRTP approach? Kokkos could have several CRTP classes that internal space classes could inherit from and so enforce some sorts of behaviors. But that approach seems quite difficult to implement...
* I guess the current PR does not enforce all internal space classes to have `verify_is_initialized`, it just proposes that those currently having it use the same implementation :smile:

Side question: `verify_is_initialized` returns a `bool`, but in fact the return is `true` or it aborts, so the return value is kind of useless...

dalg24: I would add a death test that potentially checks the error message with a regex.

Then I would delegate to backends developers and tell them to do whatever they want but that the test needs to pass.  I am not convinced that there is a need for a shared facility to achieve the desired behavior.  If I was maintaining a backend I would have gotten rid of that `verify_is_initialized` function a long time ago.  I don't like anything about it.

dalg24:
I agree that it would be nice to check the QOI of backends with respect of error messages when constructing/deleting Kokkos objects and invoking functions before Kokkos::initialize or after Kokkos::finalize.
That said I don't think that `Impl::verify_is_initialized()` is the way to go.

Issue #6439, opened 2023-09-13
Can't compile SIMD tests with ROCm 5.6.0 and AVX2


masterleinad:
Looks like the problem with `ROCm` 5.6.0 not being able to compile the SIMD tests with AVX2 persists. I disabled AVX2 support in that case now as a workaround. Unfortunately, I don't have access to that compiler version to easily debug this further.

_Originally posted by @masterleinad in https://github.com/kokkos/kokkos/issues/6188#issuecomment-1716363773_

masterleinad:
Fixed in #6449.

PR #6449, opened 2023-09-15
Workaround for ROCm 5.6+ failing to compile with AVX2 SIMD support

masterleinad:
Workaround for https://github.com/kokkos/kokkos/issues/6439. Tested in https://github.com/kokkos/kokkos/pull/6188.

It turns out that `hipcc` for `ROCm` 5.6 struggles to compile `_mm_maskload_epi32` and `_mm256_maskload_epi64`. This pull request proposes to use `_mm_loadu_si128` with `_mm_and_si128` and `_mm256_loadu_si256` with `_mm256_and_si256` instead as a workaround.
Also, see https://stackoverflow.com/a/59650871 and https://stackoverflow.com/questions/67997645/most-efficient-way-of-implementing-mm256-mask-add-ps-on-avx2 for why this should be OK. Note that we don't want to use a mask for most replaced cases anyway (but we need it since the masked intrinsics have only been introduced with AVX512).

### DIFF
--- a/simd/src/Kokkos_SIMD_AVX2.hpp
+++ b/simd/src/Kokkos_SIMD_AVX2.hpp
@@ -938,7 +938,14 @@ class simd<std::int32_t, simd_abi::avx2_fixed_size<4>> {
   }
   KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION void copy_from(value_type const* ptr,
                                                        element_aligned_tag) {
+    // FIXME_HIP ROCm 5.6 can't compile with the intrinsic used here.
+#if defined(__HIPCC__) &&       \
+    ((HIP_VERSION_MAJOR > 5) || \
+     ((HIP_VERSION_MAJOR == 5) && (HIP_VERSION_MINOR >= 6)))
+    m_value = _mm_loadu_si128(reinterpret_cast<__m128i const*>(ptr));
+#else
     m_value = _mm_maskload_epi32(ptr, static_cast<__m128i>(mask_type(true)));
+#endif
   }
   KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION void copy_to(
       value_type* ptr, element_aligned_tag) const {
@@ -1079,8 +1086,15 @@ class simd<std::int64_t, simd_abi::avx2_fixed_size<4>> {
   }
   KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION void copy_from(value_type const* ptr,
                                                        element_aligned_tag) {
+    // FIXME_HIP ROCm 5.6 can't compile with the intrinsic used here.
+#if defined(__HIPCC__) &&       \
+    ((HIP_VERSION_MAJOR > 5) || \
+     ((HIP_VERSION_MAJOR == 5) && (HIP_VERSION_MINOR >= 6)))
+    m_value = _mm256_loadu_si256(reinterpret_cast<__m256i const*>(ptr));
+#else
     m_value = _mm256_maskload_epi64(reinterpret_cast<long long const*>(ptr),
                                     static_cast<__m256i>(mask_type(true)));
+#endif
   }
   KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION void copy_to(
       value_type* ptr, element_aligned_tag) const {
@@ -1232,8 +1246,15 @@ class simd<std::uint64_t, simd_abi::avx2_fixed_size<4>> {
   }
   KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION void copy_from(value_type const* ptr,
                                                        element_aligned_tag) {
+    // FIXME_HIP ROCm 5.6 can't compile with the intrinsic used here.
+#if defined(__HIPCC__) &&       \
+    ((HIP_VERSION_MAJOR > 5) || \
+     ((HIP_VERSION_MAJOR == 5) && (HIP_VERSION_MINOR >= 6)))
+    m_value = _mm256_loadu_si256(reinterpret_cast<__m256i const*>(ptr));
+#else
     m_value = _mm256_maskload_epi64(reinterpret_cast<long long const*>(ptr),
                                     static_cast<__m256i>(mask_type(true)));
+#endif
   }
   KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION constexpr explicit operator __m256i()
       const {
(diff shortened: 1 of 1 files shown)


simd/src/Kokkos_SIMD_AVX2.hpp
```diff
@@ -938,7 +938,14 @@ class simd<std::int32_t, simd_abi::avx2_fixed_size<4>> {
   }
   KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION void copy_from(value_type const* ptr,
                                                        element_aligned_tag) {
+    // FIXME_HIP ROCm 5.6 can't compile with the intrinsic used here.
+#if defined(__HIPCC__) &&       \
+    ((HIP_VERSION_MAJOR > 5) || \
+     ((HIP_VERSION_MAJOR == 5) && (HIP_VERSION_MINOR >= 6)))
```
ldh4, line 944:
```suggestion
     ((HIP_VERSION_MAJOR >= 5) && (HIP_VERSION_MINOR >= 6)))
```
Don't we want to only guard against rocm 5.6+ and not necessarily for everything after rocm 5.0?

masterleinad: What you are proposing wouldn't use the workaround for, say, 6.0, right? I think the current version is correct for guarding >=5.6. For 5.0 we wouldn't use the workaround since both `HIP_VERSION_MAJOR > 5` and `HIP_VERSION_MINOR >=6` are false.

ldh4 (approved):

simd/src/Kokkos_SIMD_AVX2.hpp
dalg24, line 944: This is the kind of guards we don't want to read multiple times.  Consider doing this
```suggestion
#if defined(__HIPCC__) &&       \
    ((HIP_VERSION_MAJOR > 5) || \
     ((HIP_VERSION_MAJOR == 5) && (HIP_VERSION_MINOR >= 6)))
#define KOKKOS_IMPL_WORKAROUND_ROCM_AVX2_ISSUE
#endif

#ifdef KOKKOS_IMPL_WORKAROUND_ROCM_AVX2_ISSUE
```
and `#undef` wherever appropriate.

masterleinad: Sure. 65260e0ef37b496b7a95da49af62fd350e309127

cz4rs (approved):

simd/src/Kokkos_SIMD_AVX2.hpp
```diff
@@ -25,6 +25,13 @@
 
 #include <immintrin.h>
 
+// FIXME_HIP ROCm 5.6 can't compile with the intrinsic used here.
+#if defined(__HIPCC__) &&       \
+    ((HIP_VERSION_MAJOR > 5) || \
+     ((HIP_VERSION_MAJOR == 5) && (HIP_VERSION_MINOR >= 6)))
```
Rombur, line 31: Why don't you limit the workaround to ROCm 5.6? There is the same problem with 5.7?

masterleinad: I am being conservative. We can restrict the workaround if we know that 5.7 works.

Rombur: If you haven't check 5.7, then we should try it before merging this.

masterleinad: Please do it yourself then. I don't have access to 5.7 on a machine with AMD GPUs.

dalg24:
There is a docker image for it. It was uploaded 2 days ago

simd/src/Kokkos_SIMD_AVX2.hpp
masterleinad: Just trying with `Serial`, I am still seeing the error with 5.7:
```
cd /app/kokkos/build_hip/simd/unit_tests && /opt/rocm/bin/hipcc  -DKOKKOS_DEPENDENCE -I/app/kokkos/simd/unit_tests/include -I/app/kokkos/tpls/gtest -I/app/kokkos/build_hip -I/app/kokkos/build_hip/core/src -I/app/kokkos/core/src -I/app/kokkos/core/src/../../tpls/desul/include -I/app/kokkos/build_hip/containers/src -I/app/kokkos/containers/src -I/app/kokkos/build_hip/algorithms/src -I/app/kokkos/algorithms/src -I/app/kokkos/build_hip/simd/src -I/app/kokkos/simd/src  -O2 -g -DNDEBUG   -pthread -march=znver2 -mtune=znver2 -std=gnu++17 -o CMakeFiles/Kokkos_UnitTest_SIMD.dir/TestSIMD.cpp.o -c /app/kokkos/simd/unit_tests/TestSIMD.cpp
clang++: error: clang frontend command failed with exit code 70 (use -v to see invocation)
AMD clang version 17.0.0 (https://github.com/RadeonOpenCompute/llvm-project roc-5.7.0 23352 d1e13c532a947d0cbfc94759c00dcf152294aa13)
Thread model: posix
InstalledDir: /opt/rocm-5.7.0/llvm/bin
clang++: note: diagnostic msg: Error generating preprocessed source(s).
```

masterleinad:
The HIP CI builds are passing (here and in https://github.com/kokkos/kokkos/pull/6188).

simd/src/Kokkos_SIMD_AVX2.hpp
```diff
@@ -1728,4 +1763,10 @@ class where_expression<simd_mask<std::uint64_t, simd_abi::avx2_fixed_size<4>>,
 }  // namespace Experimental
 }  // namespace Kokkos
 
+#if defined(__HIPCC__) &&       \
+    ((HIP_VERSION_MAJOR > 5) || \
+     ((HIP_VERSION_MAJOR == 5) && (HIP_VERSION_MINOR >= 6)))
```
crtrott, line 1768:
```suggestion
#if defined(KOKKOS_IMPL_WORKAROUND_ROCM_AVX2_ISSUE)
```

crtrott (commented):
I think it's basically ok. I have a suggestion at the the end, though so we don't have mismatching conditions for defined and undef potentially. My proposed thing would work right?

simd/src/Kokkos_SIMD_AVX2.hpp
masterleinad: I just undefined it unconditionally now.

ldh4:
We could also consider replacing current implementation with proposed workarounds and avoid introducing these `ifdefs`. As stated in the description, there are cases where masks aren't really needed, and it's also notable that `_mm256_loadu_si256` is noted to have lower latency than `_mm256_maskload_epi64` according to the intel intrinsics guide. I think it could be a suitable replacement.

masterleinad:
> We could also consider replacing current implementation with proposed workarounds and avoid introducing these `ifdefs`. As stated in the description, there are cases where masks aren't really needed, and it's also notable that `_mm256_loadu_si256` is noted to have lower latency than `_mm256_maskload_epi64` according to the intel intrinsics guide. I think it could be a suitable replacement.

Yes, for the cases where we don't use the mask. I'm not quite sure about the cases where the mask is actually used.

simd/src/Kokkos_SIMD_AVX2.hpp
Rombur: Let's change the condition to only check for ROCm 5.6 and 5.7. I've talked with AMD and it's easier for the compiler team to fix the issue if they can just pull Kokkos without having to mess with our code to reproduce the error.

Rombur (approved):

PR #6452, opened 2023-09-18
OpenMPTarget: Disable check for SIMD compiler macros

masterleinad:
https://godbolt.org/z/3qb1xcs9x demonstrates that the `-fopenmp-targets=nvptx64` flag disables SIMD macros (in the device compile pass) and this also holds for other targets.
Since we can still compile and run the unit tests, this pull request proposes to disable the checks for matching SIMD compiler macros (introduced in https://github.com/kokkos/kokkos/pull/6248) if we enable support for the `OpenMPTarget` backend.

### DIFF
--- a/simd/src/Kokkos_SIMD.hpp
+++ b/simd/src/Kokkos_SIMD.hpp
@@ -30,6 +30,20 @@
 
 #include <Kokkos_Macros.hpp>
 
+// FIXME_OPENMPTARGET The device pass disables all compiler macros checked
+#ifdef KOKKOS_ENABLE_OPENMPTARGET
+#if defined(KOKKOS_ARCH_AVX2)
+#include <Kokkos_SIMD_AVX2.hpp>
+#endif
+
+#if defined(KOKKOS_ARCH_AVX512XEON)
+#include <Kokkos_SIMD_AVX512.hpp>
+#endif
+
+#ifdef __ARM_NEON
+#include <Kokkos_SIMD_NEON.hpp>
+#endif
+#else  // KOKKOS_ENABLE_OPENMPTARGET
 #if defined(KOKKOS_ARCH_AVX) && !defined(__AVX__)
 #error "__AVX__ must be defined for KOKKOS_ARCH_AVX"
 #endif
@@ -50,6 +64,7 @@
 
 #ifdef __ARM_NEON
 #include <Kokkos_SIMD_NEON.hpp>
+#endif
 #endif
 
 #if defined(KOKKOS_COMPILER_NVCC) && KOKKOS_COMPILER_NVCC < 1130


rgayatri23 (approved):

simd/src/Kokkos_SIMD.hpp
```diff
@@ -30,6 +30,20 @@
...
+#ifdef KOKKOS_ENABLE_OPENMPTARGET
+#if defined(KOKKOS_ARCH_AVX2)
+#include <Kokkos_SIMD_AVX2.hpp>
+#endif
+
+#if defined(KOKKOS_ARCH_AVX512XEON)
+#include <Kokkos_SIMD_AVX512.hpp>
+#endif
+
+#ifdef __ARM_NEON
```
crtrott, line 43: how does this help?

crtrott (changes requested):
I assume we just don't have a KOKKOS_ARCH_NEON thing?

simd/src/Kokkos_SIMD.hpp
```diff
@@ -30,6 +30,20 @@
...
+#include <Kokkos_SIMD_NEON.hpp>
+#endif
+#else  // KOKKOS_ENABLE_OPENMPTARGET
 #if defined(KOKKOS_ARCH_AVX) && !defined(__AVX__)
```
dalg24, line 47:
```suggestion
#if defined(KOKKOS_ARCH_AVX) && !defined(__AVX__) && !defined(KOKKOS_ENABLE_OPENMPTARGET)
```

dalg24:
Is that an issue when enabling non-nvidia archs?

masterleinad:
> I assume we just don't have a KOKKOS_ARCH_NEON thing?

Coming in https://github.com/kokkos/kokkos/pull/6394.

masterleinad:
> Is that an issue when enabling non-nvidia archs?

According to @rgayatri23 we have the same issue whenever we specify `-fopenmp-targets=*`.

simd/src/Kokkos_SIMD.hpp
masterleinad: It just mirrors what we do below as well. We could pull it out but I would expect us to guard this in a similar way in the near future as well.

simd/src/Kokkos_SIMD.hpp
masterleinad: I didn't want to repeat the FIXME everywhere.

crtrott (approved):

rgayatri23:
> > Is that an issue when enabling non-nvidia archs?
> 
> According to @rgayatri23 we have the same issue whenever we specify `-fopenmp-targets=*`.

Yes I encountered the same issue even on AMD MI250x `-fopenmp-targets=gfx90a`

Issue #6442, opened 2023-09-13
Nightly failure with intel/19.0.5: TestStdAlgorithmsTeamFindIf error: class template name must be a placeholder for the complete type being initialized


ndellingwood:
**Describe the bug**
Nightly intel/19.0.5 builds fail to compile with output:

```
12:58:20 /home/jenkins/node/workspace/kokkos_osrhel7_intel_1905_openmp_pthread_serial/kokkos/algorithms/unit_tests/TestStdAlgorithmsTeamFindIf.cpp(165): error: class template name must be a placeholder for the complete type being initialized (not for a component of that type)
12:58:20       const GreaterEqualFunctor unaryPred{val};
12:58:20             ^
12:58:20           detected during instantiation of "void Test::stdalgos::TeamFindIf::run_all_scenarios<LayoutTag,ValueType>(bool) [with LayoutTag=Test::stdalgos::DynamicTag, ValueType=double]" at line 197
```

The job was down for several days and just came back up today, the code above merged in with PR #6207 

**Please also include the following items to support reproducing the bug**
**Reproducer**:
```bash
module load sems-cmake sems-intel/19.0.5

$KOKKOS_PATH/generate_makefile.bash --with-devices=Serial  --compiler=icpc --cxxflags="-O3 -Wall -Wunused-parameter -Wshadow -pedantic -Werror -Wsign-compare -Wtype-limits -Wuninitialized " --cxxstandard="17" --no-examples   --with-options=enable_large_mem_tests
```

ndellingwood:
Also occurs in `TestStdAlgorithmsTeamFindIfNot`

```
14:37:51 /home/jenkins/node/workspace/kokkos_osrhel7_intel_1905-dbg/kokkos/algorithms/unit_tests/TestStdAlgorithmsTeamFindIfNot.cpp(160): error: class template name must be a placeholder for the complete type being initialized (not for a component of that type)
14:37:51       const GreaterEqualFunctor unaryPred{val};
14:37:51             ^
14:37:51           detected during instantiation of "void Test::stdalgos::TeamFindIfNot::run_all_scenarios<LayoutTag,ValueType>(bool) [with LayoutTag=Test::stdalgos::DynamicTag, ValueType=double]" at line 192
```

fnrizzi:
closing this since the PR was merged, will reopen if something else comes up

ndellingwood:
Nightlies passed, thanks @fnrizzi !

PR #6455, opened 2023-09-19
fix intel compile error for CTAD

fnrizzi:
fixes https://github.com/kokkos/kokkos/issues/6442

Was also able to reproduce on godbolt: https://godbolt.org/z/MaaaeWxYT
Looks like this version of the intel compiler did not like the const qualifier. 
Also verified it works for higher versions of intel https://godbolt.org/z/GEWocj71q


masterleinad (approved):

nliber (approved):

crtrott (approved):

crtrott:
Unrelated HIP build system issues.

PR #6276, opened 2023-07-12
SIMD: added unit tests for reductions ops and a few unary ops

ldh4:
This PR adds unit tests for:
- `T hmin(const_where_expression<simd_mask<T, Abi>, simd<T, Abi>> const& x)`
- `T hmax(const_where_expression<simd_mask<T, Abi>, simd<T, Abi>> const& x)`
- `T reduce(const_where_expression<simd_mask<T, Abi>, simd<T, Abi>> const& x, T, std::plus<>)`

and also for following unary ops from `Intel svml intrinsics` when built with `icc`/`icpc`
- `simd Kokkos::cbrt(const simd& lhs)`
- `simd Kokkos::exp(const simd& lhs)`
- `simd Kokkos::log(const simd& lhs)`

PR #6465, opened 2023-09-22
SIMD: Math functions should be in namespace Kokkos

masterleinad:
This fixes the confusion in https://github.com/kokkos/kokkos/pull/6276#discussion_r1316444804.
At the moment, we have fallback SIMD math functions (those that are also defined in `cmath`) defined in namespace `Kokkos` but the specialized ones in namespace `Kokkos::Experimental`.
This pull request makes sure to have all these functions in namespace `Kokkos` so that users don't have to use different namespaces based on the SIMD type having specializations or not. If `DEPRECATED_CODE_4` is enabled, all of these functions are also defined  in namespace `Kokkos::Experimental` similar to what we are doing for the non-SIMD `Kokkos` math functions.

### DIFF
--- a/simd/src/Kokkos_SIMD_Common.hpp
+++ b/simd/src/Kokkos_SIMD_Common.hpp
@@ -378,6 +378,18 @@ template <class T, class Abi>
   return result;
 }
 
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+namespace Experimental {
+template <class T, class Abi>
+[[nodiscard]] KOKKOS_DEPRECATED KOKKOS_FORCEINLINE_FUNCTION
+    Experimental::simd<T, Abi>
+    min(Experimental::simd<T, Abi> const& a,
+        Experimental::simd<T, Abi> const& b) {
+  return Kokkos::min(a, b);
+}
+}  // namespace Experimental
+#endif
+
 template <class T, class Abi>
 [[nodiscard]] KOKKOS_FORCEINLINE_FUNCTION Experimental::simd<T, Abi> max(
     Experimental::simd<T, Abi> const& a, Experimental::simd<T, Abi> const& b) {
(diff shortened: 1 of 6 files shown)


simd/src/Kokkos_SIMD_Common.hpp
```diff
@@ -388,23 +400,54 @@ template <class T, class Abi>
...
+    for (std::size_t i = 0; i < Experimental::simd<T, Abi>::size(); ++i) {   \
+      result[i] = Kokkos::FUNC(a[i]);                                        \
+    }                                                                        \
+    return result;                                                           \
+  }                                                                          \
+  namespace Experimental {                                                   \
+  template <class T, class Abi>                                              \
+  [[nodiscard]] KOKKOS_DEPRECATED KOKKOS_FORCEINLINE_FUNCTION simd<T, Abi>   \
+  FUNC(simd<T, Abi> const& a) {                                              \
+    Kokkos::FUNC(a);                                                         \
```
ldh4, line 436: Looks like there are couple missing return statements like here.

ldh4 (commented):
Conceptually, I agree with this change, but it's unfortunate that files now have to be interleaving with `Experimental` namespaces multiple times.
Not a suggestion, but it makes me partly wonder if it may be better to aggregate all functions that are no longer in `Experimental` and place them altogether toward bottom to simplify namespace structure.

simd/src/Kokkos_SIMD_Common.hpp
masterleinad: Fixed.

masterleinad:
> Not a suggestion, but it makes me partly wonder if it may be better to aggregate all functions that are no longer in `Experimental` and place them altogether toward bottom to simplify namespace structure.

Yes, I refrained from doing anymore restructuring to avoid conflicts with other pull requests.

masterleinad:
Retest this please.

ldh4 (approved):

nliber (approved):

masterleinad:
The `gcc-8.4.0` CI is timing out and we see the usual `OpenMPTarget` test failure.

PR #6474, opened 2023-10-01
Add compatible copy assignment operator to DualView

maartenarnst:
This Kokkos PR is a companion to the merged Trilinos PR 
- https://github.com/trilinos/Trilinos/pull/12324

In that Trilinos PR, we had added a compatible copy constructor and assignment operator to `Tpetra` `WrappedDualView`. The motivation was to support reshaping copy construction and assignment of `MultiVector` filled with `Stokhos` UQ types.

As it stands, the `Kokkos` `DualView` has a compatible copy constructor, but it doesn't seem to have a compatible copy assignment operator yet. Hence, this PR adds it. 

Note that this PR doesn't add a test. Note also that the existing copy constructor doesn't seem to be tested. I tried to add a test for both. But then I found that the `DualView` seems to be missing an `operator==`, so it's not straightforward to test equality. I'd be happy to add a test, but I'm not sure how to best do it. If you should have a suggestion, I'll be happy to follow it.

Thanks!

@romintomasetti @etphipp @brian-kelley

### DIFF
--- a/containers/src/Kokkos_DualView.hpp
+++ b/containers/src/Kokkos_DualView.hpp
@@ -292,6 +292,15 @@ class DualView : public ViewTraits<DataType, Properties...> {
         d_view(src.d_view),
         h_view(src.h_view) {}
 
+  //! Copy assignment operator (shallow copy assignment)
+  template <typename DT, typename... DP>
+  DualView& operator=(const DualView<DT, DP...>& src) {
+    modified_flags = src.modified_flags;
+    d_view         = src.d_view;
+    h_view         = src.h_view;
+    return *this;
+  }
+
   //! Subview constructor
   template <class DT, class... DP, class Arg0, class... Args>
   DualView(const DualView<DT, DP...>& src, const Arg0& arg0, Args... args)
--- a/containers/unit_tests/TestDualView.hpp
+++ b/containers/unit_tests/TestDualView.hpp
@@ -65,14 +65,14 @@ struct test_dualview_copy_construction_and_assignment {
   using scalar_type     = Scalar;
   using execution_space = Device;
 
-  void run_me() {
+  void operator()() {
     constexpr unsigned int n = 10;
     constexpr unsigned int m = 5;
 
     using SrcViewType = Kokkos::DualView<Scalar**, Kokkos::LayoutLeft, Device>;
     using DstViewType =
         Kokkos::DualView<const Scalar * [m], Kokkos::LayoutLeft, Device>;
-    
+
     SrcViewType a("A", n, m);
 
     // Copy construction
@@ -99,8 +99,6 @@ struct test_dualview_copy_construction_and_assignment {
       a.clear_sync_state();
     }
   }
-
-  test_dualview_copy_construction_and_assignment() { run_me(); }
 };
 
 template <typename Scalar, class Device>
@@ -424,7 +422,7 @@ void test_dualview_alloc(unsigned int size) {
 
 template <typename Scalar, typename Device>
 void test_dualview_copy_construction_and_assignment() {
-  Impl::test_dualview_copy_construction_and_assignment<Scalar, Device>();
+  Impl::test_dualview_copy_construction_and_assignment<Scalar, Device>()();
 }
 
 template <typename Scalar, typename Device>
--- a/containers/unit_tests/TestDualView.hpp
+++ b/containers/unit_tests/TestDualView.hpp
@@ -65,18 +65,21 @@ struct test_dualview_copy_construction_and_assignment {
   using scalar_type     = Scalar;
   using execution_space = Device;
 
-  template <typename ViewType>
   void run_me() {
-    const unsigned int n = 10;
-    const unsigned int m = 5;
+    constexpr unsigned int n = 10;
+    constexpr unsigned int m = 5;
 
-    ViewType a("A", n, m);
+    using SrcViewType = Kokkos::DualView<Scalar**, Kokkos::LayoutLeft, Device>;
+    using DstViewType =
+        Kokkos::DualView<const Scalar * [m], Kokkos::LayoutLeft, Device>;
+    
+    SrcViewType a("A", n, m);
 
     // Copy construction
-    ViewType b(a);
+    DstViewType b(a);
 
     // Copy assignment
-    ViewType c = a;
+    DstViewType c = a;
 
     // Check equality (shallow) of the host and device views
     ASSERT_EQ(a.view_host(), b.view_host());
@@ -87,7 +90,7 @@ struct test_dualview_copy_construction_and_assignment {
 
     // We can't test shallow equality of modified_flags because it's protected.
     // So we test it indirectly through sync state behavior.
-    if (!std::decay_t<ViewType>::impl_dualview_is_single_device::value) {
+    if (!std::decay_t<SrcViewType>::impl_dualview_is_single_device::value) {
       a.clear_sync_state();
       a.modify_host();
       ASSERT_TRUE(a.need_sync_device());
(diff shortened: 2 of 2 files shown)


fnrizzi:
ok to test

fnrizzi:
> it doesn't seem to have a compatible copy assignment operator yet. 

Did you check in the issues if there was anything mentioning this? 


> Note that this PR doesn't add a test. Note also that the existing copy constructor doesn't seem to be tested. 
> I tried to add a test for both. But then I found that the DualView seems to be missing an operator==, 
> so it's not straightforward to test equality. I'd be happy to add a test, but I'm not sure how to best do it. 

My suggestion would be to split the test into two parts: 
1. a test that checks traits and types, for example that assigning to a non-mutable view works as expected

2. a test for the numerics: for this, one suggestion would be to either use/access `d_view` and `h_view` (since they are public members for now) and check that those are equal. Since this is a test, relying on these members to be public is ok I think. Or using `view_host` and `view_device` and checking for equality directly on these two.

masterleinad:
>  Or using view_host and view_device and checking for equality directly on these two. My preference is the second.

I would do that.

maartenarnst:
Hi @fnrizzi and @masterleinad. Thank you for your suggestions. 

- I have just added a test that checks equality of the host and device views. 
- I've added an indirect test of the modified_flags member variable through a test of the sync behavior.

@fnrizzi, I'm not sure what you mean with traits tests. Do you have a specific test in mind? Also, just a s a note, because the copy constructor and assignment operator of DualView essentially delegate to those of View, I'm not sure we should go too far in testing traits like assignability. But I'll be happy to follow your suggestion.

masterleinad:
You have to fix the indentation (using `clang-format 8`):
```diff
diff --git a/containers/unit_tests/TestDualView.hpp b/containers/unit_tests/TestDualView.hpp
index 796791c3d..54086914b 100644
--- a/containers/unit_tests/TestDualView.hpp
+++ b/containers/unit_tests/TestDualView.hpp
@@ -79,15 +79,15 @@ struct test_dualview_copy_construction_and_assignment {
     ViewType c = a;
 
     // Check equality (shallow) of the host and device views
-    ASSERT_EQ(a.view_host(),   b.view_host());
+    ASSERT_EQ(a.view_host(), b.view_host());
     ASSERT_EQ(a.view_device(), b.view_device());
 
-    ASSERT_EQ(a.view_host(),   c.view_host());
+    ASSERT_EQ(a.view_host(), c.view_host());
     ASSERT_EQ(a.view_device(), c.view_device());
 
     // We can't test shallow equality of modified_flags because it's protected.
     // So we test it indirectly through sync state behavior.
-    if ( ! std::decay_t<ViewType>::impl_dualview_is_single_device::value) {
+    if (!std::decay_t<ViewType>::impl_dualview_is_single_device::value) {
       a.clear_sync_state();
       a.modify_host();
       ASSERT_TRUE(a.need_sync_device());
```
Otherwise, this looks OK to me.

maartenarnst:
Hi @masterleinad. OK, thanks for letting me know about the issue with the formatting. I've just corrected it.

fnrizzi:
> @fnrizzi, I'm not sure what you mean with traits tests. Do you have a specific test in mind? Also, just a s a note, because the copy constructor and assignment operator of DualView essentially delegate to those of View, I'm not sure we should go too far in testing traits like assignability. But I'll be happy to follow your suggestion.

That is fine, I am ok with the test as is

containers/unit_tests/TestDualView.hpp
```diff
@@ -60,6 +60,48 @@ struct test_dualview_alloc {
...
+    const unsigned int n = 10;
+    const unsigned int m = 5;
+
+    ViewType a("A", n, m);
+
+    // Copy construction
+    ViewType b(a);
+
+    // Copy assignment
+    ViewType c = a;
```
masterleinad, line 79: This should have worked before already. Can you modify this or add another case so that the View types are compatible but not equal?

maartenarnst: Hi @masterleinad. OK, indeed. I've just updated it so that the views are compatible but not of the same type.

masterleinad (approved):

maartenarnst:
Hi @masterleinad. I've just improved the indentation. Sorry for that.

containers/unit_tests/TestDualView.hpp
```diff
@@ -60,6 +60,49 @@ struct test_dualview_alloc {
...
+  using scalar_type     = Scalar;
+  using execution_space = Device;
+
+  void run_me() {
+    constexpr unsigned int n = 10;
+    constexpr unsigned int m = 5;
+
+    using SrcViewType = Kokkos::DualView<Scalar**, Kokkos::LayoutLeft, Device>;
+    using DstViewType =
++        Kokkos::DualView<const Scalar * [m], Kokkos::LayoutLeft, Device>;
```
masterleinad, line 74:
```suggestion
        Kokkos::DualView<const Scalar * [m], Kokkos::LayoutLeft, Device>;
```

containers/unit_tests/TestDualView.hpp
```diff
@@ -379,6 +422,11 @@ void test_dualview_alloc(unsigned int size) {
   ASSERT_TRUE(test.result);
 }
 
+template <typename Scalar, typename Device>
+void test_dualview_copy_construction_and_assignment() {
+  Impl::test_dualview_copy_construction_and_assignment<Scalar, Device>();
```
fnrizzi, line 427:
```suggestion
  Impl::test_dualview_copy_construction_and_assignment<Scalar, Device>()();
```
and we get rid of the constructor and replace run_me wiht operator().

containers/unit_tests/TestDualView.hpp
```diff
@@ -404,6 +452,10 @@ TEST(TEST_CATEGORY, dualview_alloc) {
   test_dualview_alloc<int, TEST_EXECSPACE>(10);
 }
 
+TEST(TEST_CATEGORY, test_dualview_copy_construction_and_assignment) {
+  test_dualview_copy_construction_and_assignment<int, TEST_EXECSPACE>();
```
fnrizzi, line 456: similar to above

fnrizzi (approved):
the suggested changes to test are minor, i am ok with or without

containers/unit_tests/TestDualView.hpp
fnrizzi: yeah that was the thing i was saying in my suggestion for checking assignment of a compatible view

containers/unit_tests/TestDualView.hpp
maartenarnst: OK. Made the change.

containers/unit_tests/TestDualView.hpp
maartenarnst: I think no change is needed because it's still calling the version with `()` which is then in turn calling the new way with `()()`.

containers/src/Kokkos_DualView.hpp
```diff
@@ -292,6 +292,15 @@ class DualView : public ViewTraits<DataType, Properties...> {
...
         h_view(src.h_view) {}
 
+  //! Copy assignment operator (shallow copy assignment)
+  template <typename DT, typename... DP>
+  DualView& operator=(const DualView<DT, DP...>& src) {
+    modified_flags = src.modified_flags;
+    d_view         = src.d_view;
+    h_view         = src.h_view;
+    return *this;
+  }
```
dalg24, line 302: Sorry I am a bit late to the party, just going through the changelog to prepare the release 4.2 briefing, and I don't think this is needed.  Converting assignment will just work and do the same thing without us defining it.  It just goes through the converting constructor and regular copy assignment.
Am I missing anything?

https://godbolt.org/z/zehfP4brM

maartenarnst: It seems it's not needed indeed.

I think I had looked at the case of the view when I made this PR. The view has both the converting copy constructor and the converting copy assignment operator. But I now think I understand that the case if the view is different because it needs to be able to assign unmanaged from managed.

dalg24: I proposed getting rid of it in #6649

PR #6477, opened 2023-10-02
Cuda: Avoid repeated template instantiations for UniqueToken and idForInstance

masterleinad:
https://github.com/dealii/dealii/issues/16065 shows that in a `deal.II` build with `Kokkos+Cuda` some `Kokkos` templates are the most expensive templates. Specifically, these are `idForInstance` and the `Kokkos::View` used in `UniqueToken`.
This pull request provides explicit template instantiations for the former one and pushes the implementation of the `UniqueToken` constructor to a `*.cpp` file avoiding compiling these templates repeatedly.

I restricted this pull request to `Cuda` since that is what I looked at first. If this finds agreement, I'm happy to do similar changes to all backends.


core/src/Cuda/Kokkos_Cuda_UniqueToken.cpp
```diff
@@ -0,0 +1,36 @@
...
+
+#include <Kokkos_Core.hpp>
+#include <Cuda/Kokkos_Cuda_UniqueToken.hpp>
+
+namespace Kokkos::Experimental {
+
+UniqueToken<Cuda, UniqueTokenScope::Global>::UniqueToken(size_type max_size) {
+  m_locks = Kokkos::View<uint32_t*, Kokkos::CudaSpace>(
+      std::string("Kokkos::UniqueToken::m_locks"), max_size);
+}
```
dalg24, line 29:
```suggestion
}

```

core/src/Cuda/Kokkos_Cuda_UniqueToken.cpp
```diff
@@ -0,0 +1,36 @@
...
+UniqueToken<Cuda, UniqueTokenScope::Global>::UniqueToken(size_type max_size,
+                                                         Cuda const& exec) {
+  m_locks = Kokkos::View<uint32_t*, Kokkos::CudaSpace>(
+      Kokkos::view_alloc(exec, std::string("Kokkos::UniqueToken::m_locks")),
+      max_size);
+}
```
dalg24, line 35:
```suggestion
}

```

dalg24 (commented):
Did you measure improvement in build time?

masterleinad:
> Did you measure improvement in build time?

https://github.com/dealii/dealii/issues/16065#issuecomment-1745015746 includes some other fixes but none of them changed build time significantly but the respective compile items vanished from the report. Note that the builds were done with 64 processes which might hide improvements. 
I hope that introducing the `Kokkos_Abort.hpp` header helps with
```
1241653 ms: /home/darndt/dealii/include/deal.II/base/exceptions.h (included 780 times, avg 1591 ms), included via: [...]
```

core/src/Cuda/Kokkos_Cuda_UniqueToken.cpp
```diff
@@ -0,0 +1,38 @@
...
+#endif
+
+#include <Kokkos_Core.hpp>
+#include <Cuda/Kokkos_Cuda_UniqueToken.hpp>
+
+namespace Kokkos::Experimental {
+
+UniqueToken<Cuda, UniqueTokenScope::Global>::UniqueToken(size_type max_size) {
+  m_locks = Kokkos::View<uint32_t*, Kokkos::CudaSpace>(
+      std::string("Kokkos::UniqueToken::m_locks"), max_size);
```
dalg24, line 28: Why did you add explicit std::string construction?

masterleinad: Trying to have fewer template instantiations for different lengths of the C-style array. I'm happy to switch to C-style arrays again if you prefer.

dalg24: Please undo.  This denotes an issue with the View constructors, not something that should be dealt with at call site.

masterleinad: f3b1667ce

dalg24 (commented):
Please investigate why the gets instantiated at all.  It is not clear where that would be used from.

masterleinad:
Not crucial.

Issue #6473, opened 2023-10-01
HIPInternal::concurrency inconsistent with Impl::HIPInternal::m_maxWavesPerCU 


maartenarnst:
**Describe the bug**

Kokkos overrides in `Kokkos_HIP.cpp` the maximum number of warps per CU:

- https://github.com/kokkos/kokkos/blob/41cf2e51cd8e63a614aa1b4a9564a35934b805d5/core/src/HIP/Kokkos_HIP.cpp#L70-L72

However, later in the code in `Kokkos_HIP_Instance.cpp`, the overridden value is not used when computing the concurrency:

- https://github.com/kokkos/kokkos/blob/41cf2e51cd8e63a614aa1b4a9564a35934b805d5/core/src/HIP/Kokkos_HIP_Instance.cpp#L78-L82

For instance, on our Radeon Pro VII, `m_deviceProp.maxThreadsPerMultiProcessor / WarpSize` is 40, whereas `Impl::HIPInternal::m_maxWavesPerCU` is set by Kokkos to 32.

It seems this issue can be straightforwardly fixed by doing in `Kokkos_HIP_Instance.cpp`:

```
int HIPInternal::concurrency() {
  static int const concurrency = m_maxWavesPerCU * Impl::HIPTraits::WarpSize * m_deviceProp.multiProcessorCount;
  return concurrency;
}
```

If you agree with such a change or you have another suggestion, I would be happy to make a PR. Thanks in advance!

Rombur:
@maartenarnst Yes, please open a PR,

maartenarnst:
Thanks, @Rombur. It's PR #6479.

maartenarnst:
Closing as https://github.com/kokkos/kokkos/pull/6479 was merged.

PR #6479, opened 2023-10-03
HIP: Fix concurrency

maartenarnst:
This PR fixes 
- #6473.

i.e., Kokkos hardodes in `Kokkos_HIP.cpp` the maximum number of waves per CU, and the goal of this PR is to make computations of the concurrency consistent with this hardcoded number.

Longer term, should we try to work towards removing this hardcoded value, and ping somebody from AMD to bring this issue to their attention and ask if there is a way for the hip function to ideally return directly the appropriate value for `maxThreadsPerMultiProcessor`?

### DIFF
--- a/core/src/HIP/Kokkos_HIP_Instance.cpp
+++ b/core/src/HIP/Kokkos_HIP_Instance.cpp
@@ -76,8 +76,8 @@ std::size_t scratch_count(const std::size_t size) {
 //----------------------------------------------------------------------------
 
 int HIPInternal::concurrency() {
-  static int const concurrency = m_deviceProp.maxThreadsPerMultiProcessor *
-                                 m_deviceProp.multiProcessorCount;
+  static int const concurrency = m_maxThreadsPerSM * m_multiProcCount;
+
   return concurrency;
 }
 


masterleinad:
OK to test.

masterleinad:
The `Cuda` backend has the same code. It probably makes sense for that part to be consistent across the two backends.

Rombur:
> Longer term, should we try to work towards removing this hardcoded value, and ping somebody from AMD to bring this issue to their attention

It is someone from AMD that wrote that code... 

@IanBogle do you know if now there is a way to get the sustainable number of wavefronts / CU instead of the maximum of WF / CU?

Rombur (approved):

IanBogle:
I don't know off the top of my head, but I'm asking around. What do you mean by "sustainable number of wavefronts/CU"? Is it sustainable in terms of occupancy or some other characteristic?

Rombur:
Yes sustainable in term of occupancy. @arghdos wrote that code a while back
https://github.com/kokkos/kokkos/blob/41cf2e51cd8e63a614aa1b4a9564a35934b805d5/core/src/HIP/Kokkos_HIP.cpp#L70-L72

I wonder if there is a new property that would give us that number now.

IanBogle:
I don't know of any property that exposes this number, but I'll check with @arghdos when he's back in the office. I have checked recent versions of the file that is referenced in the code snippet, and it is still looks accurate.

Rombur:
Thanks for looking @IanBogle!

masterleinad (approved):

dalg24:
Unrelated OpenMPTarget failure
```
4:   sum1
4:     Which is: 7642095
4:   sum2
4:     Which is: 0
4: [  FAILED  ] openmptarget.partitioning_by_args (0 ms)

PR #6483, opened 2023-10-04
Fix NVCC warnings

masterleinad:
This pull request fixes a bunch of warnings we are currently seeing in our CI, particularly for calling `__host__` functions from `__host__` `__device__` functions and  "missing return type".
- It seems that the `if constexpr (use_std_sort)` in `BinSort` didn't entirely suppress the warning, at least not for @stanmoore1.
- `nvcc` allows using `builtin_unreachable` from 11.3.0 on which helps with warnings of type "missing return type".
- We shouldn't use `std`  functions such as `std::min` in `KOKKOS_FUNCTION`s.


algorithms/unit_tests/TestStdAlgorithmsCommon.hpp
```diff
@@ -198,7 +198,8 @@ auto create_deep_copyable_compatible_view_with_same_extent(ViewType view) {
 
   // this is needed for intel to avoid
   // error #1011: missing return statement at end of non-void function
-#if defined KOKKOS_COMPILER_INTEL
+#if defined KOKKOS_COMPILER_INTEL || \
+    (defined(KOKKOS_COMPILER_NVCC) && KOKKOS_COMPILER_NVCC >= 1130)
```
fnrizzi, line 202: nvcc complains about this too?

masterleinad: Yes, that's what I was seeing.

core/src/Kokkos_BitManipulation.hpp
dalg24, line 1: What warnings did we get on these functions?

masterleinad: "missing return statement at end of non-void function"

dalg24 (approved):

fnrizzi (approved):

stanmoore1:
I finally got a chance to check this PR and it fixes all the warnings I was seeing, thanks @masterleinad!

PR #6485, opened 2023-10-05
Fix compiling SIMD unit tests on NVIDIA

masterleinad:
This pull request fixes the issue with using SIMD `Experimental` math functions with `nvcc` we are currently seeing on `develop`:
```
/var/jenkins/workspace/Kokkos/install/include/Kokkos_SIMD_Common.hpp(452): error: "Kokkos::Experimental::simd<int32_t, Kokkos::Experimental::simd_abi::avx512_fixed_size<8>>" contains a vector, which is not supported in device code
          detected during:
            instantiation of "Kokkos::Experimental::simd<T, Abi> Kokkos::abs(const Kokkos::Experimental::simd<T, Abi> &) [with T=int32_t, Abi=Kokkos::Experimental::simd_abi::avx512_fixed_size<8>]" 
(452): here
            instantiation of "Kokkos::Experimental::simd<T, Abi> Kokkos::Experimental::abs(const Kokkos::Experimental::simd<T, Abi> &) [with T=int32_t, Abi=Kokkos::Experimental::simd_abi::avx512_fixed_size<8>]" 
/var/jenkins/workspace/Kokkos/simd/unit_tests/include/TestSIMD_MathOps.hpp(69): here
            instantiation of "void host_check_math_op_one_loader<Abi,Loader,UnaryOp,T>(UnaryOp, std::size_t, const T *) [with Abi=Kokkos::Experimental::simd_abi::avx512_fixed_size<8>, Loader=load_element_aligned, UnaryOp=absolutes, T=int32_t]" 
/var/jenkins/workspace/Kokkos/simd/unit_tests/include/TestSIMD_MathOps.hpp(77): here
            instantiation of "void host_check_math_op_all_loaders<Abi,Op,T...>(Op, std::size_t, const T *...) [with Abi=Kokkos::Experimental::simd_abi::avx512_fixed_size<8>, Op=absolutes, T=<int32_t>]" 
/var/jenkins/workspace/Kokkos/simd/unit_tests/include/TestSIMD_MathOps.hpp(93): here
            instantiation of "void host_check_all_math_ops<Abi,DataType,n>(const DataType (&)[n], const DataType (&)[n]) [with Abi=Kokkos::Experimental::simd_abi::avx512_fixed_size<8>, DataType=int32_t, n=11UL]" 
/var/jenkins/workspace/Kokkos/simd/unit_tests/include/TestSIMD_MathOps.hpp(112): here
            instantiation of "void host_check_math_ops<Abi,DataType>() [with Abi=Kokkos::Experimental::simd_abi::avx512_fixed_size<8>, DataType=int32_t]" 
/var/jenkins/workspace/Kokkos/simd/unit_tests/include/TestSIMD_MathOps.hpp(123): here
            instantiation of "void host_check_math_ops_all_types<Abi,DataTypes...>(Kokkos::Experimental::Impl::data_types<DataTypes...>) [with Abi=Kokkos::Experimental::simd_abi::avx512_fixed_size<8>, DataTypes=<int32_t, uint32_t, int64_t, uint64_t, double, float>]" 
/var/jenkins/workspace/Kokkos/simd/unit_tests/include/TestSIMD_MathOps.hpp(130): here
            instantiation of "void host_check_math_ops_all_abis(Kokkos::Experimental::Impl::abi_set<Abis...>) [with Abis=<Kokkos::Experimental::simd_abi::scalar, Kokkos::Experimental::simd_abi::avx512_fixed_size<8>>]" 
/var/jenkins/workspace/Kokkos/simd/unit_tests/include/TestSIMD_MathOps.hpp(253): here
```
We don't overload the `Experimental` math function so it makes sense that the compiler complains in the `device` pass for host-only SIMD types. Since specializations are provided for `host`-only SIMD types, this pull request annotates the `Experimental` (deprecated) functions as `host`-only.
It seems, however, that the compiler doesn't see that the correct overload referred to in the `Kokkos::Experimental` math function is annotated as host-only function. Therefore, we never call the `Experimental` function for `nvcc`.

The pull request also fixes a bunch of other warnings where we were calling host functions from host-device functions.

### DIFF
--- a/simd/unit_tests/include/SIMDTesting_Ops.hpp
+++ b/simd/unit_tests/include/SIMDTesting_Ops.hpp
@@ -80,7 +80,7 @@ class absolutes {
   template <typename T>
   auto on_host(T const& a) const {
     if constexpr (std::is_signed_v<typename T::value_type>) {
-#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+#if defined(KOKKOS_ENABLE_DEPRECATED_CODE_4) && !defined(KOKKOS_COMPILER_NVCC)
       return Kokkos::Experimental::abs(a);
 #else
       return Kokkos::abs(a);
@@ -95,11 +95,7 @@ class absolutes {
   template <typename T>
   KOKKOS_INLINE_FUNCTION auto on_device(T const& a) const {
     if constexpr (std::is_signed_v<typename T::value_type>) {
-#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
-      return Kokkos::Experimental::abs(a);
-#else
       return Kokkos::abs(a);
-#endif
     }
     return a;
   }
--- a/simd/unit_tests/include/TestSIMD_GeneratorCtors.hpp
+++ b/simd/unit_tests/include/TestSIMD_GeneratorCtors.hpp
@@ -36,20 +36,33 @@ inline void host_check_gen_ctor() {
     expected[i] = (init_mask[i]) ? init[i] * 9 : init[i];
   }
 
-  simd_type basic(KOKKOS_LAMBDA(std::size_t i) { return init[i]; });
-  mask_type mask(KOKKOS_LAMBDA(std::size_t i) { return init_mask[i]; });
-
   simd_type rhs;
   rhs.copy_from(init, Kokkos::Experimental::element_aligned_tag());
-  host_check_equality(basic, rhs, lanes);
-
-  simd_type lhs(KOKKOS_LAMBDA(std::size_t i) { return init[i] * 9; });
-  simd_type result(
-      KOKKOS_LAMBDA(std::size_t i) { return (mask[i]) ? lhs[i] : rhs[i]; });
 
   simd_type blend;
   blend.copy_from(expected, Kokkos::Experimental::element_aligned_tag());
-  host_check_equality(blend, result, lanes);
+
+  if constexpr (std::is_same_v<Abi, Kokkos::Experimental::simd_abi::scalar>) {
+    simd_type basic(KOKKOS_LAMBDA(std::size_t i) { return init[i]; });
+    host_check_equality(basic, rhs, lanes);
+
+    simd_type lhs(KOKKOS_LAMBDA(std::size_t i) { return init[i] * 9; });
+    mask_type mask(KOKKOS_LAMBDA(std::size_t i) { return init_mask[i]; });
+    simd_type result(
+        KOKKOS_LAMBDA(std::size_t i) { return (mask[i]) ? lhs[i] : rhs[i]; });
+
+    host_check_equality(blend, result, lanes);
+  } else {
+    simd_type basic([=](std::size_t i) { return init[i]; });
+    host_check_equality(basic, rhs, lanes);
+
+    simd_type lhs([=](std::size_t i) { return init[i] * 9; });
+    mask_type mask([=](std::size_t i) { return init_mask[i]; });
+    simd_type result(
+        [=](std::size_t i) { return (mask[i]) ? lhs[i] : rhs[i]; });
+
+    host_check_equality(blend, result, lanes);
+  }
 }
 
 template <typename Abi, typename... DataTypes>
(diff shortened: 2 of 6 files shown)


crtrott (approved):

simd/unit_tests/include/SIMDTesting_Ops.hpp
```diff
@@ -80,7 +80,7 @@ class absolutes {
   template <typename T>
   auto on_host(T const& a) const {
     if constexpr (std::is_signed_v<typename T::value_type>) {
-#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+#if defined(KOKKOS_ENABLE_DEPRECATED_CODE_4) && !defined(KOKKOS_COMPILER_NVCC)
```
ldh4, line 83: Do we need to guard only against nvcc? Will this be problematic for hipcc and sycl compiler as well?

simd/unit_tests/include/TestSIMD_GeneratorCtors.hpp
```diff
@@ -36,20 +36,33 @@ inline void host_check_gen_ctor() {
...
+    simd_type basic([=](std::size_t i) { return init[i]; });
+    host_check_equality(basic, rhs, lanes);
+
+    simd_type lhs([=](std::size_t i) { return init[i] * 9; });
+    mask_type mask([=](std::size_t i) { return init_mask[i]; });
+    simd_type result(
+        [=](std::size_t i) { return (mask[i]) ? lhs[i] : rhs[i]; });
+
+    host_check_equality(blend, result, lanes);
+  }
```
ldh4, line 65: I'm guessing this separation was needed because `KOKKOS_LAMBDA` being `__host__ __device__` function in device enabled builds but is being called from host-only generator constructors (non-scalar simd backends)?

ldh4 (approved):

simd/unit_tests/include/SIMDTesting_Ops.hpp
masterleinad: It seems to work fine for `HIP`, `SYCL`, and `Clang+Cuda`.

simd/unit_tests/include/TestSIMD_GeneratorCtors.hpp
masterleinad: Yes.

simd/unit_tests/include/SIMDTesting_Ops.hpp
dalg24: What is the issue here?

simd/unit_tests/include/TestSIMD_GeneratorCtors.hpp
dalg24: Is it less disgusting if you do KOKKOS_IF_ON_HOST inside the KOKKOS_LAMBDA?

masterleinad: IMHO, it's conceptionally correct (and consistent with the other constructors) for the generator constructors to be host-only.

simd/unit_tests/include/SIMDTesting_Ops.hpp
masterleinad: > t seems, however, that the compiler doesn't see that the correct overload referred to in the Kokkos::Experimental math function is annotated as host-only function. Therefore, we never call the Experimental function for nvcc.

dalg24: I do not understand the answer.  Also why do we guard here and just use `Kokkos::abs` in the next changeset?

simd/unit_tests/include/TestSIMD_GeneratorCtors.hpp
dalg24: Please clarify what builds are broken without that change.
My concern is that this code become less readable/maintainable.

dalg24: All versions?

crtrott: > Is it less disgusting if you do KOKKOS_IF_ON_HOST inside the KOKKOS_LAMBDA?

Not clear how that would work. The problem is about handing a host device function to the generator isn't it?

crtrott: Lets go with this for now. I think the splitting is clean enough.

simd/unit_tests/include/SIMDTesting_Ops.hpp
masterleinad: #6487 should fix this for good.

PR #6487, opened 2023-10-06
SIMD: Split math functions from SIMD_Common.hpp

masterleinad:
Follow-up to https://github.com/kokkos/kokkos/pull/6485#pullrequestreview-1660793788.

Splitting the math functions from `SIMD_Common.hpp` and including the new header after including the SIMD types header files fixes the error we were still seeing when using `Kokkos::Experimental::abs` on the host for host-only types with `nvcc`; with this split, `Kokkos::Experimental::abs` finds the host-only specialization for `Kokkos::abs`.


ldh4 (approved):

simd/src/Kokkos_SIMD_Common_Math.hpp
```diff
@@ -0,0 +1,256 @@
...
+//@HEADER
+
+#ifndef KOKKOS_SIMD_COMMON_MATH_HPP
+#define KOKKOS_SIMD_COMMON_MATH_HPP
+
+#include <cmath>
+#include <cstring>
+
+#include <Kokkos_Core.hpp>
+
```
fnrizzi, line 24: does this header not need `Kokkos_SIMD_Common.hpp`?

masterleinad: I made it self-contained in 9abaa33.

simd/unit_tests/include/SIMDTesting_Ops.hpp
```diff
@@ -80,7 +80,7 @@ class absolutes {
   template <typename T>
   auto on_host(T const& a) const {
     if constexpr (std::is_signed_v<typename T::value_type>) {
-#if defined(KOKKOS_ENABLE_DEPRECATED_CODE_4) && !defined(KOKKOS_COMPILER_NVCC)
+#if defined(KOKKOS_ENABLE_DEPRECATED_CODE_4)
```
dalg24, line 83: What about https://github.com/kokkos/kokkos/pull/6485/files#diff-66339256bc16b14bde61943b2f7e184363640137e9a9b03ccf93c7f1a89e3fa9L98-L102 ?
Should we revert it as well?

simd/src/Kokkos_SIMD_Common_Math.hpp
```diff
@@ -0,0 +1,262 @@
...
+namespace Experimental {
+
+template <class T, class Abi>
+class simd;
+
+template <class T, class Abi>
+class simd_mask;
+
+template <class M, class T>
+class const_where_expression;
```
dalg24, line 33: Why did you go for forward declarations rather than including `<Kokkos_SIMD_Common.hpp>`?

simd/unit_tests/include/SIMDTesting_Ops.hpp
masterleinad: Reverting that change would make us call `__host__` functions from a `__host__ __device__` function (for `host`-only types) and we would potentially run into the same error since we would be trying to use vector types in a `__host__ __device__` function.

simd/src/Kokkos_SIMD_Common_Math.hpp
masterleinad: To minimize dependencies (even though it probably doesn't matter much here). We should be able to use this header just using the `SIMD` specializations and not the fallback. 
In the end, this pull request shows that the order of overload declarations matters, and including `Kokkos_SIMD_Common.hpp` might suggest that it would be fine to include that file before the overloads for the `simd` specializations.

simd/src/Kokkos_SIMD_Common_Math.hpp
crtrott: Should we check that this is not included without Kokkos_SIMD.hpp being included first?

crtrott (changes requested):

crtrott (approved):

fnrizzi (approved):

simd/src/Kokkos_SIMD_Common_Math.hpp
```diff
@@ -0,0 +1,262 @@
...
+  return Kokkos::max(a, b);
+}
+}  // namespace Experimental
+#endif
+
+// fallback implementations of <cmath> functions.
+// individual Abi types may provide overloads with more efficient
+// implementations.
+// These are not in the Experimental namespace because their double
+// overloads are not either
```
nliber, line 122: Is this comment correct?  If I'm reading it correctly, it looks like lines 135-143 put the same named function in `Experimental` (albeit deprecated).

nliber (changes requested):

simd/src/Kokkos_SIMD_Common_Math.hpp
masterleinad: This pull request just moves code but 229b43dab fixes it anyway.

crtrott:
Retest this please.

ldh4:
The CI testing is timing out on an unrelated unit test.
```
The following tests FAILED:
	 12 - Kokkos_CoreUnitTest_LogicalSpaces (Timeout)
Error: Process completed with exit code 8.
```

Issue #6453, opened 2023-09-19
Implement Team `parallel_scan` with return value


cz4rs:
Follow-up to #5635, tracking implementation PRs.
Suggested merging order is from top to bottom.

| Backend | TeamThreadRange | ThreadVectorRange |
| -- | -- | -- |
| Serial + OpenMP | #6090 (merged) | #6292 (merged) |
| CUDA | #6301 (merged) | #6235 (merged) |
| HIP | #6302 (merged) | #6242 (merged) |
| Threads | #6303 (merged) | #6305 (merged) |
| SYCL | #6307 (merged) | #6308 (merged) |

cz4rs:
Closing after all relevant PRs were merged.

PR #6491, opened 2023-10-09
NOTHING TO SEE YET 

fnrizzi:
`PartitionedCopy` was implemented for the team level using `Kokkos::Single` because, at the time the teamlevel effort started, the proper overloads for the team scan accepting a return value were missing. 
This was obviously a temporary solution. 
Now that all the teamlevel scan overloads have been added (see https://github.com/kokkos/kokkos/issues/6453), we can fix this so that the performance is obviously better. 

FINISH THIS


fnrizzi:
superseeded by https://github.com/kokkos/kokkos/pull/6494

PR #6493, opened 2023-10-09
team-level `unique_copy`, `copy_if`: impl: replace the use of `Single` with proper parallel_scan

fnrizzi:
`UniqueCopy`, `CopyIf`  were implemented for the team level using `Kokkos::Single` because, at the time the teamlevel effort started, the proper overloads for the team scan accepting a return value were missing. 
This was obviously a temporary solution. 
Now that all the teamlevel scan overloads have been added (see https://github.com/kokkos/kokkos/issues/6453), we can fix this so that the performance is obviously better.

### DIFF
--- a/algorithms/src/std_algorithms/impl/Kokkos_MustUseKokkosSingleInTeam.hpp
+++ b/algorithms/src/std_algorithms/impl/Kokkos_MustUseKokkosSingleInTeam.hpp
@@ -1,0 +1,64 @@
+//@HEADER
+// ************************************************************************
+//
+//                        Kokkos v. 4.0
+//       Copyright (2022) National Technology & Engineering
+//               Solutions of Sandia, LLC (NTESS).
+//
+// Under the terms of Contract DE-NA0003525 with NTESS,
+// the U.S. Government retains certain rights in this software.
+//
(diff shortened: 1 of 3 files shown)


algorithms/src/std_algorithms/impl/Kokkos_MustUseKokkosSingleInTeam.hpp
```diff
@@ -0,0 +1,64 @@
...
+#define KOKKOS_STD_ALGORITHMS_MUSTUSEKOKKOSSINGLEINTEAM_HPP
+
+#include <Kokkos_Core.hpp>
+
+namespace Kokkos {
+namespace Experimental {
+namespace Impl {
+
+template <typename T>
+struct stdalgo_must_use_kokkos_single_for_team_scan : std::false_type {};
```
nliber, line 27: Add and use:

```c++
template<typename T>
inline constexpr bool stdalgo_must_use_kokkos_single_for_team_scan_v = stdalgo_must_use_kokkos_single_for_team_scan<T>::value;
```

to match the C++17 type traits.

nliber (changes requested):

nliber (approved):

crtrott (approved):

PR #6496, opened 2023-10-09
SYCL: Allow setting subgroup size

masterleinad:
There is demand for being able to set the subgroup size per kernel. This is mostly for discussion and for people to try it out. At the moment, only `RangePolicy` with `parallel_for` is supported.

### DIFF
--- a/core/src/Kokkos_Concepts.hpp
+++ b/core/src/Kokkos_Concepts.hpp
@@ -121,6 +121,13 @@ struct LaunchBounds {
   static constexpr unsigned int minBperSM{minB};
 };
 
+template <int size>
+struct SubGroupSize {
+  using subgroup_size        = SubGroupSize;
+  using type                 = SubGroupSize<size>;
+  static constexpr int value = size;
+};
+
 }  // namespace Kokkos
 
 //----------------------------------------------------------------------------
--- a/core/src/traits/Kokkos_LaunchBoundsTrait.hpp
+++ b/core/src/traits/Kokkos_LaunchBoundsTrait.hpp
@@ -62,6 +62,43 @@ struct PolicyTraitMatcher<LaunchBoundsTrait, LaunchBounds<maxT, minB>>
 // </editor-fold> end PolicyTraitMatcher specialization }}}1
 //==============================================================================
 
+//==============================================================================
+// <editor-fold desc="trait specification"> {{{1
+
+struct SubGroupSizeTrait : TraitSpecificationBase<SubGroupSizeTrait> {
+  struct base_traits {
+    static constexpr bool launch_bounds_is_defaulted = true;
+    static constexpr int subgroup_size               = -1;
+
+    KOKKOS_IMPL_MSVC_NVCC_EBO_WORKAROUND
+  };
+  template <class SubGroupSizeParam, class AnalyzeNextTrait>
+  struct mixin_matching_trait : AnalyzeNextTrait {
+    using base_t = AnalyzeNextTrait;
+    using base_t::base_t;
+
+    static constexpr bool launch_bounds_is_defaulted = false;
+
+    static_assert(base_t::launch_bounds_is_defaulted,
+                  "Kokkos Error: More than one launch_bounds given");
+
+    static constexpr int subgroup_size = SubGroupSizeParam::size;
+  };
+};
+
+// </editor-fold> end trait specification }}}1
+//==============================================================================
+
+//==============================================================================
+// <editor-fold desc="PolicyTraitMatcher specialization"> {{{1
+
+template <int size>
+struct PolicyTraitMatcher<SubGroupsSizeTrait, SubGroupSize<size>>
+    : std::true_type {};
+
+// </editor-fold> end PolicyTraitMatcher specialization }}}1
+//==============================================================================
+
 }  // end namespace Impl
 }  // end namespace Kokkos
 
--- a/core/src/traits/Kokkos_LaunchBoundsTrait.hpp
+++ b/core/src/traits/Kokkos_LaunchBoundsTrait.hpp
@@ -67,7 +67,7 @@ struct PolicyTraitMatcher<LaunchBoundsTrait, LaunchBounds<maxT, minB>>
 
 struct SubGroupSizeTrait : TraitSpecificationBase<SubGroupSizeTrait> {
   struct base_traits {
-    static constexpr bool launch_bounds_is_defaulted = true;
+    static constexpr bool subgroup_size_is_defaulted = true;
     static constexpr int subgroup_size               = -1;
 
     KOKKOS_IMPL_MSVC_NVCC_EBO_WORKAROUND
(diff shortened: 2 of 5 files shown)


core/src/traits/Kokkos_LaunchBoundsTrait.hpp
```diff
@@ -62,6 +62,47 @@ struct PolicyTraitMatcher<LaunchBoundsTrait, LaunchBounds<maxT, minB>>
 // </editor-fold> end PolicyTraitMatcher specialization }}}1
 //==============================================================================
 
+//==============================================================================
+// <editor-fold desc="trait specification"> {{{1
+
+#ifdef KOKKOS_ENABLE_SYCL
```
dalg24, line 68: Just observing that this would be the first policy trait that is only defined if a specific backend is enabled.

core/src/traits/Kokkos_LaunchBoundsTrait.hpp
```diff
@@ -62,6 +62,47 @@ struct PolicyTraitMatcher<LaunchBoundsTrait, LaunchBounds<maxT, minB>>
...
+
+struct SubGroupSizeTrait : TraitSpecificationBase<SubGroupSizeTrait> {
+  struct base_traits {
+    static constexpr bool subgroup_size_is_defaulted = true;
+    static constexpr int subgroup_size               = -1;
+
+    KOKKOS_IMPL_MSVC_NVCC_EBO_WORKAROUND
```
dalg24, line 75: If we stick with only defined with SYCL, that macro is unnecessary.

core/src/SYCL/Kokkos_SYCL_ParallelReduce_Range.hpp
```diff
@@ -322,9 +322,23 @@ class Kokkos::Impl::ParallelReduce<
         auto reduction_lambda = reduction_lambda_factory(
             local_mem, num_teams_done, results_ptr, values_per_thread);
 
+#if defined(__INTEL_LLVM_COMPILER) && __INTEL_LLVM_COMPILER >= 20230100
```
dalg24, line 325: Why are the guards different on this one?
All other occurrences so far were doing `#ifdef SYCL_EXT_ONEAPI_KERNEL_PROPERTIES`

core/src/SYCL/Kokkos_SYCL_ParallelFor_MDRange.hpp
```diff
@@ -152,7 +153,20 @@ class Kokkos::Impl::ParallelFor<FunctorType, Kokkos::MDRangePolicy<Traits...>,
             {n_global_x, n_global_y, n_global_z},
             {global_x, global_y, global_z}, {local_x, local_y, local_z})
             .exec_range();
-      });
+      };
+#ifdef SYCL_EXT_ONEAPI_KERNEL_PROPERTIES
```
dalg24, line 157: You might want to include this in the information printed by `SYCL::print_configuration()`

core/src/Kokkos_Concepts.hpp
```diff
@@ -120,6 +120,15 @@ struct LaunchBounds {
   static constexpr unsigned int minBperSM{minB};
 };
 
+#ifdef KOKKOS_ENABLE_SYCL
+template <int size>
+struct SubGroupSize {
```
dalg24, line 125: Do we want to make it `Experimental::` ?

dalg24 (commented):
> There is demand for being able to set the subgroup size per kernel.

Please elaborate.

Would you please discuss the choice of making that new property only defined if SYCL is enabled?

masterleinad:
> > There is demand for being able to set the subgroup size per kernel.
> Please elaborate.

I have been asked multiple times for a was to force the subgroup size per kernel instead of forcing globally for every kernel. Forcing a smaller subgroup size is a standard technique when dealing with register spills (to effectively double the number of registers available). Thus, this feature would be used in similar situations as `LaunchBounds` is for `Cuda` and `HIP`.

dalg24:
> We were also thinking about doing the same for `LaunchBounds` later possibly.

I don't understand what that means.

core/src/SYCL/Kokkos_SYCL_ParallelFor_MDRange.hpp
masterleinad: Sure.

dalg24:
> I have been asked multiple times for a was to force the subgroup size per kernel instead of forcing globally for every kernel. Forcing a smaller subgroup size is a standard technique when dealing with register spills (to effectively double the number of registers available). Thus, this feature would be used in similar situations as `LaunchBounds` is for `Cuda` and `HIP`.

Asked multiple times by the same person or this is coming from many?  Also you don't say anything about performance impact.  Was this tentatively deployed downstream wherever that ask was coming from?  Any feedback from it?

masterleinad:
> Asked multiple times by the same person or this is coming from many? Also you don't say anything about performance impact. Was this tentatively deployed downstream wherever that ask was coming from? Any feedback from it?

Most recently, Tim Wiliams asked about it and we compared a bunch of kernels in LAMMPS with forcing different subgroups sizes a while ago observing some performance improvements with @cjknight in comparison to the compiler's choices.

core/src/SYCL/Kokkos_SYCL.cpp
```diff
@@ -119,6 +119,27 @@ void SYCL::print_configuration(std::ostream& os, bool verbose) const {
...
+#ifdef SYCL_EXT_ONEAPI_ENQUEUE_FUNCTIONS
+  os << "macro  SYCL_EXT_ONEAPI_ENQUEUE_FUNCTIONS : defined\n";
+#else
+  os << "macro  SYCL_EXT_ONEAPI_ENQUEUE_FUNCTIONS : undefined\n";
+#endif
+#ifdef SYCL_EXT_ONEAPI_AUTO_LOCAL_RANGE
+  os << "macro  SYCL_EXT_ONEAPI_AUTO_LOCAL_RANGE : defined\n";
+#else
+  os << "macro  SYCL_EXT_ONEAPI_AUTO_LOCAL_RANGE : undefined\n";
+#endif
```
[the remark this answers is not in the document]
masterleinad: I think it's useful to print all of them. I'm happy to open another pull request for the ones not added here.

core/src/Kokkos_Concepts.hpp
```diff
@@ -120,6 +120,15 @@ struct LaunchBounds {
...
+  using subgroup_size        = SubGroupSize;
```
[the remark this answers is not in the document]
masterleinad: I took the `LaunchBounds` trait as template. Looks like we don't need the alias, though.

core/src/Kokkos_Concepts.hpp
masterleinad: I'm fine either way especillay if that mean that we can evaluate its usefulness some more before committing to keeping it.

core/src/SYCL/Kokkos_SYCL_ParallelFor_MDRange.hpp
```diff
@@ -152,7 +153,20 @@ class Kokkos::Impl::ParallelFor<FunctorType, Kokkos::MDRangePolicy<Traits...>,
...
+      auto get_properties = [&]() {
```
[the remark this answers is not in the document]
masterleinad: Fixed the capture.

core/src/SYCL/Kokkos_SYCL_ParallelReduce_Range.hpp
masterleinad: Fixed. Most have been an artifact from rebasing/merging.

masterleinad:
Needs more discussion.

PR #5020, opened 2022-05-11
Add cmake support for nvc++ as CUDA compiler

dalg24:
cc @scottbiersdorff

PR #6499, opened 2023-10-11
Allow NVHPC as device compiler only with Kokkos_ENABLE_IMPL_NVHPC_AS_DEVICE_COMPILER=ON

masterleinad:
This pull request introduces an option to allow `nvc++` as `Cuda` device compiler (with the default being "OFF").
This partially reverts https://github.com/kokkos/kokkos/pull/5020 that allowed using `nvc++` as `Cuda` device compiler.


crtrott (approved):

crtrott:
I think this is good enough for now.

masterleinad:
Retest this please.

masterleinad:
`HIP-ROCm-5.6-C++20` is timing out. all other `jenkins` CI builds, particularly the `NVHPC` builds, pass.

lucbv (approved):
That looks fine but we should try to keep these IMPL options limited as there is now way to prevent users from accessing them...

crtrott:
This is a stopgap until we get more clarity on the direction of the nvc++.

PR #6500, opened 2023-10-11
SYCL: Use host-pinned memory to copy reduction/scan result

masterleinad:
For dot-type kernels, I see an improvement between 10% and 50% in runtime when using host-pinned memory with `memcpy`  vs. device memory with `sycl::queue::memcpy` on Intel GPUs. This is analogous to what we are doing for `Cuda` and `HIP`.

Noticed while looking at #5334.

### DIFF
--- a/core/src/SYCL/Kokkos_SYCL_ParallelReduce_MDRange.hpp
+++ b/core/src/SYCL/Kokkos_SYCL_ParallelReduce_MDRange.hpp
@@ -305,6 +305,7 @@ class Kokkos::Impl::ParallelReduce<CombinedFunctorReducerType,
     // At this point, the reduced value is written to the entry in results_ptr
     // and all that is left is to copy it back to the given result pointer if
     // necessary.
+    // Using DeepCopy instead of fence+memcpy turned out to be up to 2x slower.
     if (host_result_ptr) {
       m_space.fence(
           "Kokkos::Impl::ParallelReduce<SYCL, MDRangePolicy>::execute: result "
--- a/core/src/SYCL/Kokkos_SYCL_ParallelReduce_MDRange.hpp
+++ b/core/src/SYCL/Kokkos_SYCL_ParallelReduce_MDRange.hpp
@@ -78,7 +78,7 @@ class Kokkos::Impl::ParallelReduce<CombinedFunctorReducerType,
         m_result_ptr_device_accessible(
             MemorySpaceAccess<Kokkos::Experimental::SYCLDeviceUSMSpace,
                               typename View::memory_space>::accessible),
-        m_host_scratch_lock(
+        m_scratch_buffers_lock(
             m_space.impl_internal_space_instance()->m_mutexScratchSpace) {}
 
  private:
@@ -348,7 +348,7 @@ class Kokkos::Impl::ParallelReduce<CombinedFunctorReducerType,
 
   // Only let one ParallelReduce instance at a time use the host scratch memory.
   // The constructor acquires the mutex which is released in the destructor.
-  std::scoped_lock<std::mutex> m_host_scratch_lock;
+  std::scoped_lock<std::mutex> m_scratch_buffers_lock;
 };
 
 #endif /* KOKKOS_SYCL_PARALLEL_REDUCE_MDRANGE_HPP */
--- a/core/src/SYCL/Kokkos_SYCL_ParallelReduce_MDRange.hpp
+++ b/core/src/SYCL/Kokkos_SYCL_ParallelReduce_MDRange.hpp
@@ -95,6 +95,11 @@ class Kokkos::Impl::ParallelReduce<CombinedFunctorReducerType,
     const unsigned int value_count =
         m_functor_reducer.get_reducer().value_count();
     sycl::device_ptr<value_type> results_ptr;
+    auto host_result_ptr =
+        (m_result_ptr && !m_result_ptr_device_accessible)
+            ? static_cast<sycl::host_ptr<value_type>>(
+                  instance.scratch_host(sizeof(value_type) * value_count))
+            : nullptr;
 
     sycl::event last_reduction_event;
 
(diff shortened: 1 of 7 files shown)


masterleinad:
`CUDA-12.2-NVHPC` and `OPENACC-NVHPC-CUDA-12.2` were timing out, `OPENMPTARGET-Clang` failed `openmptarget.partitioning_by_vector` which we see regularly. Everything else, In particular the `SYCL` build, passes.

Rombur (approved):

core/src/SYCL/Kokkos_SYCL_ParallelReduce_MDRange.hpp
```diff
@@ -95,6 +95,11 @@ class Kokkos::Impl::ParallelReduce<CombinedFunctorReducerType,
     const unsigned int value_count =
         m_functor_reducer.get_reducer().value_count();
     sycl::device_ptr<value_type> results_ptr;
+    auto host_result_ptr =
+        (m_result_ptr && !m_result_ptr_device_accessible)
+            ? static_cast<sycl::host_ptr<value_type>>(
+                  instance.scratch_host(sizeof(value_type) * value_count))
```
dalg24, line 101: Thread safety comes from the shared memory mutex?
If you don't rename it, at least add a comment that it also guards for the host-pinned buffer.

core/src/SYCL/Kokkos_SYCL_ParallelReduce_MDRange.hpp
```diff
@@ -296,11 +305,12 @@ class Kokkos::Impl::ParallelReduce<CombinedFunctorReducerType,
...
-      Kokkos::Impl::DeepCopy<Kokkos::Experimental::SYCLDeviceUSMSpace,
-                             Kokkos::Experimental::SYCLDeviceUSMSpace>(
-          m_space, m_result_ptr, results_ptr,
-          sizeof(*m_result_ptr) * value_count);
+    if (host_result_ptr) {
+      m_space.fence(
+          "Kokkos::Impl::ParallelReduce<SYCL, MDRangePolicy>::execute: result "
+          "not device-accessible");
+      std::memcpy(m_result_ptr, host_result_ptr,
+                  sizeof(value_type) * value_count);
```
dalg24, line 313: Why fence+std::memcpy rather than some SYCL async copy?  Is that because the lock would be released too early?

core/src/SYCL/Kokkos_SYCL_ParallelReduce_MDRange.hpp
masterleinad: 2ea1f0e16da449b714962ce331086b61afadecf2

core/src/SYCL/Kokkos_SYCL_ParallelReduce_MDRange.hpp
masterleinad:
```diff
diff --git a/core/src/SYCL/Kokkos_SYCL_ParallelReduce_Range.hpp b/core/src/SYCL/Kokkos_SYCL_ParallelReduce_Range.hpp
index c6b6fb393..be4fd7d00 100644
--- a/core/src/SYCL/Kokkos_SYCL_ParallelReduce_Range.hpp
+++ b/core/src/SYCL/Kokkos_SYCL_ParallelReduce_Range.hpp
@@ -328,9 +328,10 @@ class Kokkos::Impl::ParallelReduce<CombinedFunctorReducerType,
     // and all that is left is to copy it back to the given result pointer if
     // necessary.
     if (host_result_ptr) {
-      space.fence();
-      std::memcpy(m_result_ptr, host_result_ptr,
-                  sizeof(*m_result_ptr) * value_count);
+      Kokkos::Impl::DeepCopy<Kokkos::Experimental::SYCLDeviceUSMSpace,
+                             Kokkos::Experimental::SYCLDeviceUSMSpace>(
+          space, m_result_ptr, host_result_ptr,
+          sizeof(*m_result_ptr) * value_count);
     }
 
     return last_reduction_event;
```
is up to 2x slower.

core/src/SYCL/Kokkos_SYCL_ParallelReduce_MDRange.hpp
```diff
@@ -78,7 +78,7 @@ class Kokkos::Impl::ParallelReduce<CombinedFunctorReducerType,
         m_result_ptr_device_accessible(
             MemorySpaceAccess<Kokkos::Experimental::SYCLDeviceUSMSpace,
                               typename View::memory_space>::accessible),
-        m_shared_memory_lock(
+        m_host_scratch_lock(
```
dalg24, line 81: I don't get the new name.  Please explain.

core/src/SYCL/Kokkos_SYCL_ParallelReduce_MDRange.hpp
dalg24: Add a comment to that effect

core/src/SYCL/Kokkos_SYCL_ParallelReduce_MDRange.hpp
masterleinad: It guards access to scratch memory allocated in host-pinned/USM host space.

core/src/SYCL/Kokkos_SYCL_ParallelReduce_MDRange.hpp
masterleinad: e66f21abd9cbe5aac2a676b6b78776cd52cccf6b

core/src/SYCL/Kokkos_SYCL_ParallelReduce_MDRange.hpp
dalg24: Try `scratch_buffers_lock`

masterleinad: b914bcf9e

dalg24 (approved):

masterleinad:
Retest this please.

masterleinad:
All CI is passing.

Issue #6317, opened 2023-07-27
Consistency between Range/MDRange/TeamPolicy reduction interface and nested reduction interface


masterleinad:
As noted in , we currently don't support functors as reducers for nested reductions but for Range/MDRange/TeamPolicy reductiions. We should consider unifying the interfaces so that an example like
```C++
#include <Kokkos_Core.hpp>

struct MyValueType {
  int value;
};

struct MyFunctor {
  KOKKOS_FUNCTION void join(MyValueType& left, const MyValueType& right) const {
    left.value += right.value;
  }

  KOKKOS_FUNCTION void init(MyValueType& arg) const { arg.value = 0; }

  KOKKOS_FUNCTION void final(MyValueType& arg) const {}

  KOKKOS_FUNCTION void operator()(const int i, MyValueType& tmp) const {
    tmp.value += i;
  }
};

template <class ExecSpace>
struct Hierarchical_Red_A {
  void run(const int pN, const int sX) {
    using team_policy = Kokkos::TeamPolicy<ExecSpace>;
    using member_type = typename Kokkos::TeamPolicy<ExecSpace>::member_type;

    Kokkos::parallel_for(
        "Team", team_policy(pN, Kokkos::AUTO),
        KOKKOS_LAMBDA(const member_type& team) {
          const int n = team.league_rank();
          MyValueType out{};

          Kokkos::parallel_reduce(Kokkos::TeamThreadRange(team, sX),
                                  MyFunctor{}, out);
        });
  }
};
```
would work. Changes for the host backends could look like
```diff
diff --git a/core/src/impl/Kokkos_HostThreadTeam.hpp b/core/src/impl/Kokkos_HostThreadTeam.hpp
index 94db1f49b..6a2575f3b 100644
--- a/core/src/impl/Kokkos_HostThreadTeam.hpp
+++ b/core/src/impl/Kokkos_HostThreadTeam.hpp
@@ -543,19 +543,21 @@ class HostThreadTeamMember {
     team_reduce(reducer, reducer.reference());
   }
 
-  template <typename ReducerType>
-  KOKKOS_INLINE_FUNCTION std::enable_if_t<is_reducer<ReducerType>::value>
-  team_reduce(ReducerType const& reducer,
-              typename ReducerType::value_type contribution) const noexcept {
-    KOKKOS_IF_ON_HOST((
-        if (1 < m_data.m_team_size) {
-          using value_type = typename ReducerType::value_type;
[...]

PR #6511, opened 2023-10-13
Fix TeamThreadMDRange parallel_reduce

### LINKED ISSUE #6513: parallel_reduce on TeamthreadMDRangeMDRange does not initialize the value to identity
I under Kokkos 4.1, HIP exec space, MI250X, ROCm 5.5.1 and amdclang the following code will not initialize the reduction variable:

```c++
#include <Kokkos_Core.hpp>
#include <iostream>

int main() {
    Kokkos::ScopeGuard scope_gurad;

    Kokkos::parallel_for(
        Kokkos::TeamPolicy<>(1, Kokkos::AUTO), KOKKOS_LAMBDA(Kokkos::TeamPolicy<>::member_type a_team_handle) {
            double team_sum = 10.0;
            Kokkos::parallel_reduce(
                Kokkos::TeamThreadMDRange<Kokkos::Rank<2, Kokkos::Iterate::Default, Kokkos::Iterate::Default>,
                                          std::remove_reference_t<decltype(a_team_handle)>>(a_team_handle,
                                                                                            4, 8),
                [&](auto a, auto b, auto& thread_sum) {
                    thread_sum += 1.0;
                },
                team_sum);
            if(threadIdx.y == 0)
                // This could print garbage or whatever the value was set to
                // before the reduction, here 10.
                printf("team_sum:%f\n", team_sum);
        });

    Kokkos::fence();

    return 0;
}
```

AFAIK, the expected behavior is that the `team_sum` be initialized to the identity (0 for sum).

Initially mentioned in https://kokkosteam.slack.com/archives/C5BGU5NDQ/p1697442571746009

### LINKED ISSUE #6530: TeamThreadMDRange, TeamVectorMDRange, and ThreadVectorMDRange parallel_reduce don't reduce
As discussed in https://kokkosteam.slack.com/archives/C5BGU5NDQ/p1697194377868819, `TeamThreadMDRange`, `TeamVectorMDRange`, and `ThreadVectorMDRange` `parallel_reduce` don't actually accumulate values across threads, they rather behave as a `parallel_for` with a thread local, user-provided variable. In the end, we are missing to do the final reduction across threads. The results in
```C++
#include <Kokkos_Core.hpp>
#include <iostream>

int main()
{
  Kokkos::ScopeGuard scope_gurad;

  Kokkos::parallel_for(Kokkos::TeamPolicy<>(1, 1), KOKKOS_LAMBDA(Kokkos::TeamPolicy<>::member_type a_team_handle) {
   double team_sum = 0.0;
   Kokkos::parallel_reduce(
      Kokkos::TeamThreadMDRange<Kokkos::Rank<2, Kokkos::Iterate::Default, Kokkos::Iterate::Default>,
                              std::remove_reference_t<decltype(a_team_handle)>>(a_team_handle,
                                                                                                4, 8),
                            [&](auto a, auto b, auto& thread_sum) {
                                                thread_sum += 1.0;
                            },
       team_sum);
    Kokkos::printf("team_sum:%f\n", team_sum);
  });
  Kokkos::fence();

  return 0;
}
```
depend on the number of threads.
https://github.com/kokkos/kokkos/pull/6511 fixes this.

masterleinad:
Fixes #6513, fixes #6530. I decided both in the same pull request to minimize the number of pull request/load on the CI.

It came up in https://kokkosteam.slack.com/archives/C5BGU5NDQ/p1697207384819959?thread_ts=1697194377.868819&cid=C5BGU5NDQ that the behavior of `TeamThreadMDRange` `parallel_reduce` is surprising. It just does the reduction for a single thread and is missing the reduction across multiple team members. Also, we don't initialize the result variable.
This pull request fixes that. It appears that the same holds true for `ThreadVectorMDRange` and `TeamVectorMDRange` but we are missing a uniform `vector_reduce` implementation so that is commented at the moment.


etiennemlb:
The fix resolved the issue I posted in the above mentioned Slack discussion.

scripts/diff_files
```diff
@@ -1 +0,0 @@
-
```
cz4rs, line 1: This change somehow got included.

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -983,7 +983,10 @@ template <typename Rank, typename TeamHandle, typename Lambda,
 KOKKOS_INLINE_FUNCTION void parallel_reduce(
     TeamThreadMDRange<Rank, TeamHandle> const& policy, Lambda const& lambda,
     ReducerValueType& val) {
+  val = ReducerValueType{};
```
ldh4, line 986: I have a very vague memory that we intentionally did not default initialize the result variable here, expecting it to be handled further down the line when eventually non-md team policies are called. But seeing that the result does not reflect that thought, I think initializing the result value here works.

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -997,7 +1000,23 @@ template <typename Rank, typename TeamHandle, typename Lambda,
...
+                || std::is_same_v<typename TeamHandle::execution_space,
+                                  Kokkos::Cuda>
+#elif defined(KOKKOS_ENABLE_HIP)
+                || std::is_same_v<typename TeamHandle::execution_space,
+                                  Kokkos::HIP>
+#elif defined(KOKKOS_ENABLE_SYCL)
+                || std::is_same_v<typename TeamHandle::execution_space,
+                                  Kokkos::Experimental::SYCL>
+#endif
+  )
```
ldh4, line 1016: Do we prefer this over `if constexpr(!std::is_same_v<typename TeamHandle::execution_space, Kokkos::Serial>)`?

core/unit_test/CMakeLists.txt
```diff
@@ -214,35 +214,7 @@ foreach(Tag Threads;Serial;OpenMP;Cuda;HPX;OpenMPTarget;OpenACC;HIP;SYCL)
...
-      ViewHooks
-      ViewLayoutStrideAssignment
-      ViewMapping_a
-      ViewMapping_b
-      ViewMapping_subview
-      ViewMemoryAccessViolation
-      ViewOfClass
-      ViewResize
-      WorkGraph
-      WithoutInitializing
```
ldh4, line 245: Why are these deleted?

core/src/Kokkos_ExecPolicy.hpp
masterleinad: Let's see for which of the backends we need this in the end for the test to pass. We can discuss separately if we need a uniform `vector_reduce` implementation for all backends or not.

core/unit_test/TestTeamMDRange.hpp
```diff
@@ -169,7 +169,13 @@ struct TestTeamThreadMDRangeParallelFor : public TestTeamMDParallelFor {
...
 
     Kokkos::parallel_for(
-        Kokkos::TeamPolicy<ExecSpace>(leagueSize, Kokkos::AUTO),
+        Kokkos::TeamPolicy<ExecSpace>(
+            leagueSize, Kokkos::AUTO
+#ifndef KOKKOS_ENABLE_OPENMPTARGET
+            ,
+            Kokkos::TeamPolicy<ExecSpace>::vector_length_max()
+#endif
+                ),
```
masterleinad, line 178: We need at least 32 threads in a team and setting `vector_length_max` would give us 1 thread with `vector_length` 32.

ldh4: Does that happen even when you manually specify a number for `team_size`, or is it only when `Kokkos::AUTO` is used along with `vector_length_max`?

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Parallel.hpp
```diff
@@ -149,8 +155,9 @@ class OpenMPTargetExecTeamMember {
       }
 #pragma omp barrier
     }
-    return team_scratch[0];
+    value = team_scratch[0];
```
rgayatri23, line 158: Shouldnt this have given a warning at-least since it is now a `void` function.

masterleinad: Discussed offline that an earlier change in this pull request missed writing back to `value` when removing the `return` statement.

rgayatri23: nvm wrong question

dalg24 (commented):
Why is the OpenMPTarget not proposed as its own PR?
Are they not more unit tests that can be enabled with that fix?

masterleinad:
> Why is the OpenMPTarget not proposed as its own PR?
> Are they not more unit tests that can be enabled with that fix?

My motivation was to minimize the number of places where we would disable/modify a test (that wasn't using the respective functionality in the intended way) and this fix in the `OpenMPTarget` was the only one to make the adjusted tests pass on all backends (the test was enabled for in the first place). I would expect there to be some more tests that could be enabled after fixing the `team_reduce` interface to be in line with the other backends but don't have a strong opinion if @rgayatri23 (or someone else) does that before or after this pull request.

core/unit_test/TestTeamMDRange.hpp
masterleinad: We are now using `vector_length==2` if `OpenMPTarget` was enabled which seems to work fine as well.

core/unit_test/TestTeamMDRange.hpp
```diff
@@ -923,7 +1049,13 @@ struct TestTeamThreadMDRangeParallelReduce : public TestTeamMDParallelReduce {
...
               },
               teamSum);
-          leagueSum += teamSum;
+      // FIXME_OPENMPTARGET
+#ifdef KOKKOS_ENABLE_OPENMPTARGET
+          if (team.team_rank() == 0) leagueSum += teamSum;
+#else
+          Kokkos::single(Kokkos::PerTeam(team),
+                         [&]() { leagueSum += teamSum; });
+#endif
```
ldh4, line 1058: Could you remind me why `Kokkos::single` with `PerTeam` can't be used for `OMPTarget` here?

ldh4 (approved):

core/unit_test/TestTeamMDRange.hpp
masterleinad: It failed to compile.

core/src/Kokkos_ExecPolicy.hpp
dalg24: Do we exclusively support sum reductions?
Otherwise initializing this way does not look right to me.

masterleinad: Yes, we currently only allow sum reductions for this interface.

dalg24: Do we have a static assert to detect if a non-sum reducer or if a functor with a init/join is passed?

masterleinad: I added some `static_assert`s for the `value_type` in 1c85dd0b71fd1b0a2f4eb2d4a655f524e0fb4c96.
It shouldn't be that difficult to extend the interface to support the full reduction interface but I would rather do that in follow-up pull request.

masterleinad: Also, note that the nested reduction interface doesn't allow functors as reducers yet, see https://github.com/kokkos/kokkos/issues/6317.

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -983,7 +983,16 @@ template <typename Rank, typename TeamHandle, typename Lambda,
 KOKKOS_INLINE_FUNCTION void parallel_reduce(
     TeamThreadMDRange<Rank, TeamHandle> const& policy, Lambda const& lambda,
     ReducerValueType& val) {
+  static_assert(/*!Kokkos::is_view_v<ReducerValueType> &&*/
```
masterleinad, line 986: `is_view_v` requires pulling in `Kokkos_View.hpp` which in turn pulls in this header. To avoid the circular dependency, I just commented this particular check.

masterleinad:
> Why is the OpenMPTarget not proposed as its own PR?
> Are they not more unit tests that can be enabled with that fix?

Not quite sure. The motivation here was to not disable more tests. I wanted to let @rgayatri23 explore (in a different pull request) if more unit tests could be enabled when providing the correct interface but the one enabled additionally here is the only one that seemed obvious.

crtrott (approved):

PR #6514, opened 2023-10-16
Threads: replace enum with constexpr int and enum class

Rombur:
This PR replaces the `enum` in `Kokkos_Threads_Instance` with `constexpr int` and `enum class` and move the `enum class` to its own file. The PR also moves `Kokkos_Spinwait.hpp` from `Impl` to `Threads`. `spinwait_while_equal` now takes a `ThreadState` as an argument. The template parameter has been removed. The following functions, which where not used, have been removed:
 - `root_spinwait_while_equal`
 - `root_spinwait_until_equal`
 - `spinwait_until_equal`
 - `yield_until_equal`

### DIFF
--- a/core/src/Threads/Kokkos_Threads_Instance.cpp
+++ b/core/src/Threads/Kokkos_Threads_Instance.cpp
@@ -690,7 +690,7 @@ void ThreadsInternal::initialize(int thread_count_arg) {
     for (unsigned ith = thread_spawn_begin; ith < thread_count; ++ith) {
       // Try to protect against cache coherency failure by casting to volatile.
       ThreadsInternal *const th =
-          ((ThreadsInternal *volatile *)s_threads_exec)[ith];
+          ((ThreadsInternal * volatile *)s_threads_exec)[ith];
       if (th) {
         wait_yield(th->m_pool_state, ThreadState::Active);
       } else {
--- a/core/src/Threads/Kokkos_Threads_Instance.cpp
+++ b/core/src/Threads/Kokkos_Threads_Instance.cpp
@@ -133,7 +133,7 @@ void ThreadsInternal::global_unlock() {
 
 //----------------------------------------------------------------------------
 
-void ThreadsInternal::wait_yield(volatile int &flag, const int value) {
+void ThreadsInternal::wait_yield(volatile State &flag, const State value) {
   while (value == flag) {
     std::this_thread::yield();
   }
@@ -146,13 +146,13 @@ void ThreadsInternal::driver() {
 
   ThreadsInternal this_thread;
 
-  while (this_thread.m_pool_state == ThreadsInternal::Active) {
+  while (this_thread.m_pool_state == State::Active) {
     (*s_current_function)(this_thread, s_current_function_arg);
 
     // Deactivate thread and wait for reactivation
-    this_thread.m_pool_state = ThreadsInternal::Inactive;
+    this_thread.m_pool_state = State::Inactive;
 
-    wait_yield(this_thread.m_pool_state, ThreadsInternal::Inactive);
+    wait_yield(this_thread.m_pool_state, State::Inactive);
   }
 }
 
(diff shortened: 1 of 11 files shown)


core/src/Threads/Kokkos_Threads_Instance.cpp
```diff
@@ -278,12 +278,11 @@ void ThreadsInternal::execute_sleep(ThreadsInternal &exec, const void *) {
   const int rank_rev = exec.m_pool_size - (exec.m_pool_rank + 1);
 
   for (int i = 0; i < n; ++i) {
-    Impl::spinwait_while_equal<int>(
-        exec.m_pool_base[rank_rev + (1 << i)]->m_pool_state,
-        ThreadsInternal::Active);
+    Impl::spinwait_while_equal<State>(
+        exec.m_pool_base[rank_rev + (1 << i)]->m_pool_state, State::Active);
```
dalg24, line 282:
```suggestion
    Impl::spinwait_while_equal(m_pool_base[rev_rank + (1 << i)]->m_pool_state,
                               State::Active);
```
Same comment everywhere.  I don't think there is any point on specifying the template parameters.

core/src/impl/Kokkos_Spinwait.hpp
```diff
@@ -38,8 +38,8 @@ enum class WaitMode : int {
 void host_thread_yield(const uint32_t i, const WaitMode mode);
 
 template <typename T>
-std::enable_if_t<std::is_integral<T>::value, void> root_spinwait_while_equal(
-    T const volatile& flag, const T value) {
+std::enable_if_t<std::is_integral_v<T> || std::is_enum_v<T>, void>
+root_spinwait_while_equal(T const volatile& flag, const T value) {
```
dalg24, line 42: Why are these templated at all?  Was it some way of avoiding circular dependencies?

Rombur: Yeah but now that I think about it more, I could also move the `enum` outside of the class into a separate file, move `Kokkos_Spinwait` to the `Threads` backend folder, and remove the template parameter. Not sure if one solution is better than the other.

Rombur:
I moved `Kokkos_Spinwait` to `Threads` and removed the unused spinwait/yield functions .

core/src/Threads/Kokkos_Threads_Spinwait.hpp
```diff
@@ -0,0 +1,48 @@
...
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
+#ifndef KOKKOS_THREADS_SPINWAIT_HPP
+#define KOKKOS_THREADS_SPINWAIT_HPP
+
+#include <Kokkos_Macros.hpp>
+#include <Kokkos_Atomic.hpp>
```
masterleinad, line 21: Why do you need `atomics` here?

core/src/Threads/Kokkos_Threads_Spinwait.hpp
```diff
@@ -0,0 +1,48 @@
...
+
+#include <Threads/Kokkos_Threads_State.hpp>
+
+#include <cstdint>
+
+#include <type_traits>
```
masterleinad, line 27: What do you need this header for?

core/src/Threads/Kokkos_Threads_Spinwait.hpp
```diff
@@ -0,0 +1,48 @@
...
+// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
+#ifndef KOKKOS_THREADS_SPINWAIT_HPP
+#define KOKKOS_THREADS_SPINWAIT_HPP
+
+#include <Kokkos_Macros.hpp>
```
masterleinad, line 20: Are you using ay macros explicitly?

Rombur:
Retest this please

dalg24 (commented):
Please update the description and document what functions are being dropped because they were not used.

dalg24 (approved):

masterleinad (approved):

dalg24:
Ignoring all the timeouts

Issue #6515, opened 2023-10-16
Nightly build errors with cuda/11.4.2+gcc/8.3.0: Kokkos_Bitset.hpp:38:26: error: '__T0' was not declared in this scope


ndellingwood:
**Describe the bug**
Nightly Cuda builds with cuda/11.4.2+gcc/8.3.0 are failing to compile with error output:
```
02:06:15 /data/jenkins-new/workspace/kokkos_semsrhel7gpu01_cuda1142_gcc830_cpp17/kokkos/containers/unit_tests/../src/Kokkos_Bitset.hpp:38:26: error: '__T0' was not declared in this scope
02:06:15      auto new_ctor_props(view_ctor_prop);
02:06:15                      ~~~~~^~~~~~~~~~~~~~~  
02:06:15 /data/jenkins-new/workspace/kokkos_semsrhel7gpu01_cuda1142_gcc830_cpp17/kokkos/containers/unit_tests/../src/Kokkos_Bitset.hpp:38:26: note: suggested alternative: '__T1'
02:06:15      auto new_ctor_props(view_ctor_prop);
02:06:15                      ~~~~~^~~~~~~~~~~~~~~  
02:06:15                      __T1
02:06:15 /data/jenkins-new/workspace/kokkos_semsrhel7gpu01_cuda1142_gcc830_cpp17/kokkos/containers/unit_tests/../src/Kokkos_Bitset.hpp:41:8: error: unable to deduce 'auto' from 'new_ctor_props'
02:06:15      return new_ctor_props;
02:06:15         ^~~~~~~~~~~~~~
02:06:15 /data/jenkins-new/workspace/kokkos_semsrhel7gpu01_cuda1142_gcc830_cpp17/kokkos/containers/unit_tests/../src/Kokkos_Bitset.hpp: In instantiation of 'Kokkos::Bitset<Device>::Bitset(const Kokkos::Impl::ViewCtorProp<Args ...>&, unsigned int) [with P = {}; Device = Kokkos::Serial]':
02:06:15 /data/jenkins-new/workspace/kokkos_semsrhel7gpu01_cuda1142_gcc830_cpp17/kokkos/containers/unit_tests/../src/Kokkos_Bitset.hpp:96:79:   required from 'Kokkos::Bitset<Device>::Bitset(unsigned int) [with Device = Kokkos::Serial]'
[...]

PR #6518, opened 2023-10-17
`Kokkos_BitSet.hpp`: spell out type explicitly inside `Impl::with_updated_label`

fnrizzi:
Fixes https://github.com/kokkos/kokkos/issues/6515.
Still exploring why auto fails. But since here the issue is just a copy constructor and we know the type, i don't see why not making things explicit. 

This works on kokkos-dev-2.

### DIFF
--- a/containers/src/Kokkos_Bitset.hpp
+++ b/containers/src/Kokkos_Bitset.hpp
@@ -33,9 +33,10 @@ namespace Impl {
 template <typename... P>
 auto with_updated_label(const ViewCtorProp<P...>& view_ctor_prop,
                         const std::string& label) {
+  using vcp_t = ViewCtorProp<P...>;
   //! If the label property is already set, append. Otherwise, set label.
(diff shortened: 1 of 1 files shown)


masterleinad (approved):

masterleinad:
> Still exploring why auto fails. But since here the issue is just a copy constructor and we know the type, i don't see why not making things explicit.

Looks very much like a `nvcc` compiler bug. 🙂

ndellingwood:
This will need cherry-pick to release-candidate-4.2.00

ndellingwood:
CI failure due to timeout:

```
Timeout has been exceeded
```

Does this merit a retest, or is this a known spurious issue in which case this is ready for merge?

ndellingwood:
The timeout was with the OPENACC-NVHPC-CUDA-12.2 job, though all tests passed:

```
100% tests passed, 0 tests failed out of 53
```

Issue #6467, opened 2023-09-25
Bitset: should it be default constructed ?


romintomasetti:
## Summary

This issue's purpose is to discuss whether or not `Kokkos::Bitset` should be:
- [:one:] Default constructed *by the compiler*, much like `Kokkos::View` is. It is therefore completely **default-initialized**.
- [:two:] Default constructed *with a defaulted zero size*, as it is the case now. All members of `Kokkos::Bitset` (like views) are initialized **to a zero size**.

### Context

In PR #6198, I made `Kokkos::Bitset` completely default constructed *by the compiler* (more or less along these lines):
 ```c++
 KOKKOS_DEFAULTED_FUNCTION
  Bitset() = default;

  /// constructor
  /// arg_size := number of bit in set
  explicit Bitset(unsigned arg_size) : Bitset(Kokkos::view_alloc(), arg_size) {}

  template <class... P>
  Bitset(const Impl::ViewCtorProp<P...>& arg_prop, unsigned arg_size)
```

but  @masterleinad  wanted to discuss this change in a seperate issue/PR. The current constructors of `Kokkos::Bitset` are:
```c++
  /// constructor
  /// arg_size := number of bit in set
  Bitset(unsigned arg_size = 0u) : Bitset(Kokkos::view_alloc(), arg_size) {}

  template <class... P>
  Bitset(const Impl::ViewCtorProp<P...>& arg_prop, unsigned arg_size)
```

### Proposition

Make `Kokkos::Bitset` default constructed.

### Possible backward compatibility issues

The only issue I see transitioning from situation :two: to :one: is that if someone wrote code like:
```c++
auto my_bitset = Kokkos::Bitset(); // for now, this is equivalent to writing Bitset(0)
ASSERT(my_bitset.is_allocated()); // for now this is OK
```
then using the defaulted compiler constructor would break the assertion. Not sure anyone wrote such a code though (except in `TestBitset.hpp` of course :wink:).

However, I think having the size defaulted to 0 is weird, because it then sets `Kokkos::Bitset::m_blocks` to a size 0, so I guess it goes through the whole `Kokkos::View` initialization mechanisms, which is quite unexpected for a default constructed device container... right?

If Kokkos team agrees on making the changes to situration :one: I can take care of the PR.

ajpowelsnl:
Thanks, @romintomasetti.  I'll put it on the agenda for discussion.

masterleinad:
[...]

PR #6524, opened 2023-10-18
Prefer defaulted default constructor for Bitset

dalg24:
The default argument to the constructor that takes the size of the bitset was deferring to another constructor that creates an empty view with a label argument.  This alocates 128 bits for the view header. This showed when constructing an UnorderedMap with a pointless 128-bit "header-only" allocation which implies an unnecessary fence.

cc @uliegecsm @romintomasetti

### DIFF
--- a/containers/src/Kokkos_Bitset.hpp
+++ b/containers/src/Kokkos_Bitset.hpp
@@ -92,9 +92,10 @@ class Bitset {
   using block_view_type = View<unsigned*, Device, MemoryTraits<RandomAccess>>;
 
  public:
-  /// constructor
+  Bitset() = default;
+
   /// arg_size := number of bit in set
-  Bitset(unsigned arg_size = 0u) : Bitset(Kokkos::view_alloc(), arg_size) {}
+  Bitset(unsigned arg_size) : Bitset(Kokkos::view_alloc(), arg_size) {}
 
   template <class... P>
   Bitset(const Impl::ViewCtorProp<P...>& arg_prop, unsigned arg_size)
@@ -310,8 +311,8 @@ class Bitset {
   }
 
  private:
-  unsigned m_size;
-  unsigned m_last_block_mask;
+  unsigned m_size            = 0;
+  unsigned m_last_block_mask = 0;
   block_view_type m_blocks;
 
  private:
--- a/containers/unit_tests/TestBitset.hpp
+++ b/containers/unit_tests/TestBitset.hpp
@@ -155,7 +155,7 @@ void test_bitset() {
 
   {
     unsigned ts = 100u;
-    bitset_type b1;
+    bitset_type b1(Kokkos::view_alloc("MyBitset"), 0);
     ASSERT_TRUE(b1.is_allocated());
 
     b1 = bitset_type(ts);
@@ -165,6 +165,9 @@ void test_bitset() {
     ASSERT_TRUE(b1.is_allocated());
     ASSERT_TRUE(b2.is_allocated());
     ASSERT_TRUE(b3.is_allocated());
+
+    bitset_type b4;
+    ASSERT_FALSE(b4.is_allocated());
   }
 
   std::array<unsigned, 7> test_sizes = {
--- a/containers/unit_tests/TestBitset.hpp
+++ b/containers/unit_tests/TestBitset.hpp
@@ -260,26 +260,6 @@ TEST(TEST_CATEGORY, bitset_default_constructor_no_alloc) {
   listen_tool_events(Config::DisableAll());
 }
 
-TEST(TEST_CATEGORY, bitset_zero_size_does_alloc) {
-  using namespace Kokkos::Test::Tools;
-  listen_tool_events(Config::DisableAll(), Config::EnableAllocs());
-
-  auto success = validate_existence(
-      [&]() {
-        Kokkos::Bitset bs1(0);
-        EXPECT_TRUE(bs1.is_allocated());
-
-        Kokkos::Bitset bs2(Kokkos::view_alloc("MyBitset"), 0);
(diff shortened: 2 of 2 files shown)


containers/src/Kokkos_Bitset.hpp
```diff
@@ -92,9 +92,10 @@ class Bitset {
   using block_view_type = View<unsigned*, Device, MemoryTraits<RandomAccess>>;
 
  public:
-  /// constructor
+  Bitset() = default;
```
romintomasetti, line 95: Just wondering... Is `KOKKOS_DEFAULTED_FUNCTION` still needed to make the bitset default constructible on device?

romintomasetti:
> Fixes #6467.

@masterleinad I fixes part of #6467 I think. What about https://github.com/kokkos/kokkos/issues/6467#issuecomment-1749076147 ?

containers/src/Kokkos_Bitset.hpp
dalg24: It was not constructible from device before and I see no good reason to change it here

masterleinad:
> @masterleinad I fixes part of #6467 I think. What about [#6467 (comment)](https://github.com/kokkos/kokkos/issues/6467#issuecomment-1749076147) ?

That's a separate discussion and isn't reflected in the title or description of the pull request and only serves as a motivation. Open a separate issue for it if you want to discuss it in a larger context.

containers/src/Kokkos_Bitset.hpp
```diff
@@ -92,9 +92,10 @@ class Bitset {
...
+
   /// arg_size := number of bit in set
-  Bitset(unsigned arg_size = 0u) : Bitset(Kokkos::view_alloc(), arg_size) {}
+  Bitset(unsigned arg_size) : Bitset(Kokkos::view_alloc(), arg_size) {}
```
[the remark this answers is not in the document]
dalg24: I tend to agree with the suggested change but I will note that it was not explicit before and that is besides the scope of this PR.
That said I will apply the suggested change if other reviewers would like me to.

containers/src/Kokkos_Bitset.hpp
romintomasetti: Ok!

romintomasetti:
@dal24 @masterleinad Would you like me to open a new issue with https://github.com/kokkos/kokkos/issues/6467#issuecomment-1749076147 ?

dalg24:
> @dal24 @masterleinad Would you like me to open a new issue with [#6467 (comment)](https://github.com/kokkos/kokkos/issues/6467#issuecomment-1749076147) ?

Please do so

romintomasetti:
> > @dal24 @masterleinad Would you like me to open a new issue with [#6467 (comment)](https://github.com/kokkos/kokkos/issues/6467#issuecomment-1749076147) ?
> 
> Please do so

Just did, it is #6526. Please tell me if the title is OK with you :wink:

masterleinad:
```
/__w/kokkos/kokkos/containers/unit_tests/TestBitset.hpp:159: Failure
Value of: b1.is_allocated()
  Actual: false
Expected: true
```
which was anticipated in https://github.com/kokkos/kokkos/issues/6467#issue-1911018394.

masterleinad (approved):

dalg24:
SYCL build timed out but I intend to ignore if I can get a 2nd approval

containers/unit_tests/TestBitset.hpp
```diff
@@ -155,7 +155,7 @@ void test_bitset() {
 
   {
     unsigned ts = 100u;
-    bitset_type b1;
+    bitset_type b1(Kokkos::view_alloc("MyBitset"), 0);
```
romintomasetti, line 158: For such tests, shouldn't `Kokkos::Test::Tools` with `Config::EnableAllocs` be used ?

dalg24: Sure

romintomasetti (approved):

containers/unit_tests/TestBitset.hpp
```diff
@@ -237,6 +242,44 @@ void test_bitset() {
...
+        Kokkos::Bitset bs1(0);
+        EXPECT_TRUE(bs1.is_allocated());
+
+        Kokkos::Bitset bs2(Kokkos::view_alloc("MyBitset"), 0);
+        EXPECT_TRUE(bs2.is_allocated());
+      },
+      [&](AllocateDataEvent) {
+        return MatchDiagnostic{true, {"Found alloc event"}};
+      });
+  ASSERT_TRUE(success);
```
romintomasetti, line 278: Shouldn't this test be done in 2 distinct `validate_existence` ? If I understand it correctly, in its current state, the first allocation `bs1(0)` could in fact not allocate anything and the test would still pass if `bs2` allocates something?

Not requesting any change though, I just wanted to make sure I understand `validate_existence` :smile:

BTW is there any helper in the `Kokkos` testing toolbox that would not only allow us to ensure that some event occurred (in this case `AllocateDataEvent`) but also to count them ?

dalg24: You are correct and that is how I had written the test initially but I didn't think it was worth the boiler plate code.

I don't know that there is currently something to count the number of allocations even though writing directly a tool that does that would be trivial.

Rombur (approved):

containers/src/Kokkos_Bitset.hpp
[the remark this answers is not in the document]
crtrott: May need to do this with a "deprecated" guard.

crtrott:
Three of the CUDA builds seg faulted during compilation. The CUDA 12 build was fine.

dalg24:
Retest this please

PR #6198, opened 2023-06-08
UnorderedMap(space instance): proposal for #6067

romintomasetti:
This is a proposal for:
- #6067

1. I needed to modify both `UnorderedMap` and `Bitset` to ensure no global fencing occurs.
2. Also, the member views are now always named either as `<user provided object label> - <member hardcoded label>` or if no user label is given then a default is used. This is all done using the new `with_updated_label`.
3. Now, `Bitset` is not default initialized in the default constructor (size was before hardcoded to zero)
4. Miscellaneous changes in `Bitset` due to `clang format` 
5. 2 tests in `TestUnorderedMap` were not running using the device though I think they should so I modified them

- [ ] Check that `with_updated_label` is semantically OK and check where it should be (not sure it should be defined in `Kokkos_UnorderMap.hpp` nor in `Kokkos_Bitset.hpp`)

PR #6525, opened 2023-10-18
Do not append " - blocks" to the bitset label

dalg24:
Fixup for #6198

Rational:  Unnecessary as there is only one view (in contrast to UnorderedMap that has multiple accessory views).  IMO this is something that should rather be dealt with at call site.

code/label | Before this PR | After this PR
--- | --- | ---
`Bitset<D> bs(N);` | "Bitset - blocks" | "Bitset"
`Bitset<D> bs(view_alloc("MyLbl"), N);` | "MyLbl - blocks" | "MyLbl"
`UnorderedMap<K,V,D> m;` | "UnorderedMap - bitset - blocks" | "UnorderedMap - bitset"
`UnorderedMap<K,V,D> m(view_alloc("MyLbl"));` | "MyLbl - bitset - blocks" | "MyLbl - bitset"

cc @uliegecsm @romintomasetti

### DIFF
--- a/containers/src/impl/Kokkos_UnorderedMap_impl.hpp
+++ b/containers/src/impl/Kokkos_UnorderedMap_impl.hpp
@@ -27,6 +27,22 @@
 namespace Kokkos {
 namespace Impl {
 
+//! Either append to the label if the property already exists, or set it.
+template <typename... P>
+auto with_updated_label(const ViewCtorProp<P...>& view_ctor_prop,
+                        const std::string& label) {
+  using vcp_t = ViewCtorProp<P...>;
+  //! If the label property is already set, append. Otherwise, set label.
+  if constexpr (vcp_t::has_label) {
+    vcp_t new_ctor_props(view_ctor_prop);
+    static_cast<ViewCtorProp<void, std::string>&>(new_ctor_props)
+        .value.append(label);
+    return new_ctor_props;
+  } else {
+    return Impl::with_properties_if_unset(view_ctor_prop, label);
+  }
+}
+
 uint32_t find_hash_size(uint32_t size);
 
 template <typename Map>
--- a/containers/src/impl/Kokkos_UnorderedMap_impl.hpp
+++ b/containers/src/impl/Kokkos_UnorderedMap_impl.hpp
@@ -27,20 +27,16 @@
 namespace Kokkos {
 namespace Impl {
 
-//! Either append to the label if the property already exists, or set it.
+//! Append to the label contained in view_ctor_prop.
 template <typename... P>
-auto with_updated_label(const ViewCtorProp<P...>& view_ctor_prop,
-                        const std::string& label) {
+auto append_to_label(const ViewCtorProp<P...>& view_ctor_prop,
+                     const std::string& label) {
   using vcp_t = ViewCtorProp<P...>;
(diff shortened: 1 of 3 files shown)


romintomasetti:
I can't approve, but you get my approval :+1:

containers/src/impl/Kokkos_UnorderedMap_impl.hpp
```diff
@@ -27,6 +27,22 @@
...
+  //! If the label property is already set, append. Otherwise, set label.
+  if constexpr (vcp_t::has_label) {
+    vcp_t new_ctor_props(view_ctor_prop);
+    static_cast<ViewCtorProp<void, std::string>&>(new_ctor_props)
+        .value.append(label);
+    return new_ctor_props;
+  } else {
+    return Impl::with_properties_if_unset(view_ctor_prop, label);
+  }
+}
```
masterleinad, line 44: AFAICT the label exists whenever this function is called.
```suggestion
//! Append to the label.
template <typename... P>
auto append_to_label(const ViewCtorProp<P...>& view_ctor_prop,
                        const std::string& label) {
  static_assert (vcp_t::has_label); {
  ViewCtorProp<P...> new_ctor_props(view_ctor_prop);
  static_cast<ViewCtorProp<void, std::string>&>(new_ctor_props)
        .value.append(label);
  return new_ctor_props;
}
```

dalg24: Feel free to push to my branch (make sure you format it properly)

romintomasetti (approved):

containers/src/impl/Kokkos_UnorderedMap_impl.hpp
```diff
@@ -27,6 +27,18 @@
...
+auto append_to_label(const ViewCtorProp<P...>& view_ctor_prop,
+                     const std::string& label) {
+  using vcp_t = ViewCtorProp<P...>;
+  static_assert(vcp_t::has_label);
+  vcp_t new_ctor_props(view_ctor_prop);
+  static_cast<ViewCtorProp<void, std::string>&>(new_ctor_props)
+      .value.append(label);
+  return new_ctor_props;
+}
+
```
fnrizzi, line 41: just a thought: seems like this function (or variation of it) is being moved around where needed, 
and right here this is only used for unorderedmap. However this function has clearly no ties to unoredered map but just pertain to the ViewCtorProp. What about moving it closer to where ViewCtorProp is defined? It could be that, maybe, using this appending fucntion is something that could potentially be used elsewhere. No idea yet where, but keeping it closer to ViewCtorProp and maybe renaming it makes more sense to me and we don't have to keep moving it around.

fnrizzi (approved):
left a comment but not blocking on it

masterleinad (approved):

containers/src/impl/Kokkos_UnorderedMap_impl.hpp
dalg24: I generally agree with your suggestion but I would rather have this tackled in some other follow up PR.
Are they other such functions that would need to be relocated?

fnrizzi: no sure about other functions, i can try to look into it

PR #6531, opened 2023-10-18
SYCL: Use SYCL_EXT_ONEAPI_DEVICE_GLOBAL to detect support for device global variables

masterleinad:
When `SYCL_EXT_ONEAPI_DEVICE_GLOBAL` is defined, we can at least get rid of
```
sycl::ext::oneapi::experimental::properties(sycl::ext::oneapi::experimental::device_image_scope)
``` 
which forces us to use `-fsycl-device-code-split=off` instead of, say, `-fsycl-device-code-split=per_kernel`. Being able to use the latter flag vastly improves compile time in combination with `-fsycl-max-parallel-link-jobs=N`. Unfortunately, `sycl::device_global` variables still don't work when compiling with shared libraries.

Corresponds to https://github.com/desul/desul/pull/113.

### DIFF
--- a/cmake/kokkos_arch.cmake
+++ b/cmake/kokkos_arch.cmake
@@ -588,32 +588,39 @@ IF (KOKKOS_ENABLE_SYCL)
 ENDIF()
 
 # Check support for device_global variables
-# FIXME_SYCL Once the feature test macro SYCL_EXT_ONEAPI_DEVICE_GLOBAL is
-#            available, use that instead.
+# FIXME_SYCL Even if SYCL_EXT_ONEAPI_DEVICE_GLOBAL is defined, we still can't
+#            use device global variables with shared libraries
 IF(KOKKOS_ENABLE_SYCL AND NOT BUILD_SHARED_LIBS)
-  INCLUDE(CheckCXXSourceCompiles)
   STRING(REPLACE ";" " " CMAKE_REQUIRED_FLAGS "${KOKKOS_COMPILE_OPTIONS}")
-  CHECK_CXX_SOURCE_COMPILES("
-    #include <sycl/sycl.hpp>
-    using namespace sycl::ext::oneapi::experimental;
-    using namespace sycl;
+  INCLUDE(CheckCXXSymbolExists)
+  CHECK_CXX_SYMBOL_EXISTS(SYCL_EXT_ONEAPI_DEVICE_GLOBAL "sycl/sycl.hpp" KOKKOS_IMPL_HAVE_SYCL_EXT_ONEAPI_DEVICE_GLOBAL)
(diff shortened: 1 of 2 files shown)


cmake/kokkos_arch.cmake
```diff
@@ -588,32 +588,39 @@ IF (KOKKOS_ENABLE_SYCL)
 ENDIF()
 
 # Check support for device_global variables
-# FIXME_SYCL Once the feature test macro SYCL_EXT_ONEAPI_DEVICE_GLOBAL is
-#            available, use that instead.
+# FIXME_SYCL Even if SYCL_EXT_ONEAPI_DEVICE_GLOBAL is defined, we still can't
+#            use device global variables with shared libraries
```
trws, line 592: This is true for all the models except the host right? Or is there some extra restriction for sycl?

masterleinad: I couldn't find much about that restriction in the SYCL specs and didn't get a clear answer from Intel. I'm happy to explore if we can now have a non-`SYCL_EXTERNAL`/rdc option (like we have for the other backends), which might allow the use of shared libraries. 
I only mentioned it specifically since some functionality needs shared libraries (and `ZES_ENABLE_SYSMAN=1`) like querying free and total device memory.

trws: When you say “needs shared libraries” what do you mean? As far as I know, none of the backends support having device code in one shared object call device code in another one.  Is it more than that, like code in one shared object can’t dereference an address from another? Sorry if I’m being obtuse, I’m trying to tease out what the specific limitation is so I can keep track of it, and thank you for the explanation so far.

masterleinad: > When you say “needs shared libraries” what do you mean?

`KokkosKernels` needs/has/wants functionality for querying free and total device memory and the implementation requires `Kokkos` and `KokkosKernels` to be built as shared libraries (and setting `ZES_ENABLE_SYSMAN=1`). That is unrelated to how global device variables are used. That means that we can't support arbitrary-sized atomics and that memory query at the same time.

> As far as I know, none of the backends support having device code in one shared object call device code in another one. Is it more than that, like code in one shared object can’t dereference an address from another?

I don't know about more restrictions. I normally don't use `SYCL_EXTERNAL`/rdc much and I have only seen linker errors for this feature. Neither https://www.intel.com/content/www/us/en/developer/articles/technical/use-the-sycl-external-macro.html nor https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#subsec:syclexternal mention shared libraries in particular, but only talk about the need to use `SYCL_EXTERNAL` if device functions are used in a different translation unit.

dalg24 (approved):

Rombur (approved):

dalg24:
Retest this please

dalg24:
SYCL build did pass

PR #6532, opened 2023-10-19
SYCL: Restrict workaround for is_device_copyable to oneAPI versions before 2024.0.0

masterleinad:
It turns out that the workaround in https://github.com/kokkos/kokkos/pull/6009 (triggered by https://github.com/intel/llvm/pull/8195) was not as robust as expected since `Intel` decided in https://github.com/intel/llvm/pull/11388 to drop the second template parameter for `sycl::is_device_copyable` (in accordance with the SYCL2020 specs). I expect that commit to land in `oneAPI 2024.0.0` (or later) so we restrict the workaround to versions before that.

### DIFF
--- a/core/src/SYCL/Kokkos_SYCL_Instance.hpp
+++ b/core/src/SYCL/Kokkos_SYCL_Instance.hpp
@@ -330,8 +330,8 @@ struct sycl::is_device_copyable<
     Kokkos::Experimental::Impl::SYCLFunctionWrapper<Functor, Storage, false>>
     : std::true_type {};
 
-// FIXME_SYCL Remove when this specialization when specializations for
-// sycl::device_copyable also apply to const-qualified types.
+#if defined(KOKKOS_ENABLE_SYCL) && defined(__INTEL_LLVM_COMPILER) && \
+    __INTEL_LLVM_COMPILER < 20240000
 template <typename>
 struct NonTriviallyCopyableAndDeviceCopyable {
   NonTriviallyCopyableAndDeviceCopyable(
@@ -354,5 +354,6 @@ struct sycl::is_device_copyable<
     std::enable_if_t<!sycl::is_device_copyable_v<
         const NonTriviallyCopyableAndDeviceCopyable<Functor>>>>
     : std::true_type {};
+#endif
 #endif
 #endif
--- a/core/src/SYCL/Kokkos_SYCL_Instance.hpp
+++ b/core/src/SYCL/Kokkos_SYCL_Instance.hpp
@@ -330,8 +330,9 @@ struct sycl::is_device_copyable<
     Kokkos::Experimental::Impl::SYCLFunctionWrapper<Functor, Storage, false>>
     : std::true_type {};
 
-#if defined(KOKKOS_ENABLE_SYCL) && defined(__INTEL_LLVM_COMPILER) && \
-    __INTEL_LLVM_COMPILER < 20240000
+#if defined(KOKKOS_ENABLE_SYCL) &&                                           \
+    ((defined(__INTEL_LLVM_COMPILER) && __INTEL_LLVM_COMPILER < 20240000) || \
+     (defined(__LIBSYCL_MAJOR_VERSION) && __LIBSYCL_MAJOR_VERSION >= 7))
 template <typename>
 struct NonTriviallyCopyableAndDeviceCopyable {
   NonTriviallyCopyableAndDeviceCopyable(
--- a/core/src/SYCL/Kokkos_SYCL_Instance.hpp
+++ b/core/src/SYCL/Kokkos_SYCL_Instance.hpp
@@ -331,7 +331,7 @@ struct sycl::is_device_copyable<
     : std::true_type {};
 
 #if (defined(__INTEL_LLVM_COMPILER) && __INTEL_LLVM_COMPILER < 20240000) || \
-    (defined(__LIBSYCL_MAJOR_VERSION) && __LIBSYCL_MAJOR_VERSION >= 7)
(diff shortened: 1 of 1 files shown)


core/src/SYCL/Kokkos_SYCL_Instance.hpp
```diff
@@ -330,8 +330,8 @@ struct sycl::is_device_copyable<
     Kokkos::Experimental::Impl::SYCLFunctionWrapper<Functor, Storage, false>>
     : std::true_type {};
 
-// FIXME_SYCL Remove when this specialization when specializations for
-// sycl::device_copyable also apply to const-qualified types.
+#if defined(KOKKOS_ENABLE_SYCL) && defined(__INTEL_LLVM_COMPILER) && \
+    __INTEL_LLVM_COMPILER < 20240000
```
aelovikov-intel, line 334: Better macros to check would be `__LIBSYCL_MAJOR_VERSION` and `__LIBSYCL_MINOR_VERSION`. They could be used to distinguish between release versions and would work with current development trunk. Won't work for arbitrary commits between two adjacent releases though, but nothing would to the best of my knowledge.

masterleinad: How do I know which commit on `intel/llvm` corresponds to which `__LIBSYCL_MAJOR_VERSION/__LIBSYCL_MINOR_VERSION`?

aelovikov-intel: It's only meaningful for releases and the latest commit in intel/llvm, so I don't think the question is valid. I'd argue that there should be no expectations to support old commits of intel/llvm outside official releases (neither for us at intel/llvm nor for any library like kokkos).

masterleinad: Let me rephrase my questions: What is the mapping from compiler releases to `__LIBSYCL_MAJOR_VERSION/__LIBSYCL_MINOR_VERSION`? I am also having a hard time mapping `intel/llvm` releases to `oneAPI` releases (and the other way around).

aelovikov-intel: Current intel/llvm trunk: 7/1, will be the same for OneAPI 2024.1
OneAPI 2024.0 7/0

masterleinad: Hmmm... It's arguably easier to understand `__INTEL_LLVM_COMPILER < 20240000` than `_LIBSYCL_MAJOR_VERSION < 7`.

aelovikov-intel: I think the former is specific to OneAPI releases and not available for github.com/intel/llvm compilers.

masterleinad: I added a commit to also check for that macro.

core/src/SYCL/Kokkos_SYCL_Instance.hpp
```diff
@@ -356,3 +357,4 @@ struct sycl::is_device_copyable<
     : std::true_type {};
 #endif
 #endif
+#endif
```
dalg24, line 360: We don't want
```C++
template <typename Functor, typename Storage>
struct sycl::is_device_copyable<
    const Kokkos::Experimental::Impl::SYCLFunctionWrapper<Functor, Storage,
                                                          false>>
    : std::true_type {};
```
any more?

masterleinad: Correct. The whole point of this workaround was to figure out if a `const`-qualified type is implicitly device-copyable if the non-`const` type is since this behavior changed between `oneAPI` versions where the latest versions don't allow using `SFINAE` anymore and do the "right thing" anyway.

core/src/SYCL/Kokkos_SYCL_Instance.hpp
```diff
@@ -330,8 +330,9 @@ struct sycl::is_device_copyable<
     Kokkos::Experimental::Impl::SYCLFunctionWrapper<Functor, Storage, false>>
     : std::true_type {};
 
-// FIXME_SYCL Remove when this specialization when specializations for
-// sycl::device_copyable also apply to const-qualified types.
+#if defined(KOKKOS_ENABLE_SYCL) &&                                           \
```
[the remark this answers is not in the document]
masterleinad: Dropped in 52dd77e0fa4aef043a9ae18ff1edce686867df3b.

core/src/SYCL/Kokkos_SYCL_Instance.hpp
```diff
@@ -330,8 +330,8 @@ struct sycl::is_device_copyable<
     Kokkos::Experimental::Impl::SYCLFunctionWrapper<Functor, Storage, false>>
     : std::true_type {};
 
-// FIXME_SYCL Remove when this specialization when specializations for
-// sycl::device_copyable also apply to const-qualified types.
+#if (defined(__INTEL_LLVM_COMPILER) && __INTEL_LLVM_COMPILER < 20240000) || \
+    (defined(__LIBSYCL_MAJOR_VERSION) && __LIBSYCL_MAJOR_VERSION >= 7)
```
dalg24, line 334: Please confirm that this not meant to be if the SYCL runtime library version is *less than*

masterleinad: Ah, good catch. Fixed in 5c1094933c21e832636fc910be00fbae0909fdd0.

dalg24 (approved):
I don't know how to check the version numbers in the macro guard but I am fine with the general direction

cz4rs (approved):

PR #6534, opened 2023-10-24
SYCL: Implement DESUL_ATOMICS_ENABLE_SYCL_SEPARABLE_COMPILATION path

masterleinad:
Follow-up to https://github.com/desul/desul/pull/113 and https://github.com/kokkos/kokkos/pull/6531#discussion_r1366248056. Based on https://github.com/desul/desul/pull/113.
With `SYCL_EXT_ONEAPI_DEVICE_GLOBAL` being defined, we can also implement the `DESUL_ATOMICS_ENABLE_SYCL_SEPARABLE_COMPILATION` path analogously to the `Cuda` and `HIP` implementation. This now also seems to work fine when using shared libraries.

Corresponds to https://github.com/desul/desul/pull/114.


dalg24:
Please rebase and make sure it has the version that was merged into desul

masterleinad:
> Please rebase and make sure it has the version that was merged into desul

Done.

cmake/kokkos_arch.cmake
```diff
@@ -588,16 +588,18 @@ IF (KOKKOS_ENABLE_SYCL)
 ENDIF()
 
 # Check support for device_global variables
-# FIXME_SYCL Even if SYCL_EXT_ONEAPI_DEVICE_GLOBAL is defined, we still can't
-#            use device global variables with shared libraries
-IF(KOKKOS_ENABLE_SYCL AND NOT BUILD_SHARED_LIBS)
+# FIXME_SYCL If SYCL_EXT_ONEAPI_DEVICE_GLOBAL is defined, we can
+#            use device global variables with shared libraries using the
+#            non-separable compilation implementation. We get compile-time
+#            errors for the separable compilation implementation in either case.
```
dalg24, line 594:
```suggestion
# FIXME_SYCL If SYCL_EXT_ONEAPI_DEVICE_GLOBAL is not defined and we
#            are building shared libraries
#            use device global variables with shared libraries using the
#            non-separable compilation implementation. We get compile-time
#            errors for the separable compilation implementation in either case.
```

cmake/kokkos_arch.cmake
```diff
@@ -618,7 +620,7 @@ IF(KOKKOS_ENABLE_SYCL AND NOT BUILD_SHARED_LIBS)
 
     IF(KOKKOS_IMPL_SYCL_DEVICE_GLOBAL_SUPPORTED)
       COMPILER_SPECIFIC_FLAGS(
-        DEFAULT -fsycl-device-code-split=off -DDESUL_SYCL_DEVICE_GLOBAL_SUPPORTED
+        DEFAULT -fsycl-device-code-split=off -DDESUL_ATOMICS_ENABLE_SYCL_SEPARABLE_COMPILATION -DDESUL_SYCL_DEVICE_GLOBAL_SUPPORTED
```
dalg24, line 623: Comment why we enable RDC in Desul in that case

cmake/kokkos_arch.cmake
masterleinad: This suggestion might be (mis)interpreted as "When  `SYCL_EXT_ONEAPI_DEVICE_GLOBAL` is defined, shared libraries can be used." I'd prefer to be more specific and emphasize that we still need the non-separable implementation for shared libraries to be supported (which we need for functionalities in `KokkosKernels`) and this motivates the choice below.

dalg24: Feel free to rephrase.  I just don't understand the comment as it is now.

cmake/kokkos_arch.cmake
```diff
@@ -617,8 +621,9 @@ IF(KOKKOS_ENABLE_SYCL AND NOT BUILD_SHARED_LIBS)
       KOKKOS_IMPL_SYCL_DEVICE_GLOBAL_SUPPORTED)
 
     IF(KOKKOS_IMPL_SYCL_DEVICE_GLOBAL_SUPPORTED)
+      # Only the separable compilation implementation is supported.
       COMPILER_SPECIFIC_FLAGS(
-        DEFAULT -fsycl-device-code-split=off -DDESUL_SYCL_DEVICE_GLOBAL_SUPPORTED
+        DEFAULT -fsycl-device-code-split=off -DDESUL_ATOMICS_ENABLE_SYCL_SEPARABLE_COMPILATION -DDESUL_SYCL_DEVICE_GLOBAL_SUPPORTED
```
dalg24, line 626: How can we capture this setting within Kokkos source code?
(`#ifdef DESUL_ATOMICS_ENABLE_SYCL_SEPARABLE_COMPILATION` is not the answer, I will move these compile definitions to `<Kokkos_Atomics_Desul_Config.hpp>` where there belong in a follow up.)

Also is it not something you would want to see when the configuration is printed?

masterleinad: > How can we capture this setting within Kokkos source code?

We might want to define a corresponding Kokkos `CMake` option as we have for `Cuda` and `HIP`, see https://github.com/kokkos/kokkos/pull/5993, which enables some more optimizations on the compiler level but also has some downsides as we are using some low-level functions in `Kokkos` itself.
Without this option, I think it's fine just to configure `desul/atomics/Config.hpp` accordingly.

> Also is it not something you would want to see when the configuration is printed?

Not sure. My gut feeling is to document choices that are somewhat directly done by the user.

dalg24: > > How can we capture this setting within Kokkos source code?
> 
> We might want to define a corresponding Kokkos `CMake` option as we have for `Cuda` and `HIP`, see #5993, which enables some more optimizations on the compiler level but also has some downsides as we are using some low-level functions in `Kokkos` itself. Without this option, I think it's fine just to configure `desul/atomics/Config.hpp` accordingly.

Never mind I realize now we can do
```C++
#if defined(KOKKOS_IMPL_SYCL_DEVICE_GLOBAL_SUPPORTED) && defined(SYCL_EXT_ONEAPI_DEVICE_GLOBAL)
```
Let's not get into the `Kokkos_ENABLE_SYCL_RELOCATABLE_DEVICE_CODE` option discussion just yet

masterleinad: > Let's not get into the Kokkos_ENABLE_SYCL_RELOCATABLE_DEVICE_CODE option discussion just yet

I agree. 🙂

dalg24: > > Also is it not something you would want to see when the configuration is printed?
> 
> Not sure. My gut feeling is to document choices that are somewhat directly done by the user.

You are missing the point of `Kokkos::print_configuration`.  It is not so much for users it is more for developers to know for sure what was built.
And I expect that knowing whether `SYCL_EXT_ONEAPI_DEVICE_GLOBAL` is defined or not is useful information.

masterleinad:
We could do
```diff
diff --git a/cmake/kokkos_arch.cmake b/cmake/kokkos_arch.cmake
index 86ca0a315..2a7dfcb51 100644
--- a/cmake/kokkos_arch.cmake
+++ b/cmake/kokkos_arch.cmake
@@ -620,7 +620,7 @@ IF(KOKKOS_ENABLE_SYCL)
 
     IF(KOKKOS_IMPL_SYCL_DEVICE_GLOBAL_SUPPORTED)
       COMPILER_SPECIFIC_FLAGS(
-        DEFAULT -fsycl-device-code-split=off -DDESUL_ATOMICS_ENABLE_SYCL_SEPARABLE_COMPILATION -DDESUL_SYCL_DEVICE_GLOBAL_SUPPORTED
+        DEFAULT -fsycl-device-code-split=off -DDESUL_SYCL_DEVICE_GLOBAL_SUPPORTED
       )
     ENDIF()
   ENDIF()
diff --git a/core/src/CMakeLists.txt b/core/src/CMakeLists.txt
index 012af0a7d..a4edf1ba1 100644
--- a/core/src/CMakeLists.txt
+++ b/core/src/CMakeLists.txt
@@ -18,6 +18,9 @@ IF (NOT desul_FOUND)
   ENDIF()
   IF(KOKKOS_ENABLE_SYCL)
     SET(DESUL_ATOMICS_ENABLE_SYCL ON)
+    IF(KOKKOS_IMPL_SYCL_DEVICE_GLOBAL_SUPPORTED AND NOT KOKKOS_IMPL_HAVE_SYCL_EXT_ONEAPI_DEVICE_GLOBAL)
+      SET(DESUL_ATOMICS_ENABLE_SYCL_SEPARABLE_COMPILATION ON)
+    ENDIF()
   ENDIF()
   IF(KOKKOS_ENABLE_OPENMPTARGET)
     SET(DESUL_ATOMICS_ENABLE_OPENMP ON) # not a typo Kokkos OpenMPTarget -> Desul OpenMP
```
already.

dalg24:
Yes that looks more like it

dalg24:
Retest this please

dalg24:
Planning to ignore OpenMPTarget failure
```
48/48 Test #48: Kokkos_UnitTest_SIMD ........................Subprocess aborted***Exception:   3.20 sec
```
Everything else passed.

dalg24 (approved):

fnrizzi (approved):
not an expert on this, but looks good to me

PR #6536, opened 2023-10-24
Reduce number of `View` constructor instantiations

dalg24:
Templating the `View` constructors on the label type is a bad idea.  It yields a bunch of pointless instantiations.
```C++
Kokkos::View<int> v("v"); // -> instantiate for char[2]
Kokkos::View<int> w("MyLabel"); // -> instantiate for char[8]
Kokkos::View<int> x(std::string("MyLabel")); //  -> instantiate for std::string
```
even though the label is just converted to `std::string` when deferring to the `View` constructor that takes `Impl::ViewCtorProp`.

I suggest to take a const reference to `std::string` instead.
We could also take it by value and move.  Let me know if you want to this instead.

For reference, our current documentation says we have
```C++
View(const std::string &name, const IntType&... indices);
View(const std::string &name, const array_layout &layout);
```
https://kokkos.github.io/kokkos-core-wiki/API/core/view/view.html#_CPPv44ViewRKNSt6stringEDpRK7IntType
and
https://kokkos.github.io/kokkos-core-wiki/API/core/view/view.html#_CPPv44ViewRKNSt6stringERK12array_layout

This was prompted in https://github.com/kokkos/kokkos/pull/6477#discussion_r1352791393

### DIFF
--- a/core/src/Kokkos_View.hpp
+++ b/core/src/Kokkos_View.hpp
@@ -1489,26 +1489,20 @@ class View : public ViewTraits<DataType, Properties...> {
   }
 
   // Allocate with label and layout
-  template <typename Label>
-  explicit inline View(
-      const Label& arg_label,
-      std::enable_if_t<Kokkos::Impl::is_view_label<Label>::value,
-                       typename traits::array_layout> const& arg_layout)
+  explicit inline View(std::string const& arg_label,
+                       typename traits::array_layout const& arg_layout)
       : View(Impl::ViewCtorProp<std::string>(arg_label), arg_layout) {}
 
   // Allocate label and layout, must disambiguate from subview constructor.
-  template <typename Label>
-  explicit inline View(
-      const Label& arg_label,
-      std::enable_if_t<Kokkos::Impl::is_view_label<Label>::value, const size_t>
-          arg_N0          = KOKKOS_IMPL_CTOR_DEFAULT_ARG,
-      const size_t arg_N1 = KOKKOS_IMPL_CTOR_DEFAULT_ARG,
-      const size_t arg_N2 = KOKKOS_IMPL_CTOR_DEFAULT_ARG,
-      const size_t arg_N3 = KOKKOS_IMPL_CTOR_DEFAULT_ARG,
-      const size_t arg_N4 = KOKKOS_IMPL_CTOR_DEFAULT_ARG,
-      const size_t arg_N5 = KOKKOS_IMPL_CTOR_DEFAULT_ARG,
-      const size_t arg_N6 = KOKKOS_IMPL_CTOR_DEFAULT_ARG,
-      const size_t arg_N7 = KOKKOS_IMPL_CTOR_DEFAULT_ARG)
+  explicit inline View(std::string const& arg_label,
+                       const size_t arg_N0 = KOKKOS_IMPL_CTOR_DEFAULT_ARG,
+                       const size_t arg_N1 = KOKKOS_IMPL_CTOR_DEFAULT_ARG,
+                       const size_t arg_N2 = KOKKOS_IMPL_CTOR_DEFAULT_ARG,
+                       const size_t arg_N3 = KOKKOS_IMPL_CTOR_DEFAULT_ARG,
+                       const size_t arg_N4 = KOKKOS_IMPL_CTOR_DEFAULT_ARG,
+                       const size_t arg_N5 = KOKKOS_IMPL_CTOR_DEFAULT_ARG,
+                       const size_t arg_N6 = KOKKOS_IMPL_CTOR_DEFAULT_ARG,
+                       const size_t arg_N7 = KOKKOS_IMPL_CTOR_DEFAULT_ARG)
       : View(Impl::ViewCtorProp<std::string>(arg_label),
              typename traits::array_layout(arg_N0, arg_N1, arg_N2, arg_N3,
                                            arg_N4, arg_N5, arg_N6, arg_N7)) {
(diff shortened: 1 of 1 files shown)


masterleinad (commented):
I would prefer if we would do this change uniformly across all containers as well and get rid of `is_view_label`.

dalg24:
> I would prefer if we would do this change uniformly across all containers as well 

Sure

> and get rid of `is_view_label`.

This is more involved and I am not even sure it is possible.

masterleinad:
> This is more involved and I am not even sure it is possible.

Right, we would need to duplicate code in `ViewCtor.hpp` but that's the only file using it.

masterleinad (dismissed):

nmm0 (approved):
I like this, looks good to me.

So this will enable passing in `const char *` as the label which weirdly was not possible before.

I checked and this wouldn't suddenly make the following mean something different:

```cpp
auto v = Kokkos::View<double*>(0,5);
```

(It instantiates a view with no label and a nullptr. It's really gross and we probably shouldn't allow it, but at least this PR won't make it ambiguous)

nmm0 (changes requested):
Scratch my earlier comment, this means views of `char *` break -- it's shown in the CI. This fails because suddenly for char literals (not `const char *`) suddenly the view constructor that takes a pointer is a better match than `std::string`. You can use a string literal, but this would still break existing code. We could add a `const char *` constructor I suppose, but that would break pointer construction for views of `char *`.

I think the only way around this is to make a templated overload for string literals, i.e.:

```cpp
template<unsigned N>
View(const char (&label)[N], ...)
```

crtrott:
How about: 

```
diff --git a/core/src/Kokkos_View.hpp b/core/src/Kokkos_View.hpp
index bcbb28014..27896c3f1 100644
--- a/core/src/Kokkos_View.hpp
+++ b/core/src/Kokkos_View.hpp
@@ -1497,11 +1497,9 @@ class View : public ViewTraits<DataType, Properties...> {
       : View(Impl::ViewCtorProp<std::string>(arg_label), arg_layout) {}
 
   // Allocate label and layout, must disambiguate from subview constructor.
-  template <typename Label>
   explicit inline View(
-      const Label& arg_label,
-      std::enable_if_t<Kokkos::Impl::is_view_label<Label>::value, const size_t>
-          arg_N0          = KOKKOS_IMPL_CTOR_DEFAULT_ARG,
+      const std::string& arg_label,
+      const size_t arg_N0 = KOKKOS_IMPL_CTOR_DEFAULT_ARG,
       const size_t arg_N1 = KOKKOS_IMPL_CTOR_DEFAULT_ARG,
       const size_t arg_N2 = KOKKOS_IMPL_CTOR_DEFAULT_ARG,
       const size_t arg_N3 = KOKKOS_IMPL_CTOR_DEFAULT_ARG,
@@ -1565,8 +1563,10 @@ class View : public ViewTraits<DataType, Properties...> {
         arg_N0, arg_N1, arg_N2, arg_N3, arg_N4, arg_N5, arg_N6, arg_N7));
   }
 
+  template<class PointerType>
   explicit KOKKOS_INLINE_FUNCTION View(
-      pointer_type arg_ptr, const size_t arg_N0 = KOKKOS_IMPL_CTOR_DEFAULT_ARG,
+      PointerType arg_ptr,
+      std::enable_if_t<std::is_convertible_v<PointerType, pointer_type> ,const size_t> arg_N0 = KOKKOS_IMPL_CTOR_DEFAULT_ARG,
       const size_t arg_N1 = KOKKOS_IMPL_CTOR_DEFAULT_ARG,
       const size_t arg_N2 = KOKKOS_IMPL_CTOR_DEFAULT_ARG,
       const size_t arg_N3 = KOKKOS_IMPL_CTOR_DEFAULT_ARG,
```

crtrott:
https://godbolt.org/z/7qqb7qxda

So my proposed thing would make this guy now do something:

```c++
Kokkos::View<const char*> a("Test",4); 
```
This will create an unmanaged view of the string "Test". 
Before it would fail to compile because it would try to run the initialization functor, which can't compile because it assigns to the elements. 
Note that `const char*` behaved in a special way here, because:

```c++
const int values[5] = {1,2,3,4,5};
Kokkos::View<const int*> a(values, 5);
```
Would compile and give you an unmanaged view of the values.

Note that:
```c++
Kokkos::View<char*> a("TEST", 5);
```
will after this change still allocate, because `const char[]` is not convertible to `char*`.

dalg24:
> > I would prefer if we would do this change uniformly across all containers as well
> 
> Sure

I changed my mind and dropped the changes to `DynRankView` and `OffsetView`.
The `DynRankView` constructors update looks straightforward but it is more complicated for `OffsetView`
See https://github.com/kokkos/kokkos/blob/189aaa6da8620abe4b257e3eb1ad8832cba3f940/containers/src/Kokkos_OffsetView.hpp#L1053-L1079

These can be tackled in another PR if there is interest.

dalg24:
> So my proposed thing would make this guy now do something:
> 
> ```c++
> Kokkos::View<const char*> a("Test",4); 
> ```
> 
> This will create an unmanaged view of the string "Test". Before it would fail to compile because it would try to run the initialization functor, which can't compile because it assigns to the elements.

Actually that is not true.  The behavior does not change with this PR.  The code above does compile and it gives the somewhat surprising runtime unmanaged view wrapping the string literal.

So I suppose that's good.

core/src/Kokkos_View.hpp
```diff
@@ -1565,8 +1559,10 @@ class View : public ViewTraits<DataType, Properties...> {
         arg_N0, arg_N1, arg_N2, arg_N3, arg_N4, arg_N5, arg_N6, arg_N7));
   }
 
+  template <class PointerType, class = std::enable_if_t<std::is_convertible_v<
```
dalg24, line 1562: Should we do `std::is_same` instead?

nmm0:
> > So my proposed thing would make this guy now do something:
> > ```c++
> > Kokkos::View<const char*> a("Test",4); 
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > This will create an unmanaged view of the string "Test". Before it would fail to compile because it would try to run the initialization functor, which can't compile because it assigns to the elements.
> 
> Actually that is not true. The behavior does not change with this PR. The code above does compile and it gives the somewhat surprising runtime unmanaged view wrapping the string literal.
> 
> So I suppose that's good.

What's even more wild is that if you change that to

```cpp
Kokkos::View<const char*> a("Test",4); 
```

It creates a view with the label "Test"

nmm0 (approved):
Looks good to me! I believe the issue with char views should be resolved by this

core/src/Kokkos_View.hpp
nmm0: I think it needs to be convertible to work with e.g. non const pointer for a const view

dalg24: Yes Christian already pointed to me it was a silly suggestion

masterleinad (approved):
Fine with me but I don't think it helps much with the observations in #6477 (which isn't necessarily the point).

crtrott (approved):

PR #6537, opened 2023-10-25
Remove sleep and wake functions

Rombur:
This PR removes `sleep`, `wake`, and `execute_sleep` from the `Threads` backend. These functions are not used. The CUDA and SYCL backends declare `sleep` and `wake` but they don't implement them.

### DIFF
--- a/core/src/Cuda/Kokkos_Cuda.hpp
+++ b/core/src/Cuda/Kokkos_Cuda.hpp
@@ -137,26 +137,6 @@ class Cuda {
 #endif
   }
 
-  /** \brief  Set the device in a "sleep" state.
-   *
-   * This function sets the device in a "sleep" state in which it is
-   * not ready for work.  This may consume less resources than if the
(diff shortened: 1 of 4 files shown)


masterleinad:
Why is it OK to remove these functions from the `Threads` backend without deprecating them first?

dalg24:
> Why is it OK to remove these functions from the `Threads` backend without deprecating them first?

Because they are defined for Impl::ThreadsInternal.
The Cuda and SYCL member functions have no implementation.

dalg24 (approved):

masterleinad (approved):

masterleinad:
We should probably already have done it in https://github.com/kokkos/kokkos/commit/89247b6532137693926a187684976c2644d253b5 where we removed public member functions.

dalg24:
Ignoring Cuda+OpenMPTarget failures that are related to machine issues

PR #5127, opened 2022-06-17
Drop numa args in threads backend initialization

dalg24:
Re-opening #5121 that I accidentally closed

PR #6541, opened 2023-10-26
Cleanup ThreadsInternal

Rombur:
This PR does a few unrelated thing but it is pretty small that's why I didn't split it. The two most interesting changes in this PR are:

1. Cleanup of `ThreadsInternal::initialize`. This mainly consists in removing variables that have been unnecessary since https://github.com/kokkos/kokkos/pull/5127
2. Remove the `Sentinel` structure. This structure throws an error if we don't call `Kokkos::finalize` or call it in the wrong scope. It's pretty neat but it's the only backend that does this and I am not convinced that it's something that should be done at the backend level.

It's easier to review the PR if you hide the whitespace changes

### DIFF
--- a/core/src/Threads/Kokkos_Threads.hpp
+++ b/core/src/Threads/Kokkos_Threads.hpp
@@ -40,7 +40,6 @@ static_assert(false,
 
 namespace Kokkos {
 namespace Impl {
-class ThreadsInternal;
 enum class fence_is_static { yes, no };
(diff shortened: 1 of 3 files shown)


dalg24 (commented):
Most changes are trivial but I need more time to understand 7d31c22

crtrott (approved):

dalg24 (approved):

PR #6543, opened 2023-10-26
Fix infinity, quiet_NaN, signaling_Nan, isfinite, isnan, isinf for half_t and bhalf_t

### LINKED ISSUE #6476: Incorrect half_t and bhalf_t nan and infinity numeric trait values?
CC: @lucbv

These values are currently set to:

### half_t
https://github.com/kokkos/kokkos/blob/31c060ab505983ea08ca554604c89982c5f674a9/core/src/impl/Kokkos_Half_NumericTraits.hpp#L158-L170

https://github.com/kokkos/kokkos/blob/31c060ab505983ea08ca554604c89982c5f674a9/core/src/impl/Kokkos_Half_NumericTraits.hpp#L64-L74

### bhalf_t
https://github.com/kokkos/kokkos/blob/31c060ab505983ea08ca554604c89982c5f674a9/core/src/impl/Kokkos_Half_NumericTraits.hpp#L302-L307

https://github.com/kokkos/kokkos/blob/31c060ab505983ea08ca554604c89982c5f674a9/core/src/impl/Kokkos_Half_NumericTraits.hpp#L261-L271

Note that casting from float's values for nan and infinity to half_t and bhalf_t may give different values than what is listed above.

Related to https://github.com/lucbv/kokkos-kernels/pull/2.

masterleinad:
Fixes #6476.

This is a proposal for fixing the broken support for special values in half_t and bhalf_t. We basically introduce a bit comparison type that we use for `infinity`, `quite_NaN` and `signaling_NaN` and `bit_cast` the `impl_type` for comparison. This type is implicitly convertible to `Kokkos::Experimental::[b]half_t` but doesn't define any comparison operators or arithmetic operations on its own. 
Thus, `auto tmp = Kokkos::Experimental::infinity<half_t>::value` can be directly used to use the correct overload in `Kokkos::isinf(tmp)`. But comparisons such as `tmp==tmp` or operations like `tmp+=2` won't compile and the `tmp` must first be converted to `Kokkos::Experimental::[b]half_t`.
Note, that this limitation is a result of favoring `Kokkos::Experimental::infinity<half_t>::value` (and similar) to be `constexpr` over returning the correct type.

### DIFF
--- a/containers/unit_tests/CMakeLists.txt
+++ b/containers/unit_tests/CMakeLists.txt
@@ -46,6 +46,11 @@ foreach(Tag Threads;Serial;OpenMP;HPX;Cuda;HIP;SYCL)
      LIST(REMOVE_ITEM UnitTestSources ${dir}/TestCuda_DynViewAPI_generic.cpp)
     endif()
 
+    # FIXME_NVHPC: NVC++-S-0000-Internal compiler error. extractor: bad opc       0
+    if(KOKKOS_ENABLE_CUDA AND KOKKOS_CXX_COMPILER_ID STREQUAL NVHPC)
+     LIST(REMOVE_ITEM UnitTestSources ${dir}/TestCuda_WithoutInitializing.cpp)
+    endif()
+
     KOKKOS_ADD_EXECUTABLE_AND_TEST(ContainersUnitTest_${Tag} SOURCES ${UnitTestSources})
   endif()
 endforeach()
--- a/core/src/Kokkos_BitManipulation.hpp
+++ b/core/src/Kokkos_BitManipulation.hpp
@@ -115,7 +115,7 @@ bit_cast(From const& from) noexcept {
   return sycl::bit_cast<To>(from);
 #else
   To to;
-  memcpy(&to, &from, sizeof(To));
+  memcpy(static_cast<void*>(&to), static_cast<const void*>(&from), sizeof(To));
   return to;
 #endif
 }
--- a/core/src/impl/Kokkos_Half_FloatingPointWrapper.hpp
+++ b/core/src/impl/Kokkos_Half_FloatingPointWrapper.hpp
@@ -220,6 +220,36 @@ namespace Impl {
 template <typename FloatType>
 struct BitComparisonWrapper {
   std::uint16_t value;
+
+  template <typename Number>
+  KOKKOS_FUNCTION friend bool operator==(BitComparisonWrapper a, Number b) {
+    return static_cast<FloatType>(a) == b;
+  }
+
+  template <typename Number>
+  KOKKOS_FUNCTION friend bool operator!=(BitComparisonWrapper a, Number b) {
+    return static_cast<FloatType>(a) != b;
+  }
+
+  template <typename Number>
+  KOKKOS_FUNCTION friend bool operator<(BitComparisonWrapper a, Number b) {
+    return static_cast<FloatType>(a) < b;
+  }
+
+  template <typename Number>
+  KOKKOS_FUNCTION friend bool operator<=(BitComparisonWrapper a, Number b) {
+    return static_cast<FloatType>(a) <= b;
+  }
+
+  template <typename Number>
+  KOKKOS_FUNCTION friend bool operator>(BitComparisonWrapper a, Number b) {
+    return static_cast<FloatType>(a) > b;
+  }
+
+  template <typename Number>
+  KOKKOS_FUNCTION friend bool operator>=(BitComparisonWrapper a, Number b) {
+    return static_cast<FloatType>(a) >= b;
+  }
 };
 
 template <typename FloatType>
--- a/core/src/impl/Kokkos_Half_FloatingPointWrapper.hpp
+++ b/core/src/impl/Kokkos_Half_FloatingPointWrapper.hpp
@@ -223,32 +223,25 @@ struct BitComparisonWrapper {
 };
 
 template <typename FloatType>
-static constexpr BitComparisonWrapper<FloatType> exponent_mask;
-
-#ifdef KOKKOS_IMPL_HALF_TYPE_DEFINED
-template <>
-constexpr BitComparisonWrapper<Kokkos::Experimental::half_t>
-    exponent_mask<Kokkos::Experimental::half_t>{0b0'11111'0000000000};
-#endif
-
-#ifdef KOKKOS_IMPL_BHALF_TYPE_DEFINED
-template <>
-constexpr BitComparisonWrapper<Kokkos::Experimental::bhalf_t>
-    exponent_mask<Kokkos::Experimental::bhalf_t>{0b0'11111111'0000000};
-#endif
-
+inline constexpr BitComparisonWrapper<FloatType> exponent_mask;
 template <typename FloatType>
-constexpr BitComparisonWrapper<FloatType> fraction_mask;
+inline constexpr BitComparisonWrapper<FloatType> fraction_mask;
 
 #ifdef KOKKOS_IMPL_HALF_TYPE_DEFINED
 template <>
-constexpr BitComparisonWrapper<Kokkos::Experimental::half_t>
+inline constexpr BitComparisonWrapper<Kokkos::Experimental::half_t>
+    exponent_mask<Kokkos::Experimental::half_t>{0b0'11111'0000000000};
+template <>
+inline constexpr BitComparisonWrapper<Kokkos::Experimental::half_t>
     fraction_mask<Kokkos::Experimental::half_t>{0b0'00000'1111111111};
 #endif
 
 #ifdef KOKKOS_IMPL_BHALF_TYPE_DEFINED
 template <>
-constexpr BitComparisonWrapper<Kokkos::Experimental::bhalf_t>
+inline constexpr BitComparisonWrapper<Kokkos::Experimental::bhalf_t>
+    exponent_mask<Kokkos::Experimental::bhalf_t>{0b0'11111111'0000000};
+template <>
+inline constexpr BitComparisonWrapper<Kokkos::Experimental::bhalf_t>
     fraction_mask<Kokkos::Experimental::bhalf_t>{0b0'00000000'1111111};
 #endif
 
--- a/core/src/impl/Kokkos_Half_FloatingPointWrapper.hpp
+++ b/core/src/impl/Kokkos_Half_FloatingPointWrapper.hpp
@@ -222,11 +222,41 @@ struct BitComparisonWrapper {
   std::uint16_t value;
 };
 
+template <typename FloatType>
+static constexpr BitComparisonWrapper<FloatType> exponent_mask;
+
+#ifdef KOKKOS_IMPL_HALF_TYPE_DEFINED
+template <>
+static constexpr BitComparisonWrapper<Kokkos::Experimental::half_t>
+    exponent_mask<Kokkos::Experimental::half_t>{0b0'11111'0000000000};
+#endif
+
+#ifdef KOKKOS_IMPL_BHALF_TYPE_DEFINED
+template <>
+static constexpr BitComparisonWrapper<Kokkos::Experimental::bhalf_t>
+    exponent_mask<Kokkos::Experimental::bhalf_t>{0b0'11111111'0000000};
+#endif
+
+template <typename FloatType>
+static constexpr BitComparisonWrapper<FloatType> fraction_mask;
+
+#ifdef KOKKOS_IMPL_HALF_TYPE_DEFINED
+template <>
+static constexpr BitComparisonWrapper<Kokkos::Experimental::half_t>
+    fraction_mask<Kokkos::Experimental::half_t>{0b0'00000'1111111111};
+#endif
+
+#ifdef KOKKOS_IMPL_BHALF_TYPE_DEFINED
+template <>
+static constexpr BitComparisonWrapper<Kokkos::Experimwental::bhalf_t>
+    fraction_mask<Kokkos::Experimental::bhalf_t>{0b0'00000000'1111111};
+#endif
+
 template <class FloatType>
 class alignas(FloatType) floating_point_wrapper {
  public:
   using impl_type           = FloatType;
-  using bit_comparison_type = BitComparisonWrapper<FloatType>;
+  using bit_comparison_type = BitComparisonWrapper<floating_point_wrapper>;
 
  private:
   impl_type val;
--- a/core/src/impl/Kokkos_Half_FloatingPointWrapper.hpp
+++ b/core/src/impl/Kokkos_Half_FloatingPointWrapper.hpp
@@ -18,6 +18,7 @@
 #define KOKKOS_HALF_FLOATING_POINT_WRAPPER_HPP_
 
 #include <Kokkos_Macros.hpp>
+#include <Kokkos_BitManipulation.hpp>  // bit_cast
 
 #include <type_traits>
 #include <iosfwd>  // istream & ostream for extraction and insertion ops
@@ -215,10 +216,17 @@ cast_from_wrapper(const Kokkos::Experimental::bhalf_t& x);
 /************************** END forward declarations **************************/
 
 namespace Impl {
+
+template <typename FloatType>
+struct BitComparisonWrapper {
+  std::uint16_t value;
+};
+
 template <class FloatType>
 class alignas(FloatType) floating_point_wrapper {
  public:
-  using impl_type = FloatType;
+  using impl_type           = FloatType;
+  using bit_comparison_type = BitComparisonWrapper<FloatType>;
 
  private:
   impl_type val;
@@ -267,6 +275,11 @@ class alignas(FloatType) floating_point_wrapper {
     const fixed_width_integer_type rv_val = *rv_ptr;
     val       = reinterpret_cast<const impl_type&>(rv_val);
 #endif  // KOKKOS_HALF_IS_FULL_TYPE_ON_ARCH
+  }
+
+  KOKKOS_FUNCTION
+  floating_point_wrapper(bit_comparison_type rhs) {
+    val = Kokkos::bit_cast<impl_type>(rhs);
   }
 
   // Don't support implicit conversion back to impl_type.
--- a/core/src/impl/Kokkos_Half_FloatingPointWrapper.hpp
+++ b/core/src/impl/Kokkos_Half_FloatingPointWrapper.hpp
@@ -227,28 +227,28 @@ static constexpr BitComparisonWrapper<FloatType> exponent_mask;
 
 #ifdef KOKKOS_IMPL_HALF_TYPE_DEFINED
 template <>
-static constexpr BitComparisonWrapper<Kokkos::Experimental::half_t>
+constexpr BitComparisonWrapper<Kokkos::Experimental::half_t>
     exponent_mask<Kokkos::Experimental::half_t>{0b0'11111'0000000000};
 #endif
 
 #ifdef KOKKOS_IMPL_BHALF_TYPE_DEFINED
 template <>
-static constexpr BitComparisonWrapper<Kokkos::Experimental::bhalf_t>
+constexpr BitComparisonWrapper<Kokkos::Experimental::bhalf_t>
     exponent_mask<Kokkos::Experimental::bhalf_t>{0b0'11111111'0000000};
 #endif
 
 template <typename FloatType>
-static constexpr BitComparisonWrapper<FloatType> fraction_mask;
+constexpr BitComparisonWrapper<FloatType> fraction_mask;
 
 #ifdef KOKKOS_IMPL_HALF_TYPE_DEFINED
 template <>
-static constexpr BitComparisonWrapper<Kokkos::Experimental::half_t>
+constexpr BitComparisonWrapper<Kokkos::Experimental::half_t>
     fraction_mask<Kokkos::Experimental::half_t>{0b0'00000'1111111111};
 #endif
 
 #ifdef KOKKOS_IMPL_BHALF_TYPE_DEFINED
 template <>
-static constexpr BitComparisonWrapper<Kokkos::Experimwental::bhalf_t>
+constexpr BitComparisonWrapper<Kokkos::Experimental::bhalf_t>
     fraction_mask<Kokkos::Experimental::bhalf_t>{0b0'00000000'1111111};
 #endif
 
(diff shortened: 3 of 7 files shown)


core/src/impl/Kokkos_Half_NumericTraits.hpp
```diff
@@ -166,7 +166,7 @@ struct Kokkos::Experimental::Impl::norm_min_helper<
 template <>
 struct Kokkos::Experimental::Impl::quiet_NaN_helper<
     Kokkos::Experimental::half_t> {
-  static constexpr float value = 0xfc000;
+  static constexpr Kokkos::Experimental::half_t::bit_comparison_type value{0xfc00};
```
masterleinad, line 169: I verified that all these values have the correct bit pattern and fixed it in some cases.

core/src/impl/Kokkos_Half_NumericTraits.hpp
```diff
@@ -303,13 +303,13 @@ struct Kokkos::Experimental::Impl::norm_min_helper<
...
-  static constexpr float value = 0x7fc000;
+  static constexpr Kokkos::Experimental::bhalf_t::bit_comparison_type value{0xffc1};
 };
 // Signaling not a bhalf number
 template <>
 struct Kokkos::Experimental::Impl::signaling_NaN_helper<
     Kokkos::Experimental::bhalf_t> {
-  static constexpr float value = 0x7fe000;
+  static constexpr Kokkos::Experimental::bhalf_t::bit_comparison_type value{0xff81};
 };
```
masterleinad, line 313: See https://en.wikipedia.org/wiki/Bfloat16_floating-point_format#Examples.

core/src/impl/Kokkos_Half_MathematicalFunctions.hpp
```diff
@@ -155,10 +157,59 @@ KOKKOS_IMPL_MATH_HALF_FUNC_WRAPPER(KOKKOS_IMPL_MATH_BINARY_FUNCTION_HALF, nextaf
...
+
+#if defined(KOKKOS_BHALF_T_IS_FLOAT) && !KOKKOS_BHALF_T_IS_FLOAT
+KOKKOS_INLINE_FUNCTION bool isnan(Kokkos::Experimental::bhalf_t x) {
+  using bit_type = Kokkos::Experimental::half_t::bit_comparison_type;
+  bit_type bit_pattern_x         = bit_cast<bit_type>(static_cast<Kokkos::Experimental::bhalf_t::impl_type>(x));
+  bit_type bit_pattern_quiet_nan = Kokkos::Experimental::quiet_NaN_v<Kokkos::Experimental::bhalf_t>;
+  if (bit_pattern_x.value == bit_pattern_quiet_nan.value)
+    return true;
+  bit_type bit_pattern_signaling_nan = Kokkos::Experimental::signaling_NaN_v<Kokkos::Experimental::bhalf_t>;
+  return (bit_pattern_x.value == bit_pattern_signaling_nan.value);
```
masterleinad, line 199: We probably need to make a couple more values `NaN`, see https://github.com/intel/llvm/blob/ebd1d1fd296f07694f8dfe53bc7cd9b9906be5c5/sycl/include/sycl/ext/oneapi/experimental/bfloat16_math.hpp#L33-L39.

core/src/impl/Kokkos_Half_MathematicalFunctions.hpp
```diff
@@ -155,10 +157,59 @@ KOKKOS_IMPL_MATH_HALF_FUNC_WRAPPER(KOKKOS_IMPL_MATH_BINARY_FUNCTION_HALF, nextaf
...
+   return (bit_pattern_x.value == bit_pattern_inf.value);
+ }
+#endif
+
+#if defined(KOKKOS_BHALF_T_IS_FLOAT) && !KOKKOS_BHALF_T_IS_FLOAT
+KOKKOS_INLINE_FUNCTION bool isinf(Kokkos::Experimental::bhalf_t x) {
+   using bit_type = Kokkos::Experimental::half_t::bit_comparison_type;
+   bit_type bit_pattern_x   = bit_cast<bit_type>(static_cast<Kokkos::Experimental::bhalf_t::impl_type>(x));
+   bit_type bit_pattern_inf = Kokkos::Experimental::infinity_v<Kokkos::Experimental::bhalf_t>;
+   return (bit_pattern_x.value == bit_pattern_inf.value);
```
masterleinad, line 175: We also need to treat `-inf`.

core/unit_test/TestMathematicalFunctions.hpp
```diff
@@ -1581,21 +1581,15 @@ struct TestIsNaN {
...
     !defined(KOKKOS_ENABLE_HIP)  // FIXME_SYCL, FIXME_HIP
-    if (isnan(static_cast<KE::half_t>(2.f))
-#if !defined(KOKKOS_ENABLE_CUDA)
-        || !isnan(quiet_NaN<KE::half_t>::value) ||
-        !isnan(signaling_NaN<KE::half_t>::value)
-#endif
-    ) {
+    if (isnan(static_cast<KE::half_t>(2.f)) ||
+        !isnan(static_cast<KE::half_t>(quiet_NaN<KE::half_t>::value)) ||
+        !isnan(static_cast<KE::half_t>(signaling_NaN<KE::half_t>::value))) {
```
crtrott, line 1586: Does this work: 

```c++
KE::half_t val = quiet_NaN<KE::half_t>::value;
assert(val == quiet_NaN<KE::half_t>::value);
```

I would assume it does right?

dalg24: NaN does not compare equal to itself

e10harvey (commented):
Thank you for taking care of this, @masterleinad !

core/src/impl/Kokkos_Half_NumericTraits.hpp
dalg24, line 1: Not asking for change in this PR but would `finite_{min,max}` and `epsilon` also need to be expressed as uint16?

core/src/impl/Kokkos_Half_MathematicalFunctions.hpp
```diff
@@ -155,10 +156,77 @@ KOKKOS_IMPL_MATH_HALF_FUNC_WRAPPER(KOKKOS_IMPL_MATH_BINARY_FUNCTION_HALF, nextaf
...
 // fpclassify
-KOKKOS_IMPL_MATH_HALF_FUNC_WRAPPER(KOKKOS_IMPL_MATH_UNARY_PREDICATE_HALF, isfinite)
-KOKKOS_IMPL_MATH_HALF_FUNC_WRAPPER(KOKKOS_IMPL_MATH_UNARY_PREDICATE_HALF, isinf)
-#if !defined(KOKKOS_ENABLE_SYCL) && !defined(KOKKOS_ENABLE_HIP) // FIXME_SYCL, FIXME_HIP
-KOKKOS_IMPL_MATH_HALF_FUNC_WRAPPER(KOKKOS_IMPL_MATH_UNARY_PREDICATE_HALF, isnan)
+
+#if defined(KOKKOS_HALF_T_IS_FLOAT) && !KOKKOS_HALF_T_IS_FLOAT
+KOKKOS_INLINE_FUNCTION bool isfinite(Kokkos::Experimental::half_t x) {
+  using bit_type = Kokkos::Experimental::half_t::bit_comparison_type;
+  const bit_type exponent_mask{0b0'11111'0000000000};
```
dalg24, line 163: I would define these masks as variable templates (templated on the half type) next to the definitions of their respective half type.

core/unit_test/TestMathematicalFunctions.hpp
```diff
@@ -1290,12 +1290,12 @@ struct TestAbsoluteValueFunction {
     if (abs(static_cast<KE::half_t>(4.f)) != static_cast<KE::half_t>(4.f) ||
         abs(static_cast<KE::half_t>(-4.f)) != static_cast<KE::half_t>(4.f)) {
       ++e;
-      KOKKOS_IMPL_DO_NOT_USE_PRINTF("failed abs(KE::half_t)\n");
+      Kokkos::printf("failed abs(KE::half_t)\n");
```
dalg24, line 1293: I am generally fine with 22ea496674a2ac44d6eb77d519fa3b0b726a6688 but these changes do not seem to conflict with the rest of the PR and therefore probably should have been proposed in some other PR to avoid the noise.

core/unit_test/TestMathematicalFunctions.hpp
```diff
@@ -1554,9 +1552,163 @@ TEST(TEST_CATEGORY, mathematical_functions_ieee_remainder_function) {
...
+      ++e;
+      Kokkos::printf("failed isfinite(integral)\n");
+    }
+    if (!isfinite(2.f) || isfinite(quiet_NaN<float>::value) ||
+        isfinite(signaling_NaN<float>::value) ||
+        isfinite(infinity<float>::value)) {
+      ++e;
+      Kokkos::printf("failed isfinite(float)\n");
+    }
+    if (!isfinite(static_cast<KE::half_t>(2.f))
```
dalg24, line 1578: Why did you interleave half types between float an double?

core/unit_test/TestMathematicalFunctions.hpp
```diff
@@ -1554,9 +1552,163 @@ TEST(TEST_CATEGORY, mathematical_functions_ieee_remainder_function) {
...
+#ifndef KOKKOS_COMPILER_NVHPC  // FIXME_NVHPC 23.7
```
dalg24, line 1579: Fails to compile?

dalg24 (commented):
Overall I like the solution proposed

containers/unit_tests/CMakeLists.txt
```diff
@@ -46,6 +46,11 @@ foreach(Tag Threads;Serial;OpenMP;HPX;Cuda;HIP;SYCL)
      LIST(REMOVE_ITEM UnitTestSources ${dir}/TestCuda_DynViewAPI_generic.cpp)
     endif()
 
+    # FIXME_NVHPC: NVC++-S-0000-Internal compiler error. extractor: bad opc       0
+    if(KOKKOS_ENABLE_CUDA AND KOKKOS_CXX_COMPILER_ID STREQUAL NVHPC)
+     LIST(REMOVE_ITEM UnitTestSources ${dir}/TestCuda_WithoutInitializing.cpp)
```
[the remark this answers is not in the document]
masterleinad: Good question. I have no idea why it started failing in the CI. I couldn't reproduce it locally.

core/src/Kokkos_BitManipulation.hpp
```diff
@@ -115,7 +115,7 @@ bit_cast(From const& from) noexcept {
   return sycl::bit_cast<To>(from);
 #else
   To to;
-  memcpy(&to, &from, sizeof(To));
+  memcpy(static_cast<void*>(&to), static_cast<const void*>(&from), sizeof(To));
```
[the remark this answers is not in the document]
masterleinad: We got some `-Wclass-memaccess` warnings for the `impl_type` with `Cuda`. AFAICT, the SFINAE we have in place is sufficient and can ignore those warnings.

core/src/impl/Kokkos_Half_MathematicalFunctions.hpp
masterleinad: Yeah, I thought about it but couldn't quite decide one way or the other.

core/unit_test/TestMathematicalFunctions.hpp
masterleinad: Yes, this is the same as we have for the existing tests.

dalg24:
Do we have tests like?
1+inf==inf max<inf
nan!=nan nan!=0 nan!=inf
etc.

core/src/impl/Kokkos_Half_NumericTraits.hpp
masterleinad: I don't think they need to be expressed as `uint16` since they are representable as `float` as well.
Returning the `bit_conversion_type` has the advantage that its only implicitly convertible to `[b]half_t`. So that  the correct overload is selected for functions that accept all floating point types, like `isinf`, when called with these traits.

masterleinad:
> Do we have tests like?
> 1+inf==inf max<inf
> nan!=nan nan!=0 nan!=inf
> etc.

They are not yet enabled for `half_t` and `bhalf_t`. I looked into it yesterday and they seemed to work at least for `SYCL`. I wasn't quite sure if making sure that the comparison operators work as expected should be part of this pull request or rather a follow-up.

core/unit_test/TestMathematicalFunctions.hpp
masterleinad: I used the same order as in the existing tests.

dalg24: Understood

dalg24:
> > Do we have tests like?
> > 1+inf==inf max<inf
> > nan!=nan nan!=0 nan!=inf
> > etc.
> 
> They are not yet enabled for `half_t` and `bhalf_t`. I looked into it yesterday and they seemed to work at least for `SYCL`. I wasn't quite sure if making sure that the comparison operators work as expected should be part of this pull request or rather a follow-up.

I'm not sure.  How big of a change would that be?  It would be nice if we can fix it all here but don't necessarily want to postpone too much.

masterleinad:
> I'm not sure. How big of a change would that be? It would be nice if we can fix it all here but don't necessarily want to postpone too much.

I just enabled the tests we had for other arithmetic types. Let's see if it passes CI for all backends.

core/src/impl/Kokkos_Half_FloatingPointWrapper.hpp
```diff
@@ -215,10 +216,47 @@ cast_from_wrapper(const Kokkos::Experimental::bhalf_t& x);
...
 
 namespace Impl {
+
+template <typename FloatType>
+struct BitComparisonWrapper {
+  std::uint16_t value;
+};
+
+template <typename FloatType>
+static constexpr BitComparisonWrapper<FloatType> exponent_mask;
```
dalg24, line 226: Not sure if that will work.  You might need to do something like
```C++
template <class> struct HalfTypeBitMasksHelper {};
template <> struct HalfTypeBitMasksHelper<half_t> {
  using value_type = BitComparisonWrapper<half_t>;
  static constexpr value_type exponent_mask = 0b11...;
  static constexpr value_type fraction_mask = 0b00...;
};

template <class T>
inline constexpr exponent_mask = HalfTypeBitMasksHelper<T>::exponent_mask;
template <class T>
inline constexpr fraction_mask = HalfTypeBitMasksHelper<T>::fraction_mask;

File: core/unit_test/TestNumericTraits.hpp [hunk 13]
```diff
@@ -101,8 +101,8 @@ struct TestNumericTraits {
 
   KOKKOS_FUNCTION void operator()(Infinity, int, int& e) const {
     using Kokkos::Experimental::infinity;
-    auto const inf  = infinity<T>::value;
-    auto const zero = T(0);
+    T const inf  = infinity<T>::value;
```
dalg24, line 104: T -> auto here everywhere else below.
What was the issue with that code?

core/src/impl/Kokkos_Half_FloatingPointWrapper.hpp
dalg24, line 226: Declare both masks next to each other and specialize them both in one shot, it means less macro guards and will improve readability.

dalg24 (commented):
Minor comments (non blocking)
I see the inf=1+inf test but not the NaN comparison.  Are they anywhere?

core/unit_test/TestNumericTraits.hpp
masterleinad: `infinity` returns the `bit_comparison_type` type which doesn't have any comparisons or math functions defined.

masterleinad:
> I see the inf=1+inf test but not the NaN comparison. Are they anywhere?

We have `nan != nan` and `nan != zero`.

masterleinad:
A follow-up enabling all suitable traits tests and converting the traits to using the `bit_comparison_type` is at https://github.com/masterleinad/kokkos/tree/half_numeric_traits.

masterleinad:
```
NVC++-F-0000-Internal compiler error. process_acc_put_dinit: unexpected datatype       8  (/var/jenkins/workspace/Kokkos/build/core/unit_test/cuda/TestCuda_NumericTraits.cpp)
```

masterleinad:
`CUDA-11.0.3-Clang-Tidy` was failing in `defaultdevicetype.shared_space` and the `github-Linux / CI (ubuntu:latest , clang++, RelwithDebInfo, THREADS, ...` build showed multiple errors (restarted now) but all of that appears to be spurious/unrelated.

core/unit_test/TestNumericTraits.hpp
dalg24: Document that observable behavior change in the description of your PR

dalg24: I suppose that was already true when it was defined as float...

masterleinad: I added more details to the description.

core/src/impl/Kokkos_Half_MathematicalFunctions.hpp
```diff
@@ -155,10 +156,77 @@ KOKKOS_IMPL_MATH_HALF_FUNC_WRAPPER(KOKKOS_IMPL_MATH_BINARY_FUNCTION_HALF, nextaf
...
+
+#if defined(KOKKOS_BHALF_T_IS_FLOAT) && !KOKKOS_BHALF_T_IS_FLOAT
+KOKKOS_INLINE_FUNCTION bool isfinite(Kokkos::Experimental::bhalf_t x) {
+  using bit_type = Kokkos::Experimental::bhalf_t::bit_comparison_type;
+  constexpr bit_type exponent_mask = Kokkos::Experimental::Impl::exponent_mask<Kokkos::Experimental::bhalf_t>;
+  const bit_type bit_pattern_x = bit_cast<bit_type>(
+      static_cast<Kokkos::Experimental::bhalf_t::impl_type>(x));
+  return (bit_pattern_x.value & exponent_mask.value) != exponent_mask.value;
+}
+#endif
```
dalg24, line 178: We could avoid repeating ourselves with something along these lines
```C++
namespace Impl {
template <class Half>
KOKKOS_FUNCTION bool isfinite_helper(T x) {
  constexpr auto exponent_mask = Kokkos::Experimental::Impl::exponent_mask<Half>;
  auto const bit_pattern_x = Kokkos::bit_cast<decltype(exponent_mask)>(
      static_cast<Half>(x));
  return (bit_pattern_x.value & exponent_mask.value) != exponent_mask.value;
}
}  // namespace Impl

File: core/unit_test/TestNumericTraits.hpp (hunk 13 above)
Comment (dalg24) [reply core/unit_test/TestNumericTraits.hpp:104]: It is late so it might be a bad idea but did you consider defining the comparison for the bit comparison type that defers to the half type comparison?
Then we might not need these changes.

**Review by dalg24 — COMMENTED** [review]
I'm mostly ok with that patch.
BTW do we have tests with conversions to and from float or double for these singular values?  Not necessarily asking for them in this PR but would like to know how these behave.

File: core/unit_test/TestNumericTraits.hpp (hunk 13 above)
Comment (masterleinad) [reply core/unit_test/TestNumericTraits.hpp:104]: I would rather have the numeric traits return the correct type and return `static const` and switch to `static constexpr` when `bit_cast` is `constexpr`. If we can't construct a constexpr `[b]half_type`, there doesn't seem much of a benefit of having the returned implementation type be `constexpr`.

Comment (masterleinad) [reply core/unit_test/TestNumericTraits.hpp:104]: Turns out that doesn't work very well since we need `static` variables to be zero- or constant-initialized to be useable on the device.

Comment (masterleinad) [reply core/unit_test/TestNumericTraits.hpp]: dd1d3e708690523e6c73d4469bb7a483c9e47f24 introduces some comparison operators which seems to work for the contexts we test.

**Comment by masterleinad** [comment]
Only the `THREADS` CI is failing but we didn't touch anything for that here.

File: core/unit_test/TestNumericTraits.hpp [hunk 15]
```diff
@@ -147,8 +147,8 @@ struct TestNumericTraits {
   KOKKOS_FUNCTION void operator()(QuietNaN, int, int& e) const {
 #ifndef KOKKOS_COMPILER_NVHPC  // FIXME_NVHPC 23.7 nan
     using Kokkos::Experimental::quiet_NaN;
-    constexpr auto nan  = quiet_NaN<T>::value;
-    constexpr auto zero = T(0);
+    auto const nan  = quiet_NaN<T>::value;
```
dalg24, line 150: Remind me why it cannot be constexpr?

masterleinad: Ah, oversight, fixed in 9d840255f84a8ba5e035f957fd8d4c855b8b7ced.

fnrizzi: so this was reverted to const in https://github.com/kokkos/kokkos/pull/6543/commits/4a81ff0d107c768080899e866f8fdaedbb49afbc right?

core/unit_test/TestNumericTraits.hpp
```diff
@@ -387,6 +391,14 @@ TEST(TEST_CATEGORY, numeric_traits_min_max_exponent10) {
 #endif
 }
 TEST(TEST_CATEGORY, numeric_traits_quiet_and_signaling_nan) {
+#ifndef KOKKOS_COMPILER_NVHPC  // FIXME_NVHPC 23.7
```
fnrizzi, line 394: what happens with nvhpc 23.7?

core/unit_test/TestNumericTraits.hpp
masterleinad: `zero` can't be `constexpr` for `half_t` but `nan` can (and is).

masterleinad:
Only `github-Linux / CI (ubuntu:latest, g++, RelWithDebInfo, THREADS) (pull_request)` is failing (which we see on almost all pull requests now).

core/src/impl/Kokkos_Half_FloatingPointWrapper.hpp
```diff
@@ -215,10 +216,70 @@ cast_from_wrapper(const Kokkos::Experimental::bhalf_t& x);
 /************************** END forward declarations **************************/
 
 namespace Impl {
+
+template <typename FloatType>
+struct BitComparisonWrapper {
```
fnrizzi, line 221: would it make sense to add a staitic assert here to ensure the floatype is actually a float type?

fnrizzi (approved):
left one comment but it is kind of minor - thanks Daniel!

core/src/impl/Kokkos_Half_FloatingPointWrapper.hpp
masterleinad: It's mostly templated so that they get different types for `half_t` and `bhalf_t`. All contexts where it's actually used use `bit_cast` which checks that the size of the types (this one and `[b]half_t::impl_type`) match. I could add a check in `FloatingPointWrapper` (which `half_t` and `bhalf_t` are aliases from) that the contained `impl_type` really has 16 bits but I don't think adding any checks here would be very useful.

ndellingwood:
I tested a Trilinos build with this PR, kokkos-kernels@release-candidate-4.2.00 and enabling the KokkosKernels, Belos, and Ifpack2 tests and did not see any new regressions compared with Trilinos develop branch

dalg24 (approved):

PR #6545, opened 2023-10-27
OpenMPTarget: CI compiler upgrade.

rgayatri23:
The PR updates CI to the latest compiler version of clang/17.0.3.
It also enables a few of the blocked unit tests.


dalg24 (commented):
This is fine by me but I don't understand your strategy.  The fact these are now passing with the latest does not mean they'd pass with 16.0

rgayatri23:
> This is fine by me but I don't understand your strategy. The fact these are now passing with the latest does not mean they'd pass with 16.0

Frankly I am bit confused myself on how to organize it. The set of unit tests that pass with various versions.subversions from 16.x to 17.x is neither the same nor is it completely disjoint. I also noticed that the `RelWithDebugInfo` build (which is in our CI) passes on V100 but crashes on A100 (only `Release` build passes on A100). So right now I went with the easiest way of making sure that at-least the Kokkos CI passes.

masterleinad (commented):
I think we should decide which tests to guard for when deciding on the minimum compiler version. If we want to use something higher in our CI, we might need extra checks.

rgayatri23:
> I think we should decide which tests to guard for when deciding on the minimum compiler version. If we want to use something higher in our CI, we might need extra checks.

It would be nice to use at-least 17.x to allow for the usage of scratch memory extensions in clang.
We can start with 15 and block only for 16.0 and 17.subversions to limit the bloating of guards if needed.

rgayatri23:
I have created an issue #6554 which lists the unit tests that do not work with the various versions of `clang` compiler. 
In the current PR, I have only blocked unit tests for version 17.0.3 which is what our CI checks.

core/unit_test/CMakeLists.txt
```diff
@@ -372,20 +372,15 @@ if(Kokkos_ENABLE_OPENMPTARGET)
...
     ${CMAKE_CURRENT_BINARY_DIR}/openmptarget/TestOpenMPTarget_TeamReductionScan.cpp
     ${CMAKE_CURRENT_BINARY_DIR}/openmptarget/TestOpenMPTarget_WorkGraph.cpp
-    IF (KOKKOS_CXX_COMPILER_ID STREQUAL "Clang" AND KOKKOS_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 16.0.0)
-        ${CMAKE_CURRENT_BINARY_DIR}/openmptarget/TestOpenMPTarget_SubView_c01.cpp
-        ${CMAKE_CURRENT_BINARY_DIR}/openmptarget/TestOpenMPTarget_SubView_c02.cpp
-        ${CMAKE_CURRENT_BINARY_DIR}/openmptarget/TestOpenMPTarget_SubView_c03.cpp
-        ${CMAKE_CURRENT_BINARY_DIR}/openmptarget/TestOpenMPTarget_Reducers_d.cpp
-    endif()
-    IF (KOKKOS_CXX_COMPILER_ID STREQUAL "Clang" AND KOKKOS_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 16.0.0)
+    IF (KOKKOS_CXX_COMPILER_ID STREQUAL "Clang" AND KOKKOS_CXX_COMPILER_VERSION VERSION_EQUAL 17.0.3)
```
masterleinad, line 375:
```suggestion
    IF (KOKKOS_CXX_COMPILER_ID STREQUAL "Clang" AND KOKKOS_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 17.0.3)
```
We can explicitly disable if `clang` manages to have regressions in future versions.

dalg24 (approved):
I wish you proposed 9932d4c5d6326898e42e2f3d350de4730046638d as its own PR but sure

core/unit_test/CMakeLists.txt
```diff
@@ -372,20 +372,15 @@ if(Kokkos_ENABLE_OPENMPTARGET)
...
     ${CMAKE_CURRENT_BINARY_DIR}/openmptarget/TestOpenMPTarget_TeamReductionScan.cpp
     ${CMAKE_CURRENT_BINARY_DIR}/openmptarget/TestOpenMPTarget_WorkGraph.cpp
-    IF (KOKKOS_CXX_COMPILER_ID STREQUAL "Clang" AND KOKKOS_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 16.0.0)
-        ${CMAKE_CURRENT_BINARY_DIR}/openmptarget/TestOpenMPTarget_SubView_c01.cpp
-        ${CMAKE_CURRENT_BINARY_DIR}/openmptarget/TestOpenMPTarget_SubView_c02.cpp
-        ${CMAKE_CURRENT_BINARY_DIR}/openmptarget/TestOpenMPTarget_SubView_c03.cpp
-        ${CMAKE_CURRENT_BINARY_DIR}/openmptarget/TestOpenMPTarget_Reducers_d.cpp
-    endif()
-    IF (KOKKOS_CXX_COMPILER_ID STREQUAL "Clang" AND KOKKOS_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 16.0.0)
+    IF (KOKKOS_CXX_COMPILER_ID STREQUAL "Clang" AND KOKKOS_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 17.0.3)
```
masterleinad, line 375: These work for Clang 17.0.3 and later, right? In that case you should remove the test for earlier versions
```suggestion
    IF (KOKKOS_CXX_COMPILER_ID STREQUAL "Clang" AND KOKKOS_CXX_COMPILER_VERSION VERSION_LESS 17.0.3)
```

rgayatri23: No these are the ones that don't work even with 17.0.3, the ones that work have been removed from the list already.

masterleinad: OK. So these work with earlier versions and that's why you are guarding them this way?

rgayatri23: Are the changes clear now?

masterleinad (approved):

core/unit_test/CMakeLists.txt
masterleinad: Yes, thank you!

dalg24:
Ignoring HIP timeout

masterleinad:
`HIP-ROCm-5.6-C++20` timed out. We see the usual `github-Linux` THREADS failures.

Issue #6546, opened 2023-10-27
HPX CI builds are failings


dalg24:
Build logs all fail with
```
In file included from /home/runner/work/kokkos/kokkos/hpx/libs/core/thread_pool_util/src/thread_pool_suspension_helpers.cpp:7:
In file included from /home/runner/work/kokkos/kokkos/hpx/libs/core/async_local/include/hpx/async_local/apply.hpp:11:
In file included from /home/runner/work/kokkos/kokkos/hpx/libs/core/executors/include/hpx/executors/apply.hpp:11:
In file included from /home/runner/work/kokkos/kokkos/hpx/libs/core/execution/include/hpx/execution/executors/execution.hpp:14:
In file included from /home/runner/work/kokkos/kokkos/hpx/libs/core/execution_base/include/hpx/execution_base/execution.hpp:14:
In file included from /home/runner/work/kokkos/kokkos/hpx/libs/core/iterator_support/include/hpx/iterator_support/counting_shape.hpp:10:
/home/runner/work/kokkos/kokkos/hpx/libs/core/iterator_support/include/hpx/iterator_support/counting_iterator.hpp:58:53: error: no member named 'intmax_t' in namespace 'std'
                    (sizeof(Integer) >= sizeof(std::intmax_t)),
                                               ~~~~~^
/home/runner/work/kokkos/kokkos/hpx/libs/core/iterator_support/include/hpx/iterator_support/counting_iterator.hpp:61:39: error: expected ';' after alias declaration
                        std::intmax_t>,
                                      ^
                                      ;
/home/runner/work/kokkos/kokkos/hpx/libs/core/iterator_support/include/hpx/iterator_support/counting_iterator.hpp:64:46: error: no member named 'intmax_t' in namespace 'std'
                        std::ptrdiff_t, std::intmax_t>>::type::type;
                                        ~~~~~^
/home/runner/work/kokkos/kokkos/hpx/libs/core/iterator_support/include/hpx/iterator_support/counting_iterator.hpp:64:55: error: expected member name or ';' after declaration specifiers
                        std::ptrdiff_t, std::intmax_t>>::type::type;
~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
4 errors generated.
```
presumably due to an update of the testing environment since HPX version is pinned.  I supposed HPX was accidentally relying on a transitive include of `<cstdint>`.

cc @msimberg

[...]

PR #6547, opened 2023-10-27
Bump HPX version used in CI to 1.9.0

msimberg:
Avoids ("fixes") https://github.com/kokkos/kokkos/issues/6546.

HPX is unlikely to get a patch release for < 1.9.0 and users can easily enough work around the problem if they would like to keep using HPX < 1.9.0. I don't think it's worth a workaround in Kokkos for this.

### DIFF
--- a/.github/workflows/continuous-integration-workflow-hpx.yml
+++ b/.github/workflows/continuous-integration-workflow-hpx.yml
@@ -36,7 +36,7 @@ jobs:
         uses: actions/checkout@v3
         with:
           repository: STELLAR-GROUP/hpx
-          ref: 1.8.0
+          ref: v1.9.0
           path: hpx
       - uses: actions/cache@v3
         id:   cache-hpx


masterleinad (approved):
Fine with me.

dalg24:
Ignoring unrelated failure
```
4: /var/jenkins/workspace/Kokkos/core/unit_test/TestExecSpacePartitioning.hpp:105: Failure
4:   sum1
4:     Which is: 7642095
4:   sum2
4:     Which is: 0
4: [  FAILED  ] openmptarget.partitioning_by_args (0 ms)

dalg24:
Please confirm that we only want to bump the version used in the CI without changing the minimum requirement
https://github.com/kokkos/kokkos/blob/189aaa6da8620abe4b257e3eb1ad8832cba3f940/cmake/Modules/FindTPLHPX.cmake#L2

msimberg:
> Please confirm that we only want to bump the version used in the CI without changing the minimum requirement

Yep, confirm.

dalg24 (approved):

PR #6553, opened 2023-10-30
Only define STDALGO_TEAM_SOURCES_* once

masterleinad:
These variables actually don't use anything from the `foreach` loop over all backends and we just define the corresponding variables redundantly as many times as we have backends enabled. This pull request fixes this confusing behavior by pulling the relevant code out of the loops.


fnrizzi:
failure is unrelated and ROCm timed out

fnrizzi (approved):

cwpearson (approved):

Issue #5609, opened 2022-11-01
Implementability of logical spaces


dalg24:
Logical spaces are our attempt to mock GPU to enable debugging from a host-only build.

As we rediscovered in #5544 (adding a `static_assert` to check that the view memory space is accessible from the execution space that does the initialization) we currently only have logical memory spaces.
Logical memory spaces are implemented as a strongly-typed wrapper around another space which changes accessibility and has its own separate allocation tracking facility.  We don't have logical execution spaces from which we can legally access logical memory spaces so we can't do much with them.

The goal of this issue is to determine whether it is possible to make them useful.  If so what are we missing?  Otherwise should we withdraw the whole thing?

Here are previous work on the subject that I was able to dig up with a quick search:
* Issue #783 first discussion of a debug facility that would be available from a serial build 
* PR #2229 reviving the original discussion and stating 
* PR #2307 prototype allocation in the debug space with separate tracking mechanism
* PR #2822 attempt to pass the meta information from the policy to the tools via `beginParallel{For,Reduce,Scan)`
* PR #2918 draft of the logical memory space that is currently in `develop`
* PR #3546 version finally merged into release 3.4

dalg24:
This snippet of code
https://github.com/kokkos/kokkos/blob/3c842fb0a46df51dd6bcd012316a9171993e7a05/core/src/impl/Kokkos_ViewMapping.hpp#L4124-L4135
illustrates why I currently think it cannot work.  This is how we determine memory access violations in views.  In short, we cannot infer from the kernel in what context it is being called, that is we cannot distinguish a logical execution space from its underlying host execution space.

masterleinad:
I could see that the class would be more usable with something like
```diff
diff --git a/core/src/Kokkos_LogicalSpaces.hpp b/core/src/Kokkos_LogicalSpaces.hpp
index 2f5aa2985..30d73c1e4 100644
--- a/core/src/Kokkos_LogicalSpaces.hpp
+++ b/core/src/Kokkos_LogicalSpaces.hpp
@@ -71,6 +71,14 @@ struct LogicalSpaceSharesAccess {
   struct no_shared_access {};
 };
 
+template <class BaseExecutionSpace, class MemorySpace>
+class LogicalExecutionSpace : public BaseExecutionSpace {
+ public:
+  using BaseExecutionSpace::BaseExecutionSpace;
+  using execution_space = LogicalExecutionSpace;
+  using memory_space    = MemorySpace;
+};
+
[...]

PR #6555, opened 2023-10-31
Relax accessibility mandate in `View` constructor

dalg24:
Addressing https://github.com/kokkos/kokkos/issues/5609#issuecomment-1490777893

As Phil pointed out, it makes little sense to enforce accessibility when the user explicitly requests allocation with no initialization.

Consider the code below that Phil had suggested in his original comment
```C++
struct TestSpaceNamer {
  static constexpr const char* get_name() { return "TestSpace"; }
};

using fake_memory_space = Kokkos::Experimental::LogicalMemorySpace<
    Kokkos::HostSpace, Kokkos::DefaultHostExecutionSpace, TestSpaceNamer,
    Kokkos::Experimental::LogicalSpaceSharesAccess::no_shared_access>;

TEST(defaultdevicetypeKokkosViewContentsTest, test_logical_device_view_contents) {
  // Create an inaccessible View
  using LogicalViewType = Kokkos::View<int*, fake_memory_space>;
  auto lv = LogicalViewType(Kokkos::view_alloc("lv", Kokkos::WithoutInitializing), 1); // static_assert fails

  // Initialize a value
  Kokkos::deep_copy(lv, 3);
}
```
With this the change proposed here the code will now compile.

You may ask why I did not add the code snippet as a unit test.  The answer is that I actually plan to propose removing `LogicalSpace` as soon as I am done opening this PR.  I think that nevertheless Phil was right, that the static assertion was too strict, and it is worth fixing.

### DIFF
--- a/containers/src/Kokkos_DynRankView.hpp
+++ b/containers/src/Kokkos_DynRankView.hpp
@@ -1096,7 +1096,7 @@ class DynRankView : public ViewTraits<DataType, Properties...> {
         prop_copy,
         Impl::DynRankDimTraits<typename traits::specialize>::
             template createLayout<traits, P...>(arg_prop, arg_layout),
-        Impl::ViewCtorProp<P...>::has_execution_space);
+        std::bool_constant<Impl::ViewCtorProp<P...>::has_execution_space>());
 
     // Setup and initialization complete, start tracking
     m_track.assign_allocated_record_to_uninitialized(record);
--- a/containers/src/Kokkos_OffsetView.hpp
+++ b/containers/src/Kokkos_OffsetView.hpp
@@ -1196,7 +1196,8 @@ class OffsetView : public ViewTraits<DataType, Properties...> {
 
     Kokkos::Impl::SharedAllocationRecord<>* record = m_map.allocate_shared(
         prop_copy, arg_layout,
-        Kokkos::Impl::ViewCtorProp<P...>::has_execution_space);
+        std::bool_constant<
+            Kokkos::Impl::ViewCtorProp<P...>::has_execution_space>());
 
     // Setup and initialization complete, start tracking
     m_track.assign_allocated_record_to_uninitialized(record);
--- a/core/src/Kokkos_View.hpp
+++ b/core/src/Kokkos_View.hpp
@@ -1419,7 +1419,8 @@ class View : public ViewTraits<DataType, Properties...> {
         i4, i5, i6, i7, alloc_name);
 
     Kokkos::Impl::SharedAllocationRecord<>* record = m_map.allocate_shared(
-        prop_copy, arg_layout, Impl::ViewCtorProp<P...>::has_execution_space);
+        prop_copy, arg_layout,
+        std::bool_constant<Impl::ViewCtorProp<P...>::has_execution_space>());
 
     // Setup and initialization complete, start tracking
     m_track.m_tracker.assign_allocated_record_to_uninitialized(record);
--- a/core/src/impl/Kokkos_ViewArray.hpp
+++ b/core/src/impl/Kokkos_ViewArray.hpp
@@ -319,18 +319,19 @@ class ViewMapping<Traits, Kokkos::Array<>> {
 
   //----------------------------------------
 
-  template <class... P>
+  template <class... P, bool B>
   Kokkos::Impl::SharedAllocationRecord<> *allocate_shared(
       Kokkos::Impl::ViewCtorProp<P...> const &arg_prop,
       typename Traits::array_layout const &arg_layout,
-      bool execution_space_specified) {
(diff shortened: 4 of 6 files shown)


PhilMiller (approved):

PhilMiller:
The logical space stuff was actually useful for testing code in the DARMA serialization library Magistrate. Specifically, we wanted to be able to write tests that we could instantiate code that would work for CudaSpace view instances and never access them other than via `deep_copy`, without needing a Cuda compiler to build and run the tests. So, consider this a request not to go ahead with removing that.

masterleinad (commented):
Given recent discussions on slack, I think it makes sense to relax this for `WithoutInitializing`. I would move it down to line 3450, though.

masterleinad:
For reference, we initially decided in https://github.com/kokkos/kokkos/pull/5544 to the guard this way.

masterleinad (dismissed):
> You may ask why I did not add the code snippet as a unit test. The answer is that I actually plan to propose removing `LogicalSpace` as soon as I am done opening this PR. I think that nevertheless Phil was right, that the static assertion was too strict, and it is worth fixing.

We should still add a test that would have errored out before.

dalg24:
> Assuming that `LogicalSpace` is in discussion for deprecation/removal, I would prefer a test that doesn't rely on it so we wouldn't remove the test with `LogicalSpace`.

Can you think of another way of testing it?

masterleinad:
> Can you think of another way of testing it?

You can do
```
Kokkos::View<int*, Kokkos::DefaultExecutionSpace::memory_space> v(Kokkos::view_alloc("v", Kokkos::WithoutInitializing, Kokkos::DefaultHostExecutionSpace{}), 1);
```
somewhere which would have failed before whenever a `GPU` backend is enabled.

dalg24:
> You can do
> 
> ```
> Kokkos::View<int*, Kokkos::DefaultExecutionSpace::memory_space> v(Kokkos::view_alloc("v", Kokkos::WithoutInitializing, Kokkos::DefaultHostExecutionSpace{}), 1);
> ```
> 
> somewhere which would have failed before whenever a `GPU` backend is enabled.

This is not valid code IMO.  What does the execution space argument mean with respect to allocation if it is not related to the memory space?

masterleinad:
> This is not valid code IMO. What does the execution space argument mean with respect to allocation if it is not related to the memory space?

The discussion in https://kokkosteam.slack.com/archives/C5BGU5NDQ/p1698104109094269 (which I assumed triggered this pull request) discusses using an execution space instance for `create_mirror_view_and_copy` that might not be compatible with the memory space for allocating but only for the `deep_copy`. In this case, it might be desirable to allow doing that even when using a different execution space instance for the allocation than provided since we are still fencing the one we are actually using. In general, the intent is to use the execution space instance to allocate within the correct context and possibly avoid device synchronization/`Kokkos::fence`. Thus, this becomes important for multi-GPU support.

dalg24:
> The discussion in https://kokkosteam.slack.com/archives/C5BGU5NDQ/p1698104109094269 (which I assumed triggered this pull request)

No.  See the description.
Also not sure why you bring up `create_mirror_view_and_copy` here.  The code you posted above is not something that I think we should support.

crtrott:
Thinking about this deeper I think I am very uncomfortable with having a memory_space where memory_space::execution_space can't access memory_space, or the other way around having an execution_space which can't access execution_space::memory_space.

Possibly we could have NullExecSpace which is marked for accessibility to something but can't ever be used to compile a kernel.

masterleinad:
> No. See the description.

I'm struggling to understand how the example would look without using LogicalSpaces.

> Also not sure why you bring up `create_mirror_view_and_copy` here. The code you posted above is not something that I think we should support.

AFAICT, the example I gave is exactly what this pull request would allow to do and I'm on the brink if we should allow it or not.

dalg24:
> > No. See the description.
> 
> I'm struggling to understand how the example would look without using LogicalSpaces.

There would be no test without LogicalSpace

> > Also not sure why you bring up `create_mirror_view_and_copy` here. The code you posted above is not something that I think we should support.
> 
> AFAICT, the example I gave is exactly what this pull request would allow to do and I'm on the brink if we should allow it or not.

I am leaning towards withdrawing this PR.

Rombur:
What's the use case of this PR outside of LogicalSpace? I understand that conceptually we don't need to check for accessibility but I find it strange to allow to use an execution space that does not have access to the memory space.

dalg24:
> What's the use case of this PR outside of LogicalSpace? I understand that conceptually we don't need to check for accessibility but I find it strange to allow to use an execution space that does not have access to the memory space.

At the moment nothing other than theoretical.

crtrott:
Not consensus to do this.

PR #6557, opened 2023-10-31
Remove `Experimental::LogicalMemorySpace`

dalg24:
Proposed resolution for #5609
In short, coming to the conclusion that `LogicalMemorySpace` cannot achieve its design goal (that is a debugging facility for accessibility issues on systems without a discrete GPU) and suggesting to just remove it.
The code sits in the `Experimental::` namespace.  I am not aware of any serious use of this facility.

I did a [quick search on GH](https://github.com/search?q=LogicalMemorySpace&type=code) and AFAICT the only code we'd break is kokkos-remote-spaces but we could fix it ourselves.  It looks like Kokkos-resillience would not be affected by that change.  @janciesko @nmm0 

Let me know if you think I am too aggressive and that we should consider going through deprecation.


dalg24:
@PhilMiller voiced an objection to this in https://github.com/kokkos/kokkos/pull/6555#issuecomment-1786442085
I expect he meant to comment here so I am pasting the comment below

> The logical space stuff was actually useful for testing code in the DARMA serialization library Magistrate. Specifically, we wanted to be able to write tests that we could instantiate code that would work for CudaSpace view instances and never access them other than via deep_copy, without needing a Cuda compiler to build and run the tests. So, consider this a request not to go ahead with removing that.

PhilMiller:
Thanks for copying that over - I was on my phone and hadn't seen this PR put up yet.

Magistrate is here: https://github.com/DARMA-tasking/magistrate/
The test code fragment that you quoted is actually present in https://github.com/DARMA-tasking/magistrate/blob/develop/tests/unit/test_kokkos_serialize_special.cc

I realize it's disabled at the moment, but as I understand (having stepped away from the project) there are broader design changes being done in the stuff that it would be testing. It should still be applicable. One of the longer-standing NGA folks should be able to follow up on that point.

fnrizzi:
@cz4rs since you worked a bit on magistrate, any comments on this?

masterleinad:
I would prefer to just deprecate. After all, the class is pretty isolated and it's not obvious that its existence would impact any refactoring or cause any other problems.

masterleinad (approved):
We agreed in the team meeting to remove `LogicalMemorySpace` without deprecating it. The removal looks fine to me.

PR #6558, opened 2023-10-31
Enable `Experimental::LogicalMemorySpace` with OpenACC and OpenMPTarget backends

dalg24:
Since there is at least an objection to removing `Experimental::LogicalMemorySpace` and it is not clear at this time if we will remove or not, I figured I would attempt to enable.  I noticed these friend declarations were missing and assumed it would just work if I add them.


dalg24:
HIP timeouts can be ignored

PR #6569, opened 2023-11-01
Cleanup for static_asserts with empty error

masterleinad:
Follow-up to https://github.com/kokkos/kokkos/pull/6543#discussion_r1378738193.
This pull request removes all `static_assert`s with an empty error message ~~and replaces `KOKKOS_IMPL_DO_NOT_USE_PRINTF` with `Kokkos::printf` for the few places we introduced it after the initial cleanup~~.

### DIFF
--- a/algorithms/unit_tests/TestStdAlgorithmsModOps.cpp
+++ b/algorithms/unit_tests/TestStdAlgorithmsModOps.cpp
@@ -48,7 +48,7 @@ struct MyMovableType {
 TEST(std_algorithms_mod_ops_test, move) {
   MyMovableType a;
   using move_t = decltype(std::move(a));
-  static_assert(std::is_rvalue_reference<move_t>::value, "");
+  static_assert(std::is_rvalue_reference<move_t>::value);
 
   // move constr
   MyMovableType b(std::move(a));
@@ -70,7 +70,7 @@ struct StdAlgoModSeqOpsTestMove {
   void operator()(const int index) const {
     typename ViewType::value_type a{11};
     using move_t = decltype(std::move(a));
-    static_assert(std::is_rvalue_reference<move_t>::value, "");
+    static_assert(std::is_rvalue_reference<move_t>::value);
     m_view(index) = std::move(a);
   }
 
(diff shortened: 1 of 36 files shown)


dalg24:
Note that in the original changes it made sense to combine these because they were in the same source file...

masterleinad:
> Note that in the original changes it made sense to combine these because they were in the same source file...

Just doing that `static_assert` now.

dalg24 (commented):
You missed those broken over multilines, e.g.
https://github.com/kokkos/kokkos/blob/0e5aa150355cc5163198ae1cbba7d89fef2ab61f/core/unit_test/hip/TestHIP_Spaces.cpp#L31-L33

Use modernize-unary-static-assert

masterleinad:
> Use modernize-unary-static-assert

If only we would run `clang-tidy` for all backends... I ran it locally and fixed the remaining occurrences with line breaks manually.

dalg24:
> > Use modernize-unary-static-assert
> 
> If only we would run `clang-tidy` for all backends... I ran it locally and fixed the remaining occurrences with line breaks manually.

Still worth adding
```diff
diff --git a/.clang-tidy b/.clang-tidy
index 2b0d6e51d..6047b9d35 100644
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,3 +1,9 @@
-Checks: '-*,kokkos-*,modernize-use-using,modernize-use-nullptr,cppcoreguidelines-pro-type-cstyle-cast'
+Checks: "-*,\
+kokkos-*,\
+modernize-use-using,\
+modernize-use-nullptr,\
+modernize-unary-static-assert,\
+cppcoreguidelines-pro-type-cstyle-cast,\
+"
 FormatStyle: file
 HeaderFilterRegex: '.*/*.hpp'

dalg24:
But can do in a follow up

masterleinad:
> But can do in a follow up

Maybe, we discuss some more checks like `modernize-type-traits` as well.

fnrizzi (approved):
assuming all instances were fixed, looks good!

dalg24:
Ignoring the OpenMPTarget failure

dalg24 (approved):

PR #6570, opened 2023-11-02
simd: cleanup fallback implementations

ldh4:
Related to #5674.

Current fallback implementations for `operator*` and `operator>>` that are placed in `Kokkos_SIMD_Common` are only needed for AVX2 backend as corresponding intrinsics are not available in AVX2 and are provided in AVX512 instead.

Because these fallback impls are not for general use, moved these from SIMD_Common header to AVX2 backend file.

### DIFF
--- a/simd/src/Kokkos_SIMD_AVX2.hpp
+++ b/simd/src/Kokkos_SIMD_AVX2.hpp
@@ -1111,6 +1111,11 @@ class simd<std::int32_t, simd_abi::avx2_fixed_size<4>> {
     return simd(
         _mm_add_epi32(static_cast<__m128i>(lhs), static_cast<__m128i>(rhs)));
   }
+  [[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION friend simd operator*(
+      simd const& lhs, simd const& rhs) noexcept {
+    return simd(
+        _mm_mullo_epi32(static_cast<__m128i>(lhs), static_cast<__m128i>(rhs)));
+  }
 
   [[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION friend simd operator>>(
       simd const& lhs, int rhs) noexcept {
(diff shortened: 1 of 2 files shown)


masterleinad (approved):

dalg24:
OpenMPTarget CI builds fails with
```
48/48 Test #48: Kokkos_UnitTest_SIMD ........................Subprocess aborted***Exception:   3.21 sec
```
I think I might have seen that failure elsewhere but I want to hear from @ldh4 @rgayatri23 whether they want to rerun or ignore

rgayatri23:
> OpenMPTarget CI builds fails with
> 
> ```
> 48: [ RUN      ] simd.device_math_ops
> 48: "PluginInterface" error: Failure to synchronize stream (nil): Error in cuStreamSynchronize: an illegal memory access was encountered
> 48: Libomptarget error: Consult https://openmp.llvm.org/design/Runtimes.html for debugging options.
> 48: Kokkos_OpenMPTarget_ParallelFor_Range.hpp:54:1: Libomptarget fatal error 1: failure of target construct while offloading is mandatory
> 48/48 Test #48: Kokkos_UnitTest_SIMD ........................Subprocess aborted***Exception:   3.21 sec
> ```
> 
> I think I might have seen that failure elsewhere but I want to hear from @ldh4 @rgayatri23 whether they want to rerun or ignore

I think we saw something similar for a unit test in a different SIMD related PR. I believe it was because of issues with SIMD length. If the failure repeats everytime, we might have to protect against it.

ldh4:
> > OpenMPTarget CI builds fails with
> > ```
> > 48: [ RUN      ] simd.device_math_ops
> > 48: "PluginInterface" error: Failure to synchronize stream (nil): Error in cuStreamSynchronize: an illegal memory access was encountered
> > 48: Libomptarget error: Consult https://openmp.llvm.org/design/Runtimes.html for debugging options.
> > 48: Kokkos_OpenMPTarget_ParallelFor_Range.hpp:54:1: Libomptarget fatal error 1: failure of target construct while offloading is mandatory
> > 48/48 Test #48: Kokkos_UnitTest_SIMD ........................Subprocess aborted***Exception:   3.21 sec
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I think I might have seen that failure elsewhere but I want to hear from @ldh4 @rgayatri23 whether they want to rerun or ignore
> 
> I think we saw something similar for a unit test in a different SIMD related PR. I believe it was because of issues with SIMD length. If the failure repeats everytime, we might have to protect against it.

But it looks like it's not a deterministic failure, which confuses me. There are some currently open SIMD PRs that are not producing this error, but on the other hand, I see that there are non-SIMD PRs that are having this error in their CI test result.

We can have `simd.device_math_ops` be skipped if `OpenMPTarget` is enabled, if that's how this should be handled.

rgayatri23:
I don't think it was exactly the same unit test. But if its not deterministic then we can ignore it for now.

dalg24 (approved):

PR #6572, opened 2023-11-02
view: restore constexprness

romintomasetti:
This PR brings back some `Kokkos::View` *constexprness*.

Indeed, it seems a lot of `Kokkos` functionalities have been marked `constexpr`, but that's not tested thoroughly. For instance, it was broken for the `Kokkos::View`.

I think the changes are legitimate because they are mostly writing `constructor() = default;`.

This is a draft, because it only works in C++20 for the moment. The culprit seems the `SharedAllocationTracker` destructor that can be made *constexpr* in `c++20`, but not in `c++17`.

### DIFF
--- a/core/src/impl/Kokkos_SharedAlloc.hpp
+++ b/core/src/impl/Kokkos_SharedAlloc.hpp
@@ -414,10 +414,11 @@ union SharedAllocationTracker {
     m_record_bits = DO_NOT_DEREF_FLAG;
   }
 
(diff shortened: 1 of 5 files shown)


dalg24:
I think that you could explore having unmanaged views in constant expressions.
I don't see how you can have atomics (reference counters) and allocations be constexpr at this time.
Which brings up the question: should you be using std::mdspan directly instead.

PR #6574, opened 2023-11-02
simd: temporarily skip device math ops unit test for OpenMPTarget build

ldh4:
Follow-up from https://github.com/kokkos/kokkos/pull/6570#issuecomment-1791677168.

The CI test has been reporting following simd test failures for OpenMPTarget build:
```
48/48 Test #48: Kokkos_UnitTest_SIMD ........................Subprocess aborted***Exception:   3.21 sec
```
However, this failure seems non-deterministic, seeing that only a few PRs, including non-SIMD related PRs, are producing this particular error during the CI testing while some other SIMD PRs are not producing this error.
Placing a temporary skip in `simd.device_math_ops` for `OpenMPTarget` build to avoid noise in the CI test results.

### DIFF
--- a/simd/unit_tests/include/TestSIMD_MathOps.hpp
+++ b/simd/unit_tests/include/TestSIMD_MathOps.hpp
@@ -298,7 +298,7 @@ TEST(simd, device_math_ops) {
   GTEST_SKIP()
       << "skipping because of a non-deterministic failure reporting: "
          "Failure to synchronize stream (nil): Error in "
-         "cuStreamSynchronize: an illegal memory access was encountered"
+         "cuStreamSynchronize: an illegal memory access was encountered";
 #endif
       Kokkos::parallel_for(1, simd_device_math_ops_functor());
 }
--- a/simd/unit_tests/include/TestSIMD_MathOps.hpp
+++ b/simd/unit_tests/include/TestSIMD_MathOps.hpp
@@ -300,7 +300,7 @@ TEST(simd, device_math_ops) {
          "Failure to synchronize stream (nil): Error in "
          "cuStreamSynchronize: an illegal memory access was encountered";
 #endif
-      Kokkos::parallel_for(1, simd_device_math_ops_functor());
+  Kokkos::parallel_for(1, simd_device_math_ops_functor());
 }
 
 #endif
--- a/simd/unit_tests/include/TestSIMD_MathOps.hpp
+++ b/simd/unit_tests/include/TestSIMD_MathOps.hpp
@@ -294,6 +294,9 @@ TEST(simd, host_math_ops) {
 }
 
 TEST(simd, device_math_ops) {
+#ifdef KOKKOS_ENABLE_OPENMPTARGET
+  GTEST_SKIP();
+#endif
   Kokkos::parallel_for(1, simd_device_math_ops_functor());
 }
 
--- a/simd/unit_tests/include/TestSIMD_MathOps.hpp
+++ b/simd/unit_tests/include/TestSIMD_MathOps.hpp
@@ -294,10 +294,13 @@ TEST(simd, host_math_ops) {
 }
 
 TEST(simd, device_math_ops) {
-#ifdef KOKKOS_ENABLE_OPENMPTARGET
-  GTEST_SKIP();
+#ifdef KOKKOS_ENABLE_OPENMPTARGET  // FIXME_OPENMPTARGET
+  GTEST_SKIP()
+      << "skipping because of a non-deterministic failure reporting: "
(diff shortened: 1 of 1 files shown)


ldh4:
In the meantime, I will look into this to find out what may be causing the problem.

rgayatri23 (approved):
Fine for now. I will review what the issue is with this unit test.

dalg24:
Please resolve the conflicts.
Also this is not the right way to disable a test.
If you use a big hammer you disable outside the body of the unit test so it does not give the impression that the test ran.
If it is only a runtime error but you still wish to ensure it compiles you use GTEST_SKIP().

ldh4:
> Please resolve the conflicts. Also this is not the right way to disable a test. If you use a big hammer you disable outside the body of the unit test so it does not give the impression that the test ran. If it is only a runtime error but you still wish to ensure it compiles you use GTEST_SKIP().

Rebased and resolved merge conflicts.
Placed in a `GTEST_SKIP()` since this is to suppress a runtime error.

simd/unit_tests/include/TestSIMD_MathOps.hpp
```diff
@@ -294,6 +294,9 @@ TEST(simd, host_math_ops) {
 }
 
 TEST(simd, device_math_ops) {
+#ifdef KOKKOS_ENABLE_OPENMPTARGET
+  GTEST_SKIP();
+#endif
```
dalg24, line 299:
```suggestion
#ifdef KOKKOS_ENABLE_OPENMPTARGET  // FIXME_OPENMPTARGET
  GTEST_SKIP() << "some helpful message";
#endif
```

dalg24 (commented):
I would prefer if you added a FIXME_OPENMPTARGET tag and injected a message into the SKIP.

simd/unit_tests/include/TestSIMD_MathOps.hpp
ldh4: Added a short description about the reported failure. c0c64b1.

simd/unit_tests/include/TestSIMD_MathOps.hpp
```diff
@@ -294,7 +294,13 @@ TEST(simd, host_math_ops) {
 }
 
 TEST(simd, device_math_ops) {
-  Kokkos::parallel_for(1, simd_device_math_ops_functor());
+#ifdef KOKKOS_ENABLE_OPENMPTARGET  // FIXME_OPENMPTARGET
+  GTEST_SKIP()
+      << "skipping because of a non-deterministic failure reporting: "
+         "Failure to synchronize stream (nil): Error in "
+         "cuStreamSynchronize: an illegal memory access was encountered"
```
dalg24, line 301:
```suggestion
         "cuStreamSynchronize: an illegal memory access was encountered";
```

simd/unit_tests/include/TestSIMD_MathOps.hpp
```diff
@@ -294,7 +294,13 @@ TEST(simd, host_math_ops) {
...
 
 TEST(simd, device_math_ops) {
-  Kokkos::parallel_for(1, simd_device_math_ops_functor());
+#ifdef KOKKOS_ENABLE_OPENMPTARGET  // FIXME_OPENMPTARGET
+  GTEST_SKIP()
+      << "skipping because of a non-deterministic failure reporting: "
+         "Failure to synchronize stream (nil): Error in "
+         "cuStreamSynchronize: an illegal memory access was encountered";
+#endif
+      Kokkos::parallel_for(1, simd_device_math_ops_functor());
```
dalg24, line 303:
```suggestion
  Kokkos::parallel_for(1, simd_device_math_ops_functor());
```

dalg24: clang-format failed
```diff

diff --git a/simd/unit_tests/include/TestSIMD_MathOps.hpp b/simd/unit_tests/include/TestSIMD_MathOps.hpp
index 6ef780e69..6f8a8aa0f 100644
--- a/simd/unit_tests/include/TestSIMD_MathOps.hpp
+++ b/simd/unit_tests/include/TestSIMD_MathOps.hpp
@@ -300,7 +300,7 @@ TEST(simd, device_math_ops) {
          "Failure to synchronize stream (nil): Error in "
          "cuStreamSynchronize: an illegal memory access was encountered";
 #endif
-      Kokkos::parallel_for(1, simd_device_math_ops_functor());
+  Kokkos::parallel_for(1, simd_device_math_ops_functor());
 }
 
 #endif

masterleinad (approved):

dalg24:
omptarget build passed

PR #6577, opened 2023-11-03
SYCL: Address deprecations after oneAPI 2023.2.0

masterleinad:
The only new API function introduced in `oneAPI 2023.2.0` that addresses later deprecations (for `sycl::accessor::get_pointer`) is `sycl::accessor<>::get_multi_ptr`, see the changelog at https://www.intel.com/content/www/us/en/developer/articles/release-notes/intel-oneapi-dpc-c-compiler-release-notes.html. `get_multi_ptr` is conforming to the SYCL2020 specifications, see https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_interface_for_buffer_command_accessors.
We can use decorated pointers here, so we can make it easier for the compiler to figure out the correct address space (although that wasn't a problem before).

### DIFF
--- a/core/src/SYCL/Kokkos_SYCL_ParallelFor_Team.hpp
+++ b/core/src/SYCL/Kokkos_SYCL_ParallelFor_Team.hpp
@@ -76,14 +76,8 @@ class Kokkos::Impl::ParallelFor<FunctorType, Kokkos::TeamPolicy<Properties...>,
 
       auto lambda = [=](sycl::nd_item<2> item) {
         const member_type team_member(
-#if defined(KOKKOS_COMPILER_INTEL_LLVM) && \
-    KOKKOS_COMPILER_INTEL_LLVM >= 20230200
-            team_scratch_memory_L0
-                .get_multi_ptr<sycl::access::decorated::yes>(),
-#else
-            team_scratch_memory_L0.get_pointer(),
-#endif
-            shmem_begin, scratch_size[0],
+            KOKKOS_IMPL_SYCL_GET_MULTI_PTR(team_scratch_memory_L0), shmem_begin,
+            scratch_size[0],
             global_scratch_ptr + item.get_group(1) * scratch_size[1],
             scratch_size[1], item, item.get_group_linear_id(),
             item.get_group_range(1));
--- a/core/src/SYCL/Kokkos_SYCL_ParallelFor_Team.hpp
+++ b/core/src/SYCL/Kokkos_SYCL_ParallelFor_Team.hpp
@@ -76,7 +76,14 @@ class Kokkos::Impl::ParallelFor<FunctorType, Kokkos::TeamPolicy<Properties...>,
 
       auto lambda = [=](sycl::nd_item<2> item) {
         const member_type team_member(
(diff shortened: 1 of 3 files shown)


dalg24 (commented):
Did you consider defining some `KOKKOS_IMPL_SYCL_USE_ACCESSOR_GET_MULTI_PTR` macro or whatnot in `Kokkos_Setup_SYCL.hpp`?

masterleinad:
> Did you consider defining some `KOKKOS_IMPL_SYCL_USE_ACCESSOR_GET_MULTI_PTR` macro or whatnot in `Kokkos_Setup_SYCL.hpp`?

No, but 2535a63 does that.

dalg24:
> > Did you consider defining some `KOKKOS_IMPL_SYCL_USE_ACCESSOR_GET_MULTI_PTR` macro or whatnot in `Kokkos_Setup_SYCL.hpp`?
> 
> No, but [2535a63](https://github.com/kokkos/kokkos/commit/2535a632afa1a05b97762bc1b0d889a03ca68c27) does that.

If that is the approach you chose then why are you a macro (rather than a function call)

masterleinad:
> If that is the approach you chose then why are you a macro (rather than a function call)

Mostly because it fits directly what I do in the first commit anyway and it just provides some shorthand notation to be used in a very limited number of places. I don't see much of an advantage in converting it to a function.

dalg24:
The function-like macro is not the right tool for the job here IMO.  You are welcome to wait and see if others agree with you.

masterleinad:
> The function-like macro is not the right tool for the job here IMO. You are welcome to wait and see if others agree with you.

What do you propose? Did you like the initial version better?

dalg24:
I suggested
```C++
// Kokkos_Setup_SYCL.hpp
#if <WHATEVS>
#define KOKKOS_IMPL_SYCL_USE_ACCESSOR_GET_MULTI_PTR
#endif

// everywhere else so you don't repeat the <WHATEV> and you can easily audit where it is used
#ifdef KOKKOS_IMPL_SYCL_USE_ACCESSOR_GET_MULTI_PTR
foo.get_pointer()
#else
foo.get_multi_ptr<...>()
#endif
```

masterleinad:
That doesn't make the code more concise, though. We use the same approach, e.g.,  for `KOKKOS_IMPL_OMPX_DYN_CGROUP_MEM`.

dalg24:
> That doesn't make the code more concise, though.

I did not go for conciseness, I went for not repeating the condition on the minimum toolchain version.  I judged that I'd rather see the two variants at call site but note that I did not block and I will merge if other reviewers like your approach better.

> We use the same approach, e.g., for `KOKKOS_IMPL_OMPX_DYN_CGROUP_MEM`.

I am glad you bring up that one.  It is a good example of the function-like macro being a good fit in a different situation.  Not repeating a complex pragma, just conditionally appending another directive.

crtrott (approved):
If it were just two places I would have definitely gone with two places of #ifdef with 10 places I would have clearly preferred function, or function macro approach. With 4 I am on the fence. But I gonna approve it as is so we can move forward.

dalg24 (approved):
Approving for the sake of moving forward

Issue #6517, opened 2023-10-16
Silent failure when reaching the maximum iteration count


petschge:
The following code works

```
#include <Kokkos_Vector.hpp>

int main(int argc, char* argv[]) {
  Kokkos::initialize(argc, argv);

  const  int64_t N = pow(2LL, 39LL) - pow(2LL, 8LL);
  double nu = 0;

  Kokkos::parallel_reduce("sample reduction",
    Kokkos::RangePolicy <Kokkos::IndexType<int64_t>> (0,N),
    KOKKOS_LAMBDA(const int64_t i, double & mynu_) { mynu_ += 1.0; },
    nu);
  Kokkos::fence();

  printf("N=%lu, nu=%g\n", N, nu);

  Kokkos::finalize();
  return 0;
}
```

and gives "N=549755813632, nu=5.49756e+11" as expected but changing the definition of N to
`const  int64_t N = pow(2LL, 39LL) - pow(2LL, 8LL) + 1;`
breaks things and results in "N=549755813633, nu=0" on GPU (it work on CPU / the OpenMP backend, but is of course much slower).

Kokkos config looks as follows:
```
#define KOKKOS_VERSION 40100
#define KOKKOS_VERSION_MAJOR 4
#define KOKKOS_VERSION_MINOR 1
#define KOKKOS_VERSION_PATCH 0
#define KOKKOS_ENABLE_SERIAL
#define KOKKOS_ENABLE_OPENMP
#define KOKKOS_ENABLE_CUDA
#define KOKKOS_ENABLE_CXX17
#define KOKKOS_ENABLE_CUDA_LAMBDA  // deprecated
#define KOKKOS_ENABLE_DEPRECATED_CODE_4
#define KOKKOS_ENABLE_DEPRECATION_WARNINGS
#define KOKKOS_ENABLE_COMPLEX_ALIGN
#define KOKKOS_ENABLE_LIBDL
#define KOKKOS_ARCH_TURING75
```

As discussed on slack, this is (probably) an issue of reaching the 2^31-1 iteration limit with a block_size of 256. It would be nice to get an explicit error message when reaching that limit instead of silent failure.

Rombur:
We need to check but I believe that you get an error if you enable `Kokkos_ENABLE_DEBUG_BOUNDS_CHECK`.

fnrizzi:
[...]

PR #6578, opened 2023-11-06
cuda: reduction with `RangePolicy`: fix grid dimensions to work for large values and avoid overflow

fnrizzi:
This PR addresses https://github.com/kokkos/kokkos/issues/6517 

## what is the issue
The reduction did not work in that test case because, currently, the code setting the grid in the cuda reduction via Range policy is using the following logic (note the use of `int`):
```cpp
// REQUIRED ( 1 , N , 1 )
dim3 block(1, block_size, 1);
// Required grid.x <= block.y
dim3 grid(std::min(int(block.y), int((nwork + block.y - 1) / block.y)), 1, 1);
```
so when `block_size` is 256 (which happens to be the case for the use case reported in 6517), this evaluates to: 
```
block = 1,256,1
grid = -2147483648 1 1
```
so the min is taking 2147483648 and the grid is set incorrectly. 
So the issue was not a too large value of N for the reduction, but the actual logic in setting the grid dim internally. 

## solution 
This PR replaces the code above with: 
```cpp
// REQUIRED ( 1 , N , 1 )
dim3 block(1, block_size, 1);
// Required grid.x <= block.y
dim3 grid(std::min(index_type(block.y), index_type((nwork + block.y - 1) / block.y)), 1, 1);
```
so that `min` operates on meaningful values, and `grid` is set correctly even for large values of `nwork`.

## test 

The test added reproduces the original issue and should pass now

### DIFF
--- a/core/src/Cuda/Kokkos_Cuda_Parallel_Range.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_Parallel_Range.hpp
@@ -115,7 +115,6 @@ class ParallelFor<FunctorType, Kokkos::RangePolicy<Traits...>, Kokkos::Cuda> {
         std::min(typename Policy::index_type((nwork + block.y - 1) / block.y),
                  typename Policy::index_type(maxGridSizeX)),
         1, 1);
-
 #ifdef KOKKOS_IMPL_DEBUG_CUDA_SERIAL_EXECUTION
     if (Kokkos::Impl::CudaInternal::cuda_use_serial_execution()) {
       block = dim3(1, 1, 1);
--- a/core/src/Cuda/Kokkos_Cuda_Parallel_Range.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_Parallel_Range.hpp
@@ -32,7 +32,6 @@
 
 #include <impl/Kokkos_Tools.hpp>
 #include <typeinfo>
-#include <iostream>
 
 namespace Kokkos {
 namespace Impl {
--- a/core/src/Cuda/Kokkos_Cuda_Parallel_Range.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_Parallel_Range.hpp
@@ -37,8 +37,8 @@ namespace Kokkos {
 namespace Impl {
 
 template <typename IndexType>
-void admissible_iteration_count_else_throw(IndexType /*nwork*/, int blockSize,
-                                           const std::string& /*patternName*/,
+void admissible_iteration_count_else_throw(IndexType nwork, int blockSize,
+                                           const std::string& patternName,
                                            Impl::CudaInternal* instance) {
   const int max_grid_size        = instance->m_deviceProp.maxGridSize[0];
   const std::size_t max_num_iter = (std::size_t)max_grid_size * blockSize;
--- a/core/src/Cuda/Kokkos_Cuda_Parallel_Range.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_Parallel_Range.hpp
@@ -42,7 +42,8 @@ void admissible_iteration_count_else_throw(IndexType nwork, int blockSize,
                                            Impl::CudaInternal* instance) {
   const int max_grid_size        = instance->m_deviceProp.maxGridSize[0];
   const std::size_t max_num_iter = (std::size_t)max_grid_size * blockSize;
-  if ((std::size_t)nwork / blockSize > (std::size_t)max_grid_size) {
+
+  if ((std::size_t)nwork > max_num_iter) {
     const std::string msg =
         patternName + ": the target number of iterations " +
         std::to_string(nwork) + " > the maximum number of iterations " +
--- a/core/src/Cuda/Kokkos_Cuda_Parallel_Range.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_Parallel_Range.hpp
@@ -971,8 +971,7 @@ class ParallelScanWithTotal<FunctorType, Kokkos::RangePolicy<Traits...>,
     if (nwork) {
       enum { GridMaxComputeCapability_2x = 0x0ffff };
 
-      const int block_size =
-          256;  // local_block_size(m_functor_reducer.get_functor());
+      const int block_size = local_block_size(m_functor_reducer.get_functor());
       KOKKOS_ASSERT(block_size > 0);
 
       const int grid_max =
--- a/core/src/Cuda/Kokkos_Cuda_Parallel_Range.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_Parallel_Range.hpp
@@ -36,6 +36,21 @@
 namespace Kokkos {
 namespace Impl {
 
+template <typename IndexType>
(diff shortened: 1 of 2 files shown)


core/unit_test/TestReduce.hpp
```diff
@@ -625,4 +625,22 @@ TEST(TEST_CATEGORY, int_combined_reduce_mixed) {
 }
 #endif
 #endif
+
+#if defined(KOKKOS_ENABLE_CUDA)
```
masterleinad, line 629: Why restrict the test to just `Cuda`? How are the other backends behaving?

crtrott: agreed we should check this always.

crtrott: I will update the PR

crtrott (changes requested):
I will update this to have the test always on.

core/unit_test/TestReduce.hpp
fnrizzi: @masterleinad sorry i saw your comment but forgot to answer: the reason i kept this guarded is that originally i wanted to proceed one backend at a time, so i first wanted to ensure the cuda case worked since it was reported in the issue and get that merged. Then i was going to try this on the other backends . 

@crtrott i am updating it now to run alwasy

fnrizzi:
i think the serial bakend is timing out since the loop is very large 
https://github.com/kokkos/kokkos/actions/runs/7471719285/job/20332477996?pr=6578

crtrott:
Ok lets restrict to GPU backends and Release mode.

core/unit_test/TestReduce.hpp
```diff
@@ -625,4 +625,42 @@ TEST(TEST_CATEGORY, int_combined_reduce_mixed) {
...
+struct EnableTestForIssue6517<Kokkos::Experimental::SYCL> : std::true_type {};
+#endif
+
+struct FunctorIssue6517 {
+  KOKKOS_FUNCTION void operator()(const int64_t /*i*/, double& update) const {
+    update += 1.0;
+  }
+};
+
+TEST(TEST_CATEGORY, issue6517) {
```
masterleinad, line 652: Can we find a better, more descriptive name than `issue617`?

fnrizzi: i like it because we can easily find the issue :) 
you have any suggestions?

masterleinad: `large_iteration_count` or something like that.

fnrizzi:
> Ok lets restrict to GPU backends and Release mode.

i restricted to cuda, hip and sycl -  is it  enough or want me to extend to openacc too maybe?

masterleinad (commented):
Let's also enable `OpenMPTarget` and `OpenACC` and see if they do the right thing.

fnrizzi:
retest this please

core/unit_test/TestReduce.hpp
```diff
@@ -625,4 +625,63 @@ TEST(TEST_CATEGORY, int_combined_reduce_mixed) {
 }
 #endif
 #endif
+
+#if defined(NDEBUG)
+// the following test was made for:
+// https://github.com/kokkos/kokkos/issues/6517
+
+template <class T>
+struct EnableTestForReductionWithLargeIterationCount : std::false_type {};
```
dalg24, line 634: I would have defined directly a variable template w/o going through a trait class.

core/unit_test/TestReduce.hpp
```diff
@@ -625,4 +625,63 @@ TEST(TEST_CATEGORY, int_combined_reduce_mixed) {
...
+                    TEST_EXECSPACE>::value) {
+    GTEST_SKIP() << "Disabling for host backends";
+  }
+
+  const int64_t N = pow(2LL, 39LL) - pow(2LL, 8LL) + 1;
+  Kokkos::RangePolicy<TEST_EXECSPACE, Kokkos::IndexType<int64_t>> p(0, N);
+  double nu = 0;
+  EXPECT_NO_THROW(Kokkos::parallel_reduce(
+      "sample reduction", p, FunctorReductionWithLargeIterationCount(), nu));
+  ASSERT_DOUBLE_EQ(nu, double(N));
```
dalg24, line 683: Does strict comparison actually fail?

I would have gone for reducing into a 64-bit integer that way we don't have to go through that macro and remember how many bits go to the mantissa of a IEEE double to be sure that we stay in the range of integers that can be represented exactly.

dalg24 (approved):

core/unit_test/TestReduce.hpp
fnrizzi: I took the test from the issue that was posted without major changes

core/unit_test/TestReduce.hpp
fnrizzi: Yeah good point

core/unit_test/TestReduce.hpp
```diff
@@ -625,4 +625,63 @@ TEST(TEST_CATEGORY, int_combined_reduce_mixed) {
...
+  KOKKOS_FUNCTION void operator()(const int64_t /*i*/, double& update) const {
+    update += 1.0;
+  }
+};
+
+TEST(TEST_CATEGORY, reduction_with_large_iteration_count) {
+  if constexpr (!EnableTestForReductionWithLargeIterationCount<
+                    TEST_EXECSPACE>::value) {
+    GTEST_SKIP() << "Disabling for host backends";
+  }
```
masterleinad, line 676:
```suggestion
  if constexpr (std::is_same_v<TEST_EXECSPACE::memory_space, Kokkos::HostSpace>) {
    GTEST_SKIP() << "Disabling for host backends";
  }
```

masterleinad (approved):

core/src/Cuda/Kokkos_Cuda_Parallel_Range.hpp
```diff
@@ -308,8 +308,9 @@ class ParallelReduce<CombinedFunctorReducerType, Kokkos::RangePolicy<Traits...>,
       // REQUIRED ( 1 , N , 1 )
       dim3 block(1, block_size, 1);
       // Required grid.x <= block.y
-      dim3 grid(std::min(int(block.y), int((nwork + block.y - 1) / block.y)), 1,
-                1);
+      dim3 grid(std::min(index_type(block.y),
+                         index_type((nwork + block.y - 1) / block.y)),
```
cedricchevalier19, line 312: Hopping the addition does not overflow.

crtrott: yeah, for now this is ok: we could in debug mode add some more comprehensive checks later

core/unit_test/TestReduce.hpp
crtrott: I think it's unlikely to fail, but it doesn't hurt either. double is 11 bits for exponent. You take away 1 bit for sign too, that means up to 52bits of integral types should be fine I think?

crtrott (approved):

PR #6579, opened 2023-11-06
Remove Kokkos::[b]half_t volatile overloads

masterleinad:
`Kokkos::[b]half_t` is the only `Kokkos`-defined type that still supports `volatile` semantics (without deprecations). In particular, we deprecated the overloads for the more widely used `Kokkos::complex` in https://github.com/kokkos/kokkos/pull/5412. `fixed_width_integer_type` is defined only for these overloads and we need to do some `reinterpret_cast` to work around missing support for the backend-specific types. #6543 introduces `bit_comparison_type` that serves a similar purpose but replacing the `reinterpret_casts` with `bit_cast` is difficult. Thus, to clean up the interface, this pull request suggests to ~~deprecate~~ remove the `volatile` overloads.

### DIFF
--- a/core/src/impl/Kokkos_Half_FloatingPointWrapper.hpp
+++ b/core/src/impl/Kokkos_Half_FloatingPointWrapper.hpp
@@ -196,20 +196,12 @@ KOKKOS_INLINE_FUNCTION
 
 template <class T>
 static KOKKOS_INLINE_FUNCTION Kokkos::Experimental::half_t cast_to_wrapper(
-    T x, const
-#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
-    volatile
-#endif
-    Kokkos::Impl::half_impl_t::type&);
+    T x, const Kokkos::Impl::half_impl_t::type&);
 
 #ifdef KOKKOS_IMPL_BHALF_TYPE_DEFINED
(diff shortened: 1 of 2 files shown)


dalg24:
@masterleinad why should we deprecate rather than just remove?

masterleinad:
> @masterleinad why should we deprecate rather than just remove?

Because I am trying to be conservative and give people an opportunity to complain. 🙂

dalg24:
Half types belong to namespace Experimental and I suspect we have already removed the `volatile` overload in other classes.

masterleinad:
I'm happy to remove them already, see 06f1766.

dalg24 (approved):

crtrott (approved):

crtrott:
@lucbv any reasons to have the volatile things?

dalg24:
Per conversation on Slack, Evan said the `volatile` overloads were added to support reductions.  Since that requirement was lifted a while ago, we should be good to remove.

dalg24:
Retest this please

dalg24:
Retest this please

dalg24:
SYCL-OneAPI build was aborted and CUDA-11.6-NVCC-DEBUG was killed

PR #6584, opened 2023-11-07
containers(UnorderedMap): correct conditionals for the case of a set

romintomasetti:
This PR:
- fixes two conditionals that protect the case of a set (i.e., just keys, no values); these conditionals were missing the `!` and hence worked in the wrong direction
- adds a test specific to the case of a set; it inserts, erases at random keys, and rehashes
- ~~add several missing execution space as arguments in `UnorderedMap`~~ (done in #6610)
- ~~fix missing `Kokkos::view_alloc` in `Kokkos::unique`~~ (done in #6598)

### DIFF
--- a/containers/src/impl/Kokkos_UnorderedMap_impl.hpp
+++ b/containers/src/impl/Kokkos_UnorderedMap_impl.hpp
@@ -91,7 +91,7 @@ struct UnorderedMapErase {
       next                     = m_map.m_next_index[curr];
       m_map.m_next_index[curr] = invalid_index;
       m_map.m_keys[curr]       = key_type();
-      if (m_map.is_set) m_map.m_values[curr] = value_type();
+      if constexpr (!map_type::is_set) m_map.m_values[curr] = value_type();
       curr                  = next;
       m_map.m_hash_lists(i) = next;
     }
@@ -110,7 +110,7 @@ struct UnorderedMapErase {
           m_map.m_next_index[prev] = next;
           m_map.m_next_index[curr] = invalid_index;
           m_map.m_keys[curr]       = key_type();
-          if (map_type::is_set) m_map.m_values[curr] = value_type();
+          if constexpr (!map_type::is_set) m_map.m_values[curr] = value_type();
         }
         curr = next;
       }
--- a/containers/unit_tests/TestUnorderedSet.hpp
+++ b/containers/unit_tests/TestUnorderedSet.hpp
@@ -82,15 +82,10 @@ TEST(TEST_CATEGORY, UnorderedSet_insert_erase_and_rehash) {
 
   // Always try to erase truly random indices by initializing the random pool
   // randomly.
-  std::random_device dev;
-  std::mt19937 rng(dev());
-  std::uniform_int_distribution<uint64_t> dist(
-      0, std::numeric_limits<uint64_t>::max());
+  std::mt19937 rng(testing::UnitTest::GetInstance()->random_seed());
+  const uint64_t seed = std::uniform_int_distribution<uint64_t>{}(rng);
 
   // Generate random indices between 0 and size_all. Those will be erased.
-  const uint64_t seed = dist(rng);
-  SCOPED_TRACE("Random seed: " + std::to_string(seed));
-
   Kokkos::Random_XorShift64_Pool<TEST_EXECSPACE> generator(seed);
   key_view_type keys_erased(
       Kokkos::view_alloc(Kokkos::WithoutInitializing, exec,
--- a/containers/unit_tests/TestUnorderedSet.hpp
+++ b/containers/unit_tests/TestUnorderedSet.hpp
@@ -1,0 +1,134 @@
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+// SPDX-FileCopyrightText: Copyright Contributors to the Kokkos project
+
+#ifndef KOKKOS_TEST_UNORDERED_SET_HPP
+#define KOKKOS_TEST_UNORDERED_SET_HPP
+
+#include <random>
+
+#include <gtest/gtest.h>
+#include <Kokkos_Macros.hpp>
+#ifdef KOKKOS_ENABLE_EXPERIMENTAL_CXX20_MODULES
+import kokkos.core;
+import kokkos.random;
+import kokkos.sort;
+import kokkos.std_algorithms;
+import kokkos.unordered_map;
+#else
+#include <Kokkos_Core.hpp>
+#include <Kokkos_Random.hpp>
+#include <Kokkos_Sort.hpp>
+#include <Kokkos_StdAlgorithms.hpp>
+#include <Kokkos_UnorderedMap.hpp>
+#endif
+
+namespace {
+
+template <typename MapType>
+struct Inserter {
+  MapType map;
+
+  template <std::integral T>
+  KOKKOS_FUNCTION void operator()(const T idx) const {
+    Kokkos::UnorderedMapInsertResult res = map.insert(idx);
+    if (!res.success()) Kokkos::abort("Inserter: insert failed.");
+  }
+};
+
+template <typename MapType, typename KeyViewType>
+struct Eraser {
+  MapType map;
+  KeyViewType keys;
+
+  template <std::integral T>
+  KOKKOS_FUNCTION void operator()(const T idx) const {
+    if (!map.erase(keys(idx))) Kokkos::abort("Eraser: erase failed.");
+  }
+};
+
+// Ensure that inserting into, erasing from, and rehashing an unordered set
+// works.
+TEST(TEST_CATEGORY, UnorderedSet_insert_erase_and_rehash) {
+  // First, size_all entries will be inserted into the set.
+  // Then, size_erased will be erased from it before rehash.
+  // Once rehashing is done, the capacity of the set is expected to
+  // be smaller than the one of the initial set.
+  constexpr size_t size_all    = 2579;
+  constexpr size_t size_erased = 568;
+
+  using uset_type = Kokkos::UnorderedMap<size_t, void, TEST_EXECSPACE>;
+  static_assert(uset_type::is_set);
+
+  using key_view_type =
+      Kokkos::View<typename uset_type::key_type*, TEST_EXECSPACE>;
+
+  const TEST_EXECSPACE exec{};
+
+  // Initialize the set.
+  uset_type uset(Kokkos::view_alloc(exec, "test uset"), size_all);
+  const auto initial_capacity = uset.capacity();
+
+  ASSERT_GE(uset.capacity(), size_all);
+  ASSERT_EQ(uset.size(), 0u);
+
+  // Insert size_all keys.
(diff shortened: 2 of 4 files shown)


containers/src/impl/Kokkos_UnorderedMap_impl.hpp
```diff
@@ -98,7 +102,7 @@ struct UnorderedMapErase {
       next                     = m_map.m_next_index[curr];
       m_map.m_next_index[curr] = invalid_index;
       m_map.m_keys[curr]       = key_type();
-      if (m_map.is_set) m_map.m_values[curr] = value_type();
+      if constexpr ( ! map_type::is_set) m_map.m_values[curr] = value_type();
```
romintomasetti, line 105: Note for the reviewers: here was the bug I think...

containers/unit_tests/TestUnorderedSet.hpp
```diff
@@ -0,0 +1,122 @@
...
+TEST(TEST_CATEGORY, UnorderedSet_insert_erase_and_rehash) {
+  /**
+   * The set will be inserted for @ref size_all entries,
+   * and then @ref size_erased will be erased from it before rehash.
+   * Once rehasing is done, the capacity of the set is expected to
+   * be smaller than the one of the initial set.
+   */
+  constexpr size_t size_all    = 2579;
+  constexpr size_t size_erased = 568;
+  static_assert(size_all > size_erased);
```
masterleinad, line 37: This comparison looks pretty pointless.

containers/unit_tests/TestUnorderedSet.hpp
```diff
@@ -0,0 +1,122 @@
...
+  using set_type = Kokkos::UnorderedMap<size_t, void, TEST_EXECSPACE>;
+  static_assert(set_type::is_set);
+
+  using policy_type = Kokkos::RangePolicy<TEST_EXECSPACE>;
+  using index_type  = typename policy_type::index_type;
+
+  using index_view_type = Kokkos::View<index_type[size_erased], TEST_EXECSPACE>;
+
+  //! Ensure that the space instance is not the default one.
+  const auto space = Kokkos::Experimental::partition_space(TEST_EXECSPACE{}, 1)[0];
```
masterleinad, line 48: `partition_space` doesn't guarantee that. 🙂 Also, why does it matter to use an execution space instance different from the default one here?

containers/unit_tests/TestUnorderedSet.hpp
```diff
@@ -0,0 +1,122 @@
...
+  namespace KE = Kokkos::Experimental;
+  const auto indices_erased_end  = KE::unique(space, KE::begin(indices_erased), KE::end(indices_erased));
+  const auto indices_erased_size = KE::distance(KE::begin(indices_erased), indices_erased_end);
+  const auto expected_uset_size  = size_all - indices_erased_size;
+
+  //! Debug information.
+  std::cout << "> initial uset   size: " << size_all            << std::endl;
+  std::cout << "> initial erased size: " << size_erased         << std::endl;
+  std::cout << "> final   erased size: " << indices_erased_size << std::endl;
+  std::cout << "> final   used   size: " << expected_uset_size  << std::endl;
```
masterleinad, line 93: Let's not clutter the test output with debug output.
```suggestion
```

containers/unit_tests/TestUnorderedSet.hpp
```diff
@@ -0,0 +1,122 @@
...
+
+  //! Ensure that the space instance is not the default one.
+  const auto space = Kokkos::Experimental::partition_space(TEST_EXECSPACE{}, 1)[0];
+
+  //! Initialize the set.
+  set_type uset(Kokkos::view_alloc(space, "test uset"), size_all);
+  const auto initial_capacity = uset.capacity();
+
+  ASSERT_GE(uset.capacity(), size_all);
+  ASSERT_EQ(uset.size()    , 0);
```
masterleinad, line 55:
```suggestion
  ASSERT_EQ(uset.size()    , 0u);
```

masterleinad (commented):
Also, make sure to run `clang-format-8`.

containers/unit_tests/TestUnorderedSet.hpp
romintomasetti: Indeed, removed :heavy_check_mark:

containers/unit_tests/TestUnorderedSet.hpp
romintomasetti: I think at least for `Cuda` and `HIP`, `partition_space` guarantees that we are now on a new stream, that is different from the default one. Am I right?

As for why I'm doing that, it is because in our own code, we noticed that it was the only easy trick to detect code paths executing in a given `space` calling code that runs on the default space, thereby triggering (only from time to time, only on GPUs) a failure in our tests. We already spotted several missing fences in our code thanks to this method.

containers/unit_tests/TestUnorderedSet.hpp
romintomasetti: Ok, removed :heavy_check_mark:

containers/unit_tests/TestUnorderedSet.hpp
romintomasetti: Just wondering why this is necessary...

masterleinad: Because we enable warnings for comparisons between signed and unsigned types and gtest doesn't handle this gracefully for us. 🙂

containers/unit_tests/TestUnorderedSet.hpp
masterleinad: Yes, that's what it does for `Cuda`, `HIP` and `SYCL`.

masterleinad:
OK to test.

masterleinad:
The indentation is still not `clang-format-8` conforming.

dalg24 (dismissed):
Split this up into several focused PRs.  Avoid drive-by changes.

romintomasetti:
> Split this up into several focused PRs.  Avoid drive-by changes.

@dalg24 Are you OK with splitting it in 3 as follows?
- [ ] change in `unique`
- [ ] missing `space` argument in *e.g.* `begin_erase`
- [ ] the fix for the uset

janciesko:
@maartenarnst Thanks to picking this up. Could you also update the title/PR description please?

containers/unit_tests/TestUnorderedSet.hpp
maartenarnst: We don't use partition_space anymore.

maartenarnst:
Hey @janciesko. Voila voila! The PR title and description are updated!

containers/unit_tests/TestUnorderedSet.hpp
```diff
@@ -0,0 +1,129 @@
...
+      "uset test - insert " + std::to_string(size_all) + " values",
+      Kokkos::RangePolicy(exec, 0, size_all), Inserter<uset_type>{.map = uset});
+  exec.fence();
+
+  ASSERT_GE(uset.capacity(), size_all);
+  ASSERT_EQ(uset.size(), size_all);
+
+  // Always try to erase truly random indices by initializing the random pool
+  // randomly.
+  std::random_device dev;
```
janciesko: Tiny nit-pick: we probably want to use a constant seed in testing for reproducibility if something starts failing.

janciesko (approved):
If CI passes, LGTM

containers/unit_tests/TestUnorderedSet.hpp
JBludau: Maybe just printing in case of test failure is the way to go here

maartenarnst: Thanks! Printing a nice message with the random seed now indeed!

JBludau:
The CI is raising a valid point:
```
[CUDA-12.2-NVHPC-AS-HOST-COMPILER] "/var/jenkins/workspace/Kokkos_PR-6584/containers/unit_tests/TestUnorderedSet.hpp", line 57: error: variable "size_erased" was declared but never referenced [declared_but_not_referenced]
[CUDA-12.2-NVHPC-AS-HOST-COMPILER]   constexpr size_t size_erased = (568); 
```

maartenarnst:
> The CI is raising a valid point:
> 
> ```
> [CUDA-12.2-NVHPC-AS-HOST-COMPILER] "/var/jenkins/workspace/Kokkos_PR-6584/containers/unit_tests/TestUnorderedSet.hpp", line 57: error: variable "size_erased" was declared but never referenced [declared_but_not_referenced]
> [CUDA-12.2-NVHPC-AS-HOST-COMPILER]   constexpr size_t size_erased = (568); 
> ```

Okay, that's weird, it's used on the next line in the static extent :) Let's try to solve by just using a view with a dynamic extent.

masterleinad (approved):

maartenarnst:
The failure in the mi300 pipeline appears unrelated as it is in a core unit test, and not in a containers unit test.

JBludau (commented):
The changes look good.
I know it is a fix, but it also is a change of behavior ... If someone created a map without initialization, erasing called the default ctor previously but will not do anything now. Should we guard it with a deprecation?

crtrott:
> The changes look good. I know it is a fix, but it also is a change of behavior ... If someone created a map without initialization, erasing called the default ctor previously but will not do anything now. Should we guard it with a deprecation?

How so? I don't see the change in behavior?

JBludau:
> How so? I don't see the change in behavior?

Doesn't 
```
if constexpr (!map_type::is_set) m_map.m_values[curr] = value_type();
```
mean when erasing a map we now overwrite with a default constructed value vs. we did nothing previously?
That would require that `value_type` is default constructable on the device. https://kokkos.org/kokkos-core-wiki/API/containers/Unordered-Map.html#description states we only want `trivially copyable` and when using `SequentialHostInit` `value_type` can be a view.

masterleinad:
> That would require that value_type is default constructable on the device. [kokkos.org/kokkos-core-wiki/API/containers/Unordered-Map.html#description](https://kokkos.org/kokkos-core-wiki/API/containers/Unordered-Map.html#description) states we only want trivially copyable and when using SequentialHostInit value_type can be a view.

Since the `if` wasn't a `if constexpr` before, it needed to be default-constructable already.

JBludau:
> Since the `if` wasn't a `if constexpr` before, it needed to be default-constructable already.

Yeah, I agree, but we still did not do it when erasing

maartenarnst:
With @romintomasetti, we have improved a bit how the random seed is generated and printed. 

We now rely on the google test random seed:
- if google test is run in shuffle mode, it prints the random seed by default
- otherwise, the random seed is 0

There is no impact on the test itself. The changes are in a new commit.

janciesko:
Thanks! @romintomasetti @maartenarnst

crtrott:
> > Since the `if` wasn't a `if constexpr` before, it needed to be default-constructable already.
> 
> 
> 
> Yeah, I agree, but we still did not do it when erasing

Which arguably was a bug right?

PR #6586, opened 2023-11-08
Desul atomics: Trade SYCL-specific compile definition for a macro defintion in the configuration header

dalg24:
Background:
The `KOKKOS_IMPL_SYCL_DEVICE_GLOBAL_SUPPORTED` macro need to be translated to `DESUL_SYCL_DEVICE_GLOBAL_SUPPORTED` in the Desul atomics library.  We were achieving this via extra compile definitions.  This PR suggest to instead define the Desul macro in the `<Kokkos_Atomics_Desul_Config.hpp> header that is serving that exact purpose.  This is more in line with what we did for NVIDIA CUDA architectures and makes the compile line less complicated.


masterleinad (approved):
Fine with me. Whenever we feel ready (the minimum `oneAPI` version supports it) we can drop that variable and only use `SYCL_EXT_ONEAPI_DEVICE_GLOBAL`

Rombur (approved):

dalg24:
SYCL build passed

PR #6587, opened 2023-11-09
Remove duplicated code in Threads::ParallelScan_Range

Rombur:
This PR just moves duplicated code to two free functions.

### DIFF
--- a/core/src/Threads/Kokkos_Threads_ParallelScan_Range.hpp
+++ b/core/src/Threads/Kokkos_Threads_ParallelScan_Range.hpp
@@ -21,6 +21,34 @@
 
 namespace Kokkos {
 namespace Impl {
+namespace Details {
+template <class TagType, class FunctorType, class Member, class ReferenceType>
+inline std::enable_if_t<std::is_void<TagType>::value> exec_range(
+    FunctorType const &functor, Member const &ibeg, Member const &iend,
+    ReferenceType update, bool const final) {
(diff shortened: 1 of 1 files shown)


masterleinad (commented):
This still duplicates a majority of the implementation for the two classes. For other backends, we derive the two classes from a `ParallelScan*Base`, see https://github.com/kokkos/kokkos/blob/91ee4e1e145b4daf69c6457ab35c7be0a4cda9f2/core/src/SYCL/Kokkos_SYCL_ParallelScan_Range.hpp#L362-L413.

core/src/Threads/Kokkos_Threads_ParallelScan_Range.hpp
```diff
@@ -21,6 +21,34 @@
 
 namespace Kokkos {
 namespace Impl {
+namespace Details {
```
ldh4, line 24: Is it necessary to introduce this namespace? Is `Impl` namespace not enough?

crtrott: Yeah we don't really need more or?

dalg24:
What is the status of this PR?

dalg24:
HIP runtime failures are unrelated (machine issue)

Rombur:
> What is the status of this PR?

It's on my TODO list to finish. There are things more important than this PR that I want to finish.

Rombur:
This is too outdated. Closing.

PR #6593, opened 2023-11-10
 Remove KOKKOS_IMPL_DO_NOT_USE_PRINTF 

masterleinad:
We introduced `Kokkos::printf` as a replacement in the last release so we can get rid of `KOKKOS_IMPL_DO_NOT_USE_PRINTF` now.

### DIFF
--- a/core/src/Kokkos_Macros.hpp
+++ b/core/src/Kokkos_Macros.hpp
@@ -339,12 +339,6 @@
 #define KOKKOS_IMPL_DEVICE_FUNCTION
 #endif
 
-// Temporary solution for SYCL not supporting printf in kernels.
-// Might disappear at any point once we have found another solution.
-#if !defined(KOKKOS_IMPL_DO_NOT_USE_PRINTF)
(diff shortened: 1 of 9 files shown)


cz4rs:
Can examples be updated to simply use `Kokkos::printf` as well?
There's a bunch of them with a following construct:
https://github.com/kokkos/kokkos/blob/1a3ea28f6e97b4c9dd2c8ceed53ad58ed5f94dfe/example/tutorial/01_hello_world/hello_world.cpp#L61-L66

masterleinad:
> Can examples be updated to simply use `Kokkos::printf` as well?

Sure, see e89c603f5.

cz4rs (approved):
Looks good to me!

Rombur (approved):

dalg24 (approved):
For a second I was worried it was used in KokkosKernels but it looks like they properly guarded it.
No usage in the rest of Trilinos so we are good.  Thx

PR #6594, opened 2023-11-13
OpenMPTarget: Remove OpenMPTargetExec

masterleinad:
Follow-up to https://github.com/kokkos/kokkos/pull/6585#discussion_r1390462744.
When moving to execution space instances the variables in `OpenMPTargetExec` have to be non-`static`.
We also move the whole `OpenMPTargetExec` into `OpenMPTargetInstance`.


masterleinad:
The OpenMPTarget build passes.

dalg24 (commented):
I will leave it up to @rgayatri23 but I think that the right thing to do is refactor to get rid of `OpenMPTargetExec` and align further with what was done in the OpenACC backend

masterleinad:
Let's wait for #7174.

masterleinad:
> I will leave it up to @rgayatri23 but I think that the right thing to do is refactor to get rid of `OpenMPTargetExec` and align further with what was done in the OpenACC backend

I pushed a commit to remove `OpenMPTargetExec`. I'm happy to propose as a follow-up, though.

rgayatri23:
> > I will leave it up to @rgayatri23 but I think that the right thing to do is refactor to get rid of `OpenMPTargetExec` and align further with what was done in the OpenACC backend
> 
> I pushed a commit to remove `OpenMPTargetExec`. I'm happy to propose as a follow-up, though.

Thanks. I was going to do the pushing of code into `OpenMPTarget_Instance` soon. 
I am working on making it more similar to OpenACC backend.

masterleinad:
> Thanks. I was going to do the pushing of code into `OpenMPTarget_Instance` soon.

Let me know which commits you want me to retain here.

rgayatri23:
> > Thanks. I was going to do the pushing of code into `OpenMPTarget_Instance` soon.
> 
> Let me know which commits you want me to retain here.

I think the current changes look good. If the CI passes, I am good.

core/src/OpenMPTarget/Kokkos_OpenMPTarget.hpp
```diff
@@ -31,6 +31,7 @@ static_assert(false,
 #include <cstddef>
 #include <iosfwd>
 #include <OpenMPTarget/Kokkos_OpenMPTargetSpace.hpp>
+//#include <OpenMPTarget/Kokkos_OpenMPTarget_Instance.hpp>
```
dalg24, line 34:
```suggestion
```

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Instance.cpp
```diff
@@ -27,11 +27,11 @@
 // constructor. undef'ed at the end
 #define KOKKOS_IMPL_OPENMPTARGET_WORKAROUND
 
+#include <Kokkos_Core.hpp>
 #include <OpenMPTarget/Kokkos_OpenMPTarget.hpp>
```
dalg24, line 31: This looks somewhat excessive.

core/src/OpenMPTarget/Kokkos_OpenMPTarget_Parallel.hpp
```diff
@@ -749,33 +749,28 @@ namespace Impl {
...
-  static uint32_t* m_uniquetoken_ptr;
-};
+  void* m_scratch_ptr = nullptr;
+  std::mutex m_mutex_scratch_ptr;
+  int64_t m_scratch_size      = 0;
+  uint32_t* m_uniquetoken_ptr = nullptr;
+};*/
 
 }  // namespace Impl
 }  // namespace Kokkos
```
dalg24, line 776: Remove instead of commenting

dalg24 (commented):
Did the scope of this PR change?
I don't necessarily have an issue with the new direction, just observing that it diverges from the PR title/description.

masterleinad:
> I don't necessarily have an issue with the new direction, just observing that it diverges from the PR title/description.

Updated.

rgayatri23 (approved):
OpenMPTarget build passes.

PR #6597, opened 2023-11-14
simd: add missing multiplication operators in neon simd

ldh4:
Caused by #6570.
Removal of the fallback implementation of `operator*` caused an invalid operand problem in NEON build.
This PR adds missing `operator*` in NEON using either NEON intrinsics or a fallback implementation using the generator constructor.

### DIFF
--- a/simd/src/Kokkos_SIMD_NEON.hpp
+++ b/simd/src/Kokkos_SIMD_NEON.hpp
@@ -868,7 +868,11 @@ class simd<std::int32_t, simd_abi::neon_fixed_size<2>> {
     return simd(
         vadd_s32(static_cast<int32x2_t>(lhs), static_cast<int32x2_t>(rhs)));
   }
-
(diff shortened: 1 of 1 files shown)


masterleinad (approved):

crtrott (approved):

crtrott:
Not sure whats going on with the threads backend, but don't think this is related here.

crtrott:
@ldh4 can you cherry pick on release-candidate

masterleinad:
> @ldh4 can you cherry pick on release-candidate

This was broken in #6570 which isn't part of the release candidate AFAICT.

ldh4:
> > @ldh4 can you cherry pick on release-candidate
> 
> This was broken in #6570 which isn't part of the release candidate AFAICT.

Yes, @masterleinad is right. #6570 isn't part of 4.2 release candidate. It was merged into `develop` after the release candidate was branched off. So cherry-picking this isn't necessary.
I just checked on an arm platform just to confirm, and this problem doesn't show up in the release candidate branch.

PR #6598, opened 2023-11-14
kokkos(unique): fix allocation of temporary view to enforce using the provided space instance and skip initialization 

romintomasetti:
This PR is extracted from #6584 as requested by @dalg24.

### DIFF
--- a/algorithms/src/std_algorithms/impl/Kokkos_Unique.hpp
+++ b/algorithms/src/std_algorithms/impl/Kokkos_Unique.hpp
@@ -105,7 +105,9 @@ IteratorType unique_exespace_impl(const std::string& label,
       // using the same algorithm used for unique_copy but we now move things
       using value_type    = typename IteratorType::value_type;
       using tmp_view_type = Kokkos::View<value_type*, ExecutionSpace>;
-      tmp_view_type tmp_view("std_unique_tmp_view", num_elements_to_explore);
+      tmp_view_type tmp_view(Kokkos::view_alloc(ex, Kokkos::WithoutInitializing,
+                                                "std_unique_tmp_view"),
+                             num_elements_to_explore);
 
       // scan extent is: num_elements_to_explore - 1
       // for same reason as the one explained in unique_copy


masterleinad (approved):

dalg24 (approved):

algorithms/src/std_algorithms/impl/Kokkos_Unique.hpp
```diff
@@ -105,7 +105,9 @@ IteratorType unique_exespace_impl(const std::string& label,
       // using the same algorithm used for unique_copy but we now move things
       using value_type    = typename IteratorType::value_type;
       using tmp_view_type = Kokkos::View<value_type*, ExecutionSpace>;
-      tmp_view_type tmp_view("std_unique_tmp_view", num_elements_to_explore);
+      tmp_view_type tmp_view(Kokkos::view_alloc(ex, Kokkos::WithoutInitializing,
```
dalg24, line 108: Please update the description and say this also skip initialization.

romintomasetti: Ok, done :wink:

fnrizzi: why do we want to skip the initialization?

romintomasetti: @fnrizzi I would say that for "scalar" types (a `double` for instance), it might not make a difference. For a custom type, default constructing it using the `ViewValueFunctor` might be a cost we want to avoid. Also note that when initializing, I think what happens is really backend-dependent and is driven by `ZeroMemset`.

All in all, there might be no noticeable performance improvement, but at least now the view is initialized correctly (I mean the following code does not require that the view is zero initialized).

fnrizzi: i was for a moment concerned about first touch allocation, but since the tmp_view is used immediately below in the par_scan without doing anything particular, i guess this is fine

fnrizzi (approved):

romintomasetti:
@dalg24 @masterleinad @fnrizzi Thanks for approving! I think one of needs to add "Ok to test" :smile:

fnrizzi:
> I think one of needs to add "Ok to test" 😄

the jenkins is temporarily down, that is why it is not running

fnrizzi:
> @fnrizzi Is it still down ? 😄

yes it is see here:
```
continuous-integration/jenkin-ci/pr Expected — Waiting for status to be reported
```

PR #6599, opened 2023-11-15
operator== for Kokkos::Array.

maartenarnst:
This PR adds `operator==` to `Kokkos::Array`. 

The implementation is outside the class and it is templated. This way, it can be used to check equality between the different specialisations of `Kokkos::Array` (primary, zero size, contiguous, strided).

### DIFF
--- a/core/src/Kokkos_Array.hpp
+++ b/core/src/Kokkos_Array.hpp
@@ -320,6 +320,26 @@ struct Array<T, KOKKOS_INVALID_INDEX, Array<>::strided> {
       : m_elem(arg_ptr), m_size(arg_size), m_stride(arg_stride) {}
 };
 
+template <typename T,
+          size_t LN, typename LP, 
+          size_t RN, typename RP>
+KOKKOS_INLINE_FUNCTION constexpr bool operator==(const Array<T, LN, LP>& lhs, 
+                                                 const Array<T, RN, RP>& rhs) {
+  if (lhs.size() != rhs.size()) return false;
+    for (int i = 0; i < (int)(lhs.size()); ++i) {
+      if (lhs[i] != rhs[i]) return false;
+    }
+    return true;
+}
+
+template <typename T,
+          size_t LN, typename LP, 
+          size_t RN, typename RP>
+KOKKOS_INLINE_FUNCTION constexpr bool operator!=(const Array<T, LN, LP>& lhs, 
+                                                 const Array<T, RN, RP>& rhs) {
+  return !(lhs == rhs);
+}
+
 }  // namespace Kokkos
 
 //<editor-fold desc="Support for structured binding">
--- a/core/src/Kokkos_Array.hpp
+++ b/core/src/Kokkos_Array.hpp
@@ -320,23 +320,20 @@ struct Array<T, KOKKOS_INVALID_INDEX, Array<>::strided> {
       : m_elem(arg_ptr), m_size(arg_size), m_stride(arg_stride) {}
 };
 
-template <typename T,
-          size_t LN, typename LP, 
-          size_t RN, typename RP>
-KOKKOS_INLINE_FUNCTION constexpr bool operator==(const Array<T, LN, LP>& lhs, 
-                                                 const Array<T, RN, RP>& rhs) {
-  if (lhs.size() != rhs.size()) return false;
-    for (int i = 0; i < (int)(lhs.size()); ++i) {
-      if (lhs[i] != rhs[i]) return false;
-    }
-    return true;
+template <typename T, size_t N>
+KOKKOS_INLINE_FUNCTION constexpr bool operator==(const Array<T, N, void>& lhs,
+                                                 const Array<T, N, void>& rhs) {
+  // The cast to int is necessary to avoid a warning about a pointless
+  // comparison with zero in the case that an unsigned type is used.
+  for (int i = 0; i < static_cast<int>(lhs.size()); ++i) {
+    if (lhs[i] != rhs[i]) return false;
+  }
+  return true;
 }
 
-template <typename T,
-          size_t LN, typename LP, 
-          size_t RN, typename RP>
-KOKKOS_INLINE_FUNCTION constexpr bool operator!=(const Array<T, LN, LP>& lhs, 
-                                                 const Array<T, RN, RP>& rhs) {
+template <typename T, size_t N>
+KOKKOS_INLINE_FUNCTION constexpr bool operator!=(const Array<T, N, void>& lhs,
+                                                 const Array<T, N, void>& rhs) {
   return !(lhs == rhs);
 }
 
--- a/core/unit_test/TestArrayOps.hpp
+++ b/core/unit_test/TestArrayOps.hpp
@@ -92,6 +92,84 @@ TEST(TEST_CATEGORY, array_element_access) {
   ASSERT_EQ(ca.data()[index], a[index]);
 }
 
+TEST(TEST_CATEGORY, array_operator_equal) {
+  using A = Kokkos::Array<int, 2>;
+  constexpr A a{{3, 5}};
+  constexpr A b{{3, 5}};
+  constexpr A c{{5, 3}};
+
+  static_assert(a == b);
+  static_assert(!(a == c));
+  static_assert(a != c);
+
+  ASSERT_TRUE(a == b);
+  ASSERT_FALSE(a == c);
+  ASSERT_TRUE(a != c);
+
+  using E = Kokkos::Array<int, 0>;
+  constexpr E e;
+  constexpr E f;
+
+  static_assert(e == f);
+  static_assert(!(e != f));
+
+  ASSERT_TRUE(e == f);
+  ASSERT_FALSE(e != f);
+
+  static_assert(a != e);
+
+  ASSERT_TRUE(a != e);
+
+  using AC =
+    Kokkos::Array<int, KOKKOS_INVALID_INDEX, Kokkos::Array<>::contiguous>;
+
+  int ac_[] = {3, 5};
+  const AC ac(ac_, std::size(ac_));
+  int bc_[] = {3, 5};
+  const AC bc(bc_, std::size(bc_));
+
+  int dc_[] = {11, 13, 17, 19};
+  const AC dc(dc_, std::size(dc_));
+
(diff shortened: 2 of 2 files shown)


maartenarnst:
Hi @fnrizzi. Thanks for looking at the PR. Sorry for the test failures. I have corrected the assert that led to the test failures. 

It's a very small PR, so I'm not sure it's worth it to convert it to draft though. If the tests fail again, I'll be sure to do it. If you prefer to make it a draft regardless, please just let me know.

fnrizzi:
I think it is good , you fixed it right away so no need to change!
I was just mentioning that because, in general, if a pr is marked ready I/we assume it is ready for review so the last kind of implies the CI should be passing already.

core/src/Kokkos_Array.hpp
```diff
@@ -320,6 +320,24 @@ struct Array<T, KOKKOS_INVALID_INDEX, Array<>::strided> {
       : m_elem(arg_ptr), m_size(arg_size), m_stride(arg_stride) {}
 };
 
+template <typename LT, size_t LN, typename LP, 
+          typename RT, size_t RN, typename RP>
+KOKKOS_INLINE_FUNCTION constexpr bool operator==(const Array<LT, LN, LP>& lhs, 
+                                                 const Array<RT, RN, RP>& rhs) {
```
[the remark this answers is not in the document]
fnrizzi: i would vote to be consistent with the c++ one.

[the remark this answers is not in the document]
maartenarnst: Indeed, it's probably best to implement it for the same value type only. I've just made change.

maartenarnst:
Some of the builds with the intel compiler failed because `for (size_t i = 0; i < lhs.size(); ++i) {` led to a "pointless comparison of an unsigned type with 0" warning for the specialisation of `Kokkos::Array` of zero size. I've solved it by changing the line to `for (int i = 0; i < (int)(lhs.size()); ++i) {`, similarly to what was done in PR #6423.

core/src/Kokkos_Array.hpp
```diff
@@ -320,6 +320,26 @@ struct Array<T, KOKKOS_INVALID_INDEX, Array<>::strided> {
       : m_elem(arg_ptr), m_size(arg_size), m_stride(arg_stride) {}
 };
 
+template <typename T,
+          size_t LN, typename LP, 
+          size_t RN, typename RP>
+KOKKOS_INLINE_FUNCTION constexpr bool operator==(const Array<T, LN, LP>& lhs, 
+                                                 const Array<T, RN, RP>& rhs) {
```
masterleinad, line 327: `std::array` also doesn't allow compring `std::array`s with different numbers of elements, see https://godbolt.org/z/Y5b3fzGfq.
```suggestion
template <typename T,
          size_t N, typename LP, 
          typename RP>
KOKKOS_INLINE_FUNCTION constexpr bool operator==(const Array<T, N, LP>& lhs, 
                                                 const Array<T, N, RP>& rhs) {
```
I'm not quite sure about the `Proxy` argument, though.

maartenarnst: I'm not sure what's best to do.

The four versions of `Kokkos::Array` are:

```
template <class T = void, size_t N = KOKKOS_INVALID_INDEX, class Proxy = void> 
struct Array;

template <class T, class Proxy> 
struct Array<T, 0, Proxy>;

template <class T> 
struct Array<T, KOKKOS_INVALID_INDEX, Array<>::contiguous>;

template <class T>
struct Array<T, KOKKOS_INVALID_INDEX, Array<>::strided>;
```

And so the `Proxy` influences `N`. I.e., for the last two specialisations (wraps of dynamically allocated arrays), the size is set to the invalid value. If we limit `operator==` to only comparisons of equal size, I think we won't be able to compare `Kokkos::Array` that wrap a statically allocated array with `Kokkos::Array` that wrap a dynamically allocated array.

The fact that the `Kokkos::Array` can wrap a dynamically allocated array seems to be a difference with the `std::Array`. So to me, it's OK that there is also a difference in what `operator==` can do. But, if you prefer, I can update the implementation and limit it to comparisons of the same size.

maartenarnst: Hi @masterleinad. I just wanted to ask if you should have had a chance to think about what would be best to do? Should we limit the comparison to the same size to be consistent with the std::array as you had suggested? Or should we allow comparisons between kokkos::array with different size template parameter to allow comparisons of kokkos::array of different proxy template parameter, such as static and dynamic ones? Thanks in advance.

masterleinad: Ideally, I would want to eliminate the `Proxy` template argument but that's a different discussion. Unless you have a use case where you need it, I would still prefer to follow the restrictions set by `std::array` and only provide the comparison operator if `Proxy==void`. It's easier to relax it later than to restrict it.

maartenarnst: OK. I've just revised it as you suggest. The comparison is now only provided if the arrays have equal size and `Proxy==void`. 

(Just a thought: in a discussion about removing `Proxy`, perhaps the possibility of replacing the statically sized `Kokkos::Array` with `std::Array` could also be explored; in our codes, `std::Array` seems to work at least in HIP and Cuda kernels, but I don't know about the other backends).

masterleinad (approved):

maartenarnst:
Hi @fnrizzi. Would you have a moment to take another look? Thanks in advance.

core/src/Kokkos_Array.hpp
```diff
@@ -320,6 +320,21 @@ struct Array<T, KOKKOS_INVALID_INDEX, Array<>::strided> {
       : m_elem(arg_ptr), m_size(arg_size), m_stride(arg_stride) {}
 };
 
+template <typename T, size_t N>
+KOKKOS_INLINE_FUNCTION constexpr bool operator==(const Array<T, N, void>& lhs,
+                                                 const Array<T, N, void>& rhs) {
+  for (int i = 0; i < (int)(lhs.size()); ++i) {
```
fnrizzi, line 326: i would prefer we use the size_type here instead of int

fnrizzi: yes but that was for casting a var looping over at most Rank , here we are looping over the whole array. 
i realize this in practice will not be an issue, but generally IMO the fewer casts the better. 
ok i think i am fine with this

fnrizzi, line 326: can you please just add a comment here to say that the casting is necessary to avoid the pointless comparison warning?

maartenarnst: Voila voila. I've just added the comment.

core/src/Kokkos_Array.hpp
```diff
@@ -320,6 +320,23 @@ struct Array<T, KOKKOS_INVALID_INDEX, Array<>::strided> {
       : m_elem(arg_ptr), m_size(arg_size), m_stride(arg_stride) {}
 };
 
+template <typename T, size_t N>
+KOKKOS_INLINE_FUNCTION constexpr bool operator==(const Array<T, N, void>& lhs,
+                                                 const Array<T, N, void>& rhs) {
+  // The cast to int is necessary to avoid a warning about a pointless
+  // comparison with zero in the case that an unsigned type is used.
+  for (int i = 0; i < (int)(lhs.size()); ++i) {
```
nliber, line 328: Do you still get the pointless comparison warning if you use `!=` instead of `<`?  `!=` is more idiomatic C++ anyway.

If you still have to cast, use `static_cast` instead of the C-style cast.

Also, it really should be a hidden `friend` instead of a free function in the namespace.

nliber (changes requested):

core/src/Kokkos_Array.hpp
maartenarnst: Hi @nliber. Thanks for the review and suggestions! And sorry for replying with some delay.

- The pointless comparison warning disappears if I use `!=` instead of `<`. That's interesting. For now, I've kept the use of `<` and the cast to remain consistent with what was done in PR #6423.
- I've replaced the C-style cast with a `static_cast`.
- I'm not sure how to best make it work as a hidden `friend`. An issue is that the `Kokkos::Array` has several specialisations. And we want the equality operator to work with both the specialisation for a fixed (non-zero) size and the one for the fixed zero size. So, if I move the equality operator inside the class definition, I think I'll have to define it twice. Also, because the equality operator uses public member functions, I think it wouldn't need to be a friend.

nliber: Hidden `friend`s are about reducing the size of overload sets (faster compiles) by only using ADL (argument dependent lookup) and avoiding accidental implicit conversions.  It is conflated into the straightforward usage of `friend` (granting access to non-public members of a class).

Anthony Williams has a good article on hidden friends at [https://www.justsoftwaresolutions.co.uk/cplusplus/hidden-friends.html](https://www.justsoftwaresolutions.co.uk/cplusplus/hidden-friends.html).

Here are some examples:

```c++
class X {
    // hidden friend
    friend bool operator==(X const&, X const&) { return true; }
};
// Not hidden friend
inline bool operator!=(X const& lhs, X const& rhs) { return !(lhs == rhs); }

struct Y {
    operator X() const { return X(); }  // implicit conversion to X
};

X x;
Y y;
assert(  x == x );
assert(!(x != x));
assert(  x == y );
assert(!(x != y));
assert(  y == x );
assert(!(y != x));
assert(  y == y );  // error: no match for 'operator==' (operand types are 'Y' and 'Y')
assert(!(y != y));
```

```c++
class X {
    // hidden friend
    friend bool operator==(X const&, X const&) { return true; }
};
// Not hidden friend
inline bool operator!=(X const& lhs, X const& rhs) { return !(lhs == rhs); }

struct Y {
    operator X() const { return X(); }  // implicit conversion to X
};

X x;
Y y;
  operator==(x, x);
::operator==(x, x); // error: no member named 'operator==' in the global namespace
  operator!=(x, x);
::operator!=(x, x);
```

To be a hidden `friend`, it has to be the first declaration of a function and that declaration must contain the definition.  You'll have to repeat it in every specialization.

It also needs to be a `friend` because that is the only way to declare a non-member function inside a class.

dalg24:
Superseded by #7148

PR #6601, opened 2023-11-15
Threads: add missing broadcast to TeamThreadRange parallel_scan

fnrizzi:
Fixes some of the failures for the threads backend seen for example in https://github.com/kokkos/kokkos/actions/runs/6881218385/job/18717130472?pr=6600

Related to https://github.com/kokkos/kokkos/issues/6470

PR #6604, opened 2023-11-16
Fix threads loop in `Kokkos_CoreUnitTest_DeviceAndThreads`

fnrizzi:
after #6601 will be merged, the last (currently) issue with the threads CI remaining is: 
```
test_device_id (TestDeviceAndThreads.KokkosInitializationTestCase) ... skipped 'no device detected'
test_disable_warnings (TestDeviceAndThreads.KokkosInitializationTestCase) ... ok
test_num_threads (TestDeviceAndThreads.KokkosInitializationTestCase) ... ERROR
test_tune_internals (TestDeviceAndThreads.KokkosInitializationTestCase) ... ok

======================================================================
ERROR: test_num_threads (TestDeviceAndThreads.KokkosInitializationTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/__w/kokkos/kokkos/builddir/core/unit_test/TestDeviceAndThreads.py", line 47, in test_num_threads
    GetFlag(
  File "/__w/kokkos/kokkos/builddir/core/unit_test/TestDeviceAndThreads.py", line 29, in GetFlag
    raise Exception(p.stderr.decode("utf-8"))
Exception: terminate called after throwing an instance of 'std::runtime_error'
  what():  Kokkos::Threads::initialize HWLOC ERROR(s) : thread_count(3) imbalanced among cores(2)
```

It is hitting the constraint imposed internally on the imbalance. 
Currently, the loop over threads in the test is
```py
def GetNumThreads(max_threads):
    for x in [1, 2, 3, 5, 7]:
        if x >= max_threads:
            break
        yield x
    yield max_threads
# ...

def test_num_threads(self):
    max_threads = GetFlag("max_threads")
    if max_threads == 1:
        self.skipTest("no host parallel backend enabled")
    for num_threads in GetNumThreads(max_threads):
        val = GetFlag("num_threads", "--kokkos-num-threads={}".format(num_threads))
        self.assertEqual(num_threads, GetFlag("num_threads", "--kokkos-num-threads={}".format(num_threads)))
```       

# Solution (also based on comments below) 

I am using now a loop over threads based on the physical core count like so:
```
looplist = [1] + [i*phys_cores_count for i in [1,2,3,4,5,6,7]]
```
The phys core count is detected via system calls to the running OS.

Note: we could just use the python package `psutil` via `phys_cores_count = psutil.cpu_count(logical=False)` but doing that triggered an error because that package was not found in the python installation.

### DIFF
--- a/core/unit_test/TestDeviceAndThreads.py
+++ b/core/unit_test/TestDeviceAndThreads.py
@@ -30,7 +30,7 @@ def GetFlag(flag, *extra_args):
     return int(p.stdout)
 
 def GetNumThreads(max_threads):
-    for x in [1, 2, 3, 5, 7]:
+    for x in [1, 2, 4, 6, 8]:
         if x >= max_threads:
             break
         yield x
--- a/core/unit_test/TestDeviceAndThreads.py
+++ b/core/unit_test/TestDeviceAndThreads.py
@@ -17,7 +17,7 @@
 
 import unittest
 import subprocess
-import psutil
+import os #psutil
 
 PREFIX = "$<TARGET_FILE_DIR:Kokkos_CoreUnitTest_DeviceAndThreads>"
 EXECUTABLE = "$<TARGET_FILE_NAME:Kokkos_CoreUnitTest_DeviceAndThreads>"
@@ -31,7 +31,16 @@ def GetFlag(flag, *extra_args):
     return int(p.stdout)
 
 def GetNumThreads(max_threads):
-    phys_cores_count = psutil.cpu_count(logical=False)
+    #phys_cores_count = psutil.cpu_count(logical=False)
+    args = ['sysctl', '-n', 'hw.physicalcpu_max']
+    if os.name == 'nt':
+        args = ['wmic', 'cpu', 'get', 'NumberOfCores']
+
+    result = subprocess.run(args, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    output = result.stdout.decode('utf-8')
+    phys_cores_count = int(output)
+    print(phys_cores_count)
+
     looplist = [1] + [i*phys_cores_count for i in [1,2,3,4,5,6,7]]
     for x in looplist:
         if x >= max_threads:
--- a/core/unit_test/TestDeviceAndThreads.py
+++ b/core/unit_test/TestDeviceAndThreads.py
@@ -17,7 +17,7 @@
 
 import unittest
 import subprocess
-import platform #psutil
+import platform
 
 PREFIX = "$<TARGET_FILE_DIR:Kokkos_CoreUnitTest_DeviceAndThreads>"
 EXECUTABLE = "$<TARGET_FILE_NAME:Kokkos_CoreUnitTest_DeviceAndThreads>"
@@ -31,7 +31,6 @@ def GetFlag(flag, *extra_args):
     return int(p.stdout)
 
 def GetNumThreads(max_threads):
-    #phys_cores_count = psutil.cpu_count(logical=False)
     args = []
     name = platform.system()
     if name == 'Darwin':
--- a/core/unit_test/TestDeviceAndThreads.py
+++ b/core/unit_test/TestDeviceAndThreads.py
@@ -32,7 +32,7 @@ def GetFlag(flag, *extra_args):
 
 def GetNumThreads(max_threads):
     #phys_cores_count = psutil.cpu_count(logical=False)
-    args = ['sysctl', '-n', 'hw.physicalcpu_max']
+    args = ['nproc', '--all'] #'sysctl', '-n', 'hw.physicalcpu_max']
     if os.name == 'nt':
         args = ['wmic', 'cpu', 'get', 'NumberOfCores']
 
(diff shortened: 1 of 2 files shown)


Rombur:
We should relax the imbalance check. If tomorrow the CI has access to three cores, we will get the same issue again.

fnrizzi:
> We should relax the imbalance check. If tomorrow the CI has access to three cores, we will get the same issue again.

It seems to me the imbalance check is quite an important constraint to have, why getting rid of it? Are you suggesting to replace it wirh a warning?

fnrizzi:
Ok so if you think that mapping is acceptable, you vote for removing the check right? IMO we should either keep the check for imbalance and trigger like we do now, or relax it but still warn that the mapping is not optimal

Rombur:
> Ok so if you think that mapping is acceptable, then you vote for removing the check right?

Not really. We should still check that we don't have a situation where we have three threads on one core and one thread on the other core. It's just that if it's impossible to have perfect balance, we should accept an imbalance of one thread.

fnrizzi:
Ah I see, I thought you were ok with any imbalance!  
You are ok with imbalance of 1.

Rombur:
>You are ok with imbalance of 1.

Correct. I think if we say that the imbalance is at most 1, we will get an error if there is a balanced solution and we don't use. If there is no perfectly balanced solution, we will have the next best thing.

fnrizzi:
IMO, i think we should not allow an imbalance. If the user sets the num threads to cause the imbalance, i think it easy for them to just set it correctly. IMO, doing this and so keeping the constraint is safer than just relaxing it since this has performance implications. But maybe someone else wants to pitch in on this

(as far as the test, i can work around it to guard for odd/even cores by querying the # of cores and deciding the loop based on that)

crtrott:
Lets modify the test range, and only test values which would be balanced for now. Is it hard to do that?

fnrizzi:
> Lets modify the test range, and only test values which would be balanced for now. Is it hard to do that?

yes, so basically conditioning the loop on even/odd cases. let me try

fnrizzi:
failure is unrelated:
```
  program.  Verify that Intel:2021.10.0.20230609 is set up correctly (e.g.,
  check that correct library headers are being used).
```

core/unit_test/TestDeviceAndThreads.py
```diff
@@ -30,7 +31,20 @@ def GetFlag(flag, *extra_args):
...
+        args = ['sysctl', '-n', 'hw.physicalcpu_max']
+    elif name == 'Linux':
+        args = ['nproc', '--all']
+    else:
+        args = ['wmic', 'cpu', 'get', 'NumberOfCores']
+
+    result = subprocess.run(args, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+    output = result.stdout.decode('utf-8')
+    phys_cores_count = int(output)
+    looplist = [1] + [i*phys_cores_count for i in [1,2,3,4,5,6,7]]
```
crtrott, line 46: Does that mean for 2 cores you test 1, 2, 4, 6 etc?

I explicitly want to test uneven number of threads at least on our nightly platforms maybe we should just do a loop from 1 to min(core_count, 8) or so?

crtrott (changes requested):
I don't want to run into a situation where we never test uneven number of threads because all our platforms happen to have even number of cores.

core/unit_test/TestDeviceAndThreads.py
fnrizzi: Yes that is what the loop does. 
But for numcores = 3 we get odd numbers too.  

If we test odd numbers of threads if numcores = 2, the test triggers the imbalance.

fnrizzi: Also, it seems to me this test aims at testing that what is passed via num-threads actually matches the hat is used internally, so why is my solution not general enough?

masterleinad: I think we should just make sure that we test with an uneven number of threads in a build that doesn't use `hwloc`.

fnrizzi: how about this solution: https://github.com/kokkos/kokkos/pull/6604/commits/374064ab75aa4fd1727781b01c8c1724470788ac?

masterleinad (approved):
Looks OK to me.

crtrott (approved):

core/unit_test/TestDeviceAndThreads.py
```diff
@@ -30,7 +31,22 @@ def GetFlag(flag, *extra_args):
...
 def GetNumThreads(max_threads):
-    for x in [1, 2, 3, 5, 7]:
+    args = []
+    name = platform.system()
+    if name == 'Darwin':
+        args = ['sysctl', '-n', 'hw.physicalcpu_max']
+    elif name == 'Linux':
+        args = ['nproc', '--all']
+    else:
+        args = ['wmic', 'cpu', 'get', 'NumberOfCores']
```
dalg24, line 41: Why is `os.cpu_count()` no sufficient?
https://docs.python.org/3/library/os.html#os.cpu_count

fnrizzi: That gives us the logical count, while above I am using the physical count. I figured it was better to rely on the physical core count.

Issue #6605, opened 2023-11-16
Problem using gnu_generate_makefile.sh when GNU make version is >= 4.3


pkestene:
**Describe the bug**
 
When trying to build kokkos using `gnu_generate_makefile.bash`, some generated unit test source files are ill-formed due to a change of behavior in GNU make version 4.3. E.g. when the following code gets expanded by Makefile
```make
# this works with make 4.2 but not with 4.3
$(shell echo "\#include <TestSort.hpp>" >> Test.cpp)
```
the slash gets printed in the output `Test.cpp`

One solution to fix the problem is mentioned e.g. here:
https://github.com/CEED/libCEED/blob/6dbc39456a768b0ea4055927c94c474873ea8bf6/common.mk#L34

just use this instead (it works for all version of make, I tested 4.2 and 4.3)
```make
HASH=\#
$(shell echo "$(HASH)include <TestSort.hpp>" >> Test.cpp)
```

**Please also include the following items to support reproducing the bug**

To reproduce, e.g.:
```shell 
mkdir _build
cd _build
../gnu_generate_makefile.bash --with_devices=OpenMP
make
g++ -I./ -I/data/pkestene/install/trilinos/kokkos/github/kokkos_test/core/src -I/data/pkestene/install/trilinos/kokkos/github/kokkos_test/containers/src -I/data/pkestene/install/trilinos/kokkos/github/kokkos_test/algorithms/src -I/data/pkestene/install/trilinos/kokkos/github/kokkos_test/tpls/desul/include -I/data/pkestene/install/trilinos/kokkos/github/kokkos_test/tpls/mdspan/include -std=c++17 -fopenmp -I./ -I/data/pkestene/install/trilinos/kokkos/github/kokkos_test/core/src -I/data/pkestene/install/trilinos/kokkos/github/kokkos_test/containers/src -I/data/pkestene/install/trilinos/kokkos/github/kokkos_test/algorithms/src -I/data/pkestene/install/trilinos/kokkos/github/kokkos_test/tpls/desul/include -I/data/pkestene/install/trilinos/kokkos/github/kokkos_test/tpls/mdspan/include -I/data/pkestene/install/trilinos/kokkos/github/kokkos_test/tpls/gtest -I/data/pkestene/install/trilinos/kokkos/github/kokkos_test/core/unit_test -I/data/pkestene/install/trilinos/kokkos/github/kokkos_test/core/unit_test/category_files -O3  -c TestOpenMP_Init.cpp
[...]

PR #2972, opened 2020-04-24
Fix Makefile.kokkos compat with make 4.3

pgrete:
GNU make 4.3 introduced a backward incompatible change with version 4.3, see https://lists.gnu.org/archive/html/info-gnu/2020-01/msg00004.html
```
* WARNING: Backward-incompatibility!
  Number signs (#) appearing inside a macro reference or function invocation
  no longer introduce comments and should not be escaped with backslashes:
  thus a call such as:
    foo := $(shell echo '#')
  is legal.  Previously the number sign needed to be escaped, for example:
    foo := $(shell echo '\#')
  Now this latter will resolve to "\#".  If you want to write makefiles
  portable to both versions, assign the number sign to a variable:
    H := \#
    foo := $(shell echo '$H')
```

With these changes I can successfully use make 4.3 and <4.3 again.

Addresses #2870

PR #6606, opened 2023-11-16
Fix Makefile when using gnu_generate_makefile.sh and make >= 4.3

pkestene:
Fix issue #6605


masterleinad (approved):
This is similar to what we are doing in `Makefile.kokkos`.

Rombur (approved):

dalg24:
> This is similar to what we are doing in `Makefile.kokkos`.

Daniel is referring to #2972
Should we should use `$H` then instead of introducing a new variable for it?
https://github.com/kokkos/kokkos/blob/04bdcd247673eb9ed7afdc0f25005f7f6419063e/Makefile.kokkos#L462-L463

pkestene:
@dalg24 
Sorry, I was not aware that a 'hash' variable was already defined.
I can remove definition of "HASH" use "H" instead

Issue #5980, opened 2023-03-12
Different behavior when using Kokkos::SharedSpace and Kokkos::CudaSpace for virtual functions.


wangxi0706:
I am wrapping derived class as shown in doc (https://kokkos.github.io/kokkos-core-wiki/ProgrammingGuide/Kokkos-and-Virtual-Functions.html?highlight=virtual+function). When I am using ShareSpace or CUDAUVM Space,

wangxi0706:
When usinng  ShareSpace or CUDAUVM Space, Code and output is like this.
int main(int argc, char *argv[])
{
  Kokkos::initialize(argc, argv);
  Kokkos::DefaultExecutionSpace{}.print_configuration(std::cout);
  {
    Kokkos::View<Foo_1[1], Kokkos::SharedSpace> f_1("f_1");
    Kokkos::View<Foo_2[1], Kokkos::SharedSpace> f_2("f_1");
    int value_1, value_2;
    Kokkos::parallel_reduce(
        "CheckValues", 1,
        KOKKOS_LAMBDA(const int &, int &lsum) { Foo *pf_1=&f_1[0]; lsum = pf_1->value(); }, value_1);
    Kokkos::parallel_reduce(
        "CheckValues", 1,
        KOKKOS_LAMBDA(const int &, int &lsum) { Foo *pf_2=&f_2[0]; lsum = pf_2->value(); }, value_2);
    printf("Values: %i %i\n", value_1, value_2);
  }
  Kokkos::finalize();
}

And the output reads as follows:
Device Execution Space:
  KOKKOS_ENABLE_CUDA: yes
Cuda Atomics:
  KOKKOS_ENABLE_CUDA_ATOMICS: no
Cuda Options:
  KOKKOS_ENABLE_CUDA_LAMBDA: yes
  KOKKOS_ENABLE_CUDA_LDG_INTRINSIC: yes
  KOKKOS_ENABLE_CUDA_RELOCATABLE_DEVICE_CODE: yes
  KOKKOS_ENABLE_CUDA_UVM: yes
  KOKKOS_ENABLE_CXX11_DISPATCH_LAMBDA: yes
Cuda Runtime Configuration:
macro  KOKKOS_ENABLE_CUDA      : defined
macro  CUDA_VERSION          = 12000 = version 12.0
[...]

PR #6610, opened 2023-11-20
unorderedmap: missing `space` instance for erasure and rehash ops

romintomasetti:
This PR is extracted from #6584 as requested by @dalg24.

It adds a `space` instance argument to map erasure and rehash operators for `Kokkos::UnorderedMap`.

~~It also makes `invalid_index` public, otherwise it cannot be used by client code, though the `invalid_index` is used *e.g.* by `Kokkos::UnorderedMap::find`.~~ Topic moved to #6641.

### DIFF
--- a/containers/src/Kokkos_Bitset.hpp
+++ b/containers/src/Kokkos_Bitset.hpp
@@ -130,10 +130,11 @@ class Bitset {
 
   /// set all bits to 1
   /// can only be called from the host
-  void set() {
-    Kokkos::deep_copy(m_blocks, ~0u);
+  void set(const execution_space& exec) {
+    Kokkos::deep_copy(exec, m_blocks, ~0u);
 
     if (m_last_block_mask) {
+      exec.fence("todo message");
       // clear the unused bits in the last block
       Kokkos::Impl::DeepCopy<typename Device::memory_space, Kokkos::HostSpace>(
           m_blocks.data() + (m_blocks.extent(0) - 1u), &m_last_block_mask,
@@ -146,11 +147,11 @@ class Bitset {
 
   /// set all bits to 0
   /// can only be called from the host
-  void reset() { Kokkos::deep_copy(m_blocks, 0u); }
+  void reset(const execution_space& exec) { Kokkos::deep_copy(exec, m_blocks, 0u); }
 
   /// set all bits to 0
   /// can only be called from the host
-  void clear() { Kokkos::deep_copy(m_blocks, 0u); }
+  void clear(const execution_space& exec) { Kokkos::deep_copy(exec, m_blocks, 0u); }
 
   /// set i'th bit to 1
   /// can only be called from the device
--- a/containers/src/Kokkos_UnorderedMap.hpp
+++ b/containers/src/Kokkos_UnorderedMap.hpp
@@ -380,21 +380,30 @@ class UnorderedMap {
   histogram_type get_histogram() { return histogram_type(*this); }
 
   //! Clear all entries in the table.
-  void clear() {
+  void clear(const execution_space& exec) {
     m_bounded_insert = true;
 
     if (capacity() == 0) return;
 
-    m_available_indexes.clear();
+    m_available_indexes.clear(exec);
 
-    Kokkos::deep_copy(m_hash_lists, invalid_index);
-    Kokkos::deep_copy(m_next_index, invalid_index);
+    Kokkos::deep_copy(exec, m_hash_lists, invalid_index);
+    Kokkos::deep_copy(exec, m_next_index, invalid_index);
     {
       const key_type tmp = key_type();
-      Kokkos::deep_copy(m_keys, tmp);
+      Kokkos::deep_copy(exec, m_keys, tmp);
     }
-    Kokkos::deep_copy(m_scalars, 0);
+    Kokkos::deep_copy(exec, m_scalars, 0);
     m_size() = 0;
+  }
+
+  //! @overload
+  void clear() {
+    Kokkos::fence(
+        "Kokkos::UnorderedMap::clear: fence before");
+    execution_space exec {};
+    clear(exec);
+    exec.fence("Kokkos::UnorderedMap::clear: fence after");
   }
 
   KOKKOS_INLINE_FUNCTION constexpr bool is_allocated() const {
@@ -412,24 +421,44 @@ class UnorderedMap {
   ///
   /// This is <i>not</i> a device function; it may <i>not</i> be
   /// called in a parallel kernel.
-  bool rehash(size_type requested_capacity = 0) {
+  bool rehash(const size_type requested_capacity = 0) {
+    Kokkos::fence(
+        "Kokkos::UnorderedMap::rehash: fence before since no execution space instance provided");
+    execution_space exec {};
+    bool result = rehash(exec, requested_capacity);
+    exec.fence("Kokkos::UnorderedMap::rehash: fence after since no execution space instance provided");
+    return result;
+  }
+
+  bool rehash(const execution_space &exec, const size_type requested_capacity = 0) {
     const bool bounded_insert = (capacity() == 0) || (size() == 0u);
-    return rehash(requested_capacity, bounded_insert);
+    return rehash(exec, requested_capacity, bounded_insert);
   }
 
   bool rehash(size_type requested_capacity, bool bounded_insert) {
-    if (!is_insertable_map) return false;
+    Kokkos::fence(
+        "Kokkos::UnorderedMap::rehash: fence before since no execution space instance provided");
+    execution_space exec{};
+    bool result = rehash(exec, requested_capacity, bounded_insert);
+    exec.fence("Kokkos::UnorderedMap::rehash: fence after since no execution space instance provided");
+    return result;
+  }
+
+  bool rehash(const execution_space &exec, size_type requested_capacity,
+              bool bounded_insert) {
+    if constexpr (!is_insertable_map) return false;
 
     const size_type curr_size = size();
     requested_capacity =
         (requested_capacity < curr_size) ? curr_size : requested_capacity;
 
-    insertable_map_type tmp(requested_capacity, m_hasher, m_equal_to);
+    insertable_map_type tmp(Kokkos::view_alloc(exec, "Kokkos::UnorderedMap::rehash: temporary view"), requested_capacity,
+                            m_hasher, m_equal_to);
 
     if (curr_size) {
       tmp.m_bounded_insert = false;
-      Impl::UnorderedMapRehash<insertable_map_type> f(tmp, *this);
-      f.apply();
+      Impl::UnorderedMapRehash<insertable_map_type> f{tmp, *this};
+      f.apply(exec);
     }
     tmp.m_bounded_insert = bounded_insert;
 
@@ -466,26 +495,37 @@ class UnorderedMap {
   }
 
   bool begin_erase() {
+    Kokkos::fence(
+        "Kokkos::UnorderedMap::begin_erase: fence before since no execution space instance provided");
+    execution_space exec {};
+    bool result = begin_erase(exec);
+    exec.fence("Kokkos::UnorderedMap::begin_erase: fence after since no execution space instance provided");
+    return result;
+  }
+
+  bool begin_erase(const execution_space &exec) {
     bool result = !erasable();
     if (is_insertable_map && result) {
-      execution_space().fence(
-          "Kokkos::UnorderedMap::begin_erase: fence before setting erasable "
-          "flag");
-      set_flag(erasable_idx);
+      set_flag(exec, erasable_idx);
     }
     return result;
   }
 
   bool end_erase() {
+    Kokkos::fence(
+        "Kokkos::UnorderedMap::end_erase: fence before");
+    execution_space exec {};
+    bool result = end_erase(exec);
+    exec.fence("Kokkos::UnorderedMap::end_erase: fence after");
+    return result;
+  }
+
+  bool end_erase(const execution_space &exec) {
     bool result = erasable();
     if (is_insertable_map && result) {
-      execution_space().fence(
-          "Kokkos::UnorderedMap::end_erase: fence before erasing");
-      Impl::UnorderedMapErase<declared_map_type> f(*this);
-      f.apply();
-      execution_space().fence(
-          "Kokkos::UnorderedMap::end_erase: fence after erasing");
-      reset_flag(erasable_idx);
+      Impl::UnorderedMapErase<declared_map_type> f{*this};
+      f.apply(exec);
+      reset_flag(exec, erasable_idx);
     }
     return result;
   }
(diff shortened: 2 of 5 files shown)


romintomasetti:
@dalg24 @fnrizzi @masterleinad Would you have a minute to look at this one? Thanks!

containers/src/impl/Kokkos_UnorderedMap_impl.hpp
```diff
@@ -223,16 +222,14 @@ struct UnorderedMapPrint {
 
   map_type m_map;
 
-  UnorderedMapPrint(map_type const& map) : m_map(map) {}
```
masterleinad, line 226: Why did you change?

containers/src/Kokkos_UnorderedMap.hpp
```diff
@@ -243,16 +243,19 @@ class UnorderedMap {
...
-  static const bool is_insertable_map =
+  static constexpr bool is_insertable_map =
       !has_const_key && (is_set || !has_const_value);
-  static const bool is_modifiable_map = has_const_key && !has_const_value;
-  static const bool is_const_map      = has_const_key && has_const_value;
+  static constexpr bool is_modifiable_map = has_const_key && !has_const_value;
+  static constexpr bool is_const_map      = has_const_key && has_const_value;
+
+  static constexpr size_type invalid_index =
+      KOKKOS_INVALID_INDEX_TYPE(size_type);
```
masterleinad, line 258: Why is it important for `invalid_index` to be public? How would it be used in user code?

containers/src/impl/Kokkos_UnorderedMap_impl.hpp
```diff
@@ -76,19 +75,19 @@ struct UnorderedMapErase {
   using size_type       = typename map_type::size_type;
   using key_type        = typename map_type::key_type;
   using value_type      = typename map_type::impl_value_type;
+  using policy_type =
+      Kokkos::RangePolicy<execution_space, IndexType<size_type>>;
 
   map_type m_map;
 
-  UnorderedMapErase(map_type const& map) : m_map(map) {}
```
masterleinad, line 82: Why did you remove?

romintomasetti: 1. `UnorderedMapErase` is an aggregate and as such can be initialized using aggregate initialization.
2. In my opinion, the constructor was not correct in the sense that, bearing in mind that the map is a "shared pointer to some memory", it should have been `UnorderedMapErase(map_type map) : m_map(std::move(map)) {}` to clearly show that shared ownership is requested.

All in all I decided to remove the constructor :pencil2:

containers/src/Kokkos_UnorderedMap.hpp
romintomasetti: In one of my tests, I do:
```c++
using map_t = Kokkos::UnorderedMap<int, double, execution_space>;

...

map_t map(...);

...

Kokkos::parallel_for(
    ...,
    KOKKOS_LAMBDA(...) {

        ....

        if(const auto direct_index = map.find(my_key); direct_index != ~static_cast<typename umap_t::size_type>(0))
        {
            <use `direct_index` here>
        }
    }
);

...
```

So in this case it would be nice to have access to `invalid_index`, which in fact is a value that can be returned by `find`, which is **a public member function**. IMO we don't even need a use case to decide if `invalid_index` should be public. I think it should because it can be returned by a public member function.

containers/src/impl/Kokkos_UnorderedMap_impl.hpp
romintomasetti: For the same reason as for `UnorderedMapErasure`, see my comment https://github.com/kokkos/kokkos/pull/6610#discussion_r1402100981 #5980

containers/src/impl/Kokkos_UnorderedMap_impl.hpp
romintomasetti: @masterleinad What's your opinion on this then ? :)

masterleinad: My preference is to minimize drive-by changes. 🙂

romintomasetti: @masterleinad I already extracted these changes from #6584... I'm not against making another "extracted from" PR with the "remove useless constructors" change if you really want to and in general I agree with the "one PR per topic without drive-by change", but it makes the contribution process very lengthy for such a "deletion of useless line of code change" case... Especially that in this PR I already modify the `UnorderedMapErase` & Co. functors, so deleting their useless constructor can arguably not be a drive-by change :angel:

What do you think?

containers/src/Kokkos_UnorderedMap.hpp
```diff
@@ -465,26 +476,24 @@ class UnorderedMap {
...
+  bool end_erase(const execution_space &space = execution_space{}) {
     bool result = erasable();
     if (is_insertable_map && result) {
-      execution_space().fence(
-          "Kokkos::UnorderedMap::end_erase: fence before erasing");
-      Impl::UnorderedMapErase<declared_map_type> f(*this);
-      f.apply();
-      execution_space().fence(
-          "Kokkos::UnorderedMap::end_erase: fence after erasing");
+      space.fence("Kokkos::UnorderedMap::end_erase: fence before erasing");
```
masterleinad, line 493: Same here.

containers/src/impl/Kokkos_UnorderedMap_impl.hpp
```diff
@@ -47,16 +47,15 @@ struct UnorderedMapRehash {
...
   map_type m_dst;
   const_map_type m_src;
 
-  UnorderedMapRehash(map_type const& dst, const_map_type const& src)
-      : m_dst(dst), m_src(src) {}
-
-  void apply() const {
-    parallel_for("Kokkos::Impl::UnorderedMapRehash::apply", m_src.capacity(),
-                 *this);
+  void apply(const execution_space& space = execution_space{}) const {
```
masterleinad, line 56: Is there a reason to keep the default argument? Since this is not user-facing, I would rather add passing an execution space instance on the call site.

containers/src/impl/Kokkos_UnorderedMap_impl.hpp
```diff
@@ -76,19 +75,19 @@ struct UnorderedMapErase {
...
   using value_type      = typename map_type::impl_value_type;
+  using policy_type =
+      Kokkos::RangePolicy<execution_space, IndexType<size_type>>;
 
   map_type m_map;
 
-  UnorderedMapErase(map_type const& map) : m_map(map) {}
-
-  void apply() const {
+  void apply(const execution_space& space = execution_space{}) const {
```
masterleinad, line 83: I would similarly avoid the default argument here.

core/src/Kokkos_Macros.hpp
```diff
@@ -532,7 +532,11 @@ static constexpr bool kokkos_omp_on_host() { return false; }
 #define KOKKOS_ENABLE_CUDA_LDG_INTRINSIC
 #endif
 
-#define KOKKOS_INVALID_INDEX (~std::size_t(0))
+//! Invalid index of a given @p __type__.
+#define KOKKOS_INVALID_INDEX_TYPE(__type__) ~static_cast<__type__>(0)
```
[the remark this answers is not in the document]
masterleinad: If we are to provide this macro at all, this should be prefixed with `KOKKOS_IMPL`.

containers/src/impl/Kokkos_UnorderedMap_impl.hpp
masterleinad: You likely get a pull request merged if you change fewer lines of code but it's not a showstopper for me.

containers/src/Kokkos_UnorderedMap.hpp
```diff
@@ -465,26 +476,24 @@ class UnorderedMap {
     return is_insertable_map ? get_flag(erasable_idx) : false;
   }
 
-  bool begin_erase() {
+  bool begin_erase(const execution_space &space = execution_space{}) {
     bool result = !erasable();
     if (is_insertable_map && result) {
-      execution_space().fence(
+      space.fence(
```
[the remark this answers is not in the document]
romintomasetti: Ok, I'll remove the fence. Should I provide an overload that does not expect an execution space instance?

containers/src/Kokkos_UnorderedMap.hpp
romintomasetti: Ok, I'll remove the fences. Should I provide an overload that does not expect an execution space instance?

containers/src/impl/Kokkos_UnorderedMap_impl.hpp
romintomasetti: I agree, I removed the default.

containers/src/impl/Kokkos_UnorderedMap_impl.hpp
romintomasetti: Indeed, consider it done :wink:

containers/src/Kokkos_UnorderedMap.hpp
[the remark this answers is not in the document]
romintomasetti: BTW, `begin_erase` calls `set_flag`. And `set_flag` makes a `Kokkos::fence(..)`.

I guess I should create an issue for that ?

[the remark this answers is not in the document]
masterleinad: Sure.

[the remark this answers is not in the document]
romintomasetti: Sure, for the issue, the overload or both ? :smile:

containers/src/Kokkos_UnorderedMap.hpp
```diff
@@ -413,23 +414,33 @@ class UnorderedMap {
   /// This is <i>not</i> a device function; it may <i>not</i> be
   /// called in a parallel kernel.
   bool rehash(size_type requested_capacity = 0) {
+    return rehash(execution_space{}, requested_capacity);
```
masterleinad, line 417: We should probably fence after calling the overload taking an execution space instance.

containers/src/Kokkos_UnorderedMap.hpp
```diff
@@ -413,23 +414,33 @@ class UnorderedMap {
...
+  }
+
+  bool rehash(const execution_space &space, size_type requested_capacity = 0) {
     const bool bounded_insert = (capacity() == 0) || (size() == 0u);
-    return rehash(requested_capacity, bounded_insert);
+    return rehash(space, requested_capacity, bounded_insert);
   }
 
   bool rehash(size_type requested_capacity, bool bounded_insert) {
+    return rehash(execution_space{}, requested_capacity, bounded_insert);
```
masterleinad, line 426: Here as well.

containers/src/Kokkos_UnorderedMap.hpp
```diff
@@ -465,26 +476,19 @@ class UnorderedMap {
     return is_insertable_map ? get_flag(erasable_idx) : false;
   }
 
-  bool begin_erase() {
+  bool begin_erase(const execution_space &space) {
```
masterleinad, line 479: We still need the overload without execution space instance for backward-compatibility.

containers/src/Kokkos_UnorderedMap.hpp
romintomasetti: Do you agree that it's a `Kokkos::fence` that's needed?
```c++
bool rehash(size_type requested_capacity = 0) {
  bool result = rehash(execution_space{}, requested_capacity);
  Kokkos::fence("Kokkos::UnorderedMap::rehash: fence after rehasing");
  return result;
}
```

containers/src/Kokkos_UnorderedMap.hpp
romintomasetti: Ok, I've written one. But I'm not sure it is done in the correct way. Could you check please?

core/src/Kokkos_Macros.hpp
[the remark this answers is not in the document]
romintomasetti: If you agree, I can make a separate PR with this change (as suggested by @dal24). The macro's name would then be `KOKKOS_IMPL_INVALID_INDEX`.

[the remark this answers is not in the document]
masterleinad: Sure as long as you remove the changes you make in other pull requests here.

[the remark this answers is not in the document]
romintomasetti: Done in #6641.

dalg24 (changes requested):
Move the `Impl::UnorderedMapErase` aggregate initialization debate to another PR (or explain why it belongs here)
Withdraw all the "invalid type" change form this PR and solely focus on asynchronous behavior

containers/src/Kokkos_UnorderedMap.hpp
masterleinad: That's what we are normally doing.

masterleinad:
> Would a proposal as follows make sense (at least the concept)? It would be the implementation of the pattern mentioned above:

I prefer just calling `fence` before and after explicitly. Given that there might be a lot of code in between the fences, I find the introduction of a wrapper more obfuscating than helpful.

romintomasetti:
Hi @masterleinad ! I think this PR is now more focused on the topic "add `exec` argument to rehash, erase, etc".

I think I used the same style as you use *e.g.* in `Kokkos::fill_random` (when no `exec` given, first use `Kokkos::fence`, then call the overload with the default `exec`, then fence on `exec`).

masterleinad (commented):
Please add/modify some tests so that the new overloads are actually tested.

containers/unit_tests/TestUnorderedMap.hpp
```diff
@@ -158,10 +157,8 @@ struct TestErase {
   TestErase(map_type map, uint32_t num_erases, uint32_t num_duplicates)
       : m_map(map), m_num_erase(num_erases), m_num_duplicates(num_duplicates) {}
 
-  void testit() {
-    execution_space().fence();
-    Kokkos::parallel_for(m_num_erase, *this);
-    execution_space().fence();
+  void testit(const execution_space& exec) {
+    Kokkos::parallel_for(Kokkos::RangePolicy<execution_space>(exec, 0, m_num_erase), *this);
```
masterleinad, line 161: All the other `tesit`s fence the execution space instance.
```suggestion
    Kokkos::parallel_for(Kokkos::RangePolicy<execution_space>(exec, 0, m_num_erase), *this);
    exec.fence();
```

masterleinad (commented):
Have you checked that erase and rehash don't fence internally when called with execution space instances? It would be good to add a corresponding check since that seems to be the purpose of this pull request.

containers/src/Kokkos_Bitset.hpp
```diff
@@ -130,10 +130,11 @@ class Bitset {
 
   /// set all bits to 1
   /// can only be called from the host
-  void set() {
-    Kokkos::deep_copy(m_blocks, ~0u);
+  void set(const execution_space& exec) {
```
dalg24, line 133: This would be a breaking change.
Same comment about reset() and clear()

romintomasetti:
This is not a priority, and I might not even need this anymore. Closing.

PR #6614, opened 2023-11-20
GitHub Workflows: Use Ubuntu 22.04 instead of Fedora for Intel compiler testing

masterleinad:
Ubuntu 22.04 should have a better lifetime than Fedora, also see https://github.com/kokkos/ci-containers/pull/46#issue-2002295766.


cz4rs (approved):
Looks good to me.

fnrizzi (approved):

crtrott:
For the record Fedora Intel builds started failing (I believe some issue with the std lib headers).

crtrott (approved):

PR #6615, opened 2023-11-21
nvcc(wrapper): adding "long" and "short" versions for all flags

romintomasetti:
This PR adds "long" and "short" versions of valid `nvcc` flags to the `nvcc_wrapper` script.

### DIFF
--- a/bin/nvcc_wrapper
+++ b/bin/nvcc_wrapper
@@ -229,7 +229,7 @@ do
     fi
     ;;
   #Handle known nvcc args
-  --dryrun|--verbose|--keep|--source-in-ptx|-src-in-ptx|--keep-dir*|-G|-lineinfo|-extended-lambda|-expt-extended-lambda|-expt-relaxed-constexpr|--resource-usage|--fmad=*|--use_fast_math|--Wext-lambda-captures-this|-Wext-lambda-captures-this)
+  --dryrun|-dryrun|--verbose|-v|--keep|-keep|--source-in-ptx|-src-in-ptx|--keep-dir*|-keep-dir*|-G|-lineinfo|--generate-line-info|-extended-lambda|-expt-extended-lambda|-expt-relaxed-constexpr|--resource-usage|-res-usage|-fmad=*|--use_fast_math|-use_fast_math|--Wext-lambda-captures-this|-Wext-lambda-captures-this)
     cuda_args="$cuda_args $1"
     ;;
   #Handle more known nvcc args


masterleinad (approved):
Fine with me.

bin/nvcc_wrapper
```diff
@@ -229,7 +229,7 @@ do
     fi
     ;;
   #Handle known nvcc args
-  --dryrun|--verbose|--keep|--source-in-ptx|-src-in-ptx|--keep-dir*|-G|-lineinfo|-extended-lambda|-expt-extended-lambda|-expt-relaxed-constexpr|--resource-usage|--fmad=*|--use_fast_math|--Wext-lambda-captures-this|-Wext-lambda-captures-this)
+  --dryrun|--verbose|--keep|--source-in-ptx|-src-in-ptx|--keep-dir*|-G|-lineinfo|--generate-line-info|-extended-lambda|-expt-extended-lambda|-expt-relaxed-constexpr|--resource-usage|--fmad=*|--use_fast_math|--Wext-lambda-captures-this|-Wext-lambda-captures-this)
```
fnrizzi, line 232: seems to me this is already present via the shortcut `-lineinfo`

fnrizzi (changes requested):
see my comment please

bin/nvcc_wrapper
romintomasetti: @fnrizzi You're right that `--generate-line-info` and `-lineinfo` are the same thing.

However, when reading the output of `nvcc --help` or reading [the documentation](https://docs.nvidia.com/cuda/pdf/CUDA_Compiler_Driver_NVCC.pdf), it seems that `--generate-line-info` is the preferred choice (since `-lineinfo` is always in parentheses).

My own opinion (for what it's worth) is that:
1. `nvcc_wrapper` should be able to allow **any** valid `nvcc` argument.
2. It tend to prefer the `--generate-line-info` which makes more sense when reading it ("generate line information"), while `-lineinfo` is slightly more cryptic...

fnrizzi: i think it is just an alias for it, not necessarily a preferred way. 
In any case, I agree that nvcc_wrapper should accept any valid arg. So I would suggest to modify thjis PR so that for every flag, we allow both the "long" version as well as the "short" one.

romintomasetti: Ideally that's what I should do... However, I must admit I'm out of time this week :timer_clock: On a very pragmatic side of the things, I guess `nvcc_wrapper` could be enriched "on the go" with users' requests. What do you think?

fnrizzi: ok, how about a follow up PR for this?

romintomasetti: Fine with me. How would I allow you to modify my branch ?

fnrizzi, line 232:
```suggestion
  --dryrun|-dryrun|--verbose|-v|--keep|-keep|--source-in-ptx|-src-in-ptx|--keep-dir*|-keep-dir*|-G|-lineinfo|--generate-line-info|--extended-lambda|-extended-lambda|--expt-extended-lambda|-expt-extended-lambda|--expt-relaxed-constexpr|-expt-relaxed-constexpr|--resource-usage|-res-usage|--fmad=*|-fmad=*|--use_fast_math|-use_fast_math|--Wext-lambda-captures-this|-Wext-lambda-captures-this)
```

fnrizzi: i just posted the suggestion  below should be easy to just commit that

romintomasetti: Ok, just tell me when you're OK with the suggestion and I'll push it :wink:

fnrizzi: from https://docs.nvidia.com/cuda/cuda-compiler-driver-nvcc/index.html

fnrizzi: try now please :)

fnrizzi: please change the description and i am good ! thank you for the PR!

fnrizzi (approved):

masterleinad (changes requested):

bin/nvcc_wrapper
```diff
@@ -229,7 +229,7 @@ do
     fi
     ;;
   #Handle known nvcc args
-  --dryrun|--verbose|--keep|--source-in-ptx|-src-in-ptx|--keep-dir*|-G|-lineinfo|-extended-lambda|-expt-extended-lambda|-expt-relaxed-constexpr|--resource-usage|--fmad=*|--use_fast_math|--Wext-lambda-captures-this|-Wext-lambda-captures-this)
+  --dryrun|-dryrun|--verbose|-v|--keep|-keep|--source-in-ptx|-src-in-ptx|--keep-dir*|-keep-dir*|-G|-lineinfo|--generate-line-info|--extended-lambda|-extended-lambda|--expt-extended-lambda|-expt-extended-lambda|--expt-relaxed-constexpr|-expt-relaxed-constexpr|--resource-usage|-res-usage|--fmad=*|-fmad=*|--use_fast_math|-use_fast_math|--Wext-lambda-captures-this|-Wext-lambda-captures-this)
```
fnrizzi, line 232:
```suggestion
--dryrun|-dryrun|--verbose|-v|--keep|-keep|--source-in-ptx|-src-in-ptx|--keep-dir*|-keep-dir*|-G|-lineinfo|--generate-line-info|-extended-lambda|-expt-extended-lambda|-expt-relaxed-constexpr|--resource-usage|-res-usage|-fmad=*|--use_fast_math|-use_fast_math|--Wext-lambda-captures-this|-Wext-lambda-captures-this)
```
this is without duplicates from the lines below

cz4rs (approved):

masterleinad (approved):
Fine with me.

romintomasetti:
Is there anything we can do to make the pipeline happy ? Should I rebase on the develop to fetch a fix or something?

crtrott (approved):

crtrott:
Its a bit of a mess but I guess we just need to clean this up in a holistic manner at some point.

PR #6617, opened 2023-11-22
Add bound checks in RangePolicy and MDRangePolicy

### LINKED ISSUE #6616: Possible infinite loop with MDRangePolicy when upper bound is smaller than lower bound
**Describe the bug**

Currently, when using this (e.g. with Kokkos::Serial backend):
```c++
    Kokkos::parallel_for(
      "test_mdrange",
      Kokkos::MDRangePolicy<Kokkos::Rank<2>>({ 100, 100 }, { 95, 105 }),
      KOKKOS_LAMBDA(const int i, const int j)
      {
          printf("i,j=%d,%d\n", i, j);
      });
```

it leads to an infinite loop. Probably that  `MDRangePolicy::init_helper` should do a `Kokkos::Abort` if there is a rank `i` such that `m_upper[i] < m_lower[i]`

1. Compilers (with versions)
2. Kokkos release or commit used (i.e., the sha1 number)
3. Platform, architecture and backend
4. CMake configure command
5. Output from CMake configure command
6. Minimum, complete code needed to reproduce the bug
7. Command line needed to reproduce the bug
8. `KokkosCore_config.h` header file (generated during the build)
9. Please provide any additional relevant error logs

ldh4:
Resolves #6616.

Added a simple check in MDRangePolicy that checks if any of the `end index` is less than its corresponding `start index` for all ranks.

### DIFF
--- a/core/src/KokkosExp_MDRangePolicy.hpp
+++ b/core/src/KokkosExp_MDRangePolicy.hpp
@@ -30,6 +30,7 @@ static_assert(false,
 #include <impl/KokkosExp_Host_IterateTile.hpp>
 #include <Kokkos_ExecPolicy.hpp>
 #include <type_traits>
+#include <sstream>
 
 namespace Kokkos {
 
@@ -328,10 +329,13 @@ struct MDRangePolicy : public Kokkos::Impl::PolicyTraits<Properties...> {
     for (int i = rank_start; i != rank_end; i += increment) {
       const index_type length = m_upper[i] - m_lower[i];
 
-      if (m_upper[i] < m_lower[i])
-        Kokkos::abort(
-            "Kokkos::MDRangePolicy bounds error: One of the lower bounds is "
-            "greater than the upper bound of its rank.");
+      if (m_upper[i] < m_lower[i]) {
+        std::stringstream msg;
+        msg << "Kokkos::MDRangePolicy bounds error: The lower bound ("
+            << m_lower[i] << ") is greater than its upper bound (" << m_upper[i]
+            << ") in rank " << i + 1 << ".";
+        Kokkos::abort(msg.str().c_str());
+      }
 
       if (m_tile[i] <= 0) {
         m_tune_tile_size = true;
--- a/core/src/KokkosExp_MDRangePolicy.hpp
+++ b/core/src/KokkosExp_MDRangePolicy.hpp
@@ -30,7 +30,6 @@ static_assert(false,
 #include <impl/KokkosExp_Host_IterateTile.hpp>
 #include <Kokkos_ExecPolicy.hpp>
 #include <type_traits>
-#include <sstream>
 
 namespace Kokkos {
 
@@ -330,11 +329,12 @@ struct MDRangePolicy : public Kokkos::Impl::PolicyTraits<Properties...> {
       const index_type length = m_upper[i] - m_lower[i];
 
       if (m_upper[i] < m_lower[i]) {
-        std::stringstream msg;
-        msg << "Kokkos::MDRangePolicy bounds error: The lower bound ("
-            << m_lower[i] << ") is greater than its upper bound (" << m_upper[i]
-            << ") in rank " << i + 1 << ".";
-        Kokkos::abort(msg.str().c_str());
+        std::string msg =
+            "Kokkos::MDRangePolicy bounds error: The lower bound (" +
+            std::to_string(m_lower[i]) + ") is greater than its upper bound (" +
+            std::to_string(m_upper[i]) + ") in rank " + std::to_string(i + 1) +
+            ".";
+        Kokkos::abort(msg.c_str());
       }
 
       if (m_tile[i] <= 0) {
--- a/core/src/KokkosExp_MDRangePolicy.hpp
+++ b/core/src/KokkosExp_MDRangePolicy.hpp
@@ -327,6 +327,12 @@ struct MDRangePolicy : public Kokkos::Impl::PolicyTraits<Properties...> {
     }
     for (int i = rank_start; i != rank_end; i += increment) {
       const index_type length = m_upper[i] - m_lower[i];
+
+      if (m_upper[i] < m_lower[i])
+        Kokkos::abort(
+            "Kokkos::MDRangePolicy bounds error: One of the lower bounds is "
+            "greater than the upper bound of its rank.");
+
       if (m_tile[i] <= 0) {
         m_tune_tile_size = true;
         if ((inner_direction == Iterate::Right && (i < rank - 1)) ||
--- a/core/src/KokkosExp_MDRangePolicy.hpp
+++ b/core/src/KokkosExp_MDRangePolicy.hpp
@@ -332,7 +332,7 @@ struct MDRangePolicy : public Kokkos::Impl::PolicyTraits<Properties...> {
         std::string msg =
             "Kokkos::MDRangePolicy bounds error: The lower bound (" +
             std::to_string(m_lower[i]) + ") is greater than its upper bound (" +
-            std::to_string(m_upper[i]) + ") in rank " + std::to_string(i + 1) +
+            std::to_string(m_upper[i]) + ") in dimension " + std::to_string(i) +
             ".";
         Kokkos::abort(msg.c_str());
       }
--- a/core/src/Kokkos_ExecPolicy.hpp
+++ b/core/src/Kokkos_ExecPolicy.hpp
@@ -117,16 +117,18 @@ class RangePolicy : public Impl::PolicyTraits<Properties...> {
   inline RangePolicy(const typename traits::execution_space& work_space,
                      const member_type work_begin, const member_type work_end)
       : m_space(work_space),
-        m_begin(work_begin < work_end ? work_begin : 0),
-        m_end(work_begin < work_end ? work_end : 0),
+        m_begin(work_begin),
+        m_end(work_end),
         m_granularity(0),
         m_granularity_mask(0) {
+    check_bounds_validity();
     set_auto_chunk_size();
   }
 
   /** \brief  Total range */
   inline RangePolicy(const member_type work_begin, const member_type work_end)
       : RangePolicy(typename traits::execution_space(), work_begin, work_end) {
+    check_bounds_validity();
     set_auto_chunk_size();
   }
 
@@ -136,10 +138,11 @@ class RangePolicy : public Impl::PolicyTraits<Properties...> {
                      const member_type work_begin, const member_type work_end,
                      Args... args)
       : m_space(work_space),
-        m_begin(work_begin < work_end ? work_begin : 0),
-        m_end(work_begin < work_end ? work_end : 0),
+        m_begin(work_begin),
+        m_end(work_end),
         m_granularity(0),
         m_granularity_mask(0) {
+    check_bounds_validity();
     set_auto_chunk_size();
     set(args...);
   }
@@ -149,6 +152,7 @@ class RangePolicy : public Impl::PolicyTraits<Properties...> {
   inline RangePolicy(const member_type work_begin, const member_type work_end,
                      Args... args)
       : RangePolicy(typename traits::execution_space(), work_begin, work_end) {
+    check_bounds_validity();
     set_auto_chunk_size();
     set(args...);
   }
@@ -216,6 +220,16 @@ class RangePolicy : public Impl::PolicyTraits<Properties...> {
     }
     m_granularity      = new_chunk_size;
     m_granularity_mask = m_granularity - 1;
+  }
+
+  inline void check_bounds_validity() {
+    if (m_end < m_begin) {
+      std::string msg = "Kokkos::RangePolicy bounds error: The lower bound (" +
+                        std::to_string(m_begin) +
+                        ") is greater than the upper bound (" +
+                        std::to_string(m_end) + ").";
+      Kokkos::abort(msg.c_str());
+    }
   }
 
  public:
--- a/core/unit_test/TestMDRangePolicyConstructors.hpp
+++ b/core/unit_test/TestMDRangePolicyConstructors.hpp
@@ -93,6 +93,25 @@ TEST(TEST_CATEGORY_DEATH, policy_bounds_unsafe_narrowing_conversions) {
       },
       "unsafe narrowing conversion");
 }
+
+TEST(TEST_CATEGORY_DEATH, policy_invalid_bounds) {
+  using Policy = Kokkos::MDRangePolicy<TEST_EXECSPACE, Kokkos::Rank<5>>;
+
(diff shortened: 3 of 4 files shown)


core/src/KokkosExp_MDRangePolicy.hpp
```diff
@@ -327,6 +327,12 @@ struct MDRangePolicy : public Kokkos::Impl::PolicyTraits<Properties...> {
     }
     for (int i = rank_start; i != rank_end; i += increment) {
       const index_type length = m_upper[i] - m_lower[i];
+
+      if (m_upper[i] < m_lower[i])
+        Kokkos::abort(
+            "Kokkos::MDRangePolicy bounds error: One of the lower bounds is "
+            "greater than the upper bound of its rank.");
```
fnrizzi, line 334: i would also print the index so that we can tell users more explicitly which bound was triggering this

ldh4: Modified the wording to print the input bounds and the rank. See 0e9ee83.

core/src/KokkosExp_MDRangePolicy.hpp
```diff
@@ -327,6 +328,15 @@ struct MDRangePolicy : public Kokkos::Impl::PolicyTraits<Properties...> {
...
     for (int i = rank_start; i != rank_end; i += increment) {
       const index_type length = m_upper[i] - m_lower[i];
+
+      if (m_upper[i] < m_lower[i]) {
+        std::stringstream msg;
+        msg << "Kokkos::MDRangePolicy bounds error: The lower bound ("
+            << m_lower[i] << ") is greater than its upper bound (" << m_upper[i]
+            << ") in rank " << i + 1 << ".";
+        Kokkos::abort(msg.str().c_str());
+      }
```
masterleinad, line 338: Did you consider just treating it as an empty range if the upper bound is lower than the lower bound?
Isn't that what we are doing for `RangePolicy`?

ldh4: I didn't realize that `RangePolicy` behaves this way, but the idea of silently changing a policy to be an empty range seems unhelpful for the users, although I admit that given a lower bound that's higher than its upper bound, making it an empty range is most logical.

In case of `MDRangePolicy`, since the lower bound and upper bound checks will be needed for all ranks, should all ranges become empty ranges upon finding a failed bound check or should only failing ranks be empty ranges?

fnrizzi: I personally think we should do something that is the most useful for the users, which IMO is catching the problem as early as possible. Adopting the solution proposed in this PR seems good to me because the problem is caught immediately during the construction of the policy.

core/src/KokkosExp_MDRangePolicy.hpp
```diff
@@ -327,6 +328,15 @@ struct MDRangePolicy : public Kokkos::Impl::PolicyTraits<Properties...> {
     }
     for (int i = rank_start; i != rank_end; i += increment) {
       const index_type length = m_upper[i] - m_lower[i];
+
+      if (m_upper[i] < m_lower[i]) {
+        std::stringstream msg;
+        msg << "Kokkos::MDRangePolicy bounds error: The lower bound ("
+            << m_lower[i] << ") is greater than its upper bound (" << m_upper[i]
+            << ") in rank " << i + 1 << ".";
+        Kokkos::abort(msg.str().c_str());
```
fnrizzi, line 337: can we avoid pulling the sstream header into this and just use a regular string?

core/src/KokkosExp_MDRangePolicy.hpp
```diff
@@ -30,6 +30,7 @@ static_assert(false,
 #include <impl/KokkosExp_Host_IterateTile.hpp>
 #include <Kokkos_ExecPolicy.hpp>
 #include <type_traits>
+#include <sstream>
```
fnrizzi, line 33: can we avoid pulling the sstream header into this and just use a regular string? (see my comment below)

core/src/KokkosExp_MDRangePolicy.hpp
pkestene: FWIW, at first I was also thinking that aborting was the right thing to do; but just emptying the range is also a good solution, afterall when doing `for(int i=0; i<-10;i++)` this is valid code that may happen when implementation an algorithm where the bound are computed using analytical expressions, and the desired behavior is doing nothing.

fnrizzi: "when implementating an algorithm where the bound are computed using analytical expressions, and the desired behavior is doing nothing." what do you mean by: 
1. using analytical expressions,
2. the desired behavor is doing nothing

fnrizzi: I think we need to differentiate what concerns the algirithm vs how the MDRangePolicy works. 
I don't think it would be right to tie the implementation/semantics of the MDRangePolicy to something that pertains a certain algorithm. 
It should be up to the algorithm to translate its requirements into something that makes sense for MDRangePolicy. 
if you algorithm has meaning for a range like [15, -20], then it should be up to you to translate that into an empty MDRangePolicy. 
IMO, the MDRangePolicy should either accept a valid set of bounds or empty bounds but NOT "flipped" bounds. 
Does it make sense what I am trying to convey?

pkestene: I agree with your last option (3 bullets above) and that it should be clearly stated in the documentation.
IMHO, if one rank has upper bound < lower bound then the right behavior, is don't do any iteration at all; afterall this is what happen when you have two nest for loops, and one of them has invalid range.

Somehow, about your 3rd bullet above, when upper bound is < lower bound, it is not really an invalid range, just an empty range, but for a different reason.

The last point is that Range and MDRange policy should behave the same. If you think aborting is best in case of invalid range in MDRange, then Range policy should behave likewise. 

Finally, I don't have strong opinion here, just saying that first I was thinking aborting was the right behavior, but finally I can live fine with emptying the range.

masterleinad: Maybe, we just discuss it on Wednesday in the meeting.

core/src/KokkosExp_MDRangePolicy.hpp
ldh4: Removed in [341b464](https://github.com/kokkos/kokkos/pull/6617/commits/341b464d38638c07fb4767c4becb35ab78859dbd).

core/src/KokkosExp_MDRangePolicy.hpp
ldh4: Converted to a string in [341b464](https://github.com/kokkos/kokkos/pull/6617/commits/341b464d38638c07fb4767c4becb35ab78859dbd).

core/src/KokkosExp_MDRangePolicy.hpp
```diff
@@ -327,6 +327,16 @@ struct MDRangePolicy : public Kokkos::Impl::PolicyTraits<Properties...> {
     }
     for (int i = rank_start; i != rank_end; i += increment) {
       const index_type length = m_upper[i] - m_lower[i];
+
+      if (m_upper[i] < m_lower[i]) {
+        std::string msg =
+            "Kokkos::MDRangePolicy bounds error: The lower bound (" +
+            std::to_string(m_lower[i]) + ") is greater than its upper bound (" +
+            std::to_string(m_upper[i]) + ") in rank " + std::to_string(i + 1) +
```
dalg24, line 335: How does
rank -> dimension
i+1 -> I
sound?

core/unit_test/TestMDRangePolicyConstructors.hpp
```diff
@@ -93,6 +93,27 @@ TEST(TEST_CATEGORY_DEATH, policy_bounds_unsafe_narrowing_conversions) {
       },
       "unsafe narrowing conversion");
 }
+
+TEST(TEST_CATEGORY_DEATH, policy_invalid_bounds) {
+  using Policy = Kokkos::MDRangePolicy<TEST_EXECSPACE, Kokkos::Rank<5>>;
```
dalg24, line 98: Why rank 5 if we can trigger with 2?

core/src/KokkosExp_MDRangePolicy.hpp
ldh4: Changed error output this.

core/unit_test/TestMDRangePolicyConstructors.hpp
ldh4: No real reason. Changed to 2 ranks and removed the second test case.

core/unit_test/TestMDRangePolicyConstructors.hpp
```diff
@@ -93,6 +93,19 @@ TEST(TEST_CATEGORY_DEATH, policy_bounds_unsafe_narrowing_conversions) {
...
+  using Policy = Kokkos::MDRangePolicy<TEST_EXECSPACE, Kokkos::Rank<2>>;
+
+  ::testing::FLAGS_gtest_death_test_style = "threadsafe";
+  ASSERT_DEATH(
+      {
+        (void)Policy({100, 100}, {90, 90});
+      },
+      "Kokkos::MDRangePolicy bounds error: The lower bound \\(100\\) is "
+      "greater "
+      "than its upper bound \\(90\\) in dimension 1\\.");
```
dalg24, line 107: I hope the test fails...

Rombur: The failing dimension depends if you use `Iterate::Left` or `Iterate::Right`

dalg24: So we expect the test as is should not pass on all backends?

Rombur: Yes, the test should pass for some backends and fail for others

masterleinad: Have we decided anyway that the bounds check should always produce an error?

ldh4: @dalg24 @Rombur is correct here. It's `Iterate::Right` by default here. So the check would start from `rank-1` dimension first.
https://github.com/kokkos/kokkos/blob/c3dde624d4228b1dd7f8952bc2439a64db5207ee/core/src/KokkosExp_MDRangePolicy.hpp#L318-L327

@masterleinad No, we haven't decided on that. If we do an abort here for `MDRangePolicy`, it's inconsistent from how `RangePolicy` handles this, so I think we definitely should discuss at some point before merging.

fnrizzi: I would vote to abort and also change range policy to reflect this for the same reason of what I said above.

ldh4: Fixed to check against the policy's iteration pattern.

> I would vote to abort and also change range policy to reflect this for the same reason of what I said in the other comment

I agree with this approach as well. My only concern is if that will cause many breakages for the users. Although the change is for a corner case, it's a sudden behavioral change in a commonly used function.

dalg24 (approved):
Please update the API doc to clearly state that `begin[d] <= end[d]` for all dimensions is a precondition for the constructors.

core/unit_test/TestMDRangePolicyConstructors.hpp
```diff
@@ -93,6 +93,23 @@ TEST(TEST_CATEGORY_DEATH, policy_bounds_unsafe_narrowing_conversions) {
...
+
+  auto dim = (Policy::inner_direction == Kokkos::Iterate::Right) ? 1 : 0;
+
+  ASSERT_DEATH(
+      {
+        (void)Policy({100, 100}, {90, 90});
+      },
+      "Kokkos::MDRangePolicy bounds error: The lower bound \\(100\\) is "
+      "greater "
+      "than its upper bound \\(90\\) in dimension " +
```
dalg24, line 110:
```suggestion
      "greater than its upper bound \\(90\\) in dimension " +
```

dalg24:
If merged, must appear in the changelog under "incpompatibilies"

dalg24:
Also if we proceed with this PR, we'd probably want to also enforce the precondition for `RangePolicy`.

dalg24:
We might need to guard against `KOKKOS_ENABLE_DEPRECATED_CODE_4` to be safe.

ldh4:
> Also if we proceed with this PR, we'd probably want to also enforce the precondition for RangePolicy.

Yes, we still need to have a discussion on that to come to a decision.
For consistency, either `RangePolicy` will need to have a similar precondition with an abort or `MDRangePolicy` will need to use a silently emptied range upon violating its precondition.

core/src/KokkosExp_MDRangePolicy.hpp
```diff
@@ -327,6 +327,16 @@ struct MDRangePolicy : public Kokkos::Impl::PolicyTraits<Properties...> {
     }
     for (int i = rank_start; i != rank_end; i += increment) {
       const index_type length = m_upper[i] - m_lower[i];
+
+      if (m_upper[i] < m_lower[i]) {
```
crtrott, line 331: we definitely need to allow 0 to 0 etc.

dalg24: this check is fine

ldh4:
Updated `RangePolicy` to enforce the same precondition.

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -117,16 +117,18 @@ class RangePolicy : public Impl::PolicyTraits<Properties...> {
...
         m_granularity(0),
         m_granularity_mask(0) {
+    check_bounds_validity();
     set_auto_chunk_size();
   }
 
   /** \brief  Total range */
   inline RangePolicy(const member_type work_begin, const member_type work_end)
       : RangePolicy(typename traits::execution_space(), work_begin, work_end) {
+    check_bounds_validity();
```
dalg24, line 131: You'd never get here since the construction is deferred to the other `RangePolicy` constructor that would error out.

ldh4: I did notice it before, but I placed it in just to have this check logic be paired with each `set_auto_chunk_size()` call, which makes uses of `m_begin` and `m_end`.
If you feel strongly that we don't want this in this constructor (and the other similar constructor below), I'll remove the lines.

dalg24: I can live with it.  There are things that bother me more in that file :)

ldh4:
Retest this please.

tcclevenger (approved):

dalg24:
CI failures are unrelated

PR #6619, opened 2023-11-24
view(move): fixing move semantics

romintomasetti:
This PR fixes the move semantics of `Kokkos::View`.

To better understand the problem, please look at the new test `TestViewMove.hpp`.

* `Kokkos::Impl::ViewTracker` was not fulfilling the rules for an implicitly defined move constructor/assignment, so it must be manually defined. Before that, calling `Kokkos::View::use_count` on a moved view was not returning 0, as one would have expected. Now it does.
* `Kokkos::Impl::ViewMapping` was fulfilling the rules for an implicitly defined move constructor/assignment, but the implicitly defined ones would not set `m_impl_handle` back to `nullptr`. Therefore, calling `Kokkos::View::is_allocated` on a moved view was not returning `false`, as one would have expected. Now it does.

Note that now, `Kokkos::View`'s move semantics follow the move semantics of *e.g.* `std::shared_ptr`:
```c++
std::shared_ptr<T> a(...);

std::shared_ptr<T> b(std::move(a));

if(a != nullptr || a.use_count() != 0) throw std::runtime_error("This is per the standard not expected.");
```

Join work with @maartenarnst.

### DIFF
--- a/core/src/impl/Kokkos_ViewMapping.hpp
+++ b/core/src/impl/Kokkos_ViewMapping.hpp
@@ -3131,8 +3131,8 @@ class ViewMapping<
 
   using handle_type = typename ViewDataHandle<Traits>::handle_type;
 
-  handle_type m_impl_handle;
-  offset_type m_impl_offset;
+  handle_type m_impl_handle = nullptr;
+  offset_type m_impl_offset {};
 
  private:
   template <class, class...>
@@ -3348,14 +3348,25 @@ class ViewMapping<
   //----------------------------------------
 
   KOKKOS_DEFAULTED_FUNCTION ~ViewMapping() = default;
-  KOKKOS_INLINE_FUNCTION ViewMapping() : m_impl_handle(), m_impl_offset() {}
+  KOKKOS_DEFAULTED_FUNCTION ViewMapping()  = default;
 
   KOKKOS_DEFAULTED_FUNCTION ViewMapping(const ViewMapping&) = default;
   KOKKOS_DEFAULTED_FUNCTION ViewMapping& operator=(const ViewMapping&) =
       default;
 
-  KOKKOS_DEFAULTED_FUNCTION ViewMapping(ViewMapping&&) = default;
-  KOKKOS_DEFAULTED_FUNCTION ViewMapping& operator=(ViewMapping&&) = default;
+  //! Move constructor. Leaves @p other in an empty state.
+  KOKKOS_INLINE_FUNCTION
+  ViewMapping(ViewMapping&& other)
+      : m_impl_handle(std::exchange(other.m_impl_handle, nullptr)),
+        m_impl_offset(std::move(other.m_impl_offset)) {}
+
+  //! Move assignment. Leaves @p other in an empty state.
+  KOKKOS_INLINE_FUNCTION
+  ViewMapping& operator=(ViewMapping&& other) {
+    m_impl_handle = std::exchange(other.m_impl_handle, nullptr);
+    m_impl_offset = std::move(other.m_impl_offset);
+    return *this;
+  }
 
   //----------------------------------------
 
--- a/core/src/impl/Kokkos_ViewTracker.hpp
+++ b/core/src/impl/Kokkos_ViewTracker.hpp
@@ -42,9 +42,10 @@ struct ViewTracker {
 
   track_type m_tracker;
 
-  KOKKOS_INLINE_FUNCTION
-  ViewTracker() : m_tracker() {}
+  KOKKOS_DEFAULTED_FUNCTION
+  ViewTracker() = default;
 
+  //! Copy constructor.
   KOKKOS_INLINE_FUNCTION
   ViewTracker(const ViewTracker& vt) noexcept
       : m_tracker(vt.m_tracker, view_traits::is_managed) {}
@@ -73,6 +74,7 @@ struct ViewTracker {
     KOKKOS_IF_ON_DEVICE((m_tracker.assign_force_disable(vt.m_track.m_tracker);))
   }
 
+  //! Copy assignment.
   KOKKOS_INLINE_FUNCTION ViewTracker& operator=(
       const ViewTracker& rhs) noexcept {
     if (this == &rhs) return *this;
@@ -89,6 +91,14 @@ struct ViewTracker {
   KOKKOS_INLINE_FUNCTION
   explicit ViewTracker(const track_type& tt) noexcept
       : m_tracker(tt, view_traits::is_managed) {}
+
+  //! Move constructor.
+  KOKKOS_DEFAULTED_FUNCTION
+  ViewTracker(ViewTracker&&) = default;
+
+  //! Move assignment.
+  KOKKOS_DEFAULTED_FUNCTION
+  ViewTracker& operator=(ViewTracker&& other) = default;
 };
 
 }  // namespace Impl
--- a/core/unit_test/TestViewMove.hpp
+++ b/core/unit_test/TestViewMove.hpp
@@ -1,0 +1,59 @@
+//@HEADER
+// ************************************************************************
+//
+//                        Kokkos v. 4.0
+//       Copyright (2022) National Technology & Engineering
+//               Solutions of Sandia, LLC (NTESS).
+//
+// Under the terms of Contract DE-NA0003525 with NTESS,
+// the U.S. Government retains certain rights in this software.
+//
+// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+#ifndef TESTVIEWMOVE_HPP_
+#define TESTVIEWMOVE_HPP_
+
+#include <gtest/gtest.h>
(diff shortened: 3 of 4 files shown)


core/unit_test/TestViewMove.hpp
```diff
@@ -0,0 +1,59 @@
...
+  using tester_t        = TestViewMove<view_t>;
+
+  view_t view("view move test", 5);
+
+  EXPECT_EQ(view.use_count(), 1);
+  EXPECT_TRUE(view.is_allocated());
+
+  tester_t tester{std::move(view)};
+
+  //! As @ref view was moved, it's left in a pristine (a.k.a. *empty*) state.
```
romintomasetti, line 49:
```suggestion
  //! As the view was moved, it's left in a pristine (a.k.a. *empty*) state.
```

romintomasetti:
@dalg24 @masterleinad

core/src/impl/Kokkos_ViewMapping.hpp
```diff
@@ -3348,14 +3348,25 @@ class ViewMapping<
...
+  //! Move constructor. Leaves @p other in an empty state.
+  KOKKOS_INLINE_FUNCTION
+  ViewMapping(ViewMapping&& other)
+      : m_impl_handle(std::exchange(other.m_impl_handle, nullptr)),
+        m_impl_offset(std::move(other.m_impl_offset)) {}
+
+  //! Move assignment. Leaves @p other in an empty state.
+  KOKKOS_INLINE_FUNCTION
+  ViewMapping& operator=(ViewMapping&& other) {
+    m_impl_handle = std::exchange(other.m_impl_handle, nullptr);
```
masterleinad, line 3366: Same here.

core/src/impl/Kokkos_ViewTracker.hpp
```diff
@@ -89,6 +91,17 @@ struct ViewTracker {
   KOKKOS_INLINE_FUNCTION
   explicit ViewTracker(const track_type& tt) noexcept
       : m_tracker(tt, view_traits::is_managed) {}
+
+  //! Move constructor.
+  KOKKOS_INLINE_FUNCTION
+  ViewTracker(ViewTracker&& other) : m_tracker(std::move(other.m_tracker)) {}
```
masterleinad, line 97: Isn't that the default behavior?

core/src/impl/Kokkos_ViewTracker.hpp
```diff
@@ -89,6 +91,17 @@ struct ViewTracker {
...
+
+  //! Move assignment.
+  KOKKOS_INLINE_FUNCTION
+  ViewTracker& operator=(ViewTracker&& other) {
+    m_tracker = std::move(other.m_tracker);
+    return *this;
+  }
```
masterleinad, line 104: Same here. How does the behavior differ from the `default` implementation?

core/src/impl/Kokkos_ViewMapping.hpp
```diff
@@ -3131,8 +3131,8 @@ class ViewMapping<
 
   using handle_type = typename ViewDataHandle<Traits>::handle_type;
 
-  handle_type m_impl_handle;
-  offset_type m_impl_offset;
+  handle_type m_impl_handle = nullptr;
+  offset_type m_impl_offset{};
```
masterleinad, line 3135: Why are these changes necessary?

core/src/impl/Kokkos_ViewMapping.hpp
```diff
@@ -3348,14 +3348,25 @@ class ViewMapping<
   //----------------------------------------
 
   KOKKOS_DEFAULTED_FUNCTION ~ViewMapping() = default;
-  KOKKOS_INLINE_FUNCTION ViewMapping() : m_impl_handle(), m_impl_offset() {}
+  KOKKOS_DEFAULTED_FUNCTION ViewMapping()  = default;
```
masterleinad, line 3351: Why did you change the default constructor (when needing to provide initialization defaults anyway)?

core/src/impl/Kokkos_ViewTracker.hpp
romintomasetti: I also thought it would work using `ViewTracker(ViewTracker&&) = default;`.

However, doing so breaks the test I added in `TestViewMove.hpp`:
```bash
Note: Google Test filter = *view_move*
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from openmp
[ RUN      ] openmp.view_move
/workspaces/kokkos/core/unit_test/TestViewMove.hpp:50: Failure
  view.use_count()
    Which is: 2
  0
/workspaces/kokkos/core/unit_test/TestViewMove.hpp:53: Failure
  tester.view.use_count()
    Which is: 2
  1
[  FAILED  ] openmp.view_move (0 ms)
[----------] 1 test from openmp (0 ms total)

[----------] Global test environment tear-down
[==========] 1 test from 1 test suite ran. (0 ms total)
[  PASSED  ] 0 tests.
[  FAILED  ] 1 test, listed below:
[  FAILED  ] openmp.view_move

 1 FAILED TEST
```

For the test's output, my conclusion is that instead of being moved from one view to another, the view tracker was copied.

Reading [the rules for Implicitly-declared move constructor](https://en.cppreference.com/w/cpp/language/move_constructor), it seems that a defaulted move constructor is not eligible due to `ViewTracker` having user-declared copy constructors.

Am I wrong? Am I missing something?

core/src/impl/Kokkos_ViewMapping.hpp
romintomasetti: I think the change I need is only `handle_type m_impl_handle = nullptr;`.

These changes are related to your comment below. I defaulted the constructor, so I wanted to provide proper initialization of the members.

core/src/impl/Kokkos_ViewMapping.hpp
romintomasetti: Because I find it clearer to have proper default member values alongside with the member definition (see your comment above).

Also, in #6524, @dalg24 didn't use `KOKKOS_DEFAULTED_FUNCTION` for the default `Bitset` constructor. And in fact `KOKKOS_DEFAULTED_FUNCTION` is always defined empty. So is there any reason why `Kokkos` is keeping this macro alive? Should it be deprecated?

core/src/impl/Kokkos_ViewTracker.hpp
masterleinad: https://godbolt.org/z/saKdjrW3z shows that a `default` move constructor is selected even when a user-defined copy constructor is provided and this `default` move constructor calls the move constructor for member variables. Only when removing the `default` move constructor, the copy constructor is selected instead.

core/src/impl/Kokkos_ViewMapping.hpp
dalg24: I did not use `KOKKOS_DEFAULTED_FUNCTION` because `Bitset` is not intended to be default constructible on the device.

core/src/impl/Kokkos_ViewTracker.hpp
romintomasetti: I applied your suggestion (using the `default`). Thanks!

core/src/impl/Kokkos_ViewMapping.hpp
romintomasetti: > I did not use KOKKOS_DEFAULTED_FUNCTION because Bitset is not intended to be default constructible on the device.

Ah, OK! It's clearer now :mag:

dalg24:
What do you mean by "fixi(ng) the move semantics of `Kokkos::View`"?
Are you specifically looking at the moved-from state?

core/src/impl/Kokkos_ViewMapping.hpp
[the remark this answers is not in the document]
masterleinad: Looks like it works for `CUDA` and `SYCL` at least.

romintomasetti:
> What do you mean by "fixi(ng) the move semantics of `Kokkos::View`"? Are you specifically looking at the moved-from state?

@dalg24 I think there are 2 points:
1. This PR really fixes the "move" action of a view. This is the fix in `ViewTracker` (the move operators were not implicitly defined by the compiler, so we needed to explicitly default them). Without that fix, the view tracker was shared between the moved from and the moved to views.
2. This PR also addresses the state in which a moved view is left. Especially, it is ensuring that `is_allocated` returns `false` and `use_count` returns 0.

romintomasetti:
There are a lot of failing tests and compilation failures. Are they related to the changes induced by this PR ?

masterleinad:
> There are a lot of failing tests and compilation failures. Are they related to the changes induced by this PR ?

I would think so. There are a bunch of failures that we haven't seen elsewhere.

romintomasetti:
We agreed with @dalg24 that I would re-compile the Trilinos SHA we're using and running the tests.

This would help us detect any potential issue in downstream codes.

maartenarnst:
Just as a follow up to this thread. We noted that the View's move constructor also appears to be missing a version that converts. E.g. moving a statically sized view to a dynamically sized view.

dalg24:
Closing in favor #6636

PR #6620, opened 2023-11-24
Kokkos::DualView(semantics): allow moving host and device views into a dual view

romintomasetti:
This PR enhances the `Kokkos::DualView` constructor taking the host and device views by allowing it to take moving views.

This PR needs #6619.


dalg24:
Blocked on #6636

PR #6624, opened 2023-11-27
graph(HIP): adding inline keyword to fix #6623

### LINKED ISSUE #6623: `Kokkos::Graph`(`HIP`): declaration and definition in the same file leads to `multiple definition of`
**Describe the bug**

This happens since `Kokkos::Graph` was specialized for `HIP` in #6370.

Since member functions of `Kokkos::Impl::GraphImpl<Kokkos::HIP>` are both declared and defined in the same header file (without `inline` keyword), we get tons of `multiple definition of` when compiling our library:
```bash
multiple definition of `Kokkos::Impl::GraphImpl<Kokkos::HIP>::submit()`
...
multiple definition of `Kokkos::Impl::GraphImpl<Kokkos::HIP>::get_execution_space() const`
...
```

I think the problem can be solved by either defining the member functions inside the class (as done in `Kokkos_Cuda_Graph_Impl.hpp`) or by adding the `inline` keyword.

I can make a PR with whatever fix you prefer.

I'm tagging @Rombur @crtrott @masterleinad because they were involved in #6370.

romintomasetti:
This PR adds missing `inline` keyword in the specialization of `Kokkos::Impl::GraphImpl` for `Kokkos::HIP`.

Fixes #6623.

### DIFF
--- a/core/src/HIP/Kokkos_HIP_Graph_Impl.hpp
+++ b/core/src/HIP/Kokkos_HIP_Graph_Impl.hpp
@@ -83,7 +83,7 @@ class GraphImpl<Kokkos::HIP> {
   hipGraphExec_t m_graph_exec = nullptr;
 };
 
-GraphImpl<Kokkos::HIP>::~GraphImpl() {
+inline GraphImpl<Kokkos::HIP>::~GraphImpl() {
   m_execution_space.fence("Kokkos::GraphImpl::~GraphImpl: Graph Destruction");
   KOKKOS_EXPECTS(m_graph);
   if (m_graph_exec) {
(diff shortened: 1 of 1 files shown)


core/src/HIP/Kokkos_HIP_Graph_Impl.hpp
```diff
@@ -48,26 +48,26 @@ class GraphImpl<Kokkos::HIP> {
...
   GraphImpl& operator=(GraphImpl&&) = delete;
 
-  ~GraphImpl();
+  inline ~GraphImpl();
 
-  explicit GraphImpl(Kokkos::HIP instance);
+  explicit inline GraphImpl(Kokkos::HIP instance);
 
-  void add_node(std::shared_ptr<aggregate_node_impl_t> const& arg_node_ptr);
+  inline void add_node(std::shared_ptr<aggregate_node_impl_t> const& arg_node_ptr);
```
masterleinad, line 55: I would prefer if you added `inline` to the definition instead of the declaration since the former is what causes the multiple defintions.

dalg24 (dismissed):
The `inline` specifier goes on the definition.
https://isocpp.org/wiki/faq/inline-functions#where-to-put-inline-keyword

core/src/HIP/Kokkos_HIP_Graph_Impl.hpp
romintomasetti: Indeed, done!

masterleinad (approved):

Issue #6628, opened 2023-11-28
Nightly Cuda+OpenMP builds fail to link


ndellingwood:
**Describe the bug**

Nightly Cuda+OpenMP builds fail to link after recent updates (output below is for cuda/12.0+gcc/9.3.0):
```
20:19:31 CMakeFiles/Kokkos_CoreUnitTest_DeviceAndThreads.dir/UnitTest_DeviceAndThreads.cpp.o: In function `print_flag(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&)':
20:19:31 tmpxft_0003d11a_00000000-6_UnitTest_DeviceAndThreads.cudafe1.cpp:(.text+0x7f1): undefined reference to `omp_get_max_threads'
20:19:31 CMakeFiles/Kokkos_CoreUnitTest_DeviceAndThreads.dir/UnitTest_DeviceAndThreads.cpp.o: In function `get_max_threads()':
20:19:31 tmpxft_0003d11a_00000000-6_UnitTest_DeviceAndThreads.cudafe1.cpp:(.text+0x171): undefined reference to `omp_get_max_threads'
20:19:31 ../src/libkokkoscore.a(Kokkos_OpenMP.cpp.o): In function `Kokkos::OpenMP::concurrency() const':
20:19:31 tmpxft_0003cfc0_00000000-6_Kokkos_OpenMP.cudafe1.cpp:(.text+0x180): undefined reference to `omp_get_level'
20:19:31 ../src/libkokkoscore.a(Kokkos_OpenMP.cpp.o): In function `Kokkos::OpenMP::in_parallel(Kokkos::OpenMP const&)':
20:19:31 tmpxft_0003cfc0_00000000-6_Kokkos_OpenMP.cudafe1.cpp:(.text+0x2e8): undefined reference to `omp_get_level'
20:19:31 ../src/libkokkoscore.a(Kokkos_OpenMP.cpp.o): In function `Kokkos::OpenMP::impl_thread_pool_size() const':
20:19:31 tmpxft_0003cfc0_00000000-6_Kokkos_OpenMP.cudafe1.cpp:(.text+0x310): undefined reference to `omp_get_level'
20:19:31 ../src/libkokkoscore.a(Kokkos_OpenMP.cpp.o): In function `Kokkos::OpenMP::OpenMP(int)':
[...]

PR #6629, opened 2023-11-28
nvcc wrapper: remove troubling flag to fix 6628

fnrizzi:
fixes https://github.com/kokkos/kokkos/issues/6628 , but still investigating why

### DIFF
--- a/bin/nvcc_wrapper
+++ b/bin/nvcc_wrapper
@@ -229,7 +229,7 @@ do
     fi
     ;;
   #Handle known nvcc args
-  --dryrun|-dryrun|--verbose|-v|--keep|-keep|--source-in-ptx|-src-in-ptx|--keep-dir*|-keep-dir*|-G|-lineinfo|--generate-line-info|-extended-lambda|-expt-extended-lambda|-expt-relaxed-constexpr|--resource-usage|-res-usage|-fmad=*|--use_fast_math|-use_fast_math|--Wext-lambda-captures-this|-Wext-lambda-captures-this)
+  --dryrun|-dryrun|--verbose|--keep|-keep|--source-in-ptx|-src-in-ptx|--keep-dir*|-keep-dir*|-G|-lineinfo|--generate-line-info|-extended-lambda|-expt-extended-lambda|-expt-relaxed-constexpr|--resource-usage|-res-usage|-fmad=*|--use_fast_math|-use_fast_math|--Wext-lambda-captures-this|-Wext-lambda-captures-this)
     cuda_args="$cuda_args $1"
     ;;
   #Handle more known nvcc args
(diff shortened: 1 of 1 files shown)


ndellingwood (approved):

fnrizzi:
> I's like to understand better what the problem here is.

i agree, i was looking into it

ndellingwood:
For reference, the nvcc_wrapper impacted Cuda+OpenMP nightly builds for multiple combos of cuda + gcc that were tested (cuda/11.0,1,2,4,7 and cuda/12.0 + gcc/8.3,5 and gcc/9.2)

masterleinad:
Ultimately, `CMake`s `find_package` doesn't seem to handle to handle `-v` well when used for `nvcc` (instead of the host compiler). The behavior for `--verbose` is the same but requesting verbose `Makefiles` adds `-v` to the compiler flags and not `--verbose`. It would remove `--verbose` as well, though, to be consistent if we decide to remove `-v` from the flags `nvcc` handles.

ndellingwood:
@masterleinad interesting, so the issue is cmake-nvcc interaction? I suppose this did not occur for example in the CUDA-12.2-NVHPC CI build (which tests with OpenMP and CUDA) since nvc++ was the compiler rather than nvcc?

masterleinad:
> I suppose this did not occur for example in the CUDA-12.2-NVHPC CI build (which tests with OpenMP and CUDA) since nvc++ was the compiler rather than nvcc?

Possibly, I think `-DCMAKE_VERBOSE_MAKEFILE=ON` triggered the issue with `nvcc` at least.

ndellingwood:
@masterleinad are you okay to move forward with this PR for avoiding the issue?

masterleinad:
> @masterleinad are you okay to move forward with this PR for avoiding the issue?

As I said before, I would prefer to remove `--verbose` for consistency (even if it's not triggered right now).

fnrizzi:
@masterleinad done

masterleinad (approved):

cz4rs (approved):

dalg24:
> > @masterleinad are you okay to move forward with this PR for avoiding the issue?
> 
> As I said before, I would prefer to remove `--verbose` for consistency (even if it's not triggered right now).

Why is it ok to remove?  That flag was supported before #6615

masterleinad:
> Why is it ok to remove? That flag was supported before #6615

My only concern is that we would run into the same problems if `CMake` decides to use `--verbose` instead of `-v`. Note that these flags would still not be ignored but just passed to the host compiler instead.

crtrott (changes requested):
I am against removing --verbose

crtrott (approved):

PR #6631, opened 2023-11-28
[4.2.01] Add warp sync for Cuda parallel reduce to avoid race condition #6630

tcclevenger:
Cherry-picking https://github.com/kokkos/kokkos/pull/6630.


masterleinad:
Why is this crucial for the patch release?

crtrott (approved):

PR #6632, opened 2023-11-29
Moving a View outside a parallel region should not increase the use count

dalg24:
Initially reported here https://github.com/kokkos/kokkos/pull/6619#issuecomment-1831402112

`ViewTracker` has a user-defined copy constructor which prevents the compiler to generate the move constructor and a move assignment operator.  This means that the view reference counter increases when the view is moved which leads to resources not being released (that is memory leak).

This PR fixes the issue and ensures that use count is not increased when a view is moved.

### DIFF
--- a/core/src/impl/Kokkos_ViewTracker.hpp
+++ b/core/src/impl/Kokkos_ViewTracker.hpp
@@ -42,12 +42,18 @@ struct ViewTracker {
 
   track_type m_tracker;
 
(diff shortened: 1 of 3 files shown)


masterleinad (approved):

dalg24:
To clarify, memory leaks did not actually occur because the shared allocation tracker was preserved on the moved-from view.

Issue #6381, opened 2023-08-23
Update `clang-format` version


cz4rs:
Follow-up to #6376, see https://github.com/kokkos/kokkos/issues/6376#issuecomment-1690544398.

It seems that he have agreed on using `constexpr West` in Kokkos. Update `clang-format` version used in CI and adjust `.clang-format` accordingly.

Relevant option: https://clang.llvm.org/docs/ClangFormatStyleOptions.html#qualifierorder.

cz4rs:
Note: we need to pick `clang-format` version, we are currently using `8.0.0`. `QualifierOrder` was introduced in 14.

masterleinad:
Switching to clang-format 16 would change around 1000 lines. That's really not too bad in my opinion.

PR #6634, opened 2023-11-29
Replace ubuntu:18.04 with ubuntu:20.04 as base image of clang-format check

Rombur:
We use Ubuntu 18.04 as base image for clang-format check in our CI but it has reached end-of-line. This is fine if the machine as a cached image of Ubuntu 18.04 but when it doesn't there is an error when we try to install new packages. This PR bumps the image to Ubuntu 20.04.

### DIFF
--- a/scripts/docker/Dockerfile.clang
+++ b/scripts/docker/Dockerfile.clang
@@ -1,49 +1,13 @@
-FROM ubuntu:18.04
+FROM ubuntu:20.04
 
 RUN apt-get update && apt-get install -y \
         bc \
         git \
         build-essential \
+        clang-format-8 \
         wget \
-        ccache \
         && \
     apt-get clean && \
     rm -rf /var/lib/apt/lists/*
 
-RUN KEYDUMP_URL=https://cloud.cees.ornl.gov/download && \
-    KEYDUMP_FILE=keydump && \
-    wget --quiet ${KEYDUMP_URL}/${KEYDUMP_FILE} && \
-    wget --quiet ${KEYDUMP_URL}/${KEYDUMP_FILE}.sig && \
-    gpg --import ${KEYDUMP_FILE} && \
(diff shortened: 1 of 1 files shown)


dalg24:
> I don't think we need cmake to run clang-format, so I've removed it. I will add it back if it's necessary.

You are correct.  It dates back from when we were using that same image in another build.

dalg24 (commented):
https://github.com/kokkos/kokkos/blob/573d697302c3a23f6a376f14cc937925249a28c0/scripts/docker/Dockerfile.clang#L23
We explicitly download LLVM/Clang binaries that were created for Ubuntu 18.04
Is that not going to be an issue?

Rombur:
> Is that not going to be an issue?

Yeah it doesn't work. LLVM does not provide binaries for Ubuntu versions newer than 18.04. We will need to install clang-format differently

dalg24:
Maybe it is time to upgrade to a newer clang-format 🤮

masterleinad:
`Ubuntu-20.04` seems to have `clang-format-8`, see https://packages.ubuntu.com/search?suite=default&section=all&arch=any&keywords=clang-format-8&searchon=names.

masterleinad:
What's the problem with Ubuntu 18.04 anyway? AFAICT, `docker` still has the corresponding image.

cz4rs:
Related: #6381. I guess it's the right time to do it since the `4.2.0` release is out.

Rombur:
> What's the problem with Ubuntu 18.04 anyway? AFAICT, docker still has the corresponding image.

Sure the image is still there but like I said in the description of the PR, it crashes when we do `apt update`. There are no issue if the image was cached but if you get a machine where the image was not cached we get an error.

masterleinad (approved):

cz4rs (approved):

PR #6636, opened 2023-11-29
Revisit view move semantics

dalg24:
This is a draft, intended as a basis for discussion.
Such a change in behavior would require guarding the code to allow reverting to the old behavior.

The current behavior (i.e. in the develop branch) is the following:
* moving a view `a` to a view `b` effectively makes a (shallow) copy, the use counter is increased by one and both `a` and `b` manage the same original allocation.

The proposed behavior (this PR), originally suggested in #6619, is:
* moving a view `a` to a view `b` transfers ownership to `b` and `a` changes its state to a default constructed object.

### DIFF
--- a/core/unit_test/TestViewMove.hpp
+++ b/core/unit_test/TestViewMove.hpp
@@ -55,4 +55,53 @@ TEST(TEST_CATEGORY, view_move_and_use_count) {
           v2.data(), v2.extent(0), v2.extent(1)));
 }
 
+// Check that moving a View leaves the moved-from object in a state equivalent
+// to being default constructed
+// returns the number of errors encountered
+template <class ViewType>
+KOKKOS_FUNCTION int check_moved_from_view_state(ViewType v) {
+  int err = 0;
+
+  ViewType w(std::move(v));  // move construction
+  if (v != ViewType()) {
+    Kokkos::printf("failed moved-from view after calling move constructor\n");
+    ++err;
+  }
+
+  v = std::move(w);  // move assignment
+  if (w != ViewType()) {
+    Kokkos::printf(
+        "failed moved-from view after calling move assignment operator\n");
+    ++err;
+  }
+
+  return err;
+}
+
+template <class ViewType>
+void test_moved_from_view(ViewType v) {
+  EXPECT_EQ(check_moved_from_view_state(v), 0) << "outside parallel region";
+
+  using ExexutionSpace = typename ViewType::execution_space;
+  int errors;
+  Kokkos::parallel_reduce(
+      Kokkos::RangePolicy<ExexutionSpace>(0, 1),
+      KOKKOS_LAMBDA(int, int& err) { err += check_moved_from_view_state(v); },
+      errors);
+  EXPECT_EQ(errors, 0) << "within parallel region";
+}
+
+TEST(TEST_CATEGORY, view_moved_from) {
+  using ExecutionSpace = TEST_EXECSPACE;
+
+  test_moved_from_view(Kokkos::View<int, ExecutionSpace>("v0"));
+  test_moved_from_view(Kokkos::View<float*, ExecutionSpace>("v1", 1));
+  Kokkos::View<double**, ExecutionSpace> v2("v2", 1, 2);
+  test_moved_from_view(Kokkos::View<double**, ExecutionSpace>(
+      v2.data(), v2.extent(0), v2.extent(1)));
+  test_moved_from_view(Kokkos::View<double**, ExecutionSpace,
+                                    Kokkos::MemoryTraits<Kokkos::Unmanaged>>(
+      v2.data(), v2.extent(0), v2.extent(1)));
+}
+
 }  // namespace
--- a/core/unit_test/TestViewMove.hpp
+++ b/core/unit_test/TestViewMove.hpp
@@ -1,0 +1,58 @@
+//@HEADER
+// ************************************************************************
+//
+//                        Kokkos v. 4.0
+//       Copyright (2022) National Technology & Engineering
+//               Solutions of Sandia, LLC (NTESS).
+//
+// Under the terms of Contract DE-NA0003525 with NTESS,
+// the U.S. Government retains certain rights in this software.
+//
+// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
+#include <Kokkos_Core.hpp>
+
+#include <gtest/gtest.h>
+
+namespace {
+
+// Check that moving a View outside a parallel region does not increase the
+// number of views managing the allocation.
+template <class ViewType>
+void test_moving_view_does_not_change_use_count(ViewType v) {
+  auto* const ptr = v.data();
+  auto const cnt  = v.use_count();
+
+  ViewType w(std::move(v));  // move construction
+  EXPECT_EQ(w.use_count(), cnt);
+  EXPECT_EQ(w.data(), ptr);
+
+  v = std::move(w);  // move assignment
+  EXPECT_EQ(v.use_count(), cnt);
+  EXPECT_EQ(v.data(), ptr);
+}
+
+TEST(TEST_CATEGORY, view_move_and_use_count) {
+  using ExecutionSpace = TEST_EXECSPACE;
+
+  test_moving_view_does_not_change_use_count(
+      Kokkos::View<int, ExecutionSpace>("v0"));
+
+  test_moving_view_does_not_change_use_count(
+      Kokkos::View<float*, ExecutionSpace>("v1", 1));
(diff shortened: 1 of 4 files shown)


romintomasetti:
It seems that since #7427, we get the move semantic by default (because the `mdspan`-based implementation is used).

Meaning that the following test now is fine:
```c++
// This is fine with KOKKOS_ENABLE_IMPL_MDSPAN defined.
TEST(View, move_semantics)
{
    using view_t = Kokkos::View<double*, execution_space>;

    view_t source("to be moved", 5);

    EXPECT_EQ  (source.use_count(), 1);
    EXPECT_TRUE(source.is_allocated());

    const view_t dest = std::move(source);

    EXPECT_EQ  (source.use_count(), 0) << "Move semantic broken.";
    EXPECT_TRUE(source.is_allocated()) << "I'm expecting 'false' here but...";

    EXPECT_EQ  (dest.use_count(), 1) << "Move semantic broken.";
    EXPECT_TRUE(dest.is_allocated()) << "This is not possible.";
}
```

However, there is still something weird for the `is_allocated`. I'm expecting that `is_allocated` returns `false` if the `use_count` is 0?

masterleinad:
This needs
```diff
diff --git a/core/src/View/MDSpan/Kokkos_MDSpan_Accessor.hpp b/core/src/View/MDSpan/Kokkos_MDSpan_Accessor.hpp
index b501690c2..26b1c681e 100644
--- a/core/src/View/MDSpan/Kokkos_MDSpan_Accessor.hpp
+++ b/core/src/View/MDSpan/Kokkos_MDSpan_Accessor.hpp
@@ -301,6 +301,28 @@ class ReferenceCountedDataHandle {
       const ReferenceCountedDataHandle<OtherElementType, memory_space>& other)
       : m_tracker(other.m_tracker), m_handle(other.m_handle) {}
 
+  KOKKOS_FUNCTION ReferenceCountedDataHandle(
+      const ReferenceCountedDataHandle& other) = default;
+
+  KOKKOS_FUNCTION ReferenceCountedDataHandle(ReferenceCountedDataHandle&& other)
+      : m_tracker(std::move(other.m_tracker)),
+        m_handle(std::move(other.m_handle)) {
+    other.m_handle = nullptr;
+  }
+
+  KOKKOS_FUNCTION ReferenceCountedDataHandle& operator=(
+      const ReferenceCountedDataHandle& other) = default;
+
+  KOKKOS_FUNCTION ReferenceCountedDataHandle& operator=(
+      ReferenceCountedDataHandle&& other) {
+    m_tracker      = std::move(other.m_tracker);
+    m_handle       = std::move(other.m_handle);
+    other.m_handle = nullptr;
+    return *this;
+  }
+
+  ~ReferenceCountedDataHandle() = default;
+
   template <
       class OtherElementType, class OtherSpace,
       class = std::enable_if_t<
```
for the new `View` implementation. Moving a member pointer doesn't set the source pointer to `nullptr` by default.

romintomasetti:
@dalg24 Now is the time to revive this one ! :crossed_swords: 

Here is why:
```c++
TEST(view, moveable) {
  using view_t = Kokkos::View<int, Kokkos::HostSpace>;

  view_t src("src");
  view_t dst(std::move(src));

#if defined(KOKKOS_ENABLE_IMPL_MDSPAN)
  printf("Use mdspan.\n");
  ASSERT_EQ(dst.use_count(), 1);
#elif defined(KOKKOS_ENABLE_IMPL_VIEW_LEGACY)
  printf("Using legacy view.");
  ASSERT_EQ(dst.use_count(), 2);
#else
  #error "This should not happen."
#endif
}
```

masterleinad:
Given that we are close to Kokkos 5, I'm not sure we care about consistency for the move behavior wrt the use count between the mdspan-based and the legacy implementation. There are other differences that are much less subtle.

masterleinad:
We want to test the behavior, though.

core/unit_test/TestViewMove.hpp
```diff
@@ -0,0 +1,137 @@
...
+  }
+
+  // NOLINTEND(bugprone-use-after-move)
+
+  return err;
+}
+
+template <class ViewType>
+void test_moved_from_view(ViewType v) {
+  // EXPECT_EQ(check_moved_from_view_state(v), 0) << "outside parallel region";
```
masterleinad: We don't reset `extents` or `span` after moving.

core/unit_test/TestViewMove.hpp
```diff
@@ -0,0 +1,137 @@
...
+template <class ViewType>
+void test_moving_view_does_not_change_use_count(ViewType v) {
+  auto* const ptr = v.data();
+  auto const cnt  = v.use_count();
+
+  // NOLINTBEGIN(bugprone-use-after-move)
+
+  ViewType w(std::move(v));  // move construction
+#ifdef KOKKOS_ENABLE_IMPL_VIEW_LEGACY
+  if (w.use_count() != 0) EXPECT_EQ(w.use_count(), cnt + 1);
```
dalg24: Drop the guard and use a ternary or have another assertion in the else-branch

dalg24 (commented):
What behavior do we get when/o the changes in the source files?

masterleinad:
> What behavior do we get when/o the changes in the source files?

The data pointer doesn't change (movnig a pointer doesn't change it's value).

dalg24:
> > What behavior do we get when/o the changes in the source files?
> 
> The data pointer doesn't change (movnig a pointer doesn't change it's value).

Why did you change it as part of this PR?
I thought we agreed on matching the test to the current behavior and to propose changes in follow ups.

masterleinad:
> Why did you change it as part of this PR?

I thought I had pushed these changes before (they are dated from July 11) already when trying to remember the exact status of the pull request. Since it turned out I had found a solution, I wanted to share what it takes to make the tests (almost) work.

dalg24:
> > Why did you change it as part of this PR?
> 
> I thought I had pushed these changes before (they are dated from July 11) already when trying to remember the exact status of the pull request. Since it turned out I had found a solution, I wanted to share what it takes to make the tests (almost) work.

Please save them to the side and have the test match the current behavior.

core/unit_test/TestViewMove.hpp
```diff
@@ -0,0 +1,148 @@
...
+  auto const cnt  = v.use_count();
+
+  // NOLINTBEGIN(bugprone-use-after-move)
+
+  ViewType w(std::move(v));  // move construction
+#ifdef KOKKOS_ENABLE_IMPL_VIEW_LEGACY
+  if (w.use_count() != 0)
+    EXPECT_EQ(w.use_count(), cnt + 1);
+  else
+    EXPECT_EQ(w.use_count(), 0);
```
crtrott, line 23: Why? You already checked that its zero.

core/unit_test/TestViewMove.hpp
```diff
@@ -0,0 +1,148 @@
...
+  EXPECT_EQ(w.use_count(), cnt);
+#endif
+  EXPECT_EQ(w.data(), ptr);
+#ifdef KOKKOS_ENABLE_IMPL_VIEW_LEGACY
+  EXPECT_EQ(v.use_count(), w.use_count());
+  EXPECT_EQ(v.data(), w.data());
+#else
+  EXPECT_EQ(v.use_count(), 0);
+  // FIXME should be nullptr
+  EXPECT_EQ(v.data(), ptr);
```
crtrott: What about the extents?

dalg24 (commented):
Thank you
I can't formally approve but I am ok with the current version

core/unit_test/TestViewMove.hpp
masterleinad: Damien wanted to add it in https://github.com/kokkos/kokkos/pull/6636#discussion_r2389439076.

core/unit_test/TestViewMove.hpp
masterleinad: What do you mean? Are you referring to line 121?

crtrott: ah nvm test says only testing use count.

cwpearson (approved):

core/unit_test/TestViewMove.hpp
nmm0: Maybe we should fix (I presume the data handle) so that it is nullptr, this seems like it could mess someone up if it isn't?

dalg24: The intent here is to capture the current behavior and to consider whether and how to change the status quo in follow up work.

masterleinad: For reference, [`780b8ca` (#6636)](https://github.com/kokkos/kokkos/pull/6636/commits/780b8cab32981c6e8ba444bcb2c696471e4857dc) would fix it.

core/unit_test/TestViewMove.hpp
dalg24: I know it looks silly but it comes handy if you retrace a failure in the gtest logs

PR #6641, opened 2023-12-01
UnorderedMap: promote `invalid_index` to public and use `Experimental::finite_max_v` for its value

romintomasetti:
As discussed in #6610 with @masterleinad, it makes sense that `Kokkos::UnorderedMap::invalid_index` is promoted to a `public` member, because `Kokkos::UnorderedMap::find` must be able to return a distinguished value to denote that no element with the given key could be found. As `Kokkos::UnorderedMap::find` is `public`, so does ``Kokkos::UnorderedMap::invalid_index`.

As discussed with @dalg24, the value of `invalid_index` is now computed with the nice `Kokkos::Experimental::finite_max_v` that is more robust against the type than `~static_cast<type>(0)`.

### DIFF
--- a/containers/src/Kokkos_UnorderedMap.hpp
+++ b/containers/src/Kokkos_UnorderedMap.hpp
@@ -254,6 +254,9 @@ class UnorderedMap {
   static constexpr bool is_modifiable_map = has_const_key && !has_const_value;
   static constexpr bool is_const_map      = has_const_key && has_const_value;
 
+  static constexpr size_type invalid_index =
+      Experimental::finite_max_v<size_type>;
+
   using insert_result = UnorderedMapInsertResult;
 
   using HostMirror =
@@ -263,7 +266,6 @@ class UnorderedMap {
   //@}
 
  private:
-  enum : size_type { invalid_index = ~static_cast<size_type>(0) };
 
   using impl_value_type = std::conditional_t<is_set, int, declared_value_type>;
 


dalg24:
Please provide a rational explaining why this is needed (in the description)

containers/src/Kokkos_UnorderedMap.hpp
```diff
@@ -263,7 +263,9 @@ class UnorderedMap {
   //@}
 
  private:
-  enum : size_type { invalid_index = ~static_cast<size_type>(0) };
+  enum : size_type {
+    invalid_index = KOKKOS_IMPL_INVALID_INDEX_TYPE(size_type)
```
romintomasetti, line 267: @dalg24 proposes to use `Kokkos::Experimental::NumericTraits::FINITE_MAX` (see doc, core numerics)

dalg24:
```C++
public:
  static constexpr size_type invalid_index = Experimental::finite_max_v<size_type>;

**Review by dalg24 — COMMENTED** [review]
Make sure you describe the need for `invalid_index` to be public (`UnorderedMap::find` returning a distinguished value to denote that no element with the given key could be found)

File: containers/src/Kokkos_UnorderedMap.hpp (hunk 1 above)
Comment (romintomasetti) [reply containers/src/Kokkos_UnorderedMap.hpp:267]: Thanks!

**Comment by dalg24** [comment]
Digging more into this, we define `UnorderedMapInvalidIndex`
https://github.com/kokkos/kokkos/blob/a41dba5860c019f503822854f5ba8485937e1733/containers/src/Kokkos_UnorderedMap.hpp#L43
and `size_type` is always unsigned
https://github.com/kokkos/kokkos/blob/a41dba5860c019f503822854f5ba8485937e1733/containers/src/Kokkos_UnorderedMap.hpp#L232
so users can already compare the value returned by `find()` with `UnorderedMapInvalidIndex`.
```C++
auto index = map.find(key);
if (index == UnorderedMapInvalidIndex) {
  // ...
```

It is unclear to me what we are fixing here.  It is a bit silly we define that private `invalid_index` and not use `UnorderedMapInvalidIndex` but user-facing gap as far as I can tell.
What am I missing?

romintomasetti:
I wasn't even aware of `UnorderedMapInvalidIndex`...

dalg24:
> What about moving these two structs into `UnorderedMap` and removing `UnorderedMapInvalidIndex`? Users would then use:
> 
> * `Kokkos::Unorderedmap::invalid_index`
> * `Kokkos::UnorderedMap::InsertResult`
> * `Kokkos::UnorderedMap::InsertOpTypes`
> 
> I don't know if this is a nice style or not, but considering that these 3 types are in the end intrinsic to `UnorderedMap`...

I think the current design leads to less symbols being emitted.  If you make them member you will have one version per instantiation of `UnorderedMap`.  And the user could not refer to them directly, they would need to spell out at least the `Key` and `Value` template parameters.

romintomasetti:
> > What about moving these two structs into `UnorderedMap` and removing `UnorderedMapInvalidIndex`? Users would then use:
> > 
> > * `Kokkos::Unorderedmap::invalid_index`
> > * `Kokkos::UnorderedMap::InsertResult`
> > * `Kokkos::UnorderedMap::InsertOpTypes`
> > 
> > I don't know if this is a nice style or not, but considering that these 3 types are in the end intrinsic to `UnorderedMap`...
> 
> I think the current design leads to less symbols being emitted. If you make them member you will have one version per instantiation of `UnorderedMap`. And the user could not refer to them directly, they would need to spell out at least the `Key` and `Value` template parameters.

If you care about the number of symbols emitted, would something like this be OK for you ? Other "stuff independent from the key/value type" like the invalid index could be encapsulated in this base class...
```c++
class UnorderedMapBase {
    public:
        struct InsertResult {
            void print() const { printf("%s\n", __PRETTY_FUNCTION__); }
        };
};

template <typename T>
class UnorderedMap : public UnorderedMapBase {
    public:
        InsertResult insert() const { return InsertResult {} ; }
};
```

Let me try to summarize the discussion:
1. There are 2 different invalid index variables, one in the `Kokkos` namespace named `UnorderedMapInvalidIndex` and one defined in the unordered map class (`invalid_index`) and used internally. I bet this will lead to strange bugs in the future so this duplication should be tackled IMO.
2. Intrinsics of the unordered map (`UnorderedMapInvalidIndex`, `UnorderedMapInsertResult`, `UnorderedMapInsertOpTypes`) are defined in the `Kokkos` namespace because "they don't depend on the key/value type". I guess this is clearly the case for `UnorderedMapInvalidIndex` and `UnorderedMapInsertResult`, but not really for `UnorderedMapInsertOpTypes` that is in fact templated over the value type and is therefore tightly bound to the value type of the unordered map.

With the code above, we would be able to define intrinsics of the unordered map that do not depend on the key/value types in the `UnorderedMapBase`, and intrinsics that do depend on the key/value type in the templated `UnorderedMap`.

What do you think ?

dalg24:
> 1. There are 2 different invalid index variables, one in the `Kokkos` namespace named `UnorderedMapInvalidIndex` and one defined in the unordered map class (`invalid_index`) and used internally. I bet this will lead to strange bugs in the future so this duplication should be tackled IMO.
> 2. Intrinsics of the unordered map (`UnorderedMapInvalidIndex`, `UnorderedMapInsertResult`, `UnorderedMapInsertOpTypes`) are defined in the `Kokkos` namespace because "they don't depend on the key/value type". I guess this is clearly the case for `UnorderedMapInvalidIndex` and `UnorderedMapInsertResult`, but not really for `UnorderedMapInsertOpTypes` that is in fact templated over the value type and is therefore tightly bound to the value type of the unordered map.
> 
> With the code above, we would be able to define intrinsics of the unordered map that do not depend on the key/value types in the `UnorderedMapBase`, and intrinsics that do depend on the key/value type in the templated `UnorderedMap`.
> 
> What do you think ?

1. Sure.  Then we are talking about a refactor and is a matter of much lesser urgency.
2. What issue would this resolve?  Why would that be better than the status quo?

romintomasetti:
Let's close this PR. The benefits are not really interesting, and no one uses this whatsoever.

PR #6642, opened 2023-12-01
tools(profiling): fix typo Kokkos_Tools_Optim[i]zationGoal

romintomasetti:
This PR accompanies kokkos/kokkos-tools/pull/221.

It fixes the typo `Kokkos_Tools_OptimzationGoal` to `Kokkos_Tools_OptimizationGoal`.

### Risks

- No risk: For people fetching this change, there would be a clear compilation error if they relied on the old name. They would simply need to fix the typo on their side as well.
- Minor risk: For tools that have been compiled in the past, this might be a breaking change. However, we deem there are not so many tools out there that rely on this "optimization goal" feature. We could mitigate this risk by making the change enclosed in a "deprecated code" region.

### DIFF
--- a/core/src/impl/Kokkos_Profiling_C_Interface.h
+++ b/core/src/impl/Kokkos_Profiling_C_Interface.h
@@ -154,7 +154,7 @@ enum Kokkos_Tools_OptimizationType {
   Kokkos_Tools_Maximize
 };
 
-struct Kokkos_Tools_OptimzationGoal {
+struct Kokkos_Tools_OptimizationGoal {
   size_t type_id;
   enum Kokkos_Tools_OptimizationType goal;
 };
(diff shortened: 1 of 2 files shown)


dalg24 (commented):
Although obviously this was a typo, this would be a breaking change.
We need to discuss whether it is acceptable/worth to proceed.

masterleinad (commented):
I think we should just add a deprecated alias.

dalg24:
Tuning is still considered experimental and it is unlikely to be broadly used so I am not necessarily opposed to fix.
I supposed if we are really worried we could increment `KOKKOSP_INTERFACE_VERSION`.

dalg24:
> I think we should just add a deprecated alias.

A C alias? :)

romintomasetti:
So, what's next then ? :laughing:

masterleinad (approved):

dalg24:
Ignoring Jenkins CI

PR #6645, opened 2023-12-04
add missing header, fix #6644

### LINKED ISSUE #6644: Use of std::setprecision without including iomanip in TestStdAlgorithms[X]Scan.cpp
The following test files in `kokkos/algorithms/unit_tests` have calls to `std::setprecision`, but there is no `#include<iomanip>` in any of them or in `TestStdAlgorithmsCommon.hpp`, which they include.

    TestStdAlgorithmsExclusiveScan.cpp
    TestStdAlgorithmsInclusiveScan.cpp
    TestStdAlgorithmsTransformExclusiveScan.cpp
    TestStdAlgorithmsTransformInclusiveScan.cpp

fnrizzi:


fnrizzi:
I think this is just a byproduct of when this was done first time. 
Now that I look again, we test also for the custom value type, but we could just convert it to double explicitly without static cast. 

I can do a follow up PR to fix it

dalg24 (approved):

PR #6581, opened 2023-11-06
Kokkos SIMD: macros for LLVM-based oneAPI compilers

japlews:

PR #6647, opened 2023-12-05
Fixup test math functions ulp should be an integral number

dalg24:
ulp stands for Unit in the Last Place.
Making it a double was a poor choice.  This PR suggest to change it to an int.

Triggered by https://github.com/kokkos/kokkos/pull/6581/files#r1416118599


crtrott (approved):

dalg24:
Ignoring failures

PR #6650, opened 2023-12-06
SYCL: Force inlining of Kokkos::printf

masterleinad:
It turns out that `sycl::ext::oneapi::experimental::printf` needs to have a compile-time formatting string and this only works well right now if calls to `Kokkos::printf` are inlined. Using `-O0`, this isn't the case in practice, and compilation would fail.
This, this pull request proposes to force inline `Kokkos::printf` for `SYCL`.

Just duplicating the whole function instead of using multiple `#if KOKKOS_ENABLE_SYCL` switches, seemed the cleaner solution (since most of the source code lines need to be duplicated anyway).

### DIFF
--- a/core/src/Kokkos_Printf.hpp
+++ b/core/src/Kokkos_Printf.hpp
@@ -30,15 +30,18 @@ namespace Kokkos {
 // In contrast to std::printf, return void to get a consistent behavior across
 // backends. The GPU backends always return 1 and NVHPC only compiles if we
 // don't ask for the return value.
-template <typename... Args>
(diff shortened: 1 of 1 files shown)


dalg24 (commented):
Any reason not to force inline when deferring to :::printf as well?

masterleinad:
Retest this please.

crtrott:
Can we try for OpenMPTarget with Intel GPUs 

```c++
namespace Kokkos {
 using ::printf;
}
```

cz4rs (approved):

PR #6651, opened 2023-12-06
CI: Don't use rocm-docker runners

masterleinad:
The `rocm-docker` runners are not working right now and cause spurious CI failures. In particular, the `clang-format` check is failing when running on those machines preventing from any other `jenkins` CI to run. Thus, this pull request proposes to
- not use `rocm-docker` for `clang-fromat`. We should have enough other runners so we don't need to burden these scarce resources with it.
- disable `HIP` CI for now until those machines are working again.

### DIFF
--- a/.jenkins
+++ b/.jenkins
@@ -22,7 +22,7 @@ pipeline {
                 dockerfile {
                     filename 'Dockerfile.clang'
                     dir 'scripts/docker'
-                    label 'nvidia-docker || rocm-docker || docker'
+                    label 'nvidia-docker || docker'
                     args '-v /tmp/ccache.kokkos:/tmp/ccache'
                 }
             }
--- a/.jenkins
+++ b/.jenkins
@@ -135,82 +135,82 @@ pipeline {
                         }
                     }
                 }
-                stage('HIP-ROCm-5.2') {
-                    agent {
-                        dockerfile {
-                            filename 'Dockerfile.hipcc'
-                            dir 'scripts/docker'
-                            additionalBuildArgs '--build-arg BASE=rocm/dev-ubuntu-20.04:5.2'
-                            label 'rocm-docker && vega'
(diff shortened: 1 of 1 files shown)


cz4rs (approved):
Looks good to me. Spurious `Clang-Format` failures are pretty annoying right now.

fnrizzi (approved):
i think make sense

fnrizzi:
maybe it is worth just omitting the clang format on all jenkins runs since we have it as github check?

dalg24:
> maybe it is worth just omitting the clang format on all jenkins runs since we have it as github check?

No this was intentional to reduce the pressure on the CI.  If it is not properly formatted, fail fast.

fnrizzi:
> > maybe it is worth just omitting the clang format on all jenkins runs since we have it as github check?
> 
> No this was intentional to reduce the pressure on the CI. If it is not properly formatted, fail fast.

ok, but what is the point of having the same format check run on both github and jenkins?

fnrizzi:
or you mean that the format is used a kind of "filter" to reduce pressure?

masterleinad:
> or you mean that the format is used a kind of "filter" to reduce pressure? that seems to me just procrastinating the issue though

Yes, we run it first so that we at least avoid one extra testing cycle if the indentation needs to be updated in a pull request.
The GitHub workflow just makes it more convenient to see that the indentation needs to be fixed so that pull request authors don't even have to bother clicking through the `jenkins` interface.

fnrizzi:
> > or you mean that the format is used a kind of "filter" to reduce pressure? that seems to me just procrastinating the issue though
> 
> Yes, we run it first so that we at least avoid one extra testing cycle if the indentation needs to be updated in a pull request. The GitHub workflow just makes it more convenient to see that the indentation needs to be fixed so that pull request authors don't even have to bother clicking through the `jenkins` interface.

ok!

PR #6652, opened 2023-12-06
Improve handling of printf in OMPT on Intel GPUs

masterleinad:
Addressing https://github.com/kokkos/kokkos/pull/6650#issuecomment-1843581455.
We are currently not printing anything when using `OpenMPTarget` on `Intel` GPUs and a non-zero number of arguments. We discussed that we rather give up on controlling the interface in this case and import `printf` from the global namespace.

### DIFF
--- a/core/src/Kokkos_Printf.hpp
+++ b/core/src/Kokkos_Printf.hpp
@@ -30,6 +30,9 @@ namespace Kokkos {
 // In contrast to std::printf, return void to get a consistent behavior across
 // backends. The GPU backends always return 1 and NVHPC only compiles if we
 // don't ask for the return value.
+#if defined(KOKKOS_ENABLE_OPENMPTARGET) && defined(KOKKOS_ARCH_INTEL_GPU)
+using ::printf;
+#else
 template <typename... Args>
 KOKKOS_FORCEINLINE_FUNCTION void printf(const char* format, Args... args) {
 #ifdef KOKKOS_ENABLE_SYCL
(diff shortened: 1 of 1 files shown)


masterleinad:
@rgayatri23 ping

rgayatri23 (approved):

core/src/Kokkos_Printf.hpp
```diff
@@ -39,15 +42,13 @@ KOKKOS_FORCEINLINE_FUNCTION void printf(const char* format, Args... args) {
   else
     sycl::ext::oneapi::experimental::printf(format, args...);
 #else
-  if constexpr (sizeof...(Args) == 0) ::printf("%s", format);
-    // FIXME_OPENMPTARGET non-string-literal argument used in printf is not
-    // supported for spir64
-#if !(defined(KOKKOS_ENABLE_OPENMPTARGET) && defined(KOKKOS_ARCH_INTEL_GPU))
+  if constexpr (sizeof...(Args) == 0)
+    ::printf("%s", format);
```
dalg24, line 46: Do we still need handle no arguments passed separately with these changes?

masterleinad: Yes, this still avoids warnings (I don't exactly recall which exact compiler and compiler versions) and was part of the implementation before we made changes for `OpenMPTarget`, see https://github.com/kokkos/kokkos/pull/6083/files.

dalg24 (approved):

dalg24:
Ignoring failures as they are due to storage issue on the test machines.

PR #6653, opened 2023-12-06
Remove deprecation warning for AllocationMechanism for gcc <11.0 

masterleinad:
Related to the conversation in https://kokkosteam.slack.com/archives/C5BGU5NDQ/p1701727560890859.
We see deprecation warnings for GCC versions before release 11 every time a compilation unit includes `Kokkos_HostSpace.hpp`.
This pull request proposes removing the deprecation attribute for `AllocationMechanism` to avoid these warnings. It seems very unlikely that the enum would be used without the `HostSpace` constructor in question that still has the deprecation attribute.

### DIFF
--- a/core/src/Kokkos_HostSpace.hpp
+++ b/core/src/Kokkos_HostSpace.hpp
@@ -75,12 +75,19 @@ class HostSpace {
   /**\brief  Non-default memory space instance to choose allocation mechansim,
    * if available */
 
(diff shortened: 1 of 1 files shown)


dalg24 (approved):
In case it is useful to anyone else https://godbolt.org/z/bbenczfbE

dalg24:
Will need to cherry-pick into the patch release 4.2.01

cz4rs (approved):
Looks good to me.

PR #6659, opened 2023-12-09
Fix msvc cuda develop

crtrott:
Various small fixups for MSVC/CUDA build:
- builtin_unreachable is not defined when using NVCC/MSVC
- missing include in Kokkos_Serial
- host/device lambdas in constexpr if branches not supported on NVCC?MSVC
- some default CTAD not working for some reason with NVCC/MSVC
- half/bhalf type is_nan/is_finite/is_inf etc. hit some non-working system header intrinsic - I suspect it has something to do with the bit comparison manipulation stuff but didn't investigate much

PR #6661, opened 2023-12-11
Add CI for MSVC+Cuda

masterleinad:
Based on top of #6659 that is needed for this new build to pass.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -1,0 +1,36 @@
+name: github-windows
+
+on:
+  push:
+  pull_request:
+
+concurrency:
+  group: ${ {github.event_name }}-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{github.event_name == 'pull_request'}}
+
+permissions:
+  contents: read
+
+jobs:
+  windows-cuda:
+    # Cuda build on Windows
+    name: Windows Cuda
+    runs-on: windows-2022
+
+    steps:
+    - uses: Jimver/cuda-toolkit@v0.2.11
+      id: cuda-toolkit
+      with:
+        cuda: '12.1.0'
+    - uses: actions/checkout@v4
+    - name: configure
+      shell: bash
+      run: |
+        mkdir build
+        mkdir c:/project
+        cd build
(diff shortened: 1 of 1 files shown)


.github/workflows/windows.yml
```diff
@@ -0,0 +1,36 @@
...
+  group: ${ {github.event_name }}-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{github.event_name == 'pull_request'}}
+
+permissions:
+  contents: read
+
+jobs:
+  windows-cuda:
+    # Cuda build on Windows
+    name: Windows Cuda
```
fnrizzi, line 17: can we make the name consistent with the names of the other CI pipelines?

fnrizzi: so that is looks consistent with the other names

.github/workflows/windows.yml
```diff
@@ -0,0 +1,36 @@
...
+      shell: bash
+      run: |
+        mkdir build
+        mkdir c:/project
+        cd build
+        cmake -DKokkos_ENABLE_CUDA=ON -DKokkos_ARCH_VOLTA70=ON -DKokkos_ENABLE_TESTS=ON -DKokkos_ENABLE_COMPILE_AS_CMAKE_LANGUAGE=ON ..
+    - name: build library
+      shell: bash
+      run: |
+        cmake --build build --parallel 2 --config Release
```
fnrizzi, line 36: why not adding both a debug and release build?

dalg24:
Any particular reason for using GitHub workflows rather than AppVeyor for that build?
(Naively asking since the Windows testing is currently on there)

masterleinad:
> Any particular reason for using GitHub workflows rather than AppVeyor for that build?

I managed to do it with `GitHub` workflows and could use an existing action for installing `Cuda` (that caches it). I tried previously with `appveyor`, see https://github.com/kokkos/kokkos/pull/5266 but never really managed to get both builds going the way I wanted to (also it didn't cache `cuda`).
I would consider moving the `appveyor` CI we have right now to `GitHub` as well so that we avoid being split across too many different systems.

.github/workflows/windows.yml
masterleinad: What do you want it to be? `github-windows-cuda / CI`?

crtrott (approved):

PR #6662, opened 2023-12-12
Add CMakeLists.txt for `benchmarks/stream`

cwpearson:
```bash
cmake -DKokkos_ENABLE_BENCHMARKS=ON
make
./benchmarks/stream/Kokkos_stream
```


masterleinad (approved):

dalg24 (approved):
Looks good.

A few thoughts to improve benchmarks in follow up PRs:
* currently we only add executables, meaning we build but we don't actually run them.
* we might want to include "benchmark" in the exe name and more importantly when we register it as a test so that it runs when we do ctest.

The same comments apply to examples/tutorials.

dalg24:
Ignoring unrelated HIP failures (` No space left on device` at checkout time)

romintomasetti:
> Looks good.
> 
> A few thoughts to improve benchmarks in follow up PRs:
> 
>     * currently we only add executables, meaning we build but we don't actually run them.

@dalg24 @cwpearson  In our code base, any benchmark we add is compiled and runs for one iteration in our CICD actions (using `Benchmark::Iterations(IterationCount n)`).

>     * we might want to include "benchmark" in the exe name and more importantly when we register it as a test so that it runs when we do ctest.

We do this as well. Benchmarks are prefixed with `benchmark_*` and added to the `CTest` list.

PR #6664, opened 2023-12-13
Do not negate the dependent true traits helper

dalg24:
This came up in some other work and I don't want to see that practice spread so I am replacing it with a better alternative.

I considered adding an `Impl::always_false` traits but there is a number of things I don't like with `always_true` (starting with the fact it is variadic).

I suggest to assert that the argument type is `void` which would not be legal C++.

### DIFF
--- a/core/src/OpenACC/Kokkos_OpenACC_ParallelReduce_MDRange.hpp
+++ b/core/src/OpenACC/Kokkos_OpenACC_ParallelReduce_MDRange.hpp
@@ -31,8 +31,7 @@ template <class Functor, class Reducer, class Policy,
 struct OpenACCParallelReduceMDRangeHelper {
   OpenACCParallelReduceMDRangeHelper(Functor const&, Reducer const&,
                                      Policy const&) {
-    static_assert(!Kokkos::Impl::always_true<Functor>::value,
-                  "not implemented");
+    static_assert(std::is_void_v<Functor>, "not implemented");
   }
 };
 }  // namespace Kokkos::Experimental::Impl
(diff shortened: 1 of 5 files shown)


crtrott (approved):

cwpearson (approved):

masterleinad (commented):
In my opinion, `always_true`/`always_false` is more expressive.

dalg24:
> In my opinion, `always_true`/`always_false` is more expressive.

Are you saying "not always true" is expressive enough or that you prefer "always false"?

masterleinad:
> Are you saying "not always true" is expressive enough or that you prefer "always false"?

`always_false` is better, I would understand the opposite of `always_true` to be `sometimes_false` instead of `always_false`. I find the intent behind using `void_v` not that intuitive.

dalg24:
What would you prefer then?

I would have suggested
```C++
template<typename T>
inline constexpr bool dependent_false = false;
```
but we have
https://github.com/kokkos/kokkos/blob/cedbf56f64e3cb1e6f11e934c01bff7b1c09964b/core/src/impl/Kokkos_Utilities.hpp#L49-L50

so maybe 
```diff
diff --git a/core/src/impl/Kokkos_Utilities.hpp b/core/src/impl/Kokkos_Utilities.hpp
index 7e2f13056..43c98ed72 100644
--- a/core/src/impl/Kokkos_Utilities.hpp
+++ b/core/src/impl/Kokkos_Utilities.hpp
@@ -49,6 +49,11 @@ struct integral_constant {
 template <typename... Is>
 struct always_true : std::true_type {};

+// type-dependent exrpession that is always false intended for use in
+// static_assert to check "we should never get there"
+template <typename... Deps>
+struct always_false : std::false_type {};
+
 //==============================================================================

 #if defined(__cpp_lib_type_identity)

**Comment by masterleinad** [comment]
Yeah, I would prefer
```
template <typename... Deps>
struct always_false : std::false_type {};
```
and would also be fine with making it non-variadic (even if it'd be inconsistent with `Impl::always_true`).

masterleinad (approved):

seyonglee (approved):

PR #6672, opened 2023-12-14
Don't disable desul atomics if only the Serial backend is enabled

masterleinad:
The tests in https://github.com/kokkos/kokkos/pull/6151 (in particular the Windows build, as well as `ThreadSanitizer` and `AddressSanitizer`) show that we need to use proper atomics for at least `HostSharedPointer` and `SharedAllocationRecord` even in the case that only the `Serial` backend is enabled if multiple threads are used.
This pull request proposes to just remove the special handling of this case to avoid other hard-to-find bugs.
The alternative is, of course, to just fix the two classes mentioned above.

### DIFF
--- a/core/src/Kokkos_Atomics_Desul_Volatile_Wrapper.hpp
+++ b/core/src/Kokkos_Atomics_Desul_Volatile_Wrapper.hpp
@@ -25,173 +25,166 @@ static_assert(false,
 #include <Kokkos_Atomics_Desul_Config.hpp>
 #include <desul/atomics.hpp>
 
-#ifdef KOKKOS_INTERNAL_NOT_PARALLEL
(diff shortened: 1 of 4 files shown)


dalg24:
You are free to investigate how to resolve the race conditions in the unit test but know that this is a much bigger discussion.
I would tend to agree with you, even more so in the context of multiple serial execution space instances but this is in-place because atomics where prohibitively expensive for some downstream who didn't want to do atomic-if-needed in their functors.

crtrott (changes requested):
At a minimum we need an option to enable the current behavior with Serial only. Long term (i.e. Kokkos 5) we could talk to apps to explicitly do this on their side by directly using desul atomics with proper scope.

masterleinad:
Let's make this a draft for now and first fix the relevant spots in #6674.

masterleinad:
Superseded by #6692.

PR #6674, opened 2023-12-14
Replace atomics in HostSharedPtr and SharedAllocationRecord

masterleinad:
Alternative to #6672, extracted from #6151. Instead of removing the special case for atomics when only the `Serial` backend is enabled, this pull request just explicitly uses `desul` atomics for `HostSharedPtr` and `SharedAllocationRecord` which were causing problems in  #6151.


dalg24 (changes requested):
Not wanting to treat these differently

masterleinad:
We discussed that we want to use `desul` explicitly and rather rely on #6692.

PR #6675, opened 2023-12-15
Improve "deep_copy: no copy mechanism" exception message

brian-kelley:
The cases where deep_copy can't be used can be confusing to newer users. This was just brought up at KUG, and I also see a recent issue about it.

Since these cases - cross-space copies with incompatible layouts or non-contiguous - all come down to a single point in the code, so it makes sense to provide some more info in this error message.

Tested with a deep_copy between two non-contiguous subviews.

Before:
```
  what():  Error: Kokkos::deep_copy with no available copy mechanism: v2 to v1
```

After:
```
  what():  Error: Kokkos::deep_copy with no available copy mechanism: from src ("v2") to dst ("v1").
There is no common execution space that can access both src's space
(Host) and dst's space (Cuda), so src and dst
must be contiguous and have the same layout.
```

We could even add more info: which view(s) are noncontiguous? What the strides are? But this is already long for an exception message.

### DIFF
--- a/core/src/Kokkos_CopyViews.hpp
+++ b/core/src/Kokkos_CopyViews.hpp
@@ -22,6 +22,7 @@ static_assert(false,
 #ifndef KOKKOS_COPYVIEWS_HPP_
 #define KOKKOS_COPYVIEWS_HPP_
 #include <string>
+#include <sstream>
 #include <Kokkos_Parallel.hpp>
 #include <KokkosExp_MDRangePolicy.hpp>
 #include <Kokkos_Layout.hpp>
--- a/core/src/Kokkos_CopyViews.hpp
+++ b/core/src/Kokkos_CopyViews.hpp
@@ -614,15 +614,15 @@ void view_copy(const DstType& dst, const SrcType& src) {
 
   if (!DstExecCanAccessSrc && !SrcExecCanAccessDst) {
     std::ostringstream ss;
-    ss << "Error: Kokkos::deep_copy with no available copy mechanism: ";
-    ss << "from src (\"" << src.label() << "\") to dst (\"" << dst.label()
-       << "\").\n";
-    ss << "There is no common execution space that can access both src's "
-          "space\n";
-    ss << "(" << src_memory_space().name() << ") and dst's space ("
-       << dst_memory_space().name() << "), ";
-    ss << "so src and dst\n";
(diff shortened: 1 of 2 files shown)


dalg24 (commented):
I like it.  Please add a test.

(not requested) I would like us to consider calling abort instead of throwing an exception here

brian-kelley:
@dalg24 Just added a test

masterleinad (approved):

dalg24 (approved):

core/src/Kokkos_CopyViews.hpp
```diff
@@ -612,12 +612,17 @@ void view_copy(const DstType& dst, const SrcType& src) {
   };
 
   if (!DstExecCanAccessSrc && !SrcExecCanAccessDst) {
-    std::string message(
-        "Error: Kokkos::deep_copy with no available copy mechanism: ");
-    message += src.label();
-    message += " to ";
-    message += dst.label();
-    Kokkos::Impl::throw_runtime_exception(message);
+    std::ostringstream ss;
```
dalg24, line 615: Include what you use `<iosfwd>`

brian-kelley: Thanks, fixed

core/src/Kokkos_CopyViews.hpp
```diff
@@ -612,12 +613,17 @@ void view_copy(const DstType& dst, const SrcType& src) {
...
+    ss << "Error: Kokkos::deep_copy with no available copy mechanism: ";
+    ss << "from src (\"" << src.label() << "\") to dst (\"" << dst.label()
+       << "\").\n";
+    ss << "There is no common execution space that can access both src's "
+          "space\n";
+    ss << "(" << src_memory_space().name() << ") and dst's space ("
+       << dst_memory_space().name() << "), ";
+    ss << "so src and dst\n";
```
fnrizzi, line 624: since we write the message cleanly, I would suggest spelling out "source view" instead of "src" and "destination view" instead of "dst" so that we avoid using names that are specific to the func parameters

fnrizzi (changes requested):
just a minor request

fnrizzi (approved):

fnrizzi:
retest this please

PR #6676, opened 2023-12-15
GitHub CI: Test with AddressSanitizer

masterleinad:
Related to #6151 and #6674. AddressSanitizer would have detected problems in #6151 without #6674 (if only the `Serial` backend was enabled). Since we are treating atomics special if only the `Serial` backend was enabled (also see #6672), I added a build with AddressSanitizer that only enables `Serial`. I also enabled it for another build (with the `Threads` backend) for which it was convenient.

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -124,8 +124,6 @@ jobs:
           cmake --build builddir --parallel 2
           ccache -s
       - name: Tests
-        env:
-          ASAN_OPTIONS: allocator_may_return_null=1
         working-directory: builddir
         run: ctest --output-on-failure
       - name: Test linking against build dir
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -47,9 +47,17 @@ jobs:
             backend: 'OPENMP'
           - distro: 'ubuntu:latest'
             cxx: 'clang++'
+            cxx_extra_flags: '-fsanitize=address'
+            extra_linker_flags: '-fsanitize=address'
             cmake_build_type: 'RelWithDebInfo'
             backend: 'THREADS'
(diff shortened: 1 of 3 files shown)


masterleinad:
All GitHub CI passes.

ldh4 (approved):
I'm not sure about needing this for `threads`, but not particularly against it either.

.github/workflows/continuous-integration-workflow.yml
```diff
@@ -115,9 +124,12 @@ jobs:
           cmake --build builddir --parallel 2
           ccache -s
       - name: Tests
+        env:
+          ASAN_OPTIONS: allocator_may_return_null=1
```
dalg24, line 128: What is this option needed for?

masterleinad: We intentionally allocate more memory than reasonably available to check the error message in some tests. It seemed reasonable to not care too much about this particular check.

masterleinad: See [`09f12c6` (#6676)](https://github.com/kokkos/kokkos/pull/6676/commits/09f12c6983657b606489b5a824918a8c02e71c3d).

dalg24: Why did you not use an attribute?
https://clang.llvm.org/docs/AddressSanitizer.html#disabling-instrumentation-with-attribute-no-sanitize-address

masterleinad: I'm not quite sure how exactly I would use it for just this unit test and disabling the unit test seemed good enough.

masterleinad: Looks like the attempt in 335f7f2  for using attributes doesn't really work.

dalg24: I never managed to get it to work when I tried on my Mac a while ago.  Thanks for trying anyway.

dalg24 (approved):

PR #6678, opened 2023-12-17
Provide `Kokkos::swap` utility function as part of Core

dalg24:
Move the `Kokkos_Swap.hpp` header from Algorithms to Core and promote from namespace `Experimental::` to `Kokkos::`.
In the process, made the following changes (to align with the standard library)
* Constexprify
* Fix the noexcept expression
* Instead of failed a static_assert, non-movable types do not participate in overload resolution (we had implemented it as a "mandate" but it was meant to be a "constraint")
* Added an overload for arrays


core/src/Kokkos_Swap.hpp
```diff
@@ -0,0 +1,67 @@
...
+#include <cstddef>
+#include <type_traits>
+#include <utility>
+
+namespace Kokkos {
+
+template <class T>
+KOKKOS_FUNCTION constexpr std::enable_if_t<std::is_move_constructible_v<T> &&
+                                           std::is_move_assignable_v<T>>
+swap(T& a, T& b) noexcept(std::is_nothrow_move_constructible_v<T>&&
```
fnrizzi, line 31: i understand that it would be more appropriate to make it a constraint, but in this case we only have effectively only single overload , so using sfinea makes the error message very cryptic. can we jsut leave it as a static assert ?

core/src/Kokkos_Swap.hpp
```diff
@@ -0,0 +1,67 @@
...
+template <class T>
+struct is_swappable {
+  template <class U>
+  static decltype(swap(std::declval<T&>(), std::declval<T&>())) test_swap(int);
+  struct Nope;
+  template <class U>
+  static Nope test_swap(long);
+  static constexpr bool value =
+      !std::is_same_v<decltype(test_swap<T>(0)), Nope>;
+};
```
fnrizzi, line 49: is this copied from the std more or less?

dalg24: The standard implements `is_swappable` but we can't really use it because it would not add `Kokkos::swap` as one of the possible overloads (unless the arguments belong to namespace Kokkos::).
But yes more or less equivalent to what you'd do in the standard implementation.

core/src/Kokkos_Swap.hpp
dalg24: > Can we jsut leave it as a static assert ?

No we would not be able to detect if the underlying element type is swappable in the overload for arrays.

fnrizzi:
why are we moving it to core?

dalg24:
> why are we moving it to core?

Because that is a fundamental utility function.

core/unit_test/TestSwap.hpp
```diff
@@ -0,0 +1,68 @@
...
+        ++err;
+      }
+    }
+  }
+
+  TestSwap() {
+    int errors;
+    Kokkos::parallel_reduce(
+        "TestSwap", Kokkos::RangePolicy<ExecutionSpace>(0, 1), *this, errors);
+    EXPECT_EQ(errors, 0);
```
fnrizzi, line 62: last comment: shouldn't we also test calling swap from on the host?

core/src/Kokkos_Swap.hpp
```diff
@@ -0,0 +1,67 @@
...
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
+#ifndef KOKKOS_SWAP_HPP
+#define KOKKOS_SWAP_HPP
+
+#include <Kokkos_Macros.hpp>
+
+#include <cstddef>
```
ldh4, line 22: Since `<utility>` is being included, might not need to include `<cstddef>`.

dalg24:
I added it because some builds were failing without it

ldh4 (approved):

core/unit_test/TestSwap.hpp
dalg24: It gets instantiated for all backends so it will run on the host as well.  I didn't see too much benefit from calling it from outside a parallel region.

fnrizzi:
clang format is wrong on jenkins?

dalg24:
> clang format is wrong on jenkins?

It was an issue with the machine.

Retest this please

fnrizzi:
retest this please

fnrizzi (approved):
obviously with the caveat the ci passes

masterleinad:
> clang format is wrong on jenkins?

Also, see #6651.

core/src/Kokkos_Swap.hpp
```diff
@@ -0,0 +1,67 @@
...
+// the U.S. Government retains certain rights in this software.
+//
+// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
+#ifndef KOKKOS_SWAP_HPP
+#define KOKKOS_SWAP_HPP
```
fnrizzi, line 18: i was looking at something, and realized that here you don't have the following lines: 
```
#ifndef KOKKOS_IMPL_PUBLIC_INCLUDE                                                                            |
#define KOKKOS_IMPL_PUBLIC_INCLUDE                                                                            |
#define KOKKOS_IMPL_PUBLIC_INCLUDE_NOTDEFINED_SWAP                                                            |
#endif
```
and at the bottom, like in other headers, is it intentional?

masterleinad: We don't use any private Kokkos headers here so there is no reason to define `KOKKOS_IMPL_PUBLIC_INCLUDE` (and undefining it at the end).

dalg24:
Retest this please

dalg24:
This introduces an ambiguity
```
Building CXX object core/src/CMakeFiles/kokkoscore.dir/Cuda/Kokkos_Cuda_Instance.cpp.o
[197](https://cloud.cees.ornl.gov/jenkins-ci/job/Kokkos/job/PR-6678/4/pipeline-console/?start-byte=0&selected-node=272#log-197)
/usr/include/c++/9/bits/unique_ptr.h(400): error: more than one instance of overloaded function "swap" matches the argument list:
[198](https://cloud.cees.ornl.gov/jenkins-ci/job/Kokkos/job/PR-6678/4/pipeline-console/?start-byte=0&selected-node=272#log-198)
            function template "std::enable_if_t<<expression>, void> Kokkos::swap(T &, T &)"
[199](https://cloud.cees.ornl.gov/jenkins-ci/job/Kokkos/job/PR-6678/4/pipeline-console/?start-byte=0&selected-node=272#log-199)
/var/jenkins/workspace/Kokkos_PR-6678/core/src/Kokkos_Swap.hpp(31): here
[200](https://cloud.cees.ornl.gov/jenkins-ci/job/Kokkos/job/PR-6678/4/pipeline-console/?start-byte=0&selected-node=272#log-200)
            function template "std::enable_if<std::__and_<std::__not_<std::__is_tuple_like<_Tp>>, std::is_move_constructible<_Tp>, std::is_move_assignable<_Tp>>::value, void>::type std::swap(_Tp &, _Tp &)"
[201](https://cloud.cees.ornl.gov/jenkins-ci/job/Kokkos/job/PR-6678/4/pipeline-console/?start-byte=0&selected-node=272#log-201)
/usr/include/c++/9/bits/move.h(182): here
[202](https://cloud.cees.ornl.gov/jenkins-ci/job/Kokkos/job/PR-6678/4/pipeline-console/?start-byte=0&selected-node=272#log-202)
            argument types are: (Kokkos::Cuda *, Kokkos::Cuda *)
[203](https://cloud.cees.ornl.gov/jenkins-ci/job/Kokkos/job/PR-6678/4/pipeline-console/?start-byte=0&selected-node=272#log-203)
          detected during:
[204](https://cloud.cees.ornl.gov/jenkins-ci/job/Kokkos/job/PR-6678/4/pipeline-console/?start-byte=0&selected-node=272#log-204)
            instantiation of "void std::unique_ptr<_Tp, _Dp>::reset(std::unique_ptr<_Tp, _Dp>::pointer) noexcept [with _Tp=Kokkos::Cuda, _Dp=std::default_delete<Kokkos::Cuda>]" 
[205](https://cloud.cees.ornl.gov/jenkins-ci/job/Kokkos/job/PR-6678/4/pipeline-console/?start-byte=0&selected-node=272#log-205)
(307): here
[206](https://cloud.cees.ornl.gov/jenkins-ci/job/Kokkos/job/PR-6678/4/pipeline-console/?start-byte=0&selected-node=272#log-206)
            instantiation of "std::unique_ptr<_Tp, _Dp> &std::unique_ptr<_Tp, _Dp>::operator=(std::unique_ptr<_Tp, _Dp> &&) noexcept [with _Tp=Kokkos::Cuda, _Dp=std::default_delete<Kokkos::Cuda>]" 
[207](https://cloud.cees.ornl.gov/jenkins-ci/job/Kokkos/job/PR-6678/4/pipeline-console/?start-byte=0&selected-node=272#log-207)
/var/jenkins/workspace/Kokkos_PR-6678/core/src/Cuda/Kokkos_CudaSpace.cpp(56): here
[208](https://cloud.cees.ornl.gov/jenkins-ci/job/Kokkos/job/PR-6678/4/pipeline-console/?start-byte=0&selected-node=272#log-208)
[209](https://cloud.cees.ornl.gov/jenkins-ci/job/Kokkos/job/PR-6678/4/pipeline-console/?start-byte=0&selected-node=272#log-209)
1 error detected in the compilation of "/var/jenkins/workspace/Kokkos_PR-6678/core/src/Cuda/Kokkos_CudaSpace.cpp".

dalg24:
I think that we don't really have many option besides not calling it `swap`.
Here is an example illustrating the issue with Kokkos 4.2, swapping value type that is in the same namespace than the user-defined (generic) swap yields ambiguity compile error:
https://godbolt.org/z/qaGq579aq

dalg24:
After talking with Christian, we will attempt a best effort to provide `Kokkos::swap` overload for classes in namespace Kokkos.

crtrott (changes requested):
This doesn't work

core/src/Kokkos_Swap.hpp
romintomasetti, line 31: I am not sure providing a custom `Kokkos::swap` implementation for *any* object type is a good idea.

For example, any host compiler I know supports `std::swap` (and `Kokkos` is on `c++17` so `std::swap` is OK on host code whatsoever).

For device compilers, some are already supporting it. Those I tested:
- `nvcc` (`Build cuda_12.2.r12.2/compiler.33191640_0`)
- `hipcc` (`HIP version: 5.7.31921-d1770ee1b`)

dalg24: As far as I can tell it is not supported by NVCC https://godbolt.org/z/9s3Px96rM

romintomasetti: Sorry, it works on our side for 2 reasons:
1. We are on `c++20`, so `std::swap` is `constexpr`.
2. We compile with the flag `--expt-relaxed-constexpr` when using `nvcc`

So yes, it seems you'll need to bother with re-implementing the wheel once again :wheel_of_dharma:

dalg24:
Closing in favor of #6697

PR #6683, opened 2023-12-20
simd: add compound assignments and reductions

ldh4:
This PR adds following functions that are described in [p1928r7](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/p1928r7.pdf):

- Compound assignments
    - `simd<T, Abi>& operator/=(simd<T, Abi>& lhs, U&& rhs)` (enabled)
    - `simd<T, Abi>& operator>>=(simd<T, Abi>& lhs, U&& rhs)` (added)
    - `simd<T, Abi>& operator<<=(simd<T, Abi>& lhs, U&& rhs)` (added)

- Constructor that accepts a pointer to the src memory
    - `constexpr explicit simd(value_type const* ptr, FlagType flag)`

- Fallback simd reducers
    - `constexpr T reduce(const simd<T, Abi>& x, const typename simd<T, Abi>::mask_type& mask, T identity_element, BinaryOperation binary_op)`
    - `constexpr T reduce(const simd<T, Abi>& x, const typename simd<T, Abi>::mask_type& mask, std::plus<> binary_op = {}) noexcept`
    - `constexpr T reduce(const simd<T, Abi>& x, const typename simd<T, Abi>::mask_type& mask, std::bit_and<> binary_op) noexcept`
    - `constexpr T reduce(const simd<T, Abi>& x, const typename simd<T, Abi>::mask_type& mask, std::bit_or<> binary_op) noexcept`
    - `constexpr T reduce(const simd<T, Abi>& x, const typename simd<T, Abi>::mask_type& mask, std::bit_xor<> binary_op) noexcept`
    - `constexpr T reduce_min(const simd<T, Abi>& x) noexcept`
    - `constexpr T reduce_min(const simd<T, Abi>& x, const typename simd<T, Abi>::mask_type& mask) noexcept`
    - `constexpr T reduce_max(const simd<T, Abi>& x) noexcept`
    - `constexpr T reduce_max(const simd<T, Abi>& x, const typename simd<T, Abi>::mask_type& mask) noexcept`

- `Max`, `Min`, `Sum` reducers for AVX512
    - `T reduce_min(const simd<T, Abi>& x)`
    - `T reduce_max(const simd<T, Abi>& x)`
    - `T reduce(const_where_expression<simd_mask<T, Abi>, simd<T, Abi>> const& x, T, std::plus<>)`

- Deprecated `hmin` and `hmax` in favor of `reduce_min` and `reduce_max`

### DIFF
--- a/simd/src/Kokkos_SIMD_Common_Math.hpp
+++ b/simd/src/Kokkos_SIMD_Common_Math.hpp
@@ -32,6 +32,7 @@ class simd_mask;
 template <class M, class T>
 class const_where_expression;
 
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
 template <typename T, typename Abi>
 [[nodiscard]] KOKKOS_DEPRECATED KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION T
 hmin(const_where_expression<simd_mask<T, Abi>, simd<T, Abi>> const& x) {
@@ -55,6 +56,7 @@ hmax(const_where_expression<simd_mask<T, Abi>, simd<T, Abi>> const& x) {
   }
   return result;
 }
+#endif
 
 template <typename T, typename Abi>
 [[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION T
--- a/simd/src/Kokkos_SIMD_Common_Math.hpp
+++ b/simd/src/Kokkos_SIMD_Common_Math.hpp
@@ -23,6 +23,10 @@ namespace Kokkos {
 
 namespace Experimental {
 
+namespace simd_abi {
+class scalar;
+}
+
 template <class T, class Abi>
 class simd;
 
@@ -58,7 +62,9 @@ hmax(const_where_expression<simd_mask<T, Abi>, simd<T, Abi>> const& x) {
 }
 #endif
 
-template <typename T, typename Abi>
+template <
+    typename T, typename Abi,
+    std::enable_if_t<!std::is_same_v<Abi, simd_abi::scalar>, bool> = false>
 [[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION T
 reduce_min(const_where_expression<simd_mask<T, Abi>, simd<T, Abi>> const& x) {
   auto const& v = x.impl_get_value();
@@ -70,7 +76,9 @@ reduce_min(const_where_expression<simd_mask<T, Abi>, simd<T, Abi>> const& x) {
   return result;
 }
 
-template <class T, class Abi>
+template <
+    class T, class Abi,
+    std::enable_if_t<!std::is_same_v<Abi, simd_abi::scalar>, bool> = false>
 [[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION T
 reduce_max(const_where_expression<simd_mask<T, Abi>, simd<T, Abi>> const& x) {
   auto const& v = x.impl_get_value();
@@ -82,7 +90,9 @@ reduce_max(const_where_expression<simd_mask<T, Abi>, simd<T, Abi>> const& x) {
   return result;
 }
 
-template <class T, class Abi, class BinaryOperation = std::plus<>>
+template <
+    class T, class Abi, class BinaryOperation = std::plus<>,
+    std::enable_if_t<!std::is_same_v<Abi, simd_abi::scalar>, bool> = false>
 [[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION T
 reduce(const_where_expression<simd_mask<T, Abi>, simd<T, Abi>> const& x,
        BinaryOperation op = {}) {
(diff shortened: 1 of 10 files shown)


masterleinad:
The windows build shows
```
[ RUN      ] simd.device_shift_ops
SIMD unit test equality condition failed on device
Traceback functionality not available
```

simd/unit_tests/include/TestSIMD_ShiftOps.hpp
```diff
@@ -245,13 +256,23 @@ KOKKOS_INLINE_FUNCTION void device_check_shift_ops() {
...
       if constexpr (std::is_signed_v<DataType>) {
         for (std::size_t i = 0; i < width; ++i) test_vals[i] *= -1;
         device_check_shift_op_all_loaders<Abi>(shift_right(), test_vals,
                                                shift_by, num_cases);
+        device_check_shift_op_all_loaders<Abi>(shift_right_eq(), test_vals,
+                                               shift_by, num_cases);
+        device_check_shift_op_all_loaders<Abi>(shift_left(), test_vals,
+                                               shift_by, num_cases);
+        device_check_shift_op_all_loaders<Abi>(shift_left_eq(), test_vals,
+                                               shift_by, num_cases);
```
masterleinad, line 275: Left shifting negative values is undefined behavior (as the CI detects as well).
```suggestion
```

masterleinad:
```
/var/jenkins/workspace/Kokkos_PR-6683/simd/src/Kokkos_SIMD_Common.hpp(220): error: "Kokkos::Experimental::simd<double, Kokkos::Experimental::simd_abi::avx512_fixed_size<8>> &" contains a vector, which is not supported in device code
```
```
/var/jenkins/workspace/Kokkos_PR-6683/simd/src/Kokkos_SIMD_Common.hpp:350:10: error: reference to __host__ function 'reduce<int, Kokkos::Experimental::simd_abi::scalar, std::multiplies<>>' in __host__ __device__ function
  return reduce(v, binary_op);
```

simd/unit_tests/include/TestSIMD_ShiftOps.hpp
nliber: Right shifting negative values is also UB: [[expr.shift]](https://eel.is/c++draft/expr.shift#1)

nliber (changes requested):

simd/unit_tests/include/TestSIMD_ShiftOps.hpp
masterleinad: cppreference says in https://en.cppreference.com/w/cpp/language/operator_arithmetic:
> For negative a, the value of a >> b is implementation-defined (in most implementations, this performs arithmetic right shift, so that the result remains negative). 

Doesn't [[expr.shift]](https://eel.is/c++draft/expr.shift#1) only talk about the right operand?

nliber: You are correct; it only talks about the right operand.

FYI: somewhere between C++17 and the current WD, right shifting negative numbers became well-defined (It shouldn't matter for this PR though).

ldh4: Removed tests for left shifting signed values.
For right shifting, right operands are set to be 0 or positive values in this test, so shouldn't be a problem.

ldh4:
Retest this please.

simd/src/Kokkos_SIMD_Common_Math.hpp
```diff
@@ -55,20 +60,60 @@ hmax(const_where_expression<simd_mask<T, Abi>, simd<T, Abi>> const& x) {
...
+
+template <
+    class T, class Abi, class BinaryOperation = std::plus<>,
+    std::enable_if_t<!std::is_same_v<Abi, simd_abi::scalar>, bool> = false>
+[[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION T
+reduce(const_where_expression<simd_mask<T, Abi>, simd<T, Abi>> const& x,
+       BinaryOperation op = {}) {
+  auto const& v = x.impl_get_value();
+  auto const& m = x.impl_get_mask();
+  auto result   = v[0];
```
cwpearson, line 101: If I understand page 43/44 of the linked PDF, with an empty mask this is supposed to return the identity, not `v[0]`? If so, test will need to be modified as well I think.

simd/unit_tests/include/SIMDTesting_Ops.hpp
```diff
@@ -259,116 +331,149 @@ class log_op {
...
+    auto w = Kokkos::Experimental::where(mask, a);
+    return Kokkos::Experimental::reduce(w, BinaryOperation());
+  }
+  template <typename T, typename MaskType>
+  KOKKOS_INLINE_FUNCTION auto on_device_serial(T const& a,
+                                               MaskType mask) const {
+    auto w        = Kokkos::Experimental::where(mask, a);
+    auto const& v = w.impl_get_value();
+    auto const& m = w.impl_get_mask();
+    auto result   = v[0];
```
cwpearson, line 365: See comment about returning identity vs v[0]. Other places in tests as well.

cwpearson (changes requested):

simd/src/Kokkos_SIMD_Common_Math.hpp
ldh4: You are correct. Fixed to return the identity element instead if no bits are on in input mask.

simd/unit_tests/include/SIMDTesting_Ops.hpp
ldh4: Fixed this to return identity element if no bits are on in the mask.
This was replaced and moved to a reduction test that takes in a simd value and a mask instead of `const_where_expression`.

nliber (approved):

cwpearson (approved):

ldh4:
Confirmed that changes in this PR for NEON backend build and pass all tests.

Issue #6670, opened 2023-12-14
Nightly failures with openmp.partitioning_by_args , partitioning_by_vector subtests


ndellingwood:
**Describe the bug**
The openmp `partitioning_by_args` and `partitioning_by_vector` subtests fail in Cuda_OpenMP builds with cuda/11.2.2 and gcc/8.3.0 following recent update; presumably this will apply to other OpenMP jobs after nightly testing completes

**Test failures**:
```
[ RUN      ] openmp.partitioning_by_args
/home/jenkins/weaver/workspace/Kokkos_develop_weaver_cuda_openmp_cuda1122_gcc830_volta-debug/kokkos/core/unit_test/TestExecSpacePartitioning.hpp:105: Failure
Expected equality of these values:
  sum1
    Which is: -294063648
  sum2
    Which is: 2016
[  FAILED  ] openmp.partitioning_by_args (20 ms)
[ RUN      ] openmp.partitioning_by_vector
/home/jenkins/weaver/workspace/Kokkos_develop_weaver_cuda_openmp_cuda1122_gcc830_volta-debug/kokkos/core/unit_test/TestExecSpacePartitioning.hpp:105: Failure
Expected equality of these values:
  sum1
    Which is: -1367109152
  sum2
    Which is: 3297
[  FAILED  ] openmp.partitioning_by_vector (11 ms)
```

This follows merge of the following commits:

Fix builtin_unreachable use for MSVC/CUDA ([details](https://jenkins-son.sandia.gov/job/Kokkos_develop_weaver_cuda_openmp_cuda1122_gcc830_volta-debug/334/changes#fb0380b91b79b5744ea2b12e28a5663c59aeacc4) / [githubweb](https://github.com/kokkos/kokkos/commit/fb0380b91b79b5744ea2b12e28a5663c59aeacc4))
    Fix missing include on msvc/cuda ([details](https://jenkins-son.sandia.gov/job/Kokkos_develop_weaver_cuda_openmp_cuda1122_gcc830_volta-debug/334/changes#458910fbf37ba7fae78e54a2e5b78522cb1b2ccb) / [githubweb](https://github.com/kokkos/kokkos/commit/458910fbf37ba7fae78e54a2e5b78522cb1b2ccb))
[...]

PR #6666, opened 2023-12-13
Use omp_get_max_active_levels() when supported

masterleinad:
Similar to https://github.com/kokkos/kokkos/pull/2843 (and fixed in https://github.com/kokkos/kokkos/pull/3142), `omp_get_max_active_levels` exists OpenMP 4.5, see [openmp.org/wp-content/uploads/openmp-4.5.pdf](https://www.openmp.org/wp-content/uploads/openmp-4.5.pdf) page 248, and `omp_get_nested` has been deprecated since.
Extracted from #6151.

PR #6685, opened 2023-12-21
OpenMP: Use `omp_get_nested` for older gcc versions

rgayatri23:
`omp_get_max_active_level` fails with gcc versions lower than 11.2.0. 
So reverting to the prior implementation which was changed here #6666 

This fixes the issue seen in #6670

### DIFF
--- a/core/src/OpenMP/Kokkos_OpenMP_Instance.hpp
+++ b/core/src/OpenMP/Kokkos_OpenMP_Instance.hpp
@@ -45,13 +45,13 @@ namespace Kokkos {
 namespace Impl {
 
 inline bool execute_in_serial(OpenMP const& space = OpenMP()) {
-// FIXME_OPENMP - `omp_get_max_active_levels` fails with gcc version lower
-// than 11.1.0
+// The default value returned by `omp_get_max_active_levels` with gcc version lower
+// than 11.1.0 is 2147483647 instead of 1.
 #if (!defined(KOKKOS_COMPILER_GNU) || KOKKOS_COMPILER_GNU >= 1110) && \
     _OPENMP >= 201511
-  int is_nested = omp_get_max_active_levels() > 1;
+  bool is_nested = omp_get_max_active_levels() > 1;
 #else
-  int is_nested = omp_get_nested();
+  bool is_nested = static_cast<bool>(omp_get_nested());
 #endif
   return (OpenMP::in_parallel(space) && !(is_nested && (omp_get_level() == 1)));
 }
--- a/core/src/OpenMP/Kokkos_OpenMP_Instance.hpp
+++ b/core/src/OpenMP/Kokkos_OpenMP_Instance.hpp
@@ -45,6 +45,9 @@ namespace Kokkos {
 namespace Impl {
 
 inline bool execute_in_serial(OpenMP const& space = OpenMP()) {
+  // FIXME_OPENMP - `omp_get_max_active_levels` fails with gcc version lower
+  // than 11.2.0
+#if KOKKOS_COMPILER_GNU >= 1120
   return (OpenMP::in_parallel(space) && !(
 #if _OPENMP >= 201511
                                             (omp_get_max_active_levels() > 1)
@@ -52,6 +55,10 @@ inline bool execute_in_serial(OpenMP const& space = OpenMP()) {
                                             omp_get_nested()
 #endif
                                             && (omp_get_level() == 1)));
+#else
+  return (OpenMP::in_parallel(space) &&
+          !(omp_get_nested() && (omp_get_level() == 1)));
+#endif
 }
 
 }  // namespace Impl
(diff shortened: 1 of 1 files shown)


core/src/OpenMP/Kokkos_OpenMP_Instance.hpp
```diff
@@ -45,13 +45,20 @@ namespace Kokkos {
...
 #if _OPENMP >= 201511
                                             (omp_get_max_active_levels() > 1)
 #else
                                             omp_get_nested()
 #endif
                                             && (omp_get_level() == 1)));
+#else
+  return (OpenMP::in_parallel(space) &&
+          !(omp_get_nested() && (omp_get_level() == 1)));
+#endif
```
masterleinad, line 61: Maybe
```suggestion
  // FIXME_OPENMP - `omp_get_max_active_levels` fails with gcc version lower
  // than 11.2.0
#if KOKKOS_COMPILER_GNU >= 1120 && _OPENMP >= 201511
  int is_nested = omp_get_max_active_levels() > 1;
#else
  int is_nested = omp_get_nested();
 #endif
  return (OpenMP::in_parallel(space) &&
          !(is_nested && (omp_get_level() == 1)));
```
to keep it a little more compact?

dalg24:
How did you determine what versions of GCC were broken?

rgayatri23:
> How did you determine what versions of GCC were broken?

I tested it with 10.3.0 and 11.0 and 11.2.0. The current `develop` branch only passes with the last version.

dalg24:
> > How did you determine what versions of GCC were broken?
> 
> I tested it with 10.3.0 and 11.0 and 11.2.0. The current `develop` branch only passes with the last version.

Is that something you would be able to reproduce on Compiler Explorer?

rgayatri23:
I tested this on godbolt (https://godbolt.org/z/zK6d971Gn) and it looks like anything less than gcc/11.1 gives garbage value.
Everything up from gcc/11.1 and higher gives the expected result. 
I updated the code to use newer version from gcc/11.1 and higher.

core/src/OpenMP/Kokkos_OpenMP_Instance.hpp
rgayatri23: Done. Thanks for the suggestion.

core/src/OpenMP/Kokkos_OpenMP_Instance.hpp
```diff
@@ -45,13 +45,15 @@ namespace Kokkos {
 namespace Impl {
 
 inline bool execute_in_serial(OpenMP const& space = OpenMP()) {
-  return (OpenMP::in_parallel(space) && !(
-#if _OPENMP >= 201511
-                                            (omp_get_max_active_levels() > 1)
+// FIXME_OPENMP - `omp_get_max_active_levels` fails with gcc version lower
+// than 11.1.0
```
masterleinad, line 49:
```suggestion
// The default value returned by `omp_get_max_active_levels` with gcc version lower
// than 11.1.0 is 2147483647 instead of 1.
```

masterleinad:
It seems that only the default value is problematic. As soon as `omp_set_max_active_levels(int)` is used `omp_get_max_active_levels()` also returns the expected values for GCC versions before 11.

core/src/OpenMP/Kokkos_OpenMP_Instance.hpp
```diff
@@ -45,13 +45,15 @@ namespace Kokkos {
...
+#if (!defined(KOKKOS_COMPILER_GNU) || KOKKOS_COMPILER_GNU >= 1110) && \
+    _OPENMP >= 201511
+  int is_nested = omp_get_max_active_levels() > 1;
```
dalg24, line 52:
```suggestion
  bool is_nested = omp_get_max_active_levels() > 1;
```

core/src/OpenMP/Kokkos_OpenMP_Instance.hpp
```diff
@@ -45,13 +45,15 @@ namespace Kokkos {
...
 #else
-                                            omp_get_nested()
+  int is_nested = omp_get_nested();
```
dalg24, line 54:
```suggestion
  bool is_nested = static_cast<bool>(omp_get_nested());
```

masterleinad (approved):

core/src/OpenMP/Kokkos_OpenMP_Instance.hpp
```diff
@@ -45,13 +45,15 @@ namespace Kokkos {
 namespace Impl {
 
 inline bool execute_in_serial(OpenMP const& space = OpenMP()) {
-  return (OpenMP::in_parallel(space) && !(
-#if _OPENMP >= 201511
-                                            (omp_get_max_active_levels() > 1)
+// The default value returned by `omp_get_max_active_levels` with gcc version lower
+// than 11.1.0 is 2147483647 instead of 1.
```
dalg24, line 49:
```suggestion
// The default value returned by `omp_get_max_active_levels` with gcc version
// lower than 11.1.0 is 2147483647 instead of 1.
```

PR #6689, opened 2024-01-02
Drop unnecessary if-conditions in `Profiling::ProfilingSection`

dalg24:
The `invoke_kokkosp_callback` internal handles the guarding for whether a tool library is loaded or not, and more importantly whether the tool actually handles sections at all.
https://github.com/kokkos/kokkos/blob/cbbe09b93db49e081ab4d400975aca8c007c4223/core/src/impl/Kokkos_Profiling.cpp#L477-L500

I don't think that the `if (Kokkos::Profiling::profileLibraryLoaded())` is needed but if the reviewers disagree then we probably should add it to `Profiling::ScopedRegion`.
https://github.com/kokkos/kokkos/blob/cbbe09b93db49e081ab4d400975aca8c007c4223/core/src/Kokkos_Profiling_ScopedRegion.hpp#L39-L42

Drive-by change: removed pointless header include

### DIFF
--- a/core/src/Kokkos_Profiling_ProfileSection.hpp
+++ b/core/src/Kokkos_Profiling_ProfileSection.hpp
@@ -22,7 +22,6 @@
 #endif
 
 #include <Kokkos_Macros.hpp>
-#include <impl/Kokkos_Profiling_Interface.hpp>
 #include <impl/Kokkos_Profiling.hpp>
 
 #include <string>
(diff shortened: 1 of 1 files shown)


dalg24:
My rational for dropping the guards is that we don't expect users to do
```C++
  Kokkos::Profiling::pushRegion("MyApp::do_stuff");
}
// do work
  Kokkos::Profiling::popRegion();
}
```
and I don't see why sections would be any different.

masterleinad (approved):
Fine with me.

dalg24:
Failures are unrelated (Disk quota exceeded on one of the test machine)

PR #6692, opened 2024-01-03
Add `ATOMICS_BYPASS` configuration option to disable atomics

dalg24:
Do not disable atomics unless the user opts in.

We are currently disabling atomics when no host parallel nor device backend is enabled (i.e. only the Serial backend is enabled)
https://github.com/kokkos/kokkos/blob/71a9bcae52543bd065522bf3e41b5bfa467d8015/core/src/Kokkos_Macros.hpp#L87-L92
https://github.com/kokkos/kokkos/blob/71a9bcae52543bd065522bf3e41b5bfa467d8015/core/src/Kokkos_Atomics_Desul_Wrapper.hpp#L52-L56
This gets problematic with our recent enabling multiple Serial execution instance and advertising they can be used from threads and we already had thread sanitizers yell at us (reference counting).
Here, just like in #6672, I suggest to change the current behavior to do the safe thing by default but I introduce an option to get back the old behavior and disable atomics.  I made sure atomics can only be disabled for Serial-only build.

I am not sure about the `NAME_TBD_UNSAFE_ATOMICS` option name and would like to hear suggestions.  I do not like that it is sort of a "negated" option (because it disables something when it is ON) but I judged it was more desirable to have a macro defined when we depart from the default behavior and none otherwise.


dalg24:
Forgot to refer to #6674 
The approach in that PR was to make our internal reference counting thread safe by not using the Kokkos atomics.

masterleinad:
> Forgot to refer to #6674
> The approach in that PR was to make our internal reference counting thread safe by not using the Kokkos atomics.

Yes, these are places where we always need to use proper atomics (also see #6151) which I think we should handle explicitly independent of the configuration option proposed here.

masterleinad:
> I am not sure about the `NAME_TBD_UNSAFE_ATOMICS` option name and would like to hear suggestions. I do not like that it is sort of a "negated" option (because it disables something when it is ON) but I judged it was more desirable to have a macro defined when we depart from the default behavior and none otherwise.

What about `Kokkos_ENABLE_SERIAL_SAFE_ATOMICS` (default `ON`) since it only really matters when (solely) the `Serial` backend is enabled?

crtrott:
This approach looks good to me. How about: `Kokkos_ENABLE_ATOMICS_BYPASS`

dalg24:
I updated the PR with the name suggestion from Christian. Feel free to suggest a better name.

Also wanted to mention I did not forget generated makefiles, I chose not to update. Not trying to sweep it under the rug, let me know if you disagree.

cmake/kokkos_arch.cmake
```diff
@@ -1139,3 +1139,14 @@ MESSAGE(STATUS "Architectures:")
 FOREACH(Arch ${KOKKOS_ENABLED_ARCH_LIST})
   MESSAGE(STATUS " ${Arch}")
 ENDFOREACH()
+
+
+IF(KOKKOS_ENABLE_ATOMICS_BYPASS)
+  IF(NOT _HOST_PARALLEL STREQUAL "NoTypeDefined" OR NOT _DEVICE_PARALLEL STREQUAL "NoTypeDefined")
+    MESSAGE(FATAL_ERROR "Not allowed to disable atomics if any host parallel or device backend is enabled")
```
masterleinad, line 1146:
```suggestion
    MESSAGE(FATAL_ERROR "Kokkos_ENABLE_ATOMICS_BYPASS can only be enabled if neither a host parallel nor a device backend is enabled!")
```

cmake/kokkos_enable_options.cmake
```diff
@@ -73,6 +73,7 @@ KOKKOS_ENABLE_OPTION(HIP_MULTIPLE_KERNEL_INSTANTIATIONS OFF "Whether multiple ke
 
 # This option will go away eventually, but allows fallback to old implementation when needed.
 KOKKOS_ENABLE_OPTION(DESUL_ATOMICS_EXTERNAL OFF "Whether to use an external desul installation")
+KOKKOS_ENABLE_OPTION(ATOMICS_BYPASS OFF "**NOT RECOMMENDED** Whether to disable atomics")
```
masterleinad, line 76: I find it hard to understand what "disabling atomics" is supposed to mean. Maybe
```suggestion
KOKKOS_ENABLE_OPTION(ATOMICS_BYPASS OFF "**NOT RECOMMENDED** Whether to make atomics operations no-ops (only allowed for Serial-only builds).")
```

dalg24: They are not no-ops, they do perform the operation, just not atomically.   That said I agree "disable atomics" is not great.

masterleinad: Yes, maybe we just go with "bypass" matching the options' name.

dalg24: I find repeating the exact same word as the option not useful.  Because either you understand what "atomics bypass" mean and you don't care; or you are confused and you read the description and great that is not helpful.

core/src/Kokkos_Macros.hpp
```diff
@@ -84,11 +84,18 @@
...
     !defined(KOKKOS_ENABLE_OPENMPTARGET) && !defined(KOKKOS_ENABLE_HIP) && \
     !defined(KOKKOS_ENABLE_SYCL) && !defined(KOKKOS_ENABLE_OPENACC)
-#define KOKKOS_INTERNAL_NOT_PARALLEL
+// OK no other backend than the Serial one is enabled
+// atomics will be disabled
+#else
+#error Not allowed to disable atomics with a host parallel and/or ar device backend enabled
+#endif
+
 #endif
```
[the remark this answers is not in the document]
crtrott: Do we need KOKKOS_INTERNAL_NOT_PARALLEL in your suggestion its removed isn't it?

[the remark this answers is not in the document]
masterleinad: The diff shows the old version first, not the one in the current status of the pull request. AFAICT, all occurrences of `KOKKOS_INTERNAL_NOT_PARALLEL` have already been replaced.

cmake/kokkos_arch.cmake
dalg24: From Christian "Make atomics non-atomic for non-threaded MPI-only use cases"

crtrott (approved):
I can make a follow on after talking with Stan in case we need this for the Makefile.

core/src/Serial/Kokkos_Serial.cpp
```diff
@@ -153,7 +153,7 @@ void Serial::print_configuration(std::ostream& os, bool /*verbose*/) const {
   os << "Host Serial Execution Space:\n";
   os << "  KOKKOS_ENABLE_SERIAL: yes\n";
 
-#ifdef KOKKOS_INTERNAL_NOT_PARALLEL
+#ifdef KOKKOS_ENABLE_ATOMICS_BYPASS
   os << "Kokkos atomics disabled\n";
```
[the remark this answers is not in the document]
dalg24: Not planning on updating the message.  I think it is good enough.

cmake/kokkos_enable_options.cmake
dalg24: Updated with Christian's suggestion

cmake/kokkos_arch.cmake
dalg24: Updated.  Let me know if you still don't like it.

core/src/Kokkos_Macros.hpp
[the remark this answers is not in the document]
masterleinad: @dalg24 What about this one?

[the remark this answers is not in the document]
dalg24: Updated.  Let me know if you still don't like it.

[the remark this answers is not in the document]
masterleinad: I still find this more complicated than necessary with the two `#if`s.

masterleinad (approved):

dalg24:
Retest this please

dalg24:
Retest this please

dalg24:
Everything passed except the SYCL build that was aborted.

Issue #6694, opened 2024-01-03
Check whether reduction_identity is used in team reduce


crtrott:
Its not quite clear to me why the point_t default constructor is changed, considering that there is a `reduction_identity` overload. Maybe the team reduce does not use reduction identity??

_Originally posted by @crtrott in https://github.com/kokkos/kokkos/issues/5334#issuecomment-1875789682_

masterleinad:
Our reduction interface only uses `reduction_identity` for the built-in reducers. If we don't provide a reducer (and the functor doesn't implement `join`), the reduction uses `operator+=` for `join` and uses the default constructor for `init`. Hence, the behavior seems to be inline with this behavior.
We could still discuss if we would wan to use `Kokkos::Sum`(using `reduction_identity`) if no reducer was provided.

dalg24:
It is used as `test_sum` does both reductions w/o a reducer and with `Kokkos::Sum`.
https://github.com/kokkos/kokkos/blob/f485cfa53cb2d25cab98e7a1d41a116027ec2363/core/unit_test/TestReducers.hpp#L403-L498

Thanks Daniel for the explanation.

PR #6696, opened 2024-01-03
Cleanup: remove unused `reduction_identity<Test::point_t>` specialization

dalg24:
Address #6694
We thought the `reduction_identity` specialization for `point_t` looked fishy (e.g. the return type of the member functions is not `point_t`) when reviewing #5334 and were trying to understand why the `point_t` default constructor needed changes.  The answer was "it is never used".

Daniel recalled how custom reductions work in https://github.com/kokkos/kokkos/issues/6694#issuecomment-1875876600

For the record, `point_t` and the unnecessary specialization of `reduction_identity` were introduced in  #4156

### DIFF
--- a/core/unit_test/TestNonTrivialScalarTypes.hpp
+++ b/core/unit_test/TestNonTrivialScalarTypes.hpp
@@ -270,21 +270,5 @@ struct reduction_identity<Test::array_reduce<scalar_t, N>> {
     return Test::array_reduce<scalar_t, N>(t_red_ident::prod());
   }
 };
-
-template <>
-struct reduction_identity<Test::point_t> {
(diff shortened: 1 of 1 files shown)


dalg24:
Apparently something does not quite go like we thought
```
/__w/kokkos/kokkos/core/src/Kokkos_Parallel_Reduce.hpp:61:11: error: implicit instantiation of undefined template 'Kokkos::reduction_identity<Test::point_t>'
   61 |     val = reduction_identity<value_type>::sum();

dalg24:
https://github.com/kokkos/kokkos/blob/f485cfa53cb2d25cab98e7a1d41a116027ec2363/core/unit_test/TestReducers.hpp#L480-L497
We also do a reduction with the Sum reducer

PR #6697, opened 2024-01-04
Provide `kokkos_swap` as part of Core and deprecate `Experimental::swap` in Algorithms

dalg24:
Alternative to #6678 

`Kokkos::Experimental::swap -> Kokkos::kokkos_swap`

I had overlooked that removing `Kokkos::Experimental::swap` was a breaking change so I reintroduced it but marked it as deprecated.  I moved the definition to `<std_algorithms/Kokkos_IterSwap.hpp>` because I wanted it to remain in Algorithms (moving it to Core would not be a breaking change if you prefer to keep it in `<Kokkos_Swap.hpp>`).  Let me know if you want me to change this.


dalg24:
Retest this please

dalg24:
Retest this please

algorithms/src/std_algorithms/Kokkos_IterSwap.hpp
```diff
@@ -58,6 +57,16 @@ void iter_swap(IteratorType1 a, IteratorType2 b) {
...
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+template <class T>
+KOKKOS_DEPRECATED_WITH_COMMENT("Use Kokkos::kokkos_swap instead!")
+KOKKOS_FUNCTION
+    void swap(T& a, T& b) noexcept(::Kokkos::kokkos_swap(std::declval<T&>(),
+                                                         std::declval<T&>())) {
+  ::Kokkos::kokkos_swap(a, b);
+}
+#endif
+
```
ldh4, line 69: I'm guessing this in moved here just for the sake of deprecation warning and to delete `std_algorithms/Kokkos_Swap.hpp` right away?

core/src/Kokkos_Swap.hpp
```diff
@@ -0,0 +1,68 @@
...
+};
+
+template <class T>
+inline constexpr bool is_nothrow_swappable_v =
+    noexcept(kokkos_swap(std::declval<T&>(), std::declval<T&>()));
+
+}  // namespace Impl
+
+template <class T, std::size_t N>
+KOKKOS_FUNCTION constexpr std::enable_if_t<Impl::is_swappable<T>::value>
```
ldh4, line 59: Should there be `is_swappable_v` as well?

dalg24: Not worth it.  I don't see a good reason to make it public and this is the only place where we would use it.

algorithms/src/std_algorithms/Kokkos_IterSwap.hpp
dalg24: It is in order to by more time to anyone who would be using `Kokkos::Experimental::swap`

dalg24:
The SYCL CI build was aborted.  Everything else passed.  I plan to ignore.

crtrott (approved):

ldh4 (approved):

PR #6701, opened 2024-01-04
Try linking against CUDA libararies even with KOKKOS_ENABLE_COMPILE_AS_CMAKE_LANGUAGE

masterleinad:
The MSVC+Cuda CI wasn't tested with #6392 and is failing since we are not linking against the Cuda driver libraries. This pull request tries to just use our regular TPL interface for that.


dalg24 (approved):

dalg24:
@ibaned please review

ibaned:
@dalg24 I will try to build with this soon!

ibaned (approved):
Our application still compiles with these changes!

PR #6711, opened 2024-01-12
Drop pointless `Impl::CudaExec` forward declaration that has no matching definition

dalg24:
Clearly an oversight


fnrizzi (approved):

fnrizzi:
i think github ci did not run because of the change in the targt branch but the jenkins ran

Rombur (approved):

dalg24:
Ignoring SYCL build that was aborted

PR #6710, opened 2024-01-12
Deprecate `{Cuda,HIP}::detect_device_count()` and `Cuda::[detect_]device_arch()`

dalg24:
* Deprecate
  * `Cuda::detect_device_count()`
  * `Cuda::device_arch()`
  * `Cuda::detect_device_arch()`
  * `HIP::detect_device_count()`
* Inline getting the device count with CUDA and HIP (cannot be generic and indirection brings no benefit)
* Get rid of `CudaInternalDevices`
* Cleanup `Cuda::print_configuration`
* Get rid of a couple CUDA wrapper functions where setting the device made no sense   
  * `CudaIntenal::cuda_get_device_count_wrapper` 
  * `CudaIntenal::cuda_get_device_properties_wrapper `

PR #6713, opened 2024-01-12
Add runtime function to query the number of devices and make device ID consistent with `KOKKOS_VISIBLE_DEVICES`

dalg24:
`Kokkos::num_devices()` is intended as a replacement for `{Cuda,HIP}::detect_device_count()` deprecated in #6710 
I propose to make it legal to call it before `Kokkos::initialize` so it may be leveraged to select what GPU to use.

This PR also fixes a defect in `Kokkos::device_id()` when the `KOKKOS_VISIBLE_DEVICES` environment variable is set.
The Kokkos runtime was returning the device id according to the underlying backend runtime but not taking into consideration whether the devices were masked out.  This was an oversight when we introduced it in #5855 (first released in 4.1)
Assume the current system has 4 GPUs, the table below shows the value returned by the `device_id()` runtime
initialization settings | current | this PR
--- | --- | ---
\<none\> | 0 | 0
`device_id=1` | 1 | 1
`KOKKOS_VISIBLE_DEVICES=0` | 0 | 0
`KOKKOS_VISIBLE_DEVICES=3` | 3 | 0
`KOKKOS_VISIBLE_DEVICES=1,0` | 1 | 0
`device_id=1` `KOKKOS_VISIBLE_DEVICES=1,0` | 0 | 1

I also went ahead and masked out devices in `{Cuda,HIP}::print_config`

### DIFF
--- a/core/src/impl/Kokkos_Core.cpp
+++ b/core/src/impl/Kokkos_Core.cpp
@@ -198,7 +198,12 @@ std::vector<int> const& Kokkos::Impl::get_visible_devices() {
 }
 
 [[nodiscard]] int Kokkos::num_devices() noexcept {
-  return Impl::get_visible_devices().size();
+  if constexpr (std::is_same_v<DefaultExecutionSpace,
+                               DefaultHostExecutionSpace>) {
+    return -1;  // no GPU backend enabled
+  } else {
+    return Impl::get_visible_devices().size();
+  }
 }
 
 [[nodiscard]] int Kokkos::num_threads() noexcept {
--- a/core/src/impl/Kokkos_Core.cpp
+++ b/core/src/impl/Kokkos_Core.cpp
@@ -167,6 +167,11 @@ bool is_valid_map_device_id_by(std::string const& x) {
 
 }  // namespace
 
+std::vector<int> const& Kokkos::Impl::get_visible_devices() {
+  static auto devices = get_visible_devices(get_device_count());
+  return devices;
+}
+
 [[nodiscard]] int Kokkos::device_id() noexcept {
 #if defined(KOKKOS_ENABLE_CUDA)
   int device = Cuda().cuda_device();
@@ -190,11 +195,6 @@ bool is_valid_map_device_id_by(std::string const& x) {
   }
   Kokkos::abort("implementation bug");
   return -1;
-}
-
-std::vector<int> const& Kokkos::Impl::get_visible_devices() {
-  static auto devices = get_visible_devices(get_device_count());
-  return devices;
 }
 
 [[nodiscard]] int Kokkos::num_threads() noexcept {
--- a/core/src/impl/Kokkos_Core.cpp
+++ b/core/src/impl/Kokkos_Core.cpp
@@ -193,7 +193,7 @@ std::vector<int> const& Kokkos::Impl::get_visible_devices() {
       return i;
     }
   }
-  Kokkos::abort("implementation bug");
+  Kokkos::abort("Unexpected error: cannot determine device id");
   return -1;
 }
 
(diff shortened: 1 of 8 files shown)


Rombur (approved):
Looks good. How did you test that the mask for the visible devices works correctly? We don't have a machine in the CI with multiple visible GPUS.

dalg24:
> How did you test that the mask for the visible devices works correctly? We don't have a machine in the CI with multiple visible GPUS.

We have a test for the infamous `Impl::get_visible_devices(int device_count) -> std::vector<int>`
https://github.com/kokkos/kokkos/blob/d560c4719df6fcf664cdb5c98de29fdd159f27d0/core/unit_test/TestParseCmdLineArgsAndEnvVars.cpp#L392-L439
and I had check by hand on a system back when we introduced `KOKKOS_VISIBLE_DEVICES`.
That is not bulletproof obviously but it is something at least.
Do you have suggestions on how to improve?

core/unit_test/CMakeLists.txt
```diff
@@ -1235,12 +1235,10 @@ if (NOT KOKKOS_HAS_TRILINOS)
       INPUT TestDeviceAndThreads.py
       ${USE_SOURCE_PERMISSIONS_WHEN_SUPPORTED}
     )
-    if(NOT Kokkos_ENABLE_OPENMPTARGET)  # FIXME_OPENMPTARGET does not select the right device
```
dalg24, line 1238: Oversight in #5492

Rombur:
> Do you have suggestions on how to improve?

No I was just wondering how you tested it. Maybe it's something Sandia could test in their nightly

core/unit_test/CMakeLists.txt
dalg24: Looks like OpenMPTarget still doesn't do the right thing.
@rgayatri23 please fix.
Withdrawing the commit in the meantime.

core/src/impl/Kokkos_Core.cpp
```diff
@@ -167,20 +167,43 @@ bool is_valid_map_device_id_by(std::string const& x) {
...
+  int device = -1;
+  return device;
 #endif
+  auto const& visible_devices = Impl::get_visible_devices();
+  for (std::size_t i = 0; i < visible_devices.size(); ++i) {
+    if (visible_devices[i] == device) {
+      return i;
+    }
+  }
+  Kokkos::abort("implementation bug");
```
ldh4, line 196: I suggest making this abort message more instructive. I'm not sure how likely this abort can be triggered, but since this is a publicly callable function, I think it should be more informative. (something like `Unexpected error; device backend is enabled but no visible device found`?)

dalg24: this is a safeguard that would only occur if we (kokkos) screw up the logic.

dalg24: but yes "Unexpected error" sounds good.  essentially it could happen if a backend messes up

ldh4 (approved):

core/src/impl/Kokkos_Core.cpp
```diff
@@ -167,20 +167,43 @@ bool is_valid_map_device_id_by(std::string const& x) {
 
 }  // namespace
 
+std::vector<int> const& Kokkos::Impl::get_visible_devices() {
+  static auto devices = get_visible_devices(get_device_count());
+  return devices;
+}
```
masterleinad, line 173: Any objections to specialize for `SYCL` (in a folow-up) to store sycl::devices in that case? I am just now looking at implementing https://kokkosteam.slack.com/archives/C5BGU5NDQ/p1705683106903069 for filtering suitable `SYCL` devices.

masterleinad (approved):

core/src/impl/Kokkos_Core.cpp
dalg24: How do you want to specialize?
Also I did not figure how to you use your slack link.

masterleinad: See https://github.com/kokkos/kokkos/pull/6758.

dalg24: Nevermind you do `get_sycl_devices`

dalg24:
Ignoring the CI that is headed straight to timeout following the massive retrigger of testing all PRs.

PR #6721, opened 2024-01-16
Let the smart pointer manage the CUDA/HIP stream

dalg24:
Instead of storing a boolean in `{Cuda,HIP}Internal` to determine whether the stream needs to be deleted by the destructor, let the `HostSharedPtr` that manages it also deal with that stream.

Rational for doing this: simplifying logic and also preparing for getting rid of `{Cuda,HIP}Internal::{initialize,finalize}()` altogether.

### DIFF
--- a/core/src/HIP/Kokkos_HIP.cpp
+++ b/core/src/HIP/Kokkos_HIP.cpp
@@ -115,13 +115,14 @@ HIP::HIP()
 }
 
 HIP::HIP(hipStream_t const stream, Impl::ManageStream manage_stream)
-    : m_space_instance(new Impl::HIPInternal, [](Impl::HIPInternal* ptr) {
-        ptr->finalize();
-        if (static_cast<bool>(manage_stream)) {
-          KOKKOS_IMPL_HIP_SAFE_CALL(cudaStreamDestroy(ptr->m_stream));
-        }
-        delete ptr;
-      }) {
+    : m_space_instance(
+          new Impl::HIPInternal, [manage_stream](Impl::HIPInternal* ptr) {
+            ptr->finalize();
+            if (static_cast<bool>(manage_stream)) {
+              KOKKOS_IMPL_HIP_SAFE_CALL(hipStreamDestroy(ptr->m_stream));
+            }
+            delete ptr;
+          }) {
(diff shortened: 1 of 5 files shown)


core/src/HIP/Kokkos_HIP.cpp
```diff
@@ -115,11 +117,14 @@ HIP::HIP()
 HIP::HIP(hipStream_t const stream, Impl::ManageStream manage_stream)
     : m_space_instance(new Impl::HIPInternal, [](Impl::HIPInternal* ptr) {
         ptr->finalize();
+        if (static_cast<bool>(manage_stream)) {
+          KOKKOS_IMPL_HIP_SAFE_CALL(cudaStreamDestroy(ptr->m_stream));
```
dalg24, line 121:
```suggestion
          KOKKOS_IMPL_HIP_SAFE_CALL(hipStreamDestroy(ptr->m_stream));
```

core/src/HIP/Kokkos_HIP.cpp
```diff
@@ -115,11 +117,14 @@ HIP::HIP()
 HIP::HIP(hipStream_t const stream, Impl::ManageStream manage_stream)
     : m_space_instance(new Impl::HIPInternal, [](Impl::HIPInternal* ptr) {
```
crtrott, line 118: doesn't this need to capture manage_stream

dalg24, line 118: Capture clause missing

core/src/HIP/Kokkos_HIP.cpp
crtrott, line 121: and this likely should be cudaStreamDestroy

crtrott (dismissed):
Assuming that someone checks it actually works on HIP (after fixing the two things I pointed out) I am happy with this change.

Rombur:
This doesn't compile on HIP. Error line 92 in core/src/HIP/Kokkos_HIP.cpp

Rombur (approved):
Looks good and passes the testsuite.

tcclevenger (approved):
Looks alright to me.

PR #6714, opened 2024-01-12
Make initialize and finalize of the Cuda/HIP singleton less special

dalg24:
Working towards moving initialization and finalization of the execution instance internals to the constructor and destructor respectively.

Moving all logic depending on whether the current exec space instance is the default one (the one we refer to as age singleton) to `{Cuda,HIP}::impl_{initialize,finalize}`.  Note that in CUDA I chose not to use the `cuda_func_wrapper` functions because they tied to an execution space instance.

PR #6722, opened 2024-01-16
Kokkos_HIP.cpp: include Kokkos_Core.hpp to resolve errors

ndellingwood:
Resolves errors of form:
/home/jenkins/caraway-new/workspace/KokkosKernels_PullRequest_VEGA90A_ROCM560/kokkos/core/src/HIP/Kokkos_HIP.cpp:96:15: error: no member named 'hip_global_unique_token_locks' in namespace 'Kokkos::Impl'
  (void)Impl::hip_global_unique_token_locks(true);

Note: include of Kokkos_HIP_UniqueToken.hpp was insufficient as this triggered new errors: /home/ndellin/kokkos/core/src/HIP/Kokkos_HIP_UniqueToken.hpp:40:29: error: implicit instantiation of undefined template 'Kokkos::View<unsigned int *, Kokkos::HIPSpace>'
  View<uint32_t*, HIPSpace> m_locks;

### DIFF
--- a/core/src/HIP/Kokkos_HIP.cpp
+++ b/core/src/HIP/Kokkos_HIP.cpp
@@ -18,6 +18,7 @@
 #define KOKKOS_IMPL_PUBLIC_INCLUDE
 #endif
 
+#include <Kokkos_Core.hpp>
 #include <HIP/Kokkos_HIP.hpp>
 #include <HIP/Kokkos_HIP_Instance.hpp>
 


ndellingwood:
The error above began after merge of #6714 and first popped up in kokkos-kernels CI testing against kokkos@develop, will eventually show in the nightlies

core/src/HIP/Kokkos_HIP.cpp
```diff
@@ -18,6 +18,7 @@
 #define KOKKOS_IMPL_PUBLIC_INCLUDE
 #endif
 
+#include <Kokkos_Core.hpp>
```
crtrott, line 21: @ I gonna approve this for now but we need to fix this properly.

dalg24 (approved):
Probably should figure out minimal includes needed or drop the next two

core/src/HIP/Kokkos_HIP.cpp
crtrott: And properly could mean removing the other includes below, or actually figuring out the minimal set we need.

crtrott (approved):

core/src/HIP/Kokkos_HIP.cpp
ndellingwood: Yeah, this was the "heavy hammer" quick fix (including `Kokkos_HIP_UniqueToken.hpp` introduced new problems)

PR #6726, opened 2024-01-18
Guard `[MD]RangePolicy` precondition check for deprecated code 4

dalg24:
Fixup for #6617 that I merged a bit hastily

This is a breaking change so we need to guard it with `#ifndef KOKKOS_ENABLE_DEPRECATED_CODE_4`.
I opted for raising warnings otherwise and did some cleanup to avoid duplicated warning messages.

### DIFF
--- a/core/src/KokkosExp_MDRangePolicy.hpp
+++ b/core/src/KokkosExp_MDRangePolicy.hpp
@@ -333,8 +333,12 @@ struct MDRangePolicy : public Kokkos::Impl::PolicyTraits<Properties...> {
             "Kokkos::MDRangePolicy bounds error: The lower bound (" +
             std::to_string(m_lower[i]) + ") is greater than its upper bound (" +
             std::to_string(m_upper[i]) + ") in dimension " + std::to_string(i) +
-            ".";
+            ".\n";
+#if !defined(KOKKOS_ENABLE_DEPRECATED_CODE_4)
         Kokkos::abort(msg.c_str());
+#elif defined(KOKKOS_ENABLE_DEPRECATION_WARNINGS)
+        Kokkos::Impl::log_warning(msg);
+#endif
       }
 
       if (m_tile[i] <= 0) {
--- a/core/src/Kokkos_ExecPolicy.hpp
+++ b/core/src/Kokkos_ExecPolicy.hpp
@@ -128,7 +128,6 @@ class RangePolicy : public Impl::PolicyTraits<Properties...> {
   /** \brief  Total range */
   inline RangePolicy(const member_type work_begin, const member_type work_end)
       : RangePolicy(typename traits::execution_space(), work_begin, work_end) {
-    check_bounds_validity();
     set_auto_chunk_size();
   }
 
@@ -227,8 +226,12 @@ class RangePolicy : public Impl::PolicyTraits<Properties...> {
       std::string msg = "Kokkos::RangePolicy bounds error: The lower bound (" +
                         std::to_string(m_begin) +
                         ") is greater than the upper bound (" +
-                        std::to_string(m_end) + ").";
+                        std::to_string(m_end) + ").\n";
+#if !defined(KOKKOS_ENABLE_DEPRECATED_CODE_4)
       Kokkos::abort(msg.c_str());
+#elif defined(KOKKOS_ENABLE_DEPRECATION_WARNINGS)
+      Kokkos::Impl::log_warning(msg);
+#endif
     }
   }
 
--- a/core/src/Kokkos_ExecPolicy.hpp
+++ b/core/src/Kokkos_ExecPolicy.hpp
@@ -231,6 +231,8 @@ class RangePolicy : public Impl::PolicyTraits<Properties...> {
       Kokkos::abort(msg.c_str());
 #elif defined(KOKKOS_ENABLE_DEPRECATION_WARNINGS)
       Kokkos::Impl::log_warning(msg);
+      m_begin = 0;
+      m_end   = 0;
 #endif
     }
   }
--- a/core/src/Kokkos_ExecPolicy.hpp
+++ b/core/src/Kokkos_ExecPolicy.hpp
@@ -227,12 +227,13 @@ class RangePolicy : public Impl::PolicyTraits<Properties...> {
                         std::to_string(m_begin) +
                         ") is greater than the upper bound (" +
                         std::to_string(m_end) + ").\n";
-#if !defined(KOKKOS_ENABLE_DEPRECATED_CODE_4)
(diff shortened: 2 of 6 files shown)


core/src/KokkosExp_MDRangePolicy.hpp
```diff
@@ -333,8 +333,12 @@ struct MDRangePolicy : public Kokkos::Impl::PolicyTraits<Properties...> {
             "Kokkos::MDRangePolicy bounds error: The lower bound (" +
             std::to_string(m_lower[i]) + ") is greater than its upper bound (" +
             std::to_string(m_upper[i]) + ") in dimension " + std::to_string(i) +
-            ".";
+            ".\n";
+#if !defined(KOKKOS_ENABLE_DEPRECATED_CODE_4)
         Kokkos::abort(msg.c_str());
+#elif defined(KOKKOS_ENABLE_DEPRECATION_WARNINGS)
```
dalg24, line 339: I am on the fence whether to have just 
```suggestion
#elif
```
here instead.  `Impl::log_warning()` already honors the runtime option to disable warning messages.;
Let me know what you think.

ldh4: Unless `KOKKOS_ENABLE_DEPRECATION_WARNINGS` is set to be on by default, I want to vote for changing to `#else` and always produce warning messages.
It may potentially produce noisy outputs, but it'll notify that there are definitely incorrectly set bounds for a policy somewhere in the code (which hopefully were set by mistake and are trivial to correct).

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -227,8 +226,12 @@ class RangePolicy : public Impl::PolicyTraits<Properties...> {
       std::string msg = "Kokkos::RangePolicy bounds error: The lower bound (" +
                         std::to_string(m_begin) +
                         ") is greater than the upper bound (" +
-                        std::to_string(m_end) + ").";
+                        std::to_string(m_end) + ").\n";
+#if !defined(KOKKOS_ENABLE_DEPRECATED_CODE_4)
       Kokkos::abort(msg.c_str());
+#elif defined(KOKKOS_ENABLE_DEPRECATION_WARNINGS)
+      Kokkos::Impl::log_warning(msg);
+#endif
```
ldh4, line 234: For `MDRangePolicy`, there weren't any preconditions for the bounds at all prior to the merged PR, but for `RangePolicy`, its original behavior was to silently fallback to an empty range.
If to follow through a proper deprecation cycle, should enabling `KOKKOS_ENABLE_DEPRECATED_CODE` preserve its original behavior?

core/unit_test/TestRangePolicyConstructors.hpp
```diff
@@ -74,13 +74,38 @@ TEST(TEST_CATEGORY_DEATH, range_policy_invalid_bounds) {
   using Policy    = Kokkos::RangePolicy<TEST_EXECSPACE>;
   using ChunkSize = Kokkos::ChunkSize;
 
-  ASSERT_DEATH({ (void)Policy(100, 90); },
-               "Kokkos::RangePolicy bounds error: The lower bound \\(100\\) is "
-               "greater than the upper bound \\(90\\)\\.");
+  char msg[] =
+      R"WARNING(Kokkos::RangePolicy bounds error: The lower bound (100) is greater than the upper bound (90).
+)WARNING";
```
ldh4, line 79: > R"WARNING(Kokkos::RangePolicy

Is `R` at front a typo or is it for a regex magic?

dalg24: It is a [raw string literal](https://en.cppreference.com/w/cpp/language/string_literal).
I went for it when I saw all the character escaping but now I realize none of it is actually needed.

core/src/KokkosExp_MDRangePolicy.hpp
dalg24: It is `ON` by default
https://github.com/kokkos/kokkos/blob/179d2e67fd42eeb894cf9b5e976104a13de990c2/cmake/kokkos_enable_options.cmake#L52

core/src/Kokkos_ExecPolicy.hpp
dalg24: Yes it should preserve the original behavior.  I see your point.  Will fix.

core/src/impl/Kokkos_Error.cpp
```diff
@@ -64,7 +71,8 @@ std::string human_memory_size(size_t arg_bytes) {
 
 void Experimental::RawMemoryAllocationFailure::print_error_message(
     std::ostream &o) const {
-  o << "Allocation of size " << Impl::human_memory_size(m_attempted_size);
+  o << "Allocation of size "
+    << ::Kokkos::Impl::human_memory_size(m_attempted_size);
```
dalg24, line 75: I was getting error on my local build.  I do not understand why it was fine before and what changed now but it is not something worth digging into.

core/unit_test/TestMDRangePolicyConstructors.hpp
```diff
@@ -99,15 +99,37 @@ TEST(TEST_CATEGORY_DEATH, policy_invalid_bounds) {
...
   ::testing::FLAGS_gtest_death_test_style = "threadsafe";
 
-  auto dim = (Policy::inner_direction == Kokkos::Iterate::Right) ? 1 : 0;
+  auto [dim0, dim1] = (Policy::inner_direction == Kokkos::Iterate::Right)
+                          ? std::make_pair(1, 0)
+                          : std::make_pair(0, 1);
+  std::string msg1 =
+      "Kokkos::MDRangePolicy bounds error: The lower bound (100) is greater "
+      "than its upper bound (90) in dimension " +
+      std::to_string(dim0) + ".\n";
```
ldh4, line 108:
```suggestion
      std::to_string(dim0) + ".";
```

Looking at the CI error output:
```
Death test: { (void)Policy({100, 100}, {90, 90}); }
    Result: died but not with expected error.
  Expected: contains regular expression "Kokkos::MDRangePolicy bounds error: The lower bound (100) is greater than its upper bound (90) in dimension 1.\n"
Actual msg:
[  DEATH   ] Kokkos::MDRangePolicy bounds error: The lower bound (100) is greater than its upper bound (90) in dimension 1.
```
You might not need `\n`.

dalg24: Well we do strict comparison in the non death test code path.
Anyway I tried dropping it here in the definition and appending it back if not doing death tests.

crtrott (approved):

ldh4 (approved):

PR #6731, opened 2024-01-19
Reenable HIP testing on AMD GPUs

Rombur:
This PRs reenables CI on AMD using AMD GPUs. I had to disable the memory requirement test. It passes on my machine but not on the CI.

### DIFF
--- a/core/unit_test/hip/TestHIP_Memory_Requirements.cpp
+++ b/core/unit_test/hip/TestHIP_Memory_Requirements.cpp
@@ -48,6 +48,9 @@ TEST(hip, memory_requirements) {
   // we want all user-facing memory in hip to be coarse grained. As of
   // today(07.01.22) the documentation is not reliable/correct, we test the
   // memory on the device and host
+  // FIXME_HIP
+  GTEST_SKIP() << "skipping the test because the CI on MI100 returns:  error( "
+                  "hipErrorInvalidValue)";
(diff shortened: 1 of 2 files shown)


Rombur:
retest this please

core/unit_test/hip/TestHIP_Memory_Requirements.cpp
```diff
@@ -48,6 +48,9 @@ TEST(hip, memory_requirements) {
   // we want all user-facing memory in hip to be coarse grained. As of
   // today(07.01.22) the documentation is not reliable/correct, we test the
   // memory on the device and host
+  // FIXME_HIP
+  GTEST_SKIP() << "skipping the test because the CI on MI100 returns:  error( "
+                  "hipErrorInvalidValue)";
```
ldh4, line 53: If it fails only for MI100, should this skip be specific for MI100 rather than always skipping (i.e. by using `KOKKOS_ENABLE_ARCH_GFX908` or so)?

Rombur: I don't know why it fails. It could be the architecture, the version of the driver installed, it could be related to something else in the CI. For now, I just want to get the CI back up and look at that later. We should get some MI210 soon that should help to understand the issue better,

dalg24 (approved):

ldh4 (approved):

masterleinad (approved):

PR #6740, opened 2024-01-24
Make Kokkos::Tools::Impl::begin_parallel_* internally specializable

nmm0:
The rational for these changes are for both Kokkos Resilience and Kokkos Tools.

In Kokkos Resilience, we extend Kokkos with custom execution spaces that are resilient (i.e. ResOpenMP). We internally dispatch to the base execution space (OpenMP), duplicating or triplicating the kernel. This breaks Kokkos Tools because Kokkos Tools sees two begin_parallel_* in a row. This change will allow Kokkos Resilience (or also Kokkos Tools) to specialize these begin and end hooks depending on the execution space. That way, Kokkos Resilience (for example) could implement a no-op for the outer begin/end hooks. The intention is, because everything is in Impl, is that this should **never** be specialized by users.

**Design** I changed the template parameters to be identical to those that we use for specializing `Parallel*`. I did this so that the implementation can have the same specialization control as the `Parallel*` construct.

### DIFF
--- a/core/src/Kokkos_Parallel.hpp
+++ b/core/src/Kokkos_Parallel.hpp
@@ -135,7 +135,9 @@ inline void parallel_for(const std::string& str, const ExecPolicy& policy,
   uint64_t kpID = 0;
 
   ExecPolicy inner_policy = policy;
-  Kokkos::Tools::Impl::begin_parallel_for(inner_policy, functor, str, kpID);
+  Kokkos::Tools::Impl::ParallelForToolsHook<
+      FunctorType, ExecPolicy>::begin_parallel_for(inner_policy, functor, str,
+                                                   kpID);
 
   Kokkos::Impl::shared_allocation_tracking_disable();
   Impl::ParallelFor<FunctorType, ExecPolicy> closure(functor, inner_policy);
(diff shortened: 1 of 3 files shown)


dalg24:
How does that "break" the tools.  Each kernel, duplicated or not, still get a unique kernel ID.
Also you can give them different names (append a sub id or whatnot) in your resilience exec space before you forward to the underlying space.

nmm0:
> How does that "break" the tools. Each kernel, duplicated or not, still get a unique kernel ID. Also you can give them different names (append a sub id or whatnot) in your resilience exec space before you forward to the underlying space.

@dalg24 The outer kernel will have incorrect timings including NaNs

dalg24:
> > How does that "break" the tools. Each kernel, duplicated or not, still get a unique kernel ID. Also you can give them different names (append a sub id or whatnot) in your resilience exec space before you forward to the underlying space.
> 
> @dalg24 The outer kernel will have incorrect timings including NaNs

How so?

nmm0:
The reason this breaks Kokkos tools tracking is because Kokkos tools uses a global "currentEntry" to point to the current entry being timed. In the Kokkos Resilience use case, this results in the outer entry being uninitialized because it is never modified

dalg24:
It sound more like bug on the tool-side, not a core issue

nmm0:
Closing because we decided to address this in Tools

Issue #6736, opened 2024-01-23
Compilation flags seemingly passed twice to Trilinos


maartenarnst:
**Describe the bug**

It appears that when Kokkos is built as part of Trilinos, it passes compilation flags twice to Trilinos.

- https://github.com/kokkos/kokkos/blob/71a9bcae52543bd065522bf3e41b5bfa467d8015/CMakeLists.txt#L295-L301

It may be behind an issue I am seeing. 

In fact, I'm trying to compile Trilinos with the Cray compiler on the European computer LUMI, which is similar to Frontier. Here, the linking doesn't work, and the reason appears to be that in passing these flags twice, Kokkos is setting "-x hip" as part of the `CMAKE_CXX_FLAGS`. It seems CMake then passes these flags to the linker too. And the issue is then that when it's called to link, the Cray compiler doesn't react well to `-x hip`.

- https://github.com/kokkos/kokkos/blob/71a9bcae52543bd065522bf3e41b5bfa467d8015/cmake/kokkos_arch.cmake#L225 

It appears a straightforward fix would be to remove `${KOKKOSCORE_CXX_FLAGS}` from the two lines that set the `CMAKE_CXX_FLAGS` in the snippet above. The flags would then still be passed to Trilinos through the `set(KOKKOS_ALL_COMPILE_OPTIONS $<$<COMPILE_LANGUAGE:CXX> ...`.  

The duplication seems to come from PR 
- #2779.

If you agree with the suggestion to remove the `${KOKKOSCORE_CXX_FLAGS}` from the two lines that set the `CMAKE_CXX_FLAGS` or you have another suggestion, I'll be happy to make a PR.

Also tagging
- https://github.com/kokkos/kokkos/issues/5921
- @romintomasetti

masterleinad:
Fixed in #6742.

PR #6742, opened 2024-01-24
Trilinos: Don't let Kokkos set CMAKE_CXX_FLAGS

masterleinad:
Corresponds to https://github.com/trilinos/Trilinos/pull/12572. Fixes https://github.com/kokkos/kokkos/issues/6736.
We shouldn't set any `CMAKE_CXX_FLAGS` in `Kokkos`, not even for `Trilinos` since that messes with any target that doesn't depend on `Kokkos`. Also, we are already propagating flags properly from `Kokkos` to `Trilinos` so those would be duplicated.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -252,7 +252,6 @@ ENDIF()
 # subpackages
 
 ## This restores the old behavior of ProjectCompilerPostConfig.cmake
-# It sets the CMAKE_CXX_FLAGS globally to those used by Kokkos
 # We must do this before KOKKOS_PACKAGE_DECL
 IF (KOKKOS_HAS_TRILINOS)
   # Overwrite the old flags at the top-level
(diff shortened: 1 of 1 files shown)


ndellingwood:
Reminder note that we'll need to cherry-pick to release-candidate-4.2.01 to avoid clobber of https://github.com/trilinos/Trilinos/pull/12572 by the patch release

masterleinad:
@ndellingwood Feel free to block the Trilinos pull request. I mostly care about testing there (and the OK from Trilinos folks in general). I'm happy to just update with the next `Kokkos` update, in particular, if we can make it part of the patch update and that happens soon.

crtrott (approved):

ndellingwood (approved):
Trilinos CI in https://github.com/trilinos/Trilinos/pull/12572 will serve as the testing for this as changes only impact Trilinos builds

cedricchevalier19:
I think that not modifying `CMAKE_CXX_FLAGS` is a very good progress for users, even outside Trilinos.
Thanks @masterleinad .

dalg24:
The issue in the builds was machine related and at runtime.  This is OK to ignore.

dalg24 (approved):

ndellingwood:
Cherry-pick to release-candidate-4.2.01 https://github.com/kokkos/kokkos/pull/6756

PR #6747, opened 2024-01-25
std(remove-if): fixing tmp view alloc + avoid evaluating twice the predicate during final pass

romintomasetti:
We were looking to `Kokkos::Experimental::remove_if` implementation with @maartenarnst and we think we found 3 small issues that this PR seeks to fix:
1. The temporary view should be allocated on the given execution space instance.
2. The temporary view needs not be initialized, because all its elements will be set later on.
3. The `Kokkos::Experimental::Impl::StdRemoveIfStage1Functor` functor should not evaluate the predicate twice during the final pass (as this predicate can be arbitrarily costly).

### DIFF
--- a/algorithms/src/std_algorithms/impl/Kokkos_RemoveAllVariants.hpp
+++ b/algorithms/src/std_algorithms/impl/Kokkos_RemoveAllVariants.hpp
@@ -46,15 +46,14 @@ struct StdRemoveIfStage1Functor {
   void operator()(const IndexType i, IndexType& update,
                   const bool final_pass) const {
     auto& myval = m_first_from[i];
-    if (final_pass) {
-      if (!m_must_remove(myval)) {
+
+    if (!m_must_remove(myval)) {
+      if (final_pass) {
         // calling move here is ok because we are inside final pass
         // we are calling move assign as specified by the std
         m_first_dest[update] = std::move(myval);
       }
-    }
 
-    if (!m_must_remove(myval)) {
       update += 1;
     }
   }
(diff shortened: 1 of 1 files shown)


algorithms/src/std_algorithms/impl/Kokkos_RemoveAllVariants.hpp
```diff
@@ -46,15 +46,14 @@ struct StdRemoveIfStage1Functor {
   void operator()(const IndexType i, IndexType& update,
                   const bool final_pass) const {
     auto& myval = m_first_from[i];
-    if (final_pass) {
-      if (!m_must_remove(myval)) {
+
+    if (!m_must_remove(myval)) {
+      if(is_final) {
```
masterleinad, line 51:
```suggestion
      if(final_pass) {
```

romintomasetti: Indeed !

romintomasetti: Done.

masterleinad (approved):

dalg24 (approved):

dalg24:
```diff
diff --git a/algorithms/src/std_algorithms/impl/Kokkos_RemoveAllVariants.hpp b/algorithms/src/std_algorithms/impl/Kokkos_RemoveAllVariants.hpp
index 850601fec..456df43ae 100644
--- a/algorithms/src/std_algorithms/impl/Kokkos_RemoveAllVariants.hpp
+++ b/algorithms/src/std_algorithms/impl/Kokkos_RemoveAllVariants.hpp
@@ -48,7 +48,7 @@ struct StdRemoveIfStage1Functor {
     auto& myval = m_first_from[i];

     if (!m_must_remove(myval)) {
-      if(final_pass) {
+      if (final_pass) {
         // calling move here is ok because we are inside final pass
         // we are calling move assign as specified by the std
         m_first_dest[update] = std::move(myval);
@@ -107,7 +107,9 @@ IteratorType remove_if_exespace_impl(const std::string& label,
     // create helper tmp view
     using value_type    = typename IteratorType::value_type;
     using tmp_view_type = Kokkos::View<value_type*, ExecutionSpace>;
-    tmp_view_type tmp_view(Kokkos::view_alloc(Kokkos::WithoutInitializing, ex, "std_remove_if_tmp_view"), keep_count);
+    tmp_view_type tmp_view(Kokkos::view_alloc(Kokkos::WithoutInitializing, ex,
+                                              "std_remove_if_tmp_view"),
+                           keep_count);
     using tmp_readwrite_iterator_type = decltype(begin(tmp_view));

     // in stage 1, *move* all elements to keep from original range to tmp

dalg24 (changes requested):
Fix the format

dalg24 (approved):

dalg24:
Not waiting for SYCL build that cannot find a machine to run

PR #6754, opened 2024-01-27
Add checks for unsafe implicit conversions in RangePolicy

### LINKED ISSUE #6709: Check for unintended conversion of policy range arguments
Consider the following:
```
RangePolicy<IndexType<unsigned>> p(-1, 10); 
parallel_for(p, KOKKOS_LAMBDA(int i) { printf("%i\n",i); });
```
This will implicitly convert the -1 to a 4B and thus the parallel for will simply not execute anything, but its not considered invalid input. We could intercept that by templating on the constructor arguments, and then do a safe checked conversion to index type. 

https://godbolt.org/z/8rfo19G4d

ldh4:
Resolves #6709 

Added checks in RangePolicy to detect non value-preserving implicit conversions between input bounds and the policy's `IndexType`.

Cases considered:
 - `signed` arg type to `unsigned` index types
 - `unsigned` arg type to `signed` index types
 - Narrowing conversions

### DIFF
--- a/core/src/Kokkos_ExecPolicy.hpp
+++ b/core/src/Kokkos_ExecPolicy.hpp
@@ -115,29 +115,18 @@ class RangePolicy : public Impl::PolicyTraits<Properties...> {
         m_granularity_mask(0) {}
 
   /** \brief  Total range */
+#ifndef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  template <typename IndexType1, typename IndexType2,
+            std::enable_if_t<(std::is_convertible_v<IndexType1, member_type> &&
+                              std::is_convertible_v<IndexType2, member_type>),
+                             bool> = false>
+  inline RangePolicy(const typename traits::execution_space& work_space,
+                     const IndexType1 work_begin, const IndexType2 work_end)
+#else
   inline RangePolicy(const typename traits::execution_space& work_space,
                      const member_type work_begin, const member_type work_end)
+#endif
       : m_space(work_space),
-        m_begin(work_begin),
-        m_end(work_end),
-        m_granularity(0),
-        m_granularity_mask(0) {
-    check_bounds_validity();
-    set_auto_chunk_size();
-  }
-
-  /** \brief  Total range */
-  inline RangePolicy(const member_type work_begin, const member_type work_end)
-      : RangePolicy(typename traits::execution_space(), work_begin, work_end) {}
-
-#ifndef KOKKOS_ENABLE_DEPRECATED_CODE_4
-  template <typename IndexType,
-            std::enable_if_t<(std::is_integral_v<IndexType> &&
-                              !std::is_same_v<IndexType, member_type> &&
-                              std::is_convertible_v<IndexType, member_type>),
-                             bool> = false>
-  inline RangePolicy(const IndexType work_begin, const IndexType work_end)
-      : m_space(typename traits::execution_space()),
         m_begin(work_begin),
         m_end(work_end),
         m_granularity(0),
@@ -147,50 +136,59 @@ class RangePolicy : public Impl::PolicyTraits<Properties...> {
     check_bounds_validity();
     set_auto_chunk_size();
   }
-#endif
 
   /** \brief  Total range */
+#ifndef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  template <typename IndexType1, typename IndexType2,
+            std::enable_if_t<(std::is_convertible_v<IndexType1, member_type> &&
+                              std::is_convertible_v<IndexType2, member_type>),
+                             bool> = false>
+  inline RangePolicy(const IndexType1 work_begin, const IndexType2 work_end)
+#else
+  inline RangePolicy(const member_type work_begin, const member_type work_end)
+      : RangePolicy(typename traits::execution_space(), work_begin, work_end) {}
+
+  /** \brief  Total range */
+#ifndef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  template <typename IndexType1, typename IndexType2, typename... Args,
+            std::enable_if_t<(std::is_convertible_v<IndexType1, member_type> &&
+                              std::is_convertible_v<IndexType2, member_type>),
+                             bool> = false>
+  inline RangePolicy(const typename traits::execution_space& work_space,
+                     const IndexType1 work_begin, const IndexType2 work_end,
+                     Args... args)
+#else
   template <class... Args>
   inline RangePolicy(const typename traits::execution_space& work_space,
                      const member_type work_begin, const member_type work_end,
                      Args... args)
+#endif
       : m_space(work_space),
         m_begin(work_begin),
         m_end(work_end),
         m_granularity(0),
         m_granularity_mask(0) {
+    check_conversion_safety(work_begin);
+    check_conversion_safety(work_end);
     check_bounds_validity();
     set_auto_chunk_size();
     set(args...);
   }
 
   /** \brief  Total range */
+#ifndef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  template <typename IndexType1, typename IndexType2, typename... Args,
+            std::enable_if_t<(std::is_convertible_v<IndexType1, member_type> &&
+                              std::is_convertible_v<IndexType2, member_type>),
+                             bool> = false>
+  inline RangePolicy(const IndexType1 work_begin, const IndexType2 work_end,
+                     Args... args)
+#else
   template <class... Args>
   inline RangePolicy(const member_type work_begin, const member_type work_end,
                      Args... args)
       : RangePolicy(typename traits::execution_space(), work_begin, work_end,
                     args...) {}
-
-#ifndef KOKKOS_ENABLE_DEPRECATED_CODE_4
-  template <typename IndexType, typename... Args,
-            std::enable_if_t<(std::is_integral_v<IndexType> &&
-                              !std::is_same_v<IndexType, member_type> &&
-                              std::is_convertible_v<IndexType, member_type>),
-                             bool> = false>
-  inline RangePolicy(const IndexType work_begin, const IndexType work_end,
-                     Args... args)
-      : m_space(typename traits::execution_space()),
-        m_begin(work_begin),
-        m_end(work_end),
-        m_granularity(0),
-        m_granularity_mask(0) {
-    check_conversion_safety(work_begin);
-    check_conversion_safety(work_end);
-    check_bounds_validity();
-    set_auto_chunk_size();
-    set(args...);
-  }
-#endif
 
  private:
   inline void set() {}
@@ -275,12 +273,14 @@ class RangePolicy : public Impl::PolicyTraits<Properties...> {
   }
 
   template <typename IndexType>
-  void check_conversion_safety(IndexType bound) {
+  static void check_conversion_safety(const IndexType bound) {
     std::string msg =
         "Kokkos::RangePolicy bound type error: unsafe implicit conversion may "
         "not preserve the original value.\n";
     bool warn = false;
 
+#if !defined(KOKKOS_ENABLE_DEPRECATED_CODE_4) || \
+    defined(KOKKOS_ENABLE_DEPRECATION_WARNINGS)
     if constexpr (std::is_signed_v<IndexType> !=
                   std::is_signed_v<member_type>) {
       // check signed to unsigned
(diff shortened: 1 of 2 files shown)


core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -131,6 +132,25 @@ class RangePolicy : public Impl::PolicyTraits<Properties...> {
     set_auto_chunk_size();
   }
 
+#ifndef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  template <typename IndexType,
+            std::enable_if_t<(std::is_integral_v<IndexType> &&
+                              !std::is_same_v<IndexType, member_type> &&
+                              std::is_convertible_v<IndexType, member_type>),
+                             bool> = false>
```
dalg24, line 140: Why did you chose to add a new templated constructors instead of updating the existing ones?

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -232,6 +273,40 @@ class RangePolicy : public Impl::PolicyTraits<Properties...> {
...
       m_begin = 0;
       m_end   = 0;
+#ifdef KOKKOS_ENABLE_DEPRECATION_WARNINGS
+      Kokkos::Impl::log_warning(msg);
+#endif
+    }
+  }
+
+  template <typename IndexType>
+  void check_conversion_safety(IndexType bound) {
```
dalg24, line 283: Can it be static?

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -232,6 +273,40 @@ class RangePolicy : public Impl::PolicyTraits<Properties...> {
...
+    std::string msg =
+        "Kokkos::RangePolicy bound type error: unsafe implicit conversion may "
+        "not preserve the original value.\n";
```
dalg24, line 286: Can the error message print the value and the type it is trying to convert to?

core/src/Kokkos_ExecPolicy.hpp
ldh4: There wasn't a strong reason for it. Part of the minor reasoning was for better readability so that the existing constructor isn't going to be heavily interleaved by multiple `ifndef KOKKOS_ENABLE_DEPRECATED_CODE_4`.
But I just realized that the constructor with an execution space should also have this check. I'll make the updates and integrate this conversion into the existing constructors instead.

ldh4: Moved this change into existing constructors. [67fcdb0](https://github.com/kokkos/kokkos/pull/6754/commits/67fcdb02ef547355bce81dea0fc8942842c13680).

core/src/Kokkos_ExecPolicy.hpp
ldh4: Changed to static. 67fcdb0.

core/src/Kokkos_ExecPolicy.hpp
dalg24: It is good enough for now

core/unit_test/TestRangePolicyConstructors.hpp
```diff
@@ -121,4 +122,42 @@ TEST(TEST_CATEGORY_DEATH, range_policy_invalid_bounds) {
...
+
+#else
+  if (!Kokkos::show_warnings()) {
+    GTEST_SKIP() << "Kokkos warning messages are disabled";
+  }
+
+  {
+    UIntPolicy policy(-1, 10);
+    ASSERT_EQ(policy.begin(), 0u);
+    ASSERT_EQ(policy.end(), 0u);
```
dalg24, line 152: Check that warnings are raised if deprecation warnings are enabled and show_warnings() returns true

crtrott:
I believe the logic isn't quite right for what it should check. I would suggest we just copy the stuff from mdspan (and backs-porting):

https://github.com/llvm/llvm-project/blob/0e8eb445db0cc2552d9d077b527a43c779785cb9/libcxx/include/__mdspan/extents.h#L228-L252

Also the function for the representability check should only be called in debug mode maybe?

dalg24:
DongHun and I met earlier today.  I shared an alternative implementation https://godbolt.org/z/WEKM444n4 and we rediscovered that MDRange had https://github.com/kokkos/kokkos/blob/4c94f089bd636dbcec5cee6e41d7c131ce3b79cc/core/src/KokkosExp_MDRangePolicy.hpp#L74-L87
We also discussed that C++20 has `std::in_range` that does pretty much what we want.

We decided to focus on making sure that we got the tests right and agreed we can refactor as needed later.

ldh4:
> I believe the logic isn't quite right for what it should check. I would suggest we just copy the stuff from mdspan (and backs-porting):
> https://github.com/llvm/llvm-project/blob/0e8eb445db0cc2552d9d077b527a43c779785cb9/libcxx/include/__mdspan/extents.h#L228-L252

I think there may be some disparities between what the mdspan extent is requiring and what is currently accepted for indicies in `RangePolicy`.

For example of this case:
```
Kokkos::RangePolicy<Kokkos::IndexType<int>> p(-3, -1);
```
([in godbolt](https://godbolt.org/z/jEjTEYd5T))
Although not a very intuitive example, currently it's valid to pass in negative values as bounds in `RangePolicy` if `IndexType` is set as a signed integral type. Also the [kokkos documentation](https://kokkos.org/kokkos-core-wiki/API/core/policies/RangePolicy.html?highlight=rangepolicy#common-arguments-for-all-execution-policies) on `RangePolicy` doesn't seem to be requiring `IndexType` to be a positive value. (`IndexType` defaults to `int64_t`)

But looking at the comment in the mdspan extent's representability check, the precondition is that an extent must be a positive integer:
```
// Function to check whether a value is representable as another type
// value must be a positive integer otherwise returns false
```
So, there is an upfront check to disallow any negative extents, and because of this requirement, the rest of representability check would allow `signed` to `unsigned` conversion (which would be an unsafe implicit conversion for `RangePolicy` described in #6709 that this PR aims to catch).

I think we would need to establish that `RangePolicy` should not be taking in negative values for bounds in order to use that mdspan extent check.

core/unit_test/TestRangePolicyConstructors.hpp
ldh4: There is a minor problem with this test. When `KOKKOS_ENABLE_DEPRECATED_CODE_4` is not enabled, the `RangePolicy` constructors are not templated on the bound types. So, the bound args are implicitly converted to `member_type` and will not be producing the unsafe conversion warning from this PR. Instead, it will only raise the warning from `check_bounds_validity` check.

In the interest of outputting the warning message even when `KOKKOS_ENABLE_DEPRECATED_CODE_4` is defined, the tentative solution is to use this condition instead:
```
#if !defined(KOKKOS_ENABLE_DEPRECATED_CODE_4) || defined(KOKKOS_ENABLE_DEPRECATION_WARNINGS)
```
See [c8befbd](https://github.com/kokkos/kokkos/pull/6754/commits/c8befbdfa8c752ac885edfa5a1f3c4379c9c7654). I don't think this is a desired flag condition to leave it as.

dalg24:
DongHun is correct.  `std::mdspan` says
> representable as a nonnegative value of type `index_type`

and we allow negative values in RangePolicy

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -114,42 +115,92 @@ class RangePolicy : public Impl::PolicyTraits<Properties...> {
...
   inline RangePolicy(const typename traits::execution_space& work_space,
                      const member_type work_begin, const member_type work_end,
                      Args... args)
+#endif
       : m_space(work_space),
         m_begin(work_begin),
         m_end(work_end),
         m_granularity(0),
         m_granularity_mask(0) {
+#ifdef KOKKOS_ENABLE_DEBUG
```
dalg24, line 178: Why did you guard for `KOKKOS_ENABLE_DEBUG `?
I think we want to enable only based on `KOKKOS_ENABLE_DEPRECATED_CODE_4` and `KOKKOS_ENABLE_DEPRECATION_WARNINGS`.
`KOKKOS_ENABLE_DEBUG` is an "opt-in" option but we would really want users to get the runtime warnings unless they disable warnings.

core/unit_test/TestRangePolicyConstructors.hpp
```diff
@@ -121,4 +122,82 @@ TEST(TEST_CATEGORY_DEATH, range_policy_invalid_bounds) {
...
 
+TEST(TEST_CATEGORY_DEATH, range_policy_implicitly_converted_bounds) {
+  using UIntIndexType = Kokkos::IndexType<unsigned>;
+  using IntIndexType  = Kokkos::IndexType<int>;
+  using UIntPolicy    = Kokkos::RangePolicy<TEST_EXECSPACE, UIntIndexType>;
+  using IntPolicy     = Kokkos::RangePolicy<TEST_EXECSPACE, IntIndexType>;
+
+#ifndef KOKKOS_ENABLE_DEBUG
+  GTEST_SKIP();
+#endif
```
dalg24, line 133: Just drop this and the guard in the source file and (assuming that the tests are passing) I will approve.

core/src/Kokkos_ExecPolicy.hpp
ldh4: Removed this guard and in the unit test.

> Also the function for the representability check should only be called in debug mode maybe?

It was added in attempt to address this feedback.

core/unit_test/TestRangePolicyConstructors.hpp
ldh4: Removed the ifdef guard and gtest_skip.

dalg24:
Sometimes it’s best to dismiss the reviewer’s suggestion:)

dalg24 (approved):

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -114,42 +115,88 @@ class RangePolicy : public Impl::PolicyTraits<Properties...> {
         m_granularity_mask(0) {}
 
   /** \brief  Total range */
+#if !defined(KOKKOS_ENABLE_DEPRECATED_CODE_4) || \
+    defined(KOKKOS_ENABLE_DEPRECATION_WARNINGS)
```
masterleinad, line 119: Whether or not we disable deprecation warnings shouldn't affect the interface exposed. Whether we warn or error out is determined in the implementation of `check_conversion_safety` anyway and there the guards make sense.
```suggestion
#ifndef KOKKOS_ENABLE_DEPRECATED_CODE_4
```
What happens if `is_convertible` is false anyway? Wouldn't that already result in a compilation error? In that case, we can just disable the new interface unconditionally anyway, right?

masterleinad (changes requested):

core/src/Kokkos_ExecPolicy.hpp
masterleinad: Updated my comment.

ldh4: > Whether we warn or error out is determined in the implementation of `check_conversion_safety` anyway and there the guards make sense.
> What happens if `is_convertible` is false anyway? Wouldn't that already result in a compilation error? In that case, we can just enable the new interface unconditionally anyway, right?

Yes, since the old interface simply accepted arguments that are implicitly convertible to the `member_type`, semantically it should be no different from enforcing the explicit `is_convertible` check to the interface. I placed in those ifdef guards as a form of compatibility safety and mainly to follow our regular process for changing user facing interface.
But I am for using the new interface right away and leave the ifdef gurads only in `check_conversion_safety` as well.

ldh4: > >Whether or not we disable deprecation warnings shouldn't affect the interface exposed.

> I agree with this statement in general. But the change you suggest would inhibit diagnosing narrowing conversions. (I am not saying we cannot come up with something else, just that the suggestion does not work.)

@dalg24, any opinion on replacing the old interface with the new interface right away and leave the two ifdef guards only in `check_conversion_saftey`?

dalg24: > @dalg24, any opinion on replacing the old interface with the new interface right away and leave the two ifdef guards only in check_conversion_saftey?

OK with it.

ldh4: Switched to use the new interfaces and removed ifdef guards. [3065552](https://github.com/kokkos/kokkos/pull/6754/commits/3065552efd27d9a9782707b8804b394f2af2bbdc).

masterleinad (approved):

dalg24:
Ignoring failure of one of the HIP build (machine issue) and 
```
5: [ RUN      ] cuda.workgraph_fib
5: /var/jenkins/workspace/Kokkos_PR-6754/core/unit_test/TestWorkGraph.hpp:127: Failure
5: Expected equality of these values:
5:   h_values(0)
5:     Which is: 5168
5:   full_fibonacci(m_input)
5:     Which is: 6765
5: [  FAILED  ] cuda.workgraph_fib (1416 ms)

PR #3966, opened 2021-04-26
Fence Profiling Support in all backends

DavidPoliakoff:
So, this one needs conversation, we're really at a "design review" point.

One thing I've been promising for ages, but haven't gotten around to doing, is to add fence profiling to Kokkos. We have the callbacks (#2891), but they're currently unused.

What I see, is that we basically need a Kokkos wrapper around any fencing function. cudaDeviceSynchronize,cudaStreamSynchronize, that take in at a minimum a string name, _probably_ also a deviceID? Then we never directly call the fencing function, only the wrapper. But this change radiates out. If cudaStreamSynchronize takes a name, then so must Kokkos::Cuda::fence(), so that it can forward it to cudaStreamSynchronize. We also need compatibility, so we need a Kokkos::Cuda::fence with no args, that calls the named version.

I've sketched what this might look like in Cuda. Note that the only changes I make are to Cuda, we also need to modify generic code (CopyViews and the like), but those changes wouldn't compile.

*What I need to proceed* is for somebody to sign off on the design. Once you do that, I will write _the infrastructure for this_ in every backend. My hope is that we can merge at that point. Once we have the infrastructure, I (or a backend owner) can start labeling fences, and starting to put this in generic code as well (like that in CopyViews).

PR #6762, opened 2024-01-30
Avoid calling wrapper functions with singleton in Kokkos_CudaSpace.cpp where we want a different device

masterleinad:
Extracted from #6737.  This pull request replaces all instances of calling CudaInternal::singleton with a wrapper in `Kokkos_CudaSpace.cpp` where we can get the correct device id (all other places should be fine using the singleton anyway).

We inline `cuda_stream_synchronize` with `Kokkos::Tools::Experimental::SpecialSynchronizationCases` since that is only used in one place and doesn't provide a `device_id` parameter.

### DIFF
--- a/core/src/Cuda/Kokkos_Cuda_Error.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_Error.hpp
@@ -27,10 +27,6 @@
 namespace Kokkos {
 namespace Impl {
 
-void cuda_stream_synchronize(
-    const cudaStream_t stream,
-    Kokkos::Tools::Experimental::SpecialSynchronizationCases reason,
-    const std::string& name);
 void cuda_device_synchronize(const std::string& name);
 void cuda_stream_synchronize(const cudaStream_t stream,
                              const std::string& name);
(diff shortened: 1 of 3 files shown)


core/src/Cuda/Kokkos_Cuda_Error.hpp
```diff
@@ -27,10 +27,6 @@
 namespace Kokkos {
 namespace Impl {
 
-void cuda_stream_synchronize(
```
dalg24, line 30: Introduced in #3966

dalg24 (commented):
Please improve the title and the description as these are not all the instances of reaching out to singleton to call the CUDA API wrapper functions.
This handles:
a/ the obvious one that are the allocation/free since the associated memory spaces now have a device and a stream stored as data members
b/ the one inlining the `cuda_stream_synchronize` call which was exclusively used in the 2-arguments `DeepCopyAsyncCuda`

masterleinad:
> Please improve the title and the description as these are not all the instances of reaching out to singleton to call the CUDA API wrapper functions.

I updated both.

tcclevenger (approved):

dalg24 (approved):

masterleinad:
All Cuda builds except for `CUDA-11.0.3-Clang-Tidy` are passing and the latter reports "disk quota exceeded".

PR #6764, opened 2024-01-30
Drop 2-arguments `ZeroMemset` constructor overloads

dalg24:
Rational: Only used once at a place where we should have passed a default-constructed execution space to be consistent with what we do in the else-clause when we call `contiguous_fill`.
The Cuda 2-args `ZeroMemset` was problematic for the multi-gpu support (see https://github.com/kokkos/kokkos/pull/6737#discussion_r1471959006).

This work supersedes https://github.com/kokkos/kokkos/pull/6187

Drive-by change: leverage CTAD when calling ZeroMemset to improve code readability.


masterleinad (approved):

dalg24:
Retest this please

Rombur (approved):

tcclevenger (approved):
Thanks, @dalg24. I'll close the other PR.

PR #6765, opened 2024-01-31
Remove redundant calls in rangepolicy constructors

ldh4:
Removed redundant function calls in delegating constructors in RangePolicy.

### DIFF
--- a/core/src/Kokkos_ExecPolicy.hpp
+++ b/core/src/Kokkos_ExecPolicy.hpp
@@ -127,9 +127,7 @@ class RangePolicy : public Impl::PolicyTraits<Properties...> {
 
   /** \brief  Total range */
   inline RangePolicy(const member_type work_begin, const member_type work_end)
-      : RangePolicy(typename traits::execution_space(), work_begin, work_end) {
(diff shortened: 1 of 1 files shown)


core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -150,11 +148,7 @@ class RangePolicy : public Impl::PolicyTraits<Properties...> {
   template <class... Args>
   inline RangePolicy(const member_type work_begin, const member_type work_end,
                      Args... args)
-      : RangePolicy(typename traits::execution_space(), work_begin, work_end) {
-    check_bounds_validity();
-    set_auto_chunk_size();
-    set(args...);
```
dalg24, line 156: The `set(args...);` would not be executed with the change you propose.

ldh4: Fixed to call the constructor that takes in a parameter pack `args`.

masterleinad (approved):

dalg24:
Ignoring unrelated failures (exceeded disk quota) in the CUDA builds

PR #6767, opened 2024-01-31
Cuda multi-GPU support: Pass the correct device id to get_cuda_kernel_func_attributes

masterleinad:
Part of #6091. This pull request makes sure that we pass the correct device id to `get_cuda_kernel_func_attributes`. I decided to replace the `static` `func_attr` variables with a `std::map` (similar to #6753) so that we still cache the computed value the first time we run the respective functor on a particular device.

### DIFF
--- a/core/src/Cuda/Kokkos_Cuda_KernelLaunch.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_KernelLaunch.hpp
@@ -128,21 +128,19 @@ inline void check_shmem_request(CudaInternal const* cuda_instance, int shmem) {
 // These functions need to be templated on DriverType and LaunchBounds
 // so that the static bool is unique for each type combo
 // KernelFuncPtr does not necessarily contain that type information.
-// FIXME_CUDA_MULTIPLE_DEVICES
 template <class DriverType, class LaunchBounds, class KernelFuncPtr>
 const cudaFuncAttributes& get_cuda_kernel_func_attributes(
-    const KernelFuncPtr& func) {
+    int cuda_device, const KernelFuncPtr& func) {
   // Only call cudaFuncGetAttributes once for each unique kernel
   // by leveraging static variable initialization rules
-  auto wrap_get_attributes = [&]() -> cudaFuncAttributes {
+  static std::map<int, cudaFuncAttributes> func_attr;
+  if (func_attr.find(cuda_device) == func_attr.end()) {
     cudaFuncAttributes attr;
-    KOKKOS_IMPL_CUDA_SAFE_CALL(
-        (CudaInternal::singleton().cuda_func_get_attributes_wrapper(&attr,
-                                                                    func)));
-    return attr;
-  };
-  static cudaFuncAttributes func_attr = wrap_get_attributes();
-  return func_attr;
+    KOKKOS_IMPL_CUDA_SAFE_CALL(cudaSetDevice(cuda_device));
+    KOKKOS_IMPL_CUDA_SAFE_CALL(cudaFuncGetAttributes(&attr, func));
+    func_attr.insert(std::pair{cuda_device, attr});
+  }
+  return func_attr[cuda_device];
 }
 
 template <class DriverType, class LaunchBounds, class KernelFuncPtr>
@@ -682,10 +680,10 @@ struct CudaParallelLaunchImpl<
     }
   }
 
-  static cudaFuncAttributes get_cuda_func_attributes() {
+  static cudaFuncAttributes get_cuda_func_attributes(int cuda_device) {
     return get_cuda_kernel_func_attributes<
         DriverType, Kokkos::LaunchBounds<MaxThreadsPerBlock, MinBlocksPerSM>>(
-        base_t::get_kernel_func());
+        cuda_device, base_t::get_kernel_func());
   }
 };
 
--- a/core/src/Cuda/Kokkos_Cuda_Parallel_MDRange.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_Parallel_MDRange.hpp
@@ -332,9 +332,9 @@ class ParallelReduce<CombinedFunctorReducerType,
     using closure_type =
         Impl::ParallelReduce<CombinedFunctorReducer<FunctorType, ReducerType>,
                              Policy, Kokkos::Cuda>;
-    cudaFuncAttributes attr =
-        CudaParallelLaunch<closure_type,
-                           LaunchBounds>::get_cuda_func_attributes();
+    cudaFuncAttributes attr = CudaParallelLaunch<closure_type, LaunchBounds>::
+        get_cuda_func_attributes(
+            m_policy.space().impl_internal_space_instance()->m_cudaDev);
     while (
         (n && (maxShmemPerBlock < shmem_size)) ||
         (n >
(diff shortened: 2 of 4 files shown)


dalg24 (commented):
Why did you choose to go this route VS setting the right device when entering the `Parallel{For,Reduce,Scan}` construct and stripping the wrappers.

masterleinad:
> Why did you choose to go this route VS setting the right device when entering the `Parallel{For,Reduce,Scan}` construct and stripping the wrappers.

We want to be explicit about which device to use and if a function doesn't set the device it is implied that we mean to use the default one (which is oftentimes made more clear by using `CudaInternal::singleton()` or `Cuda()`).
I want to evaluate all the wrappers after this pull request when we are pretty much happy with all the places to use wrappers or not.

dalg24:
I think that we should adopt the "setting the right device when entering the Parallel{For,Reduce,Scan} construct and stripping the wrappers" approach.

masterleinad:
> I think that we should adopt the "setting the right device when entering the Parallel{For,Reduce,Scan} construct and stripping the wrappers" approach.

You want to get rid of `get_cuda_kernel_func_attributes` and `get_cuda_func_attributes`?

dalg24:
> You want to get rid of `get_cuda_kernel_func_attributes` and `get_cuda_func_attributes`?

I do not want to change their signature.  Why are you asking?

masterleinad:
> I do not want to change their signature. Why are you asking?

I'm trying to understand what you want. I don't it's a good idea to silently assume we have set the device correctly. IIRC we discussed in our last meeting that we want to pass the device id explicitly in case it's not the default one and we can't deduce it from other arguments.

core/src/Cuda/Kokkos_Cuda_Parallel_MDRange.hpp
```diff
@@ -332,9 +332,9 @@ class ParallelReduce<CombinedFunctorReducerType,
     using closure_type =
         Impl::ParallelReduce<CombinedFunctorReducer<FunctorType, ReducerType>,
                              Policy, Kokkos::Cuda>;
-    cudaFuncAttributes attr =
-        CudaParallelLaunch<closure_type,
-                           LaunchBounds>::get_cuda_func_attributes();
+    cudaFuncAttributes attr = CudaParallelLaunch<closure_type, LaunchBounds>::
+        get_cuda_func_attributes(
+            m_policy.space().impl_internal_space_instance()->m_cudaDev);
```
dalg24, line 337: that is not the fastest way to get the device id.
```suggestion
            m_policy.space().cuda_device());
```

dalg24: Fast -> concise, less convoluted

dalg24 (commented):
Change the getting the device id from the policy and I will approve

core/src/Cuda/Kokkos_Cuda_KernelLaunch.hpp
```diff
@@ -128,21 +128,19 @@ inline void check_shmem_request(CudaInternal const* cuda_instance, int shmem) {
...
-    KOKKOS_IMPL_CUDA_SAFE_CALL(
-        (CudaInternal::singleton().cuda_func_get_attributes_wrapper(&attr,
-                                                                    func)));
-    return attr;
-  };
-  static cudaFuncAttributes func_attr = wrap_get_attributes();
-  return func_attr;
+    KOKKOS_IMPL_CUDA_SAFE_CALL(cudaSetDevice(cuda_device));
+    KOKKOS_IMPL_CUDA_SAFE_CALL(cudaFuncGetAttributes(&attr, func));
+    func_attr.insert(std::pair{cuda_device, attr});
```
dalg24, line 141:
```suggestion
    func_attr.emplace(cuda_device, attr);
```

core/src/Cuda/Kokkos_Cuda_KernelLaunch.hpp
```diff
@@ -128,21 +128,19 @@ inline void check_shmem_request(CudaInternal const* cuda_instance, int shmem) {
...
-                                                                    func)));
-    return attr;
-  };
-  static cudaFuncAttributes func_attr = wrap_get_attributes();
-  return func_attr;
+    KOKKOS_IMPL_CUDA_SAFE_CALL(cudaSetDevice(cuda_device));
+    KOKKOS_IMPL_CUDA_SAFE_CALL(cudaFuncGetAttributes(&attr, func));
+    func_attr.insert(cuda_device, attr);
+  }
+  return func_attr[cuda_device];
```
dalg24, line 143: If we are worried about latency we should do
```c++
auto search = func_attr.find(cuda_device);
if (search == func_attr.end()) {
  ...
  return attr;
}
return search->second;

**Review by dalg24 — APPROVED** [review]

File: core/src/Cuda/Kokkos_Cuda_KernelLaunch.hpp (hunk 3 above)
Comment (masterleinad) [reply core/src/Cuda/Kokkos_Cuda_KernelLaunch.hpp:143]: Running `Kokkos Launch Latency Test`:
This pull request
```
N V K M time_no_fence time_fence (time_no_fence_fenced)
100000000 1 1 20 parallel_for: 3.410850 3239.667250 ( 3237.208250 )
100000000 16 1 20 parallel_for: 3.512550 1010.526850 ( 1004.075800 )
100000000 200 1 20 parallel_for: 964.569050 1016.016200 ( 1014.913050 )
100000000 3000 1 20 parallel_for: 967.518000 1019.864300 ( 1017.874550 )
100000000 30000 1 20 parallel_for: 875.876850 1050.608750 ( 1028.366200 )
```
before
```
N V K M time_no_fence time_fence (time_no_fence_fenced)
100000000 1 1 20 parallel_for: 3.399300 3238.960950 ( 3237.724650 )
100000000 16 1 20 parallel_for: 3.511050 1011.017300 ( 1004.226550 )
100000000 200 1 20 parallel_for: 964.840100 1016.107300 ( 1015.046800 )
100000000 3000 1 20 parallel_for: 967.730900 1019.796150 ( 1018.087500 )
100000000 30000 1 20 parallel_for: 879.111450 1051.797000 ( 1031.848700 )
```
I don't see

**Review by crtrott — APPROVED** [review]

**Comment by dalg24** [comment]
The Clang+CUDA build failed with
```
/var/jenkins/workspace/Kokkos_PR-6767/core/src/Cuda/Kokkos_Cuda_Parallel_MDRange.hpp:45:66: error: too few arguments to function call, single argument 'cuda_device' was not specified [clang-diagnostic-error]
                         LaunchBounds>::get_cuda_func_attributes();

dalg24:
You missed https://github.com/kokkos/kokkos/blob/8d553299210361ee3b5f4c41025f2791487a81a2/core/src/Cuda/Kokkos_Cuda_Parallel_MDRange.hpp#L43-L45

dalg24 (changes requested):
Please fix the MDRange `get_cuda_func_attributes()` occurrence you missed and audit that there are no other places that were missed.

masterleinad:
> Please fix the MDRange `get_cuda_func_attributes()` occurrence you missed and audit that there are no other places that were missed.

Fixed. I didn't find any other place.

dalg24 (approved):

dalg24:
Ignoring failures in HIP ROCm 5.6 build

dalg24 (commented):
Forgot to formally approved

PR #6769, opened 2024-01-31
Get rid of `ZeroMemset`'s silly trailing value argument

dalg24:
Follow up on #6764 
`ZeroMemset` was talking a trailing value argument that was ignored in virtually all implementation and intended to be all bits set to zero.  This PR drops that value argument.
I think this is more intuitive.

`ZeroMemset(exec, v, /*ignored=*/ValueType(0)) -> ZeroMemset(exec, v)`

### DIFF
--- a/core/src/Kokkos_CopyViews.hpp
+++ b/core/src/Kokkos_CopyViews.hpp
@@ -1336,13 +1336,12 @@ inline void contiguous_fill(
 // Default implementation for execution spaces that don't provide a definition
 template <typename ExecutionSpace, class ViewType>
 struct ZeroMemset {
-  ZeroMemset(const ExecutionSpace& exec_space, const ViewType& dst,
-             typename ViewType::const_value_type& value) {
-    contiguous_fill(exec_space, dst, value);
-  }
-
-  ZeroMemset(const ViewType& dst, typename ViewType::const_value_type& value) {
-    contiguous_fill(ExecutionSpace(), dst, value);
+  ZeroMemset(const ExecutionSpace& exec_space, const ViewType& dst) {
+    using ValueType = typename ViewType::value_type;
+    alignas(alignof(
+        ValueType)) char zero_initialized_storage[sizeof(ValueType)] = {};
+    contiguous_fill(exec_space, dst,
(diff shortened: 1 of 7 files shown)


core/src/Kokkos_CopyViews.hpp
```diff
@@ -1336,13 +1336,12 @@ inline void contiguous_fill(
 // Default implementation for execution spaces that don't provide a definition
 template <typename ExecutionSpace, class ViewType>
 struct ZeroMemset {
-  ZeroMemset(const ExecutionSpace& exec_space, const ViewType& dst,
-             typename ViewType::const_value_type& value) {
-    contiguous_fill(exec_space, dst, value);
-  }
-
-  ZeroMemset(const ViewType& dst, typename ViewType::const_value_type& value) {
-    contiguous_fill(ExecutionSpace(), dst, value);
```
dalg24, line 1345: This 2nd constructor is not used anymore.  I should have removed it in the previous PR but I missed it.

core/src/Kokkos_CopyViews.hpp
```diff
@@ -1336,13 +1336,12 @@ inline void contiguous_fill(
...
+  ZeroMemset(const ExecutionSpace& exec_space, const ViewType& dst) {
+    using ValueType = typename ViewType::value_type;
+    alignas(alignof(
+        ValueType)) char zero_initialized_storage[sizeof(ValueType)] = {};
```
dalg24, line 1342: Probably should have picked `unsigned char` instead of `char`.  Will amend the change.

dalg24:
> I don't think it's worth changing (apart from removing the unused constructor).

My perspective it is better to have the complexity at a single place when we want an object with all zero bits and the interface fill natural rather than the current version where we pay for that odd signature everywhere.

nmm0 (approved):
Looks good to me! (I agree with removing to simplify code)

tcclevenger (approved):
I agree with removing extra arg for simplicity.

crtrott (approved):

crtrott:
ROCM 5.6 build failed restart?

dalg24:
> ROCM 5.6 build failed restart?

Nah.  The other HIP build passed.
image

Issue #6323, opened 2023-07-31
Add RISC-V support to the build system


diehlpk:
I played around testing Kokkos on RISC-V. Kokkos built on RISC-V using gcc out of the box. However, CMake needed some changes for the architecture to add correct compiler flags. 

Is there interest to support RISC-V? 

If so I am willing to prepare a pull request.

I assume I would need to add a new architecture flag for the RISC-V CPU or board.

diehlpk:
Kokkos scales well on RISC V, I will post the preprint of our SC workshop paper soon.

lucbv:
@diehlpk this sounds like a good idea but in general this has not happened because there is no hardware beyond development boards are at least there was not until recently?
I would be happy to have a discussion on the topic if you want.
I am guessing that you are sending your paper to the RISC-V for HPC workshop?

diehlpk:
Yes, your guess it correct.

ajpowelsnl:
> Kokkos scales well on RISC V, I will post the preprint of our SC workshop paper soon.

Many thanks for the suggestion, @diehlpk.  It would be great if you could develop a PR for this feature.

diehlpk:
There might be soon a desktop machine available:

https://www.crowdsupply.com/milkv/milk-v-pioneer

However, our idea was to evaluate HPX and Kokkos on this new architecture.

diehlpk:
@lucbv The paper got accepted, and here is the [preprint](https://arxiv.org/abs/2309.06530)

diehlpk:
@ajpowelsnl I finished traveling and will now have time to work on the pull request.

diehlpk:
It took a while but here is the promised [pull request ](https://github.com/kokkos/kokkos/pull/6771).

diehlpk:
This can be closed since the pull request is merged.

PR #6771, opened 2024-01-31
Add suport for RISCV and the Milk-V's Pioneer

diehlpk:
This pull request adds support to compile Kokkos on RISC-V for the [Milk-V's Pioneer](https://www.hackster.io/news/milk-v-s-pioneer-is-a-tall-glass-of-cold-risc-v-with-64-cores-up-to-128gb-of-ram-fabf9f22747f) machine.

One can add `-DKOKKOS_ARCH_RISC64_rv64imafdcv=ON` and Kokkos builds on the machine.

I am open to discussing a better name for `RISC64_rv64imafdcv` since I am not quite happy with it.


diehlpk:
refs #6323

dalg24 (changes requested):
We do not take PRs against master.
Please rebase your PR onto develop and edit the destination branch.

cmake/kokkos_arch.cmake
```diff
@@ -378,6 +378,17 @@ IF (KOKKOS_ARCH_HSW)
...
 ENDIF()
 
+IF (KOKKOS_ARCH_RISC64_rv64imafdcv)
+  COMPILER_SPECIFIC_FLAGS(
+    COMPILER_ID KOKKOS_CXX_HOST_COMPILER_ID
+        Cray    NO-VALUE-SPECIFIED
+        Intel   NO-VALUE-SPECIFIED
+        MSVC    NO-VALUE-SPECIFIED
+        NVHPC   NO-VALUE-SPECIFIED
+        DEFAULT -march=rv64imafdcv
```
dalg24, line 388: You require to use whitespaces instead of tabs.
See https://cloud.cees.ornl.gov/jenkins-ci/job/Kokkos/job/PR-6771/1/pipeline-console/log?nodeId=25

cmake/kokkos_arch.cmake
```diff
@@ -378,6 +378,17 @@ IF (KOKKOS_ARCH_HSW)
   )
 ENDIF()
 
+IF (KOKKOS_ARCH_RISC64_rv64imafdcv)
```
dalg24, line 381: Why is the suffix in lowercase?  We generally use uppercase for these.
Also you did not actually register a `KOKKOS_ARCH_OPTION()` for it.

core/src/impl/Kokkos_Core.cpp
```diff
@@ -683,6 +683,8 @@ void pre_initialize_internal(const Kokkos::InitializationSettings& settings) {
 #elif defined(KOKKOS_ARCH_AMD_ZEN3)
   declare_configuration_metadata("architecture", "CPU architecture",
                                  "AMD_ZEN3");
+#elif defined(KOKKOS_ARCH_RISC64_rv64imafdcv)
```
dalg24, line 686: You would need to add a `#cmakedefine` in `cmake/KokkosCore_config.h.in` for this macro to be defined.

cmake/kokkos_arch.cmake
lucbv, line 381: I do not like the prefix `KOKKOS_ARCH_RISC64`, I think `KOKKOS_ARCH_RISCV` is better considering that other flavors of risc exist out there that are unrelated to RISC-V.

masterleinad (commented):
For this to actually be visible in `CMake` you need to add a line such as
```
DECLARE_AND_CHECK_HOST_ARCH(RISC64_rv64imafdcv               "Milk-V (RISC-V) CPUs")
```
would it make sense to just call it `MILKV` instead of `RISC64_rv64imafdcv`? How specialized is this architecture? Could it be used more broadly than in `Milk-V`?

lucbv:
`RISC64_rv64imafdcv` is not an actual architecture, it's a set of RISC-V extensions that are required by Milk-V. I agree with @masterleinad that it is a bit too prone to user error with the given name!

cmake/kokkos_arch.cmake
diehlpk: Ok, I can change that.

core/src/impl/Kokkos_Core.cpp
diehlpk: Ok, I added the define

cmake/kokkos_arch.cmake
diehlpk: No, I did not register the KOKKOS_ARCH_OPTION() yet.

diehlpk:
> We do not take PRs against master. Please rebase your PR onto develop and edit the destination branch.

Rebasing my PR onto develop with the current master as the base has too many conflicts. I will close this pull request and do a new one.

diehlpk:
New pull request is #6773

PR #4349, opened 2021-09-23
Fence after DeepCopy without execution space

masterleinad:
We discussed on `slack` that fencing should happen at the call side of `DeepCopy`. 
edit: In particular, there is no need to call synchronous variants in the `DeepCopy` specializations since the overloads with overloads that take an (accessible) execution space are supposed to be asynchronous anyway (e.g. such that the user is responsible for the correct fences when using the three-argument `deep_copy`) and the `DeepCopy` that doesn't take an execution space is supposed to be fenced internally at the call site.

PR #6772, opened 2024-01-31
Change the default execution policy behavior of the OpenACC backend

seyonglee:
Change the default execution policy behavior of the OpenACC backend from synchronous to asynchronous executions.

- Change the default OpenACC async_arg value from acc_async_sync to acc_async_noval.
- Add acc_wait(async_arg) to scalar reduction operations (parallel_reduce()).

### DIFF
--- a/core/src/OpenACC/Kokkos_OpenACC_DeepCopy.hpp
+++ b/core/src/OpenACC/Kokkos_OpenACC_DeepCopy.hpp
@@ -34,7 +34,7 @@ struct Kokkos::Impl::DeepCopy<Kokkos::Experimental::OpenACCSpace,
     // value checking is added as a safeguard. (The current NVHPC (V22.5)
     // supports OpenACC V2.7.)
     if (n > 0) {
-      acc_memcpy_device(dst, const_cast<void*>(src), n);
+      acc_memcpy_device_async(dst, const_cast<void*>(src), n, acc_async_noval);
     }
   }
   DeepCopy(const Kokkos::Experimental::OpenACC& exec, void* dst,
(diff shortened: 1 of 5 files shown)


core/src/OpenACC/Kokkos_OpenACC_Instance.hpp
```diff
@@ -35,15 +35,15 @@ class OpenACCInternal {
 
  public:
   static int m_acc_device_num;
-  int m_async_arg = acc_async_sync;
+  int m_async_arg = acc_async_noval;
```
masterleinad, line 38: What's the difference between `acc_async_noval` and `acc_async_sync`?

masterleinad (commented):
Are kernels still executed in the same order they were submitted (and non-overlapping) with this changes?

core/src/OpenACC/Kokkos_OpenACC_Instance.hpp
seyonglee: `acc_async_novel` means that the kernels are submitted to a default asynchronous queue, and `acc_async_sync` means that kernels are submitted to a default synchronous queue.

masterleinad (approved):

crtrott (approved):
I think this is ok. We could possibly add a comment in the places where `acc_async_noval` is used directly that these are places where execution space instances (as a representation of the queue) may need to be passed in.

core/src/OpenACC/Kokkos_OpenACC_DeepCopy.hpp
```diff
@@ -34,7 +34,7 @@ struct Kokkos::Impl::DeepCopy<Kokkos::Experimental::OpenACCSpace,
     // value checking is added as a safeguard. (The current NVHPC (V22.5)
     // supports OpenACC V2.7.)
     if (n > 0) {
-      acc_memcpy_device(dst, const_cast<void*>(src), n);
+      acc_memcpy_device_async(dst, const_cast<void*>(src), n, acc_async_noval);
```
dalg24, line 37: This overload of `DeepCopy`, without the leading execution space argument, is meant to be synchronous IIRC.

masterleinad: We have fences at the call side, also see https://github.com/kokkos/kokkos/pull/4349.

dalg24:
The OpenACC build passed

PR #6562, opened 2023-10-31
SYCL: Improve team_reduce implementation

masterleinad:
Related to https://github.com/kokkos/kokkos/pull/5334#issuecomment-1787497628.
#5334 exposed that `team_reduce` in `SYCL` doesn't use arbitrarily large types in general.
The current implementation relies on using some fixed-size local memory (that is also used for `team_scan`) since the SYCL2020 standard doesn't support requesting local memory at function scope. That means that the best we could do was to allocate some fixed-size local memory when launching the kernel but that allocation might not be big enough.
`intel/llvm` proposes an extension that allows this, see https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/supported/sycl_ext_oneapi_local_memory.asciidoc.
With this extension, we can implement `team_reduce` in a similar way as we do for `Cuda` and `HIP`. The relevant `Cuda` backend function is `cuda_inter_warp_reduction`: https://github.com/kokkos/kokkos/blob/54f2e7f23dff4dcd4675e6dcf35106a46a063381/core/src/Cuda/Kokkos_Cuda_ReduceScan.hpp#L61-L92

The strategy is to first do a shuffle reduction within warps, then load warp results in chunks of 4 in shared memory and do the final reduction for all threads redundantly. This also what we now do with this pull request. Note that it's not clear to me whether using a chunk size of 8 is optimal. We have a maximum of 1024 threads per workgroup which means 32 subgroups for subgroup size 32 and 64 for subgroup size 16. That means that loading all subgroup results into local memory might take up to 32/8=4, or 64/8=8 rounds.
Also, note that we are using a chunk size of 8 for HIP (max 1024 threads, 64 threads per wavefront, max 16 wavefronts, maximum of 16/8=2 rounds) and 4 for CUDA (max 1024 threads, 32 threads per warp, max 32 warps, maximum of 32/4=8 rounds).

PR #6779, opened 2024-02-01
SYCL: Clean up TeamPolicy implementation

masterleinad:
This pull request is based on #6562 and avoids using `m_team_reduce`. The advantage is that we don't need to explicitly allocate memory for `TeamPolicy` but only when needed. The problem with that is that we can't know for sure how much shared memory `Kokkos` uses internally which makes calculating the maximum team size difficult (if limited by the shared memory). The current approach just doesn't modify those functions.

This should be applicable to `HIP` and `Cuda` as well with the advantage of being able to introspect the functor for implicit shared memory being used.


dalg24:
Are you still pursuing this?

masterleinad:
This would lielly create more problems than it wopuld solve for the SYCL backend.

PR #6782, opened 2024-02-02
multi-GPU support: Add test for all policies

masterleinad:
This should be the last step for #6091. In particular, this pull request extends the test in `TestCuda_interOp_StreamsMultiGPU.cpp` to test all policies by effectively copying the test in `TestCuda_InterOp_Streams.cpp` and running all the kernels on the default device and the last visible device (as reported by `cudaGetDevciceCount`) using managed and unmanaged `View`s.

Future steps are:
- arbitrary size atomics, see https://github.com/desul/desul/pull/110 and https://github.com/kokkos/kokkos/pull/6469.
- ~~managed Views (requires storing the device id in `SharedAllocationRecord`).~~

### DIFF
--- a/core/unit_test/CMakeLists.txt
+++ b/core/unit_test/CMakeLists.txt
@@ -787,7 +787,7 @@ if(Kokkos_ENABLE_CUDA)
   KOKKOS_ADD_EXECUTABLE_AND_TEST(
     CoreUnitTest_CudaInterOpStreamsMultiGPU
     SOURCES
-      UnitTestMain.cpp
+      UnitTestMainInit.cpp
       cuda/TestCuda_InterOp_StreamsMultiGPU.cpp
   )
   KOKKOS_ADD_EXECUTABLE_AND_TEST(
--- a/core/unit_test/cuda/TestCuda_InterOp_StreamsMultiGPU.cpp
+++ b/core/unit_test/cuda/TestCuda_InterOp_StreamsMultiGPU.cpp
@@ -48,6 +48,8 @@ void test_policies(TEST_EXECSPACE exec0, Kokkos::View<int *, TEST_EXECSPACE> v0,
   Kokkos::deep_copy(exec, v, 5);
   Kokkos::deep_copy(exec0, v0, 5);
 
+  Kokkos::deep_copy(v, v0);
+
   int sum;
   int sum0;
 
--- a/core/unit_test/cuda/TestCuda_InterOp_StreamsMultiGPU.cpp
+++ b/core/unit_test/cuda/TestCuda_InterOp_StreamsMultiGPU.cpp
@@ -16,28 +16,47 @@
 
 #include <TestCuda_Category.hpp>
 #include <Test_InterOp_Streams.hpp>
+#include <memory>
 
 namespace {
 
-std::array<TEST_EXECSPACE, 2> get_execution_spaces(int n_devices) {
-  KOKKOS_IMPL_CUDA_SAFE_CALL(cudaSetDevice(0));
-  cudaStream_t stream0;
-  KOKKOS_IMPL_CUDA_SAFE_CALL(cudaStreamCreate(&stream0));
+struct StreamsAndDevices {
+  std::shared_ptr<std::array<cudaStream_t, 2>> streams;
+  std::array<int, 2> devices;
 
-  KOKKOS_IMPL_CUDA_SAFE_CALL(cudaSetDevice(n_devices - 1));
-  cudaStream_t stream;
-  KOKKOS_IMPL_CUDA_SAFE_CALL(cudaStreamCreate(&stream));
+  StreamsAndDevices() {
+    int n_devices;
+    KOKKOS_IMPL_CUDA_SAFE_CALL(cudaGetDeviceCount(&n_devices));
 
-  TEST_EXECSPACE exec0(stream0);
-  TEST_EXECSPACE exec(stream);
+    devices = {0, n_devices - 1};
+
+    streams = std::shared_ptr<std::array<cudaStream_t, 2>>(
+        new std::array<cudaStream_t, 2>{},
+        [this](std::array<cudaStream_t, 2> *s) {
+          for (int i = 0; i < 2; ++i) {
+            KOKKOS_IMPL_CUDA_SAFE_CALL(cudaSetDevice(devices[i]));
+            KOKKOS_IMPL_CUDA_SAFE_CALL(cudaStreamDestroy((*s)[i]));
+          }
+        });
+    for (int i = 0; i < 2; ++i) {
+      KOKKOS_IMPL_CUDA_SAFE_CALL(cudaSetDevice(devices[i]));
+      KOKKOS_IMPL_CUDA_SAFE_CALL(cudaStreamCreate(&((*streams)[i])));
+    }
+  }
+};
+
+std::array<TEST_EXECSPACE, 2> get_execution_spaces(
+    const StreamsAndDevices &streams_and_devices) {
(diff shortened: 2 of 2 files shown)


masterleinad:
@tcclevenger Can you please have a look?

core/unit_test/cuda/TestCuda_InterOp_StreamsMultiGPU.cpp
```diff
@@ -32,14 +28,139 @@ TEST(cuda, multi_gpu) {
...
+  int n_devices;
+  KOKKOS_IMPL_CUDA_SAFE_CALL(cudaGetDeviceCount(&n_devices));
   {
-    TEST_EXECSPACE space0(stream0);
-    ASSERT_EQ(space0.cuda_device(), 0);
-    TEST_EXECSPACE space(stream);
-    ASSERT_EQ(space.cuda_device(), n_devices - 1);
+    std::array<TEST_EXECSPACE, 2> execs = get_execution_spaces(n_devices);
+
+    ASSERT_EQ(execs[0].cuda_device(), 0);
```
tcclevenger, line 116: These checks are repeated below. Would it make sense just to add 
```
ASSERT_EQ(exec0.cuda_device(), 0);
ASSERT_EQ(exec.cuda_device(), ndev-1);
```
after creating the spaces in `get_execution_spaces()`?

core/unit_test/CMakeLists.txt
```diff
@@ -787,7 +787,7 @@ if(Kokkos_ENABLE_CUDA)
   KOKKOS_ADD_EXECUTABLE_AND_TEST(
     CoreUnitTest_CudaInterOpStreamsMultiGPU
     SOURCES
-      UnitTestMain.cpp
+      UnitTestMainInit.cpp
```
tcclevenger, line 790: What is this change for?

tcclevenger (commented):
Maybe also testing `deep_copy` to and from host makes sense to add? I also can't remember, but is deep-copying between devices possible yet?

core/unit_test/CMakeLists.txt
masterleinad: This ensures that `Kokkos` is initialized before any tests are run and finalized after. The file used previously leaves that to the individual tests (which means that only one test case can be used since `Kokkos` can only be initialized and finalized once).

masterleinad:
> I also can't remember, but is deep-copying between devices possible yet?

See 49a7ea6

tcclevenger (approved):

core/unit_test/cuda/TestCuda_InterOp_StreamsMultiGPU.cpp
```diff
@@ -32,14 +28,141 @@ TEST(cuda, multi_gpu) {
...
+    test_policies(execs[0], view0, execs[1], view);
+    stream0 = execs[0].cuda_stream();
+    stream  = execs[1].cuda_stream();
   }
-  Kokkos::finalize();
+  KOKKOS_IMPL_CUDA_SAFE_CALL(cudaSetDevice(0));
+  KOKKOS_IMPL_CUDA_SAFE_CALL(cudaStreamDestroy(stream0));
+
+  KOKKOS_IMPL_CUDA_SAFE_CALL(cudaSetDevice(n_devices - 1));
+  KOKKOS_IMPL_CUDA_SAFE_CALL(cudaStreamDestroy(stream));
```
dalg24, line 137: It is good that you attempted to reuse code to create the exec spaces but the cleanup is still messy.

core/unit_test/cuda/TestCuda_InterOp_StreamsMultiGPU.cpp
```diff
@@ -16,34 +16,158 @@
...
+#include <memory>
 
 namespace {
-TEST(cuda, multi_gpu) {
-  Kokkos::initialize();
 
-  int n_devices;
-  KOKKOS_IMPL_CUDA_SAFE_CALL(cudaGetDeviceCount(&n_devices));
+struct StreamsAndDevices {
+  std::shared_ptr<std::array<cudaStream_t, 2>> streams;
```
dalg24, line 24: Why did you go for a `shared_ptr` rather than other alternatives?
* `unique_ptr`
* Move cleanup logic to the class destructor and delete copy constructor and copy assignment operator
* GTest's native support for "fixtures"

I am not necessarily blocking but the approach looks weird to me

masterleinad: - It was more convenient to use `shared_ptr` than `unique_ptr` with a custom deleter.
- could do class destructor + delete special member functions
- don't like messing with gtest fixtures

dalg24: > * It was more convenient to use `shared_ptr` than `unique_ptr` with a custom deleter.

How so?  I am not very familiar with the differences.

> * don't like messing with gtest fixtures

Why not?  My understanding is that it was precisely developed for that use case you have.

masterleinad: > How so? I am not very familiar with the differences.

For `unique_ptr` the deleter type is part of the type but not for `shared_ptr`. Also, I don't have to care about using special member functions with a `shared_ptr` member.

> Why not? My understanding is that it was precisely developed for that use case you have.

I rather start from a more clean slate for the two unit tests.

In the end, these are all small differences. If you are unhappy with this approach, I can do whatever. I just don't think that would change/improve the quality of the unit test much.

dalg24: If you don't want to use GTest fixture, then please move the cleanup logic to the destructor and make sure to make it non-copyable.

masterleinad: See 6440a25

tcclevenger:
Writing some use cases based on these tests, I made the following mistake
```
space1 = Kokkos::Cuda(stream on device 1)
v1 = some view on space1

// Error
Kokkos::deep_copy(v1, 0);

// No error
Kokkos::deep_copy(space1, v1, 0);
```
The error message
> terminate called after throwing an instance of 'std::runtime_error'
  what():  (exec_space_instance.impl_internal_space_instance() ->cuda_memset_async_wrapper( dst.data(), 0, dst.size() * sizeof(typename View<T, P...>::value_type))) error( cudaErrorInvalidValue): invalid argument /home/tccleve/Kokkos/install/include/Cuda/Kokkos_Cuda_ZeroMemset.hpp:34

Is this expected that `deep_copy()` cannot correctly copy to non-default device without passing an exespace? 

If this is the case, it would be nice to have a more descriptive error message, but wanted to make sure this is not a bug before starting on that. I imagine that could become a common mistake for the user..

masterleinad:
> Is this expected that `deep_copy()` cannot correctly copy to non-default device without passing an exespace?

Anything that doesn't have an execution space instance is assumed to be executed on the default device. AFAICT, the problem here is that `Kokkos::deep_copy(v1, 0);` calls `cudaMemsetAsync` with the default execution space instance and that instance doesn't have access to `v1`'s allocation. `cudaMemcpy[Async]` doesn't have that problem and that's why the `deep_copy` in the tests in this pull request works.

In general, we need to teach users to provide an execution space instance everywhere when using multiple devices and the most prominent place is probably the `View` constructor.

We are storing the memory space instance used for allocating the `View` so we could use that to detect problems when using the wrong device for a particular `View`.

dalg24:
All CUDA builds passed. Ignoring HIP and SYCL failures.

Issue #6705, opened 2024-01-07
question: allow `local_deep_copy` to work at team and thread level


maartenarnst:
A recurrent pattern in hierarchical parallelism is to copy values from a subview to a scratch pad view. 

Currently, one way of implementing such an operation is by calling `local_deep_copy`. Currently, this function hardcodes the use of `TeamVectorRange`. Thus, it is well-suited to copy to a "team" scratch.

It would be interesting to allow `local_deep_copy` to work with a `ThreadVectorRange` too. It would allow patterns whereby a "thread" scratch is filled. E.g, it would allow something like:

```
parallel_outer(
    TeamPolicy<>(numberOfTeams, teamSize, vectorLength),
    KOKKOS_LAMBDA (const member_type & teamMember [, ...]) {
        parallel_middle(
            TeamThreadRange(teamMember, thisTeamsRangeSize),
            [=] (const int indexWithinBatch [, ...]) {
                const int idx = teamMember.team_size() * teamMember.league_rank() + indexWithinBatch;

                const auto my_subview = Kokkos::subview(my_view, pair(idx * someSize, (idx + 1) * someSize));

                const scratchpad_view_t scratch(teamMember.thread_scratch(0), someSize);

                local_deep_copy_thread(teamMember, scratch, my_subview);

                ...
            } [, ...])
    } [, ...]);
```

One difficulty is that `ThreadVectorRange` is implemented as a function, and so it's not clear whether a nice implementation could be obtained by templating `local_deep_copy` on a policy type. 

A possibility may be to do something like what was done in `Kokkos_NestedSort.hpp`. There would be two functions, one with a `_team` suffix, and one with a `_thread` suffix, and they would call a common implementation.

Do you have suggestions for the best way to proceed? Thanks in advance.

Also tagging @romintomasetti.

masterleinad:
So you basically want to wrap
```
  Kokkos::parallel_for(Kokkos::ThreadVectorRange(team, N),
                       [&](const int& i) { dst(i) = src(i); });
  team.team_barrier();
}
```
?

> One difficulty is that ThreadVectorRange is implemented as a function, and so it's not clear whether a nice implementation could be obtained by templating local_deep_copy on a policy type.

`ThreadVectorRange` returns an `Impl::ThreadVectorRangeBoundariesStruct` object.

crtrott:
Actually we should just make all of these things overloads of deep_copy with the corresponding policy arguments ...

[...]

PR #6783, opened 2024-02-02
Adding `local_deep_copy_thread` for deep copy at threads level

cedricchevalier19:
#6705 asks for the feature, making the use of thread level scratch easier.

@crtrott proposal to be able to call all the deep copy functionalities with `deep_copy` instead of `local_deep_copy` for nested contexts is not implemented in this PR.

Most of the work in this PR is about refactoring local_deep_copy tests. By itself `local_deep_copy_thread` implementation is basically just a naive copy and paste of `local_deep_copy`. We cannot directly template with policies as they are functions that take the iteration space in argument and we need to set it internally (the length of the view).

Few related remarks:
1. I think `local_deep_copy*` functions should be hardened, their only check is if the two views share the same rank. Copying from a smaller view should result in out-of-bounds accesses. Do we have to do a premature return or an assert or do we need a better error handling?

2. `local_deep_copy` decompositions of indices does not take into account the view layout. I am not sure if the smart decomposition always makes sense performance-wise.

3. `local_deep_copy_contiguous` can be related to ~~`std::copy`~~ `Kokkos::Experimental::copy` from algorithm

### DIFF
--- a/core/src/Kokkos_CopyViews.hpp
+++ b/core/src/Kokkos_CopyViews.hpp
@@ -1791,19 +1791,31 @@ namespace Experimental {
 /** \brief  A local deep copy between views of the default specialization,
  * compatible type, same non-zero rank.
  */
-template <class TeamType, class DT, class... DP, class ST, class... SP>
-void KOKKOS_INLINE_FUNCTION local_deep_copy_contiguous(
-    const TeamType& team, const View<DT, DP...>& dst,
-    const View<ST, SP...>& src) {
+template <class TeamType, class iType, class MemberType, class DT, class... DP,
+          class ST, class... SP>
+void KOKKOS_INLINE_FUNCTION deep_copy_contiguous(
+    const TeamType& team,
+    Kokkos::Impl::ThreadVectorRangeBoundariesStruct<iType, MemberType>,
+    const View<DT, DP...>& dst, const View<ST, SP...>& src) {
+  Kokkos::parallel_for(Kokkos::ThreadVectorRange(team, src.span()),
+                       [&](const int& i) { dst.data()[i] = src.data()[i]; });
+}
+//----------------------------------------------------------------------------
+template <class TeamType, class iType, class MemberType, class DT, class... DP,
+          class ST, class... SP>
+void KOKKOS_INLINE_FUNCTION deep_copy_contiguous(
+    const TeamType& team,
+    Kokkos::Impl::TeamVectorRangeBoundariesStruct<iType, MemberType>,
+    const View<DT, DP...>& dst, const View<ST, SP...>& src) {
   Kokkos::parallel_for(Kokkos::TeamVectorRange(team, src.span()),
                        [&](const int& i) { dst.data()[i] = src.data()[i]; });
 }
 //----------------------------------------------------------------------------
 template <class TeamType, class DT, class... DP, class ST, class... SP>
-void KOKKOS_INLINE_FUNCTION local_deep_copy_contiguous_thread(
-    const TeamType& team, const View<DT, DP...>& dst,
-    const View<ST, SP...>& src) {
-  Kokkos::parallel_for(Kokkos::ThreadVectorRange(team, src.span()),
+void KOKKOS_INLINE_FUNCTION
+local_deep_copy_contiguous(const TeamType& team, const View<DT, DP...>& dst,
+                           const View<ST, SP...>& src) {
+  Kokkos::parallel_for(Kokkos::TeamVectorRange(team, src.span()),
                        [&](const int& i) { dst.data()[i] = src.data()[i]; });
 }
 //----------------------------------------------------------------------------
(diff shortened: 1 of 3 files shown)


core/src/Kokkos_CopyViews.hpp
```diff
@@ -1984,6 +2146,41 @@ void KOKKOS_INLINE_FUNCTION local_deep_copy(
...
+      int i2                          = itmp % dst.extent(2);
+      itmp                            = itmp / dst.extent(2);
+      int i3                          = itmp % dst.extent(3);
+      itmp                            = itmp / dst.extent(3);
+      int i4                          = itmp % dst.extent(4);
+      itmp                            = itmp / dst.extent(4);
+      int i5                          = itmp % dst.extent(5);
+      int i6                          = itmp / dst.extent(5);
+      dst(i0, i1, i2, i3, i4, i5, i6) = src(i0, i1, i2, i3, i4, i5, i6);
+    });
```
masterleinad, line 2179: Is there a good reason not to use `ThreadVectorMDRange`?

core/unit_test/TestLocalDeepCopy.hpp
```diff
@@ -14,894 +14,272 @@
...
+  Kokkos::parallel_reduce(
+      "Check B", B.span(),
+      KOKKOS_LAMBDA(int i, double& lsum) { lsum += B.data()[i]; },
+      Kokkos::Sum<double>(sum_all));
 
-  double sum_all = 0.0;
-  for (size_t i = 0; i < B.span(); i++) {
-    sum_all += h_B.data()[i];
+  auto correct_sum = fill_value;
+  for (auto i = 0; i < ViewType::rank; i++) {
```
masterleinad, line 136: You are getting warnings about comparing signed and unsigned types.
```suggestion
  for (decltype(ViewType::rank) i = 0; i < ViewType::rank; ++i) {
```

core/src/Kokkos_CopyViews.hpp
cedricchevalier19: For this example, I think `MDRange` only goes up to rank 6, and `local_deep_copy` was already implemented this way.

cedricchevalier19:
> You can't use `std::copy` in device code for all backends AFAIK. `memcpy` would be usable but only makes sense of the types match and are trivially copyable.

I wanted to talk about `Kokkos::Experimental::copy` of (std) algorithms.

cedricchevalier19:
> Why can't we template on the type returned by TeamThreadRange, TeamVectorRange, and ThreadVectorRange? That might require storing the team handle but that should be about it.

The call for `*VectorRange` is inside `local_deep_copy` because it requires the size of the iteration space. The user can know the kind of policy he wants inside `local_deep_copy` but I do not know how to pass it without having to set the number of elements to loop on.

core/src/Kokkos_CopyViews.hpp
masterleinad: Right. That doesn't mean that we can't try to do better than the current implementation. 🙂

masterleinad:
> I wanted to talk about `Kokkos::Experimental::copy` of (std) algorithms.

Ah, that would create a dependency from `core` to `algorithms` which might be problematic.

> The call for *VectorRange is inside local_deep_copy because it requires the size of the iteration space. The user can know the kind of policy he wants inside local_deep_copy but I do not know how to pass it without having to set the number of elements to loop on.

I was thinking about
```C++
template <class iType, class MemberType, class DT, class... DP, class ST, class... SP>
void KOKKOS_INLINE_FUNCTION deep_copy(
    Impl::ThreadVectorRangeBoundariesStruct<iType, MemberType> member,
    const View<DT, DP...>& dst,
    const View<ST, SP...>& src) {
  Kokkos::parallel_for(memberKokkos::ThreadVectorRange(member.team, src.span()),
                       [&](const int& i) { dst.data()[i] = src.data()[i]; });
}
```
or similar ignoring the range stored in `ThreadVectorRangeBoundariesStruct`.

core/src/Kokkos_CopyViews.hpp
```diff
@@ -1847,7 +1849,9 @@ template <class TeamType, class iType, class MemberType, class DT, class... DP,
           class ST, class... SP>
 void KOKKOS_INLINE_FUNCTION very_deep_copy(
     TeamType team,
-    Kokkos::Impl::TeamThreadRangeBoundariesStruct<iType, MemberType> member,
+    [[maybe_unused]] Kokkos::Impl::TeamThreadRangeBoundariesStruct<iType,
+                                                                   MemberType>
+        member,
```
masterleinad, line 1854:
```suggestion
    Kokkos::Impl::TeamThreadRangeBoundariesStruct<iType,
                                                                   MemberType>
        /*member*/,
```
would be simpler.

cedricchevalier19: You are definitely true. Thanks!

cedricchevalier19:
I think it is simpler to start over and do it after some refactoring of `deep_copy.`

PR #6787, opened 2024-02-05
simd: temporarily skip simd reduction unit test for omptarget build

ldh4:
Related: #6574 

Based on recent CI outputs from few PRs, it seems that the simd reduction unit test is suddenly producing the same non-deterministic error in the same manner that the simd math ops tests were occasionally producing. Temporarily skipping the test for OpenMPTarget build in the interest of clearing the CI, until further triage.


dalg24:
Ignoring HIP failures

dalg24 (approved):

PR #6738, opened 2024-01-23
Untangle `SharedAllocationRecord` spaghetti code

dalg24:
Traded specializations in all. backend for trivial definitions deferring to a generic implementation and explicit instantiations.

PR #6789, opened 2024-02-06
Fix SharedAllocationRecord to allocate using the correct execution space instance

masterleinad:
Since https://github.com/kokkos/kokkos/pull/6738 we are not using the correct execution space instance for allocating anymore. This pull request fixes this behavior by again forwarding the execution space instance argument in the `SharedAllocationCommon` constructor to `checked_allocation_with_header`.

Before that pull request, only some memory spaces implemented `allocate(ExecutionSpace, ...)` because only some of them could actually take advantage of using a particular execution space instance. That's why only some of them forwarded the execution space instance in the `SharedAllocationCommon` constructor.

With the unification of all the `SharedAllocationCommon` implementations in #6738, we now have to implement `allocate(ExecutionSpace,...)` for all memory spaces which is what this pull request does. The ones missing were:
- `CudaUVMSpace`
- `CudaHostPinnedSpace`
- `HIPSpace`
- `HIPHostPinnedSpace`
- `HIPManagedSpace`
- `HostSpace`
- `OpenMPTargetSpace`
- `SYCLSharedUSMSpace`
- `SYCLHostUSMSpace`

Note that I decided to only implement the overload for the single possible execution space instance type if there is only one. All the others are just templated on an arbitrary execution space type.

This pull request is particularly important for multi-GPU support since we would otherwise allocate on the wrong device.

### DIFF
--- a/core/src/impl/Kokkos_SharedAlloc.hpp
+++ b/core/src/impl/Kokkos_SharedAlloc.hpp
@@ -252,14 +252,14 @@ class SharedAllocationRecordCommon : public SharedAllocationRecord<void, void> {
   ~SharedAllocationRecordCommon();
   template <class ExecutionSpace>
   SharedAllocationRecordCommon(
-      ExecutionSpace const&, MemorySpace const& space, std::string const& label,
-      std::size_t alloc_size,
+      ExecutionSpace const& exec, MemorySpace const& space,
+      std::string const& label, std::size_t alloc_size,
       record_base_t::function_type dealloc = &deallocate)
       : SharedAllocationRecord<void, void>(
 #ifdef KOKKOS_ENABLE_DEBUG
             &s_root_record,
 #endif
-            checked_allocation_with_header(space, label, alloc_size),
+            checked_allocation_with_header(exec, space, label, alloc_size),
             sizeof(SharedAllocationHeader) + alloc_size, dealloc, label),
         m_space(space) {
     auto& header = *SharedAllocationRecord<void, void>::m_alloc_ptr;
@@ -315,7 +315,7 @@ class HostInaccessibleSharedAllocationRecordCommon
 #ifdef KOKKOS_ENABLE_DEBUG
             &s_root_record,
 #endif
-            checked_allocation_with_header(space, label, alloc_size),
+            checked_allocation_with_header(exec, space, label, alloc_size),
             sizeof(SharedAllocationHeader) + alloc_size, dealloc, label),
         m_space(space) {
     SharedAllocationHeader header;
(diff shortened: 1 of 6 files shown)


dalg24 (commented):
Please expand the description and explain that this PR is correcting an inconsistency (defect?) in the memory spaces interface, that is, some memory spaces (CudaUVM, CudaHostPinned, complete-list-here) were missing `allocate()` overload taking an execution space.

Adding these let us do the right thing when creating shared allocation records.

masterleinad:
I updated the description.

dalg24 (approved):

cedricchevalier19:
Just to be curious, `Kokkos_HBWSpace` is not concerned because we should not use it?

core/src/impl/Kokkos_SharedAlloc.hpp
```diff
@@ -252,14 +252,14 @@ class SharedAllocationRecordCommon : public SharedAllocationRecord<void, void> {
   ~SharedAllocationRecordCommon();
   template <class ExecutionSpace>
   SharedAllocationRecordCommon(
-      ExecutionSpace const&, MemorySpace const& space, std::string const& label,
-      std::size_t alloc_size,
+      ExecutionSpace const& exec, MemorySpace const& space,
+      std::string const& label, std::size_t alloc_size,
```
tcclevenger, line 256: IMO using `exec_space` and `mem_space` since both are "spaces" would be better. Or `exec` and `mem`. But maybe this is more common in the code than I am aware that `space` implies memory space.

tcclevenger (approved):
Feel free to ignore my comment if there is no other support for the name change. I see a lot of function signatures in `Kokkos_SharedAlloc.hpp` that use `exec_space` and `space`, so there is already precedent for using `space` for memory space when execution space is present.

masterleinad:
> To be fair, we should do either make HBW compile in our CI or drop it. But that's not something for this PR to fix.

`HBWSpace` doesn't work right now. Also, see https://github.com/kokkos/kokkos/pull/4889. At this point, we should probably just drop it.

core/src/impl/Kokkos_SharedAlloc.hpp
masterleinad: `SharedAllocationRecord` is probably the only place where memory space instances are used. If there are any other changes requested, I'm happy to change. Otherwise, I'd prefer to possibly improve in a different pul request so we can make progress with pull requests that depend on this one.

dalg24:
Ignoring runtime failures in HIP builds

PR #6790, opened 2024-02-07
Let `Impl::get_gpu` return a `std::optional<int>`

dalg24:
Motivation: Being able to tell when the user did not specify a device and we did not detect MPI and delegate device selection to the GPU backend.


masterleinad (approved):
Fine with me.

Rombur (approved):

core/src/impl/Kokkos_Core.cpp
```diff
@@ -16,6 +16,7 @@
 
 #ifndef KOKKOS_IMPL_PUBLIC_INCLUDE
 #define KOKKOS_IMPL_PUBLIC_INCLUDE
+#include <optional>
```
dalg24, line 19: Didn't mean to add
```suggestion
```

PR #6797, opened 2024-02-08
Drop `KOKKOS_ENABLE_INTEL_MM_ALLOC` macro

dalg24:
The `KOKKOS_ENABLE_INTEL_MM_ALLOC` macro has no effect since #6341
We should have removed it then.  It was an oversight.

### DIFF
--- a/core/src/impl/Kokkos_Core.cpp
+++ b/core/src/impl/Kokkos_Core.cpp
@@ -568,14 +568,6 @@ void pre_initialize_internal(const Kokkos::InitializationSettings& settings) {
                                  "no");
 #endif
 
-#ifdef KOKKOS_ENABLE_INTEL_MM_ALLOC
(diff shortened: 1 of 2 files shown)


masterleinad (approved):

cedricchevalier19 (commented):
Looks there is also a mention in `Kokkos_HostSpace.hpp` in deprecated section:

https://github.com/kokkos/kokkos/blob/71a9bcae52543bd065522bf3e41b5bfa467d8015/core/src/Kokkos_HostSpace.hpp#L78-L83

dalg24:
> Looks there is also a mention in `Kokkos_HostSpace.hpp` in a deprecated section:

That's correct but that deprecated enumerator has to stay until the next major release.

Rombur (approved):

dalg24:
Ignoring HIP (ROCm 5.6) failures

PR #1440, opened 2018-02-27
Enhance portability by using std::chrono in Timer

wdeconinck:

PR #6798, opened 2024-02-09
Drop librt TPL and associated `KOKKOS_ENABLE_LIBRT` macro

dalg24:
As far as I can tell, last use for it was removed in #1440


masterleinad (approved):
Were you running into any problems on a system that doesn't have `librt`?

Rombur (approved):

dalg24:
> Were you running into any problems on a system that doesn't have `librt`?

No.  As I was removing memkind, I just looked at the list of TPLs and wondered what librt was for.

dalg24:
Ignoring HIP ROCM 5.6 failure.

Christian had no objection.

PR #6802, opened 2024-02-10
simd: allow wider vector width for 32 bit types

ldh4:
Currently in Kokkos SIMD, all simd types use a uniformly set size per simd backend. This size is determined by the size of the largest simd register available in a simd backend divided by 64 (bit). However, 32 bit types can take advantage of this and pack twice as much of data than 64 bit types could in a given simd register.

This PR adds simd types with wider vector for:

- `AVX2`: `float`, `int32_t` (size 8)
- `AVX512`: `float`, `int32_t`, `uint32_t` (size 16)
- `NEON`: `float`, `int32_t` (size 4)

### DIFF
--- a/simd/src/Kokkos_SIMD_AVX2.hpp
+++ b/simd/src/Kokkos_SIMD_AVX2.hpp
@@ -237,7 +237,13 @@ class simd_mask<float, simd_abi::avx2_fixed_size<8>> {
     __m256& m_mask;
     int m_lane;
     KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION __m256 bit_mask() const {
+      // FIXME_HIP ROCm 5.6, 5.7, and 6.0 can't compile with the intrinsic used
+      // here.
+#ifdef KOKKOS_IMPL_WORKAROUND_ROCM_AVX2_ISSUE
+      return _mm256_cvtepi32_ps(_mm256_setr_epi32(
+#else
       return _mm256_castsi256_ps(_mm256_setr_epi32(
+#endif
           -std::int32_t(m_lane == 0), -std::int32_t(m_lane == 1),
           -std::int32_t(m_lane == 2), -std::int32_t(m_lane == 3),
           -std::int32_t(m_lane == 4), -std::int32_t(m_lane == 5),
--- a/simd/src/Kokkos_SIMD_AVX2.hpp
+++ b/simd/src/Kokkos_SIMD_AVX2.hpp
@@ -229,6 +229,100 @@ class simd_mask<float, simd_abi::avx2_fixed_size<4>> {
 };
 
 template <>
+class simd_mask<float, simd_abi::avx2_fixed_size<8>> {
+  __m256 m_value;
+
+ public:
+  class reference {
+    __m256& m_mask;
+    int m_lane;
+    KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION __m256 bit_mask() const {
+      return _mm256_castsi256_ps(_mm256_setr_epi32(
+          -std::int32_t(m_lane == 0), -std::int32_t(m_lane == 1),
+          -std::int32_t(m_lane == 2), -std::int32_t(m_lane == 3),
+          -std::int32_t(m_lane == 4), -std::int32_t(m_lane == 5),
+          -std::int32_t(m_lane == 6), -std::int32_t(m_lane == 7)));
+    }
+
+   public:
+    KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION reference(__m256& mask_arg,
+                                                    int lane_arg)
+        : m_mask(mask_arg), m_lane(lane_arg) {}
+    KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION reference
+    operator=(bool value) const {
+      if (value) {
+        m_mask = _mm256_or_ps(bit_mask(), m_mask);
+      } else {
+        m_mask = _mm256_andnot_ps(bit_mask(), m_mask);
+      }
+      return *this;
+    }
+    KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION operator bool() const {
+      return (_mm256_movemask_ps(m_mask) & (1 << m_lane)) != 0;
+    }
+  };
+  using value_type = bool;
+  using abi_type   = simd_abi::avx2_fixed_size<8>;
+  KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION simd_mask() = default;
+  KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION explicit simd_mask(value_type value)
+      : m_value(_mm256_castsi256_ps(_mm256_set1_epi32(-std::int32_t(value)))) {}
+  template <class G,
+            std::enable_if_t<
+                std::is_invocable_r_v<value_type, G,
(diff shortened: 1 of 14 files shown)


crtrott:
Windows CUDA wasn't passing because of restricitons of NVCC with MSVC. I think you may need to guard the test against Windows + CUDA

ldh4:
Modified the `CMakeList.txt` to prevent simd unit test files from building in Windows+CUDA build.
Let's see if this makes the CI for Windows CUDA build pass.

masterleinad:
Works on my Mac M1 with `ARM_NEON`.

ldh4:
Retest this please.

masterleinad:
The last CI results show
```
clang-16: error: clang frontend command failed with exit code 70 (use -v to see invocation)
AMD clang version 16.0.0 (https://github.com/RadeonOpenCompute/llvm-project roc-5.6.0 23243 be997b2f3651a41597d7a41441fff8ade4ac59ac)
Thread model: posix
InstalledDir: /opt/rocm-5.6.0/llvm/bin
```
for HIP-ROCm-5.6-C++20. Do you need a workaround similar to https://github.com/kokkos/kokkos/pull/6449?

simd/unit_tests/CMakeLists.txt
```diff
@@ -1,7 +1,9 @@
 KOKKOS_INCLUDE_DIRECTORIES(${KOKKOS_SOURCE_DIR}/simd/unit_tests/include)
 
-KOKKOS_ADD_EXECUTABLE_AND_TEST(
-  UnitTest_SIMD
-  SOURCES
-  UnitTestMain.cpp
-  TestSIMD.cpp)
+IF((NOT (Kokkos_ENABLE_CUDA AND WIN32)))
```
masterleinad, line 3: Where does this restriction come from?

simd/unit_tests/include/SIMDTesting_Utilities.hpp
```diff
@@ -181,4 +181,10 @@ class load_as_scalars {
   }
 };
 
+template <typename T, typename = void>
+constexpr bool is_type_v = false;
+
+template <typename T>
+constexpr bool is_type_v<T, decltype(void(sizeof(T)))> = true;
```
masterleinad, line 188: What is this used for? Would you mind adding some comments?

ldh4: This is to loosely check that the type `T` is a complete type.
Because of how `abi_set` and `data_type_set` are currently defined, not all data types can be paired with an abi type with an extended vector width. So, this check is simply used to skip compiling tests for those datatype+abi pairs that are not defined.

ldh4 (commented):
> The last CI results show
> 
> ```
> fatal error: error in backend: Instruction Combining seems stuck in an infinite loop after 1000 iterations.
> clang-16: error: clang frontend command failed with exit code 70 (use -v to see invocation)
> AMD clang version 16.0.0 (https://github.com/RadeonOpenCompute/llvm-project roc-5.6.0 23243 be997b2f3651a41597d7a41441fff8ade4ac59ac)
> Target: x86_64-unknown-linux-gnu
> Thread model: posix
> InstalledDir: /opt/rocm-5.6.0/llvm/bin
> ```
> 
> for HIP-ROCm-5.6-C++20. Do you need a workaround similar to #6449?

Possibly, although neither `_mm_maskload_epi32` nor `_mm256_maskload_epi64` was used in this PR. I'm not sure which intrinsics are causing this issue, but I'll see if the same error occurs with a rebase.

ldh4:
It seems like rocm 5.6-6.0 can't compile `_mm256_castsi256_ps` when used outside of constructors. Applying the same treatment as #6449 and converted to use `_mm256_cvtepi32_ps` instead for rocm 5.6-6.0 builds.

simd/unit_tests/CMakeLists.txt
masterleinad: Fine with me.

simd/unit_tests/include/SIMDTesting_Utilities.hpp
masterleinad: Would you mind adding a comment to that effect?

simd/src/Kokkos_SIMD_AVX2.hpp
```diff
@@ -1045,29 +1248,30 @@ class simd<std::int32_t, simd_abi::avx2_fixed_size<4>> {
   KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION simd& operator=(simd const&) = default;
   KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION simd& operator=(simd&&) = default;
   KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION static constexpr std::size_t size() {
-    return 4;
+    return 8;
   }
-  template <class U, std::enable_if_t<std::is_convertible_v<U, value_type>,
-                                      bool> = false>
+  template <typename U, std::enable_if_t<std::is_convertible_v<U, value_type>,
```
masterleinad, line 1253:
```suggestion
  template <class U, std::enable_if_t<std::is_convertible_v<U, value_type>,
```
Looks like we are almost exclusively using `class U` so far.

masterleinad (commented):
Apart from https://github.com/kokkos/kokkos/pull/6802#discussion_r1607306815, this looks OK to me skimming through the implementation details. I haven't checked that all the intrinsic used are actually correct, though.

simd/unit_tests/include/SIMDTesting_Utilities.hpp
ldh4: Added a comment explaining the use case of these structs.

simd/src/Kokkos_SIMD_AVX2.hpp
ldh4: Changed to `class U`.

simd/src/Kokkos_SIMD_AVX512.hpp
```diff
@@ -944,12 +1332,230 @@ class simd<std::uint32_t, simd_abi::avx512_fixed_size<8>> {
...
+      _mm512_cvtepi32_ps(static_cast<__m512i>(a)));
+}
+
+[[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION Experimental::simd<
+    float, Experimental::simd_abi::avx512_fixed_size<16>>
+ceil(Experimental::simd<
+     std::int32_t, Experimental::simd_abi::avx512_fixed_size<16>> const& a) {
+  return Experimental::simd<float,
+                            Experimental::simd_abi::avx512_fixed_size<16>>(
+      _mm512_cvtepi32_ps(static_cast<__m512i>(a)));
```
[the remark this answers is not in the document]
ldh4: Because the value type here is `int32_t`, no real rounding operation is necessary.
But since the returning simd value type is expected to be `float`, a simple conversion from `int32_t` to `float` is done for these rounding operations.

Rombur (approved):

ldh4:
Retest this please.

ndellingwood:
@ldh4 could you add an entry for the 4.4 changelog [issue](https://github.com/kokkos/kokkos/issues/6914) please?

ldh4:
@ndellingwood Added to the changelog.

Issue #6775, opened 2024-01-31
Nightly test failure, Intel/19.0.5: Kokkos_ViewMapping.hpp(3053): error: no instance of constructor "ZeroMemset" matches the argument list


ndellingwood:
**Describe the bug**
Nightly intel/19.0.5 builds fail to compile (serial, openmp, threads backends) with output:
```
/home/jenkins/node/workspace/kokkos_osrhel7_intel_1905_openmp_pthread_serial/kokkos/core/src/impl/Kokkos_ViewMapping.hpp(3053): error: no instance of constructor "ZeroMemset" matches the argument list
            argument types are: (Kokkos::HostSpace::execution_space, Kokkos::View<int *, Kokkos::HostSpace::memory_space, Kokkos::MemoryTraits<1U>>, int)
            space,
            ^
/home/jenkins/node/workspace/kokkos_osrhel7_intel_1905_openmp_pthread_serial/kokkos/core/src/Kokkos_CopyViews.hpp(1338): note: this candidate was rejected because mismatch in count of arguments
  struct ZeroMemset {
         ^
/home/jenkins/node/workspace/kokkos_osrhel7_intel_1905_openmp_pthread_serial/kokkos/core/src/Kokkos_CopyViews.hpp(1339): note: this candidate was rejected because arguments do not match
    ZeroMemset(const ExecutionSpace& exec_space, const ViewType& dst,
    ^
/home/jenkins/node/workspace/kokkos_osrhel7_intel_1905_openmp_pthread_serial/kokkos/core/src/Kokkos_CopyViews.hpp(1344): note: this candidate was rejected because mismatch in count of arguments
    ZeroMemset(const ViewType& dst, typename ViewType::const_value_type& value) {
    ^
          detected during:
            instantiation of "std::enable_if_t<<expression>, void> Kokkos::Impl::ViewValueFunctor<DeviceType, ValueType, true>::construct_shared_allocation() [with DeviceType=Kokkos::Device<Kokkos::HostSpace::execution_space, Kokkos::HostSpace::memory_space>, ValueType=int, Dummy=int]" at line 3442
[...]

PR #6807, opened 2024-02-12
Explicity pass template params to ZeroMemset for intel icpc compilers

ndellingwood:
Avoid apparent CTAD issue with older icpc intel compilers
Resolves issue https://github.com/kokkos/kokkos/issues/6775

### DIFF
--- a/core/src/Kokkos_CopyViews.hpp
+++ b/core/src/Kokkos_CopyViews.hpp
@@ -1359,7 +1359,7 @@ contiguous_fill_or_memset(
       && !std::is_same_v<ExecutionSpace, Kokkos::OpenMP>
 #endif
   )
-#if defined(KOKKOS_COMPILER_INTEL) && && (KOKKOS_COMPILER_INTEL < 2100)
+#if defined(KOKKOS_COMPILER_INTEL) && &&(KOKKOS_COMPILER_INTEL < 2100)
     // icpc needs extra ctad help to compile
     // See https://github.com/kokkos/kokkos/issues/6775
     ZeroMemset<ExecutionSpace, View<DT, DP...>>(exec_space, dst);
@@ -1397,7 +1397,7 @@ contiguous_fill_or_memset(
 // leading to the significant performance issues
 #ifndef KOKKOS_ARCH_A64FX
   if (Impl::is_zero_byte(value))
-#if defined(KOKKOS_COMPILER_INTEL) && && (KOKKOS_COMPILER_INTEL < 2100)
+#if defined(KOKKOS_COMPILER_INTEL) && &&(KOKKOS_COMPILER_INTEL < 2100)
     // icpc needs extra ctad help to compile
     // See https://github.com/kokkos/kokkos/issues/6775
     ZeroMemset<exec_space_type, ViewType>(exec, dst);
--- a/core/src/Kokkos_CopyViews.hpp
+++ b/core/src/Kokkos_CopyViews.hpp
@@ -1359,13 +1359,13 @@ contiguous_fill_or_memset(
       && !std::is_same_v<ExecutionSpace, Kokkos::OpenMP>
 #endif
   )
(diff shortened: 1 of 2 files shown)


ndellingwood:
I made the change to explicitly guard and deal with the older intel icpc ctad issues. If a cleaner change is preferred I can drop the guarded path and just pass the template parameters in all cases

masterleinad (commented):
Are you not running into the same problem in `Kokkos_ViewMapping.hpp`?

ndellingwood:
> Are you not running into the same problem in `Kokkos_ViewMapping.hpp`?

@masterleinad no, the compilation error only occurred with `ZeroMemset` calls in the `Kokkos_CopyViews.hpp` header. I can apply the same changes there for consistency if you'd like

masterleinad:
> @masterleinad no, the compilation error only occurred with `ZeroMemset` calls in the `Kokkos_CopyViews.hpp` header. I can apply the same changes there for consistency if you'd like

I'd prefer that.

dalg24:
I would prefer specifying template parameter only where strictly necessary.

dalg24 (approved):

masterleinad (approved):

ndellingwood:
HIP-ROCM-5.2 job failed in the unit tests (I think the driver or runtime still might have issues?)
SYCL-OneAPI job failed in Shell Script stage, unclear to me what the problem was (but hopefully unrelated to this PR?)
```
Total Test time (real) = 354.77 sec
[Pipeline] }
[Pipeline] // withEnv
[Pipeline] }
$ docker stop --time=1 78849b170a737dedc8e55eb7fefb460603c4ea69e32304603a34498d1ccdf0ac
$ docker rm -f --volumes 78849b170a737dedc8e55eb7fefb460603c4ea69e32304603a34498d1ccdf0ac
[Pipeline] // withDockerContainer
[Pipeline] }
[Pipeline] // withEnv
[Pipeline] }
[Pipeline] // node
[Pipeline] }
[Pipeline] // stage
[Pipeline] }
[Pipeline] // parallel
[Pipeline] }
[Pipeline] // stage
[Pipeline] }
[Pipeline] // timeout
[Pipeline] }
[Pipeline] // withEnv
[Pipeline] End of Pipeline
ERROR: script returned exit code 2

GitHub has been notified of this commit’s build result

Finished: FAILURE
```

ndellingwood:
@dalg24 @masterleinad I don't think the CI fails are related to this PR, let me know if there are any concerns

dalg24:
I do not agree with the changes requested by Daniel and the last two commits that were pushed afterI approved.  I am waiting on others to weigh in on this.

crtrott:
In particular with all the ifdefing I agree: only do it where strictly necessary.

ndellingwood:
@crtrott @dalg24 @masterleinad how about I revise and just unconditionally explicitly pass the template params to the ZeroMemset calls? The if-guard litter allowed leaving some "documentation" of what was problematic with the older intel/19 compiler in this case, but isn't really necessary

crtrott:
It would still need a comment of why we are passing them explicitly even though they should be deducable. And I would still only do it in the place where it is actually needed.

masterleinad:
> @crtrott @dalg24 @masterleinad how about I revise and just unconditionally explicitly pass the template params to the ZeroMemset calls? The if-guard litter allowed leaving some "documentation" of what was problematic with the older intel/19 compiler in this case, but isn't really necessary

Fine with me if you leave a FIXME explaining why using CTAD doesn't work here so we can change when we require newer compiler versions.

ndellingwood:
Ready for review. Should I squash down to a single commit, or can that be during merge after approvals etc?

crtrott (approved):

dalg24:
> Ready for review. Should I squash down to a single commit, or can that be during merge after approvals etc?

We can do a squash-merge

dalg24:
Ignoring runtimes issues with the 2 HIP builds and the SYCL build that did not start.

PR #6810, opened 2024-02-12
Update workflow permissions to meet OSSF scorecard requirements

crtrott:
This puts the permissions of workflows explicitly to read only


crtrott:
Actually good point: i thought it was that we inherit from an action or so: but more likely scorecard was checking these things only on the default branch ...

So I updated all of them to use what the scorecard thing uses. Lets see if that works.

masterleinad (approved):

lucbv (approved):

PR #6812, opened 2024-02-13
Only copy data in UnorderedMap deep_copy

tcclevenger:
Previously, `deep_copy(UnorderedMap,UnorderedMap)` acted much more similar to `create_mirror_view_and_copy()` which allocated the dst view and then deep_copy the data. Now deep_copy requires the capacity of the src and dst UnorderedMaps are identical and deep copies all data from src to dst.

Also changed
- Deprecated old behavior
- Separate `UnorderedMap::create_copy_view()` into `allocate_view()` and `deep_copy_view()`
- Create a `create_mirror(UnorderedMap)` specialization

### DIFF
--- a/containers/src/Kokkos_UnorderedMap.hpp
+++ b/containers/src/Kokkos_UnorderedMap.hpp
@@ -856,8 +856,20 @@ class UnorderedMap {
                    std::is_same<std::remove_const_t<SValue>, value_type>::value>
   deep_copy_view(
       UnorderedMap<SKey, SValue, SDevice, Hasher, EqualTo> const &src) {
+#ifndef KOKKOS_ENABLE_DEPRECATED_CODE_4
     // To deep copy UnorderedMap, capacity must be identical
     KOKKOS_EXPECTS(capacity() == src.capacity());
+#else
+    if (capacity() != src.capacity()) {
+      allocate_view(src);
+#ifdef KOKKOS_ENABLE_DEPRECATION_WARNINGS
+      Kokkos::Impl::log_warning(
+          "Warning: deep_copy_view() allocating views is deprecated. Must call "
+          "with UnorderedMaps of identical capacity, or use "
+          "create_copy_view().\n");
+#endif
+    }
+#endif
 
     if (m_hash_lists.data() != src.m_hash_lists.data()) {
       Kokkos::deep_copy(m_available_indexes, src.m_available_indexes);
@@ -964,11 +976,7 @@ template <typename DKey, typename DT, typename DDevice, typename SKey,
 inline void deep_copy(
     UnorderedMap<DKey, DT, DDevice, Hasher, EqualTo> &dst,
     const UnorderedMap<SKey, ST, SDevice, Hasher, EqualTo> &src) {
-#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
-  dst.create_copy_view(src);
-#else
   dst.deep_copy_view(src);
-#endif
 }
 
 // Specialization of create_mirror() for an UnorderedMap object.
--- a/containers/src/Kokkos_UnorderedMap.hpp
+++ b/containers/src/Kokkos_UnorderedMap.hpp
@@ -856,8 +856,6 @@ class UnorderedMap {
                    std::is_same<std::remove_const_t<SValue>, value_type>::value>
   deep_copy_view(
       UnorderedMap<SKey, SValue, SDevice, Hasher, EqualTo> const &src) {
-    using src_type = UnorderedMap<SKey, SValue, SDevice, Hasher, EqualTo>;
-
     // To deep copy UnorderedMap, capacity must be identical
     KOKKOS_EXPECTS(capacity() == src.capacity());
 
--- a/containers/src/Kokkos_UnorderedMap.hpp
+++ b/containers/src/Kokkos_UnorderedMap.hpp
@@ -805,56 +805,84 @@ class UnorderedMap {
     return *this;
   }
 
+  // Re-allocate the views of the calling UnorderedMap according to src
+  // capacity, and deep copy the src data.
   template <typename SKey, typename SValue, typename SDevice>
   std::enable_if_t<std::is_same<std::remove_const_t<SKey>, key_type>::value &&
                    std::is_same<std::remove_const_t<SValue>, value_type>::value>
   create_copy_view(
       UnorderedMap<SKey, SValue, SDevice, Hasher, EqualTo> const &src) {
     if (m_hash_lists.data() != src.m_hash_lists.data()) {
-      insertable_map_type tmp;
+      allocate_view(src);
+      deep_copy_view(src);
+    }
+  }
 
-      tmp.m_bounded_insert    = src.m_bounded_insert;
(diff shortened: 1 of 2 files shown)


containers/unit_tests/TestUnorderedMap.hpp
```diff
@@ -424,7 +425,7 @@ TEST(TEST_CATEGORY, UnorderedMap_valid_empty) {
   Map n{};
   n = Map{m.capacity()};
   n.rehash(m.capacity());
-  Kokkos::deep_copy(n, m);
+  n.create_copy_view(m);
   ASSERT_TRUE(m.is_allocated());
   ASSERT_TRUE(n.is_allocated());
```
tcclevenger, line 430: I kept this test as using the same behavior as the old deep_copy.

tcclevenger:
Fails on SYCL and HIP builds appear unrelated. I manually tested container unit tests on MI100 (caraway) and got passes for Serial and HIP tests.

containers/src/Kokkos_UnorderedMap.hpp
```diff
@@ -805,56 +805,82 @@ class UnorderedMap {
...
+
+  // Deep copy view data from src. This requires that the src capacity is
+  // identical to the capacity of the calling UnorderedMap.
+  template <typename SKey, typename SValue, typename SDevice>
+  std::enable_if_t<std::is_same<std::remove_const_t<SKey>, key_type>::value &&
+                   std::is_same<std::remove_const_t<SValue>, value_type>::value>
+  deep_copy_view(
+      UnorderedMap<SKey, SValue, SDevice, Hasher, EqualTo> const &src) {
+    // To deep copy UnorderedMap, capacity must be identical
+    KOKKOS_EXPECTS(capacity() == src.capacity());
```
masterleinad, line 860: If the capacities don't match and deprecated code is enabled maybe call `allocate_view` here and print a deprecation message if deprecation warnings are enabled.

containers/src/Kokkos_UnorderedMap.hpp
```diff
@@ -932,13 +960,29 @@ class UnorderedMap {
...
 template <typename DKey, typename DT, typename DDevice, typename SKey,
           typename ST, typename SDevice, typename Hasher, typename EqualTo>
 inline void deep_copy(
     UnorderedMap<DKey, DT, DDevice, Hasher, EqualTo> &dst,
     const UnorderedMap<SKey, ST, SDevice, Hasher, EqualTo> &src) {
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
   dst.create_copy_view(src);
+#else
+  dst.deep_copy_view(src);
+#endif
```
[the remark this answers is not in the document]
masterleinad: Maybe just call `deep_copy_view` here?

containers/src/Kokkos_UnorderedMap.hpp
```diff
@@ -805,56 +805,82 @@ class UnorderedMap {
...
+    }
+  }
+
+  // Allocate views of the calling UnorderedMap with the same capacity as the
+  // src.
+  template <typename SKey, typename SValue, typename SDevice>
+  std::enable_if_t<std::is_same<std::remove_const_t<SKey>, key_type>::value &&
+                   std::is_same<std::remove_const_t<SValue>, value_type>::value>
+  allocate_view(
+      UnorderedMap<SKey, SValue, SDevice, Hasher, EqualTo> const &src) {
```
masterleinad, line 827: Does this need to be `public`?

containers/src/Kokkos_UnorderedMap.hpp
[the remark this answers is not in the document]
tcclevenger: Sorry, I should have put this comment on the deprecated portion. Should I have a warning message for this chunk:
```
// add some message
#else
#endif
```

containers/src/Kokkos_UnorderedMap.hpp
tcclevenger: I need to use `allocate_view()` in the unit test, but actually I do not think `deep_copy_view()` needs to be public.

containers/src/Kokkos_UnorderedMap.hpp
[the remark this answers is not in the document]
masterleinad: I would only warn if the new behavior would error out.

containers/src/Kokkos_UnorderedMap.hpp
tcclevenger: Never mind, both `allocate_view()` and `deep_copy_view()` need to be public since they are called by `deep_copy()` and `create_mirror()` respectively.

masterleinad: I was suggesting to call `create_copy_view` inside of `deep_copy_view` if necessary and that `deep_copy` would only call `deep_copy_view`. Would that work?

tcclevenger: oh I see now, sorry I was confused. Yeah, that should work.

crtrott (approved):

crtrott:
Retest this please

masterleinad:
Retest this please

tcclevenger:
I think the only fail is a CI issue, not with this PR.

masterleinad:
Only `CUDA-11.0-NVCC-RDC` failed on `fetnat07` with a jenkins problem.

ndellingwood:
@dalg24 @tcclevenger does this merit an entry for the 4.4 changelog? Since it is a change of behavior, including deprecation, it seems notable to add?

dalg24:
Deferring to Christian and Conrad.

tcclevenger:
Yeah, this probably deserves an entry. I'll add one.

containers/src/Kokkos_UnorderedMap.hpp
```diff
@@ -932,13 +970,25 @@ class UnorderedMap {
...
 
-// Specialization of deep_copy for two UnorderedMap objects.
+// Specialization of deep_copy() for two UnorderedMap objects.
 template <typename DKey, typename DT, typename DDevice, typename SKey,
           typename ST, typename SDevice, typename Hasher, typename EqualTo>
 inline void deep_copy(
     UnorderedMap<DKey, DT, DDevice, Hasher, EqualTo> &dst,
     const UnorderedMap<SKey, ST, SDevice, Hasher, EqualTo> &src) {
-  dst.create_copy_view(src);
+  dst.deep_copy_view(src);
```
MalachiTimothyPhillips, line 979: Assuming `KOKKOS_ENABLE_DEPRECATED_CODE_4` is enabled, this is actually a change in behavior and will not work as expected if using the `deep_copy` to also perform the view allocations on the destination `UnorderedMap`.

MalachiTimothyPhillips: Scratch that, I see the `allocate_view` call here: https://github.com/kokkos/kokkos/pull/6812/files#diff-2c8acfc1ac5cbbb4cc3da14e466f42c85e415b540c815bf2630c42ec2ac09d59R864.

PR #6813, opened 2024-02-13
fix constness for views in std algorithms 

fnrizzi:
as pointed out in https://github.com/kokkos/kokkos/pull/6801#discussion_r1488242863, this PR fixes all instances so that a View is always taken by `const & `

### DIFF
--- a/algorithms/src/std_algorithms/Kokkos_CopyIf.hpp
+++ b/algorithms/src/std_algorithms/Kokkos_CopyIf.hpp
@@ -54,7 +54,8 @@ template <
     std::enable_if_t<::Kokkos::is_execution_space_v<ExecutionSpace>, int> = 0>
 auto copy_if(const ExecutionSpace& ex,
              const ::Kokkos::View<DataType1, Properties1...>& source,
-             ::Kokkos::View<DataType2, Properties2...>& dest, Predicate pred) {
+             const ::Kokkos::View<DataType2, Properties2...>& dest,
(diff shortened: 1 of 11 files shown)


algorithms/src/std_algorithms/Kokkos_CopyIf.hpp
```diff
@@ -54,7 +54,8 @@ template <
     std::enable_if_t<::Kokkos::is_execution_space_v<ExecutionSpace>, int> = 0>
 auto copy_if(const ExecutionSpace& ex,
              const ::Kokkos::View<DataType1, Properties1...>& source,
-             ::Kokkos::View<DataType2, Properties2...>& dest, Predicate pred) {
+             const ::Kokkos::View<DataType2, Properties2...>& dest,
+             Predicate pred) {
```
dalg24, line 58: what are the preconditions?  we don't allocate and expect that the destination view has enough storage?

dalg24: we don't resize the destination view to fit

dalg24 (approved):
The discussion about copy_if doesn't need to be addressed here

nliber (approved):

masterleinad (approved):

dalg24:
The doc will need to be updated to reflect this

dalg24:
ORNL's Jenkins CI server is down

PR #6815, opened 2024-02-13
Fix fence in Kokkos::sort when using std::sort

masterleinad:
Fixes the HPX test failures in https://github.com/kokkos/kokkos/pull/6801. We need to fence before using `std::sort` if an execution space instance was provided.

### DIFF
--- a/algorithms/src/sorting/Kokkos_SortPublicAPI.hpp
+++ b/algorithms/src/sorting/Kokkos_SortPublicAPI.hpp
@@ -29,7 +29,7 @@ namespace Kokkos {
 // ---------------------------------------------------------------
 
 template <class ExecutionSpace, class DataType, class... Properties>
-void sort([[maybe_unused]] const ExecutionSpace& exec,
+void sort(const ExecutionSpace& exec,
           const Kokkos::View<DataType, Properties...>& view) {
   // constraints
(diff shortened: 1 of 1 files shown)


algorithms/src/sorting/Kokkos_SortPublicAPI.hpp
```diff
@@ -52,6 +52,7 @@ void sort([[maybe_unused]] const ExecutionSpace& exec,
   }
 
   if constexpr (Impl::better_off_calling_std_sort_v<ExecutionSpace>) {
+    exec.fence("Kokkos::sort without comparator use std::sort");
```
dalg24, line 55: shouldn't that be a `Kokkos::fence()`?

masterleinad: We are only expecting work to be enqueued on the provided execution space instance. It's not clear to me why a global fence would be necessary here.

dalg24: I misread the signature.  Sorry

dalg24 (approved):

nliber (approved):

dalg24:
ORNL's cloud is down but all builds on GitHub passed so I am just going to merge.

PR #6816, opened 2024-02-13
Adding SECURITY.md file

crtrott:
This is based on the one from OSSF Scorecard.

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,12 @@
+# Reporting Security Issues
+
+To report a security issue, please email
+[lebrungrandt@ornl.gov](mailto:lebrungrandt@ornl.gov)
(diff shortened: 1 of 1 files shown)


dalg24 (approved):

dalg24:
Couldn't we use [GitHub's private reporting feature](https://docs.github.com/en/code-security/security-advisories/guidance-on-reporting-and-writing-information-about-vulnerabilities/privately-reporting-a-security-vulnerability) instead?

masterleinad:
My understanding is that this file complements the private reporting feature (and we could point to it).

nliber:
The OSSF scorecard uses a mail alias.  Would we want to go that route instead of hardcoded email addresses for individuals?

PR #6824, opened 2024-02-16
SYCL CI: Avoid setvars.sh

masterleinad:
This PR tries to avoid the SYCL CI issues we are seeing when running `setvars.sh` by setting the necessary environment variables in `Dockerfile` already.

### DIFF
--- a/scripts/docker/Dockerfile.sycl
+++ b/scripts/docker/Dockerfile.sycl
@@ -55,3 +55,12 @@ RUN wget https://registrationcenter-download.intel.com/akdlm/irc_nas/19133/l_one
     chmod +x ./l_oneDPL_p_2022.0.0.25335.sh && \
     ./l_oneDPL_p_2022.0.0.25335.sh -a -s --eula accept && \
     rm l_oneDPL_p_2022.0.0.25335.sh
+
+# clang++
+ENV PATH=/opt/intel/oneapi/compiler/latest/linux/bin-llvm/:$PATH
+# sycl-ls, icpx
+ENV PATH=/opt/intel/oneapi/compiler/latest/linux/bin/:$PATH
+# libsycl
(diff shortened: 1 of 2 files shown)


masterleinad:
Seems to work.

scripts/docker/Dockerfile.sycl
```diff
@@ -55,3 +55,12 @@ RUN wget https://registrationcenter-download.intel.com/akdlm/irc_nas/19133/l_one
...
     rm l_oneDPL_p_2022.0.0.25335.sh
+
+# clang++
+ENV PATH=/opt/intel/oneapi/compiler/latest/linux/bin-llvm/:$PATH
+# sycl-ls, icpx
+ENV PATH=/opt/intel/oneapi/compiler/latest/linux/bin/:$PATH
+# libsycl
+ENV LD_LIBRARY_PATH=/opt/intel/oneapi/compiler/latest/linux/lib:$LD_LIBRARY_PATH
+# libsvml
+ENV LD_LIBRARY_PATH=/opt/intel/oneapi/compiler/latest/linux/compiler/lib/intel64_lin:$LD_LIBRARY_PATH
```
dalg24, line 66: Did you consider doing something like
```
RUN echo 'source /opt/intel/oneapi/setvars.sh --force' > $HOME/.bashrc
```
instead.
Is there still no "official" container we could use?

masterleinad: This is the same approach as in https://github.com/kokkos/ci-containers/blob/main/ubuntu which has been working well and we do the same for, say `CMake` in `Dockerfiles` already.
Last I checked, we would not parse any of the `.bashrc`, `.profile`, etc. files when testing with `jenkins`. An alternative might be messing with ENTRYPOINT but that is way more messy in my opinion.

I'm not aware of an "official" container that would contain the Codeplay plugin.

dalg24 (approved):

crtrott (approved):

crtrott:
SYCL passed in jenkins

PR #6825, opened 2024-02-16
Cleanup KOKKOS_CONFIGURE_CORE

masterleinad:
While verifying https://github.com/kokkos/kokkos/pull/6823/files#r1492824399, I noticed that we have a couple of unused variables and include files. In particular,
- KOKKOS_MEMSPACE_LIST
- KOKKOS_BACKEND_POST_INCLUDE_LIST

never contain anything and we can thus remove `KokkosCore_Config_PostInclude.hpp>` and only print the backends once.


crtrott:
Was the post part maybe something the resilience plugin needed? @nmm0

nmm0 (approved):
These are fine for me, we don't use these in resilience

crtrott (approved):

PR #6831, opened 2024-02-21
Fix scorecard workflow

masterleinad:


crtrott (approved):

dalg24 (approved):

dalg24:
It passed after we retriggered

dalg24 (approved):

masterleinad:
We are still seeing
```
refs/heads/develop not supported with push event.
Only the default branch master is supported.
```
It seems we need to wait until we make `develop` the default branch before we get updates (otherwise only for new releases).

PR #6834, opened 2024-02-21
Fix Kokkos README's FENL link

mhoemmen:
FENL example lives in TrilinosCouplings now.

### DIFF
--- a/example/README
+++ b/example/README
@@ -1,7 +1,7 @@
 This directory contains example application proxies that use different
 parts of Kokkos.  If you are looking for the FENL ("finite element
-nonlinear" solve) example, it has moved into the LinAlg subpackage of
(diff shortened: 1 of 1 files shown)


masterleinad (dismissed):
You have to target `develop`.

masterleinad:
The alternative would be to just delete these lines, of course, since we don't have any control over the future (location) of this example anyway.

cedricchevalier19:
I think FENL example should also be listed in https://github.com/kokkos/kokkos-miniapps if this repo is still active.

dalg24 (approved):

masterleinad (approved):

dalg24:
The format check passed.  Ignoring everything else

Issue #6794, opened 2024-02-08
Nightly test failure, SYCL backend on PVC: kokkosp.parallel_scan_no_fence failure


ndellingwood:
**Describe the bug**
Nightly Sycl builds on PVC arch with intel/oneapi/2023.1.0 are failing in the `kokkosp.parallel_scan_no_fence` sub-test:
```
[ RUN      ] kokkosp.parallel_scan_no_fence
Test failure: encountered unwanted events
BeginFence { "Kokkos::Tools::invoke_kokkosp_callback: Kokkos Profile Tool Fence", 0,10}
EndFence { 10}
BeginFence { "Kokkos::Tools::invoke_kokkosp_callback: Kokkos Profile Tool Fence", 117440512,11}
EndFence { 11}
BeginParallelScan { "dogs", 117440513,12}
BeginFence { "Kokkos::Impl::sycl_deallocate: fence before deallocate", 117440512,13}
EndFence { 13}
BeginFence { "Kokkos::Tools::invoke_kokkosp_callback: Kokkos Profile Tool Fence", 0,14}
EndFence { 14}
BeginFence { "Kokkos::Tools::invoke_kokkosp_callback: Kokkos Profile Tool Fence", 117440512,15}
EndFence { 15}
EndParallelScan { 12}
/home/jenkins/blake-new/workspace/Kokkos_develop_blake_sycl_pv_oneapi_2023_1_0/kokkos/core/unit_test/tools/TestEventCorrectness.hpp:429: Failure
Value of: success
  Actual: false
Expected: true
[  FAILED  ] kokkosp.parallel_scan_no_fence (0 ms)
```

**Merged commits since previous pass**

Fix SharedAllocationRecord to allocate using the correct execution space instance (#6789) — [github](https://jenkins-son.sandia.gov/user/github/) / [githubweb](https://github.com/kokkos/kokkos/commit/391f2d12f2d3e5f896d497a11d0752aa47be7277)

Temporarily disable simd_reduction test for omptarget build — [Dong Hun Lee](https://jenkins-son.sandia.gov/user/donlee/) / [githubweb](https://github.com/kokkos/kokkos/commit/97997807de7f592a0845f96ee0adb4644a57efb2)
[...]

PR #3084, opened 2020-06-08
Move allocation profiling to the allocate/deallocate calls

DavidPoliakoff:
Addresses #3064 . Pushing to catch the inevitable preprocessor branches leading to undefined arguments, will mark ready to review when that's fixed

PR #6835, opened 2024-02-21
Remove MemorySpace::[de]allocate overloads without labels

masterleinad:
I noticed in https://github.com/kokkos/kokkos/issues/6794#issuecomment-1957666389 that we call `MemorySpace::deallocate` without a label only from internal allocations where we can do better.
Thus, this pull request suggests streamlining the interface by removing all overloads for `MemorySpace::[de]allocate` that don't take a label. So far, I only made sure to provide labels for `SYCL` in all places that need them now.


ldh4:
Should these `[de]allocate` functions go through a proper deprecation process? They are technically public functions in a non-impl namespace, available to be called in places external to kokkos.

masterleinad:
> Should these `[de]allocate` functions go through a proper deprecation process? They are technically public functions in a non-impl namespace, available to be called in places external to kokkos.

Yeah, that's up for debate.

ldh4 (approved):
Other than the need for a discussion for a potential deprecation, looks good to me.

masterleinad:
> Other than the need for a discussion for a potential deprecation, looks good to me.

Let me add a commit with the deprecations.

core/src/Cuda/Kokkos_CudaSpace.cpp
```diff
@@ -193,7 +196,8 @@ void *impl_allocate_common(const int device_id,
         KOKKOS_IMPL_CUDA_SAFE_CALL(cudaStreamSynchronize(stream));
       } else {
         Impl::cuda_device_synchronize(
-            "Kokkos::Cuda: backend fence after async malloc");
+            std::string("Kokkos::Cuda: backend fence after async malloc ") +
```
dalg24, line 199: Why did you change here and below every time we were passing a C string?

dalg24 (commented):
Since when do we have these `MemSpace::[de]allocate` overload with a label vs the ones w/o a label?

What is problematic about having both?

Why is requiring a label the right thing to do?

masterleinad:
> Since when do we have these MemSpace::[de]allocate overload with a label vs the ones w/o a label?

The version with labels was introduced in https://github.com/kokkos/kokkos/pull/3084. The one without labels was the only overload before that.

> What is problematic about having both?

There doesn't seem to be a good case where you cannot provide a meaningful label. We require a label for `View` allocations anyway. Not providing a label (possibly by accident, not knowing that the other overload existed) makes debugging [de]allocations events harder (provided the respective backend reports back to Tools events).

> Why is requiring a label the right thing to do?

This pull request is motivated by https://github.com/kokkos/kokkos/issues/6794#issuecomment-1957666389. It's very hard to understand Kokkos Tools output for allocation/deallocation events if the label is omitted. In that issue, we were trying to hunt down a specific deallocation that made a test introspecting Tools events fail (since there was an unexpected fence). With the changes in this pull request, we would have caught the offending deallocation directly.

core/src/Cuda/Kokkos_CudaSpace.cpp
masterleinad: What would be the better way to add to strings? The idea is to report the label for the respective allocation to Tools.

tcclevenger (approved):
I like the change.

crtrott (changes requested):
I am blocking for now. I am not sure I agree with removing the allocate function without label. I am even less sure I agree with removing the deallocate function with label - since it means that a user has to track the label around somewhere. And what would it mean to provide a different label at deallocate than was used on allocate.

ajpowelsnl:
@masterleinad , @crtrott -- on the agenda for discussion

ajpowelsnl:
- **TODO:**
  - Tackle with #6918 post Kokkos-4.4

dalg24:
There was no progress on this for a while and there was no consensus to keep working on this today at the developer meeting.
One broader design question is whether we meant for these [de]allocate member function to be public or if it still and foremost intended as the implementation detail for backends.

PR #6840, opened 2024-02-23
Add benchmarks to CMake

cedricchevalier19:
Small PR that enables building `atomic`, `bytes_and_flops`, `gather` and `policy_performance` benchmarks when `Kokkos_ENABLE_BENCHMARKS` is set.

Small changes are:
- local includes use `"` instead of `<>` to avoid setting `-I.` ,
- renaming some local variables whose names where colliding with others or keyword,
- deleting unneeded function arguments to avoid warnings.


masterleinad (approved):

nmm0 (approved):
Looks good to me!

dalg24:
Seeing
```
fatal error: error in backend: Cannot select: 0x5555645ae330: i64,ch = stacksave 0x5555645ae1e0, core/src/OpenMPTarget/Kokkos_OpenMPTarget_ParallelFor_Team.hpp:150:11 @[ core/src/OpenMPTarget/Kokkos_OpenMPTarget_ParallelFor_Team.hpp:148:1 @[ core/src/OpenMPTarget/Kokkos_OpenMPTarget_ParallelFor_Team.hpp:144:5 ] ]
In function: __omp_offloading_30_4d9e13fa__ZNK6Kokkos4Impl11ParallelForIZN3RunIdLi1ELi1EE3runEiiiiiiiiEUlRKNS0_26OpenMPTargetExecTeamMemberEE_NS_10TeamPolicyIJEEENS_12Experimental12OpenMPTargetEE12execute_implIvEEvv_l144
clang++: error: clang frontend command failed with exit code 70 (use -v to see invocation)
clang version 17.0.3
Thread model: posix
InstalledDir: /opt/llvm/bin
[ 96%] Building CXX object algorithms/unit_tests/CMakeFiles/Kokkos_AlgorithmsUnitTest_StdSet_Team_H.dir/TestStdAlgorithmsTeamRemoveCopy.cpp.o
clang++: note: diagnostic msg: 
********************

PLEASE ATTACH THE FOLLOWING FILES TO THE BUG REPORT:
Preprocessed source(s) and associated run script(s) are located at:
clang++: note: diagnostic msg: /tmp/bench_double-691ca2.cpp
clang++: note: diagnostic msg: /tmp/bench_double-946f6a.cpp
clang++: note: diagnostic msg: /tmp/bench_double-691ca2.sh
clang++: note: diagnostic msg: 

********************
make[2]: *** [benchmarks/bytes_and_flops/CMakeFiles/Kokkos_bytes_and_flops.dir/build.make:76: benchmarks/bytes_and_flops/CMakeFiles/Kokkos_bytes_and_flops.dir/bench_double.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:5285: benchmarks/bytes_and_flops/CMakeFiles/Kokkos_bytes_and_flops.dir/all] Error 2
```
in the OpenMPTarget build

rgayatri23:
@cedricchevalier19 - Can you provide me with your cmake configuration line, I have been using this command
```bash
cmake -DCMAKE_CXX_COMPILER=clang++ -DKokkos_ENABLE_OPENMPTARGET=ON -DCMAKE_BUILD_TYPE=RelWithDebugInfo -DKokkos_ARCH_AMPERE80=ON -DKokkos_ENABLE_BENCHMARKS=ON ../
```

And this leads to ICE even when disabling all benchmarks. I tested this with clang/16 and clang/17.0.6 on NVIDIA A100. 

Just wanted to see if there is any difference in the options used by you.

rgayatri23:
Ok I see what the issue is, any `debug` symbol leads to an ICE. This is something that has been noticed even earlier. So the `Release` version works fine except for 2 benchmarks. I am not able to push to your branch but here is the diff that should resolve the issue

```bash
diff --git a/benchmarks/CMakeLists.txt b/benchmarks/CMakeLists.txt
index 4faeb0a20..43f8ae808 100644
--- a/benchmarks/CMakeLists.txt
+++ b/benchmarks/CMakeLists.txt
@@ -1,7 +1,11 @@
 KOKKOS_ADD_BENCHMARK_DIRECTORIES(atomic)
-KOKKOS_ADD_BENCHMARK_DIRECTORIES(bytes_and_flops)
 KOKKOS_ADD_BENCHMARK_DIRECTORIES(gather)
 KOKKOS_ADD_BENCHMARK_DIRECTORIES(gups)
 KOKKOS_ADD_BENCHMARK_DIRECTORIES(launch_latency)
-KOKKOS_ADD_BENCHMARK_DIRECTORIES(policy_performance)
 KOKKOS_ADD_BENCHMARK_DIRECTORIES(stream)
+
+#FIXME_OPENMPTARGET - These two benchmarks cause ICE. Commenting them for now but a deeper analysis on the cause and a possible fix will follow.
+IF(NOT Kokkos_ENABLE_OPENMPTARGET)
+KOKKOS_ADD_BENCHMARK_DIRECTORIES(policy_performance)
+KOKKOS_ADD_BENCHMARK_DIRECTORIES(bytes_and_flops)
+ENDIF()
```

benchmarks/CMakeLists.txt
```diff
@@ -1,11 +1,22 @@
+if (Kokkos_ENABLE_OPENMPTARGET AND CMAKE_BUILD_TYPE!="Release" AND Kokkos_ENABLE_BENCHMARKS)
```
dalg24, line 1: > `CMAKE_BUILD_TYPE!="Release"`

I don't think that is valid cmake

benchmarks/CMakeLists.txt
```diff
@@ -1,11 +1,22 @@
...
+if (NOT Kokkos_ENABLE_OPENMPTARGET)
     KOKKOS_ADD_BENCHMARK_DIRECTORIES(policy_performance)
     KOKKOS_ADD_BENCHMARK_DIRECTORIES(bytes_and_flops)
-ENDIF()
\ No newline at end of file
+else ()
+    # FIXME_OPENMPTARGET - These two benchmarks cause ICE.
+    # Commenting them for now but a deeper analysis on the cause and a possible fix will follow.
+    if (Kokkos_ENABLE_BENCHMARKS)
+        message(WARNING "OpenMPTarget is enabled: disabling 2 benchmarks to avoid ICE.")
```
dalg24, line 20: I would't raise a warning for this.  I think the previous version was good enough.

benchmarks/CMakeLists.txt
```diff
@@ -1,11 +1,22 @@
...
+    endif ()
+endif ()
```
dalg24, line 22: Add an empty new line

benchmarks/CMakeLists.txt
cedricchevalier19: Ok, I revert the last commit, and rebase to have an updated SYCL test.

masterleinad (approved):

benchmarks/bytes_and_flops/bench_unroll_stride.hpp
```diff
@@ -87,25 +87,25 @@ struct Run<Scalar, UNROLL, STRIDE> {
                     C(n, i, 0) = a1;
 #endif
 #if (UNROLL == 2)
-                    C(n, i, 0) = a1 + a2;
+                    C(n, iter, 0) = a1 + a2;
```
dalg24, line 90: What about L87?
Also that does not look right to me.  It gives out-of-bounds accesses if the number of iterations exceeds the view extent in this dimension.  Also it changes the meaning of the code.

crtrott: yeah this all needs to stay `i` that said the rename of the shadowed variable is probably a good idea :-)

benchmarks/bytes_and_flops/main.cpp
```diff
@@ -86,7 +86,7 @@ int main(int argc, char* argv[]) {
     printf("D must be one of 1,2,4,8,16,32\n");
     return 0;
   }
-  if ((P < 1) && (P > 2)) {
+  if ((P < 1) || (P > 4)) {
```
crtrott, line 89: haha that condition above is funny :-)

crtrott (dismissed):
Yeah the change in the loop as Damien found is wrong - needs to be `i` other than that looks good.

benchmarks/bytes_and_flops/bench_unroll_stride.hpp
cedricchevalier19: Oups, very sorry but thank you for this good catch @dalg24.

dalg24:
Retest this please

dalg24:
Willing to ignore Clang+CUDA and OpenMPTarget builds that were interrupted and one of the HIP build that had (unrelated) runtime issues.

benchmarks/CMakeLists.txt
```diff
@@ -1,3 +1,12 @@
+#FIXME_OPENMPTARGET - compiling in debug mode causes ICE.
```
dalg24, line 1: @rgayatri23 do you want this to be guarded?

rgayatri23: Yeah that would be a good idea to protect it.

crtrott (approved):
I think this is ok without an extra guard. Will remind us that there is actually an issue.

PR #6803, opened 2024-02-11
Add test for current CTAD support with RangePolicy

dalg24:
Extracted from #5517
Intended as a base for discussion when adding deduction guides in a follow up PR.

Adding a compile-only test to showcase what currently works before we make changes.

PR #6841, opened 2024-02-23
Remove redundant RangePolicy constructor

masterleinad:
Following the discussion in https://kokkosteam.slack.com/archives/C5BGU5NDQ/p1708708198675449, there are two possible constructors for
```
Kokkos::RangePolicy        policy(ExecutionSpace(), 0, N);
```
which only differ by a trailing variadic template argument, which is why some compilers have trouble with CTAD in this case. Since the variadic template argument can be empty, we don't actually need the constructor without the variadic template argument.

### DIFF
--- a/core/src/Kokkos_ExecPolicy.hpp
+++ b/core/src/Kokkos_ExecPolicy.hpp
@@ -119,24 +119,6 @@ class RangePolicy : public Impl::PolicyTraits<Properties...> {
             std::enable_if_t<(std::is_convertible_v<IndexType1, member_type> &&
                               std::is_convertible_v<IndexType2, member_type>),
                              bool> = false>
-  inline RangePolicy(const typename traits::execution_space& work_space,
-                     const IndexType1 work_begin, const IndexType2 work_end)
-      : m_space(work_space),
-        m_begin(work_begin),
-        m_end(work_end),
-        m_granularity(0),
-        m_granularity_mask(0) {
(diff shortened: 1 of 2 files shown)


masterleinad:
I rebased after #6803 was merged and added the respective test case.

masterleinad:
Retest this please.

masterleinad:
> Does that change anything about NVCC support?

No, it turns out that actually none of the test cases in that file work with `nvcc 11.0` (and `gcc 9.4.0`). I updated the test case to guard it completely in that case.

nliber (approved):
This is fine.

Do we want to go further (not necessarily in this P/R) and update the constructor we are forwarding to?  That constructor takes a parameter pack for `Args`, which is overkill, as the only thing you can pass to it is `ChunkSize`.  You can (IMO nonsensically) pass `ChunkSize` multiple times, so technically that would be a breaking change.

masterleinad:
> Do we want to go further (not necessarily in this P/R) and update the constructor we are forwarding to? That constructor takes a parameter pack for `Args`, which is overkill, as the only thing you can pass to it is `ChunkSize`. You can (IMO nonsensically) pass `ChunkSize` multiple times, so technically that would be a breaking change.

@nliber Please take care of that in a follow-up pull request.

dalg24:
Taking a leap of faith.  We are having issues with the CI but we want to make progress on this.

PR #6845, opened 2024-02-28
Remove variadic range policy constructor

nliber:
To be added after #6841.

This removes the `RangePolicy` variadic constructor (in favor of two non-variadic constructors) and associated `set` functions, as the only extra parameter was `ChunkSize`.

`ChunkSize` is now pass-by-value in `set` as well to be consistent, as it only wraps an `int`.

In the member initializer for the constructor that takes `ChunkSize`, `m_granularity_mask` is assigned from `m_granularity - 1` instead of `chunk_size.value - 1` to reduced the likelihood of type mismatch (they should be the same value).

### DIFF
--- a/core/src/Kokkos_ExecPolicy.hpp
+++ b/core/src/Kokkos_ExecPolicy.hpp
@@ -150,11 +150,12 @@ class RangePolicy : public Impl::PolicyTraits<Properties...> {
       : m_space(work_space),
         m_begin(work_begin),
         m_end(work_end),
-        m_granularity(chunk_size.value),
-        m_granularity_mask(m_granularity - 1) {
+        m_granularity(0),
+        m_granularity_mask(0) {
     check_conversion_safety(work_begin);
     check_conversion_safety(work_end);
     check_bounds_validity();
+    set_chunk_size(chunk_size.value);
   }
 
   /** \brief  Total range */
@@ -168,10 +169,13 @@ class RangePolicy : public Impl::PolicyTraits<Properties...> {
                     chunk_size) {}
 
  public:
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  KOKKOS_DEPRECATED_WITH_COMMENT("Use set_chunk_size instead")
   inline void set(ChunkSize chunksize) {
     m_granularity      = chunksize.value;
     m_granularity_mask = m_granularity - 1;
   }
+#endif
 
  public:
   /** \brief return chunk_size */
--- a/core/src/Kokkos_ExecPolicy.hpp
+++ b/core/src/Kokkos_ExecPolicy.hpp
@@ -123,13 +123,12 @@ class RangePolicy : public Impl::PolicyTraits<Properties...> {
       : RangePolicy(typename traits::execution_space(), work_begin, work_end) {}
 
   /** \brief  Total range */
-  template <typename IndexType1, typename IndexType2, typename... Args,
+  template <typename IndexType1, typename IndexType2,
             std::enable_if_t<(std::is_convertible_v<IndexType1, member_type> &&
                               std::is_convertible_v<IndexType2, member_type>),
                              bool> = false>
   inline RangePolicy(const typename traits::execution_space& work_space,
-                     const IndexType1 work_begin, const IndexType2 work_end,
-                     Args... args)
+                     const IndexType1 work_begin, const IndexType2 work_end)
       : m_space(work_space),
         m_begin(work_begin),
         m_end(work_end),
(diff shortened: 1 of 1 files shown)


nliber:
Rebased.

masterleinad (approved):

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -148,26 +163,14 @@ class RangePolicy : public Impl::PolicyTraits<Properties...> {
...
-  template <class... Args>
-  inline void set(Args...) {
-    static_assert(
-        0 == sizeof...(Args),
-        "Kokkos::RangePolicy: unhandled constructor arguments encountered.");
-  }
-
-  template <class... Args>
-  inline void set(const ChunkSize& chunksize, Args... args) {
+  inline void set(ChunkSize chunksize) {
```
dalg24, line 171: Drop the function.  We have a `set_chunkl_size()` and don;t really need this one

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -139,7 +138,23 @@ class RangePolicy : public Impl::PolicyTraits<Properties...> {
...
+                     const IndexType1 work_begin, const IndexType2 work_end,
+                     const ChunkSize chunk_size)
+      : m_space(work_space),
+        m_begin(work_begin),
+        m_end(work_end),
+        m_granularity(chunk_size.value),
+        m_granularity_mask(m_granularity - 1) {
+    check_conversion_safety(work_begin);
+    check_conversion_safety(work_end);
+    check_bounds_validity();
```
dalg24, line 157:
```suggestion
    check_bounds_validity();
    set_chunk_size(chunk_size.value);
```

core/src/Kokkos_ExecPolicy.hpp
nliber: That is a breaking change, and `set(ChunkSize)` is a different interface than `set_chunk_size(int)`.

I don't mind making the change I just wanted to point out that it is reasonable for folks to call `set(ChunkSize)` (I have no idea if they actually do or not).

core/src/Kokkos_ExecPolicy.hpp
nliber: I'm setting the chunk size in the member initializers for `m_granularity` and `m_granularity_mask`.  I want to confirm with you that you'd rather I do it via the `set_chunk_size(int)` function.

core/src/Kokkos_ExecPolicy.hpp
masterleinad: We can deprecate `set()` but I wouldn't outright remove it.

core/src/Kokkos_ExecPolicy.hpp
dalg24: I would since technically this avoid repeating the logic.

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -139,7 +138,23 @@ class RangePolicy : public Impl::PolicyTraits<Properties...> {
...
     check_bounds_validity();
     set_auto_chunk_size();
-    set(args...);
+  }
+
+  template <typename IndexType1, typename IndexType2,
+            std::enable_if_t<(std::is_convertible_v<IndexType1, member_type> &&
+                              std::is_convertible_v<IndexType2, member_type>),
+                             bool> = false>
+  inline RangePolicy(const typename traits::execution_space& work_space,
```
dalg24, line 147: While we are at it how about we drop these pointless `inline` specifiers (here and in the previous constructor)?

core/src/Kokkos_ExecPolicy.hpp
dalg24: Yes we would need to go through the normal deprecation process.  It is ok to do as a follow up if you prefer.

core/src/Kokkos_ExecPolicy.hpp
nliber: It's all over the file, and I didn't want to do it piecemeal.  How much clean up should I do here?  (It is simple to get rid of them all.)

dalg24: I would do the 2 constructors in this PR

core/src/Kokkos_ExecPolicy.hpp
nliber: Done.  I also removed setting it in the member initialisers (it would be misleading to set it in both places).

core/src/Kokkos_ExecPolicy.hpp
nliber: I just did it now.  It really is part of this.

core/src/Kokkos_ExecPolicy.hpp
dalg24: >  I also removed setting it in the member initialisers (it would be misleading to set it in both places).

I think you forgot to push these changes.  But whatever, this can be done as a follow up.

dalg24:
Retest this please

dalg24:
The SYCL build failed with
```
/var/jenkins/workspace/Kokkos_PR-6845/core/unit_test/TestRangePolicyCTAD.cpp:87:37: error: unused variable 'des' [-Werror,-Wunused-const-variable]
const Kokkos::DefaultExecutionSpace des =
                                    ^
1 error generated.
make[2]: *** [core/unit_test/CMakeFiles/Kokkos_CoreTestCompileOnly.dir/build.make:160: core/unit_test/CMakeFiles/Kokkos_CoreTestCompileOnly.dir/TestRangePolicyCTAD.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:1943: core/unit_test/CMakeFiles/Kokkos_CoreTestCompileOnly.dir/all] Error 2

dalg24:
This is broken in develop so I will merge

PR #6847, opened 2024-02-29
Validate environment variables "KOKKOS_..."

brian-kelley:
For any environment variables prefixed with ``KOKKOS_`` (case insensitive), check that it is one of the recognized variables that kokkos or kokkos-tools uses. Give a warning if it's unrecognized, or if it's recognized but not in the correct case (all-caps).

This is based off work @cwpearson did for Tpetra in https://github.com/trilinos/Trilinos/pull/12722.

Example usage:
```
export KOKKOS_NOT_A_VALID_VAR=123
export KOKKOS_device_id=0
./example 10
```
prints:
```
Warning: environment variable "KOKKOS_device_id" is
not in all-caps, but is otherwise one of the valid
environment variables used by Kokkos. If you intended
to set "KOKKOS_DEVICE_ID", this variable
will not have the same effect.

Warning: environment variable "KOKKOS_NOT_A_VALID_VAR"
is prefixed like a variable that controls Kokkos,
but is not in the list of valid
Kokkos environment variables.
...
```

### DIFF
--- a/core/src/impl/Kokkos_Core.cpp
+++ b/core/src/impl/Kokkos_Core.cpp
@@ -34,6 +34,7 @@
 #include <cstdlib>
 #include <stack>
 #include <functional>
+#include <unordered_set>
 #include <list>
 #include <cerrno>
 #include <random>
@@ -42,6 +43,14 @@
 #include <unistd.h>
 #else
 #include <windows.h>
+#endif
+
+// environ is included in the C runtime on posix platforms.
+// It contains a null-terminated list of existing environment variables.
+// This is used below in validate_prefixed_env_vars().
+#if not(defined(WIN) && (_MSC_VER >= 1900))
+// This declaration needs to be in the global namespace.
+extern char** environ;
 #endif
 
 //----------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


dalg24 (commented):
* We need some way of programmatically of disabling warnings for environment variables that match a user-specified regex like we do for command line arguments
https://github.com/kokkos/kokkos/blob/bb734012e2011165187b930f055f0c5c14cff4da/core/src/impl/Kokkos_Command_Line_Parsing.cpp#L243-L252
* You need to add tests.  See examples here
https://github.com/kokkos/kokkos/blob/04a5334c699cb9b87293d27bc73090b3b7c13019/core/unit_test/TestParseCmdLineArgsAndEnvVars.cpp#L327-L336
* We need to make sure we don't raise warnings for env variables that are commonly set by modules on leadership computing facilities.  I expect things like `Kokkos_ROOT`, `Kokkos_DIR`, `Kokkos_HOME`, etc.

brian-kelley:
@dalg24 I think I addressed your points. Can you think of other variables that should be ignored (from modules etc.)?

core/src/impl/Kokkos_Command_Line_Parsing.cpp
```diff
@@ -24,9 +24,16 @@
 #include <cstring>
 #include <iostream>
 #include <regex>
+#include <unordered_set>
```
dalg24, line 27: Forgot to remove

core/src/impl/Kokkos_Command_Line_Parsing.cpp
```diff
@@ -241,19 +248,55 @@ void Kokkos::Impl::warn_deprecated_command_line_argument(
...
+    std::regex{"KOKKOS_LIBRARY_DIRS=.*",
+               std::regex::egrep | std::regex_constants::icase},
+    std::regex{"KOKKOS_LIBRARY_DIRECTORIES=.*",
+               std::regex::egrep | std::regex_constants::icase},
+    std::regex{"KOKKOS_SOURCE=.*",
+               std::regex::egrep | std::regex_constants::icase},
+    std::regex{"KOKKOS_SRC=.*",
+               std::regex::egrep | std::regex_constants::icase},
+    std::regex{"KOKKOS_INSTALL=.*",
+               std::regex::egrep | std::regex_constants::icase},
```
dalg24, line 279: Let's discuss that list at the next developer meeting.
We need to check with other projects in the Kokkos ecosystem that this works for them.

I purposefully omitted `KOKKOS_VARIORUM_FUNC_TYPE`, I don't think we should accommodate this one.

core/unit_test/TestParseCmdLineArgsAndEnvVars.cpp
```diff
@@ -389,6 +389,64 @@ TEST(defaultdevicetype, env_vars_tune_internals) {
...
+        {"KOKKOS_EXTENSION_WHATEVER", "123"},  // user explicitly asked not to
+                                               // warn about that prefix
+    }};
+    SKIP_IF_ENVIRONMENT_VARIABLE_ALREADY_SET(ev);
+    ::testing::internal::CaptureStderr();
+    Kokkos::InitializationSettings settings;
+    Kokkos::Impl::parse_environment_variables(settings);
+    auto captured = ::testing::internal::GetCapturedStderr();
+    EXPECT_TRUE(captured.find("KOKKOS_EXTENSION_WHATEVER") ==
+                std::string::npos);
```
dalg24, line 446: This is kinda awkward.  We cannot actually assert that there will be no warnings at all.  I am wondering whether we should add a leading `char* envp[]` argument to `Impl::parse_environment_variables()` so we can control more fine-grained.  What do you think?

core/unit_test/TestParseCmdLineArgsAndEnvVars.cpp
```diff
@@ -389,6 +389,64 @@ TEST(defaultdevicetype, env_vars_tune_internals) {
...
+        {"KOKKOS_device_id", "0"},  // wrong case
+    }};
+    SKIP_IF_ENVIRONMENT_VARIABLE_ALREADY_SET(ev);
+    ::testing::internal::CaptureStderr();
+    Kokkos::InitializationSettings settings;
+    Kokkos::Impl::parse_environment_variables(settings);
+    auto captured = ::testing::internal::GetCapturedStderr();
+    EXPECT_TRUE(captured.find("not recognized") != std::string::npos &&
+                captured.find("KOKKOS_device_id=0") != std::string::npos &&
+                !settings.has_device_id());
```
dalg24, line 418: The windows build failed with
```
[ RUN      ] defaultdevicetype.env_vars_unrecognized
C:\projects\source\core\unit_test\TestParseCmdLineArgsAndEnvVars.cpp(418): error: Value of: captured.find("not recognized") != std::string::npos && captured.find("KOKKOS_device_id=0") != std::string::npos && !settings.has_device_id()
  Actual: false
Expected: true
[  FAILED  ] defaultdevicetype.env_vars_unrecognized (18 ms)

nliber:
The only variable exported form kokkos modules at ALCF is KOKKOS_HOME

dalg24:
> The only variable exported form kokkos modules at ALCF is KOKKOS_HOME

OLCF prefixed them with `OLCF_` so no issue

Issue #6843, opened 2024-02-28
Nightly test failure, cuda/12.2+gcc/8.5: TestRangePolicyCTAD.cpp(23): error: no instance of constructor "RangePolicy" matches the argument list


ndellingwood:
**Describe the bug**

Nightly builds with cuda/12.2+gcc/8.5.0 fail to compile `TestRangePolicyCTAD` with output:

```
22:34:29 /home/jenkins/caraway-new/workspace/Kokkos_develop_carawayA100_cuda1170_cuda_serial/kokkos/core/unit_test/TestRangePolicyCTAD.cpp(23): error: no instance of constructor "RangePolicy" matches the argument list
22:34:29             argument types are: (int64_t, int64_t)
22:34:29   using PolicyMaker = decltype(::Kokkos::RangePolicy(std::declval<Args>()...));
22:34:29                                                      ^
22:34:29           detected during:
22:34:29             instantiation of type "<unnamed>::PolicyMaker<<unnamed>::TestRangePolicyCTAD::i64, <unnamed>::TestRangePolicyCTAD::i64>" at line 27
22:34:29             instantiation of "const bool <unnamed>::IsSamePolicy [with Policy=Kokkos::RangePolicy<>, Args=<<unnamed>::TestRangePolicyCTAD::i64, <unnamed>::TestRangePolicyCTAD::i64>]" at line 58
22:34:29 
22:34:29 /home/jenkins/caraway-new/workspace/Kokkos_develop_carawayA100_cuda1170_cuda_serial/kokkos/core/unit_test/TestRangePolicyCTAD.cpp(23): error: cannot deduce class template arguments
22:34:29   using PolicyMaker = decltype(::Kokkos::RangePolicy(std::declval<Args>()...));
22:34:29                                ^
22:34:29           detected during:
22:34:29             instantiation of type "<unnamed>::PolicyMaker<<unnamed>::TestRangePolicyCTAD::i64, <unnamed>::TestRangePolicyCTAD::i64>" at line 27
22:34:29             instantiation of "const bool <unnamed>::IsSamePolicy [with Policy=Kokkos::RangePolicy<>, Args=<<unnamed>::TestRangePolicyCTAD::i64, <unnamed>::TestRangePolicyCTAD::i64>]" at line 58
22:34:29 
[...]

PR #6850, opened 2024-02-29
Disable NVCC CTAD tests

nliber:
This is addressing issue #6843.


crtrott:
Would the following work or still inside decltype?

```c++
template <class... Args>
using PolicyMaker = decltype(::Kokkos::RangePolicy(std::declval<Args>()...));
template <class Policy, class... Args>
inline constexpr bool IsSamePolicy =
    std::is_same_v<Policy, PolicyMaker<Args...>>;
```

with 

```c++
template <class... Args>
constexpr auto PolicyMaker(Args ... args) { return  ::Kokkos::RangePolicy(std::declval<Args>()...); }
template <class Policy, class... Args>
inline constexpr bool IsSamePolicy =
    std::is_same_v<Policy, decltype(PolicyMaker(declval<Args>()...))>;
```

nliber:
That didn't work.  Going back to my original tests (without using a macro).

nliber:
Update:

I didn't notice until now that the explicit CTAD guides are not in Kokkos_ExecPolicy.hpp.

What I think is happening (I've tried checking the standard wording on CTAD, but I find the wording on implicit guides to be inscrutable :-(():

Because `RangePolicy` takes a parameter pack, there are no template parameters to deduce, so as long as there is a matching constructor in the primary template for `RangePolicy`, `RangePolicy(x, y)` always deduces to `RangePolicy<>::RangePolicy(x, y)`.

All the current tests only check for things that should deduce to `RangePolicy<>`, and those updated tests compile under both gcc 12.2 and clang 17.0.3.

However, nvcc has a bug where that won't deduce to anything.  If I add back in the explicit guides:

```c++
RangePolicy()->RangePolicy<>;

RangePolicy(int64_t, int64_t)->RangePolicy<>;
RangePolicy(int64_t, int64_t, ChunkSize const&)->RangePolicy<>;

RangePolicy(DefaultExecutionSpace const&, int64_t, int64_t)->RangePolicy<>;
RangePolicy(DefaultExecutionSpace const&, int64_t, int64_t, ChunkSize const&)
    ->RangePolicy<>;

template <typename ES, typename = std::enable_if_t<is_execution_space_v<ES>>>
RangePolicy(ES const&, int64_t, int64_t)->RangePolicy<ES>;

template <typename ES, typename = std::enable_if_t<is_execution_space_v<ES>>>
RangePolicy(ES const&, int64_t, int64_t, ChunkSize const&)->RangePolicy<ES>;
```

This compiles under gcc 12.2, clang 17.0.3 and cuda 12.3 + gcc 12.2.

I still have other combos to check, and need to add back in non-DefaultExecutionSpace tests, but this path is promising...

core/unit_test/TestRangePolicyCTAD.cpp
```diff
@@ -17,76 +17,134 @@
...
-
-template <class Policy, class... Args>
-inline constexpr bool IsSamePolicy =
-    std::is_same_v<Policy, PolicyMaker<Args...>>;
+struct TestRangePolicyCTAD {
+  struct SomeExecutionSpace {
+    using execution_space = SomeExecutionSpace;
+    using size_type       = size_t;
 
-#define KOKKOS_TEST_RANGE_POLICY(...) static_assert(IsSamePolicy<__VA_ARGS__>)
```
masterleinad, line 31: Can you try if this approach also works to make the test a little more concise?

nliber: The current version appears to be working under all the compilers we care about.  I've checked all the logs in the CI, and all those builds are passing w/o warnings in this area.

We've been down a similar path already.  That is a material change (no instance variables) that may not work across the board.  I'm fine with trying that in the future, but not part of this P/R.

nliber:
@ndellingwood can you confirm this works?

ndellingwood:
@nliber the culprit build successfully completed 👍  tested with sha d78a7d4383786359ee8692af5b30aac973fca0da

core/unit_test/TestRangePolicyCTAD.cpp
masterleinad: I'm fine with a follow-up pull request for that.

masterleinad (approved):

masterleinad:
Supersedes https://github.com/kokkos/kokkos/pull/5517.

ndellingwood (approved):

dalg24 (commented):
Fine by me but would like @ndellingwood to confirm this resolves the nightly failures

dalg24:
Ignoring the HIP failures

ndellingwood:
@dalg24 I tested this with the culprit build and was able to successfully compile

PR #6849, opened 2024-02-29
Create mirror view of the keys on the host in `sort_by_key` fallback implementation

dalg24:
Fix #6848 
Creating a mirror view and deep copying was guarded for OpenMPTarget but it really should be done unconditionally.
It will be a no-op if the keys are accessible from the host.

PR #6856, opened 2024-03-01
Fix fallback implementation for sort_by_key

masterleinad:
Alternative to #6849. For the fallback implementation of `sort_by_key` using  `Kokkos::sort`, we need to know in advance if the sort executes on the same device or the host so we can copy the keys used in the comparator accordingly.
This pull request makes this decision explicitly and we only call `sort` on the device in cases where we are sure that it doesn't execute in the host. Note that this disregards runtime decisions like when using `SYCL` with `LayoutStride` and we just play it safe by executing on the host in this case. Making it a compile-time decision might improve compile-time and avoids dealing with SYCL device-copyable issues in both branches.

### DIFF
--- a/algorithms/src/sorting/impl/Kokkos_SortByKeyImpl.hpp
+++ b/algorithms/src/sorting/impl/Kokkos_SortByKeyImpl.hpp
@@ -116,7 +116,7 @@ void sort_by_key_cudathrust(
 }
 #endif
 
-#if defined(KOKKOS_ENABLE_SYCL)
+#if defined(KOKKOS_ENABLE_ONEDPL)
 template <class Layout>
 struct sort_on_device<Kokkos::Experimental::SYCL, Layout>
     : std::bool_constant<std::is_same_v<Layout, Kokkos::LayoutLeft> ||
--- a/algorithms/src/sorting/impl/Kokkos_SortByKeyImpl.hpp
+++ b/algorithms/src/sorting/impl/Kokkos_SortByKeyImpl.hpp
@@ -87,7 +87,19 @@ static_assert_is_admissible_to_kokkos_sort_by_key(const ViewType& /* view */) {
                 "LayoutRight, LayoutLeft or LayoutStride.");
 }
 
+// For the fallback sort_by_key implementation we might need to execute
+// Kokkos::sort on the host. This type trait determines at compile-time where we
+// want to execute.
+template <class ExecutionSpace, class Layout>
+struct sort_on_device : std::false_type {};
+
+template <class T, class Layout>
+inline constexpr bool sort_on_device_v = sort_on_device<T, Layout>::value;
+
 #if defined(KOKKOS_ENABLE_CUDA)
+template <class Layout>
+struct sort_on_device<Kokkos::Cuda, Layout> : std::true_type {};
+
 template <class KeysDataType, class... KeysProperties, class ValuesDataType,
           class... ValuesProperties, class... MaybeComparator>
 void sort_by_key_cudathrust(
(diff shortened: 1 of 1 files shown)


algorithms/src/sorting/impl/Kokkos_SortByKeyImpl.hpp
```diff
@@ -87,7 +87,19 @@ static_assert_is_admissible_to_kokkos_sort_by_key(const ViewType& /* view */) {
                 "LayoutRight, LayoutLeft or LayoutStride.");
 }
 
+// For the fallback sort_by_key implementation we might need to execute
+// Kokkos::sort on the host. This type trait determines at compile-time where we
+// want to execute.
+template <class ExecutionSpace, class Layout>
+struct sort_on_device : std::false_type {};
```
dalg24, line 94: Why bother with the trait?  It doesn't seem like you use it.  You can directly specialize the variable template.

algorithms/src/sorting/impl/Kokkos_SortByKeyImpl.hpp
```diff
@@ -152,6 +171,8 @@ void sort_by_key_via_sort(
     const Kokkos::View<KeysDataType, KeysProperties...>& keys,
     const Kokkos::View<ValuesDataType, ValuesProperties...>& values,
     MaybeComparator&&... maybeComparator) {
+  static_assert(sizeof...(MaybeComparator) <= 1);
```
dalg24, line 174: Sure

algorithms/src/sorting/impl/Kokkos_SortByKeyImpl.hpp
```diff
@@ -87,7 +87,19 @@ static_assert_is_admissible_to_kokkos_sort_by_key(const ViewType& /* view */) {
                 "LayoutRight, LayoutLeft or LayoutStride.");
 }
 
+// For the fallback sort_by_key implementation we might need to execute
+// Kokkos::sort on the host. This type trait determines at compile-time where we
```
dalg24, line 91: > we might need to execute Kokkos::sort on the host

"We need to consider if Kokkos::sort defers to the fallback implementation that copies the array to the host and uses std::sort.  The value is determined by the implementation choices in the `impl/Kokkos_SortImpl.hpp` which mean they need to be kept in sync (sigh...)"

algorithms/src/sorting/impl/Kokkos_SortByKeyImpl.hpp
```diff
@@ -104,6 +116,12 @@ void sort_by_key_cudathrust(
 }
 #endif
 
+#if defined(KOKKOS_ENABLE_SYCL)
```
dalg24, line 119: Does it not depend on whether oneDPL was found or not?

algorithms/src/sorting/impl/Kokkos_SortByKeyImpl.hpp
```diff
@@ -87,7 +87,18 @@ static_assert_is_admissible_to_kokkos_sort_by_key(const ViewType& /* view */) {
                 "LayoutRight, LayoutLeft or LayoutStride.");
 }
 
+// For the fallback implementation for sort_by_key using Kokkos::sort, we need
+// to consider if Kokkos::sort defers to the fallback implementation that copies
+// the array to the host and uses std::sort. This decision is made in
+// impl/Kokkos_SortImpl.hpp and this type trait decides at runtime when it is
```
dalg24, line 93: > this type trait decides at runtime when it is safe to assume that Kokkos::sort doesn't manually copy to the host

I don't understand this sentence.  There is no more trait and the variable template does not "decide" anything.  Do you mean it is used somewhere to determine what to do?

dalg24: Please improve/update the comment

masterleinad: d5c747872f1f2d77b765462c4230408e4481720d

dalg24 (approved):

dalg24:
Retest this please

masterleinad:
Only the CUDA-12.2-NVHPC build on fetnat 05 and the OpenMPTarget build on fetnat04 failed due to disk space issues.

ldh4 (approved):

dalg24:
Ignoring `Disk quota exceeded` errors in the OpenMPTarget and one of the CUDA builds.

PR #6857, opened 2024-03-01
Fix amdclang++ compilation

tpadioleau:
I came across a link time issue when compiling Kokkos 4.2.01 tests with both Cray and AMD clang compilers for the HIP backend when enabling the `Kokkos_ENABLE_HIP_RELOCATABLE_DEVICE_CODE` option. I got `undefined reference to __hip_fatbin`
while it was working fine with the `hipcc` compiler driver.

I don't know if these two compilers are officially supported by Kokkos but adding the `--hip-link` link flag seems to fix the issue:
- suggested [here](https://github.com/ROCm/HIP/issues/2203)
- `hipcc` seems to always add this flag [here](https://github.com/ROCm/HIPCC/blob/3555aa113d350e94c8d17c5ff519d1acaf9aefb9/src/hipBin_amd.h#L854).

I only tested it on a rocm 5.7.1 environment on Adastra. As a result, `hipcc`, `amdclang++` and `cce` are all able to compile the Kokkos tests with `Kokkos_ENABLE_HIP_RELOCATABLE_DEVICE_CODE` disabled or enabled. Unfortunately when this latter option is enabled the two tests `hip.mathspecialfunc_cbesselh1stkind` and `hip.mathspecialfunc_cbesselh2ndkind` do not pass with `amdclang++` and `cce` even though they pass with `hipcc`. I don't know the reason yet.

What do you think of adding this flag by default ?


masterleinad (approved):

cmake/kokkos_arch.cmake
```diff
@@ -571,6 +571,9 @@ IF (KOKKOS_ENABLE_HIP)
     COMPILER_SPECIFIC_FLAGS(
       DEFAULT -fgpu-rdc
     )
+    COMPILER_SPECIFIC_LINK_OPTIONS(
```
dalg24, line 574: @Rombur  What's the best way to do this?
* Comment + add unconditionally
* Put the call within a `IF(NOT KOKKOS_CXX_COMPILER_ID STREQUAL HIPCC)`
* Specify a non-default clause for hipcc with an empty flag?

Rombur: First, I want to try on Frontier. We have seen in the past the cray compiler behaving differently on different machines. If it works correctly,  we should guard this flag with `IF(NOT KOKKOS_CXX_COMPILER_ID STREQUAL HIPCC)`

tpadioleau:
cc @etiennemlb

Rombur:
I can reproduce the error. Unfortunately this PR does not really fix the problem for me. I've tried the following compilers:
`amdclang` with ROCm 5.2, `CCE 15.0` with ROCm 5.2, `amdclang` with ROCm 6.0, and `CCE 17.0` with ROCm 6.0. This PR fixed the issue for `amdclang` with ROCm 6.0. For the other compilers, the compilation stops faster with PR. I am getting errors from DESUL as soon as we link @tpadioleau which version of the cray compiler are you using?

tpadioleau:
I have just updated the PR with guards for hipcc. Let me know if it needs further modifications.

I may have been confused about the Cray environment, I am not sure to have correctly tested the Cray compiler. I am testing it again.

Rombur (approved):

dalg24:
Ignoring (unrelated) HIP runtime failures and couple CUDA builds that did not start because the test machine had no disk space.

dalg24:
@Rombur please add this to the changelog

PR #6860, opened 2024-03-05
Fix C-style cast warnings with clang-tidy

masterleinad:
Running `clang-tidy` locally with our current settings, I still see some complaints about C-style casts.

### DIFF
--- a/core/src/Serial/Kokkos_Serial_Task.hpp
+++ b/core/src/Serial/Kokkos_Serial_Task.hpp
@@ -121,7 +121,7 @@ class TaskQueueSpecializationConstrained<
     using task_base_type = TaskBase;
     using queue_type     = typename scheduler_type::queue_type;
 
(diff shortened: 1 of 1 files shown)


crtrott (approved):
I don't really see the reason to have `auto*` there why not just `auto`:
https://godbolt.org/z/6Eo178jjx

Not blocking, I mean it might make it a bit clearer that we are looking at a const ptr, not a ptr of const. But still, if anyone else feels like get rid of * please feel free to say.

dalg24 (approved):
I agree with Christian suggesting to use `auto` instead of `auto*` in general but I don't think it is worth arguing about it, especially for that specific code.

masterleinad:
Note that there is another clang-tidy check for this and that the LLVM Coding Standards advise to make it obvious if an `auto`-typed variable is a pointer, see https://clang.llvm.org/extra/clang-tidy/checks/readability/qualified-auto.html.

Issue #870, opened 2017-06-06
Add view label to `View bounds error` message for CUDA backend


stanmoore1:
It would be *really* helpful for debugging if the the view label was added to a `View bounds error` on the device with the CUDA backend. Right now, all I get is:

<pre>
:0: : block: [0,0,0], thread: [0,26,0] Assertion `View bounds error` failed.
terminate called after throwing an instance of 'std::runtime_error'
  what():  cudaDeviceSynchronize() error( cudaErrorAssert): device-side assert triggered ../../lib/kokkos/core/src/Cuda/Kokkos_Cuda_Impl.cpp:122
Traceback functionality not available
</pre>

ibaned:
Hmmm... one problem would be that `View::label()` returns a `std::string` right now, which is a no-go inside GPU code. The second problem being that even simple string appending without `std::string` or even `strcpy` becomes a real chore on the GPU. There may be ways around those obstacles, just noting that this is non-trivial to do.

stanmoore1:
Well, trying to find out which view is out of bounds without any hints from Kokkos is also non-trivial :)

ibaned:
yep, yep. I'll try to figure it out, its not impossible yet.

stanmoore1:
This would be a non-issue if `gdb` or `cuda-gdb` could give me the line number in my functor where it is crashing, but the last line in my code that the stack trace shows is where I call the parallel_for and then the stack trace goes into `Kokkos_Parallel.hpp`, which isn't helpful at all.

stanmoore1:
Using `cuda-gdb` only gives a single frame strack trace of `Kokkos_Cuda_abort.hpp:75`.

stanmoore1:
The view label is stored as a `char array`, not as `std::string`, so why not just copy that up to device memory and `printf` it out as a `C-string`? Seems pretty trivial to me...

ibaned:
It is actually already in device memory, _for some views_. Much of what I've been doing for the past day is looking for the cases of views that don't have a label, this includes `Memory::Unmanaged` and views that don't have `Memory::Unmanaged` but were constructed by being given a pointer. I _think_ that I can detect all these edge cases safely inside GPU code...

[...]

PR #6861, opened 2024-03-06
Improve message on view out of bounds access and always abort

### LINKED ISSUE #6851: Consider aborting instead of throwing an exception in view out-of-bound accesses
(Discussion started by Stan on Slack)

Throwing an exception makes things awkward when using multiple MPI ranks.
There does not seem to be a good reason to throw form the host side.
The code that does the out-of-bound check is there
https://github.com/kokkos/kokkos/blob/04a5334c699cb9b87293d27bc73090b3b7c13019/core/src/impl/Kokkos_ViewMapping.hpp#L4016-L4023

We should just call abort instead of throwing.

dalg24:
Fix #6851 

* Always call abort when out-of-bounds access is detected.  (We were throwing a runtime exception when it occurred on the host side.)
* Improve the error message

Before this PR
```
View bounds error of view my_label ( 10 < 1 , 0 < 2 )
```
changed to
```
Kokkos::View ERROR: out of bounds access (label="mylabel") with indices [10,0] but extents [1,2]
```

TLDR; Stan's wishes are my command

### DIFF
--- a/core/src/impl/Kokkos_ViewMapping.hpp
+++ b/core/src/impl/Kokkos_ViewMapping.hpp
@@ -3541,7 +3541,7 @@ template <class Map, class... Indices, std::size_t... Enumerate>
 KOKKOS_FUNCTION bool within_range(Map const& map,
                                   std::index_sequence<Enumerate...>,
                                   Indices... indices) {
-  return (... && ((std::size_t)indices < map.extent(Enumerate)));
+  return (((std::size_t)indices < map.extent(Enumerate)) && ...);
 }
 
 template <class... Indices>
--- a/core/src/impl/Kokkos_ViewMapping.hpp
+++ b/core/src/impl/Kokkos_ViewMapping.hpp
@@ -3545,7 +3545,7 @@ KOKKOS_FUNCTION bool check_bounds(Map const& map,
 }
 
 template <class... Indices>
-KOKKOS_FUNCTION constexpr char* append_formated_multidimensional_index(
+KOKKOS_FUNCTION constexpr char* append_formatted_multidimensional_index(
     char* dest, Indices... indices) {
   char* d = dest;
   Kokkos::Impl::strcat(d, "[");
@@ -3565,7 +3565,7 @@ KOKKOS_FUNCTION constexpr char* append_formated_multidimensional_index(
 template <class Map, class... Indices, std::size_t... Enumerate>
 KOKKOS_FUNCTION void print_extents(char* dest, Map const& map,
                                    std::index_sequence<Enumerate...>) {
-  (void)append_formated_multidimensional_index(dest, map.extent(Enumerate)...);
+  (void)append_formatted_multidimensional_index(dest, map.extent(Enumerate)...);
 }
 
 template <class T>
@@ -3616,7 +3616,7 @@ KOKKOS_INLINE_FUNCTION void view_verify_operator_bounds(
       }
     }();)
     Kokkos::Impl::strcat(err, "\") with indices ");
-    Kokkos::Impl::append_formated_multidimensional_index(err, args...);
+    Kokkos::Impl::append_formatted_multidimensional_index(err, args...);
     Kokkos::Impl::strcat(err, " but extents ");
     Kokkos::Impl::print_extents(err, map,
                                 std::make_index_sequence<sizeof...(Args)>());
--- a/core/src/impl/Kokkos_ViewMapping.hpp
+++ b/core/src/impl/Kokkos_ViewMapping.hpp
@@ -2737,6 +2737,7 @@ class ViewMapping<
       : m_impl_handle(arg_handle), m_impl_offset(arg_offset) {}
 
  public:
+  using printable_label_typedef = void;
   enum { is_managed = Traits::is_managed };
 
   //----------------------------------------
@@ -3567,6 +3568,9 @@ KOKKOS_FUNCTION void print_extents(char* dest, Map const& map,
   (void)append_formated_multidimensional_index(dest, map.extent(Enumerate)...);
 }
 
+template <class T>
+using printable_label_typedef_t = typename T::printable_label_typedef;
+
 template <class MemorySpace, class ViewType, class MapType, class... Args>
 KOKKOS_INLINE_FUNCTION void view_verify_operator_bounds(
     Kokkos::Impl::ViewTracker<ViewType> const& tracker, const MapType& map,
@@ -3576,14 +3580,41 @@ KOKKOS_INLINE_FUNCTION void view_verify_operator_bounds(
     char err[256] = "";
     Kokkos::Impl::strcat(err, "Kokkos::View ERROR: out of bounds access");
     Kokkos::Impl::strcat(err, " label=(\"");
-    if (tracker.m_tracker.has_record()) {
-      KOKKOS_IF_ON_HOST(
-          Kokkos::Impl::strncat(
-              err, tracker.m_tracker.template get_label<void>().c_str(), 128);)
-      KOKKOS_IF_ON_DEVICE(Kokkos::Impl::strcat(err, "**UNAVAILABLE**");)
-    } else {
+    KOKKOS_IF_ON_HOST([&] {
+      if (tracker.m_tracker.has_record()) {
+        Kokkos::Impl::strncat(
+            err, tracker.m_tracker.template get_label<void>().c_str(), 128);
+        return;
+      }
       Kokkos::Impl::strcat(err, "**UNMANAGED**");
-    }
+    }();)
+    KOKKOS_IF_ON_DEVICE([&] {
+      // Check #1: is there a SharedAllocationRecord?  (we won't use it, but
+      // if its not there then there isn't a corresponding
+      // SharedAllocationHeader containing a label).  This check should cover
+      // the case of Views that don't have the Unmanaged trait but were
+      // initialized by pointer.
+      if (!tracker.m_tracker.has_record()) {
+        Kokkos::Impl::strcat(err, "**UNMANAGED**");
+        return;
+      }
+      // Check #2: does the ViewMapping have the printable_label_typedef
+      // defined? See above that only the non-specialized standard-layout
+      // ViewMapping has this defined by default. The existence of this
+      // alias indicates the existence of MapType::is_managed
+      if constexpr (is_detected_v<printable_label_typedef_t, MapType>) {
+        // Check #3: is the View managed as determined by the MemoryTraits?
+        if constexpr (MapType::is_managed != 0) {
+          SharedAllocationHeader const* const header =
+              SharedAllocationHeader::get_header(
+                  static_cast<void const*>(map.data()));
+          char const* const label = header->label();
+          Kokkos::Impl::strcat(err, label);
+          return;
+        }
+        Kokkos::Impl::strcat(err, "**UNMANAGED**");
+      }
+    }();)
     Kokkos::Impl::strcat(err, "\") with indices ");
     Kokkos::Impl::append_formated_multidimensional_index(err, args...);
     Kokkos::Impl::strcat(err, " but extents ");
--- a/core/src/impl/Kokkos_ViewMapping.hpp
+++ b/core/src/impl/Kokkos_ViewMapping.hpp
@@ -3548,24 +3548,24 @@ template <class... Indices>
 KOKKOS_FUNCTION constexpr char* append_formatted_multidimensional_index(
     char* dest, Indices... indices) {
   char* d = dest;
-  Kokkos::Impl::strcat(d, "[");
+  strcat(d, "[");
   (
       [&] {
-        d += Kokkos::Impl::strlen(d);
-        (void)Kokkos::Impl::to_chars_i(d,
-                                       d + 20,  // 20 digits ought to be enough
-                                       indices);
-        Kokkos::Impl::strcat(d, ",");
+        d += strlen(d);
+        to_chars_i(d,
+                   d + 20,  // 20 digits ought to be enough
+                   indices);
+        strcat(d, ",");
       }(),
       ...);
-  d[Kokkos::Impl::strlen(d) - 1] = ']';  // overwrite trailing comma
+  d[strlen(d) - 1] = ']';  // overwrite trailing comma
   return dest;
 }
 
 template <class Map, class... Indices, std::size_t... Enumerate>
 KOKKOS_FUNCTION void print_extents(char* dest, Map const& map,
                                    std::index_sequence<Enumerate...>) {
-  (void)append_formatted_multidimensional_index(dest, map.extent(Enumerate)...);
+  append_formatted_multidimensional_index(dest, map.extent(Enumerate)...);
 }
 
 template <class T>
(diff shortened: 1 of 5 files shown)


dalg24:
(I am unsure about the code to deduce the view label so I pushed only the tests that I expect to fail to confirm what the current behavior, i.e. before this PR, is)

stanmoore1:
I'm a little concerned with this change: [Do not bother with the label on the device side](https://github.com/kokkos/kokkos/pull/6861/commits/40348e88d265f7dfaf42354db5be18db21d55808). I originally asked for this enhancement in https://github.com/kokkos/kokkos/issues/870. I know it isn't general, but having the label in the error message on the GPU even in some cases is very helpful.

core/unit_test/TestViewOutOfBoundsAccess.hpp
```diff
@@ -0,0 +1,171 @@
...
+  using V2 = Kokkos::View<int**,       ExecutionSpace>;
+  using V3 = Kokkos::View<int***,      ExecutionSpace>;
+  using V4 = Kokkos::View<int****,     ExecutionSpace>;
+  using V5 = Kokkos::View<int*****,    ExecutionSpace>;
+  using V6 = Kokkos::View<int******,   ExecutionSpace>;
+  using V7 = Kokkos::View<int*******,  ExecutionSpace>;
+  using V8 = Kokkos::View<int********, ExecutionSpace>;
+  constexpr bool is_device =
+      std::is_same_v<ExecutionSpace, Kokkos::DefaultExecutionSpace> &&
+      std::is_same_v<Kokkos::DefaultExecutionSpace,
```
dalg24, line 113: I had gotten it wrong anyway it should have been and default exec is NOT the same as default host exec

dalg24:
> I'm a little concerned with this change: [Do not bother with the label on the device side](https://github.com/kokkos/kokkos/pull/6861/commits/40348e88d265f7dfaf42354db5be18db21d55808). I originally asked for this enhancement of printing out device side labels in #870. I know it isn't general, but having the label in the error message on the GPU even in some cases is very helpful.

This is fair.  The rational for not bothering with it was that the label info being printed will be garbage if indexing into a subview there the pointer was shifted from the start of the original view.
I reverted the changes.

core/src/impl/Kokkos_ViewMapping.hpp
```diff
@@ -3537,110 +3537,90 @@ class ViewMapping<
...
-         view_verify_operator_bounds<R + 1>(map, args...);
+template <class... Indices>
+KOKKOS_FUNCTION constexpr char* append_formated_multidimensional_index(
+    char* dest, Indices... indices) {
+  char* d = dest;
+  Kokkos::Impl::strcat(d, "[");
+  (
+      [&] {
+        d += Kokkos::Impl::strlen(d);
+        (void)Kokkos::Impl::to_chars_i(d,
```
masterleinad, line 3555: Why do you cast to `void` here? Does any compiler complain about the unused return type?

core/src/impl/Kokkos_ViewMapping.hpp
```diff
@@ -3537,110 +3537,90 @@ class ViewMapping<
...
+                                       d + 20,  // 20 digits ought to be enough
+                                       indices);
+        Kokkos::Impl::strcat(d, ",");
+      }(),
+      ...);
```
masterleinad, line 3560: Why not just
```suggestion
      for(index: {indices...}) {
        d += Kokkos::Impl::strlen(d);
        Kokkos::Impl::to_chars_i(d,
                                       d + 20,  // 20 digits ought to be enough
                                       index);
        Kokkos::Impl::strcat(d, ",");
      }
```
for a little more "standard" syntax. If we really want to allow different types we can cast `indices` in the initializer list to `std::size_t` or so.

core/src/impl/Kokkos_ViewMapping.hpp
```diff
@@ -3537,110 +3537,90 @@ class ViewMapping<
...
-inline void view_error_operator_bounds(char* buf, int len, const MapType& map,
-                                       const iType& i, Args... args) {
-  const int n = snprintf(
-      buf, len, " %ld < %ld %c", static_cast<unsigned long>(i),
-      static_cast<unsigned long>(map.extent(R)), (sizeof...(Args) ? ',' : ')'));
-  view_error_operator_bounds<R + 1>(buf + n, len - n, map, args...);
+template <class Map, class... Indices, std::size_t... Enumerate>
+KOKKOS_FUNCTION void print_extents(char* dest, Map const& map,
+                                   std::index_sequence<Enumerate...>) {
+  (void)append_formated_multidimensional_index(dest, map.extent(Enumerate)...);
```
masterleinad, line 3568: Why `(void)`?

core/src/impl/Kokkos_ViewMapping.hpp
```diff
@@ -3537,110 +3537,90 @@ class ViewMapping<
...
+    Kokkos::Impl::strcat(err, "Kokkos::View ERROR: out of bounds access");
+    Kokkos::Impl::strcat(err, " label=(\"");
+    KOKKOS_IF_ON_HOST([&] {
+      if (tracker.m_tracker.has_record()) {
+        Kokkos::Impl::strncat(
+            err, tracker.m_tracker.template get_label<void>().c_str(), 128);
+        return;
+      }
+      Kokkos::Impl::strcat(err, "**UNMANAGED**");
+    }();)
```
masterleinad, line 3590: We don't use immediately invoked lambdas elsewhere with `KOKKOS_IF_ON_HOST`. Did you introduce it because of the extra control flow (`return;`) used below?

core/src/impl/Kokkos_ViewMapping.hpp
```diff
@@ -3537,110 +3537,90 @@ class ViewMapping<
 namespace Kokkos {
 namespace Impl {
 
-template <unsigned, class MapType>
-KOKKOS_INLINE_FUNCTION bool view_verify_operator_bounds(const MapType&) {
-  return true;
+template <class Map, class... Indices, std::size_t... Enumerate>
+KOKKOS_FUNCTION bool check_bounds(Map const& map,
```
masterleinad, line 3541:
```suggestion
KOKKOS_FUNCTION bool bounds_satisfied(Map const& map,
```
to make it clearer that we are returning a `bool`.

core/src/impl/Kokkos_ViewMapping.hpp
```diff
@@ -3537,110 +3537,90 @@ class ViewMapping<
...
-           (we won't use it, but if its not there then there isn't
-            a corresponding SharedAllocationHeader containing a label).
-           This check should cover the case of Views that don't
-           have the Unmanaged trait but were initialized by pointer. */
-        if (tracker.m_tracker.has_record()) {
-          operator_bounds_error_on_device(map);
-        } else { Kokkos::abort("View bounds error"); }))
+  if (!check_bounds(map, std::make_index_sequence<sizeof...(Args)>(),
+                    args...)) {
+    char err[256] = "";
```
masterleinad, line 3580: What is the impact on kernel size? Do we only get here in `DEBUG` mode? I guess it doesn't matter much since we only get here with `KOKKOS_ENABLE_DEBUG_BOUNDS_CHECK` anyway.

core/src/impl/Kokkos_ViewMapping.hpp
```diff
@@ -3537,110 +3537,90 @@ class ViewMapping<
...
+      [&] {
+        d += Kokkos::Impl::strlen(d);
+        (void)Kokkos::Impl::to_chars_i(d,
+                                       d + 20,  // 20 digits ought to be enough
+                                       indices);
+        Kokkos::Impl::strcat(d, ",");
+      }(),
+      ...);
+  d[Kokkos::Impl::strlen(d) - 1] = ']';  // overwrite trailing comma
+  return dest;
```
masterleinad, line 3562: Is the return value used anywhere?

core/src/impl/Kokkos_ViewMapping.hpp
```diff
@@ -3537,110 +3537,90 @@ class ViewMapping<
...
+template <class... Indices>
+KOKKOS_FUNCTION constexpr char* append_formated_multidimensional_index(
+    char* dest, Indices... indices) {
+  char* d = dest;
+  Kokkos::Impl::strcat(d, "[");
+  (
+      [&] {
+        d += Kokkos::Impl::strlen(d);
+        (void)Kokkos::Impl::to_chars_i(d,
+                                       d + 20,  // 20 digits ought to be enough
```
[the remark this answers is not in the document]
dalg24: Because it can represent the largest representable signed 64-bit integer.
How would you call that named variable?  I thought the comment was good enough but I am open to change if you have a suggestion that improves readability.

core/src/impl/Kokkos_ViewMapping.hpp
dalg24: Out of habit.  I have not actually tried without it.  Do you want me to drop the cast to `void`?

core/src/impl/Kokkos_ViewMapping.hpp
dalg24: Out of habit.  I do not know if any compiler actually warns.  Do you want me to remove?

core/src/impl/Kokkos_ViewMapping.hpp
dalg24: I introduced it to reduce the complexity of the control flow in the device side and I decided to use it as well for the host code where the improvement is less obvious but I judged that consistency with the device code was a good thing.

core/src/impl/Kokkos_ViewMapping.hpp
dalg24: No impact unless you enable debug_bound_check at configuration time.

core/src/impl/Kokkos_ViewMapping.hpp
dalg24: Not in the current version.  I chose to return the char pointer argument because that was what a number of string manipulation routine do (e.g. strcat)

core/src/impl/Kokkos_ViewMapping.hpp
masterleinad: Yeah, I would prefer that. AFAIK you would only ever get a warning from ignoring a return value if the function is declared with the [[nodiscard]] attribute.

core/src/impl/Kokkos_ViewMapping.hpp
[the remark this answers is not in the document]
dalg24: I will change it but I am not a big fan of pulling in `<limits>` into our headers (admittedly we already to that from the exec policy header).
Also I think that it does not necessarily convey that at the time that code was written, we picked a number of digits that we think will suffice whereas I thought the code as I initially wrote does.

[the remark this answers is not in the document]
dalg24: On 2nd thought, not changing unless more reviewers ask for it.

core/src/impl/Kokkos_ViewMapping.hpp
dalg24: Also the current implementation would yield
```C++
  Kokkos::View<int**> v("v",2,2);
  std::cout << v(1u, -1);  // Kokkos::View ERROR: out of bounds access label=("v") with indices [1,-1] but extents [2,2]
```
which is better than `Kokkos::View ERROR: out of bounds access label=("v") with indices [1,18446744073709551615] but extents [2,2]`

core/src/impl/Kokkos_ViewMapping.hpp
dalg24: I renamed it to `within_range`.  Let me know if you still do not like the name.

core/src/impl/Kokkos_ViewMapping.hpp
dalg24: I was not planning to change.  Please unresolved and comment if you want me to change.

core/src/impl/Kokkos_ViewMapping.hpp
crtrott: agreed

core/src/impl/Kokkos_ViewMapping.hpp
```diff
@@ -3537,110 +3537,88 @@ class ViewMapping<
...
 namespace Impl {
 
-template <unsigned, class MapType>
-KOKKOS_INLINE_FUNCTION bool view_verify_operator_bounds(const MapType&) {
-  return true;
+template <class Map, class... Indices, std::size_t... Enumerate>
+KOKKOS_FUNCTION bool within_range(Map const& map,
+                                  std::index_sequence<Enumerate...>,
+                                  Indices... indices) {
+  return (... && ((std::size_t)indices < map.extent(Enumerate)));
```
crtrott, line 3544: still stumbling slightly over the left fold, but no reason to block

crtrott (approved):
Looks largely fine to me. I kinda agree with Daniel on the host lambda thing, don't see a point in it, and I am also stumbling over the left fold. But its not like I care that much

core/src/impl/Kokkos_ViewMapping.hpp
dalg24: Oh yes sorry I had agreed to change

core/src/impl/Kokkos_ViewMapping.hpp
dalg24: here is what it looks like without the lambda trick
```diff
@@ -3580,13 +3580,12 @@ KOKKOS_INLINE_FUNCTION void view_verify_operator_bounds(
     char err[256] = "";
     strcat(err, "Kokkos::View ERROR: out of bounds access");
     strcat(err, " label=(\"");
-    KOKKOS_IF_ON_HOST([&] {
-      if (tracker.m_tracker.has_record()) {
-        strncat(err, tracker.m_tracker.template get_label<void>().c_str(), 128);
-        return;
-      }
-      strcat(err, "**UNMANAGED**");
-    }();)
+    KOKKOS_IF_ON_HOST(
+        if (tracker.m_tracker.has_record()) {
+          strncat(err, tracker.m_tracker.template get_label<void>().c_str(),
+                  128);
+          return;
+        } else { strcat(err, "**UNMANAGED**"); })
     KOKKOS_IF_ON_DEVICE([&] {
       // Check #1: is there a SharedAllocationRecord?  (we won't use it, but
       // if its not there then there isn't a corresponding

Comment (dalg24) [reply core/src/impl/Kokkos_ViewMapping.hpp]: Anyway I changed it since both Christian and you complained about it

File: core/src/impl/Kokkos_ViewMapping.hpp (hunk 9 above)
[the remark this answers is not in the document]
Comment (masterleinad) [reply core/src/impl/Kokkos_ViewMapping.hpp]:
```diff
diff --git a/core/src/impl/Kokkos_ViewMapping.hpp b/core/src/impl/Kokkos_ViewMapping.hpp
index 89eff343c..b4894a8ec 100644
--- a/core/src/impl/Kokkos_ViewMapping.hpp
+++ b/core/src/impl/Kokkos_ViewMapping.hpp
@@ -3549,15 +3549,12 @@ KOKKOS_FUNCTION constexpr char* append_formatted_multidimensional_index(
     char* dest, Indices... indices) {
   char* d = dest;
   strcat(d, "[");
-  (
-      [&] {
-        d += strlen(d);
-        to_chars_i(d,
-                   d + 20,  // 20 digits ought to be enough
-                   (size_t)indices);
-        strcat(d, ",");
-      }(),
-      ...);
+  for (auto index : {static_cast<size_t>(indices)...}) {
+    d += strlen(d);
+    constexpr int max_digits = 20;  // enough for uint64_t
+    to_chars_i(d, d + max_digits, index);
+    strcat(d, ",");
+  }
   d[strlen(d) - 1] = ']';  // overwrite trailing comma
   return dest;
 }
```
would also look better to me.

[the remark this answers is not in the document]
masterleinad: or just the `max_digits` part.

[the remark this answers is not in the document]
dalg24: The `static_cast` would potentially lead to poor quality error messages (see https://github.com/kokkos/kokkos/pull/6861#discussion_r1517992352).
I don't think the `max_digits` is better than the current version and I am not planning to change unless more reviewers ask me to change it.

Rombur (approved):

dalg24:
Unrelated runtime failures in the ROCm 5.2 HIP build

dalg24:
Ignoring
```
[100%] Linking Fortran executable example
gmake[2]: *** [CMakeFiles/example.dir/build.make:119: example] Error 1
gmake[1]: *** [CMakeFiles/Makefile2:83: CMakeFiles/example.dir/all] Error 2
gmake: *** [Makefile:101: all] Error 2
Error: Process completed with exit code 2.
``` in Fedora rawhide builds

PR #6866, opened 2024-03-08
Scratch space fix for MultiGPU

tcclevenger:
For cuda, use `CudaSpace::(de)allocate()` functions instead of `Kokkos::kokkos_malloc`, `Kokkos::kokkos_free`, and `Kokkos::realloc` to avoid using `Kokkos_SharedAlloc.hpp` classes, which allocate only on default execution space.

### DIFF
--- a/core/src/Cuda/Kokkos_Cuda_Instance.cpp
+++ b/core/src/Cuda/Kokkos_Cuda_Instance.cpp
@@ -428,21 +428,21 @@ void *CudaInternal::resize_team_scratch_space(int scratch_pool_id,
   // Multiple ParallelFor/Reduce Teams can call this function at the same time
   // and invalidate the m_team_scratch_ptr. We use a pool to avoid any race
   // condition.
+  auto mem_space = Kokkos::CudaSpace::impl_create(m_cudaDev, m_stream);
   if (m_team_scratch_current_size[scratch_pool_id] == 0) {
     m_team_scratch_current_size[scratch_pool_id] = bytes;
     m_team_scratch_ptr[scratch_pool_id] =
-        Kokkos::kokkos_malloc<Kokkos::CudaSpace>(
-            "Kokkos::CudaSpace::TeamScratchMemory",
-            m_team_scratch_current_size[scratch_pool_id]);
+        mem_space.allocate("Kokkos::CudaSpace::TeamScratchMemory",
+                           m_team_scratch_current_size[scratch_pool_id]);
   }
   if ((bytes > m_team_scratch_current_size[scratch_pool_id]) ||
       ((bytes < m_team_scratch_current_size[scratch_pool_id]) &&
        (force_shrink))) {
+    mem_space.deallocate(m_team_scratch_ptr[scratch_pool_id],
+                         m_team_scratch_current_size[scratch_pool_id]);
     m_team_scratch_current_size[scratch_pool_id] = bytes;
     m_team_scratch_ptr[scratch_pool_id] =
-        Kokkos::kokkos_realloc<Kokkos::CudaSpace>(
-            m_team_scratch_ptr[scratch_pool_id],
-            m_team_scratch_current_size[scratch_pool_id]);
+        mem_space.allocate("Kokkos::CudaSpace::TeamScratchMemory", bytes);
   }
   return m_team_scratch_ptr[scratch_pool_id];
 }
@@ -459,8 +459,8 @@ void CudaInternal::finalize() {
 
   was_finalized = true;
 
+  auto cuda_mem_space = Kokkos::CudaSpace::impl_create(m_cudaDev, m_stream);
   if (nullptr != m_scratchSpace || nullptr != m_scratchFlags) {
-    auto cuda_mem_space = Kokkos::CudaSpace::impl_create(m_cudaDev, m_stream);
     auto host_mem_space =
         Kokkos::CudaHostPinnedSpace::impl_create(m_cudaDev, m_stream);
     cuda_mem_space.deallocate(m_scratchFlags,
@@ -476,7 +476,8 @@ void CudaInternal::finalize() {
 
   for (int i = 0; i < m_n_team_scratch; ++i) {
     if (m_team_scratch_current_size[i] > 0)
-      Kokkos::kokkos_free<Kokkos::CudaSpace>(m_team_scratch_ptr[i]);
+      cuda_mem_space.deallocate(m_team_scratch_ptr[i],
+                                m_team_scratch_current_size[i]);
   }
 
   m_scratchSpaceCount   = 0;
--- a/core/unit_test/cuda/TestCuda_InterOp_StreamsMultiGPU.cpp
+++ b/core/unit_test/cuda/TestCuda_InterOp_StreamsMultiGPU.cpp
@@ -196,10 +196,11 @@ struct ScratchFunctor {
 
     // Check that each scratch entry has been incremented exactly R times
     int team_error_accum;
+    auto R_loc = R;  // avoid implicit capture of this
     Kokkos::parallel_reduce(
         Kokkos::TeamVectorRange(team, 0, scratch_size),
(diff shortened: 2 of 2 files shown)


core/src/Cuda/Kokkos_Cuda_Instance.cpp
```diff
@@ -428,11 +428,13 @@ void *CudaInternal::resize_team_scratch_space(int scratch_pool_id,
   // Multiple ParallelFor/Reduce Teams can call this function at the same time
   // and invalidate the m_team_scratch_ptr. We use a pool to avoid any race
   // condition.
+  auto mem_space = Kokkos::CudaSpace::impl_create(m_cudaDev, m_stream);
+  Kokkos::Cuda exec_space(m_stream);
```
tcclevenger, line 432: Is this odd to create a Kokkos::Cuda object inside `CudaInternal`? We could always implement this function in `Cuda` class since we have access to the execution space where this is called, and just hand the calling Cuda object to `kokkos_malloc/realloc` below.

core/src/Cuda/Kokkos_Cuda.hpp
```diff
@@ -171,7 +171,18 @@ class Cuda {
 
   KOKKOS_DEPRECATED Cuda(cudaStream_t stream, bool manage_stream);
 
+  ~Cuda() {
+    // In case of multi-gpu, we must free scratch before the
+    // internal space is deleted so that we can pass a valid execution
+    // space to kokkos_free().
+    free_scratch();
+  }
```
tcclevenger, line 179: We get a segfault if we attempt to create a non-default Cuda object inside `CudaInternal::finalize()`, so we cannot free the scratch space on the correct device. 

Not sure exactly why we segfault, but I can look into that more if we don't like this solution.

core/src/Cuda/Kokkos_Cuda_Instance.cpp
masterleinad: I would prefer if we just used `mem_space.allocate` and `mem_space.deallocate` here as we do for the other scratch allocations. This doesn't require adding an overload for `kokkos_malloc` with an execution space which we could decide on elsewhere.

masterleinad: and yes, I don't think it's a good idea to create a `Cuda` execution space instance from within `CudaInternal`.

masterleinad:
It should be enough to do
```diff
diff --git a/core/src/Cuda/Kokkos_Cuda_Instance.cpp b/core/src/Cuda/Kokkos_Cuda_Instance.cpp
index f439f4fd0..78064b0e5 100644
--- a/core/src/Cuda/Kokkos_Cuda_Instance.cpp
+++ b/core/src/Cuda/Kokkos_Cuda_Instance.cpp
@@ -428,21 +428,22 @@ void *CudaInternal::resize_team_scratch_space(int scratch_pool_id,
   // Multiple ParallelFor/Reduce Teams can call this function at the same time
   // and invalidate the m_team_scratch_ptr. We use a pool to avoid any race
   // condition.
+  auto mem_space = Kokkos::CudaSpace::impl_create(m_cudaDev, m_stream);
   if (m_team_scratch_current_size[scratch_pool_id] == 0) {
     m_team_scratch_current_size[scratch_pool_id] = bytes;
     m_team_scratch_ptr[scratch_pool_id] =
-        Kokkos::kokkos_malloc<Kokkos::CudaSpace>(
+        mem_space.allocate(
             "Kokkos::CudaSpace::TeamScratchMemory",
             m_team_scratch_current_size[scratch_pool_id]);
   }
   if ((bytes > m_team_scratch_current_size[scratch_pool_id]) ||
       ((bytes < m_team_scratch_current_size[scratch_pool_id]) &&
        (force_shrink))) {
-    m_team_scratch_current_size[scratch_pool_id] = bytes;
-    m_team_scratch_ptr[scratch_pool_id] =
-        Kokkos::kokkos_realloc<Kokkos::CudaSpace>(
-            m_team_scratch_ptr[scratch_pool_id],
-            m_team_scratch_current_size[scratch_pool_id]);
+     mem_space.deallocate(m_team_scratch_ptr[scratch_pool_id],
+        m_team_scratch_current_size[scratch_pool_id]);
+     m_team_scratch_current_size[scratch_pool_id] = bytes;
+     m_team_scratch_ptr[scratch_pool_id] = mem_space.allocate(
+        "Kokkos::CudaSpace::TeamScratchMemory", bytes);
   }
   return m_team_scratch_ptr[scratch_pool_id];
 }
@@ -475,8 +476,9 @@ void CudaInternal::finalize() {
   }
 
   for (int i = 0; i < m_n_team_scratch; ++i) {
+    auto cuda_mem_space = Kokkos::CudaSpace::impl_create(m_cudaDev, m_stream);
     if (m_team_scratch_current_size[i] > 0)
-      Kokkos::kokkos_free<Kokkos::CudaSpace>(m_team_scratch_ptr[i]);
+      cuda_mem_space.deallocate(m_team_scratch_ptr[i], m_team_scratch_current_size[i]);
   }
 
   m_scratchSpaceCount   = 0;
```
which is what I would propose instead. I'm happy to open a different pull request if we want to compare the two options.

tcclevenger:
@masterleinad I like your suggestion much better. Didn't realize we didn't need the Kokkos_SharedAlloc tracking stuff. I'll make the change here and maybe add a few more tests and set it ready for review. Thanks!

masterleinad (approved):

core/unit_test/cuda/TestCuda_InterOp_StreamsMultiGPU.cpp
```diff
@@ -168,4 +168,108 @@ TEST(cuda_multi_gpu, unmanaged_views) {
...
+  Kokkos::parallel_for("test_scratch_device_1", policy1, f1);
+
+  Kokkos::fence();
+  int64_t error0, error1;
+  Kokkos::deep_copy(error0, f0.error_count);
+  Kokkos::deep_copy(error1, f1.error_count);
+  ASSERT_EQ(error0, 0);
+  ASSERT_EQ(error1, 0);
+
+  // Request larget scratch size to trigger a realloc and test
```
ldh4, line 240:
```suggestion
  // Request largest scratch size to trigger a realloc and test
```

ldh4 (commented):
> Add overloads for Kokkos::kokkos_malloc, Kokkos::kokkos_free, and Kokkos::kokkos_realloc which take in an execution space to pass to Kokkos_SharedAlloc.hpp classes

Are the added overloads in this PR?

tcclevenger:
@ldh4 I forgot to update the description. From Daniels comments, we went a different route (not using `Kokkos_SharedAlloc.hpp`). Update the description now.

core/unit_test/cuda/TestCuda_InterOp_StreamsMultiGPU.cpp
```diff
@@ -168,4 +168,108 @@ TEST(cuda_multi_gpu, unmanaged_views) {
     KOKKOS_IMPL_CUDA_SAFE_CALL(cudaFree(p));
   }
 }
+
+struct ScratchFunctor {
+  int scratch_size;
+  int R;
+  Kokkos::View<int64_t, Kokkos::Cuda, Kokkos::MemoryTraits<Kokkos::Atomic>>
+      error_count;
```
dalg24, line 176: I would prefer if you did a reduction rather than an atomic error counter.

dalg24 (commented):
Not blocking but thinking that using a reduction would simplify your test.
The fix looks good.  Good catch.

tcclevenger:
Some failures in Cuda test. Investigating now.

ldh4 (approved):

dalg24:
Retest this please

tcclevenger:
@dalg24  All Cuda tests passing, I believe all fails are unrelated.

PR #6868, opened 2024-03-11
Add HIP specialization for sort-by-key

Rombur:
The new code is almost identical to the CUDA version. I just needed to replace CUDA with HIP to make it work.


masterleinad (approved):

dalg24 (approved):

dalg24:
Ignoring
```
```
in fedora rawhide builds

PR #6870, opened 2024-03-11
SYCL: Make sure to call find_dependency for oneDPL if necessary 

masterleinad:
Without `KOKKOS_EXPORT_CMAKE_TPL(oneDPL)` we are missing finding `oneDPL` in downstream code in case `oneDPL` is not included in the `oneAPI` installation used for the `SYCL` backend but is installed as CMake package elsewhere (which is uncommon). This pull request is fixing a bug in the exported configuration for this case.
Also, move linking with `oneDPL` to `algorithms`.

Support for `oneDPL` was initially added in https://github.com/kokkos/kokkos/pull/5229 (Kokkos 4.1.00).


dalg24 (approved):
Please spell out (in the description) when support for oneDPL was added and make it clear this is resolving a bug in the exported configuration.

Rombur (approved):

dalg24:
Please add to the 4.3 changelog

PR #6871, opened 2024-03-11
Fix linking with rothrust in downstream applications

masterleinad:
Without this, we get link errors like
```
CMake Error at /opt/cmake/share/cmake-3.16/Modules/CMakeFindDependencyMacro.cmake:47 (find_package):
  By not providing "FindROCTHRUST.cmake" in CMAKE_MODULE_PATH this project
  has asked CMake to find a package configuration file provided by
  "ROCTHRUST", but CMake did not find one.

  Could not find a package configuration file provided by "ROCTHRUST" with
  any of the following names:

    ROCTHRUSTConfig.cmake
    rocthrust-config.cmake

  Add the installation prefix of "ROCTHRUST" to CMAKE_PREFIX_PATH or set
  "ROCTHRUST_DIR" to a directory containing one of the above files.  If
  "ROCTHRUST" provides a separate development package or SDK, be sure it has
  been installed.
  /app/kokkos-install-hip/lib/cmake/Kokkos/KokkosConfig.cmake:39 (FIND_DEPENDENCY)
  CMakeLists.txt:10 (find_package)


See also "/app/kokkos/example/build_cmake_installed/build/CMakeFiles/CMakeOutput.log".
See also "/app/kokkos/example/build_cmake_installed/build/CMakeFiles/CMakeError.log".
```
in downstream code if `rocthrust` is not part of the `rocm` installation but installed separately.

edit: This doesn't fix the issue in tools, but https://github.com/kokkos/kokkos-tools/pull/244 does.


masterleinad:
Without the last commit `Fortran` can't be found in the CI. I don't think we need to include `Fortran` in all of the examples.

masterleinad:
I'm fine with removing the last two commits if we don't want to make those changes to `build_cmake_installed`.

cmake/Modules/FindTPLROCTHRUST.cmake
```diff
@@ -12,4 +12,4 @@ FIND_PACKAGE(rocthrust REQUIRED)
 KOKKOS_CREATE_IMPORTED_TPL(ROCTHRUST INTERFACE LINK_LIBRARIES roc::rocthrust)
 
 # Export ROCTHRUST as a Kokkos dependency
-KOKKOS_EXPORT_CMAKE_TPL(ROCTHRUST)
+KOKKOS_EXPORT_CMAKE_TPL(rocthrust)
```
Rombur, line 15: I don't understand this change. We should export the TPL that we import. I tested the current code with ArborX when I created the PR and it's working fine.

Rombur: This change also works with ArborX. Are we supposed to export the TPL that we created or the underlying library?

masterleinad: I noticed the problem first when working on https://github.com/kokkos/kokkos/pull/6870. If I used `KOKKOS_EXPORT_CMAKE_TPL` with the new TPL created by `Kokkos`, `find_dependency` didn't have a hint where to find the underlying library the interface target tried to link with. Basically, `find_dependency` here would call `find_package(ROCTHRUST)` with the `ROCTHRUST_DIR` set (which doesn't exist since it's an interface target). These changes should only make a difference if the `ROCm` installation doesn't define the `roc::rocthrust` target already and you need to provide hints. Also, we should try to find `rocprim` and link to it, see https://github.com/ROCm/rocThrust?tab=readme-ov-file#using-rocthrust-in-a-project.

cmake/Modules/FindTPLROCTHRUST.cmake
```diff
@@ -8,8 +8,10 @@
 # the values are not cached, FIND_PACKAGE(rocthrust) will overwrite them.
 SET(AMDGPU_TARGETS "" CACHE STRING "AMD GPU targets to compile for")
 SET(GPU_TARGETS "" CACHE STRING "GPU targets to compile for")
+FIND_PACKAGE(rocprim REQUIRED)
```
Rombur, line 11: If you need this, your installation of ROCm is probably messed up.

Rombur (approved):

dalg24 (approved):

dalg24:
Both HIP builds passed

ndellingwood:
@masterleinad 
I tested Trilinos with this merged and I am running into configuration errors:
```
20:07:16 CMake Error at kokkos/cmake/kokkos_tpls.cmake:29 (MESSAGE):
20:07:16   Enabled TPL ROCTHRUST inside TriBITS build, but this can only be enabled in
20:07:16   a standalone build
20:07:16 Call Stack (most recent call first):
20:07:16   kokkos/cmake/kokkos_tpls.cmake:43 (KOKKOS_TPL_OPTION)
20:07:16   kokkos/cmake/kokkos_tribits.cmake:208 (INCLUDE)
20:07:16   kokkos/CMakeLists.txt:226 (KOKKOS_SETUP_BUILD_ENVIRONMENT)
```

I didn't add any cmake config options to enable ROCTHRUST, any suggestions why this may be getting triggered?

ndellingwood:
I tried this change which allowed me to configure Trilinos, not sure if this is the right path but I'll put in a PR to start discussion
```diff
diff --git a/cmake/kokkos_tpls.cmake b/cmake/kokkos_tpls.cmake
index 3998eaa9b..42416117c 100644
--- a/cmake/kokkos_tpls.cmake
+++ b/cmake/kokkos_tpls.cmake
@@ -36,11 +36,13 @@ KOKKOS_TPL_OPTION(CUDA    ${Kokkos_ENABLE_CUDA} TRIBITS CUDA)
 IF(KOKKOS_ENABLE_HIP AND NOT KOKKOS_CXX_COMPILER_ID STREQUAL HIPCC AND NOT
     KOKKOS_HAS_TRILINOS)
   SET(ROCM_DEFAULT ON)
+  SET(ROCTHRUST_DEFAULT ON)
 ELSE()
   SET(ROCM_DEFAULT OFF)
+  SET(ROCTHRUST_DEFAULT OFF)
 ENDIF()
 KOKKOS_TPL_OPTION(ROCM    ${ROCM_DEFAULT})
-KOKKOS_TPL_OPTION(ROCTHRUST ${KOKKOS_ENABLE_HIP})
+KOKKOS_TPL_OPTION(ROCTHRUST ${ROCTHRUST_DEFAULT})
 
 IF(KOKKOS_ENABLE_SYCL AND NOT KOKKOS_HAS_TRILINOS)
   SET(ONEDPL_DEFAULT ON)
```

dalg24:
Having rocThrust can be a big deal for the performance of Kokkos::sort

ndellingwood:
@dalg24 yeah, definitely preferable to have rocthrust available. I opened #6879 as a heavy-handed workaround option for now, not the desired route to go but I'm not certain what is necessary to make it work

masterleinad:
> I didn't add any cmake config options to enable ROCTHRUST, any suggestions why this may be getting triggered?

Our current `CMake` system with `TriBITS` supports only allows `TPLs` in `Kokkos` that correspond to something in `Trilinos`. https://github.com/kokkos/kokkos/pull/6164 would get rid of this restriction.
Depending on how important support for `rocThrust` is we could just avoid using the current wrappers and do the more `CMake` way (which will be pretty hacky without #6164).

masterleinad:
Also, this pull request should not have changed anything for `Trilinos` but https://github.com/kokkos/kokkos/pull/6793.

ndellingwood:
@masterleinad sorry for the miscategorizing the issue occurrence, our Hip CI was down and I mistakenly picked out this PR over #6793 when working back through the triggering of the config issue

PR #6864, opened 2024-03-07
Tailor Existing Kokkos "build_cmake_installed" Example for Quick Start 

ajpowelsnl:
This PR proposes updates to the "build_cmake_installed" example to integrate with the Quick Start (documentation).

In the future, we would like to change the "build_cmake_installed" directory name to "quick_start_example", or suchlike.

Here are the steps for the updated example:

- Configure / build / install Kokkos (nota bene:  Kokkos build recipe will be the modern CMake version in the "Quick Start"; it is necessary to turn CMake extensions off in the Kokkos build to avoid the `find_package` warning in the example)
```
cmake \
-DKokkos_ENABLE_CUDA=ON \
-DKokkos_ARCH_VOLTA70=ON \
-DKokkos_ENABLE_SERIAL=ON \
-DCMAKE_EXTENSIONS=OFF \
-DCMAKE_INSTALL_PREFIX=$INSTALL_DIR \
-DCMAKE_BUILD_TYPE=Release ..
```

- Configure example

```
cmake -B <build-dir-name> -S . -DKokkos_ROOT=<kokkos-installation>
```
- Build example

```
cmake --build <build-dir-name>
```

- Run example
```
cd  <build-dir-name>
$ ./cmake_example 
Device Execution Space:
  KOKKOS_ENABLE_CUDA: yes
Cuda Options:
  KOKKOS_ENABLE_CUDA_LAMBDA: yes
  KOKKOS_ENABLE_CUDA_LDG_INTRINSIC: yes
  KOKKOS_ENABLE_CUDA_RELOCATABLE_DEVICE_CODE: no
  KOKKOS_ENABLE_CUDA_UVM: no
  KOKKOS_ENABLE_CXX11_DISPATCH_LAMBDA: yes
  KOKKOS_ENABLE_IMPL_CUDA_MALLOC_ASYNC: yes

Cuda Runtime Configuration:
macro  KOKKOS_ENABLE_CUDA      : defined
macro  CUDA_VERSION          = 12000 = version 12.0
Kokkos::Cuda[ 0 ] Tesla V100-PCIE-16GB capability 7.0, Total Global Memory: 15.77 G, Shared Memory per Block: 48 K : Selected
Kokkos::Cuda[ 1 ] Tesla V100-PCIE-16GB capability 7.0, Total Global Memory: 15.77 G, Shared Memory per Block: 48 K
Usage: ./cmake_example [<kokkos_options>] <size>
```

PR #6873, opened 2024-03-12
Try fixing Fortran linking in build_cmake_installed for Fedora rawhide

masterleinad:
Our GitHub CI started failing with linking errors like
```
```
and this is not the first time the fedora rawhide build has caused problems.

Also see https://github.com/llvm/llvm-project/pull/74139.


ajpowelsnl:
#6864 is the primary PR.  Could we please fix that one up?

example/build_cmake_installed/CMakeLists.txt
```diff
@@ -10,8 +10,10 @@ project(Example CXX Fortran)
 find_package(Kokkos REQUIRED)
 
 add_executable(example cmake_example.cpp foo.f)
+MESSAGE(STATUS "XXX COMPILER_ID: ${CMAKE_Fortran_COMPILER_ID}")
```
dalg24, line 13: In case that helps, it looks like the `LLVMFlang` was only added in CMake 3.24 https://cmake.org/cmake/help/v3.24/variable/CMAKE_LANG_COMPILER_ID.html

masterleinad:
Supersede by #6982.

PR #6876, opened 2024-03-13
Don't use Fedora development version in GitHub CI

masterleinad:
Our GitHub CI started failing with linking errors like
```
```
a couple days ago. and this is not the first time the fedora rawhide build has caused problems. Note that this is not a stable but a development version. This pull request suggests removing testing with this development version and only considering stable OSs.

### DIFF
--- a/.github/workflows/continuous-integration-workflow.yml
+++ b/.github/workflows/continuous-integration-workflow.yml
@@ -20,7 +20,7 @@ jobs:
     continue-on-error: true
     strategy:
       matrix:
-        distro: ['fedora:latest', 'fedora:rawhide', 'ubuntu:latest']
+        distro: ['fedora:latest', 'ubuntu:latest']
         cxx: ['g++', 'clang++']
         cxx_extra_flags: ['']
         cmake_build_type: ['Release', 'Debug']


nmm0:
We may also want to pin a specific version rather than using `:latest`, as the version can change between CI runs

masterleinad:
> We may also want to pin a specific version rather than using `:latest`, as the version can change between CI runs

I think it's kind of the point that we always want to test with the latest and don't have to care about updating a pinned version.

nmm0:
> > We may also want to pin a specific version rather than using `:latest`, as the version can change between CI runs
> 
> I think it's kind of the point that we always want to test with the latest and don't have to care about updating a pinned version.

We run the danger of the tested platform breaking out from under us; i.e. if there is an incompatibility with a new version of an OS it will starting failing CI jobs that may have nothing to do with the platform version. If we instead pin a version and create a PR to change to a newer version when needed, we can contain any changes in an isolated PR and random people's CI jobs won't break.

ajpowelsnl:
> > > We may also want to pin a specific version rather than using `:latest`, as the version can change between CI runs
> > 
> > 
> > I think it's kind of the point that we always want to test with the latest and don't have to care about updating a pinned version.
> 
> We run the danger of the tested platform breaking out from under us; i.e. if there is an incompatibility with a new version of an OS it will starting failing CI jobs that may have nothing to do with the platform version. If we instead pin a version and create a PR to change to a newer version when needed, we can contain any changes in an isolated PR and random people's CI jobs won't break.

Isn't there also risk (to Kokkos) in looking backwards (i.e., to a pinned version) vs. forward (either to `latest` / `stable release`), plus, adjusting versions "by hand" is a bit of trap.

Rombur (approved):

nmm0 (approved):

nmm0:
> > > > We may also want to pin a specific version rather than using `:latest`, as the version can change between CI runs
> > > 
> > > 
> > > I think it's kind of the point that we always want to test with the latest and don't have to care about updating a pinned version.
> > 
> > 
> > We run the danger of the tested platform breaking out from under us; i.e. if there is an incompatibility with a new version of an OS it will starting failing CI jobs that may have nothing to do with the platform version. If we instead pin a version and create a PR to change to a newer version when needed, we can contain any changes in an isolated PR and random people's CI jobs won't break.
> 
> Isn't there also risk (to Kokkos) in looking backwards (i.e., to a pinned version) vs. forward (either to `latest` / `stable release`), plus, adjusting versions "by hand" is a bit of trap. Do we have a general policy , in terms of OS for CI? If not, maybe something along the lines of using `latest` / `stable release` for the OS we test ?

@ajpowelsnl The problem with `latest` is that it will propagate failures to unrelated PRs. By all means we should test latest versions in a separate PR as they come out. I'm speaking of experience from other projects as our CI kept breaking because system packages changed out from under us. It's also the whole reason why this CI failure came up in the first place.

dalg24:
Please comment on the current failures
Are they related?

masterleinad:
> Are they related?

The other failures in this pull request are not related AFAICT. I mean I'm only removing a build with errors that don't show up anywhere else.

masterleinad:
Also, see #6873 for tracking the issue that is side-stepped here.

PR #6877, opened 2024-03-13
MI300 support unified memory support

crtrott:
This provides a (for now) implementation only opt-in to leveraging unified memory by making HIPSpace host accessible.


crtrott:
Retest this please

Rombur:
Retest this please

Rombur:
There was an issue with the machine, I had to reboot it. Now you have a real issue with clang-format.

algorithms/src/sorting/impl/Kokkos_SortImpl.hpp
```diff
@@ -399,9 +399,14 @@ sort_device_view_with_comparator(
 
   using ViewType = Kokkos::View<DataType, Properties...>;
   using MemSpace = typename ViewType::memory_space;
+// Note with HIP unified memory this code path is still the right thing to do
+// We don't have thrust in yet, and the create_mirror_view_and_copy will do
+// the right thing.
```
masterleinad, line 404: We do have `rocthrust` support in `develop`.

core/src/HIP/Kokkos_HIP_SharedAllocationRecord.hpp
```diff
@@ -20,8 +20,12 @@
 #include <HIP/Kokkos_HIP_Space.hpp>
 #include <impl/Kokkos_SharedAlloc.hpp>
 
+#if !defined(KOKKOS_ENABLE_IMPL_HIP_UNIFIED_MEMORY)
 KOKKOS_IMPL_HOST_INACCESSIBLE_SHARED_ALLOCATION_SPECIALIZATION(
     Kokkos::HIPSpace);
+#else
+KOKKOS_IMPL_SHARED_ALLOCATION_SPECIALIZATION(Kokkos::HIPSpace);
+#endif
```
masterleinad, line 28:
```suggestion
#if defined(KOKKOS_ENABLE_IMPL_HIP_UNIFIED_MEMORY)
KOKKOS_IMPL_SHARED_ALLOCATION_SPECIALIZATION(Kokkos::HIPSpace);
#else
KOKKOS_IMPL_HOST_INACCESSIBLE_SHARED_ALLOCATION_SPECIALIZATION(
    Kokkos::HIPSpace);
#endif
```
for readability.

core/src/HIP/Kokkos_HIP_Space.hpp
```diff
@@ -65,6 +65,18 @@ class HIPSpace {
...
+  void* allocate(const ExecutionSpace&, const size_t arg_alloc_size) const {
+    return allocate(arg_alloc_size);
+  }
+  template <typename ExecutionSpace>
+  void* allocate(const ExecutionSpace&, const char* arg_label,
+                 const size_t arg_alloc_size,
+                 const size_t arg_logical_size = 0) const {
+    return allocate(arg_label, arg_alloc_size, arg_logical_size);
+  }
+#else
```
masterleinad, line 79:
```suggestion
#endif
```
Going forward, it still makes sense to treat the overload with a `HIP` execution space instance special.

core/src/HIP/Kokkos_HIP_Space.hpp
```diff
@@ -74,6 +86,7 @@ class HIPSpace {
                  const size_t arg_logical_size = 0) const {
     return allocate(arg_label, arg_alloc_size, arg_logical_size);
   }
+#endif
```
masterleinad, line 89:
```suggestion
```

masterleinad (changes requested):

core/src/HIP/Kokkos_HIP_Space.hpp
crtrott: how so? It does the exact same thing as the one below?

masterleinad: Yes, at the moment, but we will need a separate overload when using `hipMallocAsync` later.

crtrott: Then lets introduce it when we need it.

algorithms/src/sorting/impl/Kokkos_SortImpl.hpp
Rombur: Yes, at the minimum the comment needs to be updated.

core/src/HIP/Kokkos_HIP.cpp
```diff
@@ -146,6 +146,10 @@ void HIP::print_configuration(std::ostream& os, bool /*verbose*/) const {
 #else
   os << "no\n";
 #endif
+#ifdef KOKKOS_ENABLE_IMPL_HIP_UNIFIED_MEMORY
+  os << "  KOKKOS_ENABLE_IMPL_HIP_UNIFIED_MEMORY: ";
+  os << "yes\n";
+#endif
```
Rombur, line 152:
```suggestion
 os << "  KOKKOS_ENABLE_IMPL_HIP_UNIFIED_MEMORY: ";
#ifdef KOKKOS_ENABLE_IMPL_HIP_UNIFIED_MEMORY
   os << "yes\n";
#else
   os << "no\n";
#endif
```

Rombur (commented):
Did anyone run this code on a MI300?

stanmoore1:
>Did anyone run this code on a MI300?

I did a couple weeks ago--seems to be working and it gave measurable speedup by making host/device transfers a no-op.

core/src/HIP/Kokkos_HIP.cpp
dalg24: Actually I like Christian's version better.
I would prefer if we would only print something when macros are defined and nothing otherwise, or at least that it looks closer to what the cmake-generated configuration file contains.  When we print more, we make the relevant part possibly less salient.
Maybe (not here) we could do something like only print undefined macros in verbose mode.
That said I don't want this discussion to delay progress on this work.

core/src/HIP/Kokkos_HIP_SharedAllocationRecord.hpp
dalg24: Agree with Daniel

algorithms/src/sorting/impl/Kokkos_SortImpl.hpp
crtrott: Do you have a suggestion? I.e. are you expecting we never got reach this thing?

crtrott, line 404:
```suggestion
// Thrust is optional but the create_mirror_view_and_copy will do
// the right thing if we reach this point with unified memory.
```

dalg24: Sounds good.  Maybe `Thrust -> rocThrust`

Rombur (approved):

masterleinad (approved):

PR #6887, opened 2024-03-19
SYCL: Fix multi-GPU support and add test

masterleinad:
This is analogous to https://github.com/kokkos/kokkos/pull/6866 and https://github.com/kokkos/kokkos/pull/6782 for `Cuda`. It's known to fail with `Intel GPUs` for some older compiler versions but works with the latest official release and on `polaris` with `A100`s.
The test added reuses and refactors the tests for CUDA multi-GPUs.

### DIFF
--- a/core/unit_test/cuda/TestCuda_InterOp_StreamsMultiGPU.cpp
+++ b/core/unit_test/cuda/TestCuda_InterOp_StreamsMultiGPU.cpp
@@ -15,7 +15,7 @@
 //@HEADER
 
 #include <TestCuda_Category.hpp>
-#include <Test_InterOp_Streams.hpp>
+#include <TestMultiGPU.hpp>
 
 namespace {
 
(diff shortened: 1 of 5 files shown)


core/unit_test/cuda/TestCuda_InterOp_StreamsMultiGPU.cpp
dalg24, line 1: Did you mean to commit this?

core/src/SYCL/Kokkos_SYCL_Instance.hpp
```diff
@@ -53,7 +53,7 @@ class SYCLInternal {
   void register_team_scratch_event(int scratch_pool_id, sycl::event event);
 
   uint32_t impl_get_instance_id() const;
-  static int m_syclDev;
+  int m_syclDev;
```
dalg24, line 57: So what's the intended meaning of that now non-static device ID?
It looks like it is uninitialized for non-default instances.

dalg24 (commented):
I don't get the changes to `m_syclDev`.

It would be good if we avoid duplicating code in unit tests.

core/src/SYCL/Kokkos_SYCL_Instance.hpp
masterleinad: I just reverted those changes now.

masterleinad:
> I don't get the changes to `m_syclDev`.

Reverted.

> It would be good if we avoid duplicating code in unit tests.

See [`f433a74` (#6887)](https://github.com/kokkos/kokkos/pull/6887/commits/f433a74bb4364569a013bed99596c6c1617e3dbc).

ndellingwood:
@masterleinad testing this PR with intel/2023.1.0 on PVC (OneDPL disabled) was successful - 100% tests passed

masterleinad:
@ndellingwood Does the system you tested on have multiple GPUs?

ndellingwood:
@masterleinad no, only one GPU available per node

core/unit_test/sycl/TestSYCL_InterOp_StreamsMultiGPU.cpp
```diff
@@ -0,0 +1,78 @@
...
+  Kokkos::View<int *, TEST_EXECSPACE> view(Kokkos::view_alloc("v", execs[1]),
+                                           100);
+
+  test_policies(execs[0], view0, execs[1], view);
+}
+
+TEST(sycl_multi_gpu, unmanaged_views) {
+  std::array<TEST_EXECSPACE, 2> execs = get_execution_spaces();
+
+  int *p0 = sycl::malloc_device<int>(100, execs[0].sycl_queue());
```
dalg24, line 62: Would you remind me why we need to test both for managed and unmanaged views?

masterleinad: So we test interoperability with GPU memory pointers as well as allocations going through managed Views.
If only the one with managed Views fails, we have a good idea that something in the View allocation is problematic.

core/unit_test/sycl/TestSYCL_InterOp_StreamsMultiGPU.cpp
```diff
@@ -0,0 +1,78 @@
...
+        std::rethrow_exception(e);
+      } catch (sycl::exception const &e) {
+        std::cerr << e.what() << '\n';
+        asynchronous_error = true;
+      }
+    }
+    if (asynchronous_error)
+      Kokkos::Impl::throw_runtime_exception(
+          "There was an asynchronous SYCL error!\n");
+  };
```
dalg24, line 38: Why are we specifying an asynchronous error handler?
Any good reason not to just use the constructor that takes a device and a property list?

core/unit_test/sycl/TestSYCL_InterOp_StreamsMultiGPU.cpp
dalg24: I am not blocking but I feel like it is not worth the complexity of having both.

dalg24: In any case please update the description and mention the refactor to reuse the tests for CUDA multi-GPUs

core/unit_test/sycl/TestSYCL_InterOp_StreamsMultiGPU.cpp
masterleinad: It's just the default we are using for the internal queues. It doesn't really matter here.

dalg24: Then drop it

core/unit_test/sycl/TestSYCL_InterOp_StreamsMultiGPU.cpp
masterleinad: Done.

core/unit_test/sycl/TestSYCL_InterOp_StreamsMultiGPU.cpp
masterleinad: Here you go, see https://github.com/kokkos/kokkos/pull/6887/commits/0f04a5fca52a045a8f110255e05786f0e80ff798!

nliber (approved):

PR #6891, opened 2024-03-20
Use recommended/max team size functions in Cuda ParallelFor and Reduce constructors

### LINKED ISSUE #6814: [Bug] Parallel for on Cuda not calling team_size_recommended
**Describe the bug**

If the user doesn't provide the team size, the specialisation of `ParallelFor` on `Cuda` sets it as follows:

- https://github.com/kokkos/kokkos/blob/391f2d12f2d3e5f896d497a11d0752aa47be7277/core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp#L545-L552

This is quite similar but not exactly the same as what `team_size_recommended` does:

- https://github.com/kokkos/kokkos/blob/391f2d12f2d3e5f896d497a11d0752aa47be7277/core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp#L142-L148

It may also be worth noting that the documentation suggests that `ParalelFor` will call `team_size_recommended` (which it doesn't in this case).

- https://kokkos.org/kokkos-core-wiki/API/core/parallel-dispatch/ParallelForTag.html?highlight=team_size_recommended

And it may also be worth noting that on HIP, the behavior is more what would be expected:
- https://github.com/kokkos/kokkos/blob/e0dc0128e04f18c2bbbaefceef3616e7ddcfa3c4/core/src/HIP/Kokkos_HIP_Parallel_Team.hpp#L510-L520

It is not immediately clear why the constructor of `ParallelFor` doesn't call `team_size_recommended`, i.e. why there is duplication.  And it is not immediately clear why there are the differences, i.e. the sizes of the double that are added. 

Joint observations with @romintomasetti.

tcclevenger:
Fixes #6814. Previous computation of `m_team_size` in cuda `ParallelFor()` and `ParallelReduce()` was intended to match `policy.team_size_recommended()`, but
missing extra scratch space allocation. Easiest to use `team_size_recommended()` instead. This matches exactly what is done for HIP backend.

Also, we were manually implementing `policy.team_max_size()` to verify `m_team_size` was not too large in `ParallelFor()`. Use `team_max_size()` instead.

### DIFF
--- a/core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
@@ -542,14 +542,9 @@ class ParallelFor<FunctorType, Kokkos::TeamPolicy<Properties...>,
     cudaFuncAttributes attr =
         CudaParallelLaunch<ParallelFor, LaunchBounds>::get_cuda_func_attributes(
             internal_space_instance->m_cudaDev);
-    m_team_size =
-        m_team_size >= 0
-            ? m_team_size
-            : Kokkos::Impl::cuda_get_opt_block_size<FunctorType, LaunchBounds>(
-                  internal_space_instance, attr, m_functor, m_vector_size,
-                  m_policy.team_scratch_size(0),
-                  m_policy.thread_scratch_size(0)) /
-                  m_vector_size;
+    m_team_size = m_team_size >= 0 ? m_team_size
+                                   : arg_policy.team_size_recommended(
+                                         arg_functor, ParallelForTag());
 
     m_shmem_begin = (sizeof(double) * (m_team_size + 2));
     m_shmem_size =
--- a/core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
@@ -906,15 +906,11 @@ class ParallelReduce<CombinedFunctorReducerType,
         m_policy.space().impl_internal_space_instance();
     cudaFuncAttributes attr = CudaParallelLaunch<ParallelReduce, LaunchBounds>::
         get_cuda_func_attributes(internal_space_instance->m_cudaDev);
-    m_team_size =
-        m_team_size >= 0
-            ? m_team_size
-            : Kokkos::Impl::cuda_get_opt_block_size<FunctorType, LaunchBounds>(
-                  internal_space_instance, attr,
-                  m_functor_reducer.get_functor(), m_vector_size,
-                  m_policy.team_scratch_size(0),
-                  m_policy.thread_scratch_size(0)) /
-                  m_vector_size;
+    m_team_size = m_team_size >= 0 ? m_team_size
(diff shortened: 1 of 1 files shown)


core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
```diff
@@ -539,17 +539,9 @@ class ParallelFor<FunctorType, Kokkos::TeamPolicy<Properties...>,
...
-    m_team_size =
-        m_team_size >= 0
-            ? m_team_size
-            : Kokkos::Impl::cuda_get_opt_block_size<FunctorType, LaunchBounds>(
-                  internal_space_instance, attr, m_functor, m_vector_size,
-                  m_policy.team_scratch_size(0),
-                  m_policy.thread_scratch_size(0)) /
-                  m_vector_size;
+    m_team_size = m_team_size >= 0 ? m_team_size
+                                   : arg_policy.team_size_recommended(
```
romintomasetti, line 543: The changes looks good overall ! (I'm always happier with code reuse :+1:)

However, I wonder about one thing that this PR will change.

I think https://github.com/kokkos/kokkos/blob/9fff1e066a7c3f2759d251e327fc08963f4f8c2b/core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp#L144-L147 will set the team and thread scratch sizes regardless of whether it is a parallel for or a parallel reduce.

But I guess that for a parallel for, there is no need for the tiny additional scratch size.

romintomasetti: A possible fix would be to modify the recommended team size function along these lines:
```c++
template <class FunctorType, class ParallelTagType>
  int team_size_recommended(const FunctorType& f, const ParallelTagType&) const {
    using closure_type =
        Impl::ParallelFor<FunctorType, TeamPolicy<Properties...>>;
    cudaFuncAttributes attr =
        CudaParallelLaunch<closure_type, typename traits::launch_bounds>::
            get_cuda_func_attributes(space().cuda_device());
    const int block_size =
        Kokkos::Impl::cuda_get_opt_block_size<FunctorType,
                                              typename traits::launch_bounds>(
            space().impl_internal_space_instance(), attr, f,
            (size_t)impl_vector_length(),
            (size_t)team_scratch_size(0) + is it reduce ? 2 * sizeof(double) : 0,
            (size_t)thread_scratch_size(0) + is it reduce ? sizeof(double) : 0);
    return block_size / impl_vector_length();
  }
```

romintomasetti (changes requested):
One major concern...

masterleinad (approved):

core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
crtrott: What Daniel said.

crtrott (approved):

ndellingwood:
Does this merit a changelog entry for 4.4 https://github.com/kokkos/kokkos/issues/6914 ?

masterleinad:

done

PR #6892, opened 2024-03-20
Fix a bug in Makefile when using AMD GPU architectures

Rombur:
There is a bug in the Makefile.kokkos when we try to use `AMD_GFXYYY` for the architecture. The issue is that the `or` we are using does not work like I expected. From the documentation "Each argument is expanded, in order. If an argument expands to a non-empty string the processing stops and the result of the expansion is that string." We ended up evaluating only the first argument. This PR fixes the issue.

The bug was introduced in 4.2. We should cherry-pick this for 4.3 because the error message is really unhelpful.


Makefile.kokkos
```diff
@@ -400,13 +400,29 @@ ifeq ($(KOKKOS_INTERNAL_USE_ARCH_ZEN3), 0)
...
+ifeq ($(KOKKOS_INTERNAL_USE_ARCH_AMD_GFX90A), 0)
+  KOKKOS_INTERNAL_USE_ARCH_AMD_GFX90A := $(call kokkos_has_string,$(KOKKOS_ARCH),VEGA90A)
+endif
 KOKKOS_INTERNAL_USE_ARCH_AMD_GFX940 := $(call kokkos_has_string,$(KOKKOS_ARCH),AMD_GFX940)
 KOKKOS_INTERNAL_USE_ARCH_AMD_GFX942 := $(call kokkos_has_string,$(KOKKOS_ARCH),AMD_GFX942)
-KOKKOS_INTERNAL_USE_ARCH_AMD_GFX1030 := $(or $(call kokkos_has_string,$(KOKKOS_ARCH),NAVI1030),$(call kokkos_has_string,$(KOKKOS_ARCH),AMD_GFX1030))
-KOKKOS_INTERNAL_USE_ARCH_AMD_GFX1100 := $(or $(call kokkos_has_string,$(KOKKOS_ARCH),NAVI1100),$(call kokkos_has_string,$(KOKKOS_ARCH),AMD_GFX1100))
+KOKKOS_INTERNAL_USE_ARCH_AMD_GFX1030 := $(call kokkos_has_string,$(KOKKOS_ARCH),AMD_GFX1030)
+ifeq ($(KOKKOS_INTERNAL_USE_ARCH_AMD_GFX1030), 0)
+	KOKKOS_INTERNAL_USE_ARCH_AMD_GFX1030 := $(call kokkos_has_string,$(KOKKOS_ARCH),NAVI1030)
```
dalg24, line 420: Indentation is weird here and below

masterleinad (approved):

stanmoore1:
I confirm this fixes the issue I was seeing. @Rombur if possible could you fix the comment in the Makefile that gives options for `KOKKOS_ARCH`, the AMD and Intel GPUs need `AMD_` and `Intel_` prefix respectively, like this:

```
# AMD-GPUS: AMD_GFX906,AMD_GFX908,AMD_GFX90A,AMD_GFX940,AMD_GFX942,AMD_GFX1030,AMD_GFX1100
# AMD-CPUS: AMDAVX,Zen,Zen2,Zen3
# Intel-GPUs: Intel_Gen9,Intel_Gen11,Intel_Gen12LP,Intel_DG1,Intel_XeHP,Intel_PVC
```

Rombur:
@stanmoore1 Yes, I will fix the comments. 

@masterleinad The Intel architecture have the format `IntelFoo` except for `PVC` which is just `PVC` and not `IntelPVC`. Should I change it to match the other Intel architectures?

stanmoore1:
> @masterleinad The Intel architecture have the format `IntelFoo` except for `PVC` which is just `PVC` and not `IntelPVC`. Should I change it to match the other Intel architectures?

Ah I missed that. Can everything be made consistent between CMake and Makefile for `KOKKOS_ARCH`? It makes the documentation for LAMMPS/SPARTA a lot simpler. The Makefile options are not case sensitive we can keep the lowercase. In CMake the Intel options are: `INTEL_PVC`, can that be propagated to Makefile as well?

masterleinad:
> @masterleinad The Intel architecture have the format `IntelFoo` except for `PVC` which is just `PVC` and not `IntelPVC`. Should I change it to match the other Intel architectures?

I would just leave it at this point. With the current code, `IntelPVC` would also be recognized. I'm hesitant to break setups for people who are only specifying `PVC`.

stanmoore1:
> > @masterleinad The Intel architecture have the format `IntelFoo` except for `PVC` which is just `PVC` and not `IntelPVC`. Should I change it to match the other Intel architectures?
> 
> I would just leave it at this point. With the current code, `IntelPVC` would also be recognized. I'm hesitant to break setups for people who are only specifying `PVC`.

This breaks our assumption in the LAMMPS and SPARTA documentation that the CMake and Makefile options are the same (disregarding capitalization since the Makefile is case-insensitive), see https://docs.lammps.org/Build_extras.html#kokkos and https://github.com/kokkos/kokkos/blob/develop/cmake/kokkos_arch.cmake#L126 and

masterleinad:
> This breaks our assumption in the LAMMPS and SPARTA documentation that the CMake and Makefile options are the same (disregarding capitalization since the Makefile is case-insensitive), see [docs.lammps.org/Build_extras.html#kokkos](https://docs.lammps.org/Build_extras.html#kokkos) and [`develop`/cmake/kokkos_arch.cmake#L126](https://github.com/kokkos/kokkos/blob/develop/cmake/kokkos_arch.cmake?rgh-link-date=2024-03-22T15%3A03%3A28Z#L126).

But the CMake arch `IntelPVC` still works for the `Makefile`, right?

stanmoore1:
>But the CMake arch IntelPVC still works for the Makefile, right?

Perhaps, but is this documented anywhere? It's not listed here: https://kokkos.org/kokkos-core-wiki/keywords.html#keywords-enable-options

masterleinad:
> It's not listed here: [kokkos.org/kokkos-core-wiki/keywords.html#keywords-enable-options](https://kokkos.org/kokkos-core-wiki/keywords.html#keywords-enable-options)

That's all for `CMake` anyway. I don't know about `Makefile` documentation.
What I was going to suggest is that we accept both `PVC` and `IntelPVC` in the `Makefile` for backward compatibility but it turns out that the current `Makefile` code is doing that already. We are just trying to find `PVC` in the arch string and that is matched both for `PVC` and `IntelPVC`. :shrug:

stanmoore1:
> That's all for `CMake` anyway. I don't know about `Makefile` documentation. What I was going to suggest is that we accept both `PVC` and `IntelPVC` in the `Makefile` for backward compatibility but it turns out that the current `Makefile` code is doing that already. We are just trying to find `PVC` in the arch string and that is matched both for `PVC` and `IntelPVC`. 🤷

I don't think there is any documentation for Makefile 😂  Every other Kokkos Arch besides the Intel GPUs have the same value for CMAKE and Makefile, e.g. `AMD_GFX942`, `SKX`,  `HOPPER90`. It is really confusing to have something different for CMake vs Makefile.

Rombur:
> It is really confusing to have something different for CMake vs Makefile.

I agree. Unless there is a good reason, we should try to use the same name.

masterleinad:
> I agree. Unless there is a good reason, we should try to use the same name.

I agree and I'm saying that you can already use `Intel_PVC` in `Makefiles` as architecture.

stanmoore1:
Yes this is true for `PVC` but not for any other Intel GPU, e.g. `INTEL_GEN` won't work, it has to be `IntelGen`:

```
KOKKOS_INTERNAL_USE_ARCH_INTEL_GEN := $(call kokkos_has_string,$(KOKKOS_ARCH),IntelGen)
KOKKOS_INTERNAL_USE_ARCH_INTEL_GEN9 := $(call kokkos_has_string,$(KOKKOS_ARCH),IntelGen9)
KOKKOS_INTERNAL_USE_ARCH_INTEL_GEN11 := $(call kokkos_has_string,$(KOKKOS_ARCH),IntelGen11)
KOKKOS_INTERNAL_USE_ARCH_INTEL_GEN12LP := $(call kokkos_has_string,$(KOKKOS_ARCH),IntelGen12LP)
KOKKOS_INTERNAL_USE_ARCH_INTEL_DG1 := $(call kokkos_has_string,$(KOKKOS_ARCH),IntelDG1)
KOKKOS_INTERNAL_USE_ARCH_INTEL_XEHP := $(call kokkos_has_string,$(KOKKOS_ARCH),IntelXeHP)
KOKKOS_INTERNAL_USE_ARCH_INTEL_PVC := $(call kokkos_has_string,$(KOKKOS_ARCH),PVC)
```

masterleinad:
> Yes this is true for `PVC` but not for any other Intel GPU, e.g. `INTEL_GEN` won't work, it has to be `IntelGen`:

Yes, we should fix that.

ajpowelsnl (approved):
Looks good - I support standardizing (to the extent possible) architectures for CMake and the Makefile.

dalg24:
Plz cherry-pick into 4.3 RC branch

ndellingwood:
Cherry-pick PR #6899

PR #6895, opened 2024-03-22
Update Intel GPU architectures in Makefile

masterleinad:
Addresses https://github.com/kokkos/kokkos/pull/6892#issuecomment-2015629714.
We should be able to use the same architecture options for `Makefile` and `CMake` builds. For `Intel` GPUs these are currently different, though. The syntax for `CMake` is, e.g., `INTEL_XEHP` while we have `IntelXeHp` for the `Makefile`.
This pull request makes sure that both variants are recognized. Note that `INTEL_PVC` is special since we allowed only `PVC` in the Makefile previously which means that `PVC`, `IntelPVC`, and `Intel_PVC` are all already recognized as the same architectures. For the `Intel_GEN*` architectures we now also make sure that `INTEL_GEN9`, `INTEL_GEN11`, or `INTEL_GEN12LP` would not enable the JIT "architecture" `INTEL_GEN` as well.
Also, note that capitalization is ignored in the `Makefile`.

### DIFF
--- a/Makefile.kokkos
+++ b/Makefile.kokkos
@@ -15,7 +15,7 @@ KOKKOS_DEVICES ?= "Threads"
 # IBM:      Power8,Power9
 # AMD-GPUS: GFX906,GFX908,GFX90A,GFX940,GFX942,GFX1030,GFX1100
 # AMD-CPUS: AMDAVX,Zen,Zen2,Zen3
-# Intel-GPUs: Gen9,Gen11,Gen12LP,DG1,XeHP,PVC
+# Intel-GPUs: Intel_Gen,Intel_Gen9,Intel_Gen11,Intel_Gen12LP,Intel_DG1,Intel_XeHP,Intel_PVC
 KOKKOS_ARCH ?= ""
 # Options: yes,no
 KOKKOS_DEBUG ?= "no"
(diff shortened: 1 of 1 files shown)


stanmoore1:
Thank you @masterleinad!!

Makefile.kokkos
```diff
@@ -318,12 +318,39 @@ KOKKOS_INTERNAL_USE_ARCH_ICL := $(call kokkos_has_string,$(KOKKOS_ARCH),ICL)
...
+endif
 KOKKOS_INTERNAL_USE_ARCH_INTEL_DG1 := $(call kokkos_has_string,$(KOKKOS_ARCH),IntelDG1)
+ifeq ($(KOKKOS_INTERNAL_USE_ARCH_INTEL_DG1), 0)
+  KOKKOS_INTERNAL_USE_ARCH_INTEL_DG1 := $(call kokkos_has_string,$(KOKKOS_ARCH),Intel_DG1)
+endif
 KOKKOS_INTERNAL_USE_ARCH_INTEL_XEHP := $(call kokkos_has_string,$(KOKKOS_ARCH),IntelXeHP)
+ifeq ($(KOKKOS_INTERNAL_USE_ARCH_INTEL_XEHP), 0)
+  KOKKOS_INTERNAL_USE_ARCH_INTEL_XEHP := $(call kokkos_has_string,$(KOKKOS_ARCH),Intel_XeHP)
+endif
 KOKKOS_INTERNAL_USE_ARCH_INTEL_PVC := $(call kokkos_has_string,$(KOKKOS_ARCH),PVC)
```
ajpowelsnl, line 355: Would it make sense to standardize architectural specifications for the Makefile and CMake, such that [CMake settings](https://kokkos.org/kokkos-core-wiki/keywords.html) would be the same, (minus the `Kokkos_ARCH_` part of the variable name)?

ajpowelsnl: I reviewed the PR out-of-order, so it was only later that I appreciated the fact that you are standardizing.

Makefile.kokkos
```diff
@@ -318,12 +318,39 @@ KOKKOS_INTERNAL_USE_ARCH_ICL := $(call kokkos_has_string,$(KOKKOS_ARCH),ICL)
 KOKKOS_INTERNAL_USE_ARCH_ICX := $(call kokkos_has_string,$(KOKKOS_ARCH),ICX)
 KOKKOS_INTERNAL_USE_ARCH_SPR := $(call kokkos_has_string,$(KOKKOS_ARCH),SPR)
 
-KOKKOS_INTERNAL_USE_ARCH_INTEL_GEN := $(call kokkos_has_string,$(KOKKOS_ARCH),IntelGen)
 KOKKOS_INTERNAL_USE_ARCH_INTEL_GEN9 := $(call kokkos_has_string,$(KOKKOS_ARCH),IntelGen9)
+ifeq ($(KOKKOS_INTERNAL_USE_ARCH_INTEL_GEN9), 0)
+  KOKKOS_INTERNAL_USE_ARCH_INTEL_GEN9 := $(call kokkos_has_string,$(KOKKOS_ARCH),Intel_Gen9)
```
dalg24, line 323: Do we want to support both with and without the underscore?
Iff the answer is "yes", then consider adding a comment above that block to capture what the intent is

Makefile.targets
```diff
@@ -81,7 +81,7 @@ ifeq ($(KOKKOS_INTERNAL_USE_THREADS), 1)
 Kokkos_Threads_Instance.o: $(KOKKOS_CPP_DEPENDS) $(KOKKOS_PATH)/core/src/Threads/Kokkos_Threads_Instance.cpp
 	$(CXX) $(KOKKOS_CPPFLAGS) $(KOKKOS_CXXFLAGS) $(CXXFLAGS) -c $(KOKKOS_PATH)/core/src/Threads/Kokkos_Threads_Instance.cpp
 Kokkos_Threads_Spinwait.o: $(KOKKOS_CPP_DEPENDS) $(KOKKOS_PATH)/core/src/Threads/Kokkos_Threads_Spinwait.cpp
-	$(CXX) $(KOKKOS_CPPFLAGS) $(KOKKOS_CXXFLAGS) $(CXXFLAGS) -c $(KOKKOS_PATH)/core/src/Threads/Kokkos_Spinwait.cpp
+	$(CXX) $(KOKKOS_CPPFLAGS) $(KOKKOS_CXXFLAGS) $(CXXFLAGS) -c $(KOKKOS_PATH)/core/src/Threads/Kokkos_Threads_Spinwait.cpp
```
Rombur, line 84: I am fine with this change but add a comment about it in the PR description

example/make_buildlink/Makefile
```diff
@@ -1,6 +1,6 @@
-KOKKOS_DEVICES=OpenMP
-KOKKOS_CUDA_OPTIONS=enable_lambda
-KOKKOS_ARCH = "SNB,Kepler35"
+#KOKKOS_DEVICES=OpenMP
+#KOKKOS_CUDA_OPTIONS=enable_lambda
+#KOKKOS_ARCH = "SNB,Kepler35"
```
Rombur, line 3: Did you mean to commit this?

masterleinad: Removed.

Makefile.targets
masterleinad: This is #6896. Removed.

Rombur (approved):

dalg24 (approved):

ndellingwood:

masterleinad:

done

PR #6904, opened 2024-03-28
graph: allow access to native graph object

romintomasetti:
### Summary

The native graph objects should be accessible for advanced users.

### Description

For `Cuda`, `HIP` and `SYCL` backends, the specialization of `Kokkos::Impl::Graph` always has 2 backend-specific members that are of interest for advanced usage:
1. the *topological* graph (`cudaGraph_t`, `hipGraph_t`, `the too loooong SYCL command graph type`)
2. the *executable* graph (`cudaGraphExec_t`, and so on)

Since we cannot easily guarantee `constness` semantics, all getters are non-`const`. 

Usage of these backend objects include:
- export graph to `DOT` (this PR tests this)
- add nodes
- check whatever the vendor API allows

> :warning: Please note that for now, we don't test interoperability in the sense of adding nodes through the backend API. This will be done later to keep this PR small.

### DIFF
--- a/cmake/kokkos_arch.cmake
+++ b/cmake/kokkos_arch.cmake
@@ -763,6 +763,8 @@ if(KOKKOS_ENABLE_SYCL)
       compiler_specific_flags(DEFAULT -fsycl-device-code-split=off -DDESUL_SYCL_DEVICE_GLOBAL_SUPPORTED)
     endif()
   endif()
+
+  check_cxx_symbol_exists(SYCL_EXT_ONEAPI_GRAPH "sycl/sycl.hpp" KOKKOS_IMPL_HAVE_SYCL_EXT_ONEAPI_GRAPH)
 endif()
 
 set(CUDA_ARCH_ALREADY_SPECIFIED "")
--- a/core/src/Cuda/Kokkos_Cuda_Graph_Impl.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_Graph_Impl.hpp
@@ -206,6 +206,9 @@ struct GraphImpl<Kokkos::Cuda> {
         m_execution_space, _graph_node_kernel_ctor_tag{},
         aggregate_kernel_impl_t{});
   }
+
+  cudaGraph_t cuda_graph() { return m_graph; }
+  cudaGraphExec_t cuda_graph_exec() { return m_graph_exec; }
 };
 
 }  // end namespace Impl
--- a/core/src/HIP/Kokkos_HIP_Graph_Impl.hpp
+++ b/core/src/HIP/Kokkos_HIP_Graph_Impl.hpp
@@ -79,6 +79,9 @@ class GraphImpl<Kokkos::HIP> {
     KOKKOS_ENSURES(m_graph_exec);
   }
 
+  hipGraph_t hip_graph() { return m_graph; }
+  hipGraphExec_t hip_graph_exec() { return m_graph_exec; }
+
  private:
   Kokkos::HIP m_execution_space;
   hipGraph_t m_graph          = nullptr;
--- a/core/src/Kokkos_Graph.hpp
+++ b/core/src/Kokkos_Graph.hpp
@@ -97,6 +97,10 @@ struct [[nodiscard]] Graph {
   }
 
   void submit() const { submit(get_execution_space()); }
+
+  decltype(auto) native_graph();
+
+  decltype(auto) native_graph_exec();
 };
 
 // </editor-fold> end Graph }}}1
@@ -167,6 +171,42 @@ create_graph(Closure&& arg_closure) {
 
 // </editor-fold> end create_graph }}}1
 //==============================================================================
+
+template <class ExecutionSpace>
+decltype(auto) Graph<ExecutionSpace>::native_graph() {
+  KOKKOS_EXPECTS(bool(m_impl_ptr));
+#if defined(KOKKOS_ENABLE_CUDA)
+  if constexpr (std::is_same_v<ExecutionSpace, Kokkos::Cuda>) {
+    return m_impl_ptr->cuda_graph();
+  }
+#elif defined(KOKKOS_ENABLE_HIP) && defined(KOKKOS_IMPL_HIP_NATIVE_GRAPH)
+  if constexpr (std::is_same_v<ExecutionSpace, Kokkos::HIP>) {
+    return m_impl_ptr->hip_graph();
+  }
+#elif defined(KOKKOS_ENABLE_SYCL) && defined(SYCL_EXT_ONEAPI_GRAPH)
+  if constexpr (std::is_same_v<ExecutionSpace, Kokkos::SYCL>) {
+    return m_impl_ptr->sycl_graph();
+  }
+#endif
+}
+
+template <class ExecutionSpace>
+decltype(auto) Graph<ExecutionSpace>::native_graph_exec() {
+  KOKKOS_EXPECTS(bool(m_impl_ptr));
+#if defined(KOKKOS_ENABLE_CUDA)
+  if constexpr (std::is_same_v<ExecutionSpace, Kokkos::Cuda>) {
+    return m_impl_ptr->cuda_graph_exec();
+  }
+#elif defined(KOKKOS_ENABLE_HIP) && defined(KOKKOS_IMPL_HIP_NATIVE_GRAPH)
+  if constexpr (std::is_same_v<ExecutionSpace, Kokkos::HIP>) {
+    return m_impl_ptr->hip_graph_exec();
+  }
+#elif defined(KOKKOS_ENABLE_SYCL) && defined(SYCL_EXT_ONEAPI_GRAPH)
+  if constexpr (std::is_same_v<ExecutionSpace, Kokkos::SYCL>) {
+    return m_impl_ptr->sycl_graph_exec();
+  }
+#endif
+}
 
 }  // end namespace Experimental
 }  // namespace Kokkos
--- a/core/unit_test/CMakeLists.txt
+++ b/core/unit_test/CMakeLists.txt
@@ -749,6 +749,10 @@ if(Kokkos_ENABLE_CUDA)
   kokkos_add_executable_and_test(
     CoreUnitTest_CudaInterOpStreamsMultiGPU SOURCES UnitTestMainInit.cpp cuda/TestCuda_InterOp_StreamsMultiGPU.cpp
   )
+
+  kokkos_add_executable_and_test(
+    CoreUnitTest_CudaInterOpGraph SOURCES UnitTestMainInit.cpp cuda/TestCuda_InterOp_Graph.cpp
+  )
 endif()
 
 if(Kokkos_ENABLE_HIP)
@@ -771,6 +775,10 @@ if(Kokkos_ENABLE_HIP)
   kokkos_add_executable_and_test(CoreUnitTest_HIPInterOpInit SOURCES UnitTestMain.cpp hip/TestHIP_InterOp_Init.cpp)
   kokkos_add_executable_and_test(
     CoreUnitTest_HIPInterOpStreams SOURCES UnitTestMain.cpp hip/TestHIP_InterOp_Streams.cpp
+  )
+
+  kokkos_add_executable_and_test(
+    CoreUnitTest_HIPInterOpGraph SOURCES UnitTestMainInit.cpp hip/TestHIP_InterOp_Graph.cpp
   )
 endif()
 
@@ -810,6 +818,12 @@ if(Kokkos_ENABLE_SYCL)
   kokkos_add_executable_and_test(
     CoreUnitTest_SYCLInterOpStreamsMultiGPU SOURCES UnitTestMainInit.cpp sycl/TestSYCL_InterOp_StreamsMultiGPU.cpp
   )
+
+  if(KOKKOS_IMPL_HAVE_SYCL_EXT_ONEAPI_GRAPH)
+    kokkos_add_executable_and_test(
+      CoreUnitTest_SYCLInterOpGraph SOURCES UnitTestMainInit.cpp sycl/TestSYCL_InterOp_Graph.cpp
+    )
+  endif()
 endif()
 
 set(DEFAULT_DEVICE_SOURCES
--- a/core/unit_test/TestGraph.hpp
+++ b/core/unit_test/TestGraph.hpp
@@ -507,7 +507,18 @@ TEST_F(TEST_CATEGORY_FIXTURE(graph), force_global_launch) {
             false, {"Either the name or pointer or size did not match"}};
       }));
 
+  listen_tool_events(Config::DisableAll());
 #endif
+}
+
+// Ensure that an empty graph on the default host execution space
+// can be submitted.
+TEST_F(TEST_CATEGORY_FIXTURE(graph), empty_graph_default_host_exec) {
+  auto graph =
+      Kokkos::Experimental::create_graph(Kokkos::DefaultHostExecutionSpace{});
+  graph.instantiate();
+  graph.submit();
+  graph.get_execution_space().fence();
 }
 
 template <typename ViewType, size_t TargetIndex, size_t NumIndices = 0>
--- a/core/unit_test/cuda/TestCuda_InterOp_Graph.cpp
+++ b/core/unit_test/cuda/TestCuda_InterOp_Graph.cpp
@@ -1,0 +1,143 @@
+//@HEADER
+// ************************************************************************
+//
+//                        Kokkos v. 4.0
+//       Copyright (2022) National Technology & Engineering
+//               Solutions of Sandia, LLC (NTESS).
+//
+// Under the terms of Contract DE-NA0003525 with NTESS,
+// the U.S. Government retains certain rights in this software.
+//
+// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
+#include <filesystem>
+#include <fstream>
+#include <regex>
+
+#include <TestCuda_Category.hpp>
+#include <Kokkos_Core.hpp>
+#include <Kokkos_Graph.hpp>
+
+#include <gtest/gtest.h>
+
+namespace {
+
+template <typename ViewType>
+struct Increment {
+  ViewType data;
+
+  KOKKOS_FUNCTION
+  void operator()(const int) const { ++data(); }
+};
+
+class TEST_CATEGORY_FIXTURE(GraphInterOp) : public ::testing::Test {
+ public:
+  using execution_space = Kokkos::Cuda;
+  using view_t =
+      Kokkos::View<int, execution_space, Kokkos::MemoryTraits<Kokkos::Atomic>>;
+  using graph_t = Kokkos::Experimental::Graph<execution_space>;
+
+  void SetUp() override {
+    data = view_t(Kokkos::view_alloc(exec, "witness"));
+
+    graph = Kokkos::Experimental::create_graph(exec, [&](const auto& root) {
+      root.then_parallel_for(1, Increment<view_t>{data});
+    });
+  }
+
+ protected:
+  execution_space exec{};
+  view_t data;
+  std::optional<graph_t> graph;
+};
+
+// This test checks the promises of Kokkos::Graph against its
+// underlying Cuda native objects.
+TEST_F(TEST_CATEGORY_FIXTURE(GraphInterOp), promises_on_native_objects) {
+  // Before instantiation, the Cuda graph is valid, but the Cuda executable
+  // graph is still null.
+  cudaGraph_t cuda_graph = graph->native_graph();
+
+  ASSERT_NE(cuda_graph, nullptr);
+  ASSERT_EQ(graph->native_graph_exec(), nullptr);
+
+  // After instantiation, both native objects are valid.
+  graph->instantiate();
+
+  cudaGraphExec_t cuda_graph_exec = graph->native_graph_exec();
+
+  ASSERT_EQ(graph->native_graph(), cuda_graph);
+  ASSERT_NE(cuda_graph_exec, nullptr);
+
+  // Submission should not affect the underlying objects.
+  graph->submit();
+
+  ASSERT_EQ(graph->native_graph(), cuda_graph);
+  ASSERT_EQ(graph->native_graph_exec(), cuda_graph_exec);
+}
+
+// Count the number of nodes. This is useful to ensure no spurious
+// (possibly empty) node is added.
+TEST_F(TEST_CATEGORY_FIXTURE(GraphInterOp), count_nodes) {
+  graph->instantiate();
+
+  size_t num_nodes;
+
+  KOKKOS_IMPL_CUDA_SAFE_CALL(
+      cudaGraphGetNodes(graph->native_graph(), nullptr, &num_nodes));
+
+  ASSERT_EQ(num_nodes, 2u);
+}
+
+// Use native Cuda graph to generate a DOT representation.
+TEST_F(TEST_CATEGORY_FIXTURE(GraphInterOp), debug_dot_print) {
+#if CUDA_VERSION < 11600
+  GTEST_SKIP() << "Export a graph to DOT requires Cuda 11.6.";
+#else
+  graph->instantiate();
+
+  const auto dot = std::filesystem::temp_directory_path() / "cuda_graph.dot";
+
+  // Convert path to string then to const char * to make it work on Windows.
+  KOKKOS_IMPL_CUDA_SAFE_CALL(
+      cudaGraphDebugDotPrint(graph->native_graph(), dot.string().c_str(),
+                             cudaGraphDebugDotFlagsVerbose));
+
+  ASSERT_TRUE(std::filesystem::exists(dot));
+  ASSERT_GT(std::filesystem::file_size(dot), 0u);
+
+  // We could write a check against the full kernel's function signature, but
+  // it would make the test rely too much on internal implementation details.
+  // Therefore, we just look for the functor and policy. Note that the
+  // signature is mangled in the 'dot' output.
+  const std::string expected("[A-Za-z0-9_]+Increment[A-Za-z0-9_]+RangePolicy");
+
+  std::stringstream buffer;
+  buffer << std::ifstream(dot).rdbuf();
+
+  ASSERT_TRUE(std::regex_search(buffer.str(), std::regex(expected)))
+      << "Could not find expected signature regex " << std::quoted(expected)
+      << " in " << dot;
+#endif
+}
+
+// Ensure that the graph has been instantiated with the default flag.
+TEST_F(TEST_CATEGORY_FIXTURE(GraphInterOp), instantiation_flags) {
+#if CUDA_VERSION < 12000
+  GTEST_SKIP() << "Graph instantiation flag inspection requires Cuda 12.";
+#else
+  graph->instantiate();
+  unsigned long long flags =
+      Kokkos::Experimental::finite_max_v<unsigned long long>;
+  KOKKOS_IMPL_CUDA_SAFE_CALL(
+      cudaGraphExecGetFlags(graph->native_graph_exec(), &flags));
+
+  ASSERT_EQ(flags, 0u);
+#endif
+}
+
+}  // namespace
--- a/core/unit_test/hip/TestHIP_InterOp_Graph.cpp
+++ b/core/unit_test/hip/TestHIP_InterOp_Graph.cpp
@@ -1,0 +1,119 @@
+//@HEADER
+// ************************************************************************
+//
+//                        Kokkos v. 4.0
+//       Copyright (2022) National Technology & Engineering
+//               Solutions of Sandia, LLC (NTESS).
+//
+// Under the terms of Contract DE-NA0003525 with NTESS,
+// the U.S. Government retains certain rights in this software.
+//
+// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
+#include <filesystem>
+#include <fstream>
+#include <regex>
+
+#include <TestHIP_Category.hpp>
+#include <Kokkos_Core.hpp>
+#include <Kokkos_Graph.hpp>
+
+#include <gtest/gtest.h>
+
+namespace {
+
+template <typename ViewType>
+struct Increment {
+  ViewType data;
+
+  KOKKOS_FUNCTION
+  void operator()(const int) const { ++data(); }
+};
+
+// This test checks the promises of Kokkos::Graph against its
+// underlying HIP native objects.
+TEST(TEST_CATEGORY, graph_promises_on_native_objects) {
+#if !defined(KOKKOS_IMPL_HIP_NATIVE_GRAPH)
+  GTEST_SKIP() << "This test will not work without native graph support";
+#else
+  auto graph = Kokkos::Experimental::create_graph<Kokkos::HIP>();
+
+  auto root = Kokkos::Impl::GraphAccess::create_root_ref(graph);
+
+  // Before instantiation, the HIP graph is valid, but the HIP executable
+  // graph is still null.
+  hipGraph_t hip_graph = graph.native_graph();
+
+  ASSERT_NE(hip_graph, nullptr);
+  ASSERT_EQ(graph.native_graph_exec(), nullptr);
+
+  // After instantiation, both native objects are valid.
+  graph.instantiate();
+
+  hipGraphExec_t hip_graph_exec = graph.native_graph_exec();
+
+  ASSERT_EQ(graph.native_graph(), hip_graph);
+  ASSERT_NE(hip_graph_exec, nullptr);
+
+  // Submission should not affect the underlying objects.
+  graph.submit();
+
+  ASSERT_EQ(graph.native_graph(), hip_graph);
+  ASSERT_EQ(graph.native_graph_exec(), hip_graph_exec);
+#endif
+}
+
+// Use native HIP graph to generate a DOT representation.
+TEST(TEST_CATEGORY, graph_instantiate_and_debug_dot_print) {
+#if !defined(KOKKOS_IMPL_HIP_NATIVE_GRAPH)
+  GTEST_SKIP() << "This test will not work without native graph support";
+#else
+  using view_t = Kokkos::View<int, Kokkos::HIP>;
+
+  const Kokkos::HIP exec{};
+
+  view_t data(Kokkos::view_alloc(exec, "witness"));
+
+  auto graph = Kokkos::Experimental::create_graph(exec);
+
+  auto root = Kokkos::Impl::GraphAccess::create_root_ref(graph);
+
+  root.then_parallel_for(1, Increment<view_t>{data});
+
+  graph.instantiate();
+
+  size_t num_nodes;
+
+  KOKKOS_IMPL_HIP_SAFE_CALL(
+      hipGraphGetNodes(graph.native_graph(), nullptr, &num_nodes));
+
+  ASSERT_EQ(num_nodes, 2u);
+
+  const auto dot = std::filesystem::temp_directory_path() / "hip_graph.dot";
+
+  KOKKOS_IMPL_HIP_SAFE_CALL(hipGraphDebugDotPrint(
+      graph.native_graph(), dot.c_str(), hipGraphDebugDotFlagsVerbose));
+
+  ASSERT_TRUE(std::filesystem::exists(dot));
+  ASSERT_GT(std::filesystem::file_size(dot), 0u);
+
+  // We could write a check against the full kernel's function signature, but
+  // it would make the test rely too much on internal implementation details.
+  // Therefore, we just look for the functor and policy. Note that the
+  // signature is mangled in the 'dot' output.
+  const std::string expected("[A-Za-z0-9_]+Increment[A-Za-z0-9_]+RangePolicy");
+
+  std::stringstream buffer;
+  buffer << std::ifstream(dot).rdbuf();
+
+  ASSERT_TRUE(std::regex_search(buffer.str(), std::regex(expected)))
+      << "Could not find expected signature regex " << std::quoted(expected)
+      << " in " << dot;
+#endif
+}
+
+}  // namespace
--- a/core/unit_test/sycl/TestSYCL_InterOp_Graph.cpp
+++ b/core/unit_test/sycl/TestSYCL_InterOp_Graph.cpp
@@ -1,0 +1,104 @@
+//@HEADER
+// ************************************************************************
+//
+//                        Kokkos v. 4.0
+//       Copyright (2022) National Technology & Engineering
+//               Solutions of Sandia, LLC (NTESS).
+//
+// Under the terms of Contract DE-NA0003525 with NTESS,
+// the U.S. Government retains certain rights in this software.
+//
+// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
+#include <filesystem>
+#include <fstream>
+#include <regex>
+
+#include <TestSYCL_Category.hpp>
+#include <Kokkos_Core.hpp>
+#include <Kokkos_Graph.hpp>
+
+#include <gtest/gtest.h>
+
+namespace {
+
+template <typename ViewType>
+struct Increment {
+  ViewType data;
+
+  KOKKOS_FUNCTION
+  void operator()(const int) const { ++data(); }
+};
+
+TEST(TEST_CATEGORY, graph_get_native_return_types_are_references) {
+  using graph_t = Kokkos::Experimental::Graph<Kokkos::SYCL>;
+  static_assert(
+      std::is_reference_v<decltype(std::declval<graph_t>().native_graph())>);
+  static_assert(std::is_reference_v<
+                decltype(std::declval<graph_t>().native_graph_exec())>);
+}
+
+// This test checks the promises of Kokkos::Graph against its
+// underlying SYCL native objects.
+TEST(TEST_CATEGORY, graph_promises_on_native_objects) {
+  auto graph = Kokkos::Experimental::create_graph<Kokkos::SYCL>();
+
+  auto root = Kokkos::Impl::GraphAccess::create_root_ref(graph);
+
+  // Before instantiation, the SYCL graph is valid, but the SYCL executable
+  // graph is still null. Since the SYCL command graph is a regular object,
+  // no check is needed.
+  // However, the executable SYCL command graph is stored as an optional,
+  // so let's check it is empty for now.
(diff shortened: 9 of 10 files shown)


core/src/Cuda/Kokkos_Cuda_Graph_Impl.hpp
```diff
@@ -200,6 +200,9 @@ struct GraphImpl<Kokkos::Cuda> {
         m_execution_space, _graph_node_kernel_ctor_tag{},
         aggregate_kernel_impl_t{});
   }
+
+  const cudaGraph_t     getNativeGraph()           const { return m_graph; }
```
romintomasetti, line 204: @dalg24 I don't know what's best. Make `m_graph` public or use a getter?

I don't have a strong opinion on any.

core/src/HIP/Kokkos_HIP_Graph_Impl.hpp
```diff
@@ -69,6 +69,9 @@ class GraphImpl<Kokkos::HIP> {
   template <class... PredecessorRefs>
   auto create_aggregate_ptr(PredecessorRefs&&...);
 
+  const hipGraph_t     getNativeGraph()           const { return m_graph; }
+  const hipGraphExec_t getNativeExecutableGraph() const { return m_graph_exec; }
```
masterleinad, line 73: Matching the local style
```suggestion
  const hipGraph_t     get_native_graph()           const { return m_graph; }
  const hipGraphExec_t get_native_executable_graph() const { return m_graph_exec; }
```

core/src/Cuda/Kokkos_Cuda_Graph_Impl.hpp
dalg24: A getter for sure.
Why is your return type const-qualified?

core/src/HIP/Kokkos_HIP_Graph_Impl.hpp
romintomasetti: Indeed ! Done :heavy_check_mark:

core/src/Cuda/Kokkos_Cuda_Graph_Impl.hpp
```diff
@@ -200,6 +200,9 @@ struct GraphImpl<Kokkos::Cuda> {
         m_execution_space, _graph_node_kernel_ctor_tag{},
         aggregate_kernel_impl_t{});
   }
+
+  const cudaGraph_t get_native_graph() const { return m_graph; }
+  const cudaGraphExec_t get_native_executable_graph() const { return m_graph_exec; }
```
dalg24, line 205: I am tempted to suggest to call these `get_cuda_graph[_exec]` because I am not convinced there is much generic code one could write that get the native graph nor that the concept of "executable" graph will always make sense.

romintomasetti: - Agreed for `_exec` suffix.
- For the `get_cuda_graph`, it would mean `get_cuda_graph`/`get_hip_graph`, and so on (*i.e.* not portable across backends). I guess the question is if you want something that looks like `space.cuda_stream()`/`space.hip_stream`, or if you want something like `get_graph_impl` ("portable" at least across `Cuda` and `HIP` backends)?

core/src/Cuda/Kokkos_Cuda_Graph_Impl.hpp
romintomasetti: Sorry, `const` qualification makes no sense here. Removed.

core/src/Cuda/Kokkos_Cuda_Graph_Impl.hpp
dalg24: That's my point, it is more about interoperability than portability here.

romintomasetti: Fair enough !

romintomasetti:
@dalg24 @masterleinad If you don't have any further comment, this one is good to go!

masterleinad (commented):
Can you add a test that demonstrates a use case?

romintomasetti:
> Can you add a test that demonstrates a use case?

Would something along these lines be OK for you ?
```c++
//! @test This test ensures that the underlying @c CUDA objects can be retrieved from a @c Kokkos::Graph.
TEST_F(TEST_CATEGORY, cuda_graph) {
  //! Build a non-empty graph. The content does not matter as it won't be executed.
  auto graph =
      Kokkos::Experimental::create_graph<TEST_EXECSPACE>([&](auto root) {
        root.then_parallel_for(1, CountTestFunctor<TEST_EXECSPACE>{});
      });

  //! Retrieve the @c Kokkos::Impl::Graph.
  auto root_node_ref  = Kokkos::Impl::GraphAccess::create_root_ref(graph);
  auto graph_ptr_impl = Kokkos::Impl::GraphAccess::get_graph_weak_ptr(root_node_ref);

  //! Instantiate the graph manually.
  cudaGraphInstantiate(
    graph_ptr_impl->get_cuda_graph_exec(),
    graph_ptr_impl->get_cuda_graph()
  );
}
```
(in `TestCuda_Graph.cpp`, and similarly in `testHIP_Graph.cpp`)

masterleinad:
Yeah, something like that, maybe also call `cudaGraphDebugDotPrint`?

romintomasetti:
- I wrote a test for `Cuda`, and I basically copy-pasted it for `HIP`. Is the copy-paste the expected testing strategy ?
- The test creates a graph through `Kokkos` API, then retrieves the underlying backend objects to instantiate them manually, and then uses the `Kokkos` API again to submit the graph. It seems this test is more of an "interoperability" test. `Kokkos` has some `*InterOp*` test files. Should I move my tests to some `TestCuda_InterOp_Graph.cpp` ?

Rombur:
There is an error with CUDA 11.0
```
/var/jenkins/workspace/Kokkos_PR-6904/core/unit_test/cuda/TestCuda_Graph.cpp(79): error: identifier "cudaGraphDebugDotFlagsVerbose" is undefined
/var/jenkins/workspace/Kokkos_PR-6904/core/unit_test/cuda/TestCuda_Graph.cpp(78): error: identifier "cudaGraphDebugDotPrint" is undefined
```
There is also a problem with ROCM 5.2:
```
/var/jenkins/workspace/Kokkos_PR-6904/core/unit_test/hip/TestHIP_Graph.cpp:49:29: error: no member named 'get_hip_graph_exec' in 'Kokkos::Impl::GraphImpl<Kokkos::HIP>'
  ASSERT_EQ(graph_ptr_impl->get_hip_graph_exec(), nullptr);
            ~~~~~~~~~~~~~~~~^
```
The graph library is bugged in ROCm 5.2, so we are not using it. `Kokkos_HIP_Graph_Impl.hpp` is not included when you compile with ROCm 5.2, so you need to guard your test.

core/src/Cuda/Kokkos_Cuda_Graph_Impl.hpp
```diff
@@ -200,6 +200,9 @@ struct GraphImpl<Kokkos::Cuda> {
         m_execution_space, _graph_node_kernel_ctor_tag{},
         aggregate_kernel_impl_t{});
   }
+
+  cudaGraph_t& get_cuda_graph() { return m_graph; }
+  cudaGraphExec_t& get_cuda_graph_exec() { return m_graph_exec; }
```
dalg24, line 205: Why did you change and now return a reference?  AFAIK `cudaGraph[Exec]_t` are handle types, some typedef for a pointer to an implementation-defined struct and they are intended to be passed by value.

dalg24: That said I was about to ask about the const-qualifier in the previous version.

core/unit_test/cuda/TestCuda_Graph.cpp
```diff
@@ -16,3 +16,67 @@
...
+  auto graph =
+      Kokkos::Experimental::create_graph<TEST_EXECSPACE>([&](auto root) {
+        root.then_parallel_for(
+            1, SetViewToValueFunctor<TEST_EXECSPACE, int>{data, 1});
+      });
+
+  //! Retrieve the @c Kokkos::Impl::Graph.
+  auto root_node_ref = Kokkos::Impl::GraphAccess::create_root_ref(graph);
+  auto graph_ptr_impl =
+      Kokkos::Impl::GraphAccess::get_graph_weak_ptr(root_node_ref).lock();
```
dalg24, line 45: Just noting that we are reaching into an `Impl::` facility here so users won't technically be allowed to use it.

dalg24 (commented):
Does not necessarily need to be addressed here, but so far we have only looked into interoperability in one direction Kokkos ->  raw CUDA graph but we might want to think about ways to build a root node from and existing CUDA graph or whatnot.

core/unit_test/cuda/TestCuda_Graph.cpp
romintomasetti: Indeed. I guess this convoluted way of accessing the backend specialization stems from the original implementation strategy which I think aimed at providing some sort of *type erasure*.

core/src/Cuda/Kokkos_Cuda_Graph_Impl.hpp
romintomasetti: ### `get_cuda_graph`

I think that if we want to allow interoperability in both direction, it has to be
```c++
cudaGraph_t& get_cuda_graph() { return m_graph; }
```
(*i.e.* by reference).

Because if I have an existing `cudaGraph_t`, and want to "set it" to my `Kokkos::Graph`, there is no other way I guess. Or maybe you'd want a `set_cuda`graph` ? Up to you :smile:

### `get_cuda_graph_exec`

For this one, one of the primary reasons to handle `cudaGraphExec_t` "manually" would be when one wants to *instantiate the graph with parameters* (something the `Kokkos` API does not allow us to do for now).

The signature of one of the overloads of `cudaGraphInstantiate` from the `Cuda` documentation:
```c++
__host__ ​cudaError_t cudaGraphInstantiateWithParams ( cudaGraphExec_t* pGraphExec, cudaGraph_t graph, cudaGraphInstantiateParams* instantiateParams );
```

dalg24: OK so that was your way of addressing https://github.com/kokkos/kokkos/pull/6904#pullrequestreview-1972501757

I am not a big fan of the getter returning a reference.  I would have preferred if we could find a way to set the root node at construction and not allow anyone to change the graph from underneath Kokkos otherwise.  This makes it possible to break invariances such as mismatch between the graph and the "executable".

Regarding the setter for the graph executable.  Do you actually need it?  Are you not able to set properties from the handle when it is returned by value?

romintomasetti: #### Graph handle

For the `cudaGraph_t` object, I guess it's enough for now to be able to retrieve it *by value*. I suggest we use this PR to add just the regualr getter  as follows:
```c++
cudaGraph_t get_cuda_graph() const { return m_graph; }
```
This would already enable interoperability "from `Kokkos` to outside". And we could discuss in another issue how to best enable interoperability "from outside to `Kokkos`", *i.e.* how to allow construction of a `Kokkos::Graph` from an existing `cudaGraph_t` handle.

#### Executable graph handle

The function used by `Kokkos` with the error node and logs has been "removed" in `CUDA 12` (https://developer.nvidia.com/blog/cuda-toolkit-12-0-released-for-general-availability/, still in https://docs.nvidia.com/cuda/archive/11.1.0/cuda-runtime-api/group__CUDART__GRAPH.html#group__CUDART__GRAPH but not anymore in https://docs.nvidia.com/cuda/archive/12.4.0/cuda-runtime-api/group__CUDART__GRAPH.html#group__CUDART__GRAPH).

I looked in my include files (I'm on `Cuda 12.3.2`):
```c++
static __inline__ __host__ cudaError_t cudaGraphInstantiate( cudaGraphExec_t *pGraphExec,  cudaGraph_t graph,  cudaGraphNode_t *pErrorNode,  char *pLogBuffer,  size_t bufferSize){
  (void)pErrorNode;
  (void)pLogBuffer;
  (void)bufferSize;
  return ::cudaGraphInstantiate(pGraphExec, graph, 0);
}
```
In the new `Cuda` versions, these are the API functions that can be used to instantiate the graph:
```c++
__host__ ​cudaError_t cudaGraphInstantiate ( cudaGraphExec_t* pGraphExec, cudaGraph_t graph, unsigned long long flags = 0 )

__host__ ​cudaError_t cudaGraphInstantiateWithFlags ( cudaGraphExec_t* pGraphExec, cudaGraph_t graph, unsigned long long flags = 0 )

__host__ ​cudaError_t cudaGraphInstantiateWithParams ( cudaGraphExec_t* pGraphExec, cudaGraph_t graph, cudaGraphInstantiateParams* instantiateParams ) 
```

So we could now pass the parameters through `Kokkos::Impl::instantiate` more easily. I pushed the changes just for you to see. Note that we might enter into some `CUDA` API versioning issues (the `CUDA` graph API changed in `CUDA 12`).

core/src/Cuda/Kokkos_Cuda_Graph_Impl.hpp
```diff
@@ -199,6 +199,9 @@ struct GraphImpl<Kokkos::Cuda> {
         m_execution_space, _graph_node_kernel_ctor_tag{},
         aggregate_kernel_impl_t{});
   }
+
+  cudaGraph_t get_cuda_graph() const { return m_graph; }
+  cudaGraphExec_t get_cuda_graph_exec() const { return m_graph_exec; }
```
dalg24, line 204: No preconditions?

dalg24 (commented):
What about the SYCL backend?

romintomasetti:
I guess `SYCL` can be tackled by someone else in another PR if there is any interest, right ?

core/src/Cuda/Kokkos_Cuda_Graph_Impl.hpp
romintomasetti: I am not sure to see a good reason for precondition here.

Why would it be disallowed for an advanced user to query the `m_graph_exec` at anytime ?

core/unit_test/CMakeLists.txt
```diff
@@ -825,6 +825,13 @@ if(Kokkos_ENABLE_CUDA)
       UnitTestMainInit.cpp
       cuda/TestCuda_InterOp_StreamsMultiGPU.cpp
   )
+
+  KOKKOS_ADD_EXECUTABLE_AND_TEST(
+    CoreUnitTest_CudaInterOpGraph
+    SOURCES
+      UnitTestMainInit.cpp
+      cuda/TestCuda_InterOp_Graph.cpp
+  )
```
romintomasetti, line 834: `SYCL` seems to have [`print_graph`](https://github.com/intel/llvm/blob/fbb1fb0010ec128c93148461c951dccf4c01b5b7/sycl/include/sycl/ext/oneapi/experimental/graph.hpp#L317), let's not forget `SYCL`.

romintomasetti: Done.

masterleinad:
```
/__w/kokkos/kokkos/core/unit_test/TestGraph.hpp:73:30: error: 'SetViewToValueFunctor' does not name a type
   73 |   using set_functor        = SetViewToValueFunctor<TEST_EXECSPACE, int>;
      |                              ^~~~~~~~~~~~~~~~~~~~~
```

cmake/kokkos_enable_options.cmake
```diff
@@ -58,6 +58,7 @@ KOKKOS_ENABLE_OPTION(IMPL_CUDA_UNIFIED_MEMORY OFF "Whether to leverage unified m
 KOKKOS_ENABLE_OPTION(DEPRECATED_CODE_4    ON "Whether code deprecated in major release 4 is available" )
 KOKKOS_ENABLE_OPTION(DEPRECATION_WARNINGS ON "Whether to emit deprecation warnings" )
 KOKKOS_ENABLE_OPTION(HIP_RELOCATABLE_DEVICE_CODE  OFF "Whether to enable relocatable device code (RDC) for HIP")
+KOKKOS_ENABLE_OPTION(IMPL_SYCL_GRAPH OFF "Whether to enable graph for SYCL (sycl_ext_oneapi_graph)")
```
masterleinad, line 61: I don't think this is a good idea. We want support to be enabled if the implementation supports SYCL Command Graphs.

romintomasetti: I need a way to skip the `TestSYCL_InterOp_Graph.cpp` if there is no support for it.

But OK, I could just keep the `KOKKOS_IMPL_HAVE_SYCL_EXT_ONEAPI_GRAPH` and drop `KOKKOS_ENABLE_IMPL_SYCL_GRAPH` if you don't the new option. I don't feel strongly about it anyway.

romintomasetti: Still I think as a user I might want to ensure that `Kokkos` will use the `SYCL` specialization, and so maybe the user would like to provide `KOKKOS_ENABLE_IMPL_SYCL_GRAPH=ON` to ensure that `Kokkos` errors out if this cannot be done?

core/src/HIP/Kokkos_HIP_Graph_Impl.hpp
```diff
@@ -79,6 +79,9 @@ class GraphImpl<Kokkos::HIP> {
     KOKKOS_ENSURES(m_graph_exec);
   }
 
+  hipGraph_t get_hip_graph() { return m_graph; }
+  hipGraphExec_t get_hip_graph_exec() { return m_graph_exec; }
```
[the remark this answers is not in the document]
romintomasetti: Let's make them all non-`const`.

cmake/kokkos_enable_options.cmake
romintomasetti: `SYCL_EXT_ONEAPI_GRAPH` is defined within `SYCL` headers, so I need `KOKKOS_IMPL_HAVE_SYCL_EXT_ONEAPI_GRAPH` to discard the test. I think I prefer discarding the `TestSYCL_InterOp_Graph.cpp`, rather than haing it but empty.

masterleinad (approved):
Fine with me.

core/unit_test/hip/TestHIP_InterOp_Graph.cpp
```diff
@@ -0,0 +1,97 @@
...
+
+  ASSERT_EQ(num_nodes, 2u);
+
+  const auto dot = std::filesystem::temp_directory_path() / "hip_graph.dot";
+
+  KOKKOS_IMPL_HIP_SAFE_CALL(
+      hipGraphDebugDotPrint(graph_ptr_impl->get_hip_graph(), dot.c_str(),
+                            hipGraphDebugDotFlagsVerbose));
+
+  ASSERT_TRUE(std::filesystem::exists(dot));
```
Rombur, line 94: How good is that test? Is a file, possibly empty, created if the graph is invalid?

romintomasetti: I could check if the file is not empty, but checking the content is harder (and I guesswe don't want to hardcode the expected result, as it depends on pointer addresses and stuff).

Rombur: I am asking because if `hipGraphDebugDotPrint` creates a file no matter what, the test is useless.

romintomasetti: OK, so I've just written a check for what's the content of the exported `dot`. But the test relies too much on internals (like the threshold for constant vs global memory launch).

So I will revert this.

I think that the test in its current form is indeed not very meaningful. However, noting that the graph has been instantiated just before, this means the backend had a chance to shout if the graph was invalid.

Making a "very meaningful" use of the executable graph would make the test much more complex (for instance I could [change the kernel parameters](https://rocm.docs.amd.com/projects/HIP/en/latest/doxygen/html/group___graph.html#ga5b1918dae65224863b7370e6d4ad3f2a)), so I think just adding a check that the file is not empty would be enough. Do you agree ?

romintomasetti:
I have explicitly kept the commit with the "complicated test that relies too much on internals", and added a revert commit.

@masterleinad Do you think that it's OK / brings value or should I remove it completely ? Thanks !

masterleinad:
> @masterleinad Do you think that it's OK / brings value or should I remove it completely ? Thanks !

I'm expecting that we would squash all commits when merging anyway.

romintomasetti:
I just dropped them. It's easy to re-write it if need be.

core/unit_test/sycl/TestSYCL_InterOp_Graph.cpp
```diff
@@ -0,0 +1,82 @@
...
+  graph.instantiate();
+
+  ASSERT_EQ(graph_ptr_impl->get_sycl_graph().get_nodes().size(), 2u);
+
+  const auto dot = std::filesystem::temp_directory_path() / "sycl_graph.dot";
+
+  graph_ptr_impl->get_sycl_graph().print_graph(dot, true);
+
+  ASSERT_TRUE(std::filesystem::exists(dot));
+  ASSERT_GT(std::filesystem::file_size(dot), 0u);
```
Rombur, line 79: Why didn't you check that the node is in the `dot` file for SYCL?

romintomasetti: Done. I was able to compile it with my setup but could not execute the test.

Rombur: Did you forget to push the change?

romintomasetti: Oups. I pushed to my own fork, but this PR is from a branch of our organization :dizzy_face:

Rombur (dismissed):

core/unit_test/sycl/TestSYCL_InterOp_Graph.cpp
masterleinad: I check that this works on the testbeds.

core/unit_test/TestGraph_helpers.hpp
```diff
@@ -0,0 +1,35 @@
...
+//
+// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
+#include <Kokkos_Core.hpp>
+
+namespace {
```
dalg24, line 19: We use anonymous namespaces we want to avoid symbol collision.
Not sure it makes much sense here

dalg24 (commented):
If the intent is to allow users to access these native graph handle, then the PR does not achieve its goal because it currently requires naming implementation details which by definition are out of reach of users.

romintomasetti:
Following @dalg24 advice, I've refactored a few things:
- Allow native objects to be retrieved with using any `Kokkos::Impl`. This is done by adding *e.g.* `cuda_graph` method to `Kokkos::Graph` if `Cuda` is enabled. This sounds much like `Kokkos::Cuda::cuda_stream` method. Still to be done however is to treat the case I want a `Kokkos::Graph<Serial>` when `Cuda` is enabled. In that case it sounds really awkward to have a `cuda_graph` getter.
- Get rid of the test helper.

core/src/Kokkos_Graph.hpp
```diff
@@ -97,6 +97,40 @@ struct [[nodiscard]] Graph {
...
+#if defined(KOKKOS_ENABLE_CUDA)
+  cudaGraph_t cuda_graph() {
+    KOKKOS_EXPECTS(bool(m_impl_ptr))
+    return m_impl_ptr->cuda_graph();
+  }
+  cudaGraphExec_t cuda_graph_exec() {
+    KOKKOS_EXPECTS(bool(m_impl_ptr))
+    return m_impl_ptr->cuda_graph_exec();
+  }
+#endif
```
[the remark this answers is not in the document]
romintomasetti: I followed your suggestion. Thanks !

core/unit_test/TestGraph_helpers.hpp
romintomasetti: I got rid of this helper file, as per your suggestions. Thanks !

masterleinad:
You'll have to run `clang-format` again.

core/src/Cuda/Kokkos_Cuda_Graph_Impl.hpp
romintomasetti: I guess we can close this one. There is no precondition in the `GraphImpl`, but there are in the `Graph`.

cmake/kokkos_enable_options.cmake
romintomasetti: Closing.

romintomasetti:
The Jenkins jobs revealed a few issues:
- The interoperability test for `HIP` must be deactivated for `HIP 5.2` because for that version `Kokkos::Graph` uses the defaulted implementation.
- Some parts of the `Cuda` interoperability test need to be disabled for older Cuda.

cmake/kokkos_arch.cmake
```diff
@@ -613,6 +613,17 @@ IF (KOKKOS_ENABLE_HIP)
...
+  set(KOKKOS_IMPL_HAVE_HIP_WORKING_GRAPH_CODE "
+    #include <hip/hip_runtime_api.h>
+
+    #if HIP_VERSION_MAJOR == 5 && HIP_VERSION_MINOR == 2
+      #error \"HIP does not have a working implementation of 'hipGraph'.\"
+    #endif
+    int main() {}
+  ")
+  check_cxx_source_runs(KOKKOS_IMPL_HAVE_HIP_WORKING_GRAPH_CODE KOKKOS_IMPL_HAVE_HIP_WORKING_GRAPH_IMPLEMENTATION)
+  unset(KOKKOS_IMPL_HAVE_HIP_WORKING_GRAPH_CODE)
```
masterleinad, line 626: Why are you not using `KOKKOS_CXX_COMPILER_VERSION`?

romintomasetti: The only doubt I have: is it true that `hipcc` is the only compiler supported by `Kokkos` to compile `HIP` ?

romintomasetti: @Rombur

romintomasetti:
To me, the Windows job failure is a compiler error. Could anyone have a look ?
```bash
2024-09-17T22:44:17.6145119Z D:\a\kokkos\kokkos\core\unit_test\cuda\TestCuda_InterOp_Graph.cpp(92): error : argument of type "const std::filesystem::path::value_type *" is incompatible with parameter of type "const char *" [D:\a\kokkos\kokkos\build\core\unit_test\Kokkos_CoreUnitTest_CudaInterOpGraph.vcxproj]
2024-09-17T22:44:17.6149755Z       Kokkos::Impl::cuda_internal_safe_call(cudaGraphDebugDotPrint( graph.native_graph(), dot.c_str(), cudaGraphDebugDotFlagsVerbose), "cudaGraphDebugDotPrint( graph.native_graph(), dot.c_str(), cudaGraphDebugDotFlagsVerbose)", "D:\\a\\kokkos\\kokkos\\core\\unit_test\\cuda\\TestCuda_InterOp_Graph.cpp", 93);
2024-09-17T22:44:17.6152515Z                                                                                           ^
2024-09-17T22:44:17.6153255Z   
2024-09-17T22:44:17.7623016Z   1 error detected in the compilation of "D:/a/kokkos/kokkos/core/unit_test/cuda/TestCuda_InterOp_Graph.cpp".
```

masterleinad:
> To me, the Windows job failure is a compiler error. Could anyone have a look ?

`const std::filesystem::path::value_type` is `wchar_t` on Windows according to https://en.cppreference.com/w/cpp/filesystem/path. So the type of `dot` should explicitly be `const char*`.

cmake/kokkos_arch.cmake
Rombur: >  is it true that hipcc is the only compiler supported by Kokkos to compile HIP ?

No, we also support `amdclang` and Cray Clang

romintomasetti: OK, so the proposed `if` condition based on the `HIPCC` compiler version would not be generic, right ?

I guess we've got 2 choices:
- be pragmatic and say that the `if` will discard the test for `hipcc < 5.3` **and** if the compiler is not `hipcc`
- come up with another way of detecting the version of `HIP`

Rombur: You could do something like this in the test itself
```C++
#if ((HIP_VERSION_MAJOR == 5) && (HIP_VERSION_MINOR == 2))
GTEST_SKIP();
#else 
Your test
#endif
```

romintomasetti: Yeah I know, but I really don't want to add a unit test executable whose all tests are skipped. It makes people think it is tested, until they carefully look at the test output (which no one does when `ctest` reports that all is fine).

romintomasetti:
Failures seem unrelated. Could you have a look @Rombur @masterleinad @dalg24  ? Thanks !

romintomasetti:
The only Jenkins job failing did for some unrelated reason.

Could you please guys review this one again ?

core/unit_test/TestGraph.hpp
```diff
@@ -398,6 +398,16 @@ TEST_F(TEST_CATEGORY_FIXTURE(graph), empty_graph) {
   ex.fence();
 }
 
+// Ensure that an empty graph on the default host execution space
+// can be submitted.
+TEST_F(TEST_CATEGORY_FIXTURE(graph), empty_graph_default_host_exec) {
```
dalg24, line 403: Skip if default host exec is the same as default exec?

Also why is this part of this PR?

core/src/Kokkos_Graph.hpp
```diff
@@ -168,6 +172,45 @@ create_graph(Closure&& arg_closure) {
...
+#elif defined(KOKKOS_ENABLE_HIP) && defined(KOKKOS_HIP_NATIVE_GRAPH)
+  if constexpr (std::is_same_v<ExecutionSpace, Kokkos::HIP>) {
+    return m_impl_ptr->hip_graph();
+  }
+#elif defined(KOKKOS_ENABLE_SYCL) && defined(SYCL_EXT_ONEAPI_GRAPH)
+  if constexpr (std::is_same_v<ExecutionSpace, Kokkos::SYCL>) {
+    return m_impl_ptr->sycl_graph();
+  }
+#endif
+  Kokkos::abort("Native graph object not available for this execution space");
```
dalg24, line 191: They'll get a compile error if you don't return anything.

core/unit_test/cuda/TestCuda_InterOp_Graph.cpp
```diff
@@ -0,0 +1,125 @@
...
+  graph.instantiate();
+
+  size_t num_nodes = 0;
+
+  KOKKOS_IMPL_CUDA_SAFE_CALL(
+      cudaGraphGetNodes(graph.native_graph(), nullptr, &num_nodes));
+
+  ASSERT_EQ(num_nodes, 2u);
+
+#if CUDA_VERSION >= 11600
```
dalg24, line 89: Skip the test in that case so we don;t get the false impression it did what the label says

core/unit_test/cuda/TestCuda_InterOp_Graph.cpp
```diff
@@ -0,0 +1,125 @@
...
+#endif
+
+#if CUDA_VERSION >= 12000
+  unsigned long long flags =
+      Kokkos::Experimental::finite_max_v<unsigned long long>;
+  KOKKOS_IMPL_CUDA_SAFE_CALL(
+      cudaGraphExecGetFlags(graph.native_graph_exec(), &flags));
+
+  ASSERT_EQ(flags, 0u);
+#endif
```
dalg24, line 122: What is the added value?
Does it belong in that unit test and is it worth having?

core/src/setup/Kokkos_Setup_HIP.hpp
```diff
@@ -43,6 +43,11 @@
 #endif
 // clang-format on
 
+// The implementation of hipGraph in ROCm 5.2 is bugged, so we cannot use it.
+#if !((HIP_VERSION_MAJOR == 5) && (HIP_VERSION_MINOR == 2))
+#define KOKKOS_HIP_NATIVE_GRAPH 1
```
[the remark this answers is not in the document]
romintomasetti: @Rombur  ?

core/unit_test/TestGraph.hpp
romintomasetti: I wrote this test when I realized that for backends such as `HIP` or `SYCL`, `Kokkos::Graph<backend>` could potentially go on the defaulted implementation. My main concern was that at some point we could potentially make things wrong enough such that it would not be possible anymore to instantiate the graph for the default host execution space type.

core/unit_test/cuda/TestCuda_InterOp_Graph.cpp
romintomasetti: It might be better to do it in a new unit test, indeed.

Worth having, I'd say not really for now because we always instantiate with the defaults. So this tests that we instantiate with defaults (we don't do anything smart / experimental).

core/src/setup/Kokkos_Setup_HIP.hpp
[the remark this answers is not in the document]
Rombur: > This naming (macro KOKKOS_BACKEND_BLA) is a bit unorthodox.
We typically say "enabled" followed by underscore backend and we don't actually give it a value.

Usually we use "enabled" when the user has control over the macro through CMake. Here there is nothing that the user can do, we just check the ROCm version. That's why we went with `KOKKOS_HIP_NATIVE_GRAPH`. I am fine using "enabled" for consistency.

> Do we intend it to be public?

Yes, it needs to be public because it changes the behavior of public functions.

> If so, should we including it in the information outputted by HIP::print_configuration()?

I would say no. It's useless if someone reports a bug because we already know that information thanks to the ROCm version. The macro will always be defined except for ROCm  5.2 which is not available on Frontier anymore.

cmake/kokkos_arch.cmake
romintomasetti: I think we can close this one too.

core/src/Kokkos_Graph.hpp
```diff
@@ -168,6 +172,45 @@ create_graph(Closure&& arg_closure) {
...
 //==============================================================================
 
+template <class ExecutionSpace>
+decltype(auto) Graph<ExecutionSpace>::native_graph() {
+  KOKKOS_EXPECTS(bool(m_impl_ptr));
+#if defined(KOKKOS_ENABLE_CUDA)
+  if constexpr (std::is_same_v<ExecutionSpace, Kokkos::Cuda>) {
+    return m_impl_ptr->cuda_graph();
+  }
+#elif defined(KOKKOS_ENABLE_HIP) && defined(KOKKOS_IMPL_HIP_GRAPH_ENABLED)
```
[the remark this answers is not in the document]
romintomasetti: This was tackled in #7378. Closing.

core/src/setup/Kokkos_Setup_HIP.hpp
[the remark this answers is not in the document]
romintomasetti: This was tackled in #7378. Closing.

core/src/Kokkos_Graph.hpp
romintomasetti: OK, let's remove the abort then :heavy_check_mark:

core/unit_test/cuda/TestCuda_InterOp_Graph.cpp
romintomasetti: Done :heavy_check_mark:

core/unit_test/cuda/TestCuda_InterOp_Graph.cpp
romintomasetti: I've put that assertion in a separated unit test "instantiation flags inspection".

Rombur:
You need to fix cmake-format

core/src/Kokkos_Graph.hpp
```diff
@@ -179,7 +219,6 @@ create_graph(Closure&& arg_closure) {
 #include <impl/Kokkos_Default_Graph_Impl.hpp>
 #include <Cuda/Kokkos_Cuda_Graph_Impl.hpp>
 #if defined(KOKKOS_ENABLE_HIP)
-// The implementation of hipGraph in ROCm 5.2 is bugged, so we cannot use it.
```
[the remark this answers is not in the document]
romintomasetti: :heavy_check_mark:

core/unit_test/CMakeLists.txt
```diff
@@ -730,6 +730,10 @@ if(Kokkos_ENABLE_CUDA)
   kokkos_add_executable_and_test(
     CoreUnitTest_CudaInterOpStreamsMultiGPU SOURCES UnitTestMainInit.cpp cuda/TestCuda_InterOp_StreamsMultiGPU.cpp
   )
+
+  KOKKOS_ADD_EXECUTABLE_AND_TEST(
```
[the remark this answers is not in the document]
romintomasetti: :heavy_check_mark:

core/unit_test/CMakeLists.txt
```diff
@@ -787,6 +795,12 @@ if(Kokkos_ENABLE_SYCL)
   kokkos_add_executable_and_test(
     CoreUnitTest_SYCLInterOpStreamsMultiGPU SOURCES UnitTestMainInit.cpp sycl/TestSYCL_InterOp_StreamsMultiGPU.cpp
   )
+
+  if(KOKKOS_IMPL_HAVE_SYCL_EXT_ONEAPI_GRAPH)
+    KOKKOS_ADD_EXECUTABLE_AND_TEST(
```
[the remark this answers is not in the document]
romintomasetti: :heavy_check_mark:

core/unit_test/cuda/TestCuda_InterOp_Graph.cpp
```diff
@@ -0,0 +1,144 @@
...
+  using view_t =
+      Kokkos::View<int, execution_space, Kokkos::MemoryTraits<Kokkos::Atomic>>;
+  using graph_t = Kokkos::Experimental::Graph<execution_space>;
+
+  void SetUp() override {
+    data = view_t(Kokkos::view_alloc(exec, "witness"));
+
+    graph = Kokkos::Experimental::create_graph(exec);
+
+    auto root = Kokkos::Impl::GraphAccess::create_root_ref(*graph);
```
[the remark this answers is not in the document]
romintomasetti: No particular reason. I'm fine either way. Would you prefer it written like that ?
```c++
graph = Kokkos::Experimental::create_graph(exec, [&](auto& root){
    root.then_parallel_for(1, Increment<view_t>{data});
});
```

core/unit_test/cuda/TestCuda_InterOp_Graph.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+    root.then_parallel_for(1, Increment<view_t>{data});
+  }
+
+ protected:
+  const execution_space exec{};
```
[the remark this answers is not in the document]
romintomasetti: Leftover, removed.

core/unit_test/cuda/TestCuda_InterOp_Graph.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+#include <TestCuda_Category.hpp>
+#include <Kokkos_Core.hpp>
+#include <Kokkos_Graph.hpp>
+
+#include <gtest/gtest.h>
+
+namespace {
+
+template <typename ViewType>
```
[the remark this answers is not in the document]
romintomasetti: I did not wanted it to be hardcoded. For instance, some tests would need an atomic accessor, others not.

core/unit_test/cuda/TestCuda_InterOp_Graph.cpp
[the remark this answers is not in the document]
dalg24: Yes

core/unit_test/cuda/TestCuda_InterOp_Graph.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+ protected:
+  const execution_space exec{};
+  view_t data;
+  std::optional<graph_t> graph;
+};
+
+// This test checks the promises of Kokkos::Graph against its
+// underlying Cuda native objects.
+TEST_F(TEST_CATEGORY_FIXTURE(GraphInterOp), promises_on_native_objects) {
```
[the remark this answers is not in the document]
romintomasetti: Indeed, it's checking pre and post conditions right ?

core/unit_test/cuda/TestCuda_InterOp_Graph.cpp
```diff
@@ -0,0 +1,144 @@
...
+  ASSERT_EQ(graph->native_graph(), cuda_graph);
+  ASSERT_EQ(graph->native_graph_exec(), cuda_graph_exec);
+}
+
+// Count the number of nodes. This is useful to ensure no spurious
+// (possibly empty) node is added.
+TEST_F(TEST_CATEGORY_FIXTURE(GraphInterOp), count_nodes) {
+  graph->instantiate();
+
+  size_t num_nodes = 0;
```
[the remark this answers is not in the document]
romintomasetti: Why ?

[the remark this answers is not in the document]
dalg24: Because that value is meaningless and it expresses better the intent that num_nodes is to be initialized by the CUDA API call below.

core/unit_test/cuda/TestCuda_InterOp_Graph.cpp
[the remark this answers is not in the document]
masterleinad: I agree with @dalg24's comment.

core/unit_test/cuda/TestCuda_InterOp_Graph.cpp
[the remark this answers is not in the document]
romintomasetti: Sure. I thought I'd get some warning related to `Wuninitialized`, so let's see.

core/unit_test/cuda/TestCuda_InterOp_Graph.cpp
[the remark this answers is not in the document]
romintomasetti: OK, done :heavy_check_mark:

dalg24:
```
12: [ RUN      ] cuda_GraphInterOp.instantiation_flags
12: unknown file: Failure
12: C++ exception with description "cudaGraphExecGetFlags(graph->native_graph_exec(), &flags) error( cudaErrorInvalidValue): invalid argument /var/jenkins/workspace/Kokkos_PR-6904/core/unit_test/cuda/TestCuda_InterOp_Graph.cpp:136" thrown in the test body.
12: [  FAILED  ] cuda_GraphInterOp.instantiation_flags (0 ms)
```
in CUDA build with NVC++ as the host compiler

core/unit_test/TestGraph.hpp
```diff
@@ -507,9 +507,20 @@ TEST_F(TEST_CATEGORY_FIXTURE(graph), force_global_launch) {
             false, {"Either the name or pointer or size did not match"}};
       }));
 
+  listen_tool_events(Config::DisableAll());
```
dalg24, line 510: Fine

dalg24:
Boldly ignoring the NVC++ segfaults in the OpenACC and CUDA with NVHPC as host compiler builds

PR #1731, opened 2018-08-08
LayoutTiled: Initial implementation

ndellingwood:
Added LayoutTiled to Kokkos_Layouts;
ViewOffset for LayoutTiled with
operator(), ViewMapping overloads and
tile_subview functions for ranks 2-8.

Initial unit testing on host for ranks 2 and 3.

1. Add unit tests for remaining ranks
2. Add missing ExecSpace testing
3. Consider specializations of LayoutTiled to support default iterate
patterns - update ViewOffset, etc
4. Move Iterate enum type to common location
6. Decide what to name the ViewLayoutTiled file, or where contents go

 Changes to be committed:
	modified:   core/src/Kokkos_Layout.hpp
	new file:   core/src/impl/Kokkos_ViewLayoutTiled.hpp
	new file:   core/unit_test/TestViewLayoutTiled.hpp
	modified:   core/unit_test/serial/TestSerial_Other.cpp

PR #6907, opened 2024-03-29
Remove `Experimental::LayoutTiled` class template and deprecate `is_layouttiled` trait

dalg24:
Rational: Currently not usable (see https://godbolt.org/z/K7q6zz8Pc) because the `ViewMapping` implementation is only provided in a private header that is never included in Kokkos internals.  Removing reduces complexity before the `mdspan` view refactor.

I marked the `Kokkos::is_layouttiled` trait as deprecated and I just removed the `Experimental::LayoutTiled` class template (instead of going through deprecation) because, even though one could use that type directly, there is really noting useful that can be done with it (that is cannot use it as View template argument).  Let me know if you think that is too aggressive.


dalg24:
I forgot to mention neither the trait nor the layout are used in Trilnos outside Kokkos

crtrott:
I pinged Victor to see if there is any use of this he can find, but generally I am ok to rip it out for now. We could ask for feedback at the release briefing.

dalg24:
If that helps deciding, since its introduction in https://github.com/kokkos/kokkos/pull/1731, the `Experimental::LayoutTiled` implementation has never been included outside in the tests.

crtrott:
We couldn't identify any use. Its in experimental namespace. And you had to include a impl/ header, which didn't even work anymore since we made the private/public header distinction. So yank it.

crtrott (approved):
Follow up we should move the `layout_iterate_type_selector` into `Impl` namespace. There is no reason it should be public. Note this thing is used in stuff like ViewCopy to figure out the type of the `MDRangePolicy`.

ldh4 (approved):

dalg24:
CUDA failures were unrelated (machine issues)

PR #6910, opened 2024-04-02
Do not return a copy of the input functor for `Kokkos::Experimental::for_each`

tpadioleau:
If I understand correctly, the Kokkos algorithms match the standard algorithms with an execution policy. From https://eel.is/c++draft/alg.foreach#10, the `for_each` version taking an execution policy should not return a copy of the input functor.

The PR makes all `Kokkos::Experimental::for_each` versions return `void`. Should `Kokkos::Serial` be a special case ?

### DIFF
--- a/algorithms/src/std_algorithms/Kokkos_ForEach.hpp
+++ b/algorithms/src/std_algorithms/Kokkos_ForEach.hpp
@@ -29,49 +29,46 @@ namespace Experimental {
 template <
     class ExecutionSpace, class IteratorType, class UnaryFunctorType,
     std::enable_if_t<Kokkos::is_execution_space_v<ExecutionSpace>, int> = 0>
-UnaryFunctorType for_each(const std::string& label, const ExecutionSpace& ex,
-                          IteratorType first, IteratorType last,
-                          UnaryFunctorType functor) {
-  return Impl::for_each_exespace_impl(label, ex, first, last,
-                                      std::move(functor));
+void for_each(const std::string& label, const ExecutionSpace& ex,
+              IteratorType first, IteratorType last, UnaryFunctorType functor) {
+  Impl::for_each_exespace_impl(label, ex, first, last, std::move(functor));
 }
 
 template <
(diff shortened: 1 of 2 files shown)


masterleinad:
> Should Kokkos::Serial be a special case?

No, we would encourage non-portable code.

masterleinad (approved):

dalg24:
Did you survey if other algorithms had similar issues?
Does it look like an oversight or are they legitimate use of this signature?

fnrizzi:
I think this one is a local slip.
But definitely a global check is useful...

tpadioleau:
> Did you survey if other algorithms had similar issues? Does it look like an oversight or are they legitimate use of this signature?

I only checked a few other algorithms, I could not find other differences. Following this document https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2016/p0024r2.html, `std::for_each` seems to be the only algorithm that was modified due to the introduction of execution policies ?

> Does it make sense? I am not saying we should not do this, but I don't think we should interpret the execution space as the "twin" of std policy. And so kind of decide on a consistent approach everywhere

Makes sense to me, so we agree that all versions of `Kokkos::Experimental::for_each(...)`, even for `Kokkos::Serial`, should be as close as possible to the signature of `std::for_each(std::execution::par_unseq, ...)` ?

fnrizzi:
> Makes sense to me, so we agree that all versions of `Kokkos::Experimental::for_each(...)`, even for `Kokkos::Serial`, should be as close as possible to the signature of `std::for_each(std::execution::par_unseq, ...)` ?

yes i agree with that

dalg24:
Ignoring unrelated HIP runtime failure

dalg24 (approved):
Interpreting Francesco answers as an implicit approval and not seeing any good reason to block on my end.

We need to update the doc and include in the 4.4 changelog

fnrizzi:
Sorry forgot to approve!

PR #6919, opened 2024-04-04
[ci skip] CHANGELOG.md: 4.3.00 update to develop

ndellingwood:


CHANGELOG.md
```diff
@@ -1,5 +1,105 @@
 # CHANGELOG
 
+## [4.3.00](https://github.com/kokkos/kokkos/tree/4.3.00) (2024-03-19)
```
masterleinad, line 3: I noticed that the dates in this file don't really match https://github.com/kokkos/kokkos/releases. Maybe, we just remove those in here in a follow-up?

masterleinad (approved):

CHANGELOG.md
ndellingwood: Yeah, the date you marked is from when I created the changelog PR to the rc. I think removing the dates in a follow up (and excluding in following releases) sounds like a good plan, otherwise that needs to be updated right before final approval+merge

PR #6750, opened 2024-01-25
SYCL: Unroll shuffle loops for top-level parallel_reduce and parallel_scan

masterleinad:
Similar to #6562, unrolling the loops for shuffles improves performance significantly. For a simple dot product benchmark we are seeing on Intel PVC

| elements  | old shuffle (GB/s) | new shuffle (GB/s) | local memory (GB/s) |
| -- | -- | -- | -- |
| 2 |    0.00195408 | 0.00196642 | 0.00195283
| 4 |    0.00416202 | 0.00425014 | 0.00421528
| 8 |    0.00820138 | 0.00832427 | 0.0083891
| 16 |     0.0138029 | 0.0146039 | 0.0142298
| 32 |    0.0294672 | 0.0335499 | 0.0324571
| 64 |    0.0581855 | 0.064666 | 0.0623012
| 128 |  0.119849 | 0.133714 | 0.125332
| 256 |     0.229995 | 0.258772 | 0.237334
| 512 |    0.465701 | 0.509693 | 0.440304
| 1024 |  0.855585 | 1.01661 | 0.734111
| 2048 |    1.49275 | 1.72464 | 1.47962
| 4096 |    2.88409 | 3.56523 | 3.15379
| 8192 |    5.46391 | 6.73428 | 6.30253
| 16384 |    10.6889 | 13.2914 | 12.1061
| 32768 |   20.9248 | 26.2938 | 24.5314
| 65536 |   40.7319 | 49.5028 | 46.1789
| 131072 |   70.2748 | 95.7102 | 86.0199
| 262144 |     103.283 | 143.788 | 130.812
| 524288 |   160.729 | 289.661 | 259.618
| 1048576 |    219.284 | 402.79 | 349.205
| 2097152 |    420.953 | 750.37 | 656.928
| 4194304 |    776.545 | 1291.27 | 1196.56
| 8388608 |     1352.93 | 2073.44 | 1999.47
| 16777216 |   2125.82 | 2919.15 | 2990.47
| 33554432 |    1646.12 | 1686.77 | 1599.83
| 67108864 |    1621.1 | 1643.43 | 1587.29
| 134217728 |      1613.23 | 1607.5 | 1580.22
| 268435456 |    1533.8 | 1537.64 | 1532.46
| 536870912 |    1638.35 | 1641.31 | 1637.63
| 1073741824 |   1671.84 | 1670.2 | 1676.66

[...]

Issue #6914, opened 2024-04-03
CHANGELOG for 4.4


dalg24:
This is the changelog for the upcoming 4.4 release

### Features:
* Add converting constructors from and conversion operators to [`std::mdspan`](https://en.cppreference.com/w/cpp/container/mdspan) [\#6830](https://github.com/kokkos/kokkos/pull/6830) [\#7069](https://github.com/kokkos/kokkos/pull/7069)

#### CUDA:
* nvcc(wrapper): Adding ability to process `--disable-warnings` flag [6936](https://github.com/kokkos/kokkos/issues/6936)
*  Use recommended/max team size functions in Cuda ParallelFor and Reduce constructors [\#6891](https://github.com/kokkos/kokkos/issues/6891) 

#### HIP:
* MI300 support unified memory support [\#6877](https://github.com/kokkos/kokkos/pull/6877) **PATCH RELEASE 4.3.1**
* Use builtin atomics in the HIP backend [\#6882](https://github.com/kokkos/kokkos/pull/6882) [\#7000](https://github.com/kokkos/kokkos/pull/7000)
*  Add expert CMake options to control compilation flags for AMD GPUs [\#7127 ](https://github.com/kokkos/kokkos/pull/7127)

#### SYCL:
* Fix multi-GPU support [\#6887](https://github.com/kokkos/kokkos/pull/6887)
* Improve team_reduce performance [\#6562](https://github.com/kokkos/kokkos/pull/6562) 
* Unroll shuffle loops for top-level parallel_reduce and parallel_scan [\#6750](https://github.com/kokkos/kokkos/pull/6750)
* Fix lock for guarding scratch space in `TeamPolicy` `parallel_reduce` [\#6988](https://github.com/kokkos/kokkos/pull/6988)
* Print submission command queue property [\#7004](https://github.com/kokkos/kokkos/pull/7004)
* SYCL: Add support for Graphs [\#6912](https://github.com/kokkos/kokkos/pull/6912)

#### OpenMPTarget:
* Honor user requested loop ordering in MDRange policy [\#6925](https://github.com/kokkos/kokkos/pull/6925)
* Use mutex lock for guarding scratch space in parallel_scan. [\#6998](https://github.com/kokkos/kokkos/pull/6998)

#### OpenACC:
* Change the default execution behavior of the parallel_for(team-policy) constructs in the OpenACC backend [\#7012](https://github.com/kokkos/kokkos/pull/7012)

#### HPX:
[...]

PR #6924, opened 2024-04-08
SYCL: Fix range in subgroup scan for workgroup_scan

masterleinad:
After https://github.com/kokkos/kokkos/pull/6750, https://github.com/kokkos/kokkos/blob/55c5757502bdc5d5f6b9658a91114dcf56df6a4b/core/src/SYCL/Kokkos_SYCL_ParallelScan_Range.hpp#L51 triggers in `Debug` mode. This has been wrong before but we never noticed. We must only run over the strides that can possibly happen in a subgroup, not the strides possible in the full workgroup.
After this pull request, the implementation matches `workgroup_reduce` better.

### DIFF
--- a/core/src/SYCL/Kokkos_SYCL_ParallelScan_Range.hpp
+++ b/core/src/SYCL/Kokkos_SYCL_ParallelScan_Range.hpp
@@ -35,10 +35,11 @@ void workgroup_scan(sycl::nd_item<dim> item, const FunctorType& final_reducer,
   auto sg               = item.get_sub_group();
   const int sg_group_id = sg.get_group_id()[0];
   const int id_in_sg    = sg.get_local_id()[0];
+  const int local_range = std::min<int>(sg.get_local_range()[0], global_range);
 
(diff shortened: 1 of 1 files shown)


crtrott:
so why did we never get a wrong answer in our testing?

masterleinad:
> so why did we never get a wrong answer in our testing?

Because we still have a
```
if (id_in_sg >= stride) final_reducer.join(&local_value, &tmp);
```
guard which prevented these (garbage) values from ever being used.

dalg24:
Your description refers twice to the same PR.

masterleinad:
> Your description refers twice to the same PR.

Removed the duplication.

dalg24 (approved):

dalg24:
Unrelated HIP runtime failures.

masterleinad:
The `SYCL` build passed.

crtrott (approved):

ndellingwood:

masterleinad:

No, it's a fix for a pull request that is already in the changelog.

PR #6925, opened 2024-04-10
OpenMPTarget: Update loop order in MDRange

rgayatri23:
The PR updates the loop order to honor user requested loop ordering in the MDRange policy of the OpenMPTarget backend. 
It also deletes the unused (and not working) tiling implementation for now. 

FYI - Tiling implementation as-is gives wrong results in MDRange policy for the backend. Looking to fix it in a different PR later.

### DIFF
--- a/core/src/OpenMPTarget/Kokkos_OpenMPTarget_ParallelFor_MDRange.hpp
+++ b/core/src/OpenMPTarget/Kokkos_OpenMPTarget_ParallelFor_MDRange.hpp
@@ -1,0 +1,385 @@
+//@HEADER
+// ************************************************************************
+//
+//                        Kokkos v. 4.0
+//       Copyright (2022) National Technology & Engineering
+//               Solutions of Sandia, LLC (NTESS).
+//
+// Under the terms of Contract DE-NA0003525 with NTESS,
+// the U.S. Government retains certain rights in this software.
+//
+// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
+#ifndef KOKKOS_OPENMPTARGET_PARALLELFOR_MDRANGE_HPP
+#define KOKKOS_OPENMPTARGET_PARALLELFOR_MDRANGE_HPP
+
+#include <omp.h>
+#include <Kokkos_Parallel.hpp>
+#include "Kokkos_OpenMPTarget_MDRangePolicy.hpp"
+#include <OpenMPTarget/Kokkos_OpenMPTarget_Parallel.hpp>
+#include <OpenMPTarget/Kokkos_OpenMPTarget_Parallel_Common.hpp>
+
+//----------------------------------------------------------------------------
+//----------------------------------------------------------------------------
+
+namespace Kokkos {
+namespace Impl {
+
+template <class FunctorType, class... Traits>
+class ParallelFor<FunctorType, Kokkos::MDRangePolicy<Traits...>,
+                  Kokkos::Experimental::OpenMPTarget> {
+ private:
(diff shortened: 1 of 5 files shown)


rgayatri23:
> `MDRangePolicy` has outer and inner direction properties with backend-specific defaults but that can be overridden by the user. The default is up to the backend implementers but we need to see that the user choice is honored. You can look how we added support in the OpenACC backend (we still need to add support for tiles).

I addressed the issue. Separated the MDRange implementation into 2 files for `parallel_for` and `parallel_reduce`. Deleted the non-working and blocked tiling implementation for now. Will add it in a later PR.

core/src/OpenMPTarget/Kokkos_OpenMPTarget_ParallelFor_MDRange.hpp
```diff
@@ -0,0 +1,385 @@
...
+//@HEADER
+
+#ifndef KOKKOS_OPENMPTARGET_PARALLELFOR_MDRANGE_HPP
+#define KOKKOS_OPENMPTARGET_PARALLELFOR_MDRANGE_HPP
+
+#include <omp.h>
+#include <Kokkos_Parallel.hpp>
+#include "Kokkos_OpenMPTarget_MDRangePolicy.hpp"
+#include <OpenMPTarget/Kokkos_OpenMPTarget_Parallel.hpp>
+#include <OpenMPTarget/Kokkos_OpenMPTarget_Parallel_Common.hpp>
```
dalg24, line 24: What do you use from these headers?

dalg24 (approved):
Please update the title line and description to reflect what that PR does.

You might want to open an issue to track what is missing (e.g. support for non-sum reductions, etc.)
There is a lot of improvement/cleanup you can borrow from the OpenACC backend implementation (not asking you to do this here)

core/src/OpenMPTarget/Kokkos_OpenMPTarget_ParallelFor_MDRange.hpp
rgayatri23: Nothing, I removed those.

crtrott (changes requested):
What about having just one loop ordering you collapse but then make the index decisison in the middle:

```c++
if constexpr (std::is_same_v<Order,  OpenMPTargetIterateRight>) {
          if constexpr (std::is_void<typename Policy::work_tag>::value)
            functor(i0, i1, i2);
          else
            functor(typename Policy::work_tag(), i0, i1, i2);
        }
} else {
          if constexpr (std::is_void<typename Policy::work_tag>::value)
            functor(i2, i1, i0);
          else
            functor(typename Policy::work_tag(), i2, i1, i0);
        }
}
```
You need to also swap the loop bounds of course:

```c++
if constexpr (std::is_same_v<Order,  OpenMPTargetIterateRight>) {
    const Index begin_0 = policy.m_lower[0];
    const Index begin_1 = policy.m_lower[1];
    const Index begin_2 = policy.m_lower[2];

    const Index end_0 = policy.m_upper[0];
    const Index end_1 = policy.m_upper[1];
    const Index end_2 = policy.m_upper[2];
} else {
    const Index begin_0 = policy.m_lower[2];
    const Index begin_1 = policy.m_lower[1];
    const Index begin_2 = policy.m_lower[0];

    const Index end_0 = policy.m_upper[2];
    const Index end_1 = policy.m_upper[1];
    const Index end_2 = policy.m_upper[0];
}

crtrott:
Not 100% sure if its much cleaner what I suggested: Damien, Rahul?

dalg24:
It’s elegant but would not work (bounds getting out of scope before the
tightly nested loops). I suppose you could declare the bounds before you
assign their value in the if constexpr or even use a ternary to initialize
them but at that point I think what we did in the OpenACC backend is more
readable.

crtrott (approved):

PR #6926, opened 2024-04-10
Add ROCm 6.0 in the nightly CI

Rombur:
I'd like to have the latest release of ROCm in the nightly CI.

### DIFF
--- a/.jenkins_nightly
+++ b/.jenkins_nightly
@@ -102,6 +102,39 @@ pipeline {
                           '''
                     }
                 }
+                stage('HIP-ROCM-6.0') {
+                    agent {
+                        dockerfile {
+                            filename 'Dockerfile.hipcc'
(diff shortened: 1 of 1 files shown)


dalg24 (commented):
Don't forget to add `[ci skip]` in your commit message next time for this kind of changes.
You don't want an install and follow up step where you build something likes an example or tutorial against the installed package to ensure everything looks fine?

Rombur:
> You don't want an install and follow up step where you build something likes an example or tutorial against the installed package to ensure everything looks fine?

I can add it if you want. I personally don't see the point in doing that.

dalg24 (approved):
Considering that the post step is not present in other builds either.
I still think that it would be good to add but that is beyond the scope of that PR.

crtrott (approved):

masterleinad (approved):

PR #6934, opened 2024-04-11
Deprecate trailing `Proxy` template argument in `Kokkos::Array`

dalg24:
Following up on #6906 and deprecating the `Proxy` template argument in `Kokkos::Array<T, N, Proxy>`.

Rational:
* Further alignment with `std::array<T, N>`, `Kokkos::Array` is provided as a drop-in replacement for the facility provided by the standard library.  One caveat is we still spell it differently with that leading uppercase A...
* This rather obscure facility was solely there to support the view specialization that we removed in #6906.  We deprecate it instead of removing to be safe (I could not find any use when searching through Trilinos).  Doing so will reduce the code complexity and make it trivial to document `Kokkoss::Array` as we can just refer to the standard library documentation (we never bothered documenting it).

### DIFF
--- a/core/src/Kokkos_Array.hpp
+++ b/core/src/Kokkos_Array.hpp
@@ -80,7 +80,11 @@ struct ArrayBoundsCheck<Integral, false> {
 /**\brief  Derived from the C++17 'std::array'.
  *         Dropping the iterator interface.
  */
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
 template <class T = void, size_t N = KOKKOS_INVALID_INDEX, class Proxy = void>
+#else
+template <class T, size_t N>
+#endif
 struct Array {
  public:
   /**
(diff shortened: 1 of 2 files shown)


masterleinad (approved):

nliber (commented):
Do we want to add a check in the primary template that `Proxy == void`, or is that overkill?

nliber (approved):

dalg24:
> Do we want to add a check in the primary template that `Proxy == void`, or is that overkill?

Like `static_assert` or somehow emit a warning?
When would we raise that (for what config)?

nliber (changes requested):
I don't think `<limits>` is used anywhere.

dalg24:
> I don't think `<limits>` is used anywhere.

Not seeing how it relates to this PR.  Can you elaborate?

nliber (approved):

PR #6936, opened 2024-04-12
Add support for `--disable-warnings` flag into `nvcc_wrapper`

### LINKED ISSUE #6874: nvcc_wrapper does not treat '--disable-warnings' properly
If `--disable-warnings` is passed to `nvcc_wrapper`, it cannot process the option properly and eventually throws an error during compilation:

`cc1plus: error: unknown pass warnings specified in '-fdisable'`

It can be reproduced by compiling an arbitrary source file with the option.

rgayatri23:
The PR attempts to fix #6874

### DIFF
--- a/bin/nvcc_wrapper
+++ b/bin/nvcc_wrapper
@@ -233,7 +233,7 @@ do
     cuda_args="$cuda_args $1"
     ;;
   #Handle more known nvcc args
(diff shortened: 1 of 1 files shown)


masterleinad (approved):
We already support the shorthand version `-w`.

dalg24:
@rgayatri23 please add it to the changelog

PR #6938, opened 2024-04-15
Add thread-safety tests

masterleinad:
Part of https://github.com/kokkos/kokkos/pull/6151 that only adds the tests from there.
The tests check that kernels submitted to one execution space instance by multiple threads are not running concurrently.

### DIFF
--- a/core/unit_test/TestExecSpaceThreadSafety.hpp
+++ b/core/unit_test/TestExecSpaceThreadSafety.hpp
@@ -33,8 +33,8 @@ void run_threaded_test(const Lambda1 l1, const Lambda2 l2) {
 #elif !defined(KOKKOS_ENABLE_THREADS) && !defined(KOKKOS_ENABLE_HPX)
 template <class Lambda1, class Lambda2>
 void run_threaded_test(const Lambda1 l1, const Lambda2 l2) {
-  std::thread t1(std::move(l1));
-  std::thread t2(std::move(l2));
+  std::thread t1(l1);
+  std::thread t2(l2);
   t1.join();
   t2.join();
 }
--- a/core/unit_test/TestExecSpaceThreadSafety.hpp
+++ b/core/unit_test/TestExecSpaceThreadSafety.hpp
@@ -98,11 +98,6 @@ void run_exec_space_thread_safety_range() {
 }
 
 TEST(TEST_CATEGORY, exec_space_thread_safety_range) {
-// FIXME_OPENACC
-#ifdef KOKKOS_ENABLE_OPENACC
-  if (std::is_same_v<TEST_EXECSPACE, Kokkos::Experimental::OpenACC>)
-    GTEST_SKIP() << "skipping since test is known to fail for OpenACC";
-#endif
 #ifdef KOKKOS_ENABLE_OPENMPTARGET
   if (std::is_same_v<TEST_EXECSPACE, Kokkos::Experimental::OpenMPTarget>)
     GTEST_SKIP() << "skipping since test is known to fail for OpenMPTarget";
@@ -166,11 +161,6 @@ void run_exec_space_thread_safety_mdrange() {
 }
 
 TEST(TEST_CATEGORY, exec_space_thread_safety_mdrange) {
-// FIXME_OPENACC
-#ifdef KOKKOS_ENABLE_OPENACC
-  if (std::is_same_v<TEST_EXECSPACE, Kokkos::Experimental::OpenACC>)
-    GTEST_SKIP() << "skipping since test is known to fail for OpenACC";
-#endif
 #ifdef KOKKOS_ENABLE_OPENMPTARGET
   if (std::is_same_v<TEST_EXECSPACE, Kokkos::Experimental::OpenMPTarget>)
     GTEST_SKIP() << "skipping since test is known to fail for OpenMPTarget";
@@ -232,11 +222,6 @@ void run_exec_space_thread_safety_team_policy() {
 }
 
 TEST(TEST_CATEGORY, exec_space_thread_safety_team_policy) {
-// FIXME_OPENACC
-#ifdef KOKKOS_ENABLE_OPENACC
-  if (std::is_same_v<TEST_EXECSPACE, Kokkos::Experimental::OpenACC>)
-    GTEST_SKIP() << "skipping since test is known to fail for OpenACC";
-#endif
 // FIXME_OPENMPTARGET
 #ifdef KOKKOS_ENABLE_OPENMPTARGET
   if (std::is_same_v<TEST_EXECSPACE, Kokkos::Experimental::OpenMPTarget>)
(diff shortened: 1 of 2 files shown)


core/unit_test/TestExecSpaceThreadSafety.hpp
```diff
@@ -0,0 +1,537 @@
...
+  auto [iteration, element] = Kokkos::pair<int, int>(host_error_index());
+  ASSERT_EQ(iteration, 0) << " failing at " << iteration << ", " << element;
+  auto host_view =
+      Kokkos::create_mirror_view_and_copy(Kokkos::HostSpace{}, view);
+  for (int i = 0; i < N; ++i)
+    ASSERT_EQ(host_view(i), 2 * M) << " failing at " << i;
+}
+
+TEST(TEST_CATEGORY, exec_space_thread_safety_range) {
+// FIXME_OPENACC
```
Rombur, line 109: It looks like `OpenACC` does not pass any of the tests, so maybe just remove the test at the CMake level like you did for `Serial` and `OpenMP`

Rombur (approved):
Looks good. I just have a small cosmetic comment.

core/unit_test/TestExecSpaceThreadSafety.hpp
```diff
@@ -0,0 +1,537 @@
...
+#include <iostream>
+#include <thread>
+
+#include <Kokkos_Core.hpp>
+
+namespace Test {
+
+#include <Kokkos_Core.hpp>
+#include <iostream>
+#include <thread>
```
dalg24, line 28: Uggg?

dalg24: Do anonymous space and do not include headers within an open namespace

core/unit_test/TestExecSpaceThreadSafety.hpp
dalg24: Ditto

core/unit_test/TestExecSpaceThreadSafety.hpp
dalg24: Do anonymous space and do not include headers within an open namespace

core/unit_test/TestExecSpaceThreadSafety.hpp
dalg24: (Or clearly state why these are treated differently)

core/unit_test/TestExecSpaceThreadSafety.hpp
```diff
@@ -0,0 +1,499 @@
...
+    if (omp_get_thread_num() == 0) l1();
+    if (omp_get_thread_num() == 1) l2();
+  }
+}
+// We cannot run the multithreaded test when threads or HPX is enabled because
+// we cannot launch a thread from inside another thread
+#elif !defined(KOKKOS_ENABLE_THREADS) && !defined(KOKKOS_ENABLE_HPX)
+template <class Lambda1, class Lambda2>
+void run_threaded_test(const Lambda1 l1, const Lambda2 l2) {
+  std::thread t1(std::move(l1));
```
dalg24, line 36: If you care about the moving instead of copying please change that code.  Otherwise just drop the `std::move`.
The lambda arguments are const-qualified so this is always going to copy.  Hence the code in its current version is misleading.

core/unit_test/TestExecSpaceThreadSafety.hpp
```diff
@@ -0,0 +1,499 @@
...
+      "error_index");
+
+  run_threaded_test(
+      [=]() {
+        TEST_EXECSPACE exec;
+        for (int j = 0; j < M; ++j) {
+          Kokkos::parallel_for(
+              "bar", Kokkos::RangePolicy<TEST_EXECSPACE>(exec, 0, N),
+              KOKKOS_LAMBDA(int i) {
+                if (i < N - 1 &&
```
dalg24, line 64: Why make the condition more complicated instead of picking [0, N-1) for the range?

core/unit_test/TestExecSpaceThreadSafety.hpp
```diff
@@ -0,0 +1,499 @@
...
+                    (Kokkos::atomic_load(view.data() + (i + 1)) - view(i)) > 1)
```
dalg24, line 65: Why are you doing 
`Kokkos::atomic_load(view.data() + (i + 1))` instead of 
`Kokkos::atomic_load(&view(i + 1))`?

dalg24: Also I just don't get what this is doing.  Why are we mixing atomic reads and non-atomic writes?

core/unit_test/TestExecSpaceThreadSafety.hpp
```diff
@@ -0,0 +1,499 @@
...
+}
+#else
+template <class Lambda1, class Lambda2>
+void run_threaded_test(const Lambda1 l1, const Lambda2 l2) {
+  l1();
+  l2();
+}
+#endif
+
+void run_exec_space_thread_safety_range() {
```
dalg24, line 49: I stared at it for 10min and I am still not sure what your test does.
This is a hint that comments are a must here.

dalg24 (commented):
I don't even

core/unit_test/TestExecSpaceThreadSafety.hpp
masterleinad: Added comments.

core/unit_test/TestExecSpaceThreadSafety.hpp
masterleinad: We want to update all elements in this kernel and check for the difference with respect to the next element.

core/unit_test/TestExecSpaceThreadSafety.hpp
masterleinad: Could do either `view.data() + i+1` or `&view(i+1)` here.

The idea is that the atomics shouldn't be necessary if everything works as expected but only if we have two kernels running concurrently. Let me just also do the increment with atomics.

core/unit_test/TestExecSpaceThreadSafety.hpp
```diff
@@ -0,0 +1,539 @@
...
+              "bar", Kokkos::RangePolicy<TEST_EXECSPACE>(exec, 0, N),
+              KOKKOS_LAMBDA(int i) {
+                // check that values in the View don't differ by more than one
+                // which would indicate that the other kernel has written to it
+                // while this one is running.
+                if (i < N - 1 && (Kokkos::atomic_load(&view(i + 1)) -
+                                  Kokkos::atomic_load(&view(i))) > 1)
+                  Kokkos::atomic_store(error_index.data(),
+                                       Kokkos::pair{j + 1, i});
+                Kokkos::atomic_inc(&view(i));
```
dalg24, line 76: Why did you choose to increment value rather than set them to different distinguished values in the two kernels?
The incrementing sanity check logic is more complex.

crtrott (approved):
I think the test logic is fine with the increment.

Rombur (approved):

Issue #6941, opened 2024-04-16
Nightly build failures: gcc/9.2, clang/9 with c++20: Kokkos_RandomAccessIterator.hpp:64:17: error: expected identifier...


ndellingwood:
**Describe the bug**

Nightly gcc/9.2 and clang/9 builds with c++20 standard are failing to compile with output:
```
23:41:19 /home/jenkins/jenkins-new/workspace/Kokkos_develop_kokkosdev2_gcc920_pthread_serial_cpp20/kokkos/algorithms/src/std_algorithms/impl/Kokkos_RandomAccessIterator.hpp:64:17: error: expected identifier
23:41:19    64 |   requires(std::is_constructible_v<view_type, OtherViewType>) KOKKOS_FUNCTION
23:41:19       |                 ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
23:41:19 /home/jenkins/jenkins-new/workspace/Kokkos_develop_kokkosdev2_gcc920_pthread_serial_cpp20/kokkos/algorithms/src/std_algorithms/impl/Kokkos_RandomAccessIterator.hpp:64:61: error: expected unqualified-id before ‘)’ token
23:41:19    64 |   requires(std::is_constructible_v<view_type, OtherViewType>) KOKKOS_FUNCTION
23:41:19       |                                                             ^
23:41:19 /home/jenkins/jenkins-new/workspace/Kokkos_develop_kokkosdev2_gcc920_pthread_serial_cpp20/kokkos/algorithms/src/std_algorithms/impl/Kokkos_RandomAccessIterator.hpp:65:7: error: expected constructor, destructor, or type conversion before ‘explicit’
23:41:19    65 |       explicit(!std::is_convertible_v<OtherViewType, view_type>)
```

This follows after merge of #6929  

**Please include the following for a minimal reproducer**

**Reproducer configuration sample** (kokkos-dev-2):
```
module load sems-archive-env sems-archive-cmake/3.17.1 sems-archive-gcc/9.2.0

$KOKKOS_PATH/generate_makefile.bash --with-devices=Pthread,Serial --compiler=g++ --cxxflags="-O3 -Wall -Wunused-parameter -Wshadow -pedantic -Werror -Wsign-compare -Wtype-limits -Wignored-qualifiers -Wempty-body -Wclobbered -Wuninitialized " --cxxstandard="20" --kokkos-path=$KOKKOS_PATH  --no-examples
```

[...]

PR #6942, opened 2024-04-16
Prefer standard C++ feature testing to guard the C++20 requires expression

dalg24:
Fixup for #6929 that broke our nightlies (see #6941)

Temporary fix for our nightly builds so we can make decision on minimum CXX20 compiler requirements when we see fit.

The issue is that Clang 9 and GCC 9 do not support concepts.

### DIFF
--- a/algorithms/src/std_algorithms/impl/Kokkos_RandomAccessIterator.hpp
+++ b/algorithms/src/std_algorithms/impl/Kokkos_RandomAccessIterator.hpp
@@ -59,7 +59,15 @@ class RandomAccessIterator< ::Kokkos::View<DataType, Args...> > {
                                                 ptrdiff_t current_index)
       : m_view(view), m_current_index(current_index) {}
 
-#ifndef KOKKOS_ENABLE_CXX17  // C++20 and beyond
+// FIXME The C++20 requires expression is not supported with Clang 9 and GCC 9
+// The following guards is unsufficient until we increase our minimum CXX20
+// compiler requirements.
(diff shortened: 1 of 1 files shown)


algorithms/src/std_algorithms/impl/Kokkos_RandomAccessIterator.hpp
```diff
@@ -59,7 +59,15 @@ class RandomAccessIterator< ::Kokkos::View<DataType, Args...> > {
...
-#ifndef KOKKOS_ENABLE_CXX17  // C++20 and beyond
+// FIXME The C++20 requires expression is not supported with Clang 9 and GCC 9
+// The following guards is unsufficient until we increase our minimum CXX20
+// compiler requirements.
+//   #ifndef KOKKOS_ENABLE_CXX17  // C++20 and beyond
+// We replace the Kokkos guards with standard C++ feature testing in the
+// meantime.
+#if (defined(__cpp_concepts) && (__cpp_concepts >= 201907L)) && \
+    (defined(__cpp_conditional_explicit) &&                     \
+     (__cpp_conditional_explicit >= 201806L))
```
dalg24, line 70: See https://en.cppreference.com/w/cpp/feature_test

Admittedly, checking the integer literal values might be a bit excessive but I don't think it hurts either.
Let me know if you'd like me to change the new guards.

masterleinad (approved):
Fine with me.

Rombur (approved):

dalg24:
Unrelated failures (HIP runtime and Cuda fail to launch)

PR #6943, opened 2024-04-16
Add `kokkos_swap(Array<T, N>)` specialization

dalg24:
Specializing the swap algorithm for Kokkos arrays was initially proposed in #6697 but we dropped it to focus on the Kokkos swap ADL ordeal. Somehow we overlooked a stray <Kokkos_Swap.hpp> header include in the Kokkos::Array header file.  This PR reintroduce a
`Kokkos::kokkos_swap(Kokkos::Array)` specialization, following closely what the standard library does for `std::swap(std::array)`.

### DIFF
--- a/core/src/Kokkos_Array.hpp
+++ b/core/src/Kokkos_Array.hpp
@@ -133,6 +133,17 @@ struct Array {
   KOKKOS_INLINE_FUNCTION constexpr const_pointer data() const {
     return &m_internal_implementation_private_member_data[0];
   }
+
+ private:
+  template <class U = T>
+  friend KOKKOS_INLINE_FUNCTION constexpr std::enable_if_t<
+      Impl::is_swappable<T>::value>
+  kokkos_swap(Array<T, N>& a,
+              Array<T, N>& b) noexcept(Impl::is_nothrow_swappable_v<T>) {
+    for (std::size_t i = 0; i < N; ++i) {
+      kokkos_swap(a[i], b[i]);
+    }
+  }
 };
 
 #ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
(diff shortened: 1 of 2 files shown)


Rombur (approved):

core/src/Kokkos_Array.hpp
```diff
@@ -185,6 +196,10 @@ struct Array<T, 0> {
   // for default move constructor and move assignment operator.
   // Array( Array && ) = default ;
   // Array & operator = ( Array && ) = default ;
+
+ private:
+  friend KOKKOS_INLINE_FUNCTION constexpr void kokkos_swap(
+      Array<T, 0>&, Array<T, 0>&) noexcept(Impl::is_nothrow_swappable_v<T>) {}
```
nliber, line 202: Why is this dependent on `is_nothrow_swappable<T>` and not just `noexcept(true)`?

It should have the same exception specification as [[array.zero]](https://eel.is/c++draft/array.zero#4).

nliber (dismissed):

core/src/Kokkos_Array.hpp
```diff
@@ -133,6 +133,17 @@ struct Array {
   KOKKOS_INLINE_FUNCTION constexpr const_pointer data() const {
     return &m_internal_implementation_private_member_data[0];
   }
+
+ private:
+  template <class U = T>
```
maartenarnst, line 138: It seems the template argument `U` is not used (?). I may be missing something.

dalg24: You're right I should have used in the SFINAE condition below instead of `T`

dalg24: Not sure why the compiler didn't yell at me

core/src/Kokkos_Array.hpp
dalg24: You're right, I had looked in cppreference and read the framed `noexcept` specifications and missed the next clause that clearly stated:

> For zero-sized arrays, [noexcept](https://en.cppreference.com/w/cpp/language/noexcept_spec) specification:  noexcept

Will fix.  Thanks.

crtrott (approved):

PR #6947, opened 2024-04-17
Deprecate specialization of `Kokkos::pair` for a single element

dalg24:
This specialization is not documented, does not follow the standard library, it is not tested and has no known usage in Trilinos. `Kokkos::pair`, as we generally describe it, was intended as a drop-in replacement for `std::pair`.  Hence, obscure departure from the standard implementation do not look like a good idea.
This PR suggest to deprecate that `T2=void` specialization for degenerate pair that only hold one element.

### DIFF
--- a/core/src/Kokkos_Pair.hpp
+++ b/core/src/Kokkos_Pair.hpp
@@ -413,12 +413,13 @@ KOKKOS_FORCEINLINE_FUNCTION pair<T1&, T2&> tie(T1& x, T2& y) {
   return (pair<T1&, T2&>(x, y));
 }
 
+#ifndef KOKKOS_ENABLE_DEPRECATED_CODE_4
 //
 // Specialization of Kokkos::pair for a \c void second argument.  This
 // is not actually a "pair"; it only contains one element, the first.
 //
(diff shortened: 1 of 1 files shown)


masterleinad (approved):

nliber (approved):

dalg24:
Ignoring unrelated failure (HIP and 2 timeouts SYCL/CUDA)

core/src/Kokkos_Pair.hpp
```diff
@@ -413,12 +413,13 @@ KOKKOS_FORCEINLINE_FUNCTION pair<T1&, T2&> tie(T1& x, T2& y) {
   return (pair<T1&, T2&>(x, y));
 }
 
+#ifndef KOKKOS_ENABLE_DEPRECATED_CODE_4
```
nmm0, line 416: Should it be `#ifdef`? I'm getting build failures when `KOKKOS_ENABLE_DEPRECATED_CODE_4` is OFF

masterleinad: Looks like it.

PR #6951, opened 2024-04-17
Serial: Use the provided execution space instance in TeamPolicy

masterleinad:
https://kokkosteam.slack.com/archives/C5BGU5NDQ/p1713386069368429 noticed serialization of `TeamPolicy` kernels from different `Serial` execution space instances. The culprit was that `TeamPolicy` returned the default executions pace instance instead of the one provided in the constructor.

### DIFF
--- a/core/src/Serial/Kokkos_Serial_Parallel_Team.hpp
+++ b/core/src/Serial/Kokkos_Serial_Parallel_Team.hpp
@@ -37,6 +37,8 @@ class TeamPolicyInternal<Kokkos::Serial, Properties...>
   int m_league_size;
   int m_chunk_size;
 
+  Kokkos::Serial m_space;
+
  public:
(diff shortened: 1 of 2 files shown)


PhilMiller (approved):
Change looks good to me as written.

It doesn't look like any corresponding changes need to be made for `RangePolicy` or `MDRangePolicy`, since those are taken directly from core Kokkos definitions.

PhilMiller:
(re my review: I helped debug this)

crtrott (approved):
We still need a test for this though.

masterleinad:
> We still need a test for this though.

Here you go!

ndellingwood:
I think this fix merits a changelog entry for 4.4 #6914 ?

masterleinad:
> I think this fix merits a changelog entry for 4.4 #6914 ?

Will be part of the patch release.

PR #6954, opened 2024-04-18
Introduce `KOKKOS_DEDUCTION_GUIDE` macro to allow user-defined deduction guide in device code for clang compiler

tpadioleau:
We recently came across a compiler error when using the deduction guide of `TeamThreadMDRange` on a HIP compilation error

```
error: reference to __host__ function '<deduction guide for TeamThreadMDRange><Kokkos::Impl::HIPTeamMember, int, int>' in __host__ __device__ function
```

I could reproduce it on godbolt https://godbolt.org/z/GTdeY77KT. My understanding is that clang requires the user-defined deduction guide to be annotated to be used inside a device function.

Based on the suggestions on slack, this PR annotates the user-defined deduction guides using a new macro `KOKKOS_DEDUCTION_GUIDE`. I have used this macro to annotate guides associated to constructors annotated `KOKKOS_FUNCTION`.

Should we introduce tests for that ?

cc @etiennemlb


core/src/setup/Kokkos_Setup_HIP.hpp
```diff
@@ -27,6 +27,8 @@
 #define KOKKOS_LAMBDA [=] __host__ __device__
 #define KOKKOS_CLASS_LAMBDA [ =, *this ] __host__ __device__
 
+#define KOKKOS_DEDUCTION_GUIDE __host__ __device__
```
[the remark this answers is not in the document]
tpadioleau: I understand your point, just wondering if annotating something that is not a function by `KOKKOS_FUNCTION` would be confusing ?

dalg24:
Ok, plz update title accordingly

core/src/setup/Kokkos_Setup_HIP.hpp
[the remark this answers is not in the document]
masterleinad: It just follows the semantics for deduction guides so it's as confusing and more consistent.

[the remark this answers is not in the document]
masterleinad: Also, we should limit the number of (public) macros we define since they are a pain to deal with later if we consider removing them since we can't deprecate them.

tpadioleau:
Back in draft the time to add tests

core/src/setup/Kokkos_Setup_HIP.hpp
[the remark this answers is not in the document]
Rombur: > , using KOKKOS_FUNCTION instead of introducing a new macro matches more closely the intent of deduction guides (by replicating the constructor's signature).

I agree with @masterleinad on this. 

I don't really see how introducing a new macro improves the code. Worse it makes a bug more likely. Both this PR and mdspan fix the issue for HIP but not for CUDA clang...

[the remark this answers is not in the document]
tpadioleau: Indeed i forgot the case of CUDA clang.

Is it safe though to assume that compilers need the same annotations between deduction guides and constructors ?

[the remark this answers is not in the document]
masterleinad: > Is it safe though to assume that compilers need the same annotations between deduction guides and constructors ?

That's what we are observing, right?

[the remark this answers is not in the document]
tpadioleau: So far yes i agree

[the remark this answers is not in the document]
dalg24: > It just follows the semantics for deduction guides

I don't understand that argument.  Did you mean **syntax**?  Please clarify.

To the consistency argument, I would like to point out there is prior art in Kokkos, with `KOKKOS_DEFAULTED_FUNCTION` and `KOKKOS_INLINE_DELETED_FUNCTION`
<details>
<summary>(Blame game for naming that `KOKKOS_INLINE_DELETED_FUNCTION` macro)</summary>
https://github.com/kokkos/kokkos/commit/3a439317142573f0f6b39e2cd61934fb5e56ca72
</details>

You could argue that these are not needed any more and we are stuck with them, but I also want to look back and see that these were also special syntaxes for the compiler for which it was also not clear what to do about `__host__ __device__` annotations (especially for deleted functions!).  Experience has shown that compilers behavior changed and we had to discriminate workarounds based on vendor toolchain versions.

If we use `KOKKOS_FUNCTION` to decorate our deduction guides, teach our users to do so, and there is even only one compiler that we need to support and that does not like the host device annotation, then we will have to do the distinction.  We will be able to fix it in our codebase but it will be harder to propagate downstream.  
That said I have no evidence it will break and I haven't surveyed what currently works.

> Worse it makes a bug more likely. Both this PR and mdspan fix the issue for HIP but not for CUDA clang...

I don't see how it makes it more bug prone.  Please elaborate.
Regarding Clang+CUDA, did you check that we see the issue when the annotation are missing.  I would be surprised if that was not considered in the reference implementation of `mdspan`.  That is possible but I expect someone would have asked the question.  @crtrott do you remember?

I don't like the proliferation of public macros either but I tend to think it is justified in this case.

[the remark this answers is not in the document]
Rombur: > I don't see how it makes it more bug prone. Please elaborate.

Different people wrote different versions of this PR and they both forgot to fix CUDA clang. If you didn't introduce the new macro and instead use `KOKKOS_FUNCTION`, the PR would have fixed the issue.

> Regarding Clang+CUDA, did you check that we see the issue when the annotation are missing

If you look at the godbolt code that @tpadioleau posted. You can see that he is using CUDA clang to reproduce the error not HIP.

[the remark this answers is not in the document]
dalg24: > > Regarding Clang+CUDA, did you check that we see the issue when the annotation are missing
> 
> If you look at the godbolt code that @tpadioleau posted. You can see that he is using CUDA clang to reproduce the error not HIP.

Haha damn.  Missed the compiler explorer link, I was on my cell.

crtrott:
I believe a new macro is warranted. NVCC currently does not require the markup on constructors (though it works with them too). But my experience with markup on defaulted ctors is a thing which I believe warrants caution here. NVCC changed its requirements regarding requiring/allowing markup on defaulted functions over time. And generally speaking: deduction guides are NOT functions. In fact the signatures don't even need to match the signatures of any existing CTOR. 

E.g. look at this: https://godbolt.org/z/fYPhK5Ec5

```
template <class T>
struct Foo {
    __host__ __device__ Foo(float ,int = 0) {}
};

template <class T>
__host__ __device__ Foo(T) -> Foo<T>; 
```

The deduction guide looks nothing like the ctor, but it will work as long as you try to construct from something convertible to float. 

So it's not a function: it's a rule. Introducing a new macro for that seems warranted to me.

crtrott (approved):

crtrott:
There is a real issue with OpenMPTarget @rgayatri23 can you take a look too?

core/src/setup/Kokkos_Setup_Cuda.hpp
```diff
@@ -56,6 +56,10 @@
 #define KOKKOS_LAMBDA [=] __host__ __device__
 #define KOKKOS_CLASS_LAMBDA [ =, *this ] __host__ __device__
 
+#if defined(__clang__)
```
dalg24, line 59:
```suggestion
#ifdef KOKKOS_COMPILER_CLANG
```

core/unit_test/TestNestedReducerCTAD.cpp
```diff
@@ -0,0 +1,248 @@
...
+          Kokkos::StdPartitionPoint<IndexType, MemorySpace>;
+      using ValueType = ReducerTypeExpected::value_type;
+      Kokkos::View<ValueType, MemorySpace> view;
+      Kokkos::StdPartitionPoint reducer(view);
+      check_types<ReducerTypeExpected>(team_handle, reducer);
+    }
+  }
+
+  static void test_ctad_inside_parallel_for() {
+    Kokkos::parallel_for(TeamPolicy(0, Kokkos::AUTO), TestNestedReducerCTAD());
```
dalg24, line 244: Do we need the parallel_for?  Are we not able reproduce the issue by simply using CTAD within a KOKKOS_FUNCTION?

crtrott:
@tpadioleau the OpenMPTarget failure looks real. Not sure when @rgayatri23 will have time so if you can track it down that would be great. Possibly we can just disable the test for now for OpenMPTarget.

core/unit_test/TestNestedReducerCTAD.cpp
crtrott: If you don't compile with RDC I think it would not necessarily compile the function depending on the optimization level

dalg24: Didn't you manage to reproduce on the compiler explorer?  Did you actually write a kernel launch?

core/unit_test/TestNestedReducerCTAD.cpp
```diff
@@ -0,0 +1,248 @@
...
+    {
+      using ReducerTypeExpected =
+          Kokkos::MaxFirstLocCustomComparator<ScalarType, IndexType,
+                                              FakeComparator, MemorySpace>;
+      using ValueType = ReducerTypeExpected::value_type;
+      Kokkos::View<ValueType, MemorySpace> view;
+      FakeComparator comparator;
+      Kokkos::MaxFirstLocCustomComparator reducer(view, comparator);
+      check_types<ReducerTypeExpected>(team_handle, reducer);
+    }
```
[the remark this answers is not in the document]
rgayatri23: Right, we will have to implement `MaxFirstLocCustomComparator` for the OpenMPTarget backend.

rgayatri23:
@tpadioleau - Can you disable that test for the OpenMPTarget backend for now?

core/unit_test/TestNestedReducerCTAD.cpp
tpadioleau: If I put the same code in an anonymous namespace, the compiler does not generate any assembly, https://godbolt.org/z/8fxEnM6P1.

Do you prefer that I remove the anonymous namespace and make a dedicated executable for this test ?

core/src/setup/Kokkos_Setup_Cuda.hpp
tpadioleau: Should we actually also add the same guard for the HIP backend ?

core/unit_test/TestNestedReducerCTAD.cpp
dalg24: We don't care about assembly being generated.  Don't.

tpadioleau: I meant the compiler does not generate any error if I remove the call to `parallel_for`.

core/src/setup/Kokkos_Setup_Cuda.hpp
tpadioleau: It does not work on a CUDA clang compilation. My understanding is that the macro `KOKKOS_COMPILER_CLANG` is not yet defined at this stage.

Files in the core/src/setup directory should only depend on the backend, not the compiler right ?

dalg24: Oh you're right, it only gets defined later, these setup files get included at the top of the Kokkos_Macros.hpp header.  Sorry.

core/unit_test/TestNestedReducerCTAD.cpp
dalg24: OK.  Then withdrawn.

tpadioleau:
> @tpadioleau - Can you disable that test for the OpenMPTarget backend for now?

Done in d954dc4b2aa790740b8db871240dfe55daab8c30

I also notice a failure on the OpenACC backend, should I disable it ?

core/src/setup/Kokkos_Setup_Cuda.hpp
```diff
@@ -56,6 +56,10 @@
...
+#define KOKKOS_DEDUCTION_GUIDE __host__ __device__
+#endif
```
masterleinad, line 61: I'd prefer if we removed the guard. `nvcc` seems to be fine with it as well.

masterleinad:
> I also notice a failure on the OpenACC backend, should I disable it ?

The errors look like
> /var/jenkins/workspace/Kokkos_PR-6954/core/src/OpenACC/Kokkos_OpenACC_ParallelReduce_Team.hpp", line 242: error: no instance of constructor "Kokkos::Sum<Scalar, Space>::Sum [with Scalar=<unnamed>::TestNestedReducerCTAD::ScalarType, Space=<unnamed>::TestNestedReducerCTAD::MemorySpace]" matches the argument list
    ValueType tmp = ValueType();
                    ^
          detected during:
            instantiation of "void Kokkos::parallel_reduce(const Kokkos::Impl::TeamVectorRangeBoundariesStruct<iType, Kokkos::Impl::OpenACCTeamMember> &, const Lambda &, ValueType &) [with iType=int, Lambda=<unnamed>::TestNestedReducerCTAD::FakeFunctor<int>, ValueType=const Kokkos::Sum<<unnamed>::TestNestedReducerCTAD::ScalarType, <unnamed>::TestNestedReducerCTAD::MemorySpace>]" at line 46 of "/var/jenkins/workspace/Kokkos_PR-6954/core/unit_test/TestNestedReducerCTAD.cpp"
            instantiation of "void <unnamed>::TestNestedReducerCTAD::check_types<ReducerTypeExpected,TeamHandle,ReducerTypeToCheck>(const TeamHandle &, const ReducerTypeToCheck &) [with ReducerTypeExpected=Kokkos::Sum<<unnamed>::TestNestedReducerCTAD::ScalarType, <unnamed>::TestNestedReducerCTAD::MemorySpace>, TeamHandle=<unnamed>::TestNestedReducerCTAD::TeamHandle, ReducerTypeToCheck=Kokkos::Sum<<unnamed>::TestNestedReducerCTAD::ScalarType, <unnamed>::TestNestedReducerCTAD::MemorySpace>]" at line 55 of "/var/jenkins/workspace/Kokkos_PR-6954/core/unit_test/TestNestedReducerCTAD.cpp"

and we have to do something about it. If we can't figure it out, we could consider disabling the test for `OpenACC`.

nliber (approved):

nliber:
@masterleinad @tpadioleau I think the error in https://github.com/kokkos/kokkos/pull/6954#issuecomment-2102675742 may be a legitimate bug unrelated to deduction guides.

`ValueType` is deducing to `Sum<...>` (which doesn't have a default constructor), but there is no OpenACC `parallel_reduce` for `TeamVectorRangeBoundariesStruct` with a constrained overload for a reducer.

Contrast this with the `parallel_reduce` overloads for `TeamThreadRangeBoundariesStruct` and `ThreadVectorRangeBoundariesStruct` which do have different overloads for reducers and non-reducers.

masterleinad:
> @masterleinad @tpadioleau I think the error in [#6954 (comment)](https://github.com/kokkos/kokkos/pull/6954#issuecomment-2102675742) may be a legitimate bug unrelated to deduction guides.

I was sure that it was unrelated to this pull request. 🙂 We still have to fix the CI by disabling the test with a FIXME for OpenACC.

tpadioleau:
@masterleinad @nliber Sorry for my late reply and thank you for looking at the issue.

In the meantime the OpenACC issue is solved I could have a look at how to slightly change the test. I have in mind to remove the inner parallel_reduce. I know that I could not remove the outer parallel_for without breaking the test.

core/src/setup/Kokkos_Setup_Cuda.hpp
tpadioleau: I am fine with both solutions though this introduces a change for a working compiler.

masterleinad:
> In the meantime the OpenACC issue is solved I could have a look at how to slightly change the test. I have in mind to remove the inner parallel_reduce. I know that I could not remove the outer parallel_for without breaking the test.

I don't think anyone is working on a fix for the OpenACC issue.

Issue #6842, opened 2024-02-26
Inconsistency in the behavior of `create_mirror_view`


pzehner:
We observed a difference of behavior for `Kokkos::create_mirror_view` on the host memory space when the view to mirror is of constant items.

If `create_mirror_view` is called *without an execution space*, the mirrored view is duplicated, and its elements are not constant.
Otherwise, if it is called *with an execution space*, the mirrored view is the same as the input view, and its elements are constant.

Here is a MWE:

```cpp
// test.cpp

#include <cassert>
#include <Kokkos_Core.hpp>

void without_exec_space(Kokkos::View<const int*, Kokkos::HostSpace> a)
{
    // call without execution space
    auto b = Kokkos::create_mirror_view(a);

    static_assert(
            Kokkos::SpaceAccessibility<Kokkos::HostSpace, decltype(a)::memory_space>::accessible);

    // the mirrored view doesn't contain const ints anymore
    static_assert(std::is_same_v<decltype(b(0)), int&>);

    // views are different
    assert(b.data() != a.data());
}

void with_exec_space(Kokkos::View<const int*, Kokkos::HostSpace> a)
{
    // call with execution space
    auto b = Kokkos::create_mirror_view(Kokkos::DefaultHostExecutionSpace(), a);

    static_assert(std::is_same_v<
                  Kokkos::DefaultHostExecutionSpace::memory_space,
                  decltype(a)::memory_space>);
    static_assert(
            Kokkos::SpaceAccessibility<Kokkos::HostSpace, decltype(a)::memory_space>::accessible);

    // the mirrored view contains const ints
    static_assert(std::is_same_v<decltype(b(0)), const int&>);

    // views are the same
    assert(b.data() == a.data());
}

int main(int argc, char* argv[]) {
    Kokkos::initialize(argc, argv);
    {
        Kokkos::View<int*, Kokkos::HostSpace> a("A", 10);
        without_exec_space(a);
        with_exec_space(a);
    }
    Kokkos::finalize();
}
```

```cmake
# CMakeLists.txt

cmake_minimum_required(VERSION 3.18)

project(
    TestCreateMirrorView
    LANGUAGES CXX
)

find_package(Kokkos)

add_executable(
    test
    test.cpp
)

target_link_libraries(
    test
    PRIVATE
        Kokkos::kokkos
)
```

Any of the two behaviors may be valid, this is not the point of this issue, but only one should prevail.

[...]

PR #5533, opened 2022-10-07
Cleanup create_mirror* using C++17

masterleinad:
Follow-up to https://github.com/kokkos/kokkos/pull/5488.

PR #6955, opened 2024-04-19
Refactor: Replace SFINAE by `if constexpr` for `create_mirror*` functions

pzehner:
This PR is a refactor without side effects.

The idea is to simplify the code of the `create_mirror*` functions by applying techniques of the C++ 17 standard. A refactor with the similar approach was drafted by @masterleinad in https://github.com/kokkos/kokkos/pull/5533 about two years ago, but never merged.

The multiple overloaded functions of `Impl::create_mirror`, `Impl::create_mirror_view`, and `Impl::create_mirror_view_and_copy` that use SFINAE are replaced by one function that uses `if constexpr` instead (for Kokkos views, offset views, dynamic views and dynamic rank views). Eventually, the return type of these functions is replaced by `auto`, and decided depending on which branch of the `if constexpr` is used.

The logic of the public `create_mirror*` functions (which is sometimes redundant with the logic of the private functions) is not altered in this PR and will be assessed in an upcoming one.

This PR is part of a of a larger effort which aims to assess https://github.com/kokkos/kokkos/issues/6842.

We successfully built and tested the refactored code for the following configurations:

| Compiler      | Version  | *Min. ver.* | Backend | Version | *Min. ver.* |
|---------------|----------|-------------|---------|---------|-------------|
| Clang         | 14.0.0   | *8.0.0*     | Serial  |         |             |
| GCC           | 8.4.0    | *8.2.0*     | Serial  |         |             |
| GCC           | 8.4.0    | *8.2.0*     | OpenMP  |         |             |
| GCC           | 8.4.0    | *8.2.0*     | CUDA    | 11.4    | *11.0*      |
| Intel Classic | 2023.2.1 | *19.0.5*    | OpenMP  |         |             |
| Intel LLVM    | 2023.2.1 | *2021.1.1*  | OpenMP  |         |             |
| Intel LLVM    | 2024.0.2 | *2022.0.0*  | SYCL    |         |             |
| HIP           | 5.2.0    | *5.2.0*     | ROCM    |         |             |

where "Compiler" is the used compiler, next "Version" is the version of the compiler, next "Min. Vers." is the minimum version of the compiler for Kokkos 4.3.0 according to the documentation, "Backend" is the name of the backend, next "Version" is the version for the backend if necessary, and next "Min. Vers." is the minimum version of the backend supported.

Unfortunately, we could not test any earlier version of the compilers, nor of the backends, on our test machines. Also, please note that some configurations could not be used:

- Clang with OpenMP or CUDA;
- MSVC; and
- NVHPC.

Finally, the SYCL configuration was built, but not tested.


containers/src/Kokkos_DynRankView.hpp
```diff
@@ -2058,71 +2044,39 @@ inline auto create_mirror(
...
-
-template <class T, class... P, class... ViewCtorArgs,
-          class = std::enable_if_t<
-              Impl::ViewCtorProp<ViewCtorArgs...>::has_memory_space>>
-inline std::enable_if_t<
-    Kokkos::is_space<
-        typename Impl::ViewCtorProp<ViewCtorArgs...>::memory_space>::value &&
+            typename DynRankView<T, P...>::HostMirror::data_type>::value
+        ) {
+      return src;
```
masterleinad, line 2063: Is this the same type as `typename DynRankView<T, P...>::HostMirror`? If we don't intend to change behavior, we should be explicit about keeping the same return types.

containers/src/Kokkos_DynRankView.hpp
```diff
@@ -2058,71 +2044,39 @@ inline auto create_mirror(
...
-        typename Impl::ViewCtorProp<ViewCtorArgs...>::memory_space, T,
-        P...>::view_type>
-create_mirror_view(const Kokkos::DynRankView<T, P...>& src,
-                   const typename Impl::ViewCtorProp<ViewCtorArgs...>&) {
-  return src;
+            P...>::is_same_memspace
+        ) {
+      return src;
+    } else {
+      return Kokkos::Impl::create_mirror(src, arg_prop);
```
masterleinad, line 2075:
```suggestion
      return typename Impl::MirrorDRViewType<
        typename Impl::ViewCtorProp<ViewCtorArgs...>::memory_space, T,
        P...>::view_type{Kokkos::Impl::create_mirror(src, arg_prop)};
```

containers/src/Kokkos_DynRankView.hpp
```diff
@@ -1933,15 +1933,10 @@ struct MirrorDRVType {
...
-    const DynRankView<T, P...>& src,
-    const Impl::ViewCtorProp<ViewCtorArgs...>& arg_prop,
-    std::enable_if_t<!Impl::ViewCtorProp<ViewCtorArgs...>::has_memory_space>* =
-        nullptr) {
-  using src_type = DynRankView<T, P...>;
-  using dst_type = typename src_type::HostMirror;
 
+// collection of static asserts for create_mirror
+template <class... ViewCtorArgs>
+void check_view_ctor_args_create_mirror_drv() {
```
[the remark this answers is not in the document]
pzehner: We re-used the `check_view_ctor_args*` functions from `core`.

pzehner:
The code style was fixed using `clang-format`.

pzehner:
Merging done. For merging, please squash the commits, as the history of this branch is quite messy.

masterleinad:
> Merging done. For merging, can you squash the commits, as the history of this branch is quite messy?

Yes, we normally squash commits before merging.

pzehner:
> we should test with Trilinos/Sacado/Stokhos just to make sure this doesn't break their customizations.

Could you recommend any build script for Trilinos, in order to test it?

crtrott:
Retest this please!

pzehner:
@crtrott Could you validate this PR? It seems the failing jobs are due to external factors.

crtrott (approved):

PR #6958, opened 2024-04-22
Replace deprecated sycl::device_ptr/sycl::host_ptr

masterleinad:
See https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/supported/sycl_ext_intel_usm_address_spaces.asciidoc#feature-test-macro. `sycl::ext::intel::device_ptr/host_ptr` was introduced in https://github.com/intel/llvm/pull/7680 and `sycl::device_ptr/host_ptr` was removed in https://github.com/intel/llvm/pull/13240.
This pull request introduces aliases in `Kokkos::Impl::SYCLTypes` based on the level of feature support.

### DIFF
--- a/core/src/SYCL/Kokkos_SYCL_Instance.cpp
+++ b/core/src/SYCL/Kokkos_SYCL_Instance.cpp
@@ -166,7 +166,7 @@ int SYCLInternal::acquire_team_scratch_space() {
   return current_team_scratch;
 }
 
-sycl::device_ptr<void> SYCLInternal::resize_team_scratch_space(
+sycl::ext::intel::device_ptr<void> SYCLInternal::resize_team_scratch_space(
     int scratch_pool_id, std::int64_t bytes, bool force_shrink) {
   // Multiple ParallelFor/Reduce Teams can call this function at the same time
   // and invalidate the m_team_scratch_ptr. We use a pool to avoid any race
@@ -251,7 +251,8 @@ void SYCLInternal::finalize() {
   m_queue.reset();
 }
 
-sycl::device_ptr<void> SYCLInternal::scratch_space(const std::size_t size) {
+sycl::ext::intel::device_ptr<void> SYCLInternal::scratch_space(
+    const std::size_t size) {
   if (verify_is_initialized("scratch_space") &&
       m_scratchSpaceCount < scratch_count(size)) {
     auto mem_space = Kokkos::Experimental::SYCLDeviceUSMSpace(*m_queue);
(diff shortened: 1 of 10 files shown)


masterleinad:
The SYCL build passes.

core/src/setup/Kokkos_Setup_SYCL.hpp
```diff
@@ -45,4 +45,21 @@
 #define KOKKOS_IMPL_SYCL_GET_MULTI_PTR(accessor) accessor.get_pointer()
 #endif
 
+// FIXME_SYCL Use type directly once it has stabilized in SYCL.
+namespace Kokkos::Impl::SYCLTypes {
+#ifndef SYCL_EXT_INTEL_USM_ADDRESS_SPACES
+#error SYCL_EXT_INTEL_USM_ADDRESS_SPACES undefined!
+#elif SYCL_EXT_INTEL_USM_ADDRESS_SPACES >= 2
+template <typename T>
+using device_ptr = sycl::ext::intel::device_ptr<T>;
```
dalg24, line 54: I would prefer if didn't introduced that nested typename and renamed 
`Kokkos::Impl::SYCLTypes::{host,device}_ptr` to `Kokkos::Impl::sycl_{host,device}_ptr` which I find more readable and consistent with common practice throughout the codebase.

dalg24: It also makes most if not all of the changes in this PR be simply
`sycl::device_ptr -> sycl_device_ptr` and so on.

masterleinad: I don't have a strong opinion. If that gets me your approval, I'm happy to change it.

masterleinad: See 6333d7d

core/src/SYCL/Kokkos_SYCL_Instance.cpp
```diff
@@ -166,7 +166,7 @@ int SYCLInternal::acquire_team_scratch_space() {
   return current_team_scratch;
 }
 
-sycl::device_ptr<void> SYCLInternal::resize_team_scratch_space(
+Kokkos::Impl::sycl_device_ptr<void> SYCLInternal::resize_team_scratch_space(
```
dalg24, line 169: You don't need the `Kokkos::Impl::` qualification?
Same question everywhere

dalg24 (approved):
I would not qualify when not necessary.  I find it unnecessarily verbose.

core/src/SYCL/Kokkos_SYCL_Instance.cpp
masterleinad: Yes, I do. I only qualified where necessary. `Kokkos_SYCL_Instance.[cpp|hpp]` implementations are in a `Kokkos::Experimental::Impl` namespace. As soon as we remove `Experimental` from the `SYCL` backend, it wouldn't be necessary anymore.

masterleinad: Then I need to qualify the other half of the occurrences.

masterleinad:
The SYCL build passes.

crtrott (approved):

ndellingwood:

PR #6959, opened 2024-04-23
SYCL: Avoid deprecated floating-point number abs overloads

masterleinad:
The floating-point number overloads for `sycl::abs` were deprecated in https://github.com/intel/llvm/pull/11774, see https://github.com/steffenlarsen/llvm/blob/4497895152baca8212549e645d1d7c5c9ce96988/sycl/include/sycl/builtins_legacy_scalar.hpp#L468-L474.

### DIFF
--- a/core/src/Kokkos_MathematicalFunctions.hpp
+++ b/core/src/Kokkos_MathematicalFunctions.hpp
@@ -278,7 +278,7 @@ KOKKOS_INLINE_FUNCTION long long abs(long long n) {
 }
 KOKKOS_INLINE_FUNCTION float abs(float x) {
 #ifdef KOKKOS_ENABLE_SYCL
-  return sycl::fabs(x);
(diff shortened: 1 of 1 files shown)


core/src/Kokkos_MathematicalFunctions.hpp
```diff
@@ -277,12 +277,20 @@ KOKKOS_INLINE_FUNCTION long long abs(long long n) {
 #endif
 }
 KOKKOS_INLINE_FUNCTION float abs(float x) {
+#ifdef KOKKOS_ENABLE_SYCL
+  return sycl::fabs(x);
```
dalg24, line 281:
```suggestion
  return sycl::fabs(x);  // sycl::abs is only provided for integral types
```

dalg24 (approved):
Consider adding a comment (both places)

crtrott (approved):

crtrott:
Why oh why ....

PR #6964, opened 2024-04-24
Add Tutorial for Matrix Operations Handling 

### LINKED ISSUE #6963: Add Tutorial for Matrix Operations Handling
Adding tutorial involving matrix multiplication
 - This example demonstrates handling memory spaces and utilizing parallel computation for a matrix-matrix multiplication operation
 - Includes Makefile and CMakeLists.txt

khoinguyyyen:
Adding tutorial involving matrix multiplication 
resolves #6963 

This example demonstrates handling memory spaces and utilizing parallel computation for a matrix-matrix multiplication operation
Includes Makefile and CMakeLists.txt

### DIFF
--- a/example/tutorial/04_simple_memoryspaces_matrix/CMakeLists.txt
+++ b/example/tutorial/04_simple_memoryspaces_matrix/CMakeLists.txt
@@ -1,0 +1,9 @@
+
+INCLUDE_DIRECTORIES(${CMAKE_CURRENT_BINARY_DIR})
+INCLUDE_DIRECTORIES(${CMAKE_CURRENT_SOURCE_DIR})
+
+# This is a tutorial, not a test, so we don't ask CTest to run it.
+KOKKOS_ADD_EXECUTABLE(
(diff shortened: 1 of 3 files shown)


crtrott:
Hi,

thanks for the contribution. Generally in tutorials we are trying to have a specific educational thing we want to convey. In the existing memory space tutorial we are introducing the host mirror mechanism. This example does the same thing for a different type of kernel (matrix multiply). However, the title seems to indicate that the example wants to teach something about matrix multiply? But the kernel is not how we would recommend even a simple one. So the question is what is the intended takeaway from this example compared to the existing one.

ajpowelsnl:
Hi @khoinguyyyen -- does your tutorial provide concepts beyond those in the matrix - vector multiply [example](https://github.com/kokkos/kokkos-tutorials/tree/main/Exercises/03) (using `HostMirror`)?  If not, may we close this PR?

ajpowelsnl:
Closing PR because no response from authors

PR #6965, opened 2024-04-25
Do not require OpenMP support for languages other than CXX

dalg24:
This change was prompted by a discussion on Slack https://kokkosteam.slack.com/archives/C5BGU5NDQ/p1714063365919929

Specify CXX component when searching for OpenMP so that OpenMP support is not required for other languages with CMake.

One caveat with that change is that finding the OpenMP dependency downstream now requires a CMake minimum version of 3.10

The CMake module documentation can be found at https://cmake.org/cmake/help/latest/module/FindOpenMP.html

### DIFF
--- a/cmake/kokkos_tpls.cmake
+++ b/cmake/kokkos_tpls.cmake
@@ -103,13 +103,13 @@ if (Kokkos_ENABLE_IMPL_MDSPAN AND Kokkos_ENABLE_MDSPAN_EXTERNAL)
 endif()
 
 IF (Kokkos_ENABLE_OPENMP)
-  find_package(OpenMP REQUIRED)
+  find_package(OpenMP REQUIRED COMPONENTS CXX)
   # FIXME_TRILINOS Trilinos doesn't allow for Kokkos to use find_dependency
   # so we just append the flags here instead of linking with the OpenMP target.
   IF(KOKKOS_HAS_TRILINOS)
     COMPILER_SPECIFIC_FLAGS(DEFAULT ${OpenMP_CXX_FLAGS})
   ELSE()
-    KOKKOS_EXPORT_CMAKE_TPL(OpenMP REQUIRED)
+    KOKKOS_EXPORT_CMAKE_TPL(OpenMP REQUIRED COMPONENTS CXX)
   ENDIF()
 ENDIF()
 


masterleinad (approved):

dalg24:
(Haven't tested on a Cuda+OpenMP build with `Kokkos_ENABLE_COMPILE_AS_CMAKE_LANGUAGE` but we need to check that it works or whether we need to register `CUDA` as a language as well in that case)

crtrott (changes requested):
Just read your comment: why is it not applicable with CUDA/HIP as language?

masterleinad (changes requested):
Looks like OpenMP support is already not working if `Cuda` is the language and we need something like
```CMake
index 34e9f0598..7011fd870 100644
--- a/cmake/kokkos_arch.cmake
+++ b/cmake/kokkos_arch.cmake
@@ -172,6 +172,9 @@ GLOBAL_SET(KOKKOS_CUDA_OPTIONS)
 IF(KOKKOS_CXX_COMPILER_ID STREQUAL NVIDIA)
   GLOBAL_APPEND(KOKKOS_CUDA_OPTIONS "-extended-lambda")
+  IF(Kokkos_ENABLE_OPENMP AND Kokkos_ENABLE_COMPILE_AS_CMAKE_LANGUAGE) 
+    GLOBAL_APPEND(KOKKOS_CUDA_OPTIONS -Xcompiler ${OpenMP_CXX_FLAGS})
+  ENDIF()
 ENDIF()
 
 IF (KOKKOS_ENABLE_CUDA_CONSTEXPR)
```
which should also work here.

dalg24:
> Looks like OpenMP support is already not working if `Cuda` is the language and we need something like
> [diff]
> which should also work here.

Please clarify.  My editing and striking out my own comment above was me expecting this PR would not break the use case openmp+cuda with compile as cmake language support enabled (regardless of whether it works or not at the moment in develop).
Are you reporting that as a matter of fact "Cuda+OpenMP with `Kokkos_ENABLE_COMPILE_AS_CMAKE_LANGUAGE` `ON`" does not work (w or w/o this PR)?

masterleinad:
> Are you reporting that as a matter of fact "Cuda+OpenMP with `Kokkos_ENABLE_COMPILE_AS_CMAKE_LANGUAGE` `ON`" does not work (w or w/o this PR)?

Yes, I'm seeing
```bash
cd /app/kokkos/build/core/src && /usr/local/cuda/bin/nvcc -forward-unknown-to-host-compiler -DKOKKOS_DEPENDENCE -I/app/kokkos/build/core/src -I/app/kokkos/core/src -I/app/kokkos/build -I/app/kokkos/core/src/../../tpls/desul/include -isystem=/usr/local/cuda/include -O2 -g -DNDEBUG --generate-code=arch=compute_80,code=[compute_80,sm_80] -extended-lambda -Wext-lambda-captures-this -std=c++17 -MD -MT core/src/CMakeFiles/kokkoscore.dir/impl/Kokkos_Core.cpp.o -MF CMakeFiles/kokkoscore.dir/impl/Kokkos_Core.cpp.o.d -x cu -c /app/kokkos/core/src/impl/Kokkos_Core.cpp -o CMakeFiles/kokkoscore.dir/impl/Kokkos_Core.cpp.o
In file included from /app/kokkos/core/src/OpenMP/Kokkos_OpenMP.hpp:203,
                 from /app/kokkos/core/src/decl/Kokkos_Declare_OPENMP.hpp:21,
                 from /app/kokkos/build/KokkosCore_Config_DeclareBackend.hpp:22,
                 from /app/kokkos/core/src/Kokkos_Core.hpp:45,
                 from /app/kokkos/core/src/impl/Kokkos_Core.cpp:21:
/app/kokkos/core/src/OpenMP/Kokkos_OpenMP_Instance.hpp:23:2: error: #error "You enabled Kokkos OpenMP support without enabling OpenMP in the compiler!"
   23 | #error \
      |  ^~~~~
make[2]: *** [core/src/CMakeFiles/kokkoscore.dir/build.make:118: core/src/CMakeFiles/kokkoscore.dir/impl/Kokkos_Core.cpp.o] Error 1
make[2]: Leaving directory '/app/kokkos/build'
make[1]: *** [CMakeFiles/Makefile2:1059: core/src/CMakeFiles/kokkoscore.dir/all] Error 2
make[1]: Leaving directory '/app/kokkos/build'
make: *** [Makefile:146: all] Error 2
```

dalg24:
That's what I thought, that the changes here do not have effect on CUDA (nor HIP) with compile as cmake language.

I am tempted to argue that this PR stands on it own and that I would not necessarily block it for loosely related pre-existing bugs.  That said if you and/or Christian would prefer to fix these in the process I am not going to argue too much.

Did you check whether it is also broken with HIP?

dalg24:
As far as I can tell HIP suffers from the same issue.
I am still not convinced this PR is the right place to fix these.

masterleinad:
> As far as I can tell HIP suffers from the same issue.

Yes, confirmed.

> I am still not convinced this PR is the right place to fix these.

I'm fine either way.

masterleinad (approved):
We can fix the issue with `Kokkos_ENABLE_COMPILE_AS_CMAKE_LANGUAGE` elsewhere.

dalg24:
> Just read your comment: why is it not applicable with CUDA/HIP as language?

Because the only effect of that change is that OpenMP support will be detected only for CXX, instead of potentially for C and Fortran if they are enabled at all.
If you look at the documentation of the cmake find module for openmp (link an above in the description) neither CUDA nor HIP matters.  And again this change is about **detection**, the logic that add special flags to enable OpenMP support stays the same.  Does that make sense?

crtrott:
Doesn't really make sense to me other than CMake being broken and not considering CUDA as a thing to add OpenMP too if it is enabled, even if you don't specify which languages to add it too ...

But sure if its anyway broken :-(

crtrott (approved):

PR #6969, opened 2024-04-29
Threads: Don't silently allow m_instance to be a nullptr

masterleinad:
https://github.com/kokkos/kokkos/actions/runs/8881763629/job/24384885382?pr=6955 is diagnosing that the reduction return value might not be initialized. While looking at that, I noticed that we are silently allowing `m_instance` to be a `nullptr` which doesn't seem to make much sense. Returning early in this case would not initialize the return variable indeed.

### DIFF
--- a/core/src/Threads/Kokkos_Threads_Team.hpp
+++ b/core/src/Threads/Kokkos_Threads_Team.hpp
@@ -352,6 +352,7 @@ class ThreadsExecTeamMember {
         m_chunk_size(team.chunk_size()),
         m_league_chunk_end(0),
         m_team_alloc(team.team_alloc()) {
+    KOKKOS_ASSERT(m_instance != nullptr);
     if (team.league_size()) {
       // Execution is using device-team interface:
 
--- a/core/src/Threads/Kokkos_Threads_Team.hpp
+++ b/core/src/Threads/Kokkos_Threads_Team.hpp
@@ -188,8 +188,6 @@ class ThreadsExecTeamMember {
         using type =
             typename if_c<sizeof(Type) < TEAM_REDUCE_SIZE, Type, void>::type;
 
-        if (m_instance == nullptr) return value;
-
         if (team_rank() != team_size() - 1) *
             ((volatile type*)m_instance->scratch_memory()) = value;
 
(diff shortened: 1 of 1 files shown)


Rombur (approved):

cedricchevalier19 (approved):
I was wondering about

```c++
  inline ThreadsInternal& threads_exec_team_base() const {
    return m_team_base ? **m_team_base : *m_instance;
  }
```

I am not sure if this function is used anywhere, but if it is, perhaps it should be hardened as well.

masterleinad:
> I am not sure if this function is used anywhere, but if it is, perhaps it should be hardened as well.

It's not part of the interface for team members and not used anywhere in the library so we should be able to just remove it.

core/src/Threads/Kokkos_Threads_Team.hpp
```diff
@@ -188,8 +188,6 @@ class ThreadsExecTeamMember {
         using type =
             typename if_c<sizeof(Type) < TEAM_REDUCE_SIZE, Type, void>::type;
 
-        if (m_instance == nullptr) return value;
```
nmm0, line 191: Should we Kokkos Assert here or something instead then? (and elsewhere)

masterleinad: The only place where `m_instance` is set, is in the constructor and we are using it directly after this line. Hence, I don't see much value in adding an assert here (but we could).

nmm0: ah got it, I'm fine with not checking if it is part of the class invariant that it is not null and not an individual precondition for one of the member functions

nmm0: Ah, but lets at least check it in the constructor then https://github.com/kokkos/kokkos/pull/6969/files#diff-3bf5843decd4b71c16ceb24b342b912271c46008d99ccc601d8f5b395d4fbb74L348

nmm0 (commented):
I think we should assert that instance is non-null in the constructor if we are not going to test in the functions: https://github.com/kokkos/kokkos/pull/6969/files#diff-3bf5843decd4b71c16ceb24b342b912271c46008d99ccc601d8f5b395d4fbb74L348

nmm0 (approved):
Looks good to me!

ndellingwood:
Does this merit a fixup changelog entry for 4.4 #6914 ?

masterleinad:
> Does this merit a fixup changelog entry for 4.4 #6914 ?

No.

PR #6982, opened 2024-05-02
Fix fedora CI builds with flang-new

masterleinad:
This fixes the fedora CI builds that started failing this week. Linking with `-fno-fortran-main`, prevents creating a main function from the `Fortran` code.
```bash
```


lucbv (approved):
Looks good to me.

crtrott (approved):

PR #6989, opened 2024-05-03
Fix deprecation warnings with GCC for `pair<T1,void>` comparison operators

dalg24:
Fixup for  #6947

GCC emits deprecation warnings in the definition of comparison operators for the deprecated specialization `Kokkos::pair<T1, void>`.  The fix is simple: add the `[[deprecated]]` attribute for all of these comparison operators.

The issue was spotted in the ArborX nightlies and is also present in our nighties on ORNL Jenkins, we just didn't notice.
@ndellingwood did the SNL nighty builds not catch this issue?!!

Reproducer [here](https://godbolt.org/z/4sfEjoWbG) does not seem to be sensitive to the GCC version.

### DIFF
--- a/core/src/Kokkos_Pair.hpp
+++ b/core/src/Kokkos_Pair.hpp
@@ -450,37 +450,37 @@ struct KOKKOS_DEPRECATED pair<T1, void> {
 //
 
 template <class T1>
-KOKKOS_FORCEINLINE_FUNCTION constexpr bool operator==(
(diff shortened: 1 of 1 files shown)


masterleinad (approved):

crtrott (approved):

ndellingwood:
@dalg24 most of the nightly jobs are building with deprecated code off (several Trilinos integration builds have deprecated code on but no -Werror due to noise in other packages). I'll update testing for more deprecated code == on testing coverage

PR #6990, opened 2024-05-06
Fix TPL_LIBRARY_SUFFIXES for 32-bit build

masterleinad:
For 32-bit builds, we should search for TPL libraries in `lib32` instead of `lib64`. This should fix the last failing GitHub CI build.

The `openmp.parallel_scan_with_reducers` test is failing with
```bash
 RUN      ] openmp.parallel_scan_with_reducers
/__w/kokkos/kokkos/core/unit_test/TestTeamVector.hpp:1010: Failure
  host_outputs(i)
    Which is: 1.0302e-12
  expected(i)
    Which is: 1.0302e-12
differ at index 103
/__w/kokkos/kokkos/core/unit_test/TestTeamVector.hpp:1010: Failure
  host_outputs(i)
    Which is: 1.0302e-12
  expected(i)
    Which is: 1.0302e-12
differ at index 102
```
and I decided to just skip since we have multiple other tests disabled for this build already (and it's not related to the build issue fixed here).


masterleinad:
The 32-bit build is passing.

crtrott (approved):

dalg24 (approved):

dalg24:
Ignoring unrelated runtime issues in one HIP build and one CUDA build that timed out

PR #6998, opened 2024-05-08
OpenMPTarget: Use mutex lock for parallel scan.

rgayatri23:
The PR adds a mutex lock so that only one instance of `parallel_scan` can access scratch memory at a given time.

### DIFF
--- a/core/src/OpenMPTarget/Kokkos_OpenMPTarget_ParallelScan_Range.hpp
+++ b/core/src/OpenMPTarget/Kokkos_OpenMPTarget_ParallelScan_Range.hpp
@@ -48,6 +48,10 @@ class ParallelScan<FunctorType, Kokkos::RangePolicy<Traits...>,
   value_type* m_result_ptr;
   const bool m_result_ptr_device_accessible;
 
+  // Only let one ParallelScan instance at a time use the scratch memory.
+  // The constructor acquires the mutex which is released in the destructor.
+  std::scoped_lock<std::mutex> m_scratch_memory_lock;
+
   template <class TagType>
   std::enable_if_t<std::is_void<TagType>::value> call_with_tag(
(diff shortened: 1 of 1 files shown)


masterleinad (approved):
Looks good to me. Let's see what CI says.

dalg24 (approved):
Please make sure you attach enough information in the description of your PR.
Clearly state that you are trying to fix a **bug**.
Describe the bug, here a race condition that has been observed on multiple occurrences recently in the CI builds.
You might want to paste a snippet from the failing logs
```
4: [ RUN      ] openmptarget.exec_space_thread_safety_range_scan
4: /var/jenkins/workspace/Kokkos_PR-6988/core/unit_test/TestExecSpaceThreadSafety.hpp:312: Failure
4:   host_error()
4:     Which is: 1
4:   0
4: [  FAILED  ] openmptarget.exec_space_thread_safety_range_scan (5907 ms)
```
You probably should also discuss where this bug is coming from.  Here the bug was always there and the recent merge of #6938 exposed it.  We just didn't realize then that it was there.

Then if you feel extra zealous you can explain that the scoped_lock data member technique is used in other places in your backend implementation
```
core/src/OpenMPTarget/Kokkos_OpenMPTarget_ParallelReduce_MDRange.hpp:78:        m_scratch_memory_lock(OpenMPTargetExec::m_mutex_scratch_ptr) {}
core/src/OpenMPTarget/Kokkos_OpenMPTarget_ParallelReduce_Range.hpp:112:        m_scratch_memory_lock(OpenMPTargetExec::m_mutex_scratch_ptr) {}
core/src/OpenMPTarget/Kokkos_OpenMPTarget_ParallelReduce_Team.hpp:525:        m_scratch_memory_lock(OpenMPTargetExec::m_mutex_scratch_ptr) {}
```

dalg24:
Please make sure you add this to the change log

PR #6999, opened 2024-05-08
Suppress deprecated warnings via pragma push/pop in the tests

dalg24:
Introduce `KOKKOS_IMPL_DISABLE_DEPRECATED_WARNINGS_{PUSH,POP}` macros to suppress diagnostics when appropriate and fix all the warnings I could see in our tests.

Just in case we need to rework the suppressions macros https://godbolt.org/z/Pjd7hYba9

### DIFF
--- a/core/src/Kokkos_Macros.hpp
+++ b/core/src/Kokkos_Macros.hpp
@@ -562,6 +562,28 @@ static constexpr bool kokkos_omp_on_host() { return false; }
 #define KOKKOS_IMPL_WARNING(desc) KOKKOS_IMPL_DO_PRAGMA(message(#desc))
 #endif
 
+// clang-format off
+#if defined(__EDG__)
+  #define KOKKOS_IMPL_DISABLE_DEPRECATED_WARNINGS_PUSH() \
+    _Pragma("push")                                      \
+    _Pragma("diag_suppress 1223")
+  #define KOKKOS_IMPL_DISABLE_DEPRECATED_WARNINGS_POP() \
+    _Pragma("pop")
+#elif defined(__GNUC__) || defined(__clang__)
+  #define KOKKOS_IMPL_DISABLE_DEPRECATED_WARNINGS_PUSH() \
+    _Pragma("GCC diagnostic push")                       \
+    _Pragma("GCC diagnostic ignored \"-Wdeprecated-declarations\"")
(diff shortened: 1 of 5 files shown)


core/src/Kokkos_Macros.hpp
```diff
@@ -562,6 +562,28 @@ static constexpr bool kokkos_omp_on_host() { return false; }
...
+    _Pragma("GCC diagnostic ignored \"-Wdeprecated-declarations\"")
+  #define KOKKOS_IMPL_DISABLE_DEPRECATED_WARNINGS_POP() \
+    _Pragma("GCC diagnostic pop")
+#elif defined(_MSC_VER)
+  #define KOKKOS_IMPL_DISABLE_DEPRECATED_WARNINGS_PUSH() \
+    _Pragma("warning(push)")                             \
+    _Pragma("warning(disable: 4996)")
+  #define KOKKOS_IMPL_DISABLE_DEPRECATED_WARNINGS_POP() \
+    _Pragma("warning(pop)")
+#endif
```
dalg24, line 584: Need to define them empty in the else-clause

masterleinad (approved):
Looks OK to me.

simd/unit_tests/include/SIMDTesting_Ops.hpp
```diff
@@ -81,7 +81,9 @@ class absolutes {
   auto on_host(T const& a) const {
     if constexpr (std::is_signed_v<typename T::value_type>) {
 #if defined(KOKKOS_ENABLE_DEPRECATED_CODE_4)
+      KOKKOS_IMPL_DISABLE_DEPRECATED_WARNINGS_PUSH()
       return Kokkos::Experimental::abs(a);
+      KOKKOS_IMPL_DISABLE_DEPRECATED_WARNINGS_POP()
```
Rombur, line 86: Can the compiler decide to ignore the pragma because it's after the return and the code cannot be "executed"?

nmm0 (approved):
Looks good to me!

simd/unit_tests/include/SIMDTesting_Ops.hpp
nmm0: I don't think it would be pruned since it's all preprocessor based and not actually executed

PR #7003, opened 2024-05-09
SYCL: Add Kokkos_ENABLE_SYCL_NO_IMMEDIATE_COMMAND_LISTS option

masterleinad:
Related to #6912. Prior to https://github.com/intel/llvm/pull/12279 and https://github.com/oneapi-src/unified-runtime/pull/1218, the oneAPI implementation didn't support immediate command lists. Thus, this pull request provides a way to disable their use with a `CMake` option. More information on immediate command lists (which is the default) can be found at https://www.intel.com/content/www/us/en/developer/articles/guide/level-zero-immediate-command-lists.html. In short, the option controls whether kernels are executed in batched before being submitted for execution or not.
Depending on the kernel's complexity/weight this might be beneficial or not.

The alternative is for the user to define `SYCL_PI_LEVEL_ZERO_USE_IMMEDIATE_COMMANDLISTS` as an environment variable but that seems to be worse in terms of figuring out the configuration a user is using.

We could make this flag `EXPERIMENTAL` or `IMPL` since it's not clear to me that we need this apart from Graph support since Kokkos' kernels are normally not very lightweight.


core/src/SYCL/Kokkos_SYCL.hpp
```diff
@@ -159,8 +159,14 @@ std::vector<SYCL> partition_space(const SYCL& sycl_space, Args...) {
   std::vector<SYCL> instances;
   instances.reserve(sizeof...(Args));
   for (unsigned int i = 0; i < sizeof...(Args); ++i)
-    instances.emplace_back(
-        sycl::queue(context, device, sycl::property::queue::in_order()));
+    instances.emplace_back(sycl::queue(
+        context, device,
+        {sycl::property::queue::in_order()
```
dalg24, line 164: Was it intentional that the use of "in order" queue was not guarded here?

dalg24 (commented):
Do we have any experience with the performance implications of forcing the no_immediate_command_list?

What are your thoughts on whether to commit to make it a public configuration option?

masterleinad:
> Do we have any experience with the performance implications of forcing the no_immediate_command_list?

Not really. It's just about launch overhead.

 > What are your thoughts on whether to commit to make it a public configuration option?

I would only expose it to provide a way for `Graphs` to work.

core/src/SYCL/Kokkos_SYCL.hpp
masterleinad: Related pull requests are https://github.com/kokkos/kokkos/pull/6189 and https://github.com/kokkos/kokkos/pull/6246. The latter former made the default execution space instances in order and the latter dealt with interoperability (without the guard). Thus, it's at least consistent with the interoperability test but I could see that would also be consistent to guard it here.

dalg24: I don't understand your answer.  If we have a `KOKKOS_IMPL_SYCL_USE_IN_ORDER_QUEUES` macro and we use it internally to decide whether to pass the `sycl::property::queue::in_order` for the default instance, why are we unconditionally specifying `in_order` in `partition_space`?

cmake/kokkos_enable_options.cmake
```diff
@@ -48,6 +48,7 @@ KOKKOS_ENABLE_OPTION(CUDA_LAMBDA ${CUDA_LAMBDA_DEFAULT} "Whether to allow lambda
 # resolved but we keep the option around a bit longer to be safe.
 KOKKOS_ENABLE_OPTION(IMPL_CUDA_MALLOC_ASYNC ON  "Whether to enable CudaMallocAsync (requires CUDA Toolkit 11.2)")
 KOKKOS_ENABLE_OPTION(IMPL_NVHPC_AS_DEVICE_COMPILER OFF "Whether to allow nvc++ as Cuda device compiler")
+KOKKOS_ENABLE_OPTION(SYCL_NO_IMMEDIATE_COMMAND_LISTS OFF "Whether to disallow immediate command lists in the SYCL backend (necessary for Graph support)")
```
dalg24, line 51: Considering that this does not seem to be a long term solution, please make it an IMPL_ option

core/src/SYCL/Kokkos_SYCL.cpp
```diff
@@ -110,6 +110,11 @@ void SYCL::print_configuration(std::ostream& os, bool verbose) const {
 #else
   os << "macro  KOKKOS_IMPL_SYCL_USE_IN_ORDER_QUEUES : undefined\n";
 #endif
+#ifdef KOKKOS_ENABLE_SYCL_NO_IMMEDIATE_COMMAND_LISTS
+  os << "macro  KOKKOS_ENABLE_SYCL_NO_IMMEDIATE_COMMAND_LISTS : defined\n";
+#else
+  os << "macro  KOKKOS_ENABLE_SYCL_NO_IMMEDIATE_COMMAND_LISTS : undefined\n";
+#endif
```
dalg24, line 117: You might want to print the runtime property of the queue as well.

masterleinad:
Superseded by #7004.

PR #7006, opened 2024-05-10
Test DualView resize/realloc with WithoutInitializing for types without default constructor

masterleinad:
This pull request adds another test for `DualView` for https://github.com/kokkos/kokkos/pull/6993 that fixed `Kokkos::realloc/resize` with Kokkos::WithouttInitializing such that it can be used with types that don't define a default constructor.

### DIFF
--- a/containers/unit_tests/TestDualView.hpp
+++ b/containers/unit_tests/TestDualView.hpp
@@ -282,15 +282,20 @@ struct test_dualview_resize {
     const unsigned int m      = 5;
     const unsigned int factor = 2;
 
-    ViewType a("A", n, m);
+    ViewType a;
+    if constexpr (Initialize)
+      a = ViewType("A", n, m);
+    else
+      a = ViewType(Kokkos::view_alloc(Kokkos::WithoutInitializing, "A"), n, m);
+
     Kokkos::deep_copy(a.d_view, 1);
 
     /* Covers case "Resize on Device" */
     a.modify_device();
-    if (Initialize)
-      Kokkos::resize(Kokkos::WithoutInitializing, a, factor * n, factor * m);
-    else
+    if constexpr (Initialize)
       Kokkos::resize(a, factor * n, factor * m);
+    else
+      Kokkos::resize(Kokkos::WithoutInitializing, a, factor * n, factor * m);
     ASSERT_EQ(a.extent(0), n * factor);
     ASSERT_EQ(a.extent(1), m * factor);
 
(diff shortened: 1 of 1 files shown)


containers/unit_tests/TestDualView.hpp
```diff
@@ -463,12 +472,36 @@ TEST(TEST_CATEGORY, dualview_deep_copy) {
...
+struct Kokkos::reduction_identity<Test::NoDefaultConstructor> {
+  static Test::NoDefaultConstructor sum() { return {0}; }
+};
+
+namespace Test {
+
 TEST(TEST_CATEGORY, dualview_realloc) {
   test_dualview_realloc<int, TEST_EXECSPACE>();
+  Impl::test_dualview_realloc<NoDefaultConstructor, TEST_EXECSPACE,
+                              /* Initialize */ false>();
```
cedricchevalier19, line 498: A little bit disturbing to have `Impl::` on this line and not on the one above.

containers/unit_tests/TestDualView.hpp
```diff
@@ -463,12 +472,36 @@ TEST(TEST_CATEGORY, dualview_deep_copy) {
...
 }
 
 TEST(TEST_CATEGORY, dualview_resize) {
   test_dualview_resize<int, TEST_EXECSPACE>();
+  Impl::test_dualview_resize<NoDefaultConstructor, TEST_EXECSPACE,
+                             /* Initialize */ false>();
```
cedricchevalier19, line 504: A little bit disturbing to have `Impl::` on this line and not on the one above.

cedricchevalier19 (approved):

containers/unit_tests/TestDualView.hpp
masterleinad: `test_dualview_realloc` calls `Impl::test_dualview_realloc` both with `Initialiize=true` and `Initialize=false`. We can only test with `Initialize=false` for `NoDefualtConstructor`, though.

crtrott (approved):

crtrott (changes requested):
This doesn't work yet the failures in Jenkins are real.

dalg24:
Retest this please

crtrott (approved):

crtrott (changes requested):
There is a SYCL segfault is that real?

masterleinad:
> There is a SYCL segfault is that real?

No,
> [ RUN      ] sycl_graph.launch_six
 3/57 Test  #3: Kokkos_CoreUnitTest_SYCL1A .......................***Exception: SegFault 30.80 sec

was fixed in a different pull request and this one only modifies container tests anyway.

crtrott (approved):

PR #7009, opened 2024-05-13
SYCL: Don't use shuffles for top-level reductions

masterleinad:
Partly reverts #6750. It turns out that using `shuffles` using more generic types such as in
```C++
Kokkos::parallel_reduce("kk_wup", N, KOKKOS_LAMBDA(int i, double& lsum1, double& lsum2) {
                      lsum1 += x_data_[i]*y_data_[i];
                      lsum2 += x_data_[i]*y_data_[i];},
                      result1, result2);
```
is much slower. Given that #6750 brought shuffle reductions only on par with local memory reductions, it makes sense to always use the latter for now. I have a slight preference for keeping the code for shuffle reductions (like we do for Cuda and HIP).

### DIFF
--- a/core/src/SYCL/Kokkos_SYCL_WorkgroupReduction.hpp
+++ b/core/src/SYCL/Kokkos_SYCL_WorkgroupReduction.hpp
@@ -21,9 +21,11 @@
 
 namespace Kokkos::Impl::SYCLReduction {
 
+// FIXME_SYCL It appears that using shuffles is slower than going through local
+// memory.
 template <class ReducerType>
-inline constexpr bool use_shuffle_based_algorithm =
-    std::is_reference_v<typename ReducerType::reference_type>;
+inline constexpr bool use_shuffle_based_algorithm = false;
+// std::is_reference_v<typename ReducerType::reference_type>;
 
 template <typename ValueType, typename ReducerType, int dim>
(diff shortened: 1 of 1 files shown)


core/src/SYCL/Kokkos_SYCL_WorkgroupReduction.hpp
```diff
@@ -21,9 +21,11 @@
 
 namespace Kokkos::Impl::SYCLReduction {
 
+// FIXME_SYCL It appears that using shuffles is slower than going through local
+// memory.
```
dalg24, line 25: Do you want to say more here?  Because if I understand correctly you are reported that it is slower for reductions on user-defined types (in your report that type is produced internally to handle the doing multiple reductions at once).  Because one might run an experiment on scalar types just like you did in your original work and conclude this is not needed any more.

masterleinad: Sure, see c6dedc1c30a4af3ecfd56ef42bfe8286b2a8edd1.

masterleinad:
The SYCL build passed.

dalg24 (approved):

crtrott (approved):

ndellingwood:
@masterleinad does this need a changelog entry for 4.4? Looks like it may have impacts on performance?

PR #7010, opened 2024-05-14
HPCBIND: Prevent hwloc-ls from opening graphic windows unintentionally

DanielRichtmann:
I noticed that `hpcbind` unconditionally opens two lstopo X windows when a display is present. The command I ran to trigger this behavior was

`hpcbind --distribute=2 --distribute-partition=1 -v`

in combination with hwloc version 2.7.0 on ubuntu 22.04 lts.

This PR changes this behavior to only open these windows when the `--lstopo` flag is passed at the command line.

### DIFF
--- a/bin/hpcbind
+++ b/bin/hpcbind
@@ -635,7 +635,9 @@ if [[ ${HPCBIND_TEE} -eq 0 || ${HPCBIND_VERBOSE} -eq 0 ]]; then
 
   if [[ ${HPCBIND_HAS_HWLOC} -eq 1 ]]; then
     echo "[BINDINGS]" >> ${HPCBIND_LOG}
-    hwloc-ls --restrict "${HPCBIND_HWLOC_CPUSET}" >> ${HPCBIND_LOG}
+    if [[ ${HPCBIND_LSTOPO} -eq 1 && ! -z ${DISPLAY} ]]; then
+      hwloc-ls --restrict "${HPCBIND_HWLOC_CPUSET}" >> ${HPCBIND_LOG}
+    fi
   else
     echo "Unable to show bindings, hwloc not available." >> ${HPCBIND_LOG}
   fi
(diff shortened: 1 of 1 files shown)


bin/hpcbind
```diff
@@ -635,7 +635,9 @@ if [[ ${HPCBIND_TEE} -eq 0 || ${HPCBIND_VERBOSE} -eq 0 ]]; then
 
   if [[ ${HPCBIND_HAS_HWLOC} -eq 1 ]]; then
     echo "[BINDINGS]" >> ${HPCBIND_LOG}
-    hwloc-ls --restrict "${HPCBIND_HWLOC_CPUSET}" >> ${HPCBIND_LOG}
+    if [[ ${HPCBIND_LSTOPO} -eq 1 && ! -z ${DISPLAY} ]]; then
+      hwloc-ls --restrict "${HPCBIND_HWLOC_CPUSET}" >> ${HPCBIND_LOG}
```
dalg24, line 639: Would passing `--output-format console` resolve your issue?
(I have no experience with `hpcbind` nor `hwloc`)

ajpowelsnl, line 639: Isn't `$DISPLAY` set / unset on the user's machine?  Assuming that's the case, then conditioning `hwloc-ls --restrict  "${HPCBIND_HWLOC_CPUSET}" >> ${HPCBIND_LOG}` upon `${DISPLAY}` being non zero (if it's set at all) could have negative downstream effects for users who wish to use `hpcbind` , but do not have `$DISPLAY` set.  I checked a couple of (non-Windows) machines, and `$DISPLAY` wasn't set, appearing to be the default scenario.  Would it be possible for you to simply `unset $DISPLAY` before using `hpcbind`?

bin/hpcbind
```diff
@@ -660,7 +662,9 @@ else
 
   if [[ ${HPCBIND_HAS_HWLOC} -eq 1 ]]; then
     echo "[BINDINGS]" > >(tee -a ${HPCBIND_LOG})
-    hwloc-ls --restrict "${HPCBIND_HWLOC_CPUSET}" --no-io --no-bridges > >(tee -a ${HPCBIND_LOG})
+    if [[ ${HPCBIND_LSTOPO} -eq 1 && ! -z ${DISPLAY} ]]; then
+      hwloc-ls --restrict "${HPCBIND_HWLOC_CPUSET}" --no-io --no-bridges > >(tee -a ${HPCBIND_LOG})
+    fi
```
ajpowelsnl, line 667: See comment above above the logic and potential risk of the proposed change.

ajpowelsnl (commented):
Maybe  a fix that should occur on the user's machine?

ajpowelsnl:
Closing because merging the proposed change creates risk for non-Windows users

masterleinad:
> Closing because merging the proposed change creates risk for non-Windows users

This pull request is not about `WIndows`, but about `hpcbind` opening windows/GUIs on `Ubuntu`.

PR #7017, opened 2024-05-21
Improve `Impl::is_zero_byte()`

dalg24:
Follow up on #7014
Simplify the logic, do not copy the value and use `memcmp`,  trust that the standard library implementers know what they are doing.
I assumed that no one will stuff elements of size exceeding 256 bytes in views.  It just means that the `ZeroMemset` code path would not be exerted if that happens, code correctness would not be affected.

### DIFF
--- a/core/src/impl/Kokkos_ViewMapping.hpp
+++ b/core/src/impl/Kokkos_ViewMapping.hpp
@@ -17,11 +17,10 @@
 #ifndef KOKKOS_EXPERIMENTAL_VIEW_MAPPING_HPP
 #define KOKKOS_EXPERIMENTAL_VIEW_MAPPING_HPP
 
+#include <cstring>
 #include <type_traits>
 #include <initializer_list>
 
-#include <Kokkos_Array.hpp>
-#include <Kokkos_BitManipulation.hpp>
 #include <Kokkos_Core_fwd.hpp>
 #include <Kokkos_DetectionIdiom.hpp>
 #include <Kokkos_Pair.hpp>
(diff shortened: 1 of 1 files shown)


core/src/impl/Kokkos_ViewMapping.hpp
```diff
@@ -2530,21 +2529,12 @@ struct ViewDataHandle<
 namespace Kokkos {
 namespace Impl {
 
+inline static constexpr std::byte all_zeroes[256] = {};
```
masterleinad, line 2532:
```suggestion
```

core/src/impl/Kokkos_ViewMapping.hpp
```diff
@@ -2530,21 +2529,12 @@ struct ViewDataHandle<
...
-              sizeof(T) % sizeof(int) == 0, int,
-              std::conditional_t<sizeof(T) % sizeof(short int) == 0, short int,
-                                 char>>>>;
-  auto bit_values = Kokkos::bit_cast<
-      Kokkos::Array<comparison_type, sizeof(T) / sizeof(comparison_type)>>(t);
-  for (std::size_t i = 0; i < sizeof(T) / sizeof(comparison_type); ++i)
-    if (bit_values[i] != 0) return false;
-  return true;
+bool is_zero_byte(const T& x) {
+  if (sizeof(T) > sizeof(all_zeroes)) return false;  // out of luck
```
masterleinad, line 2536: I'd prefer
```suggestion
  constexpr std::byte all_zeroes[sizeof(T)] = {};
```
to be more general. The generated instructions for all versions are pretty similar, see https://godbolt.org/z/6Ef97eKhv.

masterleinad (approved):

crtrott (approved):

dalg24:
Boldly ignoring all the timeouts and other failures

PR #7020, opened 2024-05-21
OpenMPTarget: Fix compiling Graph tests

masterleinad:
Moving the locks into `execute` fixes the issue and makes the relevant classes copyable. We only mess with scratch memory in `execute` and not in the constructor anyway. For consistency, I changed all classes even those that are not yet tested with Graphs.

### DIFF
--- a/core/unit_test/TestGraph.hpp
+++ b/core/unit_test/TestGraph.hpp
@@ -130,7 +130,12 @@ TEST_F(TEST_CATEGORY_FIXTURE(graph), launch_six) {
         Kokkos::MDRangePolicy<TEST_EXECSPACE, Kokkos::Rank<2>>{{0, 0}, {1, 1}},
         count_functor{count, bugs, 0, 6});
     //----------------------------------------
-    ready.then_parallel_for(Kokkos::TeamPolicy<TEST_EXECSPACE>{1, 1},
+#ifdef KOKKOS_ENABLE_OPENMPTARGET
+    int team_size = 32;
+#else
+    int team_size = 1;
+#endif
+    ready.then_parallel_for(Kokkos::TeamPolicy<TEST_EXECSPACE>{1, team_size},
                             count_functor{count, bugs, 0, 6});
     //----------------------------------------
     ready.then_parallel_for(2, count_functor{count, bugs, 0, 6});
--- a/core/unit_test/TestGraph.hpp
+++ b/core/unit_test/TestGraph.hpp
@@ -113,6 +113,10 @@ TEST_F(TEST_CATEGORY_FIXTURE(graph), launch_one_rvalue) {
 }
 
 TEST_F(TEST_CATEGORY_FIXTURE(graph), launch_six) {
+#ifdef KOKKOS_ENABLE_OPENMPTARGET  // FIXME_OPENMPTARGET team_size incompatible
+  if (std::is_same_v<TEST_EXECSPACE, Kokkos::Experimental::OpenMPTarget>)
+    GTEST_SKIP() << "skipping since OpenMPTarget can't use team_size 1";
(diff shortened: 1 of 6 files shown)


crtrott (approved):

dalg24:
Did compile but failed with
```
3: [ RUN      ] openmptarget_graph.launch_six
3: OpenMPTarget backend requires a minimum of 32 threads per team.
3: .
 3/48 Test  #3: Kokkos_CoreUnitTest_OpenMPTarget ............Subprocess aborted***Exception: 259.66 sec

File: core/unit_test/TestGraph.hpp [hunk 1]
```diff
@@ -130,7 +130,12 @@ TEST_F(TEST_CATEGORY_FIXTURE(graph), launch_six) {
         Kokkos::MDRangePolicy<TEST_EXECSPACE, Kokkos::Rank<2>>{{0, 0}, {1, 1}},
         count_functor{count, bugs, 0, 6});
     //----------------------------------------
-    ready.then_parallel_for(Kokkos::TeamPolicy<TEST_EXECSPACE>{1, 1},
+#ifdef KOKKOS_ENABLE_OPENMPTARGET
```
dalg24, line 133:
```suggestion
#ifdef KOKKOS_ENABLE_OPENMPTARGET  // FIXME_OPENMPTARGET
```

rgayatri23 (approved):
LGTM, If the unit tests pass. The 32 thread limit might not be needed with newer upcoming llvm compiler versions.

masterleinad:
> LGTM, If the unit tests pass. The 32 thread limit might not be needed with newer upcoming llvm compiler versions.

Well, we are enforcing it. It doesn't come from the compiler so feel free to remove that check when we can.

rgayatri23:
> > LGTM, If the unit tests pass. The 32 thread limit might not be needed with newer upcoming llvm compiler versions.
> 
> Well, we are enforcing it. It doesn't come from the compiler so feel free to remove that check when we can.

We are enforcing it because the llvm compiler generates a minimum of 32 threads. Even if we request less than that (lets say 16 threads), the compiler will silently generate 32 and not give a warning. This might impact applications that depend on exact number of threads in a `TeamThreadRange`.

masterleinad:
The OpenMPTarget build passed.

core/unit_test/TestGraph.hpp
```diff
@@ -113,6 +113,10 @@ TEST_F(TEST_CATEGORY_FIXTURE(graph), launch_one_rvalue) {
 }
 
 TEST_F(TEST_CATEGORY_FIXTURE(graph), launch_six) {
+#ifdef KOKKOS_ENABLE_OPENMPTARGET  // FIXME_OPENMPTARGET team_size incompatible
+  if (std::is_same_v<TEST_EXECSPACE, Kokkos::Experimental::OpenMPTarget>)
+    GTEST_SKIP() << "skipping since OpenMPTarget can't use team_size 1";
```
dalg24, line 118: And that team size cannot be easily set to an acceptable value?  Why did you revert your initial attention to set it to 32?  What was the issue?

dalg24:
> The OpenMPTarget build passed.

Would you want this PR merged before #7024 ?

core/unit_test/TestGraph.hpp
masterleinad: The test was failing with a larger team size (and I didn't investigate more after that).

dalg24: Fine but the comment does not really capture that you tried the usual "if using openmptarget then set a team size of 32" and that it did not resolve the issue.

dalg24 (approved):

PR #7012, opened 2024-05-15
Change the default execution behavior of the parallel_for(team-policy) constructs in the OpenACC backend

seyonglee:
This PR changes the default execution behavior of the parallel_for(team-policy) constructs in the OpenACC backend.
(This PR handles a missing case not covered by the previous PR #6772)
The update in this PR fixes the OpenACC backend error in the thread safety test in PR #6938.

PR #7022, opened 2024-05-21
OpenACC: Skip exec_space_thread_safety_range_scan

masterleinad:
Even after https://github.com/kokkos/kokkos/pull/7012 `exec_space_thread_safety_range_scan` is still failing for `OpenACC`. This pull request disables the test for now to get the CI running again.

### DIFF
--- a/core/unit_test/TestExecSpaceThreadSafety.hpp
+++ b/core/unit_test/TestExecSpaceThreadSafety.hpp
@@ -77,6 +77,10 @@ void run_exec_space_thread_safety_range() {
 }
 
 TEST(TEST_CATEGORY, exec_space_thread_safety_range) {
+#ifdef KOKKOS_ENABLE_OPENACC  // FIXME_OPENACC
+  if (std::is_same_v<TEST_EXECSPACE, Kokkos::Experimental::OpenACC>)
+    GTEST_SKIP() << "skipping since test is known to fail with OpenACC";
+#endif
 #ifdef KOKKOS_ENABLE_OPENMPTARGET
   if (std::is_same_v<TEST_EXECSPACE, Kokkos::Experimental::OpenMPTarget>)
     GTEST_SKIP() << "skipping since test is known to fail for OpenMPTarget";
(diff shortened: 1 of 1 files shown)


seyonglee:
The current OpenACC implementation of the `parallel_scan()` construct is NOT thread-safe.
The OpenACC parallel_scan() is implemented using 5 subtasks, each of which is thread-safe, but to make the `parallel_scan()` thread-safe, the 5 subtasks should be executed in an atomic manner (like OpenMP critical section), but the current OpenACC implementation does not guarantee that.

seyonglee (approved):

core/unit_test/TestExecSpaceThreadSafety.hpp
```diff
@@ -77,6 +77,10 @@ void run_exec_space_thread_safety_range() {
 }
 
 TEST(TEST_CATEGORY, exec_space_thread_safety_range) {
+#ifdef KOKKOS_ENABLE_OPENACC  // FIXME_OPENACC
```
seyonglee, line 80: Why disabled here? As I know, exec_space_thread_safety_range will work for the OpenACC backend.

masterleinad:
```
[ RUN      ] openacc.exec_space_thread_safety_range
3: Kokkos::Impl::SharedAllocationRecord '' failed decrement count = 0
```
If you could investigate this further that would be great. I just want the CI to pass again without investigating the cause of the failures too much.

masterleinad:
The OpenACC build passed.

core/unit_test/TestExecSpaceThreadSafety.hpp
masterleinad: The test checks if kernels submitted to the same execution space instance from multiple threads are enqueued and don't run concurrently.

nmm0 (approved):
Looks good to me, with @seyonglee 's go ahead of course

PR #7028, opened 2024-05-23
Add tests to `ScopeGuard`

pzehner:
This PR aims to test the `Kokkos::ScopeGuard` class, that was untested.

No new features nor modifications were added to the core source code. A new test file `core/unit_test/UnitTest_ScopeGuard.cpp` was created and tests the scope guard on various configurations. Especially, faulty situations were tested, e.g. when creating two scope guards.

### DIFF
--- a/core/unit_test/UnitTest_ScopeGuard.cpp
+++ b/core/unit_test/UnitTest_ScopeGuard.cpp
@@ -147,4 +147,8 @@ TEST_F(scope_guard_DeathTest, destroy_after_finalize) {
 static_assert(!std::is_copy_assignable<Kokkos::ScopeGuard>());
 static_assert(!std::is_copy_constructible<Kokkos::ScopeGuard>());
 
+// Test scope guard is not movable.
+static_assert(!std::is_move_assignable<Kokkos::ScopeGuard>());
+static_assert(!std::is_move_constructible<Kokkos::ScopeGuard>());
+
 }  // namespace
--- a/core/unit_test/UnitTest_ScopeGuard.cpp
+++ b/core/unit_test/UnitTest_ScopeGuard.cpp
@@ -1,0 +1,95 @@
+#include <cstdlib>
+#include <gtest/gtest.h>
+#include <Kokkos_Core.hpp>
+
+namespace Test {
+
+/**
+ * Test to create a scope guard normally.
+ */
+TEST(scope_guard, create) {
+  ASSERT_FALSE(Kokkos::is_initialized());
+  ASSERT_FALSE(Kokkos::is_finalized());
+
+  // run it in a different process so side effects are not kept
+  EXPECT_EXIT(
+      {
+        {
+          Kokkos::ScopeGuard guard{};
+
+          ASSERT_TRUE(Kokkos::is_initialized());
+          ASSERT_FALSE(Kokkos::is_finalized());
+        }
+
+        ASSERT_FALSE(Kokkos::is_initialized());
+        ASSERT_TRUE(Kokkos::is_finalized());
+
+        std::exit(EXIT_SUCCESS);
+      },
+      testing::ExitedWithCode(0), "");
+
+  ASSERT_FALSE(Kokkos::is_initialized());
+  ASSERT_FALSE(Kokkos::is_finalized());
+}
+
+/**
+ * Test to create another scope guard when one has been created.
+ */
+TEST(scope_guard, create_while_initialize) {
+  ASSERT_FALSE(Kokkos::is_initialized());
+  ASSERT_FALSE(Kokkos::is_finalized());
+
+  EXPECT_DEATH(
+      {
+        Kokkos::ScopeGuard guard1{};
+
+        // create a second scope guard while there is one already existing
+        Kokkos::ScopeGuard guard2{};
+      },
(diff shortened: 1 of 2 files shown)


core/unit_test/UnitTest_ScopeGuard.cpp
```diff
@@ -0,0 +1,123 @@
...
+    ASSERT_FALSE(Kokkos::is_finalized());
+  }
+
+  void TearDown() override {
+    ASSERT_FALSE(Kokkos::is_initialized());
+    ASSERT_FALSE(Kokkos::is_finalized());
+  }
+};
+
+namespace Test {
```
dalg24, line 38: Please make it the anonymous namespace 
```suggestion
namespace {
```
and move the definition of `AssertEnvironmentTest` into it

dalg24 (commented):
You want to static assert that the scope guard is not copyable?
```c++
static_assert(!std::is_copy_constructible<Kokkos::ScopeGuard>
              && !std::is_copy_assignable<Kokkos::ScopeGuard>);
```
(Add at namespace scope)

I think our convention is to suffix death tests with `_DeathTest`.
I am not sure how to do it with fixtures though.

pzehner:
> Look a bit what we did for other death tests. I think we generally remove them from the list with the OpenACC backend.

But the OpenACC backend works fine on those death tests! This is weird, I should check what was done for the other death tests.

pzehner:
Retest this please

core/unit_test/UnitTest_ScopeGuard.cpp
```diff
@@ -0,0 +1,123 @@
...
+          if (!Kokkos::is_initialized()) std::exit(EXIT_FAILURE);
+          if (Kokkos::is_finalized()) std::exit(EXIT_FAILURE);
+        }
+
+        if (Kokkos::is_initialized()) std::exit(EXIT_FAILURE);
+        if (!Kokkos::is_finalized()) std::exit(EXIT_FAILURE);
+
+        std::exit(EXIT_SUCCESS);
+      },
+      testing::ExitedWithCode(0), "");
```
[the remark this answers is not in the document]
dalg24: You're right.  It is indeed not provided by google test.  Reading the doc again and it looks like they imply `NormalExit()` would be defined in the user code.
Sorry about the noise.

core/unit_test/UnitTest_ScopeGuard.cpp
```diff
@@ -0,0 +1,150 @@
...
+          Kokkos::ScopeGuard guard{};
+
+          if (!Kokkos::is_initialized()) std::exit(EXIT_FAILURE);
+          if (Kokkos::is_finalized()) std::exit(EXIT_FAILURE);
+        }
+
+        if (Kokkos::is_initialized()) std::exit(EXIT_FAILURE);
+        if (!Kokkos::is_finalized()) std::exit(EXIT_FAILURE);
+
+        std::exit(EXIT_SUCCESS);
```
dalg24, line 60: I would have slightly preferred if we used consistently either `EXIT_SUCCESS` or `0` as the exit code as argument of `std::exit()` and `ExitedWithCode()`

dalg24 (approved):

core/unit_test/UnitTest_ScopeGuard.cpp
pzehner: Sure, I'll harmonize that.

masterleinad (approved):

core/unit_test/UnitTest_ScopeGuard.cpp
```diff
@@ -0,0 +1,150 @@
...
+      "Destroying a ScopeGuard after Kokkos was finalized");
+}
+
+/**
+ * Static tests
+ */
+
+// Test scope guard is not copyable.
+static_assert(!std::is_copy_assignable<Kokkos::ScopeGuard>());
+static_assert(!std::is_copy_constructible<Kokkos::ScopeGuard>());
```
nliber, line 148: Please also test that it is not moveable.

nliber: I'm wondering if those `static_asserts` should be in the code, not the test (theoretically, one could have a reference counted `ScopeGuard` with reasonable copy semantics).

nliber (dismissed):

core/unit_test/UnitTest_ScopeGuard.cpp
pzehner: I'll add the assert to this aim, but I'm not sure to be able to tell where it should be.

pzehner:
Ok, I'll add some guards to prevent those tests to run with NVHPC.

Yet, I'm surprised that NVHPC with the OpenACC backend works.

core/unit_test/UnitTest_ScopeGuard.cpp
dalg24: Having them in the tests where they currently are is fine.

pzehner: Ok.

pzehner:
It seems to work now for NVHPC. The only failing job is for HIP, which segfaults for a lot of tests.

dalg24 (approved):

PR #7030, opened 2024-05-23
CTAD TeamPolicy (v3)

nliber:
Starting with the deduction guides & test cases from ctad-teampolicy-v3
This compiles under llvm12, llvm18, gcc11 and gcc14.


dalg24 (approved):
Passing `Kokkos::AUTO_t` by const reference hurts my eyes but we do the same thing in the actual constructors and we can fix it all later.

dalg24:
@nliber please add this to the changelog and update the doc

PR #7032, opened 2024-05-24
Memory Pool size setting for cudaMallocAsync

bjoo:
# A patch to allow setting the pool size for cudaMallocAsync 

## Implementation
The first time cudaAsyncMalloc is called (in void* impl_alloc_common() in Kokkos_CudaSpace.cpp)
we check the environment variable KOKKOS_CUDA_MEMPOOL_SIZE . We overallocate this by 64 bytes (an arbitrary amount, just to get us over the size), then set some properties 
on the device default mempool to retain KOKKOS_CUDA_MEMPOOL_SIZE of memory after an async free. Subsequent allocations are 
faster (by between 1 to 2 orders of magnitude) for sufficiently large sized chunks of memory which fit in the Pool.

## Efficacy
A benchmark test has been placed in kokkos/benchmarks/async_test which can be built using the 'Makefile' setup. 
To execute it,  export the KOKKOS_CUDA_MEMPOOL_SIZE and run `async_alloc.cuda`. The utility will range through allocations from 8B to 16GB.
And collect timing of allocating (and freeing) a Kokkos::View.

The `-d` flag to `async_alloc.cuda` can be used to specify cycling downwards i.e. from 16GB to 8B.

The attached PDF shows the benchmark times, sweeping up from 0 to 8GB sizes, with various mempool settings show the gains from the async allocator from allocation sizes of 512KB upwards 
* at about 16MiB the allocator with an unspecified (0) poolsize becomes as expensive as using cudaMalloc and in fact becomes worse
* using a pool maintains an advantage of between 1 to 2 orders of magnitude depending on the allocation size
* after 4GB the allocation efficiency with the 4.2GB pool starts to deteriorate as we run out of pool space.

This data is from an Ada L40S GPU. Other GPU architecture benchmarks are work in progress just now.
[AsyncAllocUp.pdf](https://github.com/kokkos/kokkos/files/15436278/AsyncAllocUp.pdf)


core/src/Cuda/Kokkos_CudaSpace.cpp
```diff
@@ -172,6 +174,115 @@ void *CudaSpace::allocate(const char *arg_label, const size_t arg_alloc_size,
...
+  if (std::isalpha(last_letter)) {
+    mempool_size_string.pop_back();
+    char last_letter_cap = std::toupper(last_letter);
+    switch (last_letter_cap) {
+      case 'K': factor = 1024; break;
+      case 'M': factor = 1024 * 1024; break;
+      case 'G': factor = 1024 * 1024 * 1024; break;
+      default:
+        // Not permitted
+        return false;
```
masterleinad, line 224: This should probably print the problem.
```suggestion
        std::cerr << "KOKKOS_CUDA_MEMPOOL_SIZE couldn't be parsed properly!\n"
        // Not permitted
        return false;
```

core/src/Cuda/Kokkos_CudaSpace.cpp
```diff
@@ -172,6 +174,115 @@ void *CudaSpace::allocate(const char *arg_label, const size_t arg_alloc_size,
...
 
 namespace {
+
+#ifndef CUDART_VERSION
+#error CUDART_VERSION undefined!
+#elif (defined(KOKKOS_ENABLE_IMPL_CUDA_MALLOC_ASYNC) && CUDART_VERSION >= 11020)
+
+/**
+ *  If a device has been initialized we will drop it in the set
+ * so we don't initialize it again
```
masterleinad, line 184:
```suggestion
 * If a device has been initialized we will drop it in the set
 * so we don't initialize it again
```

core/src/Cuda/Kokkos_CudaSpace.cpp
```diff
@@ -172,6 +174,115 @@ void *CudaSpace::allocate(const char *arg_label, const size_t arg_alloc_size,
...
+    requested_size = static_cast<size_t>(std::stod(mempool_size_string));
+  } catch (...) {
+    std::cerr << "Unable to convert " << mempool_size_string
+              << " to a number\n";
+    return false;
+  }
+
+  requested_size *= factor;
+  size_t n_bytes = static_cast<size_t>(std::ceil(requested_size));
+  if (!(n_bytes > 0)) return false;
```
masterleinad, line 244: `size_t` is unsigned so this equivalent to
```suggestion
  if (n_bytes == 0) return false;
```
but do we need this check then or would you rather wan to check that `requested_size` is not greater than the largest number representable by `size_t`?

core/src/Cuda/Kokkos_CudaSpace.cpp
```diff
@@ -172,6 +174,115 @@ void *CudaSpace::allocate(const char *arg_label, const size_t arg_alloc_size,
...
+
+  std::cout << "Allocating " << n_bytes << " bytes for mempool\n";
```
masterleinad, line 246:
```suggestion
```

core/src/Cuda/Kokkos_CudaSpace.cpp
```diff
@@ -172,6 +174,115 @@ void *CudaSpace::allocate(const char *arg_label, const size_t arg_alloc_size,
...
+ *    @param stream     A specific stream for cudaMallocAsync
+ * 		@param error_code return any failure codes for cuda APIs
+ *    @return  				  true for success, false for failure
+ */
+bool initializeMempool(const int device_id, const cudaStream_t stream,
+                       cudaError_t &error_code) {
+  error_code = cudaSuccess;
+
+  std::cout << "Initializing Default Memory Pool for device " << device_id
+            << "\n";
```
masterleinad, line 201:
```suggestion;
```

core/src/Cuda/Kokkos_CudaSpace.cpp
```diff
@@ -172,6 +174,115 @@ void *CudaSpace::allocate(const char *arg_label, const size_t arg_alloc_size,
...
+  // d) free the memory, ensuring the pool retains what we asked for
+  // All the API calls need to return cudaSuccess or we fail
+
+  // First get the default memory pool
+  error_code = cudaSuccess;
+  cudaMemPool_t default_mempool;  // The memory pool handle
+
+  // Cuda API call to get the memory pool for device with device_id
+  error_code = cudaDeviceGetDefaultMemPool(&default_mempool, device_id);
+  if (error_code != cudaSuccess) return false;  // handle failure
```
masterleinad, line 261:
```suggestion
  KOKKOS_IMPL_CUDA_SAFE_CALL((cudaDeviceGetDefaultMemPool(&default_mempool, device_id)));
```
and similarly below.

core/src/Cuda/Kokkos_CudaSpace.cpp
```diff
@@ -185,20 +296,54 @@ void *impl_allocate_common(const int device_id,
...
 #error CUDART_VERSION undefined!
 #elif (defined(KOKKOS_ENABLE_IMPL_CUDA_MALLOC_ASYNC) && CUDART_VERSION >= 11020)
-  if (arg_alloc_size >= memory_threshold_g) {
-    error_code = cudaMallocAsync(&ptr, arg_alloc_size, stream);
-
-    if (error_code == cudaSuccess) {
-      if (stream_sync_only) {
-        KOKKOS_IMPL_CUDA_SAFE_CALL(cudaStreamSynchronize(stream));
+  // Always use the Async. Don't use the 40K Lower limit- may well be arch
+  // dependent.
```
masterleinad, line 300:
```suggestion
  // Always use the Async. Don't use the 40K Lower limit - may well be arch
  // dependent.
```
Is `memory_threshold_g` used anymore?

core/src/Cuda/Kokkos_CudaSpace.cpp
```diff
@@ -185,20 +296,54 @@ void *impl_allocate_common(const int device_id,
...
+
+  // Check the set of allocated devices, to check if our device is already
+  // allocate it
+  if (_mem_pool_initialized.find(device_id) == _mem_pool_initialized.end()) {
+    // Not found: so initialize it
+    //
```
masterleinad, line 306:
```suggestion
    // Not found: so initialize it
```

core/src/Cuda/Kokkos_CudaSpace.cpp
```diff
@@ -185,20 +296,54 @@ void *impl_allocate_common(const int device_id,
...
+        throw Experimental::CudaRawMemoryAllocationFailure(
+            arg_alloc_size, error_code,
+            Experimental::RawMemoryAllocationFailure::AllocationMechanism::
+                CudaMallocAlignedPoolSetup);
       } else {
-        Impl::cuda_device_synchronize(
-            "Kokkos::Cuda: backend fence after async malloc");
+        // if the API calls all returned cudaSuccess, the strtod() could
+        // have thrown an exception which we caught. In this case we
+        // throw a dumb exception ehere
```
masterleinad, line 320:
```suggestion
        // throw a dumb exception here
```

core/src/Cuda/Kokkos_CudaSpace.cpp
bjoo: What is the comment here? It seems empty

core/src/Cuda/Kokkos_CudaSpace.cpp
bjoo: So I can change it to that. Currently what will happen is that false will be returned but the `error_code` will still be `cudaSuccess` and the error is dealt with as an exception on line 320. The logic is that if this routine returns false either a CUDA API failed or the parsing failed. The failure of the CUDA API can be determined by looking at `error_code`

However I am happy to accommodate whichever way you prefer to promulgate the error to the user. I can put the error here, or at the point of raising the exception.

core/src/Cuda/Kokkos_CudaSpace.cpp
bjoo: Should be resolved in: 7e6b25a59

core/src/Cuda/Kokkos_CudaSpace.cpp
bjoo: Added a brief report here about not being able to parse units in af41845ef

bjoo (commented):
Hi @masterleinad 
 I accepted the cosmetic changes, and pushed changes to deal with the some error reporting. One real issue remains I guess which is with the 'safe call' v.s. my return fals approach. The latter one is designed to return false either if there is a parse error or if there is a Cuda error. The Cuda error is reflected in the value of `error_code`. Let me know how you want to deal with that. I can do whichever way is best for you.

core/src/Cuda/Kokkos_CudaSpace.cpp
```diff
@@ -172,6 +175,132 @@ void *CudaSpace::allocate(const char *arg_label, const size_t arg_alloc_size,
...
+    requested_size = static_cast<size_t>(std::stod(mempool_size_string));
+  } catch (...) {
+    std::cerr << "Unable to convert " << mempool_size_string
+              << " to a number\n";
+    return false;
+  }
+
+  // Check for non-positive size memory requests
+  if (requested_size <= 0) {
+    std::cerr << "Negative amount of memory requested in allocation\n";
```
masterleinad, line 245:
```suggestion
  // Check for non-positive size memory requests
  if (requested_size == 0) {
    std::cerr << "No memory requested in allocation\n";
```
`size_t` is unsigned and can't be negative. Should it be an error to request 0 bytes here?

core/src/Cuda/Kokkos_CudaSpace.cpp
masterleinad: Fair enough.

core/src/Cuda/Kokkos_CudaSpace.cpp
```diff
@@ -172,6 +175,132 @@ void *CudaSpace::allocate(const char *arg_label, const size_t arg_alloc_size,
...
+  double max_size_t = static_cast<double>(std::numeric_limits<size_t>::max());
+  if (requested_size > max_size_t) {
+    std::cerr << "Requested amount of memory " << requested_size << " exceeds "
+              << "maximum alloatable size " << max_size_t << "\n";
+    return false;
+  }
+
+  // At this point requested_size should be appropriate
+  // neither too big nor negative.
+  size_t n_bytes = static_cast<size_t>(std::ceil(requested_size));
```
masterleinad, line 264: This could be larger than the largest representable `size_t`, though.

core/src/Cuda/Kokkos_CudaSpace.cpp
```diff
@@ -172,6 +175,132 @@ void *CudaSpace::allocate(const char *arg_label, const size_t arg_alloc_size,
...
+  if (error_code != cudaSuccess) return false;  // handle failure
+
+  // Now we allocate the pool size amount + a little more (64 bytes is arbitrary
+  // it just needs to be more than the retention size.
+  void *dummy_ptr = nullptr;
+  error_code      = cudaMallocAsync(&dummy_ptr, n_bytes + 64, stream);
+  if (error_code != cudaSuccess) return false;  // handle failure
+
+  // Now we free the memory, and our poolsize amount will be retained in the
+  // pool
```
masterleinad, line 294:
```suggestion
  // Now we free the memory, and our pool size amount will be retained in the
  // pool
```

core/src/Cuda/Kokkos_CudaSpace.cpp
```diff
@@ -185,20 +314,53 @@ void *impl_allocate_common(const int device_id,
...
 #error CUDART_VERSION undefined!
 #elif (defined(KOKKOS_ENABLE_IMPL_CUDA_MALLOC_ASYNC) && CUDART_VERSION >= 11020)
-  if (arg_alloc_size >= memory_threshold_g) {
-    error_code = cudaMallocAsync(&ptr, arg_alloc_size, stream);
-
-    if (error_code == cudaSuccess) {
-      if (stream_sync_only) {
-        KOKKOS_IMPL_CUDA_SAFE_CALL(cudaStreamSynchronize(stream));
+  // Always use the Async. Don't use the 40K Lower limit - may well be arch
+  // dependent.
```
masterleinad, line 318: We should check `cudaDevAttrMemoryPoolsSupported` before going here, see https://github.com/kokkos/kokkos-core-wiki/pull/525#discussion_r1615644650, but probably not in this pull request.

benchmarks/async_alloc/async_alloc.cpp
```diff
@@ -0,0 +1,104 @@
...
+using MemorySpace = Kokkos::DefaultExecutionSpace::memory_space;
+
+// Store the individual timings
+std::vector<std::pair<size_t, double>> inner_loop_times;
+
+// The pair returned are the first_alloc_time and the loop time
+//
+std::pair<double, double> test(bool up) {
+  int iters      = 50;
+  size_t minimum = 8 / sizeof(float);  // 64K
```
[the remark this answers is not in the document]
bjoo: Yes. originally I was going for more. I see 'comment-rot' there. The comment about 64K is not true. I wanted to start there but others suggested 8 bytes. I will fix.

benchmarks/async_alloc/async_alloc.cpp
```diff
@@ -0,0 +1,104 @@
...
+  std::vector<size_t> sizes;
+  if (up) {
+    for (size_t num = minimum; num <= maximum; num *= 2) sizes.push_back(num);
+  } else {
+    for (size_t num = maximum; num >= minimum; num /= 2) sizes.push_back(num);
+  }
+
+  Kokkos::Timer first_alloc_timer;
+  {  // Prime the pump - first long alloc -- Time it.
+    Kokkos::View<float *, MemorySpace> dummy("unlabeled", 64);
```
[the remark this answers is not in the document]
bjoo: The size doesn't matter, yes I went for 256K here, but it is not important. I just want one first call that will call the allocator. Then it will set the poolsize. As you see the view is then immediately freed.

benchmarks/async_alloc/async_alloc.cpp
```diff
@@ -0,0 +1,104 @@
...
+
+  // FIXME: Exponential stepping here
+  Kokkos::Timer inner_loop_timer;
+  Kokkos::Timer alloc_loop_timer;
+
+  // Measured
+  for (size_t num : sizes) {
+    inner_loop_timer.reset();
+    for (int i = 0; i < iters; i++) {
+      Kokkos::View<float *, MemorySpace> a("unlabeled", num);
```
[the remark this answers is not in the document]
bjoo: Right. That is a good suggestion. I should accept this and redraw my graph.

benchmarks/async_alloc/async_alloc.cpp
```diff
@@ -0,0 +1,104 @@
...
+    }
+    double inner_loop_time = inner_loop_timer.seconds();
+
+    // Store in vector
+    inner_loop_times.push_back(std::make_pair<>(
+        num * sizeof(float), inner_loop_time / static_cast<double>(iters)));
```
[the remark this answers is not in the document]
bjoo: Yeah... Maybe... I think push_back is just as good in this instance?

benchmarks/async_alloc/async_alloc.cpp
```diff
@@ -0,0 +1,104 @@
...
+  std::cout << "First Alloc: " << 64 << " bytes, " << first_alloc_time
+            << " sec\n";
+  std::cout << "Test Alloc Loop Total: " << alloc_loop_time << " sec\n";
+  std::cout << "Alloc Loop Timings:\n";
+  std::cout << "===================\n";
+
+  std::cout << "# size (B) \t time (sec) \n";
+  std::cout << "# -----------------------\n";
+  std::sort(inner_loop_times.begin(), inner_loop_times.end(),
+            [=](const auto &a, const auto &b) { return a.first < b.first; });
```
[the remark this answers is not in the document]
bjoo: Sure. I was not aware of `revert`. Does it make much difference? It is not in a time critical part.

benchmarks/async_alloc/async_alloc.cpp
```diff
@@ -0,0 +1,104 @@
...
+  for (auto pair : inner_loop_times) {
+    std::cout << pair.first << ", " << pair.second << "\n";
```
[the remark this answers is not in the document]
bjoo: Happy to accept it but is this important? all the rest of the line is with the "-s

core/src/Cuda/Kokkos_CudaSpace.cpp
```diff
@@ -30,8 +30,11 @@
 #include <sstream>
 #include <algorithm>
 #include <atomic>
+#include <cctype>
+#include <unordered_set>
+#include <limits>
 
-//#include <Cuda/Kokkos_Cuda_BlockSize_Deduction.hpp>
+// #include <Cuda/Kokkos_Cuda_BlockSize_Deduction.hpp>
```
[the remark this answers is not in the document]
bjoo: I think that was left there from before. Happy to remove.

core/src/Cuda/Kokkos_CudaSpace.cpp
bjoo: It is true that `size_t` cannot be negative. However the user can still either wilfully or otherwise set the env var to "-1.4" -- that is allowed for a double. So I check the double for negativity before converting it to `size_t`. I am not sure whether requesting zero memory is allowed or not.

core/src/Cuda/Kokkos_CudaSpace.cpp
```diff
@@ -172,6 +175,132 @@ void *CudaSpace::allocate(const char *arg_label, const size_t arg_alloc_size,
...
+  if (requested_size <= 0) {
+    std::cerr << "Negative amount of memory requested in allocation\n";
+    return false;
+  }
+
+  // Multiply in units
+  requested_size *= factor;
+
+  // Check we are not asking for memory that is greater than what size_t can
+  // hold. Since requested can be the larger I convert the maximum of size_t
```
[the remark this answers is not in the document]
bjoo: OK.

core/src/Cuda/Kokkos_CudaSpace.cpp
```diff
@@ -172,6 +175,132 @@ void *CudaSpace::allocate(const char *arg_label, const size_t arg_alloc_size,
...
+  // to a double and compare those
+  double max_size_t = static_cast<double>(std::numeric_limits<size_t>::max());
+  if (requested_size > max_size_t) {
+    std::cerr << "Requested amount of memory " << requested_size << " exceeds "
+              << "maximum alloatable size " << max_size_t << "\n";
+    return false;
+  }
+
+  // At this point requested_size should be appropriate
+  // neither too big nor negative.
```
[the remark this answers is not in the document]
bjoo: I would prefer to keep the initial version here.  It's either "either-or"  or "neither-nor".

core/src/Cuda/Kokkos_CudaSpace.cpp
bjoo: On the other hand if I floor it we may be short. I can set the check on requested size to be   < max

core/src/Cuda/Kokkos_CudaSpace.cpp
```diff
@@ -172,6 +175,132 @@ void *CudaSpace::allocate(const char *arg_label, const size_t arg_alloc_size,
...
+  if (error_code != cudaSuccess) return false;  // handle failure
+
+  // Here we set the size of the retained memory in the pool after a
+  // free by setting the cudaMemPoolAttrReleaseThresold property of the mempool
+  error_code = cudaMemPoolSetAttribute(
+      default_mempool, cudaMemPoolAttrReleaseThreshold, &n_bytes);
+  if (error_code != cudaSuccess) return false;  // handle failure
+
+  // Now we allocate the pool size amount + a little more (64 bytes is arbitrary
+  // it just needs to be more than the retention size.
```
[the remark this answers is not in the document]
bjoo: Good catch... Thanks!

core/src/Cuda/Kokkos_CudaSpace.cpp
bjoo: OK

benchmarks/async_alloc/async_alloc.cpp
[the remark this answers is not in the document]
masterleinad: Just more concise. 🙂

benchmarks/async_alloc/async_alloc.cpp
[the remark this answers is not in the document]
masterleinad: It just took me a second to understand why you would sort here which is clearer if you use  `std::revert` or add a comment.

benchmarks/async_alloc/async_alloc.cpp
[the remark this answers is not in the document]
masterleinad: Not that important.

core/src/Cuda/Kokkos_CudaSpace.cpp
masterleinad: What you are testing here is the value after calling
```
static_cast<size_t>(std::stod(mempool_size_string));
```
though, `std::stod(mempool_size_string)` could be negative but isn't after casting to `size_t`. You might just want to skip that cast.

bjoo: Yes indeed.

crtrott (changes requested):
Blocking for now: I am not convinced that this is the right programmatic path forward for this for a couple reasons.

1. I am not comfortable adding a bunch of backend specific environment variables and command line options for Kokkos - this becomes very fast untestable in particular if people than start to ask for all the other options of stuff like the default memory pool.
2. I am not convinced that making the default allocators behave more like actual memory pools is the right thing to do. Generally, I want them to be just "system allocators". 
3. People would need to understand the implications of what that threshold does and my gut feeling is that in isolation it just becomes a "make it larger and then allocations are fast" without folks realizing what that means (e.g. that this is a fixed overhead on the GPU potentially causing issues in multi-tenant cases).
4. I think we should expose memory pool functionality as memory spaces which advertise that they are backed by a memory pool and which you pass around. These things can then have all the typical memory pool property and we could map it to backend specific ones if available instead of writing our own impl.

crtrott:
We may wanna organize a meeting for a larger discussion.

bjoo:
Hi Christian, Thanks for your comments and reasoning. I will reach out to you on the slack or direct email, and we can organize a meeting. Best wishes, B

core/src/Cuda/Kokkos_CudaSpace.cpp
```diff
@@ -172,6 +174,130 @@ void *CudaSpace::allocate(const char *arg_label, const size_t arg_alloc_size,
...
+    char last_letter_cap = std::toupper(last_letter);
+    switch (last_letter_cap) {
+      case 'K': factor = 1024; break;
+      case 'M': factor = 1024 * 1024; break;
+      case 'G': factor = 1024 * 1024 * 1024; break;
+      default:
+        std::cerr << "Unknown units for requested size " << last_letter << "\n";
+        return false;
+        break;
+    }
```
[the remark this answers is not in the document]
romintomasetti: Also, this function feels like it's the "opposite" of https://github.com/kokkos/kokkos/blob/662043966274936f2c7a367e58975f06c2814cf7/core/src/impl/Kokkos_Error.cpp#L47-L71, so it might be nice to be consistent with the notations ? (I mean `B`, `KiB` and so on)

core/src/Cuda/Kokkos_CudaSpace.cpp
```diff
@@ -172,6 +174,130 @@ void *CudaSpace::allocate(const char *arg_label, const size_t arg_alloc_size,
...
+  // c) allocate our desired poolsize amount plus a little more
+  // d) free the memory, ensuring the pool retains what we asked for
+  // All the API calls need to return cudaSuccess or we fail
+
+  // First get the default memory pool
+  error_code = cudaSuccess;
+  cudaMemPool_t default_mempool;  // The memory pool handle
+
+  // Cuda API call to get the memory pool for device with device_id
+  error_code = cudaDeviceGetDefaultMemPool(&default_mempool, device_id);
```
romintomasetti, line 275: As `Cuda` allows creating memory pool, and actually allows much more attributes to be set, I would propose to get this code a bit more "structured". Something like:
```c++
class CudaMemoryPool {
    cudaMemPool_t handle;

    static auto get_default(const Cuda& exec) {
        CudaMemoryPool pool;
        KOKKOS_IMPL_CUDA_SAFE_CALL(cudaDeviceGetDefaultMemPool(&pool, exec.device_id));
        return pool;
    }

    template <cudaMemPoolAttr Attribute, typename ValueType>
    void set(const ValueType& value) const {
        KOKKOS_IMPL_CUDA_SAFE_CALL(cudaMemPoolSetAttribute(handle, Attribute, &value));
    }

    void initialize(const Cuda& exec, const size_t n_bytes) const {
        void *dummy_ptr = nullptr;
        KOKKOS_IMPL_CUDA_SAFE_CALL(cudaMallocAsync(&dummy_ptr, n_bytes, exec.stream));
        KOKKOS_IMPL_CUDA_SAFE_CALL(cudaFreeAsync(dummy_ptr, exec.stream));
    }
};

std::unordered_set<CudaMemoryPool> memory_pools;
```

bjoo:
[like]  Balint Joo reacted to your message:
________________________________

crtrott:
Closing: we anyway moved away again from cudaMallocAsync due to too much problems downstream since it doesn't allow IPC. We will revisit memory pools as an explicit feature.

Issue #6918, opened 2024-04-04
`kokkos_malloc` should accept an execution space instance


romintomasetti:
### Summary

`Kokkos::kokkos_malloc` can be used to allocate memory in a memory space, allowing the memory to be tracked by `Kokkos` (among other benefits such as label).

However, it seems it is not possible to pass an execution space instance (that would allow one to do stream ordered allocation).

Currently, the code looks like:
https://github.com/kokkos/kokkos/blob/a833fb00ba79b36354fd45e3fbae1e1621e187d0/core/src/Kokkos_Core.hpp#L152-L165

### Actions

1. Add overloads that allow an execution space instance to be provided.
2. Rename `Kokkos::kokkos_malloc` to `Kokkos::malloc` ?

Joint work with @maartenarnst while thinking of the requirements for a copy-to-device helper, that helps users generating the *vtable* on device.

romintomasetti:
@dalg24 @masterleinad @crtrott What do you think ?

PR #7035, opened 2024-05-28
Add ExecSpace constructor from MemSpace

tcclevenger:
This is the least evasive solution I thought of for the following problem.

### Problem: ###
Constructing a view from a cuda memory space defined on a non-default device does not allocate memory correctly:
```
Kokkos::View v(Kokkos::view_alloc(..., mem_space), N); // Allocation will be on default dev
                                                       // even if mem_space is not
```

The problem exists because inside `Kokkos_ViewMapping.hpp` and `Kokkos_SharedAlloc_tmpi.hpp`, default execution spaces are created and passed to `ZeroMemset` and `Impl::DeepCopy`, respectively.

This is a relatively common tripping point for multi-gpu, namely, code **not** in `core/src/Cuda` which work by creating default constructed exec/mem spaces which are obviously not on correct device (e.g. `kokkos_malloc()`, which poses similar problem for multi-gpu as mentioned in #6918).

### Solution: ###
Give each backend a constructor for execution space which takes in a memory space. Every other backend just default constructs, but it allows cuda to extract the correct stream.

### Notes: ###
- Leaving this as Draft since more tests would be needed. Didn't want to write them if people do not like this solution.
- Other solutions I considered...
  - Things like `ZeroMemset` and `DeepCopy` could have an overload that takes a memory space, then Cuda could internally extract the stream and create an execution space.
  - Memory spaces could contain a fall-back execution space instance that it points too instead of using a default constructed execution space.

### DIFF
--- a/containers/src/Kokkos_DynRankView.hpp
+++ b/containers/src/Kokkos_DynRankView.hpp
@@ -1096,7 +1096,8 @@ class DynRankView : public ViewTraits<DataType, Properties...> {
         prop_copy,
         Impl::DynRankDimTraits<typename traits::specialize>::
             template createLayout<traits, P...>(arg_prop, arg_layout),
-        Impl::ViewCtorProp<P...>::has_execution_space);
+        Impl::ViewCtorProp<P...>::has_execution_space,
+        Impl::ViewCtorProp<P...>::has_memory_space);
 
     // Setup and initialization complete, start tracking
     m_track.assign_allocated_record_to_uninitialized(record);
--- a/containers/src/Kokkos_OffsetView.hpp
+++ b/containers/src/Kokkos_OffsetView.hpp
@@ -1189,7 +1189,8 @@ class OffsetView : public ViewTraits<DataType, Properties...> {
 
     Kokkos::Impl::SharedAllocationRecord<>* record = m_map.allocate_shared(
         prop_copy, arg_layout,
-        Kokkos::Impl::ViewCtorProp<P...>::has_execution_space);
+        Kokkos::Impl::ViewCtorProp<P...>::has_execution_space,
+        Kokkos::Impl::ViewCtorProp<P...>::has_memory_space);
 
     // Setup and initialization complete, start tracking
     m_track.assign_allocated_record_to_uninitialized(record);
--- a/core/src/Cuda/Kokkos_Cuda.hpp
+++ b/core/src/Cuda/Kokkos_Cuda.hpp
@@ -171,6 +171,11 @@ class Cuda {
 
   KOKKOS_DEPRECATED Cuda(cudaStream_t stream, bool manage_stream);
 
+  template <class MemSpace>
+  Cuda(const MemSpace& mem_space,
+       std::enable_if_t<Kokkos::is_memory_space<MemSpace>::value>* = nullptr)
+      : Cuda(mem_space.get_stream()) {}
+
   //--------------------------------------------------------------------------
   //! Free any resources being consumed by the device.
   static void impl_finalize();
--- a/core/src/Cuda/Kokkos_CudaSpace.hpp
+++ b/core/src/Cuda/Kokkos_CudaSpace.hpp
@@ -118,6 +118,8 @@ class CudaSpace {
   /**\brief Return Name of the MemorySpace */
   static constexpr const char* name() { return m_name; }
 
+  cudaStream_t get_stream() const { return m_stream; }
+
  private:
   int m_device;
   cudaStream_t m_stream;
@@ -213,6 +215,8 @@ class CudaUVMSpace {
     return CudaUVMSpace(device_id, stream);
   }
 
+  cudaStream_t get_stream() const { return m_stream; }
+
  private:
   int m_device;
   cudaStream_t m_stream;
@@ -298,6 +302,8 @@ class CudaHostPinnedSpace {
  public:
   /**\brief Return Name of the MemorySpace */
   static constexpr const char* name() { return m_name; }
+
+  cudaStream_t get_stream() const { return m_stream; }
 
  private:
   int m_device;
--- a/core/src/HIP/Kokkos_HIP.hpp
+++ b/core/src/HIP/Kokkos_HIP.hpp
@@ -52,6 +52,11 @@ class HIP {
       Impl::ManageStream manage_stream = Impl::ManageStream::no);
   KOKKOS_DEPRECATED HIP(hipStream_t stream, bool manage_stream);
 
+  template <class MemSpace>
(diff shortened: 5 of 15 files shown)


masterleinad:
The only way to create a non-default `MemorySpace` instance is by using
```C++
CudaSpace::impl_create(int device_id, cudaStream_t stream)
```
which users are not supposed to use. The first question is if we want to give users a legal way of specifying appropriate resources for a memory space instance. 

If so, in my opinion, the memory space should never own those resources and therefore I find it somewhat weird to be able to construct a memory space from an execution space. For me, device resources are encapsulated in an execution space object (even if the cuda stream within might not be managed by `Kokkos`). Thus, it would make more sense to me to provide the memory space with an execution space constructor.
In any case, we need to clarify what the behavior for a `View` constructor is if
- an execution space instance is provided
- a memory space instance is provided
- an execution space instance and a memory space instance are provided (and the underlying device/stream combos possibly don't match)
- neither an execution space instance nor a memory space instance is provided.

tcclevenger:
> The only way to create a non-default `MemorySpace` instance is by using
> 
> ```c++
> CudaSpace::impl_create(int device_id, cudaStream_t stream)
> ```
> 
> which users are not supposed to use. The first question is if we want to give users a legal way of specifying appropriate resources for a memory space instance.

I used `CudaSpace::impl_create` just to create a testing scenario, but is there not any other way for a user to be in possession of a memory space instance? Or is memory space itself a sort of internal object that the user only knows of the type?

> If so, in my opinion, the memory space should never own those resources and therefore I find it somewhat weird to be able to construct a memory space from an execution space. For me, device resources are encapsulated in an execution space object (even if the cuda stream within might not be managed by `Kokkos`). Thus, it would make more sense to me to provide the memory space with an execution space constructor.

Here it is the other way around, I'm constructing an exec space from a mem space. But I may be misunderstanding what you're saying. Alternatively, each memory space could have a `get_default_exec_space()` function that returns what a default exec space should look like for a given memory space.

>  In any case, we need to clarify what the behavior for a `View` constructor is if
> 
>     * an execution space instance is provided
> 
>     * a memory space instance is provided
> 
>     * an execution space instance and a memory space instance are provided (and the underlying device/stream combos possibly don't match)
> 
>     * neither an execution space instance nor a memory space instance is provided.

I agree, a lot more scenarios would need to be tested for the different `view_alloc()` args.

masterleinad:
> I used `CudaSpace::impl_create` just to create a testing scenario, but is there not any other way for a user to be in possession of a memory space instance? Or is memory space itself a sort of internal object that the user only knows of the type?

No, memory space instances aren't internal objects but are hardly ever used with a state. We added device and stream members so that allocations and deallocations would use the correct stream in cases where we don't have a full execution space available but setting those can only be done through `impl_create`. 

> Here it is the other way around, I'm constructing an exec space from a mem space. But I may be misunderstanding what you're saying. Alternatively, each memory space could have a get_default_exec_space() function that returns what a default exec space should look like for a given memory space.

The existing API has member functions that (optionally) take an execution space instance that would take precedence over the one stored. Thus, we are communicating the resource to be used for memory spaces at the user-level in a portable way through execution space instances already.
Hence, we could allow the user to set the default resource with a constructor taking an execution space instance which would make more sense to me than providing execution space instances with a constructor from a memory space instance. 

Also note that constructing an executions space instance is a lot more expensive than constructing a memory space instance.

> I agree, a lot more scenarios would need to be tested for the different view_alloc() args.

We just need to agree on the intended behavior first. To me, it makes sense to say: if a memory space instance is provided use that for allocations (although only the device really matters since we make sure that allocation is always synchronous), if an execution space instance is provided use that for initialization. If no memory space is provided construct one from the execution space instance for allocation (matching the current behavior that we are using the same stream for allocating as for initializing). 

I'm just not sure about the case we are discussing here where no execution space instance is provided. Should we make the expensive effort to create one internally or assume the default one is fine?

tcclevenger:
> I'm just not sure about the case we are discussing here where no execution space instance is provided. Should we make the expensive effort to create one internally or assume the default one is fine?

I want to make sure I understand, what we are doing currently is creating a default exec space (to pass to `ZeroMemset` and `DeepCopy`), but for example, creating a default `Kokkos::Cuda` instance is just internally pointing to the singleton `CudaInternal` which is already initialized, but creating a new `Kokkos::Cuda` (w/ a stream) is initializing a new `CudaInternal` which would be more expensive? I wasn't paying attention to the cost difference, so that makes sense.

> Ultimately, the question is if there are any places where providing a stateful memory space instance instead of an execution space instance is beneficial.

I think my question is, if it is not beneficial (for multi-gpu, I think it doesn't matter for all other cases) how should we handle it? Because I think it is a problem if someone reads documentation and nothing tells them otherwise that they can't use a memory space to construct a view. 

Since allocation is not the issue, only initialization, could we require that a view constructed without an exec space is used with `Kokkos::WithoutInitialization` tag?

masterleinad:
> I want to make sure I understand, what we are doing currently is creating a default exec space (to pass to `ZeroMemset` and `DeepCopy`), but for example, creating a default `Kokkos::Cuda` instance is just internally pointing to the singleton `CudaInternal` which is already initialized, but creating a new `Kokkos::Cuda` (w/ a stream) is initializing a new `CudaInternal` which would be more expensive? I wasn't paying attention to the cost difference, so that makes sense.

Yes, a default constructed execution space instance always gives you the same object that is initialized once when Kokkos is initialized; constructing from a stream always initializes a new object which includes various internal allocations. In particular, constructing execution space instances from the same stream gives you distinct objects but since they are sharing the same underlying resource they can't execute kernels concurrently (which isn't a problem).

> I think my question is, if it is not beneficial (for multi-gpu, I think it doesn't matter for all other cases) how should we handle it? Because I think it is a problem if someone reads documentation and nothing tells them otherwise that they can't use a memory space to construct a view.

The current situation is that users can only construct a default memory space instance which is (hopefully) effectively ignored internally. What we should fix for sure (if it doesn't work) is passing a non-default execution space instance (using a different device) and a default constructed memory space instance via `Kokkos::view_alloc`.

> Since allocation is not the issue, only initialization, could we require that a view constructed without an exec space is used with `Kokkos::WithoutInitialization` tag?

No, that would break pretty much every `Kokkos` code since people normally neither pass `Kokkos::WithoutInitializing` nor an execution space instance via `Kokkos::view_alloc`.

tcclevenger:
Closing. The approach I took here is not going to be used. 

> What we should fix for sure (if it doesn't work) is passing a non-default execution space instance (using a different device) and a default constructed memory space instance via Kokkos::view_alloc.

I will look into this and any changes needed will be in a separate PR.

PR #7039, opened 2024-05-28
Improve Kokkos Graphs

masterleinad:
Part of #6912. This pull request removes some strings from `static_assert`s that are not helpful anymore and expands the Kokkos::Graphs tests to cover all policies that the `Cuda` and `HIP` implementation support, i.e., everything except `parallel_scan`. This requires some fixes to `HIP` parallel constructs.

### DIFF
--- a/core/src/Cuda/Kokkos_Cuda_Graph_Impl.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_Graph_Impl.hpp
@@ -117,10 +117,7 @@ struct GraphImpl<Kokkos::Cuda> {
   //  requires NodeImplPtr is a shared_ptr to specialization of GraphNodeImpl
   //  Also requires that the kernel has the graph node tag in it's policy
   void add_node(std::shared_ptr<NodeImpl> const& arg_node_ptr) {
-    static_assert(
-        NodeImpl::kernel_type::Policy::is_graph_kernel::value,
-        "Something has gone horribly wrong, but it's too complicated to "
-        "explain here.  Buy Daisy a coffee and she'll explain it to you.");
+    static_assert(NodeImpl::kernel_type::Policy::is_graph_kernel::value);
     KOKKOS_EXPECTS(bool(arg_node_ptr));
     // The Kernel launch from the execute() method has been shimmed to insert
     // the node into the graph
--- a/core/unit_test/TestGraph.hpp
+++ b/core/unit_test/TestGraph.hpp
@@ -21,6 +21,21 @@
 
 namespace Test {
 
+template <class ExecSpace, class ValueType>
+struct NoOpReduceFunctor {
+  KOKKOS_FUNCTION void operator()(int, ValueType&) const {
+    Kokkos::abort("Should never be called!");
+  }
+  KOKKOS_FUNCTION void operator()(int, int, ValueType&) const {
+    Kokkos::abort("Should never be called!");
+  }
+  KOKKOS_FUNCTION void operator()(
+      const typename Kokkos::TeamPolicy<ExecSpace>::member_type&,
+      ValueType&) const {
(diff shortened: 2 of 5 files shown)


core/unit_test/TestGraph.hpp
```diff
@@ -208,9 +223,25 @@ TEST_F(TEST_CATEGORY_FIXTURE(graph), DISABLED_repeat_chain) {
...
+        NoOpReduceFunctor<TEST_EXECSPACE, int> no_op_functor;
+        root.then_parallel_reduce(Kokkos::RangePolicy<TEST_EXECSPACE>(0, 0),
+                                  no_op_functor, count)
+#if !defined(KOKKOS_ENABLE_SYCL)  // FIXME_SYCL
+#if !defined(KOKKOS_ENABLE_CUDA) && \
+    !defined(KOKKOS_ENABLE_HIP)  // FIXME_CUDA FIXME_HIP
+            .then_parallel_reduce(
+                Kokkos::MDRangePolicy<TEST_EXECSPACE, Kokkos::Rank<2>>{{0, 0},
+                                                                       {0, 0}},
+                no_op_functor, count)
```
dalg24, line 237: Why does this one not work with HIP nor CUDA?
What about SYCL?  Shouldn't it work if it is using the default graph implementation?

dalg24 (commented):
Please improve your PR description

core/unit_test/TestGraph.hpp
masterleinad: For SYCL in the default implementation, it's the same problem that already required skipping the `launch_six` test (and the problem wasn't obvious to me).
For Cuda and HIP, we don't treat empty kernel launches correctly, i.e., they are not inserted into the graph.

dalg24: Why is it incorrect to skip inserting them if they do nothing?

masterleinad: Because we are asserting that the node is not empty after setting it up and we actually need to reset the result variable (which we do correctly for the other policies).

Rombur: Does the test fail with HIP and CUDA? If it passes, we should change the test to expose the problem

masterleinad: It fails with HIP and CUDA if not disabling the MDRangePolicy test. After disabling the MDRangePolicy test it passes.

core/src/Cuda/Kokkos_Cuda_Graph_Impl.hpp
```diff
@@ -117,10 +117,7 @@ struct GraphImpl<Kokkos::Cuda> {
   //  requires NodeImplPtr is a shared_ptr to specialization of GraphNodeImpl
   //  Also requires that the kernel has the graph node tag in it's policy
   void add_node(std::shared_ptr<NodeImpl> const& arg_node_ptr) {
-    static_assert(
-        NodeImpl::kernel_type::Policy::is_graph_kernel::value,
-        "Something has gone horribly wrong, but it's too complicated to "
-        "explain here.  Buy Daisy a coffee and she'll explain it to you.");
```
ldh4, line 123: I agree that this error message is not useful to the readers.
Is it feasible to replace it with some other useful message instead of completely removing it?

masterleinad: I wouldn't know what else to say. We basically add this trait to kernels that should be run in a graph and copy the kernel in the process. Thus, this `static_assert` could trigger if we forget to add this trait (or pass the original kernel here).

ldh4: Sounds fair.

ldh4 (approved):

PR #7040, opened 2024-05-28
Structured binding support for Kokkos::complex

nliber:
Following the model in [P2819R2](https://wg21..link/P2819R2) (voted into C++26 on 2023-11), this adds structured binding support for Kokkos::complex.

### DIFF
--- a/core/src/Kokkos_Complex.hpp
+++ b/core/src/Kokkos_Complex.hpp
@@ -38,8 +38,10 @@ template <size_t I, typename ComplexForwardingRef>
 KOKKOS_FUNCTION constexpr auto&& complex_get(
     ComplexForwardingRef&& z) noexcept {
   static_assert(I < 2);
-  if constexpr (I == 0) return static_cast<ComplexForwardingRef&&>(z).re_;
-  if constexpr (I == 1) return static_cast<ComplexForwardingRef&&>(z).im_;
+  if constexpr (I == 0)
+    return static_cast<ComplexForwardingRef&&>(z).re_;
+  else
+    return static_cast<ComplexForwardingRef&&>(z).im_;
 }
 }  // namespace Impl
 
--- a/core/src/Kokkos_Complex.hpp
+++ b/core/src/Kokkos_Complex.hpp
@@ -442,7 +442,8 @@ class
 
 }  // namespace Kokkos
 
-// Tuple protocol for complex based on https://wg21.link/p2819r2
+// Tuple protocol for complex based on https://wg21.link/p2819r2 (voted into
+// the C++26 working draft on 2023-11)
 template <typename RealType>
 struct std::tuple_size<Kokkos::complex<RealType>>
     : std::integral_constant<size_t, 2> {};
--- a/core/src/Kokkos_Complex.hpp
+++ b/core/src/Kokkos_Complex.hpp
@@ -301,24 +301,24 @@ class
 
   template <size_t I>
   KOKKOS_FUNCTION friend constexpr RealType& get(complex& z) noexcept {
-    return get_ref<I>(z);
+    return complex::get_ref<I>(z);
   }
 
   template <size_t I>
   KOKKOS_FUNCTION friend constexpr RealType&& get(complex&& z) noexcept {
-    return get_ref<I>(std::move(z));
+    return complex::get_ref<I>(std::move(z));
   }
 
   template <size_t I>
   KOKKOS_FUNCTION friend constexpr const RealType& get(
       const complex& z) noexcept {
-    return get_ref<I>(z);
+    return complex::get_ref<I>(z);
   }
 
   template <size_t I>
   KOKKOS_FUNCTION friend constexpr const RealType&& get(
       const complex&& z) noexcept {
-    return get_ref<I>(std::move(z));
+    return complex::get_ref<I>(std::move(z));
   }
 
 #ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
--- a/core/src/Kokkos_Complex.hpp
+++ b/core/src/Kokkos_Complex.hpp
@@ -31,39 +31,6 @@
 
 namespace Kokkos {
 
-template <typename RealType>
-class complex;
-
-// Tuple protocol for complex based on https://wg21.link/p2819r2 (voted into
-// the C++26 working draft on 2023-11)
-
-// get<...>(...) forward declared so as not to be hidden friends
-template <size_t I, typename RealType>
-constexpr RealType& get(complex<RealType>& z) noexcept;
-
-template <size_t I, typename RealType>
-constexpr RealType&& get(complex<RealType>&& z) noexcept;
-
-template <size_t I, typename RealType>
-constexpr const RealType& get(const complex<RealType>& z) noexcept;
-
-template <size_t I, typename RealType>
-constexpr const RealType&& get(const complex<RealType>&& z) noexcept;
-
-}  // namespace Kokkos
-
-template <typename RealType>
-struct std::tuple_size<Kokkos::complex<RealType>>
-    : std::integral_constant<size_t, 2> {};
-
-template <size_t I, typename RealType>
-struct std::tuple_element<I, Kokkos::complex<RealType>> {
-  static_assert(I < 2);
-  using type = RealType;
-};
-
-namespace Kokkos {
-
 /// \class complex
 /// \brief Partial reimplementation of std::complex that works as the
 ///   result of a Kokkos::parallel_reduce.
@@ -81,16 +48,6 @@ class
                     std::is_same_v<RealType, std::remove_cv_t<RealType>>,
                 "Kokkos::complex can only be instantiated for a cv-unqualified "
                 "floating point type");
-
-  template <size_t I, typename ComplexForwardingRef>
-  KOKKOS_FUNCTION static constexpr auto&& get_ref(
-      ComplexForwardingRef&& z) noexcept {
-    static_assert(I < 2);
-    if constexpr (I == 0)
-      return std::forward<ComplexForwardingRef>(z).re_;
-    else
-      return std::forward<ComplexForwardingRef>(z).im_;
-  }
 
  private:
   RealType re_{};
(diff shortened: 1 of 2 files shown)


ldh4:
I think the link in the description to P2819R2 is not correct.

masterleinad:
> I think the link in the description to P2819R2 is not correct.

It forwards me to https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/p2819r2.pdf.

ldh4:
> > I think the link in the description to P2819R2 is not correct.
> 
> It forwards me to https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/p2819r2.pdf.



> > I think the link in the description to P2819R2 is not correct.
> 
> It forwards me to https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/p2819r2.pdf.



> > I think the link in the description to P2819R2 is not correct.
> 
> It forwards me to https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2023/p2819r2.pdf.

It fails to forward me to anywhere on any of my browsers. But then maybe it's just my machine.

core/src/Kokkos_Complex.hpp
```diff
@@ -31,6 +31,20 @@
...
 
+// complex_get is a helper for tuple protocol
+// Needs to be declared before complex because complex grants it friendship
+namespace Impl {
+template <size_t I, typename ComplexForwardingRef>
+KOKKOS_FUNCTION constexpr auto&& complex_get(
+    ComplexForwardingRef&& z) noexcept {
+  static_assert(I < 2);
+  if constexpr (I == 0)
+    return static_cast<ComplexForwardingRef&&>(z).re_;
```
masterleinad, line 42: Any particular reason not to use
```suggestion
    return std::forward<ComplexForwardingRef>(z).re_;
```
(which might be more familiar for perfect forwarding)? Concerns about usability in device code?

core/unit_test/TestComplex.hpp
```diff
@@ -532,4 +531,68 @@ TEST(TEST_CATEGORY, complex_operations_arithmetic_types_overloads) {
...
+
+  Z z(2., 3.);
+
+  // get lvalue
+  Z &l = z;
+  static_assert(std::is_same_v<decltype(Kokkos::get<0>(l)), Z::value_type &>);
+  static_assert(std::is_same_v<decltype(Kokkos::get<1>(l)), Z::value_type &>);
+  auto &[lr, li] = l;
+  ASSERT_EQ(lr, l.real());
+  ASSERT_EQ(li, l.imag());
```
masterleinad, line 553: Does this also work in device code?

core/src/Kokkos_Complex.hpp
nliber: It was more that I'm not actually forwarding it; rather, I just want to make it the same kind of reference that was passed in.

I know I'm not consistent about this, using `std::move` in other places...

Should I change it?

core/unit_test/TestComplex.hpp
nliber: Why wouldn't it?

core/src/Kokkos_Complex.hpp
```diff
@@ -423,6 +440,43 @@ class
...
+struct std::tuple_element<I, Kokkos::complex<RealType>> {
+  static_assert(I < 2);
+  using type = RealType;
+};
+
+namespace Kokkos {
+
+template <size_t I, typename RealType>
+KOKKOS_FUNCTION constexpr RealType& get(complex<RealType>& z) noexcept {
+  return Impl::complex_get<I>(z);
```
dalg24, line 460: Why did you chose to introduce `Impl::complex_get` instead of having the various `get` be declared as hidden friends?

dalg24 (commented):
Please clearly states that the WG21 paper you refer to has been voted into the draft for C++26

core/src/Kokkos_Complex.hpp
nliber: It doesn't work as hidden friends.  That was the mistake LEWG made with P2819R1 (and was corrected by LWG for P2819R2).

dalg24: I had briefly looked in libcxx and they did friends but public

dalg24:
> Please clearly states that the WG21 paper you refer to has been voted into the draft for C++26

I meant in the description of your PR but the comment you added in the code is fine

core/unit_test/TestComplex.hpp
masterleinad: It seems to work fine, see https://godbolt.org/z/TGEvzYx5e (and our `Array` implementation). I'd still prefer if we also test in device code.

core/src/Kokkos_Complex.hpp
masterleinad: `std::move` wouldn't quite behave as expected, though, see (https://godbolt.org/z/hWq1M13r5), or am I missing something? I'd prefer `std::forward` over `static_assert` here.

nliber: In the two forms of `get` that return r-value references, I use `std::move(z)` instead of `static_cast<complex<RealType>&&>(z)` / `static_cast<const complex<const RealType>&&>(z)`, even though there is no intent to move, because it doesn't require stating the destination type (as that is deduced), so it is less error-prone.  This is inconsistent with the argument I presented for using `static_cast<ComplexForwardingRef&&>(z)` vs. `std::forward<ComplexForwardingRef>(z)`. 

Anyway, I'll change it to `std::forward<ComplexForwardingRef>(z)`.  What is an extra template instantiation in the grand scheme of things? :-)

core/unit_test/TestComplex.hpp
nliber: Done.

core/src/Kokkos_Complex.hpp
nliber: Done.

core/unit_test/TestComplex.hpp
```diff
@@ -532,4 +531,111 @@ TEST(TEST_CATEGORY, complex_operations_arithmetic_types_overloads) {
...
+                                 value_type &&>);
+    static_assert(std::is_same_v<decltype(Kokkos::get<1>(complex_type(5., 7.))),
+                                 value_type &&>);
+
+    auto &&[rr, ri] = complex_type(5., 7.);
+    d_results[2]    = complex_type(rr, ri);
+
+    // get const rvalue
+    auto rc = []() -> complex_type const && {
+      static const complex_type zz(11., 13.);
```
masterleinad, line 617:
```
dynamic initialization is not supported for a function-scope static __device__ variable within a __device__/__global__ function
```
```suggestion
      const complex_type zz(11., 13.);
```

nliber: I'll figure out a workaround.  It is `static` to get around an MSVC warning.

core/src/Kokkos_Complex.hpp
nliber: One can also do SBs via member `get` functions (harder to use in general, but that doesn’t matter if the intent is only to have these function to support SBs).

I initially had that approach working, but abandoned it as I didn’t want the interface to depart from that of `std::complex` any more than necessary.  Should I resurrect it?

dalg24: > It doesn't quite work? @crtrott, you were present for the LWG discussion of P2819R1 in Kona (I was not) when LWG told LEWG they were making it non-hidden. Thoughts?
> 
> As for "public", are you just saying you want me to move where the friend declaration is in the class?

I am suggesting you define `get` in the body of the class as a public friend function and you inline the definition and drop the `Impl::complex_get` helper.

nliber: @dalg24 So you are basically saying that you want me to implement [P2819R1](https://wg21.link/P2819R1), not [P2819R2](https://wg21.link/P2819R2) (and I assume remove the comment about this being approved for C++26, because P2819R1 was rejected by LWG).

I don't see the justification for going against what was ratified for C++26.

masterleinad (approved):

core/src/Kokkos_Complex.hpp
dalg24: I expect if you forward declare the get function and implement it as a friend that will follow the spec from R2 that is in the current draft. I slightly prefer doing it that way but won't fight if you disagree,

nliber: I'll do that (I thought you were proposing an interface change, not an implementation change).

I think I'll also change `Impl::complex_get` to a `private static` function in `complex` so it won't need friendship.  It was in `Impl` because it is somewhat independent of `T`, as its main concern is the correct reference type, but that required it being a `friend` because `.real()` and `.imag()` don't always return references.

nliber:
retest this please

crtrott:
Retest this please!

crtrott:
@nliber assuming the ORNL jenkins jobs are in decent shape, can you rebase on top of develop so we can test with the fixed Windows CUDA build.

core/src/Kokkos_Complex.hpp
nliber: I don't think you can forward declare `friend` functions for a templated class, which means they are still hidden friends.

If I use `__PRETTY_FUNCTION__` on:

```c++
template <size_t I, typename RealType>
constexpr RealType& get(complex<RealType>& z) noexcept { /* ... */ }
```
It returns

`RealType &Kokkos::get(complex<RealType> &) [I = 0UL, RealType = double]`

If I use `__PRETTY_FUNCTION__` on:

```c++
  template <size_t I>
friend constexpr RealType& get(complex& z) noexcept { /* ... */ }
```

it returns

`RealType &Kokkos::get(complex<RealType> &) [I = 0UL]`

If I put both definitions in, structured bindings calls the `friend` function.

I'll give it some thought over the weekend, but I'm leaning towards reverting it, as this involves some dark corners of the language...

nliber: Changed.  Made `get` a direct friend declaration in the two cases it is needed.

core/unit_test/TestComplex.hpp
```diff
@@ -545,6 +549,143 @@ TEST(TEST_CATEGORY, complex_operations_arithmetic_types_overloads) {
...
+  static_assert(std::is_same_v<std::tuple_size<complex_type>::type,
+                               std::integral_constant<size_t, 2>>);
+
+  // tuple_element
+  static_assert(
+      std::is_same_v<std::tuple_element_t<0, complex_type>, value_type>);
+  static_assert(
+      std::is_same_v<std::tuple_element_t<1, complex_type>, value_type>);
+
+  static void testgetreturnreferencetypes() {
```
dalg24, line 573: I would have preferred if you used underscores between words to increase readability

core/unit_test/TestComplex.hpp
```diff
@@ -31,6 +30,11 @@
 #endif
 #endif
 
+namespace {
+template <typename... Ts>
+KOKKOS_FUNCTION constexpr void maybe_unused(Ts &&...) noexcept {}
+}  // namespace
```
dalg24, line 36: I would have gone all the way and move everything in this file into the anonymous namespace

dalg24 (approved):
Looks good

dalg24:
Jenkins failures can be ignored
```
3: [ RUN      ] openacc.mdspan_minimal_functional
3: Failing in Thread:1
3:  File: (OpenACC API)
3:  Function: acc_copyout:1
3:  Line: 1
3: 
 3/20 Test  #3: Kokkos_CoreUnitTest_OpenACC .................***Failed   77.52 sec
```
is clearly unrelated,
all tests segfaulting in the HIP-ROCm-5.6-C++20 build, and 
GPU not detected by Kokkos::initialize in the OpenMPTarget build mid-way through running the test suite

dalg24:
The failure with ICPC on GH actions is also clearly unrelated
```
[  FAILED  ] openmp.Random_XorShift1024_0
```
Not worth rerunning

dalg24:
Make sure you add it to the change log

ndellingwood:
@nliber could you add an entry for the 4.4 changelog?

ndellingwood:
I added an entry to the 4.4 changelog, simple copy+paste of the PR title, please update if desired

PR #7042, opened 2024-05-29
Avoid using "#if not defined"

masterleinad:
MSVC doesn't like "#if not defined" and emits warnings like
```
D:\a\kokkos\kokkos\algorithms\unit_tests\TestStdAlgorithmsTeamMinElement.cpp(172): warning C4067: unexpected tokens following preprocessor directive - expected a newline [D:\a\kokkos\kokkos\build\algorithms\unit_tests\Kokkos_AlgorithmsUnitTest_StdSet_Team_D.vcxproj]
```
It's not clear to me if this is valid preprocessor syntax or not but replacing it with `#ifndef` clearly is more standard.

Note that `MSVC` actually doesn't treat this correctly, see https://godbolt.org/z/x9nhG1sWq.

### DIFF
--- a/algorithms/unit_tests/TestStdAlgorithmsTeamExclusiveScan.cpp
+++ b/algorithms/unit_tests/TestStdAlgorithmsTeamExclusiveScan.cpp
@@ -85,7 +85,7 @@ struct TestFunctorA {
         break;
       }
 
-#if not defined KOKKOS_ENABLE_OPENMPTARGET
+#ifndef KOKKOS_ENABLE_OPENMPTARGET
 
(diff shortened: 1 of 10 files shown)


nliber:
`not` is a keyword that is a synonym for `!` and is supposed to be valid.

I vaguely recall that MSVC made them macros, not keywords, and require `#include <iso646.h>` for that.

That being said, because MSVC doesn't take them by default, we should change our guards.

nliber (approved):

dalg24 (approved):

Issue #7037, opened 2024-05-28
Support building with no RTTI and no exception


dalg24:
In order to be able to support exotic architectures, we should consider avoiding the use of RTTI (runtime type information) and exceptions.
The goal of this is issue is mostly to gauge the interest.  We can discuss later the details of what we would need to change in the codebase to achieve this but in a nutshell, we'd ban the use of `dynamic_cast<>`, the operator `typeid` and "raw" throw expressions (we already mostly encapsulate the error reporting throw utilities such as `Impl::throw_runtime_exception()`).

PR #7043, opened 2024-05-29
Introduce `Impl::TypeInfo<T>::name()` as a replacement for `typeid(T).name()`

dalg24:
Partially addressing #7037 
Propose an alternative to the `typeid` operator.
We will need to discuss impact on tools before this can be merged.


core/src/Kokkos_TypeInfo.hpp
```diff
@@ -0,0 +1,92 @@
...
+}  // namespace Kokkos
+
+#else  // out of luck, using Intel C++ Compiler Classic
+
+namespace Kokkos {
+
+template <class T>
+class TypeInfo {
+ public:
+  static constexpr std::string_view name() noexcept { return "not supported"; }
```
crtrott, line 85: I would have returned here: `typeid(T).name();`, I get it that means its not constexpr but do we really need this string has a constexpr thing? I.e. does Kokkos::TypeInfo need to be compile time where ISO type_info is not? This would also isolate the ifdefing to just this one place not everywhere where it's used.

core/src/Kokkos_TypeInfo.hpp
```diff
@@ -0,0 +1,92 @@
...
+#ifndef KOKKOS_TYPE_INFO_HPP
+#define KOKKOS_TYPE_INFO_HPP
+
+#include <array>
+#include <string_view>
+#include <utility>
+
+#include <Kokkos_Macros.hpp>
+
+#ifndef KOKKOS_COMPILER_INTEL
```
crtrott, line 26: Doesn't that also exclude the clang based Intel compiler?

crtrott (changes requested):
Not sure I agree with the approach fully (i.e. where to diverge based on whether Intel is the compiler).

core/src/Kokkos_TypeInfo.hpp
masterleinad: No, that one defines `KOKKOS_COMPILER_INTEL_LLVM`.

nliber:
Note:  you can abstract out the clang/gcc differences.  I've done it here in the `pretty_type` function in [pretty_name.h](https://github.com/nliber/cool/blob/main/pretty_name.h) (feel free to take it).

core/src/Kokkos_TypeInfo.hpp
dalg24: It would not be usable in a constant expression in that case (that is the solution you suggest would not have the `constexpr` which means already a departure from the non ICC code path).

dalg24:
I purposefully did not yet update the `kokkosp.kernel_name.*` tests to highlight how this PR impacts the tooling interface.

Here is a summary of what changed for a few compilers:
**MSVC**
```
[ RUN      ] kokkosp.kernel_name_parallel_for
C:\projects\source\core\unit_test\tools\TestKernelNames.cpp(61): error: Expected equality of these values:
  last_parallel_for
    Which is: "const class `anonymous-namespace'::test_kernel_name_parallel_for::<lambda_80851a54a66e22be4a29bd358a7e54b9>"
  typeid(my_lambda).name()
    Which is: "class <lambda_80851a54a66e22be4a29bd358a7e54b9>"
[  FAILED  ] kokkosp.kernel_name_parallel_for (0 ms)
[ RUN      ] kokkosp.kernel_name_parallel_reduce
C:\projects\source\core\unit_test\tools\TestKernelNames.cpp(114): error: Expected equality of these values:
  last_parallel_reduce.find(suffix)
    Which is: 18446744073709551615
  last_parallel_reduce.length() - suffix.length()
    Which is: 336
[  FAILED  ] kokkosp.kernel_name_parallel_reduce (0 ms)
[ RUN      ] kokkosp.kernel_name_parallel_scan
C:\projects\source\core\unit_test\tools\TestKernelNames.cpp(135): error: Expected equality of these values:
  last_parallel_scan
    Which is: "const class `anonymous-namespace'::test_kernel_name_parallel_scan::<lambda_e4bbb049a597eaa1f6e277d59491d5c2>"
  typeid(my_lambda).name()
    Which is: "class <lambda_e4bbb049a597eaa1f6e277d59491d5c2>"
[  FAILED  ] kokkosp.kernel_name_parallel_scan (0 ms)
[ RUN      ] kokkosp.kernel_name_internal
C:\projects\source\core\unit_test\tools\TestKernelNames.cpp(169): error: Expected equality of these values:
  empty_pcn.get()
    Which is: "struct `anonymous-namespace'::kokkosp_kernel_name_internal_Test::TestBody::ThisType"
  typeid(ThisType).name()
    Which is: "struct `private: virtual void __cdecl `anonymous namespace'::kokkosp_kernel_name_internal_Test::TestBody(void) __ptr64'::`2'::ThisType"
[  FAILED  ] kokkosp.kernel_name_internal (0 ms)
[ RUN      ] kokkosp.profiling_section
[       OK ] kokkosp.profiling_section (0 ms)
```
**LLVM Clang**
```
[ RUN      ] kokkosp.kernel_name_parallel_for
/__w/kokkos/kokkos/core/unit_test/tools/TestKernelNames.cpp:61: Failure
  last_parallel_for
    Which is: "const (lambda at /__w/kokkos/kokkos/core/unit_test/tools/TestKernelNames.cpp:55:28)"
  typeid(my_lambda).name()
    Which is: "ZN12_GLOBAL__N_129test_kernel_name_parallel_forEvE3$_0"

[  FAILED  ] kokkosp.kernel_name_parallel_for (0 ms)
[ RUN      ] kokkosp.kernel_name_parallel_reduce
/__w/kokkos/kokkos/core/unit_test/tools/TestKernelNames.cpp:97: Failure
Expected: (last_parallel_reduce.find(typeid(my_lambda).name())) != (std::string::npos), actual: 18446744073709551615 vs 18446744073709551615
Kokkos::Impl::CombinedFunctorReducer<(lambda at /__w/kokkos/kokkos/core/unit_test/tools/TestKernelNames.cpp:89:28), Kokkos::Impl::FunctorAnalysis<Kokkos::Impl::FunctorPatternInterface::REDUCE, Kokkos::RangePolicy<Kokkos::Threads>, (lambda at /__w/kokkos/kokkos/core/unit_test/tools/TestKernelNames.cpp:89:28), float>::Reducer> does not contain ZN12_GLOBAL__N_132test_kernel_name_parallel_reduceEvE3$_0

[  FAILED  ] kokkosp.kernel_name_parallel_reduce (0 ms)
[ RUN      ] kokkosp.kernel_name_parallel_scan
/__w/kokkos/kokkos/core/unit_test/tools/TestKernelNames.cpp:135: Failure
  last_parallel_scan
    Which is: "const (lambda at /__w/kokkos/kokkos/core/unit_test/tools/TestKernelNames.cpp:129:28)"
  typeid(my_lambda).name()
    Which is: "ZN12_GLOBAL__N_130test_kernel_name_parallel_scanEvE3$_0"

[  FAILED  ] kokkosp.kernel_name_parallel_scan (0 ms)
[ RUN      ] kokkosp.kernel_name_internal
/__w/kokkos/kokkos/core/unit_test/tools/TestKernelNames.cpp:169: Failure
  empty_pcn.get()
    Which is: "ThisType"
  typeid(ThisType).name()
    Which is: "ZN12_GLOBAL__N_133kokkosp_kernel_name_internal_Test8TestBodyEvE8ThisType"

[  FAILED  ] kokkosp.kernel_name_internal (0 ms)
```
**GCC**
```
[ RUN      ] kokkosp.kernel_name_parallel_for
/__w/kokkos/kokkos/core/unit_test/tools/TestKernelNames.cpp:61: Failure
  last_parallel_for
    Which is: "const {anonymous}::test_kernel_name_parallel_for()::<lambda(int)>"
  typeid(my_lambda).name()
    Which is: "ZN12_GLOBAL__N_129test_kernel_name_parallel_forEvEUliE_"

[  FAILED  ] kokkosp.kernel_name_parallel_for (0 ms)
[ RUN      ] kokkosp.kernel_name_parallel_reduce
/__w/kokkos/kokkos/core/unit_test/tools/TestKernelNames.cpp:96: Failure
Expected: (last_parallel_reduce.find(typeid(my_lambda).name())) != (std::string::npos), actual: 18446744073709551615 vs 18446744073709551615
CombinedFunctorReducer<{anonymous}::test_kernel_name_parallel_reduce()::<lambda(int, float&)>, FunctorAnalysis<FunctorPatternInterface::REDUCE, Kokkos::RangePolicy<Kokkos::OpenMP>, {anonymous}::test_kernel_name_parallel_reduce()::<lambda(int, float&)>, float>::Reducer, void> does not contain ZN12_GLOBAL__N_132test_kernel_name_parallel_reduceEvEUliRfE_

[  FAILED  ] kokkosp.kernel_name_parallel_reduce (0 ms)
[ RUN      ] kokkosp.kernel_name_parallel_scan
/__w/kokkos/kokkos/core/unit_test/tools/TestKernelNames.cpp:135: Failure
  last_parallel_scan
    Which is: "const {anonymous}::test_kernel_name_parallel_scan()::<lambda(int, float&, bool)>"
  typeid(my_lambda).name()
    Which is: "ZN12_GLOBAL__N_130test_kernel_name_parallel_scanEvEUliRfbE_"

[  FAILED  ] kokkosp.kernel_name_parallel_scan (0 ms)
[ RUN      ] kokkosp.kernel_name_internal
/__w/kokkos/kokkos/core/unit_test/tools/TestKernelNames.cpp:169: Failure
  empty_pcn.get()
    Which is: "{anonymous}::kokkosp_kernel_name_internal_Test::TestBody()::ThisType"
  typeid(ThisType).name()
    Which is: "ZN12_GLOBAL__N_133kokkosp_kernel_name_internal_Test8TestBodyEvE8ThisType"

[  FAILED  ] kokkosp.kernel_name_internal (0 ms)
```

dalg24:
Extracting here one of the test case:

|  | Before | After|
| --- | --- | --- |
| MSVC | "class <lambda_80851a54a66e22be4a29bd358a7e54b9>" | "const class \`anonymous-namespace'::test_kernel_name_parallel_for::<lambda_80851a54a66e22be4a29bd358a7e54b9>" |
| LLVM Clang | "ZN12_GLOBAL__N_129test_kernel_name_parallel_forEvE3$_0" | "const (lambda at /__w/kokkos/kokkos/core/unit_test/tools/TestKernelNames.cpp:55:28)" |
| GCC | "ZN12_GLOBAL__N_129test_kernel_name_parallel_forEvEUliE_" | "const {anonymous}::test_kernel_name_parallel_for()::<lambda(int)>" |

dalg24:
Note that while the length of defaulted kernel names generally increases, it often decreases for templated functors.
That is to say, I am showing you the worse case scenarios.

crtrott:
I did some compile time check and this new thing wins over typeid by a mile:

```
#include<Kokkos_Core.hpp>
#include<typeinfo>

template<size_t Idx>
bool instantiate_one() {
  #ifdef TYPEID
    auto s = typeid(Kokkos::View<int[Idx]>).name();
    return s[0]=='K';
  #else
    auto s = Kokkos::TypeInfo<Kokkos::View<int[Idx]>>::name();
    return s[0]=='K';
  #endif
}

// Packs can only be soo large
template<size_t Offset, size_t ... Idx>
bool instantiation_expander(std::index_sequence<Idx...>) {
  return (instantiate_one<Offset + Idx>() && ...);
}


  bool result = true;
  int R = argc > 1?atoi(argv[1]):10;
  Kokkos::initialize(argc,argv);
  {
    Kokkos::Timer timer;
    for(int r=0; r<R; r++) {
    result = result &&
             instantiation_expander<1>(std::make_index_sequence<64>()) &&
             instantiation_expander<100>(std::make_index_sequence<64>()) &&
             instantiation_expander<200>(std::make_index_sequence<64>()) &&
             instantiation_expander<300>(std::make_index_sequence<64>()) &&
             instantiation_expander<400>(std::make_index_sequence<64>()) &&
             instantiation_expander<500>(std::make_index_sequence<64>()) &&
             instantiation_expander<600>(std::make_index_sequence<64>()) &&
             instantiation_expander<700>(std::make_index_sequence<64>()) &&
             instantiation_expander<800>(std::make_index_sequence<64>()) &&
             instantiation_expander<900>(std::make_index_sequence<64>()) &&
             instantiation_expander<1000>(std::make_index_sequence<64>());
    }
    printf("Time %lf\n",time);
  }
  Kokkos::finalize();
  return result?0:1;
}
```

1.6s for TypeInfo vs 3.3s for typeid on my Mac with apple clang.

crtrott:
NVCC 11.7 its 7.7s vs 1m37s

dalg24:
Retest this please

dalg24:
Retest this please

core/unit_test/CMakeLists.txt
```diff
@@ -104,6 +105,20 @@ if(WIN32)
   LIST(APPEND COMPILE_ONLY_SOURCES TestWindowsInclude.cpp)
 endif()
 
+if(Kokkos_INSTALL_TESTING)  # FIXME Kokkos_ and KOKKOS_ variables are out of sync
```
cwpearson, line 108: Brief comment why there is some special case handling here (even just "See Kokkos_TypeInfo.hpp")

core/src/impl/Kokkos_Profiling_C_Interface.h
```diff
@@ -28,7 +28,7 @@
 #include <stdbool.h>
 #endif
 
-#define KOKKOSP_INTERFACE_VERSION 20211015
+#define KOKKOSP_INTERFACE_VERSION 20240906
```
cwpearson, line 31: Do we have a way to (or do we care to) document what has changed when this interface version changes?

cwpearson (approved):
I think this is a good. Two non-blocking suggestions.

core/src/impl/Kokkos_Profiling_C_Interface.h
dalg24: I don't think we currently have a way to do it.
Here the change would capture that fallback kernel names are not generated using the `typeid` operator any more and that tool could in principle skip demangling.

core/unit_test/CMakeLists.txt
dalg24: https://github.com/kokkos/kokkos/issues/7305 is intended to keep track of that issue.
I was not planning on adding a link in the code, it feels unnecessary.

crtrott (approved):
I had looked a bunch at this code also to fix the windows thing, so feel reasonably comfortable with it.

dalg24:
Ignoring NVHPC segfault in the OpenACC build.  Everything else passed.

PR #7047, opened 2024-05-30
Minor Kokkos::Array structured bindings improvement

nliber:
Added `static_assert`s so that `tuple_element` and `get` are always in range when called.


dalg24 (approved):

masterleinad (approved):

dalg24:
Failures are all clearly unrelated (at code checkout)

PR #7048, opened 2024-05-31
Add nightly CI on Frontier

Rombur:
This PR adds a `.gitlab-ci.yml` configuration file to run a nightly CI on Frontier. I use a minimal configuration because that's the configuration used to create the Kokkos module on Frontier. I have disable the `DeviceAndThreads` test because the test uses python and there is an issue when we try to load some python module.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,0 +1,13 @@
+test:
+  stage: test
+  tags: [frontier, shell]
+  id_tokens:
+    OLCF_ID_TOKEN:
+      aud: https://code.olcf.ornl.gov
+  script:
+    - module load rocm/6.0
+    - mkdir build
+    - cd build
+    - cmake -DCMAKE_CXX_COMPILER=hipcc -DKokkos_ENABLE_HIP=ON -DKokkos_ENABLE_TESTS=ON ..
+    - make -j16
+    - ctest -E Kokkos_CoreUnitTest_DeviceAndThreads -V
--- a/.olcf-gitlab-ci.yml
+++ b/.olcf-gitlab-ci.yml
@@ -6,8 +6,7 @@ test:
       aud: https://code.olcf.ornl.gov
   script:
     - module load rocm/6.0
-    - mkdir build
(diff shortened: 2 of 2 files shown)


.gitlab-ci.yml
dalg24, line 1: Maybe we should use a less generic name and include "olcf" in the file name.

masterleinad (approved):
Looks OK to me.

.gitlab-ci.yml
```diff
@@ -0,0 +1,13 @@
+test:
+  stage: test
+  tags: [frontier, shell]
+  id_tokens:
+    OLCF_ID_TOKEN:
+      aud: https://code.olcf.ornl.gov
+  script:
+    - module load rocm/6.0
+    - mkdir build
+    - cd build
```
dalg24, line 10: Do we need to clean up and delete the build directory?

.gitlab-ci.yml
```diff
@@ -0,0 +1,13 @@
...
+    - cmake -DCMAKE_CXX_COMPILER=hipcc -DKokkos_ENABLE_HIP=ON -DKokkos_ENABLE_TESTS=ON ..
```
dalg24, line 11: I'd prefer doing `cmake -B build [options...]` and `cmake --build build`

.gitlab-ci.yml
```diff
@@ -0,0 +1,13 @@
...
+    - make -j16
```
crtrott, line 12: since I assume this builds on a compute node anyway: would it reduce runtime to go to -j 40 or so?

dalg24: or just `cmake --build -j` with no number of processes argument

dalg24 (approved):

crtrott (approved):

dalg24:
Did you change anything?  We (or at least I) don;t care about the CI for this change set

Rombur:
> Did you change anything?

There was a couple of typos when I switched to `cmake -B` and `cmake --build`. It's working now

PR #7053, opened 2024-06-03
Update mdspan tpl and fix accessor type in mdspan conversion test

nmm0:
This is a mostly a update of mdspan to https://github.com/kokkos/mdspan/commit/18c332e07efc66991707a4d443fd7562b5746ccf

The update breaks one test in mdspan so this has a fix (please review core/unit_test/TestMDSpanConversion.hpp)

### DIFF
--- a/core/unit_test/TestMDSpanConversion.hpp
+++ b/core/unit_test/TestMDSpanConversion.hpp
@@ -22,6 +22,17 @@
 
 namespace {
 
+template <class Accessor>
+struct MakeConstAccessorImpl;
+
+template <template <class> class Accessor, class T>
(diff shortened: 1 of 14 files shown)


crtrott:
we need to fix the shadow thing

crtrott (changes requested):
We need to fix the shadow warning on the mdspan side first.

nmm0:
> we need to fix the shadow thing

Yeah I have a WIP on the mdspan side but our CI on mdspan doesn't use (I guess a recent enough version?) of gcc to test it. I get the warnings with GCC 12 on my local machine but CI doesn't fail even with -Wshadow

nmm0:
Switching to draft until mdspan changes get merged

PR #7046, opened 2024-05-29
Add dependabot for GitHub Actions

masterleinad:
Adding `dependabot` should help avoid warnings in GitHub actions about outdated dependencies and increase the scorecard value. Thus, we wouldn't need pull requests like https://github.com/kokkos/kokkos/pull/6826 anymore.

Example: https://github.com/dealii/dealii/pull/16514, https://github.com/kokkos/kokkos-kernels/pull/2218. 

We (someone with write-access to the repository) should also check the configuration at https://github.com/kokkos/kokkos/settings/security_analysis). Also, see https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuring-dependabot-version-updates#enabling-dependabot-version-updates.

PR #7054, opened 2024-06-03
Update scorecard GitHub workflow

masterleinad:
This hopefully fixes issues like
```
[...]
2024/06/01 01:50:48 error signing scorecard json results: error signing payload: getting key from Fulcio: verifying SCT: updating local metadata and targets: error updating to TUF remote mirror: invalid key
[...]
```
in the `scorecard` CI, see https://github.com/kokkos/kokkos/actions/runs/9326621637/job/25675392376. Also, see https://github.com/ossf/scorecard-action/issues/998 and https://github.com/ossf/scorecard-action/issues/997. Basically, updating helped there.

### DIFF
--- a/.github/workflows/scorecard.yml
+++ b/.github/workflows/scorecard.yml
@@ -33,12 +33,12 @@ jobs:
 
     steps:
       - name: "Checkout code"
-        uses: actions/checkout@8e5e7e5ab8b370d6c329ec480221332ada57f0ab # v3.5.2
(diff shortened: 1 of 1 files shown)


nmm0 (approved):
ci errors are clearly spurious. Let's give it a try!

crtrott (approved):

masterleinad:
> Isn;t it something that should be handled by Dependabot that we added in #7046 ?

Yes, except that we didn't put it in the correct directory (and it probably wouldn't have triggered in time anyway). Also, the motivation here was to fix a build error rather than updating for safety (but updating fixed the issue).

PR #7061, opened 2024-06-06
Refactor: Move logic of `create_mirror*` to `Impl::create_mirror*`

pzehner:
This PR is a refactor.

Its idea is to simplify the code by moving the duplicated logic of the various `create_mirror` and `create_mirror_view` (public functions) to `Impl::create_mirror` and `Impl::create_mirror_view` (private functions). The public functions now behave as pass-troughs that only call their private counterpart. All the public functions have an `auto` type, which means their actual type is also decided by the private functions they call.

This PR makes the code cleaner and more maintainable, as it removes most of the extraneous SFINAE of the public functions, and merges some of them.

As deducing the type of the public functions is more indirect, this PR induces more work for the compiler, which may have consequences.

This PR is part of a of a larger effort which aims to assess #6842, and is the following of #6955.

We faced two difficulties when creating this PR:

1. The `core/unit_test/TestViewAPI.hpp` test file used the [implementation object](https://github.com/kokkos/kokkos/blob/63f05204d0eaeec83e25f87eeb8025a188683f92/core/unit_test/TestViewAPI.hpp#L842) `Impl::MirrorViewType` incorrectly by passing it a whole `Space` when a `Space::memory_space` should be passed (note that this test file is the only one explicitly mentioning an implementation object, which is used in a functor for one test). The test code was simplified by removing this hard-coded type and obtaining it from a template argument instead; and
2. The `containers/src/Kokkos_DualView.hpp` source file defined [lambda functions](https://github.com/kokkos/kokkos/blob/63f05204d0eaeec83e25f87eeb8025a188683f92/containers/src/Kokkos_DualView.hpp#L1037) using `if constexpr` on template arguments to decide how to call `Kokkos::create_mirror_view`. We encountered a very surprising bug with GCC 11 and CUDA 12.2, where both branches where evaluated by the compiler when the type of `create_mirror_view` was `auto`, resulting in a compilation error. The problem was fixed by moving the `if constexpr` part in a templated function, which unfortunately makes the code less readable.

As discussed below, this PR reintroduced that public `create_mirror` functions check the input view is not `specialize`d, and extended this concept to offset views and dynamic views.

### DIFF
--- a/containers/src/Kokkos_DualView.hpp
+++ b/containers/src/Kokkos_DualView.hpp
@@ -1039,7 +1039,7 @@ class DualView : public ViewTraits<DataType, Properties...> {
       if (sizeMismatch) {
         ::Kokkos::resize(properties, d_view, n0, n1, n2, n3, n4, n5, n6, n7);
         // this part of the lambda was relocated in a method as it contains a
-        // `if constexpr` whose, in some cases, both branches can be evaluated,
+        // `if constexpr`. In some cases, both branches were evaluated
         // leading to a compile error
         resync_host(properties);
 
@@ -1053,7 +1053,7 @@ class DualView : public ViewTraits<DataType, Properties...> {
       if (sizeMismatch) {
         ::Kokkos::resize(properties, h_view, n0, n1, n2, n3, n4, n5, n6, n7);
         // this part of the lambda was relocated in a method as it contains a
-        // `if constexpr` whose, in some cases, both branches can be evaluated,
+        // `if constexpr`. In some cases, both branches were evaluated
         // leading to a compile error
         resync_device(properties);
 
@@ -1096,9 +1096,8 @@ class DualView : public ViewTraits<DataType, Properties...> {
 
  private:
   // resync host mirror from device
-  // this code was relocated from a lambda as it contains a `if constexpr`
-  // whose, in some cases, both branches can be evaluated, leading to a compile
-  // error
+  // this code was relocated from a lambda as it contains a `if constexpr`.
+  // In some cases, both branches were evaluated, leading to a compile error
   template <class... ViewCtorArgs>
   inline void resync_host(Impl::ViewCtorProp<ViewCtorArgs...> const&) {
     using alloc_prop_input = Impl::ViewCtorProp<ViewCtorArgs...>;
@@ -1113,8 +1112,7 @@ class DualView : public ViewTraits<DataType, Properties...> {
 
   // resync device mirror from host
   // this code was relocated from a lambda as it contains a `if constexpr`
-  // whose, in some cases, both branches can be evaluated, leading to a compile
-  // error
+  // In some cases, both branches were evaluated leading to a compile error
   template <class... ViewCtorArgs>
   inline void resync_device(Impl::ViewCtorProp<ViewCtorArgs...> const&) {
     using alloc_prop_input = Impl::ViewCtorProp<ViewCtorArgs...>;
--- a/containers/src/Kokkos_DualView.hpp
+++ b/containers/src/Kokkos_DualView.hpp
@@ -1038,7 +1038,7 @@ class DualView : public ViewTraits<DataType, Properties...> {
       /* Resize on Device */
       if (sizeMismatch) {
         ::Kokkos::resize(properties, d_view, n0, n1, n2, n3, n4, n5, n6, n7);
-        resync_host<ViewCtorArgs...>();
+        resync_host(properties);
 
         /* Mark Device copy as modified */
         ++modified_flags(1);
@@ -1049,7 +1049,7 @@ class DualView : public ViewTraits<DataType, Properties...> {
       /* Resize on Host */
       if (sizeMismatch) {
         ::Kokkos::resize(properties, h_view, n0, n1, n2, n3, n4, n5, n6, n7);
-        resync_device<ViewCtorArgs...>();
+        resync_device(properties);
 
         /* Mark Host copy as modified */
         ++modified_flags(0);
@@ -1088,11 +1088,13 @@ class DualView : public ViewTraits<DataType, Properties...> {
     }
   }
 
+  private:
+
+  // resync host mirror from device
   template <class... ViewCtorArgs>
-  inline void resync_host() {
+  inline void resync_host(Impl::ViewCtorProp<ViewCtorArgs...> const &) {
     using alloc_prop_input = Impl::ViewCtorProp<ViewCtorArgs...>;
 
-    // resync host mirror from device
     if constexpr (alloc_prop_input::initialize) {
       h_view = create_mirror_view(typename t_host::memory_space(), d_view);
     } else {
@@ -1101,11 +1103,11 @@ class DualView : public ViewTraits<DataType, Properties...> {
     }
   }
 
+  // resync device mirror from host
   template <class... ViewCtorArgs>
-  inline void resync_device() {
+  inline void resync_device(Impl::ViewCtorProp<ViewCtorArgs...> const &) {
     using alloc_prop_input = Impl::ViewCtorProp<ViewCtorArgs...>;
 
-    // resync device mirror from host
     if constexpr (alloc_prop_input::initialize) {
       d_view = create_mirror_view(typename t_dev::memory_space(), h_view);
 
@@ -1114,6 +1116,8 @@ class DualView : public ViewTraits<DataType, Properties...> {
                                   typename t_dev::memory_space(), h_view);
     }
   }
+
+  public:
 
   void resize(const size_t n0 = KOKKOS_IMPL_CTOR_DEFAULT_ARG,
               const size_t n1 = KOKKOS_IMPL_CTOR_DEFAULT_ARG,
--- a/containers/src/Kokkos_DualView.hpp
+++ b/containers/src/Kokkos_DualView.hpp
@@ -1038,6 +1038,9 @@ class DualView : public ViewTraits<DataType, Properties...> {
       /* Resize on Device */
       if (sizeMismatch) {
         ::Kokkos::resize(properties, d_view, n0, n1, n2, n3, n4, n5, n6, n7);
+        // this part of the lambda was relocated in a method as it contains a
+        // `if constexpr` whose, in some cases, both branches can be evaluated,
+        // leading to a compile error
         resync_host(properties);
 
         /* Mark Device copy as modified */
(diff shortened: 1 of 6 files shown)


core/src/Kokkos_CopyViews.hpp
```diff
@@ -3478,7 +3478,9 @@ void check_view_ctor_args_create_mirror() {
 // create a mirror
 // private interface that accepts arbitrary view constructor args passed by a
 // view_alloc
-template <class T, class... P, class... ViewCtorArgs>
+template <class T, class... P, class... ViewCtorArgs,
+          class = std::enable_if_t<
+              std::is_void<typename ViewTraits<T, P...>::specialize>::value>>
```
masterleinad, line 3483: `specialize` only matters for the public interface. Some `Trilinos` packages `specialize` `create_mirror` for their purposes and so it's important that `create_mirror_view` calls `create_mirror` (and not any function in the `Impl` namespace).

pzehner: Ok, I will reintroduce the `specialize` check for the public functions only.

I didn't get from the [documentation](https://kokkos.org/kokkos-core-wiki/API/core/view/view.html?highlight=specialize#_CPPv410specialize) what was the purpose of this tag.

containers/src/Kokkos_DualView.hpp
```diff
@@ -1099,6 +1088,33 @@ class DualView : public ViewTraits<DataType, Properties...> {
...
+    using alloc_prop_input = Impl::ViewCtorProp<ViewCtorArgs...>;
+
+    // resync host mirror from device
+    if constexpr (alloc_prop_input::initialize) {
+      h_view = create_mirror_view(typename t_host::memory_space(), d_view);
+    } else {
+      h_view = create_mirror_view(Kokkos::WithoutInitializing,
+                                  typename t_host::memory_space(), d_view);
+    }
+  }
```
tpadioleau, line 1102:
```suggestion
  template <class... ViewCtorArgs>
  void resync_host(Impl::ViewCtorProp<ViewCtorArgs> const&) {
    using alloc_prop_input = Impl::ViewCtorProp<ViewCtorArgs...>;

    // resync host mirror from device
    if constexpr (alloc_prop_input::initialize) {
      h_view = create_mirror_view(typename t_host::memory_space(), d_view);
    } else {
      h_view = create_mirror_view(Kokkos::WithoutInitializing,
                                  typename t_host::memory_space(), d_view);
    }
  }
```

I would favor template deduction even though the input parameter is not used.

Can you also ensure this function is not public ?

containers/src/Kokkos_DualView.hpp
```diff
@@ -1099,6 +1088,33 @@ class DualView : public ViewTraits<DataType, Properties...> {
...
+
+    // resync device mirror from host
+    if constexpr (alloc_prop_input::initialize) {
+      d_view = create_mirror_view(typename t_dev::memory_space(), h_view);
+
+    } else {
+      d_view = create_mirror_view(Kokkos::WithoutInitializing,
+                                  typename t_dev::memory_space(), h_view);
+    }
+  }
```
tpadioleau, line 1116: Same remarks here as `resync_host`

core/src/Kokkos_CopyViews.hpp
tpadioleau, line 3483:
```suggestion
              std::is_void_v<typename ViewTraits<T, P...>::specialize>>>
```

core/src/Kokkos_CopyViews.hpp
```diff
@@ -3558,7 +3545,9 @@ namespace Impl {
 // create a mirror view
 // private interface that accepts arbitrary view constructor args passed by a
 // view_alloc
-template <class T, class... P, class... ViewCtorArgs>
+template <class T, class... P, class... ViewCtorArgs,
+          class = std::enable_if<
+              std::is_void<typename ViewTraits<T, P...>::specialize>::value>>
```
tpadioleau, line 3550:
```suggestion
              std::is_void_v<typename ViewTraits<T, P...>::specialize>>>
```

core/unit_test/TestViewAPI.hpp
```diff
@@ -837,18 +837,15 @@ struct TestViewMirror {
...
   struct CopyUnInit {
-    using mirror_view_type = typename Kokkos::Impl::MirrorViewType<
-        Space, double *, Layout, Kokkos::HostSpace, MemoryTraits>::view_type;
-
-    mirror_view_type a_d;
+    View a_d;
 
     KOKKOS_INLINE_FUNCTION
-    CopyUnInit(mirror_view_type &a_d_) : a_d(a_d_) {}
+    explicit CopyUnInit(View &a_d_) : a_d(a_d_) {}
```
tpadioleau, line 845:
```suggestion
    explicit CopyUnInit(View const& a_d_) : a_d(a_d_) {}
```

That should allow you to use CTAD later `CopyUnInit(a_d)`.

core/unit_test/TestViewAPI.hpp
```diff
@@ -837,18 +837,15 @@ struct TestViewMirror {
...
   }
 
-  template <class MemoryTraits, class Space>
+  template <class View>
   struct CopyUnInit {
-    using mirror_view_type = typename Kokkos::Impl::MirrorViewType<
-        Space, double *, Layout, Kokkos::HostSpace, MemoryTraits>::view_type;
-
-    mirror_view_type a_d;
+    View a_d;
```
tpadioleau, line 842: Maybe add a `static_cast(Kokkos::is_view_v<View>, "Expecting a Kokkos::View")` ?

core/src/Kokkos_CopyViews.hpp
masterleinad: Ideally, yes. It's up to you if you want to fix it here or not.

tpadioleau (changes requested):

core/unit_test/TestViewAPI.hpp
pzehner: This is a test functor used in only one test. I don't think we need extra security.

core/src/Kokkos_CopyViews.hpp
pzehner: This part was removed from @masterleinad review. I followed your recommendation, however.

core/src/Kokkos_CopyViews.hpp
pzehner: This part was removed from @masterleinad review. I followed your recommendation, however.

pzehner: I think it's ok to generalize it here.

core/unit_test/TestViewAPI.hpp
pzehner: It didn't help with Intel classics.

containers/src/Kokkos_DualView.hpp
pzehner: Done.

pzehner:
Retest this please

pzehner:
Retest this please!

pzehner:
Retest this please

yasahi-hpc:
Retest this please

pzehner:
Retest this please

pzehner:
Retest this please

pzehner:
Retest this please

pzehner:
Retest this please

pzehner:
Retest this please

pzehner:
Retest this please

pzehner:
Retest this please

containers/src/Kokkos_DualView.hpp
```diff
@@ -1038,12 +1038,7 @@ class DualView : public ViewTraits<DataType, Properties...> {
       /* Resize on Device */
       if (sizeMismatch) {
         ::Kokkos::resize(properties, d_view, n0, n1, n2, n3, n4, n5, n6, n7);
-        if constexpr (alloc_prop_input::initialize) {
-          h_view = create_mirror_view(typename t_host::memory_space(), d_view);
-        } else {
-          h_view = create_mirror_view(Kokkos::WithoutInitializing,
-                                      typename t_host::memory_space(), d_view);
-        }
+        resync_host(properties);
```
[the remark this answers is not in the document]
pzehner: Indeed. I added a comment about it.

pzehner:
Retest this please

pzehner:
Retest this please

core/src/Kokkos_CopyViews.hpp
```diff
@@ -3592,68 +3587,49 @@ inline auto create_mirror_view(
...
 template <class Space, class T, class... P,
-          class Enable = std::enable_if_t<Kokkos::is_space<Space>::value>>
-typename Impl::MirrorViewType<Space, T, P...>::view_type create_mirror_view(
-    const Space& space, const Kokkos::View<T, P...>& src,
-    std::enable_if_t<!Impl::MirrorViewType<Space, T, P...>::is_same_memspace>* =
-        nullptr) {
-  return Kokkos::create_mirror(space, src);
+          class Enable = std::enable_if_t<Kokkos::is_space<Space>::value>,
+          typename     = std::enable_if_t<
+              std::is_void_v<typename ViewTraits<T, P...>::specialize>>>
```
crtrott, line 3611:
```suggestion
          class Enable = std::enable_if_t<Kokkos::is_space<Space>::value &&
              std::is_void_v<typename ViewTraits<T, P...>::specialize>>>
```

Doesn't this work too?

crtrott (changes requested):
There are a bunch of places where two constraints are applied via two `enable_if_t`, can't that be done via a single `enable_if_t` with a a compound condition? Is there a reason to prefer two template arguments, i.e. better error messages?

crtrott:
I am almost good with this, checked it pretty carefully. @pzehner see my above question: if there is a good reason to do it as two separate template argument for dual constraint overloads than I am happy to merge it as is, if not I'd rather have a compound condition with a single enable_if_t

core/src/Kokkos_CopyViews.hpp
pzehner: Let me check this. I think I just re-pasted the non-`specialize` check and forget to merge the `enable_if_t`s.

pzehner: Fixed.

masterleinad (approved):
Looks OK to me.

crtrott (approved):

Issue #7029, opened 2024-05-23
Assignemnt operator fails within `Kokkos_Cuda_BlockSize_Deduction.hpp` only when a Kokkos function is called from `Python` wrapping


s-sajid-ali:
When running a particular `kokkos::parallel_for` functor as part of a larger routine via it's Python wrapping, the copy by reference for the cuda device properties struct at the following line fails and leaves `prop` in a null initialized state which then causes a kernel launch failure:
https://github.com/kokkos/kokkos/blob/24b24d07561972697d882176db0ee16436e12cec/core/src/Cuda/Kokkos_Cuda_BlockSize_Deduction.hpp#L210

Here are the addresses of the two variables in question, right after line 210 of `kokkos/core/src/Cuda/Kokkos_Cuda_BlockSize_Deduction.hpp`:
```
(gdb) print &(prop)
$1 = (const cudaDeviceProp *) 0x7ffe0a11ac80 <Kokkos::Impl::CudaInternal::m_deviceProp>
(gdb) print &(cuda_instance->m_deviceProp)
$2 = (cudaDeviceProp *) 0x7ffea0ff6d80 <Kokkos::Impl::CudaInternal::m_deviceProp>
```
and they are clearly different!

<details>
<summary>And here are the values of `cuda_instance->m_deviceProp` and `prop` </summary>
<br>

```
(gdb) print cuda_instance->m_deviceProp
$4 = {name = "Tesla V100-PCIE-32GB", '\000' <repeats 235 times>, uuid = {
    bytes = "\017i\202bڍ\323\313}\t\344\343Ĺ\210", <incomplete sequence \342>}, luid = "\000\000\000\000\000\000\000", 
  luidDeviceNodeMask = 0, totalGlobalMem = 34079899648, sharedMemPerBlock = 49152, regsPerBlock = 65536, warpSize = 32, 
  memPitch = 2147483647, maxThreadsPerBlock = 1024, maxThreadsDim = {1024, 1024, 64}, maxGridSize = {2147483647, 65535, 65535}, 
  clockRate = 1380000, totalConstMem = 65536, major = 7, minor = 0, textureAlignment = 512, texturePitchAlignment = 32, 
  deviceOverlap = 1, multiProcessorCount = 80, kernelExecTimeoutEnabled = 0, integrated = 0, canMapHostMemory = 1, 
[...]

PR #5473, opened 2022-09-20
Use inline static member variables for CudaInternal

masterleinad:
Addressing https://github.com/kokkos/kokkos/pull/5442#discussion_r974176530 in correspondence of doing the same changes to the `HIP` backend in [`891fe55` (#5444)](https://github.com/kokkos/kokkos/pull/5444/commits/891fe55638db9f4c59d661b24728af5a1a6fb331).

PR #7065, opened 2024-06-07
Fix using shared libraries and -fvisibility=hidden

### LINKED ISSUE #6178: `inline static` variables in headers break CUDA kernels of a library loaded with `RTLD_DEEPBIND`

masterleinad:
Fixes https://github.com/kokkos/kokkos/issues/7029. Fixes #6178. Related to https://github.com/kokkos/kokkos/pull/6224. As reported in https://github.com/kokkos/kokkos/pull/5473#issuecomment-1299102792, we run into issues if using `inline static` variables when downstream code compiles with `-fvisibility-hidden`. This pull request makes sure that we test this case (by adding that flag to all the tests when building with shared libraries) and replace the respective variables for `Cuda` and `HIP`.

### DIFF
--- a/cmake/kokkos_tribits.cmake
+++ b/cmake/kokkos_tribits.cmake
@@ -160,10 +160,11 @@ FUNCTION(KOKKOS_ADD_EXECUTABLE_AND_TEST ROOT_NAME)
             )
         ENDIF()
     ENDIF()
-    # We noticed problems with -fvisibility-hidden for inline static variables
+    # We noticed problems with -fvisibility=hidden for inline static variables
     # if Kokkos was built as shared library.
-    IF(BUILD_SHARED_LIBS AND NOT KOKKOS_CXX_HOST_COMPILER_ID MATCHES MSVC)
-      TARGET_COMPILE_OPTIONS(${PACKAGE_NAME}_${ROOT_NAME} PRIVATE -fvisibility=hidden)
+    IF(BUILD_SHARED_LIBS)
+      SET_PROPERTY(TARGET ${PACKAGE_NAME}_${ROOT_NAME} PROPERTY VISIBILITY_INLINES_HIDDEN ON)
+      SET_PROPERTY(TARGET ${PACKAGE_NAME}_${ROOT_NAME} PROPERTY CXX_VISIBILITY_PRESET hidden)
     ENDIF()
 ENDFUNCTION()
 
--- a/cmake/kokkos_tribits.cmake
+++ b/cmake/kokkos_tribits.cmake
@@ -160,6 +160,11 @@ FUNCTION(KOKKOS_ADD_EXECUTABLE_AND_TEST ROOT_NAME)
             )
         ENDIF()
     ENDIF()
+    # We noticed problems with -fvisibility-hidden for inline static variables
+    # if Kokkos was built as shared library.
+    IF(BUILD_SHARED_LIBS AND NOT KOKKOS_CXX_HOST_COMPILER_ID MATCHES MSVC)
+      TARGET_COMPILE_OPTIONS(${PACKAGE_NAME}_${ROOT_NAME} PRIVATE -fvisibility=hidden)
+    ENDIF()
 ENDFUNCTION()
 
 FUNCTION(KOKKOS_SET_EXE_PROPERTY ROOT_NAME)
--- a/.jenkins
+++ b/.jenkins
@@ -155,6 +155,7 @@ pipeline {
                         sh 'echo "/opt/rocm/llvm/lib" > /etc/ld.so.conf.d/llvm.conf && ldconfig'
                         sh '''rm -rf build && mkdir -p build && cd build && \
                               cmake \
+                                -DBUILD_SHARED_LIBS=ON \
                                 -DCMAKE_BUILD_TYPE=Debug \
                                 -DCMAKE_CXX_COMPILER=hipcc \
                                 -DCMAKE_CXX_FLAGS="-Werror -Wno-unused-command-line-argument -DNDEBUG" \
@@ -445,6 +446,7 @@ pipeline {
                         sh 'ccache --zero-stats'
                         sh '''rm -rf build && mkdir -p build && cd build && \
                               cmake \
+                                -DBUILD_SHARED_LIBS=ON \
                                 -DCMAKE_BUILD_TYPE=Debug \
                                 -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
                                 -DCMAKE_CXX_COMPILER=$WORKSPACE/bin/nvcc_wrapper \
--- a/.jenkins
+++ b/.jenkins
@@ -155,7 +155,6 @@ pipeline {
                         sh 'echo "/opt/rocm/llvm/lib" > /etc/ld.so.conf.d/llvm.conf && ldconfig'
                         sh '''rm -rf build && mkdir -p build && cd build && \
                               cmake \
-                                -DBUILD_SHARED_LIBS=ON \
                                 -DCMAKE_BUILD_TYPE=Debug \
                                 -DCMAKE_CXX_COMPILER=hipcc \
                                 -DCMAKE_CXX_FLAGS="-Werror -Wno-unused-command-line-argument -DNDEBUG" \
@@ -191,6 +190,7 @@ pipeline {
                         sh 'ccache --zero-stats'
                         sh '''rm -rf build && mkdir -p build && cd build && \
                               cmake \
+                                -DBUILD_SHARED_LIBS=ON \
                                 -DCMAKE_BUILD_TYPE=RelWithDebInfo \
                                 -DCMAKE_CXX_COMPILER=hipcc \
                                 -DCMAKE_CXX_FLAGS="-Werror -Wno-unused-command-line-argument" \
--- a/core/src/Cuda/Kokkos_Cuda_Instance.cpp
+++ b/core/src/Cuda/Kokkos_Cuda_Instance.cpp
@@ -737,6 +737,14 @@ namespace Impl {
 int g_cuda_space_factory_initialized =
     initialize_space_factory<Cuda>("150_Cuda");
 
+int CudaInternal::m_cudaArch = -1;
+cudaDeviceProp CudaInternal::m_deviceProp;
+std::set<int> CudaInternal::cuda_devices = {};
+std::map<int, unsigned long *> CudaInternal::constantMemHostStagingPerDevice =
+    {};
+std::map<int, cudaEvent_t> CudaInternal::constantMemReusablePerDevice = {};
+std::map<int, std::mutex> CudaInternal::constantMemMutexPerDevice     = {};
+
 }  // namespace Impl
 
 }  // namespace Kokkos
--- a/core/src/Cuda/Kokkos_Cuda_Instance.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_Instance.hpp
@@ -91,10 +91,10 @@ class CudaInternal {
   int m_cudaDev = -1;
 
   // Device Properties
-  inline static int m_cudaArch = -1;
+  static int m_cudaArch;
   static int concurrency();
 
-  inline static cudaDeviceProp m_deviceProp;
+  static cudaDeviceProp m_deviceProp;
 
   // Scratch Spaces for Reductions
   mutable std::size_t m_scratchSpaceCount;
@@ -120,11 +120,10 @@ class CudaInternal {
   bool was_initialized = false;
   bool was_finalized   = false;
 
-  inline static std::set<int> cuda_devices = {};
-  inline static std::map<int, unsigned long*> constantMemHostStagingPerDevice =
-      {};
-  inline static std::map<int, cudaEvent_t> constantMemReusablePerDevice = {};
-  inline static std::map<int, std::mutex> constantMemMutexPerDevice     = {};
+  static std::set<int> cuda_devices;
+  static std::map<int, unsigned long*> constantMemHostStagingPerDevice;
+  static std::map<int, cudaEvent_t> constantMemReusablePerDevice;
+  static std::map<int, std::mutex> constantMemMutexPerDevice;
 
   static CudaInternal& singleton();
 
--- a/core/src/HIP/Kokkos_HIP_Instance.cpp
+++ b/core/src/HIP/Kokkos_HIP_Instance.cpp
@@ -353,6 +353,22 @@ void HIPInternal::finalize() {
   m_num_scratch_locks = 0;
 }
 
+int HIPInternal::m_hipDev                                     = -1;
+unsigned HIPInternal::m_multiProcCount                        = 0;
+unsigned HIPInternal::m_maxWarpCount                          = 0;
+std::array<HIPInternal::size_type, 3> HIPInternal::m_maxBlock = {0, 0, 0};
+unsigned HIPInternal::m_maxWavesPerCU                         = 0;
+int HIPInternal::m_shmemPerSM                                 = 0;
+int HIPInternal::m_maxShmemPerBlock                           = 0;
+int HIPInternal::m_maxThreadsPerSM                            = 0;
+
+hipDeviceProp_t HIPInternal::m_deviceProp;
+
+std::mutex HIPInternal::scratchFunctorMutex;
+unsigned long *HIPInternal::constantMemHostStaging = nullptr;
+hipEvent_t HIPInternal::constantMemReusable        = nullptr;
+std::mutex HIPInternal::constantMemMutex;
+
 //----------------------------------------------------------------------------
 
 Kokkos::HIP::size_type hip_internal_multiprocessor_count() {
--- a/core/src/HIP/Kokkos_HIP_Instance.hpp
+++ b/core/src/HIP/Kokkos_HIP_Instance.hpp
@@ -70,16 +70,16 @@ class HIPInternal {
  public:
   using size_type = ::Kokkos::HIP::size_type;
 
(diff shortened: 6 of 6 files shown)


masterleinad:
Retest this please.

cmake/kokkos_tribits.cmake
```diff
@@ -160,6 +160,11 @@ FUNCTION(KOKKOS_ADD_EXECUTABLE_AND_TEST ROOT_NAME)
             )
         ENDIF()
     ENDIF()
+    # We noticed problems with -fvisibility-hidden for inline static variables
+    # if Kokkos was built as shared library.
+    IF(BUILD_SHARED_LIBS AND NOT KOKKOS_CXX_HOST_COMPILER_ID MATCHES MSVC)
+      TARGET_COMPILE_OPTIONS(${PACKAGE_NAME}_${ROOT_NAME} PRIVATE -fvisibility=hidden)
+    ENDIF()
```
dalg24, line 167: Why did you decide to do this instead of using `CMAKE_CXX_VISIBILITY_PRESET=hidden` and `CMAKE_VISIBILITY_INLINES_HIDDEN=ON`?

dalg24 (commented):
Plz confirm that the Cuda and HIP builds were indeed failing with shared libraries w/o our replacing inline static data members.

cmake/kokkos_tribits.cmake
masterleinad: I wasn't aware of `CMake` attributes using them now.

cmake/kokkos_tribits.cmake
```diff
@@ -160,6 +160,11 @@ FUNCTION(KOKKOS_ADD_EXECUTABLE_AND_TEST ROOT_NAME)
             )
         ENDIF()
     ENDIF()
+    # We noticed problems with -fvisibility-hidden for inline static variables
```
nmm0: We could do: [CXX_VISIBILITY_PRESET](https://cmake.org/cmake/help/latest/prop_tgt/LANG_VISIBILITY_PRESET.html) it's slighlty more idomatic

nmm0: Also, the comment here is a bit weird, since it looks like we are enabling hidden visibility by default (which is a good idea, since it's keeps visibility the same as MSVC)

nmm0 (commented):
Minor comments

masterleinad:
> Plz confirm that the Cuda and HIP builds were indeed failing with shared libraries w/o our replacing inline static data members.

Yes

crtrott (approved):

cmake/kokkos_tribits.cmake
```diff
@@ -160,6 +160,12 @@ FUNCTION(KOKKOS_ADD_EXECUTABLE_AND_TEST ROOT_NAME)
             )
         ENDIF()
     ENDIF()
+    # We noticed problems with -fvisibility=hidden for inline static variables
+    # if Kokkos was built as shared library.
+    IF(BUILD_SHARED_LIBS)
+      SET_PROPERTY(TARGET ${PACKAGE_NAME}_${ROOT_NAME} PROPERTY VISIBILITY_INLINES_HIDDEN ON)
+      SET_PROPERTY(TARGET ${PACKAGE_NAME}_${ROOT_NAME} PROPERTY CXX_VISIBILITY_PRESET hidden)
```
dalg24, line 167: Why did you chose set target properties rather than set the CMAKE_ prefixed variables globally?

masterleinad: I preferred the per-target property to better control which targets should use it. It needs precisely to be the test targets and none of the implementation files. With CMake variables this feels harder to control.

nmm0: Yeah I think we should avoid setting CMAKE_* variables if possible

nmm0 (approved):
Looks good to me!

cmake/kokkos_tribits.cmake
dalg24: Why are you saying that.  I am talking about passing it in a few CI builds as argument, not setting it in our cmake files.

Issue #6923, opened 2024-04-08
Move `layout_iterate_type_selector` into `Impl` namespace


masterleinad:
Follow up we should move the `layout_iterate_type_selector` into `Impl` namespace. There is no reason it should be public. Note this thing is used in stuff like ViewCopy to figure out the type of the `MDRangePolicy`.

_Originally posted by @crtrott in https://github.com/kokkos/kokkos/pull/6907#pullrequestreview-1984244986_

jbadwaik:
I'll be interested to work on it.

dalg24:
We need to mark it as deprecated and defer to the trait class that definition that we move into `Impl::`.

nliber:
Any reason this issue is still open?

PR #7076, opened 2024-06-18
Move `layout_iterate_type_selector` into Impl namespace

jbadwaik:
- Address issue #6923
  - Copied the `layout_iterate_type_selector` into the Impl namespace
  - Deprecated the `layout_iterate_type_selector` in parent namespace

The CI runs successfully on [github-windows](https://github.com/jbadwaik/kokkos/actions/runs/9562090958) and [clang-format-check](https://github.com/jbadwaik/kokkos/actions/runs/9562492103). The first attempt had the wrong code format so running it again with the correctly formatted code. 

A question is that, should the `Impl` namespace be put in a separate Impl file? `impl/Kokkos_Layout.hpp` for example?

### DIFF
--- a/core/src/Kokkos_Layout.hpp
+++ b/core/src/Kokkos_Layout.hpp
@@ -222,38 +222,6 @@ template <typename Layout, class Enable = void>
 struct KOKKOS_DEPRECATED is_layouttiled : std::false_type {};
 #endif
 
-// The layout_iterate_type_selector is now deprecated.
-// Please use the Kokkos::Impl::layout_iterate_type_selector instead.
-#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
-template <typename... Layout>
-struct KOKKOS_DEPRECATED layout_iterate_type_selector {
-  static const Kokkos::Iterate outer_iteration_pattern =
-      Kokkos::Iterate::Default;
-  static const Kokkos::Iterate inner_iteration_pattern =
-      Kokkos::Iterate::Default;
-};
-
-template <>
-struct KOKKOS_DEPRECATED layout_iterate_type_selector<Kokkos::LayoutRight> {
-  static const Kokkos::Iterate outer_iteration_pattern = Kokkos::Iterate::Right;
-  static const Kokkos::Iterate inner_iteration_pattern = Kokkos::Iterate::Right;
-};
-
-template <>
-struct KOKKOS_DEPRECATED layout_iterate_type_selector<Kokkos::LayoutLeft> {
-  static const Kokkos::Iterate outer_iteration_pattern = Kokkos::Iterate::Left;
-  static const Kokkos::Iterate inner_iteration_pattern = Kokkos::Iterate::Left;
-};
-
-template <>
-struct KOKKOS_DEPRECATED layout_iterate_type_selector<Kokkos::LayoutStride> {
-  static const Kokkos::Iterate outer_iteration_pattern =
-      Kokkos::Iterate::Default;
-  static const Kokkos::Iterate inner_iteration_pattern =
-      Kokkos::Iterate::Default;
-};
-#endif
-
 namespace Impl {
 // For use with view_copy
 template <typename... Layout>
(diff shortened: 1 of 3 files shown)


core/src/Kokkos_Layout.hpp
```diff
@@ -222,6 +222,36 @@ template <typename Layout, class Enable = void>
...
+  static const Kokkos::Iterate inner_iteration_pattern = Kokkos::Iterate::Left;
+};
+
+template <>
+struct KOKKOS_DEPRECATED layout_iterate_type_selector<Kokkos::LayoutStride> {
+  static const Kokkos::Iterate outer_iteration_pattern =
+      Kokkos::Iterate::Default;
+  static const Kokkos::Iterate inner_iteration_pattern =
+      Kokkos::Iterate::Default;
+};
```
masterleinad, line 252: Please wrap in `#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4`.

jbadwaik: Done

core/src/Kokkos_Layout.hpp
```diff
@@ -222,6 +222,39 @@ template <typename Layout, class Enable = void>
...
+template <typename... Layout>
+struct KOKKOS_DEPRECATED layout_iterate_type_selector {
+  static const Kokkos::Iterate outer_iteration_pattern =
+      Kokkos::Iterate::Default;
+  static const Kokkos::Iterate inner_iteration_pattern =
+      Kokkos::Iterate::Default;
+};
+
+template <>
+struct KOKKOS_DEPRECATED layout_iterate_type_selector<Kokkos::LayoutRight> {
```
dalg24, line 237: IIRC we need to be careful with specializations of deprecated classes, some compilers warn about it.

masterleinad (dismissed):

core/src/Kokkos_Layout.hpp
dalg24: https://godbolt.org/z/T1f5KczGb

masterleinad: We can just deprecate the `static const` members instead. That doesn't trigger these spurious warnings.

dalg24: I would do a deprecated alias https://godbolt.org/z/E7cnTMPzq

jbadwaik:
Updated the code to use alias instead of duplicated code. Tested it with clang and gcc.

masterleinad (approved):

crtrott (approved):

crtrott (changes requested):
@jbadwaik thanks for your contribution, its almost good to go but it seems like it crashes the NVHPC compiler (i.e. the compiler segfaults ...) I will have a look now.

crtrott:
Args doesn't reproduce on my machine with the same compiler version ...

core/src/Kokkos_Layout.hpp
```diff
@@ -250,6 +251,15 @@ struct layout_iterate_type_selector<Kokkos::LayoutStride> {
   static const Kokkos::Iterate inner_iteration_pattern =
       Kokkos::Iterate::Default;
 };
+}  // namespace Impl
+
+// The layout_iterate_type_selector is now deprecated.
+// Please use the Kokkos::Impl::layout_iterate_type_selector instead.
```
dalg24, line 257: We should not suggest to use things in `Kokkos::Impl::`.  It will be obvious enough to developers and users are not allowed to reach for it.
```suggestion
```

dalg24 (commented):
I would prefer if we dropped the comment.  Looks good otherwise.

dalg24:
The OpenACC failure is clearly unrelated but logging it here because that is the 2nd time I see it this morning
```
3: [ RUN      ] openacc.mdspan_minimal_functional
3: Failing in Thread:1
3:  File: (OpenACC API)
3:  Function: acc_copyout:1
3:  Line: 1
```
I don't know what to think of the NVHPC segfault.  Let's just re-run.

Retest this please

PR #7080, opened 2024-06-19
Fix thread-safety for the Serial backend

masterleinad:
Splits the changes for the Serial backend from https://github.com/kokkos/kokkos/pull/6151.
This pull request ensures that kernels submitted to the same execution space instance from multiple threads don't run concurrently.


core/src/Serial/Kokkos_Serial.hpp
```diff
@@ -52,6 +52,9 @@ class SerialInternal {
  public:
   SerialInternal() = default;
 
+  SerialInternal(const SerialInternal&) = delete;
```
nliber, line 55: Why just these two instead of the Rule-of-5?

Technically this is just for emphasis, as having a `mutex` as a non-static member will delete all the move/copy operations.

core/src/Serial/Kokkos_Serial.cpp
```diff
@@ -59,6 +68,16 @@ void SerialInternal::finalize() {
...
   m_is_initialized = false;
+
+  // guard erasing from all_instances
+  {
+    std::scoped_lock lock(all_instances_mutex);
+    auto it = std::find(all_instances.begin(), all_instances.end(), this);
+    if (it == all_instances.end())
+      Kokkos::abort(
+          "Execution space instance to be removed couldn't be found!");
+    all_instances.erase(it);
```
nliber, line 79: I don't know if we are concerned about minimizing the amount of time the lock is held.  If so,
```suggestion
    swap(*it, all_instances.back());
    all_instances.pop_back();
```
is O(1) (although his doesn't change the overall complexity of `finalize()` as `find` is still O(N)).

nliber (changes requested):

core/src/Serial/Kokkos_Serial.hpp
masterleinad: > Technically this is just for emphasis, as having a mutex as a non-static member will delete all the move/copy operations.

Correct. I'm happy to remove if you prefer.

nliber: Let's remove it.

masterleinad: > Let's remove it.

Done.

nliber (approved):

core/src/Serial/Kokkos_Serial.cpp
```diff
@@ -59,6 +68,17 @@ void SerialInternal::finalize() {
...
   m_is_initialized = false;
+
+  // guard erasing from all_instances
+  {
+    std::scoped_lock lock(all_instances_mutex);
+    auto it = std::find(all_instances.begin(), all_instances.end(), this);
+    if (it == all_instances.end())
+      Kokkos::abort(
+          "Execution space instance to be removed couldn't be found!");
+    std::swap(*it, all_instances.back());
```
nliber, line 79: Giving it more thought:  you don't even need `swap`, as assignment will do:

```suggestion
    *it = all_instances.back();
```

Sorry about that.  The usual reason to prefer `swap` is when it is non-throwing, but pointer assignment doesn't throw either.

core/src/Serial/Kokkos_Serial.cpp
```diff
@@ -147,7 +170,9 @@ Serial::Serial(NewInstance)
     : m_space_instance(new Impl::SerialInternal, [](Impl::SerialInternal* ptr) {
         ptr->finalize();
         delete ptr;
-      }) {}
+      }) {
+  m_space_instance->initialize();
```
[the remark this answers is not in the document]
dalg24: `Kokkos::initialize` only initializes default instance.

dalg24, line 174: So this was an oversight right?  It means that the `m_is_initialized` member was never set to `true` but that somehow didn't cause an error :/

dalg24 (approved):

core/src/Serial/Kokkos_Serial.cpp
masterleinad: Right, before this pull request that only affected ` SerialInternal::is_initialized()`.

crtrott (approved):

masterleinad:
It's a bugfix. I'll add it to the changelog.

PR #7081, opened 2024-06-19
Use GCC 14.1 and C++ 26 in Jenkins nightly

Rombur:
Update the nightly CI to use the latest GCC

### DIFF
--- a/.jenkins_nightly
+++ b/.jenkins_nightly
@@ -70,10 +70,10 @@ pipeline {
                           '''
                     }      
                 }   
-                stage('GCC-13') {
+                stage('GCC-14') {
                     agent {
                         docker {
-                            image 'gcc:13.1'
+                            image 'gcc:14.1'
                             label 'docker'
                         }
                     }
@@ -88,7 +88,7 @@ pipeline {
                           mkdir -p build && cd build && \
                           cmake \
                             -DCMAKE_BUILD_TYPE=Release \
-                            -DCMAKE_CXX_STANDARD=23 \
+                            -DCMAKE_CXX_STANDARD=26 \
                             -DCMAKE_CXX_FLAGS=-Werror \
                             -DKokkos_ARCH_NATIVE=ON \
                             -DKokkos_ENABLE_COMPILER_WARNINGS=ON \


dalg24 (commented):
Please elaborate on why we want to test for a version of the standard that is not yet released rather than C++23 that we support.

Try to remember to use `[ci skip]...` next time

Rombur:
The entire point of this build is to have one build on the bleeding edge of what the compiler can do. If we are going to hold back the compiler/standard, we may as well remove the build.

crtrott:
Do we have C++23 coverage with GCC elsewhere? But I agree that I want a C++26 build too, though we may wanna use clang for that?

Rombur:
> Do we have C++23 coverage with GCC elsewhere? 

No, we don't really test different C++ standard. We have a single C++20 build and it's for HIP.

> we may wanna use clang for that?

We can but on Linux clang uses `libstdc++` not `libc++`, so we won't have mdspan by default. We should be able to make it work if it's what you want.

crtrott (approved):

ajpowelsnl (approved):
looks good

dalg24:
```
+ cmake -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_STANDARD=26 -DCMAKE_CXX_FLAGS=-Werror -DKokkos_ARCH_NATIVE=ON -DKokkos_ENABLE_COMPILER_WARNINGS=ON -DKokkos_ENABLE_BENCHMARKS=ON -DKokkos_ENABLE_EXAMPLES=ON -DKokkos_ENABLE_TESTS=ON -DKokkos_ENABLE_DEPRECATED_CODE_4=ON -DKokkos_ENABLE_DEPRECATION_WARNINGS=ON -DKokkos_ENABLE_SERIAL=ON ..
-- Setting default Kokkos CXX standard to 26
-- Kokkos version: 4.3.99
-- The project name is: Kokkos
-- Using internal gtest for testing
CMake Warning at cmake/kokkos_test_cxx_std.cmake:54 (MESSAGE):
  C++26 is not supported as a compiler feature.  We will choose custom flags
  for now, but this behavior has been deprecated.  Please open an issue at
  https://github.com/kokkos/kokkos/issues reporting that GNU 13.1.0 failed
  for 26, preferably including your CMake command.
  cmake/kokkos_test_cxx_std.cmake:78 (kokkos_set_cxx_standard_feature)
  cmake/kokkos_tribits.cmake:209 (INCLUDE)
  CMakeLists.txt:226 (KOKKOS_SETUP_BUILD_ENVIRONMENT)


CMake Error at cmake/kokkos_test_cxx_std.cmake:60 (MESSAGE):
  Compiler GNU should support cxx_std_26, but CMake reports feature not
  supported
  cmake/kokkos_test_cxx_std.cmake:78 (kokkos_set_cxx_standard_feature)
  cmake/kokkos_tribits.cmake:209 (INCLUDE)
  CMakeLists.txt:226 (KOKKOS_SETUP_BUILD_ENVIRONMENT)

PR #7088, opened 2024-06-21
Fix SpaceAwareAccessor based on usage experiment in View

crtrott:
- fix offset type of SpaceAwareAccesor
- Improved test to catch the mistake, and added a typedef for nested_accessor which I believe every wrapping accessor should have
- fix convertibility properties to match View convertibility

### DIFF
--- a/core/src/View/MDSpan/Kokkos_MDSpan_Accessor.hpp
+++ b/core/src/View/MDSpan/Kokkos_MDSpan_Accessor.hpp
@@ -50,21 +50,14 @@ struct SpaceAwareAccessor {
   KOKKOS_DEFAULTED_FUNCTION
   constexpr SpaceAwareAccessor() = default;
 
-  template <class OtherNestedAccessorType,
-            std::enable_if_t<std::is_constructible_v<NestedAccessor,
-                                                     OtherNestedAccessorType>,
-                             int> = 0>
+  template <
+      class OtherMemorySpace, class OtherNestedAccessorType,
+      std::enable_if_t<
+          MemorySpaceAccess<MemorySpace, OtherMemorySpace>::assignable &&
+              std::is_constructible_v<NestedAccessor, OtherNestedAccessorType>,
+          int> = 0>
   KOKKOS_FUNCTION constexpr SpaceAwareAccessor(
-      const SpaceAwareAccessor<MemorySpace, OtherNestedAccessorType>&
-          other) noexcept
-      : nested_acc(other.nested_acc) {}
-
-  template <class OtherNestedAccessorType,
-            std::enable_if_t<std::is_constructible_v<NestedAccessor,
-                                                     OtherNestedAccessorType>,
-                             int> = 0>
-  KOKKOS_FUNCTION constexpr SpaceAwareAccessor(
-      const SpaceAwareAccessor<AnonymousSpace, OtherNestedAccessorType>&
+      const SpaceAwareAccessor<OtherMemorySpace, OtherNestedAccessorType>&
           other) noexcept
       : nested_acc(other.nested_acc) {}
 
--- a/core/src/View/MDSpan/Kokkos_MDSpan_Accessor.hpp
+++ b/core/src/View/MDSpan/Kokkos_MDSpan_Accessor.hpp
@@ -76,8 +76,17 @@ struct SpaceAwareAccessor {
   }
 
   KOKKOS_FUNCTION
-  constexpr auto offset(data_handle_type p, size_t i) const noexcept {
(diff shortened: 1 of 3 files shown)


core/src/View/MDSpan/Kokkos_MDSpan_Accessor.hpp
```diff
@@ -81,8 +83,7 @@ struct SpaceAwareAccessor {
   }
 
   KOKKOS_FUNCTION
-  constexpr data_handle_type offset(data_handle_type p, size_t i) const
-      noexcept {
+  constexpr auto offset(data_handle_type p, size_t i) const noexcept {
```
masterleinad, line 86: The return type should really be `typename offset_policy​::​data_handle_type` (according to https://eel.is/c++draft/mdspan.accessor.reqmts#9), right?

nliber (approved):

core/src/View/MDSpan/Kokkos_MDSpan_Accessor.hpp
crtrott: jup but that is the same as nested_acc.offset return type

core/src/View/MDSpan/Kokkos_MDSpan_Accessor.hpp
```diff
@@ -34,35 +34,30 @@ namespace Impl {
...
-  template <class OtherNestedAccessorType,
-            std::enable_if_t<std::is_constructible_v<NestedAccessor,
-                                                     OtherNestedAccessorType>,
-                             int> = 0>
+  template <
+      class OtherMemorySpace, class OtherNestedAccessorType,
+      std::enable_if_t<
+          MemorySpaceAccess<MemorySpace, OtherMemorySpace>::assignable &&
+              std::is_constructible_v<NestedAccessor, OtherNestedAccessorType>,
+          int> = 0>
```
dalg24, line 58: This change does not seem to belong to this PR which the description advertise as being about fixing the return type of the `SpaceAwareAccessor::offset` member function.
Please withdraw it or make the case that it belongs here and update the title and description.

core/src/View/MDSpan/Kokkos_MDSpan_Accessor.hpp
```diff
@@ -34,35 +34,30 @@ namespace Impl {
...
   using reference        = typename NestedAccessor::reference;
   using data_handle_type = typename NestedAccessor::data_handle_type;
-
   using offset_policy =
       SpaceAwareAccessor<MemorySpace, typename NestedAccessor::offset_policy>;
 
-  using memory_space = MemorySpace;
+  // Specific to SpaceAwareAccessor
+  using memory_space         = MemorySpace;
+  using nested_accessor_type = NestedAccessor;
```
dalg24, line 46: I don't necessary have an issue with adding that public member type but you don't say we might want to have it and it is not used.

core/src/View/MDSpan/Kokkos_MDSpan_Accessor.hpp
masterleinad: I would prefer spelling out the return type here instead of using `auto`.

core/src/View/MDSpan/Kokkos_MDSpan_Accessor.hpp
crtrott: I updated title and description

core/src/View/MDSpan/Kokkos_MDSpan_Accessor.hpp
crtrott: done

core/src/View/MDSpan/Kokkos_MDSpan_Accessor.hpp
crtrott: I still think we should have it, and I did add the nested_accessor() function as in https://eel.is/c++draft/linalg.scaled

masterleinad (approved):

core/src/View/MDSpan/Kokkos_MDSpan_Accessor.hpp
dalg24: `std::linalg::scaled_accessor` does not define a member type for it.  Anyway, I don't feel too strongly about it.

crtrott: Yeah I know, the comma in my reply was basically saying "I do think we should have the typedef" and orthogonal to that "I also added the nested_accessor() function as in scaled_accessor"

PR #7091, opened 2024-06-21
Remove cmake options to change default spaces at configure time

JBludau:
It looks like these options are somewhat broken, as both `KOKKOS_DEVICE_SPACE_ACTIVE` and `KOKKOS_HOSTPARALLEL_SPACE_ACTIVE` seem to be unused.

Furthermore, it seems like the cmake options to change the default execution and memory space are undocumented.


masterleinad:
This was introduced in https://github.com/kokkos/kokkos/pull/2941.

dalg24 (approved):
That code dates back from the refactor in #2941
It is not clear to me how that was supposed to work.  I am ok with removing it.

aprokop (approved):

nliber (approved):

JBludau:
Some builds are failing but it seems unrelated. But a qualified second opinion would be good, as I have not followed CI problems recently

dalg24:
CI failures unrelated
```
3: Failing in Thread:1
3:  File: (OpenACC API)
3:  Function: acc_copyout:1
3:  Line
```
and all unit test segfaulting in one of the HIP builds

PR #7096, opened 2024-06-22
Cleanup KokkosP hooks in `Profiling::`

dalg24:
Prefer using-declarations to reduce boiler plate code.
In the process I realized that the definitions of `Profiling::{print_help, parse_args}` had no matching declarations.
I elected to add them as it looks as an oversight, let me know if you prefer to omit them.


aprokop (approved):
LGTM

masterleinad (approved):

crtrott (approved):

PR #7098, opened 2024-06-24
Copy print_configuration when combining two kokkos settings.

DerNils-git:
[Combining](https://github.com/kokkos/kokkos/blob/develop/core/src/impl/Kokkos_Core.cpp#L83) two `Kokkos::InitializationSettings` skipped the `print_configuration` setting. 

We encountered this bug when initializing `Kokkos::ScopeGuard` using `Kokkos::InitializationSettings` with `set_print_configuration(true)`.
It also seems to occur when using `Kokkos::initialize` with the above setting.

According to the documentation setting this variable should be a valid member of the initialization setting.
https://kokkos.org/kokkos-core-wiki/API/core/initialize_finalize/InitializationSettings.html


dalg24 (approved):
Thanks for reporting and fixing.
In case it's useful for other reviewers, the complete list of initialization settings is here.

masterleinad (approved):

Issue #7084, opened 2024-06-20
Nightly build errors, intel (icpc) using "-no-ip" flag


ndellingwood:
**Describe the bug**

Nightly builds (particularly Trilinos) with intel classic compilers and using the "-no-ip" flag to disable IPO fail to compile with internal errors with the TestSerial_ViewCopy_c.cpp test:

```
": internal error: 101003_1112

compilation aborted for /home/ndellin/kokkos/testing/SemsIntel2021_3/core/unit_test/serial/TestSerial_ViewCopy_c.cpp (code 4)
make[2]: *** [core/unit_test/CMakeFiles/Kokkos_CoreUnitTest_Serial2.dir/build.make:345: core/unit_test/CMakeFiles/Kokkos_CoreUnitTest_Serial2.dir/serial/TestSerial_ViewCopy_c.cpp.o] Error 4
make[2]: Leaving directory '/home/ndellin/kokkos/testing/SemsIntel2021_3'
make[1]: *** [CMakeFiles/Makefile2:1501: core/unit_test/CMakeFiles/Kokkos_CoreUnitTest_Serial2.dir/all] Error 2
```

The error does not occur so far in builds where "-no-ip" is not used.

This is a bit of a special case that has been used as a workaround of internal errors for builds of Trilinos when using intel icpc compilers and setting architecture flags (I can search back through PRs / issues if needed to find where this was discussed)

This is the list of commits since the previous night's passing builds:

**Changes**

    Add SpaceAwareAccessor ([details](https://jenkins-son.sandia.gov/view/Kokkos/job/Kokkos_develop_blake_intel_2023_2_0-icpc_openmp_serial_cpp17/19/changes#afbff6c532aef15665636073251328f37c354f51) / [githubweb](https://github.com/kokkos/kokkos/commit/afbff6c532aef15665636073251328f37c354f51))
    Add basic tests for SpaceAwareAccessor ([details](https://jenkins-son.sandia.gov/view/Kokkos/job/Kokkos_develop_blake_intel_2023_2_0-icpc_openmp_serial_cpp17/19/changes#b2046a40e4c9d109b2004abec91d542998a03859) / [githubweb](https://github.com/kokkos/kokkos/commit/b2046a40e4c9d109b2004abec91d542998a03859))
[...]

PR #7106, opened 2024-06-28
Workaround to ice with icpc when using -no-ip

ndellingwood:
Remove the ViewCopy_c test when compiling with icpc See issue: https://github.com/kokkos/kokkos/issues/7084

### DIFF
--- a/core/unit_test/CMakeLists.txt
+++ b/core/unit_test/CMakeLists.txt
@@ -270,15 +270,20 @@ foreach(Tag Threads;Serial;OpenMP;Cuda;HPX;OpenMPTarget;OpenACC;HIP;SYCL)
       WorkGraph
       WithoutInitializing
       )
-      set(file ${dir}/Test${Tag}_${Name}.cpp)
-      # Write to a temporary intermediate file and call configure_file to avoid
-      # updating timestamps triggering unnecessary rebuilds on subsequent cmake runs.
-      file(WRITE ${dir}/dummy.cpp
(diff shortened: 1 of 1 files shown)


ndellingwood:
I was able compile several Trilinos packages that have been sensitive to ICE without the kokkos tests; removing this test with intel classic compilers is sufficient to allow the rest of the Kokkos tests to compile and avoids wasting time on clunky workarounds for a deprecated compiler

crtrott (changes requested):
can't we just add a "remove from list thing"?

ndellingwood:
> can't we just add a "remove from list thing"?

@crtrott the change I added skips `ViewCopy_c` in the for loop when intel compilers are used; I can try removing the generated file from the list L285 `list(APPEND ${Tag}_SOURCES2A ${file})`, but will that leave behind some generated file cruft from the configuration file write/configuration L280-284?

crtrott:
One sec I am pushing an alternative.

crtrott:
Can you try this? It mirrors how we do things for the sources 1B

ndellingwood:
@crtrott yeah I like that better, I'll test it out

ndellingwood:
@crtrott that worked, thanks!

crtrott (approved):

nmm0 (approved):
Looks good to me though I'm really curious as to what exactly is causing the ICE

crtrott:
Bunch of issues with container bringup. I will merge this.

PR #7107, opened 2024-06-28
Adapt jenkins CI to hopper node

masterleinad:
We are seeing
```
nvcc fatal   : Value 'sm_90' is not defined for option 'gpu-architecture'
```
in some builds so we have to restrict existing builds using `Cuda` version earlier than 11.8 (all of them) not to use the `hopper` runner. At the same time, we are adding a build to test the `hopper` runner.


dalg24:
Please discuss node availability considerations.
As far as I know we have preview access, no blanket promise about long-term availability of the GPU.
Also, restricting a build to run on a single agent tends to make the overall Jenkins job more brittle.

It doesn't mean we are not allowed to explicitly target the H100 GPU but we need to be careful doing so.
Please explain why you think that is fine.

masterleinad:
> Please discuss node availability considerations.
As far as I know we have preview access, no blanket promise about long-term availability of the GPU.
Also, restricting a build to run on a single agent tends to make the overall Jenkins job more brittle.

> It doesn't mean we are not allowed to explicitly target the H100 GPU but we need to be careful doing so.
Please explain why you think that is fine.

At this point, I just want to make sure that we can actually use the runner and it turns out that we need at least cuda-11.8 for H100. Thus, we need to update the label restrictions for the jobs that were only asking for any kind of Nvidia GPU (since they were using Cuda before 11.8). We can discuss if we want to keep the new build and restricting it to the `hopper` node later.

masterleinad:
Let's close until we have (permanent? reliable?) access to that node.

PR #7109, opened 2024-07-02
simd: add initial support for SVE

ldh4:
WIP: testing, refactoring, some consolidations

This PR adds initial kokkos simd types for sve intrinsic, initially targeting for A64fx with sve 1.0 (and not sve 2).
Uses vector length specific (VLS) instead of vector length agnostic (VLA).

Assumes 512 bits sve vector length based on specification in kokkos cmake file:

https://github.com/kokkos/kokkos/blob/d9f7dfe85c4bf8ff5a0360ac7144ea50de680984/cmake/kokkos_arch.cmake#L292-L297


diehlpk:
I have access to A64FX nodes and will test the pull request.

PR #7113, opened 2024-07-03
For the HIP C++20 build use libc++

crtrott:


crtrott:
@masterleinad @rombur looks like the ROCM install comes without libc++: 
```c++
/opt/rocm-5.6.0/llvm/lib/clang/16.0.0/include/cuda_wrappers/cmath:27:15: fatal error: 'cmath' file not found
    #include_next <cmath>
```

Rombur:
@crtrott  You need to install it in `Dockerfile.hipcc`

crtrott:
Doesn't

PR #7114, opened 2024-07-03
Add missing concepts include in test

crtrott:
Note this was implicitly likely there when mdspan is enabled.

### DIFF
--- a/core/unit_test/TestSpaceAwareAccessor.hpp
+++ b/core/unit_test/TestSpaceAwareAccessor.hpp
@@ -18,6 +18,9 @@
 #include <type_traits>
 
 #include <gtest/gtest.h>
+#ifndef KOKKOS_ENABLE_CXX17
+#include <concepts>
+#endif
 
 template <class T>
 struct funky_data_handle {
@@ -91,8 +94,7 @@ void test_space_aware_accessor() {
             std::is_same_v<typename acc_t::memory_space, memory_space_t>);
         static_assert(
             std::is_same_v<typename acc_t::nested_accessor_type, FunkyAcc<T>>);
-#if defined(KOKKOS_ENABLE_CXX20) || defined(KOKKOS_ENABLE_CXX23) || \
-    defined(KOKKOS_ENABLE_CXX26)
+#ifndef KOKKOS_ENABLE_CXX17
         static_assert(std::copyable<acc_t>);
         static_assert(std::is_empty_v<acc_t>);
 #endif


masterleinad (approved):

dalg24 (approved):

masterleinad (commented):
This will not be sufficient for some compilers. What about we explicitly check for `concepts` support in this release and guard accordingly and enforce `concepts` support for C++20 and beyond after the next release?

crtrott:
What do you mean sufficient for some compilers? Basically what I am saying is: if you try to enable C++20 for Kokkos with a compiler that doesn't have concepts its on you and I am ok with the build failing. We do NOT want to get into the business of checking for every feature in the standard individually.

masterleinad:
> What do you mean sufficient for some compilers? Basically what I am saying is: if you try to enable C++20 for Kokkos with a compiler that doesn't have concepts its on you and I am ok with the build failing. We do NOT want to get into the business of checking for every feature in the standard individually.

Then we should be consistent (see https://github.com/kokkos/kokkos/blob/83f975a9486ee9432c560c31a62566bf09752d67/algorithms/src/std_algorithms/impl/Kokkos_RandomAccessIterator.hpp#L62-L92) and probably already fail early in the CMake configuration.

masterleinad:
Something like
```diff
diff --git a/cmake/kokkos_test_cxx_std.cmake b/cmake/kokkos_test_cxx_std.cmake
index b075a3e36..f638aad4e 100644
--- a/cmake/kokkos_test_cxx_std.cmake
+++ b/cmake/kokkos_test_cxx_std.cmake
@@ -66,20 +66,27 @@ IF(KOKKOS_CXX_STANDARD STREQUAL "17")
   kokkos_set_cxx_standard_feature(17)
   SET(KOKKOS_CXX_INTERMEDIATE_STANDARD "1Z")
   SET(KOKKOS_ENABLE_CXX17 ON)
-ELSEIF(KOKKOS_CXX_STANDARD STREQUAL "20")
-  kokkos_set_cxx_standard_feature(20)
-  SET(KOKKOS_CXX_INTERMEDIATE_STANDARD "2A")
-  SET(KOKKOS_ENABLE_CXX20 ON)
-ELSEIF(KOKKOS_CXX_STANDARD STREQUAL "23")
-  kokkos_set_cxx_standard_feature(23)
-  SET(KOKKOS_CXX_INTERMEDIATE_STANDARD "2B")
-  SET(KOKKOS_ENABLE_CXX23 ON)
-ELSEIF(KOKKOS_CXX_STANDARD STREQUAL "26")
-  kokkos_set_cxx_standard_feature(26)
-  SET(KOKKOS_CXX_INTERMEDIATE_STANDARD "2C")
-  SET(KOKKOS_ENABLE_CXX26 ON)
 ELSE()
-  MESSAGE(FATAL_ERROR "Kokkos requires C++17 or newer but requested ${KOKKOS_CXX_STANDARD}!")
+  IF(KOKKOS_CXX_STANDARD STREQUAL "20")
+    kokkos_set_cxx_standard_feature(20)
+    SET(KOKKOS_CXX_INTERMEDIATE_STANDARD "2A")
+    SET(KOKKOS_ENABLE_CXX20 ON)
+  ELSEIF(KOKKOS_CXX_STANDARD STREQUAL "23")
+    kokkos_set_cxx_standard_feature(23)
+    SET(KOKKOS_CXX_INTERMEDIATE_STANDARD "2B")
+    SET(KOKKOS_ENABLE_CXX23 ON)
+  ELSEIF(KOKKOS_CXX_STANDARD STREQUAL "26")
+    kokkos_set_cxx_standard_feature(26)
+    SET(KOKKOS_CXX_INTERMEDIATE_STANDARD "2C")
+    SET(KOKKOS_ENABLE_CXX26 ON)
+  ELSE()
+    MESSAGE(FATAL_ERROR "Kokkos requires C++17 or newer but requested ${KOKKOS_CXX_STANDARD}!")
+  ENDIF()
+  INCLUDE(CheckIncludeFileCXX)
+  CHECK_INCLUDE_FILE_CXX("<concepts>" KOKKOS_COMPILER_HAS_CONCEPTS)
+  IF (NOT KOKKOS_COMPILER_HAS_CONCEPTS)
+    MESSAGE(FATAL_ERROR "C++20 or later requested but the compiler doesn't support concepts!")
+  ENDIF()
 ENDIF()
 
 # Enforce that we can compile a simple C++17 program
```

crtrott:
I think we should back out that thing in algorithms and simply enable it if C++20 is enabled. I do not think we should check for everything in cmake. I mean do you really want to add: check for concepts, check for conditional explicit, check for if consteval, check for elifdef check for span, etc.? I am pretty strongly on the no side for this. If we want to exit early: I think we can simply already define minimum compiler versions for C++20 now.

crtrott:
Did the right thing: error message is now about <concepts> not found not std::copyable.

PR #2707, opened 2020-01-31
Remove uvm allocation limit

masterleinad:
Fixes #2641.

PR #7120, opened 2024-07-07
Deprecate `RawMemoryAllocationFailure::FailureMode::MaximumCudaUVMAllocationsExceeded`

dalg24:
Oversight in #2707
It is in namespace `Experimental::` so we could just remove it but I suggest we mark it as deprecated to be safe.


dalg24:
> It is in namespace `Experimental::` so we could just remove it but I suggest we mark it as deprecated to be safe.

I had not anticipated we'd need a workaround in the switch statement.  Let me know if you'd prefer to just drop the enumerator.  I think that is fine.

cedricchevalier19:
I think it is simpler to drop it.

It is funny to have a user-facing data structure in the `impl` directory!

dalg24:
> I think it is simpler to drop it.

Done

cedricchevalier19 (approved):

PR #7123, opened 2024-07-08
Check for LIBCXX 10 or later for C++20 and later

masterleinad:
In response to https://github.com/kokkos/kokkos/pull/7114#issuecomment-2206731884.
We are already checking for ` __GNUC__` and Intel compiler versions in this file.

`LIBCXX_RELEASE`  might be defined for compilers other than `g++` that are using `libstdc++`. Version 10 is sufficient both for conditional explicit and concepts.

### DIFF
--- a/algorithms/src/std_algorithms/impl/Kokkos_RandomAccessIterator.hpp
+++ b/algorithms/src/std_algorithms/impl/Kokkos_RandomAccessIterator.hpp
@@ -59,15 +59,7 @@ class RandomAccessIterator< ::Kokkos::View<DataType, Args...> > {
                                                 ptrdiff_t current_index)
       : m_view(view), m_current_index(current_index) {}
 
-// FIXME The C++20 requires expression is not supported with Clang 9 and GCC 9
-// The following guards is unsufficient until we increase our minimum CXX20
-// compiler requirements.
-//   #ifndef KOKKOS_ENABLE_CXX17  // C++20 and beyond
-// We replace the Kokkos guards with standard C++ feature testing in the
-// meantime.
-#if (defined(__cpp_concepts) && (__cpp_concepts >= 201907L)) && \
-    (defined(__cpp_conditional_explicit) &&                     \
-     (__cpp_conditional_explicit >= 201806L))
+#ifndef KOKKOS_ENABLE_CXX17  // C++20 and beyond
   template <class OtherViewType>
   requires(std::is_constructible_v<view_type, OtherViewType>) KOKKOS_FUNCTION
       explicit(!std::is_convertible_v<OtherViewType, view_type>)
--- a/core/src/Kokkos_Macros.hpp
+++ b/core/src/Kokkos_Macros.hpp
@@ -58,15 +58,16 @@
 #include <impl/Kokkos_NvidiaGpuArchitectures.hpp>
 #endif
 
+#if !defined(KOKKOS_ENABLE_CXX17)
 #if __has_include(<version>)
 #include <version>
 #else
 #include <iso646.h>
 #endif
-#if !defined(KOKKOS_ENABLE_CXX17) && defined(_GLIBCXX_RELEASE) && \
-    _GLIBCXX_RELEASE < 10
+#if defined(_GLIBCXX_RELEASE) && _GLIBCXX_RELEASE < 10
 #error \
     "Compiling with support for C++20 or later requires a libstdc++ version later than 9"
+#endif
 #endif
 
 //----------------------------------------------------------------------------
--- a/core/src/Kokkos_Macros.hpp
+++ b/core/src/Kokkos_Macros.hpp
@@ -62,7 +62,7 @@
 #if __has_include(<version>)
 #include <version>
 #else
-#include <iso646.h>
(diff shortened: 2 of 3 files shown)


algorithms/src/std_algorithms/impl/Kokkos_RandomAccessIterator.hpp
```diff
@@ -59,15 +59,7 @@ class RandomAccessIterator< ::Kokkos::View<DataType, Args...> > {
...
-// FIXME The C++20 requires expression is not supported with Clang 9 and GCC 9
-// The following guards is unsufficient until we increase our minimum CXX20
-// compiler requirements.
-//   #ifndef KOKKOS_ENABLE_CXX17  // C++20 and beyond
-// We replace the Kokkos guards with standard C++ feature testing in the
-// meantime.
-#if (defined(__cpp_concepts) && (__cpp_concepts >= 201907L)) && \
-    (defined(__cpp_conditional_explicit) &&                     \
-     (__cpp_conditional_explicit >= 201806L))
+#ifndef KOKKOS_ENABLE_CXX17  // C++20 and beyond
```
cedricchevalier19, line 62: Checking C++20 features with `CXX17` in the macro name is strange.

core/src/Kokkos_Macros.hpp
```diff
@@ -58,6 +58,13 @@
 #include <impl/Kokkos_NvidiaGpuArchitectures.hpp>
 #endif
 
+#include <ciso646>
```
cedricchevalier19, line 61: `<ciso646>` does not exist anymore in C++20 (https://en.cppreference.com/w/cpp/header/ciso646) . Is it ok to unconditionally include it?

cedricchevalier19 (commented):
I can't help to think that library requirements should be checked by cmake.
But I understand that you do not want duplicate things for the makefiles.

core/src/Kokkos_Macros.hpp
masterleinad:
```suggestion
#include <iso646.h>
```

algorithms/src/std_algorithms/impl/Kokkos_RandomAccessIterator.hpp
masterleinad: That's what we are doing whenever we are checking for C++20 and later.

crtrott:
> I can't help to think that library requirements should be checked by cmake. But I understand that you do not want duplicate things for the makefiles.

By checking in the header we also catch the situation where you installed Kokkos with a valid C++ stdlib but then compiled an application with an invalid one (i.e. you forgot to add the -gcc-toolchain to your compiler flags, or only loaded the intel module without loading also a new enough GCC etc.)

core/src/Kokkos_Macros.hpp
```diff
@@ -58,6 +58,13 @@
 #include <impl/Kokkos_NvidiaGpuArchitectures.hpp>
 #endif
 
+#include <iso646.h>
```
masterleinad, line 61: The C++20 way would be using `version` but that might not exist for problematic toolchains.

dalg24: Any good reason to prefer the version you proposed to
```C++
#ifndef KOKKOS_ENABLE_CXX17
#include <version>
#if defined(_GLIBCXX_RELEASE) && _GLIBCXX_RELEASE < 10
#error \
    "Compiling with support for C++20 or later requires a libstdc++ version later than 9"
#endif
#endif

Comment (masterleinad) [reply core/src/Kokkos_Macros.hpp:61]: `version` might not exist and we would get a better error message with the current code. I'm fine with splitting the if in two and only conditionally include `iso646.h`.

Comment (dalg24) [reply core/src/Kokkos_Macros.hpp:61]:
```C++
#if __has_include(<version>)
#include <version>
#else
#include <ciso646>
#endif
```
then?

masterleinad: Sure, just in case there is a case where `iso646.h` doesn't define `_GLIBCXX_RELEASE` but `<version>` exists.

core/src/Kokkos_Macros.hpp
```diff
@@ -58,6 +58,18 @@
 #include <impl/Kokkos_NvidiaGpuArchitectures.hpp>
 #endif
 
+#if !defined(KOKKOS_ENABLE_CXX17)
+#if __has_include(<version>)
+#include <version>
+#else
+#include <iso646.h>
```
dalg24, line 65:
```suggestion
#include <ciso646>
```

masterleinad: https://en.cppreference.com/w/cpp/header/ciso646 says:
> [...] \<ciso646> is removed in C++20. Corresponding <iso646.h> is still available in C++20. [...]

so it seems safer to use `iso646.h`.

dalg24: The whole thing is guarded for C++20 enabled

dalg24 (approved):

nliber (approved):

PR #7130, opened 2024-07-11
Allow using multiple HIP devices

tcclevenger:
Allow HIP to use multiple devices. Here we follow the implementation in Cuda.

Performance: Testing a HIP version of the multigpu tutorial (computing yAx with 2 different sets of y,A,x, first both sets on a single device, second on 2 different devices).
```
MI250, rocm 5.2

1 device:
  N( 10000 ) nrepeat ( 100 ) problem( 1600.32 MB ) time( 1.21464 s ) bandwidth( 263.504 GB/s )

2 device:
  N( 10000 ) nrepeat ( 100 ) problem( 1600.32 MB ) time( 0.610534 s ) bandwidth( 524.236 GB/s )

```

I can split each of these commits into separate PRs if that is preferred.


core/src/HIP/Kokkos_HIP.hpp
```diff
@@ -52,6 +52,9 @@ class HIP {
       Impl::ManageStream manage_stream = Impl::ManageStream::no);
   KOKKOS_DEPRECATED HIP(hipStream_t stream, bool manage_stream);
 
+  HIP(const int hip_device, hipStream_t stream,
+      Impl::ManageStream manage_stream = Impl::ManageStream::no);
```
tcclevenger, line 56: For anything before rocm5.6, we do not have access to a function which gets a device id from a stream (that I am aware of), so we need a constructor to take a device id.

masterleinad: What about we only provide multi-GPU support for `ROCm` 5.6 and later then?

tcclevenger: I would be ok with that.

tcclevenger:
I changed to only querying device from stream in the case of rocm5.6+. If rocm5.5- then we just use the singleton device.

The negative for me is that there is no way for a user to know they can't use multigpu (outside of comments or documentation, which will need to be added) since it's just a `stream` they are creating which gets paired to a device id that might be incorrect. But I guess this is the case right now with the stream constructor.

Another option is to make multi GPU less of an expert method and provide a constructor
```
HIP(dev_id)
```
which internally creates a stream that is automatically managed. Then multi GPU could be used without the user ever creating a stream, and it would work for all rocm5.2+.

tcclevenger:
```
MI250, rocm 5.6
  - 1 device
      N( 10000 ) nrepeat ( 100 ) problem( 1600.32 MB ) time( 1.19394 s ) bandwidth( 268.073 GB/s )
  - 2 device
      N( 10000 ) nrepeat ( 100 ) problem( 1600.32 MB ) time( 0.612265 s ) bandwidth( 522.754 GB/s )


```

core/src/HIP/Kokkos_HIP_Instance.cpp
```diff
@@ -170,8 +170,31 @@ void HIPInternal::initialize(hipStream_t stream) {
...
+
   m_stream = stream;
 
+  // Allocate a staging buffer for constant mem in pinned host memory
+  // and an event to avoid overwriting driver for previous kernel launches
+  if (not constantMemHostStaging[m_hipDev])
+    KOKKOS_IMPL_HIP_SAFE_CALL((hip_host_malloc_wrapper(
+        reinterpret_cast<void **>(&constantMemHostStaging[m_hipDev]),
+        Impl::HIPTraits::ConstantMemoryUsage)));
+  if (not constantMemReusable[m_hipDev])
```
cwpearson, line 194: `if (!...` (as elsewhere in Kokkos)

core/src/HIP/Kokkos_HIP_Instance.cpp
```diff
@@ -170,8 +170,31 @@ void HIPInternal::initialize(hipStream_t stream) {
...
+  m_hipDev = singleton().m_hipDev;
+#endif
+  KOKKOS_IMPL_HIP_SAFE_CALL(hipSetDevice(m_hipDev));
+  hip_devices.insert(m_hipDev);
+
   m_stream = stream;
 
+  // Allocate a staging buffer for constant mem in pinned host memory
+  // and an event to avoid overwriting driver for previous kernel launches
+  if (not constantMemHostStaging[m_hipDev])
```
cwpearson, line 190: `if (!...` (as elsewhere in Kokkos)

cwpearson (changes requested):

core/unit_test/CMakeLists.txt
```diff
@@ -844,6 +844,14 @@ if(Kokkos_ENABLE_HIP)
       UnitTestMain.cpp
       hip/TestHIP_InterOp_Streams.cpp
   )
+  if(KOKKOS_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 5.6)
```
romintomasetti, line 847: It seems usually the following is done:
```cmake
if(KOKKOS_CXX_COMPILER_ID STREQUAL "<name-of-hipcc-I-guess" AND KOKKOS_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 5.6)
```

romintomasetti: It would also help people reasoning about your condition and understand it is for `hipcc`.

masterleinad: This is more aboutthe ROCm version than the compiler.

core/src/HIP/Kokkos_HIP_ZeroMemset.hpp
```diff
@@ -26,9 +26,10 @@ namespace Impl {
...
   ZeroMemset(const HIP& exec_space, const View<T, P...>& dst) {
-    KOKKOS_IMPL_HIP_SAFE_CALL(hipMemsetAsync(
-        dst.data(), 0, dst.size() * sizeof(typename View<T, P...>::value_type),
-        exec_space.hip_stream()));
+    KOKKOS_IMPL_HIP_SAFE_CALL(
+        (exec_space.impl_internal_space_instance()->hip_memset_async_wrapper(
+            dst.data(), 0,
+            dst.size() * sizeof(typename View<T, P...>::value_type))));
   }
 };
```
romintomasetti, line 34: Please do this in a separate PR, as multi-GPU support is not required for this change.

Also, please consider doing it as @dalg24 did in #7226.

romintomasetti (changes requested):

core/src/HIP/Kokkos_HIP_ZeroMemset.hpp
masterleinad: This pull request needs rebasing anyway.

tcclevenger: This was resolved in the rebase.

core/unit_test/CMakeLists.txt
tcclevenger: nevermind, HIP macros aren't available. Working on a different solution.

tcclevenger: Used the suggestion to match other places using `KOKKOS_CXX_COMPILER_ID STREQUAL HIPCC`.

cwpearson (approved):

core/src/HIP/Kokkos_HIP_Instance.cpp
```diff
@@ -170,8 +170,31 @@ void HIPInternal::initialize(hipStream_t stream) {
   if (was_finalized)
     Kokkos::abort("Calling HIP::initialize after HIP::finalize is illegal\n");
 
+    // Get the device ID. If this is rocm5.6 or later, we can query this from
+    // the provided stream and potentially use multiple GPU devices, if rocm5.5
+    // or earlier, we must use the singleton device id and there are no checks
+    // possible for the device id matching the device the stream was created on.
```
masterleinad, line 176:
```suggestion
    // Get the device ID. If this is ROCm 5.6 or later, we can query this from
    // the provided stream and potentially use multiple GPU devices. For ROCm 5.5
    // or earlier, we must use the singleton device id and there are no checks
    // possible for the device id matching the device the stream was created on.
```

core/src/HIP/Kokkos_HIP_Instance.cpp
```diff
@@ -217,7 +240,7 @@ Kokkos::HIP::size_type *HIPInternal::scratch_space(const std::size_t size) {
 Kokkos::HIP::size_type *HIPInternal::scratch_flags(const std::size_t size) {
   if (verify_is_initialized("scratch_flags") &&
       m_scratchFlagsCount < scratch_count(size)) {
-    Kokkos::HIPSpace mem_space;
+    auto mem_space = Kokkos::HIPSpace::impl_create(m_hipDev);
```
masterleinad, line 243: No stream? You will need that for `hipMallocAsync`, right?

tcclevenger: Right now `hipMallocAsync` isn't used and `HIPSpace` doesn't have a `hipStream m_stream` member variable.

masterleinad: I'm fine with either adding it here or the pull request adding `hipMallocAsync` support.

masterleinad:
You'll have to fix the indentation again.

core/src/HIP/Kokkos_HIP_Instance.cpp
tcclevenger: Looks like this is already in the works https://github.com/kokkos/kokkos/pull/7324. There has been action on that PR, so maybe we wait on it to merge and then I rebase and make my necessary changes.

masterleinad (approved):

tcclevenger:
Converting to draft while I resolve errors after rebase.

tcclevenger:
Updated performance with the changes. 

Testing a HIP version of the multigpu tutorial (computing yAx with 2 different sets of y,A,x, first both sets on a single device, second on 2 different devices).
```
MI250, rocm 5.6
  - 1 device
      N( 10000 ) nrepeat ( 100 ) problem( 1600.32 MB ) time( 1.1884 s ) bandwidth( 269.323 GB/s )
  - 2 device
      N( 10000 ) nrepeat ( 100 ) problem( 1600.32 MB ) time( 0.611997 s ) bandwidth( 522.983 GB/s )

MI250, rocm 6.0
  - 1 device
      N( 10000 ) nrepeat ( 100 ) problem( 1600.32 MB ) time( 1.19467 s ) bandwidth( 267.91 GB/s )
  - 2 device
      N( 10000 ) nrepeat ( 100 ) problem( 1600.32 MB ) time( 0.613716 s ) bandwidth( 521.518 GB/s )
```

core/src/HIP/Kokkos_HIP_Space.cpp
```diff
@@ -101,18 +111,33 @@ void* HIPSpace::impl_allocate(
...
+                              arg_label, arg_alloc_size, arg_logical_size,
+                              arg_handle, true);
+}
+void* HIPSpace::impl_allocate(
+    const char* arg_label, const size_t arg_alloc_size,
+    const size_t arg_logical_size,
+    const Kokkos::Tools::SpaceHandle arg_handle) const {
+  return impl_allocate_common(m_device, m_stream, arg_label, arg_alloc_size,
+                              arg_logical_size, arg_handle, false);
+}
```
masterleinad, line 140: It seems easier to just inline `HIPSpace::impl_allocate` where it's used.

core/src/HIP/Kokkos_HIP_Space.hpp
```diff
@@ -93,10 +101,15 @@ class HIPSpace {
...
 
  private:
-  void* impl_allocate(const hipStream_t stream, const char* arg_label,
+  void* impl_allocate(const HIP& exec_space, const char* arg_label,
                       const size_t arg_alloc_size,
-                      const size_t arg_logical_size,
-                      bool stream_sync_only) const;
+                      const size_t arg_logical_size = 0,
+                      const Kokkos::Tools::SpaceHandle =
+                          Kokkos::Tools::make_space_handle(name())) const;
```
masterleinad, line 108: Why is it better to create the space handle here instead of locally when required? Is this mostly to match the `Cuda` implementation?

tcclevenger: Yeah, I wanted to match Cuda, but I can change that.

core/src/HIP/Kokkos_HIP_Space.cpp
tcclevenger: This again was me matching Cuda, but wouldn't mind changing. I'll have `allocate()` call a single `impl_allocate()` where the signature is
```
impl_allocate(int dev_id, hipStream_t stream,...) {}
```

masterleinad (approved):

masterleinad (approved):

core/src/HIP/Kokkos_HIP_KernelLaunch.hpp
```diff
@@ -402,15 +411,16 @@ struct HIPParallelLaunchKernelInvoker<DriverType, LaunchBounds,
       params.kernelParams   = (void **)args;
       params.extra          = nullptr;
 
-      KOKKOS_IMPL_HIP_SAFE_CALL(hipGraphAddKernelNode(
-          &graph_node, graph, /* dependencies = */ nullptr,
-          /* numDependencies = */ 0, &params));
+      KOKKOS_IMPL_HIP_SAFE_CALL(
+          (hip_instance->hip_graph_add_kernel_node_wrapper(
+              &graph_node, graph, /* dependencies = */ nullptr,
+              /* numDependencies = */ 0, &params)));
```
Rombur, line 417: @romintomasetti do you know if we can have nodes on different GPUs?

core/src/HIP/Kokkos_HIP_Instance.cpp
```diff
@@ -178,8 +178,32 @@ void HIPInternal::initialize(hipStream_t stream) {
   if (was_finalized)
     Kokkos::abort("Calling HIP::initialize after HIP::finalize is illegal\n");
 
+    // Get the device ID. If this is ROCm 5.6 or later, we can query this from
+    // the provided stream and potentially use multiple GPU devices. For
+    // ROCm 5.5 or earlier, we must use the singleton device id and there are no
+    // checks possible for the device id matching the device the stream was
+    // created on.
```
Rombur, line 185: So with ROCm 5.5 and earlier, we just send everything on a single device? Is there a way to warn the user about it?

core/src/HIP/Kokkos_HIP_KernelLaunch.hpp
romintomasetti: `Cuda` supports it and knows apps are using it, for `HIP` I don't know yet.

core/src/HIP/Kokkos_HIP_Instance.cpp
tcclevenger: I believe the answer is no, there is not way to query the device ID from the stream until rocm5.6, so we couldn't tell the user. But currently in develop the user is free to create and pass a stream on the incorrect device without warning.

We could have a `HIP(dev_id)` constructor that takes in a device id and then internally creates and manages a stream, but this is very different to how we do this in Cuda and I think there has been push back on this in the past. But that would work for all rocm we support in Kokkos.

Rombur: This will need to be documented in the [Known issues](https://kokkos.org/kokkos-core-wiki/known-issues.html)

core/unit_test/hip/TestHIP_InterOp_StreamsMultiGPU.cpp
```diff
@@ -0,0 +1,108 @@
...
+  }
+};
+
+std::array<TEST_EXECSPACE, 2> get_execution_spaces(
+    const StreamsAndDevices &streams_and_devices) {
+  TEST_EXECSPACE exec0(streams_and_devices.streams[0]);
+  TEST_EXECSPACE exec1(streams_and_devices.streams[1]);
+
+  // Must return void to use ASSERT_EQ
+  [&]() {
```
Rombur, line 52: Why do you create a lambda function here?

tcclevenger: `ASSERT_EQ` must be enclosed in a function that returns void. I changed to using `EXPECT_EQ`.

core/src/HIP/Kokkos_HIP_Instance.hpp
```diff
@@ -128,6 +131,161 @@ class HIPInternal {
...
+
+  // Set the device in to the device stored by this instance for HIP API calls.
+  void set_hip_device() const {
+    verify_is_initialized("set_hip_device");
+    KOKKOS_IMPL_HIP_SAFE_CALL(hipSetDevice(m_hipDev));
+  }
+
+  // Return this instances stream after setting the correct device
+  hipStream_t get_stream() const {
+    set_hip_device();
```
Rombur, line 152: Why is it necessary to set the device? We return a stored variable.

core/src/HIP/Kokkos_HIP_Instance.hpp
```diff
@@ -128,6 +131,161 @@ class HIPInternal {
...
+    return m_stream;
+  }
+
+  // HIP API wrappers where we set the correct device id before calling the HIP
+  // API functions. For HIP API functions that take a stream, an optional input
+  // stream is available. If no stream is given, the stream for this instance is
+  // used.
+
+  // Helper function for selecting correct input stream
+  hipStream_t get_input_stream(hipStream_t s) const {
```
Rombur, line 162: I don't understand why this function and `get_input_stream` exist. Neither of this function is necessary. They just obfuscate the code.

tcclevenger: This is left over from when we were calling HIP API wrapper function from the singleton, where we might need to pass in a stream. I removed it and we now assume wrapper want to use the internal stream.

core/src/HIP/Kokkos_HIP_Instance.hpp
tcclevenger: If I recall correctly, using the stream in the kernel launcher, e.g., 
```
(base_t::get_kernel_func())<<<grid, block, shmem,
                                  hip_instance->get_stream()>>>();
```
required the device to be set. Let me investigate and verify this.

Rombur: If that's the only place, I would prefer if we do something like:
```c++
hip_instance->set_device();
(base_t::get_kernel_func())<<<grid, block, shmem, hip_instance->m_stream>>>();
```

tcclevenger: For cuda, not setting the correct device causes computation in those kernels to not be done/written on correct device. In HIP there isn't that same issue, at least for the unit test we have. I also did some timings and there seems to be no timing degradation.

But I am still not sure why this isn't an issue for HIP. I know initially running HIP on non-default device but never setting the correct device ID for API calls was not an "issue" in the sense that computation all happened in the correct place, but that the time to execute kernels was very slow. My guess is that HIP allowed each device to access memory on other devices and that was slow?

I'm going to continue to look through the kernel launch impl in Cuda and HIP to try and reason with myself why we would/wouldn't need to set device for Cuda/HIP.

janciesko (approved):
LGTM

core/src/HIP/Kokkos_HIP_Instance.hpp
janciesko: I can live with how this is. Is this still a thing to discuss?

masterleinad:
You'll have to fixe the formatting in the `CMake` files:
```diff
diff --git a/core/unit_test/CMakeLists.txt b/core/unit_test/CMakeLists.txt
index c3fcdc3..93df041 100644
--- a/core/unit_test/CMakeLists.txt
+++ b/core/unit_test/CMakeLists.txt
@@ -801,14 +801,10 @@ if(Kokkos_ENABLE_HIP)
   kokkos_add_executable_and_test(
     CoreUnitTest_HIPInterOpGraph SOURCES UnitTestMainInit.cpp hip/TestHIP_InterOp_Graph.cpp
   )
-  if(KOKKOS_CXX_COMPILER_ID STREQUAL HIPCC AND
-     KOKKOS_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 5.6)
+  if(KOKKOS_CXX_COMPILER_ID STREQUAL HIPCC AND KOKKOS_CXX_COMPILER_VERSION VERSION_GREATER_EQUAL 5.6)
     # MultiGPU only works for rocm 5.6+
-    KOKKOS_ADD_EXECUTABLE_AND_TEST(
-      CoreUnitTest_HIPInterOpStreamsMultiGPU
-      SOURCES
-        UnitTestMainInit.cpp
-        hip/TestHIP_InterOp_StreamsMultiGPU.cpp
+    kokkos_add_executable_and_test(
+      CoreUnitTest_HIPInterOpStreamsMultiGPU SOURCES UnitTestMainInit.cpp hip/TestHIP_InterOp_StreamsMultiGPU.cpp
     )
   endif()
 endif()
```

masterleinad (approved):
Looks OK to me.

core/src/HIP/Kokkos_HIP_Instance.cpp
JBludau: Is there a documentation pr yet?

Rombur (approved):

tcclevenger:
Investigating a fail on MI300A when allocating a managed view on dev1 in the test.

masterleinad:
> Investigating a fail on MI300A when allocating a managed view on dev1 in the test.

Do all the other tests work for you? In particular, the one reakted to https://github.com/kokkos/kokkos/pull/7567? Have you made sure that `HSA_XNACK=1` is set in the environment?

tcclevenger:
Update: Not related to XNACK. We are passing an incorrect dev ID for managed views, working on a solution.

tcclevenger:
Update: Issue (which is only on MI300A) anytime we synchronizing streams. It's complaining that 
```
Memory access fault by GPU node-7 (Agent handle: 0x10f0a60) on address 0x7f4a603db000. Reason: Write access to a read-only page.
```
at  
```
KOKKOS_IMPL_HIP_SAFE_CALL(hipStreamSynchronize(m_stream));
```
in `HIPInternal::fence()`.

This line isn't changed in this PR, but it's clearly related to the streams being allocated on different devices since single GPU tests doesn't fail the same way. 

It fails when fencing `exec_space` on dev 0. And setting the correct devid before sync doesn't solve the issue.

Rombur:
@tcclevenger Did you try with different compiler version? It may be worth to check if ROCm 6.3.2 fixes the issue

tcclevenger:
Update: The `HSA_TOOLS_LIB=/opt/rocm-6.3.2/lib/librocm-debug-agent.so.2 HSA_ENABLE_DEBUG=1 ./exec` tool pointed to the memory access fault being here https://github.com/kokkos/kokkos/blob/44f8e6c3a1f0f49152e9b6bbd81dfb8f5737211a/core/src/HIP/Kokkos_HIP_ZeroMemset.cpp#L32
This was introduced in https://github.com/kokkos/kokkos/commit/bd63122775a1c31dfca14868e456edfbddccb657.

Reverting that commit (calling:
```
    KOKKOS_IMPL_HIP_SAFE_CALL(
        (exec_space.impl_internal_space_instance()->hip_memset_async_wrapper(
            dst, 0, cnt)));
```
instead) "solves" the issue, but I'm guessing only because I'm using rocm>6.2.1. Investigating what in multi gpu impl is causing the issue, must be something with the allocation of `dst`?

masterleinad:
> Investigating what in multi gpu impl is causing the issue, must be something with the allocation of `dst`?

Is the memory allocated using the same stream as the functor uses?

tcclevenger:
@Rombur @masterleinad This is ready for re-review. `HIPSpace::allocate(HIP& exec_space,...)` functions were being guarded out in the case of unified mem, so we were only allocating on default device (using the `HIPSpace::allocate(ExecutionSpace&,...)` overload). This should be ready to go pending CI.

MI300A is having timeouts for `Kokkos_CoreUnitTest_HIP`, but it looks like from other PRs this is unrelated to this PR.

tcclevenger:
@masterleinad From the output I don't think `export HSA_XNACK=1` in the .gitlab script is being applied correctly. We get the warning
```
Kokkos::HIP::initialize WARNING: Could not determine that xnack is enabled.
                                 Kokkos requires xnack to be enabled for
                                 ARCH_AMD_GFX942_APU (MI300A) to access host
                                 allocations from the device. Set HSA_XNACK=1
                                 in your environment and ensure
                                 \"CONFIG_HMM_MIRROR=y\" is in the /boot/config
                                 file and that file is readable
```

masterleinad:
> @masterleinad From the output I don't think `export HSA_XNACK=1` in the .gitlab script is being applied correctly. We get the warning

See https://github.com/kokkos/kokkos/issues/7727. The issues is a file in `/boot`.

masterleinad (approved):
Looks good to me.

Rombur (approved):
Good catch.

core/src/HIP/Kokkos_HIP_Instance.hpp
```diff
@@ -70,7 +74,7 @@ class HIPInternal {
  public:
   using size_type = ::Kokkos::HIP::size_type;
 
-  static int m_hipDev;
+  int m_hipDev = -1;
   static int m_maxThreadsPerSM;
 
   static hipDeviceProp_t m_deviceProp;
```
dalg24, line 80: Just noting that this is still static.
Not blocking considering that the CUDA backend has the same issue but this is something that needs to be looked into.

core/src/HIP/Kokkos_HIP_DeepCopy.cpp
```diff
@@ -41,7 +42,8 @@ void DeepCopyHIP(void* dst, void const* src, size_t n) {
 void DeepCopyAsyncHIP(const HIP& instance, void* dst, void const* src,
                       size_t n) {
   KOKKOS_IMPL_HIP_SAFE_CALL(
-      hipMemcpyAsync(dst, src, n, hipMemcpyDefault, instance.hip_stream()));
+      (instance.impl_internal_space_instance()->hip_memcpy_async_wrapper(
```
dalg24, line 45: Why did we add extra parentheses?

core/src/HIP/Kokkos_HIP_KernelLaunch.hpp
```diff
@@ -188,13 +188,21 @@ struct HIPParallelLaunchKernelFuncData {
     return hip_func_attributes.localSizeBytes;
   }
 
-  static hipFuncAttributes get_hip_func_attributes(void const *kernel_func) {
-    static hipFuncAttributes attr = [=]() {
+  // These functions need to be templated on DriverType and LaunchBounds
+  // so that the static bool is unique for each type combo
+  // KernelFuncPtr does not necessarily contain that type information.
+  static hipFuncAttributes get_hip_func_attributes(const int hip_device,
```
dalg24, line 194: These const-qualifiers on argument passed by value are somewhat pointless.

core/src/HIP/Kokkos_HIP_Space.hpp
```diff
@@ -76,12 +84,12 @@ class HIPSpace {
                  const size_t arg_logical_size = 0) const {
     return allocate(arg_label, arg_alloc_size, arg_logical_size);
   }
-#else
+#endif
```
dalg24, line 87: Please confirm that this change is intentional

masterleinad: Yes, we always need the member function below to allocate on the correct device. The ones abover are only necessary for unified memory.

core/src/HIP/Kokkos_HIP.cpp
```diff
@@ -105,18 +104,10 @@ Kokkos::HIP::initialize WARNING: Could not determine that xnack is enabled.
...
-
-  void* constant_mem_void_ptr = nullptr;
-  KOKKOS_IMPL_HIP_SAFE_CALL(hipHostMalloc(
-      &constant_mem_void_ptr, Impl::HIPTraits::ConstantMemoryUsage));
-  Impl::HIPInternal::constantMemHostStaging =
-      static_cast<unsigned long*>(constant_mem_void_ptr);
-
-  KOKKOS_IMPL_HIP_SAFE_CALL(
-      hipEventCreate(&Impl::HIPInternal::constantMemReusable));
+  // Set singleton device id
```
crtrott, line 107: Not really a singleton, but maybe we can fix that naming issue in a separate PR later (also for CUDA etc.)

dalg24:
Planning to ignore the timeout in the MI300A build on HPSF CI

core/src/HIP/Kokkos_HIP_DeepCopy.cpp
tcclevenger: Actually, I'm not reproducing it... I'm thinking a) I only saw this on the Cuda side of things or b) it was a relic from when I use using a function pointer impl. Let me test removing all of them and I'll push the change.

Rombur:
@tcclevenger please add an entry in the changelog

JBludau:
@tcclevenger can you add documentation on this for the upcoming release?

PR #7131, opened 2024-07-12
Do not bother catching a memory allocation failure exception and rethrow it

dalg24:
If a "bad alloc" exception is thrown when resizing private scratch pads, there is little benefit to catch it and convert it to a `std::runtime_error` exception.
Instead prefer specifying a label when calling MemorySpace::allocate() and leave the exception alone.

This is loosely related to #7037 
Note that we don't bother with this pattern of intercepting allocation failure exception in device backends.

### DIFF
--- a/core/src/OpenMP/Kokkos_OpenMP_Instance.cpp
+++ b/core/src/OpenMP/Kokkos_OpenMP_Instance.cpp
@@ -126,14 +126,7 @@ void OpenMPInternal::resize_thread_data(size_t pool_reduce_bytes,
         space.deallocate(m_pool[rank], old_alloc_bytes);
       }
 
-      void *ptr = nullptr;
-      try {
-        ptr = space.allocate(alloc_bytes);
-      } catch (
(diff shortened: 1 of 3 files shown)


stanmoore1:
@dalg24 is it going to abort now so one could get a core dump (very helpful for debugging)?

dalg24:
Note that there a couple more instances of catch `Experimental::RawMemoryAllocationFailure`, mess with it and rethrow, in `impl/Kokkos_SharedAlloc.hpp` but that these are slightly different and I intend to handle them in future work.

crtrott (approved):

ldh4 (approved):

dalg24:
Ignoring
```
3: [ RUN      ] openacc.exec_space_thread_safety_mdrange_reduce
3: Kokkos::Impl::SharedAllocationRecord '' failed decrement count = 0
 3/20 Test  #3: Kokkos_CoreUnitTest_OpenACC .................Subprocess aborted***Exception:  29.55 sec
```
Everything else passed

PR #7132, opened 2024-07-12
Do not swallow errors when deallocating memory with CUDA

dalg24:
Get rid of
```C++
try {
  /* call CUDA runtime to deallocate and throw if an error code is returned */
} catch(...) { /* gulp */ }
```
This code is in place since the dawn of times and hasn't been challenged until now.

Tip to reviewers: hide whitespaces

### DIFF
--- a/core/src/Cuda/Kokkos_CudaSpace.cpp
+++ b/core/src/Cuda/Kokkos_CudaSpace.cpp
@@ -333,27 +333,24 @@ void CudaSpace::impl_deallocate(
     Kokkos::Profiling::deallocateData(arg_handle, arg_label, arg_alloc_ptr,
                                       reported_size);
   }
-  try {
 #ifndef CUDART_VERSION
 #error CUDART_VERSION undefined!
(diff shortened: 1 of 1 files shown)


cedricchevalier19 (approved):
Thanks for the whitespace tip.

masterleinad (approved):

dalg24:
One of the CUDA build (`'CUDA-11.0-NVCC-RDC`) has unrelated failures
```
      Start  3: Kokkos_CoreUnitTest_Cuda1

3: Test command: /var/jenkins/workspace/Kokkos_PR-7132/build-tests/core/unit_test/Kokkos_CoreUnitTest_Cuda1
3: Test timeout computed to be: 1500
3:   what():  cudaGetDeviceCount(&count) error( cudaErrorNoDevice): no CUDA-capable device is detected /var/jenkins/workspace/Kokkos_PR-7132/core/src/impl/Kokkos_Core.cpp:134
 3/47 Test  #3: Kokkos_CoreUnitTest_Cuda1 ........................Child aborted***Exception:   0.19 sec
```
Everything else passes

PR #7133, opened 2024-07-12
Simplify finalize logic

dalg24:
The original intent was to get rid of an unnecessary try-catch block, arguing that it can be achieved with a `noexcept` specifier and that the custom message is not critical (feel free to speak up if you don't agree).

It turns out the test was disabled so I re-enabled it.  I elected to use raw CMake instead of TriBITS to set it up.  It passes on develop and with the getting-rid the try-catch block.  I am willing to split it up if you want me to.

**edit**
I forgot to mention that I removed the `KOKKOS_ADD_ADVANCED_TEST` wrapper function since I removed the only "use" in the codebase.


core/src/impl/Kokkos_Core.cpp
```diff
@@ -786,34 +786,18 @@ void initialize_internal(const Kokkos::InitializationSettings& settings) {
   post_initialize_internal(settings);
 }
 
-void pre_finalize_internal() {
-  typename decltype(finalize_hooks)::size_type numSuccessfulCalls = 0;
+// declared noexcept such that std::terminate is called if any of the registered
+// function throws
+void call_registered_finalize_hook_functions() noexcept {
   while (!finalize_hooks.empty()) {
-    auto f = finalize_hooks.top();
```
dalg24, line 792: The astute reviewer will note that we were unnecessarily making copies of the registered functions.

masterleinad:
It seems the MSVC CI has trouble dealing with the exception
```
 8/37 Test  #8: Kokkos_CoreUnitTest_PushFinalizeHookTerminateRegex ...***Timeout 1500.03 sec
PASSED: I am the custom std::terminate handler.
      Start  9: Kokkos_CoreUnitTest_PushFinalizeHookTerminateFails
```
for some reason and doesn't escape the while loop.

dalg24:
(I had forgotten to commit a note about why I disabled on Windows)

masterleinad (approved):

nliber (approved):

dalg24:
Ignoring the CUDA failures on one of the builds (clearly unrelated)
```
test 34
      Start 34: Kokkos_Benchmark_Atomic_MinMax

34: Test command: /var/jenkins/workspace/Kokkos_PR-7133/build/core/perf_test/Kokkos_Benchmark_Atomic_MinMax "--benchmark_counters_tabular=true" "--benchmark_out=Kokkos_Benchmark_Atomic_MinMax_2024-07-15_T18-41-49.json"
34: Test timeout computed to be: 1500
34: terminate called after throwing an instance of 'std::runtime_error'
34:   what():  cudaGetDeviceCount(&count) error( cudaErrorNoDevice): no CUDA-capable device is detected /var/jenkins/workspace/Kokkos_PR-7133/core/src/impl/Kokkos_Core.cpp:134
34/60 Test #34: Kokkos_Benchmark_Atomic_MinMax .......................Child aborted***Exception:   0.09 sec
```

PR #7139, opened 2024-07-16
Get rid of `RawMemoryAllocationFailure::AllocationMechanism` and derived backend-specific exceptions

dalg24:
All allocation errors are raised using `Impl::throw_bad_alloc()`.
We drop `RawMemoryAllocationFailure::AllocationMechanism` to get rid of the anti-pattern where core depends on backend-specific details.
I couldn't come up with a good way to split this up.  I plan on cleaning up further the `RawMemoryAllocationFailure` exception class in follow-up work.

### DIFF
--- a/core/src/Cuda/Kokkos_CudaSpace.cpp
+++ b/core/src/Cuda/Kokkos_CudaSpace.cpp
@@ -73,8 +73,8 @@ void throw_cuda_allocation_failure(size_t alloc_size, cudaError_t error_code,
   msg += " returned error code \"";
   msg += cudaGetErrorName(error_code);
   msg += "\"";
-  throw_bad_alloc(alloc_size, std::align_val_t{1}, get_failure_mode(error_code),
-                  std::move(msg));
+  Kokkos::Impl::throw_bad_alloc(alloc_size, std::align_val_t{1},
+                                get_failure_mode(error_code), std::move(msg));
 }
 
 }  // namespace
@@ -221,7 +221,7 @@ void *impl_allocate_common(const int device_id,
     // we should do here since we're turning it into an
     // exception here
     cudaGetLastError();
-    throw_cuda_allocation_failure(arg_alloc_size, errot_code, "cudaMalloc()");
+    throw_cuda_allocation_failure(arg_alloc_size, error_code, "cudaMalloc()");
   }
 
   if (Kokkos::Profiling::profileLibraryLoaded()) {
@@ -275,7 +275,7 @@ void *CudaUVMSpace::impl_allocate(
       // we should do here since we're turning it into an
       // exception here
       cudaGetLastError();
-      throw_cuda_allocation_failure(arg_alloc_size, errot_code,
+      throw_cuda_allocation_failure(arg_alloc_size, error_code,
                                     "cudaMallocManaged()");
     }
 
@@ -317,7 +317,7 @@ void *CudaHostPinnedSpace::impl_allocate(
     // we should do here since we're turning it into an
     // exception here
     cudaGetLastError();
-    throw_cuda_allocation_failure(arg_alloc_size, errot_code,
+    throw_cuda_allocation_failure(arg_alloc_size, error_code,
                                   "cudaHostMalloc()");
   }
   if (Kokkos::Profiling::profileLibraryLoaded()) {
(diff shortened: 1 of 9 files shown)


core/src/SYCL/Kokkos_SYCL_Space.cpp
```diff
@@ -75,17 +89,20 @@ SYCLHostUSMSpace::SYCLHostUSMSpace()
...
 
-  if (hostPtr == nullptr)
-    throw RawMemoryAllocationFailure(
-        arg_alloc_size, 1, RawMemoryAllocationFailure::FailureMode::Unknown,
-        failure_tag);
+  if (hostPtr == nullptr) {
+    using FailureMode =
+        Kokkos::Experimental::RawMemoryAllocationFailure::FailureMode;
+    Kokkos::Impl::throw_bad_alloc(arg_alloc_size, std::align_val_t{1},
+                                  FailureMode::Unknown,
```
dalg24, line 103: I preserved the current behavior but this will need fixing.
Apparently `FailureMode::Unknown` was more intended as an invalid state if we look at the definition of `RawMemoryAllocationFailure::what()`.

core/src/Cuda/Kokkos_CudaSpace.cpp
```diff
@@ -56,6 +56,29 @@ const std::unique_ptr<Kokkos::Cuda> &Kokkos::Impl::cuda_get_deep_copy_space(
...
+    case cudaErrorMemoryAllocation: return FailureMode::OutOfMemoryError;
+    case cudaErrorInvalidValue: return FailureMode::InvalidAllocationSize;
+    default: return FailureMode::Unknown;
+  }
+}
+
+void throw_cuda_allocation_failure(size_t alloc_size, cudaError_t error_code,
+                                   std::string msg) {
+  msg += " returned error code \"";
+  msg += cudaGetErrorName(error_code);
```
masterleinad, line 74: Why don't we map to a std::string?

core/src/Cuda/Kokkos_CudaSpace.cpp
```diff
@@ -56,6 +56,29 @@ const std::unique_ptr<Kokkos::Cuda> &Kokkos::Impl::cuda_get_deep_copy_space(
...
+  msg += "\"";
+  Kokkos::Impl::throw_bad_alloc(alloc_size, std::align_val_t{1},
```
masterleinad, line 76:
```suggestion
  Kokkos::Impl::throw_bad_alloc(alloc_size, std::align_val_t{256},
```
According to https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#device-memory-accesses:
> Any address of a variable residing in global memory or returned by one of the memory allocation routines from the driver or runtime API is always aligned to at least 256 bytes.

core/src/HIP/Kokkos_HIP_Space.cpp
```diff
@@ -43,6 +43,25 @@ namespace {
...
+  }
+}
+
+void throw_hip_allocation_failure(size_t alloc_size, hipError_t error_code,
+                                  std::string msg) {
+  msg += "returned error code \"";
+  msg += hipGetErrorName(error_code);
+  msg += "\"";
+  Kokkos::Impl::throw_bad_alloc(alloc_size, std::align_val_t{1},
+                                get_failure_mode(error_code), std::move(msg));
```
masterleinad, line 62: At least add a FIXME for the alignment? Ultimately, we at least assume an alignment of 64 byes throughout the code base anyway.

masterleinad (commented):
I'd prefer if you added a FIXME to all the places where we are reporting trying to align to 1 byte.

dalg24:
> I'd prefer if you added a FIXME to all the places where we are reporting trying to align to 1 byte.

I plan to propose to abandon the `memory_failure` and the `alignment` arguments in a follow up.
Therefore I would prefer not bothering with the FIXMEs.

The `alignment` is a static property, it's not like a specific call to allocate would specifically fail because of its value.
If we can agree that is not needed, then we are left with "out of memory" and "invalid size".
"Invalid size" sounds more like as an "argument not within domain" precondition violation.
If you push the reasoning, one can even question whether we need more than one type of exception with a message.
I think it is more valuable to consider adding things like attaching the source location and stack trace to the exception than having categories of errors.

masterleinad (approved):

Rombur (approved):

PR #7141, opened 2024-07-17
Fix NVHPC CI

masterleinad:
Follow-up to https://github.com/kokkos/kokkos/pull/6987. Testing with `nvhpc-23.9` the `numeric_traits_quiet_and_signaling_nan` tests are failing for unknown reasons. This looks almost like the implicit `-fp-model=fast` for `icpx` but I couldn't find that the compiler does something like that.

`no_device_stack` doesn't seem to be recognized as suppressable warning anymore.

### DIFF
--- a/.jenkins
+++ b/.jenkins
@@ -82,7 +82,7 @@ pipeline {
                                 -DCMAKE_BUILD_TYPE=RelWithDebInfo \
                                 -DCMAKE_CXX_COMPILER=nvc++ \
                                 -DCMAKE_CXX_STANDARD=17 \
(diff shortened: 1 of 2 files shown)


dalg24:
The OpenACC failure is unrelated
```
3: [ RUN      ] openacc.exec_space_thread_safety_mdrange_reduce
3: Kokkos::Impl::SharedAllocationRecord '' failed decrement count = 0

dalg24 (approved):

aprokop (approved):

PR #7145, opened 2024-07-19
Drop `Experimental::RawMemoryAllocationFailure` and don't catch exceptions to rethrow them in shared alloc

dalg24:
Motivation to get rid of `Experimental::RawMemoryAllocationFailure`:
* backends do not really distinguish failure modes and it is not clear it is useful
* the alignment is not controlled by the user, its value is static, so there is little point in attaching it to the exception
* we were actually catching them to rethrow a regular runtime error
* in the end, the most important thing is to raise an error with a good quality error message

The changes in this PR simplify the bad alloc error reporting.
New error messages are along these lines:
```
Kokkos ERROR: Host memory space failed to allocate 1.678e+07 TiB (label="hello_world_failure").
```

I also got rid of the "catch the exception, append to the error message, and rethrow" pattern in shared alloc because we have all the necessary information when we first throw.

### DIFF
--- a/core/src/Cuda/Kokkos_CudaSpace.cpp
+++ b/core/src/Cuda/Kokkos_CudaSpace.cpp
@@ -58,23 +58,13 @@ const std::unique_ptr<Kokkos::Cuda> &Kokkos::Impl::cuda_get_deep_copy_space(
 
 namespace {
 
-auto get_failure_mode(cudaError_t error_code) {
-  using FailureMode =
-      Kokkos::Experimental::RawMemoryAllocationFailure::FailureMode;
-  switch (error_code) {
-    case cudaErrorMemoryAllocation: return FailureMode::OutOfMemoryError;
-    case cudaErrorInvalidValue: return FailureMode::InvalidAllocationSize;
-    default: return FailureMode::Unknown;
-  }
-}
-
 void throw_cuda_allocation_failure(size_t alloc_size, cudaError_t error_code,
                                    std::string msg) {
   msg += " returned error code \"";
   msg += cudaGetErrorName(error_code);
   msg += "\"";
   Kokkos::Impl::throw_bad_alloc(alloc_size, std::align_val_t{1},
-                                get_failure_mode(error_code), std::move(msg));
+                                std::move(msg));
 }
 
 }  // namespace
--- a/core/src/Cuda/Kokkos_CudaSpace.cpp
+++ b/core/src/Cuda/Kokkos_CudaSpace.cpp
@@ -56,18 +56,6 @@ const std::unique_ptr<Kokkos::Cuda> &Kokkos::Impl::cuda_get_deep_copy_space(
   return space;
 }
 
-namespace {
-
-void throw_cuda_allocation_failure(size_t alloc_size, cudaError_t error_code,
-                                   std::string msg) {
-  msg += " returned error code \"";
-  msg += cudaGetErrorName(error_code);
-  msg += "\"";
-  Kokkos::Impl::throw_bad_alloc(alloc_size, std::move(msg));
-}
-
-}  // namespace
-
 namespace Kokkos {
 namespace Impl {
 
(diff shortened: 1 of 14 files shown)


core/unit_test/TestViewAPI.hpp
```diff
@@ -1571,51 +1571,6 @@ class TestViewAPI {
...
   }
-
-  static void run_test_error() {
-// FIXME_MSVC_WITH_CUDA
-// This test doesn't behave as expected on Windows with CUDA
-#if defined(_WIN32) && defined(KOKKOS_ENABLE_CUDA)
-    if (std::is_same<typename dView1::memory_space,
-                     Kokkos::CudaUVMSpace>::value)
-      return;
-#endif
```
dalg24, line 1582: Noting that this was never true in the current develop but I explicitly instantiated the new test on shared memory spaces so we will see if that is still an issue.

core/src/Cuda/Kokkos_CudaSpace.cpp
```diff
@@ -221,7 +198,7 @@ void *impl_allocate_common(const int device_id,
     // we should do here since we're turning it into an
     // exception here
     cudaGetLastError();
-    throw_cuda_allocation_failure(arg_alloc_size, error_code, "cudaMalloc()");
+    Kokkos::Impl::throw_bad_alloc(arg_handle.name, arg_alloc_size, arg_label);
```
dalg24, line 201: Could also do
```suggestion
    Kokkos::Impl::throw_bad_alloc(Kokkos::CudaSpace::name(), arg_alloc_size, arg_label);
```

core/src/impl/Kokkos_Error.hpp
```diff
@@ -17,99 +17,20 @@
 #ifndef KOKKOS_IMPL_ERROR_HPP
 #define KOKKOS_IMPL_ERROR_HPP
 
-#include <new>  // align_val_t
 #include <string>
-#include <iosfwd>
 #include <Kokkos_Macros.hpp>
```
Rombur, line 21: You only need to include `<string>, the other includes are not necessary in the header file.

dalg24: You're correct, but a lot of headers relied on these transitive includes and I did not really want to fix it here.
Is that ok to leave it as future cleanup work?

Rombur: Yes, we can do it another time.

masterleinad (approved):
Looks good to me.

dalg24:
Ignoring OpenMPTarget failure
```
clang++: error: no library 'libomptarget-nvptx-sm_70.bc' found in the default clang lib directory or in LIBRARY_PATH; use '--libomptarget-nvptx-bc-path' to specify nvptx bitcode library
make[2]: *** [core/src/CMakeFiles/kokkoscore.dir/build.make:104: core/src/CMakeFiles/kokkoscore.dir/impl/Kokkos_Command_Line_Parsing.cpp.o] Error 1
clang++: error: no library 'libomptarget-nvptx-sm_70.bc' found in the default clang lib directory or in LIBRARY_PATH; use '--libomptarget-nvptx-bc-path' to specify nvptx bitcode library
make[2]: *** [core/src/CMakeFiles/kokkoscore.dir/build.make:76: core/src/CMakeFiles/kokkoscore.dir/impl/Kokkos_Abort.cpp.o] Error 1
[  1%] Building CXX object _deps/googlebenchmark-build/src/CMakeFiles/benchmark.dir/benchmark_api_internal.cc.o
clang++: error: no library 'libomptarget-nvptx-sm_70.bc' found in the default clang lib directory or in LIBRARY_PATH; use '--libomptarget-nvptx-bc-path' to specify nvptx bitcode library
[  1%] Built target AlwaysCheckGit
make[2]: *** [core/src/CMakeFiles/kokkoscore.dir/build.make:90: core/src/CMakeFiles/kokkoscore.dir/impl/Kokkos_CPUDiscovery.cpp.o] Error 1
make[1]: *** [CMakeFiles/Makefile2:1561: core/src/CMakeFiles/kokkoscore.dir/all] Error 2

PR #7146, opened 2024-07-22
SYCL: Use sycl::shift_group_[left|right] and sycl::select_from_group

masterleinad:
https://github.com/intel/llvm/pull/13666 removes `shuffle_*` which is replaced by `sycl::shift_group*` and `sycl::select_from_group` to be conformant with SYCL2020, see https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#_shift_left_and_shift_right. Unfortunately, the new functions check that the type is trivially copyable which creates some problems for us since we don't impose that restriction in implementations that use shuffles.
The current approach is to `reinterpret_cast` to some trivially-copyable type. The underlying implementation is using
- __spirv_SubgroupShuffleINTEL
- __spirv_SubgroupShuffleDownINTEL
- __spirv_SubgroupShuffleUpINTEL

### DIFF
--- a/core/src/SYCL/Kokkos_SYCL_WorkgroupReduction.hpp
+++ b/core/src/SYCL/Kokkos_SYCL_WorkgroupReduction.hpp
@@ -21,6 +21,47 @@
 
 namespace Kokkos::Impl::SYCLReduction {
 
+template <int N>
+struct TrivialWrapper {
+  std::byte array[N];
+};
+
+template <typename T>
+T shift_group_left(sycl::sub_group sg, T x,
+                   sycl::sub_group::linear_id_type delta) {
+  if constexpr (std::is_trivially_copyable_v<T>)
+    return sycl::shift_group_left(sg, x, delta);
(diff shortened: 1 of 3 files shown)


nliber (approved):

core/src/SYCL/Kokkos_SYCL_WorkgroupReduction.hpp
```diff
@@ -21,6 +21,47 @@
 
 namespace Kokkos::Impl::SYCLReduction {
 
+template <int N>
+struct TrivialWrapper {
+  std::byte array[N];
+};
+
+template <typename T>
+T shift_group_left(sycl::sub_group sg, T x,
```
dalg24, line 30: Add a comment that this is "shuffle down"
Same for the other functions

dalg24 (commented):
Adding comment is optional.

dalg24 (approved):

PR #7148, opened 2024-07-22
Hidden friend operator== for Kokkos::Array

nliber:
This supercedes #6599 (it incorporates the run time tests that @maartenarnst wrote there).

Adds hidden friends `==` and `!=` to `Kokkos::Array<T, N>` for all `N`, including 0.

Deliberately does not add them to the deprecated forms of `Kokkos::Array`.

Adds a compile time test for equality comparable.

### DIFF
--- a/core/unit_test/TestArray.cpp
+++ b/core/unit_test/TestArray.cpp
@@ -15,6 +15,7 @@
 //@HEADER
 
 #include <Kokkos_Array.hpp>
+#include <Kokkos_DetectionIdiom.hpp>
 
 namespace {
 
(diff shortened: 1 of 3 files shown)


core/unit_test/TestArray.cpp
```diff
@@ -23,6 +24,9 @@ namespace {
 template <typename... Ts>
 KOKKOS_FUNCTION constexpr void maybe_unused(Ts&&...) {}
 
+template <typename T, typename U = T>
+using equality_comparable = decltype(std::declval<T&>() == std::declval<U&>());
```
dalg24, line 28:
```suggestion
using equality_comparable = decltype(std::declval<T const&>() == std::declval<U const&>());
```
(not formatted)

dalg24 (approved):
I would prefer if you added the const qualifiers in your definition of equality_comparable

masterleinad (approved):

Issue #7142, opened 2024-07-17
RangePolicy constructors may accidentally implicitly convert optional last parameter to ChunkSize


nliber:
In Release 4.3, we removed the templated variadic constructors for `RangePolicy` in favor of taking `ChunkSize` as a final parameter, as that was the only use of the variadic constructors.

Because the original `RangePolicy` constructors were templated, that last parameter had to be `ChunkSize` and not just convertible to `ChunkSize`.  Now that it is a non-templated constructor, the constructors that take `ChunkSize` take anything implicitly convertible to `ChunkSize` in that position.  Specifically, because `ChunkSize` is _implicitly_ constructible from an `int`, just providing an `int` in that position will now compile where it wouldn't before, which is not the intention.

dalg24:
I would like us to investigate declaring `ChunkSize` constructor from an integer `explicit` as a possible resolution.

nliber:
Although that is a breaking change (see [https://godbolt.org/z/4bxEvMqv3](https://godbolt.org/z/4bxEvMqv3)), I concur.

PR #7151, opened 2024-07-23
Make struct "ChunkSize" constructor explicit to avoid implicit construction in RangePolicy

tcclevenger:
Release 4.3 unintentionally introduced the ability for `RangePolicy` constructor to take as it's final argument any type convertible to `ChunkSize` and implicitly construct a `ChunkSize` object for the input. Remove that by making the `ChunkSize(int)` constructor explicit.

### DIFF
--- a/core/src/Kokkos_ExecPolicy.hpp
+++ b/core/src/Kokkos_ExecPolicy.hpp
@@ -40,7 +40,7 @@ struct ParallelReduceTag {};
 
 struct ChunkSize {
   int value;
-  ChunkSize(int value_) : value(value_) {}
+  explicit ChunkSize(int value_) : value(value_) {}
 };
 
 /** \brief  Execution policy for work over a range of an integral type.
--- a/core/src/Kokkos_ExecPolicy.hpp
+++ b/core/src/Kokkos_ExecPolicy.hpp
@@ -38,9 +38,22 @@ struct ParallelForTag {};
 struct ParallelScanTag {};
 struct ParallelReduceTag {};
 
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+namespace Impl {
+struct OverloadHelper {};
+}  // namespace Impl
+#endif
+
 struct ChunkSize {
   int value;
   explicit ChunkSize(int value_) : value(value_) {}
+
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  KOKKOS_DEPRECATED_WITH_COMMENT(
+      "In RangePolicy constructor, ChunkSize struct was constructed implicitly "
+      "from a convertable type. Instead, pass Kokkos::ChunkSize(int).")
+  ChunkSize(int value_, Impl::OverloadHelper = {}) : value(value_) {}
+#endif
 };
 
 /** \brief  Execution policy for work over a range of an integral type.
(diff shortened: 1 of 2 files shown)


tcclevenger:
@dalg24 I'm trying to find a clean way to deprecate the old behavior. Is this necessary? I'm guessing in a single release cycle not too high of a chance that it is being used unintentionally.

dalg24:
When I said "investigate" in https://github.com/kokkos/kokkos/issues/7142#issuecomment-2233767695, I meant search downstream (Trilinos?) whether code breaks, since as Nevin pointed out, it is a breaking change even though we never intended to let someone construct `RangePolicy([exec,] (integral)begin, (integral)end, (integral)chunk_size)`.
Considering that we only accidentally enabled that in 4.3, usage should be limited.
A quick search in Trilinos yields
```
packages/stokhos/src/sacado/kokkos/Stokhos_Tpetra_Utilities.hpp:343:      const size_t ChunkSize = 16;
packages/stokhos/src/sacado/kokkos/Stokhos_Tpetra_Utilities.hpp:344:      const size_t M = (m+ChunkSize-1)/ChunkSize;
packages/stokhos/src/sacado/kokkos/Stokhos_Tpetra_Utilities.hpp:346:      Policy policy(M,ChunkSize,ChunkSize);
```
But interestingly, this code was committed in 2018 which means we need to find out if that code was actually built and if so why it worked(?)

dalg24:
If we decide (I expect you will make recommendations) we need to go through deprecations, we could do something along these lines https://godbolt.org/z/6TWrfx78a

tcclevenger:
In those Trilinos files, the `Policy` is a `TeamPolicy`, it is just a coincidence that "chunk size" is used as a variable name. I'll do a little more searching in Trilinos.

dalg24:
> In those Trilinos files, the `Policy` is a `TeamPolicy`, it is just a coincidence that "chunk size" is used as a variable name. I'll do a little more searching in Trilinos.

That makes more sense.  Thanks.

dalg24:
> If we decide (I expect you will make recommendations) we need to go through deprecations, we could do something along these lines https://godbolt.org/z/6TWrfx78a

I just realized this won't work. `ChunkSize cs{1};` will be ambiguous.

tcclevenger:
Yeah, I'm testing now we need something different.

masterleinad (approved):
We'll need to update the documentation accordingly. 
I doubt that anyone noticed any difference but I think it's fair to make the constructor explicit either way.

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -40,7 +40,12 @@ struct ParallelReduceTag {};
 
 struct ChunkSize {
   int value;
+  explicit ChunkSize(int value_) : value(value_) {}
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  template <typename T = void>
```
nliber, line 45: I didn't know that would work...

nliber (approved):
So, I tried the equivalent of this, but the RangePolicy constructor would now implicitly call the explicit ChunkSize constructor.  I don't know why (both under clang 18.1 and gcc 14.1).

dalg24 (changes requested):
Add a compile time test that check that when deprecated code is disabled, one cannot build a RangePolicy with a trailing integral argument

nmm0 (approved):
Thanks for doing this, implicit conversions are pretty unfortunate here

masterleinad (approved):

dalg24:
Please make sure you list it under "incompatibilities" in the change log

Issue #7153, opened 2024-07-24
Nightly build failure, intel/19.0.5 icpc: TestArrayOps.hpp(110): error #854: const variable "e" requires an initializer


ndellingwood:
**Describe the bug**

Nightly intel/19.0.5 jobs are having trouble handling updates from #7148

```
14:35:33 In file included from /home/jenkins/node/workspace/kokkos_osrhel7_intel_1905_openmp_pthread_serial/TestAll_2024-07-23_14.28.04/intel/19.0.5/Serial-release/core/unit_test/serial/TestSerial_ArrayOps.cpp(2):
14:35:33 /home/jenkins/node/workspace/kokkos_osrhel7_intel_1905_openmp_pthread_serial/kokkos/core/unit_test/TestArrayOps.hpp(110): error #854: const variable "e" requires an initializer -- class "Kokkos::Array<int, 0UL>" has no user-provided default constructor
14:35:33     constexpr E e;
14:35:33                 ^
14:35:33 
14:35:33 In file included from /home/jenkins/node/workspace/kokkos_osrhel7_intel_1905_openmp_pthread_serial/TestAll_2024-07-23_14.28.04/intel/19.0.5/Serial-release/core/unit_test/serial/TestSerial_ArrayOps.cpp(2):
14:35:33 /home/jenkins/node/workspace/kokkos_osrhel7_intel_1905_openmp_pthread_serial/kokkos/core/unit_test/TestArrayOps.hpp(111): error #854: const variable "f" requires an initializer -- class "Kokkos::Array<int, 0UL>" has no user-provided default constructor
14:35:33     constexpr E f;
```

**Please include the following for a minimal reproducer**

**Reproducer configuration**:
```
module load sems-cmake sems-intel/19.0.5

$KOKKOS_PATH/generate_makefile.bash --with-devices=Serial  --compiler=icpc --cxxflags="-O3 -Wall -Wunused-parameter -Wshadow -pedantic -Werror -Wsign-compare -Wtype-limits -Wuninitialized " --cxxstandard="17" --kokkos-path=$KOKKOS_PATH --no-examples   --with-options=enable_large_mem_tests
```

PR #7154, opened 2024-07-24
Fix Kokkos::Array<T, 0> default initialization for icpc

nliber:
This should address Issue #7153.

icpc incorrectly errors on `constexpr Kokkos::Array<T, 0>` needing initialization when it has a default constructor, as it is an empty specialization.

Prefer the suggestion of @masterleinad of removing the special member functions to address this issue.

### DIFF
--- a/core/unit_test/TestArrayOps.hpp
+++ b/core/unit_test/TestArrayOps.hpp
@@ -16,7 +16,6 @@
 
 #include <gtest/gtest.h>
 #include <Kokkos_Core.hpp>
-#include <numeric>
 
 namespace {
 
@@ -107,8 +106,8 @@ TEST(TEST_CATEGORY, array_operator_equal) {
   ASSERT_TRUE(a != c);
 
   using E = Kokkos::Array<int, 0>;
-  constexpr E e;
-  constexpr E f;
+  constexpr E e{};
+  constexpr E f{};
 
   static_assert(e == f);
   static_assert(!(e != f));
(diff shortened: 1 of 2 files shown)


masterleinad:
@ndellingwood Can you check that this indeed fixes #7153?

core/unit_test/TestArrayOps.hpp
```diff
@@ -16,7 +16,6 @@
 
 #include <gtest/gtest.h>
 #include <Kokkos_Core.hpp>
-#include <numeric>
```
dalg24, line 19: OK but generally avoid mixing in unrelated changes

dalg24 (approved):

masterleinad:
It seems the problem is the defaulted constructor, see https://godbolt.org/z/dTv5nTP8e. Can we just remove that instead?

ndellingwood (approved):
This resolved the compilation errors in the nightly intel/19.0.5 configuration, thanks

nliber:
@masterleinad We'd have to get rid of all the special member functions.

dalg24:
> @masterleinad We'd have to get rid of all the special member functions.

Any good reason not to get rid of them?

masterleinad (approved):

ndellingwood (approved):
The most recent changes compiled without issue in the intel/19.0.5 build

dalg24 (approved):

dalg24:
The SYCL build looks like it is headed for timeout town.  Everything else passed.

Issue #7152, opened 2024-07-24
Execution space constructors should be explicit


dalg24:
Background: We ran into disconcerting implicit conversions in #7151, `int` implicitly converting to `OpenMP`.

We should consider adding the `explicit` specifier to execution spaces constructors.

These are the backends that allow implicit conversions:
https://github.com/kokkos/kokkos/blob/bc7e02a16c779be323dd3f7673332d0ffd61b63c/core/src/Serial/Kokkos_Serial.hpp#L119
https://github.com/kokkos/kokkos/blob/bc7e02a16c779be323dd3f7673332d0ffd61b63c/core/src/OpenMP/Kokkos_OpenMP.hpp#L70
https://github.com/kokkos/kokkos/blob/bc7e02a16c779be323dd3f7673332d0ffd61b63c/core/src/HPX/Kokkos_HPX.hpp#L171-L180
https://github.com/kokkos/kokkos/blob/bc7e02a16c779be323dd3f7673332d0ffd61b63c/core/src/Cuda/Kokkos_Cuda.hpp#L169-L172
https://github.com/kokkos/kokkos/blob/bc7e02a16c779be323dd3f7673332d0ffd61b63c/core/src/HIP/Kokkos_HIP.hpp#L51-L53

nliber:
I'm in favor of fixing this now.

However, I think we should split up the constructors (since some of them are both single argument and multi argument due to defaulting parameters) so that the single argument constructors are `explicit` while the multi argument constructors are implicit.

tcclevenger:
Work in progress on this. Just putting some notes here. 

The serial impl does not allow explicit conversion since `NewInstance` is only explicitly constructible. 
https://github.com/kokkos/kokkos/blob/7d0e3e8184779c6f6b0d3ad7eac0341dbf273a2c/core/src/Serial/Kokkos_Serial.hpp#L79-L81
https://github.com/kokkos/kokkos/blob/bc7e02a16c779be323dd3f7673332d0ffd61b63c/core/src/Serial/Kokkos_Serial.hpp#L119

masterleinad:
> The serial impl does not allow explicit conversion since `NewInstance` is only explicitly constructible.

You could still do
```
Kokkos::RangePolicy<Kokkos::Serial>policy (NewInstance{}, begin, end);
```
instead of
```
Kokkos::RangePolicy<Kokkos::Serial> policy(Kokkos::Serial{NewInstance{}}, begin, end);
```

[...]

PR #7156, opened 2024-07-26
Make ExecutionSpace constructors explicit

tcclevenger:
Add deprecation constructor using the trick from @masterleinad 
```
explicit ExecSpace(Arg a) {}

template<typename T=void>
[[deprecated]]
ExecSpace(Arg a) : ExecSpace(a) {}
```
Notes:
- `OpenMP` was the only backend actually using implicit conversion internally.
- `Cuda(cudaStream_t stream, ManageStream manage_stream = ManageStream::no)` was split into 
```
Cuda(cudaStream_t stream, ManageStream manage_stream);
Cuda(cudaStream_t stream) : Cuda(stream, no) {}
```
per @nliber suggestion https://github.com/kokkos/kokkos/issues/7152#issuecomment-2248298545. Same for HIP. (Let me know if you had a different vision)
- Added test for SYCL to match non-deprecated from other backends since it was already using explicit conversions.

I still plam to search through Trilinos and see if any of these are being used implicitly.

### DIFF
--- a/core/src/Cuda/Kokkos_Cuda.hpp
+++ b/core/src/Cuda/Kokkos_Cuda.hpp
@@ -166,8 +166,17 @@ class Cuda {
 
   Cuda();
 
-  Cuda(cudaStream_t stream,
-       Impl::ManageStream manage_stream = Impl::ManageStream::no);
+  explicit Cuda(cudaStream_t stream) : Cuda(stream, Impl::ManageStream::no) {}
+
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  template <typename T = void>
+  KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Cuda execution space should be constructed explicitly.")
+  Cuda(cudaStream_t stream)
+      : Cuda(stream) {}
+#endif
+
+  Cuda(cudaStream_t stream, Impl::ManageStream manage_stream);
 
   KOKKOS_DEPRECATED Cuda(cudaStream_t stream, bool manage_stream);
 
--- a/core/unit_test/TestExecutionSpace.hpp
+++ b/core/unit_test/TestExecutionSpace.hpp
@@ -87,6 +87,9 @@ constexpr bool test_execspace_explicit_construction() {
 #endif
 #endif
 
+#ifdef KOKKOS_ENABLE_OPENACC
+  static_assert(!std::is_convertible_v<int, Kokkos::Experimental::OpenACC>);
+#endif
 #ifdef KOKKOS_ENABLE_SYCL
   static_assert(
       !std::is_convertible_v<sycl::queue, Kokkos::Experimental::SYCL>);
--- a/core/unit_test/TestExecutionSpace.hpp
+++ b/core/unit_test/TestExecutionSpace.hpp
@@ -44,9 +44,8 @@ TEST(TEST_CATEGORY, execution_space_as_class_data_member) {
 }
 #endif
 
-constexpr bool test_execspace_explicit() {
+constexpr bool test_execspace_explicit_construction() {
 #ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
-
 #ifdef KOKKOS_ENABLE_SERIAL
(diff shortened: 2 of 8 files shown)


core/unit_test/TestRangePolicyConstructors.hpp
```diff
@@ -204,8 +204,6 @@ constexpr bool test_chunk_size_explicit() {
 #ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
   static_assert(std::is_convertible_v<int, ChunkSize>);
   static_assert(std::is_constructible_v<ChunkSize, int>);
-  // FIXME some execution spaces are implicitly constructible from int
-  // which makes the constructor call ambiguous.
```
masterleinad, line 208: I would keep the comment without the FIXME.
```suggestion
  // Some execution spaces were implicitly constructible from int
  // which made the constructor call ambiguous.
```

masterleinad (approved):

cedricchevalier19 (approved):

core/unit_test/TestExecutionSpace.hpp
```diff
@@ -44,4 +44,57 @@ TEST(TEST_CATEGORY, execution_space_as_class_data_member) {
...
+  static_assert(!std::is_convertible_v<
+                hpx::execution::experimental::unique_any_sender<>&&,
+                Kokkos::Experimental::HPX>);
+#endif
+#endif
+
+#ifdef KOKKOS_ENABLE_SYCL
+  static_assert(
+      !std::is_convertible_v<sycl::queue, Kokkos::Experimental::SYCL>);
+#endif
```
dalg24, line 93: I am not a big fan of these backend-specific ifdefs but I don't have a better idea at the moment.
Note that OpenACC also had properly used explicit on a constructor from `int`:
https://github.com/kokkos/kokkos/blob/7d0e3e8184779c6f6b0d3ad7eac0341dbf273a2c/core/src/OpenACC/Kokkos_OpenACC.hpp#L76

core/unit_test/TestRangePolicyConstructors.hpp
dalg24: Agreed please keep the comment

core/src/Cuda/Kokkos_Cuda.hpp
```diff
@@ -166,8 +166,17 @@ class Cuda {
...
   Cuda();
 
-  Cuda(cudaStream_t stream,
-       Impl::ManageStream manage_stream = Impl::ManageStream::no);
+  explicit Cuda(cudaStream_t stream) : Cuda(stream, Impl::ManageStream::no) {}
+
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  template <typename T = void>
+  KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Cuda execution space should be constructed explicitly.")
```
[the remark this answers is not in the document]
tcclevenger: You are correct.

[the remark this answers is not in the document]
tcclevenger: Oh wait, sorry I read this backwards. I used "should" because if they have Deprecated code turned on it's technically not a "must" that they use the correct construction. But it is a must after v4. So I don't mind changing them.

core/unit_test/TestExecutionSpace.hpp
tcclevenger: I agree. Another option was individual tests in the backend specific folders, but that seemed to add a lot more lines.

tcclevenger: I'll add an OpenACC test to the group.

core/src/Cuda/Kokkos_Cuda.hpp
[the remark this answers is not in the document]
dalg24: Keep it as is then

dalg24 (approved):

dalg24:
Too impatient to wait for the NVCC debug build.  Everything else passed.

dalg24:
Please add to the 4..4 changelog under "incompatibility"

PR #7160, opened 2024-07-26
Add support for SpacemiT K60 (RISC-V)

diehlpk:
Recently, the first RISC_V CPU (SpacemiT K60) which supports vectorization was released. I am attempting to compile Kokkos with scalar types for now and this allows to compile Kokkos on the boards. 

A second pull request with support for SVE vectorization will follow.


dalg24 (commented):
I know virtually nothing about RISC-V architectures but it seems like you conflated a product name with the architecture.
Should it be called "64GCVB" instead?

lucbv:
@dalg24 do we have a clear idea of how we name these though? Some of the ARM ones are just `ARMVX_PRODUCT` others are `ARMVX`? On the other hand most `x86_64` architectures are just named after the vendor product line name?

dalg24:
> @dalg24 do we have a clear idea of how we name these though? Some of the ARM ones are just `ARMVX_PRODUCT` others are `ARMVX`? On the other hand most `x86_64` architectures are just named after the vendor product line name?

I don't.  Suggestion on how to improve the status quo are welcome.

diehlpk:
> I know virtually nothing about RISC-V architectures but it seems like you conflated a product name with the architecture. Should it be called "64GCVB" instead?

As far as I understand the full name of the CPU is `SpacemiT K1 Otca-core X60(RV64GCVB),RVA22,RVV1.0`.

So I am happy to use `RV64GCVB` instead.

diehlpk:
@dalg24 I updated the name.

diehlpk:
> > @dalg24 do we have a clear idea of how we name these though? Some of the ARM ones are just `ARMVX_PRODUCT` others are `ARMVX`? On the other hand most `x86_64` architectures are just named after the vendor product line name?
> 
> I don't. Suggestion on how to improve the status quo are welcome.

For the first RISC-V support, we use `RISCV_CPUNAME` and I liked that approach.

lucbv:
My worry is that we might have more than one `RISCV_CPUNAME` that ends up with the same compiler flags which is a bit unfortunate. I like RVA22V it states the profile and the optional V extension that is supported, RVA22V is also a superset of RV64GCBV so technically more precise than the later...

diehlpk:
Screenshot 2024-08-07 at 2 22 40 PM

It seems that vectorization works and we see improvements. Scaling does not work so well but it is a single-board computer anyways.

diehlpk:
> My worry is that we might have more than one `RISCV_CPUNAME` that ends up with the same compiler flags which is a bit unfortunate. I like RVA22V it states the profile and the optional V extension that is supported, RVA22V is also a superset of RV64GCBV so technically more precise than the later...

Ok, I can use  RVA22V and will do soon a new commit.

diehlpk:
@lucbv I changed the name. Could you please have a look again?

lucbv (approved):

lucbv:
@diehlpk that's okay with me, thanks for taking the time to look at this and discuss it!

diehlpk:
@dalg24 could you please have a look? I would like to get it merge to setup some weekly tests.

dalg24:
Retest this please (worried about the format)

crtrott (approved):
I agree with the name, the reasoning seems as good as any.

ndellingwood:
Looks like this needs a 4.5 changelog entry

PR #7164, opened 2024-07-29
 Add support for CUDA unified memory architectures i.e. Grace Hopper (2)

diehlpk:


cedricchevalier19:
Ok, so let's make a quick summary.

The idea is to exploit the memory capacities of the Grace-Hopper hardware: depending on the memory allocation mechanism, the CPU (resp. GPU) can access the GPU (resp. CPU) memory.

Since @crtrott's last commit, the current state is to rely on `cudaMallocManaged` to allocate memory. Technically speaking, this means that the effect of this PR is to substitute the default memory space with `shared space`.

crtrott:
Yes, using cudaMallocManaged with men advise, vs malloc with men advise was based on revised recommendations from NVIDIA. However, so far it looks like this comes with performance costs we may not want to accept. Currently, I am leaning almost more to not do anything specific for Grace Hopper, though maybe it might be good to have this thing here as an "impl" option to allow us to track how well the unified memory works.

diehlpk:
Close this pull request and work on the previous on again.

PR #7165, opened 2024-07-29
Add arg(...), norm(...) and proj(...) to Kokkos::complex

nliber:
Adding to `Kokkos::complex` the three missing functions from [[cmplx.over]](https://eel.is/c++draft/cmplx.over):

- `arg` phase angle
- `norm` - squared magnitude
- `proj` - projection onto Riemann sphere

Each of these has overloads for:
- `Kokkos::complex<T>`
- `RealType` (effectively casts its argument to `Kokkos::complex<RealType>)
- `IntegerType` (effectively casts its argument to `Kokkos::complex<double>`)

The standard woefully underspecifies these functions (corner cases, which things are overloads and which are templates, etc.), and it is unclear if cppreference.com is correct in its interpretation of the standard.

Notes:
- While the functions have been marked `constexpr`, this hasn't been tested yet.  Still have to track down which particular standard made a particular function `constexpr` (for instance, the implementation of `proj` uses `copysign`, and that wasn't made `constexpr` until C++23).
- In the tests, there are statements of the form `ASSERT_EQ(cds.size(), projcds.size());`.  While this really should be a `static_assert`, that was failing under clang18, and I didn't want to spend more time investigating it (it'll probably all optimize out anyway).
- I still need to write documentation.


nliber:
retest this please

dalg24:
Please resolve the conflicts and notify reviewers when this is ready

PR #7171, opened 2024-07-30
Move SYCL out of Experimental

masterleinad:
The SYCL backend has been stable for quite some time already. We are still supporting the 2023.0.0 release but we also can't promise that the compiler won't break Kokkos in the future. In any case, with `Aurora` getting production-ready the next release might be a good time to move the SYCL backend out of `Experimental`.
This pull request retains aliases in the `Experimental` namespace without deprecating them.


dalg24:
HIP was promoted in 4.0
Please elaborate why you think we should consider promoting the SYCL backend from Experimental to the Kokkos namespace now rather than wait about a year for the next major release.

masterleinad:
> Please elaborate why you think we should consider promoting the SYCL backend from Experimental to the Kokkos namespace now rather than wait about a year for the next major release.

We could but I was rather going by the anticipated date of acceptance for Aurora and that INCITE allocations on Aurora are starting in calendar year 2025. This pull request is not backward incompatible so I'm not sure it needs to align with a major release. We can discuss in a developer meeting.

crtrott:
Generally, I think I am in favor of this (i.e. aligning it with Aurora acceptance rather than major Kokkos version). We should consider to explicitly put the minimum supported OneAPI version to whatever is on Aurora for acceptance? What do you think?

masterleinad:
> We should consider to explicitly put the minimum supported OneAPI version to whatever is on Aurora for acceptance?

I expect there will be another update for the default compiler versions soon. At the moment, the oldest compiler available is 2023.1.0. We should evaluate if any of the guards we are using could be removed if we update the minimum
```
$ git grep -n "INTEL_LLVM" | grep "[<>]"
core/src/Kokkos_BitManipulation.hpp:104:    __INTEL_LLVM_COMPILER < 20240000
core/src/Kokkos_BitManipulation.hpp:114:    __INTEL_LLVM_COMPILER >= 20240000
core/src/Kokkos_Macros.hpp:195:    KOKKOS_COMPILER_INTEL_LLVM >= 20230100
core/src/SYCL/Kokkos_SYCL_Instance.hpp:336:#if (defined(__INTEL_LLVM_COMPILER) && __INTEL_LLVM_COMPILER < 20240000) || \
core/src/setup/Kokkos_Setup_SYCL.hpp:41:#if defined(__INTEL_LLVM_COMPILER) && __INTEL_LLVM_COMPILER >= 20230200
core/unit_test/TestBitManipulation.cpp:504:    (defined(__INTEL_LLVM_COMPILER) && __INTEL_LLVM_COMPILER >= 20240000)
```

dalg24:
I find the Aurora acceptance argument compelling and I would be ok with proceeding now.

masterleinad:
I would discuss raising the minimum compiler requirement separately from this pull request but I also don't see a lot if benefits from it at this point in time.

core/src/decl/Kokkos_Declare_SYCL.hpp
```diff
@@ -35,6 +35,16 @@
...
+/*
+namespace Kokkos {
+namespace Experimental {
+using SYCLDeviceUSMSpace           = ::Kokkos::SYCLDeviceUSMSpace;
+using SYCLHostUSMSpace = ::Kokkos::SYCLHostUSMSpace;
+using SYCLSharedUSMSpace    = ::Kokkos::SYCLSharedUSMSpace;
+using SYCL                = ::Kokkos::SYCL;
+}  // namespace Experimental
+}  // namespace Kokkos
+*/
```
dalg24, line 47: Did you define these aliases elsewhere?

masterleinad: They were not meant tobe  commented here. I checked before that defining these aliases and not changing anything in the tests would work.

dalg24 (approved):

tcclevenger (approved):

ldh4 (approved):
Looks good to me.

PR #7172, opened 2024-07-30
Skip Rangepolicy conversion check if roundtrip convertibility is not provided

### LINKED ISSUE #7005: RangePolicy construction fails if input bound types are not convertible to the policy's IndexType
Reported from https://github.com/ECP-copa/Cabana/pull/753#issuecomment-2103376470.
 
Changes made in #6754 expects `RangePolicy`'s input bound types to be convertible to the `RangePolicy`'s `IndexType`.

However, the conversion safety check in `RangePolicy` actually requires a full roundtrip convertibility:
https://github.com/kokkos/kokkos/blob/506da184f0aa8af9fbbca5bccd8c500b81865af5/core/src/Kokkos_ExecPolicy.hpp#L271-L272
and fails if only one way conversion is available.

Few options to mitigate this include:
- Adding constraints to the `RangePolicy` constructors (more conditions in `enable_if`)
- Making roundtrip convertibility a mandate (with a `static_assert`)
- Skipping the value preserving/narrowing check if only convertible one way (output a warning message about skipping the check)

ldh4:
Resolves #7005

`check_conversion_safety` in `RangePolicy` was requiring roundtrip convertibility between `member_type` and `IndexType`. The RangePolicy construction would fail to compile if convertibility is not supported for either direction.

This PR loosens that requirement, allowing RangePolicy be constructible as long as the user input index type (`IndexType`) is convertible to `member_type`. If `member_type` is not convertible to `IndexType`, the conversion safety check will be skipped.

### DIFF
--- a/core/src/Kokkos_ExecPolicy.hpp
+++ b/core/src/Kokkos_ExecPolicy.hpp
@@ -261,8 +261,10 @@ class RangePolicy : public Impl::PolicyTraits<Properties...> {
           "not preserve its original value.\n";
       bool warn = false;
 
-      if constexpr (std::is_signed_v<IndexType> !=
-                    std::is_signed_v<member_type>) {
+      if constexpr (std::is_arithmetic_v<IndexType> &&
+                    std::is_arithmetic_v<member_type> &&
+                    (std::is_signed_v<IndexType> !=
+                     std::is_signed_v<member_type>)) {
         // check signed to unsigned
         if constexpr (std::is_signed_v<IndexType>)
           warn |= (bound < static_cast<IndexType>(
@@ -334,20 +336,20 @@ class RangePolicy : public Impl::PolicyTraits<Properties...> {
   };
 };
 
-RangePolicy() -> RangePolicy<>;
+RangePolicy()->RangePolicy<>;
 
-RangePolicy(int64_t, int64_t) -> RangePolicy<>;
-RangePolicy(int64_t, int64_t, ChunkSize const&) -> RangePolicy<>;
+RangePolicy(int64_t, int64_t)->RangePolicy<>;
+RangePolicy(int64_t, int64_t, ChunkSize const&)->RangePolicy<>;
 
-RangePolicy(DefaultExecutionSpace const&, int64_t, int64_t) -> RangePolicy<>;
+RangePolicy(DefaultExecutionSpace const&, int64_t, int64_t)->RangePolicy<>;
 RangePolicy(DefaultExecutionSpace const&, int64_t, int64_t, ChunkSize const&)
-    -> RangePolicy<>;
+    ->RangePolicy<>;
 
 template <typename ES, typename = std::enable_if_t<is_execution_space_v<ES>>>
-RangePolicy(ES const&, int64_t, int64_t) -> RangePolicy<ES>;
+RangePolicy(ES const&, int64_t, int64_t)->RangePolicy<ES>;
 
 template <typename ES, typename = std::enable_if_t<is_execution_space_v<ES>>>
-RangePolicy(ES const&, int64_t, int64_t, ChunkSize const&) -> RangePolicy<ES>;
+RangePolicy(ES const&, int64_t, int64_t, ChunkSize const&)->RangePolicy<ES>;
 
 }  // namespace Kokkos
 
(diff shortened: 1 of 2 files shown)


core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -249,45 +249,55 @@ class RangePolicy : public Impl::PolicyTraits<Properties...> {
...
+      }
 #else
-    (void)bound;
+      (void)bound;
+#endif
+    } else {
+#ifdef KOKKOS_ENABLE_DEPRECATION_WARNINGS
+      std::string msg =
+          "Kokkos RangePolicy bound type warning: Two-way convertibility requirement on policy index type and input bound type is not met.
+          Conversion safety check is skipped."
```
ldh4, line 297: I think this could be better worded. If there's any suggestion, please let me know.

core/unit_test/TestRangePolicyConstructors.hpp
```diff
@@ -128,6 +128,8 @@ TEST(TEST_CATEGORY_DEATH, range_policy_implicitly_converted_bounds) {
   using IntIndexType  = Kokkos::IndexType<int>;
   using UIntPolicy    = Kokkos::RangePolicy<TEST_EXECSPACE, UIntIndexType>;
   using IntPolicy     = Kokkos::RangePolicy<TEST_EXECSPACE, IntIndexType>;
+  using IntAtomicDataElementType =
+      Kokkos::Impl::AtomicDataElement<Kokkos::ViewTraits<int>>;
```
dalg24, line 132: Please define a type that is convertible to the range policy integral type but not constructible from it instead.
It will improve readability and avoid introducing an unnecessary dependency with implementation details from other facilities in the codebase.

core/src/Kokkos_ExecPolicy.hpp
dalg24: I don't think we should warn at all.
We only want to raise warnings when we know for sure that something is wrong.  The current version would yield false positive and would force users to disable warnings and potential miss things they should be aware of.

dalg24 (changes requested):

core/src/Kokkos_ExecPolicy.hpp
ldh4: That's a fair point, that this warning is a false positive in many cases.
Removed this warning message.

core/unit_test/TestRangePolicyConstructors.hpp
ldh4: Replaced to use a simple struct that is convertible to rangepolicy index_type instead.

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -249,45 +249,48 @@ class RangePolicy : public Impl::PolicyTraits<Properties...> {
   // To be replaced with std::in_range (c++20)
   template <typename IndexType>
   static void check_conversion_safety(const IndexType bound) {
+    if constexpr (std::is_convertible_v<member_type, IndexType>) {
```
dalg24, line 252: Please add a comment above this line, something like
"Checking that the round-trip conversion preserve the value"

core/unit_test/TestRangePolicyConstructors.hpp
```diff
@@ -158,6 +166,14 @@ TEST(TEST_CATEGORY_DEATH, range_policy_implicitly_converted_bounds) {
...
                  get_error_msg(expected, test_val));
   }
+  {
+    typename IntIndexType::type test_val = -1;
+    ASSERT_NO_THROW((void)IntPolicy(
+        -2,
+        TestIndexConversionCheck<typename IntPolicy::index_type>(test_val)));
+    ASSERT_NO_THROW((void)UIntPolicy(
+        0, TestIndexConversionCheck<typename IntPolicy::index_type>(test_val)));
+  }
```
dalg24, line 176: Shouldn't that pass regardless of whether deprecated code 4 is enabled or not?

core/unit_test/TestRangePolicyConstructors.hpp
```diff
@@ -158,6 +166,14 @@ TEST(TEST_CATEGORY_DEATH, range_policy_implicitly_converted_bounds) {
     ASSERT_DEATH({ (void)UIntPolicy(test_val, 10, Kokkos::ChunkSize(2)); },
                  get_error_msg(expected, test_val));
   }
+  {
+    typename IntIndexType::type test_val = -1;
+    ASSERT_NO_THROW((void)IntPolicy(
+        -2,
+        TestIndexConversionCheck<typename IntPolicy::index_type>(test_val)));
+    ASSERT_NO_THROW((void)UIntPolicy(
+        0, TestIndexConversionCheck<typename IntPolicy::index_type>(test_val)));
```
dalg24, line 175:
```suggestion
        0, TestIndexConversionCheck<typename UIntPolicy::index_type>(test_val)));
```

core/unit_test/TestRangePolicyConstructors.hpp
dalg24: Also I expect we only care about testing that a non-value preserving construction goes through, that is just the 2nd assertion?

core/src/Kokkos_ExecPolicy.hpp
ldh4: Added a comment.

core/unit_test/TestRangePolicyConstructors.hpp
ldh4: Ended up removing this line.

core/unit_test/TestRangePolicyConstructors.hpp
ldh4: Fixed the test. Also removed a redundant test.
Whether `KOKKOS_ENABLE_DEPRECATED_CODE_4` is enabled or not doesn't affect the test, but it now will fail to compile without `if constexpr (std::is_convertible_v<member_type, IndexType>)` at the beginning of `check_conversion_safety()`.

nliber (approved):

core/unit_test/TestRangePolicyConstructors.hpp
```diff
@@ -158,6 +166,15 @@ TEST(TEST_CATEGORY_DEATH, range_policy_implicitly_converted_bounds) {
...
                  get_error_msg(expected, test_val));
   }
+  {
+    using test_type = TestIndexConversionCheck<typename IntPolicy::index_type>;
+
+    static_assert(std::is_convertible_v<test_type, IntPolicy::index_type>);
+    static_assert(!std::is_convertible_v<IntPolicy::index_type, test_type>);
+
+    typename IntIndexType::type test_val = -1;
+    ASSERT_NO_THROW((void)IntPolicy(-2, test_type(&test_val)));
```
dalg24, line 176:
```suggestion
    using test_type = TestIndexConversionCheck<UIntPolicy::index_type>;

    static_assert(std::is_convertible_v<test_type, UIntPolicy::index_type>);
    static_assert(!std::is_convertible_v<UIntPolicy::index_type, test_type>);

    UIntPolicy::index_type test_val = -1;
    ASSERT_NO_THROW((void)UIntPolicy(-2, &test_val));
```
When you pick the signed integer version, you only test that it compiles.  If you pick the unsigned one, you also reveal that we would not currently catch that the issue.  I think it makes it a better choice.

core/unit_test/TestRangePolicyConstructors.hpp
dalg24: > Whether `KOKKOS_ENABLE_DEPRECATED_CODE_4` is enabled or not doesn't affect the test, but it now will fail to compile without `if constexpr (std::is_convertible_v<member_type, IndexType>)` at the beginning of `check_conversion_safety()`.

My point is the test you added is guarded by `#ifndef KOKKOS_ENABLE_DEPRECATED_CODE_4` and that it should not be guarded.

ldh4: Moved this out of `#ifndef KOKKOS_ENABLE_DEPRECATED_CODE_4` guard and made into its own test.

core/unit_test/TestRangePolicyConstructors.hpp
ldh4: Converted to unsigned.

masterleinad (approved):

masterleinad:
```
/__w/kokkos/kokkos/core/unit_test/TestRangePolicyConstructors.hpp(143): error #68: integer conversion resulted in a change of sign
    UIntIndexType::type test_val = -1;
                                   ^

/__w/kokkos/kokkos/core/src/Kokkos_ExecPolicy.hpp(251): error #869: parameter "bound" was never referenced
    static void check_conversion_safety(const IndexType bound) {
                                                        ^
          detected during:
            instantiation of "void Kokkos::RangePolicy<Properties...>::check_conversion_safety(IndexType) [with Properties=<Kokkos::OpenMP, Kokkos::IndexType<unsigned int>>, IndexType=<unnamed>::TestIndexConversionCheck<unsigned int>]" at line 143
            instantiation of "Kokkos::RangePolicy<Properties...>::RangePolicy(const Kokkos::Impl::PolicyTraits<Properties...>::execution_space &, IndexType1, IndexType2) [with Properties=<Kokkos::OpenMP, Kokkos::IndexType<unsigned int>>, IndexType1=int, IndexType2=<unnamed>::TestIndexConversionCheck<unsigned int>, <unnamed>=false]" at line 128
            instantiation of "Kokkos::RangePolicy<Properties...>::RangePolicy(IndexType1, IndexType2) [with Properties=<Kokkos::OpenMP, Kokkos::IndexType<unsigned int>>, IndexType1=int, IndexType2=<unnamed>::TestIndexConversionCheck<unsigned int>, <unnamed>=false]" at line 144 of "/__w/kokkos/kokkos/core/unit_test/Te
```

core/unit_test/TestRangePolicyConstructors.hpp
```diff
@@ -131,6 +131,19 @@ struct TestIndexConversionCheck {
   T* const value;
 };
 
+TEST(TEST_CATEGORAY, range_policy_one_way_convertible_bounds) {
```
nliber, line 134:
```suggestion
TEST(TEST_CATEGORY, range_policy_one_way_convertible_bounds) {
```

nliber (changes requested):

nliber (approved):

masterleinad:
Segfaulting in `range_policy_round_trip_conversion_fires`.

dalg24:
Please fix

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -247,46 +247,50 @@ class RangePolicy : public Impl::PolicyTraits<Properties...> {
...
-    }
+      std::string msg =
+          "Kokkos::RangePolicy bound type error: an unsafe implicit conversion "
+          "is performed on a bound (" +
+          std::to_string(bound) +
+          "), which may "
+          "not preserve its original value.\n";
+      bool warn = false;
+
+      if constexpr (std::is_arithmetic_v<IndexType> &&
```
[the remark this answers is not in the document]
ldh4: Removed `std::is_arithmetic_v<IndexType>` from this condition.

dalg24:
Ignoring unrelated failures (NVC++ segfault, Clang+CUDA and CUDA with generated makefiles with GPU not detected at runtime, and host function called from host device function warning treated as error)

dalg24:
Please add to the change log under bug fix

PR #7175, opened 2024-07-31
Hide `IMPL_REF_COUNT_BRANCH_UNLIKELY` option

dalg24:
Fixup for #6730

On 2nd thought, I don't think we should expose that option/macro.  It unnecessarily increase the cognitive load on users.
That option is really a contingency plan.


masterleinad (approved):

nmm0 (approved):
:+1:

nliber (approved):

dalg24:
Ignoring unrelated compile time error in the NVHPC build

PR #7176, opened 2024-07-31
remove usage of ENABLE_CXX11_DISPATCH_LAMBDA

tjhei:
The macro KOKKOS_ENABLE_CXX11_DISPATCH_LAMBDA is set unconditionally for a long time, so there should not be any reason to check it in the examples or tests.

FYI @masterleinad

### DIFF
--- a/algorithms/unit_tests/TestStdAlgorithmsForEach.cpp
+++ b/algorithms/unit_tests/TestStdAlgorithmsForEach.cpp
@@ -55,7 +55,6 @@ void test_for_each(const ViewType view) {
   std::for_each(KE::begin(expected), KE::end(expected), non_mod_functor);
   compare_views(expected, view);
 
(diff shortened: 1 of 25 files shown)


dalg24 (approved):
In case that is useful to anyone else.

https://github.com/kokkos/kokkos/blob/dbb0abb677554ed8339cef41825424fe94cbaae3/core/src/Kokkos_Macros.hpp#L107

tjhei:
Yes, I left the macro and the additional checks in https://github.com/kokkos/kokkos/blob/dbb0abb677554ed8339cef41825424fe94cbaae3/core/unit_test/TestCompilerMacros.cpp#L33
as is.

Also see #5964

masterleinad (approved):

dalg24:
Ignoring SYCL failure (clearly unrelated)

PR #7180, opened 2024-07-31
mention indent/formatting script

### LINKED ISSUE #7177: document how to indent
Would it be possible to better document how one is supposed to indent/format the code before contributing? I looked in the following places before discovering ./scripts/apply-clang-format:
- CMake generated make target
- project README
- top of .clang-format file
- https://github.com/kokkos/kokkos/blob/dbb0abb677554ed8339cef41825424fe94cbaae3/CONTRIBUTING.md 

It is also somewhat difficult to acquire clang-format-8: homebrew and debian/ubuntu packages do not contain such an old version.

tjhei:
closes #7177

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -7,6 +7,8 @@ We actively welcome pull requests.
 3. If you've changed APIs, update the documentation.
 4. Ensure the test suite passes.
 
+Before sending your patch for review, please try to ensure that it is formatted properly. We use clang-format version 8.0.0 for this.
+
 ## Issues
 We use GitHub issues to track public bugs. Please ensure your description is clear and has sufficient instructions to be able to reproduce the issue.
 


masterleinad (approved):

CONTRIBUTING.md
```diff
@@ -3,9 +3,10 @@
 ## Pull Requests
 We actively welcome pull requests.
 1. Fork the repo and create your branch from `develop`.
-2. If you've added code that should be tested, add tests.
-3. If you've changed APIs, update the documentation.
-4. Ensure the test suite passes.
+2. Apply code formatting by running ./scripts/apply-clang-format
```
dalg24, line 6: I don't know that I agree with that step.
I would only mention something like "Before sending your patch for review, please try to ensure that it is formatted properly.  We use clang-format version 8.0.0 for this."

.clang-format
dalg24, line 1: I think that this file was fine.

dalg24 (commented):
I have mixed feelings about recommending to apply `scripts/apply-clang-format`.

tjhei:
> I have mixed feelings about recommending to apply `scripts/apply-clang-format`.

I am confused. In #7177 you write "I like all the suggestions above".

> Before sending your patch for review, please try to ensure that it is formatted properly. We use clang-format version 8.0.0 for this."

The script also fixes other issues (copyright, trailing whitespace, etc.) and is much easier to use than trying to figure out how to run clang-format manually on all files.

dalg24:
> > I have mixed feelings about recommending to apply `scripts/apply-clang-format`.
> 
> I am confused. In #7177 you write "I like all the suggestions above".

Diagonal reading

The portability of `apply-clang-format` is limited. It was just put together to check the format in the CI. I find that the important part is letting prospective contributors know that we use clang-format. If they are familiar with the tool, they certainly don't need that script.

The no trailing whitespaces and check the copyright part is detail, it can be covered in the extended developer information.

tjhei:
> The portability of `apply-clang-format` is limited. It was just put together to check the format in the CI. I find that the important part is letting prospective contributors know that we use clang-format. If they are familiar with the tool, they certainly don't need that script.

Alright. If you think that makes more sense, then that is what I will do. I was assuming that using the script is a good default recommendation for new users (it certainly is for me).

nliber (approved):

ldh4 (approved):

dalg24:
(format step passed on Jenkins)

PR #7186, opened 2024-08-01
Avoid nesting fences into parallel_for when initializing/deleting views

dalg24:
At this time, this PR is more support for discussion than me saying this is a bug.

I interpret the version in develop as the parallel_for accounts for the full view initialization/deletion.
But I tend to think it is wrong and that we shouldn't have nested the fence into the parallel_for.

I am interested to hear what others think.


crtrott (approved):

masterleinad (approved):
This makes more sense to me as well. The View initialization kernel should not be treated differently from other kernels. For `parallel_reduce` we also fence after `endParallelReduce`.

dalg24:
I was going to omit in the change log.  Let me know if you disagree.

masterleinad:
> I was going to omit in the change log. Let me know if you disagree.

Yeah, I don't think we need that.

PR #7189, opened 2024-08-02
Fix atomic accessor for pre-volta GPU architectures

### LINKED ISSUE #7187: Cuda+Pascal: feature .relaxed requires sm_70 or higher
**Describe the bug**

When compiling programs with parallel_reduce for OpenMP + Cuda, with Pascal61 architecture, ptxas throws out errors like
```
ptxas: Feature '.relaxed' requires .target sm_70 or higher
```

A bisect points to #7089 as the first commit where this happens.

1. CUDA 12.5 with GCC 11.4.1 as host compiler
2. Replicates with today's develop branch, but b84125ed6 is the earliest commit where it happens
3. Architectures: SKX and Pascal61

Full reproducer: replace contents of example/build_cmake_in_tree/cmake_example.cpp with
```
#include <Kokkos_Core.hpp>
int main() {
  Kokkos::initialize();
  {
    double result;
    Kokkos::parallel_reduce(
        Kokkos::RangePolicy<>(0, 100),
        KOKKOS_LAMBDA(int i, double& lsum)
        {
          lsum += i*i - 1;
        }, result);
    printf("%f\n", result);
  }
  Kokkos::finalize();
  return 0;
}
```
Then configure with this and build.
```
cmake \
-DKokkos_ENABLE_CUDA=ON \
-DKokkos_ENABLE_OPENMP=ON \
-DKokkos_ARCH_SKX=ON \
-DKokkos_ARCH_PASCAL61=ON \
../kokkos/example/build_cmake_in_tree
```
Should see
```
ptxas /tmp/tmpxft_003a502d_00000000-6_cmake_example.ptx, line 273;
error   : Feature '.relaxed' requires .target sm_70 or higher
```

dalg24:
Fix #7187 

We forgot to include that "Kokkos_Atomics_Desul_Config.hpp" configuration helper that translates Kokkos GPU macro architectures into Desul ones.

Admittedly this is not great that Desul expects these define macros to just be properly set by the user.  IIRC we didn't really want to handle architectures in the configuration which granted a lot of freedom downstream but alas also complicates usage.

### DIFF
--- a/core/src/Kokkos_Atomics_Desul_Volatile_Wrapper.hpp
+++ b/core/src/Kokkos_Atomics_Desul_Volatile_Wrapper.hpp
@@ -22,7 +22,6 @@ static_assert(false,
 #ifndef KOKKOS_DESUL_ATOMICS_VOLATILE_WRAPPER_HPP_
 #define KOKKOS_DESUL_ATOMICS_VOLATILE_WRAPPER_HPP_
 #include <Kokkos_Macros.hpp>
-#include <Kokkos_Atomics_Desul_Config.hpp>
 #include <desul/atomics.hpp>
 
 #ifdef KOKKOS_ENABLE_ATOMICS_BYPASS
--- a/core/src/Kokkos_Atomics_Desul_Wrapper.hpp
+++ b/core/src/Kokkos_Atomics_Desul_Wrapper.hpp
@@ -22,8 +22,6 @@ static_assert(false,
 #ifndef KOKKOS_DESUL_ATOMICS_WRAPPER_HPP_
 #define KOKKOS_DESUL_ATOMICS_WRAPPER_HPP_
 #include <Kokkos_Macros.hpp>
-
-#include <Kokkos_Atomics_Desul_Config.hpp>
 #include <desul/atomics.hpp>
 
 #include <impl/Kokkos_Volatile_Load.hpp>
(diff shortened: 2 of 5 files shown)


dalg24:
https://github.com/kokkos/kokkos/blob/f96d5b5f551560d05d70b4a8c801567e38bc7ffb/core/src/Kokkos_Atomics_Desul_Config.hpp#L24-L33

dalg24:
> What about just merging `Kokkos_Atomics_Desul_Config.hpp` with `Kokkos_Macros.hpp`?

I would push back against "merging".  These do not really belong into that macros header which currently stays clear of architectures considerations.
The merit of what you suggest is that this would avoid reintroducing that same issue in the future.  I suppose I would consider including the desul configuration here
https://github.com/kokkos/kokkos/blob/f96d5b5f551560d05d70b4a8c801567e38bc7ffb/core/src/Kokkos_Macros.hpp#L56-L59

masterleinad:
> The merit of what you suggest is that this would avoid reintroducing that same issue in the future. I suppose I would consider including the desul configuration here

I would prefer that over the current approach.

dalg24:
Note that `Kokkos_Macros.hpp` still must be included before `desul/atomics.hpp`

dalg24:
(had forgotten to commit the changes to Kokkos_Macros.hpp)

masterleinad (approved):

dalg24:
CI failures are clearly unrelated (at code checkout)

PR #7195, opened 2024-08-05
launch_bounds example: support arch with lower thread count

### LINKED ISSUE #7191: CUDA: ptxas warning in launch bounds example with compute capabilities 86 and 89
> launch_bounds produces a compiler warning on my machine:
> ```
> ptxas warning : Value of threads per SM for entry
> _ZN6Kokkos4Impl33cuda_parallel_launch_local_memoryINS0_14ParallelReduceINS0_22CombinedFunctorReducerI9collisionNS0_15FunctorAnalysisINS0_23FunctorPatternInterface6REDUCEENS_11RangePolicyIJNS_12LaunchBoundsILj512ELj4EEEEEES4_iE7ReducerEvEESB_NS_4CudaEEELj512ELj4EEEvT_
> is out of range. .minnctapersm will be ignored

_Originally posted by @tjhei in https://github.com/kokkos/kokkos/pull/7179#issue-2440684778

(Timo's report was about cc 89 but Daniel confirmed to me on Slack he could reproduce with 86)

tjhei:
fixes #7191

### DIFF
--- a/example/tutorial/launch_bounds/launch_bounds_reduce.cpp
+++ b/example/tutorial/launch_bounds/launch_bounds_reduce.cpp
@@ -2,7 +2,6 @@
 // SPDX-FileCopyrightText: Copyright Contributors to the Kokkos project
 
 #include <Kokkos_Core.hpp>
-#include <cstdio>
 
 //
 // First reduction (parallel_reduce) example:
@@ -101,8 +100,7 @@ int main(int argc, char* argv[]) {
   Kokkos::initialize(argc, argv);
   const int n = 10000;
 
-  // Compute and count hash collisions in
-  // parallel, using Kokkos.
+  // Compute and count hash collisions in parallel, using Kokkos.
   // This is not really a useful algorithm, but it demonstrates the
   // LaunchBounds functionality
   int sum1 = 0;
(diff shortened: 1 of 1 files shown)


masterleinad:
I'm having a very hard time actually seeing any improvements with different launch bounds. For me the point here is to show how to use `LaunchBounds` and it's less of a problem if we don't actually see fewer registers being used (since that is also very much dependent on the compiler version and compute capability).

crtrott:
This example needs to be completely reworked, it's meaningless right now - NVCC will use only 26 register for the kernel either way.

janciesko:
This addresses a compiler warning RE occupancy as https://github.com/kokkos/kokkos/issues/7191. I'll mark this ready for review. Edit: Noting that the test needs to be reworked.

masterleinad:
As it stands, I would rather remove the example.

janciesko:
It might make sense to keep this example, remove specific mention of registers and actually fix the warning (this PR). 

The narrative for the tutorial is more: "This is how you use this thing and even though launch bounds have changed and register usage might have changes, results are still the same."

We could introspect via `get_cuda_func_attributes` and actually print out the compiler hints but not in this PR.

janciesko (approved):
This PR fixes a compiler warning. I have added some minor changes to improve the test beyond the original intent of this PR given the comments.

ldh4 (approved):
I agree that this tutorial is now more of:
> For me the point here is to show how to use LaunchBounds

and

> The narrative for the tutorial is more: "This is how you use this thing and even though launch bounds have changed and register usage might have changes, results are still the same."

I think these changes are fine for the purpose of addressing the warning from #7191
Reworking or completely removing this example could be handled separately as a follow-up.

PR #7198, opened 2024-08-05
Update cmake and make configurations for Clacc/LLVM compiler

seyonglee:
Update cmake and make configurations for Clacc/LLVM compiler to compile the OpenACC backend.
 - Disable the OpenACC atomic implementations when using Clacc.
 - Disable unsupported unit tests for Clacc.
 - Support optional CUDA_PATH and ROCM_PATH, which are used for the OpenACC backend to calculate concurrency for NVIDIA GPUs and AMD GPUs (actual OpenACC backend implementation will be added in a separate PR).


Makefile.kokkos
```diff
@@ -168,7 +168,7 @@ KOKKOS_INTERNAL_COMPILER_CRAY        := $(strip $(shell $(CXX) -craype-verbose 2
 KOKKOS_INTERNAL_COMPILER_NVCC        := $(strip $(shell echo "$(shell export OMPI_CXX=$(OMPI_CXX); export MPICH_CXX=$(MPICH_CXX); $(CXX) --version 2>&1 | grep -c nvcc)>0" | bc))
 KOKKOS_INTERNAL_COMPILER_NVHPC       := $(strip $(shell $(CXX) --version 2>&1 | grep -c "nvc++"))
 KOKKOS_INTERNAL_COMPILER_CLANG       := $(call kokkos_has_string,$(KOKKOS_CXX_VERSION),clang)
-KOKKOS_INTERNAL_COMPILER_CRAY_CLANG  := $(strip $(shell $(CXX) -craype-verbose 2>&1 | grep -c "clang++"))
+KOKKOS_INTERNAL_COMPILER_CRAY_CLANG  := $(strip $(shell $(CXX) -craype-verbose 2>&1 | grep -v "error:" | grep -c "clang++"))
```
masterleinad, line 171: Why would we filter out "error:"?

Makefile.kokkos
```diff
@@ -272,6 +272,8 @@ ifeq ($(KOKKOS_INTERNAL_USE_OPENMPTARGET), 1)
     KOKKOS_INTERNAL_OPENMPTARGET_FLAG := -fiopenmp -Wno-openmp-mapping
   else ifeq ($(KOKKOS_INTERNAL_COMPILER_NVHPC), 1)
     KOKKOS_INTERNAL_OPENMPTARGET_FLAG := -mp=gpu 
+  else ifeq ($(KOKKOS_INTERNAL_COMPILER_CLANG), 1)
+    KOKKOS_INTERNAL_OPENMPTARGET_FLAG := -fopenmp -Xopenmp-target
```
masterleinad, line 276: Why are you modifying `OpenMPTarget` flags here?

Makefile.kokkos
```diff
@@ -401,13 +405,16 @@ ifeq ($(KOKKOS_INTERNAL_USE_ARCH_NVIDIA), 0)
...
   ifeq ($(KOKKOS_INTERNAL_COMPILER_CLANG), 1)
-    KOKKOS_INTERNAL_NVCC_PATH := $(shell which nvcc)
     CUDA_PATH ?= $(KOKKOS_INTERNAL_NVCC_PATH:/bin/nvcc=)
     ifeq ($(KOKKOS_INTERNAL_USE_OPENMPTARGET), 1)
       KOKKOS_INTERNAL_OPENMPTARGET_FLAG := $(KOKKOS_INTERNAL_OPENMPTARGET_FLAG) --cuda-path=$(CUDA_PATH)
     endif
   endif
+  ifeq ($(KOKKOS_INTERNAL_COMPILER_NVHPC), 1)
+    CUDA_PATH ?= $(KOKKOS_INTERNAL_NVCC_PATH:/bin/nvcc=)
+  endif
```
masterleinad, line 417: Why is this necessary? Don't we set `KOKKOS_INTERNAL_NVCC_PATH` in the same way already in line 148?

Makefile.kokkos
```diff
@@ -456,6 +463,14 @@ ifeq ($(KOKKOS_INTERNAL_USE_ARCH_AMD_GFX1100), 0)
...
 endif
 KOKKOS_INTERNAL_USE_ARCH_AMD_GFX1103 := $(call kokkos_has_string,$(KOKKOS_ARCH),AMD_GFX1103)
+KOKKOS_INTERNAL_USE_ARCH_AMD := $(shell expr $(KOKKOS_INTERNAL_USE_ARCH_AMD_GFX906)  \
+                                              + $(KOKKOS_INTERNAL_USE_ARCH_AMD_GFX908)  \
+                                              + $(KOKKOS_INTERNAL_USE_ARCH_AMD_GFX90A)  \
+                                              + $(KOKKOS_INTERNAL_USE_ARCH_AMD_GFX940)  \
+                                              + $(KOKKOS_INTERNAL_USE_ARCH_AMD_GFX942) \
+                                              + $(KOKKOS_INTERNAL_USE_ARCH_AMD_GFX1030) \
+                                              + $(KOKKOS_INTERNAL_USE_ARCH_AMD_GFX1100) \
+                                              + $(KOKKOS_INTERNAL_USE_ARCH_AMD_GFX1103))
```
masterleinad, line 473: Should this be in a separate pull request?

Makefile.kokkos
```diff
@@ -1014,86 +1029,123 @@ ifeq ($(KOKKOS_INTERNAL_USE_OPENMPTARGET), 1)
   endif
 endif
 
+ifeq ($(KOKKOS_INTERNAL_USE_OPENACC), 1)
+  ifeq ($(KOKKOS_INTERNAL_COMPILER_CLANG), 1)
+      KOKKOS_INTERNAL_CUDA_ARCH_FLAG=-fopenmp-targets=nvptx64-nvidia-cuda -Xopenmp-target=nvptx64-nvidia-cuda -march
+      KOKKOS_INTERNAL_AMD_ARCH_FLAG=-fopenmp-targets=amdgcn-amd-amdhsa -Xopenmp-target=amdgcn-amd-amdhsa -march
```
masterleinad, line 1035: So `OpenACC` using `openmp-target` flags?

cmake/kokkos_arch.cmake
```diff
@@ -845,14 +845,35 @@ IF (KOKKOS_ENABLE_OPENACC)
       Clang -Xopenmp-target=nvptx64-nvidia-cuda -march=${CLANG_CUDA_ARCH}
             -fopenmp-targets=nvptx64-nvidia-cuda
     )
+    IF(DEFINED ENV{CUDA_PATH})
+      COMPILER_SPECIFIC_LINK_OPTIONS(Clang -L$ENV{CUDA_PATH}/lib64)
+    ENDIF()
+    COMPILER_SPECIFIC_LIBS(
+      Clang -lcudart
+      NVHPC -cuda
+    )
```
masterleinad, line 854: Why is `FindCUDATookit` not sufficient?

cmake/kokkos_arch.cmake
```diff
@@ -845,14 +845,35 @@ IF (KOKKOS_ENABLE_OPENACC)
...
   ELSEIF(KOKKOS_AMDGPU_ARCH_FLAG)
     COMPILER_SPECIFIC_FLAGS(
       Clang -Xopenmp-target=amdgcn-amd-amdhsa -march=${KOKKOS_AMDGPU_ARCH_FLAG}
             -fopenmp-targets=amdgcn-amd-amdhsa
     )
+    IF(DEFINED ENV{ROCM_PATH})
+      COMPILER_SPECIFIC_FLAGS(Clang -I$ENV{ROCM_PATH}/include)
+      COMPILER_SPECIFIC_LINK_OPTIONS(Clang -L$ENV{ROCM_PATH}/lib)
+    ENDIF()
+    COMPILER_SPECIFIC_LIBS(Clang -lamdhip64)
```
masterleinad, line 864: Is finding `ROCm` not sufficient?

Makefile.kokkos
seyonglee: it is because without that filtering, output will contain multiple lines depending on which compilers to use.

Makefile.kokkos
seyonglee: I agree that this change is unrelated to the OpenACC backend; I added that since we can compile the OpenMPTarget backend using the public Clang/LLVM but not included in the Makefile.
Or, I can delete this and submit it as a separate PR.

Makefile.kokkos
seyonglee: The line 148 is for the CUDA backend, which does not include the cases where other backends (OpenACC and OpenMPTarget) use NVIDIA GPUs.

Makefile.kokkos
seyonglee: The Clacc compiler is built on top of the LLVM OpenMP compiler, and thus it internally uses the existing LLVM OpenMP flags.

Makefile.kokkos
seyonglee: `KOKKOS_INTERNAL_USE_ARCH_AMD` is used by the OpenACC backend when targting AMD GPUs.

cmake/kokkos_arch.cmake
seyonglee: NVHPC installs the `nvcc` compiler in two locations: one is in `${NVHPCROOT}/cuda/bin` and the other is in `${NVHPCROOT}/compilers/bin`. 
As I know, for the CUDA backend, either path will be OK, but the OpenACC Clacc compiler should use the first path.
That's why a separate `CUDA_PATH` is used, but it is still possible to use the `CUDA_ROOT` returned by `FindCUDAToolkit` for the OpenACC Clacc compiler too.

cmake/kokkos_arch.cmake
seyonglee: As I understand, the ROCM module is loaded only for the HIP backend, but I don't know how to set cmake configurations so that it is also loaded when the OpenACC backend targets AMD GPUs; can you (or anyone) help on this?

core/unit_test/CMakeLists.txt
```diff
@@ -640,6 +641,8 @@ IF(KOKKOS_ENABLE_OPENACC AND KOKKOS_CXX_COMPILER_ID STREQUAL Clang)
     ${CMAKE_CURRENT_BINARY_DIR}/openacc/TestOpenACC_ViewMemoryAccessViolation.cpp
     ${CMAKE_CURRENT_BINARY_DIR}/openacc/TestOpenACC_WithoutInitializing.cpp
     ${CMAKE_CURRENT_BINARY_DIR}/openacc/TestOpenACC_ViewAPI_d.cpp
+	#Below test is disabled because it uses atomic operations not supported by Clacc.
+    ${CMAKE_CURRENT_BINARY_DIR}/openacc/TestOpenACC_ExecSpaceThreadSafety.cpp
```
dalg24, line 645:
```suggestion
    #Below test is disabled because it uses atomic operations not supported by Clacc.
    ${CMAKE_CURRENT_BINARY_DIR}/openacc/TestOpenACC_ExecSpaceThreadSafety.cpp
```

cmake/kokkos_arch.cmake
```diff
@@ -845,14 +845,35 @@ IF (KOKKOS_ENABLE_OPENACC)
...
   ELSE()
+    # When not compiling for offload to any GPU, we're compiling for kernel
+    # execution on the host.  In that case, memory is shared between the OpenACC
+    # space and the host space.
```
dalg24, line 868: I thought that this else clause meant that we may run anywhere?
Did I misunderstand?
If not should we then distinguish between forcing to run on host and the generic case?

Makefile.kokkos
dalg24, line 1: Makefile support is optional.  Do you really want it?

tpls/desul/include/desul/atomics/Macros.hpp
```diff
@@ -58,8 +58,11 @@ SPDX-License-Identifier: (BSD-3-Clause)
 #endif
 
 #if defined(DESUL_ATOMICS_ENABLE_OPENACC)
+#ifdef __NVCOMPILER
```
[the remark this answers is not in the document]
seyonglee: The intention is to use the OpenACC atomic implementation only when compiled by NVHPC. (Clacc compiler can compile the Kokkos OpenACC backend if the application does not use atomics.)

cmake/kokkos_arch.cmake
seyonglee: The intention is to follow the standard Kokkos style (decide the target device at compilation time).
If the target GPU architecture is explicitly specified (e.g., -DKokkos_ARCH_HOPPER90=ON), the Openacc backend will offload Kokkos parallel constructs to the target device. (Users can still offload the parallel constructs to the host CPU by inserting the `acc_set_device_type()` API in the application, but `acc_set_device_type() API is not Kokkos API and thus not portable.
If the target GPU architecture is NOT specified, the OpenACC backend will offload Kokkos parallel constructs to the host CPU as a multicore target.

Makefile.kokkos
seyonglee: Some Kokkos mini-apps that I'm using do not provide cmake compilation, and thus it will be better to keep the make compilation option for such applications.

tpls/desul/include/desul/atomics/Macros.hpp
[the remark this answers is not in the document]
dalg24: I am saying don't turn it ON here
https://github.com/kokkos/kokkos/blob/195f45f8f1c641f551bff791044eaf99c709a029/core/src/CMakeLists.txt#L28-L30
if you detect we are not using NVHPC and then you don't need to change anything in Desul.

cmake/kokkos_arch.cmake
dalg24: Does that mean you want to drop the logic that allows us to fallback to host?
https://github.com/kokkos/kokkos/blob/195f45f8f1c641f551bff791044eaf99c709a029/core/src/OpenACC/Kokkos_OpenACC.cpp#L49-L59

You could very well preserve that logic and just introduce a new OpenACC-only option to force using the host.

tpls/desul/include/desul/atomics/Macros.hpp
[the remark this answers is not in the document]
seyonglee: Updated; thanks!

core/src/CMakeLists.txt
```diff
@@ -26,7 +26,9 @@ IF (NOT desul_FOUND)
     SET(DESUL_ATOMICS_ENABLE_OPENMP ON) # not a typo Kokkos OpenMPTarget -> Desul OpenMP
   ENDIF()
   IF(KOKKOS_ENABLE_OPENACC)
-    SET(DESUL_ATOMICS_ENABLE_OPENACC ON)
+    IF(KOKKOS_CXX_COMPILER_ID STREQUAL NVHPC)
```
dalg24, line 29: You want a comment, something greppable so you remember to adjust whenever Clacc catches up?

cmake/kokkos_arch.cmake
seyonglee: It will be good to have a new OpenACC-only option to force using the host; @dalg24 can you suggest a good Kokkos CMake option name for that? (I'm not familiar with the Kokkos CMake option naming rule).

core/src/CMakeLists.txt
seyonglee: Added a comment.

tpls/desul/include/desul/atomics/Macros.hpp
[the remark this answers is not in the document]
dalg24: Plz revert this change then

cmake/kokkos_arch.cmake
dalg24: What about `Kokkos_ENABLE_OPENACC_FORCE_HOST_AS_DEVICE`?

cmake/kokkos_arch.cmake
dalg24: I think you would need to configure with -DKokkos_ENABLE_ROCM=ON to tell Kokkos to find it

dalg24 (commented):
We would probably need to update https://github.com/kokkos/kokkos/blob/195f45f8f1c641f551bff791044eaf99c709a029/core/src/OpenACC/Kokkos_OpenACC_Traits.hpp#L24-L34 as well

Makefile.kokkos
seyonglee: @rgayatri23: you are right; the flag was wrong (and redundant). Therefore, this flag is deleted.

cmake/kokkos_arch.cmake
seyonglee: Updated cmake/make configurations to use `Kokkos_ENABLE_OPENACC_FORCE_HOST_AS_DEVICE`

seyonglee:
> We would probably need to update
> 
> https://github.com/kokkos/kokkos/blob/195f45f8f1c641f551bff791044eaf99c709a029/core/src/OpenACC/Kokkos_OpenACC_Traits.hpp#L24-L34
> 
> as well

Updated.

cmake/kokkos_enable_options.cmake
```diff
@@ -71,6 +71,7 @@ KOKKOS_ENABLE_OPTION(AGGRESSIVE_VECTORIZATION OFF "Whether to aggressively vecto
 KOKKOS_ENABLE_OPTION(COMPILE_AS_CMAKE_LANGUAGE OFF "Whether to use native cmake language support")
 KOKKOS_ENABLE_OPTION(HIP_MULTIPLE_KERNEL_INSTANTIATIONS OFF "Whether multiple kernels are instantiated at compile time - improve performance but increase compile time")
 KOKKOS_ENABLE_OPTION(IMPL_HIP_UNIFIED_MEMORY OFF "Whether to leverage unified memory architectures for HIP")
+KOKKOS_ENABLE_OPTION(OPENACC_FORCE_HOST_AS_DEVICE OFF "Whether to force to use host as a target device for OpenACC")
```
dalg24, line 74: Plz add this L142
```cmake
CHECK_DEVICE_SPECIFIC_OPTIONS(DEVICE OPENACC OPTIONS OPENACC_FORCE_HOST_AS_DEVICE)

File: core/src/OpenACC/Kokkos_OpenACC.cpp [hunk 13]
```diff
@@ -59,11 +70,52 @@ void Kokkos::Experimental::OpenACC::impl_initialize(
...
   } else {
     using Kokkos::Impl::get_visible_devices;
+    acc_set_device_type(Impl::OpenACC_Traits::dev_type);
     std::vector<int> const& visible_devices = get_visible_devices();
     using Kokkos::Impl::get_gpu;
     int const dev_num = get_gpu(settings).value_or(visible_devices[0]);
     acc_set_device_num(dev_num, Impl::OpenACC_Traits::dev_type);
     Impl::OpenACCInternal::m_acc_device_num = dev_num;
+#if defined(KOKKOS_IMPL_ARCH_NVIDIA_GPU)
+    struct cudaDeviceProp deviceProp;
```
dalg24, line 80:
```suggestion
    cudaDeviceProp deviceProp;
```

core/src/OpenACC/Kokkos_OpenACC.cpp
```diff
@@ -59,11 +70,52 @@ void Kokkos::Experimental::OpenACC::impl_initialize(
...
+    }
+    Impl::OpenACCInternal::m_concurrency =
+        deviceProp.maxThreadsPerMultiProcessor * deviceProp.multiProcessorCount;
+#elif defined(KOKKOS_ENABLE_OPENACC_FORCE_HOST_AS_DEVICE)
+    Impl::OpenACCInternal::m_concurrency = std::thread::hardware_concurrency();
+    if (!Impl::OpenACCInternal::m_concurrency) {
+      Kokkos::Impl::host_abort(
+          "Error: During OpenACC backend initialization, failed to retrieve "
+          "CPU "
+          "hardware concurrency");
```
dalg24, line 109:
```suggestion
          "CPU hardware concurrency");
```

core/src/OpenACC/Kokkos_OpenACC.cpp
```diff
@@ -59,11 +70,52 @@ void Kokkos::Experimental::OpenACC::impl_initialize(
...
+      msg << "Error: During OpenACC backend initialization, failed to retrieve "
+          << "HIP device properties: (" << hipGetErrorName(error)
+          << "): " << hipGetErrorString(error);
+      Kokkos::Impl::host_abort(msg.str().c_str());
+    }
+    Impl::OpenACCInternal::m_concurrency =
+        deviceProp.maxThreadsPerMultiProcessor * deviceProp.multiProcessorCount;
+#elif defined(KOKKOS_ENABLE_OPENACC_FORCE_HOST_AS_DEVICE)
+    Impl::OpenACCInternal::m_concurrency = std::thread::hardware_concurrency();
+    if (!Impl::OpenACCInternal::m_concurrency) {
```
dalg24, line 105: I would prefer
```suggestion
    if (Impl::OpenACCInternal::m_concurrency == 0) {
```

core/src/OpenACC/Kokkos_OpenACC.cpp
```diff
@@ -59,11 +70,52 @@ void Kokkos::Experimental::OpenACC::impl_initialize(
...
+          "CPU "
+          "hardware concurrency");
+    }
+#else
+    // FIXME_OPENACC: Compute Impl::OpenACCInternal::m_concurrency correctly.
+    if (show_warnings()) {
+      std::cerr << "Warning: OpenACC backend concurrency detection not "
+                   "implemented for target architecture"
+                << std::endl;
+    }
```
dalg24, line 117: Please confirm that you want to start raising warning in that case

core/src/OpenACC/Kokkos_OpenACC.cpp
```diff
@@ -59,11 +70,52 @@ void Kokkos::Experimental::OpenACC::impl_initialize(
         acc_get_device_num(acc_device_host);
   } else {
     using Kokkos::Impl::get_visible_devices;
+    acc_set_device_type(Impl::OpenACC_Traits::dev_type);
```
dalg24, line 73: This actually would break the falling back to host logic

cmake/kokkos_enable_options.cmake
seyonglee: Added.

core/src/OpenACC/Kokkos_OpenACC.cpp
seyonglee: The falling back to host logic is handled in line 60, and thus this will not break it.
The `else` part in line 71 ~ 118 handles the case where device type is set to NVIDIA/AMD GPUs, host CPU, or unknown accelerator (`acc_device_not_host`).

core/src/OpenACC/Kokkos_OpenACC.cpp
seyonglee: This part is for an unknown accelerator type (`acc_device_not_host`) where we don't know the concurrency.

core/src/OpenACC/Kokkos_OpenACC_Traits.hpp
```diff
@@ -23,14 +23,25 @@ namespace Kokkos::Experimental::Impl {
 
 struct OpenACC_Traits {
 #if defined(KOKKOS_IMPL_ARCH_NVIDIA_GPU)
-  static constexpr acc_device_t dev_type     = acc_device_nvidia;
-  static constexpr bool may_fallback_to_host = false;
+  static constexpr acc_device_t dev_type            = acc_device_nvidia;
+  static constexpr bool may_fallback_to_host        = false;
+  static constexpr OpenACCSpace::size_type WarpSize = 32;
```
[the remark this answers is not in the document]
seyonglee: The next PR will include codes using this.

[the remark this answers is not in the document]
dalg24: Include it then.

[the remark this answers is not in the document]
dalg24: (By that I mean please remove it and introduce it back when you need it)

core/src/OpenACC/Kokkos_OpenACC.cpp
dalg24: I understand that this is for the case where we use that 256000 fallback value.
I am just wondering if that is excessive and we would now be "spamming" the user about something that is not unexpectedly happening at runtime but that we can "diagnose" at configuration time and only warn then.
I don't feel too strongly about it if you want it though. Just observing that this somewhat departs from the usual situations in which we currently decide to warn users in the rest of the codebase.

core/src/OpenACC/Kokkos_OpenACC.cpp
dalg24: I am not familiar with the spec but I expected that if we call `acc_set_device_type(acc_device_not_host)` the runtime will error out if there is no device available for execution?
I that not how it works?  Is the runtime still allowed to fallback to host in that case?

core/src/OpenACC/Kokkos_OpenACCSpace.hpp
```diff
@@ -85,16 +85,26 @@ class OpenACCSpace {
 template <>
 struct Kokkos::Impl::MemorySpaceAccess<Kokkos::HostSpace,
                                        Kokkos::Experimental::OpenACCSpace> {
+#if defined(KOKKOS_ENABLE_OPENACC_FORCE_HOST_AS_DEVICE)
+  enum : bool{assignable = true};
+  enum : bool{accessible = true};
+#else
```
masterleinad, line 91: Do you really want to allow accessing `OpenACCSpace` from the host? This will likely break code when you are first forcing OpenACC to target the host and later a proper device.

cmake/kokkos_arch.cmake
seyonglee: @masterleinad Can you help on this (separating the Cuda backend from using FindCUDAToolkit)? I don't fully understand how the Kokkos CMake handles the CUDA backend.

Makefile.kokkos
masterleinad: Isn't that problematic?

core/src/OpenACC/Kokkos_OpenACC_Traits.hpp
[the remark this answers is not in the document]
seyonglee: Deleted as suggested.

Makefile.kokkos
seyonglee: I don't think that will be problematic; what I did here is just to fix an error in setting `KOKKOS_INTERNAL_COMPILER_CRAY_CLANG`.

Makefile.kokkos
```diff
@@ -282,6 +286,8 @@ ifeq ($(KOKKOS_INTERNAL_USE_OPENACC), 1)
   # Set OpenACC flags.
   ifeq ($(KOKKOS_INTERNAL_COMPILER_NVHPC), 1)
     KOKKOS_INTERNAL_OPENACC_FLAG := -acc
+  else ifeq ($(KOKKOS_INTERNAL_COMPILER_CLANG), 1)
+    KOKKOS_INTERNAL_OPENACC_FLAG := -fopenacc -fopenacc-fake-async-wait -fopenacc-implicit-worker=vector -Wno-openacc-and-cxx -Wno-openmp-mapping -Wno-unknown-cuda-version -Wno-pass-failed
```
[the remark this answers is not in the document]
seyonglee: There is no good user manual, but you can find the basic information on these Clacc compiler options from the READM file in the GitHub Clacc repository (https://github.com/llvm-doe-org/llvm-project/tree/clacc/main).

cmake/kokkos_arch.cmake
seyonglee: Just adding `Kokkos_ENABLE_ROCM=ON` does not seem to work. (cmake/Modules/FindTPLROCM.cmake does not set include path.)

core/src/OpenACC/Kokkos_OpenACC.cpp
seyonglee: You are right, `acc_set_device_type(acc_device_not_host)` will issue an error if no device available for execution. 
In that sense, it will be better to use `acc_device_default` so that we can you whatever available in the target system. (changed the `OpenACC_Traits::dev_type` to `acc_device_default`)

cmake/kokkos_arch.cmake
dalg24: You would still have to register that your target depend on it.

core/src/OpenACC/Kokkos_OpenACC.cpp
```diff
@@ -46,6 +56,7 @@ Kokkos::Experimental::OpenACC::OpenACC(int async_arg)
 
 void Kokkos::Experimental::OpenACC::impl_initialize(
     InitializationSettings const& settings) {
+  Impl::OpenACCInternal::m_concurrency = 256000;  // guess when cannot compute
```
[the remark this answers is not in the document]
seyonglee: It is a random guess; a comment is added.

core/src/OpenACC/Kokkos_OpenACCSpace.hpp
seyonglee: This is based on an assumption that if a user explicitly specify `KOKKOS_ENABLE_OPENACC_FORCE_HOST_AS_DEVICE` but without a NVIDIA/AMD GPU architecture (`KOKKOS_ARCH`), all the Kokkos parallel constructs will be executed on the host CPU. If the user wants to use both CPU and GPU as offloading targets, the user should not specify `KOKKOS_ENABLE_OPENACC_FORCE_HOST_AS_DEVICE`.

core/src/OpenACC/Kokkos_OpenACC.cpp
seyonglee: Removed the warning message; instead added a message in the CMake configuration (Kokkos_arch.cmake) regarding the automatic fallback mode.

core/src/OpenACC/Kokkos_OpenACC.cpp
```diff
@@ -23,7 +23,17 @@
 #include <impl/Kokkos_ExecSpaceManager.hpp>
 #include <impl/Kokkos_DeviceManagement.hpp>
 
+#if defined(KOKKOS_IMPL_ARCH_NVIDIA_GPU)
+#include <cuda_runtime.h>
+#elif defined(KOKKOS_ARCH_AMD_GPU)
+#define __HIP_PLATFORM_AMD__
```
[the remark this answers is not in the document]
seyonglee: `hip_runtime_api.h` contains two implementations: one for AMD GPUs and the other for NVIDIA GPUs; the macro, `__HIP_PLATFORM_AMD__` is needed to choose AMD GPUs. (We put it here since we don't know what kind of side effects it will have if it becomes a global macro.)
Added a comment.

Makefile.kokkos
```diff
@@ -401,13 +407,16 @@ ifeq ($(KOKKOS_INTERNAL_USE_ARCH_NVIDIA), 0)
...
   ifeq ($(KOKKOS_INTERNAL_COMPILER_CLANG), 1)
-    KOKKOS_INTERNAL_NVCC_PATH := $(shell which nvcc)
     CUDA_PATH ?= $(KOKKOS_INTERNAL_NVCC_PATH:/bin/nvcc=)
     ifeq ($(KOKKOS_INTERNAL_USE_OPENMPTARGET), 1)
       KOKKOS_INTERNAL_OPENMPTARGET_FLAG := $(KOKKOS_INTERNAL_OPENMPTARGET_FLAG) --cuda-path=$(CUDA_PATH)
     endif
   endif
+  ifeq ($(KOKKOS_INTERNAL_COMPILER_NVHPC), 1)
+    CUDA_PATH ?= $(KOKKOS_INTERNAL_NVCC_PATH:/bin/nvcc=)
+  endif
```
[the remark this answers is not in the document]
seyonglee: `CUDA_PATH` will be needed for `nvc++` if it is used for compile the OpenACC backend.
(As you said, this change is not for `clacc`; should it be deleted for now?)

cmake/kokkos_arch.cmake
seyonglee: How to register it? (FindTPLROCM.cmake also needs to be updated to include the path for the hip runtime header.)

core/src/OpenACC/Kokkos_OpenACC_Traits.hpp
```diff
@@ -28,8 +28,14 @@ struct OpenACC_Traits {
 #elif defined(KOKKOS_ARCH_AMD_GPU)
   static constexpr acc_device_t dev_type     = acc_device_radeon;
   static constexpr bool may_fallback_to_host = false;
+#if defined(KOKKOS_ARCH_AMD_GFX1030) || defined(KOKKOS_ARCH_AMD_GFX1100)
 #else
-  static constexpr acc_device_t dev_type     = acc_device_not_host;
+#endif
```
dalg24, line 33: Remove L31-33

simd/unit_tests/include/TestSIMD_MathOps.hpp
```diff
@@ -316,6 +316,14 @@ TEST(simd, device_math_ops) {
       << "skipping because of a non-deterministic failure reporting: "
          "Failure to synchronize stream (nil): Error in "
          "cuStreamSynchronize: an illegal memory access was encountered";
+#endif
+#ifdef KOKKOS_ENABLE_OPENACC
+#ifdef KOKKOS_COMPILER_CLANG  // FIXME_CLACC
```
dalg24, line 321: I would prefer
```suggestion
#if defined(KOKKOS_ENABLE_OPENACC) && defined(KOKKOS_COMPILER_CLANG)  // FIXME_CLACC
```
(not formatted and need to delete `#endif`)
Same comment in the next file.

core/unit_test/TestAbort.hpp
```diff
@@ -103,5 +103,12 @@ void test_abort_from_device() {
 
 TEST(TEST_CATEGORY_DEATH, abort_from_device) {
   ::testing::FLAGS_gtest_death_test_style = "threadsafe";
+// FIXME_OPENACC FIXME_NVHPC: NVHPC fails when targetting CPUs.
+#if defined(KOKKOS_ENABLE_OPENACC) && defined(KOKKOS_COMPILER_NVHPC) && \
+    defined(KOKKOS_ENABLE_OPENACC_FORCE_HOST_AS_DEVICE)
+  GTEST_SKIP()
```
dalg24, line 109: I think we would want the
```C++
if (std::is_same_v<TEST_EXECSPACE, Kokkos::Experimental::OpenACC>)
```
because we still want to test the host backends regardless that the ACC backend is targeting the host

core/unit_test/TestExecSpaceThreadSafety.hpp
```diff
@@ -78,9 +78,11 @@ void run_exec_space_thread_safety_range() {
 
 TEST(TEST_CATEGORY, exec_space_thread_safety_range) {
 #ifdef KOKKOS_ENABLE_OPENACC  // FIXME_OPENACC
+#ifdef KOKKOS_ENABLE_OPENACC_FORCE_HOST_AS_DEVICE
```
dalg24, line 81: Please confirm that this means that the test works EXCEPT when targeting the host with the OpenACC backend.
Meaning we could have gotten of these guards if it wasn't for that new option we introduce.

core/src/OpenACC/Kokkos_OpenACC_Traits.hpp
seyonglee: Removed.

core/unit_test/TestAbort.hpp
seyonglee: Added.

core/unit_test/TestExecSpaceThreadSafety.hpp
seyonglee: Yes, the tests work except when the OpenACC backend targets the host.

simd/unit_tests/include/TestSIMD_MathOps.hpp
seyonglee: Updated TestSIMD_Reductions.hpp too.

Makefile.kokkos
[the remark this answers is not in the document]
seyonglee: In fact, `CUDA_PATH` may not be necessary for `nvc++` if `-cuda` option is used instead.
(Deleted the above three lines and added `-cuda` to `nvc++`)

core/unit_test/TestAbort.hpp
```diff
@@ -103,5 +103,14 @@ void test_abort_from_device() {
 
 TEST(TEST_CATEGORY_DEATH, abort_from_device) {
   ::testing::FLAGS_gtest_death_test_style = "threadsafe";
+// FIXME_OPENACC FIXME_NVHPC: NVHPC fails when targetting CPUs.
+#if defined(KOKKOS_ENABLE_OPENACC) && defined(KOKKOS_COMPILER_NVHPC) && \
+    defined(KOKKOS_ENABLE_OPENACC_FORCE_HOST_AS_DEVICE)
+  if (std::is_same<ExecutionSpace, Kokkos::Experimental::OpenACC>::value) {
```
dalg24, line 109: Will conflict with #7227
Maybe use is_same_v

cmake/kokkos_arch.cmake
```diff
@@ -845,15 +845,42 @@ IF (KOKKOS_ENABLE_OPENACC)
...
+    )
+    COMPILER_SPECIFIC_DEFS(
+      Clang KOKKOS_ENABLE_OPENACC_FORCE_HOST_AS_DEVICE
+    )
   ELSE()
+    # Automatic fallback mode; try to offload any available GPU, and fall back
+    # to the host CPU if no available GPU is found.
     COMPILER_SPECIFIC_FLAGS(
-      NVHPC -acc
+      NVHPC -acc=gpu,multicore
```
dalg24, line 881: Presumably this change is not actually needed per se.
Would you explain why it is preferable to specify the target architectures?
Is that a Clacc thing?
(I am not necessary objecting to the change, just wanting to make sure I understand)

cmake/kokkos_arch.cmake
```diff
@@ -845,15 +845,42 @@ IF (KOKKOS_ENABLE_OPENACC)
...
+    # memory is shared between the OpenACC space and the host space.
+    COMPILER_SPECIFIC_FLAGS(
+      NVHPC -acc=multicore
+    )
+    COMPILER_SPECIFIC_DEFS(
+      NVHPC KOKKOS_ENABLE_OPENACC_FORCE_HOST_AS_DEVICE
+    )
+    COMPILER_SPECIFIC_DEFS(
+      Clang KOKKOS_ENABLE_OPENACC_FORCE_HOST_AS_DEVICE
+    )
```
dalg24, line 876: Remove and add them to `cmake/KokkosCore_config.h.in` instead

core/unit_test/TestAbort.hpp
seyonglee: Updated to `is_same_v`

cmake/kokkos_arch.cmake
seyonglee: `KOKKOS_ENABLE_OPENACC_FORCE_HOST_AS_DEVICE` is supposed to be enabled only when none of NVIDIA and AMD GPU architecture is specified. If we add it to `cmake/kokkosCore_config.h.in`, it will be defined regardless of GPU architecture setting.

cmake/kokkos_arch.cmake
seyonglee: The behavior of `-acc` is not the same as that of `-acc=gpu,multicore`; when host fallback occurs, with `-acc`, OpenACC kernels are executed serially on the host CPU while executed in parallel with `-acc=gpu,multicore`. (GPU execution behavior is the same.)

core/unit_test/TestAbort.hpp
dalg24: I meant just the occurrence you added not the other two that were already there

core/unit_test/TestAbort.hpp
dalg24: I meant just this occurrence that was added, not the other two that were already there

cmake/kokkos_arch.cmake
dalg24: Then you could/should have that cmakedefine in the config file and error out at configuration time if both an NVIDIA/AMD GPU arch and that FORCE_HOST_AS_DEVICE option are enabled

core/unit_test/TestAbort.hpp
```diff
@@ -75,13 +75,13 @@ struct TestAbortCausingAbnormalProgramTerminationAndPrinting {
...
 #if defined(KOKKOS_ENABLE_OPENMPTARGET)  // FIXME_OPENMPTARGET
-  if (std::is_same<ExecutionSpace, Kokkos::Experimental::OpenMPTarget>::value) {
+  if (std::is_same_v<ExecutionSpace, Kokkos::Experimental::OpenMPTarget>) {
     TestAbortPrintingToStdout<ExecutionSpace>();
   } else {
     TestAbortCausingAbnormalProgramTerminationAndPrinting<ExecutionSpace>();
   }
 #elif defined(KOKKOS_ENABLE_OPENACC)  // FIXME_OPENACC
-  if (std::is_same<ExecutionSpace, Kokkos::Experimental::OpenACC>::value) {
+  if (std::is_same_v<ExecutionSpace, Kokkos::Experimental::OpenACC>) {
```
dalg24, line 84: I'd recommend undoing that change and the one above to reduce conflicts

core/unit_test/TestAbort.hpp
seyonglee: Undo the changes for the other two already-existing ones.

cmake/kokkos_arch.cmake
seyonglee: Updated CMake/Make so that it will be added to `KokkosCore_config.h` when specified. 
CMake will print a warning message if both `KOKKOS_ENABLE_OPENACC_FORCE_HOST_AS_DEVICE` and GPU architecture are specified, and the OpenACC backend will ignore  `KOKKOS_ENABLE_OPENACC_FORCE_HOST_AS_DEVICE` in that case.

Makefile.kokkos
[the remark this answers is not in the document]
crtrott: so this is obviously not for main-line clang. Are you saying these also work for cray-clang or just for Clacc?

seyonglee:
> I still don't get fully why we need that force host option, in particular if it's ignored if you set a GPU Arch. Is it that you are saying without setting it the OpenACC would at runtime discover a GPU if you didn't set an arch, and start using that - i.e. you want something like a "JIT-GPU-Arch" capability preserved?

Yes, we want to keep the "JIT-GPU-Arch" capability (OpenACC backend automatically chooses the target GPU) as an available option but not the default option, so that users can choose between targeting host CPU and letting the OpenACC backend to choose the target device.

Makefile.kokkos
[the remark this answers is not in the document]
seyonglee: These options work only for Clacc.

Makefile.kokkos
seyonglee: `-craype-verbose` is an Cray-Clang-specific option; running this command will cause an error if non-Cray Clang compiler (e.g., Clacc) is used.

cmake/kokkos_arch.cmake
```diff
@@ -838,22 +838,49 @@ ENDIF()
 
 IF (KOKKOS_ENABLE_OPENACC)
   IF(KOKKOS_CUDA_ARCH_FLAG)
+    IF(KOKKOS_ENABLE_OPENACC_FORCE_HOST_AS_DEVICE)
+      MESSAGE(WARNING "If a GPU architecture is specified, Kokkos_ENABLE_OPENACC_FORCE_HOST_AS_DEVICE option will be ignored.")
```
dalg24, line 842: I think that should be a hard error (same comment in the AMDGPU branch)

core/src/OpenACC/Kokkos_OpenACCSpace.hpp
```diff
@@ -85,16 +85,28 @@ class OpenACCSpace {
 template <>
 struct Kokkos::Impl::MemorySpaceAccess<Kokkos::HostSpace,
                                        Kokkos::Experimental::OpenACCSpace> {
+#if defined(KOKKOS_ENABLE_OPENACC_FORCE_HOST_AS_DEVICE) && \
+    !defined(KOKKOS_IMPL_ARCH_NVIDIA_GPU) && !defined(KOKKOS_ARCH_AMD_GPU)
```
dalg24, line 89: Let's only guard with `#ifdef KOKKOS_ENABLE_OPENACC_FORCE_HOST_AS_DEVICE` and make that incompatible with a NVIDIA or AMD gpu architecture specified.
(same comment below)

core/src/OpenACC/Kokkos_OpenACC.cpp
```diff
@@ -86,6 +135,13 @@ void Kokkos::Experimental::OpenACC::print_configuration(std::ostream& os,
   os << "yes\n";
 #else
   os << "no\n";
+#endif
+  os << "  KOKKOS_ENABLE_OPENACC_FORCE_HOST_AS_DEVICE: ";
+#if defined(KOKKOS_ENABLE_OPENACC_FORCE_HOST_AS_DEVICE) && \
+    !defined(KOKKOS_IMPL_ARCH_NVIDIA_GPU) && !defined(KOKKOS_ARCH_AMD_GPU)
```
dalg24, line 141:
```suggestion
#ifdef KOKKOS_ENABLE_OPENACC_FORCE_HOST_AS_DEVICE
```

dalg24 (commented):
Haven't looked at the makefile but I am basically OK with this PR if we make `OPENACC_FORCE_HOST_AS_DEVICE` incompatible with a GPU arch being specified (i.e. error at configuration time instead of a warning and simpler guards in the code base)

cmake/kokkos_arch.cmake
seyonglee: Changed.

core/src/OpenACC/Kokkos_OpenACC.cpp
seyonglee: Applied.

seyonglee:
> Haven't looked at the makefile but I am basically OK with this PR if we make `OPENACC_FORCE_HOST_AS_DEVICE` incompatible with a GPU arch being specified (i.e. error at configuration time instead of a warning and simpler guards in the code base)

Changed both CMake and Make configurations to issue configuration errors if both are specified.

dalg24 (approved):
Approving the non-Makefile part

Makefile.kokkos
crtrott: So would this flag now set to be to true if you use Clacc? Or was it set to true without the filtering out the error?

seyonglee: In old Makefile.kokkos, `KOKKOS_INTERNAL_COMPILER_CRAY_CLANG  := $(strip $(shell $(CXX) -craype-verbose 2>&1 | grep -c "clang++"))` sets `KOKKOS_INTERNAL_COMPILER_CRAY_CLANG` to `2` if non-Cray-Clang compilers (e.g., Clacc) are used. However, in later parts of the Makefile, `KOKKOS_INTERNAL_COMPILER_CRAY_CLANG` is used in two cases: 1) true case (`ifeq ($(KOKKOS_INTERNAL_COMPILER_CRAY_CLANG), 1)`) and 2) false case (`ifeq ($(KOKKOS_INTERNAL_COMPILER_CRAY_CLANG), 0)`). However, the false case will not work as intended since the value of `KOKKOS_INTERNAL_COMPILER_CRAY_CLANG` becomes `2` instead of `0` for non-Cray-Clang compilers. This change set `KOKKOS_INTERNAL_COMPILER_CRAY_CLANG` to 0 if non-Cray-Clang compiler is used.

cmake/kokkos_arch.cmake
masterleinad: We can try to figure this out in a follow-up pull request.

Makefile.kokkos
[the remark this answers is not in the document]
crtrott: I guess we don't care to reuse Clacc here since it's the only compiler identified as clang which supports OpenACC anyway.

Makefile.kokkos
```diff
@@ -1118,41 +1175,49 @@ ifeq ($(KOKKOS_INTERNAL_USE_ARCH_AMD_GFX906), 1)
...
   tmp := $(call kokkos_append_header,"$H""define KOKKOS_ARCH_AMD_GFX1100")
   tmp := $(call kokkos_append_header,"$H""define KOKKOS_ARCH_AMD_GPU")
   KOKKOS_INTERNAL_HIP_ARCH_FLAG := --offload-arch=gfx1100
+  KOKKOS_INTERNAL_AMD_ARCH_FLAG := $(KOKKOS_INTERNAL_AMD_ARCH_FLAG)=gfx1100
 endif
 ifeq ($(KOKKOS_INTERNAL_USE_ARCH_AMD_GFX1103), 1)
   tmp := $(call kokkos_append_header,"$H""define KOKKOS_ARCH_AMD_GFX1103")
   tmp := $(call kokkos_append_header,"$H""define KOKKOS_ARCH_AMD_GPU")
   KOKKOS_INTERNAL_HIP_ARCH_FLAG := --offload-arch=gfx1103
+  KOKKOS_INTERNAL_AMD_ARCH_FLAG := $(KOKKOS_INTERNAL_AMD_ARCH_FLAG)=gfx1103
```
crtrott, line 1220: Couldn't we combine this? Just make KOKKOS_INTERNAL_AMD_ARCH_FLAG=--offload-arch when HIP is enabled? And then use KOKKOS_INTERNAL_AMD_ARCH_FLAG for OpenACC, OpenMP-Offload and HIP?

seyonglee: You are right; we can use `KOKKOS_INTERNAL_AMD_ARCH_FLAG` for OpenACC, HIP, and OpenMPTarget.
I updated `Makefile.kokkos` for both OpenACC and HIP backends to use `KOKKOS_INTERNAL_AMD_ARCH_FLAG`.
(I didn't change the OpenMPTarget backend configuration since the OpenMPTarget backend does not seem to support AMD GPUs (there were several compilation errors.))

crtrott (approved):

dalg24:
One CUDA build failed because the GPU was not detected at runtime.  Looks good.

dalg24:
Make sure you add to the 4.5 change log

ndellingwood:
Looks like this still needs a changelog entry

seyonglee:
> Looks like this still needs a changelog entry

Updated.

PR #7210, opened 2024-08-09
Pin dependencies in Dockerfiles

dalg24:
Drive-by fix in one of the GitHub Actions that raised a warning in our last OpenSSF Scorecard report.

Pin all Docker images to increase our score and pickup good habits.
Switch to NVIDIA registry for all CUDA images.

Also computed the checksum of the oneAPI binaries we host on our server and added a check.  This was showing up in the reports.


masterleinad (approved):

aprokop (approved):

dalg24:
One unrelated failure at runtime (GPU not detected)

PR #7211, opened 2024-08-09
OpenMPTarget: FunctorAdapter bug in AMD GPU code path.

rgayatri23:
The PR fixes a bug introduced in #7200


masterleinad (approved):

Rombur (approved):
This PR looks good but I am not convinced about your guard for the AMD architectures. I am guessing you don't (want to) support these architectures but with the current guards, we go through the `#else`.

rgayatri23:
> This PR looks good but I am not convinced about your guard for the AMD architectures. I am guessing you don't (want to) support these architectures but with the current guards, we go through the `#else`.

I think this led to some memory issues at the time we put the guard in. I do not have access to those architectures to test now.

dalg24:
Failures were at runtime (GPU not detected)

PR #7217, opened 2024-08-12
Cuda: Check if device support cudaMallocAsync

### LINKED ISSUE #7044: Check for cudaMallocAsync support before using it
Currently, we are always using `cudaMallocAsync` for allocating memory if the compiler is recent enough (and the user doesn't forcefully disable it). As pointed out in https://github.com/kokkos/kokkos-core-wiki/pull/525#discussion_r1615644650, this is not quite enough as there still might be GPUs that don't support memory pools (independent of the compiler version).
Thus, we should check `cudaDevAttrMemoryPoolsSupported` when initializing the execution space instance to decide if `cudaMallocAsync` can be used.

masterleinad:
Fixes #7044. According to https://developer.nvidia.com/blog/cuda-pro-tip-the-fast-way-to-query-device-properties, this should be fast so checking for every allocation should not be noticeable.

### DIFF
--- a/core/src/Cuda/Kokkos_CudaSpace.cpp
+++ b/core/src/Cuda/Kokkos_CudaSpace.cpp
@@ -179,6 +179,7 @@ void *impl_allocate_common(const int device_id,
 #ifndef CUDART_VERSION
 #error CUDART_VERSION undefined!
 #elif (defined(KOKKOS_ENABLE_IMPL_CUDA_MALLOC_ASYNC) && CUDART_VERSION >= 11020)
+  // FIXME_KEPLER Everything after Kepler should support cudaMallocAsync
   int device_supports_cuda_malloc_async;
   KOKKOS_IMPL_CUDA_SAFE_CALL(
       cudaDeviceGetAttribute(&device_supports_cuda_malloc_async,
--- a/core/src/Cuda/Kokkos_CudaSpace.cpp
+++ b/core/src/Cuda/Kokkos_CudaSpace.cpp
@@ -185,7 +185,7 @@ void *impl_allocate_common(const int device_id,
                              cudaDevAttrMemoryPoolsSupported, device_id));
 
   if (arg_alloc_size >= memory_threshold_g &&
-      device_supports_cuda_malloc_async) {
+      device_supports_cuda_malloc_async == 1) {
     error_code = cudaMallocAsync(&ptr, arg_alloc_size, stream);
 
     if (error_code == cudaSuccess) {
--- a/core/src/Cuda/Kokkos_CudaSpace.cpp
+++ b/core/src/Cuda/Kokkos_CudaSpace.cpp
@@ -179,7 +179,13 @@ void *impl_allocate_common(const int device_id,
 #ifndef CUDART_VERSION
 #error CUDART_VERSION undefined!
 #elif (defined(KOKKOS_ENABLE_IMPL_CUDA_MALLOC_ASYNC) && CUDART_VERSION >= 11020)
-  if (arg_alloc_size >= memory_threshold_g) {
+  int device_supports_cuda_malloc_async;
+  KOKKOS_IMPL_CUDA_SAFE_CALL(
+      cudaDeviceGetAttribute(&device_supports_cuda_malloc_async,
+                             cudaDevAttrMemoryPoolsSupported, device_id));
+
+  if (arg_alloc_size >= memory_threshold_g &&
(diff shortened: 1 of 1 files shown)


ldh4 (approved):
Looks good to me, but have you tried running this with `-DKokkos_ENABLE_IMPL_CUDA_MALLOC_ASYNC=ON` on a machine where `cudaMallocAsync` is disabled by any chance? Just to check if querying for `cudaDevAttrMemoryPoolsSupported` returns as expected.

core/src/Cuda/Kokkos_CudaSpace.cpp
```diff
@@ -179,7 +179,13 @@ void *impl_allocate_common(const int device_id,
...
 #error CUDART_VERSION undefined!
 #elif (defined(KOKKOS_ENABLE_IMPL_CUDA_MALLOC_ASYNC) && CUDART_VERSION >= 11020)
-  if (arg_alloc_size >= memory_threshold_g) {
+  int device_supports_cuda_malloc_async;
+  KOKKOS_IMPL_CUDA_SAFE_CALL(
+      cudaDeviceGetAttribute(&device_supports_cuda_malloc_async,
+                             cudaDevAttrMemoryPoolsSupported, device_id));
+
+  if (arg_alloc_size >= memory_threshold_g &&
+      device_supports_cuda_malloc_async) {
```
dalg24, line 188:
```suggestion
      device_supports_cuda_malloc_async == 1) {
```

core/src/Cuda/Kokkos_CudaSpace.cpp
```diff
@@ -179,7 +179,13 @@ void *impl_allocate_common(const int device_id,
 #ifndef CUDART_VERSION
 #error CUDART_VERSION undefined!
 #elif (defined(KOKKOS_ENABLE_IMPL_CUDA_MALLOC_ASYNC) && CUDART_VERSION >= 11020)
-  if (arg_alloc_size >= memory_threshold_g) {
+  int device_supports_cuda_malloc_async;
```
dalg24, line 182: Do we have a feel for what GPUs do not support it?
Should we be planning for dropping that check when/if it becomes unnecessary?

masterleinad:
> Looks good to me, but have you tried running this with `-DKokkos_ENABLE_IMPL_CUDA_MALLOC_ASYNC=ON` on a machine where `cudaMallocAsync` is disabled by any chance? Just to check if querying for `cudaDevAttrMemoryPoolsSupported` returns as expected.

No, I don't know which machine wouldn't allow it. Everything we are testing with certainly does.

core/src/Cuda/Kokkos_CudaSpace.cpp
masterleinad: According to https://github.com/kokkos/kokkos-core-wiki/pull/525#discussion_r1603636387, it's difficult to know.

core/src/Cuda/Kokkos_CudaSpace.cpp
masterleinad: I didn't compare with 1 because the result is logically a `bool` (but the interface always returns an int for all attributes).

dalg24: From the [doc](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DEVICE.html#group__CUDART__DEVICE_1gb22e8256592b836df9a9cc36c9db7151)
> [cudaDevAttrMemoryPoolsSupported](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__TYPES.html#group__CUDART__TYPES_1gg49e2f8c2c0bd6fe264f2fc970912e5cd0876515d3b274cb0b40f8a7d913fb1aa): 1 if the device supports using the cudaMallocAsync and cudaMemPool family of APIs, and 0 otherwise

dalg24:
Retest this please

core/src/Cuda/Kokkos_CudaSpace.cpp
dalg24: Maybe say we can probably get rid of it next time we bump our requirements and drop support for Kepler

masterleinad: Sure

dalg24 (approved):

dalg24:
Ignoring OpenMPTarget failures at runtime (GPU not detected)

PR #7219, opened 2024-08-13
Improve View for non-scalar trivial types

dalg24:
As [cppreference points out](https://en.cppreference.com/w/cpp/types/is_destructible#Notes)
> Storage occupied by [trivially destructible](https://en.cppreference.com/w/cpp/language/destructor#Trivial_destructor) objects [may be reused](https://en.cppreference.com/w/cpp/language/lifetime#Storage_reuse) without calling the destructor.

so we can "extend" the specialization for scalars to trivial types.

This has two benefits:
1. it improves performance (skip a fence and a kernel launch to delete the elements in the view)
2. it let us get rid of more constraints on `Impl::ViewValueFunctor`'s member function which simplifies the code

### DIFF
--- a/core/src/View/Kokkos_ViewAlloc.hpp
+++ b/core/src/View/Kokkos_ViewAlloc.hpp
@@ -236,8 +236,7 @@ struct ViewValueFunctor<DeviceType, ValueType, true /* is_scalar */> {
         name(std::move(arg_name)),
         default_exec_space(true) {}
 
-  template <typename Dummy = ValueType>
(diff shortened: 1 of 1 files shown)


nliber (approved):
I agree with this direction.  `is_trivial` is a generalization of `is_scalar` and is the property we care about here.

nmm0 (approved):
Looks good to me!

dalg24:
CUDA failure is clearly unrelated (GPU not detected at runtime)

PR #7224, opened 2024-08-13
Hands off reserved identifiers

dalg24:
Avoid our using identifiers that contain double underscore or that begin with an underscore followed by an uppercase letter.
Any program using them is "ill-formed, no diagnostic required".


ldh4 (approved):

nliber (approved):
I knew starting with double underscores was reserved, but didn't know that any double underscores in an identifier was also reserved.

aprokop (approved):

dalg24:
I was not planning to add to the change log.  Let me know if you disagree.

PR #7225, opened 2024-08-14
Improve further view initialization/destruction

dalg24:
Following up on #7219

Refactoring `Impl::ViewValueFunctor` by getting rid of the specializations and consolidating everything into the primary template to avoid repeating ourselves and improve readability.

Separate the initialization vs destruction concerns based on the triviality of the default constructor and the triviality of the destructor which allows us to skip a kernel launch for non trivially-copyable types that are trivially destructible.

**edit** some toolchain did not like the properties of typed pointer passed into native memcpy functionality so I went back to guarding the initialization for trivial types.  I will attempt to relax in a follow up.

### DIFF
--- a/core/src/View/Kokkos_ViewAlloc.hpp
+++ b/core/src/View/Kokkos_ViewAlloc.hpp
@@ -69,14 +69,13 @@ struct ViewValueFunctor<DeviceType, ValueType, /*IsTrivial=*/false> {
   bool default_exec_space;
 
   template <class SameValueType = ValueType>
-  KOKKOS_INLINE_FUNCTION
-      std::enable_if_t<std::is_default_constructible<SameValueType>::value>
-      operator()(ConstructTag const&, const size_t i) const {
+  KOKKOS_FUNCTION
+      std::enable_if_t<std::is_default_constructible_v<SameValueType>>
+      operator()(ConstructTag, const size_t i) const {
     new (ptr + i) ValueType();
   }
 
-  KOKKOS_INLINE_FUNCTION void operator()(DestroyTag const&,
-                                         const size_t i) const {
+  KOKKOS_FUNCTION void operator()(DestroyTag, const size_t i) const {
     (ptr + i)->~ValueType();
   }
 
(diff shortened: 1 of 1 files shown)


dalg24:
Retest this please

core/src/View/Kokkos_ViewAlloc.hpp
```diff
@@ -139,132 +120,56 @@ struct ViewValueFunctor<DeviceType, ValueType, /*IsTrivial=*/false> {
...
+                            Kokkos::MemoryTraits<Kokkos::Unmanaged>>(ptr, n));
+
+    if (Kokkos::Profiling::profileLibraryLoaded()) {
+      Kokkos::Profiling::endParallelFor(kpID);
+    }
+    if (default_exec_space) {
+      space.fence("Kokkos::View::initialization via memset");
     }
-#endif
   }
```
masterleinad, line 168: What happened to this function?

dalg24: I got rid of it to see if we still need it (last commit)

dalg24: I withdrew it.  I had tried on Summit with CUDA 12.2 and I would have sworn it did not segfault.

dalg24:
Retest this please

dalg24:
Retest this please

nliber (approved):

aprokop (approved):
LGTM

core/src/View/Kokkos_ViewAlloc.hpp
```diff
@@ -139,19 +124,52 @@ struct ViewValueFunctor<DeviceType, ValueType, /*IsTrivial=*/false> {
...
+      Kokkos::Profiling::endParallelFor(kpID);
+    }
+    if (default_exec_space) {
+      space.fence("Kokkos::View::initialization via memset");
+    }
+  }
+
   void construct_shared_allocation() {
-    parallel_for_implementation<ConstructTag>();
+#ifndef KOKKOS_ARCH_A64FX
```
nmm0, line 153:
```suggestion
// On A64FX memset seems to do the wrong thing with regards to first touch
// leading to the significant performance issues
#ifndef KOKKOS_ARCH_A64FX
```

I think we should keep the comment here, just to document why this is necessary

nmm0 (approved):
Only a minor comment

masterleinad (approved):

dalg24:
Ignoring OpenMPTarget failures at runtime (GPU not detected)

PR #7227, opened 2024-08-15
Enforce modernize-type-traits

masterleinad:
This pull request enforces using C++14's and C++17's abbreviated type traits which we want to use.
Most of the changes in this pull request are mechanical and contained in one commit. 
However, MSVC and nvcc required flipping some SFINAE constraints around so that the constraint that is dependent on a template parameter comes first. 
Also, this pull request makes sure we only consider headers in our own implementations and not in the `tpls` (we have to list the directories explicitly since a negative regex will only be available from clang-tidy 19 on). For some reason, I still had to replace some type traits in gtest.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,3 +1,3 @@
-Checks: '-*,kokkos-*,modernize-use-using,modernize-use-nullptr,cppcoreguidelines-pro-type-cstyle-cast'
+Checks: '-*,kokkos-*,modernize-type-traits,modernize-use-using,modernize-use-nullptr,cppcoreguidelines-pro-type-cstyle-cast'
 FormatStyle: file
 HeaderFilterRegex: '(algorithms|benchmarks|containers|core|example|simd).*\.hpp'
--- a/containers/src/Kokkos_DynRankView.hpp
+++ b/containers/src/Kokkos_DynRankView.hpp
@@ -599,12 +599,12 @@ class DynRankView : public ViewTraits<DataType, Properties...> {
   // This assumes a contiguous underlying memory (i.e. no padding, no
   // striding... AND a Trilinos/Sacado scalar type )
   template <typename iType>
-  KOKKOS_INLINE_FUNCTION
-      std::enable_if_t<!std::is_same_v<typename drvtraits::value_type,
-                                       typename drvtraits::scalar_array_type> &&
-                           std::is_integral_v<iType>,
-                       reference_type>
-      operator[](const iType& i0) const {
+  KOKKOS_INLINE_FUNCTION std::enable_if_t<
+      std::is_integral_v<iType> &&
+          !std::is_same_v<typename drvtraits::value_type,
+                          typename drvtraits::scalar_array_type>,
+      reference_type>
+  operator[](const iType& i0) const {
     //      auto map = impl_map();
     const size_t dim_scalar = m_map.dimension_scalar();
     const size_t bytes      = this->span() / dim_scalar;
--- a/containers/src/Kokkos_DynRankView.hpp
+++ b/containers/src/Kokkos_DynRankView.hpp
@@ -94,8 +94,8 @@ struct DynRankDimTraits {
   // Non-strided Layout
   template <typename Layout>
   KOKKOS_INLINE_FUNCTION static std::enable_if_t<
-      (std::is_same<Layout, Kokkos::LayoutRight>::value ||
-       std::is_same<Layout, Kokkos::LayoutLeft>::value),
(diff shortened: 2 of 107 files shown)


nliber:
I'm generally in favor.

Is the intention of this just for standard traits, or Kokkos traits as well?

masterleinad:
> Is the intention of this just for standard traits, or Kokkos traits as well?

We should do the same for Kokkos traits but that's not that easily enforceable with clang-tidy.

.clang-tidy
```diff
@@ -1,3 +1,3 @@
-Checks: '-*,kokkos-*,modernize-use-using,modernize-use-nullptr,cppcoreguidelines-pro-type-cstyle-cast'
+Checks: '-*,kokkos-*,modernize-type-traits,modernize-use-using,modernize-use-nullptr,cppcoreguidelines-pro-type-cstyle-cast'
 FormatStyle: file
-HeaderFilterRegex: '.*/*.hpp'
```
masterleinad, line 3: This regex doesn't make much sense. We basically only check that the file ends in `hpp` and contains an arbitrary number of `/` and then an arbitrary character before that.

aprokop (commented):
I do like the intent. However, it seems like a lot of changes that will result in conflicts. I wonder if it's possible to pass only new patches through the check.

masterleinad:
> I do like the intent. However, it seems like a lot of changes that will result in conflicts. I wonder if it's possible to pass only new patches through the check.

I wouldn't know how (and if that's desirable). Ultimately, almost all of the changes are mechanical. Everything that required manual intervention is in separate commits.

masterleinad:
Retest this please.

.clang-tidy
dalg24: I wonder who added it 😉

masterleinad: I was young and naive. 🙂

tpls/gtest/gtest/gtest.h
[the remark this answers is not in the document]
masterleinad: I couldn't figure out why it still complained about this file.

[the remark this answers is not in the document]
dalg24: Oh that's a macro, that's why...

core/src/impl/Kokkos_Traits.hpp
```diff
@@ -102,8 +102,7 @@ struct are_integral<T, Args...> {
         // Accept std::is_integral OR std::is_enum as an integral value
         // since a simple enum value is automically convertible to an
         // integral value.
-    (std::is_integral<T>::value || std::is_enum<T>::value) &&
-    are_integral<Args...>::value
+    (std::is_integral_v<T> || std::is_enum_v<T>)&&are_integral<Args...>::value
```
dalg24, line 105: The format looks weird but if clang-format is happy I guess I don't really care

dalg24 (approved):
Approving so I don't need to look at this again

aprokop (approved):

containers/src/Kokkos_DynRankView.hpp
```diff
@@ -602,9 +600,9 @@ class DynRankView : public ViewTraits<DataType, Properties...> {
   // striding... AND a Trilinos/Sacado scalar type )
   template <typename iType>
   KOKKOS_INLINE_FUNCTION std::enable_if_t<
-      !std::is_same<typename drvtraits::value_type,
-                    typename drvtraits::scalar_array_type>::value &&
-          std::is_integral<iType>::value,
+      std::is_integral_v<iType> &&
+          !std::is_same_v<typename drvtraits::value_type,
+                          typename drvtraits::scalar_array_type>,
```
dalg24, line 605: What happened here?

dalg24:
Didn't you apply clang-tidy fixits?!
Does that mean you have been doing these or at least some by hand?

masterleinad:
I used clang-tidy fixes but had to manually change some occurrences for MSVC and nvcc in the last two commits:
>  However, MSVC and nvcc required flipping some SFINAE constraints around so that the constraint that is dependent on a template parameter comes first.

containers/src/Kokkos_DynRankView.hpp
dalg24: Oh that's just moved.  Nevermind

ndellingwood:
Does this need a 4.5 changelog entry, or should it be skipped as it is internal updates and cleanup?

masterleinad:
> Does this need a 4.5 changelog entry, or should it be skipped as it is internal updates and cleanup?

No need for a changelog entry.

PR #7228, opened 2024-08-15
Hide `IMPL_CUDA_MALLOC_ASYNC` option when CUDA is not enabled

dalg24:
Before this PR a build that does not have CUDA enabled would show that `IMPL_CUDA_MALLOC_ASYNC` was `ON` which is just silly.
Instead we adopt the same approach as we do for options like `CUDA_LAMBDA` and we only turn it `ON` when the CUDA backend is actually enabled.

**edit:** I added some drive-by changes for the HIP backend.  I was going to do another PR for it but they would conflict and it looked like a minor fixup that is not worth the trouble.  Let me know if you disagree and I will split them up again.


dalg24:
I am tempted to call it a bug

masterleinad (approved):

aprokop (approved):

Rombur (approved):

dalg24:
The OpenMPTarget build ran out of disk space and the CUDA+Clang build failures were at runtime (GPU not detected)

PR #7229, opened 2024-08-15
Introduce new `SequentialHostInit` view allocation property

dalg24:
Motivation: View of views support

```C++
{
  View<TypeContainsView, HostSpace> v(view_alloc("v", SequentialHostInit));
  v() = TypeContainsView(/* ... */);
  // no cleanup, the destructor will handle it
}
```

Note that, in the current form, non-default-constructible element types are not supported.

cc @stanmoore1 @rppawlo @vbrunini

### DIFF
--- a/core/src/View/Kokkos_ViewAlloc.hpp
+++ b/core/src/View/Kokkos_ViewAlloc.hpp
@@ -193,7 +193,6 @@ struct ViewValueFunctorSequentialHostInit {
   using ExecSpace = typename DeviceType::execution_space;
   using MemSpace  = typename DeviceType::memory_space;
   static_assert(SpaceAccessibility<HostSpace, MemSpace>::accessible);
-  static_assert(std::is_same_v<ExecSpace, typename MemSpace::execution_space>);
 
   ValueType* ptr;
   size_t n;
--- a/core/src/View/Kokkos_ViewAlloc.hpp
+++ b/core/src/View/Kokkos_ViewAlloc.hpp
@@ -196,10 +196,6 @@ struct ViewValueFunctorSequentialHostInit {
   size_t n;
 
   ViewValueFunctorSequentialHostInit() = default;
-  ViewValueFunctorSequentialHostInit(
-      const ViewValueFunctorSequentialHostInit&) = default;
-  ViewValueFunctorSequentialHostInit& operator=(
-      const ViewValueFunctorSequentialHostInit&) = default;
 
   ViewValueFunctorSequentialHostInit(ExecSpace const& arg_space,
                                      ValueType* const arg_ptr,
--- a/core/src/View/Kokkos_ViewAlloc.hpp
+++ b/core/src/View/Kokkos_ViewAlloc.hpp
@@ -189,29 +189,31 @@ struct ViewValueFunctor {
 };
 
 template <class DeviceType, class ValueType>
-struct ViewValueFunctorHostSerialInit {
+struct ViewValueFunctorSequentialHostInit {
   using ExecSpace = typename DeviceType::execution_space;
 
   ValueType* ptr;
   size_t n;
 
-  ViewValueFunctorHostSerialInit() = default;
-  ViewValueFunctorHostSerialInit(const ViewValueFunctorHostSerialInit&) =
-      default;
-  ViewValueFunctorHostSerialInit& operator=(
-      const ViewValueFunctorHostSerialInit&) = default;
+  ViewValueFunctorSequentialHostInit() = default;
+  ViewValueFunctorSequentialHostInit(
+      const ViewValueFunctorSequentialHostInit&) = default;
+  ViewValueFunctorSequentialHostInit& operator=(
+      const ViewValueFunctorSequentialHostInit&) = default;
 
-  ViewValueFunctorHostSerialInit(ExecSpace const& arg_space,
-                                 ValueType* const arg_ptr, size_t const arg_n,
-                                 std::string /*arg_name*/)
+  ViewValueFunctorSequentialHostInit(ExecSpace const& arg_space,
+                                     ValueType* const arg_ptr,
+                                     size_t const arg_n,
+                                     std::string /*arg_name*/)
       : ptr(arg_ptr), n(arg_n) {
     (void)arg_space;
     KOKKOS_ASSERT(arg_space == ExecSpace() &&
                   "FIXME if attached better be the default instance");
   }
 
-  ViewValueFunctorHostSerialInit(ValueType* const arg_ptr, size_t const arg_n,
-                                 std::string /*arg_name*/)
+  ViewValueFunctorSequentialHostInit(ValueType* const arg_ptr,
+                                     size_t const arg_n,
+                                     std::string /*arg_name*/)
       : ptr(arg_ptr), n(arg_n) {}
 
   void construct_shared_allocation() {
--- a/core/src/View/Kokkos_ViewAlloc.hpp
+++ b/core/src/View/Kokkos_ViewAlloc.hpp
@@ -191,6 +191,8 @@ struct ViewValueFunctor {
 template <class DeviceType, class ValueType>
 struct ViewValueFunctorSequentialHostInit {
   using ExecSpace = typename DeviceType::execution_space;
+  using MemSpace  = typename DeviceType::memory_space;
+  static_assert(SpaceAccessibility<HostSpace, MemSpace>::accessible);
 
   ValueType* ptr;
   size_t n;
@@ -203,8 +205,7 @@ struct ViewValueFunctorSequentialHostInit {
                                      std::string /*arg_name*/)
       : ptr(arg_ptr), n(arg_n) {
     (void)arg_space;
-    KOKKOS_ASSERT(arg_space == ExecSpace() &&
-                  "FIXME if attached better be the default instance");
+    KOKKOS_ASSERT(arg_space == ExecSpace());
   }
 
   ViewValueFunctorSequentialHostInit(ValueType* const arg_ptr,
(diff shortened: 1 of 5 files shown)


dalg24:
Older GCC versions gave
```
/var/jenkins/workspace/Kokkos_PR-7229/core/src/View/Kokkos_ViewAlloc.hpp:204:34: error: expected unqualified-id before '[' token
   ViewValueFunctorHostSerialInit([[maybe_unused]] ExecSpace const& arg_space,
                                  ^
/var/jenkins/workspace/Kokkos_PR-7229/core/src/View/Kokkos_ViewAlloc.hpp:204:34: error: expected ')' before '[' token
   ViewValueFunctorHostSerialInit([[maybe_unused]] ExecSpace const& arg_space,
                                 ~^
                                  )

**Comment by rppawlo** [comment]
Thanks @dalg24 !

File: core/src/impl/Kokkos_ViewCtor.hpp [hunk 1]
```diff
@@ -199,11 +203,22 @@ struct ViewCtorProp : public ViewCtorProp<void, P>... {
...
   using memory_space    = typename var_memory_space::type;
   using execution_space = typename var_execution_space::type;
   using pointer_type    = typename var_pointer::type;
 
+  // FIXME can't quite do that to validate user in[ut because we sometimes
+  // attach default instances ourselves
+  // static_assert(!host_serial_init || !has_execution_space,
+  //              "Incompatible HostSerialInit view "
+  //              "alloc property with execution space argument");
+
```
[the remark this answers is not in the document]
rppawlo: @dalg24 - we don't need this PR if you want to drop it. Only have one instance so far in empire that wasn't using the phalanx object that is already fixed.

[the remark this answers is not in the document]
dalg24: Rewriting Phalanx' VoV abstraction in terms of this might not be the worse idea.
In any case there seems to be a need for this feature.  Turns out the misuse of VoVs is more widespread than we suspected.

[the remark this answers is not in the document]
vbrunini: Was the "I am tempted to withdraw this change." comment just referring to the commented out static_assert here, not the whole feature then?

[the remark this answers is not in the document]
dalg24: The FIXME comment with the assertion

core/src/View/Kokkos_ViewAlloc.hpp
```diff
@@ -188,6 +188,54 @@ struct ViewValueFunctor {
   }
 };
 
+template <class DeviceType, class ValueType>
+struct ViewValueFunctorHostSerialInit {
+  using ExecSpace = typename DeviceType::execution_space;
```
masterleinad, line 193: I didn't find a check that the execution space can access host space anywhere else.
```suggestion
  using ExecSpace = typename DeviceType::execution_space;
  static_assert(Kokkos::SpaceAccessibility<ExecSpace, Kokkos::HostSpace>::accessible);
```

masterleinad (commented):
I'd like to hear from a couple of users that actually want to use this property.

vbrunini:
> I'd like to hear from a couple of users that actually want to use this property.

What would you like to hear from us?

core/src/View/Kokkos_ViewAlloc.hpp
crtrott: Yeah looks like we just pass the View device type in. But this initialization scheme should be valid for any View whose memory space is host accessible.

crtrott:
This looks pretty good. Two questions:
1. Daniel triggered this: should we pass the Views DeviceType (and thus execution space in there) or do something special so that the execution space we pass to the initialization functor actually is a host execution space?
2. Is `HostSerialInit` a good name? One could think it implies we use the Serial backend - which we don't. Maybe `SequentialHostInit` ?

crtrott:
Regarding the thing Daniel brought up: I am tending towards ignoring the execution space instance passed in, and maybe just statically check that the memory space is host accessible.

core/src/View/Kokkos_ViewAlloc.hpp
```diff
@@ -188,6 +188,54 @@ struct ViewValueFunctor {
...
+
+  ValueType* ptr;
+  size_t n;
+
+  ViewValueFunctorHostSerialInit() = default;
+  ViewValueFunctorHostSerialInit(const ViewValueFunctorHostSerialInit&) =
+      default;
+  ViewValueFunctorHostSerialInit& operator=(
+      const ViewValueFunctorHostSerialInit&) = default;
```
masterleinad, line 202: Any reason to declare (only) some of the special member functions explicitly?
```suggestion
```

masterleinad:
> * Daniel triggered this: should we pass the Views DeviceType (and thus execution space in there) or do something special so that the execution space we pass to the initialization functor actually is a host execution space?

I think it makes sense to just template on the memory space and check for accessbility since we are essentially ignoring the execution space (instance) anyway.

> * Is `HostSerialInit` a good name? One could think it implies we use the Serial backend - which we don't. Maybe `SequentialHostInit` ?

I think that's a more fitting name.

masterleinad:
> What would you like to hear from us?

I'm mostly curious if there are many users that are interested in this functionality (as opposed to the way we currently support View of Views) even after making sure they are compatible (in one way or the other) with the latest release.

vbrunini:
> > What would you like to hear from us?
> 
> I'm mostly curious if there are many users that are interested in this functionality (as opposed to the way we currently support View of Views) even after making sure they are compatible (in one way or the other) with the latest release.

Yes, this was motivated by seeing the amount of cleanup work several of our apps would have to do to fix up view of view usage for the 4.4 changes. Our current plan is to use the cmake option to disable the additional locks that were added to view construction/destruction in 4.4 and wait until this feature is available to leverage it to clean up our view of view usage.

core/src/View/Kokkos_ViewAlloc.hpp
dalg24: > this initialization scheme should be valid for any View whose memory space is host accessible

See L195

core/src/View/Kokkos_ViewAlloc.hpp
dalg24: Removed

dalg24:
> 1. Daniel triggered this: should we pass the Views DeviceType (and thus execution space in there) or do something special so that the execution space we pass to the initialization functor actually is a host execution space?

I added an assertion that the view memory space is accessible from host.
Not attempting to diagnose that the user passed an execution space to `view_alloc()` but will document that it is not legal.  It is unclear at that time how much refactor would be needed to achieve this.

> 2. Is `HostSerialInit` a good name? One could think it implies we use the Serial backend - which we don't. Maybe `SequentialHostInit` ?

I followed your suggestion.  I cannot come up with anything better at this time.

dalg24:
> Not attempting to diagnose that the user passed an execution space to `view_alloc()` but will document that it is not legal. It is unclear at that time how much refactor would be needed to achieve this.

On 2nd thought, one might want to asynchronously allocate in shared space from a specific exec space instance.  In which case we need to fence before we initialize.

masterleinad:
> On 2nd thought, one might want to asynchronously allocate in shared space from a specific exec space instance. In which case we need to fence before we initialize.

We always make sure that the allocation is available directly after allocating; the allocation mechanism is never observable to be asynchronous.

core/src/View/Kokkos_ViewAlloc.hpp
```diff
@@ -195,26 +195,28 @@ struct ViewValueFunctorSequentialHostInit {
...
   ViewValueFunctorSequentialHostInit(ValueType* const arg_ptr,
                                      size_t const arg_n,
                                      std::string /*arg_name*/)
-      : ptr(arg_ptr), n(arg_n) {}
+      : space(ExecSpace{}), ptr(arg_ptr), n(arg_n), default_exec_space(true) {}
 
   void construct_shared_allocation() {
+    if (!default_exec_space) {
+      space.fence("Kokkos::View::initialization after allocate");
+    }
```
masterleinad, line 219: Do we actually need this fence? It's impossible for a provided execution space instance to access the allocation at this point and we guarantee that the allocation mechanism itself always behaves synchronously.

dalg24: It looks like you are right and that CUDA/HIP shared and host pinned allocations are synchronous.
It was unclear to me whether we fence for host accessible SYCL allocations (not saying that is a reason not to drop that fence, just that we might want to check on that).

masterleinad: All examples in https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html seem to suggets that `sycl::malloc*` is synchronous: queues are out-of-order by default and there is no fence between kernels and memory allocations (see https://registry.khronos.org/SYCL/specs/sycl-2020/html/sycl-2020.html#sec:usm).

core/src/View/Kokkos_ViewAlloc.hpp
```diff
@@ -188,6 +188,56 @@ struct ViewValueFunctor {
   }
 };
 
+template <class DeviceType, class ValueType>
+struct ViewValueFunctorSequentialHostInit {
+  using ExecSpace = typename DeviceType::execution_space;
+  using MemSpace  = typename DeviceType::memory_space;
+  static_assert(SpaceAccessibility<HostSpace, MemSpace>::accessible);
+  static_assert(std::is_same_v<ExecSpace, typename MemSpace::execution_space>);
```
crtrott, line 196:
```suggestion
```

This shouldn't be here.

core/src/View/Kokkos_ViewAlloc.hpp
crtrott: Yeah, I think we don't need that.

masterleinad (approved):
Fine with me.

crtrott (approved):
Yeah this looks good. The tests cover the different use cases we talked about.

dalg24:
Ignoring unrelated failures (nvc++ segfault, shared space test failure in the Clang+CUDA build, and GPU not detected at runtime in RDC build)

PR #7236, opened 2024-08-20
Fix clang-tidy header guard to ignore tpls

masterleinad:
Part of #7227. Currently, `clang-tidy` takes all header files (that end in `hpp`) in the source directory into account. This pull request makes sure that we only check header files in the
- algorithms,
- benchmarks,
- containers,
- core,
- example, and
- simd

directories to ignore `tpls`. A negative regex is only available from `clang-tidy` 19 on. For this to work properly, we also can't include tpl directories via a relative path from one of the above directories but must use absolute paths via `KOKKOS_SOURCE_DIR`.


dalg24 (approved):

ldh4 (approved):

dalg24:
Ignoring unrelated failure in one of the CUDA build (GPU not detected at runtime)

PR #7237, opened 2024-08-20
Allow disabling tests via CMake option

masterleinad:
Part of #6164. Addressing https://github.com/kokkos/kokkos/pull/6164#issuecomment-1739953719.
TriBITS allos disabling tests via
```
-D <fullTestName>_DISABLE:BOOL=ON
```
and this pull request mirrors that behavior so that `Trilinos` can continue disable `Kokkos` tests even after #6164.

edit:
Also added options for running in serial and setting environment variables.


dalg24:
Found it https://tribitspub.github.io/TriBITS/users_guide/index.html#disabling-tests-externally-tribits-add-test

cmake/kokkos_tribits.cmake
```diff
@@ -222,12 +222,15 @@ MACRO(KOKKOS_ADD_TEST_EXECUTABLE ROOT_NAME)
...
     "SOURCES"
     ${ARGN})
-  KOKKOS_ADD_EXECUTABLE(${ROOT_NAME}
-    SOURCES ${PARSE_SOURCES}
-    ${PARSE_UNPARSED_ARGUMENTS}
-    TESTONLYLIBS ${KOKKOS_GTEST_LIB}
-  )
-  SET(EXE_NAME ${PACKAGE_NAME}_${ROOT_NAME})
+  # Don't do anything if the user disabled the test
+  IF(NOT ${PACKAGE_NAME}_${ROOT_NAME}_DISABLE)
```
dalg24, line 226: That does not look right.  As I understand it, disabling a test means not running it but still building the executable.

dalg24 (commented):
Why did you decide to guard adding the test versus setting the [DISABLED](https://cmake.org/cmake/help/latest/prop_test/DISABLED.html) property on it?

cmake/kokkos_tribits.cmake
masterleinad: Yes, I think you are right that TriBITS still compiles the test. The relevant code is https://github.com/TriBITSPub/TriBITS/blob/474375581690f008aa78a55f0fb213a3a53df98a/tribits/core/test_support/TribitsAddTestHelpers.cmake#L603-L627.

masterleinad:
> Why did you decide to guard adding the test versus setting the [DISABLED](https://cmake.org/cmake/help/latest/prop_test/DISABLED.html) property on it?

`TriBITS` doesn't even give the test to `CMake` when this option is set.

masterleinad:
We should consult `Trilinos` maintainers and @ndellingwood if this is good enough to deal with `Kokkos` tests in `Trilinos`.

dalg24:
I would be curious to hear why this support for disabling tests is needed that cannot just be done with an exclude regex argument to ctest.
Any idea?

masterleinad:
> I would be curious to hear why this support for disabling tests is needed that cannot just be done with an exclude regex argument to ctest.

https://tribits.org/doc/TribitsBuildReference.html#disabling-specific-tests even talks about using `ctest -E` as alternative to `-D <fullTestName>_SET_DISABLED_AND_MSG="<messageWhyDisabled>` (which allows for more information). It's not clear to me why both `DISABLE` and `SET_DISABLED_AND_MSG` exist. Note that there is also `EXE_DISABLE` which avoids building the executable.

ndellingwood:
@masterleinad @dalg24 the `-D <TestName>_DISABLE=ON` option is a useful option used by some Trilinos configurations, the removal of the option would show up and potentially impact configurations with the Trilinos AutoTester. For example, disabling timing-based tests is done to avoid spurious failures that force a lengthy retest, e.g.
https://github.com/trilinos/Trilinos/blob/bf922e75428ef7dd925811f0a532d05dd9cff808/packages/framework/ini-files/config-specs.ini#L1466

There are a couple other options that similarly help control how tests are run to avoid spurious failures requiring retest:

* The option `-D <TestName>_EXTRA_ARGS="<some string>"` allows finer-grained disable of sub-tests in a test, for example using gtest_filter, which is also used with the Trilinos AutoTester for similar reasons to above
https://github.com/trilinos/Trilinos/blob/bf922e75428ef7dd925811f0a532d05dd9cff808/packages/framework/ini-files/config-specs.ini#L360-L362

* The option `-D <TestName>_SET_RUN_SERIAL=ON` allows tests to be launched sequentially, used to mitigate cases of resource contention when tests are run in parallel (i.e. `ctest -jN`)
https://github.com/trilinos/Trilinos/blob/bf922e75428ef7dd925811f0a532d05dd9cff808/packages/framework/ini-files/config-specs.ini#L1060-L1062

dalg24:
OK we can do DISABLE and RUN_SERIAL as properties on the tests.  Any good reason why you would need these extra args rather than setting the ENVIRONMENT property.  You would also be able to set Kokkos options or GTest filters that way.

cmake/fake_tribits.cmake
```diff
@@ -94,6 +94,18 @@ FUNCTION(KOKKOS_ADD_TEST)
...
     SET(TEST_NAME ${PACKAGE_NAME}_${TEST_NAME})
+
+    # For compatibility with Trilinos testing
+    IF(${TEST_NAME}_DISABLE)
+      RETURN()
+    ENDIF()
+    IF(${TEST_NAME}_SET_RUN_SERIAL)
+      SET_TESTS_PROPERTIES(${TEST_NAME} PROPERTIES RUN_SERIAL ON)
+    ENDIF()
+    IF(${TEST_NAME}_ENVIRONMENT)
```
dalg24, line 105: Maybe add a comment that TriBITS supports `_EXTRA_ARGS` instead of `_ENVIRONMENT`

dalg24 (approved):
On 2nd thought I am tempted to just go ahead and add support for EXTRA_ARGS (I'd keep ENVIRONMENT)
What do you think?

masterleinad:
> On 2nd thought I am tempted to just go ahead and add support for EXTRA_ARGS (I'd keep ENVIRONMENT)

Sure, added in the last commit.

cmake/fake_tribits.cmake
```diff
@@ -94,12 +94,26 @@ FUNCTION(KOKKOS_ADD_TEST)
     # Prepend package name to the test name
     # These should be the full target name
     SET(TEST_NAME ${PACKAGE_NAME}_${TEST_NAME})
+
+    # For compatibility with Trilinos testing
```
dalg24, line 98:
```suggestion
    # For compatibility with Trilinos testing, we support:
    #  * `-D <fullTestName>_EXTRA_ARGS="<arg0>;<arg1>;<arg2>;..."`
    #  * `-D <fullTestName>_SET_RUN_SERIAL=ON`
```

cmake/fake_tribits.cmake
```diff
@@ -94,12 +94,26 @@ FUNCTION(KOKKOS_ADD_TEST)
...
+    IF(${TEST_NAME}_DISABLE)
+      RETURN()
+    ENDIF()
+    IF(${TEST_NAME}_SET_RUN_SERIAL)
+      SET_TESTS_PROPERTIES(${TEST_NAME} PROPERTIES RUN_SERIAL ON)
+    ENDIF()
+    # Trilinos doesn't have this option but only ${${TEST_NAME}_EXTRA_ARGS}, see below.
```
dalg24, line 105:
```suggestion
    # TriBITS doesn't actually currently support `-D <fullTestName>_ENVIRONMENT`
    # but we decided to add it anyway
```

crtrott:
I don't think we care about testing this or? I assume Trilinos will complain if it breaks.

crtrott (approved):

ndellingwood:
I'll run some checks with Trilinos to confirm the behavior of the added options and update later today

ndellingwood:
I tested this PR directly in a simple build of Trilinos (sems-gcc/8.5.0, Serial backend, only Kokkos enabled)

* The `_DISABLE`, `_SET_RUN_SERIAL` and `_EXTRA_ARGS` options worked as expected
* The `_ENVIRONMENT` option was not recognized; I attempted to use it to filter out a rng subtest but it was not recognized at configuration and the test:
```
CMake Warning:

    Kokkos_UnitTest_Random_ENVIRONMENT
```

Here is the configuration I tested:

```bash
export TRILINOS_DIR=<path-to-source>

module purge
module load sems-cmake sems-gcc/8.5.0 sems-openmpi/4.1.6 sems-ninja sems-python/3.9.0

# Configure Trilinos
cmake \
 -GNinja \
 -DCMAKE_INSTALL_PREFIX="${PWD}/install" \
 -DCMAKE_CXX_COMPILER=$(which g++) \
 -DCMAKE_CXX_STANDARD="17" \
 -DTrilinos_ENABLE_TESTS=ON \
 -DTrilinos_ENABLE_ALL_PACKAGES=OFF \
 -DTrilinos_ENABLE_Kokkos=ON \
 -DKokkos_ENABLE_SERIAL=ON \
 -DKokkos_SOURCE_DIR_OVERRIDE:STRING=kokkos \
 -DKokkosKernels_SOURCE_DIR_OVERRIDE:STRING=kokkos-kernels \
\
 -D Kokkos_CoreUnitTest_Serial1_DISABLE=ON \
 -D Kokkos_CoreUnitTest_Default_SET_RUN_SERIAL=ON \
 -D Kokkos_UnitTest_Random_EXTRA_ARGS="--gtest_filter=-serial.Random_XorShift64" \
 -D Kokkos_UnitTest_Random_ENVIRONMENT="--gtest_filter=-serial.Random_XorShift1024_0" \
$TRILINOS_DIR
```

@masterleinad should I also test with these changed pulled into #6164 to confirm that the options are recognized after the TriBITS removal?

masterleinad:
@ndellingwood I fixed the order so that the options are applied past test generation. The syntax should be
```
cmake -D Kokkos_UnitTest_Random_ENVIRONMENT="GTEST_FILTER=-serial.Random_XorShift1024_0" ..
```

ndellingwood:
@masterleinad thanks, I'll retest now

masterleinad:
> @masterleinad should I also test with these changed pulled into #6164 to confirm that the options are recognized after the TriBITS removal?

I expect that you would run into merge conflicts. Let's test after e have merged this and rebased #6164.

ndellingwood:
@masterleinad I tested a clean build with sha 3fddbfb6b0012391eb6aa7086a858dc131b531af and using ` -D Kokkos_UnitTest_Random_ENVIRONMENT="GTEST_FILTER=-serial.Random_XorShift1024_0"` , but I still get the warning output implying the option is unrecognized
```
CMake Warning:

    Kokkos_UnitTest_Random_ENVIRONMENT
```

dalg24:
Are you sure about the name? Shouldn't it be Kokkos_AlgorithmsUnitTest_Bla

masterleinad:
All the changes here are in the branch for Kokkos-standalone so it shouldn't actually have any impact for a Trilinos build as of now.

ndellingwood:
@dalg24 when I run `ctest -N` I see `Kokkos_UnitTest_Random` as the name

dalg24:
> All the changes here are in the branch for Kokkos-standalone so it shouldn't actually have any impact for a Trilinos build as of now. 🤦

So close this and merge into the overhaul PR?

masterleinad:
> So close this and merge into the overhaul PR?

Yeah I checked that the ENVIRONMENT option works as expected for `Kokkos` and `Trilinos`.

PR #7239, opened 2024-08-20
OpenMPTarget: Update dockerfile for clang.

rgayatri23:
Draft:
The PR attempts to fix the random errors of "No GPU found" by updating the docker with options that build the NVIDIA target arch.

### DIFF
--- a/scripts/docker/Dockerfile.openmptarget
+++ b/scripts/docker/Dockerfile.openmptarget
@@ -55,6 +55,7 @@ RUN LLVM_URL=https://github.com/llvm/llvm-project/archive &&\
       -DCMAKE_CXX_COMPILER=g++ \
       -DLLVM_ENABLE_PROJECTS="clang" \
       -DLLVM_ENABLE_RUNTIMES="openmp" \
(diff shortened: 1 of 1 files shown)


diehlpk (approved):
LGTM!

aprokop (commented):
The variables seem [correct](https://llvm.org/docs/CMake.html#llvm-related-variables). It seems that OpenMPTarget builds, though the tests fail due `51: Error: no GPU available for execution.`

crtrott:
And you can't reproduce this locally?

crtrott (changes requested):
Doesn't fix the issue.

rgayatri23:
Yeah this does not fix the issue for which it was originally created. However I think this is necessary to get all the OpenMP libs in the docker build. I  tested this locally and the current docker build from `develop` does not seem to have OpenMP libs in the final install. Unsure how the tests are running actually in such a case. 

When I tried to replicate this locally, I needed this option to build the correct LLVM. However I never faced the "No GPU" error. Unsure at this point on whether the problem is with the build or CI resources.

crtrott (approved):

PR #7242, opened 2024-08-21
Allow extracting host and device views from DualView with const value type

masterleinad:
We discussed in https://kokkosteam.slack.com/archives/C5BGU5NDQ/p1724189696168339 that you can't get the underlying host and device `View`s of a `DualView` with `const` value type due to a mismatch between the respective return type and the the actual `View` returned. This pull request fixes that by just deducing the return type via `auto`.

### DIFF
--- a/containers/src/Kokkos_DualView.hpp
+++ b/containers/src/Kokkos_DualView.hpp
@@ -387,11 +387,6 @@ class DualView : public ViewTraits<DataType, Properties...> {
 
   /// \brief Return a View on a specific device \c Device.
   ///
-  /// Please don't be afraid of the nested if_c expressions in the return
-  /// value's type.  That just tells the method what the return type
-  /// should be: t_dev if the \c Device template parameter matches
-  /// this DualView's device type, else t_host.
-  ///
   /// For example, suppose you create a DualView on Cuda, like this:
   /// \code
   ///   using dual_view_type =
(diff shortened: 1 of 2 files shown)


containers/src/Kokkos_DualView.hpp
```diff
@@ -410,18 +410,7 @@ class DualView : public ViewTraits<DataType, Properties...> {
...
-          impl_device_matches_thost_device<Device>::value, t_host,
-          typename std::conditional_t<
-              impl_device_matches_thost_exec<Device>::value, t_host,
-              typename std::conditional_t<
-                  impl_device_matches_tdev_exec<Device>::value, t_dev,
-                  typename std::conditional_t<
-                      impl_device_matches_tdev_memory_space<Device>::value,
-                      t_dev, t_host>>>>>
-  view() const {
+  KOKKOS_INLINE_FUNCTION auto view() const {
```
romintomasetti, line 413: While you're at it, `inline` is implicit already:
```suggestion
  KOKKOS_FUNCTION auto view() const {
```

containers/unit_tests/TestDualView.hpp
```diff
@@ -643,6 +611,17 @@ TEST(TEST_CATEGORY,
                hvt::device_type::execution_space::name());
 }
 
+TEST(TEST_CATEGORY,
+     dualview_template_views_return_correct_views_from_const_dual_view) {
+  DualViewType dv("myView", 100);
+  ConstDualViewType const_dv = dv;
+  dv.clear_sync_state();
+  ASSERT_EQ(dv.view<typename Kokkos::DefaultHostExecutionSpace>(),
```
dalg24, line 619: Why typename?

containers/unit_tests/TestDualView.hpp
```diff
@@ -520,58 +520,26 @@ namespace {
...
+#ifdef KOKKOS_HAS_SHARED_SPACE
+template <typename ExecutionSpace>
+using SharedTestSpace = Kokkos::SharedSpace;
 #else
-template <typename ExecSpace>
-struct UVMSpaceFor {
-  using type = typename ExecSpace::memory_space;
-};
+template <typename ExecutionSpace>
+using SharedTestSpace = typename ExecutionSpace::memory_space;
```
dalg24, line 529: I don't really like that name.  What about `TestSharedSpace`?
Or do you have any other suggestion where we don't interleave a word in the alias name?

dalg24 (approved):

tcclevenger (approved):

dalg24:
Unrelated failures

PR #7245, opened 2024-08-22
Fix Deep_copy(serial_exec, dst, src) with multiple host backends

### LINKED ISSUE #7241: `deep_copy(serial_exec, dst, src)` runs on the host parallel backend default instance
When both the Serial backend and some other host parallel backend are enabled simultaneously, the deep copy is not executed on the provided serial execution space instance but runs on the default host parallel backend exec instead.
https://github.com/kokkos/kokkos/blob/16eca672987e5e20d4ccd2407d262f655eae1fcf/core/src/Kokkos_HostSpace.hpp#L204-L209
https://github.com/kokkos/kokkos/blob/f5c527d7cb12e572643e0a0810ba32fcb6d8ffcc/core/src/impl/Kokkos_HostSpace_deepcopy.cpp#L37-L44

yasahi-hpc:
Fixes #7241 

This PR aims at using the correct execution space for the `deep_copy` when both the Serial backend and some other host parallel backends are enabled.

### DIFF
--- a/core/src/Kokkos_HostSpace.hpp
+++ b/core/src/Kokkos_HostSpace.hpp
@@ -192,8 +192,8 @@ struct DeepCopy<HostSpace, HostSpace, ExecutionSpace> {
   DeepCopy(const ExecutionSpace& exec, void* dst, const void* src, size_t n) {
     // If ExecutionSpace is a Device ExecutionSpace, we should call
     // this from DefaultHostExecutionSpace
-    if constexpr (std::is_same_v<ExecutionSpace, DefaultExecutionSpace> &&
-                  !std::is_same_v<ExecutionSpace, DefaultHostExecutionSpace>) {
+    if constexpr (!Kokkos::SpaceAccessibility<ExecutionSpace,
+                                              Kokkos::HostSpace>::accessible) {
       exec.fence(
           "Kokkos::Impl::DeepCopy<HostSpace, HostSpace, "
           "ExecutionSpace>::DeepCopy: fence before copy");
--- a/core/src/Kokkos_HostSpace.hpp
+++ b/core/src/Kokkos_HostSpace.hpp
@@ -205,7 +205,13 @@ struct DeepCopy<HostSpace, HostSpace, ExecutionSpace> {
     exec.fence(
         "Kokkos::Impl::DeepCopy<HostSpace, HostSpace, "
         "ExecutionSpace>::DeepCopy: fence before copy");
-    hostspace_parallel_deepcopy_async(exec, dst, src, n);
+    // If ExecutionSpace is a Device ExecutionSpace, we should call
+    // this from DefaultHostExecutionSpace
+    if constexpr (std::is_same_v<ExecutionSpace, DefaultExecutionSpace>) {
+      hostspace_parallel_deepcopy_async(dst, src, n);
+    } else {
+      hostspace_parallel_deepcopy_async(exec, dst, src, n);
+    }
   }
 };
 
--- a/core/src/Kokkos_HostSpace.hpp
+++ b/core/src/Kokkos_HostSpace.hpp
@@ -204,7 +204,8 @@ struct DeepCopy<HostSpace, HostSpace, ExecutionSpace> {
   DeepCopy(const ExecutionSpace& exec, void* dst, const void* src, size_t n) {
     // If ExecutionSpace is a Device ExecutionSpace, we should call
     // this from DefaultHostExecutionSpace
-    if constexpr (std::is_same_v<ExecutionSpace, DefaultExecutionSpace>) {
+    if constexpr (std::is_same_v<ExecutionSpace, DefaultExecutionSpace> &&
+                  !std::is_same_v<ExecutionSpace, DefaultHostExecutionSpace>) {
       exec.fence(
           "Kokkos::Impl::DeepCopy<HostSpace, HostSpace, "
           "ExecutionSpace>::DeepCopy: fence before copy");
--- a/core/src/Kokkos_HostSpace.hpp
+++ b/core/src/Kokkos_HostSpace.hpp
@@ -190,8 +190,6 @@ struct DeepCopy<HostSpace, HostSpace, ExecutionSpace> {
   }
 
   DeepCopy(const ExecutionSpace& exec, void* dst, const void* src, size_t n) {
-    // If ExecutionSpace is a Device ExecutionSpace, we should call
-    // this from DefaultHostExecutionSpace
     if constexpr (!Kokkos::SpaceAccessibility<ExecutionSpace,
                                               Kokkos::HostSpace>::accessible) {
       exec.fence(
--- a/core/src/Kokkos_HostSpace.hpp
+++ b/core/src/Kokkos_HostSpace.hpp
@@ -183,18 +183,6 @@ namespace Kokkos {
 
 namespace Impl {
 
-template <>
-struct DeepCopy<HostSpace, HostSpace, DefaultHostExecutionSpace> {
-  DeepCopy(void* dst, const void* src, size_t n) {
-    hostspace_parallel_deepcopy(dst, src, n);
-  }
-
-  DeepCopy(const DefaultHostExecutionSpace& exec, void* dst, const void* src,
-           size_t n) {
-    hostspace_parallel_deepcopy_async(exec, dst, src, n);
-  }
-};
-
 template <class ExecutionSpace>
 struct DeepCopy<HostSpace, HostSpace, ExecutionSpace> {
   DeepCopy(void* dst, const void* src, size_t n) {
--- a/core/src/Kokkos_HostSpace.hpp
+++ b/core/src/Kokkos_HostSpace.hpp
@@ -205,7 +205,7 @@ struct DeepCopy<HostSpace, HostSpace, ExecutionSpace> {
     exec.fence(
         "Kokkos::Impl::DeepCopy<HostSpace, HostSpace, "
         "ExecutionSpace>::DeepCopy: fence before copy");
-    hostspace_parallel_deepcopy_async(dst, src, n);
+    hostspace_parallel_deepcopy_async(exec, dst, src, n);
   }
 };
 
--- a/core/src/Kokkos_HostSpace.hpp
+++ b/core/src/Kokkos_HostSpace.hpp
@@ -202,12 +202,12 @@ struct DeepCopy<HostSpace, HostSpace, ExecutionSpace> {
   }
 
   DeepCopy(const ExecutionSpace& exec, void* dst, const void* src, size_t n) {
-    exec.fence(
-        "Kokkos::Impl::DeepCopy<HostSpace, HostSpace, "
-        "ExecutionSpace>::DeepCopy: fence before copy");
     // If ExecutionSpace is a Device ExecutionSpace, we should call
     // this from DefaultHostExecutionSpace
     if constexpr (std::is_same_v<ExecutionSpace, DefaultExecutionSpace>) {
+      exec.fence(
+          "Kokkos::Impl::DeepCopy<HostSpace, HostSpace, "
+          "ExecutionSpace>::DeepCopy: fence before copy");
       hostspace_parallel_deepcopy_async(dst, src, n);
     } else {
       hostspace_parallel_deepcopy_async(exec, dst, src, n);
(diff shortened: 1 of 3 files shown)


core/src/impl/Kokkos_HostSpace_deepcopy.cpp
```diff
@@ -138,6 +137,18 @@ void hostspace_parallel_deepcopy_async(const DefaultHostExecutionSpace& exec,
...
 
+// Explicit instanciation
+template void hostspace_parallel_deepcopy_async<DefaultHostExecutionSpace>(
+    const DefaultHostExecutionSpace&, void*, const void*, ptrdiff_t);
+
+#if defined(KOKKOS_ENABLE_SERIAL) &&                                    \
+    (defined(KOKKOS_ENABLE_OPENMP) || defined(KOKKOS_ENABLE_THREADS) || \
+     defined(KOKKOS_ENABLE_HPX))
+// Instanciate only if both the Serial backend and some other host parallel
+// backend are enabled
```
masterleinad, line 148:
```suggestion
// Instantiate only if both the Serial backend and some other host parallel
// backend are enabled
```

yasahi-hpc: Fixed accordingly

core/src/Kokkos_HostSpace.hpp
```diff
@@ -205,7 +205,13 @@ struct DeepCopy<HostSpace, HostSpace, ExecutionSpace> {
...
         "Kokkos::Impl::DeepCopy<HostSpace, HostSpace, "
         "ExecutionSpace>::DeepCopy: fence before copy");
-    hostspace_parallel_deepcopy_async(dst, src, n);
+    // If ExecutionSpace is a Device ExecutionSpace, we should call
+    // this from DefaultHostExecutionSpace
+    if constexpr (std::is_same_v<ExecutionSpace, DefaultExecutionSpace>) {
+      hostspace_parallel_deepcopy_async(dst, src, n);
+    } else {
+      hostspace_parallel_deepcopy_async(exec, dst, src, n);
+    }
```
masterleinad, line 214: We don't need to fence if we are calling the `hostspace_parallel_deepcopy_async` overload taking an execution space instance.

yasahi-hpc: Something like this? Or can I completely remove this fence since this function takes an execution space argument.

```C++
    if constexpr (std::is_same_v<ExecutionSpace, DefaultExecutionSpace>) {
      exec.fence(
          "Kokkos::Impl::DeepCopy<HostSpace, HostSpace, "
          "ExecutionSpace>::DeepCopy: fence before copy");
      hostspace_parallel_deepcopy_async(dst, src, n);
    } else {
      hostspace_parallel_deepcopy_async(exec, dst, src, n);
    }
```

masterleinad: Yes, this looks good. We still need the fence since we can't enqueue the work on the given execution space instance in the first case and the deepcopy might access memory used by that execution space instance.

dalg24: More like
```C++
if (std::is_same_v<ExecutionSpace, DefaultExecutionSpace> 
  && !std::is_same_v<DefaultExecutionSpace, DefaultHostExecutionSpace>)

Comment (masterleinad) [reply core/src/Kokkos_HostSpace.hpp]: If they are the same we would use the specialization in line 187 (which we could get rid of).

File: core/src/Kokkos_HostSpace.hpp [hunk 3]
```diff
@@ -202,10 +202,17 @@ struct DeepCopy<HostSpace, HostSpace, ExecutionSpace> {
...
-        "ExecutionSpace>::DeepCopy: fence before copy");
-    hostspace_parallel_deepcopy_async(dst, src, n);
+    // If ExecutionSpace is a Device ExecutionSpace, we should call
+    // this from DefaultHostExecutionSpace
+    if constexpr (std::is_same_v<ExecutionSpace, DefaultExecutionSpace> &&
+                  !std::is_same_v<ExecutionSpace, DefaultHostExecutionSpace>) {
+      exec.fence(
+          "Kokkos::Impl::DeepCopy<HostSpace, HostSpace, "
+          "ExecutionSpace>::DeepCopy: fence before copy");
+      hostspace_parallel_deepcopy_async(dst, src, n);
```
crtrott, line 212: Uhoh I think semantically we need to do actually something more complicated than this. To be fully correct in this case I believe we need prevent any operations on exec to overlap with the deep_copy. Which means we likely need to 
- acquire exec lock
- wait for all exec operations to finish
- perform the deep_copy (and wait for it to finish)
- release the exec lock

yasahi-hpc: That is indeed more complicated.
Could you please detail why `exec.fence()` is not good enough? 
Also, is there a way to acquire exec lock for all execution spaces?
What I have in mind is something like this, which should not work
```C++
auto *internal_instance = exec.impl_internal_space_instance();
// I guess this is not available for device execution spaces
std::lock_guard<std::mutex> lock(internal_instance->m_instance_mutex);
internal_instance->fence(
    "Kokkos::Impl::DeepCopy<HostSpace, HostSpace, "
    "ExecutionSpace>::DeepCopy: fence before copy");
// How to wait for this to finish?
hostspace_parallel_deepcopy_async(dst, src, n);
```

crtrott: Essentially the problem is that the next kernel submitted to exec could overlap with the async host deep_copy, but semantically the deep_copy needs to be sequenced in the device queue. If you only have a single process thread, then the overlap could happen if host exec is async (i.e. HPX) which we can fix by fencing after the host deep_copy the host exec. If there are multiple (non-kokkos) threads, the overlap could come from another thread submitting to exec while the deep copy is running. But I am ok fixing that separately.

core/src/impl/Kokkos_HostSpace_deepcopy.cpp
```diff
@@ -43,10 +43,10 @@ void hostspace_parallel_deepcopy_async(void* dst, const void* src,
       "Kokkos::Impl::hostspace_parallel_deepcopy_async: fence after copy");
 }
 
-void hostspace_parallel_deepcopy_async(const DefaultHostExecutionSpace& exec,
-                                       void* dst, const void* src,
-                                       ptrdiff_t n) {
-  using policy_t = Kokkos::RangePolicy<Kokkos::DefaultHostExecutionSpace>;
+template <typename ExecutionSpace>
```
ajpowelsnl, line 46: Would it make sense to use `template <class ExecutionSpace>` instead of `typename`?

core/src/impl/Kokkos_HostSpace_deepcopy.cpp
```diff
@@ -138,6 +137,18 @@ void hostspace_parallel_deepcopy_async(const DefaultHostExecutionSpace& exec,
   }
 }
 
+// Explicit instanciation
```
ajpowelsnl, line 140: Suggestion:   "// Explicit instantiation"

core/src/impl/Kokkos_HostSpace_deepcopy.hpp
```diff
@@ -28,9 +28,9 @@ void hostspace_fence(const DefaultHostExecutionSpace& exec);
 void hostspace_parallel_deepcopy(void* dst, const void* src, ptrdiff_t n);
 // DeepCopy called with an execution space that can't access HostSpace
 void hostspace_parallel_deepcopy_async(void* dst, const void* src, ptrdiff_t n);
-void hostspace_parallel_deepcopy_async(const DefaultHostExecutionSpace& exec,
-                                       void* dst, const void* src, ptrdiff_t n);
-
+template <typename ExecutionSpace>
```
ajpowelsnl, line 31: Same question here:  is there a reason to use `typename` vs. `class` in templated expression?

ajpowelsnl (commented):
Looks like a good start towards resolving concerns about `deep_copy` (detailed in #7241, #7261) on host serial execution spaces

core/src/impl/Kokkos_HostSpace_deepcopy.cpp
dalg24: He is free to use whichever he likes better, as long as he does not mix in the same template declaration.
We prefer consistency at local scope but we have no strict guideline at the moment.

core/src/impl/Kokkos_HostSpace_deepcopy.cpp
masterleinad:
```suggestion
// Explicit instantiation
```

core/src/Kokkos_HostSpace.hpp
```diff
@@ -183,29 +183,24 @@ namespace Kokkos {
...
 
   DeepCopy(const ExecutionSpace& exec, void* dst, const void* src, size_t n) {
-    exec.fence(
-        "Kokkos::Impl::DeepCopy<HostSpace, HostSpace, "
-        "ExecutionSpace>::DeepCopy: fence before copy");
-    hostspace_parallel_deepcopy_async(dst, src, n);
+    // If ExecutionSpace is a Device ExecutionSpace, we should call
+    // this from DefaultHostExecutionSpace
+    if constexpr (std::is_same_v<ExecutionSpace, DefaultExecutionSpace> &&
+                  !std::is_same_v<ExecutionSpace, DefaultHostExecutionSpace>) {
```
masterleinad, line 196: I find it hard to argue about this. Do you mind using `SpaceAccessibility` instead?
```suggestion
    if constexpr (!Kokkos::SpaceAccesibility<ExecutionSpace, Kokkos::HostSpace>::accessible) {
```

core/src/Kokkos_HostSpace.hpp
```diff
@@ -183,29 +183,24 @@ namespace Kokkos {
...
     hostspace_parallel_deepcopy(dst, src, n);
   }
 
   DeepCopy(const ExecutionSpace& exec, void* dst, const void* src, size_t n) {
-    exec.fence(
-        "Kokkos::Impl::DeepCopy<HostSpace, HostSpace, "
-        "ExecutionSpace>::DeepCopy: fence before copy");
-    hostspace_parallel_deepcopy_async(dst, src, n);
+    // If ExecutionSpace is a Device ExecutionSpace, we should call
+    // this from DefaultHostExecutionSpace
```
masterleinad, line 194:
```suggestion
```

yasahi-hpc:
> Looks like a good start towards resolving concerns about `deep_copy` (detailed in #7241, #7261) on host serial execution spaces

I will start working on #7261 after this PR

core/src/impl/Kokkos_HostSpace_deepcopy.cpp
yasahi-hpc: Fixed

core/src/Kokkos_HostSpace.hpp
yasahi-hpc: This is indeed a better way. I have fixed accordingly.

core/src/Kokkos_HostSpace.hpp
yasahi-hpc: Sorry, what is the difference?

masterleinad (approved):

core/src/Kokkos_HostSpace.hpp
masterleinad: I didn't think that the comment is very useful anymore with the current version of the code so I suggested to delete it.

dalg24 (approved):

dalg24:
Unrelated failures (GPU not detected at runtime in OpenMPTargetBuild, timeout in Clang+CUDA, and host func called from host device func warning treated as error in the CUDA debug build)

dalg24:
Please add to the change log under bug fix

yasahi-hpc:
> Please add to the change log under bug fix

Thanks! I added this PR to the change log

core/src/Kokkos_HostSpace.hpp
masterleinad: We only use this branch for a device execution space instance which isn't able to access the memory in question anyway. The only problem could arise if we have another host execution space (different from the one used here) accesses the memory from a different thread but we don't promise anything in that case anyway. If we happen to use the same execution space instance, we enqueue correctly.

PR #7248, opened 2024-08-22
core(graph): allow `create_graph` without closure

romintomasetti:
This PR adds an overload of `Kokkos::Experimental::create_graph` that does not take a closure.

This helps supporting advanced use cases for which creating the graph "in one shot" is not satisfactory.

### DIFF
--- a/core/src/Kokkos_Graph.hpp
+++ b/core/src/Kokkos_Graph.hpp
@@ -148,10 +148,20 @@ Graph<ExecutionSpace> create_graph(ExecutionSpace ex, Closure&& arg_closure) {
   return rv;
 }
 
+template <class ExecutionSpace = DefaultExecutionSpace>
+std::enable_if_t<Kokkos::is_execution_space_v<ExecutionSpace>,
+                 Graph<ExecutionSpace>>
+create_graph(ExecutionSpace exec = ExecutionSpace{}) {
+  return Kokkos::Impl::GraphAccess::construct_graph(std::move(exec));
+}
+
 template <
     class ExecutionSpace = DefaultExecutionSpace,
     class Closure = Kokkos::Impl::DoNotExplicitlySpecifyThisTemplateParameter>
-Graph<ExecutionSpace> create_graph(Closure&& arg_closure) {
+std::enable_if_t<
+    !Kokkos::is_execution_space_v<Kokkos::Impl::remove_cvref_t<Closure>>,
+    Graph<ExecutionSpace>>
+create_graph(Closure&& arg_closure) {
   return create_graph(ExecutionSpace{}, (Closure&&)arg_closure);
 }
 
--- a/core/unit_test/TestGraph.hpp
+++ b/core/unit_test/TestGraph.hpp
@@ -103,28 +103,40 @@ struct TEST_CATEGORY_FIXTURE(graph) : public ::testing::Test {
   }
 };
 
+// Check if a rank-0 view contains a given value.
+template <typename Exec, typename ViewType>
+::testing::AssertionResult contains(
+    const Exec& exec, const ViewType& view,
+    const typename ViewType::value_type& expected) {
+  static_assert(ViewType::rank() == 0);
+  typename ViewType::non_const_value_type value;
+  Kokkos::deep_copy(exec, value, view);
+  exec.fence();
+  if (value != expected)
+    return ::testing::AssertionFailure()
+           << expected << " is not in " << view.label() << ", got " << value;
+  else
+    return ::testing::AssertionSuccess();
+}
+
 TEST_F(TEST_CATEGORY_FIXTURE(graph), submit_once) {
   auto graph =
       Kokkos::Experimental::create_graph<TEST_EXECSPACE>([&](auto root) {
         root.then_parallel_for(1, count_functor{count, bugs, 0, 0});
       });
   graph.submit();
-  Kokkos::deep_copy(graph.get_execution_space(), count_host, count);
-  Kokkos::deep_copy(graph.get_execution_space(), bugs_host, bugs);
-  graph.get_execution_space().fence();
-  ASSERT_EQ(1, count_host());
-  ASSERT_EQ(0, bugs_host());
+
+  ASSERT_TRUE(contains(graph.get_execution_space(), count, 1));
+  ASSERT_TRUE(contains(graph.get_execution_space(), bugs, 0));
 }
 
 TEST_F(TEST_CATEGORY_FIXTURE(graph), submit_once_rvalue) {
   Kokkos::Experimental::create_graph(ex, [&](auto root) {
     root.then_parallel_for(1, count_functor{count, bugs, 0, 0});
   }).submit();
-  Kokkos::deep_copy(ex, count_host, count);
-  Kokkos::deep_copy(ex, bugs_host, bugs);
-  ex.fence();
-  ASSERT_EQ(1, count_host());
-  ASSERT_EQ(0, bugs_host());
+
+  ASSERT_TRUE(contains(ex, count, 1));
+  ASSERT_TRUE(contains(ex, bugs, 0));
 }
 
 // Ensure that Kokkos::Graph::instantiate works.
@@ -136,11 +148,9 @@ TEST_F(TEST_CATEGORY_FIXTURE(graph), instantiate_and_submit_once) {
   });
   graph.instantiate();
   graph.submit();
-  Kokkos::deep_copy(ex, count_host, count);
-  Kokkos::deep_copy(ex, bugs_host, bugs);
-  ex.fence();
-  ASSERT_EQ(1, count_host());
-  ASSERT_EQ(0, bugs_host());
+
+  ASSERT_TRUE(contains(ex, count, 1));
+  ASSERT_TRUE(contains(ex, bugs, 0));
 }
 
 // FIXME death tests and fixtures
@@ -203,11 +213,8 @@ TEST_F(TEST_CATEGORY_FIXTURE(graph),
 
   graph.submit(execution_space_instances.at(1));
 
-  Kokkos::deep_copy(execution_space_instances.at(1), count_host, count);
-  Kokkos::deep_copy(execution_space_instances.at(1), bugs_host, bugs);
-  execution_space_instances.at(1).fence();
-  ASSERT_EQ(1, count_host());
-  ASSERT_EQ(0, bugs_host());
+  ASSERT_TRUE(contains(execution_space_instances.at(1), count, 1));
+  ASSERT_TRUE(contains(execution_space_instances.at(1), bugs, 0));
 }
 
 TEST_F(TEST_CATEGORY_FIXTURE(graph), submit_six) {
@@ -245,12 +252,9 @@ TEST_F(TEST_CATEGORY_FIXTURE(graph), submit_six) {
     //----------------------------------------
   });
   graph.submit();
-  Kokkos::deep_copy(ex, count_host, count);
-  Kokkos::deep_copy(ex, bugs_host, bugs);
-  ex.fence();
 
-  ASSERT_EQ(6, count_host());
-  ASSERT_EQ(0, bugs_host());
+  ASSERT_TRUE(contains(ex, count, 6));
+  ASSERT_TRUE(contains(ex, bugs, 0));
 }
 
 TEST_F(TEST_CATEGORY_FIXTURE(graph), when_all_cycle) {
@@ -268,13 +272,10 @@ TEST_F(TEST_CATEGORY_FIXTURE(graph), when_all_cycle) {
         .then_parallel_reduce(6, set_result_functor{count}, reduction_out);
     //----------------------------------------
   }).submit();
-  Kokkos::deep_copy(ex, bugs_host, bugs);
-  Kokkos::deep_copy(ex, count_host, count);
-  Kokkos::deep_copy(ex, reduction_host, reduction_out);
-  ex.fence();
-  ASSERT_EQ(0, bugs_host());
-  ASSERT_EQ(7, count_host());
-  ASSERT_EQ(42, reduction_host());
+
+  ASSERT_TRUE(contains(ex, bugs, 0));
+  ASSERT_TRUE(contains(ex, count, 7));
+  ASSERT_TRUE(contains(ex, reduction_out, 42));
   //----------------------------------------
 }
 
@@ -342,9 +343,8 @@ TEST_F(TEST_CATEGORY_FIXTURE(graph), zero_work_reduce) {
   if constexpr (std::is_same_v<TEST_EXECSPACE, Kokkos::Cuda>) Kokkos::fence();
 #endif
   graph.submit();
-  Kokkos::deep_copy(ex, count_host, count);
-  ex.fence();
-  ASSERT_EQ(count_host(), 0);
+
+  ASSERT_TRUE(contains(ex, count, 0));
 }
 
 // Ensure that an empty graph can be submitted.
--- a/core/unit_test/TestGraph.hpp
+++ b/core/unit_test/TestGraph.hpp
@@ -217,6 +217,48 @@ TEST_F(TEST_CATEGORY_FIXTURE(graph),
   ASSERT_TRUE(contains(execution_space_instances.at(1), bugs, 0));
 }
 
+// This test ensures that it's possible to build a Kokkos::Graph using
+// Kokkos::Experimental::create_graph without providing a closure, but giving an
+// execution space instance.
+TEST_F(TEST_CATEGORY_FIXTURE(graph), create_graph_no_closure_with_exec) {
+  auto graph = Kokkos::Experimental::create_graph(ex);
+
+  auto root = Kokkos::Impl::GraphAccess::create_root_ref(graph);
+
+  auto node = root.then_parallel_for(1, count_functor{count, bugs, 0, 0});
+
+  graph.submit(ex);
+
+  ASSERT_TRUE(contains(ex, count, 1));
+  ASSERT_TRUE(contains(ex, bugs, 0));
+}
+
+// This test ensures that it's possible to build a Kokkos::Graph using
+// Kokkos::Experimental::create_graph without any argument.
+// The test has to be skipped if the test fixture is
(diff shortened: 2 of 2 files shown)


core/unit_test/TestGraph.hpp
```diff
@@ -174,6 +171,23 @@ TEST_F(TEST_CATEGORY_FIXTURE(graph), can_instantiate_only_once) {
...
 
+// This test ensures that it's possible to build a Kokkos::Graph using
+// Kokkos::Experimental::create_graph with no argument.
+TEST_F(TEST_CATEGORY_FIXTURE(graph), create_graph_no_arg) {
+  auto graph = Kokkos::Experimental::create_graph<TEST_EXECSPACE>(ex);
+
+  // I suspect that that root node should be an implementation detail of the defaulted implementation.
+  // For Cuda and HIP, if  node is added without any ancestor, it's OK.
+  // For SYCL I guess same story.
+  auto root = Kokkos::Impl::GraphAccess::create_root_ref(graph);
```
romintomasetti, line 182: I still need to think about this more.

crtrott (changes requested):
We need to have a in-depth discussion about this. There were very good design reasons why we forced everything to happen within a single closure mostly for safety and "don't allow people to shoot themselves in the foot" reasons. I would like us to discuss this and understand why we would feel differently in the risk/benefit analysis.

romintomasetti:
> We need to have a in-depth discussion about this. There were very good design reasons why we forced everything to happen within a single closure mostly for safety and "don't allow people to shoot themselves in the foot" reasons. I would like us to discuss this and understand why we would feel differently in the risk/benefit analysis.

@crtrott Sure! I'd be happy to discuss it whenever you're available :wink:

core/unit_test/TestGraph.hpp
```diff
@@ -101,6 +101,15 @@ struct TEST_CATEGORY_FIXTURE(graph) : public ::testing::Test {
...
   }
+
+  template <typename Exec>
+  void check(const Exec& exec, const int expt_count, const int expt_bugs) {
+    Kokkos::deep_copy(exec, count_host, count);
+    Kokkos::deep_copy(exec, bugs_host, bugs);
+    exec.fence();
+    ASSERT_EQ(expt_count, count_host());
+    ASSERT_EQ(expt_bugs, bugs_host());
+  }
```
dalg24, line 112: OK with abstracting away but this approach has a big downside that is that failure reports will always point to that function instead of the call site.
Maybe a macro could help.
Also you can probably do better for the name.

And you probably don't want to tie the fate of this PR to that (somewhat unrelated) change, do you?

core/unit_test/TestGraph.hpp
```diff
@@ -219,6 +216,20 @@ TEST_F(TEST_CATEGORY_FIXTURE(graph),
   ASSERT_EQ(0, bugs_host());
 }
 
+// This test ensures that it's possible to build a Kokkos::Graph using
+// Kokkos::Experimental::create_graph without providing a closure.
+TEST_F(TEST_CATEGORY_FIXTURE(graph), create_graph_no_closure) {
+  auto graph = Kokkos::Experimental::create_graph<TEST_EXECSPACE>(ex);
```
dalg24, line 222: Make sure you have one with an exec argument, and another assertion w/o it.

core/src/Kokkos_Graph.hpp
```diff
@@ -148,10 +148,15 @@ Graph<ExecutionSpace> create_graph(ExecutionSpace ex, Closure&& arg_closure) {
   return rv;
 }
 
+template <typename Exec>
```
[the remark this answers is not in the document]
romintomasetti: :heavy_check_mark:

core/src/Kokkos_Graph.hpp
```diff
@@ -148,10 +148,15 @@ Graph<ExecutionSpace> create_graph(ExecutionSpace ex, Closure&& arg_closure) {
...
+std::enable_if_t<Kokkos::is_execution_space_v<Exec>, Graph<Exec>> create_graph(const Exec& exec = Exec{}) {
```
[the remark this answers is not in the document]
romintomasetti: :heavy_check_mark:

core/unit_test/TestGraph.hpp
romintomasetti: :heavy_check_mark:

core/unit_test/TestGraph.hpp
romintomasetti: Clearly, if we want the call site to be clearly identified, we'll need to go through a macro. Otherwise, it will always point to the `check` method location (which is not bringing much information).

core/src/Kokkos_Graph.hpp
```diff
@@ -148,10 +148,19 @@ Graph<ExecutionSpace> create_graph(ExecutionSpace ex, Closure&& arg_closure) {
...
+                 Graph<ExecutionSpace>>
+create_graph(const ExecutionSpace& exec = ExecutionSpace{}) {
+  return create_graph(exec, [](const auto& /* root */) {});
+}
+
 template <
     class ExecutionSpace = DefaultExecutionSpace,
     class Closure = Kokkos::Impl::DoNotExplicitlySpecifyThisTemplateParameter>
-Graph<ExecutionSpace> create_graph(Closure&& arg_closure) {
+std::enable_if_t<!Kokkos::is_execution_space_v<std::decay_t<Closure>>,
```
dalg24, line 161: I'd prefer if we used `Impl::remove_cvref_t` instead of `std::decay_t`

core/unit_test/TestGraph.hpp
```diff
@@ -103,28 +103,30 @@ struct TEST_CATEGORY_FIXTURE(graph) : public ::testing::Test {
   }
 };
 
+/// If the counts are correct, all nodes have been submitted.
+/// However, it does not imply that dependencies were observed.
+#define ASSERT_THAT_ALL_NODES_WERE_SUBMITTED(__exec__, __expt_count__, \
+                                             __expt_bugs__)            \
```
dalg24, line 109: Even though you could argue macro arguments are not C++ identifiers, please avoid double underscores.
I typically prefer all uppercase names but we don't really have a hard convention.
"expt" is generally not an abbreviation for expected (I suppose that what you were going for) but for experiment so you might want to come up with something else.

dalg24: About the name of the macro.  It is not just that "nodes have been submitted" is it?

romintomasetti: - For macro arguments, my convention is...to break any convention. By adding the leading and trailing double underscore, I ensure that it's very unlikely that the macro's arguments match the name of any other variable.
- For the name, agreed. If the checks are fine, it means that the nodes also successfully executed their kernel.

romintomasetti: - But reading https://timsong-cpp.github.io/cppwp/lex.name#3... yeah, it might not be a good convention indeed.

core/src/Kokkos_Graph.hpp
romintomasetti: I'll do.

romintomasetti: Done, closing.

core/unit_test/TestGraph.hpp
romintomasetti: I just changed the name (and the description), and removed the `__`.

core/unit_test/TestGraph.hpp
```diff
@@ -210,6 +208,38 @@ TEST_F(TEST_CATEGORY_FIXTURE(graph),
   ASSERT_EQ(0, bugs_host());
```
dalg24, line 208: Looks like you missed an occurrence
https://github.com/kokkos/kokkos/blob/a998687f9f1192985546333aeeafd5f5b600c915/core/unit_test/TestGraph.hpp#L204-L208

core/unit_test/TestGraph.hpp
```diff
@@ -103,28 +103,30 @@ struct TEST_CATEGORY_FIXTURE(graph) : public ::testing::Test {
...
   auto graph =
       Kokkos::Experimental::create_graph<TEST_EXECSPACE>([&](auto root) {
         root.then_parallel_for(1, count_functor{count, bugs, 0, 0});
       });
   graph.submit();
-  Kokkos::deep_copy(graph.get_execution_space(), count_host, count);
-  Kokkos::deep_copy(graph.get_execution_space(), bugs_host, bugs);
-  graph.get_execution_space().fence();
-  ASSERT_EQ(1, count_host());
-  ASSERT_EQ(0, bugs_host());
```
dalg24, line 116: I am curious what other reviewers think.  I am not 100% sold that it helps with readability.

dalg24 (commented):
I would recommend splitting concerns

core/unit_test/TestGraph.hpp
romintomasetti: Indeed !

core/unit_test/TestGraph.hpp
romintomasetti: OK. I got your point.

I changed to creating a custom "check that rank-0 view contains a given value" assertion. So it reduces the number of "repetitive" lines and to me, it also makes things easier to read :mag:

core/unit_test/TestGraph.hpp
```diff
@@ -103,28 +103,37 @@ struct TEST_CATEGORY_FIXTURE(graph) : public ::testing::Test {
   }
 };
 
+// Check if a rank-0 view contains a given value.
+template <typename Exec, typename ViewType>
+::testing::AssertionResult contains(const Exec& exec, const ViewType& view, const typename ViewType::value_type& expected) {
+  static_assert(ViewType::rank == 0);
```
dalg24, line 109:
```suggestion
  static_assert(ViewType::rank() == 0);
```

core/unit_test/TestGraph.hpp
```diff
@@ -103,28 +103,37 @@ struct TEST_CATEGORY_FIXTURE(graph) : public ::testing::Test {
...
+  typename ViewType::non_const_value_type value;
+  Kokkos::deep_copy(exec, value, view);
+  exec.fence();
+  if(value != expected)
+    return ::testing::AssertionFailure() << expected << " is not in " << view.label();
+  else
+    return ::testing::AssertionSuccess();
```
dalg24, line 116: What does the final output on failure look like?

romintomasetti:
```bash
[ RUN      ] cuda_graph.submit_once_rvalue
/workspaces/kokkos/core/unit_test/TestGraph.hpp:135: Failure
Value of: contains(ex, count, 5)
  Actual: false (5 is not in count, got 1)
Expected: true
[  FAILED  ] cuda_graph.submit_once_rvalue (0 ms)
```

core/unit_test/TestGraph.hpp
romintomasetti: :heavy_check_mark:

core/unit_test/TestGraph.hpp
dalg24: is it critical that the execution space is specified?
contains -> rank_0_view_contains?

romintomasetti: I guess it is critical, because if your test is asynchronicity-critical and `contains` makes a global fence, your test might pass thought something is semantically wrong, right ?

romintomasetti: For the naming, I think we could keep `contains` and better constraint it so that it's clear from the signature it's for rank-0 views ? (We could wait for `C++20` to constraint the `ViewType` directory)

core/src/Kokkos_Graph.hpp
```diff
@@ -148,10 +148,20 @@ Graph<ExecutionSpace> create_graph(ExecutionSpace ex, Closure&& arg_closure) {
   return rv;
 }
 
+template <class ExecutionSpace = DefaultExecutionSpace>
+std::enable_if_t<Kokkos::is_execution_space_v<ExecutionSpace>,
+                 Graph<ExecutionSpace>>
+create_graph(const ExecutionSpace& exec = ExecutionSpace{}) {
+  return create_graph(exec, [](const auto& /* root */) {});
+}
```
romintomasetti, line 156: I need to fix this. For now, that would put the use in a situation wherein we've got 2 root nodes, but only one is connected to the user's nodes. The other one is useless.

romintomasetti: Done.

core/unit_test/TestGraph.hpp
```diff
@@ -203,11 +213,50 @@ TEST_F(TEST_CATEGORY_FIXTURE(graph),
...
+    auto node = root.then_parallel_for(1, count_functor{count, bugs, 0, 0});
+  }
+
+  graph.submit(graph.get_execution_space());
+
+  if constexpr (std::is_same_v<typename decltype(graph)::execution_space,
+                               TEST_EXECSPACE>) {
+    ASSERT_TRUE(contains(graph.get_execution_space(), count, 1));
+    ASSERT_TRUE(contains(graph.get_execution_space(), bugs, 0));
+  }
```
dalg24, line 259: Same comment about readability of the condition.
I really dislike our nor asserting anything in an else-branch or skipping

core/unit_test/TestGraph.hpp
```diff
@@ -203,11 +213,50 @@ TEST_F(TEST_CATEGORY_FIXTURE(graph),
...
-  ASSERT_EQ(0, bugs_host());
+  ASSERT_TRUE(contains(execution_space_instances.at(1), count, 1));
+  ASSERT_TRUE(contains(execution_space_instances.at(1), bugs, 0));
+}
+
+// This test ensures that it's possible to build a Kokkos::Graph using
+// Kokkos::Experimental::create_graph without providing a closure, but giving an
+// execution space instance.
+TEST_F(TEST_CATEGORY_FIXTURE(graph), create_graph_no_closure_with_exec) {
+  auto graph = Kokkos::Experimental::create_graph<TEST_EXECSPACE>(ex);
```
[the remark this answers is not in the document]
romintomasetti: Indeed, and done :heavy_check_mark:

core/unit_test/TestGraph.hpp
```diff
@@ -203,11 +213,50 @@ TEST_F(TEST_CATEGORY_FIXTURE(graph),
...
+
+  auto root = Kokkos::Impl::GraphAccess::create_root_ref(graph);
+
+  auto node = root.then_parallel_for(1, count_functor{count, bugs, 0, 0});
```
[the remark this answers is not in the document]
dalg24: If there was no warning then I suppose there is nothing to do

core/unit_test/TestGraph.hpp
romintomasetti: OK, so as you suggested this morning, this test should probably be skipped from the beginning if the `TEST_EXECSPACE` is not the same as `Kokkos::DefaultExecutionSpace` ?

dalg24: Unless there is a value in checking that is is legal to build a graph w/o an argument and that when it is legal to submit it as is.

core/unit_test/TestGraph.hpp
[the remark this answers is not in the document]
romintomasetti: OK, closing then.

core/unit_test/TestGraph.hpp
romintomasetti: @dalg24  OK to close ?

core/unit_test/TestGraph.hpp
romintomasetti: @dalg24 OK to close ?

core/unit_test/TestGraph.hpp
```diff
@@ -203,11 +213,50 @@ TEST_F(TEST_CATEGORY_FIXTURE(graph),
...
+TEST_F(TEST_CATEGORY_FIXTURE(graph), create_graph_no_arg) {
+  auto graph = Kokkos::Experimental::create_graph();
+
+  static_assert(std::is_same_v<typename decltype(graph)::execution_space,
+                               Kokkos::DefaultExecutionSpace>);
+
+  auto root = Kokkos::Impl::GraphAccess::create_root_ref(graph);
+
+  if constexpr (std::is_same_v<typename decltype(graph)::execution_space,
+                               TEST_EXECSPACE>) {
```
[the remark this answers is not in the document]
romintomasetti: With the test skip, this is not needed anymore. Closing.

core/unit_test/TestGraph.hpp
dalg24: Technically they are different but I don;t think this is critical

romintomasetti: OK, I get your point. But it seems that it is indeed not critical at this point, so let's close.

dalg24 (approved):

core/src/Kokkos_Graph.hpp
```diff
@@ -148,10 +148,20 @@ Graph<ExecutionSpace> create_graph(ExecutionSpace ex, Closure&& arg_closure) {
   return rv;
 }
 
+template <class ExecutionSpace = DefaultExecutionSpace>
+std::enable_if_t<Kokkos::is_execution_space_v<ExecutionSpace>,
+                 Graph<ExecutionSpace>>
+create_graph(ExecutionSpace exec = ExecutionSpace{}) {
+  return Kokkos::Impl::GraphAccess::construct_graph(std::move(exec));
```
cedricchevalier19, line 155: Is it ok to `std::move(exec)` when `exec` is not `ExecutionSpace&&` ?

I find the interface a little bit weird about ownership here, but it looks consistent with the previous code.

romintomasetti: I guess that `ExecutionSpace` is in this case always deduced as a "bare" type (*i.e.* `std::remove_cvref_t<ExecutionSpace> == ExecutionSpace`).

dalg24: It is fine.  This is a classic "pass by value and move" idiom.

Rombur (approved):

crtrott (approved):

PR #7249, opened 2024-08-22
core(graph): allow submission onto an arbitrary exec space instance

romintomasetti:
From a semantic standpoint, `Kokkos::Graph` must be used in a three-phases way.

1. :books: *definition* (add nodes and edges according to a DAG graph to create a *topological* graph)
2. :hotsprings: *instantiation* (check DAG for flaws and prepare the *executable* graph)
3. :fire: *submission* (launch the *executable* graph)

The submission can be viewed as throwing a bunch of kernels to the device, and as such must orchestrate them in some queue. Therefore, `submit` should accept an execution space instance.

### Going further

- [x] For some backends, adding nodes to the graph effectively triggers some asynchronous copies to devices under some circumstances (*e.g.* for `Cuda` we would async copy the driver to the global memory while creating the `Cuda` graph node). Proper fencing is therefore required before submission.
- [x] I still need to check that this makes sense with [P2300](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p2300r10.html).

### DIFF
--- a/core/src/Kokkos_Graph.hpp
+++ b/core/src/Kokkos_Graph.hpp
@@ -91,10 +91,12 @@ struct [[nodiscard]] Graph {
     (*m_impl_ptr).instantiate();
   }
 
-  void submit() const {
+  void submit(const execution_space& exec) const {
     KOKKOS_EXPECTS(bool(m_impl_ptr))
-    (*m_impl_ptr).submit();
+    (*m_impl_ptr).submit(exec);
   }
+
+  void submit() const { submit(get_execution_space()); }
 };
 
 // </editor-fold> end Graph }}}1
--- a/core/src/impl/Kokkos_Default_Graph_Impl.hpp
+++ b/core/src/impl/Kokkos_Default_Graph_Impl.hpp
@@ -141,7 +141,8 @@ struct GraphImpl : private ExecutionSpaceInstanceStorage<ExecutionSpace> {
     m_has_been_instantiated = true;
   }
 
-  void submit() {
+  // FIXME The execution space instance should be passed to the nodes.
+  void submit(const ExecutionSpace& /* exec */) {
     if (!m_has_been_instantiated) instantiate();
     // This reset is gross, but for the purposes of our simple host
     // implementation...
--- a/core/unit_test/TestGraph.hpp
+++ b/core/unit_test/TestGraph.hpp
@@ -174,6 +174,40 @@ TEST_F(TEST_CATEGORY_FIXTURE(graph), can_instantiate_only_once) {
   }
 }
 
+// This test submits on an execution space instance different from the
+// one passed to the Kokkos::Graph constructor.
+TEST_F(TEST_CATEGORY_FIXTURE(graph),
+       submit_onto_another_execution_space_instance) {
+// FIXME For ROCm 5.2, the graph implementation is bugged, and Kokkos
+//       therefore uses the default implementation. The kernels of the
+//       graph will thus run on the default stream.
+#if defined(KOKKOS_ENABLE_HIP)
+#if HIP_VERSION_MAJOR == 5 && HIP_VERSION_MINOR == 2
+  if constexpr (std::is_same_v<TEST_EXECSPACE, Kokkos::HIP>)
+    GTEST_SKIP() << "Graph is not properly enqueued.";
+#endif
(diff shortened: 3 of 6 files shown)


Rombur:
Clang-format is not happy

masterleinad:
`clang-format` is still complaining. Are you using `clang-format 16`?

romintomasetti:
I guess I just forgot one last suggested fix from the previous pipeline... :innocent:

Rombur:
Retest this please

Rombur (approved):

core/src/Kokkos_Graph.hpp
```diff
@@ -91,10 +91,12 @@ struct [[nodiscard]] Graph {
...
   }
 
-  void submit() const {
+  void submit(const execution_space& exec) const {
     KOKKOS_EXPECTS(bool(m_impl_ptr))
-    (*m_impl_ptr).submit();
+    (*m_impl_ptr).submit(exec);
   }
+
+  void submit() const { submit(get_execution_space()); }
```
dalg24, line 99: Any reason to prefer a new overload over a default argument?

core/unit_test/TestGraph.hpp
```diff
@@ -174,6 +174,31 @@ TEST_F(TEST_CATEGORY_FIXTURE(graph), can_instantiate_only_once) {
   }
 }
 
+// This test submits on an execution space instance different from the
+// one passed to the Kokkos::Graph constructor.
+TEST_F(TEST_CATEGORY_FIXTURE(graph),
+       submit_onto_another_execution_space_instance) {
```
dalg24, line 180: Potentially skip it if the exec space is HIP and the ROCM version is 5.2

core/src/Kokkos_Graph.hpp
romintomasetti: None for now... I'll change that :wink:

romintomasetti: So, I wanted to write
```c++
  void submit(const execution_space& exec = get_execution_space()) const;
```
but as I expected, this isn't allowed, because
```bash
error: a nonstatic member reference must be relative to a specific object
```

romintomasetti: So we'll need both overloads if we want to continue supporting `submit()`.

core/unit_test/TestGraph.hpp
romintomasetti: Under `ROCm 5.2`, `Kokkos` uses the default graph implementation instead of the specialization for the `HIP` backend:https://github.com/kokkos/kokkos/blob/ed20bf11df3bedecff46d2bef12f5a97f486a769/core/src/Kokkos_Graph.hpp#L170-L173

Therefore, the test makes no sense and can be skipped.

@Rombur @dalg24

masterleinad (approved):
Fine with me.

romintomasetti:
It seems Jenkins failures are unrelated. Do you confirm ?

core/unit_test/TestGraph.hpp
```diff
@@ -174,6 +174,40 @@ TEST_F(TEST_CATEGORY_FIXTURE(graph), can_instantiate_only_once) {
...
+// FIXME_HIP For ROCm 5.2, the graph implementation is bugged, and Kokkos
```
dalg24, line 181: Fine but noting that the FIXME is not quite a HIP thing but seems to denote a bug in the fallback implementation

dalg24 (approved):

core/src/impl/Kokkos_Default_Graph_Impl.hpp
```diff
@@ -141,7 +141,7 @@ struct GraphImpl : private ExecutionSpaceInstanceStorage<ExecutionSpace> {
     m_has_been_instantiated = true;
   }
 
-  void submit() {
+  void submit(const ExecutionSpace& /* exec */) {
```
dalg24, line 144: That right here should be a big fat FIXME

romintomasetti: Done.

dalg24:
Unrelated failures

dalg24:
Please add to the change log

PR #7250, opened 2024-08-22
Address NVHPC CI warnings

tcclevenger:
Address NVHPC CI warnings and add `-Werror` to jenkins build.

### DIFF
--- a/core/src/Kokkos_Macros.hpp
+++ b/core/src/Kokkos_Macros.hpp
@@ -607,6 +607,18 @@ static constexpr bool kokkos_omp_on_host() { return false; }
   #define KOKKOS_IMPL_DISABLE_DEPRECATED_WARNINGS_PUSH()
   #define KOKKOS_IMPL_DISABLE_DEPRECATED_WARNINGS_POP()
 #endif
+
+#if defined(__NVCOMPILER)
+#define KOKKOS_IMPL_DISABLE_UNREACHABLE_WARNINGS_PUSH() \
+  _Pragma("diag_suppress code_is_unreachable")          \
+  _Pragma("diag_suppress initialization_not_reachable")
+#define KOKKOS_IMPL_DISABLE_UNREACHABLE_WARNINGS_POP() \
+  _Pragma("diag_default code_is_unreachable")          \
+  _Pragma("diag_default initialization_not_reachable")
+#else
+#define KOKKOS_IMPL_DISABLE_UNREACHABLE_WARNINGS_PUSH()
+#define KOKKOS_IMPL_DISABLE_UNREACHABLE_WARNINGS_POP()
+#endif
 // clang-format on
 
 #define KOKKOS_ATTRIBUTE_NODISCARD [[nodiscard]]
--- a/core/unit_test/TestNumericTraits.hpp
+++ b/core/unit_test/TestNumericTraits.hpp
@@ -405,11 +405,10 @@ TEST(TEST_CATEGORY, numeric_traits_min_max_exponent10) {
   TestNumericTraits<TEST_EXECSPACE, long double, MaxExponent10>();
 #endif
 }
-TEST(TEST_CATEGORY, numeric_traits_quiet_and_signaling_nan) {
+
 // FIXME_NVHPC
-#ifdef KOKKOS_COMPILER_NVHPC
-  GTEST_SKIP() << "This test is known to fail with the NVHPC compiler";
-#endif
+#ifndef KOKKOS_COMPILER_NVHPC
+TEST(TEST_CATEGORY, numeric_traits_quiet_and_signaling_nan) {
   TestNumericTraits<TEST_EXECSPACE, Kokkos::Experimental::half_t, QuietNaN>();
   TestNumericTraits<TEST_EXECSPACE, Kokkos::Experimental::half_t,
                     SignalingNaN>();
@@ -426,6 +425,7 @@ TEST(TEST_CATEGORY, numeric_traits_quiet_and_signaling_nan) {
   TestNumericTraits<TEST_EXECSPACE, long double, SignalingNaN>();
 #endif
 }
+#endif
 
 namespace NumericTraitsSFINAE {
 
--- a/core/unit_test/TestNumericTraits.hpp
+++ b/core/unit_test/TestNumericTraits.hpp
@@ -406,9 +406,12 @@ TEST(TEST_CATEGORY, numeric_traits_min_max_exponent10) {
 #endif
 }
 
-// FIXME_NVHPC
-#ifndef KOKKOS_COMPILER_NVHPC
+KOKKOS_IMPL_DISABLE_UNREACHABLE_WARNINGS_PUSH()
 TEST(TEST_CATEGORY, numeric_traits_quiet_and_signaling_nan) {
+#ifdef KOKKOS_COMPILER_NVHPC  // FIXME_NVHPC
+  GTEST_SKIP() << "This test is known to fail with the NVHPC compiler";
+#endif
+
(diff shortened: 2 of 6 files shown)


tpls/mdspan/include/experimental/__p2642_bits/layout_padded.hpp
```diff
@@ -59,8 +59,8 @@ MDSPAN_INLINE_FUNCTION constexpr size_t get_actual_static_padding_value() {
   } else {
     return dynamic_extent;
   }
-  // Missing return statement warning from NVCC and ICC
-#if defined(__NVCC__) || defined(__INTEL_COMPILER)
+  // Missing return statement warning from ICC
+#if defined(__INTEL_COMPILER)
```
tcclevenger, line 63: This may be naive to think we can remove for nvcc. But in src code we seem to only need these returns for intel compiler. E.g. https://github.com/kokkos/kokkos/blob/b367a71b8e9f7d4f1c89231f5b4c6ef09ee77626/core/src/Kokkos_View.hpp#L977

Something that could be related is 
https://github.com/kokkos/kokkos/blob/b367a71b8e9f7d4f1c89231f5b4c6ef09ee77626/tpls/mdspan/include/experimental/__p2630_bits/submdspan_mapping.hpp#L159 
but but I'm not sure how to trigger this setup (when do we have no CUDA_ARCH and nvcc)?

masterleinad: Just disable the check when compiling with NVHPC, i.e, when `__NVCOMPILER` is defined.

masterleinad: > when do we have no CUDA_ARCH and nvcc?

If nvcc is compiling the host path.

masterleinad: > This may be naive to think we can remove for nvcc.

We definitely need this for older `nvcc` versions.

seyonglee (approved):
It looks good to me.

core/unit_test/TestNumericTraits.hpp
```diff
@@ -405,11 +405,10 @@ TEST(TEST_CATEGORY, numeric_traits_min_max_exponent10) {
...
 #endif
 }
-TEST(TEST_CATEGORY, numeric_traits_quiet_and_signaling_nan) {
+
 // FIXME_NVHPC
-#ifdef KOKKOS_COMPILER_NVHPC
-  GTEST_SKIP() << "This test is known to fail with the NVHPC compiler";
-#endif
+#ifndef KOKKOS_COMPILER_NVHPC
+TEST(TEST_CATEGORY, numeric_traits_quiet_and_signaling_nan) {
```
dalg24, line 411: This is quite heavy handed, more like a last resort resolution.
What warning were we getting?

tcclevenger: I'll check each of the tests that are excluded.

crtrott: did you do this because you got warnings during the build from the test and the warnings were not fixable?

tcclevenger: To be clear, I didn't exclude this test as part of the PR, just changed how it was being excluded due to a warning. There are a few like this in the PR where you have 
```
TEST(TEST_CATEGORY, some_test) {
#ifdef KOKKOS_COMPILER_NVHPC  // FIXME: KOKKOS_COMPILER_NVHPC
  GTEST_SKIP() << "some error msg about not working with KOKKOS_COMPILER_NVHPC";
#endif

  // Test body
}
``` 
It complains that the `// Test body` code was not reachable because we are always exiting at `GTEST_SKIP()`. So instead, I'm ifdef-ing out the test (like is done other places in unit tests, I believe) and keeping the `// FIXME: KOKKOS_COMPILER_NVHPC

tcclevenger: So nothing is changed in this PR with what testing is run. I think addressing the test fails would make sense in a different PR.

masterleinad: Any chance that you can use `pragma push/pop` or disable `unreachable` warnings generally in the CI?

tcclevenger: I'm sure we could do something like add `--diag_suppress=code_is_unreachable` flag, but would we want to ignore these instances in source code? Or do we not care if there is unreachable code?

crtrott: What Daniel is asking to see if we can selectively in those places disable the warning. Generally we prefer the gtest skip thing over ifdefing out the whole thing because it means the tests will show up as skipped instead of just not existing in gtest output.

tcclevenger: Ah ok, didn't realize the skips provided useful info. Yeah, let me find a nice solution.

core/unit_test/incremental/Test12b_TeamScratch.hpp
```diff
@@ -86,8 +86,13 @@ struct TeamScratch {
   }
 };
 
-#ifndef KOKKOS_ENABLE_OPENACC  // FIXME_OPENACC, scratch memory not implemented
```
dalg24: ??

core/src/Kokkos_Macros.hpp
```diff
@@ -607,6 +607,18 @@ static constexpr bool kokkos_omp_on_host() { return false; }
   #define KOKKOS_IMPL_DISABLE_DEPRECATED_WARNINGS_PUSH()
   #define KOKKOS_IMPL_DISABLE_DEPRECATED_WARNINGS_POP()
 #endif
+
+#if defined(__NVCOMPILER)
+#define KOKKOS_IMPL_DISABLE_UNREACHABLE_WARNINGS_PUSH() \
+  _Pragma("diag_suppress code_is_unreachable")          \
+  _Pragma("diag_suppress initialization_not_reachable")
```
dalg24, line 614: We actually need both?

tcclevenger:
Added a Macro `KOKKOS_IMPL_DISABLE_UNREACHABLE_WARNINGS_PUSH/POP` similar to `KOKKOS_IMPL_DISABLE_DEPRECATED_WARNINGS_PUSH/POP`. Currently the only use case is `__NVCOMPILER` but if we like this solution there are many tests using other compilers that can be added so that we can get the `SKIPPED` output for tests.

tpls/mdspan/include/experimental/__p2642_bits/layout_padded.hpp
dalg24, line 1: These changes needs to be proposed to kokkos/mdspan

dalg24 (commented):
Looks good but we need to cherry-pick the mdspan changes upstream first

core/src/Kokkos_Macros.hpp
tcclevenger: I couldn't find a catch all warning for unreachable code for nvc++. We need `initialization_not_reachable` because of 
```
#ifdef KOKKOS_ENABLE_OPENACC  // FIXME_OPENACC
  GTEST_SKIP() << "skipping since scratch memory is not yet implemented in the "
                  "OpenACC backend";
#endif

  ThreadScratch<TEST_EXECSPACE> test; // warning thrown here for initing test
  ...
}
```

tpls/mdspan/include/experimental/__p2642_bits/layout_padded.hpp
tcclevenger: See https://github.com/kokkos/mdspan/pull/356

masterleinad (approved):
Looks good to me.

aprokop (approved):

tcclevenger:
OpenACC-NVCC fail unrelated.

PR #7255, opened 2024-08-23
Move Kokkos_View.hpp in prep for mdspan overhaul

crtrott:
We will have an option for some time to switch between the legacy and the new implementation (based on mdspan). Switching between them will largely just change which implementation is visible, with the legacy implementation isolated to the existing file. The new implementation will go into the Kokkos_View.hpp then. 

I.e.: 
```c++
#define KOKKOS_VIEW_HPP

#ifdef KOKKOS_ENABLE_IMPL_VIEW_LEGACY
#include <View/Kokkos_ViewLegacy.hpp>
#else

....

#endif /* KOKKOS_ENABLE_IMPL_VIEW_LEGACY */
#endif /* KOKKOS_VIEW_HPP */
```

### DIFF
--- a/core/src/Kokkos_View.hpp
+++ b/core/src/Kokkos_View.hpp
@@ -1,0 +1,27 @@
+//@HEADER
+// ************************************************************************
+//
+//                        Kokkos v. 4.0
+//       Copyright (2022) National Technology & Engineering
+//               Solutions of Sandia, LLC (NTESS).
+//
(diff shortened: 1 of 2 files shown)


crtrott:
It helps reviewing the two commits individually. Because then you see that it was really just moved.

nliber (approved):

ldh4 (approved):

masterleinad (approved):

PR #7265, opened 2024-08-26
Do not showcase using facility from `Impl::` in examples

dalg24:
Prefer `std::conditional_t` to `Impl::if_c`, even more so when it is an example

This is the only instance I could find in our examples and benchmarks


masterleinad (approved):

romintomasetti (approved):

dalg24:
Ignoring warning host only function called from host device function warning treated as an error in one of the CUDA build.
Everything else passed! ...

PR #7266, opened 2024-08-26
Prefer `std::conditional_t<T>` to `typename Impl::if_c<T>::type`

dalg24:
Prefer standard facility to homegrown solution that it would be nice to get rid of.
We are left with two use cases that use a different functionality of `Impl::if_c` that need more thought to refactor.
```
core/src/Kokkos_GraphNode.hpp:380:    using reducer_selector = Kokkos::Impl::if_c<
core/src/Kokkos_Parallel_Reduce.hpp:1501:        Kokkos::Impl::if_c<std::is_same_v<InvalidType, PassedReducerType>,

### DIFF
--- a/core/src/Threads/Kokkos_Threads_Team.hpp
+++ b/core/src/Threads/Kokkos_Threads_Team.hpp
@@ -143,8 +143,8 @@ class ThreadsExecTeamMember {
 
     KOKKOS_IF_ON_HOST((
         // Make sure there is enough scratch space:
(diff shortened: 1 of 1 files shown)


romintomasetti (approved):

dalg24:
Planning to ignore unrelated failures (GPU not detected at runtime in Clang+CUDA build and warning host from host device warning treated as error in that other CUDA build that we need to fix)

yasahi-hpc (approved):

masterleinad (approved):

PR #7267, opened 2024-08-27
update `Random_UniqueIndexget_state_idx to take locks_view_type by reference

jacobmerson:
Since the `lockview` type was not passed by reference this was causing a construction and destruction of


dalg24 (approved):

dalg24:
Failures on Jenkins CI are unrelated

dalg24:
To clarify we don't expect that reference counting happens but we don't see an issue with this change

Issue #1011, opened 2017-08-01
Kokkos::complex<T> behaves slightly differently from std::complex<T>


bjoo:
Just came accross this issue today. In some cases I want to set the real and imaginary parts of a Kokkos::complex<T> separately. This works in the intuitive way e.g.

Kokkos::complex<float> f;
f.real() = 3.0;
f.imag() = 4.0;

However, when I tried to switch over to std::complex<T> this no longer worked. Apparently 
now the only way to do this is:

std::complex<float> f;
f.real(3.0);
f.imag(4.0);

When I tried this with Kokkos::complex it didn't work:

>/global/homes/b/bjoo/MG/mg_proto_vec/test/kokkos/./kokkos_vectype.h(222): error: no instance >of overloaded function "Kokkos::complex<RealType>::imag [with RealType=REAL={REAL64=?{double}}]" matches the argument list
>            argument types are: (double)
>            object type is: MG::MyComplex<REAL>
>		adata[i].imag( adata[i].imag() - sign*bdata[i].real() );
  		         ^
>/global/homes/b/bjoo/MG/mg_proto_vec/kokkos/core/src/Kokkos_Complex.hpp(221): note: this >candidate was rejected because mismatch in count of arguments
>    KOKKOS_INLINE_FUNCTION const RealType imag () const volatile {

(here MG::MyComplex<REAL> is aliased to Kokkos::Complex<REAL> using using)

Ultimately this may not matter as folks should use Kokkos::complex, but it may trip up users migrating form std::complex . For what its worth, I prefer the Kokkos version as I can do things like 

> a.real() += b.imag();

mhoemmen:
Hi @bjoo !  Would you like to submit a pull request (against Kokkos' develop branch) in order to fix this?  Also, is the requested feature standard with C++11?  Does it work with GCC 4.8.4?

ibaned:
Here is [a relevant page of C++ documentation](http://en.cppreference.com/w/cpp/numeric/complex/real) with annotations about how things change over standards.
I suspect the way they did it this way (getters and setters) is that there is no guarantee that `std::complex` internally actually just contains two members of type `T`, it may be using some other weird encoding. That said, the getters are `constexpr` since C++11, so it can't be that weird.

ibaned:
Just added these functions, tested them, and got them approved into Kokkos develop. Record time!

bjoo:
Thank you very much Dan. I will change my code to use these new versions. As I mentioned, to 
[...]

PR #7269, opened 2024-08-27
Fix bogus icpc -Werror

ndellingwood:
Workaround error #1011: missing return statement at end of non-void function


dalg24 (approved):

dalg24:
Unrelated failures

PR #7270, opened 2024-08-28
Add proper testing for kernel names passed to the tools interface

dalg24:
Check that, the user-provided label `parallel_{for,reduce,scan}(<label>, ...)` is passed to the tool and that, when it is absent, the default is based on the information from the functor/lambda `typeid`.

This work is a preliminary step towards #7043 that would change that `std::type_info`-based fallback kernel name to something else that does not rely on RTTI (see #7037)

### DIFF
--- a/core/unit_test/CMakeLists.txt
+++ b/core/unit_test/CMakeLists.txt
@@ -1082,6 +1082,7 @@ endif()
     UnitTestMainInit.cpp
     tools/TestEventCorrectness.cpp
     tools/TestWithoutInitializing.cpp
+    tools/TestKernelNames.cpp
     tools/TestProfilingSection.cpp
     tools/TestScopedRegion.cpp
     )
--- a/core/unit_test/CMakeLists.txt
+++ b/core/unit_test/CMakeLists.txt
@@ -1081,10 +1081,10 @@ endif()
   SET(KOKKOSP_SOURCES
     UnitTestMainInit.cpp
     tools/TestEventCorrectness.cpp
-    tools/TestWithoutInitializing.cpp
     tools/TestKernelNames.cpp
     tools/TestProfilingSection.cpp
     tools/TestScopedRegion.cpp
+    tools/TestWithoutInitializing.cpp
     )
 
   # FIXME_OPENMPTARGET This test causes internal compiler errors as of 09/01/22
--- a/core/unit_test/TestRange.hpp
+++ b/core/unit_test/TestRange.hpp
@@ -60,32 +60,9 @@ struct TestRange {
     Kokkos::parallel_for(Kokkos::RangePolicy<ExecSpace, ScheduleType>(0, N),
                          *this);
 
-    {
-      using ThisType = TestRange<ExecSpace, ScheduleType>;
-      std::string label("parallel_for");
-      Kokkos::Impl::ParallelConstructName<ThisType, void> pcn(label);
-      ASSERT_EQ(pcn.get(), label);
-      std::string empty_label("");
-      Kokkos::Impl::ParallelConstructName<ThisType, void> empty_pcn(
-          empty_label);
-      ASSERT_EQ(empty_pcn.get(), typeid(ThisType).name());
-    }
-
     Kokkos::parallel_for(
         Kokkos::RangePolicy<ExecSpace, ScheduleType, VerifyInitTag>(0, N),
         *this);
-
-    {
-      using ThisType = TestRange<ExecSpace, ScheduleType>;
-      std::string label("parallel_for");
-      Kokkos::Impl::ParallelConstructName<ThisType, VerifyInitTag> pcn(label);
-      ASSERT_EQ(pcn.get(), label);
-      std::string empty_label("");
-      Kokkos::Impl::ParallelConstructName<ThisType, VerifyInitTag> empty_pcn(
-          empty_label);
-      ASSERT_EQ(empty_pcn.get(), std::string(typeid(ThisType).name()) + "/" +
-                                     typeid(VerifyInitTag).name());
-    }
 
     Kokkos::deep_copy(host_flags, m_flags);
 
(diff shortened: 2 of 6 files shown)


core/unit_test/CMakeLists.txt
```diff
@@ -1082,6 +1082,7 @@ endif()
     UnitTestMainInit.cpp
     tools/TestEventCorrectness.cpp
     tools/TestWithoutInitializing.cpp
+    tools/TestKernelNames.cpp
```
romintomasetti, line 1085: Should you place it better in the list ? It seems these lists are usually given in lexicographic order.

core/unit_test/tools/TestKernelNames.cpp
```diff
@@ -0,0 +1,170 @@
...
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
+#include <Kokkos_Core.hpp>
+
+#include <gtest/gtest.h>
+
+namespace {
```
romintomasetti, line 21: For consistency ?
```suggestion
namespace Test {
```

core/unit_test/tools/TestKernelNames.cpp
```diff
@@ -0,0 +1,170 @@
...
+  }
+
+  Kokkos::Tools::Experimental::set_begin_parallel_scan_callback(nullptr);
+}
+
+TEST(defaultdevicetype, kernel_name_parallel_for) {
+  test_kernel_name_parallel_for();
+}
+
+TEST(defaultdevicetype, kernel_name_parallel_reduce) {
```
romintomasetti, line 162: Any reason for hardcoding your test to the default execution space type ?

core/unit_test/tools/TestKernelNames.cpp
```diff
@@ -0,0 +1,170 @@
...
+}
+
+void test_kernel_name_parallel_scan() {
+  Kokkos::Tools::Experimental::set_begin_parallel_scan_callback(
+      get_parallel_scan_kernel_name);
+
+  using ExecutionSpace = Kokkos::DefaultExecutionSpace;
+
+  {
+    std::string my_label = "my_parallel_scan_range_policy";
```
romintomasetti, line 132:
```suggestion
    const std::string my_label = "my_parallel_scan_range_policy";
```
(similar occurences elsewhere)

core/unit_test/tools/TestKernelNames.cpp
```diff
@@ -0,0 +1,170 @@
...
+                            my_lambda, my_result);
+    ASSERT_EQ(last_parallel_reduce, my_label);
+
+    Kokkos::parallel_reduce(Kokkos::RangePolicy<ExecutionSpace>(0, 1),
+                            my_lambda, my_result);
+    ASSERT_NE(last_parallel_reduce.find(typeid(my_lambda).name()),
+              std::string::npos)
+        << last_parallel_reduce << " does not contain "
+        << typeid(my_lambda)
+               .name();  // internally using
```
romintomasetti, line 100: Wouldn't the test output in case of failure better if using `HasSubstr` from https://google.github.io/googletest/reference/matchers.html#string-matchers ?

romintomasetti: Along these lines:
```suggestion
    ASSERT_THAT(last_parallel_reduce, ::testing::HasSubstr(typeid(my_lambda).name()));
```

core/unit_test/tools/TestKernelNames.cpp
dalg24: Why?

dalg24: The consistent better thing to do is anonymous namespace

core/unit_test/CMakeLists.txt
dalg24: Didn't really feel like ordering the whole list and would rather not do it there

core/unit_test/tools/TestKernelNames.cpp
dalg24: That is a good suggestion, except we can't use GoogleMock yet

core/unit_test/tools/TestKernelNames.cpp
dalg24: These are all hard coded in the KokkosP unit test but as it turns out I was supposed to use `kokkosp`?

dalg24: Actually I can find both `kokkosp` and `defaultdevicetype` in there

core/unit_test/tools/TestKernelNames.cpp
dalg24: I am generally in favor of adding GMock but this likely would not play well with the ongoing TriBITS refactor and would be getting somewhat trivial after we are done with it

core/unit_test/TestRange.hpp
```diff
@@ -60,33 +60,10 @@ struct TestRange {
...
-      using ThisType = TestRange<ExecSpace, ScheduleType>;
-      std::string label("parallel_for");
-      Kokkos::Impl::ParallelConstructName<ThisType, void> pcn(label);
-      ASSERT_EQ(pcn.get(), label);
-      std::string empty_label("");
-      Kokkos::Impl::ParallelConstructName<ThisType, void> empty_pcn(
-          empty_label);
-      ASSERT_EQ(empty_pcn.get(), typeid(ThisType).name());
-    }
-
```
crtrott, line 73: Why delete these ones (instead of e.g. moving them). I mean sure we test the end result but in principle I wouldn't mind having actual unit tests for implementation details too,

crtrott (approved):
I do like the new tests, simple and effective.

core/unit_test/TestRange.hpp
dalg24: It is an `Impl::` facility, it does not guarantee that it is properly called in our parallel constructs.

crtrott: Yeah obviously, but my point is having it too tells you whether the name generation failed in some way, or whether the passing on failed.

dalg24: Re-added them into the new test file

dalg24:
Failures are clearly unrelated (machine issues)

PR #7271, opened 2024-08-29
graph(fix): defaulted graph submit control flow

romintomasetti:
This PR fixes the defaulted graph implementation `submit` control flow by ensuring that:
1. `submit` fences the received execution space instance right away before launching any work.
2. Nodes wait for their predecessors to finish (`awaitable` concept needed to make things readable).
3. `submit` waits for completion of all sinks before returning control flow to user.

This should solve:
1. The problem with `HIP 5.2` detected in #7249.
2. The problem with `HPX` backend (hopefully).

### DIFF
--- a/core/src/impl/Kokkos_Default_GraphNodeKernel.hpp
+++ b/core/src/impl/Kokkos_Default_GraphNodeKernel.hpp
@@ -36,15 +36,22 @@ struct GraphNodeKernelDefaultImpl {
   // TODO @graphs decide if this should use vtable or intrusive erasure via
   //      function pointers like in the rest of the graph interface
   virtual void execute_kernel() = 0;
+
+  GraphNodeKernelDefaultImpl() = default;
+
+  explicit GraphNodeKernelDefaultImpl(ExecutionSpace exec)
+      : m_execution_space(std::move(exec)) {}
+
+  ExecutionSpace m_execution_space;
 };
 
 // TODO Indicate that this kernel specialization is only for the Host somehow?
 template <class ExecutionSpace, class PolicyType, class Functor,
           class PatternTag, class... Args>
 class GraphNodeKernelImpl
-    : public PatternImplSpecializationFromTag<PatternTag, Functor, PolicyType,
-                                              Args..., ExecutionSpace>::type,
-      public GraphNodeKernelDefaultImpl<ExecutionSpace> {
+    : public GraphNodeKernelDefaultImpl<ExecutionSpace>,
+      public PatternImplSpecializationFromTag<PatternTag, Functor, PolicyType,
+                                              Args..., ExecutionSpace>::type {
  public:
   using base_t =
       typename PatternImplSpecializationFromTag<PatternTag, Functor, PolicyType,
@@ -61,9 +68,9 @@ class GraphNodeKernelImpl
   GraphNodeKernelImpl(std::string const &, ExecutionSpace const &,
                       Functor arg_functor, PolicyDeduced &&arg_policy,
                       ArgsDeduced &&...args)
-      : base_t(std::move(arg_functor), (PolicyDeduced &&)arg_policy,
-               (ArgsDeduced &&)args...),
-        execute_kernel_vtable_base_t() {}
+      : execute_kernel_vtable_base_t(arg_policy.space()),
+        base_t(std::move(arg_functor), (PolicyDeduced &&)arg_policy,
+               (ArgsDeduced &&)args...) {}
 
   // FIXME @graph Forward through the instance once that works in the backends
   template <class PolicyDeduced, class... ArgsDeduced>
@@ -71,7 +78,9 @@ class GraphNodeKernelImpl
                       PolicyDeduced &&arg_policy, ArgsDeduced &&...args)
       : GraphNodeKernelImpl("", ex, std::move(arg_functor),
                             (PolicyDeduced &&)arg_policy,
-                            (ArgsDeduced &&)args...) {}
+                            (ArgsDeduced &&)args...) {
+    // FIXME This constructor seem unused.
+  }
 
   void execute_kernel() final { this->base_t::execute(); }
 };
--- a/core/src/impl/Kokkos_Default_GraphNode_Impl.hpp
+++ b/core/src/impl/Kokkos_Default_GraphNode_Impl.hpp
@@ -92,6 +92,18 @@ struct GraphNodeBackendSpecificDetails {
     m_is_aggregate = true;
   }
 
+  // A node is awaitable if it can execute a kernel.
+  // A root node or an aggregate node cannot be waited for, because it does
+  // not launch anything.
+  bool awaitable() const { return (!m_is_root) && (!m_is_aggregate); }
+
+  // Retrieve the execution space instance that has been passed to
+  // the kernel at construction phase.
+  const ExecutionSpace& get_execution_space() const {
+    KOKKOS_EXPECTS(m_kernel_ptr != nullptr)
+    return m_kernel_ptr->m_execution_space;
+  }
+
   void set_predecessor(
       std::shared_ptr<GraphNodeBackendSpecificDetails<ExecutionSpace>>
           arg_pred_impl) {
@@ -104,7 +116,7 @@ struct GraphNodeBackendSpecificDetails {
     m_predecessors.push_back(std::move(arg_pred_impl));
   }
 
-  void execute_node() {
+  void execute_node(const ExecutionSpace& exec) {
     // This node could have already been executed as the predecessor of some
     // other
     KOKKOS_EXPECTS(bool(m_kernel_ptr) || m_has_executed)
@@ -115,8 +127,18 @@ struct GraphNodeBackendSpecificDetails {
       // supported semantics, but instinct I have feels like it should be...
       m_has_executed = true;
       for (auto const& predecessor : m_predecessors) {
-        predecessor->execute_node();
+        predecessor->execute_node(exec);
       }
+
+      // Before executing the kernel, be sure to fence the execution space
+      // instance of predecessors.
+      for (const auto& predecessor : m_predecessors) {
+        if (predecessor->awaitable() &&
+            predecessor->get_execution_space() != this->get_execution_space())
+          predecessor->get_execution_space().fence(
+              "Kokkos::DefaultGraphNode::execute_node: sync with predecessors");
+      }
+
       m_kernel_ptr->execute_kernel();
     }
     KOKKOS_ENSURES(m_has_executed)
--- a/core/src/impl/Kokkos_Default_Graph_Impl.hpp
+++ b/core/src/impl/Kokkos_Default_Graph_Impl.hpp
@@ -141,16 +141,29 @@ struct GraphImpl : private ExecutionSpaceInstanceStorage<ExecutionSpace> {
     m_has_been_instantiated = true;
   }
 
-  // FIXME The execution space instance should be passed to the nodes.
-  void submit(const ExecutionSpace& /* exec */) {
+  void submit(const ExecutionSpace& exec) {
     if (!m_has_been_instantiated) instantiate();
     // This reset is gross, but for the purposes of our simple host
     // implementation...
     for (auto& sink : m_sinks) {
       sink->reset_has_executed();
     }
+
+    // We don't know where the nodes will execute, so we need to fence the given
+    // execution space instance before proceeding. This is the simplest way
+    // of guaranteeing that the kernels in the graph are correctly "enqueued".
+    exec.fence(
+        "Kokkos::DefaultGraph::submit: fencing before launching graph nodes");
+
     for (auto& sink : m_sinks) {
-      sink->execute_node();
+      sink->execute_node(exec);
+    }
+
+    // Once all sinks have been executed, we need to fence them.
+    for (const auto& sink : m_sinks) {
+      if (sink->awaitable() && sink->get_execution_space() != exec)
+        sink->get_execution_space().fence(
+            "Kokkos::DefaultGraph::submit: fencing before ending graph submit");
     }
   }
 
--- a/core/unit_test/TestGraph.hpp
+++ b/core/unit_test/TestGraph.hpp
@@ -189,15 +189,6 @@ TEST_F(TEST_CATEGORY_FIXTURE_DEATH(graph), can_instantiate_only_once) {
 // one passed to the Kokkos::Graph constructor.
 TEST_F(TEST_CATEGORY_FIXTURE(graph),
        submit_onto_another_execution_space_instance) {
-// FIXME For ROCm 5.2, the graph implementation is bugged, and Kokkos
-//       therefore uses the default implementation. The kernels of the
-//       graph will thus run on the default stream.
-#if defined(KOKKOS_ENABLE_HIP)
-#if HIP_VERSION_MAJOR == 5 && HIP_VERSION_MINOR == 2
-  if constexpr (std::is_same_v<TEST_EXECSPACE, Kokkos::HIP>)
-    GTEST_SKIP() << "Graph is not properly enqueued.";
-#endif
-#endif
   const auto execution_space_instances =
       Kokkos::Experimental::partition_space(ex, 1, 1);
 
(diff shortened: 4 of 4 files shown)


Rombur:
`HPX` is failing

core/unit_test/TestGraph.hpp
```diff
@@ -349,14 +340,179 @@ TEST_F(TEST_CATEGORY_FIXTURE(graph), zero_work_reduce) {
...
+  // TODO Check that kernels are running on the execution space instance of
+  //      their policy if the defaulted graph implementation is used.
+  graph.submit(ex);
+
+  view_h_t data_host(
+      Kokkos::view_alloc(Kokkos::WithoutInitializing, "diamond - data - host"),
+      4);
+  Kokkos::deep_copy(ex, data_host, data);
+
+  ex.fence();
```
dalg24, line 432: Why not just `auto data_host =  create_mirror_view_and_copy(data);`

core/unit_test/TestGraph.hpp
```diff
@@ -349,14 +340,179 @@ TEST_F(TEST_CATEGORY_FIXTURE(graph), zero_work_reduce) {
...
+
+  ASSERT_EQ(data_host(0), value_A);
```
dalg24, line 434:
```suggestion
  ASSERT_EQ(data_host(index_A), value_A);
```

dalg24 (commented):
Still unsure how I feel about exec spaces being specified for a node parallel construct

core/src/impl/Kokkos_Default_GraphNode_Impl.hpp
```diff
@@ -92,6 +92,18 @@ struct GraphNodeBackendSpecificDetails {
     m_is_aggregate = true;
   }
 
+  // A node is awaitable if it can execute a kernel.
+  // A root node or an aggragate node cannot be waited for, because it does
```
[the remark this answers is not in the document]
romintomasetti: :heavy_check_mark:

core/src/impl/Kokkos_Default_GraphNodeKernel.hpp
```diff
@@ -36,15 +36,22 @@ struct GraphNodeKernelDefaultImpl {
   // TODO @graphs decide if this should use vtable or intrusive erasure via
   //      function pointers like in the rest of the graph interface
   virtual void execute_kernel() = 0;
+
+  GraphNodeKernelDefaultImpl() = default;
+
+  GraphNodeKernelDefaultImpl(ExecutionSpace exec)
```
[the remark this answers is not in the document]
romintomasetti: Indeed!

core/unit_test/TestGraph.hpp
```diff
@@ -349,14 +340,179 @@ TEST_F(TEST_CATEGORY_FIXTURE(graph), zero_work_reduce) {
...
-    Kokkos::fence();
 #endif
   graph.submit();
   Kokkos::deep_copy(ex, count_host, count);
   ex.fence();
   ASSERT_EQ(count_host(), 0);
 }
+
+// Ensure that an empty graph can be submitted.
+TEST_F(TEST_CATEGORY_FIXTURE(graph), empty_graph_can_be_submitted) {
```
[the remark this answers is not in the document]
romintomasetti: OK, done.

core/src/impl/Kokkos_Default_GraphNode_Impl.hpp
```diff
@@ -92,6 +92,18 @@ struct GraphNodeBackendSpecificDetails {
...
+  // not launch anything.
+  bool awaitable() const { return (!m_is_root) && (!m_is_aggregate); }
+
+  // Retrieve the execution space instance that has been passed to
+  // the kernel at construction phase.
+  const ExecutionSpace& get_execution_space() const {
+    KOKKOS_EXPECTS(m_kernel_ptr != nullptr && awaitable())
```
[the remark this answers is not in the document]
romintomasetti: Let me try to summarize this.

Looking at `Kokkos::Impl::GraphNodeImpl`, it inherits from `Kokkos::Impl::GraphNodeBackendSpecificDetails` and `Kokkos::Impl::ExecutionSpaceInstanceStorage`. So it could be tempting to do something like:
```c++
for (const auto& sink : m_sinks) {
  if (sink->awaitable()) {
    const auto & tmp = static_cast<GraphNodeImpl*>(sink.get())->execution_space_instance();
    tmp.fence("fencing before ending graph submit");
  }
}
```
and get rid of the things I've added.

The thing is, it won't do what I want.

It seems that `Kokkos::Impl::GraphNodeImpl::execution_space_instance` contains the execution space instance given to the `create_graph`.

And since our beloved kernel nodes will launch on the execution space instance of their execution policy, which i explicitly made different from the execution space instance passed to the `create_graph`, you see the problem coming.

So I had to sort of "catch" the execution space instance of the kernel execution policy `GraphNodeKernelImpl`, which is only used for kernel nods, not for root nor aggregates.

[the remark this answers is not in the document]
romintomasetti: Now, answering your question, I think `m_kernel_impl` will be `nullptr` if the node is not `awaitable`, *i.e.* it should be `nullptr` for root and aggregates. So probably `KOKKOS_EXPECTS(m_kernel_ptr != nullptr)` would be sufficient.

core/unit_test/TestGraph.hpp
romintomasetti: Good catch !

romintomasetti: Done, closing.

core/src/impl/Kokkos_Default_Graph_Impl.hpp
```diff
@@ -141,16 +141,27 @@ struct GraphImpl : private ExecutionSpaceInstanceStorage<ExecutionSpace> {
...
+  void submit(const ExecutionSpace& exec) {
     if (!m_has_been_instantiated) instantiate();
     // This reset is gross, but for the purposes of our simple host
     // implementation...
     for (auto& sink : m_sinks) {
       sink->reset_has_executed();
     }
+
+    // We don't know where the nodes will execute, so we need to fence the given
+    // execution space instance before proceeding. This is the only readable way
```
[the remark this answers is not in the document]
romintomasetti: I guess I'll replace it with "simplest way".

core/src/impl/Kokkos_Default_GraphNode_Impl.hpp
```diff
@@ -115,8 +127,17 @@ struct GraphNodeBackendSpecificDetails {
...
-        predecessor->execute_node();
+        predecessor->execute_node(exec);
       }
+
+      // Before executing the kernel, be sure to fence the execution space
+      // instance of predecessors.
+      for (const auto& predecessor : m_predecessors) {
+        if (predecessor->awaitable())
+          predecessor->get_execution_space().fence(
+              "fencing before launching child graph node");
```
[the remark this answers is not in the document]
romintomasetti: Done.

core/src/impl/Kokkos_Default_GraphNodeKernel.hpp
```diff
@@ -36,15 +36,22 @@ struct GraphNodeKernelDefaultImpl {
...
 // TODO Indicate that this kernel specialization is only for the Host somehow?
 template <class ExecutionSpace, class PolicyType, class Functor,
           class PatternTag, class... Args>
 class GraphNodeKernelImpl
-    : public PatternImplSpecializationFromTag<PatternTag, Functor, PolicyType,
-                                              Args..., ExecutionSpace>::type,
-      public GraphNodeKernelDefaultImpl<ExecutionSpace> {
+    : public GraphNodeKernelDefaultImpl<ExecutionSpace>,
+      public PatternImplSpecializationFromTag<PatternTag, Functor, PolicyType,
+                                              Args..., ExecutionSpace>::type {
```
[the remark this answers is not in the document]
romintomasetti: Because I'm extracting the execution space instance from the execution policy to pass it to the constructor of `GraphNodeKernelDefaultImpl`. Then, the policy is moved to the constructor of `PatternImplSpecializationFromTag`.

Or maybe I'm wrong.

core/unit_test/TestGraph.hpp
```diff
@@ -349,14 +340,179 @@ TEST_F(TEST_CATEGORY_FIXTURE(graph), zero_work_reduce) {
...
+  constexpr int value_A = 42, value_B = 27, value_C = 13, value_D = 147;
+  constexpr int index_A = 0, index_B = 1, index_C = 2, index_D = 3;
+
+  auto graph = Kokkos::Experimental::create_graph(ex, [&](auto root) {
+    auto node_A = root.then_parallel_for(
+        policy_t(exec_0, 0, 1), AddTo<view_t, index_A, 0>(data, value_A, {}));
+
+    auto node_B = node_A.then_parallel_for(
+        policy_t(exec_0, 0, 1),
+        AddTo<view_t, index_B, 1>(data, value_B, {index_A}));
```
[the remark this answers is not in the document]
romintomasetti: You'll have to show me how to do that.

In the mean time, I've renamed it to `FetchValuesAndContribute`. A bit longer but might help understand what it does.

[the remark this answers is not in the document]
romintomasetti: And CTAD works. Could you have a look and see if this is now OK for you ?

core/unit_test/TestGraph.hpp
romintomasetti: I guess I just copied what was done in other tests of this file.

But also thinking of the test itself, I think it's a good idea to avoid any global fencing (that are implied by `create_mirror_view_and_copy` right ?). Because we want to ensure everything was properly enqueued.

core/src/impl/Kokkos_Default_GraphNode_Impl.hpp
```diff
@@ -92,6 +92,18 @@ struct GraphNodeBackendSpecificDetails {
     m_is_aggregate = true;
   }
 
+  // A node is awaitable if it can execute a kernel.
+  // A root node or an aggregate node cannot be waited for, because it does
+  // not launch anything.
+  bool awaitable() const { return (!m_is_root) && (!m_is_aggregate); }
```
crtrott, line 98: Is this a new publicly visible API?

core/src/impl/Kokkos_Default_Graph_Impl.hpp
```diff
@@ -141,16 +141,29 @@ struct GraphImpl : private ExecutionSpaceInstanceStorage<ExecutionSpace> {
...
     for (auto& sink : m_sinks) {
-      sink->execute_node();
+      sink->execute_node(exec);
+    }
+
+    // Once all sinks have been executed, we need to fence them.
+    for (const auto& sink : m_sinks) {
+      if (sink->awaitable())
+        sink->get_execution_space().fence(
+            "Kokkos::DefaultGraph::submit: fencing before ending graph submit");
```
crtrott, line 166: We only kinda need to do that if they are different from the one passed to the graph submit or?

crtrott (commented):
I think I get it. For all of the fences: if the execution spaces of the nodes (and sinks) are the same as the one passed to the graph itself, does that mean we could not fence ever?

core/src/impl/Kokkos_Default_GraphNode_Impl.hpp
romintomasetti: It's a method of `Kokkos::Impl::GraphNodeBackendSpecificDetails` so I guess no one (that is sufficiently careful) will use it.

However, I guess I can put that `awaitable` method within the `private` methods if you want.

core/src/impl/Kokkos_Default_Graph_Impl.hpp
romintomasetti: You might possibly be right. I need to double-check.

core/src/impl/Kokkos_Default_GraphNodeKernel.hpp
```diff
@@ -61,17 +68,19 @@ class GraphNodeKernelImpl
...
 
   // FIXME @graph Forward through the instance once that works in the backends
   template <class PolicyDeduced, class... ArgsDeduced>
   GraphNodeKernelImpl(ExecutionSpace const &ex, Functor arg_functor,
                       PolicyDeduced &&arg_policy, ArgsDeduced &&...args)
       : GraphNodeKernelImpl("", ex, std::move(arg_functor),
                             (PolicyDeduced &&)arg_policy,
-                            (ArgsDeduced &&)args...) {}
+                            (ArgsDeduced &&)args...) {
+    // FIXME This constructor seem unused.
```
dalg24, line 82: So is it used?

core/src/impl/Kokkos_Default_GraphNode_Impl.hpp
dalg24: It is fine as proposed

core/src/impl/Kokkos_Default_GraphNode_Impl.hpp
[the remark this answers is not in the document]
dalg24: That's what I was thinking

core/unit_test/TestGraph.hpp
```diff
@@ -349,14 +340,207 @@ TEST_F(TEST_CATEGORY_FIXTURE(graph), zero_work_reduce) {
...
   Kokkos::deep_copy(ex, count_host, count);
   ex.fence();
   ASSERT_EQ(count_host(), 0);
 }
+
+// Ensure that an empty graph can be submitted.
+TEST_F(TEST_CATEGORY_FIXTURE(graph), empty_graph) {
+  auto graph = Kokkos::Experimental::create_graph(ex, [](auto) {});
+  graph.instantiate();
+  graph.submit(ex);
```
dalg24, line 354:
```suggestion
  ASSERT_NO_THROW({
    graph.instantiate();
    graph.submit(ex);
  });
```
(not formatted)

core/unit_test/TestGraph.hpp
```diff
@@ -349,14 +340,207 @@ TEST_F(TEST_CATEGORY_FIXTURE(graph), zero_work_reduce) {
...
+                           Kokkos::Array<size_t, NumIndices> indices_,
+                           std::integral_constant<size_t, TargetIndex>,
+                           typename ViewType::value_type value_)
+      : data(std::move(data_)), value(value_), indices(std::move(indices_)) {}
+
+  template <typename T>
+  KOKKOS_FUNCTION void operator()(const T) const {
+    if constexpr (NumIndices > 0) {
+      for (size_t index = 0; index < indices.size(); ++index)
+        data(TargetIndex) += data(indices[index]);
```
dalg24, line 382: I am tempted to tell you to can that `Kokkos::Array` and use an old fashioned C array instead so you can write a range-based for loop instead
```C++
for (auto index : indices) {
  data(TargetIndex) += data(index);
}
```
which I find more elegant in this case.  I suspect that this way we might not even need the fancy deduction guide.

core/unit_test/TestGraph.hpp
```diff
@@ -349,14 +340,207 @@ TEST_F(TEST_CATEGORY_FIXTURE(graph), zero_work_reduce) {
...
+//  \ /         fence(exec_1)               enforce dependencies.
+//   D          D(exec_0)
+TEST_F(TEST_CATEGORY_FIXTURE(graph), diamond) {
+  const auto execution_space_instances =
+      Kokkos::Experimental::partition_space(ex, 1, 1, 1, 1);
+
+  const auto exec_0 = execution_space_instances.at(0);
+  const auto exec_1 = execution_space_instances.at(1);
+  const auto exec_2 = execution_space_instances.at(2);
+  const auto exec_3 = execution_space_instances.at(3);
```
dalg24, line 410: When I see that I remember why I don't like the current interface and would like something that let the structured binding
```C++
auto const [exec_0, exec_1, exec_3] = partition_space(ex, 1, 1, 1);
```
(just a rant, no actionable item)

core/unit_test/TestGraph.hpp
```diff
@@ -349,14 +340,207 @@ TEST_F(TEST_CATEGORY_FIXTURE(graph), zero_work_reduce) {
...
+      Kokkos::view_alloc(Kokkos::WithoutInitializing, "diamond - data - host"),
+      4);
+  Kokkos::deep_copy(exec_3, data_host, data);
+
+  exec_3.fence();
+
+  ASSERT_EQ(data_host(index_A()), value_A);
+  ASSERT_EQ(data_host(index_B()), value_A + value_B);
+  ASSERT_EQ(data_host(index_C()), value_A + value_C);
+  ASSERT_EQ(data_host(index_D()), 2 * value_A + value_B + value_C + value_D);
```
dalg24, line 458:
```suggestion
  ASSERT_EQ(data_host(index_A), value_A);
  ASSERT_EQ(data_host(index_B), value_A + value_B);
  ASSERT_EQ(data_host(index_C), value_A + value_C);
  ASSERT_EQ(data_host(index_D), 2 * value_A + value_B + value_C + value_D);
```

core/unit_test/TestGraph.hpp
dalg24: Sure

core/src/impl/Kokkos_Default_GraphNode_Impl.hpp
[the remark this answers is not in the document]
romintomasetti: OK, I will change to `KOKKOS_EXPECTS(m_kernel_ptr != nullptr)`.

core/src/impl/Kokkos_Default_Graph_Impl.hpp
romintomasetti: Indeed, you're right.

I thought about it with 2 though-experiments.

# Case 1 - only sinks

There are only sinks (*i.e.* they don't have predecessors). Then, it is true that we need to fence only the sinks that where not executing on the passed `exec`.

# Case 2 - complex DAG

The sinks should already have fenced their predecessors, so we come back to the situation of case 1.


Conclusion: you must be right. I'll do it.

core/unit_test/TestGraph.hpp
romintomasetti: In the past, I also expected that the `partition_space` should return a tuple-like container if weights are passed one-by-one, and a vector-like container if weights are passed with a vector.

I can open an issue for that, to gather people's feedback. I do think I would make sense.

core/unit_test/TestGraph.hpp
romintomasetti: I know it's ugly... I was disappointed too.

But your proposal does not compile. Under the `clang`+`HPX` configuration used in the CI, you'd get:
```bash
In file included from /workspaces/kokkos/core/src/Kokkos_View.hpp:25:
/workspaces/kokkos/core/src/View/Kokkos_ViewLegacy.hpp:493:5: error: static_assert failed due to requirement 'Kokkos::Impl::are_integral<std::integral_constant<unsigned long, 5>>::value'
    static_assert(Kokkos::Impl::are_integral<Is...>::value);
    ^             ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/workspaces/kokkos/core/src/View/Kokkos_ViewLegacy.hpp:506:5: note: in instantiation of function template specialization 'Kokkos::View<int *, Kokkos::HostSpace>::check_operator_parens_valid_args<std::integral_constant<unsigned long, 5>>' requested here
    check_operator_parens_valid_args(i0);
    ^
/workspaces/kokkos/core/unit_test/TestGraph.hpp:544:22: note: in instantiation of function template specialization 'Kokkos::View<int *, Kokkos::HostSpace>::operator()<std::integral_constant<unsigned long, 5>>' requested here
  ASSERT_EQ(data_host(index_F),
                     ^
1 error generated.
```

dalg24: Ugggg
cc @crtrott

romintomasetti: I guess you'd need some "is convertible to" instead of "is integral".

core/unit_test/TestGraph.hpp
romintomasetti: Surprise surprise... I cannot, because if `NumIndices` is zero, you get:
```bash
/workspaces/kokkos/core/unit_test/TestGraph.hpp:367:18: error: zero size arrays are an extension [-Werror,-Wzero-length-array]
  size_t indices[NumIndices];
```

core/unit_test/TestGraph.hpp
romintomasetti: We will need `C++20` for structured bindings to be usable easily. For now you'd get something like (with `clang`):
```bash
error: reference to local binding 'a' declared in enclosing function 'f' 
```

So once `Kokkos` is on `c++20` we'll be able to do:
```c++
const auto [exec_0, ... exec_N] = Kokkos::Experimental::partition_space(ex, weight_0, ..., weight_N);
```

romintomasetti:
The error with `gcc 8.4` is that it cannot use the deduction guide if we don't explicitly convert the integral constants to their value in the initializer list.

core/src/impl/Kokkos_Default_GraphNode_Impl.hpp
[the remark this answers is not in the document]
romintomasetti: @dalg24 OK to close this thread ?

core/src/impl/Kokkos_Default_Graph_Impl.hpp
romintomasetti: @crtrott Done, feel free to close the thread.

core/src/impl/Kokkos_Default_GraphNodeKernel.hpp
romintomasetti: @dalg24 OK to close this thread if:
- I don't delete the constructor because we don't know if someone downstream would be impacted.
- Keep the comment because this is still something worth investigating in the future.

core/unit_test/TestGraph.hpp
romintomasetti: @dalg24 Do we wait for #7293 or not ?

core/unit_test/TestGraph.hpp
romintomasetti: @dalg24 OK to close ?

core/unit_test/TestGraph.hpp
romintomasetti: @dalg24 @crtrott Do we just ignore this, and expect the next `View` implementation to allow for this ?

core/src/impl/Kokkos_Default_GraphNodeKernel.hpp
masterleinad: How could this constructor be used given that its name contains "Impl"?

romintomasetti: I don't mind deleting this constructor if you guys think it's OK, but I would refrain doing it here because this PR's purpose is not really to cleanup the mess :wink:

core/unit_test/TestGraph.hpp
dalg24: I would say add a note as a comment would not be a bad idea but not blocking

core/unit_test/TestGraph.hpp
dalg24: Not necessarily no

romintomasetti: OK, since I left a `FIXME`-like comment, I'll close this thread.

core/unit_test/TestGraph.hpp
romintomasetti: Should I had the comment on every line for which it would be nice to write `data(index_...)` instead of `data(index_...())` ? (*i.e.* how to make the comment visible, not repeated a 100x times)

dalg24 (approved):

core/src/impl/Kokkos_Default_GraphNodeKernel.hpp
```diff
@@ -36,15 +36,22 @@ struct GraphNodeKernelDefaultImpl {
   // TODO @graphs decide if this should use vtable or intrusive erasure via
   //      function pointers like in the rest of the graph interface
   virtual void execute_kernel() = 0;
+
+  GraphNodeKernelDefaultImpl() = default;
+
+  explicit GraphNodeKernelDefaultImpl(ExecutionSpace exec)
+      : m_execution_space(std::move(exec)) {}
```
masterleinad, line 43: Any reason not to take the execution space instance by const reference as we usually do?

core/src/impl/Kokkos_Default_GraphNode_Impl.hpp
```diff
@@ -115,8 +127,17 @@ struct GraphNodeBackendSpecificDetails {
...
+        predecessor->execute_node(exec);
       }
+
+      // Before executing the kernel, be sure to fence the execution space
+      // instance of predecessors.
+      for (const auto& predecessor : m_predecessors) {
+        if (predecessor->awaitable())
+          predecessor->get_execution_space().fence(
+              "Kokkos::DefaultGraphNode::execute_node: sync with predecessors");
+      }
```
masterleinad, line 139: This is only necessary if the execution space instance the node was created with doesn't match the execution space instance used here, right?

masterleinad: This won't be sufficient for thread-safety.

core/src/impl/Kokkos_Default_Graph_Impl.hpp
```diff
@@ -141,16 +141,29 @@ struct GraphImpl : private ExecutionSpaceInstanceStorage<ExecutionSpace> {
...
     // implementation...
     for (auto& sink : m_sinks) {
       sink->reset_has_executed();
     }
+
+    // We don't know where the nodes will execute, so we need to fence the given
+    // execution space instance before proceeding. This is the simplest way
+    // of guaranteeing that the kernels in the graph are correctly "enqueued".
+    exec.fence(
+        "Kokkos::DefaultGraph::submit: fencing before launching graph nodes");
```
masterleinad, line 156: This won't be sufficient for thread-safety.

core/src/impl/Kokkos_Default_GraphNodeKernel.hpp
romintomasetti: Noting that:
1.  `Kokkos` execution space classes are behaving as shared pointers to an internal resource,
2.  `GraphNodeKernelDefaultImpl` needs to explicitly extend the lifetime of the underlying internal execution space instance,

I'm using http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#r34-take-a-shared_ptrwidget-parameter-to-express-shared-ownership.

Clearly not the standard way of doing things in `Kokkos` for sure, but definitely worth a thought.

core/src/impl/Kokkos_Default_GraphNode_Impl.hpp
romintomasetti: A node cannot execute before all its predecessors are done.

The defaulted implementation can be thought of as if each node stores an execution space instance, potentially different from the one passed to `graph.submit(...)`.

So, before a node can actually launch its kernel, it has to fence its predecessors' execution space instances to ensure they are done.

We could save a fence or two with:
```suggestion
      // Before executing the kernel, be sure to fence the execution space
      // instance of predecessors.
      for (const auto& predecessor : m_predecessors) {
        if (predecessor->awaitable() && predecessor->get_execution_space() != this->get_execution_space())
          predecessor->get_execution_space().fence(
              "Kokkos::DefaultGraphNode::execute_node: sync with predecessors");
      }
```
if you want.

core/src/impl/Kokkos_Default_Graph_Impl.hpp
romintomasetti: Could you explain what use case you have in mind ?

Here is mine.

When a user calls `graph.submit(exec_A)`, the expectation is that any node of the graph might start once all the work on `exec_A` is consumed.

masterleinad: The thread-safety guarantee we have been working towards is that kernels committed to the same execution space instance don't run concurrently. So `graph.execute(exec_space)` should make sure that there are no kernels running concurrently to the graph execution on the same execution space instance.

I'm not saying we need to fix that here. We have to define thread-safety a little better in the documentation anyway.

core/src/impl/Kokkos_Default_GraphNode_Impl.hpp
masterleinad: Yes, I think that makes sense, in particular since you have already done something similar for https://github.com/kokkos/kokkos/pull/7271/files#r1742704039. I would assume the common case is for all nodes to execute on the same execution space instance and so there wouldn't be any fences.

core/src/impl/Kokkos_Default_GraphNodeKernel.hpp
masterleinad: > Kokkos execution space classes are behaving as shared pointers to an internal resource

That's an internal detail. I'm not sure if that's what we want to tell people and teach them to do.
I won't block on this question.

core/src/impl/Kokkos_Default_GraphNode_Impl.hpp
romintomasetti: Done :heavy_check_mark:

masterleinad (approved):
Fine with me.

core/src/impl/Kokkos_Default_Graph_Impl.hpp
romintomasetti: I guess that guaranteeing that there will never be 2 nodes that hold the same execution space instance that run concurrently would:
- either increase the complexity of the defaulted graph implementation (you would probably need to scan the graph and make smart routines here and there)
- or ensure that we run nodes one by one (sequentially), killing any chance that the defaulted graph would run anything asynchronously

So yes, I'd suggest we keep things as is, but keep your remark in mind :smile:

masterleinad:
You'll have to fix the indentation.

romintomasetti:
I just added one commit to use the range-base for loop from #7293.

dalg24:
Failures are unrelated and fixed in #7299

PR #7272, opened 2024-08-29
Ignore warnings for destructor in ViewOfViews test

masterleinad:
This fixes the failing `CUDA-11.6-NVCC-DEBUG'` CI.

### DIFF
--- a/core/unit_test/CMakeLists.txt
+++ b/core/unit_test/CMakeLists.txt
@@ -300,6 +300,12 @@ foreach(Tag Threads;Serial;OpenMP;Cuda;HPX;OpenMPTarget;OpenACC;HIP;SYCL)
       list(APPEND ${Tag}_SOURCES2A ${file})
     endforeach()
 
+    # disable a warning about calling a host-only destructor in the ViewOfView test
+    # if we use nvcc_wrapper directly.
+    if(KOKKOS_CXX_COMPILER_ID STREQUAL NVIDIA AND NOT INTERNAL_USE_COMPILER_LAUNCHER)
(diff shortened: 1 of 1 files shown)


masterleinad:
Retest this please.

masterleinad:
Retest this please.

core/unit_test/CMakeLists.txt
```diff
@@ -300,6 +300,12 @@ foreach(Tag Threads;Serial;OpenMP;Cuda;HPX;OpenMPTarget;OpenACC;HIP;SYCL)
       list(APPEND ${Tag}_SOURCES2A ${file})
     endforeach()
 
+    # disable a warning about calling a host-only destructor in the ViewOfView test
+    # if we use nvcc_wrapper directly.
+    if(KOKKOS_CXX_COMPILER_ID STREQUAL NVIDIA AND NOT INTERNAL_USE_COMPILER_LAUNCHER)
+      set_source_files_properties(${dir}/Test${Tag}_ViewOfViews.cpp PROPERTIES COMPILE_FLAGS "-Xcudafe --diag_suppress=20014")
```
dalg24, line 306: I would prefer a pragma push restricted to the NVCC versions that need it

masterleinad:
> Do you understand why this is only showing up in that one CUDA 11.6 debug build?
> It looks like it is being emitted from that `ViewValueFunctor::functor_instantiate_workaround()` function.

No, but it seems reasonable for `nvcc` to complain about calling a host-only destructor from a functor (`__host__ __device__` function).

masterleinad:
We found a better workaround with @dalg24 and @crtrott.

PR #7275, opened 2024-08-29
Fix warning for host only destructor of View element on CUDA

crtrott:
```  
// When instantiating a View on host execution space with a host only
// destructor the workaround for CUDA device symbol instantiation tries to
// still compile a destruction kernel for the device, and issues a warning
// for host from host-device
```


masterleinad (approved):
Fine with me.

tcclevenger (approved):

PR #7276, opened 2024-08-29
Fix View creation in TeamPolicy array reduction test

masterleinad:
`result_type` is a 1D View so we need to provide an extent for the first dimension.

### DIFF
--- a/core/unit_test/TestTeam.hpp
+++ b/core/unit_test/TestTeam.hpp
@@ -401,7 +401,7 @@ class TestReduceTeam {
       using result_type = Kokkos::View<ScalarType *, Kokkos::HostSpace,
                                        Kokkos::MemoryUnmanaged>;
 
-      ScalarType result[Repeat][3];
+      ScalarType result[Repeat][Count];
 
       const unsigned team_size = team_exec.team_size_recommended(
           functor_type(nwork), Kokkos::ParallelReduceTag());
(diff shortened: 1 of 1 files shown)


tcclevenger (approved):

core/unit_test/TestTeam.hpp
```diff
@@ -410,7 +410,7 @@ class TestReduceTeam {
       team_exec = policy_type(league_size, team_size);
 
       for (unsigned i = 0; i < Repeat; ++i) {
-        result_type tmp(&result[i][0]);
+        result_type tmp(&result[i][0], 3);
```
tcclevenger, line 413: Should we create a `constexpr int size=3` and use it in `ScalarType result[Repeat][size]` and `result_type tmp(&result[i][0], size)`?

Probably isn't too important because it is a unit test, but could avoid confusion if the test is ever updated. Feel free to ignore.

masterleinad:
Retest this please.

aprokop (approved):
I am a bit confused. How did it work before?

masterleinad:
> I am a bit confused. How did it work before?

It was still doing accessing the memory correctly. It's just interesting that we have only one build detecting it (which as broken for other reasons at the same time).

dalg24:
> How did it work before?

Because we don't diagnose that invalid number of arguments at compile-time and that the runtime check only happens when bounds checking is enabled...

https://github.com/kokkos/kokkos/blob/662043966274936f2c7a367e58975f06c2814cf7/core/src/View/Kokkos_ViewLegacy.hpp#L1071-L1091

PR #7277, opened 2024-08-30
docker(hpx): provide `dockerfile` and use it in the ci action

romintomasetti:


crtrott:
Outdated: HPX only runs on SNL testers now.

PR #7279, opened 2024-09-01
Cleanup atomics test

dalg24:
In anticipation of upcoming changes where we avoid using the `typeid` operator, we revisit the test failure reporting in atomic tests and prefer native gtest mechanism to log additional information on failure.

### DIFF
--- a/core/unit_test/TestAtomicViews.hpp
+++ b/core/unit_test/TestAtomicViews.hpp
@@ -16,7 +16,7 @@
 
 #include <Kokkos_Core.hpp>
 
-namespace TestAtomicViews {
+namespace {
 
 //-------------------------------------------------
 //-----------atomic view api tests-----------------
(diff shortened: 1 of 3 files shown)


core/unit_test/TestAtomicViews.hpp
```diff
@@ -1044,17 +993,9 @@ bool LSEqualAtomicViewTest(const int64_t input_length) {
   T res = LSEqualAtomicView<T, DeviceType>(input_length, value, remainder);
   T resSerial = LSEqualAtomicViewCheck<T>(input_length, value, remainder);
 
-  bool passed = true;
-
-  if (resSerial != res) {
-    passed = false;
-
-    std::cout << "Loop<" << typeid(T).name()
-              << ">( test = RSEqualAtomicViewTest"
```
dalg24, line 1053: Note the bug in the code being removed

crtrott: so we need iostream because gtest likes that?

crtrott (approved):
I dislike the iostream usage in TestAtomicOperations (makes it harder in my mind to read what the expected output is), but if you all like it that's fine. Would fprintf with error stream work anyway?

core/unit_test/TestAtomicViews.hpp
dalg24: But I agree the rational for that change is much weaker than the other ones.

core/unit_test/TestAtomics.hpp
```diff
@@ -440,106 +441,80 @@ T LoopVariantSerial(int loop, int test) {
...
-
 TEST(TEST_CATEGORY, atomics) {
   const int loop_count = 1e4;
 
-  ASSERT_TRUE((TestAtomic::Loop<int, TEST_EXECSPACE>(loop_count, 1)));
-  ASSERT_TRUE((TestAtomic::Loop<int, TEST_EXECSPACE>(loop_count, 2)));
-  ASSERT_TRUE((TestAtomic::Loop<int, TEST_EXECSPACE>(loop_count, 3)));
+  Loop<int, TEST_EXECSPACE>(loop_count, 1);
+  Loop<int, TEST_EXECSPACE>(loop_count, 2);
+  Loop<int, TEST_EXECSPACE>(loop_count, 3);
```
cedricchevalier19, line 462: Perhaps using an enum instead of `1`, `2`, and `3` can make things a little bit clearer.

cedricchevalier19 (approved):
Thanks for clarifying the switch to iostream.

core/unit_test/TestAtomics.hpp
dalg24: Leaving it as a potential future refactor work

dalg24:
Retest this please

dalg24:
Ignoring NVC++ segfaulting in the OpenACC build.  Everything else passed.

PR #7284, opened 2024-09-02
Reduce visibility of `g_openmp_hardware_max_threads` symbol to avoid trouble with shared libabries

### LINKED ISSUE #7283: Building with pybind11 can't properly intilize Kokkos::Impl::g_openmp_hardware_max_threads
**Describe the bug**

When building Python module using pybind11 and linking to Kokkos with OpenMP backend, the variable `Kokkos::Impl::g_openmp_hardware_max_threads` is always `1`, which breaks `ScatterView`.

- Platform: Ubuntu 22.04 on Intel 9700k
- Compiler: clang++ 17.0.6
- Kokkos version: 4.4.0 and commit a068d826dd79e8994c7b43a3b72577e0ea704b8f
- pybind11 version: 2.12.0dev1
- CMake configure command: `CXX=clang++ cmake .. -DKokkos_ENABLE_OPENMP=ON -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=ON` (`BUILD_SHARED_LIBS=OFF` results in linkage error)

C++ file (`pytest.cpp`)
```cpp
#include <Kokkos_Core.hpp>
#include <pybind11/pybind11.h>
#include <omp.h>

void initialize() {
    printf("Value of g_openmp_hardware_max_threads before initialize is %d\n", Kokkos::Impl::g_openmp_hardware_max_threads);
    Kokkos::initialize();
    printf("Value of g_openmp_hardware_max_threads after initialize is %d, while omp_current_max_threads is %d\n",
        Kokkos::Impl::g_openmp_hardware_max_threads,
        omp_get_max_threads()
    );
}

void finalize() {
    printf("Finalizing\n");
    Kokkos::finalize();
}

PYBIND11_MODULE(pytest, m) {
    m.def("initialize", &initialize);
    m.def("finalize", &finalize);
}
```
CMakeLists.txt:
```cmake
cmake_minimum_required (VERSION 3.10)
project(pytest)

add_subdirectory(kokkos)
find_package(pybind11 REQUIRED)

pybind11_add_module(pytest pytest.cpp)
target_link_libraries(pytest PUBLIC Kokkos::kokkos)
```

test.py
```python
import pytest

pytest.initialize()
pytest.finalize()
```

Output of `python3 test.py`:
```
Value of g_openmp_hardware_max_threads before initialize is 1
Value of g_openmp_hardware_max_threads after initialize is 1, while omp_current_max_threads is 8
Finalizing
```

If add a print line in this function: https://github.com/kokkos/kokkos/blob/08ceff92bcf3a828844480bc1e6137eb74028517/core/src/OpenMP/Kokkos_OpenMP_Instance.cpp#L161
the printed value of `Kokkos::Impl::g_openmp_hardware_max_threads` is correct (equal to `omp_get_max_threads`).

Also, changing the declaration of this variable from `inline` to `extern` in the header file and defining it in `core/src/OpenMP/Kokkos_OpenMP_Instance.cpp` fix the issue.

dalg24:
Fix #7283 (not tested)

### DIFF
--- a/core/src/OpenMP/Kokkos_OpenMP_Instance.cpp
+++ b/core/src/OpenMP/Kokkos_OpenMP_Instance.cpp
@@ -31,11 +31,19 @@
 #include <sstream>
 #include <thread>
 
+namespace {
+int g_openmp_hardware_max_threads = 1;
+}
+
 namespace Kokkos {
 namespace Impl {
 
 std::vector<OpenMPInternal *> OpenMPInternal::all_instances;
 std::mutex OpenMPInternal::all_instances_mutex;
+
+int OpenMPInternal::max_hardware_threads() noexcept {
+  return g_openmp_hardware_max_threads;
+}
 
 void OpenMPInternal::clear_thread_data() {
   const size_t member_bytes =
@@ -188,9 +196,9 @@ void OpenMPInternal::initialize(int thread_count) {
     // Before any other call to OMP query the maximum number of threads
     // and save the value for re-initialization unit testing.
 
-    Impl::g_openmp_hardware_max_threads = get_current_max_threads();
+    g_openmp_hardware_max_threads = get_current_max_threads();
 
-    int process_num_threads = Impl::g_openmp_hardware_max_threads;
+    int process_num_threads = g_openmp_hardware_max_threads;
 
     if (Kokkos::hwloc::available()) {
       process_num_threads = Kokkos::hwloc::get_available_numa_count() *
@@ -203,11 +211,11 @@ void OpenMPInternal::initialize(int thread_count) {
     // process_num_threads if thread_count  > 0, set
     // g_openmp_hardware_max_threads to thread_count
     if (thread_count < 0) {
-      thread_count = Impl::g_openmp_hardware_max_threads;
+      thread_count = g_openmp_hardware_max_threads;
     } else if (thread_count == 0) {
-      if (Impl::g_openmp_hardware_max_threads != process_num_threads) {
-        Impl::g_openmp_hardware_max_threads = process_num_threads;
-        omp_set_num_threads(Impl::g_openmp_hardware_max_threads);
+      if (g_openmp_hardware_max_threads != process_num_threads) {
+        g_openmp_hardware_max_threads = process_num_threads;
+        omp_set_num_threads(g_openmp_hardware_max_threads);
       }
     } else {
       if (Kokkos::show_warnings() && thread_count > process_num_threads) {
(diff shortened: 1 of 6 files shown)


core/src/OpenMP/Kokkos_OpenMP_UniqueToken.hpp
```diff
@@ -105,7 +105,8 @@ class UniqueToken<OpenMP, UniqueTokenScope::Global> {
   /// \brief upper bound for acquired values, i.e. 0 <= value < size()
   KOKKOS_INLINE_FUNCTION
   int size() const noexcept {
-    KOKKOS_IF_ON_HOST((return Kokkos::Impl::g_openmp_hardware_max_threads;))
+    KOKKOS_IF_ON_HOST(
+        (return Kokkos::Impl::OpenMPInternal::max_hardware_threads();))
```
dalg24, line 109: Note that this now would involve a function call.
It looks like this is fine as far as `ScatterView` is concerned (not used in the hot path as far as I can tell) but we probably should discuss it a bit.

crtrott (approved):
Fine with me assuming it fixes the issue. My gut feeling is that this should not be a hot path issue with UniqueToken.

core/src/OpenMP/Kokkos_OpenMP_Instance.cpp
```diff
@@ -188,9 +196,9 @@ void OpenMPInternal::initialize(int thread_count) {
     // Before any other call to OMP query the maximum number of threads
     // and save the value for re-initialization unit testing.
 
-    Impl::g_openmp_hardware_max_threads = get_current_max_threads();
+    g_openmp_hardware_max_threads = get_current_max_threads();
 
-    int process_num_threads = Impl::g_openmp_hardware_max_threads;
+    int process_num_threads = g_openmp_hardware_max_threads;
```
masterleinad, line 201: Wouldn't it make sense to only set this for the default execution space instance?

dalg24: I think so.  But such a refactor/fix is beyond the scope of this PR.

masterleinad (approved):

masterleinad:
I verified that we see the test failure with OpenMP when building with shared libraries (and `-fvisibility=hidden`) in user code.

masterleinad:
Retest this please.

masterleinad:
Retest this please

dalg24:
Unrelated failure in one of the CUDA build (GPU not detected at runtime)

PR #7286, opened 2024-09-03
View-Refactor: New View implementation using mdspan details

crtrott:
Hoping this will skip build, because it doesn't work with C++17.

What this does is replacing Kokkos::View implementation (in particular ViewMapping) with implementation details from mdspan. Basic structure is:

View -> BasicView { data_handle, mdspan_mapping, accessor }

This also introduces a bunch of new accessors to make this all work.
Reference counting is handled by a `ReferenceCountedAccessorDataHandle` from a `ReferenceCountedAccessor` (though the name of the latter is misleading, only the data handle does reference counting, not the accessor).

The idea of this draft is to present an overview of what's going on, before issuing smaller PRs which will also work with C++17. 

This draft can also help with early feedback.


crtrott:
Note: this PR does NOT delete the old implementation at all. The files which basically can go amount to ~5k lines so when we can get rid of them this is an almost 4k lines net loss.

PR #7287, opened 2024-09-03
core(hpx): overload for `partition_space`

romintomasetti:
This PR adds an overload of `Kokkos::Experimental::partition_space` for `HPX`.

It creates new instances in "independent" mode.

### DIFF
--- a/core/src/HPX/Kokkos_HPX.hpp
+++ b/core/src/HPX/Kokkos_HPX.hpp
@@ -447,6 +447,20 @@ class HPX {
   }
 };
 
+template <typename... Args>
+std::vector<HPX> partition_space(HPX const &, Args... args) {
+  std::vector<HPX> instances(sizeof...(args));
+  for (auto &in : instances) in = HPX(HPX::instance_mode::independent);
+  return instances;
+}
+
+template <typename T>
+std::vector<HPX> partition_space(HPX const &, std::vector<T> const &weights) {
+  std::vector<HPX> instances(weights.size());
+  for (auto &in : instances) in = HPX(HPX::instance_mode::independent);
+  return instances;
+}
+
 extern template void HPX::impl_bulk_plain_erased<int>(
     bool, bool, std::function<void(int)> &&, int const,
     hpx::threads::thread_stacksize stacksize) const;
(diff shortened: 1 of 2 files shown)


romintomasetti:
@crtrott I'm not sure. I just saw that there was no specialization for `HPX`, and thought of it as if creating new "independant" streams. But I admit I'm no expert in `HPX` and don't know what's going on under the hood.

My only goal is to ensure that calling `partition_space` on `HPX` effectively creates "distinguishable" instances to ensure that tests exploiting concurrency are indeed submitting their workloads on different instances.

@msimberg ?

crtrott (approved):
I will approve anyway: this is an advance and its not a violating semantics to ignore the weights even if we could take them into account: @msimberg if there is a way to honor the weights let us know.

msimberg:
> Do all instances share the same underlying thread pool?

Yeah, they do (otherwise you'd get terrible oversubscription).

> Is there a way to split the existing thread pool, i.e. make the new instance only use some of the workers?

Yeah, there are ways of doing that, but if I remember correctly only with the old-style executors (with `post`, `async_execute, etc.), not with the P2300 scheduler that is used in Kokkos now. @hkaiser maybe you can confirm if this is still the case?

In any case, I'd recommend doing what's done here now, i.e. not try to actually partition the worker threads. It's most of the time a better idea in terms of load balancing. The main case where I can think of it being useful to use only a subset of worker threads is if you know for sure the workload won't scale to all worker threads and you're just wasting resources by using them all. You can to some extent achieve this by increasing the chunk size as well but that's of course a workaround only.

Do you know where the OpenMP/Threads space typically benefits from being partitioned (I assume the CPU ones are the only ones that really benefit...)? Might help to judge if it's something worth adding for the HPX space as well.

msimberg (approved):

dalg24:
Failures are unrelated (machine issues on builds that do not have HPX)

> Do you know where the OpenMP/Threads space typically benefits from being partitioned (I assume the CPU ones are the only ones that really benefit...)? Might help to judge if it's something worth adding for the HPX space as well.

I don't but maybe others can answer.  Merging in the meantime.

ndellingwood:
Does this merit a 4.5 changelog entry?

romintomasetti:
@ndellingwood I am not sure since this is a change to an experimental stuff. But if you deem it necessary I'll do it.

masterleinad:
I already added it.

PR #7288, opened 2024-09-03
View-Refactor: mdspan implementation updates

crtrott:
This is a piece of #7286 fixing a few small mdspan things. I will open a PR on the mdspan implementation shortly also attempting to write a test for one of the conversion issues exposed and fixed here. (`padded_stride_type::init_padding` takes an `extents_type` as argument, which previously would have required `other_mapping.extents()` to be convertible to `extents_type`. The function constraint is only "is_constructible" though, so we need to explicitly cast before calling the helper. 

MDSpan PR here: https://github.com/kokkos/mdspan/pull/358

### DIFF
--- a/tpls/mdspan/include/experimental/__p2630_bits/submdspan_extents.hpp
+++ b/tpls/mdspan/include/experimental/__p2630_bits/submdspan_extents.hpp
@@ -17,6 +17,7 @@
 #pragma once
 
 #include <tuple>
+#include <complex>
 
 #include "strided_slice.hpp"
 namespace MDSPAN_IMPL_STANDARD_NAMESPACE {
@@ -52,6 +53,33 @@ template <class OffsetType, class ExtentType, class StrideType>
 struct is_strided_slice<
     strided_slice<OffsetType, ExtentType, StrideType>> : std::true_type {};
 
+// Helper for identifying valid pair like things
+template <class T, class IndexType> struct index_pair_like : std::false_type {};
+
+template <class IdxT1, class IdxT2, class IndexType>
+struct index_pair_like<std::pair<IdxT1, IdxT2>, IndexType> {
+  static constexpr bool value = std::is_convertible_v<IdxT1, IndexType> &&
+                                std::is_convertible_v<IdxT2, IndexType>;
+};
+
+template <class IdxT1, class IdxT2, class IndexType>
+struct index_pair_like<std::tuple<IdxT1, IdxT2>, IndexType> {
+  static constexpr bool value = std::is_convertible_v<IdxT1, IndexType> &&
+                                std::is_convertible_v<IdxT2, IndexType>;
+};
+
+template <class IdxT, class IndexType>
+struct index_pair_like<std::complex<IdxT>, IndexType> {
+  static constexpr bool value = std::is_convertible_v<IdxT, IndexType>;
+};
+
+template <class IdxT, class IndexType>
+struct index_pair_like<std::array<IdxT, 2>, IndexType> {
+  static constexpr bool value = std::is_convertible_v<IdxT, IndexType>;
+};
+
+// FIXME: we actually need to pass IndexType into all of these
+
 // first_of(slice): getting begin of slice specifier range
 MDSPAN_TEMPLATE_REQUIRES(
   class Integral,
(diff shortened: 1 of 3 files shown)


tpls/mdspan/include/experimental/__p2630_bits/submdspan_extents.hpp
```diff
@@ -17,6 +17,7 @@
 #pragma once
 
 #include <tuple>
+#include <complex>
```
romintomasetti, line 20: Are you sure you want/need to include `<complex>` here ? It's surprising.

tpls/mdspan/include/experimental/__p2642_bits/layout_padded.hpp
```diff
@@ -95,6 +95,7 @@ struct padded_extent {
   using static_array_type = typename static_array_type_for_padded_extent<
       padding_value, _Extents, _ExtentToPadIdx, _Extents::rank()>::type;
 
+  MDSPAN_INLINE_FUNCTION
```
romintomasetti, line 98: `inline` is implicitly implied I think. Would be good to remove that useless specifier if so.

tpls/mdspan/include/experimental/__p2630_bits/submdspan_extents.hpp
masterleinad: See https://github.com/kokkos/mdspan/pull/358/files#r1742703222.

tpls/mdspan/include/experimental/__p2642_bits/layout_padded.hpp
masterleinad: See https://github.com/kokkos/mdspan/pull/358/files#r1742863263.

romintomasetti: My 2 cents is that doing things "for code style consistency" is a bad practice IMO. In this case, stripping out `INLINE` makes the code shorter and more to the point, and remove a clearly useless specifier.

But do whatever you want :smile:

tpls/mdspan/include/experimental/__p2630_bits/submdspan_extents.hpp
romintomasetti: But isn't this pair-like thingy used for slicing ? If so, I would be surprise that slicing is allowed with a complex... I can't image how bad a code would be if complex numbers were used for that.

crtrott: It is by the standard.

crtrott: turns out that the ISO C++ standard defines the exposition only concept pair-like which is defined through tuple-like which is defined here: https://eel.is/c++draft/tuple.like#concept:tuple-like

So `complex<double>` is a valid slice specifier for `submdspan` because its pair-like and double is convertible to `index_type`

romintomasetti: It is a draft right ? If so I would refrain supporting such a ill-formed case.

crtrott: no this is final. This was discussed in the committee and it is basically not avoidable. We want "convertible to index_type" to be valid, that means floating points will be valid. And the committee did not feel like adding an explicit "convertible to index_type but not floating point" to the specification of all these things in mdspan.

romintomasetti: OK, then I wouldn't go against the committee.

But in this specific case, you're still including `complex` in a very central file, so any downstream project using `Kokkos` would have `complex` included ?!

crtrott: That is already the case, because you get it indirectly through Kokkos_Complex

dalg24 (commented):
Please confirm that this is up to date with what we merged in kokkos/mdspan

tpls/mdspan/include/experimental/__p2630_bits/submdspan_extents.hpp
romintomasetti: I still think that `Kokkos` should not pull `<xomplex>` for me, but do what you think is best. I won't block for this of course :wink:

masterleinad (approved):

crtrott:
> Please confirm that this is up to date with what we merged in kokkos/mdspan

Obviously I intended it to be and it is, but isn't checking this part of the review ;-)

dalg24 (approved):

dalg24:
(could not think of a better way to trigger the CI and it did not work...)

dalg24:
Obviously there is a CI SKIP directive...

ndellingwood:
Should the mdspan interop updates get a changelog entry/section?

PR #7293, opened 2024-09-04
Add range-based for loop support for `Array<T, N>`

dalg24:
This is enabled by introducing `begin()` and `end()` free functions that take const and non-const  `Array` and return pointers to elements.

Whether or not to introduce them as member function can be decided elsewhere.
We would still need the free functions as the `std::` functions would lack the `__host__ __device__` annotations.

### DIFF
--- a/core/src/Kokkos_Array.hpp
+++ b/core/src/Kokkos_Array.hpp
@@ -439,7 +439,7 @@ KOKKOS_FUNCTION constexpr T const&& get(Array<T, N> const&& a) noexcept {
 namespace Kokkos {
 
 template <class T, std::size_t N>
-KOKKOS_FUNCTION constexpr T const* begin(Array<T, N> const& a) {
+KOKKOS_FUNCTION constexpr T const* begin(Array<T, N> const& a) noexcept {
   if constexpr (N == 0)
     return nullptr;
   else
@@ -447,7 +447,7 @@ KOKKOS_FUNCTION constexpr T const* begin(Array<T, N> const& a) {
 }
 
 template <class T, std::size_t N>
-KOKKOS_FUNCTION constexpr T* begin(Array<T, N>& a) {
+KOKKOS_FUNCTION constexpr T* begin(Array<T, N>& a) noexcept {
   if constexpr (N == 0)
     return nullptr;
   else
@@ -455,7 +455,7 @@ KOKKOS_FUNCTION constexpr T* begin(Array<T, N>& a) {
 }
 
 template <class T, std::size_t N>
-KOKKOS_FUNCTION constexpr T const* end(Array<T, N> const& a) {
+KOKKOS_FUNCTION constexpr T const* end(Array<T, N> const& a) noexcept {
   if constexpr (N == 0)
     return nullptr;
   else
@@ -463,7 +463,7 @@ KOKKOS_FUNCTION constexpr T const* end(Array<T, N> const& a) {
 }
 
 template <class T, std::size_t N>
-KOKKOS_FUNCTION constexpr T* end(Array<T, N>& a) {
+KOKKOS_FUNCTION constexpr T* end(Array<T, N>& a) noexcept {
   if constexpr (N == 0)
     return nullptr;
   else
--- a/core/src/Kokkos_Array.hpp
+++ b/core/src/Kokkos_Array.hpp
@@ -195,8 +195,10 @@ struct Array<T, 0> {
     return *reinterpret_cast<const_pointer>(-1);
   }
 
-  KOKKOS_INLINE_FUNCTION pointer data() { return nullptr; }
-  KOKKOS_INLINE_FUNCTION const_pointer data() const { return nullptr; }
+  KOKKOS_INLINE_FUNCTION constexpr pointer data() { return nullptr; }
+  KOKKOS_INLINE_FUNCTION constexpr const_pointer data() const {
+    return nullptr;
+  }
 
   friend KOKKOS_FUNCTION constexpr bool operator==(Array const&,
                                                    Array const&) noexcept {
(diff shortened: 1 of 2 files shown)


core/src/Kokkos_Array.hpp
```diff
@@ -435,6 +435,44 @@ KOKKOS_FUNCTION constexpr T const&& get(Array<T, N> const&& a) noexcept {
 }  // namespace Kokkos
 //</editor-fold>
 
+//<editor-fold desc="Support for range-based for loop">
+namespace Kokkos {
+
+template <class T, std::size_t N>
+constexpr T const* begin(Array<T, N> const& a) {
```
nliber, line 442: Shouldn't these all be decorated with `KOKKOS_FUNCTION`?

Also, the corresponding functions in `std::array` are marked `noexcept`.

core/unit_test/TestArray.cpp
```diff
@@ -187,6 +187,94 @@ constexpr bool test_to_array() {
...
+    return a;
+  }();
+  static_assert(accumulate(a3, 0) == 6);
+
+  return true;
+}
+
+static_assert(test_range_based_for_loop());
+
+constexpr bool test_begin_end() {
```
dalg24, line 244: Now wondering whether I should have qualified all the begin and end calls below.
Let me know if you want me to add `Kokkos::`.

nliber (changes requested):

PhilMiller (approved):
Looks good to me. Tests exercise what one hopes they would.

core/unit_test/TestArray.cpp
nliber: I'm fine either way.

PhilMiller: I think leaving them to go via ADL is preferable. You might even stick in `using std::begin;` to exercise the name resolution. I bet it turns up a bug in a vendor compiler at some point.

core/src/Kokkos_Array.hpp
dalg24: Fixed

nliber (approved):

core/src/Kokkos_Array.hpp
```diff
@@ -435,6 +435,46 @@ KOKKOS_FUNCTION constexpr T const&& get(Array<T, N> const&& a) noexcept {
...
+//<editor-fold desc="Support for range-based for loop">
+namespace Kokkos {
+
+template <class T, std::size_t N>
+KOKKOS_FUNCTION constexpr T const* begin(
+    [[maybe_unused]] Array<T, N> const& a) noexcept {
+  if constexpr (N == 0)
+    return nullptr;
+  else
+    return &a[0];
```
romintomasetti, line 447: Would the following work ?
```suggestion
  return a.data();
```

Noting that from https://en.cppreference.com/w/cpp/container/array/data:
> If [size()](https://en.cppreference.com/w/cpp/container/array/size) is ​0​, data() may or may not return a null pointer.

So you might discard my comment.

dalg24: Actually I like your idea better

core/src/Kokkos_Array.hpp
```diff
@@ -195,8 +195,10 @@ struct Array<T, 0> {
     return *reinterpret_cast<const_pointer>(-1);
   }
 
-  KOKKOS_INLINE_FUNCTION pointer data() { return nullptr; }
-  KOKKOS_INLINE_FUNCTION const_pointer data() const { return nullptr; }
+  KOKKOS_INLINE_FUNCTION constexpr pointer data() { return nullptr; }
```
romintomasetti, line 198:
```suggestion
  KOKKOS_FUNCTION constexpr pointer data() { return nullptr; }
```

(you already get `inline` implicitly)

dalg24: I'd agree with you but here consistency win and I really don't want to touch all function members.

core/unit_test/TestArray.cpp
```diff
@@ -187,6 +187,102 @@ constexpr bool test_to_array() {
...
+// also checking that it supports host device annotations
+template <class T, size_t N>
+KOKKOS_FUNCTION constexpr void iota(Kokkos::Array<T, N>& a, T value) {
+  for (auto& e : a) {
+    e = value++;
+  }
+}
+
+template <class T, size_t N>
+KOKKOS_FUNCTION constexpr T accumulate(Kokkos::Array<T, N> const& a, T init) {
```
romintomasetti, line 201: For a reference implementation that uses the `init` directly in the code (no need for `acc`): https://github.com/llvm/llvm-project/blob/c28b1a19aadff97b369889aee084073a181cfda8/libcxx/include/__numeric/accumulate.h#L27. Not requesting you to change anything if you're not willing to :smile:

romintomasetti (approved):

tpadioleau (approved):
Cool feature!

masterleinad (approved):
I'm still in favor of adding the member functions as well.

PhilMiller:
https://stackoverflow.com/questions/15608366/is-performing-arithmetic-on-a-null-pointer-undefined-behavior
`nullptr +size()` might be inherently UB

masterleinad:
> `nullptr + size()` might be inherently UB

But only if `size()` !=0.

dalg24:
> > `nullptr + size()` might be inherently UB
> 
> But only if `size()` !=0.

What Daniel said.

I should have mentioned I had tried
```C++
static_assert(static_cast<int*>(nullptr)+0==nullptr);
```
on the compiler explorer in pedantic mode on a few compiler to convince me that there was no need for special casing when I decided to follow Romin's suggestion

nliber (approved):

dalg24:
Ignoring CI

ndellingwood:
@dalg24 should this get a changelog entry?

dalg24:
Yes adding it now

ndellingwood:
Yeah, I don't see the cross-link on this PR that usually shows up. Sorry for the ping @dalg24

PR #7296, opened 2024-09-05
simd: Fix implicit copy assignment operator in AVX2 mask being deleted

### LINKED ISSUE #7278: simd_mask of int32_t don't have copy assignment on AVX2
**Describe the bug**

Assignment between `simd_mask`s of int32_t does not work on AVX2, but works on the `scalar` ABI.

- Compilers: g++ 11.4.0, icpx 2024.2.1, clang++ 17.0.6, etc. Any C++11 compiler should reproduce.
- Kokkos version: 4.4.0 release
- Platform: Ubuntu 22.04 on intel 9700K

To reproduce,
```cpp
#include <Kokkos_SIMD.hpp>

using namespace Kokkos::Experimental;

using mask_t = simd_mask<int32_t, simd_abi::native<int32_t>>;

void foo(mask_t& a, mask_t b) {
    a = b;
}
```
Compiled with `g++/clang++/icpx test.cpp -c -march=core-avx2 -DKOKKOS_ARCH_AVX2 -I/path/to/kokkos/simd/src`. `clang++-17` output:
```
simd.cpp:8:7: error: object of type 'mask_t' (aka 'simd_mask<int, avx2_fixed_size<4>>') cannot be assigned because its copy assignment operator is implicitly deleted
    8 |     a = b;
      |       ^
/home/jyren/Softwares/kokkos/kokkos-4.4.00/simd/src/Kokkos_SIMD_AVX2.hpp:366:41: note: copy assignment operator is implicitly deleted because 'simd_mask<int, Kokkos::Experimental::simd_abi::avx2_fixed_size<4>>' has a user-declared move constructor
  366 |   KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION simd_mask(simd_mask&&)      = default;
      |                                         ^
1 error generated.
```
Remove the `-DKOKKOS_ARCH_AVX2` option then the source is compiled without errors.

Also, it would be convenient if conversions between different `simd_mask` types with the same ABI could be added.

ldh4:
Fixes #7278 

Implicit copy assignment operators for few AVX2 mask types were being defined as deleted, because move constructors were declared in those mask classes.
Removed explicit move and copy constructors from those mask classes to restore implicit operators. They are now consistent with other mask types.

### DIFF
--- a/simd/src/Kokkos_SIMD_AVX2.hpp
+++ b/simd/src/Kokkos_SIMD_AVX2.hpp
@@ -361,9 +361,7 @@ class simd_mask<std::int32_t, simd_abi::avx2_fixed_size<4>> {
   };
   using value_type = bool;
   using abi_type   = simd_abi::avx2_fixed_size<4>;
-  KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION simd_mask()                 = default;
-  KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION simd_mask(simd_mask const&) = default;
-  KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION simd_mask(simd_mask&&)      = default;
+  KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION simd_mask() = default;
   KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION explicit simd_mask(value_type value)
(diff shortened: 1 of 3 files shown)


masterleinad (approved):

crtrott (approved):

dalg24:
Failures are unrelated

dalg24:
Make sure you add this to the change log under bug fixes

PR #7298, opened 2024-09-05
Prefer `ExecSpace::name()` to `typeid(ExecSpace).name()` in `print_configuration()`

dalg24:
Take advantage of the execution space static member function that returns their name when printing the configuration and avoid outputting mangled information.

On my local setup:
```diff
--- before
+++ after
@@ -3,7 +3,7 @@
   KOKKOS_COMPILER_APPLECC: 6000
 Architecture:
   CPU architecture: none
-  Default Device: N6Kokkos7ThreadsE
+  Default Device: Threads
   GPU architecture: none
   platform: 64bit
 Atomics:

### DIFF
--- a/core/src/impl/Kokkos_Core.cpp
+++ b/core/src/impl/Kokkos_Core.cpp
@@ -613,7 +613,7 @@ void pre_initialize_internal(const Kokkos::InitializationSettings& settings) {
 #endif
 
   declare_configuration_metadata("architecture", "Default Device",
-                                 typeid(Kokkos::DefaultExecutionSpace).name());
+                                 Kokkos::DefaultExecutionSpace::name());
 
 #if defined(KOKKOS_ARCH_A64FX)
   declare_configuration_metadata("architecture", "CPU architecture", "A64FX");

### CODE REVIEW

**Review by crtrott — APPROVED** [review]

**Review by aprokop — APPROVED** [review]

**Review by romintomasetti — APPROVED** [review]

File: core/src/impl/Kokkos_Core.cpp [hunk 1]
```diff
@@ -613,7 +613,7 @@ void pre_initialize_internal(const Kokkos::InitializationSettings& settings) {
 #endif
 
   declare_configuration_metadata("architecture", "Default Device",
-                                 typeid(Kokkos::DefaultExecutionSpace).name());
+                                 Kokkos::DefaultExecutionSpace::name());
```
romintomasetti, line 616: Just surprised that there is no test for that.

Could be a simple regex search in the output.

dalg24:
Failures are unrelated and fixed in #7299

core/src/impl/Kokkos_Core.cpp
dalg24: There is no test for it currently.  The 1st "incremental" test calls `ExecSpace::print_configuration()` directly but does not assert anything.
We can pursue this elsewhere, this change stands on its own.

PR #7307, opened 2024-09-09
Readme: upgrade to 4.4.00

cedricchevalier19:
Link to the latest release, Kokkos 4.4.00.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,12 +30,12 @@ To start learning about Kokkos:
 
 The latest release of Kokkos can be obtained from the [GitHub releases page](https://github.com/kokkos/kokkos/releases/latest).
 
-The current release is [4.3.01](https://github.com/kokkos/kokkos/releases/tag/4.3.01).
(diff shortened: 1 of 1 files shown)


cedricchevalier19:
I agree. The problem is that Kokkos does not follow the GitHub way to do the tarballs, so it is not very easy to provide a one-liner that downloads the latest release.

The one-liner was added when we decided to make `develop` the default branch, to provide something simple for users to copy and paste in their scripts to get the latest release.

nmm0 (approved):
Looks good, agreed it's not ideal to have to keep updating but not sure if there is a better way

masterleinad (approved):
Fine with we. We should just make this part of the release checklist (or just making sure we update when we touch `KOKKOS_VERSION`.

dalg24:
> I agree. The problem is that Kokkos does not follow the GitHub way to do the tarballs, so it is not very easy to provide a one-liner that downloads the latest release.

What is the GitHub way to do tarballs?

Issue #6974, opened 2024-05-01
layout() function broken for Views with padding -- path forward


crtrott:
We have an issue because the LayoutLeft and LayoutRight class itself (not the mapping) has no place to store stride. As a consequence the following code is wrong:

```c++
Kokkos::View<int**,Kokkos::LayoutLeft> a("A",N,M);
Kokkos::View<int**,Kokkos::LayoutLeft> a_sub =
  Kokkos::subview(a, Kokkos::pair{3,5},Kokkos::ALL());

Kokkos::View<int**,Kokkos::LayoutLeft> b_sub(a.sub.data(), a_sub.layout());
//b_sub.stride(1) != a_sub.stride(1);
```

Suggested path forward:

- restrict `layout()` to `Kokkos::LayoutLeft/Right/Stride` and deprecate
  - error out when calling `layout()` where `stride!=extent`
- introduce `View::mapping()` -> return mdspan mappings
- introduce `View(data_handle, mdspan_mapping_type map)`

masterleinad:
It's actually not stride(1) but stride(2) that fails: https://godbolt.org/z/14q5o136r

crtrott:
Because the code should have used `a_sub.layout()` corrected.

crtrott:
Also accessing stride(2) on a rank-2 view is kinda UB lol

PR #7308, opened 2024-09-09
View-refactor: Use stride in layout instance

crtrott:
This lets instances of LayoutRight and LayoutLeft carry stride information, and also use that information in LayoutRight/Left to mdspan mapping conversions. 

One issue this resolves is that the following code would actually not be valid:

```c++
using view_t = Kokkos::View<int**, Kokkos::LayoutLeft>;
view_t a("A", 20, 30);
view_t sub = Kokkos::subview(a, Kokkos::pair{5,15}, Kokkos::pair{5,25});
view_t sub2(sub.data(), sub.layout());
```

`sub2` will have lost the stride information, i.e. `sub` and `sub2` do not view the same data because `sub2` won't have the same stride as `sub`. https://godbolt.org/z/WEhxWTWE3

Note that `LayoutRight` to `layout_right_padded` is complicated by the fact that they are in-fact not the same Layout for Rank>2. `LayoutRight` has its custom stride for `stride(0)`; `layout_right_padded` has its custom stride for `stride(rank-2)`.

For a Rank-3 `View`/`mdspan` that means with `LayoutRight` the stride of dimension 0 is custom for `layout_right_padded` its the stride for dimension 1. We overlooked this for the longest time. Note that this property of `LayoutRight` is not getting exploited, by much in Kokkos. Specifically `subview` does not leverage that stride capability. 

Say you have a 3-D View and you take different types of subviews, either using a `pair` on the first, second or third dimension and otherwise use `ALL`
```c++
Kokkos::View<int***, LayoutLeft> a("A",N,M,K);
// returns LayoutLeft
auto sub_a = Kokkos::subview(a, Kokkos::pair{1,3}, Kokkos::ALL(), Kokkos::ALL());
// returns LayoutStride
auto sub_b = Kokkos::subview(a, Kokkos::ALL(), Kokkos::pair{1,3}, Kokkos::ALL());
// returns LayoutLeft
auto sub_c = Kokkos::subview(a, Kokkos::ALL(), Kokkos::ALL(), Kokkos::pair{1,3});

Kokkos::View<int***, LayoutRight> a("A",N,M,K);
// returns LayoutRight
auto sub_a = Kokkos::subview(a, Kokkos::pair{1,3}, Kokkos::ALL(), Kokkos::ALL());
// returns LayoutStride but LayoutRight would be possible
auto sub_b = Kokkos::subview(a, Kokkos::ALL(), Kokkos::pair{1,3}, Kokkos::ALL());
// returns LayoutStride but with layout_right_padded this could return layout_right_padded
auto sub_c = Kokkos::subview(a, Kokkos::ALL(), Kokkos::ALL(), Kokkos::pair{1,3});
```

So the only way you can get into the situation where you have an incompatible `LayoutRight` to `layout_right_padded` conversion is if you asked for padding during construction of a `Kokkos::View` of Rank-3 or larger.

We don't even allow this stride thing in a LayoutStride -> LayoutRight conversion right now. 

Fixes https://github.com/kokkos/kokkos/issues/6974.

### DIFF
--- a/core/src/Kokkos_Layout.hpp
+++ b/core/src/Kokkos_Layout.hpp
@@ -52,11 +52,9 @@ struct LayoutLeft {
   using array_layout = LayoutLeft;
 
   size_t dimension[ARRAY_LAYOUT_MAX_RANK];
-  // we don't have a constructor to set this directly
+  // we don't have a constructor to set the stride directly
   // but we will deprecate the class anyway (or at least using an instance of
-  // this class) so I believe its acceptable to not introduce a new ctor and
-  // simply set this value directly in the internal functions where we construct
-  // instances of LayoutLeft
+  // this class) when switching the internal implementation to use mdspan
   size_t stride;
 
   enum : bool { is_extent_constructible = true };
@@ -108,11 +106,9 @@ struct LayoutRight {
   using array_layout = LayoutRight;
 
   size_t dimension[ARRAY_LAYOUT_MAX_RANK];
-  // we don't have a constructor to set this directly
+  // we don't have a constructor to set the stride directly
   // but we will deprecate the class anyway (or at least using an instance of
-  // this class) so I believe its acceptable to not introduce a new ctor and
-  // simply set this value directly in the internal functions where we construct
-  // instances of LayoutLeft
+  // this class) when switching the internal implementation to use mdspan
   size_t stride;
 
   enum : bool { is_extent_constructible = true };
--- a/core/src/Kokkos_Layout.hpp
+++ b/core/src/Kokkos_Layout.hpp
@@ -52,6 +52,12 @@ struct LayoutLeft {
   using array_layout = LayoutLeft;
 
   size_t dimension[ARRAY_LAYOUT_MAX_RANK];
+  // we don't have a constructor to set this directly
+  // but we will deprecate the class anyway (or at least using an instance of
+  // this class) so I believe its acceptable to not introduce a new ctor and
+  // simply set this value directly in the internal functions where we construct
+  // instances of LayoutLeft
+  size_t stride;
 
   enum : bool { is_extent_constructible = true };
 
@@ -69,7 +75,8 @@ struct LayoutLeft {
                                 size_t N5 = KOKKOS_IMPL_CTOR_DEFAULT_ARG,
                                 size_t N6 = KOKKOS_IMPL_CTOR_DEFAULT_ARG,
                                 size_t N7 = KOKKOS_IMPL_CTOR_DEFAULT_ARG)
-      : dimension{N0, N1, N2, N3, N4, N5, N6, N7} {}
+      : dimension{N0, N1, N2, N3, N4, N5, N6, N7},
+        stride(KOKKOS_IMPL_CTOR_DEFAULT_ARG) {}
 
   friend bool operator==(const LayoutLeft& left, const LayoutLeft& right) {
     for (unsigned int rank = 0; rank < ARRAY_LAYOUT_MAX_RANK; ++rank)
@@ -101,6 +108,12 @@ struct LayoutRight {
   using array_layout = LayoutRight;
 
   size_t dimension[ARRAY_LAYOUT_MAX_RANK];
+  // we don't have a constructor to set this directly
+  // but we will deprecate the class anyway (or at least using an instance of
+  // this class) so I believe its acceptable to not introduce a new ctor and
+  // simply set this value directly in the internal functions where we construct
+  // instances of LayoutLeft
+  size_t stride;
 
   enum : bool { is_extent_constructible = true };
 
@@ -118,7 +131,8 @@ struct LayoutRight {
                                  size_t N5 = KOKKOS_IMPL_CTOR_DEFAULT_ARG,
                                  size_t N6 = KOKKOS_IMPL_CTOR_DEFAULT_ARG,
                                  size_t N7 = KOKKOS_IMPL_CTOR_DEFAULT_ARG)
-      : dimension{N0, N1, N2, N3, N4, N5, N6, N7} {}
+      : dimension{N0, N1, N2, N3, N4, N5, N6, N7},
+        stride{KOKKOS_IMPL_CTOR_DEFAULT_ARG} {}
 
   friend bool operator==(const LayoutRight& left, const LayoutRight& right) {
     for (unsigned int rank = 0; rank < ARRAY_LAYOUT_MAX_RANK; ++rank)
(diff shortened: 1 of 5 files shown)


masterleinad:
```
/home/runner/work/kokkos/kokkos/kokkos/core/src/View/MDSpan/Kokkos_MDSpan_Layout.hpp:174:18: error: unused parameter 'layout' [-Werror,-Wunused-parameter]
    LayoutStride layout, std::index_sequence<>) {
                 ^
```

core/src/Kokkos_Layout.hpp
```diff
@@ -101,6 +108,12 @@ struct LayoutRight {
   using array_layout = LayoutRight;
 
   size_t dimension[ARRAY_LAYOUT_MAX_RANK];
+  // we don't have a constructor to set this directly
+  // but we will deprecate the class anyway (or at least using an instance of
+  // this class) so I believe its acceptable to not introduce a new ctor and
+  // simply set this value directly in the internal functions where we construct
+  // instances of LayoutLeft
```
masterleinad, line 115:
```suggestion
  // instances of LayoutRight
```

core/src/Kokkos_Layout.hpp
```diff
@@ -52,6 +52,12 @@ struct LayoutLeft {
   using array_layout = LayoutLeft;
 
   size_t dimension[ARRAY_LAYOUT_MAX_RANK];
+  // we don't have a constructor to set this directly
+  // but we will deprecate the class anyway (or at least using an instance of
+  // this class) so I believe its acceptable to not introduce a new ctor and
+  // simply set this value directly in the internal functions where we construct
+  // instances of LayoutLeft
```
masterleinad, line 59:
```suggestion
  // we don't have a constructor to set the stride directly
  // but we will deprecate the class anyway (or at least using an instance of
  // this class) when switching the internal implementation to use mdspan
```
It's strange having a comment in the code base that says "I".

core/src/Kokkos_Layout.hpp
masterleinad: See above.

core/src/View/Kokkos_ViewMapping.hpp
```diff
@@ -1763,35 +1771,38 @@ struct ViewOffset<
...
+                                                                              ? size_t(
+                                                                                    1)
+                                                                              : m_dim.N6 *
+                                                                                    (dimension_type::
+                                                                                                 rank ==
+                                                                                             7
+                                                                                         ? size_t(
+                                                                                               1)
+                                                                                         : m_dim
+                                                                                               .N7)))))))) {
```
masterleinad, line 1805: Can you just introduce a free function that you call here to make this somewhat readable?

core/unit_test/TestMDSpanConversion.hpp
```diff
@@ -504,4 +506,53 @@ TEST(TEST_CATEGORY, view_mdspan_conversion) {
...
+    ASSERT_EQ(sub_v2.extent(1), 40);
+    ASSERT_EQ(sub_v2.extent(2), 36);
+    ASSERT_EQ(sub_v2.stride(0), 1);
+    ASSERT_EQ(sub_v2.stride(1), 20);
+    ASSERT_EQ(sub_v2.stride(2), 800);
+  }
+  {
+    // layout_right_padded<dynamic_extent> has a custom stride for
+    // stride(rank-2) LayoutRight has a custom stride for stride(0) That means
+    // the "padding" only mateches up for Rank-2 Views
```
masterleinad, line 538:
```suggestion
    // the "padding" only matches up for Rank-2 Views
```

crtrott: fixed

core/src/View/Kokkos_ViewMapping.hpp
crtrott: done - not that I think it helps much

core/src/Kokkos_Layout.hpp
crtrott: done (in both places)

core/src/View/Kokkos_ViewMapping.hpp
```diff
@@ -1754,6 +1762,43 @@ struct ViewOffset<
...
+                                                                             ? size_t(
+                                                                                   1)
+                                                                             : m_dim.N6 *
+                                                                                   (dimension_type::
+                                                                                                rank ==
+                                                                                            7
+                                                                                        ? size_t(
+                                                                                              1)
+                                                                                        : m_dim
+                                                                                              .N7)))))));
```
masterleinad, line 1799: Since it's now in a fucntion, we don't have to write this as one statement.
```suggestion
    if (arg_layout.stride != KOKKOS_IMPL_CTOR_DEFAULT_ARG)
      return arg_layout.stride;
    size_type result = m_dim.N1;
    if (dimension_type::rank >= 3)
      result *= m_dim.N2;
    if (dimension_type::rank >= 4)
      result *= m_dim.N3;
    if (dimension_type::rank >= 5)
      result *= m_dim.N4;
    if (dimension_type::rank >= 6)
      result *= m_dim.N5;
    if (dimension_type::rank >= 7)
      result *= m_dim.N6;
    if (dimension_type::rank >= 8)
      result *= m_dim.N7;
    return Padding<TrivialScalarSize>::stride(result);
```

masterleinad (approved):

core/src/View/MDSpan/Kokkos_MDSpan_Layout.hpp
```diff
@@ -77,45 +81,106 @@ KOKKOS_INLINE_FUNCTION auto array_layout_from_mapping(
...
+      layout.stride = mapping.stride(1);
+    }
+    if constexpr (std::is_same_v<typename mapping_type::layout_type,
+                                 Kokkos::Experimental::layout_right_padded<
+                                     dynamic_extent>>) {
+      if constexpr (rank == 2) {
+        layout.stride = mapping.stride(0);
+      }
+      if constexpr (rank > 2) {
+        if (mapping.stride(rank - 2) != mapping.extents().extent(rank - 1))
```
nmm0, line 106: This is from LayoutRIght having a left padding (unlike layout_right_padded) right?

nmm0 (approved):
Looks good to me, though we will need to make a custom layout for LayoutRight

core/src/View/MDSpan/Kokkos_MDSpan_Layout.hpp
crtrott: yes, and no we won't make a custom LayoutRight, we just gonna disable padding for Rank>3. As I said above we never leverage the stride capability (e.g. in subview). So we are actually fine.

PR #7309, opened 2024-09-09
SYCL: Introduce RegisterFileSize

masterleinad:
Similar to https://github.com/kokkos/kokkos/pull/6496, this pull request introduces a new policy property called `RegisterFileSize` that allows forcing the register file size to small or large for Intel GPUs with the SYCL backend. See https://www.intel.com/content/www/us/en/docs/oneapi/optimization-guide-gpu/2024-1/porting-registers.html for some cases here setting this to different values makes a measurable difference.
The implementation is based on an extension described in https://github.com/intel/llvm/blob/sycl/sycl/doc/extensions/experimental/sycl_ext_intel_grf_size.asciidoc.

Based on top of #6496.


core/src/Kokkos_Concepts.hpp
```diff
@@ -120,6 +120,20 @@ struct LaunchBounds {
   static constexpr unsigned int minBperSM{minB};
 };
 
+template <int size>
+struct SubGroupSize {
```
romintomasetti, line 124: Do these make sense for all backends ?

masterleinad: No, it's for SYCL only, just as `LaunchBounds` is for Cuda and HIP only.

masterleinad: Also, this pull request introduces `RegisterFileSize` not `SubGroupSize`.

romintomasetti: OK, but they are defined at global level, and visible always. Is it intended ? (just a question for my understanding)

romintomasetti: OK. Food for discussion then. There might be a similar `CUDA` option called [__maxnreg__](https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#maximum-number-of-registers-per-thread).

IMO anything in `Kokkos_Concepts.hpp` should have a broad adoption throughout the majority of the supported backends, otherwise it does not feel like a concepts but rather a backend-specific tuning property.

romintomasetti (changes requested):
Not sure I can find a test for the added `RegisterFileSize` "concept".

PR #7311, opened 2024-09-09
Rework event-based tool testing facility to not rely on RTTI

dalg24:
Partial fix for #7037 

Together with #7043 enables building with `-fno-rtti`

Implements some hand-rolled form of RTTI to avoid doing dynamic pointer cast to determine if an object of EventBase class is indeed of EventDerived class.  The UID values are meaningless but they get the job done.  The alternative would be to define some sort of enum that knows all the set of derived classes which is quite cumbersome.


masterleinad (approved):
Looks reasonable to me.

crtrott (approved):
OK I think I understand what's going on. And who am I to stand in the way of a PR which passed cleanly all our CI!!!

PR #7312, opened 2024-09-10
View-refactor: Remove use of View implementation details in ScatterView

crtrott:
This also introduces terminating specializations of an implementation detail of ScatterView for the standard layouts, which we will need for the View refactor.

### DIFF
--- a/containers/src/Kokkos_ScatterView.hpp
+++ b/containers/src/Kokkos_ScatterView.hpp
@@ -532,31 +532,55 @@ void args_to_array(size_t* array, int pos, T dim0, Dims... dims) {
    subview where the index specified is the largest-stride one. */
 template <typename Layout, int rank, typename V, typename... Args>
 struct Slice {
-  using next       = Slice<Layout, rank - 1, V, Kokkos::ALL_t, Args...>;
-  using value_type = typename next::value_type;
-
-  static value_type get(V const& src, const size_t i, Args... args) {
+  using next = Slice<Layout, rank - 1, V, Kokkos::ALL_t, Args...>;
+  static auto get(V const& src, const size_t i, Args... args) {
     return next::get(src, i, Kokkos::ALL, args...);
   }
 };
 
 template <typename V, typename... Args>
 struct Slice<Kokkos::LayoutRight, 1, V, Args...> {
-  using value_type =
-      typename Kokkos::Impl::ViewMapping<void, V, const size_t, Args...>::type;
-  static value_type get(V const& src, const size_t i, Args... args) {
+  static auto get(V const& src, const size_t i, Args... args) {
     return Kokkos::subview(src, i, args...);
   }
 };
 
 template <typename V, typename... Args>
 struct Slice<Kokkos::LayoutLeft, 1, V, Args...> {
-  using value_type =
-      typename Kokkos::Impl::ViewMapping<void, V, Args..., const size_t>::type;
-  static value_type get(V const& src, const size_t i, Args... args) {
+  static auto get(V const& src, const size_t i, Args... args) {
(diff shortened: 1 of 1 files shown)


containers/src/Kokkos_ScatterView.hpp
```diff
@@ -532,32 +532,56 @@ void args_to_array(size_t* array, int pos, T dim0, Dims... dims) {
...
+  }
+};
+
+template <typename V, size_t Pad, typename... Args>
+struct Slice<Kokkos::Experimental::layout_left_padded<Pad>, 1, V, Args...> {
+  static auto get(V const& src, const size_t i, Args... args) {
+    return Kokkos::subview(src, args..., i);
+  }
+};
+#endif
```
dalg24, line 583: This is not what your PR says it does

dalg24: Nevermind you touched on it in the description

masterleinad (approved):

containers/src/Kokkos_ScatterView.hpp
```diff
@@ -532,32 +532,56 @@ void args_to_array(size_t* array, int pos, T dim0, Dims... dims) {
    subview where the index specified is the largest-stride one. */
 template <typename Layout, int rank, typename V, typename... Args>
 struct Slice {
-  using next       = Slice<Layout, rank - 1, V, Kokkos::ALL_t, Args...>;
-  using value_type = typename next::value_type;
-
-  static value_type get(V const& src, const size_t i, Args... args) {
+  using next = Slice<Layout, rank - 1, V, Kokkos::ALL_t, Args...>;
```
romintomasetti, line 535: This is just a remark, not a request for change.

In our code, we also have a `slice` utility but it seems not exactly for the same purpose. Ours is more like:
```c++
// take a subview, add ALL for remaining not-given indices up to rank
template <typename view_t, typename... Indices>
auto slice(view_t&& view, Indices&&... indices);

...

auto sliced = slive(my_view_of_rank_4, 1, 2); // equivalent to subview(my_view_of_rank_4, 1, 2, ALL, ALL);
```

It is a utility that could be thought of a bit more and promoted.

romintomasetti (dismissed):
- One remark about the slicing utility.
- One remark about test coverage.

containers/src/Kokkos_ScatterView.hpp
```diff
@@ -532,32 +532,56 @@ void args_to_array(size_t* array, int pos, T dim0, Dims... dims) {
...
-  using value_type =
-      typename Kokkos::Impl::ViewMapping<void, V, Args..., const size_t>::type;
-  static value_type get(V const& src, const size_t i, Args... args) {
+  static auto get(V const& src, const size_t i, Args... args) {
+    return Kokkos::subview(src, args..., i);
+  }
+};
+
+#ifdef KOKKOS_ENABLE_IMPL_MDSPAN
+template <typename V, typename... Args>
```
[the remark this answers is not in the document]
crtrott: This can't be tested in current CI, because View does not take the new layouts yet. It will be tested when View gets changed implicitly. But I can't change View without having this specialization, and I feel it's better to introduce these changes incrementally initially in order to not touch 20 files when I introduce the new View implementation.

containers/src/Kokkos_ScatterView.hpp
crtrott: This thing here is very ScatterView specific, because it sticks the all on the appropriate side. Beyond that this PR is solely targeted at overhauling ScatterView to allow for the View refactor.

romintomasetti (commented):
I did not wanted to block.

PR #7313, opened 2024-09-10
View-refactor: Updating Offsetview to use public APIs of View. 

crtrott:
I am changing OffsetView to inherit publicly from View. Many functions such as as `extent()`, `rank()`, and what not are the same as for `View`. It really is just the additional stuff around the offsets (i.e. ctors and `operator()`) that are different.

### DIFF
--- a/containers/src/Kokkos_OffsetView.hpp
+++ b/containers/src/Kokkos_OffsetView.hpp
@@ -489,14 +489,22 @@ class OffsetView : public View<DataType, Properties...> {
                              const E& ends_, subtraction_failure)
       : base_t(Kokkos::view_wrap(p),
                typename traits::array_layout(
-                   base_t::rank() > 0 ? at(ends_, 0) - at(begins_, 0) : 0,
-                   base_t::rank() > 1 ? at(ends_, 1) - at(begins_, 1) : 0,
-                   base_t::rank() > 2 ? at(ends_, 2) - at(begins_, 2) : 0,
-                   base_t::rank() > 3 ? at(ends_, 3) - at(begins_, 3) : 0,
-                   base_t::rank() > 4 ? at(ends_, 4) - at(begins_, 4) : 0,
-                   base_t::rank() > 5 ? at(ends_, 5) - at(begins_, 5) : 0,
-                   base_t::rank() > 6 ? at(ends_, 6) - at(begins_, 6) : 0,
-                   base_t::rank() > 7 ? at(ends_, 7) - at(begins_, 7) : 0)) {
+                   base_t::rank() > 0 ? at(ends_, 0) - at(begins_, 0)
+                                      : KOKKOS_IMPL_CTOR_DEFAULT_ARG,
+                   base_t::rank() > 1 ? at(ends_, 1) - at(begins_, 1)
+                                      : KOKKOS_IMPL_CTOR_DEFAULT_ARG,
+                   base_t::rank() > 2 ? at(ends_, 2) - at(begins_, 2)
+                                      : KOKKOS_IMPL_CTOR_DEFAULT_ARG,
+                   base_t::rank() > 3 ? at(ends_, 3) - at(begins_, 3)
+                                      : KOKKOS_IMPL_CTOR_DEFAULT_ARG,
+                   base_t::rank() > 4 ? at(ends_, 4) - at(begins_, 4)
+                                      : KOKKOS_IMPL_CTOR_DEFAULT_ARG,
+                   base_t::rank() > 5 ? at(ends_, 5) - at(begins_, 5)
+                                      : KOKKOS_IMPL_CTOR_DEFAULT_ARG,
+                   base_t::rank() > 6 ? at(ends_, 6) - at(begins_, 6)
+                                      : KOKKOS_IMPL_CTOR_DEFAULT_ARG,
+                   base_t::rank() > 7 ? at(ends_, 7) - at(begins_, 7)
+                                      : KOKKOS_IMPL_CTOR_DEFAULT_ARG)) {
     for (size_t i = 0; i != m_begins.size(); ++i) {
       m_begins[i] = at(begins_, i);
     };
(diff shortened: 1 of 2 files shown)


containers/src/Kokkos_OffsetView.hpp
```diff
@@ -180,30 +180,27 @@ void runtime_check_rank_device(const size_t rank_dynamic, const size_t rank,
...
+
+  using begins_type = Kokkos::Array<int64_t, base_t::rank()>;
 
   template <typename iType,
             std::enable_if_t<std::is_integral_v<iType>, iType> = 0>
   KOKKOS_FUNCTION int64_t begin(const iType local_dimension) const {
-    return local_dimension < Rank ? m_begins[local_dimension]
-                                  : KOKKOS_INVALID_OFFSET;
+    return local_dimension < base_t::rank() ? m_begins[local_dimension]
+                                            : KOKKOS_INVALID_OFFSET;
```
dalg24, line 203:
```
/__w/kokkos/kokkos/containers/unit_tests/../src/Kokkos_OffsetView.hpp:202:28: error: comparison of integer expressions of different signedness: 'const int' and 'Kokkos::Impl::integral_constant<long unsigned int, 2>::value_type' {aka 'long unsigned int'} [-Werror=sign-compare]
  202 |     return local_dimension < base_t::rank() ? m_begins[local_dimension]
      |            ~~~~~~~~~~~~~~~~^~~~~~~~~~~~~~~~

File: containers/unit_tests/TestOffsetView.hpp [hunk 2]
```diff
@@ -56,7 +56,10 @@ void test_offsetview_construction() {
   offset_view_type ov("firstOV", range0, range1);
 
   ASSERT_EQ("firstOV", ov.label());
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
   ASSERT_EQ(2, ov.Rank);
+#endif
+  ASSERT_EQ(2, ov.rank());
 
   ASSERT_EQ(ov.begin(0), -1);
```
dalg24, line 64:
```
/__w/kokkos/kokkos/containers/unit_tests/TestOffsetView.hpp:64:3:   required from 'void Test::test_offsetview_construction() [with Scalar = int; Device = Kokkos::OpenMP]'
   64 |   ASSERT_EQ(ov.begin(0), -1);
      |             ~~~~~~~~^~~

File: containers/src/Kokkos_OffsetView.hpp [hunk 3]
```diff
@@ -245,488 +240,47 @@ class OffsetView : public ViewTraits<DataType, Properties...> {
...
-                                 j1 * m_map.m_impl_offset.m_stride.S1];
-    }
-#if defined(KOKKOS_COMPILER_INTEL)
-    __builtin_unreachable();
+  KOKKOS_FUNCTION constexpr typename base_t::reference_type operator[](
+      const OtherIndexType& idx) const {
+#ifdef KOKKOS_ENABLE_CXX17
+    static_assert(std::is_convertible_v<OtherIndexType, index_type> &&
+                  std::is_nothrow_constructible_v<index_type, OtherIndexType> &&
+                  (base_t::rank() == 1));
```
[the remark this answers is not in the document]
crtrott: There isn't really a useful difference between the two for these purposes anyway (i.e. it's not like there is gonna be a different overload I need to disambiguate). So I did write the C++20 style constraint, and in order not to get stupid error messages in C++17 mode I added the static assert.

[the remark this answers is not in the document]
dalg24: If there was no difference then why don;t you stick to the static assertion?

containers/unit_tests/TestOffsetView.hpp
```diff
@@ -586,6 +589,7 @@ void test_offsetview_offsets_rank1() {
       KOKKOS_LAMBDA(const int ii, int& lerrors) {
         offset_view_type ov(v, {ii});
         lerrors += (ov(3) != element({3 - ii}));
+        lerrors += (ov[3] != element({3 - ii}));
```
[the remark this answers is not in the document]
crtrott: It wasn't tested before (the 1D `[]` operator, so I tested it now since I actually deleted the operator and it would have picked up the one from `View`).

containers/src/Kokkos_OffsetView.hpp
```diff
@@ -180,30 +180,27 @@ void runtime_check_rank_device(const size_t rank_dynamic, const size_t rank,
 }  // namespace Impl
 
 template <class DataType, class... Properties>
-class OffsetView : public ViewTraits<DataType, Properties...> {
- public:
-  using traits = ViewTraits<DataType, Properties...>;
-
+class OffsetView : public View<DataType, Properties...> {
```
dalg24, line 183: Please discuss that publicly inheriting from View in the description of your PR

containers/src/Kokkos_OffsetView.hpp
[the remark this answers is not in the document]
crtrott: because it originally was a constraint and the error message we get with proper requires clauses is nice

dalg24:
> I am changing OffsetView to inherit publicly from View. Many functions such as as `extent()`, `rank()`, and what not are the same as for `View`. It really is just the additional stuff around the offsets (i.e. ctors and `operator()`) that are different.



> I am changing OffsetView to inherit publicly from View. Many functions such as as `extent()`, `rank()`, and what not are the same as for `View`. It really is just the additional stuff around the offsets (i.e. ctors and `operator()`) that are different.

What about `View::access()`.  You probably want to pull that into OffsetView as is?

crtrott:
> What about `View::access()`. You probably want to pull that into OffsetView as is?

yeah probably we should :-(

crtrott:
There are two failing tests on Jenkins, one of them worrisome, but neither related to this PR. I.e. the first failure is a problem with reference counting during testing in the Kokkos_CoreUnitTest_Serial1 and the second is NVHPC compilation segfaulting in the algorithms sub package. Since this PR only touches Offsetview neither is related.

dalg24 (approved):
Christian promised me he audited that pulling the View API was ok for real now

containers/src/Kokkos_OffsetView.hpp
```diff
@@ -180,30 +180,28 @@ void runtime_check_rank_device(const size_t rank_dynamic, const size_t rank,
...
-  using track_type = Kokkos::Impl::SharedAllocationTracker;
+  using base_t = View<DataType, Properties...>;
 
  public:
-  enum { Rank = map_type::Rank };
-  using begins_type = Kokkos::Array<int64_t, Rank>;
+  // typedefs to reduce typing base_t:: further down
+  using traits = typename base_t::traits;
+  // FIXME: should be base_t::index_type after refactor
+  using index_type   = typename base_t::memory_space::size_type;
```
aprokop, line 194: Maybe move `index_type` to private, as it was not exposed in public in the previous implementation?
Also, previously, `reference_type` was part of the public interface.

crtrott: `OffsetView` is derived from `Kokkos::View` so it gets `reference_type` from there. I will actually submit a PR which introduces `index_type` for `Kokkos::View` (the legacy version). The new version will anyway have it because I am bringing it in line with `mdspan`.

aprokop (approved):

PR #7314, opened 2024-09-10
Don't report SYCL as experimental in the configuration summary anymore 

masterleinad:
Follow-up to #7171. We were still printing
```
[...]
--     Device Parallel: Kokkos::Experimental::SYCL
[...]
```
after running `CMake`.
Drive-by change: `_DEFAULT_DEVICE_MEMSPACE` was unused (and reported a non-existing type for SYCL anyway) so it's removed as well.


aprokop (approved):
LGTM

dalg24 (approved):
It took me a minute to understand what you were doing.
Thanks for cleaning up,

dalg24:
One of the CUDA build did not detect the GPU at runtime.  Everything else passed.

PR #7306, opened 2024-09-09
Disable all thread safety tests for the OpenACC backend

seyonglee:
Temporarily disable all thread safety tests for the OpenACC backend until the thread safety issue is fixed.

PR #7318, opened 2024-09-11
Disable thread safety tests on the host backends as well when OpenACC is enabled

dalg24:
Fixup for #7306

Need do skip for all exec spaces because host-side atomics have issues which is affecting the shared allocation reference counting


dalg24:
The OpenACC build did pass.  Ignoring the rest.

Issue #6799, opened 2024-02-09
[FR] `HIP` backend: missing async malloc


romintomasetti:
## Summary of the feature request

Please enable asynchronous memory allocation support for the `HIP` backend (https://rocm.docs.amd.com/en/latest/conceptual/gpu-memory.html#memory-allocation).

## Description

`Cuda` and `HIP` both support asynchronous memory allocation through `cudaMallocAsync`/`hipMallocAsync`, respectively.

However, `Kokkos` does not implement asynchronous memory allocation for `HIP`, but does for `Cuda` (it can be enabled with `Kokkos_ENABLE_IMPL_CUDA_MALLOC_ASYNC`).

I'd be glad to help if I can :wink:

Rombur:
This was fixed in Kokkos 4.5

PR #7324, opened 2024-09-11
Use hipMallocAsync when using HIP execution space

Rombur:
This PR replaces `hipMalloc` with `hipMallocAsync`. This allows to fence only one stream instead of the entire device. For CUDA, we noticed that for small allocations, it was faster to use `cudMalloc` instead of `cudaMallocAsync`. I didn't see this behavior with HIP. `hipMallocAsync` is always at least as fast as `hipMalloc`, so we unconditionally use  `hipMallocAsync`.

Related to https://github.com/kokkos/kokkos/issues/6799


core/src/HIP/Kokkos_HIP_Space.hpp
```diff
@@ -98,10 +92,10 @@ class HIPSpace {
                   const size_t arg_logical_size = 0) const;
 
  private:
-  void* impl_allocate(const char* arg_label, const size_t arg_alloc_size,
-                      const size_t arg_logical_size = 0,
-                      const Kokkos::Tools::SpaceHandle =
-                          Kokkos::Tools::make_space_handle(name())) const;
+  void* impl_allocate(const hipStream_t stream, const char* arg_label,
```
romintomasetti, line 95: Just noting that most of the methods here still take `HIP`, not `hipStream_t`.

Also noting that for the `Cuda` backend, it also takes `Cuda` instead of the `cudaStream_t`.

core/src/HIP/Kokkos_HIP_Space.hpp
```diff
@@ -98,10 +92,10 @@ class HIPSpace {
...
+                      const size_t arg_alloc_size,
+                      const size_t arg_logical_size,
+                      bool stream_sync_only) const;
```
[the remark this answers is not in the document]
Rombur: I don't find this better than using `bool`

core/src/HIP/Kokkos_HIP_Space.cpp
```diff
@@ -51,35 +51,51 @@ static std::atomic<bool> is_first_hip_managed_allocation(true);
...
-    // This is the only way to clear the last error, which we should do here
-    // since we're turning it into an exception here
-    (void)hipGetLastError();
-    Kokkos::Impl::throw_bad_alloc(name(), arg_alloc_size, arg_label);
+  KOKKOS_IMPL_HIP_SAFE_CALL(hipMallocAsync(&ptr, arg_alloc_size, stream));
+  if (stream_sync_only) {
+    KOKKOS_IMPL_HIP_SAFE_CALL(hipStreamSynchronize(stream));
+  } else {
+    KOKKOS_IMPL_HIP_SAFE_CALL(hipDeviceSynchronize());
   }
```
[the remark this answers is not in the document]
Rombur: I don't think so. The entire allocation is already captured by KokkosP.

masterleinad (approved):
Fine with me.

core/src/HIP/Kokkos_HIP_Space.cpp
```diff
@@ -51,35 +51,57 @@ static std::atomic<bool> is_first_hip_managed_allocation(true);
 
 namespace Kokkos {
 
-HIPSpace::HIPSpace() : m_device(HIP().hip_device()) {}
+HIPSpace::HIPSpace()
+    : m_device(HIP().hip_device()), m_stream(Kokkos::HIP().hip_stream()) {}
```
romintomasetti, line 55:
```suggestion
HIPSpace::HIPSpace(const HIP& exec)
    : m_device(exec.hip_device()), m_stream(exec.hip_stream()) {}
```

romintomasetti: (and probably `explicit`)

core/src/HIP/Kokkos_HIP_Space.hpp
```diff
@@ -114,6 +108,7 @@ class HIPSpace {
 
  private:
   int m_device;  ///< Which HIP device
+  hipStream_t m_stream;
```
romintomasetti, line 111: Just wondering why you would not explore this:
```suggestion
  HIP exec;
```
since `HIP` knows the device and stream already.

masterleinad: Because we want to use this class in the `HIP` execution space constructor.

core/src/HIP/Kokkos_HIP_Space.cpp
Rombur: No

romintomasetti: You still have mismatch between `HIP().hip_device()` and `Kokkos::HIP().hip_stream()`...

core/src/HIP/Kokkos_HIP_Space.hpp
romintomasetti: It just feels like a lot of the `HIP`, `HIPSpace` and `HIPInternal` types (same for `Cuda` I guess) are extracting the stream and device ID themselves.

If Kokkos wants to strongly couple device ID and stream, it might make sense that have them together in one of these objects, right ?

stanmoore1:
We've had a ton of trouble with `cudaMallocAsync` making certain versions of MPI crash, e.g. see https://github.com/kokkos/kokkos/issues/4228. Has anyone tested if `hipMallocAsync` works with Cray MPICH on Frontier?

core/src/HIP/Kokkos_HIP_Space.cpp
Rombur: I don't understand what you mean with "there is a mismatch". The extra namespace?

romintomasetti: yes

Rombur:
@stanmoore1 If you have a test that  reliably crashes, I can run it on Frontier.

core/src/HIP/Kokkos_HIP_Space.hpp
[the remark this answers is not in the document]
cwpearson: I think `bool` is okay as long as there's only one.

core/src/HIP/Kokkos_HIP_Space.hpp
ldh4: I have a similar opinion as @romintomasetti here.

Not just for consistency, but if multi-gpu support is also to be enabled for `HIP` backend in the future, this function will either need to take in a `HIP` instance to extract a device id or add another argument for it.

Rombur: @romintomasetti You are wrong. This is the same interface as the CUDA interface. Look at `impl_allocate_common`.

cedricchevalier19:
@tpadioleau, can you report if it works on Adastra?

tpadioleau:
> @tpadioleau, can you report if it works on Adastra?

Yep seems fine in an application that uses MPI device-aware communications

core/src/HIP/Kokkos_HIP_Space.hpp
tcclevenger: @romintomasetti @ldh4 I have a PR which we can merge after this one https://github.com/kokkos/kokkos/pull/7130 that will add the necessary exe_space impls for multiGPU, no need to have them here.

core/src/HIP/Kokkos_HIP_Space.hpp
```diff
@@ -77,15 +77,9 @@ class HIPSpace {
     return allocate(arg_label, arg_alloc_size, arg_logical_size);
   }
 #else
-  // FIXME_HIP Use execution space instance
-  void* allocate(const HIP&, const size_t arg_alloc_size) const {
-    return allocate(arg_alloc_size);
-  }
-  // FIXME_HIP Use execution space instance
+  void* allocate(const HIP&, const size_t arg_alloc_size) const;
```
tcclevenger, line 80:
```suggestion
  void* allocate(const HIP& exec_space, const size_t arg_alloc_size) const;
```
Should we include the name since the function impl is using `exec_space`? To be consistent with `arg_alloc_size`.

core/src/HIP/Kokkos_HIP_Space.hpp
```diff
@@ -77,15 +77,9 @@ class HIPSpace {
...
   void* allocate(const HIP&, const char* arg_label, const size_t arg_alloc_size,
```
tcclevenger, line 81:
```suggestion
  void* allocate(const HIP& exec_space, const char* arg_label, const size_t arg_alloc_size,
```
Same here.

Rombur:
I have added a commit that makes `hipMallocAsync` an opt-in option. This way we don't get any bad surprise with MPI.

Rombur:
I have talked with AMD and there shouldn't be any problem enabling `hipMallocAsync`.

dalg24:
Do you still want disabled by default then?

Rombur:
> Do you still want disabled by default then?

Yes, I want it off by default for one release. So people have time to try it  and make sure it works.

dalg24:
One CUDA build had unrelated runtime failures (GPU not detected at runtime)
HIP builds are passing

tcclevenger (approved):

dalg24:
The HIP builds are passing.

PR #7327, opened 2024-09-12
Failure to parse CTest resource allocation not recoverable

dalg24:
Proposing to call abort instead of throwing an exception to report errors in CTest resource allocation parsing

### DIFF
--- a/core/src/impl/Kokkos_Core.cpp
+++ b/core/src/impl/Kokkos_Core.cpp
@@ -271,7 +271,7 @@ int Kokkos::Impl::get_ctest_gpu(int local_rank) {
     ss << "Error: local rank " << local_rank
        << " is outside the bounds of resource groups provided by CTest. Raised"
        << " by Kokkos::Impl::get_ctest_gpu().";
-    throw_runtime_exception(ss.str());
+    abort(ss.str().c_str());
   }
 
   // Get the resource types allocated to this resource group
@@ -284,7 +284,7 @@ int Kokkos::Impl::get_ctest_gpu(int local_rank) {
     std::ostringstream ss;
     ss << "Error: " << ctest_resource_group_name << " is not specified. Raised"
        << " by Kokkos::Impl::get_ctest_gpu().";
-    throw_runtime_exception(ss.str());
+    abort(ss.str().c_str());
   }
 
   // Look for the device type specified in CTEST_KOKKOS_DEVICE_TYPE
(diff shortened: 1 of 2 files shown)


masterleinad:
What problem does this solve?

dalg24:
> What problem does this solve?

`EXPECT_THROW` cannot be used with no exceptions so I am auditing some of our error handling and all use of `EXPECT_[NO_]THROW`.

If anything here, throwing was inconsistent with how we handle similar issues in `Kokkos::initialize` and causing program termination looks like the right thing to do.

core/src/impl/Kokkos_Core.cpp
```diff
@@ -271,7 +271,7 @@ int Kokkos::Impl::get_ctest_gpu(int local_rank) {
     ss << "Error: local rank " << local_rank
        << " is outside the bounds of resource groups provided by CTest. Raised"
        << " by Kokkos::Impl::get_ctest_gpu().";
-    throw_runtime_exception(ss.str());
+    abort(ss.str().c_str());
```
romintomasetti, line 274: `get_ctest_gpu` in `core/src/impl`, seriously ?

romintomasetti: Shouldn't it be some helpers in `core/unit_test` or something like that ?

dalg24: I think you misunderstand what this does.
This is part of the GPU selection logic at initialization.

core/unit_test/TestCTestDevice.cpp
```diff
@@ -76,6 +66,8 @@ void ctest_environment::SetUp() {
   setenv("CTEST_RESOURCE_GROUP_9_GPUS", "id:4,slots:1", 1);
 }
 
+struct ctest_environment_DeathTest : public ctest_environment {};
```
Rombur, line 69: Why do you create a derived class?

dalg24: We have that convention that all death tests must be categorized with a tag ending in `_DeathTest`.
When using `TEST_F`, the name is the test fixture class so I need a strongly typed class that does the same thing as the basic fixture.

core/src/impl/Kokkos_Core.cpp
romintomasetti: I still think that this is very odd.

If someone needs to set a GPU for testing, depending on a `CTest` resource group, they should implement their own logic to map to `KOKKOS_VISIBLE_DEVICES` probably.

Or the logic of `Kokkos::Impl::get_ctest_gpu` could be make available (tweaked a bit to map to `KOKKOS_VISIBLE_DEVICES`) but definitely very awkward to find that in `core` (because it is related to tsting, and because it is bound to `CTest`).

Rombur (approved):

masterleinad (approved):
Fine with me.

core/src/impl/Kokkos_Core.cpp
dalg24: Not the place to argue about this

dalg24:
Ignoring CUDA debug build runtime issue (GPU not detected)

PR #7331, opened 2024-09-12
CI: fold HPX build into default workflow

junghans:
Due to some build failure with Fedora's HPX package in: https://koji.fedoraproject.org/koji/taskinfo?taskID=123213163


junghans:
Another side note @diehlpk figured out that:
```
export TESTS_ARGUMENTS="--hpx:ini=hpx.stacks.use_guard_pages=0"
```
helps with the hpx test.

/CC @hkaiser

masterleinad:
We might just disable the respective since `HPX` throws its own exceptions by using `std::set_new_handler` (or try to catch that exception and throw our own).

hkaiser:
Yes, currently HPX throws its own exceptions which is wrong. See https://github.com/STEllAR-GROUP/hpx/pull/6543 for a possible fix. 

Where can I see the actual runtime error generated by HPX, btw?

masterleinad:
> Where can I see the actual runtime error generated by HPX, btw?

https://github.com/kokkos/kokkos/actions/runs/10834072931/job/30062369318?pr=7331 has
> [ RUN      ] hpx.view_bad_alloc
terminate called after throwing an instance of 'hpx::detail::exception_with_info<hpx::exception>'
  what():  new allocator failed to allocate memory: HPX(out_of_memory)

and it results from https://github.com/kokkos/kokkos/blob/603aa335b898100756ed242ec958493170356e84/core/src/impl/Kokkos_HostSpace.cpp#L78-L85 we are explicitly requesting not to throw an error when allocating.

hkaiser:
> > Where can I see the actual runtime error generated by HPX, btw?
> 
> https://github.com/kokkos/kokkos/actions/runs/10834072931/job/30062369318?pr=7331 has
> 
> > [ RUN      ] hpx.view_bad_alloc
> > terminate called after throwing an instance of 'hpx::detail::exception_with_infohpx::exception'
> > what():  new allocator failed to allocate memory: HPX(out_of_memory)
> 
> and it results from
> 
> https://github.com/kokkos/kokkos/blob/603aa335b898100756ed242ec958493170356e84/core/src/impl/Kokkos_HostSpace.cpp#L78-L85
> 
> we are explicitly requesting not to throw an error when allocating.

Ok, this is something else, then. Why is a handler set by `std::set_new_handler` invoked in this case to begin with?

masterleinad:
> Ok, this is something else, then. Why is a handler set by `std::set_new_handler` invoked in this case to begin with?

It seems https://github.com/llvm/llvm-project/blob/82a36468c74a29b6154639d659550c62457e655b/libcxx/src/new.cpp#L162-L166 is executed but we don't hit the `catch` clause for some reason but abort with an unhandled exception.

hkaiser:
> > Ok, this is something else, then. Why is a handler set by `std::set_new_handler` invoked in this case to begin with?
> 
> It seems https://github.com/llvm/llvm-project/blob/82a36468c74a29b6154639d659550c62457e655b/libcxx/src/new.cpp#L162-L166 is executed but we don't hit the `catch` clause for some reason but abort with an unhandled exception.

Not sure what I can do about this... Would you have any ideas? It shouldn't matter whether the execption thrown from our new-handler is actually derived from `std::bad_alloc` (as it should, see the PR I linked above).

.github/workflows/continuous-integration-workflow.yml
```diff
@@ -73,6 +73,11 @@ jobs:
             cmake_build_type: 'RelWithDebInfo'
             backend: 'THREADS'
             stdcxx: '23'
+          - distro: 'fedora:latest'
+            cxx: 'g++'
+            cmake_build_type: 'RelWithDebInfo'
+            backend: 'HPX'
+            stdcxx: '17'
     runs-on: ubuntu-latest
```
diehlpk, line 81: @masterleinad should that not be fedora-latest as above?

masterleinad: No, we can use containers with other distros from ubuntu-latest.

hkaiser:
While I'm still not sure why this is happening, I can offer adding an option to HPX that allows disabling to set a new-handler (in the same way as one can already disable HPX's signal handling). This option could be used in environments like the Kokkos HPX backend. What do you think?

.github/workflows/continuous-integration-workflow.yml
junghans: `runs-on:` is the GitHub Action base vm, which really doesn't matter much when used with `container:`.

diehlpk (approved):

masterleinad:
Note that we already have a HPX build in our CI that I would expect to reproduce the issue if we update the HPX version there, see https://github.com/kokkos/kokkos/blob/b5509ab14f760769446ab0710dca88483eebce22/.github/workflows/continuous-integration-workflow-hpx.yml#L37-L42.

junghans:
> Note that we already have a HPX build in our CI that I would expect to reproduce the issue if we update the HPX version there, see
> 
> https://github.com/kokkos/kokkos/blob/b5509ab14f760769446ab0710dca88483eebce22/.github/workflows/continuous-integration-workflow-hpx.yml#L37-L42
> 
> .



> Note that we already have a HPX build in our CI that I would expect to reproduce the issue if we update the HPX version there, see
> 
> https://github.com/kokkos/kokkos/blob/b5509ab14f760769446ab0710dca88483eebce22/.github/workflows/continuous-integration-workflow-hpx.yml#L37-L42
> 
> .

Let's see.

junghans:
@masterleinad `github-Linux-hpx` works, but the build with the `hpx-devel` package on Fedora fails. Not sure if it is worth digging, but the fedora build logs are here:
https://koji.fedoraproject.org/koji/buildinfo?buildID=2540054

masterleinad:
> @masterleinad `github-Linux-hpx` works, but the build with the `hpx-devel` package on Fedora fails. Not sure if it is worth digging, but the fedora build logs are here:

Interesting. I can reproduce the issue locally with clang and `libc++`.

junghans:
> > @masterleinad `github-Linux-hpx` works, but the build with the `hpx-devel` package on Fedora fails. Not sure if it is worth digging, but the fedora build logs are here:
> 
> Interesting. I can reproduce the issue locally with clang and `libc++`.

Did you build with:
```
            -DCMAKE_BUILD_TYPE=Debug \
            -DHPX_WITH_UNITY_BUILD=ON \
            -DHPX_WITH_MALLOC=system \
            -DHPX_WITH_NETWORKING=OFF \
            -DHPX_WITH_EXAMPLES=OFF \
            -DHPX_WITH_TESTS=OFF \
```

junghans:
Ok, ping me if there is any patch to add to hpx package on Fedora!

masterleinad:
@diehlpk Can you try to have a look at this?

diehlpk:
@masterleinad Currently,. I have no time but maybe @hkaiser can have a look?

masterleinad:
> Now we are getting:

We just forgot to guard this test properly for deprecated code. Disabling deprecation warnings or warnings as errors (or deleting the test) is a easy workaround to investiagte the original issues here.

hkaiser:
> @hkaiser @diehlpk Now we are getting:
> 
> ```
> [ 26%] Building CXX object core/unit_test/CMakeFiles/Kokkos_CoreUnitTest_HPX.dir/hpx/TestHPX_AtomicOperations_longint.cpp.o
> In file included from /__w/kokkos/kokkos/core/unit_test/hpx/TestHPX_InParallel.cpp:17:
> /__w/kokkos/kokkos/core/unit_test/hpx/TestHPX_InParallel.cpp: In member function 'virtual void {anonymous}::hpx_in_parallel_for_range_policy_Test::TestBody()':
> /__w/kokkos/kokkos/core/unit_test/hpx/TestHPX_InParallel.cpp:31:54: error: 'static bool Kokkos::Experimental::HPX::in_parallel(const Kokkos::Experimental::HPX&)' is deprecated [-Werror=deprecated-declarations]
>    31 |   ASSERT_FALSE(Kokkos::Experimental::HPX::in_parallel());
>       |                ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^~
> ```

We need to ask @msimberg what the alternative way is to avoid using the deprecated functionality.

junghans:
Thanks @masterleinad

msimberg:
> We need to ask @msimberg what the alternative way is to avoid using the deprecated functionality.

I think @masterleinad's commit (https://github.com/kokkos/kokkos/pull/7331/commits/83160448216145313530ea03c6ccf868ea8fed40) is the only right thing to do. `HPX::in_parallel` is deprecated, but the only place where it's used (and where the warning is coming from) is the unit test for `HPX::in_parallel` itself. There's no replacement. So simply disabling the warning until the functionality, and the test, is fully removed makes sense.

junghans:
I am a little confused why the `linux-x64-hpx` test didn't catch that.

junghans:
> > I am a little confused why the `linux-x64-hpx` test didn't catch that.
> 
> That build has `-DKokkos_ENABLE_DEPRECATED_CODE_4=OFF`.

Ah, I think I will kill that job either way.

masterleinad:
```
[ RUN      ] hpx.view_bad_alloc
terminate called after throwing an instance of 'hpx::detail::exception_with_info<hpx::exception>'
  what():  new allocator failed to allocate memory: HPX(out_of_memory)
[ RUN      ] hpx.view_bad_alloc
terminate called after throwing an instance of 'hpx::detail::exception_with_info<hpx::exception>'
  what():  new allocator failed to allocate memory: HPX(out_of_memory)
```

masterleinad (changes requested):
Still failing CI.

junghans:
@hkaiser @diehlpk could you have a look at https://github.com/kokkos/kokkos/actions/runs/13203198061/job/36861088079 ?

.github/workflows/continuous-integration-linux-hpx.yml
```diff
@@ -1,89 +0,0 @@
...
-        with:
-          path:         ./hpx/install
-          key:          kokkos-hwloc-${{ github.ref }}-${{ github.sha }}
-          restore-keys: kokkos-hwloc-${{ github.ref }}
-      - name: configure hpx
-        if: steps.cache-hpx.outputs.cache-hit != 'true'
-        run: |
-          mkdir -p hpx/{build,install}
-          cd hpx/build
-          cmake \
```
junghans, line 48: @hkaiser @diehlpk is there a big difference between the cmake options here and Fedora's (https://src.fedoraproject.org/rpms/hpx/blob/rawhide/f/hpx.spec#_175), that would explain the tests running out of memory?

hkaiser: I don't think it's a real OOM condition. As we established above, the handler set by `std::set_new_handler` is invoked in this case by the library implementation and we don't know why. As promised, I have added an additional option to HPX that will prevent it from setting its OOM handler. At this point, this is available on HPX master only, though. The simplest way for this would be to set an environment variable `HPX_HANDLE_FAILED_NEW=0` while running the test.

hkaiser: RIght, that's what I had in mind. The library implementation should not call the OOM handler from inside a nothow operator new, even if it fails to allocate the memory.

junghans: I have added an export of `HPX_HANDLE_FAILED_NEW=0` to the enviorment.

junghans: @hkaiser do you understand why the test fails here (with Fedora's binary package), but not when hpx gets build?

hkaiser: @junghans could you point me to the failure logs for this, please?

junghans: https://productionresultssa10.blob.core.windows.net/actions-results/38108c36-dd3f-42db-ad34-279740b9d6fc/workflow-job-run-95be4252-462b-55b8-0a70-32019f5f54f9/logs/job/job-logs.txt?rsct=text%2Fplain&se=2025-02-11T18%3A16%3A10Z&sig=bVwQmqhe%2F46aMnR6sgT4k8m5QV6qxZ00JoSbzl7KkKM%3D&ske=2025-02-12T05%3A51%3A10Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2025-02-11T17%3A51%3A10Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-01-05&sp=r&spr=https&sr=b&st=2025-02-11T18%3A06%3A05Z&sv=2025-01-05

It still seems to fail even with `HPX_HANDLE_FAILED_NEW=0`.
```
2025-02-11T16:27:35.2815632Z env:
2025-02-11T16:27:35.2815796Z   HPX_HANDLE_FAILED_NEW: 0
2025-02-11T16:27:35.2816002Z ##[endgroup]
2025-02-11T16:27:35.3445382Z Test project /__w/kokkos/kokkos/builddir
...
2025-02-11T16:31:07.6601088Z [ RUN      ] serial.view_bad_alloc
2025-02-11T16:31:07.6601521Z terminate called after throwing an instance of 'hpx::detail::exception_with_info<hpx::exception>'
2025-02-11T16:31:07.6602251Z   what():  new allocator failed to allocate memory: HPX(out_of_memory)
2025-02-11T16:31:07.6602702Z 
```

junghans: Fixed the link above

hkaiser: > https://github.com/kokkos/kokkos/actions/runs/13266285120/job/37035838978?pr=7331
> 
> It still seems to fail even with `HPX_HANDLE_FAILED_NEW=0`.
> 
> ```
> 2025-02-11T16:27:35.2815632Z env:
> 2025-02-11T16:27:35.2815796Z   HPX_HANDLE_FAILED_NEW: 0
> 2025-02-11T16:27:35.2816002Z ##[endgroup]
> 2025-02-11T16:27:35.3445382Z Test project /__w/kokkos/kokkos/builddir
> ...
> 2025-02-11T16:31:07.6601088Z [ RUN      ] serial.view_bad_alloc
> 2025-02-11T16:31:07.6601521Z terminate called after throwing an instance of 'hpx::detail::exception_with_info<hpx::exception>'
> 2025-02-11T16:31:07.6602251Z   what():  new allocator failed to allocate memory: HPX(out_of_memory)
> 2025-02-11T16:31:07.6602702Z 
> ```

Does this use HPX master?

junghans: The current CI uses v1.9.0 (not master):
https://github.com/kokkos/kokkos/blob/1d8e45be635678aef23b30bcad01f341c4f4f631/.github/workflows/continuous-integration-linux-hpx.yml#L35
so this is a bug in v1.10.0 only?

hkaiser: > The current CI uses v1.9.0 (not master):
> 
> https://github.com/kokkos/kokkos/blob/1d8e45be635678aef23b30bcad01f341c4f4f631/.github/workflows/continuous-integration-linux-hpx.yml#L35
> 
> 
> so this is a bug in v1.10.0 only?

Now I'm confused. We started setting the OOM handler a long time ago, long before v1.9. Does it work for that version?

hkaiser: Here is a patch against v1.10:

```diff
diff --git a/libs/core/runtime_configuration/src/runtime_configuration.cpp b/libs/core/runtime_configuration/src/runtime_configuration.cpp
index 230df545614..21bea962d41 100644
--- a/libs/core/runtime_configuration/src/runtime_configuration.cpp
+++ b/libs/core/runtime_configuration/src/runtime_configuration.cpp
@@ -187,6 +187,8 @@ namespace hpx::util {
 #if !defined(HPX_WINDOWS)
             "handle_signals = ${HPX_HANDLE_SIGNALS:1}",
 #endif
+            "handle_failed_new = ${HPX_HANDLE_FAILED_NEW:1}",
+
             // arity for collective operations implemented in a tree fashion
             "[hpx.lcos.collectives]",
             "arity = ${HPX_LCOS_COLLECTIVES_ARITY:32}",
diff --git a/libs/core/runtime_local/src/runtime_local.cpp b/libs/core/runtime_local/src/runtime_local.cpp
index 526d9dd2cac..1bd5e4f792e 100644
--- a/libs/core/runtime_local/src/runtime_local.cpp
+++ b/libs/core/runtime_local/src/runtime_local.cpp
@@ -303,7 +303,10 @@ namespace hpx {
         }
 #endif
 
-        std::set_new_handler(hpx::new_handler);
+        if (hpx::util::get_entry_as<int>(cfg, "hpx.handle_failed_new", 1))
+        {
+            std::set_new_handler(hpx::new_handler);
+        }
     }
 
     ///////////////////////////////////////////////////////////////////////////
```

junghans: > > The current CI uses v1.9.0 (not master):
> > https://github.com/kokkos/kokkos/blob/1d8e45be635678aef23b30bcad01f341c4f4f631/.github/workflows/continuous-integration-linux-hpx.yml#L35
> > 
> > so this is a bug in v1.10.0 only?
> 
> Now I'm confused. We started setting the OOM handler a long time ago, long before v1.9. Does it work for that version?

Yes, currently the CI builds out hpx-1.9.0 and there the tests pass, but if we instead use Fedora's HPX package (v1.10.0) the tests fail. There could be a slight difference in how hpx was built, but I am not sure about that.

junghans: @hkaiser compare current CI setup (hpx-1.9.0) https://github.com/kokkos/kokkos/blob/1d8e45be635678aef23b30bcad01f341c4f4f631/.github/workflows/continuous-integration-linux-hpx.yml#L48 with Fedora;s package (hpx-1.10.0) https://src.fedoraproject.org/rpms/hpx/blob/rawhide/f/hpx.spec#_177.

junghans: Alternatively in a week we should have the patched hpx-1.10.0 in Fedora: https://bodhi.fedoraproject.org/updates/FEDORA-2025-7ac77a9436

junghans:
Update: Good news, the patched hpx-1.10.0 from Fedora and `HPX_HANDLE_FAILED_NEW=0` make the test pass.

Now we have 3 options:

- Use `fedora:rawhide` for the hpx - might be pretty unstable
- Merge https://github.com/kokkos/ci-containers/pull/63 (and revert it at some later stage once hpx-1.10.0-6 makes it into Fedora)
- Wait for https://bodhi.fedoraproject.org/updates/FEDORA-2025-7ac77a9436, which might be a week or so.

.github/workflows/continuous-integration-linux-hpx.yml
hkaiser: Thanks for making this happen!

masterleinad:
> Now we have 3 options:
> - Use `fedora:rawhide` for the hpx - might be pretty unstable
> - Merge https://github.com/kokkos/ci-containers/pull/63 (and revert it in the next clean up)
> - Wait for https://bodhi.fedoraproject.org/updates/FEDORA-2025-7ac77a9436, which might be a week or so.

I think we can just wait a week.

.github/workflows/continuous-integration-linux.yml
```diff
@@ -125,6 +132,10 @@ jobs:
             echo "flang-new found"
             echo "FC=flang-new" >> $GITHUB_ENV
           fi
+# workaround for handle issue in hpx-1.10.0+, remove once fixed
```
dalg24, line 135: You could use `FIXME_HPX`

.github/workflows/continuous-integration-linux.yml
```diff
@@ -89,6 +89,12 @@ jobs:
             stdcxx: '17'
             cxx_extra_flags: '-m32'
             cmake_extra_opts: '-DKokkos_ENABLE_HWLOC=OFF'
+          - distro: 'fedora:latest'
+            cxx: 'g++'
+            cmake_build_type: 'Debug'
```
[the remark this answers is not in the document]
dalg24: Sure

dalg24 (commented):
Fine by me

.github/workflows/continuous-integration-linux.yml
junghans: like this?
```suggestion
# FIXME_HPX: workaround for handle issue in hpx-1.10.0+
```

dalg24: Exactly

hkaiser: Just to clarify - this is not an HPX issue and the workaround is for the standard library implementation that invokes the OOM handler in case a nothrow new fails to allocate memory (even if it shouldn't do that, at least as far as I read the standard - see: https://eel.is/c++draft/alloc.errors#new.handler).

junghans: I will add that in 6 days once the updated hpx package is in our contailner.

junghans: > Just to clarify - this is not an HPX issue and the workaround is for the standard library implementation that invokes the OOM handler in case a nothrow new fails to allocate memory (even if it shouldn't do that, at least as far as I read the standard - see: https://eel.is/c++draft/alloc.errors#new.handler).

Can you propose a better comment?

hkaiser: > > Just to clarify - this is not an HPX issue and the workaround is for the standard library implementation that invokes the OOM handler in case a nothrow new fails to allocate memory (even if it shouldn't do that, at least as far as I read the standard - see: https://eel.is/c++draft/alloc.errors#new.handler).
> 
> Can you propose a better comment?

Perhaps:
```
# workaround for standard library calling OOM handler for failing nothrow new, remove once fixed
```
?

junghans, line 135:
```suggestion
# FIXME_HPX: workaround for standard library calling OOM handler for failing nothrow new, remove once fixed
```

junghans:
The patched hpx made it into Fedora, so this is ready to be merged.

masterleinad (approved):
Fine with me.

msimberg (approved):
Looks good to me.

This is not a blocker, but I want to note that the kokkos+hpx build took 2.5 hours on the latest workflow run (https://github.com/kokkos/kokkos/actions/runs/13357950343/job/37303340997). I don't remember it taking that long in the past and a semi-random build from a few months ago took a bit over an hour (https://github.com/kokkos/kokkos/actions/runs/12170179713/job/33945205886, without ccache as far as I can tell, and same number of threads for the build (2)). From a couple of days ago on master it's 1 hour 50 minutes (https://github.com/kokkos/kokkos/actions/runs/13340947907/job/37265301428). This could be apples to oranges if kokkos has had a lot of tests added recently, or something else in the setup has changed, but HPX may also have regressed in compile times (@hkaiser?).

junghans:
@msimberg thanks for the analysis, I noticed something similar but didn't do the historic analysis.
Could it have something to do with `Debug` vs `RelWithDebInfo`? For the Debug the build is faster, but the tests are slower and vice versa.

hkaiser:
> HPX may also have regressed in compile times (@hkaiser?).

I'm not aware of this.

msimberg:
> @msimberg thanks for the analysis, I noticed something similar but didn't do the historic analysis. Could it have something to do with `Debug` vs `RelWithDebInfo`? For the Debug the build is faster, but the tests are slower and vice versa.

Yeah, that's definitely possible. I vaguely seem to remember something similar from when I was testing things a long time ago.

> > HPX may also have regressed in compile times (@hkaiser?).
> 
> I'm not aware of this.

Ok, good to know.

In any case, as said, not a blocker. I just wanted to highlight it as something to be aware of and maybe keep an eye on.

junghans:
Ok. then let's merge this and I will do another pass to speed up the CI afterwards. One obvious things would be `continue-on-error: false`, but that is a discussion for another day.

PR #7337, opened 2024-09-13
Medley of (hopefully) uncontroversial fixups towards building with no exceptions

dalg24:
Partial fix for #7037

### DIFF
--- a/algorithms/unit_tests/TestBinSortA.hpp
+++ b/algorithms/unit_tests/TestBinSortA.hpp
@@ -247,10 +247,10 @@ TEST(TEST_CATEGORY, BinSortEmptyView) {
   Kokkos::View<int*, ExecutionSpace> v("v", 0);
 
   // test all exposed public sort methods
-  ASSERT_NO_THROW(Sorter.sort(ExecutionSpace(), v, 0, 0));
-  ASSERT_NO_THROW(Sorter.sort(v, 0, 0));
-  ASSERT_NO_THROW(Sorter.sort(ExecutionSpace(), v));
-  ASSERT_NO_THROW(Sorter.sort(v));
+  Sorter.sort(ExecutionSpace(), v, 0, 0);
+  Sorter.sort(v, 0, 0);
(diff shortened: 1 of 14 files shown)


algorithms/unit_tests/TestBinSortA.hpp
```diff
@@ -247,10 +247,10 @@ TEST(TEST_CATEGORY, BinSortEmptyView) {
...
 
   // test all exposed public sort methods
-  ASSERT_NO_THROW(Sorter.sort(ExecutionSpace(), v, 0, 0));
-  ASSERT_NO_THROW(Sorter.sort(v, 0, 0));
-  ASSERT_NO_THROW(Sorter.sort(ExecutionSpace(), v));
-  ASSERT_NO_THROW(Sorter.sort(v));
+  Sorter.sort(ExecutionSpace(), v, 0, 0);
+  Sorter.sort(v, 0, 0);
+  Sorter.sort(ExecutionSpace(), v);
+  Sorter.sort(v);
```
Rombur, line 253: Add a comment that we are just checking that these functions don't throw. It's pretty obvious what these tests do when you see `ASSERT_NO_THROW` but once you remove the macro it's not clear anymore. In 6 months, we will wonder why these tests don't check anything.

crtrott (approved):
It's fine with me, I don't think any of the places changed throwing is particularly useful in the first place.

algorithms/unit_tests/TestBinSortA.hpp
dalg24: Done

Rombur (approved):

romintomasetti (approved):
I would suggest that you state somewhere explicitly that for `noexcept` support, tests should not be using `ASSERT_NO_THROW` and `EXPECT_NO_THROW` anymore.

Writing it in the description of the PR would already help (for the future anyone wanting to understand why we forbid these 2 test macros).

dalg24:
Ignoring runtime failures (no gpu available for execution) in OpenACC build.  Everything else passed.

PR #7340, opened 2024-09-13
DynamicView: diagnose incompatible deep_copy() at compile time

dalg24:
My motivation for changing that code is related to support for no exceptions #7037
I needed the `ASSERT_THROW` to go and when looking into this I got somewhat surprised that we deferred to runtime to diagnose something we could tell at compile time.

I am not familiar with `DynamicView` (which I plan to propose to deprecate in a follow up) and could not find any usage in Trilinos, ArborX, Cabana, nor LAMMPS.

### DIFF
--- a/containers/src/Kokkos_DynamicView.hpp
+++ b/containers/src/Kokkos_DynamicView.hpp
@@ -835,21 +835,17 @@ inline void deep_copy(const View<T, DP...>& dst,
   using dst_execution_space = typename ViewTraits<T, DP...>::execution_space;
   using src_memory_space    = typename ViewTraits<T, SP...>::memory_space;
 
-  enum {
-    DstExecCanAccessSrc =
-        Kokkos::SpaceAccessibility<dst_execution_space,
-                                   src_memory_space>::accessible
-  };
+  constexpr bool DstExecCanAccessSrc =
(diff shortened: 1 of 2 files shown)


romintomasetti (approved):
LGTM.

Also agreeing on the fact that this container seems forgotten in the dustbin of history for quite some time. It's not even very clear from the documentation why you'd use it.

In fact, it's not clear what it brings compared to `Kokkos::View`. Maybe the "user-provided chunk size for memory allocation" is the only thing that was brought by it (https://github.com/kokkos/kokkos-core-wiki/blob/7a627e9671d47d24b2ad992cf1aeb6ed452a7dc5/docs/source/API/containers/DynamicView.rst?plain=1#L70).

dalg24:
Unrelated failure in CUDA-DRC build (GPU not detected at runtime)

masterleinad (approved):

rppawlo:
This was written for empire. It is the core data structure for our particle push. If you deprecate it, I will need to take ownership and move it into empire. It's a very important piece of code to us.

crtrott:
@rppawlo Yes understood. That was the thinking at the moment. We only know of Empire as user at this point.

PR #7341, opened 2024-09-13
OpenMP: Add OpenMP link flag for Cray compiler.

rgayatri23:
The PR adds `fopenmp` flag for the OpenMP builds with Cray-Clang compiler. 
Else the build fails to find OpenMP symbols at linktime.


cmake/kokkos_arch.cmake
```diff
@@ -802,6 +802,14 @@ IF(KOKKOS_ARCH_INTEL_PVC)
   CHECK_MULTIPLE_INTEL_ARCH()
 ENDIF()
 
+IF (KOKKOS_ENABLE_OPENMP)
+    IF(KOKKOS_CLANG_IS_CRAY)
+      COMPILER_SPECIFIC_LINK_OPTIONS(
+        Clang -fopenmp
+      )
+    ENDIF()
```
dalg24, line 810:
```suggestion
  IF(KOKKOS_CLANG_IS_CRAY)
    COMPILER_SPECIFIC_LINK_OPTIONS(
      Clang -fopenmp
    )
  ENDIF()
```

dalg24 (commented):
What version of CMake and of the Cray compiler were you using?
Should we consider this a defect in CMake's FindOpenMP module or in the compiler identification?
It looks like since CMake 3.28 it is able to differentiate the Clang-based Cray compiler https://cmake.org/cmake/help/v3.28/variable/CMAKE_LANG_COMPILER_ID.html

rgayatri23:
> What version of CMake and of the Cray compiler were you using? Should we consider this a defect in CMake's FindOpenMP module or in the compiler identification? It looks like since CMake 3.28 it is able to differentiate the Clang-based Cray compiler https://cmake.org/cmake/help/v3.28/variable/CMAKE_LANG_COMPILER_ID.html

I tried cmake/3.24. I can install 3.28 later and test again.

dalg24:
> I can install 3.28 later and test again.

I am fine with your patch but I would like confirmation that 3.28 works

rgayatri23:
I tested this with cmake/3.30.3 (the latest version) and still face the same issue w/o the patch
 ```bash
ld.lld: error: undefined symbol: omp_set_num_threads
>>> referenced by Kokkos_OpenMP_Instance.cpp
>>>               Kokkos_OpenMP_Instance.cpp.o:(Kokkos::Impl::OpenMPInternal::initialize(int)) in archive ../src/libkokkoscore.a

ld.lld: error: undefined symbol: omp_get_thread_num
>>> referenced by Kokkos_OpenMP_Instance.cpp
>>>               Kokkos_OpenMP_Instance.cpp.o:(_ZN6Kokkos4Impl14OpenMPInternal17clear_thread_dataEv_cray$mt$p0005) in archive ../src/libkokkoscore.a
clang++: error: linker command failed with exit code 1 (use -v to see invocation)
make[2]: *** [core/unit_test/CMakeFiles/Kokkos_CoreTestCompileOnly.dir/build.make:354: core/unit_test/Kokkos_CoreTestCompileOnly] Error 1
make[1]: *** [CMakeFiles/Makefile2:1440: core/unit_test/CMakeFiles/Kokkos_CoreTestCompileOnly.dir/all] Error 2
make: *** [Makefile:146: all] Error 2
rgayatri@perlmutter:login23:/pscratch/sd/r/rgayatri/Kokkos/rgayatri_kk/build_cray> cmake --version
cmake version 3.30.3

```

**Comment by dalg24** [comment]
Wait a minute I expect your patch would not work because now the compiler id would be CrayClang, not Clang.

**Comment by rgayatri23** [comment]
> Wait a minute I expect your patch would not work because now the compiler id would be CrayClang, not Clang.

Ah ok. Well it does not work with cmake>=3.28. And neither does replacing `Clang` with `CrayClang` (which is what `${CMAKE_CXX_COMPILER_ID}` returns. Need to look for a different solution in that case.

**Comment by crtrott** [comment]
What do you mean replacing Clang with CrayClang: for CMAKE > 3.28 can you just try to fix the Kokkos identification, i.e. have KOKKOS_CLANG_IS_CRAY defined then too? And then together with your patch does it work?

File: cmake/kokkos_arch.cmake [hunk 2]
```diff
@@ -802,6 +802,14 @@ IF(KOKKOS_ARCH_INTEL_PVC)
   CHECK_MULTIPLE_INTEL_ARCH()
 ENDIF()
 
+IF (KOKKOS_ENABLE_OPENMP)
+  IF(KOKKOS_CLANG_IS_CRAY)
+    COMPILER_SPECIFIC_LINK_OPTIONS(
+      Clang -fopenmp
```
dalg24, line 808: Did you try this?
```suggestion
      Clang -fopenmp
      CrayClang -fopenmp
```

rgayatri23: It looks like [this](https://github.com/kokkos/kokkos/blob/develop/cmake/kokkos_compiler_id.cmake#L96) condition is not satisfied for `CrayClang` with cmake>3.28. So we might have to fix Kokkos identification as @crtrott suggested.

masterleinad: You will likely also need to adjust https://github.com/kokkos/kokkos/blob/af8df138190527f4a99b632da0a58faec188e8ba/cmake/kokkos_functions.cmake#L849.

rgayatri23:
I edited the patch to accept add the OpenMP flag even for `CrayClang` whenever available.

cmake/kokkos_arch.cmake
```diff
@@ -805,7 +805,8 @@ ENDIF()
 IF (KOKKOS_ENABLE_OPENMP)
   IF(KOKKOS_CLANG_IS_CRAY)
     COMPILER_SPECIFIC_LINK_OPTIONS(
-      Clang -fopenmp
+        CrayClang -fopenmp # Needed for cmake>=3.28
+        Clang -fopenmp
```
masterleinad, line 809: The alternative would be
```suggestion
        Default -fopenmp
```
since we have already established that we have a Cray compiler.

masterleinad (commented):
I have a slight preference of eliminating `KOKKOS_CLANG_IS_CRAY` and setting `KOKKOS_CXX_COMPILER_ID` to `CrayClang` when we set the variable previously.

cmake/kokkos_arch.cmake
rgayatri23: Thats fine by me if there is no objection. I can add a comment with the details.

rgayatri23:
> I have a slight preference of eliminating `KOKKOS_CLANG_IS_CRAY` and setting `KOKKOS_CXX_COMPILER_ID` to `CrayClang` when we set the variable previously.

I have no preference on the topic. I can make the change if there is no objection.

cmake/kokkos_arch.cmake
crtrott: I mean its somewhat weird since its inside a CLANG_IS_CRAY thing. I think I prefer the more verbose one. Makes the intent clearer.

crtrott:
> > I have a slight preference of eliminating `KOKKOS_CLANG_IS_CRAY` and setting `KOKKOS_CXX_COMPILER_ID` to `CrayClang` when we set the variable previously.
> 
> I have no preference on the topic. I can make the change if there is no objection.

Not 100% sure what is better. I guess we could? @dalg24 can you weigh in?

cedricchevalier19:
I prefer @masterleinad's solution; it looks more like regular CMake, and we do not need to introduce a new variable.

crtrott:
> I prefer @masterleinad's solution; it looks more like regular CMake, and we do not need to introduce a new variable.

There is no new variable, and the variable is still necessary. What Daniel is suggesting is the following:

```
IF (KOKKOS_ENABLE_OPENMP)
  IF(KOKKOS_CLANG_IS_CRAY)
    COMPILER_SPECIFIC_LINK_OPTIONS(
        Default -fopenmp
     )
  endif()
endif()
```

So my point was that its a bit weird to say default there, because it only can be reached if the compiler identifier is `Clang` but its really a Cray compiler which isn't identified as `CrayClang` because the CMake is too old or whatever, or the compiler identifier is `CrayClang`. And we only do that because the standard OpenMP support of CMake fails with Cray, so we need to manually add the linker flag for OpenMP.

So my point was that using default there obfuscates what's going on. Sure you can read the conditions carefully and understand that you only reach that "Default" statement if you compiler is clang variant from Cray, but still.

crtrott:
Regarding the idea of resetting `KOKKOS_CXX_COMPILER_ID` I am hesitant because I think this would be the first time where `KOKKOS_CXX_COMPILER_ID` is NOT equal to `CMAKE_CXX_COMPILER_ID`. And I am not sure if that causes issues somewhere, in particular later when building something against Kokkos??

crtrott (approved):
I believe that the current state is the right approach - though I don't want to cut off discussion yet on the "default" thing and the resetting `KOKKOS_CXX_COMPILER_ID`

masterleinad:
> What Daniel is suggesting is the following:

That's slightly orthogonal. I think what Cedric meant is that I suggested to do
```
IF(<DEDUCED_THAT_CLANG_COMPILER_IS_CRAY>)
  SET(KOKKOS_CXX_COMPILER_ID CrayClang)
ENDIF()
```
and remove KOKKOS_CLANG_IS_CRAY. That's similar to what we did for `PGI`/`NVHPC`.

crtrott:
Ah did we do that for PGI/NVHPC before? Ok if that worked we could do that. Can someone check that we don't ever use KOKKOS_CXX_COMPILER_ID to compare to CMAKE_CXX_COMPILER_ID?

masterleinad:
> Ah did we do that for PGI/NVHPC before? Ok if that worked we could do that. Can someone check that we don't ever use KOKKOS_CXX_COMPILER_ID to compare to CMAKE_CXX_COMPILER_ID?

> $ git grep -n "KOKKOS_CXX_COMPILER_ID" | grep "CMAKE_CXX_COMPILER_ID"
cmake/kokkos_compiler_id.cmake:4:SET(KOKKOS_CXX_COMPILER_ID ${CMAKE_CXX_COMPILER_ID})

See https://github.com/kokkos/kokkos/pull/6038.

crtrott (changes requested):
Ok looks like preference is for KOKKOS_CLANG_IS_CRAY to die, and rather always set KOKKOS_CXX_COMPILER_ID to CrayClang: @rgayatri23 can you do that change?

rgayatri23:
I edited the patch to set `KOKKOS_CXX_COMPILER_ID` to `CrayClang` irrespective of cmake version. So the OpenMP flag can be set only for one compiler-id.

masterleinad (approved):
Im fine with removing `KOKKOS_CLANG_IS_CRAY` later. 
There some unfortunate CMake formatting changes but I hope that we can just have an automatic tool for that later anyway.

rgayatri23:
The OpenMPTarget build failed due to the known issue of "No GPU found". 
Unrelated to this PR but needs to be fixed. 
Will look into the issue.

crtrott (approved):

dalg24:
OpenMPTarget failed at runtime (GPU not detected)

dalg24:
Make sure you include this in the change log

PR #7343, opened 2024-09-15
OpenACC backend update to better support host

seyonglee:
Update the OpenACC backend to better support the host as the offloading target.
        - MDRangePolicy applies different iteration-to-thread mapping strategies for the CPU target and the GPU target.
        - TeamPolicy applies different iteration-to-thread mapping strategies for the CPU target and the GPU target.
        - Different shared allocation specializations are used for the CPU target and the GPU target.
        - Apply the Kokkos chunk_size only if it is bigger than or equal to the target device warp size.

### DIFF
--- a/core/src/OpenACC/Kokkos_OpenACC_ParallelReduce_Team.hpp
+++ b/core/src/OpenACC/Kokkos_OpenACC_ParallelReduce_Team.hpp
@@ -151,8 +151,18 @@ KOKKOS_INLINE_FUNCTION void parallel_reduce(
 #define KOKKOS_IMPL_ACC_REDUCE_TEAM_PRAGMA \
   vector vector_length(team_size* vector_length)
 #define KOKKOS_IMPL_ACC_REDUCE_TEAM_ITRS league_size* team_size* vector_length
-#define KOKKOS_IMPL_ACC_REDUCE_TEAM_LEAGUE_ID_INIT \
-  i / (team_size * vector_length)
+#ifdef KOKKOS_ENABLE_OPENACC_FORCE_HOST_AS_DEVICE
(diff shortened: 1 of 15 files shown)


core/src/OpenACC/Kokkos_OpenACC_ParallelReduce_Team.hpp
```diff
@@ -258,7 +268,10 @@ KOKKOS_INLINE_FUNCTION void parallel_reduce(
 #define KOKKOS_IMPL_ACC_REDUCE_TEAM_PRAGMA \
   num_workers(team_size) vector_length(vector_length)
 #define KOKKOS_IMPL_ACC_REDUCE_TEAM_ITRS league_size
-#define KOKKOS_IMPL_ACC_REDUCE_TEAM_LEAGUE_ID_INIT i
+#define KOKKOS_IMPL_ACC_REDUCE_TEAM_MEMBER_INIT                          \
```
cwpearson, line 271: I think this macro is only use in one place (or I missed it?)

cwpearson (changes requested):

core/src/OpenACC/Kokkos_OpenACC_ParallelReduce_Team.hpp
seyonglee: The macro `KOKKOS_IMPL_ACC_REDUCE_TEAM_MEMBER_INIT` is defined in multiple places (in line 155, 161, and 271), but used only in one place (in line 405).

crtrott:
Is there a way to split this up. It's somewhat hard to review at >1000 lines change with 15 files touched.

seyonglee:
> Is there a way to split this up. It's somewhat hard to review at >1000 lines change with 15 files touched.

I will split this PR into multiple sub PRs and submit one by one.
Close this PR.

PR #7345, opened 2024-09-16
View-Refactor: Add ReferenceCountedAccessor/DataHandle

crtrott:
This PR adds a `ReferenceCountedDataHandle` and `ReferenceCountedAccessor`. The data handle packages the pre-existing `SharedAllocationTracker` and a pointer. Note that pointer can be something else than the base pointer of the allocation to support subviews/submdspan. One tricky question is the convertibility of these things. But we can revisit that when View comes in and everyone can see which current tests would fail because conversions are allowed. In either case I am not planning to make these types public for now. I.e. they are meant as implementation details for View. The tests aren't as exhaustive as they could be yet, but cover the basics. A lot of this will be implicitly tested as well when the View replacement is in. 

One helper function added here (`make_shared_allocation_record`) returns a pointer to a record, this is what the view tracker actually consumes. Note that after that record is passed on, ownership is actually transferred, so you can't do it again. We may wanna consider rewriting this - but we have thought that about a lot of the implementation of SharedAlloc. For now this function fits into the existing scheme.

### DIFF
--- a/core/src/View/Kokkos_ViewAlloc.hpp
+++ b/core/src/View/Kokkos_ViewAlloc.hpp
@@ -26,6 +26,7 @@ static_assert(false,
 #include <cstring>
 #include <type_traits>
 #include <string>
+#include <optional>
 
 #include <impl/Kokkos_Tools.hpp>
 #include <Kokkos_Core_fwd.hpp>
@@ -246,9 +247,9 @@ template <class ElementType, class MemorySpace, class ExecutionSpace,
           bool Initialize, bool SequentialInit>
 Kokkos::Impl::SharedAllocationRecord<void, void>* make_shared_allocation_record(
     const size_t& required_span_size, std::string_view label,
-    const MemorySpace& memory_space, const ExecutionSpace* exec_space,
-    std::integral_constant<bool, Initialize>,
-    std::integral_constant<bool, SequentialInit>) {
+    const MemorySpace& memory_space,
+    const std::optional<ExecutionSpace> exec_space,
+    std::bool_constant<Initialize>, std::bool_constant<SequentialInit>) {
   static_assert(SpaceAccessibility<ExecutionSpace, MemorySpace>::accessible);
 
   // Use this for constructing and destroying the view
@@ -260,13 +261,14 @@ Kokkos::Impl::SharedAllocationRecord<void, void>* make_shared_allocation_record(
   using record_type =
       Kokkos::Impl::SharedAllocationRecord<MemorySpace, functor_type>;
 
-  static constexpr std::size_t align_mask   = 0x7;
-  static constexpr std::size_t element_size = sizeof(ElementType);
+  /* Force alignment of allocations on on 8 byte boundaries even for
+   * element types smaller than 8 bytes */
+  static constexpr std::size_t align_mask = 0x7;
 
   // Calculate the total size of the memory, in bytes, and make sure it is
   // byte-aligned
   const std::size_t alloc_size =
-      (required_span_size * element_size + align_mask) & ~align_mask;
+      (required_span_size * sizeof(ElementType) + align_mask) & ~align_mask;
 
   auto* record =
       exec_space
--- a/core/src/View/Kokkos_ViewAlloc.hpp
+++ b/core/src/View/Kokkos_ViewAlloc.hpp
@@ -242,6 +242,70 @@ struct ViewValueFunctorSequentialHostInit {
   }
 };
 
+template <class ElementType, class MemorySpace, class ExecutionSpace,
+          bool AllowPadding, bool Initialize, bool SequentialInit>
+Kokkos::Impl::SharedAllocationRecord<void, void>* make_shared_allocation_record(
+    const size_t& required_span_size, std::string_view label,
+    const MemorySpace& memory_space, const ExecutionSpace* exec_space,
+    [[maybe_unused]] std::integral_constant<bool, AllowPadding> allow_padding,
+    [[maybe_unused]] std::integral_constant<bool, Initialize> initialize,
+    [[maybe_unused]] std::integral_constant<bool, SequentialInit>
+        sequential_init) {
+  static_assert(SpaceAccessibility<ExecutionSpace, MemorySpace>::accessible);
+
+  // Use this for constructing and destroying the view
+  using device_type  = Kokkos::Device<ExecutionSpace, MemorySpace>;
+  using functor_type = std::conditional_t<
+      SequentialInit,
+      ViewValueFunctorSequentialHostInit<device_type, ElementType>,
+      ViewValueFunctor<device_type, ElementType>>;
+  using record_type =
+      Kokkos::Impl::SharedAllocationRecord<MemorySpace, functor_type>;
+
+  // If padding is allowed then pass in sizeof value type
+  // for padding computation.
+  using padding =
+      std::integral_constant<std::size_t,
+                             AllowPadding ? sizeof(ElementType) : 0>;
+
+  static constexpr std::size_t align_mask   = 0x7;
+  static constexpr std::size_t element_size = sizeof(ElementType);
+
+  // Calculate the total size of the memory, in bytes, and make sure it is
+  // byte-aligned
+  const std::size_t alloc_size =
+      (required_span_size * element_size + align_mask) & ~align_mask;
+
+  auto* record =
+      exec_space
+          ? record_type::allocate(*exec_space, memory_space, std::string{label},
+                                  alloc_size)
+          : record_type::allocate(memory_space, std::string{label}, alloc_size);
+
+  auto ptr = static_cast<ElementType*>(record->data());
+
+  auto functor =
+      exec_space ? functor_type(*exec_space, ptr, required_span_size,
+                                std::string{label})
+                 : functor_type(ptr, required_span_size, std::string{label});
+
+  //  Only initialize if the allocation is non-zero.
+  //  May be zero if one of the dimensions is zero.
+  if constexpr (Initialize) {
+    if (alloc_size) {
+      // Assume destruction is only required when construction is requested.
+      // The ViewValueFunctor has both value construction and destruction
+      // operators.
+      record->m_destroy = std::move(functor);
+
+      // Construct values
+      record->m_destroy.construct_shared_allocation();
+    }
+  }
+
+  return record;
+}
+
 }  // namespace Kokkos::Impl
 
 #endif  // KOKKOS_VIEW_ALLOC_HPP
--- a/core/src/View/Kokkos_ViewAlloc.hpp
+++ b/core/src/View/Kokkos_ViewAlloc.hpp
@@ -247,10 +247,9 @@ template <class ElementType, class MemorySpace, class ExecutionSpace,
 Kokkos::Impl::SharedAllocationRecord<void, void>* make_shared_allocation_record(
     const size_t& required_span_size, std::string_view label,
     const MemorySpace& memory_space, const ExecutionSpace* exec_space,
-    [[maybe_unused]] std::integral_constant<bool, AllowPadding> allow_padding,
-    [[maybe_unused]] std::integral_constant<bool, Initialize> initialize,
-    [[maybe_unused]] std::integral_constant<bool, SequentialInit>
-        sequential_init) {
+    std::integral_constant<bool, AllowPadding>,
+    std::integral_constant<bool, Initialize>,
+    std::integral_constant<bool, SequentialInit>) {
   static_assert(SpaceAccessibility<ExecutionSpace, MemorySpace>::accessible);
 
   // Use this for constructing and destroying the view
--- a/core/src/View/Kokkos_ViewAlloc.hpp
+++ b/core/src/View/Kokkos_ViewAlloc.hpp
@@ -243,11 +243,10 @@ struct ViewValueFunctorSequentialHostInit {
 };
 
 template <class ElementType, class MemorySpace, class ExecutionSpace,
-          bool AllowPadding, bool Initialize, bool SequentialInit>
+          bool Initialize, bool SequentialInit>
 Kokkos::Impl::SharedAllocationRecord<void, void>* make_shared_allocation_record(
     const size_t& required_span_size, std::string_view label,
     const MemorySpace& memory_space, const ExecutionSpace* exec_space,
-    std::integral_constant<bool, AllowPadding>,
     std::integral_constant<bool, Initialize>,
     std::integral_constant<bool, SequentialInit>) {
   static_assert(SpaceAccessibility<ExecutionSpace, MemorySpace>::accessible);
--- a/core/src/View/Kokkos_ViewAlloc.hpp
+++ b/core/src/View/Kokkos_ViewAlloc.hpp
@@ -262,12 +262,6 @@ Kokkos::Impl::SharedAllocationRecord<void, void>* make_shared_allocation_record(
   using record_type =
       Kokkos::Impl::SharedAllocationRecord<MemorySpace, functor_type>;
 
-  // If padding is allowed then pass in sizeof value type
-  // for padding computation.
-  using padding =
-      std::integral_constant<std::size_t,
-                             AllowPadding ? sizeof(ElementType) : 0>;
-
   static constexpr std::size_t align_mask   = 0x7;
   static constexpr std::size_t element_size = sizeof(ElementType);
 
--- a/core/src/View/MDSpan/Kokkos_MDSpan_Accessor.hpp
+++ b/core/src/View/MDSpan/Kokkos_MDSpan_Accessor.hpp
@@ -259,16 +259,6 @@ class ReferenceCountedDataHandle {
       const ReferenceCountedDataHandle<OtherElementType, memory_space>& other)
       : m_tracker(other.m_tracker), m_handle(other.m_handle) {}
 
-  KOKKOS_DEFAULTED_FUNCTION
-  ReferenceCountedDataHandle(const ReferenceCountedDataHandle&) = default;
-  KOKKOS_DEFAULTED_FUNCTION
-  ReferenceCountedDataHandle(ReferenceCountedDataHandle&&) noexcept = default;
-  KOKKOS_DEFAULTED_FUNCTION
-  ReferenceCountedDataHandle& operator=(const ReferenceCountedDataHandle&) =
-      default;
-  KOKKOS_DEFAULTED_FUNCTION
-  ReferenceCountedDataHandle& operator=(ReferenceCountedDataHandle&&) = default;
-
   KOKKOS_FUNCTION
   pointer get() const noexcept { return m_handle; }
   KOKKOS_FUNCTION
@@ -337,16 +327,6 @@ class ReferenceCountedDataHandle<ElementType, AnonymousSpace> {
   KOKKOS_FUNCTION ReferenceCountedDataHandle(
       const ReferenceCountedDataHandle& other, OtherElementType* ptr)
       : m_tracker(other.m_tracker), m_handle(ptr) {}
-
-  KOKKOS_DEFAULTED_FUNCTION
-  ReferenceCountedDataHandle(const ReferenceCountedDataHandle&) = default;
-  KOKKOS_DEFAULTED_FUNCTION
-  ReferenceCountedDataHandle(ReferenceCountedDataHandle&&) noexcept = default;
-  KOKKOS_DEFAULTED_FUNCTION
-  ReferenceCountedDataHandle& operator=(const ReferenceCountedDataHandle&) =
-      default;
-  KOKKOS_DEFAULTED_FUNCTION
-  ReferenceCountedDataHandle& operator=(ReferenceCountedDataHandle&&) = default;
 
   template <class OtherElementType, class OtherSpace,
             class = std::enable_if_t<std::is_convertible_v<
--- a/core/src/View/MDSpan/Kokkos_MDSpan_Accessor.hpp
+++ b/core/src/View/MDSpan/Kokkos_MDSpan_Accessor.hpp
@@ -212,6 +212,334 @@ struct AtomicAccessorRelaxed {
   }
 };
 
+//=====================================================================
+//============= Reference Counted Accessor and DataHandle =============
+//=====================================================================
+
+template <class ElementType, class MemorySpace>
+class ReferenceCountedDataHandle {
+ public:
+  using value_type   = ElementType;
+  using pointer      = value_type*;
+  using reference    = value_type&;
+  using memory_space = MemorySpace;
+
+  KOKKOS_DEFAULTED_FUNCTION
+  ReferenceCountedDataHandle() = default;
+  explicit ReferenceCountedDataHandle(SharedAllocationRecord<void, void>* rec) {
+    m_tracker.assign_allocated_record_to_uninitialized(rec);
+    m_handle = static_cast<pointer>(get_record()->data());
+  }
+
+  KOKKOS_FUNCTION
+  ReferenceCountedDataHandle(const SharedAllocationTracker& tracker,
+                             pointer data_handle)
+      : m_tracker(tracker), m_handle(data_handle) {}
+
+  // unmanaged ctor
+  template <class OtherElementType,
+            class = std::enable_if_t<std::is_convertible_v<
+                OtherElementType (*)[], value_type (*)[]>>>
+  KOKKOS_FUNCTION ReferenceCountedDataHandle(OtherElementType* ptr)
+      : m_tracker(), m_handle(ptr) {}
+
+  // subview ctor
+  template <class OtherElementType,
+            class = std::enable_if_t<std::is_convertible_v<
+                OtherElementType (*)[], value_type (*)[]>>>
+  KOKKOS_FUNCTION ReferenceCountedDataHandle(
+      const ReferenceCountedDataHandle& other, OtherElementType* ptr)
+      : m_tracker(other.m_tracker), m_handle(ptr) {}
+
+  // converting ctor
+  template <class OtherElementType,
+            class = std::enable_if_t<std::is_convertible_v<
+                OtherElementType (*)[], value_type (*)[]>>>
+  KOKKOS_FUNCTION ReferenceCountedDataHandle(
+      const ReferenceCountedDataHandle<OtherElementType, memory_space>& other)
+      : m_tracker(other.m_tracker), m_handle(other.m_handle) {}
+
+  KOKKOS_DEFAULTED_FUNCTION
+  ReferenceCountedDataHandle(const ReferenceCountedDataHandle&) = default;
+  KOKKOS_DEFAULTED_FUNCTION
+  ReferenceCountedDataHandle(ReferenceCountedDataHandle&&) noexcept = default;
+  KOKKOS_DEFAULTED_FUNCTION
+  ReferenceCountedDataHandle& operator=(const ReferenceCountedDataHandle&) =
+      default;
+  KOKKOS_DEFAULTED_FUNCTION
+  ReferenceCountedDataHandle& operator=(ReferenceCountedDataHandle&&) = default;
+
+  KOKKOS_FUNCTION
+  pointer get() const noexcept { return m_handle; }
+  KOKKOS_FUNCTION
+  explicit operator pointer() const noexcept { return m_handle; }
+
+  bool has_record() const { return m_tracker.has_record(); }
+  auto* get_record() const { return m_tracker.get_record<memory_space>(); }
+  int use_count() const noexcept { return m_tracker.use_count(); }
+
+  std::string get_label() const { return m_tracker.get_label<memory_space>(); }
+  KOKKOS_FUNCTION
+  const SharedAllocationTracker& tracker() const noexcept { return m_tracker; }
+
+  KOKKOS_FUNCTION
+  friend bool operator==(const ReferenceCountedDataHandle& lhs,
+                         const value_type* rhs) {
+    return lhs.m_handle == rhs;
+  }
+
+  KOKKOS_FUNCTION
+  friend bool operator==(const value_type* lhs,
+                         const ReferenceCountedDataHandle& rhs) {
+    return lhs == rhs.m_handle;
+  }
+
+ private:
+  template <class OtherElementType, class OtherSpace>
+  friend class ReferenceCountedDataHandle;
+
+  template <class OtherElementType, class OtherSpace, class NestedAccessor>
+  friend class ReferenceCountedAccessor;
+
+  SharedAllocationTracker m_tracker;
+  pointer m_handle = nullptr;
+};
+
+template <class ElementType>
+class ReferenceCountedDataHandle<ElementType, AnonymousSpace> {
+ public:
+  using value_type   = ElementType;
+  using pointer      = value_type*;
+  using reference    = value_type&;
+  using memory_space = AnonymousSpace;
+
+  KOKKOS_DEFAULTED_FUNCTION
+  ReferenceCountedDataHandle() = default;
+  explicit ReferenceCountedDataHandle(SharedAllocationRecord<void, void>* rec) {
+    m_tracker.assign_allocated_record_to_uninitialized(rec);
+    m_handle = static_cast<pointer>(get_record()->data());
+  }
+
+  KOKKOS_FUNCTION
+  ReferenceCountedDataHandle(const SharedAllocationTracker& tracker,
+                             pointer data_handle)
+      : m_tracker(tracker), m_handle(data_handle) {}
+
+  template <class OtherElementType,
+            class = std::enable_if_t<std::is_convertible_v<
+                OtherElementType (*)[], value_type (*)[]>>>
+  KOKKOS_FUNCTION ReferenceCountedDataHandle(OtherElementType* ptr)
+      : m_tracker(), m_handle(ptr) {}
+
+  template <class OtherElementType,
+            class = std::enable_if_t<std::is_convertible_v<
+                OtherElementType (*)[], value_type (*)[]>>>
+  KOKKOS_FUNCTION ReferenceCountedDataHandle(
+      const ReferenceCountedDataHandle& other, OtherElementType* ptr)
+      : m_tracker(other.m_tracker), m_handle(ptr) {}
+
+  KOKKOS_DEFAULTED_FUNCTION
+  ReferenceCountedDataHandle(const ReferenceCountedDataHandle&) = default;
+  KOKKOS_DEFAULTED_FUNCTION
+  ReferenceCountedDataHandle(ReferenceCountedDataHandle&&) noexcept = default;
+  KOKKOS_DEFAULTED_FUNCTION
+  ReferenceCountedDataHandle& operator=(const ReferenceCountedDataHandle&) =
+      default;
+  KOKKOS_DEFAULTED_FUNCTION
+  ReferenceCountedDataHandle& operator=(ReferenceCountedDataHandle&&) = default;
+
+  template <class OtherElementType, class OtherSpace,
+            class = std::enable_if_t<std::is_convertible_v<
+                OtherElementType (*)[], value_type (*)[]>>>
+  KOKKOS_FUNCTION ReferenceCountedDataHandle(
+      const ReferenceCountedDataHandle<OtherElementType, OtherSpace>& other)
+      : m_tracker(other.m_tracker), m_handle(other.m_handle) {}
+
+  KOKKOS_FUNCTION
+  pointer get() const noexcept { return m_handle; }
+  KOKKOS_FUNCTION
+  explicit operator pointer() const noexcept { return m_handle; }
+
+  bool has_record() const { return m_tracker.has_record(); }
+  auto* get_record() const { return m_tracker.get_record<memory_space>(); }
+  int use_count() const noexcept { return m_tracker.use_count(); }
+
+  std::string get_label() const { return m_tracker.get_label<memory_space>(); }
+  KOKKOS_FUNCTION
+  const SharedAllocationTracker& tracker() const noexcept { return m_tracker; }
+
+  KOKKOS_FUNCTION
+  friend bool operator==(const ReferenceCountedDataHandle& lhs,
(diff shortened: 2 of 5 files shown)


romintomasetti:
Just wondering if such an important addition should deserve a small PR description ?

core/unit_test/view/TestReferenceCountedDataHandle.hpp
```diff
@@ -0,0 +1,131 @@
...
+  using pointer      = value_type*;
+  using reference    = value_type&;
+  using memory_space = MemorySpace;
+  ...
+}
+*/
+
+namespace {
+using element_t = float;
+using mem_t     = typename TEST_EXECSPACE::memory_space;
```
romintomasetti, line 35: Noting that `mem_t` is "unconventional".

core/src/View/MDSpan/Kokkos_MDSpan_Accessor.hpp
```diff
@@ -212,6 +212,334 @@ struct AtomicAccessorRelaxed {
...
+      : m_tracker(other.m_tracker), m_handle(other.m_handle) {}
+
+  KOKKOS_DEFAULTED_FUNCTION
+  ReferenceCountedDataHandle(const ReferenceCountedDataHandle&) = default;
+  KOKKOS_DEFAULTED_FUNCTION
+  ReferenceCountedDataHandle(ReferenceCountedDataHandle&&) noexcept = default;
+  KOKKOS_DEFAULTED_FUNCTION
+  ReferenceCountedDataHandle& operator=(const ReferenceCountedDataHandle&) =
+      default;
+  KOKKOS_DEFAULTED_FUNCTION
```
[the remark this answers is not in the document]
romintomasetti: If so, I'd remove it.

core/src/View/MDSpan/Kokkos_MDSpan_Accessor.hpp
```diff
@@ -212,6 +212,334 @@ struct AtomicAccessorRelaxed {
...
+
+  // unmanaged ctor
+  template <class OtherElementType,
+            class = std::enable_if_t<std::is_convertible_v<
+                OtherElementType (*)[], value_type (*)[]>>>
+  KOKKOS_FUNCTION ReferenceCountedDataHandle(OtherElementType* ptr)
+      : m_tracker(), m_handle(ptr) {}
+
+  // subview ctor
+  template <class OtherElementType,
```
romintomasetti, line 247: Is *subview* really appropriate ?

core/src/View/MDSpan/Kokkos_MDSpan_Accessor.hpp
```diff
@@ -212,6 +212,334 @@ struct AtomicAccessorRelaxed {
...
+  KOKKOS_FUNCTION
+  pointer get() const noexcept { return m_handle; }
+  KOKKOS_FUNCTION
+  explicit operator pointer() const noexcept { return m_handle; }
+
+  bool has_record() const { return m_tracker.has_record(); }
+  auto* get_record() const { return m_tracker.get_record<memory_space>(); }
+  int use_count() const noexcept { return m_tracker.use_count(); }
+
+  std::string get_label() const { return m_tracker.get_label<memory_space>(); }
```
romintomasetti, line 281: What about this ?
```suggestion
  const std::string& get_label() const { return m_tracker.get_label<memory_space>(); }
```

core/src/View/MDSpan/Kokkos_MDSpan_Accessor.hpp
```diff
@@ -212,6 +212,334 @@ struct AtomicAccessorRelaxed {
...
+
+  template <class OtherElementType, class OtherSpace, class NestedAccessor>
+  friend class ReferenceCountedAccessor;
+
+  SharedAllocationTracker m_tracker;
+  pointer m_handle = nullptr;
+};
+
+template <class ElementType>
+class ReferenceCountedDataHandle<ElementType, AnonymousSpace> {
```
romintomasetti, line 309: It's not clear what gets specialized. I'd recommend to either add some comment on the specialized things, or if only a very tiny subset of things are specialized, probably you could do some SFINAE stuff in the primary class declaration ?

core/unit_test/view/TestReferenceCountedDataHandle.hpp
```diff
@@ -0,0 +1,131 @@
...
+using data_handle_t =
+    Kokkos::Impl::ReferenceCountedDataHandle<element_t, mem_t>;
+using const_data_handle_t =
+    Kokkos::Impl::ReferenceCountedDataHandle<const element_t, mem_t>;
+;
+}  // namespace
+
+TEST(TEST_CATEGORY, RefCountedDataHandle_Typedefs) {
+  Kokkos::parallel_for(
+      Kokkos::RangePolicy<TEST_EXECSPACE>(0, 1), KOKKOS_LAMBDA(int) {
```
romintomasetti, line 45: Do you really need to write those checks within a parallel region ? Sounds awkward.

core/unit_test/view/TestReferenceCountedDataHandle.hpp
```diff
@@ -0,0 +1,131 @@
...
+            std::is_same_v<typename data_handle_t::pointer, element_t*>);
+        static_assert(
+            std::is_same_v<typename data_handle_t::reference, element_t&>);
+        static_assert(
+            std::is_same_v<typename data_handle_t::memory_space, mem_t>);
+      });
+}
+
+TEST(TEST_CATEGORY, RefCountedDataHandle) {
+  auto shared_alloc =
```
romintomasetti, line 58: For such tests, I would have designed it with a test suite setup that makes the shared allocation record, and that would allow me to split the tests in smaller bits that are more readable and on which it's easier to reason about.

core/unit_test/view/TestReferenceCountedDataHandle.hpp
crtrott: Yes and no. In principle it is possible for typedefs to differ inside Device code and Host code. I mean it would be extremely stupid to do that but it is possible. This was we test for it.

core/unit_test/view/TestReferenceCountedDataHandle.hpp
crtrott: I thought about it, but it's still a lot of duplicated code. Essentially what it comes down to is that the only way state of RefCountedDataHandle changes is through constructors. So it all comes down to: did the constructor do what we expected it to do. Furthermore for almost anything I need to start with an allocation, so might as well use this one ...

core/src/View/MDSpan/Kokkos_MDSpan_Accessor.hpp
crtrott: I tried that but it was somewhat messy. The specialization piece has to do with convertibility i.e. it's all in the constructors. I personally felt that the specialization tack was cleaner.

crtrott: I will add some comments.

core/unit_test/view/TestReferenceCountedDataHandle.hpp
romintomasetti: You could probably initialize your allocation once (drafter pseudo-code):
```c++
class TestSuite
{
void setup() { if (!data) make share alloc;}
static data = nullptr;
}
```

core/src/View/MDSpan/Kokkos_MDSpan_Accessor.hpp
romintomasetti: Speaking of code duplication...

core/src/View/MDSpan/Kokkos_MDSpan_Accessor.hpp
crtrott: In some ways yes, in some ways no. Since we don't publicly expose this class it's generally meant as an implementation detail of View (hence subview). But mechanically it can and is used inside submdspan. So I could change that comment.

core/src/View/MDSpan/Kokkos_MDSpan_Accessor.hpp
romintomasetti: OK, so the debate might probably move to why `Kokkos` makes new strings each time we ask for the label ? I guess that some allocation shared records store a `char*` to be used on device, so why not returning a string view ?

core/src/View/MDSpan/Kokkos_MDSpan_Accessor.hpp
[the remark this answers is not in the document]
crtrott: It is currently a no-op but it also isn't deprecated yet is it? So generally we should add that unless we decide that we want the macro gone generally. It didn't use to be a no-op (CUDA 10 needed special treatment here which was different from how Clang CUDA wanted stuff to happen, or HIP).

core/src/View/MDSpan/Kokkos_MDSpan_Accessor.hpp
romintomasetti: Noting that view labels are usually quite long.

crtrott:
> Just wondering if such an important addition should deserve a small PR description ?

Sure ;-)

core/src/View/MDSpan/Kokkos_MDSpan_Accessor.hpp
crtrott: The string might actually come from device storage copied over into a temporary. Generally, labels are a debug functionality so we were never super concerned about making them fast.

core/unit_test/view/TestReferenceCountedDataHandle.hpp
crtrott: Actually that doesn't work. When you hand the thing to the data handle you pass over ownership. So it will delete the allocation when it goes out of scope.

crtrott: (or when the last one with a ref count goes out of scope)

core/unit_test/view/TestReferenceCountedDataHandle.hpp
crtrott: I should add that its likely an ODR violation if one were to do that ...

core/unit_test/view/TestReferenceCountedDataHandle.hpp
romintomasetti: Because you are initializing `Kokkos` before `GTest` in `core/unit_test/UnitTestMainInit.cpp`:
```c++
class TestSuite : public ::testing::Test
{
public:
    static void SetUpTestSuite() {
        if( ! shared_alloc) shared_alloc = Kokkos::Impl::make_shared_allocation_record<...>(...);
    }
    static void TearDownTestSuite() { /* anything to do here ? */}
protected:
    Kokkos::Impl::SharedAllocationRecord<MemorySpace, functor_type>* shared_alloc = nullptr;
};
```

core/unit_test/view/TestReferenceCountedDataHandle.hpp
crtrott: https://godbolt.org/z/fn7erb77v

its somewhat contrived obviously, but on the other hand the way we end up with the typedefs is complicated enough that I don't think its impossible to end up with weird stuff.

core/src/View/MDSpan/Kokkos_MDSpan_Accessor.hpp
romintomasetti: OK, never mind.

core/src/View/MDSpan/Kokkos_MDSpan_Accessor.hpp
romintomasetti: I still think that it's bad to specialize the whole class if you have a single difference (the conversion constructor).

core/unit_test/view/TestReferenceCountedDataHandle.hpp
romintomasetti: OK, never mind.

core/unit_test/view/TestReferenceCountedDataHandle.hpp
romintomasetti: And if you should do anything at all, at least do (enhances test readability):
```c++
auto* shared_alloc = ...
```

core/src/View/MDSpan/Kokkos_MDSpan_Accessor.hpp
crtrott: If I remember correct there was some sort of issue with a combination of ambiguity and round-trip via conversion to and construction from `default_accessor` where we couldn't figure out how to prevent conversions we did NOT want to allow doing it in a single class. I can try again and see if I can figure it out.

core/unit_test/view/TestReferenceCountedDataHandle.hpp
crtrott: Yeah it returns a pointer to a SharedAllocationRecord, but as soon as I pass that guy to the data_handle constructor the ownership of the allocation the record points to transfers. I.e. I can't pass the same pointer I get back to more than one data_handle thing. And that would break the suggestion above wouldn't it?

masterleinad:
```
C:\projects\source\core\unit_test\view/TestReferenceCountedAccessor.hpp(35,32): error C2039: 'default_accessor': is not a member of 'Kokkos' [C:\projects\source\build\core\unit_test\Kokkos_CoreUnitTest_Serial_ViewSupport.vcxproj]
C:\projects\source\core\src\Kokkos_AcquireUniqueTokenImpl.hpp(27): message : see declaration of 'Kokkos' [C:\projects\source\build\core\unit_test\Kokkos_CoreUnitTest_Serial_ViewSupport.vcxproj]
C:\projects\source\core\unit_test\view/TestReferenceCountedAccessor.hpp(35,48): error C2061: syntax error: identifier 'default_accessor' [C:\projects\source\build\core\unit_test\Kokkos_CoreUnitTest_Serial_ViewSupport.vcxproj]
C:\projects\source\core\unit_test\view/TestReferenceCountedAccessor.hpp(36,32): error C2039: 'default_accessor': is not a member of 'Kokkos' [C:\projects\source\build\core\unit_test\Kokkos_CoreUnitTest_Serial_ViewSupport.vcxproj]
C:\projects\source\core\src\Kokkos_AcquireUniqueTokenImpl.hpp(27): message : see declaration of 'Kokkos' [C:\projects\source\build\core\unit_test\Kokkos_CoreUnitTest_Serial_ViewSupport.vcxproj]
C:\projects\source\core\unit_test\view/TestReferenceCountedAccessor.hpp(36,48): error C2061: syntax error: identifier 'default_accessor' [C:\projects\source\build\core\unit_test\Kokkos_CoreUnitTest_Serial_ViewSupport.vcxproj]
C:\projects\source\core\unit_test\view/TestReferenceCountedAccessor.hpp(38,19): error C2039: 'ReferenceCountedAccessor': is not a member of 'Kokkos::Impl' [C:
```

crtrott:
> ```
> C:\projects\source\core\unit_test\view/TestReferenceCountedAccessor.hpp(35,32): error C2039: 'default_accessor': is not a member of 'Kokkos' [C:\projects\source\build\core\unit_test\Kokkos_CoreUnitTest_Serial_ViewSupport.vcxproj]
> C:\projects\source\core\src\Kokkos_AcquireUniqueTokenImpl.hpp(27): message : see declaration of 'Kokkos' [C:\projects\source\build\core\unit_test\Kokkos_CoreUnitTest_Serial_ViewSupport.vcxproj]
> C:\projects\source\core\unit_test\view/TestReferenceCountedAccessor.hpp(35,48): error C2061: syntax error: identifier 'default_accessor' [C:\projects\source\build\core\unit_test\Kokkos_CoreUnitTest_Serial_ViewSupport.vcxproj]
> C:\projects\source\core\unit_test\view/TestReferenceCountedAccessor.hpp(36,32): error C2039: 'default_accessor': is not a member of 'Kokkos' [C:\projects\source\build\core\unit_test\Kokkos_CoreUnitTest_Serial_ViewSupport.vcxproj]
> C:\projects\source\core\src\Kokkos_AcquireUniqueTokenImpl.hpp(27): message : see declaration of 'Kokkos' [C:\projects\source\build\core\unit_test\Kokkos_CoreUnitTest_Serial_ViewSupport.vcxproj]
> C:\projects\source\core\unit_test\view/TestReferenceCountedAccessor.hpp(36,48): error C2061: syntax error: identifier 'default_accessor' [C:\projects\source\build\core\unit_test\Kokkos_CoreUnitTest_Serial_ViewSupport.vcxproj]
> C:\projects\source\core\unit_test\view/TestReferenceCountedAccessor.hpp(38,19): error C2039: 'ReferenceCountedAccessor': is not a member of 'Kokkos::Impl' [C:
> ```

Teah need to protect the entire thing with MDSpan being enabled I believe

core/unit_test/view/TestReferenceCountedAccessor.hpp
```diff
@@ -0,0 +1,125 @@
...
+  const_defacc_t c_defacc;
+  acc_t acc(defacc);
+  const_acc_t c_acc1(defacc);
+  const_acc_t c_acc2(c_defacc);
+  (void)acc;
+  (void)c_acc1;
+  (void)c_acc2;
+  static_assert(!std::is_constructible_v<acc_t, const_defacc_t>);
+}
+});
```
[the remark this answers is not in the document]
crtrott: Yeah me too. Really weird.

core/src/View/MDSpan/Kokkos_MDSpan_Accessor.hpp
crtrott: I will try again and if I reencounter the real issue report it in an understandable way.

core/src/View/MDSpan/Kokkos_MDSpan_Accessor.hpp
crtrott: I will update the description of this guy and everything else a bit better (i.e. make more comments)

core/src/View/MDSpan/Kokkos_MDSpan_Accessor.hpp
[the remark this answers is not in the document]
crtrott: As Daniel said we probably can just delete them.

core/unit_test/view/TestReferenceCountedDataHandle.hpp
crtrott: `memory_space_t`?

core/src/View/MDSpan/Kokkos_MDSpan_Accessor.hpp
nmm0: I don't think making the SFINAE on the converting constructor significantly more complex is easier to read than two specializations. If y'all want two specializations we should inherit from a common type that provides most of the functionality. But given how trivial the repeated code is I don't know if it's worth it.

There's also the issue that @crtrott mentioned with round trip conversions but we wrote this code a month or so ago so I don't quite remember what it was. I can look into it.

core/unit_test/view/TestReferenceCountedDataHandle.hpp
dalg24: > In principle it is possible for typedefs to differ inside Device code and Host code

The code snippet you posted is highly unrealistic/artificial.  We are testing a Kokkos class template that may be templated on user-defined types but I fail to see for what naughty definition this test would fail (unless they specialize `std::is_same_v` in which case a/ I don't think we should care b/ your test would also break)

masterleinad:
```
/__w/kokkos/kokkos/core/unit_test/view/TestReferenceCountedDataHandle.hpp:171:20: error: unused variable 'c_ptr1' [-Werror=unused-variable]
  171 |   const element_t* c_ptr1 = ptr1;
      |                    ^~~~~~
/__w/kokkos/kokkos/core/unit_test/view/TestReferenceCountedDataHandle.hpp:187:20: error: unused variable 'c_ptr2' [-Werror=unused-variable]
  187 |   const element_t* c_ptr2 = ptr2;
      |                    ^~~~~~
```

core/src/View/Kokkos_ViewAlloc.hpp
```diff
@@ -242,6 +242,64 @@ struct ViewValueFunctorSequentialHostInit {
...
 
+template <class ElementType, class MemorySpace, class ExecutionSpace,
+          bool AllowPadding, bool Initialize, bool SequentialInit>
+Kokkos::Impl::SharedAllocationRecord<void, void>* make_shared_allocation_record(
+    const size_t& required_span_size, std::string_view label,
+    const MemorySpace& memory_space, const ExecutionSpace* exec_space,
+    [[maybe_unused]] std::integral_constant<bool, AllowPadding> allow_padding,
+    [[maybe_unused]] std::integral_constant<bool, Initialize> initialize,
+    [[maybe_unused]] std::integral_constant<bool, SequentialInit>
+        sequential_init) {
```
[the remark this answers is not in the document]
masterleinad: ping. Also older compiler struggle with `maybe_unused` as function argument.

core/unit_test/view/TestReferenceCountedDataHandle.hpp
```diff
@@ -0,0 +1,234 @@
...
+template <class ElementType, class MemorySpace>
+class ReferenceCountedDataHandle {
+ public:
+  using value_type   = ElementType;
+  using pointer      = value_type*;
+  using reference    = value_type&;
+  using memory_space = MemorySpace;
+  ...
+}
+*/
```
masterleinad, line 31: Any reason to keep this here?
```suggestion
```

core/unit_test/view/TestReferenceCountedAccessor.hpp
```diff
@@ -0,0 +1,171 @@
...
+
+void test_refcountedacc_ctors() {
+  Kokkos::parallel_for(Kokkos::RangePolicy<TEST_EXECSPACE>(0, 1), KOKKOS_LAMBDA(int) {
+      // default ctor and non-const to const
+      {
+        acc_t acc;
+        const_acc_t c_acc(acc);
+	static_assert(!std::is_constructible_v<acc_t, const_acc_t>);
+
+	unused_variable_sink(c_acc);
```
masterleinad, line 76: Why does `clang-format` not complain about the tabs used here?

core/unit_test/view/TestReferenceCountedAccessor.hpp
```diff
@@ -0,0 +1,171 @@
...
+/*
+template <class ElementType, class MemorySpace, class NestedAccessor>
+class ReferenceCountedAccessor {
+ public:
+  using element_type     = ElementType;
+  using data_handle_type = ReferenceCountedDataHandle<ElementType, MemorySpace>;
+  using reference        = typename NestedAccessor::reference;
+  using offset_policy    = ReferenceCountedAccessor;
+}
+*/
```
masterleinad, line 30: Is that really helpful here?
```suggestion
```

core/src/View/Kokkos_ViewAlloc.hpp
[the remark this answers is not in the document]
dalg24: I am also not liking the maybe_unused attribute here

core/src/View/Kokkos_ViewAlloc.hpp
```diff
@@ -242,6 +242,63 @@ struct ViewValueFunctorSequentialHostInit {
   }
 };
 
+template <class ElementType, class MemorySpace, class ExecutionSpace,
+          bool AllowPadding, bool Initialize, bool SequentialInit>
+Kokkos::Impl::SharedAllocationRecord<void, void>* make_shared_allocation_record(
+    const size_t& required_span_size, std::string_view label,
+    const MemorySpace& memory_space, const ExecutionSpace* exec_space,
+    std::integral_constant<bool, AllowPadding>,
```
masterleinad, line 250: This template parameter appears to be unused. Ar you intending to do something with it in the future?

core/unit_test/view/TestReferenceCountedAccessor.hpp
```diff
@@ -0,0 +1,156 @@
...
+  acc_t acc(defacc);
+  const_acc_t c_acc1(defacc);
+  const_acc_t c_acc2(c_defacc);
+  static_assert(!std::is_constructible_v<acc_t, const_defacc_t>);
+
+  unused_variable_sink(acc);
+  unused_variable_sink(c_acc1);
+  unused_variable_sink(c_acc2);
+}
+});
```
nmm0, line 76:
```suggestion
  Kokkos::parallel_for(Kokkos::RangePolicy<TEST_EXECSPACE>(0, 1), KOKKOS_LAMBDA(int) {
    // default ctor and non-const to const
    {
      acc_t acc;
      const_acc_t c_acc(acc);
      static_assert(!std::is_constructible_v<acc_t, const_acc_t>);

      unused_variable_sink(c_acc);
    }
    // from default_accessor
    {
      defacc_t defacc;
      const_defacc_t c_defacc;
      acc_t acc(defacc);
      const_acc_t c_acc1(defacc);
      const_acc_t c_acc2(c_defacc);
      static_assert(!std::is_constructible_v<acc_t, const_defacc_t>);

      unused_variable_sink(acc);
      unused_variable_sink(c_acc1);
      unused_variable_sink(c_acc2);
    }
  });
```

I think i got it right, no idea why clang-format wasn't complaining about this

core/src/View/MDSpan/Kokkos_MDSpan_Accessor.hpp
```diff
@@ -212,6 +212,197 @@ struct AtomicAccessorRelaxed {
...
+      const ReferenceCountedAccessor<OtherElementType, MemorySpace,
+                                     OtherNestedAccessor>&) {}
+
+  template <
+      class OtherElementType, class OtherSpace, class OtherNestedAccessor,
+      class = std::enable_if_t<
+          std::is_convertible_v<OtherElementType (*)[], element_type (*)[]> &&
+          (std::is_same_v<OtherSpace, AnonymousSpace> ||
+           std::is_same_v<memory_space, AnonymousSpace>)&&std::
+              is_constructible_v<NestedAccessor, OtherNestedAccessor>>>
```
nmm0, line 348: More weird formatting, idk why clang-format is having so many problems

core/unit_test/view/TestReferenceCountedAccessor.hpp
crtrott: It's actually worse than that. It undoes these changes again. At least it did on kokkos-dev-2

core/src/View/Kokkos_ViewAlloc.hpp
crtrott: Yeah we don't really need it here since we just pass the required_span_size in. That already includes padding if any.

masterleinad (approved):

core/unit_test/view/TestReferenceCountedDataHandle.hpp
```diff
@@ -0,0 +1,212 @@
...
+}
+
+TEST(TEST_CATEGORY, RefCountedDataHandle_Typedefs) {
+  test_ref_counted_data_handle_typedefs();
+}
+
+template <class DataHandleType, class ConstDataHandleType>
+void test_ref_counted_data_handle() {
+  auto shared_alloc =
+      Kokkos::Impl::make_shared_allocation_record<element_t, mem_t,
```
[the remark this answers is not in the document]
crtrott: For what? The data handle? Or the test? I can't create the data handle with this thing. The entire test is for implementation details only anyway.

core/unit_test/view/TestReferenceCountedDataHandle.hpp
```diff
@@ -0,0 +1,212 @@
...
+void test_ref_counted_data_handle_typedefs() {
+  static_assert(std::is_same_v<typename data_handle_t::value_type, element_t>);
+  static_assert(std::is_same_v<typename data_handle_t::pointer, element_t*>);
+  static_assert(std::is_same_v<typename data_handle_t::reference, element_t&>);
+  static_assert(std::is_same_v<typename data_handle_t::memory_space, mem_t>);
+}
+
+TEST(TEST_CATEGORY, RefCountedDataHandle_Typedefs) {
+  test_ref_counted_data_handle_typedefs();
+}
```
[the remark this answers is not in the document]
crtrott: It used to be in a parallel for so the function call was necessary. I guess I can change that.

core/unit_test/CMakeLists.txt
```diff
@@ -382,6 +382,24 @@ foreach(Tag Threads;Serial;OpenMP;Cuda;HPX;OpenMPTarget;OpenACC;HIP;SYCL)
       list(APPEND ${Tag}_SOURCES2D ${file})
     endforeach()
 
+    set(${Tag}_VIEWSUPPORT)
+    IF (Kokkos_ENABLE_IMPL_MDSPAN)
+      foreach(Name
+        ReferenceCountedAccessor
+        ReferenceCountedDataHandle
```
romintomasetti, line 389: Let's make it clear:
```cmake
set(${Tag}_VIEWSUPPORT ReferenceCountedAccessor ReferenceCountedDataHandle)
list(TRANSFORM ${Tag}_VIEWSUPPORT PREPEND ${dir}/Test${Tag}_View_)
list(TRANSFORM ${Tag}_VIEWSUPPORT APPEND  .cpp)
```

core/unit_test/CMakeLists.txt
```diff
@@ -706,6 +724,12 @@ if(Kokkos_ENABLE_SERIAL)
     UnitTestMainInit.cpp
     ${Serial_SOURCES2}
```
romintomasetti, line 725: Why don't you list the supported backends and do:
```cmake
set(ViewSupport_backends SERIAL OPENMP ...)
foreach(backend)
    if(support backend)
        add_test
    endif()
endfunction()
```

core/unit_test/view/TestReferenceCountedDataHandle.hpp
```diff
@@ -0,0 +1,212 @@
...
+    Kokkos::Impl::ReferenceCountedDataHandle<const element_t,
+                                             Kokkos::AnonymousSpace>;
+
+}  // namespace
+
+void test_ref_counted_data_handle_typedefs() {
+  static_assert(std::is_same_v<typename data_handle_t::value_type, element_t>);
+  static_assert(std::is_same_v<typename data_handle_t::pointer, element_t*>);
+  static_assert(std::is_same_v<typename data_handle_t::reference, element_t&>);
+  static_assert(std::is_same_v<typename data_handle_t::memory_space, mem_t>);
```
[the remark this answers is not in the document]
crtrott: done

core/unit_test/view/TestReferenceCountedDataHandle.hpp
```diff
@@ -0,0 +1,212 @@
...
+}
+
+template <class DataHandleType, class ConstDataHandleType>
+void test_ref_counted_data_handle() {
+  auto shared_alloc =
+      Kokkos::Impl::make_shared_allocation_record<element_t, mem_t,
+                                                  TEST_EXECSPACE>(
+          100, "Test", mem_t(), nullptr,
+          std::integral_constant<bool, true>(),    // init
+          std::integral_constant<bool, false>());  // sequential_host_init
```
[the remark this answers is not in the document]
crtrott: done

core/src/View/MDSpan/Kokkos_MDSpan_Accessor.hpp
```diff
@@ -212,6 +212,197 @@ struct AtomicAccessorRelaxed {
...
+template <class ElementType, class MemorySpace,
+          class MemoryScope = desul::MemoryScopeDevice>
+using checked_atomic_accessor_relaxed =
+    SpaceAwareAccessor<MemorySpace, AtomicAccessorRelaxed<ElementType>>;
+
+template <class ElementType, class MemorySpace,
+          class MemoryScope = desul::MemoryScopeDevice>
+using checked_reference_counted_atomic_accessor_relaxed = SpaceAwareAccessor<
+    MemorySpace, ReferenceCountedAccessor<ElementType, MemorySpace,
+                                          AtomicAccessorRelaxed<ElementType>>>;
```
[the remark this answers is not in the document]
crtrott: removed

core/src/View/MDSpan/Kokkos_MDSpan_Accessor.hpp
```diff
@@ -212,6 +212,197 @@ struct AtomicAccessorRelaxed {
...
+template <class ElementType, class MemorySpace, class NestedAccessor>
+class ReferenceCountedAccessor;
+
+template <class Accessor>
+struct IsReferenceCountedAccessorImpl : std::false_type {};
+
+template <class ElementType, class MemorySpace, class NestedAccessor>
+struct IsReferenceCountedAccessorImpl<
+    ReferenceCountedAccessor<ElementType, MemorySpace, NestedAccessor>>
+    : std::true_type {};
```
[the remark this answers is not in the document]
crtrott: done

core/src/View/MDSpan/Kokkos_MDSpan_Accessor.hpp
```diff
@@ -212,6 +212,197 @@ struct AtomicAccessorRelaxed {
...
+  KOKKOS_FUNCTION constexpr ReferenceCountedAccessor(
+      const default_accessor<OtherElementType>&) {}
+
+  template <class DstAccessor,
+            typename = std::enable_if_t<
+                !IsReferenceCountedAccessorImpl<DstAccessor>::value &&
+                std::is_convertible_v<NestedAccessor, DstAccessor>>>
+  KOKKOS_FUNCTION operator DstAccessor() const {
+    return m_nested_acc;
+  }
```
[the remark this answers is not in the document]
crtrott: Its operator DstAccessor.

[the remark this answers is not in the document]
dalg24: I missed that it was a conversion **operator**

core/unit_test/CMakeLists.txt
crtrott: I would prefer keeping it the same way we do the other tests here. We can then subsequently change them all to this more concise way of expressing stuff.

core/unit_test/CMakeLists.txt
crtrott: I moved it right up where we set the source files. The only thing is that the original separation (The way other tests are doing it) was there to do backend specific exclusions without having all the backend specific conditions in the initial list of tests.

core/src/View/Kokkos_ViewAlloc.hpp
```diff
@@ -242,6 +242,62 @@ struct ViewValueFunctorSequentialHostInit {
   }
 };
 
+template <class ElementType, class MemorySpace, class ExecutionSpace,
+          bool Initialize, bool SequentialInit>
+Kokkos::Impl::SharedAllocationRecord<void, void>* make_shared_allocation_record(
+    const size_t& required_span_size, std::string_view label,
+    const MemorySpace& memory_space, const ExecutionSpace* exec_space,
```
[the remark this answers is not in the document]
crtrott: done

core/src/View/Kokkos_ViewAlloc.hpp
```diff
@@ -242,6 +242,62 @@ struct ViewValueFunctorSequentialHostInit {
...
+  using device_type  = Kokkos::Device<ExecutionSpace, MemorySpace>;
+  using functor_type = std::conditional_t<
+      SequentialInit,
+      ViewValueFunctorSequentialHostInit<device_type, ElementType>,
+      ViewValueFunctor<device_type, ElementType>>;
+  using record_type =
+      Kokkos::Impl::SharedAllocationRecord<MemorySpace, functor_type>;
+
+  static constexpr std::size_t align_mask   = 0x7;
+  static constexpr std::size_t element_size = sizeof(ElementType);
```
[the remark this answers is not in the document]
crtrott: done

core/src/View/Kokkos_ViewAlloc.hpp
```diff
@@ -242,6 +242,62 @@ struct ViewValueFunctorSequentialHostInit {
...
+  // Use this for constructing and destroying the view
+  using device_type  = Kokkos::Device<ExecutionSpace, MemorySpace>;
+  using functor_type = std::conditional_t<
+      SequentialInit,
+      ViewValueFunctorSequentialHostInit<device_type, ElementType>,
+      ViewValueFunctor<device_type, ElementType>>;
+  using record_type =
+      Kokkos::Impl::SharedAllocationRecord<MemorySpace, functor_type>;
+
+  static constexpr std::size_t align_mask   = 0x7;
```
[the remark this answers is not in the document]
crtrott: I added a comment

core/src/View/Kokkos_ViewAlloc.hpp
```diff
@@ -242,6 +242,62 @@ struct ViewValueFunctorSequentialHostInit {
   }
 };
 
+template <class ElementType, class MemorySpace, class ExecutionSpace,
+          bool Initialize, bool SequentialInit>
+Kokkos::Impl::SharedAllocationRecord<void, void>* make_shared_allocation_record(
```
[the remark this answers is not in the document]
crtrott: for now this avoids a dependency of Kokkos_ViewAlloc.hpp on Kokkos_ViewCtor.hpp. We don't necessarily need to.

core/src/View/Kokkos_ViewAlloc.hpp
```diff
@@ -242,6 +242,62 @@ struct ViewValueFunctorSequentialHostInit {
...
+    const size_t& required_span_size, std::string_view label,
+    const MemorySpace& memory_space, const ExecutionSpace* exec_space,
+    std::integral_constant<bool, Initialize>,
+    std::integral_constant<bool, SequentialInit>) {
```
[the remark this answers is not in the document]
crtrott: done

core/unit_test/view/TestReferenceCountedDataHandle.hpp
cwpearson: Since this is in a test I don't feel too strongly about it

cwpearson (approved):

PR #6544, opened 2023-10-26
Trim some fat in `CudaInternal` (towards multiple GPUs support)

dalg24:
Motivation: cleanup and remove trip hazards for supporting multiple GPUs

Do not store a whole bunch of individual device properties as static members of `CudaInternal`.  Instead always refer to the `cudaDeviceProp` that is stored.

PR #7347, opened 2024-09-16
Remove statics variables in HIPInternal

Rombur:
This is similar to https://github.com/kokkos/kokkos/pull/6544: do not store individual device properties as static members of `HIPInternal`. Instead always refer to the `hipDeviceProp` that is stored. I mostly just moved code around except for `maxWavesPerCU` where I have updated the link to the current code in the ROCm repo and changed the value for GFX 1030, 1100, and 1103. I ran the CI on a GFX 1030 and everything passes.

### DIFF
--- a/core/src/HIP/Kokkos_HIP.cpp
+++ b/core/src/HIP/Kokkos_HIP.cpp
@@ -53,12 +53,17 @@ void HIP::impl_initialize(InitializationSettings const& settings) {
   KOKKOS_IMPL_HIP_SAFE_CALL(hipSetDevice(hip_device_id));
 
   // theoretically, we can get 40 WF's / CU, but only can sustain 32 see
-  // https://github.com/ROCm-Developer-Tools/HIP/blob/a0b5dfd625d99af7e288629747b40dd057183173/vdi/hip_platform.cpp#L742
-  Impl::HIPInternal::m_maxWavesPerCU = 32;
-  Impl::HIPInternal::m_shmemPerSM    = hipProp.maxSharedMemoryPerMultiProcessor;
+  // https://github.com/ROCm/clr/blob/4d0b815d06751735e6a50fa46e913fdf85f751f0/hipamd/src/hip_platform.cpp#L362-L366
+#if defined(KOKKOS_ARCH_AMD_GFX1030) || defined(KOKKOS_ARCH_AMD_GFX1100) || \
+    defined(KOKKOS_ARCH_AMD_GFX1103)
+  const int maxWavesPerCU = 64;
+#else
+  const int maxWavesPerCU = 32;
+#endif
+  Impl::HIPInternal::m_shmemPerSM = hipProp.maxSharedMemoryPerMultiProcessor;
   Impl::HIPInternal::m_maxShmemPerBlock = hipProp.sharedMemPerBlock;
   Impl::HIPInternal::m_maxThreadsPerSM =
-      Impl::HIPInternal::m_maxWavesPerCU * Impl::HIPTraits::WarpSize;
+      maxWavesPerCU * Impl::HIPTraits::WarpSize;
 
   // Init the array for used for arbitrarily sized atomics
   desul::Impl::init_lock_arrays();  // FIXME
--- a/core/src/HIP/Kokkos_HIP.cpp
+++ b/core/src/HIP/Kokkos_HIP.cpp
@@ -52,12 +52,6 @@ void HIP::impl_initialize(InitializationSettings const& settings) {
   const auto& hipProp = Impl::HIPInternal::m_deviceProp;
   KOKKOS_IMPL_HIP_SAFE_CALL(hipSetDevice(hip_device_id));
 
-  //----------------------------------
-  // Maximum number of blocks
-  Impl::HIPInternal::m_maxBlock[0] = hipProp.maxGridSize[0];
-  Impl::HIPInternal::m_maxBlock[1] = hipProp.maxGridSize[1];
-  Impl::HIPInternal::m_maxBlock[2] = hipProp.maxGridSize[2];
-
   // theoretically, we can get 40 WF's / CU, but only can sustain 32 see
   // https://github.com/ROCm-Developer-Tools/HIP/blob/a0b5dfd625d99af7e288629747b40dd057183173/vdi/hip_platform.cpp#L742
   Impl::HIPInternal::m_maxWavesPerCU = 32;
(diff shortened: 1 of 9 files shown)


core/src/HIP/Kokkos_HIP.cpp
```diff
@@ -49,34 +49,18 @@ void HIP::impl_initialize(InitializationSettings const& settings) {
...
-  Impl::HIPInternal::m_maxBlock[2] = hipProp.maxGridSize[2];
-
   // theoretically, we can get 40 WF's / CU, but only can sustain 32 see
-  // https://github.com/ROCm-Developer-Tools/HIP/blob/a0b5dfd625d99af7e288629747b40dd057183173/vdi/hip_platform.cpp#L742
-  Impl::HIPInternal::m_maxWavesPerCU = 32;
-  Impl::HIPInternal::m_shmemPerSM    = hipProp.maxSharedMemoryPerMultiProcessor;
-  Impl::HIPInternal::m_maxShmemPerBlock = hipProp.sharedMemPerBlock;
+  // https://github.com/ROCm/clr/blob/4d0b815d06751735e6a50fa46e913fdf85f751f0/hipamd/src/hip_platform.cpp#L362-L366
+#if defined(KOKKOS_ARCH_AMD_GFX1030) || defined(KOKKOS_ARCH_AMD_GFX1100) || \
+    defined(KOKKOS_ARCH_AMD_GFX1103)
```
ldh4, line 57: I'm guessing that this list will have to be keep appended as more supported architectures are added in the future.
Is it possible to form a condition similar to what was done in the link instead, by querying a device attribute to determine a lower WF count usage?

Rombur: Maybe but we have the same logic at other places, so you are not going to gain anything by doing that.

ldh4 (approved):

core/src/HIP/Kokkos_HIP.cpp
cwpearson: Unless we expect the `const`-ness of `maxWavesPerCU` to be important for performance, I would prefer to see this query the HIP device properties as well. Looks like the `major` and `minor` fields of the device property struct are the same as this `device.isa().versionMajor()` in the commented link.

cwpearson (changes requested):

crtrott (approved):
Fine with me.

crtrott (changes requested):
Actually agree with @cwpearson I think it should be: maxThreadsPerMultiProcessor/warpSize right?

core/src/HIP/Kokkos_HIP.cpp
Rombur: > Unless we expect knowing maxWavesPerCU at compile time to be important for performance, I would prefer to see this query the HIP device properties as well.

We are not doing this for the pleasure of keeping tables in Kokkos. The peak and the sustainable values do not match. We care about the sustainable value not the peak one, that's why we hardcode the value.

cwpearson: Sorry if I wasn't entirely clear earlier. I prefer a different method by which Kokkos selects which hard-coded `32` or `64` to use: I think we should query the HIP device properties rather than the Kokkos CMake configuration to choose between them.

The downside of querying the HIP device properties is that this hard-coded value becomes dynamic at run time (won't be `const int` since you'll switch at runtime between two hard-coded values). Unless this is expected to be a big performance problem, let's choose `32` or `64` based on the HIP device property `major` field, or some similar logic.

core/src/HIP/Kokkos_HIP.cpp
```diff
@@ -50,15 +50,11 @@ void HIP::impl_initialize(InitializationSettings const& settings) {
   KOKKOS_IMPL_HIP_SAFE_CALL(
       hipGetDeviceProperties(&Impl::HIPInternal::m_deviceProp, hip_device_id));
   KOKKOS_IMPL_HIP_SAFE_CALL(hipSetDevice(hip_device_id));
-
+  std::cout << Impl::HIPInternal::m_deviceProp.major << std::endl;
```
masterleinad, line 53:
```suggestion
```

core/src/HIP/Kokkos_HIP.cpp
```diff
@@ -49,34 +49,18 @@ void HIP::impl_initialize(InitializationSettings const& settings) {
...
-    Impl::HIPInternal::m_maxWarpCount = Impl::HIPTraits::WarpSize;
-  }
-
-  //----------------------------------
-  // Maximum number of blocks
-  Impl::HIPInternal::m_maxBlock[0] = hipProp.maxGridSize[0];
-  Impl::HIPInternal::m_maxBlock[1] = hipProp.maxGridSize[1];
-  Impl::HIPInternal::m_maxBlock[2] = hipProp.maxGridSize[2];
-
   // theoretically, we can get 40 WF's / CU, but only can sustain 32 see
```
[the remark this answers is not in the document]
masterleinad:
```suggestion
```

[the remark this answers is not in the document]
Rombur: I am not removing the comment. It's not obvious from the code why we are not using 40.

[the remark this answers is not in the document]
masterleinad: I would still refer to the device version than like
```suggestion
  // for older devices, we can theoretically get 40 WF's / CU, but only can sustain 32 see
```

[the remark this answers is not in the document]
Rombur: I'll update the comment. Note that this has nothing to do with the generation of GPU.

masterleinad (approved):

cwpearson (approved):

core/src/HIP/Kokkos_HIP.cpp
```diff
@@ -49,34 +49,14 @@ void HIP::impl_initialize(InitializationSettings const& settings) {
...
-  // theoretically, we can get 40 WF's / CU, but only can sustain 32 see
-  // https://github.com/ROCm-Developer-Tools/HIP/blob/a0b5dfd625d99af7e288629747b40dd057183173/vdi/hip_platform.cpp#L742
-  Impl::HIPInternal::m_maxWavesPerCU = 32;
-  Impl::HIPInternal::m_shmemPerSM    = hipProp.maxSharedMemoryPerMultiProcessor;
-  Impl::HIPInternal::m_maxShmemPerBlock = hipProp.sharedMemPerBlock;
+  // theoretically on GFX 9XX GPUs, we can get 40 WF's / CU, but only can
+  // sustain 32 see
+  // https://github.com/ROCm/clr/blob/4d0b815d06751735e6a50fa46e913fdf85f751f0/hipamd/src/hip_platform.cpp#L362-L366
+  const int maxWavesPerCU =
+      Impl::HIPInternal::m_deviceProp.major <= 9 ? 32 : 64;
```
dalg24, line 57: Please make sure that this change is captured in the 4.5 changelog

PR #7348, opened 2024-09-17
Clean up OpenMPTarget

masterleinad:
- remove the reducer `volatile` overloads. The respective overloads for the builtin reducers were removed in https://github.com/kokkos/kokkos/commit/067c2593b58e1d9204d097432e221cfd524d7978. Related to https://github.com/kokkos/kokkos/pull/7330#discussion_r1763675491.
- `execute_tile` doesn't used its first argument in `ParallelFor` `MDRange`.

### DIFF
--- a/core/src/OpenMPTarget/Kokkos_OpenMPTarget_ParallelFor_MDRange.hpp
+++ b/core/src/OpenMPTarget/Kokkos_OpenMPTarget_ParallelFor_MDRange.hpp
@@ -52,21 +52,19 @@ class ParallelFor<FunctorType, Kokkos::MDRangePolicy<Traits...>,
 
     Policy policy = m_policy;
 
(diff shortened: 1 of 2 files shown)


rgayatri23 (approved):
LGTM. You are removing the `volatile` overloads and the `offset` parameter in MDRange which is supposed to be used for tiling. We can add that back when we have the implementation.

masterleinad:
> `offset` parameter in MDRange which is supposed to be used for tiling. We can add that back when we have the implementation.

I noticed some compiler warnings about using uninitialized variables and it turned out that the first argument os never used.

ldh4 (approved):

crtrott (approved):
Fine with me, let's see if it passes.

crtrott:
OpenMPTarget passed. Will merge.

Issue #7302, opened 2024-09-06
Nightly test failure: sycl_graph.diamond, sycl_graph.end_of_submit_control_flow failures with Sycl backend, PVC arch, intel/oneapi/2023.1.0


ndellingwood:
**Describe the bug**

A nightly Sycl job on PVC arch failed in the `sycl_graph.end_of_submit_control_flow` unit test

```
03:36:30 [ RUN      ] sycl_graph.diamond
03:36:30 /home/jenkins/blake-new/workspace/Kokkos_develop_blake_sycl_pv_oneapi_2023_1_0/kokkos/core/unit_test/TestGraph.hpp:454: Failure
03:36:30 Expected equality of these values:
03:36:30   data_host(index_A())
03:36:30     Which is: 84
03:36:30   value_A
03:36:30     Which is: 42
03:36:30 [  FAILED  ] sycl_graph.diamond (0 ms)
03:36:30 [ RUN      ] sycl_graph.end_of_submit_control_flow
03:36:30 /home/jenkins/blake-new/workspace/Kokkos_develop_blake_sycl_pv_oneapi_2023_1_0/kokkos/core/unit_test/TestGraph.hpp:540: Failure
03:36:30 Expected equality of these values:
03:36:30   data_host(index_E())
03:36:30     Which is: 257676512
03:36:30   value_B + value_E
03:36:30     Which is: 523
03:36:30 [  FAILED  ] sycl_graph.end_of_submit_control_flow (0 ms)
```

The failure began following merge of these commits

**Changes**
    Fix implicit copy assignment operator for AVX2 mask being deleted ([details](https://jenkins-son.sandia.gov/job/Kokkos_develop_blake_sycl_pv_oneapi_2023_1_0/381/changes#8daac7ce8bf849495cc7606fc5bec49585737e1f) / [githubweb](https://github.com/kokkos/kokkos/commit/8daac7ce8bf849495cc7606fc5bec49585737e1f))
    Update clang-format-check to fix it ([details](https://jenkins-son.sandia.gov/job/Kokkos_develop_blake_sycl_pv_oneapi_2023_1_0/381/changes#307e13ca51e2c73bac938506a23c91373497847a) / [githubweb](https://github.com/kokkos/kokkos/commit/307e13ca51e2c73bac938506a23c91373497847a))
[...]

PR #7349, opened 2024-09-17
Restrict OpenMP tests for OMP_NUM_THREADS detected at initialization

masterleinad:
Fixes https://github.com/kokkos/kokkos/issues/7302#issuecomment-2349285849.

It's not quite clear why `exec_space_thread_safety_team_policy` and `exec_space_thread_safety_team_policy` also need `Kokkos` to be configure with at least two threads.

The backtrace looks like
```
 bt
* thread #2, stop reason = EXC_BAD_ACCESS (code=1, address=0x28)
  * frame #0: 0x00000001000d4b04 Kokkos_CoreUnitTest_OpenMP`Kokkos::Impl::HostThreadTeamData::team_shared(this=0x0000000000000000) const at Kokkos_HostThreadTeam.hpp:284:9
    frame #1: 0x00000001000d4a6c Kokkos_CoreUnitTest_OpenMP`Kokkos::Impl::HostThreadTeamMember<Kokkos::OpenMP>::HostThreadTeamMember(this=0x0000000170602790, arg_data=0x0000000000000000, arg_league_rank=0, arg_league_size=1) at Kokkos_HostThreadTeam.hpp:427:28
    frame #2: 0x00000001000d4420 Kokkos_CoreUnitTest_OpenMP`Kokkos::Impl::HostThreadTeamMember<Kokkos::OpenMP>::HostThreadTeamMember(this=0x0000000170602790, arg_data=0x0000000000000000, arg_league_rank=0, arg_league_size=1) at Kokkos_HostThreadTeam.hpp:431:40
    frame #3: 0x00000001003635ac Kokkos_CoreUnitTest_OpenMP`std::__1::enable_if<std::is_void<void>::value, void>::type Kokkos::Impl::ParallelFor<(anonymous namespace)::run_exec_space_thread_safety_team_policy()::$_0::operator()() const::'lambda'(Kokkos::Impl::HostThreadTeamMember<Kokkos::OpenMP> const&), Kokkos::TeamPolicy<Kokkos::OpenMP>, Kokkos::OpenMP>::exec_team<void>(functor=0x00000001706028c8, data=0x0000000000000000, league_rank_begin=0, league_rank_end=1, league_size=1) at Kokkos_OpenMP_Parallel_For.hpp:300:15
    frame #4: 0x0000000100362fc8 Kokkos_CoreUnitTest_OpenMP`Kokkos::Impl::ParallelFor<(anonymous namespace)::run_exec_space_thread_safety_team_policy()::$_0::operator()() const::'lambda'(Kokkos::Impl::HostThreadTeamMember<Kokkos::OpenMP> const&), Kokkos::TeamPolicy<Kokkos::OpenMP>, Kokkos::OpenMP>::execute(this=0x00000001706028c0) const at Kokkos_OpenMP_Parallel_For.hpp:348:7
    frame #5: 0x0000000100362cc4 Kokkos_CoreUnitTest_OpenMP`void Kokkos::parallel_for<Kokkos::TeamPolicy<Kokkos::OpenMP>, (anonymous namespace)::run_exec_space_thread_safety_team_policy()::$_0::operator()() const::'lambda'(Kokkos::Impl::HostThreadTeamMember<Kokkos::OpenMP> const&), void>(str="", policy=0x0000000170602a88, functor=0x0000000170602a48) at Kokkos_Parallel.hpp:146:11
    frame #6: 0x0000000100362bec Kokkos_CoreUnitTest_OpenMP`void Kokkos::parallel_for<Kokkos::TeamPolicy<Kokkos::OpenMP>, (anonymous namespace)::run_exec_space_thread_safety_team_policy()::$_0::operator()() const::'lambda'(Kokkos::Impl::HostThreadTeamMember<Kokkos::OpenMP> const&)>(policy=0x0000000170602a88, functor=0x0000000170602a48, (null)=0x0000000000000000) at Kokkos_Parallel.hpp:155:3
    frame #7: 0x0000000100362ac8 Kokkos_CoreUnitTest_OpenMP`(anonymous namespace)::run_exec_space_thread_safety_team_policy()::$_0::operator()(this=0x000000016fdfed38) const at TestExecSpaceThreadSafety.hpp:152:7
    frame #8: 0x0000000100362a24 Kokkos_CoreUnitTest_OpenMP`::_ZN12_GLOBAL__N_117run_threaded_testIZNS_40run_exec_space_thread_safety_team_policyEvE3$_0S1_EEvT_T0_.omp_outlined_debug__(.global_tid.=0x0000000170602b9c, .bound_tid.=0x0000000170602b98, l1=0x000000016fdfed68, l2=0x000000016fdfed38) &, const (unnamed class) &) at TestExecSpaceThreadSafety.hpp:34:36
    frame #9: 0x0000000100362ba4 Kokkos_CoreUnitTest_OpenMP`::_ZN12_GLOBAL__N_117run_threaded_testIZNS_40run_exec_space_thread_safety_team_policyEvE3$_0S1_EEvT_T0_.omp_outlined(.global_tid.=0x0000000170602b9c, .bound_tid.=0x0000000170602b98, l1=0x000000016fdfed68, l2=0x000000016fdfed38) at TestExecSpaceThreadSafety.hpp:31:1
    frame #10: 0x00000001091421bc libomp.dylib`__kmp_invoke_microtask + 156
    frame #11: 0x00000001090e2878 libomp.dylib`__kmp_invoke_task_func + 332
    frame #12: 0x00000001090e1554 libomp.dylib`__kmp_launch_thread + 420
    frame #13: 0x0000000109125d94 libomp.dylib`__kmp_launch_worker(void*) + 280
    frame #14: 0x0000000198e0df94 libsystem_pthread.dylib`_pthread_start + 136
```

### DIFF
--- a/core/unit_test/TestExecSpacePartitioning.hpp
+++ b/core/unit_test/TestExecSpacePartitioning.hpp
@@ -79,7 +79,7 @@ void check_distinctive([[maybe_unused]] ExecSpace exec1,
 #ifdef KOKKOS_ENABLE_OPENMP
 template <class Lambda1, class Lambda2>
 void run_threaded_test(const Lambda1 l1, const Lambda2 l2) {
-  if (omp_get_max_threads() == 1)
+  if (omp_get_max_threads() < 2)
     GTEST_SKIP() << "insufficient number of supported concurrent threads";
 
 #pragma omp parallel num_threads(2)
--- a/core/unit_test/TestExecSpacePartitioning.hpp
+++ b/core/unit_test/TestExecSpacePartitioning.hpp
@@ -79,6 +79,11 @@ void check_distinctive([[maybe_unused]] ExecSpace exec1,
 #ifdef KOKKOS_ENABLE_OPENMP
 template <class Lambda1, class Lambda2>
 void run_threaded_test(const Lambda1 l1, const Lambda2 l2) {
+#ifdef KOKKOS_ENABLE_OPENMP
+  if (Kokkos::OpenMP().concurrency() == 1)
+    GTEST_SKIP() << "test needs at least two OpenMP threads";
+#endif
+
 #pragma omp parallel num_threads(2)
   {
     if (omp_get_thread_num() == 0) l1();
@@ -126,6 +131,10 @@ void test_partitioning(std::vector<TEST_EXECSPACE>& instances) {
 }
 
 TEST(TEST_CATEGORY, partitioning_by_args) {
+#ifdef KOKKOS_ENABLE_OPENMP
+  if (Kokkos::OpenMP().concurrency() == 1)
+    GTEST_SKIP() << "test needs at least two OpenMP threads";
+#endif
   auto instances =
       Kokkos::Experimental::partition_space(TEST_EXECSPACE(), 1, 1);
   ASSERT_EQ(int(instances.size()), 2);
@@ -133,6 +142,10 @@ TEST(TEST_CATEGORY, partitioning_by_args) {
 }
 
 TEST(TEST_CATEGORY, partitioning_by_vector) {
+#ifdef KOKKOS_ENABLE_OPENMP
+  if (Kokkos::OpenMP().concurrency() == 1)
+    GTEST_SKIP() << "test needs at least two OpenMP threads";
+#endif
   // Make sure we can use a temporary as argument for weights
   auto instances = Kokkos::Experimental::partition_space(
       TEST_EXECSPACE(), std::vector<int> /*weights*/ {1, 1});
--- a/core/unit_test/TestExecSpacePartitioning.hpp
+++ b/core/unit_test/TestExecSpacePartitioning.hpp
@@ -79,10 +79,8 @@ void check_distinctive([[maybe_unused]] ExecSpace exec1,
 #ifdef KOKKOS_ENABLE_OPENMP
 template <class Lambda1, class Lambda2>
 void run_threaded_test(const Lambda1 l1, const Lambda2 l2) {
-#ifdef KOKKOS_ENABLE_OPENMP
-  if (Kokkos::OpenMP().concurrency() == 1)
-    GTEST_SKIP() << "test needs at least two OpenMP threads";
-#endif
+  if (omp_get_max_threads() == 1)
+    GTEST_SKIP() << "insufficient number of supported concurrent threads";
 
 #pragma omp parallel num_threads(2)
   {
@@ -131,10 +129,6 @@ void test_partitioning(std::vector<TEST_EXECSPACE>& instances) {
 }
 
 TEST(TEST_CATEGORY, partitioning_by_args) {
-#ifdef KOKKOS_ENABLE_OPENMP
-  if (Kokkos::OpenMP().concurrency() == 1)
-    GTEST_SKIP() << "test needs at least two OpenMP threads";
-#endif
   auto instances =
       Kokkos::Experimental::partition_space(TEST_EXECSPACE(), 1, 1);
   ASSERT_EQ(int(instances.size()), 2);
@@ -142,10 +136,6 @@ TEST(TEST_CATEGORY, partitioning_by_args) {
 }
 
 TEST(TEST_CATEGORY, partitioning_by_vector) {
-#ifdef KOKKOS_ENABLE_OPENMP
-  if (Kokkos::OpenMP().concurrency() == 1)
-    GTEST_SKIP() << "test needs at least two OpenMP threads";
-#endif
   // Make sure we can use a temporary as argument for weights
   auto instances = Kokkos::Experimental::partition_space(
       TEST_EXECSPACE(), std::vector<int> /*weights*/ {1, 1});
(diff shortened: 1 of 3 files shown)


crtrott (approved):
Fine for now: though we should figure out WHY it does need multiple threads when partitioning. Semantically from the Kokkos perspective it should work without, but we may do something like omp_set_num_threads(0) if the numbers don't work out. @masterleinad can you open an issue for that?

masterleinad:
> though we should figure out WHY it does need multiple threads when partitioning.

What else would we do if we can't assign any positive of threads to a execution space instance? According to https://www.openmp.org/spec-html/5.0/openmpsu110.html, the behavior for `omp_set_num_threads(0)` is implementation-defined.

masterleinad:
> @masterleinad can you open an issue for that?

See #7357.

dalg24 (changes requested):
I do not understand why the Graph tests require more than 1 omp thread to execute.

masterleinad:
> I do not understand why the Graph tests require more than 1 omp thread to execute.

We want ti use multiple independent execution space instances and we are using `partition_space` with an `OpenMP` execution space instance to do that. That means that we distribute the number of threads for the input instance to between the output execution space instances and each of them most have at least one thread assigned which means the input execution space instance must have at least as many execution space instances assigned as the number of execution space instances we wan to create.

dalg24:
How isn't it just a bug in `create_OpenMP_instances()` rather than a property of that test?

dalg24 (commented):
On 2nd thought I don't see why we need the skip in the other tests either considering that we do
https://github.com/kokkos/kokkos/blob/94d211a617ea655e774d3460c1bf2ee86431b2f6/core/unit_test/TestExecSpacePartitioning.hpp#L87-L91

And to be clear I am saying that the current implementation of `partition_space()` for OpenMP is wrong, that even if you have omp num threads equal to 1, partitioning the default execution space instance into multiple instances should work and that all created instances should also have 1 thread.

crtrott:
Just to make that absolutely clear: I was ok with disabling the test until we have fixed partition space. Yes the threaded test which sets omp_num_threads(2) does need explicitly skip out (though we could check something like omp_max_threads() or whatever it's called instead of concurrency), but the ones which just fail because of partition space: for those we should fix partition space. As a stop gap I am ok with merging a test disabling.

core/unit_test/TestExecSpacePartitioning.hpp
```diff
@@ -79,6 +79,11 @@ void check_distinctive([[maybe_unused]] ExecSpace exec1,
 #ifdef KOKKOS_ENABLE_OPENMP
 template <class Lambda1, class Lambda2>
 void run_threaded_test(const Lambda1 l1, const Lambda2 l2) {
+#ifdef KOKKOS_ENABLE_OPENMP
+  if (Kokkos::OpenMP().concurrency() == 1)
+    GTEST_SKIP() << "test needs at least two OpenMP threads";
+#endif
```
dalg24, line 85:
```suggestion
  if (omp_get_max_threads() == 1)
    GTEST_SKIP() << "insufficient number of supported concurrent threads";
```
Same comment everywhere in this file and the thread safety one

core/unit_test/TestExecSpacePartitioning.hpp
```diff
@@ -79,6 +79,9 @@ void check_distinctive([[maybe_unused]] ExecSpace exec1,
 #ifdef KOKKOS_ENABLE_OPENMP
 template <class Lambda1, class Lambda2>
 void run_threaded_test(const Lambda1 l1, const Lambda2 l2) {
+  if (omp_get_max_threads() == 1)
```
dalg24, line 82:
```suggestion
  if (omp_get_max_threads() < 2)
```

core/unit_test/TestExecSpaceThreadSafety.hpp
```diff
@@ -28,6 +28,9 @@ namespace {
 #ifdef KOKKOS_ENABLE_OPENMP
 template <class Lambda1, class Lambda2>
 void run_threaded_test(const Lambda1 l1, const Lambda2 l2) {
+  if (omp_get_max_threads() == 1)
```
dalg24, line 31:
```suggestion
  if (omp_get_max_threads() < 2)
```

core/unit_test/TestGraph.hpp
```diff
@@ -200,6 +200,11 @@ TEST_F(TEST_CATEGORY_FIXTURE_DEATH(graph), can_instantiate_only_once) {
 // one passed to the Kokkos::Graph constructor.
 TEST_F(TEST_CATEGORY_FIXTURE(graph),
        submit_onto_another_execution_space_instance) {
+#ifdef KOKKOS_ENABLE_OPENMP  // FIXME_OPENMP partition_space
+  if (ex.concurrency() == 1)
```
dalg24, line 204:
```suggestion
  if (ex.concurrency() < 2)
```

dalg24 (approved):
Sure

masterleinad:
Only `Stage 'CUDA-12.2-NVHPC-AS-HOST-COMPILER'` is failing with an unrelated segmentation fault in the compiler.

PR #7351, opened 2024-09-17
ViewRefactor: Fix issues with view refactor in Sacado

crtrott:
Somehow the nested dimensions wouldn't be correclty represented in strides, and DynRankView wasn't accounting for all the specialized helper things in Sacado.

There is a few nasty things here: DynRankView can either take the hidden dimension as an additional argument to its constructor, or pass it through the ViewAllocation properties, via a specialization of these properties in Sacado. Also the DynRankDimTraits are specialized etc. 

It's all fairly messy, but maybe we can get rid of all of it after the full View refactor by rewriting Sacado via using Accessors.

### DIFF
--- a/containers/src/Kokkos_DynRankView.hpp
+++ b/containers/src/Kokkos_DynRankView.hpp
@@ -108,6 +108,9 @@ struct DynRankDimTraits {
   }
 
   // Create the layout for the rank-7 view.
+  // Because the underlying View is rank-7, preserve "unspecified" for
+  // dimension 8.
+
   // Non-strided Layout
   template <typename Layout>
   KOKKOS_INLINE_FUNCTION static std::enable_if_t<
@@ -123,7 +126,7 @@ struct DynRankDimTraits {
         layout.dimension[4] != unspecified ? layout.dimension[4] : 1,
         layout.dimension[5] != unspecified ? layout.dimension[5] : 1,
         layout.dimension[6] != unspecified ? layout.dimension[6] : 1,
-        layout.dimension[7] != unspecified ? layout.dimension[7] : 1);
+        layout.dimension[7] != unspecified ? layout.dimension[7] : unspecified);
     new_layout.stride = layout.stride;
     return new_layout;
   }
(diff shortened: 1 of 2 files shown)


containers/src/Kokkos_DynRankView.hpp
```diff
@@ -921,9 +929,14 @@ KOKKOS_INLINE_FUNCTION auto subdynrankview(
...
-                     typename sub_t::device_type,
-                     typename sub_t::memory_traits>(sub, new_rank);
+
+  using return_type =
+      DynRankView<typename sub_t::value_type, Kokkos::LayoutStride,
+                  typename sub_t::device_type, typename sub_t::memory_traits>;
+  return static_cast<return_type>(
+      DynRankView<typename sub_t::value_type, typename sub_t::array_layout,
+                  typename sub_t::device_type, typename sub_t::memory_traits>(
+          sub, new_rank));
```
masterleinad, line 939: Why did you need to change `return_type`?

romintomasetti (commented):
Do you think it's possible to come up with a test that somehow mimics `Sacado`'s use case ?

containers/src/Kokkos_DynRankView.hpp
crtrott: Because DynRankView used to only ever return LayoutStride in subviews, and some code downstream was relying on that (doing ternary expression with different dimensionality subview, which resulted in different types).

crtrott:
Yes but the test would be roughly the size of Sacado. Essentially Sacado partially specializes almost every implementation detail of Kokkos::View. And as soon as the View refactor is done all these details are gonna be gone and we switch to much cleaner accessors to achieve the same goal. I don't believe it's useful for us to work out those tests. What we could to is an add a CI pass doing Trilinos with just Sacado enabled. It's not terribly long to compile. I will put our folks on that as soon as we have figured out the new testing mechanism on Sandia's machines.

containers/src/Kokkos_DynRankView.hpp
masterleinad: If that's something to change later again (after refactoring Sacado) we should add a comment to that extent.

crtrott: Its not actually Sacado its in Intrepid2

crtrott: We can make that part of the DynRank Review later.

masterleinad (approved):

janciesko (approved):

ndellingwood:
Tested this PR to check if results of failing tests in https://trilinos-cdash.sandia.gov/viewTest.php?onlyfailed&buildid=1861613 were resolved, only 3 failures (big improvement, thanks @crtrott !):

```
	3122 - PanzerAdaptersSTK_tQuad8ToQuad4Factory_MPI_2 (Failed)
	3123 - PanzerAdaptersSTK_tQuadraticToLinearMeshFactory_MPI_2 (Failed)
	3159 - PanzerAdaptersSTK_tPointBasisValuesEvaluator_MPI_1 (Failed)
```

I'm rebuilding Panzer to compare those tests against a Trilinos build without kokkos@develop to further triage

ndellingwood:
The first two tests failed with Trilinos@develop (not correlated with kokkos@develop changes), however `PanzerAdaptersSTK_tPointBasisValuesEvaluator_MPI_1` passes with kokkos@develop and needs further investigation

ndellingwood (approved):
I confirmed most of the failing tests from e.g. https://trilinos-cdash.sandia.gov/viewTest.php?onlyfailed&buildid=1861613 pass with this PR
One failing test `PanzerAdaptersSTK_tPointBasisValuesEvaluator_MPI_1` needs more investigation

ndellingwood:
@crtrott I think this is ready for merge?

crtrott:
Failed with a GPU issue one test. Will merge.

Issue #7294, opened 2024-09-05
Add a way to disable `cudaMallocAsync` with Spack


cedricchevalier19:
**Describe the bug**

`cray-mpich` does not work well with the default enabled `cudaMallocAsync`.
When building Kokkos with Spack, there is no way to disable `cudaMallocAsync`.

see : spack/spack#46090

Note:
The current spack recipe contains an older [workaround](https://github.com/spack/spack/blob/2107a88514d8afe8a72f879cbde1351ce65180f7/var/spack/repos/builtin/packages/kokkos/package.py#L366C4-L375) that is not working anymore. 

```python 
        # Kokkos 4.2.00+ changed the default to Kokkos_ENABLE_IMPL_CUDA_MALLOC_ASYNC=on
        # which breaks GPU-aware with Cray-MPICH
        # See https://github.com/kokkos/kokkos/pull/6402
        # TODO: disable this once Cray-MPICH is fixed
        if (
            self.spec.satisfies("@4.2.00:")
            and "mpi" in self.spec
            and self.spec["mpi"].name == "cray-mpich"
        ):
            options.append(self.define("Kokkos_ENABLE_IMPL_CUDA_MALLOC_ASYNC", False))
```

Kokkos does not depend on `mpi`, so it cannot adapt its configuration against it.

dalg24:
Does Spack support appending user-specified cmake flags in the configuration step?

cedricchevalier19:
> Does Spack support appending user-specified cmake flags in the configuration step?

Not to my knowledge (in fact I almost hope it is not possible to by-pass Spack configuration when building things).

I think a `alloc_async` variant, enabled by default, should be added to the recipe. It can work with either cuda and hip/rocm backends.

rbberger:
Another data point. Based on this https://developer.nvidia.com/blog/using-cuda-stream-ordered-memory-allocator-part-2/#interprocess_communication_support it seems that `cudaIpcGetMemHandle` (which `mpich`, `cray-mpich`, `openmpi`, `libfabric` use interally) won't work with `cudaMallocAsync` memory at all. Apparently only UCX is aware of MemPools. I hope I'm wrong.

cedricchevalier19:
Always disabled.

PR #7353, opened 2024-09-18
By default, do not use cudaMallocAsync

Rombur:
Given all the problems that users reported on slack, stop using `cudaMallocAsync` by default. This is a partial revert of https://github.com/kokkos/kokkos/pull/6402

### DIFF
--- a/.jenkins
+++ b/.jenkins
@@ -465,7 +465,7 @@ pipeline {
                                 -DKokkos_ENABLE_LIBDL=OFF \
                                 -DKokkos_ENABLE_OPENMP=ON \
                                 -DKokkos_ENABLE_IMPL_MDSPAN=OFF \
-                                -DKokkos_ENABLE_IMPL_CUDA_MALLOC_ASYNC=OFF \
+                                -DKokkos_ENABLE_IMPL_CUDA_MALLOC_ASYNC=ON \
                               .. && \
                               make -j8 && ctest --verbose && \
                               cd ../example/build_cmake_in_tree && \
--- a/Makefile.kokkos
+++ b/Makefile.kokkos
@@ -30,7 +30,7 @@ KOKKOS_TRIBITS ?= "no"
 KOKKOS_STANDALONE_CMAKE ?= "no"
 
 # Default settings specific options.
-# Options: force_uvm,use_ldg,rdc,enable_lambda,enable_constexpr,disable_malloc_async
+# Options: force_uvm,use_ldg,rdc,enable_lambda,enable_constexpr,enable_malloc_async
 KOKKOS_CUDA_OPTIONS ?= ""
 
 # Options: rdc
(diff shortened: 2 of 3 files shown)


crtrott (approved):

cedricchevalier19 (approved):

Rombur:
Can we merge this. The failing test seems unrelated.

dalg24:
> Can we merge this. The failing test seems unrelated.

I just went over the last developer meeting note and don't quite follow what was the rational for changing the default, and what the future roadmap would be.
Without more context I tend to disagree with the direction.  I am not blocking but I won't merge until I am brought up to speed.

stanmoore1:
Spack is broken (https://github.com/kokkos/kokkos/issues/7294), LAMMPS is crashing and I'm tired of telling users to turn it off (https://matsci.org/t/lammps-snap-performance-drop-with-llvm-clang-on-nvidia-a100/57196/16?u=stamoor), so now we are disabling it ourselves by default. It is broken even on high profile machines like ALCF Polaris. It would be nice if it wasn't the default in Kokkos.

Rombur:
> I just went over the last developer meeting note and don't quite follow what was the rational for changing the default

There are two problems:
 1. more people than expected are using MPI libraries built on old UCX libraries
 2. Nvidia's [IPC](https://developer.nvidia.com/blog/using-cuda-stream-ordered-memory-allocator-part-2/#interprocess_communication_support) does not currently work if the option is enabled. We need to rework how the memory pool works.

fglines-nv:
This default causes on Cray's MPICH and thus Venado, another new high-profile system. You should merge this one in soon.

I can understand using asynchronous mallocs by default for CUDA since it should speed up certain malloc-heavy workflows. Because most Kokkos applications (at least that I've worked with) use only a single CUDA stream there's no danger accessing GPU data before it's been malloced.

However since it breaks most MPI libraries due to changes needed for NVIDIA's IPC we should disable this by default. Maybe once the memory pool changes are made to work with IPC+`cudaMallocAsync` we can enable this by default for MPI libraries that also work with the new interface -- that might only be HPCX at time of writing.

dalg24:
@crtrott is able to merge if he'd decide to do so.  As far as I am concerned I would wait the developer meeting (tomorrow) and demand more background information as well as an update to the CUDA known issues section in the [doc](https://kokkos.org/kokkos-core-wiki/known-issues.html#cuda)

cedricchevalier19:
> There are two problems:
> 
>     1. more people than expected are using MPI libraries built on old UCX libraries

It is not always with UCX anyway.

>     2. Nvidia's [IPC](https://developer.nvidia.com/blog/using-cuda-stream-ordered-memory-allocator-part-2/#interprocess_communication_support) does not currently work if the option is enabled. We need to rework how the memory pool works.

This is more tricky. It is not, per se, the job of a library to change the behavior of the default memory pool (https://developer.nvidia.com/blog/using-cuda-stream-ordered-memory-allocator-part-2/#library_composability).

PR #7354, opened 2024-09-18
OpenACC update for host part 1

seyonglee:
This PR contains a subset of PR #7343 : 
- Update cmake/make for Clacc to use `--offload-arch=native` option in the automatic fallback mode.
- Use different shared allocation specializations for the CPU target and the GPU target.
- Change MemorySpaceAccess behavior of OpenACCSpace when targeting the host to mimic the CudaUVMSpace behavior.
- Apply different iteration-to-thread mapping strategies for TeamPolicy targeting CPU and GPU.

### DIFF
--- a/core/src/OpenACC/Kokkos_OpenACCSpace.hpp
+++ b/core/src/OpenACC/Kokkos_OpenACCSpace.hpp
@@ -32,11 +32,18 @@ class OpenACCSpace {
   OpenACCSpace() = default;
 
   /**\brief  Allocate untracked memory in the space */
-  void* allocate(const Kokkos::Experimental::OpenACC& exec_space,
-                 const size_t arg_alloc_size) const;
-  void* allocate(const Kokkos::Experimental::OpenACC& exec_space,
-                 const char* arg_label, const size_t arg_alloc_size,
-                 const size_t arg_logical_size = 0) const;
+  template <typename ExecutionSpace>
+  void* allocate(const ExecutionSpace& exec_space,
+                 const size_t arg_alloc_size) const {
+    return allocate(exec_space, "[unlabeled]", arg_alloc_size);
+  }
+  template <typename ExecutionSpace>
+  void* allocate(const ExecutionSpace& exec_space, const char* arg_label,
+                 const size_t arg_alloc_size,
+                 const size_t arg_logical_size = 0) const {
+    return impl_allocate(exec_space, arg_label, arg_alloc_size,
+                         arg_logical_size);
+  }
   void* allocate(const size_t arg_alloc_size) const;
   void* allocate(const char* arg_label, const size_t arg_alloc_size,
                  const size_t arg_logical_size = 0) const;
@@ -50,11 +57,16 @@ class OpenACCSpace {
   static constexpr char const* name() { return "OpenACCSpace"; }
 
  private:
-  void* impl_allocate(const Kokkos::Experimental::OpenACC& exec_space,
-                      const char* arg_label, const size_t arg_alloc_size,
+  template <typename ExecutionSpace>
+  void* impl_allocate(const ExecutionSpace& exec_space, const char* arg_label,
+                      const size_t arg_alloc_size,
                       const size_t arg_logical_size = 0,
-                      const Kokkos::Tools::SpaceHandle =
-                          Kokkos::Tools::make_space_handle(name())) const;
+                      const Kokkos::Tools::SpaceHandle arg_handle =
+                          Kokkos::Tools::make_space_handle(name())) const {
+    (void)exec_space;
+    return impl_allocate(arg_label, arg_alloc_size, arg_logical_size,
+                         arg_handle);
+  }
   void* impl_allocate(const char* arg_label, const size_t arg_alloc_size,
                       const size_t arg_logical_size = 0,
                       const Kokkos::Tools::SpaceHandle =
--- a/core/src/OpenACC/Kokkos_OpenACCSpace.hpp
+++ b/core/src/OpenACC/Kokkos_OpenACCSpace.hpp
@@ -84,25 +84,21 @@ class OpenACCSpace {
 template <>
 struct Kokkos::Impl::MemorySpaceAccess<Kokkos::HostSpace,
                                        Kokkos::Experimental::OpenACCSpace> {
+  enum : bool { assignable = false };
 #if defined(KOKKOS_ENABLE_OPENACC_FORCE_HOST_AS_DEVICE)
-  enum : bool{assignable = false};
   enum : bool{accessible = true};
 #else
-  enum : bool { assignable = false };
   enum : bool { accessible = false };
 #endif
+  enum : bool { deepcopy = true };
 };
 
 template <>
 struct Kokkos::Impl::MemorySpaceAccess<Kokkos::Experimental::OpenACCSpace,
                                        Kokkos::HostSpace> {
-#if defined(KOKKOS_ENABLE_OPENACC_FORCE_HOST_AS_DEVICE)
-  enum : bool{assignable = false};
-  enum : bool{accessible = false};
-#else
   enum : bool { assignable = false };
   enum : bool { accessible = false };
-#endif
+  enum : bool { deepcopy = true };
 };
 
 template <>
@@ -110,6 +106,7 @@ struct Kokkos::Impl::MemorySpaceAccess<Kokkos::Experimental::OpenACCSpace,
                                        Kokkos::Experimental::OpenACCSpace> {
   enum : bool { assignable = true };
   enum : bool { accessible = true };
+  enum : bool { deepcopy = true };
 };
 /*--------------------------------------------------------------------------*/
 
--- a/core/src/OpenACC/Kokkos_OpenACCSpace.hpp
+++ b/core/src/OpenACC/Kokkos_OpenACCSpace.hpp
@@ -58,12 +58,11 @@ class OpenACCSpace {
 
  private:
   template <typename ExecutionSpace>
-  void* impl_allocate(const ExecutionSpace& exec_space, const char* arg_label,
+  void* impl_allocate(const ExecutionSpace&, const char* arg_label,
                       const size_t arg_alloc_size,
                       const size_t arg_logical_size = 0,
                       const Kokkos::Tools::SpaceHandle arg_handle =
                           Kokkos::Tools::make_space_handle(name())) const {
-    (void)exec_space;
     return impl_allocate(arg_label, arg_alloc_size, arg_logical_size,
                          arg_handle);
   }
--- a/core/src/OpenACC/Kokkos_OpenACCSpace.hpp
+++ b/core/src/OpenACC/Kokkos_OpenACCSpace.hpp
@@ -86,7 +86,7 @@ template <>
 struct Kokkos::Impl::MemorySpaceAccess<Kokkos::HostSpace,
                                        Kokkos::Experimental::OpenACCSpace> {
 #if defined(KOKKOS_ENABLE_OPENACC_FORCE_HOST_AS_DEVICE)
-  enum : bool{assignable = true};
+  enum : bool{assignable = false};
   enum : bool{accessible = true};
 #else
   enum : bool { assignable = false };
@@ -98,8 +98,8 @@ template <>
 struct Kokkos::Impl::MemorySpaceAccess<Kokkos::Experimental::OpenACCSpace,
                                        Kokkos::HostSpace> {
 #if defined(KOKKOS_ENABLE_OPENACC_FORCE_HOST_AS_DEVICE)
-  enum : bool{assignable = true};
-  enum : bool{accessible = true};
+  enum : bool{assignable = false};
+  enum : bool{accessible = false};
 #else
   enum : bool { assignable = false };
   enum : bool { accessible = false };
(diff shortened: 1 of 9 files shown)


core/src/OpenACC/Kokkos_OpenACCSpace.hpp
```diff
@@ -62,11 +69,16 @@ class OpenACCSpace {
...
-                      const char* arg_label, const size_t arg_alloc_size,
+  template <typename ExecutionSpace>
+  void* impl_allocate(const ExecutionSpace& exec_space, const char* arg_label,
+                      const size_t arg_alloc_size,
                       const size_t arg_logical_size = 0,
-                      const Kokkos::Tools::SpaceHandle =
-                          Kokkos::Tools::make_space_handle(name())) const;
+                      const Kokkos::Tools::SpaceHandle arg_handle =
+                          Kokkos::Tools::make_space_handle(name())) const {
+    (void)exec_space;
```
tcclevenger, line 78: Slight preference for me, to be consistent with the other memory space allocate functions, we would have in the function declaration `void* impl_allocate(const ExecutionSpace&, ...)` instead of having `(void)exec_space`.

core/src/OpenACC/Kokkos_OpenACC_Team.hpp
```diff
@@ -134,17 +141,30 @@ class OpenACCTeamMember {
...
-      : m_team_size(team_size),
-        m_league_rank(league_rank),
+      : m_league_rank(league_rank),
         m_league_size(league_size),
+        m_team_size(team_size),
         m_vector_length(vector_length) {
 #ifdef KOKKOS_COMPILER_NVHPC
     m_team_rank = __pgi_vectoridx();
+#else
+#ifdef KOKKOS_COMPILER_CLANG
```
dalg24, line 151:
```suggestion
#elif defined(KOKKOS_COMPILER_CLANG)
```
Presumably this never worked so please double check you need that change

core/src/OpenACC/Kokkos_OpenACC_Team.hpp
```diff
@@ -134,17 +141,30 @@ class OpenACCTeamMember {
...
+    m_team_rank = omp_get_thread_num();
 #else
     m_team_rank = 0;
+#endif
```
dalg24, line 155:
```suggestion
```

core/src/OpenACC/Kokkos_OpenACC_ParallelReduce_Team.hpp
```diff
@@ -372,31 +385,29 @@ KOKKOS_INLINE_FUNCTION void parallel_reduce(
...
-                                        vector_length);                    \
-      functor(team, val);                                                  \
-    }                                                                      \
-    acc_wait(async_arg);                                                   \
-    aval = val;                                                            \
-  }                                                                        \
+    KOKKOS_IMPL_ACC_PRAGMA(parallel loop gang num_gangs(league_size) KOKKOS_IMPL_ACC_REDUCE_TEAM_PRAGMA reduction(OPERATOR : val) copyin(functor) async(async_arg))                                            \
+    /* clang-format on */                                               \
+    for (int i = 0; i < KOKKOS_IMPL_ACC_REDUCE_TEAM_ITRS; i++) {        \
+      KOKKOS_IMPL_ACC_REDUCE_TEAM_MEMBER_INIT                           \
```
dalg24, line 405: Why that change?
I feel like it obfuscate the code

core/src/OpenACC/Kokkos_OpenACC_ParallelReduce_Team.hpp
```diff
@@ -151,8 +151,18 @@ KOKKOS_INLINE_FUNCTION void parallel_reduce(
 #define KOKKOS_IMPL_ACC_REDUCE_TEAM_PRAGMA \
   vector vector_length(team_size* vector_length)
 #define KOKKOS_IMPL_ACC_REDUCE_TEAM_ITRS league_size* team_size* vector_length
-#define KOKKOS_IMPL_ACC_REDUCE_TEAM_LEAGUE_ID_INIT \
-  i / (team_size * vector_length)
+#ifdef KOKKOS_ENABLE_OPENACC_FORCE_HOST_AS_DEVICE
+#define KOKKOS_IMPL_ACC_REDUCE_TEAM_MEMBER_INIT                          \
+  int league_rank = i / (team_size * vector_length);                     \
+  int team_rank   = i % (team_size * vector_length);                     \
```
dalg24, line 157: Why is it here that we need to compute the team rank ourselves?

core/src/OpenACC/Kokkos_OpenACC_Team.hpp
```diff
@@ -134,17 +141,30 @@ class OpenACCTeamMember {
...
 #endif
   }
 
+  OpenACCTeamMember(const int league_rank, const int league_size,
+                    const int team_rank, const int team_size,
+                    const int vector_length)
```
dalg24, line 161: If that constructor is only needed when forcing the most as device I am tempted to guard it's definition since it is potentially confusing to have both weakly-typed constructor coexist.
If applicable hide the other definition for host as device.

dalg24 (commented):
This still looks like a medley of unrelated changes.
In particular please spell out why we want OpenACCSpace to be more like a SharedSpace and why we need to compute ourselves the team rank when forcing the host as device.

core/src/OpenACC/Kokkos_OpenACC_Team.hpp
```diff
@@ -42,10 +45,10 @@ class OpenACCTeamMember {
 
   scratch_memory_space m_team_shared;
   int m_team_scratch_size[2];
-  int m_team_rank;
-  int m_team_size;
   int m_league_rank;
   int m_league_size;
+  int m_team_rank;
+  int m_team_size;
```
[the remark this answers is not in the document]
seyonglee: Just to be aligned with the initialization order in the `OpenACCTEamMember` constructor. (This change is NOT required but for personal preference.)

core/src/OpenACC/Kokkos_OpenACC_Team.hpp
```diff
@@ -67,7 +70,11 @@ class OpenACCTeamMember {
 
   KOKKOS_FUNCTION int league_rank() const { return m_league_rank; }
   KOKKOS_FUNCTION int league_size() const { return m_league_size; }
+#ifdef KOKKOS_COMPILER_CLANG
+  KOKKOS_FUNCTION int team_rank() const { return omp_get_thread_num(); }
+#else
   KOKKOS_FUNCTION int team_rank() const { return m_team_rank; }
+#endif
```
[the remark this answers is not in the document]
seyonglee: In the none-collapsed hierarchical implementation for the Clacc compiler, `omp_get_thread_num()` will return different values depending on the calling context. Therefore, `team_rank()` will internally call `omp_get_thread_num()` whenever `team_rank()` is called.

core/src/OpenACC/Kokkos_OpenACC_Team.hpp
```diff
@@ -21,6 +21,9 @@
 #include <impl/Kokkos_Traits.hpp>
 #include <OpenACC/Kokkos_OpenACC.hpp>
 #include <Kokkos_ExecPolicy.hpp>
+#ifdef KOKKOS_COMPILER_CLANG
+#include <omp.h>
```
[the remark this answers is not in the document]
seyonglee: The OpenACC standard does not provide a way to retrieve the gang/worker/vector IDs. For the Clacc compilation, we use the underlying OpenMP runtime API to get those ID values (e.g., `omp_get_thread_num()` for OpenACC vector ID (Kokkos team rank)).

core/src/OpenACC/Kokkos_OpenACC_ParallelReduce_Team.hpp
seyonglee: It is because when offloading to the host, the NVHPC OpenACC compiler parallelizes only the gang loop and ignores  worker/vector loops, which is not standard feature but the NVHPC-specific behavior.

core/src/OpenACC/Kokkos_OpenACC_ParallelReduce_Team.hpp
seyonglee: It is because different mapping strategies are needed for the host target and the GPU target. 
(`KOKKOS_IMPL_ACC_REDUCE_TEAM_MEMBER_INIT` is defined differently according to the target device type and compiler type.)

core/src/OpenACC/Kokkos_OpenACC_Team.hpp
seyonglee: Disabled unused constructor as suggested.

core/src/OpenACC/Kokkos_OpenACCSpace.hpp
seyonglee: Are you suggesting to separate the function template declaration from the function template definition?
If so, where should we put the function template definition? (We cannot put it in the `Kokkos_OpenACCSpace.cpp` since the definition will not be visible to the template instantiation points.)
If not suggesting separation, can you elaborate your suggestion?

core/src/OpenACC/Kokkos_OpenACCSpace.hpp
```diff
@@ -99,8 +111,8 @@ template <>
 struct Kokkos::Impl::MemorySpaceAccess<Kokkos::Experimental::OpenACCSpace,
                                        Kokkos::HostSpace> {
 #if defined(KOKKOS_ENABLE_OPENACC_FORCE_HOST_AS_DEVICE)
-  enum : bool{assignable = true};
-  enum : bool{accessible = true};
+  enum : bool{assignable = false};
+  enum : bool{accessible = false};
```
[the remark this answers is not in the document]
seyonglee: This follows CudaUVMSpace access rules.

core/src/OpenACC/Kokkos_OpenACCSpace.hpp
```diff
@@ -86,7 +98,7 @@ template <>
 struct Kokkos::Impl::MemorySpaceAccess<Kokkos::HostSpace,
                                        Kokkos::Experimental::OpenACCSpace> {
 #if defined(KOKKOS_ENABLE_OPENACC_FORCE_HOST_AS_DEVICE)
-  enum : bool{assignable = true};
+  enum : bool{assignable = false};
   enum : bool{accessible = true};
```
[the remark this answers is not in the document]
seyonglee: This follows CudaUVMSpace access rules.

core/src/OpenACC/Kokkos_OpenACCSpace.hpp
```diff
@@ -44,11 +44,18 @@ class OpenACCSpace {
...
-                 const size_t arg_alloc_size) const;
-  void* allocate(const Kokkos::Experimental::OpenACC& exec_space,
-                 const char* arg_label, const size_t arg_alloc_size,
-                 const size_t arg_logical_size = 0) const;
+  template <typename ExecutionSpace>
+  void* allocate(const ExecutionSpace& exec_space,
+                 const size_t arg_alloc_size) const {
+    return allocate(exec_space, "[unlabeled]", arg_alloc_size);
+  }
+  template <typename ExecutionSpace>
```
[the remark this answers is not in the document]
seyonglee: This follows CudaUVMSpace memory allocation implementation patterns.

core/src/OpenACC/Kokkos_OpenACCSpace.hpp
```diff
@@ -62,11 +69,16 @@ class OpenACCSpace {
   static constexpr char const* name() { return "OpenACCSpace"; }
 
  private:
-  void* impl_allocate(const Kokkos::Experimental::OpenACC& exec_space,
-                      const char* arg_label, const size_t arg_alloc_size,
+  template <typename ExecutionSpace>
+  void* impl_allocate(const ExecutionSpace& exec_space, const char* arg_label,
```
tcclevenger, line 73:
```suggestion
  void* impl_allocate(const ExecutionSpace&, const char* arg_label,
```

core/src/OpenACC/Kokkos_OpenACCSpace.hpp
tcclevenger, line 78:
```suggestion
```

tcclevenger (commented):
Apologies, my comment was confusing. In all other mem space allocate() functions we don't include `exec_space` var name in func signature instead of including and then casing `(void)exec_space`.

core/src/OpenACC/Kokkos_OpenACCSpace.hpp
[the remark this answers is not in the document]
seyonglee: Cleaned up as suggested.

core/src/OpenACC/Kokkos_OpenACCSpace.hpp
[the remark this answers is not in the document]
seyonglee: Cleaned up as suggested.

core/src/OpenACC/Kokkos_OpenACC_ParallelReduce_Team.hpp
seyonglee: `KOKKOS_IMPL_ACC_REDUCE_TEAM_MEMBER_INIT` has multiple different definitions depending on the calling contexts.

masterleinad (dismissed):
Fine with me.

janciesko:
@seyonglee - please can you rebase this?

janciesko (approved):
Rebased this thing. LGTM

PR #7356, opened 2024-09-18
Use cmake-format for enforcing uniform formatting in CMake files

masterleinad:
In reponse to https://github.com/kokkos/kokkos/pull/6164#issuecomment-2346657407. Configuration options can be found at https://cmake-format.readthedocs.io/en/latest/configopts.html. The defaults seemed pretty good to me except that I don't see much value in reflowing comments and we seem to use dangling parentheses for multi-line commands.


CMakeLists.txt
```diff
@@ -15,146 +14,169 @@ endif()
...
-      SET(DEFAULT_BUILD_TYPE "RelWithDebInfo")
-      MESSAGE(STATUS "Setting build type to '${DEFAULT_BUILD_TYPE}' as none was specified.")
-      SET(CMAKE_BUILD_TYPE "${DEFAULT_BUILD_TYPE}" CACHE STRING
+set(KOKKOS_CFG_DAG_NONE On) #sentinel to indicate no dependencies
+set(KOKKOS_CFG_DAG_DEVICES_DONE Off)
+set(KOKKOS_CFG_DAG_OPTIONS_DONE Off)
+set(KOKKOS_CFG_DAG_ARCH_DONE Off)
+set(KOKKOS_CFG_DAG_CXX_STD_DONE Off)
+set(KOKKOS_CFG_DAG_COMPILER_ID_DONE Off)
+function(KOKKOS_CFG_DEPENDS SUCCESSOR PRECURSOR)
```
romintomasetti, line 56: I'm not sure that modifying the whole `CMake` base for case is a good idea.

You'll make *git* history much harder when we need to go back to understand things.

I would strongly suggest to run your format check on the new code only.

romintomasetti (changes requested):

CMakeLists.txt
aprokop: > You'll make _git_ history much harder when we need to go back to understand things.

There's a way to avoid it by using `.git-blame-ignore-revs`.

romintomasetti: Yep, but frankly it's just an additional burden. But I will not block, mess with it if you deem it necessary.

nmm0 (commented):
This looks good, I particularly like that commands are formatted lower-case by default

CMakeLists.txt
masterleinad: @romintomasetti Looks like you are currently blocking.

.github/workflows/format-check.yml
```diff
@@ -13,3 +13,9 @@ jobs:
       uses: DoozyX/clang-format-lint-action@c71d0bf4e21876ebec3e5647491186f8797fde31 # v0.18.2
       with:
         clangFormatVersion: 16
+    - name: cmake-format lint action
```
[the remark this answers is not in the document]
masterleinad: Having one workflow for formatting is more concise.

algorithms/unit_tests/CMakeLists.txt
```diff
@@ -23,21 +23,15 @@ foreach(Tag Threads;Serial;OpenMP;Cuda;HPX;HIP;SYCL;OpenMPTarget)
...
-	TestSort
-	TestSortByKey
-	TestSortCustomComp
-	TestBinSortA
-	TestBinSortB
-	TestNestedSort
-      )
+    foreach(SOURCE_Input TestSort TestSortByKey TestSortCustomComp TestBinSortA
+                         TestBinSortB TestNestedSort
+    )
```
pzehner, line 28: I guess it's really a question of taste, but I don't find this one-line list of arguments easy to read.

CMakeLists.txt
romintomasetti: Oh sorry about that !

.github/workflows/format-check.yml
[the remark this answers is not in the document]
romintomasetti: As you wish.

example/tutorial/Algorithms/01_random_numbers/CMakeLists.txt
```diff
@@ -1,9 +1,7 @@
...
+kokkos_include_directories(${CMAKE_CURRENT_BINARY_DIR})
+kokkos_include_directories(${CMAKE_CURRENT_SOURCE_DIR})
 
 # This is a tutorial, not a test, so we don't ask CTest to run it.
-KOKKOS_ADD_EXECUTABLE(
-  tutorial_algorithms_01_random_numbers
-  SOURCES random_numbers.cpp
+kokkos_add_executable(
+  tutorial_algorithms_01_random_numbers SOURCES random_numbers.cpp
 )
```
romintomasetti, line 7: I find the previous version more appealing. I guess the formatting wants you to put it on several lines because otherwise the oneliner is too long ?

cmake/msvc.cmake
```diff
@@ -1,11 +1,21 @@
...
+function(kokkos_set_msvc_flags full_standard int_standard)
+  if(CMAKE_CXX_EXTENSIONS)
+    set(KOKKOS_CXX_STANDARD_FLAG
+        ""
+        PARENT_SCOPE
+    )
+    set(KOKKOS_CXX_INTERMEDIATE_STANDARD_FLAG
+        ""
+        PARENT_SCOPE
+    )
```
romintomasetti, line 10: The previous version was better to me...

bartlettroscoe:
FYI: You might also want to consider lower-casing your CMake commands.  I had to write a tool for this [`lower_case_cmake.py`](https://github.com/TriBITSPub/TriBITS/blob/master/tribits/python_utils/lower_case_cmake.py).  It is not perfect, but it did a pretty good job on TriBITS.

nmm0 (approved):

core/src/CMakeLists.txt
```diff
@@ -1,215 +1,245 @@
...
+  append_glob(
+    KOKKOS_CORE_HEADERS ${KOKKOS_SOURCE_DIR}/tpls/desul/include/desul/*/*.hpp
+  )
+  append_glob(
+    KOKKOS_CORE_HEADERS ${KOKKOS_SOURCE_DIR}/tpls/desul/include/desul/*/*/*.hpp
+  )
+  append_glob(
+    KOKKOS_CORE_HEADERS ${KOKKOS_SOURCE_DIR}/tpls/desul/include/*/*/*.inc*
+  )
+  append_glob(KOKKOS_CORE_HEADERS ${CMAKE_CURRENT_BINARY_DIR}/desul/*.hpp)
```
dalg24, line 130: Is there way to disable the formatting?

core/src/CMakeLists.txt
```diff
@@ -1,215 +1,245 @@
...
+    set(DESUL_ATOMICS_ENABLE_SYCL ON)
+    if(KOKKOS_IMPL_SYCL_DEVICE_GLOBAL_SUPPORTED
+       AND NOT KOKKOS_IMPL_HAVE_SYCL_EXT_ONEAPI_DEVICE_GLOBAL
+    )
+      set(DESUL_ATOMICS_ENABLE_SYCL_SEPARABLE_COMPILATION ON)
+    endif()
+  endif()
+  if(KOKKOS_ENABLE_OPENMPTARGET)
+    set(DESUL_ATOMICS_ENABLE_OPENMP ON
+    )# not a typo Kokkos OpenMPTarget -> Desul OpenMP
```
dalg24, line 28: Does the format reject the space between the closing paren and the comment?
Is that something that can be changed?

dalg24 (commented):
Maybe that is the right time to decide whether we want the copyright header in the cmake files.  @crtrott what do you think?

algorithms/CMakeLists.txt
```diff
@@ -1,7 +1,9 @@
...
+if(NOT Kokkos_INSTALL_TESTING)
+  add_subdirectory(src)
+endif()
 # FIXME_OPENACC: temporarily disabled due to unimplemented features
-IF(NOT ((KOKKOS_ENABLE_OPENMPTARGET AND KOKKOS_CXX_COMPILER_ID STREQUAL NVHPC) OR KOKKOS_ENABLE_OPENACC))
-  KOKKOS_ADD_TEST_DIRECTORIES(unit_tests)
-ENDIF()
+if(NOT ((KOKKOS_ENABLE_OPENMPTARGET AND KOKKOS_CXX_COMPILER_ID STREQUAL NVHPC)
+        OR KOKKOS_ENABLE_OPENACC)
+)
```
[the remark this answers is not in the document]
masterleinad: We either use dangling parentheses for all statements that are wrapped to multiple lines or for none. There are different options where to put the dangling parenthesis.
We can increase the maximum column number to generally mitigate wrapping too much.

core/src/CMakeLists.txt
masterleinad: I think trailing comments are considered part of the command. Not sure what happened here but we can just move the comment to the line above.

core/src/CMakeLists.txt
masterleinad: You can use `# cmake-format: off` and `# cmake-format: on` comments.

cmake/msvc.cmake
cedricchevalier19: I agree.

containers/unit_tests/CMakeLists.txt
```diff
@@ -12,58 +13,55 @@ foreach(Tag Threads;Serial;OpenMP;HPX;Cuda;HIP;SYCL)
...
+      StaticCrsGraph
+      WithoutInitializing
+      UnorderedMap
+      Vector
+      ViewCtorPropEmbeddedDim
+    )
       if(NOT Kokkos_ENABLE_DEPRECATED_CODE_4 AND Name STREQUAL "Vector")
-        continue() # skip Kokkos::vector test if deprecated code 4 is not enabled
+        continue(
+        )# skip Kokkos::vector test if deprecated code 4 is not enabled
```
cedricchevalier19, line 35: This comment tied to the parenthesis and on separate line is difficult to read.

example/tutorial/Algorithms/01_random_numbers/CMakeLists.txt
cedricchevalier19: Ideally, something I would prefer something like:

```cmake
kokkos_add_executable(tutorial_algorithms_01_random_numbers
  SOURCES random_numbers.cpp
)
```

Do you think it can be configured this way?

cedricchevalier19 (commented):
It's a really huge progress to have an automated formatting, thanks @masterleinad.

If you can improve a little the format already mentioned by @romintomasetti and @dalg24 it will be perfect!

masterleinad:
> If you can improve a little the format already mentioned by @romintomasetti and @dalg24 it will be perfect!

I tweaked the settings a little also allowing up to 120 characters in a line.

cedricchevalier19 (approved):
Looks good!

.github/workflows/format-check.yml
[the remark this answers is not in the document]
dalg24: I disagree I don't think these should be steps, I think they should be individual checks (or jobs?  not sure about the right terminology) that we group together as in #7361 (I don't feel too strongly on what PR should go in first)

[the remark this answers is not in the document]
masterleinad: Fair enough.

aprokop:
Should probably add to `.git-blame-ignore-revs`.

PR #7358, opened 2024-09-19
Add CEA weekly test config

pzehner:
This PR aims to extend testing by running Kokkos tests on a weekly basis on the Ruche local cluster, by CEA means. The configuration is minimal for now and consists in a debug and release builds for the Cuda backend, on A100 GPUs.

The workflow is scheduled to run every Saturday at 2 am UTC, or manually.

The self-hosted runner on Ruche checks if the job can be executed. Scheduled runs can only take place on weekend. Manual runs have to be triggered by a white-listed user (I prefer to discuss about the members of the list on nucleus). No other job types (on push, etc.) are allowed.

Registration of the self-hosted runner was done with @crtrott.

The same configuration was tested on a dummy project, so it should run on the first try here...

### DIFF
--- a/.github/workflows/weekly-cea.yml
+++ b/.github/workflows/weekly-cea.yml
@@ -31,6 +31,8 @@ jobs:
           cmake -B build \
             -DCMAKE_BUILD_TYPE=${{ matrix.build_type }} \
             -DCMAKE_CXX_STANDARD=17 \
+            -DCMAKE_CXX_FLAGS=-Werror \
+            -DKokkos_ENABLE_COMPILER_WARNINGS=ON \
             -DKokkos_ENABLE_TESTS=ON \
             -DKokkos_ENABLE_EXAMPLES=ON \
             ${{ matrix.backend.flags }}
--- a/.github/workflows/weekly-cea.yml
+++ b/.github/workflows/weekly-cea.yml
@@ -14,9 +14,8 @@ jobs:
           - Debug
         backend:
           - name: cuda-a100
-            flags:
-              cmake: -DKokkos_ENABLE_CUDA=ON -DKokkos_ARCH_AMPERE80=ON
-              module: gcc/11.2.0/gcc-4.8.5 cuda/12.2.1/gcc-11.2.0 cmake/3.28.3/gcc-11.2.0
+            flags: -DKokkos_ENABLE_CUDA=ON -DKokkos_ARCH_AMPERE80=ON
+            modules: gcc/11.2.0/gcc-4.8.5 cuda/12.2.1/gcc-11.2.0 cmake/3.28.3/gcc-11.2.0
             queue: gpua100
             runner: [self-hosted, cuda]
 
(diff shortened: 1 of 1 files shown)


.github/workflows/weekly-cea.yml
```diff
@@ -0,0 +1,49 @@
...
+            flags: -DKokkos_ENABLE_CUDA=ON -DKokkos_ARCH_AMPERE80=ON
+            modules: gcc/11.2.0/gcc-4.8.5 cuda/12.2.1/gcc-11.2.0 cmake/3.28.3/gcc-11.2.0
+            queue: gpua100
+            runner: [self-hosted, cuda]
+
+    runs-on: ${{ matrix.backend.runner }}
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
```
romintomasetti, line 26:
```suggestion
      - uses: actions/checkout@v4
```
With this you already get meaningful step summary:
image

.github/workflows/weekly-cea.yml
```diff
@@ -0,0 +1,49 @@
...
+            -DCMAKE_CXX_STANDARD=17 \
+            -DCMAKE_CXX_FLAGS=-Werror \
+            -DKokkos_ENABLE_COMPILER_WARNINGS=ON \
+            -DKokkos_ENABLE_TESTS=ON \
+            -DKokkos_ENABLE_EXAMPLES=ON \
+            ${{ matrix.backend.flags }}
+
+      - name: Build
+        run: |
+          module load ${{ matrix.backend.modules }}
```
romintomasetti, line 42: Are you sure you need to load the modules at each job step ? I think you run them all in the same process, so I guess you could load them once in a dedicated step before the `Configure` step. Module loading failures would be easily identified then.

.github/workflows/weekly-cea.yml
pzehner: Right.

.github/workflows/weekly-cea.yml
pzehner: Job steps are executed on the same runner, but [on different processes](https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idsteps), so I cannot load modules for once. It's still possible to set environment variables from one step to the other, but [the process is tedious and has to be done for each of them](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/workflow-commands-for-github-actions#setting-an-environment-variable).

.github/workflows/weekly-cea.yml
```diff
@@ -0,0 +1,51 @@
+name: Weekly CEA CI
```
dalg24, line 1:
```suggestion
name: Weekly CEA builds
```
Technically this is not Continuous Integration :)

dalg24 (approved):

.github/workflows/weekly-cea.yml
pzehner: Right!

masterleinad (approved):
Let's give it a shot!

Issue #7350, opened 2024-09-17
ViewTraits Interface Cleanup


crtrott:
### `ViewTraits` is a helper class for implementing `View` and related classes - it defines many of the primary typedefs of `View`
- `View` publicly derived from `ViewTraits`
- Most containers take all of the traits from `ViewTraits`
- File: `core/src/View/Kokkos_ViewTraits.hpp`
- Designed to handle the quasi variadic nature of `View` template arguments (i.e. `View<int*, SharedSpace>` vs `View<int*, LayoutLeft, SharedSpace>`)

The discussion here is really about typedefs in View - even if we remove ViewTraits as a class eventually - but not in 5.0.

Prototype: 
```c++
template <class DataType, class... Properties>
struct ViewTraits;
```

### 7 `typedefs` are derived from `Properties...`

```c++
execution_space
memory_space
HostMirrorSpace
array_layout
memory_traits
specialize
hooks_policy
```

Propose for deprecation: 
- `HostMirrorSpace` replace by `host_mirror_device` -> we likely still need this, but we probably want it to consistently be a device, note that this typedef is fairly complex 
- `array_layout` replace by `layout_type` -> align with `mdspan`
- `specialize` -> drop this, since we won't allow specializing View internals anymore - and there will be no need for it since we have now standardized customization points such as `accessor`
- `hooks_policy` -> not sure but likely drop this, use `accessor` and friends?

I will use the example of `View<float*[5]` in the following

#### Data type traits
```c++
using data_type           = float*[5];
using const_data_type     = const float*[5];
using non_const_data_type = float*[5];
```
- all are useful since `const data_type` is not equal `const_data_type` necessarily.

#### Compatible array of trivial type traits:
```c++
using scalar_array_type           = float*[5];
using const_scalar_array_type     = const float*[5];
using non_const_scalar_array_type = float*[5];
```
- These only come into play for stuff like Sacado FAD
- We likely can get rid of these, only used in internals of Sacado, downstream users don't seem to use them

#### Value type traits:
```c++
using value_type           = float;
using const_value_type     = const float;
using non_const_value_type = float;
```
- these are somewhat problematic because they conflict with `mdspan`
```
mdspan::element_type <=> View::value_type
mdspan::value_type <=> View::non_const_value_type
```

#### Mapping traits:

```c++
using dimension = ...
static constexpr unsigned rank
static constexpr unsigned rank_dynamic
```
[...]

PR #7360, opened 2024-09-19
Deprecate array_type and scalar_array_type and more enums in ViewTraits

janciesko:
Partially addressing https://github.com/kokkos/kokkos/issues/7350.
Changes:
- `scalar_array_type` -> marked deprecated in favor of `data_type`
- `const_scalar_array_type` -> marked deprecated in favor of `const_data_type`
- `non_const_scalar_array_type` -> marked deprecated in favor of `non_const_data_type`
- `array_type` -> marked deprecated in favor of `type`
-  `type`, `const_type` and `non_const_type` ViewDataAnalysis deprecated in favor of `data_type`, `const_data_type` and `non_const_data_type`.
- `array_type` is deprecated in favor of `uniform_type` in DynamicView
Note: Requires updates to the documentation

### DIFF
--- a/containers/src/Kokkos_DynRankView.hpp
+++ b/containers/src/Kokkos_DynRankView.hpp
@@ -459,9 +459,14 @@ class DynRankView : private View<DataType*******, Properties...> {
   using reference_type = typename view_type::reference_type;
   using pointer_type   = typename view_type::pointer_type;
 
-  using scalar_array_type           = value_type;
-  using const_scalar_array_type     = const_value_type;
-  using non_const_scalar_array_type = non_const_value_type;
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  using scalar_array_type KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use data_type instead.") = data_type;
+  using const_scalar_array_type KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use const_data_type instead.") = const_data_type;
+  using non_const_scalar_array_type KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use non_const_data_type instead.") = non_const_data_type;
+#endif
 #ifndef KOKKOS_ENABLE_IMPL_VIEW_LEGACY
 #ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
   using specialize KOKKOS_DEPRECATED = void;
@@ -500,9 +505,14 @@ class DynRankView : private View<DataType*******, Properties...> {
   // rank 7 data_type of the traits
 
   /** \brief  Compatible view of array of scalar types */
-  using array_type = DynRankView<
-      typename drvtraits::scalar_array_type, typename drvtraits::array_layout,
+  using type = DynRankView<
+      typename drvtraits::data_type, typename drvtraits::array_layout,
       typename drvtraits::device_type, typename drvtraits::memory_traits>;
+
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  /** \brief  Compatible view of array of scalar types */
+  using array_type KOKKOS_DEPRECATED_WITH_COMMENT("Use type instead.") = type;
+#endif
 
   /** \brief  Compatible view of const data type */
   using const_type = DynRankView<
@@ -747,7 +757,7 @@ class DynRankView : private View<DataType*******, Properties...> {
 #ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
   KOKKOS_FUNCTION reference_type operator[](index_type i0) const {
     if constexpr (std::is_same_v<typename drvtraits::value_type,
-                                 typename drvtraits::scalar_array_type>) {
+                                 typename drvtraits::data_type>) {
       return view_type::data()[i0];
     } else {
       const size_t dim_scalar = view_type::impl_map().dimension_scalar();
@@ -1952,7 +1962,7 @@ namespace Experimental {
 template <class T, class... P>
 struct python_view_type<DynRankView<T, P...>> {
   using type = Kokkos::Impl::python_view_type_impl_t<
-      typename DynRankView<T, P...>::array_type>;
+      typename DynRankView<T, P...>::type>;
 };
 }  // namespace Experimental
 
--- a/containers/src/Kokkos_DynRankView.hpp
+++ b/containers/src/Kokkos_DynRankView.hpp
@@ -459,7 +459,7 @@ class DynRankView : private View<DataType*******, Properties...> {
   using reference_type = typename view_type::reference_type;
   using pointer_type   = typename view_type::pointer_type;
 
-#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_5
   using scalar_array_type KOKKOS_DEPRECATED_WITH_COMMENT(
       "Use data_type instead.") = data_type;
   using const_scalar_array_type KOKKOS_DEPRECATED_WITH_COMMENT(
@@ -509,7 +509,7 @@ class DynRankView : private View<DataType*******, Properties...> {
       typename drvtraits::data_type, typename drvtraits::array_layout,
       typename drvtraits::device_type, typename drvtraits::memory_traits>;
 
-#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_5
   /** \brief  Compatible view of array of scalar types */
   using array_type KOKKOS_DEPRECATED_WITH_COMMENT("Use type instead.") = type;
 #endif
--- a/containers/src/Kokkos_DynamicView.hpp
+++ b/containers/src/Kokkos_DynamicView.hpp
@@ -262,10 +262,15 @@ class DynamicView : public Kokkos::ViewTraits<DataType, P...> {
 
  public:
   //----------------------------------------------------------------------
-
-  /** \brief  Compatible view of array of scalar types */
-  using array_type =
+  /** \brief  Compatible view of data type */
+  using uniform_type =
       DynamicView<typename traits::data_type, typename traits::device_type>;
+
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  /** \brief  Compatible view of array of scalar types */
+  using array_type KOKKOS_DEPRECATED_WITH_COMMENT("Use uniform_type instead.") =
+      uniform_type;
+#endif
 
   /** \brief  Compatible view of const data type */
   using const_type = DynamicView<typename traits::const_data_type,
@@ -282,9 +287,8 @@ class DynamicView : public Kokkos::ViewTraits<DataType, P...> {
   using uniform_device =
       Kokkos::Device<typename traits::device_type::execution_space,
                      Kokkos::AnonymousSpace>;
-  using uniform_type               = array_type;
   using uniform_const_type         = const_type;
-  using uniform_runtime_type       = array_type;
+  using uniform_runtime_type       = uniform_type;
   using uniform_runtime_const_type = const_type;
   using uniform_nomemspace_type =
       DynamicView<typename traits::data_type, uniform_device>;
--- a/containers/src/Kokkos_DynamicView.hpp
+++ b/containers/src/Kokkos_DynamicView.hpp
@@ -266,7 +266,7 @@ class DynamicView : public Kokkos::ViewTraits<DataType, P...> {
   using uniform_type =
       DynamicView<typename traits::data_type, typename traits::device_type>;
 
-#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_5
   /** \brief  Compatible view of array of scalar types */
   using array_type KOKKOS_DEPRECATED_WITH_COMMENT("Use uniform_type instead.") =
       uniform_type;
--- a/containers/src/Kokkos_OffsetView.hpp
+++ b/containers/src/Kokkos_OffsetView.hpp
@@ -211,18 +211,15 @@ class OffsetView : public View<DataType, Properties...> {
 
  public:
   //----------------------------------------
-
-#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_5
-  /** \brief  Compatible view of array of scalar types */
-  using array_type KOKKOS_DEPRECATED_WITH_COMMENT("Use type instead.") =
-      OffsetView<typename traits::scalar_array_type,
-                 typename traits::array_layout, typename traits::device_type,
-                 typename traits::memory_traits>;
-#endif
   /** \brief  Compatible view of data type */
   using type =
       OffsetView<typename traits::data_type, typename traits::array_layout,
                  typename traits::device_type, typename traits::memory_traits>;
+
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_5
+  /** \brief  Compatible view of array of scalar types */
+  using array_type KOKKOS_DEPRECATED_WITH_COMMENT("Use type instead.") = type;
+#endif
 
   /** \brief  Compatible view of const data type */
   using const_type =
--- a/containers/src/Kokkos_OffsetView.hpp
+++ b/containers/src/Kokkos_OffsetView.hpp
@@ -211,11 +211,18 @@ class OffsetView : public View<DataType, Properties...> {
 
  public:
   //----------------------------------------
+
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
   /** \brief  Compatible view of array of scalar types */
-  using array_type =
+  using array_type KOKKOS_DEPRECATED_WITH_COMMENT("Use type instead.") =
       OffsetView<typename traits::scalar_array_type,
                  typename traits::array_layout, typename traits::device_type,
                  typename traits::memory_traits>;
+#endif
+  /** \brief  Compatible view of data type */
+  using type =
+      OffsetView<typename traits::data_type, typename traits::array_layout,
+                 typename traits::device_type, typename traits::memory_traits>;
 
   /** \brief  Compatible view of const data type */
   using const_type =
@@ -284,7 +291,7 @@ class OffsetView : public View<DataType, Properties...> {
   // interoperability with View
  private:
   using view_type =
-      View<typename traits::scalar_array_type, typename traits::array_layout,
+      View<typename traits::data_type, typename traits::array_layout,
            typename traits::device_type, typename traits::memory_traits>;
 
  public:
--- a/containers/src/Kokkos_OffsetView.hpp
+++ b/containers/src/Kokkos_OffsetView.hpp
@@ -212,7 +212,7 @@ class OffsetView : public View<DataType, Properties...> {
  public:
   //----------------------------------------
 
-#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_5
   /** \brief  Compatible view of array of scalar types */
   using array_type KOKKOS_DEPRECATED_WITH_COMMENT("Use type instead.") =
       OffsetView<typename traits::scalar_array_type,
--- a/containers/unit_tests/TestDynRankViewTypedefs.cpp
+++ b/containers/unit_tests/TestDynRankViewTypedefs.cpp
@@ -55,7 +55,7 @@ constexpr bool test_view_typedefs_impl() {
   static_assert(std::is_same_v<typename ViewType::const_data_type, typename data_analysis<DataType>::const_data_type>);
   static_assert(std::is_same_v<typename ViewType::non_const_data_type, typename data_analysis<DataType>::non_const_data_type>);
   
-  #ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  #ifdef KOKKOS_ENABLE_DEPRECATED_CODE_5
   // FIXME: these should be deprecated and for proper testing (I.e. where this is different from data_type)
   // we would need ensemble types which use the hidden View dimension facility of View (i.e. which make
   // "specialize" not void)
@@ -65,7 +65,7 @@ constexpr bool test_view_typedefs_impl() {
   static_assert(std::is_same_v<typename ViewType::non_const_scalar_array_type, typename data_analysis<DataType>::non_const_data_type>);
   KOKKOS_IMPL_DISABLE_DEPRECATED_WARNINGS_POP()
   #endif
-  #ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  #ifdef KOKKOS_ENABLE_DEPRECATED_CODE_5
   static_assert(std::is_same_v<typename ViewType::specialize, void>);
   #endif
 
@@ -98,7 +98,7 @@ constexpr bool test_view_typedefs_impl() {
 
   // FIXME: in contrast to View, hooks_policy is not propagated
   static_assert(std::is_same_v<typename ViewType::traits, ViewTraitsType>);
-#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_5
 KOKKOS_IMPL_DISABLE_DEPRECATED_WARNINGS_PUSH()
   static_assert(std::is_same_v<typename ViewType::array_type,
                                Kokkos::DynRankView<typename ViewType::scalar_array_type, typename ViewType::array_layout,
--- a/containers/unit_tests/TestDynRankViewTypedefs.cpp
+++ b/containers/unit_tests/TestDynRankViewTypedefs.cpp
@@ -54,16 +54,20 @@ constexpr bool test_view_typedefs_impl() {
   static_assert(std::is_same_v<typename ViewType::data_type, DataType>);
   static_assert(std::is_same_v<typename ViewType::const_data_type, typename data_analysis<DataType>::const_data_type>);
   static_assert(std::is_same_v<typename ViewType::non_const_data_type, typename data_analysis<DataType>::non_const_data_type>);
-
+  
+  #ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
   // FIXME: these should be deprecated and for proper testing (I.e. where this is different from data_type)
   // we would need ensemble types which use the hidden View dimension facility of View (i.e. which make
   // "specialize" not void)
+  KOKKOS_IMPL_DISABLE_DEPRECATED_WARNINGS_PUSH()
   static_assert(std::is_same_v<typename ViewType::scalar_array_type, DataType>);
   static_assert(std::is_same_v<typename ViewType::const_scalar_array_type, typename data_analysis<DataType>::const_data_type>);
   static_assert(std::is_same_v<typename ViewType::non_const_scalar_array_type, typename data_analysis<DataType>::non_const_data_type>);
-#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  KOKKOS_IMPL_DISABLE_DEPRECATED_WARNINGS_POP()
+  #endif
+  #ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
   static_assert(std::is_same_v<typename ViewType::specialize, void>);
-#endif
+  #endif
 
   // FIXME: value_type definition conflicts with mdspan value_type
   static_assert(std::is_same_v<typename ViewType::value_type, ValueType>);
@@ -94,7 +98,15 @@ constexpr bool test_view_typedefs_impl() {
 
   // FIXME: in contrast to View, hooks_policy is not propagated
   static_assert(std::is_same_v<typename ViewType::traits, ViewTraitsType>);
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+KOKKOS_IMPL_DISABLE_DEPRECATED_WARNINGS_PUSH()
   static_assert(std::is_same_v<typename ViewType::array_type,
+                               Kokkos::DynRankView<typename ViewType::scalar_array_type, typename ViewType::array_layout,
+                                            typename ViewType::device_type, //typename ViewTraitsType::hooks_policy,
+                                            typename ViewType::memory_traits>>);
+KOKKOS_IMPL_DISABLE_DEPRECATED_WARNINGS_POP()
+#endif
+  static_assert(std::is_same_v<typename ViewType::type,
                                Kokkos::DynRankView<typename ViewType::data_type, typename ViewType::array_layout,
                                             typename ViewType::device_type, //typename ViewTraitsType::hooks_policy,
                                             typename ViewType::memory_traits>>);
--- a/containers/unit_tests/TestDynRankViewTypedefs.cpp
+++ b/containers/unit_tests/TestDynRankViewTypedefs.cpp
@@ -65,7 +65,7 @@ constexpr bool test_view_typedefs_impl() {
   static_assert(std::is_same_v<typename ViewType::non_const_scalar_array_type, typename data_analysis<DataType>::non_const_data_type>);
   KOKKOS_IMPL_DISABLE_DEPRECATED_WARNINGS_POP()
   #endif
-  #ifdef KOKKOS_ENABLE_DEPRECATED_CODE_5
+  #ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
   static_assert(std::is_same_v<typename ViewType::specialize, void>);
   #endif
 
--- a/core/src/Kokkos_View.hpp
+++ b/core/src/Kokkos_View.hpp
@@ -193,21 +193,30 @@ class View : public Impl::BasicViewFromTraits<DataType, Properties...>::type {
   using raw_allocation_value_type = std::remove_pointer_t<pointer_type>;
 
  public:
-  using scalar_array_type       = typename traits::scalar_array_type;
-  using const_scalar_array_type = typename traits::const_scalar_array_type;
-  using non_const_scalar_array_type =
-      typename traits::non_const_scalar_array_type;
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  using scalar_array_type KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use data_type instead.") = data_type;
+  using const_scalar_array_type KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use const_data_type instead.") = const_data_type;
+  using non_const_scalar_array_type KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use non_const_data_type instead.") = non_const_data_type;
+#endif
 
   // typedefs from BasicView
   using typename base_t::mdspan_type;
   using reference_type = typename base_t::reference;
 
   //----------------------------------------
+  // Compatible view of a data type
+  using type = View<typename traits::data_type, typename traits::array_layout,
+                    typename traits::device_type, typename traits::hooks_policy,
+                    typename traits::memory_traits>;
+
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  //----------------------------------------
   // Compatible view of array of scalar types
-  using array_type =
-      View<typename traits::scalar_array_type, typename traits::array_layout,
-           typename traits::device_type, typename traits::hooks_policy,
-           typename traits::memory_traits>;
+  using array_type KOKKOS_DEPRECATED_WITH_COMMENT("Use type instead.") = type;
+#endif
 
   // Compatible view of const data type
   using const_type =
--- a/core/src/Kokkos_View.hpp
+++ b/core/src/Kokkos_View.hpp
@@ -193,7 +193,7 @@ class View : public Impl::BasicViewFromTraits<DataType, Properties...>::type {
   using raw_allocation_value_type = std::remove_pointer_t<pointer_type>;
 
  public:
-#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_5
   using scalar_array_type KOKKOS_DEPRECATED_WITH_COMMENT(
       "Use data_type instead.") = data_type;
   using const_scalar_array_type KOKKOS_DEPRECATED_WITH_COMMENT(
@@ -212,7 +212,7 @@ class View : public Impl::BasicViewFromTraits<DataType, Properties...>::type {
                     typename traits::device_type, typename traits::hooks_policy,
                     typename traits::memory_traits>;
 
-#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_5
   //----------------------------------------
   // Compatible view of array of scalar types
   using array_type KOKKOS_DEPRECATED_WITH_COMMENT("Use type instead.") = type;
--- a/core/src/View/Kokkos_ViewCommonType.hpp
+++ b/core/src/View/Kokkos_ViewCommonType.hpp
@@ -26,7 +26,7 @@ struct CommonViewAllocProp<void, ValueType> {
   using value_type = ValueType;
   using data_type  = ValueType;
 
-#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_5
   using scalar_array_type KOKKOS_DEPRECATED_WITH_COMMENT(
       "Use data_type instead.") = data_type;
 #endif
--- a/core/src/View/Kokkos_ViewCommonType.hpp
+++ b/core/src/View/Kokkos_ViewCommonType.hpp
@@ -23,8 +23,13 @@ struct CommonViewAllocProp;
 
 template <class ValueType>
 struct CommonViewAllocProp<void, ValueType> {
-  using value_type        = ValueType;
-  using scalar_array_type = ValueType;
+  using value_type = ValueType;
+  using data_type  = ValueType;
+
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  using scalar_array_type KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use data_type instead.") = data_type;
+#endif
 
   template <class... Views>
   KOKKOS_INLINE_FUNCTION CommonViewAllocProp(const Views&...) {}
--- a/core/src/View/Kokkos_ViewDataAnalysis.hpp
+++ b/core/src/View/Kokkos_ViewDataAnalysis.hpp
@@ -374,7 +374,7 @@ struct ViewDataAnalysis {
   using non_const_data_type =
       typename ViewDataType<non_const_value_type, dimension>::type;
 
-#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_5
   using type KOKKOS_DEPRECATED_WITH_COMMENT("Use data_type instead.") =
       data_type;
   using const_type KOKKOS_DEPRECATED_WITH_COMMENT(
@@ -383,7 +383,7 @@ struct ViewDataAnalysis {
       "Use non_const_data_type instead.") = non_const_data_type;
 #endif
 
-#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_5
   // Generate "flattened" multidimensional array specification type.
   using scalar_array_type KOKKOS_DEPRECATED_WITH_COMMENT(
       "Use data_type instead.") = data_type;
--- a/core/src/View/Kokkos_ViewDataAnalysis.hpp
+++ b/core/src/View/Kokkos_ViewDataAnalysis.hpp
@@ -381,9 +381,7 @@ struct ViewDataAnalysis {
       "Use const_data_type instead.") = const_data_type;
   using non_const_type KOKKOS_DEPRECATED_WITH_COMMENT(
       "Use non_const_data_type instead.") = non_const_data_type;
-#endif
 
-#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_5
   // Generate "flattened" multidimensional array specification type.
   using scalar_array_type KOKKOS_DEPRECATED_WITH_COMMENT(
       "Use data_type instead.") = data_type;
--- a/core/src/View/Kokkos_ViewDataAnalysis.hpp
+++ b/core/src/View/Kokkos_ViewDataAnalysis.hpp
@@ -368,15 +368,30 @@ struct ViewDataAnalysis {
   using non_const_value_type = typename array_analysis::non_const_value_type;
 
   // Generate analogous multidimensional array specification type.
-  using type       = typename ViewDataType<value_type, dimension>::type;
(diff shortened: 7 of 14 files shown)


janciesko:
Todo

- [ ] ```HostMirrorSpace``` and ```host_mirror_space``` replace by ```host_mirror_device``` -> we likely still need this, but we probably want it to consistently be a device, note that this typedef is fairly complex

- [ ] ```array_layout``` replace by ```layout_type``` -> align with mdspan

- [ ] ```specialize```-> drop this, since we won't allow specializing View internals anymore - and there will be no need for it since we have now standardized customization points such as accessor

- [ ] ```hooks_policy``` -> not sure but likely drop this, use accessor and friends?

- [ ] ```mdspan::element_type``` <=> ```View::value_type```

- [ ] ```mdspan::value_type``` <=> ```View::non_const_value_type```

- [ ] ```using size_type = typename memory_space::size_type;```

containers/src/Kokkos_OffsetView.hpp
```diff
@@ -220,9 +220,8 @@ class OffsetView : public View<DataType, Properties...> {
   //----------------------------------------
   /** \brief  Compatible view of array of scalar types */
   using array_type =
-      OffsetView<typename traits::scalar_array_type,
-                 typename traits::array_layout, typename traits::device_type,
-                 typename traits::memory_traits>;
+      OffsetView<typename traits::data_type, typename traits::array_layout,
+                 typename traits::device_type, typename traits::memory_traits>;
```
crtrott, line 224: This is wrong. The typedef itself needs to be deprecated, not changed.

containers/src/Kokkos_OffsetView.hpp
```diff
@@ -300,7 +299,7 @@ class OffsetView : public View<DataType, Properties...> {
   // interoperability with View
  private:
   using view_type =
-      View<typename traits::scalar_array_type, typename traits::array_layout,
+      View<typename traits::data_type, typename traits::array_layout,
```
crtrott, line 302: this was a bug, and the change needs to happen irrespective deprecation.

core/src/View/Kokkos_ViewLegacy.hpp
```diff
@@ -246,9 +246,9 @@ class View : public ViewTraits<DataType, Properties...> {
 
  public:
   //----------------------------------------
-  /** \brief  Compatible view of array of scalar types */
+  /** \brief  Compatible view of array of data types */
   using array_type =
-      View<typename traits::scalar_array_type, typename traits::array_layout,
+      View<typename traits::data_type, typename traits::array_layout,
            typename traits::device_type, typename traits::hooks_policy,
            typename traits::memory_traits>;
```
crtrott, line 253: don't change the typed: deprecate it.

core/unit_test/TestViewTypedefs.cpp
```diff
@@ -92,7 +95,7 @@ constexpr bool test_view_typedefs_impl() {
   // =========================================
   static_assert(std::is_same_v<typename ViewType::traits, ViewTraitsType>);
   static_assert(std::is_same_v<typename ViewType::array_type,
-                               Kokkos::View<typename ViewType::scalar_array_type, typename ViewType::array_layout,
+                               Kokkos::View<typename ViewType::data_type, typename ViewType::array_layout,
```
crtrott, line 98: This needs to be changed back with deprecation guard, when the "array_type" typedef is changed back and fixed.

example/tutorial/Advanced_Views/04_dualviews/dual_view.cpp
```diff
@@ -61,10 +61,9 @@ struct localsum {
   // double**.
   Kokkos::View<idx_type::const_data_type, idx_type::array_layout, memory_space>
       idx;
-  // "scalar_array_type" is an alias in ViewTraits (and DualView) which is the
+  // "data_type" is an alias in ViewTraits (and DualView) which is the
   // array version of the value(s) stored in the View.
-  Kokkos::View<view_type::scalar_array_type, view_type::array_layout,
-               memory_space>
+  Kokkos::View<view_type::data_type, view_type::array_layout, memory_space>
```
crtrott, line 66: This usage of scalar_array_type was nonsensical to begin with. So this change is fine.

crtrott (changes requested):

core/unit_test/TestViewTypedefs.cpp
```diff
@@ -91,10 +94,16 @@ constexpr bool test_view_typedefs_impl() {
   // in Legacy View: some helper View variants
   // =========================================
   static_assert(std::is_same_v<typename ViewType::traits, ViewTraitsType>);
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
   static_assert(std::is_same_v<typename ViewType::array_type,
                                Kokkos::View<typename ViewType::scalar_array_type, typename ViewType::array_layout,
                                             typename ViewType::device_type, typename ViewTraitsType::hooks_policy,
                                             typename ViewType::memory_traits>>);
+#endif
+  static_assert(std::is_same_v<typename ViewType::type,
```
crtrott, line 103:
```suggestion
  static_assert(std::is_same_v<typename ViewType::array_type,
```

crtrott (commented):
Almost there in my mind. We need to get broad agreement that this is the thing to do. Also note: besides deprecation this uses a ::type in View and the containers, which wasn't there before. Essentially it's the self type explicitly spelled out. It can be used in places where folks mistakenly used `array_type` before.

core/unit_test/TestViewTypedefs.cpp
janciesko: That thing is deprecated and we are testing for it in L98

containers/src/Kokkos_DynRankView.hpp
```diff
@@ -449,10 +449,15 @@ class DynRankView : private View<DataType*******, Properties...> {
...
-  using const_scalar_array_type     = const_value_type;
-  using non_const_scalar_array_type = non_const_value_type;
-  using specialize                  = typename view_type::specialize;
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  using scalar_array_type KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use data_type instead .") = value_type;
+  using const_scalar_array_type KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use const_data_type instead .") = const_value_type;
+  using non_const_scalar_array_type KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use non_const_data_type instead .") = non_const_value_type;
```
masterleinad, line 458: It's surprising to use a different alias than the one suggested in the deprecation warning. Can you clarify what's going on?
Was this just totally unconforming to the `View` example in https://github.com/kokkos/kokkos/issues/7350#issue-2532348150.

containers/src/Kokkos_DynRankView.hpp
```diff
@@ -478,9 +483,17 @@ class DynRankView : private View<DataType*******, Properties...> {
...
+  using array_type KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use Kokkos_DynRankView::type insted.") =
+      DynRankView<
+          typename drvtraits::data_type, typename drvtraits::array_layout,
+          typename drvtraits::device_type, typename drvtraits::memory_traits>;
+#endif
+  /** \brief  Compatible view of data type */
+  using type = DynRankView<
+      typename drvtraits::data_type, typename drvtraits::array_layout,
       typename drvtraits::device_type, typename drvtraits::memory_traits>;
```
masterleinad, line 497: Is this just supposed to be an alias to the current class type? Why do we need this?

containers/src/Kokkos_DynamicView.hpp
```diff
@@ -263,12 +263,17 @@ class DynamicView : public Kokkos::ViewTraits<DataType, P...> {
...
   /** \brief  Compatible view of array of scalar types */
-  using array_type =
+  using array_type KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use Kokkos_DynamicView::type insted.") =
+      DynamicView<typename traits::scalar_array_type,
+                  typename traits::device_type>;
+#endif
+  /** \brief  Compatible view of data type */
+  using type =
       DynamicView<typename traits::data_type, typename traits::device_type>;
```
masterleinad, line 276: Same here.

containers/src/Kokkos_DynRankView.hpp
janciesko: I understand your point. I think the original code was wrong and should have been data_type (vs. value_type). So we suggest the use of the correct type which is the data_type.

masterleinad: ...but we are still using the wrong type?

containers/src/Kokkos_DynRankView.hpp
janciesko: Yes. It replaces a public type so having an equivalent public replacement might be good.

containers/src/Kokkos_DynamicView.hpp
janciesko: There is a use further down in L287

containers/src/Kokkos_DynRankView.hpp
janciesko: Yes. This would be a drive-by change.

masterleinad: So right now we have
```C++
using non_const_value_type = float;
using non_const_scalar_array_type = float*[5];
using non_const_data_type = float*[5];
```
but we want people to use
```C++
float*[5];
```
although this is currently
```C++
float
```
?

janciesko: Nop, the deprecation comment should recommend value_type, is that what you have in mind?

masterleinad: The current definiton should match the recommendation so `value_type` seems more appropriate than `data_type`.

nliber (approved):

containers/src/Kokkos_DynRankView.hpp
dalg24, line 458:
```suggestion
  using scalar_array_type KOKKOS_DEPRECATED_WITH_COMMENT(
      "Use value_type instead.") = value_type;
  using const_scalar_array_type KOKKOS_DEPRECATED_WITH_COMMENT(
      "Use const_value_type instead .") = const_value_type;
  using non_const_scalar_array_type KOKKOS_DEPRECATED_WITH_COMMENT(
      "Use non_const_value_type instead.") = non_const_value_type;
```
What Daniel said.  I don't care if you keep the dot but the space before must go

containers/src/Kokkos_DynRankView.hpp
```diff
@@ -478,9 +483,17 @@ class DynRankView : private View<DataType*******, Properties...> {
   // Types below - at least the HostMirror requires the value_type, NOT the rank
   // 7 data_type of the traits
 
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
   /** \brief  Compatible view of array of scalar types */
-  using array_type = DynRankView<
-      typename drvtraits::scalar_array_type, typename drvtraits::array_layout,
+  using array_type KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use Kokkos_DynRankView::type insted.") =
```
dalg24, line 489:
```suggestion
      "Use type instead.") =
```

containers/src/Kokkos_DynamicView.hpp
```diff
@@ -263,12 +263,17 @@ class DynamicView : public Kokkos::ViewTraits<DataType, P...> {
...
 
  public:
-  //----------------------------------------------------------------------
-
+//----------------------------------------------------------------------
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
   /** \brief  Compatible view of array of scalar types */
-  using array_type =
+  using array_type KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use Kokkos_DynamicView::type insted.") =
```
dalg24, line 270:
```suggestion
      "Use type instead.") =
```

containers/src/Kokkos_OffsetView.hpp
```diff
@@ -218,11 +218,19 @@ class OffsetView : public View<DataType, Properties...> {
 
  public:
   //----------------------------------------
+
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
   /** \brief  Compatible view of array of scalar types */
-  using array_type =
+  using array_type KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use Kokkos_OffsetView::type insted.") =
```
dalg24, line 225:
```suggestion
      "Use type instead.") =
```

containers/src/Kokkos_DynRankView.hpp
dalg24: Or `Kokkos::DynRankView::type` but the prefix is superfluous.  Not the inste[a]d typos.

core/src/View/Kokkos_ViewDataAnalysis.hpp
```diff
@@ -386,10 +386,15 @@ struct ViewDataAnalysis {
...
   // Generate "flattened" multidimensional array specification type.
-  using scalar_array_type           = type;
-  using const_scalar_array_type     = const_type;
-  using non_const_scalar_array_type = non_const_type;
+  using scalar_array_type KOKKOS_DEPRECATED_WITH_COMMENT("Use type instead .") =
+      type;
+  using const_scalar_array_type KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use const_type instead .") = const_type;
+  using non_const_scalar_array_type KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use non_const_type instead .") = non_const_type;
```
dalg24, line 396:
```suggestion
  using scalar_array_type KOKKOS_DEPRECATED_WITH_COMMENT("Use type instead.") =
      type;
  using const_scalar_array_type KOKKOS_DEPRECATED_WITH_COMMENT(
      "Use const_type instead.") = const_type;
  using non_const_scalar_array_type KOKKOS_DEPRECATED_WITH_COMMENT(
      "Use non_const_type instead.") = non_const_type;
```

core/src/View/Kokkos_ViewLegacy.hpp
```diff
@@ -246,12 +246,17 @@ class View : public ViewTraits<DataType, Properties...> {
 
  public:
   //----------------------------------------
-  /** \brief  Compatible view of array of scalar types */
-  using array_type =
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  /** \brief  Compatible view of array of data types */
+  using array_type KOKKOS_DEPRECATED_WITH_COMMENT("Use Kokkos::type insted.") =
```
dalg24, line 251:
```suggestion
  using array_type KOKKOS_DEPRECATED_WITH_COMMENT("Use type instead.") =
```

core/src/View/Kokkos_ViewLegacy.hpp
```diff
@@ -1506,9 +1511,10 @@ struct CommonViewAllocProp;
 
 template <class ValueType>
 struct CommonViewAllocProp<void, ValueType> {
-  using value_type        = ValueType;
+  using value_type = ValueType;
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
   using scalar_array_type = ValueType;
```
dalg24, line 1516: Is it observable?  Used anywhere?

core/src/View/Kokkos_ViewTraits.hpp
```diff
@@ -366,15 +366,18 @@ struct ViewTraits {
...
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
   //------------------------------------
   // Compatible array of trivial type traits:
-
-  using scalar_array_type = typename data_analysis::scalar_array_type;
-  using const_scalar_array_type =
+  using scalar_array_type KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use data_type instead .") = typename data_analysis::type;
+  using const_scalar_array_type KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use const_data_type instead .") =
```
dalg24, line 375:
```suggestion
  using scalar_array_type KOKKOS_DEPRECATED_WITH_COMMENT(
      "Use data_type instead.") = typename data_analysis::type;
  using const_scalar_array_type KOKKOS_DEPRECATED_WITH_COMMENT(
      "Use const_data_type instead.") =
```

core/src/View/Kokkos_ViewTraits.hpp
```diff
@@ -366,15 +366,18 @@ struct ViewTraits {
...
       typename data_analysis::const_scalar_array_type;
-  using non_const_scalar_array_type =
+  using non_const_scalar_array_type KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use non_const_data_type instead .") =
```
dalg24, line 378:
```suggestion
  using non_const_scalar_array_type KOKKOS_DEPRECATED_WITH_COMMENT(
      "Use non_const_data_type instead.") =
```

core/unit_test/TestViewCtorPropEmbeddedDim.hpp
```diff
@@ -91,7 +94,7 @@ struct TestViewCtorProp_EmbeddedDim {
         printf("Proper common value_type\n");
       }
       else {
-        printf("WRONG common value_type\n");
+        printf("Wrong common value_type\n");
```
dalg24, line 97: Pointless change in a commented block

core/src/View/Kokkos_ViewLegacy.hpp
janciesko: Looking at Trilinos, the Sacado package uses that type. Changes to handle that on their side would be trivial.

janciesko:
Should be good now.

core/src/View/Kokkos_ViewLegacy.hpp
dalg24: I meant is that public, is there a legal way for users to refer to that type.

core/src/View/Kokkos_ViewTraits.hpp
```diff
@@ -408,9 +411,18 @@ struct ViewTraits {
 
   using size_type = typename MemorySpace::size_type;
 
-  enum { is_hostspace = std::is_same_v<MemorySpace, HostSpace> };
   enum { is_managed = MemoryTraits::is_unmanaged == 0 };
```
dalg24, line 414: Why is is_managed treated differently?

core/unit_test/TestViewMapping_a.hpp
```diff
@@ -671,11 +668,13 @@ void test_view_mapping() {
     ASSERT_TRUE((std::is_same_v<typename T::const_data_type, const int*>));
     ASSERT_TRUE((std::is_same_v<typename T::non_const_data_type, int*>));
 
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
     ASSERT_TRUE((std::is_same_v<typename T::scalar_array_type, int*>));
```
dalg24, line 672: Why aren't we getting deprecated warnings and forced to `KOKKOS_IMPL_DISABLE_DEPRECATED_WARNINGS_{POP,PUSH}()`?

core/unit_test/TestViewTypedefs.cpp
```diff
@@ -55,13 +55,16 @@ constexpr bool test_view_typedefs_impl() {
   static_assert(std::is_same_v<typename ViewType::data_type, DataType>);
   static_assert(std::is_same_v<typename ViewType::const_data_type, typename data_analysis<DataType>::const_data_type>);
   static_assert(std::is_same_v<typename ViewType::non_const_data_type, typename data_analysis<DataType>::non_const_data_type>);
-  
+
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
```
dalg24, line 59: Same question about deprecated warnings

dalg24 (commented):
What is the deal with the TODO list in https://github.com/kokkos/kokkos/pull/7360#issuecomment-2368926373
most of the items have not been addressed.

core/unit_test/TestViewCtorPropEmbeddedDim.hpp
dalg24: Ping

dalg24: Undo that change or justify it

containers/src/Kokkos_DynRankView.hpp
masterleinad: Why are you deprecating it then? Does it conflict with something?
If it's used internally, does it need to be public?

core/src/View/Kokkos_ViewLegacy.hpp
janciesko: Yes. In practice nobody probably does. The automatic differentiation thingy in Trilinos provides a specialized `CommonViewAllocProp` where they then access the `value_type`. The deprecated code also seems to be wrong already as it assigned ValueType to scalar_array_type.

janciesko:
> What is the deal with the TODO list in [#7360 (comment)](https://github.com/kokkos/kokkos/pull/7360#issuecomment-2368926373) most of the items have not been addressed.

Decided to split this into multiple PRs.

core/unit_test/TestViewTypedefs.cpp
janciesko: Good point - I think we do, https://cloud.cees.ornl.gov/jenkins-ci/job/Kokkos/job/PR-7360/9/pipeline-console/?selected-node=151
I think I need to add those macros.

janciesko:
Retest this please.

core/src/View/Kokkos_ViewDataAnalysis.hpp
```diff
@@ -386,10 +386,15 @@ struct ViewDataAnalysis {
...
-  using scalar_array_type           = type;
-  using const_scalar_array_type     = const_type;
-  using non_const_scalar_array_type = non_const_type;
+  using scalar_array_type KOKKOS_DEPRECATED_WITH_COMMENT("Use type instead.") =
+      type;
+  using const_scalar_array_type KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use const_type instead.") = const_type;
+  using non_const_scalar_array_type KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use non_const_type instead.") = non_const_type;
+#endif
```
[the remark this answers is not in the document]
janciesko: We're accessing these members in public deprecated code in Kokkos_ViewTraits so we need to keep them around until the deprecated code is removed.

core/src/View/Kokkos_ViewTraits.hpp
```diff
@@ -408,9 +411,18 @@ struct ViewTraits {
...
+
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  enum {
+    is_hostspace KOKKOS_DEPRECATED = std::is_same_v<MemorySpace, HostSpace>
+  };
+  enum {
+    is_random_access KOKKOS_DEPRECATED_WITH_COMMENT(
+        "Use memory_traits::is_random_access instead.") =
+        MemoryTraits::is_random_access == 1
+  };
```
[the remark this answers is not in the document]
janciesko: For the sake of this PR we do the deprecations only.

dalg24:
I expect this fails now that we merged https://github.com/kokkos/kokkos/pull/7440

janciesko:
Maybe. I didn't get the usecase for the triple stacked macros in #7440 still.

janciesko:
@masterleinad Like this?https://github.com/kokkos/kokkos/blob/6501d34f731558ef171f84a05da13ad1acb47b51/core/src/View/Kokkos_ViewLegacy.hpp#L257-L280

I wonder if the macro works on all platforms. See here: https://cloud.cees.ornl.gov/jenkins-ci/job/Kokkos/job/PR-7360/14/pipeline-console/?selected-node=151

masterleinad:
> I wonder if the macro works on all platforms. See here: [cloud.cees.ornl.gov/jenkins-ci/job/Kokkos/job/PR-7360/14/pipeline-console?selected-node=151](https://cloud.cees.ornl.gov/jenkins-ci/job/Kokkos/job/PR-7360/14/pipeline-console/?selected-node=151)

We fixed that in https://github.com/kokkos/kokkos/pull/7467/commits/8941659645a7edd4d68e33a03e74fd317f70ed34.

masterleinad:
> @masterleinad Like this?

Yes. We decided to also guard with `KOKKOS_ENABLE_DEPRECATION_WARNINGS` (which make sure `KOKKOS_DEPRECATED` expands to `[[deprecated]]`) so that we would still catch other deprecation warnings if only ours are deactivated.

janciesko:
Ok, I've rebased this.

masterleinad:
> Ok, I've rebased this.

Looks like you last pushed on October 15.

containers/src/Kokkos_DynRankView.hpp
masterleinad: ping

containers/src/Kokkos_DynamicView.hpp
masterleinad: Then you can declare `uniform_type` accordingly and void declaring a `type` alias.

containers/src/Kokkos_OffsetView.hpp
masterleinad: same here.

janciesko:
Retest this please.

core/src/View/Kokkos_ViewLegacy.hpp
```diff
@@ -246,12 +246,19 @@ class View : public ViewTraits<DataType, Properties...> {
...
-  using array_type =
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  /** \brief  Compatible view of array of data types */
+  using array_type KOKKOS_DEPRECATED_WITH_COMMENT("Use type instead.") =
       View<typename traits::scalar_array_type, typename traits::array_layout,
            typename traits::device_type, typename traits::hooks_policy,
            typename traits::memory_traits>;
+#endif
 
+  KOKKOS_IMPL_DISABLE_DEPRECATED_WARNINGS_PUSH()
```
dalg24, line 257: What warnings are we ignoring?!
I am surprised this is needed.

janciesko: In this block we're ignoring the warnings that come from using ViewTraits that include the deprecated stuff when defining the View type in L259.

janciesko: That said, I agree that the macros should ideally be in the unit tests and not here. I still see CI test errors (https://cloud.cees.ornl.gov/jenkins-ci/job/Kokkos/job/PR-7360/18/pipeline-graph/) coming from the unit tests, any idea why they still occur? I feel like I've added the right macros.

janciesko:
Retest this please.

janciesko:
Rebasing

containers/src/Kokkos_DynRankView.hpp
janciesko: Hi @masterleinad, sorry for the delay. We use the alias as in https://github.com/kokkos/kokkos/blob/855db509a1e5632d9291023fcceb134fc3263277/core/unit_test/TestViewTypedefs.cpp#L108

containers/src/Kokkos_DynRankView.hpp
```diff
@@ -485,9 +490,16 @@ class DynRankView : private View<DataType*******, Properties...> {
...
 
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
   /** \brief  Compatible view of array of scalar types */
-  using array_type = DynRankView<
-      typename drvtraits::scalar_array_type, typename drvtraits::array_layout,
+  using array_type KOKKOS_DEPRECATED_WITH_COMMENT("Use type instead.") =
+      DynRankView<
+          typename drvtraits::data_type, typename drvtraits::array_layout,
+          typename drvtraits::device_type, typename drvtraits::memory_traits>;
+#endif
```
dalg24, line 499: Move this definition below and do
```C++
 using array_type KOKKOS_DEPRECATED_WITH_COMMENT("Use type instead.") = type;
```
which more clearly established they are the same as compared to repeating the definition and excepting the reader to compare them.

core/src/View/Kokkos_ViewTraits.hpp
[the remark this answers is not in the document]
dalg24: These are not covered in the description.  Did you mean to include that change set in this PR?

containers/src/Kokkos_DynRankView.hpp
janciesko: Updated.

science-enthusiast:
`array_layout` appears in many changes that are to be made in this PR. [7401](https://github.com/kokkos/kokkos/pull/7401) is taking care of `array_layout` and has progressed quite well. That could be merged soon. Then this PR can be worked on, after rebasing.

containers/src/Kokkos_DynRankView.hpp
```diff
@@ -479,9 +479,14 @@ class DynRankView : private View<DataType*******, Properties...> {
   using reference_type = typename view_type::reference_type;
   using pointer_type   = typename view_type::pointer_type;
 
-  using scalar_array_type           = value_type;
-  using const_scalar_array_type     = const_value_type;
-  using non_const_scalar_array_type = non_const_value_type;
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
```
science-enthusiast: Here and in other places:
`KOKKOS_ENABLE_DEPRECATED_CODE_4` --> `KOKKOS_ENABLE_DEPRECATED_CODE_5`

containers/src/Kokkos_DynRankView.hpp
science-enthusiast: Apart from the tests in `containers/unit_tests/TestDynRankViewTypedefs.cpp` and `core/unit_test/TestViewTypedefs.cpp`, `DynRankView::array_type` is used internally here:
https://github.com/kokkos/kokkos/blob/035744eb210c3e89d158912642d1f52cfdcb2213/containers/src/Kokkos_DynRankView.hpp#L1951-L1957
and maybe here:
https://github.com/kokkos/kokkos/blob/035744eb210c3e89d158912642d1f52cfdcb2213/core/src/KokkosExp_InterOp.hpp#L74-L76

Most likely `array_type` may be hardly used publicly, indeed but I agree with @janciesko's reasoning, there is a public member type called `type` in many parts of Kokkos.

The change `array_type` to `type` is being made uniformly everywhere: `containers/src/Kokkos_DynamicView.hpp`, `containers/src/Kokkos_OffsetView.hpp` and `core/src/Kokkos_View.hpp`.

core/src/Kokkos_View.hpp
```diff
@@ -196,21 +196,28 @@ class View : public Impl::BasicViewFromTraits<DataType, Properties...>::type {
   using raw_allocation_value_type = std::remove_pointer_t<pointer_type>;
 
  public:
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
   using scalar_array_type       = typename traits::scalar_array_type;
   using const_scalar_array_type = typename traits::const_scalar_array_type;
   using non_const_scalar_array_type =
       typename traits::non_const_scalar_array_type;
+#endif
```
science-enthusiast, line 203:
```
#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_5
  using scalar_array_type KOKKOS_DEPRECATED_WITH_COMMENT(
      "Use data_type instead.") = data_type;
  using const_scalar_array_type KOKKOS_DEPRECATED_WITH_COMMENT(
      "Use const_data_type instead.") = const_data_type;
  using non_const_scalar_array_type KOKKOS_DEPRECATED_WITH_COMMENT(
      "Use non_const_data_type instead.") = non_const_data_type;
#endif
```

containers/unit_tests/TestDynRankViewTypedefs.cpp
```diff
@@ -96,7 +98,15 @@ constexpr bool test_view_typedefs_impl() {
...
   // FIXME: in contrast to View, hooks_policy is not propagated
   static_assert(std::is_same_v<typename ViewType::traits, ViewTraitsType>);
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+KOKKOS_IMPL_DISABLE_DEPRECATED_WARNINGS_PUSH()
   static_assert(std::is_same_v<typename ViewType::array_type,
+                               Kokkos::DynRankView<typename ViewType::scalar_array_type, typename ViewType::array_layout,
+                                            typename ViewType::device_type, //typename ViewTraitsType::hooks_policy,
+                                            typename ViewType::memory_traits>>);
+KOKKOS_IMPL_DISABLE_DEPRECATED_WARNINGS_POP()
+#endif
```
science-enthusiast, line 108: Could we simplify this code as `static_assert(std::is_same_v<typename ViewType::array_type, typename ViewType::type>);`, since we are explicitly static_asserting the type of `ViewType::type`?

core/src/View/Kokkos_ViewCommonType.hpp
```diff
@@ -36,8 +36,7 @@ struct CommonViewAllocProp;
 
 template <class ValueType>
 struct CommonViewAllocProp<void, ValueType> {
-  using value_type        = ValueType;
-  using scalar_array_type = ValueType;
+  using value_type = ValueType;
```
science-enthusiast, line 26: We will also need to add:

```
#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_5
  using scalar_array_type KOKKOS_DEPRECATED_WITH_COMMENT("Use value_type instead.") = value_type;
#endif
```

core/src/View/Kokkos_ViewDataAnalysis.hpp
```diff
@@ -381,15 +381,21 @@ struct ViewDataAnalysis {
...
-  using type       = typename ViewDataType<value_type, dimension>::type;
-  using const_type = typename ViewDataType<const_value_type, dimension>::type;
-  using non_const_type =
+  using data_type = typename ViewDataType<value_type, dimension>::type;
+  using const_data_type =
+      typename ViewDataType<const_value_type, dimension>::type;
+  using non_const_data_type =
       typename ViewDataType<non_const_value_type, dimension>::type;
 
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
```
science-enthusiast: We will need to add the following:

```
  using type KOKKOS_DEPRECATED_WITH_COMMENT(
      "Use data_type instead.")  = data_type;
  using const_type KOKKOS_DEPRECATED_WITH_COMMENT(
      "Use const_data_type instead.") = const_data_type;
  using non_const_type KOKKOS_DEPRECATED_WITH_COMMENT(
      "Use non_const_data_type instead.")  = non_const_data_type;
```

containers/unit_tests/TestDynRankViewTypedefs.cpp
```diff
@@ -56,16 +56,18 @@ constexpr bool test_view_typedefs_impl() {
   static_assert(std::is_same_v<typename ViewType::data_type, DataType>);
   static_assert(std::is_same_v<typename ViewType::const_data_type, typename data_analysis<DataType>::const_data_type>);
   static_assert(std::is_same_v<typename ViewType::non_const_data_type, typename data_analysis<DataType>::non_const_data_type>);
-
+  
+  #ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
   // FIXME: these should be deprecated and for proper testing (I.e. where this is different from data_type)
   // we would need ensemble types which use the hidden View dimension facility of View (i.e. which make
   // "specialize" not void)
```
science-enthusiast, line 61:
```
  // FIXME: For proper testing (I.e. where this is different from data_type) we would need
  // ensemble types which use the hidden View dimension facility of View (i.e. which make
  // "specialize" not void)
```

core/src/View/Kokkos_ViewDataAnalysis.hpp
```diff
@@ -381,15 +381,21 @@ struct ViewDataAnalysis {
   using non_const_value_type = typename array_analysis::non_const_value_type;
 
   // Generate analogous multidimensional array specification type.
-  using type       = typename ViewDataType<value_type, dimension>::type;
-  using const_type = typename ViewDataType<const_value_type, dimension>::type;
-  using non_const_type =
+  using data_type = typename ViewDataType<value_type, dimension>::type;
+  using const_data_type =
+      typename ViewDataType<const_value_type, dimension>::type;
+  using non_const_data_type =
```
science-enthusiast, line 374: Shall we document that this PR is also deprecating `type`, `const_type` and `non_const_type` in `ViewDataAnalysis`?

core/unit_test/TestViewCtorPropEmbeddedDim.hpp
```diff
@@ -75,10 +75,7 @@ struct TestViewCtorProp_EmbeddedDim {
         Kokkos::deep_copy(hcv1, cv1);
 
         ASSERT_EQ((std::is_same_v<CommonViewValueType, double>), true);
-        ASSERT_EQ((std::is_same_v<
-                      typename decltype(view_alloc_arg)::scalar_array_type,
-                      CommonViewValueType>),
-                  true);
+
```
science-enthusiast, line 73: Won't this assert work?
```
        ASSERT_EQ((std::is_same_v<
                      typename decltype(view_alloc_arg)::data_type,
                      CommonViewValueType>),
                  true);
```

core/unit_test/TestViewTypedefs.cpp
```diff
@@ -55,13 +55,17 @@ constexpr bool test_view_typedefs_impl() {
...
   // FIXME: these should be deprecated and for proper testing (I.e. where this is different from data_type)
   // we would need ensemble types which use the hidden View dimension facility of View (i.e. which make
   // "specialize" not void)
+  KOKKOS_IMPL_DISABLE_DEPRECATED_WARNINGS_PUSH()
   static_assert(std::is_same_v<typename ViewType::scalar_array_type, DataType>);
   static_assert(std::is_same_v<typename ViewType::const_scalar_array_type, typename data_analysis<DataType>::const_data_type>);
   static_assert(std::is_same_v<typename ViewType::non_const_scalar_array_type, typename data_analysis<DataType>::non_const_data_type>);
+  KOKKOS_IMPL_DISABLE_DEPRECATED_WARNINGS_POP()
+#endif
```
science-enthusiast, line 66: These are also needed:

```
  static_assert(std::is_same_v<typename ViewType::data_type, DataType>);
  static_assert(std::is_same_v<typename ViewType::const_data_type, typename data_analysis<DataType>::const_data_type>);
  static_assert(std::is_same_v<typename ViewType::non_const_data_type, typename data_analysis<DataType>::non_const_data_type>);
```

core/unit_test/TestViewTypedefs.cpp
```diff
@@ -95,10 +99,18 @@ constexpr bool test_view_typedefs_impl() {
   // in Legacy View: some helper View variants
   // =========================================
   static_assert(std::is_same_v<typename ViewType::traits, ViewTraitsType>);
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+KOKKOS_IMPL_DISABLE_DEPRECATED_WARNINGS_PUSH()
   static_assert(std::is_same_v<typename ViewType::array_type,
                                Kokkos::View<typename ViewType::scalar_array_type, typename ViewType::array_layout,
                                             typename ViewType::device_type, typename ViewTraitsType::hooks_policy,
                                             typename ViewType::memory_traits>>);
```
science-enthusiast, line 105: The static_assert can be simplified as:

`static_assert(std::is_same_v<typename ViewType::array_type, typename ViewType::type>);`

dalg24 (commented):
Keep it as "deprecated code 4" for now
There is very limited usage known downstream, that is being replaced by Christian as part of the porting to the new View layout customization point.
Also, the facility to use instead has been available for a while so code can be updated without guarding for the Kokkos version.

dalg24 (commented):
Please update the PR description to reflect the current state

core/unit_test/TestViewTypedefs.cpp
```diff
@@ -55,13 +55,17 @@ constexpr bool test_view_typedefs_impl() {
   static_assert(std::is_same_v<typename ViewType::data_type, DataType>);
   static_assert(std::is_same_v<typename ViewType::const_data_type, typename data_analysis<DataType>::const_data_type>);
   static_assert(std::is_same_v<typename ViewType::non_const_data_type, typename data_analysis<DataType>::non_const_data_type>);
-  
+
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
   // FIXME: these should be deprecated and for proper testing (I.e. where this is different from data_type)
   // we would need ensemble types which use the hidden View dimension facility of View (i.e. which make
   // "specialize" not void)
```
science-enthusiast, line 60: Will the following comment work?
```
  // FIXME: For proper testing (I.e. where this is different from data_type) we would need
  // ensemble types which use the hidden View dimension facility of View (i.e. which make
  // "specialize" not void)
```

containers/unit_tests/TestDynRankViewTypedefs.cpp
janciesko: We want to explicitly construct the `DynRankView` type here to test if we're constructing the correct `DynRankView::type` member in `DynRankView` itself.

core/src/View/Kokkos_ViewCommonType.hpp
janciesko: Note that this is in `Kokkos::Impl`, we should be ok just dropping it.

core/src/View/Kokkos_ViewDataAnalysis.hpp
janciesko: See L392-397. Should be there already.

core/unit_test/TestViewCtorPropEmbeddedDim.hpp
janciesko: Good catch, it should. Updating.

core/unit_test/TestViewTypedefs.cpp
janciesko: Same as the other comment RE simplification

core/unit_test/TestViewTypedefs.cpp
janciesko: We'll get rid of the comment once we remove the deprecated code block.

core/unit_test/TestViewTypedefs.cpp
janciesko: This is included already, see lines L55-L57

core/src/View/Kokkos_ViewDataAnalysis.hpp
janciesko: ~No, `::Impl`, we should also be able to just drop those once the deprecated tests go away.~
Document in the PR, yes.

janciesko: I've added a deprecation block for those members vs just dropping them. I believe this is what you mean by "documenting".

containers/src/Kokkos_DynRankView.hpp
janciesko: Yes, that should be the recommendation.

janciesko:
@dalg24  - We're inclined to mark all of this for deprecation in 5.0 now.  Fixing the CI issues.

containers/src/Kokkos_DynRankView.hpp
janciesko: Yes

core/src/View/Kokkos_ViewCommonType.hpp
```diff
@@ -23,8 +23,13 @@ struct CommonViewAllocProp;
...
+  using data_type  = ValueType;
+
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_5
+  using scalar_array_type KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use data_type instead.") = data_type;
+#endif
```
dalg24:
```suggestion
```

core/src/View/Kokkos_ViewDataAnalysis.hpp
science-enthusiast: No, I meant mentioning this change in the PR description.

core/src/View/Kokkos_ViewDataAnalysis.hpp
```diff
@@ -368,15 +368,30 @@ struct ViewDataAnalysis {
...
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_5
+  using type KOKKOS_DEPRECATED_WITH_COMMENT("Use data_type instead.") =
+      data_type;
+  using const_type KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use const_data_type instead.") = const_data_type;
+  using non_const_type KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use non_const_data_type instead.") = non_const_data_type;
+#endif
+
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_5
```
science-enthusiast: We could combine these two `#ifdef` blocks.

containers/unit_tests/TestDynRankViewTypedefs.cpp
science-enthusiast: We can have it this way also. No problem.

containers/src/Kokkos_DynRankView.hpp
science-enthusiast: I don't follow this discussion. Please, clarify. What we are trying to push is this:

https://github.com/kokkos/kokkos/blob/6e54624e1eb4fceb2e25cc4a0272ce2f23dc25c8/containers/src/Kokkos_DynRankView.hpp#L462-L469

core/src/View/Kokkos_ViewDataAnalysis.hpp
janciesko: done

core/src/View/Kokkos_ViewDataAnalysis.hpp
janciesko: done

crtrott (approved):
*there was some old comment from me in the review box*

containers/unit_tests/TestDynRankViewTypedefs.cpp
science-enthusiast: This change is not needed. The comment is inside deprecated code that will be eventually deleted.

containers/src/Kokkos_DynamicView.hpp
science-enthusiast, line 1: Could the change in this case be documented in the PR description, please?

`array_type` is deprecated in favour of `uniform_type`, which is different from other places.

containers/src/Kokkos_OffsetView.hpp
```diff
@@ -211,11 +211,18 @@ class OffsetView : public View<DataType, Properties...> {
...
   //----------------------------------------
+
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_5
   /** \brief  Compatible view of array of scalar types */
-  using array_type =
+  using array_type KOKKOS_DEPRECATED_WITH_COMMENT("Use type instead.") =
       OffsetView<typename traits::scalar_array_type,
                  typename traits::array_layout, typename traits::device_type,
                  typename traits::memory_traits>;
+#endif
```
science-enthusiast: Sorry for not noticing it earlier. This block could be moved after the using declaration of `type`, in the following form:

```
#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_5
  /** \brief  Compatible view of array of scalar types */
  using array_type KOKKOS_DEPRECATED_WITH_COMMENT("Use type instead.") = type;
#endif
```

containers/src/Kokkos_DynamicView.hpp
janciesko: Done

containers/src/Kokkos_OffsetView.hpp
janciesko: Done

science-enthusiast (approved):
Looks good to me.

PR #7361, opened 2024-09-19
Move to Staged Testing

crtrott:
This introduced more staged testing both for GitHub actions and for Jenkins

Github Actions: I made the primary files not trigger on PRs but on "workflow_call" and introduced an overarching `continuous-integration-stager.yml` which stages the sub-workflows.

First stage: clang-format, codeql, simple GCC Serial build
Second stage: everything else

For Jenkins we already had a clang-format stage, I introduced a second one, and move one Hip, one CUDA, and the GCC 8.4 build there.


crtrott:
Fixed in #7362

PR #7364, opened 2024-09-20
`ViewCtorProp`: default constructor and move semantics

maartenarnst:
The `ViewCtorProp` class is important in the implementation of the often used `view_alloc` function.

We have recently looked at this class in more detail and noted several issues:

- As it stands, the `ViewCtorProp` is not default constructible. For example, the following code doesn't compile:

```
using vcp_t = Kokkos::Impl::ViewCtorProp<std::string>;
vcp_t prop; // compilation errror
```

- Even though some of the base classes of `ViewCtorProp` have constructors that take arguments by rvalue reference, these constructors do not subsequently move the passed arguments. Hence, they may ultimately not move. For example, the following test fails:

```
using vcp_t = Kokkos::Impl::ViewCtorProp<std::string>;
std::string label("my label we like to move around because it is soooo long !");
const vcp_t prop(std::move(label));
ASSERT_TRUE(label.empty()); // runtime test failure
```

- Even though the functionalities of `ViewCtorProp` and `view_alloc` are tested implicitly by numerous tests that use them, it seems there are no unit tests of these functionalities just by themselves.

Just for context/use case, we noted these issues while we explored how to make the new `SequentialHostInit` property work with `DualView`. This led us to think about passing different view constructor properties to the host and device sides. And this is how default constructing view constructor properties, moving them, and so on came up.

Thus, this PR: 
- adds a default constructor to `ViewCtorProp`
- modifies a constructor and adds a move constructor to `ViewCtorProp`
- modifies a constructor of the `ViewCtorProp` base class dealing with labels to support move semantics
- adds unit tests for `ViewCtorProp`

As you will see, the modifications are limited to the main `ViewCtorProp` class and the base class dealing with labels. In this way, the PR remains small for a first review. Similar modifications could be made for the other base classes. I would be happy to do it in this PR or in a follow up.

Joint work with @romintomasetti

### DIFF
--- a/core/src/View/Kokkos_ViewCtor.hpp
+++ b/core/src/View/Kokkos_ViewCtor.hpp
@@ -113,7 +113,7 @@ struct ViewCtorProp<std::enable_if_t<is_view_label<Label>::value>, Label> {
   using type = std::string;
 
   ViewCtorProp(const type &arg) : value(arg) {}
-  ViewCtorProp(type &&arg) : value(arg) {}
+  ViewCtorProp(type &&arg) : value(std::move(arg)) {}
 
   type value;
 };
@@ -213,11 +213,16 @@ struct ViewCtorProp : public ViewCtorProp<void, P>... {
   using execution_space = typename var_execution_space::type;
   using pointer_type    = typename var_pointer::type;
 
-  /*  Copy from a matching argument list.
-   *  Requires  std::is_same< P , ViewCtorProp< void , Args >::value ...
-   */
-  template <typename... Args>
-  inline ViewCtorProp(Args const &...args) : ViewCtorProp<void, P>(args)... {}
+  // Construct from a matching argument list.
+  //
+  // Note that if P is empty, this constructor is the default constructor.
+  // On the other hand, if P is not empty, the constraint implies that
+  // there is no default constructor.
+  template <typename... Args,
+            typename = std::enable_if_t<std::conjunction_v<
+                std::is_constructible<view_ctor_prop_base<P>, Args &&>...>>>
+  ViewCtorProp(Args &&...args)
+      : ViewCtorProp<void, P>(std::forward<Args>(args))... {}
 
   template <typename... Args>
   KOKKOS_FUNCTION ViewCtorProp(pointer_type arg0, Args const &...args)
@@ -454,15 +459,14 @@ inline constexpr Kokkos::Impl::AllowPadding_t AllowPadding{};
  * alignment
  */
 template <class... Args>
-inline Impl::ViewCtorProp<typename Impl::ViewCtorProp<void, Args>::type...>
-view_alloc(Args const &...args) {
-  using return_type =
-      Impl::ViewCtorProp<typename Impl::ViewCtorProp<void, Args>::type...>;
+auto view_alloc(Args &&...args) {
+  using return_type = Impl::ViewCtorProp<typename Impl::ViewCtorProp<
+      void, Kokkos::Impl::remove_cvref_t<Args>>::type...>;
 
   static_assert(!return_type::has_pointer,
                 "Cannot give pointer-to-memory for view allocation");
 
-  return return_type(args...);
+  return return_type(std::forward<Args>(args)...);
 }
 
 template <class... Args>
--- a/core/unit_test/TestViewCtorProp.hpp
+++ b/core/unit_test/TestViewCtorProp.hpp
@@ -1,0 +1,95 @@
+//@HEADER
+// ************************************************************************
+//
+//                        Kokkos v. 4.0
+//       Copyright (2022) National Technology & Engineering
+//               Solutions of Sandia, LLC (NTESS).
+//
+// Under the terms of Contract DE-NA0003525 with NTESS,
+// the U.S. Government retains certain rights in this software.
+//
+// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
+#include <gtest/gtest.h>
+
+#include <Kokkos_Core.hpp>
+
+namespace {
+
+using vcp_empty_t      = Kokkos::Impl::ViewCtorProp<>;
+using vcp_label_base_t = Kokkos::Impl::ViewCtorProp<void, std::string>;
+using vcp_label_t      = Kokkos::Impl::ViewCtorProp<std::string>;
+
+// Check traits of Kokkos::Impl::ViewCtorProp<>.
+TEST(TEST_CATEGORY, vcp_empty_traits) {
+  // Check that the empty view constructor properties class is default
+  // constructible. This is needed for calls of Kokkos::view_alloc().
+  static_assert(std::is_default_constructible_v<vcp_empty_t>);
+  static_assert(std::is_same_v<decltype(Kokkos::view_alloc()), vcp_empty_t>);
+}
+
+// Check Kokkos::Impl::is_view_label.
+TEST(TEST_CATEGORY, is_view_label) {
+  static_assert(Kokkos::Impl::is_view_label<std::string>::value);
+
+  static_assert(Kokkos::Impl::is_view_label<const char[3]>::value);
+  static_assert(Kokkos::Impl::is_view_label<char[3]>::value);
+
+  // A char* is not a label. Thus, a label is distinguished from a pointer type.
+  static_assert(!Kokkos::Impl::is_view_label<char*>::value);
+}
+
+// Check traits of base class Kokkos::Impl::ViewCtorProp<void, std::string>.
+TEST(TEST_CATEGORY, vcp_label_base_traits) {
+  static_assert(std::is_same_v<typename vcp_label_base_t::type, std::string>);
+
+  // Check that the base class is default constructible. The default constructor
+  // may be called by the copy constructor of derived classes, such as when
+  // copy constructing a view constructor properties object from another view
+  // constructor properties object that holds fewer properties.
+  static_assert(std::is_default_constructible_v<vcp_label_base_t>);
+}
+
+// Check traits of derived class Kokkos::Impl::ViewCtorProp<std::string>.
+TEST(TEST_CATEGORY, vcp_label_traits) {
+  static_assert(std::is_base_of_v<vcp_label_base_t, vcp_label_t>);
+
+  static_assert(vcp_label_t::has_label);
+
+  // Check that the derived class is not default constructible. It is a design
+  // choice to not allow the default constructor to be called.
+  static_assert(!std::is_default_constructible_v<vcp_label_t>);
+}
+
+// Check that Kokkos::view_alloc perfect forwards a label passed by
+// rvalue reference, and check that the constructor
+// of Kokkos::Impl::ViewCtorProp<std::string> moves this label.
+TEST(TEST_CATEGORY, view_alloc_can_perfect_forward_label) {
+  std::string label("our label");
+
+  auto prop = Kokkos::view_alloc(std::move(label));
+
+  ASSERT_TRUE(label.empty());
+  ASSERT_EQ(Kokkos::Impl::get_property<Kokkos::Impl::LabelTag>(prop),
+            "our label");
+}
+
+// Check the copy constructor of Kokkos::Impl::ViewCtorProp<std::string>.
+TEST(TEST_CATEGORY, vcp_label_copy_constructor) {
+  // Copy construction from a view constructor properties object with a label.
+  static_assert(std::is_copy_constructible_v<vcp_label_t>);
+
+  vcp_label_t prop = Kokkos::view_alloc("our label");
+  vcp_label_t prop_copy(prop);
+
+  ASSERT_EQ(Kokkos::Impl::get_property<Kokkos::Impl::LabelTag>(prop),
+            "our label");
+  ASSERT_EQ(Kokkos::Impl::get_property<Kokkos::Impl::LabelTag>(prop_copy),
+            "our label");
+}
+
+}  // namespace
--- a/core/unit_test/CMakeLists.txt
+++ b/core/unit_test/CMakeLists.txt
@@ -261,6 +261,7 @@ foreach(Tag Threads;Serial;OpenMP;Cuda;HPX;OpenMPTarget;OpenACC;HIP;SYCL)
         ViewCopy_b
         ViewCopy_c
         ViewCtorDimMatch
+        ViewCtorProp
         ViewEmptyRuntimeUnmanaged
         ViewHooks
         ViewLayoutStrideAssignment


core/src/View/Kokkos_ViewCtor.hpp
```diff
@@ -236,11 +235,12 @@ struct ViewCtorProp : public ViewCtorProp<void, P>... {
   template <typename... Args>
   ViewCtorProp(view_ctor_prop_args<Args...> const &arg)
       : view_ctor_prop_base<Args>(
-            static_cast<view_ctor_prop_base<Args> const &>(arg))... {
-    // Suppress an unused argument warning that (at least at one point) would
-    // show up if sizeof...(Args) == 0
-    (void)arg;
```
maartenarnst, line 242: Just a note for the reviewers: it looks like this is no longer needed after adding the default constructor.

core/src/View/Kokkos_ViewCtor.hpp
```diff
@@ -234,13 +233,14 @@ struct ViewCtorProp : public ViewCtorProp<void, P>... {
...
-  }
+            static_cast<view_ctor_prop_base<Args> const &>(arg))... {}
+
+  template <typename... Args>
+  ViewCtorProp(view_ctor_prop_args<Args...> &&arg)
```
romintomasetti, line 241: Could these 2 constructors actually be merged into one ?

core/unit_test/TestViewCtorProp.hpp
```diff
@@ -0,0 +1,136 @@
...
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
+#include <gtest/gtest.h>
+
+#include <Kokkos_Core.hpp>
+
+namespace Test {
```
romintomasetti, line 21: As @dalg24 would say, you probably don't need `Test`.

core/unit_test/TestViewCtorProp.hpp
```diff
@@ -0,0 +1,136 @@
...
+
+// Check traits of Kokkos::Impl::ViewCtorProp<void, std::string>.
+TEST(TEST_CATEGORY, vcp_label_base_traits) {
+  using vcp_label_base_t = Kokkos::Impl::ViewCtorProp<void, std::string>;
+
+  static_assert(std::is_default_constructible_v<vcp_label_base_t>);
+  static_assert(std::is_copy_constructible_v<vcp_label_base_t>);
+  static_assert(std::is_copy_assignable_v<vcp_label_base_t>);
+  static_assert(std::is_move_constructible_v<vcp_label_base_t>);
+  static_assert(std::is_move_assignable_v<vcp_label_base_t>);
```
romintomasetti, line 42: Would it be worth adding a single trait that does all these checks ? (at least for this test)

core/src/View/Kokkos_ViewCtor.hpp
```diff
@@ -213,11 +211,12 @@ struct ViewCtorProp : public ViewCtorProp<void, P>... {
   using execution_space = typename var_execution_space::type;
   using pointer_type    = typename var_pointer::type;
 
-  /*  Copy from a matching argument list.
-   *  Requires  std::is_same< P , ViewCtorProp< void , Args >::value ...
-   */
+  ViewCtorProp() = default;
```
masterleinad, line 214: Why is it a good idea to create a default-constructed object (possibly with invalid values)?

core/src/View/Kokkos_ViewCtor.hpp
```diff
@@ -213,11 +211,12 @@ struct ViewCtorProp : public ViewCtorProp<void, P>... {
   using execution_space = typename var_execution_space::type;
   using pointer_type    = typename var_pointer::type;
 
-  /*  Copy from a matching argument list.
-   *  Requires  std::is_same< P , ViewCtorProp< void , Args >::value ...
-   */
```
masterleinad, line 218: Why did you remove this comment?

core/src/View/Kokkos_ViewCtor.hpp
romintomasetti: I thin k it was possible before this PR already...

But if you ask me, I'd say a defaulted view property has no label, and basically all the rest is defaulted too (memory space, execution space, initialization and so on).

masterleinad: > I thin k it was possible before this PR already...

No, it wasn't. You would get a mismatch in the size of the `Args..` and `P...` parameter packs.

masterleinad: > But if you ask me, I'd say a defaulted view property has no label, and basically all the rest is defaulted too (memory space, execution space, initialization and so on).

Sure it's the same as not using a `ViewCtorProp` argument  in the respective API calls. The point is that you would specify template arguments for which you would explicitly want to set a non-default value and so a default constructor somewhat defeats that purpose.

dalg24:
I can see how `Impl::ViewCtorProp` could be improved but would you please clarify what legal user code you are trying to write today and that does not compile or does not behave as one might expect it should.

maartenarnst:
> I can see how `Impl::ViewCtorProp` could be improved but would you please clarify what legal user code you are trying to write today and that does not compile or does not behave as one might expect it should.

Hi @dalg24. The user code we are trying to write today is a utility function to create a dual view using the `SequentialHostInit` property for the host side. A simplified version of our function looks like:

```
template <typename DualViewType, typename ViewCtorPropType, typename... Args>
auto create_dualview_seqhostinit(ViewCtorPropType&& prop, Args&&... args)
{
    static_assert(std::remove_cvref_t<ViewCtorPropType>::has_label);
    
    auto prop_h = Kokkos::view_alloc(
        Kokkos::SequentialHostInit,
        Kokkos::Impl::get_property<Kokkos::Impl::LabelTag>(std::forward<ViewCtorPropType>(prop))    
    );

    typename DualViewType::t_host view_h(std::move(prop_h), std::forward<Args>(args)...);

    // ... view constructor props (mem, exec, WithoutInitializing) for the device view

    // ... construction of the device view using create_mirror_view

    DualViewType dual_view(std::move(view_d), std::move(view_h));

    // ... sync state of the dual view

    return dual_view;
}
``` 

Then, we wanted to write tests using a fixture class like:

```
class DualViewSeqHostInitTest : public ::testing::Test
{
protected:
    using dual_view_t = ...
    
    using vcp_label_t = decltype(Kokkos::view_alloc(std::declval<std::string>()));

protected:
    void SetUp() override {
        prop = Kokkos::view_alloc("my dual view");
    }

protected:
    vcp_label_t prop; 
};

TEST_F(DualViewSeqHostInitTest, ...)
{
    dual_view_t dual_view = create_dualview_seqhostinit<dual_view_t>(this->prop, ...);

    ...
}

...
```

This way of setting up the tests didn't work because `Impl::ViewCtorProp` is not default constructible. I.e., when we tried it this way, we got the error "You would get a mismatch in the size of the Args.. and P... parameter packs." that @masterleinad mentioned. In our code, we just avoided the issue by using a constructor for the fixture instead of the `SetUp` method. But this is how we ended up looking in more detail at the implementation of `Impl::ViewCtorProp`.

Looking at the implementation of `Impl::ViewCtorProp`, we noticed that all the base classes have default constructors, and, where relevant, they have constructors that take an argument by rvalue reference. E.g.:

- https://github.com/kokkos/kokkos/blob/6cc8ceaa2a95856d4b6a36f813e115196a8b7414/core/src/View/Kokkos_ViewCtor.hpp#L106-L119

And so this could be interpreted as if the code was written with the intent that `Impl::ViewCtorProp` is itself also default constructible and can move. Hence, this PR to try to "fix" small pieces that seemed to be missing.

Thinking of "code you are trying to write today and that ... does not behave as one might expect it should". In the snippet above, we may want the label to be moved from `prop` into `prop_h` (same for other view ctor props). Because of how `view_alloc` and `get_property` are currently implemented, I think the current implementation would copy. If we'd want to change that, I think we would have to modify also `view_alloc`, `get_property`, ...

core/src/View/Kokkos_ViewCtor.hpp
maartenarnst: I removed the comment because now that the function passes by universal reference, I thought the comment might become less readable, because we'd have to remove references and const. All in all, it looked like by keeping the line "requires a matching argument list", things would remain sufficiently clear. (just a detail, it seems the code snippet is also missing the type alias `type` from `ViewCtorProp<void , Args>`)

core/src/View/Kokkos_ViewCtor.hpp
romintomasetti: I might have something wrong in my setting, but this compiles (current develop branch):
```c++
TEST(try, it)
{
    using ctor_t = Kokkos::Impl::ViewCtorProp<void, std::string>;
    static_assert(std::is_default_constructible_v<ctor_t>);
}
```

So it seems that it's already default constructible, and this PR does not change the current behavior.

masterleinad: We are considerd about
```C++
using ctor_t = Kokkos::Impl::ViewCtorProp<std::string>;
static_assert(std::is_default_constructible_v<ctor_t>);
```
which would also work but
```C++
ctor_t bla;
```
fails. `is_default_constructible_v` only detects if there is a default constructor. It doesn't detect if it's instantiable (if it's a variadic template), also see https://godbolt.org/z/b4dMb91Pz.

maartenarnst:
Hi @dalg24 and @masterleinad. Thanks a lot for your feedback and the discussion! With @romintomasetti, we have worked the PR. We have focused it specifically on ensuring that when a user calls `view_alloc` with a label, then that label can be moved. 

This has required four changes:
- add missing `std::move` in the constructor of the base class constructor that passes by rvalue reference
- pass by forwarding reference in the derived class constructor
- pass by forwarding reference in `view_alloc`
- add a constraint to the derived class constructor to ensure it works well alongside the existing copy constructor

This PR no longer changes what is default constructible and what is not. However, note that as @masterleinad had shown on Godbolt, the `std::is_default_constructible_v` was not detecting that the derived class was not default constructible if there is at least one view constructor property. The constraint that we added fixed that issue (at least in the tests we've run so far on our machines :)).

core/unit_test/TestViewCtorProp.hpp
maartenarnst: OK. Done. More consistent indeed!

core/src/View/Kokkos_ViewCtor.hpp
maartenarnst: Reverted.

core/src/View/Kokkos_ViewCtor.hpp
maartenarnst: Reverted.

core/src/View/Kokkos_ViewCtor.hpp
maartenarnst: We added a modified version of this constraint directly in the constructor in the code.

core/unit_test/TestViewCtorProp.hpp
```diff
@@ -0,0 +1,145 @@
...
+
+  // Copy construction from an empty view constructor properties object.
+  static_assert(std::is_constructible_v<vcp_label_t, const vcp_empty_t&>);
+
+  vcp_empty_t prop_empty;
+  vcp_label_t prop_empty_copy(prop_empty);
+
+  ASSERT_TRUE(
+      Kokkos::Impl::get_property<Kokkos::Impl::LabelTag>(prop_empty_copy)
+          .empty());
```
dalg24, line 131: Not sure what is the goal of these checks.
Just like default constructing a ViewCtorProp that contains a label, constructing from the base class makes little sense to me.

masterleinad (approved):
Fine with me.

core/unit_test/TestViewCtorProp.hpp
```diff
@@ -0,0 +1,145 @@
...
+  // Copy construction from a view constructor properties object with a label.
+  static_assert(std::is_copy_constructible_v<vcp_label_t>);
+
+  auto prop = Kokkos::view_alloc("our label");
+  vcp_label_t prop_copy(prop);
+
+  ASSERT_EQ(Kokkos::Impl::get_property<Kokkos::Impl::LabelTag>(prop),
+            "our label");
+  ASSERT_EQ(Kokkos::Impl::get_property<Kokkos::Impl::LabelTag>(prop_copy),
+            "our label");
```
dalg24, line 142: Also not quite convinced this is super useful.  This would pass with the current develop.

masterleinad:
```
/__w/kokkos/kokkos/core/src/View/Kokkos_ViewCtor.hpp(436): error: no instance of constructor "Kokkos::Impl::ViewCtorProp<P...>::ViewCtorProp [with P=<Kokkos::Impl::WithoutInitializing_t, std::string>]" matches the argument list
            argument types are: (Kokkos::Impl::WithoutInitializing_t, std::string)
        : ViewCtorProp<WithoutInitializing_t, std::string>(
                                                          ^
/__w/kokkos/kokkos/core/src/View/Kokkos_ViewCtor.hpp(182): note: this candidate was rejected because mismatch in count of arguments
  struct ViewCtorProp : public ViewCtorProp<void, P>... {
         ^
/__w/kokkos/kokkos/core/src/View/Kokkos_ViewCtor.hpp(182): note: this candidate was rejected because mismatch in count of arguments
  struct ViewCtorProp : public ViewCtorProp<void, P>... {
         ^
/__w/kokkos/kokkos/core/src/View/Kokkos_ViewCtor.hpp(242): note: this candidate was rejected because mismatch in count of arguments
    ViewCtorProp(view_ctor_prop_args<Args...> const &arg)
    ^
/__w/kokkos/kokkos/core/src/View/Kokkos_ViewCtor.hpp(233): note: this candidate was rejected because mismatch in count of arguments
    KOKKOS_FUNCTION ViewCtorProp(pointer_type arg0)
                    ^
/__w/kokkos/kokkos/core/src/View/Kokkos_ViewCtor.hpp(228): note: this candidate was rejected because arguments do not match
    KOKKOS_FUNCTION ViewCtorProp(pointer_type arg0, Args const &...args)
                    ^
/__w/kokkos/kokkos/core/src/View/Kokkos_ViewCtor.hpp(224): note: this candidate was rejected because at least one template argument could not be deduced
    ViewCtorProp(Args &&...args)
    ^
```

core/src/View/Kokkos_ViewCtor.hpp
```diff
@@ -213,11 +213,16 @@ struct ViewCtorProp : public ViewCtorProp<void, P>... {
...
-  template <typename... Args>
-  inline ViewCtorProp(Args const &...args) : ViewCtorProp<void, P>(args)... {}
+  // Construct from a matching argument list.
+  //
+  // Note that if P is empty, this constructor is the default constructor.
+  // On the other hand, if P is not empty, the constraint implies that
+  // the default constructor cannot be called.
```
romintomasetti, line 220:
```suggestion
  // On the other hand, if P is not empty, the constraint implies that
  // there is no default constructor.
```

core/unit_test/TestViewCtorProp.hpp
```diff
@@ -0,0 +1,145 @@
...
+  static_assert(std::is_constructible_v<vcp_label_t, std::string>);
+  static_assert(std::is_constructible_v<vcp_label_t, const std::string&>);
+  static_assert(std::is_constructible_v<vcp_label_t, std::string&&>);
+
+  constexpr unsigned N = 3;
+  static_assert(std::is_constructible_v<vcp_label_t, const char[N]>);
+  static_assert(std::is_constructible_v<vcp_label_t, char[N]>);
+
+  static_assert(std::is_constructible_v<vcp_label_t, char*>);
+}
```
dalg24, line 91: Not finding this super useful.

core/unit_test/TestViewCtorProp.hpp
```diff
@@ -0,0 +1,145 @@
...
+  ASSERT_TRUE(label.empty());
+  ASSERT_EQ(Kokkos::Impl::get_property<Kokkos::Impl::LabelTag>(prop),
+            "our label");
+}
+
+// Check that Kokkos::view_alloc moves a label passed by rvalue reference.
+TEST(TEST_CATEGORY, vcp_label_view_alloc_can_move) {
+  std::string label("our label");
+
+  auto prop = Kokkos::view_alloc(std::move(label));
```
dalg24, line 111: This is equivalent to L100 in the previous unit test.  What am I missing?

core/unit_test/TestViewCtorProp.hpp
```diff
@@ -0,0 +1,145 @@
...
+// Check traits of Kokkos::Impl::ViewCtorProp<>.
+TEST(TEST_CATEGORY, vcp_empty_traits) {
+  using vcp_empty_t = Kokkos::Impl::ViewCtorProp<>;
+
+  // Check that the empty view constructor properties class is default
+  // constructible. This is needed for calls of Kokkos::view_alloc().
+  static_assert(std::is_default_constructible_v<vcp_empty_t>);
+
+  static_assert(std::is_same_v<decltype(Kokkos::view_alloc()), vcp_empty_t>);
+}
```
dalg24, line 44: What is the goal of that test?

core/src/View/Kokkos_ViewCtor.hpp
```diff
@@ -213,11 +213,16 @@ struct ViewCtorProp : public ViewCtorProp<void, P>... {
...
+  template <typename... Args,
+            typename = std::enable_if_t<std::conjunction_v<
+                std::is_constructible<ViewCtorProp<void, P>, Args &&>...>>>
+  ViewCtorProp(Args &&...args)
+      : ViewCtorProp<void, P>(std::forward<Args>(args))... {}
```
masterleinad, line 225: It seems this doesn't work with the Intel Classic compiler
```suggestion
// The Intel Classic Compiler struggles deducing the variadic template parameters with the SFINAE template parameter 
#ifdef KOKKOS_COMPILER_INTEL
 template <typename... Args>
#else
  template <typename... Args,
            typename = std::enable_if_t<std::conjunction_v<
                std::is_constructible<ViewCtorProp<void, P>, Args &&>...>>>
#endif
 ViewCtorProp(Args &&...args)
      : ViewCtorProp<void, P>(std::forward<Args>(args))... {}
```

maartenarnst:
Thanks @masterleinad for posting the error message. It looks like the intel compiler got confused about different ways in which `ViewCtorProp` is used. The issue had already been noted in the same file:

- https://github.com/kokkos/kokkos/blob/ca4325fdac0c3d8a97a9fa4c1fda3692c2a57ecf/core/src/View/Kokkos_ViewCtor.hpp#L231-L235

With @romintomasetti, we applied the fix suggested in this comment, i.e. use the type alias, in our constraint, and it looks like this fixes the issue of the compilation error with icpc.

core/unit_test/TestViewCtorProp.hpp
romintomasetti: I guess checking that `vcp_empty_t` is default constructible is a must, because this is not straightforward (and is needed).

Checking that `Kokkos::view_alloc()` returns a `vcp_empty` is somewhat a check of the internal and could be removed. However, `is_default_constructible_v` checks that the default constructor exists, but it does not instantiate it. At some point, we had a working version that was telling "yeah it is default constructible", but any call site that would actually instantiate the default constructor would not compile. So I guess this second part of the test is about this: check that instantiating the default constructor works ?

core/unit_test/TestViewCtorProp.hpp
```diff
@@ -0,0 +1,126 @@
...
+TEST(TEST_CATEGORY, vcp_label_base_traits) {
+  using vcp_label_base_t = Kokkos::Impl::ViewCtorProp<void, std::string>;
+
+  static_assert(std::is_same_v<typename vcp_label_base_t::type, std::string>);
+
+  // Check that the base class is default constructible. The default constructor
+  // may be called by the copy constructor of deriveded classes, such as when
+  // copy constructing a view constructor properties object from another view
+  // constructor properties object that holds fewer properties.
+  static_assert(std::is_default_constructible_v<vcp_label_base_t>);
```
romintomasetti, line 56: @maartenarnst Please double-check that this is also the case on the develop branch. Otherwise, you'll need to add a constraint to your variadic constructor that you need at least on argument.

romintomasetti: Sorry, my mistake. It is already default constructible on the develop. Closing.

core/unit_test/TestViewCtorProp.hpp
romintomasetti: One test checks move semantics with the internals. The other one goes through the public API `Kokkos::view_alloc`.

I would argue that the second test could be enough. Indeed, if `auto prop = Kokkos::view_alloc(std::move(label));` does not move, it means that either `view_alloc` does not perfect forward, or that the first test should also fail.

core/unit_test/TestViewCtorProp.hpp
maartenarnst: We have shortened the tests. The tests of the traits are now focused on testing the properties of being "default constructible" and "copy constructible" for the specific cases that are needed for existing use cases in Kokkos. We deleted the tests of the other traits.

core/src/View/Kokkos_ViewCtor.hpp
maartenarnst: Since this discussion, the PR has been reworked. The PR now leaves unchanged what is default constructible and what is not. In short, the base classes have an explicitly defaulted constructor. The derived class is default constructible if it is empty. But it is not default constructible if it holds one or more properties.

core/unit_test/TestViewCtorProp.hpp
maartenarnst: The goal of this test is to ensure that use case like this works:
- https://github.com/kokkos/kokkos/blob/fa09fa8713d44cef31dcbb7bd5aace19a28a1244/core/src/View/Kokkos_ViewCtor.hpp#L267

Here, a view constructor properties object "with more properties" is copy constructed from a view constructor properties object "with less properties". The base classes holding the missing properties are default constructed.

core/unit_test/TestViewCtorProp.hpp
maartenarnst: The goal of this test is to ensure that the way this PR turns the constructor of the view constructor properties class into a forwarding constructor works well with the existing copy constructor. To ensure that it works well, we had to add a constraint to the forwarding constructor. We found that if we remove the constraint, the compiler will prefer the forwarding constructor over the copy constructor, and compilation of this test will fail.

core/unit_test/TestViewCtorProp.hpp
maartenarnst: Indeed! Mostly all deleted. We kept the check of the label being stored as `std::string`, the check of `has_label`, and the check of the inheritance.

core/unit_test/TestViewCtorProp.hpp
maartenarnst: Indeed. We merged the two tests into a single tests. The combined tests now checks both `view_alloc` (public API) and the constructor (in the `Impl` namespace).

core/unit_test/TestViewCtorProp.hpp
maartenarnst: There are several use cases where `view_alloc` is called without arguments in `Kokkos`. And so this test ensures that such use case works.  @romintomasett's comment explains well with more details why it was needed.

core/src/View/Kokkos_ViewCtor.hpp
maartenarnst: Indeed. We found that it can be solved by using the `view_ctor_prop_args` typedef. So we preferred that solution over using a preprocessor conditional.

maartenarnst:
Hi @dalg24 and @masterleinad. With @romintomasetti, we have gone over the PR once more. We have shortened/motivated the tests, following @dalg24 's suggestions. Would you have a moment for another look?

maartenarnst:
Hi @dalg24. Just a reminder :) Would you have a moment to take a look if you get a chance? Thanks in advance!

core/src/View/Kokkos_ViewCtor.hpp
```diff
@@ -113,7 +113,7 @@ struct ViewCtorProp<std::enable_if_t<is_view_label<Label>::value>, Label> {
   using type = std::string;
 
   ViewCtorProp(const type &arg) : value(arg) {}
-  ViewCtorProp(type &&arg) : value(arg) {}
+  ViewCtorProp(type &&arg) : value(std::move(arg)) {}
```
nliber, line 116: Another possibility is to replace these two constructors with
```c++
ViewCtorProp(type arg) : value(std::move(arg)) {}
```

I'm not sure which is better though...

core/src/View/Kokkos_ViewCtor.hpp
```diff
@@ -213,11 +213,16 @@ struct ViewCtorProp : public ViewCtorProp<void, P>... {
...
-   */
-  template <typename... Args>
-  inline ViewCtorProp(Args const &...args) : ViewCtorProp<void, P>(args)... {}
+  // Construct from a matching argument list.
+  //
+  // Note that if P is empty, this constructor is the default constructor.
+  // On the other hand, if P is not empty, the constraint implies that
+  // there is no default constructor.
+  template <typename... Args,
+            typename = std::enable_if_t<std::conjunction_v<
```
nliber, line 222: I don't quite understand this.

`conjunction` with just one argument is unusual.

And the comment doesn't see right, either.  Whether or not this acts as a default constructor depends not only on the constraint but also on `sizeof...(Args) == 0`.

nliber (changes requested):

core/src/View/Kokkos_ViewCtor.hpp
crtrott: leave it like it is

core/src/View/Kokkos_ViewCtor.hpp
crtrott: The arguments are the expansion of this: `std::is_constructible<view_ctor_prop_base<P>, Args &&>...`

crtrott (approved):
Looks fine to me now. @nliber I think I answered your question or?

crtrott (dismissed):
I am actually fine with this: but one of the test configs has a warning which makes it not compile in the test. Needs to be fixed. Feel free to dismiss my change-request when it is fixed.

dalg24 (commented):
Not blocking
(Just realized I had forgotten to send my review)

core/unit_test/TestViewCtorProp.hpp
```diff
@@ -0,0 +1,106 @@
...
+TEST(TEST_CATEGORY, vcp_label_copy_constructor) {
+  // Copy construction from an empty view constructor properties object.
+  static_assert(std::is_constructible_v<vcp_label_t, const vcp_empty_t&>);
+
+  vcp_empty_t prop_empty;
+  vcp_label_t prop_empty_copy(prop_empty);
+
+  ASSERT_TRUE(
+      Kokkos::Impl::get_property<Kokkos::Impl::LabelTag>(prop_empty_copy)
+          .empty());
```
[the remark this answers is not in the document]
maartenarnst: This behavior is currently needed because of how `Impl::with_properties_if_unset` is currently implemented:

- https://github.com/kokkos/kokkos/blob/22fea8395ac617d3b5a933e8cc2905a4c3b3b6f4/core/src/View/Kokkos_ViewCtor.hpp#L252-L270

There are current use cases where `view_ctor_prop` on line 267 is empty, and `new_view_ctor_prop` on this line is not empty, e.g. it has a label, or an execution space instance, or ... Thus, with this test, we want to make sure that such current use cases work.

I agree that this behavior is a bit strange. It may make sense to refactor `Impl::with_properties_if_unset` to avoid it. The test could then be deleted. May be best in a follow-up PR? I would be happy to work on it.

It's fine either way for me. I can also delete the test now if you prefer.

core/unit_test/TestViewCtorProp.hpp
```diff
@@ -0,0 +1,106 @@
...
+  static_assert(!std::is_default_constructible_v<vcp_label_t>);
+}
+
+// Check that Kokkos::view_alloc perfect forwards a label passed by
+// rvalue reference, and check that the constructor
+// of Kokkos::Impl::ViewCtorProp<std::string> moves this label.
+TEST(TEST_CATEGORY, view_alloc_can_perfect_forward_label) {
+  std::string label("our label");
+
+  vcp_label_t prop = Kokkos::view_alloc(std::move(label));
```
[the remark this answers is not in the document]
maartenarnst: We spelled it out to make the test a bit stronger. It's fine either way for me too. I can change it to `auto` if you prefer.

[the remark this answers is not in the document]
dalg24: The way I see it, this unit test is about checking that, if a string is "moved" into view_alloc(), then it is indeed in a moved-from state which (hopefully) means no copy occurred.
The actual type returned by view_alloc() is an implementation detail, it is non obvious from the place you read the test what that return type is and that naming the type like you do does not trigger a conversion.

core/unit_test/TestViewCtorProp.hpp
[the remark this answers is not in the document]
dalg24: As I understand it, his PR is about properly moving label strings into view alloc property construct and it is orthogonal to what `Impl::with_properties_if_unset()` does, meaning that test addition does not really belong here.

core/unit_test/TestViewCtorProp.hpp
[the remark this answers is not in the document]
maartenarnst: OK. No problem. I have just replaced the `vcp_label_t` with `auto`.

core/unit_test/TestViewCtorProp.hpp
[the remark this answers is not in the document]
maartenarnst: OK. No problem. I have just deleted this test that constructs a copy from an empty view alloc prop. Perhaps when we have a chance, we can discuss live whether or not we'd like to do a follow-up pr. It could be a follow-up pr that adds move semantics to `Impl::with_properties_if_unset()` and also addresses the issue of how it relies on the strange behavior of the copy constructor of the view alloc prop.

nliber (approved):

dalg24:
Unrelated NVHPC seagull

PR #7368, opened 2024-09-23
Fix using OpenMP TeamPolicy from OpenMP parallel region

### LINKED ISSUE #7357: Revisit OpenMP thread count restrictions for partition_space and TeamPolicies constructs called nested.
#7349 restricted tests if the number of detected threads is less than the numbers of partitions requested for an OpenMP execution space instance. However, its' not clear why `exec_space_thread_safety_team_policy` and `exec_space_thread_safety_team_policy` using the same execution space instance are also erroring out when called in a nested context.

masterleinad:
Fixes #7357. When Kokkos OpenMP parallel constructs are called from a OpenMP parallel region, we want to serialize the kernel. This also means that we should explicitly use resources dedicated for that one thread running (and not ask for the thread number).
This is similar to https://github.com/kokkos/kokkos/blob/1d92d9753a8c727632a7d9238eb9189d95e9364f/core/src/OpenMP/Kokkos_OpenMP_Parallel_Scan.hpp#L84-L92 and https://github.com/kokkos/kokkos/blob/1d92d9753a8c727632a7d9238eb9189d95e9364f/core/src/OpenMP/Kokkos_OpenMP_Parallel_Scan.hpp#L210-L220


ldh4 (commented):
Changes look logical, but are there tests that would expose the described issue?

masterleinad:
> Changes look logical, but are there tests that would expose the described issue?

Not easily. I found this when setting `OMP_NUM_THREADS=1` since we then don't even allocate memory for a second thread in the default execution space instance.

diehlpk (approved):
I think it looks good but the question with the tests is valid.

masterleinad:
It might be useful to have an `OpenMP` with `OMP_NUM_THREADS=1` since we ran into multiple problems with that.

masterleinad:
> I am not sure about this change because `get_thread_data()` should be the same as `get_thread_data(0)` in the case you describe. The real bug might be in `get_thread_data()`.

It's a question of what we detect as the active thread number (which doesn't give the value we want here).

Rombur:
When we are inside a parallel region, `get_thread_data()` is not supposed to detect the thread number. It should use thread 0.

https://github.com/kokkos/kokkos/blob/256777d4630daec087b54328634850d857918eba/core/src/OpenMP/Kokkos_OpenMP_Instance.hpp#L94-L96

The condition might be wrong.

masterleinad:
> The condition might be wrong.

From my understanding the logic in `get_thread_data()` is for the case that `Kokkos` causes nested parallelism, not when we have an external OpenMP parallel region.

Rombur:
Then every time we use `get_thread_data()`, there is potentially a bug. We should either fix `get_thread_data()` or check that it does the correct thing everywhere.

masterleinad:
I aam now using `execute_in_serial` inside `get_thread_data()`. The last commit also removes `m_level` but we can discuss if we want to support creating `OpenMP` execution space instances inside `OpenMP` parallel regions.

masterleinad:
We discussed that we want to disallow using an OpenMP execution space instance in a OpenMP parallel region unless OMP_NESTED is set. This implies that we error out if `execute_in_serial` returns `true`.

dalg24:
Is this PR considered a follow up on https://github.com/kokkos/kokkos/pull/7417 ?
Do we want to continue working on it?

masterleinad:
> Is this PR considered a follow up on #7417 ?
> Do we want to continue working on it?

I would still want to discuss disallowing creating OpenMP execution space instances inside OpenMP parallel regions to clean our implementation. That can be here (which has relate changes but not a great pull description) or elsewhere.

dalg24:
> > Is this PR considered a follow up on #7417 ?
> > Do we want to continue working on it?
> 
> I would still want to discuss disallowing creating OpenMP execution space instances inside OpenMP parallel regions to clean our implementation. That can be here (which has relate changes but not a great pull description) or elsewhere.

Please put it on the agenda for next week's dev meeting

Issue #7268, opened 2024-08-27
Serial backend: performance regression


brian-kelley:
The 4.4 release caused a performance regression on the Serial backend, for the Trilinos Intrepid2 Sierra test.
Bisecting showed that the first commit with the regression was the merge of #7080.
The full performance test has a bunch of kernels but the reproducer below is just one of them run in a loop.

1. Compiler: GCC 10.3.0
2. Kokkos release or commit used: good = 892e13c8c4, bad = 3309d9332 (good + #7080)
3. Arch: Intel Cascade Lake (Xeon 8268), but observe similar slowdowns on Broadwell and Skylake
4. Configure: ``cmake -DCMAKE_BUILD_TYPE=Release -DKokkos_ENABLE_SERIAL=ON ../kokkos/example/build_cmake_in_tree``

Good version output: ``Time: 5.14894``
Bad version output: ``Time: 6.99087``

@masterleinad @crtrott This is the issue to continue the thread from slack

~~Full reproducer, to copy into kokkos/example/build_cmake_in_tree/cmake_example.cpp~~
**This reproducer has a bug - it accesses ``_output`` out of bounds in the last dimension. Fixed reproducer in comments below**
```
#include <Kokkos_Core.hpp>
#include <Kokkos_DynRankView.hpp>
#include <iostream>

template <typename OutputViewType, typename leftInputViewType,
          typename rightInputViewType>
struct Functor {
  OutputViewType _output;
  const leftInputViewType _leftInput;
  const rightInputViewType _rightInput;

  const int _iend;
  const int _jend;

  using value_type = typename OutputViewType::value_type;

  KOKKOS_INLINE_FUNCTION
  Functor(OutputViewType output_, leftInputViewType leftInput_,
          rightInputViewType rightInput_)
      : _output(output_),
        _leftInput(leftInput_),
        _rightInput(rightInput_),
        _iend(output_.extent_int(rank(output_) - 1)),
        _jend(rightInput_.extent_int(4 - 1)) {}

  KOKKOS_INLINE_FUNCTION
  void operator()(const int cl, const int bf, const int pt) const {
    apply_matvec_product(cl, bf, pt);
  }

  KOKKOS_FORCEINLINE_FUNCTION
  void apply_matvec_product(const int& cl, const int& bf, const int& pt) const {
    const auto lpt = (_leftInput.extent(1) == 1 ? size_t(0) : pt);
    for (int i = 0; i < _iend; ++i) {
      value_type tmp(0);
[...]

PR #7369, opened 2024-09-23
Serial: Kokkos_ENABLE_ATOMICS_BYPASS removes mutex around parallel regions

cwpearson:
This PR causes `Kokkos_ENABLE_ATOMICS_BYPASS` to remove mutexes introduced around parallel regions in the Serial backend. These mutexes were introduced to preserve Kokkos semantics when multiple threads are hitting the Serial backend. This is not needed in the usual case where a single thread is using the Serial backend.

Experiments in https://github.com/kokkos/kokkos/issues/7268 suggest that the mere presence of some memory operations in these locations can mess up codegen (?) and introduce a 20% slowdown.

These "alternatives" for a lock / unlock were tried and found not to fix the regression.

* incrementing / decrementing `volatile char *`
* atomic increment/decrement
* acquire / release atomic fence

Removing the mutex reliably resolves the problem.

### DIFF
--- a/core/src/Serial/Kokkos_Serial_Parallel_MDRange.hpp
+++ b/core/src/Serial/Kokkos_Serial_Parallel_MDRange.hpp
@@ -44,11 +44,16 @@ class ParallelFor<FunctorType, Kokkos::MDRangePolicy<Traits...>,
 
  public:
   inline void execute() const {
+    // caused a possibly codegen-related slowdown, especially in GCC 9-11
+    // with KOKKOS_ARCH_NATIVE
+    // https://github.com/kokkos/kokkos/issues/7268
+#ifndef KOKKOS_ENABLE_ATOMICS_BYPASS
     // Make sure kernels are running sequentially even when using multiple
     // threads
     auto* internal_instance =
         m_iter.m_rp.space().impl_internal_space_instance();
     std::lock_guard<std::mutex> lock(internal_instance->m_instance_mutex);
+#endif
     this->exec();
   }
   template <typename Policy, typename Functor>
@@ -112,10 +117,15 @@ class ParallelReduce<CombinedFunctorReducerType,
     auto* internal_instance =
         m_iter.m_rp.space().impl_internal_space_instance();
 
+    // caused a possibly codegen-related slowdown, especially in GCC 9-11
+    // with KOKKOS_ARCH_NATIVE
+    // https://github.com/kokkos/kokkos/issues/7268
+#ifndef KOKKOS_ENABLE_ATOMICS_BYPASS
     // Make sure kernels are running sequentially even when using multiple
     // threads, lock resize_thread_team_data
     std::lock_guard<std::mutex> instance_lock(
         internal_instance->m_instance_mutex);
+#endif
     internal_instance->resize_thread_team_data(
         pool_reduce_size, team_reduce_size, team_shared_size,
         thread_local_size);
--- a/core/src/Serial/Kokkos_Serial_Parallel_Range.hpp
+++ b/core/src/Serial/Kokkos_Serial_Parallel_Range.hpp
@@ -49,10 +49,15 @@ class ParallelFor<FunctorType, Kokkos::RangePolicy<Traits...>, Kokkos::Serial> {
 
  public:
   inline void execute() const {
+    // caused a possibly codegen-related slowdown, especially in GCC 9-11
+    // with KOKKOS_ARCH_NATIVE
+    // https://github.com/kokkos/kokkos/issues/7268
+#ifndef KOKKOS_ENABLE_ATOMICS_BYPASS
     // Make sure kernels are running sequentially even when using multiple
     // threads
     auto* internal_instance = m_policy.space().impl_internal_space_instance();
     std::lock_guard<std::mutex> lock(internal_instance->m_instance_mutex);
+#endif
     this->template exec<typename Policy::work_tag>();
   }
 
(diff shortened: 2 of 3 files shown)


masterleinad:
Kokkos_ENABLE_ATOMICS_BYPASS sounds like a misnomer here. Also, you can only enable it if only the `Serial` backend is enabled. I would be curious how widespread these performance regressions are. The analysis in #7268 seems very specific for one particular application/kernel.
I wouldn't be opposed to have a dedicated flag for saying "I don't care about any thread-safety issues" if there is demand.

cwpearson:
Some performance measurements on sapphire rapids with the following reproducer

| GCC | 4.3 | 4.4 | 4.4 + This PR |
|-|-|-|-|
| 9.5.0 | 1.70 | 1.70 | 1.14 |
| 10.2.0 | 1.60 | 2.0 | 1.15 |
| 10.5.0 | 1.60 | 1.95 | 1.15 |
| 11.5.0 | 1.40 | 1.93 | 1.22 |
| 12.4.0 | 1.85 | 1.90 | 1.40 |
| 13.3.0 | 1.75 | 1.81 | 1.43|
| 14.2.0 | 1.70 | 1.79 | 1.43 |

```
 -DCMAKE_BUILD_TYPE=Release \
 -DCMAKE_CXX_FLAGS="-march=native -mtune=native" \
```

```c++
#include <Kokkos_Core.hpp>
#include <Kokkos_DynRankView.hpp>
#include <iostream>

template <typename OutputViewType, typename leftInputViewType,
          typename rightInputViewType>
struct Functor {
  OutputViewType _output;
  const leftInputViewType _leftInput;
  const rightInputViewType _rightInput;

  const int _iend;
  const int _jend;

  using value_type = typename OutputViewType::value_type;

  KOKKOS_INLINE_FUNCTION
  Functor(OutputViewType output_, leftInputViewType leftInput_,
          rightInputViewType rightInput_)
      : _output(output_),
        _leftInput(leftInput_),
        _rightInput(rightInput_),
        _iend(output_.extent_int(3)),
        _jend(rightInput_.extent_int(3)) {}

  KOKKOS_INLINE_FUNCTION
  void operator()(const int cl, const int bf, const int pt) const {
    apply_matvec_product(cl, bf, pt);
  }

  KOKKOS_FORCEINLINE_FUNCTION
  void apply_matvec_product(const int& cl, const int& bf, const int& pt) const {
    for (int i = 0; i < _iend; ++i) {
      value_type tmp(0);
      for (int j = 0; j < _jend; ++j)
        tmp += _leftInput(cl, pt, j, i) * _rightInput(cl, bf, pt, j);
      _output(cl, bf, pt, i) = tmp;
    }
  }
};

template <typename Output, typename Left, typename Right>
void run(const Output& ov, const Left& lv, const Right& rv) {
  using Policy = Kokkos::MDRangePolicy<Kokkos::Serial, Kokkos::Rank<3>,
                                       Kokkos::IndexType<int>>;
  Policy policy({0, 0, 0}, {ov.extent(0), ov.extent(1), ov.extent(2)});
  Functor f(ov, lv, rv);
  Kokkos::parallel_for(policy, f);
}

  {
    using DRV =
        Kokkos::DynRankView<double,
                            Kokkos::Device<Kokkos::Serial, Kokkos::HostSpace>>;
    DRV output("output", 512, 8, 8, 3);
    DRV lv("lv", 512, 8, 3, 3);
    DRV rv("rv", 512, 8, 8, 3);
    Kokkos::Timer t;
    for (int i = 0; i < 6000; i++) run(output, lv, rv);
    double elapsed = t.seconds();
    std::cout << "Time: " << t.seconds() << '\n';
  }
  Kokkos::finalize();
  return 0;
}
```

dalg24 (commented):
As I understand it, the performance regression reported in #7268 was without bypassing atomics so the current approach would not resolve the issue.

cwpearson:
The only successful fix I found so far in https://github.com/kokkos/kokkos/issues/7268 was to remove the mutex entirely - replacing it with various non-atomic and atomic alternatives all keep the regression in place.

This seems like a suitable workaround for the common case where folks are not hitting the serial backend from multiple threads - I will follow up with the actual use-case for https://github.com/kokkos/kokkos/issues/7268 to see if it will work for them

In the mean time we can keep https://github.com/kokkos/kokkos/issues/7268 open, and if anyone has any good ideas I'm happy to keep at it.

dalg24:
Briefly summarizing discussions at the developer meeting:
GCC versions that are problematic are going to be around for a while.  Other compilers like Clang do not have the issue.
Introducing a separate option was considered and we were reluctant to go that way.
It was pointed out that "atomics bypass" might not be a good name for that option which basically is an opt-in making Kokkos constructs unsafe to use in multithreaded contexts.  We decided to keep the name as is.

@brian-kelley is using `-DKokkos_ENABLE_ATOMICS_BYPASS=ON` a workable solution?

dalg24 (approved):
Fine by me if that is good enough for Brian.
Noting that we still acquire the locks when fencing but that I agree with not guarding there if that is not observably affecting performance.

masterleinad (approved):

crtrott (approved):
I believe this is the right thing to do since we can't find a non-interfering mutex alternative.

crtrott:
Failure is unrelated.

ndellingwood:
Should this get a 4.5 changelog entry?

dalg24:
Yes

dalg24:
Under bug fixes or serial backend, something like "remediate performance regression from 4.4"

PR #7372, opened 2024-09-24
Fix node memory value in CEA weekly tests

pzehner:
This PR fixes an incomplete configuration for CEA weekly tests, where the amount of memory was left to default on Ruche, and was too low for `*.view_allocation_large_rank` tests to pass.

### DIFF
--- a/.github/workflows/weekly-cea.yml
+++ b/.github/workflows/weekly-cea.yml
@@ -19,7 +19,7 @@ jobs:
           - name: cuda-a100
             flags: -DKokkos_ENABLE_CUDA=ON -DKokkos_ARCH_AMPERE80=ON
             modules: gcc/11.2.0/gcc-4.8.5 cuda/12.2.1/gcc-11.2.0 cmake/3.28.3/gcc-11.2.0
-            queue: gpua100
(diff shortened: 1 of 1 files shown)


.github/workflows/weekly-cea.yml
```diff
@@ -19,7 +19,7 @@ jobs:
           - name: cuda-a100
             flags: -DKokkos_ENABLE_CUDA=ON -DKokkos_ARCH_AMPERE80=ON
             modules: gcc/11.2.0/gcc-4.8.5 cuda/12.2.1/gcc-11.2.0 cmake/3.28.3/gcc-11.2.0
```
masterleinad, line 21: Just seeing this now `gcc/11.2.0/gcc-4.8.5` looks pretty weird. What do you need it for?

pzehner: To get GCC v11.2.0, otherwise the default GCC of the OS is used, which is v4.8.5. I know, the module names are pretty weird as they also contain what version of GCC that was used to build the underlying program (at least for GCC and CMake, for Cuda I'm not sure).

dalg24 (approved):

PR #7379, opened 2024-09-25
HIP: Check that Kokkos is running on the architecture it was compiled for

Rombur:
Check at initialization that we are running on the architecture that we compiled Kokkos for.


Rombur:
Retest this please

core/src/HIP/Kokkos_HIP.cpp
```diff
@@ -50,6 +50,43 @@ void HIP::impl_initialize(InitializationSettings const& settings) {
...
+#endif
+#ifdef KOKKOS_ARCH_AMD_GFX1030
+  check_architecture("gfx1030");
+#endif
+#ifdef KOKKOS_ARCH_AMD_GFX1100
+  check_architecture("gfx1100");
+#endif
+#ifdef KOKKOS_ARCH_AMD_GFX1103
+  check_architecture("gfx1100");
+#endif
```
dalg24, line 88: How about an approach a la https://github.com/kokkos/kokkos/blob/daaee368fe4b8fbac079e0b398339b8573947b4c/core/src/impl/Kokkos_NvidiaGpuArchitectures.hpp#L17-L56

You define a `KOKKOS_IMPL_ARCH_AMD_GPU` macro as appropriate (integral value GCN arch or full arch name I haven't thought too much about it) which can be leveraged for other purposes as well.

core/src/HIP/Kokkos_HIP.cpp
```diff
@@ -50,6 +50,43 @@ void HIP::impl_initialize(InitializationSettings const& settings) {
   KOKKOS_IMPL_HIP_SAFE_CALL(
       hipGetDeviceProperties(&Impl::HIPInternal::m_deviceProp, hip_device_id));
   KOKKOS_IMPL_HIP_SAFE_CALL(hipSetDevice(hip_device_id));
+
+  // Check that we are running on the expected architecture
+  auto check_architecture = [](const std::string& gfx_name) {
+    std::string arch_name = Impl::HIPInternal::m_deviceProp.gcnArchName;
```
dalg24, line 56: I think these are good use cases for std::string_view (even though obviously it is not critical)

core/src/HIP/Kokkos_HIP.cpp
dalg24: I would recommend you pick integral value in which case your sanity check comes down to
```C++
constexpr std::string_view required_arch =
    "gfx" KOKKOS_IMPL_TOSTRING(KOKKOS_IMPL_ARCH_AMD_GPU);
if (std::string_view detected_arch = Impl::HIPInternal::m_deviceProp.gcnArchName;
    detected_arch != required_arch) {
  // abort
}

Comment (Rombur) [reply core/src/HIP/Kokkos_HIP.cpp:88]: > You define a KOKKOS_IMPL_ARCH_AMD_GPU macro as appropriate 

I am not convinced about this. There is no equivalent to compute capabilities in HIP. You cannot check that you have at least a compute capability of 3.7 and then assume that some features are available. In HIP, there are macros for each individual feature. So I don't see how introducing this macro helps.

> I would recommend you pick integral value

That will not work for MI250 ( `gfx90a`)

Comment (dalg24) [reply core/src/HIP/Kokkos_HIP.cpp:88]: I think there is still value in having a macro that when defines denotes that we are targeting an AMD GPU.

Good point about `gfx90a`.

Comment (Rombur) [reply core/src/HIP/Kokkos_HIP.cpp:88]: We already have `KOKKOS_ARCH_AMD_GPU`

**Review by cwpearson — APPROVED** [review]

File: core/src/HIP/Kokkos_HIP.cpp (hunk 1 above)
Comment (dalg24) [reply core/src/HIP/Kokkos_HIP.cpp:88]: Per our discussion at the office this morning let's define `KOKKOS_ARCH_AMD_GPU` as a GCN arch name literal string

File: core/src/HIP/Kokkos_HIP.cpp [hunk 3]
```diff
@@ -50,6 +50,16 @@ void HIP::impl_initialize(InitializationSettings const& settings) {
   KOKKOS_IMPL_HIP_SAFE_CALL(
       hipGetDeviceProperties(&Impl::HIPInternal::m_deviceProp, hip_device_id));
   KOKKOS_IMPL_HIP_SAFE_CALL(hipSetDevice(hip_device_id));
+
+  // Check that we are running on the expected architecture
+  std::string arch_name = Impl::HIPInternal::m_deviceProp.gcnArchName;
+  if (arch_name != KOKKOS_ARCH_AMD_GPU) {
```
dalg24, line 56:
```suggestion
  if (std::string arch_name = Impl::HIPInternal::m_deviceProp.gcnArchName;
      arch_name != KOKKOS_ARCH_AMD_GPU) {
```
(not formatted)
would prefer a string_view and a narrower scope for arch_name but not blocking

dalg24 (approved):

core/src/HIP/Kokkos_HIP.cpp
Rombur: string_view is annoying to use when creating the error message because you cannot concatenate it with `char`.

crtrott (approved):

core/src/HIP/Kokkos_HIP.cpp
```diff
@@ -50,6 +50,16 @@ void HIP::impl_initialize(InitializationSettings const& settings) {
   KOKKOS_IMPL_HIP_SAFE_CALL(
       hipGetDeviceProperties(&Impl::HIPInternal::m_deviceProp, hip_device_id));
   KOKKOS_IMPL_HIP_SAFE_CALL(hipSetDevice(hip_device_id));
+
+  // Check that we are running on the expected architecture
+  if (std::string arch_name = Impl::HIPInternal::m_deviceProp.gcnArchName;
+      arch_name != KOKKOS_ARCH_AMD_GPU) {
```
dalg24, line 56: Well
```
Kokkos::HIP::initialize ERROR: running kernels compiled for gfx90a on gfx90a:sramecc+:xnack- device.
```

dalg24:
```suggestion
      arch_name.find(KOKKOS_ARCH_AMD_GPU) == 0) {
```

dalg24:
Formatting and HIP builds are passing.  Not waiting for the rest.

Issue #7377, opened 2024-09-25
`hipMemsetAsync` error in scalar deep-copy into host-allocated unmanaged view on MI300A


cwpearson:
On MI300A with `DKokkos_ENABLE_IMPL_HIP_UNIFIED_MEMORY=ON`, it is permitted to do a host allocation (e.g. `calloc`), wrap it in an unmanaged `Kokkos::SharedSpace` view, and run a GPU parallel-for that modifies it.

However, if you `deep_copy` a scalar into that view, Kokkos ultimately tries to use `hipMemsetAsync`, which in rocm <= 6.2.0 comes back with the following error:

```
terminate called after throwing an instance of 'std::runtime_error'
  what():  hipMemsetAsync( dst.data(), 0, dst.size() * sizeof(typename View<T, P...>::value_type), exec_space.hip_stream()) error( hipErrorInvalidValue): invalid argument /usr/WS1/shavano/spack-environments/ATS4CCE_GPU/2.5.2/spack/opt/spack/linux-rhel8-zen3/cce-18.0.0b-magic/kokkos-4.4.00-fwvvrmm/include/HIP/Kokkos_HIP_ZeroMemset.hpp:31
```

I think probably `hipMemsetAsync` is wrongly erroring out when it sees we're trying to operate on a host pointer, even though the GPU can access this data. I've reported this to AMD.

In the mean time, we can probably work around this by using a parallel_for rather than hipMemsetAsync in this specific situation.

```c++
#include <Kokkos_Core.hpp>
#include <Kokkos_OffsetView.hpp>

#include <iostream>

int main(void) {

    static_assert(std::is_same_v<Kokkos::HIPManagedSpace, Kokkos::SharedSpace>, "");

    Kokkos::initialize(); {

    double *ptr = (double*)calloc(1, 10 * sizeof(double));
    Kokkos::fence();

    Kokkos::View<double*, Kokkos::SharedSpace, Kokkos::MemoryUnmanaged> v(ptr, 10);
    Kokkos::fence();

    std::cerr << "parallel_for...\n";
    Kokkos::parallel_for(
         Kokkos::RangePolicy<Kokkos::HIP>(0, 10),
         KOKKOS_LAMBDA(int i) {v[i] = 1.0;}
    );
    Kokkos::fence();

    std::cerr << "deep_copy...\n";
    Kokkos::deep_copy(v, 0.0); // ERROR HERE
    std::cerr << "fence...\n";
    Kokkos::fence();
[...]

PR #7380, opened 2024-09-25
Use a kernel to zero memory on mi300a

cwpearson:
Use a `parallel_for` to implement an alternative to `hipMemsetAsync` for MI300a, since `hipMemsetAsync` doesn't work for host-allocated pointers in ROCm <= 6.2.0 (and possibly later versions).

Potential fix for https://github.com/kokkos/kokkos/issues/7377

### DIFF
--- a/core/src/HIP/Kokkos_HIP_ZeroMemset.cpp
+++ b/core/src/HIP/Kokkos_HIP_ZeroMemset.cpp
@@ -27,7 +27,7 @@ namespace Impl {
 // alternative to hipMemsetAsync, which sets the first `cnt` bytes of `dst` to 0
 void zero_with_hip_kernel(const HIP& exec_space, void* dst, size_t cnt) {
   Kokkos::parallel_for(
-      "zero_with_hip_kernel",
+      "Kokkos::ZeroMemset via parallel_for",
       Kokkos::RangePolicy<Kokkos::HIP>(exec_space, 0, cnt),
       KOKKOS_LAMBDA(size_t i) { static_cast<char*>(dst)[i] = 0; });
 }
--- a/core/src/HIP/Kokkos_HIP_ZeroMemset.cpp
+++ b/core/src/HIP/Kokkos_HIP_ZeroMemset.cpp
@@ -1,0 +1,36 @@
+//@HEADER
+// ************************************************************************
+//
+//                        Kokkos v. 4.0
+//       Copyright (2022) National Technology & Engineering
+//               Solutions of Sandia, LLC (NTESS).
+//
+// Under the terms of Contract DE-NA0003525 with NTESS,
+// the U.S. Government retains certain rights in this software.
+//
+// Part of Kokkos, under the Apache License v2.0 with LLVM Exceptions.
+// See https://kokkos.org/LICENSE for license information.
+// SPDX-License-Identifier: Apache-2.0 WITH LLVM-exception
+//
+//@HEADER
+
+#ifndef KOKKOS_IMPL_PUBLIC_INCLUDE
+#define KOKKOS_IMPL_PUBLIC_INCLUDE
+#endif
+
+#include <HIP/Kokkos_HIP_ZeroMemset.hpp>
+#include <HIP/Kokkos_HIP_ParallelFor_Range.hpp>
+
+namespace Kokkos {
+namespace Impl {
+
+// alternative to hipMemsetAsync, which sets the first `cnt` bytes of `dst` to 0
+void zero_with_hip_kernel(const HIP& exec_space, void* dst, size_t cnt) {
+  Kokkos::parallel_for(
+      "zero_with_hip_kernel",
+      Kokkos::RangePolicy<Kokkos::HIP>(exec_space, 0, cnt),
+      KOKKOS_LAMBDA(size_t i) { static_cast<char*>(dst)[i] = 0; });
+}
+
+}  // namespace Impl
+}  // namespace Kokkos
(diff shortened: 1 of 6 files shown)


core/src/HIP/Kokkos_HIP_ZeroMemset.hpp
```diff
@@ -23,11 +23,22 @@
...
+
 template <>
 struct ZeroMemset<HIP> {
   ZeroMemset(const HIP& exec_space, void* dst, size_t cnt) {
+    // in ROCm <= 6.2.0, hipMemsetAsync on a host-allocated pointer
+    // returns an invalid value error, but accessing the data via a
+    // GPU kernel works.
+#if defined(KOKKOS_ENABLE_IMPL_HIP_UNIFIED_MEMORY) && \
+    defined(KOKKOS_ARCH_AMD_GFX942)
+    zero_with_hip_kernel(exec_space, dst, cnt);
```
masterleinad, line 37: The implementation is so short. Does it make sense to create a new file for it?

cwpearson: Happy to do an `inline` in the header instead if you prefer. Let me know.

core/src/HIP/Kokkos_HIP_ZeroMemset.cpp
```diff
@@ -0,0 +1,36 @@
...
+#include <HIP/Kokkos_HIP_ZeroMemset.hpp>
+#include <HIP/Kokkos_HIP_ParallelFor_Range.hpp>
+
+namespace Kokkos {
+namespace Impl {
+
+// alternative to hipMemsetAsync, which sets the first `cnt` bytes of `dst` to 0
+void zero_with_hip_kernel(const HIP& exec_space, void* dst, size_t cnt) {
+  Kokkos::parallel_for(
+      "zero_with_hip_kernel",
```
dalg24, line 30: Must have `Kokkos::` prefix
maybe spell out `"Kokkos::ZeroMemset via parallel_for"`

core/src/HIP/Kokkos_HIP_ZeroMemset.cpp
```diff
@@ -0,0 +1,36 @@
...
+      Kokkos::RangePolicy<Kokkos::HIP>(exec_space, 0, cnt),
+      KOKKOS_LAMBDA(size_t i) { static_cast<char*>(dst)[i] = 0; });
```
dalg24, line 32: Take `dst` as `std::byte*` so you don't have to cast

dalg24 (changes requested):
Do not forget generated makefiles support

Rombur (approved):

dalg24 (commented):
Also can you come up with a test that is not tool-based that fails on MI300A?

crtrott:
> Also can you come up with a test that is not tool-based that fails on MI300A?

crtrott:
> Also can you come up with a test that is not tool-based that fails on MI300A?

Yeah something like this I believe:
```
int* ptr = new int[N];
View<int*, Kokkos::HipSpace> a(ptr, N);
deep_copy(a, 0);
```

core/src/HIP/Kokkos_HIP_ZeroMemset.cpp
cwpearson: This ended up being kind of clumsy because `std::byte` is not an arithmetic type and also can't implicitly cast `void*` to `byte*` (so there was an explicit cast somewhere anyway), so I left it as is.

https://godbolt.org/z/z46c17hqs

core/unit_test/hip/TestHIP_UnifiedMemory_ZeroMemset.cpp
```diff
@@ -0,0 +1,47 @@
...
+// On MI300a with ROCM <= 6.2.0, hipMemsetAsync was failing with an error when
+// called on host-allocated buffers. The fix was in PR 7380 to use a
+// parallel_for to zero memory
+TEST(hip, unified_memory_zero_memset) {
+#if !defined(KOKKOS_ENABLE_IMPL_HIP_UNIFIED_MEMORY)
+#error this test should only be run with HIP unified memory enabled
+#endif
+
+  constexpr size_t N = 1024 * 1024;  // size doesn't matter
+  int* ptr           = new int[N];
```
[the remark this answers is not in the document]
cwpearson: oops 😊

core/unit_test/hip/TestHIP_UnifiedMemory_ZeroMemset.cpp
```diff
@@ -0,0 +1,43 @@
...
+#if !defined(KOKKOS_ENABLE_IMPL_HIP_UNIFIED_MEMORY)
+#error this test should only be run with HIP unified memory enabled
+#endif
+
+  constexpr size_t N = 1024 * 1024;  // size doesn't matter
+  std::vector<int> v(N, 1);          // initialize to non-zero
+  Kokkos::View<int*, Kokkos::HIPSpace> a(v.data(), N);
+
+  // zero with deep_copy (this is where the error occurs)
+  Kokkos::deep_copy(a, 0);
```
ldh4, line 35: What's the current resulting behavior of this line without the workaround in this PR? Does this line crash or does it not correctly zero-out memory?

cwpearson: Neither, it just errors out with a HIP runtime error, you can see it here: https://github.com/kokkos/kokkos/issues/7377

crtrott (approved):
This is looking good, does what we discussed.

masterleinad (approved):

Issue #21, opened 2015-06-11
Simplify teams + shared memory histogram example to remove vectorization


mhoemmen:
kokkos/example/tutorial/Hierarchical_Parallelism/03_vectorization/vectorization.cpp

It would be helpful for the Kokkos tutorial to have a histogram example that uses shared memory, but doesn't have three levels of parallelism.

crtrott:
This will be addressed in the formal tutorial development project to be done by the end of this summer.

Issue #20, opened 2015-06-09
Kokkos View dimension_() for long returning unsigned int


jdbooth:
Ticket: Regarding issue posted on Kokkos-developers
view.dimiension_0() returning unsigned int, when using long type to index.

Email

Ok that said. It actually does do something incorrect. It seems that somewhere internally an unsigned int is used. 4294967296 gives a view with dimension_0() being reported as 0 while 4294967295 gives back the correct length. Can you open a ticket on github for us?

Thanks
Christian

Test Code:

#include <iostream>
//Kokkos
#include <Kokkos_Core.hpp>
#include <impl/Kokkos_Timer.hpp>

int main(int argc, char* argv[])
{
  typedef long long                         Int;
  typedef double                            Entry;
  typedef Kokkos::OpenMP                    Node;
  typedef Kokkos::View<Entry*, Node>        View;
  typedef Kokkos::ViewAllocateWithoutInitializing NO_INIT;
  Kokkos::initialize(argc, argv);
  Node::print_configuration(std::cout, true);
  Int size = 67449499257;
  View a = View(NO_INIT("a"), size);
  std::cout << "Size: " << size << " View.size: " << a.dimension_0() << std::en\
dl;
  for(Int i = 0; i < size-1; i++)
    {
      a(i) = 0.0;
    }
}//end main()

jgfouca:
I have a pull request in to fix this:
https://github.com/kokkos/kokkos/pull/22

crtrott:
This is merged in and seems to fix the issue.

PR #19, opened 2015-05-20
Kokkos: Modification for task data parallel interface.

kyungjoo-kim:
- member_single() is created in Serial and Threads execution spaces
    - ThreadsTeamMember default constructor is initialized with
      team_size = 1 and league_size = 1.

  - Assignment operator of TaskPolicy is moved from private to public.

PR #7381, opened 2024-09-26
Add TriBITS compatibility tests

dalg24:
As it says on the tin

### DIFF
--- a/core/unit_test/CMakeLists.txt
+++ b/core/unit_test/CMakeLists.txt
@@ -1289,6 +1289,12 @@ KOKKOS_ADD_EXECUTABLE(
   SOURCES UnitTest_CMakeTriBITSCompatibility.cpp
 )
 
+KOKKOS_ADD_TEST(
+  NAME CoreUnitTest_CMakeTriBITSCompatibilityWillFail
+  EXE CoreUnitTest_CMakeTriBITSCompatibility
+)
+set_property(TEST Kokkos_CoreUnitTest_CMakeTriBITSCompatibilityWillFail PROPERTY WILL_FAIL TRUE)
+
 SET(Kokkos_CoreUnitTest_CMakeTriBITSCompatibilityDisable_DISABLE ON)
 KOKKOS_ADD_TEST(
   NAME CoreUnitTest_CMakeTriBITSCompatibilityDisable
--- a/core/unit_test/CMakeLists.txt
+++ b/core/unit_test/CMakeLists.txt
@@ -1284,6 +1284,29 @@ KOKKOS_ADD_EXECUTABLE_AND_TEST(
   ARGS "one 2 THREE"
 )
 
+KOKKOS_ADD_EXECUTABLE(
+  CoreUnitTest_CMakeTriBITSCompatibility
+  SOURCES UnitTest_CMakeTriBITSCompatibility.cpp
+)
+
+SET(Kokkos_CoreUnitTest_CMakeTriBITSCompatibilityDisable_DISABLE ON)
+KOKKOS_ADD_TEST(
+  NAME CoreUnitTest_CMakeTriBITSCompatibilityDisable
+  EXE CoreUnitTest_CMakeTriBITSCompatibility
+)
+
+SET(Kokkos_CoreUnitTest_CMakeTriBITSCompatibilityExtraArgs_EXTRA_ARGS "--kokkos-test-tribits-compatibility=1")
+KOKKOS_ADD_TEST(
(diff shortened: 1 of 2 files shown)


core/unit_test/CMakeLists.txt
```diff
@@ -1284,6 +1284,29 @@ KOKKOS_ADD_EXECUTABLE_AND_TEST(
...
+KOKKOS_ADD_TEST(
+  NAME CoreUnitTest_CMakeTriBITSCompatibilityExtraArgs
+  EXE CoreUnitTest_CMakeTriBITSCompatibility
+)
+
+SET(Kokkos_CoreUnitTest_CMakeTriBITSCompatibilityEnvironment_ENVIRONMENT "KOKKOS_TEST_TRIBITS_COMPATIBILITY=1")
+KOKKOS_ADD_TEST(
+  NAME CoreUnitTest_CMakeTriBITSCompatibilityEnvironment
+  EXE CoreUnitTest_CMakeTriBITSCompatibility
+)
```
masterleinad, line 1308: Setting `ENVIRONMENT` via a CMake variable is not supported by `TriBITS` but only by `tribits_add_test` directly. Do we want to keep it? I now lean slightly to removing until there is demand.

Also, we could add `<fullTestName>_SET_DISABLED_AND_MSG` for better compatibility with `TriBITS`.

dalg24: I know it is not a TriBITS option but could not think of a better way to categorize it.
I could add a comment if you want me to.

As far as _SET_DISABLED_AND_MSG is concerned, I was not planning to tackle it here.

core/unit_test/CMakeLists.txt
```diff
@@ -1284,6 +1284,29 @@ KOKKOS_ADD_EXECUTABLE_AND_TEST(
   ARGS "one 2 THREE"
 )
 
+KOKKOS_ADD_EXECUTABLE(
+  CoreUnitTest_CMakeTriBITSCompatibility
+  SOURCES UnitTest_CMakeTriBITSCompatibility.cpp
+)
+
+SET(Kokkos_CoreUnitTest_CMakeTriBITSCompatibilityDisable_DISABLE ON)
```
ndellingwood, line 1292: To check my understanding, the confirmation that this test passes (i.e. does not run thanks to the _DISABLE property) is that the unit test will not terminate with `EXIT_FAILURE`, even though the test lacks the required  `KOKKOS_TEST_TRIBITS_COMPATIBILITY` or `--kokkos-test-tribits-compatibility` matches required for success, does that sound correct?

ndellingwood (approved):
Nice tests, thanks for adding

core/unit_test/CMakeLists.txt
ndellingwood: I'm referring to L1292-1296, my highlight block didn't carry over when I commented

dalg24: Correct, if we actually tried to run that command it would fail.

dalg24: I added a test with the WILL_FAIL property to ensure that it would indeed fail

dalg24:
```
$ ctest --test-dir cmake-build-default -C Debug -R TriBITS

Internal ctest changing into directory: /Users/qdi/CLionProjects/kokkos/cmake-build-default
Test project /Users/qdi/CLionProjects/kokkos/cmake-build-default
    Start 19: Kokkos_CoreUnitTest_CMakeTriBITSCompatibilityWillFail
1/3 Test #19: Kokkos_CoreUnitTest_CMakeTriBITSCompatibilityWillFail ......   Passed    0.62 sec
    Start 20: Kokkos_CoreUnitTest_CMakeTriBITSCompatibilityExtraArgs
2/3 Test #20: Kokkos_CoreUnitTest_CMakeTriBITSCompatibilityExtraArgs .....   Passed    0.02 sec
    Start 21: Kokkos_CoreUnitTest_CMakeTriBITSCompatibilityEnvironment
3/3 Test #21: Kokkos_CoreUnitTest_CMakeTriBITSCompatibilityEnvironment ...   Passed    0.01 sec

100% tests passed, 0 tests failed out of 3

Label Time Summary:
Kokkos    =   0.64 sec*proc (3 tests)

Total Test time (real) =   0.66 sec

dalg24:
Failure was unrelated (GPU not detected at runtime in one of the CUDA builds)

crtrott (approved):

PR #7386, opened 2024-09-27
Add support for no exceptions

### LINKED ISSUE #1074: make[1]: *** No rule to make target 'Kokkos_spinwait.o', needed by 'libkokkos.a'.  Stop.
# Problem
The latest version is broken.
```
$ git log
commit 2b7e9c2013d70660512daae6d084dc8254a8a45d
Merge: 54eb75c 32fb8ee
Author: Dan Ibanez <daibane@sandia.gov>
Date:   Wed Aug 16 16:56:21 2017 -0600

    Merge develop for 2.04.00
```

# Reproducer
```
KOKKOS/github$ git reset --hard ; git checkout . ; git clean -dfx ; mkdir -p build && cd build && ../generate_makefile.bash --kokkos-path=.. --prefix=/opt/kokkos/intel/openmp --with-openmp --arch=HSW --compiler=icpc && make test && make install && cd ..
```

# Error
```
icpc -I./ -I/home/jrhammon/Work/DOE/KOKKOS/github/core/src -I/home/jrhammon/Work/DOE/KOKKOS/github/containers/src -I/home/jrhammon/Work/DOE/KOKKOS/github/algorithms/src  --std=c++11 -xCORE-AVX2 -fopenmp -O3 -c /home/jrhammon/Work/DOE/KOKKOS/github/core/src/impl/Kokkos_Error.cpp
make[1]: *** No rule to make target 'Kokkos_spinwait.o', needed by 'libkokkos.a'.  Stop.
make[1]: Leaving directory '/home/jrhammon/Work/DOE/KOKKOS/github/build/core'
Makefile:12: recipe for target 'kokkoslib-test' failed
make: *** [kokkoslib-test] Error 2
```

It is busted with other compilers, too, e.g. GCC 7.

dalg24:
Final step to fix #1074


cmake/kokkos_enable_options.cmake
```diff
@@ -86,6 +86,8 @@ mark_as_advanced(Kokkos_ENABLE_IMPL_MDSPAN)
 mark_as_advanced(Kokkos_ENABLE_MDSPAN_EXTERNAL)
 mark_as_advanced(Kokkos_ENABLE_IMPL_SKIP_COMPILER_MDSPAN)
 
+KOKKOS_ENABLE_OPTION(NO_EXCEPTIONS ON BOOL "Whether to ban the use of exceptions")
```
dalg24, line 89: Planing to mark it as advanced.  Let me know if you disagree.

dalg24: (And will default to OFF obviously)

cedricchevalier19: Good idea to mark it as advanced.

masterleinad: I find `NO_EXCEPTIONS=OFF/ON` harder to read than `USE_EXCEPTIONS=ON/OFF` but I can also understand this choice since the flag is `-fno-exceptions` for a lot of compilers.

masterleinad:
https://godbolt.org/z/er5Ycb7M6 shows errors with thrust and -fno-exceptions for all CUDA versions and that is consistent with https://github.com/NVIDIA/thrust/blob/756c5afc0750f1413da05bd2b6505180e84c53d4/thrust/system/detail/system_error.inl#L85-L106.

```
# nvcc -x cu -std=c++17 -Xcompiler -fno-exceptions blabla.cpp && ./a.out 
/opt/nvidia/hpc_sdk/Linux_x86_64/23.7/cuda/12.2/include/thrust/system/detail/system_error.inl: In member function 'virtual const char* thrust::system::system_error::what() const':
/opt/nvidia/hpc_sdk/Linux_x86_64/23.7/cuda/12.2/include/thrust/system/detail/system_error.inl:99:8: error: exception handling disabled, use '-fexceptions' to enable
   99 |     catch(...)
      |        ^~~
```
with nvcc-12.2 and
```
# nvcc -ccbin nvc++ -x cu -std=c++17 -Xcompiler -fno-exceptions blabla.cpp && ./a.out 
"/opt/nvidia/hpc_sdk/Linux_x86_64/23.7/cuda/12.2/include/thrust/system/detail/system_error.inl", line 90: error: support for exception handling is disabled
  try 
  ^

"/opt/nvidia/hpc_sdk/Linux_x86_64/23.7/cuda/12.2/include/thrust/system/cuda/detail/util.h", line 237: error: support for exception handling is disabled
  { throw system::system_error(status, thrust::cuda_category()); } ; 
    ^

"/opt/nvidia/hpc_sdk/Linux_x86_64/23.7/cuda/12.2/include/thrust/system/cuda/detail/util.h", line 281: error: support for exception handling is disabled
  { throw system::system_error(status, thrust::cuda_category(), msg); } ; 
    ^
```
with `nvc++`

https://github.com/NVIDIA/thrust/issues/1147 says (back in 2020) they don’t care about supporting -fno-expections in thrust (if the compiler wouldn’t just ignore try-catch  blocks).

masterleinad:
Currently, this pull request doesn't add flags for the `Makefile` build and that's the reason why CI passes. Adding `-Xcompiler -fno-exceptions` gives:
```
/app/kokkos_new/core/unit_test/TestViewAPI.hpp:1366:3: error: exception handling disabled, use '-fexceptions' to enable
 1366 |     } catch (const std::bad_alloc &) {
      |   ^ ~~~
```
because we didn't define `KOKKOS_ENABLE_NO_EXCEPTIONS`

masterleinad:
It seems we should just not test with -fno-exceptions with nvcc.

masterleinad:
`SYCL` seems to be fine even on the testbeds. `-fno-excepotions` is visible in the compilation flags.

cmake/kokkos_tribits.cmake
```diff
@@ -116,6 +116,13 @@ function(KOKKOS_ADD_EXECUTABLE_AND_TEST ROOT_NAME)
       target_compile_options(${PACKAGE_NAME}_${ROOT_NAME} PRIVATE "-fno-rtti")
     endif()
   endif()
+  if(KOKKOS_ENABLE_NO_EXCEPTIONS AND NOT KOKKOS_CXX_COMPILER_ID STREQUAL "NVIDIA")
```
dalg24, line 119: Please add a comment that it does not work with CUDA on Microsoft Windows and that the Thrust library that we use in particular for sorting algorithms on the device with CUDA does not support building without exceptions as of today

core/src/impl/Kokkos_Error.cpp
```diff
@@ -25,8 +25,21 @@
 #include <Kokkos_Core.hpp>  // show_warnings
 #include <impl/Kokkos_Error.hpp>
 
+#ifndef KOKKOS_COMPILER_MSVC
+#define KOKKOS_IMPL_BUILTIN_UNREACHABLE() __builtin_unreachable()
+#else
+#define KOKKOS_IMPL_BUILTIN_UNREACHABLE() \
+  do {                                    \
+  } while (false)
```
dalg24, line 33: Now that I look at this again I wonder if that is not potentially confusing as one might think I had written an infinite loop as a hint to the compiler that it cannot be reached but my actual intent was to have some dummy code that needs to be followed by an underscore.
Maybe I should do `static_assert(true, "no-op to require trailing semicolon")` instead.  Thoughts?

cmake/kokkos_enable_options.cmake
```diff
@@ -103,6 +103,8 @@ mark_as_advanced(Kokkos_ENABLE_IMPL_MDSPAN)
 mark_as_advanced(Kokkos_ENABLE_MDSPAN_EXTERNAL)
 mark_as_advanced(Kokkos_ENABLE_IMPL_SKIP_COMPILER_MDSPAN)
 
+kokkos_enable_option(NO_EXCEPTIONS ON BOOL "Whether to ban the use of exceptions")
```
masterleinad, line 106: The current wording doesn't make it clear to me that the scope is limited to `Kokkos` itself and we wouldn't just add `-fno-excpetions` to the compiler flags.
```suggestion
kokkos_enable_option(NO_EXCEPTIONS ON BOOL "Whether to allow Kokkos to use exceptions")
```

core/src/impl/Kokkos_Error.cpp
masterleinad: We are using `__builtin_unreachable` in a bunch of other places so it makes sense to me to define it in a more central place.

masterleinad: https://en.cppreference.com/w/cpp/utility/unreachable shows that you could try using
```
__assume(false)
```
for MSVC instead.

dalg24: Does not belong to this work

PR #7387, opened 2024-09-27
feat: enforce `override` for `virtual` methods

romintomasetti:
This PR adds `-Wsuggest-override` to the compiler warnings.

In addition to legitimately missing `override`, `gcc-8` also warns for (implicitly declared) virtual functions that are already marked `final`, `i.e.` `gcc-8` requires `override final` in many places though `final` was already fine. I just decided to follow its suggestions to make the CI happy.

Also note that I had to make the `tpl/gtest` a `SYSTEM` include directory, because it does not make sense to warn for anything happening in a TPL. I removed the *global include directory* in favor of the *target-based* one. It is more `CMake`-target compliant.


dalg24 (approved):
Fine by me if all the compilers in the CI eat that flag

cmake/kokkos_arch.cmake
```diff
@@ -147,6 +147,7 @@ if(KOKKOS_ENABLE_COMPILER_WARNINGS)
       "-Wsign-compare"
       "-Wtype-limits"
       "-Wuninitialized"
+      -Wsuggest-override
```
dalg24, line 150:
```suggestion
      "-Wsuggest-override"
```

romintomasetti: The quotes are superfluous in this case. It would have been nice if your `CMake` tidy had told you that :angel:

dalg24: I know they are but this is about consistency

romintomasetti: OK.

core/unit_test/CMakeLists.txt
```diff
@@ -6,7 +6,7 @@ if(NOT GTest_FOUND) # fallback to internal gtest
   set(GTEST_SOURCE_DIR ${Kokkos_SOURCE_DIR}/tpls/gtest)
 
   #need here for tribits
-  kokkos_include_directories(${GTEST_SOURCE_DIR})
+  kokkos_include_directories(SYSTEM ${GTEST_SOURCE_DIR})
```
romintomasetti, line 9: `gcc 8` warns inside `gtest.h`. I don't want to modify a TPL.

If we make the `gtest` directory a system include directory, compilers should not warn about it.

cmake/fake_tribits.cmake
```diff
@@ -185,8 +185,12 @@ function(KOKKOS_ADD_TEST_LIBRARY NAME)
 endfunction()
 
 function(KOKKOS_INCLUDE_DIRECTORIES)
-  cmake_parse_arguments(INC "REQUIRED_DURING_INSTALLATION_TESTING" "" "" ${ARGN})
-  include_directories(${INC_UNPARSED_ARGUMENTS})
+  cmake_parse_arguments(INC "REQUIRED_DURING_INSTALLATION_TESTING;SYSTEM" "" "" ${ARGN})
```
romintomasetti, line 188: 1. `kokkos_include_directories` seems to be a (useless?) wrapper for `include_directories`.
2. Using `include_directories` is a bad idea, it goes against `CMake` target principles.
3. `REQUIRED_DURING_INSTALLATION_TESTING` seems a dead option.

dalg24 (approved):
Fine

cmake/fake_tribits.cmake
romintomasetti: In the end, I don't need to modify this function but still, it seems useless. @masterleinad What do you think ? Is it for some `TriBits` compatibility reason ?

masterleinad: The point of removing the special path for `Trilinos`/`TriBITS` builds was to be able to start cleaning up our `CMake` files. There are only a couple of lines left that we need for `TriBITS` interoperability but this shouldn't be one of them.

core/src/impl/Kokkos_ExecSpaceManager.hpp
```diff
@@ -123,14 +123,14 @@ template <class ExecutionSpace>
 struct ExecSpaceDerived : ExecSpaceBase {
   static_assert(check_valid_execution_space<ExecutionSpace>());
   static_assert(check_is_regular<ExecutionSpace>());
-  void initialize(InitializationSettings const& settings) final {
+  void initialize(InitializationSettings const& settings) override final {
```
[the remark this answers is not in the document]
dalg24: I hadn't paid attention to the override+final
Not loving it

core/unit_test/CMakeLists.txt
```diff
@@ -6,12 +6,11 @@ if(NOT GTest_FOUND) # fallback to internal gtest
   set(GTEST_SOURCE_DIR ${Kokkos_SOURCE_DIR}/tpls/gtest)
 
   #need here for tribits
```
dalg24, line 8:
```suggestion
```

cmake/fake_tribits.cmake
romintomasetti: OK, thanks for the insight :+1:

core/src/impl/Kokkos_ExecSpaceManager.hpp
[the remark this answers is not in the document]
romintomasetti: @dalg24 I can introduce a macro like that:
```c++
#if compiler is gcc and version is less than 9.2
    #define MARK_OVERRIDE_WHEN_ALREADY_FINAL override
#else
    #define MARK_OVERRIDE_WHEN_ALREADY_FINAL
#endif
```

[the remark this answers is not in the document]
dalg24: Not worth the trouble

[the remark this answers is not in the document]
dalg24: You could add a comment or do nothing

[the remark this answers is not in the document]
romintomasetti: Ok, let's do nothing :angel:

romintomasetti:
My guess is that when you're creating some "imported TPL", you should add interface include directories as `SYSTEM`. Haven't tested locally though, but it would make sense.

masterleinad:
Somehow the las push didn't trigger jenkins CI. The latest build there is https://cloud.cees.ornl.gov/jenkins-ci/job/Kokkos/job/PR-7387/10/ with SHA 600bfac4ecc964471309f9143ff8d75d906702c1.

dalg24:
Retest this please

tpadioleau (dismissed):

cmake/kokkos_functions.cmake
```diff
@@ -347,7 +347,7 @@ macro(kokkos_create_imported_tpl NAME)
       target_link_libraries(${NAME} INTERFACE ${TPL_LINK_LIBRARIES})
     endif()
     if(TPL_INCLUDES)
-      target_include_directories(${NAME} INTERFACE ${TPL_INCLUDES})
+      target_include_directories(${NAME} SYSTEM INTERFACE ${TPL_INCLUDES})
```
romintomasetti, line 350: @dalg24 This change was introduce in an attempt to silence some `Thrust` warnings, did not work. Should I revert ?

tpadioleau:
> It seems that the remaining warnings for `Cuda 11.6.2` job:
> 
> ```shell
> /usr/local/cuda/bin/../targets/x86_64-linux/include/thrust/system/detail/bad_alloc.h:46:21: error: 'virtual const char* thrust::system::detail::bad_alloc::what() const' can be marked override [-Werror=suggest-override]
>    46 |     inline virtual const char *what(void) const throw()
>       |                     ^~~~
> cc1plus: all warnings being treated as errors
> ```
> 
> are generated by includes from `Thrust`. Not much we can do about it except adding ignore pragmas I guess...

I had a look at it, I don't understand the situation. I tried to reproduce what happens in the jenkins, building the docker image and building the cmake target `Kokkos_HeaderSelfContained_Kokkos_AnonymousSpace_hpp` but no warning for me. Am I missing something ?

tpadioleau:
My bad, the problematic target was `Kokkos_HeaderSelfContained_Kokkos_Sort_hpp`. No need for a `find_package(Thrust ...)` ? Is the compiler supposed to bring it ?

masterleinad:
> My bad, the problematic target was `Kokkos_HeaderSelfContained_Kokkos_Sort_hpp`. No need for a `find_package(Thrust ...)` ? Is the compiler supposed to bring it ?

Yes, and the headers are implicitly included via `-I` not via`-isystem`. https://github.com/NVIDIA/cccl/pull/531 made help with that for newer `CUDA` versions.

tpadioleau:
> Yes, and the headers are implicitly included via `-I` not via`-isystem`. [NVIDIA/cccl#531](https://github.com/NVIDIA/cccl/pull/531) made help with that for newer `CUDA` versions.

I see thanks, too bad. So the only solution for now is to add `#pragma GCC diagnostic ignored "-Wsuggest-override"` ? As it was done for `-Wshadow` I guess ?

cmake/kokkos_functions.cmake
crtrott: I am not sure it's a good idea for us to inject "system" things into stuff downstream users get. Is this doing that?

crtrott (commented):
Not sure about that system thing for Thrust, considering that it didn't fix the thing it was supposed to fix maybe revert?

tpadioleau:
I suggest
```
diff --git a/algorithms/src/sorting/impl/Kokkos_SortByKeyImpl.hpp b/algorithms/src/sorting/impl/Kokkos_SortByKeyImpl.hpp
index 55d2c9c59..2a8f761d9 100644
--- a/algorithms/src/sorting/impl/Kokkos_SortByKeyImpl.hpp
+++ b/algorithms/src/sorting/impl/Kokkos_SortByKeyImpl.hpp
@@ -30,6 +30,7 @@
 
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wshadow"
+#pragma GCC diagnostic ignored "-Wsuggest-override"
 
 #if defined(KOKKOS_COMPILER_CLANG)
 // Some versions of Clang fail to compile Thrust, failing with errors like
diff --git a/algorithms/src/sorting/impl/Kokkos_SortImpl.hpp b/algorithms/src/sorting/impl/Kokkos_SortImpl.hpp
index ad0644cea..83f155555 100644
--- a/algorithms/src/sorting/impl/Kokkos_SortImpl.hpp
+++ b/algorithms/src/sorting/impl/Kokkos_SortImpl.hpp
@@ -34,6 +34,7 @@
 
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wshadow"
+#pragma GCC diagnostic ignored "-Wsuggest-override"
 
 #if defined(KOKKOS_COMPILER_CLANG)
 // Some versions of Clang fail to compile Thrust, failing with errors like
```

cmake/kokkos_functions.cmake
romintomasetti: @crtrott Indeed, this does not solve the problem of the `thrust` related warnings. I'll revert.

romintomasetti: Done, closing :heavy_check_mark:

romintomasetti:
@tpadioleau Indeed, it's the only way forward. I made the changes. Let's see !

dalg24:
Agreeing that failures were unrelated (NVHPC compiler segfaulting and GPU not detected in OpenACC builds)

dalg24 (approved):
Fine

tpadioleau (approved):
LGTM

PR #7401, opened 2024-10-03
ViewTraits Interface Cleanup Part 2

janciesko:
Partially addressing https://github.com/kokkos/kokkos/issues/7350.

Changes:
- `array_layout` -> deprecated
-  Adds `layout_type` (This is now consistent with layout types in mdspan)
-  Unit tests testing for those types -> deprecated


core/src/Kokkos_ScratchSpace.hpp
```diff
@@ -65,8 +65,13 @@ class ScratchMemorySpace {
...
 
-  using array_layout = typename ExecSpace::array_layout;
-  using size_type    = typename ExecSpace::size_type;
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  using array_layout KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use layout_type instead.") = typename ExecSpace::array_layout;
+#endif
+  using layout_type = typename ExecSpace::layout_type;
+
+  using size_type = typename ExecSpace::size_type;
```
cwpearson, line 74:
```suggestion
  using layout_type = typename ExecSpace::layout_type;
  using size_type   = typename ExecSpace::size_type;
```

cwpearson (approved):
One optional suggestion

containers/src/Kokkos_DynRankView.hpp
```diff
@@ -439,7 +439,11 @@ class DynRankView : private View<DataType*******, Properties...> {
   using const_value_type     = typename view_type::const_value_type;
   using non_const_value_type = typename view_type::non_const_value_type;
   using traits               = typename view_type::traits;
-  using array_layout         = typename view_type::array_layout;
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  using array_layout KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use layout_type instead.") = typename view_type::array_layout;
```
science-enthusiast, line 444: `typename view_type::array_layout` --> `typename view_type::layout_type`

algorithms/src/std_algorithms/impl/Kokkos_RandomAccessIterator.hpp
```diff
@@ -71,7 +71,18 @@ class RandomAccessIterator<::Kokkos::View<DataType, Args...>> {
...
+=======
+                    (std::is_same_v<typename view_type::traits::layout_type,
+                                    Kokkos::LayoutLeft> ||
+                     std::is_same_v<typename view_type::traits::layout_type,
+                                    Kokkos::LayoutRight> ||
+                     std::is_same_v<typename view_type::traits::layout_type,
+                                    Kokkos::LayoutStride>),
+                "RandomAccessIterator only supports 1D Views with LayoutLeft, "
+                "LayoutRight, LayoutStride.");
+>>>>>>> 86c5f1e38 (Depracate array_layout in favor of layout_type)
```
science-enthusiast, line 85: Merge conflict to be fixed.

core/src/View/Kokkos_ViewTraits.hpp
```diff
@@ -275,10 +280,14 @@ struct ViewTraits<void> {
...
   using memory_space    = void;
   using HostMirrorSpace = void;
-  using array_layout    = void;
-  using memory_traits   = void;
-  using specialize      = void;
-  using hooks_policy    = void;
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  using array_layout KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use layout_type instead.") = void;
+#endif
```
science-enthusiast, line 286: Better to move this after `layout_type` alias definition and assign to `layout_type`.

core/src/View/Kokkos_ViewTraits.hpp
```diff
@@ -287,10 +296,13 @@ struct ViewTraits<void, void, Prop...> {
   using execution_space = typename ViewTraits<void, Prop...>::execution_space;
   using memory_space    = typename ViewTraits<void, Prop...>::memory_space;
   using HostMirrorSpace = typename ViewTraits<void, Prop...>::HostMirrorSpace;
-  using array_layout    = typename ViewTraits<void, Prop...>::array_layout;
-  using memory_traits   = typename ViewTraits<void, Prop...>::memory_traits;
-  using specialize      = typename ViewTraits<void, Prop...>::specialize;
-  using hooks_policy    = typename ViewTraits<void, Prop...>::hooks_policy;
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  using array_layout = typename ViewTraits<void, Prop...>::array_layout;
```
science-enthusiast, line 300: Needs to be `  using array_layout KOKKOS_DEPRECATED_WITH_COMMENT("Use layout_type instead.") = layout_type;` and placed after `layout_type` alias is defined.

core/src/View/Kokkos_ViewTraits.hpp
```diff
@@ -300,24 +312,34 @@ struct ViewTraits<
...
   using memory_space    = typename ViewTraits<void, Prop...>::memory_space;
   using HostMirrorSpace = typename ViewTraits<void, Prop...>::HostMirrorSpace;
-  using array_layout    = typename ViewTraits<void, Prop...>::array_layout;
-  using memory_traits   = typename ViewTraits<void, Prop...>::memory_traits;
-  using specialize      = typename ViewTraits<void, Prop...>::specialize;
-  using hooks_policy    = HooksPolicy;
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  using array_layout KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use layout_type instead.") =
+      typename ViewTraits<void, Prop...>::array_layout;
```
science-enthusiast, line 318: Same as earlier case.

Needs to be  using array_layout KOKKOS_DEPRECATED_WITH_COMMENT("Use layout_type instead.") = layout_type; and placed after layout_type alias is defined.

core/src/View/Kokkos_ViewTraits.hpp
```diff
@@ -300,24 +312,34 @@ struct ViewTraits<
...
   using memory_space    = typename ViewTraits<void, Prop...>::memory_space;
   using HostMirrorSpace = typename ViewTraits<void, Prop...>::HostMirrorSpace;
-  using array_layout    = ArrayLayout;
-  using memory_traits   = typename ViewTraits<void, Prop...>::memory_traits;
-  using specialize      = typename ViewTraits<void, Prop...>::specialize;
-  using hooks_policy    = typename ViewTraits<void, Prop...>::hooks_policy;
+
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  using array_layout KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use layout_type instead.") = LayoutType;
```
science-enthusiast, line 337: Same as earlier cases.

Needs to be `using array_layout KOKKOS_DEPRECATED_WITH_COMMENT("Use layout_type instead.") = layout_type;` and placed after `layout_type` alias is defined.

core/src/View/Kokkos_ViewTraits.hpp
```diff
@@ -332,15 +354,18 @@ struct ViewTraits<std::enable_if_t<Kokkos::is_space<Space>::value>, Space,
...
       "Only one View Execution or Memory Space template argument");
 
   using execution_space = typename Space::execution_space;
   using memory_space    = typename Space::memory_space;
   using HostMirrorSpace =
       typename Kokkos::Impl::HostMirror<Space>::Space::memory_space;
-  using array_layout  = typename execution_space::array_layout;
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  using array_layout KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use layout_type instead.") = typename execution_space::array_layout;
```
science-enthusiast, line 366: Same as earlier cases.

Needs to be `using array_layout KOKKOS_DEPRECATED_WITH_COMMENT("Use layout_type instead.") = layout_type;` and placed after `layout_type` alias is defined.

core/src/View/Kokkos_ViewUniformType.hpp
```diff
@@ -36,17 +36,29 @@ struct ViewScalarToDataType<ScalarType, 0> {
 
 template <class LayoutType, int Rank>
 struct ViewUniformLayout {
-  using array_layout = LayoutType;
```
science-enthusiast, line 39: This is the same comment as in `core/src/View/Kokkos_ViewTraits.hpp`. The logic around lines 39, 44 and 49 should be changed in the same manner as it is done around line 68:

The type alias `array_layout` needs to be defined as `using array_layout KOKKOS_DEPRECATED_WITH_COMMENT("Use layout_type instead.") = layout_type;` and needs to be placed after `layout_type` alias.

core/src/View/MDSpan/Kokkos_MDSpan_Layout.hpp
```diff
@@ -50,37 +50,37 @@ template <size_t Pad>
...
 template <>
-struct LayoutFromArrayLayout<LayoutStride> {
+struct LayoutFromLayoutType<LayoutStride> {
   using type = layout_stride;
 };
 
-template <class ArrayLayout, class MDSpanType>
-KOKKOS_INLINE_FUNCTION auto array_layout_from_mapping(
+template <class LayoutType, class MDSpanType>
+KOKKOS_INLINE_FUNCTION auto layout_type_from_mapping(
```
science-enthusiast, line 61: `Impl::array_layout_from_mapping` is called in `core/src/Kokkos_View.hpp` and `core/src/View/Kokkos_ViewLegacy.hpp`. The call in `core/src/Kokkos_View.hpp` still needs to be changed.

containers/src/Kokkos_DynRankView.hpp
```diff
@@ -487,24 +491,24 @@ class DynRankView : private View<DataType*******, Properties...> {
   // Types below - at least the HostMirror requires the value_type, NOT the rank
   // 7 data_type of the traits
 
-  /** \brief  Compatible view of array of scalar types */
+  /** \brief  Compatible view of data type */
   using array_type = DynRankView<
-      typename drvtraits::scalar_array_type, typename drvtraits::array_layout,
+      typename drvtraits::data_type, typename drvtraits::layout_type,
```
science-enthusiast, line 496: A similar change has to be around line 720. `operator[]` . Phalanx.

containers/src/Kokkos_DynRankView.hpp
```diff
@@ -1457,13 +1460,20 @@ struct MirrorDRViewType {
...
   };
+
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
   // The array_layout
-  using array_layout = typename src_view_type::array_layout;
+  using array_layout KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use layout_type instead.") = typename src_view_type::array_layout;
+#endif
+  // The layout type
+  using layout_type = typename src_view_type::layout_type;
```
science-enthusiast, line 1470: Like line 442: `array_layout` alias to be defined as `using array_layout KOKKOS_DEPRECATED_WITH_COMMENT("Use layout_type instead.") = layout_type;` and after definition of `layout_type` alias.

containers/src/Kokkos_DynamicView.hpp
```diff
@@ -266,12 +266,16 @@ class DynamicView : public Kokkos::ViewTraits<DataType, P...> {
...
  public:
-  //----------------------------------------------------------------------
-
+//----------------------------------------------------------------------
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
   /** \brief  Compatible view of array of scalar types */
-  using array_type =
+  using array_type KOKKOS_DEPRECATED_WITH_COMMENT("Use type instead.") =
+      DynamicView<typename traits::scalar_array_type,
+                  typename traits::device_type>;
```
science-enthusiast, line 274: `using array_type KOKKOS_DEPRECATED_WITH_COMMENT("Use type instead.") = type;` after the definition of the alias `type`.

containers/src/Kokkos_DynamicView.hpp
```diff
@@ -577,14 +581,19 @@ struct MirrorDynamicViewType {
...
         std::is_same_v<memory_space, typename src_view_type::memory_space>
   };
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
   // The array_layout
-  using array_layout = typename src_view_type::array_layout;
+  using array_layout KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use layout_type instead.") = typename src_view_type::array_layout;
+#endif
+  // The layout type
+  using layout_type = typename src_view_type::layout_type;
```
science-enthusiast, line 590: Similar to other places, `using array_layout KOKKOS_DEPRECATED_WITH_COMMENT(
      "Use layout_type instead.") = layout_type;`, after the definition of `layout_type`.

janciesko:
@science-enthusiast Thanks for the review. I've pushed a commit implementing the suggestions.

core/src/Cuda/Kokkos_Cuda.hpp
```diff
@@ -115,7 +115,11 @@ class Cuda {
   using size_type = memory_space::size_type;
 
   //! This execution space's preferred array layout.
-  using array_layout = LayoutLeft;
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  using array_layout KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use layout_type instead.") = LayoutLeft;
+#endif
+  using layout_type = LayoutLeft;
```
science-enthusiast: Again, here and in `core/src/HIP/Kokkos_HIP.hpp`, `core/src/HPX/Kokkos_HPX.hpp`, `layout_type` should be defined first and then we should have:

> #ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
>   using array_layout KOKKOS_DEPRECATED_WITH_COMMENT(
>       "Use layout_type instead.") = layout_type;
> #endif

core/src/Kokkos_Crs.hpp
```diff
@@ -65,18 +65,23 @@ class Crs {
...
 
  public:
-  using data_type       = DataType;
-  using array_layout    = typename traits::array_layout;
+  using data_type = DataType;
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  using array_layout KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use layout_type instead.") = typename traits::array_layout;
+#endif
+  using layout_type = typename traits::layout_type;
```
science-enthusiast: using layout_type = typename traits::layout_type;
  using array_layout KOKKOS_DEPRECATED_WITH_COMMENT(
      "Use layout_type instead.") = layout_type;
#endif

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -1021,7 +1021,11 @@ struct TeamThreadMDRange<Rank<N, OuterDir, InnerDir>, TeamHandle> {
   using BoundaryType   = int;
   using TeamHandleType = TeamHandle;
   using ExecutionSpace = typename TeamHandleType::execution_space;
-  using ArrayLayout    = typename ExecutionSpace::array_layout;
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  using ArrayLayout KOKKOS_DEPRECATED_WITH_COMMENT("Use LayoutType instead.") =
+      typename ExecutionSpace::array_layout;
+#endif
+  using LayoutType = typename ExecutionSpace::layout_type;
```
science-enthusiast: using LayoutType = typename ExecutionSpace::layout_type;
  using ArrayLayout KOKKOS_DEPRECATED_WITH_COMMENT("Use LayoutType instead.") =
      LayoutType;
#endif

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -1057,7 +1061,11 @@ struct ThreadVectorMDRange<Rank<N, OuterDir, InnerDir>, TeamHandle> {
...
```
science-enthusiast: using LayoutType = typename ExecutionSpace::layout_type;
  using ArrayLayout KOKKOS_DEPRECATED_WITH_COMMENT("Use LayoutType instead.") =
      LayoutType;
#endif

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -1094,7 +1102,11 @@ struct TeamVectorMDRange<Rank<N, OuterDir, InnerDir>, TeamHandle> {
...
```
science-enthusiast: using LayoutType = typename ExecutionSpace::layout_type;
using ArrayLayout KOKKOS_DEPRECATED_WITH_COMMENT("Use LayoutType instead.") =
LayoutType;
#endif

core/src/Kokkos_Layout.hpp
```diff
@@ -30,7 +30,13 @@ static_assert(false,
...
 namespace Kokkos {
 
-enum { ARRAY_LAYOUT_MAX_RANK = 8 };
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+enum {
+  ARRAY_LAYOUT_MAX_RANK KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use LAYOUT_TYPE_MAX_RANK instead.") = 8
+};
+#endif
+enum { LAYOUT_TYPE_MAX_RANK = 8 };
```
science-enthusiast: enum { LAYOUT_TYPE_MAX_RANK = 8 };
enum {
  ARRAY_LAYOUT_MAX_RANK KOKKOS_DEPRECATED_WITH_COMMENT(
      "Use LAYOUT_TYPE_MAX_RANK instead.") = LAYOUT_TYPE_MAX_RANK
};
#endif

core/src/Kokkos_Layout.hpp
```diff
@@ -48,10 +54,15 @@ enum { ARRAY_LAYOUT_MAX_RANK = 8 };
...
 /// major."
 struct LayoutLeft {
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
   //! Tag this class as a kokkos array layout
-  using array_layout = LayoutLeft;
+  using array_layout KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use layout_type instead.") = LayoutLeft;
+#endif
+  //! Tag this class as a kokkos layout
+  using layout_type = LayoutLeft;
```
science-enthusiast, line 45: using layout_type = LayoutLeft;
  using array_layout KOKKOS_DEPRECATED_WITH_COMMENT(
      "Use layout_type instead.") = layout_type;
#endif

core/src/Kokkos_Layout.hpp
```diff
@@ -102,10 +113,15 @@ struct LayoutLeft {
...
 /// two-dimensional array, "layout right" is also called "row major."
 struct LayoutRight {
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
   //! Tag this class as a kokkos array layout
-  using array_layout = LayoutRight;
+  using array_layout KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use layout_type instead.") = LayoutRight;
+#endif
+  //! Tag this class as a kokkos layout
+  using layout_type = LayoutRight;
```
science-enthusiast: using layout_type = LayoutRight;
  using array_layout KOKKOS_DEPRECATED_WITH_COMMENT(
      "Use layout_type instead.") = layout_type;
#endif

core/src/Kokkos_Layout.hpp
```diff
@@ -146,11 +162,16 @@ struct LayoutRight {
...
 ///         multi-index mapping into contiguous memory.
 struct LayoutStride {
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
   //! Tag this class as a kokkos array layout
-  using array_layout = LayoutStride;
+  using array_layout KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use layout_type instead.") = LayoutStride;
+#endif
+  //! Tag this class as a kokkos layout
+  using layout_type = LayoutStride;
```
science-enthusiast, line 151: using layout_type = LayoutStride;
  using array_layout KOKKOS_DEPRECATED_WITH_COMMENT(
      "Use layout_type instead.") = LayoutStride;
#endif

core/src/Kokkos_ScratchSpace.hpp
```diff
@@ -65,8 +65,13 @@ class ScratchMemorySpace {
   //! This execution space preferred device_type
   using device_type = Kokkos::Device<execution_space, memory_space>;
 
-  using array_layout = typename ExecSpace::array_layout;
-  using size_type    = typename ExecSpace::size_type;
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  using array_layout KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use layout_type instead.") = typename ExecSpace::array_layout;
+#endif
+  using layout_type = typename ExecSpace::layout_type;
```
science-enthusiast: using layout_type = typename ExecSpace::layout_type;
  using array_layout KOKKOS_DEPRECATED_WITH_COMMENT(
      "Use layout_type instead.") = layout_type;
#endif

core/src/OpenACC/Kokkos_OpenACC.hpp
```diff
@@ -66,9 +66,12 @@ class OpenACC {
...
   using memory_space    = OpenACCSpace;
   using device_type     = Kokkos::Device<execution_space, memory_space>;
-
-  using array_layout = LayoutLeft;
-  using size_type    = memory_space::size_type;
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  using array_layout KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use layout_type instead.") = LayoutLeft;
+#endif
+  using layout_type = LayoutLeft;
```
science-enthusiast: using layout_type = LayoutLeft;
  using array_layout KOKKOS_DEPRECATED_WITH_COMMENT(
      "Use layout_type instead.") = layout_type;
#endif

core/src/OpenMP/Kokkos_OpenMP.hpp
```diff
@@ -59,8 +59,13 @@ class OpenMP {
...
 
   //! This execution space preferred device_type
-  using device_type          = Kokkos::Device<execution_space, memory_space>;
-  using array_layout         = LayoutRight;
+  using device_type = Kokkos::Device<execution_space, memory_space>;
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  using array_layout KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use layout_type instead.") = LayoutRight;
+#endif
+  using layout_type = LayoutRight;
```
science-enthusiast: using layout_type = LayoutRight;
  using array_layout KOKKOS_DEPRECATED_WITH_COMMENT(
      "Use layout_type instead.") = layout_type;
#endif

core/src/OpenMPTarget/Kokkos_OpenMPTarget.hpp
```diff
@@ -58,9 +58,12 @@ class OpenMPTarget {
...
   //! This execution space preferred device_type
   using device_type = Kokkos::Device<execution_space, memory_space>;
-
-  using array_layout = LayoutLeft;
-  using size_type    = memory_space::size_type;
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  using array_layout KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use layout_type instead.") = LayoutLeft;
+#endif
+  using layout_type = LayoutLeft;
```
science-enthusiast: using layout_type = LayoutLeft;
  using array_layout KOKKOS_DEPRECATED_WITH_COMMENT(
      "Use layout_type instead.") = layout_type;
#endif

core/src/SYCL/Kokkos_SYCL.hpp
```diff
@@ -56,8 +56,13 @@ class SYCL {
   using memory_space    = SYCLDeviceUSMSpace;
   using device_type     = Kokkos::Device<execution_space, memory_space>;
 
-  using array_layout = LayoutLeft;
-  using size_type    = memory_space::size_type;
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  using array_layout KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use layout_type instead.") = LayoutLeft;
+#endif
+  using layout_type = LayoutLeft;
```
science-enthusiast: using layout_type = LayoutLeft;
  using array_layout KOKKOS_DEPRECATED_WITH_COMMENT(
      "Use layout_type instead.") = layout_type;
#endif

core/src/Serial/Kokkos_Serial.hpp
```diff
@@ -106,7 +106,11 @@ class Serial {
   using device_type = Kokkos::Device<execution_space, memory_space>;
 
   //! This device's preferred array layout.
-  using array_layout = LayoutRight;
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  using array_layout KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use layout_type instead.") = LayoutRight;
+#endif
+  using layout_type = LayoutRight;
```
science-enthusiast: using layout_type = LayoutRight;
  using array_layout KOKKOS_DEPRECATED_WITH_COMMENT(
      "Use layout_type instead.") = layout_type;
#endif

core/src/View/Kokkos_ViewTraits.hpp
```diff
@@ -368,10 +391,14 @@ struct ViewTraits<
...
   using HostMirrorSpace = void;
-  using array_layout    = void;
-  using memory_traits   = MemoryTraits;
-  using specialize      = void;
-  using hooks_policy    = void;
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  using array_layout KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use layout_type instead.") = void;
+#endif
+  using layout_type   = void;
```
science-enthusiast: using layout_type   = void;
  using array_layout KOKKOS_DEPRECATED_WITH_COMMENT(
      "Use layout_type instead.") = layout_type;
#endif

core/src/View/Kokkos_ViewTraits.hpp
```diff
@@ -441,8 +472,12 @@ struct ViewTraits {
   //------------------------------------
   // Mapping traits:
 
-  using array_layout = ArrayLayout;
-  using dimension    = typename data_analysis::dimension;
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  using array_layout KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use layout_type instead.") = LayoutType;
+#endif
+  using layout_type = LayoutType;
```
science-enthusiast: using layout_type = LayoutType;
  using array_layout KOKKOS_DEPRECATED_WITH_COMMENT(
      "Use layout_type instead.") = layout_type;
#endif

core/src/View/Kokkos_ViewUniformType.hpp
```diff
@@ -36,17 +36,29 @@ struct ViewScalarToDataType<ScalarType, 0> {
...
 };
 
 template <class LayoutType>
 struct ViewUniformLayout<LayoutType, 0> {
-  using array_layout = Kokkos::LayoutLeft;
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  using array_layout KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use layout_type instead.") = Kokkos::LayoutLeft;
+#endif
+  using layout_type = Kokkos::LayoutLeft;
```
science-enthusiast: using layout_type = Kokkos::LayoutLeft;
  using array_layout KOKKOS_DEPRECATED_WITH_COMMENT(
      "Use layout_type instead.") = layout_type;
#endif

core/src/View/Kokkos_ViewUniformType.hpp
```diff
@@ -36,17 +36,29 @@ struct ViewScalarToDataType<ScalarType, 0> {
...
 };
 
 template <>
 struct ViewUniformLayout<Kokkos::LayoutRight, 1> {
-  using array_layout = Kokkos::LayoutLeft;
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  using array_layout KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use layout_type instead.") = Kokkos::LayoutRight;
+#endif
+  using layout_type = Kokkos::LayoutRight;
```
science-enthusiast: using layout_type = Kokkos::LayoutRight;
  using array_layout KOKKOS_DEPRECATED_WITH_COMMENT(
      "Use layout_type instead.") = layout_type;
#endif

core/src/Threads/Kokkos_Threads.hpp
```diff
@@ -52,8 +52,12 @@ class Threads {
   //! This execution space preferred device_type
   using device_type = Kokkos::Device<execution_space, memory_space>;
 
-  using array_layout = Kokkos::LayoutRight;
-  using size_type    = memory_space::size_type;
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  using array_layout KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use layout_type instead.") = LayoutRight;
+#endif
+  using layout_type = LayoutRight;
```
science-enthusiast: using layout_type = LayoutRight;
  using array_layout KOKKOS_DEPRECATED_WITH_COMMENT(
      "Use layout_type instead.") = layout_type;
#endif

containers/src/Kokkos_StaticCrsGraph.hpp
```diff
@@ -279,25 +279,29 @@ class StaticCrsGraph {
...
 
  public:
-  using data_type       = DataType;
-  using array_layout    = typename traits::array_layout;
+  using data_type = DataType;
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  using array_layout KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use layout_type instead.") = typename traits::array_layout;
+#endif
+  using layout_type     = typename traits::layout_type;
```
science-enthusiast: using layout_type     = typename traits::layout_type;
  using array_layout KOKKOS_DEPRECATED_WITH_COMMENT(
      "Use layout_type instead.") = layout_type;
#endif

containers/src/Kokkos_OffsetView.hpp
```diff
@@ -1270,14 +1269,21 @@ struct MirrorOffsetViewType {
...
   };
+
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
   // The array_layout
-  using array_layout = typename src_view_type::array_layout;
+  using array_layout KOKKOS_DEPRECATED_WITH_COMMENT(
+      "Use layout_type instead.") = typename src_view_type::array_layout;
+#endif
+  // The layout type
+  using layout_type = typename src_view_type::layout_type;
```
science-enthusiast, line 1279: using layout_type = typename src_view_type::layout_type;
  using array_layout KOKKOS_DEPRECATED_WITH_COMMENT(
      "Use layout_type instead.") = layout_type;
#endif

science-enthusiast (commented):
Please have a look at changes to be done.

Also, in `algorithms/src/std_algorithms/impl/Kokkos_RandomAccessIterator.hpp`
`ViewType::traits::array_layout` in many places should be replaced by `ViewType::traits::layout_type`, I think.

core/src/View/MDSpan/Kokkos_MDSpan_Layout.hpp
janciesko: Done

containers/src/Kokkos_DynamicView.hpp
```diff
@@ -287,9 +291,9 @@ class DynamicView : public Kokkos::ViewTraits<DataType, P...> {
   using uniform_device =
       Kokkos::Device<typename traits::device_type::execution_space,
                      Kokkos::AnonymousSpace>;
-  using uniform_type               = array_type;
+  using uniform_type               = type;
```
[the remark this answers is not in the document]
janciesko: We certainly could.

janciesko:
Rebasing

containers/src/Kokkos_DynamicView.hpp
```diff
@@ -267,10 +267,18 @@ class DynamicView : public Kokkos::ViewTraits<DataType, P...> {
...
+  /** \brief  Compatible view of data type */
+  using type KOKKOS_DEPRECATED_WITH_COMMENT("Use uniform_type instead.") =
+      uniform_type;
+#endif
 
   /** \brief  Compatible view of array of scalar types */
-  using array_type =
-      DynamicView<typename traits::data_type, typename traits::device_type>;
+  using array_type KOKKOS_DEPRECATED_WITH_COMMENT("Use uniform_type instead.") =
+      uniform_type;
```
masterleinad, line 272: If it's depreacted, it shouldn't exist when `Kokkos_ENABLE_DEPRECATED_COE_4=OFF` is set.

masterleinad (commented):
```
using array_type KOKKOS_DEPRECATED_WITH_COMMENT("Use uniform_type instead.") =
      uniform_type;
```
isn't properly guarded.

containers/src/Kokkos_DynamicView.hpp
science-enthusiast: If possible, it is better to make the changes related to `array_type` (including `uniform_type` and `type`) as part of [7360](https://github.com/kokkos/kokkos/pull/7360).

core/unit_test/TestViewCtorPropEmbeddedDim.hpp
```diff
@@ -75,10 +75,13 @@ struct TestViewCtorProp_EmbeddedDim {
         Kokkos::deep_copy(hcv1, cv1);
 
         ASSERT_EQ((std::is_same_v<CommonViewValueType, double>), true);
+
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
         ASSERT_EQ((std::is_same_v<
                       typename decltype(view_alloc_arg)::scalar_array_type,
                       CommonViewValueType>),
                   true);
+#endif
```
science-enthusiast: It is better to make this change as part of https://github.com/kokkos/kokkos/pull/7360.

containers/unit_tests/TestDynRankViewTypedefs.cpp
```diff
@@ -72,8 +72,10 @@ constexpr bool test_view_typedefs_impl() {
   static_assert(std::is_same_v<typename ViewType::const_value_type, const ValueType>);
   static_assert(std::is_same_v<typename ViewType::non_const_value_type, std::remove_const_t<ValueType>>);
 
+  #ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
   // FIXME: should maybe be deprecated
   static_assert(std::is_same_v<typename ViewType::array_layout, Layout>);
+  #endif
```
science-enthusiast:
```
#ifdef KOKKOS_ENABLE_DEPRECATION_WARNINGS
  KOKKOS_IMPL_DISABLE_DEPRECATED_WARNINGS_PUSH()
#endif
  static_assert(std::is_same_v<typename ViewType::array_layout, Layout>);
#ifdef KOKKOS_ENABLE_DEPRECATION_WARNINGS
  KOKKOS_IMPL_DISABLE_DEPRECATED_WARNINGS_POP()
#endif
#endif
```

containers/src/Kokkos_DynRankView.hpp
```diff
@@ -1579,13 +1582,20 @@ struct MirrorDRViewType {
     is_same_memspace =
         std::is_same_v<memory_space, typename src_view_type::memory_space>
   };
+
+  // The layout type
+  using layout_type = typename src_view_type::layout_type;
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
   // The array_layout
```
science-enthusiast, line 1573: We could remove this comment also.

containers/src/Kokkos_DynamicView.hpp
```diff
@@ -267,10 +267,18 @@ class DynamicView : public Kokkos::ViewTraits<DataType, P...> {
 
  public:
   //----------------------------------------------------------------------
+  using uniform_type =
+      DynamicView<typename traits::data_type, typename traits::device_type>;
+
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  /** \brief  Compatible view of data type */
```
science-enthusiast: Doxygen related comments could also be removed for deprecated code.

containers/src/Kokkos_DynamicView.hpp
```diff
@@ -577,14 +584,20 @@ struct MirrorDynamicViewType {
     is_same_memspace =
         std::is_same_v<memory_space, typename src_view_type::memory_space>
   };
+  // The layout type
+  using layout_type = typename src_view_type::layout_type;
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
   // The array_layout
```
science-enthusiast, line 586: This comment could be removed.

containers/unit_tests/TestDynRankViewTypedefs.cpp
```diff
@@ -72,8 +72,10 @@ constexpr bool test_view_typedefs_impl() {
...
```
science-enthusiast, line 87: It looks like the following static assert could be added:

`static_assert(std::is_same_v<typename ViewType::layout_type, Layout>);`

core/unit_test/TestViewTypedefs.cpp
```diff
@@ -71,8 +71,11 @@ constexpr bool test_view_typedefs_impl() {
   static_assert(std::is_same_v<typename ViewType::const_value_type, const ValueType>);
   static_assert(std::is_same_v<typename ViewType::non_const_value_type, std::remove_const_t<ValueType>>);
 
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
   // FIXME: should maybe be deprecated
```
science-enthusiast, line 78:
```
#ifdef KOKKOS_ENABLE_DEPRECATION_WARNINGS
  KOKKOS_IMPL_DISABLE_DEPRECATED_WARNINGS_PUSH()
#endif
  static_assert(std::is_same_v<typename ViewType::array_layout, Layout>);
#ifdef KOKKOS_ENABLE_DEPRECATION_WARNINGS
  KOKKOS_IMPL_DISABLE_DEPRECATED_WARNINGS_POP()
#endif
#endif
```

core/src/View/Kokkos_ViewTraits.hpp
```diff
@@ -266,20 +271,18 @@ using accessor_from_view_traits_t =
...
+        typename LayoutFromLayoutType<typename Traits::layout_type>::type>
 struct MDSpanViewTraits {
   using index_type = std::size_t;
   using extents_type =
       typename ExtentsFromDataType<index_type,
                                    typename Traits::data_type>::type;
-  using mdspan_layout_type = Layout;
-  static_assert(
-      std::is_same_v<Layout, typename LayoutFromArrayLayout<
-                                 typename Traits::array_layout>::type>);
```
science-enthusiast, line 269: Do we want to retain the static assert?

core/src/View/Kokkos_ViewTraits.hpp
```diff
@@ -346,10 +353,13 @@ struct ViewTraits<void, void, Prop...> {
   using execution_space = typename ViewTraits<void, Prop...>::execution_space;
   using memory_space    = typename ViewTraits<void, Prop...>::memory_space;
   using HostMirrorSpace = typename ViewTraits<void, Prop...>::HostMirrorSpace;
-  using array_layout    = typename ViewTraits<void, Prop...>::array_layout;
-  using memory_traits   = typename ViewTraits<void, Prop...>::memory_traits;
-  using specialize      = typename ViewTraits<void, Prop...>::specialize;
-  using hooks_policy    = typename ViewTraits<void, Prop...>::hooks_policy;
+  using layout_type     = typename ViewTraits<void, Prop...>::layout_type;
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  using array_layout = layout_type;
```
science-enthusiast: `using array_layout KOKKOS_DEPRECATED_WITH_COMMENT("Use layout_type instead.") = layout_type;`

containers/src/Kokkos_DynamicView.hpp
janciesko: Yes

janciesko: This is an oversight from splitting the PRs. Rolling this change back. Sorry about this.

containers/src/Kokkos_DynamicView.hpp
janciesko: The comment is within the deprecated code block so we will remove it in the future

core/src/View/Kokkos_ViewTraits.hpp
janciesko: Yes

containers/unit_tests/TestDynRankViewTypedefs.cpp
janciesko: Yes

janciesko:
Rebasing

core/unit_test/TestViewCtorPropEmbeddedDim.hpp
ldh4: I agree with this comment. This deprecation block doesn't seem related to deprecating `ViewTraits::array_layout`.
Any reason this deprecation block was needed here, especially since this is a unit test file?

containers/src/Kokkos_DynRankView.hpp
janciesko: Yes

janciesko:
This should be ready as well

core/src/View/MDSpan/Kokkos_MDSpan_Layout.hpp
```diff
@@ -51,38 +51,38 @@ struct IsLayoutLeftPadded<Experimental::layout_left_padded<Pad>>
     : std::true_type {};
 
 template <class ArrayLayout>
-struct LayoutFromArrayLayout {
+struct LayoutFromLayoutType {
```
nmm0: So something here can be a little confusing to people reading the code -- we have two types of layouts -- the Kokkos layouts and the mdspan layouts. For example a `LayoutLeft` vs `Experimental::layout_left_padded`. It's not clear from the name of this metafunction (and similarly for other conversion utilities) that it's converting from a Kokkos layout to an mdspan layout or vice versa. I wonder if the best thing here would be to rename to the rather verbose `MDSpanLayoutFromLayoutType` and something similar for the others.

containers/src/Kokkos_DynamicView.hpp
masterleinad: Didn't you want to address these changes in a different pull request?

containers/unit_tests/TestDynRankViewTypedefs.cpp
masterleinad: This comment doesn't seem to be addressed. I think it would be OK to only test the new alias.

core/src/View/Kokkos_ViewDataAnalysis.hpp
```diff
@@ -386,7 +386,6 @@ struct ViewDataAnalysis {
   using non_const_type =
       typename ViewDataType<non_const_value_type, dimension>::type;
 
-  // Generate "flattened" multidimensional array specification type.
```
masterleinad, line 385: What happened here?

core/src/View/Kokkos_ViewLegacy.hpp
```diff
@@ -246,27 +246,26 @@ class View : public ViewTraits<DataType, Properties...> {
 
  public:
   //----------------------------------------
-  /** \brief  Compatible view of array of scalar types */
+  /** \brief  Compatible view of data type */
```
masterleinad, line 239: Why did you change the comment?

core/src/View/Kokkos_ViewTraits.hpp
```diff
@@ -206,7 +206,12 @@ constexpr void customize_view_arguments(
 
 #ifdef KOKKOS_ENABLE_IMPL_MDSPAN
 namespace Impl {
+
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
 struct UnsupportedKokkosArrayLayout;
+#endif
```
masterleinad: Why do you need to deprecate if it's in `Impl`?

core/src/View/Kokkos_ViewTraits.hpp
```diff
@@ -554,13 +586,11 @@ struct ViewTraits {
...
       !std::is_same_v<void,
                       decltype(customize_view_arguments(
-                          Impl::ViewArguments<value_type, array_layout,
+                          Impl::ViewArguments<value_type, layout_type,
                                               device_type, memory_traits>()))>;
 
   enum { is_hostspace = std::is_same_v<MemorySpace, HostSpace> };
   enum { is_managed = MemoryTraits::is_unmanaged == 0 };
-  enum { is_random_access = MemoryTraits::is_random_access == 1 };
-
```
masterleinad: What happened here?

core/unit_test/TestViewCtorPropEmbeddedDim.hpp
masterleinad: This comment still applies.

core/unit_test/TestViewTypedefs.cpp
masterleinad: The comment still applies but it also seems to be OK to not test the deprecated alias and just the new one.

core/src/View/Kokkos_ViewLegacy.hpp
janciesko: Artifact from splitting PRs. Rolling back.

core/src/View/Kokkos_ViewTraits.hpp
janciesko: I don't. Removed.

core/src/View/Kokkos_ViewTraits.hpp
janciesko: Artifact. Restored.

core/unit_test/TestViewCtorPropEmbeddedDim.hpp
janciesko: Yes

algorithms/src/std_algorithms/impl/Kokkos_RandomAccessIterator.hpp
science-enthusiast, line 1: The accesses to `array_layout` in lines 56, 58 and 60 have to be fixed.

This will help to fix errors in the CUDA-11.6-NVCC-DEBUG and HIP-ROCm-5.3 tests in Jenkins.

core/src/View/MDSpan/Kokkos_MDSpan_Layout.hpp
science-enthusiast: If we can think of a name that retains the target (say, MDSpan) and what is being converted (say, layout) that should be enough. The source is understood from the context. Will `MDSpanLayoutAdapter` work?

core/src/View/MDSpan/Kokkos_MDSpan_Layout.hpp
```diff
@@ -51,38 +51,38 @@ struct IsLayoutLeftPadded<Experimental::layout_left_padded<Pad>>
     : std::true_type {};
 
 template <class ArrayLayout>
```
science-enthusiast: The parameter name can be changed to `LayoutType`.

core/src/View/Kokkos_ViewTraits.hpp
```diff
@@ -266,20 +267,18 @@ using accessor_from_view_traits_t =
 template <
```
science-enthusiast, line 254: In the comment above, "View's ArrayLayout" can be changed to "View's LayoutType".

science-enthusiast:
In `core/src/View/Kokkos_ViewDataAnalysis.hpp` there is a struct template called `ViewDataAnalysis` with a  template parameter called `ArrayLayout` that is used as a tag in other places (e.g., `core/src/View/Kokkos_ViewDataAnalysis.hpp`, `core/unit_test/TestViewMapping_a.hpp`).

core/unit_test/TestViewTypedefs.cpp
janciesko: Yes

core/src/View/MDSpan/Kokkos_MDSpan_Layout.hpp
janciesko: Given we use "this-from-this" naming in the file (`LayoutFromLayoutType`), `MDSpanLayoutFromLayoutType` seem to be more consistent than using adapter.

janciesko: It's that's ok, I'll rename to `MDSpanLayoutFromLayoutType`.

janciesko:
This should be ready :)

science-enthusiast:
There is this `extents_type` missing error involving `Kokkos::LayoutStride` that is appearing in a few CI tests [linux / CI (x64, ubuntu:latest, g++-12, RelWithDebInfo, OPENMP, 17)](https://github.com/kokkos/kokkos/actions/runs/16759166296/job/47451006419?pr=7401#logs) and [CUDA-11.7-NVCC](https://cloud1.cees.ornl.gov/jenkins-ci/job/Kokkos/job/PR-7401/18/pipeline-overview/?selected-node=389). When I tested locally on my machine, similar two tests (`core/perf_test/PerfTest_ViewCopy_*.cpp` and `core/unit_test/*/Test*_MDSpanConversion.cpp`) are failing for all the backends that I tested for: CPU-Serial, CPU-Threads, CPU-OpenMP and CPU-Serial + GPU-CUDA. The nature of the error is the same - involving `Kokkos::LayoutStride` and the `is_mapping_of` check in `tpls/mdspan/include/experimental/__p0009_bits/layout_stride.hpp`. The `develop` branch builds successfully locally for all these backends. The C++ compiler is GCC 11.4.

janciesko:
It seems like overload resolution for `typename Mapping` fails for some reason in
```
template<class Layout, class Mapping>
  constexpr bool is_mapping_of =
    std::is_same<typename Layout::template mapping<typename Mapping::extents_type>, Mapping>::value;
```
where Mapping is a  `StridedLayoutMapping`.

Weird that the renaming is causing this?
Note: This seems to be a compiler specific issue. In my case, clang16 compiles fine, GCC>12 works, GCC12 fails.
@nmm0 - Do you see a reason why this might be happening?

science-enthusiast:
> Note: This seems to be a compiler specific issue. In my case, clang16 compiles fine, GCC12 fails.

I observe the same behavior as what @janciesko is observing. The backends CPU-Serial, CPU-Threads and CPU-OpenMP are building correctly with Clang-16.

nmm0:
> It seems like overload resolution for `typename Mapping` fails for some reason in
> 
> ```
> template<class Layout, class Mapping>
>   constexpr bool is_mapping_of =
>     std::is_same<typename Layout::template mapping<typename Mapping::extents_type>, Mapping>::value;
> ```
> 
> where Mapping is a `StridedLayoutMapping`.
> 
> Weird that the renaming is causing this? Note: This seems to be a compiler specific issue. In my case, clang16 compiles fine, GCC12 fails. @nmm0 - Do you see a reason why this might be happening?

@janciesko if you are talking about this error:

```
 /__w/kokkos/kokkos/tpls/mdspan/include/experimental/__p0009_bits/layout_stride.hpp:53:95: error: no type named 'extents_type' in 'struct Kokkos::LayoutStride'
   53 |     std::is_same<typename Layout::template mapping<typename Mapping::extents_type>, Mapping>::value;
      |                                                                                               ^~~~~
```

Looks like a Kokkos layout (LayoutStride) didn't get translated into the corresponding mdspan layout (layout_stride) somewhere

janciesko:
@nmm0, yes that error. It fails once we hit the conversion test in:
core/unit_test/TestMDSpanConversion.hpp:292
```
    // LayoutStride
    {
      const size_t strides[] = {2};
      test_conversion_from_mdspan<Kokkos::LayoutStride>(
          Kokkos::View<value_type *, Kokkos::LayoutStride, ExecutionSpace>(
              "ref", Kokkos::LayoutStride{7, 2}),
          Kokkos::layout_stride::mapping<Kokkos::dextents<std::size_t, 1>>{
              Kokkos::mdspan_non_standard, Kokkos::dextents<std::size_t, 1>{7},
              strides});
    }
```
For LL and LR is seems to work.

core/src/View/MDSpan/Kokkos_MDSpan_Layout.hpp
```diff
@@ -50,39 +50,39 @@ template <size_t Pad>
...
 template <>
-struct LayoutFromArrayLayout<LayoutRight> {
+struct MDSpanLayoutFromLayoutType<LayoutRight> {
   using type = Experimental::layout_right_padded<dynamic_extent>;
 };
 
 template <>
-struct LayoutFromArrayLayout<LayoutStride> {
+struct MDSpanLayoutFromLayoutType<LayoutStride> {
   using type = layout_stride;
```
science-enthusiast, line 57: Could we specify `layout_stride` with complete namespace scope and see whether it helps in addressing the compilation error with GCC?

janciesko: Yep, but same error occurs (specific to GCC12).

janciesko: Maybe compiler bug

science-enthusiast: Thanks for trying. Indeed, it appears to be a compiler bug. Trévis (@tretre91) tried on his machine, which has GCC 13 and it was successful. Could you say more on this, Trévis?

tretre91: It works with gcc 13.3 but gives an error with gcc 11.4. I tried to play with a small reproducer, and it didn't seem to give an error when building a `BasicView` with the same parameters, I tried to investigate further but I'm not that familiar with the View internals
```cpp
int main(int argc, char *argv[]) {
  Kokkos::ScopeGuard guard(argc, argv);

  using view_type = Kokkos::View<double *, Kokkos::LayoutStride,
                                 Kokkos::DefaultExecutionSpace>;

  // This works
  Kokkos::Impl::BasicViewFromTraits<double *, Kokkos::LayoutStride,
                                    Kokkos::DefaultExecutionSpace>::type
      v("foo",
        Kokkos::Impl::mapping_from_layout_type<
            view_type::mdspan_type::mapping_type>(Kokkos::LayoutStride{7, 2}));

  // This gives an error
  Kokkos::View<double *, Kokkos::LayoutStride, Kokkos::DefaultExecutionSpace>
    view("foo", Kokkos::LayoutStride{7, 2});
}
```

science-enthusiast: Nice to see that you tried the following construct:
```
  Kokkos::Impl::BasicViewFromTraits<double *, Kokkos::LayoutStride,
                                    Kokkos::DefaultExecutionSpace>::type
      v("foo",
        Kokkos::Impl::mapping_from_layout_type<
            view_type::mdspan_type::mapping_type>(Kokkos::LayoutStride{7, 2}));
```
The following also works:
```
  Kokkos::View<double *, Kokkos::LayoutStride, Kokkos::DefaultExecutionSpace>
    view("foo", Kokkos::Impl::mapping_from_layout_type<
                  view_type::mdspan_type::mapping_type>(Kokkos::LayoutStride{7, 2}));
```

janciesko:
The error message seems to make sense. Kokkos' LayoutStride does not have that member. I do not see why calling `mapping_from_layout_type` via the View ctor is different to calling it explicitly. Seems like `LayoutStride` is not converted to`layout_stride`.
[...]
```
In instantiation of 'constexpr const bool Kokkos::detail::is_mapping_of<Kokkos::LayoutStride, Kokkos::LayoutStride>':
/ascldap/users/jciesko/Kokkos/kokkos/tpls/mdspan/include/mdspan/../experimental/__p0009_bits/layout_stride.hpp:53:95: error: no type named 'extents_type' in 'struct Kokkos::LayoutStride'
   53 |     std::is_same<typename Layout::template mapping<typename Mapping::extents_type>, Mapping>::value;
```

science-enthusiast:
> The error message seems to make sense. Kokkos' LayoutStride does not have that member. I do not see why calling `mapping_from_layout_type` via the View ctor is different to calling it explicitly. Seems like `LayoutStride` is not converted to`extents_type`. [...]

I agree with you. The function `Kokkos::Impl::mapping_from_layout_type` correctly returns an MDSpan layout_stride which has the member `extents_type`. However, `Kokkos::LayoutStride` is not getting mapped to MDSpan layout_stride.

nmm0:
I suspect the bug is coming from compiler substitution as it is trying to find an appropriate constructor for `layout_stride`. I'm not 100% sure how it is reaching that, I'm still trying to figure that out to see if there is a workaround. It's worth noting that this works with GCC 12 if you build with C++20 as the condition on that constructor switches to a `requires` rather than `enable_if` -- which is why I'm pretty sure it's an issue with how the compiler is doing substitution. As for why it's happening now and not before, or where it is originating from, I'm not sure yet.

nmm0:
Ok so just updating with my findings -- gonna try and find a workaround. But basically, for some reason gcc 12 is considering this BasicView constructor:

```cpp
  template <class... P>
  explicit inline BasicView(
      const Impl::ViewCtorProp<P...> &arg_prop,
      std::enable_if_t<!Impl::ViewCtorProp<P...>::has_pointer &&
                           !Impl::ViewCtorProp<P...>::has_accessor_arg,
                       typename mdspan_type::mapping_type> const &arg_mapping)
      : BasicView(create_data_handle(arg_prop, arg_mapping, accessor_type{}),
                  arg_mapping) {}
```

I guess it's considering converting `LayoutStride` to the mdspan `layout_stride` -- that of course would normally be rejected as a candidate since it's not a valid conversion, but I guess GCC 12 chokes on it while evaluating whether it can convert.

nmm0:
I believe I found a workaround. It's not pretty, but will at least change GCC 12 from failing to compile to just (rightfully) failing substittion when considering that overload. If we change the definition of `is_mapping_of` to:


```cpp
  template<class, class Enabled = void>
  struct is_mapping_of_impl {
    template<class>
    static constexpr bool value_for_layout = false;
  };

  template<class Mapping>
  struct is_mapping_of_impl<Mapping, std::void_t<typename Mapping::extents_type>>
  {
    template<class Layout>
    static constexpr bool value_for_layout = std::is_same<typename Layout::template mapping<typename Mapping::extents_type>, Mapping>::value;
  };

  template<class Layout, class Mapping>
  constexpr bool is_mapping_of =
    is_mapping_of_impl<Mapping>::template value_for_layout<Layout>;
 ```

The tests compile for me with GCC12. Let's try that, but hold off on merging this until I can get the workaround in mdspan as we should merge and test it there first.

**Comment by janciesko** [comment]
Yes, this technically works. Not sure if we should restrict this to GCC12 and what the value for `value_for_layout` in the defaulted case should be.

File: core/src/View/MDSpan/Kokkos_MDSpan_Layout.hpp (hunk 55 above)
Comment (science-enthusiast) [reply core/src/View/MDSpan/Kokkos_MDSpan_Layout.hpp:57]: Addressed by @nmm0 in https://github.com/kokkos/mdspan/pull/423.

**Comment by science-enthusiast** [comment]
Awaiting [#8320](https://github.com/kokkos/kokkos/pull/8320) to be merged. It is needed to handle the mdspan related error for GCC 12, 11.4 etc.

**Comment by janciesko** [comment]
@science-enthusiast, this should be good now.

**Comment by science-enthusiast** [comment]
> @science-enthusiast, this should be good now.

Happy to see that all tests are passing. Most likely, all is well with this PR. We went over the various places to change quite carefully.

**Review by nmm0 — COMMENTED** [review]
Looks good to me though I did push some to this some time ago so I'm going to wait for other ppl to approve

File: containers/unit_tests/TestDynRankViewTypedefs.cpp [hunk 56]
```diff
@@ -94,20 +102,21 @@ constexpr bool test_view_typedefs_impl() {
 
   // FIXME: in contrast to View, hooks_policy is not propagated
   static_assert(std::is_same_v<typename ViewType::traits, ViewTraitsType>);
+
   static_assert(std::is_same_v<typename ViewType::array_type,
-                               Kokkos::DynRankView<typename ViewType::data_type, typename ViewType::array_layout,
+                               Kokkos::DynRankView<typename ViewType::scalar_array_type, typename ViewType::layout_type,
```
ldh4: Why was `ViewType::data_type` changed to `scalar_array_type` here? Is it related to the `array_layout` deprecation or is it just for consistency?

science-enthusiast: Good catch :thumbsup:

janciesko: Updated, thanks @ldh4

ldh4 (approved):

janciesko:
That is because in 5.0, "Deprecated code 4" is off by default right?

janciesko:
Changing to `KOKKOS_ENABLE_DEPRECATED_CODE_5`

janciesko:
Retest this please

core/src/Kokkos_Core.cppm
```diff
@@ -102,9 +102,9 @@ export {
   using ::Kokkos::InvalidType;
   using ::Kokkos::is_always_assignable;
   using ::Kokkos::is_always_assignable_v;
-  using ::Kokkos::is_array_layout;
-  using ::Kokkos::is_array_layout_v;
```
masterleinad, line 106: You still need those when `KOKKOS_ENABLE_DEPRECATED_CODE_5` is enabled.

janciesko: Yes

ndellingwood:
Some compatibility changes that'll be needed in Trilinos after this PR merges, possibly more needed (below needs guards for KOKKOS_ENABLE_DEPRECATED_CODE_5 or maybe by KOKKOS_VERSION >= 50099 or so):

Sacado, with new View impl enabled:
```diff
diff --git a/packages/sacado/src/Kokkos_LayoutContiguous.hpp b/packages/sacado/src/Kokkos_LayoutContiguous.hpp
index 316b7796749..8c466eb0812 100644
--- a/packages/sacado/src/Kokkos_LayoutContiguous.hpp
+++ b/packages/sacado/src/Kokkos_LayoutContiguous.hpp
@@ -50,8 +50,8 @@ struct LayoutContiguous : public Layout {
 #ifdef SACADO_HAS_NEW_KOKKOS_VIEW_IMPL
 namespace Impl {
 template <class Layout, unsigned Stride>
-struct LayoutFromArrayLayout<LayoutContiguous<Layout, Stride>> {
-  using type = typename LayoutFromArrayLayout<Layout>::type;
+struct MDSpanLayoutFromLayoutType<LayoutContiguous<Layout, Stride>> {
+  using type = typename MDSpanLayoutFromLayoutType<Layout>::type;
 };
 }
 #endif
diff --git a/packages/sacado/src/Sacado_Fad_Kokkos_Specialization.hpp b/packages/sacado/src/Sacado_Fad_Kokkos_Specialization.hpp
index eaf27171646..6727dcdf21c 100644
--- a/packages/sacado/src/Sacado_Fad_Kokkos_Specialization.hpp
+++ b/packages/sacado/src/Sacado_Fad_Kokkos_Specialization.hpp
@@ -185,7 +185,7 @@ subdynrankview(const DynRankView<T, LayoutContiguous<LayoutSrc, StrideSrc>,
   return return_type{
       typename return_type::view_type(
           sub.data_handle(),
-          Impl::mapping_from_array_layout<typename return_type::mapping_type>(
+          Impl::mapping_from_layout_type<typename return_type::mapping_type>(
               layout),
           sub.accessor()),
       new_rank};
```

Sacado with legacy View (maybe not important, since legacy view code is set for removal):
```diff
diff --git a/packages/sacado/src/KokkosExp_View_Fad.hpp b/packages/sacado/src/KokkosExp_View_Fad.hpp
index 104b85dbb0c..e04aec11d4e 100644
--- a/packages/sacado/src/KokkosExp_View_Fad.hpp
+++ b/packages/sacado/src/KokkosExp_View_Fad.hpp
@@ -2079,7 +2079,7 @@ public:
             && Sacado::IsStaticallySized< typename Sacado::ValueType< TraitsValueType >::type >::value
            )
         {
-          typedef typename DstType::offset_type::array_layout DstLayoutType;
+          typedef typename DstType::offset_type::layout_type DstLayoutType;
           //typedef typename ViewArrayLayoutSelector<typename DstType::offset_type::array_layout>::type DstLayoutType;
           typedef typename SrcFadType::array_offset_type::dimension_type SrcViewDimension;
```

core/src/Kokkos_Concepts.hpp
```diff
@@ -146,7 +146,10 @@ KOKKOS_IMPL_DEFINE_CONCEPT_AND_TRAIT_FROM_TYPEDEF(execution_space,
                                                   ExecutionSpace)
 KOKKOS_IMPL_DEFINE_CONCEPT_AND_TRAIT_FROM_TYPEDEF(execution_policy,
                                                   ExecutionPolicy)
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_5
 KOKKOS_IMPL_DEFINE_TRAIT_FROM_TYPEDEF(array_layout)
+#endif
+KOKKOS_IMPL_DEFINE_TRAIT_FROM_TYPEDEF(layout_type)
```
masterleinad:
```suggestion
KOKKOS_IMPL_DEFINE_CONCEPT_AND_TRAIT_FROM_TYPEDEF(layout_type, LayoutType)
```
and add tests for it.

JBludau:
It seems to have broken the module build in jenkins.
@janciesko can you post the command you used to make all these changes? it is quite hard to review

core/src/Kokkos_Concepts.hpp
```diff
@@ -146,7 +146,12 @@ KOKKOS_IMPL_DEFINE_CONCEPT_AND_TRAIT_FROM_TYPEDEF(execution_space,
                                                   ExecutionSpace)
 KOKKOS_IMPL_DEFINE_CONCEPT_AND_TRAIT_FROM_TYPEDEF(execution_policy,
                                                   ExecutionPolicy)
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_5
+KOKKOS_IMPL_DISABLE_DEPRECATED_WARNINGS_PUSH()
```
science-enthusiast, line 150: Maybe, this point was discussed earlier. I don't remember. Sorry, to bring it up at a late stage. I would like to know if
```
KOKKOS_IMPL_DISABLE_DEPRECATED_WARNINGS_PUSH()
```
must always be
```
#ifdef KOKKOS_ENABLE_DEPRECATION_WARNINGS
  KOKKOS_IMPL_DISABLE_DEPRECATED_WARNINGS_PUSH()
#endif
```

The same for `KOKKOS_IMPL_DISABLE_DEPRECATED_WARNINGS_POP()`.

masterleinad: Only if you still need to use guarded deprecated code (tests or any other code that is deprecated).

science-enthusiast: Thanks for the clarification :+1:

masterleinad:
@janciesko Looks like you need to fix merge conflicts.

janciesko:
Rebased. The diff came from merging https://github.com/kokkos/kokkos/pull/8614 which reintroduced the deprecated array_layout as a side-effect.

science-enthusiast:
Retest this please

science-enthusiast:
I would like to know why this PR is not yet merged? It carries changes in many places. It is better to merge it without too much delay, please.

I understand that #8614 has taken care of compatibility with Trilinos and @janciesko has rebased w.r.t. those changes.

The two Jenkins failures are linking issues:
* `CUDA-12.2-NVHPC-AS-HOST-COMPILER`: Could not open input file '/usr/lib/x86_64-linux-gnu/libcuda.so'
* `CUDA-11.8-Clang-15`
   /usr/bin/ld: /usr/bin/ld: DWARF error: invalid or unhandled FORM value: 0x23
   ../src/libkokkoscore.a(Kokkos_Cuda_Instance.cpp.o): in function `Kokkos::Impl::CudaInternal::initialize(CUstream_st*)':
   Kokkos_Cuda_Instance.cpp:(.text+0x113c): undefined reference to `cuStreamGetCtx'
   ...

masterleinad:
@janciesko Can you rebase and resolve merge conflicts?

science-enthusiast:
@janciesko has rebased till 0f5c86cc3e811ea23fcd7edd04d5d240fcbbacde. Is that enough?

masterleinad:
No, there are still conflicts in `core/src/Kokkos_View.hpp` that must be resolved.

janciesko:
Rebased

janciesko:
Retest this please

janciesko:
Sorry all, we're a bit stalled on this one.
The issue is the following.
```C++

    using view_t = Kokkos::View<int**>;
    view_t v{};
    auto mds = v.to_mdspan();
    static_assert(!std::is_same_v<typename decltype(mds)::layout_type, decltype(v.layout())>);
    static_assert(!std::is_same_v<typename decltype(mds)::layout_type, typename decltype(v)::array_layout /*layout_type in this PR*/>);
    
    using mdspan_t = Kokkos::mdspan<double, Kokkos::dextents<int, 2>, typename view_t::layout_type>; 
    using mdspan2_t = Kokkos::mdspan<double, Kokkos::dextents<int, 2>, typename view_t::array_layout /*layout_type in this PR*/>; 
    mdspan_t b1; //OK
    mdspan2_t b2; //fails

```

Renaming `View::array_layout` to `View::layout_type` collides with `BasicView`'s `layout_type` which we derive View from.
Example above:
```
// app/kokkos/include/experimental/__p0009_bits/mdspan.hpp:77:9: error: no class template named 'mapping' in 'using Kokkos::mdspan<double, Kokkos::extents<int, 18446744073709551615, 18446744073709551615>, Kokkos::LayoutRight>::layout_type = struct Kokkos::LayoutRight' {aka 'struct Kokkos::LayoutRight'}
// using mapping_type = typename layout_type::template mapping<extents_type>;
```

janciesko:
Closing for now.

Issue #7090, opened 2024-06-21
Kokkos::Threads backend is broken on M2 MacBook (ARM) in develop


crtrott:
It still works in 4.3.01 but in develop a lot of the tests segfault when launched with more than one thread (export KOKKOS_NUM_THREADS=4). 

For example:
```
s1087574:build crtrott$ containers/unit_tests/Kokkos_ContainersUnitTest_Threads 
[==========] Running 65 tests from 1 test suite.
[----------] Global test environment set-up.
[----------] 65 tests from threads
[ RUN      ] threads.bitset
Segmentation fault: 11
```

Configured:

```
cmake -DKokkos_ENABLE_THREADS=ON -DKokkos_ENABLE_TESTS=ON
```

dalg24:
Were you able to reproduce on another platform or is it Mac specific?

crtrott:
My gut feeling is that it's ARM specific, maybe Power (i.e. relaxed memory semantics), haven't explicitly tried though.

dalg24:
I was able to reproduce on M1 Pro.  It looks indeterministic.
Note that I did get a segfault on 4.3.01 after re-running a couple times.
(Just a data point, I am not investigating further for the time being)

janciesko:
For the mentioned configuration and architecture, a variety of tests is failing if running a few times, incl unit tests in core. Are we not catching this in CI?

masterleinad:
> For the mentioned configuration and architecture, a variety of tests is failing if running a few times, incl unit tests in core. Are we not catching this in CI?

What are you seeing? For reference, we are using `hwloc` in the CI with the Threads backend (https://github.com/kokkos/kokkos/blob/develop/.github/workflows/continuous-integration-workflow.yml so we are initializing with multiple threads.

masterleinad:
```
17: [ RUN      ] THREADS.incr_14_MDrangeReduce
17: /Users/6da/Software/kokkos/core/unit_test/incremental/Test14_MDRangeReduce.hpp:115: Failure
17: Expected equality of these values:
17:   h_result
17:     Which is: 1012.5
17:   d_result
17:     Which is: 720
```
Is what I'm seeing locally on my Mac M1 with #7402.

dalg24:
Anyone actually checked that this does not occur say on X86?

masterleinad:
> Anyone actually checked that this does not occur say on X86?

That's what we have in our CI if I'm not mistaken.

dalg24:
[...]

PR #7402, opened 2024-10-04
Use std::thread::hardware_concurrency as default for the Threads backend

masterleinad:
Related to https://github.com/kokkos/kokkos/issues/7090. It's strane that the `Threads` backend only uses one thread by default whereas we autodetect the number of hardware concurrent threads for other host parallel backends.

### DIFF
--- a/core/src/Threads/Kokkos_Threads_Instance.cpp
+++ b/core/src/Threads/Kokkos_Threads_Instance.cpp
@@ -493,7 +493,7 @@ void ThreadsInternal::initialize(int thread_count_arg) {
                          ? Kokkos::hwloc::get_available_numa_count() *
                                Kokkos::hwloc::get_available_cores_per_numa() *
                                Kokkos::hwloc::get_available_threads_per_core()
-                         : 1;
+                         : std::thread::hardware_concurrency();
     }
 
     const bool allow_asynchronous_threadpool = false;


crtrott:
std::thread::hardware_conccurency is bad I think. I thought we investigated this before. I.e. it did not seem to take into account stuff like process masks?

masterleinad:
> std::thread::hardware_conccurency is bad I think. I thought we investigated this before. I.e. it did not seem to take into account stuff like process masks?

That's what we are using for `OpenACC` as concurrency (if targeting host) and for `cores per node` when detecting MPI over-subscription with `Threads` and `OpenMP`. It's not necessarily a very good value but better than just 1. When configuring with hwloc, we take more information into account.

crtrott:
Ok I guess I remember wrong.

janciesko:
I don't have a critical dependency on the default behavior w/o hwloc. The change looks good to me and seems consistent with other backends.

janciesko (approved):

dalg24 (changes requested):
This has been discussed in the past and we had agreed that the status quo was better than wildly oversubscribing nodes with the default setting.

We should consider revisiting/cleaning up our hwloc wrappers and potentially bundling it by default with Kokkos.

masterleinad:
> This has been discussed in the past and we had agreed that the status quo was better than wildly oversubscribing nodes with the default setting.

We do have warnings for oversubscription with MPI and they use `std::thread::hardware_concurrency` as well, see https://github.com/kokkos/kokkos/blob/4e88587265240078b2d6d94261d118e7a4a15cea/core/src/Threads/Kokkos_Threads_Instance.cpp#L600-L615.

dalg24:
> > This has been discussed in the past and we had agreed that the status quo was better than wildly oversubscribing nodes with the default setting.
> 
> We do have warnings for oversubscription with MPI and they use `std::thread::hardware_concurrency` as well, see

Sure.  We had it back then too.
To be clear I am not saying this can never change, just that there we are relitigating something we looked at in the past and that there is no new information being brought up so far.

janciesko:
Given that `std::thread::hardware_concurrency` hints on how many threads can be running in parallel at any given moment in time (vs. how many can be created), I wonder if a heuristic that adjusts that default value relative to reported ranks might be better?

masterleinad:
> Given that `std::thread::hardware_concurrency` hints on how many threads can be running in parallel at any given moment in time (vs. how many can be created), I wonder if a heuristic that adjusts that default value relative to reported ranks might be better?

Feel free to suggest something better.

PR #7406, opened 2024-10-04
Workaround for AVX2 evaluation with CrayClang Compiler

rgayatri23:
Co-authored-by: Dong Hun Lee @ldh4 

The PR tries to work around the issue with CrayClang compiler where we need to do a bitwise comparison to evaluate the presence of `__AVX2__` macro. 

Not reproducible with simple test case.


simd/src/Kokkos_SIMD.hpp
```diff
@@ -48,8 +48,12 @@
 #error "__AVX__ must be defined for KOKKOS_ARCH_AVX"
 #endif
 
+// FIXME_CRAY_LLVM - For some reason CrayClang needs bitwise comparison here to
+// correctly evaluate the AVX2 flag.
 #if defined(KOKKOS_ARCH_AVX2)
-#if !defined(__AVX2__)
+#if !defined(__AVX2__) && !defined(KOKKOS_COMPILER_CRAY_LLVM)
+#error "__AVX2__ must be defined for KOKKOS_ARCH_AVX2"
+#elif defined(KOKKOS_COMPILER_CRAY_LLVM) && (__AVX2__ & 0x0)
```
masterleinad, line 56: Doesn't `(__AVX2__ & 0x0)` always evaluate to 0?

rgayatri23: You are right. Thinking of a different way to address this issue.

rgayatri23: Avoiding the check for the CRAY_LLVM compiler for now.

simd/src/Kokkos_SIMD.hpp
```diff
@@ -48,8 +48,13 @@
 #error "__AVX__ must be defined for KOKKOS_ARCH_AVX"
 #endif
 
+// FIXME_CRAY_LLVM - For some reason CrayClang does not correctly evaluate the
+// `__AVX2__` macro at this location even when the compiler supports it on the
+// given architecture. Avoiding this check for the (architecture, compiler)
+// combination for now. The code would fail if the macro is unavailable although
+// the error message might not be clean in the case.
```
masterleinad, line 55: What backend is this for? Note that we are disabling all these checks for `OpenMPTarget` already.

rgayatri23: This is for enabling SIMD unit tests.

masterleinad: Sure. Since you get here, you must try with a different backend than `OpenMPTarget`, right?

ldh4: I think this was reproduced even with just Serial backend.
I now wonder if this check against the cray compiler should be moved to L34 as well.

masterleinad: It would make sense to me that the issue is not specific to `AVX2` but I'm surprised that AVX2 intrinsic are usable although the macro isn't defined.

ldh4: I don't know how to logically describe this behavior, but the macro was defined.
We checked that `__AVX2__` was in the list of preprocessor defined flags by dumping the output from CC.
Then we tested `#ifdef __AVX2__` right above this block, and it was correctly recognizing the macro to be defined; however, as soon as that added `ifdef` block was commented back out, the error on L58 was being outputted.

I had suspected that it could be something specific with the environment, but it sounded like @rgayatri23 was able to reproduce the same error on another machine with cray compiler.

rgayatri23: Ok an update - I was not able to reproduce it on a different machine, i.e., frontier which has the same CPU arch and same compiler/version. So I think it might be something on Perlmutter that is probably interfering with the `AVX2` macro evaluation.

masterleinad:
@rgayatri23 What's the status here?

rgayatri23:
I am going to close this for now since I observe this behavior only on Perlmutter.

Issue #7399, opened 2024-10-03
Better error message for HIP parallel_reduce


jewatkins:
If the reducer size is too large inside a HIP parallel_reduce, Kokkos errors out with the following message: "Kokkos::Impl::ParallelReduce< HIP > could not find a valid execution configuration." It wasn't clear that too much shared memory was being requested.

I had to print values here to get an idea of what was happening (and it looks like maybe some better error handling was envisioned): https://github.com/kokkos/kokkos/blob/develop/core/src/HIP/Kokkos_HIP_BlockSize_Deduction.hpp#L158

Also, the error message was much more clear when I simply removed the original error: "HIPParallelLaunch FAILED: shared memory request is too large". Would it be possible to return a bit more information here? At the very least a message about too much shared memory but a kernel name, requested sizes, etc would be very useful.

dalg24:
Not sure how much we can promise here but agreeing that we should looking into it.

cwpearson:
#7414 changed the message so it says how much is requested / how much can be provided.

PR #7407, opened 2024-10-04
Cuda: Error out for non-positive team_size_max/recommended

masterleinad:
Related to #7399. While working on #5879, I run into cases where the kernel just wasn't launched.

### DIFF
--- a/core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
@@ -106,7 +106,12 @@ class TeamPolicyInternal<Kokkos::Cuda, Properties...>
             (size_t)impl_vector_length(),
             (size_t)team_scratch_size(0) + 2 * sizeof(double),
             (size_t)thread_scratch_size(0) + sizeof(double));
-    return block_size / impl_vector_length();
+    int max_block_size = block_size / impl_vector_length();
+    if (max_block_size <= 0)
+      Kokkos::Impl::throw_runtime_exception(std::string(
+          "Kokkos::Impl::ParallelFor<Cuda, TeamPolicy> could not find a "
+          "valid execution configuration."));
+    return max_block_size;
   }
 
   template <class FunctorType>
@@ -119,7 +124,14 @@ class TeamPolicyInternal<Kokkos::Cuda, Properties...>
         CombinedFunctorReducer<FunctorType,
                                typename functor_analysis_type::Reducer>,
         TeamPolicy<Properties...>, Kokkos::Cuda>;
-    return internal_team_size_max<closure_type>(f);
+
+    int max_block_size = internal_team_size_max<closure_type>(f);
+
+    if (max_block_size <= 0)
+      Kokkos::Impl::throw_runtime_exception(std::string(
+          "Kokkos::Impl::ParallelReduce<Cuda, TeamPolicy> could not find a "
+          "valid execution configuration."));
+    return max_block_size;
   }
 
   template <class FunctorType, class ReducerType>
@@ -128,7 +140,14 @@ class TeamPolicyInternal<Kokkos::Cuda, Properties...>
     using closure_type =
         Impl::ParallelReduce<CombinedFunctorReducer<FunctorType, ReducerType>,
                              TeamPolicy<Properties...>, Kokkos::Cuda>;
-    return internal_team_size_max<closure_type>(f);
+
+    int max_block_size = internal_team_size_max<closure_type>(f);
+
+    if (max_block_size <= 0)
+      Kokkos::Impl::throw_runtime_exception(std::string(
+          "Kokkos::Impl::ParallelReduce<Cuda, TeamPolicy> could not find a "
+          "valid execution configuration."));
+    return max_block_size;
   }
 
   template <class FunctorType>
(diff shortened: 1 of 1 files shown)


core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
```diff
@@ -106,7 +106,12 @@ class TeamPolicyInternal<Kokkos::Cuda, Properties...>
             (size_t)impl_vector_length(),
             (size_t)team_scratch_size(0) + 2 * sizeof(double),
             (size_t)thread_scratch_size(0) + sizeof(double));
-    return block_size / impl_vector_length();
+    int max_block_size = block_size / impl_vector_length();
+    if (max_block_size <= 0)
+      Kokkos::Impl::throw_runtime_exception(std::string(
```
crtrott, line 111: Isn't Damien trying to get rid of all exceptions?

crtrott (changes requested):

core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
masterleinad: https://github.com/kokkos/kokkos/pull/7386/files#diff-96b736d454cbbe8522519e7a007cb1b70529f444131e0d02dfee3ff77c5b8d2d aborts in `Kokkos::Impl::throw_runtime_exception` if exceptions are disabled so this is not conflicting.

I choose to throw an exception here because that is what we are doing in similar cases. If users should be able to catch these exceptions (e.g., to try again with less local scratch memory), we should probably use a distinguishable exception type here but that seems out-of-scope for this pull request.

masterleinad:
Only `SNL-CI` is "failing" (it didn't run due to missing permissions).

core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
```diff
@@ -106,7 +106,12 @@ class TeamPolicyInternal<Kokkos::Cuda, Properties...>
             (size_t)impl_vector_length(),
             (size_t)team_scratch_size(0) + 2 * sizeof(double),
             (size_t)thread_scratch_size(0) + sizeof(double));
-    return block_size / impl_vector_length();
+    int max_block_size = block_size / impl_vector_length();
+    if (max_block_size <= 0)
```
romintomasetti, line 110: How could this ever be negative ?!

masterleinad: It's an `int` so testing for `<=0` seems to be more appropriate than testing for `==0`.

Rombur (approved):

ldh4 (approved):

ndellingwood:
> Only `SNL-CI` is "failing" (it didn't run due to missing permissions).

@masterleinad I added the SNL-AT2-SPECIAL-APPROVAL label, I think it is needed to launch the job on an SNL system. The job should not be blocking

masterleinad:
It seems we should just not test with `-fno-exceptions` with `nvcc`.

masterleinad:
> Can you come up with a test that fails?

With
```diff
diff --git a/core/unit_test/incremental/Test12a_ThreadScratch.hpp b/core/unit_test/incremental/Test12a_ThreadScratch.hpp
index 61d2fd307..383bc57ca 100644
--- a/core/unit_test/incremental/Test12a_ThreadScratch.hpp
+++ b/core/unit_test/incremental/Test12a_ThreadScratch.hpp
@@ -36,7 +36,7 @@ struct ThreadScratch {
   int sX, sY;
   data_t v;
 
-  const int scratch_level = 1;
+  const int scratch_level = 0;
   KOKKOS_FUNCTION
   void operator()(const team_t &team) const {
     // Allocate and use scratch pad memory
```
I noticed that the kernel was silently not launched.

masterleinad:
A pretty minimal example is
```C++
template <class ExecSpace>
struct ThreadScratch {
  KOKKOS_FUNCTION
  void operator()(const typename Kokkos::TeamPolicy<ExecSpace>::member_type &team) const {
  }
};

  using policy_t = Kokkos::TeamPolicy<ExecSpace>;
  using team_t   = typename Kokkos::TeamPolicy<ExecSpace>::member_type;
  using data_t   = Kokkos::View<size_t **, ExecSpace>;

  using scratch_t = Kokkos::View<size_t *, ExecSpace,
                                 Kokkos::MemoryTraits<Kokkos::Unmanaged> >;

  int scratchSize = scratch_t::shmem_size(321);
    policy_t policy =
        policy_t(1, Kokkos::AUTO, 1)
            .set_scratch_size(0, Kokkos::PerThread(scratchSize));

  int max_team_size = policy.team_size_max(ThreadScratch<ExecSpace>{}, Kokkos::ParallelForTag());
}
```

dalg24 (changes requested):
Withdraw all unrelated std::string changes

masterleinad:
> Withdraw all unrelated std::string changes
 
Here you go.

dalg24 (changes requested):
This approach looks just wrong to me.
It introduces preconditions on user-facing `TeamPolicy` query member functions.  Returning zero to denote that the kernel launch would fail with the current configuration is reasonable.  The changes proposed here would force the user to use try catch blocks.
We should raise  an error if the user actually attempt to launch with a bad configuration, not if he queries information about it.

masterleinad:
> We should raise an error if the user actually attempt to launch with a bad configuration, not if he queries information about it.

The tricky part is that we allow the user to manually set `team_size==0` with the effect of not launching the kernel. Maybe, we just discuss in the meeting tomorrow.

dalg24:
Obviously if the number of teams was not 0 (assuming 0 thread per team is not legal) and the block size internally deduced is zero we ought to error out, silently doing nothing is a terrible thing to do here.  I am just arguing that `TeamPolicy::team_size_{max,recommended}()` is not the place to raise an error.

masterleinad:
> assuming 0 thread per team is not legal

I think that is the important discussion point. Currently, we allow this and just don't execute. This pull request tries to say if you are obtaining the recommended or maximum team size then that can't be zero but otherwise, it's still OK.
We should discuss if it's OK for the user to explicitly set `team_size=0` (implying that the kernel isn't executed) or error out even in that case.

masterleinad:
Superseded by #7423.

PR #7414, opened 2024-10-08
HIP: Improve the error message when we cannot find an acceptable error block size

Rombur:
This PR improves the error message when we cannot find a `BlockSize` in a `parallel_reduce`.

The error message looks like:
```
  what():  Kokkos::Impl::ParallelReduce< HIP > could not find a valid execution configuration: your kernel requires 65552 bytes of shared memory per block but only 65536 bytes per block are available.
```
We could also add the name of the kernel but that would require passing it to the backends. Since we can already get this information using the `KernelLogger`, I didn't do it.

Here is the code I used to trigger the error (I am not convinced it is useful to make a test out of it)
```c++
#include <Kokkos_Core.hpp>

int constexpr array_size = 241; // Change the amount of shared memory required
int constexpr size = 1e6;

struct array_type
{
  double array[array_size];

  KOKKOS_INLINE_FUNCTION array_type()
  {
    for (int i=0; i<array_size; ++i)
      array[i] = 0.;
  }

  KOKKOS_INLINE_FUNCTION array_type(array_type const &rhs)
  {
    for (int i=0; i<array_size; ++i)
      array[i] = rhs.array[i];
  }

  KOKKOS_INLINE_FUNCTION array_type& operator += (array_type const &src)
  {
    for (int i=0; i<array_size; ++i)
      array[i] += src.array[i];
    
    return *this;
  }
};

namespace Kokkos
{
  template<>
  struct reduction_identity<array_type>
  {
    KOKKOS_FORCEINLINE_FUNCTION static array_type sum()
    {
      return array_type();
    }
  };
}

int main(int argc, char**argv)
{
  Kokkos::ScopeGuard scope(argc, argv);

  {
    Kokkos::View<array_type[size], Kokkos::HIPSpace> a("a");
    Kokkos::View<array_type[size], Kokkos::HIPSpace> b("b");
    array_type val;
    Kokkos::parallel_reduce("reduce", Kokkos::RangePolicy<Kokkos::HIP>(0, size),
                         KOKKOS_LAMBDA(int i , array_type &tmp)
                         {
                           for (int j=0; j<array_size; ++j)
                           tmp.array[j] += a[i].array[j] + b[i].array[j];
                         }, Kokkos::Sum<array_type>(val));
    Kokkos::fence();
  }

  return 0;
}
```

This is related to https://github.com/kokkos/kokkos/issues/7399.

### DIFF
--- a/core/src/HIP/Kokkos_HIP_ParallelReduce_Range.hpp
+++ b/core/src/HIP/Kokkos_HIP_ParallelReduce_Range.hpp
@@ -232,9 +232,21 @@ class ParallelReduce<CombinedFunctorReducerType, Kokkos::RangePolicy<Traits...>,
     if ((nwork > 0) || need_device_set) {
       const int block_size = local_block_size(m_functor_reducer.get_functor());
       if (block_size == 0) {
+        const unsigned int shared_memory_required =
+            hip_single_inter_block_reduce_scan_shmem<false, WorkTag,
+                                                     value_type>(
+                m_functor_reducer.get_functor(), HIPTraits::WarpSize);
+        const unsigned int shared_memory_available =
+            m_policy.space()
+                .impl_internal_space_instance()
+                ->m_deviceProp.maxSharedMemoryPerMultiProcessor;
         Kokkos::Impl::throw_runtime_exception(
             std::string("Kokkos::Impl::ParallelReduce< HIP > could not find a "
-                        "valid execution configuration."));
+                        "valid execution configuration: your kernel requires " +
+                        std::to_string(shared_memory_required) +
+                        " bytes of shared memory per block but only " +
+                        std::to_string(shared_memory_available) +
+                        " bytes per block are available."));
       }
 
       // REQUIRED ( 1 , N , 1 )


cwpearson (approved):

core/src/HIP/Kokkos_HIP_ParallelReduce_Range.hpp
```diff
@@ -232,9 +232,21 @@ class ParallelReduce<CombinedFunctorReducerType, Kokkos::RangePolicy<Traits...>,
...
+                .impl_internal_space_instance()
+                ->m_deviceProp.maxSharedMemoryPerMultiProcessor;
         Kokkos::Impl::throw_runtime_exception(
             std::string("Kokkos::Impl::ParallelReduce< HIP > could not find a "
-                        "valid execution configuration."));
+                        "valid execution configuration: your kernel requires " +
+                        std::to_string(shared_memory_required) +
+                        " bytes of shared memory per block but only " +
+                        std::to_string(shared_memory_available) +
+                        " bytes per block are available."));
```
tcclevenger, line 249: Does `block_size==0` exactly correspond to `shared_memory_required > shared_memory_available`?

Rombur: Yes. We use the required shared memory to calculate the block size. If we compute a `block_size` smaller than one warp, we set the `block_size` to zero.

tcclevenger (approved):

dalg24 (commented):
Don't we run into the same issue with ParallelScan?

In the original report, it was pointed out that if dropping the check in ParallelReduce then we would trigger https://github.com/kokkos/kokkos/blob/1df216f63db580a2d8172f44c05a440be982165a/core/src/HIP/Kokkos_HIP_KernelLaunch.hpp#L556-L559
where we could in principle print more information.
Can you briefly comment on here why the approach in your PR is more suited.

Rombur:
> Don't we run into the same issue with ParallelScan?

Yes, we do. I checked if we had the same issue with `parallel_for` but I forgot about `parallel_scan`. Do you want me to do it in this PR or a new one?

> In the original report, it was pointed out that if dropping the check in ParallelReduce then we would trigger 
Can you briefly comment on here why the approach in your PR is more suited.

At first I wanted to simply remove the check but we divide by the `block_size` before we get to the next check. I guess they just got lucky the program didn't stop earlier.

crtrott (approved):

crtrott:
Let's rerun testing and wait with merging. Both HIP builds failed with a version checkout issue, so no test actually compiled this thing.

crtrott:
Retest this please.

dalg24:
Unrelated OpenACC failure.  Both HIP builds passed.

PR #7417, opened 2024-10-08
OpenMP: Deprecate support for Kokkos parallel inside OpenMP parallel without nested OpenMP enabled

masterleinad:
Addressing https://github.com/kokkos/kokkos/pull/7368#issuecomment-2389330353.
This avoids using the buggy code path fixed in #7368 when we compile without deprecated code enabled.

### DIFF
--- a/core/src/OpenMP/Kokkos_OpenMP_Instance.hpp
+++ b/core/src/OpenMP/Kokkos_OpenMP_Instance.hpp
@@ -122,8 +122,18 @@ inline bool execute_in_serial(OpenMP const& space = OpenMP()) {
 #else
   bool is_nested = static_cast<bool>(omp_get_nested());
 #endif
-  return (space.impl_internal_space_instance()->get_level() < omp_get_level() &&
-          !(is_nested && (omp_get_level() == 1)));
+  bool max_parallel_level_exceeded =
+      (space.impl_internal_space_instance()->get_level() < omp_get_level() &&
+       !(is_nested && (omp_get_level() == 1)));
+
+#ifndef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  if (max_parallel_level_exceeded)
+    Kokkos::abort(
+        "Kokkos::OpenMP: Nested parallelism requires the maximum active levels "
+        "to be larger than 1!");
+#endif
+
+  return max_parallel_level_exceeded;
 }
 
 }  // namespace Impl
--- a/core/src/OpenMP/Kokkos_OpenMP_Instance.hpp
+++ b/core/src/OpenMP/Kokkos_OpenMP_Instance.hpp
@@ -130,7 +130,7 @@ inline bool execute_in_serial(OpenMP const& space = OpenMP()) {
   if (max_parallel_level_exceeded)
     Kokkos::abort(
         "Kokkos::OpenMP: Nested parallelism requires the maximum active levels "
-        "to be larger than 1!");
+        "to be larger than 1 and not more than two levels are supported!");
 #endif
 
   return max_parallel_level_exceeded;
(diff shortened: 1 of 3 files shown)


janciesko (approved):

core/src/OpenMP/Kokkos_OpenMP_Instance.hpp
```diff
@@ -122,8 +122,18 @@ inline bool execute_in_serial(OpenMP const& space = OpenMP()) {
...
-          !(is_nested && (omp_get_level() == 1)));
+  bool max_parallel_level_exceeded =
+      (space.impl_internal_space_instance()->get_level() < omp_get_level() &&
+       !(is_nested && (omp_get_level() == 1)));
+
+#ifndef KOKKOS_ENABLE_DEPRECATED_CODE_4
+  if (max_parallel_level_exceeded)
+    Kokkos::abort(
+        "Kokkos::OpenMP: Nested parallelism requires the maximum active levels "
+        "to be larger than 1!");
```
tcclevenger, line 133: Wouldn't this error message be confusing in the case that `omp_get_max_active_levels()>1` but we've exceeded that level in the internal instance?

masterleinad: The level stored in the internal instance is the OMP level the execution space instance was created on. If that level is higher than the maximum level, we are clearly already attempting to use nested parallelism. Let me try to improve the message a bit.

masterleinad: I'd like to propose deprecating creating (OpenMP) execution space instances inside a parallel OpenMP region as well which would make this check clearer as well.

tcclevenger (approved):

JBludau:
Maybe I am not understanding the issue here, but this looks like a partial bugfix not a deprecation to me. And it does not look like there is a clear path to if it achieving what it is intending (from what I understand)

JBludau (commented):
Maybe I am not understanding the issue here, but this looks like a bugfix not a deprecation to me. Why are we only fixing this for >4? Abort would indicate something serious is violated

masterleinad:
> Maybe I am not understanding the issue here, but this looks like a bugfix not a deprecation to me. 

As #7368 shows, we don't support nested parallelism (using nested OpenMP regions) properly when the number of parallel levels is 1 but it kind of worked before if you weren't setting `OMP_NUM_THREADS` to very small values.
We discussed that we wouldn't want to support any parallelism that OpenMP internally serializes. Deprecating gave codes that might be using OpenMP in this way a chance to address the new restriction instead of outright failing. That being said, I don't have a problem with just changing it for the Kokkos 5 release.

JBludau:
> > Maybe I am not understanding the issue here, but this looks like a bugfix not a deprecation to me.
> 
> As #7368 shows, we don't support nested parallelism (using nested OpenMP regions) properly when the number of parallel levels is 1 but it kind of worked before if you weren't setting `OMP_NUM_THREADS` to very small values. We discussed that we wouldn't want to support any parallelism that OpenMP internally serializes. Deprecating gave codes that might be using OpenMP in this way a chance to address the new restriction instead of outright failing. That being said, I don't have a problem with just changing it for the Kokkos 5 release.

hmm I guess that is fair

JBludau (approved):

crtrott (approved):
Yeah I think this is OK. This feature supports calling Kokkos parallel algorithms inside an OpenMP parallel region, without nested OpenMP enabled. This was always fragile and is not tested on our side. We do test it with nested OpenMP enabled and support that situation. So I think this is fine. Folks would just need to set OMP_NESTED=true and set up their OpenMP threads as N x 1 to effectively restore the old behavior in most situations I believe.

dalg24:
Make sure you add this to the 5.0 changelog

PR #7420, opened 2024-10-09
GitLab CI: Add SYCL build

masterleinad:
See https://gitlab.spack.io/masterleinad/kokkos/-/jobs/12902035 for the passing build.

### DIFF
--- a/.gitlab/hpsf-gitlab-ci.yml
+++ b/.gitlab/hpsf-gitlab-ci.yml
@@ -1,9 +1,21 @@
-test:
+NVIDIA-GH200:
   stage: test
   tags: [nvidia-gh200]
   image: masterleinad/kokkos-nvcc:12.6.1
   script:
     - cmake -B build -DKokkos_ENABLE_CUDA=ON -DKokkos_ARCH_HOPPER90=ON -DKokkos_ENABLE_IMPL_CUDA_UNIFIED_MEMORY=ON -DKokkos_ENABLE_TESTS=ON
+    - cmake --build build -j48
+    - cd build
+    - ctest -V
+
+INTEL-DATA-CENTER-MAX-1100:
+  stage: test
+  tags: [intel-data-center-max-1100]
+  image: intel/oneapi-basekit:2024.2.1-0-devel-ubuntu22.04
+  script:
+    - sycl-ls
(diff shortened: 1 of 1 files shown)


.gitlab/hpsf-gitlab-ci.yml
```diff
@@ -7,3 +7,15 @@ test:
     - cmake --build build -j48
     - cd build
     - ctest -V
+
+INTEL-DATA-CENTER-MAX-1100:
+  stage: test
+  tags: [intel-data-center-max-1100]
+  image: intel/oneapi-basekit:2024.2.1-0-devel-ubuntu22.04
+  script:
+    - sycl-ls
```
masterleinad, line 16: Not necessary but this helped having a good idea what backends (with names) are available on the machine.

Rombur (approved):

.gitlab/hpsf-gitlab-ci.yml
```diff
@@ -7,3 +7,15 @@ test:
...
+    - export ONEAPI_DEVICE_SELECTOR=level_zero:gpu
+    - cmake -B build -DCMAKE_BUILD_TYPE=Release -DCMAKE_CXX_COMPILER=icpx -DKokkos_ENABLE_SYCL=ON -DKokkos_ARCH_INTEL_PVC=ON -DKokkos_ENABLE_TESTS=ON -DCMAKE_CXX_FLAGS="-fsycl-device-code-split=per_kernel -fp-model=precise"
+    - cmake --build build -j48
+    - cd build
+    - ctest -V
```
dalg24, line 21: Fine but looking at https://gitlab.spack.io/kokkos/kokkos/-/jobs/12916716 we got ~4500L of output so we might want to use `--output-on-failure`?
Also there is a `--test-dir` option that would avoid the need for changing directory.

masterleinad: I'm not sure if having that amount of output is a problem. I would rather reduce it later (to have more information about the tests that ran) if it turns out to be a problem.

dalg24: Sure

dalg24 (approved):

PR #7418, opened 2024-10-08
tests(graph): filesystem not supported by old compilers so skip these tests

romintomasetti:
Fixes #7416.

Basically, `gcc` older than `9.1.0` and `LLVM` older than `11` would require additional link flags (that sometimes do not even fix all the undefined references).

Therefore, this PR disables the problematic tests for old host compilers that fall into the above condition.

See https://github.com/kokkos/kokkos/issues/7416#issuecomment-2400691182 for more details.

PR #7421, opened 2024-10-09
tests(graph): filesystem not supported by old compilers so skip these tests (followup of #7418)

romintomasetti:
This PR is a followup of #7418.

Instead of comparing the compiler version, we need to check the `libstdc++` version. Indeed, the most recent `hipcc` compiler could be used with an old `libstdc++` for instance.

I also applied the proposed skip messages that @dalg24 proposed in #7418.

### DIFF
--- a/core/unit_test/cuda/TestCuda_InterOp_Graph.cpp
+++ b/core/unit_test/cuda/TestCuda_InterOp_Graph.cpp
@@ -97,10 +97,12 @@ TEST_F(TEST_CATEGORY_FIXTURE(GraphInterOp), count_nodes) {
 TEST_F(TEST_CATEGORY_FIXTURE(GraphInterOp), debug_dot_print) {
 #if CUDA_VERSION < 11600
   GTEST_SKIP() << "Export a graph to DOT requires Cuda 11.6.";
-#elif KOKKOS_COMPILER_GNU < 910
-  GTEST_SKIP() << "'filesystem' is not fully supported prior to GCC 9.1.0.";
-#elif KOKKOS_COMPILER_CLANG < 1100
-  GTEST_SKIP() << "'filesystem' is not fully supported prior to LLVM 11.";
+#elif defined(_GLIBCXX_RELEASE) && _GLIBCXX_RELEASE < 9
+  GTEST_SKIP() << "The GNU C++ Library (libstdc++) versions less than 9.1 "
+                  "require linking with `-lstdc++fs`";
+#elif defined(_LIBCPP_VERSION) && _LIBCPP_VERSION < 110000
+  GTEST_SKIP() << "The LLVM C++ Standard Library (libc++) versions less than "
+                  "11 require linking with `-lc++fs`";
 #else
   graph->instantiate();
 
--- a/core/unit_test/hip/TestHIP_InterOp_Graph.cpp
+++ b/core/unit_test/hip/TestHIP_InterOp_Graph.cpp
@@ -71,10 +71,12 @@ TEST(TEST_CATEGORY, graph_promises_on_native_objects) {
 TEST(TEST_CATEGORY, graph_instantiate_and_debug_dot_print) {
 #if !defined(KOKKOS_IMPL_HIP_NATIVE_GRAPH)
   GTEST_SKIP() << "This test will not work without native graph support";
-#elif KOKKOS_COMPILER_GNU < 910
-  GTEST_SKIP() << "'filesystem' is not fully supported prior to GCC 9.1.0.";
-#elif KOKKOS_COMPILER_CLANG < 1100
-  GTEST_SKIP() << "'filesystem' is not fully supported prior to LLVM 11.";
+#elif defined(_GLIBCXX_RELEASE) && _GLIBCXX_RELEASE < 9
+  GTEST_SKIP() << "The GNU C++ Library (libstdc++) versions less than 9.1 "
+                  "require linking with `-lstdc++fs`";
+#elif defined(_LIBCPP_VERSION) && _LIBCPP_VERSION < 110000
+  GTEST_SKIP() << "The LLVM C++ Standard Library (libc++) versions less than "
+                  "11 require linking with `-lc++fs`";
 #else
   using view_t = Kokkos::View<int, Kokkos::HIP>;
 


core/unit_test/cuda/TestCuda_InterOp_Graph.cpp
```diff
@@ -97,10 +97,10 @@ TEST_F(TEST_CATEGORY_FIXTURE(GraphInterOp), count_nodes) {
 TEST_F(TEST_CATEGORY_FIXTURE(GraphInterOp), debug_dot_print) {
 #if CUDA_VERSION < 11600
```
dalg24, line 98: How is that one fine but not the ones below?

romintomasetti: It is a `Cuda` interoperability test, so I guess compiling this test without `CUDA_VERSION` could not happen ?

tcclevenger (approved):

dalg24:
FYI @lucbv (upcoming revision will hopefully help with the issues you are seeing in Kokkos-kernels)

core/unit_test/hip/TestHIP_InterOp_Graph.cpp
```diff
@@ -71,10 +71,12 @@ TEST(TEST_CATEGORY, graph_promises_on_native_objects) {
...
-#elif KOKKOS_COMPILER_GNU < 910
-  GTEST_SKIP() << "'filesystem' is not fully supported prior to GCC 9.1.0.";
-#elif KOKKOS_COMPILER_CLANG < 1100
-  GTEST_SKIP() << "'filesystem' is not fully supported prior to LLVM 11.";
+#elif defined(KOKKOS_COMPILER_GNU) && KOKKOS_COMPILER_GNU < 910
+  GTEST_SKIP()
+      << "GNU implementation prior to 9.1 requires linking with '-lstdc++fs'.";
+#elif defined(KOKKOS_COMPILER_CLANG) && KOKKOS_COMPILER_CLANG < 1100
+  GTEST_SKIP() << "LLVM implementation prior to LLVM 11.0 requires linking "
+                  "with '-lc++fs'.";
```
tcclevenger, line 79: An issue I'm running across is that clang16 on the machine I'm running with requires this link as well, even though it is version compliant. It looks like it is because I am  having to use an older versions of the GNU standard library libstdc++ (gcc 8.5).

romintomasetti: I changed the guarding conditions to compare the `libstdc++` version instead of the compiler. Does it help ?

tcclevenger: No, this did not work. I think that we cant use `__GLIBCXX__`. Here's what I get
```
[tccleve@fat1 unit_test]$ gcc --version
gcc (GCC) 8.5.0 20210514 (Red Hat 8.5.0-18)
Copyright (C) 2018 Free Software Foundation, Inc.
This is free software; see the source for copying conditions.  There is NO
warranty; not even for MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.
```
I'm not sure, but I'm guessing that 8.5 could have been updated after releases that are compliant?

nmm0: Yeah looks like that's right, we can't use `__GLIBCXX__`. GCC continues to add minor updates to older major versions and `__GLIBCXX__` just is the release date.

core/unit_test/cuda/TestCuda_InterOp_Graph.cpp
```diff
@@ -97,10 +97,12 @@ TEST_F(TEST_CATEGORY_FIXTURE(GraphInterOp), count_nodes) {
...
-#elif KOKKOS_COMPILER_GNU < 910
-  GTEST_SKIP() << "'filesystem' is not fully supported prior to GCC 9.1.0.";
-#elif KOKKOS_COMPILER_CLANG < 1100
-  GTEST_SKIP() << "'filesystem' is not fully supported prior to LLVM 11.";
+#elif defined(__GLIBCXX__) && __GLIBCXX__ < 20190503
+  GTEST_SKIP()
+      << "GNU implementation prior to 9.1 requires linking with '-lstdc++fs'.";
+#elif defined( _LIBCPP_VERSION) &&  _LIBCPP_VERSION < 110000
+  GTEST_SKIP() << "LLVM implementation prior to LLVM 11.0 requires linking "
+                  "with '-lc++fs'.";
```
dalg24, line 105:
```suggestion
  GTEST_SKIP() << "The LLVM C++ Standard Library (libc++) versions less than 11 require linking with `-lc++fs`";
```

core/unit_test/cuda/TestCuda_InterOp_Graph.cpp
```diff
@@ -97,10 +97,12 @@ TEST_F(TEST_CATEGORY_FIXTURE(GraphInterOp), count_nodes) {
 TEST_F(TEST_CATEGORY_FIXTURE(GraphInterOp), debug_dot_print) {
 #if CUDA_VERSION < 11600
   GTEST_SKIP() << "Export a graph to DOT requires Cuda 11.6.";
-#elif KOKKOS_COMPILER_GNU < 910
-  GTEST_SKIP() << "'filesystem' is not fully supported prior to GCC 9.1.0.";
-#elif KOKKOS_COMPILER_CLANG < 1100
-  GTEST_SKIP() << "'filesystem' is not fully supported prior to LLVM 11.";
+#elif defined(__GLIBCXX__) && __GLIBCXX__ < 20190503
+  GTEST_SKIP()
+      << "GNU implementation prior to 9.1 requires linking with '-lstdc++fs'.";
```
dalg24, line 102:
```suggestion
      << "The GNU C++ Library (libstdc++) versions less than 9.1 require linking with `-lstdc++fs`";
```

core/unit_test/hip/TestHIP_InterOp_Graph.cpp
masterleinad: What about `_GLIBCXX_RELEASE`?

romintomasetti: Yeah, let's use `_GLIBCXX_RELEASE` as it's done at other locations in the code base.

core/unit_test/cuda/TestCuda_InterOp_Graph.cpp
romintomasetti: :heavy_check_mark:

core/unit_test/cuda/TestCuda_InterOp_Graph.cpp
romintomasetti: :heavy_check_mark:

masterleinad:
Any reason not to apply the same fix for `SYCL`?

tcclevenger (approved):
This fixed it.

masterleinad (dismissed):
Any reason not to apply the same fix for SYCL?

romintomasetti:
> Any reason not to apply the same fix for `SYCL`?

Nobody complained for this I guess ? If you know how you would produce the same bug for the `SYCL` toolchain of your choice, you can apply the fix on my branch or in a followup. But I would do nothing for now since no one complained :angel:

dalg24:
Let's only apply if we actually see it fails so we don;t unnecessarily put more stress on the CI.

nmm0 (approved):

masterleinad:
It's probably more unlikely to trigger the same problem with `SYCL` but I think we should apply the same fix for consistency. In our CI, I see `_GLIBCXX_RELEASE` is 11 and on the testbeds 12 but it's conceivable to use an older libstdc++. For completeness, `_LIBCPP_VERSION` isn't defined.
I'm fine with adding the same workaround in a follow-up but the CI isn't very far in testing anyway.

dalg24:
Ignoring NVC++ segfault in the OpenACC build and GPU not detected in a couple CUDA builds

PR #7423, opened 2024-10-09
Check that internal uses of team_size_recommended return positive values

masterleinad:
Supersedes #7407. Related to #7399.
```
template <class ExecSpace>
struct ThreadScratch {
  KOKKOS_FUNCTION
  void operator()(const typename Kokkos::TeamPolicy<ExecSpace>::member_type &team) const {
  }
};

  using policy_t = Kokkos::TeamPolicy<ExecSpace>;
  using team_t   = typename Kokkos::TeamPolicy<ExecSpace>::member_type;
  using data_t   = Kokkos::View<size_t **, ExecSpace>;

  using scratch_t = Kokkos::View<size_t *, ExecSpace,
                                 Kokkos::MemoryTraits<Kokkos::Unmanaged> >;

  int scratchSize = scratch_t::shmem_size(321);
    policy_t policy =
        policy_t(1, Kokkos::AUTO, 1)
            .set_scratch_size(0, Kokkos::PerThread(scratchSize));

  int max_team_size = policy.team_size_max(ThreadScratch<ExecSpace>{}, Kokkos::ParallelForTag());

  Kokkos::parallel_for(
        "Test12a_ThreadScratch",
        policy_t(1, max_team_size).set_scratch_size(0, Kokkos::PerThread(scratchSize)),
        *this);
}
```
would currently silently not run. We discussed that it's the users' job to make sure that `max_team_size` is positive (which is what the drive-by change does). On the other hand, we should make sure that internal uses of `team_size_max/recommended` are checked. Since `team_size_max` isn't used internally, that only leaves `team_sizee_recommended`.

### DIFF
--- a/core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
@@ -539,9 +539,14 @@ class ParallelFor<FunctorType, Kokkos::TeamPolicy<Properties...>,
         m_vector_size(arg_policy.impl_vector_length()) {
     auto internal_space_instance =
         m_policy.space().impl_internal_space_instance();
-    m_team_size = m_team_size >= 0 ? m_team_size
-                                   : arg_policy.team_size_recommended(
-                                         arg_functor, ParallelForTag());
+    if (m_team_size < 0) {
+      m_team_size =
+          arg_policy.team_size_recommended(arg_functor, ParallelForTag());
+      if (m_team_size <= 0)
+        Kokkos::Impl::throw_runtime_exception(
+            "Kokkos::Impl::ParallelFor<Cuda, TeamPolicy> could not find a "
+            "valid execution configuration.");
+    }
 
     m_shmem_begin = (sizeof(double) * (m_team_size + 2));
     m_shmem_size =
@@ -896,11 +901,16 @@ class ParallelReduce<CombinedFunctorReducerType,
         m_vector_size(arg_policy.impl_vector_length()) {
     auto internal_space_instance =
         m_policy.space().impl_internal_space_instance();
-    m_team_size = m_team_size >= 0 ? m_team_size
-                                   : arg_policy.team_size_recommended(
-                                         arg_functor_reducer.get_functor(),
-                                         arg_functor_reducer.get_reducer(),
-                                         ParallelReduceTag());
+
+    if (m_team_size < 0) {
+      m_team_size = arg_policy.team_size_recommended(
+          arg_functor_reducer.get_functor(), arg_functor_reducer.get_reducer(),
+          ParallelReduceTag());
+      if (m_team_size <= 0)
+        Kokkos::Impl::throw_runtime_exception(
+            "Kokkos::Impl::ParallelReduce<Cuda, TeamPolicy> could not find a "
+            "valid execution configuration.");
+    }
 
     m_team_begin =
         UseShflReduction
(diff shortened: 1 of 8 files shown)


core/src/SYCL/Kokkos_SYCL_ParallelFor_Team.hpp
```diff
@@ -164,10 +164,14 @@ class Kokkos::Impl::ParallelFor<FunctorType, Kokkos::TeamPolicy<Properties...>,
         m_league_size(arg_policy.league_size()),
         m_team_size(arg_policy.team_size()),
         m_vector_size(arg_policy.impl_vector_length()) {
-    // FIXME_SYCL optimize
```
nmm0, line 167: What was this comment referring to? Is it something these changes fix?

core/unit_test/incremental/Test12a_ThreadScratch.hpp
```diff
@@ -75,7 +75,8 @@ struct ThreadScratch {
             .set_scratch_size(scratch_level, Kokkos::PerThread(scratchSize));
 
     int max_team_size = policy.team_size_max(*this, Kokkos::ParallelForTag());
-    v                 = data_t("Matrix", pN, max_team_size);
+    ASSERT_GT(max_team_size, 0);
+    v = data_t("Matrix", pN, max_team_size);
```
nmm0, line 79: Should we add a test for the throw case? Is it reproduceable across platforms?

core/src/SYCL/Kokkos_SYCL_ParallelFor_Team.hpp
masterleinad: In my opinion there is just nothing to fix anymore. We can of course try to improve `team_size_recommended` but the logic here is fine.

core/unit_test/incremental/Test12a_ThreadScratch.hpp
masterleinad: I don't think there is a very consistent and backend-independent way to trigger the exceptions added here. Note that there is no requirement for the `Parallel(For|Reduce)` implementations to call `team_size_recommended` or `team_size_max` (even though we basically want the team size from `team_size_recommended` when `Kokkos::AUTO` is provided).
The host backends might just always use `team_size=1` for example.

lucbv (approved):

nmm0 (approved):

core/unit_test/incremental/Test12a_ThreadScratch.hpp
crtrott: this change seems unnecessary.

masterleinad: Right. My point was rather to demonstrate that checking is important here since switching the scratch level here silently avoided running the kernel.

core/src/Cuda/Kokkos_Cuda_Parallel_Team.hpp
```diff
@@ -539,9 +539,14 @@ class ParallelFor<FunctorType, Kokkos::TeamPolicy<Properties...>,
...
-    m_team_size = m_team_size >= 0 ? m_team_size
-                                   : arg_policy.team_size_recommended(
-                                         arg_functor, ParallelForTag());
+    if (m_team_size < 0) {
+      m_team_size =
+          arg_policy.team_size_recommended(arg_functor, ParallelForTag());
+      if (m_team_size <= 0)
+        Kokkos::Impl::throw_runtime_exception(
+            "Kokkos::Impl::ParallelFor<Cuda, TeamPolicy> could not find a "
+            "valid execution configuration.");
```
ldh4, line 548: Should this check and throw be part of `team_size_recommended`, seeing that this check is always paired with  that function call in this PR?

ldh4 (approved):

dalg24:
Fine but we really need to add tests that purposefully try to request a team size that exceeds the max value and other that request more scratch memory that available.

Issue #7422, opened 2024-10-09
Could not build with OpenMP on Windows (error : unknown type name 'handle_type')


hhrecode001:
**Describe the bug**
I'm trying to install Kokkos on Windows with OpenMP.

Here is my cmake commands:
```
cmake -S ".." -B . -DCMAKE_INSTALL_PREFIX="..\install_dir_openmp" -DKokkos_ENABLE_OPENMP=ON
```
Result:
```
-- Building for: Visual Studio 17 2022
-- Setting default Kokkos CXX standard to 17
-- Selecting Windows SDK version 10.0.22621.0 to target Windows 10.0.22631.
-- The CXX compiler identification is MSVC 19.41.34120.0
-- Detecting CXX compiler ABI info
-- Detecting CXX compiler ABI info - done
-- Check for working CXX compiler: C:/Program Files/Microsoft Visual Studio/2022/Professional/VC/Tools/MSVC/14.41.34120/bin/Hostx64/x64/cl.exe - skipped
-- Detecting CXX compile features
-- Detecting CXX compile features - done
-- Setting build type to 'RelWithDebInfo' as none was specified.
-- Kokkos version: 4.4.99
-- The project name is: Kokkos
-- Using internal gtest for testing
-- Configured git information in C:/Users/huyho/source/repos/new/kokkos/build/generated/Kokkos_Version_Info.cpp
-- Using -std:c++17 for C++17 extensions as feature
-- Built-in Execution Spaces:
--     Device Parallel: NoTypeDefined
--     Host Parallel: Kokkos::OpenMP
--       Host Serial: NONE
--
-- Architectures:
-- Found OpenMP_CXX: -openmp (found version "2.0")
-- Found OpenMP: TRUE (found version "2.0") found components: CXX
-- Using internal desul_atomics copy
-- Experimental mdspan support is enabled
-- Looking for C++ include experimental/mdspan
-- Looking for C++ include experimental/mdspan - not found
-- Looking for C++ include mdspan
-- Looking for C++ include mdspan - found
-- Using internal mdspan directory C:/Users/huyho/source/repos/new/kokkos/tpls/mdspan/include
-- Kokkos Backends: OPENMP
-- Configuring done (6.1s)
-- Generating done (0.3s)
...
```

Here is my build command: `cmake --build . --config Release` , but then I got this error message:
```
...
[...]

PR #7424, opened 2024-10-09
Enforce minimum required version for the OpenMP implementation detected at configuration time

dalg24:
Related to #7422 
Recent MSVC compilers support OpenMP 2.0 which is not sufficient to build our OpenMP backend.
We should diagnose that at configuration time.

I propose to pass a version argument to the find_package() call.


aprokop (approved):

nmm0 (approved):

romintomasetti:
I guess that this would diserve an entry to the change log, or to the documentation of the backend ? Making a note that `OpenMP` 3.0 is the minimum required standard would be useful.

Issue #7403, opened 2024-10-04
OpenACC issue with MDRangePolicy


crtrott:
In the BasicView PR (#7385) we ran into an issue at the end with OpenACC. 
The actual problem is hard to pin down. Two tests are under some conditions crashing (i.e. `basic_view_access` and `basic_view_atomic_access`) but if one disables (gtest filter) the extents ctor test (which runs fine before the other tests crash) everything else passes. Running any of the tests individually also passes. However, `basic_view_access` (and the atomic version) are not clean in compute-sanitizer (the ctor test is clean).

The problem identified by compute-sanitizer (where it will actually crash the test even running it alone) does not go away even when the test is compiled standalone. In fact the following very simple test is not clean in compute-sanitizer running standalone either, and failing with the same problem:

```c++
void foo() {
  Kokkos::parallel_for(Kokkos::MDRangePolicy<Kokkos::Rank<2>>({0,0},{64,64}), KOKKOS_LAMBDA(int i, int j) {});
  Kokkos::fence();
}

TEST(TEST_CATEGORY, basic_view_access) {
  foo();
}
```

Printing stuff like this:
```
========= COMPUTE-SANITIZER
[==========] Running 1 test from 1 test suite.
[----------] Global test environment set-up.
[----------] 1 test from defaultdevicetype
[ RUN      ] defaultdevicetype.basic_view_access
========= Invalid __shared__ read of size 8 bytes
=========     at void Kokkos::Experimental::Impl::OpenACCParallelForMDRangePolicy_37_gpu<Kokkos::Experimental::Impl::FunctorAdapter<Test::foo()::[lambda(int, int) (instance 1)], Kokkos::MDRangePolicy<Kokkos::Rank<(unsigned int)2, (Kokkos::Iterate)0, (Kokkos::Iterate)0>>, (Kokkos::Experimental::Impl::RoutineClause)1>>(Kokkos::Experimental::Impl::OpenACCCollapse, std::integral_constant<Kokkos::Iterate, (Kokkos::Iterate)1>, const T1 &, const Kokkos::Array<long, (unsigned long)2, void> &, const Kokkos::Array<long, (unsigned long)2, void> &, int)+0x970 in /ascldap/users/crtrott/Kokkos/kokkos/core/src/OpenACC/Kokkos_OpenACC_ParallelFor_MDRange.hpp:37
=========     by thread (0,0,0) in block (0,0,0)
=========     Address 0x0 is out of bounds
[...]

PR #7425, opened 2024-10-10
Fix OpenACC MDRangePolicy errors (issue #7403)

seyonglee:
This PR fixes the OpenACC MDRangePolicy errors (issue #7403).
OpenACC parallel constructs with MDRangePolicy and the new BasicView suffer from dangling pointer problems when compiled by NVHPC compiler (tested with V23.9 and 24.5). This problem disappears if loop-collapsing transformation is manually applied to the MDRangePolicy loops. This PR applies the manual loop-collapsing transformation only when compiled by NVHPC compiler.

### DIFF
--- a/core/src/OpenACC/Kokkos_OpenACC_ParallelFor_MDRange.hpp
+++ b/core/src/OpenACC/Kokkos_OpenACC_ParallelFor_MDRange.hpp
@@ -41,7 +41,7 @@ void OpenACCParallelForMDRangePolicy(OpenACCCollapse, OpenACCIterateLeft,
 // clang-format off
 #pragma acc parallel loop gang vector copyin(functor) async(async_arg)
   // clang-format on
-  for (auto m = 0; m < nIter; ++m) {
+  for (decltype(nIter) m = 0; m < nIter; ++m) {
     auto i1 = m / dim0 + begin1;
     auto i0 = m % dim0 + begin0;
     functor(i0, i1);
@@ -75,7 +75,7 @@ void OpenACCParallelForMDRangePolicy(OpenACCCollapse, OpenACCIterateRight,
 // clang-format off
 #pragma acc parallel loop gang vector copyin(functor) async(async_arg)
   // clang-format on
-  for (auto m = 0; m < nIter; ++m) {
+  for (decltype(nIter) m = 0; m < nIter; ++m) {
     auto i0 = m / dim1 + begin0;
     auto i1 = m % dim1 + begin1;
     functor(i0, i1);
@@ -158,7 +158,7 @@ void OpenACCParallelForMDRangePolicy(OpenACCCollapse, OpenACCIterateLeft,
 // clang-format off
 #pragma acc parallel loop gang vector copyin(functor) async(async_arg)
   // clang-format on
-  for (auto m = 0; m < nIter; ++m) {
+  for (decltype(nIter) m = 0; m < nIter; ++m) {
     auto tmp1 = dim1 * dim0;
     auto i2   = m / tmp1 + begin2;
     auto tmp2 = m % tmp1;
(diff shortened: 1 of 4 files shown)


dalg24:
Any idea what that "dangling pointer" issue is about?

seyonglee:
> Any idea what that "dangling pointer" issue is about?

When an OpenACC parallel construct with MDRangePolicy and the new BasicView is compiled by the NVHPC compiler, the execution fails with illegal address errors (`CUDA_ERROR_ILLEGAL_ADDRESS`) and NVIDIA compute-sanitizer complains multiple invalid memory accesses with the following comment: "Launch failed (often invalid pointer dereference)".
I guess that there is a NVHPC compiler bug in handling OpenACC `collapse` clause when deeply nested loops have complex functor body. (Actual behavior varies depending on the optimization level (`-O2` or higher), the complexity of functor body, and loop nesting level.) However, if collapse transformation is manually applied, those problem disappears.

core/src/OpenACC/Kokkos_OpenACC_ParallelFor_MDRange.hpp
```diff
@@ -30,10 +30,23 @@ void OpenACCParallelForMDRangePolicy(OpenACCCollapse, OpenACCIterateLeft,
...
+  auto begin0 = begin[0];
+  auto end0   = end[0];
+#if defined(KOKKOS_ENABLE_OPENACC_COLLAPSE_MDRANGE_LOOPS)
+  auto dim1  = end1 - begin1;
+  auto dim0  = end0 - begin0;
+  auto nIter = dim1 * dim0;
+// clang-format off
+#pragma acc parallel loop gang vector copyin(functor) async(async_arg)
+  // clang-format on
+  for (auto m = 0; m < nIter; ++m) {
```
crtrott, line 44: I appreciate the sentiment of the "auto" but `m` is going to be `int` this way, and if nIter turns out to be unsigned you still get a warning.

crtrott, line 44:
```suggestion
  for (decltype(nIter) m = 0; m < nIter; ++m) {
```

crtrott (commented):
This feels like a LOT of extra code is it worth it?

core/src/OpenACC/Kokkos_OpenACC_ParallelFor_MDRange.hpp
crtrott: You need to check for other places like this.

seyonglee: Updated.

seyonglee: Updated.

core/src/OpenACC/Kokkos_OpenACC_ParallelFor_MDRange.hpp
```diff
@@ -30,10 +30,23 @@ void OpenACCParallelForMDRangePolicy(OpenACCCollapse, OpenACCIterateLeft,
...
-  int begin0 = begin[0];
-  int end0   = end[0];
+  auto begin1 = begin[1];
+  auto end1   = end[1];
+  auto begin0 = begin[0];
+  auto end0   = end[0];
+#if defined(KOKKOS_ENABLE_OPENACC_COLLAPSE_MDRANGE_LOOPS)
+  auto dim1  = end1 - begin1;
+  auto dim0  = end0 - begin0;
+  auto nIter = dim1 * dim0;
```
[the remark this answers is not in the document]
seyonglee: As I know, there is no easy way to detect overflow; we will have to test values (dim0, dim1, … in this case) before performing operations (multiplication in this case), which requires to check multiple cases that can cause the overflow. Please let me know if there is any easy way to check the overflow.

tcclevenger (approved):

tcclevenger (approved):
Manually compared these indices calculation from the flat index to the multi-dim version for some test bounds and did not find any errors.

crtrott (approved):

PR #7426, opened 2024-10-10
SYCL: InterOp_Graph skip dot file when filesystem isn't fully supported

masterleinad:
Addressing https://github.com/kokkos/kokkos/pull/7421#issuecomment-2403484898:
> It's probably more unlikely to trigger the same problem with SYCL but I think we should apply the same fix for consistency. In our CI, I see _GLIBCXX_RELEASE is 11 and on the testbeds 12 but it's conceivable to use an older libstdc++. For completeness, _LIBCPP_VERSION isn't defined.

### DIFF
--- a/core/unit_test/sycl/TestSYCL_InterOp_Graph.cpp
+++ b/core/unit_test/sycl/TestSYCL_InterOp_Graph.cpp
@@ -80,6 +80,15 @@ TEST(TEST_CATEGORY, graph_instantiate_and_debug_dot_print) {
 
   ASSERT_EQ(graph.native_graph().get_nodes().size(), 2u);
 
+#if defined(_GLIBCXX_RELEASE) && _GLIBCXX_RELEASE < 9
+  GTEST_SKIP()
+      << "The GNU C++ Library (libstdc++) versions less than 9.1 "
+         "require linking with `-lstdc++fs` when using std::filesystem";
+#elif defined(_LIBCPP_VERSION) && _LIBCPP_VERSION < 110000
+  GTEST_SKIP()
+      << "The LLVM C++ Standard Library (libc++) versions less than "
+         "11 require linking with `-lc++fs` when using std::filesystem";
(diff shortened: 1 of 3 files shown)


dalg24:
Then why did you accept https://github.com/kokkos/kokkos/pull/7421#issuecomment-2403461976

masterleinad:
> Then why did you accept [#7421 (comment)](https://github.com/kokkos/kokkos/pull/7421#issuecomment-2403461976)

I didn't agree with it. My comment said that I'm fine doing it for `SYCL` in a follow-up and do I dismissed my review to avoid blocking.

core/unit_test/sycl/TestSYCL_InterOp_Graph.cpp
```diff
@@ -80,6 +80,15 @@ TEST(TEST_CATEGORY, graph_instantiate_and_debug_dot_print) {
 
   ASSERT_EQ(graph.native_graph().get_nodes().size(), 2u);
 
+#if defined(_GLIBCXX_RELEASE) && _GLIBCXX_RELEASE < 9
+  GTEST_SKIP()
+      << "The GNU C++ Library (libstdc++) versions less than 9.1 "
+         "require linking with `-lstdc++fs` when using std::filesystem";
```
romintomasetti, line 86:
```suggestion
         "require linking with `-lstdc++fs`";
```
Please keep messages aligned with what we did elsewhere.

masterleinad: This is a better comment in my opinion. I'm happy to change it for the CUDA and HIP test as well if you prefer.

romintomasetti: Yes, please.

nmm0 (approved):

romintomasetti (approved):

PR #7430, opened 2024-10-11
More multiplication result converted to large type fixes

dalg24:
These are were the last ones in our headers.
After this we still need to look into benchmarks and tests.

It is a bit like warnings, they are quite annoying at times but if we don't strive to avoid them we might miss the one that matters.

### DIFF
--- a/core/src/Kokkos_MemoryPool.hpp
+++ b/core/src/Kokkos_MemoryPool.hpp
@@ -196,9 +196,10 @@ class MemoryPool {
 
         stats.consumed_superblocks++;
         stats.consumed_blocks += block_used;
-        stats.consumed_bytes += block_used * block_size;
+        stats.consumed_bytes += static_cast<size_t>(block_used) * block_size;
         stats.reserved_blocks += block_count - block_used;
-        stats.reserved_bytes += (block_count - block_used) * block_size;
+        stats.reserved_bytes +=
+            static_cast<size_t>(block_count - block_used) * block_size;
       }
     }
 
--- a/core/src/Kokkos_ScratchSpace.hpp
+++ b/core/src/Kokkos_ScratchSpace.hpp
@@ -110,7 +110,7 @@ class ScratchMemorySpace {
     // Note: for team scratch m_offset is 0, since every
     // thread will get back the same shared pointer
     void* tmp           = m_iter + m_offset * size;
-    uintptr_t increment = size * m_multiplier;
+    uintptr_t increment = static_cast<uintptr_t>(size) * m_multiplier;
 
     // Cast to uintptr_t to avoid problems with pointer arithmetic using SYCL
     const auto end_iter =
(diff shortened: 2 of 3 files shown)


core/src/Kokkos_MemoryPool.hpp
```diff
@@ -196,9 +196,10 @@ class MemoryPool {
 
         stats.consumed_superblocks++;
         stats.consumed_blocks += block_used;
-        stats.consumed_bytes += block_used * block_size;
+        stats.consumed_bytes += static_cast<size_t>(block_used) * block_size;
         stats.reserved_blocks += block_count - block_used;
-        stats.reserved_bytes += (block_count - block_used) * block_size;
+        stats.reserved_bytes +=
+            static_cast<size_t>(block_count - block_used) * block_size;
```
dalg24, line 202: Definitions here
https://github.com/kokkos/kokkos/blob/7d64c3d6581d2c7be03c48e8feea9408f89ca735/core/src/Kokkos_MemoryPool.hpp#L154-L156

core/src/impl/Kokkos_HostThreadTeam.hpp
```diff
@@ -358,7 +358,7 @@ class HostThreadTeamData {
     int const num  = (m_work_end + m_work_chunk - 1) / m_work_chunk;
     int const part = (num + m_league_size - 1) / m_league_size;
 
-    m_work_range.first  = part * m_league_rank;
+    m_work_range.first  = static_cast<int64_t>(part) * m_league_rank;
```
dalg24, line 361: https://github.com/kokkos/kokkos/blob/7d64c3d6581d2c7be03c48e8feea9408f89ca735/core/src/impl/Kokkos_HostThreadTeam.hpp#L78-L80

core/src/Kokkos_ScratchSpace.hpp
```diff
@@ -110,7 +110,7 @@ class ScratchMemorySpace {
     // Note: for team scratch m_offset is 0, since every
     // thread will get back the same shared pointer
     void* tmp           = m_iter + m_offset * size;
-    uintptr_t increment = size * m_multiplier;
+    uintptr_t increment = static_cast<uintptr_t>(size) * m_multiplier;
```
romintomasetti, line 113: Probably worth tested.

romintomasetti: So, is it a compiler warning/error you're fixing or rather some runtime issue?

dalg24: https://github.com/kokkos/kokkos/security/code-scanning/24

romintomasetti: I guess it's worth adding this `clang-tidy` check once you're done.

dalg24: Sure.  It looks like it was added in LLVM 13.

crtrott (approved):

masterleinad (approved):

dalg24:
Ignoring shared space failure in CUDA build

core/src/Kokkos_ScratchSpace.hpp
romintomasetti: @dalg24 Followup then ?

dalg24: What are you asking?

dalg24: Sure we probably should work towards enabling all the bugprone-* checks, but no I am not working on it.

PR #7440, opened 2024-10-14
CI: Enable deprecation warnings

masterleinad:
We now transitioned to using pragma push/pop to avoid deprecation warnings in tests which means that we don't have to disable deprecation warnings via a compiler flag for the CI.

- Drive-by: Don't use `-Wsuggest-override` with `icpc`.
- the current suppression macro doesn't work properly with `icpc` so we just `-diag-disable` instead in order not too make the CI matrix more complicated
- `Kokkos::Impl::ThreadsInternal::in_paralel` was marked deprecated but can just removed when deprecated code is disabled. It's sufficient for `Kokkos::Threads::in_parallel` to be deprecated. The deprecation of the former caused warnings when used in the latter.


janciesko:
Thanks, I've seen this issue on some of the CI runs.

containers/unit_tests/TestOffsetView.hpp
```diff
@@ -57,11 +57,15 @@ void test_offsetview_construction() {
 
   ASSERT_EQ("firstOV", ov.label());
 
+#ifdef KOKKOS_ENABLE_DEPRECATION_WARNINGS
```
nliber, line 60: Question (and not a blocker):  why not put the `KOKKOS_ENABLE_DEPRECATION_WARNINGS` check directly in the `KOKKOS_IMPL_DISABLE_DEPRECATED_WARNINGS_PUSH` and `KOKKOS_IMPL_DISABLE_DEPRECATED_WARNINGS_POP` macros?

nliber (approved):

containers/unit_tests/TestOffsetView.hpp
dalg24: Because we might want to disable deprecated warnings not emitted by us and because then the macro does what it says.

Rombur:
Retest this please

janciesko (approved):

core/perf_test/test_taskdag.cpp
```diff
@@ -226,6 +226,8 @@ int main(int argc, char* argv[]) {
   return 0;
 }
 
+#ifdef KOKKOS_ENABLE_DEPRECATION_WARNINGS
```
janciesko, line 229: It looks weird that we enable deprecation warnings just to disable them in next line. Can you please elaborate? Also I am not sure if I see the matching `KOKKOS_IMPL_DISABLE_DEPRECATED_WARNINGS_PUSH()`

masterleinad: `KOKKOS_ENABLE_DEPRECARTION_WARNINGS` makes `KOKKOS_DEPRECATED` expand to `[[deprecated]]`.
`KOKKOS_IMPL_DISABLE_DEPRECATED_WARNINGS_PUSH/POP` however suppresses compiler warnings like `-Wdeprecated-declarations` and thus covers more than just the deprecation warnings coming from `Kokkos`.

masterleinad: > Also I am not sure if I see the matching `KOKKOS_IMPL_DISABLE_DEPRECATED_WARNINGS_PUSH()`

Fixed in fb7810747.

containers/unit_tests/TestOffsetView.hpp
janciesko: Is this the use-case for the variety of macros you have in mind? 

```c++
#ifdef KOKKOS_ENABLE_DEPRECATION_WARNINGS //expand KOKKOS_DEPRECATED macro ([[deprecated]])
using old_type KOKKOS_DEPRECATED_WITH_COMMENT(
      "Use new type") = int; //Compiler emits warning
KOKKOS_IMPL_DISABLE_DEPRECATED_WARNINGS_PUSH()  //Selectively disable warning just for this
using another_old_type KOKKOS_DEPRECATED_WITH_COMMENT(
      "Use another new type") = int;
KOKKOS_IMPL_DISABLE_DEPRECATED_WARNINGS_POP()
#endif
#endif
```

core/unit_test/TestArrayOps.hpp
```diff
@@ -137,7 +137,9 @@ TEST(TEST_CATEGORY, array_zero_data_nullptr) {
 }
 
 #ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+#ifdef KOKKOS_ENABLE_DEPRECATION_WARNINGS
```
dalg24, line 140: Not worth the trouble

core/src/Threads/Kokkos_Threads_Instance.hpp
```diff
@@ -403,7 +403,7 @@ class ThreadsInternal {
   static void start(void (*)(ThreadsInternal &, const void *), const void *);
 
 #ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
-  KOKKOS_DEPRECATED static int in_parallel();
+  static int in_parallel();
 #endif
```
dalg24, line 407: ???
not captured in the description

cmake/kokkos_arch.cmake
```diff
@@ -163,6 +163,11 @@ if(KOKKOS_ENABLE_COMPILER_WARNINGS)
     endif()
   endif()
 
+  # ICPC doesn't support -Wsuggest-override
+  if(KOKKOS_CXX_COMPILER_ID STREQUAL Intel)
+    list(REMOVE_ITEM COMMON_WARNINGS "-Wsuggest-override")
+  endif()
```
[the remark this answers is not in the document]
masterleinad: Drive-by change. The compiler just complains for very target but doesn't error out.

.github/workflows/continuous-integration-linux.yml
```diff
@@ -24,13 +24,13 @@ jobs:
         include:
           - distro: 'ubuntu:intel'
             cxx: 'icpc'
-            cxx_extra_flags: '-diag-disable=177,10441'
+            cxx_extra_flags: '-diag-disable=177,1478,1786,10441'
```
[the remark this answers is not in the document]
masterleinad: Deprecation warnings. Apparently, our current macro doesn't work properly with `icpx` although we are using the same warning numbers for `EDG`. I didn't attempt to diagnose this further.

core/src/Threads/Kokkos_Threads_Instance.hpp
masterleinad: It's internal in the `Impl` namespace so we don't need to deprecate (which causes warnings when used in the non-`Impl` overload).

masterleinad: Updated the description of the pull request.

.github/workflows/continuous-integration-linux.yml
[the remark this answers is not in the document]
dalg24: Both?  Please post a reference to the doc and/or to the warning message

.github/workflows/continuous-integration-linux.yml
dalg24, line 1: Did we preserve any build with deprecation warnings off?

masterleinad: No, we don't.

dalg24: Why wouldn't we?  We also care about that case.

masterleinad: The effect of disabling deprecation warnigs is that `KOKKOS_DEPRECATED` and `KOKKOS_DEPRECATED_WITH_COMMENT` expand to nothing. Therefore, we strictly test less when we disable deprecation warnings. The only advantage would be that we don't hide other deprecations not caused by Kokkos in guarded code (which we try to keep pretty minimal). I don't mind if we do that for some builds that where the configuration is mostly redundant anyway.

dalg24: We test that if we are not supposed to emit deprecation warnings we don't

.github/workflows/continuous-integration-linux.yml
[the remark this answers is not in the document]
masterleinad: See https://github.com/masterleinad/kokkos/actions/runs/11355055477/job/31584316036 and https://github.com/masterleinad/kokkos/actions/runs/11355055477/job/31584316417. They show:
```
/__w/kokkos/kokkos/core/src/impl/Kokkos_MultipleTaskQueue.hpp(441): error #1478: class "Kokkos::SimpleTaskScheduler<ExecSpace, Kokkos::Impl::MultipleTaskQueue<ExecSpace, Kokkos::Impl::default_tasking_memory_space_for_execution_space_t<ExecSpace>, Kokkos::Impl::TaskQueueTraitsLockBased, Kokkos::MemoryPool<Kokkos::Device<Kokkos::HostSpace::execution_space, Kokkos::HostSpace::memory_space>>>>" (declared at line 67 of "/__w/kokkos/kokkos/core/src/impl/Kokkos_SimpleTaskScheduler.hpp") was declared
          deprecated
      using specialization = Impl::TaskQueueSpecialization<
            ^
          detected during:
            instantiation of "size_t={unsigned long} Kokkos::Impl::MultipleTaskQueue<ExecSpace, MemorySpace, TaskQueueTraits, MemoryPool>::task_queue_allocation_size(const Kokkos::Impl::TaskQueueMemoryManager<ExecSpace, MemorySpace, MemoryPool>::execution_space &, const Kokkos::Impl::TaskQueueMemoryManager<ExecSpace, MemorySpace, MemoryPool>::memory_space &, const Kokkos::Impl::TaskQueueMemoryManager<ExecSpace, MemorySpace, MemoryPool>::memory_pool &) [with ExecSpace=ExecSpace,
                      MemorySpace=Kokkos::Impl::default_tasking_memory_space_for_execution_space_t<ExecSpace>, TaskQueueTraits=Kokkos::Impl::TaskQueueTraitsLockBased, MemoryPool=Kokkos::MemoryPool<Kokkos::Device<Kokkos::HostSpace::execution_space, Kokkos::HostSpace::memory_space>>]" at line 193 of "/__w/kokkos/kokkos/core/src/impl/Kokkos_SimpleTaskScheduler.hpp"
            instantiation of "Kokkos::SimpleTaskScheduler<ExecSpace, QueueType>::SimpleTaskScheduler(const Kokkos::SimpleTaskScheduler<ExecSpace, QueueType>::execution_space &, const Kokkos::SimpleTaskScheduler<ExecSpace, QueueType>::memory_space &, const Kokkos::SimpleTaskScheduler<ExecSpace, QueueType>::memory_pool &) [with ExecSpace=ExecSpace, QueueType=Kokkos::Impl::MultipleTaskQueue<ExecSpace, Kokkos::Impl::default_tasking_memory_space_for_execution_space_t<ExecSpace>,
                      Kokkos::Impl::TaskQueueTraitsLockBased, Kokkos::MemoryPool<Kokkos::Device<Kokkos::HostSpace::execution_space, Kokkos::HostSpace::memory_space>>>]" at line 233 of "/__w/kokkos/kokkos/core/src/impl/Kokkos_SimpleTaskScheduler.hpp"
            instantiation of "Kokkos::SimpleTaskScheduler<ExecSpace, QueueType>::SimpleTaskScheduler(const Kokkos::SimpleTaskScheduler<ExecSpace, QueueType>::memory_space &, size_t={unsigned long}, unsigned int, unsigned int, unsigned int) [with ExecSpace=ExecSpace, QueueType=Kokkos::Impl::MultipleTaskQueue<ExecSpace, Kokkos::Impl::default_tasking_memory_space_for_execution_space_t<ExecSpace>, Kokkos::Impl::TaskQueueTraitsLockBased,
                      Kokkos::MemoryPool<Kokkos::Device<Kokkos::HostSpace::execution_space, Kokkos::HostSpace::memory_space>>>]" at line 169 of "/__w/kokkos/kokkos/core/perf_test/test_taskdag.cpp"
```
and
```
/__w/kokkos/kokkos/containers/unit_tests/TestOffsetView.hpp(64): error #1786: constant "Kokkos::View<DataType, Properties...>::Rank [with DataType=int **, Properties=<Kokkos::OpenMP>]" (declared at line 321 of "/__w/kokkos/kokkos/core/src/View/Kokkos_ViewLegacy.hpp") was declared deprecated ("Use rank instead.")
    ASSERT_EQ(2u, ov.Rank);
    ^
          detected during instantiation of "void Test::test_offsetview_construction<Scalar,Device>() [with Scalar=int, Device=Kokkos::OpenMP]" at line 746
```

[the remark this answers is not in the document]
dalg24: ok

dalg24:
Ignoring NVHPC segfault

containers/unit_tests/TestOffsetView.hpp
janciesko: :(

PR #7441, opened 2024-10-14
Introduce index_type and standardize size_type across Kokkos backends

janciesko:
This PR introduces a signed `index_type` for each memory and execution space and standardizes `size_type` as unsigned across all Kokkos backends. This aligns Kokkos with the mdspan model where `index_type` is signed and `size_type` is unsigned.

### Motivation
- **Signed indexing**: Signed indices are safer for arithmetic (e.g., negative ranges) and avoid undefined behavior in overflow cases.
- **Consistency**: Different backends previously used different index/size semantics; this makes them uniform.
- **mdspan compatibility**: `std::mdspan` uses `index_type` (signed) and `size_type` (unsigned); Kokkos Views should mirror this.

### Changes
Memory spaces
- Add `index_type = std::make_signed_t<size_type>` to all memory spaces:
  - Cuda: `CudaSpace`, `CudaUVMSpace`, `CudaHostPinnedSpace`
  - HIP: `HIPSpace`, `HIPHostPinnedSpace`, `HIPManagedSpace`
  - Host: `HostSpace`, `AnonymousSpace`
  - OpenACC: `OpenACCSpace`
  - SYCL: `SYCLDeviceUSMSpace`, `SYCLSharedUSMSpace`, `SYCLHostUSMSpace`
Execution spaces
- Propagate `index_type` from memory space to execution spaces (Cuda, HIP, HPX, OpenACC, OpenMP, SYCL, Serial, Threads).
Policies
- Default policy `index_type` is now `execution_space::index_type` instead of `execution_space::size_type`.
- Improved overflow handling in `RangePolicy` for signed index types.
Views and CRS
- `BasicView`: Use `mdspan_type::index_type` for `index_type`.
- `ViewTraits`: Add `index_type`.
- `Crs`: Use `index_type` in transpose and count/fill utilities.

### DIFF
--- a/core/src/Cuda/Kokkos_CudaSpace.hpp
+++ b/core/src/Cuda/Kokkos_CudaSpace.hpp
@@ -49,8 +49,8 @@ class CudaSpace {
   using memory_space    = CudaSpace;
   using execution_space = Kokkos::Cuda;
   using device_type     = Kokkos::Device<execution_space, memory_space>;
-
-  using size_type = unsigned int;
+  using size_type       = unsigned int;
+  using index_type      = std::make_signed_t<size_type>;
 
   /*--------------------------------*/
 
@@ -139,6 +139,7 @@ class CudaUVMSpace {
   using execution_space = Cuda;
   using device_type     = Kokkos::Device<execution_space, memory_space>;
   using size_type       = unsigned int;
+  using index_type      = std::make_signed_t<size_type>;
 
   /*--------------------------------*/
 
@@ -227,6 +228,7 @@ class CudaHostPinnedSpace {
   using memory_space    = CudaHostPinnedSpace;
   using device_type     = Kokkos::Device<execution_space, memory_space>;
   using size_type       = unsigned int;
+  using index_type      = std::make_signed_t<size_type>;
 
   /*--------------------------------*/
 
--- a/core/src/HIP/Kokkos_HIP_Space.hpp
+++ b/core/src/HIP/Kokkos_HIP_Space.hpp
@@ -39,8 +39,8 @@ class HIPSpace {
   using memory_space    = HIPSpace;
   using execution_space = HIP;
   using device_type     = Kokkos::Device<execution_space, memory_space>;
-
-  using size_type = unsigned int;
+  using size_type       = unsigned int;
+  using index_type      = std::make_signed_t<size_type>;
 
   /*--------------------------------*/
 
@@ -122,6 +122,7 @@ class HIPHostPinnedSpace {
   using memory_space    = HIPHostPinnedSpace;
   using device_type     = Kokkos::Device<execution_space, memory_space>;
   using size_type       = unsigned int;
+  using index_type      = std::make_signed_t<size_type>;
 
   /*--------------------------------*/
 
@@ -202,6 +203,7 @@ class HIPManagedSpace {
   using execution_space = HIP;
   using device_type     = Kokkos::Device<execution_space, memory_space>;
   using size_type       = unsigned int;
+  using index_type      = std::make_signed_t<size_type>;
 
   /*--------------------------------*/
 
--- a/core/src/Kokkos_AnonymousSpace.hpp
+++ b/core/src/Kokkos_AnonymousSpace.hpp
@@ -23,6 +23,7 @@ class AnonymousSpace {
   using memory_space    = AnonymousSpace;
   using execution_space = Kokkos::DefaultExecutionSpace;
   using size_type       = size_t;
+  using index_type      = std::make_signed_t<size_type>;
 
   //! This memory space preferred device_type
   using device_type = Kokkos::Device<execution_space, memory_space>;
--- a/core/src/Kokkos_ExecPolicy.hpp
+++ b/core/src/Kokkos_ExecPolicy.hpp
@@ -251,18 +251,18 @@ class ImplRangePolicy<ExecSpace, Properties...>
         if constexpr (std::is_signed_v<IndexType> !=
                       std::is_signed_v<member_type>) {
           // check signed to unsigned
-          if constexpr (std::is_signed_v<IndexType>) warn |= (bound < 0);
+          if constexpr (std::is_signed_v<IndexType>) error |= (bound < 0);
 
           // check unsigned to signed
           if constexpr (std::is_signed_v<member_type>) {
             // avoid overflow warnings by checking the size of the types e.g.
-            // conversion ‘long int’ to ‘int’ changes value
-            // from ‘9223372036854775807’ to ‘-1’
+            // conversion ‘unsigned int’ to ‘int’ changes value
+            // from ‘4294967295’ to ‘-1’ (when IndexType is narrower)
             if constexpr (sizeof(member_type) <= sizeof(IndexType))
               // safely cast member_type max to IndexType because member_type is
               // the same size or smaller.
-              warn |= (bound > static_cast<IndexType>(
-                                   std::numeric_limits<member_type>::max()));
+              error |= (bound > static_cast<IndexType>(
+                                    std::numeric_limits<member_type>::max()));
           }
         }
       }
--- a/core/src/Kokkos_ExecPolicy.hpp
+++ b/core/src/Kokkos_ExecPolicy.hpp
@@ -247,28 +247,25 @@ class ImplRangePolicy<ExecSpace, Properties...>
     if constexpr (std::is_convertible_v<member_type, IndexType>) {
       bool error = false;
 
-      if constexpr (std::is_arithmetic_v<member_type> &&
-                    (std::is_signed_v<IndexType> !=
-                     std::is_signed_v<member_type>)) {
-        // check signed to unsigned
-        if constexpr (std::is_signed_v<IndexType>)
-          error |= (bound < static_cast<IndexType>(
-                                std::numeric_limits<member_type>::min()));
+      if constexpr (std::is_arithmetic_v<member_type>) {
+        if constexpr (std::is_signed_v<IndexType> !=
+                      std::is_signed_v<member_type>) {
+          // check signed to unsigned
+          if constexpr (std::is_signed_v<IndexType>) warn |= (bound < 0);
 
-        // check unsigned to signed
-        if constexpr (std::is_signed_v<member_type>) {
-          // avoid overflow warnings by checking the size of the types e.g.
-          // conversion ‘long int’ to ‘int’ changes value
-          // from ‘9223372036854775807’ to ‘-1’
-          if constexpr (sizeof(member_type) <= sizeof(IndexType))
-            warn |= (bound > static_cast<IndexType>(
-                                 std::numeric_limits<member_type>::max()));
-          else {
-            warn |= (bound > std::numeric_limits<IndexType>::max());
+          // check unsigned to signed
+          if constexpr (std::is_signed_v<member_type>) {
+            // avoid overflow warnings by checking the size of the types e.g.
+            // conversion ‘long int’ to ‘int’ changes value
+            // from ‘9223372036854775807’ to ‘-1’
+            if constexpr (sizeof(member_type) <= sizeof(IndexType))
+              // safely cast member_type max to IndexType because member_type is
+              // the same size or smaller.
+              warn |= (bound > static_cast<IndexType>(
+                                   std::numeric_limits<member_type>::max()));
           }
         }
       }
-
       // check narrowing
       error |=
           (static_cast<IndexType>(static_cast<member_type>(bound)) != bound);
--- a/core/src/Kokkos_ExecPolicy.hpp
+++ b/core/src/Kokkos_ExecPolicy.hpp
@@ -240,42 +240,44 @@ class ImplRangePolicy<ExecSpace, Properties...>
     }
   }
 
-  // To be replaced with std::in_range (c++20)
+  // Arithmetic member_type and IndexType: signedness / numeric_limits checks.
+  // Always run the round-trip check below as well; for non-arithmetic IndexType
+  // only that line applies inside the is_convertible gate.
   template <typename IndexType>
   static void check_conversion_safety([[maybe_unused]] const IndexType bound) {
-    // Checking that the round-trip conversion preserves input index value
     if constexpr (std::is_convertible_v<member_type, IndexType>) {
       bool error = false;
 
-      if constexpr (std::is_arithmetic_v<member_type>) {
+      if constexpr (std::is_arithmetic_v<member_type> &&
+                    std::is_arithmetic_v<IndexType>) {
         if constexpr (std::is_signed_v<IndexType> !=
                       std::is_signed_v<member_type>) {
-          // check signed to unsigned
           if constexpr (std::is_signed_v<IndexType>) error |= (bound < 0);
-
-          // check unsigned to signed
           if constexpr (std::is_signed_v<member_type>) {
-            // avoid overflow warnings by checking the size of the types e.g.
-            // conversion ‘unsigned int’ to ‘int’ changes value
-            // from ‘4294967295’ to ‘-1’ (when IndexType is narrower)
             if constexpr (sizeof(member_type) <= sizeof(IndexType))
-              // safely cast member_type max to IndexType because member_type is
-              // the same size or smaller.
               error |= (bound > static_cast<IndexType>(
                                     std::numeric_limits<member_type>::max()));
+            else {
+              error |= (bound > std::numeric_limits<IndexType>::max());
+            }
           }
         }
       }
-      // check narrowing
+
       error |=
           (static_cast<IndexType>(static_cast<member_type>(bound)) != bound);
 
       if (error) {
+        std::string bound_to_text = []<typename T>(const T& b) {
+          if constexpr (std::is_arithmetic_v<T>)
+            return std::to_string(b);
+          else
+            return std::to_string(static_cast<member_type>(b));
+        }(bound);
         std::string msg =
             "Kokkos::RangePolicy bound type error: an unsafe implicit "
             "conversion is performed on a bound (" +
-            std::to_string(bound) +
-            "), which may not preserve its original value.\n";
+            bound_to_text + "), which may not preserve its original value.\n";
 
         Kokkos::abort(msg.c_str());
       }
--- a/core/src/Kokkos_ExecPolicy.hpp
+++ b/core/src/Kokkos_ExecPolicy.hpp
@@ -256,9 +256,17 @@ class ImplRangePolicy<ExecSpace, Properties...>
                                 std::numeric_limits<member_type>::min()));
 
         // check unsigned to signed
-        if constexpr (std::is_signed_v<member_type>)
-          error |= (bound > static_cast<IndexType>(
-                                std::numeric_limits<member_type>::max()));
+        if constexpr (std::is_signed_v<member_type>) {
+          // avoid overflow warnings by checking the size of the types e.g.
+          // conversion ‘long int’ to ‘int’ changes value
+          // from ‘9223372036854775807’ to ‘-1’
+          if constexpr (sizeof(member_type) <= sizeof(IndexType))
+            warn |= (bound > static_cast<IndexType>(
+                                 std::numeric_limits<member_type>::max()));
+          else {
+            warn |= (bound > std::numeric_limits<IndexType>::max());
+          }
+        }
       }
 
       // check narrowing
--- a/core/src/Kokkos_HostSpace.hpp
+++ b/core/src/Kokkos_HostSpace.hpp
@@ -38,6 +38,7 @@ class HostSpace {
   //! Tag this class as a kokkos memory space
   using memory_space = HostSpace;
   using size_type    = size_t;
+  using index_type   = std::make_signed_t<size_type>;
 
   /// \typedef execution_space
   /// \brief Default execution space for this memory space.
--- a/core/src/OpenACC/Kokkos_OpenACCSpace.hpp
+++ b/core/src/OpenACC/Kokkos_OpenACCSpace.hpp
@@ -26,7 +26,8 @@ class OpenACCSpace {
   using execution_space = OpenACC;
   using device_type     = Kokkos::Device<execution_space, memory_space>;
 
-  using size_type = size_t;
+  using size_type  = size_t;
+  using index_type = std::make_signed_t<size_type>;
 
   OpenACCSpace() = default;
 
--- a/core/src/OpenACC/Kokkos_OpenACC_ParallelReduce_Range.hpp
+++ b/core/src/OpenACC/Kokkos_OpenACC_ParallelReduce_Range.hpp
@@ -171,8 +171,8 @@ void OpenACCParallelReduceCustom(Schedule<Static>, Policy const& apolicy,
 KOKKOS_IMPL_ACC_PRAGMA(parallel loop gang num_gangs(n_chunks) num_workers(1) vector_length(chunk_size) KOKKOS_IMPL_OPENACC_VECTOR_RED_TEMP_CLAUSE1 present(functor, reducer, gang_red_temp) async(async_arg))
   /* clang-format on */
   for (IndexType team_id = 0; team_id < n_chunks; ++team_id) {
-    std::size_t tSize = static_cast<std::size_t>(chunk_size);
-    std::size_t tStep;
+    IndexType tSize = chunk_size;
+    IndexType tStep;
 #pragma acc loop vector
     for (IndexType thread_id = 0; thread_id < chunk_size; ++thread_id) {
       const IndexType local_offset = team_id * chunk_size + begin;
@@ -184,7 +184,6 @@ KOKKOS_IMPL_ACC_PRAGMA(parallel loop gang num_gangs(n_chunks) num_workers(1) vec
       }
       KOKKOS_IMPL_OPENACC_VECTOR_RED_TEMP_ACCESS1(thread_id) = temp;
     }
-#pragma acc loop seq
     for (tStep = (chunk_size >> 1); tStep > 0; tStep >>= 1) {
 #pragma acc loop vector
       for (IndexType thread_id = 0; thread_id < chunk_size; ++thread_id) {
@@ -210,20 +209,18 @@ KOKKOS_IMPL_ACC_PRAGMA(parallel loop gang num_gangs(n_chunks) num_workers(1) vec
 KOKKOS_IMPL_ACC_PRAGMA(parallel num_gangs(1) num_workers(1) vector_length(chunk_size) KOKKOS_IMPL_OPENACC_VECTOR_RED_TEMP_CLAUSE2 present(reducer, gang_red_temp) copyin(m_result_view) async(async_arg))
   /* clang-format on */
   {
-    std::size_t tSize = static_cast<std::size_t>(chunk_size);
-    std::size_t tStep;
+    IndexType tSize = chunk_size;
+    IndexType tStep;
 #pragma acc loop vector
     for (IndexType thread_id = 0; thread_id < chunk_size; ++thread_id) {
       IndexType idx;
       ValueType temp;
       temp = initVal;
-#pragma acc loop seq
       for (idx = thread_id; idx < n_chunks; idx += chunk_size) {
         reducer.join(&temp, &gang_red_temp(idx));
       }
       KOKKOS_IMPL_OPENACC_VECTOR_RED_TEMP_ACCESS2(thread_id) = temp;
     }
-#pragma acc loop seq
     for (tStep = (chunk_size >> 1); tStep > 0; tStep >>= 1) {
 #pragma acc loop vector
       for (IndexType thread_id = 0; thread_id < chunk_size; ++thread_id) {
--- a/core/src/OpenACC/Kokkos_OpenACC_ParallelReduce_Range.hpp
+++ b/core/src/OpenACC/Kokkos_OpenACC_ParallelReduce_Range.hpp
@@ -171,8 +171,8 @@ void OpenACCParallelReduceCustom(Schedule<Static>, Policy const& apolicy,
 KOKKOS_IMPL_ACC_PRAGMA(parallel loop gang num_gangs(n_chunks) num_workers(1) vector_length(chunk_size) KOKKOS_IMPL_OPENACC_VECTOR_RED_TEMP_CLAUSE1 present(functor, reducer, gang_red_temp) async(async_arg))
   /* clang-format on */
   for (IndexType team_id = 0; team_id < n_chunks; ++team_id) {
-    IndexType tSize = chunk_size;
-    IndexType tStep;
+    std::size_t tSize = static_cast<std::size_t>(chunk_size);
+    std::size_t tStep;
 #pragma acc loop vector
     for (IndexType thread_id = 0; thread_id < chunk_size; ++thread_id) {
       const IndexType local_offset = team_id * chunk_size + begin;
@@ -210,8 +210,8 @@ KOKKOS_IMPL_ACC_PRAGMA(parallel loop gang num_gangs(n_chunks) num_workers(1) vec
 KOKKOS_IMPL_ACC_PRAGMA(parallel num_gangs(1) num_workers(1) vector_length(chunk_size) KOKKOS_IMPL_OPENACC_VECTOR_RED_TEMP_CLAUSE2 present(reducer, gang_red_temp) copyin(m_result_view) async(async_arg))
   /* clang-format on */
   {
-    IndexType tSize = chunk_size;
-    IndexType tStep;
+    std::size_t tSize = static_cast<std::size_t>(chunk_size);
+    std::size_t tStep;
 #pragma acc loop vector
     for (IndexType thread_id = 0; thread_id < chunk_size; ++thread_id) {
       IndexType idx;
--- a/core/src/View/Kokkos_BasicView.hpp
+++ b/core/src/View/Kokkos_BasicView.hpp
@@ -134,15 +134,13 @@ class BasicView {
  public:
   using mdspan_type =
       mdspan<ElementType, Extents, LayoutPolicy, AccessorPolicy>;
-  using extents_type  = typename mdspan_type::extents_type;
-  using layout_type   = typename mdspan_type::layout_type;
-  using accessor_type = typename mdspan_type::accessor_type;
-  using mapping_type  = typename mdspan_type::mapping_type;
-  using element_type  = typename mdspan_type::element_type;
-  using value_type    = typename mdspan_type::value_type;
-  // FIXME: backwards compatibility, should be changed to the same as mdspan
-  // index_type
-  using index_type       = typename mdspan_type::size_type;
+  using extents_type     = typename mdspan_type::extents_type;
+  using layout_type      = typename mdspan_type::layout_type;
+  using accessor_type    = typename mdspan_type::accessor_type;
+  using mapping_type     = typename mdspan_type::mapping_type;
+  using element_type     = typename mdspan_type::element_type;
+  using value_type       = typename mdspan_type::value_type;
+  using index_type       = typename mdspan_type::index_type;
   using size_type        = typename mdspan_type::size_type;
   using rank_type        = typename mdspan_type::rank_type;
   using data_handle_type = typename mdspan_type::data_handle_type;
--- a/core/src/View/Kokkos_ViewAlloc.hpp
+++ b/core/src/View/Kokkos_ViewAlloc.hpp
@@ -40,12 +40,11 @@ struct ViewValueFunctor {
   template <class SameValueType = ValueType>
   KOKKOS_FUNCTION
       std::enable_if_t<std::is_default_constructible_v<SameValueType>>
-      operator()(ConstructTag, const ExecSpace::index_type i) const {
+      operator()(ConstructTag, const size_t i) const {
     new (ptr + i) ValueType();
   }
 
-  KOKKOS_FUNCTION void operator()(DestroyTag,
-                                  const ExecSpace::index_type i) const {
+  KOKKOS_FUNCTION void operator()(DestroyTag, const size_t i) const {
     // When instantiating a View on host execution space with a host only
     // destructor the workaround for CUDA device symbol instantiation tries to
     // still compile a destruction kernel for the device, and issues a warning
@@ -85,9 +84,11 @@ struct ViewValueFunctor {
 
   template <typename Tag>
   void parallel_for_implementation() {
-    using index_type = typename ExecSpace::index_type;
-    using PolicyType = Kokkos::RangePolicy<ExecSpace, index_type, Tag>;
-    PolicyType policy(space, index_type(0), index_type(n));
+    // Use a 64-bit index range: allocations can exceed what the execution
+    // space default index_type (e.g. 32-bit on CUDA/HIP) can represent.
+    using PolicyType =
+        Kokkos::RangePolicy<ExecSpace, Kokkos::IndexType<size_t>, Tag>;
+    PolicyType policy(space, size_t(0), n);
     uint64_t kpID = 0;
     if (Kokkos::Profiling::profileLibraryLoaded()) {
       const std::string functor_name =
--- a/core/src/View/Kokkos_ViewAlloc.hpp
+++ b/core/src/View/Kokkos_ViewAlloc.hpp
@@ -40,11 +40,12 @@ struct ViewValueFunctor {
   template <class SameValueType = ValueType>
   KOKKOS_FUNCTION
       std::enable_if_t<std::is_default_constructible_v<SameValueType>>
-      operator()(ConstructTag, const size_t i) const {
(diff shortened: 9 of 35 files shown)


janciesko:
Retest this please.

core/src/HIP/Kokkos_HIP_Space.hpp
```diff
@@ -211,7 +212,8 @@ class HIPManagedSpace {
   using memory_space    = HIPManagedSpace;
   using execution_space = HIP;
   using device_type     = Kokkos::Device<execution_space, memory_space>;
-  using size_type       = unsigned int;
+  using index_type      = unsigned int;
+  using size_type       = index_type;
```
dalg24: This one is wrong
```suggestion
  using size_type       = unsigned int;
  using index_type      = std::make_signed_t<size_type>;
```

core/src/Kokkos_AnonymousSpace.hpp
```diff
@@ -23,6 +23,7 @@ class AnonymousSpace {
   using memory_space    = AnonymousSpace;
   using execution_space = Kokkos::DefaultExecutionSpace;
   using size_type       = size_t;
+  using index_type      = size_type;
```
dalg24:
```suggestion
  using index_type      = std::make_signed_t<size_type>;
```

dalg24 (dismissed):
We agreed `index_type` was going to be signed, defined as `std::make_signed_t<size_type>;`

core/src/View/Kokkos_ViewTraits.hpp
```diff
@@ -536,7 +536,8 @@ struct ViewTraits {
   using host_mirror_space = HostMirrorSpace;
   using hooks_policy      = HooksPolicy;
 
-  using size_type = typename MemorySpace::size_type;
+  using size_type  = typename MemorySpace::size_type;
+  using index_type = typename MemorySpace::index_type;
```
dalg24: Why do we want it in the View traits?
Why is it part of this PR?

core/src/OpenACC/Kokkos_OpenACCSpace.hpp
```diff
@@ -26,7 +26,8 @@ class OpenACCSpace {
   using execution_space = OpenACC;
   using device_type     = Kokkos::Device<execution_space, memory_space>;
 
-  using size_type = size_t;
+  using size_type  = size_t;
+  using index_type = size_type;
```
dalg24:
```suggestion
  using index_type = std::make_signed_t<size_type>;
```

core/src/Kokkos_HostSpace.hpp
```diff
@@ -38,6 +38,7 @@ class HostSpace {
   //! Tag this class as a kokkos memory space
   using memory_space = HostSpace;
   using size_type    = size_t;
+  using index_type   = size_type;
```
dalg24:
```suggestion
  using index_type   = std::make_signed_t<size_type>;
```

core/src/View/Kokkos_ViewTraits.hpp
cwpearson: it might be surprising for `size_type` to exist and `index_type` to not exist if they're side-by-side everywhere else.

cwpearson (commented):
Looks like we refer to `size_type` in the docs as what we now mean by `index_type`?

https://github.com/kokkos/kokkos-core-wiki/blob/08d3bfd041455efa4154290a578a707f13e15ae9/docs/source/API/core/view/view.rst?plain=1#L191-L193

(that's not the only place, e.g. DynRankView)

masterleinad:
> Looks like we refer to `size_type` in the docs as what we now mean by `index_type`?

Yes, the idea is to replace `size_type` with `index_type` and possibly deprecate `size_type`.

core/src/View/Kokkos_ViewTraits.hpp
janciesko: We don't have a use for it, @dalg24 is right, rolled back change.

janciesko:
@masterleinad, I've added two separate tests but simplified the `check_conversion_safety`.

masterleinad:
> @masterleinad, I've added two separate tests but simplified the `check_conversion_safety`.

This is failing on pretty much every single CI build. Does the tests pass for you locally?

janciesko:
Fixing

core/src/impl/Kokkos_HostSpace_deepcopy.cpp
```diff
@@ -128,8 +128,7 @@ template <typename ExecutionSpace>
 void hostspace_parallel_zeromemset(const ExecutionSpace& exec, void* dst,
                                    size_t n) {
   constexpr uint8_t z_u8 = 0x00;
-  using policy_t =
-      Kokkos::RangePolicy<ExecutionSpace, Kokkos::IndexType<size_t>>;
+  using policy_t         = Kokkos::RangePolicy<ExecutionSpace>;
```
dalg24: This would break if we changed the definition of `HostSpace::index_type`.
Any particular reason to touch this code?

core/src/View/Kokkos_BasicView.hpp
```diff
@@ -136,15 +136,13 @@ class BasicView {
...
-  // FIXME: backwards compatibility, should be changed to the same as mdspan
-  // index_type
-  using index_type       = typename mdspan_type::size_type;
+  using extents_type     = typename mdspan_type::extents_type;
+  using layout_type      = typename mdspan_type::layout_type;
+  using accessor_type    = typename mdspan_type::accessor_type;
+  using mapping_type     = typename mdspan_type::mapping_type;
+  using element_type     = typename mdspan_type::element_type;
+  using value_type       = typename mdspan_type::value_type;
+  using index_type       = typename mdspan_type::index_type;
```
dalg24, line 143: Why does that change belong to this PR?

core/src/impl/Kokkos_HostSpace_deepcopy.cpp
```diff
@@ -128,8 +128,7 @@ template <typename ExecutionSpace>
 void hostspace_parallel_zeromemset(const ExecutionSpace& exec, void* dst,
                                    size_t n) {
   constexpr uint8_t z_u8 = 0x00;
-  using policy_t =
-      Kokkos::RangePolicy<ExecutionSpace, Kokkos::IndexType<size_t>>;
```
masterleinad, line 132: Why did you change?

core/src/traits/Kokkos_IndexTypeTrait.hpp
```diff
@@ -47,7 +47,7 @@ struct IndexTypePolicyMixin<Kokkos::IndexType<IntegralIndexType>,
   static constexpr auto show_index_type_error_in_compilation_message =
       show_extra_index_type_erroneously_given_to_execution_policy<
           std::conditional_t<base_t::index_type_is_defaulted, void,
-                             typename base_t::schedule_type>>{};
```
masterleinad, line 50: That was just wrong it seems.

core/src/View/Kokkos_ViewAlloc.hpp
```diff
@@ -84,9 +87,9 @@ struct ViewValueFunctor {
 
   template <typename Tag>
   void parallel_for_implementation() {
-    using PolicyType =
-        Kokkos::RangePolicy<ExecSpace, Kokkos::IndexType<int64_t>, Tag>;
-    PolicyType policy(space, 0, n);
+    using index_type = typename ExecSpace::index_type;
+    using PolicyType = Kokkos::RangePolicy<ExecSpace, index_type, Tag>;
+    PolicyType policy(space, index_type(0), index_type(n));
```
dalg24: The original code was explicitly requesting 64-bit integers.
Why are we changing it to something that is backend-specific?
Isn't going to fall over say with a 9GB allocation of some 4-byte non-trivially constructible user-defined type (e.g. `n > INT_MAX` on CUDA or HIP).

core/src/View/Kokkos_ViewAlloc.hpp
```diff
@@ -40,11 +40,12 @@ struct ViewValueFunctor {
   template <class SameValueType = ValueType>
   KOKKOS_FUNCTION
       std::enable_if_t<std::is_default_constructible_v<SameValueType>>
-      operator()(ConstructTag, const size_t i) const {
+      operator()(ConstructTag, const ExecSpace::index_type i) const {
```
dalg24: It looks like we were explicitly requesting 64-bit integers to index into that array we are allocating.

core/src/View/Kokkos_ViewTraits.hpp
dalg24: I am confused.  What was rolled back?  Why is this change still part of this PR?

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -280,9 +280,17 @@ class RangePolicy : public Impl::PolicyTraits<Properties...> {
...
+          // avoid overflow warnings by checking the size of the types e.g.
+          // conversion ‘long int’ to ‘int’ changes value
+          // from ‘9223372036854775807’ to ‘-1’
+          if constexpr (sizeof(member_type) <= sizeof(IndexType))
+            warn |= (bound > static_cast<IndexType>(
+                                 std::numeric_limits<member_type>::max()));
+          else {
+            warn |= (bound > std::numeric_limits<IndexType>::max());
+          }
+        }
```
dalg24, line 267: Was there a gap in our RangePolicy constructor testing?

core/src/traits/Kokkos_IndexTypeTrait.hpp
```diff
@@ -47,7 +47,7 @@ struct IndexTypePolicyMixin<Kokkos::IndexType<IntegralIndexType>,
...
+                             typename base_t::index_type>>{};
```
dalg24, line 50: So if I understand correctly, what was wrong was the error message we would produce, if specifying twice the integer type to use with the policy (via template arguments)
https://godbolt.org/z/raGMTsxrd

core/src/impl/Kokkos_HostSpace_deepcopy.cpp
janciesko: It would, hence the index type is part of the exec space. We cannot arbitrarily change that.

dalg24: I don't understand the answer.  I am saying was there a particular harm in being explicit?  Is it better to have that code work implicitly because the default index type is (hopefully) going to be large enough.

core/src/Kokkos_ExecPolicy.hpp
janciesko: See comment in-line

core/src/impl/Kokkos_HostSpace_deepcopy.cpp
janciesko: "Being explicit" and "index type is (hopefully) going to be large enough" are good considerations but I think they could be addressed when defining the default `index_type` (which could be size_t) for that exec space.

core/src/View/Kokkos_ViewAlloc.hpp
janciesko: For consistency we should probably use index type here.

core/src/View/Kokkos_ViewAlloc.hpp
janciesko: This is an interesting case. Looking for more input if we should hard-code such loops to use 64 bits indices or infer from index_type.

core/src/impl/Kokkos_HostSpace_deepcopy.cpp
masterleinad: No, it's not. We explicitly want the index type here tp be `size_t` since that's the maximum range of the allocation. `IndexType` neesd to be kernel-specific (with a default as `ExecutionSpace::index_type`).

core/src/traits/Kokkos_IndexTypeTrait.hpp
janciesko: I think so

core/src/View/Kokkos_ViewAlloc.hpp
masterleinad: You want to use the `index_type` of the policy here but since that isn't available we need tp be explicit and use `size_t`.

core/src/impl/Kokkos_HostSpace_deepcopy.cpp
janciesko: aah, I mixed up the words. Yes, `size_type` here which is potentially larger than index_type. Good point, thanks Daniel.

core/src/Kokkos_ExecPolicy.hpp
dalg24: I can see the comment, I am asking why we did not realize this before and whever we need to update `core/unit_test/TestRangePolicyConstructors.hpp`

core/src/impl/Kokkos_HostSpace_deepcopy.cpp
janciesko: Probably not even `size_type` but `size_t`.

core/src/View/Kokkos_ViewTraits.hpp
janciesko: ~~Forgot about this and put it back. Reverted.~~

`[‎core/src/Kokkos_Crs.hpp]` seems to have a dependency on `traits::index_type` now.

core/src/Kokkos_ExecPolicy.hpp
janciesko: We did not realize this before as our tests tested for positive (or zero) indices only. Testing with negative indices produces those warnings now. The new tests are in `TestRangePolicyConstructors.hpp`.

core/src/traits/Kokkos_IndexTypeTrait.hpp
masterleinad: Could also be in a separate pull request.

core/src/View/Kokkos_ViewAlloc.hpp
janciesko: Agreed, reverted, thanks for the rational.

core/src/View/Kokkos_ViewAlloc.hpp
janciesko: Same rational applies here I think, thanks for pointing it out Damien. Still we likely want to use `size_t` as the index type.

masterleinad: Generally, you should probaby leave `size_t` alone (if it's not the alias declaration `using size_type = size_t`).

core/src/View/Kokkos_BasicView.hpp
janciesko: It seems reasonable to 
`using index_type = typename mdspan_type::size_type;` -> 
`using index_type = typename mdspan_type::index_type;`

core/src/Cuda/Kokkos_CudaSpace.hpp
```diff
@@ -49,8 +49,8 @@ class CudaSpace {
   using memory_space    = CudaSpace;
   using execution_space = Kokkos::Cuda;
   using device_type     = Kokkos::Device<execution_space, memory_space>;
-
-  using size_type = unsigned int;
+  using size_type       = unsigned int;
+  using index_type      = std::make_signed_t<size_type>;
```
crtrott, line 53: Should we turn the logic around? mdspan defines index_type and then makes size_type the unsigned version. I do agree that we should have a default index type for the execution spaces which is signed, in order to support negative indices.

crtrott: Note this is a question, not a mandate to change. Ending up with `unsigned int` / `int` respectively for our GPU spaces and the 64bit variants for host is fine with me.

core/src/View/Kokkos_BasicView.hpp
crtrott: I think this is fine, right now it's anyway hardcoded to size_t and it falls into the cleanup of index_type being consistently used.

core/src/View/Kokkos_ViewAlloc.hpp
crtrott: It's not yet reverted is it? And the rational for using 64bit indexing in all these guys is that we are instantiating these kernels to loop over the entries of a View during initialization and destruction. This thus needs to cover the range of the index_type we are using in the View, which currently is actually 64bit. And it needs to be to cover more than 2B elements. Since we are not passing the View in here lets leave these alone for now, and rework this when we actually can change the View index_type, thus we could know here at compile time that we wouldn't have more than 2B iterations.

core/src/View/Kokkos_ViewAlloc.hpp
crtrott: Same needs to be reverted since we may have allocations of more than 2B elements.

core/src/View/Kokkos_ViewLegacy.hpp
```diff
@@ -323,7 +323,7 @@ class View : public ViewTraits<DataType, Properties...> {
   // using mapping_type -> not applicable
   using element_type = typename traits::value_type;
   // using value_type -> conflicts with traits::value_type
-  using index_type = typename traits::memory_space::size_type;
+  using index_type = typename traits::index_type;
```
crtrott: Let's leave this guy alone.

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -248,20 +248,25 @@ class RangePolicy : public Impl::PolicyTraits<Properties...> {
...
+      if constexpr (std::is_arithmetic_v<member_type>) {
+        if constexpr (std::is_signed_v<IndexType> !=
+                      std::is_signed_v<member_type>) {
+          // check signed to unsigned
+          if constexpr (std::is_signed_v<IndexType>) warn |= (bound < 0);
+
+          // check unsigned to signed
+          if constexpr (std::is_signed_v<member_type>) {
+            // avoid overflow warnings by checking the size of the types e.g.
+            // conversion ‘long int’ to ‘int’ changes value
```
crtrott:
```suggestion
            // conversion ‘unsigned int’ to ‘int’ changes value
```

crtrott: We are in the IndexType is unsigned case right now.

crtrott (changes requested):
Couple comments to be addressed. But there is also one bigger one: we are changing the sign of RangePolicy begin/end. This might be a bit worrisome, and will pop up warnings in downstream code everywhere. On the other hand I am also feeling really uncomfortable with negative begin values not working by default. Thoughts?

core/src/View/Kokkos_ViewAlloc.hpp
janciesko: Agreed. Reverted in: https://github.com/kokkos/kokkos/pull/7441/changes/53b132f5df4492a1456115bc5ea74121c1734ccb

core/unit_test/TestRangePolicyConstructors.hpp
```diff
@@ -169,12 +195,16 @@ TEST(TEST_CATEGORY_DEATH, range_policy_implicitly_converted_bounds) {
...
   }
   {
-    long long test_val = std::numeric_limits<long long>::max();
-    ASSERT_DEATH({ (void)IntPolicy(0LL, test_val); },
+    unsigned long long test_val =
+        std::numeric_limits<unsigned long long>::max();
+    ASSERT_DEATH({ (void)IntPolicy(0, test_val); },
+                 get_error_msg(expected, test_val));
+
+    ASSERT_DEATH({ (void)DefaultPolicy(0, test_val); },
```
masterleinad: `DefaultPolicy` isn't defined.

masterleinad:
> But there is also one bigger one: we are changing the sign of RangePolicy begin/end. This might be a bit worrisome, and will pop up warnings in downstream code everywhere. On the other hand I am also feeling really uncomfortable with negative begin values not working by default. Thoughts?

Since `index_type` didn't exist before, I don't think (`-Wsign-compare`-type) warnings are a concern downstream generally. The change that appears in user programs is assgnment between values of different signedness but I'm not aware that the major compilers generally warn about that. Compiling with `-Wsign-conversion` in `Kokkos` would already give a lot of extra warnings.

For  the `SYCL` backend, we had used `size_type=int` and that triggered warnings in user codes whenever `ExecutionSpace::size_type` was explicitly used for a loop variable (or similar). IIn https://github.com/kokkos/kokkos/pull/8542, we changed that type to be unsigned for consistency with other backends but I haven't heard anyone complaining about more warnings.

core/src/impl/Kokkos_HostSpace_deepcopy.cpp
```diff
@@ -147,9 +147,10 @@ void hostspace_parallel_zeromemset(const ExecutionSpace& exec, void* dst,
   // Zero-fill with 8-byte words in parallel
   uint64_t* dst_p  = reinterpret_cast<uint64_t*>(dst_c);
   const size_t cnt = (n - count) / 8;
+  using index_t    = typename ExecutionSpace::index_type;
   Kokkos::parallel_for("Kokkos::Impl::hostspace_parallel_zeromemset",
                        policy_t(exec, 0, cnt),
-                       [=](const size_t i) { dst_p[i] = z_u64; });
+                       [=](const index_t i) { dst_p[i] = z_u64; });
```
masterleinad: This still wrong. There are no changes necessary in this file.

core/src/View/Kokkos_ViewAlloc.hpp
```diff
@@ -84,9 +84,11 @@ struct ViewValueFunctor {
 
   template <typename Tag>
   void parallel_for_implementation() {
+    // Use a 64-bit index range: allocations can exceed what the execution
+    // space default index_type (e.g. 32-bit on CUDA/HIP) can represent.
     using PolicyType =
-        Kokkos::RangePolicy<ExecSpace, Kokkos::IndexType<int64_t>, Tag>;
-    PolicyType policy(space, 0, n);
+        Kokkos::RangePolicy<ExecSpace, Kokkos::IndexType<size_t>, Tag>;
+    PolicyType policy(space, size_t(0), n);
```
masterleinad, line 91: Why did you change?

core/unit_test/view/TestBasicView.hpp
```diff
@@ -18,7 +18,9 @@ namespace {
 template <class ExecutionSpace, class Extents>
 auto make_spanning_mdrange_policy_from_extents_impl(const Extents &extents,
                                                     std::index_sequence<0>) {
-  return Kokkos::RangePolicy<ExecutionSpace>{0, extents.extent(0)};
+  using index_type = typename Kokkos::RangePolicy<ExecutionSpace>::index_type;
+  return Kokkos::RangePolicy<ExecutionSpace>{
+      0, static_cast<index_type>(extents.extent(0))};
```
masterleinad: Why did you change?

masterleinad (dismissed):

core/src/impl/Kokkos_HostSpace_deepcopy.cpp
janciesko: Oversight probably, thanks! Rolled back.

core/unit_test/view/TestBasicView.hpp
janciesko: This is unnecessary. Rolled back.

core/src/View/Kokkos_ViewAlloc.hpp
janciesko: This seems like an oversight in the original code.

masterleinad: I'd be suprised if you can create an allocation larger than `std::numeric_limits<std::int64_t>::max()`. 🙂 

Note that `size_t` isn't necessarily a 64-bit type so the new comment seems to apply better to the old code. If you keep the change here, I would remove the comment.

core/src/Cuda/Kokkos_CudaSpace.hpp
dalg24: You mean this
```suggestion
  using index_type      = int;
  using size_type       = std::make_unsigned_t<index_type>;
```

Fine by me.

core/src/View/Kokkos_ViewTraits.hpp
dalg24: Why are the Crs changes part of this PR in the first place?

core/src/View/Kokkos_ViewAlloc.hpp
dalg24: I agree with Daniel, I don't see any good reason for this change.
On second thought I don't really care to revert.
I don't find the comment particularly useful.

core/unit_test/TestTeamPolicyCTAD.cpp
```diff
@@ -19,6 +19,7 @@ struct TestTeamPolicyCTAD {
   struct SomeExecutionSpace {
     using execution_space = SomeExecutionSpace;
     using size_type       = size_t;
+    using index_type      = std::make_signed_t<size_type>;
```
dalg24:
```suggestion
    using index_type      = int;
```

core/unit_test/TestRangePolicyCTAD.cpp
```diff
@@ -15,6 +15,7 @@ struct TestRangePolicyCTAD {
   struct SomeExecutionSpace {
     using execution_space = SomeExecutionSpace;
     using size_type       = size_t;
+    using index_type      = std::make_signed_t<size_type>;
```
dalg24:
```suggestion
    using index_type      = int;
```

core/unit_test/TestMDRangePolicyCTAD.cpp
```diff
@@ -21,6 +21,7 @@ struct TestMDRangePolicyCTAD {
   struct SomeExecutionSpace {
     using execution_space = SomeExecutionSpace;
     using size_type       = size_t;
+    using index_type      = std::make_signed_t<size_type>;
```
dalg24:
```suggestion
    using index_type      = int;
```

core/unit_test/TestRange.hpp
```diff
@@ -145,8 +180,15 @@ struct TestRange {
...
-                                                                N + offset),
+        Kokkos::RangePolicy<ExecSpace, ScheduleType, PosOffsetTag>(offset,
+                                                                   N + offset),
+        *this, total);
+    // sum( 1 .. N )
+    ASSERT_EQ(size_t((N) * (N + 1) / 2), size_t(total));
+
+    Kokkos::parallel_reduce(
+        Kokkos::RangePolicy<ExecSpace, ScheduleType, NegOffsetTag>(
+            (-1) * offset, N - offset),
```
dalg24:
```suggestion
            -offset, N - offset),
```

core/unit_test/TestRangePolicyConstructors.hpp
```diff
@@ -18,9 +18,10 @@ import kokkos.core_impl;
 namespace {
 
 TEST(TEST_CATEGORY, range_policy_runtime_parameters) {
-  using Policy     = Kokkos::RangePolicy<>;
-  using Index      = Policy::index_type;
-  Index work_begin = 5;
+  using Policy = Kokkos::RangePolicy<>;
+  using Index  = Policy::index_type;
+
+  Index work_begin = -5;
```
dalg24, line 24: Why change this unit test?

dalg24 (commented):
Don't forget to also update the NextSilicon backend

core/unit_test/TestTeamPolicyCTAD.cpp
```diff
@@ -19,6 +19,7 @@ struct TestTeamPolicyCTAD {
   struct SomeExecutionSpace {
     using execution_space = SomeExecutionSpace;
     using size_type       = size_t;
```
dalg24:
```suggestion
```

core/unit_test/TestRangePolicyCTAD.cpp
```diff
@@ -15,6 +15,7 @@ struct TestRangePolicyCTAD {
   struct SomeExecutionSpace {
     using execution_space = SomeExecutionSpace;
     using size_type       = size_t;
```
dalg24:
```suggestion
```

core/unit_test/TestMDRangePolicyCTAD.cpp
```diff
@@ -21,6 +21,7 @@ struct TestMDRangePolicyCTAD {
   struct SomeExecutionSpace {
     using execution_space = SomeExecutionSpace;
     using size_type       = size_t;
```
dalg24:
```suggestion
```

core/unit_test/TestRangePolicyConstructors.hpp
janciesko: I think we can reuse this test, just extending the ranges.

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -247,20 +247,25 @@ class ImplRangePolicy<ExecSpace, Properties...>
...
+                      std::is_signed_v<member_type>) {
+          // check signed to unsigned
+          if constexpr (std::is_signed_v<IndexType>) error |= (bound < 0);
+
+          // check unsigned to signed
+          if constexpr (std::is_signed_v<member_type>) {
+            // avoid overflow warnings by checking the size of the types e.g.
+            // conversion ‘unsigned int’ to ‘int’ changes value
+            // from ‘4294967295’ to ‘-1’ (when IndexType is narrower)
+            if constexpr (sizeof(member_type) <= sizeof(IndexType))
```
ldh4, line 261: If this produced an overflow warning when `IndexType` is narrower than `member_type`, wouldn't there be a same warning produced from L270, which does the same casting check?

Also, I'm not sure if relying on sizeof is safe, since IndexType could be any user defined type (could be implicitly convertible to and from `member_type` but may not necessarily an arithmetic type)

core/src/View/Kokkos_ViewTraits.hpp
dalg24: Ping

core/src/Kokkos_ExecPolicy.hpp
ldh4: I may be little confused on the provided comments,
but because of this condition in the check:
```
        if constexpr (std::is_signed_v<IndexType> !=
                      std::is_signed_v<member_type>) {
```
this portion of check is for when signedness of `IndexType` and `member_type` don't match (except when IndexType is a user-defined type)
The unit tests you added in `TestRangePolicyConstructors.hpp` seems to be only signed (`long long`) to signed (`int`) conversion.

core/src/Kokkos_ExecPolicy.hpp
janciesko: The `sizeof` split is only trying to avoid the specific diagnostic error you get from  
`bound > static_cast<IndexType>(std::numeric_limits<member_type>::max())` when `IndexType` is narrower than `member_type` (e.g. `member_type = unsigned long long`, `IndexType = int`): the cast of `max()` to a narrower signed type is what triggers the “value changes on conversion” style warning. 

The conversion is safe given we're in the `if constexpr (std::is_arithmetic_v<member_type>)`  branch.

core/src/Kokkos_ExecPolicy.hpp
janciesko: Yes, the `long long` + `RangePolicy<int>` death tests look like “signed to signed”; note that we're testing the round-trip / narrowing path there. We test for the signedness mismatch between the bound type and the policy index type in the `range_policy_implicitly_converted_bounds` tests.

core/src/View/Kokkos_ViewTraits.hpp
janciesko: Hey Damien, CRS uses ViewTraits and RangePolicy with an explicit index type; when the default policy index becomes execution_space::index_type, the CRS helpers should probably use that too.  It is for consistency with the index-type rollout.

dalg24: I do not think that we should handle Crs as part of this PR.
Please withdraw these changes.

core/unit_test/TestRangePolicyConstructors.hpp
```diff
@@ -100,11 +124,13 @@ TEST(TEST_CATEGORY_DEATH, range_policy_round_trip_conversion_fires) {
...
-      "Kokkos::RangePolicy bound type error: an unsafe implicit conversion is "
-      "performed";
-  ASSERT_DEATH((void)Policy(0, W(&n)), msg);
+  // FIXME: The following ASSERT_DEATH was disabled because the round-trip
+  // conversion check does not trigger with the current check_conversion_safety
+  // logic. Re-enable once the expected failure is understood and implemented.
+  // [[maybe_unused]] std::string msg =
+  //     "Kokkos::RangePolicy bound type error: an unsafe implicit conversion is
+  //     " "performed";
+  // ASSERT_DEATH((void)Policy(0, W(&n)), msg);
```
ldh4, line 133: I would uncomment this and disable the test by prepending the test name with `DISABLED` (`DISABLED_range_policy_round_trip_conversion_fires`).
Or, place an extra `long` var in the struct `W`, and it will trigger the expected abort.

core/src/Kokkos_ExecPolicy.hpp
ldh4: I get that this is to avoid the overflow warning and that adding this additional constraint doesn't alter the upper bound check.
However, I have a tangential concern on the usage of `sizeof`. This condition is fine when `IndexType` is an arithmetic type but not when `IndexType` is a UDT padded with extra variables.

For instance, in the test that was disabled in this PR `range_policy_round_trip_conversion_fires`, padding the struct `W` with an extra `long` will satisfy the condition `(sizeof(member_type) <= sizeof(IndexType))` even though the actual used index type is `int`. So `sizeof` seems unreliable here when `IndexType` isn't arithmetic.

janciesko: Aah, yes, that's right, I have not thought that through. The entire block should be replaceable with `std::in_range`.

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -240,37 +240,35 @@ class ImplRangePolicy<ExecSpace, Properties...>
...
+  // round-trip line applies.
   template <typename IndexType>
   static void check_conversion_safety([[maybe_unused]] const IndexType bound) {
-    // Checking that the round-trip conversion preserves input index value
     if constexpr (std::is_convertible_v<member_type, IndexType>) {
       bool error = false;
 
+      // For arithmetic member_type and IndexType, use std::in_range,
+      // use the round-trip check below for UDT
       if constexpr (std::is_arithmetic_v<member_type> &&
```
janciesko, line 253: @ldh4 How's this?

ldh4: Yeah maybe this is how it'll have to be, bypassing this check if `IndexType` is a UDT. I wasn't sure if there's a way to keep the bounds check correct while considering all cases of UDTs.

As for `std::in_range`, I believe it only works for integral types. We could definitely check with `std::in_range` if both `member_type` and `IndexType` are integrals. Otherwise, maybe fallback to the previous implementation (given that both types are still arithmetic types)?

janciesko: `std::in_range` does not buy us anything. Reverted.

ldh4 (approved):

core/src/OpenACC/Kokkos_OpenACCSpace.hpp
```diff
@@ -26,7 +26,8 @@ class OpenACCSpace {
   using execution_space = OpenACC;
   using device_type     = Kokkos::Device<execution_space, memory_space>;
 
-  using size_type = size_t;
+  using size_type  = size_t;
+  using index_type = std::make_signed_t<size_type>;
```
nmm0, line 30: I think this is okay, but worth mentioning that this will likely be a signed 64 bit int instead of the other device execution spaces which are signed 32 bit ints

core/src/View/Kokkos_ViewAlloc.hpp
```diff
@@ -84,9 +84,11 @@ struct ViewValueFunctor {
 
   template <typename Tag>
   void parallel_for_implementation() {
+    // Use a 64-bit index range: allocations can exceed what the execution
+    // space default index_type (e.g. 32-bit on CUDA/HIP) can represent.
     using PolicyType =
-        Kokkos::RangePolicy<ExecSpace, Kokkos::IndexType<int64_t>, Tag>;
-    PolicyType policy(space, 0, n);
+        Kokkos::RangePolicy<ExecSpace, Kokkos::IndexType<size_t>, Tag>;
```
nmm0, line 90: This may or may not be 64 bit, I'd rather not driveby change this

core/src/Kokkos_ExecPolicy.hpp
```diff
@@ -240,37 +240,44 @@ class ImplRangePolicy<ExecSpace, Properties...>
...
-        // check signed to unsigned
-        if constexpr (std::is_signed_v<IndexType>)
-          error |= (bound < static_cast<IndexType>(
-                                std::numeric_limits<member_type>::min()));
-
-        // check unsigned to signed
-        if constexpr (std::is_signed_v<member_type>)
-          error |= (bound > static_cast<IndexType>(
-                                std::numeric_limits<member_type>::max()));
+                    std::is_arithmetic_v<IndexType>) {
```
nmm0, line 252: If I understand this correctly, and this doesn't have to work on device we may be able to simplify the representability part with https://en.cppreference.com/cpp/utility/in_range

nmm0 (commented):
First pass, I will take another look tomorrow

core/src/View/Kokkos_ViewAlloc.hpp
crtrott: there is a conversion thing otherwise going on, because the functor called already was using size_t.

core/src/View/Kokkos_ViewAlloc.hpp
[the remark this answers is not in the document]
crtrott: its fine, only on 32bit machines would it not be 64bit but then we don't need it either.

crtrott (approved):

crtrott:
Retest this please.

core/src/OpenACC/Kokkos_OpenACCSpace.hpp
janciesko: The reduction seems to be broken

janciesko: @nmm0 Should be fixed.

core/src/OpenACC/Kokkos_OpenACC_ParallelReduce_Range.hpp
```diff
@@ -171,8 +171,8 @@ void OpenACCParallelReduceCustom(Schedule<Static>, Policy const& apolicy,
 KOKKOS_IMPL_ACC_PRAGMA(parallel loop gang num_gangs(n_chunks) num_workers(1) vector_length(chunk_size) KOKKOS_IMPL_OPENACC_VECTOR_RED_TEMP_CLAUSE1 present(functor, reducer, gang_red_temp) async(async_arg))
   /* clang-format on */
   for (IndexType team_id = 0; team_id < n_chunks; ++team_id) {
-    IndexType tSize = chunk_size;
-    IndexType tStep;
+    std::size_t tSize = static_cast<std::size_t>(chunk_size);
+    std::size_t tStep;
```
masterleinad: What's going on here?

janciesko: For OpenACC, the reduction was producing wrong results in our unit tests with the defaulted `index_type` which now is a `signed` type. 

It seems that what the compiler produces for those binary operations in that reduction is wrong. Given that `chunk_size` is always positive, I am not sure which particular binary op causes this.

Casting to an unsigned type resolved this issue. This applies only to the OpenACC backend.

masterleinad: @seyonglee Can you provide more insight into this?

seyonglee: I don't know why but sometimes explicit `acc loop seq` directive causes incorrect code generation; deleting them will fix the error.

core/src/View/Kokkos_BasicView.hpp
dalg24: Again I do not understand why this change belong here.  Please revert it and propose it as a follow up PR

janciesko:
I think this is ready to go. @masterleinad, @seyonglee could you finalize the review?

core/src/OpenACC/Kokkos_OpenACC_ParallelReduce_Range.hpp
janciesko: This seems to have fixed it. Great to see this collaboration. Thanks again @seyonglee @masterleinad

seyonglee (approved):

PR #7451, opened 2024-10-15
Ensure RandomAccessIterator fulfills the std::random_access_iterator concept

masterleinad:
Addressing https://github.com/kokkos/kokkos/pull/7304#discussion_r1747883125. 
~~This pull request also makes sure that we can construct a `RandomAccessIterator` from a `const`-qualified `View`.~~

### DIFF
--- a/algorithms/src/std_algorithms/impl/Kokkos_RandomAccessIterator.hpp
+++ b/algorithms/src/std_algorithms/impl/Kokkos_RandomAccessIterator.hpp
@@ -164,6 +164,10 @@ class RandomAccessIterator<::Kokkos::View<DataType, Args...>> {
     return it;
   }
 
+  friend iterator_type operator+(difference_type n, iterator_type other) {
+    return other + n;
+  }
+
   KOKKOS_FUNCTION
   iterator_type operator-(difference_type n) const {
     auto it = *this;
--- a/algorithms/unit_tests/TestStdAlgorithmsCompileOnly.cpp
+++ b/algorithms/unit_tests/TestStdAlgorithmsCompileOnly.cpp
@@ -507,6 +507,20 @@ struct TestStruct {
   }
 };
 
+#ifndef KOKKOS_ENABLE_CXX17
+template <typename ViewType>
+constexpr bool
+test_kokkos_iterator_satify_std_random_access_iterator_concept() {
+  return std::random_access_iterator<
+      Kokkos::Experimental::Impl::RandomAccessIterator<ViewType>>;
+}
+
+static_assert(test_kokkos_iterator_satify_std_random_access_iterator_concept<
+              Kokkos::View<int *>>());
+static_assert(test_kokkos_iterator_satify_std_random_access_iterator_concept<
+              Kokkos::View<const int *>>());
+#endif
+
 }  // namespace compileonly
 }  // namespace stdalgos
 }  // namespace Test


algorithms/src/std_algorithms/impl/Kokkos_Constraints.hpp
```diff
@@ -24,7 +24,7 @@ namespace Kokkos {
 namespace Experimental {
 namespace Impl {
 
-template <class T>
+template <class T, class Enable = void>
```
masterleinad, line 27: Relevant change

algorithms/src/std_algorithms/impl/Kokkos_RandomAccessIterator.hpp
```diff
@@ -26,13 +26,14 @@ namespace Kokkos {
...
 class RandomAccessIterator;
 
-template <class DataType, class... Args>
-class RandomAccessIterator< ::Kokkos::View<DataType, Args...> > {
+template <class ViewType>
+class RandomAccessIterator<ViewType,
+                           std::enable_if_t<Kokkos::is_view_v<ViewType>>> {
  public:
-  using view_type     = ::Kokkos::View<DataType, Args...>;
+  using view_type     = ViewType;
```
masterleinad, line 36: Relevant change.

algorithms/src/std_algorithms/impl/Kokkos_RandomAccessIterator.hpp
```diff
@@ -111,80 +119,94 @@ class RandomAccessIterator< ::Kokkos::View<DataType, Args...> > {
...
-  reference operator*() const { return m_view(m_current_index); }
+  pointer data() const { return m_data; }
 
   KOKKOS_FUNCTION
-  view_type view() const { return m_view; }
+  int stride() const { return m_stride; }
+
+  friend iterator_type operator+(difference_type n, iterator_type other) {
+    return other + n;
+  }
```
masterleinad, line 197: Relevant change.

algorithms/unit_tests/TestStdAlgorithmsCompileOnly.cpp
```diff
@@ -110,6 +110,21 @@ struct TestStruct {
...
+  static_assert(std::random_access_iterator<
+                Kokkos::Experimental::Impl::RandomAccessIterator<
+                    Kokkos::View<const T *>>>);
+  static_assert(std::random_access_iterator<
+                Kokkos::Experimental::Impl::RandomAccessIterator<
+                    const Kokkos::View<T *>>>);
+  static_assert(std::random_access_iterator<
+                Kokkos::Experimental::Impl::RandomAccessIterator<
+                    const Kokkos::View<const T *>>>);
+#endif
```
masterleinad, line 126: Relevant change.

.clang-tidy
```diff
@@ -1,3 +1,3 @@
-Checks: '-*,kokkos-*,modernize-type-traits,modernize-use-using,modernize-use-nullptr,cppcoreguidelines-pro-type-cstyle-cast'
+Checks: '-*,kokkos-*,modernize-type-traits,modernize-use-using,modernize-use-nullptr,cppcoreguidelines-pro-type-cstyle-cast,bugprone-*,-bugprone-macro-parentheses,-bugprone-reserved-identifier,-bugprone-easily-swappable-parameters,-bugprone-unhandled-self-assignment'
```
crtrott, line 1: Uhm I don't think you should change the clang-tidy checks here,

algorithms/src/sorting/Kokkos_SortPublicAPI.hpp
```diff
@@ -53,13 +53,9 @@ void sort(const ExecutionSpace& exec,
...
-    if (view.span_is_contiguous()) {
-      std::sort(view.data(), view.data() + view.size());
-    } else {
-      auto first = ::Kokkos::Experimental::begin(view);
-      auto last  = ::Kokkos::Experimental::end(view);
-      std::sort(first, last);
-    }
+    auto first = ::Kokkos::Experimental::begin(view);
+    auto last  = ::Kokkos::Experimental::end(view);
+    std::sort(first, last);
```
crtrott, line 58: Not sure this is a good idea. I wouldn't be at ALL surprised if std::sort with pointers is faster/specialized than with arbitrary iterators.

crtrott (changes requested):
This change makes RandomAccessIterator very different from before: it will not work for non-strided layouts anymore and it will not reference count. We definitely can' just do this without a larger discussion ...

masterleinad:
> This change makes RandomAccessIterator very different from before: it will not work for non-strided layouts anymore and it will not reference count. We definitely can' just do this without a larger discussion ...

I marked the relevant parts of the pull request.

crtrott (changes requested):
Since we only specialize RandomAccessIterator for View, how about we don't do enable_if and just take a "ViewType" template argument and then static assert that it is a View? I would like to avoid dummy template arguments and enable ifs if there is not good reason to have them. We can always come back and change that if we have a reason to do specializations. What are other folks thoughts? Also this is impl only so we can do what we want.

algorithms/src/std_algorithms/impl/Kokkos_RandomAccessIterator.hpp
```diff
@@ -164,6 +163,10 @@ class RandomAccessIterator<::Kokkos::View<DataType, Args...>> {
     return it;
   }
 
+  friend iterator_type operator+(difference_type n, iterator_type other) {
+    return other + n;
+  }
```
masterleinad, line 168: This is the relevant change to make the class fulfill the std::random_access_iterator concept. Thhe rest of the changes are for allowing `const View`s.

algorithms/unit_tests/TestStdAlgorithmsCompileOnly.cpp
dalg24, line 126: Book keeping: Create you own constexpr templated function that returns true, call it something like `test_kokkos_iterator_satify_std_random_access_iterator_concept()`, and static assert at namespace scope

dalg24 (changes requested):

algorithms/unit_tests/TestStdAlgorithmsCompileOnly.cpp
masterleinad: Here you go: c82e30e401dc34ccbd786bb51913d00ed921ae34

algorithms/unit_tests/TestStdAlgorithmsCompileOnly.cpp
```diff
@@ -507,6 +507,24 @@ struct TestStruct {
   }
 };
 
+#ifndef KOKKOS_ENABLE_CXX17
+template <typename ViewType>
+constexpr bool
+test_kokkos_iterator_satify_std_random_access_iterator_concept() {
+  return std::random_access_iterator<
```
dalg24, line 514: Where do we get the `<concepts>` header from?  Is it guaranteed that it is always transitively included?

masterleinad: According to https://en.cppreference.com/w/cpp/iterator/random_access_iterator, `random_access_iterator` comes from `<iterator>` and we include that header in `Kokkos_RandomAccessIterator.hpp` which is included in `Kokkos_StdAlgorithms.hpp` that we include at the top of the file. 
In short, `RandomAccessIterator` needs the `iterator` header and so `random_access_iterator` is available whenever a definition of `RandomAccessIterator` is needed.

algorithms/unit_tests/TestStdAlgorithmsCompileOnly.cpp
```diff
@@ -507,6 +507,24 @@ struct TestStruct {
...
+}
+
+static_assert(test_kokkos_iterator_satify_std_random_access_iterator_concept<
+              Kokkos::View<int *>>());
+static_assert(test_kokkos_iterator_satify_std_random_access_iterator_concept<
+              Kokkos::View<const int *>>());
+static_assert(test_kokkos_iterator_satify_std_random_access_iterator_concept<
+              const Kokkos::View<int *>>());
+static_assert(test_kokkos_iterator_satify_std_random_access_iterator_concept<
+              const Kokkos::View<const int *>>());
```
dalg24, line 525: Why do we do const-qualified views as template params?

masterleinad: It just looked like a gap in support.

dalg24: How so? It does look like a conscious decision when we did implement it as a specialization.
I fail to see what use case this supports.

The way I understand it is that we said we have `[c]{begin, end}` free functions that take a Kokkos view and return some sort of random iterator whose type is not specified.
Ensuring that the standard library random iterator concept is modeled by that this undisclosed type looks worthwhile to me.
But I am reluctant to approve that change that lets the view template argument have cv-qualifiers.  If anything I like the status quo better.

masterleinad: Then let's just postpone `const` support (possibly until we decide to move the class out of `Impl`).

masterleinad: Here you go: https://github.com/kokkos/kokkos/pull/7451/commits/e1d3d2d2dfd5a320931abc47d222928d2851cb00

dalg24 (approved):

algorithms/src/std_algorithms/impl/Kokkos_RandomAccessIterator.hpp
dalg24: (other changes were dropped)

crtrott (approved):

aprokop (approved):

PR #7447, opened 2024-10-15
tests: enable random shuffling on a subset of the CI

romintomasetti:
Tackling #7435.

PR #4811, opened 2022-02-21
Test with nvcc -Werror

masterleinad:
Follow-up to #4735. Since we now have the ability to test for `nvcc`'s `-Werror`, we should also use it in our CI which is what this pull request does.

PR #7021, opened 2024-05-21
Cuda: Fix nvcc warnings

masterleinad:
Fixes #6991. Currently, we only turn warnings into errors. We should incorporate all the fixes for the warnings here as well, though.

PR #4642, opened 2022-01-04
Suppress CUDA RDC warning about stack size

masterleinad:
Fixes https://github.com/kokkos/kokkos/issues/2039. It's still debatable if we rather have downstream code deal with it or if we only want to apply it in the test suite instead.

PR #7455, opened 2024-10-16
jenkins: fixing cxx flags that needed quoting

romintomasetti:
Spotted while working on #7447.

The bug has been introduced in #4811.

### DIFF
--- a/.jenkins
+++ b/.jenkins
@@ -147,7 +147,7 @@ pipeline {
                                 -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
                                 -DCMAKE_CXX_COMPILER=$WORKSPACE/bin/nvcc_wrapper \
                                 -DCMAKE_CXX_FLAGS="-Werror --Werror=all-warnings -Xcudafe --diag_suppress=3159 -Xcudafe --diag_suppress=940" \
-                                -DCMAKE_EXE_LINKER_FLAG="-Xnvlink -suppress-stack-size-warning" \
+                                -DCMAKE_EXE_LINKER_FLAGS="-Xnvlink -suppress-stack-size-warning" \
                                 -DCMAKE_CXX_STANDARD=17 \
                                 -DKokkos_INSTALL_TESTING=ON \
                               .. && \
--- a/.jenkins
+++ b/.jenkins
@@ -146,7 +146,8 @@ pipeline {
                                 -DCMAKE_BUILD_TYPE=Release \
                                 -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
                                 -DCMAKE_CXX_COMPILER=$WORKSPACE/bin/nvcc_wrapper \
-                                -DCMAKE_CXX_FLAGS=-Werror --Werror=all-warnings -Xcudafe --diag_suppress=3159 \
+                                -DCMAKE_CXX_FLAGS="-Werror --Werror=all-warnings -Xcudafe --diag_suppress=3159 -Xcudafe --diag_suppress=940" \
+                                -DCMAKE_EXE_LINKER_FLAG="-Xnvlink -suppress-stack-size-warning" \
                                 -DCMAKE_CXX_STANDARD=17 \
                                 -DKokkos_INSTALL_TESTING=ON \
                               .. && \
--- a/core/unit_test/TestComplex.hpp
+++ b/core/unit_test/TestComplex.hpp
@@ -25,7 +25,7 @@
 #else
 #ifdef __CUDA_ARCH__
 #pragma diagnostic push
-#pragma diag_suppress 20208
+#pragma diag_suppress 3245
 #endif
 #endif
 #endif
--- a/core/unit_test/TestViewRank.cpp
+++ b/core/unit_test/TestViewRank.cpp
@@ -33,9 +33,7 @@ constexpr bool test_view_rank_and_dynamic_rank() {
   static_assert(!std::is_integral_v<decltype(rank)>);
   auto rank_preferred = View::rank();  // since 4.0.01
   static_assert(std::is_same_v<decltype(rank_preferred), size_t>);
-  (void)rank;
-  (void)rank_preferred;
-  return true;
+  return rank == rank_preferred;
 }
 
 // clang-format off
--- a/core/unit_test/TestNumericTraits.hpp
+++ b/core/unit_test/TestNumericTraits.hpp
@@ -29,7 +29,7 @@
 #else
 #ifdef __CUDA_ARCH__
 #pragma diagnostic push
-#pragma diag_suppress 20208
+#pragma diag_suppress 3245
 #endif
 #endif
 #endif
--- a/core/unit_test/TestRangePolicyConstructors.hpp
+++ b/core/unit_test/TestRangePolicyConstructors.hpp
@@ -234,7 +234,7 @@ TEST(TEST_CATEGORY_DEATH, range_policy_implicitly_converted_bounds) {
       "Kokkos::RangePolicy bound type error: an unsafe implicit conversion is "
       "performed on a bound (), which may not preserve its original value.\n";
 
-  auto get_error_msg = [](auto str, auto val) {
+  [[maybe_unused]] auto get_error_msg = [](auto str, auto val) {
     return str.insert(str.find("(") + 1, std::to_string(val).c_str());
   };
 #ifndef KOKKOS_ENABLE_DEPRECATED_CODE_4


core/unit_test/TestComplex.hpp
```diff
@@ -25,7 +25,13 @@
...
 #ifdef __CUDA_ARCH__
 #pragma diagnostic push
-#pragma diag_suppress 20208
+// This line triggers "invalid error number" on Jenkins job "CUDA-11.0-NVCC-RDC".
+// But commenting it triggers the error "'long double' is treated as 'double' in device code".
+// It seems that in https://github.com/kokkos/kokkos/pull/7021/files, the fix
+// was to add at the command line.
+// Here, I just tried to add "-Xcudafe --display_error_number" to get the actual error code and got:
+//  Error #3245-D: 'long double' is treated as 'double' in device code
+#pragma diag_suppress 3245
```
romintomasetti, line 34: @masterleinad I think you worked on this in #7021. Could you please review / test on your side ?

romintomasetti: Haven't checked yet but could it be that error codes are not the same if using `nv_diag_suppress` *vs* `diag_suppress` ?

containers/src/Kokkos_DualView.hpp
```diff
@@ -441,8 +441,12 @@ class DualView : public ViewTraits<DataType, Properties...> {
...
       }
     }
-#ifdef KOKKOS_COMPILER_INTEL
-    __builtin_unreachable();
+    // Got:
+    //  error #940-D: missing return statement at end of non-void function
+    // so I added '__builtin_unreachable' and now I get the calling host
+    // function from host device function.
+#if defined(KOKKOS_COMPILER_NVCC) && KOKKOS_COMPILER_NVCC == 1100
+    Kokkos::abort("This should never happen.");
```
romintomasetti, line 449: Decide what to do here.

.jenkins
```diff
@@ -148,7 +148,7 @@ pipeline {
                                 -DCMAKE_BUILD_TYPE=Release \
                                 -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
                                 -DCMAKE_CXX_COMPILER=$WORKSPACE/bin/nvcc_wrapper \
-                                -DCMAKE_CXX_FLAGS=-Werror --Werror=all-warnings -Xcudafe --diag_suppress=3159 \
+                                -DCMAKE_CXX_FLAGS="-Werror --Werror=all-warnings -Xcudafe --diag_suppress=3159 -Xcudafe --diag_suppress=940" \
```
romintomasetti, line 151: I've added the `diag_suppress=940` because otherwise we get to fix:
> error #940-D: missing return statement at end of non-void function
in `Kokkos::DualView::view` method.

But it is not "trivial", because this method is `__host__ __device__`, and the usual `__builtin_unreachable` trick is `__host__`.

My first attempt was then to add `Kokkos::abort("This never happens.");` but I'm a bit reluctant adding this.

I find ignoring the `940` error code a good compromise.

core/unit_test/CMakeLists.txt
```diff
@@ -759,6 +759,15 @@ if(Kokkos_ENABLE_CUDA)
     ${Cuda_SOURCES_SHAREDSPACE}
   )
 
+  # Suppress warning:
+  #   "Stack size for entry function 'xxx' cannot be statically determined"
+  # See https://github.com/kokkos/kokkos/issues/2039.
+  if(KOKKOS_ENABLE_CUDA_RELOCATABLE_DEVICE_CODE)
+    target_link_options(Kokkos_CoreUnitTest_Cuda3
```
romintomasetti, line 766: @dalg24 Do we want this ?

containers/src/Kokkos_DualView.hpp
romintomasetti: I decided to use `diag_suppress=940` at CLI level to avoid this (see comment below).

.jenkins
```diff
@@ -148,7 +148,7 @@ pipeline {
                                 -DCMAKE_BUILD_TYPE=Release \
                                 -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
                                 -DCMAKE_CXX_COMPILER=$WORKSPACE/bin/nvcc_wrapper \
-                                -DCMAKE_CXX_FLAGS=-Werror --Werror=all-warnings -Xcudafe --diag_suppress=3159 \
+                                -DCMAKE_CXX_FLAGS="-Werror --Werror=all-warnings -Xcudafe --diag_suppress=3159" \
```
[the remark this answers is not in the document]
romintomasetti: My gut feeling is that since this is one of the jobs that run on the first stage of the `Jenkins` CI, it's not a very good idea to disable device "warning as error" and defer potential issues to a later stage.

algorithms/unit_tests/TestStdAlgorithmsCommon.hpp
```diff
@@ -199,7 +199,7 @@ auto create_deep_copyable_compatible_view_with_same_extent(ViewType view) {
   // this is needed for intel to avoid
   // error #1011: missing return statement at end of non-void function
 #if defined KOKKOS_COMPILER_INTEL ||                                  \
-    (defined(KOKKOS_COMPILER_NVCC) && KOKKOS_COMPILER_NVCC >= 1130 && \
+    (defined(KOKKOS_COMPILER_NVCC) && KOKKOS_COMPILER_NVCC >= 1100 && \
```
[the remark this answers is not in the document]
romintomasetti: I guess where I changed this, it's host code so we get `__builtin_unreachable` from the host compiler (`g++-8` in this case).

The link you mention is when `nvcc` introduces `__builtin_unreachable` for *device code*, right ?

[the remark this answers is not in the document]
romintomasetti: So, my first attempt was to solve the errors by making changes.

Then came this issue that `__builtin_unreachable` is not available for device code for `nvcc < 11.3`. So I was stuck.

So I added `--diag_suppress=940` for this only place where I was stuck (because I couldn't use `__builtin_unreachable` in a device function).

So you're right that because of this `diag_suppress`, I should be able to revert other changes. Let's do it and see.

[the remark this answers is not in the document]
romintomasetti: Seems fine.

.jenkins
romintomasetti: See discussion in https://github.com/kokkos/kokkos/pull/7455#discussion_r1804591172.

masterleinad (dismissed):
Fine with me if it's passing CI.

core/unit_test/TestComplex.hpp
dalg24: ping (both of you)

core/unit_test/CMakeLists.txt
```diff
@@ -759,6 +759,13 @@ if(Kokkos_ENABLE_CUDA)
     ${Cuda_SOURCES_SHAREDSPACE}
   )
 
+  # Suppress warning:
+  #   "Stack size for entry function 'xxx' cannot be statically determined"
+  # See https://github.com/kokkos/kokkos/issues/2039.
+  if(KOKKOS_ENABLE_CUDA_RELOCATABLE_DEVICE_CODE)
+    target_link_options(Kokkos_CoreUnitTest_Cuda3 PRIVATE -Xnvlink -suppress-stack-size-warning)
+  endif()
```
dalg24, line 767: Was that warning actually turned into an error?
If so, I am leaning towards using CMAKE_EXE_LINKER_FLAG instead.

Not that this had been proposed and rejected in the past https://github.com/kokkos/kokkos/pull/4642#issuecomment-1006021381

core/unit_test/TestViewRank.cpp
```diff
@@ -28,13 +28,12 @@ constexpr bool test_view_rank_and_dynamic_rank() {
...
+  [[maybe_unused]] auto rank =
+      View::rank;  // not an integral type in contrast to Kokkos version
   // less than 4.0.01
   static_assert(!std::is_integral_v<decltype(rank)>);
-  auto rank_preferred = View::rank();  // since 4.0.01
+  [[maybe_unused]] auto rank_preferred = View::rank();  // since 4.0.01
   static_assert(std::is_same_v<decltype(rank_preferred), size_t>);
-  (void)rank;
-  (void)rank_preferred;
   return true;
```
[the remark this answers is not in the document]
romintomasetti: Much better indeed.

core/unit_test/CMakeLists.txt
romintomasetti: It generated a `nvlink` **error**. OK to pass it in the `.jenkins` through `-DCMAKE_EXE_LINKER_FLAG="-Xnvlink -suppress-stack-size-warning"` but beware it will be applied to *all* targets.

core/unit_test/TestViewRank.cpp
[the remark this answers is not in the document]
romintomasetti: Done.

core/unit_test/TestViewRank.cpp
```diff
@@ -28,13 +28,12 @@ constexpr bool test_view_rank_and_dynamic_rank() {
   static_assert(std::is_same_v<decltype(View::rank()), size_t>);
   static_assert(std::is_convertible_v<decltype(View::rank_dynamic), size_t>);
   static_assert(std::is_same_v<decltype(View::rank_dynamic()), size_t>);
-  auto rank = View::rank;  // not an integral type in contrast to Kokkos version
+  [[maybe_unused]] auto rank =
+      View::rank;  // not an integral type in contrast to Kokkos version
   // less than 4.0.01
```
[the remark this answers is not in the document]
romintomasetti: With your suggestion from https://github.com/kokkos/kokkos/pull/7455#discussion_r1805343051 I could revert the changes. Closing.

core/unit_test/TestComplex.hpp
romintomasetti: I just cross-checked in clean `Docker` containers and got:
```c++
#include "cuda_runtime.h"

/**
 * Compiling with:
 *      ./bin/nvcc_wrapper -Werror test.cpp
 *
 * With @c nvcc 12.6 this generates
 *      test.cpp(5): warning #20208-D: 'long double' is treated as 'double' in device code
 * and can be silenced therefore with
 *      #pragma nv_diag_suppress 20208
 *
 * With @c nvcc 11.0, we need
 *      #pragma diag_suppress 3245
*/
#ifdef __NVCC_DIAG_PRAGMA_SUPPORT__
#pragma nv_diagnostic push
#pragma nv_diag_suppress 20208
#else
#ifdef __CUDA_ARCH__
#pragma diagnostic push
#pragma diag_suppress 3245
#endif
#endif

__device__ void test()
{
    [[maybe_unused]]long double var = 0.5;
}

int main() {}
```

romintomasetti: So indeed the error codes are different.

core/unit_test/CMakeLists.txt
romintomasetti: Seems not, see https://github.com/kokkos/kokkos/pull/7455#discussion_r1805362482. Closing this comment in favor of the mentioned one.

core/unit_test/TestComplex.hpp
dalg24: What does that mean?  Do we need both?

romintomasetti: It means that to suppress this "long double treated as double in device code" warning, we need 20208 when `__NVCC_DIAG_PRAGMA_SUPPORT__` and 3245 otherwise. More precisely you need 20208 if you use `nv_diag_suppress*` and 3245 if using `diag_suppress`.

core/unit_test/CMakeLists.txt
romintomasetti: Quoting from a failed job that does not have this `nvlink` option:
```bash
nvlink error   : Stack size for entry function '_ZN6Kokkos4Impl72_GLOBAL__N__48_tmpxft_00001f73_00000000_7_TestCuda_Task_cpp1_ii_6ea0a04923cuda_task_queue_executeINS_18BasicTaskSchedulerINS_4CudaENS0_9TaskQueueIS4_NS_12CudaUVMSpaceEEEEEEEvT_i' cannot be statically determined
```

core/unit_test/TestComplex.hpp
dalg24: OK I get it now.  Hadn't expanded the code above.

.jenkins
```diff
@@ -146,7 +146,8 @@ pipeline {
                                 -DCMAKE_BUILD_TYPE=Release \
                                 -DCMAKE_CXX_COMPILER_LAUNCHER=ccache \
                                 -DCMAKE_CXX_COMPILER=$WORKSPACE/bin/nvcc_wrapper \
-                                -DCMAKE_CXX_FLAGS=-Werror --Werror=all-warnings -Xcudafe --diag_suppress=3159 \
+                                -DCMAKE_CXX_FLAGS="-Werror --Werror=all-warnings -Xcudafe --diag_suppress=3159 -Xcudafe --diag_suppress=940" \
+                                -DCMAKE_EXE_LINKER_FLAG="-Xnvlink -suppress-stack-size-warning" \
```
dalg24, line 150:
```suggestion
                                -DCMAKE_EXE_LINKER_FLAGS="-Xnvlink -suppress-stack-size-warning" \
```

dalg24 (approved):
Thanks, this is clearly an improvement over the status quo, the GitLab warnings can be handled later

crtrott (approved):

PR #7456, opened 2024-10-16
Allow using SequentialHostInit with Kokkos::DualView

### LINKED ISSUE #7373: `SequentialHostInit` does not compile in a `DualView`
**Describe the bug**

`SequentialHostInit` does not compile in a `DualView`:

`k_view.resize(Kokkos::view_alloc(Kokkos::SequentialHostInit),n);`

```
../../lib/kokkos/core/src/View/Kokkos_ViewAlloc.hpp(321): error: static assertion failed
static_assert(SpaceAccessibility<HostSpace, MemSpace>::accessible);
```

It seems like the Device view inside the DualView should ignore the `SequentialHostInit` property so it can compile.

masterleinad:
Fixes #7373. @stanmoore1 Can you confirm that this works for you?

This fixes using `SequentialHostInit` with `DualView` for the constructor and `resize`. `realloc` doesn't work properly with `View` of `Views` anyway so I decided to leave it alone for this pull request.

### DIFF
--- a/containers/src/Kokkos_DualView.hpp
+++ b/containers/src/Kokkos_DualView.hpp
@@ -942,12 +942,21 @@ class DualView : public ViewTraits<DataType, Properties...> {
         Impl::size_mismatch(h_view, h_view.rank_dynamic, new_extents);
 
     if (sizeMismatch) {
-      ::Kokkos::realloc(arg_prop, d_view, n0, n1, n2, n3, n4, n5, n6, n7);
-      if constexpr (alloc_prop_input::initialize) {
-        h_view = create_mirror_view(typename t_host::memory_space(), d_view);
+      if constexpr (alloc_prop_input::sequential_host_init) {
+        static_assert(alloc_prop_input::initialize,
+                      "DualView: SequentialHostInit isn't compatible with "
+                      "WithoutInitializing!");
+        ::Kokkos::realloc(arg_prop, h_view, n0, n1, n2, n3, n4, n5, n6, n7);
+        d_view =
+            create_mirror_view_and_copy(typename t_dev::memory_space(), h_view);
       } else {
-        h_view = create_mirror_view(Kokkos::WithoutInitializing,
-                                    typename t_host::memory_space(), d_view);
+        ::Kokkos::realloc(arg_prop, d_view, n0, n1, n2, n3, n4, n5, n6, n7);
+        if constexpr (alloc_prop_input::initialize) {
+          h_view = create_mirror_view(typename t_host::memory_space(), d_view);
+        } else {
+          h_view = create_mirror_view(Kokkos::WithoutInitializing,
+                                      typename t_host::memory_space(), d_view);
+        }
       }
     } else if constexpr (alloc_prop_input::initialize) {
       if constexpr (alloc_prop_input::has_execution_space) {
--- a/containers/src/Kokkos_DualView.hpp
+++ b/containers/src/Kokkos_DualView.hpp
@@ -1072,10 +1072,8 @@ class DualView : public ViewTraits<DataType, Properties...> {
       if (sizeMismatch) {
         sync<typename t_host::memory_space>();
         ::Kokkos::resize(arg_prop, h_view, n0, n1, n2, n3, n4, n5, n6, n7);
-        resync_device(Kokkos::view_alloc(Kokkos::WithoutInitializing));
-
-        /* Mark Host copy as modified */
-        ++modified_flags(0);
+        d_view =
+            create_mirror_view_and_copy(typename t_dev::memory_space(), h_view);
       }
       return;
     } else if constexpr (alloc_prop_input::has_execution_space) {
--- a/containers/src/Kokkos_DualView.hpp
+++ b/containers/src/Kokkos_DualView.hpp
@@ -279,11 +279,11 @@ class DualView : public ViewTraits<DataType, Properties...> {
     if constexpr (Impl::ViewCtorProp<P...>::sequential_host_init) {
       h_view = t_host(arg_prop, n0, n1, n2, n3, n4, n5, n6, n7);
       static_assert(Impl::ViewCtorProp<P...>::initialize,
-                    "DualView: Kokkos::SequentialHostInit isn't compatible "
-                    "with Kokkos::WithoutInitializing!");
+                    "DualView: SequentialHostInit isn't compatible with "
+                    "WithoutInitializing!");
       static_assert(!Impl::ViewCtorProp<P...>::has_execution_space,
-                    "DualView: Kokkos::SequentialHostInit isn't compatible "
-                    "with providing an execution space instance!");
+                    "DualView: SequentialHostInit isn't compatible with "
+                    "providing an execution space instance!");
 
       d_view = Kokkos::create_mirror_view_and_copy(
           typename traits::memory_space{}, h_view);
@@ -1063,11 +1063,11 @@ class DualView : public ViewTraits<DataType, Properties...> {
 
     if constexpr (alloc_prop_input::sequential_host_init) {
       static_assert(alloc_prop_input::initialize,
-                    "DualView: Kokkos::SequentialHostInit isn't compatible "
-                    "with Kokkos::WithoutInitializing!");
+                    "DualView: SequentialHostInit isn't compatible with "
+                    "WithoutInitializing!");
       static_assert(!alloc_prop_input::has_execution_space,
-                    "DualView: Kokkos::SequentialHostInit isn't compatible "
-                    "with providing an execution space instance!");
+                    "DualView: SequentialHostInit isn't compatible with "
+                    "providing an execution space instance!");
 
       if (sizeMismatch) {
         sync<typename t_host::memory_space>();
--- a/containers/src/Kokkos_DualView.hpp
+++ b/containers/src/Kokkos_DualView.hpp
@@ -275,14 +275,29 @@ class DualView : public ViewTraits<DataType, Properties...> {
            const size_t n5                   = KOKKOS_IMPL_CTOR_DEFAULT_ARG,
            const size_t n6                   = KOKKOS_IMPL_CTOR_DEFAULT_ARG,
            const size_t n7                   = KOKKOS_IMPL_CTOR_DEFAULT_ARG)
-      : modified_flags(t_modified_flags("DualView::modified_flags")),
-        d_view(arg_prop, n0, n1, n2, n3, n4, n5, n6, n7) {
-    // without UVM, host View mirrors
-    if constexpr (Kokkos::Impl::has_type<Impl::WithoutInitializing_t,
-                                         P...>::value)
-      h_view = Kokkos::create_mirror_view(Kokkos::WithoutInitializing, d_view);
-    else
-      h_view = Kokkos::create_mirror_view(d_view);
+      : modified_flags(t_modified_flags("DualView::modified_flags")) {
+    if constexpr (Impl::ViewCtorProp<P...>::sequential_host_init) {
+      h_view = t_host(arg_prop, n0, n1, n2, n3, n4, n5, n6, n7);
+      static_assert(Impl::ViewCtorProp<P...>::initialize,
+                    "DualView: Kokkos::SequentialHostInit isn't compatible "
+                    "with Kokkos::WithoutInitializing!");
+      static_assert(!Impl::ViewCtorProp<P...>::has_execution_space,
+                    "DualView: Kokkos::SequentialHostInit isn't compatible "
+                    "with providing an execution space instance!");
+
+      d_view =
+          Kokkos::create_mirror_view(typename traits::memory_space{}, h_view);
+    } else {
+      d_view = t_dev(arg_prop, n0, n1, n2, n3, n4, n5, n6, n7);
+
+      // without UVM, host View mirrors
+      if constexpr (Kokkos::Impl::has_type<Impl::WithoutInitializing_t,
+                                           P...>::value)
+        h_view =
+            Kokkos::create_mirror_view(Kokkos::WithoutInitializing, d_view);
+      else
+        h_view = Kokkos::create_mirror_view(d_view);
+    }
   }
 
   //! Copy constructor (shallow copy)
@@ -1046,9 +1061,11 @@ class DualView : public ViewTraits<DataType, Properties...> {
       }
     };
 
-    constexpr bool has_execution_space = alloc_prop_input::has_execution_space;
-
-    if constexpr (has_execution_space) {
+    if constexpr (alloc_prop_input::sequential_host_init) {
+      sync<typename t_host::memory_space>();
+      resize_on_host(arg_prop);
+      return;
+    } else if constexpr (alloc_prop_input::has_execution_space) {
       using ExecSpace = typename alloc_prop_input::execution_space;
       const auto& exec_space =
           Impl::get_property<Impl::ExecutionSpaceTag>(arg_prop);
--- a/containers/src/Kokkos_DualView.hpp
+++ b/containers/src/Kokkos_DualView.hpp
@@ -285,8 +285,8 @@ class DualView : public ViewTraits<DataType, Properties...> {
                     "DualView: Kokkos::SequentialHostInit isn't compatible "
                     "with providing an execution space instance!");
 
-      d_view =
-          Kokkos::create_mirror_view(typename traits::memory_space{}, h_view);
+      d_view = Kokkos::create_mirror_view_and_copy(
+          typename traits::memory_space{}, h_view);
     } else {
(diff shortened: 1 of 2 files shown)


containers/unit_tests/TestDualView.hpp
```diff
@@ -622,6 +622,16 @@ TEST(TEST_CATEGORY,
             const_dv.view<Kokkos::DefaultExecutionSpace>());
 }
 
+TEST(TEST_CATEGORY, dualview_sequential_host_init) {
+  Kokkos::DualView<double*> dv(
+      Kokkos::view_alloc("myView", Kokkos::SequentialHostInit), 100u);
```
maartenarnst, line 627: In our cuda build, this line doesn't compile. 

Because the constructor of `DualView` forwards the properties to the constructor of the device view:

- https://github.com/kokkos/kokkos/blob/3fe17d96d42eb2bd84f1f2a835a2ed0ba87606b6/containers/src/Kokkos_DualView.hpp#L279

And `SequentialHostInit` cannot be used for a device view:

- https://github.com/kokkos/kokkos/blob/3fe17d96d42eb2bd84f1f2a835a2ed0ba87606b6/core/src/View/Kokkos_ViewAlloc.hpp#L202-L206

containers/unit_tests/TestDualView.hpp
```diff
@@ -622,6 +622,16 @@ TEST(TEST_CATEGORY,
             const_dv.view<Kokkos::DefaultExecutionSpace>());
 }
 
+TEST(TEST_CATEGORY, dualview_sequential_host_init) {
```
maartenarnst, line 625: It would probably make sense to consider in the test not only a case with a type like `int` but also a case with a type with a non-trivial default constructor. E.g. a "dual view of views". To make sure that we trigger checks that the destructor is not called in a parallel region.

containers/src/Kokkos_DualView.hpp
```diff
@@ -928,11 +928,26 @@ class DualView : public ViewTraits<DataType, Properties...> {
...
+        if constexpr (alloc_prop_input::initialize) {
+          h_view = create_mirror_view(
+              Kokkos::view_alloc(typename t_host::memory_space(),
+                                 Kokkos::SequentialHostInit),
+              d_view);
+        } else {
+          h_view = create_mirror_view(
+              Kokkos::view_alloc(Kokkos::WithoutInitializing,
+                                 typename t_host::memory_space(),
+                                 Kokkos::SequentialHostInit),
```
maartenarnst, line 941: There is a `static_assert` in the constructor of `ViewCtorProp` that seems to prevent passing both `WithoutInitializing` and `SequentialHostInit`:

- https://github.com/kokkos/kokkos/blob/3fe17d96d42eb2bd84f1f2a835a2ed0ba87606b6/core/src/View/Kokkos_ViewCtor.hpp#L208-L210

containers/src/Kokkos_DualView.hpp
```diff
@@ -275,14 +275,29 @@ class DualView : public ViewTraits<DataType, Properties...> {
...
-      : modified_flags(t_modified_flags("DualView::modified_flags")),
-        d_view(arg_prop, n0, n1, n2, n3, n4, n5, n6, n7) {
-    // without UVM, host View mirrors
-    if constexpr (Kokkos::Impl::has_type<Impl::WithoutInitializing_t,
-                                         P...>::value)
-      h_view = Kokkos::create_mirror_view(Kokkos::WithoutInitializing, d_view);
-    else
-      h_view = Kokkos::create_mirror_view(d_view);
+      : modified_flags(t_modified_flags("DualView::modified_flags")) {
+    if constexpr (Impl::ViewCtorProp<P...>::sequential_host_init) {
```
romintomasetti, line 279: I think it's time to discuss if we want to allow the dual view constructor to be passed both host and device `arg_prop`.

With this change, it means I cannot build the device view on a given execution space instance.

masterleinad: You can but not with `SequentialHostInit` (which didn't work properly before anyway) so this doesn't create any new restrictions.

> I think it's time to discuss if we want to allow the dual view constructor to be passed both host and device arg_prop.

Sure but not here.

romintomasetti: I was suggesting a constructor that takes 2 `ViewCtorProp`, since we want to build 2 views.

masterleinad: That's out-of-scope. The point of this pull request is to make sure we can use `SequentialHostInit` with `DualView` in a reasonable way. Adding another `ViewCtorProp` is orthogonal to that.

containers/src/Kokkos_DualView.hpp
```diff
@@ -275,14 +275,29 @@ class DualView : public ViewTraits<DataType, Properties...> {
...
+      h_view = t_host(arg_prop, n0, n1, n2, n3, n4, n5, n6, n7);
+      static_assert(Impl::ViewCtorProp<P...>::initialize,
+                    "DualView: Kokkos::SequentialHostInit isn't compatible "
+                    "with Kokkos::WithoutInitializing!");
+      static_assert(!Impl::ViewCtorProp<P...>::has_execution_space,
+                    "DualView: Kokkos::SequentialHostInit isn't compatible "
+                    "with providing an execution space instance!");
+
+      d_view =
+          Kokkos::create_mirror_view(typename traits::memory_space{}, h_view);
```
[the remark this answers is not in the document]
crtrott: @dalg24

containers/src/Kokkos_DualView.hpp
```diff
@@ -1046,9 +1061,11 @@ class DualView : public ViewTraits<DataType, Properties...> {
       }
     };
 
-    constexpr bool has_execution_space = alloc_prop_input::has_execution_space;
-
-    if constexpr (has_execution_space) {
+    if constexpr (alloc_prop_input::sequential_host_init) {
+      sync<typename t_host::memory_space>();
+      resize_on_host(arg_prop);
```
crtrott, line 1066: Does this work? It will call resync_device forwarding the properties. Doesn't that resent then need to do without initializing?

crtrott (changes requested):
I think this isn't quite right yet.

containers/src/Kokkos_DualView.hpp
masterleinad: `resync_device` largely ignores the properties. We only forward `WithoutInitiailzing`.

containers/src/Kokkos_DualView.hpp
dalg24: Let's take that discussion elsewhere

dalg24: By that I agree this is out of scope of this work, please open a separate issue

containers/unit_tests/TestDualView.hpp
```diff
@@ -622,6 +622,22 @@ TEST(TEST_CATEGORY,
...
+  dv_v.h_view(0) = v;
+  dv_v.h_view(1) = w;
+
+  dv_v.modify_host();
+  dv_v.sync_device();
+
+  dv_v.resize(Kokkos::view_alloc(Kokkos::SequentialHostInit), 2u);
+  ASSERT_EQ(dv_v.d_view.size(), 2u);
+  ASSERT_EQ(dv_v.h_view.size(), 2u);
+}
```
dalg24, line 640: What about realloc?

containers/unit_tests/TestDualView.hpp
```diff
@@ -622,6 +622,22 @@ TEST(TEST_CATEGORY,
             const_dv.view<Kokkos::DefaultExecutionSpace>());
 }
 
+TEST(TEST_CATEGORY, dualview_of_view) {
```
dalg24, line 625: Isn't this more about the use of the `SequentialHostInit` view alloc property?
The inner element type doesn't need to be views nor contain views, does it?

containers/src/Kokkos_DualView.hpp
```diff
@@ -275,14 +275,29 @@ class DualView : public ViewTraits<DataType, Properties...> {
...
-                                         P...>::value)
-      h_view = Kokkos::create_mirror_view(Kokkos::WithoutInitializing, d_view);
-    else
-      h_view = Kokkos::create_mirror_view(d_view);
+      : modified_flags(t_modified_flags("DualView::modified_flags")) {
+    if constexpr (Impl::ViewCtorProp<P...>::sequential_host_init) {
+      h_view = t_host(arg_prop, n0, n1, n2, n3, n4, n5, n6, n7);
+      static_assert(Impl::ViewCtorProp<P...>::initialize,
+                    "DualView: Kokkos::SequentialHostInit isn't compatible "
+                    "with Kokkos::WithoutInitializing!");
```
dalg24, line 283: Drop the Kokkos:: qualifiers in the comment?

containers/src/Kokkos_DualView.hpp
maartenarnst: I agree with @romintomasetti that it is important that it works to pass an execution space instance. The current behavior is that if the user passes an execution space instance to the constructor in the constructor properties, this execution space instance is forwarded to the constructor of the device view, but not to the constructor of the host view. So it seems fair to try to maintain this behavior with `SequentialHostInit`.

A solution with two constructor properties looks like a new approach. So I can understand that such a solution would be out of scope.

Perhaps there are other solutions. With the constructor currently in this PR, if the user passes an `arg_prop` with `SequentialHostInit`, that `arg_prop` is forwarded "as is" to the constructor of the host view. And so there can't be a device execution space in that `arg_prop`. Does this suggest that we should not forward that `arg_prop` "as is" to the constructor of the host view? E.g. a solution whereby we'd create a new `arg_prop_h` (without the passed exec, if any) that we forward to the constructor of the host view, and a new `arg_prop_d` (with the passed exec, if any), that we pass to the `create_mirror_view` to create the device view.

dalg24: The status quo (what is currently in develop) is that the SequentialHostInit is essentially incompatible with DualView.
This PR stands on its own.  It enables the most straightforward use cases, bringing basic support.
Romin is asking for more advanced use cases which raises a lot more questions.
My ruling that this is out of scope is not a judgment of the value/importance of advance use cases, just pragmatism.  The current PR is an improvement and does not prevent further extensions.

containers/src/Kokkos_DualView.hpp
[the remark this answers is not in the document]
dalg24: Agreeing with Christian, this looks fishy

containers/src/Kokkos_DualView.hpp
maartenarnst: Hi @dalg24. Sorry, it looks like you replied while I was typing my comment. OK to take it elsewhere indeed.

containers/src/Kokkos_DualView.hpp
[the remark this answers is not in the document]
dalg24: Try a type that contains a view that has a non empty view when default constructing

containers/unit_tests/TestDualView.hpp
masterleinad: > What about realloc?

See pull requets description. In more detail: it's broken anyway for View of View since we are doing `Kokkos::deep_copy(v, v::value_type{})` which isn't interpreted as "copy a scalar to a View" but as "copy between two Views"

containers/unit_tests/TestDualView.hpp
masterleinad: No, it doesn't need to be a View of Views but it makes for a better test (and that one kind of needs `SequentialHostInit` so I rather called it `dualview_of_view` instead of `dualview_sequential_host_alloc` but I don't feel strongly about the name).

containers/src/Kokkos_DualView.hpp
masterleinad: No strong opinion. Users probably use `Kokkos::WithoutInitializing` instead of the unqualified `WithoutInitializing`.

containers/src/Kokkos_DualView.hpp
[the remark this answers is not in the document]
masterleinad: I think it's reasonable to assume that value types for `DualView` have the same behavior on the host and the device, i.e., they are either default constructible on both the host and the device or not constructible on either of them (in which case `SequentialHostInit` can't be used anyway).

containers/unit_tests/TestDualView.hpp
```diff
@@ -622,6 +622,42 @@ TEST(TEST_CATEGORY,
...
+
+  return dv_v;
+}
+
+TEST(TEST_CATEGORY, dualview_sequential_host_init) {
+  auto dv_v =
+      test_sequential_host_init<Kokkos::View<double*, TEST_EXECSPACE>>();
+  dv_v.resize(Kokkos::view_alloc(Kokkos::SequentialHostInit), 2u);
+  ASSERT_EQ(dv_v.d_view.size(), 2u);
+  ASSERT_EQ(dv_v.h_view.size(), 2u);
```
masterleinad, line 657: See https://github.com/kokkos/kokkos/issues/7466 for issues when trying `resize` with the `S` type.

crtrott (approved):
I think this is ok.

crtrott:
Retest this please.

containers/src/Kokkos_DualView.hpp
```diff
@@ -1046,9 +1061,22 @@ class DualView : public ViewTraits<DataType, Properties...> {
...
+      static_assert(!alloc_prop_input::has_execution_space,
+                    "DualView: SequentialHostInit isn't compatible with "
+                    "providing an execution space instance!");
 
-    if constexpr (has_execution_space) {
+      if (sizeMismatch) {
+        sync<typename t_host::memory_space>();
+        ::Kokkos::resize(arg_prop, h_view, n0, n1, n2, n3, n4, n5, n6, n7);
+        d_view =
+            create_mirror_view_and_copy(typename t_dev::memory_space(), h_view);
```
tcclevenger, line 1076: Should we be updating the modify flags for host and device here?

masterleinad: `sync` should make sure that the flags are set appropriately and we then modify host and deviove view consistently.

containers/unit_tests/TestDualView.hpp
```diff
@@ -622,6 +622,42 @@ TEST(TEST_CATEGORY,
...
+  V v_;
+
+ public:
+  template <class... Extents>
+  S(std::string label, Extents... extents) : v_(std::move(label), extents...) {}
+  S() : v_("v", 10) {}
+};
+
+template <typename V>
+auto test_sequential_host_init() {
```
nmm0, line 637: This function is named `test_sequential_host_init` but doesn't have any assertions or tests, maybe rename to `initialize_dual_view_sequential_host_init`

janciesko (approved):
Looks good to me

containers/unit_tests/TestDualView.hpp
```diff
@@ -622,6 +622,42 @@ TEST(TEST_CATEGORY,
...
+  dv_v.modify_host();
+  dv_v.sync_device();
+
+  return dv_v;
+}
+
+TEST(TEST_CATEGORY, dualview_sequential_host_init) {
+  auto dv_v =
+      test_sequential_host_init<Kokkos::View<double*, TEST_EXECSPACE>>();
+  dv_v.resize(Kokkos::view_alloc(Kokkos::SequentialHostInit), 2u);
```
nmm0, line 655: This seems like its missing more tests, can we use the non view-of-view reproducer you made in the issue?

containers/unit_tests/TestDualView.hpp
nmm0: Can we test realloc without view of views?

nmm0 (approved):
Looks good to me!

tcclevenger (approved):

crtrott:
@masterleinad doesn't seem to compile on HIP anymore?

crtrott:
```
/var/jenkins/workspace/Kokkos_PR-7456/containers/unit_tests/../src/Kokkos_DualView.hpp:978:5: note: in instantiation of function template specialization 'Kokkos::DualView<double *>::impl_realloc<Kokkos::Impl::SequentialHostInit_t>' requested here
    impl_realloc(n0, n1, n2, n3, n4, n5, n6, n7, arg_prop);
    ^
/var/jenkins/workspace/Kokkos_PR-7456/containers/unit_tests/TestDualView.hpp:665:6: note: in instantiation of function template specialization 'Kokkos::DualView<double *>::realloc<Kokkos::Impl::SequentialHostInit_t>' requested here
  dv.realloc(Kokkos::view_alloc(Kokkos::SequentialHostInit), 3);
```

crtrott (approved):

Issue #7375, opened 2024-09-25
Nightly Trilinos test failures: Phalanx and Panzer package unit tests, gcc compilers


ndellingwood:
**Describe the bug**

The following Phalanx tests fail in OpenMP builds with gcc compilers when testing with kokkos@develop
**Phalanx_tKokkos_MPI_1** (gdb stacktrace)
```
...
3. kokkos_DynRankView_UnitTest ... [Passed] (0.00154 sec)
malloc_consolidate(): invalid chunk size

Thread 1 "Phalanx_tKokkos" received signal SIGABRT, Aborted.
0x00007ffff3e8b94c in __pthread_kill_implementation () from /lib64/libc.so.6
Missing separate debuginfos, use: dnf debuginfo-install blas-3.9.0-10.el9.x86_64 glibc-2.34-100.el9_4.2.x86_64 lapack-3.9.0-10.el9.x86_64 nvidia-driver-cuda-libs-560.35.03-1.el9.x86_64
(gdb) bt
#0  0x00007ffff3e8b94c in __pthread_kill_implementation () from /lib64/libc.so.6
#1  0x00007ffff3e3e646 in raise () from /lib64/libc.so.6
#2  0x00007ffff3e287f3 in abort () from /lib64/libc.so.6
#3  0x00007ffff3e29130 in __libc_message.cold () from /lib64/libc.so.6
#4  0x00007ffff3e959f7 in malloc_printerr () from /lib64/libc.so.6
#5  0x00007ffff3e96728 in malloc_consolidate () from /lib64/libc.so.6
#6  0x00007ffff3e982f8 in _int_malloc () from /lib64/libc.so.6
#7  0x00007ffff3e99809 in malloc () from /lib64/libc.so.6
#8  0x00007ffff42af82c in operator new (sz=2049) at /tmp/root/spack-stage/spack-stage-gcc-11.4.0-eptquy7sg7d42upooqgi5ganz7uqoauu/spack-src/libstdc++-v3/libsupc++/new_op.cc:50
#9  0x00007ffff43428c5 in std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >::reserve (this=this@entry=0x7fffffff3f70, __res=<optimized out>)
[...]

PR #7457, opened 2024-10-17
Reintroduce Phalanx special DynRankView operator[]

crtrott:
@ndellingwood @rppawlo 

I am reintroducing some special sauce for Phalanx which got lost in the rewrite despite the very explicit comments in the old code: https://github.com/kokkos/kokkos/blob/15dc143e5f39949eece972a798e175c4b463d4b8/containers/src/Kokkos_DynRankView.hpp#L585-L621

However, the comment also indicates that that was meant as more of a temporary accommodation not a thing we support in perpetua so this PR introduces this behavior back in ONLY with deprecated code enabled, and in debug mode will check that the operator[] is only called if the dynamic rank is 1.

@rppawlo if this is absolutely unacceptable we can talk about this. But I have fairly strong reservations about carrying this behavior forward. 

Hopefully fixes all of: https://github.com/kokkos/kokkos/issues/7375

### DIFF
--- a/containers/src/Kokkos_DynRankView.hpp
+++ b/containers/src/Kokkos_DynRankView.hpp
@@ -580,9 +580,37 @@ class DynRankView : private View<DataType*******, Properties...> {
              index_type i6 = 0) const {
     return view_type::operator()(i0, i1, i2, i3, i4, i5, i6);
   }
+
+// This is an accomodation for Phalanx, that is usint the operator[] to access
+// all elements in a linear fashion even when the rank is not 1
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
   KOKKOS_FUNCTION reference_type operator[](index_type i0) const {
+    if constexpr (std::is_same_v<typename drvtraits::value_type,
+                                 typename drvtraits::scalar_array_type>) {
+      return view_type::data()[i0];
+    } else {
+      const size_t dim_scalar = view_type::impl_map().dimension_scalar();
+      const size_t bytes      = view_type::span() / dim_scalar;
+
+      using tmp_view_type =
+          Kokkos::View<DataType*, typename traits::array_layout,
+                       typename traits::device_type,
+                       Kokkos::MemoryTraits<traits::memory_traits::impl_value |
+                                            unsigned(Kokkos::Unmanaged)>>;
+      tmp_view_type rankone_view(view_type::data(), bytes, dim_scalar);
+      return rankone_view(i0);
+    }
+  }
+#else
+  KOKKOS_FUNCTION reference_type operator[](index_type i0) const {
+#ifdef KOKKOS_ENABLE_DEBUG
+    if (rank() != 1u)
+      Kokkos::abort("DynRankView operator[] can only be used for rank-1");
+#endif
     return view_type::operator()(i0, 0, 0, 0, 0, 0, 0);
   }
+#endif
+
   KOKKOS_FUNCTION reference_type access(index_type i0 = 0, index_type i1 = 0,
                                         index_type i2 = 0, index_type i3 = 0,
                                         index_type i4 = 0, index_type i5 = 0,


ndellingwood:
Thanks @crtrott ! Changes look reasonable, I'll follow up after testing

ndellingwood (approved):
The phalanx tests failing in https://github.com/kokkos/kokkos/issues/7375 are passing, thanks @crtrott ! Should wait on Roger's feedback before final approval/merge

ndellingwood:
Huh, this also fixed the failing Panzer subtest from #7375 as well, even though that test didn't use Fad types...

rppawlo:
Thanks for fixing @crtrott and @ndellingwood ! I'm all for deprecating this code. In the failing phalanx unit test, I found this comment I made in 2016 during the initial transition to kokkos:
```
    // In the initial implementation of the Kokkos::DynRankView                                                                                                                          
    // bracket operator, you could only use it on a rank-1 view. This                                                                                                                    
    // is not how Intrepid uses it. Intrepid expects access to the                                                                                                                       
    // entire size of the array. This is dangerous and will not work                                                                                                                     
    // with subviews/padding. So the bracket op was changed to support                                                                                                                   
    // this but is considered a hack until we can remove all bracket                                                                                                                     
    // op use.               
```

containers/src/Kokkos_DynRankView.hpp
```diff
@@ -580,9 +580,38 @@ class DynRankView : private View<DataType*******, Properties...> {
...
+      tmp_view_type rankone_view(view_type::data(), bytes, dim_scalar);
+      return rankone_view(i0);
+    }
+  }
+#else
+  KOKKOS_FUNCTION reference_type operator[](index_type i0) const {
+#ifdef KOKKOS_ENABLE_DEBUG
+    if (view_type::rank() != 1u)
+      Kokkos::abort("DynRankView operator[] can only be used for rank-1");
+#endif
```
masterleinad, line 610: This looks like a static property so we don't need a runtime check
```suggestion
    static_assert(view_type::rank() == 1u, "DynRankView operator[] can only be used for rank-1");
```
but it's also defined as
```
  using view_type = View<DataType*******, Properties...>;
```
so it wouldn't trigger anyway. What am I missing?

crtrott: yeah lol you are right it needs to call `rank()`

containers/src/Kokkos_DynRankView.hpp
```diff
@@ -580,9 +580,38 @@ class DynRankView : private View<DataType*******, Properties...> {
...
+    } else {
+      const size_t dim_scalar = view_type::impl_map().dimension_scalar();
+      const size_t bytes      = view_type::span() / dim_scalar;
+
+      using tmp_view_type = Kokkos::View<
+          DataType*, typename traits::array_layout,
+          typename traits::device_type,
+          Kokkos::MemoryTraits<traits::memory_traits::is_unmanaged |
+                               traits::memory_traits::is_random_access |
+                               traits::memory_traits::is_atomic>>;
```
nmm0, line 600: Maybe a dumb question, but aren't these all booleans? Thus the memory traits would either be 0 or Unmanaged?

containers/src/Kokkos_DynRankView.hpp
masterleinad:
```suggestion
#ifdef KOKKOS_ENABLE_DEBUG
    if (rank() != 1u)
      Kokkos::abort("DynRankView operator[] can only be used for rank-1");
#endif
```

containers/src/Kokkos_DynRankView.hpp
masterleinad: No, you probably want `MemoryTraitsFlags`
```suggestion
          Kokkos::MemoryTraits<Kokkos::Unmanaged |
                               Kokkos::RandomAccess |
                               Kokkos::Atomic>>;
```

nmm0: Would you want to take the values from traits, rather than enabling all three flags?

masterleinad: Like
```
(traits::memory_traits::is_unmanaged ? Kokkos::Unmanaged : 0) | ...
```

nmm0: yeah but I'm honestly not sure what this function is supposed to do

masterleinad: To me it looks like we want to reinterpret the DynRankView as 1-D View to access all elements preserving the memory traits (ignoring `Restrict` and `Aligned`).

nmm0: @crtrott just copied the old code over, so it looks like this might have been broken for the last 5 years? :P

masterleinad: I mean it apparently worked good enough for `Trilinos`. 🙂

nmm0 (changes requested):
I guess we should probably fix the memory traits bug

crtrott:
Fixed both the Rank and the MemoryTraits thing.

masterleinad (approved):

ndellingwood (approved):

nmm0 (approved):
Looks good to me!

containers/src/Kokkos_DynRankView.hpp
```diff
@@ -580,9 +580,37 @@ class DynRankView : private View<DataType*******, Properties...> {
              index_type i6 = 0) const {
     return view_type::operator()(i0, i1, i2, i3, i4, i5, i6);
   }
+
+// This is an accomodation for Phalanx, who use the operator[] to access all
```
dalg24, line 584: "who uses" or "that is using"?

containers/src/Kokkos_DynRankView.hpp
```diff
@@ -580,9 +580,37 @@ class DynRankView : private View<DataType*******, Properties...> {
...
+// memory
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
   KOKKOS_FUNCTION reference_type operator[](index_type i0) const {
+    if constexpr (std::is_same<typename drvtraits::value_type,
+                               typename drvtraits::scalar_array_type>::value) {
```
dalg24, line 589:
```suggestion
    if constexpr (std::is_same_v<typename drvtraits::value_type,
                               typename drvtraits::scalar_array_type>) {
```
(not formatted)

I thought we were enforcing this?!

containers/src/Kokkos_DynRankView.hpp
```diff
@@ -580,9 +580,37 @@ class DynRankView : private View<DataType*******, Properties...> {
...
+      return view_type::data()[i0];
+    } else {
+      const size_t dim_scalar = view_type::impl_map().dimension_scalar();
+      const size_t bytes      = view_type::span() / dim_scalar;
+
+      using tmp_view_type =
+          Kokkos::View<DataType*, typename traits::array_layout,
+                       typename traits::device_type,
+                       Kokkos::MemoryTraits<traits::memory_traits::impl_value |
+                                            unsigned(Kokkos::Unmanaged)>>;
```
dalg24, line 599: I can't believe we did not mandate the underlying integral type of `MemoryTraitsFlags` is not `unsigned` in the first place and we cast everywhere...

dalg24: Yes sorry I am just ruminating.  Didn't mean to suggest there is anything to fix here

containers/src/Kokkos_DynRankView.hpp
```diff
@@ -580,9 +580,37 @@ class DynRankView : private View<DataType*******, Properties...> {
              index_type i6 = 0) const {
     return view_type::operator()(i0, i1, i2, i3, i4, i5, i6);
   }
+
+// This is an accomodation for Phalanx, that is usint the operator[] to access
```
dalg24, line 584: Haha you misspelled usin~~t~~g
(I don't really care to fix)

PR #7460, opened 2024-10-17
graph(diagnostic): enable compile-time diagnostic of illegal reduction target

romintomasetti:
This PR ensures that a user trying to add a parallel-reduce node whose reduction target is not device accessible will be welcomed with a nice compile-time message.

This is a fundamental difference with regular parallel-reduce for which we make sure that the reduction target will be set to the expected value after the kernel finished using deep copies.

> ⚠️ This PR re-enables a test! The test name was `DISABLED_repeat_chain`, so it was compiled but **always** skipped. This PR re-enables the test, that due to the new `static_assert` cannot be compiled and has to be skipped for illegal configurations, *i.e.* when the graph execution space cannot access one of the reduction target memory space (*e.g.* `bugs_host`).

### DIFF
--- a/core/unit_test/TestGraph.hpp
+++ b/core/unit_test/TestGraph.hpp
@@ -329,32 +329,41 @@ TEST_F(TEST_CATEGORY_FIXTURE(graph), when_all_cycle) {
   //----------------------------------------
 }
 
-// This test is disabled because we don't currently support copying to host,
+// This test requires that the graph execution space can access
+// the host memoy space because we don't currently support copying to host,
 // even asynchronously. We _may_ want to do that eventually?
-TEST_F(TEST_CATEGORY_FIXTURE(graph), DISABLED_repeat_chain) {
-  auto graph = Kokkos::Experimental::create_graph(
-      ex, [&, count_host = count_host](auto root) {
-        //----------------------------------------
-        root.then_parallel_for(1, set_functor{count, 0})
-            .then_parallel_for(1, count_functor{count, bugs, 0, 0})
-            .then_parallel_for(1, count_functor{count, bugs, 1, 1})
-            .then_parallel_reduce(1, set_result_functor{count}, count_host)
-            .then_parallel_reduce(
-                1, set_result_functor{bugs},
-                Kokkos::Sum<int, Kokkos::HostSpace>{bugs_host});
-        //----------------------------------------
-      });
+TEST_F(TEST_CATEGORY_FIXTURE(graph), repeat_chain) {
+  constexpr bool result_not_accessible_by_exec = !Kokkos::SpaceAccessibility<
+      TEST_EXECSPACE, decltype(bugs_host)::memory_space>::accessible;
 
-  //----------------------------------------
-  constexpr int repeats = 10;
+  if constexpr (result_not_accessible_by_exec) {
+    GTEST_SKIP() << "The graph requires the reduction targets like 'bugs_host' "
+                    "to be accessible by the execution space.";
+  } else {
+    auto graph = Kokkos::Experimental::create_graph(
+        ex, [&, count_host = count_host](auto root) {
+          //----------------------------------------
+          root.then_parallel_for(1, set_functor{count, 0})
+              .then_parallel_for(1, count_functor{count, bugs, 0, 0})
+              .then_parallel_for(1, count_functor{count, bugs, 1, 1})
+              .then_parallel_reduce(1, set_result_functor{count}, count_host)
+              .then_parallel_reduce(
+                  1, set_result_functor{bugs},
+                  Kokkos::Sum<int, Kokkos::HostSpace>{bugs_host});
+          //----------------------------------------
+        });
 
-  for (int i = 0; i < repeats; ++i) {
-    graph.submit();
-    ex.fence();
-    EXPECT_EQ(2, count_host());
-    EXPECT_EQ(0, bugs_host());
+    //----------------------------------------
+    constexpr int repeats = 10;
+
+    for (int i = 0; i < repeats; ++i) {
+      graph.submit();
+      ex.fence();
+      EXPECT_EQ(2, count_host());
+      EXPECT_EQ(0, bugs_host());
+    }
+    //----------------------------------------
   }
-  //----------------------------------------
 }
 
 TEST_F(TEST_CATEGORY_FIXTURE(graph), zero_work_reduce) {
--- a/core/src/Kokkos_GraphNode.hpp
+++ b/core/src/Kokkos_GraphNode.hpp
@@ -358,6 +358,23 @@ class GraphNodeRef {
                       Kokkos::is_reducer<return_type_remove_cvref>::value,
                   "Output argument to parallel reduce in a graph must be a "
                   "View or a Reducer");
+
+    if constexpr (Kokkos::is_reducer_v<return_type_remove_cvref>) {
+      static_assert(
+          Kokkos::SpaceAccessibility<
+              ExecutionSpace, typename return_type_remove_cvref::
+                                  result_view_type::memory_space>::accessible,
+          "The reduction target must be accessible by the graph execution "
+          "space.");
+    } else {
+      static_assert(
+          Kokkos::SpaceAccessibility<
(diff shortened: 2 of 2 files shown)


romintomasetti:
@maartenarnst will be interested.

core/unit_test/TestGraph.hpp
```diff
@@ -329,32 +329,38 @@ TEST_F(TEST_CATEGORY_FIXTURE(graph), when_all_cycle) {
...
-            .then_parallel_reduce(1, set_result_functor{count}, count_host)
-            .then_parallel_reduce(
-                1, set_result_functor{bugs},
-                Kokkos::Sum<int, Kokkos::HostSpace>{bugs_host});
-        //----------------------------------------
-      });
+TEST_F(TEST_CATEGORY_FIXTURE(graph), repeat_chain) {
+  constexpr bool skip = !Kokkos::SpaceAccessibility<
+      Kokkos::Impl::remove_cvref_t<decltype(ex)>,
+      typename Kokkos::Impl::remove_cvref_t<
```
dalg24, line 338: typename keyword is not needed

core/unit_test/TestGraph.hpp
```diff
@@ -329,32 +329,38 @@ TEST_F(TEST_CATEGORY_FIXTURE(graph), when_all_cycle) {
...
+          decltype(bugs_host)>::memory_space>::accessible;
```
dalg24, line 339: Why are we not just saying "host"?

romintomasetti: What do you mean ?

core/unit_test/TestGraph.hpp
romintomasetti: OK, removed.

core/unit_test/TestGraph.hpp
```diff
@@ -329,32 +329,38 @@ TEST_F(TEST_CATEGORY_FIXTURE(graph), when_all_cycle) {
...
-            .then_parallel_reduce(
-                1, set_result_functor{bugs},
-                Kokkos::Sum<int, Kokkos::HostSpace>{bugs_host});
-        //----------------------------------------
-      });
+TEST_F(TEST_CATEGORY_FIXTURE(graph), repeat_chain) {
+  constexpr bool skip = !Kokkos::SpaceAccessibility<
+      Kokkos::Impl::remove_cvref_t<decltype(ex)>,
+      Kokkos::Impl::remove_cvref_t<
+          decltype(bugs_host)>::memory_space>::accessible;
```
masterleinad, line 339:
```suggestion
  constexpr bool skip = !Kokkos::SpaceAccessibility<
      TEST_EXESPACE,
      Kokkos::HostSpace>::accessible;
```

romintomasetti: I agree for `TEST_EXESPACE`.

For `bugs_host`, I would tend to argue that it helps relate the condition to the graph creation a few lines below.

core/unit_test/TestGraph.hpp
dalg24: that I don't understand why we have such a complex deduction of the space if in the end you know/you rely on the fact that this is the host.  I am basing this on the skip message you provided.

romintomasetti: I think I should be more specific in the message:
> The reduction target 'bugs_host' memory space cannot be accessed by the graph.

I just want to make sure that if someone changes the `bugs_host` to be in a shared space, this check will enable the test.

romintomasetti:
@dalg24 @masterleinad Please read the PR description again (I just updated it and it should help you).

core/unit_test/TestGraph.hpp
```diff
@@ -329,32 +329,36 @@ TEST_F(TEST_CATEGORY_FIXTURE(graph), when_all_cycle) {
...
 }
 
-// This test is disabled because we don't currently support copying to host,
+// This test requires that the graph execution space can access
+// the host memoy space because we don't currently support copying to host,
 // even asynchronously. We _may_ want to do that eventually?
-TEST_F(TEST_CATEGORY_FIXTURE(graph), DISABLED_repeat_chain) {
-  auto graph = Kokkos::Experimental::create_graph(
-      ex, [&, count_host = count_host](auto root) {
-        //----------------------------------------
```
masterleinad, line 337: There doesn't seem to be a reason for changing these lines and below.

core/unit_test/TestGraph.hpp
```diff
@@ -329,32 +329,36 @@ TEST_F(TEST_CATEGORY_FIXTURE(graph), when_all_cycle) {
...
-            .then_parallel_for(1, count_functor{count, bugs, 0, 0})
-            .then_parallel_for(1, count_functor{count, bugs, 1, 1})
-            .then_parallel_reduce(1, set_result_functor{count}, count_host)
-            .then_parallel_reduce(
-                1, set_result_functor{bugs},
-                Kokkos::Sum<int, Kokkos::HostSpace>{bugs_host});
-        //----------------------------------------
-      });
+TEST_F(TEST_CATEGORY_FIXTURE(graph), repeat_chain) {
+  constexpr bool skip = !Kokkos::SpaceAccessibility<
```
masterleinad, line 336:
```suggestion
  constexpr bool result_not_host_accessible = !Kokkos::SpaceAccessibility<
```

core/unit_test/TestGraph.hpp
romintomasetti: What do you mean ?

core/unit_test/TestGraph.hpp
romintomasetti: OK, done :heavy_check_mark:

core/unit_test/TestGraph.hpp
masterleinad: No, you deleted a bunch of lines like
```
```
which just made the diff bigger (ignoring whitespace changes) than necessary.

core/unit_test/TestGraph.hpp
```diff
@@ -329,32 +329,36 @@ TEST_F(TEST_CATEGORY_FIXTURE(graph), when_all_cycle) {
...
+      TEST_EXECSPACE, decltype(bugs_host)::memory_space>::accessible;
+
+  if constexpr (skip) {
+    GTEST_SKIP() << "The host memory space cannot be accessed by the graph.";
```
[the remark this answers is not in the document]
romintomasetti: Agreed, I just modified the sentence slightly... OK with you now ? :smile:

core/unit_test/TestGraph.hpp
romintomasetti: To me these lines are useless and make the code harder to read (I guess people would use that to expand/collapse that region, but in general it's just distracting). But if you want them back, I'll do.

core/unit_test/TestGraph.hpp
```diff
@@ -329,32 +329,37 @@ TEST_F(TEST_CATEGORY_FIXTURE(graph), when_all_cycle) {
...
-            .then_parallel_reduce(
-                1, set_result_functor{bugs},
-                Kokkos::Sum<int, Kokkos::HostSpace>{bugs_host});
-        //----------------------------------------
-      });
+TEST_F(TEST_CATEGORY_FIXTURE(graph), repeat_chain) {
+  constexpr bool result_not_accessible_by_exec = !Kokkos::SpaceAccessibility<
+      TEST_EXECSPACE, decltype(bugs_host)::memory_space>::accessible;
+
+  if constexpr (skip) {
```
masterleinad, line 339:
```suggestion
  if constexpr (result_not_accessible_by_exec) {
```

core/unit_test/TestGraph.hpp
masterleinad: Yeah, let's just put them back since you have to fix the `skip` variable anyway.

core/unit_test/TestGraph.hpp
romintomasetti: :heavy_check_mark:

masterleinad (approved):

dalg24 (approved):

PR #7461, opened 2024-10-17
Move BasicView to Impl and disable for classic Intel compiler

crtrott:
We had issues with the classic Intel compiler where BasicView exposes bugs which are hard to impossible to work around. 
Hence we disable BasicView here, and back out the attempted workaround in SharedAllocationTracker. That workaround had resulted in Intel compiler ICE when enabling AVX2 in other code.

I am also moving BasicView into Impl, so disabling it is not actually a user facing difference when using the Intel compiler. But BasicView was anyway meant to be in the Impl namespace, it was a mistake that it wasn't.

### DIFF
--- a/core/unit_test/CMakeLists.txt
+++ b/core/unit_test/CMakeLists.txt
@@ -96,7 +96,7 @@ set(COMPILE_ONLY_SOURCES
     view/TestExtentsDatatypeConversion.cpp
 )
 
-if(NOT Kokkos_ENABLE_IMPL_MDSPAN)
+if(NOT Kokkos_ENABLE_IMPL_MDSPAN OR KOKKOS_CXX_COMPILER_ID STREQUAL "Intel")
   list(REMOVE_ITEM COMPILE_ONLY_SOURCES view/TestBasicViewMDSpanConversion.cpp)
 endif()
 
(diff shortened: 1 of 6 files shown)


masterleinad (approved):

dalg24 (commented):
Error out at configuration time if `Kokkos_ENABLE_IMPL_MDSPAN` is enabled and the Intel C++ Compiler Classic is used

core/unit_test/CMakeLists.txt
```diff
@@ -358,6 +358,9 @@ foreach(Tag Threads;Serial;OpenMP;Cuda;HPX;OpenMPTarget;OpenACC;HIP;SYCL)
         configure_file(${dir}/dummy.cpp ${file})
         list(APPEND ${Tag}_VIEWSUPPORT ${file})
       endforeach()
+      if(Kokkos_CXX_COMPILER_ID STREQUAL "Intel")
```
nmm0, line 361: We probably should also guard this with `Kokkos_ENABLE_IMPL_MDSPAN`

nmm0 (approved):
Looks good to me

core/unit_test/CMakeLists.txt
ndellingwood, line 361:
```suggestion
      if(NOT Kokkos_ENABLE_IMPL_MDSPAN OR KOKKOS_CXX_COMPILER_ID STREQUAL "Intel")
```

crtrott: It already is a few lines up

dalg24 (approved):

ndellingwood (approved):
Tested icpc (intel/2023.1.0), serial, mdspan=on, Kokkos builds complete and tests pass 100% 
Trilinos build in progress

PR #7462, opened 2024-10-17
Introduce AMD_GFX942_APU and add autodetection

Rombur:
This PR introduces a new architecture `AMD_GFX942_APU` which corresponds to MI300A. When this architecture is set, we enable the unified memory. I have updated the autodetection to detect if we use an APU or a discrete GPU. One thing that we need to decide is the value of the macro `KOKKOS_ARCH_AMD_GPU`. Right now, the value is `gfx942` for both the APU and the discrete GPU. Since this is the flag that both architectures use, I would keep it as is but we could set the macro to `gfx942_apu`.

I still need to update the Makefile.

I don't have access to MI300, so I need someone to test this PR.


cmake/kokkos_arch.cmake
```diff
@@ -1163,6 +1163,26 @@ if(KOKKOS_ARCH_HOPPER90)
...
+  )
+
+  if(NOT COMPILE_RESULT OR NOT RESULT EQUAL 0)
+    message(SEND_ERROR "Autodetection of AMD APU failed."
+                       "Please manually specify one AMD GPU architecture via -DKokkos_ARCH_{..}=ON'."
+    )
+  endif()
+
+  if(AMD_APU)
+    set(${ARCH} AMD_GFX942_APU PARENT_SCOPE)
```
crtrott, line 1182: Maybe we should check here also that the little agent reporter return gfx942? Not sure if there are other APUs, now or in the future?

crtrott: ah never mind we only do this if its already gfx942 sounds good, we can later expand if and when needed.

cmake/kokkos_arch.cmake
```diff
@@ -1255,6 +1279,10 @@ foreach(ARCH IN LISTS SUPPORTED_AMD_ARCHS)
   endif()
 endforeach()
 
+if(KOKKOS_ARCH_AMD_GFX942_APU)
+  set(KOKKOS_ENABLE_IMPL_HIP_UNIFIED_MEMORY ON)
```
crtrott, line 1283: does this need to be 'Kokkos_ENABLE_IMPL_HIP_UNIFIED_MEMORY'? Also I think we could just remove the cmake level option for this and define it inside the source code based on the architecture macro. Thoughts?

cmake/kokkos_arch.cmake
Rombur: I didn't want to make the code more general/complicated not knowing if this will be a one-off.

cmake/kokkos_arch.cmake
Rombur: I thought we wanted to keep the CMake option for the next release and then remove. I am fine removing the CMake option now.

cmake/kokkos_arch.cmake
crtrott: As I said its fine, we are already inside a gfx942 clause and we can generalize if and when necessary.

cmake/kokkos_arch.cmake
crtrott: Yeah let's remove it.

crtrott (changes requested):
Can we add a "warning" level check during Kokkos::initialize if you are on an MI300A but have not chosen the APU arch?

crtrott (approved):

cmake/compile_tests/amd_apu.cc
```diff
@@ -0,0 +1,38 @@
...
+  hipError_t error = hipGetDeviceProperties(&hipProp, 0);
+
+  if (error != hipSuccess) {
+    std::cout << hipGetErrorString(error) << '\n';
+    return error;
+  }
+
+  if (hipProp.integrated == 1) {
+    // We detected an APU
+    std::cout << "ON";
```
cedricchevalier19, line 31: On the MI300A I have access to, `hipProp.integrated == 0` when using rocm 6.1. 
But it works great with rocm 6.2.

cedricchevalier19 (commented):
Weird results for the automatic detection depending on rocm version.

Note for later: we will have to figure out how to deal with the APU in the spack recipe. Adding `_apu` to the architecture is not really an option there.

ldh4:
I'm seeing the same result as @cedricchevalier19 on MI300A that I have access to.
With rocm 6.2, the cmake autodetection correctly identifies `AMD_GFX942_APU` to be the architecture, but with rocm 6.1 and below, the autodetection sets the arch as `AMD_GFX942`.

core/src/HIP/Kokkos_HIP_ZeroMemset.hpp
```diff
@@ -32,8 +32,7 @@ struct ZeroMemset<HIP> {
     // in ROCm <= 6.2.0, hipMemsetAsync on a host-allocated pointer
     // returns an invalid value error, but accessing the data via a
     // GPU kernel works.
-#if defined(KOKKOS_ENABLE_IMPL_HIP_UNIFIED_MEMORY) && \
-    defined(KOKKOS_ARCH_AMD_GFX942)
+#if defined(KOKKOS_ENABLE_IMPL_HIP_UNIFIED_MEMORY)
```
ldh4, line 35: Isn't `KOKKOS_ENABLE_IMPL_HIP_UNIFIED_MEMORY` being removed as part of this PR?
Should this be ifdef'ed on `KOKKOS_ARCH_AMD_GFX942_APU` instead?

crtrott: No I want that as an internal typedef for all APU versions. I.e. if there ever is another AMD APU we would use the same option, even though the arch is different. I just don't really want it as a CMake option.

core/src/HIP/Kokkos_HIP.cpp
```diff
@@ -66,6 +66,15 @@ void HIP::impl_initialize(InitializationSettings const& settings) {
...
   }
 
+  // Print a warning if the user did not select the APU architecture when using
+  // a MI300A
+#ifdef KOKKOS_ARCH_AMD_GFX942
+  if ((Kokkos::show_warnings()) && (hipProp.integrated == 1)) {
+    std::cerr << "Kokkos::HIP::initialize WARNING: running kernels for MI300X "
+                 "(discrete GPU) on a MI300A (APU).\n";
+  }
+#endif
```
crtrott, line 76: Should we abort in the other direction? I.e. if you have set KOKKOS_ARCH_AMD_GFX942_APU (or for that matter if the UNIFIED_MEMORY thing is defined) but hipProp.integrated==0 we abort?

Rombur: I was thinking that if you set KOKKOS_ARCH_AMD_GFX942_APU on a MI300X, the code will crash right away and you will realize that you did something wrong. While if you set KOKKOS_ARCH_AMD_GFX942 on a MI300A, you won't get the behavior that you expect but it won't be obvious.

cmake/compile_tests/amd_apu.cc
Rombur: It looks like `hipProp.integrated` was added in [ROCm 6.2](https://rocm.docs.amd.com/en/docs-6.2.0/about/release-notes.html#id3)... It seems that before that version there is no way to know if you are using an APU.

core/src/HIP/Kokkos_HIP.cpp
crtrott: Will it crash? Even if you have like the Linux memory management thingy?

Rombur: You are right. It won't crash in that case. I'll just change the warning to always warn if you pick the wrong architecture.

core/src/HIP/Kokkos_HIP.cpp
```diff
@@ -66,14 +66,19 @@ void HIP::impl_initialize(InitializationSettings const& settings) {
...
   if ((Kokkos::show_warnings()) && (hipProp.integrated == 1)) {
     std::cerr << "Kokkos::HIP::initialize WARNING: running kernels for MI300X "
                  "(discrete GPU) on a MI300A (APU).\n";
   }
 #endif
+#ifdef KOKKOS_ARCH_AMD_GFX942_APU
+  if ((Kokkos::show_warnings()) && (hipProp.integrated == 0)) {
+    std::cerr << "Kokkos::HIP::initialize WARNING: running kernels for MI300A "
+                 "(APU) on a MI300X (discrete GPU).\n";
+  }
```
ldh4, line 80: I get an error when enabling `KOKKOS_ARCH_AMD_GFX942_APU` with rocm 6.1:
```
kokkos/core/src/HIP/Kokkos_HIP.cpp:77:37: error: use of undeclared identifier 'hipProp'
   77 |   if ((Kokkos::show_warnings()) && (hipProp.integrated == 0)) {
      |   
```

Should there be something like this:
```
hipDeviceProp_t hipProp;
hipGetDeviceProperties(&hipProp, 0);
```

Rombur: It should be fixed. I should have use `Impl::HIPInternal::m_deviceProp`

crtrott (changes requested):
I think we still need to sort out the macros, you deleted the line from cmakedefine so is the macro actually defined (UNIFIED_MEMORY)? I want it defined if any APU (right now only GFX942) is defined. But I'd be fine with it only defined in source code not in the cmake.

Rombur:
I've moved the declaration of the macro inside `setup/Kokkos_Setup_HIP.hpp` with the other HIP dedicated macros. I've also renamed it from `KOKKOS_ENABLE_IMPL_HIP_UNIFIED_MEMORY` to `KOKKOS_IMPL_HIP_UNIFIED_MEMORY` to match the other macros.

crtrott (approved):
Looks good to me @ldh4 if you could retest one more time that would be great.

crtrott:
@cedricchevalier19 if you could also rerun the test on MI300A that would be great. But I think this looks good now.

cmake/kokkos_arch.cmake
```diff
@@ -1163,6 +1163,26 @@ if(KOKKOS_ARCH_HOPPER90)
...
+  file(REMOVE_RECURSE ${BINARY_TEST_DIR})
+  file(MAKE_DIRECTORY ${BINARY_TEST_DIR})
+
+  try_run(RESULT COMPILE_RESULT ${BINARY_TEST_DIR} ${CMAKE_CURRENT_SOURCE_DIR}/cmake/compile_tests/amd_apu.cc
+          RUN_OUTPUT_VARIABLE AMD_APU
+  )
+
+  if(NOT COMPILE_RESULT OR NOT RESULT EQUAL 0)
+    message(SEND_ERROR "Autodetection of AMD APU failed."
+                       "Please manually specify one AMD GPU architecture via -DKokkos_ARCH_{..}=ON'."
```
masterleinad, line 1177:
```suggestion
    message(SEND_ERROR "Autodetection of AMD APU failed."
                       "Please manually specify either '-DKokkos_ARCH_GFX942=ON' or '-DKokkos_ARCH_GFX942_APU=ON'"
```

core/src/HIP/Kokkos_HIP.cpp
masterleinad: If users are manually requesting `KOKKOS_ARCH_AMD_GFX942_APU` and are using `ROCm` before 6.2, they will run into a compilation error for `m_deviceProp.integrated` here since that property was only introduced in 6.2?

core/unit_test/TestWithoutInitializing.hpp
```diff
@@ -223,8 +223,7 @@ TEST(TEST_CATEGORY, deep_copy_zero_memset) {
 
   // for MI300A with unified memory, ZeroMemset uses a parallel for
   auto success =
-#if defined(KOKKOS_ENABLE_IMPL_HIP_UNIFIED_MEMORY) && \
-    defined(KOKKOS_ARCH_AMD_GFX942)
+#if defined(KOKKOS_IMPL_HIP_UNIFIED_MEMORY) && defined(KOKKOS_ARCH_AMD_GFX942)
```
masterleinad, line 226: The only way `KOKKOS_IMPL_HIP_UNIFIED_MEMORY` is defined is for `KOKKOS_ARCH_AMD_GFX942_APU`, right?
```suggestion
#if defined(KOKKOS_IMPL_HIP_UNIFIED_MEMORY)
```

core/src/HIP/Kokkos_HIP.cpp
Rombur: It compiles fine. It just returns 0. We already have the same code in `print_configuration` and it works.

ldh4:
Retested on MI300A.
With rocm 6.2, it correctly auto-detected `AMD_GFX942_APU`. Manually setting `-DKokkos_ARCH_AMD_GFX942_APU=ON` also worked fine.
With rocm 6.1, auto-detected `AMD_GFX942`. Manually setting `-DKokkos_ARCH_AMD_GFX942_APU=ON` and running a short test outputted a warning message: `Kokkos::HIP::initialize WARNING: running kernels for MI300A (APU) on a MI300X (discrete GPU).`

Does this mean that rocm 6.2 will be the required version for kokkos on MI300A?

crtrott:
@ldh4 no it is only a warning it will actually just work. Don't think its really worth it to ifdef the warning for ROCM 6.1 and earlier. By the time we run this system in production we will have 6.x as the standard for users anyway.

crtrott:
Retest this please.

cedricchevalier19:
The PR works to choose the correct architecture with ROCM 6.2 (in fact the `integrated` property work if you use a ROCM 6.1 compiler and a 6.2 runtime).
However, I am not able to compile the tests in APU mode, getting this error (here with HIP+OpenMP, but I have the same issue with HIP+Serial):

```

[  5%] Building CXX object core/unit_test/CMakeFiles/Kokkos_CoreTestCompileOnly.dir/TestCreateMirror.cpp.o
In file included from /kokkos/core/unit_test/TestCreateMirror.cpp:17:
In file included from /kokkos/core/src/Kokkos_Core.hpp:45:
In file included from /kokkos/build-mi300-test-a/KokkosCore_Config_DeclareBackend.hpp:22:
In file included from /kokkos/core/src/decl/Kokkos_Declare_OPENMP.hpp:21:
In file included from /kokkos/core/src/OpenMP/Kokkos_OpenMP.hpp:30:
In file included from /kokkos/core/src/Kokkos_HostSpace.hpp:36:
/kokkos/core/src/impl/Kokkos_SharedAlloc.hpp:212:13: error: no matching member function for call to 'allocate'
  212 |       space.allocate(exec_space, label.c_str(),
      |       ~~~~~~^~~~~~~~
/kokkos/core/src/impl/Kokkos_SharedAlloc.hpp:247:13: note: in instantiation of function template specialization 'Kokkos::Impl::checked_allocation_with_header<Kokkos::OpenMP, Kokkos::HIPSpace>' requested here
  247 |             checked_allocation_with_header(exec, space, label, alloc_size),
      |             ^
/kokkos/core/src/impl/Kokkos_SharedAlloc.hpp:408:9: note: in instantiation of function template specialization 'Kokkos::Impl::SharedAllocationRecordCommon<Kokkos::HIPSpace>::SharedAllocationRecordCommon<Kokkos::OpenMP>' requested here
  408 |       : SharedAllocationRecord<MemorySpace, void>(
      |         ^
/kokkos/core/src/impl/Kokkos_SharedAlloc.hpp:445:21: note: in instantiation of function template specialization 'Kokkos::Impl::SharedAllocationRecord<Kokkos::HIPSpace, Kokkos::Impl::ViewValueFunctor<Kokkos::Device<Kokkos::OpenMP, Kokkos::HIPSpace>, int>>::SharedAllocationRecord<Kokkos::OpenMP>' requested here
  445 |         (return new SharedAllocationRecord(exec_space, arg_space, arg_label,
      |                     ^
/kokkos/core/src/View/Kokkos_ViewMapping.hpp:2831:28: note: in instantiation of function template specialization 'Kokkos::Impl::SharedAllocationRecord<Kokkos::HIPSpace, Kokkos::Impl::ViewValueFunctor<Kokkos::Device<Kokkos::OpenMP, Kokkos::HIPSpace>, int>>::allocate<Kokkos::OpenMP>' requested here
 2831 |             ? record_type::allocate(exec_space, mem_space, alloc_name,
      |                            ^
/kokkos/core/src/View/Kokkos_ViewLegacy.hpp:1039:60: note: in instantiation of function template specialization 'Kokkos::Impl::ViewMapping<Kokkos::ViewTraits<int *, Kokkos::LayoutLeft, Kokkos::Device<Kokkos::OpenMP, Kokkos::HIPSpace>, Kokkos::Experimental::EmptyViewHooks>, void>::allocate_shared<Kokkos::Impl::WithoutInitializing_t, std::basic_string<char>, Kokkos::HIPSpace, Kokkos::OpenMP>' requested here
 1039 |     Kokkos::Impl::SharedAllocationRecord<>* record = m_map.allocate_shared(
/kokkos/core/src/Kokkos_CopyViews.hpp:3481:12: note: in instantiation of function template specialization 'Kokkos::View<int *, Kokkos::LayoutLeft, Kokkos::Device<Kokkos::OpenMP, Kokkos::HIPSpace>, Kokkos::Experimental::EmptyViewHooks>::View<Kokkos::Impl::WithoutInitializing_t, std::basic_string<char>>' requested here
 3481 |     return dst_type(prop_copy, src.layout());
      |            ^
/kokkos/core/unit_test/TestCreateMirror.cpp:95:3: note: in instantiation of function template specialization 'test_create_mirror_properties<Kokkos::View<int *, Kokkos::HIP>>' requested here
   95 |   test_create_mirror_properties(device_view);
      |   ^
/kokkos/core/src/HIP/Kokkos_HIP_Space.hpp:81:9: note: candidate function not viable: no known conversion from 'const Kokkos::OpenMP' to 'const HIP' for 1st argument
   81 |   void* allocate(const HIP& exec_space, const char* arg_label,
      |         ^        ~~~~~~~~~~~~~~~~~~~~~
/kokkos/core/src/HIP/Kokkos_HIP_Space.hpp:86:9: note: candidate function not viable: requires at most 3 arguments, but 4 were provided
   86 |   void* allocate(const char* arg_label, const size_t arg_alloc_size,
      |         ^        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
   87 |                  const size_t arg_logical_size = 0) const;
      |                  ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/kokkos/core/src/HIP/Kokkos_HIP_Space.hpp:80:9: note: candidate function not viable: requires 2 arguments, but 4 were provided
   80 |   void* allocate(const HIP& exec_space, const size_t arg_alloc_size) const;
      |         ^        ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
/kokkos/core/src/HIP/Kokkos_HIP_Space.hpp:85:9: note: candidate function not viable: requires single argument 'arg_alloc_size', but 4 arguments were provided
   85 |   void* allocate(const size_t arg_alloc_size) const;
      |         ^        ~~~~~~~~~~~~~~~~~~~~~~~~~~~
1 error generated when compiling for host.
```

Rombur:
I know what's the issue. I'll have a patch soon.

Rombur:
It should be good now.

masterleinad (approved):
Fine with me.

ldh4 (dismissed):

crtrott:
@ldh4 did you retest after Brunos fix?

ldh4:
> @ldh4 did you retest after Brunos fix?

Just retested after pulling from this branch.
I now get this error when building tests with rocm 6.2 with `AMD_GFX942_APU`:
```
/kokkos/core/src/HIP/Kokkos_HIP_Space.cpp:61:17: error: out-of-line definition of 'allocate' does not match any declaration in 'Kokkos::HIPSpace'
   61 | void* HIPSpace::allocate(const HIP& exec_space,
      |                 ^~~~~~~~
/kokkos/core/src/HIP/Kokkos_HIP_Space.cpp:66:17: error: out-of-line definition of 'allocate' does not match any declaration in 'Kokkos::HIPSpace'
   66 | void* HIPSpace::allocate(const HIP& exec_space, const char* arg_label,
      |                 ^~~~~~~~
```

Rombur:
@ldh4 can you try again please

algorithms/src/sorting/impl/Kokkos_SortImpl.hpp
```diff
@@ -412,7 +412,7 @@ sort_device_view_with_comparator(
 // Note with HIP unified memory this code path is still the right thing to do
 // if we end up here when RocThrust is not enabled.
 // The create_mirror_view_and_copy will do the right thing (no copy).
-#ifndef KOKKOS_ENABLE_IMPL_HIP_UNIFIED_MEMORY
+#ifndef KOKKOS_IMPL_HIP_UNIFIED_MEMORY
```
ldh4, line 415: Can you move the `memspace` type alias above to be in this `ifndef`?
I'm getting warnings on this:

```
/kokkos/algorithms/unit_tests/../src/sorting/./impl/Kokkos_SortImpl.hpp:411:9: warning: unused type alias 'MemSpace' [-Wunused-local-typedef]
  411 |   using MemSpace = typename ViewType::memory_space;
      |         ^
 ```

**Review by ldh4 — COMMENTED** [review]
All tests built, but there were two failures:
```
4 - Kokkos_CoreUnitTest_Serial2 (Failed)
32 - Kokkos_ContainersUnitTest_HIP (Failed)

[ RUN      ] serial.deep_copy_zero_memset
Test failure: Didn't encounter wanted events
/kokkos/core/unit_test/TestWithoutInitializing.hpp:245: Failure
Value of: success
  Actual: false
Expected: true
[  FAILED  ] serial.deep_copy_zero_memset (0 ms)

[ RUN      ] hip.create_mirror_view_and_copy_dynamicview
Test failure: encountered unwanted events
  Found fence event
BeginFence { "DynamicView::resize_serial: Fence after copying chunks to the device", 50331649,105}
EndFence { 105}
BeginFence { "Kokkos::Tools::invoke_kokkosp_callback: Kokkos Profile Tool Fence", 0,106}
EndFence { 106}
BeginFence { "Kokkos::Tools::invoke_kokkosp_callback: Kokkos Profile Tool Fence", 50331648,107}
EndFence { 107}
BeginParallelFor { "Kokkos::ViewCopy-1D", 1,108}
BeginFence { "Kokkos::Tools::invoke_kokkosp_callback: Kokkos Profile Tool Fence", 0,109}
EndFence { 109}
BeginFence { "Kokkos::Tools::invoke_kokkosp_callback: Kokkos Profile Tool Fence", 50331648,110}
EndFence { 110}
EndParallelFor { 108}
/kokkos/containers/unit_tests/TestWithoutInitializing.hpp:705: Failure
Value of: success
  Actual: false
Expected: true
[  FAILED  ] hip.create_mirror_view_and_copy_dynamicview (0 ms)
```

File: algorithms/src/sorting/impl/Kokkos_SortImpl.hpp (hunk 9 above)
Comment (Rombur) [reply algorithms/src/sorting/impl/Kokkos_SortImpl.hpp:415]: Yes, I'll do it.

**Comment by Rombur** [comment]
@ldh4 can you check that these two tests pass on develop. I think that they should also fail if you enable unified memory there.

**Review by ldh4 — COMMENTED** [review]
@Rombur The two failing tests fail on develop branch as well with `-DKokkos_ENABLE_IMPL_HIP_UNIFIED_MEMORY=ON` (I had to apply a similar change to [5940680]; otherwise, the develop branch currently doesn't build).

I made a few changes to the failing tests to make them pass on MI300A, but it seems that I can't push a commit to this branch. Here's the diff:
```
diff --git a/containers/unit_tests/TestWithoutInitializing.hpp b/containers/unit_tests/TestWithoutInitializing.hpp
index d9c2aaa4a..293289855 100644
--- a/containers/unit_tests/TestWithoutInitializing.hpp
+++ b/containers/unit_tests/TestWithoutInitializing.hpp
@@ -44,6 +44,12 @@
                                Kokkos::CudaSpace>)                       \
     GTEST_SKIP() << "skipping since unified memory requires additional " \
                     "fences";
+#elif defined(KOKKOS_IMPL_HIP_UNIFIED_MEMORY)
+#define GTEST_SKIP_IF_UNIFIED_MEMORY_SPACE                               \
+  if constexpr (std::is_same_v<typename TEST_EXECSPACE::memory_space,    \
+                               Kokkos::HIPSpace>)                        \
+    GTEST_SKIP() << "skipping since unified memory requires additional " \
+                    "fences";
 #else
 #define GTEST_SKIP_IF_UNIFIED_MEMORY_SPACE
 #endif
diff --git a/core/unit_test/TestWithoutInitializing.hpp b/core/unit_test/TestWithoutInitializing.hpp
index 87cd77c62..da8ed9d50 100644
--- a/core/unit_test/TestWithoutInitializing.hpp
+++ b/core/unit_test/TestWithoutInitializing.hpp
@@ -222,26 +222,29 @@ TEST(TEST_CATEGORY, deep_copy_zero_memset) {
   Kokkos::View<int*, TEST_EXECSPACE> bla("bla", 8);
 
   // for MI300A with unified memory, ZeroMemset uses a parallel for
-  auto success =
-#if defined(KOKKOS_IMPL_HIP_UNIFIED_MEMORY)
-      validate_existence(
-          [&]() { Kokkos::deep_copy(bla, 0); },
-          [&](BeginParallelForEvent e) {
-            const bool found =
-                (e.descriptor().find("Kokkos::ZeroMemset via parallel_for") !=
-                 std::string::npos);
-            return MatchDiagnostic{found,
-                                   {"Found expected parallel_for label"}};
-          });
-#else
-      validate_absence([&]() { Kokkos::deep_copy(bla, 0); },
-                       [&](BeginParallelForEvent) {
-                         return MatchDiagnostic{true, {"Found begin event"}};
-                       },
-                       [&](EndParallelForEvent) {
-                         return MatchDiagnostic{true, {"Found end event"}};
-                       });
+  auto success = false;
+#ifdef KOKKOS_IMPL_HIP_UNIFIED_MEMORY
+  if constexpr (!std::is_same_v<TEST_EXECSPACE::memory_space,
+                                Kokkos::HostSpace>)
+    success = validate_existence(
+        [&]() { Kokkos::deep_copy(bla, 0); },
+        [&](BeginParallelForEvent e) {
+          const bool found =
+              (e.descriptor().find("Kokkos::ZeroMemset via parallel_for") !=
+               std::string::npos);
+          return MatchDiagnostic{found, {"Found expected parallel_for label"}};
+        });
+  else
 #endif
+    success =
+        validate_absence([&]() { Kokkos::deep_copy(bla, 0); },
+                         [&](BeginParallelForEvent) {
+                           return MatchDiagnostic{true, {"Found begin event"}};
+                         },
+                         [&](EndParallelForEvent) {
+                           return MatchDiagnostic{true, {"Found end event"}};
+                         });
+
   ASSERT_TRUE(success);
   listen_tool_events(Config::DisableAll());
 }
 ```

Rombur:
>  it seems that I can't push a commit to this branch

I don't know why. You should be able to do it. Anyway, I've pushed your patch.

crtrott:
Retest this please.

crtrott (approved):

PR #7468, opened 2024-10-18
Update external desul version SHA for CI

crtrott:
This is needed for the View refactor, where we rely on a fix in AtomicRef


masterleinad (approved):

dalg24 (approved):
Sure

PR #7469, opened 2024-10-18
Split TestHalfOperators kernels to improve compilation time with SYCL

masterleinad:
This improved the compilation time of the tests by 60 min (to around 20 min). The runtime of the kernel is negligible.

### DIFF
--- a/core/unit_test/TestHalfOperators.hpp
+++ b/core/unit_test/TestHalfOperators.hpp
@@ -419,7 +419,7 @@ struct Functor_TestHalfOperators {
     float tmp_s_lhs;
     half_impl_type half_tmp;
 
-    // Initialze output views to catch missing test invocations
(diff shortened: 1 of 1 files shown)


crtrott:
So its not even splitting the test into multiple object files its just splitting the kernel 5 ways?

crtrott (approved):
Not sure I am loving the approach but I also don't immediately have a better idea for low overhead (in terms of lines of code) partitioning of the test.

masterleinad:
> So its not even splitting the test into multiple object files its just splitting the kernel 5 ways?

Yes, since we can compile with `-fsycl-device-code-split=per_kernel` that does the trick.

dalg24:
Looks like the build on the spack gltlab was significantly faster but I did not see improvement on the Jenkins build

dalg24 (approved):

PR #7470, opened 2024-10-19
CI: Gather test results

masterleinad:
This pull request makes the unit test results visible in the respective sections for the `jenkins` and `GItLab` CI.

Examples in https://cloud.cees.ornl.gov/jenkins-ci/job/Kokkos/job/PR-7470/8/testReport/ and https://gitlab.spack.io/kokkos/kokkos/-/pipelines/863294/test_report?job_name=NVIDIA-GH200.

### DIFF
--- a/.jenkins
+++ b/.jenkins
@@ -211,7 +211,6 @@ pipeline {
                     }
                     post {
                         always {
-                            sh 'ccache --show-stats'
                             xunit([CTest(deleteOutputFiles: true, failIfNotNew: true, pattern: 'build/Testing/**/Test.xml', skipNoTestFiles: false, stopProcessingIfError: true)])
                         }
                     }
@@ -253,7 +252,6 @@ pipeline {
                     }
                     post {
                         always {
-                            sh 'ccache --show-stats'
                             xunit([CTest(deleteOutputFiles: true, failIfNotNew: true, pattern: 'build/Testing/**/Test.xml', skipNoTestFiles: false, stopProcessingIfError: true)])
                         }
                     }
@@ -533,7 +531,6 @@ pipeline {
                     post {
                         always {
                             sh 'ccache --show-stats'
-                            xunit([CTest(deleteOutputFiles: true, failIfNotNew: true, pattern: 'build/Testing/**/Test.xml', skipNoTestFiles: false, stopProcessingIfError: true)])
                         }
                     }
                 }
--- a/.jenkins
+++ b/.jenkins
@@ -166,7 +166,7 @@ pipeline {
                                 -DCMAKE_CXX_FLAGS=-Werror \
                                 -DCMAKE_CXX_STANDARD=17 \
                               .. && \
-                              make -j8 && ctest -T Test --verbose && \
+                              make -j8 && ctest --verbose && \
                               cd ../.. && \
                               cmake -B build_cmake_installed_different_compiler/build -DCMAKE_CXX_COMPILER=clang++ -DCMAKE_CXX_FLAGS=-Werror -DCMAKE_CXX_STANDARD=17 build_cmake_installed_different_compiler && \
                               cmake --build build_cmake_installed_different_compiler/build --target all && \
@@ -175,7 +175,7 @@ pipeline {
                     post {
                         always {
                             sh 'ccache --show-stats'
-                            xunit([CTest(deleteOutputFiles: true, failIfNotNew: true, pattern: 'build/Testing/**/Test.xml', skipNoTestFiles: false, stopProcessingIfError: true)])
+                            xunit([CTest(deleteOutputFiles: true, failIfNotNew: true, pattern: 'build-tests/Testing/**/Test.xml', skipNoTestFiles: false, stopProcessingIfError: true)])
                         }
                     }
                 }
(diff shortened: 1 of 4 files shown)


masterleinad:
> For the gitlab ci I do see a Tests tab with statistics, that said the log output has this:

https://gitlab.spack.io/kokkos/kokkos/-/pipelines/862573/test_report looks fine to me. No idea why it needed three attempts to succeed uploading the artifact. The Intel GPU build succeeded on the first try.

crtrott:
Question: what do we gain with this? I looked at the test overview thing (in both cases) and its not clear that it would give you an easy way to get to the test-output in case of failure (nothing failed so maybe that is why?). I.e. if it only gives you a number or even a list of failed tests, without being able to link back to the output of the test I am not sure that we gain anything useful, since you would still need to click on the raw log file and scroll through it to see error report.

masterleinad:
https://gitlab.spack.io/kokkos/kokkos/-/pipelines/863294/test_report?job_name=NVIDIA-GH200 has useful results and I'm used to seeing similar output for jenkins but that currently doesn't look quite right (https://cloud.cees.ornl.gov/jenkins-ci/job/Kokkos/job/PR-7470/7/testReport/). On the other hand, even having the list of failed tests there is more useful than needing to go through all the logs in my opinion.

masterleinad:
Now, the test reports on `jenkins` are also good, see https://cloud.cees.ornl.gov/jenkins-ci/job/Kokkos/job/PR-7470/8/testReport/.

crtrott (approved):
Yeah that seems like useful output now :-)

Rombur (approved):

.jenkins
```diff
@@ -91,11 +97,12 @@ pipeline {
                                 -DKokkos_ENABLE_BENCHMARKS=ON \
                                 -DKokkos_ENABLE_HIP=ON \
                               .. && \
-                              make -j8 && ctest --verbose'''
+                              make -j8 && ctest --no-compress-output -T Test --verbose'''
```
dalg24, line 100: Did you look at the size of these XML files that are produced as compared to the size of the logs that are uploaded to the server?
Should we be worried about disk space?

masterleinad: The `xml` files are between 56 KiB and 128 KiB per report:
- https://cloud.cees.ornl.gov/jenkins-ci/job/Kokkos/job/PR-7470/9/execution/node/179/ws/build/Testing/20241022-1416/ 56.07 KiB
- https://cloud.cees.ornl.gov/jenkins-ci/job/Kokkos/job/PR-7470/9/execution/node/48/ws/build/Testing/20241022-1330/ 104.55 KiB
- https://cloud.cees.ornl.gov/jenkins-ci/job/Kokkos/job/PR-7470/9/execution/node/180/ws/build/Testing/20241022-1511/ 127.70 KiB
- https://cloud.cees.ornl.gov/jenkins-ci/job/Kokkos/job/PR-7470/9/execution/node/183/ws/build/Testing/20241022-1423/ 117.89 KiB
- https://cloud.cees.ornl.gov/jenkins-ci/job/Kokkos/job/PR-7470/9/execution/node/184/ws/build/Testing/20241022-1404/ 111.32 KiB
- https://cloud.cees.ornl.gov/jenkins-ci/job/Kokkos/job/PR-7470/9/execution/node/185/ws/build/Testing/20241022-1435/ 113.52 KiB

That's negligible compared with the size of the build directories, KokkosCore_UnitTest_Cuda alone is 154.53 MiB in https://cloud.cees.ornl.gov/jenkins-ci/job/Kokkos/job/PR-7470/9/execution/node/186/ws/build/core/unit_test.

The size of the whole test output for that build is 4.3 MiB

.jenkins
```diff
@@ -62,7 +62,13 @@ pipeline {
                                 -DKokkos_ENABLE_LIBQUADMATH=ON \
                                 -DKokkos_ENABLE_SERIAL=ON \
                               .. && \
-                              make -j8 && ctest --verbose && gcc -I$PWD/../core/src/ ../core/unit_test/tools/TestCInterface.c'''
+                              make -j8 && ctest --no-compress-output -T Test --verbose && gcc -I$PWD/../core/src/ ../core/unit_test/tools/TestCInterface.c'''
+                    }
+                    post {
+                        always {
+                            sh 'ccache --show-stats'
```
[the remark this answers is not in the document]
masterleinad: Fixed. There are still various builds where we zero stats and reports stats that don't use `ccache`. If we think it matters, we can fix elsewhere.

.jenkins
dalg24: > negligible compared with the size of the build directories

Build directories are not uploaded to the server.  I suspect these xml files are.

masterleinad: So we are talking about a maximum of 128*10 KiB or so in `xml` files which is around a quarter of the log file.

dalg24: Should we trade --verbose for --output-on-failure?
One downside is we would not be able to follow progress on a single test :/
What do you think?

dalg24 (commented):
I am fine to merge as is.  You might want to investigate using the Jenkins JUnit support to simplify the configuration

PR #7472, opened 2024-10-21
simd: refactor simd classes and hide simd_abi

ldh4:
This PR makes following changes in kokkos simd to keep its interface more aligned with [P1928R12](https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2024/p1928r12.pdf):

- Renames `simd` classes to `basic_simd` (Section 1.6)
- Renames `simd_mask` classes to `basic_simd_mask` (Section 1.6)
- Hides/deprecates `simd` interface that exposes `simd_abi` (Section 4.1)
- Introduces a `simd<T, N>` type alias where `T` is data type and `N` is vector size (Section 4.1)

- Replaced the type alias `Kokkos::Experimental::simd<T, simd_abi>` with `Kokkos::Experimental::simd<T, N>`

### DIFF
--- a/simd/src/Kokkos_SIMD.hpp
+++ b/simd/src/Kokkos_SIMD.hpp
@@ -84,13 +84,24 @@ namespace simd_abi {
 namespace Impl {
 
 #if defined(KOKKOS_ARCH_AVX512XEON)
-using host_native = avx512_fixed_size<8>;
+using host_fixed_native = avx512_fixed_size<8>;
+template <int N>
+using host_native_abi = avx512_fixed_size<N>;
+
 #elif defined(KOKKOS_ARCH_AVX2)
-using host_native  = avx2_fixed_size<4>;
+using host_fixed_native = avx2_fixed_size<4>;
+template <int N>
+using host_native_abi = avx2_fixed_size<N>;
+
 #elif defined(KOKKOS_ARCH_ARM_NEON)
-using host_native  = neon_fixed_size<2>;
+using host_fixed_native = neon_fixed_size<2>;
+template <int N>
+using host_native_abi = neon_fixed_size<N>;
+
 #else
-using host_native   = scalar;
+using host_fixed_native = scalar;
+template <int N>
+using host_native_abi = scalar;
 #endif
 
 template <class T>
@@ -99,7 +110,10 @@ struct ForSpace;
 #ifdef KOKKOS_ENABLE_SERIAL
 template <>
 struct ForSpace<Kokkos::Serial> {
-  using type = host_native;
+  using type = host_fixed_native;
+
+  template <int N>
+  using simd_abi = host_native_abi<N>;
 };
 #endif
 
@@ -107,13 +121,19 @@ struct ForSpace<Kokkos::Serial> {
 template <>
 struct ForSpace<Kokkos::Cuda> {
   using type = scalar;
+
+  template <int N>
+  using simd_abi = scalar;
 };
 #endif
 
 #ifdef KOKKOS_ENABLE_THREADS
 template <>
 struct ForSpace<Kokkos::Threads> {
-  using type = host_native;
+  using type = host_fixed_native;
+
+  template <int N>
+  using simd_abi = host_native_abi<N>;
 };
 #endif
 
@@ -121,13 +141,19 @@ struct ForSpace<Kokkos::Threads> {
 template <>
 struct ForSpace<Kokkos::Experimental::HPX> {
   using type = scalar;
+
+  template <int N>
+  using simd_abi = scalar;
 };
 #endif
 
 #ifdef KOKKOS_ENABLE_OPENMP
 template <>
 struct ForSpace<Kokkos::OpenMP> {
-  using type = host_native;
+  using type = host_fixed_native;
+
+  template <int N>
+  using simd_abi = host_native_abi<N>;
 };
 #endif
 
@@ -135,6 +161,9 @@ struct ForSpace<Kokkos::OpenMP> {
 template <>
 struct ForSpace<Kokkos::Experimental::OpenMPTarget> {
   using type = scalar;
+
+  template <int N>
+  using simd_abi = scalar;
 };
 #endif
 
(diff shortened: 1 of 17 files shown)


simd/src/Kokkos_SIMD.hpp
```diff
@@ -99,83 +110,143 @@ struct ForSpace;
...
+
+template <class T, class Abi>
+using simd_mask KOKKOS_DEPRECATED_WITH_COMMENT(
+    "Temporarily use Impl::simd_mask<T> instead") = basic_simd_mask<T, Abi>;
+#endif
 
 namespace Impl {
 
+template <class T, int N = 0>
+using simd =
```
nmm0, line 240: I don't quite get why these are in Impl?

nmm0 (approved):
Looks good to me! I'm only a bit iffy about the Impl::simd thing but if that's what was already decided on let's just go with it

simd/src/Kokkos_SIMD.hpp
ldh4: This is the part where I'm looking for some more opinions from reviewers.

In P1928, `simd<T, Abi>` is removed and `simd<T, N = native-size>` became an alias for `basic_simd<T, Abi>`.
However, in Kokkos simd, replacing `simd<T, Abi>` with `simd<T, N = native-size>` first needs to go through a deprecation. Until `simd<T, Abi>` is fully deprecate and removed, I thought `simd<T, N = native-size>` could be introduced in the `Impl` namespace temporarily and later be moved out of impl.

simd/src/Kokkos_SIMD.hpp
```diff
@@ -99,83 +110,143 @@ struct ForSpace;
...
 template <class Space>
 using ForSpace = typename Impl::ForSpace<typename Space::execution_space>::type;
 
 template <class T>
 using native = ForSpace<Kokkos::DefaultExecutionSpace>;
+#endif
 
 }  // namespace simd_abi
 
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
```
tcclevenger, line 218: Do we need this deprecation guards if all the typedefs have `KOKKOS_DEPRECATED_WITH_COMMENT`?

simd/src/Kokkos_SIMD.hpp
```diff
@@ -99,83 +110,143 @@ struct ForSpace;
...
 
+template <class Space = Kokkos::DefaultExecutionSpace>
+using native_fixed_abi = typename ForSpace<Space>::type;
+
+template <int N, class Space = Kokkos::DefaultExecutionSpace>
+using native_abi = typename ForSpace<Space>::template simd_abi<N>;
+
 }  // namespace Impl
 
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
```
tcclevenger, line 208: Instead of this guard, could we just add `KOKKOS_DEPRRECATED` for the `using` statements like you do below?

ldh4: That was my initial approach, but placing the `deprecated` attributes on these aliases would cause every simd compilation to output deprecation errors, because both `simd_abi::ForSpace` and `simd_abi::native` are used for `native_simd` and `native_simd_mask`, which are now also deprecated.

simd/src/Kokkos_SIMD.hpp
ldh4: I can see that we don't necessarily need it, but I think we could still have this deprecation guard. To some extent, it serves as an indicator for us to quickly grep for and remove by the end of Kokkos 4.0.

simd/src/Kokkos_SIMD.hpp
janciesko: I think that's ok.

janciesko (approved):
This looks good to me.

nliber (approved):

ldh4:
Retest this please.

simd/src/Kokkos_SIMD.hpp
JBludau: I would prefer we do not promote people actively using anything that is in `Impl`. Is there no other way?

nmm0: I guess one question is how many people would we be actively breaking here if we didn't put this in Impl? Aren't our customers still using an old version of this (and anyway it's in Experimental)? I guess it's some of the same arguments as removing reference

ldh4: > I would prefer we do not promote people actively using anything that is in `Impl`. Is there no other way?

I think an alternative approach could be to name the new alias to be something else temporarily and place it in the `Kokkos::Experimental` namespace. Then, once `simd<T, Abi>` is fully deprecated and removed, introduce `simd<T, N>`, followed by the deprecation of the temporary alias.

> I guess one question is how many people would we be actively breaking here if we didn't put this in Impl? Aren't our customers still using an old version of this (and anyway it's in Experimental)?

I think the most common use of Kokkos simd is through `native_simd<...>` (which is being deprecated here), but it's still been valid to use `simd<T, Abi>` for specifying simd vector width.
Although very tempting, I'm a bit hesitant to skip the deprecation step and replace `simd<T, Abi>` with `simd<T, N>` right away.

JBludau: > it's still been valid to use `simd<T, Abi>` for specifying simd vector width. Although very tempting, I'm a bit hesitant to skip the deprecation step and replace `simd<T, Abi>` with `simd<T, N>` right away.

hmm ... as it is still in `Experimental` I would actually prefer to make it compliant with the P1928 and not have a state where we tell people to use something out of `Impl`. Especially since it would break user code twice.

ldh4: Removed `simd<T, Abi>` from `Kokkos::Experimental` namespace and replaced it with `simd<T, N>` without deprecation as discussed.

JBludau: Hmm ... can we add some `simd` and `simd_mask` in some tests

JBludau (approved):
lgtm

ldh4:
Retest this please.

ldh4:
Retest this please.

PR #7473, opened 2024-10-21
Fix Kokkos::realloc for View of Views

### LINKED ISSUE #7466: SequentialHostInit issues with resize and realloc
```diff
diff --git a/core/unit_test/TestViewOfViews.hpp b/core/unit_test/TestViewOfViews.hpp
index 1d53bca33..2311fe93a 100644
--- a/core/unit_test/TestViewOfViews.hpp
+++ b/core/unit_test/TestViewOfViews.hpp
@@ -23,12 +23,12 @@ namespace {
 // User-defined types with a View data member
 template <class V>
 class S {
-  V v_;
+  V v_ = V("v", 5);
 
  public:
   template <class... Extents>
   S(std::string label, Extents... extents) : v_(std::move(label), extents...) {}
-  KOKKOS_DEFAULTED_FUNCTION S() = default;
+  S() = default;
 };
 
 template <class V>
@@ -99,6 +99,8 @@ void test_view_of_views_sequential_host_init() {
   vov(0, 0) = a;
   vov(1, 0) = a;
   vov(0, 1) = b;
+
+  Kokkos::resize(Kokkos::view_alloc(Kokkos::SequentialHostInit),vov,1,1);
 }
 
 TEST(TEST_CATEGORY, view_of_views_default) {
```
triggers a deadlock and we see the same problem in #7456. The issue is that tracking is enabled for the inside view and copying elements inside resize (via `parallel_for`) causes `deallocate` calls.

For `realloc` the problem is a call to `deep_copy(v, typename V::value_type{})` which is intended to copy the scalar value to all elements of the View given. If the `value_type` is a `View` however, we try to copy every element of the source View to the destination View which of course fails miserably since neither the range nor the value type of the two Views match.

masterleinad:
Fixes #7466 partly. We must make sure to select the overload copying from/to a scalar for
```
deep_copy(View<View<Scalar>>, View<Scalar>);
```
and not the ones copying from one view to another view of a similar type.

### DIFF
--- a/core/src/Kokkos_CopyViews.hpp
+++ b/core/src/Kokkos_CopyViews.hpp
@@ -958,6 +958,71 @@ void contiguous_fill_or_memset(
 
   contiguous_fill_or_memset(exec_space_type(), dst, value);
 }
+
+template <class ViewType, class ValueType, int Dim = 0>
+struct ViewSequentialHostFill {
+  const ViewType& dst;
+  const ValueType& value;
+
+  template <class... Indices>
+  void apply(Indices... idx) const {
+    if constexpr (Dim == int(ViewType::rank)) {
+      if constexpr (ViewType::rank == 0) {
+        dst() = value;
+      } else {
+        dst(idx...) = value;
+      }
+    } else {
+      for (size_t i = 0; i < dst.extent(Dim); ++i) {
+        ViewSequentialHostFill<ViewType, ValueType, Dim + 1>{dst, value}.apply(
+            idx..., i);
+      }
+    }
+  }
+
+  void operator()() const { apply(); }
+};
+
+template <class ViewType>
+void sequential_host_fill(const ViewType& dst,
+                          const typename ViewType::const_value_type& value) {
+  static_assert(
+      Kokkos::SpaceAccessibility<Kokkos::HostSpace,
+                                 typename ViewType::memory_space>::accessible,
+      "sequential_host_fill requires host-accessible memory");
+  static_assert(std::is_same_v<typename ViewType::non_const_value_type,
+                               typename ViewType::value_type>,
+                "deep_copy requires non-const destination type");
+
+  if (dst.data() == nullptr || dst.span() == 0) {
+    return;
+  }
+
+  if constexpr (ViewType::rank == 0) {
+    dst() = value;
+    return;
+  }
+
+  if (dst.span_is_contiguous() && !ViewType::traits::impl_is_customized) {
+    using value_type      = typename ViewType::value_type;
+    value_type* const ptr = dst.data();
+    const size_t n        = dst.span();
+    if constexpr (std::is_trivially_copyable_v<value_type>) {
+      if (has_all_zero_bits(value)) {
+        std::memset(static_cast<void*>(ptr), 0, n * sizeof(value_type));
+        return;
+      }
+    }
+    for (size_t i = 0; i < n; ++i) {
+      ptr[i] = value;
+    }
+    return;
+  }
+
+  ViewSequentialHostFill<ViewType, typename ViewType::const_value_type> fill{
+      dst, value};
+  fill();
+}
 }  // namespace Impl
 
 /** \brief  Deep copy a value from Host memory into a view.  */
@@ -1199,13 +1264,15 @@ inline void deep_copy(
 template <class DT, class... DP, class ST, class... SP>
 inline void deep_copy(
     const View<DT, DP...>& dst, const View<ST, SP...>& src,
-    std::enable_if_t<
-        (std::is_void_v<typename ViewTraits<DT, DP...>::specialize> &&
-         std::is_void_v<typename ViewTraits<ST, SP...>::specialize> &&
-         (is_view_v<typename View<DT, DP...>::value_type> ==
-          is_view_v<typename View<ST, SP...>::value_type>)&&  //
-         (unsigned(ViewTraits<DT, DP...>::rank) != 0 ||
-          unsigned(ViewTraits<ST, SP...>::rank) != 0))>* = nullptr) {
+    std::enable_if_t<(
+        std::is_void_v<typename ViewTraits<DT, DP...>::specialize> &&
+        std::is_void_v<typename ViewTraits<ST, SP...>::specialize> &&
+        (is_view_v<typename View<DT, DP...>::value_type> ==
+         is_view_v<typename View<
+             ST, SP...>::value_type>)&&(unsigned(ViewTraits<DT, DP...>::rank) !=
+                                            0 ||
+                                        unsigned(ViewTraits<ST, SP...>::rank) !=
+                                            0))>* = nullptr) {
   Impl::check_deep_copy_view_arguments_are_distinct(std::addressof(dst),
                                                     std::addressof(src));
 
@@ -2408,14 +2475,16 @@ template <class ExecSpace, class DT, class... DP, class ST, class... SP>
 inline void deep_copy(
     const ExecSpace& exec_space, const View<DT, DP...>& dst,
     const View<ST, SP...>& src,
-    std::enable_if_t<
-        (Kokkos::is_execution_space<ExecSpace>::value &&
-         std::is_void_v<typename ViewTraits<DT, DP...>::specialize> &&
-         std::is_void_v<typename ViewTraits<ST, SP...>::specialize> &&
-         (is_view_v<typename View<DT, DP...>::value_type> ==
-          is_view_v<typename View<ST, SP...>::value_type>)&&  //
-         (unsigned(ViewTraits<DT, DP...>::rank) != 0 ||
-          unsigned(ViewTraits<ST, SP...>::rank) != 0))>* = nullptr) {
+    std::enable_if_t<(
+        Kokkos::is_execution_space<ExecSpace>::value &&
+        std::is_void_v<typename ViewTraits<DT, DP...>::specialize> &&
+        std::is_void_v<typename ViewTraits<ST, SP...>::specialize> &&
+        (is_view_v<typename View<DT, DP...>::value_type> ==
+         is_view_v<typename View<
+             ST, SP...>::value_type>)&&(unsigned(ViewTraits<DT, DP...>::rank) !=
+                                            0 ||
+                                        unsigned(ViewTraits<ST, SP...>::rank) !=
+                                            0))>* = nullptr) {
   Impl::check_deep_copy_view_arguments_are_distinct(std::addressof(dst),
                                                     std::addressof(src));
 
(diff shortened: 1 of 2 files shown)


dalg24 (changes requested):

core/unit_test/TestViewOfViews.hpp
```diff
@@ -126,4 +126,107 @@ TEST(TEST_CATEGORY, test_view_of_views_sequential_host_init) {
...
+  {
+    using VoV = Kokkos::View<V**, Kokkos::HostSpace>;
+    VoV vov(Kokkos::view_alloc("vov", Kokkos::SequentialHostInit), 2, 2);
+    {
+      V a("a");
+      vov(0, 0) = a;
+    }
+    Kokkos::realloc(Kokkos::view_alloc(Kokkos::SequentialHostInit), vov, 1, 1);
+    ASSERT_EQ(vov.size(), 1u);
+  }
```
dalg24, line 150: These add a lot of boiler plate and it is unclear how this coverage is important.

dalg24 (commented):
Did you consider realloc to a larger size and realloc to current size?

core/unit_test/TestViewOfViews.hpp
masterleinad: The internal implementation could depend on the rank (like it does for `resize`).

masterleinad:
> Did you consider realloc to a larger size and realloc to current size?

It doesn't seem to add much in my opinion. We don't do anything if it's the same size (which is tested elsewhere) and does the same thing whether we request a larger or smaller size. Checking for different ranks seems more important in comparison.

janciesko:
Retest this please

janciesko (approved):
This is a relevant fix

janciesko:
Additional bug: For the case that the size of the VoV stays the same, we keep the allocation and call deep_copy(v, value_type{}),  which ends of using calling a pF in ViewFill which hits that mutex and then deadlocks. We probably have to add a path for that case and use something like a SequentialHostFill.

masterleinad:
> Additional bug: For the case that the size of the VoV stays the same, we keep the allocation and call deep_copy(v, value_type{}), which ends of using calling a pF in ViewFill which hits that mutex and then deadlocks. We probably have to add a path for that case and use something like a SequentialHostFill.

Doesn't https://github.com/kokkos/kokkos/pull/7514 fix that?

janciesko:
Doesn't seem like it, #7514 doesn't touch on impl_realloc.

> > Additional bug: For the case that the size of the VoV stays the same, we keep the allocation and call deep_copy(v, value_type{}), which ends of using calling a pF in ViewFill which hits that mutex and then deadlocks. We probably have to add a path for that case and use something like a SequentialHostFill.
> 
> Doesn't #7514 fix that?

janciesko:
For now I'd prefer to have this PR fixed and then rebase https://github.com/kokkos/kokkos/pull/7514 if neccessary

core/src/Kokkos_CopyViews.hpp
```diff
@@ -1199,11 +1264,15 @@ inline void deep_copy(
...
-          unsigned(ViewTraits<ST, SP...>::rank) != 0))>* = nullptr) {
+    std::enable_if_t<(
+        std::is_void_v<typename ViewTraits<DT, DP...>::specialize> &&
+        std::is_void_v<typename ViewTraits<ST, SP...>::specialize> &&
+        (is_view_v<typename View<DT, DP...>::value_type> ==
+         is_view_v<typename View<
+             ST, SP...>::value_type>)&&(unsigned(ViewTraits<DT, DP...>::rank) !=
+                                            0 ||
+                                        unsigned(ViewTraits<ST, SP...>::rank) !=
+                                            0))>* = nullptr) {
```
masterleinad: We don't need the changes here...

core/src/Kokkos_CopyViews.hpp
```diff
@@ -2406,12 +2475,16 @@ template <class ExecSpace, class DT, class... DP, class ST, class... SP>
...
+    std::enable_if_t<(
+        Kokkos::is_execution_space<ExecSpace>::value &&
+        std::is_void_v<typename ViewTraits<DT, DP...>::specialize> &&
+        std::is_void_v<typename ViewTraits<ST, SP...>::specialize> &&
+        (is_view_v<typename View<DT, DP...>::value_type> ==
+         is_view_v<typename View<
+             ST, SP...>::value_type>)&&(unsigned(ViewTraits<DT, DP...>::rank) !=
+                                            0 ||
+                                        unsigned(ViewTraits<ST, SP...>::rank) !=
+                                            0))>* = nullptr) {
```
masterleinad: ... and here.

core/src/Kokkos_CopyViews.hpp
```diff
@@ -958,6 +958,71 @@ void contiguous_fill_or_memset(
...
+      if constexpr (ViewType::rank == 0) {
+        dst() = value;
+      } else {
+        dst(idx...) = value;
+      }
+    } else {
+      for (size_t i = 0; i < dst.extent(Dim); ++i) {
+        ViewSequentialHostFill<ViewType, ValueType, Dim + 1>{dst, value}.apply(
+            idx..., i);
+      }
```
masterleinad, line 978:
```suggestion
      ViewSequentialHostFill<ViewType, ValueType, Dim + 1> view_sequential_host_fill{dst, value}.
      for (size_t i = 0; i < dst.extent(Dim); ++i) {
        view_sequential_host_fill.apply(idx..., i);
      }
```

core/src/Kokkos_CopyViews.hpp
```diff
@@ -958,6 +958,71 @@ void contiguous_fill_or_memset(
...
+template <class ViewType, class ValueType, int Dim = 0>
+struct ViewSequentialHostFill {
+  const ViewType& dst;
+  const ValueType& value;
+
+  template <class... Indices>
+  void apply(Indices... idx) const {
+    if constexpr (Dim == int(ViewType::rank)) {
+      if constexpr (ViewType::rank == 0) {
+        dst() = value;
```
masterleinad: This case can be moved to the call operator since it never participates in the recursion. You also already handle that case in `sequential_host_fill`.

core/src/Kokkos_CopyViews.hpp
```diff
@@ -958,6 +958,71 @@ void contiguous_fill_or_memset(
...
+      }
+    } else {
+      for (size_t i = 0; i < dst.extent(Dim); ++i) {
+        ViewSequentialHostFill<ViewType, ValueType, Dim + 1>{dst, value}.apply(
+            idx..., i);
+      }
+    }
+  }
+
+  void operator()() const { apply(); }
```
masterleinad: Why not just use the constructor to call `apply`?

core/src/Kokkos_CopyViews.hpp
```diff
@@ -958,6 +958,71 @@ void contiguous_fill_or_memset(
...
+                "deep_copy requires non-const destination type");
+
+  if (dst.data() == nullptr || dst.span() == 0) {
+    return;
+  }
+
+  if constexpr (ViewType::rank == 0) {
+    dst() = value;
+    return;
+  }
```
masterleinad: Since rank-0 views are always contiguous, this special case isn't really necessary, right?

core/src/Kokkos_CopyViews.hpp
janciesko: @masterleinad I removed the trailing `//` in your previous commit and clang-format updated the formatting.

core/src/Kokkos_CopyViews.hpp
```diff
@@ -2408,14 +2475,16 @@ template <class ExecSpace, class DT, class... DP, class ST, class... SP>
 inline void deep_copy(
     const ExecSpace& exec_space, const View<DT, DP...>& dst,
     const View<ST, SP...>& src,
-    std::enable_if_t<
-        (Kokkos::is_execution_space<ExecSpace>::value &&
-         std::is_void_v<typename ViewTraits<DT, DP...>::specialize> &&
-         std::is_void_v<typename ViewTraits<ST, SP...>::specialize> &&
-         (is_view_v<typename View<DT, DP...>::value_type> ==
-          is_view_v<typename View<ST, SP...>::value_type>)&&  //
```
janciesko: @masterleinad Do you see the `//` ?

core/src/Kokkos_CopyViews.hpp
```diff
@@ -958,6 +958,71 @@ void contiguous_fill_or_memset(
...
+  if constexpr (ViewType::rank == 0) {
+    dst() = value;
+    return;
+  }
+
+  if (dst.span_is_contiguous() && !ViewType::traits::impl_is_customized) {
+    using value_type      = typename ViewType::value_type;
+    value_type* const ptr = dst.data();
+    const size_t n        = dst.span();
+    if constexpr (std::is_trivially_copyable_v<value_type>) {
```
ldh4, line 1001: Is this check for trivially copyable necessary?
The following operation in this block is checking if `value` is zero-bits and zeroing-out all values at `ptr`. Whether `value_type` is trivially copyable or not seems unrelated.

core/src/Kokkos_CopyViews.hpp
```diff
@@ -958,6 +958,71 @@ void contiguous_fill_or_memset(
...
+      if (has_all_zero_bits(value)) {
+        std::memset(static_cast<void*>(ptr), 0, n * sizeof(value_type));
+        return;
+      }
+    }
+    for (size_t i = 0; i < n; ++i) {
+      ptr[i] = value;
+    }
```
ldh4, line 1009: Maybe this is where the trivially copyable check at L1010 needs to be checked for.

core/src/Kokkos_CopyViews.hpp
masterleinad: Yes, but with your changes this doesn't need to be touched in the first place.

core/src/Kokkos_CopyViews.hpp
janciesko: @masterleinad Yes. Could you maybe make these changes and finish up the PR? 
The idea in general here is that we need to call a sequential host fill in `impl_realloc` for the case where sizes match and it is a VoV.

core/src/Kokkos_CopyViews.hpp
masterleinad: `has_all_zero_bits` requires `std::is_trivially_copyable_v`. Of course, we could also just always use `ViewSequentialHostFill` here.

janciesko: We could but the idea is to benefit from this distinction for performance.

core/src/Kokkos_CopyViews.hpp
janciesko: No, here we still want to support non-trivially copy-able objects which invoke a custom assignment operator but do not need to go through a more expensive indexing in `ViewSequentialHostFill`

janciesko:
I do not see a way to make this simpler or to apply the same idea as in https://github.com/kokkos/kokkos/pull/7514.

ldh4 (approved):

crtrott:
Want to delay this until after RC for 5.2 - this could have subtle downstream impacts.

masterleinad:
> Want to delay this until after RC for 5.2 - this could have subtle downstream impacts.

I mean it's an entire different code path so there shouldn't be an impact on any downstream users.

PR #7477, opened 2024-10-23
Fix AVX512 constexpr issue with GCC 8.5

crtrott:
I am also changing the SNL workflow to make sure we test AVX512 guaranteed (the machine we are running on has AVX512)

### DIFF
--- a/simd/src/Kokkos_SIMD_AVX512.hpp
+++ b/simd/src/Kokkos_SIMD_AVX512.hpp
@@ -30,6 +30,18 @@
     "Kokkos_SIMD_AVX512.hpp must be included before Kokkos_SIMD_Common_Math.hpp!"
 #endif
 
+// FIXME: Some intrinsics we are using are not marked constexpr in GCC
+// prior to some version. 8.5 fails 9.5 works.
+#ifdef KOKKOS_COMPILER_GNU
+#if KOKKOS_COMPILER_GNU < 950
+#define KOKKOS_IMPL_SIMD_CONSTEXPR
+#endif
+#endif
+
+#ifndef KOKKOS_IMPL_SIMD_CONSTEXPR
+#define KOKKOS_IMPL_SIMD_CONSTEXPR constexpr
+#endif
+
 namespace Kokkos {
 namespace Experimental {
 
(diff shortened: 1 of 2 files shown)


simd/src/Kokkos_SIMD_AVX512.hpp
```diff
@@ -30,6 +30,18 @@
...
+// FIXME: Some intrinsics we are using are not marked constexpr in GCC
+// prior to some version. 8.5 fails 9.5 works.
+#ifdef KOKKOS_COMPILER_GNU
+#if KOKKOS_COMPILER_GNU < 950
+#define KOKKOS_IMPL_SIMD_CONSTEXPR
+#endif
+#endif
+
+#ifndef KOKKOS_IMPL_SIMD_CONSTEXPR
+#define KOKKOS_IMPL_SIMD_CONSTEXPR constexpr
```
nliber, line 42: Should we `#undef` this at the end of the file?

nliber (changes requested):

nmm0 (approved):
Looks good to me once nevin's request is addressed

ldh4:
The constexpr-change was made to all simd backends in #6683. Should changes in this PR be applied to Scalar, AVX2 and NEON as well?

simd/src/Kokkos_SIMD_AVX512.hpp
```diff
@@ -30,6 +30,18 @@
...
+#endif
```
masterleinad, line 43:
```suggestion
#if defined(KOKKOS_COMPILER_GNU) && KOKKOS_COMPILER_GNU < 950
#define KOKKOS_IMPL_SIMD_CONSTEXPR
#else
#define KOKKOS_IMPL_SIMD_CONSTEXPR constexpr
#endif
```

masterleinad:
https://godbolt.org/z/vvnEWW5zb demonstrates that the issue is already resolved with 9.1. Removing the extra check for `none_of` shows that we would get a warning even with recent g++ versions, see https://godbolt.org/z/zTcG95cEz. Hence, I would drop the `constexpr` annotation altogether (unless we want to provide a `consteval` fall back).

simd/src/Kokkos_SIMD_AVX512.hpp
dalg24: Ping

masterleinad: Depends on https://github.com/kokkos/kokkos/pull/7486.

ldh4: `constexpr` changes won't be part of #7486 (I will cherry-pick 7445d06bc7a7239f66ca59dc135d989287150d40 for reliable AVX512 testing).
I plan on revisiting constexpr-ness of simd interfaces in a separate PR later.
So, I think this PR could be closed for now.

masterleinad:
constexpr changes won't be part of https://github.com/kokkos/kokkos/pull/7486

Issue #7412, opened 2024-10-07
Bug or feature? deep copying views that lie about their memory spaces


dalg24:
Consider this
```C++
size_t const N = 100;
float* d_x;
cudaMalloc(&d_x, N*sizeof(float));
auto* x = new float[N];
for (size_t i=0; i<N; ++i) x[i]=I;
using DeviceSpace = Kokkos::DefaultExecutionSpace::memory_space;
Kokkos::deep_copy(Kokkos::View<float*, Kokkos::HostSpace>(d_x, N),
    Kokkos::View<float*, DeviceSpace>(x, N)); // oops I swapped host and device memory spaces
```
When executing this code will not do what my code meant, it will just "work" because internally we pass `cudaMemcpyDefault` in our `cudaMemcpy` wrappers (instead of `cudaMemcpyDeviceToHost` like one could expect from the snippet above).

Can we get away with saying this is a precondition violation or should we consider this poor quality of implementation?
I am leaning towards we should have preferred `cudaMemcpyHostToDevice` in our internals.  Thoughts?

dalg24:
I forgot to say the HIP backend has the same behavior.
Note that if using the 3-argument overload of deep_copy() and passing an execution space, the error would be async and only caught (presumably) in the next runtime API call.

tpadioleau:
Not really addressing your main question but should we try to detect if the pointer is coherent with the memory space ? I think with CUDA we can do that checking the type in `cudaPointerAttributes` ? I don't know about other backends.

masterleinad:
> Not really addressing your main question but should we try to detect if the pointer is coherent with the memory space ? I think with CUDA we can do that checking the type in `cudaPointerAttributes` ? I don't know about other backends.

I think that's a good idea, at least with `Kokkos_ENABLE_DEBUG`.

crtrott:
Yeah as far as I remember the attribute query does add some latency, so it's something we could do in debug mode. Might be worthwhile to check whether not using cudaMemcpyDefault also improved performance.

dalg24:
I am going to interpret the feedback so far as "not a feature".

As far as the debug checks are concerned, ideally these would happen at (unmanaged) view construction but we don't want backend-specific code there and we currently don't have such thing as a memory space checking whether a pointer is "compatible" (we would need to define what this even mean, something like may have been allocated by that space instance).  Were you guys thinking to have these checks in the Impl::DeepCopy specializations only?

masterleinad:
[...]

Issue #7452, opened 2024-10-15
`print_configuration`: add kernel module information like `HMM` support


romintomasetti:
The `Kokkos::print_configuration` function mainly shows information from the "compile-time" configuration (target architectures, compilers and what not).

However, it lacks *runtime* information that are vital.

For instance, `Cuda` support for `HMM` might be very important to give as feedback (`cudaDevAttrPageableMemoryAccess`) because for the same compiled binaries, same hardware, if the driver is different (even same version but you change some kernel module parameters - see Gist below), **the code can behave differently**.

For instance, with `HMM` support (see https://gist.github.com/romintomasetti/b8472f574e1407096466e55aede8bfd7 for all necessary links and stuff), it is fine to use host pageable memory on the device without hitting the "expected" illegal read/write stopper.

So, with `Cuda` supporting `HMM`, building some unmanaged view from a host pointer will work:
```c++
std::vector<int> my_vector(size);
Kokkos::View<int*, Kokkos::CudaSpace> my_unmanaged(my_vector.data());
```

Some links extracted from the `Gist`:
* https://www.kernel.org/doc/html/v5.0/vm/hmm.html
* https://docs.nvidia.com/cuda/cuda-c-programming-guide/#system-requirements-for-unified-memory
* https://developer.nvidia.com/blog/simplifying-gpu-application-development-with-heterogeneous-memory-management/

cedricchevalier19:
Note that these capabilities are also available for architectures like GH200, without `HMM.`

Do you want to gather a list of properties we should check or at least print?

PR #7480, opened 2024-10-24
CUDA/HIP/SYCL: Check memory space in unmanaged View

masterleinad:
Addresses #7412. This pull requests distinguishes between
- `HostSpace`
- `Default::ExecutionSpace::memory_space`
- `SharedSpace`
- `SharedHostPinnedSpace`

~~and aborts if we pass in a pointer from a different memory space for the Cuda, HIP, and SYCL backend (since those have facilities to query a pointer for it's allocation mechanism). We can discuss if we want to be more lenient and make the checks follow `SpaceAccessibility::assignable`.~~ 
and aborts if the target space is not assignable from the detected space.

### DIFF
--- a/core/src/Cuda/Kokkos_CudaSpace.cpp
+++ b/core/src/Cuda/Kokkos_CudaSpace.cpp
@@ -482,6 +482,9 @@ void check_memory_space(cudaMemoryType deduced_memory_type) {
                                          RequestedMemorySpace>();
       return;
     case cudaMemoryTypeUnregistered:
+#ifdef KOKKOS_ENABLE_IMPL_CUDA_UNIFIED_MEMORY
+      return;
+#endif
       if (!Kokkos::SpaceAccessibility<RequestedMemorySpace,
                                       Kokkos::HostSpace>::assignable)
         abort_incompatible_memory_spaces<Kokkos::HostSpace,
--- a/core/src/Cuda/Kokkos_CudaSpace.cpp
+++ b/core/src/Cuda/Kokkos_CudaSpace.cpp
@@ -455,33 +455,40 @@ namespace {
 template <typename RequestedMemorySpace>
 void check_memory_space(cudaMemoryType deduced_memory_type) {
   switch (deduced_memory_type) {
-    case cudaMemoryTypeHost: 
-            if (!Kokkos::SpaceAccessibility<RequestedMemorySpace, Kokkos::CudaHostPinnedSpace>::assignable)
-                    Kokkos::abort(("Detected CudaHostPinnedSpace but requested incompatible " +
-                   std::string(RequestedMemorySpace::name())).c_str());
-            return;
-    case cudaMemoryTypeDevice: 
-            if (!Kokkos::SpaceAccessibility<RequestedMemorySpace, Kokkos::CudaSpace>::assignable)
-                    Kokkos::abort(("Detected CudaSpace but requested incompatible " +
-                   std::string(RequestedMemorySpace::name())).c_str());
-            return;
-    case cudaMemoryTypeManaged: 
-            if (!Kokkos::SpaceAccessibility<RequestedMemorySpace, Kokkos::CudaUVMSpace>::assignable)
-                    Kokkos::abort(("Detected CudaUVMSpace but requested incompatible " +
-                   std::string(RequestedMemorySpace::name())).c_str());
-            return;
-    case cudaMemoryTypeUnregistered: 
-            if (!Kokkos::SpaceAccessibility<RequestedMemorySpace, Kokkos::HostSpace>::assignable)
-                    Kokkos::abort(("Detected HostSpace but requested incompatible " +
-                   std::string(RequestedMemorySpace::name())).c_str());
-            return;
-    default:
-      Kokkos::abort("bug: unknown Cuda memory type");
+    case cudaMemoryTypeHost:
+      if (!Kokkos::SpaceAccessibility<RequestedMemorySpace,
+                                      Kokkos::CudaHostPinnedSpace>::assignable)
+        Kokkos::abort(
+            ("Detected CudaHostPinnedSpace but requested incompatible " +
+             std::string(RequestedMemorySpace::name()))
+                .c_str());
+      return;
+    case cudaMemoryTypeDevice:
+      if (!Kokkos::SpaceAccessibility<RequestedMemorySpace,
+                                      Kokkos::CudaSpace>::assignable)
+        Kokkos::abort(("Detected CudaSpace but requested incompatible " +
+                       std::string(RequestedMemorySpace::name()))
+                          .c_str());
+      return;
+    case cudaMemoryTypeManaged:
+      if (!Kokkos::SpaceAccessibility<RequestedMemorySpace,
+                                      Kokkos::CudaUVMSpace>::assignable)
+        Kokkos::abort(("Detected CudaUVMSpace but requested incompatible " +
+                       std::string(RequestedMemorySpace::name()))
+                          .c_str());
+      return;
+    case cudaMemoryTypeUnregistered:
+      if (!Kokkos::SpaceAccessibility<RequestedMemorySpace,
+                                      Kokkos::HostSpace>::assignable)
+        Kokkos::abort(("Detected HostSpace but requested incompatible " +
+                       std::string(RequestedMemorySpace::name()))
+                          .c_str());
+      return;
+    default: Kokkos::abort("bug: unknown Cuda memory type");
   }
 }
 
-}
-
+}  // namespace
 
 template <>
 void Kokkos::Impl::runtime_check_memory_space<Kokkos::CudaHostPinnedSpace>(
--- a/core/src/Cuda/Kokkos_CudaSpace.cpp
+++ b/core/src/Cuda/Kokkos_CudaSpace.cpp
@@ -452,37 +452,40 @@ void cuda_prefetch_pointer(cudaStream_t stream, const void *ptr, size_t bytes,
 
 namespace {
 
+template <typename DetectedMemorySpace, typename RequestedMemorySpace>
+void abort_incompatible_memory_spaces() {
+  Kokkos::abort(("Detected " + std::string(DetectedMemorySpace::name()) +
+                 " but " + "requested incompatible " +
+                 std::string(RequestedMemorySpace::name()))
+                    .c_str());
+}
+
 template <typename RequestedMemorySpace>
 void check_memory_space(cudaMemoryType deduced_memory_type) {
   switch (deduced_memory_type) {
     case cudaMemoryTypeHost:
       if (!Kokkos::SpaceAccessibility<RequestedMemorySpace,
                                       Kokkos::CudaHostPinnedSpace>::assignable)
-        Kokkos::abort(
-            ("Detected CudaHostPinnedSpace but requested incompatible " +
-             std::string(RequestedMemorySpace::name()))
-                .c_str());
+        abort_incompatible_memory_spaces<Kokkos::CudaHostPinnedSpace,
+                                         RequestedMemorySpace>();
       return;
     case cudaMemoryTypeDevice:
       if (!Kokkos::SpaceAccessibility<RequestedMemorySpace,
                                       Kokkos::CudaSpace>::assignable)
-        Kokkos::abort(("Detected CudaSpace but requested incompatible " +
-                       std::string(RequestedMemorySpace::name()))
-                          .c_str());
+        abort_incompatible_memory_spaces<Kokkos::CudaSpace,
+                                         RequestedMemorySpace>();
       return;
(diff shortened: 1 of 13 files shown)


core/unit_test/TestViewMapping_a.hpp
```diff
@@ -647,7 +647,7 @@ void test_view_mapping() {
 
   //----------------------------------------
 
-  {
+  if (std::is_same_v<typename Space::memory_space, Kokkos::HostSpace>) {
```
masterleinad, line 650: We might want to check for accessibility instead.

core/src/View/Kokkos_ViewCtor.hpp
```diff
@@ -151,12 +151,13 @@ struct ViewCtorProp<void, T *> {
...
-  static constexpr bool has_label           = false;
-  static constexpr bool allow_padding       = false;
-  static constexpr bool initialize          = true;
+  static constexpr bool has_memory_space     = false;
+  static constexpr bool has_execution_space  = false;
+  static constexpr bool has_pointer          = true;
+  static constexpr bool has_label            = false;
+  static constexpr bool allow_padding        = false;
+  static constexpr bool initialize           = true;
+  static constexpr bool sequential_host_init = false;
```
masterleinad, line 160: We should fix this either way.

core/unit_test/TestViewEmptyRuntimeUnmanaged.hpp
```diff
@@ -25,12 +25,6 @@ void test_empty_view_runtime_unmanaged() {
   T d{};
   auto* p = reinterpret_cast<T*>(0xABADBABE);
 
-  (void)Kokkos::View<T>(p);
-  (void)Kokkos::View<T>(&d);
-  (void)Kokkos::View<T>(nullptr);
-  (void)Kokkos::View<T>(NULL);  // NOLINT(modernize-use-nullptr)
-  (void)Kokkos::View<T>(0);     // NOLINT(modernize-use-nullptr)
-
```
masterleinad, line 33: These `Views` aren't in fact empty.

core/unit_test/TestViewMapping_a.hpp
```diff
@@ -667,45 +667,45 @@ void test_view_mapping() {
     ASSERT_EQ(vr1.data(), &data[0]);
     ASSERT_EQ(cr1.data(), &data[0]);
 
-    ASSERT_TRUE((std::is_same_v<typename T::data_type, int*>));
-    ASSERT_TRUE((std::is_same_v<typename T::const_data_type, const int*>));
-    ASSERT_TRUE((std::is_same_v<typename T::non_const_data_type, int*>));
+    static_assert(std::is_same_v<typename T::data_type, int*>);
```
masterleinad, line 670: Some what of a drive-by change but we are now guarding the test more restrictively at runtime the checks here need to be compile-time to still work despite the guard.

masterleinad:
Just triggering GitLab CI.

core/src/View/Kokkos_ViewCtor.hpp
masterleinad: https://github.com/kokkos/kokkos/pull/7605.

core/src/SYCL/Kokkos_SYCL_Space.cpp
```diff
@@ -62,14 +63,80 @@ std::string_view get_memory_space_name(sycl::usm::alloc allocation_kind) {
     case sycl::usm::alloc::host: return Kokkos::SYCLHostUSMSpace::name();
     case sycl::usm::alloc::device: return Kokkos::SYCLDeviceUSMSpace::name();
     case sycl::usm::alloc::shared: return Kokkos::SYCLSharedUSMSpace::name();
+    case sycl::usm::alloc::unknown: return "sycl::usm::alloc::unknown";
```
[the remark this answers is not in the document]
masterleinad: Could also remove this.

core/src/Cuda/Kokkos_CudaSpace.cpp
```diff
@@ -465,6 +465,79 @@ void cuda_prefetch_pointer(const Cuda &space, const void *ptr, size_t bytes,
...
+                          .c_str());
+      return;
+    case cudaMemoryTypeManaged:
+      if (!Kokkos::SpaceAccessibility<RequestedMemorySpace,
+                                      Kokkos::CudaUVMSpace>::assignable)
+        Kokkos::abort(("Detected CudaUVMSpace but requested incompatible " +
+                       std::string(RequestedMemorySpace::name()))
+                          .c_str());
+      return;
+    case cudaMemoryTypeUnregistered:
```
JBludau, line 496: So a pointer allocated using `malloc` on a machine with CUDA 12 and a kernel that supports HMM would hit the `Managed` path?

core/src/HIP/Kokkos_HIP_Space.hpp
```diff
@@ -375,7 +375,28 @@ struct MemorySpaceAccess<HIPManagedSpace, HIPHostPinnedSpace> {
   enum : bool { deepcopy = true };
 };
 
+template <typename MemorySpace>
+void runtime_check_memory_space(const void* ptr, const MemorySpace&);
```
JBludau, line 379: Since we check for assignability, should it be represented in the name?

core/src/Cuda/Kokkos_CudaSpace.cpp
masterleinad: I would expect the runtime to say `Unregistered` but I haven't checked.

JBludau: If I understand https://developer.nvidia.com/blog/simplifying-gpu-application-development-with-heterogeneous-memory-management/#system_allocated_memory_and_cuda_allocators correctly, basically all memory allocated with traditional stuff like `mmap` or `malloc` should be accessible on the GPU and CPU if HMM is available and the driver supports it (independently of the GPU) ... I guess we would need to test if it results in `Unregistered` or `Managed`.

masterleinad: See https://github.com/kokkos/kokkos/issues/7597#issuecomment-2541733930 for Grace Hopper. In short, the runtime detection only cares about the allocation mechanism. `malloc` is mapped to `Unregistered`.

masterleinad: Also note that we don't change any of the `SpaceAccessibility<...>::assignable` values for MI300A or GH200.

JBludau: Thanks for the tests, @masterleinad. I guess my point is: should we treat something that the runtime reports `Unregistered` as `Managed`, if it is malloced on coda 12.2 with HMM? because it can migrate similar to UVM.

masterleinad: In my opinion, all considerations should go into `SpaceAccessibility`/`MemorySpaceAccess`. We can certainly say that `HostSpace` is assignable to `CudaUVMSpace` but I wouldn't mess with the detection here.

JBludau: yeah, I am all for deciding in the `SpaceAccessibility`. I just think we should make a plan how to deal with the somewhat new memory behavior in general. That was the intention of https://github.com/kokkos/kokkos/issues/7607

core/src/HIP/Kokkos_HIP_Space.hpp
masterleinad: Sure, see 0a911e4e49313f1005145c7682292fb4d6b7cad9.

cedricchevalier19 (commented):
This PR makes things a little bit stricter than what the system can do but it should help #7452 .

We have to clarify `accessibility` properties: is it memory that can be accessed from the device or memory that is valid from any device-library functions like `*memcpy/memset/...`

core/unit_test/TestViewEmptyRuntimeUnmanaged.hpp
dalg24: Agreed

masterleinad:
We need to decide first on SpaceAccessibility for HMM like in https://github.com/kokkos/kokkos/issues/7607 and #7567 for this not to break intended use cases.

cedricchevalier19:
Using `accessible` will make some codes valid with GH enabled but invalid without enabling it.

masterleinad:
> We have to clarify `accessibility` properties: is it memory that can be accessed from the device or memory that is valid from any device-library functions like `*memcpy/memset/...`

I would argue, we should check`assignable` but yes we have to clarify what that means for all combinations.

masterleinad:
We decided that we want for now to allow people to to wrap raw pointers on APUs that come from any allocation and that we would only check the allocation type for non-APU setups.

janciesko:
Retest this please

janciesko (approved):

core/unit_test/TestViewMemorySpaceViolation.cpp
```diff
@@ -0,0 +1,137 @@
...
+       defined(KOKKOS_ENABLE_SYCL)) &&                              \
+      defined(KOKKOS_ENABLE_DEBUG))
+  GTEST_SKIP() << "memory space violation only detected for Cuda, HIP, or SYCL "
+                  "with Kokkos_ENABLE_DEBUG";
+#else
+
+#if defined(KOKKOS_ENABLE_IMPL_CUDA_UNIFIED_MEMORY) || \
+    defined(KOKKOS_IMPL_HIP_UNIFIED_MEMORY)
+  GTEST_SKIP()
+      << "We don't want to check pointer memory spaces for APU-like setups";
```
cedricchevalier19: At least for Nvidia, it might be of interest to test whether a "Cuda pointer" is not passed to construct a new host space as `CudaMalloc*` pointers are not readable from the CPU (on GH200).

janciesko: Yes, we should be able to cover this, probably by just removing [this](
https://github.com/kokkos/kokkos/pull/7480/changes/BASE..5f00b58ad84cede2702f1862a85def56306b3831#diff-6e45f19e60b403ba42a8031814fe8b23473d8dad69937b6dfb17c71d1412c575R423-R424)

janciesko: @masterleinad thoughts?

janciesko: Thanks Daniel, I think this looks good. @cedricchevalier19 please see if this addresses your concern.

core/src/Cuda/Kokkos_CudaSpace.cpp
```diff
@@ -449,6 +449,86 @@ void cuda_prefetch_pointer(cudaStream_t stream, const void *ptr, size_t bytes,
...
+                 std::string(RequestedMemorySpace::name()))
+                    .c_str());
+}
+
+template <typename RequestedMemorySpace>
+void check_memory_space(cudaMemoryType deduced_memory_type) {
+  switch (deduced_memory_type) {
+    case cudaMemoryTypeHost:
+      if (!Kokkos::SpaceAccessibility<RequestedMemorySpace,
+                                      Kokkos::CudaHostPinnedSpace>::assignable)
```
ldh4, line 468: These `if` blocks could all be `if constexpr`.

core/src/HIP/Kokkos_HIP_Space.cpp
```diff
@@ -306,3 +306,154 @@ void HIPManagedSpace::impl_deallocate(
...
+        abort_incompatible_memory_spaces<Kokkos::HIPSpace,
+                                         RequestedMemorySpace>();
+      return;
+    case hipMemoryTypeManaged:
+      if (!Kokkos::SpaceAccessibility<RequestedMemorySpace,
+                                      Kokkos::HIPManagedSpace>::assignable)
+        abort_incompatible_memory_spaces<Kokkos::HIPManagedSpace,
+                                         RequestedMemorySpace>();
+      return;
+#if HIP_VERSION_MAJOR >= 6
```
ldh4: The minimum requirement is now rocm 6.2. Maybe we no longer need these various version checks in this file?

ldh4 (approved):

core/src/Cuda/Kokkos_CudaSpace.cpp
crtrott: OK coming back to this discussion: I think this path should likely not check. There is a difference between what we say you are allowed to do with strongly typed things, and what inter-op via ptrs looks like. I.e. no: we shouldn't allow assigning HostSpace to CudaManagedSpace in my mind, but if some user wants to stuff a malloced ptr into a SharedSpace View when HMM is enabled that's on them.

crtrott (changes requested):
Don't agree with the unregistered path checks.

PR #7484, opened 2024-10-24
Revert 6683:  simd: add compound assignments and reductions

masterleinad:
Testing with AVX512 is currebtly broken and it seems to make sense to revert, add consistent AVX512 testing (see #7477), and fix then again.


ldh4 (approved):
I agree with reverting this, especially since this hinders the release. Clearly, there are errors that were not caught in the CI.

nmm0 (approved):
ok, if @crtrott agrees

crtrott (approved):

PR #7486, opened 2024-10-25
simd: add compound assignments and update reductions

ldh4:
This is a clone of a reverted PR: #6683.

Descriptions from the original PR:

This PR adds following functions described in P1928

- Compound assignments
    - `simd<T, Abi>& operator/=(simd<T, Abi>& lhs, U&& rhs)` (enabled)
    - `simd<T, Abi>& operator>>=(simd<T, Abi>& lhs, U&& rhs)` (added)
    - `simd<T, Abi>& operator<<=(simd<T, Abi>& lhs, U&& rhs)` (added)

- Constructor that accepts a pointer to the src memory
    - `constexpr explicit simd(value_type const* ptr, FlagType flag)`

- Fallback simd reducers
    - `T reduce(const simd<T, Abi>& x, const typename simd<T, Abi>::mask_type& mask, T identity_element, BinaryOperation binary_op)`
    - `T reduce(const simd<T, Abi>& x, const typename simd<T, Abi>::mask_type& mask, std::plus<> binary_op = {}) noexcept`
    - `T reduce(const simd<T, Abi>& x, const typename simd<T, Abi>::mask_type& mask, std::bit_and<> binary_op) noexcept`
    - `T reduce(const simd<T, Abi>& x, const typename simd<T, Abi>::mask_type& mask, std::bit_or<> binary_op) noexcept`
    - `T reduce(const simd<T, Abi>& x, const typename simd<T, Abi>::mask_type& mask, std::bit_xor<> binary_op) noexcept`
    - `T reduce_min(const simd<T, Abi>& x) noexcept`
    - `T reduce_min(const simd<T, Abi>& x, const typename simd<T, Abi>::mask_type& mask) noexcept`
    - `T reduce_max(const simd<T, Abi>& x) noexcept`
    - `T reduce_max(const simd<T, Abi>& x, const typename simd<T, Abi>::mask_type& mask) noexcept`

- `Max`, `Min`, `Sum` reducers for AVX512
    - `T reduce_min(const simd<T, Abi>& x)`
    - `T reduce_max(const simd<T, Abi>& x)`
    - `T reduce(const_where_expression<simd_mask<T, Abi>, simd<T, Abi>> const& x, T, std::plus<>)`

- Deprecated `hmin` and `hmax` in favor of `reduce_min` and `reduce_max`

### DIFF
--- a/simd/src/Kokkos_SIMD_Common.hpp
+++ b/simd/src/Kokkos_SIMD_Common.hpp
@@ -25,6 +25,10 @@ namespace Kokkos {
 
 namespace Experimental {
 
+namespace simd_abi {
+class scalar;
+}
+
 template <class T, class Abi>
 class basic_simd;
 
@@ -135,7 +139,7 @@ template <class T>
 
 template <class T, class U, class Abi,
           std::enable_if_t<std::is_arithmetic_v<U>, bool> = false>
-[[nodiscard]] KOKKOS_FORCEINLINE_FUNCTION auto operator+(
+[[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION auto operator+(
     Experimental::basic_simd<T, Abi> const& lhs, U rhs) {
   using result_member = decltype(lhs[0] + rhs);
   return Experimental::basic_simd<result_member, Abi>(lhs) +
@@ -144,15 +148,17 @@ template <class T, class U, class Abi,
 
 template <class T, class U, class Abi,
           std::enable_if_t<std::is_arithmetic_v<U>, bool> = false>
-[[nodiscard]] KOKKOS_FORCEINLINE_FUNCTION auto operator+(
+[[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION auto operator+(
     U lhs, Experimental::basic_simd<T, Abi> const& rhs) {
   using result_member = decltype(lhs + rhs[0]);
   return Experimental::basic_simd<result_member, Abi>(lhs) +
          Experimental::basic_simd<result_member, Abi>(rhs);
 }
 
-template <class T, class U, class Abi>
-KOKKOS_FORCEINLINE_FUNCTION basic_simd<T, Abi>& operator+=(
+template <
+    class T, class U, class Abi,
+    std::enable_if_t<!std::is_same_v<Abi, simd_abi::scalar>, bool> = false>
+KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION basic_simd<T, Abi>& operator+=(
     basic_simd<T, Abi>& lhs, U&& rhs) {
   lhs = lhs + std::forward<U>(rhs);
   return lhs;
@@ -167,7 +173,7 @@ KOKKOS_FORCEINLINE_FUNCTION where_expression<M, T>& operator+=(
 
 template <class T, class U, class Abi,
           std::enable_if_t<std::is_arithmetic_v<U>, bool> = false>
-[[nodiscard]] KOKKOS_FORCEINLINE_FUNCTION auto operator-(
+[[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION auto operator-(
     Experimental::basic_simd<T, Abi> const& lhs, U rhs) {
   using result_member = decltype(lhs[0] - rhs);
   return Experimental::basic_simd<result_member, Abi>(lhs) -
@@ -176,22 +182,24 @@ template <class T, class U, class Abi,
 
 template <class T, class U, class Abi,
           std::enable_if_t<std::is_arithmetic_v<U>, bool> = false>
-[[nodiscard]] KOKKOS_FORCEINLINE_FUNCTION auto operator-(
+[[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION auto operator-(
     U lhs, Experimental::basic_simd<T, Abi> const& rhs) {
   using result_member = decltype(lhs - rhs[0]);
   return Experimental::basic_simd<result_member, Abi>(lhs) -
          Experimental::basic_simd<result_member, Abi>(rhs);
 }
 
-template <class T, class U, class Abi>
-KOKKOS_FORCEINLINE_FUNCTION basic_simd<T, Abi>& operator-=(
+template <
+    class T, class U, class Abi,
+    std::enable_if_t<!std::is_same_v<Abi, simd_abi::scalar>, bool> = false>
+KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION basic_simd<T, Abi>& operator-=(
     basic_simd<T, Abi>& lhs, U&& rhs) {
   lhs = lhs - std::forward<U>(rhs);
   return lhs;
 }
 
 template <class M, class T, class U>
-KOKKOS_FORCEINLINE_FUNCTION where_expression<M, T>& operator-=(
+KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION where_expression<M, T>& operator-=(
     where_expression<M, T>& lhs, U&& rhs) {
   lhs = lhs.value() - std::forward<U>(rhs);
   return lhs;
@@ -199,7 +207,7 @@ KOKKOS_FORCEINLINE_FUNCTION where_expression<M, T>& operator-=(
 
 template <class T, class U, class Abi,
           std::enable_if_t<std::is_arithmetic_v<U>, bool> = false>
-[[nodiscard]] KOKKOS_FORCEINLINE_FUNCTION auto operator*(
+[[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION auto operator*(
     Experimental::basic_simd<T, Abi> const& lhs, U rhs) {
   using result_member = decltype(lhs[0] * rhs);
   return Experimental::basic_simd<result_member, Abi>(lhs) *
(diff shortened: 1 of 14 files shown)


ldh4:
Few simd unit tests in this PR are currently producing different results depending on optimization level.
It seems that turning on -O3 is making a few tests to have undefined behaviors somewhere.

WIP

masterleinad:
Can you cherry-pick https://github.com/kokkos/kokkos/pull/7477/commits/7445d06bc7a7239f66ca59dc135d989287150d40 here so that we have reliable AVX512 testing for this pull request?

ldh4:
> Can you cherry-pick [7445d06](https://github.com/kokkos/kokkos/commit/7445d06bc7a7239f66ca59dc135d989287150d40) here so that we have reliable AVX512 testing for this pull request?

Cherry-picked ([0a2f72e](https://github.com/kokkos/kokkos/pull/7486/commits/0a2f72e297c744965099f53e93b4c66033815ba9))

There seems to be some issues with load instructions in simd unit tests when specifically built by gcc with `-O2` or above and `-g`. Values don't seem to be properly loaded when simd instrinsics are used before accessing individual values in each simd vector. This doesn't seem to be a problem for clang, however.

Placed in some ugly workarounds. Unit tests for `where_expression` could be not mitigated this way. Thus, disabled the test for non-debug gcc builds.

dalg24:
What is the status of this?

ldh4:
> What is the status of this?

@dalg24 This is ready for reviews again. The issue that I described above still exists but is handled for simd unit tests by placing memory fences. This issue is not introduced by changes in this PR. So, it will have to be examined further separately.

masterleinad:
@ldh4 Can you remove and set the special approval tag and then retrigger the SNL CI?

ldh4:
Retest this please.

masterleinad:
@ldh4 I meant retriggering the (failing) GitHub CI, not the jenkins CI.

simd/src/Kokkos_SIMD_NEON.hpp
```diff
@@ -2687,6 +2722,196 @@ class where_expression<simd_mask<std::uint64_t, simd_abi::neon_fixed_size<2>>,
   }
 };
 
+// [[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION std::int32_t
```
janciesko, line 2725: What's the reason for the commented blocks here?

simd/src/Kokkos_SIMD_Common.hpp
```diff
@@ -134,7 +138,7 @@ template <class T>
 
 template <class T, class U, class Abi,
           std::enable_if_t<std::is_arithmetic_v<U>, bool> = false>
-[[nodiscard]] KOKKOS_FORCEINLINE_FUNCTION auto operator+(
+[[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION auto operator+(
```
masterleinad, line 141: Why are this and the one below now a host-only function? I guess we have a different overload for device-compatible types?

simd/src/Kokkos_SIMD_Common_Math.hpp
```diff
@@ -32,8 +36,9 @@ class simd_mask;
 template <class M, class T>
 class const_where_expression;
 
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
 template <typename T, typename Abi>
-[[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION T
+[[nodiscard]] KOKKOS_DEPRECATED KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION T
```
masterleinad, line 41: Can you use `KOKKOS_DEPRECATED_WITH_COMMENT` to provide some guidance on the replacement?

simd/src/Kokkos_SIMD_Scalar.hpp
```diff
@@ -294,6 +347,58 @@ KOKKOS_FORCEINLINE_FUNCTION simd<T, simd_abi::scalar> condition(
...
+}
+
+template <class T, class BinaryOperation>
+[[nodiscard]] KOKKOS_FORCEINLINE_FUNCTION T
+reduce(Experimental::simd<T, Experimental::simd_abi::scalar> const& x,
+       BinaryOperation binary_op) noexcept {
+  return reduce(
+      x, Experimental::simd<T, Experimental::simd_abi::scalar>::mask_type(true),
+      T(0), binary_op);
+}
```
masterleinad, line 366: Is this overload necessary? It seems dangerous to implicitly assume `0` to be the neutral element.

simd/unit_tests/include/SIMDTesting_Utilities.hpp
```diff
@@ -191,4 +191,27 @@ constexpr bool is_type_v = false;
...
+// Zero initializing each simd lane to avoid gcc's uninitialized variable
+// warnings on default constructed simd variables.
+// Using a for-loop to initialize:
+//
+// SIMDType a;
+// for(int i=0; i<SIMDType::size(); ++i)
+//   a[i] = 0;
+// return a;
+//
+// does not seem to suppress gcc's warnings.
```
masterleinad, line 203: I guess that's also an artifact of the reinterpret_cast issue?

simd/unit_tests/include/TestSIMD_MathOps.hpp
```diff
@@ -29,21 +29,34 @@ void host_check_math_op_one_loader(BinaryOp binary_op, std::size_t n,
...
-    simd_type expected_result;
+    simd_type expected_result(zero_init<simd_type>());
+
+    // gcc build with cxxflag of -g and -O2 or above doesn't seem to properly
+    // load values into simd vectors until simd values are directly accessed.
+    // Placing a memory fence to ensure that simd values are fully loaded
+    // before executing simd instructions.
+#if defined(KOKKOS_COMPILER_GNU) && defined(NDEBUG)
+    __sync_synchronize();
+#endif
```
masterleinad, line 50: That issue should go away if we don't try to access elements via `reinterpret_cast`, right?

simd/src/Kokkos_SIMD_Scalar.hpp
ldh4: The overload is necessary as this is part of the reduction interface in std::simd.
As for the identity element, it does look like `0` would not be a correct value to put in according to std::simd:
```
The default argument for identity_element is equal to
  • T() if BinaryOperation is plus<>,
  • T(1) if BinaryOperation is multiplies<>,
  • T(~T()) if BinaryOperation is bit_and<>,
  • T() if BinaryOperation is bit_or<>, or
  • T() if BinaryOperation is bit_xor<>.
```
I'll get it fixed.

simd/unit_tests/include/SIMDTesting_Utilities.hpp
ldh4: Yes, these zero inits may no longer be needed depending on the outcome of how the reinterpret_cast issue is be handled.

simd/unit_tests/include/TestSIMD_MathOps.hpp
ldh4: Yes, these memory fences will be deleted once the reinterpret cast issue is fixed.

simd/src/Kokkos_SIMD_Scalar.hpp
ldh4: Fixed along with the latest rebase.

simd/unit_tests/include/SIMDTesting_Utilities.hpp
ldh4: These are removed now.

simd/unit_tests/include/TestSIMD_MathOps.hpp
ldh4: Memory fences are removed now.

simd/src/Kokkos_SIMD_Scalar.hpp
```diff
@@ -412,6 +523,17 @@ template <class T>
   return a == basic_simd_mask<T, Kokkos::Experimental::simd_abi::scalar>(false);
 }
 
+#ifdef KOKKOS_ENABLE_DEPRECATED_CODE_4
+template <class T>
+[[nodiscard]] KOKKOS_DEPRECATED KOKKOS_FORCEINLINE_FUNCTION T
```
masterleinad, line 528: Would you mind replacing `KOKKOS_DEPRECATED` with `KOKKOS_DEPRECATED_WITH_COMMENT` like you do in some overloads?

masterleinad (commented):
Apart from the consistency in `KOKKOS_DEPRECATED[_WITH_COIMMENT])`, this looks good to me.

simd/src/Kokkos_SIMD_Scalar.hpp
ldh4: Added deprecation messages for `hmin` and `hmax`.

masterleinad (approved):

simd/src/Kokkos_SIMD_Scalar.hpp
```diff
@@ -291,6 +345,63 @@ KOKKOS_FORCEINLINE_FUNCTION basic_simd<T, simd_abi::scalar> condition(
...
+}
+
+template <class T>
+[[nodiscard]] KOKKOS_FORCEINLINE_FUNCTION T
+reduce_min(Experimental::basic_simd<T, Experimental::simd_abi::scalar> const&
+               x) noexcept {
+  return reduce_min(
+      x, Experimental::basic_simd<T, Experimental::simd_abi::scalar>::mask_type(
+             true));
+}
```
crtrott, line 385: Why not simply return x[0]? You call the above function with true, which return x[0]

simd/src/Kokkos_SIMD_Scalar.hpp
```diff
@@ -291,6 +345,63 @@ KOKKOS_FORCEINLINE_FUNCTION basic_simd<T, simd_abi::scalar> condition(
...
+  return x[0];
+}
+
+template <class T>
+[[nodiscard]] KOKKOS_FORCEINLINE_FUNCTION T
+reduce_min(Experimental::basic_simd<T, Experimental::simd_abi::scalar> const&
+               x) noexcept {
+  return reduce_min(
+      x, Experimental::basic_simd<T, Experimental::simd_abi::scalar>::mask_type(
+             true));
```
crtrott, line 384:
```suggestion
  return x[0];
```

simd/src/Kokkos_SIMD_Scalar.hpp
```diff
@@ -291,6 +345,63 @@ KOKKOS_FORCEINLINE_FUNCTION basic_simd<T, simd_abi::scalar> condition(
...
+  return x[0];
+}
+
+template <class T>
+[[nodiscard]] KOKKOS_FORCEINLINE_FUNCTION T
+reduce_max(Experimental::basic_simd<T, Experimental::simd_abi::scalar> const&
+               x) noexcept {
+  return reduce_max(
+      x, Experimental::basic_simd<T, Experimental::simd_abi::scalar>::mask_type(
+             true));
```
crtrott, line 402:
```suggestion
  return x[0];
```

simd/src/Kokkos_SIMD_Common_Math.hpp
```diff
@@ -57,17 +64,47 @@ hmax(const_where_expression<basic_simd_mask<T, Abi>, basic_simd<T, Abi>> const&
...
+    class T, class Abi, class BinaryOperation = std::plus<>,
+    std::enable_if_t<!std::is_same_v<Abi, simd_abi::scalar>, bool> = false>
+[[nodiscard]] KOKKOS_IMPL_HOST_FORCEINLINE_FUNCTION T
+reduce(basic_simd<T, Abi> const& v,
+       typename basic_simd<T, Abi>::mask_type const& m, T identity,
+       BinaryOperation op = {}) {
+  if (none_of(m)) {
+    return identity;
+  }
+  auto result = v[0];
```
crtrott, line 105: what if mask[0] is false?

simd/src/Kokkos_SIMD_Common_Math.hpp
```diff
@@ -57,17 +64,47 @@ hmax(const_where_expression<basic_simd_mask<T, Abi>, basic_simd<T, Abi>> const&
...
+  for (std::size_t i = 1; i < v.size(); ++i) {
+    if (m[i]) result = op(result, v[i]);
   }
```
crtrott, line 108:
```suggestion
  auto result = identity;
  for (std::size_t i = 0; i < v.size(); ++i) {
    if (m[i]) result = op(result, v[i]);
  }
```

crtrott (changes requested):
I think I found one bug and two places which could be simplified. You may wanna add a test for the masked reduction where the first element of the mask is false.

simd/src/Kokkos_SIMD_Common_Math.hpp
ldh4: Good catch. Fixed to use the identity value instead.

simd/src/Kokkos_SIMD_Common_Math.hpp
ldh4: Using `identity` wasn't entirely correct as `identity` could be a user provided arbitrary value. Modified to use a true identity element to fix this. See [d3211e1](https://github.com/kokkos/kokkos/pull/7486/commits/d3211e143d048ebd9e4e1fbc0171b112794bdb5b).

simd/src/Kokkos_SIMD_Scalar.hpp
ldh4: Applied. [d3211e1](https://github.com/kokkos/kokkos/pull/7486/commits/d3211e143d048ebd9e4e1fbc0171b112794bdb5b)

simd/src/Kokkos_SIMD_Scalar.hpp
ldh4: Yes, switched to simply return `x[0]`.

simd/src/Kokkos_SIMD_Scalar.hpp
ldh4: Applied. [d3211e1](https://github.com/kokkos/kokkos/pull/7486/commits/d3211e143d048ebd9e4e1fbc0171b112794bdb5b)

ldh4 (commented):
> I think I found one bug and two places which could be simplified. You may wanna add a test for the masked reduction where the first element of the mask is false.

Modified the existing reduction test with more combination of mask values, including mask[0] being false.

crtrott (approved):

dalg24:
The SYCL failure on ORNL Jenkins CI seems unrelated

PR #7490, opened 2024-10-28
ViewTraits Interface Cleanup Part 3: Deprecate HostMirrorSpace in favor of host_mirror_device

janciesko:
[WIP]
Changes:
- `HostMirrorSpace` -> deprecated in favor of `host_mirror_device`
- `HostMirror` -> deprecated in favor of `host_mirror_type`(member of View)
- Added `host_mirror_device`
- `HostMirror` requires a Device

TODO:
[] Properly deprecate previous members and maintain behavior
[] Make work with C++17 (no use concepts)
[] Put back manual creation of devices in TypeDef* tests


algorithms/src/Kokkos_Random.hpp
```diff
@@ -915,13 +915,14 @@ class Random_XorShift64_Pool {
     state_ = state_data_type("Kokkos::Random_XorShift64::state", num_states_,
                              padding_);
 
-    typename state_data_type::HostMirror h_state =
+    typename state_data_type::host_mirror_type h_state =
```
romintomasetti, line 918: I would do a separate PR for just this change (*i.e.* search replace of `HostMirror` with `host_mirror_type`.

romintomasetti (commented):
Please separate concerns to ease the review.

algorithms/src/Kokkos_Random.hpp
janciesko: Will do, thanks for looking into this.

janciesko: I think these are the minimal changes to cover in this PR. Note: replacing `HostMirror` with `host_mirror_type` requires to define that type which depends then on the other changes. Please see if this makes sense to you.

containers/src/Kokkos_UnorderedMap.hpp
```diff
@@ -256,8 +256,8 @@ class UnorderedMap {
 
   using insert_result = UnorderedMapInsertResult;
 
-  using HostMirror =
-      UnorderedMap<Key, Value, host_mirror_space, Hasher, EqualTo>;
+  using host_mirror_type =
+      UnorderedMap<Key, Value, host_mirror_device, Hasher, EqualTo>;
```
crtrott, line 260: We may need to hide this change (i.e. changing `host_mirror_space` to `host_mirror_device`) behind a deprecation guard. We are changing types here, which means that in principle it's a breaking change.

containers/unit_tests/TestDynRankViewTypedefs.cpp
```diff
@@ -180,93 +181,74 @@ constexpr bool test_view_typedefs_impl() {
...
-  // HostMirrorSpace is a mess so: if the default exec is a host exec, that is it
-  using host_mirror_space = std::conditional_t<is_host_exec, Kokkos::DefaultExecutionSpace,
-  // otherwise if unified memory is not on its HostSpace
-                               std::conditional_t<!has_unified_mem_space, Kokkos::HostSpace,
-  // otherwise its the following Device type
-                               Kokkos::Device<Kokkos::DefaultHostExecutionSpace, typename Kokkos::DefaultExecutionSpace::memory_space>>>;
-  static_assert(test_view_typedefs<layout_type, space, memory_traits, host_mirror_space, int, int&>(
+  using space         = Kokkos::DefaultExecutionSpace;
+  using device        = Kokkos::Device<space, typename space::memory_space>;
+  using host_mirror_device = Kokkos::Impl::HostMirror<device>::Device;
```
crtrott, line 199: The test previously very explicitly did not use the internal pieces of Kokkos to figure out what the expected host type is. This was done in order to make sure our internals do what we expect. Right now a bug in Kokkos::Impl::HostMirror would make the test pass, since both the View and the test use the same implementation. So say `HostMirror::Device` as `Device<Cuda, CudaSpace>` would make the test still pass even though the implementation is faulty.

containers/unit_tests/TestDynRankViewTypedefs.cpp
```diff
@@ -180,93 +181,74 @@ constexpr bool test_view_typedefs_impl() {
...
-  using host_mirror_space = std::conditional_t<is_host_exec, Kokkos::HostSpace,
-  // otherwise if unified memory is not on its also HostSpace!
-                               std::conditional_t<!has_unified_mem_space, Kokkos::HostSpace,
-  // otherwise its the following memory space ...
-                               Kokkos::DefaultExecutionSpace::memory_space>>;
-  static_assert(test_view_typedefs<layout_type, space, memory_traits, host_mirror_space, int, int&>(
-                     ViewParams<int, Kokkos::DefaultExecutionSpace>{}));
+  using space         = Kokkos::DefaultExecutionSpace;
+  using device        = Kokkos::Device<space, typename space::memory_space>;
+  using host_mirror_device = Kokkos::Impl::HostMirror<device>::Device;
```
crtrott, line 210: same as above

core/src/Kokkos_Concepts.hpp
```diff
@@ -286,9 +286,10 @@ struct is_space {
 
  public:
   static constexpr bool value = is_exe::value || is_mem::value || is_dev::value;
-
+  static constexpr bool is_exec_space() { return is_exe::value; }
+  static constexpr bool is_mem_space() { return is_mem::value; }
+  static constexpr bool is_device() { return is_dev::value; }
```
crtrott, line 291: Not sure how I feel about this given that this is supposed to be a minimal type trait, on the other hand it does contain to public typedefs already. Other folks thoughts?

crtrott: Also assume we keep this: are these tested?

core/src/Kokkos_Crs.hpp
```diff
@@ -73,10 +73,10 @@ class Crs {
   using size_type       = SizeType;
 
   using staticcrsgraph_type = Crs<DataType, Arg1Type, Arg2Type, SizeType>;
-  using HostMirror =
-      Crs<DataType, array_layout, typename traits::host_mirror_space, SizeType>;
-  using row_map_type = View<size_type*, array_layout, device_type>;
-  using entries_type = View<DataType*, array_layout, device_type>;
+  using host_mirror_type    = Crs<DataType, array_layout,
+                               typename traits::host_mirror_device, SizeType>;
```
crtrott, line 77: Same comment as for containers: may need deprecation guard.

core/src/Kokkos_HostSpace.hpp
```diff
@@ -138,39 +138,47 @@ namespace Kokkos {
 
 namespace Impl {
 
-static_assert(Kokkos::Impl::MemorySpaceAccess<Kokkos::HostSpace,
-                                              Kokkos::HostSpace>::assignable);
```
crtrott, line 142: Why delete this?

core/src/Kokkos_HostSpace.hpp
```diff
@@ -138,39 +138,47 @@ namespace Kokkos {
...
-
-template <typename S>
+template <class S>
 struct HostMirror {
  private:
+  using is_space = typename Kokkos::is_space<S>;
+  static_assert(is_space());
```
crtrott, line 145:
```suggestion
  static_assert(is_space::value);
```
I don't really like the reliance on implicit conversion to bool here.

core/src/Kokkos_HostSpace.hpp
```diff
@@ -138,39 +138,47 @@ namespace Kokkos {
...
                                         typename S::memory_space>::accessible
   };
 
  public:
-  using Space = std::conditional_t<
-      keep_exe && keep_mem, S,
+  // Construct a device mirror type
+  using Device = std::conditional_t<
+      keep_exe && keep_mem,
+      Kokkos::Device<typename S::execution_space, typename S::memory_space>,
```
crtrott, line 165:
```suggestion
      typename S::device_type,
```

core/src/Kokkos_HostSpace.hpp
```diff
@@ -138,39 +138,47 @@ namespace Kokkos {
...
       std::conditional_t<keep_mem,
                          Kokkos::Device<Kokkos::HostSpace::execution_space,
                                         typename S::memory_space>,
-                         Kokkos::HostSpace>>;
+                         Kokkos::Device<Kokkos::HostSpace::execution_space,
+                                        Kokkos::HostSpace::memory_space>>>;
```
crtrott, line 170:
```suggestion
                         Kokkos::HostSpace::device_type>>;
```

core/src/Kokkos_HostSpace.hpp
```diff
@@ -138,39 +138,47 @@ namespace Kokkos {
...
 namespace Impl {
 
-static_assert(Kokkos::Impl::MemorySpaceAccess<Kokkos::HostSpace,
-                                              Kokkos::HostSpace>::assignable);
-
-template <typename S>
+template <class S>
 struct HostMirror {
  private:
+  using is_space = typename Kokkos::is_space<S>;
```
crtrott, line 144: Let's not do this. Need a different name so we don't shadow stuff.

core/src/Kokkos_View.hpp
```diff
@@ -165,7 +165,7 @@ class View : public Impl::BasicViewFromTraits<DataType, Properties...>::type {
   using execution_space      = typename traits::execution_space;
   using memory_space         = typename traits::memory_space;
   using memory_traits        = typename traits::memory_traits;
-  using host_mirror_space    = typename traits::host_mirror_space;
```
crtrott, line 168: This is a public type or? Can't just remove it, need to deprecate it.

core/src/Kokkos_View.hpp
```diff
@@ -210,7 +210,7 @@ class View : public Impl::BasicViewFromTraits<DataType, Properties...>::type {
   using host_mirror_type =
       View<typename traits::non_const_data_type, typename traits::array_layout,
            Device<DefaultHostExecutionSpace,
-                  typename traits::host_mirror_space::memory_space>,
+                  typename traits::host_mirror_device::memory_space>,
```
crtrott, line 213: As I said for other classes this may need a deprecation guard.

crtrott: Don't you need to deprecate the `HostMirror` below?

core/src/View/Kokkos_ViewTraits.hpp
```diff
@@ -455,12 +459,12 @@ struct ViewTraits {
...
-  using device_type       = Kokkos::Device<ExecutionSpace, MemorySpace>;
-  using memory_traits     = MemoryTraits;
-  using host_mirror_space = HostMirrorSpace;
-  using hooks_policy      = HooksPolicy;
+  using execution_space    = ExecutionSpace;
+  using memory_space       = MemorySpace;
+  using device_type        = Kokkos::Device<ExecutionSpace, MemorySpace>;
+  using host_mirror_device = HostMirrorDevice;
+  using memory_traits      = MemoryTraits;
+  using hooks_policy       = HooksPolicy;
```
crtrott, line 467: HostMirrorSpace was a public typedef inside public ViewTraits, so it still needs to be here (in deprecated form).

core/unit_test/TestViewTypedefs.cpp
```diff
@@ -183,103 +183,78 @@ constexpr bool test_view_typedefs_impl() {
...
-  // HostMirrorSpace is a mess so: if the default exec is a host exec, that is it
-  using host_mirror_space = std::conditional_t<is_host_exec, Kokkos::DefaultExecutionSpace,
-  // otherwise if unified memory is not on its HostSpace
-                               std::conditional_t<!has_unified_mem_space, Kokkos::HostSpace,
-  // otherwise its the following Device type
-                               Kokkos::Device<Kokkos::DefaultHostExecutionSpace, typename Kokkos::DefaultExecutionSpace::memory_space>>>;
-  static_assert(test_view_typedefs<layout_type, space, memory_traits, host_mirror_space, int, int&>(
+  using space         = Kokkos::DefaultExecutionSpace;
+  using device        = Kokkos::Device<space, typename space::memory_space>;
+  using host_mirror_device = Kokkos::Impl::HostMirror<device>::Device;
```
crtrott, line 201: Same comment as in the other tests.

crtrott (changes requested):
I left a few comments: I do think too that we should split up introduction of `host_mirror_type` and deprecation of `HostMirror` and the actual change of what `HostMirror` is. I spent an hour reviewing this, and most of the files we could merge in if this was split up properly and I had just reviewed the renaming.

containers/unit_tests/TestDynRankViewTypedefs.cpp
janciesko: Yep, restoring the original  test / code.

algorithms/src/Kokkos_Random.hpp
janciesko: @romintomasetti  - I've moved changes regarding View::HostMirror to https://github.com/kokkos/kokkos/pull/8232
I was referring to Kokkos::Impl::HostMirror in my comment above which we'll handle in this PR. Sorry for the confusion.

core/src/Kokkos_HostSpace.hpp
janciesko: We have an explicit `( )` operator which returns `value`.

core/src/Kokkos_HostSpace.hpp
janciesko: Ups, oversight.

core/src/View/Kokkos_ViewTraits.hpp
janciesko: Fixed

core/unit_test/TestViewTypedefs.cpp
janciesko: Fixed

containers/src/Kokkos_UnorderedMap.hpp
janciesko: Fixed

core/src/Kokkos_View.hpp
janciesko: Yes, this deprecation is included in https://github.com/kokkos/kokkos/pull/8232 where we deal with View::HostMirror.

janciesko: For the above, we likely do not need a deprecation guard since the behavior did not change: memory_space is still memory_space in either HostMirror implementation.

core/src/Kokkos_View.hpp
janciesko: Fixed.

janciesko:
Closing in favor of:
- https://github.com/kokkos/kokkos/pull/8232
- https://github.com/kokkos/kokkos/pull/8243
- https://github.com/kokkos/kokkos/pull/8252

PR #7491, opened 2024-10-29
Use absolute paths for CXX compiler in launcher script

MarcelKoch:
This PR uses absolute paths for the CXX compiler and the third argument of the compiler launcher script.

Before, the following would not choose the `nvcc_wrapper`:
```
kokkos_launch_compiler nvcc_wrapper /usr/bin/g++ g++
```
even though `$(which g++) == /usr/bin/g++`.


crtrott (approved):
Seems reasonable

masterleinad (approved):

PR #7492, opened 2024-10-29
Ignore gcc assembler options `-Wa,*` in the nvcc_wrapper.

MarcelKoch:
This PR ignores the `-Wa,*` options passed to the nvcc_wrapper.

These options can't be passed through `-Xcompiler`, since those options are only for the gcc preprocessor or compiler. None of the other `-X<something>` nvcc options allow passing options to the gcc assembler, so ignoring the options seems like the only reasonable choice.

### DIFF
--- a/bin/nvcc_wrapper
+++ b/bin/nvcc_wrapper
@@ -500,8 +500,19 @@ do
     xlinker_args="$xlinker_args -Xlinker ${1:4:${#1}}"
     host_linker_args="$host_linker_args ${1:4:${#1}}"
     ;;
-  #Ignore options passed to host assembler, as nvcc has no forwarding option
+  #Handle host assembler options
   -Wa,*)
+    #To pass the -Wa options to the host compiler via -Xcompiler it is necessary
+    #to use '\\,' for each comma in the options. As users might already add escapes
+    #to the comma by themselves, the escapes are first removed and then only the
+    #required number of \ are added back.
+    xcompiler_args_wa=$(echo -e "$1" | sed -E 's/\\\+,/,/g' | sed -E 's/,/\\\\\\\,/g')
+    if [ $first_xcompiler_arg -eq 1 ]; then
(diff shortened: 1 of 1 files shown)


cwpearson:
I've never personally needed this before, but I observe that:

```bash
nvcc --compiler-options="-Wa,blah" ...
```
expands to (`-v`)
```
gcc ... -Wa blah ...
```
which I suspect is not what you want. However,

```bash
nvcc --compiler-options="-Wa\,blah" ...
```

(note escaped comma) expands to

```bash
gcc ...  -Wa,blah ...
```

which may be what you want?

cwpearson (dismissed):
I've never personally needed this before, but I observe that:

```bash
nvcc --compiler-options="-Wa,blah" ...
```
expands to (`-v`)
```
gcc ... -Wa blah ...
```
which I suspect is not what you want. However,

```bash
nvcc --compiler-options="-Wa\,blah" ...
```

(note escaped comma) expands to

```bash
gcc ...  -Wa,blah ...
```

which may be what you want?

MarcelKoch:
Thanks for the hints @cwpearson. I changed it now, so that the options are preserved. Since the wrapper uses `-Xcompiler` without additional quotes, it is necessary to replace the `,` with `\\,`.

crtrott (approved):

masterleinad:
I'm seeing
```
$ ./bin/nvcc_wrapper -Wa,blah,blub --show foo.cpp
nvcc -ccbin g++ -arch=sm_70 -Xcompiler -Wa\\\,blah\\\,blub -x cu foo.cpp
```
is that really what is intended?

MarcelKoch:
@masterleinad I'm no expert with bash, so I just added enough `\` until the CXX compiler stopped complaining. When `--show` shows only `\\,`, then the compiler still fails. I'm not sure if this has to do with how the `\` are rendered, but when `--show` shows `\\\,` it works.

PR #7493, opened 2024-10-29
Use atomics for volatile store loads to fix Threads on ARM

### LINKED ISSUE #7090: Kokkos::Threads backend is broken on M2 MacBook (ARM) in develop

crtrott:
This should fix threads backend on ARM.
Fixes #7090

### DIFF
--- a/core/src/Kokkos_MemoryPool.hpp
+++ b/core/src/Kokkos_MemoryPool.hpp
@@ -498,13 +498,14 @@ class MemoryPool {
     //   hint_sb_id_ptr[0] is the dynamically changing hint
     //   hint_sb_id_ptr[1] is the static start point
 
-    volatile uint32_t *const hint_sb_id_ptr =
+    uint32_t *const hint_sb_id_ptr =
         m_sb_state_array      /* memory pool state array */
         + m_hint_offset       /* offset to hint portion of array */
         + HINT_PER_BLOCK_SIZE /* number of hints per block size */
               * (block_size_lg2 - m_min_block_size_lg2); /* block size id */
 
-    const int32_t sb_id_begin = int32_t(hint_sb_id_ptr[1]);
+    const int32_t sb_id_begin =
+        int32_t(Kokkos::atomic_load(&hint_sb_id_ptr[1]));
 
     // Fast query clock register 'tic' to pseudo-randomize
     // the guess for which block within a superblock should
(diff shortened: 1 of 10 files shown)


masterleinad:
Still seeing
```
2: [ RUN      ] threads.team_scratch_request
2: /Users/6da/Software/kokkos_new/core/unit_test/TestTeam.hpp:908: Failure
2: Expected equality of these values:
2:   error_count
2:     Which is: 34
2:   0
```
after running a couple of times.

masterleinad:
```
[ RUN      ] threads.unique_token_team_acquire
/tmp/kokkos/core/unit_test/TestUniqueToken.hpp:246: Failure
  host_errors(0)
    Which is: 1
  0
max reuse was 136
[  FAILED  ] threads.unique_token_team_acquire (0 ms)
```

janciesko:
Still:
```
[ RUN      ] threads.unique_token_team_acquire
/Users/jciesko/RAndD/kokkos_round2/core/unit_test/TestUniqueToken.hpp:246: Failure
  host_errors(0)
    Which is: 1
  0
max reuse was 111
[  FAILED  ] threads.unique_token_team_acquire (7 ms)
```

janciesko:
I glanced over the internal_fence and I wonder if we understand how we include the root thread in this fence. It seems that we set the state of the root thread to inactive in L306. The root thread (`s_threads_exec[0]`) then immediately enters the fence and likely proceeds before other workers reach that fence. 

```C++
// Wait for root thread to become inactive
void ThreadsInternal::internal_fence() {
  memory_fence();
  if (s_thread_pool_size[0]) {
    // Wait for the root thread to complete:
    Impl::spinwait_while_equal(s_threads_exec[0]->m_pool_state,
                               ThreadState::Active);
  }

  Kokkos::atomic_store(&s_current_function, nullptr);
  Kokkos::atomic_store(&s_current_function_arg, nullptr);

  // Make sure function and arguments are cleared before
  // potentially re-activating threads with a subsequent launch.
  memory_fence();
}
```

janciesko:
Alternatively, get rid of more volatile types in favor of atomics.

PR #7498, opened 2024-10-30
Fixing Threads backend on ARM - use std::atomic for thread coordination

crtrott:
Fixes threads backend failures on ARM.
This approach seems to be most consistently to fix the issues. I suspect that largely that is due to inserting seqcst load/stores in critical places.

I did some performance testing and didn't see any obvious bad things. In fact this is as fast as develop, while the other PR #7493 show significant slowdowns for this code:

```c++
#include<Kokkos_Core.hpp>

  {

    int N = 100000;
    int R = 100;
    Kokkos::View<double**> A("A",N, R);

    Kokkos::Timer timer;
    Kokkos::parallel_for(Kokkos::TeamPolicy<>(N, 4), KOKKOS_LAMBDA(const typename Kokkos::TeamPolicy<>::member_type& team) {
      for(int r=0; r<R; r++) {
        double result;
        Kokkos::parallel_reduce(Kokkos::TeamThreadRange(team, 0, 100), [&](int i, double& lres) {
          lres += i%r;
        },result);
        A(team.league_rank(), r) = result;
      }
    });
    printf("%lf %lf\n",A(1,1),time);
  }
  Kokkos::finalize();
}
```

### DIFF
--- a/core/src/Threads/Kokkos_Threads_Instance.cpp
+++ b/core/src/Threads/Kokkos_Threads_Instance.cpp
@@ -67,8 +67,9 @@ std::pair<unsigned, unsigned>
 
 int s_thread_pool_size[3] = {0, 0, 0};
 
-void (*volatile s_current_function)(ThreadsInternal &, const void *);
-const void *volatile s_current_function_arg = nullptr;
+using s_current_function_type = void (*)(ThreadsInternal &, const void *);
+std::atomic<s_current_function_type> s_current_function;
+std::atomic<const void *> s_current_function_arg = nullptr;
 
 inline unsigned fan_size(const unsigned rank, const unsigned size) {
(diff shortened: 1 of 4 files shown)


masterleinad:
> This approach seems to be most consistently to fix the issues. I suspect that largely that is due to inserting seqcst load/stores in critical places.

I ran the `Kokkos_CoreUnitTest_Threads` more than 150 times and `threads.workgraph_fib` 1000 times without failure on a Grace CPU with 32 threads.

janciesko:
Seems to make it work also on the Apple M1.

janciesko (approved):

masterleinad (approved):

romintomasetti:
I would still change the PR title before merging @crtrott :wink:

PR #7501, opened 2024-11-01
Try junit in jenkins

masterleinad:


masterleinad:
This doen't appear to give any better output than the xunit approach we are currently using and would require more modern CMake versions.

PR #7502, opened 2024-11-01
oneDPL: Sort on device using Kokkos::RandomAccessIterator

masterleinad:
Currently, `Kokkos::sort/sort_by_key` perform the sort on the device using oneDPL for SYCL only if the data is contiguous. This pull request removes that restrictions by using `RandomAccessIterator` like we do for `Cuda` and `HIP`.
The key to making this work is that `oneDPL` avoids assuming host data for the given random access iterator if it has an `is_passed_directly = std::true_type` alias, see, e.g., https://github.com/oneapi-src/oneDPL/issues/855#issuecomment-1483643869. A less obscure approach is proposed in https://github.com/oneapi-src/oneDPL/pull/861. With this approach I ran into some compilation errors, see https://github.com/oneapi-src/oneDPL/issues/1922 which were eventually resolved in https://github.com/oneapi-src/oneDPL/pull/1927. Thus, we need a `oneDPL` that hasn't appeared in any tag or release for this pull request to work.

Based on top of #7304 and #7451.


masterleinad:
This fixes some performance problems in `LAMMPS` as confirmed by @rarensu.

algorithms/src/sorting/impl/Kokkos_SortByKeyImpl.hpp
```diff
@@ -142,8 +142,10 @@ void sort_by_key_rocthrust(
 
 #if defined(KOKKOS_ENABLE_ONEDPL)
 
-#if ONEDPL_VERSION_MAJOR > 2022 || \
-    (ONEDPL_VERSION_MAJOR == 2022 && ONEDPL_VERSION_MINOR > 7)
+#if ONEDPL_VERSION_MAJOR > 2022 ||   \
+    (ONEDPL_VERSION_MAJOR == 2022 && \
+     (ONEDPL_VERSION_MINOR > 7 ||    \
+      (ONEDPL_VERSION_MINOR == 7 && ONEDPL_VERSION_PATCH >= 1)))
```
dalg24, line 148: This is so ugly that it makes me think we might want to define some semantic versioning comparison macros.
(Obviously not blocking and would need to be done elsewhere)

In the meantime consider defining a `KOKKOS_IMPL_SYCL_ALWAYS_USE_ONEDPL_SORTING` or whatnot

algorithms/src/std_algorithms/impl/Kokkos_RandomAccessIterator.hpp
```diff
@@ -63,6 +63,10 @@ class RandomAccessIterator<::Kokkos::View<DataType, Args...>> {
   using pointer           = typename view_type::pointer_type;
   using reference         = typename view_type::reference_type;
 
+#ifdef KOKKOS_ENABLE_SYCL
+  using is_passed_directly = std::true_type;
+#endif
```
dalg24, line 68: What is that for?

masterleinad: As the pull request description says, this is needed to tell `oneDPL` not to assume that the data is on the host but on the device.

dalg24: Please add a comment in the code to capture that this tag is picked up by oneDPL with a reference to https://github.com/uxlfoundation/oneDPL/blob/a045eac689f9107f50ba7b42235e9e927118e483/include/oneapi/dpl/pstl/hetero/dpcpp/utils_ranges_sycl.h#L210-L214

algorithms/src/sorting/impl/Kokkos_SortByKeyImpl.hpp
```diff
@@ -65,13 +65,19 @@
...
 
-#if defined(KOKKOS_ENABLE_ONEDPL) && \
-    (ONEDPL_VERSION_MAJOR > 2022 ||  \
-     (ONEDPL_VERSION_MAJOR == 2022 && ONEDPL_VERSION_MINOR >= 2))
+#ifdef KOKKOS_ENABLE_ONEDPL
+#define KOKKOS_ONEDPL_VERSION                                 \
+  ONEDPL_VERSION_MAJOR * 10000 + ONEDPL_VERSION_MINOR * 100 + \
+      ONEDPL_VERSION_PATCH
+#define KOKKOS_ONEDPL_VERSION_GREATER_EQUAL(MAJOR, MINOR, PATCH) \
+  (KOKKOS_ONEDPL_VERSION >= ((MAJOR)*10000 + (MINOR)*100 + (PATCH)))
```
dalg24, line 73: This is not what I suggested in https://github.com/kokkos/kokkos/pull/7502#discussion_r1939937540

I would still slightly prefer having a named macro that more clearly expresses what we guard for and let us avoid to repeat the version (not blocking though).
Please add the `KOKKOS_IMPL_*` prefix.

dalg24 (approved):

Rombur (approved):

dalg24:
Make sure you add this to the 4.6 changelog

PR #7509, opened 2024-11-04
HIP: prefer smaller block size

Rombur:
This PR fixes an issue reported by AMD and the AthenaK team. Unlike the CUDA backend, in the HIP backend we prefer larger block size. Because of this when people use a TeamPolicy with `Kokkos::AUTO`, we also prefer large team size and we end up launching too many threads. This PR changes the behavior to prefer smaller block size, similar to what is done in CUDA.

### DIFF
--- a/core/src/HIP/Kokkos_HIP_BlockSize_Deduction.hpp
+++ b/core/src/HIP/Kokkos_HIP_BlockSize_Deduction.hpp
@@ -115,7 +115,8 @@ unsigned hip_internal_get_block_size(const HIPInternal *hip_instance,
   const unsigned min_threads_per_sm = min_waves_per_eu * HIPTraits::WarpSize;
   const unsigned shmem_per_sm =
       hip_instance->m_deviceProp.maxSharedMemoryPerMultiProcessor;
-  unsigned block_size = tperb_reg;
+  unsigned block_size     = tperb_reg;
+  unsigned min_block_size = 0;
   do {
     unsigned total_shmem = f(block_size);
     // find how many threads we can fit with this blocksize based on LDS usage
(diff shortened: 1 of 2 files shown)


crtrott (changes requested):
If the register limited block size is large we already exit early right? So you wouldn't get the smaller block size after all? Or is this just trying to fix an invalid configuration? And when would you end up invalid?

Rombur:
Yes, we still exit early if we use too many register. The issue that AthenaK has is that the code is much slower with new versions of ROCm compared to old ones. The reason is that the compiler uses fewer registers and so the block size is not register-bound anymore. Because of that the team-size jumps from 256 to 1024 which really slows down their application.

crtrott:
No I mean in that case Athena describes don't we still exit early and end up with 1024?

Rombur:
I am confused about your questions. The way the code works is the following: 
 - we check if we will spill register
 - if we spill, we pick a block size of 256 (This is what AthenaK gets with the old ROCm version)
 - if we don't spill register, we compute the block size (This is what AthenaK gets with the new ROCm version)

In the old block size deduction function, we would pick the largest block size that works given the shared memory required by the functor. In the new function, we don't exit the loop if we find a block size that works. We save the value and keep going.

crtrott:
This:
```
   unsigned block_size     = tperb_reg;
   unsigned min_block_size = 0;
   do {
     unsigned total_shmem = f(block_size);
     // find how many threads we can fit with this blocksize based on LDS usage
    unsigned tperb_shmem = total_shmem > shmem_per_sm ? 0 : block_size;
    if constexpr (BlockSize == BlockType::Max) {
      // we want the maximum blocksize possible
      // just wait until we get a case where we can fit the LDS per SM
      if (tperb_shmem) return block_size;
    } else {
      if (block_size == tperb_reg && tperb_shmem >= tperb_reg) {
        // fast path for exit on first iteration if registers are more limiting
        // than LDS usage, just use the register limited size
        return tperb_reg;
      }
```

If `tperb_reg == 1024` and `tperb_shmem == 1024` you return 1024 or?

core/src/HIP/Kokkos_HIP_ParallelReduce_Range.hpp
```diff
@@ -241,7 +241,7 @@ class ParallelReduce<CombinedFunctorReducerType, Kokkos::RangePolicy<Traits...>,
       dim3 block(1, block_size, 1);
       // use a slightly less constrained, but still well bounded limit for
       // scratch
-      int nblocks = (nwork + block.y - 1) / block.y;
+      index_type nblocks = (nwork + block.y - 1) / block.y;
```
dalg24, line 244: What's the deal with this change and the conversion below?

core/src/HIP/Kokkos_HIP_BlockSize_Deduction.hpp
```diff
@@ -126,37 +126,28 @@ unsigned hip_internal_get_block_size(const HIPInternal *hip_instance,
...
-          tperb > HIPTraits::ConservativeThreadsPerBlock) {
-        return block_size;
+      // The logic prefers smaller blocks sizes over larger ones to give more
+      // flexibility to the scheduler and to decrease the number of threads
+      // launched when using Kokkos::AUTO in TeamPolicy. If the block size is
+      // smaller than 256, fall back to BlockType::Max condition.
+      if (blocks_per_cu_shmem > min_waves_per_eu &&
+          tperb >= HIPTraits::ConservativeThreadsPerBlock) {
+        min_block_size = block_size;
+      } else if ((min_block_size == 0) && (tperb_shmem)) {
```
dalg24, line 142: Should this be checking that tperb is non null (instead of tperb_shmem)?

core/src/HIP/Kokkos_HIP_ParallelReduce_Range.hpp
Rombur: Now that we favor smaller bloc size, there are cases where we overflow `nblock`. The conversion is because the two arguments of `std::min` need to be the same type.

crtrott (approved):
Ok I think I am on-board now with the logic in here as in the last version. I tried writing a test code exploring the range of parameters for registers vs shmem but failed to produce a code which can reliably change registers used. One thing I noticed however (which is orthogonal from this PR) is that team_policy.team_size_max(f, tag) throws if it can't find a valid team_size. I think that needs to get addressed but in a different PR?

Rombur:
> One thing I noticed however (which is orthogonal from this PR) is that team_policy.team_size_max(f, tag) throws if it can't find a valid team_size. I think that needs to get addressed but in a different PR?

Yes, I'll do it in a followup

dalg24:
Retest this please

dalg24 (approved):

PR #7514, opened 2024-11-05
Fix Kokkos::resize with View of Views

### LINKED ISSUE #7466: SequentialHostInit issues with resize and realloc

masterleinad:
Fixes https://github.com/kokkos/kokkos/issues/7466.
As discussed in the meeting last week (https://github.com/kokkos/kokkos/issues/7466#issuecomment-2447610652), this pull request runs the copy operation sequentially on the host. To be able to use the existing infrastructure for using MDRangePolicy/HostIterateTile the attribute needs to be propagated to a number of layers. To make this feasible, about half of the commits simplify classes in `Kokkos_CopyViews.hpp`. These changes can, of course, be merged first independently if preferred.

### DIFF
--- a/core/src/Kokkos_CopyViews.hpp
+++ b/core/src/Kokkos_CopyViews.hpp
@@ -314,9 +314,10 @@ struct ViewFill<ViewType, Layout, ExecSpace, 8, iType> {
   }
 };
 
-template <class ViewTypeA, class ViewTypeB, class Layout, class ExecSpace,
-          typename iType>
-struct ViewCopy<ViewTypeA, ViewTypeB, Layout, ExecSpace, 1, iType> {
+template <bool SequentialHostInit, class ViewTypeA, class ViewTypeB,
+          class Layout, class ExecSpace, typename iType>
+struct ViewCopy<SequentialHostInit, ViewTypeA, ViewTypeB, Layout, ExecSpace, 1,
+                iType> {
   ViewTypeA a;
   ViewTypeB b;
 
@@ -341,9 +342,10 @@ struct ViewCopy<ViewTypeA, ViewTypeB, Layout, ExecSpace, 1, iType> {
   }
 };
 
-template <class ViewTypeA, class ViewTypeB, class Layout, class ExecSpace,
-          typename iType>
-struct ViewCopy<ViewTypeA, ViewTypeB, Layout, ExecSpace, 2, iType> {
+template <bool SequentialHostInit, class ViewTypeA, class ViewTypeB,
+          class Layout, class ExecSpace, typename iType>
+struct ViewCopy<SequentialHostInit, ViewTypeA, ViewTypeB, Layout, ExecSpace, 2,
+                iType> {
   ViewTypeA a;
   ViewTypeB b;
   static const Kokkos::Iterate outer_iteration_pattern =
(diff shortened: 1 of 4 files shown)


masterleinad:
Retest this please

janciesko:
I think this approach for host-serializing `resize` via differentiating in view_copy is good.

janciesko (approved):

core/src/Kokkos_Core_fwd.hpp
```diff
@@ -269,8 +269,8 @@ template <class ViewType, class Layout = typename ViewType::array_layout,
           int Rank = ViewType::rank, typename iType = int64_t>
 struct ViewFill;
 
-template <class ViewTypeA, class ViewTypeB, class Layout, class ExecSpace,
-          int Rank, typename iType>
+template <bool SequentialHostInit, class ViewTypeA, class ViewTypeB,
+          class Layout, class ExecSpace, int Rank, typename iType>
```
ldh4: Any thoughts on defaulting `SequentialHostInit` template parameter to false, so that we avoid needing to construct an instance of ViewCopy starting always with a bool parameter?

core/src/Kokkos_CopyViews.hpp
```diff
@@ -748,11 +860,12 @@ struct CommonSubview {
       : dst_sub(dst, args...), src_sub(src, args...) {}
 };
 
-template <class DstType, class SrcType, int Rank = DstType::rank>
+template <bool SequentialHostInit, class DstType, class SrcType,
+          int Rank = DstType::rank>
```
ldh4: Same here. Any opinion on defaulting the `SequentialHostInit` parameter to false?

ldh4 (approved):
Looks good to me.
Minor complaint on the `SequentialHostInit` parameter. But okay with it either way.

core/src/Kokkos_Core_fwd.hpp
janciesko: Sounds good to me. We'd need to update the template param order.

janciesko: I'll let @masterleinad chime in here as well.

masterleinad: IIRC it was just convenient for making sure we catch all places. We can swap back.

crtrott:
Lets delay until after 5.2: risk of subtle downstream issues.

crtrott (changes requested):
Coming back to this: assuming we still need this (we do right?) I don't think the name SequentialHostInit is really good at the `view_copy` function and `ViewCopy` class level. These functions don't "init" they copy. The name is appropriate for `ViewRemap`. I also want to see sanity check in `view_copy` and `ViewCopy` that the Views are host accessible when you have that "SequentialHostExec" flag set. I.e. static_assert that if that thing is true, that the views are accesible something like `static_assert((SequentialHostExec==false) || SpacAccessibility<HostSpace, ViewSpace>::accessible)` or however that needs to be spelled.

Also the order of template arguments is partially iffy. I get that one would want that new flag up front for `view_copy` (so we don't have to spell out the view argument types) and I agree with that. But on the `ViewCopy` partial specializations make the order of template args the same as it appears in the actual specialization (e.g. line 562, put the new bool last).

PR #7518, opened 2024-11-06
Only use public `deep_copy` inside containers

pzehner:
This PR is a refactor which is part of a greater effort to simplify the use of `Impl::DeepCopy`.

It replaces the `Impl::DeepCopy` used in containers by the public `deep_copy`.

### DIFF
--- a/containers/src/Kokkos_Bitset.hpp
+++ b/containers/src/Kokkos_Bitset.hpp
@@ -400,10 +400,7 @@ void deep_copy(Bitset<DstDevice>& dst, Bitset<SrcDevice> const& src) {
     Kokkos::Impl::throw_runtime_exception(
         "Error: Cannot deep_copy bitsets of different sizes!");
   }
-
-  Kokkos::fence("Bitset::deep_copy: fence before copy operation");
   Kokkos::deep_copy(dst.m_blocks, src.m_blocks);
-  Kokkos::fence("Bitset::deep_copy: fence after copy operation");
 }
 
 template <typename DstDevice, typename SrcDevice>
@@ -412,10 +409,7 @@ void deep_copy(Bitset<DstDevice>& dst, ConstBitset<SrcDevice> const& src) {
     Kokkos::Impl::throw_runtime_exception(
         "Error: Cannot deep_copy bitsets of different sizes!");
   }
-
-  Kokkos::fence("Bitset::deep_copy: fence before copy operation");
   Kokkos::deep_copy(dst.m_blocks, src.m_blocks);
-  Kokkos::fence("Bitset::deep_copy: fence after copy operation");
 }
 
 template <typename DstDevice, typename SrcDevice>
@@ -424,10 +418,7 @@ void deep_copy(ConstBitset<DstDevice>& dst, ConstBitset<SrcDevice> const& src) {
     Kokkos::Impl::throw_runtime_exception(
         "Error: Cannot deep_copy bitsets of different sizes!");
   }
-
-  Kokkos::fence("Bitset::deep_copy: fence before copy operation");
   Kokkos::deep_copy(dst.m_blocks, src.m_blocks);
-  Kokkos::fence("Bitset::deep_copy: fence after copy operation");
 }
 
 }  // namespace Kokkos
--- a/containers/src/Kokkos_Bitset.hpp
+++ b/containers/src/Kokkos_Bitset.hpp
@@ -135,9 +135,9 @@ class Bitset {
 
     if (m_last_block_mask) {
       // clear the unused bits in the last block
-      Kokkos::Impl::DeepCopy<typename Device::memory_space, Kokkos::HostSpace>(
-          m_blocks.data() + (m_blocks.extent(0) - 1u), &m_last_block_mask,
-          sizeof(unsigned));
+      auto last_block = Kokkos::subview(m_blocks, m_blocks.extent(0) - 1u);
+      Kokkos::deep_copy(typename Device::execution_space{}, last_block,
+                        m_last_block_mask);
       Kokkos::fence(
           "Bitset::set: fence after clearing unused bits copying from "
           "HostSpace");
@@ -402,10 +402,7 @@ void deep_copy(Bitset<DstDevice>& dst, Bitset<SrcDevice> const& src) {
   }
 
   Kokkos::fence("Bitset::deep_copy: fence before copy operation");
-  Kokkos::Impl::DeepCopy<typename DstDevice::memory_space,
-                         typename SrcDevice::memory_space>(
-      dst.m_blocks.data(), src.m_blocks.data(),
-      sizeof(unsigned) * src.m_blocks.extent(0));
+  Kokkos::deep_copy(dst.m_blocks, src.m_blocks);
   Kokkos::fence("Bitset::deep_copy: fence after copy operation");
 }
 
@@ -417,10 +414,7 @@ void deep_copy(Bitset<DstDevice>& dst, ConstBitset<SrcDevice> const& src) {
   }
 
   Kokkos::fence("Bitset::deep_copy: fence before copy operation");
-  Kokkos::Impl::DeepCopy<typename DstDevice::memory_space,
-                         typename SrcDevice::memory_space>(
-      dst.m_blocks.data(), src.m_blocks.data(),
-      sizeof(unsigned) * src.m_blocks.extent(0));
+  Kokkos::deep_copy(dst.m_blocks, src.m_blocks);
   Kokkos::fence("Bitset::deep_copy: fence after copy operation");
 }
 
@@ -432,10 +426,7 @@ void deep_copy(ConstBitset<DstDevice>& dst, ConstBitset<SrcDevice> const& src) {
   }
 
   Kokkos::fence("Bitset::deep_copy: fence before copy operation");
-  Kokkos::Impl::DeepCopy<typename DstDevice::memory_space,
-                         typename SrcDevice::memory_space>(
-      dst.m_blocks.data(), src.m_blocks.data(),
-      sizeof(unsigned) * src.m_blocks.extent(0));
+  Kokkos::deep_copy(dst.m_blocks, src.m_blocks);
   Kokkos::fence("Bitset::deep_copy: fence after copy operation");
 }
 
--- a/containers/src/Kokkos_DynamicView.hpp
+++ b/containers/src/Kokkos_DynamicView.hpp
@@ -195,11 +195,13 @@ struct ChunkedArrayManager {
   void deep_copy_to(
       const ExecutionSpace& exec_space,
       ChunkedArrayManager<OtherMemorySpace, ValueType> const& other) const {
-    if (other.m_chunks != m_chunks) {
-      Kokkos::Impl::DeepCopy<OtherMemorySpace, MemorySpace, ExecutionSpace>(
-          exec_space, other.m_chunks, m_chunks,
-          sizeof(pointer_type) * (m_chunk_max + 2));
-    }
+    // use of ad-hoc unmanaged views
+    Kokkos::deep_copy(
+        exec_space,
+        Kokkos::View<uintptr_t*, OtherMemorySpace>(
+            reinterpret_cast<uintptr_t*>(other.m_chunks), m_chunk_max + 2),
+        Kokkos::View<uintptr_t*, MemorySpace>(
+            reinterpret_cast<uintptr_t*>(m_chunks), m_chunk_max + 2));
   }
 
   KOKKOS_INLINE_FUNCTION
--- a/containers/src/Kokkos_UnorderedMap.hpp
+++ b/containers/src/Kokkos_UnorderedMap.hpp
@@ -874,22 +874,16 @@ class UnorderedMap {
     if (m_hash_lists.data() != src.m_hash_lists.data()) {
       Kokkos::deep_copy(m_available_indexes, src.m_available_indexes);
 
-      using raw_deep_copy =
-          Kokkos::Impl::DeepCopy<typename device_type::memory_space,
-                                 typename SDevice::memory_space>;
+      // do the other deep copies asynchronously if possible
+      typename device_type::execution_space exec_space{};
 
-      raw_deep_copy(m_hash_lists.data(), src.m_hash_lists.data(),
-                    sizeof(size_type) * src.m_hash_lists.extent(0));
-      raw_deep_copy(m_next_index.data(), src.m_next_index.data(),
-                    sizeof(size_type) * src.m_next_index.extent(0));
-      raw_deep_copy(m_keys.data(), src.m_keys.data(),
-                    sizeof(key_type) * src.m_keys.extent(0));
+      Kokkos::deep_copy(exec_space, m_hash_lists, src.m_hash_lists);
+      Kokkos::deep_copy(exec_space, m_next_index, src.m_next_index);
+      Kokkos::deep_copy(exec_space, m_keys, src.m_keys);
       if (!is_set) {
-        raw_deep_copy(m_values.data(), src.m_values.data(),
-                      sizeof(impl_value_type) * src.m_values.extent(0));
+        Kokkos::deep_copy(exec_space, m_values, src.m_values);
       }
-      raw_deep_copy(m_scalars.data(), src.m_scalars.data(),
-                    sizeof(int) * num_scalars);
+      Kokkos::deep_copy(exec_space, m_scalars, src.m_scalars);
 
       Kokkos::fence(
           "Kokkos::UnorderedMap::deep_copy_view: fence after copy to dst.");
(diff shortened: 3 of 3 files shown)


containers/src/Kokkos_Bitset.hpp
```diff
@@ -135,9 +135,13 @@ class Bitset {
 
     if (m_last_block_mask) {
       // clear the unused bits in the last block
-      Kokkos::Impl::DeepCopy<typename Device::memory_space, Kokkos::HostSpace>(
-          m_blocks.data() + (m_blocks.extent(0) - 1u), &m_last_block_mask,
-          sizeof(unsigned));
+      Kokkos::parallel_for(
+          "clear the unused bits in the last block",
```
Rombur, line 139: The label should be prefixed with `Kokkos::`

containers/src/Kokkos_UnorderedMap.hpp
```diff
@@ -901,33 +924,34 @@ class UnorderedMap {
   bool modified() const { return get_flag(modified_idx); }
 
   void set_flag(int flag) const {
-    using raw_deep_copy =
-        Kokkos::Impl::DeepCopy<typename device_type::memory_space,
-                               Kokkos::HostSpace>;
-    const int true_ = true;
-    raw_deep_copy(m_scalars.data() + flag, &true_, sizeof(int));
+    Kokkos::parallel_for(
+        "set flag",
```
Rombur, line 928: Here and below prefix the label  with `Kokkos::`

containers/src/Kokkos_Bitset.hpp
pzehner: Done.

containers/src/Kokkos_UnorderedMap.hpp
pzehner: Done.

pzehner:
I wonder if the fences are still necessary.

containers/src/Kokkos_Bitset.hpp
```diff
@@ -402,10 +406,7 @@ void deep_copy(Bitset<DstDevice>& dst, Bitset<SrcDevice> const& src) {
   }
 
   Kokkos::fence("Bitset::deep_copy: fence before copy operation");
-  Kokkos::Impl::DeepCopy<typename DstDevice::memory_space,
-                         typename SrcDevice::memory_space>(
-      dst.m_blocks.data(), src.m_blocks.data(),
-      sizeof(unsigned) * src.m_blocks.extent(0));
+  Kokkos::deep_copy(dst.m_blocks, src.m_blocks);
   Kokkos::fence("Bitset::deep_copy: fence after copy operation");
```
masterleinad, line 410: You don't eed the fences here anymore.

containers/src/Kokkos_Bitset.hpp
```diff
@@ -417,10 +418,7 @@ void deep_copy(Bitset<DstDevice>& dst, ConstBitset<SrcDevice> const& src) {
...
```
masterleinad, line 422: These can go as well.

containers/src/Kokkos_Bitset.hpp
```diff
@@ -432,10 +430,7 @@ void deep_copy(ConstBitset<DstDevice>& dst, ConstBitset<SrcDevice> const& src) {
...
```
masterleinad, line 434: ... and here

containers/src/Kokkos_UnorderedMap.hpp
```diff
@@ -874,22 +906,13 @@ class UnorderedMap {
...
-        raw_deep_copy(m_values.data(), src.m_values.data(),
-                      sizeof(impl_value_type) * src.m_values.extent(0));
+        Kokkos::deep_copy(m_values, src.m_values);
       }
-      raw_deep_copy(m_scalars.data(), src.m_scalars.data(),
-                    sizeof(int) * num_scalars);
+      Kokkos::deep_copy(m_scalars, src.m_scalars);
 
       Kokkos::fence(
           "Kokkos::UnorderedMap::deep_copy_view: fence after copy to dst.");
```
masterleinad, line 918: This fence can also be removed.

containers/src/Kokkos_UnorderedMap.hpp
```diff
@@ -901,33 +924,34 @@ class UnorderedMap {
...
-        Kokkos::Impl::DeepCopy<Kokkos::HostSpace,
-                               typename device_type::memory_space>;
-    int result = false;
-    raw_deep_copy(&result, m_scalars.data() + flag, sizeof(int));
+    bool result = false;
+    Kokkos::parallel_reduce(
+        "Kokkos::get flag",
+        Kokkos::RangePolicy(typename device_type::execution_space{}, flag,
+                            flag + 1),
+        Impl::FlagGetFunctor(m_scalars), result);
```
masterleinad, line 954: Why can't this be a lambda?

containers/src/Kokkos_UnorderedMap.hpp
```diff
@@ -901,33 +924,34 @@ class UnorderedMap {
...
-    using raw_deep_copy =
-        Kokkos::Impl::DeepCopy<typename device_type::memory_space,
-                               Kokkos::HostSpace>;
-    const int false_ = false;
-    raw_deep_copy(m_scalars.data() + flag, &false_, sizeof(int));
+    Kokkos::parallel_for(
+        "Kokkos::reset flag",
+        Kokkos::RangePolicy(typename device_type::execution_space{}, flag,
+                            flag + 1),
+        Impl::FlagSetFunctor(m_scalars, false));
```
masterleinad, line 942: Why can't this be a lambda?

containers/src/Kokkos_UnorderedMap.hpp
```diff
@@ -901,33 +924,34 @@ class UnorderedMap {
...
-    using raw_deep_copy =
-        Kokkos::Impl::DeepCopy<typename device_type::memory_space,
-                               Kokkos::HostSpace>;
-    const int true_ = true;
-    raw_deep_copy(m_scalars.data() + flag, &true_, sizeof(int));
+    Kokkos::parallel_for(
+        "Kokkos::set flag",
+        Kokkos::RangePolicy(typename device_type::execution_space{}, flag,
+                            flag + 1),
+        Impl::FlagSetFunctor(m_scalars, true));
```
masterleinad, line 931: Why can't this be a lambda?

containers/src/Kokkos_UnorderedMap.hpp
```diff
@@ -148,6 +148,38 @@ struct UnorderedMapInsertOpTypes {
...
+struct FlagGetFunctor {
+  View m_view;
+
+  explicit FlagGetFunctor(View const &view) : m_view(view) {}
+
+  KOKKOS_FUNCTION
+  void operator()(unsigned const i, bool &value_inner) const {
+    value_inner = static_cast<bool>(m_view(i));
+  }
+};
```
masterleinad, line 179: These are so short that it hsould be fine to inline them.

containers/src/Kokkos_DynamicView.hpp
```diff
@@ -195,11 +195,13 @@ struct ChunkedArrayManager {
...
-          exec_space, other.m_chunks, m_chunks,
-          sizeof(pointer_type) * (m_chunk_max + 2));
-    }
+    // use of ad-hoc unmanaged views
+    Kokkos::deep_copy(
+        exec_space,
+        Kokkos::View<value_type*, OtherMemorySpace>(
+            reinterpret_cast<value_type*>(other.m_chunks), m_chunk_max + 2),
+        Kokkos::View<value_type*, MemorySpace>(
+            reinterpret_cast<value_type*>(m_chunks), m_chunk_max + 2));
```
masterleinad, line 204: Why is the `interpret_cast` the right thing to do here?

containers/src/Kokkos_UnorderedMap.hpp
pzehner: Because when doing so, NVCC complains with:

```
error: The enclosing parent function ("get_flag") for an extended __host__ __device__ lambda cannot have private or protected access within its class
```

containers/src/Kokkos_UnorderedMap.hpp
pzehner: Same as above.

containers/src/Kokkos_UnorderedMap.hpp
pzehner: Same as above.

containers/src/Kokkos_Bitset.hpp
pzehner: Done.

containers/src/Kokkos_Bitset.hpp
pzehner: Done.

containers/src/Kokkos_Bitset.hpp
pzehner: Done.

containers/src/Kokkos_DynamicView.hpp
pzehner: `m_chunks` is a double pointer that is treated as `void *` by `Dynk::DeepCopy`. `Kokkos::View`, even unmanaged, requires to set the appropriate rank and dimensions, the later being unknown. So I use `reinterpret_cast` to force treating `m_chunks` as 1-rank data. As the view is temporary and only used for the call of `deep_copy`, I think it is OK to do so.

containers/src/Kokkos_UnorderedMap.hpp
pzehner: Do you mean using `inline`?

Rombur (approved):

containers/src/Kokkos_DynamicView.hpp
masterleinad: If I'm not mistaken, the original call did something like
```C++
memcpy(reinterpret_cast<pointer_type*>(&other.m_chunks), reinterpret_cast<pointer_type*>(&m_chunks), sizeof(pointer_type) * (m_chunk_max + 2));
```
(where the reinterpret_cast is irrelevant of course)
while the new implementation does
```C++
memcpy(reinterpret_cast<value_type*>(&other.m_chunks), reinterpret_cast<value_type*>(&m_chunks), sizeof(value_type) * (m_chunk_max + 2));
```
and these would only do the same if `sizeof(value_type) == `sizeof(pointer_type)`. In other words, the Views data type should be `pointer_type` (which is problematic since int** would be treated as a two-dimensional View) or something with the same size like `uintptr_t`.

containers/src/Kokkos_UnorderedMap.hpp
pzehner: Eventually replaced it by a call to `deep_copy` + a scalar subview.

containers/src/Kokkos_UnorderedMap.hpp
pzehner: Eventually replaced it by a call to `deep_copy` + a scalar subview.

containers/src/Kokkos_UnorderedMap.hpp
pzehner: Eventually replaced it by a call to `deep_copy` + a scalar subview.

containers/src/Kokkos_UnorderedMap.hpp
pzehner: Eventually replaced it by a call to `deep_copy` + a scalar subview.

containers/src/Kokkos_DynamicView.hpp
pzehner: To clarify things out, `pointer_type` is the same as `value_type*`, and `m_chunks` is a `pointer_type*`.

I see what you mean, so my PR is very likely copying a wrong amount of data (since `sizeof(value_type*) == sizeof(pointer_type)`), then?

masterleinad: > I see what you mean, so my PR is very likely copying a wrong amount of data (since `sizeof(value_type*) == sizeof(pointer_type)`), then?

That's what it looks like to me.

pzehner: Do you have any idea on how to fix that?

masterleinad: Use `uintptr_t` for the `View`'s data type (and reinterpret_cast to `uintptr_t*` instead of `value_type*`).

containers/src/Kokkos_UnorderedMap.hpp
```diff
@@ -874,25 +874,13 @@ class UnorderedMap {
...
-        raw_deep_copy(m_values.data(), src.m_values.data(),
-                      sizeof(impl_value_type) * src.m_values.extent(0));
+        Kokkos::deep_copy(m_values, src.m_values);
       }
-      raw_deep_copy(m_scalars.data(), src.m_scalars.data(),
-                    sizeof(int) * num_scalars);
-
-      Kokkos::fence(
-          "Kokkos::UnorderedMap::deep_copy_view: fence after copy to dst.");
+      Kokkos::deep_copy(m_scalars, src.m_scalars);
```
masterleinad, line 883: Note that we now have up to 12 global fences where 2 would be sufficient.

containers/src/Kokkos_DynamicView.hpp
pzehner: There is no guarantee that `value_type` is an `int`, since it's template parameter. In the tests I see, it's a `double`.

masterleinad: `uintptr_t` does have the same size as a pointer. It doesn't matter what the type pointed to is. Also, round-trip conversion between a pointer and `uintptr_t` is fine and that's what we are effectively doing here. The `memcpy` never tries to interpret values stored in `uintptr_t`.

pzehner: Ok, I didn't know that. Seems like it works.

containers/src/Kokkos_UnorderedMap.hpp
pzehner: That's a lot indeed. Is it possible to call `deep_copy` without fencing?

masterleinad: You could at least use a three-argument deep_copy for all those calls except the first and the last.

masterleinad: > That's a lot indeed. Is it possible to call deep_copy without fencing?

That'd what the three-argument `deep_copy` is for.

containers/src/Kokkos_DynamicView.hpp
pzehner: Ew, the code segfaults now...

pzehner: Fixed.

containers/src/Kokkos_UnorderedMap.hpp
pzehner: Ok, I think I found a way.

masterleinad (approved):
Fine with me.

masterleinad (approved):
Still looks good to me.

containers/src/Kokkos_UnorderedMap.hpp
```diff
@@ -874,24 +874,18 @@ class UnorderedMap {
...
-        raw_deep_copy(m_values.data(), src.m_values.data(),
-                      sizeof(impl_value_type) * src.m_values.extent(0));
+        Kokkos::deep_copy(exec_space, m_values, src.m_values);
       }
-      raw_deep_copy(m_scalars.data(), src.m_scalars.data(),
-                    sizeof(int) * num_scalars);
+      Kokkos::deep_copy(exec_space, m_scalars, src.m_scalars);
 
-      Kokkos::fence(
+      exec_space.fence(
```
Rombur, line 888: Why did you change to `exec_space`? Currently we synchronize everything (`Kokkos::fence()` calls `static_fence()` which ends up calling `cudaDeviceSynchronize`). With this change,  if you use `--default-stream per-thread` with the CUDA backend, you will only fence the default streams not the others, right?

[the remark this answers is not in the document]
pzehner: I don't have a strong opinion about this

Rombur: `deep_copy_view` does not take an execution space, so I would expect to have a global fence. Unless there is a good reason, I think we should revert this change.

pzehner: Ok, I'll revert this.

Rombur (approved):

containers/src/Kokkos_Bitset.hpp
```diff
@@ -400,13 +400,7 @@ void deep_copy(Bitset<DstDevice>& dst, Bitset<SrcDevice> const& src) {
     Kokkos::Impl::throw_runtime_exception(
         "Error: Cannot deep_copy bitsets of different sizes!");
   }
```
crtrott, line 402: Should we remove these throws here? We have checks inside Kokkos::deep_copy for matching size. Though having it here would keep the more specific message.

crtrott (approved):
Approving but want to give other folks a chance to weigh in on the question of the throw in Bitset.

containers/src/Kokkos_Bitset.hpp
dalg24: Not feeling too strongly about it.  I would merge the PR as is.

crtrott: it would have been in scope since the check clearly accompanies the impl deep copy call same as the fences

PR #7517, opened 2024-11-06
Fix DynRankView in ScratchSpace

crtrott:
This fixes using DynRankView in scratch space. Two issues are fixed:

* `shmem_size` didn't report correct size.
* ctor taking scratch space called host only function.

PR #7523, opened 2024-11-08
Add PR 7517 to changelog

tcclevenger:


masterleinad (approved):

dalg24:
Relying to https://github.com/kokkos/kokkos/pull/7517#issuecomment-2463790170
> @dalg24 Should I add this to the changelog under fixes?

I am not sure.  My understanding is that this bug would not trigger before the view refactor changes that come in 4.5, in which case I would omit that entry.
@crtrott should be able to advise on this.

crtrott:
Should have worked before so we don't need this.

PR #7530, opened 2024-11-11
Fix some clang-tidy complaints

masterleinad:
Looking at https://github.com/kokkos/kokkos/pull/7430#discussion_r1800001727, I noticed that I am still seeing `clang-tidy` complaints locally. This pull request fixes those when compiling with `HPX`, `Threads`, `OpenMP` and `Serial`.

### DIFF
--- a/.github/workflows/continuous-integration-linux.yml
+++ b/.github/workflows/continuous-integration-linux.yml
@@ -19,6 +19,7 @@ jobs:
         cxx_extra_flags: ['']
         cmake_build_type: ['Release', 'Debug']
         backend: ['OPENMP']
+        clang-tidy: ['-DCMAKE_CXX_CLANG_TIDY="clang-tidy;-warnings-as-errors=*"']
         stdcxx: [17]
         include:
           - distro: 'ubuntu:intel'
@@ -26,12 +27,14 @@ jobs:
             cxx_extra_flags: '-diag-disable=177,1478,1786,10441'
             cmake_build_type: 'Release'
             backend: 'OPENMP'
+            clang-tidy: ''
             stdcxx: '17'
           - distro: 'ubuntu:intel'
             cxx: 'icpc'
             cxx_extra_flags: '-diag-disable=177,1478,1786,10441'
             cmake_build_type: 'Debug'
             backend: 'OPENMP'
+            clang-tidy: ''
             stdcxx: '17'
           - distro: 'ubuntu:intel'
             cxx: 'icpx'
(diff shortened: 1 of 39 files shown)


tpadioleau (approved):
LGTM!

tpadioleau:
Do you plan to continue working on it ? Running `grep -R "std::is_[a-Z_]*[^v]<"` I could catch more of them

dalg24:
Did you consider adding clang-tidy checks to a few more builds to increase "coverage"?

masterleinad:
> Do you plan to continue working on it ? Running `grep -R "std::is_[a-Z_]*[^v]<"` I could catch more of them

No, I don't intend to continue looking manually. Those are all instances I could find with local `clang-tidy` configurations.

masterleinad:
> Did you consider adding clang-tidy checks to a few more builds to increase "coverage"?

I'm not opposed to doing that in a follow-up but didn't intend to add more here.

dalg24:
I am not blocking but won't approve either if no build is added to enforce.

tpadioleau:
Do we understand why clang-tidy did not catch issues from the file `algorithms/src/sorting/Kokkos_BinOpsPublicAPI.hpp` before ? Isn't it used by the `Threads` backend used in the CI ?

masterleinad:
> Do we understand why clang-tidy did not catch issues from the file `algorithms/src/sorting/Kokkos_BinOpsPublicAPI.hpp` before ? Isn't it used by the `Threads` backend used in the CI ?

You can find the `clang-tidy` results running all builds in `continuous-integration-linux.yml` in https://github.com/masterleinad/kokkos/actions/runs/11937162091/job/33272485805.

.github/workflows/continuous-integration-linux.yml
```diff
@@ -19,7 +19,6 @@ jobs:
         cxx_extra_flags: ['']
         cmake_build_type: ['Release', 'Debug']
         backend: ['OPENMP']
-        clang-tidy: ['']
```
tpadioleau, line 22: Do you think we should exclude `g++` builds ? I don't know how reliable is the support compilers other than `clang++`.

masterleinad: It seems to work now after removing `-Wclobbered` and I think we reevaluate when problems appear.

masterleinad:
> Can I also add a test for HPX ?

Sure, go ahead.

.github/workflows/continuous-integration-linux.yml
masterleinad: `-Wclobbered` is included in `-Wextra` and also not really relevant for `Kokkos` AFAICT.

masterleinad: I don't have a strong opinion about disabling for `g++`. It's not clear to me if the build time is significantly longer and it makes writing the CI file more difficult.

cwpearson (approved):

PR #7537, opened 2024-11-19
core(graph): cleaning `is_graph_kernel`

romintomasetti:
This PR cleans `is_graph_kernel`.

Changes are twofold:
1. For specialized graph implementations, the aggregate node type is now much simpler, because it does not need to define a policy. This change also makes sense, because an aggregate node does not do any work, so it should not have a policy, nor should it be defined as a "graph kernel".
2. `GraphImpl::add_node` is now better constrained. Note that `add_node` is **not** part of the user-facing API, so we don't need to make a nice error message. Also note that I'm strengthening the constraints on this function because I'll introduce a new type of node in a followup PR, requiring `add_node` to be specialized.

### DIFF
--- a/core/src/Cuda/Kokkos_Cuda_Graph_Impl.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_Graph_Impl.hpp
@@ -121,10 +121,11 @@ struct GraphImpl<Kokkos::Cuda> {
   }
 
   template <class NodeImpl>
-  //  requires NodeImplPtr is a shared_ptr to specialization of GraphNodeImpl
-  //  Also requires that the kernel has the graph node tag in its policy
-  void add_node(std::shared_ptr<NodeImpl> const& arg_node_ptr) {
-    static_assert(NodeImpl::kernel_type::Policy::is_graph_kernel::value);
+  std::enable_if_t<
+      Kokkos::Impl::is_graph_kernel_v<typename NodeImpl::kernel_type>>
+  add_node(std::shared_ptr<NodeImpl> const& arg_node_ptr) {
+    static_assert(
+        Kokkos::Impl::is_specialization_of_v<NodeImpl, GraphNodeImpl>);
     KOKKOS_EXPECTS(bool(arg_node_ptr));
     // The Kernel launch from the execute() method has been shimmed to insert
     // the node into the graph
--- a/core/src/impl/Kokkos_Default_Graph_Impl.hpp
+++ b/core/src/impl/Kokkos_Default_Graph_Impl.hpp
@@ -80,9 +80,9 @@ struct GraphImpl : private ExecutionSpaceInstanceStorage<ExecutionSpace> {
   // <editor-fold desc="required customizations"> {{{2
 
   template <class NodeImpl>
-  //  requires NodeImplPtr is a shared_ptr to specialization of GraphNodeImpl
   void add_node(std::shared_ptr<NodeImpl> const& arg_node_ptr) {
-    static_assert(NodeImpl::kernel_type::Policy::is_graph_kernel::value);
+    static_assert(
+        Kokkos::Impl::is_specialization_of_v<NodeImpl, GraphNodeImpl>);
     // Since this is always called before any calls to add_predecessor involving
     // it, we can treat this node as a sink until we discover otherwise.
     arg_node_ptr->node_details_t::set_kernel(arg_node_ptr->get_kernel());
--- a/core/src/impl/Kokkos_Utilities.hpp
+++ b/core/src/impl/Kokkos_Utilities.hpp
@@ -120,6 +120,10 @@ struct is_specialization_of : std::false_type {};
 template <template <class...> class Template, class... Args>
 struct is_specialization_of<Template<Args...>, Template> : std::true_type {};
 
+template <typename T, template <typename...> class U>
+inline constexpr bool is_specialization_of_v =
+    is_specialization_of<T, U>::value;
+
 // </editor-fold> end is_specialization_of }}}1
 //==============================================================================
 
--- a/core/src/Cuda/Kokkos_Cuda_GraphNodeKernel.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_GraphNodeKernel.hpp
@@ -105,12 +105,6 @@ class GraphNodeKernelImpl<Kokkos::Cuda, PolicyType, Functor, PatternTag,
 
 struct CudaGraphNodeAggregateKernel {
   using graph_kernel = CudaGraphNodeAggregateKernel;
-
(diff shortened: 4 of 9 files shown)


masterleinad (dismissed):
```
/builds/kokkos/kokkos/core/src/SYCL/Kokkos_SYCL_Graph_Impl.hpp:111:26: error: out-of-line definition of 'add_node' does not match any declaration in 'Kokkos::Impl::GraphImpl<Kokkos::SYCL>'
  111 | GraphImpl<Kokkos::SYCL>::add_node(
      |                          ^~~~~~~~
```

core/src/impl/Kokkos_Utilities.hpp
```diff
@@ -120,6 +120,10 @@ struct is_specialization_of : std::false_type {};
 template <template <class...> class Template, class... Args>
 struct is_specialization_of<Template<Args...>, Template> : std::true_type {};
 
+template <typename T, template <typename...> class U>
+inline constexpr bool is_specialization_of_v =
+    is_specialization_of<T, U>::value;
+
```
JBludau, line 126: I think this should be a separate pr

core/src/impl/Kokkos_Default_Graph_Impl.hpp
```diff
@@ -80,9 +80,9 @@ struct GraphImpl : private ExecutionSpaceInstanceStorage<ExecutionSpace> {
   // <editor-fold desc="required customizations"> {{{2
 
   template <class NodeImpl>
-  //  requires NodeImplPtr is a shared_ptr to specialization of GraphNodeImpl
-  void add_node(std::shared_ptr<NodeImpl> const& arg_node_ptr) {
-    static_assert(NodeImpl::kernel_type::Policy::is_graph_kernel::value);
+  std::enable_if_t<
+      Kokkos::Impl::is_specialization_of_v<NodeImpl, GraphNodeImpl>>
+  add_node(std::shared_ptr<NodeImpl> const& arg_node_ptr) {
```
JBludau, line 85: hmm ... can we give an easy to read error message to our users? Do you really need SFINAE here?

romintomasetti: SFINAE is not 100% needed now, but a following PR will introduce a new type of node, for which we'll need to specialize `add_node`;

core/src/impl/Kokkos_Utilities.hpp
romintomasetti: Given that `is_specialization_of` is only used in the graph stuff, I would say that it's OK not to make a separate PR with just this change. Up to you.

JBludau: yeah, it is probably fine

core/src/impl/Kokkos_Default_Graph_Impl.hpp
JBludau: ah ok. well, if users never hit this I guess it is fine.

core/src/impl/Kokkos_Utilities.hpp
romintomasetti: Thanks, closing then.

core/src/impl/Kokkos_Default_Graph_Impl.hpp
romintomasetti: @JBludau i updated the PR description. If it's OK for you, please close this thread :wink:

romintomasetti:
@masterleinad @JBludau I don't know if we care about the Windows build failure. It seems unrelated to my changes, isn't ?

core/src/Cuda/Kokkos_Cuda_Graph_Impl.hpp
```diff
@@ -121,10 +121,10 @@ struct GraphImpl<Kokkos::Cuda> {
...
 
   template <class NodeImpl>
-  //  requires NodeImplPtr is a shared_ptr to specialization of GraphNodeImpl
-  //  Also requires that the kernel has the graph node tag in its policy
-  void add_node(std::shared_ptr<NodeImpl> const& arg_node_ptr) {
-    static_assert(NodeImpl::kernel_type::Policy::is_graph_kernel::value);
+  std::enable_if_t<
+      Kokkos::Impl::is_specialization_of_v<NodeImpl, GraphNodeImpl> &&
+      Kokkos::Impl::is_graph_kernel_v<typename NodeImpl::kernel_type>>
+  add_node(std::shared_ptr<NodeImpl> const& arg_node_ptr) {
```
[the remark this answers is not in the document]
masterleinad: Then, it seems a `static_assert` for `is_specialization_of_v<NodeImpl, GraphNodeImpl` instead using SFINAE for it makes sense.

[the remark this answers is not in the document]
romintomasetti: Indeed, and done :heavy_check_mark:

masterleinad (approved):
Fine with me.

core/src/Cuda/Kokkos_Cuda_Graph_Impl.hpp
```diff
@@ -121,10 +121,11 @@ struct GraphImpl<Kokkos::Cuda> {
   }
 
   template <class NodeImpl>
-  //  requires NodeImplPtr is a shared_ptr to specialization of GraphNodeImpl
-  //  Also requires that the kernel has the graph node tag in its policy
-  void add_node(std::shared_ptr<NodeImpl> const& arg_node_ptr) {
-    static_assert(NodeImpl::kernel_type::Policy::is_graph_kernel::value);
+  std::enable_if_t<
+      Kokkos::Impl::is_graph_kernel_v<typename NodeImpl::kernel_type>>
```
dalg24, line 125: I am confused.  Where is that `Impl::is_graph_kernel_v` helper variable template defined?

dalg24 (changes requested):

core/src/Cuda/Kokkos_Cuda_Graph_Impl.hpp
romintomasetti: Not easy to find indeed.

https://github.com/kokkos/kokkos/blob/bc6581798c5efe1abc8b1b937baa7f2185414665/core/src/Kokkos_Concepts.hpp#L168

dalg24 (approved):
Summarizing the changes here:

* Turn the `GraphImpl::add_node()` **mandate** (i.e. hard requirement on type enforced via a static assertion) that the node pointer argument is tagged as a graph kernel into a **constraint** (i.e. control on overload resolution) in preparation of adding a new overload that takes some other kind of graph node
* Get rid of the dummy `Policy` "tag" on aggregate nodes
* Add `Impl::is_specialization_v` helper variable template

PR #7539, opened 2024-11-21
SYCL: Add clang-tidy check

masterleinad:
Follow-up to #7530.

### DIFF
--- a/core/src/Cuda/Kokkos_Cuda_Half_Conversion.hpp
+++ b/core/src/Cuda/Kokkos_Cuda_Half_Conversion.hpp
@@ -97,25 +97,25 @@ half_t cast_to_half(unsigned long val) {
 }
 
 template <class T>
-KOKKOS_INLINE_FUNCTION std::enable_if_t<std::is_same<T, float>::value, T>
+KOKKOS_INLINE_FUNCTION std::enable_if_t<std::is_same_v<T, float>, T>
 cast_from_half(half_t val) {
   return __half2float(half_t::impl_type(val));
 }
 
 template <class T>
-KOKKOS_INLINE_FUNCTION std::enable_if_t<std::is_same<T, bool>::value, T>
+KOKKOS_INLINE_FUNCTION std::enable_if_t<std::is_same_v<T, bool>, T>
 cast_from_half(half_t val) {
   return static_cast<T>(cast_from_half<float>(val));
 }
 
 template <class T>
-KOKKOS_INLINE_FUNCTION std::enable_if_t<std::is_same<T, double>::value, T>
+KOKKOS_INLINE_FUNCTION std::enable_if_t<std::is_same_v<T, double>, T>
 cast_from_half(half_t val) {
   return static_cast<T>(__half2float(half_t::impl_type(val)));
 }
 
 template <class T>
-KOKKOS_INLINE_FUNCTION std::enable_if_t<std::is_same<T, short>::value, T>
+KOKKOS_INLINE_FUNCTION std::enable_if_t<std::is_same_v<T, short>, T>
 cast_from_half(half_t val) {
   KOKKOS_IF_ON_DEVICE((return __half2short_rz(half_t::impl_type(val));))
   KOKKOS_IF_ON_HOST(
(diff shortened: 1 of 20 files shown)


core/unit_test/sycl/TestSYCL_InterOp_Graph.cpp
```diff
@@ -14,16 +14,20 @@
...
 
 #include <gtest/gtest.h>
 
+#include <fstream>
+#include <regex>
+
+#if !(defined(_GLIBCXX_RELEASE) && _GLIBCXX_RELEASE < 9) && \
+    !(defined(_LIBCPP_VERSION) && _LIBCPP_VERSION < 110000)
+#include <filesystem>
+#endif
```
masterleinad, line 29: Trying to include `filesystem` caused problems with the underlying `clang` version used for `clang-tidy`.

dalg24: Ok

dalg24 (approved):
I do not love the change to the Cuda half type header but I don't think it is worth re-running the tests

core/src/Cuda/Kokkos_Cuda_Half_Conversion.hpp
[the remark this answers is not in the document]
masterleinad: I needed that to make the CI happy.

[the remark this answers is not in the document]
dalg24: How so?  Was it something about the signature of the definition not matching the declaration?

core/src/impl/KokkosExp_IterateTileGPU.hpp
```diff
@@ -875,7 +875,7 @@ struct is_array_type<T[]> : std::true_type {
 template <typename T>
 using value_type_storage_t =
     typename std::conditional_t<is_array_type<T>::value, std::decay<T>,
-                                std::add_lvalue_reference<T> >::type;
+                                std::add_lvalue_reference<T>>::type;
```
tpadioleau, line 878: What do you think of the following ?

```cpp
template <typename T>
using value_type_storage_t =
    std::conditional_t<is_array_type<T>::value, std::decay_t<T>,
                                std::add_lvalue_reference_t<T>>;
```

core/unit_test/sycl/TestSYCL_InterOp_Graph.cpp
tpadioleau: I am surprised we are using an old clang here, doesn't it come from the oneapi install ? I would expect a recent clang/clang-tidy ?

core/src/impl/Kokkos_Half_FloatingPointWrapper.hpp
tpadioleau, line 1: How does half type works ? This code was disabled in the other builds ?

tpadioleau (commented):
Looks better :)

core/unit_test/sycl/TestSYCL_InterOp_Graph.cpp
masterleinad: I could still use an older stdlibc++. I haven't investigated further, though.

core/src/impl/Kokkos_Half_FloatingPointWrapper.hpp
masterleinad: For most configurations the half types are just aliases to `float`. Only `Cuda`, `HIP` and `SYCL` use special types.

core/src/impl/KokkosExp_IterateTileGPU.hpp
masterleinad: Sure, see https://github.com/kokkos/kokkos/pull/7539/commits/2980f51adb93388e4feb02567273b884b42ce320.

Rombur (approved):

tpadioleau (approved):

core/unit_test/sycl/TestSYCL_InterOp_Graph.cpp
tpadioleau: I didn't see that the file had already such `ifdef` logic in the tests

masterleinad:
I had to fix `HIP` in the same way as `Cuda`.

dalg24:
Retest this please

Issue #7526, opened 2024-11-11
Cmake .in files are not formatted with `cmake-format`


JBludau:
should be done after open PRs e.g. #7525 that touch the files are merged

pzehner:
This is a noob comment, but

> - Putting all kokkos dependent files of a library into a subdirectory (if you do not expose Kokkos objects in your own interface), and linking with `target_link_libraries(myLib PRIVATE Kokkos::kokkos)`.

subdirectory part aside, isn't this what should be done normally anyway?

JBludau:
> This is a noob comment, but
> 
> > * Putting all kokkos dependent files of a library into a subdirectory (if you do not expose Kokkos objects in your own interface), and linking with `target_link_libraries(myLib PRIVATE Kokkos::kokkos)`.
> 
> subdirectory part aside, isn't this what should be done normally anyway?

yes, you are right. The important thing is the "expose Kokkos objects in your interface". Meaning, you can not have any function that accepts e.g. a `Kokkos::View` in its function interface. So if you do not want to write your own wrapper for some data storage like `View` you can not link `PRIVATE` to `Kokkos::kokkos`. In this case everyone that uses your library will need to compile their code with the compiler and flags that we set (which might not work for them). Now the question is "how do we allow users to only compile files that actually use/expose kokkos" with the compiler we set vs. setting the compiler globally (like we do now). Does this make sense?

cedricchevalier19:
> yes, you are right. The important thing is the "expose Kokkos objects in your interface". Meaning, you can not have any function that accepts e.g. a `Kokkos::View` in its function interface. So if you do not want to write your own wrapper for some data storage like `View` you can not link `PRIVATE` to `Kokkos::kokkos`.

I thought it was the CMake definition `PRIVATE` to not expose anything from underneath. In your case you have also to be `INTERFACE` and CMake rules that `PRIVATE+INTERFACE = PUBLIC`.

JBludau:
> > yes, you are right. The important thing is the "expose Kokkos objects in your interface". Meaning, you can not have any function that accepts e.g. a `Kokkos::View` in its function interface. So if you do not want to write your own wrapper for some data storage like `View` you can not link `PRIVATE` to `Kokkos::kokkos`.
> 
> I thought it was the CMake definition `PRIVATE` to not expose anything from underneath. In your case you have also to be `INTERFACE` and CMake rules that `PRIVATE+INTERFACE = PUBLIC`.

[...]

PR #7541, opened 2024-11-21
Make `build_cmake_installed_kk_as_language` example backend agnostic

JBludau:
I tried to use the vars we export to make it agnostic (at least for `CUDA` and `HIP`).
We also should specify the std and architectures Kokkos was configured for on the target. Otherwise we end up with the std one depending on the toolkit which might not provide stuff we use inside kokkos

### DIFF
--- a/example/build_cmake_installed_kk_as_language/CMakeLists.txt
+++ b/example/build_cmake_installed_kk_as_language/CMakeLists.txt
@@ -7,7 +7,7 @@ cmake_minimum_required(VERSION 3.19)
 project(Example CXX Fortran)
 
 # Look for an installed Kokkos
-find_package(Kokkos COMPONENTS)
+find_package(Kokkos)
 enable_language(${Kokkos_COMPILE_LANGUAGE})
 
 #the language has to be set on the files
--- a/example/build_cmake_installed_kk_as_language/CMakeLists.txt
+++ b/example/build_cmake_installed_kk_as_language/CMakeLists.txt
@@ -4,14 +4,19 @@ cmake_minimum_required(VERSION 3.19)
 
 # Projects can safely mix languages - must have C++ support
 # Kokkos flags will only apply to C++ files
-project(Example CXX Fortran CUDA)
+project(Example CXX Fortran)
 
 # Look for an installed Kokkos
-find_package(Kokkos COMPONENTS separable_compilation)
-set_source_files_properties(cmake_example.cpp PROPERTIES LANGUAGE CUDA)
+find_package(Kokkos COMPONENTS)
+enable_language(${Kokkos_COMPILE_LANGUAGE})
+
+#the language has to be set on the files
+set_source_files_properties(cmake_example.cpp PROPERTIES LANGUAGE ${Kokkos_COMPILE_LANGUAGE})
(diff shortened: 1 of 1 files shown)


example/build_cmake_installed_kk_as_language/CMakeLists.txt
```diff
@@ -4,14 +4,21 @@ cmake_minimum_required(VERSION 3.19)
 
 # Projects can safely mix languages - must have C++ support
 # Kokkos flags will only apply to C++ files
-project(Example CXX Fortran CUDA)
+project(Example CXX Fortran)
 
 # Look for an installed Kokkos
-find_package(Kokkos COMPONENTS separable_compilation)
-set_source_files_properties(cmake_example.cpp PROPERTIES LANGUAGE CUDA)
+find_package(Kokkos COMPONENTS)
```
masterleinad, line 10:
```suggestion
find_package(Kokkos)
```

masterleinad (commented):
What's the relation to #7525?

example/build_cmake_installed_kk_as_language/CMakeLists.txt
JBludau: whoops

JBludau:
> What's the relation to #7525?

I am holding back on this one because I think we have to discuss some strategy first. I tried to give a quick outline in #7526

masterleinad (approved):

example/build_cmake_installed_kk_as_language/CMakeLists.txt
```diff
@@ -4,14 +4,21 @@ cmake_minimum_required(VERSION 3.19)
...
+enable_language(${Kokkos_COMPILE_LANGUAGE})
+
+#the language has to be set on the files
+set_source_files_properties(cmake_example.cpp PROPERTIES LANGUAGE ${Kokkos_COMPILE_LANGUAGE})
 add_executable(example cmake_example.cpp bar.cpp foo.f)
 
-# This is the only thing required to set up compiler/linker flags
+# The architectures and std have to be set on the target
+set_target_properties(
+  example PROPERTIES ${Kokkos_COMPILE_LANGUAGE}_ARCHITECTURES ${Kokkos_${Kokkos_COMPILE_LANGUAGE}_ARCHITECTURES}
```
dalg24, line 19: What happens with cmake when you set a property that does not exists?
As I understand it, there is no such thing as a CXX_ARCHITECTURES property and I am somewhat surprised it does not error out that the value argument is missing (trailing arg does not exists and presumably does not exists)

masterleinad: It does
```
CMake Error at CMakeLists.txt:18 (set_target_properties):
  set_target_properties called with incorrect number of arguments.
```
In my understanding this example is only meant to be used (at this point) when the `HIP` or `Cuda` backend is enabled.

JBludau: hmm `set_properties` just silently does nothing for CXX as language ... but not sure if we want to it this way

crtrott (approved):

crtrott:
Retest this please.

dalg24:
Retest this please

PR #7542, opened 2024-11-22
call thrust::inclusive_scan when available

science-enthusiast:
Update 6/1/25: The PR now takes care of AMD GPUs also. Thus, in total 4 functions have been added, 2 for CUDA and 2 for HIP. For both CUDA and HIP, two overloads of Kokkos::Experimental::inclusive_scan have been implemented which call the corresponding thrust::inclusive_scan functions, under device execution space.

Also, a benchmark code has been implemented inside the `algorithms/perf_test`. Note that the directory `algorithms/perf_test` has been newly created. Going forward perf tests regarding algorithms are expected to be added here, just like `core/perf_test`.

Rough performance numbers (based on just a single run of the scan operation) can be observed by using the following two test codes:

Thrust: [thrust_single_test_cpp.txt](https://github.com/user-attachments/files/18504552/thrust_single_test_cpp.txt)

Kokkos: [kokkos_single_test_cpp.txt](https://github.com/user-attachments/files/18504562/kokkos_single_test_cpp.txt)

Total number of elements in the input = 1e8. The input contains values 0, 1, 2, ..., 1e8 - 1. Scans based on sum and max binary operations were tested.

thrust::inclusive_scan 
sum (default): 2467 us, sum (functor): 5460 us, max (functor): 6165 us

With modified Kokkos (calling thrust::inclusive_scan)
sum (default): 3829 us, sum (functor): 4667 us, max (functor): 2683 us

With original Kokkos (calling Kokkos::parallel_scan)
sum (default): 11488 us, sum (functor): 9618 us, max (functor): 9743 us

As a point of reference, Kokkos::parallel_scan with a summation based functor takes 12998 us based on the following code: 
[test-kokkos_cpp.txt](https://github.com/user-attachments/files/17870598/test-kokkos_cpp.txt)


masterleinad:
You'll have to fix the formatting:
```diff
--- ./algorithms/src/std_algorithms/impl/Kokkos_InclusiveScan.hpp	(original)
+++ ./algorithms/src/std_algorithms/impl/Kokkos_InclusiveScan.hpp	(reformatted)
@@ -149,9 +149,8 @@
 #if defined(KOKKOS_ENABLE_CUDA)
 template <class InputIteratorType, class OutputIteratorType>
 OutputIteratorType inclusive_scan_default_op_exespace_impl(
-    const std::string& label, const Cuda& ex,
-    InputIteratorType first_from, InputIteratorType last_from,
-    OutputIteratorType first_dest) {
+    const std::string& label, const Cuda& ex, InputIteratorType first_from,
+    InputIteratorType last_from, OutputIteratorType first_dest) {
   const auto thrust_ex = thrust::cuda::par.on(ex.cuda_stream());
 
   thrust::inclusive_scan(thrust_ex, first_from, last_from, first_dest);
@@ -204,17 +203,16 @@
 // -------------------------------------------------------------
 
 #if defined(KOKKOS_ENABLE_CUDA)
-template <class InputIteratorType, class OutputIteratorType,
-	  class BinaryOpType>
+template <class InputIteratorType, class OutputIteratorType, class BinaryOpType>
 OutputIteratorType inclusive_scan_custom_binary_op_exespace_impl(
-    const std::string& label, const Cuda& ex,
-    InputIteratorType first_from, InputIteratorType last_from,
-    OutputIteratorType first_dest, BinaryOpType binary_op) {
+    const std::string& label, const Cuda& ex, InputIteratorType first_from,
+    InputIteratorType last_from, OutputIteratorType first_dest,
+    BinaryOpType binary_op) {
   const auto thrust_ex = thrust::cuda::par.on(ex.cuda_stream());
 
-  thrust::inclusive_scan(thrust_ex, first_from, last_from,
-		         first_dest, binary_op);
- 
+  thrust::inclusive_scan(thrust_ex, first_from, last_from, first_dest,
+                         binary_op);
+
   const auto num_elements = thrust::distance(first_from, last_from);
 
   return first_dest + num_elements;
```

> thrust::inclusive_scan - sum (functor) Time taken: 5460 us
> thrust::inclusive_scan - max (functor) Time taken: 6165 us
>
> With modified Kokkos (calling thrust::inclusive_scan):
>
> KE::inclusive_scan - sum (functor) - time taken: 4667 us
> KE::inclusive_scan - max (functor) - time taken: 2683 us

How do you explain that you are observing `Kokkos` to be faster if it's calling `thrust::inclusive_scan`?

science-enthusiast:
@masterleinad Yes, I am working on the formatting errors.

Regarding the performance numbers. The only reason that I can think of is that these numbers are based on a single run of the scan in each of the cases. I could use Google Benchmark and report those numbers.

masterleinad:
> I could use Google Benchmark and report those numbers.

Or just run the benchmark 100 times or so manually.

algorithms/src/std_algorithms/impl/Kokkos_InclusiveScan.hpp
```diff
@@ -24,6 +24,47 @@
...
+#pragma GCC diagnostic ignored "-Wsuggest-override"
+
+#if defined(KOKKOS_COMPILER_CLANG)
+// Some versions of Clang fail to compile Thrust, failing with errors like
+// this:
+//    <snip>/thrust/system/cuda/detail/core/agent_launcher.h:557:11:
+//    error: use of undeclared identifier 'va_printf'
+// The exact combination of versions for Clang and Thrust (or CUDA) for this
+// failure was not investigated, however even very recent version combination
+// (Clang 10.0.0 and Cuda 10.0) demonstrated failure.
```
cedricchevalier19, line 47: Just curious, but is this error still present with the latest versions of clang or cuda?

science-enthusiast: Good point. I will check.

algorithms/src/std_algorithms/impl/Kokkos_InclusiveScan.hpp
```diff
@@ -101,9 +142,25 @@ struct InclusiveScanDefaultFunctor {
...
-// exespace impl
-//
+// -------------------------------------------------------------
+// inclusive_scan_default_op_exespace_impl
+// -------------------------------------------------------------
+
+#if defined(KOKKOS_ENABLE_CUDA)
+template <class InputIteratorType, class OutputIteratorType>
+OutputIteratorType inclusive_scan_default_op_exespace_impl(
+    const std::string& label, const Cuda& ex, InputIteratorType first_from,
```
masterleinad, line 152: The label is unused here. Maybe, add something like
```C++
 uint64_t kpID = 0;
      Kokkos::Profiling::beginParallelScan(
          label + " via thrust::inclusive_scan",
          Kokkos::Profiling::Experimental::device_id(ex), &kpID);
    }
```
before the call to `thrust::inclusize_scan` and
```C++
      Kokkos::Profiling::endParallelScan(kpID);
    }
```
after it. Similarly below.

aprokop (commented):
Why only CUDA? What about rocThrust? What's the difference there?

algorithms/src/std_algorithms/impl/Kokkos_InclusiveScan.hpp
```diff
@@ -24,6 +24,47 @@
...
+#define _CubLog
+#include <thrust/distance.h>
+#include <thrust/scan.h>
+#pragma pop_macro("_CubLog")
+#else
+#include <thrust/distance.h>
+#include <thrust/scan.h>
+#endif
+
+#pragma GCC diagnostic pop
```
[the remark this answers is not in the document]
science-enthusiast: Yes, it will be good to run tests on AMD GPUs. I will come back on this.

This code snippet was copied from the sort implementation. Yes, refactoring this part into a separate header is the better approach. Also, as @cedricchevalier19 pointed out, I will check whether the problem still exists with recent Clang and Cuda.

tpadioleau (commented):
I have a few remarks/questions

1. Can you share the latest version of your test codes ? You recently added repetition loops right ?

2. Can you benchmark again the Thrust implementation using `thrust::device_vector<size_t>` for the input array ?

3. Can you use the same functors for `SumFunctor` and `MaxFunctor` both in the Thrust and Kokkos implementations ?

science-enthusiast:
@tpadioleau 

Here are the codes as .txt files. Don't mind the WET-ness in the code.
[test_in_scan_avg_cpp.txt](https://github.com/user-attachments/files/17920904/test_in_scan_avg_cpp.txt)
[test_thrust_avg_cu.txt](https://github.com/user-attachments/files/17920905/test_thrust_avg_cu.txt)

Sure, I will get back to you with the benchmarks for the case that directly calls thrust::inclusive_scan:
- by creating an input device_vector (and not using the counting iterators).
- with functors that accept parameters by const reference. const ref parameters are needed by Kokkos. Otherwise, matching instances of functions are not being found.

dalg24:
Please briefly discuss in the description the status of exclusive_scan

science-enthusiast:
Results generated after ensuring that in all cases:
- the input is a device_vector of ints.
- the binary operator functor takes its parameters by const reference.

These results have to be compared with those given [here](https://github.com/kokkos/kokkos/pull/7542#issuecomment-2494521173). As pointed out by @tpadioleau [here](https://github.com/kokkos/kokkos/pull/7542#pullrequestreview-2461629886), in that case, the code directly calling thrust::inclusive_scan took the inputs based on counting_iterators (and not device vectors) and the parameters in the binary operator functor were passed by value and not const reference.

Average time taken over 100 runs

Kokkos::Experimental::inclusive_scan calling Kokkos::parallel_scan in the backend
KE::inclusive_scan - sum (default) - avg. time: 4993 us
KE::inclusive_scan - sum (functor) - avg. time: 9392 us
KE::inclusive_scan - max (functor) - avg. time: 9390 us

Kokkos::Experimental::inclusive_scan calling thrust::inclusive_scan in the backend
KE::inclusive_scan - sum (default) - avg. time: 3359 us
KE::inclusive_scan - sum (functor) - avg. time: 4518 us
KE::inclusive_scan - max (functor) - avg. time: 3395 us

thrust::inclusive_scan
thrust::inclusive_scan - sum (default) - avg. time: 3240 us
thrust::inclusive_scan - sum (functor) - avg. time: 2327 us
thrust::inclusive_scan - max (functor) - avg. time: 2789 us

Kokkos::parallel_scan - sum functor - avg. time: 3396 us

Observation:
- Calling thrust::inclusive_scan directy is generally more efficient.
- thrust::inclusive_scan taking the input as a device_vector is slower compared to counting_iterators.

The codes used:
[test-thrust-avg_cu.txt](https://github.com/user-attachments/files/18024266/test-thrust-avg_cu.txt)
[test-in-scan-avg_cpp.txt](https://github.com/user-attachments/files/18024267/test-in-scan-avg_cpp.txt)

science-enthusiast:
@dalg24
A separate PR would be needed to implement a set of perf tests (similar to the tests in `core/perf_tests`) based on Google Benchmark.
Would it be ok, to merge the current PR and then work on the PR for the perf test or should the perf test be ready first?

masterleinad:
```
Kokkos::Experimental::inclusive_scan calling thrust::inclusive_scan in the backend
KE::inclusive_scan - sum (default) - avg. time: 3359 us
KE::inclusive_scan - sum (functor) - avg. time: 4518 us
KE::inclusive_scan - max (functor) - avg. time: 3395 us

thrust::inclusive_scan
thrust::inclusive_scan - sum (default) - avg. time: 3240 us
thrust::inclusive_scan - sum (functor) - avg. time: 2327 us
thrust::inclusive_scan - max (functor) - avg. time: 2789 us
```
Can you explain these discrepancies, particularly in the second case? What do you see running NSight?

science-enthusiast:
> Can you explain these discrepancies, particularly in the second case? What do you see running NSight?

@masterleinad, I am also surprised by the second case (sum scan using a functor). I have rerun the test and get similar numbers. I haven't analysed with NSight. I will come back on this.

cedricchevalier19:
Can you have more data than just the average (like min, max, std)? Experiments are long enough to be meaningful even if run one by one.

algorithms/src/std_algorithms/impl/Kokkos_InclusiveScan.hpp
science-enthusiast: An issue has been created (#7649) to track this.

algorithms/src/std_algorithms/impl/Kokkos_InclusiveScan.hpp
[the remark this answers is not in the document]
science-enthusiast: 2 functions have been added to called thrust::inclusive_scan when using HIP backend.

science-enthusiast:
> Why only CUDA? What about rocThrust? What's the difference there?

2 functions have been added to called thrust::inclusive_scan when using HIP backend.

algorithms/src/std_algorithms/impl/Kokkos_InclusiveScan.hpp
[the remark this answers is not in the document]
science-enthusiast: An issue has been created (#7649) to keep track of this problem. It is expected to be resolved by release 5.0.

algorithms/perf_test/CMakeLists.txt
```diff
@@ -0,0 +1,65 @@
...
+if((KOKKOS_ENABLE_OPENMPTARGET OR KOKKOS_ENABLE_OPENACC) AND KOKKOS_CXX_COMPILER_ID STREQUAL NVHPC)
+  return()
+endif()
+if(KOKKOS_ENABLE_OPENACC AND KOKKOS_CXX_COMPILER_ID STREQUAL Clang)
+  return()
+endif()
+
+if(NOT Kokkos_ENABLE_BENCHMARKS)
+  return()
+endif()
```
masterleinad, line 12: You are already checking for this in `algorithms/CMakeLists.txt`.
```suggestion
```

algorithms/perf_test/CMakeLists.txt
```diff
@@ -0,0 +1,65 @@
+# FIXME_OPENMPTARGET - the NVIDIA HPC compiler nvc++ in the OpenMPTarget backend does not pass the perf_tests.
+# FIXME_OPENACC - temporarily disabled due to unimplemented features
+if((KOKKOS_ENABLE_OPENMPTARGET OR KOKKOS_ENABLE_OPENACC) AND KOKKOS_CXX_COMPILER_ID STREQUAL NVHPC)
+  return()
+endif()
+if(KOKKOS_ENABLE_OPENACC AND KOKKOS_CXX_COMPILER_ID STREQUAL Clang)
+  return()
+endif()
```
masterleinad, line 8: It looks like you just copied this block from `core/perf_test/CMakeLists.txt`. Have you actually checked if it's necessary here?

algorithms/perf_test/CMakeLists.txt
```diff
@@ -0,0 +1,65 @@
...
+
+  foreach(SOURCE_FILE ${BENCHMARK_SOURCES})
+    set_source_files_properties(${SOURCE_FILE} PROPERTIES LANGUAGE ${KOKKOS_COMPILE_LANGUAGE})
+  endforeach()
+
+  string(TIMESTAMP BENCHMARK_TIME "%Y-%m-%d_T%H-%M-%S" UTC)
+  set(BENCHMARK_ARGS --benchmark_counters_tabular=true --benchmark_out=${BENCHMARK_NAME}_${BENCHMARK_TIME}.json)
+
+  add_test(NAME ${BENCHMARK_NAME} COMMAND ${BENCHMARK_NAME} ${BENCHMARK_ARGS})
+endfunction()
```
masterleinad, line 63: I'd prefer not to duplicate all of this but share it in a common location with `core/perf_test/CMakeLists.txt`

algorithms/perf_test/test_inclusive_scan.cpp
```diff
@@ -0,0 +1,143 @@
...
+  }
+};
+
+// Helper to obtain last element of a view
+template <class T>
+T obtainLastElem(const Kokkos::View<T*, exec_space>& out) {
+  auto h_out = Kokkos::create_mirror_view(out);
+  Kokkos::deep_copy(h_out, out);
+  auto view_size = h_out.extent(0);
+  return h_out(view_size - 1);
```
masterleinad, line 62: You don't have to copy the whole `View` if you only want the last element.
```suggestion
  T last_element;
  Kokkos::deep_copy(last_element, Kokkos::subview(out, out.extent(0)-1);
  return last_element;
```

algorithms/perf_test/test_inclusive_scan.cpp
```diff
@@ -0,0 +1,143 @@
...
+}
+
+template <class T>
+void InclusiveScan(benchmark::State& state,
+                   const Kokkos::View<T*, exec_space>& in,
+                   Kokkos::View<T*, exec_space>& out) {
+  // Default prefix sum scan
+  Kokkos::Timer timer;
+  KE::inclusive_scan("default scan", exec_space(), KE::cbegin(in), KE::cend(in),
```
masterleinad, line 71: Maybe fence before for good measure to make sure nothing else is included in the time recorded.
```suggestion
  exec_space().fence();
  // Default prefix sum scan
  Kokkos::Timer timer;
  KE::inclusive_scan("default scan", exec_space(), KE::cbegin(in), KE::cend(in),
```

algorithms/perf_test/test_inclusive_scan.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+  T res_def = obtainLastElem(out);
+
+  // Prefix sum scan using a sum functor
+  SumFunctor<T> sum_functor;
+
+  timer.reset();
+  KE::inclusive_scan("sum functor scan", exec_space(), KE::cbegin(in),
+                     KE::cend(in), KE::begin(out), sum_functor);
+  exec_space().fence();
```
masterleinad, line 84: Without an explicit fence right before `timer.reset()`, this relies on `obtainLastElem` to fence (which it currently does).

algorithms/perf_test/test_inclusive_scan.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+  T res_sum_func = obtainLastElem(out);
+
+  // Scan using a max functor
+  MaxFunctor<T> max_functor;
+
+  timer.reset();
+  KE::inclusive_scan("max functor scan", exec_space(), KE::cbegin(in),
+                     KE::cend(in), KE::begin(out), max_functor);
+  exec_space().fence();
```
masterleinad, line 95: Same here.

algorithms/perf_test/test_inclusive_scan.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+  // Checking expected results
+  std::size_t res_max_func = static_cast<std::size_t>(obtainLastElem(out));
+  std::size_t exp_max_val  = in.size() - 1;
+  bool passed = (res_def == res_sum_func) && (res_max_func == exp_max_val);
+
+  state.counters["Passed"]             = benchmark::Counter(passed);
+  state.counters["Value default scan"] = benchmark::Counter(res_def);
+  state.counters["Value sum functor"]  = benchmark::Counter(res_sum_func);
+  state.counters["Value max functor"]  = benchmark::Counter(res_max_func);
```
masterleinad, line 106: This feels out-of-place. I would rather error out if the results aren't correct.

algorithms/perf_test/test_inclusive_scan.cpp
```diff
@@ -0,0 +1,143 @@
...
+  state.counters["Time default scan"]  = benchmark::Counter(time_def);
+  state.counters["Time sum functor"]   = benchmark::Counter(time_sum_func);
+  state.counters["Time max functor"]   = benchmark::Counter(time_max_func);
+  state.counters["Size of type"]       = benchmark::Counter(sizeof(T));
```
masterleinad, line 110: Why are you printing the size of the type as a counter?
```suggestion
```

algorithms/perf_test/test_inclusive_scan.cpp
```diff
@@ -0,0 +1,143 @@
...
+//
+//@HEADER
+
+#include <algorithm>
+#include <cstddef>
+#include <cstdio>
+#include <cstdint>
+#include <cstring>
+#include <cstdlib>
+#include <random>
```
masterleinad, line 23: Where is this header used? Do you need all the others above?

algorithms/perf_test/test_inclusive_scan.cpp
dalg24, line 110: Why did you pick this approach instead of having individual benchmarks?

algorithms/perf_test/test_inclusive_scan.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+#include "../../core/perf_test/Benchmark_Context.hpp"
+
+namespace KE = Kokkos::Experimental;
+
+using exec_space = Kokkos::DefaultExecutionSpace;
+
+template <class ValueType>
+struct SumFunctor {
+  KOKKOS_INLINE_FUNCTION
```
dalg24, line 39:
```suggestion
  KOKKOS_FUNCTION
```
Same comment below

masterleinad:
Do you have some more performance results for `NVIDIA` and `AMD` GPUs?

algorithms/perf_test/CMakeLists.txt
science-enthusiast: That is true. I could check by removing these conditions (one set at a time) and seeing whether it breaks the build.

algorithms/perf_test/test_inclusive_scan.cpp
science-enthusiast: Yes, I was wondering about this. I could have asked people around, instead of committing it in a PR. Thank you.

algorithms/perf_test/CMakeLists.txt
science-enthusiast: I will look into this.

algorithms/perf_test/test_inclusive_scan.cpp
science-enthusiast: I thought, the input and output views could be allocated only once and reused.

algorithms/perf_test/test_inclusive_scan.cpp
```diff
@@ -0,0 +1,143 @@
...
+};
+
+// Helper to obtain last element of a view
+template <class T>
+T obtainLastElem(const Kokkos::View<T*, exec_space>& out) {
+  auto h_out = Kokkos::create_mirror_view(out);
+  Kokkos::deep_copy(h_out, out);
+  auto view_size = h_out.extent(0);
+  return h_out(view_size - 1);
+}
```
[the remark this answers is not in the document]
science-enthusiast: I was wondering about this. I search for `back()`-like function and didn't find it. I should have thought of using subview.

algorithms/perf_test/test_inclusive_scan.cpp
science-enthusiast: While looking at the results, I thought this will give an idea about the size of the array involved. However, this information can be easily inferred.

algorithms/perf_test/test_inclusive_scan.cpp
```diff
@@ -0,0 +1,143 @@
...
+    InclusiveScan<T>(state, in, out);
+  }
+}
+
+static constexpr std::size_t PROB_SIZE  = 100'000'000;
+static constexpr std::uint32_t ITER_CNT = 100;
+
+BENCHMARK(Test_InclusiveScan<std::uint64_t>)
+    ->Arg(PROB_SIZE)
+    ->Iterations(ITER_CNT);
```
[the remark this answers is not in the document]
science-enthusiast: I was doing this to check whether I got similar numbers to my earlier tests which manually ran the code for 100 iterations. Then I just left it like that, without removing it. It may be a better practice to let Google Benchmark select the number of iterations.

algorithms/perf_test/test_inclusive_scan.cpp
science-enthusiast: I was thinking that the fence that just follows each of the calls to inclusive_scan was enough. To be sure, I will fence before the call to inclusive_scan also.

algorithms/perf_test/test_inclusive_scan.cpp
science-enthusiast: `<cstdint>` is for `std::uint64_t`. I wanted to implement a logic to create an input array that will have random values picked randomly between 0 and N-1. But then I didn't implement it. Even then, `std::shuffle` may be better than using functions in `<random>`. I think I used `<cstring>` while debugging but I am not sure. I don't remember why `<cstdlib>` was included, since, `<cstddef>` (which is also included) is the more reliable header for the definition of `std::size_t`. In short, among the four listed above, I will retain `<cstdint>`.

algorithms/perf_test/CMakeLists.txt
tpadioleau: In the `cmake` directory ?

tpadioleau (changes requested):

algorithms/perf_test/test_inclusive_scan.cpp
```diff
@@ -0,0 +1,143 @@
...
+  double time_def = timer.seconds();
+
+  T res_def = obtainLastElem(out);
+
+  // Prefix sum scan using a sum functor
+  SumFunctor<T> sum_functor;
+
+  timer.reset();
+  KE::inclusive_scan("sum functor scan", exec_space(), KE::cbegin(in),
+                     KE::cend(in), KE::begin(out), sum_functor);
```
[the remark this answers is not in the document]
science-enthusiast: Yes, this will also work.

algorithms/perf_test/test_inclusive_scan.cpp
```diff
@@ -0,0 +1,143 @@
...
+#include <cstdlib>
+#include <random>
+
+#include <benchmark/benchmark.h>
+
+#include <Kokkos_Core.hpp>
+#include <Kokkos_Timer.hpp>
+#include <Kokkos_StdAlgorithms.hpp>
+
+#include "../../core/perf_test/Benchmark_Context.hpp"
```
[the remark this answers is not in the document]
science-enthusiast: I am already reviewing the headers included in the code. Let me see which all need to be retained.

algorithms/perf_test/test_inclusive_scan.cpp
```diff
@@ -0,0 +1,143 @@
...
+  auto h_out = Kokkos::create_mirror_view(out);
+  Kokkos::deep_copy(h_out, out);
+  auto view_size = h_out.extent(0);
+  return h_out(view_size - 1);
+}
+
+template <class T>
+void InclusiveScan(benchmark::State& state,
+                   const Kokkos::View<T*, exec_space>& in,
+                   Kokkos::View<T*, exec_space>& out) {
```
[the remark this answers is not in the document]
science-enthusiast: Good point about constness of views.

science-enthusiast:
https://github.com/kokkos/kokkos/pull/7542#discussion_r1905121379

Could this comment be clarified?

algorithms/perf_test/CMakeLists.txt
science-enthusiast: Which cmake directory?

science-enthusiast: Yes, it is a good idea to move these common functions to the root directory and make `core/perf_test`, `algorithms/perf_test` and the perf tests in the future, to refer to these common functions. This is better than making `core/perf_test` as the source of reference.

algorithms/perf_test/test_inclusive_scan.cpp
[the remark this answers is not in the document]
science-enthusiast: Good point about manual timing and passing the measured time to `SetIterationTime` etc. I noticed this being done in some of the codes in `core/perf_test` and then I forgot about it. Since we are using `Kokkos::Timer`, properly implementing manual timing is the better approach.

algorithms/perf_test/test_inclusive_scan.cpp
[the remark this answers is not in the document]
science-enthusiast: @tpadioleau In practice, there is a correction regarding this. @dalg24 was also mentioning it. When a function template parameter has a View with const element type (i.e, Kokkos::View<const T*, ...>), it is not possible to pass a View with a non-const element type (e.g., Kokkos::View<double *, ...>). The compiler is not able to find a match. I tried this with the function `obtainLastElem`, since, the function doesn't modify the elements of the View and faced this error.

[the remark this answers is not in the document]
tpadioleau:
```suggestion
                   const Kokkos::View<T*, exec_space>& in,
                   const Kokkos::View<T*, exec_space>& out) {
```

You are right the compiler cannot both make a conversion and a deduction of the template parameter. I suggest we only make the object `const`.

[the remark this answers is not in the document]
science-enthusiast: I will clarify my thoughts further. I have heard about not performing conversion and deduction at the same time, in the context of multiple function parameters with the same template parameter type. e.g., if one argument is `double` and the other is `int`, while both function parameters are of type `T`, a matching function will not be found. I couldn't understand well enough, how not performing conversion and deduction at the same time causes the problem in our case. For example, on compiler explorer, I tried a simple function template with a `const T` parameter. It could find a match when a non-const `double` was passed. Why is it not working out with a Kokkos View?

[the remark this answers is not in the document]
masterleinad: https://godbolt.org/z/eorvfcrbn would be a simple example.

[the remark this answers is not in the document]
science-enthusiast: Thank you for the illustrative example.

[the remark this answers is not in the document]
science-enthusiast: There is this article: [Coercing deep const-ness](https://brevzin.github.io/c++/2021/09/10/deep-const/)

He mentions how exception has been made to the rule of no conversion during deduction only when it comes to pointer/reference to const. For class templates like `std::span`, the rule still applies. We are facing the same issue with `Kokkos::View`. He describes how one could possibly overcome this problem. Most likely, Kokkos is already making efforts in this direction.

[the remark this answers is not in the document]
tpadioleau: While I understand the problem in the language I am fine with the "completely useless" `takes_generic_const_span`. The user is responsible to pass a `span<const int>`, using `views::as_const` if necessary.

algorithms/perf_test/test_inclusive_scan.cpp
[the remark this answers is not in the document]
science-enthusiast: I missed this change in [e7ca1e4](https://github.com/kokkos/kokkos/commit/e7ca1e490c8ce319f75e3e7d26c3fa05d4862cd7). I would like to make this change and will do it soon.

algorithms/perf_test/CMakeLists.txt
science-enthusiast: A FIX_ME has been added to keep track of this issue.

algorithms/perf_test/CMakeLists.txt
science-enthusiast: As suggested by @dalg24, refactoring such common logic should be handled in a separate PR. An issue has been created: [7670](https://github.com/kokkos/kokkos/issues/7670). Also, FIX_MEs have been added in the relevant places.

algorithms/perf_test/CMakeLists.txt
masterleinad: Just try to remove these guards and see if it fails in the CI.

masterleinad (changes requested):
I still don't see new performance results.

science-enthusiast:
> I still don't see new performance results.

Yes, I am on it. I have numbers for CUDA. I need to access a super computer (Adastra) to run on an AMD GPU. I have been facing some issues as a first time user of Adastra. But as of now, I have managed to build Kokkos for HIP backend (but some warnings have come. Hopefully, it should not be a problem). I will be submitting a SLURM script soon.

science-enthusiast:
Common settings:
Input size: 100000000
min_warmup_time: 0.1 seconds
time measured using manual_time (i.e., Kokkos::Timer)

The time resolution is set as seconds in the custom benchmark main function in `core/perf_test/BenchmarkMain.cpp`. Thus there is a loss in significant digits in the time. A FIXME has been added to see whether this can be removed in a future PR.

With Nvidia V100:
----------------------------------------------------------------------------------------------------------------------------------------------------
Benchmark                                                        Time          CPU      Iterations      UserCounters...
----------------------------------------------------------------------------------------------------------------------------------------------------
Test_InclusiveScan<std::uint64_t>                0.003 s      0.002 s      234 FOM: GB/s=636.203/s MB=1.6k Passed=1
Test_InclusiveScan<std::int64_t>                  0.002 s       0.002 s      247 FOM: GB/s=654.122/s MB=1.6k Passed=1
Test_InclusiveScan<double>                         0.003 s        0.002 s     264 FOM: GB/s=514.215/s MB=1.6k Passed=1
Test_InclusiveScan<std::uint64_t, SumFunctor>  0.003 s  0.002 s  286 FOM: GB/s=613.193/s MB=1.6k Passed=1
Test_InclusiveScan<std::int64_t, SumFunctor>  0.003 s  0.002 s  260 FOM: GB/s=596.954/s MB=1.6k Passed=1
Test_InclusiveScan<double, SumFunctor>         0.003 s  0.002 s  232 FOM: GB/s=618.953/s MB=1.6k Passed=1
Test_InclusiveScan<std::uint64_t, MaxFunctor>  0.003 s  0.002 s  279 FOM: GB/s=607.155/s MB=1.6k Passed=1
Test_InclusiveScan<std::int64_t, MaxFunctor>    0.002 s  0.002 s  213 FOM: GB/s=645.614/s MB=1.6k Passed=1
Test_InclusiveScan<double, MaxFunctor>           0.003 s  0.002 s  209 FOM: GB/s=500.297/s MB=1.6k Passed=1

With AMD MI300A:
----------------------------------------------------------------------------------------------------------------------------------------------------
Benchmark                                                        Time          CPU      Iterations      UserCounters...
----------------------------------------------------------------------------------------------------------------------------------------------------
Test_InclusiveScan<std::uint64_t>                  0.001 s     0.001 s       692 FOM: GB/s=1.58853k/s MB=1.6k Passed=1
Test_InclusiveScan<std::int64_t>                    0.001 s     0.001 s       691 FOM: GB/s=1.58809k/s MB=1.6k Passed=1
Test_InclusiveScan<double>                           0.001 s     0.001 s       692 FOM: GB/s=1.59124k/s MB=1.6k Passed=1
Test_InclusiveScan<std::uint64_t, SumFunctor>  0.001 s  0.001 s  692 FOM: GB/s=1.58757k/s MB=1.6k Passed=1
Test_InclusiveScan<std::int64_t, SumFunctor>    0.001 s  0.001 s  691 FOM: GB/s=1.58822k/s MB=1.6k Passed=1
Test_InclusiveScan<double, SumFunctor>           0.001 s  0.001 s  688 FOM: GB/s=1.59125k/s MB=1.6k Passed=1
Test_InclusiveScan<std::uint64_t, MaxFunctor>  0.001 s  0.001 s  692 FOM: GB/s=1.59091k/s MB=1.6k Passed=1
Test_InclusiveScan<std::int64_t, MaxFunctor>    0.001 s  0.001 s  693 FOM: GB/s=1.59023k/s MB=1.6k Passed=1
Test_InclusiveScan<double, MaxFunctor>           0.001 s  0.001 s  691 FOM: GB/s=1.5864k/s MB=1.6k Passed=1

CC: @masterleinad

science-enthusiast:
Common settings:
Input size: 100000000
min_warmup_time: 0.1 seconds
time measured using manual_time (i.e., Kokkos::Timer)

The time resolution is set as seconds in the custom benchmark main function in `core/perf_test/BenchmarkMain.cpp`. Thus there is a loss in significant digits in the time. A FIXME has been added to see whether this can be removed in a future PR.

With Nvidia V100:
Benchmark$`~~~~~~~~~~~~~~~~~~~~~~~~~`$Time$`~~~~~~~~`$CPU$`~~~`$Iterations$`~~~~~~`$UserCounters...
std::uint64_t$`~~~~~~~~~~~~~~~~~~~~`$0.003 s$`~~~~~~`$0.002 s$`~~~`$234$`~`$FOM: GB/s=636.203/s MB=1.6k Passed=1
std::int64_t$`~~~~~~~~~~~~~~~~~~~~`$0.002 s$`~~~~~~~`$0.002 s$`~~~`$247$`~`$FOM: GB/s=654.122/s MB=1.6k Passed=1
double$`~~~~~~~~~~~~~~~~~~~~~~~~~`$0.003 s$`~~~~~~`$0.002 s$`~~~`$264$`~`$FOM: GB/s=514.215/s MB=1.6k Passed=1
std::uint64_t, SumFunctor$`~~`$0.003 s$`~~`$0.002 s$`~~~`$286$`~`$FOM: GB/s=613.193/s MB=1.6k Passed=1
std::int64_t, SumFunctor$`~~`$0.003 s$`~~`$0.002 s$`~~~`$260$`~`$FOM: GB/s=596.954/s MB=1.6k Passed=1
double, SumFunctor$`~~~~~~~~~`$0.003 s$`~~`$0.002 s$`~~~`$232$`~`$FOM: GB/s=618.953/s MB=1.6k Passed=1
std::uint64_t, MaxFunctor$`~~`$0.003 s$`~~`$0.002 s$`~~~`$279$`~`$FOM: GB/s=607.155/s MB=1.6k Passed=1
std::int64_t, MaxFunctor$`~~~~`$0.002 s$`~~`$0.002 s$`~~~`$213$`~`$FOM: GB/s=645.614/s MB=1.6k Passed=1
double, MaxFunctor$`~~~~~~~~~~~`$0.003 s$`~~`$0.002 s$`~~~`$209$`~`$FOM: GB/s=500.297/s MB=1.6k Passed=1

With AMD MI300A:
Benchmark$`~~~~~~~~~~~~~~~~~~~~~~~~~`$Time$`~~~~~~~~`$CPU$`~~~`$Iterations$`~~~~~~`$UserCounters...
std::uint64_t$`~~~~~~~~~~~~~~~~~~~~`$0.001 s$`~~~~~~`$0.001 s$`~~~`$692$`~`$FOM: GB/s=1.58853k/s MB=1.6k Passed=1
std::int64_t$`~~~~~~~~~~~~~~~~~~~~`$0.001 s$`~~~~~~`$0.001 s$`~~~`$691$`~`$FOM: GB/s=1.58809k/s MB=1.6k Passed=1
double$`~~~~~~~~~~~~~~~~~~~~~~~~~`$0.001 s$`~~~~~~`$0.001 s$`~~~`$692$`~`$FOM: GB/s=1.59124k/s MB=1.6k Passed=1
std::uint64_t, SumFunctor$`~~`$0.001 s$`~`$0.001 s$`~~~`$692$`~`$FOM: GB/s=1.58757k/s MB=1.6k Passed=1
std::int64_t, SumFunctor$`~~`$0.001 s$`~`$0.001 s$`~~~`$691$`~`$FOM: GB/s=1.58822k/s MB=1.6k Passed=1
double, SumFunctor$`~~~~~~~~~`$0.001 s$`~`$0.001 s$`~~~`$688$`~`$FOM: GB/s=1.59125k/s MB=1.6k Passed=1
std::uint64_t, MaxFunctor$`~~`$0.001 s$`~`$0.001 s$`~~~`$692$`~`$FOM: GB/s=1.59091k/s MB=1.6k Passed=1
std::int64_t, MaxFunctor$`~~~~`$0.001 s$`~`$0.001 s$`~~~`$693$`~`$FOM: GB/s=1.59023k/s MB=1.6k Passed=1
double, MaxFunctor$`~~~~~~~~~~~`$0.001 s$`~`$0.001 s$`~~~`$691$`~`$FOM: GB/s=1.5864k/s MB=1.6k Passed=1
CC: @masterleinad

masterleinad:
What do the results look like before this pull request?

algorithms/perf_test/CMakeLists.txt
masterleinad: ping

science-enthusiast:
If the following check is removed inside `CMakeLists.txt` within `algorithms/perf_test`, then there are failures in CI.

```
if((KOKKOS_ENABLE_OPENMPTARGET OR KOKKOS_ENABLE_OPENACC) AND KOKKOS_CXX_COMPILER_ID STREQUAL NVHPC)
   return()
endif()

```
The failure cases are:
linux-x64 / CI (fedora:latest, clang++, Debug, OPENMP, -DCMAKE_CXX_CLANG_TIDY="clang-tidy;-warnings-as-errors...
linux-x64 / CI (fedora:latest, clang++, Release, OPENMP, -DCMAKE_CXX_CLANG_TIDY="clang-tidy;-warnings-as-erro...
linux-x64 / CI (fedora:latest, g++, Debug, OPENMP, -DCMAKE_CXX_CLANG_TIDY="clang-tidy;-warnings-as-errors=*",...
linux-x64 / CI (fedora:latest, g++, Release, OPENMP, -DCMAKE_CXX_CLANG_TIDY="clang-tidy;-warnings-as-errors=*...

The four failures happen even after re-running them. All the four failures are of the same nature and concern Google Benchmark.

Capture d’écran du 2025-01-17 13-05-59
The text from all the four cases are below for reference:
[clang++_debug_openmp.txt](https://github.com/user-attachments/files/18454352/clang%2B%2B_debug_openmp.txt)
[clang++_release_openmp.txt](https://github.com/user-attachments/files/18454354/clang%2B%2B_release_openmp.txt)
[g++_debug_openmp.txt](https://github.com/user-attachments/files/18454355/g%2B%2B_debug_openmp.txt)
[g++_release_openmp.txt](https://github.com/user-attachments/files/18454356/g%2B%2B_release_openmp.txt)

masterleinad:
> If the following check is removed inside `CMakeLists.txt` within `algorithms/perf_test`, then there are failures in CI.

That doesn't make sense. None of these builds use `nvhpc` as the compiler.

science-enthusiast:
> > If the following check is removed inside `CMakeLists.txt` within `algorithms/perf_test`, then there are failures in CI.
> 
> That doesn't make sense. None of these builds use `nvhpc` as the compiler.

Yes, I also noticed that. The CI is running now with the check (KOKKOS_ENABLE_OPENACC AND KOKKOS_CXX_COMPILER_ID STREQUAL Clang) removed. Let us see what happens. Then I will also run CI with both the checks in place, as before and see whether CI passes as expected.

masterleinad:
Note that the jenkins build was fine and that's the only one using `nvhpc` (as host compiler). `KOKKOS_ENABLE_OPENACC AND KOKKOS_CXX_COMPILER_ID STREQUAL Clang)` isn't tested in the CI. I would just remove all the checks and get a verbpse log if the problem in the `GitHub` CI persists.

masterleinad:
The fedora containers has a recent update, see https://github.com/kokkos/ci-containers/pkgs/container/ci-containers%2Ffedora. I woulnd't be surprised if that's the problem.

science-enthusiast:
> Note that the jenkins build was fine and that's the only one using `nvhpc` (as host compiler). `KOKKOS_ENABLE_OPENACC AND KOKKOS_CXX_COMPILER_ID STREQUAL Clang)` isn't tested in the CI. I would just remove all the checks and get a verbpse log if the problem in the `GitHub` CI persists.

Thanks for your message. I noticed `nvhpc` in one of the cases within the Jenkins build but I didn't know that it was the only place. Noted your point about "OpenACC + Clang" being not tested. I will see how to get a verbose log. I think I saw that option when I re-ran the tests. I didn't enable it. Let me check.

science-enthusiast:
The same errors associated with the Fedora containers are happening when the check for (KOKKOS_ENABLE_OPENACC AND KOKKOS_CXX_COMPILER_ID STREQUAL Clang) is removed. I re-ran the tests after enabling `debug logging`. I don't see any extra information. I dowloaded the logs and checked also. I cannot find any extra information.

As such the error is clang-tidy related, which results in an error when warnings as errors is enabled. When checking online for "use c++14 style type templates", other people are also facing this issue. Not clear why it is a problem for us with these Fedora containers.

algorithms/perf_test/CMakeLists.txt
```diff
@@ -0,0 +1,69 @@
...
+# inclusive_scan and other future tests?
+# FIXME_OPENMPTARGET - the NVIDIA HPC compiler nvc++ in the OpenMPTarget backend does not pass the perf_tests.
+# FIXME_OPENACC - temporarily disabled due to unimplemented features
+if((KOKKOS_ENABLE_OPENMPTARGET OR KOKKOS_ENABLE_OPENACC) AND KOKKOS_CXX_COMPILER_ID STREQUAL NVHPC)
+  return()
+endif()
+# if(KOKKOS_ENABLE_OPENACC AND KOKKOS_CXX_COMPILER_ID STREQUAL Clang)
+#   return()
+# endif()
+
```
masterleinad, line 11:
```suggestion
```

science-enthusiast: I will remove these checks in the PR.

masterleinad (approved):
FIne with me.

algorithms/perf_test/test_inclusive_scan.cpp
```diff
@@ -0,0 +1,211 @@
...
+
+  auto h_in = Kokkos::create_mirror_view(in);
+
+  for (std::size_t i = 0; i < kProbSize; ++i) {
+    h_in(i) = i;
+  }
+
+  Kokkos::deep_copy(in, h_in);
+
+  return std::make_tuple(in, out, h_in);
```
Rombur, line 74: Why are you returning `h_in`? You are not using it later on.

science-enthusiast: Thanks for your review. I did that with the intention to pass it on to the serialScan function. Then I have forgotten about it. I wrote serialScan which accepts a view in the default execution space. I will retain serialScan as such and modify this function to return only in and out.

science-enthusiast: The function serialScan has been updated.

Rombur (approved):

tpadioleau (approved):
lgtm

science-enthusiast:
Retest this please!

science-enthusiast:
Retest this please!

algorithms/src/std_algorithms/impl/Kokkos_InclusiveScan.hpp
```diff
@@ -101,9 +147,62 @@ struct InclusiveScanDefaultFunctor {
...
+    const std::string& label, const Cuda& ex, InputIteratorType first_from,
+    InputIteratorType last_from, OutputIteratorType first_dest) {
+  const auto thrust_ex = thrust::cuda::par.on(ex.cuda_stream());
+
+  std::uint64_t kpID = 0;
+  if (Kokkos::Profiling::profileLibraryLoaded()) {
+    Kokkos::Profiling::beginParallelScan(
+        label + " via thrust::inclusive_scan",
+        Kokkos::Profiling::Experimental::device_id(ex), &kpID);
+  }
```
dalg24, line 166: Why did you opt for this versus `{push,pop}Region()`?

algorithms/perf_test/CMakeLists.txt
```diff
@@ -0,0 +1,58 @@
+# FIXME: The following logic should be moved from here and also from `core/perf_test/CMakeLists.txt` to
+# the root `CMakeLists.txt` in the form of a macro
+# Find or download google/benchmark library
```
dalg24, line 3: OK
If we haven't already done so please open an issue to keep track of this reorg.

algorithms/perf_test/test_inclusive_scan.cpp
```diff
@@ -0,0 +1,207 @@
...
+  Kokkos::View<T*, ExecSpace> in{"input", kProbSize};
+  Kokkos::View<T*, ExecSpace> out{"output", kProbSize};
+
+  auto h_in = Kokkos::create_mirror_view(in);
+
+  for (std::size_t i = 0; i < kProbSize; ++i) {
+    h_in(i) = i;
+  }
+
+  Kokkos::deep_copy(in, h_in);
```
dalg24, line 73: Why not initialize these directly on the device?

algorithms/perf_test/test_inclusive_scan.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+  T res_scan  = obtainLastElem(out);
+  bool passed = (res_check == res_scan);
+
+  return std::make_tuple(time_scan, passed);
+}
+
+// Benchmark: Default (prefix sum) scan
+template <class T>
+static void Test_InclusiveScan(benchmark::State& state) {
```
dalg24, line 134: Why static?

algorithms/perf_test/test_inclusive_scan.cpp
```diff
@@ -0,0 +1,207 @@
...
+    const auto [time_scan, passed] =
+        InclusiveScan<T, ScanFunctor<T>>(in, out, res_check, scan_functor);
+
+    KokkosBenchmark::report_results(state, in, 2, time_scan);
+    state.counters["Passed"] = passed;
+  }
+}
+
+static constexpr std::size_t PROB_SIZE = 100'000'000;
+static constexpr double WARMUP_TIME    = 0.1;  // warm up of 100 milli-seconds
```
dalg24, line 170: If we want to avoid symbol collisions, the right thing to do is probably just define everything here within the anonymous namespace like we do for tests.

algorithms/perf_test/test_inclusive_scan.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+BENCHMARK(Test_InclusiveScan<std::uint64_t>)
+    ->Arg(PROB_SIZE)
+    ->MinWarmUpTime(WARMUP_TIME)
```
dalg24, line 174: Is that something specific to that benchmark or something we would like to set globally?

algorithms/src/std_algorithms/impl/Kokkos_InclusiveScan.hpp
science-enthusiast: I will discuss with @masterleinad about this.

algorithms/perf_test/test_inclusive_scan.cpp
```diff
@@ -0,0 +1,207 @@
...
+             ScanFunctor scan_functor = ScanFunctor()) {
+  std::size_t view_size = h_in.extent(0);
+
+  Kokkos::View<T*, HostExecSpace> h_out("output", view_size);
+
+  h_out(0) = h_in(0);
+
+  for (std::size_t i = 1; i < view_size; ++i) {
+    h_out(i) = scan_functor(h_in(i), h_out(i - 1));
+  }
```
dalg24, line 90: You could have use `std::inclusive_scan` (not blocking)

algorithms/perf_test/test_inclusive_scan.cpp
dalg24: If you add a FIXME comment saying that need to be revisited and should be apply wholesale to the full exe I am ok with merging and fixing as a follow up

algorithms/perf_test/CMakeLists.txt
```diff
@@ -0,0 +1,58 @@
...
+  # they can be used by all performance tests.
+  list(APPEND BENCHMARK_SOURCES ../../core/perf_test/BenchmarkMain.cpp ../../core/perf_test/Benchmark_Context.cpp)
+
+  add_executable(${BENCHMARK_NAME} ${BENCHMARK_SOURCES})
+  target_link_libraries(${BENCHMARK_NAME} PRIVATE benchmark::benchmark Kokkos::kokkos impl_git_version)
+  target_include_directories(${BENCHMARK_NAME} SYSTEM PRIVATE ${benchmark_SOURCE_DIR}/include)
+
+  foreach(SOURCE_FILE ${BENCHMARK_SOURCES})
+    set_source_files_properties(${SOURCE_FILE} PROPERTIES LANGUAGE ${KOKKOS_COMPILE_LANGUAGE})
+  endforeach()
```
JBludau, line 50: This alone will not work ... as it might need an architecture and standard which need to be defined on target level. But this could be marked with a FIXME as it will potentially go away with https://github.com/kokkos/kokkos/pull/7582

algorithms/perf_test/CMakeLists.txt
```diff
@@ -0,0 +1,58 @@
...
+  endif()
+
+  target_compile_options(benchmark PRIVATE -w)
+  target_compile_options(benchmark_main PRIVATE -w)
+endif()
+
+# FIXME: This function should be moved from here and also from `core/perf_test/CMakeLists.txt` to
+# the root `CMakeLists.txt`
+function(KOKKOS_ADD_BENCHMARK NAME)
+  cmake_parse_arguments(BENCHMARK "" "" "SOURCES" ${ARGN})
```
JBludau, line 34: can it be called without any args?

algorithms/perf_test/CMakeLists.txt
```diff
@@ -0,0 +1,58 @@
...
+  FetchContent_MakeAvailable(googlebenchmark)
+  list(POP_BACK CMAKE_MESSAGE_INDENT)
+
+  # Suppress clang-tidy diagnostics on code that we do not have control over
+  if(CMAKE_CXX_CLANG_TIDY)
+    set_target_properties(benchmark PROPERTIES CXX_CLANG_TIDY "")
+  endif()
+
+  target_compile_options(benchmark PRIVATE -w)
+  target_compile_options(benchmark_main PRIVATE -w)
```
JBludau, line 28: why do we need this?

algorithms/perf_test/CMakeLists.txt
```diff
@@ -0,0 +1,58 @@
...
+endif()
+
+# FIXME: This function should be moved from here and also from `core/perf_test/CMakeLists.txt` to
+# the root `CMakeLists.txt`
+function(KOKKOS_ADD_BENCHMARK NAME)
```
JBludau, line 33: why is the name an extra arg and not in the `cmake_parse_arguments`

algorithms/perf_test/test_inclusive_scan.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+  Kokkos::deep_copy(in, h_in);
+
+  return std::make_tuple(in, out, h_in);
+}
+
+// Perform scan in a serial manner. Used as reference.
+template <class T, class ScanFunctor = SumFunctor<T>>
+T serialScan(const Kokkos::View<T*, HostExecSpace>& h_in,
```
JBludau, line 80: this naming does not follow your convention ... I would prefer following the convention we use in the rest of Kokkos

algorithms/perf_test/test_inclusive_scan.cpp
```diff
@@ -0,0 +1,207 @@
...
+  KE::inclusive_scan("Default scan", ExecSpace(), KE::cbegin(in), KE::cend(in),
+                     KE::begin(out));
+  ExecSpace().fence();
+  double time_scan = timer.seconds();
+
+  T res_scan  = obtainLastElem(out);
+  bool passed = (res_scan == res_check);
+
+  return std::make_tuple(time_scan, passed);
+}
```
JBludau, line 110: could you unify this with the non-default below? At the moment this is repeating a lot of code

algorithms/perf_test/test_inclusive_scan.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+  T res_check = serialScan<T>(h_in);
+
+  for (auto _ : state) {
+    const auto [time_scan, passed] = InclusiveScan<T>(in, out, res_check);
+
+    KokkosBenchmark::report_results(state, in, 2, time_scan);
+    state.counters["Passed"] = passed;
+  }
+}
```
JBludau, line 147: can you unify the two tests vs repeating most of the code?

algorithms/src/std_algorithms/impl/Kokkos_InclusiveScan.hpp
```diff
@@ -24,6 +24,52 @@
...
+// Some versions of Clang fail to compile Thrust, failing with errors like
+// this:
+//    <snip>/thrust/system/cuda/detail/core/agent_launcher.h:557:11:
+//    error: use of undeclared identifier 'va_printf'
+// The exact combination of versions for Clang and Thrust (or CUDA) for this
+// failure was not investigated, however even very recent version combination
+// (Clang 10.0.0 and Cuda 10.0) demonstrated failure.
+//
+// Defining _CubLog here locally allows us to avoid that code path, however
+// disabling some debugging diagnostics
```
JBludau, line 50: I have a hard time understanding this comment

core/perf_test/BenchmarkMain.cpp
```diff
@@ -24,6 +24,7 @@
 int main(int argc, char** argv) {
   Kokkos::initialize(argc, argv);
   benchmark::Initialize(&argc, argv);
+  // FIXME: seconds as default time unit leads to precision loss
```
JBludau, line 27: could you explain why you do it?

algorithms/perf_test/CMakeLists.txt
masterleinad: This file is copied from https://github.com/kokkos/kokkos/blob/develop/core/perf_test/CMakeLists.txt.

algorithms/src/std_algorithms/impl/Kokkos_InclusiveScan.hpp
masterleinad: That's pre-existing in the sortimpl and sortbykeyimpl files.

core/perf_test/BenchmarkMain.cpp
masterleinad: This is pre-existing in `core`.

algorithms/perf_test/test_inclusive_scan.cpp
science-enthusiast: Thank you for your review comments. I also thought about the repeating codes.

One solution that I tried was a variadic template based benchmark function. In the [User Guide](https://github.com/google/benchmark/blob/main/docs/user_guide.md), there are two different ways in which benchmark function templates can be defined.
1. [variadic templates](https://github.com/google/benchmark/blob/main/docs/user_guide.md#passing-arbitrary-arguments-to-a-benchmark) that are used to pass arbitrary arguments to the benchmark function.
2. [function templates](https://github.com/google/benchmark/blob/main/docs/user_guide.md#templated-benchmarks) that are not variadic in nature, through which benchmarks for different types could be defined etc.

In my case, I need a variadic function template with a fixed first argument that is the type of the elements in the view. Google Benchmark is not allowing me to register such a function.

One other solution that I could think of was to pass a flag that indicates whether it is the default case or the case that accepts a functor. However, in this case there will always be a functor that also needs to be passed around. This functor will not be used in the default sum case. This didn't look like a nice solution.

core/perf_test/BenchmarkMain.cpp
science-enthusiast: Thank you for clarifying, @masterleinad.

I noticed that the output times have just one or two significant digits due to representing them in seconds. The FIXME has been added to address this problem in the future. When addressing issue [7670](https://github.com/kokkos/kokkos/issues/7670), this file (core/perf_test/BenchmarkMain.cpp) needs to be moved to a common place. At that time, this precision issue could also be addressed.

algorithms/perf_test/CMakeLists.txt
science-enthusiast: I will document possible question/issues with the logic in this CMakeLists.txt in [7670](https://github.com/kokkos/kokkos/issues/7670).

algorithms/perf_test/test_inclusive_scan.cpp
science-enthusiast: I will change the name to ref_scan and call `std::inclusive_scan` inside.

algorithms/perf_test/test_inclusive_scan.cpp
JBludau: So you can not use ?
```
template<typename T, typename Functor = void>
Test_InclusiveScan(...)
{
if constexpr (std::is_same_v<Functor,void>)
  //default case
else
 //special case
}
```

algorithms/perf_test/CMakeLists.txt
science-enthusiast: It is being tracked in issue [7670](https://github.com/kokkos/kokkos/issues/7670).

algorithms/perf_test/test_inclusive_scan.cpp
science-enthusiast: The input array is needed both on the device (where it is passed to Kokkos::inclusive_scan) and on the host (where it is passed to a reference implementation of inclusive_scan). Thus there is both a `h_in` and `h_out`.

algorithms/perf_test/CMakeLists.txt
science-enthusiast: It can be called with just one argument that corresponds to NAME.

JBludau: but this would not really create anything helpful would it?

algorithms/perf_test/CMakeLists.txt
JBludau: Did you copy this from somewhere?
How it is now it does nothing for `CXX` and not enough for any other language.

masterleinad: The whole file is copied from https://github.com/kokkos/kokkos/blob/develop/core/perf_test/CMakeLists.txt

JBludau: ok, in this case I am fine with the FIXME

algorithms/perf_test/CMakeLists.txt
science-enthusiast: I have added a FIXME to consider making NAME a one_value_argument for cmake_parse_arguments.

algorithms/perf_test/CMakeLists.txt
science-enthusiast: :)

algorithms/perf_test/CMakeLists.txt
science-enthusiast: A FIXME has been added to consider this.

algorithms/perf_test/test_inclusive_scan.cpp
science-enthusiast: A FIXME comment has been added.

algorithms/perf_test/test_inclusive_scan.cpp
science-enthusiast: The anonymous namespace based approach has been implemented.

algorithms/perf_test/test_inclusive_scan.cpp
science-enthusiast: As noted below, the anonymous namespace based approach has been implemented now.

algorithms/src/std_algorithms/impl/Kokkos_InclusiveScan.hpp
science-enthusiast: The call to `inclusive_scan` is now enclosed by a `pushRegion()` and a `popRegion()`.

algorithms/perf_test/test_inclusive_scan.cpp
science-enthusiast: `std::inclusive_scan` is not available in GCC 8.4.0. The Jenkins CI has a GCC 8.4.0 based check. A detailed FIXME has been added.

algorithms/perf_test/test_inclusive_scan.cpp
science-enthusiast: Please take a look at the updated code.

science-enthusiast:
Retest this please!

algorithms/src/std_algorithms/impl/Kokkos_InclusiveScan.hpp
```diff
@@ -101,9 +148,48 @@ struct InclusiveScanDefaultFunctor {
...
+#endif
+
+#if defined(KOKKOS_ENABLE_ROCTHRUST)
+template <class InputIteratorType, class OutputIteratorType>
+OutputIteratorType inclusive_scan_default_op_exespace_impl(
+    const std::string& label, const HIP& ex, InputIteratorType first_from,
+    InputIteratorType last_from, OutputIteratorType first_dest) {
+  const auto thrust_ex = thrust::hip::par.on(ex.hip_stream());
+
+  Kokkos::Profiling::pushRegion(label + " via thrust::inclusive_scan");
```
masterleinad, line 181: If you use this approach, please add the regions to all the other paths `inclusive_scan` could take.

science-enthusiast: I see your point. However, the path is a call to a function in the Thrust library. I am not sure what can be done in this case.

Or is there some other approach for profiling that is more suitable here?

masterleinad: If we don't have a special path, you would just do
```
Kokkos::Profiling::pushRegion(label + " via Kokkos::parallel_scan");
```
making sure that `label` contains `Kokkos::inclusive_scan`.

science-enthusiast: Thank you for your quick reply. Do you mean:

`Kokkos::Profiling::pushRegion(label + " via Kokkos::inclusive_scan");`

Since this `inclusive_scan` function doesn't call `Kokkos::parallel_scan`.

masterleinad: No, I meant to add a corresponding call to the other code paths that are using `Kokkos::parallel_scan`.

science-enthusiast: Thanks. I see your point better. I will make the change and let you know.

algorithms/perf_test/test_inclusive_scan.cpp
```diff
@@ -0,0 +1,191 @@
...
+constexpr std::size_t PROB_SIZE = 100'000'000;
+
+}  // anonymous namespace
+
+// FIXME: Add logic to pass min. warm-up time. Also, the value should be set
+// by the user. Say, via the environment variable BENCHMARK_MIN_WARMUP_TIME.
+
+BENCHMARK(BM_inclusive_scan<std::uint64_t>)->Arg(PROB_SIZE)->UseManualTime();
+BENCHMARK(BM_inclusive_scan<std::int64_t>)->Arg(PROB_SIZE)->UseManualTime();
+BENCHMARK(BM_inclusive_scan<double>)->Arg(PROB_SIZE)->UseManualTime();
```
JBludau, line 173: Do we also want to benchmark this for `float`, `half` or `complex`?

science-enthusiast: I thought about `float`. The output type, in the case of operations like sum, may have to be of higher precision. I could add a FIXME about revisiting later about further types like `float`, `half` and `complex`. There is a lot of work that could be done with regards to benchmark testing of Kokkos algorithms. We could start in a small way and build on it.

JBludau: Sounds good to me. I would prefer an issue rather than a fixme as it might require some discussion

algorithms/src/std_algorithms/impl/Kokkos_InclusiveScan.hpp
science-enthusiast: Could you have a look at the updated code?

masterleinad: It would if you wouldn't squash all the commits so it's easier to review changes to the pull request. 🙂

science-enthusiast: Sorry about that. I will try to do it differently going forward. The changes are regarding the addition of push and pop regions to all the inclusive_scan functions.

science-enthusiast: I also thought that since the changes are small, it will not be difficult for the reviewer.

masterleinad:
```
/builds/kokkos/kokkos/algorithms/unit_tests/../src/std_algorithms/impl/Kokkos_InclusiveScan.hpp:395:22: error: reference to __host__ function 'pushRegion' in __host__ __device__ function
  395 |   Kokkos::Profiling::pushRegion(
      |                      ^
/builds/kokkos/kokkos/algorithms/unit_tests/../src/std_algorithms/Kokkos_InclusiveScan.hpp:252:16: note: called by 'inclusive_scan<Kokkos::Impl::HIPTeamMember, Kokkos::Experimental::Impl::RandomAccessIterator<Kokkos::View<const double *, Kokkos::LayoutStride, Kokkos::Device<Kokkos::HIP, Kokkos::HIPSpace>, Kokkos::Experimental::EmptyViewHooks, Kokkos::MemoryTraits<0>>>, Kokkos::Experimental::Impl::RandomAccessIterator<Kokkos::View<double *, Kokkos::LayoutStride, Kokkos::Device<Kokkos::HIP, Kokkos::HIPSpace>, Kokkos::MemoryTraits<0>>>, 0>'
  252 |   return Impl::inclusive_scan_default_op_team_impl(teamHandle, first, last,
      |                ^
/builds/kokkos/kokkos/algorithms/unit_tests/TestStdAlgorithmsTeamInclusiveScan.cpp:74:26: note: called by 'operator()<Kokkos::Impl::HIPTeamMember>'
   74 |         auto it    = KE::inclusive_scan(member, first, last, firstDest);
      |                          ^
/builds/kokkos/kokkos/core/src/HIP/Kokkos_HIP_ParallelFor_Team.hpp:64:5: note: called by 'exec_team<void>'
   64 |     m_functor(member);
      |     ^
/builds/kokkos/kokkos/core/src/HIP/Kokkos_HIP_ParallelFor_Team.hpp:89:22: note: called by 'operator()'
   89 |       this->template exec_team<work_tag>(typename Policy::member_type(
      |                      ^
/builds/kokkos/kokkos/core/src/HIP/Kokkos_HIP_KernelLaunch.hpp:76:3: note: called by 'hip_parallel_launch_local_memory<Kokkos::Impl::ParallelFor<Test::stdalgos::TeamInclusiveScan::TestFunctorA<Kokkos::View<double **, Kokkos::HIPSpace>, Kokkos::View<double **>, Kokkos::View<unsigned long *>, Kokkos::View<bool *>, Kokkos::View<double *, Kokkos::LayoutRight, Kokkos::HIP>, Test::stdalgos::TeamInclusiveScan::PlusFunctor<double>>, Kokkos::TeamPolicy<Kokkos::HIP>>>'
   76 |   driver();
      |   ^
/builds/kokkos/kokkos/core/src/impl/Kokkos_Profiling.hpp:95:6: note: 'pushRegion' declared here
   95 | void pushRegion(const std::string& kName);
      |      ^
```

algorithms/src/std_algorithms/impl/Kokkos_InclusiveScan.hpp
```diff
@@ -247,10 +391,16 @@ KOKKOS_FUNCTION OutputIteratorType inclusive_scan_default_op_team_impl(
...
+
+  Kokkos::Profiling::pushRegion(
+      "Kokkos::inclusive_scan via TeamPolicy and Kokkos::parallel_scan");
+
   ::Kokkos::parallel_scan(TeamThreadRange(teamHandle, 0, num_elements),
                           func_type(first_from, first_dest));
   teamHandle.team_barrier();
 
+  Kokkos::Profiling::popRegion();
+
```
masterleinad, line 403: Revert the changes to the team implementation.

science-enthusiast: Sorry, I should have caught this.

science-enthusiast: I would like to know how you got the output of the compilation error that you pasted above. I am able to get that error output by building the code locally. I don't see it in the logs of the failed CIs.

masterleinad: You squashed again which makes it much harder to find the results from the previous run. In any case, I saw this in the GitLab CI. For the last commit, the CI results can be found at https://gitlab.spack.io/kokkos/kokkos/-/pipelines/980661.

science-enthusiast: I found this article: https://jacobtomlinson.dev/posts/2022/dont-prematurely-squash/rebase-and-force-push-your-prs/

I suppose Kokkos is taking the same approach.

science-enthusiast: I also see the merit in making the process easier for the reviewers.

science-enthusiast:
@dalg24 
Please take a look at the changes made. I had already commented in the individual conversations about the changes. They are summarized below:

* Profiling is supported now via push and pop Regions. All the code paths have been handled. 
* Anonymous namespace has been used to enforce internal linkage.
* A FIXME has been added to accept warm-up time as a user defined parameter (say, as an environment variable).
* `std::span` is not supporteed by GCC 8.4.0 (Jenkins CI). Thus the reference scan implementation is still implemented using a for loop. A FIXME has been added.

algorithms/perf_test/test_inclusive_scan.cpp
```diff
@@ -0,0 +1,191 @@
...
+    h_out(i) = scan_functor(h_in(i), h_out(i - 1));
+  }
+
+  return h_out(view_size - 1);
+}
+
+// Inclusive Scan with default binary operation (sum) or user provided functor
+// Note: The nature of the functor must be compatible with the
+// elements in the input and output views
+template <class T, template <class> class ScanFunctor = ImpSumBinOp>
```
JBludau, line 109: Since you never use the default for `ScanFunctor` I would remove it to make it more expressive. (Non-blocking)

science-enthusiast: The default type `ImpSumBinOp` for `ScanFunctor` is used in this check `if constexpr (std::is_same_v<ScanFunctor<T>, ImpSumBinOp<T>>)` when only the parameter `T` is specified when calling `inclusive_scan`.

JBludau: Oh, my bad. You actually use this in line 171-173. I just looked at the later invocations

dalg24:
Retest this please

JBludau:
@science-enthusiast can you open a pr adding documentation for this, please?

masterleinad:
> Do we have a general policy of spelling out when we forward to vendor implementations? That is basically where I was coming from

We only list TPLs as options in the documentation but don't document in which cases interfaces might use them (which are implementation details IMHO).

JBludau:
Ok, then this does not need to be documented

PR #7550, opened 2024-11-26
Add support for Zen 4 AMD microarchitecture

dalg24:
**Not tested**

Zen 4 was introduced by AMD a couple years ago and is supported on recent compilers (as far as I can tell since 
nvc++ 23.9,
Clang 16.0,
and GCC 12.3)

This is intended to support El Capitan.

@masterleinad As far as for testing it looks like Frank cluster at UO might have a couple systems with Epyc Genoa CPUs so we could look into this (preferably as a follow up)

@ldh4 it looks like Zen4 supports AVX-512 but I wasn't sure what to do about it.


masterleinad:
> @ldh4 it looks like Zen4 supports AVX-512 but I wasn't sure what to do about it.

I can confirm that AVX512 was autodetected for MI300A which seems to be a Zen4.

masterleinad:
> @masterleinad As far as for testing it looks like Frank cluster at UO might have a couple systems with Epyc Genoa CPUs so we could look into this (preferably as a follow up)

or we enable it for the MI300A build.

cmake/kokkos_arch.cmake
```diff
@@ -412,6 +413,24 @@ if(KOKKOS_ARCH_ZEN3)
...
 endif()
 
+if(KOKKOS_ARCH_ZEN4)
+  compiler_specific_flags(
+    COMPILER_ID
+    KOKKOS_CXX_HOST_COMPILER_ID
+    Intel
+    -mavx2
+    MSVC
+    /arch:AVX2
```
masterleinad, line 423:
```suggestion
    Intel
    -mavx512f
    MSVC
    /arch:AVX512
```

cmake/kokkos_arch.cmake
```diff
@@ -412,6 +413,24 @@ if(KOKKOS_ARCH_ZEN3)
...
+    NVHPC
+    -tp=zen4
+    DEFAULT
+    -march=znver4
+    -mtune=znver4
+  )
+  set(KOKKOS_ARCH_AMD_ZEN4 ON)
+  set(KOKKOS_ARCH_AVX2 ON)
```
masterleinad, line 431:
```suggestion
  set(KOKKOS_ARCH_AVX512 ON)
```

Makefile.kokkos
```diff
@@ -474,7 +477,7 @@ KOKKOS_INTERNAL_USE_ARCH_AMD := $(shell expr $(KOKKOS_INTERNAL_USE_ARCH_AMD_GFX9
 
 # Any AVX?
 KOKKOS_INTERNAL_USE_ARCH_AVX        := $(shell expr $(KOKKOS_INTERNAL_USE_ARCH_SNB) + $(KOKKOS_INTERNAL_USE_ARCH_AMDAVX))
-KOKKOS_INTERNAL_USE_ARCH_AVX2       := $(shell expr $(KOKKOS_INTERNAL_USE_ARCH_HSW) + $(KOKKOS_INTERNAL_USE_ARCH_BDW) + $(KOKKOS_INTERNAL_USE_ARCH_ZEN) + $(KOKKOS_INTERNAL_USE_ARCH_ZEN2) + $(KOKKOS_INTERNAL_USE_ARCH_ZEN3))
+KOKKOS_INTERNAL_USE_ARCH_AVX2       := $(shell expr $(KOKKOS_INTERNAL_USE_ARCH_HSW) + $(KOKKOS_INTERNAL_USE_ARCH_BDW) + $(KOKKOS_INTERNAL_USE_ARCH_ZEN) + $(KOKKOS_INTERNAL_USE_ARCH_ZEN2) + $(KOKKOS_INTERNAL_USE_ARCH_ZEN3) + $(KOKKOS_INTERNAL_USE_ARCH_ZEN4))
```
masterleinad, line 480: I would just omit changing the Makefile for AVX* here.

cmake/kokkos_arch.cmake
dalg24: That would have no effect.  There is no such macro.

masterleinad: We are using `AVX512XEON` pretty much for all architectures and that's the guard used for `SIMD`.

dalg24: Sure, the variable being set in your suggestion is still wrong as far as I can tell (I am on the phone though)

masterleinad: > Sure, the variable being set in your suggestion is still wrong as far as I can tell (I am on the phone though)

Updated.

ldh4: It does feel like we should perhaps consolidate or refactor AVX512 macros. There are currently 
 `KOKKOS_ARCH_AVX512MIC` and `KOKKOS_ARCH_AVX512XEON`, but `KOKKOS_ARCH_AVX512MIC` is currently not in use anywhere.

Makefile.kokkos
ldh4: I think ZEN4 should be omitted here or should go under another variable (i.e. KOKKOS_INTERNAL_USE_ARCH_AVX512).

Makefile.kokkos
```diff
@@ -845,6 +848,19 @@ ifeq ($(KOKKOS_INTERNAL_USE_ARCH_ZEN3), 1)
   endif
 endif
 
+ifeq ($(KOKKOS_INTERNAL_USE_ARCH_ZEN4), 1)
+  tmp := $(call kokkos_append_header,"$H""define KOKKOS_ARCH_AMD_ZEN4")
+  tmp := $(call kokkos_append_header,"$H""define KOKKOS_ARCH_AVX2")
+
+  ifeq ($(KOKKOS_INTERNAL_COMPILER_INTEL), 1)
+    KOKKOS_CXXFLAGS += -mavx2
+    KOKKOS_LDFLAGS += -mavx2
```
ldh4, line 857: Should these be `-xCORE-AVX512` instead?

cmake/kokkos_arch.cmake
masterleinad: I agree and just rename all of them to `KOKKOS_IMPL_AVX512` or so.

Makefile.kokkos
dalg24: I suggest we do the refactor later when we rename the macro

cmake/kokkos_arch.cmake
dalg24: Let's do the renaming as a follow up

Makefile.kokkos
dalg24: done

masterleinad (approved):

JBludau (approved):
looks good

dalg24:
Retest this please

dalg24:
Retest this please

